PR #12913, opened 2021-07-19
Add boo#1188444 to list of known bugs

ggardet:
- Verification run: https://openqa.opensuse.org/t1847427

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -45,6 +45,13 @@
         },
         "type": "bug"
     },
+    "bsc#1188444": {
+        "description": "Bootloader left irq enabled: bank 1 irq 9",
+        "products": {
+            "opensuse": ["15.2"]
+        },
+        "type": "bug"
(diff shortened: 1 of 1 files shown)


data/journal_check/bug_refs.json
```diff
@@ -45,6 +45,13 @@
         },
         "type": "bug"
     },
+    "bsc#1188444": {
+        "description": "[Firmware Bug]: Bootloader left irq enabled: bank 1 irq 9",
```
Vogtinator, line 49: Needs regex escaping, but let's just avoid having to deal with that part:

```suggestion
        "description": "Bootloader left irq enabled: bank 1 irq 9",
```

Vogtinator (changes requested):

data/journal_check/bug_refs.json
ggardet: Updated

Vogtinator (approved):

PR #12916, opened 2021-07-19
Extend AutoYaST file(s) preparation for Rules and Classes

jknphy:
- Related ticket: [poo#95392](https://progress.opensuse.org/issues/95392)
- Verification run: [autoyast_mini](https://openqa.suse.de/tests/overview?build=jknphy%2Fos-autoinst-distri-opensuse%23expand_prepare_profile&version=15-SP3&distri=sle)

Note: it would require some adjustment in follow-up tasks: [checking of existence of profile/second stage ay](https://openqa.suse.de/tests/6483301#step/installation/1)

### DIFF
--- a/tests/autoyast/prepare_profile.pm
+++ b/tests/autoyast/prepare_profile.pm
@@ -7,13 +7,8 @@
 # notice and this notice are preserved.  This file is offered as-is,
 # without any warranty.
 
-# Summary: Expand variables in the autoyast profiles and make it accessible for SUT
-#
-# - Get profile from autoyast template
-# - Map version names
-# - Get IP address from system variables
-# - Get values from SCC_REGCODE SCC_REGCODE_HA SCC_REGCODE_GEO SCC_REGCODE_HPC SCC_URL ARCH LOADER_TYPE
-# - Modify profile with obtained values and upload new autoyast profile
+# Summary: Prepare AutoYaST xml profile by expanding variables
+# before installation and setting correct URL for the installer.
 # Maintainer: QE YaST <qa-sle-yast@suse.de>
 
 use strict;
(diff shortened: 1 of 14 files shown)


tests/autoyast/prepare_profile.pm
```diff
@@ -26,25 +26,41 @@ use autoyast qw(
...
-    die $profile                         if $profile->isa('Mojo::Exception');
-
-    $profile = expand_version($profile);
-    $profile = adjust_network_conf($profile);
-    $profile = expand_variables($profile);
-    upload_profile(profile => $profile, path => $path);
+    my $files;
+    my $ay_path = get_required_var('AUTOYAST');
+
+    if ($ay_path =~ /\/$/) {
```
OleksandrOrlov, line 36: As far as I understand, here we rely only on the last "/" in the path. So, case `AUTOYAST=path/to/folder` will not work. Maybe it is better to check if the path is a directory?

Something, like

```perl
if ( -d $ay_path ) {
    # When using rules and classes, multiple xml files will be processed
    $files = get_test_data_files($ay_path);
}
```

jknphy: I has to be written like that, is in the YaST documentation with the last '/', so if someone try that we might find some better way to handle, but it should only read one file and fail, here is the doc: https://documentation.suse.com/sles/15-SP3/html/SLES-all/rulesandclass.html

OleksandrOrlov: Oh, I see. Could you please highlight that this slash in the end is important? At least a comment, but better a check with meaningful error message. Because otherwise it may confuse even us in future.

Right now if `AUTOYAST=path/to/file` will be provided, the code execution will go to `else` block and will try to find a file with that name. It may be hard to debug in future.

jknphy: that is exactly how is supposed to be, to point to a file and if it is the wrong one some error is expected (but of course that requires some knowledge of ay and features rules and classes), but I got your point, the only way to give an error on on particular path is to fix expectation as we just get a string, and the path is evaluated later in different ways. I wanted to be a bit conservative as the code is not new at old, it has been with us for years, but at the end I could separate in two files and reuse the common part in the loop and give the user/tester meaningful error when the settings is not as expected.

OleksandrOrlov: I thought it is simple and expected we can do something like this:

```perl
elsif  ( -d $ay_path && $ay_path !~ /\/$/) {
   die "Folder path is used without "/" in the end. Please add "/" in case you want to use rules in AutoYaST"
}
```

Will that work?

jknphy: no, as I mentioned before it cannot work, because the $ay_path is a relative path that is resolved later to its full path, so -d will always tell us that doesn't exists. The worker is checking with Perl using -d if the file exists but at that point is just an string unprocessed, is another way to say it.

okurz:
Seems like you triggered problems in SLE maintenance tests, see https://progress.opensuse.org/issues/95804 , e.g. tests https://openqa.suse.de/tests/6508464

PR #12919, opened 2021-07-19
Soft-fail vsftpd module in s390x

punkioudi:
Because of this [bug](https://bugzilla.suse.com/show_bug.cgi?id=1176813), the scenarios with the anonymous user case, will be excluded from the testsuite only for s390x run, where the issue occurs. 
The reason that the `run_s390x.sh` was created, is to keep the rest of the testsuite running. When the issue will be solved, then the condition for the s390x and the `run_s390x.sh` will be removed. 

- Related ticket: https://progress.opensuse.org/issues/93270
- Needles: N/A
- Verification run: 15SP3 [s390x](https://openqa.suse.de/tests/6493400#step/vsftpd/31) | [x86_64](https://openqa.suse.de/tests/6493393#step/vsftpd/30)


tests/console/vsftpd.pm
```diff
@@ -33,6 +33,7 @@ use utils 'zypper_call';
 sub run {
     my $self = shift;
     $self->select_serial_terminal;
+    my $arch = get_var('ARCH');
```
foursixnine, line 36: if you import `Utils::Architectures 'is_s390x';` you can get rid of this variable

tests/console/vsftpd.pm
```diff
@@ -45,7 +46,13 @@ sub run {
     assert_script_run 'cd /tmp';
     assert_script_run 'wget ' . data_url('qam/vsftpd.tar.gz');
     assert_script_run 'tar xzfv vsftpd.tar.gz';
-    assert_script_run 'bash run.sh |& tee run.log', 300;
+    # Remove the cases with anonymous user in run_s390x because of bsc#1176813
+    if ($arch eq 's390x') {
+        record_soft_failure 'bsc#1176813 - vsftpd: security: one_process_model needs a better OS for the anonymous user scenarios';
+        assert_script_run 'bash run_s390x.sh |& tee run.log', 300;
```
foursixnine, line 52: I doubt this will allow the next test to continue and would still mark them as failed, maybe script_run is a better choice, and reverse the condition:

soft fail if the runsh fails, die if the rush didn't fail (so we know that the test is passing, and can remove the soft failure).

Also I see that you've changed something in the data file, but do not explain the change anywhere that's visible.

punkioudi: I see what you say, but I wanted to keep the scenarios of the test suite that pass, to continue running. These changes are located in the vsftpd tarball, which will be soon visible in this PR (I have marked it as WIP ;) )

dzedro (approved):

foursixnine (approved):

PR #12921, opened 2021-07-19
Validate beta popup

OleksandrOrlov:
Required changes to support BETA popup while using `select_product` test module.

- Verification run: https://openqa.suse.de/tests/6484522

### DIFF
--- a/lib/Installation/Warnings/WarningsController.pm
+++ b/lib/Installation/Warnings/WarningsController.pm
@@ -1,0 +1,46 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved. This file is offered as-is,
+# without any warranty.
+
+# Summary: The class introduces business actions for Warning Popups
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::Warnings::WarningsController;
+use strict;
+use warnings;
+use YuiRestClient;
+use Installation::Warnings::ConfirmationWarningRichText;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {}, $class;
+    return $self->init();
+}
+
+sub init {
+    my ($self) = @_;
+    $self->{ConfirmationWarningRichText} = Installation::Warnings::ConfirmationWarningRichText->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_confirmation_warning_rich_text {
+    my ($self) = @_;
+    return $self->{ConfirmationWarningRichText};
+}
+
+sub get_text {
+    my ($self) = @_;
+    $self->get_confirmation_warning_rich_text->text();
+}
+
+sub accept_warning {
+    my ($self) = @_;
+    $self->get_confirmation_warning_rich_text->press_ok();
+}
+
+1;
--- a/schedule/staging/ext4@s390x-staging.yaml
+++ b/schedule/staging/ext4@s390x-staging.yaml
@@ -10,6 +10,7 @@ vars:
 schedule:
   - installation/bootloader_start
   - installation/setup_libyui
+  - installation/validate_beta_popup
   - installation/product_selection/select_product
   - installation/licensing/accept_license
   - installation/scc_registration
--- a/tests/installation/validate_beta_popup.pm
+++ b/tests/installation/validate_beta_popup.pm
@@ -1,0 +1,31 @@
+# Copyright (C) 2021 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+# Summary: Test module to validate beta popup and accept it.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use Test::Assert ':all';
+
+sub run {
+    my $warnings = $testapi::distri->get_warnings_controller();
+    assert_matches(qr/You are accessing our Beta Distribution/, $warnings->get_text(),
+        "Warning Dialog for BETA did not appear, while it was expected.");
+    $warnings->accept_warning();
+}
+
+1;
--- a/lib/Distribution/Opensuse/Tumbleweed.pm
+++ b/lib/Distribution/Opensuse/Tumbleweed.pm
@@ -31,6 +31,7 @@ use Installation::Partitioner::LibstorageNG::v4_3::GuidedSetupController;
 use Installation::Partitioner::LibstorageNG::GuidedSetupController;
 use Installation::Partitioner::LibstorageNG::v4_3::ExpertPartitionerController;
 use Installation::SystemRole::SystemRoleController;
+use Installation::Warnings::WarningsController;
 use YaST::Bootloader::BootloaderController;
 use YaST::Firstboot::ConfigurationCompletedController;
 use YaST::Firstboot::HostNameController;
@@ -135,6 +136,10 @@ sub get_bootloader {
 
 sub get_license_agreement {
     return Installation::License::Opensuse::LicenseAgreementController->new();
+}
+
+sub get_warnings_controller {
+    return Installation::Warnings::WarningsController->new();
 }
 
 1;
--- a/lib/Installation/Warnings/ConfirmationWarningRichText.pm
+++ b/lib/Installation/Warnings/ConfirmationWarningRichText.pm
@@ -20,12 +20,18 @@ sub init {
     my $self = shift;
     $self->SUPER::init();
     $self->{rt_warning} = $self->{app}->label({type => 'YRichText'});
+    $self->{btn_ok}     = $self->{app}->button({id => 'ok'});
     return $self;
 }
 
 sub text {
     my ($self) = @_;
     return $self->{rt_warning}->text();
+}
+
+sub press_ok {
+    my ($self) = @_;
+    $self->{btn_ok}->click();
 }
 
 1;
--- a/schedule/staging/minimal+base@s390x-staging.yaml
+++ b/schedule/staging/minimal+base@s390x-staging.yaml
@@ -12,6 +12,7 @@ vars:
 schedule:
   - installation/bootloader_start
   - installation/setup_libyui
+  - installation/validate_beta_popup
   - installation/product_selection/select_product
   - installation/licensing/accept_license
   - installation/scc_registration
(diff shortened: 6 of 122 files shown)


lib/Installation/Warnings/WarningsController.pm
```diff
@@ -0,0 +1,49 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved. This file is offered as-is,
+# without any warranty.
+
+# Summary: The class introduces business actions for Warning Popups
```
sofiasyria, line 10: I don't agree with using the word "business" everywhere. It can be confusing for people who are not so familiar with testing theory. Plus, if I google "business action" and "software testing" together, I don't get a clear definition.

tests/installation/validate_beta_popup.pm
```diff
@@ -0,0 +1,31 @@
...
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use Test::Assert ':all';
+
+sub run {
+    my $warnings = $testapi::distri->get_warnings_controller();
+    assert_matches(qr/You are accessing our Beta Distribution/, $warnings->get_warning_rich_text(),
+        "Warning Dialog for 'BIOS boot partition has too small size to install bootloader' did not appear, while it is expected.");
```
sofiasyria, line 27: Grammatically, it's better to use the same tense: "did not appear, while it was expected"

schedule/staging/ext4@s390x-staging.yaml
```diff
@@ -34,6 +35,11 @@ schedule:
   - console/validate_blockdevices
   - console/validate_free_space
   - console/validate_read_write
+conditional_schedule:
+  validate_beta_popup:
+    BETA:
+      1:
```
sofiasyria, line 41: Maybe it would make it less complicated to add the condition inside the module, instead of turning every schedule into a conditional one.

sofiasyria:
Did you add the correct VR links?

lib/Installation/Warnings/WarningsController.pm
OleksandrOrlov: Thanks for the review and thanks for the point!;)

First of all, reading or adding new tests using [the approach](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/ui-framework-documentation.md), assumes a person got familiar with it, e.g. by reading our documentation where it is mentioned and described (Please, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/ui-framework-documentation.md#context section). 

Also, if you will google not just software testing but something like `"business actions" in automation testing`, you get description in several places.

If you think that we cover that part not enough in the documentation, feel free to raise that point as a ticket or update the documentation so that it will be more clear.

schedule/staging/ext4@s390x-staging.yaml
OleksandrOrlov: I was also thinking about that and it is for sure rather simple solution.

But at the same time it breaks our strategy of not having conditional logic that changes behavior of the test. We are in progress of refactoring our tests to be atomic and to remove extra knowledge about product (i.e. do not have steps that do not belong to the test case that is covered in perl test module). The condition will hide the steps performed for beta popup handling inside of the module, so that we'll not be able to distinguish where we cover that test case without reviewing logs. 

This is one of the things that we try to achieve: make test suites coverage transparent and delegate knowledge of what should be tested to test suite schedule.

I also agree that having the same conditional logic in all the files is not that good from DRY prospective, but this is the drawback of the decision we made some time ago: duplication over complexity in test schedule. We also started discussions on how that could be improved, so that we'll not have so many places to change, once something that is needed for all the test suites is added.

tests/installation/validate_beta_popup.pm
OleksandrOrlov: Good point! It is copy-paste from another test module. Updated.

OleksandrOrlov:
> Did you add the correct VR links?

Yes. Probably failures confused you. 

Main goal of these VRs is to show that `validate_beta_popup` is scheduled when BETA=1 and not scheduled when BETA=0. Also for the case when BETA=1, VR shows that test passes on this test module.

The failures on `scc_registration` does not belong to the change.

schedule/staging/ext4@s390x-staging.yaml
OleksandrOrlov: > Maybe it would make it less complicated to add the condition inside the module

One more thing, that I forgot to mention, is that the popup does not appear on Product Selection Page. So, technically, it does not belong to Product Select Page and cannot be handled by `select_product` test module.

lib/Installation/Warnings/WarningsController.pm
```diff
@@ -0,0 +1,49 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::Warnings::WarningsController;
+use strict;
+use warnings;
+
+use strict;
```
jknphy, line 17: two imports of `strict`

lib/Installation/Warnings/WarningsController.pm
```diff
@@ -0,0 +1,49 @@
...
+    return $self->init();
+}
+
+sub init {
+    my ($self) = @_;
+    $self->{ConfirmationWarningRichText} = Installation::Warnings::ConfirmationWarningRichText->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_confirmation_warning_rich_text {
```
jknphy, line 34: the caller might not care about if it is rich text, or you plan to use in future `get_confirmation_warning_text` ? Could be just get_text(), but it will be answer above about the reason to have this controller.

lib/Installation/Warnings/WarningsController.pm
```diff
@@ -0,0 +1,49 @@
...
+package Installation::Warnings::WarningsController;
+use strict;
+use warnings;
+
+use strict;
+use warnings;
+use YuiRestClient;
+use Installation::Warnings::ConfirmationWarningRichText;
+
+sub new {
```
jknphy, line 22: do we really need this controller? We could have some controller named for example yast_initialization or installer_initialization (or similar) which would use this warning, because we always use this as part of the test.
what about this way? https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/7d6fdcdc3b9b6917689b61e29856998af8830cdb/tests/installation/authentication/root_simple_pwd.pm#L27
Because I'm not sure how we guarantee that the warning is there, does it work in other archs?

schedule/staging/ext4@s390x-staging.yaml
jknphy: I also 50/50 with that, I think we have a whole new life cycle of the product to figure out better solution, I would just go with adding the test unconditionally, instead of having more mess in every yaml and just create a ticket to work on it.

On the other hand I like that you make the problem more visible and not hide it, so can encourage us to find soon some solution. 
Ideally we should not depend on any openQA settings, so we can just have some notification before it would happen at the end of product life cycle (that is feasible) and enable in single place. It could also helps if some intermediate build doesn't have and we want to continue testing (but I'm not sure if that case make sense...as the build could be invalid). So that is the thing, that you prepare it for not having Beta as well and that might not needed for now.

In summary, I would prefer to have just the test unconditionally, but I'm fine also with this solution.

jknphy (commented):
I add some feedback about the structure, but nothing that would stop us to merge as we need to continue testing further steps.

schedule/staging/ext4@s390x-staging.yaml
sofiasyria: But this particular solution would be very costly if applied in more occasions. I do not propose to merge it with select_product module. I would prefer to keep it separate and the worker would just check a condition at the beginning of the module and do nothing in case of BETA=0. It's not ideal indeed, but keeps the schedules clean.

sofiasyria: Maybe we can add some extra feature in scheduling for conditions manipulated by vars.

jknphy: > But this particular solution would be very costly if applied in more occasions. I do not propose to merge it with select_product module. I would prefer to keep it separate and the worker would just check a condition at the beginning of the module and do nothing in case of BETA=0. It's not ideal indeed, but keeps the schedules clean.

long history of conditions in modules make me disagree :) I still would prefer this solution (despite some of my comments) to follow with the design that we are currently working on.

jknphy: > Maybe we can add some extra feature in scheduling for conditions manipulated by vars.

that will be done in future, but now cannot be seen the pattern, we need first to have complete installation to do it.

foursixnine: One thing is that we also have the GMC flag, which is something of the future, but will also need to happen. In any case, does the merging of yaml schedule suport conditional schedules? if so all that could ble handled within one and then imported as yast installation base or something, from what I see, it's always the same three lines.

OleksandrOrlov: @foursixnine Yeah, this is one of the solutions we are right now thinking about. We'll address this in another ticket, as right now we don't have some well-established structure for inheritance in `schedule` section of YAML schedule, and we want to think about the proper design first.

Currently this change will affect only YaST Job Group where we'll be able to distinguish all pros and cons of the solution, then we'll be able to apply it for test suites that use installation modules in another Job Groups.

So, in the end, we discussed in the team and agreed that the simplest solution in terms of adapting it in future for new schedule design, will be scheduling of `validate_beta_warning` in all the tests (in YaST job group), as soon as this warning will be with us for a long period of time. Until it will disappear we'll have some solution how to manage it in a convenient way.

lib/Installation/Warnings/WarningsController.pm
OleksandrOrlov: As soon as we are using the term `business actions` all over the project, I guess this is not a topic of this particular PR. In case you see any better term, we can discuss this and plan activities to rename it everywhere.

lib/Installation/Warnings/WarningsController.pm
OleksandrOrlov: Agree. I used the approach taken from another place in our framework, but as we identify better approaches constantly, I would say, having just `get_text()` is the proper way.

In the past it was not so clear on how it will affect us while having different warning popups, so we used this approach with separating them. Now we identified we should go towards unifying things that are same from the business logic of product. So, warning is warning, client shouldn't care about it's internals.

lib/Installation/Warnings/WarningsController.pm
OleksandrOrlov: I would go with this controller, as I would like to experiment with this approach. It is similar to what we have in NavigationController, which seems to be working good and saved us a lot of potential duplication.

Here I see two improvements:
1. When popup is shown, LibyuiClient see it as the topmost page and does not see the page under it. So, we anyways can guarantee that the warning is there only by getting some its elements (e.g. label or warning text). So, with the approach of having WarningController, nothing will be changed on this matter.
2. Warnings may appear on any page, also there may be plenty of variants of the warning. e.g. like we have for Create User Page. We can have only `warning->get_text()` and then compare the received text with the expected one, provided in test module. So, no duplication, everything related to warnings will be in one controller, like we have for NavigationController.

jknphy (approved):

schedule/staging/ext4@s390x-staging.yaml
foursixnine: @OleksandrOrlov awesome, please if you remember... keep me in the loop :D

PR #12883, opened 2021-07-13
Make handle_login to take care of reaching the desktop

GraceWang571:
Extract the code of reaching the desktop into handle_login
Handle welcome screen in handle_login including gnome 40 activities
Remove the codes that handle assert_screen 'generic-desktop' and
gnome40 activities in specific test cases

- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/738
- Verification run: 
- Test suites contain cases that invoke wait_boot_past_bootloader, handle_login and handle_relogin are verified.
- create_hdd_gnome-wayland (contains first_boot which invokes wait_boot_past_bootloader ) on TW https://openqa.opensuse.org/tests/1838019

- kde-wayland@64bit_virtio (contains first_boot which invokes wait_boot_past_bootloader and start_wayland_plasma5 which invokes handle_login) on TW https://openqa.opensuse.org/tests/1838021

- extra_tests_gome (contains multi_users_dm which invokes handle_login and vino_screensharing_available which contains handle_relogin) on TW https://openqa.opensuse.org/tests/1838016

- gnome-x11-ibus (contails ibus_installation which invokes handle_relogin) on TW  https://openqa.opensuse.org/tests/1838017

- desktopapps-gnome (contains application_starts_on_login which invokes handle_relogin) on TW: https://openqa.opensuse.org/tests/1838018
on SLE: https://openqa.suse.de/tests/6424500

- desktopapps-remote-client1 (contains onetime_vncsession_xvnc_remmina, onetime_vncsession_multilogin_failed and onetime_vncsession_xvnc_tigervnc which invokes handle_login) on SLE: https://openqa.suse.de/tests/6424524

- desktopapps-remote-client2 (contains persistent_vncsession_xvnc which invokes handle_login) on SLE: https://openqa.suse.de/tests/6424930

- test case user_gui_login only contais handle_relogin, so no verification run provided.
- test case boot_rt_kernel contains handle_login but no verification run provided.

PR #12922, opened 2021-07-19
lib/opensusebasetest.pm: Handle gnome-activities in wait_boot_past_bootloader

Vogtinator:
Missing in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12883.

Verification run: https://openqa.opensuse.org/tests/1847753

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -1024,6 +1024,7 @@ sub wait_boot_past_bootloader {
 
     my @tags = qw(generic-desktop emergency-shell emergency-mode);
     push(@tags, 'opensuse-welcome') if opensuse_welcome_applicable;
+    push(@tags, 'gnome-activities') if check_var('DESKTOP', 'gnome');
 
     # boo#1102563 - autologin fails on aarch64 with GNOME on current Tumbleweed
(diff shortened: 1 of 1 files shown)


lib/opensusebasetest.pm
```diff
@@ -1050,7 +1051,10 @@ sub wait_boot_past_bootloader {
     # Starting with GNOME 40, upon login, the activities screen is open (assuming the
     # user will want to start something. For openQA, we simply press 'esc' to close
     # it again and really end up on the desktop
-    send_key('esc') if match_has_tag('gnome-activities');
+    if (match_has_tag('gnome-activities')) {
+        send_key 'esc';
+        check_screen 'generic-desktop', $check_interval;
```
DimStar77, line 1056: as you already noted: in case we close activities with opensuse-welcome open, we fail here now

Vogtinator: Needs some perl fu to filter out `gnome-activities` from `@tags`

DimStar77: `@tags = grep {!/gnome-activites/} @tags;`

DimStar77:
Moved to #12924 - (as Fabian had to leave)

PR #12924, opened 2021-07-19
lib/opensusebasetest.pm: Handle gnome-activities in wait_boot_past_bootloader

DimStar77:
Missing in #12883 - continuation of #12922 (Fabian had to run away)

- Verification run: https://openqa.opensuse.org/tests/1847840

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -1024,6 +1024,7 @@ sub wait_boot_past_bootloader {
 
     my @tags = qw(generic-desktop emergency-shell emergency-mode);
     push(@tags, 'opensuse-welcome') if opensuse_welcome_applicable;
+    push(@tags, 'gnome-activities') if check_var('DESKTOP', 'gnome');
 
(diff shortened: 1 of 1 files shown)


DimStar77:
There most be more things missing even with this change.

E.g finish_desktop does not cope with gnome-activities

GraceWang571:
LGTM.
Thanks for finding and fixing the omission. I will fix the finish_desktop failure caused by #12883 later.
I will also check if there's any other omissions left.

DimStar77:
> LGTM.
> Thanks for finding and fixing the omission. I will fix the finish_desktop failure caused by #12883 later.
> I will also check if there's any other omissions left.

Thanks!

Another one found: https://openqa.opensuse.org/tests/1847940#step/reboot_gnome/24 (reboot_gnome)

GraceWang571:
https://openqa.opensuse.org/tests/1847098#step/multi_users_dm/18 this would also fail.
gnome-activities need to be handled in ensure_unlocked_desktop as well.
I am working on the fix.

PR #12925, opened 2021-07-19
Add test modules for timezone page using LibyuiClient

OleksandrOrlov:
The PR adds the appropriate test modules for Clock And Timezone Page and updates YAML schedule files accordingly.

- Related ticket: https://progress.opensuse.org/issues/94889
- Verification runs: 
   - openSUSE: https://openqa.opensuse.org/tests/1847869
   - SLE: https://openqa.suse.de/tests/6486078

### DIFF
--- a/schedule/yast/installer_extended/installer_extended.yaml
+++ b/schedule/yast/installer_extended/installer_extended.yaml
@@ -27,7 +27,7 @@ schedule:
   - installation/system_role/select_role
   - installation/partitioning/accept_proposed_layout
   - installation/releasenotes
-  - installation/installer_timezone
+  - installation/clock_and_timezone/accept_timezone_configuration
   - installation/authentication/default_user_simple_pwd
   - installation/authentication/root_simple_pwd
   - installation/resolve_dependency_issues
--- a/tests/installation/clock_and_timezone/validate_timezone_configuration.pm
+++ b/tests/installation/clock_and_timezone/validate_timezone_configuration.pm
@@ -7,24 +7,24 @@
 # notice and this notice are preserved.  This file is offered as-is,
 # without any warranty.
 
-# Summary: Handles Time and Date dialog in YaST Firstboot Configuration.
-#
+# Summary: Collects all the info about current Timezone configuration
+# and validates it against the one provided by test_data.
 # Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
 
-use base 'y2_firstboot_basetest';
+use base 'y2_installbase';
 use strict;
 use warnings;
 use scheduler 'get_test_suite_data';
 use cfg_files_utils;
 
 sub run {
-    my $test_data           = get_test_suite_data()->{time_and_date};
+    my $test_data = get_test_suite_data()->{time_and_date};
+
     my $clock_and_time_zone = $testapi::distri->get_clock_and_time_zone();
     compare_settings({
             expected      => $test_data,
             current       => $clock_and_time_zone->collect_current_clock_and_time_zone_info(),
             suppress_info => 1});
-    $clock_and_time_zone->proceed_with_current_configuration();
 }
 
 1;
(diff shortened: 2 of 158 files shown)


schedule/yast/installer_extended/installer_extended.yaml
```diff
@@ -27,7 +27,8 @@ schedule:
   - installation/system_role/select_role
   - installation/partitioning/accept_proposed_layout
   - installation/releasenotes
-  - installation/installer_timezone
+  - installation/clock_and_timezone/validate_timezone_configuration
```
jknphy, line 30: I think we should have a dedicated test suite for this to check defaults, I can see this installer_extended for doing some more actions, clicking there, open other windows, etc. Also it would need to be commented in the header of this yaml and make the other installer_extended yamls consistent, plus adding for it test data, my recommendation would be to not added here.

tests/installation/clock_and_timezone/validate_timezone_configuration.pm
```diff
@@ -7,24 +7,24 @@
...
 use warnings;
 use scheduler 'get_test_suite_data';
 use cfg_files_utils;
 
 sub run {
-    my $test_data           = get_test_suite_data()->{time_and_date};
-    my $clock_and_time_zone = $testapi::distri->get_clock_and_time_zone();
+    my $test_data = get_test_suite_data()->{time_and_date};
+
+    my $clock_and_time_zone = $testapi::distri->get_clock_and_time_zone()->get_clock_and_time_zone_page();
```
jknphy, line 23: it looks like you wanted to just do `$testapi::distri->get_clock_and_time_zone()->get_clock_and_time_zone_page();` to ensure to be in the page. But you still need a variable for the controller to run the method below `collect_current_clock_and_time_zone_info` as you are using the page which does not have that method. am I right?

OleksandrOrlov: Right! Thanks for the catch.

schedule/yast/installer_extended/installer_extended.yaml
OleksandrOrlov: ok, removed. We should revisit this `installer_extended`. Also, I do not like the name, because it says nothing about what it tests inside. Maybe we can make it more specific, because right now it is collector of the test modules that do not fit another test suites:)

tests/installation/clock_and_timezone/validate_timezone_configuration.pm
```diff
@@ -7,24 +7,24 @@
...
 sub run {
-    my $test_data           = get_test_suite_data()->{time_and_date};
+    my $test_data = get_test_suite_data()->{time_and_date};
+
     my $clock_and_time_zone = $testapi::distri->get_clock_and_time_zone();
     compare_settings({
             expected      => $test_data,
             current       => $clock_and_time_zone->collect_current_clock_and_time_zone_info(),
             suppress_info => 1});
-    $clock_and_time_zone->proceed_with_current_configuration();
```
jknphy, line 27: could you please delete after this change [proceed_with_current_configuration](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/dbd780e4b3df2278e28bc9e113803864382c3d8f/lib/Installation/ClockAndTimeZone/ClockAndTimeZoneController.pm#L47) [btn_next](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/dbd780e4b3df2278e28bc9e113803864382c3d8f/lib/Installation/ClockAndTimeZone/ClockAndTimeZonePage.pm#L27) and [press_next](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/dbd780e4b3df2278e28bc9e113803864382c3d8f/lib/Installation/ClockAndTimeZone/ClockAndTimeZonePage.pm#L54)

OleksandrOrlov: Done.

PR #12927, opened 2021-07-20
Add oracle linux guest profiles and installation files

waynechen55:
* **Add new oracle linux 6/7/8 guest parameters files:**
  * data/virt_autotest/guest_params_xml_files/oracle_linux_6_10_graphical_server_x86_64.xml
  * data/virt_autotest/guest_params_xml_files/oracle_linux_7_9_graphical_server_x86_64.xml
  * data/virt_autotest/guest_params_xml_files/oracle_linux_8_4_graphical_server_x86_64.xml
* **Add new oracle linux 6/7/8 guest unattended installations files:**
  * data/virt_autotest/guest_unattended_installation_files/oracle_linux_6_10_graphical_server.ks
  * data/virt_autotest/guest_unattended_installation_files/oracle_linux_7_9_graphical_server.ks
  * data/virt_autotest/guest_unattended_installation_files/oracle_linux_8_4_graphical_server.ks
* **Rename guest_installation_autoyast_file to guest_installation_automation_file in:**
  * data/virt_autotest/guest_params_xml_files/sles_15_sp3_64_kvm_hvm_uefi_with_power_management_x86_64.xml
  * data/virt_autotest/guest_params_xml_files/sles_15_sp3_64_kvm_hvm_uefi_without_power_management_x86_64.xml
  * data/virt_autotest/guest_params_xml_files/sles_15_sp3_64_xen_hvm_uefi_with_power_management_x86_64.xml
  * data/virt_autotest/guest_params_xml_files/sles_15_sp3_64_xen_hvm_uefi_without_power_management_x86_64.xml
* **Rename** data/virt_autotest/guest_autoyast_xml_files folder to data/virt_autotest/guest_unattended_installation_files

* **Verification runs:**
  * [oracle linux 6/7/8 pass](http://10.67.129.106/tests/1537)
  * [oracle linux 6/7/8 fail](http://10.67.129.106/tests/1510)
  * [sles guest pass](http://10.67.129.106/tests/1596)
  * [sles guest fail](10.67.129.106/tests/1597)
  * [oracle linux 6/7/8 pass with 15sp4](http://10.67.129.106/tests/1606)
  * [oracle linux 6/7/8 fail with 15sp4](http://10.67.129.106/tests/1614)
  * [sles guest pass with 15sp4](http://10.67.129.106/tests/1613)
  * [sles guest fail with 15sp4](http://10.67.129.106/tests/1616)


waynechen55:

nanzhg:
LGTM.

Julie-CAO:
LGTM. I need to change <guest_installation_autoyast_file> element name in TW param files as well

guoxuguang:
LGTM. Please update Verification runs on sles15sp4 will be better

alice-suse:
Can we oraganize all guest unattended installation files in a way like this?

data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}

waynechen55:
> Can we oraganize all guest unattended installation files in a way like this?
> 
> data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}

From my POV data/virt_autotest/guest_unattended_installation_files is enough. Additional layers just additional layers and additional changes to adapt these layers. To be honest, there is no use because "guest files" are not parsed based on layers(like the legacy qa lib virtauto test package), now they are given directly and conveniently. If "the layers as you said" was introduced, it is useless and only means you wrap "guest files" in a "folder" and pass it transparently to another entity which does not even care whether "the guest files" in one folder or another because there is no dependency on folder layers. There is no point to do this.

alice-suse:
> > Can we oraganize all guest unattended installation files in a way like this?
> > data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}
> 
> From my POV data/virt_autotest/guest_unattended_installation_files is enough. Additional layers just additional layers and additional changes to adapt these layers. To be honest, there is no use because "guest files" are not parsed based on layers(like the legacy qa lib virtauto test package), now they are given directly and conveniently. If "the layers as you said" was introduced, it is useless and only means you wrap "guest files" in a "folder" and pass it transparently to another entity which does not even care whether "the guest files" in one folder or another because there is no dependency on folder layers. There is no point to do this.

I think the point is to have better structured data. Nothing to do with how it is consumed.

waynechen55:
> > > Can we oraganize all guest unattended installation files in a way like this?
> > > data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}
> > 
> > 
> > From my POV data/virt_autotest/guest_unattended_installation_files is enough. Additional layers just additional layers and additional changes to adapt these layers. To be honest, there is no use because "guest files" are not parsed based on layers(like the legacy qa lib virtauto test package), now they are given directly and conveniently. If "the layers as you said" was introduced, it is useless and only means you wrap "guest files" in a "folder" and pass it transparently to another entity which does not even care whether "the guest files" in one folder or another because there is no dependency on folder layers. There is no point to do this.
> 
> I think the point is to have better structured data. Nothing to do with how it is consumed.

But there is no need of this structured data from my POV. It is not for you to see it as "structured data", actually it is already structured data because it is the place to store unattended installation files, there is no need to add extra folder structure. And those files in it can also tell what it is used for, let alone automation which does not even care. To be honest, you should change your angle to see it instead of sticking to "a rigid structured data". Additional folder layer does not mean clearer structure, it can also mean meaningless and redundant hierarchy. It is already structured data and it is enough to be clear. The point here is "your own definition for structured data" is not very sutiable here from my humble opinion.

alice-suse:
> > > > Can we oraganize all guest unattended installation files in a way like this?
> > > > data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}
> > > 
> > > 
> > > From my POV data/virt_autotest/guest_unattended_installation_files is enough. Additional layers just additional layers and additional changes to adapt these layers. To be honest, there is no use because "guest files" are not parsed based on layers(like the legacy qa lib virtauto test package), now they are given directly and conveniently. If "the layers as you said" was introduced, it is useless and only means you wrap "guest files" in a "folder" and pass it transparently to another entity which does not even care whether "the guest files" in one folder or another because there is no dependency on folder layers. There is no point to do this.
> > 
> > 
> > I think the point is to have better structured data. Nothing to do with how it is consumed.
> 
> But there is no need of "this structured data comes from addition folder layers" from my POV. It is not for you to see it as "structured data", actually it is already structured data because it is the place to store unattended installation files, there is no need to add extra folder structure, let alone all unattended installation files are unique and clear enough(they are not the "def" things in legacy qa lib virtauto test package, all "def" things are the same so they should be put in different folder to be unique. Such additional folder layer is not necessary here and useless because things are already unique and clear). And those files in it can also tell what it is used for, let alone automation which does not even care. To be honest, you should change your angle to see it instead of sticking to "a rigid structured data". Additional folder layer does not mean clearer structure, it can also mean meaningless and redundant hierarchy. On the other hand, unique files all serve the same purpose in the same folder can also means clear. It is already structured data and it is enough to be clear. The point here is "your own definition for structured data" is not very sutiable here from my humble opinion.

Well, I do not think it is related to who proposes/defines this.  And no one has absolute right to decide what is suitable either. The review process is just to see other's ideas in case the author does not notice or not aware of something , either good or bad.  IMO, directory layers make the config files better structured/organized, at least better than it is now. This is how I see it. But if you do not agree, totally fine. No force for anything.

waynechen55:
> > > > > Can we oraganize all guest unattended installation files in a way like this?
> > > > > data/virt_autotest/guest_unattended_installation_files/{suse{sles,tw,leap},oracle,windows,...}
> > > > 
> > > > 
> > > > From my POV data/virt_autotest/guest_unattended_installation_files is enough. Additional layers just additional layers and additional changes to adapt these layers. To be honest, there is no use because "guest files" are not parsed based on layers(like the legacy qa lib virtauto test package), now they are given directly and conveniently. If "the layers as you said" was introduced, it is useless and only means you wrap "guest files" in a "folder" and pass it transparently to another entity which does not even care whether "the guest files" in one folder or another because there is no dependency on folder layers. There is no point to do this.
> > > 
> > > 
> > > I think the point is to have better structured data. Nothing to do with how it is consumed.
> > 
> > 
> > But there is no need of "this structured data comes from addition folder layers" from my POV. It is not for you to see it as "structured data", actually it is already structured data because it is the place to store unattended installation files, there is no need to add extra folder structure, let alone all unattended installation files are unique and clear enough(they are not the "def" things in legacy qa lib virtauto test package, all "def" things are the same so they should be put in different folder to be unique. Such additional folder layer is not necessary here and useless because things are already unique and clear). And those files in it can also tell what it is used for, let alone automation which does not even care. To be honest, you should change your angle to see it instead of sticking to "a rigid structured data". Additional folder layer does not mean clearer structure, it can also mean meaningless and redundant hierarchy. On the other hand, unique files all serve the same purpose in the same folder can also means clear. It is already structured data and it is enough to be clear. The point here is "your own definition for structured data" is not very sutiable here from my humble opinion.
> 
> Well, I do not think it is related to who proposes/defines this. And no one has absolute right to decide what is suitable either. The review process is just to see other's ideas in case the author does not notice or not aware of something , either good or bad. IMO, directory layers make the config files better structured/organized, at least better than it is now. This is how I see it. But if you do not agree, totally fine. No force for anything.

I can not say "additional folder layers" is better than without them. It depends on whether it is needed. "No need" means already "clear" after thoughtful thinking. If "additional folder layers" is also added in in this situation, I do not see it as "clearer", I say it is "over engineering" for this case. It might be habit issue or personal preference, so I have no strong feeling to do any judgement. IMHO, there is no need at all for this case and there is no need to change on this aspect. Hope you can understand but I am forcing you to change your angle if you do not. I do my best to explain.

alice-suse:
It is not a big deal, fine to be merged as it is. 
Now nearly all files under data/virt_autotest are plain structured(under it there are various shell scripts, xml files, autoyast, etc), rather than overall hierarchical structured with directory layers according to their file serving purpose. It is ok with current size. When it grows wild to some quantity, may worth a ticket to setup structure for all.

waynechen55:
> It is not a big deal, fine to be merged as it is.
> Now nearly all files under data/virt_autotest are plain structured(under it there are various shell scripts, xml files, autoyast, etc), rather than overall hierarchical structured with directory layers according to their file serving purpose. It is ok with current size. When it grows wild to some quantity, may worth a ticket to setup structure for all.

I agree to merge. 
As far as whether "extra structure" is needed is concerned, I think it should depend on whether it is really needed. For example, if you find it is very difficult to extend or it is not clear at all anymore. From what I have seen till now, there is no such issue. And I am not short-sighted as well, current folder structure is concise and clear. But anyway I can not rule out all possibilities to make some change in some day, I will be very happy to do that if really necessary. By the way, all guest profiles are in a folder, all guest unattended installation files are in another folder, they are different folders, they already have "structure and layer".

alice-suse:
>By the way, all guest profiles are in a folder, all guest unattended installation files are in another folder, they are different folders, they already have "structure and layer", and very concise structure and layer.

Well, if I get correctly from PR, the sles autoyast files are in a different folder from "guest unattended installation files" -- I gave in initial suggestion that we'd better put sles autoyast files also under "guest unattended installation files"  umbrella.

waynechen55:
> > By the way, all guest profiles are in a folder, all guest unattended installation files are in another folder, they are different folders, they already have "structure and layer", and very concise structure and layer.
> 
> Well, if I get correctly from PR, the sles autoyast files are in a different folder from "guest unattended installation files" -- I gave in initial suggestion that we'd better put sles autoyast files also under "guest unattended installation files" umbrella. But from your later comments, seems you do not want to change that way. I am confused now, how do you want to organize these unattended files including autoyast?

All unattended installation files are already in "guest unattended installation files" including sles autoyast files.

alice-suse:
> All unattended installation files are already in "guest unattended installation files" folder(data/virt_autotest/guest_unattended_installation_files), including sles autoyast files.

Right. I checked again, the files are renamed and under this top umbrella. Sorry for missing this before.
Merge it now.

PR #12933, opened 2021-07-20
Add logs for variables parsed from yaml schedule

sofiasyria:
There are some openQA variables, parsed from yaml schedule. This commit creates logs for these particular VARS.  Also, modifies message for YAML test_data, in order to be more clear.

- Related ticket: https://progress.opensuse.org/issues/94955
- Verification run: http://falafel.suse.cz/tests/499/logfile?filename=autoinst-log.txt

### DIFF
--- a/lib/scheduler.pm
+++ b/lib/scheduler.pm
@@ -35,6 +35,7 @@ my $root_project_dir = dirname(__FILE__) . '/../';
 my $include          = YAML::PP::Schema::Include->new(paths => ($root_project_dir));
 my $ypp              = YAML::PP->new(schema => ['Core', $include, 'Merge']);
 $include->yp($ypp);
+$Data::Dumper::Terse = 1;
 
 sub parse_vars {
     my ($schedule) = shift;
@@ -43,6 +44,8 @@ sub parse_vars {
         $value =~ s/%(.*?)%/get_var($1)/eg;
         $vars{$var} = $value;
     }
+    my $varlog = Dumper(\%vars);
+    diag("parse_vars (variables parsed from YAML schedule): " . $varlog);
     return %vars;
 }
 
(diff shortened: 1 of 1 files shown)


lib/scheduler.pm
```diff
@@ -43,6 +44,8 @@ sub parse_vars {
         $value =~ s/%(.*?)%/get_var($1)/eg;
         $vars{$var} = $value;
     }
+    my $varlog = Dumper(\%vars);
+    diag("parse_vars (VARS parsed from YAML schedule): " . $varlog);
```
jknphy, line 48: neatpicking: they are called Settings in openQA, instead of VARS, and we call it 'vars' in our yaml schedule, so something like this would be more clear: `openQA settings parsed from 'vars' section in YAML schedule`. wdyt?
Could we apply the same in [line 132](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12933/files#diff-bbd0c458d484c80e3bf564bd65378809c63c2268756729a9f9319a034783b1abR132) ?

jknphy: ok, documentation also use 'variables' so for that is called vars.json that generated file, so, we could use `Variables parsed from 'vars' section in YAML schedule` or `Variables parsed from YAML schedule`, but it is just a suggestion.

OleksandrOrlov (approved):
Thanks for the change. Please, address Joaquin's comment and we can merge.

jknphy:
> Thanks for the change. Please, address Joaquin's comment and we can merge.

only missing point is if you want to add that additional info in line 132 @sofiasyria

sofiasyria:
> > Thanks for the change. Please, address Joaquin's comment and we can merge.
> 
> only missing point is if you want to add that additional info in line '#132' as well @sofiasyria

I don't think it's needed because parse_test_suite_data is a clear function name. Adding "Data parsed from YAML test data" wouldn't add valuable info. WDYT?

jknphy:
> > > Thanks for the change. Please, address Joaquin's comment and we can merge.
> > 
> > 
> > only missing point is if you want to add that additional info in line '#132' as well @sofiasyria
> 
> I don't think it's needed because parse_test_suite_data is a clear function name. Adding "Data parsed from YAML test data" wouldn't add valuable info. WDYT?

For collaborators not knowing much openQA might help, we also have http://open.qa/api/testapi/#_get_test_data, so to indicate that the data comes from yaml and not from data/ folder might reduce confusion.

PR #12935, opened 2021-07-20
Add mce-inject module at beginning of the test

dzedro:
MCE event is not recorded after injection, not sure why does that happen.
Add mce-inject module at beginnig of test, to make sure it's loaded when
injecting errors.

- Related ticket: https://progress.opensuse.org/issues/95126
- Verification run:
http://dzedro.suse.cz/tests/18668 15 SP2
http://dzedro.suse.cz/tests/18671 15 SP1
http://dzedro.suse.cz/tests/18670 15
https://openqa.suse.de/tests/6533663 15 SP2

### DIFF
--- a/tests/hpc/rasdaemon.pm
+++ b/tests/hpc/rasdaemon.pm
@@ -30,8 +30,6 @@ use testapi;
 use utils;
 
 sub inject_error {
-    # load kernel module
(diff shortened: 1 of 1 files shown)


czerw:
Thanks a lot!

PR #12946, opened 2021-07-21
Schedule vsftpd in s390x maintenance runs

punkioudi:
When this [PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12919) is merged, vsftpd can be also scheduled for s390x architecture in maintenance runs.

- Related ticket: https://progress.opensuse.org/issues/93270
- Needles: N/A
- Verification run: SLES [15SP3](https://openqa.suse.de/tests/6506800#step/vsftpd/31) | [15SP2](https://openqa.suse.de/tests/6506801#step/vsftpd/31) | [15SP1](https://openqa.suse.de/tests/6506802#step/vsftpd/31) | [15](https://openqa.suse.de/tests/6506803#step/vsftpd/31) | [12SP5](https://openqa.suse.de/tests/6506804#step/vsftpd/31) | [12SP4](https://openqa.suse.de/tests/6506805#step/vsftpd/31)

### DIFF
--- a/schedule/qam/common/mau-extratests2.yaml
+++ b/schedule/qam/common/mau-extratests2.yaml
@@ -30,6 +30,7 @@ schedule:
   - console/aaa_base
   - console/gd
   - console/systemtap
+  - console/vsftpd
   - '{{version_specific}}'
   - console/coredump_collect
 conditional_schedule:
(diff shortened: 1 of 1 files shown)


foursixnine:
@punkioudi the only issue I have is that we will end up with an unconditional soft fail in s390x regardless of whether the bug is present or not, fortunately it won't mask other issues which is good, could you add a separate ticket to remove the soft failure??.

foursixnine (approved):

punkioudi:
> @punkioudi the only issue I have is that we will end up with an unconditional soft fail in s390x regardless of whether the bug is present or not, fortunately it won't mask other issues which is good, could you add a separate ticket to remove the soft failure??.

Yes, sure omw. Actually the ticket will be : 1. Remove the soft-failure, 2. Remove the `run_s390x.sh`

PR #12948, opened 2021-07-21
Handle autologin broken on aarch64 Tumbleweed for xfce - boo#1188559

ggardet:
https://bugzilla.opensuse.org/show_bug.cgi?id=1188559

- Verification run: https://openqa.opensuse.org/t1850680


ggardet:
Manual login also fail, so no workaround is possible.

PR #12949, opened 2021-07-21
SLE Micro: add migration test from 5.0 to 5.1

jlausuch:
Reuse existing zypper_migration test and adapting zypper commands
to transactional server equivalent.

- Related ticket: https://progress.opensuse.org/issues/95383
- Verification run:  https://openqa.suse.de/tests/6512108

### DIFF
--- a/schedule/sle-micro/migration.yaml
+++ b/schedule/sle-micro/migration.yaml
@@ -1,0 +1,19 @@
+name:           sle_micro_containers
+description:    >
+    Maintainer: qa-c@suse.de.
+    SUSE Linux Enterprise Micro tests
+schedule:
+  - microos/disk_boot
+  - transactional/disable_grub
+  - console/suseconnect_scc
+  - transactional/install_updates
+  - migration/online_migration/zypper_migration
+  - microos/networking
+  - microos/libzypp_config
+  - microos/image_checks
+  - microos/one_line_checks
+  - microos/services_enabled
+  - microos/cockpit_service
+  - microos/toolbox
+  - console/journal_check
+  - shutdown/shutdown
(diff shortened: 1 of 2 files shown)


jlausuch:
@tinawang123 @hjluo

mloviska:
Isn't this almost the same https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/transactional/tdup.pm ?
Even migration is distro upgrade, or not ?

jlausuch:
> Isn't this almost the same https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/transactional/tdup.pm ?
> Even migration is distro upgrade, or not ?

`transactional-update dup` calls `zypper dup`  and `transactional-update migration` calls `zypper migration` :) 
I'm not 100% sure about the difference but I think for `zypper dup` you need to setup the repos yourself. Migration allows you to select a target to migrate to. So, let's put it this way: we want to test the `migration` subcommand from `transactional-update`.

asmorodskyi (approved):
LGTM

mloviska:
> > Isn't this almost the same https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/transactional/tdup.pm ?
> > Even migration is distro upgrade, or not ?
> 
> `transactional-update dup` calls `zypper dup` and `transactional-update migration` calls `zypper migration` :)
> I'm not 100% sure about the difference but I think for `zypper dup` you need to setup the repos yourself. Migration allows you to select a target to migrate to. So, let's put it this way: we want to test the `migration` subcommand from `transactional-update`.
> 
Ah, yeah. TW user :), sorry
> Interesting reading: https://documentation.suse.com/sles/15-SP3/html/SLES-all/cha-upgrade-online.html#sec-upgrade-online-zypper

PR #12951, opened 2021-07-21
Revert "Extend AutoYaST file(s) preparation for Rules and Classes" (poo#95804)

okurz:
Reverts os-autoinst/os-autoinst-distri-opensuse#12916
due to https://progress.opensuse.org/issues/95804


OleksandrOrlov (changes requested):
Thanks Oliver for being proactive. We'll consider that revert, once we'll have chance to investigate the issue.

PR #12953, opened 2021-07-22
Added leap upgrade to sle test workflow for uefi

tinawang123:
Added leap upgrade to sle test workflow for uefi

Need support uefi test scenario for leap upgrade to sle

- Related ticket:https://progress.opensuse.org/issues/92650
- Verification run: https://openqa.suse.de/tests/6514986#
Verified opensuse: https://openqa.opensuse.org/tests/1856894#
Verified security case: https://openqa.suse.de/tests/6566995

### DIFF
--- a/tests/installation/bootloader_uefi.pm
+++ b/tests/installation/bootloader_uefi.pm
@@ -106,8 +106,8 @@ sub run {
         # Select boot from HDD
         send_key_until_needlematch 'inst-bootmenu-boot-harddisk', 'up';
         send_key 'ret';
-        # use firmware boot manager of aarch64 to boot HDD
-        $self->handle_uefi_boot_disk_workaround if (check_var('ARCH', 'aarch64'));
+        # use firmware boot manager of aarch64 and uefi to boot HDD
+        $self->handle_uefi_boot_disk_workaround if (check_var('ARCH', 'aarch64') || get_var('UEFI'));
         assert_screen("grub2");
         return;
     }


tests/installation/bootloader_uefi.pm
```diff
@@ -106,8 +106,8 @@ sub run {
         # Select boot from HDD
         send_key_until_needlematch 'inst-bootmenu-boot-harddisk', 'up';
         send_key 'ret';
-        # use firmware boot manager of aarch64 to boot HDD
-        $self->handle_uefi_boot_disk_workaround if (check_var('ARCH', 'aarch64'));
+        # use firmware boot manager of aarch64 and uefi to boot HDD
+        $self->handle_uefi_boot_disk_workaround if (check_var('ARCH', 'aarch64') || get_var('UEFI'));
```
rfan1, line 110: Yutao,

For x86_64, we should use the same workaround, right?

lemon-suse: I just wonder this setting will make some failure that some cases with 'UEFI' may not want to use this workaround, so are you sure such cases with UEFI could use this workaround?

lemon-suse: I just wonder this setting will make some failure that some cases with 'UEFI' may not want to use this workaround, so are you sure such cases with UEFI could use this workaround?

tinawang123: I have verified https://openqa.suse.de/tests/6532122#

tinawang123: > Yutao,
> 
> For x86_64, we should use the same workaround, right?

It won't impact x86 case.

lemon-suse: There are so many testcases on O3 with the bootloader_uefi.pm, so it is better to run verification logs on O3 to avoid regression.

lemon-suse (approved):

rfan1 (approved):
LGTM

PR #12955, opened 2021-07-22
Unified guest installation installer

waynechen55:
* **Rename** uefi_guest_installation.pm to unified_guest_installation.pm because it is really an unified guest installation installer instead of can only be used with uefi guest installation.
* **Rename** UEFI_GUEST_LIST and UEFI_GUEST_PROFILES to UNIFIED_GUEST_LIST and UNIFIED_GUEST_PROFILES to also indicate that these two arguments can be used with any kind of guest.
* **Introduce** VIRT_UNIFIED_GUEST_INSTALL to indicate that the new module virt_autotest/unified_guest_installation is preferred other than legacy one.
* **VIRT_UEFI_GUEST_INSTALL** is kept to indicate that the installed guest is uefi guest so uefi_guest_verification needs to be called to verify it.
* **Replace** the old "VIRT_UEFI_GUEST_INSTALL" in tests/virt_autotest/virt_utils.pm and tests/virtualization/universal/hotplugging.pm to "VIRT_UNIFIED_GUEST_INSTALL" because integrated dhcp/dns is being used in new guest installation module instead of manipulating them separately."VIRT_UNIFIED_GUEST_INSTALL" is the one to be used here.

* **Verification runs:**
  * [oracle linux 6 with features](http://10.67.129.106/tests/1495)
  * [oracle linux 8 uefi guest with features](http://10.67.129.106/tests/1521)

### DIFF
--- a/products/sle/main.pm
+++ b/products/sle/main.pm
@@ -600,10 +600,10 @@ sub load_yast2_registration_tests {
 }
 
 sub load_virt_guest_install_tests {
-    if (get_var("VIRT_UEFI_GUEST_INSTALL")) {
-        loadtest "virt_autotest/uefi_guest_installation";
+    if (get_var("VIRT_UNIFIED_GUEST_INSTALL")) {
+        loadtest "virt_autotest/unified_guest_installation";
         loadtest "virt_autotest/set_config_as_glue";
-        loadtest "virt_autotest/uefi_guest_verification";
+        loadtest "virt_autotest/uefi_guest_verification" if get_var("VIRT_UEFI_GUEST_INSTALL");
     }
     else {
         loadtest "virt_autotest/guest_installation_run";
--- a/tests/virt_autotest/unified_guest_installation.pm
+++ b/tests/virt_autotest/unified_guest_installation.pm
@@ -7,15 +7,15 @@
 # notice and this notice are preserved. This file is offered as-is,
 # without any warranty.
 #
-# Summary: This module supports concurrent multiple uefi virtual machine
-# installations with vm names and profiles obtained from UEFI_GUEST_LIST
-# and UEFI_GUEST_PROFILES respectively. There is no restriction on vm names
-# to be used, so any desired vm names can be given to UEFI_GUEST_LIST=
+# Summary: This module supports concurrent multiple virtual machines
+# installations with vm names and profiles obtained from UNIFIED_GUEST_LIST
+# and UNIFIED_GUEST_PROFILES respectively. There is no restriction on vm names
+# to be used, so any desired vm names can be given to UNIFIED_GUEST_LIST=
 # "vm_name_1,vm_name_2,vm_name_3". Similary,any vm profile names can be
-# given to UEFI_GUEST_PROFILES,as long as there are corresponding profile
+# given to UNIFIED_GUEST_PROFILES,as long as there are corresponding profile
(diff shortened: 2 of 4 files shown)


waynechen55:

alice-suse:
LGTM

nanzhg:
LGTM

Julie-CAO:
Thanks for the renaming. 

VIRT_UNIFIED_GUEST_INSTALL & VIRT_UEFI_GUEST_INSTALL look very much alike. Also, we may have more tests on UEFI guest rather than install, so how about name it VIRT_UEFI_GUEST?

I am going to do the corresponding changes to my factory-first PR...

waynechen55:
> Thanks for the renaming.
> 
> VIRT_UNIFIED_GUEST_INSTALL & VIRT_UEFI_GUEST_INSTALL look very much alike. Also, we may have more tests on UEFI guest rather than install, so how about name it VIRT_UEFI_GUEST?
> 
> I am going to do the corresponding changes to my factory-first PR...

Changed VIRT_UEFI_GUEST_INSTALL to VIRT_UEFI_GUEST_TEST. So it can be used anywhere uefi guest specific test applies.

Julie-CAO:
LGTM. thanks.

waynechen55:
> LGTM. thanks.

I changed VIRT_UEFI_GUEST_TEST back to VIRT_UEFI_GUEST_INSTALL. Although this is not a big problem, I am willing to explain in more details. I think keep it as the original VIRT_UEFI_GUEST_INSTALL is better to reflect what is included in load_virt_guest_install_tests function. virt_autotest/uefi_guest_verification is used to do basic checking on installed uefi guest, VIRT_UEFI_GUEST_INSTALL is more like a parameter paired with VIRT_UNIFIED_GUEST_INSTALL to indicate what kind of guest is to be installed during guest installation phase instead of controlling feature test. And  virt_autotest/uefi_guest_verification is more like a paired module with installed uefi guest instead of a feature test because it tests whether uefi guest is installed successfully, not whether a feature works for uefi guest. I suggest you use another parameter like "VIRT_UNIFIED_GUEST_TEST" to control feature test if necessary in the future. Thanks.

Julie-CAO:
> > LGTM. thanks.
> 
> I changed VIRT_UEFI_GUEST_TEST back to VIRT_UEFI_GUEST_INSTALL. Although this is not a big problem, I am willing to explain in more details. I think keep it as the original VIRT_UEFI_GUEST_INSTALL is better to reflect what is included in load_virt_guest_install_tests function. virt_autotest/uefi_guest_verification is used to do basic checking on installed uefi guest, VIRT_UEFI_GUEST_INSTALL is more like a parameter paired with VIRT_UNIFIED_GUEST_INSTALL to indicate what kind of guest is to be installed beside which guest installation module is to be used during guest installation phase, it is not supposed to control any following feature test. And virt_autotest/uefi_guest_verification is more like a paired module with installed uefi guest instead of a feature test because it tests whether uefi guest is installed successfully, not whether a feature works for uefi guest. I think this way is better. I suggest you use another parameter like "VIRT_UEFI_GUEST_TEST" or "ENABLE_TESTING_UEFI_GUEST" to control feature test if necessary in the future. Thanks.

It is ok to me.

alice-suse:
Seems all comments resolved, merge it.

PR #12957, opened 2021-07-22
Set the filesystem on root partition to test different filesystem

lemon-suse:
We'd like to test on different file system such as ext4,xfs besides the default btrfs on root partition.

Related ticket: https://progress.opensuse.org/issues/93600
Needles: Already merged in OSD.
Verification run:
  x86_64:
  https://openqa.nue.suse.com/tests/6506809#step/partitioning/2 - ext4
  http://openqa.nue.suse.com/tests/6506810#step/partitioning/2 - xfs
  later will support more verification log on other Arches.

### DIFF
--- a/tests/installation/partitioning.pm
+++ b/tests/installation/partitioning.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2017 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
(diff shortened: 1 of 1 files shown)


tests/installation/partitioning.pm
```diff
@@ -22,15 +22,18 @@ use partition_setup qw(%partition_roles is_storage_ng_newui);
 
 sub run {
     assert_screen 'partitioning-edit-proposal-button', 40;
-    if (check_var('PARTITION_EDIT', 'ext4_btrfs')) {
+    # We need to test on different filesystems on root partition
+    # and do not care the filesystem on home partition.
+    my $my_filesys = get_var('PARTITION_EDIT');
```
tinawang123, line 27: If don't set 'PARTITION_EDIT', the $my_filesys will get undefined.

lemon-suse: Not set then we won't run the following code flow.

lemon-suse: undefine should not be '1', so the condition should work. :)

tinawang123 (approved):
LGTM

PR #12960, opened 2021-07-22
Revert is_tumbleweed from is_storage_ng

jknphy:
This condition was added by me recently and even if makes sense, doesn't get along well with the 'monster' code that we have in `is_boot_encrypted`. We will improve that part in the future, for now, this change will fix Leap upgrades and adding setting FULL_LVM_ENCRYPT to encryptlvm in TW should allow correct booting using current module.

- Related ticket: [poo#94904](https://progress.opensuse.org/issues/94904)
- Verification run: [cryptlvm + upgrade_Leap_42.3_cryptlvm](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=jknphy%2Fos-autoinst-distri-opensuse%23fix_encrypt_scenarios_o3)

### DIFF
--- a/lib/version_utils.pm
+++ b/lib/version_utils.pm
@@ -376,7 +376,7 @@ sub is_staging {
 Returns true if storage_ng is used
 =cut
 sub is_storage_ng {
-    return get_var('STORAGE_NG') || is_sle('15+') || is_tumbleweed;
+    return get_var('STORAGE_NG') || is_sle('15+');
 }
 
 =head2 is_upgrade


lib/version_utils.pm
```diff
@@ -376,7 +376,7 @@ sub is_staging {
 Returns true if storage_ng is used
 =cut
 sub is_storage_ng {
-    return get_var('STORAGE_NG') || is_sle('15+') || is_tumbleweed;
+    return get_var('STORAGE_NG') || is_sle('15+');
```
okurz, line 379: Our tests should always consider openSUSE Tumbleweed as the default product without a version. That means any differing behaviour should have an explicit exclude rule for the "older products" and potentially an explanation why the behaviour should differ. You already remove `is_tumbleweed` but you can also try to negate the logic check.

jknphy: yes, I got your point, it is a good remainder, but the best solution is to not use this sort of conditions at all. This function was created wrongly without the default you mentioned of 1 and the exceptions and collaborators just pile up more conditions without having scope of all products.

PR #12969, opened 2021-07-23
Do not run the --add-samba-data testcase in sle 15

ge0r:
sle 15 fails with unrecognized option '--add-samba-data', since the --add-samba-data option is not available in adcli for this product.

- Needles: No needles
- Verification run: [before](http://angmar.suse.de/tests/3577#step/samba_adcli/108) | [after](http://angmar.suse.de/tests/3649#)

### DIFF
--- a/tests/network/samba/samba_adcli.pm
+++ b/tests/network/samba/samba_adcli.pm
@@ -133,7 +133,7 @@ sub run {
     assert_script_run "expect -c 'spawn ssh -l geekouser\@geeko.com localhost -t;expect sword:;send Nots3cr3t\\n;expect geekouser>;send exit\\n;interact'";
 
     # poo#91950 (update machine password with adcli --add-samba-data option)
-    update_password();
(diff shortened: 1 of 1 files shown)


tests/network/samba/samba_adcli.pm
```diff
@@ -133,7 +133,7 @@ sub run {
     assert_script_run "expect -c 'spawn ssh -l geekouser\@geeko.com localhost -t;expect sword:;send Nots3cr3t\\n;expect geekouser>;send exit\\n;interact'";
 
     # poo#91950 (update machine password with adcli --add-samba-data option)
-    update_password();
+    update_password() if (!is_sle('=15'));    # sle 15 does not support the `--add-samba-data` option
```
punkioudi, line 136: you could just use unless ^^

ge0r: good point, much clearer with unless

punkioudi (approved):

PR #12972, opened 2021-07-26
CONTRIBUTING: Add rule suggestions about avoiding TODO/FIXME

okurz:


CONTRIBUTING.md
```diff
@@ -91,6 +91,9 @@ and additionally the following rules:
 * Avoid "dead code": Don't add disabled code as nobody but you will understand
   why this is not active. Better leave it out and keep in your local git
   repository, either in `git stash` or a temporary "WIP"-commit.
+* Do not add TODO/FIXME references: TODO comments will rot. Use plain comments
+  to explain or link to external issue trackers. Also see
+  https://blog.usejournal.com/stop-writing-code-comments-28fef5272752
```
asmorodskyi, line 96: I don't think that including blog posts of random people is good idea : 
1. This is just personal opinion of single person . Which may or may not be accepted by all contributors of this project. This not some law which everyone should obey without doubts.
2. No guarantee for how long this URL will be valid . Also no guarantee that after some time this URL will point to some different article

okurz: > I don't think that including blog posts of random people is good idea :
> 
>     1. This is just personal opinion of single person . Which may or may not be accepted by all contributors of this project. This not some law which everyone should obey without doubts.

Exactly. This is why I am referencing an article that is not written by me and I ask reviewers of the project if they agree with the rule.

>     2. No guarantee for how long this URL will be valid . Also no guarantee that after some time this URL will point to some different article

Yes, nothing is forever :) Should I include the date when I accessed the blog article? I doubt we can easily include a checksum of the blog content here to ensure it's the very same over time ;)

foursixnine: You can include https://wiki.c2.com/?TodoComments which has arguments pro and against todo/fixme notes.

Having a visible TODO, FIXME comment gives a better opportunity for that idea to be implemented, or deleted altogether if necessary, whilst a ticket in the tracker again, loses relevance and is there rotting for years.

While I can understand how a TODO can be an antipattern in the lib/ or a baseclass, I don't agree that we shouldn't allow todos at all.

foursixnine (changes requested):
It's a negative from my side, While TODO's and FIXME can rot. In our case, we're not just writing code, but expressing user behavior for tests, at the moment of writting, and because something was in a small scope or a proof of concept one could have the need to express the idea, instead of having a ticket in the issue tracker that will also rot and lose it's relevance as time goes by, and tests are updated and changed.

Makefile
```diff
@@ -13,6 +13,8 @@ prepare:
 	$(MAKE) check-links
 	cd os-autoinst && cpanm -nq --installdeps .
 	cpanm -nq --installdeps .
+	# ensure high enough limit for rename detection
+	git config diff.renameLimit || git config diff.renameLimit 4000
```
foursixnine, line 17: This is an unrelated change

okurz: Well unrelated to the one git commit, yes. But I put it into a separate git commit: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12972/commits/c0e3e8d1426f09b719c5259d1934471a8e923832

CONTRIBUTING.md
okurz: I like https://wiki.c2.com/ but https://wiki.c2.com/?TodoComments as well most of the related pages are applicable to single-person hacking projects it seems. Here we want to have contribution guidelines for  a big community project.

Obviously tickets in ticket trackers can stay there for long but this is my point: At the time of writing people think it's important so they add TODO/FIXME but the next day you do something different and years later that TODO/FIXME is still there, the code is doing no harm and fixing it is just no priority. Making the code less clear to read and confusing other developers is not helpful. Having a ticket that you close after years realizing that the issue actually turned out to be not relevant anymore even though we might all agree that the code is not perfect or not beautiful is just realistic and should be expected.

Would you agree if I say " … especially in lib/ code or a baseclass"?

asmorodskyi (changes requested):
I don't see the point of finding common ground on question "Is it right to add TODO/FIXME comments or not ? ". This discussion will be too long and I am not sure if it is feasible to find common answer. 
But I think you will agree that there several opinion on this regard so please avoid bold statements . Something like "Try to avoid using TODO/FIXME" sound more approriate for me. 
Also I am against adding any references except some wikipedia articles or resources with same level of trust

CONTRIBUTING.md
foursixnine: I would agree with: " Avoid using TODO/FIXME especially in lib/ code or a baseclass, as these will likely rot, it is not forbidden, but recommended not to. Whenever an idea needs to be expressed, use plain comments or add a reference to a ticket in the issue tracker those may rot too, but that's also expected and doesn't pollute the code."

schlad: I much more prefer this phrasing: "recommended not to"

schlad:
I do not agree with this: "or link to external issue trackers."
As a matter of fact I would personally prefer to have code and git log free of any urls linking with anything which would be 'external' to the git repo. IMO git repo should be self-sufficient.

schlad (changes requested):
Please re-phrase the sentence around TODO. Also please remove the encouragement  towards any external issue tracker.

foursixnine:
@okurz https://github.com/os-autoinst/os-autoinst-distri-opensuse/blame/1454b4f9c6dedf645bad4094a897651c0f5a79f9/lib/main_common.pm#L1092 really?.

okurz:
> @okurz https://github.com/os-autoinst/os-autoinst-distri-opensuse/blame/1454b4f9c6dedf645bad4094a897651c0f5a79f9/lib/main_common.pm#L1092 really? (old commit, I just wanna tease)

:D That was old me, I have learned since then ;) Honestly I learned to avoid TODOs 20 years ago, no idea what made me think in this commit someone would look into the TODO anytime soon

CONTRIBUTING.md
okurz: Updated along the suggestion of @foursixnine with minor changes

okurz:
@os-autoinst/tests-maintainer I think I have all the suggestions from @foursixnine @schlad @asmorodskyi incorporated but received neither further suggestions nor approval. Can someone else please help to move this forward, i.e. either merge or tell me what I need to change?

jlausuch:
I don't have a strong opinion about this, but I would prefer not enforcing this rule. Sometimes when you start debugging the code trying to find the cause of a failure you step into that `TODO/FIXME` which gives you a hint of what might happen. I would prefer having a comment instead of a url to some issue tracker. But if we keep this as a suggestion, I'm fine with it... My vote is neutral.
BTW, if we accept it, there is some work to do afterwards :)
```
$ cd os-autoinst-distri-opensuse
$ grep -rn "# TODO\|#TODO" ./|wc -l
      80

$ grep -rn "# FIXME\|#FIXME" ./|wc -l
      25
```

b10n1k:
>  Here we want to have contribution guidelines for a big community project.

Considering those tags as legal programming and useful technique:

There are SM and PO who mainly decide how they like to do some things. 
And there are reviewers who can/should be able to say if a tag like TODO make sense or not
And there are maintainers who, ideally, go through the code and their modules. 

Personally, i could suggest a few things, directly and indirectly:
- create a TODO file which will work as a org-mode file. Have an entry with a deadline which can be review easily. (ask me if you need more arguments)
- To help the contributors and the reviewers, create another sub-section in the `Coding style`. What is *must* to do/have and what is *suggestion* or *recommended*. In general any organization which ppl can easily helped finding a specific rule. As extension could be the distinction among perl coding things, general comments and commit stuff

grisu48 (changes requested):
I'm against this change because this will not make people improve their code but only remove the TODO in the comment or omit that particular comment as a whole.

I'd much rather have stale TODO's in the code that at least point me in the direction of what the thought at this point was, than having someone push code without those comments because we enforce them to not be there.

In my experience this does not lead to better code but even less comments and this would be a step backwards.

***

That being said, we could apply this to the `lib` folder, where we also enforce then higher standards. Simply filtering `TODO/FIXME` comments won't be enough though.

okurz:
@grisu48
> That being said, we could apply this to the `lib` folder, where we also enforce then higher standards. Simply filtering `TODO/FIXME` comments won't be enough though.

The text already says "Avoid using TODO/FIXME especially in library code, e.g. in the lib/ folder or a baseclass,". Do you have a suggestion for a better phrasing?

foursixnine:
> @grisu48
> 
> > That being said, we could apply this to the `lib` folder, where we also enforce then higher standards. Simply filtering `TODO/FIXME` comments won't be enough though.
> 
> The text already says "Avoid using TODO/FIXME especially in library code, e.g. in the lib/ folder or a baseclass,". Do you have a suggestion for a better phrasing?

@okurz rephrase it with "suggest not to" instead of avoid, after all... the sentiment is clear: Nobody should be adding 300 TODO/FIXME and nobody should be harassing anybody with comments of "Avoid..."... or at best, make them an info message in the CI or something like that.

okurz:
> > @grisu48
> > > That being said, we could apply this to the `lib` folder, where we also enforce then higher standards. Simply filtering `TODO/FIXME` comments won't be enough though.
> > 
> > 
> > The text already says "Avoid using TODO/FIXME especially in library code, e.g. in the lib/ folder or a baseclass,". Do you have a suggestion for a better phrasing?
> 
> @okurz rephrase it with "suggest not to" instead of avoid, after all... the sentiment is clear: Nobody should be adding 300 TODO/FIXME and nobody should be harassing anybody with comments of "Avoid..."... or at best, make them an info message in the CI or something like that.

"suggest not to …" does not form a complete sentence. Also for the sake of consistency I would use "Avoid" as this is what we use in multiple other sentences.

PR #12978, opened 2021-07-27
Update perltidy to 20210717

okurz:
See related change https://github.com/os-autoinst/os-autoinst/pull/1732


okurz:
wait for https://github.com/os-autoinst/os-autoinst/pull/1732 first

PR #12982, opened 2021-07-27
Public Cloud: Add --progress to rsync and timeout command before

pdostal:
- Related ticket: [poo#95104](https://progress.opensuse.org/issues/95104)
- Failure: [transfer_repos](https://openqa.suse.de/tests/6578724#step/transfer_repos/7)
- Verification run: http://pdostal-server.suse.cz/tests/12384#step/transfer_repos/6

### DIFF
--- a/tests/publiccloud/transfer_repos.pm
+++ b/tests/publiccloud/transfer_repos.pm
@@ -32,7 +32,8 @@ sub run {
         record_info('Skip download', 'Skipping download triggered by setting (QAM_PUBLICCLOUD_SKIP_DOWNLOAD = 1)');
     } else {
         assert_script_run('du -sh ~/repos');
-        assert_script_run("rsync -uva -e ssh ~/repos root@" . $args->{my_instance}->public_ip . ":'/tmp/repos'", timeout => 1800);
+        my $timeout = 2400;
+        assert_script_run("rsync --timeout=$timeout -uvahP -e ssh ~/repos root@" . $args->{my_instance}->public_ip . ":'/tmp/repos'", timeout => $timeout + 10);
         $args->{my_instance}->run_ssh_command(cmd => "sudo find /tmp/repos/ -name *.repo -exec sed -i 's,http://,/tmp/repos/repos/,g' '{}' \\;");
         $args->{my_instance}->run_ssh_command(cmd => "sudo find /tmp/repos/ -name *.repo -exec zypper ar '{}' \\;");
         $args->{my_instance}->run_ssh_command(cmd => "sudo find /tmp/repos/ -name *.repo -exec echo '{}' \\;");


tests/publiccloud/transfer_repos.pm
```diff
@@ -32,7 +32,7 @@ sub run {
         record_info('Skip download', 'Skipping download triggered by setting (QAM_PUBLICCLOUD_SKIP_DOWNLOAD = 1)');
     } else {
         assert_script_run('du -sh ~/repos');
-        assert_script_run("rsync -uva -e ssh ~/repos root@" . $args->{my_instance}->public_ip . ":'/tmp/repos'", timeout => 1800);
+        assert_script_run("timeout 2400 rsync -uvaP -e ssh ~/repos root@" . $args->{my_instance}->public_ip . ":'/tmp/repos'", timeout => 2400);
```
asmorodskyi, line 35: I would say that situation is the same as with curl progress , this will spam output quite a lot so I would keep it optional ( e.g. via some variable ) . We can even use same variable as for curl or add a separate one ...

also not sure what the point having timeout which duplicate functionality which we already have out of the box ( failing after defined period of time ) ?

okurz: it can make sense to have internal timeouts as they can be more fine-grained. In this case I would say the bash-level timeout command is not adding benefits on top of the timeout in the os-autoinst level timeout. On top internal timeouts should always be lower than any higher level of timeout. rsync has an internal timeout parameter. How about:

```suggestion
        my $timeout = 2400;
        assert_script_run("rsync --timeout=$timeout -uvaP -e ssh ~/repos root@" . $args->{my_instance}->public_ip . ":'/tmp/repos'", timeout => $timeout + 10);
```

pdostal: @asmorodskyi the `rsync` currently prints every file it transfers. The `-P` arguments only adds additional data there. This patch updates one specific test module - I don't think we need special variable for that.

@okurz right now, when the `assert_script_run` times out, the `curl` still runs which makes the terminal unusable. Thank you for the internal timeout parameter.

asmorodskyi: ok agree , and I like what Oli suggesting

asmorodskyi (approved):
LGTM

tests/publiccloud/transfer_repos.pm
pdostal: Me too, I already implemented it. Now running verification runs.

jlausuch (approved):

pdostal:
I added the verification run and also added `-h` for human readable output.
```
sent 4.03G bytes  received 85.30K bytes  34.28M bytes/sec
total size is 4.03G  speedup is 1.00
```

PR #12984, opened 2021-07-27
Ignore orphan kernel symlink in livepatch tests

mdoucha:
A few older kernel packages create `/boot/vmlinux` or `/boot/vmlinuz` symlink but fail to set package ownership for it. This packaging mistake causes the check for unexpected kernel packages to fail and blocks livepatch testing. Since there's nothing we can do about these already released packages, ignore the symlink altogether in livepatch tests and only check it in new kernel updates.

- Related ticket: N/A
- Needles: N/A
- Verification run:
  - SLE-15SP1 livepatch: https://openqa.suse.de/tests/6583334
  - SLE-15SP2 livepatch: https://openqa.suse.de/tests/6583337
  - SLE-15SP2 KOTD: https://openqa.suse.de/tests/6583339
  - SLE-15SP2 kernel-rt: https://openqa.suse.de/tests/6583341

### DIFF
--- a/tests/kernel/update_kernel.pm
+++ b/tests/kernel/update_kernel.pm
@@ -30,7 +30,11 @@ sub check_kernel_package {
     my $kernel_name = shift;
 
     script_run('ls -1 /boot/vmlinu[xz]*');
-    my $packs = script_output('rpm -qf --qf "%{NAME}\n" /boot/vmlinu[xz]*');
+    # Only check versioned kernels in livepatch tests. Some old kernel
+    # packages install /boot/vmlinux symlink but don't set package ownership.
+    my $glob  = get_var('KGRAFT', 0) ? '-*' : '*';
+    my $cmd   = 'rpm -qf --qf "%{NAME}\n" /boot/vmlinu[xz]' . $glob;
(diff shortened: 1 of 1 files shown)


mdoucha:
All verification runs have passed, this PR is ready for review and merge.

a-kpappas:
Can't this not be fixed by another update? It seems as if the test found a problem and we are fixing the test instead of the problem.

mdoucha:
> Can't this be fixed by another update? It seems as if the test found a problem and we are fixing the test instead of the problem.

Livepatch tests need to install specific kernel version that can be up to 12 months old so releasing a new kernel will do exactly nothing to fix this problem. Future kernel releases will still be covered by this check so it won't happen again but we need a workaround for the old broken kernels at least for the next year.

foursixnine (approved):

foursixnine:
> Livepatch tests need to install specific kernel version that can be up to 12 months old so releasing a new kernel will do exactly nothing to fix this problem. Future kernel releases will still be covered by this check so it won't happen again but we need a workaround for the old broken kernels at least for the next year.

PR #12988, opened 2021-07-28
Public Cloud: increase max retries for Vault login

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/95266

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -556,14 +556,14 @@ Wrapper around C<<$self->vault_api()>> to get retry capability.
 sub vault_api {
     my ($self, $path, %args) = @_;
     my $ret;
-    my $tries = 3;
+    my $tries = get_var('PUBLICCLOUD_VAULT_TRIES', 3);
 
     while ($tries-- > 0) {
         eval {
             $ret = $self->__vault_api($path, %args);
         };
         return $ret unless ($@);
-        sleep 10;
+        sleep get_var('PUBLICCLOUD_VAULT_TIMEOUT', 60);
     }
     die("Maximum number of Vault request retries exceeded. Check Vault Server is up and running");
 }


lib/publiccloud/provider.pm
```diff
@@ -556,14 +556,14 @@ Wrapper around C<<$self->vault_api()>> to get retry capability.
 sub vault_api {
     my ($self, $path, %args) = @_;
     my $ret;
-    my $tries = 3;
+    my $tries = 10;
```
asmorodskyi, line 559:
```suggestion
    my $tries = get_var(PUBLICCLOUD_VAULT_TRIES, 3);
```

lib/publiccloud/provider.pm
```diff
@@ -556,14 +556,14 @@ Wrapper around C<<$self->vault_api()>> to get retry capability.
...
 
     while ($tries-- > 0) {
         eval {
             $ret = $self->__vault_api($path, %args);
         };
         return $ret unless ($@);
-        sleep 10;
+        sleep 60;
```
asmorodskyi, line 566:
```suggestion
        sleep get_var('PUBLICCLOUD_VAULT_TIMEOUT', 60);
```

lib/publiccloud/provider.pm
jlausuch: yet another 2 new variables? why not deciding in default values for all jobs? These type of failures might happen for any provider, any test... we would need to add the same variable/value to all jobs anyway. What is the difference?

asmorodskyi:
I was investigating this issue , I haven't found anything certain but my gut feel telling me that it can not be solved by such approach. But because it is just gut feel which I can not prove let's do something in the middle :) - leave old behavior as default but allow temporarily change it at run-time or in job group settings

asmorodskyi (changes requested):
let's switch to variables so we don't need more PR ( at least in this repo :) )

asmorodskyi (approved):
LGTM

jlausuch:
VRs: 
without variables: https://openqa.suse.de/tests/6592003
with variables: https://openqa.suse.de/tests/6592005

lib/publiccloud/provider.pm
pdostal: I agree with Jose.

pdostal:
Let's at least add the variables to `variables.md`

jlausuch:
> Let's at least add the variables to `variables.md`

there are no `PUBLIC_CLOUD_` variables defined... I'm not gonna add these 2 there... if we want that, let's add them ALL in a separate PR.

PR #12989, opened 2021-07-28
Public Cloud: Do not upload zypper log in instance_overview.pm

pdostal:
- Related ticket: [poo#95104](https://progress.opensuse.org/issues/95104)


jlausuch:
why, isn't this log useful?

pdostal:
> why, isn't this log useful?

It is too detailed and too big and then there is problem in `zypper` that file is uploaded in `post_fail_hook` anyways.

asmorodskyi (approved):
LGTM

PR #13002, opened 2021-07-30
Containers: schedule PackageHub related tests only if BETA=1

jlausuch:
We started QA for 15-SP4 but PackageHub is not available until Beta
phase, so we should stop running the tests that need PH available.

- Related ticket:  https://progress.opensuse.org/issues/95626

### DIFF
--- a/schedule/containers/engines_and_tools_docker.yaml
+++ b/schedule/containers/engines_and_tools_docker.yaml
@@ -20,21 +20,31 @@ conditional_schedule:
         - containers/docker_runc
   supported:
     HOST_VERSION:
+      15-SP4:
+        - containers/buildah_docker
       15-SP3:
         - containers/buildah_docker
       15-SP2:
         - containers/buildah_docker
       15-SP1:
         - containers/buildah_docker
+  package_hub_dependent:
+    FLAVOR:
+      Server-DVD-Updates:
+        - containers/docker_compose
+        - containers/registry
+    BETA:
+      0:
+        - containers/docker_compose
+        - containers/registry
 schedule:
   - '{{boot}}'
   - boot/boot_to_desktop
   - containers/docker
   - '{{runc}}'
   - '{{supported}}'
-  - containers/docker_compose
   - containers/zypper_docker
   - containers/docker_3rd_party_images
-  - containers/registry
+  - '{{package_hub_dependent}}'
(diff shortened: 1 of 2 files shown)


pdostal:
What if `BETA` is not set?

jlausuch:
> What if `BETA` is not set?

Good point... Yet another reason to use `containers_main.pm` where we could just simply do `if ! get_var('BETA')` :D @asmorodskyi

asmorodskyi:
> > What if `BETA` is not set?
> 
> Good point... Yet another reason to use `containers_main.pm` where we could just simply do `if ! get_var('BETA')` :D @asmorodskyi

I am ready to merge such PR even from vacation :D

pdostal:
Can you please clone for example https://openqa.suse.de/tests/6617406 _

jlausuch:
> Can you please clone for example https://openqa.suse.de/tests/6617406 _

already did the test locally.  I tried to do :

```
  test_undef:
    ABC:
      0:
        - containers/docker_compose
      undef:
        - containers/registry
 ```
 
 in both cases the tests are not scheduled if the variable is NOT set. So, we have a problem here...

**Comment by jlausuch** [comment]
Or we can force `BETA=0` to all our maintenance jobs...

**Comment by pdostal** [comment]
What about this?
```yaml
  package_hub_dependent:
     FLAVOR:
       Server-DVD-Updates:
         - containers/docker_compose
         - containers/registry
     BETA:
       0:
         - containers/docker_compose
         - containers/registry
```

jlausuch:
VRs for engines and tools: 
https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_fix_15sp4&distri=sle&version=15-SP4

pdostal (approved):

schedule/containers/engines_and_tools_docker.yaml
```diff
@@ -20,21 +20,31 @@ conditional_schedule:
         - containers/docker_runc
   supported:
     HOST_VERSION:
+      15-SP4:
+        - containers/buildah_docker
```
jlausuch, line 24: I also had to add this... yet again another reason to avoid these ugly conditions if we could just have `if is_sle(>=15-SP1)` ...

PR #13006, opened 2021-07-30
Makefile: split commands and use xargs

perlpunk:
To avoid having a too long argument list


OleksandrOrlov (approved):
LGTM. Tina, thanks for you assistance!:)

PR #13009, opened 2021-08-02
Use mouse to login gdm if username is not highlighted.

zhaocongjia:
And remove the softfailure.
Upstream has changed the design of gdm, although the first
username in gdm is highlighted after boot, it is not always
like that under other circumstance.  So when the username
is not highlighted, use 'assert_and_click'.

Also update `dm_login` a bit to avoid using hard-coded value and remove outdated comment.

- Related ticket: https://progress.opensuse.org/issues/64947
- Needles: None
- Verification run:
 Tumbleweed: https://openqa.opensuse.org/tests/1865638
 SLE15SP4: https://openqa.suse.de/tests/6631109#step/gdm_session_switch/26

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -19,6 +19,7 @@ use LWP::Simple;
 use Config::Tiny;
 use utils;
 use version_utils qw(is_sle is_leap is_tumbleweed);
+use x11utils 'select_user_gnome';
 use POSIX 'strftime';
 use mm_network;
 
@@ -30,16 +31,13 @@ sub post_run_hook {
 
 sub dm_login {
     assert_screen('displaymanager', 60);
-    # The keyboard focus was losing in gdm of SLE15 bgo#657996
-    mouse_set(520, 350) if is_sle('15+');
-    send_key('ret');
+    select_user_gnome;
     assert_screen('originUser-login-dm');
     type_password;
 }
 
 # logout and switch window-manager
 sub switch_wm {
-    mouse_set(1000, 30);
     assert_and_click "system-indicator";
     assert_and_click "user-logout-sector";
     assert_and_click "logout-system";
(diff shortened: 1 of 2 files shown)


GraceWang571:
Does the code change work on the previous release (e.g. SLE12SP5 or SLE15SP3...)?
We need to make sure the code change won't break the maintenance test.

zhaocongjia:
Basically, this commit just removes a softfailure, which only happens sporadically.  So this should not break any older releases.

GraceWang571:
> Basically, this commit just removes a softfailure, which only happens sporadically. So this should not break any older releases.

Yeah, the thing is that we need to make sure the bsc#1086425 is really fixed on the potentially affected versions (e.g. TW, Leap, and SLE). That's to say, the status of bsc#1086425 is VERIFIED FIXED. 
And then, we are safe to remove the softfailure here.

zhaocongjia:
bsc#1086425 is fixed -- username is highlighted in gdm, but that is not stable for our testing purpose -- if the mouse cursor changes a little or if we switch console, then the focus can be lost.  So in testing we should use mouse as a fallback and don't consider this as softfail.

GraceWang571:
I know we can see from the  https://openqa.suse.de/tests/6631109/file/autoinst-log.txt that the `mouse_set` code is gone.
But the `gdm_session_switch` is not enabled for TW, so it might not be a good choice to use https://openqa.opensuse.org/tests/1865638 as a verification run for TW.

In terms of the softfailure part, it would be better to use https://openqa.suse.de/tests/6599163#step/first_boot/6 since it's still marked as softfailure due to bsc#1086425. Run the test suite with your code change and see if the softfailure disappears, and see if there are any side-effect.

zhaocongjia:
There seems no such softfail on TW.

The updated VR on SLE is: https://openqa.suse.de/tests/6645210#step/first_boot/5

PR #13011, opened 2021-08-02
Fix gnome-x11-ibus failure

GraceWang571:
GNOME40: input sources are now under 'keyboard'
Extract add_input_resource into x11test.pm
Update ibus_test_cn/kr/jp.pm to use add_input_resource
Simplify remove_input_source in ibus_clean.pm
Update ibus_clean.pm to use the new location of the input source

- Related ticket: https://progress.opensuse.org/issues/95120
- Needles: already added when debugging on o3
- Verification run: https://openqa.opensuse.org/tests/1865693

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -999,4 +999,32 @@ sub start_gnome_tweak_tool {
     }
 }
 
+# Since GNOME 40 or later, the 'Input Sources' is no longer in the 'Region & Language' panel
+# The 'Input Sources' is in the gnome-control-center 'Keyboard' panel now
+# Navigate to the 'Keyboard' panel first and then add the input source that to be tested
+sub add_input_resource {
+    my ($self, $tag) = @_;
+
+    # send_key 'super';
+    # wait_still_screen;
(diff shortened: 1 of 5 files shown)


DimStar77:
Is there no single product (maintenance/QR tests) running the test on something GNOME < 40?

GraceWang571:
> Is there no single product (maintenance/QR tests) running the test on something GNOME < 40?

Currently, the ibus tests are only enabled on TW.
We have a plan to enable in on SLE and will take the GNOME < 40 into consideration at that time.

GraceWang571:
Latest verification run:
TW: https://openqa.opensuse.org/tests/1870139
SLE15SP3: http://10.163.42.54/tests/1916

GraceWang571:
@DimStar77 I update the code so that it can handle GNOME < 40 situations. 
But I still haven't enabled it for SLE since I found a bug on SLE which is https://bugzilla.suse.com/show_bug.cgi?id=1189192
I'd like to enable it on SLESP4 maybe after the bug is fixed.

zhaocongjia:
LGTM.

PR #13020, opened 2021-08-03
boot_ltp: Take framebuffer console screenshot after boot

mdoucha:
This is debug code for [poo#81142](https://progress.opensuse.org/issues/81142). It'll gzip the contents of console framebuffer and upload them to OpenQA.

The framebuffer contains raw pixel data of what the kernel thinks the screen output should look like. The raw screenshot will not rule out a display device driver bug in the kernel but it will tell us whether the bug is in the higher level console rendering code.

- Related ticket: https://progress.opensuse.org/issues/81142
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/6784854

### DIFF
--- a/tests/kernel/boot_ltp.pm
+++ b/tests/kernel/boot_ltp.pm
@@ -40,6 +40,11 @@ sub run {
     }
 
     $self->select_serial_terminal;
+
+    # Debug code for poo#81142
+    script_run('gzip -9 </dev/fb0 >framebuffer.dat.gz');
+    upload_logs('framebuffer.dat.gz', failok => 1);
+
     assert_secureboot_status(1) if (get_var('SECUREBOOT'));
 
     log_versions;


mdoucha:
From the Logs & Assets tab you can download [boot_ltp-framebuffer.dat.gz](https://openqa.suse.de/tests/6641327/file/boot_ltp-framebuffer.dat.gz) which contains raw framebuffer data that should be identical to the `linux-login` screenshot in OpenQA. 

After downloading the framebuffer file, you can convert it to PNG using the following commands:

    gzip -d boot_ltp-framebuffer.dat.gz
    fb2png.py 1024 768 RGBX boot_ltp-framebuffer.dat

Where `fb2png.py` is the following Python script:

    #!/usr/bin/python3
    
    import sys
    from PIL import Image
    
    def convert_image(width, height, mode, filename):
        with open(filename, 'rb') as fr:
            img = Image.frombytes(mode, (width, height), fr.read())
        name = filename.rpartition('.')[0] + '.png'
        if mode != 'RGB':
            img = img.convert('RGB')
        img.save(name)
    
    if __name__ == '__main__':
        if len(sys.argv) < 5:
            print('Usage: %s width height pixelformat filename [...]' % sys.argv[0],
                file=sys.stderr)
            exit(1)
        width = int(sys.argv[1])
        height = int(sys.argv[2])
        for filename in sys.argv[4:]:
            convert_image(width, height, sys.argv[3], filename)

If the resulting PNG file looks distorted or the script throws an exception about image data length, you'll need to analyze the framebuffer pixel format yourself and adjust the script arguments accordingly.

tests/kernel/boot_ltp.pm
```diff
@@ -40,6 +40,11 @@ sub run {
     }
 
     $self->select_serial_terminal;
+
+    # Debug code for poo#81142
+    script_run('cat /dev/fb0 | gzip -9 >framebuffer.dat.gz');
```
okurz, line 45: how about putting that into the post_fail_hook? If you do that as the first action within the post_fail_hook nothing should change what is within the framebuffer

mdoucha: Your workaround needle prevents `boot_ltp` failure most of the time so `post_fail_hook()` will not be called.
https://openqa.suse.de/tests/6630340#step/install_ltp/3

This is the most reliable way to gather debug data right now. The gzipped framebuffer file is about 20-30KB in size so the disk overhead is negligible.

okurz: sure, fine.

kalikiana:
> Where `fb2png.py` is the following Python script:

Where did this script come from? I think this should be available somewhere in the repo, unless it's already packaged/hosted somewhere else.

mdoucha:
> > Where `fb2png.py` is the following Python script:
> 
> Where did this script come from? I think this should be available somewhere in the repo, unless it's already packaged/hosted somewhere else.

I wrote it to make the framebuffer data usable. Ideally, the screenshot would be generated by a C program like `fbcat` or `fbgrab` that can query framebuffer settings and write the screenshot in a standard image format but SLES doesn't ship either of those. And since I expect this PR to be reverted after a week or so after it finishes its purpose, I don't see the point of packaging the script.

PR #13022, opened 2021-08-03
MicroOS: allow failure for "audit -a" command

jlausuch:
audit -a returns !=0 if no events are found and the script_output fails the test.

- Failure: https://openqa.opensuse.org/tests/1865228#step/enable_selinux/33
- Verification run: https://openqa.opensuse.org/tests/1866577#live


cfconrad:
LGTM, 
by default there are just no `AVC Report` entries.

PR #13025, opened 2021-08-04
Fix host_configuration module against rhel

b10n1k:
Running dhclient on all-good networking of the rhel image is messing
with the route table.

Running dhclient shouldnt cause any problem on an already configured route table
IIRC and is propably distro specific issue and it can ignored for now and workaround
exclude the execution when HDD is rhel image.

Maybe it would be nice to have an self-agnostic method here and run config
commands when it is needed.

- Related ticket: https://progress.opensuse.org/issues/96116
- Verification run: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=b10n1k%2Fos-autoinst-distri-opensuse%2313025

### DIFF
--- a/tests/containers/host_configuration.pm
+++ b/tests/containers/host_configuration.pm
@@ -8,7 +8,7 @@
 # without any warranty.
 
 # Summary: Setup system which will host containers
-# - setup networking via dhclient
+# - setup networking via dhclient when is needed
 # - make sure that ca certifications were installed
 # - import SUSE CA certificates
 # Maintainer: qa-c team <qa-c@suse.de>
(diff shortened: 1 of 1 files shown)


tests/containers/host_configuration.pm
```diff
@@ -30,7 +30,7 @@ sub run {
         disable_and_stop_service(opensusebasetest::firewall, ignore_failure => 1);
     }
     else {
-        assert_script_run "dhclient -v";
+        assert_script_run "dhclient -v" unless get_var("HDD_1") =~ /rhel/;
```
grisu48, line 33: I'd add a comment here as well, so that we know in a month from know why we added this. Something like

    # Re-running dhclient on RHEL is confusing the routing tables

or something similar. Apart from that the PR looks good to me, I have troubles in understanding the VR. I'm assuming https://openqa.suse.de/tests/6645363 is the only relevant one? And the failure in `podman_image` there is not related to this PR.

b10n1k: > I'd add a comment here as well, so that we know in a month from know why we added this. Something like
> 
> ```
> # Re-running dhclient on RHEL is confusing the routing tables
> ```

Ready for review

> 
> or something similar. Apart from that the PR looks good to me, I have troubles in understanding the VR. I'm assuming https://openqa.suse.de/tests/6645363 is the only relevant one? And the failure in `podman_image` there is not related to this PR.

correct. the podman problem is different and i think there is separate ticket as it fails elsewhere as well. 
Also tests/6645363 is the main VR but i used to try to run a few random ones usually to catch side effect/breaking changes

grisu48 (approved):
LGTM! :rocket:

tests/containers/host_configuration.pm
jlausuch: Can we just just disable this module in rhel?

b10n1k: first sorry i didnt see this earlier. 
this is also possible. i will provide a commit even if that one is already merged.

PR #13026, opened 2021-08-04
Update size string when creating KVM backing image file 

mloviska:
SLEM z/KVM failures [sle-micro-5.1-MicroOS-Image-s390x-Build46.1_10.9-sle-micro_update@s390x-kvm-sle12](https://openqa.suse.de/tests/6645372) and [sle-micro-5.1-MicroOS-Image-s390x-Build46.1_10.9-sle-micro_qemu@s390x-kvm-sle12](https://openqa.suse.de/tests/6645371) occur due to wrong string size parameter in the caller.

```
[2021-08-04T09:10:59.446 CEST] [debug] [run_ssh_cmd(qemu-img create '/var/lib/libvirt/images/openQA-SUT-2a.img' -f qcow2 -b '/var/lib/libvirt/images//SUSE-MicroOS.s390x-5.1.0-Default-kvm-Build10.9.qcow2' 30GG)] stderr:
  qemu-img: Invalid image size specified! You may use k, M, G, T, P or E suffixes for
  qemu-img: kilobytes, megabytes, gigabytes, terabytes, petabytes and exabytes.
```

VRs:
- [sle-micro-5.1-MicroOS-Image-s390x-Build46.1_10.9-sle-micro_qemu@s390x-kvm-sle12](http://kepler.suse.cz/tests/6277#step/boot_to_desktop/1)
- [sle-micro-5.1-MicroOS-Image-s390x-Build46.1_10.9-sle-micro_update@s390x-kvm-sle12](http://kepler.suse.cz/tests/6278#)

### DIFF
--- a/tests/installation/bootloader_svirt.pm
+++ b/tests/installation/bootloader_svirt.pm
@@ -139,6 +139,7 @@ sub run {
         if (my $full_hdd = get_var('HDD_' . $n)) {
             my $hdd     = basename($full_hdd);
             my $hddpath = search_image_on_svirt_host($svirt, $hdd, $hdddir);
+            $size_i = get_var("HDDSIZEGB_$n") if get_var("HDDSIZEGB_$n");
             if ($hddpath =~ m/vmdk\.xz$/) {
                 my $nfs_ro = $hddpath;
                 $hddpath = "$vmware_openqa_datastore/$hdd" =~ s/vmdk\.xz/vmdk/r;
@@ -154,7 +155,8 @@ sub run {
                 {
                     backingfile => 1,
                     dev_id      => $dev_id,
-                    file        => $hddpath
+                    file        => $hddpath,
+                    size        => $size_i
                 });
         }
         else {
@@ -162,7 +164,7 @@ sub run {
                 {
                     create => 1,
                     dev_id => $dev_id,
-                    size   => $size_i . 'G'
+                    size   => $size_i
                 });
         }
         $dev_id = chr((ord $dev_id) + 1);    # return next letter in alphabet
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -1000,7 +1000,6 @@ sub zkvm_add_disk {
 
                 enter_cmd("# copying image ($basename)...");
                 if (my $size = get_var("HDDSIZEGB_$di")) {
-                    $size .= "G";
                     $svirt->add_disk({file => $hdd_path, backingfile => 1, dev_id => $dev_id, size => $size});
                 } else {
                     $svirt->add_disk({file => $hdd_path, backingfile => 1, dev_id => $dev_id});
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -1007,7 +1007,7 @@ sub zkvm_add_disk {
             } else {
                 # Create a new image, most likely it can be image for installation
                 # or additional optional drive for further testing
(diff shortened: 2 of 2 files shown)


mloviska:
Follow-up for https://github.com/os-autoinst/os-autoinst/pull/1736

tests/installation/bootloader_svirt.pm
```diff
@@ -139,6 +139,7 @@ sub run {
         if (my $full_hdd = get_var('HDD_' . $n)) {
             my $hdd     = basename($full_hdd);
             my $hddpath = search_image_on_svirt_host($svirt, $hdd, $hdddir);
+            $size_i = get_var("HDDSIZEGB_$n") if get_var("HDDSIZEGB_$n");
```
b10n1k, line 142: why not put this in L137
`my $size_i = get_var("HDDSIZEGB_$n") ? 
                      get_var("HDDSIZEGB_$n") : 
                      get_var('HDDSIZEGB', '10');`

or something like that...

mloviska: If `HDDSIZEGB_$n` is not set, it is better to keep the default image size, as we most likely have larger images than 10G. And in such a case the parameter would be ignored because the image won't be bootable if the backing file is smaller than the original image.
And I prefer to keep mapping `HDD_$n` to `HDDSIZEGB_$n`.

b10n1k (changes requested):
There are two commits.
- `Add arbitrary backing file size for svirt images` missing information which seems that it is included in the second one but it would be nice to keep it neat. 
- `Update size string when creating KVM backing image file` would you delete the VR links from the commit. i dont think are "valuable"(i cant come with a proper term) in the git log history.

grisu48 (approved):
LGTM :rocket: 

Took me a little bit to figure out, that the suffix `G` is added by the backend, so we don't loose it. The actual error is that `G` is added twice, resulting in something like `30GG` (as can be seen in Martin's error message on the right end).

tests/installation/bootloader_svirt.pm
b10n1k: Still i think it makes sense to have the assignment in one place. And you could work out the default case.
Something like that could be `my $size_i = get_var("HDDSIZEGB_$n") ? get_var("HDDSIZEGB_$n") : get_var('HDDSIZEGB') // 10;`
but it s up to you how would you like to implement it. 
The point is that the code is easier readable and it is easy-made change. wdyt?

mloviska:
> LGTM rocket
> 
> Took me a little bit to figure out, that the suffix `G` is added by the backend, so we don't loose it. The actual error is that `G` is added twice, resulting in something like `30GG` (as can be seen in Martin's error message on the right end).

Thanks, there might be a case I have missed, I will need to do more testing. Labelling this one as WIP for now.

Martchus (commented):
Since the variable is called `HDDSIZE*GB*`: Wouldn't it make sense to still add the `G` suffix by default?

mloviska:
> Since the variable is called `HDDSIZE*GB*`: Wouldn't it make sense to still add the `G` suffix by default?

Yes, it does and sorry for the mess! I am not really sure how to approach as in case of Qemu it is handled in the `os-autoinst` repo, and when it comes to `svirt` it is mostly mixed. It is a bit obscure situation.
https://github.com/os-autoinst/os-autoinst/blob/master/OpenQA/Qemu/Proc.pm#L188
https://github.com/os-autoinst/os-autoinst/blob/master/consoles/sshVirtsh.pm#L339

Martchus:
You could append the `G` only conditionally when `looks_like_number` from `Scalar::Util` returns true.

PR #13033, opened 2021-08-05
Add test for a failing migration from SLES to SLES4SAP and the successful rollback

bs-suse:
This test checks if a successful rollback happens on a failing migration from SLES to SLES4SAP in case that an invalid registration key is used.

- Verification run: http://keks.qa.suse.de/tests/206#step/migrate_sles_to_sles4sap/37
- Related ticket: https://jira.suse.com/browse/TEAM-2123


tests/sles4sap/migrate_sles_to_sles4sap.pm
```diff
@@ -48,8 +52,16 @@ sub run {
     send_key 'ret';
     wait_serial "Please enter your activation code", timeout => 5;
     enter_cmd "${regcode}";
-    assert_script_run "ls /tmp/OK";
-    zypper_call "in -y -t pattern sap_server";
+
+    # test either a failing migration or a working one
+    if ("${regcode}" eq "invalid_key") {
```
ricardobranco777, line 57: Perl is not Bash :)

I think just `$regcode` will work.

bs-suse:
The test as it worked before the change is as well still working: http://keks.qa.suse.de/tests/203#step/migrate_sles_to_sles4sap/37

ricardobranco777 (commented):
Don't know why we have Bashisms like `"${regcode} "`.

Otherwise LGTM

tests/sles4sap/migrate_sles_to_sles4sap.pm
bs-suse: Good point, I just wanted to be consistent with what was already there 3 lines above.

bs-suse: I changed both occurrences of the bashism :-)

alvarocarvajald (commented):
Some general comments:

- Due to the size and complexity of the changes, it will be better to squash all 4 commits into one.
- You can add in the description the Jira task as the related ticket. That way we have a reference as to where this test scenario was requested.

I have some comments below regarding the code, but it is looking good so far.

tests/sles4sap/migrate_sles_to_sles4sap.pm
alvarocarvajald: Sometimes "bashisms" are necessary:
```
# perl
$msg = 'Hello ';
print "Without bashisms: $msgWorld\n";
print "With    bashisms: ${msg}World\n";
Without bashisms: 
With    bashisms: Hello World
```
Or when doing some more complex data structures handling:
```
    for my $arg (@{$param_file}) {
        next unless $arg =~ $PARAM_RE;
        push @{$params{$1}}, path($2)->slurp;
    }
```
But agree with @ricardobranco777 .... it does look better without the curly braces. :)

ricardobranco777 (approved):

alvarocarvajald (approved):
LGTM

PR #13035, opened 2021-08-05
[WIP] Add utilities to manage esxi hosts and vms

nanzhg:
To manage ESXi hosts and VMs, creating a new module to wrap up vim-cmd commands and add some useful utilities for open-vm-tools tests.

1. Power operations (On, Off, Reboot, Reset, Shutdown)
2. Network information (Guest IP, VSwitch device binding)
3. Clock synchronization (Host timestamp, Disable time sync settings)

- Related ticket: https://progress.opensuse.org/issues/78810
- Verification run:  [To be added]

PR #13036, opened 2021-08-05
Add new tests for open-vm-tools against vmware esxi

nanzhg:
**[tests/virt_autotest/esxi_open_vm_tools.pm]** According to the test design [https://confluence.suse.com/display/~nanzhang/Test+design+of+open-vm-tools], adding 3 parts to extend this test module for testing open-vm-tools.
1. Power management tests
2. Networking tests
3. Clock Sync tests

**[lib/virt_autotest/esxi_utils.pm]** To manage ESXi hosts and VMs, creating a new module to wrap up vim-cmd commands and add some useful utilities for open-vm-tools tests.
1. Power operations (On, Off, Reboot, Reset, Shutdown)
2. Network information (Guest IP, VSwitch device binding)
3. Clock synchronization (Host timestamp, Disable time sync settings)

**[lib/main_common.pm]** Load the tests under the consoletests suite for vmware only. To enable this test without any impact on other console test suites, introduced a new parameter "OPEN_VM_TOOLS=1" as a temporary solution, and will use YAML job template instead.

- Related ticket:  https://progress.opensuse.org/issues/78810
- Verification run:  http://10.67.129.66/tests/356

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1085,6 +1085,8 @@ sub load_consoletests {
     loadtest "console/prepare_test_data";
     loadtest "console/consoletest_setup";
     loadtest 'console/integration_services' if is_hyperv || is_vmware;
+    # This is a temporary solution for enabling open-vm-tools tests, will use YAML job template instead.
+    loadtest 'virt_autotest/esxi_open_vm_tools' if is_vmware && get_var('OPEN_VM_TOOLS');
 
     if (get_var('IBM_TESTS')) {
         # prepare tarballs for the testcase
--- a/lib/virt_autotest/esxi_utils.pm
+++ b/lib/virt_autotest/esxi_utils.pm
@@ -1,0 +1,89 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Utilities for running ESXi commands
+# Maintainer: Nan Zhang <nan.zhang@suse.com>
+
+package virt_autotest::esxi_utils;
+
+use base Exporter;
+use Exporter;
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+our @EXPORT = qw(
+  esxi_vm_get_vmid
+  esxi_vm_power_getstate
+  esxi_vm_power_ops
+  esxi_vm_network_binding
+  esxi_vm_public_ip
+  get_host_timestamp
+  disable_vm_time_synchronization
+);
+
+sub esxi_vm_get_vmid {
+    my $vm_name = shift;
+    my $vim_cmd = "vim-cmd vmsvc/getallvms | grep -i $vm_name | awk -F' ' '{print \$1}'";
+    my (undef, $vmid) = console('svirt')->run_cmd($vim_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    chomp($vmid);
+    return $vmid;
+}
+
+sub esxi_vm_power_getstate {
+    my $vmid    = shift;
+    my $vim_cmd = "vim-cmd vmsvc/power.getstate";
+    my (undef, $power_state) = console('svirt')->run_cmd("$vim_cmd $vmid", domain => 'sshVMwareServer', wantarray => 1);
+    return $power_state;
+}
+
+sub esxi_vm_power_ops {
+    my ($vmid, $powerops) = @_;
+    my $vim_cmd = "vim-cmd vmsvc/$powerops";
+    return console('svirt')->run_cmd("$vim_cmd $vmid", domain => 'sshVMwareServer', wantarray => 1);
+}
+
+sub esxi_vm_network_binding {
+    my $vmid    = shift;
+    my $vim_cmd = "vim-cmd vmsvc/get.environment $vmid | grep vswitch | sed -n 1p | cut -d'\"' -f2";
+    my (undef, $vswitch) = console('svirt')->run_cmd($vim_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    return $vswitch;
+}
+
+sub esxi_vm_public_ip {
+    my $vmid    = shift;
+    my $vim_cmd = "vim-cmd vmsvc/get.guest $vmid | grep ipAddress | sed -n 1p | cut -d'\"' -f2";
+    my (undef, $vm_ip) = console('svirt')->run_cmd($vim_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    return $vm_ip;
+}
+
+sub get_host_timestamp {
+    my $date_cmd = shift // "date -u +'\%F \%T'";    # Default to get UTC time
+    my (undef, $host_time) = console('svirt')->run_cmd($date_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    return $host_time;
+}
+
+sub disable_vm_time_synchronization {
+    my $vm_name  = shift;
+    my $vmx_file = "/vmfs/volumes/datastore1/openQA/$vm_name.vmx";
+
+    # Set all time synchronization properties to FALSE
+    console('svirt')->run_cmd("sed -ie 's/tools.syncTime.*/tools.syncTime=\"FALSE\"/' $vmx_file", domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.continue=\"FALSE\" >> $vmx_file",            domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.restore=\"FALSE\" >> $vmx_file",             domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.resume.disk=\"FALSE\" >> $vmx_file",         domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.shrink=\"FALSE\" >> $vmx_file",              domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.tools.startup=\"FALSE\" >> $vmx_file",       domain => 'sshVMwareServer', wantarray => 1);
+    console('svirt')->run_cmd("echo time.synchronize.resume.host=\"FALSE\" >> $vmx_file",         domain => 'sshVMwareServer', wantarray => 1);
+}
+
+1;
--- a/tests/virt_autotest/esxi_open_vm_tools.pm
+++ b/tests/virt_autotest/esxi_open_vm_tools.pm
@@ -1,0 +1,293 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Run open-vm-tools testing against VMware ESXi
+# Maintainer: Nan Zhang <nan.zhang@suse.com>
+
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+use virt_autotest::esxi_utils;
+use Time::Local;
+
+our $VM_POWER_ON  = 'Powered on';
+our $VM_POWER_OFF = 'Powered off';
+
+sub run {
+    my $vm_name = console('svirt')->name;
+    my $vm_id   = esxi_vm_get_vmid($vm_name);
+    my $vm_ip   = esxi_vm_public_ip($vm_id);
+    chomp($vm_ip);
+
+    record_info('Guest Name', $vm_name);
+    record_info('Guest ID',   $vm_id);
+    record_info('Guest IP',   $vm_ip);
+
+    die "The variable \$vm_id or \$vm_ip cannot be empty." if ($vm_id eq "" || $vm_ip eq "");
+
+    do_sanity_checks();
+    do_power_management_tests($vm_id, $vm_ip);
+    do_networking_tests($vm_id, $vm_ip);
+    do_clock_sync_tests($vm_name, $vm_id, $vm_ip);
+}
+
+sub do_sanity_checks {
+    assert_script_run('zypper -n in open-vm-tools') if (script_run('rpmquery open-vm-tools'));
+
+    assert_script_run("/usr/bin/vmware-checkvm | grep 'good'");
+
+    systemctl('status vmtoolsd');
+    systemctl('status vgauthd');
+
+    systemctl('restart vmtoolsd');
+    assert_script_run("systemctl status vmtoolsd | grep 'Started open-vm-tools'");
+
+    assert_script_run("/usr/bin/vmtoolsd -v | grep 'VMware Tools daemon, version'");
+
+    assert_script_run('vmware-toolbox-cmd logging level set vmtoolsd message');
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = message'");
+
+    assert_script_run('vmware-toolbox-cmd logging level set vmtoolsd debug');
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = debug'");
+}
+
+sub take_vm_power_ops {
+    # $powerops:     this value is from the esxi server command 'vim-cmd vmsvc/'
+    # $if_vm_state:  $VM_POWER_ON or $VM_POWER_OFF; check if it's the current vm power state
+    my ($vm_id, $powerops, $if_vm_state) = @_;
+
+    if (esxi_vm_power_getstate($vm_id) =~ m/$if_vm_state/) {
+        if (esxi_vm_power_ops($vm_id, $powerops)) {
+            return 0;
+        } else {
+            return 1;
+        }
+    }
+}
+
+sub check_vm_power_state {
+    # $powerops_ret:     return value from the subroutine esxi_vm_power_ops
+    # $if_pingable:      1 or 0; check if the vm network is pingable or not
+    # $expect_vm_state:  $VM_POWER_ON or $VM_POWER_OFF; expected vm power state after taking power action
+    my ($vm_id, $vm_ip, $powerops, $powerops_ret, $if_pingable, $expect_vm_state) = @_;
+
+    if ($powerops_ret) {
+        wait_for_vm_network($vm_ip, $if_pingable);
+    } elsif ($powerops_ret == undef) {
+        return;
+    } else {
+        die "Not the correct VM state for guest $powerops.";
+    }
+    die "Failed to $powerops the guest." if (!esxi_vm_power_getstate($vm_id) =~ m/$expect_vm_state/);
+}
+
+sub wait_for_vm_network {
+    my $vm_ip       = shift;         # vm ip address
+    my $if_pingable = shift;         # 1 or 0; check if the vm network is pingable or not
+    my $times       = shift // 9;    # the number of ping execution
+    my $count       = shift // 3;    # ping waits for the count
+    my $interval    = shift // 5;    # wait secs between each ping
+    my $cmd;
+
+    # This ping command is used for waiting VM bootup or shutdown
+    if ($if_pingable) {
+        $cmd = "ping -c$count -i$interval $vm_ip";
+    } else {
+        $cmd = "! ping -c$count -i$interval $vm_ip";
+    }
+
+    # Monitor the networking after bring up/down the VM
+    while ($times && console('svirt')->run_cmd($cmd)) {
+        $times--;
+    }
+}
+
+sub login_vm_console {
+    reset_consoles;
+    console('svirt')->start_serial_grab;
+    select_console('sut');
+    assert_screen('grub2', 90);
+    send_key 'ret';
+    assert_screen('linux-login', 90);
+    select_console('root-console');
+}
+
+sub get_epoch_time {
+    my ($datetime, $epoch, $sec, $min, $hour, $mday, $mon, $year);
+    $datetime = $_[0];
+
+    my @datetime = reverse(split /-|:|\s|\//, $datetime);
+    foreach (@datetime) {
+        s/^0//;
+        $epoch .= $_ . ",";
+    }
+    $epoch =~ s/,$//;
+    ($sec, $min, $hour, $mday, $mon, $year) = split /,/, $epoch;
+    $year -= 1900;
+    $mon  -= 1;
+    $epoch = timelocal($sec, $min, $hour, $mday, $mon, $year);
+
+    return $epoch;
+}
+
+sub init_guest_time {
+    my $h_datetime = get_host_timestamp();
+    # Get last day of the host time
+    my $last_day = script_output("date -u -d '$h_datetime last day' +'\%F \%T'");
+    # Set the guest time by using the variable $last_day
+    my $g_datetime = script_output("date -u -s '$last_day' +'\%F \%T'");
+
+    return $g_datetime;
+}
+
+sub get_diff_seconds {
+    my $h_datetime = get_host_timestamp();
+    my $g_datetime = script_output("date -u +'\%F \%T'");
+    my $h_timesec  = get_epoch_time($h_datetime);
+    my $g_timesec  = get_epoch_time($g_datetime);
+    my $diff_secs  = abs(int($g_timesec - $h_timesec));
+
+    record_info('Host time',                       $h_datetime);
+    record_info('Guest time after sync with host', $g_datetime);
+    record_info('Diff secs',                       $diff_secs);
+
+    return $diff_secs;
+}
+
+sub do_power_management_tests {
+    my ($vm_id, $vm_ip) = @_;
+    my ($powerops, $powerops_ret);
+
+    record_info('Power Manangement Tests');
+    select_console('svirt');
+
+    record_info('Guest Power Shutdown');
+    $powerops     = 'power.shutdown';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_ON);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 0, $VM_POWER_OFF);
+
+    record_info('Guest Power On');
+    $powerops     = 'power.on';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_OFF);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    record_info('Guest Power Restart');
+    $powerops     = 'power.restart';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_ON);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    record_info('Guest Power Reset');
+    $powerops     = 'power.reset';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_ON);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    record_info('Guest Power Off');
+    $powerops     = 'power.off';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_ON);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 0, $VM_POWER_OFF);
+}
+
+sub do_networking_tests {
+    my ($vm_id, $vm_ip) = @_;
+    my ($powerops, $powerops_ret);
+
+    record_info('Networking Tests');
+    # Boot up the VM if it's powered off
+    $powerops     = 'power.on';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_OFF);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    my $vswitch_name = esxi_vm_network_binding($vm_id);
+    record_info('VSwitch Name', $vswitch_name);
+
+    login_vm_console();
+
+    # Test network via assigned IPv4 address
+    assert_script_run("ping -I $vm_ip -4 -c3 openqa.suse.de");
+    assert_script_run("ping -I $vm_ip -4 -c3 www.suse.com");
+}
+
+sub do_clock_sync_tests {
+    my ($vm_name, $vm_id, $vm_ip) = @_;
+    my ($powerops, $powerops_ret, $g_init_time, $diff_secs);
+
+    record_info('Clock Sync Tests');
+    # Boot up the VM if it's powered off
+    $powerops     = 'power.on';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_OFF);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    select_console('sut') if (current_console() ne 'sut');
+
+    # Set the last day of host time in guest before enabling timesync service
+    $g_init_time = init_guest_time();
+    record_info('Guest current time before time sync enabled', $g_init_time);
+
+    # Enable timesync service in guest VM
+    if (script_output('vmware-toolbox-cmd timesync status', proceed_on_failure => 1, timeout => 90) eq 'Disabled') {
+        assert_script_run('vmware-toolbox-cmd timesync enable', timeout => 90);
+    }
+
+    # Check the time difference between host and guest after timesync service enabled
+    $diff_secs = get_diff_seconds();
+
+    if ($diff_secs < 10) {
+        record_info('Clock synchronization is successful.');
+    } else {
+        die 'Clock synchronization failed.';
+    }
+
+    # Disable clock sync tests
+    select_console('svirt');
+
+    $powerops     = 'power.shutdown';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_ON);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 0, $VM_POWER_OFF);
+
+    disable_vm_time_synchronization($vm_name);
+
+    $powerops     = 'power.on';
+    $powerops_ret = take_vm_power_ops($vm_id, $powerops, $VM_POWER_OFF);
+    check_vm_power_state($vm_id, $vm_ip, $powerops, $powerops_ret, 1, $VM_POWER_ON);
+
+    login_vm_console();
+
+    # Set guest time with a given time which different from host time
+    $g_init_time = init_guest_time();
+    record_info('Set guest time after time sync disabled', $g_init_time);
+
+    # Guest time will not be synced up with host after timesync service disabled
+    $diff_secs = get_diff_seconds();
+
+    if ($diff_secs > 86400 && $diff_secs < 86420) {
+        record_info('Clock synchronization was disabled successfully.');
+    } else {
+        die 'Disabling clock synchronization failed.';
+    }
+}
+
+sub post_fail_hook {
+    select_console 'log-console';
+
+    # Upload open-vm-tools backend service logs
+    upload_logs '/var/log/vmware-network.log';
+    upload_logs '/var/log/vmware-vgauthsvc.log.0';
+    upload_logs '/var/log/vmware-vmsvc-root.log';
+    upload_logs '/var/log/vmware-vmtoolsd-root.log';
+
+    # Upload guest diagnostic information
+    if (script_run('/usr/bin/vm-support') == 0) {
+        my $vm_logs_tarball = script_output('ls vm-*.tar.gz');
+        upload_logs $vm_logs_tarball;
+    }
+}
+
+1;
--- a/tests/virtualization/universal/open_vm_tools.pm
+++ b/tests/virtualization/universal/open_vm_tools.pm
@@ -22,6 +22,7 @@ use virt_autotest::common;
 sub run {
     my ($self) = @_;
 
+    # Note: This test module will be deprecated, and the following test code has been moved to tests/virt_autotest/esxi_open_vm_tools.pm
     script_retry "ssh root\@$_ zypper -n in open-vm-tools", delay => 30, retry => 6 foreach (keys %virt_autotest::common::guests);
 
     assert_script_run "ssh root\@$_ /usr/bin/vmware-checkvm | grep 'good'" foreach (keys %virt_autotest::common::guests);


Julie-CAO:
I viewed the implementation of the vm-tools test, some concerns or questions: 

- Why do you create a separate block for "OPEN_VM_TOOLS"?  Is your test not compatible with qem tests? If so, the existing vm tools test written by qem is workable in your tests?
- The two PRs(this pr and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13035) implement one thing, they should be submitted in on PR.
- you created esxi_utils.pm under lib/, but the functions in it are all specific to vm guests. So I think lib/virt_auto/ is the better place.

nanzhg:
@alice-suse @guoxuguang @Julie-CAO @waynechen55 please help to review this PR. Thanks

nanzhg:
> I viewed the implementation of the vm-tools test, some concerns or questions:
> 
> * Why do you create a separate block for "OPEN_VM_TOOLS"?  Is your test not compatible with qem tests? If so, the existing vm tools test written by qem is workable in your tests?

From the existing test code, suppose qem tests based on mutliple guests. Our checks are starting from guest installation, and it's only one guest. So I don't think it can compatible with qem test cases. Just utilize this test module and run with our specific needs.

> * The two PRs(this pr and [[WIP] Add utilities to manage esxi hosts and vms #13035](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13035)) implement one thing, they should be submitted in on PR.

It's ok to combine into one PR.

> * you created esxi_utils.pm under lib/, but the functions in it are all specific to vm guests. So I think lib/virt_auto/ is the better place.

All those functions in esxi_utils.pm are needed to be run on ESXi server, that means they are server side commands.

Julie-CAO:
> > I viewed the implementation of the vm-tools test, some concerns or questions:
> > 
> > * Why do you create a separate block for "OPEN_VM_TOOLS"?  Is your test not compatible with qem tests? If so, the existing vm tools test written by qem is workable in your tests?
> 
> From the existing test code, suppose qem tests based on mutliple guests. Our checks are starting from guest installation, and it's only one guest. So I don't think it can compatible with qem test cases. Just utilize this test module and run with our specific needs.
> 
Reuse is a principle for us to write tests, especially for tests on the same virtualization tool. Reusing tests with qem team can extend test coverage and save effort for each team.

I don't think that The use of multiple guests by qem and starting guest installation by us you mentioned above are the reason why we are unable to be compatible with QEM test, because these two problems can be solved.

> > * you created esxi_utils.pm under lib/, but the functions in it are all specific to vm guests. So I think lib/virt_auto/ is the better place.
> 
> All those functions in esxi_utils.pm are needed to be run on ESXi server, that means they are server side commands.

Even it is server-side functions, but they all control guests. they are virtualization specific rather than common libs.

nanzhg:
@Julie-CAO OK, will update with a new version. Thanks

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+use utils;
+use virt_autotest::esxi_utils;
+use Time::Local;
+
+sub run {
+    assert_script_run('rpmquery open-vm-tools');
+    systemctl('status vmtoolsd');
+    systemctl('status vgauthd');
+
+    my $vm_name = console('svirt')->name;
```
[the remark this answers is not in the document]
nanzhg: Yes, this test is designed for run testing after one guest installation. For now, these are required info for the test running like name, id and ip. We can add more if there is any other requirement.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+    $h_timesec  = get_timelc($h_datetime);
+    $diff_sec   = int($g_timesec - $h_timesec);
+
+    record_info('Diff secs', $diff_sec);
+    if (abs($diff_sec) > 300) {
+        record_info("Clock synchronization was disabled.");
+    } else {
+        die "Disabling clock synchronization failed.";
+    }
+}
```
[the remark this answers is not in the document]
nanzhg: Yes, post_fail_hook was triggered from the parent method and related logs were saved, refer to the job http://10.67.129.66/tests/243#downloads.

[2021-08-27T20:59:48.128 CST] [debug] lib/consoletest.pm:53 called opensusebasetest::post_fail_hook -> lib/opensusebasetest.pm:1315 called utils::show_tasks_in_blocked_state -> lib/utils.pm:1467 called testapi::send_key

lib/main_common.pm
```diff
@@ -1085,7 +1085,8 @@ sub load_consoletests {
     loadtest "console/system_state";
     loadtest "console/prepare_test_data";
     loadtest "console/consoletest_setup";
-    loadtest 'console/integration_services' if is_hyperv || is_vmware;
+    loadtest 'console/integration_services'     if is_hyperv || is_vmware;
+    loadtest "virt_autotest/esxi_open_vm_tools" if is_vmware;
```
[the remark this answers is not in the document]
nanzhg: OK, to enable this test without any impact on console test suite, will introduce a new parameter "OPEN_VM_TOOLS=1" as the temporary solution.

lib/virt_autotest/esxi_utils.pm
```diff
@@ -0,0 +1,116 @@
...
+    my (undef, $vm_ip) = console('svirt')->run_cmd($vim_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    return $vm_ip;
+}
+
+sub get_host_timestamp {
+    my (undef, $host_time) = console('svirt')->run_cmd("date -u +'\%Y-\%m-\%d \%H:\%M:\%S'", domain => 'sshVMwareServer', wantarray => 1);
+    return $host_time;
+}
+
+sub disable_all_clock_synchronization {
```
alice-suse, line 102: Do I get it correctly that this function disable clock sync on a single vm? If yes, suggest to spin vm into the function name.

alice-suse, line 102: Do I get it correctly that this function disable clock sync on a single vm? If yes, suggest to spin vm into the function name.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+    $timelc =~ s/,$//;
+    ($sec, $min, $hour, $mday, $mon, $year) = split /,/, $timelc;
+    $year -= 1900;
+    $mon  -= 1;
+    $timelc = timelocal($sec, $min, $hour, $mday, $mon, $year);
+
+    return $timelc;
+}
+
+sub power_management_tests {
```
alice-suse, line 59: For sub/function name, it is better to be an action, rather than a noun, eg this sub can be named as do_power_management_tests. So does others.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+        while ($num && console('svirt')->run_cmd("ping -c1 $vm_ip")) {
+            sleep(10);
+            $num--;
+        }
+    }
+
+    my $vswitch_name = esxi_vm_network_binding($vm_id);
+    record_info('VSwitch Name', $vswitch_name);
+
+    reset_consoles;
```
alice-suse, line 131: I am not sure whether I get this code block(L131~138) correctly. Would you please help explain what's the purpose for the block and what are done to achieve this? BTW is this the part that are blocked by the svirt backend issue which you mentioned in today's team meeting?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+    my ($g_datetime, $h_datetime, $g_difftime);
+
+    # Enable timesync service in guest VM
+    if (script_output("vmware-toolbox-cmd timesync status", proceed_on_failure => 1, timeout => 90) == "Disabled") {
+        $g_datetime = script_output("date -u +'\%Y-\%m-\%d \%H:\%M:\%S'");
+        record_info('Guest current time before syncing up', $g_datetime);
+        assert_script_run("vmware-toolbox-cmd timesync enable", timeout => 90);
+        $h_datetime = get_host_timestamp();
+        record_info('Host time', $h_datetime);
+        $g_datetime = script_output("date -u +'\%Y-\%m-\%d \%H:\%M:\%S'");
```
alice-suse, line 152: Is it correct that this function should be executed directly on the vm's console? It seems in code you have switched to "select_console("sut");" in networking_tests, so here the code to switch to sut console is not given, right? If yes, I'd suggest to give a comment for it, to let people more easily understand the test code.

An even better way is to decouple the close relationship(or env dependency) between network test and clock test, so that they can be more flexible.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+        record_info('Guest time after syncing up with host', $g_datetime);
+    }
+
+    my $h_timesec = get_timelc($h_datetime);
+    my $g_timesec = get_timelc($g_datetime);
+    my $diff_sec  = int($g_timesec - $h_timesec);
```
alice-suse, line 158: How fast is the clock syncing? From code, after enabling clock sync, the host and guest time are retrieved one by one immediately, without dedicated time left for the syncing. Is this stable?

lib/virt_autotest/esxi_utils.pm
nanzhg: Updated the function name with "disable_vm_time_synchronization".

nanzhg: Updated the function name with "disable_vm_time_synchronization".

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, adding prefix "do_" to all test functions.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: The purpose of this code block is that running ping commands inside the vm to check its guest network. We have to select vm console and do some login actions after vm booted up, then we could run those ping commands.
Actually, the ping command(assert_script_run("ping -c3 www.suse.com")) was blocked after vm shutdown and booted up(Referring defect #97517).

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+    # Guest time will not be synced up with the hypervisor after time synchronization disabled
+    $g_datetime = script_output("date -u +'\%Y-\%m-\%d \%H:\%M:\%S'");
+    $h_datetime = get_host_timestamp();
+    $g_timesec  = get_timelc($g_datetime);
+    $h_timesec  = get_timelc($h_datetime);
+    $diff_sec   = int($g_timesec - $h_timesec);
+
+    record_info('Diff secs', $diff_sec);
+    if (abs($diff_sec) > 300) {
+        record_info("Clock synchronization was disabled.");
```
alice-suse, line 212: I really like the above design that tests disabled clock sync. Can similar way (set differentiated time on guest, then start clock sync, then check the time diff with host)  be used to test enabled clock sync? IMHO, it will be better and more accurate. BTW is guest reboot necessary here?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+            sleep(10);
+            $num--;
+        }
+    }
+
+    reset_consoles;
+    select_console('sut');
+    assert_screen("grub2", 90);
+    send_key 'ret';
+    select_console("root-console", timeout => 180);
```
alice-suse, line 194: This code block(L190~194) seems to repeat, suggest to put it in a function.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+
+    select_console('svirt');
+
+    # Guest Power Off
+    if (esxi_vm_power_getstate($vm_id) =~ m/Powered on/) {
+        record_info('Guest Power Off');
+        esxi_vm_power_off($vm_id);
+        my $num = 6;
+        while ($num && console('svirt')->run_cmd("! ping -c1 $vm_ip")) {
+            sleep(10);
```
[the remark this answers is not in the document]
nanzhg: Thanks. It's a good idea! I've encapsulated the code block as a function "wait_for_vm_bootup" for reuse.

lib/main_common.pm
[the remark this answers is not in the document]
Julie-CAO: "OPEN_VM_TOOLS=1" can be a temporary solution, we will have better solution in the future, just to enable integration tests and open-vm-tools tests in our virtualization tests.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, I've put it in a function called "login_vm_console".

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Yes, we can do that before enabling clock sync service.
From the official document, power off guest is the prerequisites before disabling clock sync. Referring to the link below.
https://docs.vmware.com/en/VMware-Tools/11.2.5/com.vmware.vsphere.vmwaretools.doc/GUID-678DF43E-5B20-41A6-B252-F2E13D1C1C49.html

lib/main_common.pm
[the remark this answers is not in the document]
nanzhg: Yes, we will use YAML job template for instead in future.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: As tested from my local openqa, it will sync immediately once enabling the timesync service, and it's stable enough. But anyway we could set a time on guest before retrieving which different from the host.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Yes, it's the open-vm-tools function within the guest.  Maybe adding a judgment to check the current console before test running can decouple the relationship between each tests.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,218 @@
...
+            $num--;
+        }
+    }
+    # Guest Power On
+    if (esxi_vm_power_getstate($vm_id) =~ m/Powered off/) {
+        record_info('Guest Power On');
+        esxi_vm_power_on($vm_id);
+        my $num = 6;
+        while ($num && console('svirt')->run_cmd("ping -c1 $vm_ip")) {
```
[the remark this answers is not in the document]
nanzhg: The statement console('svirt')->run_cmd only returns 0 or not zero value. Added below statements follow by the "if branch" to confirm the vm state before running power actions.
i.e.
```
    } else {
        die "Not the correct VM state for testing Guest Power Off.";
    }
```
Originally, I considered this problem. But actually this test is designed to run after a new guest install, the only pre-condition is to check if the required package is installed correctly and its service status is in running without errors at the background. On the other hand, if the guest is in abnormal state, I think the pre-condition will post failures before the test beginning.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Added following statements to ensure the test is under the correct console.
`    record_info('Clock Sync Tests');
     select_console("sut") if (current_console() != "sut");
`

nanzhg:
@alice-suse @guoxuguang @Julie-CAO @waynechen55 I've updated as per comments, please help to review again. Thanks

lib/virt_autotest/esxi_utils.pm
```diff
@@ -0,0 +1,117 @@
...
+    my $vmid    = shift;
+    my $vim_cmd = "vim-cmd vmsvc/power.reset";
+    return console('svirt')->run_cmd("$vim_cmd $vmid", domain => 'sshVMwareServer', wantarray => 1);
+}
+
+sub esxi_vm_power_shutdown {
+    my $vmid    = shift;
+    my $vim_cmd = "vim-cmd vmsvc/power.shutdown";
+    return console('svirt')->run_cmd("$vim_cmd $vmid", domain => 'sshVMwareServer', wantarray => 1);
+}
```
waynechen55, line 81: What about aggregate all these functions to one ? esxi_vm_power_cycle which takes $vim_cmd or 'shutdown/reset' as arguments ?

nanzhg: Thanks, that's a good idea! Already aggregated them into one function, refer to the latest commit.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,214 @@
...
+    record_info('Networking Tests');
+    check_power_state($vm_id, $vm_ip, 'power.on', 1, $VM_POWERED_OFF, $VM_POWERED_ON);
+
+    my $vswitch_name = esxi_vm_network_binding($vm_id);
+    record_info('VSwitch Name', $vswitch_name);
+
+    login_vm_console();
+
+    assert_script_run('ping -c3 openqa.suse.de');
+    assert_script_run('ping -c3 www.suse.com');
```
waynechen55, line 150: I am not very sure what are you testing here. Because it seems that some services start to provide services with ipv6 address. So your ping may also get a response from ipv6 address. And if your guest has multiple interfaces, it is very difficult to tell the interface on which icmp reply is received. What about make this ping test more specific ? For example, use some options like -4, -I and many others ?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,214 @@
...
+    check_power_state($vm_id, $vm_ip, 'power.on', 1, $VM_POWERED_OFF, $VM_POWERED_ON);
+
+    select_console('sut') if (current_console() ne 'sut');
+
+    # Set the last day of host time within guest before enabling timesynce
+    $g_datetime = set_guest_time();
+    record_info('Guest current time before syncing up', $g_datetime);
+
+    # Enable timesync service in guest VM
+    if (script_output('vmware-toolbox-cmd timesync status', proceed_on_failure => 1, timeout => 90) eq 'Disabled') {
```
waynechen55, line 167: I recommend to use proceed_on_failure => 0 here. "vmware-toolbox-cmd timesync status" should always return the actual status. If not, then there is a problem. 
So I think:
my $variable = script_output('vmware-toolbox-cmd timesync status', proceed_on_failure => 0, timeout => 90);
if ($variable eq 'Disabled')
is better.

lib/virt_autotest/esxi_utils.pm
```diff
@@ -0,0 +1,93 @@
...
+  get_host_timestamp
+  disable_vm_time_synchronization
+);
+
+sub esxi_vm_get_vmid {
+    my $vm_name = shift;
+    my $vim_cmd = "vim-cmd vmsvc/getallvms | grep -i $vm_name | awk -F' ' '{print \$1}'";
+    my (undef, $vmid) = console('svirt')->run_cmd($vim_cmd, domain => 'sshVMwareServer', wantarray => 1);
+    chomp($vmid);
+    return $vmid;
```
waynechen55, line 43: It seems that console('svirt')->run_cmd is api which is called very often. So is there any possibility that it returns empty or not the thing you want ? If this happens, your test can run as expected.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,214 @@
...
+
+sub login_vm_console {
+    reset_consoles;
+    select_console('sut');
+    assert_screen('grub2', 90);
+    send_key 'ret';
+    select_console('root-console', timeout => 180);
+}
+
+sub set_guest_time {
```
waynechen55, line 84: Although you only use this function after login_vm_console, this functions seems can be used everywhere. To be honest, the function name does not match with what it is supposed to do.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,214 @@
...
+    my @time = reverse(split /-|:|\s|\//, $time);
+    foreach (@time) {
+        s/^0//;
+        $timelc .= $_ . ",";
+    }
+    $timelc =~ s/,$//;
+    ($sec, $min, $hour, $mday, $mon, $year) = split /,/, $timelc;
+    $year -= 1900;
+    $mon  -= 1;
+    $timelc = timelocal($sec, $min, $hour, $mday, $mon, $year);
```
waynechen55, line 106: What does $timelc stand for ? lc for local ? local clock ?

lib/virt_autotest/esxi_utils.pm
nanzhg: It's used to run server side commands, and it depends on the input parameters. I think it's unlikely to return an empty value if the input is correct.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Actually, this is the command provided from open-vm-tools, and it will return '69' if its service status is 'Disabled'. So this is a specific return value, but we would not expect this command runs into a failure.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Yes, you are right. We should use the option "-4" for ping via ipv4 address. This test is just for testing if assigned ip can work normally. Besides, the guest is designed and assigned only one network interface.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: How about the function name 'set_random_time'?

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: It doesn't stand for a local time, the function timelocal() converts a formatted time to total seconds. Here it's just a variable name, no special meaning. It is only used to receive the value returned from the function timelocal().

waynechen55: So 'lc" comes out of nothing ? If this is the case, why not use something more meaningful.

lib/virt_autotest/esxi_utils.pm
waynechen55: I see. Maybe this is just used this way for a long time. I still think it would be better to check whether the returned value is empty from run_cmd. This can be done in the function that tries to retrieve a value or the function that is going to use this value. Or in the run_cmd itself to check whether non-emtpy value is obtained. Or obtained empty value is allowed in run_cmd but it will fail or an option like the one used in script_output, proceed_on_failure, should b introduced. It gives me the feeling that run_cmd is not very good at error handling which should be enhanced from my POV.

tests/virt_autotest/esxi_open_vm_tools.pm
waynechen55: I see. If I understand correctly, this test will enable timesync anyway if "the if" fails or true.

nanzhg: Yes, that's right. We just want to match the string of the service status via this command.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, using  the variable name 'second_str' instead of 'timelc', and changing the function name to "get_second_str()".

waynechen55: If this generates and returns epoch time, why not use get_time_epoch ?

lib/virt_autotest/esxi_utils.pm
nanzhg: OK, I can understand your concern. It makes sense to check if any empty value which has impact during testing. Actually we already have part of inspection for the return value, such as esxi_vm_power_getstate() invoked in the function check_power_state(). The function run_cmd() is a wrapper for baseclass:run_ssh_cmd(),  seems we can't do more on it. So I would suggest to add related checks in the test function. As I said, the expected result we could get if the input is correct. My opinion is to ensure that the key input parameter is not empty string like $vm_id which is frequently used as a input parameter with run_cmd(), and such checking can be added at beginning of the tests.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Great idea! Thanks for the suggestion. Will update with the name.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,209 @@
...
+    my ($vm_id, $vm_ip, $power_action, $if_pingable, $check_vm_state, $if_vm_state, $expect_vm_state) = @_;
+
+    if (esxi_vm_power_getstate($vm_id) =~ m/$if_vm_state/) {
+        esxi_vm_power_cycle($vm_id, $power_action);
+        wait_for_vm_bootup($vm_ip, $if_pingable);
+    } elsif ($check_vm_state) {
+        die "Not the correct VM state for guest $power_action.";
+    }
+    die "Failed to $power_action the guest." if (!esxi_vm_power_getstate($vm_id) =~ m/$expect_vm_state/);
+}
```
waynechen55, line 54: I was wondering would you please go through all your functions and check whether they serve host or guest and indicate this in their names ? For example, this one check_power_state to check_guest_power_state  or check_vm_power_state ?

nanzhg: Updated.

lib/virt_autotest/esxi_utils.pm
waynechen55: I see. You can go with your current way of checking. What I am concerned comes from what a API function is supposed to be.

lib/main_common.pm
```diff
@@ -1087,6 +1087,8 @@ sub load_consoletests {
     loadtest "console/prepare_test_data";
     loadtest "console/consoletest_setup";
     loadtest 'console/integration_services' if is_hyperv || is_vmware;
+    # This is a temporary solution for enabling open-vm-tools tests, will use YAML job template instead.
```
waynechen55, line 1090: There will be a standalone open_vm_tools test suite in yaml job template ?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,209 @@
...
+    do_clock_sync_tests($vm_name, $vm_id, $vm_ip);
+}
+
+sub check_vm_power_state {
+    my ($vm_id, $vm_ip, $power_action, $if_pingable, $check_vm_state, $if_vm_state, $expect_vm_state) = @_;
+
+    if (esxi_vm_power_getstate($vm_id) =~ m/$if_vm_state/) {
+        esxi_vm_power_cycle($vm_id, $power_action);
+        wait_for_vm_bootup($vm_ip, $if_pingable);
+    } elsif ($check_vm_state) {
```
waynechen55, line 50: I am not very sure what this "elsif ($check_vm_state)" is used for ? It seems that it can be removed ?

lib/main_common.pm
nanzhg: For now, a lot of test modules are included in console tests. To save time and get rid of unnecessary test modules, we'll only keep integration service and open-vm-tools tests with textmode test suite in future.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,209 @@
...
+    record_info('Networking Tests');
+    check_vm_power_state($vm_id, $vm_ip, 'power.on', 1, 0, $VM_POWERED_OFF, $VM_POWERED_ON);
+
+    my $vswitch_name = esxi_vm_network_binding($vm_id);
+    record_info('VSwitch Name', $vswitch_name);
+
+    login_vm_console($vm_name);
+
+    assert_script_run('ping -4 -c3 openqa.suse.de');
+    assert_script_run('ping -4 -c3 www.suse.com');
```
waynechen55, line 147: What about "ping -I $vm_ip -4 -c3 xxxx" ? So your ping bind to your preferred ipv4 address ?

nanzhg: It's a good suggestion. Agree on this. Updated.

tests/virt_autotest/esxi_open_vm_tools.pm
waynechen55: It is more like get_random_time ?

nanzhg: This function invokes 'date' command with '-s' option, it's actually for setting a time. So I gave the prefix 'set_' for the name.
-s, --set=STRING           set time described by STRING

waynechen55: That's OK.

pdostal:
Hello,

after talking with @nanzhg, I would like to join this effort and help with the review. As this module uses `svirt` instead of `qemu + ssh` I suggest to continue here and deprecate `tests/virtualization/universal/open_vm_tools.pm`.

There are few suggestions you can implement in this new module:
 * Check the `vmtoolsd` systemd service so it starts.
 * Check the `vmtoolsd -v` command so it starts.
 * Check the `vmware-toolbox-cmd logging` command.

And please add a note to `tests/virtualization/universal/open_vm_tools.pm` about this new module.

Thank you

nanzhg:
@pdostal, Thanks for the review and comments. Will do related updates as we discussed. Thank you.

nanzhg:
@pdostal, I've merged the checks you suggested into this new module, and also added a note to tests/virtualization/universal/open_vm_tools.pm. Please help for review. Thanks
Verification run: http://10.67.129.66/tests/339

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+    assert_script_run("/usr/bin/vmware-checkvm | grep 'good'");
+    systemctl('status vmtoolsd');
+    systemctl('status vgauthd');
+    systemctl('restart vmtoolsd');
+    assert_script_run("systemctl status vmtoolsd | grep 'Started open-vm-tools'");
+    assert_script_run("/usr/bin/vmtoolsd -v | grep 'VMware Tools daemon, version'");
+    assert_script_run('vmware-toolbox-cmd logging level set vmtoolsd debug');
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = debug'");
+    assert_script_run('vmware-toolbox-cmd logging level set vmtoolsd message');
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = message'");
```
alice-suse, line 46: How about moving above block into some functions, similar with below sub-tests like do_xx_tests?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+
+    do_power_management_tests($vm_id, $vm_ip);
+    do_networking_tests($vm_id, $vm_ip);
+    do_clock_sync_tests($vm_name, $vm_id, $vm_ip);
+}
+
+sub check_vm_power_state {
+    my ($vm_id, $vm_ip, $power_action, $if_pingable, $check_vm_state, $if_vm_state, $expect_vm_state) = @_;
+
+    if (esxi_vm_power_getstate($vm_id) =~ m/$if_vm_state/) {
```
alice-suse, line 56: What does these parameters $if_pingable, $check_vm_state, $if_vm_state, $expect_vm_state stand for? Can you help give some explanation as comment?

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+    my $interval    = shift // 5;    # wait secs between each ping
+    my $cmd;
+
+    if ($if_pingable) {
+        $cmd = "ping -c$count -i$interval $vm_ip";
+    } else {
+        $cmd = "! ping -c$count -i$interval $vm_ip";
+    }
+
+    while ($times && console('svirt')->run_cmd($cmd)) {
```
alice-suse, line 79: The function name is wait_for_vm_bootup, I guess you want to wait until vm bootup, that is until ping finally works? When $if_pingable is 0, the cmd is statically "! ping", how can you tell whether vm is up or not? BTW this function can not differ timeout and cmd works, which I guess it should.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+    $epoch =~ s/,$//;
+    ($sec, $min, $hour, $mday, $mon, $year) = split /,/, $epoch;
+    $year -= 1900;
+    $mon  -= 1;
+    $epoch = timelocal($sec, $min, $hour, $mday, $mon, $year);
+
+    return $epoch;
+}
+
+sub do_power_management_tests {
```
alice-suse, line 121: This function is much clearer. Nice job!

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = debug'");
+    assert_script_run('vmware-toolbox-cmd logging level set vmtoolsd message');
+    assert_script_run("vmware-toolbox-cmd logging level get vmtoolsd | grep 'vmtoolsd.level = message'");
+
+    do_power_management_tests($vm_id, $vm_ip);
+    do_networking_tests($vm_id, $vm_ip);
+    do_clock_sync_tests($vm_name, $vm_id, $vm_ip);
+}
+
+sub check_vm_power_state {
```
alice-suse, line 53: I read it twice, since this function is not to check_xx, but to ensure_xx, right? If yes, suggest to make the function name more accurate.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,217 @@
...
+
+    # Guest time will not be synced up with the hypervisor after time synchronization disabled
+    $g_datetime = script_output("date -u +'\%F \%T'");
+    $h_datetime = get_host_timestamp();
+    $g_timesec  = get_epoch_time($g_datetime);
+    $h_timesec  = get_epoch_time($h_datetime);
+    $diff_secs  = int($g_timesec - $h_timesec);
+
+    record_info('Diff secs', $diff_secs);
+    if (abs($diff_secs) > 300) {
```
alice-suse, line 210: The guest time is set randomly, although chance is low, but in theory it is possible to be within 300 seconds difference. Suggest to set guest time with given time difference with host time, eg host time -1 day.

Besides, in this do_clock_sync_tests, there are obvious duplicated code lines which can be saved by small function, eg compare time difference code block. Can you improve it?

tests/virt_autotest/esxi_open_vm_tools.pm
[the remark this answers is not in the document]
alice-suse: @nanzhg I checked the job, I do not get quite relevant log for open-vm-tools test. Actually the consoletest is a quite generic base class, I am afraid it does not have open-vm-tools needed specific logs. Can you point out which files are the debug logs for it when test fail?

alice-suse (changes requested):
Nan, thanks a lot for the great rework. It is quite close to be merged. I finish my second round full review, and some minor comments are given. I will vote for merge after them fixed.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Agree. These commands looks more like some checks, so might be better call it do_sanity_checks().

tests/virt_autotest/esxi_open_vm_tools.pm
pdostal: This subroutine does just too much. I suggest to split it into:
1) Check the state
2) Change the state

And I also suggest to not `ping` inside this subroutine - that should be done elsewhere.

tests/virt_autotest/esxi_open_vm_tools.pm
pdostal: Also you checking the time difference twice - I suggest to create a subroutine for this.

pdostal (commented):
I am very happy with this VMware specific test. It's on good path! Thank you @nanzhg :rocket:

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, already added empty lines.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Updated to set guest time with the last day of host time, and creating a subroutine "init_guest_time()" for both the case of  timesync enable and disable. And created another subroutine "get_diff_seconds()" for checking the time difference code block.

lib/virt_autotest/esxi_utils.pm
pdostal: Why is the routine called like that? power cycle means turning it off and on again. Just call it `esxi_vm_power`.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Split into two subroutines: take_vm_power_action() & check_vm_power_state()

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Related description for the parameters has been added.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: This function provides two options that depends on what the power action did. For example, to test the VM shutdown,  we need to give 0 for $if_pingable, as we expected the ping command failed to exit the while loop. And also this ping command can be used for waiting VM bootup or shutdown, it will be run at most 90 seconds by default till it's timeout or break. After this period of time, invoking esxi_vm_power_getstate() to check if it's the expected VM power state. I know there are many ways to check VM is up or not, but just want to choose a simple way in this case because most of tests are based on running ssh command. So we could run those checks once the VM network bringing up or down. Maybe we can extend the check points in future.

nanzhg:
Verification run passed: http://10.67.129.66/tests/351

tests/virt_autotest/esxi_open_vm_tools.pm
alice-suse: I see, then what this function does actually differs a little from its function name -- wait_for_vm_bootup, right? It is more like wait_for_vm_bootup_or_powerdown?

tests/virt_autotest/esxi_open_vm_tools.pm
alice-suse: Thanks for the rework. It well matches the suggestion comments . 

One more minor suggestion, with 1 day/3600 seconds difference, shall the diff check here be closer to 3600 seconds, so as to ensure that the setting and syncing disablement actually works?

tests/virt_autotest/esxi_open_vm_tools.pm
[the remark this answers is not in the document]
alice-suse: @nanzhg  I do not see any further reply here. Any idea?

alice-suse (changes requested):
Most comments are resolved. Nice job!
I closed those resolved discussions. Now only 3 minor ones open, and need some attention.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Here the guest time is set one day earlier than host, the purpose of this test is to ensure the time will not be synced with host, but it's not able to avoid any time drift from guest. As we know, some intermediate operations(e.g. getting timestamp from host&guest) will cost a few seconds. I think as long as we observe the time is out of sync, that's enough.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, can we name it wait_for_vm_network? Maybe it's better.

tests/virt_autotest/esxi_open_vm_tools.pm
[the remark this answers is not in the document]
nanzhg: @alice-suse Sorry, I missed this thread. The following are the specific logs for open-vm-tools, do you think we need to copy them from guest if post_fail_hook triggered?

/var/log/vmware-network.log
/var/log/vmware-vgauthsvc.log.0
/var/log/vmware-vmsvc-root.log
/var/log/vmware-vmtoolsd-root.log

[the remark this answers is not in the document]
nanzhg: Another option is to use the vm-support command-line utility produces a gzipped tarball locally on the guest, which will collect diagnostic information for analysis. Then uploading this tarball to its openqa job.

E.g. vm-2021-09-26.10900.tar.gz
tmp/vm-support.A7bgZV/tmp/vm-support-version.txt
tmp/vm-support.A7bgZV/tmp/df.txt
tmp/vm-support.A7bgZV/tmp/ifconfig.txt
tmp/vm-support.A7bgZV/tmp/mount.txt
tmp/vm-support.A7bgZV/tmp/dmesg.txt
tmp/vm-support.A7bgZV/tmp/ulimit-a.txt
tmp/vm-support.A7bgZV/tmp/uptime.txt
tmp/vm-support.A7bgZV/tmp/date.txt
tmp/vm-support.A7bgZV/tmp/umask.txt
tmp/vm-support.A7bgZV/tmp/ps-auwwx.txt
tmp/vm-support.A7bgZV/tmp/lspci1.txt
tmp/vm-support.A7bgZV/tmp/lspci2.txt
tmp/vm-support.A7bgZV/tmp/lspci3.txt
tmp/vm-support.A7bgZV/tmp/modules.txt
tmp/vm-support.A7bgZV/tmp/uname.txt
tmp/vm-support.A7bgZV/tmp/issue.txt
tmp/vm-support.A7bgZV/tmp/rpm-qa.txt
tmp/vm-support.A7bgZV/tmp/netstat-lan.txt
tmp/vm-support.A7bgZV/tmp/route.txt
tmp/vm-support.A7bgZV/tmp/free.txt

tests/virt_autotest/esxi_open_vm_tools.pm
alice-suse: I do not quite get what the function does just from this name without knowing your previous explanation. Maybe you can also add some comment to tell what the function does?

tests/virt_autotest/esxi_open_vm_tools.pm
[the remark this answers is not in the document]
alice-suse: Seems these two options actually provides different debug logs. IMHO, what debug logs should be uploaded depends on what logs are needed to find root cause for failure in these open-vm-tools test. You should know this better than me. Sorry, I do not have any experience to share.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: OK, let's take an example. Refer to the verification run - http://10.67.129.66/tests/351#step/esxi_open_vm_tools/224

After timesync disabled and reset the guest time, we can see the deviation should be within 10s. I think it is acceptable.
Host time: 2021-09-26 07:21:50
Guest time: 2021-09-25 07:21:40

tests/virt_autotest/esxi_open_vm_tools.pm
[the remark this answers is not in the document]
nanzhg: OK, will upload both logs once the subroutine post_fail_hook triggered.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Got it. Updated.

tests/virt_autotest/esxi_open_vm_tools.pm
nanzhg: Added comments in this function.

nanzhg:
@alice-suse Thanks for review again. I've committed my updates again. Please help to merge if no any further comments from others.

tests/virt_autotest/esxi_open_vm_tools.pm
```diff
@@ -0,0 +1,293 @@
...
+    login_vm_console();
+
+    # Set guest time with a given time which different from host time
+    $g_init_time = init_guest_time();
+    record_info('Set guest time after time sync disabled', $g_init_time);
+
+    # Guest time will not be synced up with host after timesync service disabled
+    $diff_secs = get_diff_seconds();
+
+    if ($diff_secs > 86420) {
```
alice-suse, line 270: @nanzhg This diff_secs should be within a range, sth like 86400 < diff_secs < 86420?

nanzhg: Corrected. Thanks for pointing out. Will update the verification run once it's done.

alice-suse (approved):

alice-suse:
Wait for final validation run before merge.

nanzhg:
Verification run for the final commit got passed: http://10.67.129.66/tests/356

PR #13039, opened 2021-08-06
Fix build_and_run_image for non-suse hosts

b10n1k:
The `build_and_run_image` fails due to the Dockerfile context which
requires suse repos available but the base images are not supported
outside of the suse ecosystem.

The solution in this commit restores the Dockerfile that used to use
python image and no package installation was necessary. Use this when
host is not suse.

- Related ticket: https://progress.opensuse.org/issues/95786
- Verification run: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=b10n1k%2Fos-autoinst-distri-opensuse%2313039
https://openqa.opensuse.org/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313039&distri=opensuse&version=Tumbleweed

[opensuse](https://openqa.opensuse.org/tests/overview?distri=opensuse&build=b10n1k%2Fos-autoinst-distri-opensuse%2313039&version=Tumbleweed)


grisu48:
Remind me again, the reason from switching from the python app to apache2 was to reduce the build time only, right?

data/containers/app.py
```diff
@@ -0,0 +1,28 @@
...
+
+app = Flask(__name__)
+message = ""
+
+@app.route("/")
+def index():
+    return render_template("index.html", message = message)
+
+if __name__ == "__main__":
+    ctx = ssl.create_default_context()
```
grisu48, line 12: I don't see where the context `ctx` is actually used. Is this still required?

lib/containers/container_images.pm
```diff
@@ -57,6 +57,11 @@ sub build_and_run_image {
     record_info('Downloading', "Dockerfile: containers/$dockerfile\nHTML: containers/index.html");
     assert_script_run "mkdir -p $dir/BuildTest";
     assert_script_run "curl -f -v " . data_url("containers/$dockerfile") . " > $dir/BuildTest/Dockerfile";
+    if ($dockerfile eq 'Dockerfile.python3') {
```
grisu48, line 60: For now ok, but in the future we should move away from hardcoded filenames and bundle the test containers in their own directory or tarball and set the Dockerfile as variable.

This is beyond the scope of this PR though.

grisu48 (commented):
Nothing big, some small comments.

b10n1k:
> Remind me again, the reason from switching from the python app to apache2 was to reduce the build time only, right?

this is what i remember too. @pdostal also wanted to unify the Dockerfile IIRC

data/containers/app.py
b10n1k: `ssl.create_default_context()` will create an instance with secure default settings which are overrode below, to avoid ssl warnings etc..

lib/containers/container_images.pm
b10n1k: > For now ok, but in the future we should move away from hardcoded filenames and bundle the test containers in their own directory or tarball and set the Dockerfile as variable.

i agree. hardcoded strings do not look good. but it was straight forward in this case. it is interesting to me how this can improved. 

> 
> This is beyond the scope of this PR though.

thank you

data/containers/app.py
b10n1k: ready for review

pdostal:
> Remind me again, the reason from switching from the python app to apache2 was to reduce the build time only, right?

No. The reason was to reduce the number of dependencies.
My opinion is tho that I did not achieve that as we still depend on SCC :D

data/containers/app.py
```diff
@@ -0,0 +1,24 @@
+from flask import Flask, render_template
```
grisu48, line 1: Since not all people know Flask, we might want to add a small explanatory header about the function, something like

```suggestion
#!/usr/bin/env python3
##
## Simple Flask test program (see https://flask.palletsprojects.com/en/2.0.x/quickstart/)
## Serve 'index.html' from the `templates` directory
##
## Usage: python3 app.py
## Bind to port 80 or to the PORT environment variable, if set

from flask import Flask, render_template
```

data/containers/app.py
```diff
@@ -0,0 +1,24 @@
...
+import os, sys, ssl
+
+app = Flask(__name__)
+message = ""
+
+@app.route("/")
+def index():
+    return render_template("index.html", message = message)
```
grisu48, line 9: IMHO the [index.html](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/containers/index.html) file does not use the passed message argument at all but is a static file:

    <html>
    The test shall pass
    </html>

So, the Balrog always wins :slightly_smiling_face: 

As a suggestion: either lines 12-23 are not needed or that the html template should be edited.

grisu48 (changes requested):
We might have a small Balrog problem in the `app.py` :wink:

data/containers/app.py
b10n1k: ready for review

data/containers/app.py
```diff
@@ -0,0 +1,31 @@
...
+
+app = Flask(__name__)
+message = ""
+
+@app.route("/")
+def index():
+    return render_template("index.html")
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1:
```
grisu48, line 19: if `message` is not needed in line 16, the lines 19-30 are useless, right? See my previous comment.

Sorry for the nitpicking :slightly_smiling_face:

b10n1k: nope. this is irrelevant. 
`ENTRYPOINT ["python", "./app.py", "www.google.com"]` in dockerfile should explain it for you. which is used to `curl` later.

grisu48: Yes, and the outcome of this `curl` call is then stored in the `message` variable, which is never used.

You could replace `www.google.com` in the `ENTRYPOINT` with some bogus domain and the container application would still run successfully, e.g.

    ENTRYPOINT ["python", "./app.py", "http://www.thisdomaindoesnot.exist/"]

Here the container would behave just the same as with `www.google.com`. In the current state, this argument is irrelevant, and that makes me wonder why it is even there in the first place.

pdostal: We already have `containers/index.html` so please just read it and return it over HTTP. No need for any parameter here.

b10n1k: i removed a lot of the confusion. mark as `this change is ready for review`

b10n1k: https://openqa.opensuse.org/tests/1872711#

pdostal (approved):
This is nice, thank you @b10n1k :rocket:

grisu48 (approved):

PR #12947, opened 2021-07-21
Handle specific test checks on SLES12-SP5-JeOS

b10n1k:
- `Server Applications Module` is not available on SLES12-SP5-JeOS
- `Select System Local` comes before keyboard selection

- Related ticket: https://progress.opensuse.org/issues/95410
- Needles: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1548
- Verification run: 
https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2312947&distri=sle&version=12-SP5

https://openqa.suse.de/tests/overview?version=15-SP3&build=b10n1k%2Fos-autoinst-distri-opensuse%2312947&distri=sle

PR #13042, opened 2021-08-06
Fix curl_https defect from #12947

b10n1k:
Revert a change on glibc_locale which breaks curl_https in o3.
Change should have been removed on #12947

- Related ticket: https://progress.opensuse.org/issues/95410


b10n1k:
dummy commit.

PR #13043, opened 2021-08-06
[WIP] Use libyui-rest-api to register during installation

JRivrain:
Related ticket https://progress.opensuse.org/issues/95476

Job Groups: YaST
Screen: Registration
Test Cases:
1 - Provide only registration code;
3 - Skip registration for Full medium.
Ticket with Test Cases: https://progress.opensuse.org/issues/94895
Task:
 Implement test modules for both test cases using Page Object Model and LibyuiClient; Split existing scc_registration module to two parts: system registration and module selection, leaving the part for module selection as is (The rewrite of module selection will be covered in another ticket);
        As the scope is only YaST Job Group for now, we cannot simply remove registration part from scc_registration, as it is used in a lot of places. So, as a possible solution, the part for module selection may be extracted to separate function and moved to registration.pm library file then used in scc_registration and in new module (something like select_extensions_and_modules), which will be used altogether with the test module for registration.
    Update schedule files with the new modules.

VRs:
[ext4_yast@64bit](http://waaa-amazing.suse.cz/tests/15630)

Other VRs to see if the changes in library work as expected:
[skip_registration+workaround_modules@64bit](http://waaa-amazing.suse.cz/tests/15629)
[gnome+proxy_SCC+allmodules@64bit](http://waaa-amazing.suse.cz/tests/15577)
[wsm+dev_tools+textmode@64bit](http://waaa-amazing.suse.cz/tests/15579)

Update to schedule files in Yast group will be done in a separate PR for clarity.

### DIFF
--- a/lib/Installation/Registration/RegistrationController.pm
+++ b/lib/Installation/Registration/RegistrationController.pm
@@ -1,0 +1,59 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved. This file is offered as-is,
+# without any warranty.
+
+# Summary: The class introduces business actions for Registration dialog
+# when the system is already registered.
+#
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::Registration::RegistrationController;
+use strict;
+use warnings;
+use Installation::Registration::RegistrationPage;
+use Installation::Warnings::ConfirmationWarning;
+use YuiRestClient;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {}, $class;
+    return $self->init($args);
+}
+
+sub init {
+    my ($self, $args) = @_;
+    $self->{RegistrationPage}    = Installation::Registration::RegistrationPage->new({app => YuiRestClient::get_app()});
+    $self->{UseUpdateReposPopup} = Installation::Warnings::ConfirmationWarning->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_registration_page {
+    my ($self) = @_;
+    die "Registration page is not displayed" unless $self->{RegistrationPage}->is_shown();
+    return $self->{RegistrationPage};
+}
+
+sub get_update_repository_popup {
+    my ($self, $expected_popup) = @_;
+    # Check that that a "yesno" popup is shown and that it contains expected text.
+    $self->{UseUpdateReposPopup}->check_text($expected_popup) if $self->{UseUpdateReposPopup}->is_shown;
+    return $self->{UseUpdateReposPopup};
+}
+
+sub register_product_with_regcode {
+    my ($self, $reg_code) = @_;
+    $self->get_registration_page->enter_reg_code($reg_code);
+    $self->get_registration_page->press_next();
+}
+
+sub enable_update_repositories {
+    my ($self, $expected_popup) = @_;
+    $self->get_update_repository_popup($expected_popup)->press_yes();
+}
+
+1;
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2020 SUSE LLC
+# Copyright (C) 2015-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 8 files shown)


lib/registration.pm
```diff
@@ -565,25 +567,19 @@ sub process_scc_register_addons {
...
-}
-
 sub show_development_versions {
     assert_screen('scc-beta-filter-checkbox');
     send_key('alt-i');
     assert_screen('scc-beta-filter-unchecked');
 }
 
 sub fill_in_registration_data {
-    my ($addon, $uc_addon);
```
JRivrain, line 584: dead code

lib/registration.pm
```diff
@@ -565,25 +567,19 @@ sub process_scc_register_addons {
     }
 }
 
-sub process_scc_register {
```
JRivrain, line 568: Moved at the end of process_modules

lib/registration.pm
```diff
@@ -668,38 +664,6 @@ sub fill_in_registration_data {
             }
         }
     }
-
```
JRivrain, line 671: Moved in "process_modules"

tests/installation/registration/registration_with_regcode.pm
```diff
@@ -0,0 +1,26 @@
...
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi qw(check_var get_var send_key save_screenshot);
+use registration qw(handle_scc_popups process_modules skip_registration);
+
+sub run {
+    my $reg_code = get_var('SCC_REGCODE');
+    $testapi::distri->get_registration()->register_product_with_regcode($reg_code);
+    handle_scc_popups();
```
JRivrain, line 23: Temporary solution. With theis we keep the old way of deciding whether we we use or not the online repositories, and it deals with the intermediate popups. Or do we want to deal with these pop-ups with libyui ? with strict expectations, we should only have to deal with this piece below as the rest is for SLEl 12:

            while ($counter--) {
            die 'Registration repeated too much. Check if SCC is down.' if ($counter eq 1);
            assert_screen(\@tags, timeout => 360);
            if (match_has_tag('import-untrusted-gpg-key')) {
                handle_untrusted_gpg_key;
                next;
            }
                # sometimes nvidia driver repos are unreliable
                send_key 'alt-y';
                record_soft_failure 'bsc#1144831';
                next;
            }
            elsif (match_has_tag('contacting-registration-server')) {
                # sometimes SCC just takes its time - just continue looking after a while
                sleep 5;
                next;
            }
            elsif (match_has_tag('refreshing-repository')) {
                # it takes some time to refresh repos
                sleep 5;
                next;
            }

OleksandrOrlov: We should definitely get rid of approach with `match_has_tag` in our new test modules, as it makes test unpredictable. If we would do automatic installation but not a test, it may be useful to make the program to be very smart and decide what to do with some particular popup, which appear. But I guess we discussed that in a team and agreed that test should be dumb and perform expected steps. 

If some popup should be handled in test suite, it should not be some random thing, it is a separate test and probably we need to create a separate test module for this case.

So, for new tests we should define if some popups should be handled and handle them in separate test module. For `scc_registration` keep it as is for now, as in your task other job groups are out of scope and we cannot say right now if some popups need to be handled there.

JRivrain: I agree, but what to do with soft-failures ?

JRivrain: eg "# sometimes nvidia driver repos are unreliable"

OleksandrOrlov: Do the issues happen? In case it is really happen and it is something sporadic, leave the condition for this particular issue for now and create a ticket to investigate how can we improve our test module on that.

But I guess there should be no issues in our job group. At least I do not remember them.

JRivrain: For example in functional group, they are not even using yaml scheduling (eg here https://openqa.suse.de/tests/6599569#settings). I'd rather not use libyui in that module as we'd need to find every single test suite that uses it, create tons of yaml schedules, or mess with main.pm...

lib/Installation/Registration/RegistrationController.pm
```diff
@@ -0,0 +1,47 @@
...
+sub get_registration_page {
+    my ($self) = @_;
+    die "Registration page is not displayed" unless $self->{RegistrationPage}->is_shown();
+    return $self->{RegistrationPage};
+}
+
+sub register_product_with_regcode {
+    my ($self, $reg_code) = @_;
+    $self->get_registration_page->enter_reg_code($reg_code);
+    save_screenshot();
```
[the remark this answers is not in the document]
JRivrain: I thought we should use only methods that describe business logic in controllers ?

lib/registration.pm
JRivrain: I find this more clear like this, "process_scc_register" is confusing : it handles the modules, so it makes sense to put it in process_modules. otherwise renaming it would make two different functions that handle modules, so would be equally confusing.

PR #13050, opened 2021-08-10
Unify publiccloud variables

grisu48:
In an attempt to get rid of the old QAM variables, this PR introduced a
new unified `PUBLIC_CLOUD_` variable to skip usage of the maintenance
updates.

The old `QAM_PUBLICCLOUD_SKIP_DOWNLOAD` variable will be removed once we
are convinced that it is not used anymore.

- Verification run: http://duck-norris.qam.suse.de/tests/6808 and http://duck-norris.qam.suse.de/t6814


cfconrad (approved):
LGTM!

PR #13054, opened 2021-08-11
 Unset `_EXPECT_EFI_MOK_MANAGER` before selecting console

mloviska:
Failure as [during reboot](https://openqa.suse.de/tests/6790389#step/verify_efi_mok/120)
prevents accessing another console in the post fail hook subroutine.
Therefore the test case has to unset expected MOK Manager variable as
first action.


mloviska:
VR: http://kepler.suse.cz/tests/6425#

PR #13055, opened 2021-08-11
Adjust validate_btrfs module to use a single opensuse image

b10n1k:
`engines_and_tools_docker` tries to fetch an image from *registry.suse.com*
which doesnt exist at the moment. However when the module runs against sle15sp4
(latest jobgroup) it should use the images from the internal repo.

The solution proposed by @jlausuch is to pick up some openSUSE image as the test
    is always running against the host-based tests to test the btrfs storage feature
    of docker.

- Related ticket: https://progress.opensuse.org/issues/96383
- Verification run: TBD - no totest sp4 yet

### DIFF
--- a/tests/containers/validate_btrfs.pm
+++ b/tests/containers/validate_btrfs.pm
@@ -101,10 +101,10 @@ sub run {
     my ($running_version, $sp, $host_distri) = get_os_release;
     install_docker_when_needed($host_distri);
     allow_selected_insecure_registries(runtime => 'docker');
-    my $docker    = containers::runtime->new(runtime => 'docker');
-    my $btrfs_dev = '/var/lib/docker';
-    my ($untested_images, $released_images) = get_suse_container_urls();
-    _sanity_test_btrfs($docker, $btrfs_dev, $released_images->[0]);
+    my $docker         = containers::runtime->new(runtime => 'docker');
+    my $btrfs_dev      = '/var/lib/docker';
+    my $images_to_test = 'registry.opensuse.org/opensuse/leap:15';
+    _sanity_test_btrfs($docker, $btrfs_dev, $images_to_test);
     _test_btrfs_thin_partitioning($docker, $btrfs_dev);
     _test_btrfs_device_mgmt($docker, $btrfs_dev);
     _test_btrfs_balancing($btrfs_dev);


tests/containers/validate_btrfs.pm
```diff
@@ -104,7 +104,9 @@ sub run {
     my $docker    = containers::runtime->new(runtime => 'docker');
     my $btrfs_dev = '/var/lib/docker';
     my ($untested_images, $released_images) = get_suse_container_urls();
-    _sanity_test_btrfs($docker, $btrfs_dev, $released_images->[0]);
+    my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
+
+    _sanity_test_btrfs($docker, $btrfs_dev, $images_to_test->[0]);
```
grisu48, line 109: If `$images_to_test` is an array, why only test the first URL?

b10n1k: When the module implemented, it was one image in the list. i keep that as it is. 
Now this is scheduled on host-based tests. which means that it doesnt have to run against various images. if it works for one it should be fine

pdostal (approved):

b10n1k:
> This is not to be executed against untested images, it's a host test.
> What you could do is actually get the list of images to test:
> 
> ```
>     my $versions   = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
>     for my $version (split(/,/, $versions)) {
>         my ($untested_images, $released_images) = get_suse_container_urls($version);
> ```
> 
> and use `untested_images` array only.

i think i was not clear in my description.
i modified the module and i use the same technique as we do for podman_image and docker_image.
the difference:
- before: it used always `released_images`
- after: the images are used based on CONTAINERS_UNTESTED_IMAGES . Such as host-based < 15sp4 can use released and host based == 15sp4 (via CONTAINERS_UNTESTED_IMAGES) can use totest

tests/containers/validate_btrfs.pm
jlausuch: Yes, we need to find a trade-off here... We already run `podman_image.pm` for a bunch of released images and it's time consuming. So, testing single image is fine for me, and it doesn't have to be a sle image, it could be some random image...

jlausuch:
I would approach this PR differently. Since this test will be executed always for container host validation (not for container image validation), since it's testing a host feature, instead of calling `get_suse_container_urls` I would just pick some openSUSE image or something. 

So, what I would do is simply replace this:
```
    my ($untested_images, $released_images) = get_suse_container_urls();
    _sanity_test_btrfs($docker, $btrfs_dev, $released_images->[0]);
```
by this:
```
    my $image = 'registry.opensuse.org/opensuse/leap';
    _sanity_test_btrfs($docker, $btrfs_dev, $image);
```

What do you think?

jlausuch:
LGTM. You have 2 commits, want to squash them? Also, since you updated the commit message, can you also update the title of the PR to the same? Thx!

b10n1k:
> LGTM. You have 2 commits, want to squash them? Also, since you updated the commit message, can you also update the title of the PR to the same? Thx!

Done. but i am waiting for a job that i can clone and provide a VR

pdostal:
Please rebase Yianni to get rid of unrelated Michael's commit.

b10n1k:
https://openqa.suse.de/tests/7084829

jlausuch (approved):

jlausuch (commented):
You still have to rebase or squash :)

b10n1k:
> You still have to rebase or squash :)

But i rebased last night. what do i miss?

jlausuch:
> > You still have to rebase or squash :)
> 
> But i rebased last night. what do i miss?

Your PR contains 2 commits, your actual change and 
[915478b918f3c19f7c92b6b365c61a091bcebb1e](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/915478b918f3c19f7c92b6b365c61a091bcebb1e)
which I don't know where it comes from, it's not even merged in the main repo. You need to squash them to avoid confusion in the main repo.
I normally do `git rebase -i HEAD~2` and `pick` the commit I want to keep and `squash` the commit I want to disappear.

jlausuch (approved):

PR #13058, opened 2021-08-11
lib/containers/common: Ignore warnings from buildah from

Vogtinator:
script_output using the serial terminal captures both stdout and stderr, but
stderr may contain some warnings which need to be ignored.

- Related ticket: https://progress.opensuse.org/issues/96753
- Verification run:
bulidah_docker: https://openqa.opensuse.org/t1873975
bulidah_podman: https://openqa.opensuse.org/t1873976


Vogtinator:
To get 20210810 out, this confirms that buildah isn't actually broken and the warning is harmless.

PR #12697, opened 2021-06-11
Allow root to login using password

rfan1:
Due to bsc#1173067, openssh now no longer allows RootLogin
using password auth on Tumbleweed, the latest Leap 16 will sync
with Tumbleweed as well.

So fix the logic here, I didn't test Leap and SLES for now, will check the code once sles15sp4 is out.

- Related ticket: https://progress.opensuse.org/issues/93832
                          https://progress.opensuse.org/issues/93835
- Needles: n/a
- Verification run: https://openqa.opensuse.org/tests/1787735 (389-ds server)
                            https://openqa.opensuse.org/tests/1787736 (389-ds client)
                            https://openqa.opensuse.org/tests/1787686 (swtpm-create-hdd)
                            https://openqa.opensuse.org/tests/1787687 (swtpm-verify)

PR #13060, opened 2021-08-12
Permit ssh root login and pubkey access for Tumbleweed

Julie-CAO:
sshd default configuration was changed since a TW snapshot(between 20210521 and 20210727). Root user login, password login and authentic keys login have not been allowed by default any more. This PR is to modify the sshd configuration as soon as the SLE is installed as ssh is the primary way we interact with the system.

- permit root login to host via sol console
- permit root login and authentic keys login to guest via autoyast
- minor improvement
  --use 'ip link' instead 'brctl'(no extra package has to be installed)
  --remove workaround for libvirtd.service as the issue is fixed

- Related ticket: https://progress.opensuse.org/issues/96374
- needles: [sol-console-wait-typing-ret](http://10.67.129.51/tests/2580#step/switch_to_ssh_and_install_hypervisor/2)
- Verification run: 
[TW kvm guest installation](http://10.67.129.51/tests/2566)
[TW XEN guest installation](http://10.67.129.51/tests/2580)
[SLE 12sp5 xen fv guest installation](https://openqa.nue.suse.com/tests/6828263)

@alice-suse @guoxuguang @waynechen55 @nanzhg

### DIFF
--- a/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
+++ b/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
@@ -174,5 +174,29 @@
       <username>root</username>
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- /etc/ssh/sshd_config/sshd_config is not default installed in TW -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config"
+if [ ! -f $sshd_config_file ]; then
+    mkdir -p `dirname $sshd_config_file`
+    echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
+else
+    keys="PermitRootLogin PubkeyAuthentication PasswordAuthentication"
+    for key in $keys; do
+        sed -i "/^[# ]*$key */{h;s/^[# ]*$key *.*\$/$key yes/};\${x;/^\$/{s//$key yes/;H};x}" $sshd_config_file
+    done
+fi
+[ -d /root/.ssh ] || mkdir -p /root/.ssh; chmod 700 /root/.ssh
+touch /root/.ssh/authorized_keys; chmod 600 /root/.ssh/authorized_keys
+echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
+]]>
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
 </profile>
 
--- a/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml
+++ b/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml
@@ -174,5 +174,30 @@
       <username>root</username>
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- /etc/ssh/sshd_config/sshd_config is not default installed in TW -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config"
+if [ ! -f $sshd_config_file ]; then
+    mkdir -p `dirname $sshd_config_file`
+    echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
+else
+    keys="PermitRootLogin PubkeyAuthentication PasswordAuthentication"
+    for key in $keys; do
+        sed -i "/^[# ]*$key */{h;s/^[# ]*$key *.*\$/$key yes/};\${x;/^\$/{s//$key yes/;H};x}" $sshd_config_file
+    done
+fi
+[ -d /root/.ssh ] || mkdir -p /root/.ssh; chmod 700 /root/.ssh
+touch /root/.ssh/authorized_keys; chmod 600 /root/.ssh/authorized_keys
+echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
+]]>
+
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
 </profile>
 
--- a/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
+++ b/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
@@ -174,5 +174,29 @@
       <username>root</username>
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- /etc/ssh/sshd_config/sshd_config is not default installed in TW -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config"
+if [ ! -f $sshd_config_file ]; then
+    mkdir -p `dirname $sshd_config_file`
+    echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
+else
+    keys="PermitRootLogin PubkeyAuthentication PasswordAuthentication"
+    for key in $keys; do
+        sed -i "/^[# ]*$key */{h;s/^[# ]*$key *.*\$/$key yes/};\${x;/^\$/{s//$key yes/;H};x}" $sshd_config_file
+    done
+fi
+[ -d /root/.ssh ] || mkdir -p /root/.ssh; chmod 700 /root/.ssh
+touch /root/.ssh/authorized_keys; chmod 600 /root/.ssh/authorized_keys
+echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
+]]>
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
 </profile>
 
--- a/lib/guest_installation_and_configuration_base.pm
+++ b/lib/guest_installation_and_configuration_base.pm
@@ -792,7 +792,7 @@ sub config_guest_network_bridge_device {
                 #NIC in openSUSE TW guest is unable to get the IP from its network configration file with 'wicked ifup' or 'ifup'
                 #Not sure if it is a bug yet. This is just a temporary solution.
                 my $_bridge_ipaddr = script_output("grep IPADDR $_bridge_device_config_file | cut -d \"'\" -f2");
-                script_retry("brctl addbr $_bridge_device && ip addr add $_bridge_ipaddr dev $_bridge_device && ip link set $_bridge_device up", retry => 3, die => 0);
+                script_retry("ip link add $_bridge_device type bridge && ip addr add $_bridge_ipaddr dev $_bridge_device && ip link set $_bridge_device up", retry => 3, die => 0);
             }
             else {
                 script_retry("wicked ifup $_bridge_device_config_file $_bridge_device", retry => 3, die => 0);
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -99,6 +99,7 @@ our @EXPORT = qw(
   assert_secureboot_status
   susefirewall2_to_firewalld
   permit_root_ssh
+  permit_root_ssh_in_sol
 );
 
 =head1 SYNOPSIS
@@ -2030,6 +2031,22 @@ sub permit_root_ssh {
         assert_script_run("echo 'PermitRootLogin yes' > /etc/ssh/sshd_config.d/root.conf");
         assert_script_run("systemctl restart sshd");
     }
+}
+
+=head2 permit_root_ssh_in_sol
+    permit_root_ssh_in_sol();
+
+for ipmi backend, PermitRootLogin has to be set in sol console
+however, assert_script_run and script_run is not stable in sole console
+enter_cmd or type_string are acceptable
+
+=cut
+
+sub permit_root_ssh_in_sol {
+    my $sshd_config_file = shift;
+
+    $sshd_config_file //= "/etc/ssh/sshd_config";
+    enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
 }
 
 1;
--- a/tests/virt_autotest/reboot_and_wait_up.pm
+++ b/tests/virt_autotest/reboot_and_wait_up.pm
@@ -44,7 +44,7 @@ sub reboot_and_wait_up {
         switch_from_ssh_to_sol_console(reset_console_flag => 'off');
         #on some SUTs, fg. a HP machine in Beijing lab, screen of sol console is cleared at the second reboot.
         #and the 'return' key must be typed then 'login' prompt appears in sol console
-        send_key 'ret' unless check_screen('text-logged-in-root');
+        send_key 'ret' if check_screen('sol-console-wait-typing-ret');
 
         my ($package_name, $file_name, $line_num) = caller;
         diag("The package $package_name defined from file $file_name called me at line $line_num");
@@ -60,14 +60,16 @@ sub reboot_and_wait_up {
             ipmi_backend_utils::ipmitool 'chassis power reset';
         }
         else {
-            #login
-            #The timeout can't be too small since autoyast installation
-            assert_screen "text-login", 600;
-            enter_cmd "root";
-            assert_screen "password-prompt";
-            type_password;
-            send_key('ret');
-            assert_screen "text-logged-in-root";
+            #login is required when sol console is used for the first time
+            unless (check_screen('text-logged-in-root')) {
+                #The timeout can't be too small since autoyast installation
+                assert_screen "text-login", 600;
+                enter_cmd "root";
+                assert_screen "password-prompt";
+                type_password;
+                send_key('ret');
+                assert_screen "text-logged-in-root";
+            }
 
             #type reboot
             enter_cmd("reboot");
--- a/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
+++ b/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
@@ -20,18 +20,25 @@ use base 'virt_autotest_base';
 use testapi;
 use ipmi_backend_utils;
 use Utils::Backends qw(use_ssh_serial_console is_remote_backend);
-use utils qw(zypper_call systemctl);
+use utils qw(zypper_call systemctl permit_root_ssh_in_sol);
 use virt_autotest::utils qw(is_kvm_host is_xen_host);
 
 sub run {
 
-    #skip TW host installation and directly login if you'd like to run test on an SUT with TW installed
-    if (get_var('IPMI_DO_NOT_RESTART_HOST')) {
-        select_console 'sol', await_console => 0;
-        send_key 'ret';
+    #enable ssh root access for openSUSE TW in sol console
+    #root access is not permitted by default in TW, see bsc#1173067#c2
+    select_console 'sol', await_console => 1;
+    send_key 'ret' if check_screen('sol-console-wait-typing-ret');
+    if (check_screen('text-login')) {
+        enter_cmd "root";
+        assert_screen "password-prompt";
+        type_password;
+        send_key('ret');
     }
-
-    use_ssh_serial_console unless check_screen('text-logged-in-root');
+    assert_screen "text-logged-in-root";
+    #skip TW host installation and directly login if you'd like to run test on an SUT with TW installed
+    permit_root_ssh_in_sol unless get_var('IPMI_DO_NOT_RESTART_HOST');
+    select_console('root-ssh');
 
     return if get_var('IPMI_DO_NOT_RESTART_HOST');
 
@@ -42,13 +49,6 @@ sub run {
     set_serial_console_on_vh('', '', $hypervisor);
 
     virt_autotest::utils::install_default_packages();
-    zypper_call('in bridge-utils');
-
-    #check status of libvirtd
-    systemctl 'status libvirtd', ignore_failure => 1;
-    save_screenshot;
-    systemctl 'restart libvirtd', ignore_failure => 1;
-    systemctl 'status libvirtd',  ignore_failure => 1;
     save_screenshot;
 }
 
--- a/tests/virt_autotest/virt_utils.pm
+++ b/tests/virt_autotest/virt_utils.pm
@@ -41,9 +41,9 @@ sub enable_debug_logging {
     #log filter is set to store component logs with different levels.
     my $libvirtd_conf_file = "/etc/libvirt/libvirtd.conf";
     if (!script_run "ls $libvirtd_conf_file") {
-        script_run "sed -i '/log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $libvirtd_conf_file";
-        script_run "sed -i '/log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%};\${x;/^\$/{s%%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%;H};x}' $libvirtd_conf_file";
-        script_run "sed -i '/log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%};\${x;/^\$/{s%%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $libvirtd_conf_file";
         script_run "grep -e log_level -e log_outputs -e log_filters $libvirtd_conf_file";
     }
     save_screenshot;
@@ -51,7 +51,7 @@ sub enable_debug_logging {
     # enable journal log with prvious reboot
     my $journald_conf_file = "/etc/systemd/journald.conf";
     if (!script_run "ls $journald_conf_file") {
-        script_run "sed -i '/Storage *=/{h;s/^[# ]*Storage *=.*\$/Storage=persistent/};\${x;/^\$/{s//Storage=persistent/;H};x}' $journald_conf_file";
+        script_run "sed -i '/^[# ]*Storage *=/{h;s/^[# ]*Storage *=.*\$/Storage=persistent/};\${x;/^\$/{s//Storage=persistent/;H};x}' $journald_conf_file";
         script_run "grep Storage $journald_conf_file";
         script_run 'systemctl restart systemd-journald';
     }


data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
```diff
@@ -174,5 +174,17 @@
       <username>root</username>
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes\nAuthorizedKeysFile      .ssh/authorized_keys" > /etc/ssh/sshd_config
+[ -d /root/.ssh ] || mkdir /root/.ssh; echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
```
alice-suse, line 183: Can it be made more tolerable -- no matter how these settings be changed in future(on or off, partially or all), the script can set up the required configurations correctly?  Same as below.

Julie-CAO: sshd_config has been removed in TW. this shell script is the easiest way to create a new sshd_config over autoyast. We may have more cautious way to check if the file exist then blabla.., like 

if sshd_config exists
  sed /xxx/xxx/s if xxx exists
else
  echo ...

autoyast debugging is more complex than simple shell environment ---- have to run an entire autoyast installation each time and there is no detail debug information. I'll have a try, If it goes well, I will improve it. Otherwise, I'd better keep the simplest method to save effort :-).

lib/guest_installation_and_configuration_base.pm
```diff
@@ -792,7 +792,7 @@ sub config_guest_network_bridge_device {
                 #NIC in openSUSE TW guest is unable to get the IP from its network configration file with 'wicked ifup' or 'ifup'
                 #Not sure if it is a bug yet. This is just a temporary solution.
                 my $_bridge_ipaddr = script_output("grep IPADDR $_bridge_device_config_file | cut -d \"'\" -f2");
-                script_retry("brctl addbr $_bridge_device && ip addr add $_bridge_ipaddr dev $_bridge_device && ip link set $_bridge_device up", retry => 3, die => 0);
+                script_retry("ip link add $_bridge_device type bridge && ip addr add $_bridge_ipaddr dev $_bridge_device && ip link set $_bridge_device up", retry => 3, die => 0);
```
waynechen55, line 795: This is a change that based on the change that you made to opensuse ?  ip link is better than brctl in which way ?

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml
```diff
@@ -174,5 +174,17 @@
...
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes\nAuthorizedKeysFile      .ssh/authorized_keys" > /etc/ssh/sshd_config
+[ -d /root/.ssh ] || mkdir /root/.ssh; echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
+]]>
```
waynechen55, line 184: I think you'd better refer to this article:
https://developers.redhat.com/blog/2018/11/02/how-to-manually-copy-ssh-keys-rhel
It seems that it introduces the complete steps to setup passwordless ssh login, including permission issues

waynechen55: Also this one:
https://phoenixnap.com/kb/setup-passwordless-ssh

lib/guest_installation_and_configuration_base.pm
Julie-CAO: yes, for TW only, and may change after I have more thought about wicked in TW.

'ip link' is default installed in TW, no need to install an extra bridge-utils package to run 'brctl'.

tests/virt_autotest/reboot_and_wait_up.pm
```diff
@@ -44,7 +44,7 @@ sub reboot_and_wait_up {
         switch_from_ssh_to_sol_console(reset_console_flag => 'off');
         #on some SUTs, fg. a HP machine in Beijing lab, screen of sol console is cleared at the second reboot.
         #and the 'return' key must be typed then 'login' prompt appears in sol console
-        send_key 'ret' unless check_screen('text-logged-in-root');
+        send_key 'ret' if check_screen('sol-console-wait-typing-ret');
```
waynechen55, line 47: To be honest, I do not think this new needle name is intuitive. I think "sol-console-activated" is better based what you are   capturing. Additionally, "unless check_screen('text-logged-in-root')" already covers "if check_screen('sol-console-wait-typing-ret')", so why bother changing this part ?

tests/virt_autotest/reboot_and_wait_up.pm
```diff
@@ -60,14 +60,16 @@ sub reboot_and_wait_up {
...
+            #login is required when sol console is used for the first time
+            unless (check_screen('text-logged-in-root')) {
+                #The timeout can't be too small since autoyast installation
+                assert_screen "text-login", 600;
+                enter_cmd "root";
+                assert_screen "password-prompt";
+                type_password;
+                send_key('ret');
+                assert_screen "text-logged-in-root";
+            }
```
waynechen55, line 72: What about 
```
elsif (! check_screen('text-logged-in-root')) 
xxxxx
else 
yyyyy
```

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml
Julie-CAO: the script in autoyast runs inside guest, while 'ssh-copy-id' needs to be run in host. Unless you modify the unified guest installation base class to support this operation.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -25,13 +25,27 @@ use virt_autotest::utils qw(is_kvm_host is_xen_host);
...
+    send_key 'ret' if check_screen('sol-console-wait-typing-ret');
+    if (check_screen('text-login')) {
+        enter_cmd "root";
+        assert_screen "password-prompt";
+        type_password;
+        send_key('ret');
+    }
+    assert_screen "text-logged-in-root";
+    enter_cmd("echo \"PermitRootLogin yes\" > /etc/ssh/sshd_config");
+    enter_cmd("systemctl restart sshd");
```
waynechen55, line 40: What about a new function called 
enable_root_ssh_login ?
So any host needs it can just call it directly.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml
waynechen55: Ignore ssh-copy-id, just manual part

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: By the way, I think there is already functionality can help do this, you do not need a new needle:

```
    select_console 'sol', await_console => 1;
    send_key 'ret';
```
This code snippet can help do the same thing.

Julie-CAO: two needles here because I have met the similiar problem you had met where you opened a progress ticket to openqa tool team.

Note the screenshots below:
text logged in: http://10.67.129.51/tests/2573#step/switch_to_ssh_and_install_hypervisor/5
'sol-console-wait-typing-ret'(not type 'ret' yet): http://10.67.129.51/tests/2573#step/switch_to_ssh_and_install_hypervisor/2
sol-console-activated('ret' typed or no need to type 'ret' on some suts):  http://10.67.129.51/tests/2573#step/reboot_and_wait_up_normal/3

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
guoxuguang: if removed sshd_config file as expected in TW, would be better to add comments in opensuse_tumbleweed_XX.xml file to let us clear this changed.

tests/virt_autotest/reboot_and_wait_up.pm
Julie-CAO: No actition is needed for 'else', so there is no 'else' here.  Could you give an example if I did not get your meaning exactly?

rfan1:
There is already a function to permit root ssh:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12697/
https://progress.opensuse.org/issues/93829
Please double check :)

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: we have already a permit_root_ssh function in utils which is recently added and can be used in SLE, I thought a little bit and am afraid our 'sol' console is not so friendly to assert_script_run... 

sub permit_root_ssh {
    if (is_sle('<16') || is_leap('<16.0')) {
        my $results = script_run("grep 'PermitRootLogin yes' /etc/ssh/sshd_config");
        if (!$results) {
            assert_script_run("sed -i 's/^PermitRootLogin.*\$/PermitRootLogin yes/' /etc/ssh/sshd_config");
            assert_script_run("systemctl restart sshd");
        }
    }
    else {
        assert_script_run("echo 'PermitRootLogin yes' > /etc/ssh/sshd_config.d/root.conf");
        assert_script_run("systemctl restart sshd");
    }
}

Julie-CAO:
> There is already a function to permit root ssh:
> #12697
> https://progress.opensuse.org/issues/93829
> Please double check :)

thanks for reminding. I'll evaluate if sol console supports assert_script_run() well.

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: This is what I think:
```
sub switch_from_ssh_to_sol_console {
    my (%opts) = @_;

    #close root-ssh console
    prepare_system_shutdown;
    #switch to sol console
    set_var('SERIALDEV', '');
    $serialdev = 'ttyS1';
    bmwqemu::save_vars();
    if ($opts{'reset_console_flag'} eq "on") {
        reset_consoles;
    }
    select_console 'sol', await_console => 1; ##### You need to change to await console
    save_screenshot;
}
```
Then remove 
```
send_key 'ret' if check_screen('sol-console-wait-typing-ret');
```

or just
```
send_key 'ret'
```
Because "await_console" already makes sure you are in sol.

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: I do not know either. After thinking twice, I think your logic here is enough.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
waynechen55: Have you tried ?  If you are sure there is problem in sol, could you help develop a sol version of permit_root_ssh ? So it can also be reused.

Julie-CAO: I met problem with assert_script_run() in the past in sol console, I am unable to remember the detail, it is likely the return value is not always correct, so assertion sometimes fails but actually the command is executed fine. 'type_string' or 'enter_cmd' are more stable in sol console. I'll try again and figure out the best approach.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
waynechen55: @Julie-CAO suse's sshd is just openssh deamon. According to openssh sshd document:
https://man.openbsd.org/sshd.8

-f config_file
Specifies the name of the configuration file. The default is /etc/ssh/sshd_config. sshd refuses to start if there is no configuration file.

So if sshd config is not available, then there is no chance that sshd  can be started, let alone "whether root login is permitted". So I think the root cause might be sshd config is missing instead of root login is not allowed.

Julie-CAO: I checked sshd.service is running without errors when TW is installed.

root login is really not permitted, please refer to the bug in my comment: bsc#1173067, the progress ticket from other team https://progress.opensuse.org/issues/93829, and the PR(https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12697) which says 'not permit root login' will be adapted by SLE16 and Leap 16.

Julie-CAO:
> > There is already a function to permit root ssh:
> > #12697
> > https://progress.opensuse.org/issues/93829
> > Please double check :)
> 
> thanks for reminding. I'll evaluate if sol console supports assert_script_run() well.
Just tried, assert_script_run() malfunctions in sol console. 
http://10.67.129.51/tests/2583#step/switch_to_ssh_and_install_hypervisor/5
`[2021-08-16T16:14:36.823 CST] [info] ::: basetest::runtest: # Test died: command 'echo 'PermitRootLogin yes' > /etc/ssh/sshd_config.d/root.conf' timed out at /usr/lib/os-autoinst/testapi.pm line 959`

I am going to improve permit_root_ssh() to adapt sol console.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: Just tried, assert_script_run() does not work in sol console, please see my reply https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13060#issuecomment-899329620. I am going to improve the existing permit_root_ssh function to be workable on sol console.

tests/virt_autotest/reboot_and_wait_up.pm
Julie-CAO: SUT behaves differently with select_console 'sol' regardless await_console => 1, for example,  sut A(such as suts in OSD) changes to sol console and '#' prompt directly(like this https://openqa.nue.suse.com/tests/6598987#step/reboot_and_wait_up_normal/4); sut B(such as vh015/vh016 HP machines) stops at 'sol-console-wait-typing-ret'(http://10.67.129.51/tests/2573#step/switch_to_ssh_and_install_hypervisor/2),  after receives send_key 'ret' then '#' prompts.

We have to differentiate them because 'ret' is only required for sut B. sut A is prompting login, 'ret' will make login fail on sut A.
The meaning of check_screen('sol-console-wait-typing-ret')  is to distinguish between SUT A and SUT B

here is the example I just run(with 'await_console => 1' and without sending 'ret' on vh016):
http://10.67.129.51/tests/2584#step/switch_to_ssh_and_install_hypervisor/3

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
waynechen55: This bug mentions that something found in config file, but the config file now is missing from TW I think. So how can I know what default settings are ? I think there might be a product issue if default config file is not available for sshd.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
waynechen55: Are you going to add alternatives like type_string or similar for sol ? If that is case, I think this may deserve a new function or even a new module to store functions that can be reused in sol.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
Julie-CAO: @alice-suse do you know where the upstream is adopted in TW? where or how can I know the TW we tested is supposed to be? I did not find any document for TW, I referenced the docs for openSUSE Leap.

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: Why the original
```
send_key 'ret' unless check_screen('text-logged-in-root');
```
is replaced with
```
send_key 'ret' if check_screen('sol-console-wait-typing-ret');
```
Have you considered any consequence that may follow ? There is no chance to "ret' anymore unless "sol-console-wating-type-ret".

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -25,13 +25,27 @@ use virt_autotest::utils qw(is_kvm_host is_xen_host);
...
+
     #skip TW host installation and directly login if you'd like to run test on an SUT with TW installed
     if (get_var('IPMI_DO_NOT_RESTART_HOST')) {
         select_console 'sol', await_console => 0;
         send_key 'ret';
     }
 
-    use_ssh_serial_console unless check_screen('text-logged-in-root');
+    use_ssh_serial_console;
```
waynechen55, line 48: Can we just use api here, like select_console('root-ssh' or someting) ?  I do not know why we need to stick with use_ssh_serial_console

tests/virt_autotest/reboot_and_wait_up.pm
Julie-CAO: > Why the original
> 
> ```
> send_key 'ret' unless check_screen('text-logged-in-root');
> ```
> 
> is replaced with
> 
> ```
> send_key 'ret' if check_screen('sol-console-wait-typing-ret');
> ```
> 
'sol-console-wait-typing-ret' is a new introduced needle which can avoid the problem you met in arm. the selected area can be large enough to distinguish 'text-logged-in-root' and 'text-login'.

> Have you considered any consequence that may follow ? There is no chance to "ret' anymore unless 'sol-console-wait-typing-ret"

Yes, that is exactly what I want. no 'ret' for those suts which has already been in sol console, type 'ret' for thost suts which required 'ret' for command prompts.

waynechen55: OK if you know what you are going to do.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -25,13 +25,27 @@ use virt_autotest::utils qw(is_kvm_host is_xen_host);
...
+        send_key('ret');
+    }
+    assert_screen "text-logged-in-root";
+    enter_cmd("echo \"PermitRootLogin yes\" > /etc/ssh/sshd_config");
+    enter_cmd("systemctl restart sshd");
+
     #skip TW host installation and directly login if you'd like to run test on an SUT with TW installed
     if (get_var('IPMI_DO_NOT_RESTART_HOST')) {
         select_console 'sol', await_console => 0;
         send_key 'ret';
```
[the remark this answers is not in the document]
Julie-CAO: yes, I have changed.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
Julie-CAO: Talked with @rfan1 offline that the removal of sshd_config is the fact for TW and it is expected for SLE16 and Leap 16, the info is coming from developer, no Jira feature or document about it. Anyway, sshd is not our test target but a necessary procedure for virtualization testing, so I prefer to let it work rather than investigate whether it is reasonable.

waynechen55: It is acceptable at the moment I think.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: create a separate permit_root_ssh_in_sol() function in utils.pm. It can be used in our future SLE16 test. Verification tests are in progress.

Julie-CAO:
> There is already a function to permit root ssh:
> #12697
> https://progress.opensuse.org/issues/93829
> Please double check :)

A new permit_root_ssh_in_sol() has been created for ipmi backend in utils.pm.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
alice-suse: > @alice-suse do you know where the upstream is adopted in TW? where or how can I know the TW we tested is supposed to be? I did not find any document for TW, I referenced the docs for openSUSE Leap.

@Julie-CAO You can get official product changes(mean to have changes, rather than random) according to below quote from https://confluence.suse.com/pages/viewpage.action?spaceKey=qasleapac2&title=Factory-first+discussion+meeting+20210527. 
```
TW product change notice: there can be several ways to follow:
1) check in https://planet.opensuse.org/en/ the  posts with title "openSUSE Tumbleweed – Review of the week YYYY/WW", eg http://dominique.leuenberger.net/blog/2021/05/opensuse-tumbleweed-review-of-the-week-2021-21/
2) https://download.opensuse.org/tumbleweed/iso/ files with name Changes.[6 digit date].txt  
3) check "News portals" and "Mailing lists" in https://en.opensuse.org/openSUSE:Communication_channels
4) https://build.opensuse.org/project/show/openSUSE:Factory package changelogs

Note: confirmed with Dominique  Leuenberger, 1) and 2) can  give a rather complete insight of what happened and also a glimpse at what is coming.
```

lib/utils.pm
```diff
@@ -2032,4 +2033,19 @@ sub permit_root_ssh {
...
+
+for ipmi backend, PermitRootLogin has to be set in sol console
+however, assert_script_run and script_run is not stable in sole console
+enter_cmd or type_string are acceptable
+
+=cut
+
+sub permit_root_ssh_in_sol {
+
+    my $sshd_config_file = "/etc/ssh/sshd_config";
```
waynechen55, line 2047: Possible to make this part more resilient ? For example,
```
sub permit_root_ssh_in_sol {
       my $sshd_config_file = shift;

       $sshd_config_file //= "/etc/ssh/sshd_config";
}

And also check whether password login is permitted ?

Comment (Julie-CAO) [reply lib/utils.pm:2047]: > Possible to make this part more resilient ? For example,
> 
> ```
> sub permit_root_ssh_in_sol {
>        my $sshd_config_file = shift;
> 
>        $sshd_config_file //= "/etc/ssh/sshd_config";
> 
>        xxxxx
> }
> ```
Good suggestion.

> 
> And also check whether password login is permitted ?

root can login with password with this change. We'd better not make the function complex as the running commands on sol console is risky.

File: data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml (hunk 1 above)
Comment (Julie-CAO) [reply data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml]: > if removed sshd_config file as expected in TW, would be better to add comments in opensuse_tumbleweed_XX.xml file to let us clear this changed.

more comments added.

File: data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml (hunk 3 above)
Comment (Julie-CAO) [reply data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_hvm_guest_x86_64.xml:195]: add more steps in autoyast about the permission of the ssh files.

File: data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml (hunk 1 above)
Comment (Julie-CAO) [reply data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml]: > Can it be made more tolerable -- no matter how these settings be changed in future(on or off, partially or all), the script can set up the required configurations correctly? Same as below.

changed the shell script in autoyast in following cases:
case1: "#key=value" => "key=value"
case2: key does not exist => add the key on the end of sshd_config
case3: sshd_config does not exist => create and add all the keys

File: tests/virt_autotest/virt_utils.pm [hunk 10]
```diff
@@ -41,17 +41,17 @@ sub enable_debug_logging {
...
-        script_run "sed -i '/log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%};\${x;/^\$/{s%%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%;H};x}' $libvirtd_conf_file";
+        script_run "sed -i '/^[# ]*log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $libvirtd_conf_file";
         script_run "grep -e log_level -e log_outputs -e log_filters $libvirtd_conf_file";
```
waynechen55, line 47: Can your regex cover 
```
log_level=xxxxx
log_outputs=xxxxx
log_filters=xxxxx
`````
?

If not, what about
```
sed -i '/^.*log_level *=/xxxxx;
```
?

File: tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm (hunk 7 above)
Comment (Julie-CAO) [reply tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm]: yes. select_console('root-ssh') is enough here.

File: tests/virt_autotest/virt_utils.pm (hunk 10 above)
Comment (Julie-CAO) [reply tests/virt_autotest/virt_utils.pm:47]: Yes, it does. "^[# ]*log_level *=" matches "log_level=xxxxx".  '*' means zero or more '#' or ' ' here.

"/^.*log_level *=/" may mismatch some comments like "#you can specify log_level = xxx to blabla".

Comment (waynechen55) [reply tests/virt_autotest/virt_utils.pm:47]: Yes, I see.

File: lib/guest_installation_and_configuration_base.pm (hunk 2 above)
Comment (waynechen55) [inline lib/guest_installation_and_configuration_base.pm:795]: I just came across something about wicked. Hope this can be helpful;
https://en.opensuse.org/Portal:Wicked/Using_Wicked

Comment (waynechen55) [reply lib/guest_installation_and_configuration_base.pm:795]: And also this may help resolve your problem ? Anyway this looks like a quite old problem in SLE-12
https://www.suse.com/support/kb/doc/?id=000019064

File: data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml (hunk 1 above)
Comment (Julie-CAO) [reply data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml]: find the change about sshd in TW changelogs:
'''OpenSSH: move of the distro default config to /usr/etc. Admin config is supposed to be in /etc. Make sure to check the .rpmsave files after the update. Best to put config snippets in /etc/ssh/sshd_config.d/*.conf files to inherit distro config changes'''
'''- Change vendor configuration dir from /usr/share/ssh/ to
  /usr/etc/ssh/.
- Remove upgrade enablement hack. This has been fixed in
  systemd-rpm-macros (bsc#1180083).
- Add support for vendor provided configuration files in
  /usr/share/ssh/ (openssh-8.4p1-vendordir.patch)
- Move configuration files from /etc/ssh/ to /usr/share/ssh/
- Drop openssh-7.7p1-allow_root_password_login.patch to prevent login
  as root via password by default (is also upstream default). Comment
  indicates that this was a temporary meassure that we now had for
  five years, time to get rid of it (bsc#1173067)'''

tried /etc/ssh/sshd_config.d/sshd_config does not take effect, so still keep it /etc/ssh/sshd_config.

File: data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml [hunk 11]
```diff
@@ -174,5 +174,28 @@
...
     </user>
   </users>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- /etc/ssh/sshd_config/sshd_config is not default installed in TW -->
+    <init-scripts config:type="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config"
+if [ ! -f $sshd_config_file ]; then
```
waynechen55, line 184: What about add another line:
```
mkdir `dirname $sshd_config_file`
```

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
```diff
@@ -174,5 +174,28 @@
...
+    echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
+else
+    keys="PermitRootLogin PubkeyAuthentication PasswordAuthentication"
+    for key in $keys; do
+        sed -i "/^[# ]*$key */{h;s/^[# ]*$key *.*\$/$key yes/};\${x;/^\$/{s//$key yes/;H};x}" $sshd_config_file
+    done
+fi
+[ -d /root/.ssh ] || mkdir /root/.ssh; chmod 700 /root/.ssh
```
waynechen55, line 192: mkdir -p is better

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
```diff
@@ -174,5 +174,28 @@
...
+touch /root/.ssh/authorized_keys; chmod 600 /root/.ssh/authorized_keys
+echo "##Authorized-Keys##" >> /root/.ssh/authorized_keys
+]]>
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
```
waynechen55, line 199: By the way, there seems to be more elegant way to enable root ssh login with public key. Have you tried it ?
https://doc.opensuse.org/projects/autoyast/#Configuration-Security-users-and-groups

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
Julie-CAO: Yes, thanks.

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
Julie-CAO: ok

data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_xen_pv_guest_x86_64.xml
Julie-CAO: there is already the 'root' user and its auth keys in uefi guest autoyast control file, I coped that one and did not make change to <user> section. It worked in a previous snapshot but it did not take effect in the snapshot when sshd was changed. I guess the user creation procedure is prior to the sshd configuration files creation. Just my guess.

The autoyast document for opensuse Leap 15 seems to be outdated for TW. TW may be more like Leap 16.

lib/guest_installation_and_configuration_base.pm
Julie-CAO: Thanks. I'll try to fix the bridge problem and the network issue in another pr.

Julie-CAO:
verification runs with the latest commits passed. Rr is ready for merge. @alice-suse  Other remained issues listed in https://progress.opensuse.org/issues/96374 are still in progress.
[TW xen](http://10.67.129.51/tests/2596)
[TW kvm](http://10.67.129.51/tests/2594)
[SLE kvm](https://openqa.nue.suse.com/tests/6880571)

PR #13061, opened 2021-08-12
Add missing metadata in modules

punkioudi:
While working on the CI extension, it was observed that in some perl modules the `Summary` and `Maintainer` areas were missing from the file header.

- Related ticket: https://progress.opensuse.org/issues/96716
- Needles: N/A
- Verification run: N/A


tjyrinki (approved):
These are always welcome, thank you!

PR #13065, opened 2021-08-13
publiccloud::run_ltp Use LTP::WhiteList utils to evaluate failures

cfconrad:
Use LTP::WhiteList utility to evaluate failures.
Also extend skipped ltp test.

- Related ticket: https://progress.opensuse.org/issues/92470
- Verification run: 
  - http://cfconrad-vm.qa.suse.de/tests/8261 (whit LTP_KNOWN_ISSUES set)
  - ~http://cfconrad-vm.qa.suse.de/t8262 (without known issues)~
  - http://cfconrad-vm.qa.suse.de/tests/8263 (without known issues)

### DIFF
--- a/lib/LTP/WhiteList.pm
+++ b/lib/LTP/WhiteList.pm
@@ -1,4 +1,4 @@
-# Copyright © 2019 SUSE LLC
+# Copyright © 2019-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -28,7 +28,14 @@ use Mojo::UserAgent;
 use Mojo::JSON;
 use Mojo::File 'path';
 
-our @EXPORT = qw(download_whitelist override_known_failures is_test_disabled);
+our @EXPORT = qw(
+  download_whitelist
+  find_whitelist_testsuite
+  find_whitelist_entry
+  override_known_failures
+  is_test_disabled
+  list_skipped_tests
+);
 
 sub download_whitelist {
     my $path = get_var('LTP_KNOWN_ISSUES');
@@ -38,6 +45,7 @@ sub download_whitelist {
     unless ($res->is_success) {
         record_info("File not downloaded!", $res->message, result => 'softfail');
         set_var('LTP_KNOWN_ISSUES', undef);
+        return;
     }
     my $basename = $path =~ s#.*/([^/]+)#$1#r;
     save_tmp_file($basename, $res->body);
@@ -49,31 +57,21 @@ sub download_whitelist {
 sub find_whitelist_entry {
     my ($env, $suite, $test) = @_;
 
-    my $path = get_var('LTP_KNOWN_ISSUES');
-    return undef unless defined($path);
-
-    my $content = path($path)->slurp;
-    my $issues  = Mojo::JSON::decode_json($content);
-    return undef unless $issues;
-    return undef unless exists $issues->{$suite};
+    $suite = find_whitelist_testsuite($suite);
+    return undef unless ($suite);
 
     my @issues;
-    if (ref($issues->{$suite}) eq 'ARRAY') {
-        @issues = @{$issues->{$suite}};
+    if (ref($suite) eq 'ARRAY') {
+        @issues = @{$suite};
     }
     else {
         $test =~ s/_postun$//g if check_var('KGRAFT', 1) && check_var('UNINSTALL_INCIDENT', 1);
-        return undef unless exists $issues->{$suite}->{$test};
-        @issues = @{$issues->{$suite}->{$test}};
+        return undef unless exists $suite->{$test};
+        @issues = @{$suite->{$test}};
     }
 
-  ISSUE:
     foreach my $cond (@issues) {
-        foreach my $filter (qw(product ltp_version revision arch kernel backend retval flavor)) {
-            next ISSUE if exists $cond->{$filter} and $env->{$filter} !~ m/$cond->{$filter}/;
-        }
-
-        return $cond;
+        return $cond if (whitelist_entry_match($cond, $env));
     }
 
     return undef;
@@ -81,12 +79,29 @@ sub find_whitelist_entry {
 
 sub override_known_failures {
     my ($self, $env, $suite, $test) = @_;
-    my $entry = find_whitelist_entry($env, $suite, $test);
+    my $entry;
+
+    if ($env->{retval} && ref($env->{retval}) eq 'ARRAY') {
+        my %local_env = %$env;
+
+        my @retvals = grep { $_ ne 0 } @{$env->{retval}};
+        # if all retvals are zero, we might catch one of the `retval=>'^0$'` filters.
+        @retvals = (0) unless (@retvals);
+
+        for my $retval (@retvals) {
+            my $tmp = find_whitelist_entry({%$env, retval => $retval}, $suite, $test);
+            return 0 unless ($tmp);
+            $entry //= $tmp;
+        }
+    } else {
+        $entry = find_whitelist_entry($env, $suite, $test);
+    }
 
     return 0 unless defined($entry);
-    bmwqemu::diag("Failure in LTP:$suite:$test is known, overriding to softfail");
+    my $msg = "Failure in LTP:$suite:$test is known, overriding to softfail";
+    bmwqemu::diag($msg);
     $self->{result} = 'softfail';
-    $self->record_soft_failure_result($entry->{message}) if exists $entry->{message};
+    $self->record_soft_failure_result(join("\n", $msg, ($entry->{message} // ())));
     return 1;
 }
 
@@ -95,6 +110,45 @@ sub is_test_disabled {
 
     return 1 if defined($entry) && exists $entry->{skip} && $entry->{skip};
     return 0;
+}
+
+sub find_whitelist_testsuite {
+    my ($suite) = @_;
+
+    my $path = get_var('LTP_KNOWN_ISSUES');
+    return undef unless defined($path) and -e $path;
+
+    my $content = path($path)->slurp;
+    my $issues  = Mojo::JSON::decode_json($content);
+    return undef unless $issues;
+    return $issues->{$suite};
+}
+
+sub list_skipped_tests {
+    my ($env, $suite) = @_;
+    my @skipped_tests;
+    $suite = find_whitelist_testsuite($suite);
+    return @skipped_tests unless ($suite);
+    return @skipped_tests if (ref($suite) eq 'ARRAY');
+
+    for my $test (keys(%$suite)) {
+        my @entrys = grep { $_->{skip} && whitelist_entry_match($_, $env) } @{$suite->{$test}};
+        push @skipped_tests, $test if @entrys;
+    }
+    return @skipped_tests;
+}
+
+sub whitelist_entry_match
+{
+    my ($entry, $env) = @_;
+    my @attributes = qw(product ltp_version revision arch kernel backend retval flavor);
+
+    foreach my $attr (@attributes) {
+        next         unless defined $entry->{$attr};
+        return undef unless defined $env->{$attr};
+        return undef if ($env->{$attr} !~ m/$entry->{$attr}/);
+    }
+    return $entry;
 }
 
 1;
--- a/tests/publiccloud/run_ltp.pm
+++ b/tests/publiccloud/run_ltp.pm
@@ -12,20 +12,19 @@
 #
 # Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>, qa-c team <qa-c@suse.de>
 
-use base "publiccloud::basetest";
-use strict;
-use warnings;
+use Mojo::Base 'publiccloud::basetest';
 use testapi;
 use utils;
 use repo_tools 'generate_version';
 use Mojo::UserAgent;
 use LTP::utils qw(get_ltproot get_ltp_version_file);
-use LTP::WhiteList qw(download_whitelist);
+use LTP::WhiteList qw(download_whitelist find_whitelist_testsuite find_whitelist_entry list_skipped_tests override_known_failures);
 use Mojo::File;
+use Mojo::JSON;
 use publiccloud::utils qw(is_byos select_host_console);
+use Data::Dumper;
 
-our $root_dir      = '/root';
-our $ver_linux_log = '';
+our $root_dir = '/root';
 
 sub get_ltp_rpm
 {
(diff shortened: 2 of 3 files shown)


lib/LTP/WhiteList.pm
```diff
@@ -97,4 +96,46 @@ sub is_test_disabled {
...
+    }
+    return @skipped_tests;
+}
+
+sub whitelist_entry_match
+{
+    my ($entry, $env) = @_;
+    my @mandatory_attributes = qw(product ltp_version revision arch kernel backend retval flavor);
+
+    if (grep { !exists $env->{$_} } @mandatory_attributes) {
```
cfconrad, line 131: I'm not sure if we wanna be that strict here. The old version just produced a error like:
```
Use of uninitialized value $env->{$filter} in pattern match (m//) at -e line 1.
```
So maybe, we want just skip a filter, if the environment doesn't have that value?!
@baierjan  WDJT?

baierjan: Wow, this code changed a lot since I saw it last time... 😄  

Yes, it would be nice to skip missing filters to not generate warnings, however as far as I know, none of them is really a mandatory filter (and usually only a few are used for a single entry). The only mandatory attribute is the name of the test (variables `$suite` and `$test` from the `find_whitelist_entry`) no other assumption can be made. So making the attributes mandatory is not a good choice.

If you need more deep clarification, @mdoucha can help you here as he is more familiar with the code and current usage.

cfconrad: Thx, I will put @mdoucha into the reviewer list. To make my point clear here is a example:

ltp_known_issues.json:
```json
{
  "syscalls": {
    "ioctl_loop01": [
      {
        "message": "Kernel bug. WONTFIX for livepatches. bsc#1170794",
        "retval": "^1$",
        "product": "sle:15-SP2$",
        "revision": "kernel-livepatch-SLE15-SP2_Update_0$"
      }
    ]
}
```

The given `$env` would look like:
```perl
{
   product => 'sle:15-SP2',
   revision => 'kernel-livepatch-SLE15-SP2_Update_666'
   retval => 1
}
```

When we now call `whitelist_entry_match()` on that entry, it won't match because the `revision` doesn't.

**The question is, how should `whitelist_entry_match()` acct, if the `revision` isn't inside the `$env` like:**
```perl
whitelist_entry_match(
     {
        message => "Kernel bug. WONTFIX for livepatches. bsc#1170794",
        retval => '^1$',
        product =>  'sle:15-SP2$',
        revision => 'kernel-livepatch-SLE15-SP2_Update_0$'
      },
      {
        product => 'sle:15-SP2',
      }
);
```

cfconrad: While writing this, I think we should **not** `die()` if the env doesn't have all filters set, but plz comment!

baierjan: The whitelist entry should match if all provided filters match, so I would say if any of the filter is not in the environment, it is probably impossible for the filter to match (unless the filter says `'^$'`, but that's in my opinion a useless filter and can be omitted completely). Correct me if I am wrong, but the old version just emitted a warning and did the match with an empty string, I pretty much think that is the expected behavior. It would be probably nice to create a user-readable warning to let the user know, he is using either a non-standard filter or filter which is not valid in this context (and of course there might be also a typo in that filter).

We should probably emit `record_info` about missing env and assume an empty string for the filter evaluation, no reason to `die()`. Do you agree?

cfconrad: Yes I agree that we should not `die()`.

The old implementation did the match with a `undef` string (so warning appeared), which is always false, or? 

So we would like to keep that behavior, that the above example should *not* match, because we miss `retval` and `revision` in the environment?!

baierjan: > The old implementation did the match with a `undef` string (so warning appeared), which is always false, or? 

Yes, I would think so.

> So we would like to keep that behavior, that the above example should *not* match, because we miss `retval` and `revision` in the environment?!

Exactly, we want to match for the specific environment only. If they are missing, it is different environment and we should *not* match. Again, @mdoucha can correct me, if my assumptions are wrong.

mdoucha: The old implementation was [paired with code](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a910395f0804f8dc0b1c003f0c57afa9355e90a1/lib/LTP/utils.pm#L234) which always set all the environment variables to _something_. If you want to allow whitelist lookup with incomplete environment, which is *not* currently supported at all, the I suggest treating undefined env variables as silent match failure.

lib/LTP/WhiteList.pm
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2019 SUSE LLC
+# Copyright © 2021 SUSE LLC
```
pdostal, line 1:
```suggestion
# Copyright © 2019-2021 SUSE LLC
```

tests/publiccloud/run_ltp.pm
```diff
@@ -90,7 +130,20 @@ sub run {
...
+    download_whitelist();
+    my $ltp_env = gen_ltp_env($instance);
+    $self->{ltp_env} = $ltp_env;
+
+    # Use lib/LTP/WhiteList module to exclude tests
+    if (get_var('LTP_KNOWN_ISSUES')) {
+        my $exclude       = get_required_var('COMMAND_EXCLUDE');
+        my @skipped_tests = list_skipped_tests($ltp_env, get_required_var('COMMAND_FILE'));
+        if (@skipped_tests) {
+            $exclude .= '|' if (length($exclude) > 0);
```
pdostal, line 142:
```suggestion
            $exclude .= '|' if (length($exclude) > 1);
```

pdostal (approved):

lib/LTP/WhiteList.pm
cfconrad: Thanks for pointing it out and correcting me! I will change the code.

tests/publiccloud/run_ltp.pm
cfconrad: not sure about this. Why you want to be sure that the $exclude string is at least 2 chars long?

pdostal: I assume that this line is adding the `|` separator but it is needed only when there are at least two things to be separated.

mdoucha: `$exclude` is a string (regex). `|` needs to be added when the current string is not empty.

cfconrad:
ready from my side

tests/publiccloud/run_ltp.pm
```diff
@@ -48,6 +47,47 @@ sub instance_log_args
...
+    eval {
+        require OpenQA::Parser::Format::LTP;
+
+        my $ltp_log = Mojo::JSON::decode_json($log_file->slurp());
+        my $parser  = OpenQA::Parser::Format::LTP->new()->load($log_file->to_string);
+
+        if (find_whitelist_testsuite($ltp_testsuite)) {
+            for my $result (@{$parser->results()}) {
+                next unless ($result->{result} ne 'ok');
+                my ($ltp_test) = grep { $_->{test_fqn} eq $result->{test_fqn} } @{$ltp_log->{results}};
```
mdoucha, line 71: It'd be more efficient to turn `$ltp_log->{results}` into a hash. Syscalls testsuite has enough tests to make this O(n^2) array lookup slow if a significant number of them fails.

tests/publiccloud/run_ltp.pm
```diff
@@ -48,6 +47,47 @@ sub instance_log_args
...
+                for my $ret (@{$ltp_test->{test}->{retval}}) {
+                    if (override_known_failures($self, {%{$self->{ltp_env}}, result => $ret}, $ltp_testsuite, $result->{test_fqn})) {
```
mdoucha, line 73: I think the condition should be reversed. Softfail only if **all** retvals are zero or match a whitelist entry.

tests/publiccloud/run_ltp.pm
```diff
@@ -90,7 +130,20 @@ sub run {
...
     }
 
-    $ver_linux_log = $instance->run_ssh_command(cmd => get_ltproot() . "/ver_linux");
+    download_whitelist();
+    my $ltp_env = gen_ltp_env($instance);
+    $self->{ltp_env} = $ltp_env;
+
+    # Use lib/LTP/WhiteList module to exclude tests
+    if (get_var('LTP_KNOWN_ISSUES')) {
+        my $exclude       = get_required_var('COMMAND_EXCLUDE');
```
mdoucha, line 139: Why is this var required?

mdoucha (changes requested):

tests/publiccloud/run_ltp.pm
cfconrad: Agree here, the reason for this is "historical". 

We use the `runltp-ng --json_filter=openqa`. This produce a json file which can be consumed by https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Parser/Format/LTP.pm .
From my point of view, we (tests/publiccloud/run_ltp.pm) are the only user of that Parser, thus there is actually no need to have `--json_filter=openqa` anymore, instead adopt the `OpenQA::Parser::Format::LTP` to understand the latest produced output from `runltp-ng` or create a LTP-ng Parser.

But I can add that temp hash...

tests/publiccloud/run_ltp.pm
cfconrad: Makes sense, so if one of the retval's, do not match whitelist, we fail! THX!

tests/publiccloud/run_ltp.pm
cfconrad: Because we tell it is required later in the script as well. But not needed, thx!

lib/LTP/WhiteList.pm
```diff
@@ -97,4 +96,44 @@ sub is_test_disabled {
...
+    return $issues->{$suite};
+}
+
+sub list_skipped_tests {
+    my ($env, $suite) = @_;
+    my @skipped_tests;
+    $suite = find_whitelist_testsuite($suite);
+    return @skipped_tests unless ($suite);
+
+    die 'Unsupported format for `list_skipped_tests()`' if (ref($suite) eq 'ARRAY');
```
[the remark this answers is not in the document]
cfconrad: ok

cfconrad:
@mdoucha thx for review, I added your advises in `fixup!` commits.
Plz let me know if it is ok, then I'm going to squash it.

tests/publiccloud/run_ltp.pm
```diff
@@ -66,13 +66,15 @@ sub upload_ltp_logs
...
-                next unless ($result->{result} ne 'ok');
-                my ($ltp_test) = grep { $_->{test_fqn} eq $result->{test_fqn} } @{$ltp_log->{results}};
-                for my $ret (@{$ltp_test->{test}->{retval}}) {
-                    if (override_known_failures($self, {%{$self->{ltp_env}}, result => $ret}, $ltp_testsuite, $result->{test_fqn})) {
+                # Mark only as softfail, if _all_ `retval` have a "known_issue" entry or the retval was 0
+                my @failed_attempts = grep { $_ ne 0 } @{$ltp_log_results{$result->{test_fqn}}->{retval}};
+                my @known_failures  = grep { find_whitelist_entry({%{$self->{ltp_env}}, result => $_}, $ltp_testsuite, $result->{test_fqn}) } @failed_attempts;
+
+                if (@failed_attempts > 0 && @failed_attempts == @known_failures) {
+                    if (override_known_failures($self, {%{$self->{ltp_env}}, result => shift @known_failures}, $ltp_testsuite, $result->{test_fqn})) {
```
mdoucha, line 76: The code above is a bit too complicated. It'd be easier to find retvals that are *not* whitelisted like this:

    my @new_fails = grep { $_ ne 0 && !find_whitelist_entry(...) } @{$ltp_log_results{$result->{test_fqn}}->{retval}};

then call `override_known_failures()` if the array is empty. You don't need to check the return value then because you've already verified it'll be 1.

Also, the env variable is `retval`, not `result`.

cfconrad: I would like to put the evaluation logic into `LTP::WhiteList` so it is done "always" the same way, WDYT about 1233dcbdaea626c464b4367528c77bf4a7ba3618

I have a problem with `$retval eq 0`. This actually makes the `override_known_failures()` behaves inconsistently see this https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13065/commits/49e87aff74ad3449fdd88bad0372ac28c2bfa659#diff-9f0479ba67450b0dd9dd303a261b4bcdef42036ff154db485b12dbde2600ad89R85 I would expect here, that we do not overwrite, like we do with `$retval = [ 0 ];`. 
But I would like to double check with you, if you also think we should return 0, if someone pass a environment with `$retval = 0;` to `override_known_failures()`. 
Or we keep it up to the caller and don't extract `0` from `$retval` if the retval is an array.

Also added a unit test for it :face_with_thermometer:

lib/LTP/WhiteList.pm
```diff
@@ -49,44 +57,50 @@ sub download_whitelist {
...
 sub override_known_failures {
     my ($self, $env, $suite, $test) = @_;
-    my $entry = find_whitelist_entry($env, $suite, $test);
+    my $entry;
+
+    if ($env->{retval} && ref($env->{retval}) eq 'ARRAY') {
+        my %local_env = %$env;
+
+        my @retvals = grep { $_ ne 0 } @{$env->{retval}};
+        return 0 unless (@retvals);
```
mdoucha, line 88: Hmm, this is a bit more complicated. We have some whitelist entries with `retval: ^0$` for tests which pass but trigger a harmless kernel warning. You should still check for those if all retvals are zero.

cfconrad: Lets just not exclude zero at all then!?

mdoucha: That's the fun thing: This may not work either. If you have a retval array like `[0, 0, 0, 1, 0]` and there is a whitelist entry for `retval: ^1$` which is caused by a known random error, you *need* to ignore the zeroes.

mdoucha: IOW: when `@retvals` is empty here, just search for whitelist entry with `retval => 0` instead of returning.

cfconrad: ok

cfconrad: plz check latest, thx

mdoucha (approved):
OK, this should work.

cfconrad:
VR: 
  * http://cfconrad-vm.qa.suse.de/t8276 (with known issues)
  * http://cfconrad-vm.qa.suse.de/t8277 (without known issues)

PR #13068, opened 2021-08-13
Move opensuse tests schedule to separate folder

OleksandrOrlov:
Currently, almost all the schedule files for opensuse from YaST job group contain _opensuse suffix and placed under yast directory.

While making bulk change in yast schedules it requires a lot of attention to not introduce unnecessary changes in opensuse schedules.

Also, spreading those tests over yast folder makes them harder to find and identify of what schedules already created for opensuse and what are still use main.pm directly.

This only affects the schedules that are located in `schedule/yast/` folder.

- Related ticket: https://progress.opensuse.org/issues/95748
- Verification runs: not needed as there is no changes in files. Only changes in Job Group settings should be made.

**Important:** `schedule/yast/` should be replaced with `schedule/yast/opensuse/` in all the Job Group settings on o3. I kept the same directory structure inside of `schedule/yast/opensuse/` as in `schedule/yast/`, so the replacement should be easy.


DimStar77:
https://openqa.opensuse.org/tests/1908812

Reason: tests died: unable to load main.pm, check the log for the cause (e.g. syntax error)

OleksandrOrlov:
I've changed path in YAML_SCHEDULE to the correct one and restarted the job. It passes now: https://openqa.opensuse.org/tests/1909005

Thanks for your report!

PR #13074, opened 2021-08-16
Add test scenario for PC-AMD-SEV

grisu48:
Adding a test scenario for running a smoke test on the AMD-SEV secured
publiccloud instances. Also introduce the `smoketest` test scenario, which can be used for basic smoketest on more non-standard machine types.

- Related ticket: https://progress.opensuse.org/issues/94634
- Verification run: 15-SP3 [GCE AMD-SEV](http://duck-norris.qam.suse.de/tests/6843) | [GCE default](http://duck-norris.qam.suse.de/tests/6844)

### DIFF
--- a/data/publiccloud/terraform/gce.tf
+++ b/data/publiccloud/terraform/gce.tf
@@ -74,6 +74,10 @@ variable "tags" {
     default = {}
 }
 
+variable "enable_confidential_vm" {
+	default=false
+}
+
 resource "random_id" "service" {
     count = var.instance_count
     keepers = {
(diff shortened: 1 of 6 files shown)


grisu48:
I acknowledge @pdostal comment, that `smoketest.pm` is functional a subset of the `instance_overview.pm`, however those two testcases have two different purposes:

* `instance_overview` is collecting information for debugging
* `smoketest` is a collection of very basic tests to check the health of the instance

This is why I keep them separated.

pdostal:
> This is why I keep them separated.

You might improve the `instance_overview` instead ... maybe even rename it :smile:

pdostal (approved):

grisu48:
> > This is why I keep them separated.
> 
> You might improve the `instance_overview` instead ... maybe even rename it smile

I consider this as a prototype for https://progress.opensuse.org/issues/95557. However it is unclear to me if we want to keep this one or simply use `imgproof` instead. I'm in favour of the later, but this is outside the scope of this PR.

PR #13075, opened 2021-08-16
Softfail known samba_adcli failures

ge0r:
A number of samba_adcli failures have been identified as seen in
poo#96980.

Those are known failures and should be marked as soft failures,
in order to avoid coverage loss, as they are sporadic and currently
there is less benefit to unschedule the module.

- Related ticket: https://progress.opensuse.org/issues/96992
- Needles: No needles
- Verification run: [example soft fail in x86_64](https://openqa.suse.de/tests/6865752#step/samba_adcli/65) | [different soft fail in x86_64](http://angmar.suse.de/tests/3701#step/samba_adcli/166)

### DIFF
--- a/tests/network/samba/samba_adcli.pm
+++ b/tests/network/samba/samba_adcli.pm
@@ -75,7 +75,10 @@ sub update_password {
                 record_soft_failure("bsc#1188390");
 
                 # Retrying the adcli join is needed, probably due to https://bugs.freedesktop.org/show_bug.cgi?id=55487
-                script_retry 'adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 5, timeout => 60;
+                if (script_retry('adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 15, timeout => 60, die => 0) != 0) {
+                    record_soft_failure('poo#96983');
+                    return;
+                }
             }
         }
 
@@ -121,7 +124,10 @@ sub run {
     assert_script_run "expect kinit.exp";
 
     # Retrying the adcli join is needed, probably due to https://bugs.freedesktop.org/show_bug.cgi?id=55487
-    script_retry 'adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 5, timeout => 60;
+    if (script_retry('adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 15, timeout => 60, die => 0) != 0) {
+        record_soft_failure('poo#96983');
+        return;
+    }
 
     #Verify if machine already added
     assert_script_run "adcli info -D geeko.com  -S $AD_hostname -v";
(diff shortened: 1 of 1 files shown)


tests/network/samba/samba_adcli.pm
```diff
@@ -75,7 +75,10 @@ sub update_password {
                 record_soft_failure("bsc#1188390");
 
                 # Retrying the adcli join is needed, probably due to https://bugs.freedesktop.org/show_bug.cgi?id=55487
-                script_retry 'adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 5, timeout => 60;
+                if (script_retry('adcli join -v -W --domain geeko.com -U Administrator -C', delay => 10, retry => 15, timeout => 60, die => 0) != 0) {
+                    record_soft_failure('poo#96983');
```
okurz, line 79: I recommend against using soft-fails referencing test issues. You (or your team) should already know that the ticket still exists. What's the point in informing test reviewers about test issues? You can use `record_info` though to explain workarounds

tests/network/samba/samba_adcli.pm
```diff
@@ -90,7 +93,7 @@ sub update_password {
                 record_soft_failure("bsc#1188575");
                 last;
             } else {
-                die("wbinfo output does not contain $substr");
+                record_soft_failure("wbinfo output does not contain $substr");
```
okurz, line 96: use `record_soft_failure` only with ticket references, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/Makefile#L146

okurz (changes requested):

okurz:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13072 is the alternative proposal to disable tests temporarily

tests/network/samba/samba_adcli.pm
ge0r: you are right! thanks

tests/network/samba/samba_adcli.pm
ge0r: If I use record_info it gives the idea of providing some extra info in an otherwise normal test run.
In that case we need to communicate that the test run is not without issues and that the potential softfailed run deviates from the intended one.

okurz: yes but you got to be honest: Who cares? Only the test maintainers care and they already know the ticket. Whoever creates a maintenance submission that is tested same as release managers of new builds of products will just care about "can we release or not"? And *maybe* some of them care about which product issues still exist despite releasing. This is where soft-fails should be used, pointing to real product issues. But who of these people should really care or have the time to care about "the test is not behaving exactly as the test writers intended"?

ge0r: Since I am on a tight deadline here (as was mentioned in the rocketchat thread) I will mark this as resolved since I believe that my reply explains the reasoning behind choosing `record_soft_failure` over `record_info`.
If you do not agree please do revert this tmrw and we can discuss it further.

ge0r: Even if it is just for the test maintainers, I believe that it is a valid semantic to show that the test case has not been run as intended. A soft failed test sends a visible, clear message that there is something wrong with the way the test case was covered on the specific run. I believe we should not downgrade those cases to just recording information of progress issues, since it is an ambiguous and unclear message to those who care (the maintainers)

punkioudi (approved):
Lgtm George, I agree with the test being indicated as soft-failed (when it does soft-fails), as the rest of the code might not be executed.

tests/network/samba/samba_adcli.pm
ge0r: Since I am on a tight deadline (as was mentioned on the thread on rocketchat) and I cannot wait for a further response, I will mark this as resolved. If however you find that my comment does not back why I chose soft_failing over record_info, feel free to revert this PR if it is merged.

PR #13078, opened 2021-08-17
Fix SLE15SP4 staging project with GNOME40 upgrade

zhaocongjia:
This is to fix problems in "sle-15-SP4-Online-E-Staging" on o.s.d, which includes newly introduced GNOME40 stack.

The code paths to handle GNOME40 was already added by previous Tumbleweed tests, this PR just enables them for SLE15SP4+ and Leap15.4+.

- Related ticket: https://progress.opensuse.org/issues/96875
- Needles: Created a few in osd by developer mode
- Verification run: https://openqa.suse.de/tests/6857221#step/x_vt/7 https://openqa.suse.de/tests/6857221#step/reboot_gnome/2

### DIFF
--- a/tests/console/x_vt.pm
+++ b/tests/console/x_vt.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2019 SUSE LLC
+# Copyright © 2012-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
@@ -15,15 +15,16 @@ use base "consoletest";
 use strict;
 use warnings;
 use testapi;
-use version_utils qw(is_tumbleweed is_leap);
+use version_utils qw(is_tumbleweed is_leap is_sle);
 
 sub run {
+    # First, list all X processes, including user process and gdm process
+    script_run('ps -ef | grep bin/X');
     if (script_run("ps -ef | grep bin/X | egrep 'tty7|wayland'") == 1) {
-        if ((is_tumbleweed || is_leap('>=15.2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
-            record_soft_failure('boo#1138327');
+        if ((is_tumbleweed || is_leap('>=15.2') || is_sle('>=15-SP2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
+            diag('user X process runs on tty2 for systems with GNOME 3.32+, see boo#1138327');
         }
         else {
-            script_run('ps -ef | grep bin/X');
             die 'Expected tty7 used by X or wayland not found';
         }
     }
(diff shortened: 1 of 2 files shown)


tests/console/x_vt.pm
```diff
@@ -15,15 +15,16 @@ use base "consoletest";
...
     if (script_run("ps -ef | grep bin/X | egrep 'tty7|wayland'") == 1) {
-        if ((is_tumbleweed || is_leap('>=15.2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
-            record_soft_failure('boo#1138327');
+        if ((is_tumbleweed || is_leap('>=15.2') || is_sle('>=15-SP2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
+            diag('user X process runs on tty2 for systems with GNOME 3.32+, see boo#1138327');
         }
         else {
-            script_run('ps -ef | grep bin/X');
             die 'Expected tty7 used by X or wayland not found';
         }
```
GraceWang571, line 29: How about using the below code (which I think is briefer)?

```
if (script_run("ps -ef | grep bin/X | egrep 'tty2|tty7|wayland'") == 1) {
    die 'Expected tty2 or tty7 used by X or wayland not found';
}
```

GraceWang571: Sorry, I quoted the wrong code. I mean as a replacement of the below part of the code 
```
    if (script_run("ps -ef | grep bin/X | egrep 'tty7|wayland'") == 1) {
        if ((is_tumbleweed || is_leap('>=15.2') || is_sle('>=15-SP2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
            diag('user X process runs on tty2 for systems with GNOME 3.32+, see boo#1138327');
        }
        else {
            die 'Expected tty7 used by X or wayland not found';
        }
```

zhaocongjia: Different system expects different tty.  If SLE15SP4 starts at tty7, that would be a bug.  So the testing of system version is necessary.

GraceWang571: So, what about SLE15SP3 or before? Is it supposed to start at tty7 or tty2?

zhaocongjia: Before SLE15SP2, they are handled by the first `if`, expected at tty7.  After SLE15SP2, they are handled by the nested `if`, expected at tty2.

GraceWang571: Yes, that's what confused me. 
See the below link for SLE15SP3 https://openqa.suse.de/tests/5990447#step/x_vt/1, it shows tty7.

zhaocongjia: Good catch.  That is because it is `minimal_x`, xdm starts at tty7.  For SLE15SP3 with gdm, gdm starts at tty7 and user X session starts at tty2.

GraceWang571: aha, I see. Thanks for the explanation. I have no more questions.

PR #13079, opened 2021-08-17
Fix lvm_encrypt_separate_boot QAM test_data

sofiasyria:
After commit 601b2ed37e , the partitioning test_data where changed for ppc and 64bit as well. The two archs need separate test data in order for the new_partitioning_gpt to work properly.
 
- Related ticket: https://progress.opensuse.org/issues/96842
- Verification runs:
64bit: http://falafel.suse.cz/tests/524
ppc: https://openqa.suse.de/tests/6869561


tjyrinki (approved):
Looks great, thank you for testing too. Will look at the related yaml schedule change separately.

PR #13087, opened 2021-08-18
Add workaround to move down in pattern selection while bsc#1189550

jknphy:
- Related tickets: 
  - https://progress.opensuse.org/issues/97004
  - https://bugzilla.suse.com/show_bug.cgi?id=1189550
- Needles: [MR#1554](https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1554)
- Verification run: 
  - [mru-install-minimal-with-addons](https://openqa.suse.de/tests/6881542)

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -267,6 +267,11 @@ sub select_specific_patterns_by_iteration {
     # delete special 'all' and 'default' keys from the check
     delete $patterns{default};
     delete $patterns{all};
+
+    if (is_sle('>=15-sp3')) {
+        record_soft_failure('bsc#1189550 - Problem with scroll bar event in pattern selection');
+    }
+
     while (1) {
         die "looping for too long" unless ($counter--);
         my $needs_to_be_selected;
@@ -297,6 +302,7 @@ sub select_specific_patterns_by_iteration {
         my $selected = check_screen([qw(current-pattern-selected on-category)], 0);
         if ($selected && $selected->{needle}->has_tag('on-category')) {
             move_down;
+            workaround_bsc1189550() if is_sle('>=15-sp3');
             next;
         }
         if ($needs_to_be_selected && !$selected) {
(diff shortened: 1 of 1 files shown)


lib/y2_installbase.pm
```diff
@@ -267,6 +267,11 @@ sub select_specific_patterns_by_iteration {
     # delete special 'all' and 'default' keys from the check
     delete $patterns{default};
     delete $patterns{all};
+
+    if (is_sle('>=15-sp3')) {
+        record_soft_failure('bsc#1189550 - Problem with scroll bar event in pattern selection');
```
czerw, line 272: I would move record_info into `workaround_bsc1189550` function

lib/y2_installbase.pm
```diff
@@ -297,6 +302,7 @@ sub select_specific_patterns_by_iteration {
         my $selected = check_screen([qw(current-pattern-selected on-category)], 0);
         if ($selected && $selected->{needle}->has_tag('on-category')) {
             move_down;
+            workaround_bsc1189550() if is_sle('>=15-sp3');
```
czerw, line 305: What about 15-SP4, do we know it is impacted in the same way?

czerw:
Faulty needle `first_boot-emergency-mode-20210813` should be removed as part of workaround.

lib/y2_installbase.pm
jknphy: it would appear too many times, to many soft-failed boxes. We need to remember to delete here, is the disadvantage but it is more clear as it is not trivial to find a code to detect it, as it happens always for the last item when scrolling.

lib/y2_installbase.pm
jknphy: yes, it reached staging too: https://progress.opensuse.org/issues/97088

jknphy:
> Faulty needle `first_boot-emergency-mode-20210813` should be removed as part of workaround.

https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1554

lib/y2_installbase.pm
czerw: ok

michaelgrifalconi:
Change was quite straight-forward so I just merged, regardless of the broken pipeline :) @jknphy

PR #13088, opened 2021-08-18
kernel: Switch kexec test to select_serial_terminal

czerw:
Fix poo#97004: Use select_serial_terminal instead of root console to
make test more robust. Test maintainer was updated.

- Related ticket: https://progress.opensuse.org/issues/97004
- Needles: none
- Verification run: 
15-SP4 x86_64: https://openqa.suse.de/tests/6882461
15-SP4 aarch64: https://openqa.suse.de/tests/6882466
15-SP4 ppc64le: https://openqa.suse.de/tests/6882467
15-SP3 x86_64: https://openqa.suse.de/tests/6882463
15-SP2 x86_64: https://openqa.suse.de/tests/6882480
15 x86_64: https://openqa.suse.de/tests/6882544
12-SP5 x86_64: https://openqa.suse.de/tests/6882647
12-SP4 x86_64: https://openqa.suse.de/t6882548

### DIFF
--- a/tests/kernel/kernel_kexec.pm
+++ b/tests/kernel/kernel_kexec.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2017-2018 SUSE LLC
+# Copyright (C) 2017-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -15,7 +15,7 @@
 
 # Package: kexec-tools systemd
 # Summary:  [qa_automation] kexec testsuite
-# Maintainer: Nathan Zhao <jtzhao@suse.com>
+# Maintainer: QE Kernel <kernel-qa@suse.de>
 
 use base "opensusebasetest";
 use strict;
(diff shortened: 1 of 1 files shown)


czerw:
@mdoucha @pevik please check

mdoucha:
It looks good but `select_serial_terminal()` after `kexec` will only wait 95 seconds for login prompt. Do you think it'll be enough in all cases?

pevik:
> It looks good but `select_serial_terminal()` after `kexec` will only wait 95 seconds for login prompt. Do you think it'll be enough in all cases?

Good point, but shouldn't this be handled somewhere in https://github.com/os-autoinst/os-autoinst/blob/master/consoles/serial_screen.pm or https://github.com/os-autoinst/os-autoinst/blob/master/consoles/virtio_terminal.pm (add proper timeout based on backend/arch?)
I suppose one of them handles the timeout. Or is the timeout based on needles setup?

mdoucha:
This is normally handled by `wait_boot()` which ensures that the system will be fully booted before calling `select_serial_terminal()`. But `kexec` bypasses Grub so `$self->wait_boot_past_bootloader()` would be the correct method to call here.

pevik (approved):

czerw:
Added missing "$self->wait_boot_past_bootloader()"  and updated verifications in description.

mdoucha (approved):
LGTM

PR #13092, opened 2021-08-18
Provide test_data for validate_blockdevices test

rakoenig:
- on openSUSE the validate_blockdevices test seems to just check for
  the prompt of the root shell.
- this file provides the default disk layout, so that the
  validate_blockdevices test can check against it
- You need to provide this file as YAML_TEST_DATA setting

- Related ticket: https://progress.opensuse.org/issues/97142
- Verification run: https://openqa.opensuse.org/tests/1880239

### DIFF
--- a/test_data/yast/xfs/xfs_partition_opensuse.yaml
+++ b/test_data/yast/xfs/xfs_partition_opensuse.yaml
@@ -1,0 +1,22 @@
+---
+disks:
+  - name: vda
+    table_type: gpt
+    allowed_unpartitioned: 0.00GB
+    partitions:
+      - name: vda2
+        role: operating-system
+        formatting_options:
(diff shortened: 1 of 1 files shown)


test_data/yast/xfs/xfs_partition_opensuse.yaml
```diff
@@ -0,0 +1,24 @@
+---
+disks:
+  - name: vda
+    table_type: gpt
+    allowed_unpartitioned: 0.00GB
+    partitions:
+      - name: vda2
+        role: operating-system
+        partition_type: primary
```
jknphy, line 9: could you please delete the partition_type key in the file? that is used just for msdos. `should_mount` is still needed in `validate_read_write`.

PR #13094, opened 2021-08-19
JeOS: fix schedule for fips test case

jlausuch:
bootloader_uefi was executed for all the cases. It should be
only executed in specific machine types.


jlausuch:
VR: https://openqa.suse.de/tests/6893157

PR #13095, opened 2021-08-19
Do not fail when fips_enabled file does not exist

ge0r:
File `/proc/sys/crypto/fips_enabled` does not exist in some systems.
Do not fail when getting it's content while it does not exist.

- Related ticket: https://progress.opensuse.org/issues/97187
- Needles: No needles
- Verification run: pending


Vogtinator (approved):

Vogtinator:
VR is still pending, but code looks good and unlikely to make it worse.

ge0r:
Updated VR

PR #13097, opened 2021-08-19
console/publiccloud: Increase timeout to download data dir

cfconrad:
Current data/ directory size is 230M, if I create a tar.gz it gets only
reduced to 215M.
So give PUBLIC_CLOUD tests 10min to download the directory to the
running instance.

- Related ticket: https://progress.opensuse.org/issues/97208
- Verification run: https://openqa.suse.de/tests/6893493

### DIFF
--- a/tests/console/prepare_test_data.pm
+++ b/tests/console/prepare_test_data.pm
@@ -25,7 +25,7 @@ sub run {
     check_var("BACKEND", "ipmi") ? use_ssh_serial_console : select_console 'root-console';
 
     select_console 'user-console';
-    assert_script_run "curl -L -v -f " . autoinst_url('/data') . " | cpio -id", timeout => 300;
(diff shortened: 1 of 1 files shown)


cfconrad:
This is one, where it would take 7.5minutes https://openqa.suse.de/tests/6890171#step/prepare_test_data/3

pdostal:
The problem of this module is: Do we need it?
Most of the consoletests are downloading just their own assets.

I actually thought we excluded this already.

cc: @grisu48

pdostal:
Different approach in #13115

cfconrad:
This is the better solution: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13115 thx @pdostal

pdostal:
Thank you Clemens.

PR #13098, opened 2021-08-19
Add test_metadata in ci check

punkioudi:
It checks that the Copyright, Summary and Maintainer fields in the file header of the tests are added.

- Related ticket: https://progress.opensuse.org/issues/96716
- Needles: N/A
- Verification run: N/A

### DIFF
--- a/tools/check_metadata
+++ b/tools/check_metadata
@@ -4,6 +4,7 @@ success=1
 for file in $files; do
     grep -q '# Summary: .\+' $file || (echo "Missing '# Summary: <multi line summary of test>'" in $file && exit 1) || success=0
     grep -q '# Maintainer: .\+@.\+' $file || (echo "Missing '# Maintainer: <email address>'" in $file && exit 1) || success=0
+    grep -q '# Copyright .\+' $file || (echo "Missing Copyright © <year> SUSE LLC" in $file && exit 1) || success=0
 done
 [ $success = 1 ] && echo "SUCCESS" && exit 0
 exit 1
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ test-spec:
 	tools/update_spec --check
 
 .PHONY: test-static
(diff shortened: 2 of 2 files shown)


tools/check_metadata
```diff
@@ -4,6 +4,7 @@ success=1
 for file in $files; do
     grep -q '# Summary: .\+' $file || (echo "Missing '# Summary: <multi line summary of test>'" in $file && exit 1) || success=0
     grep -q '# Maintainer: .\+@.\+' $file || (echo "Missing '# Maintainer: <email address>'" in $file && exit 1) || success=0
+    grep -q '# Copyright .\+' $file || (echo "Missing Copyright © <year> SUSE LLC" in $file && exit 1) || success=0
```
foursixnine, line 7: is within the scope of this pr to also check if one of the files being modified, has an updated header? otherwise lgtm

punkioudi: no, not yet, it will be handled in another pr (when I am back from vacation :) )

ge0r (approved):
LGTM.
One side question, not related to your changes, shouldn't [this line](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/cfa428315959b4a696dfb1d0f03fcc3c56af1073/Makefile#L78) also apply for `.pm` files in the `lib/` directory?
It seems like they also have similar metadata in need of checking.

punkioudi:
> LGTM.
> One side question, not related to your changes, shouldn't [this line](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/cfa428315959b4a696dfb1d0f03fcc3c56af1073/Makefile#L78) also apply for `.pm` files in the `lib/` directory?
> It seems like they also have similar metadata in need of checking.

Hm, nice catch George! I will add another progress issue to the story of the ci ticket, as I suspect that there are a lot more in the `lib/` directory that don't have the proper file header.

PR #13099, opened 2021-08-19
xfstests: Add xfstests basic tests in o3

SeroSun:
Add xfstests basic tests into o3.
- Modify opensuse/main.pm to goto xfstests test path
- Modify xfstests/install.pm to use obs for non-sle test
- Correct test path for xfstests in non-sle test, because default install path not the same

- Verification run: 
create_hdd_xfstests:             http://10.67.133.102/tests/360
xfstests_btrfs-btrfs-001-050: http://10.67.133.102/tests/361


lansuse:
LGTM

PR #13101, opened 2021-08-19
container: Fix version string validation in test_opensuse_based_image()

cfconrad:
We use openQA `BETA` variable to determine if a beta Version string is
available or not. But this function is also used to verify containers
running on a host. So it doesn't belong to the `BETA` varialbe of the
test.

- Ticked:  https://progress.opensuse.org/issues/97223
- Verification run: TBD

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -152,8 +152,9 @@ sub test_opensuse_based_image {
     my $image   = $args{image};
     my $runtime = $args{runtime};
 
-    my $distri  = $args{distri}  //= get_required_var("DISTRI");
-    my $version = $args{version} //= get_required_var("VERSION");
+    my $distri  = $args{distri}  // get_required_var("DISTRI");
+    my $version = $args{version} // get_required_var("VERSION");
+    my $beta    = $args{beta}    // get_var('BETA', 0);
 
     die 'Argument $image not provided!'   unless $image;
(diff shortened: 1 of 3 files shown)


lib/containers/container_images.pm
```diff
@@ -152,8 +152,9 @@ sub test_opensuse_based_image {
     my $image   = $args{image};
     my $runtime = $args{runtime};
 
-    my $distri  = $args{distri}  //= get_required_var("DISTRI");
-    my $version = $args{version} //= get_required_var("VERSION");
+    my $distri  = $args{distri}  // get_required_var("DISTRI");
+    my $version = $args{version} // get_required_var("VERSION");
+    my $beta    = $args{beta}    // get_var('BETA', 1);
```
jlausuch, line 157: why default to 1? I would say if BETA is not set, we should consider that default is 0.

jlausuch:
s/varialbe/variable

pdostal:
Please update also `docker_image` test module.

lib/containers/container_images.pm
cfconrad: oh yes... thx!

jlausuch:
Looks good. Can you please run some VRs?

cfconrad:
The VR is failing on something different. But I t hink the fix works: http://cfconrad-vm.qa.suse.de/tests/8285#step/podman_image/95

pdostal (approved):

PR #13102, opened 2021-08-19
Ignore bsc#1126272 in journal_check for SLE Micro and MicroOS

jlausuch:
The bug has been reported as Won't fix and it appears on different
products besides MicroOS.


Vogtinator (approved):
Unfortunately I agree...

jlausuch:
Agreed with Jiri as well.

PR #13104, opened 2021-08-19
Fix condition which assigns the proper dockerfile template

b10n1k:
With the condition as it was before the `$dockerfile` variable was always
assigned to the one that uses the python image. That may not make the tests
fail but it was not the expected behavior.

The `dockerfile.python3` should be used only for the hosts which do not support
`container-suseconnect-zypp`. Use `dockerfile.python3` to workaround wherever we
cant perform apache installation as it happens in `Dockerfile`.

- Related ticket: https://progress.opensuse.org/issues/95786
Verification run: 
https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313104&version=15-SP3&distri=sle
- [sle](https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313104&distri=sle&version=15-SP2)
- [tw](https://openqa.opensuse.org/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313104&distri=opensuse&version=Tumbleweed)

### DIFF
--- a/tests/containers/docker_image.pm
+++ b/tests/containers/docker_image.pm
@@ -38,7 +38,7 @@ sub run {
 
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGES
     my $versions   = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
-    my $dockerfile = ($host_distri ne 'suse') ? 'Dockerfile.python3' : 'Dockerfile';
+    my $dockerfile = $host_distri !~ m/^(sle|opensuse)/i ? 'Dockerfile.python3' : 'Dockerfile';
     for my $version (split(/,/, $versions)) {
         my ($untested_images, $released_images) = get_suse_container_urls($version);
         my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
--- a/tests/containers/podman_image.pm
+++ b/tests/containers/podman_image.pm
@@ -28,9 +28,8 @@ sub run {
     allow_selected_insecure_registries(runtime => $runtime);
 
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGES
-    my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
-
-    my $dockerfile = ($host_distri ne 'suse') ? 'Dockerfile.python3' : 'Dockerfile';
+    my $versions   = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
+    my $dockerfile = $host_distri !~ m/^(sle|opensuse)/i ? 'Dockerfile.python3' : 'Dockerfile';
     for my $version (split(/,/, $versions)) {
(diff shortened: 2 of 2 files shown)


tests/containers/docker_image.pm
```diff
@@ -38,7 +38,7 @@ sub run {
 
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGES
     my $versions   = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
-    my $dockerfile = ($host_distri ne 'suse') ? 'Dockerfile.python3' : 'Dockerfile';
+    my $dockerfile = !check_os_release('suse', 'PRETTY_NAME') ? 'Dockerfile.python3' : 'Dockerfile';
```
cfconrad, line 41: I'm not sure, but for me it took me a time to understand it. What about be more explicit, so it is easy to read like:
```suggestion
    my $dockerfile = $host_distri !~ m/^(sle|opensuse-.*)$/ ? 'Dockerfile.python3' : 'Dockerfile';
```

tests/containers/podman_image.pm
```diff
@@ -28,16 +28,15 @@ sub run {
...
+    my $dockerfile = !check_os_release('suse', 'PRETTY_NAME') ? 'Dockerfile.python3' : 'Dockerfile';
     for my $version (split(/,/, $versions)) {
         my ($untested_images, $released_images) = get_suse_container_urls($version);
         my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
         for my $iname (@{$images_to_test}) {
             record_info "IMAGE", "Testing image: $iname";
-            test_container_image(image => $iname, runtime => $runtime);
-            test_rpm_db_backend(image => $iname, runtime => $runtime);
+            #test_container_image(image => $iname, runtime => $runtime);
+            #test_rpm_db_backend(image => $iname, runtime => $runtime);
```
pdostal, line 39: Why this?

tests/containers/docker_image.pm
b10n1k: `check_os_release` has been tested extensively IMO and is quite solid. I would rather leave it as it is. However your suggestion might be faster. but i think it doesnt matter here.

tests/containers/podman_image.pm
b10n1k: stupid me man. i had done that for testing local and i forgot it. thats why

b10n1k: **this change is ready for review**

tests/containers/docker_image.pm
pdostal: I agree with Clemens. We already have `$host_distri` so let's use it please.

b10n1k: thats funny argument as we already use `check_os_release` too. which means that we do not have to import anything mew too. tespa. i did the change. *this change is ready for review*

pdostal (approved):

jlausuch (approved):

PR #13105, opened 2021-08-19
container: Use serial terminal for image test modules

cfconrad:
Simple exchange `root-console` with `$self->select_serial_console()`
reduce the execution time on s390 from 1h8m to 21m.

- Verification run: 
  - https://openqa.suse.de/tests/6898866 (s390 with serial_terminal) => 21m

Compared with this https://openqa.suse.de/tests/6893576 using `root-console` it takes 1h8m.

### DIFF
--- a/tests/containers/docker_image.pm
+++ b/tests/containers/docker_image.pm
@@ -28,7 +28,9 @@ use containers::urls 'get_suse_container_urls';
 use version_utils qw(get_os_release check_os_release is_tumbleweed);
 
 sub run {
-    select_console "root-console";
+    my $self = shift;
+    $self->select_serial_terminal();
+
     my ($running_version, $sp, $host_distri) = get_os_release;
     my $runtime = "docker";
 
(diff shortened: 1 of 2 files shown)


pdostal:
Wow! 🎉 🚀

pdostal (approved):

jlausuch:
I think it's a good improvement, however I think some people might not like it (we are sharing these tests with the community), since this moves the console output from the OpenQA clickable frames to the serial_terminal.txt which is sometimes difficult to find things. 
In any case, the time efficiency we gain with this change makes it worth it.

pdostal:
I don't think it moves it out of 'frames' - it just moves it out of image frames to text frames 😄

jlausuch:
> I don't think it moves it out of 'frames' - it just moves it out of image frames to text frames 😄

Yes, and I am perfectly fine with that option, I just hope it's fine with everyone else using these tests :)

PR #13117, opened 2021-08-23
Improve default installation using libyui-rest-api

jknphy:
Related ticket: [poo#97973](https://progress.opensuse.org/issues/97973)
Verification run: [SLE](https://openqa.suse.de/tests/overview?version=15-SP4&build=jknphy%2Fos-autoinst-distri-opensuse%23improve_default_installation&distri=sle) | [openSUSE](https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=jknphy%2Fos-autoinst-distri-opensuse%23improve_default_installation&distri=opensuse)


tests/installation/registration/register_via_scc.pm
```diff
@@ -7,19 +7,21 @@
...
-# Summary: Register the system with scc in the installer with registration code.
+# Summary: Register the system via SCC in the installer with registration
+# code, optional email and enabling update repositories.
 #
 # Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
 
 use base 'y2_installbase';
 use strict;
 use warnings;
-use testapi 'get_var';
```
JRivrain, line 17: I use to specify just the functions I need in this case, I suppose it gives a tiny bit of memory optimization by loading in memory only what we need. Though this is just an assumption, I could not find a doc to confirm it, and I guess we would barely notice the difference on modern hardware anyway.

jknphy: It is so basic library, that once you use some function you might need some other when you need to fix this test, so it is more comfortable in this case to have the general one for maintenance and debugging.

tests/installation/registration/register_via_scc.pm
```diff
@@ -7,19 +7,21 @@
...
-    my $reg_code = get_var('SCC_REGCODE');
-    $testapi::distri->get_registration()->register_product_with_regcode($reg_code);
-    $testapi::distri->get_registration()->enable_update_repositories('The registration server offers update repos.*');
+    $testapi::distri->get_registration()->register_via_scc({
+            email    => get_var('SCC_EMAIL'),
+            reg_code => get_var('SCC_REGCODE')});
+    $testapi::distri->get_registration()->enable_update_repositories();
 }
 
 1;
```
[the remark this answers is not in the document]
jknphy: suseconnect is a name of a tool/script, and register_via_scc is one of the ways you can use the screen in the installer, register_system doesn't have that info, that there are different methods of registration during installation.

jknphy:
> Very nice, though I don't understand why to rename modules_selection to select_module_desktop ? I guess we never select anything else ?

Yes, I didn't figure out the whole thing, I mean all the cases in our job group, but it is clear that in many places we just enable the desktop module, so we can just have an atomic one for doing it and it is clear what we do without looking at the code. Also avoid to create test data "for now".

sofiasyria:
LGTM, just a couple of renaming cases, I preferred the old ones as they describe better. The module validate_beta_popup and the function select_product instead of install_product.

jknphy:
> LGTM, just a couple of renaming cases, I preferred the old ones as they describe better. The module validate_beta_popup and the function select_product instead of install_product.

Regarding `select_product` looks like we don't really proceed with the installation, just select something in the screen and perhaps there is another action or a next, `select` is a verb more for the UI IMO, calling `install_product` the navigation is implicit.
In the case of validate_beta_popup I think you are right, that still looks like a validation and I missed to apply the same philosophy. I just did in another commit, so we don't 'validate' anything there, and we do some action more functional related, if we would like to avoid to rewrite test we need to be closer to the functionality I believe.

PR #13120, opened 2021-08-23
Fix podman pull network issues in rootless_podman

pdostal:
- Error: https://openqa.suse.de/tests/6920228#step/rootless_podman/60
- Verification run: http://pdostal-server.suse.cz/tests/12462#

### DIFF
--- a/tests/containers/rootless_podman.pm
+++ b/tests/containers/rootless_podman.pm
@@ -25,6 +25,7 @@ use utils;
 use containers::common;
 use containers::container_images;
 use containers::urls 'get_suse_container_urls';
+use containers::utils 'registry_url';
 use version_utils qw(get_os_release);
 use version_utils 'is_sle';
 
@@ -63,7 +64,8 @@ sub run {
 }
 
 sub softfail_and_skip_on_bsc1182874 {
-    if (script_run("podman run alpine", timeout => 120) != 0) {
+    my $alpine = registry_url('alpine');
+    if (script_run("podman run $alpine", timeout => 180) != 0) {
         record_soft_failure "bsc#1182874 - container fails to run in rootless mode";
         return 1;
     }


tests/containers/rootless_podman.pm
```diff
@@ -63,7 +64,9 @@ sub run {
 }
 
 sub softfail_and_skip_on_bsc1182874 {
-    if (script_run("podman run alpine", timeout => 120) != 0) {
+    my $alpine = registry_url('alpine');
+    script_retry("podman pull $alpine", delay => 15, retry => 6);
```
b10n1k, line 68: IMO i do not know why do we need this exactly. 
- adds delay. alpine is really small anyway
- reduce coverage of the `run` capabilities, especially in `engines_and_tools`
- using our private *registry* the most of the problems should be resolved regarding pulling. Otherwise we should look at the registry itself.

Again i am not sure what/or how much improvement will be. However if you have concerns i would recommend to run a few stress tests without this line and see it you will find any.

tests/containers/rootless_podman.pm
```diff
@@ -63,7 +64,9 @@ sub run {
...
+    if (script_run("podman run $alpine", timeout => 120) != 0) {
```
b10n1k, line 69: in case you go on with the above line, does the `timeout` still make sense here?

tests/containers/rootless_podman.pm
pdostal: > * adds delay. alpine is really small anyway
To see my 'concern' please see the error linked in the description.

> * reduce coverage of the `run` capabilities, especially in `engines_and_tools`
I understand that. It is questionable.

> * using our private _registry_ the most of the problems should be resolved regarding pulling. Otherwise we should look at the registry itself.
Yes, this will already improve the situation.

Let's talk about this tomorrow.

tests/containers/rootless_podman.pm
pdostal: Maybe not, maybe something lower like 60?

tests/containers/rootless_podman.pm
grisu48: The problem is not the image size, the problem are often network infra problems. Therefore I'd also increase the delay actually to at least 30 seconds, otherwise we risk to see failures related to the same problem multiple times.

```suggestion
    script_retry("podman pull $alpine", delay => 30, retry => 3);
```

WDYT?

grisu48 (approved):
LGTM

tests/containers/rootless_podman.pm
b10n1k: its up to your judge. not a big deal IMO. i am leaving on you.

tests/containers/rootless_podman.pm
b10n1k: What i suggest is of making sure that the images pulls from `18.156.2.117:5000`. and if this doesnt improve things then try something (maybe `script_retry`) in another step.

pdostal: Let's go with Yianni's suggestion :smile: Please review again.

b10n1k (approved):
+1

tests/containers/rootless_podman.pm
b10n1k: diki

pdostal:
Verification run succeeded.

PR #13123, opened 2021-08-24
Sle micro containers fix

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/97292


jlausuch:
VR: https://openqa.suse.de/tests/6928283
failure is due to something else, but the booting problem is solved.

PR #13124, opened 2021-08-24
Public Cloud: Add debug output and wait longer when entering tunneled console

pdostal:
The `test` command has now longer timeout so the `ssh` has some time to establish.
The `ssh` has the verbose (`-v` flag) so we get more details when establishing the console.

- Related error: https://openqa.suse.de/tests/6924828#step/ssh_interactive_start/14
- Verification run: http://pdostal-server.suse.cz/tests/12461#step/ssh_interactive_start/13

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -839,7 +839,8 @@ sub activate_console {
     }
     if (get_var('TUNNELED') && $name !~ /tunnel/) {
         die "Console '$console' activated in TUNNEL mode activated but tunnel(s) are not yet initialized, use the 'tunnel' console and call 'setup_ssh_tunnels' first" unless get_var('_SSH_TUNNELS_INITIALIZED');
-        $self->script_run('ssh -t sut', 0);
+        # The verbose output is visible only at the tunnel-console - it doesn't interfere with tests as it isn't piped to /dev/sshserial
+        $self->script_run('ssh -vt sut', 0);
         ensure_user($user) unless (get_var('PUBLIC_CLOUD'));
     }
     set_var('CONSOLE_JUST_ACTIVATED', 1);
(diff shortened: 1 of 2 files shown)


lib/susedistribution.pm
```diff
@@ -839,7 +839,7 @@ sub activate_console {
     }
     if (get_var('TUNNELED') && $name !~ /tunnel/) {
         die "Console '$console' activated in TUNNEL mode activated but tunnel(s) are not yet initialized, use the 'tunnel' console and call 'setup_ssh_tunnels' first" unless get_var('_SSH_TUNNELS_INITIALIZED');
-        $self->script_run('ssh -t sut', 0);
+        $self->script_run('ssh -vt sut', 0);
```
grisu48, line 842: Won't this interfere with a lot of test cases, where the additional debug output might be unexpected and causes calls like `validate_script_output` ecc to fail in the future?

pdostal: No, one verbose flag won't produce any output when the session is established.
If I'd use it multiple times then the amount of output could start to interfere.

At least that's the conclusion of my research :smile:

cfconrad: I agree, if we have connection problems, then we might get something. If we do not have ssh problems during the test, there should not be more messages in the meantime.

If it will be different, we could try using `-E <logfile>` and upload the log

cfconrad (approved):

lib/susedistribution.pm
grisu48: The actual output of the ssh tunnel goes via the fifo in `/dev/sshserial`, so the debug output here should not have any functional implications on the running test cases.

Pavel's suggestion is IMHO valid and should not interfere with our test runs. LGTM!

grisu48 (approved):
LGTM :rocket:

lib/susedistribution.pm
pdostal: We use mostly serial terminal. This will stay in the VNC.

pdostal:
OK, I was wrong: https://openqa.suse.de/tests/6927034#step/command_not_found/7

I will try to use `-E` as Clemens suggested in another pull request.

PR #13128, opened 2021-08-25
Create br0 after host installation and fix network interface names

Julie-CAO:
- create br0 just after host installation as it is not default installed
- support opensuse as guest name
- allow more network interface names with udev naming scheme in virtual network tests

- Related ticket: https://progress.opensuse.org/issues/96374
- Verification run: 
[TW kvm](http://10.67.129.51/tests/2607)
[TW xen](http://10.67.129.51/tests/2608)
[SLE kvm](https://openqa.nue.suse.com/tests/6935263)

@alice-suse @xguo @waynechen55 @nanzhg   Could you help review?

### DIFF
--- a/tests/virt_autotest/libvirt_virtual_network_init.pm
+++ b/tests/virt_autotest/libvirt_virtual_network_init.pm
@@ -84,7 +84,14 @@ sub run_test {
         save_guest_ip($guest, name => "br123");
         virt_autotest::utils::ssh_copy_id($guest);
         #Prepare the new guest network interface files for libvirt virtual network
-        assert_script_run("ssh root\@$guest 'cd /etc/sysconfig/network/; cp ifcfg-eth0 ifcfg-eth1; cp ifcfg-eth0 ifcfg-eth2; cp ifcfg-eth0 ifcfg-eth3; cp ifcfg-eth0 ifcfg-eth4; cp ifcfg-eth0 ifcfg-eth5; cp ifcfg-eth0 ifcfg-eth6'");
+        #for some guests, interfaces are named eth0, eth1, eth2, ...
+        #for TW kvm guest, they are enp1s0, enp2s0, enp3s0, ...
+        my $primary_nic = script_output("ssh root\@$guest \"ip a|awk -F': ' '/state UP/ {print \\\$2}'|head -n1\"");
+        $primary_nic =~ /([a-zA-Z]*)(\d)(\w*)/;
+        for (my $i = 1; $i <= 6; $i++) {
+            my $nic = $1 . (int($2) + $i) . $3;
+            assert_script_run("ssh root\@$guest 'cp /etc/sysconfig/network/ifcfg-$primary_nic /etc/sysconfig/network/ifcfg-$nic'");
+        }
         #enable guest wickedd debugging
         assert_script_run "ssh root\@$guest \"sed -i 's/^WICKED_DEBUG=.*/WICKED_DEBUG=\"all\"/g' /etc/sysconfig/network/config\"";
         assert_script_run "ssh root\@$guest 'grep 'WICKED_DEBUG' /etc/sysconfig/network/config'";
--- a/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
+++ b/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
@@ -50,9 +50,21 @@ sub run {
     zypper_call('--gpg-auto-import-keys ref');
     zypper_call("in -t pattern ${hypervisor}_server ${hypervisor}_tools", 1800);
     set_serial_console_on_vh('', '', $hypervisor);
+    systemctl 'enable libvirtd', ignore_failure => 1;
 
     virt_autotest::utils::install_default_packages();
     save_screenshot;
+
+    #set up necessary network bridge br0 for kvm/xen hypervisor.
+    #SLE sets up br0 during installation as kvm/xen role, while openSUSE TW does not.
+    #change MACAddressPolicy to none, or the network bridge will get a different MAC address from its enslave
+    #it is specific to TW because it uses udev naming scheme v247, see bsc#1136600
+    my $bridge_rule_file = "/usr/lib/systemd/network/98-default-bridge.link";
+    assert_script_run("echo \"[Match]\nDriver=bridge\n\n[Link]\nMACAddressPolicy=none\" > $bridge_rule_file", 60);
+    my $default_netdevice = get_var('SUT_NETDEVICE', 'eno1');
+    script_run("echo \"BOOTPROTO='dhcp'\nSTARTMODE='auto'\nBRIDGE='yes'\nBRIDGE_PORTS='$default_netdevice'\" > /etc/sysconfig/network/ifcfg-br0");
+    script_run("echo \"BOOTPROTO='none'\nSTARTMODE='auto'\" > /etc/sysconfig/network/ifcfg-$default_netdevice");
+
 }
 
 sub test_flags { {fatal => 1} }
--- a/lib/virt_autotest/virtual_network_utils.pm
+++ b/lib/virt_autotest/virtual_network_utils.pm
@@ -254,7 +254,7 @@ sub restart_network {
 
 sub restore_guests {
     return if get_var('INCIDENT_ID');    # QAM does not recreate guests every time
-    my $get_vm_hostnames   = "virsh list  --all | grep sles | awk \'{print \$2}\'";
+    my $get_vm_hostnames   = "virsh list --all | grep -e sles -e opensuse | awk \'{print \$2}\'";
     my $vm_hostnames       = script_output($get_vm_hostnames, 30, type_command => 0, proceed_on_failure => 0);
     my @vm_hostnames_array = split(/\n+/, $vm_hostnames);
     foreach (@vm_hostnames_array)
--- a/tests/virt_autotest/virt_utils.pm
+++ b/tests/virt_autotest/virt_utils.pm
@@ -635,7 +635,7 @@ sub check_guest_disk_type {
 sub recreate_guests {
     my $based_guest_dir = shift;
     return if get_var('INCIDENT_ID');    # QAM does not recreate guests every time
-    my $get_vm_hostnames   = "virsh list  --all | grep sles | awk \'{print \$2}\'";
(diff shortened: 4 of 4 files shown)


tests/virt_autotest/libvirt_virtual_network_init.pm
```diff
@@ -84,7 +84,14 @@ sub run_test {
         save_guest_ip($guest, name => "br123");
         virt_autotest::utils::ssh_copy_id($guest);
         #Prepare the new guest network interface files for libvirt virtual network
-        assert_script_run("ssh root\@$guest 'cd /etc/sysconfig/network/; cp ifcfg-eth0 ifcfg-eth1; cp ifcfg-eth0 ifcfg-eth2; cp ifcfg-eth0 ifcfg-eth3; cp ifcfg-eth0 ifcfg-eth4; cp ifcfg-eth0 ifcfg-eth5; cp ifcfg-eth0 ifcfg-eth6'");
+        #for some guests, interfaces are named eth0, eth1, eth2, ...
+        #for TW kvm guest, they are enp1s0, enp2s0, enp3s0, ...
+        my $primary_nic = script_output("ssh root\@$guest \"ip a|awk -F': ' '/state UP/ {print \\\$2}'|head -n1\"");
+        $primary_nic =~ /([a-zA-Z]*)(\d)(\w*)/;
```
waynechen55, line 90: Why this regex matching is needed ?  And it seems that it actually does not match anything ? It just searches some characters and numbers globally.

tests/virt_autotest/libvirt_virtual_network_init.pm
```diff
@@ -84,7 +84,14 @@ sub run_test {
...
+        for (my $i = 1; $i <= 6; $i++) {
+            my $nic = $1 . (int($2) + $i) . $3;
+            assert_script_run("ssh root\@$guest 'cd /etc/sysconfig/network/; cp ifcfg-$primary_nic ifcfg-$nic'");
+        }
```
waynechen55, line 94: You are creating new interfaces from $primary_nic ?  What if $3 empty ?

waynechen55: What about "cd ~" at the end ?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -50,9 +50,21 @@ sub run {
...
 
     virt_autotest::utils::install_default_packages();
     save_screenshot;
+
+    #set up necessary network bridge br0 for virtualization tests.
+    #SLE sets up br0 during installation as kvm/xen role, while openSUSE TW does not.
+    #change MACAddressPolicy to none, or the network bridge will get a different MAC address from its enslave
+    #it is specific to TW because it uses udev naming scheme v247, see bsc#1136600
+    my $bridge_rule_file = "/usr/lib/systemd/network/98-default-bridge.link";
+    assert_script_run("echo \"[Match]\nDriver=bridge\n\n[Link]\nMACAddressPolicy=none\" > $bridge_rule_file", 60);
```
waynechen55, line 63: Be cautious that you overwrite $bridge_rule_file

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -50,9 +50,21 @@ sub run {
...
+    my $default_netdevice = get_var('SUT_NETDEVICE', 'eno1');
```
waynechen55, line 64: Why en01 is the last default value ?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -50,9 +50,21 @@ sub run {
...
+    script_run("echo \"BOOTPROTO='dhcp'\nSTARTMODE='auto'\nBRIDGE='yes'\nBRIDGE_PORTS='$default_netdevice'\" > /etc/sysconfig/network/ifcfg-br0");
+    script_run("echo \"BOOTPROTO='none'\nSTARTMODE='auto'\" > /etc/sysconfig/network/ifcfg-$default_netdevice");
+
```
waynechen55, line 67: What about a new function like create_network_interface ? It will responsible for filling out interface config file, and take arguments like config file, interface name, those parameters in config file.

tests/virt_autotest/libvirt_virtual_network_init.pm
Julie-CAO: > You are creating new interfaces from $primary_nic ? What if $3 empty ?

fg. $primary_nic='eth0',  $3 is empty,  the interfaces will be eth1, eth2, eth3, ...., eth6. There is a verification run with eth0: SLE kvm

Julie-CAO: > What about "cd ~" at the end ?

Yes, better not to change the current work directory.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: 98-default-bridge.link is created as a new file. The existing one is 99-default.link. 98-default-bridge.link is loaded ahead to override the rules for 'bridges', it is deliberate as we'd like the MAC of the bridge is consistent to its enslave network device the way in SLE.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: 'eno1' in TW is like 'em1' in SLE. It happens that all my three test machines are using 'on board network card'.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: There are only two lines and a pile of parameters. SLE doesn't have this requirement yet. Don't bother. If there is a similar demand in the future, let's consider it again.

tests/virt_autotest/libvirt_virtual_network_init.pm
alice-suse: @Julie-CAO Would you please explain the purpose of the for loop? Are you creating 6 new ifcfg-xx files? What are they for?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -50,9 +50,21 @@ sub run {
     zypper_call('--gpg-auto-import-keys ref');
     zypper_call("in -t pattern ${hypervisor}_server ${hypervisor}_tools", 1800);
     set_serial_console_on_vh('', '', $hypervisor);
+    systemctl 'enable libvirtd', ignore_failure => 1;
 
     virt_autotest::utils::install_default_packages();
     save_screenshot;
+
+    #set up necessary network bridge br0 for virtualization tests.
```
alice-suse, line 58: This file goes to contain more and more minor functions that can not be indicated from the file name. May need to consider changing the file name at some time.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
alice-suse: Why are different nics here and in above file? eno1 vs   enp1s0? Do they both exist on tw host, or either one?

tests/virt_autotest/libvirt_virtual_network_init.pm
Julie-CAO: yes, I created 6 new ifcfg-xx files. That's not what I did initially, I just adapted the test scenarios to TW from SLE. @guoxuguang , could you give more input about why we created 6 ifcfg-xx files in virtual network tests as you discussed with Pavel?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: 'eno1' is default interface for TW host, 'enp1s0' is for TW guest in TW kvm host. 'eth0' is for TW guest in TW XEN host.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: br0 is still one of the necessary part of installing hypervisor. it is created by default in SLE when kvm/xen role is chosen during installation. while in TW, you have to zypper in kvm/xen, then set up br0 manually. the procedure is still conforms to the file name yet.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
waynechen55: What a new rule file like 98-default-bridge.link is needed ? You can just modify the existing one ?

Julie-CAO: I just follow the recommendation in https://bugzilla.opensuse.org/show_bug.cgi?id=1136600#c34. It is safer and easier.

tests/virt_autotest/libvirt_virtual_network_init.pm
waynechen55: What about wrap up this code snippet into a function ? create_network_interfaces_from_primary_nic or other name ? It takes the number of interfaces to be created as argument.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
waynechen55: So this module switch_to_ssh_and_install_hypervisor.pm is only for TW to use ?

Julie-CAO: Yes, it is specific to TW only as there is no kvm/xen role selection during installation.

tests/virt_autotest/libvirt_virtual_network_init.pm
Julie-CAO: To copy&paste numbers of network configuration files does not look like a decent way to implement the test, I prfer that this is a temporary and simple solution. So It is not necessary to make this immature code that is not needed elsewhere into a function.

PR #13129, opened 2021-08-25
Check for auth window in "handle_login" for S390x.

zhaocongjia:
The issue is that test fails at https://openqa.suse.de/tests/6934935#step/first_boot/10

It is expecting 'generic-desktop' but get 'gnome-activities' instead.

The reason is there was an extra polkit auth window, as you can see in the video at https://openqa.suse.de/tests/6934935/video?filename=video.ogv.

So the solution is to check for this extra window and close it.

This extra window is expected on S390x and other test platform if vncserver is providing the graphical session: https://bugzilla.suse.com/show_bug.cgi?id=1177446

- Related ticket: https://progress.opensuse.org/issues/97451
- Needles: None
- Verification run: https://openqa.suse.de/tests/6935288#step/first_boot/6 https://openqa.suse.de/tests/6935289#step/first_boot/8


lib/x11utils.pm
```diff
@@ -233,6 +233,14 @@ sub handle_login {
     assert_screen 'displaymanager-password-prompt';
     type_password($mypwd);
     send_key 'ret';
+    wait_still_screen;
+    # for S390x testing, since they are not using qemu built-in vnc, it is
+    # expected that polkit authentication window can open for first time login.
+    # see bsc#1177446 for more information.
+    if (check_screen(qw[authentication-required-user-settings authentication-required-modify-system])) {
```
GraceWang571, line 240: You need to give a timeout value to check_screen, otherwise, you will get errors
 `Argument "authentication-required-modify-system" isn't numeric in multiplication (*) at /usr/lib/os-autoinst/bmwqemu.pm line 324.`
See https://openqa.suse.de/tests/6935288/file/autoinst-log.txt

GraceWang571: Besides, have you tried the `handle_additional_polkit_windows` (https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/234b496213ed9bc282e0279362aa8d32ed85088c/lib/opensusebasetest.pm#L971) ?

zhaocongjia: Thanks, I'll change `check_screen`.

About `handle_additional_polkit_windows`, it will introduce extra softfailure, and it is too heavy, so not very suitable here.

PR #13130, opened 2021-08-25
RStudio: kill firefox if it hangs

dcermak:
This pull request does two things:
1. At the end of the RStudio Server tests we quit Firefox. However, Firefox will sometimes hang and won't quit, in that case we just `kill -9` it and be done with that.
2. fix a `assert_screen` to `assert_and_click` so that a the flamegraph tab is closed and the rstudio_server and rstudio_desktop tests can be run independently of each other (atm they have an implicit dependency due to the needles).

- Related ticket: https://progress.opensuse.org/issues/96593
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/743
- Verification run: https://openqa.opensuse.org/tests/1885829

### DIFF
--- a/tests/x11/rstudio_server.pm
+++ b/tests/x11/rstudio_server.pm
@@ -62,13 +62,17 @@ sub run() {
     # log out at last and close firefox
     assert_and_click("rstudio_server-sign-out");
     assert_screen("rstudio_server-login-username");
-    send_key('alt-f4');
 
-    # optionally click away the close all tabs window
-    assert_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)]);
-    if (match_has_tag('rstudio_server-firefox_quit-and-close-tabs')) {
-        click_lastmatch();
-        assert_screen('generic-desktop');
+    for my $i (1 .. 5) {
+        send_key('ctrl-w', wait_screen_change => 1);
+        last if defined(check_screen('generic-desktop'));
+    }
+
+    if (!defined(check_screen('generic-desktop'))) {
+        # firefox is stuck and just greys out but does not display the "close
+        # all tabs" dialog => kill it
+        record_info('Workaround', 'Firefox froze for 60s and does not want to close');
+        script_run('pkill -9 firefox');
     }
 }
 
--- a/lib/rstudio.pm
+++ b/lib/rstudio.pm
@@ -131,7 +131,7 @@ sub rstudio_run_profiler {
     type_string("profvis(expr = {sub.test1 <- system.time(for (i in 1:50000) {df[3, 2]})})");
     send_key("ret");
 
-    assert_screen("$prefix-profile_flame-graph_close");
+    assert_and_click("$prefix-profile_flame-graph_close");
 }
 
 sub rstudio_test_notebook {


tests/x11/rstudio_server.pm
```diff
@@ -62,13 +62,19 @@ sub run() {
...
-    assert_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)]);
+    check_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)], timeout => 60);
     if (match_has_tag('rstudio_server-firefox_quit-and-close-tabs')) {
+        # optionally click away the close all tabs window => FF should close
         click_lastmatch();
         assert_screen('generic-desktop');
+    } elsif (!match_has_tag('generic-desktop')) {
+        # firefox is stuck and just greys out but does not display the "close
+        # all tabs" dialog => kill it
+        script_run('pkill -9 firefox');
```
b10n1k, line 76: i dont know much about rstudio but this is what i am thinking. this sounds like a bug. So it would make sense to mark it with a softfail. 
Also, as it is now, 
- will check screen once 
- if doesnt find a match in 60s it will execute the `pkill`

what about something like
```
do {
        check_screen([qw(rstudio_server-firefox_quit-and-close-tabs rstudio_server-login-username generic-desktop)]);
        if (match_has_tag('rstudio_server-firefox_quit-and-close-tabs')) {
           click_lastmatch();
        }
        if (match_has_tag('rstudio_server-login-username')) {
              # if match means stalling
              #softfail or pkill or...
         }
    } while (not match_has_tag('generic-desktop'));
```

in additional i didnt find the `pkill` in the VR so i am not sure if this change actually works.

tests/x11/rstudio_server.pm
```diff
@@ -62,13 +62,19 @@ sub run() {
...
+        assert_screen('generic-desktop');
```
b10n1k, line 77: this seems it belong outside the conditional block. and can be removed from L72 too

dcermak: Indeed, thanks!

tests/x11/rstudio_server.pm
dcermak: I thought `check_screen` would check continuously in the same way that `assert_screen` does. Does it really work differently?

b10n1k: they are the same in this matter. `check_screen` will try to match within a timeout period and it will continue. 
But based on the documentation If you take a look
, `check_screen` is preferrable  for `for optional matches` and `is recommended to use the lowest possible timeout`.

what i suggested was to run continually `check_screen` with default timeout , so you will continue as soon as it matches while waiting for some time.

dcermak: I really don't see the difference between running `check_screen([...], timeout => 60)` and `check_screen([...])` in a loop. What am I missing here?

tests/x11/rstudio_server.pm
```diff
@@ -62,14 +62,20 @@ sub run() {
...
-    assert_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)]);
+    check_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)], timeout => 60);
     if (match_has_tag('rstudio_server-firefox_quit-and-close-tabs')) {
+        # optionally click away the close all tabs window => FF should close
         click_lastmatch();
-        assert_screen('generic-desktop');
+    } elsif (!match_has_tag('generic-desktop')) {
+        # firefox is stuck and just greys out but does not display the "close
+        # all tabs" dialog => kill it
+        record_soft_failure('Firefox froze for 60s and does not want to close');
```
okurz, line 75: have you seen https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13130/checks?check_run_id=3461849409#step:6:1769 ? soft-fails must reference valid product issues. If you don't have a bug report for that use

```suggestion
        record_info('Workaround', 'Firefox froze for 60s and does not want to close');
```

tests/x11/rstudio_server.pm
okurz: Only use `check_screen` if you want to check the return code. 

We also listen the preference for `assert_screen` in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style.

What the documentation means with "Use this for optional matches" is that you can use `check_screen` if there might be something on the screen and maybe not. But if there is something on the screen and it could be "foo" or "bar" use a multi-tag assert_screen (as in before). @b10n1k I am not sure if maybe you misunderstood what the documentation wants to say as well?

And, yes, a `check_screen 'foo', 60` can be equivalent to `for i in (1 .. 60) { last if check_screen 'foo', 60 }`. Both cases can waste 60s and should be avoided.

tests/x11/rstudio_server.pm
dcermak: Thanks! I've applied this fix, but I must confess that I wasn't able to figure out what the test suite was complaining about from the output.

dcermak:
I've rewritten this test now so that it no longer tries to alt-f4 firefox but instead closes one tab after the other until it ends up at the desktop. If it does not, then firefox is killed via `pkill -9`.

tests/x11/rstudio_server.pm
```diff
@@ -62,13 +62,17 @@ sub run() {
...
     assert_screen("rstudio_server-login-username");
-    send_key('alt-f4');
 
-    # optionally click away the close all tabs window
-    assert_screen([qw(rstudio_server-firefox_quit-and-close-tabs generic-desktop)]);
-    if (match_has_tag('rstudio_server-firefox_quit-and-close-tabs')) {
-        click_lastmatch();
-        assert_screen('generic-desktop');
+    for my $i (1 .. 5) {
+        send_key('ctrl-w', wait_screen_change => 1);
```
okurz, line 67: Instead of this with the wait screen I still suggest to use a multi-tag assert screen. Exit the loop as soon as you match the desktop or if the loop runs out

dcermak:
See #13153

PR #13132, opened 2021-08-25
Wireguard: Remove /boot/config- file check

pdostal:
- Related error: https://openqa.opensuse.org/tests/1885684#step/wireguard/16
- Verification run: http://pdostal-server.suse.cz/tests/12475

### DIFF
--- a/tests/network/wireguard.pm
+++ b/tests/network/wireguard.pm
@@ -75,7 +75,8 @@ sub run {
         assert_script_run 'modprobe wireguard';
     }
 
-    assert_script_run 'grep -i CONFIG_WIREGUARD /boot/config-$(uname -r)';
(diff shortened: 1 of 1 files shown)


tests/network/wireguard.pm
```diff
@@ -75,7 +75,6 @@ sub run {
         assert_script_run 'modprobe wireguard';
     }
```
grisu48, line 77: How about instead of completely removing this test, just skip the check if the file doesn't exists?

```suggestion
assert_script_run('grep -i CONFIG_WIREGUARD /boot/config-$(uname -r)') unless(script_run('stat /boot/config-$(uname -r)') != 0);
```

grisu48 (approved):
LGTM :rocket:

PR #13138, opened 2021-08-26
Replace encryption_volume_activate module with libuyi based

sofiasyria:
- Related ticket: https://progress.opensuse.org/issues/96809
- Verification run: 
http://falafel.suse.cz/tests/overview?version=15-SP4&build=26.1&distri=sle
https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=sofiasyria%2Fos-autoinst-distri-opensuse%23encrypt_oop


tests/installation/encrypted_volume/cancel_encrypted_volume.pm
```diff
@@ -0,0 +1,29 @@
...
+# Summary: Test module to cancel encrypted volume activation.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    my $encrypted_volume = $testapi::distri->get_encrypted_volume();
```
jknphy, line 25: let's use here our common approach to not include `use testapi` as it is not needed when we use variable `$testapi` and also make this one line: `$testapi::distri->get_encrypted_volume_activation()->do_not_activate()`.

lib/Distribution/Sle/15_current.pm
```diff
@@ -24,6 +24,7 @@ use Installation::License::Sle::LicenseAgreementController;
 use Installation::License::Sle::Firstboot::LicenseAgreementController;
 use Installation::ProductSelection::ProductSelectionController;
 use Installation::Registration::RegistrationController;
+use Installation::EncryptedVolume::EncryptedVolumeController;
```
jknphy, line 27: this is not SLE specific, even if we don't have a job there (don't we, right?) we should add it for TW distri.

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
```diff
@@ -0,0 +1,53 @@
...
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved. This file is offered as-is,
+# without any warranty.
+
+# Summary: Controller for YaST Encrypted Volume module.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::EncryptedVolume::EncryptedVolumeController;
```
jknphy, line 13: just a suggestion here: s/EncryptedVolumeController/EncryptedVolumeActivationController/
The activation part is really the key of the logic of this screen, at least as minimum, let's add it to the Summary.

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
```diff
@@ -0,0 +1,53 @@
...
+use strict;
+use warnings;
+use YuiRestClient;
+use Installation::EncryptedVolume::EncryptedVolumePage;
```
jknphy, line 17: IMO it should be `Installation::SystemProbing::EncryptedVolumeActivationPage` (with SystemProbing as folder) as this screen could contains other popups and it is clear by the name that belongs to that phase of system probing by YaST.

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
```diff
@@ -0,0 +1,53 @@
...
+use testapi;
```
jknphy, line 18: not needed.

schedule/yast/encryption/activate_encrypted_volume+force_recompute.yaml
```diff
@@ -17,7 +16,7 @@ schedule:
   - installation/product_selection/select_product
   - installation/licensing/accept_license
   - installation/scc_registration
-  - installation/encrypted_volume_activation
+  - installation/encrypted_volume/activate_encrypted_volume
```
jknphy, line 19: similar here, I think the intermediate folder should be /system_probing/

tests/installation/encrypted_volume/activate_encrypted_volume.pm
```diff
@@ -0,0 +1,30 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    my $encrypted_volume = $testapi::distri->get_encrypted_volume();
+    $encrypted_volume->enter_volume_encryption_password($password);
```
jknphy, line 26: this is fine at the level of the page that we enter something and we accept it but from the point of view of the test should look more like one single operation: `$testapi::distri->get_encrypted_volume_activation()->activate($password)`

lib/Distribution/Sle/15_current.pm
sofiasyria: Wouldn't it make more sense to add it there once we add a relevant job for TW? With the same logic, should I add it for Leap too?

tests/installation/encrypted_volume/cancel_encrypted_volume.pm
sofiasyria: "use testapi" is needed for the $password. I know you like patterns and mathematics and tidy details and everything, but in this case, if we apply a strict pattern everywhere it would only make sense to use a similar approach as I am using because in a module that will call more than 2 commands, it wouldn't make sense to have this short of repetitions. I will not change it because it helps readability of the code a lot.

lib/Distribution/Sle/15_current.pm
jknphy: TW in our distribution pattern is the default, we put in the other products when there is some exception or overwrite. As this logic is not specific (let's assume that the screen doesn't change much there) should go to Factory (TW) let's say and we can use it without changes when there will be some openSUSE job. So, in other words we should not put it in SLE or Leap, and put this code directly in TW.

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
sofiasyria: Done

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
sofiasyria: Done

lib/Installation/EncryptedVolume/EncryptedVolumeController.pm
sofiasyria: Done

tests/installation/encrypted_volume/cancel_encrypted_volume.pm
jknphy: you are not using $password for this test module and the library is not needed to be loaded, we can save that.
The less variable you declare in Perl, the better, is kind of the rule of the thumb, it is a very readable line, is not even long the one proposed, the plan is not touch the tests, so the idea is not to think about needing more than one line here, but let's leave it as it is.

lib/Distribution/Sle/15_current.pm
sofiasyria: I didn't know. Thanks for the info. Done.

tests/installation/encrypted_volume/cancel_encrypted_volume.pm
sofiasyria: Done

JRivrain:
LGTM

PR #12698, opened 2021-06-11
Containers: use untested or official container images accordingly

jlausuch:
There are 2 scenarios:
1) SUT=Host updates
   Here we should test the host container features and use official
   stable images for that purpose.
   These tests are triggered every day as part of maintenance updates.

2) SUT=container image updates
   Here we are targetting the images that are not released yet, so
   we use an official released HOST to test the untested images.
   For these tests, we have the variable CONTAINER_IMAGE_TYPE=untested
   These tests will be triggered whenever there is a new container image
   build in :/Update:/CR:/ToTest/images/

The PR is putting this into practice and using the right arrays depending
on the situation.

[type description here, PLEASE, REMOVE THIS LINE, PLACEHOLDER, BEFORE SUBMITTING THIS PULL REQUEST]

- Related ticket: https://progress.opensuse.org/issues/93480

PR #13141, opened 2021-08-27
Refactor get_suse_container_urls using data structure for each distri

b10n1k:
I have approach this problem organizing the data into a hash. The data
structure can hold information explicitly, so it is more ease to handle.

Each <distri><version> hold three information
- released_registry
- totest_registry
- available_arch

I use anonumous functions for `registries` to provide flexibility. For instance
you can choose the different registry based on aarch. or it is possible to
modify the sub to return an array in the future.

I tried to group some of the versions as they provide commonalities but regular
hash accept scalar as a key.

- Related ticket: https://progress.opensuse.org/issues/97232
Verification run:
- https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313141&distri=sle&version=15-SP4
- https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=b10n1k%2Fos-autoinst-distri-opensuse%2313141
- https://openqa.suse.de/tests/overview?version=5.1&distri=sle-micro&build=b10n1k%2Fos-autoinst-distri-opensuse%2313141
- https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313141&version=5.0&distri=sle-micro
- https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.3&build=b10n1k%2Fos-autoinst-distri-opensuse%2313141
- https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2313141&distri=microos
- https://openqa.opensuse.org/tests/overview?version=15.2&distri=opensuse&build=b10n1k%2Fos-autoinst-distri-opensuse%2313141

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -52,7 +52,7 @@ sub get_opensuse_registry_prefix {
     }
 }
 
-our %registries = (
+our %images_uri = (
     sle => {
         '12-SP3' => {
             released => sub { 'registry.suse.com/suse/sles12sp3' },
@@ -234,7 +234,7 @@ our %registries = (
 
 sub supports_image_arch {
     my ($distri, $version, $arch) = @_;
-    (grep { $_ eq $arch } @{$registries{$distri}{$version}{available_arch}}) ? 1 : 0;
+    (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
 
 # Returns a tuple of image urls and their matching released "stable" counterpart.
@@ -250,8 +250,8 @@ sub get_suse_container_urls {
     my @released_images = ();
 
     if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
-        push @untested_images, $registries{$args{distri}}{$args{version}}{totest}->($args{arch});
-        push @released_images, $registries{$args{distri}}{$args{version}}{released}->($args{arch});
+        push @untested_images, $images_uri{$args{distri}}{$args{version}}{totest}->($args{arch});
+        push @released_images, $images_uri{$args{distri}}{$args{version}}{released}->($args{arch});
     } else {
         die("Unknown combination of distro/arch.");
     }
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright © 2018 SUSE LLC
+# Copyright © 2018-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
@@ -8,7 +8,7 @@
 # without any warranty.
 
 # Summary: Database for URLs of container images to be tested
-# Maintainer: Fabian Vogt <fvogt@suse.com>
+# Maintainer: Fabian Vogt <fvogt@suse.com>, qa-c team <qa-c@suse.de>
 
 package containers::urls;
 
@@ -52,64 +52,207 @@ sub get_opensuse_registry_prefix {
     }
 }
 
+our %registries = (
+    sle => {
+        '12-SP3' => {
+            released => sub { 'registry.suse.com/suse/sles12sp3' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-12-sp3/docker/update/cr/totest/images/suse/sles12sp3';
+            },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
+        },
+        '12-SP4' => {
+            released => sub { 'registry.suse.com/suse/sles12sp4' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-12-sp4/docker/update/cr/totest/images/suse/sles12sp4';
+            },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
+        },
+        '12-SP5' => {
+            released => sub { 'registry.suse.com/suse/sles12sp5' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-12-sp5/docker/update/cr/totest/images/suse/sles12sp5';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15' => {
+            released => sub { 'registry.suse.com/suse/sle15:15.0' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15/update/cr/totest/images/suse/sle15:15.0';
+            },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
+        },
+        '15-SP1' => {
+            released => sub { 'registry.suse.com/suse/sle15:15.1' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp1/update/cr/totest/images/suse/sle15:15.1';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP2' => {
+            released => sub { 'registry.suse.com/suse/sle15:15.2' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp2/update/cr/totest/images/suse/sle15:15.2';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP3' => {
+            released => sub { 'registry.suse.com/suse/sle15:15.3' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP4' => {
+            released => sub { },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15:15.4 ';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        }
+    },
+    opensuse => {
+        Tumbleweed => {
+            released => sub { 'registry.opensuse.org/opensuse/tumbleweed' },
+            totest   => sub {
+                'registry.opensuse.org/' . get_opensuse_registry_prefix . 'opensuse/tumbleweed';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
+        },
+        '15.0' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.0' },
+            totest   => sub {
+                'registry.opensuse.org/opensuse/leap/15.0/images/totest/images/opensuse/leap:15.0';
+            },
+            available_arch => ['x86_64']
+        },
+        '15.1' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.1' },
+            totest   => sub {
+                my $arch = shift;
+                if ($arch eq 'x86_64') {
+                    'registry.opensuse.org/opensuse/leap/15.1/images/totest/containers/opensuse/leap:15.1';
+                } elsif ($arch eq 'arm') {
+                    'registry.opensuse.org/opensuse/leap/15.1/arm/images/totest/containers/opensuse/leap:15.1';
+                }
+            },
+            available_arch => ['x86_64', 'arm']
+        },
+        '15.2' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.2' },
+            totest   => sub {
+                my $arch = shift;
+                if ($arch eq 'x86_64') {
+                    'registry.opensuse.org/opensuse/leap/15.2/images/totest/containers/opensuse/leap:15.2';
+                } elsif ($arch eq 'arm') {
+                    'registry.opensuse.org/opensuse/leap/15.2/arm/images/totest/containers/opensuse/leap:15.2';
+                }
+            },
+            available_arch => ['x86_64', 'arm']
+        },
+        '15.3' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.3' },
+            totest   => sub {
+                my $arch = shift;
+                if (grep { $_ eq $arch } qw/x86_64 aarch64 ppc64le s390x/) {
+                    'registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3';
+                } elsif ($arch eq 'arm') {
+                    'registry.opensuse.org/opensuse/leap/15.3/arm/images/totest/containers/opensuse/leap:15.3';
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
+        }
+    },
+    'sle-micro' => {
+        '15' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.0' },
+            totest         => sub { },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
+        },
+        '15-SP1' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.1' },
+            totest         => sub { },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP2' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.2' },
+            totest         => sub { },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP3' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.3' },
+            totest         => sub { },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        }
+    },
+    microos => {
+        Tumbleweed => {
+            released => sub { 'registry.opensuse.org/opensuse/tumbleweed' },
+            totest   => sub {
+                'registry.opensuse.org/' . get_opensuse_registry_prefix . 'opensuse/tumbleweed';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
+        },
+        '15.1' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.1' },
+            totest   => sub {
+                my $arch = shift;
+                if ($arch eq 'x86_64') {
+                    'registry.opensuse.org/opensuse/leap/15.1/images/totest/containers/opensuse/leap:15.1';
+                } elsif (grep { $_ eq $arch } qw/aarch64 arm/) {
+                    'registry.opensuse.org/opensuse/leap/15.1/arm/images/totest/containers/opensuse/leap:15.1';
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        },
+        '15.2' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.2' },
+            totest   => sub {
+                my $arch = shift;
+                if ($arch eq 'x86_64') {
+                    'registry.opensuse.org/opensuse/leap/15.2/images/totest/containers/opensuse/leap:15.2';
+                } elsif (grep { $_ eq $arch } qw/aarch64 arm/) {
+                    'registry.opensuse.org/opensuse/leap/15.2/arm/images/totest/containers/opensuse/leap:15.2';
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        },
+        '15.3' => {
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.3' },
+            totest   => sub {
+                my $arch = shift;
+                if ($arch eq 'x86_64') {
+                    'registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3';
+                } elsif (grep { $_ eq $arch } qw/aarch64 arm/) {
+                    'registry.opensuse.org/opensuse/leap/15.3/arm/images/totest/containers/opensuse/leap:15.3';
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        }
+    }
+);
+
+sub supports_image_arch {
+    my ($distri, $version, $arch) = @_;
+    (grep { $_ eq $arch } @{$registries{$distri}{$version}{available_arch}}) ? 1 : 0;
+}
+
 # Returns a tuple of image urls and their matching released "stable" counterpart.
 # If empty, no images available.
 sub get_suse_container_urls {
-    my $version    = shift // get_required_var('VERSION');
-    my $dotversion = $version =~ s/-SP/./r;                    # 15 -> 15, 15-SP1 -> 15.1
-    $dotversion = "${dotversion}.0" if $dotversion !~ /\./;    # 15 -> 15.0
-
+    my %args = (
+        version => get_required_var('VERSION'),
+        arch    => get_required_var('ARCH'),
+        distri  => get_required_var('DISTRI'),
+        @_
+    );
     my @untested_images = ();
     my @released_images = ();
-    if (is_sle(">=12-sp3", $version) && is_sle('<15', $version)) {
-        my $lowerversion  = lc $version;
-        my $nodashversion = $version =~ s/-sp/sp/ir;
-        # No aarch64 image
-        if (!is_aarch64) {
-            push @untested_images, "registry.suse.de/suse/sle-${lowerversion}/docker/update/cr/totest/images/suse/sles${nodashversion}";
-            push @released_images, "registry.suse.com/suse/sles${nodashversion}";
-        }
-    }
-    elsif (is_sle(">=15", $version) && is_released) {
-        my $lowerversion = lc $version;
-        # Location for maintenance builds
-        push @untested_images, "registry.suse.de/suse/sle-${lowerversion}/update/cr/totest/images/suse/sle15:${dotversion}";
-        push @released_images, "registry.suse.com/suse/sle15:${dotversion}";
-    }
-    elsif (is_sle(">=15-sp4", $version)) {
-        my $lowerversion = lc $version;
-        # Location for GA builds
-        push @untested_images, "registry.suse.de/suse/sle-${lowerversion}/ga/test/images/suse/sle15:${dotversion}";
-        push @released_images, "registry.suse.com/suse/sle15:${dotversion}";
-    }
-    elsif (is_sle_micro) {
-        # Untested images are not validated in SLE Micro, so leave it empty
-        push @released_images, "registry.suse.com/suse/sle15:${dotversion}";
-    }
-    elsif (is_tumbleweed || is_microos("Tumbleweed")) {
-        push @untested_images, "registry.opensuse.org/" . get_opensuse_registry_prefix . "opensuse/tumbleweed";
-        push @released_images, "registry.opensuse.org/opensuse/tumbleweed";
-    }
-    elsif (is_leap(">=15.3")) {
-        # All archs in the same location
-        push @untested_images, "registry.opensuse.org/opensuse/leap/${version}/images/totest/containers/opensuse/leap:${version}";
-        push @released_images, "registry.opensuse.org/opensuse/leap:${version}";
-    }
-    elsif ((is_leap(">15.0") || is_microos(">15.0")) && is_x86_64) {
-        push @untested_images, "registry.opensuse.org/opensuse/leap/${version}/images/totest/containers/opensuse/leap:${version}";
-        push @released_images, "registry.opensuse.org/opensuse/leap:${version}";
-    }
-    elsif ((is_leap(">15.0") || is_microos(">15.0")) && (is_aarch64 || is_arm)) {
-        push @untested_images, "registry.opensuse.org/opensuse/leap/${version}/arm/images/totest/containers/opensuse/leap:${version}";
-        push @released_images, "registry.opensuse.org/opensuse/leap:${version}";
-    }
-    elsif (is_leap(">15.0") && is_ppc64le) {
-        # No image set up yet :-(
-    }
-    elsif (is_sle("<=12-sp2", $version)) {
-        # No images for old SLE
-    }
-    else {
+
+    if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
+        push @untested_images, $registries{$args{distri}}{$args{version}}{totest}->($args{arch});
+        push @released_images, $registries{$args{distri}}{$args{version}}{released}->($args{arch});
+    } else {
         die("Unknown combination of distro/arch.");
     }
 
@@ -143,3 +286,5 @@ sub get_3rd_party_images {
 
     return (\@images);
 }
+
+1;
--- a/tests/containers/buildah_docker.pm
+++ b/tests/containers/buildah_docker.pm
@@ -34,7 +34,7 @@ sub run {
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGES
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
-        my ($untested_images, $released_images) = get_suse_container_urls($version);
+        my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
         my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
         for my $iname (@{$images_to_test}) {
             record_info "IMAGE", "Testing image: $iname";
--- a/tests/containers/buildah_podman.pm
+++ b/tests/containers/buildah_podman.pm
@@ -34,7 +34,7 @@ sub run {
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGES
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
-        my ($untested_images, $released_images) = get_suse_container_urls($version);
+        my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
         my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
         for my $iname (@{$images_to_test}) {
             record_info "IMAGE", "Testing image: $iname";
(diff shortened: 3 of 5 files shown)


lib/containers/urls.pm
```diff
@@ -51,64 +52,229 @@ sub get_opensuse_registry_prefix {
...
 # If empty, no images available.
 sub get_suse_container_urls {
-    my $version    = shift // get_required_var('VERSION');
-    my $dotversion = $version =~ s/-SP/./r;                    # 15 -> 15, 15-SP1 -> 15.1
-    $dotversion = "${dotversion}.0" if $dotversion !~ /\./;    # 15 -> 15.0
+    my $version = shift // get_required_var('VERSION');
+    my $arch    = shift // get_required_var('ARCH');
+    my $distri  = shift // get_required_var('DISTRI');
 
     my @untested_images = ();
```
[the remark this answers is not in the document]
b10n1k: `untested_images` as well as `released_images`  came from a previous team agreement. i dont want to interfere and change them. i think it is messy to change the names frequently. in the other side i think the structure's name as quite good and they do not mess with the already implementation.

lib/containers/urls.pm
```diff
@@ -142,3 +308,60 @@ sub get_3rd_party_images {
 
     return (\@images);
 }
+
+sub fetch_image {
+    my () = @_;
+    return "registry.opensuse.org/" . get_opensuse_registry_prefix . "opensuse/tumbleweed";
+
+}
```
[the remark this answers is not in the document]
b10n1k: i think this is not valid anymore. Can you resolve conversation?

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: I like your `totest_` and `released_` prefixes. But I'd like to use the same prefixes everywhere.

Maybe another point: I'd rename `%registries` to `%images` and `released_registry` to `released_image` and `totest_registry` to `totest_image`.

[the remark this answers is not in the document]
pdostal: @jlausuch @Vogtinator can you help here so we agree on same convention and use it everywhere?

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
     }
 }
 
+our %registries = (
+    sle => {
+        '12-SP3' => {
+            released_registry => sub { "registry.suse.com/suse/sles12sp3" },
```
cfconrad, line 57: I think the hash is already about registries, so lets call it `released` `test`, wdyt?
```suggestion
            released => sub { "registry.suse.com/suse/sles12sp3" },
```

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
...
 # Returns a tuple of image urls and their matching released "stable" counterpart.
 # If empty, no images available.
 sub get_suse_container_urls {
-    my $version    = shift // get_required_var('VERSION');
-    my $dotversion = $version =~ s/-SP/./r;                    # 15 -> 15, 15-SP1 -> 15.1
-    $dotversion = "${dotversion}.0" if $dotversion !~ /\./;    # 15 -> 15.0
-
+    my $version         = shift // get_required_var('VERSION');
+    my $arch            = shift // get_required_var('ARCH');
+    my $distri          = shift // get_required_var('DISTRI');
```
cfconrad, line 251: I would prefer named parameters like:
```suggestion
    my (%args) = @_;
    $args{version} //= get_required_var('VERSION');
    $args{arch} //=  get_required_var('ARCH');
    $args{distri} //=  get_required_var('DISTRI');
```

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
...
+                    "";
+                }
+
+            },
+            available_arch => ['x86_64', 'arm']
+        },
+        '15.2' => {
+            released_registry => sub { "registry.opensuse.org/opensuse/leap:15.2" },
+            totest_registry   => sub {
+                if (check_var('ARCH', 'x86_64')) {
```
cfconrad, line 145: Using `check_var()` inhere is bad, as it would not use the given parameter to `get_suse_container_urls()`

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
     }
 }
 
+our %registries = (
```
cfconrad, line 54: I would go with
```perl
{
     '<flavor>' => {
               '<version>' => {
                        released => [ARRAY_REF | CODE],
                        staging => [ARRAY_REF | CODE],
                        supported_arch => ARRAY_REF  # <= OPTIONAL, if not given all archs are supported
                     },
               '<version_b>' => {
                      <arch_a> => {
                        released => [ARRAY_REF | CODE],
                        staging => [ARRAY_REF | CODE],
                      }
               }
   '<flavor_b>' => {
...
   }
}
```
CODE will always get a HASH as parameter looking like: `( arch => 'xxx', flavor => 'XXX', version => 'XxX' )` and **have to** return a ARRAY_REF.

I also thought about making the hash keys as regex, but I think this will endup as unreadable as before :)

The distinguished between `flavor->version->arch->released` and `flavor->version->released` can be done by checking for the key `released`...

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
...
+                    "";
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        },
+        '15.3' => {
+            released_registry => sub { "registry.opensuse.org/opensuse/leap:15.3" },
+            totest_registry   => sub {
+                if (check_var('ARCH', 'x86_64')) {
+                    "registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3";
```
cfconrad, line 229:
```suggestion
                    'registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3';
```
Use single ticks where every you can!

lib/containers/urls.pm
```diff
@@ -51,64 +51,211 @@ sub get_opensuse_registry_prefix {
...
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        }
+    }
+);
+
+sub check_image_arch_in_registry {
+    my ($distri, $version, $arch) = @_;
+    (grep /\b$arch\b/, @{$registries{$distri}{$version}{available_arch}}) ? 1 : 0;
```
cfconrad, line 243: there is no need for regex here
```suggestion
    (grep { $_ eq $arch } @{$registries{$distri}{$version}{available_arch}}) ? 1 : 0;
```

cfconrad (changes requested):
I like the idea to have a well formatted structure with all supported version/flavor/archs

But using `check_var()` inside these functions are wrong, as we do not consider the given parameters, which one of the errros we have currently too. That's why I put it to `request change`
Everything else is just loud thinking...

lib/containers/urls.pm
cfconrad: This is, how it could look like if we still would use regex as hash keys
```perl
       '15.3' => {
            released_registry => sub { "registry.opensuse.org/opensuse/leap:15.3" },
            totest_registry   => sub {
                if (check_var('ARCH', 'x86_64') || check_var('ARCH', 'aarch64') || check_var('ARCH', 'ppc64le') || check_var('ARCH', 's390x')) {
                    "registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3";
                } elsif (check_var('ARCH', 'arm')) {
                    "registry.opensuse.org/opensuse/leap/15.3/arm/images/totest/containers/opensuse/leap:15.3";
                } else {
                    "";
                }
            },
```
will be 
```perl
       '15.3' =>  {
          qr/^x86_64|aarch64|ppc64le$/ => {
              released_registry => [ 'registry.opensuse.org/opensuse/leap:15.3'],
              totest_registry   => ['registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3]'
          'arm' => {
                  released => [ 'registry.opensuse.org/opensuse/leap:15.3'],
                  staging => ['registry.opensuse.org/opensuse/leap/15.3/arm/images/totest/containers/opensuse/leap:15.3'
        }
}
````

Comment (cfconrad) [reply lib/containers/urls.pm:54]: **I need to correct my self** 
> CODE will always get a HASH as parameter looking like: ( arch => 'xxx', flavor => 'XXX', version => 'XxX' ) and have to return a ARRAY_REF.
I think we do not need a ARRAY as return value, so replace all the ARRAY_REF with simple SCALAR.

File: lib/containers/urls.pm (hunk 3 above)
Comment (b10n1k) [reply lib/containers/urls.pm:57]: could be. but i would like to keep the `totest`. we used it even in verbal communication when we talk about images and it is less general than `test`. is it ok?

Comment (cfconrad) [reply lib/containers/urls.pm:57]: sure!

File: lib/containers/urls.pm (hunk 6 above)
Comment (cfconrad) [reply lib/containers/urls.pm:54]: regarding regex:
maybe we should not go with full regex, but with multiple strings separated by some delimiter like `|`.

File: lib/containers/urls.pm (hunk 4 above)
Comment (b10n1k) [reply lib/containers/urls.pm:251]: This will add a few more files in the PR FYI

File: lib/containers/urls.pm (hunk 5 above)
Comment (b10n1k) [reply lib/containers/urls.pm:145]: i think that s ok!!! 
we want `x86_64` container images on `x86_64` hosts.
or.. missing smth?

File: lib/containers/urls.pm (hunk 3 above)
Comment (b10n1k) [reply lib/containers/urls.pm]: *this change is ready for review*

File: lib/containers/urls.pm (hunk 7 above)
Comment (b10n1k) [reply lib/containers/urls.pm]: *this change is ready for review*

File: lib/containers/urls.pm (hunk 8 above)
Comment (b10n1k) [reply lib/containers/urls.pm]: *this change is ready for review*

File: lib/containers/urls.pm (hunk 4 above)
Comment (b10n1k) [reply lib/containers/urls.pm]: *this change is ready for review* but i will have to make another commit for the affected files. leave the conversation open

File: lib/containers/urls.pm (hunk 5 above)
Comment (cfconrad) [reply lib/containers/urls.pm:145]: no, I think you don't. Still I woundn't do that, as this apply only for `ARCH` but not for `VERSION` or `FLAVOR` which are agruments as well.
And maybe there will be a helper host running on `x86_64` which connects to a `arm` one via `ssh` and installing there some containers... now it belongs to the one who create the job what `ARCH` means... HOST/SUT what ever.
I think it is just more clean to not use it here, as we already got `arch` once on the argument handling.

Comment (b10n1k) [reply lib/containers/urls.pm]: *this change is ready for review *
i replace the `check_var` with a conditional check of the passing `$arch`

File: lib/containers/urls.pm (hunk 6 above)
Comment (b10n1k) [reply lib/containers/urls.pm:54]: *this change is ready for review*
for future reference: 
- We tried to use regexp as key for `<version>`
- @cfconrad provided alternatives https://github.com/cfconrad/os-autoinst-distri-opensuse/commit/4c71ec13b577f4f1246816a6d807faeabc774b97 and https://github.com/cfconrad/os-autoinst-distri-opensuse/blob/play_with_container_urls/containers_urls.pl

File: lib/containers/urls.pm [hunk 9]
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+            released => sub { 'registry.opensuse.org/opensuse/leap:15.3' },
+            totest   => sub {
+                my $arch = shift;
+                if (grep { $_ eq $arch } qw/x86_64 aarch64 ppc64le s390x/) {
+                    'registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3';
+                } elsif ($arch eq 'arm') {
+                    'registry.opensuse.org/opensuse/leap/15.3/arm/images/totest/containers/opensuse/leap:15.3';
+                } else {
+                    '';
+                }
```
pdostal, line 167: Why do you return `''` instead of just skipping?

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+            },
+            available_arch => ['x86_64', 'aarch64', 'arm']
+        }
+    }
+);
+
+sub check_image_arch_in_registry {
```
pdostal, line 247:
```suggestion
sub return_image {
```

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
-    my $version    = shift // get_required_var('VERSION');
-    my $dotversion = $version =~ s/-SP/./r;                    # 15 -> 15, 15-SP1 -> 15.1
-    $dotversion = "${dotversion}.0" if $dotversion !~ /\./;    # 15 -> 15.0
-
+    my %args = (
+        version => get_required_var('VERSION'),
+        arch    => get_required_var('ARCH'),
+        distri  => get_required_var('DISTRI'),
+        @_
+    );
```
pdostal, line 260: My previous intention was to make this able to take specific parameters instead of variables as well.

WDYT @jlausuch?

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: Ping. I would still be for unifying those names. E.g. `released_images` and `totest_images` as we have above.

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
+        }
+    },
+    'sle-micro' => {
+        '15' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.0' },
+            totest         => sub { },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
```
cfconrad, line 176: That we do not have `aarch64` is new in the refactored code, Is it expected?

lib/containers/urls.pm
cfconrad: thats what it is, It is just a little bit obfuscated written. The code is more or less similar to:
```perl
    my (%args) = @_;
    $args{version} //= get_required_var('VERSION');
    $args{arch}    //= get_required_var('ARCH');
    $args{distri}  //= get_required_var('DISTRI');
```
with the exception, that the openqa variables are required, even if we call the function with parameters like `version => 'foo'`.

lib/containers/urls.pm
b10n1k: `check_image_arch_in_registry` is more descriptive. in the other side this function doesnt return any kind of image. it just checks whether the container image is built against the arch which is run on.

lib/containers/urls.pm
b10n1k: thats a good point. 
suse/containers/sle-server/15/containers/suse/sle15 with tag `15.0`, `arm64` is missing. Although if you check `latest` tag the `arm64` seems available. `latest` is tricky because it might be different build than `15.0`. Also the repo in obs, there are images built against aarch64 but not for `15_0`. 
So imo `aarch` shouldnt be in this list.

lib/containers/urls.pm
b10n1k: is it a big issue? i didnt overthink about it to be honest. it can be neglected.

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+            released => sub { 'registry.suse.com/suse/sle15:15.3' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
+        },
+        '15-SP4' => {
+            released => sub { '' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-15-sp4/update/cr/totest/images/suse/sle15:15.4 ';
```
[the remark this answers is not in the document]
b10n1k: i have replaced it. I might be wrong here but i think sle15sp4 should test `suse/sle-15-sp4/update/cr/totest/images/suse/sle15`. 
registry.s.d doesnt provide `suse/sle-15-sp4/ga/test/images/suse/sle15` (unless i have really overseen it) anyway.

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+                'registry.suse.de/suse/sle-12-sp4/docker/update/cr/totest/images/suse/sles12sp4';
+            },
+            available_arch => ['x86_64', 'ppc64le', 's390x']
+        },
+        '12-SP5' => {
+            released => sub { 'registry.suse.com/suse/sles12sp5' },
+            totest   => sub {
+                'registry.suse.de/suse/sle-12-sp5/docker/update/cr/totest/images/suse/sles12sp5';
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x']
```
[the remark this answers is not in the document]
b10n1k: i think it was wrong. the registry provides aarch for sle12sp5. Someone needs to confirm that i am wrong

lib/containers/urls.pm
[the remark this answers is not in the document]
cfconrad: I trust you!

lib/containers/urls.pm
cfconrad: ok

cfconrad (approved):
LGTM

lib/containers/urls.pm
pdostal: I think it does, imagine some new image will arrive for testing, someone will schedule it but forget to add it here.
The tests may still pass and nobody will see that the image is actually not tested. I would skip the `else` block.

Then at the end you can check if some image has been returned and if not we can do something like:
```perl
die("No image for DISTRI=$distr VERSION=$version ARCH=$arch available");
```

lib/containers/urls.pm
pdostal: OK, let's be descriptive: It does not check any registry, it checks hash. And the word `arch` does not fit there either.

What about `check_image_in_hash`?

lib/containers/urls.pm
pdostal: OK! That is super cool!

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: @b10n1k : you are right here, there is indeed 12-SP5 image for aarch64.

There are ONLY 3 gaps in our images, and they are ONLY for aarch64:
- 12-SP3
- 12-SP4
- 15 (SP0)

The rest of the versions have images in all the architectures.

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Yes, Marcus Meissner enabled this project for 15-SP4 images. I am not sure if it's wrong or not, cause we were told to use  `registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15:15.4` in the beginning... Let's keep it like this and change it later on if we are told so.

lib/containers/urls.pm
jlausuch: I agree with Pavel here. Better to use `die` here to not give a false green (the test won't test any image).

lib/containers/urls.pm
jlausuch: That's right, as in previous comment, 15(SP0) aarch64 doesn't exist.

lib/containers/urls.pm
```diff
@@ -51,64 +51,220 @@ sub get_opensuse_registry_prefix {
...
+                    '';
+                }
+            },
+            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
+        }
+    },
+    'sle-micro' => {
+        '15' => {
+            released       => sub { 'registry.suse.com/suse/sle15:15.0' },
+            totest         => sub { },
```
jlausuch, line 175: In case someone is wondering why this is empty, we are not doing container image validation on SLE Micro, so we shouldn't return any image here (there shouldn't be any test requesting it).

lib/containers/urls.pm
jlausuch: This sounds like a "tab or spaces" discussion :)
The function is basically checking if a given image exists in the hash which represents the registry contents... 

Honestly, not in favor of `check_image_in_hash` but I understand it's more descriptive... 

I can give you some examples that are descriptive and easy for the user to understand, but I would vote for simplicity and easiness. Why not `image_exists` (it returns true or false if agiven image exists or not).

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Sorry for the late answer. 
I am in favor of renaming `%registries` to `%images`.

And now that I hear Pavel's opinion, I kind of like it. I am also in favor of using `totest_images` instead of `unreleased_images`, but @b10n1k if you don't want to do that change here, I'm fine with doing that adaptation in a later PR. We probably need to rename some variable names, etc.

jlausuch:
This is actually a good improvement to have cleaner url configuration for each case. Good job. Some comments here and there, but looks very good.

lib/containers/urls.pm
cfconrad: I think this code will never be hitten, as we first call `check_image_arch_in_registry()` and thus, if someone give a arch which isn't there we already `die("Unknown combination of distro/arch.");`.

jlausuch: This is also true. The condition will be `if('')` which will be evaluated to `false` and it will hit the `die`. So, up to you how to implement it.

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: Let's just not forget about it.

lib/containers/urls.pm
pdostal: Exactly. If this won't be hitten I'd just remove the `else {}` completely.

lib/containers/urls.pm
b10n1k: Neither to be honest.
`check_image_in_hash` => it is not that bad but it does reflect the purpose in my opinion. 
`image_exists` => we return a tuple of released/totest but there are cases that one of them can be empty
i will rename it `supports_image_arch` => the function is checking the `available_arch` which contains the arch that are supported for the image based on the distri and the version.

lib/containers/urls.pm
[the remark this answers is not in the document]
b10n1k: My objection is that arbitrary change of the variable names or unrelated request of the scope is bad practice. The reason is because usually you trigger side effects and it might cause a chain of reactions+additional checks,VR and etc. Second, you request to rename some variables, unrelated to the scope of this PR, which as for beginning they are not bad ones and the discussion should be in the case the PR have made changes on them. Which it doesnt and a very important reason of it is that there was a [PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12698) which reviewed and approved (requester was one of them) after a long talk on the rocket.chat if i remember. i just think is really bad to approve something and someone changed it without a good "reason". this causes more issues and inconsistency in the reviews and waste our time.

tespa. i would like to leave it as it is for the reasons i just mentioned. and you can decide later what its destiny will be :)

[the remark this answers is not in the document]
b10n1k: *this change is ready for review*

lib/containers/urls.pm
b10n1k: *this change is ready for review*

lib/containers/urls.pm
b10n1k: *this change is ready for review*

lib/containers/urls.pm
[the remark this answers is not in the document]
Vogtinator: From 15 SP3 on, the locations before and after GA (maintenance testing) are different. For now it has to be `registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15` for SP4.

Vogtinator (changes requested):
"+202 -57", 145 more lines while also being less flexible :-/

Having a hash of subs means that it's not just plain data, but still code, IMO the shorter code was simpler, but that's just my view.

It's no longer possible to distinguish between "not available, skip" (e.g. 15.2 PowerPC) and "not known, abort" (none of the cases match).

The biggest issue IMO however is that it needs manual changes for every new version of SLE or Leap, while currently it handles later version of known distributions automatically.

Vogtinator:
> It's no longer possible to distinguish between "not available, skip" (e.g. 15.2 PowerPC) and "not known, abort" (none of the cases match).

Actually, that is possible. `push @foo, bar();` is a noop if `bar` doesn't return anything. `push @foo, undef;` however isn't. Perl is weird.

lib/containers/urls.pm
jlausuch: > `image_exists` => we return a tuple of released/totest but there are cases that one of them can be empty

I don't understand. I don't want you to return anything, just true or false as you are already doing :)

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Why is this resolved? Let's just update it to the right location.

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Another "tab or spaces" discussion :D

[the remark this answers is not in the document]
jlausuch: > My objection is that arbitrary change of the variable names or unrelated request of the scope is bad practice. The reason is because usually you trigger side effects and it might cause a chain of reactions+additional checks,VR and etc. Second, you request to rename some variables, unrelated to the scope of this PR, which as for beginning they are not bad ones and the discussion should be in the case the PR have made changes on them. Which it doesnt and a very important reason of it is that there was a [PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12698) which reviewed and approved (requester was one of them) after a long talk on the rocket.chat if i remember. i just think is really bad to approve something and someone changed it without a good "reason". this causes more issues and inconsistency in the reviews and waste our time.
> 
> tespa. i would like to leave it as it is for the reasons i just mentioned. and you can decide later what its destiny will be :)

I agree with this, it's not something for this PR as I mentioned.

jlausuch:
> "+202 -57", 145 more lines while also being less flexible :-/
> 
> Having a hash of subs means that it's not just plain data, but still code, IMO the shorter code was simpler, but that's just my view.

Yes, it's more code but it's also more clear. I am in favor of this format. Maybe, in the future we could put it in json/yaml/xml format in data directory and just read it here, so we don't create so many lines of code with that big hash. Something similar to what we do for journal_check test. But it's fine if that is done later.

lib/containers/urls.pm
b10n1k: > > `image_exists` => we return a tuple of released/totest but there are cases that one of them can be empty
> 
> I don't understand. I don't want you to return anything, just true or false as you are already doing :)

ignore. you can review the new name

lib/containers/urls.pm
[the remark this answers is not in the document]
b10n1k: @Vogtinator [from another pr which uses that registry](https://openqa.suse.de/tests/7061832#step/buildah_podman/83) you can see that it fails there. but i found that the path is also wrong. should it be `suse/sle-15-sp4/ga/images/suse/sle15` instead of `registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15`?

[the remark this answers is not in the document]
Vogtinator: No, it needs to be `registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15`. It was only fixed in IBS yesterday and there's no build released to test yet.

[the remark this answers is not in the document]
jlausuch: So, let's then update it to `registry.suse.de/suse/sle-15-sp4/ga/test/images/suse/sle15:15.4`. The next IBS build for 15-SP4 should contain the image already.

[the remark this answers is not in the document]
b10n1k: **this change is ready for review**

jlausuch (approved):

lib/containers/urls.pm
b10n1k: *additional commit added* 
thanks @cfconrad you can close this conversation if you are ok with the latest commit. i touched only files with a positional parameter.

Vogtinator (changes requested):
> The biggest issue IMO however is that it needs manual changes for every new version of SLE or Leap, while currently it handles later version of known distributions automatically.

Is still valid

b10n1k:
> > The biggest issue IMO however is that it needs manual changes for every new version of SLE or Leap, while currently it handles later version of known distributions automatically.
> 
> Is still valid

Sorry i thought that @jlausuch has answered on that and seems to be in favor of this structure. is this somehow different?

My opinion is that this is a more neat approach because you can "describe" what is your data.  there is a trade off with the procedural logic. the writing/reading is getting harder because the decision made by mixing version, distri and arch. the task was to simplify this. 

To avoid manual changes we could use this struct a little bit different.

```
our %registries = (
    sle => {
        '12' => {
            released => sub { 'registry.suse.com/suse/sles12sp3' },
            totest   => sub {
                'registry.suse.de/' . get_sle12($version);
            },
            available_arch => ['x86_64', 'ppc64le', 's390x']
        },
        15' => {
            released => sub { 'registry.suse.com/suse/sle15:15.0' },
            totest   => sub {
                'registry.suse.de/'' . get_sle15($version);
            },
            available_arch => ['x86_64', 'ppc64le', 's390x']
        },
    },
    opensuse => {
        Tumbleweed => {
            released => sub { 'registry.opensuse.org/opensuse/tumbleweed' },
            totest   => sub {
                'registry.opensuse.org/' . get_opensuse_registry_prefix . 'opensuse/tumbleweed';
            },
            available_arch => ['x86_64', 'aarch64', 'ppc64le', 's390x', 'arm']
        },
        'leap => {
            released => sub { 'registry.opensuse.org/opensuse/leap:15.0' },
            totest   => sub {
                'registry.opensuse.org/ . get_leap($version)';
            },
            available_arch => ['x86_64']
        }
       }
```

or something like that. 
So for sure it will be nice to have it, it is possible on my point of view but i think there is time to figure this out.

lib/containers/urls.pm
pdostal: There are still occasions like `released => sub { }` or `released => sub { }` and I'd remove those as well.

lib/containers/urls.pm
pdostal: OK. cc: @jlausuch

lib/containers/urls.pm
```diff
@@ -52,64 +52,207 @@ sub get_opensuse_registry_prefix {
     }
 }
 
+our %registries = (
```
pdostal, line 55:
```suggestion
our %images = (
```
I'm sorry, but we're not listing **registries** there. We're listing image names.

ilausuch (approved):
For me is good, only that I agree to change the registries variable name to images, more clear what contains

jlausuch:
> > > The biggest issue IMO however is that it needs manual changes for every new version of SLE or Leap, while currently it handles later version of known distributions automatically.
> > 
> > 
> > Is still valid
> 
> Sorry i thought that @jlausuch has answered on that and seems to be in favor of this structure. is this somehow different?

Yes, I am in favor. If we need to add new releases it won't be a big deal, maybe once or twice a year? For new releases there are actually bigger changes we need to do, adding a few lines here is not that much of a problem. 

The only thing against having this hash here is just that it takes too much lines in a `.pm` module, so I would be also in favor to transform this into json file and parse it here, but this can be done in another PR.

lib/containers/urls.pm
b10n1k: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13141#discussion_r705007863

in brief: `registry` is the location that an image is stored. the *image* is defined by the whole uri (local: locahost/alpine, remote registry.whatever.com/alipine). We cant name everything *images* and i thing *%registries* solves that.

b10n1k:
> For me is good, only that I agree to change the registries variable name to images, more clear what contains

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13141#discussion_r707099288. feel free to suggests something different. for instance *%images_uri* but i still think that registries describes it just fine.

lib/containers/urls.pm
jlausuch: As I said, I don't care about the name, but answering to this comment, you are actually providing the full uri of the images, e.g. `registry.suse.de/suse/sle-12-sp3/docker/update/cr/totest/images/suse/sles12sp3`, you are not providing just the registry `registry.suse.de`.

lib/containers/urls.pm
b10n1k: those are fine and they have their purpose. Avoid `Can't use an undefined value`. simplifies invocation and error handling

pdostal:
> > For me is good, only that I agree to change the registries variable name to images, more clear what contains
> 
> 
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13141#discussion_r707099288. feel free to suggests something different. for instance *%images_uri* but i still think that registries describes it just fine.

Either `%images_uri` or just `%images` work for me. Thx Yianni

jlausuch:
It seems this is being blocked because of the hash name. Let's just let people vote:
`%registries` or `%images`.

I vote for `%images`

ilausuch:
I vote for %images, %registries confuses for me

b10n1k:
> It seems this is being blocked because of the hash name. Let's just let people vote:
> `%registries` or `%images`.
> 
> I vote for `%images`

are we going to vote for each review conflict? i even proposed another name... and i will add a few here in case you want to pick something from that which will define the hash uniquely 
- %images_uri
- %images_struct
- %registry_images
- %suse_images

Again consider that the hash is not a private variable and it is worth to give a better name than %images

jlausuch:
>     * %images_uri
>     * %images_struct
>     * %registry_images
>     * %suse_images
>

 %images_uri +1

b10n1k:
> > ```
> > * %images_uri
> > * %images_struct
> > * %registry_images
> > * %suse_images
> > ```
> 
> %images_uri +1

First comes First Served?? :) *this change is ready for review*

pdostal:
> > The biggest issue IMO however is that it needs manual changes for every new version of SLE or Leap, while currently it handles later version of known distributions automatically.
> 
> Is still valid

This is in my opinion out of scope of this PR and also quite complex problem as we test prerelease versions too.

pdostal (approved):

b10n1k:
follow up ticket  to address manual changes to the hash https://progress.opensuse.org/issues/98658

DimStar77:
This seems responsible for the staging failures, like
https://openqa.opensuse.org/tests/1918468#step/podman_image/28

PR #13147, opened 2021-08-30
yast2_system_settings: Simplify xterm and grep calls

okurz:
As observed in
https://openqa.opensuse.org/tests/1887939#step/yast2_system_settings/43
an openQA test ran into a problem after running `validate_script_output`
within an xterm session. X11 applications seem to be more prone to
mistyping characters so we should be careful to type less. This might
help in the future to reduce mistyping failures in this test module. A
next step could be to use the "root-console" or a proper serial
terminal.


okurz:
Mee, too bad. Sorry. Reverting in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13429

PR #13148, opened 2021-08-30
containers: Simplify 'zypper ref' calls

okurz:
'zypper ref' correctly returns a zero exit code for success only if
repositories can be successfully refreshed. So there is no need to grep
the output. Also this prevents some problems which ask for the
non-interactive mode anyway.

Related progress issue: https://progress.opensuse.org/issues/95311

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -258,17 +258,15 @@ sub test_zypper_on_container {
     assert_script_run("$runtime run $image zypper lr -s", 120);
 
     if ($runtime =~ /buildah/) {
-        # zypper ref
-        assert_script_run("$runtime run $image -- zypper -v ref | grep \"All repositories have been refreshed\"", 120);
+        assert_script_run("$runtime run $image -- zypper -nv ref", 120);
 
         # Create new image and remove the working container
         assert_script_run("$runtime commit --rm $image refreshed", 120);
 
         # Verify the new image works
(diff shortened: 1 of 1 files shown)


pdostal (approved):

lib/containers/container_images.pm
```diff
@@ -258,17 +258,15 @@ sub test_zypper_on_container {
...
     if ($runtime =~ /buildah/) {
-        # zypper ref
-        assert_script_run("$runtime run $image -- zypper -v ref | grep \"All repositories have been refreshed\"", 120);
+        assert_script_run("$runtime run $image -- zypper -n ref", 120);
 
         # Create new image and remove the working container
         assert_script_run("$runtime commit --rm $image refreshed", 120);
 
         # Verify the new image works
-        assert_script_run("$runtime run \$($runtime from refreshed) -- zypper -v ref | grep \"All repositories have been refreshed\" ", 120);
```
b10n1k, line 268: i dont see any reason to remove the verbose mode. Could you add it back?

b10n1k (changes requested):
if you have any good reason for the remove of the verbose flag let me know. Otherwise lets not change that

lib/containers/container_images.pm
okurz: done

okurz:
> if you have any good reason for the remove of the verbose flag let me know. Otherwise lets not change that

Sure, we can keep the "-v" flag

b10n1k (approved):
+1

PR #13155, opened 2021-08-31
Rewrite workaround for scrolling for bsc#1189550

jknphy:
Based on feedback from [bsc#1189550](https://bugzilla.suse.com/show_bug.cgi?id=1189550#c16) we can improve the workaround and avoid failure in latest build in SLE-15-SP4.

- Related ticket: [poo#97760](https://progress.opensuse.org/issues/97760) | [previous workaround](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13087)
- Verification run: [mru-install-minimal-with-addons + select_modules_and_patterns](https://openqa.suse.de/tests/overview?distri=sle&build=jknphy%2Fos-autoinst-distri-opensuse%23change_workaround_scrollbar)

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -27,6 +27,8 @@ use utils qw(type_string_slow zypper_call);
 use lockapi;
 use mmapi;
 
+my $workaround_bsc1189550_done;
+
 =head1 y2_installbase
 
 C<y2_installbase> - Base class for Yast installer related functionality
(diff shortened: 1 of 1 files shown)


czerw:
Pleas do verification also on lower versions (15-sp2, and maybe 12-SP5) to see that `mode_down` function works fine after the change.

jknphy:
> Pleas do verification also on lower versions (15-sp2, and maybe 12-SP5) to see that `mode_down` function works fine after the change.

in lower version the workaround will never be executed and it was only from SP3 and on. I think we are safe, it doesn't modify previous behavior.

czerw:
> > Pleas do verification also on lower versions (15-sp2, and maybe 12-SP5) to see that `mode_down` function works fine after the change.
> 
> in lower version the workaround will never be executed and it was only from SP3 and on. I think we are safe, it doesn't modify previous behavior.

I know, but we are touching code which runs everywhere, just to make sure we will not break it.

czerw (approved):

PR #13167, opened 2021-09-02
Adapt check_resume to new implemented logic

JRivrain:
Failing scenario: https://openqa.suse.de/tests/6950236

According to https://bugzilla.suse.com/show_bug.cgi?id=1188731

    Hibernation is proposed by default in x86_64
    Hibernation is NOT proposed for any other architecture

Instead on unschedule test for x86_64 we should rename this negative test and probably create a positive one for x86_64.
Please, consider other products, as the check is present for SLE-15-SP1.
Suggestion: try to avoid test data and just be explicit with the check hardcoded in the test module.

VRs: https://openqa.suse.de/tests/overview?build=JRivrain%2Fos-autoinst-distri-opensuse%2313167

### DIFF
--- a/schedule/yast/btrfs/btrfs_sle_libstorage-ng.yaml
+++ b/schedule/yast/btrfs/btrfs_sle_libstorage-ng.yaml
@@ -38,12 +38,22 @@ schedule:
   - installation/first_boot
   - console/system_prepare
   - console/force_scheduled_tasks
-  - console/check_resume
+  - '{{validate_hibernation}}'
   - console/validate_product_installed_SLES
   - console/validate_no_cow_attribute
   # On all the backends except s390x, /home is located on a separate partition
   - '{{validate_home_partition}}'
 conditional_schedule:
+  validate_hibernation:
+    ARCH:
+      aarch64:
+        - console/hibernation_disabled
+      ppc64le:
+        - console/hibernation_disabled
+      s390x:
+        - console/hibernation_disabled
+      x86_64:
+        - console/hibernation_enabled
   isosize:
     BACKEND:
       qemu:
--- a/schedule/yast/btrfs/btrfs_sle_libstorage-ng@ipmi.yaml
+++ b/schedule/yast/btrfs/btrfs_sle_libstorage-ng@ipmi.yaml
@@ -32,7 +32,7 @@ schedule:
   - installation/first_boot
   - console/system_prepare
   - console/force_scheduled_tasks
-  - console/check_resume
+  - console/hibernation_enabled
   - console/validate_product_installed_SLES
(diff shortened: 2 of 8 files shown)


schedule/yast/btrfs/btrfs_sle_libstorage-ng.yaml
```diff
@@ -38,12 +38,25 @@ schedule:
   - installation/first_boot
   - console/system_prepare
   - console/force_scheduled_tasks
-  - console/check_resume
+  - '{{hibernation_enabled}}'
```
jknphy, line 41: you just need 1 placeholder to replace as both test would not run at the same time.

JRivrain: It runs both on x86_64 and other architectures, so it needs to be this way. Or I am I missing something ?

jknphy: we can just do it in one block, like in other examples in the repo, it is designed in that way:
```
  validate_hibernation:
    ARCH:
      aarch64:
        - console/hibernation_disabled
      ppc64le:
        - console/hibernation_disabled
      s390x:
        - console/hibernation_disabled
      x86_64:
        - console/hibernation_enabled
 ```

Comment (JRivrain) [reply schedule/yast/btrfs/btrfs_sle_libstorage-ng.yaml:41]: Ah, I did not know, thanks !

File: schedule/yast/btrfs/btrfs_sle_libstorage-ng@ipmi.yaml [hunk 2]
```diff
@@ -32,7 +32,6 @@ schedule:
   - installation/first_boot
   - console/system_prepare
   - console/force_scheduled_tasks
-  - console/check_resume
```
JRivrain, line 35: Excluded for ipmi as I could not manage to have a working VR. Not sure hibernation would work on ipmi (or hyperv) but we don't really validate it anyway, as we only check if the parameter is present, we don't actually try to resume after hibernation.

OleksandrOrlov: According to [bsc#1188731](https://bugzilla.suse.com/show_bug.cgi?id=1188731):

```
- Hibernation is proposed by default in x86_64 
- Hibernation is NOT proposed for any other architecture
```

So, it seems like 'resume=' parameter should be present for ipmi.

JRivrain: I can add the check for ipmi, but that will be without VR then.

PR #13171, opened 2021-09-02
Increase timeout for kdump restart info

czerw:
Fix poo#96702: Default timeout is not enough, it is set to 90 seconds now.

- Related ticket: https://progress.opensuse.org/issues/96702
- Needles: none
- Verification run: 
ppc64le: https://openqa.suse.de/tests/7002076
aarch64:  https://openqa.suse.de/tests/7002075


czerw:
@mloviska please review

PR #13176, opened 2021-09-03
Edit registry url for sle15sp4

b10n1k:
registry.suse.de has `suse/sle-15-sp4/update/cr/totest/images/suse/sle15`
    which i think we should use for the sle15sp4
    
    as for now we have two issues in latest group:
    - missing CONTAINERS_UNTESTED_IMAGES
    - wrong url for totest registry
    - 


b10n1k:
invalid as it was fix in another commit

PR #13186, opened 2021-09-03
Enable s390x tests for staging

sofiasyria:
Enable s390x tests RAID1 and cryptlvm_minimal_x for staging

- Related ticket: https://progress.opensuse.org/issues/95751
- Verification run fail due to s390x problem: 
RAID1 - https://openqa.suse.de/tests/7094875
cryptlvm_minimal_x - https://openqa.suse.de/tests/7110931 (Disabled temporarily failing modules)
     x86_64 : https://openqa.suse.de/tests/7111254

### DIFF
--- a/schedule/staging/RAID1@s390x-staging.yaml
+++ b/schedule/staging/RAID1@s390x-staging.yaml
@@ -1,0 +1,68 @@
+---
+name:           RAID1_s390x@s390x-staging
+description:    >
+  Test for RAID1
+  Installation of RAID1 using expert partitioner.
+vars:
+  RAIDLEVEL: 1
+  YUI_REST_API: 1
+schedule:
+  - installation/bootloader_start
+  - installation/setup_libyui
+  - installation/product_selection/access_beta_distribution
+  - installation/product_selection/install_SLES
+  - installation/licensing/accept_license
+  - installation/registration/register_via_scc
+  - installation/module_selection/skip_module_selection
+  - installation/add_on_product/skip_install_addons
+  - installation/system_role
+  - installation/partitioning
+  - installation/partitioning/raid_gpt
+  - installation/installer_timezone
+  - installation/hostname_inst
+  - installation/user_settings
+  - installation/user_settings_root
+  - installation/resolve_dependency_issues
+  - installation/installation_overview
+  - installation/disable_grub_timeout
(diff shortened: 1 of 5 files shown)


schedule/staging/RAID1@s390x-staging.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - installation/user_settings
+  - installation/user_settings_root
+  - installation/resolve_dependency_issues
+  - installation/installation_overview
+  - installation/disable_grub_timeout
+  - installation/start_install
+  - installation/await_install
+  - installation/logs_from_installation_system
+  - installation/reboot_after_installation
+  - installation/grub_test
```
jknphy, line 30: I think it should be `handle_reboot`. Same for the other test suite.

jknphy (commented):
LGTM, on the HDDSIZEGB for both we could go for 20GB same than its x86_64 partner, might worth to test it with that size.

schedule/staging/RAID1@s390x-staging.yaml
sofiasyria: Done

jknphy:
It might be that we don't need to select desktop module (we do similar for powerVM and s390x (basically remote workers) in our group as far as I could check), could you try that? or just comment it in the same way than for the other case.

jknphy:
btw, just noticed that for the encrypt scenario we didn't apply libyui, it would be easier to do the change now we have #13192 13192 Did you find some issue with that?

jknphy:
> The failure might be related with [selecting desktop module](https://openqa.suse.de/tests/7043920#step/scc_registration/14) (we do similar for remote workers, powerVM, s390x in our group as far as I could check), could you try not selecting it? or just comment it in the same way than for the other case.

I'm not sure if the test suite make sense s390x, but I double-check and we have some scenarios with desktop, so it could be.

sofiasyria:
Selecting deskop module is essential for the particular test. I enabled libyui and used the new modules where I could.

jknphy (approved):
LGTM, still there is a few modules to be replaced but you already did more that expected when switching to libyui, we will revisit the leftovers in the future for our groups as well. Good job!
You can proceed to configure job groups.

PR #13188, opened 2021-09-06
Remove unstable tests from Staging test runs

zhaocongjia:
https://progress.opensuse.org/issues/98135

- Related ticket: https://progress.opensuse.org/issues/98135
- Needles: None
- Verification run: (None. First, this commit only changes schedule yaml file; second, the original test is unable to rerun due to asset file being removed.)


GraceWang571:
Since the frequently failed yast2_snapper blocks SLE15SP4-Staging, I agree with removing it from the Staging test run temporarily and add it back after making the test case more robust.

PR #12842, opened 2021-07-05
Apply runtime.pm in containers libraries

b10n1k:
Some hybrid approach to integrate the runtime.pm into the container tests.
This intends to improve a bit the design of the tests regarding containers.
In theory it gives you the possibility to initialize a runtime and implement
specifics of that particular Class instead of if/else inside the test.
For now the decoupling is not in place though.

- Use runtime.pm and its derived subclasses
- Some things added to runtime.pm but not everything is modified to avoid huge changes for now
- no change in the behavior or in the test logic

**I started this [pad](https://etherpad.nue.suse.com/p/runtime.pm) to collect questions, concerns, ideas for improvements) 
put there all you have. This is to answer questions and come together with a design which everybody understand. Then we can create either tickets for request or have a meeting with specific agenda or...**

Verification run: 
[15sp1](https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2312842&version=15-SP1&distri=sle)
[15sp3](https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2312842&version=15-SP3&distri=sle)
[12sp5](https://openqa.suse.de/tests/overview?version=12-SP5&build=b10n1k%2Fos-autoinst-distri-opensuse%2312842&distri=sle)
[TW](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2312842)

11.09.21:
[o3-tw](https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2312842&distri=opensuse)
[osd-sle15sp2](https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2312842&distri=sle&version=15-SP2)

PR #13189, opened 2021-09-06
POC: BCI test integration

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/94003

**NOTE**: This is a PoC, so it's tested only on 15-SP3 host as an example. However, the code is valid to be enabled on other hosts. Therefore, I created [this ticket](https://progress.opensuse.org/issues/98183) to extend the coverage to all the hosts we support. However, there are some things to be fixed in the upstream repository and the needed adaptations in this module can be done in later PRs.

### DIFF
--- a/tests/containers/bci_tests.pm
+++ b/tests/containers/bci_tests.pm
@@ -1,0 +1,117 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: bci-tests runner
+#   SUSE Linux Enterprise Base Container Images (SLE BCI)
+#   provides truly open, flexible and secure container images and application
+#   development tools for immediate use by developers and integrators without
+#   the lock-in imposed by alternative offerings.
+#
+#   This module is used to test BCI repository within a container.
+#
+#   It installs the required python packages and uses the existing BCI-test
+#   repository defined by BCI_TESTS_REPO. Then, it will run the different
+#   tests environments defined by BCI_TEST_ENVS.
+#   - install needed python environment
+#   - clone bci-tests repo
+#   - build project
+#   - run test for each given environment
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use Mojo::Base qw(consoletest);
+use XML::LibXML;
+use utils qw(systemctl zypper_call ensure_ca_certificates_suse_installed);
+use version_utils qw(get_os_release);
+use containers::common;
+use testapi;
+
+our $test_envs = get_var('BCI_TEST_ENVS', 'base,init,dotnet,python,node,go,multistage');
+
+sub parse_logs {
+    my $self = @_;
+
+    upload_logs('junit_build_serial.xml');
+    upload_logs('junit_build_parallel.xml');
+
+    # bci-tests produce separate XUnit results files for each environment.
+    # We need tp merge all together into a single xml file that will
+    # be used by OpenQA to represent the results in "External results"
+    my $dom  = XML::LibXML::Document->new('1.0', 'utf-8');
+    my $root = $dom->createElement('testsuites');
+    $dom->setDocumentElement($root);
+
+    # Dump xml contents to a location where we can access later using data_url
+    for my $env (split(/,/, $test_envs)) {
+        upload_logs('junit_' . $env . '_serial.xml');
+        my $log_file = upload_logs('junit_' . $env . '_parallel.xml');
+        my $dom      = XML::LibXML->load_xml(location => "ulogs/$log_file");
+        for my $node ($dom->findnodes('//testsuite')) {
+            # Replace default attribute name "pytest" by its env name
+            $node->{name} =~ s/pytest/$env/;
+            # Append test results to the resulting xml file
+            $root->appendChild($node);
+        }
+    }
+    $dom->toFile('result.xml', 1);
+    # Download file from host pool to the instance
+    assert_script_run('curl ' . autoinst_url('/files/result.xml') . ' -o /tmp/result.txt');
+    parse_extra_log('XUnit', '/tmp/result.txt');
+}
+
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $runtime        = get_required_var('CONTAINER_RUNTIME');
+    my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
+    my $bci_devel_repo = get_var('BCI_DEVEL_REPO');
+    my $bci_timeout    = get_var('BCI_TIMEOUT', 900);
+
+    ensure_ca_certificates_suse_installed;
+
+    my ($running_version, $sp, $host_distri) = get_os_release;
+    if ($runtime eq 'podman') {
+        install_podman_when_needed($host_distri);
+        install_buildah_when_needed($host_distri);
+    }
+    elsif ($runtime eq 'docker') {
+        install_docker_when_needed($host_distri);
+    }
+    else {
+        die("Runtime $runtime not given or not supported");
+    }
+
+    record_info('Install', 'Install needed packages');
+    zypper_call('--quiet in git-core python39', timeout => 600);
+    assert_script_run('pip3.9 --quiet install tox pytest', timeout => 600);
+
+    record_info('Clone', 'Clone BCI tests repository');
+    assert_script_run("git clone -q --depth 1 $bci_tests_repo");
+
+    record_info('Build', 'Build bci-tests project');
+    assert_script_run('cd bci-tests');
+    assert_script_run("export CONTAINER_RUNTIME=$runtime");
+    assert_script_run("export BCI_DEVEL_REPO=$bci_devel_repo") if $bci_devel_repo;
+    assert_script_run('tox -e build', timeout => 600);
+
+    # Run the tests for each environment
+    for my $env (split(/,/, $test_envs)) {
+        record_info("Test: $env");
+        assert_script_run("tox -e $env", timeout => $bci_timeout);
+    }
+
+    $self->parse_logs();
+}
+
+sub test_flags {
+    return {fatal => 1, milestone => 1};
+}
+
+1;
--- a/schedule/containers/sle_image_on_sle_host_docker.yaml
+++ b/schedule/containers/sle_image_on_sle_host_docker.yaml
@@ -7,9 +7,14 @@ conditional_schedule:
     ARCH:
       's390x':
         - installation/bootloader_start
+  bci:
+    HOST_VERSION:
+      '15-SP3':
+        - containers/bci_tests
 schedule:
   - '{{boot}}'
   - boot/boot_to_desktop
   - containers/host_configuration
   - containers/docker_image
+  - '{{bci}}'
   - containers/container_diff
--- a/schedule/containers/sle_image_on_sle_host_podman.yaml
+++ b/schedule/containers/sle_image_on_sle_host_podman.yaml
@@ -7,8 +7,13 @@ conditional_schedule:
     ARCH:
       's390x':
         - installation/bootloader_start
+  bci:
+    HOST_VERSION:
+      '15-SP3':
+        - containers/bci_tests
 schedule:
   - '{{boot}}'
   - boot/boot_to_desktop
   - containers/host_configuration
   - containers/podman_image
+  - '{{bci}}'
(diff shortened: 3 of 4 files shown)


jlausuch:
VR with isolated schedule: http://fromm.arch.suse.de/tests/2034

jlausuch:
2 more VRs in the expected schedule:

- [podman](http://fromm.arch.suse.de/tests/2041)
- [docker](http://fromm.arch.suse.de/tests/2043)

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+    record_info('Clone', 'Clone BCI tests repository');
+    assert_script_run("git clone -q --depth 1 $bci_tests_repo");
+
+    record_info('Build', 'Build bci-tests project');
+    assert_script_run('cd bci-tests');
+    assert_script_run("export CONTAINER_RUNTIME=$runtime");
+    assert_script_run("export BCI_DEVEL_REPO=$bci_devel_repo") if $bci_devel_repo;
+    assert_script_run('tox -e build', timeout => 600);
+    upload_logs('junit_build_serial.xml');
+    upload_logs('junit_build_parallel.xml');
```
mloviska, line 65: Is it necessary to have this part outside of test case loop? I mean, it feels like build tests should be always as the first element of `BCI_TEST_ENVS` list.

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+
+=head2 BCI_TEST_ENVS
+
+The list of environments to be tested.
+e.g. base,init,dotnet,python,node,go,multistage
+
+=head2 BCI_DEVEL_REPO
+
+This parameter is given to the bci-tests to inject a different SLE_BCI
+repository url to the container image instead of the default one.
```
mloviska, line 129: as `my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');` what should be the default repository? Should it be provided by some other tool as SCC?

mloviska: As sorry misread the variable name.

tests/containers/bci_tests.pm
jlausuch: `build` is a pre-step you need to do before running any tests, it basically sets up the needed packages used by the tests. I would like to keep it separate as it's not an actual test. Maybe @dcermak can comment on this?

tests/containers/bci_tests.pm
jlausuch: Good point. For now, it's in gitlab, and in an individual's namespace. So, I'd keep it like this until we have a more official placement for the repo.

jlausuch: And I misunderstood your question :D

mloviska: Mondays are hard :D

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $runtime        = get_required_var('RUNTIME');
+    my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
+    my $test_envs      = get_required_var('BCI_TEST_ENVS');
+    my $bci_devel_repo = get_var('BCI_DEVEL_REPO');
+    my $bci_timeout    = get_var('BCI_TIMEOUT', 900);
+
+    ensure_ca_certificates_suse_installed;
```
asmorodskyi, line 38: from this line and till 50th - this code snippet which I have seen in several places . maybe not related to this PR but maybe we can extract it into function ?

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+    }
+    elsif ($runtime eq 'docker') {
+        install_docker_when_needed($host_distri);
+    }
+    else {
+        die("Runtime $runtime not given or not supported");
+    }
+
+    record_info('Install', 'Install needed packages');
+    zypper_call('--quiet in git-core python39', timeout => 600);
```
asmorodskyi, line 53: hardcoding certain python version ? there should be some good reasons for this . do we have them ? can we add some comment here regarding this ?

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+
+    record_info('Clone', 'Clone BCI tests repository');
+    assert_script_run("git clone -q --depth 1 $bci_tests_repo");
+
+    record_info('Build', 'Build bci-tests project');
+    assert_script_run('cd bci-tests');
+    assert_script_run("export CONTAINER_RUNTIME=$runtime");
+    assert_script_run("export BCI_DEVEL_REPO=$bci_devel_repo") if $bci_devel_repo;
+    assert_script_run('tox -e build', timeout => 600);
+    upload_logs('junit_build_serial.xml');
```
asmorodskyi, line 64: maybe I am missing something because I have not much knowledge about what BCI is and how actually it works . but my understanding that start from this line there is a mix of two things which will looks better separated : 
1. actually running BCI tests
2. processing results to display it in openQA 

I think if we will split this into two fully isolated steps it will make code more clear for support and debug of failures . So first simply run the tests without any LibXML/ upload_logs actions , later scan init libXML magic and scan VM for result xml's and upload them all . Yeah this will lead to more code ( maybe ... ) 

P.S. it is just idea feel free to ignore it if you don't like it :)

tests/containers/bci_tests.pm
jlausuch: It's a PoC, and this is to be investigated in [this ticket](https://progress.opensuse.org/issues/98183) 
It comes from [requierements of the project](https://gitlab.suse.de/dancermak/bci-tests#what-do-i-need-to-contribute). But I agree with you, but this is a working PoC, and this is to be investigated in [this ticket](https://progress.opensuse.org/issues/98183) when we need to enable it on other hosts.

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,135 @@
...
+use containers::common;
+use testapi;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $runtime        = get_required_var('RUNTIME');
+    my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
+    my $test_envs      = get_required_var('BCI_TEST_ENVS');
```
grisu48, line 34: Can we add a comment here which environments are supported?

```suggestion
    my $test_envs      = get_required_var('BCI_TEST_ENVS');       # supported enviornments are: base,init,dotnet,python,node,go,multistage
```

grisu48 (approved):
Minor issue other than that LGTM for a PoC. Good job Jose!

tests/containers/bci_tests.pm
asmorodskyi: fare enough

asmorodskyi (approved):
LGTM

tests/containers/bci_tests.pm
jlausuch: There is already an example in line 124, but I will add all the possible envs.

tests/containers/bci_tests.pm
jlausuch: It's a good idea. I was actually making use of the same loop :) 
I will create another one for parsing the results. Thanks!

jlausuch:
2 new VRs with latest update
http://fromm.arch.suse.de/tests/2048
http://fromm.arch.suse.de/tests/2049

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,149 @@
...
+use version_utils qw(get_os_release);
+use containers::common;
+use testapi;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $runtime        = get_required_var('RUNTIME');
+    my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
```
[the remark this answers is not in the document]
pdostal: Also don't forget to add it to the variables list.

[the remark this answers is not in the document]
jlausuch: This variable is used only in this test, there are plenty of variables that are not in mentioned variables.md, I'm not sure it's a good idea if it's used in single place. 

It's required cause the location might change and currently it's pointing to some suse internal repo. I would prefer not to hardcode links to gitlab.suse.de.

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,149 @@
...
+    my $bci_devel_repo = get_var('BCI_DEVEL_REPO');
+    my $bci_timeout    = get_var('BCI_TIMEOUT',   900);
+    my $test_envs      = get_var('BCI_TEST_ENVS', 'base,init,dotnet,python,node,go,multistage');
+
+    ensure_ca_certificates_suse_installed;
+
+    my ($running_version, $sp, $host_distri) = get_os_release;
+    if ($runtime eq 'podman') {
```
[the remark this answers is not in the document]
jlausuch: I know, and I really refuse to have 2 files with the exact same code to only vary 1 line.

tests/containers/bci_tests.pm
jlausuch: I'm ok with subrutine for treating the logs, which I'll do. But for running the test, it's just 4 lines of code, not sure it's worth it :)

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
ilausuch: I was following this comment, and I checked that neither PUBLIC_CLOUD_* variables are in variables.md

ilausuch (approved):

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
jlausuch: Right, and those are used in several places :)

tests/containers/bci_tests.pm
```diff
@@ -0,0 +1,158 @@
...
+            # Replace default attribute name "pytest" by its env name
+            $node->{name} =~ s/pytest/$env/;
+            # Append test results to the resulting xml file
+            $root->appendChild($node);
+        }
+    }
+    $dom->toFile("os-autoinst-distri-opensuse/data/result.xml", 1);
+    # Download file from host pool to the instance
+    assert_script_run 'curl -f -v ' . data_url('result.xml') . " > /tmp/result.txt";
+    parse_extra_log('XUnit', '/tmp/result.txt');
```
cfconrad, line 57:
```suggestion
    $dom->toFile('result.xml', 1);
    # Download file from host pool to the instance
    assert_script_run('(cd /tmp/ && curl -O ' . autoinst_url('/files/result.xml') .  ')');
    parse_extra_log('XUnit', '/tmp/result.txt');
```

I think this is more common.

jlausuch:
VRs with latest update:
http://fromm.arch.suse.de/tests/2080
http://fromm.arch.suse.de/tests/2079

tests/containers/bci_tests.pm
jlausuch: I changed it, thanks for the suggestion :)

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
pdostal: I know that it's not good approach. But at least we use it everywhere so it's the current standard.

If you have better idea, please let's create a ticket and let's apply it for all our test modules.

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
pdostal: Guys there is this file and it should contain all variables we use. That is the rule we simply have to follow.
If you introduce a variable and not include it there you create mess and work for others.
I don't consider that funny, I consider that as problem.

Regarding the `PUBLIC_CLOUD_` variables I plan to collect them as the last part of [poo#97742](https://progress.opensuse.org/issues/97742).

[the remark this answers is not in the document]
pdostal: I was thinking about this and:
What if we use this approach with `RUNTIME` everywhere?

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
jlausuch: Maybe, with Yannis approach with runtime class will be solved... but for this file I would like to keep it like this, not duplicate it, as you said, it's not good approach, so I don't want to start a new test with the bad approach :)

tests/containers/bci_tests.pm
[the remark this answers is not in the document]
jlausuch: No one said it was funny :)

So, we agree that public cloud variables should end up in variables.md. I have created [poo#98292](https://progress.opensuse.org/issues/98292) for that so we don't forget.

But for this specific case, I don't think it makes sense. I have a section at the end of this file to explain all the variables because they are used only in this file. I'm not sure we have such strict rule, but I might be wrong.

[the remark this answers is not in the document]
jlausuch: > I was thinking about this and:
> What if we use this approach with `RUNTIME` everywhere?

We should find an agreement about https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12842 fist. But maybe it makes sense, yes.

jlausuch:
@pdostal I moved variables to variables.md

pdostal (approved):
Ready to fly :rocket:

PR #13190, opened 2021-09-06
Focus on auth window password prompt before typing password.

zhaocongjia:
After GNOME40 upgrade, sometimes the auth window needs focus before typing password.

- Related ticket: https://progress.opensuse.org/issues/98171
- Needles: None
- Verification run: https://openqa.suse.de/tests/7029655#step/consoletest_finish/24

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -118,6 +118,10 @@ sub ensure_unlocked_desktop {
         if (match_has_tag('authentication-required-user-settings') || match_has_tag('authentication-required-modify-system')) {
             wait_still_screen;    # Check again as the pop-up may be just a glitch, see bsc#1168979
             if (check_screen([qw(authentication-required-user-settings authentication-required-modify-system)])) {
+                if (is_sle('>=15-SP4') || is_leap('>=15.4')) {
+                    # auth window needs focus before typing passoword for GNOME40+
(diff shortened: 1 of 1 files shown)


lib/x11utils.pm
```diff
@@ -118,6 +118,7 @@ sub ensure_unlocked_desktop {
         if (match_has_tag('authentication-required-user-settings') || match_has_tag('authentication-required-modify-system')) {
             wait_still_screen;    # Check again as the pop-up may be just a glitch, see bsc#1168979
             if (check_screen([qw(authentication-required-user-settings authentication-required-modify-system)])) {
+                assert_and_click('auth-window-password-prompt');
```
jknphy, line 121: This should fail not finding needle in product not getting new desktop? I guess you would need some condition here. Please verify for other products.

zhaocongjia: This `authentication-required-user-settings` or `authentication-required-modify-system` happens very rarely, mainly on S390x which uses xvnc, but recently another bug blocks S390x from proper finish tests.

But yes, I should add system version condition here.

zhaocongjia:
Code updated with system version check.  And Verification run for QAM 15SP3: https://openqa.suse.de/tests/7038576

PR #13192, opened 2021-09-06
Replace modules_selection & addon_products_sle when possible in YaST group

jknphy:
- Related ticket: https://progress.opensuse.org/issues/97973
- Verification run: [modules selection in YaST group](https://openqa.suse.de/tests/overview?distri=sle&build=jknphy%2Fos-autoinst-distri-opensuse%23replace_modules_selection&version=15-SP4)

For Full medium we have and issue when interacting in module selection dialog due it is using a different UI controller, instead of a richtext box we have a YMultiSelectionBox that seems to crash whatever valid rest api call we send it, so will will revisit those remaining once that is fixed


sofiasyria:
LGTM. Just a couple of thoughts. About the hide development modules checkbox, it's quite frequent that it's not there and we expect it (or shouldn't expect it, for example when GMC was out and we hadn't adjusted our tests yet). Would it make sense to have some error message displayed in case it's not there? Also, for the select multiple modules module, in general we don't want to be using test_data or vars but in that case maybe we could use the already existing var as creating different modules for each compination of add-on modules seems a bit too much.

jknphy:
> LGTM. Just a couple of thoughts. About the hide development modules checkbox, it's quite frequent that it's not there and we expect it (or shouldn't expect it, for example when GMC was out and we hadn't adjusted our tests yet). Would it make sense to have some error message displayed in case it's not there? Also, for the select multiple modules module, in general we don't want to be using test_data or vars but in that case maybe we could use the already existing var as creating different modules for each compination of add-on modules seems a bit too much.

Regarding the second question, I discussed with @OleksandrOrlov and rename the module to `select_non_conflicting_modules`, because it is precisely what it does, select as much as possible that later in pattern selection we don't see conflicts, for that reason creating individual test for each module doesn't make sense in this case due to some module depends on other modules, like development depends of desktop or some combination produces errors later in patterns.

We still have pending to create some sort of test suite that check defaults, that one should detect if the checkbox is not displayed when it should and many other things. Check defaults is out of the scope here. This PR just move the check to installer_extended which very likely will fail when we reach GMC in other things like eula languages, etc. For now we don't have better solution for that than seeing the failure (which in that test suite will continue with next test module, doesn't stop) and adapt our tests.

jknphy:
After multiple discussions offline and agree about some design choices, let's merge it and catch some missing things if any in the next build.

PR #13202, opened 2021-09-07
Push an image to Amazon ECR

ilausuch:
This test check the connectivity from the tests to the Amazon ECR
and pushes a container image

- Related ticket: https://progress.opensuse.org/issues/97715
- Verification run: http://copland.arch.suse.de/tests/225

### DIFF
--- a/lib/publiccloud/aws.pm
+++ b/lib/publiccloud/aws.pm
@@ -15,7 +15,6 @@ package publiccloud::aws;
 use Mojo::Base 'publiccloud::provider';
 use Mojo::JSON 'decode_json';
 use testapi;
-use publiccloud::utils "is_byos";
 
 use constant CREDENTIALS_FILE => '/root/amazon_credentials';
 
@@ -66,6 +65,10 @@ sub init {
         save_tmp_file(CREDENTIALS_FILE, $credentials_file);
         assert_script_run('curl -O ' . autoinst_url . "/files/" . CREDENTIALS_FILE);
     }
+
+    $self->{aws_account_id} = script_output("aws sts get-caller-identity | jq -r '.Account'");
+    die("Cannot get the UserID")                                        unless ($self->{aws_account_id});
+    die("The UserID doesn't have the correct format: $self->{user_id}") unless $self->{aws_account_id} =~ /^\d{12}$/;
 }
 
 1;
--- a/lib/publiccloud/basetest.pm
+++ b/lib/publiccloud/basetest.pm
@@ -16,6 +16,7 @@ use base 'opensusebasetest';
 use testapi;
 use publiccloud::azure;
 use publiccloud::ec2;
+use publiccloud::eks;
 use publiccloud::gce;
 use strict;
 use warnings;
@@ -27,12 +28,21 @@ sub provider_factory {
     die("Provider already initialized") if ($self->{provider});
 
     if (check_var('PUBLIC_CLOUD_PROVIDER', 'EC2')) {
-        $provider = publiccloud::ec2->new(
-            key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
-            key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
-            region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
-            username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
-        );
+        if (check_var('PUBLIC_CLOUD_SERVICE', 'EKS')) {
+            $provider = publiccloud::eks->new(
+                key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
+                key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
+                region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
+                username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
+            );
+        } else {
+            $provider = publiccloud::ec2->new(
+                key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
+                key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
+                region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
+                username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
+            );
+        }
 
     }
     elsif (check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE')) {
--- a/lib/publiccloud/eks.pm
+++ b/lib/publiccloud/eks.pm
@@ -1,0 +1,50 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Helper class for Amazon Elastic Container Registry (ECR)
+#
+# Maintainer: Ivan Lausuch <ilausuch@suse.de>, qa-c team <qa-c@suse.de>
+# Documentation: https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html
+
+package publiccloud::eks;
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+=head2 push_container_image
+Upload a container image to the ECR. Required parameter is the
+name of the image, previously stored in the local registry. And
+the tag (name) in the public cloud containers repository
+Retrieves the full name of the uploaded image or die.
+=cut
+sub push_container_image {
+    my ($self, $image, $tag) = @_;
+
+    my $region     //= $self->region();
+    my $repository //= get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REPO", 'suse-qec-testing');
+    my $aws_account_id   = $self->{aws_account_id};
+    my $full_name_prefix = "$aws_account_id.dkr.ecr.$region.amazonaws.com";
+    my $full_name        = "$full_name_prefix/$repository:$tag";
+
+    $self->{repository} = $repository;
+    $self->{tag}        = $tag;
+
+    assert_script_run("aws ecr get-login-password --region $region | docker login --username AWS --password-stdin $full_name_prefix");
+    assert_script_run("docker tag $image $full_name");
+    assert_script_run("docker push $full_name", 180);
+    return $full_name;
+}
+
+sub cleanup {
+    my $self = shift;
+    assert_script_run("aws ecr batch-delete-image --repository-name " . $self->{repository} . " --image-ids imageTag=" . $self->{tag});
+    $self->SUPER::cleanup();
+    return;
+}
+
+1;
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -633,7 +633,7 @@ This method is called called after each test on failure or success.
 =cut
 sub cleanup {
     my ($self) = @_;
-    $self->terraform_destroy();
+    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform');
     $self->vault_revoke();
     assert_script_run "cd";
 }
--- a/tests/containers/upload_container_images_to_ecr.pm
+++ b/tests/containers/upload_container_images_to_ecr.pm
@@ -1,0 +1,33 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Push a container image to the public cloud container registry
+#
+# Maintainer: Ivan Lausuch <ilausuch@suse.com>, qa-c team <qa-c@suse.de>
+
+use Mojo::Base 'publiccloud::basetest';
+use testapi;
+use containers::urls 'get_suse_container_urls';
+use mmapi 'get_current_job_id';
+
+sub run {
+    my ($self, $args) = @_;
+
+    $self->select_serial_terminal;
+
+    my ($untested_images, $released_images) = get_suse_container_urls();
+
+    my $provider = $self->provider_factory();
+    my $image    = $untested_images->[0];
+    my $tag      = join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), get_current_job_id());
+    assert_script_run("docker pull $image", 360);
+    $provider->push_container_image($image, $tag);
+}
+
+1;
--- a/variables.md
+++ b/variables.md
@@ -239,3 +239,4 @@ ZDUPREPOS | string | | Comma separated list of repositories to be added/used for
 ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in the machine (usually 0.0.fa00 and/or 0.0.fc00)
 LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are booting an installed system from the installer medium (for example, a DVD) with the menu option "Boot Linux System" (not "boot From Hard Disk"). This option uses linuxrc.
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
+PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in CSP
--- a/schedule/containers/sle_image_on_amazon_eks.yaml
+++ b/schedule/containers/sle_image_on_amazon_eks.yaml
@@ -1,0 +1,7 @@
+---
+name: sle_image_on_amazon_eks
+description: |
+  containers Amazon on Elastic Kubernetes Service test
+schedule:
+  - boot/boot_to_desktop
+  - containers/upload_container_images_to_ecr


jlausuch (changes requested):

tests/publiccloud/upload_container_images.pm
```diff
@@ -0,0 +1,85 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Push a container image to the tested public cloud container registry
```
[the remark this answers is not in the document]
ilausuch: It is relative to: Push a container image to the public cloud container registry is being tested

tests/publiccloud/upload_container_images.pm
```diff
@@ -0,0 +1,85 @@
...
+
+1;
+
+=head1 Discussion
+
+This module tests the upload of container images to the public cloud repositories.
+
+=head1 Configuration
+
+=head2 PUBLIC_CLOUD_PROVIDER
```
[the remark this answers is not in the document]
ilausuch: Then I'll move the new ones to variables.md. But do you think I should remove the others from this document? My proposal is to keep this at is it and add the new ones to variables.md. And as you said in an other PR (do we have a ticket?) move all the common variables to the .md

lib/publiccloud/aws.pm
```diff
@@ -68,4 +69,11 @@ sub init {
     }
 }
 
+sub get_user_id {
+    zypper_call "in jq";
+    my $out = script_output("aws sts get-caller-identity | jq -r '.Account'");
```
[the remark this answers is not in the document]
ilausuch: I put this here because is related directly with the AWS and not with the ECR, is something more generic that could be used in the future for other libraries that inherit from AWS one

tests/publiccloud/upload_container_images.pm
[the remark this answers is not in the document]
ilausuch: Ok I'll keep that more simple

tests/publiccloud/upload_container_images.pm
```diff
@@ -0,0 +1,85 @@
...
+use publiccloud::ecr;
+use containers::urls 'get_suse_container_urls';
+use containers::common qw(install_docker_when_needed allow_selected_insecure_registries);
+use version_utils qw(is_sle get_os_release);
+
+sub run {
+    my ($self, $args) = @_;
+    $self->select_serial_terminal;
+
+    my ($running_version, $sp, $host_distri) = get_os_release;
```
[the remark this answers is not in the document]
ilausuch: A new ticket has been created to cover this idea https://progress.opensuse.org/issues/98325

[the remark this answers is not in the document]
ilausuch: Using podman we don't need this ticket

lib/publiccloud/aws.pm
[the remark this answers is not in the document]
ilausuch: `
# echo "das" | jq
parse error: Invalid numeric literal at line 2, column 0
# echo $?
4
`

tests/publiccloud/upload_container_images.pm
```diff
@@ -0,0 +1,85 @@
...
+    install_docker_when_needed($host_distri);
+    allow_selected_insecure_registries(runtime => 'docker') if (is_sle());
+
+    my $provider = publiccloud::ecr->new(
+        key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
+        key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
+        region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
+        username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
+    );
+    $provider->init();
```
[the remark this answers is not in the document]
ilausuch: I tested that the init() is necessary because this provider is not the same created by the basetest. And as the basetest it needs to be initialized, check:

https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/4be33ca50080a9bc942e02299ad52cde8f18ab5c/lib/publiccloud/basetest.pm#L65

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -0,0 +1,43 @@
...
+    my ($untested_images, $released_images) = get_suse_container_urls();
+
+    my $image = $untested_images->[0];
+    assert_script_run("docker pull $image", 360);
+    $provider->push_container_img($image);
+}
+
+sub cleanup {
+    my ($self) = @_;
+}
```
b10n1k, line 41: is something missing here? :)

ilausuch: no, It can be removed

ilausuch: Removed, relaunching test

lib/publiccloud/aws.pm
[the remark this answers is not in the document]
cfconrad: I like the idea of checking the output, see https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html it is a 12digit...

lib/publiccloud/aws.pm
```diff
@@ -68,4 +69,11 @@ sub init {
     }
 }
 
+sub get_user_id {
+    zypper_call "in jq";
```
cfconrad, line 73: I would prefer, this call here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/prepare_tools.pm#L78

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+package publiccloud::ecr;
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+sub push_container_img {
+    my ($self, $name) = @_;
+
+    my $aws_account_id = $self->get_user_id();
+    my $region         = get_var('PUBLIC_CLOUD_REGION',                'eu-central-1');
+    my $repository     = get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REPO", 'suse-qec-testing');
```
cfconrad, line 24: I would try to use the openqa variables in the test and pass here only parameters.
```suggestion
    my ($self, $name, %args) = @_;
    $args{region} //= $self->region();
    $args{repo} //= get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REPO", 'suse-qec-testing');

    my $aws_account_id = $self->get_user_id();
```

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+    my $tag            = (split /-/, get_required_var("NAME"))[0];
```
cfconrad, line 25: Using `get_required_var('NAME')` here is not a good Idea, I would like to avoid this!

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+    my $full_name      = "$aws_account_id.dkr.ecr.$region.amazonaws.com/$repository:$tag";
+
+    $self->{repository} = $repository;
+    $self->{tag}        = $tag;
+
+    assert_script_run("aws ecr get-login-password --region $region | docker login --username AWS --password-stdin $aws_account_id.dkr.ecr.$region.amazonaws.com");
```
cfconrad, line 31:
```suggestion
    $full_name_prefix = "$aws_account_id.dkr.ecr.$region.amazonaws.com";
    $full_name = "$full_name_prefix/$repository:$tag";
    assert_script_run("aws ecr get-login-password --region $region | docker login --username AWS --password-stdin $full_name_prefix");
```

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+# Summary: Helper class for Amazon Elastic Container Registry (ECR)
+#
+# Maintainer: Ivan Lausuch <ilausuch@suse.de>, qa-c team <qa-c@suse.de>
+# Documentation: https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html
+
+package publiccloud::ecr;
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+sub push_container_img {
```
cfconrad, line 19: Maybe you should add a stub in `publiccloud::provider` as well like: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L92

cfconrad: If this will be a method, which should be implemented by other container implementations...

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+    assert_script_run("docker push $full_name");
+    return;
+}
+
+sub cleanup {
+    my $self = shift;
+
+    assert_script_run("aws ecr batch-delete-image --repository-name " . $self->{repository} . " --image-ids imageTag=" . $self->{tag});
+    return;
+}
```
cfconrad, line 42: you overwrite this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L610 cleanup method here.
So you forget to call `revoke_vault()`.

```suggestion
sub cleanup {
    my $self = shift;

    assert_script_run("aws ecr batch-delete-image --repository-name " . $self->{repository} . " --image-ids imageTag=" . $self->{tag});
    $self->SUPER::cleanup();
}
```

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -0,0 +1,39 @@
...
+    my ($self, $args) = @_;
+    $self->select_serial_terminal;
+
+    my $provider = publiccloud::ecr->new(
+        key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
+        key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
+        region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
+        username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
+    );
+
```
cfconrad, line 29:
```suggestion
$self->{provider} = $provider;

```
Or better use the [provider_factory()](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/basetest.pm#L23) somehow.

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -0,0 +1,39 @@
...
+    $provider->init();
+
+    my ($untested_images, $released_images) = get_suse_container_urls();
+
+    my $image = $untested_images->[0];
+    assert_script_run("docker pull $image", 360);
+    $provider->push_container_img($image);
+}
```
cfconrad, line 37: This test is really nice readable!

variables.md
```diff
@@ -179,3 +179,4 @@ ZDUPREPOS | string | | Comma separated list of repositories to be added/used for
 ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in the machine (usually 0.0.fa00 and/or 0.0.fc00)
 LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are booting an installed system from the installer medium (for example, a DVD) with the menu option "Boot Linux System" (not "boot From Hard Disk"). This option uses linuxrc.
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
+PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository where the container image are stored
```
cfconrad, line 182: Just curious, does this repo somehow be created in advance?

cfconrad (changes requested):

lib/publiccloud/aws.pm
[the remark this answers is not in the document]
cfconrad: And I would go one step further, that we add a member to `publiccloud::aws` called account_id and set it on `login()`.

lib/publiccloud/ecr.pm
ilausuch: Why not is a good idea? I am interested. Anyway, I need an unique id for each job. Please, tell me if you know an other way

lib/publiccloud/ecr.pm
ilausuch: Good idea

variables.md
ilausuch: Yes, I did and it should be created in advance, but I will investigate if there is a way to create the repo directly during the test execution.

lib/publiccloud/ecr.pm
pdostal: why don't we use `lib/containers/urls.pm`?

tests/containers/upload_container_images_to_ecr.pm
pdostal: I like it too!

variables.md
pdostal: What does this repo mean?

lib/publiccloud/aws.pm
asmorodskyi: ah I wanted to write the same :) so just +1

variables.md
ilausuch: This is the container repository in the public cloud where the images will be uploaded

asmorodskyi:
```suggestion
PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in Public Cloud
```

I think it is obvious that Container Images Repo is used to store container images :)

lib/publiccloud/ecr.pm
ilausuch: Sorry, I don't understand what you mean. The problem I would like to avoid is to have two jobs uploading the container image with the same name, because one of these jobs would fail. And would have false positives

cfconrad: 1) I would pass the name as parameter to the function (with proper default) and let the name by the caller
2) If we use the same image in multiple jobs, it would be nice to reuse it. I think it is time consuming to upload these images. We do the same with vm images in CSP. And then we have a cleanup mechanize in PCW.

can you make an example of the currently used name? Maybe I'm wrong with my assumptions.

lib/publiccloud/aws.pm
ilausuch: I created this other ticket https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13252

lib/publiccloud/ecr.pm
cfconrad: Now I see, you only extract the job-id. If this image needs to be unique, then you should be unique by openqa instance.
Maybe you could use:
```perl
join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), mmapi::get_current_job_id());
```
But still, I would do it in `tests/containers/upload_container_images_to_ecr.pm`.

tests/containers/upload_container_images_to_ecr.pm
ilausuch: About using the provider factory, the problem is that we have (EC2) it should be changed to ECR but, we'll have more that one provider, one per amazon service and for the lasts tests (not this one) we'll need two ECR + EKS, and maybe more in the future. A simple solution is to have all in the EC2 but is a mess and we created the super class for AWS to avoid that. Options:
- We can add other option to PUBLIC_CLOUD_PROVIDER with the name ECR but we have to imagine what we have to do if we need to use the EKS
- We can follow the option 1 and put ECR + EKS together in the same class, then we should change the name of ecr.pm to something like ecr-eks.pm
- The test doesn't have the publiccloud::basetest base. I don't know if this is possible without having problems
- Or we keep the additional providers approach

For the next commit I will use your suggested solution

tests/containers/upload_container_images_to_ecr.pm
ilausuch: Thanks

lib/publiccloud/ecr.pm
ilausuch: Understood, good idea

lib/publiccloud/aws.pm
```diff
@@ -68,4 +69,11 @@ sub init {
     }
 }
 
+sub get_user_id {
```
asmorodskyi, line 72: so in my understanding this function will be dropped ? ( it contains of 2 lines and one goes to prepare_tools module and another in to `publiccloud::aws` init code )

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Helper class for Amazon Elastic Container Registry (ECR)
+#
+# Maintainer: Ivan Lausuch <ilausuch@suse.de>, qa-c team <qa-c@suse.de>
+# Documentation: https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html
+
+package publiccloud::ecr;
```
asmorodskyi, line 15: we have a [**test**](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/upload_image.pm) which is responsible for uploading images . Depending on provider set via variables it initialize appropriate provider class and do upload . I think it should be done same way here .

schedule/containers/sle_image_on_sle_ecr.yaml
```diff
@@ -0,0 +1,7 @@
+---
+name: sle_image_on_publiccloud
```
asmorodskyi, line 2: I find this naming a bit confusing , all other scenarios named like this meant to show host/guest OS'es . here situation is a bit different SLE ECR is not OS . I would come up with something else . 
Actually this goes together with this comment https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13202/files#r708267661 . If you accept it this scenario will have name `upload_container_image`

jlausuch, line 2:
```suggestion
name: sle_image_on_ecr
```

jlausuch: File name `sle_image_on_amazon_ecr` better? Now it seems `ecr` is something from `sle` which is wrong.

ilausuch: ok, thanks

asmorodskyi: yeah I also have concerns regarding this . even more in all other cases naming convention is `<container OS>_on_<Host OS>`  so `amazon ecr` is better than `sle ecr ` but still it is not OS . I would actually go with approach which we have for uploading "normal" images . where we have `upload_image` scenario which based on variables provided call appropriate provider function for upload. So in this case it will be `upload_container_image`  . WDYT ?

variables.md
ilausuch: done

lib/publiccloud/ecr.pm
ilausuch: done

lib/publiccloud/ecr.pm
```diff
@@ -0,0 +1,44 @@
...
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+sub push_container_img {
+    my ($self, $name) = @_;
+
+    my $aws_account_id = $self->get_user_id();
+    my $region         = get_var('PUBLIC_CLOUD_REGION',                'eu-central-1');
```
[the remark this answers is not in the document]
ilausuch: done

lib/publiccloud/ecr.pm
ilausuch: done

lib/publiccloud/ecr.pm
ilausuch: done

lib/publiccloud/ecr.pm
ilausuch: done

schedule/containers/sle_image_on_sle_ecr.yaml
ilausuch: done, I changed by sle_image_on_amazon_ecr that includes the name of amazon

lib/publiccloud/aws.pm
ilausuch: done

lib/publiccloud/aws.pm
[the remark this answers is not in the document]
ilausuch: done, but it still pendent to remove the installation of jq

lib/publiccloud/ecr.pm
pdostal: Wait, can you @ilausuch give me an example of the content of `NAME` variable?

cfconrad:
```
NAME=00000200-sle-15-SP3-Container-Image-Updates-x86_64-Build1-sle-15-SP3_image_on_ECR@64bit
```

tests/containers/upload_container_images_to_ecr.pm
cfconrad: What about `get_var('PUBLIC_CLOUD_AWS_SERVICE', 'ec2')` and then instantiate one or the other class.

lib/publiccloud/ecr.pm
pdostal: OK, Good! Thank you :)

tests/containers/upload_container_images_to_ecr.pm
ilausuch: Let's see now. It's in the new commit

ilausuch:
- I updated the last commit: Talking with @jlausuch about what tests will contain the new job group, we found that is better to change the name to the current objective of these tests to sle_image_on_amazon_eks

- I created a new commit with the idea of have only one provider in the test. Therefore we maintain the current variable PUBLIC_CLOUD_PROVIDER to select and the service is selected by the variable PUBLIC_CLOUD_SERVICE: "EKS", So the provider in the basetest will be the eks instead of ec2 for this case

ilausuch:
This PR is ready for review after the last changes and have a valid VR

lib/publiccloud/eks.pm
```diff
@@ -0,0 +1,50 @@
...
+use testapi;
+
+=head2 push_container_imgage
+Upload a container image to the ECR. Required parameter is the
+name of the image, previously stored in the local registry. And
+the tag (name) in the public cloud containers repository
+Retrieves the full name of the uploaded image or die.
+=cut
+sub push_container_imgage {
+    my ($self, $image, $tag, %arg) = @_;
```
b10n1k, line 26: nitpicking but do you really need %arg. it is not used. and if for some reason you have to leave it maybe better rename it to `%args`. but not a problem if you ignore

ilausuch: Good catch! This is a residue from where I was testing possibilities.
Done

ilausuch:
As soon you review that I could squash the commits

lib/publiccloud/aws.pm
```diff
@@ -66,6 +67,10 @@ sub init {
         save_tmp_file(CREDENTIALS_FILE, $credentials_file);
         assert_script_run('curl -O ' . autoinst_url . "/files/" . CREDENTIALS_FILE);
     }
+
+    $self->{aws_account_id} = script_output("aws sts get-caller-identity | jq -r '.Account'");
+    die("Cannot get the UserID")                                        unless ($self->{aws_account_id});
+    die("The UserID doesn't have the correct format: $self->{user_id}") unless length($self->{aws_account_id}) == 12;
```
cfconrad, line 73: I think it was 12 digits but this is nitty!
```suggestion
    die("The UserID doesn't have the correct format: $self->{user_id}") unless $self->{aws_account_id} =~ /^\d{12}$;
```

cfconrad (approved):
LGTM!

lib/publiccloud/aws.pm
```diff
@@ -16,6 +16,7 @@ use Mojo::Base 'publiccloud::provider';
 use Mojo::JSON 'decode_json';
 use testapi;
 use publiccloud::utils "is_byos";
+use utils "zypper_call";
```
pdostal, line 19: Do we use `zypper_call` in this library?

pdostal (approved):
Nice! :rocket:

lib/publiccloud/aws.pm
jlausuch: Looks like some left over. Can you remove it?

lib/publiccloud/eks.pm
```diff
@@ -0,0 +1,50 @@
...
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+=head2 push_container_imgage
+Upload a container image to the ECR. Required parameter is the
+name of the image, previously stored in the local registry. And
+the tag (name) in the public cloud containers repository
+Retrieves the full name of the uploaded image or die.
+=cut
+sub push_container_imgage {
```
jlausuch, line 25:
```suggestion
sub push_container_image {
```

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -0,0 +1,33 @@
...
+
+    $self->select_serial_terminal;
+
+    my ($untested_images, $released_images) = get_suse_container_urls();
+
+    my $provider = $self->provider_factory();
+    my $image    = $untested_images->[0];
+    my $tag      = join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), get_current_job_id());
+    assert_script_run("docker pull $image", 360);
+    $provider->push_container_imgage($image, $tag);
```
jlausuch, line 30:
```suggestion
    $provider->push_container_image($image, $tag);
```

lib/publiccloud/eks.pm
```diff
@@ -0,0 +1,50 @@
...
+# Summary: Helper class for Amazon Elastic Container Registry (ECR)
+#
+# Maintainer: Ivan Lausuch <ilausuch@suse.de>, qa-c team <qa-c@suse.de>
+# Documentation: https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html
+
+package publiccloud::eks;
+use Mojo::Base 'publiccloud::aws';
+use testapi;
+
+=head2 push_container_imgage
```
jlausuch, line 19:
```suggestion
=head2 push_container_image
```

lib/publiccloud/eks.pm
```diff
@@ -0,0 +1,50 @@
...
+Upload a container image to the ECR. Required parameter is the
+name of the image, previously stored in the local registry. And
+the tag (name) in the public cloud containers repository
+Retrieves the full name of the uploaded image or die.
+=cut
+sub push_container_imgage {
+    my ($self, $image, $tag) = @_;
+
+    my $region     //= $self->region();
```
jlausuch, line 28: /sarcasm mode on: perltidy is so funny... it doesn't complain that this is not in perfect alignment with lines below :D

lib/publiccloud/eks.pm
ilausuch: Oh! typo

jlausuch (commented):
Looks very good already. Some minor typos to correct.

lib/publiccloud/aws.pm
ilausuch: And `use publiccloud::utils "is_byos";` is not used neither

ilausuch: done

lib/publiccloud/eks.pm
ilausuch: done and related ones

lib/publiccloud/aws.pm
ilausuch: done

ilausuch:
I am waiting for VR

ilausuch:
VR passed. I pasted the link

grisu48 (approved):

jlausuch:
Your test fails in `cleanup` http://copland.arch.suse.de/tests/221#step/upload_container_images_to_ecr/81 
because it calls [`$self->terraform_destroy();`](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L636) when there hasn't been any terraform action, so you need to prevent calling this function for EKS case (any any other not using terraform).

ilausuch:
> Your test fails in `cleanup` http://copland.arch.suse.de/tests/221#step/upload_container_images_to_ecr/81
> because it calls [`$self->terraform_destroy();`](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L636) when there hasn't been any terraform action, so you need to prevent calling this function for EKS case (any any other not using terraform).

I solve the problem avoiding terraform_destron in EKS case

lib/publiccloud/basetest.pm
```diff
@@ -27,12 +28,21 @@ sub provider_factory {
...
-            username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
-        );
+        if (check_var('PUBLIC_CLOUD_SERVICE', 'EKS')) {
+            $provider = publiccloud::eks->new(
+                key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
+                key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
+                region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
+                username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
+            );
+        } else {
```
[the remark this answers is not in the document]
ilausuch: I understand. Maybe this is better, but I don't like too much indexations. What do you think?

`if (check_var('PUBLIC_CLOUD_PROVIDER', 'EC2')) {
        if (get_var('PUBLIC_CLOUD_SERVICE')) {
            if (check_var('PUBLIC_CLOUD_SERVICE', 'EKS')) {
                $provider = publiccloud::eks->new(
                    key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
                    key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
                    region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
                    username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
                );
            }
			else { die "Invalid PUBLIC_CLOUD_SERVICE value"; }
        } else {
            $provider = publiccloud::ec2->new(
                key_id     => get_var('PUBLIC_CLOUD_KEY_ID'),
                key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
                region     => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1'),
                username   => get_var('PUBLIC_CLOUD_USER',   'ec2-user')
            );
        }
    }`

lib/publiccloud/provider.pm
```diff
@@ -609,7 +609,7 @@ This method is called called after each test on failure or success.
 =cut
 sub cleanup {
     my ($self) = @_;
-    $self->terraform_destroy();
+    $self->terraform_destroy() unless get_var('PUBLIC_CLOUD_SERVICE') ~~ ["EKS"];
```
jlausuch, line 612: why not?

```suggestion
    $self->terraform_destroy() unless check_var('PUBLIC_CLOUD_SERVICE', 'EKS');
```

ilausuch: The idea is to have an array of excepted services. Is expected we'll have one for each public cloud.

ilausuch: The idea is to have an array of excepted services. Is expected we'll have one for each public cloud.

jlausuch: Ok, makes sense. Anyway, in a later PR we could improve the cleanup function based on @cfconrad idea: not running terraform destroy if terraform directory exists `script_run('test -d ' . TERRAFORM_DIR . '/.terraform');`

jlausuch (approved):

lib/publiccloud/provider.pm
jlausuch: Seems that pipeline complains about this. I haven't never seen this way of checking if "string is in array" in perl and I am not perl expert, but what I would probably do is:

`unless (grep $_ eq get_var('PUBLIC_CLOUD_SERVICE'), ('X', 'Y', 'Z')) {`

@cfconrad wdyt?

ilausuch: I implemented the first solution and the VR is green.

ilausuch:
I increased the timeout when the image is pushed. See http://copland.arch.suse.de/tests/223

lib/publiccloud/provider.pm
```diff
@@ -633,7 +633,7 @@ This method is called called after each test on failure or success.
 =cut
 sub cleanup {
     my ($self) = @_;
-    $self->terraform_destroy();
+    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform');
```
ldevulder, line 636: Why is this needed? It breaks the HA/SAP cleanup....

ilausuch: we understood that if there is not terraform directory the terraform_destroy command will fail

ilausuch: proposed solution https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13328

PR #13183, opened 2021-09-03
Add GCP super class and move conn and auth from GCE lib

ilausuch:
Moves the common functionality (connection and auth) from GCE lib
to a super class to extend the functionality of tests over other
Google cloud services

- Related ticket: https://progress.opensuse.org/issues/97886
- Verification run: http://openqa.suse.de/tests/7014363

PR #13204, opened 2021-09-07
publiccloud: follow up fix for #13183

asmorodskyi:
Fixing minor issue caused by #13183 . 
Example of the issue https://openqa.suse.de/tests/7036605#step/img_proof/2 

VR: http://autobot.qa.suse.de/tests/2545


jlausuch (approved):
LGTM

PR #13209, opened 2021-09-08
Add sle-15-SP4 to the cluster_setup schedule

bs-suse:
Add sle-15-SP4 to the cluster_setup schedule

Would be nice to make this a wildcard setting, but I'm not sure how or even if thats possible.


alvarocarvajald (approved):
LGTM. But 15 commits?

ldevulder:
LGTM too, but yes rebase it before we can merge ;-)

PR #13210, opened 2021-09-08
Enable rules_and_classes for openSUSE

rakoenig:
Enable AutoYaST rules_and_classes for openSUSE

- Related ticket: https://progress.opensuse.org/issues/97874
- Verification run: https://openqa.opensuse.org/tests/1906987

### DIFF
--- a/data/autoyast_opensuse/rule-based_example/classes/general/software.xml
+++ b/data/autoyast_opensuse/rule-based_example/classes/general/software.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0"?>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <software>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <products config:type="list">
+      <product>openSUSE</product>
+    </products>
+    <patterns config:type="list">
+      <pattern>apparmor</pattern>
+      <pattern>base</pattern>
+      <pattern>gnome</pattern>
+    </patterns>
+    <packages config:type="list">
+      <package>apache2</package>
+    </packages>
+  </software>
+</profile>
--- a/data/autoyast_opensuse/rule-based_example/classes/general/users.xml
+++ b/data/autoyast_opensuse/rule-based_example/classes/general/users.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <login_settings>
+    <autologin_user>bernhard</autologin_user>
+  </login_settings>
+  <users config:type="list">
+    <user>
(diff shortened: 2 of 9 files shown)


data/autoyast_opensuse/rule-based_example/classes/general/software.xml
```diff
@@ -2,9 +2,15 @@
 <!DOCTYPE profile>
```
jknphy, line 2: xml accidentally ended up in this commit.

schedule/yast/autoyast/autoyast_opensuse_rules_and_classes.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - console/validate_partition_table_via_blkid
+  - console/validate_blockdevices
+  - console/zypper_lr
+  - console/zypper_ref
+  - console/ncurses
+  - update/zypper_up
+  - console/zypper_in
+  - console/zypper_log
+  - console/orphaned_packages_check
+  - autoyast/verify_cloned_profile
```
jknphy, line 29: please, exclude all test that are using test data (they are not doing anything silently), so basically:
```
...
- installation/first_boot
- console/zypper_lr
- console/zypper_ref
- console/ncurses
- update/zypper_up
- console/zypper_in
- console/zypper_log
```
But if you remove everything after first_boot also is ok. Could you create a ticket with low priority to remind us to do proper validation with this test suites?

data/autoyast_opensuse/rule-based_example/classes/general/users.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
jknphy, line 1: also please do squash with the corresponding commit, so all the changes to xml will be in the same commit. We don't need to remark the differences between openSUSE and SLE via commits, if that was the case we would prefer to share code between them, so I guess the best is having in one commit.

jknphy (commented):
LGTM except for those minors thing. Good job and investigation!

Commits are very clear and you followed well how to name it, specially the first one, we also apply 'somehow' that to the PR title, so instead of `Rules opensuse` we could find something more verbose (for example, it helps avoiding to trigger other collaborators thinking that some changes are something that at the end are not related with their domain of knowledge), For example, we could reuse the title in your commit: `Enable rules_and_classes for openSUSE `. It contains a verb, it starts by capital letter and describe with not much characters what is about, so I guess it is a good one. The other one you could add 'Add' in front, 'Add YAML file...'.

data/autoyast_opensuse/rule-based_example/classes/general/users.xml
rakoenig: Did all the editiing and rebased it so that it is just one commit.

schedule/yast/autoyast/autoyast_opensuse_rules_and_classes.yaml
jknphy: probably you closed the conversation accidentally with the squashing.

PR #13213, opened 2021-09-08
console/firewalld: Fix call to is_leap

Vogtinator:
I fell into the trap again: Unlike is_sle, it always needs the minor version.

- Verification run: https://openqa.opensuse.org/tests/1907653


Vogtinator:
VR passed

PR #13216, opened 2021-09-09
Fix tracker failures on SLE15SP4

GraceWang571:
Adjust the condition to handle different tracker versions
Add a softfailure for tracker search due to bsc#1190296

- Related ticket: https://progress.opensuse.org/issues/97940
- Needles: already added when debugging on osd
- Verification run: 

Updated on 2021-09-15
> SLE15SP4 https://openqa.suse.de/tests/7097788
> TW x86_64 https://openqa.opensuse.org/tests/1914765
> TW aarch64 https://openqa.opensuse.org/tests/1914766

### DIFF
--- a/tests/x11/tracker/tracker_by_command.pm
+++ b/tests/x11/tracker/tracker_by_command.pm
@@ -32,9 +32,16 @@ sub run {
         script_run "tracker-search newfile";
     }
     else {
-        my $trackercmd = (is_sle('<16') or is_leap('<16.0')) ? 'tracker' : 'tracker3';
+        my $trackercmd = (is_sle('<=15-sp3') or is_leap('<=15.3')) ? 'tracker' : 'tracker3';
         script_run "$trackercmd search emptyfile";
-        assert_screen('tracker-cmdsearch-emptyfile');
+        assert_screen([qw(tracker-cmdsearch-emptyfile tracker-cmdsearch-noemptyfile)]);
+        if (match_has_tag 'tracker-cmdsearch-noemptyfile') {
+            record_soft_failure 'bsc#1190296 tracker3 can not index empty files for the first time';
+            enter_cmd "clear";
+            wait_still_screen 10;
+            assert_script_run "$trackercmd search emptyfile";
+            assert_screen 'tracker-cmdsearch-emptyfile';
+        }
         script_run "$trackercmd search newfile";
     }
     assert_screen 'tracker-cmdsearch-newfile';
--- a/tests/x11/tracker/tracker_info.pm
+++ b/tests/x11/tracker/tracker_info.pm
@@ -31,7 +31,7 @@ sub run {
         script_run "tracker-info newpl.pl";
     }
     else {
-        my $trackercmd = (is_sle('<16') or is_leap('<16.0')) ? 'tracker' : 'tracker3';
+        my $trackercmd = (is_sle('<=15-sp3') or is_leap('<=15.3')) ? 'tracker' : 'tracker3';
         script_run "$trackercmd info newpl.pl";
     }
     assert_screen 'tracker-info-newpl';


tests/x11/tracker/tracker_by_command.pm
```diff
@@ -32,9 +32,12 @@ sub run {
...
     }
     else {
-        my $trackercmd = (is_sle('<16') or is_leap('<16.0')) ? 'tracker' : 'tracker3';
+        my $trackercmd = (is_sle('<=15-sp3') or is_leap('<=15.3')) ? 'tracker' : 'tracker3';
         script_run "$trackercmd search emptyfile";
-        assert_screen('tracker-cmdsearch-emptyfile');
+        assert_screen([qw(tracker-cmdsearch-emptyfile tracker-cmdsearch-noemptyfile)]);
+        record_soft_failure 'bsc#1190296 tracker3 can not index empty files for the first time.' if check_screen 'tracker-cmdsearch-noemptyfile', 30;
+        # Wait 20s for tracker to index the test file
+        wait_still_screen 20;
```
okurz, line 40:
```suggestion
        record_soft_failure 'bsc#1190296 tracker3 can not index empty files for the first time.' if match_has_tag 'tracker-cmdsearch-noemptyfile', 30;
        assert_screen '…';
```

with a proper needle tag for `assert_screen`

GraceWang571: I'm not sure if I got your meaning correctly. 
Do you mean if we meet the bug bsc#1190296, we need to execute the below code one more time?
`script_run "$trackercmd search emptyfile";
assert_screen"tracker-cmdsearch-emptyfile";`

GraceWang571: If simply add `assert_screen` here I think it's meaningless. 
Since it has `assert_screen([qw(tracker-cmdsearch-emptyfile tracker-cmdsearch-noemptyfile)]);` in front of the `record_soft_failure` and also has `assert_screen` for the `tracker3 search newfile` after it.

20s is used to give the tracker enough time to index the test file so that the following tests won't fail due to https://bugzilla.suse.com/show_bug.cgi?id=1190296

okurz: Well, the `wait_still_screen` can be equivalent to an `assert_screen` looking for the right screen with the same timeout value. Only that with `assert_screen` one does not waste time as soon as the right needle has been found. I don't know what needs to be done if the needle `tracker-cmdsearch-emptyfile` can not be found initially. It sounds like a good idea to re-execute the search command if it fails the first time assuming that it will work on the second run. Just think of what a human user would do. A human user would likely not blindly wait 20 seconds and just go on after the waiting time. A human user would likely look on the screen and depending on what is shown there do an according action. The according action can be waiting but then there would be the early-exit criteria of stopping to wait as soon as the expected result shows. So likely the following would be correct:

```perl
    assert_screen([qw(tracker-cmdsearch-emptyfile tracker-cmdsearch-noemptyfile)]);
    if (match_has_tag 'tracker-cmdsearch-noemptyfile') {
        record_soft_failure 'bsc#1190296 tracker3 can not index empty files for the first time';
        assert_script_run "$trackercmd search emptyfile";
        assert_screen 'tracker-cmdsearch-emptyfile';
    }
```

GraceWang571: Yeah, I totally agree with you and that's also the reason why I asked in my first comment.
If the needle `tracker-cmdsearch-emptyfile` can be found initially then this step of the test is PASSED.

Thanks for your detailed explanation.

okurz: Good, great. I guess you need to trigger new verification runs for that still.

GraceWang571: I am afraid I need to take ` wait_still_screen 20;` back since it takes time for the tracker to index the test file.
Without waiting 20s,  execute `assert_script_run "$trackercmd search emptyfile";` immediately, the needle tracker-cmdsearch-emptyfile still can't be found. See https://openqa.opensuse.org/tests/1912906#step/tracker_by_command/10
I will try if 10s or less is enough.

okurz: Please try to wait properly for what you actually want. Is this something that would show up in log files or maybe run some command in a loop to check if files are indexed?

tests/x11/tracker/tracker_by_command.pm
```diff
@@ -32,9 +32,16 @@ sub run {
         script_run "tracker-search newfile";
     }
     else {
-        my $trackercmd = (is_sle('<16') or is_leap('<16.0')) ? 'tracker' : 'tracker3';
+        my $trackercmd = (is_sle('<=15-sp3') or is_leap('<=15.3')) ? 'tracker' : 'tracker3';
         script_run "$trackercmd search emptyfile";
-        assert_screen('tracker-cmdsearch-emptyfile');
+        assert_screen([qw(tracker-cmdsearch-emptyfile tracker-cmdsearch-noemptyfile)]);
+        if (match_has_tag 'tracker-cmdsearch-noemptyfile') {
```
Vogtinator, line 38: Is this 100% reproducible? It might be a race condition.

tests/x11/tracker/tracker_by_command.pm
Vogtinator: My suspicion is that on older versions and TW, tracker is running in the background and listens for filesystem changes and is quick enough for the openQA test, but with 15 SP4 for some reason it's not, and so the `tracker3 search` command doesn't have that data yet.

tests/x11/tracker/tracker_by_command.pm
GraceWang571: The "tracker-cmdsearch-noemptyfile" can't be reproduced 100% if you restart the test suite multiple times.

From https://openqa.opensuse.org/tests/1913344#next_previous and https://openqa.opensuse.org/tests/1913431#next_previous, 
we can see it happens on TW aarch64 Wayland more often than x11.

tests/x11/seahorse.pm
```diff
@@ -43,7 +44,8 @@ sub run {
     }
     assert_screen "seahorse-password-dialog";         # Dialog "Passphrase for the new keyring"
     type_password;                                    # Users password (for auto unlock, it has to be the same)
-    send_key "ret";                                   # Next field (confirm PW)
+    my $confirm_field_key = (is_sle('<=15-sp3') or is_leap('<=15.3')) ? 'ret' : 'tab';
```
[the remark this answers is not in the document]
GraceWang571: Cool, thanks a lot for pointing this out.
Update the code to correct my mistake. I will update the verification runs later.
Just found the seahorse test case is only enabled for SLE current now.
Will consider enabling it for TW in the future.

tests/x11/tracker/tracker_by_command.pm
GraceWang571: Actually, this is not a new issue that happens only on 15SP4 or the latest TW, there was a similar or same bug reported on SLE15 https://bugzilla.suse.com/show_bug.cgi?id=1074582.

GraceWang571: > Please try to wait properly for what you actually want. Is this something that would show up in log files or maybe run some command in a loop to check if files are indexed?

I set it to 10s for now, but I am not sure if it's enough since issue bsc#1190296 can't be reproduced 100%
I will ask Dev if there's any log file that can be checked to avoid wasting time to wait.

tests/x11/tracker/tracker_by_command.pm
Vogtinator: I might be able to explain this. In the wayland test, the `flatpak` module fails, so it reverts back to the VM state at `consoletest_setup`. At that point, tracker is probably not up and running.

If this theory is correct, the failure should disappear once the `flatpak` failure is fixed.

tests/x11/seahorse.pm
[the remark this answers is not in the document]
GraceWang571: Don't know why the verification runs weren't updated successfully yesterday.
Please see the description for the latest VRs.

tests/x11/tracker/tracker_by_command.pm
GraceWang571: Sounds reasonable and here is another evidence for your theory https://openqa.opensuse.org/tests/1912851.
And that also explains why the tracker test case didn't get "tracker-cmdsearch-noemptyfile" after I added `EXCLUDE_MODULES=flatpal` when cloning the job.

Since there's no update for the flatpak bug for more than 5 months, I'd like to say we keep using this softfailure

tests/x11/tracker/tracker_by_command.pm
GraceWang571: I've brought this topic in the Desktop team meeting and Dev may provide more insight on this. Before that, I will merge this PR.

Here is a verification run that shows 10s works fine https://openqa.opensuse.org/tests/1923348

Vogtinator: > Actually, this is not a new issue that happens only on 15SP4 or the latest TW, there was a similar or same bug reported on SLE15 https://bugzilla.suse.com/show_bug.cgi?id=1074582.

That bug report was invalid FWICT. The openQA test had a typo which was fixed with 519188b251d43c62ca51d984b9fb4a5ab77a3142.

tests/x11/tracker/tracker_by_command.pm
Vogtinator: > Since there's no update for the flatpak bug for more than 5 months, I'd like to say we keep using this softfailure

I'd rather see the root cause worked around, which is either making sure that tracker is started, or waiting until tracker indexed that file. `wait_still_screen` does neither, it's just waiting some random time.

PR #13217, opened 2021-09-09
LTP: Warn on missing symlink of kernel config in /boot

pevik:
This further improves 2c6ead199.

Fixes: 2c6ead199 ("LTP: Warn on missing symlink of kernel config in /boot")

Verification run: http://quasar.suse.cz/tests/7112#step/install_ltp/89


pevik:
@mdoucha @richiejp @czerw I see soft failure for kernel config was a bit aggressive. I overlooked it's not only for install_ltp, but also for tests.

PR #13220, opened 2021-09-09
Fix rules_and_classes for SLE15

rakoenig:
Some small patches for rules_and_classes on SLE

- changed match rules in rules.xml because of this bug:
  https://bugzilla.suse.com/show_bug.cgi?id=1190245
- if rules_and_classes is used, then test_ayp_url appends
  "rules/rules.xml" to the URL to be tested if its reachable.
  The command server would respond with 404 if we want to
  access a direcory.

- Related ticket: https://progress.opensuse.org/issues/97871
- Verification run: https://openqa.suse.de/tests/7052844

### DIFF
--- a/data/autoyast_sle15/rule-based_example/rules/rules.xml
+++ b/data/autoyast_sle15/rule-based_example/rules/rules.xml
@@ -13,8 +13,8 @@
     </rule>
     <rule>
       <disksize>
-        <match>/dev/vda 10000-19000</match>
-        <match_type>range</match_type>
+        <match>/dev/vda 10000</match>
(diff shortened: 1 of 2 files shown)


jknphy:
seems that you just need to tidy your code with [perl-Perl-Tidy](https://software.opensuse.org/package/perl-Perl-Tidy). I'm using devel-language-perl to get it (just add the repo, zyppper in and disable the repo) but you need to find the most suitable for you.

rakoenig:
Another verification run: https://openqa.suse.de/tests/7099158

PR #13221, opened 2021-09-09
Move test_data and schedules for opensuse to separate folder

OleksandrOrlov:
Currently, the same test_data files are used for SLE and opensuse. It is hard to predict when some change is made in the test data, how the test will behave on different products.

While making bulk change in yast test_data it requires a lot of attention to not introduce unnecessary changes in opensuse test_data.

This only affects the test_data that are located in `test_data/yast/` folder.

Related ticket: https://progress.opensuse.org/issues/95748
Verification runs: not needed as there is no changes in files. Only changes in Job Group settings should be made.

Important: `test_data/yast/` should be replaced with `test_data/yast/opensuse/` in all the Job Group settings on o3. I kept the same directory structure inside of `test_data/yast/opensuse/` as in `test_data/yast/`, so the replacement should be easy.


okurz (commented):
Seems like you not only move data but also create new schedule files. How do you think it's possible to long-term maintain all this data?

OleksandrOrlov:
> Seems like you not only move data but also create new schedule files. How do you think it's possible to long-term maintain all this data?

This PR initially contained too much test_data just to show one of the proposals and discuss it more closely. In the end I just moved the test_data files that had "_opensuse" suffixes to the "opensuse" folder. So, it is just grooming changes, to have all the test _data related to openSUSE in the same folder and allow to easily manage them.

Regarding schedules, it is similar procedure with one minor exception: if the same schedule is used for SLE and openSUSE, I cloned it for opensuse and put in the appropriate folder. It is needed to more easily identify the schedules that are used for openSUSE, as we have a plan for future to implement and update for yaml schedule to have less amount of duplications, but for now we need to make some of them in order to know how many schedules we have and where.

okurz:
> Regarding schedules [...] for now we need to make some of them in order to know how many schedules we have and where.

yeah, I don't think this will fly but ok, I will trust you :)

+1

PR #13225, opened 2021-09-09
Containers: remove BCI tests due to scheduling limitations

jlausuch:
The addition of this new test is supposed to be executed only
for Container-Image-Updates, not for Server-DVD-Updates.
The problem is that the yaml schedule already has one condition
and we can't have double conditions in yaml... so let's just
remove it for now since it's breaking a lot of tests in maintenance.

broken tests: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=20210909-2&groupid=369

### DIFF
--- a/tests/containers/bci_tests.pm
+++ b/tests/containers/bci_tests.pm
@@ -69,6 +69,11 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
+    if (get_var('HOST_VERSION') !~ '15-SP3') {
+        record_info('poo#98183', 'Test not supported yet on this host.');
+        return;
+    }
+
     my $runtime        = get_required_var('CONTAINER_RUNTIME');
     my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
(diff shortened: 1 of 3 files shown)


jlausuch:
VR: (Server-DVD-Updates -> not scheduled) : https://openqa.suse.de/tests/7068368#
VR: (Image-Updates 15.3 -> scheduled) : https://openqa.suse.de/tests/7068374
VR: (Image-Updates not 15.3 -> return) : https://openqa.suse.de/tests/7068372

tests/containers/bci_tests.pm
```diff
@@ -69,6 +69,11 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
+    if (get_var('HOST_VERSION') !~ '15-SP3') {
+        record_info('poo#98183', 'Test not supported yet on this host. poo');
```
asmorodskyi, line 73: what "poo" at the end suppose to mean ? :D

jlausuch: leftover

PR #13226, opened 2021-09-10
Do registration for media upgrade cases at system_prepare module

tinawang123:
Do registration for media upgrade cases at system_prepare module.
As we need do regression test after media upgrade, we should register
the modules before regression

Changed the check conditions, as we only can do register at regression job group. The rule is very strictly. I changed it to use setting. It can do register wherever users want. It should not check scc_addons, as some module is registered by default. So I remove this condition. And the upgrade path from sles12+ to sles15+. It will add some modules by default. 

- Related ticket:https://progress.opensuse.org/issues/98195
- Verification run: https://openqa.suse.de/tests/7068280
- https://openqa.suse.de/tests/7097714#

### DIFF
--- a/tests/console/system_prepare.pm
+++ b/tests/console/system_prepare.pm
@@ -25,6 +25,7 @@ use serial_terminal 'prepare_serial_console';
 use bootloader_setup qw(change_grub_config grub_mkconfig);
 use registration;
 use services::registered_addons 'full_registered_check';
+use List::MoreUtils 'uniq';
 use strict;
 use warnings;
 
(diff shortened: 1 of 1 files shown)


hjluo:
LGTM

hjluo:
LGTM

chcao (approved):
LGTM

tests/console/system_prepare.pm
```diff
@@ -55,7 +55,11 @@ sub run {
             add_suseconnect_product(get_addon_fullname('desktop'), undef, undef, undef,                             300, 1);
             add_suseconnect_product(get_addon_fullname('we'),      undef, undef, "-r " . get_var('SCC_REGCODE_WE'), 300, 1);
         }
-        my $myaddons = get_var('SCC_ADDONS');
+        my $myaddons = get_var('SCC_ADDONS', '');
+        $myaddons .= "dev,lgm,wsm" if (is_sle('<15', get_var('ORIGIN_SYSTEM_VERSION')) && is_sle('15+'));
```
lemon-suse, line 59: if there is a SDK in SCC_ADDONS, you haven't deal with such situation.

lemon-suse, line 59: if there is a SDK in SCC_ADDONS, you haven't deal with such situation.

tinawang123: sdk       => is_sle('15+') ? 'sle-module-development-tools' : 'sle-sdk',
It did register at lib/registration.pm file

lemon-suse (approved):

PR #13227, opened 2021-09-10
Exclude empathy_irc from openSUSE

Zaoliang:
it is not supported and cannot not be tested at all.
see http://10.160.64.152/tests/3346


Zaoliang:
https://progress.opensuse.org/issues/98418

Zaoliang:
something wrong with rebase, need to create new PR

PR #13228, opened 2021-09-10
Exclude empathy_irc from openSUSE

Zaoliang:
see https://progress.opensuse.org/issues/98418


Zaoliang:
I don't understand what is wrong with run static test.

PR #13230, opened 2021-09-10
Remove emapthy_irc from openSUSE

Zaoliang:
see https://progress.opensuse.org/issues/98418


okurz (changes requested):
s/emapthy_irc/empathy_irc/

The static check fails stating "x11/empathy/empathy_irc module is not used in any schedule" because you just remove the module from the schedule but not delete the test module.

okurz:
@Zaoliang please don't merge such pull requests in the future. There is a reason for peer review

PR #13231, opened 2021-09-10
Collect /var/log from guest on installation failure

waynechen55:
* **According** to my observation, the /var/log folder on guest also contains very useful information about installation. So also collect it from guest on installation failure.
* **Needle** [linux-login](https://openqa.suse.de/tests/7087177#step/unified_guest_installation/675) is also the one monitor_guest_installation needs to detect, especially for non-sles guest.

* **Verification runs:**
  * [sles guest on sles host fail. collect log via ssh](https://openqa.suse.de/tests/7068363)
  * [sles guest on sles host fail. collect log via console](https://openqa.suse.de/tests/7088856)
  * [oracle guest on sles host fail. collect log via ssh](https://openqa.suse.de/tests/7087177)
  * [oracle guest on sles host fail. collect log via console](https://openqa.suse.de/tests/7087260)
  * [oracle guest on sles host pass](https://openqa.suse.de/tests/7087178)
  * [sles guest on sles host pass](https://openqa.suse.de/tests/7087179)

### DIFF
--- a/lib/guest_installation_and_configuration_base.pm
+++ b/lib/guest_installation_and_configuration_base.pm
@@ -1296,7 +1296,7 @@ sub monitor_guest_installation {
 
     $self->reveal_myself;
     save_screenshot;
-    if (!(check_screen([qw(text-logged-in-root guest-installation-in-progress guest-installation-failures grub2 text-login guest-console-text-login)], 180 / get_var('TIMEOUT_SCALE', 1)))) {
+    if (!(check_screen([qw(text-logged-in-root guest-installation-in-progress guest-installation-failures grub2 linux-login text-login guest-console-text-login)], 180 / get_var('TIMEOUT_SCALE', 1)))) {
         save_screenshot;
         record_info("Can not detect any interested screens on guest $self->{guest_name} installation process", "Going to detach current screen anyway");
         $self->detach_guest_installation_screen;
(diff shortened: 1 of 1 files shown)


waynechen55:

Julie-CAO:
LGTM

guoxuguang:
LGTM

lib/guest_installation_and_configuration_base.pm
```diff
@@ -1699,7 +1699,7 @@ sub post_fail_hook {
     $self->reveal_myself;
     $self->upload_guest_installation_logs;
     save_screenshot;
-    virt_utils::collect_host_and_guest_logs("", "", "/root");
+    virt_utils::collect_host_and_guest_logs("", "", "/root /var/log");
```
alice-suse, line 1702: May I ask what specific logs are useful but not included in supportconfig according to your observation? This directory has massive logs and some duplication with supportconfig. Just uploading the whole directory is not a good idea I am afraid, and also waste of disk space.

waynechen55: This collection only happens on installation failure, so it is not massive. This folder does not contain all the stuff that you think might be there. It only contains installation related logs, for example zypper history/log, mkinitrd log and etc. And it does not only server sles guest, but also other linux os. So this is necessary and absolutely not waste of disk space. I already provided verification run and hope you can have a detailed look.

PR #13234, opened 2021-09-10
Add dovecot and postfix test

mloviska:
- [sle15sp2](http://kepler.suse.cz/tests/6840#step/evolution_prepare_servers/1)
- [sle15sp4](http://kepler.suse.cz/tests/6838#step/evolution_prepare_servers/1)


jlausuch:
This would break the test: https://openqa.suse.de/tests/7085191
Have you indentified the source of the error you get in journal and kdump?

mloviska:
> This would break the test: https://openqa.suse.de/tests/7085191
> Have you indentified the source of the error you get in journal and kdump?

https://openqa.suse.de/tests/7090654# I think it works, it is only my openqa

PR #13236, opened 2021-09-10
Containers: small fixes for BCI Tests

jlausuch:
- don't fail in upload_logs commands
- collect number of failed test cases and die at the end if any fails

I created a new group in Development for BCI. Once the tests are more stable, we can move them to our image validation process.
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=17.8.1&groupid=398

### DIFF
--- a/tests/containers/bci_tests.pm
+++ b/tests/containers/bci_tests.pm
@@ -36,8 +36,8 @@ our $test_envs = get_var('BCI_TEST_ENVS', 'base,init,dotnet,python,node,go,multi
 sub parse_logs {
     my $self = @_;
 
-    upload_logs('junit_build_serial.xml');
-    upload_logs('junit_build_parallel.xml');
+    upload_logs('junit_build_serial.xml',   failok => 1);
+    upload_logs('junit_build_parallel.xml', failok => 1);
 
     # bci-tests produce separate XUnit results files for each environment.
     # We need tp merge all together into a single xml file that will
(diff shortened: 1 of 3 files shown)


asmorodskyi (approved):
LGTM

schedule/containers/sle_image_on_sle_host_podman.yaml
```diff
@@ -8,8 +8,8 @@ conditional_schedule:
       's390x':
         - installation/bootloader_start
   bci:
-    FLAVOR:
-      'Container-Image-Updates':
+    BCI_TESTS:
```
b10n1k, line 11: is this in the variables.md? 
also double check that this is the right one and not `CONTAINERS_BCI_TESTS` as the commit msg describes.

jlausuch: Good point. I will add it. VRs are on the way. Thanks.

tests/containers/bci_tests.pm
```diff
@@ -107,13 +104,21 @@ sub run {
...
+            $error_count += 1;
+            record_info('Meeeh', "There was a problem running the test: $env");
+        }
     }
 
     $self->parse_logs();
+
+    # Record soft failure not to block container image release process record
+    # soft failure. When BCI tests repo is stable enough it can be changed to "die"
+    record_soft_failure("$error_count tests failed.") if ($error_count gt 0);
```
b10n1k, line 121: this is smelly IMO.
you shouldnt need to do anything like that in first place. 
`tox` should produces those results and from this point of you the test looks wrong.

b10n1k (changes requested):
@jlausuch sorry but this looks wrong to me. if you want this to not block the release process and as the test is unstable (and not ready IMO) why not to run it in development group and keep that one stable and sound.

jlausuch:
> @jlausuch sorry but this looks wrong to me. if you want this to not block the release process and as the test is unstable (and not ready IMO) why not to run it in development group and keep that one stable and sound.

Ok, I will enable this test only in development group using same schedule.

tests/containers/bci_tests.pm
jlausuch: now we have a `die`, as it should be.

jlausuch: tox returns 0 or 1 and I dont want to use assert_script_run for those cause I want to run all of them and see what happens. If I use assert_script_run for tox commands, if 1 fails it will stop the test.

jlausuch:
VR: https://openqa.suse.de/tests/7076744

tests/containers/bci_tests.pm
```diff
@@ -69,12 +71,7 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
-    if (get_var('HOST_VERSION') !~ '15-SP3') {
-        record_info('poo#98183', 'Test not supported yet on this host.');
-        return;
-    }
-
-    my $runtime        = get_required_var('CONTAINER_RUNTIME');
+    my $runtime        = get_required_var('CONTAINER_RUNTIME', 'podman');
```
pdostal, line 74: The second parameter imo does not make sense.

pdostal (approved):
OK from my side.

schedule/containers/sle_image_on_sle_host_podman.yaml
b10n1k: resolved. thanks

tests/containers/bci_tests.pm
b10n1k: for ref: https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/419. thanks @jlausuch for that

b10n1k (approved):
+1 please keep in loop

PR #13240, opened 2021-09-13
Create hdd with "Common Criteria" System Role

lilyeyes:
Create qcow2 with "Common Criteria" System Role

- Related ticket: https://progress.opensuse.org/issues/97748
- Needles: added already
- Verification run with "CC", all passed: 
   x86: https://openqa.suse.de/tests/7123524
   aarch64: https://openqa.suse.de/tests/7123525
   s390x: https://openqa.suse.de/tests/7123523
- Verification run **without** "CC", all passed: 
   x86: https://openqa.suse.de/tests/7123532
   aarch64: https://openqa.suse.de/tests/7123528
   s390x: https://openqa.suse.de/tests/7123529
- Verification run for "audit-test" with "CC" qcow2, the it works and failed cases can be tracked by poo atm
   x86: https://openqa.suse.de/tests/7100004
   aarch64: https://openqa.suse.de/tests/7100005
   s390x: https://openqa.suse.de/tests/7105962

### DIFF
--- a/tests/installation/logs_from_installation_system.pm
+++ b/tests/installation/logs_from_installation_system.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright © 2017-2019 SUSE LLC
+# Copyright © 2017-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
@@ -35,6 +35,16 @@ sub run {
     my ($self) = @_;
     my $dasd_path = get_var('DASD_PATH', '0.0.0150');
     select_console 'install-shell';
+
+    # permit ssh root login as it is disabled in "Common Criteria" "System Role" system
+    if (check_var('SYSTEM_ROLE', 'Common_Criteria') && is_sle) {
+        record_soft_failure('workaround for poo#98715');
+        assert_script_run('mount /dev/vda2 /mnt');
+        assert_script_run("sed -i -e 's/PermitRootLogin/#PermitRootLogin/g' /mnt/etc/ssh/sshd_config");
+        assert_script_run('echo PermitRootLogin yes >> /mnt/etc/ssh/sshd_config');
+        assert_script_run('cat /mnt/etc/ssh/sshd_config');
+        assert_script_run('umount /mnt');
+    }
 
     # check for right boot-device on s390x (zVM, DASD ONLY)
     if (check_var('BACKEND', 's390x') && !check_var('S390_DISK', 'ZFCP')) {
--- a/tests/security/cc/cc_audit_test_setup.pm
+++ b/tests/security/cc/cc_audit_test_setup.pm
@@ -41,6 +41,9 @@ sub run {
     # Install audit packages
     zypper_call('in audit audit-audispd-plugins');
 
+    # Install tool packages
+    zypper_call('in wget');
+
     # Workaround for restarting audit service
     assert_script_run('sed -i \'/\[Unit\]/aStartLimitIntervalSec=0\' /usr/lib/systemd/system/auditd.service');
     assert_script_run('systemctl daemon-reload');
--- a/tests/installation/partitioning.pm
+++ b/tests/installation/partitioning.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
-# Copyright © 2009-2013 Bernhard M. Wiedemann
-# Copyright © 2012-2017 SUSE LLC
+# Copyright © 2009-2021 Bernhard M. Wiedemann
+# Copyright © 2012-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
@@ -21,6 +21,10 @@ use version_utils qw(is_leap is_storage_ng is_sle is_tumbleweed);
 use partition_setup qw(%partition_roles is_storage_ng_newui);
 
 sub run {
+    if (check_var('SYSTEM_ROLE', 'Common_Criteria')) {
+        assert_screen 'Common-Criteria-Evaluated-Configuration-RN-Next';
+        send_key 'alt-n';
+    }
     assert_screen 'partitioning-edit-proposal-button', 40;
     if (check_var('PARTITION_EDIT', 'ext4_btrfs')) {
         send_key 'alt-g';


tests/installation/logs_from_installation_system.pm
```diff
@@ -36,6 +36,14 @@ sub run {
     my $dasd_path = get_var('DASD_PATH', '0.0.0150');
     select_console 'install-shell';
 
+    if (check_var('SYSTEM_ROLE', 'Common_Criteria') && is_sle) {
+        assert_script_run('mount /dev/vda2 /mnt');
```
rfan1, line 40: Lili,

Do we have to remount root file system here?

rfan1: BTW, I suggest you add more comments here why we need this change

tests/security/cc/cc_audit_test_setup.pm
```diff
@@ -41,6 +41,9 @@ sub run {
     # Install audit packages
     zypper_call('in audit audit-audispd-plugins');
 
+    # Install tool packages
+    zypper_call('in wget');
```
jouyingbin, line 45: wget will be used in the other test case ?

jouyingbin:
Looks like the VR run is not completely finished yet. (waiting.....)

tests/security/cc/cc_audit_test_setup.pm
jouyingbin: or it is necessary for the next testing.

lilyeyes: Yes, we need wget the audit-test code to run.

tests/installation/logs_from_installation_system.pm
lilyeyes: > Lili,
> 
> Do we have to remount root file system here?

yes, as /mnt does not exist after installation (await_install).

lilyeyes:
> Looks like the VR run is not completely finished yet. (waiting.....)

Finished.

tests/installation/logs_from_installation_system.pm
lilyeyes: > BTW, I suggest you add more comments here why we need this change

Added comments.

jouyingbin:
@lilyeyes Is it able to create the gnome hdd using CC system role ?

jouyingbin:
The subject would be create **qcow2** instead of qcow :)

lilyeyes:
> @lilyeyes Is it able to create the gnome hdd using CC system role ?

I do not know. do you have any requirement for gnome hdd?

lilyeyes:
> The subject would be create **qcow2** instead of qcow :)

Done.

tests/installation/logs_from_installation_system.pm
```diff
@@ -36,6 +36,15 @@ sub run {
     my $dasd_path = get_var('DASD_PATH', '0.0.0150');
     select_console 'install-shell';
 
+    # permit ssh root login as it is disabled in "Common Criteria" "System Role" system
+    if (check_var('SYSTEM_ROLE', 'Common_Criteria') && is_sle) {
+        assert_script_run('mount /dev/vda2 /mnt');
```
rfan1, line 41: Hi Lili,
I think it is NOT the best way to add permit root login here if we have to remount root file system here?
For arm/x86_64/s390zkvm VM, root disk should be /dev/vda2 if we set the device driver to "virtio-blk", but how about zvm? how about UEFI mode? how about BM with ipmi controll?

We hardcode disk name vda here, it might introduce new issue.

jouyingbin:
> > @lilyeyes Is it able to create the gnome hdd using CC system role ?
> 
> I do not know. do you have any requirement for gnome hdd?

Currently no so far, but I think it would be also a option for installation. Such as installation of system role from the Yast2 installation. Do you think so ?  But I think it could be another poo.

rfan1:
> The subject would be create **qcow2** instead of qcow :)

How about "hdd" :1st_place_medal:

tests/installation/logs_from_installation_system.pm
lilyeyes: Then  do you have a good way to handle this?  I have no idea atm.

lilyeyes:
> > > @lilyeyes Is it able to create the gnome hdd using CC system role ?
> > 
> > 
> > I do not know. do you have any requirement for gnome hdd?
> 
> Currently no so far, but I think it would be also an option for installation. Such as installation of system role from the Yast2 installation. Do you think so ? But I think it could be another poo.
Yes currently we have no need for gnome image.
If necessary you can open another poo when you need.

lilyeyes:
@jouyingbin  @rfan1  For the requirements for "gnome mode" and "zvm? how about UEFI mode? how about BM with ipmi control" can we file poos to enhance when needed?

lilyeyes:
> > The subject would be create **qcow2** instead of qcow :)
> 
> How about "hdd"

done

lilyeyes:
@jouyingbin  @rfan1  As discussed offline I added softfail to track the workaround, please help to check again.
And the VR is good too:
https://openqa.suse.de/tests/7135785
https://openqa.suse.de/tests/7135784

rfan1 (approved):
LGTM

jouyingbin (approved):
LGTM. Thanks for the case.

tests/installation/logs_from_installation_system.pm
```diff
@@ -36,6 +36,16 @@ sub run {
     my $dasd_path = get_var('DASD_PATH', '0.0.0150');
     select_console 'install-shell';
 
+    # permit ssh root login as it is disabled in "Common Criteria" "System Role" system
+    if (check_var('SYSTEM_ROLE', 'Common_Criteria') && is_sle) {
+        record_soft_failure('workaround for poo#98715');
+        assert_script_run('mount /dev/vda2 /mnt');
+        assert_script_run("sed -i -e 's/PermitRootLogin/#PermitRootLogin/g' /mnt/etc/ssh/sshd_config");
+        assert_script_run('echo PermitRootLogin yes >> /mnt/etc/ssh/sshd_config');
+        assert_script_run('cat /mnt/etc/ssh/sshd_config');
```
Amrysliu, line 45: This line is used for debug? It's not necessary, right?

lilyeyes: yes

PR #13242, opened 2021-09-13
LTP: Record missing kernel config only in install_ltp

pevik:
It was not intended to soft fail all LTP tests.

Verification run:
* http://quasar.suse.cz/tests/7158 (opensuse-Tumbleweed-DVD-x86_64-Build20210910-ltp_net_stress_route@64bit)

### DIFF
--- a/lib/LTP/utils.pm
+++ b/lib/LTP/utils.pm
@@ -90,9 +90,10 @@ sub get_ltp_version_file {
 }
 
 sub log_versions {
-    my $kernel_pkg     = is_jeos || get_var('KERNEL_BASE') ? 'kernel-default-base' : 'kernel-default';
-    my $kernel_pkg_log = '/tmp/kernel-pkg.txt';
-    my $ver_linux_log  = '/tmp/ver_linux_before.txt';
+    my $report_missing_config = shift;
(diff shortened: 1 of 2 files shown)


czerw (approved):

lib/LTP/utils.pm
```diff
@@ -94,6 +94,7 @@ sub log_versions {
     my $kernel_pkg_log = '/tmp/kernel-pkg.txt';
     my $ver_linux_log  = '/tmp/ver_linux_before.txt';
     my $kernel_config = script_output('for f in "/boot/config-$(uname -r)" "/usr/lib/modules/$(uname -r)/config" /proc/config.gz; do if [ -f "$f" ]; then echo "$f"; break; fi; done');
+    my $is_install_ltp = (((caller 1)[3] // '') =~ /^install_ltp/);
```
mdoucha, line 97: Please don't do this. Add a parameter which controls the softfail instead.

mdoucha (changes requested):

lib/LTP/utils.pm
pevik: ok

mdoucha (approved):
LGTM

PR #13243, opened 2021-09-13
Handle VERSION invocation in ensure_ca_certificates_suse_installed

b10n1k:
In containers we use HOST_VERSION to identify the version that the host
    is using and distinguish from the version that the container is running on.
    
    `ensure_ca_certificates_suse_installed` omits the HOST_VERSION resulting
    to use wrong repo to install ca-certifications in the host.

keep `get_required_var` to not change the current behavior.

- Related ticket: https://progress.opensuse.org/issues/98553
Verification run: 
- https://openqa.suse.de/tests/7090715 check repos https://openqa.suse.de/tests/7090715#step/host_configuration/34

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1830,7 +1830,8 @@ This functions checks if ca-certificates-suse is installed and if it is not it a
 sub ensure_ca_certificates_suse_installed {
     return unless is_sle;
     if (script_run('rpm -qi ca-certificates-suse') == 1) {
-        my $distversion = get_required_var("VERSION") =~ s/-SP/_SP/r;    # 15 -> 15, 15-SP1 -> 15_SP1
+        my $host_version = get_var("HOST_VERSION") ? 'HOST_VERSION' : 'VERSION';
+        my $distversion  = get_required_var($host_version) =~ s/-SP/_SP/r;         # 15 -> 15, 15-SP1 -> 15_SP1
         zypper_call("ar --refresh http://download.suse.de/ibs/SUSE:/CA/SLE_$distversion/SUSE:CA.repo");
         zypper_call("in ca-certificates-suse");
     }


lib/utils.pm
```diff
@@ -1830,7 +1830,8 @@ This functions checks if ca-certificates-suse is installed and if it is not it a
 sub ensure_ca_certificates_suse_installed {
     return unless is_sle;
     if (script_run('rpm -qi ca-certificates-suse') == 1) {
-        my $distversion = get_required_var("VERSION") =~ s/-SP/_SP/r;    # 15 -> 15, 15-SP1 -> 15_SP1
+        my $host_version = get_var("HOST_VERSION") ? get_var("HOST_VERSION") : get_var("VERSION");
```
jlausuch, line 1833:
```suggestion
        my $host_version = get_var("HOST_VERSION", get_var("VERSION"));
```

b10n1k: :) this is what i did and i submitted wrong. awesome. let me fix it. One question tho. Do you have any other places that this variable can be used inside the container spectrum which can be adjust respectively? We call other libraries also so i am afraid that it might be hidden somewhere else as well

b10n1k: > :) this is what i did and i submitted wrong. awesome. let me fix it. One question tho. Do you have any other places that this variable can be used inside the container spectrum which can be adjust respectively? We call other libraries also so i am afraid that it might be hidden somewhere else as well

Wait. actually this is what i have done in the beginning but i changed it. i think this is more clear. and i just notice a small thingy. To address it i need to do something different. do you mind to close this and comment to the next commit?

Vogtinator:
I'm out of the loop on some of the developments, but why is there a `HOST_VERSION` variable?

There's already `VERSION` for the host. For the images we have `CONTAINER_IMAGE_VERSIONS`. Is there a need to have three independent versions?

pdostal:
I am also not aware of `HOST_VERSION` and I see no sense in it as there is `utils::get_os_release()` already.

jlausuch:
> I'm out of the loop on some of the developments, but why is there a `HOST_VERSION` variable?
> 
> There's already `VERSION` for the host. For the images we have `CONTAINER_IMAGE_VERSIONS`. Is there a need to have three independent versions?

Long version short: I wanted to keep VERSION for what's being tested, the SUT, which is the host in some cases and the image in some other cases. This would help us make what is being tested. But I was wrong and this has brought more headaches than benefits... So, we have a ticket to revert that (not easy to do in 1 commit). So, please, be patient with it :)

jlausuch:
> I am also not aware of `HOST_VERSION` and I see no sense in it as there is `utils::get_os_release()` already.

I am sure we have discussed in team several times about this variable, I remember you asked about it some time ago :) 

But yeah, as I said, it creates more confusion, so I will remove it soon.

jlausuch (approved):

PR #13245, opened 2021-09-13
publiccloud: Add openqa VARs as instance tags

cfconrad:
All added openqa vars get a prefix `openqa_var_`. The following
variables are added by default:
```
JOB_ID (retrieved by mmapi::get_current_job_id()
NAME
```

 - Ticked: https://progress.opensuse.org/issues/63604
 - Verification run: https://openqa.suse.de/tests/7090717
   - PCW view: https://publiccloud.qa.suse.de/instances?&state=UNK&state=ACTIVE&state=DELETING&state=DELETED&instance_id=i-09ef257cd523f82d1

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -16,8 +16,9 @@ use testapi qw(is_serial_terminal :DEFAULT);
 use Mojo::Base -base;
 use publiccloud::instance;
 use Data::Dumper;
-use Mojo::JSON 'decode_json';
+use Mojo::JSON qw(decode_json encode_json);
 use utils qw(file_content_replace script_retry);
+use mmapi;
 
 use constant TERRAFORM_DIR     => '/root/terraform';
 use constant TERRAFORM_TIMEOUT => 30 * 60;
@@ -360,8 +361,7 @@ sub terraform_apply {
         # Some auxiliary variables, requires for fine control and public cloud provider specifics
         for my $key (keys %{$args{vars}}) {
             my $value = $args{vars}->{$key};
-            $value =~ s/'/'"'"'/;    # ensure values are escaped properly
-            $cmd .= sprintf(q(-var '%s=%s' ), $key, $value);
+            $cmd .= sprintf(q(-var '%s=%s' ), $key, escape_single_quote($value));
         }
         $cmd .= "-var 'image_id=" . $image . "' " if ($image);
         $cmd .= "-var 'instance_count=" . $args{count} . "' ";
@@ -370,7 +370,7 @@ sub terraform_apply {
         $cmd .= "-var 'name=" . $name . "' ";
         $cmd .= "-var 'project=" . $args{project} . "' " if $args{project};
         $cmd .= "-var 'enable_confidential_vm=true' "    if $args{confidential_compute};
-        $cmd .= sprintf(q(-var 'tags={"openqa_ttl":"%d"}' ), get_var('MAX_JOB_TIME', 7200) + get_var('PUBLIC_CLOUD_TTL_OFFSET', 300));
+        $cmd .= sprintf(q(-var 'tags=%s' ), escape_single_quote($self->terraform_param_tags));
         if ($args{use_extra_disk}) {
             $cmd .= "-var 'create-extra-disk=true' ";
             $cmd .= "-var 'extra-disk-size=" . $args{use_extra_disk}->{size} . "' " if $args{use_extra_disk}->{size};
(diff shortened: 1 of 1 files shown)


lib/publiccloud/provider.pm
```diff
@@ -466,6 +466,24 @@ sub terraform_destroy {
...
+    };
+
+    return encode_json($tags);
+}
+
+sub escape_single_quote {
+    my $s = shift;
+    $s =~ s/'/'"'"'/g;
+    return $s;
+}
```
pdostal, line 485: This feel too generic to be here. Isn't there a library for this?

pdostal (approved):
OK from my side :+1:

lib/publiccloud/provider.pm
```diff
@@ -466,6 +466,24 @@ sub terraform_destroy {
...
+sub terraform_param_tags
+{
+    my ($self) = @_;
+    my $tags = {
+        openqa_ttl        => get_var('MAX_JOB_TIME', 7200) + get_var('PUBLIC_CLOUD_TTL_OFFSET', 300),
+        openqa_var_JOB_ID => get_current_job_id(),
+        map { 'openqa_var_' . $_ => get_var($_) } qw(NAME TEST VERSION FLAVOR DISTRI BUILD),
+    };
+
+    return encode_json($tags);
```
asmorodskyi, line 478:
```suggestion
    return escape_single_quote(encode_json($tags));
```

lib/publiccloud/provider.pm
```diff
@@ -370,7 +370,7 @@ sub terraform_apply {
         $cmd .= "-var 'name=" . $name . "' ";
         $cmd .= "-var 'project=" . $args{project} . "' " if $args{project};
         $cmd .= "-var 'enable_confidential_vm=true' "    if $args{confidential_compute};
-        $cmd .= sprintf(q(-var 'tags={"openqa_ttl":"%d"}' ), get_var('MAX_JOB_TIME', 7200) + get_var('PUBLIC_CLOUD_TTL_OFFSET', 300));
+        $cmd .= sprintf(q(-var 'tags=%s' ), escape_single_quote($self->terraform_param_tags));
```
asmorodskyi, line 373:
```suggestion
        $cmd .= sprintf(q(-var 'tags=%s' ), $self->terraform_param_tags());
```

lib/publiccloud/provider.pm
```diff
@@ -466,6 +466,24 @@ sub terraform_destroy {
     record_info('ERROR', 'Terraform exited with ' . $ret, result => 'fail') if ($ret != 0);
 }
 
+sub terraform_param_tags
+{
+    my ($self) = @_;
+    my $tags = {
+        openqa_ttl        => get_var('MAX_JOB_TIME', 7200) + get_var('PUBLIC_CLOUD_TTL_OFFSET', 300),
+        openqa_var_JOB_ID => get_current_job_id(),
+        map { 'openqa_var_' . $_ => get_var($_) } qw(NAME TEST VERSION FLAVOR DISTRI BUILD),
```
asmorodskyi, line 475: can we at least drop NAME ? because it is basically combination of all other variables + jobid which we also have ?

asmorodskyi: the side effect of this will be bigger records for each instance in PCW ...

cfconrad: > can we at least drop NAME ? because it is basically combination of all other variables + jobid which we also have ?

we do not have the architecture :)

> the side effect of this will be bigger records for each instance in PCW ...

We live in times of big-data, lets play that game!

Ok, what about of dropping all and keep only `NAME`. My idea is, using it as the tooltip for the `openqa` link in pcw.
I also find it good, to double check if the jobid belongs to the expected openqa instance.

lib/publiccloud/provider.pm
cfconrad: as mentioned [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13245#discussion_r707448166) I think it is wrong.

lib/publiccloud/provider.pm
cfconrad: unfortunately not 
```bash
$ grep -rn  "'"'"'"'"'"'"'"
lib/publiccloud/instance.pm:63:        $cmd =~ s/'/'"'"'/g;
lib/publiccloud/provider.pm:488:    $s =~ s/'/'"'"'/g;
lib/utils.pm:1753:        $value =~ s/'/'"'"'/g;
lib/utils.pm:1755:        $key   =~ s/'/'"'"'/g;
tests/wicked/before_test.pm.orig:33:    my $escaped                = $enable_command_logging =~ s/'/'"'"'/gr;
tests/wicked/before_test.pm:34:    my $escaped                = $enable_command_logging =~ s/'/'"'"'/gr;
```

lib/publiccloud/provider.pm
asmorodskyi: > 
> We live in times of big-data, lets play that game!

sure no problem , but let's play it in sane order :) first proper handling of tags in PCW UI than throw tons of tags into VMs :) 

> 
> Ok, what about of dropping all and keep only `NAME`. My idea is, using it as the tooltip for the `openqa` link in pcw.
> I also find it good, to double check if the jobid belongs to the expected openqa instance.

works for me :+1:

lib/publiccloud/provider.pm
asmorodskyi: I don't understand who is "callee" here ? we speaking about function which does not have input params . so can you give me more detailed example which will show that someone passing `"` so it breaks everything :thinking:

asmorodskyi:
generally LGTM after you will lower amount of variables either one way or another :) 
the thread about proper place for `escape_single_quote` call is more "academical" discussion so feel free to ignore it

asmorodskyi (approved):
LGTN

asmorodskyi:
> LGTN

oops :) just a typo :)

PR #13246, opened 2021-09-13
Rename test module ssh_interactive_init to prepare_instance

pdostal:
The test module `publiccloud/ssh_interactive_init.pm` has misleading name.
 * This module creates public cloud instance, adds `bernhard` user and enables password login.
 * This module does not prepare neither use ssh interactive terminal, neither its tunnel.

The ssh tunnel specific settings is now not executed when `TUNNELED` variable is not set.
Modules `transfer_repos`, `register_system` and `patch_and_reboot` now use normal user instead of root.

- Verification run: [containers](http://pdostal-server.suse.cz/tests/12484) [LTP](http://pdostal-server.suse.cz/tests/12494) [img-proof](http://pdostal-server.suse.cz/tests/12493)

### DIFF
--- a/tests/publiccloud/prepare_instance.pm
+++ b/tests/publiccloud/prepare_instance.pm
@@ -8,7 +8,8 @@
 # without any warranty.
 
 # Package: openssh
-# Summary: This tests will deploy the public cloud instance and prepare the ssh
+# Summary: This tests will deploy the public cloud instance, create user,
+#   prepare ssh config and permit password login
 #
 # Maintainer: <qa-c@suse.de>
 
@@ -17,27 +18,12 @@ use publiccloud::utils "select_host_console";
 use testapi;
 use utils;
 
-sub run {
-    my ($self, $args) = @_;
+sub prepare_ssh_tunnel {
+    my $instance = shift;
 
-    # If someone schedules a publiccloud run with a custom SCHEDULE this causes
-    # the test to break, because we need to pass $args, so dying earlier and with clear message about root cause
-    die('Note: Running publiccloud with a custom SCHEDULE is not supported') if (!defined $args);
-
-    select_host_console();    # select console on the host, not the PC instance
-
-    # Create public cloud instance
-    my $provider = $self->provider_factory();
-    my $instance = $provider->create_instance(check_connectivity => 1);
-    $instance->wait_for_guestregister();
-    $args->{my_provider} = $provider;
-    $args->{my_instance} = $instance;
-
-    # configure ssh client, fetch the instance ssh public key, do not use default $instance->ssh_opts
+    # configure ssh client
     my $ssh_config_url = data_url('publiccloud/ssh_config');
     assert_script_run("curl $ssh_config_url -o ~/.ssh/config");
-    assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
-    $instance->ssh_opts("");
 
     # Create the ssh alias
     assert_script_run(sprintf(q(echo -e 'Host sut\n  Hostname %s' >> ~/.ssh/config), $instance->public_ip));
@@ -47,9 +33,7 @@ sub run {
     assert_script_run("install -o $testapi::username -g users -m 0600 ~/.ssh/* /home/$testapi::username/.ssh/");
 
     # Skip setting root password for img_proof, because it expects the root password to NOT be set
-    unless (get_var('PUBLIC_CLOUD_QAM') && get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
-        $instance->run_ssh_command(qq(echo -e "$testapi::password\\n$testapi::password" | sudo passwd root));
-    }
+    $instance->run_ssh_command(qq(echo -e "$testapi::password\\n$testapi::password" | sudo passwd root));
 
     # Permit root passwordless login over SSH
     $instance->run_ssh_command('sudo sed -i "s/PermitRootLogin no/PermitRootLogin prohibit-password/g" /etc/ssh/sshd_config');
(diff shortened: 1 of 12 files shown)


tests/publiccloud/prepare_instance.pm
```diff
@@ -8,7 +8,7 @@
 # without any warranty.
 
 # Package: openssh
-# Summary: This tests will deploy the public cloud instance and prepare the ssh
```
b10n1k, line 11: this shouldnt be removed IMO. a preparation of ssh is happening if you want to be accurate.

b10n1k (commented):
i will let just my opinion. Name doesnt need to change even if `prepare_instance` is not wrong. 

i let it to others to decide but you need to update the copyright and maintainer in any case.

jlausuch:
I understand starting the tunnel is not the task of this module, but it does all the related configuration to make it possible.. 

We already have `prepare_tools.pm`, now `prepare_instance.pm`... I wonder if that will be even more confusing.

I actually like the 3 modules we have related to ssh `ssh_interactive_X`, they follow some common logic: `init`, `start` and `stop`. If you don't like `init`, what about `ssh_interactive_config.pm` ?

asmorodskyi:
> i will let just my opinion. Name doesnt need to change even if `prepare_instance` is not wrong.
> 
> i let it to others to decide but you need to update the copyright and maintainer in any case.

name *need* to be changed because we also have `ssh_interactive_start` and also we have `lib/ssh_interactive_init` . And while yes they all doing something with ssh but when you have **all of them** in same job it is hard to guess what exactly each of them is doing

cfconrad (approved):
LGTM

tests/publiccloud/prepare_instance.pm
pdostal: Review now please.

pdostal:
> We already have `prepare_tools.pm`, now `prepare_instance.pm`... I wonder if that will be even more confusing.
I in general agree, can you propose another name please?

pdostal:
@jlausuch @asmorodskyi we execute this also for LTP or img-proof where the ssh tunnel settings are not needed. Please check the code now and please check the verification runs.

asmorodskyi (approved):
LGTM

asmorodskyi:
ltp and img-proof runs failing in `register_system` ...

tests/publiccloud/register_system.pm
```diff
@@ -37,18 +37,19 @@ sub run {
...
         my $arch = get_var('PUBLIC_CLOUD_ARCH') // "x86_64";
         $arch = "aarch64" if ($arch eq "arm64");
+        my $remote = $args->{my_instance}->username . '@' . $args->{my_instance}->public_ip;
         for my $addon (@addons) {
             next if ($addon =~ /^\s+$/);
             record_info $addon, "Going to register '$addon' addon";
             if ($addon =~ /ltss/) {
                 $regcode = get_required_var('SCC_REGCODE_LTSS');
-                ssh_add_suseconnect_product($args->{my_instance}->public_ip, get_addon_fullname($addon), '${VERSION_ID}', $arch, "-r $regcode");
+                ssh_add_suseconnect_product($remote, get_addon_fullname($addon), '${VERSION_ID}', $arch, "-r $regcode");
```
asmorodskyi, line 46: just for better understanding can you please elaborate what is going on here ? why before it was fine just to pass IP and now we need to pass "username at public IP " ? I missing how your change related to this

pdostal: Before, we `ssh` as `root` but that is not needed and is actually disabled by default.
We enable `root` access for the ssh tunneled tests where it is required.

pdostal:
> ltp and img-proof runs failing in `register_system` ...

Fixed.

asmorodskyi:
CI is broken , please fix it and can be merged from my POV

tests/publiccloud/prepare_instance.pm
b10n1k: betta :)

b10n1k (commented):
Could you also update the copyrights for `register_system.pm` and `transfer_repos.pm`? seems outdated

pdostal:
The CI test usese `origin/master` where the mentioned module still exists even tho I renamed it:
```
tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
publiccloud/ssh_interactive_init module is not used in any schedule
```

The CI is in my opinion broken. @okurz, can I ask you for manual approve?

b10n1k (changes requested):
you still use `ssh_interactive_init` in some modules. take a look  on **tests/publiccloud/patch_and_reboot.pm:15**. it tries to use it as base class :)

pdostal:
> you still use `ssh_interactive_init` in some modules. take a look on **tests/publiccloud/patch_and_reboot.pm:15**. it tries to use it as base class :)

No, you probably mean:
```perl
use Mojo::Base 'publiccloud::ssh_interactive_init';
```
which resolves to `lib/publiccloud/ssh_interactive_init.pm`.

okurz (changes requested):
> The CI is in my opinion broken. @okurz, can I ask you for manual approve?

I am not feeling good about this. I don't want to take CI failures lightly and ignore them. If the CI is broken then please fix it, don't ignore it.

grisu48 (approved):
LGTM :rocket:

jlausuch:
`publiccloud/ssh_interactive_init module is not used in any schedule`

pdostal:
> `publiccloud/ssh_interactive_init module is not used in any schedule`

Yes, it has been removed here. But the Ci get's the list of modules from `master` branch.

jlausuch:
> > `publiccloud/ssh_interactive_init module is not used in any schedule`
> 
> Yes, it has been removed here. But the Ci get's the list of modules from `master` branch.

Yes, I was about to update my comment.. The module has been renamed, so this CI error is not taking into account that change...

pdostal:
@okurz please check again. The CI is now fixed.

okurz (approved):

PR #13247, opened 2021-09-13
Containers: fix hardcoded name for Dockerfile

jlausuch:
Detected failure: https://openqa.opensuse.org/tests/1913262#step/docker_image/111


jlausuch:
https://openqa.opensuse.org/tests/1914049#live

Vogtinator (changes requested):
I think that code is correct, see `BuildTest/Dockerfile` in line 50.

FWICT, `$base` is broken already and just `/python:3`, because of the special treatment of `docker.io` in `registry_url`.

jlausuch:
> I think that code is correct, see `BuildTest/Dockerfile` in line 50.
> 
> FWICT, `$base` is broken already and just `/python:3`, because of the special treatment of `docker.io` in `registry_url`.

nah... yes, you are right... We are downloading the file to `" > $dir/BuildTest/Dockerfile"`.

jlausuch:
That job shoudln't run container tests,as there is no container image for ppc

PR #13249, opened 2021-09-13
SLE Micro: add smoke tests for transactional update

jlausuch:
Basic checks for transactional update operations that
are not tested anywhere else.
This test can be extended later on if needed adding additional
changes in initrd, grub, and performing more checks.

- Related ticket: https://progress.opensuse.org/issues/98538

### DIFF
--- a/products/microos/main.pm
+++ b/products/microos/main.pm
@@ -52,6 +52,7 @@ sub load_feature_tests {
     loadtest 'microos/image_checks' if is_image_flavor;
     loadtest 'microos/one_line_checks';
     loadtest 'microos/services_enabled';
+    loadtest 'transactional/trup_smoke';
     load_transactional_role_tests;
     loadtest 'microos/cockpit_service' unless is_staging;
     loadtest 'console/journal_check';
--- a/schedule/sle-micro/dvd.yaml
+++ b/schedule/sle-micro/dvd.yaml
@@ -49,6 +49,7 @@ schedule:
   - microos/one_line_checks
   - microos/services_enabled
   - microos/cockpit_service
+  - transactional/trup_smoke
   - transactional/filesystem_ro
   - transactional/transactional_update
   - transactional/rebootmgr
--- a/schedule/sle-micro/raw_image.yaml
+++ b/schedule/sle-micro/raw_image.yaml
@@ -28,6 +28,7 @@ conditional_schedule:
     SLE_MICRO_K3S:
       '1':
         - containers/k3s_cli_check
+
 schedule:
   - '{{boot}}'
   - transactional/host_config
@@ -41,6 +42,7 @@ schedule:
   - microos/one_line_checks
   - microos/services_enabled
   - microos/cockpit_service
+  - transactional/trup_smoke
   - transactional/filesystem_ro
   - transactional/transactional_update
   - transactional/rebootmgr
--- a/tests/transactional/trup_smoke.pm
+++ b/tests/transactional/trup_smoke.pm
@@ -1,0 +1,41 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Basic smoke test to verify all basic transactional update
+#           operations work and system can properly boot.
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+use transactional;
+
+
+sub action {
+    my ($target, $text, $reboot) = @_;
+    $reboot //= 1;
+    record_info('TEST', $text);
+    trup_call($target);
+    check_reboot_changes if $reboot;
+}
+
+sub run {
+    my ($self) = @_;
+
+    select_console 'root-console';
+
+    action('bootloader', 'Reinstall bootloader');
+    action('grub.cfg',   'Regenerate grub.cfg');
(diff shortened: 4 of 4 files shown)


jlausuch:
VR: https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23sle_micro_bootloader&distri=sle-micro&version=5.1

mloviska:
LGTM, could you add also image tests, pls ?

Vogtinator:
This test is a bit weird. It's called `bootloader`, but does a lot of stuff which isn't related to bootloaders, and doesn't verify that any of the actions did what they're supposed to.

`setup-selinux` shouldn't be in here. In the case that SELinux is already enabled, it's a noop (especially with the `enable_selinux` module doing the same), and in the case that SELinux is not enabled yet, it'll have a big impact on following modules, making their results invalid.

`t-u bootloader` at the end is AFAICT not necessary, that's already done earlier and there shouldn't be any difference between those runs.

IMO this fits better into the `transactional/` directory, because it's not (SLE) Micro(OS) specific.

jsrain:
> This test is a bit weird. It's called `bootloader`, but does a lot of stuff which isn't related to bootloaders, and doesn't verify that any of the actions did what they're supposed to.

The test expects to verify that the t-u updates related to bootloader do not break the system - this is why I asked for it. Perhaps the verification could be improved, but some of the calls in the workflow showed that they can make system unbootable.
 
> `setup-selinux` shouldn't be in here. In the case that SELinux is already enabled, it's a noop (especially with the `enable_selinux` module doing the same), and in the case that SELinux is not enabled yet, it'll have a big impact on following modules, making their results invalid.

Good catch - Could the test be updated so that SELinux is not enabled during the initial installation?
 
> `t-u bootloader` at the end is AFAICT not necessary, that's already done earlier and there shouldn't be any difference between those runs.

In principle you are right - it should not make any difference, but the purpose is that it does not actually break the system. With recent bugs, in particular on s/390x, it showed that some of the calls in the test have the potential to make a machine unbootable.

> IMO this fits better into the `transactional/` directory, because it's not (SLE) Micro(OS) specific.

Cannot comment on this.

jlausuch:
Moved to transactional dir.
Not sure how to name this test... it basically tries to mess up bootloader... but I don't know a better name for it, sorry. 
I added a condition for SELinux. 

The idea is to have a separate test for this, specially for s390x. I can skip openSUSE if you consider it doesn't bring any benefit for x86_64...

tests/transactional/bootloader.pm
```diff
@@ -0,0 +1,42 @@
...
+    check_reboot_changes;
+}
+
+sub run {
+    my ($self) = @_;
+
+    select_console 'root-console';
+
+    action('bootloader',    'Reinstall bootloader');
+    action('setup-selinux', 'Enable SELinux') if (script_run("selinuxenabled") != 0);
```
Vogtinator, line 34: Is there a specific reason to do it in this test?

Simply scheduling this module after `enable_selinux` should have the same effect, especially because `setup-selinux` also does the same as `bootloader`.

tests/transactional/bootloader.pm
```diff
@@ -0,0 +1,42 @@
...
+    assert_script_run('selinuxenabled');
+    action('grub.cfg',   'Regenerate grub.cfg');
+    action('initrd',     'Regenerate initrd');
+    action('kdump',      'Regenerate kdump');
```
Vogtinator, line 38: `kdump` doesn't have an impact on the bootloader, it just regenerates the kdump initrd

Vogtinator:
What about calling this test something like `transactional_update-extra` or similar?

It currently runs commands which don't fit into the existing `transactional_update` module, and with that name it could also test `cleanup` while at it.

tests/transactional/bootloader.pm
mloviska: IMHO it is about showing that re-installing bootloader itself without enabled selinux does not break the system on any architecture. On s390x images and unless the initrd gets rebuilt, system become unbootable after enabling selinux.

jlausuch:
> What about calling this test something like `transactional_update-extra` or similar?
> 
> It currently runs commands which don't fit into the existing `transactional_update` module, and with that name it could also test `cleanup` while at it.

That sounds more reasonable, thanks for the suggestion.

What if I remove selinux enablement in this test, and we make this test run always by default (no variable to enable it), in the selinux tests and the non-selinux tests?

jsrain:
> What if I remove selinux enablement in this test, and we make this test run always by default (no variable to enable it), in the selinux tests and the non-selinux tests?

Hmm, sounds like a good idea.

jlausuch:
Updated. Let me know if you agree and I will start VRs.

tests/transactional/bootloader.pm
Vogtinator: That sounds like this is more of a regression test for a very specific bug.

Can this be generalized for a use case which runs into the bug? I think that approach results in a better kind of module.

Vogtinator:
> Updated. Let me know if you agree and I will start VRs.

Looks good to me!

jlausuch:
* [SLE-Micro](https://openqa.suse.de/tests/overview?version=5.1&distri=sle-micro&build=jlausuch%2Fos-autoinst-distri-opensuse%23sle_micro_bootloader)

* [MicroOS](https://openqa.opensuse.org/tests/overview?distri=microos&version=Tumbleweed&build=jlausuch%2Fos-autoinst-distri-opensuse%23sle_micro_bootloader)

jlausuch:
Thank you.
I'll merge when VRs are done. O3 takes forever today...

jlausuch:
All the VRs are passing. Merging

PR #13250, opened 2021-09-14
HPC package remain function need python module

tinawang123:
This is a workaround. As the code will delete the dropped module python2 at patch_sle test module. But at install_service for hpc package remain check, it need python2 module. So, it need add python2 before hpc check then de-register python2 again.

- Related ticket: https://progress.opensuse.org/issues/98162
- Verification run: https://openqa.suse.de/tests/7124205

### DIFF
--- a/lib/services/hpcpackage_remain.pm
+++ b/lib/services/hpcpackage_remain.pm
@@ -28,6 +28,7 @@ use utils;
 use strict;
 use Data::Dumper;
 use warnings;
+use registration;
 
 my @diffpkg;
 
@@ -89,10 +90,14 @@ sub full_pkgcompare_check {
     my $stage = $hash{stage};
 
     if ($stage eq 'before') {
+        # It need python2 module to compare packages
+        add_suseconnect_product("sle-module-python2", undef, undef, undef, 300, 1) if (get_var('DROPPED_MODULES', '') =~ /python2/);
         list_pkg("orignalq1w2.txt");
         install_pkg();
         list_pkg("installe3r4.txt");
         compare_pkg("/tmp/orignalq1w2.txt", "/tmp/installe3r4.txt");
+        # De-register python2 module again
+        remove_suseconnect_product('sle-module-python2') if (get_var('DROPPED_MODULES', '') =~ /python2/);
     }
     else {
         check_pkg;


hjluo:
deregister_dropped_modules if (check_var('DROPPED_MODULES', 'python2')); -> deregister_dropped_modules. it will check this in the function.

tinawang123:
> deregister_dropped_modules if (check_var('DROPPED_MODULES', 'python2')); -> deregister_dropped_modules. it will check this in the function.

As this only used to python2 module. If DROPPED_MODULES is others not python2, then it will not run those code.

lib/services/hpcpackage_remain.pm
```diff
@@ -89,10 +91,14 @@ sub full_pkgcompare_check {
     my $stage = $hash{stage};
 
     if ($stage eq 'before') {
+        # It need python2 module to compare packages
```
lemon-suse, line 94: Add the poo in the comments will be better.

lib/services/hpcpackage_remain.pm
```diff
@@ -89,10 +91,14 @@ sub full_pkgcompare_check {
...
+        add_suseconnect_product("sle-module-python2", undef, undef, undef, 300, 1) if (check_var('DROPPED_MODULES', 'python2'));
         list_pkg("orignalq1w2.txt");
         install_pkg();
         list_pkg("installe3r4.txt");
         compare_pkg("/tmp/orignalq1w2.txt", "/tmp/installe3r4.txt");
+        # De-register python2 module again
+        deregister_dropped_modules if (check_var('DROPPED_MODULES', 'python2'));
```
lemon-suse, line 101: It's better to deregister python2 directly since this is a workaround that python2 is added by yourself so no doubt.

lib/services/hpcpackage_remain.pm
tinawang123: After long time later, the poo cannot be opened. It will useless add those at here.  I can add POO at description of PR

lemon-suse: We won't delete any poo why can't open?

lib/services/hpcpackage_remain.pm
```diff
@@ -89,10 +90,14 @@ sub full_pkgcompare_check {
     my $stage = $hash{stage};
 
     if ($stage eq 'before') {
+        # It need python2 module to compare packages
+        add_suseconnect_product("sle-module-python2", undef, undef, undef, 300, 1) if (check_var('DROPPED_MODULES', 'python2'));
```
lemon-suse, line 94: I think the condition need to update since later we may have another module dropped, if it is 'python2, xxx' then can't fulfill the condition.

lib/services/hpcpackage_remain.pm
```diff
@@ -89,10 +90,14 @@ sub full_pkgcompare_check {
...
         list_pkg("orignalq1w2.txt");
         install_pkg();
         list_pkg("installe3r4.txt");
         compare_pkg("/tmp/orignalq1w2.txt", "/tmp/installe3r4.txt");
+        # De-register python2 module again
+        remove_suseconnect_product('sle-module-python2') if (check_var('DROPPED_MODULES', 'python2'));
```
lemon-suse, line 100: Same, need to update condition.

lib/services/hpcpackage_remain.pm
tinawang123: This only need python2. Just for python2 workaround. I'm not sure later if it will be impacted or not. But I think it doesn't need consider so much at here. Because it is just a workaround for special case.

lemon-suse (approved):

PR #13255, opened 2021-09-14
WIP: Don't die if releasenotes URL is not available

rakoenig:
- poo#97316: Sometimes URL for release notes is not yet available
  for SLES. In this case record a soft failure instead of dying.

- Related ticket: https://progress.opensuse.org/issues/97316

- Verification run:https://openqa.suse.de/tests/7100006

### DIFF
--- a/tests/installation/releasenotes_origin.pm
+++ b/tests/installation/releasenotes_origin.pm
@@ -25,7 +25,7 @@ sub run {
     enter_cmd "grep -o \"Got release notes.*\" /var/log/YaST2/y2log";
     assert_screen [qw(got-releasenotes-RPM got-releasenotes-URL)];
     unless (match_has_tag "got-releasenotes-$src") {
-        die "Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src";
+        record_soft_failure("poo#97316: Release notes source does NOT match expectations or not found in YaST logs, expected source: $src");
     }
     enter_cmd "exit";
     # If we don't have system role screen, release notes origin is verified on partitioning screen
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -96,7 +96,6 @@ our @EXPORT = qw(
   load_system_update_tests
   loadtest
   load_testdir
-  load_toolchain_tests
   load_virtualization_tests
   load_x11tests
   load_hypervisor_tests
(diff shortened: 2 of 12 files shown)


tests/installation/releasenotes_origin.pm
```diff
@@ -24,8 +24,10 @@ sub run {
     my $src = check_var('SCC_REGISTER', 'installation') ? "RPM" : "URL";
     enter_cmd "grep -o \"Got release notes.*\" /var/log/YaST2/y2log";
     assert_screen [qw(got-releasenotes-RPM got-releasenotes-URL)];
+    # poo#97316: It can happen that URL is not yet available for SLES
+    # so we only record a soft failure here
     unless (match_has_tag "got-releasenotes-$src") {
-        die "Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src";
+        record_soft_failure("Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src");
```
jknphy, line 30: could you fix previous typo? s/expectaions/expectations/

jknphy (commented):
for checking offline, if you can just pass SCC_REGISTER', 'installation' and the test suite remains unaffected I would go for that path, if more changes are needed, then it would be to hacky and I think it is better to create additional module for that case.

tests/installation/releasenotes_origin.pm
```diff
@@ -21,11 +21,15 @@ sub run {
     assert_screen('release-notes-button');
     send_key('ctrl-shift-alt-x');
     assert_screen('yast-xterm');
-    my $src = check_var('SCC_REGISTER', 'installation') ? "RPM" : "URL";
+    # To find out if we could retrieve release notes we try to fetch the overview page
+    my $src = (script_run("curl -o /dev/null  https://www.suse.com/releasenotes/index.html",
```
jknphy, line 25: not sure if the SUT (Software Under Test) can reach that, but it should depend on arch and product version:
`https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15-SP4/release-notes-sles.rpm`

rakoenig: Another VR: https://openqa.suse.de/tests/7142949

In normal conditions when we expect the release notes from an URL, then the SUT surely can reach that because its the top directory of where it gets the release notes from. If the SUT can't reach this URL, then also YaST won't get release notes from an URL, and the logic will then accept release notes from RPM. If the SUT can reach that, but for any reasons the release notes for the most recent release are not yet ready (as we are experiencing now) the logic will record a soft failure.

rakoenig:
I now put the code back to just record a soft failure with referencing the poo#97316 ticket.  The logic if looking for RPM or URL is stll the same relying on SCC_REGISTER, just to avoid breaking existing old job definitions.

jknphy:
> I now put the code back to just record a soft failure with referencing the poo#97316 ticket. The logic if looking for RPM or URL is stll the same relying on SCC_REGISTER, just to avoid breaking existing old job definitions.

Could you try as well the second part of the ticket and provide a verification: 'schedule the module to offline installation test suite' to see if we need to do further changes or not.

btw, you have some extra commits from previous work (probably some missing rebase).

PR #13256, opened 2021-09-14
Containers: add gcc package to the host OS

jlausuch:
https://progress.opensuse.org/issues/98183


jlausuch:
VRs:
https://openqa.suse.de/tests/7129030
https://openqa.suse.de/tests/7129034

PR #13257, opened 2021-09-14
SLE Micro: New test case for Live Patching

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/98234

### DIFF
--- a/data/autoyast_sle15/sle_micro/live_patching.xml
+++ b/data/autoyast_sle15/sle_micro/live_patching.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <suse_register>
+    <do_registration config:type="boolean">true</do_registration>
+    <install_updates config:type="boolean">false</install_updates>
+    <reg_code>{{SCC_REGCODE}}</reg_code>
+    <reg_server>{{SCC_URL}}</reg_server>
+    <addons config:type="list">
+      <addon>
+        <name>sle-module-live-patching</name>
+        <version>{{VERSION_SLE}}</version>
+        <arch>{{ARCH}}</arch>
+        <reg_code>{{SCC_REGCODE}}</reg_code>
+      </addon>
+    </addons>
+  </suse_register>
+  <bootloader>
+    <global>
+      <timeout config:type="integer">-1</timeout>
+      <hiddenmenu>false</hiddenmenu>
+    </global>
+  </bootloader>
+  <general>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+    </mode>
+  </general>
+  <software>
+    <products config:type="list">
+      <product>SUSE-MicroOS</product>
+    </products>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <packages config:type="list">
+      <package>grub2</package>
+      <package>suseconnect-ng</package>
+      <package>sle-module-live-patching-release</package>
+    </packages>
+    <patterns config:type="list">
+      <pattern>basesystem</pattern>
+      <pattern>lp_sles</pattern>
+      <pattern>microos-selinux</pattern>
+    </patterns>
+  </software>
+  <kdump>
+  <add_crash_kernel config:type="boolean">true</add_crash_kernel>
+  <crash_kernel>191M</crash_kernel>
+  <general>
+    <KDUMP_DUMPFORMAT>compressed</KDUMP_DUMPFORMAT>
+    <KDUMP_DUMPLEVEL>1</KDUMP_DUMPLEVEL>
+  </general>
+</kdump>
+  <users config:type="list">
+    <user>
+      <fullname>Bernhard M. Wiedemann</fullname>
+      <encrypted config:type="boolean">false</encrypted>
+      <user_password>nots3cr3t</user_password>
+      <username>bernhard</username>
+    </user>
+    <user>
+      <encrypted config:type="boolean">false</encrypted>
+      <user_password>nots3cr3t</user_password>
+      <username>root</username>
+    </user>
+  </users>
+</profile>
--- a/tests/microos/live_patching.pm
+++ b/tests/microos/live_patching.pm
@@ -1,0 +1,112 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: smoke test for Live Patching in SLE Micro
+#           1) Get current kernel version
+#           2) Make sure the system has been registered previously
+#           3) Register Live Patching module if not done already
+#           4) Optionally add a kernel update repo
+#           5) Update system and reboot
+#           6) Get new kernel version and compare it
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+use transactional;
+use utils qw(zypper_call ensure_ca_certificates_suse_installed);
+
+sub register_lp_module {
+    my $arch        = get_required_var('ARCH');
+    my $regcode     = get_required_var('SCC_REGCODE');
+    my $sle_version = get_required_var('VERSION_SLE');
+    my $extensions  = script_output('SUSEConnect --list-extensions');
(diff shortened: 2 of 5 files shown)


data/autoyast_sle15/sle_micro/live_patching.xml
```diff
@@ -0,0 +1,65 @@
...
+      <product>SUSE-MicroOS</product>
+    </products>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <packages config:type="list">
+      <package>grub2</package>
+    </packages>
+    <patterns config:type="list">
+      <pattern>basesystem</pattern>
+      <pattern>microos-container_runtime</pattern>
+      <pattern>microos-selinux</pattern>
```
b10n1k, line 41: i just wonder if you need to install any package from `live-patching` module. like `kernel-livepatch-tools`? not sure tho.

jlausuch: Yeah, I will need to add some stuff here.
This is still WIP and will change a lot.. actually I even shouldn't have create this PR yet... :)

tests/microos/live_patching.pm
```diff
@@ -0,0 +1,100 @@
...
+use utils qw(zypper_call);
+use version_utils qw(is_sle_micro);
+
+sub register_lp_module {
+    my $arch        = get_required_var('ARCH');
+    my $regcode     = get_required_var('SCC_REGCODE');
+    my $sle_version = get_var('VERSION_SLE', '15.3');
+    my $extensions  = script_output('SUSEConnect --list-extensions');
+    record_info('extensions', $extensions);
+    unless ($extensions =~ m/Live Patching.*Activated/) {
```
b10n1k, line 34: shouldnt have a softfail for the bug you reported, or something in case the module is not registered. in second thought the module presents in the ay profile. if module is not activated it should be reported immediately. right?

jlausuch: The bug was already fixed. Anyway, I add a `die` below if this happens.

tests/microos/live_patching.pm
```diff
@@ -0,0 +1,100 @@
...
+use warnings;
+use testapi;
+use transactional;
+use utils qw(zypper_call);
+use version_utils qw(is_sle_micro);
+
+sub register_lp_module {
+    my $arch        = get_required_var('ARCH');
+    my $regcode     = get_required_var('SCC_REGCODE');
+    my $sle_version = get_var('VERSION_SLE', '15.3');
```
b10n1k, line 31: is there any way to avoid this? for instance to get it from `SUSEConnect --list-extensions` or ...?

jlausuch: Yes, I don't like it either. Good idea, I'll try to extract it from the output.

jlausuch: I just found a nice way to do this which works like a charm:
`my ($sle_version) = $extensions =~ /sle-module-live-patching\/(.*)\/$arch/;`

But we still have 15.3 hardcoded in the autoyast profile... so I would go with job variable and I will also remove `15.3` from the profile.

jlausuch:
VRs:
[image](http://fromm.arch.suse.de/tests/2653)
[DVD](http://fromm.arch.suse.de/tests/2652)

tests/microos/live_patching.pm
b10n1k: despite the bug the module should be activated in first place by `autoyast/installation`. die when it fails as a second try. i am not sure about this part. Why?

jlausuch: There are 2 use cases:
- DVD flavor with autoyast installation -> everything should be registered and activated during installation
- Image flavor -> there is a scc_registration module already registering the main product. So, we only need to register LP here.

b10n1k: then lets move the L40 outside of that block. So it will be checked in both cases. wdyt?

tests/microos/live_patching.pm
```diff
@@ -0,0 +1,100 @@
...
+}
+
+sub setup_optional_repo {
+    # Option to use a test repository with an existing kernel update.
+    # This is useful if regular channels don't provide any kernel update
+    # and we want to force testing this feature
+    my $test_repo = get_var('SLE_MICRO_KERNEL_REPO');
+    if ($test_repo) {
+        record_info('INFO', "Adding repository $test_repo");
+        assert_script_run 'curl -k https://ca.suse.de/certificates/ca/SUSE_Trust_Root.crt -o /etc/pki/trust/anchors/SUSE_Trust_Root.crt';
```
Vogtinator, line 57: Could use `ensure_ca_certificates_suse_installed` instead. Needs to be adjusted to work with read-only / as well by copying this there

Vogtinator (changes requested):
FWICT, this does not actually do any live patching, beyond registering the extension.

What this currently does is install a kernel update, reboot and verify that the updated kernel got booted.

tests/microos/live_patching.pm
jlausuch: Makes sense, let's ensure it for both cases. Thanks!

jlausuch:
VRs:

- [image](http://fromm.arch.suse.de/tests/2700#)
- [autoyast](http://fromm.arch.suse.de/tests/2701)

jlausuch:
> FWICT, this does not actually do any live patching, beyond registering the extension.
> 
> What this currently does is install a kernel update, reboot and verify that the updated kernel got booted.

Please check again, I added some new commands I got from Jiri.
The only thing I am not sure about is that if we should use zypper up or zypper patch (the transactional way I mean)...

tests/microos/live_patching.pm
```diff
@@ -0,0 +1,112 @@
...
+    $self->setup_optional_repo;
+
+    record_info('Update', 'Updating the system');
+    trup_call('up');
+    check_reboot_changes;
+
+    # Get new kernel version
+    my $new_kernel_version = script_output('uname -r');
+    record_info('Kernel', $new_kernel_version);
+
```
mloviska, line 102: I would also try to run `klp status` (maybe before pulling updates) and `klp -v patches` to check installed patches

tests/microos/live_patching.pm
```diff
@@ -0,0 +1,112 @@
...
+        record_info('Repo', "Adding repository $test_repo");
+        zypper_call "ar $test_repo";
+        zypper_call '--gpg-auto-import-keys ref';
+    }
+}
+
+sub run {
+    my ($self) = @_;
+
+    select_console 'root-console';
```
mloviska, line 79: `select_serial_terminal` fits better as LP and also `virtio console` are supported on `x86_64` and `s390x`

Vogtinator:
> > FWICT, this does not actually do any live patching, beyond registering the extension.
> > What this currently does is install a kernel update, reboot and verify that the updated kernel got booted.
> 
> Please check again, I added some new commands I got from Jiri.

I'm not sure whether I understand what the test is currently doing and what it actually needs to do...
(CC @laenion)...

> The only thing I am not sure about is that if we should use zypper up or zypper patch (the transactional way I mean)...

I think `up`/`patch` won't matter, as long as a live patch package is installed and the live patch is active before doing a `reboot`.
I don't know what needs to be done to get a(ny) livepatch installed, but the current use of `--oldpackage` looks odd.

I don't think that many reboots should be necessary in a test for livepatching ;-) Ideally, it should install a suitable kernel, reboot and then install a live patch and confirm that it's active. Currently it reboots four times.

The module might also have to revert to the previous state (rollback + reboot, or `always_rollback=>1`?) before running other modules, in case it becomes part of a bigger job. If that's not planned it's fine as-is.

jlausuch:
> I'm not sure whether I understand what the test is currently doing and what it actually needs to do...
> (CC @laenion)...
> 
> > The only thing I am not sure about is that if we should use zypper up or zypper patch (the transactional way I mean)...
> 
> I think `up`/`patch` won't matter, as long as a live patch package is installed and the live patch is active before doing a `reboot`.
> I don't know what needs to be done to get a(ny) livepatch installed, but the current use of `--oldpackage` looks odd.
> 
> I don't think that many reboots should be necessary in a test for livepatching ;-) Ideally, it should install a suitable kernel, reboot and then install a live patch and confirm that it's active. Currently it reboots four times.
> 
> The module might also have to revert to the previous state (rollback + reboot, or `always_rollback=>1`?) before running other modules, in case it becomes part of a bigger job. If that's not planned it's fine as-is.

This is the result of last minute feature in a product, quick and not understandable flow defined :)
I will put this on hold until I get more feedback what to do with this, cause the way I was asked to add that "optional" repo is also very hacky... maybe enough for manual testing but it doesn't look nice in automation.

Vogtinator:
Yeah, this is still in early stages, with no (clear) documentation...

Are there any other live patching tests in openQA (or scripts for manual QA) which might help here?

PR #13258, opened 2021-09-15
lib/main_common: Run logs_from_installation_system also on MicroOS

Vogtinator:
I can't think of a good reason not to collect logs in those tests.

- Verification run: http://10.160.67.86/tests/1053


jlausuch (approved):
Yeah, makes sense. Not sure why this condition was there...

PR #13259, opened 2021-09-15
Remove softail net.ipv4.mem systcl check on 15SP1

juadk:
This PR removes a softfail for `net.udp.mem` on 15-SP1, I no longer see the softfail triggering in the last fifteen versions.
Let's remove the softfail first, then monitor the test before adapting the value in the reference file.

- Related ticket: https://progress.opensuse.org/issues/92566
- Needles: N/A
- Verification run: N/A


alvarocarvajald (approved):
LGTM

PR #13265, opened 2021-09-15
console/regproxy: Backdate /etc/hosts to work around boo#1190532

Vogtinator:
Modifications of /etc/hosts trigger kdump initrd generation on the next boot,
but on read-only systems that results in an error.

Instead of wasting time to perform a regeneration and reboot, just set back the
mtime.

- Verification run: https://openqa.opensuse.org/tests/1918264


DimStar77 (approved):
suitable as a workaround

PR #13266, opened 2021-09-16
Revert "Refactor get_suse_container_urls using data structure for each distri"

Vogtinator:
Reverts os-autoinst/os-autoinst-distri-opensuse#13141

Breaks TW stagings, which have `VERSION=Staging:X`.
With the hash-based nature a simple fix isn't really possible...

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright © 2018-2021 SUSE LLC
+# Copyright © 2018 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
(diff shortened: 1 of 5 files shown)


b10n1k:
@Vogtinator i did 
```
elsif (is_opensuse && (get_var("VERSION") =~ /^Staging:(?<letter>.)$/)) {                                                                                                                                                                                                  
        # Staging in opensuse sets version as Staging: where sle do not                                                                                                                                                                                                        
        push @released_images, $images_uri{$args{distri}}{Tumbleweed}{released}->($args{arch});                                                                                                                                                                                
    }
```
[VR](https://openqa.opensuse.org/tests/1919008)
would that work in your opinion?

Vogtinator:
Previously that was taken care of by `is_tumbleweed`, which also matches `Staging:X`

b10n1k:
but this will return 0 because DISTRI is microos. Should i add `return get_var('VERSION') =~ /^Staging:/;` in `is_opensuse`

jlausuch:
Actually, it was hit because of ` is_microos("Tumbleweed"))`

Vogtinator:
Actually, it was the `elsif (is_tumbleweed || is_microos("Tumbleweed")) {` case.

`is_microos` checks for `Staging:X` as well. So I'd prefer to use the existing code there

PR #13267, opened 2021-09-16
Hotfix get urls

b10n1k:
I have approach this problem organizing the data into a hash. The data
    structure can hold information explicitly, so it is more ease to handle.
    
    Each <distri><version> hold three information
    - released
    - totest
    - available_arch
    
    I use anonumous functions for `registries` to provide flexibility. For instance
    you can choose the different registry based on aarch. or it is possible to
    modify the sub to return an array in the future.
    
    I tried to group some of the versions as they provide commonalities but regular
    hash accept scalar as a key.

UPDATE: fix added for opensuse staging

Related ticket: https://progress.opensuse.org/issues/97232
Verification run: https://openqa.opensuse.org/tests/1919773

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright © 2018 SUSE LLC
+# Copyright © 2018-2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
@@ -8,7 +8,7 @@
 # without any warranty.
 
 # Summary: Database for URLs of container images to be tested
-# Maintainer: Fabian Vogt <fvogt@suse.com>
+# Maintainer: Fabian Vogt <fvogt@suse.com>, qa-c team <qa-c@suse.de>
 
 package containers::urls;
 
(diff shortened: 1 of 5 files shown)


Vogtinator:
Now it's back at doing the same thing as the code does previously...
You could add `tw-staging` as fake version+distro and use that as key for the hash.

IMO this is not the right place to use a hash or other fixed data structures.

b10n1k:
> Now it's back at doing the same thing as the code does previously...
> You could add `tw-staging` as fake version+distro and use that as key for the hash.
> 
> IMO this is not the right place to use a hash or other fixed data structures.

Do you have a suggestion? i wouldnt like to use `/data` and if i had to create a separate file in `/lib` has already this *urls.pm*(the name makes perfect sense). is there a better place? unless go with something `/lib/containers/data/urls.pm`. but i can feel already that this will not fly :)

Vogtinator:
> > Now it's back at doing the same thing as the code does previously...
> > You could add `tw-staging` as fake version+distro and use that as key for the hash.
> > IMO this is not the right place to use a hash or other fixed data structures.
> 
> Do you have a suggestion? i wouldnt like to use `/data` and if i had to create a separate file in `/lib` has already this _urls.pm_(the name makes perfect sense). is there a better place? unless go with something `/lib/containers/data/urls.pm`. but i can feel already that this will not fly :)

IMO it might be good enough to keep it as-is and just remove the obsolete entries, which had the most of the complexity. Current Leap and SLE are very regular and don't need much code. The workarounds needed to shoehorn the code into a fixed hash end up more complex and less maintainable than the current approach.

b10n1k:
> > > Now it's back at doing the same thing as the code does previously...
> > > You could add `tw-staging` as fake version+distro and use that as key for the hash.
> > > IMO this is not the right place to use a hash or other fixed data structures.
> > 
> > 
> > Do you have a suggestion? i wouldnt like to use `/data` and if i had to create a separate file in `/lib` has already this _urls.pm_(the name makes perfect sense). is there a better place? unless go with something `/lib/containers/data/urls.pm`. but i can feel already that this will not fly :)
> 
> IMO it might be good enough to keep it as-is and just remove the obsolete entries, which had the most of the complexity. Current Leap and SLE are very regular and don't need much code. The workarounds needed to shoehorn the code into a fixed hash end up more complex and less maintainable than the current approach.

*as-is* is an option(and i do not have problem with that if it looks simple and tidy). if you have read the ticket the problem is `it mix HOST and IMAGE properties, so if they are different, it will return the wrong list of urls.` 
When i start working on that i thought that the simpler way to make a clear description of our data was with this data structure. This gives you a clear representation of what provides what. i can see your point of view, and i wonder if you see any solution over this representation (to solve https://progress.opensuse.org/issues/98658)? 

i see that there are many different opinions and i dont want to take any sides. But I think that the maintenance as much the same as the old approach while the complexity is less (as with every new product or differences among products and versions becomes overwhelmed to track)

jlausuch:
Maintaining this list would be the same as current approach, cause when there is a new SP we need to update the line `elsif (is_sle(">=15-sp4", $version))` anyways... 
But I understand when Fabian says `this is not the right place to use a hash or other fixed data structures.` What I would like to see eventually is something like [this](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/journal_check/bug_refs.json) so we keep the code even simpler than current solution.

b10n1k:
> Maintaining this list would be the same as current approach, cause when there is a new SP we need to update the line `elsif (is_sle(">=15-sp4", $version))` anyways...
> But I understand when Fabian says `this is not the right place to use a hash or other fixed data structures.` What I would like to see eventually is something like [this](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/journal_check/bug_refs.json) so we keep the code even simpler than current solution.



> Maintaining this list would be the same as current approach, cause when there is a new SP we need to update the line `elsif (is_sle(">=15-sp4", $version))` anyways...
> But I understand when Fabian says `this is not the right place to use a hash or other fixed data structures.` What I would like to see eventually is something like [this](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/journal_check/bug_refs.json) so we keep the code even simpler than current solution.

it could  be a json but there is not a good reason for it as this is just internal representation which gives you 0(1) notation. i am not sure what more you gain with the json format in this case.

pdostal (approved):

b10n1k:
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=b10n1k%2Fos-autoinst-distri-opensuse%2313267
https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&version=15-SP3&distri=sle
https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&version=5.1&distri=sle-micro
https://openqa.suse.de/tests/overview?distri=sle-micro&version=5.0&build=b10n1k%2Fos-autoinst-distri-opensuse%2313267
https://openqa.opensuse.org/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&version=Tumbleweed&distri=opensuse
https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&distri=opensuse
https://openqa.opensuse.org/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&version=15.4&distri=opensuse
https://openqa.opensuse.org/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2313267&version=15.3&distri=opensuse

PR #13269, opened 2021-09-16
Required changes to mr_test.pm for testing saptune3

scmschmidt:
These are the changes on mr_test.pm to become compatible with saptune3.
I have not used the script directly! I have re-implemented it in my saptune3 Robot Framework setup, but the changed commands caused problems. Also some exit codes checks need to be altered. Sadly I haven't written down them, so expect one or two delete/rename tests to fail, because the exit codes are different.


okurz (changes requested):
Hi @scmschmidt , happy to see you here and welcome :)

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

okurz:
> @okurz Sören is on vacation and I have to take over this PR. Do you know a better way to do this in Github rather than closing this one and opening one myself?

Yes, you should be able to just push to his fork and add commits (or change) in the same branch.

PR #13273, opened 2021-09-16
Refactor select_role module to not use test_data

JRivrain:
We have nicely implemented select_role test module which already uses OOP and Libyui.
But since we've discovered that it will be more transparent to have test module to be
explicit in what it does, we need to refactor the existing select_role test module a bit

See https://progress.opensuse.org/issues/96801

VRs: http://waaa-amazing.suse.cz/tests/overview?build=select_role


ge0r:
Nice refactoring! 
One suggestion, could we also change `select_role` from `qam/QR/15-SP3/textmode/textmode.yaml` to `select_role_textmode`
and also change it in `qam/QR/15-SP3/lvm/lvm_thin_provisioning.yaml` to `select_role_sles_with_gnome` and then delete the `select_role.pm` module for good?

sofiasyria:
LGTM

tests/installation/system_role/select_role_KDE_desktop.pm
```diff
@@ -0,0 +1,33 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use scheduler 'get_test_suite_data';
+use testapi;
+
+sub run {
+    my $controller = $testapi::distri->get_system_role_controller();
```
OleksandrOrlov, line 27: I suggest to use screen name (or logical unit name, e.g. as in case of expert_partitioner) as a variable for controller. This makes your test module to be more explicit in what it is doing and where. 

Let's imagine you need to call several controllers in your test module, having the variables called by meaningful names will make it more clear when reading the test module, then having 'controller1' do something, then 'controller2' do something. 

Something like, when one say "Select Desktop KDE system role on System Role Screen", in code it's more natural to call it like `$system_role->select_system_role('Desktop with KDE Plasma')`.

Other than that, LGTM. Thanks:)

JRivrain:
> Nice refactoring!
> One suggestion, could we also change `select_role` from `qam/QR/15-SP3/textmode/textmode.yaml` to `select_role_textmode`
> and also change it in `qam/QR/15-SP3/lvm/lvm_thin_provisioning.yaml` to `select_role_sles_with_gnome` and then delete the `select_role.pm` module for good?

I mentioned it in the morning call, we said we may not touch qam things, or maybe I misunderstood. I have no objection to change it :)

schedule/yast/addon-module-ftp.yaml
```diff
@@ -15,7 +15,7 @@ schedule:
   - installation/module_selection/skip_module_selection
   - installation/addon_products_sle
   - installation/system_role/validate_default_role
-  - installation/system_role/select_role
+  - installation/system_role/select_role_textmode
```
jknphy, line 18: s/select_role_text_mode/ to be consistent.

tests/installation/system_role/validate_default_role.pm
```diff
@@ -25,10 +25,10 @@ use testapi;
 use Test::Assert 'assert_equals';
 
 sub run {
-    my $default    = get_test_suite_data()->{system_role}{default};
-    my $controller = $testapi::distri->get_system_role_controller();
+    my $default     = get_test_suite_data()->{system_role}{default};
```
jknphy, line 28: the original idea was to replace also this module (although I don't think that is reflected directly in the ticket), otherwise, we still have test data spread in each test, so there is no gain. I mean there is gain in the sense that we now can just use 'select' without test data, but the default in this case we cannot say that we will do later and put in some extended test suite, because the role selection depends on the product modules installed previously.

Could you apply the same for default role ( we could call it /system_role/default_role_text_mode.pm) and so on, and remove completely test data in each test suite. Otherwise, please file a ticket for it.

tests/installation/system_role/select_role_transactional_server.pm
```diff
@@ -0,0 +1,33 @@
...
+use strict;
+use warnings;
+use base 'y2_installbase';
+use scheduler 'get_test_suite_data';
+use testapi;
+
+sub run {
+    my $system_role = $testapi::distri->get_system_role_controller();
+
+    $system_role->select_system_role('Transactional Server');
```
jknphy, line 29: This [approach](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/24668a3a856f7d61858158edfb3463af9be8a355/lib/Installation/ProductSelection/ProductSelectionPage.pm#L27) has its advantages, I applied because I received feedback about that after first implementation of role selection. Basically we will need to change the tests if the names change or create some sort of additional mapping, it also contains spaces, so if we need for some test to have that as test data (a test that go deeper testing) it would not match our typical nomenclature. 

The idea was to represent all those controls at the level of page, so we just need to use always the same variables of the class, for example 'SLES_for_SAP' to name the variable (this is just to save typing methods) and also in the test (and notice the underscores)  and if the screen changes, the test doesn't change only we overwrite the constructor with the same variables but different search criteria.

jknphy:
> > Nice refactoring!
> > One suggestion, could we also change `select_role` from `qam/QR/15-SP3/textmode/textmode.yaml` to `select_role_textmode`
> > and also change it in `qam/QR/15-SP3/lvm/lvm_thin_provisioning.yaml` to `select_role_sles_with_gnome` and then delete the `select_role.pm` module for good?
> 
> I mentioned it in the morning call, we said we may not touch qam things, or maybe I misunderstood. I have no objection to change it :)

You can do it, just trying to avoid breaks, because a single line of test data missing can break the scenarios, and IMO there are still some things to polish here (even if it is in a follow-up task) so I would like to be those schedules as much static/fixed as possible. Makes more sense to apply the improvements later when we have most of the installation.

jknphy (commented):
I still see more potential improvements here, probably not described well in the task. Let me know wdyt.
Please, consider splitting in commits as usual. Makes the review easier and more clear.

schedule/yast/minimalx_zvm.yaml
```diff
@@ -1,46 +0,0 @@
----
-name: minimalx
```
jknphy, line 2: it would help to share the knowledge to add an additional commit why this one is not scheduled anymore (or at least I could not find it) and its schedule is removed.

tests/installation/system_role/validate_default_role.pm
JRivrain: I can do this, even though it was not indeed mentioned in the ticket at all, I assumed it might be covered by another one.

tests/installation/system_role/select_role_transactional_server.pm
JRivrain: I found this approach interesting so I wanted to give it a try, but it does not work. The type is different, not YRadiobutton but YSingleItemSelector, and a query like `curl 'http://localhost:9999/v1/widgets?type=YSingleItemSelector&label=Minimal'` returns "widget not found". here is how the json looks like:
[
  {
    "class" : "YSingleItemSelector",
    "hstretch" : true,
    "icon_base_path" : "",
    "id" : "role_selector",
    "items" : 
    [
      {
        "label" : "&Text Mode",
        "selected" : true
      },
      {
        "label" : "&Minimal"
      },
      {
        "label" : "&KVM Virtualization Host"
      },
      {
        "label" : "&XEN Virtualization Host"
      }
    ],
    "items_count" : 4,
    "vstretch" : true
  }
]

also just  `curl 'http://localhost:9999/v1/widgets?label=Minimal'` returns "widget not found" as well.

So basically, it cannot work the same way.

jknphy: ahm, right, in this case is only one control, but applies the same idea, so in order to not change the test if the labels change in future and not use test data, those strings should be at the level of the page, so we can create a method for each selection in SystemRolePage which uses those string (and corresponding methods in controller). If the labels change, we will inherit the page (or replaced if no other product was using it), if they change too often, then we will have to revisit and use test data to favor yaml edition over code edition.

tests/installation/system_role/validate_default_role.pm
jknphy: we also don't need to check more than once for the same role in the same group this default.

tests/installation/system_role/validate_default_role_null.pm
```diff
@@ -0,0 +1,31 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+# Summary: Test module to validate default System Role during installation
```
jknphy, line 16: let's use a high-level name, `validate_no_default_role_selected` for example.

tests/installation/system_role/validate_default_role_gnome_desktop.pm
```diff
@@ -0,0 +1,31 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+# Summary: Test module to validate default System Role during installation
```
jknphy, line 16: we can delete 'default' from the name of all tests, so we can use this test module to validate default or validate what we selected, so `validate_role_gnome_desktop` should be enough.

tests/installation/system_role/select_role_transactional_server.pm
```diff
@@ -0,0 +1,30 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    my $system_role = $testapi::distri->get_system_role_controller();
+    $system_role->select_system_role('transactional_server');
+    $system_role->next();
```
jknphy, line 27: let's apply here the 'next' approach to hide the next from the test itself:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13291

lib/Installation/SystemRole/SystemRolePage.pm
```diff
@@ -25,9 +25,14 @@ sub new {
 }
 
 sub init {
-    my $self = shift;
+    my ($self) = @_;
     $self->SUPER::init();
     $self->{sel_role} = $self->{app}->itemselector({id => 'role_selector'});
+    $self->{btn_next} = $self->{app}->button({id => 'next'});
+    $self->{text_mode} = 'Text Mode';
+    $self->{transactional_server} = 'Transactional Server';
```
jknphy, line 33: I did similar just using the bare names and I realize that, something like 'role_text_mode', 'role_transactional_server' would be a bit more helpful for the reader as more and more controls are included in the pages, to identify well what are these constants, but up to you, very cosmetic thingy :)

lib/Installation/SystemRole/SystemRolePage.pm
```diff
@@ -25,9 +25,14 @@ sub new {
 }
 
 sub init {
-    my $self = shift;
+    my ($self) = @_;
     $self->SUPER::init();
     $self->{sel_role} = $self->{app}->itemselector({id => 'role_selector'});
+    $self->{btn_next} = $self->{app}->button({id => 'next'});
+    $self->{text_mode} = 'Text Mode';
```
jknphy, line 32: I was checking both screen and I got confused here with Text Mode,
https://openqa.suse.de/tests/7124541#step/system_role/1
https://openqa.opensuse.org/tests/1930998#step/system_role/6
They look completely different. I would say even more than before, so you use this Page as openSUSE page, could we represent all the roles here and for sle the same, all the ones listed in that screenshot, so in follow-up task they can be applied to all yamls.

lib/Installation/SystemRole/Sle/SystemRolePage.pm
```diff
@@ -0,0 +1,34 @@
...
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
```
jknphy, line 29: let's remove this super call to init here (we can still use this class as a child to not repeat methods), because the roles from openSUSE do not apply (check the other comment below) so we can list here the ones for SLE in a more clear way.

tests/installation/system_role/validate_default_role_transactional_server.pm
```diff
@@ -20,16 +20,12 @@
...
 
 sub run {
-    my $default    = get_test_suite_data()->{system_role}{default};
-    my $controller = $testapi::distri->get_system_role_controller();
-
-    my $selected = $controller->get_selected_role();
-    assert_equals($default, $selected, 'Wrong System Role is pre-selected');
+    my $system_role = $testapi::distri->get_system_role_controller();
+    my $selected = $system_role->get_selected_role();
+    assert_matches(qr/Transactional/, $selected, 'Wrong System Role is pre-selected');
```
jknphy, line 28: Instead of writing here literal 'Transactional', we have this information stored at the page level, so we can just access the page and access its properties, it is the exception where we access from test layer to page layer, but makes it more consistent, so both string are in sync.

jknphy (commented):
It is looking really good! some comments/ideas to improve it.

tests/installation/system_role/validate_default_role_transactional_server.pm
JRivrain: I don't understand what you are suggesting here ?

tests/installation/system_role/validate_default_role_gnome_desktop.pm
JRivrain: I like the idea to shorten the names and use them to validate after, but then it will be hard to understand why this module preceeds the module selection.

jknphy: then we can use `validate_selected_role_gnome_desktop`, 'selected' would work for both, wdyt?
(I think we discuss something like this a lot of time ago in bootloader treating defaults or pre-selected values)

JRivrain: That would do, good idea !

lib/Installation/SystemRole/Sle/SystemRolePage.pm
JRivrain: Not sure if you had something else in mind, but if I remove this call, methods are not properly passed from parent : http://waaa-amazing.suse.cz/tests/15827#step/validate_selected_role_text_mode/2

lib/Installation/SystemRole/SystemRolePage.pm
```diff
@@ -25,9 +25,16 @@ sub new {
 }
 
 sub init {
-    my $self = shift;
+    my ($self) = @_;
     $self->SUPER::init();
-    $self->{sel_role} = $self->{app}->itemselector({id => 'role_selector'});
+    $self->{sel_role}                  = $self->{app}->itemselector({id => 'role_selector'});
+    $self->{btn_next}                  = $self->{app}->button({id => 'next'});
```
jknphy, line 31: we don't need this anymore.

lib/Installation/SystemRole/Sle/SystemRolePage.pm
jknphy: ahm I see, now you changed the next functionality, we don't initialize the button without this line above. The problem that I see is that we will accumulate both roles in the same object, when we should completely override them.
The methods are properly passed, this call only call the parent method for init, the rest of the methods `get_selected_roles`, etc. are passed.
The more elegant solution would be to create abstract class to not repeat the method, so that abstract class inherit from NavigationBase, and then both SystemRolePage just override the constructor.
Another way would be to do something like this `$self->{role}{GNOME_desktop} = 'SLES with GNOME';` so we override the property {role} but this is a bit more obscure.
Anyway I'm fine with this solution, we are supposed to know what to select from test layer, so this is more like technical implementation detail that I want to share with you, perhaps for next time.

lib/Installation/SystemRole/Sle/SystemRolePage.pm
```diff
@@ -0,0 +1,38 @@
...
+    $self->{role_GNOME_desktop}        = 'SLES with GNOME';
```
jknphy, line 30: here we don't need to overwrite, it could have been different keys for different products.

jknphy (approved):
Good job!, just a few comments that should no stop us to merge, more like informative or minor detail.

Btw, now without test data we are an step forward, but I believe that the part related with the validation we need to rework it and not create one test for each role. I will mention in subsequent task to enable this change everywhere. For actions is good to have atomic test, but for checking default, we should have an specific test that probably goes back in navigation and check those, but now is a bit an overkilling (it will be in our roadmap) so removing the test data how you did as suggested it is indeed a step forward and when enabling everywhere we should skip the validation for now.

lib/Installation/SystemRole/Sle/SystemRolePage.pm
JRivrain: Thanks for the ideas :)

PR #13276, opened 2021-09-17
Add notice to PackageHub enablement in early development

foursixnine:
Save time to reviewers by specifying what's the problem (or a possible cause)

https://openqa.suse.de/t7150168

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -196,8 +196,12 @@ sub add_suseconnect_product {
     if ($result != 0 && $retry) {
         if ($name =~ /PackageHub/) {
             record_soft_failure 'bsc#1124318 - Fail to get module repo metadata - running the command again as a workaround';
+
         }
-        assert_script_run("SUSEConnect -p $name/$version/$arch $params", $timeout);
+
+        my $fail_message = ($name =~ /PackageHub/ && check_var('BETA', '1')) ? "PackageHub installation might fail in early development" : undef;
+        assert_script_run("SUSEConnect -p $name/$version/$arch $params", $timeout, $fail_message);
+
     } elsif ($result && !$retry) {
         die "SUSEConnect failed activating module $name with exit code (see log output): $result.";
     }


foursixnine:
Well, at least test didn't explode :)

lib/registration.pm
```diff
@@ -196,8 +196,13 @@ sub add_suseconnect_product {
     if ($result != 0 && $retry) {
         if ($name =~ /PackageHub/) {
             record_soft_failure 'bsc#1124318 - Fail to get module repo metadata - running the command again as a workaround';
+
         }
-        assert_script_run("SUSEConnect -p $name/$version/$arch $params", $timeout);
+
+        my $fail_message = undef;
+        $fail_message = "PackageHub installation might fail in early development" if ($name =~ /PackageHub/ && check_var('BETA', '1'));
```
okurz, line 203:
```suggestion
        my $fail_message = ($name =~ /PackageHub/ && check_var('BETA', '1')) ? "PackageHub installation might fail in early development" : undef;
```

okurz (commented):
+1

lib/registration.pm
foursixnine: Done

okurz (approved):

punkioudi (approved):
In which runs can I see the fail message? Other than that lgtm!

okurz:
> In which runs can I see the fail message? Other than that lgtm!

do you mean a verification run? The message would show up whenever the `assert_script_run` fails. I would not be worried about the format of `assert_script_run`. The open point however would be if the condition is ever met. But for me at this point in code review I care that a regression is unlikely. If the condition is actually met I leave to @foursixnine to prove in production ;)

punkioudi:
Yes, I know when the message will show up, I just asked if there is any verification run

foursixnine:
> Yes, I know when the message will show up, I just asked if there is any verification run

Ideally the VR I provided was supposed to be it, however between me writing the code and pushing the changes, they decided to enable the repository :).

At least the code won't break

PR #13286, opened 2021-09-19
Containers: Simplify containers schedule into single file

jlausuch:
The reason for this proposal is to be able to follow in a
easier way the complex support matrix we have in containers
of different packages depending on sle versions and archs.
Currently, YAML schedule has a limitation in conditional
scheduling and this tries PR to solve that.

With this, we also remove all the duplication we have
currently in the yaml schedules and keep the conditions
in a single place.

Also, for HOST tests I grouped all the tests in a single job.
It will take longer time to execute but we keep all the tests
in a single place (this can be discussed).

The only downside is that we need to add `EXTRATEST=cont` to all
the container tests (it's already present in maintenance jobs)
and also `CONTAINER_RUNTIME` to help the scheduler pick docker
or podman tests.


lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+
+    # Container Image tests
+    if (get_var('FLAVOR', '') =~ /Container-Image/) {
+        loadtest_containers 'host_configuration';
+        loadtest_containers 'docker_image';
+        loadtest_containers 'podman_image';
+    }
+
+    # Container Host tests
+    else {
```
ilausuch, line 42: I understand that we don't have more flavors than two, but I think could be nice to have a elseif here checking the second flavor value

ilausuch (approved):

b10n1k (commented):
just some notes:
- i think there are a few concepts which have different scope. PR could be broke up into smaller chucks 
- We have two  PR running in the same time which try to solve similar things. this doesnt look ideal IMO

lib/containers/utils.pm
```diff
@@ -24,9 +24,15 @@ use warnings;
...
-our @EXPORT = qw(test_seccomp basic_container_tests get_vars can_build_sle_base check_docker_firewall
+our @EXPORT = qw(loadtest_containers test_seccomp basic_container_tests get_vars can_build_sle_base check_docker_firewall
   get_docker_version check_runtime_version container_ip registry_url);
 
+
+sub loadtest_containers {
+    my ($test, %args) = @_;
+    autotest::loadtest("tests/containers/$test.pm", %args);
+}
+
```
pdostal, line 35: This duplicates `load_container_tests`.

lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+        if ($runtime eq 'docker') {
+            loadtest_containers 'docker';
+            unless (is_sle("<=15") and is_aarch64) {
+                loadtest_containers 'zypper_docker';
+                loadtest_containers 'docker_runc';
+            }
+            loadtest_containers 'docker_image';
+            loadtest_containers 'buildah_docker' if is_sle('>=15-SP1');
+            loadtest_containers 'validate_btrfs' if is_x86_64;
```
pdostal, line 51: I'd move this module to the end of `docker` tests.

lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+        loadtest_containers 'docker_image';
+        loadtest_containers 'podman_image';
+    }
+
+    # Container Host tests
+    else {
+        if ($runtime eq 'docker') {
+            loadtest_containers 'docker';
+            unless (is_sle("<=15") and is_aarch64) {
+                loadtest_containers 'zypper_docker';
```
pdostal, line 46: This module should work everywhere.

lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+
+sub load_container_tests {
+    my $runtime = get_required_var('CONTAINER_RUNTIME');
+    boot_hdd_image;
+
+    # Container Image tests
+    if (get_var('FLAVOR', '') =~ /Container-Image/) {
+        loadtest_containers 'host_configuration';
+        loadtest_containers 'docker_image';
+        loadtest_containers 'podman_image';
```
pdostal, line 38: We those tests should not be loaded together.

lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+  load_container_tests
+);
+
+sub load_container_tests {
+    my $runtime = get_required_var('CONTAINER_RUNTIME');
+    boot_hdd_image;
+
+    # Container Image tests
+    if (get_var('FLAVOR', '') =~ /Container-Image/) {
+        loadtest_containers 'host_configuration';
```
pdostal, line 36: This should not run on open(SUSE) neither on RHEL.

jlausuch:
> just some notes:
> 
>     * i think there are a few concepts which have different scope. PR could be broke up into smaller chucks
It can't be broken into smaller pieces, and it's not such a big PR.  This will not have any effect if until we remove YAML_SCHEDULE variable, so it really doesn't break anything.

> 
>     * We have two  PR running in the same time which try to solve similar things. this doesnt look ideal IMO
Which one? Can you point me to it?

lib/main_containers.pm
```diff
@@ -0,0 +1,71 @@
...
+        loadtest_containers 'docker_image';
+        loadtest_containers 'podman_image';
+    }
```
pdostal, line 39: The container module `container_diff` is missing here.

lib/main_containers.pm
jlausuch: If I hardcode the second flavor to `Server-DVD-Updates` then we miss `Online` flavor for 15-SP4. Also, I think it would be good to leave the else in case we want to use it in other places in the future (openSUSE? MicroOS?).

products/sle/main.pm
```diff
@@ -655,6 +655,9 @@ if (is_jeos) {
 if (is_kernel_test()) {
     load_kernel_tests();
 }
+elsif (is_container_test) {
+    load_container_tests();
+}
```
pdostal, line 660: Please don't forget to do this change in `products/opensuse/main.pm` too as we follow the Factory First approach.

pdostal (changes requested):

lib/main_common.pm
```diff
@@ -286,6 +292,15 @@ sub is_kernel_test {
...
 
+{
+    local $@;
+
+    eval "use main_containers 'load_container_tests'";
+    if ($@) {
+        bmwqemu::fctwarn("Failed to load main_containers.pm:\n$@", 'main_common.pm');
+    }
+}
+
```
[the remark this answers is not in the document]
jlausuch: See how it is done for main_ltp.pm. I followed the same approach. Basically, if there is a problem with main_containers.pm, it will throw a warning.

lib/main_common.pm
```diff
@@ -265,6 +267,10 @@ sub is_kernel_test {
         || get_var('TUNED'));
 }
 
+sub is_container_test {
+    return (check_var('EXTRATEST', 'cont'));
+}
+
```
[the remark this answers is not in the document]
jlausuch: Yes, it is used in sle/main.pm

lib/main_common.pm
```diff
@@ -67,6 +68,7 @@ our @EXPORT = qw(
   load_common_opensuse_sle_tests
   load_common_x11
   load_consoletests
+  load_container_tests
```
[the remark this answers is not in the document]
jlausuch: but we are not importing main_containers here. See how it is done for `load_kernel_tests`

lib/containers/utils.pm
jlausuch: It's a different thing. This is the test loader. It's basically a wrapper to not specify the path every time. Same as for `loadtest_kernel`. This is pure cosmetics. I can use all the time `loadtest('containers/XYZ')` instead of `loadtest_containers('ZYZ')`

lib/main_containers.pm
jlausuch: ok

lib/main_containers.pm
jlausuch: Not for this conditions, please check the table here: https://confluence.suse.com/display/ENGCTNRSTORY/Level+Test+Plan#LevelTestPlan-PackagesfromContainersModule

lib/main_containers.pm
jlausuch: My idea was to run those in 1 job for less overhead in openqa jobs, since it's just 2 tests... but yeah, I can change as it is today and we can discuss later.

lib/main_containers.pm
jlausuch: ok for RES, but it runs in openSUSE: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/schedule/containers/sle_image_on_opensuse_host.yaml

lib/main_containers.pm
jlausuch: Right, I'll add it

products/sle/main.pm
jlausuch: O3 jobs don't use yaml schedule. My idea was to gradually adapt this to not break O3 jobs. First convert yaml schedules in OSD to this new way, and then maybe make O3 jobs use this function. 
Factory first applies to test coverage, I am not adding any new test or anything.

b10n1k:
> > just some notes:
> > ```
> > * i think there are a few concepts which have different scope. PR could be broke up into smaller chucks
> > ```
> 
> It can't be broken into smaller pieces, and it's not such a big PR. This will not have any effect if until we remove YAML_SCHEDULE variable, so it really doesn't break anything.
> 
> > ```
> > * We have two  PR running in the same time which try to solve similar things. this doesnt look ideal IMO
> > ```
> 
> Which one? Can you point me to it?

Lets discuss it after standup.

lib/main_common.pm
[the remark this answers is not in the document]
pdostal: Where? It sounds cool but I don't understand it.

lib/main_common.pm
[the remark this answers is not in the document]
pdostal: Then let's put it at least to `main_containers.pm`.

lib/containers/utils.pm
pdostal: OK, but then let's move it to `main_container.pm` too as it's used only there.

lib/main_containers.pm
pdostal: I would then be interested why `zypper_docker` does not work on `aarch64` but I guess that OK for now.

lib/containers/utils.pm
jlausuch: I can do that, but just wanted to keep container related things in container libs. Same as for kernel: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/LTP/utils.pm#L44

lib/main_common.pm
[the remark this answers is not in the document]
jlausuch: But it will fail `main_containers.pm` is not imported in main.pm
Mostly all the `is_X_test` functions are declared here: 
```
.//main_pods.pm:52:sub is_pod_test {
.//main_common.pm:253:sub is_ltp_test {
.//main_common.pm:258:sub is_kernel_test {
.//main_common.pm:270:sub is_container_test {
.//main_common.pm:312:sub is_updates_tests {
.//main_common.pm:318:sub is_updates_test_repo {
.//main_common.pm:333:sub is_memtest {
```
Would it make sense to move it to `containers/utils.pm`? If so, then I need to import it in `sle/main.pm` anyway. If we keep it here, we avoid extra imports.

lib/main_common.pm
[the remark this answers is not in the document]
jlausuch: [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/538345afc7b878c425651f5e78367b7c1fc22ed5/lib/main_common.pm#L268)

lib/main_containers.pm
jlausuch: The package `zypper_docker` doesn't exist for aarch64 for <15 (12-SP3, 12-SP4, 12-SP5, 15). You can check that yourself, just go to scc.suse.com and search for `zypper_docker` package selecting the versions and arch.

lib/containers/utils.pm
jlausuch: I have removed this function to avoid confusion.

lib/main_containers.pm
jlausuch: done

lib/main_containers.pm
jlausuch: done

lib/main_containers.pm
pdostal: I'd really change that and don't mix `podman` with `docker` tests.

lib/main_containers.pm
pdostal: It's maybe question from @b10n1k

lib/main_containers.pm
jlausuch: done

jlausuch: but for the record, see https://openqa.suse.de/tests/7163594

lib/main_containers.pm
jlausuch: done

lib/main_common.pm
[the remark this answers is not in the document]
grisu48: I'd also move it to `containers/utils.pm` similar to `is_publiccloud` (https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/utils.pm#L56)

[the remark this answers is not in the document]
jlausuch: done

products/sle/main.pm
mloviska: Does it mean removal of `load_extra_tests_docker` and `main_common.pm` clean up should come as a follow-up PR?

jlausuch: That could be the next thing to tackle, yes

mloviska:
I am in favour of this proposal as it clearly shows the benefits in container test schedule. Despite it is meant for osd only at the moment, it would be nice to make a TW VR as well. 
(LVPTM) Looks very promising to me! :+1:

pdostal:
> I am in favour of this proposal as it clearly shows the benefits in container test schedule. Despite it is meant for osd only at the moment, it would be nice to make a TW VR as well.
> (LVPTM) Looks very promising to me! +1

I am also OK with this. Let's just not forget about openSUSE.

lib/containers/utils.pm
```diff
@@ -25,7 +25,11 @@ use version_utils;
 use Mojo::Util 'trim';
 
 our @EXPORT = qw(test_seccomp basic_container_tests get_vars can_build_sle_base check_docker_firewall
-  get_docker_version check_runtime_version container_ip registry_url);
+  get_docker_version check_runtime_version container_ip registry_url is_container_test);
+
+sub is_container_test {
+    return (check_var('EXTRATEST', 'cont'));
```
cfconrad: Not sure if `cont` is already used, but I would vote for `container` as `cont` could also be `continuous`, `continental`

jlausuch:
Please do not merge until I provide enough VRs.

lib/containers/utils.pm
cfconrad: Also, maybe it is a good idea to move the `is_container_test` to `lib/main_containers.pm`

lib/main_containers.pm
```diff
@@ -0,0 +1,78 @@
...
+sub is_container_test {
+    return (check_var('EXTRATEST', 'cont'));
+}
+
+sub load_container_tests {
+    my $runtime = get_required_var('CONTAINER_RUNTIME');
+    boot_hdd_image;
+
+    # Container Image tests
+    if (get_var('FLAVOR', '') =~ /Container-Image/) {
```
cfconrad, line 39: just a flavor... 
I would go with subroutines from day 0 here,  like:

```perl
boot_hdd_image;

if (get_var('FLAVOR', '') =~ /Container-Image/) {
    if ($runtime eq 'podman') {
        load_tests_podman_container_image();
    } else {
        load_tests_docker_container_image();
    }
} else {
    if ($runtime eq 'podman') {
        load_tests_podman_common();
    } else {
        load_tests_docker_common();
    }
}
loadtest 'console/coredump_collect';
```

products/sle/main.pm
```diff
@@ -23,6 +23,7 @@ use File::Basename;
 use LWP::Simple 'head';
 use scheduler 'load_yaml_schedule';
 use Utils::Backends qw(is_hyperv is_hyperv_in_gui is_pvm);
+use main_containers qw(is_container_test load_container_tests);
```
cfconrad, line 26:
```suggestion
use main_containers;
```
is enough because of `@EXPORT` which do it for you.

lib/main_containers.pm
```diff
@@ -0,0 +1,78 @@
...
+use warnings;
+
+
+our @EXPORT = qw(
+  is_container_test
+  load_container_tests
+);
+
+sub is_container_test {
+    return (check_var('EXTRATEST', 'cont'));
```
cfconrad, line 31: Not sure if `cont` is already used, but I would vote for `container` as `cont` could also be `continuous`, `continental`

lib/main_containers.pm
jlausuch: I like that. Will do it this way. Thanks!

pdostal (approved):
Thank you, Jose. This is not that bad... for not being YAML ;) :rocket:

lib/main_containers.pm
jlausuch: done

jlausuch:
Question for all reviewers.
With current approach, all the HOST tests for maintenance will be executed together:
http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48
e.g. http://fromm.arch.suse.de/tests/2681

This is a bit different from what we have now cause everything is run in the same job... I guess we don't want such big change...
 
Currently, we are executing the module `3rd_party_images` in `engines_and_tools` job and `docker/podman_image` module in `sle_images_on_sle_host` job. I was thinking of keeping it like this, but moving `3rd_party_images`. So, something like this: 

- `engines_and_tools_docker`: all the related docker packages tests
- `engines_and_tools_docker`: all the related podman packages tests
- `images_docker`: `docker_image` + `docker_3rd_party_images`
- `images_podman`: `podman_image` + `podman_3rd_party_images`

 WDYT?

mloviska:
> Question for all reviewers.
> With current approach, all the HOST tests for maintenance will be executed together:
> http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48
> e.g. http://fromm.arch.suse.de/tests/2681
> 
> This is a bit different from what we have now cause everything is run in the same job... I guess we don't want such big change...
> 
> Currently, we are executing the module `3rd_party_images` in `engines_and_tools` job and `docker/podman_image` module in `sle_images_on_sle_host` job. I was thinking of keeping it like this, but moving `3rd_party_images`. So, something like this:
> 
>     * `engines_and_tools_docker`: all the related docker packages tests
> 
>     * `engines_and_tools_docker`: all the related podman packages tests
Typo I guess, `engines_and_tools_podman`
> 
>     * `images_docker`: `docker_image` + `docker_3rd_party_images`
> 
>     * `images_podman`: `podman_image` + `podman_3rd_party_images`
> 
> 
> WDYT?

asmorodskyi:
> Question for all reviewers.
> With current approach, all the HOST tests for maintenance will be executed together:
> http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48
> e.g. http://fromm.arch.suse.de/tests/2681
> 
> This is a bit different from what we have now cause everything is run in the same job... I guess we don't want such big change...


am I the only one who actually think that  this change make sense ?  I would say it should ease the review . Just IMHO ...

jlausuch:
> > Question for all reviewers.
> > With current approach, all the HOST tests for maintenance will be executed together:
> > http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48
> > e.g. http://fromm.arch.suse.de/tests/2681
> > This is a bit different from what we have now cause everything is run in the same job... I guess we don't want such big change...
> 
> am I the only one who actually think that this change make sense ? I would say it should ease the review . Just IMHO ...

I am also fine to keep all the host tests in a single job (well, 1 for podman, another for docker). So, it would look like something like this: http://fromm.arch.suse.de/tests/2682

pdostal:
Wait, but we have two categories: 1) Runtime tests 2) Images tests

jlausuch:
> Wait, but we have two categories: 1) Runtime tests 2) Images tests

Image tests can mean actually 2 things: image validation, or testing that "released" images (or 3rd party images) work on the host. The difference to test one or the other is with the variable [CONTAINERS_UNTESTED_IMAGES](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/containers/podman_image.pm#L38) .Yeah, it's a bit confusing, that's why I'm in favor of putting all the host test together, including `3rd party images` and `podman_image` .

lib/main_containers.pm
```diff
@@ -0,0 +1,92 @@
...
+use warnings;
+
+
+our @EXPORT = qw(
+  is_container_test
+  load_container_tests
+);
+
+sub is_container_test {
+    return (check_var('EXTRATEST', 'containers'));
```
jlausuch, line 31: I was thinking that we don't even need this variable.. we could do something like:
```
get_var('CONTAINER_RUNTIME') ? return 1 : return 0;
```

jlausuch:
VRs:
[Server-DVD-Updates](http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48)
[Image-Updates (15-SP3 only)](http://fromm.arch.suse.de/tests/overview?&groupid=34)

asmorodskyi (approved):
LGTM

lib/main_containers.pm
```diff
@@ -0,0 +1,92 @@
...
+    loadtest 'containers/podman_image';
+}
+
+sub load_tests_docker_image {
+    loadtest 'containers/docker_image';
+    loadtest 'containers/container_diff';
+}
+
+sub load_tests_podman_common {
+    unless (is_sle('<=15')) {
```
cfconrad, line 44: I know `unless` is cool, but maybe if is more readable
```suggestion
    if (is_sle('>15')) {
```

lib/main_containers.pm
```diff
@@ -0,0 +1,92 @@
...
+        loadtest 'containers/podman_image';
+        loadtest 'containers/buildah_podman';
+        loadtest 'containers/rootless_podman';
+        loadtest 'containers/podman_3rd_party_images';
+    }
+}
+
+sub load_tests_docker_common {
+    loadtest 'containers/docker';
+    unless (is_sle("<=15") and is_aarch64) {
```
cfconrad, line 55:
```suggestion
    if (is_sle(">15")  && !is_aarch64) {
```

cfconrad (approved):
LGTM, thx!

lib/main_containers.pm
jlausuch: ok

lib/main_containers.pm
jlausuch: Your approach doesn't fit with the scenario.
these 2 packages are available for ALL the cases EXCEPT (aarch64, <=15)

https://confluence.suse.com/display/ENGCTNRSTORY/Level+Test+Plan#LevelTestPlan-PackagesfromContainersModule

jlausuch:
Latest runs:
- [http://fromm.arch.suse.de/tests/overview?groupid=42&groupid=43&groupid=44&groupid=45&groupid=46&groupid=47&groupid=48](Server-DVD-Updates) (failed jobs are due to problem in  worker with infinite "uploading" state)
- [Image-Updates (15-SP3 only)](http://fromm.arch.suse.de/tests/overview?&groupid=34)

lib/main_containers.pm
cfconrad: Ups, yeah it should be `or` 
In the end I think your logic looks more sane :) so I revert my advice!

jlausuch:
> consequences:
> 
>     * https://openqa.opensuse.org/tests/1943358
> 
>     * https://openqa.opensuse.org/tests/1943359

Working on a fix

jlausuch:
Follow up fix: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13349

PR #13293, opened 2021-09-20
Split releasenotes_origin module into 2 modules for checking URL or RPM origin

rakoenig:
Fixes for fails in releasenotes_origin:

- The original releasenotes_origin test will be abandoned and replaced by
  - **releasenotes_online** which expects releasenotes coming from an URL
  - **releasenotes_offline** which expects releasenotes coming from an RPM
- If expectation is not matched the tests records a soft failure

- Related ticket: https://progress.opensuse.org/issues/97316
- Verification run: 
  - releasenotes_online: https://openqa.suse.de/tests/7173438
  - releasenotes_offline: https://openqa.suse.de/tests/7173552


tests/installation/releasenotes_online.pm
```diff
@@ -0,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright © 2017-2018 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Check origin of displayed releasenotes during installation
```
jknphy, line 10: As a suggestion 'release_notes_from_rpm' and 'release_notes_from_url' would reflect better what is about the test without opening the code than 'online' or 'offline'. In other words, as I will schedule those test modules in Online or Offline, as a tester I already have that info, so the name of the test should include the origin somehow.

tests/installation/releasenotes_online.pm
```diff
@@ -0,0 +1,37 @@
...
+# Maintainer: mgriessmeier <mgriessmeier@suse.com>, Nick Singer <nsinger@suse.de>
```
jknphy, line 11: For new tests, please borrow a 'header' that only contains copyright 2021 and Mantainer, our team, Tags is not needed, just copy/paste from another example.

tests/installation/releasenotes_offline.pm
```diff
@@ -21,11 +21,12 @@ sub run {
     assert_screen('release-notes-button');
     send_key('ctrl-shift-alt-x');
     assert_screen('yast-xterm');
-    my $src = check_var('SCC_REGISTER', 'installation') ? "RPM" : "URL";
+    # This is releasenotes_offline so we expect release notes from RPM
```
jknphy, line 24: this comment doesn't provide any valuable info which is not already obvious from the code itself, please delete it.

tests/installation/releasenotes_offline.pm
```diff
@@ -21,11 +21,12 @@ sub run {
...
+    my $src = "RPM";
     enter_cmd "grep -o \"Got release notes.*\" /var/log/YaST2/y2log";
     assert_screen [qw(got-releasenotes-RPM got-releasenotes-URL)];
     unless (match_has_tag "got-releasenotes-$src") {
-        die "Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src";
+        record_soft_failure("Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src");
```
jknphy, line 29: existing typo fixed in accidental PR reintroduced here again: s/expectaions/expectations/

tests/installation/releasenotes_offline.pm
```diff
@@ -21,11 +21,12 @@ sub run {
     assert_screen('release-notes-button');
     send_key('ctrl-shift-alt-x');
     assert_screen('yast-xterm');
-    my $src = check_var('SCC_REGISTER', 'installation') ? "RPM" : "URL";
+    # This is releasenotes_offline so we expect release notes from RPM
+    my $src = "RPM";
```
jknphy, line 25: No need for variable, we just need to put the value in `match_has_tag` and `record_soft_failure`.

Just a hint, but not everybody agree, if you don't have interpolation of strings, just use single quotes (some people prefers to use double so in the future if is expected interpolation, they are already there, but in general I've seen that not much useful).

tests/installation/releasenotes_offline.pm
jknphy: In offline scenario we don't soft-fail, we just die if there are not found by rpm, there is not network connection, if the logs would show url would be wrong and we still need to die, you could use `check_screen` for that case, but...

The correct way to do it is to switch tty instead of using needling, but that would mean to change the whole test, so I would recommend you to just check needle rpm and die, we can revisit this in future.

tests/installation/releasenotes_online.pm
```diff
@@ -0,0 +1,37 @@
...
+# Tags: fate#323273, poo#26786, #poo97316
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use version_utils ':VERSION';
```
jknphy, line 18: you can remove lines 32-34 because we don't need to check for the next screen, next test just does that, so we can also remove this line importing the function.

tests/installation/release_notes_from_rpm.pm
```diff
@@ -1,36 +1,31 @@
...
 #
-# Copyright © 2017-2018 SUSE LLC
+# Copyright © 2021 SUSE LLC
 #
 # Copying and distribution of this file, with or without modification,
 # are permitted in any medium without royalty provided the copyright
 # notice and this notice are preserved.  This file is offered as-is,
 # without any warranty.
 
 # Summary: Check origin of displayed releasenotes during installation
```
jknphy, line 10: s/releasenotes/release notes/
It doesn't check the origin, check that the origin is from rpm. Please also update the other file.

tests/installation/release_notes_from_rpm.pm
```diff
@@ -1,36 +1,31 @@
...
 use warnings;
 use testapi;
-use version_utils ':VERSION';
 
 sub run {
     assert_screen('release-notes-button');
     send_key('ctrl-shift-alt-x');
     assert_screen('yast-xterm');
-    my $src = check_var('SCC_REGISTER', 'installation') ? "RPM" : "URL";
+    # This is releasenotes_offline so we expect release notes from RPM
```
jknphy, line 22: I would recommend to remove this comment, as you see, after a refactor of the code is even less meaningful.

jknphy (commented):
LGTM, just some minor things.

jknphy:
so you renamed the test, then we must remove [this line](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/main_common.pm#L867). Also we must remove the setting from `variables.md` and `schedule/yast/releasenotes_origin+unregistered.yaml`

tests/installation/releasenotes_origin.pm
```diff
@@ -1,36 +0,0 @@
...
-    unless (match_has_tag "got-releasenotes-$src") {
-        die "Release notes source does NOT match expectaions or not found in YaST logs, expected source: $src";
-    }
-    enter_cmd "exit";
-    # If we don't have system role screen, release notes origin is verified on partitioning screen
-    my $current_screen = is_using_system_role() ? 'system-role-default-system' : 'partitioning-edit-proposal-button';
-    assert_screen $current_screen, 180;
-}
-
-1;
```
jknphy, line 36: we should include in this commit the removal from variables.md and from main.pm

tests/installation/release_notes_from_rpm.pm
```diff
@@ -0,0 +1,30 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Check origin of displayed release notes during installation
```
jknphy, line 10: You need to provide a meaningful commit message: "Fixes for poo#97316" References to ticket do not indicate what is about, for me...I don't care much, but other collaborator filter with git-log checking latest changes in the repo and they would like to see it in that way.

Please, update the summary of of both file, they are doing two difference things, so they should have different summary, one mentioning rpm the other mentioning url.

rakoenig:
Latest VRs:

- release_notes_from_url: https://openqa.suse.de/tests/7185397
- release_notes_from_rpm: https://openqa.suse.de/tests/7185396

jknphy:
LGTM, could you please resolve the conflict and we merge it?
btw, minor typo in commit title, s/Remove releasenotes_origin modul/Remove releasenotes_origin module/

PR #13295, opened 2021-09-21
Add opensuse 15.4 links to the images_uri hash

b10n1k:
opensuse 15.4 is enabled but the images are not available in the https://registry.opensuse.org/
i expect that the support is the same as 15.3 and the uris follows the same convention.

- Related ticket: https://progress.opensuse.org/issues/97232


jlausuch:
LGTM but need to squash commits.

jlausuch (approved):

PR #13301, opened 2021-09-21
Select desktop module in ext4 test suite on 390x

OleksandrOrlov:
The commit fixes [ext4 test suite](https://openqa.suse.de/tests/7125516) on s390x by selecting Desktop module
during installation, as it is expected to be installed.

- Verification run: https://openqa.suse.de/tests/7184626


jknphy:
thanks! I had it in my queue to fix this morning this error I introduced, you were fast, awesome! ;)

PR #13302, opened 2021-09-21
Fix non-existent test modules check

pdostal:
- Related pull request: #13246

The `tools/detect_unused_modules` sh script is run exclusively from `Makefile` and is used only *for test modules*:
```bash
$ tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
publiccloud/ssh_interactive_init module is not used in any schedule
```

But in the `sh` script mentioned above we check that if such *file* exists (as it may be removed / renamed - which is the case of my pull request mentioned also above). So in my case this command is executed:
```bash
$ git ls-files "*publiccloud/ssh_interactive_init*"
lib/publiccloud/ssh_interactive_init.pm
```
In my case the test module I renamed is not detected as there is similar module under `lib/` and the test fails.

As `tools/detect_unused_modules` is used only for test modules, I propose this solution.
I may also rename it to `tools/detect_unused_test_modules` if you agree.

Thank you for your cooperation.

### DIFF
--- a/tools/detect_unused_modules
+++ b/tools/detect_unused_modules
@@ -37,7 +37,7 @@ done
 
 for test in $MODULES ; do
     # Don't check non-existing modules it as it was either renamed or removed
-    if [ -z `git ls-files "*$test*"` ]; then
+    if [ -z `git ls-files "tests/$test*"` ]; then
         continue
     fi
     t=$(basename $test | sed -e 's@\.pm$@@')


b10n1k (approved):
+1
only one favor. it would be nice to add some description in the commit message. when i search in the git logs it makes it much easier to find some details. so for instance visiting  https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/5cdb32775f549a31c4e17e5cc818e30419846da2 you can read immediately what was the purpose of those lines. Which if you consider someone in the future looking for the info of your change will be a bit harder. nice catch btw

jlausuch (commented):
+1

pdostal:
Thank you @b10n1k, done.

PR #13303, opened 2021-09-21
Remove usage of nonexistent method to fix Staging

OleksandrOrlov:
The method 'has_valid_text' was recently removed, but it is still used by Registration Controller.

The PR removes its usage.

- Verification run: https://openqa.suse.de/tests/7184596

### DIFF
--- a/lib/Installation/Registration/RegistrationController.pm
+++ b/lib/Installation/Registration/RegistrationController.pm
@@ -28,11 +28,9 @@ sub init {
     my ($self, $args) = @_;
     $self->{RegistrationPage}    = Installation::Registration::RegistrationPage->new({app => YuiRestClient::get_app()});
     $self->{UseUpdateReposPopup} = Installation::Warnings::ConfirmationWarning->new({
-            app        => YuiRestClient::get_app(),
-            match_text => "The registration server offers update repos.*"});
+            app => YuiRestClient::get_app()});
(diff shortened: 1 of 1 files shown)


sofiasyria:
I didn't face any issue with staging, I just ran by chance crypt_no_lvm and it was broken. LGTM

lib/Installation/Registration/RegistrationController.pm
```diff
@@ -28,11 +28,9 @@ sub init {
     my ($self, $args) = @_;
     $self->{RegistrationPage}    = Installation::Registration::RegistrationPage->new({app => YuiRestClient::get_app()});
     $self->{UseUpdateReposPopup} = Installation::Warnings::ConfirmationWarning->new({
-            app        => YuiRestClient::get_app(),
-            match_text => "The registration server offers update repos.*"});
```
ge0r, line 32: Why is `match_text` also deleted?

ge0r: Ah I see, probably was used in `has_valid_text` and now has no usage as well

OleksandrOrlov: Because it was just used by `has_valid_text` method. If no method, then no 'match_text' needed:)

jknphy:
From the two checks removed, we should add at least the check in installer extended to not loose track of those few default that now we check.
Feel free to file a ticket to create installer_extended test suite for Full medium, as a temporal thing might help.

PR #13305, opened 2021-09-22
Permit root ssh for system with CC system role

rfan1:
In "Common Criteria System Role" system, root ssh login
is disabled by default, we need enable it on 390x to fit openQA
tests. btw, Enhance the code to locate the root disk rather than
hardcode the disk name

- Related ticket: https://progress.opensuse.org/issues/98715
- Verification run: 
x86_64: https://openqa.suse.de/tests/7200097#
arm:       https://openqa.suse.de/tests/7200098#
s390x:   https://openqa.suse.de/tests/7200087#

### DIFF
--- a/tests/installation/logs_from_installation_system.pm
+++ b/tests/installation/logs_from_installation_system.pm
@@ -36,13 +36,14 @@ sub run {
     my $dasd_path = get_var('DASD_PATH', '0.0.0150');
     select_console 'install-shell';
 
-    # permit ssh root login as it is disabled in "Common Criteria" "System Role" system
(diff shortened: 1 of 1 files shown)


lilyeyes (approved):
LGTM

jouyingbin:
@rfan1 Look good to me :) BTW, can you also provide a link in the PR description if we did not apply the patch, it will have a softfail result, right?

jouyingbin (approved):
LGTM

Amrysliu (approved):
LGTM

PR #13306, opened 2021-09-22
Containers: BCI tests now support python3.6

jlausuch:
bci-test repo was before only working in python3.9
Now, there is possibility to use python3.6 which is will help us
run the tests on different hosts (even on old ones).

- Related ticket: https://progress.opensuse.org/issues/98183

### DIFF
--- a/tests/containers/bci_tests.pm
+++ b/tests/containers/bci_tests.pm
@@ -91,8 +91,8 @@ sub run {
     }
 
     record_info('Install', 'Install needed packages');
-    zypper_call('--quiet in git-core python39 python39-devel gcc', timeout => 600);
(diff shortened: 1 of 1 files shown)


jlausuch:
VR: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=17.8.5&groupid=398

the tests are red due to a [bug in the repo](https://progress.opensuse.org/issues/98183#note-8), but it's not due to python3.6 support. I will start enabling this test also in non-SLE host where there is no SUSEConnect.

ilausuch (approved):

pdostal (approved):

PR #13307, opened 2021-09-22
pam: pam-deprecated is deprecated: do no install

DimStar77:
pam-deprecated has been dropped and no longer will be installable.

- Related ticket: https://progress.opensuse.org/issues/99039
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/1930005

### DIFF
--- a/data/pam_test/pam.sh
+++ b/data/pam_test/pam.sh
@@ -53,19 +53,7 @@ teardown() {
     [ "$status" -ne 0 ]
 }
 
-@test "deny access if too many attempts fail" {               # case 04
-    sed -i '/^auth/i\auth required pam_tally2.so onerr=fail deny=1 unlock_time=10' /etc/pam.d/pam_test
-
-    run bash -c "echo -en '$USER_ERR_PW' | pam_test auth $USER_NOR" # error password triggers this case
-    [ "$status" -ne 0 ]
-    sleep 2 # this time is less that 10s
-    run bash -c "echo -en '$USER_NOR_PW' | pam_test auth $USER_NOR"
-    [ "$status" -ne 0 ]
-    sleep 12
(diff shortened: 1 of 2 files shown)


mloviska:
The test case as is, does not fail in [sle testing](https://openqa.suse.de/tests/overview?arch=&flavor=&machine=&test=&modules=pam&module_re=&distri=sle&version=15-SP3&build=20211012-1&groupid=366#)

I guess it might be safer to skip test case rather than remove them completely. Therefore I have tried to add a skip test case flow in case the related module is not listed. 

However, the `pam_tally2.so` is not found and yet the test case passes anyway for sle. Does it have some other alias? 
http://kepler.suse.cz/tests/7856# vs http://kepler.suse.cz/tests/7838

```
+++ b/data/pam_test/pam.sh
@@ -54,6 +54,7 @@ teardown() {
 }

 @test "deny access if too many attempts fail" {               # case 04
+    (pam-config --list-modules | grep -q pam_tally2.so) || skip
     sed -i '/^auth/i\auth required pam_tally2.so onerr=fail deny=1 unlock_time=10' /etc/pam.d/pam_test

     run bash -c "echo -en '$USER_ERR_PW' | pam_test auth $USER_NOR" # error password triggers this case
@@ -114,6 +115,7 @@ teardown() {
 }

 @test "set a complex password" {                            # case 12
+    (pam-config --list-modules | grep -q pam_cracklib.so) || skip
     pam-config -a --cracklib --cracklib-retry=1 --cracklib-minlen=8 --cracklib-dcredit=-1 --cracklib-ucredit=-1 --cracklib-lcredit=-1 --cracklib-ocredit=-1

     # the length of password is less than 8
 ```

PR #13318, opened 2021-09-23
publiccloud : add workaround for LTP to detect that we are in virtual env

asmorodskyi:


tests/publiccloud/run_ltp.pm
```diff
@@ -145,6 +145,10 @@ sub run {
     my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "master");
     assert_script_run("git clone -q --single-branch -b $runltp_ng_branch --depth 1 $runltp_ng_repo");
     $instance->run_ssh_command(cmd => 'sudo CREATE_ENTRIES=1 ' . get_ltproot() . '/IDcheck.sh', timeout => 300);
+    if(is_ec2_arm()){
+        record_info('Workaround', 'AWS Nitro virtualization not recognized by systemd-detect-virt see bsc#1190440');
+        $instance->run_ssh_command(cmd => 'sudo echo "LTP_VIRT_OVERRIDE=amazon" >> /etc/environment');
```
cfconrad, line 150: this doesn't work if you are a user... I think it should be
```suggestion
        $instance->run_ssh_command('echo "LTP_VIRT_OVERRIDE=amazon" | sudo tee -a /etc/environment');
```

lib/publiccloud/utils.pm
```diff
@@ -74,6 +74,11 @@ sub is_ec2() {
     return is_publiccloud && check_var('PUBLIC_CLOUD_PROVIDER', 'EC2');
 }
 
+# Check if we are on an AWS test run and VM running on the host with ARM architecture
+sub is_ec2_arm() {
+  return is_ec2() && get_var('FLAVOR') =~ 'ARM';
```
cfconrad, line 79: I don't like it, that we check the `FLAVOR` for an architecture. But I don't have a better idea... 
And conclusion is, using a function to wrap that check is mandatory.
So +1 from me :)

tests/publiccloud/run_ltp.pm
asmorodskyi: no problem to change it but just theoretical question - what the benefit of `tee -a` in comparison with `>>` .

cfconrad: that it will work with sudo.
In your case, the `echo` command is executed as `root` but the redirect get the `USER` permissions. And a normal `user` do not have write access to `/etc`.
Just try your command on your computer.

cfconrad (approved):

lib/publiccloud/utils.pm
b10n1k: not sure if `FLAVOR` or `PUBLIC_CLOUD_ARCH` is better here. 
also you can define this from the machine type. i think all the arm use `a1`

PR #13320, opened 2021-09-23
data/delete_users: Avoid userdel -f

Vogtinator:
Instead of deleting the user forcibly, kill all processes by the to be deleted
user explicitly. The "dangling" processes cause issues like a crash of
the system dbus-daemon: https://gitlab.freedesktop.org/dbus/dbus/-/issues/343

- Verification run: https://openqa.opensuse.org/tests/1932598

### DIFF
--- a/data/delete_users
+++ b/data/delete_users
@@ -1,8 +1,10 @@
-#!/bin/bash
+#!/bin/sh
+set -eu
 
 n_users="$1"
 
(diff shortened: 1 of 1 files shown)


asdil12 (approved):
LGTM

Vogtinator:
VR passed!

PR #13323, opened 2021-09-24
Use my ($self) = @_; for parameter parsing in methods

DeepthiYV:
We are seeing the mix between my $self = shift and my ($self) = @_ for parameter parsing in functions, so the proposal would be to stick to my ($self) = @_

- Related ticket: https://progress.opensuse.org/issues/97097
- Needles: No
- Verification run:

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -47,8 +47,8 @@ C<<$instance->username()>>.
 Use argument C<rc_only> to only check for the return code of the command.
 =cut
 sub run_ssh_command {
-    my $self = shift;
-    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @_);
+    my ($self, @args) = @_;
+    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @args);
     die('Argument <cmd> missing') unless ($args{cmd});
     $args{ssh_opts} //= $self->ssh_opts() . " -i '" . $self->ssh_key . "'";
     $args{username} //= $self->username();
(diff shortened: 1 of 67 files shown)


b10n1k (changes requested):
this is not gonna fly.
1. there is not any way as for now to force this. said that, i mean you cant make sure that the next commits will not break this convention.
2. if you want to force it then automate it in CI
3. this is a breaking change. there are cases where you shoudl stick with `shift`. hard to find among 514 files but i have found one case in a publiccloud module

lib/publiccloud/instance.pm
```diff
@@ -47,7 +47,7 @@ C<<$instance->username()>>.
 Use argument C<rc_only> to only check for the return code of the command.
 =cut
 sub run_ssh_command {
-    my $self = shift;
+    my ($self) = @_;
     my %args = testapi::compat_args({cmd => undef}, ['cmd'], @_);
```
b10n1k, line 51: this is going to break the test.
think what happens when you run `run_ssh_command(cmd=>$my_cmd)`

DeepthiYV: Thank you @b10n1k .  Currently, I am still verifying the changes by executing few test suites. I am not sure how to check the above scenarios. 
I had automated these changes and will wait for the input to proceed.

foursixnine, line 51:
```suggestion
    my ($self, @args) = @_;
    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @args);
```

b10n1k: this is not always wanted in my opinion. it would work for the most of the cases but i have code which cant apply this for really. if somehow ci checks for this reduce flexibility. the only way i am thinking it can fly is through code review where everyone one know that this is the main rule but it can accept a different implementation when this is required. but it needs everyone to share this policy equally.  just my opinion.

foursixnine: @b10n1k well it's already in the contributing file, and this pr is stepping stone towards moving to having CI checks (And it's not just this, but many other things, so that we can get to a more robust codebase).

> this is not always wanted in my opinion.

Can you explain why?

PR #13328, opened 2021-09-24
Fix terraform destroy condition

ilausuch:
A bug was introduced by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13202
creating a failure in HA/SAP cleanup.

With this PR the terraform_destroy only will be omitted in case of EKS
tests

- Related ticket: progress.opensuse.org/issues/97715
- Verification run: http://copland.arch.suse.de/tests/273

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -634,7 +634,7 @@ This method is called called after each test on failure or success.
 =cut
 sub cleanup {
     my ($self) = @_;
-    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform');
+    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform') && check_var('PUBLIC_CLOUD_SERVICE', 'EKS');
     $self->vault_revoke();
     assert_script_run "cd";
 }


pdostal (approved):

jlausuch (approved):

lib/publiccloud/provider.pm
```diff
@@ -634,7 +634,7 @@ This method is called called after each test on failure or success.
 =cut
 sub cleanup {
     my ($self) = @_;
-    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform');
+    $self->terraform_destroy() unless check_var('PUBLIC_CLOUD_SERVICE', 'EKS');
```
asmorodskyi, line 637: I am not sure why you drop initial check , I think this would be correct version 
```suggestion
    $self->terraform_destroy() unless script_run('test -d ' . TERRAFORM_DIR . '/.terraform') && check_var('PUBLIC_CLOUD_SERVICE', 'EKS');
```

ilausuch: The terraform dir check was introduced by the related PR so, I don't think is necessary, but yes, is more complete.

ilausuch: Added

ldevulder: @asmorodskyi Why conditioning the destroy to this `.terraform` directory? Maybe my lab for manual test is bad, but I always have this directory, before and after an apply or destroy command. In that case the destroy in cleanup is never done.

ldevulder:
Conditioning the check to EKS works for me: http://1b210.qa.suse.de/tests/9263#step/sles4sap/391

lib/publiccloud/provider.pm
asmorodskyi: you are right , this is basically what we trying to fix . sorry I was not careful enough . @ilausuch  please revert to your initial version sorry for confusing you

ldevulder (approved):
LGTM

PR #13289, opened 2021-09-20
publiccloud: Add region variable to 'terraform destroy'

cfconrad:
With provider AWS, the destroy of the instance doesn't worked, even if
the output of terraform assume it.
```
aws_instance.openqa[0]: Destroying... [id=i-0056a077e11402492]
aws_instance.openqa[0]: Destruction complete after 0s
```
Only the time of 0s could indicate something wrong.

With this change the destruction take some time and the output looks
like:
```
aws_instance.openqa[0]: Destroying... [id=i-032c27915d1f99aab]
aws_instance.openqa[0]: Still destroying... [id=i-032c27915d1f99aab, 10s elapsed]
aws_instance.openqa[0]: Still destroying... [id=i-032c27915d1f99aab, 20s elapsed]
aws_instance.openqa[0]: Still destroying... [id=i-032c27915d1f99aab, 30s elapsed]
aws_instance.openqa[0]: Still destroying... [id=i-032c27915d1f99aab, 40s elapsed]
aws_instance.openqa[0]: Destruction complete after 41s
```

The issue appears with the current used version 3.37.0 and also with
latest 3.59.0 of terraforms AWS provider.

Ticked: https://progress.opensuse.org/issues/63604

- Verification run: 
  - https://openqa.suse.de/tests/7173555#step/run_ltp/134 (GCE)
  - https://openqa.suse.de/tests/7173554#step/run_ltp/173 (EC2)
  - https://openqa.suse.de/tests/7182287#step/run_ltp/150 (Azure)

PR #13329, opened 2021-09-24
publiccloud: Do not add region for PUBLIC_CLOUD_SLES4SAP on destroy

cfconrad:
This is a follow up for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13289
and try to address the issue with SLES4SAP terraform profiles.

- Ticked: https://progress.opensuse.org/issues/63604
- Verification run: http://1b210.qa.suse.de/t9265

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -441,18 +441,21 @@ Destroys the current terraform deployment
 =cut
 sub terraform_destroy {
     my ($self) = @_;
+    my $cmd;
     return if get_required_var('TEST') =~ /^publiccloud_upload_img/;
     record_info('INFO', 'Removing terraform plan...');
     if (get_var('PUBLIC_CLOUD_SLES4SAP')) {
         assert_script_run('cd ' . TERRAFORM_DIR . '/' . $self->conv_openqa_tf_name);
+        $cmd = 'terraform destroy -no-color -auto-approve';
     }
     else {
         assert_script_run('cd ' . TERRAFORM_DIR);
+        # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
+        $cmd = sprintf(q(terraform destroy -no-color -auto-approve -var 'region=%s'), $self->region);
     }
     # Retry 3 times with considerable delay. This has been introduced due to poo#95932 (RetryableError)
     # terraform keeps track of the allocated and destroyed resources, so its safe to run this multiple times.
-    # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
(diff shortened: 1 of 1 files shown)


lib/publiccloud/provider.pm
```diff
@@ -441,18 +441,21 @@ Destroys the current terraform deployment
 =cut
 sub terraform_destroy {
     my ($self) = @_;
+    my $cmd;
```
asmorodskyi, line 444:
```suggestion
    my $cmd = 'terraform destroy -no-color -auto-approve';
```

asmorodskyi:
but this would basically means that `destroy` will be still broken for SAP/HA because as @cfconrad  discovered in case of AWS region is *required* for destroy call ?

cfconrad:
They are using a tfvars file. And in that file the region is specified. 
From https://www.terraform.io/docs/language/values/variables.html#variable-definitions-tfvars-files you can see that a file which is called `terraform.tfvars` get's loaded automatically.

So I think it should work!

lib/publiccloud/provider.pm
```diff
@@ -441,18 +441,21 @@ Destroys the current terraform deployment
...
     else {
         assert_script_run('cd ' . TERRAFORM_DIR);
+        # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
+        $cmd = sprintf(q(terraform destroy -no-color -auto-approve -var 'region=%s'), $self->region);
     }
     # Retry 3 times with considerable delay. This has been introduced due to poo#95932 (RetryableError)
     # terraform keeps track of the allocated and destroyed resources, so its safe to run this multiple times.
-    # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
-    my $ret = script_retry(sprintf(q(terraform destroy -no-color -auto-approve -var 'region=%s'), $self->region), retry => 3, delay => 60, timeout => get_var('TERRAFORM_TIMEOUT', TERRAFORM_TIMEOUT), die => 0);
+    my $ret = script_retry($cmd, delay => 60, timeout => get_var('TERRAFORM_TIMEOUT', TERRAFORM_TIMEOUT), die => 0);
```
ldevulder, line 458: The `retry => 3` option has been removed, is it expected? If so then the comment above about this retry should also be removed.
If this is not expected then I will re-add this option in PR #13338.

asmorodskyi: nice catch , yeah I don't see good reason why to remove it , so please add it where you suggested

cfconrad: nice catch, thx

PR #13331, opened 2021-09-27
Fix gnome_session_switch failure on x11

GraceWang571:
- Related ticket: https://progress.opensuse.org/issues/97916
- Needles: already added to o.s.d when debugging
- Verification run: https://openqa.suse.de/tests/7239211


zhaocongjia:
LGTM.  This is another corner case introduced by GNOME 40 upgrade.

PR #13338, opened 2021-09-27
Allow HA/SAP PC tests to use latest SHAP tool version

ldevulder:
The changes in this commit allow the use of the latest SHAP tool version without patching it.

This commit also remove fixed bsc# workarounds.

- Related ticket: N/A
- Needles: N/A
- Merge Request: https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/438
- Verification run: http://1b210.qa.suse.de/tests/overview?build=0081&distri=sle&version=15-SP3&groupid=17 
Note: upload tests were not re-triggered, and Azure issue is because of an error in a Salt formula (fixed here: https://github.com/SUSE/habootstrap-formula/pull/86)


lib/publiccloud/provider.pm
```diff
@@ -444,18 +447,25 @@ sub terraform_destroy {
...
     }
     else {
         assert_script_run('cd ' . TERRAFORM_DIR);
-        # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
-        $cmd = sprintf(q(terraform destroy -no-color -auto-approve -var 'region=%s'), $self->region);
     }
+
+    # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
+    my $region_var  = get_var('PUBLIC_CLOUD_SLES4SAP') ? 'aws_region' : 'region';
+    my $destroy_cmd = sprintf(q(terraform destroy -no-color -auto-approve -var '%s=%s'), $region_var, $self->region);
```
cfconrad, line 459: Why this extra complexity? Is `terraform destroy` not using `terraform.tfvars` file?

lib/publiccloud/provider.pm
```diff
@@ -444,18 +447,25 @@ sub terraform_destroy {
     my $cmd;
     return if get_required_var('TEST') =~ /^publiccloud_upload_img/;
     record_info('INFO', 'Removing terraform plan...');
-    if (get_var('PUBLIC_CLOUD_SLES4SAP')) {
+    f (get_var('PUBLIC_CLOUD_SLES4SAP')) {
```
cfconrad, line 450: looks like a ooops

ldevulder: Yes, bad copy/paste... thx!

cfconrad: And can we make all "SLES4SAP" conditions within this single if.. so afterwards we just have a `$destroy_cmd` which we will use.

ldevulder: Yes, will do the change.

cfconrad (changes requested):
I would like to limit the `if SLES4SAP`, cause I'm afraid that we will break it again if they are distributed in to many places.

lib/publiccloud/provider.pm
cfconrad: thx a lot

lib/publiccloud/provider.pm
ldevulder: What do you mean? It was the code used before, I just changed the `region` variable name.
As far as I understood the `region` variable should be forced on command line because of an issue with the AWS provider. Otherwise I agree with you: the `destroy` command uses the `terraform.tfvars` file.

ldevulder: I've tried to simplify.

lib/publiccloud/provider.pm
ldevulder: Done, let me know if it's ok for you :-)

lib/publiccloud/provider.pm
```diff
@@ -444,18 +448,23 @@ sub terraform_destroy {
...
     return if get_required_var('TEST') =~ /^publiccloud_upload_img/;
     record_info('INFO', 'Removing terraform plan...');
+
+    # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
     if (get_var('PUBLIC_CLOUD_SLES4SAP')) {
         assert_script_run('cd ' . TERRAFORM_DIR . '/' . $self->conv_openqa_tf_name);
-        $cmd = 'terraform destroy -no-color -auto-approve';
+
+        # On HA/SAP tests with ha-sap-terrafrom-deployments tool we shouldn't force the region for all providers
+        $cmd = sprintf(q(terraform destroy -no-color -auto-approve -var 'aws_region=%s'), $self->region);
```
ricardobranco777, line 457: Isn't there a way to avoid `sprintf` here?

lib/publiccloud/provider.pm
```diff
@@ -444,18 +448,23 @@ sub terraform_destroy {
...
+        $cmd = 'terraform destroy -no-color -auto-approve' unless check_var('PUBLIC_CLOUD_PROVIDER', 'EC2');
     }
     else {
         assert_script_run('cd ' . TERRAFORM_DIR);
-        # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
         $cmd = sprintf(q(terraform destroy -no-color -auto-approve -var 'region=%s'), $self->region);
```
ricardobranco777, line 462: ... and same here.

ricardobranco777 (approved):
LGTM.

I didn't know sprintf was available in Perl.

lib/publiccloud/provider.pm
ldevulder: There is, but why?

lib/publiccloud/provider.pm
ldevulder: And same answer/question :D

lib/publiccloud/provider.pm
ricardobranco777: Readability? 
I'm just prejudiced against `sprintf` for being insecure in C, but Perl is not C, so feel free to ignore my suggestion.

ricardobranco777: `$cmd = "terraform destroy -no-color -auto-approve -var aws_region=" . $self->region` should work.

ldevulder: @cfconrad Which syntax do you prefer?

lib/publiccloud/provider.pm
asmorodskyi: we can go other way around - because HA/SAP can not change their variable names we can align **our** variables to HA/SAP . @cfconrad  WDYT ? ( but it is for sure out of scope for this PR )

lib/publiccloud/provider.pm
asmorodskyi: I would say it is really matter of taste , I personally like more version @ricardobranco777  but I think you can decided on your own if you want to change it or not . It is pure "eye candy" :)

lib/publiccloud/provider.pm
asmorodskyi: same comment :D

lib/publiccloud/provider.pm
ldevulder: @asmorodskyi I'm not sure that this is a good idea, simply because the variables are set in another project (ha-sap-terraform-deployments) and can changes (and in fact there are changes in every major versions...) and these changes are not from us.
So maybe better to keep it as is ;-)

lib/publiccloud/provider.pm
ldevulder: As this line was with a `sprintf` before, and as there is other `sprintf` commands in this file I prefer to keep it as is.

@ricardobranco777 there are some chances that the next change will be from you, so feel free to change it next time! ;-)

asmorodskyi: @ricardobranco777 , @ldevulder , @cfconrad  WDYT about such version ? 
```
$cmd = 'terraform destroy -no-color -auto-approve';
$cmd .= ' -var 'aws_region=' . $self->region if is_ec2();
```

Note : you will need to export `is_ec2()` from https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/utils.pm

ldevulder: In that case it's better to do this I think: `$cmd .= " -var 'aws_region=$self->region'" if is_ec2;`, as the single quote is also used for the `aws_region`, and protected all is more complicated IMHO.

I'm doing a last change and I will re-trigger jobs.

lib/publiccloud/provider.pm
```diff
@@ -441,21 +446,25 @@ Destroys the current terraform deployment
 =cut
 sub terraform_destroy {
     my ($self) = @_;
-    my $cmd;
     return if get_required_var('TEST') =~ /^publiccloud_upload_img/;
     record_info('INFO', 'Removing terraform plan...');
+
+    # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
```
asmorodskyi, line 452: I would move this comment to line 462

ldevulder: Done.

lib/publiccloud/provider.pm
ldevulder: Verification runs are OK (again Azure could fail, because of another not related bug).

asmorodskyi (approved):
LGTM

lib/publiccloud/provider.pm
```diff
@@ -441,21 +446,26 @@ Destroys the current terraform deployment
...
     record_info('INFO', 'Removing terraform plan...');
+
+    my $cmd = 'terraform destroy -no-color -auto-approve';
     if (get_var('PUBLIC_CLOUD_SLES4SAP')) {
         assert_script_run('cd ' . TERRAFORM_DIR . '/' . $self->conv_openqa_tf_name);
-        $cmd = 'terraform destroy -no-color -auto-approve';
+
+        # Add region variable also to `terraform destroy` (poo#63604) -- needed by AWS.
+        # On HA/SAP tests with ha-sap-terrafrom-deployments tool we shouldn't force the region for all providers
+        $cmd .= " -var 'aws_region=$self->region'" if is_ec2;
```
alvarocarvajald, line 458: I read you already did some verification runs with this version, so this may be minor, but I think it would be better to have `$self->region` outside of the double quotes as in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L197.

Leaving it within the double quotes could lead to funny business:

```
susetest# perl
use Mojo::File qw(path);

my $a = path('/etc/os-release');
my $b = "FILE CONTENTS: [$a->slurp]\n";
print $b;
^D
FILE CONTENTS: [/etc/os-release->slurp]
susetest# perl
use Mojo::File qw(path);

my $a = path('/etc/os-release');
my $b = "FILE CONTENTS: [" . $a->slurp . "]\n";
print $b;
^D
FILE CONTENTS: [NAME="openSUSE Leap"
VERSION="15.1"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.1"
PRETTY_NAME="openSUSE Leap 15.1"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.1"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
]
```

alvarocarvajald (approved):
LGTM

lib/publiccloud/provider.pm
cfconrad: > What do you mean? It was the code used before, I just changed the `region` variable name.
> As far as I understood the `region` variable should be forced on command line because of an issue with the AWS provider. Otherwise I agree with you: the `destroy` command uses the `terraform.tfvars` file.

From my understanding, the problem on our side occur, because we have such a line: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/publiccloud/terraform/ec2.tf#L19 and with this provider section and an unset variable `region` it fails on destroy.
I just tried our tf file with a `terraform.tfvars` file containing only the region, it worked.

Without knowing your exact tf-file, I think that you have a "hardcoded" `terraform.tfvars` file, you do not need this extra variable.
So I wold like to avoid that extra complexity if it isn't needed. Otherwise nobody will understand it later and we never get rid of it.

@asmorodskyi I think using a `terraform.tfvars` file would help here, the downside is, we cannot just c&p the terraform command to retry it manually. But we could also upload the `tfvars` file as the hasap tests do it.

lib/publiccloud/provider.pm
cfconrad: There is this interpolation syntax
```perl
my $b = "FILE CONTENTS: [${\ $a->slurp() }]";
```
which also works, but I'm not a fan of it :)

lib/publiccloud/provider.pm
ldevulder: @cfconrad ok, so yes in that case I will revert back to your version and retest to be sure. I thought it was a regression in AWS in all cases.

@ricardobranco777 @asmorodskyi if I revert back to the previous version I will let the `sprintf` as it was already here before ;-)

lib/publiccloud/provider.pm
ldevulder: @alvarocarvajald you're right, it's not good... But as @cfconrad said in fact this change is not needed for our HA/SAP tests, so I will simply revert these specific changes.

alvarocarvajald: >There is this interpolation syntax
>which also works, but I'm not a fan of it :)

Yeah, me neither :P

> @alvarocarvajald you're right, it's not good... But as @cfconrad said in fact this change is not needed for our HA/SAP tests, so I will simply revert these specific changes.

Oh, cool. Better yet.

ldevulder:
Latest verification runs are OK (still issues in Azure for another reason).

cfconrad (approved):
LGTM

PR #13339, opened 2021-09-27
[WIP] Executes a job in the EKS

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/97724
- Needles:


ilausuch:
I am refactoring this code

lib/publiccloud/eks.pm
```diff
@@ -15,6 +15,18 @@
...
 use Mojo::Base 'publiccloud::aws';
 use testapi;
+use utils;
+use publiccloud::k8s_utils;
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+
+    my $repository //= get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REPO", 'suse-qec-testing');
```
jlausuch, line 25: I guess you don't need `\\` here.
Since $repository is not an argument, you declare it here as a variable and uses `get_var`, it will take either `PUBLIC_CLOUD_CONTAINER_IMAGES_REPO` or `suse-qec-testing`

lib/publiccloud/eks.pm
```diff
@@ -15,6 +15,18 @@
...
+    $self->{repository} = $repository;
+
+    assert_script_run("aws eks update-kubeconfig --name qe-c-testing");
```
jlausuch, line 28: How long does this take? Is default timeout of 90s enough? Maybe we need to define some timeout to avoid surprises?

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
...
+    assert_script_run("$env kubectl apply -f $filename");
+}
+
+=head2 get_kubectl_env
+Get the environment variables needed to call kubectl
+=cut
+sub get_kubectl_env {
+    my $eks_key    = get_required_var('PUBLIC_CLOUD_EKS_KEY');
+    my $eks_secret = get_required_var('PUBLIC_CLOUD_EKS_SECRET');
+    'AWS_ACCESS_KEY_ID="' . $eks_key . '" AWS_SECRET_ACCESS_KEY="' . $eks_secret . '"';
```
jlausuch, line 70: Don't you need to `return` this string? (maybe not, my perl knowledge is limited and this is enough)

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
...
+    my $env      = get_kubectl_env();
+    my $filename = create_manifest($manifest);
+
+    assert_script_run("$env kubectl apply -f $filename");
+}
+
+=head2 get_kubectl_env
+Get the environment variables needed to call kubectl
+=cut
+sub get_kubectl_env {
```
jlausuch, line 67: Would it make sense to move this function to `k8s_utils.pm`?

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
...
+    my $eks_secret = get_required_var('PUBLIC_CLOUD_EKS_SECRET');
+    'AWS_ACCESS_KEY_ID="' . $eks_key . '" AWS_SECRET_ACCESS_KEY="' . $eks_secret . '"';
+}
+
+=head2 get_container_image_name
+Get the full name for a container image
+=cut
+sub get_container_image_name {
+    my ($self, $tag) = @_;
+    my $region //= $self->region();
```
jlausuch, line 78:
```suggestion
    my $region = $self->region();
```
or actually use `$self->region` later in the code.

jlausuch: Also, why is `region` a function here? Shouldn't it be `$self->region` ?

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
...
+    my $aws_account_id   = $self->{aws_account_id};
+    my $full_name_prefix = "$aws_account_id.dkr.ecr.$region.amazonaws.com";
```
jlausuch, line 80:
```suggestion
    my $full_name_prefix = "$self->{aws_account_id}.dkr.ecr.$self->region.amazonaws.com";
```

jlausuch: I would add a `record_info` with the value of this variable, it will help debugging.

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
...
+    "$full_name_prefix/$self->{repository}:$tag";
+}
+
+=head2 clean_image
+Delete a ECR image
+=cut
+sub clean_image {
```
jlausuch, line 87: I would name it `delete_image` or `remove_image` as you define above, or how the command sounds `batch-delete-image`

lib/publiccloud/k8s_utils.pm
```diff
@@ -0,0 +1,61 @@
...
+use warnings;
+use strict;
+
+our @EXPORT = qw(install_kubectl wait_for_container_log create_manifest);
+
+=head2 install_kubectl
+Install kubectl from the k8s page
+=cut
+sub install_kubectl {
+  my $version = "v1.22.2";
```
jlausuch, line 26: To allow some flexibility, you can let the user choose the version:
```suggestion
  my $version = get_var('KUBECTL_VERSION', 'v1.22.2');
```

lib/publiccloud/k8s_utils.pm
```diff
@@ -0,0 +1,61 @@
...
+  assert_script_run("curl -LO https://dl.k8s.io/release/$version/bin/linux/amd64/kubectl");
+  assert_script_run("chmod a+x kubectl");
+  assert_script_run("rm /usr/bin/kubectl");
+  assert_script_run("mv kubectl /usr/bin/kubectl");
+}
+
+=head2 wait_for_container_log
+Waits for an especific log outut or die (afeter a timeout)
```
jlausuch, line 34:
```suggestion
Waits for an especific log outut or die (after a timeout)
```

lib/publiccloud/k8s_utils.pm
```diff
@@ -0,0 +1,61 @@
...
+Creates a manifest from a string content
+=cut
+sub create_manifest {
+    my ($content) = @_;
+    my $filename = `cat /proc/sys/kernel/random/uuid | tr -d '\n'`.'.yaml';
+    my $path = "/tmp/$filename";
+
+    save_tmp_file($filename, $content);
+    assert_script_run(sprintf('curl -o "%s" "%s/files/%s"', $path, autoinst_url, $filename));
+    $path;
```
jlausuch, line 58: What does this do?

schedule/containers/sle_image_on_amazon_eks.yaml
```diff
@@ -5,3 +5,4 @@ description: |
 schedule:
   - boot/boot_to_desktop
   - containers/upload_container_images_to_ecr
+  - containers/execute_job_in_eks
```
jlausuch, line 8: maybe `run_container_in_eks` ?

lib/publiccloud/k8s_utils.pm
```diff
@@ -0,0 +1,61 @@
...
+=head2 create_manifest
+Creates a manifest from a string content
+=cut
+sub create_manifest {
+    my ($content) = @_;
+    my $filename = `cat /proc/sys/kernel/random/uuid | tr -d '\n'`.'.yaml';
+    my $path = "/tmp/$filename";
+
+    save_tmp_file($filename, $content);
+    assert_script_run(sprintf('curl -o "%s" "%s/files/%s"', $path, autoinst_url, $filename));
```
pdostal, line 57: Please place the file you download under `data/containers/` as all the others..

tests/containers/execute_job_in_eks.pm
```diff
@@ -0,0 +1,67 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      - name: main
+        image: $image
+        command: [ "cat", "/etc/os-release" ]
+      restartPolicy: Never
+  backoffLimit: 4
+EOT
```
pdostal, line 47: Please place this file into `data/containers` and then use `file_content_replace()` to place the variables.
```perl
file_content_replace("eks_manifest.yaml", IMAGE => $image);
```

I also suggest to modify `eks_apply_manifest()` so it accepts the YAML file name - not directly content.

pdostal (commented):
Hello Ivan, I really like this! I have only few small suggestions:

lib/publiccloud/eks.pm
cfconrad: @jlausuch this interpolation doesn't work. A member function need some different syntax, see https://www.oreilly.com/library/view/perl-cookbook/1565922433/ch01s11.html.
But I would still consider `sprintf` or normal string concatenation :)
```suggestion
    my $full_name_prefix = sprintf('%s.dkr.ecr.%s.amazonaws.com', $aws_account_id, $region);
```

lib/publiccloud/eks.pm
```diff
@@ -24,26 +36,57 @@ Retrieves the full name of the uploaded image or die.
 =cut
 sub push_container_image {
     my ($self, $image, $tag) = @_;
+    $self->{tag} = $tag;
```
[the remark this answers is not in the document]
cfconrad: I see it now in `clean_image()`, but if this is a variable which should not change during the usage of one instance of `publiccloud::eks`, then I would also move this as member and set it on the constructor!

lib/publiccloud/k8s_utils.pm
```diff
@@ -0,0 +1,61 @@
+# SUSE's openQA tests
+#
+# Copyright © 2021 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Summary: Base class for publiccloud kubernetes tests
```
cfconrad, line 10: If it is a baseclass, lets call it `k8sbasetests` or similar not utils.

lib/publiccloud/k8s_utils.pm
cfconrad: https://perldoc.perl.org/functions/return#return1
```
(In the absence of an explicit return, a subroutine, eval, or do FILE automatically returns the value of the last expression evaluated.)
```

tests/containers/execute_job_in_eks.pm
```diff
@@ -0,0 +1,67 @@
...
+
+sub run {
+    my ($self, $args) = @_;
+
+    $self->select_serial_terminal;
+    install_kubectl();
+
+    my $provider      = $self->provider_factory();
+    my $tag           = join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), get_current_job_id());
+    my $job_name      = $tag =~ s/_/-/r;
```
cfconrad, line 27:
```suggestion
    my $job_name      = $tag =~ s/_/-/gr;
```
you need to replace all `_`

tests/containers/execute_job_in_eks.pm
```diff
@@ -0,0 +1,67 @@
...
+    my $env           = $provider->get_kubectl_env();
+    my $image         = $provider->get_container_image_name($tag);
+    $self->{provider} = $provider;
+    $self->{job_name} = $job_name;
+    $provider->{tag}  = $tag;
```
cfconrad, line 32: I really would move this in the `factory` :)

tests/containers/execute_job_in_eks.pm
cfconrad: I would only move the file inside of `data/containers` if it is used twice... Otherwise I think it is good practice to keep the infos of *one* test in *one* place.

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -23,11 +23,18 @@ sub run {
 
     my ($untested_images, $released_images) = get_suse_container_urls();
 
-    my $provider = $self->provider_factory();
-    my $image    = $untested_images->[0];
-    my $tag      = join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), get_current_job_id());
+    my $provider      = $self->provider_factory();
+    $self->{provider} = $provider;
```
cfconrad, line 27: isn't this already done by `provider_factory()`?

tests/containers/upload_container_images_to_ecr.pm
```diff
@@ -23,11 +23,18 @@ sub run {
...
+
+    my $image = $untested_images->[0];
+    my $tag   = join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME'), get_current_job_id());
     assert_script_run("docker pull $image", 360);
     $provider->push_container_image($image, $tag);
 }
 
+sub post_fail_hook {
```
cfconrad, line 35: The idea of `publiccloud:basetest()` is to use `cleanup()` when you simply need a function which is called in any case after the tests `run()` finished, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/basetest.pm#L81

tests/containers/upload_container_images_to_ecr.pm
pdostal: This `$self` is used when you need to preserve the variable into other tests.

cfconrad: @pdostal not sure, normally we use the args from the `run()` method to move the provider from one to the other test.
I think `$self->{provider}` is needed to do a proper cleanup in `cleanup()` function.

pdostal: Ah, true. My statement would be valit for `$args` not for `$self`.

lib/publiccloud/k8s_utils.pm
jlausuch: Ah ok, just that I'm too used to "return" :)

lib/publiccloud/eks.pm
jlausuch: Yes, I know that. But I thought `region` is a parameter, not a function.

PR #13340, opened 2021-09-27
Increase timeout to return from rescue to default target

mloviska:
- [[xen-pv] snapper_undochange - snapper_nodbus_restore does not work properly](https://progress.opensuse.org/issues/68815)
- VRs: [sle15sp2@xen-pv](http://kepler.suse.cz/tests/7035#live)


mloviska:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13374

PR #13344, opened 2021-09-28
Fix sporadic issue of tomcat on Tumbleweed

Zaoliang:
see https://progress.opensuse.org/issues/96611
Increase timeout for check_screen('tomcat-click-save-login')
and add still_wait_screen after send_key('ret')
verifications:
http://10.162.30.85/tests/3719#next_previous


Zaoliang:
https://openqa.opensuse.org/tests/1943050#next_previous looks good

PR #13348, opened 2021-09-28
Replace grep -q pipelines in lib/hacluster

alvarocarvajald:
Starting with SLES+HA 15-SP4 Alpha build 39.1, some pipelines of the type
`cmd | grep -q` in `lib/hacluster` started failing with retval 141 (bash
generates a *SIGPIPE* as the left hand side command is still writing to the
pipe while the right hand side command has exited) or retval 1 (no
match), even when a match is expected. These commands were changed to
`grep -q <(cmd)` to avoid using the pipe.

- Related ticket: N/A
- Needles: N/A
- Failing tests: https://openqa.suse.de/tests/7242139#step/vg/5, https://openqa.suse.de/tests/7251378#step/clvmd_lvmlockd/3, https://openqa.suse.de/tests/7241884#step/migrate_clvmd_to_lvmlockd/11
- Verification run: [node 1](http://mango.qa.suse.de/tests/4423) & [node 2](http://mango.qa.suse.de/tests/4424)

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -361,7 +361,7 @@ true or false.
 
 sub check_rsc {
     my $rsc = shift;
-    my $ret = script_run "$crm_mon_cmd 2>/dev/null | grep -q '\\<$rsc\\>'";
+    my $ret = script_run "grep -q '\\<$rsc\\>' <($crm_mon_cmd 2>/dev/null)";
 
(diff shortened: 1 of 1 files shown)


ldevulder (approved):
LGTM

ricardobranco777 (approved):
LGTM

PR #13352, opened 2021-09-29
Modify check_kernel_config test cases to run on Tumbleweed

ShawnHa0:
According to Factory First, all test cases should run on Tumbleweed. So, we need to modify the test cases to make sure these cases can run and pass on Tumbleweed. Cases modified are under security_check_kernel_config, added version specification checking to fit Tumbleweed and SLE since there are some differentials in files and configurations between Tumbleweed and SLE.

- Related ticket: 
  https://progress.opensuse.org/issues/99102
  https://progress.opensuse.org/issues/99093
- Needles: N/A
- Verification run: 
  https://openqa.opensuse.org/tests/1942564
  https://openqa.suse.de/tests/7252012
  https://openqa.suse.de/tests/7252014

### DIFF
--- a/tests/security/check_kernel_config/CC_STACKPROTECTOR_STRONG.pm
+++ b/tests/security/check_kernel_config/CC_STACKPROTECTOR_STRONG.pm
@@ -25,13 +25,20 @@ use warnings;
 use testapi;
 use Utils::Architectures;
 use utils;
+use version_utils 'is_sle';
 
 sub run {
     # check the kernel configuration file to make sure the parameter is there
     if (!is_s390x) {
-        validate_script_output "cat /boot/config-`uname -r`|grep CONFIG_STACKPROTECTOR", qr/CONFIG_STACKPROTECTOR_STRONG=y/;
+        if (is_sle) {
+            validate_script_output "cat /boot/config-`uname -r`| grep CONFIG_STACKPROTECTOR", qr/CONFIG_STACKPROTECTOR_STRONG=y/;
+        }
     }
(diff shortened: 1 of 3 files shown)


rfan1:
Remove this line:

rfan1:
If possible, can you please also run the same tests on s390x and Power?

ShawnHa0:
> If possible, can you please also run the same tests on s390x and Power?

It seems that there are no available job to run this case on s390x and Power. If we need to run on these two platforms, maybe it is mandatory to rerun the case on s390x and generate the qcow (referring to Xiaojing yesterday)

rfan1:
> > If possible, can you please also run the same tests on s390x and Power?
> 
> It seems that there are no available job to run this case on s390x and Power. If we need to run on these two platforms, maybe it is mandatory to rerun the case on s390x and generate the qcow (referring to Xiaojing yesterday)

No need to run on O3, just run the tests on OSD should be fine :)

ShawnHa0:
> > > If possible, can you please also run the same tests on s390x and Power?
> > 
> > 
> > It seems that there are no available job to run this case on s390x and Power. If we need to run on these two platforms, maybe it is mandatory to rerun the case on s390x and generate the qcow (referring to Xiaojing yesterday)
> 
> No need to run on O3, just run the tests on OSD should be fine :)

OK, I will consult Xiaojing about how to run these cases on s390x and Power.

ShawnHa0:
Will close this one and open a new pull request with code change we discussed today.

PR #13353, opened 2021-09-29
Fix 'Printer Added' notification banner after migration

hjluo:
If we run cups service check before migration, after migration,
sometimes it will show the gnome notification banner. This will
cause assert_screen 'generic-desktop' to fail. We need to close the
gnome notification banner before the migration.

- Related ticket: https://progress.opensuse.org/issues/98799
- Needles: N/A
- Verification run: 
online:
 https://openqa.nue.suse.com/t7252329 
 https://openqa.nue.suse.com/t7252330
 https://openqa.nue.suse.com/t7259898 
 https://openqa.nue.suse.com/t7259899
offline:
https://openqa.nue.suse.com/t7260573

### DIFF
--- a/tests/migration/online_migration/pre_migration.pm
+++ b/tests/migration/online_migration/pre_migration.pm
@@ -18,7 +18,8 @@ use testapi;
 use Utils::Architectures;
 use utils;
 use migration;
-use version_utils 'is_sle';
+use version_utils;
+use x11utils 'turn_off_gnome_show_banner';
 
 sub check_or_install_packages {
     if (get_var("FULL_UPDATE") || get_var("MINIMAL_UPDATE")) {
@@ -61,6 +62,11 @@ sub run {
     remove_kgraft_patch if is_sle('<15');
     # create btrfs subvolume for aarch64 before migration
     create_btrfs_subvolume() if (is_aarch64);
+    # We need to close gnome notification banner before migration.
+    if (check_var('DESKTOP', 'gnome') && (is_aarch64 || is_ppc64le)) {
+        select_console 'user-console';
+        turn_off_gnome_show_banner;
+    }
 }
 
 sub test_flags {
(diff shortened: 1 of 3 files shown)


tests/migration/online_migration/pre_migration.pm
```diff
@@ -61,6 +62,9 @@ sub run {
     remove_kgraft_patch if is_sle('<15');
     # create btrfs subvolume for aarch64 before migration
     create_btrfs_subvolume() if (is_aarch64);
+    # We need to close gnome notification banner before migration.
+    select_console 'user-console';
+    turn_off_gnome_show_banner if check_var('DESKTOP', 'gnome');
```
lemon-suse, line 67: Is this for specific arch? If so please add the corresponding condition.

hjluo: > Is this for specific arch? If so please add the corresponding condition.

No,  now both aarch and ppc64le will hit this issue.

lemon-suse: So this is still on specific arches. limit the fix for aarch64 and ppc64le except you can prove it is needed for other arches also.

lemon-suse:
Could you add comments for the verification logs?  so that reviewer can know what the log is for before open it.

tinawang123 (approved):
LGTM

tests/migration/online_migration/pre_migration.pm
hjluo: Changed.

tests/migration/reboot_to_upgrade.pm
```diff
@@ -22,10 +22,15 @@ use utils;
 use Utils::Backends 'is_pvm';
 use power_action_utils 'power_action';
 use bootloader_setup 'stop_grub_timeout';
+use version_utils;
+use x11utils 'turn_off_gnome_show_banner';
 
 sub run {
     my ($self) = @_;
 
+    # We need to close gnome notification banner before migration.
```
lemon-suse, line 31: You'd better to include the updated code in the condition to ensure won't run on other arches.

lemon-suse (approved):

tests/migration/reboot_to_upgrade.pm
hjluo: changed.

PR #13358, opened 2021-09-29
Workaround migration bsc#1184347

dzedro:
- Related ticket: https://progress.opensuse.org/issues/97772
- Verification run: http://dzedro.suse.cz/tests/19488#step/zypper_migration/12


juadk (approved):
Thank you so much for this. As I told you, I didn't think just ignoring the issue to let the migration continue will be enough :smiley:

PR #13367, opened 2021-09-30
Containers: keep a single autoyast profile for all the hdd jobs

jlausuch:
The idea is to use this profile for ALL our container jobs.


data/autoyast_containers/autoyast_containers.xml.ep
```diff
@@ -0,0 +1,232 @@
...
+  </networking>
+  <software>
+    <products config:type="list">
+      <product>SLES</product>
+    </products>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <packages config:type="list">
+      <package>iputils</package>
+      <package>acl</package>
+      % if ($check_var->('INSTALL_DOCKER', '1')) {
```
b10n1k, line 88: I would like to ask you to make this variables a bit more generic (`DOCKER=1` `PODMAN=1`. That will give us some flexibility. for instance we can remove `CONTAINER_RUNTIME` and make use of those generic ones.

jlausuch: I would prefer to keep `CONTAINER_RUNTIME` as it really defines what it's made for.

`INSTALL_DOCKER` yeah... it's maybe too generic and I didnt have a better idea.Maybe I can reuse `CONTAINER_RUNTIME` for this..

mloviska: I actually find this `INSTALL_{DOCKER,PODMAN}` self-explanatory in deploy VM scenario with certain packages and modules.

jlausuch: if we want to be even more specific, what about `AUTOYAST_INSTALL_DOCKER=1`  and `AUTOYAST_INSTALL_PODMAN=1`

mloviska: Yes, that works, too. Given the context, `INSTALL_$runtine` is understandable well enough. Also, it is versatile as it can be used outside AY. :)

jlausuch:
VRs:
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=43.1&groupid=313&test=create_hdd_autoyast_containers

https://openqa.suse.de/tests/overview?flavor=Container-hdd&groupid=377&arch=s390x

The only issue I have is with 15-SP1 installation https://openqa.suse.de/tests/7287375#step/installation/12
which throws a kernel panic.... not sure if it's realted to the autoyast or what. But I would prefer to merge this and investigate that separately.

data/containers/autoyast_containers.xml.ep
```diff
@@ -79,6 +77,26 @@
     <keep_install_network config:type="boolean">true</keep_install_network>
     % }
   </networking>
+  % if ($check_var->('VERSION', '15-SP4')) {
```
jlausuch, line 80: For some reason, in 15-SP4 we need to have this enabled, otherwise s390x can't login to the VM afterwards... 
e.g. without this block: https://openqa.suse.de/tests/7287263#step/grub_set_bootargs/1

data/containers/autoyast_containers.xml.ep
```diff
@@ -53,13 +57,7 @@
       % } else {
       <append>splash=verbose</append>
       % }
-      % if ($check_var->('ARCH', 'aarch64')) {
-      <timeout config:type="integer">1</timeout>
-      % } elsif ($check_var->('ARCH', 's390x')) {
-      <timeout config:type="integer">15</timeout>
-      % } else {
       <timeout config:type="integer">-1</timeout>
```
jlausuch, line 60: this way we can omit `KEEP_GRUB_TIMEOUT` in all archs.

data/containers/autoyast_containers.xml.ep
```diff
@@ -29,8 +19,22 @@
...
+        % } else {
+        <version>{{VERSION}}</version>
+        % }
+        <arch>{{ARCH}}</arch>
+      </addon>
     </addons>
   </suse_register>
+  <general>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
```
jlausuch, line 35: needed to avoid `AUTOYAST_CONFIRM=1` which is not 100% reliable.

data/containers/autoyast_containers.xml.ep
```diff
@@ -29,8 +19,22 @@
         <arch>{{ARCH}}</arch>
       </addon>
       % }
+      <addon>
+        <name>sle-module-containers</name>
+        % if ($get_var->('VERSION') =~ /12-SP5|12-SP4|12-SP3/) {
+        <version>12</version>
```
jlausuch, line 25: Before you ask, for all the 12-SPx versions, we need to hardcode `12` here. Using the variable fails, e.g. `<version>12.5</version>` will through an error (see scc.suse.com for more info).

data/autoyast_containers/autoyast_containers.xml.ep
b10n1k: `INSTALL_$runtine` is self-explanatory indeed. just my point was to use just `PODMAN` and `DOCKER` which their present in the variables can indicate clearly the scope for all the actions. However i would like to resolve this and focus in other things in the PR. feel free to reopen it if you think it needs to.

PR #13368, opened 2021-09-30
Reformat code to current perltidy proposal

pevik:
[proposal](https://github.com/os-autoinst/os-autoinst/pull/1787) to use perltidy with --only-changed [1], leads to massive whitespace change.


foursixnine:
Looks ok to me

OleksandrOrlov:
LGTM

pevik:
Rebased to current master (9c290f9f6428fb3fb6e5635826286ae6a7ee6066).

pevik:
Good point. TBH I'm not sure how CI behave without it, I suppose it will be needed.

foursixnine:
> First let's clarify one thing - this PR is not required . Even without it CI pipeline will pass after merge of [os-autoinst/os-autoinst#1787](https://github.com/os-autoinst/os-autoinst/pull/1787) because we just _removed_ requirement for alignment but haven't add another which force people to remove it , right ? Or am I missing something ?

Quoting @coolo: I think you're wrong there. If that was the case, tidy wouldn't reformat the code

pevik:
Yes, this reformatting will be needed if we accept https://github.com/os-autoinst/os-autoinst/pull/1787.

okurz:
As https://github.com/os-autoinst/os-autoinst/pull/1787 was merged this is now effectively needed

PR #13372, opened 2021-10-01
Do not test ristretto on Rescue CDs

DimStar77:
https://progress.opensuse.org/issues/99585

- Related ticket: https://progress.opensuse.org/issues/99585
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/1949278 - FAIL :(


DimStar77:
the rescuecd tests seem not to set the RESCUECD variable, which is why the VR failed

PR #13376, opened 2021-10-01
Reduce timeout for qa_run jobs

michaelgrifalconi:
Change timeout from 3h to 45m

- Related ticket: https://progress.opensuse.org/issues/99591
- Needles:  no needles
- Verification run: jobs are failing anyway. This is to make them fail faster and not keep worker busy for nothing. In case more failures arises due to this change, can be reverted or re-adjusted.

### DIFF
--- a/tests/qa_automation/qa_run.pm
+++ b/tests/qa_automation/qa_run.pm
@@ -143,7 +143,7 @@ sub run {
     $self->prepare_repos();
 
     $self->start_testrun();
-    my $testrun_finished = $self->wait_testrun(timeout => 180 * 60);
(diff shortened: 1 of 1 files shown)


tjyrinki (approved):
This would be ok for now at least. I've noted in the ticket that a new ticket should be filed for possibly restoring the value, but OTOH that is now not the immediate problem, the failing of the tests and reserving openQA workers is.

okurz:
Maybe you want to have the timeout configurable by test setting, keep the old value as default and only set the test setting explicitly to a lower value within the scope of your team?

PR #13378, opened 2021-10-03
Containers: minor adjustments in autoyast profile

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/93817

- VR: https://openqa.suse.de/tests/overview?flavor=Container-Host&groupid=377&build=GM
- 15-SP4: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=43.1&groupid=313&test=create_hdd_autoyast_containers

### DIFF
--- a/data/containers/autoyast_containers.xml.ep
+++ b/data/containers/autoyast_containers.xml.ep
@@ -3,11 +3,14 @@
 <profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
   <suse_register>
     <do_registration config:type="boolean">true</do_registration>
+    % unless ($check_var->('VERSION', '15-SP4')) {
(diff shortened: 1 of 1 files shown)


jlausuch:
Needs this modification in some job groups: https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/448

pdostal (commented):
Do we really wanna have image with both `docker` and `podman` installed? They may easily interfere.

jlausuch:
> Do we really wanna have image with both `docker` and `podman` installed? They may easily interfere.

The idea is to have all preinstalled to save time in the tests. I don't see why docker and podman can work together. We already have tests in OSD and O3 that test docker and podman in the same job, they are not supposed to interfere, both should be ok to coexist. Actually if that doesn't work, then we have a serious product bug :)

cfconrad (approved):
LGTM :rocket:

PR #13379, opened 2021-10-04
lib/containers: Have get_container_logs write directly to the file

Vogtinator:
Just write it into the file directly instead of copying it manually.
This avoids various string escaping issues and improves performance.

- Verification run: https://openqa.opensuse.org/tests/1951903

`podman_3rd_party_images` went from 22m 24s back to 3m 10s.

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -201,17 +201,16 @@ sub info {
     $self->_engine_assert_script_run(sprintf("info %s %s", $property, $expected));
 }
 
-=head2 get_container_logs($container)
+=head2 get_container_logs($container, $filename)
 
 Request container's logs.
 C<container> the running container.
-C<logs> returns a string.
+C<filename> file the logs are written to.
 
 =cut
 sub get_container_logs {
-    my ($self, $container) = @_;
-    my $logs = $self->_engine_script_output("container logs $container");
-    return $logs;
+    my ($self, $container, $filename) = @_;
+    $self->_engine_assert_script_run("container logs $container | tee $filename");
 }
 
 =head2 remove_image($image_name)
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -202,8 +202,7 @@ sub test_container_image {
     return if $runtime->runtime eq 'buildah';
     $runtime->start_container('testing');
     $runtime->halt_container('testing');
-    my $logs = $runtime->get_container_logs('testing');
-    assert_script_run "echo '$logs' | tee '$logfile'";
+    $runtime->get_container_logs('testing', $logfile);
     $runtime->remove_container('testing');
     if (script_run("grep \"`uname -r`\" '$logfile'") != 0) {
         upload_logs("$logfile");


lib/containers/engine.pm
```diff
@@ -201,17 +201,16 @@ sub info {
...
-C<logs> returns a string.
+C<filename> file the logs are written to.
 
 =cut
 sub get_container_logs {
-    my ($self, $container) = @_;
-    my $logs = $self->_engine_script_output("container logs $container");
-    return $logs;
+    my ($self, $container, $filename) = @_;
+    $self->_engine_assert_script_run("container logs $container | tee $filename");
```
b10n1k, line 213: i would record_info in this case to make clear what have been done underneath the `get_container_logs` because the name doesnt implies that the logs will be found in `$filename`. or change the name if you dont mind?? wdyt?

Vogtinator: What do you suggest as new name?

Vogtinator: Alternatively, the logs could be handled entirely by the test code and not go through `grep` etc. on the SUT.

jlausuch:
I see now that serial output goes from

```
> podman container logs testing
> EOT_pmdCM
pmdCM-0-
# echo pmdCM; bash -oe pipefail /tmp/scriptpmdCM.sh ; echo SCRIPT_FINISHEDpmdCM-$?-
pmdCM
5.3.18-59.24-default
Heartbeat from registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3
  PID TTY          TIME CMD
    1 ?        00:00:00 sh
    8 ?        00:00:00 ps
SCRIPT_FINISHEDpmdCM-0-
# echo '5.3.18-59.24-default
> Heartbeat from registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3
>   PID TTY          TIME CMD
>     1 ?        00:00:00 sh
>     8 ?        00:00:00 ps' | tee '/var/tmp/container_logs'; echo BfHc5-$?-
5.3.18-59.24-default
Heartbeat from registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3
  PID TTY          TIME CMD
    1 ?        00:00:00 sh
    8 ?        00:00:00 ps
```

to

```
# podman container logs testing | tee /var/tmp/container_logs; echo yJqDz-$?-
5.3.18-59.24-default
Heartbeat from registry.opensuse.org/opensuse/leap/15.3/images/totest/containers/opensuse/leap:15.3
  PID TTY          TIME CMD
    1 ?        00:00:00 sh
    8 ?        00:00:00 ps
```

LGTM

lib/containers/engine.pm
jlausuch: But we were not doing the record_info before right? Not sure why it's needed now.

I don't mind about `$filename`... but if you want it more accurate, `$log_file` ?b

b10n1k: > What do you suggest as new name?

i dont want to force this (my comment was a humble observation). but i am gonna give a suggestion for anycase. i would say `write_log` or something in this "line"

Vogtinator:
I merged this for now to get the tests working properly again

PR #13380, opened 2021-10-04
Replace and fix output in podman_3rd_party_images

b10n1k:
the output of the script_output uses the instance variable where it is expected
the name of the runtime engine.

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -183,7 +183,7 @@ Returns an array ref with the names of the images.
 =cut
 sub get_images_by_repo_name {
     my ($self)      = @_;
-    my $repo_images = $self->_engine_script_output("images --format '{{.Repository}}'", timeout => 60);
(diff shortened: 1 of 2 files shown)


Vogtinator (changes requested):
LGTM for the actual change, but requesting changes because of the other diff.

Vogtinator:
What's the `get_images_by_repo_name` change for?

b10n1k:
> What's the `get_images_by_repo_name` change for?

http://aquarius.suse.cz/tests/7203#step/rootless_podman/105

Vogtinator (approved):
Changes look good, but the commit message doesn't mention the timeout change

jlausuch (approved):

b10n1k:
> Changes look good, but the commit message doesn't mention the timeout change

i know. i was about to do separate commit but i added that file in the first commit and i noticed afterwards. i thought it wouldnt be a big issue. but  i can add it

b10n1k:
> Changes look good, but the commit message doesn't mention the timeout change

Added

PR #13381, opened 2021-10-04
Increase container build timeout for non-x86 archs

grisu48:
Double the container build time on non-x86_64 architectures as those
sometime run into a timeout.

- Related ticket: https://progress.opensuse.org/issues/99177
- Verification run: [aach64](https://openqa.suse.de/tests/7311849) | [s390x](https://openqa.suse.de/tests/7311850) | [x86_64](https://openqa.suse.de/tests/7311851#step/docker_image/112) - relevant part passes

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -22,6 +22,7 @@ use strict;
 use warnings;
 use version_utils;
 use version;
+use Utils::Architectures;
 use containers::utils;
 use containers::common 'test_container_image';
 
@@ -66,7 +67,7 @@ sub build_and_run_image {
     assert_script_run "curl -f -v " . data_url('containers/index.html') . " > $dir/BuildTest/index.html";
 
     # At least on publiccloud, this image pull can take long and occasinally fails due to network issues
-    $builder->build($dir . "/BuildTest", "myapp", timeout => 600);
+    $builder->build($dir . "/BuildTest", "myapp", (timeout => is_x86_64 ? 600 : 1200));
     my $imgs = $builder->get_images_by_repo_name();
     die "myapp image not found in the local registry" unless (grep { $_ =~ /myapp/ } @{$imgs});
     assert_script_run("rm -rf $dir");


lib/containers/container_images.pm
```diff
@@ -66,7 +67,9 @@ sub build_and_run_image {
     assert_script_run "curl -f -v " . data_url('containers/index.html') . " > $dir/BuildTest/index.html";
 
     # At least on publiccloud, this image pull can take long and occasinally fails due to network issues
-    $builder->build($dir . "/BuildTest", "myapp", timeout => 600);
+    my $timeout = 600;
+    $timeout = 1200 unless (is_x86_64);    # Increased timeout on non-x86_64 architectures
```
b10n1k, line 71: ternary operator here will be awesome

b10n1k (changes requested):

lib/containers/container_images.pm
jlausuch, line 71:
```suggestion
    my $timeout = is_x86_64 ? 600 : 1200;
```

grisu48: Good point! I sometimes forget that perl has ternary operators.

grisu48: Handling this as duplicate of Yiannis comment. Thanks for the suggestion!

lib/containers/container_images.pm
```diff
@@ -66,7 +67,7 @@ sub build_and_run_image {
     assert_script_run "curl -f -v " . data_url('containers/index.html') . " > $dir/BuildTest/index.html";
 
     # At least on publiccloud, this image pull can take long and occasinally fails due to network issues
-    $builder->build($dir . "/BuildTest", "myapp", timeout => 600);
+    $builder->build($dir . "/BuildTest", "myapp", (timeout => is_x86_64 ? 600 : 1200));
```
cfconrad, line 70:
```suggestion
    $builder->build($dir . "/BuildTest", "myapp", timeout => is_x86_64 ? 600 : 1200);
```

cfconrad: It's not needed, but it also doesn't harm :), as the parameters of a function are always just simple array.

grisu48: I'd like to keep it that way, unless there is a clear style guide suggestion otherwise

jlausuch: I also kind of like it without (), but personal opinion too. Or if you wanna keep, at least moving the () to the actual operation, not before the variable name which looks a bit weird.. `timeout => (operation)`

b10n1k (approved):
LGTM

pdostal (approved):
Nice :rocket:

PR #13382, opened 2021-10-04
Containers: skip zypper related host tests for 15-SP4

jlausuch:
Using zypper in containers with different version as the host
doesn't work if we are registered against SCC Proxy, which
gets the assets from local storage.

To avoid this, we skip the tests on latest products that are
not released (15-SP4) for this case and we set the variable
SCC_URL pointing to the corresponding proxy url.

- Related ticket: https://progress.opensuse.org/issues/99738
- VR: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=43.1&groupid=313&test=docker_tests&test=podman_tests

### DIFF
--- a/tests/containers/buildah_docker.pm
+++ b/tests/containers/buildah_docker.pm
@@ -52,8 +52,9 @@ sub run {
                 test_opensuse_based_image(image => "${prefix_img_name}-working-container", runtime => $buildah, version => $version, beta => $beta);
                 # Due to the steps from the test_opensuse_based_image previously,
                 # the image has been committed as refreshed
-
-                build_and_run_image(runtime => $docker, builder => $buildah, base => $iname);
+                # If we are in not-released SLE host, we can't use zypper commands inside containers
+                # that are not the same version as the host, so we skip this test.
+                build_and_run_image(runtime => $docker, builder => $buildah, base => $iname) unless is_unreleased_sle;
                 $docker->cleanup_system_host(0);
                 $buildah->cleanup_system_host();
             }
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -25,9 +25,15 @@ use utils qw(zypper_call systemctl file_content_replace script_retry);
 use version_utils qw(is_sle is_leap is_microos is_sle_micro is_opensuse is_jeos is_public_cloud get_os_release check_version);
 use containers::utils qw(can_build_sle_base registry_url);
 
-our @EXPORT = qw(install_podman_when_needed install_docker_when_needed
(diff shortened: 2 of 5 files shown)


b10n1k (approved):
LGTM

jlausuch:
Related: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13384

pdostal (approved):
Can you add a comment stating what you state in description of this PR?

jlausuch:
> Can you add a comment stating what you state in description of this PR?

in all the 4 files?

jlausuch:
I can't add the same thing to both files cause pipeline complains:

```
1 duplicate(s) found with a length of 13 (>= 13 lines) and 760 bytes reduced to complete lines:
1.  File: tests/containers/podman_image.pm in lines 9..21 (bytes 253..1012)
2.  File: tests/containers/docker_image.pm in lines 10..22 (bytes 324..1083)
```
So, I will keep it only in docker flavors

tests/containers/buildah_docker.pm
```diff
@@ -52,8 +52,9 @@ sub run {
                 test_opensuse_based_image(image => "${prefix_img_name}-working-container", runtime => $buildah, version => $version, beta => $beta);
                 # Due to the steps from the test_opensuse_based_image previously,
                 # the image has been committed as refreshed
-
-                build_and_run_image(runtime => $docker, builder => $buildah, base => $iname);
+                # If "SCC_URL" is set, it means we are in not-released SLE host, which means that we can't use
+                # zypper commands inside containers that are not the same version as the host, so we skip this test.
+                build_and_run_image(runtime => $docker, builder => $buildah, base => $iname) unless get_var('SCC_URL');
```
grisu48, line 57: I'm nitpicking: What do you think about moving the check to a library function in `lib/container/common.pm` ? Something like

```
sub is_unreleased_SLE() { return get_var('SCC_URL', '') ne ''; }
```

and then use this function everywhere? Functionally it remains exactly the same, but IMHO this is better understandable, as the magic query is hidden behind a comprehensive function name.

This is a style suggestion only, so feel free to ignore :slightly_smiling_face:

grisu48 (approved):
LGTM!

tests/containers/buildah_docker.pm
jlausuch: I like it :)

jlausuch: please take a look again

jlausuch:
VR: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=43.1&groupid=313&test=docker_tests&test=podman_tests
(failures are due to https://progress.opensuse.org/issues/99732)

PR #13383, opened 2021-10-04
container: Fix check_image_in_host

cfconrad:
The value associated to `=~` must be an lvalue. And others get
interpolated. So using an array there makes no sense.

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -240,8 +240,7 @@ Returns true if host contains C<img> or false.
 =cut
 sub check_image_in_host {
     my ($self, $img) = @_;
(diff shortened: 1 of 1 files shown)


jlausuch (approved):
LGTM

jlausuch:
Will you add VR? Is there a ticket or some test failure to show why the current approach fails?

cfconrad:
unfortunately not, I just saw this code and it was obvious wrong for me. Actually I'm not sure if what jobs I should run.

jlausuch:
> unfortunately not, I just saw this code and it was obvious wrong for me. Actually I'm not sure what jobs I should run.

fair enough :)

PR #13385, opened 2021-10-04
containers::engine: Croak on timeout in _engine_script_run()

cfconrad:
The function `testapi::script_run()` it self, does not have a proper
handling if the command didn't reach the end in the given timeout. For
this case it simply return `undef`. I guess all our usecases in
container doesn't need this flexibility, so we can take care of that
error.

The if condition for `$args{timeout} > 0` is needed, as `timeout => 0`
is used to execute a command, without waiting for the prompt.

- Related ticket: https://progress.opensuse.org/issues/99765
- Verification run: 
  - https://openqa.suse.de/tests/7314786#step/rootless_podman/108 (with this change)
  - https://openqa.suse.de/tests/7311624#step/rootless_podman/116 (without this change, nothing tell me a timeout, instead that the console is still in the old command)

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -13,6 +13,7 @@
 package containers::engine;
 use Mojo::Base -base;
 use testapi;
+use Carp 'croak';
 use Test::Assert 'assert_equals';
 use containers::utils qw(registry_url);
 use utils qw(systemctl file_content_replace);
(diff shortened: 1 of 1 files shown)


jlausuch:
Why don't we just use `assert_script_run`?

cfconrad:
I guess we need the retval from the command

pdostal (approved):

grisu48 (approved):
Following Clemens's comment why `assert_script_run` would not work here I think that's a good solution for improving the container engine class.

LGTM.

jlausuch (approved):

PR #13387, opened 2021-10-05
products/microos/main: Replace repos before setting up SELinux

Vogtinator:
Setting up SELinux on the image installs packages from the repos, so better
make sure those are configured correctly at that point already.

- Verification run: https://openqa.opensuse.org/tests/1953253#

### DIFF
--- a/products/microos/main.pm
+++ b/products/microos/main.pm
@@ -37,7 +37,10 @@ sub load_boot_from_disk_tests {
     # Preparation for start testing
     loadtest 'microos/disk_boot';
     loadtest 'installation/system_workarounds' if is_aarch64;
-    loadtest 'transactional/enable_selinux'    if (get_var("ENABLE_SELINUX"));
+    replace_opensuse_repos_tests               if is_repo_replacement_required;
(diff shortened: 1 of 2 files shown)


products/microos/main.pm
```diff
@@ -37,7 +37,10 @@ sub load_boot_from_disk_tests {
     # Preparation for start testing
     loadtest 'microos/disk_boot';
     loadtest 'installation/system_workarounds' if is_aarch64;
-    loadtest 'transactional/enable_selinux'    if (get_var("ENABLE_SELINUX"));
+    replace_opensuse_repos_tests               if is_repo_replacement_required;
```
jlausuch, line 40: is this applicable for SELinux case only? If so, should we wrap this in the same condition `if (get_var("ENABLE_SELINUX"))` ?

Vogtinator: For consistency I did the change unconditionally. IMO it's weird if the order is different just in the SELinux case

jlausuch (approved):

DimStar77 (approved):

PR #13388, opened 2021-10-05
workaround: qa_automation issue with query to some dashboard

dzedro:
Disable curl <dashboard.qa2.suse.asia> until it's clarified what is with the dashboard. [1]
As far as I understand it's not required for test itself. [2]
[1] https://sd.suse.com/servicedesk/customer/portal/1/SD-62274
[2] https://github.com/SUSE/qa-testsuites/blob/master/sets/qa_testset_automation/automation/qaset/libs/msg_queue.sh

- Related ticket: https://progress.opensuse.org/issues/99591
- Verification run:
https://openqa.suse.de/tests/7319696
http://dzedro.suse.cz/tests/19518
http://dzedro.suse.cz/tests/19523

### DIFF
--- a/tests/qa_automation/kernel_multipath.pm
+++ b/tests/qa_automation/kernel_multipath.pm
@@ -142,6 +142,8 @@ sub start_testrun {
     }
 
     $self->qaset_config();
+    # workaround dashboard query https://sd.suse.com/servicedesk/customer/portal/1/SD-62274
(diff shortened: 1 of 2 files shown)


foursixnine (approved):

dzedro:
First I tried to just comment out the `curl` query but it didn't work properly, so I just deleted the file. It is more useless than I thought. :roll_eyes:

PR #13398, opened 2021-10-06
Cleanup /tmp/script* before running container commands as rootless user

jlausuch:
rootless_podman runs same commands that have been run in previous
modules as root user. Some of these commands create a script in
/tmp . When the rootless user runs those commands, it fails to
create the scripts since they are already created before by root.

- Related ticket: https://progress.opensuse.org/issues/99732


tests/containers/rootless_podman.pm
```diff
@@ -50,6 +50,11 @@ sub run {
     assert_script_run "grep $user /etc/subuid", fail_message => "subuid range not assigned for $user";
     assert_script_run "grep $user /etc/subgid", fail_message => "subgid range not assigned for $user";
     assert_script_run "setfacl -m u:$user:r /etc/zypp/credentials.d/*" if is_sle;
+
+    # Remove all previous commands generated by root. Some of these commands will be triggered
+    # by the rootless user and will generate the same file /tmp/script_X which will fail if it
```
Vogtinator, line 55:
```suggestion
    # by the rootless user and will generate the same file /tmp/scriptX which will fail if it
```

tests/containers/rootless_podman.pm
```diff
@@ -50,6 +50,11 @@ sub run {
...
+    # already exists
```
Vogtinator, line 56:
```suggestion
    # already exists owned by root
```

jlausuch:
VRs:

- https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23rootless_podman_fix&distri=sle&version=15-SP4
- https://openqa.suse.de/tests/7325186
- https://openqa.opensuse.org/t1955436

Vogtinator:
VR would be useful.

There has to be a better way, maybe making the script marker actually random or redoing the `script_output` mechanism completely without using named temporary files. That's more work and something for the future (tm) though.

mloviska:
> VR would be useful.
> 
> There has to be a better way, maybe making the script marker actually random or redoing the `script_output` mechanism completely without using named temporary files. That's more work and something for the future (tm) though.

Yup, agreed. I guess we can file a ticket in our bl and keep this workaround for a while. @jlausuch

jlausuch:
> VR would be useful.
They are running, some green already.

> 
> There has to be a better way, maybe making the script marker actually random or redoing the `script_output` mechanism completely without using named temporary files. That's more work and something for the future (tm) though.

Yes, definitely. This is a quick fix, workaround-ing the real problem that never encountered before. I'll create a poo ticket for it.

cfconrad (approved):
LGTM

PR #13400, opened 2021-10-06
wicked: Upload test logs via serial_terminal

cfconrad:
* Unify the usage of `serial_terminal::upload_file()`
* Add `wickedbase->{post_log_files}` array to add files to `post` log achive

- Verification run: http://openqa.wicked.suse.de/t14261

### DIFF
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -43,12 +43,15 @@ This function saves the command and the stdout and stderr to a file to be upload
 =cut
 sub wicked_command {
     my ($self, $action, $iface) = @_;
+    my $serial_log = '/tmp/wicked_serial.log';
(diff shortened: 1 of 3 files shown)


asmorodskyi (approved):
LGTM

cfconrad:
> Generally LGTM . one minor concern IMO what you introducing here is duplicating what we already had for ages here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/wicked/before_test.pm#L33 . Does it ?
> 
> I don't mind but I just wonder if we should drop this if no one wants to use it and inventing "new ways" ?

I don't introduce a new thing. And especially nothing related to wicked command logging. Only the upload of logs changed.

PR #13401, opened 2021-10-06
install_ltp: Use openSUSE:Factory/standard package

pevik:
x86_64 used package built against (published) Tumbleweed, which occasionally

1) required incompatible perl version (thus job failed due install error) which was fixed by removed %{perl_requires} from LTP
   package spec,

2) was built against older kernel headers, which should be mostly ok, as built LTP on slightly older headers is not a big problem s most of the errors are found in runtime. But to fix this problem properly, use package built against [openSUSE:Factory/standard](https://build.opensuse.org/package/binaries/benchmark:ltp:devel/ltp/openSUSE_Factory).

Fixes: https://progress.opensuse.org/issues/93237#note-10

Verification run: TODO

### DIFF
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -215,7 +215,7 @@ sub add_ltp_repo {
         if ((is_leap('=15.2') && is_x86_64) || (is_leap('15.3+') && !is_s390x)) {
             $repo = sprintf("Leap_%s", get_var('VERSION'));
         } elsif (is_tumbleweed) {
-            $repo = "Tumbleweed";
+            $repo = "Factory";
             $repo = "Factory_PowerPC"  if is_ppc64le();
             $repo = "Factory_zSystems" if is_s390x();
         } else {


pevik:
@Vogtinator @czerw @jlausuch @DimStar77 FYI

DimStar77:
Probably better, but skewed in the other direction.

Looking at the process we currently live with 3 snapshots in parallel:

1) The future on to be tested (Called Factory) (e.g 20211006, will start build soon)
2) The snapshot under test (build completed, so out of Factory, but not yet published, so not yet Tumbleweed) (e.g. 20211005, currently in QA)
3) the published one (called openSUSE Tumbleweed, e.g 20211001 currently published)

We can build against 1 and 3 - is the 'thing copied over to QA, waiting on a decision to be released

This in turn boils down to 'how quick the devel project would rebuild against Factory' in case of incompatible changes and if the test happened to run before or after the rebuild/publish of the devel prj. In most cases I agree, though, we will be closer to the SUT than Tumbleweed, which by nature HAS to be behind what we test)

DimStar77:
> It seems really tricky to sync all these points. We can compile LTP during test in openQA to avoid all these issues, but it is step back. Could we use something else than openSUSE:Factory/standard as target?

No, the snapshot to be tested is just laid aside waiting for a QA result, there is no repo one can build against in this case

DimStar77:
> @DimStar77 do I understand it correctly you ack this as kind of improvement (together with https://progress.opensuse.org/issues/93237 it could mostly work).

Yes - IMHO this is still better and closer to what we test. I merely wanted to point out that any 3rd-party repo can't ever be a 100% solution

PR #13404, opened 2021-10-06
Only use Dockerfile.python3 in container runtime tests

Vogtinator:
Dockerfile.python3 does not actually use the provided base image. This makes it unsuitable for base image tests.

Instead, use Dockerfile which actually uses the provided base image if possible, and don't build anything otherwise.

Verification runs:
https://openqa.opensuse.org/t1956024
(more to come)

### DIFF
--- a/tests/containers/docker.pm
+++ b/tests/containers/docker.pm
@@ -66,6 +66,9 @@ sub run {
     # Build an image from Dockerfile and test it
     build_and_run_image(runtime => $engine, base => 'registry.opensuse.org/opensuse/leap:latest');
 
+    # Build a third-party image and test it
+    build_and_run_image(runtime => $engine, dockerfile => 'Dockerfile.python3')
+
     # Clean container
     $engine->cleanup_system_host();
 }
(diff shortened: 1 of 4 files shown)


jlausuch:
One of the problems here is that originally, we had apache app build on top of the image that is tested...  and we found some problems when we started running those images on non-sle hosts because we it runs zypper command to install apache  
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/containers/Dockerfile
and if the host is not registered, the container won't get any repo (i.e. can't use zypper).
And this makes a lot of sense for SLE hosts, e.g. 
https://openqa.suse.de/tests/7324469/logfile?filename=serial_terminal.txt
```
Step 1/7 : FROM registry.suse.com/suse/sles12sp4
 ---> 32395cb5242c
Step 2/7 : RUN zypper -n in apache2 && zypper clean -a
etc
```

The approach to solve this problem was to use another Dockerfile for those cases... that's when the python3 dockerfile came into place. However, I understand now that it's completely wrong... we should use a Dockerfile where the base image is used, and not `FROM pyhton:3` which is not really testing any image... 

I am not sure this PR solves the problem cause it removes that "apache" test from SLE hosts... but I am fine as a workaround and would like to work on a REAL solution of having a Dockerfile for ANY situation which uses the CURRENT image under test.

PR #13408, opened 2021-10-07
container: Use stringify to simply access $engine->runtime member

cfconrad:
Simplify the access to $engine->runtime member variable in string
interpolation.

With this change a code like
```perl
for my $engine (containers::engine::buildah->new(), containers::engine::podman->new(), containers::engine::docker->new()) {
   print("Engine: $engine\n");
}
```
will retrive a output like:
```
Engine: buildah
Engine: podman
Engine: docker
```

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -17,6 +17,11 @@ use Carp 'croak';
 use Test::Assert 'assert_equals';
 use containers::utils qw(registry_url);
 use utils qw(systemctl file_content_replace);
+use overload
+  '""'     => sub { return shift->runtime },
+  bool     => sub { return 1 },
+  fallback => sub { return 1 };
+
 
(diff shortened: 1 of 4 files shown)


jlausuch:
`WIP` label. Shall I merge?

lib/containers/engine.pm
```diff
@@ -17,6 +17,11 @@ use Carp 'croak';
 use Test::Assert 'assert_equals';
 use containers::utils qw(registry_url);
 use utils qw(systemctl file_content_replace);
+use overload
+  '""'     => sub { return shift->runtime },
+  bool     => sub { return 1 },
+  fallback => sub { return 1 };
+
```
pdostal, line 24: Hello Clemens, how this actually works? Or do you have some good source where I can understand?

PR #13411, opened 2021-10-07
Fix broken container test app

grisu48:
The container test program `app.py` broke due to the python version switch.
This commit rewrites the app and fixes the ongoing test failures.
The new version also does not require any additional dependencies and works as stand-alone.

- Related ticket: https://progress.opensuse.org/issues/99834
- Verification runs
* [Tumbleweed docker](http://duck-norris.qam.suse.de/tests/7347) | [Tumbleweed podman](http://duck-norris.qam.suse.de/tests/7348)
* [15-SP3 docker](https://openqa.suse.de/tests/7377410)
* [15-SP3 podman](https://openqa.suse.de/tests/7377523)
* [15-SP3 podman-aarch64](https://openqa.suse.de/tests/7377524)
* [15-SP3 s390x-docker](https://openqa.suse.de/tests/7377532)
* [15-SP2 docker](https://openqa.suse.de/tests/7377538)
* [15-SP1 podman](https://openqa.suse.de/tests/7377539)
* [12-SP5 docker](https://openqa.suse.de/tests/7377540)
* [12-SP5 docker-s390x](https://openqa.suse.de/tests/7377541)
* [12-SP4 podman](https://openqa.suse.de/tests/7377542)
* [12-SP3 docker](https://openqa.suse.de/tests/7377543)


jlausuch:
> That `Dockerfile` also requires the use of a base image with `/usr/bin/python3` preinstalled, which none of our base containers have. So either it has to continue using a third-party image like `registry_url('python', '3.9');` or use `zypper` to install Python, but the only reason we have this image is to avoid having to use `zypper`...

+1
For host tests, I am fine if we use 3rd party image to test the "build" feature on the host. That can be part of docker.pm or podman.pm

For image tests, we should have a Dockerfile that doesn't use zypper or install any package (since we won't have any repo in non-registered systems). I would propose a super basic Dockerfile that only does dummy things (we just need to verifies that the given image can be built).

For example:

_script.sh_:
```
#!/bin/bash
cat /etc/os-release > /tmp/output.txt
```

_Dockerfile_:
```
FROM registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3
COPY script.sh /tmp/script.sh
RUN /tmp/script.sh
CMD cat /tmp/output.txt
```

You can complicate it as much as you want, as long as you make it isolated from the host (don't use zypper or expose ports).

jlausuch:
LGTM. Waiting for VRs

PR #13422, opened 2021-10-08
lib/utils: Fix script_run_interactive

Vogtinator:
wait_serial only returns the match, never more than that. So the return value
was missing in the returned string, which caused it to die.

Fix this by waiting for the return value up to the line ending.

- Verification run: https://openqa.opensuse.org/tests/1959708

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1641,14 +1641,15 @@ sub script_run_interactive {
         push(@words, $k->{prompt});
     }
 
-    push(@words, $endmark);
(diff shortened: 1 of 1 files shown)


mloviska:
VR is green, let's merge it.

mloviska:
> https://openqa.opensuse.org/tests/1962296
> 
> The VR was with a differentb schedule? I.e the now failing module was not in the VR

It was excluded in that run `EXCLUDE_MODULES=usr_sbin_smbd`

PR #13427, opened 2021-10-09
Run "audit-remote-libvirt" with CC System Role

lilyeyes:
Add "audit-remote-libvirt" into SLE15 SP4 job group and run with Common Criteria System Role for x86_64 arch
poo#99465 - [sle][security][sle15sp4][CC] automation: add "audit-remote-libvirt" into SLE15 SP4 job group for x86_64 arch

- Related ticket: https://progress.opensuse.org/issues/99465
- Needles: NA
- Verification run: https://openqa.suse.de/tests/7354294#

### DIFF
--- a/data/cc/nested-L2-vm.xml
+++ b/data/cc/nested-L2-vm.xml
@@ -1,0 +1,60 @@
+<domain type='kvm'>
+  <name>nested-L2-vm</name>
+  <uuid>7b256a71-1525-4b58-bdcf-e9dc779c3edd</uuid>
+  <metadata>
+  </metadata>
+  <memory unit='KiB'>1048576</memory>
+  <currentMemory unit='KiB'>1048576</currentMemory>
+  <vcpu placement='static'>1</vcpu>
+  <os>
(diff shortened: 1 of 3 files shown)


rfan1:
Lili, I would suggest you use the new qcow2 image which is built for nested vm tests [more clean status], even you can re-use the swtpm one. 

https://openqa.suse.de/tests/7262945

with this qcow2 image, you don't need to hardcode the macaddress there. Udev rules will use the default network "eth0"

lilyeyes:
> Lili, I would suggest you use the new qcow2 image which is built for nested vm tests [more clean status], even you can re-use the swtpm one.
> 
> https://openqa.suse.de/tests/7262945
> 
> with this qcow2 image, you don't need to hardcode the macaddress there. Udev rules will use the default network "eth0"

Revised code according to you comments, please help to review again.
The new verification run is passed: http://openqa.suse.de/tests/7377337

jouyingbin:
@lilyeyes 
seems CI check fail.
`This branch has conflicts that must be resolved.`

PR #13436, opened 2021-10-11
WIP Increase timeout and repeat needle match in Tomcat

Zaoliang:
We still have sporadic performance issue in Tomcat.
Increase timeout for needle match and add wait_still_screen after
send_key 'ret'. Repeat needle match to avoid performance issue. 
Add record_info  about this performance issue.
see https://progress.opensuse.org/issues/96611
verification:
http://10.162.30.85/tests/4302#step/tomcat/205


Zaoliang:
not relevant anymore.

PR #13441, opened 2021-10-12
Fix myapp image not found issue on s390x

grisu48:
This commit fixes the 'myapp image not found' error on s390x by directly
querying the image name without going the extra path of querying
repositories in a special format.

- Related ticket: https://progress.opensuse.org/issues/100826
- Verification run: https://openqa.suse.de/t7386236 https://openqa.opensuse.org/t1964266 https://openqa.opensuse.org/t1964267 https://openqa.suse.de/t7386162 https://openqa.suse.de/t7386163


grisu48:
I've been rushing this PR so that the second round of container runs has those changes already in there and we can move forward in fixing the other remaining issues.

I'm happy to discuss style issues outside of this PR and would create a new PR if required.

PR #13464, opened 2021-10-13
Add 'ncurses' into extra_tests_misc

Zaoliang:
move ncurses out of installation tests.
keep ncurses in main_common.pm for leap untill yaml schedule can be 
introduced for leap.

see https://progress.opensuse.org/issues/95365 and
https://progress.opensuse.org/issues/100868

verifications:
http://10.162.30.85/tests/3784 (sles)
http://10.162.30.85/tests/3785 (Tumbleweed)

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1157,7 +1157,7 @@ sub load_consoletests {
         loadtest "jeos/glibc_locale";
         loadtest "jeos/kiwi_templates" unless (is_leap('<15.2'));
     }
-    loadtest "console/ncurses";
+    loadtest "console/ncurses" if is_leap;
     loadtest "console/yast2_lan" unless is_bridged_networking;
     # no local certificate store
     if (!is_krypton_argon) {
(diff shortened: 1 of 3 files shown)


michaelgrifalconi:
With this comment I am not going to discuss implementation details, but only the main goal and the planning of such change that is being questioned.

**What is the goal of this change?** 
This is part of a bigger topic (I see we are missing an epic for that, would make sense to create one).
The idea is to have installation **jobs** on openQA that only create disk images, and then **tests** which are much faster to run, on created images. This is to avoid a test failing at the end of an installation job and needing several hours (instead of minutes) for subsequent runs (retry, development of fix, etc.)
Idea is to move out as much test as possible and have simple installation jobs.

This whole change would not help in case of a regression, that will need a new image to be created, but will help massively in case of hiccups of infrastructure or test issues (the main causes of test failures IMHO)

**Planning of such change**
We planned it within our squad and since it seems in our squad scope to decide what to do with tests "owned" by us (in quotes since it is still **very** unclear who owns what).
We are very happy to hear opinions and consider changes of any kind but I find very inefficient to veto a change when the investment on it is already done. Questions on the design of something shall be done before any work is done IMHO.
This shows again the need for someone on top of all squads to coordinate goals/plans/direction and maybe some sync between stakeholders **before** any work is done, to avoid the risk of wasting time on something that gets dumped at the end.

PR #13465, opened 2021-10-13
Proposal: buildah test rewrite

jlausuch:
Current `buildah_podman.pm` and `buildah_docker.pm` are more focused in image based testing instead of specific buildah command line testing. There is a loop testing all the images defined in `CONTAINER_IMAGE_VERSIONS` and basically calling `test_container_image` and `test_opensuse_based_image` methods, so basically the same test logic as in `docker_image.pm` and `podman_image.pm`, but using `buildah` command instead of `podman` or `docker`. 

What I am proposing is the following:
- reduce buildah_podman.pm and buildah_docker.pm to a single file
- focus test logic on testing the buildah CLI itself
- stop running the same tests for all the images, we are already doing that in docker_image.pm and podman_image.pm. Just use single image as reference (I picked Leap 15.3). 
- remove buildah class in engines.pm -> it's not an engine per se, it's a builder helper

This way we are testing buildah specific features instead of repeating similar things we do in other tests again and again. 
Execution example: http://fromm.arch.suse.de/tests/3514


mloviska:
> What would you suggest then?

As the image is not really the main focus of the testing, I would vote for latest TW. It should be fine as long as it is the only image present on the host.

jlausuch:
> > What would you suggest then?
> 
> As the image is not really the main focus of the testing, I would vote for latest TW. It should be fine as long as it is the only image present on the host.

Ok. I picked Leap 15:3 since it's supposed to be "more stable" than a rolling release... but in theory, everything in the public registry should be stable. But if we go TW, then I would change the python check, as the version might change in the future.

jlausuch:
VR: https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23buildah_single_module&distri=sle&version=15-SP2
(I didn't find where buildah is executed in O3...)

mloviska:
> VR: https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23buildah_single_module&distri=sle&version=15-SP2 (I didn't find where buildah is executed in O3...)

https://openqa.opensuse.org/tests/1966905

jlausuch:
> Looking good!
> 
> Main missing features are testing untested images (i.e. no `buildah_image.pm`) and building a `Dockerfile` using `buildah bud`.

If we want to do so, I wouldn't put it in this module and maybe add some extra code in `podman_image.pm` to also build the image with buildah (for untested images only).

PR #13470, opened 2021-10-14
Allow running podman tests for openSUSE using main_containers.pm

jlausuch:
By changing from `if` to `unless` we only skip the tests when the condition is fullfilled and not in any other case.

Also, change the methods names to make them more self descriptive.

VRs:
Image tests: [x86_64](https://openqa.opensuse.org/tests/1968365)  [arm](https://openqa.opensuse.org/tests/1968377)
Host tests: [x86_64](https://openqa.opensuse.org/tests/1968366)  [arm](https://openqa.opensuse.org/tests/1968376)

I would actually split the host tests into podman and docker, since there are many tests, might make sense to treat them separately. For image tests, we can keep them together...
   Something like this:
    - [docker](https://openqa.opensuse.org/tests/1968367)
    - [podman](https://openqa.opensuse.org/tests/1968368)


jlausuch:
Changes already done in [Tumbleweed](https://openqa.opensuse.org/group_overview/1) and [Tumbleweed AArch64 ](https://openqa.opensuse.org/group_overview/3)

PR #13485, opened 2021-10-18
lib/main_common: Disable plasma_browser_integration module for now

Vogtinator:
The extension got pulled by Mozilla.

- Verification run: https://openqa.opensuse.org/tests/1976937


DimStar77:
ok, that sounds like the easiest way for the time being - it can't pass on any distro anyway.

PR #13487, opened 2021-10-18
BCI: handle timeouts properly for each test call in the loop

jlausuch:
script_run returns undef if the command timesout, but it doesn't
really kill the process. The command keeps running polluting the
console and causing the next commands to fail.

By appending <timeout X> before the actual command, we make sure
the command will terminate after X seconds.

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -30,10 +30,17 @@ sub parse_logs {
     my $root = $dom->createElement('testsuites');
     $dom->setDocumentElement($root);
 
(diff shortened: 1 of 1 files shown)


jlausuch:
VRs here: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=17.8.15&groupid=398

Example of command timeout and skip logs: https://openqa.suse.de/tests/7465974#step/bci_test/43 .

dcermak:
Thanks!

PR #13494, opened 2021-10-19
assert_and_click expects timeout via keyword argument

kalikiana:
Odd number of elements in hash assignment at /usr/lib/os-autoinst/testapi.pm line 505.
    testapi::assert_and_click("firefox-enter-password-OK", 120) called at sle/tests/fips/mozilla_nss/firefox_nss.pm line 152

As seen in https://openqa.suse.de/tests/7452647/logfile?filename=autoinst-log.txt. No ticket since I noticed this by chance because the failure was being discussed in chat.

Used `rg 'assert_and_click.*,[^a-z]*\d'` to find more occurances.


tjyrinki (approved):
This is a good catch, let's see if the CI needs some kicking still.

PR #13495, opened 2021-10-19
Public Cloud: skip terraform destroy if not applied

jlausuch:
Example of the issue: https://openqa.suse.de/tests/7468470#step/upload_container_images_to_ecr/89
This is a test that doesn't use terraform at all, but it calls `terraform_destroy` after the module has finished.

VR: https://openqa.suse.de/tests/7469101

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -441,8 +441,10 @@ Destroys the current terraform deployment
 =cut
 sub terraform_destroy {
     my ($self) = @_;
+    # Do not destroy if terraform has not been applied or the environment doesn't exist
+    return unless ($self->terraform_applied);
+
     my $cmd;
-    return if get_required_var('TEST') =~ /^publiccloud_upload_img/;
     record_info('INFO', 'Removing terraform plan...');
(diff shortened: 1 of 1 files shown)


asmorodskyi:
@ricardobranco777 , @alvarocarvajald  - please review

cfconrad:
> Is there a scenario where `terraform apply` creates some resources and exits with failure? Don't we need a `terraform destroy` then?

I'm not sure here, but I think this was the reason why we did not implement it this way. So we can say that others need to take care of cleanup like https://github.com/SUSE/pcw or we need to add destroy calls on the error path as well like: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/9875b24804a8fb2f779c51f105cac2854ecb4298/lib/publiccloud/provider.pm#L405 and here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/9875b24804a8fb2f779c51f105cac2854ecb4298/lib/publiccloud/provider.pm#L401

jlausuch:
latest code VRs:
https://openqa.suse.de/tests/7469968
https://openqa.suse.de/tests/7469969

PR #13498, opened 2021-10-20
Implement cc multi-machine tests on s390x 

rfan1:
For cc multi-machine test on s390x platform, it
is different as x86_64 and aarch64, we don't need
setup "setup_multimachine" test module since it is
already in bridge network

- Related ticket: https://progress.opensuse.org/issues/99441
 https://progress.opensuse.org/issues/101157
- Needles:n/a
- Verification run: please ignore the failure on s390x, since it is due to poo#99441, we will fix it on backend worker or network layer.

390x:
https://openqa.suse.de/tests/7476984
https://openqa.suse.de/tests/7476985
x86_64:
https://openqa.suse.de/tests/7476987
https://openqa.suse.de/tests/7476988

### DIFF
--- a/schedule/security/cc_netfilter_netfilebt.yaml
+++ b/schedule/security/cc_netfilter_netfilebt.yaml
@@ -2,12 +2,24 @@ name: cc_netfilter
 description:    >
     This is for audit_test netfilter
 schedule:
+    - '{{bootloader_zkvm}}'
     - boot/boot_to_desktop
-    - network/setup_multimachine
+    - '{{setup_multimachine}}'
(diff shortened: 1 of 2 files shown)


jouyingbin:
is it a typo: backed worker?

lilyeyes:
I think you will try CC system role later when the related blocker bug get fixed, right?

rfan1:
> I think you will try CC system role later when the related blocker bug get fixed, right?

Long way to go, we have to consider the bug https://bugzilla.suse.com/show_bug.cgi?id=1190475.
But you are correct,  the target version is using CC system role

PR #13502, opened 2021-10-20
Containers: minor aesthetic improvements to EKS test

jlausuch:
- Dump some useful information using record_info for more test flow
visibility and faster troubleshooting.

- Replace assert_script_run by script_output which works better with
multi-line strings.

- Change default name for PUBLIC_CLOUD_RESOURCE_NAME to openqa.
`openqa-vm` might confuse people, as it doesn't create a VM...

VR: https://openqa.suse.de/tests/7477069

### DIFF
--- a/lib/publiccloud/aws.pm
+++ b/lib/publiccloud/aws.pm
@@ -72,7 +72,7 @@ sub init {
 }
 
 =head2 get_container_registry_prefix
-Get the full registry prefix URL for any containers image registry of ECR based on the account and region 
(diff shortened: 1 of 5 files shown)


ilausuch:
> What would you think about renaming `upload_container_images_to_ecr` to `push_image_to_ecr` ?

I like more use container_images instead of only images to not confuse with cloud or vm images. Then an other suggestion could be `push_container_image_to_ecr`

PR #13505, opened 2021-10-20
publiccloud: Prepare terrafrom env only on terraform_apply()

cfconrad:
the class `publiccloud::provider` is now used also on tests where we do
not need that terraform environment at all. But others like credential
handling and is used.
With this change, we make terraform not as one of the mandatory usages!

- Verification run : 
- https://openqa.suse.de/tests/7482707 - sle_15_SP3_image_on_eks
- https://openqa.suse.de/t7481502 - publiccloud_ltp_syscalls

### DIFF
--- a/lib/publiccloud/aws.pm
+++ b/lib/publiccloud/aws.pm
@@ -96,6 +96,6 @@ Returns a default tag for container images based of the current job id
 =cut
 sub get_default_tag {
     my ($self) = @_;
-    return join('-', get_var('PUBLIC_CLOUD_RESOURCE_NAME', 'openqa-vm'), get_current_job_id());
+    return join('-', $self->resource_name, get_current_job_id());
 }
 1;
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -24,9 +24,11 @@ has key_secret => undef;
 has region => undef;
 has username => undef;
 has prefix => 'openqa';
+has terraform_env_prepared => 0;
 has terraform_applied => 0;
 has vault_token => undef;
 has vault_lease_id => undef;
+has resource_name => sub { get_var('PUBLIC_CLOUD_RESOURCE_NAME', 'openqa-vm') };
 
 =head1 METHODS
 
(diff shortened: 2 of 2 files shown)


asmorodskyi:
the problem which I see with what we have currently is that eks/ecr are not siblings to azure/aws/gce . They almost not have any crossing points . And this PR is actually trying to solve the problem which comes from this one . If we would like to keep them in one "family" than I am ok with this PR , but I would actually consider moving eks/ecr as independent entity so it will not have provider as base class . than we will not need this PR

cfconrad:
> Looks great! One of the VRs fails: https://openqa.suse.de/tests/7481501#

Some OpenQA mistery, I used now a different job to clone https://openqa.suse.de/tests/7481511 and it worked out of the box

jlausuch:
> > > the problem which I see with what we have currently is that eks/ecr are not siblings to azure/aws/gce . They almost not have any crossing points . And this PR is actually trying to solve the problem which comes from this one . If we would like to keep them in one "family" than I am ok with this PR , but I would actually consider moving eks/ecr as independent entity so it will not have provider as base class . than we will not need this PR
> > 
> > 
> > Why different family? It uses our PC providers, it uses Vault. The only difference is that EKS test doesn't use Terraform. So, I don't really see a problem with that.
> 
> ```
> use constant TERRAFORM_DIR => '/root/terraform';
> use constant TERRAFORM_TIMEOUT => 30 * 60;
> 
> has key_id => undef;
> has key_secret => undef;
> has region => undef;
> has username => undef;
> has prefix => 'openqa';
> has terraform_applied => 0;
> has vault_token => undef;
> has vault_lease_id => undef;
> 
> sub init 
> sub conv_openqa_tf_name 
> sub find_img 
> sub upload_image
> sub img_proof 
> sub parse_img_proof_output 
> sub create_ssh_key 
> sub run_img_proof 
> sub get_image_id 
> sub create_instance 
> sub create_instances 
> sub on_terraform_apply_timeout 
> sub on_terraform_destroy_timeout 
> sub terraform_apply 
> sub terraform_destroy 
> sub terraform_param_tags
> sub escape_single_quote 
> sub __vault_login
> sub vault_login 
> sub __vault_api 
> sub vault_api 
> sub vault_get_secrets 
> sub vault_revoke 
> sub cleanup 
> sub stop_instance
> sub start_instance
> sub get_state_from_instance
> ```
> 
> this is all methods which provider has . from all this methods it uses only ONE . and it is `vault_api` just to get access to the Vault server basically and than setup auth in it's own way not related to how it is done for azure/aws/gce .
> 
> I am not sure how and where it uses PC providers . Can you point me to the code ?

What about doing Google part without using provider and see if there is more benefits than drawbacks? But if we need the vault api call, we probably need to extract it somewhere as utils method..
@ilausuch what would you think about that?

ldevulder:
> And to think that @ldevulder had these tests running and passing again last month. :(

Yes the tests were green (apart for Azure, a fix in a formula is needed and pushed in QAM). But if TF binary has been updated that could be the issue... @ricardobranco777 check this with the SHAP tean I think (for the supported version).

PR #13509, opened 2021-10-21
console/clamav: Avoid timeout failures due to upstream database timeouts

foursixnine:
- console/clamav: Avoid upstream database
- console/clamav: Attach config file when test fails

VR: https://openqa.suse.de/tests/7496514


jouyingbin (commented):
LGTM. thanks :)
Running on SLE15 SP4 FIPS job group is working well too.

PR #13511, opened 2021-10-21
Increase coverage of postfix tests

aylen384:
https://progress.opensuse.org/issues/99642


asdil12:
http://openqa.suse.de/t7497704

PR #13524, opened 2021-10-22
Design yamls with podman and docker to run as separate jobs

b10n1k:
We need to split the jobs which get scheduled with both container engines together, to run separately.
Redesign the yamls schedules which fit those conditions to be able to do so.
This follows the latest trend in the qa-c.

The need cames from another task which tries to merge podman_image and docker_image in one side and
podman__3rd_party_images and docker_3rd_party_images.

Those apparently touch pubcloud jobs and jeos and it wouldnt necessity if they were scheduled by main.pm.
make the change only for jeos for now.
This comes with https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/482

- Related ticket: https://progress.opensuse.org/issues/101210


jlausuch (commented):
As discussed, it's preferable to stick to single job for containers in JeOS.

pdostal (approved):

b10n1k:
not valid anymore

PR #13534, opened 2021-10-25
Create main_publiccloud

b10n1k:
With this PR we intent to move the scheduling from the yaml to the main.pm for the publiccloud jobs

- Related ticket: https://progress.opensuse.org/issues/101427- Verification run: 

```
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561413 YAML_SCHEDULE= PUBLIC_CLOUD_IMG_PROOF_TESTS= PUBLIC_CLOUD_CONSOLE_TESTS=1
Github oauth token provided, performing authenticated requests
Created job #7565691: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_consoletests@64bit -> https://openqa.suse.de/t7565691
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561412 YAML_SCHEDULE= PUBLIC_CLOUD_IMG_PROOF_TESTS= PUBLIC_CLOUD_CONTAINERS=1 CONTAINER_RUNTIME=1
Github oauth token provided, performing authenticated requests
Created job #7565387: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_containers@64bit -> https://openqa.suse.de/t7565387
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561410 YAML_SCHEDULE= PUBLIC_CLOUD_IMG_PROOF_TESTS= PUBLIC_CLOUD_DOWNLOAD_TESTREPO=1
Github oauth token provided, performing authenticated requests
Created job #7567468: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_download_testrepos@64bit -> https://openqa.suse.de/t7567468
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561411 YAML_SCHEDULE=
Github oauth token provided, performing authenticated requests
Created job #7565506: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_img_proof@64bit -> https://openqa.suse.de/t7565506
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561415 YAML_SCHEDULE=
Github oauth token provided, performing authenticated requests
Created job #7565570: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_ltp_cve@64bit -> https://openqa.suse.de/t7565570
❯ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534 https://openqa.suse.de/tests/7561414 YAML_SCHEDULE=
Github oauth token provided, performing authenticated requests
Created job #7565571: sle-15-SP3-EC2-ARM-Updates-x86_64-Build20211029-1-publiccloud_ltp_syscalls@64bit -> https://openqa.suse.de/t7565571```

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -66,6 +66,7 @@ our @EXPORT = qw(
   load_consoletests
   load_create_hdd_tests
   load_extra_tests
+  load_extra_tests_prepare
   load_inst_tests
   load_iso_in_external_tests
   load_jeos_tests
@@ -2804,69 +2805,6 @@ sub load_extra_tests_kernel {
     loadtest "kernel/tuned";
 }
 
-sub load_qam_publiccloud_tests {
-    my $args = OpenQA::Test::RunArgs->new();
-
-    loadtest "publiccloud/download_repos";
-    loadtest "publiccloud/prepare_instance", run_args => $args;
-    loadtest "publiccloud/register_system", run_args => $args;
-    loadtest "publiccloud/transfer_repos", run_args => $args;
-    loadtest "publiccloud/patch_and_reboot", run_args => $args;
-    if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
-        loadtest("publiccloud/img_proof", run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_LTP')) {
-        loadtest('publiccloud/run_ltp', run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME')) {
-        loadtest('publiccloud/boottime', run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
-        loadtest('publiccloud/storage_perf', run_args => $args);
-    } else {
-        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
-        loadtest "publiccloud/instance_overview" unless get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS');
-        if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
-            load_extra_tests_prepare();
-            load_extra_tests_console();
-        }
-        if (get_var('PUBLIC_CLOUD_CONTAINERS')) {
-            load_extra_tests_prepare();
-            load_extra_tests_console();
-        }
-        # Finalize
-        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
-    }
-}
-
-sub load_publiccloud_tests {
-    if (get_var('PUBLIC_CLOUD_QAM')) {
-        load_qam_publiccloud_tests();
-    } else {
-        if (get_var('PUBLIC_CLOUD_PREPARE_TOOLS')) {
-            loadtest "publiccloud/prepare_tools";
-        }
-        elsif (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
-            loadtest "publiccloud/img_proof";
-        }
-        elsif (get_var('PUBLIC_CLOUD_LTP')) {
-            loadtest 'publiccloud/run_ltp';
-        }
-        elsif (get_var('PUBLIC_CLOUD_SLES4SAP')) {
-            loadtest 'publiccloud/sles4sap';
-        }
-        elsif (get_var('PUBLIC_CLOUD_ACCNET')) {
-            loadtest 'publiccloud/az_accelerated_net';
-        }
-        elsif (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME')) {
-            loadtest "publiccloud/boottime";
-        }
-        elsif (get_var('PUBLIC_CLOUD_FIO')) {
-            loadtest 'publiccloud/storage_perf';
-        }
-        elsif (get_var('PUBLIC_CLOUD_IMAGE_LOCATION')) {
-            loadtest "publiccloud/upload_image";
-        }
-    }
-}
-
 # Scheduling set for validation of specific installation
 sub load_installation_validation_tests {
     load_system_prepare_tests;
@@ -2904,9 +2842,8 @@ sub load_transactional_role_tests {
 
 sub load_common_opensuse_sle_tests {
     load_autoyast_clone_tests if get_var("CLONE_SYSTEM");
-    load_publiccloud_tests if get_var('PUBLIC_CLOUD');
     loadtest "terraform/create_image" if get_var('TERRAFORM');
-    load_create_hdd_tests if get_var("STORE_HDD_1") || get_var("PUBLISH_HDD_1");
+    load_create_hdd_tests if (get_var("STORE_HDD_1") || get_var("PUBLISH_HDD_1")) && !get_var('PUBLIC_CLOUD');
     loadtest 'console/network_hostname' if get_var('NETWORK_CONFIGURATION');
     load_installation_validation_tests if get_var('INSTALLATION_VALIDATION');
     load_transactional_role_tests if is_transactional && (get_var('ARCH') !~ /ppc64|s390/);
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -1,0 +1,116 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: module loader of container tests
+# Maintainer: qa-c@suse.de
+
+package main_publiccloud;
+use Mojo::Base 'Exporter';
+use utils;
+use version_utils;
+use main_common qw(loadtest load_extra_tests_prepare);
+use testapi qw(check_var get_var);
+use Utils::Architectures qw(is_aarch64);
+
+our @EXPORT = qw(
+  load_publiccloud_tests
+);
+
+sub load_podman_tests() {
+    loadtest 'containers/podman';
+    loadtest 'containers/podman_3rd_party_images';
+}
+
+sub load_docker_tests() {
+    loadtest 'containers/docker';
+    loadtest 'containers/docker_runc' unless (is_aarch64 && is_sle('<=15'));
+    loadtest 'containers/docker_3rd_party_images';
+    loadtest 'containers/registry' unless (is_aarch64 && is_sle('<=15-SP1'));
+    loadtest 'containers/zypper_docker' unless (is_aarch64 && is_sle('<=15'));
+}
+
+sub load_maintenance_publiccloud_tests {
+    my $args = OpenQA::Test::RunArgs->new();
+
+    loadtest "publiccloud/download_repos";
+    loadtest "publiccloud/prepare_instance", run_args => $args;
+    loadtest "publiccloud/register_system", run_args => $args;
+    loadtest "publiccloud/transfer_repos", run_args => $args;
+    loadtest "publiccloud/patch_and_reboot", run_args => $args;
+    if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
+        loadtest("publiccloud/img_proof", run_args => $args);
+    } elsif (get_var('PUBLIC_CLOUD_LTP')) {
+        loadtest('publiccloud/run_ltp', run_args => $args);
+    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
+        loadtest('publiccloud/storage_perf', run_args => $args);
+    } else {
+        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
+        loadtest "publiccloud/instance_overview" unless get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS');
+        if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
+            load_extra_tests_prepare();
+            load_publiccloud_consoletests();
+        }
+        if (get_var('PUBLIC_CLOUD_CONTAINERS')) {
+            load_podman_tests() if is_sle('>=15-sp1');
+            load_docker_tests();
+        }
+        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    }
+}
+
+sub load_publiccloud_consoletests {
+    loadtest 'console/cleanup_qam_testrepos';
+    loadtest 'console/openvswitch';
+    loadtest 'console/rpm';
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/main_common.pm [hunk 1]
```diff
@@ -2903,9 +2841,8 @@ sub load_transactional_role_tests {
 
 sub load_common_opensuse_sle_tests {
     load_autoyast_clone_tests if get_var("CLONE_SYSTEM");
-    load_publiccloud_tests if get_var('PUBLIC_CLOUD');
     loadtest "terraform/create_image" if get_var('TERRAFORM');
-    load_create_hdd_tests if get_var("STORE_HDD_1") || get_var("PUBLISH_HDD_1");
+    load_create_hdd_tests if (get_var("STORE_HDD_1") || get_var("PUBLISH_HDD_1")) && !get_var('PUBLIC_CLOUD');
```
asmorodskyi, line 2845: Can you explain why we need this additional condition ?

asmorodskyi: no tests which has `PUBLIC_CLOUD` has `PUBLISH_HDD_1` or `STORE_HDD_1`

lib/main_publiccloud.pm
```diff
@@ -0,0 +1,118 @@
...
+        loadtest('publiccloud/storage_perf', run_args => $args);
+    } else {
+        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
+        loadtest "publiccloud/instance_overview" unless get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS');
+        if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
+            load_extra_tests_prepare();
+            load_publiccloud_consoletests();
+        }
+        if (get_var('PUBLIC_CLOUD_CONTAINERS')) {
+            load_podman_tests() if is_sle('>=15-sp1');
```
asmorodskyi, line 58: you skip `load_extra_tests_prepare`  in comparison with version which was deleted in `main_common` . was it intentional and why if it was intentional ?

lib/main_publiccloud.pm
```diff
@@ -0,0 +1,118 @@
...
+    loadtest 'containers/docker_image';
+    loadtest 'containers/docker_3rd_party_images';
+    loadtest 'containers/registry';
+    loadtest 'containers/zypper_docker';
+}
+
+sub load_qam_publiccloud_tests {
+    my $args = OpenQA::Test::RunArgs->new();
+
+    loadtest "publiccloud/download_repos";
```
asmorodskyi, line 37: If I not mistaken you already raised question that this call can be skipped and @pdostal  and @grisu48  confirmed that , right ?

lib/main_publiccloud.pm
```diff
@@ -0,0 +1,118 @@
...
+    loadtest "publiccloud/prepare_instance", run_args => $args;
+    loadtest "publiccloud/register_system", run_args => $args;
+    loadtest "publiccloud/transfer_repos", run_args => $args;
+    loadtest "publiccloud/patch_and_reboot", run_args => $args;
+    if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
+        loadtest("publiccloud/img_proof", run_args => $args);
+    } elsif (get_var('PUBLIC_CLOUD_LTP')) {
+        loadtest('publiccloud/run_ltp', run_args => $args);
+    } elsif (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME')) {
+        loadtest('publiccloud/boottime', run_args => $args);
```
asmorodskyi, line 47: AFAIK , it was considered that boottime tests does not have sense for QAM and we don't have any schedule so this can be dropped

asmorodskyi:
Overall everything looks fine , nice work @b10n1k :+1:  Some minor clarifications , see above . 

Also there is one general comment  - if this PR will be merged we will get flexibility to trigger jobs with `YAML_SCHEDULE` and `main.pm` basically by setting / removing `YAML_SCHEDULE` you can easily switch between one way or another . I personally find this flexibility dangerous and would like to avoid it . Such "dual setup" may lead to confusion in a lot of different scenarios . I would add to this PR deletion of existing YAML's .
@b10n1k , @cfconrad , @jlausuch , @ilausuch , @pdostal , @grisu48 - WDYT ?

asmorodskyi (changes requested):
IMO related YAML's needs to be dropped

lib/main_publiccloud.pm
```diff
@@ -0,0 +1,118 @@
...
+sub load_docker_tests() {
+    loadtest 'containers/docker';
+    loadtest 'containers/docker_runc';
+    loadtest 'containers/docker_image';
+    loadtest 'containers/docker_3rd_party_images';
+    loadtest 'containers/registry';
+    loadtest 'containers/zypper_docker';
+}
+
+sub load_qam_publiccloud_tests {
```
jlausuch, line 34: Is it a good time to rename `qam` by `maintenance`? I think it reflects better what it is and we eliminate the separation between the "old" departments.
`load_maintenance_publiccloud_tests`

lib/main_publiccloud.pm
```diff
@@ -0,0 +1,118 @@
...
+our @EXPORT = qw(
+  load_publiccloud_tests
+);
+
+sub load_podman_tests() {
+    loadtest 'containers/podman';
+    loadtest 'containers/podman_3rd_party_images';
+}
+
+sub load_docker_tests() {
```
jlausuch, line 25: Are we running any ARM machine in PC? If so, we should disable some of these tests for aarch64, as some packages are not available in certain versions. 
See https://confluence.suse.com/display/ENGCTNRSTORY/Level+Test+Plan#LevelTestPlan-PackagesfromContainersModule

jlausuch:
Looks good! Would like that Anton's comments are addressed before I approve.

Also, I have a suggestion for everyone to vote. Currently we have `is_public_cloud` in `version_utils.pm` [link](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/version_utils.pm#L622) but that's not really a "version" per se...  so we could move it to the new `main_publiccloud.pm` similar to `main_containers` where it contains `is_container_test`.

```
sub is_public_cloud {
    return get_var('PUBLIC_CLOUD');
}
```

WDYT?

lib/main_publiccloud.pm
pdostal: Probably not, see the [thread](https://suse.slack.com/archives/C02CGKBCGT1/p1635156292116700) where @grisu48 pointed that this module also checks that the maintenance updates are downloaded. But I'd maybe move such check to `transfer_repos.pm`. Let's create a ticket for that?

lib/main_publiccloud.pm
pdostal: Probably not, see the [thread](https://suse.slack.com/archives/C02CGKBCGT1/p1635156292116700) where @grisu48 pointed that this module also checks that the maintenance updates are downloaded. But I'd maybe move such check to `transfer_repos.pm`. Let's create a ticket for that?

pdostal:
@asmorodskyi: I vote for removing `schedule/publiccloud/` but rather in separate PR after we're sure we successfully switched.
@jlausuch: The same needs to be in my opinion done for containers as well.

lib/main_common.pm
```diff
@@ -66,6 +66,7 @@ our @EXPORT = qw(
   load_consoletests
   load_create_hdd_tests
   load_extra_tests
+  load_extra_tests_prepare
```
pdostal, line 69:
```suggestion
```

lib/main_publiccloud.pm
asmorodskyi: yes we do . for AWS we have aarch64 flavors

asmorodskyi:
>  but that's not really a "version" per se... so we could move it to the new `main_publiccloud.pm` similar to `main_containers.pn` which contains `is_container_test` 

please take a closer look to all functions in that file , majority of them are doing exactly the same : checking some variable and return true if it exists . IMHO `is_container_test` should be also moved to `version_utils`actually :) 

But I actually don't have hard feelings here we can do what Jose suggested too. I just wanted to clarify that we can not use "this function not related to  version_utils because it is not checking version " as good reason for this :)

lib/main_common.pm
asmorodskyi: I had same thought at first look , but this function is actually used in `load_qam_publiccloud_tests` which is moved out of `main_common` so we do need to export it.

lib/main_publiccloud.pm
asmorodskyi: ah right , I forgot . and I kinda agree with what Felix is saying about another place to check that we did everything right . so please ignore my comment

asmorodskyi:
> IMO related YAML's needs to be dropped

@b10n1k  please ignore this request . I think I will agree with @pdostal https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534#issuecomment-955732309 and we can do that in separate PR after some time when we will see that everything is fine . 

But I will keep "request changes " due to some other minor things which we need to do before merging ...

jlausuch:
> 
> please take a closer look to all functions in that file , majority of them are doing exactly the same : checking some variable and return true if it exists . IMHO `is_container_test` should be also moved to `version_utils`actually :)
> 

This is right, there are other functions that have nothing to do with "versions"... it looks like a messed library of mixture of different things ... so, let's not make it more messy :)

lib/main_common.pm
b10n1k: i believe the reason was that it was messing with the scheduling. need to check. i do not really remember.

lib/main_common.pm
pdostal: I see now, thank you Anton

lib/main_common.pm
b10n1k: So the job which has the var `PUBLISH_HDD_1` and breaks the scheduling is `publiccloud_download_testrepos`.

lib/main_publiccloud.pm
b10n1k: i just removed `docker_image`. This doesnt be used anywhere. the rest of the modules looks ok. VRs is against ARM btw

b10n1k: *this change is ready for review*

lib/main_publiccloud.pm
b10n1k: *this change is ready for review*

lib/main_publiccloud.pm
b10n1k: *this change is ready for review*

b10n1k:
> Looks good! Would like that Anton's comments are addressed before I approve.
> 
> Also, I have a suggestion for everyone to vote. Currently we have `is_public_cloud` in `version_utils.pm` [link](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/version_utils.pm#L622) but that's not really a "version" per se... so we could move it to the new `main_publiccloud.pm` similar to `main_containers.pn` which contains `is_container_test` [link](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/main_containers.pm#L26).
> 
> ```
> sub is_public_cloud {
>     return get_var('PUBLIC_CLOUD');
> }
> ```
> 
> WDYT?

Will it require additional modifications in various files right?

lib/main_publiccloud.pm
b10n1k: When i came to it i assumed that the containers yaml scheduling has obsoleted the function in *main_common*. *load_qam_publiccloud_tests* was not following the updates in the yaml IMO. i dont know but this is what i pressume. that said the *PUBLIC_CLOUD_CONTAINERS* var should schedule only the containers. And currently, `publiccloud_containers` jobs do not contain anything from `load_extra_tests_prepare`.

lib/main_publiccloud.pm
jlausuch: `zypper_docker` and `runc` shouldn't be run against  <=SLE 15 in aarch64.
`registry`  shouldn't be run against  <=SLE 15-SP1 in aarch64.

lib/main_common.pm
asmorodskyi: ok I see , so let's keep than

lib/main_publiccloud.pm
asmorodskyi: > this change is ready for review

can you elaborate what you have done ? idea is to remove this line and I still see it

lib/main_publiccloud.pm
asmorodskyi: yeah sure , I would use same assumptions . Just wanted to make sure that this is intentional , because you moved code to different place it is not obvious from the diff that you also dropped some code on the way

lib/main_publiccloud.pm
b10n1k: *this change is ready for review* i added conditions. thanks @jlausuch

asmorodskyi (approved):
LGTM

lib/main_publiccloud.pm
jlausuch: thanks

jlausuch (approved):

variables.md
```diff
@@ -141,6 +141,7 @@ PUBLIC_CLOUD_CLIENT_ID | string | "" | In GCE one of the variables used to authe
 PUBLIC_CLOUD_CONFIDENTIAL_VM | boolean | false | GCE Confidential VM instance
 PUBLIC_CLOUD_CONSOLE_TESTS | boolean | false | If set, console tests are added to the job.
 PUBLIC_CLOUD_CONTAINERS | boolean | false | If set, containers tests are added to the job.
+PUBLIC_CLOUD_DOWNLOAD_TESTREPO | boolean | false | If set, it schedules `publiccloud/download_repos` job.
```
pdostal, line 144: We use `PUBLIC_CLOUD_SKIP_MU` now.

PR #13550, opened 2021-10-28
[WIP] Add waiting time for ssh second xterm to show the prompt

chcao:
Add time for waiting for ssh-second-xterm to show the prompt.

- Related ticket: https://progress.opensuse.org/issues/100886
- Verification run: 
https://openqa.suse.de/tests/7547782#step/sshxterm/4
https://openqa.suse.de/tests/7547788#step/sshxterm/4

The problem is not always reproducible, running more cases to investigate it. Will update the results here.


chcao:
The problem is cased by worker performance issue, rerun or add a lager timeout will be fine, so close here.

PR #13552, opened 2021-10-28
Fix mitigation test issue for 15sp4

XGWang0:
Fix some issue and needle matching problems for 15sp4

- Related ticket: N/A
- Needles: N/A
- Verification run: http://openqa.qa2.suse.asia/tests/39469


alice-suse:
LGTM

PR #13554, opened 2021-10-28
Fix no disk space problem during the regression tests

hjluo:
We need to delete the useless snapshots in the system patch stage
and avoid creating snapshots during the 'zypper install'.

- Related ticket: https://progress.opensuse.org/issues/100958
- Needles: N/A
- Verification run:  
daily:
 https://openqa.nue.suse.com/t7563482 
 https://openqa.nue.suse.com/tests/7565791 
 https://openqa.nue.suse.com/t7563484 
 https://openqa.nue.suse.com/t7563485 
 https://openqa.nue.suse.com/t7563487 

regression:
 https://openqa.nue.suse.com/tests/7654711
 https://openqa.nue.suse.com/tests/7654712

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -89,6 +89,7 @@ our @EXPORT = qw(
   susefirewall2_to_firewalld
   permit_root_ssh
   permit_root_ssh_in_sol
+  cleanup_disk_space
 );
 
 =head1 SYNOPSIS
@@ -2042,6 +2043,35 @@ sub permit_root_ssh_in_sol {
 
     $sshd_config_file //= "/etc/ssh/sshd_config";
     enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
+}
+
+=head2 cleanup_disk_space
+    cleanup_disk_space();
+
+In fully_patch_system and minimal_patch_system, we'll create so many
+snapshots. which will cost large part of the disk space. We need to
+delete these snapshots before the migration only if the available space
+is less than DISK_LOW_WATERMARK.
+
+=cut
+
+sub cleanup_disk_space {
+    # we just do the disk clean up only if the available disk space is
+    # less than DISK_LOW_WATERMARK
+    return unless get_var("DISK_LOW_WATERMARK");
+    my $avail = script_output('findmnt -n -D -r -o avail / | awk \'{print $1+0}\'', timeout => 120);
+    diag "available space = $avail GB";
+    return if ($avail > get_var("DISK_LOW_WATERMARK"));
+
+    record_soft_failure "bsc#1192331", "Low diskspace on Filesystem root";
+
+    my @snap_lists = split /\n/, script_output("snapper list --disable-used-space | grep important= | grep -v single | awk \'{print \$1}\'");
+    foreach my $snapid (@snap_lists) {
+        assert_script_run("snapper delete -s $snapid", timeout => 120) if ($snapid > 3);
+    }
+
+    # set the snapshot number to 5-10
+    assert_script_run('snapper -croot set-config NUMBER_LIMIT=5-10');
 }
 
 1;
--- a/tests/migration/online_migration/minimal_patch.pm
+++ b/tests/migration/online_migration/minimal_patch.pm
@@ -18,6 +18,7 @@ sub run {
     select_console 'root-console';
     disable_installation_repos;
     minimal_patch_system(version_variable => 'HDDVERSION');
+    cleanup_disk_space if get_var('REMOVE_SNAPSHOTS');
     deregister_dropped_modules;
 }
 
--- a/tests/migration/online_migration/zypper_patch.pm
+++ b/tests/migration/online_migration/zypper_patch.pm
@@ -27,6 +27,7 @@ sub run {
     fully_patch_system;
     install_patterns() if (get_var('PATTERNS'));
     deregister_dropped_modules;
+    cleanup_disk_space if get_var('REMOVE_SNAPSHOTS');
     power_action('reboot', keepconsole => 1, textmode => 1);
     reconnect_mgmt_console if is_pvm;
 
--- a/tests/update/patch_sle.pm
+++ b/tests/update/patch_sle.pm
@@ -84,6 +84,9 @@ sub patching_sle {
     # create btrfs subvolume for aarch64
     create_btrfs_subvolume() if (is_aarch64);
 
+    # cleanup useless snapshots to save diskspace if we set REMOVE_SNAPSHOTS
+    cleanup_disk_space if get_var('REMOVE_SNAPSHOTS');
+
     # Remove test repos after system being patched
     remove_test_repositories;
 


lib/utils.pm
```diff
@@ -2044,4 +2045,14 @@ sub permit_root_ssh_in_sol {
     enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
 }
 
+sub cleanup_disk_space {
```
tinawang123, line 2048: I think you should add some comment to explain how to clean up disk space. It is very useful for others to use this function.

hjluo: OK.

lemon-suse (approved):

okurz (changes requested):
I general this approach seems to be fine however I suggest that you make sure there are according bugs or feature requests to prevent that problem within the product and not work around it from test side.

lib/utils.pm
```diff
@@ -2044,4 +2045,24 @@ sub permit_root_ssh_in_sol {
...
+=cut
+
+sub cleanup_disk_space {
+    my @snap_lists = split /\n/, script_output("snapper list --disable-used-space | grep important= | grep -v single | awk \'{print \$1}\'");
+    foreach my $snapid (@snap_lists) {
+        assert_script_run("snapper delete --sync $snapid") if ($snapid > 3);
+    }
+
+    # remove snapper-zypp-plugin if it was installed.
+    script_run('zypper remove -y snapper-zypp-plugin');
```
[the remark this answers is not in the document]
hjluo: > I suggest to always use `assert_script_run` unless you want to explicitly check the return value. If you want to ignore the exit code completely you could do:

changed, thanks

lib/utils.pm
```diff
@@ -2044,4 +2045,24 @@ sub permit_root_ssh_in_sol {
     enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
 }
 
+=head2 cleanup_disk_space
+    cleanup_disk_space();
+
+In fully_patch_system and minimal_patch_system, we'll create many
+snapshots. which will cost large part of the disk space. We need to 
+delete these snapshots before the migration. And we need remove 
+snapper-zypp-plugin to avoid new snapshots from 'zypper install'.
```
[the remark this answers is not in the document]
hjluo: the initiative for this is ticket is that we've hit no disk space while running regression tests.
tickets: https://progress.opensuse.org/issues/100629
            https://progress.opensuse.org/issues/97523
and I've updated the comments for this in the code snippet.

hjluo:
@okurz  Hi Oliver, can you take a look again for improvements. thanks

lib/utils.pm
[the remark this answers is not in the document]
okurz: Sure. I understand that you hit problems in tests but those progress tickets just describe the problem symptoms, not the issue with the product. We should test realistic user flows and only workaround known bugs. This is why I suggested to ensure that a corresponding bug or feature request has been reported. Please check if a corresponding bug report or feature request has been reported already or do it yourself.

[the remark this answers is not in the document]
hjluo: OK, I've searched and no bug for this kind of scenario, So I filed one bug https://bugzilla.suse.com/show_bug.cgi?id=1192331 for this.
It passed with my branch.
https://openqa.nue.suse.com/tests/7602955
https://openqa.nue.suse.com/tests/7602956

Thanks,

Huajian.Luo

lib/utils.pm
```diff
@@ -2044,4 +2045,26 @@ sub permit_root_ssh_in_sol {
...
+sub cleanup_disk_space {
+    record_soft_failure "bsc#1192331", "Low diskspace on Filesystem root";
+
+    my @snap_lists = split /\n/, script_output("snapper list --disable-used-space | grep important= | grep -v single | awk \'{print \$1}\'");
+    foreach my $snapid (@snap_lists) {
+        assert_script_run("snapper delete -s $snapid", timeout => 120) if ($snapid > 3);
+    }
+
+    # remove snapper-zypp-plugin if it was installed.
+    assert_script_run('zypper remove -y snapper-zypp-plugin ||:');
```
rfan1, line 2067: @hjluo 
I am not an expert on snapshot, however, can you please consider setting max number of the snapshots rather than removing the package?

It will disable snapshot function, correct?

hjluo: NO, To disable installation snapshots, uninstall the package snapper-zypp-plugin.

rfan1: Got it, thanks.

Anyway, removing an installed package is not a better solution here. can you try to enhance this?

rfan1 (approved):
LGTM

lib/utils.pm
```diff
@@ -2044,4 +2045,25 @@ sub permit_root_ssh_in_sol {
...
+    cleanup_disk_space();
+
+In fully_patch_system and minimal_patch_system, we'll create so many
+snapshots. which will cost large part of the disk space. We need to
+delete these snapshots before the migration.
+
+=cut
+
+sub cleanup_disk_space {
+    record_soft_failure "bsc#1192331", "Low diskspace on Filesystem root";
```
okurz, line 2058: This is now called unconditionally without checking the error condition first. You could call `script_run '[[ $(findmnt -n -D -r -o avail /) < 2G ]]'` with whatever limit you see feasible and only call `record_soft_failure` if that check returns true

tests/migration/online_migration/minimal_patch.pm
```diff
@@ -17,6 +17,7 @@ use migration;
 sub run {
     select_console 'root-console';
     disable_installation_repos;
+    cleanup_disk_space if (get_var('REMOVE_SNAPS_AND_SET_NUM'));
```
okurz, line 20:
```suggestion
    cleanup_disk_space if get_var('REMOVE_SNAPS_AND_SET_NUM');
```

okurz (changes requested):

tests/migration/online_migration/minimal_patch.pm
hjluo: changed.

lib/utils.pm
hjluo: changed. thanks

okurz (approved):

tests/migration/online_migration/zypper_patch.pm
```diff
@@ -25,8 +25,10 @@ sub run {
     disable_installation_repos;
     add_test_repositories;
     fully_patch_system;
+    cleanup_disk_space if get_var('REMOVE_SNAPS_AND_SET_NUM');
     install_patterns() if (get_var('PATTERNS'));
     deregister_dropped_modules;
+    cleanup_disk_space if get_var('REMOVE_SNAPSHOTS');
```
lemon-suse, line 31: Why not combine this line and line 28?

hjluo: changed, thanks

PR #13325, opened 2021-09-24
Save and restore fix

tbaev:
This PR fixes an issue with previous [PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13279) 

- Related ticket: NA
- Needles: NA
- Validation run : http://d488.qam.suse.de/tests/387

PR #13279, opened 2021-09-17
Extend save and restore

tbaev:
Extend the current save_and_restore test by adding a test file, after the machine state has been saved. This test file must not be present after the VM has been restored.

Related ticket: https://progress.opensuse.org/issues/97373
Needles: N/A
Verification run: http://d488.qam.suse.de/tests/113

PR #13558, opened 2021-10-29
Revert save_and_restore

grisu48:
Revert last improvements on save_and_restore as they cased a disk
corruption in some of the virtualization guests (See poo#101539).

- Related ticket: https://progress.opensuse.org/issues/101539
- Verification run: http://openqa.qam.suse.cz/tests/30010

Related PRs:

* https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13325
* https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13279

### DIFF
--- a/tests/virtualization/universal/save_and_restore.pm
+++ b/tests/virtualization/universal/save_and_restore.pm
@@ -2,15 +2,18 @@
 #
 # Copyright 2019 SUSE LLC
 # SPDX-License-Identifier: FSFAP
+#
+# Copying and distribution of this file, with or without modification,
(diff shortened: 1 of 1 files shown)


pdostal (approved):
~We need need to put this back as @tbaev did great work with this code and the only problem is that instead of `virsh save` and `virsh restore` we need to use `virsh snapshot` commands.~

After this PR all will be good as we already test internal and external snapshots elsewhere.

pdostal:
I created [poo#101776](https://progress.opensuse.org/issues/101776) for fixing the snapshotting.

PR #13566, opened 2021-10-30
Containers: create factory class

jlausuch:
This allows creating the engine object via
my $engine = $self->containers_factory('docker');
instead of
$engine = containers::engine::docker->new();

So, the test module only needs to pass 1 argument to the
constructor. This will allow us later on keep doing more
improvements.

NOTE: this is the first task after breaking down https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13370 into smaller ones.

- Related ticket: https://progress.opensuse.org/issues/101442
- VRs:
[SLE](https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_factory&distri=sle&version=15-SP3)
[Tumbleweed](https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_factory&distri=opensuse)

For now, I have only applied it to 3rd party image tests as an example and as a firs step. 

Next steps:
- [Move install_podman_when_needed and install_docker_when_needed to engine class.](https://progress.opensuse.org/issues/101444)
- [Make sure that all the container tests are using CONTAINER_RUNTIME variable](https://progress.opensuse.org/issues/101755)
- [Merge `podman_3rd_party_images.pm` and `docker_3rd_party_images.pm` into single module "3rd_party_images.pm"](https://progress.opensuse.org/issues/101213)

### DIFF
--- a/lib/containers/basetest.pm
+++ b/lib/containers/basetest.pm
@@ -1,0 +1,29 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Base class for container tests
+# Maintainer: qac team <qa-c@suse.de>
+
+package containers::basetest;
+use Mojo::Base 'opensusebasetest';
+
+sub containers_factory {
+    my ($self, $runtime) = @_;
+    my $engine;
+
+    if ($runtime eq 'docker') {
+        $engine = containers::engine::docker->new();
+    }
+    elsif ($runtime eq 'podman') {
+        $engine = containers::engine::podman->new();
+    }
+    else {
+        die("Unknown runtime $runtime. Only 'docker' and 'podman' are allowed.");
+    }
+
+    return $engine;
+}
+
+1;
--- a/tests/containers/podman_3rd_party_images.pm
+++ b/tests/containers/podman_3rd_party_images.pm
@@ -8,9 +8,7 @@
 #          Log the test results in docker-3rd_party_images_log.txt
 # Maintainer: qa-c team <qa-c@suse.de>
 
-use base 'consoletest';
-use strict;
-use warnings;
+use Mojo::Base 'containers::basetest';
 use testapi;
 use utils;
 use version_utils;
@@ -25,11 +23,11 @@ sub run {
     $self->select_serial_terminal;
 
     my ($running_version, $sp, $host_distri) = get_os_release;
-    my $engine = containers::engine::podman->new();
 
     script_run("echo 'Container base image tests:' > /var/tmp/podman-3rd_party_images_log.txt");
     # In SLE we need to add the Containers module
     install_podman_when_needed($host_distri);
+    my $engine = $self->containers_factory('podman');
     $engine->configure_insecure_registries();
     my $images = get_3rd_party_images();
     for my $image (@{$images}) {
--- a/tests/containers/docker_3rd_party_images.pm
+++ b/tests/containers/docker_3rd_party_images.pm
@@ -8,9 +8,7 @@
 #          Log the test results in docker-3rd_party_images_log.txt
 # Maintainer: qa-c team <qa-c@suse.de>
 
-use base 'consoletest';
-use strict;
-use warnings;
+use Mojo::Base 'containers::basetest';
 use testapi;
 use utils;
 use version_utils;
@@ -25,11 +23,11 @@ sub run {
     $self->select_serial_terminal;
 
     my ($running_version, $sp, $host_distri) = get_os_release;
-    my $engine = containers::engine::docker->new();
 
     script_run("echo 'Container base image tests:' > /var/tmp/docker-3rd_party_images_log.txt");
     # In SLE we need to add the Containers module
     install_docker_when_needed($host_distri);
+    my $engine = $self->containers_factory('docker');
     $engine->configure_insecure_registries();
     my $images = get_3rd_party_images();
     for my $image (@{$images}) {


lib/containers/basetest.pm
```diff
@@ -0,0 +1,43 @@
+# SUSE's openQA tests
+#
+# Copyright 2018 SUSE LLC
```
asmorodskyi, line 3:
```suggestion
# Copyright 2021 SUSE LLC
```

asmorodskyi (approved):
nice !

lib/containers/basetest.pm
```diff
@@ -0,0 +1,43 @@
...
+use strict;
+use warnings;
+use containers::common qw(install_docker_when_needed install_podman_when_needed);
+use version_utils 'get_os_release';
+
+sub containers_factory {
+    my ($self, $runtime) = @_;
+    my $engine;
+
+    die("Egnine already initialized") if ($self->{engine});
```
b10n1k, line 21: the long-plan is to be able to run a single module with different engine even when it is scheduled twice. this doesnt scale  in this case.

lib/containers/basetest.pm
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Base class for container tests
+# Maintainer: qac team <qa-c@suse.de>
+
+package containers::basetest;
+use base 'opensusebasetest';
+use testapi;
+use strict;
+use warnings;
```
b10n1k, line 13:
```suggestion
use Mojo::Base 'opensusebasetest';
use testapi;
```

lib/containers/basetest.pm
```diff
@@ -0,0 +1,43 @@
...
+
+    my ($running_version, $sp, $host_distri) = get_os_release;
+
+    if ($runtime eq 'docker') {
+        $engine = containers::engine::docker->new();
+        install_docker_when_needed($host_distri);
+    }
+    elsif ($runtime eq 'podman') {
+        $engine = containers::engine::podman->new();
+        install_podman_when_needed($host_distri);
```
b10n1k, line 31: not completely convince about that. i sort like the idea but i think it is not that simple. you mix the design with other logic. But the design should stay as clean and simply as possible to take the maximum benefits. i am afraid to propose something as it is not one or two that can be the alternatives. 
In brief keep in mind: the design pattern should focus and solve the class relationships, when the inserted subroutine focus on the configuration of the system, so i would say it is another level. flexibility is nice but i believe we need another way to show that.

lib/containers/basetest.pm
jlausuch: It does, and I think it's better to handle that in the test module, not in the factory class.

lib/containers/basetest.pm
jlausuch: I can leave this in the test module as it was, and we can think about this later on. 
My idea is also to move this into engine.pm and do something like `$engine->install()` or `$engine->setup()`.

jlausuch:
> it is not clear to me where we are heading with this. L21 is likely breaking jobs with both podman_image and docker_image. Thats why it will be nice if we could also VR from jeos, pubcloud

I will run those VRs, but it doesn't break as I'm not touching `podman_image.pm` and `docker_image.pm` in this PR, I started only the 3rd party modules.

lib/containers/basetest.pm
pdostal: See [poo#101444](https://progress.opensuse.org/issues/101444) for that

lib/containers/basetest.pm
b10n1k: i am not sure if i follow. Even if the `$self->{engine}` is defined the invocation should return itself, right? why to die?

jlausuch:
VRs with JeOS running podman and docker tests:
https://openqa.opensuse.org/tests/2006995
https://openqa.opensuse.org/tests/2006996

lib/containers/basetest.pm
jlausuch: I will change this later or tomorrow.

lib/containers/basetest.pm
jlausuch: see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/basetest.pm#L26

b10n1k:
> > it is not clear to me where we are heading with this. L21 is likely breaking jobs with both podman_image and docker_image. Thats why it will be nice if we could also VR from jeos, pubcloud
> 
> I will run those VRs, but it doesn't break as I'm not touching `podman_image.pm` and `docker_image.pm` in this PR, I started only the 3rd party modules.

https://openqa.suse.de/tests/7579646

jlausuch:
As request from @b10n1k, I have reduced the class to the bare minimum, which is the purpose of the PR. We can discus if we move the installation of the packages and registry configuration in the Factory class later on. Will run now more VRs with latest code.

jlausuch:
[SLE](https://openqa.suse.de/tests/overview?version=15-SP3&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_factory&distri=sle)
[openSUSE](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_factory)

pdostal (approved):

lib/containers/basetest.pm
```diff
@@ -0,0 +1,34 @@
...
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Base class for container tests
+# Maintainer: qac team <qa-c@suse.de>
+
+package containers::basetest;
+use Mojo::Base 'opensusebasetest';
+use containers::common qw(install_docker_when_needed install_podman_when_needed);
```
b10n1k, line 11: not used

jlausuch: fixed

b10n1k (changes requested):
i forgot to submit the review. Check comments. L11 also need to be removed, right?

lib/containers/basetest.pm
```diff
@@ -0,0 +1,31 @@
...
+# Summary: Base class for container tests
+# Maintainer: qac team <qa-c@suse.de>
+
+package containers::basetest;
+use Mojo::Base 'opensusebasetest';
+use version_utils 'get_os_release';
+
+sub containers_factory {
+    my ($self, $runtime) = @_;
+    my $engine;
```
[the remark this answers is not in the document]
jlausuch: It's only used in this single method, so let's keep it here.

lib/containers/basetest.pm
```diff
@@ -0,0 +1,33 @@
...
+        $engine = containers::engine::docker->new();
+    }
+    elsif ($runtime eq 'podman') {
+        $engine = containers::engine::podman->new();
+    }
+    else {
+        die("Unknown runtime $runtime. Only 'docker' and 'podman' are allowed.");
+    }
+
+    $self->{engine} = $engine;
```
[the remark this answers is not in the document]
jlausuch: done

tests/containers/podman_3rd_party_images.pm
```diff
@@ -8,7 +8,7 @@
 #          Log the test results in docker-3rd_party_images_log.txt
 # Maintainer: qa-c team <qa-c@suse.de>
 
-use base 'consoletest';
+use Mojo::Base 'containers::basetest';
 use strict;
 use warnings;
```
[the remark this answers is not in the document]
jlausuch: done

lib/containers/basetest.pm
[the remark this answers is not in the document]
b10n1k: i will resolve it but maybe @cfconrad should have another look. because as the class is the base `$engine` should be defined for the whole instance. Then can be called in other functions directly. 
Anyway it is fine for now i guess

b10n1k (approved):
it should be fine for now

lib/containers/basetest.pm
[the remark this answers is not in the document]
cfconrad: I would keep it as @jlausuch did, as we do not use it in other methods inside of `container::basetest`.
If we will get a common `cleanup()` or other stuff, then I would go also with a member variable.

grisu48 (commented):
A bit late for the party, still LGTM

jlausuch:
> A bit late for the party, still LGTM

You still have the follow-up https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13585
 :)

PR #13570, opened 2021-11-01
Modify test suite security_tpm2 in compliance with Factory First

ShawnHa0:
Add yaml schedule for security_tpm2. Add version specification due to differnent file location of tpm_server in Tumbleweed, change from 'systemctl start tpm2-abrmd' to 'systemctl restart tpm2-abrmd' since start service will fail on Tumbleweed while restart works fine. Add installation of expect in tpm2_env_setup, the package is not installed in Tumbleweed by default but mandatory in case tpm2_engine_self_sign

- Related ticket: https://progress.opensuse.org/issues/101590
- Needles: N/A
- Verification run: 
  - O3: https://openqa.opensuse.org/tests/2006249#
  - OSD: https://openqa.suse.de/tests/7580897

### DIFF
--- a/tests/security/tpm2/tpm2_env_setup.pm
+++ b/tests/security/tpm2/tpm2_env_setup.pm
@@ -23,7 +23,8 @@ use warnings;
 use base 'opensusebasetest';
 use testapi;
 use utils 'zypper_call';
-use power_action_utils "power_action";
+use power_action_utils 'power_action';
+use version_utils 'is_sle';
 
 sub run {
     my $self = shift;
@@ -37,7 +38,7 @@ sub run {
 
     # Install the tpm2.0 related packages
     # and then start the TPM2 Access Broker & Resource Manager
-    zypper_call "in ibmswtpm2 tpm2.0-abrmd tpm2.0-abrmd-devel openssl tpm2-0-tss tpm2-tss-engine tpm2.0-tools";
+    zypper_call "in expect ibmswtpm2 tpm2.0-abrmd tpm2.0-abrmd-devel openssl tpm2-0-tss tpm2-tss-engine tpm2.0-tools";
 
     # As we use TPM emulator, we should do some modification for tpm2-abrmd service
     # and make it connect to "--tcti=libtss2-tcti-mssim.so"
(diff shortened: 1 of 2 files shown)


lilyeyes (approved):
LGTM

tests/security/tpm2/tpm2_env_setup.pm
```diff
@@ -62,10 +63,14 @@ EOF
...
 
     # Start the emulator
-    assert_script_run "su - tss -c '/usr/lib/ibmtss/tpm_server&'";
+    if (is_sle) {
+        assert_script_run "su - tss -c '/usr/lib/ibmtss/tpm_server&'";
+    } else {
+        assert_script_run "su - tss -c '/usr/libexec/ibmtss/tpm_server&'";
+    }
 
     # Start the tpm2-abrmd service
```
rfan1, line 72: Start -> Restart

tests/security/tpm2/tpm2_env_setup.pm
```diff
@@ -24,6 +24,7 @@ use base 'opensusebasetest';
 use testapi;
 use utils 'zypper_call';
 use power_action_utils "power_action";
```
jouyingbin, line 26: Looks like we need to use the same sign (" ") or (' ') for calling a api function. 
We could need to make them as the same style in one pm file.

ShawnHa0: Got it! will do

tests/security/tpm2/tpm2_env_setup.pm
ShawnHa0: Forgot to modify the comments as well lol, will do

tests/security/tpm2/tpm2_env_setup.pm
```diff
@@ -62,10 +63,14 @@ EOF
     $self->select_serial_terminal;
 
     # Start the emulator
-    assert_script_run "su - tss -c '/usr/lib/ibmtss/tpm_server&'";
+    if (is_sle) {
+        assert_script_run "su - tss -c '/usr/lib/ibmtss/tpm_server&'";
+    } else {
+        assert_script_run "su - tss -c '/usr/libexec/ibmtss/tpm_server&'";
+    }
```
pevik, line 71: nit: I'd define server path as variable:

```
my $server = '/usr/libexec/ibmtss/tpm_server';
$server = '/usr/lib/ibmtss/tpm_server' if (is_sle);
assert_script_run "su - tss -c '$server &'";
```
Or even detect it automatically:
```
my $server = script_output('f=/usr/libexec/ibmtss/tpm_server; [ ! $f ] && f=/usr/lib/ibmtss/tpm_server; echo $f');
die 'missing tpm_server path\n' if ($server eq '');
assert_script_run "su - tss -c '$server &'";
```

ShawnHa0: Thanks for the suggestion!

pevik (approved):

tests/security/tpm2/tpm2_env_setup.pm
pevik: or even more generic (it'd work on any distro changes in the future):

```
my $server = script_output('ls /usr/*/ibmtss/tpm_server |tail -1');
die 'missing tpm_server path\n' if ($server eq '');
assert_script_run "su - tss -c '$server &'";
```

jouyingbin (commented):
LGTM. :) thanks for the update.

PR #13573, opened 2021-11-01
Change guest installation to sequential

grisu48:
Due to ongoing network problems we need to switch the guest installation from parallel to sequential

- Related ticket: https://progress.opensuse.org/issues/101542
- Verification run: http://openqa.qam.suse.cz/tests/30124

### DIFF
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -192,10 +192,11 @@ sub create_guest {
         $virtinstall .= " --events on_reboot=$on_reboot" unless ($on_reboot eq '');
         $virtinstall .= " --extra-args '$extra_args'" unless ($extra_args eq '');
         record_info("$name", "Creating $name guests:\n$virtinstall");
-        script_run "$virtinstall >> ~/virt-install_$name.txt 2>&1 & true";    # true required because & terminator is not allowed
-
-        # wait for initrd to ensure the installation is starting
-        script_retry("grep -B99 -A99 'initrd' ~/virt-install_$name.txt", delay => 15, retry => 12, die => 0);
+        # HOTFIX: Run installation sequentially due to poo#101256
+        assert_script_run("$virtinstall >> ~/virt-install_$name.txt 2>&1", timeout => 2700);
+        ## wait for initrd to ensure the installation is starting
+        # Note: This is not needed for sequential installation but left here, for when we will work on parallel installation again.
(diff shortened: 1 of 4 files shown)


lib/virt_autotest/utils.pm
```diff
@@ -192,10 +192,11 @@ sub create_guest {
         $virtinstall .= " --events on_reboot=$on_reboot" unless ($on_reboot eq '');
         $virtinstall .= " --extra-args '$extra_args'" unless ($extra_args eq '');
         record_info("$name", "Creating $name guests:\n$virtinstall");
-        script_run "$virtinstall >> ~/virt-install_$name.txt 2>&1 & true";    # true required because & terminator is not allowed
-
-        # wait for initrd to ensure the installation is starting
-        script_retry("grep -B99 -A99 'initrd' ~/virt-install_$name.txt", delay => 15, retry => 12, die => 0);
+        # HOTFIX: Run installation sequentially and retry it due to poo#101256
+        script_retry("$virtinstall >> ~/virt-install_$name.txt 2>&1", retry => 3, timeout => 2700);
```
pdostal, line 196: Will the `script_retry` work? You can't have two VMs with the same name / MAC address.

pdostal (approved):

lib/virt_autotest/utils.pm
grisu48: I think a failed `virt-install` will delete the old virtual machine, so there is no remnant.
My doubts are actually that `virt-install` will never fail, so the script_retry will never retry.

Don't know, I'd like to try it out this way and re-evaluate after a couple of months.

pdostal: I think that:
a) The `virt-install` can never fail altho the installation stucked
b) If the `virt-install` fails it will not remove the virtual machine

pdostal: But please don't count this comment as a blocker.

grisu48: Let's turn then the approach around. I replace it with `assert_script_run` and if we see that it fails, then we can replace it with `script_retry`.

grisu48:
Restarted the VR after doing some changes following Pavels comment

grisu48:
Relevant part passed: http://openqa.qam.suse.cz/tests/30124
In the older VR all tests passed.

PR #13579, opened 2021-11-02
Add check for bsc1192051

grisu48:
Add a check for bsc1192051 in `rootless_podman` and skip the test run if present. 
This bug breaks the s390x and ppc64le test runs for `rootless_podman`.

- Related ticket: https://progress.opensuse.org/issues/101280
- Verification run: [15-SP3 aarch64](https://openqa.suse.de/t7590883) | [15-SP3 s390x](https://openqa.suse.de/t7590884) | [15-SP3 x86_64](https://openqa.suse.de/t7590885) | [15-SP4 ppc64le](https://openqa.suse.de/t7590886) | [Tumbleweed](https://openqa.opensuse.org/t2009567)

### DIFF
--- a/tests/containers/rootless_podman.pm
+++ b/tests/containers/rootless_podman.pm
@@ -25,14 +25,14 @@ use containers::utils 'registry_url';
 use version_utils qw(get_os_release);
 use version_utils 'is_sle';
 use containers::engine;
-use Utils::Architectures 'is_s390x';
+use Utils::Architectures;
 
 sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
-    if (is_s390x) {
-        record_soft_failure("poo#101154 - Test images broken for s390x");
+    if ((is_s390x || is_ppc64le) && check_bsc1192051()) {
+        record_soft_failure("bsc#1192051 - Permission denied for faccessat2");
         return;
     }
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
Can you limit this softfail to the affected arches only? Otherwise failures on currently working archs won't get noticed.

pdostal (approved):
Looks good, properly commented! :100:

grisu48:
> Can you limit this softfail to the affected arches only? Otherwise failures on currently working archs won't get noticed.

I could, however [aarch64](https://openqa.suse.de/tests/7589740) and [x86_64](https://openqa.suse.de/tests/7589742) are not affected and run the whole test. And they would probably fail in the same way, if they would be affected, right?

grisu48:
I've updated it anyways and limited the check to s390x and ppc64le. New round of VRs is scheduled

Vogtinator:
> I could, however aarch64 and x86_64 (and Tumbleweed) are not affected and run the whole test. And they would probably fail in the same way, if they would be affected, right?

Yes, but a) they shouldn't b) if they fail, it needs to be investigated, instead of just solftfailing.

Vogtinator (approved):

jlausuch (approved):
I agree with faking this bug in the code because is the current approach we need to follow for maintenance ...although I am strongly against this way of making the bot happy...

PR #13580, opened 2021-11-02
console/x_vt: Check for gnome-session as well as X

Vogtinator:
On 15.2+ as well as TW, check for gnome-session running on tty2 as well.

Previously it didn't care about GNOME running on the wrong tty because of the `tty7|wayland` condition, but now without `Xwayland` it'll detect `gnome-session` running on `tty2` as well.

I didn't add code to detect `gnome-session` on `tty7` because there's nothing to test it with so far.

Also restrict the "soft softfail" to GNOME only and change deprecated `egrep` to `grep -E`.

- Related ticket: https://progress.opensuse.org/issues/99582
- Verification run: http://10.160.67.86/tests/1099

### DIFF
--- a/tests/console/x_vt.pm
+++ b/tests/console/x_vt.pm
@@ -15,13 +15,15 @@ use version_utils qw(is_tumbleweed is_leap is_sle);
 
 sub run {
     # First, list all X processes, including user process and gdm process
-    script_run('ps -ef | grep bin/X');
-    if (script_run("ps -ef | grep bin/X | egrep 'tty7|wayland'") == 1) {
-        if ((is_tumbleweed || is_leap('>=15.2') || is_sle('>=15-SP2')) && script_run('ps -ef | grep bin/X | grep tty2') == 0) {
-            diag('user X process runs on tty2 for systems with GNOME 3.32+, see boo#1138327');
+    script_run("ps -ef | grep -E 'bin/X|/gnome-session'");
+    if (script_run("ps -ef | grep bin/X | grep -E 'tty7|wayland'") == 1) {
+        if (check_var('DESKTOP', 'gnome')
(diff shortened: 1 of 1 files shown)


Vogtinator:
VR passed

tests/console/x_vt.pm
```diff
@@ -15,13 +15,15 @@ use version_utils qw(is_tumbleweed is_leap is_sle);
...
+    script_run("ps -ef | grep -E 'bin/X|/gnome-session'");
+    if (script_run("ps -ef | grep bin/X | grep -E 'tty7|wayland'") == 1) {
+        if (check_var('DESKTOP', 'gnome')
+            && (is_tumbleweed || is_leap('>=15.2') || is_sle('>=15-SP2'))
+            && script_run("ps -ef | grep -E 'bin/X|/gnome-session' | grep tty2") == 0) {
+            diag('user session runs on tty2 for systems with GNOME 3.32+, see boo#1138327');
         }
         else {
-            die 'Expected tty7 used by X or wayland not found';
+            die 'Graphical session not found on tty7';
```
dzedro, line 26: What if there is no tty2 on GNOME 3.32+ ? :)

DimStar77 (approved):

tests/console/x_vt.pm
DimStar77: if on GNOME 3.32+ you don't get a session on tty2, then our tests do something weird (could be as simple as a text login on tty2 before the graphical login)

gdm spawns on tty7, and the user session spawns on the first free tty (tty1 reserved in SUSE cases) -> tty2. If we don't end up there, something is wrong

dzedro (approved):
Wanted to wait for SLE VRs, but I think they will pass
http://dzedro.suse.cz/tests/19851 15 SP3 tty2
http://dzedro.suse.cz/tests/19852 15 SP1 tty7

PR #13589, opened 2021-11-03
Fix xdmcp_xdm and xdmcp_gdm failure on SLE15SP4

GraceWang571:
Use an available display number for Xephyr connection

Background of why the previous DISPLAY number :1 isn't available now could be found from:
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1680
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/945

- Related ticket: https://progress.opensuse.org/issues/97901
- Needles: already added to o.s.d when debugging
- Verification run: 
- desktop-desktopapps-remote-client2 https://openqa.suse.de/tests/7594929
- desktop-desktopapps-remote-client3 https://openqa.suse.de/tests/7595401

### DIFF
--- a/tests/x11/remote_desktop/xdmcp_gdm.pm
+++ b/tests/x11/remote_desktop/xdmcp_gdm.pm
@@ -31,7 +31,7 @@ sub run {
     enter_cmd "exit";
 
     # Remote access SLES via Xephyr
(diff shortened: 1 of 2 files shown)


zhaocongjia:
LGTM. As we discussed in team meeting, this is not a bug, so adjusting test case is expected here.

PR #13590, opened 2021-11-03
WIP: Fix for yast2_firewall change in zone config

rakoenig:
- According to poo#10137 the standard zone for eth0 switches
  from 'defaut' to 'public'.
- used a minimal invasive patch that tries out interface verification
  and if it does not find matching needles it tries again with 'public'
  as a zone name.
- Zone verification will match with 'no_interfaces' when previously
  the match for 'public' found that we have the new standard.
- One new needle is created because background of the trusted zone
  also changed from white to dark.
  Needle is: yast2_firewall_trusted_no_interfaces_default-20211028

- Related ticket: https://progress.opensuse.org/issues/101370
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/752
- Needles: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1577
- Verification run: (waiting for needles to merge)

### DIFF
--- a/tests/yast2_gui/yast2_firewall.pm
+++ b/tests/yast2_gui/yast2_firewall.pm
@@ -99,9 +99,23 @@ sub verify_service_started {
 
 sub verify_interface {
     my (%args) = @_;
+    my $zone = $args{zone};
 
     assert_and_click 'yast2_firewall_interfaces_menu';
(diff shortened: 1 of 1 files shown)


jknphy:
Let's close it for now, you can still save the code in your branch and we can reopen later if needed.

PR #13592, opened 2021-11-03
autoyast: Make GRUB matching more reliable

Vogtinator:
Some autoyast xml files didn't define a timeout, use 30s as a compromise
between the default value of 10s and waiting indefinitely (-1).

Now that the GRUB menu can be reliably matched by autoyast/installation,
drop the match for opensuse-welcome, which was most likely a workaround/hack
anyway (it didn't handle generic-desktop either).

Example of a failure: https://openqa.opensuse.org/tests/2010257

Verification runs:
TW autoyast_gnome: http://10.160.67.86/tests/1106
TW autoyast_minimal: http://10.160.67.86/tests/1105

### DIFF
--- a/data/autoyast_opensuse/opensuse_autoyast_local_tftp.xml
+++ b/data/autoyast_opensuse/opensuse_autoyast_local_tftp.xml
@@ -7,6 +7,11 @@
       <confirm config:type="boolean">false</confirm>
     </mode>
   </general>
+  <bootloader>
+    <global>
+      <timeout config:type="integer">30</timeout>
+    </global>
+  </bootloader>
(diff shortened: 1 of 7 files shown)


jknphy (approved):
LGTM, I would just suggest to verify against some scenario with -1 just in case, i.e.: https://openqa.opensuse.org/tests/2010334/

Vogtinator:
> LGTM, I would just suggest to verify against some scenario with -1 just in case, i.e.: https://openqa.opensuse.org/tests/2010334/

Also passes: http://10.160.67.86/tests/1111

PR #13594, opened 2021-11-03
Containers: Increase zypper timeout and fix script_run logic

pdostal:
- Errors: [docker container exec zypper in](https://openqa.suse.de/tests/7531053#), [docker container commit](https://openqa.suse.de/tests/7587608#step/docker/206)
- Related ticket: [poo#101484](https://progress.opensuse.org/issues/101484)
- Verification run: [1](http://pdostal-server.suse.cz/tests/12646)

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -155,7 +155,9 @@ sub basic_container_tests {
     die("error: missing container $container_name") unless ($output_containers =~ m/$container_name/);
 
     # Containers' state can be saved to a docker image
-    if (script_run("$runtime container exec $container_name zypper -n in curl", 300)) {
+    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
+    die('zypper inside container timed out') if (!defined($ret));
+    if ($ret != 0) {
         record_info('poo#40958 - curl install failure, try with force-resolution.');
         my $output = script_output("$runtime container exec $container_name zypper in --force-resolution -y -n curl", 600);
         die('error: curl not installed in the container') unless (($output =~ m/Installing: curl.*done/) || ($output =~ m/\'curl\' .* already installed/));


lib/containers/utils.pm
```diff
@@ -155,12 +155,14 @@ sub basic_container_tests {
     die("error: missing container $container_name") unless ($output_containers =~ m/$container_name/);
 
     # Containers' state can be saved to a docker image
-    if (script_run("$runtime container exec $container_name zypper -n in curl", 300)) {
+    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 200);
```
jlausuch, line 158: why decreasing the timeout here?

jlausuch: Should we use script_retry here? Similar to  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13542

pdostal: Sorry, I increased it to 420 now.

I don't think we need `script_retry()`.

b10n1k (approved):
LGTM

lib/containers/utils.pm
```diff
@@ -155,7 +155,9 @@ sub basic_container_tests {
     die("error: missing container $container_name") unless ($output_containers =~ m/$container_name/);
 
     # Containers' state can be saved to a docker image
-    if (script_run("$runtime container exec $container_name zypper -n in curl", 300)) {
+    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 420);
```
dzedro, line 158: I would not be so humble with timeout increase.
Sometimes it's better to have longer timeout, it has time to return reason of the error.
But you know better, great If 420 is good enough. :+1:

pdostal: You're right, I did `600`.

jlausuch (approved):

PR #13599, opened 2021-11-04
Allow undefined settings to be used in test scheduler

Julie-CAO:
An undefined setting or a setting with value=0 are unable to be passed to yaml scheduler to filter test modules currently. Take a virtualization test as an example:

the snipet in test.yaml:
`  install_guest:
    SKIP_GUEST_INSTALL:
      **_undef_**:
        - virt_autotest/unified_guest_installation`

We expect only when SKIP_GUEST_INSTALL is set in the test suite,  virt_autotest/unified_guest_installation will be skipped. However, it is impossible in current logic even with SKIP_GUEST_INSTALL=0.

With this PR, 'undef' is allowed to be used in conditional_schedule to support the setting which is not set or value is 0 in test suites.

Only verified in my own test. Need review and verification from wider users as it may affect all the tests which use yaml scheduler.

### DIFF
--- a/lib/scheduler.pm
+++ b/lib/scheduler.pm
@@ -55,7 +55,7 @@ sub parse_schedule_module {
         my $condition = $schedule->{conditional_schedule}->{$module};
         # Iterate over variables in the condition
         foreach my $var (keys %{$condition}) {
-            next unless my $val = get_var($var);
+            my $val = get_var($var, 'undef');
             # If value of the variable matched the conditions
             # Iterate over the list of the modules to be loaded
             push(@scheduled, parse_schedule_module($schedule, $_)) for (@{$condition->{$var}->{$val}});


lib/scheduler.pm
```diff
@@ -55,7 +55,7 @@ sub parse_schedule_module {
         my $condition = $schedule->{conditional_schedule}->{$module};
         # Iterate over variables in the condition
         foreach my $var (keys %{$condition}) {
-            next unless my $val = get_var($var);
+            my $val = get_var($var, 'undef');
```
foursixnine, line 58: In this case `$val` would end up  being the actual string `undef` and If I read the code properly, that undef could end up being part of the schedule... or am I reading it wrong?.

@Julie-CAO can you provide couple of examples (the cli call for clone job)?

Julie-CAO: Yes, 'undef' will be used as a string in test.yaml as the example in the pr description. Because I can not find a better solution to translate 
`loadtest "virt_autotest/unified_guest_installation" unless get_var('SKIP_GUEST_INSTALL');"`
into a yaml file.

In our test cases, we only set SKIP_GUEST_INSTALL=1 in a few of tests, and it can be switched on or off for debugging purpose. but curret yaml filter logic dose not support it.

I am trying to creat a yaml for my test by translating the test modules loading logic from main.pm, then I found the problem in yaml scheduler. I have not submit the test yaml into git repo yet, so there is not a cli call example. I have it run in my own openqa server with some debugging logs: http://10.67.129.51/tests/2692. In this case, virt_autotest/unified_guest_installation is loaded as expected without SKIP_GUEST_INSTALL setted anywhere.

But after I think over the issue, I am clined not to change our test to use yaml scheduler as yaml scheduler does not bring more convenice than using if/else statements main.pm.

foursixnine: @Julie-CAO You can get the same result if you use words: http://phobos.qa.suse.de/tests/3857611#details

```
name:           extra_tests_textmode
description:    >
    Maintainer: dheidler.
    Extra CLI tests
conditional_schedule:
    install_guest:
      SKIP_GUEST_INSTALL:
        'NO':
          - virt_autotest/unified_guest_installation
schedule:
    - '{{install_guest}}'
    - boot/boot_to_desktop
```
However, if you want to use 0 explicitly, you need to check rather if `$val` contains an undef value. 

```suggestion
            my $val = get_var($var);
            next if (!defined $val);
```

doing this, you would be able to:

- Have it empty & not scheduled: [openqa-clone-job $job  _EXIT_AFTER_SCHEDULE=1 YAML_SCHEDULE=schedule/functional/yaml_test_schedule.yaml SKIP_GUEST_INSTALL=](phobos.qa.suse.de/tests/3857613)
- Have it with your value `0` & scheduled: [openqa-clone-job $job  _EXIT_AFTER_SCHEDULE=1 YAML_SCHEDULE=schedule/functional/yaml_test_schedule.yaml SKIP_GUEST_INSTALL=0](http://phobos.qa.suse.de/tests/3857612)

Both examples are using the yaml provided above (I just changed the 0 for a no):

```
name:           extra_tests_textmode
description:    >
    Maintainer: dheidler.
    Extra CLI tests
conditional_schedule:
    install_guest:
      SKIP_GUEST_INSTALL:
        0:
          - virt_autotest/unified_guest_installation
schedule:
    - '{{install_guest}}'
    - boot/boot_to_desktop
```

Julie-CAO: Hi @foursixnine , thanks for the detail explaination. If I set SKIP_GUEST_INSTALL='NO' as you mentioned above, and make it in my test.yaml. Will get_var() still return 'false' from  in our perl test script? because there are several places where SKIP_GUEST_INSTALL is used to control our test behaviour except in the test.yaml.

such as this statement in one of our test script:
```
    if (get_var("SKIP_GUEST_INSTALL") && is_x86_64) {
        $pre_test_cmd .= " -k $vm_xml_dir";
    }
```

foursixnine: @Julie-CAO no, for that to work, you would need to add the change I suggested above (skip if $val is undef), which Is the way to go I believe, so that you can stay with perl true/false values (0/1). Ping me and let's have a quick call about this

Julie-CAO: Hi @foursixnine , I get your meaning.  The scheduler can be almost as efficient as before with your change above. Setting value '0' explicitly make my tests work as well. It is acceptable. Thank you.

foursixnine (approved):

PR #13600, opened 2021-11-04
Add console tests to latest pubcloud testing branch

asmorodskyi:
VR : https://openqa.suse.de/tests/7604525


pdostal (approved):

asmorodskyi:
VR is actually failing with `Console 'root-console' activated in TUNNEL mode activated but tunnel(s) are not yet initialized, use the 'tunnel' console and call 'setup_ssh_tunnels'` but main purpose of this PR is to fix scheduling which is achieved and this failure will requires other PR to fix it

jlausuch (approved):

PR #13605, opened 2021-11-04
Containers: create init method to call runtime installation and config

jlausuch:
By doing this, we avoid calling the installation and registry configuration methods in the test modules repeatedly.

- Related ticket: https://progress.opensuse.org/issues/101893
- VRs: [SLE](https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_init_function&version=15-SP3&distri=sle) [TW](https://openqa.opensuse.org/tests/overview?distri=opensuse&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_init_function&version=Tumbleweed)

### DIFF
--- a/lib/containers/docker.pm
+++ b/lib/containers/docker.pm
@@ -10,8 +10,16 @@ package containers::docker;
 use Mojo::Base 'containers::engine';
 use testapi;
 use containers::utils qw(registry_url);
+use containers::common qw(install_docker_when_needed);
 use utils qw(systemctl file_content_replace);
+use version_utils qw(get_os_release);
 has runtime => 'docker';
+
+sub init {
+    my ($running_version, $sp, $host_distri) = get_os_release;
+    install_docker_when_needed($host_distri);
+    configure_insecure_registries();
+}
 
 sub configure_insecure_registries {
     my ($self) = shift;
--- a/lib/containers/podman.pm
+++ b/lib/containers/podman.pm
@@ -10,8 +10,16 @@ package containers::podman;
 use Mojo::Base 'containers::engine';
 use testapi;
 use containers::utils qw(registry_url);
+use containers::common qw(install_podman_when_needed);
 use utils qw(file_content_replace);
+use version_utils qw(get_os_release);
 has runtime => "podman";
+
+sub init {
+    my ($running_version, $sp, $host_distri) = get_os_release;
+    install_podman_when_needed($host_distri);
+    configure_insecure_registries();
+}
 
 sub configure_insecure_registries {
     my ($self) = shift;
--- a/tests/containers/docker_compose.pm
+++ b/tests/containers/docker_compose.pm
@@ -21,7 +21,7 @@ use Mojo::Base 'containers::basetest';
 use testapi;
 use registration;
 use utils;
-use version_utils qw(is_sle get_os_release);
+use version_utils qw(is_sle);
 use containers::common;
 use publiccloud::utils 'is_ondemand';
 
@@ -29,9 +29,7 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
     my $docker = $self->containers_factory('docker');
-    my ($running_version, $sp, $host_distri) = get_os_release;
 
-    install_docker_when_needed($host_distri);
     add_suseconnect_product(get_addon_fullname('phub')) if is_sle();
     add_suseconnect_product(get_addon_fullname('python2')) if is_sle('=15-sp1');
 
@@ -53,7 +51,6 @@ sub run {
     assert_script_run("curl -O " . data_url("containers/docker-compose.yml"));
     assert_script_run("curl -O " . data_url("containers/haproxy.cfg"));
 
-    $docker->configure_insecure_registries();
     file_content_replace("docker-compose.yml", REGISTRY => get_var('REGISTRY', 'docker.io'));
     assert_script_run 'docker-compose pull', 600;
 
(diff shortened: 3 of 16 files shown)


b10n1k (commented):
just small things:
- tests/containers/zypper_docker.pm has out to date copyrights
- i am a bit troubled with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13605/files#diff-3b34101ea0df2e75ea46097ce9ac2b57880f89129ee82472f20b3e689d6e4132R59. i am not sure why do we do `install_docker_when_needed` there. i will check later. just mentioned in case it slipped out from you.
- like last moment thought, what would you think to name the `lib/containers/[docker|podman].pm` to ``lib/containers/[dockerbase|podmanbase].pm`?

Nonetheless, i  still dont think `$engine->init()` should be in the *basetest.pm*. (reasons explained in other PR and poo). to make another point: in case we decide to improve in the future you will end up moving part around. a single example is if you have the test create a $engine when it starts (job scope) rather in every module (module scope).

jlausuch:
> just small things:
> 
>     * tests/containers/zypper_docker.pm has out to date copyrights

Ok. Will update.

> 
>     * i am a bit troubled with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13605/files#diff-3b34101ea0df2e75ea46097ce9ac2b57880f89129ee82472f20b3e689d6e4132R59. i am not sure why do we do `install_docker_when_needed` there. i will check later. just mentioned in case it slipped out from you.

I actually removed it the first time i saw it, but apparently the command before  `zypper_call("rm $runc");` un-installs docker and the cleanup fails because docker package doesn't exist...  I think we need to revisit that module. I didn't want to do it here.

> 
>     * like last moment thought, what would you think to name the `lib/containers/[docker|podman].pm` to ``lib/containers/[dockerbase|podmanbase].pm`?

Doesn't sound bad.. let's what others think.

> 
> 
> Nonetheless, i still dont think `$engine->init()` should be in the _basetest.pm_. (reasons explained in other PR and poo). to make another point: in case we decide to improve in the future you will end up moving part around. a single example is if you have the test create a $engine when it starts (job scope) rather in every module (module scope).

We dont have much examples of needed to install-reinstall the packages and re-configure the registries.. we are just doing it every time in the modules. This PR proves that. We save a lot of code duplication by having this init method called from factory. The approach is the same in public cloud.

jlausuch:
I lost all the changes... wait for me to fix this.

lib/containers/podman.pm
```diff
@@ -10,9 +10,17 @@ package containers::podman;
 use Mojo::Base 'containers::engine';
 use testapi;
 use containers::utils qw(registry_url);
+use containers::common qw(install_podman_when_needed);
 use utils qw(file_content_replace);
+use version_utils qw(is_sle get_os_release);
```
b10n1k, line 15: `is_sle` no needed apparently

lib/containers/docker.pm
```diff
@@ -10,9 +10,17 @@ package containers::docker;
 use Mojo::Base 'containers::engine';
 use testapi;
 use containers::utils qw(registry_url);
+use containers::common qw(install_docker_when_needed);
 use utils qw(systemctl file_content_replace);
+use version_utils qw(is_sle get_os_release);
```
b10n1k, line 15: `is_sle` is not used, right?

asmorodskyi:
generally LGTM , just one small concern see another inline comment for details ...

tests/containers/docker_compose.pm
```diff
@@ -53,7 +51,6 @@ sub run {
     assert_script_run("curl -O " . data_url("containers/docker-compose.yml"));
     assert_script_run("curl -O " . data_url("containers/haproxy.cfg"));
 
-    $docker->configure_insecure_registries();
```
asmorodskyi, line 56: in this certain case it become more severe because we changing order of execution . are we sure that it won't influence on what we actually testing ?

b10n1k: it doesnt seem to make any difference. i would say it is completely save IMO

pdostal (approved):
I think that it's awesome! :)

tests/containers/docker_compose.pm
jlausuch: I also gave it a thought when I removed it. I don't know what it was placed here instead of the beginning, I guess because of convenience, you add it when you needed it. This shouldn't affect the flow of the test.

jlausuch:
VRs:
https://openqa.suse.de/tests/overview?version=15-SP3&distri=sle&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_init_function
https://openqa.opensuse.org/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_init_function&version=Tumbleweed&distri=opensuse

lib/containers/docker.pm
jlausuch: done

lib/containers/podman.pm
jlausuch: done

jlausuch:
What do you think about Yannis proposal for `dockerbase.pm` and `podmanbase.pm`? I would keep it as it is but I don't have anything against it either :)

pdostal:
> What do you think about Yannis proposal for `dockerbase.pm` and `podmanbase.pm`? I would keep it as it is but I don't have anything against it either :)

I think that it's fine like that. I won't over engineer it.

asmorodskyi (approved):
LGTM

ggardet:
Could this PR be the cause of https://progress.opensuse.org/issues/102344 ?

PR #13608, opened 2021-11-04
console/sshd: Improve background calls for port forwarding

Vogtinator:
The ssh calls in the background spam the serial terminal with debug messages,
which confuses serial input functions, leading to wait_serial not matching
and wasting a minute of waiting for each mismatch.

Write the debug output into log files and display it at the end instead.
Also make use of background_script_run.

- Verification run: http://10.160.67.86/tests/1112

Took 29s instead of 1min50s or more (https://openqa.opensuse.org/tests/2010513#step/sshd/1)

### DIFF
--- a/tests/console/sshd.pm
+++ b/tests/console/sshd.pm
@@ -116,8 +116,8 @@ sub run {
 
     # Port forwarding (bsc#1131709 bsc#1133386)
     assert_script_run "echo 'sshd.pm: Testing port forwarding' | logger";
-    assert_script_run "( ssh -vNL 4242:localhost:22 $ssh_testman\@localhost & )";
-    assert_script_run "( ssh -vNR 0.0.0.0:5252:localhost:22 $ssh_testman\@localhost & )";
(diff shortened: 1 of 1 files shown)


pdostal:
What if one of the following commands fail?

Vogtinator:
> What if one of the following commands fail?

Then the post fail hook runs, same as before?

pdostal:
> > What if one of the following commands fail?
> 
> Then the post fail hook runs, same as before?

You switched from `assert_script_run()` to `background_script_run()` and I don't see openQA to be waiting for the results of those commands. Can you please help me to understand it?

Vogtinator:
> > > What if one of the following commands fail?
> > 
> > 
> > Then the post fail hook runs, same as before?
> 
> You switched from `assert_script_run()` to `background_script_run()` and I don't see openQA to be waiting for the results of those commands. Can you please help me to understand it Fabian?

It did not do that before either, the `assert_` part was a noop.

If the commands fail, then the forwarding is not set up and one of the later calls will fail.

pdostal (approved):
Thank you @Vogtinator. I  didn't see the `& )` part in the original code.

PR #13611, opened 2021-11-04
Remove deprecated testscases for pam

punkioudi:
As pam_tally2 and pam_cracklib modules are deprecated for >=1.5.0 pam releases, the corresponding tests fail in the testsuite, so they have to be removed

- Related ticket: https://progress.opensuse.org/issues/99039
- Needles: N/A
- Verification run: [Tumbleweed](https://openqa.opensuse.org/tests/2016700#step/pam/41)
- [sles15sp3](https://openqa.suse.de/tests/7613725#step/pam/52) | [sles15sp2
](https://openqa.suse.de/tests/7613728#step/pam/52) | [sles15sp1](https://openqa.suse.de/tests/7613787#step/pam/52) | [sles15](https://openqa.suse.de/tests/7613791#step/pam/52) | [sles12sp5](https://openqa.suse.de/tests/7613792#step/pam/52) | [sles12sp4](https://openqa.suse.de/tests/7613793#step/pam/52) | [sles12sp3](https://openqa.suse.de/tests/7613794#step/pam/52) | [sles12sp2](https://openqa.suse.de/tests/7613795#step/pam/52)

### DIFF
--- a/data/pam_test/pam_deprecated.sh
+++ b/data/pam_test/pam_deprecated.sh
@@ -53,7 +53,19 @@ teardown() {
     [ "$status" -ne 0 ]
 }
 
-@test "check for valid login shell" {                       # case 04
+@test "deny access if too many attempts fail" {               # case 04
+    sed -i '/^auth/i\auth required pam_tally2.so onerr=fail deny=1 unlock_time=10' /etc/pam.d/pam_test
+
+    run bash -c "echo -en '$USER_ERR_PW' | pam_test auth $USER_NOR" # error password triggers this case
+    [ "$status" -ne 0 ]
+    sleep 2 # this time is less than 10s
+    run bash -c "echo -en '$USER_NOR_PW' | pam_test auth $USER_NOR"
+    [ "$status" -ne 0 ]
+    sleep 12
+    echo -en "$USER_NOR_PW" | pam_test auth $USER_NOR
+}
+
+@test "check for valid login shell" {                       # case 05
     sed -i '/^account/i\account required pam_shells.so' /etc/pam.d/pam_test
 
     bash -c "echo -en '$USER_NOR_PW' | pam_test auth $USER_NOR"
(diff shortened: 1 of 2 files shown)


tests/console/pam.pm
```diff
@@ -36,9 +36,6 @@ sub run {
         zypper_ar("$qa_head_repo", name => 'qa-head-repo');
     }
     zypper_call('install bats pam-test pam pam-config snapper perl');
-    if (is_tumbleweed()) {
-        zypper_call('install pam-deprecated "group(wheel)"');
```
DimStar77, line 40: group(wheel) should probably remain - as we still have the test:

```
@test "modify group access" {                                         # case 17
    echo "*;*;$USER_NOR;Al0000-2400;wheel" >> /etc/security/group.conf
```

DimStar77:
Looks almost identical to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13307 :)

tests/console/pam.pm
punkioudi: > Looks almost identical to #13307 :)

Oh indeed! Hm, what I see in that in sle, the tests don't fail yet, as  the pam version is older than the one that these 2 modules are being deprecated (eg for sle15sp3, I see `pam-1.3.0-6.50.1.x86_64` ). So I presume that we will see the same issues there and consequently, the logic of skipping will make no sense anymore.

punkioudi:
> Looks almost identical to #13307 :)

Oh indeed! Hm, what I see in that in sle, the tests don't fail yet, as  the pam version is older than the one that these 2 modules are being deprecated (eg for sle15sp3, I see `pam-1.3.0-6.50.1.x86_64` ). So I presume that we will see the same issues there and consequently, the logic of skipping will make no sense anymore.

a-kpappas:
It looks good to me but I think someone from security should also take a look before it is merged. @lilyeyes could you also take a look? 

I also think that qe-security should be added to the maintainer field.

lilyeyes:
Thanks for @ us and good to know this test case. Security QE do not use this "pam-test" for testing atm.
The changes LGTM. 
Btw, @rfan1 from our team takes responsibility for pam testing he would like to take a look too.

lilyeyes (approved):
LGTM.

data/pam_test/pam_deprecated.sh
```diff
@@ -0,0 +1,252 @@
...
+    run bash -c "echo -en '$USER_NOR_PW' | pam_test auth $USER_ERR"
+    [ "$status" -ne 0 ]
+}
+
+@test "deny access if too many attempts fail" {               # case 04
+    sed -i '/^auth/i\auth required pam_tally2.so onerr=fail deny=1 unlock_time=10' /etc/pam.d/pam_test
+
+    run bash -c "echo -en '$USER_ERR_PW' | pam_test auth $USER_NOR" # error password triggers this case
+    [ "$status" -ne 0 ]
+    sleep 2 # this time is less that 10s
```
rfan1, line 61: might be a typo, should be less than?

punkioudi:
Hi @rfan1 , thanks for taking the time to review it. The purpose though of this PR is to remove two testcases of the two pam modules that have become deprecated in the newer versions of pam. So, tbh I cannot answer to your comments, as I haven't taken part in writing this testsuite :)

What I would suggest though, is that ,for now, to give priority on fixing the failure happening in TW and at the same time, the `qe-security` squad can decide on the refactoring of this test in a separate ticket.

rfan1:
Thanks @punkioudi, that makes sense!

Can you please help check my 1st comment? it might be a typo. then I can give approve. :)

punkioudi:
> Thanks @punkioudi, that makes sense!
> 
> Can you please help check my 1st comment? it might be a typo. then I can give approve. :)

Done :)

rfan1 (approved):
LGTM

PR #13612, opened 2021-11-04
Create named subroutines for test loader

b10n1k:
That was one out of two things i left out of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13534

This is to make the scheduler readable as the else block in the `load_publiccloud_tests` was not self-explanatory.
i include also loader in replacement of `create_hdd_autoyast_pc.yaml` from the `PC Tools Image`.

- Related ticket: https://progress.opensuse.org/issues/101427
- Verification run: 
[released](https://openqa.suse.de/t7605743)
[load_publiccloud_tools](https://openqa.suse.de/t7605744)
[load_publiccloud_download_repos](https://openqa.suse.de/t7605746)
[staging](https://openqa.suse.de/t7605745)
latest - some might fail likely due to missing variables:
[released](https://openqa.suse.de/tests/7637079)
[load_publiccloud_tools](https://openqa.suse.de/tests/7637078)
[load_publiccloud_download_repos](https://openqa.suse.de/tests/7637077)
[staging](https://openqa.suse.de/tests/7637076)

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -77,56 +77,89 @@ sub load_publiccloud_consoletests {
     loadtest 'console/libgcrypt';
 }
 
-sub load_publiccloud_tests {
-    loadtest 'boot/boot_to_desktop';
-    if (check_var('PUBLIC_CLOUD_DOWNLOAD_TESTREPO', 1)) {
-        loadtest 'publiccloud/download_repos';
-        loadtest 'shutdown/shutdown';
+sub load_latest_publiccloud_tests {
+    my $args = OpenQA::Test::RunArgs->new();
+    if (get_var('PUBLIC_CLOUD_PREPARE_TOOLS')) {
+        loadtest "publiccloud/prepare_tools";
     }
-    elsif (get_var('PUBLIC_CLOUD_QAM')) {
-        load_maintenance_publiccloud_tests();
-    } else {
-        if (get_var('PUBLIC_CLOUD_PREPARE_TOOLS')) {
-            loadtest "publiccloud/prepare_tools";
+    elsif (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
+        loadtest "publiccloud/img_proof", run_args => $args;
+    }
+    elsif (get_var('PUBLIC_CLOUD_LTP')) {
+        loadtest 'publiccloud/run_ltp', run_args => $args;
+    }
+    elsif (get_var('PUBLIC_CLOUD_SLES4SAP')) {
+        loadtest 'publiccloud/sles4sap';
+    }
+    elsif (get_var('PUBLIC_CLOUD_ACCNET')) {
+        loadtest 'publiccloud/az_accelerated_net';
+    }
+    elsif (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME')) {
+        loadtest "publiccloud/boottime", run_args => $args;
+    }
+    elsif (get_var('PUBLIC_CLOUD_FIO')) {
+        loadtest 'publiccloud/storage_perf', run_args => $args;
+    }
+    elsif (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
+        loadtest "publiccloud/prepare_instance", run_args => $args;
+        loadtest "publiccloud/register_system", run_args => $args;
+        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
+        load_extra_tests_prepare();
+        load_publiccloud_consoletests();
+        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    }
+    elsif (get_var('PUBLIC_CLOUD_CONTAINERS')) {
+        loadtest "publiccloud/prepare_instance", run_args => $args;
+        loadtest "publiccloud/register_system", run_args => $args;
+        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
+        load_podman_tests();
+        load_docker_tests();
+        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    }
+    elsif (get_var('PUBLIC_CLOUD_CREATE_TOOLS_IMG')) {
+        loadtest "publiccloud/upload_image";
+    }
+    else {
+        die "*publiccloud - Latest* expects PUBLIC_CLOUD_* job variable. None is matched from the expected ones.";
+    }
+}
+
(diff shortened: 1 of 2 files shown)


asmorodskyi (changes requested):
I have bit different idea how we can organize this file . Let me come up with different PR so we can decide together which way to choose

b10n1k:
> I have bit different idea how we can organize this file . Let me come up with different PR so we can decide together which way to choose

in practice, i havent finished. just try to do small steps. and this is became a obstacle to me once again to finish something. but you are free to submit anything in any case.

asmorodskyi:
so I tried to play with the code and it didn't work out . So I am not holding this PR anymore globally but still have some suggestions so I will keep "request changes " see other inline comments

lib/main_publiccloud.pm
```diff
@@ -77,47 +77,68 @@ sub load_publiccloud_consoletests {
     loadtest 'console/libgcrypt';
 }
 
+sub load_released_publiccloud_tests {
```
asmorodskyi, line 80: we not testing anything "released" so this name sounds confusing to me .  I open to hear other ideas . Only what I have in mind is `load_latest_publiccloud_tests` . I agree it is also not very descriptive but in any case I strongly disagree with "released"

lib/main_publiccloud.pm
```diff
@@ -77,47 +77,68 @@ sub load_publiccloud_consoletests {
...
+    elsif (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
+        my $args = OpenQA::Test::RunArgs->new();
+        loadtest "publiccloud/prepare_instance", run_args => $args;
+        loadtest "publiccloud/register_system", run_args => $args;
+        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
+        load_extra_tests_prepare();
+        load_publiccloud_consoletests();
+        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    }
+    elsif (get_var('PUBLIC_CLOUD_IMAGE_LOCATION')) {
```
asmorodskyi, line 111: in a fact **every** pubcloud test has this variable , so basically it was already the case - when something is broken we got plenty of `upload_img` tests runs which silently passing because nothing to upload . so we actually "hiding" problems . this is why I want to `upload_img` to start only if someone explicitly asked for this , https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/505 .

lib/main_publiccloud.pm
grisu48: I agree with Anton, `latest` might be the better term

lib/main_publiccloud.pm
```diff
@@ -77,47 +77,68 @@ sub load_publiccloud_consoletests {
...
+        loadtest "publiccloud/upload_image";
+    }
```
grisu48, line 113: Perhaps we want to add an else branch that dies, so that we catch misconfigurations early?

```suggestion
    } else {
        die "Missing publiccloud test run definition";
    }
```

lib/main_publiccloud.pm
b10n1k: *this change is ready for review*

lib/main_publiccloud.pm
```diff
@@ -77,47 +77,68 @@ sub load_publiccloud_consoletests {
...
-        elsif (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
-            my $args = OpenQA::Test::RunArgs->new();
-            loadtest "publiccloud/prepare_instance", run_args => $args;
-            loadtest "publiccloud/register_system", run_args => $args;
-            loadtest "publiccloud/ssh_interactive_start", run_args => $args;
-            load_extra_tests_prepare();
-            load_publiccloud_consoletests();
-            loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    else {
+        loadtest 'boot/boot_to_desktop';
```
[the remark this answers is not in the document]
asmorodskyi: no because `load_create_publiccloud_tools_image` has `loadtest 'installation/bootloader'`

lib/main_publiccloud.pm
b10n1k: *this change is ready for review*

lib/main_publiccloud.pm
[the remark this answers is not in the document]
b10n1k: https://openqa.suse.de/group_overview/276 may give you an idea

asmorodskyi (approved):
LGTM

jlausuch:
Not gonna block the PR by with this change there is plenty of code duplication. I would prefer having a function to load only specific maintenance modules (download repos, install updates, etc), and use a common function for the generic test modules.

lib/main_publiccloud.pm
pdostal: THis looks good to me but let's wait for Felix.

pdostal (approved):
This looks way more readable!

lib/main_publiccloud.pm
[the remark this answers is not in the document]
grisu48: Ahhhhh, got it, makes sense.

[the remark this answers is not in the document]
grisu48: Please add a comment to the if branch, that makes this clear. Otherwise we will have the same discussion in a week from now.

Something like

    if (check_var('PUBLIC_CLOUD_TOOLS_REPO', 1)) {
        # 'installation/bootloader' is loaded instead of 'boot_to_desktop' within this subroutine.

Or something similar

grisu48 (approved):
LGTM! I'd still add the requested comment, the rest is OK

lib/main_publiccloud.pm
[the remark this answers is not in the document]
b10n1k: The comment wasnt make sense IMO, because it is obvious why the `load_create_publiccloud_tools_image` is not using `boot_to_desktop` and why then, it is defined in the *else*. However i thought that it would be a good idea to provide a small description. So *This is ready for review* from my side

cfconrad (commented):
LGTM,

did we already thought about something like `PUBLIC_CLOUD_EXTRA_TESTS` ?

b10n1k:
> LGTM,
> 
> did we already thought about something like `PUBLIC_CLOUD_EXTRA_TESTS` ?

if you mean what i think you mean, the answer is yes. and there were attempts from other contributors too. i didnt provide this because i would have to fight about the *big* changes. So i want to do it in the next step to easy the process

PR #13620, opened 2021-11-05
Fix send key issue for hyperv 2012r2 UEFI guest

nanzhg:
Hyper-V 2012UEFI guest install failure on OSD, this patch is for fixing the issue which return key was not received by console during first boot after install completed.

- Verification run: http://openqa.nue.suse.com/tests/7611188

### DIFF
--- a/lib/grub_utils.pm
+++ b/lib/grub_utils.pm
@@ -48,7 +48,10 @@ sub grub_test {
         send_key 'ret';
     }
     # Avoid return key not received occasionally for hyperv-uefi guest at first boot
-    send_key 'ret' if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
+    if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI')) {
+        sleep 5;
+        send_key 'ret';
+    }
 }
 
 =head2 handle_installer_medium_bootup


nanzhg:
@alice-suse @guoxuguang @Julie-CAO @waynechen55 please help to review this PR. Thanks

guoxuguang:
LGTM. btw, we'd better use `script_retry()` instead of `sleep`, IMHO.

waynechen55:
> LGTM. btw, we'd better use `script_retry()` instead of `sleep`, IMHO.

Something like this would be better I think. Of course, other loop can be used instead.

```
    while ($_retry_counter gt 0) {
            wait_screen_change {
                send_key 'ret';
            };
            save_screenshot;
            $_retry_counter--;
    }
```

lib/grub_utils.pm
```diff
@@ -48,7 +48,10 @@ sub grub_test {
         send_key 'ret';
     }
     # Avoid return key not received occasionally for hyperv-uefi guest at first boot
-    send_key 'ret' if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
+    if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI')) {
+        sleep 5;
```
okurz, line 52: You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

waynechen55:
I think more verification run can be done to prove that this fix can help solve "unstable issue for those 3rd-party tests"

Julie-CAO:
It is a product bug that was found by automation test. Your pr may make the test pass but it will cover up the problem.

If the puspose of your pr is to get around the bug, please add a comment to indicate that it is just a workaround, and it should be removed just after the bug is fixed.

nanzhg:
Thanks a lot for all the comments! Agree Wayne's suggestion, besides I would like to hold on this PR to see if the issue is still happened in the beta builds. Because it got passed in latest run https://openqa.nue.suse.com/tests/7619507.

nanzhg:
> It is a product bug that was found by automation test. Your pr may make the test pass but it will cover up the problem.
> 
> If the puspose of your pr is to get around the bug, please add a comment to indicate that it is just a workaround, and it should be removed just after the bug is fixed.

Thanks Julie. This is a separate issue which different from bug 1190158.

nanzhg:
Since this issue is not happened again in recent new builds, I will close this PR.

PR #13624, opened 2021-11-05
Set mouse_hide in init_desktop_runner to 1

dzedro:
Not sure why it was 200, but looks like the offset thing was broken.

- Related ticket: https://progress.opensuse.org/issues/101941
- Verification run:
https://openqa.suse.de/tests/7613203 15 SP2
https://openqa.suse.de/tests/7613204 15 SP3

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -176,7 +176,7 @@ sub init_desktop_runner {
 
     send_key($hotkey);
 
-    mouse_hide(200);
+    mouse_hide(1);
     if (!check_screen('desktop-runner', $timeout)) {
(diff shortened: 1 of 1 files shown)


okurz:
There was https://github.com/os-autoinst/os-autoinst/pull/1840 which is most likely related

dzedro:
> There was [os-autoinst/os-autoinst#1840](https://github.com/os-autoinst/os-autoinst/pull/1840) which is most likely related

Yes, that fixed it, but left the before broken `mouse_hide` in tests untouched.

dzedro:
This is the only mouse_hide with value bigger than 1, which is affected by the fix.

PR #13625, opened 2021-11-08
Don't operate firewall when system role is Common Criteria

Amrysliu:
When system role is Common Criteria, the firewall is not installed in
the system by default. Then it will report `Unit file firewalld.service
does not exist` when disabling it in setup_multimachine.pm.

- Related ticket:  https://progress.opensuse.org/issues/102116
- Needles: N/A
- Verification run:  
            https://openqa.suse.de/tests/7642614#

### DIFF
--- a/lib/Utils/Systemd.pm
+++ b/lib/Utils/Systemd.pm
@@ -15,6 +15,7 @@ our @EXPORT = qw(
   get_started_systemd_services
   clear_started_systemd_services
   systemctl
+  check_unit_file
 );
 
 =head1 Utils::Systemd
@@ -84,6 +85,18 @@ Clear the list of started systemd services
 =cut
 sub clear_started_systemd_services {
     %started_systemd_services = ();
+}
+
+=head2 check_unit_file
+
+Check if the unit file exist
+
+=cut
+
+sub check_unit_file {
+    my $unit_file = shift;
+    return 1 if (script_run("systemctl list-unit-files --all $unit_file*") == 0);
+    return 0;
 }
 
 1;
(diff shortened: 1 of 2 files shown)


mdoucha:
It might be better to check for firewall presence using `systemctl list-unit-files --all $firewall*` instead of hardcoding SYSTEM_ROLE condition. Note that the asterisk at the end is required, otherwise it won't find anything.

mdoucha (changes requested):
It might be better to check for firewall presence using `systemctl list-unit-files --all $firewall*` instead of hardcoding SYSTEM_ROLE condition. Note that the asterisk at the end is required, otherwise it won't find anything.

Amrysliu:
> It might be better to check for firewall presence using `systemctl list-unit-files --all $firewall*` instead of hardcoding SYSTEM_ROLE condition. Note that the asterisk at the end is required, otherwise it won't find anything.

Thanks. Modified.

mdoucha:
1. It'd be better to create a reusable library function for this
2. You need to pass the contents of `$self->firewall`, not the plain string "firewall".

Amrysliu:
> 1. It'd be better to create a reusable library function for this
> 2. You need to pass the contents of `$self->firewall`, not the plain string "firewall".

Good suggestions.  Thanks.

mdoucha:
The library function should go into lib/Utils/Systemd.pm

mdoucha (approved):
LGTM

Amrysliu:
> LGTM

Thanks for your help and reviewing.

rfan1 (approved):
LGTM

tinawang123 (approved):
LGTM

lilyeyes (commented):
LGTM

jouyingbin:
LGTM

jouyingbin (commented):
LGTM

PR #13627, opened 2021-11-08
Journal check: handle cases when there are no entries

jlausuch:
If there are no error messages in the journal, it will print
"-- No entries --" message, which is considered by the test module
as a failure. We should filter those cases.

- Related ticket: https://progress.opensuse.org/issues/101984
- VR: https://openqa.opensuse.org/tests/2022765 https://openqa.suse.de/tests/7629279

### DIFF
--- a/tests/console/journal_check.pm
+++ b/tests/console/journal_check.pm
@@ -52,7 +52,7 @@ sub run {
 
     $self->select_serial_terminal;
 
-    my @journal_output = split(/\n/, script_output("journalctl --no-pager -p ${\get_var('JOURNAL_LOG_LEVEL', 'err')} -o short-precise | tail -n +2"));
+    my @journal_output = split(/\n/, script_output("journalctl --no-pager --quiet -p ${\get_var('JOURNAL_LOG_LEVEL', 'err')} -o short-precise"));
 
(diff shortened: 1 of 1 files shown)


DimStar77:
Maybe filtering as part of an ignore in bug_refs.json could be considered a bit cleaner than an extra grep call

Vogtinator:
Yes, or passing the `-q` option to hide all the non-log messages. That probably drops the boot markers though.

jlausuch:
> Maybe filtering as part of an ignore in bug_refs.json could be considered a bit cleaner than an extra grep call

I can do that, but keep in mind that we will need to update the bugref for new version and will miss the first time it occurs.. If this is a generic "false negative" case, it would be better to handle these kind of situations in the test code.

DimStar77:
> Yes, or passing the `-q` option to hide all the non-log messages. That probably drops the boot markers though.

indeed, the bookmakers vanish with -q as well (tested that before too)

jlausuch:
> Yes, or passing the `-q` option to hide all the non-log messages. That probably drops the boot markers though.

Updated with this suggestion. I have also removed the `tail` as I guess it was just to filter out the same type of output.

mloviska (approved):

PR #13629, opened 2021-11-08
Move is_enabled() to Base widget parent class

ge0r:
- Related ticket: https://progress.opensuse.org/issues/100904
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/7629308#next_previous


jknphy:
As discussed offline, let's get inspired by this C++ example from section 'Composition and interfaces' in https://en.wikipedia.org/wiki/Composition_over_inheritance to move fordward.

PR #13633, opened 2021-11-08
Drop containers/container_diff from JeOS tests

mloviska:
Only HOST related container tests should run on top of JeOS


jlausuch (approved):
LGTM. Thanks, this is a left over I missed from removing this from "latest" jobs :)

PR #13636, opened 2021-11-08
Add container tests to schedule

grisu48:
Add container tests to the publiccloud schedule.

- Related ticket: https://progress.opensuse.org/issues/92779
- Verification run: https://openqa.suse.de/tests/7631389

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -116,6 +116,15 @@ sub load_publiccloud_tests {
             load_publiccloud_consoletests();
             loadtest("publiccloud/ssh_interactive_end", run_args => $args);
         }
+        elsif (get_var('PUBLIC_CLOUD_CONTAINERS')) {
+            my $args = OpenQA::Test::RunArgs->new();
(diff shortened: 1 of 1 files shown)


asmorodskyi (approved):
LGTM

lib/main_publiccloud.pm
```diff
@@ -116,6 +116,15 @@ sub load_publiccloud_tests {
             load_publiccloud_consoletests();
             loadtest("publiccloud/ssh_interactive_end", run_args => $args);
         }
+        elsif (get_var('PUBLIC_CLOUD_CONTAINERS')) {
```
jlausuch, line 119: We could get rid of this variable and use `CONTAINER_RUNTIME`. If present, we schedule container tests.

asmorodskyi: maybe but this would require also changes in maintenance setup , while this PR is simply about adding container tests to latest test suites . So I would do what you suggesting in separate PR ...

jlausuch (approved):

PR #13642, opened 2021-11-09
Wait screen to change after user selected

lemon-suse:
Sometimes the system is slow, we need to wait several seconds for the screen to change.

- Related ticket: https://progress.opensuse.org/issues/101770
- Needles: N/A
- Verification run: 
  http://openqa.nue.suse.com/tests/7629804
  http://openqa.nue.suse.com/tests/7629805
  http://openqa.nue.suse.com/tests/7629806
  http://openqa.nue.suse.com/tests/7629807
  http://openqa.nue.suse.com/tests/7629808

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -298,6 +298,8 @@ sub select_user_gnome {
     elsif (match_has_tag('displaymanager-user-selected')) {
         if ($myuser =~ 'bernhard') {
             send_key 'ret';
+            # sometimes the system is slow, need wait several seconds for the screen to change.
+            wait_still_screen 5 if is_s390x;
         }
         else {
             assert_and_click "displaymanager-$myuser";


lib/x11utils.pm
```diff
@@ -298,6 +298,8 @@ sub select_user_gnome {
     elsif (match_has_tag('displaymanager-user-selected')) {
         if ($myuser =~ 'bernhard') {
             send_key 'ret';
+            # sometimes the system is slow, need wait several seconds for the screen to change.
+            wait_still_screen 5;
```
tinawang123, line 302: I think this is performance issue. If add wait time to this common file. All of others cases will waste 5s.

lemon-suse: We need ensure test won't be blocked, I can limit the arch for s390x since currently only test on s390x have such issues. Later if other platform has the issue then remove the condition is ok.

zhaocongjia:
LGTM.  If you have concerns that it will affect other cases, you can make this s390x specific.

lemon-suse:
> LGTM. If you have concerns that it will affect other cases, you can make this s390x specific.

Ok, thanks.

tinawang123 (approved):
LGTM

PR #13643, opened 2021-11-09
Test rmt server before and after migration

tinawang123:
Test rmt server before and after migration. Before migration need
check rmt server can do sync, enable and list operation. After
migration, need check rmt server can do sync and also has expected
package.

- Related ticket: https://progress.opensuse.org/issues/101340
- Verification run: https://openqa.suse.de/tests/7635384

### DIFF
--- a/tests/console/rmt/rmt_host_migration.pm
+++ b/tests/console/rmt/rmt_host_migration.pm
@@ -1,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: rmt-server
+# Summary: This tests rmt server can work before and after migration
+#    Add rmt configuration test and basic configuration via
+#    rmt-wizard, enable repo, check repo at base system, then upgrade
+#    to latest one. It can still work fine.
+# Maintainer: Yutao wang <yuwang@suse.com>
+
+use base "consoletest";
+use strict;
+use warnings;
+use testapi;
+use repo_tools;
+
+sub run {
+    select_console 'root-console';
+
+    if (check_var('VERSION', get_required_var('ORIGIN_SYSTEM_VERSION'))) {
+        rmt_wizard();
+        # Sync from SCC
+        rmt_sync;
+        # Enable one product
+        assert_script_run("rmt-cli product enable sle-module-live-patching/15/x86_64");
+    }
+
+    # After migation, need do rmt sync again
+    rmt_sync if (check_var('VERSION', get_required_var('UPGRADE_TARGET_VERSION')));
+
+    # Before and after migration, both need rmt gets expected product
+    assert_script_run("rmt-cli product list | grep sle-module-live-patching/15/x86_64");
+}
+
+1;
--- a/schedule/migration/rmt_host_migration.yaml
+++ b/schedule/migration/rmt_host_migration.yaml
@@ -1,0 +1,38 @@
+name:    rmt_host_migration
+description:    >
+     Test rmt server before and after migration
+vars:
+  DESKTOP: gnome
+  BOOT_HDD_IMAGE: 1
+  ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
+  UPGRADE_TARGET_VERSION: '%VERSION%'
+schedule:
+  - migration/version_switch_origin_system
+  - '{{online_migration_test}}'
+  - '{{remove_ltss}}'
+  - console/rmt/rmt_host_migration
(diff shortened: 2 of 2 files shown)


tests/console/rmt/rmt_migration.pm
```diff
@@ -0,0 +1,36 @@
...
+
+sub run {
+    select_console 'root-console';
+
+    if (check_var('VERSION', get_required_var('ORIGIN_SYSTEM_VERSION'))) {
+        rmt_wizard();
+        # Sync from SCC
+        rmt_sync;
+        # Enable one product
+        assert_script_run("rmt-cli product enable sle-module-live-patching/15/x86_64");
```
lemon-suse, line 26: You can use a variable for the hard code live-patching module path, another solution is to set this in SETTING so that we can change the module easily.

tinawang123: This is just test it can be enable. Don't need set it to the setting.
Like rmt_feature, We have discussed before.

lemon-suse:
The name of rmt_migration is too general while haven't show the real test, at least this should tell other that this is a server side migration while not migration on client side normally.

tests/console/rmt/rmt_migration.pm
```diff
@@ -0,0 +1,36 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: rmt-server
+# Summary: Add rmt configuration test and basic configuration via
+#    rmt-wizard, enable repo, check repo at base system, then upgrade
+#    to latest one. Check rmt still can work and repo still exsites.
```
rfan1, line 9: Check rmt still can work and repo still exsites. -> it can still work fine

tests/console/rmt/rmt_migration.pm
rfan1, line 26: Yutao, this test only supports x86_64, correct?

tinawang123:
> The name of rmt_migration is too general while haven't show the real test, at least this should tell other that this is a server side migration while not migration on client side normally.

Has added description at the file. Users can know this is for rmt server testing.
About the file name as most of us agreed this name. I will not change it now.

chcao (approved):
LGTM

lemon-suse (approved):

tests/console/rmt/rmt_host_migration.pm
```diff
@@ -0,0 +1,37 @@
...
+    if (check_var('VERSION', get_required_var('ORIGIN_SYSTEM_VERSION'))) {
+        rmt_wizard();
+        # Sync from SCC
+        rmt_sync;
+        # Enable one product
+        assert_script_run("rmt-cli product enable sle-module-live-patching/15/x86_64");
+    }
+
+    # After migation, need do rmt sync again
+    rmt_sync if (check_var('VERSION', get_required_var('UPGRADE_TARGET_VERSION')));
```
hjluo, line 31: rmt_sync if (check_var('VERSION',  get_required_var('UPGRADE_TARGET_VERSION'))); 
-> if check_var('VERSION',  get_required_var('UPGRADE_TARGET_VERSION'));

and rmt_sync will be called before and after migration, so we can put it out the if {} block

we can use  the follow block to avoid check 2 times for before and after migtation.
if (check_var('VERSION', get_required_var('ORIGIN_SYSTEM_VERSION'))) {
  before migration
} else{
after migration
}

hjluo, line 31: rmt_sync if (check_var('VERSION',  get_required_var('UPGRADE_TARGET_VERSION'))); 
-> if check_var('VERSION',  get_required_var('UPGRADE_TARGET_VERSION'));

and rmt_sync will be called before and after migration, so we can put it out the if {} block

we can use  the follow block to avoid check 2 times for before and after migtation.
if (check_var('VERSION', get_required_var('ORIGIN_SYSTEM_VERSION'))) {
  before migration
} else{
after migration
}

tinawang123: After rmt_sync before migration need enable package, after migration don't need this, just list packages. I just worried if user don't set 'ORIGIN_SYSTEM_VERSION', then it will run else code, but it is origin system, it will get error.

hjluo:
LGTM

tests/console/rmt/rmt_host_migration.pm
hjluo: LGTM

hjluo (approved):
LGTM

PR #13644, opened 2021-11-09
[virt_autotest] Double check xen kernel for xen host at login_console

guoxuguang:
- Description
According to the latest OSD test gi-guest_win2019-on-host_developing-xen on 15-SP4 build61.1, find out that got Xen Domain0 kernel panic during login_console test. then boot up a normal kernel wihout xen kernel to continue the following test. mean that just only match virttest-bootmenu-xen-kernel needle at login_console step for xen host was not enough during getting kernel panic. 
So, we'd better double-check xen kernel module for xen host after matching virttest-bootmenu-xen-kernel needle successfully at login_console for this PR purpose. 
- Verification run: 
[gi-guest_win2019-on-host_developing-xen](https://openqa.suse.de/tests/7636500#)
wip

### DIFF
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -146,7 +146,18 @@ sub login_to_console {
     set_ssh_console_timeout_before_use if (is_remote_backend && is_aarch64 && get_var('IPMI_HW') eq 'thunderx');
     # use console based on ssh to avoid unstable ipmi
     use_ssh_serial_console;
+    # double-check xen kernel for xen host
+    $self->double_check_xen_role if is_xen_host;
 
+}
+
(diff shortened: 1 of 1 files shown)


waynechen55:
I think you'd better also check whether the booted os fs is 'read-only'.

guoxuguang:
> I think you'd better also check whether the booted os fs is 'read-only' which prevents further operation from being executed. This can happen if somehow host boots into a mismatched os or snapshot.

sure,  after match virttest-bootmenu-xen-kernel needle, double-check the xen kernel for xen vm host at login_console as this PR purpose. 

yes, try to figure out the best way to check with the 'read-only' FS.

guoxuguang:
duplicated by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13702. so, updated as closed.

PR #13645, opened 2021-11-09
Cleanup lib/main_common.pm into lib/main_containers.pm

pdostal:
As we now have `lib/main_containers.pm` we should use it.
This is about cleaning `lib/main_common.pm`.

- Related ticket: [poo#102092](https://progress.opensuse.org/issues/102092)
- Verification run: TBD

PR #13650, opened 2021-11-09
containers: Do not load main_common::loadtest

pdostal:
This is just a small fix of `lib/main_containers.pm`:
We `use main_common qw(loadtest boot_to_desktop); but:
 * We don't use `boot_to_deskotp()`
 * The `main_common::loadtest()` is just wrapper around `autotest::loadtest()`

So here I'm getting rid of this dependency.

- Verification run: TBD

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -11,17 +11,22 @@ use base 'Exporter';
 use Exporter;
 use utils;
 use version_utils;
-use main_common qw(loadtest boot_hdd_image);
 use testapi qw(check_var get_required_var get_var);
 use Utils::Architectures;
 use strict;
 use warnings;
-
+use Carp 'croak';
 
 our @EXPORT = qw(
   is_container_test
   load_container_tests
 );
+
+sub loadtest {
+    my ($test, %args) = @_;
+    croak "extensions are not allowed here '$test'" if $test =~ /\.pm$/;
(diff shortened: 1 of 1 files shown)


asmorodskyi (approved):
LGTM

lib/main_containers.pm
```diff
@@ -11,18 +11,23 @@ use base 'Exporter';
...
 use warnings;
-
+use Carp 'croak';
 
 our @EXPORT = qw(
   is_container_test
   load_container_tests
 );
 
+sub loadtest {
```
cfconrad, line 25: I would make it more clear for future reviewers that we did code duplication on purpose.
```suggestion
==head2
  container_loadtest($test)
  
We duplicate c<loadtest()> function here to not have cycle dependencies between c<main_common> and this module.
==cut
sub container_loadtest {
```

And I would use a "unique" name to make it more clear that we use our own.
WDYT?

b10n1k (commented):
i am not sure about this. the function is completely the same as the original. The `main_common` doesnt go anywhere in the near future. if someone change something there we do not get it here and we will have to maintain it in two different places, but not for good reason.

jlausuch:
After latest update of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13645 I think this is not needed.

pdostal:
I will probably close this however let's wait for #13645.

PR #13659, opened 2021-11-10
Add a module to disable grub timeout with libyui-rest-api

JRivrain:
See https://progress.opensuse.org/issues/98889

VR: 
textmode: http://waaa-amazing.suse.cz/tests/15958
prod : https://openqa.suse.de/tests/overview?build=JRivrain/os-autoinst-distri-opensuse%2313659
yast2 gui: http://waaa-amazing.suse.cz/tests/15960

### DIFF
--- a/lib/YaST/Bootloader/BootCodeOptionsPage.pm
+++ b/lib/YaST/Bootloader/BootCodeOptionsPage.pm
@@ -3,24 +3,18 @@
 # Copyright 2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
-# Summary: The class introduces all accessing methods for Bootloader
+# Summary: The class introduces all accessing methods for Boot code options tab in bootloader module
 # Maintainer: QE YaST <qa-sle-yast@suse.de>
 
 package YaST::Bootloader::BootCodeOptionsPage;
+use parent 'Installation::Navigation::NavigationBase';
 use strict;
 use warnings;
 
-sub new {
-    my ($class, $args) = @_;
-    my $self = bless {
-        app => $args->{app}
-    }, $class;
-    return $self->init();
-}
-
 sub init {
     my ($self) = @_;
-    $self->{cmb_bootloader} = $self->{app}->combobox({id => '"Bootloader::LoaderTypeWidget"'});
+    $self->SUPER::init();
+    $self->{cmb_bootloader} = $self->{app}->combobox({id => "\"Bootloader::LoaderTypeWidget\""});
     $self->{cb_write_to_partition} = $self->{app}->checkbox({id => 'boot'});
     $self->{cb_bootdev} = $self->{app}->checkbox({id => 'mbr'});
     $self->{cb_custom_boot} = $self->{app}->checkbox({id => 'custom'});
@@ -28,7 +22,7 @@ sub init {
     $self->{cb_trusted_boot} = $self->{app}->checkbox({id => '"Bootloader::TrustedBootWidget"'});
     $self->{cb_generic_to_mbr} = $self->{app}->checkbox({id => '"Bootloader::GenericMBRWidget"'});
     $self->{cb_set_active_flag} = $self->{app}->checkbox({id => '"Bootloader::ActivateWidget"'});
-    $self->{tb_boot_options} = $self->{app}->tab({id => '"CWM::DumbTabPager"'});
+    $self->{tb_boot_loader_settings} = $self->{app}->tab({id => '"CWM::DumbTabPager"'});
     $self->{btn_ok} = $self->{app}->button({id => 'next'});
     $self->{btn_cancel} = $self->{app}->button({id => 'abort'});
     return $self;
@@ -36,7 +30,7 @@ sub init {
 
 sub is_shown {
     my ($self) = @_;
-    return $self->{tb_boot_options}->selected_tab();
+    return $self->{cmb_bootloader}->exist();
 }
 
 sub get_bootloader_type {
@@ -94,14 +88,9 @@ sub uncheck_write_to_mbr {
     $self->{cb_bootdev}->uncheck();
 }
 
-sub press_ok {
+sub switch_tab_bootloader_options {
     my ($self) = @_;
-    return $self->{btn_ok}->click();
-}
-
-sub press_cancel {
-    my ($self) = @_;
-    return $self->{btn_cancel}->click();
+    $self->{tb_boot_loader_settings}->select("Boot&loader Options");
 }
 
 1;
--- a/lib/YaST/Bootloader/BootloaderOptionsPage.pm
+++ b/lib/YaST/Bootloader/BootloaderOptionsPage.pm
@@ -1,0 +1,31 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for Boot loader options tab in bootloader module
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Bootloader::BootloaderOptionsPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my $self = shift;
+    $self->SUPER::init();
+    $self->{txt_grub_timeout} = $self->{app}->textbox({id => "\"Bootloader::TimeoutWidget\""});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    $self->{txt_grub_timeout}->exist();
+}
+
+sub disable_grub_timeout {
+    my ($self) = @_;
+    $self->{txt_grub_timeout}->set("-1");
+}
+
+1;
--- a/lib/YaST/Bootloader/BootloaderSettingsController.pm
+++ b/lib/YaST/Bootloader/BootloaderSettingsController.pm
@@ -6,12 +6,12 @@
 # Summary: Controller for YaST bootloader module.
 # Maintainer: QE YaST <qa-sle-yast@suse.de>
 
-package YaST::Bootloader::BootloaderController;
+package YaST::Bootloader::BootloaderSettingsController;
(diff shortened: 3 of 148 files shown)


lib/YaST/Bootloader/BootLoaderSettingsPage.pm
```diff
@@ -0,0 +1,39 @@
...
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+    $self->{tb_boot_loader_settings} = $self->{app}->tab({id => '"CWM::DumbTabPager"'});
```
jknphy, line 25: it would be more simple to include this control in each page (not even worth it to inherit IMO) because I think it is tricky to know where you are in which tab or which tab is selected by default when you land (referring to the `is_shown` method for example), this is part of the flow of the controller for each product to use a series of steps needed.

lib/YaST/Bootloader/BootloaderController.pm
```diff
@@ -10,7 +10,9 @@ package YaST::Bootloader::BootloaderController;
 use strict;
```
jknphy, line 10: This class should be called now `BootLoaderSettingsController` according to new names we are using for most of the screens.

tests/installation/disable_grub_timeout_yui.pm
```diff
@@ -0,0 +1,18 @@
...
+# Summary: Disable grub timeout from the Installer
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    $testapi::distri->get_overview_controller()->open_booting_options();
+    $testapi::distri->get_bootloader()->disable_grub_timeout();
```
jknphy, line 15: it should be called `get_boot_loader_settings()`

lib/YaST/Bootloader/BootloaderController.pm
```diff
@@ -10,7 +10,9 @@ package YaST::Bootloader::BootloaderController;
...
 use warnings;
 use YuiRestClient;
-use YaST::Bootloader::BootCodeOptionsPage;
+use YaST::Bootloader::BootCodeOptionsTab;
+use YaST::Bootloader::BootloaderOptionsTab;
+use YaST::Bootloader::BootLoaderSettingsPage;
```
jknphy, line 15: to know where we are with is_shown we need 3 different pages. In this case I think it is a good choice not to try 1-1 (1 controller - 1 page) because all the options in all tabs seems independent, it is just about organization/visualization in the screen. In other words, 1 controller 3 pages seems fine. But each page should contain the tab repeated and inheriting from NavigationBase. Having some page that represent the frame or the tab I'm not sure it could work well because you might not know where you are in each moment, repeating the tab in each page will give us the option to navigate only to the pages that make sense, not the current one for example. wdyt?

jknphy (commented):
first round of review, let's check together offline when you have time, because that tab navigation is a bit tricky.

lib/YaST/Bootloader/BootloaderController.pm
JRivrain: It's just a logical problem for me: If I'm on the tab 1, and I want to switch to tab 2, where should I put the method to switch to tab 2 ? on Tab1Page, or on Tab2Page ? I can put it on Tab2Page, but that is somehow incorrect, as I am on tab 1 when I call this. To have a separate page representing the frame eliminates that logical issue. In this case, if we change settings within a tab, we will use a specific "is_shown" for it, so there is no ambiguity on where we are at that point, so we do not need to check it again when we press "next".

lib/YaST/Bootloader/BootCodeOptionsTab.pm
```diff
@@ -3,10 +3,10 @@
 # Copyright 2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
-# Summary: The class introduces all accessing methods for Bootloader
+# Summary: The class introduces all accessing methods for Boot code options tab in bootloader module
 # Maintainer: QE YaST <qa-sle-yast@suse.de>
 
-package YaST::Bootloader::BootCodeOptionsPage;
+package YaST::Bootloader::BootCodeOptionsTab;
```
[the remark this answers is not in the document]
JRivrain: Right, though we already started diverging largely on this with pop-ups for example.

lib/YaST/Bootloader/BootCodeOptionsTab.pm
```diff
@@ -3,10 +3,10 @@
 # Copyright 2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
-# Summary: The class introduces all accessing methods for Bootloader
+# Summary: The class introduces all accessing methods for Boot code options tab in bootloader module
```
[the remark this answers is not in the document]
JRivrain: The question is if we will cover more of that module in the installer, or in the desktop module ? Back when I implemented the bootcode options module I had the misunderstanding that we wanted to merge everything under Yast.

lib/YaST/Bootloader/BootloaderController.pm
JRivrain: I did as you suggest, In order to make sure the whole module is fully loaded, not just the tabs bar. Not resolving the conversation, so we have track of why we implemented it this way.

lib/YaST/Bootloader/BootCodeOptionsTab.pm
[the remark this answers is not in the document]
jknphy: the modules will come later in the roadmap and now the code with more quality and more homogeneous is in Installation, we don't know yet, because this module is exactly the same in both, for testing error message is great in the installer, the module doesn't help the user at all when the installer give you the pre-selected options that makes sense and error message in Installation Settings, in the installed system you are on your own, but on the other hand for the other tabs, kernel or change the theme, that are more suitable for the installed system. Feel free to leave it as it is, it is just about code organization.

lib/YaST/Bootloader/BootCodeOptionsPage.pm
```diff
@@ -20,23 +21,24 @@ sub new {
...
+    $self->{tb_boot_loader_settings} = $self->{app}->tab({id => '"CWM::DumbTabPager"'});
     $self->{btn_ok} = $self->{app}->button({id => 'next'});
     $self->{btn_cancel} = $self->{app}->button({id => 'abort'});
     return $self;
 }
 
 sub is_shown {
     my ($self) = @_;
-    return $self->{tb_boot_options}->selected_tab();
+    return $self->{cmb_bootloader}->value();
```
jknphy, line 41: like in other cases `->exist();` should work.

lib/YaST/Bootloader/BootloaderOptionsPage.pm
```diff
@@ -0,0 +1,39 @@
...
+
+# Summary: The class introduces all accessing methods for Boot loader options tab in bootloader module
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Bootloader::BootloaderOptionsPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub new {
```
jknphy, line 14: new() method is not needed, inherited from NavigationBase.

lib/YaST/Bootloader/BootloaderOptionsPage.pm
```diff
@@ -0,0 +1,39 @@
...
+sub init {
+    my $self = shift;
+    $self->SUPER::init();
+    $self->{txt_grub_timeout} = $self->{app}->textbox({id => "\"Bootloader::TimeoutWidget\""});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    $self->{txt_grub_timeout}->value();
```
jknphy, line 31: same, here, the `exists` method should do the work.

lib/YaST/Bootloader/BootloaderOptionsPage.pm
```diff
@@ -0,0 +1,39 @@
...
+}
+
+sub disable_grub_timeout {
+    my ($self) = @_;
+    $self->{txt_grub_timeout}->set("-8");
```
jknphy, line 36: why `-8` shouldn't be `-1`, maybe it goes to the value we want because it is the minimum of the range, but better be specific as the user cannot type that afir, only the libyui-rest-api can.

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -22,6 +22,8 @@ sub new {
 sub init {
     my ($self) = @_;
     $self->{BootCodePage} = YaST::Bootloader::BootCodeOptionsPage->new({app => YuiRestClient::get_app()});
+    $self->{BootloaderPage} = YaST::Bootloader::BootloaderOptionsPage->new({app => YuiRestClient::get_app()});
+    $self->{BootloaderPage} = YaST::Bootloader::BootloaderOptionsPage->new({app => YuiRestClient::get_app()});
```
jknphy, line 26: this line is repeated.

tests/installation/disable_boot_menu_timeout.pm
```diff
@@ -0,0 +1,18 @@
+# SUSE's openQA tests
```
jknphy, line 1: (optional) could we create intermediate folder under /installation/ to put there all test related with bootloader settings?

jknphy (commented):
LGTM, only some minor things, good job!
I would suggest to run some verification in openSUSE as well.

lib/YaST/Bootloader/BootloaderOptionsPage.pm
JRivrain: OMG that typo stayed right under my eyes the whole time, thanks for spotting it !

okurz:
Causing problems like in maintenance tests https://openqa.opensuse.org/tests/2036330#step/disable_boot_menu_timeout/1

PR #13663, opened 2021-11-11
console/btrfs_qgroups: Explicitly pass the reflink option to cp

Vogtinator:
Instead of relying on the default value of that option (which changed with
coreutils 9.0), pass it explicitly. This also allows testing that using
reflinks does not cause the quota to be exceeded.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1192570
- Verification run: https://openqa.opensuse.org/tests/2028742#step/btrfs_qgroups/75

### DIFF
--- a/tests/console/btrfs_qgroups.pm
+++ b/tests/console/btrfs_qgroups.pm
@@ -80,7 +80,8 @@ sub run {
     # Check limits
     assert_script_run "dd if=/dev/zero bs=10M count=3 of=nofile";
     foreach my $c ('a' .. 'b') {
-        assert_script_run "! cp nofile $c/nofile";
+        assert_script_run "cp --reflink=always nofile $c/nofile";
+        assert_script_run "! cp --reflink=never --remove-destination nofile $c/nofile";
         assert_script_run "sync && rm $c/nofile";
     }
     assert_script_run "cp nofile c/nofile";


tests/console/btrfs_qgroups.pm
```diff
@@ -80,7 +80,8 @@ sub run {
     # Check limits
     assert_script_run "dd if=/dev/zero bs=10M count=3 of=nofile";
     foreach my $c ('a' .. 'b') {
```
coolo, line 82: btw, is it just me who is confused that `$c` is a or b? :)

coolo (approved):
makes sense to me not to rely on cp default. And I checked --reflink is supported at least since 2010

tests/console/btrfs_qgroups.pm
Vogtinator: Absolutely not. It should've been called `$i`!

DimStar77 (approved):
Sounds reasonable.

dzedro:
Does that mean this should not happen ?
https://openqa.suse.de/tests/7654160#step/btrfs_qgroups/81 12 SP4
https://openqa.suse.de/tests/7654165#step/btrfs_qgroups/83 15 GA x86_64
https://openqa.suse.de/tests/7654164#step/btrfs_qgroups/83 15 GA aarch64

Vogtinator:
> Does that mean this should not happen ? https://openqa.suse.de/tests/7654160#step/btrfs_qgroups/81 12 SP4 https://openqa.suse.de/tests/7654165#step/btrfs_qgroups/83 15 GA x86_64 https://openqa.suse.de/tests/7654164#step/btrfs_qgroups/83 15 GA aarch64

No, that looks broken. Does it also happen on newer SLE? If not, it might be a kernel bug.

dzedro:
> > Does that mean this should not happen ? https://openqa.suse.de/tests/7654160#step/btrfs_qgroups/81 12 SP4 https://openqa.suse.de/tests/7654165#step/btrfs_qgroups/83 15 GA x86_64 https://openqa.suse.de/tests/7654164#step/btrfs_qgroups/83 15 GA aarch64
> 
> No, that looks broken. Does it also happen on newer SLE? If not, it might be a kernel bug.

Today it happened only on 12 SP4 and 15 GA

Vogtinator:
> > > Does that mean this should not happen ? https://openqa.suse.de/tests/7654160#step/btrfs_qgroups/81 12 SP4 https://openqa.suse.de/tests/7654165#step/btrfs_qgroups/83 15 GA x86_64 https://openqa.suse.de/tests/7654164#step/btrfs_qgroups/83 15 GA aarch64
> > 
> > 
> > No, that looks broken. Does it also happen on newer SLE? If not, it might be a kernel bug.
> 
> Today it happened only on 12 SP4 and 15 GA

Ok. I can't find any existing bugreport, so I filed one: https://bugzilla.suse.com/show_bug.cgi?id=1192630

There's still a bug with this PR, though: The `--reflink=never` value required on TW is not supported by coreutils < 8.30:

```
# ! cp --reflink=never --remove-destination nofile a/nofile; echo wx~v4-$?-
cp: invalid argument ‘never’ for ‘--reflink’
Valid arguments are:
  - ‘auto’
  - ‘always’
Try 'cp --help' for more information.
wx~v4-0-
```

I'll make a PR for that.

Vogtinator:
@dzedro

It might be a race condition with quota scanning.

Can you start a test run (or better, multiple) with https://github.com/Vogtinator/os-autoinst-distri-opensuse/tree/quotarescan?

Vogtinator:
> @dzedro
> 
> It might be a race condition with quota scanning.
> 
> Can you start a test run (or better, multiple) with https://github.com/Vogtinator/os-autoinst-distri-opensuse/tree/quotarescan?

From #13676:

> https://openqa.suse.de/tests/7654861#step/btrfs_qgroups/90
> https://openqa.suse.de/tests/7654862#step/btrfs_qgroups/90
> https://openqa.suse.de/tests/7654863#step/btrfs_qgroups/90

So they worked with the quota rescan at least three times. That either means they randomly didn't show up or waiting for the scan to complete is really necessary. I suppose the best way to find that out is to watch it for some time.

I opened https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13677.

PR #13665, opened 2021-11-11
Add azure nvme performance check

grisu48:
Adds a nmve performance check on Azure l8s_v2 instances. This test
checks the underlying performance via hdparm and compares it to the
documented values from the azure specs.

- Related ticket: https://progress.opensuse.org/issues/58646
- Verification run: https://openqa.suse.de/tests/7680720

### DIFF
--- a/tests/publiccloud/az_l8s_nvme.pm
+++ b/tests/publiccloud/az_l8s_nvme.pm
@@ -1,0 +1,52 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Test nvme performance on a L8s_v2 Azure instance
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+# Benchmark the given device with hdparm and fail, if the given threshold [MB/sec] is not reached
+sub benchmark_device {
+    my ($device, $threshold) = @_;
+
+    my $output = script_output("hdparm -t $device | grep 'Timing buffered disk reads'");
+    # e.g. 'Timing buffered disk reads: 15528 MB in  3.00 seconds = 5174.94 MB/sec'
+    if ($output =~ "Timing buffered disk reads: (.*)=( )*(?<throughput>[0-9]+\.[0-9]*) MB/sec") {
+        my $throughput = $+{throughput};
+        record_info("$device", "Throughput: $throughput MB/sec\nThreshold: $threshold MB/sec");
+        die "Threshold for $device not reached" if ($throughput < $threshold);
+    } else {
+        die "Unexpected hdparm output";
+    }
+}
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+
+    die "This test only works on Azure L instances" unless (get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'L(8|16|32|48|64|80)s_v2');
+
+    # Ensure the required disks are present
+    record_info('lsbkl', script_output('lsblk'));
+    assert_script_run('lsblk | grep sda', fail_message => "sda disk is not present");
+    assert_script_run('lsblk | grep sdb', fail_message => "sdb disk is not present");
+    assert_script_run('lsblk | grep nvme0n1', fail_message => "nvme disk is not present");
+
+    # Install required utilities
+    zypper_call("in hdparm");
+
+    # Ensure the performance is above the expected values
+    # See https://docs.microsoft.com/en-us/azure/virtual-machines/lsv2-series
+    benchmark_device("/dev/sdb", 80);
+    benchmark_device("/dev/nvme0n1", 2000);
+    # Note: L16sv2 has two, L32sv2 has three NVME disks, however we are only testing the first one. This could be extended, if required.
+}
+
+1;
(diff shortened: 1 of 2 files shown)


asmorodskyi:
nice !

asmorodskyi (approved):
LGTM

pdostal (approved):

tests/publiccloud/az_l8s_nvme.pm
```diff
@@ -0,0 +1,50 @@
...
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+
+    die "This test only works on Azure L8s instances" unless (get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'L(8|16|32|48|64|80)s_v2');
+
+    # Ensure the requested disks are present
+    assert_script_run('lsblk | grep sda', fail_message => "sda disk is not present");
+    assert_script_run('lsblk | grep sdb', fail_message => "sdb disk is not present");
+    assert_script_run('lsblk | grep nvme0n1', fail_message => "nvme disk is not present");
```
jlausuch, line 39: is this 100% sure it will be named always like this and not `nvme1n1` or similar? Same for `sda` and `sdb`.

jlausuch: Maybe, for troubleshooting purposes, it would be also nice to have a full `lsblk` output, maybe a record_info with that output?

tests/publiccloud/az_l8s_nvme.pm
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+        die "Unexpected hdparm output";
+    }
+}
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+
+    die "This test only works on Azure L8s instances" unless (get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'L(8|16|32|48|64|80)s_v2');
```
jlausuch, line 34:
```suggestion
    die "This test only works on Azure L instances" unless (get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'L(8|16|32|48|64|80)s_v2');
```
So, it works on L8, L16, etc... not only in L8.

jlausuch: Or maybe you should limit the regex to `L8` only, cause in the test below you do
`benchmark_device("/dev/nvme0n1", 2000);`  and 2000 is specific to L8 only, for L16 the numbers are different.

tests/publiccloud/az_l8s_nvme.pm
```diff
@@ -0,0 +1,50 @@
...
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+# Benchmark the given device with hdparm and fail, if the given threshold [MB/sec] is not reached
+sub benchmark_device {
+    my ($device, $threshold) = @_;
+
+    my $output = script_output("hdparm -t $device | grep 'Timing buffered disk reads'");
```
jlausuch, line 19: how long does this command take? Do we need timeout?

tests/publiccloud/az_l8s_nvme.pm
grisu48: Since this is the first nvme drive it will always be `nvme0n1`. I will add a `lsblk` output however before.

tests/publiccloud/az_l8s_nvme.pm
grisu48: L16sv2 will have two NVME's, so in all of those cases the performance test on the first nvme is still valid.

tests/publiccloud/az_l8s_nvme.pm
grisu48: No this takes only a couple of seconds, so in all sane configuration it should be well below the typical timeout of 30 seconds.

tests/publiccloud/az_l8s_nvme.pm
jlausuch: ah, ok, I thought it was just one disk, but with better performance. Ok, then my first comment sill applies :)

tests/publiccloud/az_l8s_nvme.pm
jlausuch: Ok, I thought this was doing some performance stuff :) Good

tests/publiccloud/az_l8s_nvme.pm
grisu48: I added a comment at the end of the test that makes this more clear to our future selves ;-)

grisu48:
New VR: https://openqa.suse.de/tests/7679119

tests/publiccloud/az_l8s_nvme.pm
```diff
@@ -0,0 +1,52 @@
...
+}
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+
+    die "This test only works on Azure L instances" unless (get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'L(8|16|32|48|64|80)s_v2');
+
+    # Ensure the required disks are present
+    script_run("lsblk");
```
jlausuch, line 37: I would have done `record_info('lsbkl', script_output('lsblk'))`, but ok :)

jlausuch (approved):

tests/publiccloud/az_l8s_nvme.pm
grisu48: Nothing we cannot change ;-)

grisu48:
New VR: https://openqa.suse.de/tests/7679136

PR #13666, opened 2021-11-11
Containers: Check firewall with podman

pdostal:
- Related ticket: [poo#94132](https://progress.opensuse.org/issues/94132)
- Verification run: [Docker@SLE15-SP3](http://pdostal-server.suse.cz/tests/12840), [Podman@SLE15-SP3](http://pdostal-server.suse.cz/tests/12840), [Docker@Tumbleweed](http://pdostal-server.suse.cz/tests/12837), [Podman@Tumbleweed](http://pdostal-server.suse.cz/tests/12841), [Containers@SLE15-SP3-EC2](http://pdostal-server.suse.cz/tests/12836)

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -20,7 +20,7 @@ use warnings;
 use version_utils;
 use Mojo::Util 'trim';
 
-our @EXPORT = qw(test_seccomp basic_container_tests get_vars can_build_sle_base check_docker_firewall
+our @EXPORT = qw(test_seccomp basic_container_tests get_vars can_build_sle_base
   get_docker_version check_runtime_version container_ip registry_url);
 
 sub test_seccomp {
@@ -38,29 +38,6 @@ sub test_seccomp {
     else {
         record_info('seccomp', 'Docker Engine supports seccomp');
     }
-}
-
-sub check_docker_firewall {
-    my $container_name = 'sut_container';
-    my $docker_version = get_docker_version();
-    systemctl('is-active firewalld');
-    my $running = script_output qq(docker ps -q | wc -l);
-    validate_script_output('ip a s docker0', sub { /state DOWN/ }) if $running == 0;
-    # Docker zone is created for docker version >= 20.10 (Tumbleweed), but it
-    # is backported to docker 19 for SLE15-SP3 and for Leap 15.3
-    if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {
-        assert_script_run "firewall-cmd --list-all --zone=docker";
-        validate_script_output "firewall-cmd --list-interfaces --zone=docker", sub { /docker0/ };
-    }
-    # Rules applied before DOCKER. Default is to listen to all tcp connections
-    # ex. output: "1           0        0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0"
-    validate_script_output "iptables -L DOCKER-USER -nvx --line-numbers", sub { /1.+all.+0\.0\.0\.0\/0\s+0\.0\.0\.0\/0/ };
-    assert_script_run "docker run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine');
-    my $container_ip = container_ip($container_name, 'docker');
-    # Each running container should have added a new entry to the DOCKER zone.
-    # ex. output: "1           0        0 ACCEPT     tcp  --  !docker0 docker0  0.0.0.0/0            172.17.0.2           tcp dpt:1234"
-    validate_script_output "iptables -L DOCKER -nvx --line-numbers", sub { /1.+ACCEPT.+!docker0 docker0.+$container_ip\s+tcp dpt:1234/ };
-    assert_script_run "docker kill $container_name ";
 }
 
 sub get_docker_version {
--- a/tests/containers/docker.pm
+++ b/tests/containers/docker.pm
@@ -32,6 +32,7 @@ use version_utils qw(is_jeos);
 use containers::utils;
 use containers::container_images;
 use publiccloud::utils;
+use Utils::Systemd qw(systemctl disable_and_stop_service);
 
 my $stop_firewall = 0;    # Post-run flag to stop the firewall (failsafe)
 
@@ -46,9 +47,11 @@ sub run {
     test_seccomp();
 
     if ($self->firewall() eq 'firewalld') {
-        # on publiccloud we need to install firewalld first
-        install_and_start_firewalld() if (is_publiccloud || is_jeos);
-        check_docker_firewall();
+        zypper_call('in ' . $self->firewall()) if (is_publiccloud || is_jeos);
+        systemctl('restart ' . $self->firewall());
+        systemctl('restart docker');
+        $stop_firewall = 1;
+        $engine->check_containers_firewall();
     }
 
     # Run basic runtime tests
@@ -62,26 +65,21 @@ sub run {
 
 sub post_fail_hook {
     my $self = shift;
-    cleanup();
+    cleanup($self->firewall());
     $self->SUPER::post_fail_hook;
 }
 
 sub post_run_hook {
     my $self = shift;
-    cleanup();
+    cleanup($self->firewall());
     $self->SUPER::post_run_hook;
-}
-
-sub install_and_start_firewalld() {
-    zypper_call('install firewalld');
-    systemctl('start firewalld');
-    systemctl('restart docker');
-    $stop_firewall = 1;
 }
 
 # must ensure firewalld is stopped, if it is only enabled in this test (e.g. publiccloud test runs)
 sub cleanup() {
-    script_run('systemctl stop firewalld; systemctl restart docker') if $stop_firewall;
+    my $firewall = shift;
+    disable_and_stop_service($firewall) if $stop_firewall;
+    systemctl('restart docker');
 }
 
 1;
--- a/tests/containers/podman.pm
+++ b/tests/containers/podman.pm
@@ -22,10 +22,15 @@
 use Mojo::Base 'containers::basetest';
 use testapi;
 use utils;
+use version_utils;
 use registration;
 use containers::common;
 use containers::utils;
 use containers::container_images;
+use publiccloud::utils;
+use Utils::Systemd qw(systemctl disable_and_stop_service);
+
+my $stop_firewall = 0;    # Post-run flag to stop the firewall (failsafe)
 
 sub run {
     my ($self) = @_;
@@ -33,6 +38,13 @@ sub run {
 
     my $dir = "/root/DockerTest";
     my $podman = $self->containers_factory('podman');
+
+    if ($self->firewall() eq 'firewalld') {
+        zypper_call('in ' . $self->firewall()) if (is_publiccloud || is_jeos);
+        systemctl('restart ' . $self->firewall());
+        $stop_firewall = 1;
+        $podman->check_containers_firewall();
+    }
 
     # Run basic runtime tests
     basic_container_tests(runtime => $podman->runtime);
@@ -45,10 +57,23 @@ sub run {
 
 sub post_fail_hook {
     my ($self) = @_;
+    cleanup($self->firewall());
     select_console 'log-console';
     script_run "podman version | tee /dev/$serialdev";
     script_run "podman info --debug | tee /dev/$serialdev";
     $self->SUPER::post_fail_hook;
+}
+
+sub post_run_hook {
+    my $self = shift;
+    cleanup($self->firewall());
+    $self->SUPER::post_run_hook;
+}
+
+# must ensure firewalld is stopped, if it is only enabled in this test (e.g. publiccloud test runs)
+sub cleanup() {
+    my $firewall = shift;
+    disable_and_stop_service($firewall) if $stop_firewall;
 }
 
 1;
--- a/lib/containers/docker.pm
+++ b/lib/containers/docker.pm
@@ -9,8 +9,9 @@
 package containers::docker;
 use Mojo::Base 'containers::engine';
 use testapi;
-use containers::utils qw(registry_url);
+use containers::utils qw(registry_url get_docker_version check_runtime_version container_ip);
 use containers::common qw(install_docker_when_needed);
+use version_utils qw(is_sle is_leap);
 use utils qw(systemctl file_content_replace);
 use version_utils qw(get_os_release);
 has runtime => 'docker';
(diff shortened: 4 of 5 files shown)


tests/containers/docker.pm
```diff
@@ -47,8 +47,11 @@ sub run {
 
     if ($self->firewall() eq 'firewalld') {
         # on publiccloud we need to install firewalld first
-        install_and_start_firewalld() if (is_publiccloud || is_jeos);
-        check_docker_firewall();
+        zypper_call('install firewalld') if (is_publiccloud || is_jeos);
+        systemctl('restart firewalld');
```
b10n1k, line 51: you might want to check `lib/services/firewall.pm`

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+        my $container_ip = container_ip($container_name, 'docker');
+        # Each running container should have added a new entry to the DOCKER zone.
+        # ex. output: "1           0        0 ACCEPT     tcp  --  !docker0 docker0  0.0.0.0/0            172.17.0.2           tcp dpt:1234"
+        validate_script_output "iptables -L DOCKER -nvx --line-numbers", sub { /1.+ACCEPT.+!docker0 docker0.+$container_ip\s+tcp dpt:1234/ };
+        assert_script_run "docker kill $container_name ";
+
+        assert_script_run "sysctl -a | grep '\\.forwarding'";
+        assert_script_run "firewall-cmd --list-all-zones";
+        script_run "docker run --rm " . registry_url('alpine') . " ip a s";
+        script_run "docker run --rm " . registry_url('alpine') . " ip r s";
```
b10n1k, line 71: no validation. no sure if you need to have the info or what. Also you could use the `engine` but then more changes will be needed. keep that as a comment and resolve it if you want

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+
+        # Connectivity check
+        assert_script_run "docker run --rm " . registry_url('alpine') . " wget " . get_var('OPENQA_URL');
+    } elsif ($runtime eq 'podman') {
```
b10n1k, line 75: I wonder if this can be omitted. That was the whole point of the design after all. it is up to you

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+        my $container_name = 'sut_container';
```
b10n1k, line 76: this could be one definition, right?

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+        systemctl('is-active firewalld');
+
+        # cni-podman0 interface is created when running the first container
+        assert_script_run "podman run --rm " . registry_url('alpine');
+        validate_script_output('ip a s cni-podman0', sub { /,UP/ });
+
+        # Cheking rules of specific running container
+        assert_script_run "podman run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine');
+        assert_script_run "iptables -vnL";
```
b10n1k, line 85: IIRC `iptables` has been replaced by `nftables` in some products. Idk if we run this module in any of them but keep that in mind.

b10n1k (commented):
please dont take it as a review but as some comments.

lib/containers/utils.pm
pdostal: Can you help me how to fix this? I'd like to do it properly but don't know how.

jlausuch: @b10n1k do you mean adding this logic to the engine classes and then just do `$engine->check_firewall()`?

tests/containers/podman.pm
```diff
@@ -45,10 +57,22 @@ sub run {
...
 
+sub post_run_hook {
+    my $self = shift;
+    cleanup();
+    $self->SUPER::post_run_hook;
+}
+
+# must ensure firewalld is stopped, if it is only enabled in this test (e.g. publiccloud test runs)
+sub cleanup() {
+    script_run('systemctl stop firewalld') if $stop_firewall;
```
jlausuch, line 75:
```suggestion
    disable_and_stop_service(opensusebasetest::firewall); if $stop_firewall;
```

tests/containers/docker.pm
jlausuch: I thought the same, but after a quick look at that library, it's really missing a lot of things... To start with, `install_service` installs `SuSEfirewall2` instead of `firewalld`..., then, there is no `restart` method... 

However, what I would do is replace `firewalld` by `opensusebasetest::firewall`

lib/containers/utils.pm
b10n1k: yes. this already discussed with @pdostal offline. it is up to him

jlausuch: Yes, that would be actually very nice. But I'm ok if that gets implemented in a follow-up step.

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
-    my $running = script_output qq(docker ps -q | wc -l);
-    validate_script_output('ip a s docker0', sub { /state DOWN/ }) if $running == 0;
-    # Docker zone is created for docker version >= 20.10 (Tumbleweed), but it
-    # is backported to docker 19 for SLE15-SP3 and for Leap 15.3
-    if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {
-        assert_script_run "firewall-cmd --list-all --zone=docker";
-        validate_script_output "firewall-cmd --list-interfaces --zone=docker", sub { /docker0/ };
+sub check_containers_firewall {
+    my ($runtime) = @_;
+    record_info "firewall", "Checking that firewall is enabled, properly configured and containers can reach the Internet";
```
grisu48, line 45: If such a message is required, isn't that a clear indicator that it should be a separate test module and we are doing too much in one single module already? :thinking:

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+        my $container_name = 'sut_container';
+        my $docker_version = get_docker_version();
+        systemctl('is-active firewalld');
+        my $running = script_output qq(docker ps -q | wc -l);
+        validate_script_output('ip a s docker0', sub { /state DOWN/ }) if $running == 0;
+        # Docker zone is created for docker version >= 20.10 (Tumbleweed), but it
+        # is backported to docker 19 for SLE15-SP3 and for Leap 15.3
+        if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {
+            assert_script_run "firewall-cmd --list-all --zone=docker";
+            validate_script_output "firewall-cmd --list-interfaces --zone=docker", sub { /docker0/ };
```
grisu48, line 56: I would also check here, if `docker0` is in the active firewall zones, e.g.

    validate_script_output "firewall-cmd --get-active-zones", sub { /docker.*interfaces: docker0/ };

or something similar. WDYT?

lib/containers/utils.pm
```diff
@@ -40,27 +40,59 @@ sub test_seccomp {
...
+        }
+        # Rules applied before DOCKER. Default is to listen to all tcp connections
+        # ex. output: "1           0        0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0"
+        validate_script_output "iptables -L DOCKER-USER -nvx --line-numbers", sub { /1.+all.+0\.0\.0\.0\/0\s+0\.0\.0\.0\/0/ };
```
grisu48, line 60: Ok for now, but this will break once `firewalld` uses netfilter instead of iptables.

tests/containers/docker.pm
```diff
@@ -47,8 +47,11 @@ sub run {
...
+        systemctl('restart docker');
+        $stop_firewall = 1;
+        check_containers_firewall($engine);
     }
```
grisu48, line 55: I'd add at least a `record_info` that no firewall test is performed. Swallowing tests silently is dangerous.

tests/containers/podman.pm
```diff
@@ -22,10 +22,14 @@
...
 use testapi;
 use utils;
+use version_utils;
 use registration;
 use containers::common;
 use containers::utils;
 use containers::container_images;
+use publiccloud::utils;
+
+my $stop_firewall = 0;    # Post-run flag to stop the firewall (failsafe)
```
grisu48, line 32: Nice! :slightly_smiling_face:

grisu48 (commented):
Good work! I left some comments and one major remark:

We are checking currently the functionality of the firewall by checking `iptables` rules. This is not optimal as we assume a certain underlying backend (iptables), which will probably break very soon. Instead I suggest to move the test design to a way, that we check, if the desired effect of the firewall is reached, e.g. is the container still can reach an external website and if the intra-container network is still working.

tests/containers/docker.pm
pdostal: This is done right on top of `check_containers_firewall()` - feel free if you disagree ;)

lib/containers/utils.pm
pdostal: Both `iptables` and `nftables` are using netfilter. I plan to test this PR also on Tw with `nftables`.

lib/containers/utils.pm
pdostal: You just proposed similar message in another of your suggestion.

lib/containers/utils.pm
grisu48: Ahhh, I see, perfect :+1:

lib/containers/utils.pm
grisu48: Please ignore this comment, I will have a look at this later on. This is beyond the scope of this PR.

lib/containers/utils.pm
pdostal: Please see my comment outside of this thread.

lib/containers/utils.pm
pdostal: I don't think so. This is still part of testing that the engine works. Feel free to reopen tho.

lib/containers/utils.pm
pdostal: Yes, I will comment on this outside of this thread.

pdostal:
I agree that `iptables` is slowly replaced by `nftables` but actually this is not the case for `podman` and `docker` yet. I tested this on SLE15-SP3 and Tumbleweed.  I suggest to keep the `iptables` validations in and them do the switch when upstream does.

pdostal:
1) I created the `lib/containers/operations_{docker,podman}.pm` as this subroutine does not directly touch the backend but rather test the firewall.
2) There was proposal to move this subrotine directly to `lib/containers/{docker,podman}.pm`. I'm not against it.
3) Last proposal is to create dedicated test module, e.g. `tests/containers/{docker,podman}_firewalled.pm` which is also easy to do.

Please vote! :smile:

pdostal:
The verification runs have been updated: [Docker@Tumbleweed](http://pdostal-server.suse.cz/tests/12803),[Podman@Tumbleweed](http://pdostal-server.suse.cz/tests/12804), [Docker@SLE15-SP3](http://pdostal-server.suse.cz/tests/12777), [Podman@SLE15-SP3](http://pdostal-server.suse.cz/tests/12779)

Please suggest me more verification runs!

lib/containers/operations_docker.pm
```diff
@@ -0,0 +1,50 @@
...
+    # Run container in the background
+    assert_script_run "docker run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine');
+    my $container_ip = container_ip($container_name, 'docker');
+
+    # Each running container should have added a new entry to the DOCKER zone.
+    # ex. output: "1           0        0 ACCEPT     tcp  --  !docker0 docker0  0.0.0.0/0            172.17.0.2           tcp dpt:1234"
+    validate_script_output "iptables -L DOCKER -nvx --line-numbers", sub { /1.+ACCEPT.+!docker0 docker0.+$container_ip\s+tcp dpt:1234/ };
+
+    # Connectivity to host check
+    my $default_route = script_output "docker run " . registry_url('alpine') . " ip route show default | awk \'/default/ {print \$3}\'";
```
grisu48, line 37: Clever!

lib/containers/operations_docker.pm
```diff
@@ -0,0 +1,50 @@
...
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outisde connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " wget " . get_var('OPENQA_URL');
+
+    # Kill the container running on background
+    assert_script_run "docker kill $container_name ";
```
grisu48, line 47: Since everywhere the `--rm` flag is set, I don't believe this is needed?
But if it is a safety cleanup, why not make it `script_run` instead?

lib/containers/operations_podman.pm
```diff
@@ -0,0 +1,45 @@
...
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outside connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " wget " . get_var('OPENQA_URL');
+
+    # Kill the container running on background
+    assert_script_run "podman kill $container_name";
```
grisu48, line 42: Same as for docker - See my other comment

grisu48 (approved):
LGTM! :rocket: Just some nitpicking, feel free to ignore if the comments are not applicable. Overall very good work.

lib/containers/operations_docker.pm
pdostal: No no, if you will go up, run run one container in the background. This needs to be done.

jlausuch:
I am late with this, sorry!
I would vote for option 2, we already have specifics for `docker|podman` in `lib/containers/docker|podman.pm`, let's just add it there and call it `$engine->check_firewall()`.

Also, I agree with @grisu48 about a higher level testing instead of going into the host specific configuration, which is much more delicate and error prone, and we could avoid things like
` if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {`
So, I would prefer a test case that verifies a use case from the user's prospective, not from the developer's prospective.

e.g. 
"as a user I would like to create a container and reach the outside" 
"as a user, I would like to create a container running a service on port XYZ and reach that service from outside"
etc.

Maybe for the next iteration?

pdostal:
> I am late with this, sorry! I would vote for option 2, we already have specifics for `docker|podman` in `lib/containers/docker|podman.pm`, let's just add it there and call it `$engine->check_firewall()`.

OK. I'll do that.
 
> Also, I agree with @grisu48 about a higher level testing instead of going into the host specific configuration, which is much more delicate and error prone, and we could avoid things like ` if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {` So, I would prefer a test case that verifies a use case from the user's prospective, not from the developer's prospective.
> 
> e.g. "as a user I would like to create a container and reach the outside" "as a user, I would like to create a container running a service on port XYZ and reach that service from outside" etc.
> 
> Maybe for the next iteration?

I already test connection *FROM* container.
I'll create a ticket to test connection *TO* container from outside.

pdostal:
Verification runs updated: [Docker@SLE15-SP3](http://pdostal-server.suse.cz/tests/12840), [Podman@SLE15-SP3](http://pdostal-server.suse.cz/tests/12840), [Docker@Tumbleweed](http://pdostal-server.suse.cz/tests/12837), [Podman@Tumbleweed](http://pdostal-server.suse.cz/tests/12841), [Containers@SLE15-SP3-EC2](http://pdostal-server.suse.cz/tests/12836)

This is from my opinion ready.

jlausuch (approved):

DimStar77:
https://openqa.opensuse.org/tests/2055360#step/podman/28

This - and others, sound very related

jlausuch:
We should add a check to see if firewalld exists..

DimStar77:
@sysrich looking at
https://openqa.opensuse.org/tests/2055802#step/podman/26

Is it intentional that we have no firewalld there? Or did this unhide a bug that we have?

jlausuch:
> https://openqa.opensuse.org/tests/2055360#step/podman/28
> 
> This - and others, sound very related
> 
> Seems to affect all container_host tests, also blocking stagings from turning green

BTW, SLE Micro passes this check. https://openqa.suse.de/tests/7739109#step/podman/29

sysrich:
This change is unadulterated nonsense

firewalld is not present on MicroOS or Kubic

It doesn't matter that SLE Micro passes the test, this nonsense has blocked all Tumbleweed development

I will revert it if I dont hear about a more suitable fix already being developed promptly

Vogtinator:
> This change is unadulterated nonsense
> 
> firewalld is not present on MicroOS or Kubic - the main consumers of podman in openSUSE
> 
> It doesn't matter that SLE Micro passes the test, this nonsense has blocked all Tumbleweed development
> 
> I will revert it if I dont hear about a more suitable fix already being developed promptly

Testing the combination of `podman` and `firewalld` is a good change. The only issue is that it broke if `firewalld` is not installed is simply a bug.

There's no need to react in such an unprofessional way.

PR #13667, opened 2021-11-11
Bump timeout for installer tests with LibyuiClient

OleksandrOrlov:
Tests failed on some systems when a lot of packages were installed,
because the installation takes more then 60 seconds.

The commit increases timeout to avoid that issue.

- Related ticket: https://progress.opensuse.org/issues/101957
- Verification run: https://openqa.suse.de/tests/7649112


jknphy:
I missed completely that we don't use a default timeout, we always set it, good that you found it.

PR #13668, opened 2021-11-11
Remove condition which create wrong scheduling on JeOS

b10n1k:
The main issue was the condition is `products/opensuse/main.pm` in L261
which was evaluated to *false* and then later another condition linked to *load_container_tests*.

The solution was to remove this condition, so letting *load_jeos_tests* run first and then
find *load_container_tests* which in turn will *load_jeos_containers* as it was before.

- Related ticket: https://progress.opensuse.org/issues/102272
- Verification run: 

https://openqa.opensuse.org/tests/2029918
https://openqa.opensuse.org/tests/2029917
https://openqa.opensuse.org/tests/2030191

https://openqa.suse.de/tests/7650121
https://openqa.suse.de/tests/7650120

VR for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/commits/366eb59b5758501a37278acc1dd6090ee502badd:
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2031116
Github oauth token provided, performing authenticated requests
Created job #2036430: opensuse-15.3-JeOS-for-kvm-and-xen-ppc64le-Build9.279-containers_docker_on_15.2_gm@ppc64le -> https://openqa.opensuse.org/t2036430
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2031115
Github oauth token provided, performing authenticated requests
Created job #2036431: opensuse-15.3-JeOS-for-kvm-and-xen-ppc64le-Build9.279-containers_podman_on_15.2_gm@ppc64le -> https://openqa.opensuse.org/t2036431
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2034865
Github oauth token provided, performing authenticated requests
Created job #2036432: opensuse-Tumbleweed-JeOS-for-kvm-and-xen-x86_64-Build20211113-jeos-containers@64bit_virtio -> https://openqa.opensuse.org/t2036432
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2034867
Github oauth token provided, performing authenticated requests
Created job #2036433: opensuse-Tumbleweed-JeOS-for-kvm-and-xen-x86_64-Build20211113-jeos@64bit_virtio -> https://openqa.opensuse.org/t2036433
❯ 
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2031132
Github oauth token provided, performing authenticated requests
Created job #2036434: opensuse-15.3-JeOS-for-AArch64-aarch64-Build9.279-jeos-containers@USBboot_aarch64 -> https://openqa.opensuse.org/t2036434
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2031116
Github oauth token provided, performing authenticated requests
Created job #2036436: opensuse-15.3-JeOS-for-kvm-and-xen-ppc64le-Build9.279-containers_docker_on_15.2_gm@ppc64le -> https://openqa.opensuse.org/t2036436
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.opensuse.org/tests/2031118
Github oauth token provided, performing authenticated requests
Created job #2036437: opensuse-15.3-JeOS-for-kvm-and-xen-x86_64-Build9.279-containers_on_centos_host_x86_64@64bit-2G -> https://openqa.opensuse.org/t2036437
❯ 
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.suse.de/tests/7675887
Github oauth token provided, performing authenticated requests
Created job #7678825: sle-15-SP3-JeOS-for-kvm-and-xen-Updates-x86_64-Build20211115-1-jeos-filesystem@uefi-virtio-vga -> https://openqa.suse.de/t7678825
❯  openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668/ https://openqa.suse.de/tests/7677181
Github oauth token provided, performing authenticated requests
Created job #7678827: sle-15-SP3-Server-DVD-Updates-x86_64-Build20211115-1-docker_tests@64bit -> https://openqa.suse.de/t7678827

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -554,6 +554,9 @@ sub load_system_role_tests {
     }
 }
 sub load_jeos_tests {
+    # o3 runs container jobs with jeos flavor which their schuduling is not related with this loader. Skip and let
+    # load_container_tests take care of those(lean on ppc64le and container tests on centos/ubuntu)
+    return if (get_var('CONTAINER_RUNTIME', 0) && (is_leap && is_ppc64le) || check_var('CONTAINERS_NO_SUSE_OS', 1));
     if ((is_arm || is_aarch64) && is_opensuse()) {
         # Enable jeos-firstboot, due to boo#1020019
         load_boot_tests();
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -40,6 +40,14 @@ sub is_ubuntu_host {
     return get_var("HDD_1") =~ /ubuntu/;
 }
 
+sub is_opensuse_jeos {
+    return (is_opensuse && is_jeos);
+}
+
+sub is_container_jeos {
+    return (get_var('CONTAINER_RUNTIME', 0) && ((is_leap && is_ppc64le) || check_var('CONTAINERS_NO_SUSE_OS', 1)));
+}
+
 sub load_image_tests_podman {
     loadtest 'containers/podman_image';
 }
@@ -84,23 +92,44 @@ sub load_host_tests_docker {
     loadtest "containers/container_diff" if (is_opensuse());
 }
 
+sub load_jeos_containers {
+    loadtest 'containers/podman';
+    loadtest 'containers/podman_image';
+    loadtest 'containers/podman_3rd_party_images';
+    loadtest 'containers/docker';
+    loadtest 'containers/docker_runc';
+    loadtest 'containers/docker_3rd_party_images';
+    loadtest 'containers/docker_image';
+    loadtest 'containers/docker_compose';
+    loadtest 'containers/registry' if is_x86_64;
+    loadtest 'containers/zypper_docker';
+    loadtest 'containers/rootless_podman';
+}
 
 sub load_container_tests {
     my $runtime = get_required_var('CONTAINER_RUNTIME');
     if (get_var('BOOT_HDD_IMAGE')) {
         loadtest 'installation/bootloader_zkvm' if is_s390x;
-        loadtest 'boot/boot_to_desktop';
+        loadtest 'boot/boot_to_desktop' if (is_container_jeos || !is_jeos);
     }
 
     if (is_container_image_test()) {
         # Container Image tests
-        loadtest 'containers/host_configuration' unless (is_res_host || is_ubuntu_host);
-        load_image_tests_podman() if ($runtime =~ 'podman');
-        load_image_tests_docker() if ($runtime =~ 'docker');
+        if (is_opensuse_jeos() && !is_container_jeos) {
+            load_jeos_containers();
+        } else {
+            loadtest 'containers/host_configuration' unless (is_res_host || is_ubuntu_host);
+            load_image_tests_podman() if ($runtime =~ 'podman');
+            load_image_tests_docker() if ($runtime =~ 'docker');
+        }
     } else {
         # Container Host tests
-        load_host_tests_podman() if ($runtime =~ 'podman');
-        load_host_tests_docker() if ($runtime =~ 'docker');
+        if (is_opensuse_jeos() && !is_container_jeos) {
+            load_jeos_containers();
+        } else {
+            load_host_tests_podman() if ($runtime =~ 'podman');
+            load_host_tests_docker() if ($runtime =~ 'docker');
+        }
     }
     loadtest 'console/coredump_collect';
 }
--- a/products/opensuse/main.pm
+++ b/products/opensuse/main.pm
@@ -258,7 +258,7 @@ sub load_default_tests {
 }
 
 # load the tests in the right order
-if (is_jeos && !is_container_test) {
+if (is_jeos) {
     load_jeos_tests();
 }
 


lib/main_containers.pm
```diff
@@ -84,23 +88,44 @@ sub load_host_tests_docker {
     loadtest "containers/container_diff" if (is_opensuse());
 }
 
+sub load_jeos_containers {
```
jlausuch, line 91: This is wrong. We already have a function loader for podman and docker tests. You are duplicating a lot of code here which is unnecessary.

b10n1k: As an urgent task i tried to keep the scheduling clean and isolated. in other case i would have many moving parts. Like many "if else". ofc i would like to avoid and improve this part, which appears twice 

```
if (is_opensuse_jeos()) {
            load_jeos_containers();
        } else
```

but it was the faster approach i came up initially. but it could happen later.
I also think that the *load_jeos_containers* is a good idea for now to not interfere, as jeos is kinda *special* case, and doing this so, we decouple the scheduling from sle.

products/opensuse/main.pm
```diff
@@ -258,7 +258,7 @@ sub load_default_tests {
 }
 
 # load the tests in the right order
-if (is_jeos && !is_container_test) {
+if (is_jeos) {
```
mloviska, line 261: This would break the `containers_*` tests that are actually not JeOS but have flavor `JeOS-for-kvm-and-xen`

jlausuch: my proposal is actually only having this change :)

jlausuch:
This is my proposal I had in mind before you came up with this PR. Sorry if I didn't present it before, I hope it's fine.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13672

products/opensuse/main.pm
b10n1k: @mloviska if this is one of the test cases that you had in mind, I run a VR. i cant see how the implementation affects it.

mloviska: Please compare https://openqa.opensuse.org/tests/2031118# vs https://openqa.opensuse.org/tests/2031324#

b10n1k: i did an effort to fix this with a dirty approach. I dont know. maybe @jlausuch [suggestion](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13672#issuecomment-968338314) can clean this mess.

b10n1k: *this change is ready for review* Still i can remove the duplication of the new block but feel free to go on(because the urgency) if you want and we will fix in another PR if it is needed.

jlausuch:
I would really like to fix the issue from the root, which is the openqa job setup,  not having a hacky code to handle any weird setups (e.g. booting centos with JeOS flavor). 
So, even though it's urgent, let's fix this in the right way, please. I proposed to have a new flavor for these cases where we boot "other" host images.

lib/main_containers.pm
```diff
@@ -84,23 +92,44 @@ sub load_host_tests_docker {
...
+    loadtest 'containers/zypper_docker';
+    loadtest 'containers/rootless_podman';
+}
 
 sub load_container_tests {
     my $runtime = get_required_var('CONTAINER_RUNTIME');
     if (get_var('BOOT_HDD_IMAGE')) {
         loadtest 'installation/bootloader_zkvm' if is_s390x;
-        loadtest 'boot/boot_to_desktop';
+        loadtest 'boot/boot_to_desktop' if (is_container_jeos || !is_jeos);
```
grisu48, line 113: By default we always run `boot/boot_to_desktop`, so this logic is flawed to fail in the future.
The default should always be performed `unless` the exception and not the other way around. So instead of asking "right now, boot_to_desktop needs to be executed if ...." one should approach it as "boot_to_desktop should always be executed, except for the special case, that ....

TL'DR: Please change it to something like
```suggestion
        loadtest 'boot/boot_to_desktop' unless ... ;
```

b10n1k:
> I would really like to fix the issue from the root, which is the openqa job setup, not having a hacky code to handle any weird setups (e.g. booting centos with JeOS flavor). So, even though it's urgent, let's fix this in the right way, please. I proposed to have a new flavor for these cases where we boot "other" host images.

i will unassign myself from this ticket then. i have no idea what is the right way. But i know that this would be a much bigger change, so it is up to you.

grisu48 (commented):
For a fix this is too complex and we are applying a fix plus some additional cleanup at the same time. If it is an urgent fix, then please only fix the issue and do the cleanup afterwards.

lib/main_common.pm
```diff
@@ -554,6 +554,7 @@ sub load_system_role_tests {
     }
 }
 sub load_jeos_tests {
+    return if (get_var('CONTAINER_RUNTIME', 0) && (is_leap && is_ppc64le) || check_var('CONTAINERS_NO_SUSE_OS', 1));
```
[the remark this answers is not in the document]
b10n1k: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668#discussion_r748047477

lib/main_containers.pm
b10n1k: but remains the default.  i tried with unless but there were tests failing. i would like to leave it as it is, because this is getting complex enough and i think it is going to be a better approach soon after.

lib/main_common.pm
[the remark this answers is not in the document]
pdostal: Yianni, please add `# a comment` :smile:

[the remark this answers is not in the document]
b10n1k: *added*

lib/main_containers.pm
grisu48: It does not remain the default, because it is only executed if a certain condition is fulfilled.
The opposite is that it is executed by default, unless a condition is fulfilled.

The difference is opt-in by default or opt-out by default.

Why is this important? Because if we add another test e.g. for SLE Micro, that one would not execute `boot/boot_to_desktop` unless we add that routine to the `if` loop.
If we have a exclusion logic (`unless`) then every new test runs, however weird they might be, will use this by default unless specified otherwise.

b10n1k:
deprecated from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13746/

PR #13672, opened 2021-11-12
Containers: fix schedule for JeOS

jlausuch:
This fixes:
https://progress.opensuse.org/issues/102272
https://progress.opensuse.org/issues/102278

This basically enables running the default jeos tests as it was done before, and then run the specific container tests.

In a follow-up PR, we could come up with a more smart default jeos test for container jobs only, instead of running all the default ones, but this is a different topic than this PR tries to fix.

VRs:
- [USBboot_aarch64 ](https://openqa.opensuse.org/tests/2030307)
- [x86_64](https://openqa.opensuse.org/tests/2031072)  (not sure why validate_btrfs fails)

**Some things need to be done in job configuration:**
- Add `NUMDISKS=2` for `validate_btrfs` test
- Remove `BOOT_HDD_IMAGE` -> JeOS tests don't need that variable.
- Remove `CONTAINERS_UNTESTED_IMAGES` -> this is a HOST test based to test container tools and packages on JeOS. We already have a job to validate the "untested" container image: https://openqa.opensuse.org/tests/2029146. If we want to enable also container image validation on JeOS, we should create a separate job.

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -80,7 +80,7 @@ sub load_host_tests_docker {
         loadtest 'containers/registry' if is_x86_64;
         loadtest 'containers/docker_compose';
     }
-    loadtest 'containers/validate_btrfs' if is_x86_64;
+    loadtest 'containers/validate_btrfs' if (is_x86_64 && !is_jeos);
     loadtest "containers/container_diff" if (is_opensuse());
 }
 
--- a/products/opensuse/main.pm
+++ b/products/opensuse/main.pm
@@ -258,7 +258,7 @@ sub load_default_tests {
 }
 
 # load the tests in the right order
-if (is_jeos && !is_container_test) {
+if (is_jeos) {
     load_jeos_tests();
 }
 


jlausuch:
Please be aware of this proposal: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13668
But I would really prefer to re-use code and not duplicating loadtest calls.

b10n1k:
> Please be aware of this proposal: #13668 But I would really prefer to re-use code and not duplicating loadtest calls.

i am not sure any more. i have a few observations. they are minor atm because i think it is more important to solve the issue. So in theory we can just merge this and discuss further later. However still cant decide as this PR seems to require additional commit for the variables you mentioned. ping me whenever you want

jlausuch:
> > Please be aware of this proposal: #13668 But I would really prefer to re-use code and not duplicating loadtest calls.
> 
> i am not sure any more. i have a few observations. they are minor atm because i think it is more important to solve the issue. So in theory we can just merge this and discuss further later. However still cant decide as this PR seems to require additional commit for the variables you mentioned. ping me whenever you want

To fix the scheduling issue, no new PRs are needed. The variables will be set in the UI. 

Can you have a look at validate_btrfs test? Any ideas why it fails?

jlausuch:
VR to make sure default JeOS schedule is not modified: 
https://openqa.opensuse.org/tests/2031338

pdostal:
Related to: #13645

pdostal (approved):

jlausuch:
Ok, wait. This change breaks these tests:
https://openqa.opensuse.org/tests/1995349#

But I would say that the job setup for this test is completely wrong. It's booting a CentOS HDD using JeOS flavor... I guess we need to change that, create a new flavor or re-use existing one...

jlausuch:
IMO, we should use [this test](https://openqa.opensuse.org/tests/2031047) to run the container tests after hdd creation and maybe place them in [Leap 15.3 Updates job](https://openqa.opensuse.org/group_overview/80) instead of Leap 15.3 images.

b10n1k:
> > > Please be aware of this proposal: #13668 But I would really prefer to re-use code and not duplicating loadtest calls.
> > 
> > 
> > i am not sure any more. i have a few observations. they are minor atm because i think it is more important to solve the issue. So in theory we can just merge this and discuss further later. However still cant decide as this PR seems to require additional commit for the variables you mentioned. ping me whenever you want
> 
> To fix the scheduling issue, no new PRs are needed. The variables will be set in the UI.
> 
> Can you have a look at validate_btrfs test? Any ideas why it fails?

`validate_btrfs` was not part of the scheduling there. the reason is likely the unmatched size which this module is quite sensitive.

lib/main_containers.pm
```diff
@@ -80,7 +80,7 @@ sub load_host_tests_docker {
         loadtest 'containers/registry' if is_x86_64;
         loadtest 'containers/docker_compose';
     }
-    loadtest 'containers/validate_btrfs' if is_x86_64;
+    loadtest 'containers/validate_btrfs' if (is_x86_64 && !is_jeos);
```
jlausuch, line 83: This module is not executed in SLE JeOS either, so I created this ticket to enable it later on.
https://progress.opensuse.org/issues/102377

jlausuch:
My idea is to reorder the O3 test like this:

Screenshot_20211114_190246

http://fromm.arch.suse.de/tests/overview?distri=opensuse&version=15.3&build=9.279&groupid=50
http://fromm.arch.suse.de/tests/overview?distri=opensuse&version=15.2&build=31.593&groupid=51

This would also need like this:
https://github.com/os-autoinst/openqa-trigger-from-obs/pull/145

jlausuch:
@Vogtinator @DimStar77 Would you be ok with my proposal to group container image tests so everything looks more tidy?

Vogtinator:
> @Vogtinator @DimStar77 Would you be ok with my proposal to group container image tests so everything looks more tidy?

Yes! That was pretty much the goal from the beginning, but implementing that in openqa-trigger-from-obs isn't trivial (See discussion on https://github.com/os-autoinst/openqa-trigger-from-obs/pull/145), particularly because of `BUILD` number differences.

It would be possible to decouple container images into a separate project, but that is arguably overkill and makes it harder (if not practically impossible) to use latest JeOS as host, which is simpler and quicker than using a GM image + installing updates.

jlausuch:
The following ticket has been resolved: https://progress.opensuse.org/issues/97958
So, we have now an ISOS POST with the flavor `Container-Image` to separate the container tests that boot some HDD (CentOS, Ubuntu, etc) from JeOS flavor, so we don't have that dependency any more. 
So, now, `JeOS` flavor is ONLY used for JeOS tests, and `Container-Image` flavor for image validation. 

This is ready to be merged.

PR #13649, opened 2021-11-09
Independize vault client from the publiccloud::provider

ilausuch:
Now vault client actions are part of the provider class
this forces to use code of instance creation (e.g. terraform)
in tests where are not necessary create instances (e.g. kubernetes)

- Related ticket: https://progress.opensuse.org/issues/101491
- Verification runs:
- http://copland.arch.suse.de/tests/512
- http://copland.arch.suse.de/tests/511
- (Google) http://openqa.suse.de/tests/7648540
- (Amazon) http://openqa.suse.de/t7654342
- (Azure) https://openqa.suse.de/tests/7654359

PR #13399, opened 2021-10-06
Create Public Cloud CLI tools test modules

pdostal:
Currently I create *Azure*, *Google Cloud Engine* and *AWS EC2* VMs using corresponding CLI tools.
1) I Install the tools
2) Create the VM
3) Remove the VM

So far for *Azure* and *Google* I use upstream binaries as our binaries seems to be broekn ([azure](http://pdostal-server.suse.cz/tests/12557#step/azure_cli/87), [google](http://pdostal-server.suse.cz/tests/12557#step/google_cli/25)).

- Related ticket: [poo#93097](https://progress.opensuse.org/issues/93097)
- Confluence: [Create PC test instance](https://confluence.suse.com/display/qasle/Create+PC+test+instance)
- Verification run: http://pdostal-server.suse.cz/tests/12559

PR #13680, opened 2021-11-15
Public Cloud CLI: Small fixes

pdostal:
- Related ticket: [poo#93097](https://progress.opensuse.org/issues/93097)
- Verification run: [public_cloud image](http://pdostal-server.suse.cz/tests/12882) [maintenance_update image](http://pdostal-server.suse.cz/tests/12877)


pdostal:
I was inspired by #13649 and #13399.

schedule/publiccloud/publiccloud_cli.yml
```diff
@@ -7,5 +7,5 @@ schedule:
   - publiccloud/azure_cli
   - publiccloud/aws_cli
   # We currently don't package gcloud binary
-  # - publiccloud/google_cli
+  - publiccloud/google_cli
```
asmorodskyi, line 10: can you elaborate why you think we should change this ? ( keeping in mind all discussions around gcloud in previous PR )

asmorodskyi:
> I was inspired by #13649 and #13399 smile

and what about #13534 ? shall we take it in consideration here too ?

schedule/publiccloud/publiccloud_cli.yml
pdostal: I think that we should at least keep checking if nothing breaks from our side.
Even if the package is not our it can interfere with something else.

asmorodskyi: ok

jlausuch (approved):

ilausuch (approved):

tests/publiccloud/google_cli.pm
```diff
@@ -30,7 +31,7 @@ sub run {
         assert_script_run "curl $url -o google-cloud-sdk.tar.gz";
         assert_script_run "tar xvf google-cloud-sdk.tar.gz";
         assert_script_run "google-cloud-sdk/install.sh --quiet --usage-reporting false --command-completion true";
-        assert_script_run "source ~/.bashrc";
+        assert_script_run "source ~/google-cloud-sdk/path.bash.inc";
```
grisu48, line 34: Why include the specific include instead of the generic bashrc? This invalidates the test, if the SDK is installed properly and setup in a way, that a new `bash` shell would work with it properly?

pdostal: No. I checked this and only command completion are put into `~/.bashrc` but the `gcloud` binary is still not visible under `$PATH`.

PR #13681, opened 2021-11-15
Replace pipe with process substitution in crm_mon

lpalovsky:
jsc#TEAM-5180 In lib/hacluster.pm  test check_after_reboot fails 
because of 'crm_mon' output is being piped to 'grep -g'. This results 
in 141 RC being returned even if a match was found. This PR changes 
pipe to process substitution. 

- Related ticket: https://jira.suse.com/browse/TEAM-5180
- Verification run: https://mordor.suse.cz/tests/1659#step/check_after_reboot/17

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -696,8 +696,8 @@ sub wait_until_resources_started {
     my $ret = undef;
 
     # Some CRM options can only been added on recent versions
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM

Please rebase to avoid the `Merge branch 'os-autoinst:master' into crm_mon-grep-fix` commit.

PR #13682, opened 2021-11-15
Containers: Improve stability of 'container run zypper in curl' with …

pdostal:
- Related ticket: [poo#102422](https://progress.opensuse.org/issues/102422)
- Verification run: TBD

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -151,7 +151,7 @@ sub basic_container_tests {
     die("error: missing container $container_name") unless ($output_containers =~ m/$container_name/);
 
     # Containers' state can be saved to a docker image
-    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
+    my $ret = script_retry("$runtime container exec $container_name zypper -n in curl", retry => 12, delay => 15, timeout => 600, die => 0);
     die('zypper inside container timed out') if (!defined($ret));
     if ($ret != 0) {
         record_info('poo#40958 - curl install failure, try with force-resolution.');


lib/containers/utils.pm
```diff
@@ -151,7 +151,7 @@ sub basic_container_tests {
     die("error: missing container $container_name") unless ($output_containers =~ m/$container_name/);
 
     # Containers' state can be saved to a docker image
-    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
+    my $ret = script_retry("$runtime container exec $container_name zypper -n in curl", retry => 12, delay => 15, timeout => 600, die => 0);
```
jlausuch, line 154: Why 12 retries? I have quickly checked all the places where we do retries, and I think 3 times is already a good compromise.

pdostal:
I think that we need to re-open [bsc#1192436](https://bugzilla.suse.com/show_bug.cgi?id=1192436). May I do that @jlausuch?

This is probably not good solution, see the [test run](http://pdostal-server.suse.cz/tests/12742#step/docker/205) or:
```
# docker container run -d --name tw registry.opensuse.org/opensuse/tumbleweed tail -f /dev/null
7aed81540d58d0266812e06f3fc16de2c562f771523091ea35a77754ef800f09
# docker container ls -a
CONTAINER ID   IMAGE                                       COMMAND                  CREATED                  STATUS                     PORTS     NAMES
7aed81540d58   registry.opensuse.org/opensuse/tumbleweed   "tail -f /dev/null"      Less than a second ago   Up Less than a second                tw
# timeout 600 docker container exec tw zypper -n in curl
Retrieving repository 'openSUSE-Tumbleweed-Non-Oss' metadata [.done]
Building repository 'openSUSE-Tumbleweed-Non-Oss' cache [....done]
Retrieving repository 'openSUSE-Tumbleweed-Oss' metadata [...done]
Building repository 'openSUSE-Tumbleweed-Oss' cache [....done]
Retrieving repository 'openSUSE-Tumbleweed-Update' metadata [.done]
Building repository 'openSUSE-Tumbleweed-Update' cache [....done]
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following NEW package is going to be installed:
  curl

1 new package to install.
Overall download size: 414.1 KiB. Already cached: 0 B. After the operation, additional 721.4 KiB will be used.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package curl-7.79.1-1.3.x86_64 (1/1), 414.1 KiB (721.4 KiB unpacked)
Retrieving: curl-7.79.1-1.3.x86_64.rpm [done]

Checking for file conflicts: [......done]
(1/1) Installing: curl-7.79.1-1.3.x86_64 [.
# timeout 600 docker container exec tw zypper -n in curl
System management is locked by the application with pid 8 (zypper).
Close this application before trying again.
```

jlausuch:
Yes, it doesn't look good. This is because script_run doesn't kill the process, so zypper process keeps running in the background and we trigger another "try" of the zypper command, therefore it complains with `System management is locked by the application with pid 8 (zypper).` 
I guess you can re-open that bug but I'd like to ask @Vogtinator for some feedback too.

Vogtinator:
> I think that we need to re-open bsc#1192436. May I do that @jlausuch?

The fixed libzypp is not in a released Tumbleweed snapshot yet.

Vogtinator (changes requested):
If the command fails, the test should rightfully fail instead of retrying silently.

jlausuch:
ok, let's then close this PR. Thanks @Vogtinator!

lib/containers/utils.pm
grisu48: Yes. I'd increase the delay however then
```suggestion
    my $ret = script_retry("$runtime container exec $container_name zypper -n in curl", retry => 3, delay => 60, timeout => 600, die => 0);
```

PR #13683, opened 2021-11-15
Add check for kernel/kernel-devel version mismatch

BillAnastasiadis:
Checks for kernel-default-devel installation and matches it to the kernel-default installation. The appropriate message is displayed in each case, should any of those checks fail.

- Related ticket: https://progress.opensuse.org/issues/101517
- Verification runs: 

TEST FAILURE RUNS: (test slightly altered to fail) (15sp2)
diffrent kernel-default version: http://10.161.229.209/tests/332#step/systemtap/374
no kernel-default-devel: http://10.161.229.209/tests/333#step/systemtap/371

NORMAL RUNS:
15GA: 
x86_64: https://openqa.suse.de/tests/7694449#
s390: https://openqa.suse.de/tests/7694450#
aarch64: FAILED IN PREVIOUS MODULE (mdadm) https://openqa.suse.de/tests/7694451
15sp1: 
x86_64: https://openqa.suse.de/tests/7694373#
s390: https://openqa.suse.de/tests/7694429#
aarch64: https://openqa.suse.de/tests/7694448#
15sp2: 
x86_64: https://openqa.suse.de/tests/7694291#
15sp3: 
x86_64: https://openqa.suse.de/tests/7694290#
12sp5: 
x86_64: https://openqa.suse.de/tests/7694452#
s390: https://openqa.suse.de/tests/7694453#
aarch64: https://openqa.suse.de/tests/7694454#
12sp4:
x86_64: https://openqa.suse.de/tests/7694456
s390: https://openqa.suse.de/tests/7694454#
12sp3: http://10.161.229.209/tests/309
x86_64: https://openqa.suse.de/tests/7694458#
12sp2: http://10.161.229.209/tests/310
x86_64: https://openqa.suse.de/tests/7694459#

Test is disabled on opensuse - there is another ticket to fix and enable it.

### DIFF
--- a/tests/console/systemtap.pm
+++ b/tests/console/systemtap.pm
@@ -22,6 +22,17 @@ sub run {
     $self->select_serial_terminal;
     prepare_for_kdump();
     zypper_call("in systemtap systemtap-docs kernel-devel systemtap-server");
+    script_run('ti=$(ls /lib/modules/ | grep $(uname -r) | grep -oP ".*(?=-default)") && zypper se -i -s kernel-default-devel | grep $ti > vertmp');
+    my $vers = script_output('cat vertmp');
+    if (!$vers) {
+        my $kernel_d = script_output('uname -r | grep -oP ".*(?=-default)"');
+        my $dev = zypper_call("se -i -s kernel-default-devel", exitcode => [0, 104]);
+        if ($dev ne '104') {
+            die "Installed kernel-devel does not match kernel version.\nExpected: $kernel_d, Found: Other kernel-default-devel versions";
+        } else {
+            die "kernel-devel package is required but not installed\nExpected: $kernel_d, Found: No kernel-default-devel package installed";
+        }
+    }
     assert_script_run("stap /usr/share/doc/packages/systemtap/examples/general/helloworld.stp | grep 'hello world'", 200);
     assert_script_run("stap-server condrestart | grep --line-buffered \"managed\"");
     assert_script_run("stap -v -e 'probe vfs.read {printf(\"read performed\\n\"); exit()}'");


tests/console/systemtap.pm
```diff
@@ -22,6 +22,16 @@ sub run {
...
     prepare_for_kdump();
     zypper_call("in systemtap systemtap-docs kernel-devel systemtap-server");
+    script_run('ti=$(ls /lib/modules/ | grep $(uname -r) | grep -oP ".*(?=-default)") && zypper se -i -s kernel-default-devel | grep $ti > vertmp');
+    my $vers = script_output('cat vertmp');
+    if (!$vers) {
+        my $dev = zypper_call("se -i -s kernel-default-devel", exitcode => [0, 104]);
+        if ($dev ne '104') {
+            die "Installed kernel-devel does not match kernel version";
+        } else {
+            die "kernel-devel package is required but not installed";
```
foursixnine, line 32: can you add found and expected to the die message?

BillAnastasiadis: No, good idea - going to add that now. Thanks!

punkioudi (commented):
In general lgtm, just a small comment in the VRs. It's better to run all the mau-extratests2 job, in order to check that the change you've made in systemtap, doesn't cause any issue to the other tests.

czerw:
I think we already have test for it https://openqa.suse.de/tests/7657091/modules/module_build/steps/1/src . It verifies that devel files are really for particular kernel by compilation.

tjyrinki (approved):
I think this would be a fine addition, given that dzedro ran into the problem in the first place. It's a bit different and self-explaining approach to the test as the kernel test is assuming a compilation would fail with different version while this checks actual packages.

czerw (approved):

tests/console/systemtap.pm
```diff
@@ -22,6 +22,17 @@ sub run {
     $self->select_serial_terminal;
     prepare_for_kdump();
     zypper_call("in systemtap systemtap-docs kernel-devel systemtap-server");
```
mdoucha, line 24: Wouldn't it be better to simply specify the kernel-devel version explicitly here? That's what `update_kernel` module does for kernel incident tests.

PR #13686, opened 2021-11-15
Containers: basetest: Do not rollback on Public Cloud

pdostal:
As on Public Cloud we use SSH tunnel we should not rollback the QEMU worker where this tunnel runs.

- Verification run: TBD

### DIFF
--- a/lib/containers/basetest.pm
+++ b/lib/containers/basetest.pm
@@ -10,6 +10,7 @@ package containers::basetest;
 use containers::docker;
 use containers::podman;
 use Mojo::Base 'opensusebasetest';
+use testapi 'get_var';
 
 sub containers_factory {
     my ($self, $runtime) = @_;
@@ -26,6 +27,10 @@ sub containers_factory {
     }
     $engine->init();
     return $engine;
+}
+
+sub test_flags {
+    return get_var('PUBLIC_CLOUD') ? {no_rollback => 1} : {};
 }
 
 1;


asmorodskyi (changes requested):
This PR uncover me something which raised several severe concerns in my head and questions .  It is not about this certain PR because it just repeats what we already have in similar places but generally about approach which we took here . Let's have discussion about this before merging it

grisu48 (approved):
LGTM

lib/containers/basetest.pm
```diff
@@ -28,4 +29,8 @@ sub containers_factory {
     return $engine;
 }
 
+sub test_flags {
+    return get_var('PUBLIC_CLOUD') ? {no_rollback => 1} : {};
+}
+
```
b10n1k, line 35: i wonder if this makes any difference. the `Base` defines the same `test_flags`. isnt this taken by parent class?

pdostal: You're right. We're missing the `QEMU_DISABLE_SNAPSHOTS` variable.

pdostal:
@b10n1k is right. We're missing the `QEMU_DISABLE_SNAPSHOTS` variable. The test flags are inherited from `opensusebasetest`.

jlausuch:
> @b10n1k is right. We're missing the `QEMU_DISABLE_SNAPSHOTS` variable. The test flags are inherited from `opensusebasetest`.

What do you mean "missing"?
The job that stop the whole execution has that variable: https://openqa.suse.de/tests/7676119#settings
whereas the job that continues with the other modules doesn't have the var: https://openqa.suse.de/tests/7655381#step/docker/237

pdostal:
Should I open a bug? This is in my opinion a bug in the openQA.

asmorodskyi:
> Should I open a bug? This is in my opinion a bug in the openQA.

yes please

pdostal:
[poo#102539](https://progress.opensuse.org/issues/102539)

PR #13696, opened 2021-11-16
Disable junit report in qa_automation on SLE 15-SP4 and newer

czerw:
Fix poo#102506: SLE 15-SP4 doesn't have python2 module. Junit report
script is python2 only. It was decided that old qa_automation scripts
will not be used and will be replaced in the future. It means that
script migration to python3 is not worth. We can still use
qa_automation scripts, but without human friendly openQA presentation.

- Related ticket: https://progress.opensuse.org/issues/102506
- Needles: none
- Verification run: 
15-SP4 (without report): https://openqa.suse.de/tests/7690030
15-SP3 (with report): https://openqa.suse.de/t7690046

### DIFF
--- a/tests/qa_automation/qa_run.pm
+++ b/tests/qa_automation/qa_run.pm
@@ -95,8 +95,9 @@ sub prepare_repos {
 
     add_qa_head_repo;
     add_qa_web_repo;
-    add_suseconnect_product('sle-module-python2') if is_sle('>15') && get_var('FLAVOR') !~ /-Updates$|-Incidents/;
-    zypper_call("in qa_testset_automation qa_tools python-base python-xml");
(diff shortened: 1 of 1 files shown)


czerw:
@dzedro @tjyrinki @okurz could you please check?

dzedro (approved):
I don't mind, the old qa_run should/will be dropped anyway.

okurz (approved):
It seems I can't convince people to not reinvent the wheel so, sure, go ahead and reimplement everything as flaky top-level openQA tests when tests for packages already exists for years :)

czerw:
> It seems I can't convince people to not reinvent the wheel so, sure, go ahead and reimplement everything as flaky top-level openQA tests when tests for packages already exists for years

So do you suggest to migrate junit gen to py3? It still can be done and this PR can be treated as workaround and can be merged, i will just update commit message.

okurz:
As I said I can't convince people so either way is ok. I just see it as a problem that many QA engineers have become so openQA centric that they completely lost the view of what's going on elsewhere and what is tested where already

PR #13697, opened 2021-11-16
Merge commmon loadtests in pubcloud scheduling

b10n1k:
`load_latest_publiccloud_tests` and `load_maintenance_publiccloud_tests` shares some scheduling logic.
I thought that it will make more clear if this is isolated and apply it in the subroutines without duplicate
those lines.

- Related ticket: https://progress.opensuse.org/issues/101427
- Verification run: https://openqa.suse.de/tests/overview?distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2313697&version=15-SP3

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -39,13 +39,8 @@ sub load_maintenance_publiccloud_tests {
     loadtest "publiccloud/register_system", run_args => $args;
     loadtest "publiccloud/transfer_repos", run_args => $args;
     loadtest "publiccloud/patch_and_reboot", run_args => $args;
-    if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
-        loadtest("publiccloud/img_proof", run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_LTP')) {
-        loadtest('publiccloud/run_ltp', run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
-        loadtest('publiccloud/storage_perf', run_args => $args);
-    } else {
+    my $is_other_scheduled = load_shared_publiccloud_tests();
+    unless ($is_other_scheduled != -1) {
         loadtest "publiccloud/ssh_interactive_start", run_args => $args;
         loadtest "publiccloud/instance_overview" unless get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS');
         if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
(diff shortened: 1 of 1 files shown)


lib/main_publiccloud.pm
```diff
@@ -39,13 +39,8 @@ sub load_maintenance_publiccloud_tests {
...
     loadtest "publiccloud/patch_and_reboot", run_args => $args;
-    if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
-        loadtest("publiccloud/img_proof", run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_LTP')) {
-        loadtest('publiccloud/run_ltp', run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
-        loadtest('publiccloud/storage_perf', run_args => $args);
-    } else {
+    my $is_other_scheduled = load_shared_publiccloud_tests();
+    unless ($is_other_scheduled != -1) {
```
jlausuch, line 43:
```suggestion
    unless (load_shared_publiccloud_tests()) {
```

lib/main_publiccloud.pm
```diff
@@ -77,47 +72,53 @@ sub load_publiccloud_consoletests {
...
-        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
-    }
-    elsif (get_var('PUBLIC_CLOUD_CONTAINERS')) {
-        loadtest "publiccloud/prepare_instance", run_args => $args;
-        loadtest "publiccloud/register_system", run_args => $args;
-        loadtest "publiccloud/ssh_interactive_start", run_args => $args;
-        load_podman_tests();
-        load_docker_tests();
-        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
+    return -1 if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS') || get_var('PUBLIC_CLOUD_CONTAINERS'));
```
jlausuch, line 77: why not just `0`? Normally, perl returns 1 if function is executed without errors. Just use `0`, and then you can do
`unless (load_shared_publiccloud_tests())`

lib/main_publiccloud.pm
```diff
@@ -77,47 +72,53 @@ sub load_publiccloud_consoletests {
...
+    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
+        loadtest('publiccloud/storage_perf', run_args => $args);
     }
-    elsif (get_var('PUBLIC_CLOUD_UPLOAD_IMG')) {
-        loadtest "publiccloud/upload_image";
+}
+
+sub load_latest_publiccloud_tests {
+    my $args = OpenQA::Test::RunArgs->new();
+    my $is_other_scheduled = load_shared_publiccloud_tests();
```
grisu48, line 89: I'd pass the $args argument to `load_shared_publiccloud_tests` so that we don't end up in using one $args instance for a couple of tests and other one for other tests.
```suggestion
    my $is_other_scheduled = load_shared_publiccloud_tests($args);
```

grisu48: This is present in multiple locations. We must only use one `$args` instance throughout the whole scheduling.

asmorodskyi (changes requested):
Generally I must say that it does not looks better than what we have already. And it is not about how actually you did it. It is just that logic is tricky so there is no better way than duplicate code. 
I vote for keep duplication and close PR .

lib/main_publiccloud.pm
b10n1k: The reason is that i needed another value to distinguish  *PUBLIC_CLOUD_CONSOLE_TESTS* / *PUBLIC_CLOUD_CONTAINERS* and the rest (at least in the current sequence) . 

`unless (load_shared_publiccloud_tests())` behaves as such (with `return 0 if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS') || get_var('PUBLIC_CLOUD_CONTAINERS'));`):
- consoletests - condition returns 0 and has correct scheduling
-  ltp                - calls ` loadtest('publiccloud/run_ltp)` and returns likely non-zero value and the scheduling adds ` publiccloud/ssh_interactive_start`, `publiccloud/instance_overview` and `publiccloud/ssh_interactive_end`. 

if you want to know why it is so, take a look on the [documentation](https://perldoc.perl.org/perlsub) where it writes `If you specify no return value, the subroutine returns an empty list in list context, the undefined value in scalar context, or nothing in void context.`

this is one case. ofc i tried other things too. i found the solution i present quick simple. in any case i would have to answer  questions like why i return 1 or zero, if or unless, and etc. i use `-1` as a *no-success* value which it doesnt conflict with the `true`/`false` definitions in perl, which is used to instruct a specific action when it matches.

lib/main_publiccloud.pm
b10n1k: i had same thought. thanks

b10n1k:
@asmorodskyi no problem from my side. Lets wait for others to leave their opinion and if the verdict is the _kill penalty_ lets kill it :)

lib/main_publiccloud.pm
```diff
@@ -77,47 +72,53 @@ sub load_publiccloud_consoletests {
...
+        loadtest('publiccloud/run_ltp', run_args => $args);
+    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
+        loadtest('publiccloud/storage_perf', run_args => $args);
     }
-    elsif (get_var('PUBLIC_CLOUD_UPLOAD_IMG')) {
-        loadtest "publiccloud/upload_image";
+}
+
+sub load_latest_publiccloud_tests {
+    my $args = OpenQA::Test::RunArgs->new();
```
asmorodskyi, line 88: this is not needed at all . It is mistake which passed by previous reviews . this tests don't need runargs at all . runargs needed only by tests which using ssh console and this tests are not using it.

lib/main_publiccloud.pm
```diff
@@ -77,47 +72,53 @@ sub load_publiccloud_consoletests {
...
+            loadtest "publiccloud/prepare_tools";
+        }
+        elsif (get_var('PUBLIC_CLOUD_SLES4SAP')) {
+            loadtest 'publiccloud/sles4sap';
+        }
+        elsif (get_var('PUBLIC_CLOUD_ACCNET')) {
+            loadtest 'publiccloud/az_accelerated_net';
+        }
+        elsif (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME')) {
+            loadtest "publiccloud/boottime", run_args => $args;
```
asmorodskyi, line 101: this test does not need runargs

b10n1k:
> @asmorodskyi no problem from my side. Lets wait for others to leave their opinion and if the verdict is the _kill penalty_ lets kill it :)

So should i close this?

b10n1k:
closing as per https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13697#pullrequestreview-807620502

PR #13699, opened 2021-11-16
Split image testing related classes from k8s containers ones (AWS)

ilausuch:
There is not much in common between classes related to interaction
with images ( ec2/gce/azure ) and classes related to interact with
k8s containers ( ecr/eks/gcr )

This commit splits AWS classes

- Related ticket: https://progress.opensuse.org/issues/101497
- Verification run:
-- https://openqa.suse.de/tests/7717881 (EC2)
-- http://copland.arch.suse.de/tests/572 (EKS)
-- http://copland.arch.suse.de/tests/573 (GKE)
-- https://openqa.suse.de/tests/7728639 (Azure)
-- https://openqa.suse.de/tests/7728640(Google)


lib/publiccloud/ec2.pm
```diff
@@ -28,7 +41,7 @@ sub find_img {
 sub create_keypair {
```
asmorodskyi, line 41: I think we need to do one of two : 
1. move `create_keypair` into aws_client
2. keep ssh related things fully in ec2 

currently it felt like something solid split into two pieces ...

jlausuch: What other ssh related things are outside `ec2.pm`?

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Helper class for amazon connection and authentication
+#
+# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>, qa-c team <qa-c@suse.de>
```
jlausuch, line 8: I guess we can just have qa-c contact, let's not put more burden on Clemens shoulders :)

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
...
+        $self->provider_client->init();
+    }
+    elsif ($provider eq 'GCE') {
+        # TODO
+    }
+    elsif ($provider eq 'AZURE') {
+        # TODO
+    }
+
+    $self->container_registry(get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REGISTRY", 'suse-qec-testing'));
```
jlausuch, line 48: would it make sense to move the initialization of this variable to the declaration (line19)?
```
has container_registry => sub { get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REGISTRY", 'suse-qec-testing') };
```

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
...
+}
+
+=head2 get_container_registry_prefix
+Get the full registry prefix URL for any containers image registry of ECR based on the account and region
```
jlausuch, line 52: A bit difficult to understand... maybe..
```suggestion
Get the full registry prefix URL (based on the account and region) to push container images on ECR.
```

lib/publiccloud/provider.pm
```diff
@@ -556,12 +554,31 @@ sub get_state_from_instance
     die('get_state_from_instance() isn\'t implemented');
 }
 
+
+=head2 get_container_registry_prefix
```
jlausuch, line 558: need to remove? It's the same code you added to the new file `k8s_provider.pm`, isn't it?

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
...
+=cut
+sub get_container_registry_prefix {
```
jlausuch, line 54: This method seems to be called ONLY from `get_container_image_full_name`.
Wouldn't it be better to just have 1 method?

```
sub get_container_image_uri {
    my ($self) = @_;
    my $full_name_prefix = sprintf('%s.dkr.ecr.%s.amazonaws.com', $self->provider_client->aws_account_id, $self->region);
    return "$full_name_prefix/" . $self->container_registry . ":$tag";
}
```

lib/publiccloud/ec2.pm
asmorodskyi: the field itself is in `aws_client` while code which init it is still here

lib/publiccloud/eks.pm
```diff
@@ -9,43 +9,25 @@
 # Documentation: https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html
 
 package publiccloud::eks;
-use Mojo::Base 'publiccloud::aws';
+use Mojo::Base 'publiccloud::k8s_provider';
 use testapi;
 use utils;
 
 has security_token => undef;
```
asmorodskyi, line 16: how this correlate with `security_token` which we already have in `aws_client` ?

lib/publiccloud/gcp.pm
```diff
@@ -36,6 +36,7 @@ sub vault_gcp_roles {
 sub init {
     my ($self) = @_;
     $self->SUPER::init();
+    $self->vault(publiccloud::vault->new());
```
asmorodskyi, line 39: PR is saying about AWS refactoring . is this part necessary to do now ?

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
...
+# Summary: Helper class for amazon connection and authentication
+#
+# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>, qa-c team <qa-c@suse.de>
+
+package publiccloud::k8s_provider;
+use Mojo::Base -base;
+use testapi;
+use mmapi 'get_current_job_id';
+
+has key_id => undef;
```
asmorodskyi, line 15: this line and two below already in `provider_client` , right ?

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,78 @@
...
+has region => undef;
+has resource_name => sub { get_var('PUBLIC_CLOUD_RESOURCE_NAME', 'openqa-vm') };
+has container_registry => undef;
+has provider_client => undef;
+has service => undef;
+
+sub init {
+    my ($self, $service) = @_;
+
+    die('The service must be specified') if (!$service);
```
asmorodskyi, line 26: `service` was introduced as workaround to keep k8s logic together with instance related at same flow . now after split do we really need it ? Below you actually use `$provider` anyway so maybe we can drop this variable ?

lib/publiccloud/aws_client.pm
```diff
@@ -0,0 +1,110 @@
...
+has security_token => undef;
+has region => undef;
+has ssh_key => undef;
+has ssh_key_file => undef;
+has credentials => undef;
+has vault => undef;
+has aws_account_id => undef;
+has service => undef;
+
+sub vault_create_credentials {
```
grisu48, line 29: No for this PR but we are coming closer to a point, where a independent vault class might start to make sense

lib/publiccloud/aws_client.pm
```diff
@@ -0,0 +1,110 @@
...
+    if ($self->service =~ /ECR|EC2/) {
+        my $max_tries = 6;
+        for my $i (1 .. $max_tries) {
+            my $out = script_output('aws ec2 describe-images --dry-run', 300, proceed_on_failure => 1);
+            return 1 if ($out !~ /AuthFailure/m && $out !~ /"aws configure"/m);
+            sleep 30;
+        }
+    }
+    elsif ($self->service eq "EKS") {
+        return 1;
```
grisu48, line 66: Don't we need checks as well on EKS?

lib/publiccloud/aws_client.pm
```diff
@@ -0,0 +1,110 @@
...
+
+    if (get_var('PUBLIC_CLOUD_SLES4SAP')) {
+        my $credentials_file
+          = "[default]" . $/ . 'aws_access_key_id=' . $self->key_id . $/ . 'aws_secret_access_key=' . $self->key_secret;
+
+        save_tmp_file(CREDENTIALS_FILE, $credentials_file);
+        assert_script_run('curl -O ' . autoinst_url . "/files/" . CREDENTIALS_FILE);
+    }
+
+    $self->aws_account_id(script_output("aws sts get-caller-identity | jq -r '.Account'"));
```
grisu48, line 100: I'd add a comment stating that `aws sts` is the secure token service, which is used for those credentials

lib/publiccloud/aws_client.pm
asmorodskyi: I don't think so , I think it is in right place . because to put this code into vault class you need to make it aware  about what is ECR/EC2/EKS because we have different logic for each so IMO this is right place for this logic no matter that it has `vault_create_credentials` name

lib/publiccloud/aws_client.pm
asmorodskyi: I think we do , but maybe it can be done in separate PR because this is just code refactoring and this skip of the check is already there ...

lib/publiccloud/ec2.pm
ilausuch: I prefer to remove ssh_key field from aws_client, the creation of keypair is only for ec2

lib/publiccloud/provider.pm
ilausuch: Yes, but now it is required by Google tests. So the best place is in the gcp until we have gcp_client in next PR

lib/publiccloud/k8s_provider.pm
ilausuch: We use that in other places, for instance ecr.pm

lib/publiccloud/gcp.pm
ilausuch: Yes, the creation of the vault instance was defined on provider.pm see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/55d00d82a6aeba43fd0ceb3b9dcd2ecb078106ba/lib/publiccloud/provider.pm#L45 but this will be moved to each client (aws_client and gcp_client) for now google is in the old way, so It should be maintained

lib/publiccloud/k8s_provider.pm
ilausuch: Service is needed to be passthrough to  each clients (aws_client...) so I am removing the variable from this class because is not needed to be stored

lib/publiccloud/aws_client.pm
ilausuch: Yes, I agree. In a next PR. I created a ticket https://progress.opensuse.org/issues/102774

lib/publiccloud/k8s_provider.pm
ilausuch: What I found is that in our k8s tests we don't are provided by key_id so the code always call to the vault to create them. So I removed this from k8s_provider.

lib/publiccloud/aws_client.pm
```diff
@@ -0,0 +1,110 @@
...
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Helper class for amazon connection and authentication
+#
+# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>, qa-c team <qa-c@suse.de>
+
+package publiccloud::aws_client;
+use Mojo::Base -base;
+# use Mojo::JSON 'decode_json';
```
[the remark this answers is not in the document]
ilausuch: done

lib/publiccloud/ec2.pm
ilausuch: done

lib/publiccloud/k8s_provider.pm
ilausuch: done

lib/publiccloud/k8s_provider.pm
ilausuch: Done

lib/publiccloud/eks.pm
ilausuch: Done

lib/publiccloud/k8s_provider.pm
ilausuch: Done

lib/publiccloud/k8s_provider.pm
ilausuch: Done

lib/publiccloud/aws_client.pm
ilausuch: Done

lib/publiccloud/aws_client.pm
grisu48: agreed

lib/publiccloud/aws_client.pm
grisu48: Perfect!

lib/publiccloud/k8s_provider.pm
ilausuch: done

ilausuch:
There is a problem in azure that I have to take care before of removing the WIP label

jlausuch:
LGTM

ilausuch:
Now the test of Azure has passed too

pdostal (approved):

lib/publiccloud/eks.pm
```diff
@@ -9,43 +9,24 @@
...
+    $self->SUPER::init("EKS");
 
     my $cluster = get_required_var("PUBLIC_CLOUD_K8S_CLUSTER");
     assert_script_run("aws eks update-kubeconfig --name $cluster", 120);
 }
 
+sub cleanup() {
+    my ($self) = @_;
+
+    $self->provider_client->cleanup();
```
cfconrad, line 29: You are creating the `provider_client` in k8s_provider.pm  but do cleanup here, I would try to keep new and destroy somehow connected.

lib/publiccloud/k8s_provider.pm
```diff
@@ -0,0 +1,70 @@
...
+has container_registry => sub { get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REGISTRY", 'suse-qec-testing') };
+has provider_client => undef;
+
+sub init {
+    my ($self, $service) = @_;
+    die('The service must be specified') if (!$service);
+
+    my $provider = get_required_var('PUBLIC_CLOUD_PROVIDER');
+
+    if ($provider eq 'EC2') {
```
cfconrad, line 27: I would condsider this as not oop design. The Parent should _not_ know what childs it has.
I would keep this client creation in the corresponding child classes

lib/publiccloud/eks.pm
```diff
@@ -9,43 +9,24 @@
...
-sub _check_credentials {
-    my ($self) = @_;
-    return 1;
-}
+has provider_client => undef;
 
 sub init {
     my ($self, %args) = @_;
-    $self->SUPER::init();
+    $self->SUPER::init("EKS");
```
cfconrad, line 20: see comment below, maybe it would be better to call it something like
```
    $self->SUPER::init(publiccloud::aws_client->new(
                region => $self->region,
                service => $service
            ));
```

PR #13702, opened 2021-11-17
[virt_autotest] Double check xen role for xen host at login_console

guoxuguang:
- Description
According to the latest OSD test gi-guest_win2019-on-host_developing-xen on 15-SP4 build61.1, find out that got Xen Domain0 kernel panic during login_console test. then boot up a normal kernel wihout xen kernel to continue the following test. mean that just only match virttest-bootmenu-xen-kernel needle at login_console step for xen host was not enough during getting kernel panic. 
So, we'd better double-check xen kernel module for xen host after matching virttest-bootmenu-xen-kernel needle successfully at login_console for this PR purpose. 
- Add
> Double-check xen kernel
> Check Xen hypervisor as Grub2 menuentry
> Check if start bootloader from a read-only snapshot

- Related ticket: https://progress.opensuse.org/issues/101509
- Verification run: 
validate login_console test module
xen
[gi-guest_developing-on-host_developing-xen](https://openqa.suse.de/tests/7747686#step/login_console/20)
[gi-guest_win2019-on-host_developing-xen](https://openqa.suse.de/tests/7747687#step/login_console/18)
kvm
[gi-guest_win2019-on-host_developing-kvm](https://openqa.suse.de/tests/7740988) 
reboot_and_wait_up_normal
[virt-guest-migration-sles12sp5-from-sles12sp5-to-developing-xen-src](https://openqa.suse.de/tests/7832069#step/reboot_and_wait_up_normal/25)
[virt-guest-migration-sles15sp3-from-sles15sp3-to-developing-xen-src](https://openqa.suse.de/tests/7832071#step/reboot_and_wait_up_normal/25)

### DIFF
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2012-2018 SUSE LLC
+# Copyright 2012-2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 #
 # Summary: virt_autotest: the initial version of virtualization automation test in openqa, with kvm and xen support fully
@@ -28,6 +28,25 @@ sub set_ssh_console_timeout_before_use {
     check_screen([qw(linux-login virttest-displaymanager)], 60);
     save_screenshot;
     send_key 'ret';
+}
+
+#Just only match bootmenu-xen-kernel needle was not enough for xen host if got Xen domain0 kernel panic(bsc#1192258)
+#Need to double-check xen role after matched bootmenu-xen-kernel needle successfully
+sub double_check_xen_role {
+    record_info 'INFO', 'Double-check xen kernel';
+    if (script_run('lsmod | grep xen') == 0) {
+        diag("Boot up xen kernel successfully");
+    }
+    else {
+        record_info 'INFO', 'Check Xen hypervisor as Grub2 menuentry';
+        die 'Check Xen hypervisor as Grub2 menuentry failed' if (script_run('grub2-once --list | grep Xen') != 0);
+        save_screenshot;
+        die 'Double-check xen kernel failed';
+    }
+
+    record_info 'INFO', 'Check if start bootloader from a read-only snapshot';
+    assert_script_run('touch /root/read-only.fs && rm -rf /root/read-only.fs');
+    save_screenshot;
 }
 
 sub login_to_console {
@@ -146,7 +165,8 @@ sub login_to_console {
     set_ssh_console_timeout_before_use if (is_remote_backend && is_aarch64 && get_var('IPMI_HW') eq 'thunderx');
     # use console based on ssh to avoid unstable ipmi
     use_ssh_serial_console;
-
+    # double-check xen role for xen host
+    double_check_xen_role if is_xen_host;
 }
 
 sub run {


guoxuguang:
@alice-suse @Julie-CAO @waynechen55 @nanzhg please help me review this PR if have time. Thanks

tests/virt_autotest/login_console.pm
```diff
@@ -146,6 +146,25 @@ sub login_to_console {
...
+    # double-check xen role for xen host
+    $self->double_check_xen_role if is_xen_host;
+
+}
+
+#Just only match bootmenu-xen-kernel needle was not enough for xen host if got Xen domain0 kernel panic(bsc#1192258)
+#Need to double-check xen role after matched bootmenu-xen-kernel needle successfully
+sub double_check_xen_role {
+    record_info 'INFO', 'Double-check xen kernel';
+    die 'Double-check xen kernel failed' if (script_run('lsmod | grep xen') != 0);
```
Julie-CAO, line 158: Under the current logic of these three checks, given two situations:

- if the host is bootup with xen kernel, then first check is enough, the second and third are redundent; 

- if the host is NOT bootup with xen kernel, the test will die and the second and the third check would not have the chance to be run.

So would it be better if we would change the checking logic to be like:
```
check if xen kernel is bootup
    if yes, then the host is good
    if no, then
       let's locate the reason
          is xen entry present on the grub2?
          is it bootup with a read-only filesystem?
          other reasons you may check, such as is it bootup with kvm and etc?
```

guoxuguang: Great! , but this `double_check_xen_role()` will deal with the following two situations:
situation 1:
match virttest-bootmenu-xen-kernel needle at login_console step successfully for Xen host, but, next, Xen domain0 occur kernel panic, and then boot up a normal kernel. So, we need to double-check xen kernel here 
situation 2:
unmatch grub2 needle at login_console step, xen host just boot up read-only filesystem at last.
we need to make sure if  read-only filesystem

Julie-CAO: //here 'xen kernel' means 'lsmod | grep xen'. just add these words in my comment above for clearer description.

waynechen55:
login_console.pm is called in several places. So I think you may need more verification runs. At least, it is called in reboot_and_wait_up.pm which then called by reboot_and_wait_up_upgrade and reboot_and_wait_up_normal.

guoxuguang:
> login_console.pm is called in several places. So I think you may need more verification runs. At least, it is called in reboot_and_wait_up.pm which then called by reboot_and_wait_up_upgrade and reboot_and_wait_up_normal.

bingo, next, I will add more verification runs for this PR.

tests/virt_autotest/login_console.pm
```diff
@@ -146,6 +146,25 @@ sub login_to_console {
...
+    save_screenshot;
+
+    record_info 'INFO', 'Check Xen hypervisor as Grub2 menuentry';
+    die 'Check Xen hypervisor failed' if (script_run('grub2-once --list | grep Xen') != 0);
```
waynechen55, line 162: Should be "Check Xen hypervisor as Grub2 menuentry failed' better ?

guoxuguang: sure, will update.

tests/virt_autotest/login_console.pm
```diff
@@ -146,6 +146,25 @@ sub login_to_console {
...
+    save_screenshot;
+
+    record_info 'INFO', 'Check if start bootloader from a read-only snapshot';
+    assert_script_run('touch /root/read-only.fs; rm -rf /root/read-only.fs');
```
waynechen55, line 166: Should be 'touch /root/read-only.fs && rm -rf /root/read-only.fs'. Otherwise, this command line will always return OK ?

guoxuguang: sure, had used 'touch /root/read-only.fs && rm -rf /root/read-only.fs' now.

tests/virt_autotest/login_console.pm
guoxuguang: had updated already.

tests/virt_autotest/login_console.pm
guoxuguang: short update, had updated the checking logic as below:
```
check if xen kernel is bootup   //here 'xen kernel' means 'lsmod | grep xen'
    if yes, then the host is good                                                                           ##deal with situation 1
    if no, then
       let's locate the reason
          is xen entry present on the grub2?
     at last, check with Xen VM host filesystem if it was a read-only filesystem   ##deal with situation 2
```

Julie-CAO: checking 'lsmod | grep xen', it is enough to convince it is a fine xen host. it is impossilbe that a host boots up into a XEN  read-only system. So I suggest to move the read-only check to the 'else' statement.

guoxuguang: sure,  but refer to https://progress.opensuse.org/issues/101812, just only depend on matching needles at login_conole step was not enough for XEN vm host. after double-check xen kernel successfully, we need to check with Xen vm host filesystem too.

Julie-CAO: Rather than 'matching needles', I said 'lsmod | grep xen' is enough. Have we ever experienced the situation that 'lsmod | grep xen' successfully while its filesystem is read-only?

guoxuguang: IMHO, no yet.  because of we do not check with xen kernel for xen vm host before. 
So, we have no experienced the situation that 'lsmod | grep xen' successfully while its filesystem is read-only. 
But, I had found out the situation that unmatch grub2 needle at login_console step for xen host, AFAIK. next, xen host boot up xen kernel with the read-only filesystem at last, so, we'd better check with xen kernel and make sure if filesystem was read-only together for xen vm host to deal with this situation.

guoxuguang: Refer to my local testing validation, there was existed one situation that boot up our SLE Xen vm host with xen kernel and read-only filesystem together. So, we'd better check with xen kernel and filesystem together for our Xen host.

Julie-CAO: Given the xen read-only system, we'd better check it your way. but you missed the pure read-only system in your logic, we experienced a few cases that the focus moved to the bottom of the grub menu due to xen-highlighted needle mismatch. Your only check if the xen entry exists in grub2, then the kernel panic and the needle mismatch cases both meet the checking, you can't determine it is kernel panic case or the needle mismatch case, except we don't care which one it is.

guoxuguang: AFAIK. according to the focus moved to the bottom of the grub2 menu due to xen-highlighted needle mismatch, then, there was only xen kernel + read-only filesystem on our Xen host without the situation that normal kernel +  read-only filesystem.

guoxuguang: AFAIK. there just only was the xen kernel and read-only filesystem without normal kernel + read-only filesystem after moved to the bottom of the grub2 menu due to xen-highlighted needle mismatch as your mentioned.

Julie-CAO: If that is the case, the checking logic is ok. LGTM.

tests/virt_autotest/login_console.pm
```diff
@@ -30,6 +30,27 @@ sub set_ssh_console_timeout_before_use {
...
+    record_info 'INFO', 'Double-check xen kernel';
+    if (script_run('lsmod | grep xen') == 0) {
+        diag("Boot up xen kernel successfully");
+    }
+    else {
+        record_info 'INFO', 'Check Xen hypervisor as Grub2 menuentry';
+        die 'Check Xen hypervisor as Grub2 menuentry failed' if (script_run('grub2-once --list | grep Xen') != 0);
+        save_screenshot;
+        die 'Double-check xen kernel failed';
+        save_screenshot;
```
Julie-CAO, line 45: minor issue: save_screenshot would be redundant as the test has already died...

guoxuguang: sure, had removed this `save_screenshot; `

Julie-CAO:
LGTM

guoxuguang:
Add the latest validation run:
reboot_and_wait_up_normal
[virt-guest-migration-sles12sp5-from-sles12sp5-to-developing-xen-src](https://openqa.suse.de/tests/7832069#step/reboot_and_wait_up_normal/25)
[virt-guest-migration-sles15sp3-from-sles15sp3-to-developing-xen-src](https://openqa.suse.de/tests/7832071#step/reboot_and_wait_up_normal/25)

If there was no new comment, this PR ready to merge now.

alice-suse:
@waynechen55  Does this LGTY? I am considering merging it if no hanging unresolved discussions.

tests/virt_autotest/login_console.pm
```diff
@@ -30,6 +30,26 @@ sub set_ssh_console_timeout_before_use {
...
+        record_info 'INFO', 'Check Xen hypervisor as Grub2 menuentry';
+        die 'Check Xen hypervisor as Grub2 menuentry failed' if (script_run('grub2-once --list | grep Xen') != 0);
+        save_screenshot;
+        die 'Double-check xen kernel failed';
+    }
+
+    record_info 'INFO', 'Check if start bootloader from a read-only snapshot';
+    assert_script_run('touch /root/read-only.fs && rm -rf /root/read-only.fs');
+    save_screenshot;
+
```
waynechen55, line 50: Please remove this empty line.

guoxuguang: sure, had removed this empty line now.

waynechen55: > @waynechen55 Does this LGTY? I am considering merging it if no hanging unresolved discussions.

LGTM.

waynechen55:
> @waynechen55 Does this LGTY? I am considering merging it if no hanging unresolved discussions.

LGTM

PR #13703, opened 2021-11-17
Fix for SystemSettingsPage

rakoenig:
- poo#102557
- Tab handling got the sanitize function implemented, so
   select statement for Tab labels should not have '&' in it.
- Fixed also the documentation for Tab.pm.

The first verificaton run also showed that the xterm used to become root didn't get
the focus automatically. So I added a patch for this as well. 

- Related ticket: https://progress.opensuse.org/issues/102557
- Needles: na
- Verification run: https://openqa.opensuse.org/tests/2041700


ge0r (approved):
LGTM! good catch!

PR #13706, opened 2021-11-18
Fix rpcinfo failed to show nfs on s390x

hjluo:
On s390x, the rpcinfo command sometimes failed to show the NFS information,
so we need to try more times.

- Related ticket: https://progress.opensuse.org/issues/101475  
- Needles: N/A
- Verification run: 
https://openqa.nue.suse.com/tests/7748331
https://openqa.nue.suse.com/tests/7748332
https://openqa.nue.suse.com/tests/7748333

### DIFF
--- a/lib/services/rpcbind.pm
+++ b/lib/services/rpcbind.pm
@@ -57,12 +57,20 @@ sub check_enabled {
 sub check_service {
     common_service_action('rpcbind.socket', $service_type, 'is-enabled');
     common_service_action('rpcbind.socket', $service_type, 'is-active');
+    common_service_action($nfs_server, $service_type, 'is-enabled');
+    common_service_action($nfs_server, $service_type, 'is-active');
 }
 
 sub check_function {
     assert_script_run("rpcinfo");
     # Wait for updated rpcinfo.
-    sleep(5);
+    if (script_run('rpcinfo | grep nfs') != 0) {
+        record_soft_failure('bsc#1193028', "rpcinfo can not get nfs info in time on s390x");
+        for (1 .. 10) {
+            last if (script_run('rpcinfo | grep nfs') == 0);
+            sleep 5;
+        }
+    }
     assert_script_run('rpcinfo | grep nfs');
     assert_script_run('mkdir -p /tmp/nfs');
     assert_script_run('mount -t nfs localhost:/rpcbindtest /tmp/nfs');


lib/services/rpcbind.pm
```diff
@@ -62,8 +62,10 @@ sub check_service {
 sub check_function {
     assert_script_run("rpcinfo");
     # Wait for updated rpcinfo.
-    sleep(5);
-    assert_script_run('rpcinfo | grep nfs');
+    for (1 .. 10) {
```
rfan1, line 65: Isn't a bug there? given the fact that nfs server is up and active, we can find the port

hjluo: not a bug though, just need to wait more time for nfs be shown in the rpcinfo. thanks

rfan1: Any syslog or journal logs show up before the nfs shows up?

hjluo: I've checked the log and can't find such information, thanks

rfan1: @hjluo May you please file a bug to track this "soft failure", I don't think it is a normal state

hjluo: Thanks for the suggestion, for NFS/RPC it's not always a reliable protocol, so sometimes rpcinfo may need NFS ready to get the NFS information. I'm just not sure it's a bug.

hjluo: filed bug https://bugzilla.suse.com/show_bug.cgi?id=1193028

rfan1 (approved):
lgtm

lib/services/rpcbind.pm
```diff
@@ -57,12 +57,20 @@ sub check_enabled {
...
 }
 
 sub check_function {
     assert_script_run("rpcinfo");
     # Wait for updated rpcinfo.
-    sleep(5);
+    if (script_run('rpcinfo | grep nfs') != 0) {
+        record_soft_failure 'bsc#1193028 - rpcinfo can not get nfs info in time on s390x'
+          for (1 .. 10) {
+            last unless script_run('rpcinfo | grep nfs') != 0;
```
lemon-suse, line 70: Could you use positive condition for the judgement?
Such as :
last if script_run('rpcinfo | grep nfs') == 0;

Too many negative conditions make others hard to understand the code.

lemon-suse (approved):

okurz:
To be correct, it would s390x, not s390X :)

hjluo:
> To be correct, it would s390x, not s390X :)

changed && thanks a lot

lib/services/rpcbind.pm
hjluo: changed

hjluo: changed

hjluo: changed

okurz:
you only changed the PR description title. The important part is the git commit message which will be part of the git log

hjluo:
> you only changed the PR description title. The important part is the git commit message which will be part of the git log

changed the commit message and thanks again.

lemon-suse (approved):

dzedro:
AGAIN the extensive VRs on all versions! https://openqa.suse.de/tests/7755651#step/rpcbind/70

ostruzina:
Hello, is there any news about fixing this? If not, would you consider reverting these changes?

These changes led to failures in all QE Maintenance job groups, thus blocking the approvals of maintenance updates.

hjluo:
> Hello, is there any news about fixing this? If not, would you consider reverting these changes?
> 
> These changes led to failures in all QE Maintenance job groups, thus blocking the approvals of maintenance updates.

Hello, you can see that our changes do include t the line 'grep working /tmp/nfs/test', it just timeout, our fix finds the NFS from rpcinfo command and it works. so you can file a ticket for this
timeout issue.
 
Thanks.

ostruzina:
@hjluo You claim that your code works, yet I don't see any verification run for any of the maintenance products. A simple verification run may have shown that there is an issue. This has brought us huge headaches because the approvals of maintenance updates were completely blocked.

PR #13708, opened 2021-11-18
Modify test suite aarch64_secureboot

ShawnHa0:
Modified verify_efi_mok.pm to fit Tumbleweed, mokutil --list-new returns no output on Tumbleweed while sle returns output of 'Moknew is empty'. Added condition to check if there is nothing being returned as the behavior in Tumbleweed.

- Related ticket: https://progress.opensuse.org/issues/101701
- Needles: N/A
- Verification run: 
    - O3: https://openqa.opensuse.org/tests/2040389
    - OSD: https://openqa.suse.de/tests/7698506

### DIFF
--- a/tests/console/verify_efi_mok.pm
+++ b/tests/console/verify_efi_mok.pm
@@ -110,7 +110,8 @@ sub check_mok {
     diag('Expected regex used to verify SecureBoot: ' . $state);
     validate_script_output 'mokutil --sb-state', $state;
 
-    if (script_output('mokutil --list-new', proceed_on_failure => 1) =~ /MokNew is empty/) {
+    my $output = script_output('mokutil --list-new', proceed_on_failure => 1);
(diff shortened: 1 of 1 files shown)


lilyeyes:
Usually we need offer the verification run for sle just in case. But it is OK as the changes looks good and simple.

lilyeyes (approved):
LGTM.

ShawnHa0:
> Usually we need offer the verification run for sle just in case. But it is OK as the changes looks good and simple.

I can do a verify run on sle as well, I will add it to the comment when it's done.

ShawnHa0:
> > Usually we need offer the verification run for sle just in case. But it is OK as the changes looks good and simple.
> 
> I can do a verify run on sle as well, I will add it to the comment when it's done.

OSD verify run now added.

rfan1 (approved):
LGTM

jouyingbin (approved):
LGTM

PR #13709, opened 2021-11-18
Validate output of zypper commands

grisu48:
Instead of relying on the return code only, validate part of the output
of some of the zypper commands to ensure, that the command handling
works properly.

- Related ticket: https://progress.opensuse.org/issues/102527
- Verification runs:
- [SLE15-SP3 docker](http://duck-norris.qam.suse.de/tests/7494) | [SLE15-SP3 podman](http://duck-norris.qam.suse.de/tests/7495) | [SLE15-SP3 podman aarch64](https://openqa.suse.de/tests/7728189) | [SLE15-SP3 docker aarch64](https://openqa.suse.de/tests/7728190)
- [12-SP5 docker](http://duck-norris.qam.suse.de/tests/7496)
- [12-SP3 docker](http://duck-norris.qam.suse.de/tests/7497)
- [15-SP3_image_on_sle-15-SP3_host_docker](http://duck-norris.qam.suse.de/tests/7498)
- [15-SP3_image_on_sle-15-SP3_host_podman aarch64](https://openqa.suse.de/tests/7728191)
- [Tumbleweed podman](https://openqa.opensuse.org/tests/2052545) | [Tumbleweed docker](https://openqa.opensuse.org/tests/2052546)
- [Tumbleweed podman aarch64](https://openqa.opensuse.org/tests/2052547) | [Tumbleweed docker aarch64](https://openqa.opensuse.org/tests/2052548)

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -168,12 +168,13 @@ sub test_opensuse_based_image {
             validate_script_output("$runtime run --entrypoint /bin/bash $image -c 'grep PRETTY_NAME /etc/os-release' | cut -d= -f2",
                 sub { /"SUSE Linux Enterprise Server ${pretty_version}${betaversion}"/ });
 
+            # SUSEConnect zypper service is supported only on SLE based image on SLE host
             unless (is_unreleased_sle) {
-                # SUSEConnect zypper service is supported only on SLE based image on SLE host
+                # we set --entrypoint specifically to the zypper plugin to avoid bsc#1192941
                 my $plugin = '/usr/lib/zypp/plugins/services/container-suseconnect-zypp';
-                $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1);
-                $runtime->run_container($image, cmd => "$plugin lp", keep_container => 1, timeout => 420, retry => 3, delay => 60);
-                $runtime->run_container($image, cmd => "$plugin lm", keep_container => 1, timeout => 420);
+                validate_script_output("$runtime run --entrypoint $plugin -i $image -v", sub { m/container-suseconnect version .*/ });
+                validate_script_output("$runtime run --entrypoint $plugin -i $image lp", sub { m/.*All available products.*/ });
+                validate_script_output("$runtime run --entrypoint $plugin -i $image lm", sub { m/.*Installed product.*\n.*Registration server set to.*\nAll available modules.*/ });
             }
         } else {
             record_info "non-SLE host", "This host ($host_id) does not support zypper service";
@@ -201,11 +202,17 @@ sub test_zypper_on_container {
     die 'Argument $image not provided!' unless $image;
     die 'Argument $runtime not provided!' unless $runtime;
 
-    $runtime->run_container($image, cmd => "zypper lr -s", keep_container => 1, timeout => 120);
-    $runtime->run_container($image, name => 'refreshed', cmd => "zypper -nv ref", keep_container => 1, timeout => 300);
+    # Check for bsc#1192941, which affects the docker runtime only
+    # bsc#1192941 states that the entrypoint of a derived image is set in a way, that program arguments are not passed anymore
+    # we run 'zypper ls' on a container, and if we detect the zypper usage message, we know that the 'ls' parameter was ignored
+    if ($runtime->runtime eq 'docker') {
+        record_soft_failure("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
+    }
+    validate_script_output("$runtime run -i --entrypoint '' $image zypper lr -s", sub { m/.*Alias.*Name.*Enabled.*GPG.*Refresh.*Service/ });
+    $runtime->run_container($image, name => 'refreshed', cmd => "zypper -nv ref", keep_container => 1, timeout => 300, retry => 3, delay => 60);
     $runtime->commit('refreshed', "refreshed-image", timeout => 120);
     $runtime->remove_container('refreshed');
-    $runtime->run_container($image, name => "refreshed-image", cmd => "zypper -nv ref", timeout => 300);
+    $runtime->run_container($image, name => "refreshed-image", cmd => "zypper -nv ref", timeout => 300, retry => 3, delay => 60);
     record_info "The End", "zypper test completed";
 }
 
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -114,7 +114,7 @@ sub build {
     record_info "$container_tag created", "";
 }
 
-=head2 run_container($image_name, [mode, name, remote, keep_container, timeout])
+=head2 run_container($image_name, [mode, name, remote, keep_container, timeout, retry, delay])
 
 Run a container.
 C<image_name> is required and can be the image id, the name or name with tag.
@@ -125,6 +125,8 @@ if C<cmd> found then it will execute the given command into the container.
 The container is always removed after exit.
 if C<keep_container> is 1 the container is not removed after creation. Default to get removed
 when it exits or when the daemon exits
+if C<retry> is given, the command is being repeated the given amount of times on failure
+If C<delay> is given, this defines the number of seconds between retries
 
 =cut
 sub run_container {


lib/containers/engine.pm
```diff
@@ -125,6 +130,9 @@ if C<cmd> found then it will execute the given command into the container.
 The container is always removed after exit.
 if C<keep_container> is 1 the container is not removed after creation. Default to get removed
 when it exits or when the daemon exits
+if C<retry> is given, the command is being repeaded the given amount of times on failure
```
b10n1k, line 133: typo on `repeaded`. However i am against the retry function. i cant stress this enough. from another point, also you introduce this but it is unrelated with the changes

lib/containers/engine.pm
```diff
@@ -137,8 +145,13 @@ sub run_container {
     my $params = sprintf qq(%s %s %s), $keep_container, $mode, $name;
     my $cmd = sprintf qq(run %s %s %s), $params, $image_name, $remote;
     record_info "cmd_info", "Container executes:\noptions $params $image_name $remote";
-    my $retries = $args{retry} // 1;
-    return $self->_engine_script_retry($cmd, timeout => $args{timeout}, retry => $retries, delay => $args{delay});
+    if ($args{validate}) {
+        die "retry and validate argument cannot be given at the same time" if ($args{retry});
```
b10n1k, line 149: why `die` and not continue with `validation` and print a warning, so we can fix it misused invocations later.

b10n1k (changes requested):
Even if i like your approach i would like to ask you to move the validation in a separate subroutine. i have tried in another PR which rejected by @jlausuch but i had different approach. i think and apparently wrapping `_engine_validate_script_output` makes more sense. 

however i wouldnt like the integration inside `run_container`. the benefits will be:
- readability. lets say the new function is called `validate_cmd`
`$runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });` vs `$runtime->validate_cmd($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });`
- run_container is developed to do run a container. So it is tied to `run`. you will be more flexible is you decouple the validation. For instance what would you do if you want to validate from an already running container. i think you get the idea

b10n1k:
just for idea. it looks to me that we mix `engine.pm` with other thing, which go outside of the scope of the class and i see space for making another class `containers.pm` or `images.pm` which acts on those routines. any thoughts?

lib/containers/engine.pm
grisu48: Will fix the typo. The discussion on `retry` is however outside the scope of this PR.

lib/containers/engine.pm
grisu48: Using `validate` and `retry` here is incorrect usage of the function and should by definition die.

I favor the approach that a function should fail early rather than have unexpected functionality. Here, if the user passes `validate` and `retry` (which is stated as not supported) the function will fail.
Your suggestion would silently ignore the `retry` argument, so someone could (and will) wrongly assumes that it would work otherwise.

Let's save ourselves some gray hair.

grisu48:
> Even if i like your approach i would like to ask you to move the validation in a separate subroutine. i have tried in another PR which rejected by @jlausuch but i had different approach. i think and apparently wrapping `_engine_validate_script_output` makes more sense.
> 
> however i wouldnt like the integration inside `run_container`. the benefits will be:
> 
> * readability. lets say the new function is called `validate_cmd`
>   `$runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });` vs `$runtime->validate_cmd($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });`
> * run_container is developed to do run a container. So it is tied to `run`. you will be more flexible is you decouple the validation. For instance what would you do if you want to validate from an already running container. i think you get the idea

Please not.

The `engine.pm` is already bloated enough. I was very close to replace 

    $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });

simply with the following:

    validate_script_output("$runtime run -i $plugin -v", validate => sub { m/container-suseconnect version .*/ });

The argument of let's create new functions for the purpose of we can extend them later on leads only to a bloated codebase. I don't see the advantage of duplicating every openQA API call within our engine class, just to add `$runtime` at the beginning (see e.g. `_engine_assert_script_run`, `_engine_script_run`, `_engine_script_output`, `_engine_script_retry`).

My PR is a compromise between the already existing functions, which adds an additional parameter instead of creating yet another library function.

* * *

In a nutshell: I much rather keep the changes small and the codebase tidy. I'm against refactoring a new library function simply for the argument that we might extend it in the future because this only bloats the codebase. What do the others think? @jlausuch @cfconrad @asmorodskyi @pdostal ?

pdostal:
I agree with @grisu48, but I especially agree with this:
> The argument of let's create new functions for the purpose of we can extend them later on leads only to a bloated codebase. I don't see the advantage of duplicating every openQA API call within our engine class, just to add `$runtime` at the beginning (see e.g. `_engine_assert_script_run`, `_engine_script_run`, `_engine_script_output`, `_engine_script_retry`).

lib/containers/engine.pm
pdostal: I don't want to block this, but just that you know the `script_retry()` actually can get you the output with `die => 0`, e.g.:
```
my $ret = script_retry($cmd, retry => 3, delay => 60, timeout 30, die => 0);
```

pdostal (approved):

lib/containers/engine.pm
grisu48: `script_retry` returns the return code of `$cmd` but not the output - See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1537

b10n1k:
> > Even if i like your approach i would like to ask you to move the validation in a separate subroutine. i have tried in another PR which rejected by @jlausuch but i had different approach. i think and apparently wrapping `_engine_validate_script_output` makes more sense.
> > however i wouldnt like the integration inside `run_container`. the benefits will be:
> > 
> > * readability. lets say the new function is called `validate_cmd`
> >   `$runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });` vs `$runtime->validate_cmd($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });`
> > * run_container is developed to do run a container. So it is tied to `run`. you will be more flexible is you decouple the validation. For instance what would you do if you want to validate from an already running container. i think you get the idea
> 
> Please not.

?? not what?

> 
> The `engine.pm` is already bloated enough. I was very close to replace
> 
> ```
> $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });
> ```
> 
> simply with the following:
> 
> ```
> validate_script_output("$runtime run -i $plugin -v", validate => sub { m/container-suseconnect version .*/ });
> ```
> 
> The argument of let's create new functions for the purpose of we can extend them later on leads only to a bloated codebase. I don't see the advantage of duplicating every openQA API call within our engine class, just to add `$runtime` at the beginning (see e.g. `_engine_assert_script_run`, `_engine_script_run`, `_engine_script_output`, `_engine_script_retry`).
> 

i didnt say "new function for the purpose of we can extend later on". i also didnt say of duplication of every Openqa API. i think there is a misunderstanding.

> My PR is a compromise between the already existing functions, which adds an additional parameter instead of creating yet another library function.
> 
> In a nutshell: I much rather keep the changes small and the codebase tidy. I'm against refactoring a new library function simply for the argument that we might extend it in the future because this only bloats the codebase. What do the others think? @jlausuch @cfconrad @asmorodskyi @pdostal ?

if you think the `engine.pm` is bloated i wiould like to know where. And you can see also my [comment below](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13709#issuecomment-972941626). 

Think also of the Open-close principle. A class should provide capabilities for extension but not modification in a proper design. i think this is violated if we continue like this and all the effort goes in the trash.

lib/containers/engine.pm
cfconrad: I'm not sure why retry will not be used with validate_output... Maybe you could use the retry approach from here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/wicked/wlan.pm#L324

cfconrad:
> The `engine.pm` is already bloated enough. I was very close to replace
> 
> ```
> $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });
> ```
> 
> simply with the following:
> 
> ```
> validate_script_output("$runtime run -i $plugin -v", sub { m/container-suseconnect version .*/ });
> ```

Actually I like the `validate_script_output()` here even more. It is more visible to me what it does.

lib/containers/engine.pm
grisu48: In that case we would have a functionality implemented, which is not used or tested anywhere. We could implement it now, but then we would also do add some unit tests or use it somewhere, just to make sure that this case is also functional.

At this point a failure seems to be more economical. We can implement this at any time we need it but we don't have to invest time and effort now into a functionality, which probably is never being used.

grisu48:
> > The `engine.pm` is already bloated enough. I was very close to replace
> > ```
> > $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > simply with the following:
> > ```
> > validate_script_output("$runtime run -i $plugin -v", sub { m/container-suseconnect version .*/ });
> > ```
> 
> Actually I like the `validate_script_output()` here even more. It is more visible to me what it does.

I agree. Shall we just switch to `validate_script_output` ?

lib/containers/engine.pm
grisu48: I mark this as resolved, as the original approach is not handled anymore.

grisu48:
New set of VRs are now scheduled

b10n1k:
you can go on without my approval.

lib/containers/container_images.pm
```diff
@@ -201,11 +201,12 @@ sub test_zypper_on_container {
     die 'Argument $image not provided!' unless $image;
     die 'Argument $runtime not provided!' unless $runtime;
 
-    $runtime->run_container($image, cmd => "zypper lr -s", keep_container => 1, timeout => 120);
-    $runtime->run_container($image, name => 'refreshed', cmd => "zypper -nv ref", keep_container => 1, timeout => 300);
+    record_soft_failure("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
```
jlausuch, line 204:
```suggestion
    record_soft_failure("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("$runtime run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
```

lib/containers/engine.pm
```diff
@@ -125,6 +125,7 @@ if C<cmd> found then it will execute the given command into the container.
 The container is always removed after exit.
 if C<keep_container> is 1 the container is not removed after creation. Default to get removed
 when it exits or when the daemon exits
+if C<retry> is given, the command is being repeated the given amount of times on failure
```
jlausuch, line 128: maybe need to add `delay` parameter as well here...

lib/containers/container_images.pm
grisu48: Good that you point this out. The error is a different one: This should on run for `docker` .....

grisu48:
> 

I added an extension to `run_container` as first suggestion, however that one was not accepted. My initial suggesion was to add a `validate` parameter to `run_container`:

    $runtime->run_container($image, cmd => "$plugin -v", keep_container => 1, validate => sub { m/container-suseconnect version .*/ });

Because we [didn't found a consensus](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13709#issuecomment-972973931) I went back to simply using `validate_script_output` because it is simple, readable and we [are already using it](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/containers/container_images.pm#L119). Clemens agreed on this and [my request if we should adapt this instead](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13709#issuecomment-973885274) remained unanswered.

For this PR I'd leave it as it is, because I'm not introducing a new concept but do the same as we already do. I'd leave a refactoring of this for a future PR when we have a clear idea what we want to have. Right now we have three different approaches and no consistent opinion on how this should be done.

jlausuch (approved):
Ok, Let's try to address the Classes in a different PR.

PR #13714, opened 2021-11-19
To test 389ds before and after migration

tinawang123:
To test 389ds before and after migration, use 389ds client to
connect with 389ds server before and after migration

- Related ticket: https://progress.opensuse.org/issues/101085
- Verification run: 
](https://openqa.suse.de/tests/7704170#)
https://openqa.suse.de/tests/7704171#

### DIFF
--- a/schedule/migration/tls_389ds_sssd.yaml
+++ b/schedule/migration/tls_389ds_sssd.yaml
@@ -1,0 +1,31 @@
+name: tls_389ds_sssd
+description:    >
+    This is for 389ds and sssd authentication before and after migration check
+vars:
+    ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
+    UPGRADE_TARGET_VERSION: '%VERSION%'
+schedule:
+    - '{{tls_389ds}}'
+conditional_schedule:
+    tls_389ds:
+        HOSTNAME:
+            server:
+                - migration/version_switch_origin_system
+                - installation/isosize
+                - installation/bootloader
+                - migration/online_migration/online_migration_setup
+                - console/consoletest_setup
+                - network/setup_multimachine
+                - migration/online_migration/register_system
+                - migration/online_migration/zypper_patch
+                - console/389ds/tls_389ds_server
+                - migration/version_switch_upgrade_target
+                - migration/online_migration/pre_migration
+                - migration/online_migration/zypper_migration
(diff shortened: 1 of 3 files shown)


rfan1:
CI check failed

schedule/migration/tls_389ds_sssd.yaml
```diff
@@ -0,0 +1,31 @@
+name: tls_389ds_sssd
+description:    >
+    This is for 389ds and sssd authentication before and after migration check
+vars:
+  ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
```
rfan1, line 5: yamllint --strict schedule/migration/tls_389ds_sssd.yaml
I remembered that there should be 4 blank space here :)

tests/console/389ds/tls_389ds_server.pm
```diff
@@ -0,0 +1,44 @@
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary: Implement & Integrate 389ds + sssd test case into openQA,
```
rfan1, line 4: Would you please add more comments related to migration tests?

schedule/migration/tls_389ds_sssd.yaml
```diff
@@ -0,0 +1,31 @@
+name: tls_389ds_sssd
```
lemon-suse, line 1: Why named as tls_389ds_sssd? What's the relationship between tls, sssd with this test?

tests/console/389ds/tls_389ds_server.pm
```diff
@@ -0,0 +1,45 @@
...
```
lemon-suse, line 4: Need note somewhere that this case now only supported to be run for migration path from SLES15SP3 to SLES15SP4.

tinawang123: I think it depended on your cases. I can add comment to this condition, but I don't like to add code to this condition.

schedule/migration/tls_389ds_sssd.yaml
tinawang123: tls is a protocol. sssd is part of 389ds. 
I just followed security cases' name

lemon-suse (approved):

PR #13717, opened 2021-11-22
ZFCP activation via libyui

rakoenig:
Activate ZFCP disks via libyui

- Related ticket: https://progress.opensuse.org/issues/98967
- Needles: n.a.
- Verification run: https://openqa.suse.de/tests/7755204

### DIFF
--- a/lib/Distribution/Opensuse/Tumbleweed.pm
+++ b/lib/Distribution/Opensuse/Tumbleweed.pm
@@ -14,6 +14,9 @@ use warnings FATAL => 'all';
 use parent 'susedistribution';
 use Installation::AuthenticationForRoot::AuthenticationForRootController;
 use Installation::ClockAndTimeZone::ClockAndTimeZoneController;
+use Installation::DiskActivation::DiskActivationController;
+use Installation::DiskActivation::ConfiguredZFCPDevicesController;
+use Installation::DiskActivation::AddZFCPDeviceController;
 use Installation::LanguageKeyboard::LanguageKeyboardController;
 use Installation::License::Opensuse::Firstboot::LicenseAgreementController;
 use Installation::License::Opensuse::LicenseAgreementController;
@@ -163,5 +166,16 @@ sub get_encrypted_volume_activation {
     return Installation::SystemProbing::EncryptedVolumeActivationController->new();
 }
 
+sub get_disk_activation {
+    return Installation::DiskActivation::DiskActivationController->new();
+}
+
+sub get_configured_zfcp_devices {
+    return Installation::DiskActivation::ConfiguredZFCPDevicesController->new();
+}
+
+sub get_add_new_zfcp_device {
+    return Installation::DiskActivation::AddZFCPDeviceController->new();
+}
 
 1;
--- a/lib/Installation/DiskActivation/AddZFCPDevicePage.pm
+++ b/lib/Installation/DiskActivation/AddZFCPDevicePage.pm
@@ -1,0 +1,32 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to the Add New ZFCP Device dialog
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::AddZFCPDevicePage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self, $args) = @_;
+    $self->SUPER::init($args);
+    $self->{cmb_channel} = $self->{app}->combobox({id => 'channel'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{cmb_channel}->exist();
+}
+
+sub set_channel {
+    my ($self, $channel) = @_;
+    return $self->{cmb_channel}->set($channel);
+}
+
+1;
--- a/lib/Installation/DiskActivation/DiskActivationController.pm
+++ b/lib/Installation/DiskActivation/DiskActivationController.pm
@@ -1,0 +1,53 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act on the Disk Activation page
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::DiskActivationController;
+use strict;
+use warnings;
+use Installation::DiskActivation::DiskActivationPage;
+use YuiRestClient;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {}, $class;
+    return $self->init($args);
+}
+
+sub init {
+    my ($self, $args) = @_;
+    $self->{DiskActivationPage} = Installation::DiskActivation::DiskActivationPage->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_disk_activation_page {
+    my ($self) = @_;
+    die "Disk Activation Page is not displayed" unless $self->{DiskActivationPage}->is_shown();
+    return $self->{DiskActivationPage};
+}
+
+sub configure_dasd_disks {
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_dasd();
+}
+
+sub configure_zfcp_disks {
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_zfcp();
+}
+
+sub configure_iscsi_disks {
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_iscsi();
+}
+
+sub accept_disks_configuration {
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_next();
+}
+
+1;
--- a/lib/Installation/DiskActivation/DiskActivationPage.pm
+++ b/lib/Installation/DiskActivation/DiskActivationPage.pm
@@ -1,0 +1,45 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act on the Disk Activation page
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::DiskActivationPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self, $args) = @_;
+    $self->SUPER::init($args);
+    $self->{btn_conf_dasd} = $self->{app}->button({id => 'dasd'});
+    $self->{btn_conf_zfcp} = $self->{app}->button({id => 'zfcp'});
+    $self->{btn_conf_iscsi} = $self->{app}->button({id => 'iscsi'});
+    $self->{btn_net_conf} = $self->{app}->button({id => 'network'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{btn_net_conf}->exist();
+}
+
+sub press_dasd {
+    my ($self) = @_;
+    return $self->{btn_conf_dasd}->click();
+}
+
+sub press_zfcp {
+    my ($self) = @_;
+    return $self->{btn_conf_zfcp}->click();
+}
+
+sub press_iscsi {
+    my ($self) = @_;
+    return $self->{btn_conf_iscsi}->click();
+}
+
+1;
--- a/schedule/yast/zfcp.yaml
+++ b/schedule/yast/zfcp.yaml
@@ -1,5 +1,5 @@
-name:           zfcp
-description:    >
+name: zfcp
+description: >
   Test installation on machine with zfcp multipath disk.
   Only tests succesful detection of multipath and installation.
   No functional testing of multipath itself.
@@ -12,7 +12,11 @@ schedule:
   - installation/access_beta_distribution
   - installation/product_selection/install_SLES
   - installation/licensing/accept_license
-  - installation/disk_activation
+  - installation/disk_activation/select_configure_zfcp_disks
+  - installation/disk_activation/select_add_zfcp_device
+  - installation/disk_activation/configure_zfcp_device
+  - installation/disk_activation/accept_configured_zfcp_devices
+  - installation/disk_activation/finish_disk_activation
   - installation/activate_multipath
   - installation/registration/register_via_scc
   - installation/module_selection/select_module_desktop
--- a/tests/installation/disk_activation/select_configure_zfcp_disks.pm
+++ b/tests/installation/disk_activation/select_configure_zfcp_disks.pm
@@ -1,0 +1,20 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: This test module handles ZFCP disk activation
+#          through libyui-rest-client.
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    my $disk_activation = $testapi::distri->get_disk_activation();
+    $disk_activation->configure_zfcp_disks();
+}
+
+1;
--- a/lib/Installation/DiskActivation/AddZFCPDeviceController.pm
+++ b/lib/Installation/DiskActivation/AddZFCPDeviceController.pm
@@ -1,0 +1,39 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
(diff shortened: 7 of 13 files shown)


jknphy:
Please, while you have a PR with WIP commits, you should set in the title of the PR "WIP: " as well, otherwise that commit could be merge accidentally. You also need to squash some commit there, you have repeated titles.

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
...
+    $zfcp_configuration_overview->add();
+
+    # In the Add Device page insert hard wired channel id and press next
+    my $zfcp_add_disk = $testapi::distri->get_zfcp_add_disk();
+    $zfcp_add_disk->set_channel("0.0.fa00");
+    $zfcp_add_disk->next();
+
+    # yui.log from 1st VR showed that it takes a while for
+    # device detection so lets sleep a bit.
+    sleep(8);
```
jknphy, line 32: Please, never use sleep in a test, I cannot find right now the document where this is explicitly forbidden, or ok, not recommended by contributors of this repo. In general is a very bad practice in any language, because you never know exactly how much to wait. We use it though for our framework because we need some sort of timeout but not in a test.

Also I don't thing you need it, the next page should just check with its `exists` method and this arch is not slow.

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
...
+use base 'y2_installbase';
+use YuiRestClient::Logger;
+
+sub run {
+    # in Disk Activation page press "ZFCP" button
+    my $disk_activation_selector = $testapi::distri->get_disk_activation_selector();
+    $disk_activation_selector->zfcp();
+
+    # in the ZFCP Device overview press the "Add" button
+    my $zfcp_configuration_overview = $testapi::distri->get_zfcp_configuration_overview();
```
jknphy, line 22: This two lines should be in another test module, `select_add_new_zfcp_device`. After that you will have another test module `configure_zfcp_device`. You could argue that both cases could be merge in one, but I think we will see it more clear like that IMO.

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -163,5 +166,16 @@ sub get_encrypted_volume_activation {
...
 
+sub get_disk_activation_selector {
+    return Installation::DiskActivation::DiskActivationController->new();
+}
+
+sub get_zfcp_configuration_overview {
+    return Installation::DiskActivation::ZFCPConfigurationController->new();
+}
+
+sub get_zfcp_add_disk {
```
jknphy, line 177: s/get_zfcp_add_disk/get_add_new_zfcp_device/

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
...
+    $zfcp_configuration_overview->add();
+
+    # In the Add Device page insert hard wired channel id and press next
+    my $zfcp_add_disk = $testapi::distri->get_zfcp_add_disk();
+    $zfcp_add_disk->set_channel("0.0.fa00");
```
jknphy, line 27: We have tried to hide the `next` all that we can, in this case you can just have a `configure` method with named params, i.e.: `->configure(channel => '0.0.fa00')` (btw, there is not need to double quotes when there is not interpolation, but if you prefer like that is ok which does the next inside.

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
...
+    $zfcp_add_disk->next();
+
+    # yui.log from 1st VR showed that it takes a while for
+    # device detection so lets sleep a bit.
+    sleep(8);
+
+    # Now we should be back in ZFCP Device overview page
```
jknphy, line 34: here we need a new test module, I would have not bother to check the table because we just rely on the libyui-rest-api doing well its work and then we check the installed system afterward, in other words, this is considered more like what we have called extended scenario.

I would recommend to save this code in some branch and pull out from now, it is overly complicated but I see good intention on that. We already have some function that compares arrays: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a96c5428635e34d07cf2b520e4d557529945f863/lib/utils.pm#L1159
I would definitively leave it for later.

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -163,5 +166,16 @@ sub get_encrypted_volume_activation {
     return Installation::SystemProbing::EncryptedVolumeActivationController->new();
 }
 
+sub get_disk_activation_selector {
```
jknphy, line 169: s/get_disk_activation_selector/get_disk_activation/ It is a controller not a selector or other thing, and we agreed to not use _controllers as from $testapi::distri-> all the calls we do from this layer is to Controllers' layer.

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -163,5 +166,16 @@ sub get_encrypted_volume_activation {
...
+    return Installation::DiskActivation::DiskActivationController->new();
+}
+
+sub get_zfcp_configuration_overview {
```
jknphy, line 173: s/get_zfcp_configuration_overview/get_configured_zfzp_devices/

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
...
+          YuiRestClient: Logger->get_instance()->error("$devs[$i] does not match $expectdevs[$i]");
+        }
+    }
+    if ($mismatchcount > 0) {
+        die "Wrong device names returned";
+    }
+
+    # If we reach it here then everything is fine
+    # Leave ZFCP Configuration Overview.
+    $zfcp_configuration_overview->next();
```
jknphy, line 67: here we should go with two tests as well, `accept_configured_zfcp_devices` and `finish_disk_activation`.

jknphy (commented):
It is going in the right direction, just need a bit more splitting for atomicity which would make the comments in the test redundant and we probably can skip to check the table as we did in many places in the past and a proper validation will be done in the future.

tests/installation/disk_activation_zfcp.pm
```diff
@@ -0,0 +1,73 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: This test module handles ZFCP disk activation
```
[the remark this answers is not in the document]
rakoenig: Ok. Just spotted a disk_activation_iscsi and thought then its also good for zfcp...

lib/Installation/DiskActivation/DiskActivationController.pm
```diff
@@ -0,0 +1,53 @@
...
+    return $self;
+}
+
+sub get_disk_activation_page {
+    my ($self) = @_;
+    die "Disk Activation Page is not displayed" unless $self->{DiskActivationPage}->is_shown();
+    return $self->{DiskActivationPage};
+}
+
+sub dasd {
```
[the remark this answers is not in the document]
rakoenig: Ok. Changed that, but the change will affect George as well

lib/Installation/DiskActivation/DiskActivationController.pm
```diff
@@ -0,0 +1,53 @@
...
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_zfcp();
+}
+
+sub iscsi {
+    my ($self) = @_;
+    $self->get_disk_activation_page()->press_iscsi();
+}
+
+sub next {
```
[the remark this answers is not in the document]
rakoenig: Ok, changed that.

lib/Installation/DiskActivation/DiskActivationPage.pm
```diff
@@ -0,0 +1,56 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act on the Disk Activation page
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::DiskActivationPage;
```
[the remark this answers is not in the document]
rakoenig: Done. That saves me from the press_next and next_btn things, isn't it?

lib/Installation/DiskActivation/DiskActivationPage.pm
```diff
@@ -0,0 +1,56 @@
...
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init;
+}
+
+sub init {
```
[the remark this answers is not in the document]
rakoenig: Done.

tests/installation/disk_activation_zfcp.pm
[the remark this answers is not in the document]
jknphy: please, do not resolve the comments until you change the code, otherwise is hard to review or know what is changed. Normally I can see some 'oudated' code label in top to know that.

lib/Installation/DiskActivation/DiskActivationController.pm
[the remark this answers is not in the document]
jknphy: don't worry about it at all. You didn't divide commits by layers you will have issue for sure, but ignore sync, was kind of experiment.

lib/Installation/DiskActivation/DiskActivationPage.pm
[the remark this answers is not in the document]
jknphy: yes, for each page which inherit from it.

tests/installation/disk_activation_zfcp.pm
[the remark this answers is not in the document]
jknphy: also another internal rule we use is that, we should reply to all comments by if it is enough clear and you change the code you don't need to write a reply, just with the 'outdated' file message is understood.

lib/Installation/DiskActivation/ZFCPConfigurationPage.pm
```diff
@@ -0,0 +1,53 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to the Configured ZFCP dialog dialog
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::ZFCPConfigurationPage;
```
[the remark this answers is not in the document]
rakoenig: I understand the inheritance of the press_next method. I'm still struggling to get the concept of new. If I remove new from the child class then new from the parent class (in this case NavigationBase) will be used when invoking Child::new(). This new calls a method init() but does that mean NavigationBase::init() or Chil::init(): 
I'm a bit  confused because LicenseAgreementPage.pm still has  a new method, so there should be a reason for this.

lib/Installation/DiskActivation/DiskActivationController.pm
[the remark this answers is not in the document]
rakoenig: Well, I diveded the commits by layers. DiskActivationController/Page is one, ZFCPOverviewContoller/Page and AddZFCPController/Page are individual commits.

lib/Installation/DiskActivation/ZFCPConfigurationPage.pm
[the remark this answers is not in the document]
jknphy: init is overwritten and use the child one, it is call I think initialization pattern, help to not mess up with the constructor much and have that inheritiance for initialization. The reason there are new() method exactly the same in parent and child when there is this inheritance from Navigation is that we just forget to delete them.

rakoenig:
Ok, code completely refactored, test is split up in 5 atomic steps. New VR in the start comment.

lib/Installation/DiskActivation/DiskActivationPage.pm
```diff
@@ -0,0 +1,44 @@
...
+    $self->{btn_conf_dasd} = $self->{app}->button({id => 'dasd'});
+    $self->{btn_conf_zfcp} = $self->{app}->button({id => 'zfcp'});
+    $self->{btn_conf_iscsi} = $self->{app}->button({id => 'iscsi'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{btn_conf_zfcp}->exist();
+}
```
ge0r, line 27: There seems to be a small issue with this `is_shown` because in some cases `btn_conf_zfcp` does not exist, like in [this case](https://openqa.suse.de/tests/7749086#step/configure_dasd/1), which results in test failure.
I think one good way to mitigate this is to remove this `is_shown` function completely, since it is inherited either way from the `NavigationBase` parent, which is checking if the next button exists for it's `is_shown` implementation.
[Here](https://openqa.suse.de/tests/7749121#step/configure_dasd/1) is a small verification run of how removing the `is_shown` function in `DiskActivationPage.pm` seems to work fine.

ge0r (changes requested):

lib/Installation/DiskActivation/DiskActivationPage.pm
rakoenig: I partly agree with you. Yes, the current problem is that `btn_conf_zfcp` does not exist in your case. Removing the `is_shown()` method in `DiskActivationPage.pm` would then use the `is_shown()` from `NavigationBase.pm`.

What then worries me is that `is_shown()` will be true for **any page** that has a `Next` button. So if e.g. pressing the  `btn_config_zfcp` in my test would not work and I would not end up in the "Configured ZFCP Devices" page I would not notice it if I'm just checking for a `Next` button. On the other hand we don't have a libYUI widget that for "YWizard" that could use the debug_label to determine which page is displayed. 

Of course I have no idea how big the risk is to have problems because of that. So for the moment I'm going to remove the `is_shown()` from ` DiskActivationPage.pm`.

ge0r: You have a valid point here, checking for `Next` button is not a very good check because many pages during installation have a next button so we can not be sure that the page we are interested is indeed shown.

Maybe an implementation that does not include the zfcp button then?

lib/Installation/DiskActivation/DiskActivationPage.pm
```diff
@@ -0,0 +1,45 @@
...
+use strict;
+use warnings;
+
+sub init {
+    my ($self, $args) = @_;
+    $self->SUPER::init($args);
+    $self->{btn_conf_dasd} = $self->{app}->button({id => 'dasd'});
+    $self->{btn_conf_zfcp} = $self->{app}->button({id => 'zfcp'});
+    $self->{btn_conf_iscsi} = $self->{app}->button({id => 'iscsi'});
+    $self->{btn_net_conf} = $self->{app}->butoon({id => 'network'});
```
ge0r, line 21:
```suggestion
    $self->{btn_net_conf} = $self->{app}->button({id => 'network'});
```

rakoenig: Typo fixed.

ge0r (approved):

lib/Installation/DiskActivation/AddZFCPDevicePage.pm
```diff
@@ -0,0 +1,35 @@
...
+# Summary: The module provides interface to the Add New ZFCP Device dialog
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::AddZFCPDevicePage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+
```
jknphy, line 15: only one line between areas of code, also check at the end of the rest of the files between `}` and `1`.

schedule/yast/zfcp.yaml
```diff
@@ -47,44 +51,44 @@ test_data:
         installed: 1
   zfcp:
     fcp_devices:
-      - fcp_channel: '0.0.fa00'
```
jknphy, line 50: what is the issue to change to double quotes? Seems fine, even when there are no spaces is not required to use any quotes in yaml.

jknphy (commented):
I pointed to a few very minor things. Good job @rakoenig I'm glad that you address the feedback so efficiently. You can remove the WIP and it is ready to go.

jknphy:
I checked only the final code, but per commits you need to resolve some conflicts and squash some commits yet. Once you finish would be great to retrigger VR so we don't miss anything refactored/renamed.

tests/installation/disk_activation/select_configure_zfcp_disks.pm
```diff
@@ -0,0 +1,21 @@
...
+
+# Summary: This test module handles ZFCP disk activation
+#          through libyui-rest-client.
+
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use YuiRestClient::Logger;
```
jknphy, line 14: let's remove the Logger from all test files, it is not needed ;)

PR #13718, opened 2021-11-22
Rename LTP variables with prefix for the corresponding module

b10n1k:
following the name convention that kernel modules uses and for the sake of sanity between those which are used for the same exact purpose.
that also will make the variable more specific. for instance
COMMAND_EXCLUDE - LTP_COMMAND_EXCLUDE
COMMAND_FILE - LTP_COMMAND_FILE

- Related ticket: https://progress.opensuse.org/issues/98334
- Verification run: https://openqa.suse.de/t7719617
https://openqa.suse.de/tests/7725754 for d7c99b0d659e9e17ae910bd4f6848adb72d631d9

### DIFF
--- a/variables.md
+++ b/variables.md
@@ -36,8 +36,6 @@ CLUSTER_TYPES | string | false | Set the type of cluster that have to be analyze
 CONTAINER_RUNTIME | string | | Container runtime to be used, e.g.  `docker`, `podman`, or both `podman,docker`.
 CONTAINERS_NO_SUSE_OS | boolean | false | Used by main_containers to see if the host is different than SLE or openSUSE.
 CONTAINERS_UNTESTED_IMAGES | boolean | false | Whether to use `untested_images` or `released_images` from `lib/containers/urls.pm`.
-COMMAND_FILE | string | | The LTP test command file (e.g. syscalls, cve)
-COMMAND_EXCLUDE | string | | This regex is used to exclude tests from LTP command file.
 CPU_BUGS | boolean | | Into Mitigations testing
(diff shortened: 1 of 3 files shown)


grisu48 (approved):
LGTM :rocket: 

I'd like to see one VR with a intact job schedule, then we can merge. As long as LTP runs, it should be safe, IMHO we don't need to wait for a full run.

pdostal (approved):
:rocket:

asmorodskyi (approved):
LGTM , can be merged but after VR

jlausuch (approved):

czerw:
We use `LTP_COMMAND_FILE` and `LTP_COMMAND_EXCLUDE` already and changed files are use by public cloud only. It seems ok for.  But better doublecheck @mdoucha @pevik could  you please review as well?

variables.md
```diff
@@ -96,6 +94,8 @@ LIVECD | boolean | false | Indicates live image being used.
 LIVE_INSTALLATION | boolean | false | If set, boots the live media and starts the builtin NET installer.
 LIVE_UPGRADE | boolean | false | If set, boots the live media and starts the builtin NET installer in upgrade mode.
 LIVETEST | boolean | false | Indicates test of live system.
+LTP_COMMAND_FILE | string | | The LTP test command file (e.g. syscalls, cve)
+LTP_COMMAND_EXCLUDE | string | | This regex is used to exclude tests from LTP command file.
```
pevik, line 98: LGTM. FYI we document these variables already in tests/kernel/run_ltp.pm. Also there are more variables, LTP_COMMAND_PATTERN in "our" LTP scripts. Not sure if it's a good idea to document some variables here and some in test modules.

pevik (approved):

czerw (approved):
LGTM

variables.md
jlausuch: I would favor for `variables.md`, which is more visible.

PR #13719, opened 2021-11-22
Fix for problems in disable_boot_menu_timeout.pm

rakoenig:
- Module was called from YAML schedules that don't use
  setup_libyui
- YAML schedules reverted back to disable_grub_timeout

- Related ticket: https://progress.opensuse.org/issues/102780
- Needles: n.a.
- Verification run: https://openqa.suse.de/tests/7718210

### DIFF
--- a/schedule/yast/encryption/lvm_encrypt_separate_boot_pvm.yaml
+++ b/schedule/yast/encryption/lvm_encrypt_separate_boot_pvm.yaml
@@ -27,7 +27,7 @@ schedule:
   - installation/authentication/default_user_simple_pwd
   - installation/installation_overview
   - installation/bootloader_settings/disable_plymouth
-  - installation/bootloader_settings/disable_boot_menu_timeout
+  - installation/disable_grub_timeout
   - installation/start_install
   - installation/await_install
   - installation/logs_from_installation_system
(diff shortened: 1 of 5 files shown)


ge0r (approved):
LGTM!

schedule/yast/encryption/lvm_encrypt_separate_boot_pvm.yaml
```diff
@@ -27,7 +27,7 @@ schedule:
   - installation/authentication/default_user_simple_pwd
   - installation/installation_overview
   - installation/bootloader_settings/disable_plymouth
-  - installation/bootloader_settings/disable_boot_menu_timeout
+  - installation/disable_grub_timeout
```
jknphy, line 30: looks like this change was not needed. right?

rakoenig: You are right from the point of view that disable_plymouth is using libyui as well. I guess the problem is that I selected the files to change by loooking who is using disable_boot_menu_timeout but not using setup_libyui. Now looking at the YAML again I see a innocent variable "YUI_REST_API: 1" that seems to trigger libyui without explicitly mentioning in the schedule. Now I clearly understand the very deep love for variables... ;-)

jknphy: could you do a quick revert of this line?

rakoenig: Ok, pushed a commit with reverting this to the branch.

PR #13724, opened 2021-11-23
Need more time to match gnome-setting on ARM

lemon-suse:
Through manual test on ARM, I found after click the 'setting' it will show the desktop firstly then show the gnome-setting, it is slow on ARM, so we need wait more time to catch the gnome-setting.

- Related ticket: https://progress.opensuse.org/issues/102494
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/7714299#step/install_service/342

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -740,7 +740,8 @@ sub start_gnome_settings {
 
     if (!$is_sle_12_sp1 || $settings_menu_loaded) {
         assert_and_click 'settings';
-        assert_screen 'gnome-settings';
+        my $timeout = (check_var('ARCH', 'aarch64')) ? '180' : '30';
+        assert_screen 'gnome-settings', $timeout;
     }
 }
 


lib/x11test.pm
```diff
@@ -740,7 +740,9 @@ sub start_gnome_settings {
 
     if (!$is_sle_12_sp1 || $settings_menu_loaded) {
         assert_and_click 'settings';
-        assert_screen 'gnome-settings';
+        my $timeout = 30;
+        $timeout = 180 if (check_var('ARCH', 'aarch64'));
```
tinawang123, line 744: Only aarch64?

hjluo (approved):

lib/x11test.pm
```diff
@@ -740,7 +740,8 @@ sub start_gnome_settings {
 
     if (!$is_sle_12_sp1 || $settings_menu_loaded) {
         assert_and_click 'settings';
-        assert_screen 'gnome-settings';
+        my $timeout = (check_var('ARCH', 'aarch64')) ? '180' : '30';
+        assert_screen 'gnome-settings', $timeout;
```
rfan1, line 744: Lemon, why don't we use assert_screen 'gnome-settings', 180? here?

lemon-suse: Because currently we only found this issue on ARM.

rfan1:
I suggest not using ARM, aarch64 should be better, others LGTM

rfan1 (approved):
LGTM

lib/x11test.pm
lemon-suse: This issue only happened on ARM so we shouldn't enlarge the timeout value for other arches when the value is much larger than the original value, or this will lead to need wait unexpected time when other issue happens for other arches.

PR #13725, opened 2021-11-23
Fix lpstat timeout on Aarch64

hjluo:
On the Aarch64 platform, the lpstat can not be finished in the default
time-out value. We need to extend to make it finish.

- Related ticket:  https://progress.opensuse.org/issues/102395
- Needles: N/A
- Verification run: 
  https://openqa.nue.suse.com/tests/7714940
  https://openqa.nue.suse.com/tests/7725201

### DIFF
--- a/lib/services/cups.pm
+++ b/lib/services/cups.pm
@@ -56,7 +56,7 @@ sub check_function {
     }
     # Add printers
     record_info "lpadmin", "Try to add printers and enable them";
-    validate_script_output 'lpstat -p -d -o 2>&1 || test $? -eq 1', sub { m/lpstat: No destinations added/ };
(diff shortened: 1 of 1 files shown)


tinawang123 (approved):
LGTM

lemon-suse:
Please give the summary for the largest timeout value in here or ticket, so that we will know whether the timeout value is enough or not.   Another thinking about the enlarge timeout for ARM, if the value is much larger than the original value, we'd better add arch limit since this only happened on special arch.

hjluo:
> Please give the summary for the largest timeout value in here or ticket, so that we will know whether the timeout value is enough or not. Another thinking about the enlarge timeout for ARM, if the value is much larger than the original value, we'd better add arch limit since this only happened on special arch.

thanks, already updated in the ticket.

lemon-suse (approved):

PR #13731, opened 2021-11-23
publiccloud: fix mistakes done during creation of main_publiccloud

asmorodskyi:
VRs : 

http://autobot.qa.suse.de/tests/2565
http://autobot.qa.suse.de/tests/2566
http://autobot.qa.suse.de/tests/2567
http://autobot.qa.suse.de/tests/2568
http://autobot.qa.suse.de/tests/2569
http://autobot.qa.suse.de/tests/2570

( all failures can not be related to changes because changes touching only "scheduling part" and loadtest part seems to work correct )

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -43,8 +43,6 @@ sub load_maintenance_publiccloud_tests {
         loadtest("publiccloud/img_proof", run_args => $args);
     } elsif (get_var('PUBLIC_CLOUD_LTP')) {
         loadtest('publiccloud/run_ltp', run_args => $args);
-    } elsif (get_var('PUBLIC_CLOUD_FIO')) {
-        loadtest('publiccloud/storage_perf', run_args => $args);
     } else {
         loadtest "publiccloud/ssh_interactive_start", run_args => $args;
         loadtest "publiccloud/instance_overview" unless get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS');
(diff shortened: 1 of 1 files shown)


jlausuch (approved):

lib/main_publiccloud.pm
```diff
@@ -148,7 +141,7 @@ The rest of the scheduling is divided into two separate subroutines C<load_maint
 =cut
 
 sub load_publiccloud_tests {
-    if (check_var('PUBLIC_CLOUD_TOOLS_REPO', 1)) {
+    if (check_var('PUBLIC_CLOUD_PREPARE_TOOLS', 1)) {
```
grisu48, line 144: Is this correct? Looking at https://openqa.suse.de/tests/7241026#settings and the [PC Tools image job group](https://openqa.suse.de/admin/job_templates/276) this looks wrong to me?

asmorodskyi: if you look closely , you will realize that with current state *any* change in `main_publiccloud`  is correct for https://openqa.suse.de/admin/job_templates/276 because of `YAML_SCHEDULE` variable :) so this Job Group certainly require some MR :) but again due to `YAML_SCHEDULE` variable we are safe to do it independently later

grisu48: Ahhh, I see. And you also solved another bit of headscratching. I still get confused by the mixed usage of YAMLSCHEDULE and `main_$whatever.pm` that we have.

grisu48 (approved):

PR #13732, opened 2021-11-23
Avoid failures due to un-refreshed screen & bsc#1192992 workaround

dzedro:
- Related ticket:
https://progress.opensuse.org/issues/101292
https://progress.opensuse.org/issues/102638
https://progress.opensuse.org/issues/102020
- Verification run:
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=x11 15 SP4
https://openqa.suse.de/tests/overview?distri=sle&version=15&build=x11 15 GA consoletest_finish failures are related to another issue
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP2&build=x11 15 SP2 consoletest_finish failures are related to another issue
https://openqa.opensuse.org/tests/overview?distri=opensuse&build=x11&version=Tumbleweed TW

PR #13735, opened 2021-11-24
Handle additional polkit window in S390x first_boot

zhaocongjia:
In `first_boot` on S390x platform, there might be additional polkit window after login.  Handle them by sending `esc` key.

@jknphy please review when you have time.

- Related ticket: https://progress.opensuse.org/issues/102020
https://progress.opensuse.org/issues/102740
- Needles: none
- Verification run: https://openqa.suse.de/tests/7732608 https://openqa.suse.de/tests/7732609

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -237,6 +237,11 @@ sub handle_login {
         type_password($mypwd);
         send_key 'ret';
         wait_still_screen;
+        # If there are additional polkit windows, use `esc` to discard them.
+        if (check_screen([qw(authentication-required-user-settings authentication-required-modify-system)], 10)) {
+            record_soft_failure("bsc#1192992");
+            send_key_until_needlematch('generic-desktop', 'esc');
+        }
     }
     assert_screen([qw(generic-desktop gnome-activities opensuse-welcome)], 180);
     if (match_has_tag('gnome-activities')) {


jknphy:
Could we run twice this list of verification, or at least some of them, but twice or more each one to know that is stable?
Would this work for openSUSE in s390x? but I'm not sure we have much test coverage there... just guessing due to the next lines following your change.

zhaocongjia:
Second run results: https://openqa.suse.de/tests/7732998 https://openqa.suse.de/tests/7732999

We don't have s390x for opensuse.  And this change won't affect the following code's logic.

jknphy:
should we point to the bug when doing each 'ESC' to not take it as normal behavior? Other than that LGTM, we could merge and I can restart all the failures and see. Thanks.

zhaocongjia:
Okay, I'll make some changes and do another VR.

okurz:
@okurz @Vogtinator @dzedro hi, you might be interested in this change as you previously worked on the changed code

Vogtinator:
Why discard the dialog instead of accepting it, like the other?

> We don't have s390x for opensuse. 

We do.

jknphy:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13732

dzedro:
I had already discussion with @zhaocongjia, I was thinking also about just pressing `esc`.
The behavior does not look correct with successful authentication.
Let see how the bug for it develops https://bugzilla.suse.com/show_bug.cgi?id=1192992

lib/x11utils.pm
```diff
@@ -237,6 +237,11 @@ sub handle_login {
         type_password($mypwd);
         send_key 'ret';
         wait_still_screen;
+        # If there are additional polkit windows, use `esc` to discard them.
+        if (check_screen([qw(authentication-required-user-settings authentication-required-modify-system)], 10)) {
```
okurz, line 241: Please use a multi-tag `assert_screen` with `match_has_tag` instead of `check_screen` with non-zero timeout to prevent introducing any timing dependant behaviour, to save test execution time as well as state more explicitly from the testers point of view what are the expected alternatives. For example:

```perl
assert_screen([qw(yast2_console-finished yast2_missing_package)]);
if (match_has_tag('yast2_missing_package')) {
    send_key 'alt-o';  # confirm package installation
    assert_screen 'yast2_console-finished';
}
```

okurz (changes requested):

zhaocongjia:
As discussed with dzedro offline, we'll use #13732 as the solution for `first_boot`.  Move discussion there.

PR #13737, opened 2021-11-24
Fix disks selection in Clone partition

OleksandrOrlov:
There was a fix of an issue in libyui for YMultiSelectionBox.
Previously, checkboxes may be selected with both 'select' and 'check'
actions, that was incorrect. After the fix, 'select' action highlights
element in list and 'check' action checks a checkbox. Link to the PR
with fix: https://github.com/libyui/libyui/pull/58

The commit updates testing framework to use 'check' action to select
disks in 'Clone partition layout' dialog.

Fixes https://openqa.suse.de/tests/7729237#step/raid_gpt/4

- Verification runs: 
   - staging: https://openqa.suse.de/tests/7736071
   - YaST job group (to confirm that it works on older version of libyui): https://openqa.suse.de/tests/7736072


ge0r (approved):
LGTM!

PR #13738, opened 2021-11-24
Move container registry helpers function to each client

ilausuch:
The container registry helpers behavior depends on the
provider is being testing. So it's necessary to move them
to the provider clients

- Related ticket: https://progress.opensuse.org/issues/101497
- Verification run: 
-- http://copland.arch.suse.de/tests/581 (GKE)
-- http://copland.arch.suse.de/tests/580 (EKS)

### DIFF
--- a/lib/publiccloud/aws_client.pm
+++ b/lib/publiccloud/aws_client.pm
@@ -10,6 +10,7 @@
 package publiccloud::aws_client;
 use Mojo::Base -base;
 use testapi;
+use publiccloud::vault;
 
 use constant CREDENTIALS_FILE => '/root/amazon_credentials';
 
@@ -20,6 +21,7 @@ has region => undef;
 has vault => undef;
 has aws_account_id => undef;
 has service => undef;
+has container_registry => sub { get_var("PUBLIC_CLOUD_CONTAINER_IMAGES_REGISTRY", 'suse-qec-testing') };
 
 sub vault_create_credentials {
     my ($self) = @_;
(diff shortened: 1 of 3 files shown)


jlausuch (approved):

lib/publiccloud/aws_client.pm
```diff
@@ -98,6 +100,26 @@ sub init {
...
+=cut
+sub get_container_registry_prefix {
+    my ($self) = @_;
+    my $region = $self->region;
+    my $full_name_prefix = sprintf('%s.dkr.ecr.%s.amazonaws.com', $self->aws_account_id, $region);
+    return $full_name_prefix;
+}
+
+=head2 get_container_image_full_name
+Get the full name for a container image in ECR registry
```
grisu48, line 114: The `$tag` parameter is missing as possible argument

grisu48 (approved):
LGTM! One nitpicking comment

lib/publiccloud/aws_client.pm
grisu48: I'm thinking about something like adding

    C<tag> Tag of the container to ge t the full name of

Or something similar, analog to e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/containers/engine.pm#L56

ilausuch: Done

grisu48 (approved):

jlausuch (approved):

PR #13742, opened 2021-11-25
Fix usbguard failure on aarch64

lilyeyes:
Fix poo#102921 - [sle][security][sle15sp4] test fails in usbguard on aarch64:
Test usbguard failed on aarch64, after doing "systemctl restart usbguard" at some situations the usb device "keyboard" will be blocked, then add workaround for this.

- Related ticket: https://progress.opensuse.org/issues/102921
- Needles: NA
- Verification run:
  TW: https://openqa.opensuse.org/tests/2057533
  aarch on OSD: https://openqa.suse.de/tests/7746578

### DIFF
--- a/tests/security/usbguard/usbguard.pm
+++ b/tests/security/usbguard/usbguard.pm
@@ -20,6 +20,12 @@ use warnings;
 use testapi;
 use utils;
 
+sub restart_usbguard_allow_keyboard {
+    # When doing restart usbguared service it will block the "USB Keyboard" on aarch64,
+    # so workaround it for all arches just in case (allow it permanently)
+    assert_script_run('systemctl --no-pager start usbguard.service; id=$(usbguard list-devices | grep block | grep Keyboard | cut -f1 -d :); [[ $id ]] && usbguard allow-device -p $id || echo \"No blocked keyboard found\"');
+}
+
 sub run {
     my $out = '';
     my $user = 'usbguard';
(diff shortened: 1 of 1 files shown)


rfan1:
typo: the add -> then add

rfan1:
clocked -> locked? imo

tests/security/usbguard/usbguard.pm
```diff
@@ -20,6 +20,12 @@ use warnings;
 use testapi;
 use utils;
 
+sub restart_usbguard_allow_keyboard {
```
rfan1, line 23: Lili,

Can we consider it as a bug in aarch64 platform, lgtm from my side, but it should be better we can get confirmation from developer

lilyeyes:
> typo: the add -> then add

Done. s/the/then/

tests/security/usbguard/usbguard.pm
lilyeyes: Good question. It is as expected because on aarch openQA environment there is a "keyboard" device and the default rule/policy is set to "block" so when usbguard service takes effect this device will be blocked.
It test cases I did some "rule removing" testings so after restart the usbguard will block "keyboard" too.

rfan1 (approved):
LGTM

Amrysliu (approved):

jouyingbin (approved):
LGTM

PR #13746, opened 2021-11-25
Skip host_configuration for JeOS container tests

jlausuch:
This is just to skip this unnecessary module in JeOS container tests, e.g. https://openqa.opensuse.org/tests/2056841#step/host_configuration/1

Also, skip container_diff for host tests:
We have now split host tests from image tests, so this test doesn't
make sense any more in host tests as it compares released image
from totest image.

VR:
https://openqa.opensuse.org/tests/2056867
https://openqa.opensuse.org/tests/2056868

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -81,7 +81,6 @@ sub load_host_tests_docker {
         loadtest 'containers/docker_compose';
     }
     loadtest 'containers/validate_btrfs' if (is_x86_64 && !is_jeos);
-    loadtest "containers/container_diff" if (is_opensuse());
 }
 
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
> This is just to skip this unnecessary module in JeOS container tests, e.g. https://openqa.opensuse.org/tests/2056841#step/host_configuration/1

It installs the SUSE internal CA cert, so seems necessary for JeOS as well?

> Also, skip container_diff for host tests:

+1 on that

jlausuch:
> 
> It installs the SUSE internal CA cert, so seems necessary for JeOS as well?
> 

yes, but the [VR](https://openqa.opensuse.org/tests/2056867#) shows it's not really needed... I guess because we are not pulling any SLES related image in openSUSE...

Vogtinator:
> > It installs the SUSE internal CA cert, so seems necessary for JeOS as well?
> 
> yes, but the [VR](https://openqa.opensuse.org/tests/2056867#) shows it's not really needed... I guess because we are not pulling any SLES related image in openSUSE...

Correct! It wouldn't have access to the internal network anyway.

mloviska (approved):
Let's wait for the last VR

PR #13747, opened 2021-11-25
console/verify_no_separate_home: Use findmnt to get source devices

Vogtinator:
Use findmnt to query the source device for the given mountpoints and
compare them. lsblk focuses mainly on devices, and with the new MOUNTPOINT(s)
column behaviour, doing the check is harder.

- Related ticket: https://progress.opensuse.org/issues/102795
- Verification run: https://openqa.opensuse.org/tests/2057135 (WIP)

### DIFF
--- a/tests/console/verify_no_separate_home.pm
+++ b/tests/console/verify_no_separate_home.pm
@@ -12,15 +12,14 @@ use base "consoletest";
 use strict;
 use warnings FATAL => 'all';
 use testapi;
+use Test::Assert 'assert_equals';
 
 sub run {
     select_console 'root-console';
 
-    assert_script_run("! lsblk -n | grep '/home'",
-        fail_message => "Fail!\n
-        Expected: /home is NOT on separate partition/volume.\n
-        Actual: /home is on separate partition/volume."
-    );
+    my $root_device = script_output("findmnt -nrvo SOURCE -T /");
+    my $home_device = script_output("findmnt -nrvo SOURCE -T /home");
+    assert_equals($root_device, $home_device, "/home is on a separate partition");
 }
 
 1;


tests/console/verify_no_separate_home.pm
```diff
@@ -16,11 +16,14 @@ use testapi;
...
 
-    assert_script_run("! lsblk -n | grep '/home'",
-        fail_message => "Fail!\n
-        Expected: /home is NOT on separate partition/volume.\n
-        Actual: /home is on separate partition/volume."
-    );
+    my $root_device = script_output("findmnt -nrvo SOURCE /");
+    my $home_device = script_output("findmnt -nrvo SOURCE /home");
+    record_info("Source devices", "/ is on ${root_device}, /home is on ${home_device}");
+    if ($root_device ne $home_device) {
```
OleksandrOrlov, line 22: Hi Fabian! Thank you a lot for the PR! 

Just a minor thing, could you please use here `assert_not_equals` method from [Test::Assert](https://metacpan.org/pod/Test::Assert) library, as we are trying to use it everywhere, where asserts are needed. It allows to make all asserts consistent and provide nice stacktrace in case of failure. e.g. similar to `assert_equals`, that we use here:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/d99fc545dfb161c56a105c509ad26b3d99a4283a/tests/console/validate_lvm_raid1.pm#L49

Do not forget to import `use Test::Assert 'assert_not_equals';`

Vogtinator:
I changed to `assert_equals`.

Minimal verification runs with custom `SCHEDULE`:

MicroOS passes: http://10.160.67.86/tests/1147
MicroOS with induced failure: http://10.160.67.86/tests/1146

tests/console/verify_no_separate_home.pm
Vogtinator: That way the `record_info` is also superfluous, nice!

OleksandrOrlov (approved):
Thank you! Nice solution!

OleksandrOrlov:
Fabian, I approved, but then I noticed that the case with "MicroOS with induced failure: http://10.160.67.86/tests/1146" is passing. Am I right, that it is expected to be failed?

Vogtinator:
> Fabian, I approved, but then I noticed that the case with "MicroOS with induced failure: http://10.160.67.86/tests/1146" is passing. Am I right, that it is expected to be failed?

Yep, I just pasted the wrong link: http://10.160.67.86/tests/1144#step/verify_no_separate_home/8

rakoenig:
Our daily round discovered that `findmnt` only works if there are real mount points. So the soluiton works fine on btrfs with its subvolumes but will fail on ext4 where `/home` is then just a directory below `/`.

Vogtinator:
> Our daily round discovered that `findmnt` only works if there are real mount points. So the soluiton works fine on btrfs with its subvolumes but will fail on ext4 where `/home` is then just a directory below `/`.

Indeed, should be fixed by using the `-T` option. I'll do a VR on ext4.

Vogtinator:
> > Our daily round discovered that `findmnt` only works if there are real mount points. So the soluiton works fine on btrfs with its subvolumes but will fail on ext4 where `/home` is then just a directory below `/`.
> 
> Indeed, should be fixed by using the `-T` option. I'll do a VR on ext4.

btrfs: https://openqa.opensuse.org/tests/2057656
ext4 (rescue CD): https://openqa.opensuse.org/tests/2057659

jknphy (approved):
Nice solution, thanks @Vogtinator, I didn't know how powerful was this command `findmnt`. I was doing in parallel some more hacky solution with `grep` and `grep -v` with `lsblk`. I just tested in my two VMs one with new version one with the older one and your solution works smoothly.
LGTM

Vogtinator:
I :heart: `findmnt`

PR #13748, opened 2021-11-26
Use existing certificate files for ibmtss tests

rfan1:
When test ibmtss binaries installed onto our system,
we can use the existing certificate files. however,
we should define the right working directory.
[We need test aarch64 platform only]

- Related ticket: https://progress.opensuse.org/issues/103086
- Needles: n/a
- Verification run:
 https://openqa.opensuse.org/tests/2057535 -TW
 https://openqa.suse.de/tests/7746606           -SLE

### DIFF
--- a/tests/security/ibmtss/ibmtss_basic_function.pm
+++ b/tests/security/ibmtss/ibmtss_basic_function.pm
@@ -5,7 +5,7 @@
 #          IBM has tested x86_64, s390x and ppc64le, we only need cover aarch64
 #          This test module covers basic function test
 # Maintainer: rfan1 <richard.fan@suse.com>
-# Tags: poo#101088, poo#102792, tc#1769800
+# Tags: poo#101088, poo#102792, poo#103086, tc#1769800
 
 use base 'opensusebasetest';
 use base 'consoletest';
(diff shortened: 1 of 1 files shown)


lilyeyes (approved):
LGTM

jouyingbin:
@rfan1 
I just check the VR and found there is a block with the red line, it looks no harm for the final result, just curious if it is the return value is incorrect? 
https://openqa.suse.de/tests/7746606#step/ibmtss_basic_function/38

tests/security/ibmtss/ibmtss_basic_function.pm
```diff
@@ -48,16 +48,19 @@ sub run {
...
     # Run the script
     assert_script_run('export TPM_INTERFACE_TYPE=socsim');
     my $tsslog = '/tmp/tss.log';
-    script_run("./reg.sh | tee $tsslog", timeout => 120);
+    script_run("./reg.sh -a | tee $tsslog", timeout => 240);
     upload_logs("$tsslog");
 
     # Check the test result
-    assert_script_run("cat $tsslog | grep 'Success - 0 Tests 0 Warnings'");
+    assert_script_run("cat $tsslog | grep 'Success - 35 Tests 0 Warnings'");
```
jouyingbin, line 63: The **35** test is the fixed number from the test, right?  or it is a flexible value while testing?

rfan1: Good question, for now, with emulate tpmserver engine, it will keep 35 as fixed number. so I will not change it as a variable.

rfan1:
> @rfan1 I just check the VR and found there is a block with the red line, it looks no harm for the final result, just curious if it is the return value is incorrect? https://openqa.suse.de/tests/7746606#step/ibmtss_basic_function/38

No error there, seems many lines burst in one session.

tests/security/ibmtss/ibmtss_basic_function.pm
jouyingbin: ok

jouyingbin:
> > @rfan1 I just check the VR and found there is a block with the red line, it looks no harm for the final result, just curious if it is the return value is incorrect? https://openqa.suse.de/tests/7746606#step/ibmtss_basic_function/38
> 
> No error there, seems many lines burst in one session.

got it thanks.

jouyingbin (approved):
LGTM

PR #13750, opened 2021-11-26
Handle dasd activation via libyui-rest-api

ge0r:
- Related ticket: https://progress.opensuse.org/issues/98970
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/overview?build=ge0r%2Fos-autoinst-distri-opensuse%23activate_dasd&distri=sle&version=15-SP4

### DIFF
--- a/lib/Distribution/Opensuse/Tumbleweed.pm
+++ b/lib/Distribution/Opensuse/Tumbleweed.pm
@@ -17,6 +17,7 @@ use Installation::ClockAndTimeZone::ClockAndTimeZoneController;
 use Installation::DiskActivation::DiskActivationController;
 use Installation::DiskActivation::ConfiguredZFCPDevicesController;
 use Installation::DiskActivation::AddZFCPDeviceController;
+use Installation::DiskActivation::DASDDiskManagementController;
 use Installation::LanguageKeyboard::LanguageKeyboardController;
 use Installation::License::Opensuse::Firstboot::LicenseAgreementController;
 use Installation::License::Opensuse::LicenseAgreementController;
@@ -176,6 +177,10 @@ sub get_configured_zfcp_devices {
 
 sub get_add_new_zfcp_device {
     return Installation::DiskActivation::AddZFCPDeviceController->new();
+}
+
+sub get_dasd_disk_management {
+    return Installation::DiskActivation::DASDDiskManagementController->new();
 }
 
 1;
--- a/lib/Installation/DiskActivation/DASDDiskManagementController.pm
+++ b/lib/Installation/DiskActivation/DASDDiskManagementController.pm
@@ -1,0 +1,51 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: This class introduces business actions for DASD Disk Management page
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::DASDDiskManagementController;
+use strict;
+use warnings;
+use Installation::DiskActivation::DASDDiskManagementPage;
+use YuiRestClient;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {}, $class;
+    return $self->init($args);
+}
+
+sub init {
+    my ($self, $args) = @_;
+    $self->{DASDDiskManagementPage} = Installation::DiskActivation::DASDDiskManagementPage->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_dasd_disk_management_page {
+    my ($self) = @_;
+    die "DASD Disk Management page is not displayed" unless $self->{DASDDiskManagementPage}->is_shown();
+    return $self->{DASDDiskManagementPage};
+}
+
+sub filter_channel {
+    my ($self, $channel) = @_;
+    $self->get_dasd_disk_management_page()->enter_minimum_channel($channel);
+    $self->get_dasd_disk_management_page()->enter_maximum_channel($channel);
+    $self->get_dasd_disk_management_page()->press_filter_button();
+}
+
+sub activate_device {
+    my ($self, $channel) = @_;
+    $self->get_dasd_disk_management_page()->select_device($channel);
+    $self->get_dasd_disk_management_page()->perform_action_activate();
+}
+
+sub accept_configuration {
+    my ($self) = @_;
+    $self->get_dasd_disk_management_page()->press_next();
+}
+
+1;
--- a/lib/Installation/DiskActivation/DASDDiskManagementPage.pm
+++ b/lib/Installation/DiskActivation/DASDDiskManagementPage.pm
@@ -1,0 +1,60 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles DASD Disk Management page
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::DiskActivation::DASDDiskManagementPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+    $self->{tbx_minimum_channel} = $self->{app}->textbox({id => 'min_chan'});
+    $self->{tbx_maximum_channel} = $self->{app}->textbox({id => 'max_chan'});
+    $self->{btn_filter} = $self->{app}->button({id => 'filter'});
+    $self->{tbl_devices} = $self->{app}->table({id => 'table'});
+    $self->{mco_operation} = $self->{app}->menucollection({id => 'operation'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{mco_operation}->exist();
+}
+
+sub enter_minimum_channel {
+    my ($self, $channel) = @_;
+    return $self->{tbx_minimum_channel}->set($channel);
+}
+
+sub enter_maximum_channel {
+    my ($self, $channel) = @_;
+    return $self->{tbx_maximum_channel}->set($channel);
+}
+
+sub press_filter_button {
+    my ($self) = @_;
+    return $self->{btn_filter}->click();
+}
+
+sub get_devices {
+    my ($self) = @_;
+    return $self->{tbl_devices}->items();
+}
+
+sub select_device {
+    my ($self, $channel) = @_;
+    return $self->{tbl_devices}->select(value => $channel);
+}
+
+sub perform_action_activate {
+    my ($self) = @_;
+    $self->{mco_operation}->select('Activate');
+}
+
+1;
--- a/schedule/yast/ext4/ext4@yast-s390x-disk-activation.yaml
+++ b/schedule/yast/ext4/ext4@yast-s390x-disk-activation.yaml
@@ -12,7 +12,9 @@ schedule:
   - installation/access_beta_distribution
   - installation/product_selection/install_SLES
   - installation/licensing/accept_license
-  - installation/disk_activation
+  - installation/disk_activation/select_configure_dasd_disks
+  - installation/disk_activation/configure_dasd
+  - installation/disk_activation/finish_disk_activation
   - installation/registration/register_via_scc
   - installation/module_selection/skip_module_selection
   - installation/add_on_product/skip_install_addons
--- a/tests/installation/disk_activation/configure_dasd.pm
+++ b/tests/installation/disk_activation/configure_dasd.pm
@@ -1,0 +1,19 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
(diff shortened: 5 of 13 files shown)


tests/installation/configure_dasd.pm
```diff
@@ -0,0 +1,26 @@
...
+# Summary: Activates multipath when multipath activation message appears
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use parent 'y2_installbase';
+use testapi;
+use strict;
+use warnings;
+
+sub run {
+    my $dasd_channel = "0.0.0150";
```
jknphy, line 15: We could use single quotes if not interpolation is needed, but there are not strict guidelines on that. Anyway given that we don't need to filter every time, which can just pass the value to the function, so we don't need any assignation here.

tests/installation/configure_dasd.pm
```diff
@@ -0,0 +1,26 @@
...
+
+    my $disk_activation_controller = $testapi::distri->get_disk_activation();
```
jknphy, line 17: we omit `_controller` as it is obvious that all that comes from `$testapi::distri` are controllers, and even if some day we decide to merge controllers and pages, we don't need to change tests.

tests/installation/configure_dasd.pm
```diff
@@ -0,0 +1,26 @@
...
+    $disk_activation_controller->configure_dasd_disks();
+    my $dasd_disk_management_controller = $testapi::distri->get_dasd_disk_management();
+    $dasd_disk_management_controller->filter_channel($dasd_channel);
+    $dasd_disk_management_controller->activate_channel($dasd_channel);
```
jknphy, line 21: shouldn't be `activate_disk`? I think we activate a disk, at least it what I can see in that menu, that other option like format affect to disks as well, channel seems to be a property of the disk which identify the disk, but not what is activated. wdyt?

jknphy (commented):
First round of comments, it is looking good!
I (or we :) ) should have spot that filter thingy in refinement, but it is good that we now have all in place to do it later.

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -14,6 +14,10 @@ use warnings FATAL => 'all';
 use parent 'susedistribution';
 use Installation::AuthenticationForRoot::AuthenticationForRootController;
 use Installation::ClockAndTimeZone::ClockAndTimeZoneController;
+use Installation::DiskActivation::DiskActivationController;
+use Installation::DiskActivation::ConfiguredZFCPDevicesController;
+use Installation::DiskActivation::AddZFCPDeviceController;
+use Installation::DiskActivation::DasdDiskManagementController;
```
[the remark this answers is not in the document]
ge0r: I was leaning on it as well, It's just that I found DASDDisk harder to read, but better changed it true

lib/Installation/DiskActivation/DasdDiskManagementController.pm
```diff
@@ -0,0 +1,51 @@
...
+    return $self;
+}
+
+sub get_dasd_disk_management_page {
+    my ($self) = @_;
+    die "DASD Disk Management page is not displayed" unless $self->{DasdDiskManagementPage}->is_shown();
+    return $self->{DasdDiskManagementPage};
+}
+
+sub filter_channel {
```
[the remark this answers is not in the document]
ge0r: I see what you mean, we should select the channel straight away.

lib/Installation/DiskActivation/DasdDiskManagementPage.pm
```diff
@@ -0,0 +1,61 @@
...
+    my ($self, $channel) = @_;
+    return $self->{tbl_channels}->select(value => $channel);
+}
+
+sub perform_activate_action {
+    my ($self) = @_;
+    $self->{menu_operation}->select('Activate');
+}
+
+sub is_shown {
```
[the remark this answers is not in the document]
ge0r: Yes, the thought process was that I need to check the first thing I use, which before the future changes is `tb_minimum_channel` so now it makes sense to make it `tbl_channels`

tests/installation/configure_dasd.pm
```diff
@@ -0,0 +1,26 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Activates multipath when multipath activation message appears
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use parent 'y2_installbase';
+use testapi;
```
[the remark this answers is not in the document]
ge0r: Ah I actually don't need it, I ll delete the line, I had to try something and forgot it!

tests/installation/configure_dasd.pm
ge0r: Yes good point!

tests/installation/configure_dasd.pm
ge0r: Makes sense, I ll rename it!

schedule/yast/ext4/ext4@yast-s390x-disk-activation.yaml
```diff
@@ -12,7 +12,7 @@ schedule:
   - installation/access_beta_distribution
   - installation/product_selection/install_SLES
   - installation/licensing/accept_license
-  - installation/disk_activation
+  - installation/configure_dasd
```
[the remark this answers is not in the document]
ge0r: Done! Btw this filter also includes a zfcp job, but I believe this job's yaml should not include the dasd module, right?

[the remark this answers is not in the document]
jknphy: when I filter by disk_activation I got all DASD, not sure how you get some zfcp, of course we need to exclude ssh-X because it is not libyui-rest-api.

lib/Installation/DiskActivation/DASDDiskManagementController.pm
```diff
@@ -0,0 +1,51 @@
...
+}
+
+sub filter_channel {
+    my ($self, $channel) = @_;
+    $self->get_dasd_disk_management_page()->enter_minimum_channel($channel);
+    $self->get_dasd_disk_management_page()->enter_maximum_channel($channel);
+    $self->get_dasd_disk_management_page()->press_filter_button();
+}
+
+sub activate_disk {
```
jknphy, line 40: s/activate_device/ channel is the search criteria, same for s/select_channel/select_device/

lib/Installation/DiskActivation/DASDDiskManagementController.pm
```diff
@@ -0,0 +1,51 @@
...
+    my ($self, $channel) = @_;
+    $self->get_dasd_disk_management_page()->select_channel($channel);
+    $self->get_dasd_disk_management_page()->perform_activate_action();
```
jknphy, line 43: neatpicking, but it would like to implement the rest of the option IMO it would be more readable `perform_action_<action>` in that order. wdyt?

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
```diff
@@ -0,0 +1,60 @@
...
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+    $self->{tb_minimum_channel} = $self->{app}->textbox({id => 'min_chan'});
+    $self->{tb_maximum_channel} = $self->{app}->textbox({id => 'max_chan'});
+    $self->{btn_filter} = $self->{app}->button({id => 'filter'});
+    $self->{tbl_channels} = $self->{app}->table({id => 'table'});
```
jknphy, line 20: s/tbl_channels/tbl_devices/

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
```diff
@@ -0,0 +1,60 @@
...
+
+package Installation::DiskActivation::DASDDiskManagementPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+    $self->{tb_minimum_channel} = $self->{app}->textbox({id => 'min_chan'});
```
jknphy, line 17: iirc, we agreed `txt_` instead of `tb` in a recent meeting, we need three letters, in that way differentiates better from table `tbl`. For `menu` we need to find one :) `mec` ? `mco` just `men` ? I would just try one (your choice of 3 letter) and when more menu appear we can see if makes sense to remark 'menu' or 'collection' in the abbreviation. wdyt?

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
```diff
@@ -0,0 +1,60 @@
...
+    $self->{tb_maximum_channel} = $self->{app}->textbox({id => 'max_chan'});
+    $self->{btn_filter} = $self->{app}->button({id => 'filter'});
+    $self->{tbl_channels} = $self->{app}->table({id => 'table'});
+    $self->{menu_operation} = $self->{app}->menucollection({id => 'operation'});
+    return $self;
+}
+
+sub enter_minimum_channel {
+    my ($self, $minimum_channel) = @_;
```
jknphy, line 26: should be just `$channel`, when I read it I might think that there is some validation in place.

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
```diff
@@ -0,0 +1,60 @@
...
+}
+
+sub perform_activate_action {
+    my ($self) = @_;
+    $self->{menu_operation}->select('Activate');
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{tb_minimum_channel}->exist();
```
jknphy, line 57: use the table should be more safe, did I mention that before? or even to check how the screen is loaded, sometimes the peform action might be the last because might be contextual and wait for everything else to be loaded. wdyt?

schedule/yast/ssh-x.yaml
```diff
@@ -35,4 +35,4 @@ conditional_schedule:
   disk_activation:
     BACKEND:
       s390x:
-        - installation/disk_activation
+        - installation/disk_activation/configure_dasd
```
jknphy, line 38: not here :)

tests/installation/disk_activation/configure_dasd.pm
```diff
@@ -0,0 +1,23 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Activates multipath when multipath activation message appears
```
jknphy, line 6: accidental copy/paste.

tests/installation/disk_activation/configure_dasd.pm
```diff
@@ -0,0 +1,23 @@
...
+
+use parent 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    my $disk_activation = $testapi::distri->get_disk_activation();
+    $disk_activation->configure_dasd_disks();
+    my $dasd_disk_management = $testapi::distri->get_dasd_disk_management();
+    $dasd_disk_management->filter_channel('0.0.0150');
```
jknphy, line 17: this is clear a different test, could you please create new test module for this line (i.e: /disk_activation/filter_dasd_device.pm` and for example just schedule once, not in all the yamls.
And for the remaining lines one test to access and another one for finish, which I believe it already exists.

schedule/yast/ext4/ext4@yast-s390x-disk-activation.yaml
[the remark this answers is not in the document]
ge0r: Hmmm now that I check those again it seems I got the zfcp job from somewhere else : P

schedule/yast/ssh-x.yaml
ge0r: yy! good catch

tests/installation/disk_activation/configure_dasd.pm
ge0r: yy I had not resolved those from the first time yet!

ge0r: I ll work on this

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
ge0r: Yes, I will use the table, I have not resolved the above comment yet : P

jknphy: sorry, I didn't double check that.

lib/Installation/DiskActivation/DASDDiskManagementPage.pm
ge0r: good point

tests/installation/disk_activation/configure_dasd.pm
ge0r: So as we discussed I created a progress ticket for this https://progress.opensuse.org/issues/103320

jknphy (commented):
We just apply those yaml changes to textmode and xfs test suites to the new files created recently instead of conditional schedule. Other than that, very clean work, LGTM.

PR #13751, opened 2021-11-26
Configurable timeout for prepare_test_data

ilausuch:
Sometimes the amount of data to the test is big and the curl downloading
process is interrupted by the timeout

https://progress.opensuse.org/issues/102569

- Related ticket: https://progress.opensuse.org/issues/103119
- Verification run: https://openqa.suse.de/tests/7748904

### DIFF
--- a/variables.md
+++ b/variables.md
@@ -245,3 +245,4 @@ ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in th
 LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are booting an installed system from the installer medium (for example, a DVD) with the menu option "Boot Linux System" (not "boot From Hard Disk"). This option uses linuxrc.
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
 PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in CSP
+PREPARE_TEST_DATA_TIMEOUT | integer | 300 | Download assets in the prepare_test_data module timeout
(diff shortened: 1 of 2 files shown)


jlausuch:
Since the problem seems to be on Azure only, this can be a good solution so we add this variable to Azure jobs and leave the others with the default timeout.

2 things:
- Rename the variable to `PUBLIC_CLOUD_PREPARE_TEST_DATA_TIMEOUT`, as all the related PC variables have the same prefix.
- Add the variable to https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/variables.md

ilausuch:
About the name, prepare_test_data is used in a lot of tests that are not related with public_cloud, don't think is a good idea

jlausuch:
> About the name, prepare_test_data is used in a lot of tests that are not related with public_cloud, don't think is a good idea

ah ok, sorry, I though it was just PC :)

grisu48 (approved):
LGTM! :rocket: I wasn't aware that this is a problem yet.

variables.md
```diff
@@ -245,3 +245,4 @@ ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in th
 LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are booting an installed system from the installer medium (for example, a DVD) with the menu option "Boot Linux System" (not "boot From Hard Disk"). This option uses linuxrc.
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
 PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in CSP
+PREPARE_TEST_DATA_TIMEOUT | bynber | 300 | Download assets in the prepare_test_data module timeout
```
jlausuch, line 248:
```suggestion
PREPARE_TEST_DATA_TIMEOUT | integer | 300 | Download assets in the prepare_test_data module timeout
```

ilausuch: done, thanks

jlausuch (approved):
LGTM

PR #13757, opened 2021-11-27
Enable bci-tests on Ubuntu and CentOS

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/98183

[VRs](http://fromm.arch.suse.de/tests/overview?distri=sle&version=15-SP3&build=17.8.36&groupid=41&test=bci_on_centos_8_host_docker&test=bci_on_centos_8_host_podman&test=bci_on_ubuntu_20.04_host_docker&test=bci_on_ubuntu_20.04_host_podman)

### DIFF
--- a/tests/containers/bci_prepare.pm
+++ b/tests/containers/bci_prepare.pm
@@ -27,22 +27,29 @@ use testapi;
 
 
 sub packages_to_install {
-    my $host_version = get_required_var('HOST_VERSION');
-    $host_version =~ s/-SP/./;
+    my ($version, $sp, $host_distri) = @_;
     my $arch = get_required_var('ARCH');
-
     # common packages
-    my @packages = ('git-core', 'python3', 'python3-devel', 'gcc');
-    if ($host_version eq "12.5") {
-        push @packages, 'python36-pip';
-    } elsif ($host_version eq '15') {
-        assert_script_run("SUSEConnect -p PackageHub/15/$arch");
-        push @packages, ('go1.10', 'skopeo');
-    } elsif ($host_version =~ /15\./) {
-        # Desktop module is needed for SDK module, which is required for installing go
-        assert_script_run("SUSEConnect -p sle-module-desktop-applications/$host_version/$arch");
-        assert_script_run("SUSEConnect -p sle-module-development-tools/$host_version/$arch");
-        push @packages, ('go', 'skopeo');
+    my @packages = ('git-core', 'python3', 'gcc');
+    if ($host_distri eq 'ubuntu') {
+        push @packages, ('python3-dev', 'python3-pip', 'golang');
+    } elsif ($host_distri eq 'centos') {
+        push @packages, ('python3-devel', 'python3-pip', 'golang');
+    } elsif ($host_distri eq 'sles') {
+        my $version = "$version.$sp";
+        push @packages, 'python3-devel';
+        if ($version eq "12.5") {
+            push @packages, 'python36-pip';
+        } elsif ($version eq '15.0') {
+            # On SLES15 go needs to be installed from packagehub. On later SLES it comes from the SDK module
+            assert_script_run("SUSEConnect -p PackageHub/15/$arch");
+            push @packages, ('go1.10', 'skopeo');
+        } else {
+            # Desktop module is needed for SDK module, which is required for installing go
+            assert_script_run("SUSEConnect -p sle-module-desktop-applications/$version/$arch");
+            assert_script_run("SUSEConnect -p sle-module-development-tools/$version/$arch");
+            push @packages, ('go', 'skopeo');
+        }
     } else {
         die("Host is not supported for running BCI tests.");
     }
@@ -57,9 +64,11 @@ sub run {
     my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
     my $bci_tests_branch = get_var('BCI_TESTS_BRANCH');
 
-    ensure_ca_certificates_suse_installed;
+    my ($version, $sp, $host_distri) = get_os_release;
 
-    my ($running_version, $sp, $host_distri) = get_os_release;
+    # Our Ubuntu and CentOS images have the SUSE_Trust_Root.crt already installed
+    ensure_ca_certificates_suse_installed unless ($host_distri =~ /ubuntu|centos/);
+
     if ($engine eq 'podman') {
         install_podman_when_needed($host_distri);
         install_buildah_when_needed($host_distri);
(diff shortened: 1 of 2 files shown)


dcermak:
Do you have a VR for this?

pdostal (approved):
This looks very nice to me

jlausuch:
> Do you have a VR for this?

Just updated description. Once the "retry" option in tox is ready, I will create another commit in another PR to introduce that.

tests/containers/bci_prepare.pm
```diff
@@ -27,22 +27,28 @@ use testapi;
...
+        push @packages, ('python3-dev', 'python3-pip', 'golang');
+    } elsif ($host_distri eq 'centos') {
+        push @packages, ('python3-devel', 'python3-pip', 'golang');
+    } elsif ($host_distri eq 'sles') {
+        my $version = "$version.$sp";
+        push @packages, 'python3-devel';
+        if ($version eq "12.5") {
+            push @packages, 'python36-pip';
+        } elsif ($version eq '15.0') {
+            assert_script_run("SUSEConnect -p PackageHub/15/$arch");
```
grisu48, line 44:
```suggestion
            # On SLES15 go needs to be installed from packagehub. On later SLES it comes from the SDK module
            assert_script_run("SUSEConnect -p PackageHub/15/$arch");
```

I'd add a short comment to avoid confusion in the future.

tests/containers/bci_prepare.pm
```diff
@@ -57,9 +63,11 @@ sub run {
     my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
     my $bci_tests_branch = get_var('BCI_TESTS_BRANCH');
 
-    ensure_ca_certificates_suse_installed;
+    my ($version, $sp, $host_distri) = get_os_release;
+
+    # Ubuntu and CentOS images have the SUSE_Trust_Root.crt already installed
```
grisu48, line 68:
```suggestion
    # Our Ubuntu and CentOS images have the SUSE_Trust_Root.crt already installed
```

grisu48 (approved):
LGTM! Some suggestion to make the comments clearer, other than that it can fly.

tests/containers/bci_prepare.pm
```diff
@@ -27,22 +27,29 @@ use testapi;
 
 
 sub packages_to_install {
-    my $host_version = get_required_var('HOST_VERSION');
-    $host_version =~ s/-SP/./;
+    my ($version, $sp, $host_distri) = get_os_release;
```
b10n1k, line 30: it seems that `packages_to_install` could use the outer `get_os_release`. wdyt?

lib/containers/common.pm
```diff
@@ -105,18 +105,27 @@ sub install_docker_when_needed {
...
+        } elsif ($host_os eq 'ubuntu') {
+            my $version_id = script_output('(. /etc/os-release && echo $VERSION_ID)');
+            my $ubuntu_repo = "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${version_id}";
+            assert_script_run qq(echo "deb $ubuntu_repo/ /" | tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list);
+            assert_script_run "curl -L $ubuntu_repo/Release.key | apt-key add -";
+            assert_script_run "apt-get update", timeout => 900;
+            assert_script_run "apt-get -y install buildah", timeout => 300;
+        } else {
+            activate_containers_module if $host_os =~ 'sles';
+            zypper_call('in buildah', timeout => 900);
```
b10n1k, line 125: Can you explain why the `timeout` is bigger x3 from ubuntu and x8 in centos? it just looks weird. i guess there is some reason.

tests/containers/bci_prepare.pm
```diff
@@ -73,12 +82,28 @@ sub run {
...
+            assert_script_run("apt-get -y install $pkg", timeout => 300);
+        }
+        assert_script_run('pip3 --quiet install --upgrade pip', timeout => 600);
+        assert_script_run("pip3 --quiet install tox", timeout => 600);
+    } elsif ($host_distri eq 'centos') {
+        assert_script_run("yum update -y --allowerasing", timeout => 600);
+        foreach my $pkg (@packages) {
+            assert_script_run("yum install -y $pkg", timeout => 300);
+        }
+        assert_script_run('pip3 --quiet install --upgrade pip', timeout => 600);
```
b10n1k, line 97: those lines are repeated. What about take them off to a subroutine?

b10n1k (changes requested):
not issues but i suggested some small improvements IMO

lib/containers/common.pm
jlausuch: good question... I copy pasted them from the other calls (docker and centos), but I will put everything to 300 as there is no good reasoning behind...

tests/containers/bci_prepare.pm
jlausuch: Since the pip upgrade for sles is different, I don't think we gain a lot creating a subroutine for 2 lines repeated only twice. I would leave it as it is. If later on we add another host and this repeats 3 times, yes.

tests/containers/bci_prepare.pm
jlausuch: done

ilausuch (approved):

tests/containers/bci_prepare.pm
b10n1k: ok now i am curious what difference `pip3.6` makes. it shouldnt. right? and usually the installation gives you symlinks to the current `pip`, so `pip3` should be enough or..?

jlausuch: There was a reason, it was done some months ago when I enabled BCI tests on SLEs... but I don't remember :D 
I'll try to figure out to give you a proper answer :)

lib/containers/common.pm
```diff
@@ -105,18 +105,27 @@ sub install_docker_when_needed {
...
     if (script_run("which buildah") != 0) {
-        # We may run openSUSE with DISTRI=sle and opensuse doesn't have SUSEConnect
-        activate_containers_module if $host_os =~ 'sles';
-        zypper_call "in @pkgs";
-        record_info('buildah', script_output('buildah info'));
+        if ($host_os eq 'centos') {
+            assert_script_run "dnf -y install buildah", timeout => 300;
+        } elsif ($host_os eq 'ubuntu') {
+            my $version_id = script_output('(. /etc/os-release && echo $VERSION_ID)');
+            my $ubuntu_repo = "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${version_id}";
```
mloviska, line 118: Neat picking, bash variable `$VERSION_ID` could be possibly used in the link if the subshell invocation on L#117 is removed.
In such a case, the L#117 could be just an `assert_script_call '.  /etc/os-release'`

jlausuch: I will actually modify the Ubuntu image to come with this repo in place already, no need to do it here, and also, we don't need to repeat this code in `install_podman_when needed` and here. Moving to `DRAFT` until that is done.

jlausuch:
I created [this PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13765) to isolate the pkg installation from the purpose of this PR, which is just BCI enablement.

pdostal (approved):

PR #13758, opened 2021-11-28
Containers: Test podman and docker with firewall and check connectivity

pdostal:
This pull request:
 * Extends `check_docker_firewall()` also for Podman
 * Introduces `check_container_connectivity()`
 * Introduced `podman_firewall` and `docker_firewall` test modules

This restores 3ac053b introduced in #13666 reverted by #13744 due to bug

- Related ticket: [poo#94132](https://progress.opensuse.org/issues/94132)
- Verification run: The verification runs have been updated: [SLE12-SP3-Docker](http://pdostal-server.suse.cz/tests/12979), [MicroOS-ContainerHost](http://pdostal-server.suse.cz/tests/12976), [JeOS15-SP3-Containers](http://pdostal-server.suse.cz/tests/12977), [SLE15-SP3-Podman](http://pdostal-server.suse.cz/tests/12978), [SLE15-SP3-Docker](http://pdostal-server.suse.cz/tests/12962), [Tumbleweed-Docker](http://pdostal-server.suse.cz/tests/12974), [Tumbleweed-Podman](http://pdostal-server.suse.cz/tests/12963)

### DIFF
--- a/lib/containers/docker.pm
+++ b/lib/containers/docker.pm
@@ -9,7 +9,7 @@
 package containers::docker;
 use Mojo::Base 'containers::engine';
 use testapi;
-use containers::utils qw(registry_url);
+use containers::utils qw(registry_url get_docker_version check_runtime_version container_ip);
 use containers::common qw(install_docker_when_needed);
 use utils qw(systemctl file_content_replace);
 use version_utils qw(get_os_release);
@@ -30,6 +30,28 @@ sub configure_insecure_registries {
     assert_script_run('cat /etc/docker/daemon.json');
     systemctl('restart docker');
     record_info "setup $self->runtime", "deamon.json ready";
+}
+
+sub check_containers_connectivity {
+    record_info "connectivity", "Checking that containers can connect to the host, to each other and outside of the host";
+    my $container_name = 'sut_container';
+
+    # Run container in the background
+    assert_script_run "docker run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
+    my $container_ip = container_ip($container_name, 'docker');
+
+    # Connectivity to host check
+    my $default_route = script_output "docker run " . registry_url('alpine') . " ip route show default | awk \'/default/ {print \$3}\'";
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outisde connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " wget google.com";
+
+    # Kill the container running on background
+    assert_script_run "docker kill $container_name ";
 }
 
 1;
--- a/lib/containers/podman.pm
+++ b/lib/containers/podman.pm
@@ -9,9 +9,10 @@
 package containers::podman;
 use Mojo::Base 'containers::engine';
 use testapi;
-use containers::utils qw(registry_url);
+use containers::utils qw(registry_url container_ip);
 use containers::common qw(install_podman_when_needed);
 use utils qw(file_content_replace);
+use Utils::Systemd 'systemctl';
 use version_utils qw(get_os_release);
 has runtime => "podman";
 
@@ -28,6 +29,29 @@ sub configure_insecure_registries {
     assert_script_run "curl " . data_url('containers/registries.conf') . " -o /etc/containers/registries.conf";
     assert_script_run "chmod 644 /etc/containers/registries.conf";
     file_content_replace("/etc/containers/registries.conf", REGISTRY => $registry);
+}
+
+sub check_containers_connectivity {
+    record_info "connectivity", "Checking that containers can connect to the host, to each other and outside of the host";
+    my $container_name = 'sut_container';
+
+    # Run container in the background
+    assert_script_run "podman pull " . registry_url('alpine');
+    assert_script_run "podman run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
+    my $container_ip = container_ip $container_name, 'podman';
+
+    # Connectivity to host check
+    my $default_route = script_output "podman run " . registry_url('alpine') . " ip route show default | awk \'/default/ {print \$3}\'";
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outside connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " wget google.com";
+
+    # Kill the container running on background
+    assert_script_run "podman kill $container_name";
 }
 
 1;
--- a/tests/containers/docker.pm
+++ b/tests/containers/docker.pm
@@ -28,12 +28,8 @@ use Mojo::Base 'containers::basetest';
 use testapi;
 use utils;
 use containers::common;
-use version_utils qw(is_jeos);
 use containers::utils;
 use containers::container_images;
-use publiccloud::utils;
-
-my $stop_firewall = 0;    # Post-run flag to stop the firewall (failsafe)
 
 sub run {
     my ($self) = @_;
@@ -45,11 +41,8 @@ sub run {
     my $engine = $self->containers_factory('docker');
     test_seccomp();
 
-    if ($self->firewall() eq 'firewalld') {
-        # on publiccloud we need to install firewalld first
-        install_and_start_firewalld() if (is_publiccloud || is_jeos);
-        check_docker_firewall();
-    }
+    # Test the connectivity of Docker containers
+    $engine->check_containers_connectivity();
 
     # Run basic runtime tests
     basic_container_tests(runtime => $engine->runtime);
@@ -58,30 +51,6 @@ sub run {
 
     # Clean container
     $engine->cleanup_system_host();
-}
-
-sub post_fail_hook {
-    my $self = shift;
-    cleanup();
-    $self->SUPER::post_fail_hook;
-}
-
-sub post_run_hook {
-    my $self = shift;
-    cleanup();
-    $self->SUPER::post_run_hook;
-}
-
-sub install_and_start_firewalld() {
-    zypper_call('install firewalld');
-    systemctl('start firewalld');
-    systemctl('restart docker');
-    $stop_firewall = 1;
-}
-
-# must ensure firewalld is stopped, if it is only enabled in this test (e.g. publiccloud test runs)
-sub cleanup() {
-    script_run('systemctl stop firewalld; systemctl restart docker') if $stop_firewall;
 }
 
 1;
--- a/tests/containers/podman.pm
+++ b/tests/containers/podman.pm
@@ -34,6 +34,9 @@ sub run {
     my $dir = "/root/DockerTest";
     my $podman = $self->containers_factory('podman');
 
+    # Test the connectivity of Docker containers
+    $podman->check_containers_connectivity();
+
     # Run basic runtime tests
     basic_container_tests(runtime => $podman->runtime);
     # Build an image from Dockerfile and run it
--- a/tests/containers/podman_firewall.pm
+++ b/tests/containers/podman_firewall.pm
@@ -1,0 +1,67 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: podman firewalld
+# Summary: Test podman with enabled firewall
+# Maintainer: qac team <qa-c@suse.de>
+
+use Mojo::Base 'containers::basetest';
+use testapi;
+use utils 'script_retry';
+use containers::utils qw(registry_url container_ip);
+use Utils::Systemd 'systemctl';
+
+my $stop_firewall = 0;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $podman = $self->containers_factory('podman');
+    my $container_name = 'sut_container';
+
+    # Test firewall only on systems where it's installed
+    die('Firewall is not present.') unless ($self->firewall() == 'firewalld' && script_run('which ' . $self->firewall()) == 0);
+
+    # Start firewall if it was not running before
+    if (script_run('systemctl is-active ' . $self->firewall()) != 0) {
+        systemctl('start ' . $self->firewall());
+        $stop_firewall = 1;
+    }
+
+    # cni-podman0 interface is created when running the first container
+    assert_script_run "podman pull " . registry_url('alpine');
+    assert_script_run "podman run --rm " . registry_url('alpine');
+    validate_script_output('ip a s cni-podman0', sub { /,UP/ });
+
+    # Run container in the background
+    assert_script_run "podman pull " . registry_url('alpine');
+    assert_script_run "podman run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
+
+    # Cheking rules of specific running container
+    validate_script_output("iptables -vn -t nat -L PREROUTING", sub { /CNI-HOSTPORT-DNAT/ });
+    validate_script_output("iptables -vn -t nat -L POSTROUTING", sub { /CNI-HOSTPORT-MASQ/ });
+
+    # Kill the container running on background (this may take some time)
+    assert_script_run "podman kill $container_name ";
+    script_retry "podman ps -q | wc -l | grep 0", delay => 5, retry => 12;
+
+    # Test the connectivity of Podman containers
+    $podman->check_containers_connectivity();
+
+    # Stop the firewall if it was started by this test module
+    systemctl('stop ' . $self->firewall()) if $stop_firewall;
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+
+    # Stop the firewall if it was started by this test module
+    systemctl('stop ' . $self->firewall()) if $stop_firewall;
+
+    $self->SUPER::post_fail_hook;
+}
+
+1;
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -20,7 +20,7 @@ use warnings;
 use version_utils;
 use Mojo::Util 'trim';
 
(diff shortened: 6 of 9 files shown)


asmorodskyi (approved):
LGTM

tests/containers/podman.pm
```diff
@@ -34,6 +39,14 @@ sub run {
     my $dir = "/root/DockerTest";
     my $podman = $self->containers_factory('podman');
 
+    if ($self->firewall == 'firewalld' && !is_microos()) {
+        zypper_call('in ' . $self->firewall()) if (is_publiccloud || is_jeos);
```
sysrich, line 43: Why are we force installing firewalld on public cloud and JeOS images when it's a concious decision to not include firewalld on publiccloud or JeOS Images by default?

Shouldn't we be testing the products as we produce them..nor altering them and in the process testing a different environment than what we give to customers by default?

tests/containers/docker.pm
```diff
@@ -62,26 +66,21 @@ sub run {
...
-    systemctl('start firewalld');
-    systemctl('restart docker');
-    $stop_firewall = 1;
-}
-
 # must ensure firewalld is stopped, if it is only enabled in this test (e.g. publiccloud test runs)
 sub cleanup() {
-    script_run('systemctl stop firewalld; systemctl restart docker') if $stop_firewall;
+    my $firewall = shift;
+    disable_and_stop_service($firewall) if $stop_firewall;
```
sysrich, line 82: Why do we force stop the firewall here even on systems where firewalld has been installed by default and has been started by default? Doesn't this potentially invalidate other tests later on in the run that will assume the default installed firewall is present and running?

sysrich (changes requested):

tests/containers/podman.pm
grisu48: Following this logic we could remove most of our test cases. Just by looking at [Tumbleweed-extra_tests_textmode](https://openqa.opensuse.org/tests/2065318) I see plenty test cases, where we first install the package and then run the regression test on it (`git`, `java`, `golang`, `sqlite3` just to name a few).

The choice to not ship `firewalld` by default on publiccloud is more related to the possibility on using the cloud-service provider firewall. This **choice** however does not mean that we can assume that nobody would use `firewalld`. In addition, `firewalld` and the container runtimes are supported packages, and the reason for not including them by default have nothing to do with the fact, that this is a legitimate and expected use case of our customers.

Also: we already had one docker-firewalld related [regression last year](https://bugzilla.suse.com/show_bug.cgi?id=1180401). IMHO is this a relevant and legitimate test case.

@pdostal was only fulfilling the requirement of [poo#94132](https://progress.opensuse.org/issues/94132), so I believe those concerns would be better addresses on the level of the PO @jlausuch

tests/containers/docker.pm
grisu48: But this is only executed in the case, that `firewalld` is installed, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/2a4f67b08e47ae96a1a5c9f6bc2ff092ab03c19b/tests/containers/docker.pm#L53

lib/containers/docker.pm
```diff
@@ -32,4 +33,56 @@ sub configure_insecure_registries {
...
+    # Run container in the background
+    assert_script_run "docker run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine');
+    my $container_ip = container_ip($container_name, 'docker');
+
+    # Each running container should have added a new entry to the DOCKER zone.
+    # ex. output: "1           0        0 ACCEPT     tcp  --  !docker0 docker0  0.0.0.0/0            172.17.0.2           tcp dpt:1234"
+    validate_script_output "iptables -L DOCKER -nvx --line-numbers", sub { /1.+ACCEPT.+!docker0 docker0.+$container_ip\s+tcp dpt:1234/ };
+
+    # Kill the container running on background
+    assert_script_run "docker kill $container_name ";
```
grisu48, line 63: Killing a container which does not exists (you use the `--rm` parameter before) will result in a failure:

    $ docker run --rm -ti --name hale-bopp opensuse/tumbleweed
    418a496d900f:/ # exit
    $ docker kill hale-bopp
    Error response from daemon: Cannot kill container: hale-bopp: No such container: hale-bopp

So either add a check if the container exists or change to `script_run` instead.

lib/containers/docker.pm
```diff
@@ -32,4 +33,56 @@ sub configure_insecure_registries {
...
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outisde connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " wget google.com";
+
+    # Kill the container running on background
+    assert_script_run "docker kill $container_name ";
```
grisu48, line 85: Same as above

lib/containers/podman.pm
```diff
@@ -30,4 +31,50 @@ sub configure_insecure_registries {
...
+    # Run container in the background
+    assert_script_run "podman pull " . registry_url('alpine');
+    assert_script_run "podman run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine');
+
+    # Cheking rules of specific running container
+    validate_script_output("iptables -vn -t nat -L PREROUTING", sub { /CNI-HOSTPORT-DNAT/ });
+    validate_script_output("iptables -vn -t nat -L POSTROUTING", sub { /CNI-HOSTPORT-MASQ/ });
+
+    # Kill the container running on background
+    assert_script_run "podman kill $container_name";
```
grisu48, line 54: Same as above

lib/containers/podman.pm
```diff
@@ -30,4 +31,50 @@ sub configure_insecure_registries {
...
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outside connectivity check
+    assert_script_run "podman run --rm " . registry_url('alpine') . " wget google.com";
+
+    # Kill the container running on background
+    assert_script_run "podman kill $container_name";
```
grisu48, line 77: Same as above

tests/containers/docker.pm
```diff
@@ -45,11 +46,14 @@ sub run {
     my $engine = $self->containers_factory('docker');
     test_seccomp();
 
-    if ($self->firewall() eq 'firewalld') {
-        # on publiccloud we need to install firewalld first
-        install_and_start_firewalld() if (is_publiccloud || is_jeos);
-        check_docker_firewall();
+    if ($self->firewall() == 'firewalld' && !is_microos()) {
```
grisu48, line 49: This check is a bit problematic. Because you would install `firewalld` only on publiccloud and jeos, but stop it on all flavors. I'd recommend to or move the `(is_publiccloud || is_jeos)` check to the upper `if` block or replace this check by a general check, if `firewalld` needs to be installed, e.g.

```suggestion
    if ($self->firewall() == 'firewalld' && !is_microos() && script_run('rpm -q firewalld') != 0) {
```

Otherwise we might end up at some point in a weird configuration.

grisu48 (commented):
I only see some minor issues, that should be addresses or discussed. All in all I think this is a valuable addition, good job Pavel!

tests/containers/docker.pm
grisu48: I also think this is the source of confusion for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13758#discussion_r760308575

tests/containers/docker.pm
grisu48: Ahhhh, I now understand what you mean. See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13758#discussion_r760380702, I think the source of confusion is elsewhere.

tests/containers/podman.pm
sysrich: I would have no problem with adding firewalld and testing it IN ADDITION to the default-of-no-firewall (like we do in extra tests), but this submission as it is written provides zero test coverage for networking issues on publiccloud or Jeos with no firewall, despite public cloud and Jeos with no firewall being the default

that’s not good and my concerns remain

tests/containers/docker.pm
sysrich: Indeed - we declare the variable to stop the firewall regardless of whether it’s the (IMHO also problematic) scenarios where we manually added a firewall

we should only stop the firewall on the systems that don’t normally have one, and we should probably uninstall it also, and we should probably do all that only in addition to testing the default configuration

tests/containers/podman.pm
asmorodskyi: tell you the true I was checking this PR purely from perspective of "Containers" . The fact that we blindly taking everything written for "Containers" and running it for "Containers in Public Cloud" was not taken into consideration when I was checking this PR :( . Thanks @sysrich  for highlight this part. I second all what you are saying . I would prefer that we will exclude firewalld testing in case of "Containers in Public Cloud"

lib/containers/docker.pm
jlausuch: actually, it works if you use Alpine. I guess the key is in the ENTRYPOINT. In your example, it executes "bash", so it enters the container. If you do the same with Alpine, it keeps it running.

asmorodskyi (changes requested):
After reading @sysrich comments I realize that I was not taken Public Cloud into consideration during review and my previous LGTM . And this change my statement see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13758/files#r760580565

lib/containers/docker.pm
```diff
@@ -32,4 +33,56 @@ sub configure_insecure_registries {
...
+
+    # Connectivity to host check
+    my $default_route = script_output "docker run " . registry_url('alpine') . " ip route show default | awk \'/default/ {print \$3}\'";
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $default_route;
+
+    # Cross-container connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outisde connectivity check
+    assert_script_run "docker run --rm " . registry_url('alpine') . " wget google.com";
```
jlausuch, line 82:
```suggestion
    assert_script_run "docker run --rm " . registry_url('alpine') . " ping -c3 google.com";
```

tests/containers/podman.pm
asmorodskyi: and this actually something which was already discussed and concluded that we don't want to run firewalld in public cloud . see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13472

jlausuch: I don't have a strong opinion about PC, but if the decision was to skip firewalld related topics, we should probably skip that.

For _container host_ test this PR is very much valuable. Container host testing is executed on SLE/openSUSE for both flavors, DVD and JeOS, and we are already covering non-firewall-related tests. Adding this firewall and connectivity check for me it's a plus in these flavors. SLE Micro could also benefit from this change, as firewalld comes pre-installed.

However, I could agree about uninstalling the firewall if it wasn't present by default..

jlausuch:
This adds extra value to our testing efforts in containers. Good job Pavel.
Let's see if we can reach an agreement about the open discussions.

lib/containers/docker.pm
pdostal: This does not work on my personal instance. I can do this change and test it on O3 or OSD if you persist but I consider this approach safer.

lib/containers/docker.pm
pdostal: Yes, the container runs in the background as the comment suggests but I will add a check for it.

lib/containers/docker.pm
jlausuch: Ah ok, it was just to follow the same test as in previous commands, so we test icmp port/protocol.

But now that I think it twice, it would be better to reach some common tcp port (e.g. http) in the outside world, so `wget` sounds ok.

Vogtinator:
I agree with @sysrich, the default configuration of the container runtime needs to be tested at first, then additional combinations on top can follow, as long as those combinations don't break later tests (e.g. by leaving the firewall installed and running).

BTW: JeOS has a firewall by default, except for OpenStack.

lib/containers/docker.pm
```diff
@@ -32,4 +33,56 @@ sub configure_insecure_registries {
     record_info "setup $self->runtime", "deamon.json ready";
 }
 
+sub check_containers_firewall {
+    record_info "firewall", "Checking that firewall is enabled and properly configured";
+    my $container_name = 'sut_container';
+    my $docker_version = get_docker_version();
+    systemctl('is-active firewalld');
```
Vogtinator, line 40: IMO it's not obvious that this asserts that `firewalld` is running.

Could be made more obvious by adding a comment or specifying that explicitly:

```suggestion
    systemctl('is-active firewalld', ignore_failure => 0);
```

lib/containers/docker.pm
```diff
@@ -32,4 +33,56 @@ sub configure_insecure_registries {
...
+    my $running = script_output qq(docker ps -q | wc -l);
+    validate_script_output('ip a s docker0', sub { /state DOWN/ }) if $running == 0;
+    # Docker zone is created for docker version >= 20.10 (Tumbleweed), but it
+    # is backported to docker 19 for SLE15-SP3 and for Leap 15.3
+    if (check_runtime_version($docker_version, ">=20.10") || is_sle('>=15-sp3') || is_leap(">=15.3")) {
```
Vogtinator, line 45: Sounds like you can just do

```suggestion
    unless (is_sle('<15-sp3') || is_leap("<15.3")) {
```

lib/containers/docker.pm
Vogtinator: The difference is the `-it` vs. `-id`. The `-d` makes it run in the background. There shouldn't be a difference between alpine or TW there.

Vogtinator (changes requested):
I wonder whether it's time for a `docker_firewall` resp. `podman_firewall` module. It looks like it's pretty much separate anyway, and this would allow a more custom schedule for the products as well.

pdostal:
List of today's changes: 
* The firewall is tested only when it is already installed on the host.
 * If firewall is stopped by default it gets started only for the firewall test phase.
 * The `post_run_hook()` is not needed because of the previous point.
 * The `post_fail_hook()` is not needed either as in case of failure the VM is rolled back.

tests/containers/podman.pm
pdostal: Can you please re-check the code?

lib/containers/docker.pm
pdostal: I added `sleep 30d` to the `podman run` command so it's more visible that it is running in the background. Can you please re-check?

tests/containers/docker.pm
pdostal: This should be also fixed now. Please check.

tests/containers/docker.pm
pdostal: Please can you re-check?

lib/containers/docker.pm
grisu48: Now it makes sense. Since this is a cleanup, I'd make it a `script_run` instead of an `assert_script_run`, however this is up to you.

grisu48 (approved):
LGTM.

pdostal:
@Vogtinator I did create `docker_firewall.pm` and `podman_firewall.pm`.

tests/containers/podman_firewall.pm
```diff
@@ -0,0 +1,55 @@
...
+
+    # Cheking rules of specific running container
+    validate_script_output("iptables -vn -t nat -L PREROUTING", sub { /CNI-HOSTPORT-DNAT/ });
+    validate_script_output("iptables -vn -t nat -L POSTROUTING", sub { /CNI-HOSTPORT-MASQ/ });
+
+    # Kill the container running on background (this may take some time)
+    assert_script_run "podman kill $container_name ";
+    script_retry "podman ps -q | wc -l | grep 0", delay => 5, retry => 12;
+
+    # Test the connectivity of Docker containers
```
Vogtinator, line 46:
```suggestion
    # Test the connectivity of podman containers
```

Vogtinator (approved):
Just a nitpick, LGTM otherwise. I don't think the VR links are recent enough, so I'll just assume it works.

AFAICT the `always_rollback` shouldn't be necessary anymore and could be replaced with stopping the firewall like before.

Vogtinator:
I wonder whether we run those tests on zVM or real HW, then `always_rollback` would break those.

asmorodskyi (approved):
LGTM

pdostal:
> I wonder whether we run those tests on zVM or real HW, then `always_rollback` would break those.

Maybe @jlausuch knows? If so then we need to clean up the test and reintroduce the `post_fail_hook()`.

tests/containers/podman_firewall.pm
pdostal: Ready to review

pdostal:
> AFAICT the `always_rollback` shouldn't be necessary anymore and could be replaced with stopping the firewall like before.

@Vogtinator done.

Vogtinator:
> @Vogtinator done. The tests are now stopping the firewall in case it has been started during the test.

I think you might have forgotten to `git push`

pdostal:
Pushed. I'm still running verification runs. My machine seems to be somehow slow last 2 days.

pdostal:
The verification runs have been updated: [SLE12-SP3-Docker](http://pdostal-server.suse.cz/tests/12979), [MicroOS-ContainerHost](http://pdostal-server.suse.cz/tests/12976), [JeOS15-SP3-Containers](http://pdostal-server.suse.cz/tests/12977), [SLE15-SP3-Podman](http://pdostal-server.suse.cz/tests/12978), [SLE15-SP3-Docker](http://pdostal-server.suse.cz/tests/12962), [Tumbleweed-Docker](http://pdostal-server.suse.cz/tests/12974), [Tumbleweed-Podman](http://pdostal-server.suse.cz/tests/12963)

Vogtinator (approved):
LGTM, great job!

jlausuch:
> > I wonder whether we run those tests on zVM or real HW, then `always_rollback` would break those.
> 
> Maybe @jlausuch knows? If so then we need to clean up the test and reintroduce the `post_fail_hook()`.

No that I know. We use zKVM and powerKVM for all host tests and I don't think we'll switch to real HW...

jlausuch:
Great work Pavel!

PR #13760, opened 2021-11-29
Use existing certificate files for ibmtss tests

rfan1:
When test ibmtss binaries installed onto our system,
we can use the existing certificate files. however,
we should define the right working directory.
BTW, we don't need build to test the packaged ibmtss.

- Related ticket: https://progress.opensuse.org/issues/103086
- Needles: n/a
- Verification run: 
https://openqa.suse.de/tests/7748927              -SLE
https://openqa.opensuse.org/tests/2058119    -TW


lilyeyes (approved):
lgtm

jouyingbin (approved):
LGTM

PR #13762, opened 2021-11-29
Adapt validate_blockdevices to newer version of lsblk

jknphy:
- Related ticket: [poo#102795](https://progress.opensuse.org/issues/102795)
- Verification run: [SLE](https://openqa.suse.de/tests/overview?distri=sle&build=jknphy%2Fos-autoinst-distri-opensuse%23adapt_lsblk&version=15-SP4) | [openSUSE](https://openqa.opensuse.org/tests/overview?build=jknphy%2Fos-autoinst-distri-opensuse%23adapt_lsblk&distri=opensuse)

### DIFF
--- a/lib/filesystem_utils.pm
+++ b/lib/filesystem_utils.pm
@@ -33,7 +33,8 @@ our @EXPORT = qw(
   get_used_partition_space
   lsblk_command
   validate_lsblk
-  get_partition_table_via_blkid);
+  get_partition_table_via_blkid
+  is_lsblk_able_to_display_mountpoints);
 
 =head2 str_to_mb
 
@@ -307,6 +308,26 @@ sub get_used_partition_space {
     return df_command({partition => $partition, column => '5'});
 }
 
+=head2 is_lsblk_able_to_display_mountpoints
+
+ is_lsblk_able_to_display_mountpoints();
+
+Runs utility C<lsblk> using flag --help to retrieve to check for one specific
+column 'MOUNTPOINTS'.
+
+From version util-linux-systemd-2.37 lsblk include new column MOUNTPOINTS
+with all the mountpoints including all the subvolumes in a btrfs system
+and in this case existing column MOUNTPOINT does not contain the primary
+mountpoint for the partition, the first mounted, instead contains the most
+recent one mounted. Therefore we need to workaround this issue.
+Please check bsc#1192996 for further info.
+
+=cut
+
+sub is_lsblk_able_to_display_mountpoints {
+    return script_run('lsblk --help | grep MOUNTPOINTS') == 0;
+}
+
 =head2 lsblk_command
 
  my $json = lsblk_command(output => $output);
@@ -328,7 +349,8 @@ sub lsblk_command {
 
 =head2 create_lsblk_validation_test_data
 
- my $validation_test_data = create_lsblk_validation_test_data($dev);
+ my $validation_test_data = create_lsblk_validation_test_data(
+     device => $dev, has_mountpoints_col => 1);
 
 Converts test data to test data adapted for validation using C<lsblk>.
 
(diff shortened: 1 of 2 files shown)


lib/filesystem_utils.pm
```diff
@@ -307,6 +308,26 @@ sub get_used_partition_space {
...
+with all the mountpoints including all the subvolumes in a btrfs system
+and in this case existing column MOUNTPOINT does not contain the primary
+mountpoint for the partition, the first mounted, instead contains the most
+recent one mounted. Therefore we need to workaround this issue.
+Please check bsc#1192996 for further info.
+
+=cut
+
+sub is_lsblk_able_to_display_mountpoints {
+    return script_run('! lsblk --help | grep MOUNTPOINTS');
```
Vogtinator, line 328: IMO easier to understand:

```suggestion
    return script_run('lsblk --help | grep MOUNTPOINTS') == 0;
```

lib/filesystem_utils.pm
```diff
@@ -430,7 +463,18 @@ sub validate_lsblk {
     }
 
     for my $col (keys %{$validation_test_data}) {
-        if ($validation_test_data->{$col}{value} ne $blockdev->{$col}) {
+        if ($has_mountpoints_col && $col eq 'mountpoints') {
```
Vogtinator, line 466: `$col` can only ever be `mountpoints` if `$has_mountpoints_col` is true.

```suggestion
        if ($col eq 'mountpoints') {
```

lib/filesystem_utils.pm
jknphy: right, specially if it is a pipeline.

lib/filesystem_utils.pm
jknphy: that's right I had it like that and in some intermediate code made sense but not anymore,thanks.

lib/filesystem_utils.pm
```diff
@@ -430,7 +463,18 @@ sub validate_lsblk {
     }
 
     for my $col (keys %{$validation_test_data}) {
-        if ($validation_test_data->{$col}{value} ne $blockdev->{$col}) {
+        if ($col eq 'mountpoints') {
+            my ($primary_mountpoint) = grep { /^\/$/ } @{$blockdev->{$col}};
```
Vogtinator, line 467: There is no such thing as a "primary mountpoint" and FWICT this is just a hack to treat `/` specially.

IMO this should check whether `$blockdev->{$col}` contains `$validation_test_data->{$col}{value}`, or if the latter is an array, all of them.

jknphy: right, there is more straightforward to extract. The array case I will keep it out, we don't have such cases afik and the data structure doesn't allow to use it, but it is a good point to remember, in the future we might change this validation and even use findmount as mountpoints is not very useful, if instead of showing us the first mounted is showing the last which could be any of the subvolume, is not primary but the one that is specified during installation.

Vogtinator: Yup, sounds good.

lib/filesystem_utils.pm
```diff
@@ -430,7 +463,17 @@ sub validate_lsblk {
     }
 
     for my $col (keys %{$validation_test_data}) {
-        if ($validation_test_data->{$col}{value} ne $blockdev->{$col}) {
+        if ($col eq 'mountpoints') {
+            my $col_value = $validation_test_data->{$col}{value};
+            my ($mountpoint) = grep { /^$col_value$/ } @{$blockdev->{$col}};
```
Vogtinator, line 468: Probably needs escaping here

Vogtinator (changes requested):
VR fails: https://openqa.opensuse.org/tests/2062940#step/validate_blockdevices/11

```
List of mountpoints for blockdevice /dev/vda3. does not contain expected mountpoint. Expected: 'mount_point: [SWAP]', got: 'mountpoints: [SWAP]
```

lib/filesystem_utils.pm
jknphy: was that, thanks.

Vogtinator (approved):

PR #13764, opened 2021-11-29
Update Windows image weekly

mloviska:
Based on story [[qac][research][wsl] update windows image in openqa](https://progress.opensuse.org/issues/60341), we need to update our base image weekly or bi-weekly at least.
The 3rd party tools are not covering the whole system base updates or
they do not work reliably enough, therefore simply using `control update`
should be good enough for our needs.

- Needles:

##### VRs:

* [Update with required restart](http://kepler.suse.cz/tests/9075#)
* [Minor update](http://kepler.suse.cz/tests/9267#step/update_windows/6)
* [Up to date image](http://kepler.suse.cz/tests/9268#)


jlausuch (approved):
Neat! Simple and effective!

PR #13765, opened 2021-11-29
Containers: update install_X_when_needed methods

jlausuch:
- Unify timeouts to 900 (update) and 300 (installation)
- Remove repo addition in Ubuntu, as it's already present by
default in our images used by the tests


lib/containers/common.pm
```diff
@@ -69,21 +66,18 @@ sub install_docker_when_needed {
             if ($host_os eq 'centos') {
                 assert_script_run "dnf config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo";
                 # if podman installed use flag "--allowerasing" to solve conflicts
-                assert_script_run "dnf -y install docker-ce --nobest --allowerasing", timeout => 120;
+                assert_script_run "dnf -y install docker-ce --nobest --allowerasing", timeout => 300;
```
dcermak, line 69: Out of curiosity, why are you installing docker from docker's official repo and not from the CentOS repository? I would expect that the usual user would just `dnf install moby-engine` instead of adding an extra repository.

jlausuch: It comes from the official documentation...
https://docs.docker.com/engine/install/centos/

And it's very problematic, as docker and podman can't be together in Centos since they have some dependency issues in containerd packages... very ugly... I would personally disable testing CentOS+docker... but we can't.

grisu48 (approved):
LGTM :rocket:

lib/containers/common.pm
```diff
@@ -39,13 +39,10 @@ sub install_podman_when_needed {
     my @pkgs = qw(podman);
     if (script_run("which podman") != 0) {
         if ($host_os eq 'centos') {
-            assert_script_run "dnf -y install @pkgs", timeout => 160;
+            assert_script_run "dnf -y update", timeout => 900;
+            assert_script_run "dnf -y install podman", timeout => 300;
```
pdostal, line 43:
```suggestion
            assert_script_run "dnf -y install @pkgs", timeout => 300;
```

pdostal (approved):

lib/containers/common.pm
jlausuch: done

jlausuch:
@Vogtinator with this PR, we also fix [this issue](https://openqa.opensuse.org/tests/2061746)...

lib/containers/common.pm
b10n1k: i think @dcermak is right. the docker is `moby-engine` now.

lib/containers/common.pm
```diff
@@ -105,18 +99,24 @@ sub install_docker_when_needed {
...
 sub install_buildah_when_needed {
     my $host_os = shift;
-    my @pkgs = qw(buildah);
     if (script_run("which buildah") != 0) {
-        # We may run openSUSE with DISTRI=sle and opensuse doesn't have SUSEConnect
-        activate_containers_module if $host_os =~ 'sles';
-        zypper_call "in @pkgs";
-        record_info('buildah', script_output('buildah info'));
+        if ($host_os eq 'centos') {
+            assert_script_run "dnf -y update", timeout => 900;
```
b10n1k, line 109: i am not sure but this will run twice in some cases, right? somewhere i have seen `install podman` and then `install buildah`. i wonder if we could do better here. But i dont think it is a blocker too.

jlausuch: yeah, that's true, but it won't harm either :)

lib/containers/common.pm
jlausuch: I have tried, CentOS doesn't come with moby-engine in default repos. 
I leave it up for investigation and see if it's worth it. Will create ticket for that, since it might be a good move.

ilausuch (approved):

lib/containers/common.pm
dcermak: @b10n1k It is the case for Fedora.
@jlausuch sorry, I looked over pkgs.org too quickly. `moby-engine` is available for centos, but only from Microsoft's repository. So forget about my suggestion.

PR #13766, opened 2021-11-29
Public Cloud: Increase boot disk size

pdostal:
- Related ticket: [poo#102807](https://progress.opensuse.org/issues/102807)
- Bug example: [sle-15-SP3-EC2-Containers](https://openqa.suse.de/tests/7713114)
- Verification run: [EC2](http://pdostal-server.suse.cz/tests/12889#) [GCE](http://pdostal-server.suse.cz/tests/12887) [Azure](http://pdostal-server.suse.cz/tests/12888)

### DIFF
--- a/data/publiccloud/terraform/ec2.tf
+++ b/data/publiccloud/terraform/ec2.tf
@@ -102,6 +102,11 @@ resource "aws_instance" "openqa" {
             openqa_created_date = timestamp()
             openqa_created_id = element(random_id.service.*.hex, count.index)
         }, var.tags)
+
+    ebs_block_device {
+        device_name = "/dev/sda1"
+        volume_size = 20
+    }
 }
 
 resource "aws_volume_attachment" "ebs_att" {
--- a/data/publiccloud/terraform/gce.tf
+++ b/data/publiccloud/terraform/gce.tf
@@ -100,6 +100,7 @@ resource "google_compute_instance" "openqa" {
         device_name = "${var.name}-${element(random_id.service.*.hex, count.index)}"
         initialize_params {
             image = var.image_id
+            size  = 20
         }
     }
     


grisu48 (approved):
Simple and nice. LGTM :rocket: I'd like to see a VR for one of the previously failing runs before merging

data/publiccloud/terraform/azure.tf
```diff
@@ -185,6 +185,7 @@ resource "azurerm_linux_virtual_machine" "openqa-vm" {
     name              = "${var.name}-${element(random_id.service.*.hex, count.index)}-osdisk"
     caching              = "ReadWrite"
     storage_account_type = "Standard_LRS"
+    disk_size_gb         = 30
```
b10n1k, line 188: where the 30G comes from? what is the default? looking at the gce i would say that it is 10G. Which 30 makes it too much IMO. Can you check?

jlausuch: 10GB was the default and we saw it's not enough :)

pdostal: Right, decreeing to `20G`.

b10n1k: i still see `30`

pdostal: Pushed now. However your comment applies only to GCE and EC2 as in Azure the default and smallest possible boot disk size is *30G*.

pdostal:
In *Azure* - the smallest image allowed is *30G8.*:
```
Error: creating Linux Virtual Machine "pdostal-server-c1d2606f7bc71730" (Resource Group "pdostal-server-c1d2606f7bc71730"): compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="OperationNotAllowed" Message="The specified disk size 20 GB is smaller than the size of the corresponding disk in the VM image: 30 GB. This is not allowed. Please choose equal or greater size or do not specify an explicit size." Target="osDisk.diskSizeGB"
```
So *GCE* and *EC2* are now set to *20G* and *Azure* is set to *30G*.

jlausuch:
Cool! 
What do you think about adding a `record_info` with output of `df -h` somewhere, maybe in instance_overview?

pdostal:
> Cool! What do you think about adding a `record_info` with output of `df -h` somewhere, maybe in instance_overview?

I for now excluded Azure and am now about to go out to get some fresh air.

The Google and Amazon are ready to be merged.

I'll continue tomorrow.

jlausuch (approved):

grisu48 (approved):

asmorodskyi:
too late but still LGTM :)

PR #13776, opened 2021-11-30
Add patch_sle after installation finished on zVM

lemon-suse:
We need add patch_sle after installation finished on zVM to do update and de-register modules to make migration process run well on z/VM.

- Related ticket: https://progress.opensuse.org/issues/103155
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/7761758#step/start_install/2  #the failure is for bsc#1193215

### DIFF
--- a/lib/migration.pm
+++ b/lib/migration.pm
@@ -129,7 +129,7 @@ sub deregister_dropped_modules {
 # https://documentation.suse.com/sles/15-SP2/html/SLES-all/cha-upgrade-online.html#sec-upgrade-online-zypper
 sub disable_installation_repos {
     if (is_s390x) {
-        zypper_call "mr -d `zypper lr -u | awk '/ftp:.*?openqa.suse.de|10.160.0.100/ {print \$1}'`";
+        zypper_call "mr -d `zypper lr -u | awk '/ftp:.*?openqa.suse.de|10.160.0.207|10.160.0.100/ {print \$1}'`";
     }
     else {
         zypper_call "mr -d -l";
--- a/schedule/install/sles15_s390_install.yaml
+++ b/schedule/install/sles15_s390_install.yaml
@@ -1,0 +1,47 @@
+name:           sles15_s390_install
+description:    >
+    This is prepare install task before migration.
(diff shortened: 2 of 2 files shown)


dzedro:
How is the VR related when it uses another YAML and does not run patch_sle ?

lemon-suse:
> How is the VR related when it uses another YAML and does not run patch_sle ?
In fact, for the test if without patch_sle then migration will report error that we need de-register LTSS firstly.
Below you can see the failure:
https://openqa.nue.suse.com/tests/7762317#step/upgrade_select/4

lemon-suse:
> How is the VR related when it uses another YAML and does not run patch_sle ?

In fact, for this case we need to use YAML to run patch_sle, if not, the migration process will fail for the LTSS hasn't been de-registered and will report 'no migration target found'.

Just like this https://openqa.nue.suse.com/tests/7762317#step/upgrade_select/4

> > How is the VR related when it uses another YAML and does not run patch_sle ?
> > In fact, for the test if without patch_sle then migration will report error that we need de-register LTSS firstly.
> > Below you can see the failure:
> > https://openqa.nue.suse.com/tests/7762317#step/upgrade_select/4

And we do have such test to run without yaml file and follow the process in main.pm, https://openqa.nue.suse.com/tests/7761642# We can continue use this case for test without LTSS and some modules need be de-registered before migration.

lemon-suse:
Updated the new verification log:
https://openqa.nue.suse.com/tests/7766970#  SLES15SP1 -> SLES15SP4  the failure is for bsc#1193215
https://openqa.nue.suse.com/tests/7766967#  SLES15SP3 -> SLES15SP4

dzedro (approved):
Not sure what offline means, I guess things changed since the test was created `Summary: Patch SLE qcow2 images before migration (offline)`
Good, when it works for you.

lemon-suse:
Thanks dzedro, I would like to explain that this is a chained job on z/VM, the PR is to change the first job to install, while won't touch the second job to migration. Then I will merge it. Thanks.

PR #13780, opened 2021-11-30
Public Cloud: Specify 30G boot disk for Azure

pdostal:
Our SLES images are usually 30G by default:

```
Error: Code="OperationNotAllowed" Message="The specified disk size 20 GB is smaller than the size of the corresponding disk in the VM image: 30 GB. This is not allowed. Please choose equal or greater size or do not specify an explicit size." Target="osDisk.diskSizeGB"
```

As one cannot decreese the disk size this is the minimal value.
This is not the case for EC2 and GCE see 60f1490 merged in #13766

- Verification run: [SLES15-SP2-Azure-x86_64](http://pdostal-server.suse.cz/tests/12908)

### DIFF
--- a/data/publiccloud/terraform/azure.tf
+++ b/data/publiccloud/terraform/azure.tf
@@ -182,9 +182,12 @@ resource "azurerm_linux_virtual_machine" "openqa-vm" {
   }
 
   os_disk {
-    name              = "${var.name}-${element(random_id.service.*.hex, count.index)}-osdisk"
+    name                 = "${var.name}-${element(random_id.service.*.hex, count.index)}-osdisk"
     caching              = "ReadWrite"
     storage_account_type = "Standard_LRS"
+    # SLE images are 30G by default. Uncomment this line in case we need to increase the disk size
+    # note: value can not be decreased because 30 GB is minimum allowed by Azure
+    # disk_size_gb         = 30
   }
 
   source_image_id =  var.image_id != "" ? azurerm_image.image.0.id : null


jlausuch:
I thought Azure already uses 30GB by default, why do we need this?

pdostal:
> I thought Azure already uses 30GB by default, why do we need this?

I was not sure why Azure was failing yesterday afternoon. So we can close this PR.
But we can also specify it here so we're:
1) Sure it does not shrink
2) Aware that it's 30G
3) Easily able to increase the number

No big deal tho.

pdostal:
Take it like a small documentation adjust.

grisu48:
I'm in favour of merging this PR, as it makes the default more visible to us.
It has no functional implications, but removes a source of confusion by explicitly stating the disk size. This removes the assumption that we will remember that in a year from now :slightly_smiling_face:

data/publiccloud/terraform/azure.tf
```diff
@@ -182,9 +182,11 @@ resource "azurerm_linux_virtual_machine" "openqa-vm" {
   }
 
   os_disk {
-    name              = "${var.name}-${element(random_id.service.*.hex, count.index)}-osdisk"
+    name                 = "${var.name}-${element(random_id.service.*.hex, count.index)}-osdisk"
     caching              = "ReadWrite"
     storage_account_type = "Standard_LRS"
+    # SLE images are 30G by default
```
grisu48, line 188: I'm pretty sure we will not understand this comment in a year from now so I'd remove it. Why?

If I would see this comment, then I would ask the question: "if those are 30G by default, why do we explicitly state `disk_size_gb = 30` in the next line?"
We now know why this is there, but without this knowledge this comment might be confusing.

grisu48 (approved):
LGTM :rocket: 

One little nitpicking comment, however feel free to dispute or ignore it.

data/publiccloud/terraform/azure.tf
pdostal: The comment now says:
```
# SLE images are 30G by default so the next line is currently unnecessary
```

jlausuch: I still think it's confusing..  you say that the line is unnecessary but it is still there :D 
I would simple remove it or say `size can be >=30G`

grisu48: How about

    # The default for SLE images is currently 30G. The following line is kept here in case we need to increase the disk size

asmorodskyi: if we speaking about "documentation commit" let's keep it like this . 

```
# SLE images are 30G by default. Uncomment this line in case we need to increase the disk size ( note: value can not  be decreased because 30 GB is minimum allowed by Azure ) 
# disk_size_gb         = 30
```

asmorodskyi (approved):
LGTM

pdostal:
Verification runs have been updated and this is now ready to be merged.

grisu48 (approved):
LGTM :rocket:

grisu48:
Just gonna merge it now because we have two LGTM and we could discuss about the formulation for weeks without any outcome.

PR #13785, opened 2021-11-30
Apply existing test modules using YuiRestClient to openSUSE YaST schedules

OleksandrOrlov:
The PR replaces old modules that used needles with the new ones that use LibyuiClient in opensuse schedules.

There will be more PRs with similar change. In order to make review easier and avoid mistakes, this PR contains only the test suites where `guided_setup` was used.

- Related ticket: https://progress.opensuse.org/issues/102191
- Verification runs: https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=20211129_oorlov&groupid=38

### DIFF
--- a/schedule/yast/opensuse/installer_extended/installer_extended.yaml
+++ b/schedule/yast/opensuse/installer_extended/installer_extended.yaml
@@ -15,12 +15,10 @@ schedule:
   - installation/installation_mode
   - installation/logpackages
   - installation/system_role/select_role_desktop_with_KDE_plasma
-  - installation/partitioning/select_guided_setup
-  - installation/partitioning/guided_setup
   - installation/partitioning/accept_proposed_layout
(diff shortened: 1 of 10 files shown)


schedule/yast/opensuse/installer_extended/installer_extended.yaml
```diff
@@ -15,12 +15,10 @@ schedule:
   - installation/installation_mode
   - installation/logpackages
   - installation/system_role/select_role_desktop_with_KDE_plasma
-  - installation/partitioning/select_guided_setup
-  - installation/partitioning/guided_setup
```
OleksandrOrlov, line 19: I don't know why here the test is using Guided Setup. I checked logs and video and noticed that it does not change anything there, so I guess we can skip entering to Guided Setup, as everything remains by default anyway.

ge0r (approved):
Looks good!

PR #13788, opened 2021-12-01
Add Apparmor profile "dnsmasq" test case into openQA

lilyeyes:
Add Apparmor profile "dnsmasq" test case into openQA.
Factory First will be supported later.
Nested VM related so only implement this case on x86 arch.

- Related ticket: https://progress.opensuse.org/issues/103458
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/7820554 (failure can be tracked by product bug bsc#1159626)

### DIFF
--- a/tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
+++ b/tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
@@ -1,0 +1,97 @@
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Package: apparmor-utils nscd
+# Summary: Test with "usr.sbin.dnsmasq" is in "enforce" mode and AppArmor is
+#          "enabled && active", libvirt manages the NAT network depending on
+#          dnsmasq have no error
+# - Create virtual machines with libvirt in NAT network setting
+# - Switch dnsmasq to enforce mode
+# - Check network status of the guest machine(s)
+# - Check audit.log for errors related to dnsmasq
+# Maintainer: llzhao <llzhao@suse.com>
+# Tags: poo#103458, tc#1767545
+
+use base 'apparmortest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub check_audit_log {
+    my ($self) = shift;
+    my $log_file = $apparmortest::audit_log;
+
+    # Check audit log contains no related error
+    my $script_output = script_output "cat $log_file";
+    if ($script_output =~ m/type=AVC .*apparmor=.*DENIED.* profile=.*dnsmasq.* comm=.*libvirt.*/sx) {
+        record_info('ERROR', "There are errors found in $log_file", result => 'fail');
+        $self->result('fail');
+    }
+    record_info("audit log:", "$script_output");
+}
+
+sub run {
+    my ($self) = shift;
+    my $log_file = $apparmortest::audit_log;
+    my $result_file = '/tmp/result_file';
+    my $image_path = '/var/lib/libvirt/images';
+
+    select_console 'root-console';
+
+    # Install the required packages for libvirt environment setup
+    zypper_call('in qemu dnsmasq libvirt');
+
+    # Check the version for "SLE-20353 QA: dnsmasq update to 2.83"
+    my $current_ver = script_output('zypper se -s -i dnsmasq libnettle');
+    record_info('Version', "$current_ver");
+
+    # Start libvirtd daemon and start the default libvirt network
+    assert_script_run('systemctl start libvirtd');
+    assert_script_run('virsh net-define /etc/libvirt/qemu/networks/default.xml');
+    assert_script_run('virsh net-start default');
+    assert_script_run('systemctl is-active libvirtd');
+    assert_script_run('virsh net-list | grep default | grep active');
+
+    # Download the pre-installed guest images and sample xml files
+    my $vm_name = 'nested-L2-vm';
+    my $vm_L2 = get_required_var('HDD_L2');
+    assert_script_run("wget -c -P $image_path " . autoinst_url("/assets/hdd/$vm_L2"), 900);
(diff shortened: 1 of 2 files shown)


tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
```diff
@@ -0,0 +1,95 @@
...
+use utils;
+
+sub check_audit_log {
+    my ($self) = shift;
+    my $log_file = $apparmortest::audit_log;
+
+    # Check audit log contains no related error
+    my $script_output = script_output "cat $log_file";
+    if ($script_output =~ m/type=AVC .*apparmor=.*DENIED.* profile=.*dnsmasq.* comm=.*libvirt.*/sx) {
+        record_info('ERROR', "There are errors found in $log_file", result => 'fail');
```
Amrysliu, line 28: why not use `die` directly?

lilyeyes: > why not use `die` directly?

I want test case continues executing though failed, such as I want to check audit.log https://openqa.suse.de/tests/7820554#step/usr_sbin_dnsmasq/61

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
```diff
@@ -0,0 +1,95 @@
...
+
+    # Set the AppArmor security profile to enforce mode
+    my $profile_name = 'usr.sbin.dnsmasq';
+    validate_script_output("aa-enforce $profile_name", sub { m/Setting .*$profile_name to enforce mode./ });
+
+    # Define/Start guest vm and restart libvirtd to generate audit records
+    assert_script_run("cd $image_path");
+    assert_script_run('systemctl restart libvirtd');
+    assert_script_run("virsh define $vm_name.xml");
+    assert_script_run("virsh start $vm_name");
```
rfan1, line 74: Lili,
One minor comment here [not a big deal]
If you are able to boot up the vm, You can get a dhcp ip address via dnsmasp record under dir:
"/var/lib/libvirt/dnsmasq/virbr0.status" just like below:
[
  {
    "ip-address": "192.168.122.xx",
    "mac-address": "52:54:xx:xx:xx:xx",
    "hostname": "xxxxxx",
    "client-id": "ff:00:xx:xx:xx:00:01:00:01:28:ba:xx:xx:52:54:00:xx:xx:xx",
    "expiry-time": 16393xxx0
  }
]

lilyeyes: Thanks for the info.

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
```diff
@@ -0,0 +1,95 @@
...
+sub run {
+    my ($self) = shift;
+    my $log_file = $apparmortest::audit_log;
+    my $result_file = '/tmp/result_file';
+    my $image_path = '/var/lib/libvirt/images';
+
+    select_console 'root-console';
+
+    # Check the version for "SLE-20353 QA: dnsmasq update to 2.83"
+    my $current_ver = script_output('zypper se -s dnsmasq libnettle');
```
jouyingbin, line 43: @lilyeyes 
I have 2 questions here
- You dump the dnsmasq version from the scc/proxyscc rather than the version which is installed into the system, right?
- Do you need to install the dnsmasq into the system? looks I did not find it from the code or if I missed something?

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
```diff
@@ -0,0 +1,95 @@
...
+    record_info('Version', "$current_ver");
+
+    # Install the required packages for libvirt environment setup,
```
jouyingbin, line 46: Will you remove the " , "  in the end of the line? or if there is something missing after that.

jouyingbin:
another small comment:

Could you use the fixed word usage for **AppArmor** in the subject or in git commit description?

Perhaps they are apparmor or AppArmor?  Looks like no Apparmor.  Right?

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
```diff
@@ -0,0 +1,95 @@
...
+    zypper_call('in qemu libvirt virt-install virt-manager');
```
rfan1, line 47: Lili, 

If you use 'virsh' only, may don't need to install virt-install and virt-manager here

rfan1: # rpm -qf /usr/bin/virsh
libvirt-client-7.1.0-4.1.x86_64

lilyeyes: You are right.

lilyeyes: You are right.

> another small comment:
> 
> Could you use the fixed word usage for **AppArmor** in the subject and in git commit description?
> 
> Perhaps they are apparmor or AppArmor? Looks like no Apparmor. Right?

Good pointing. I will revise.

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
lilyeyes: Deleted.

tests/security/apparmor_profile/usr_sbin_dnsmasq.pm
lilyeyes: Good question:
1. yes, but I only aim to show the available versions and then do manually check.
2. good pointing, when installing "libvirt" the "dnsmasq" will be installed automatically. I will revise code to avoid the confusion.

rfan1 (approved):
LGTM

jouyingbin (approved):
I don't have further comments now, LGTM, thanks.

Amrysliu (approved):

PR #13790, opened 2021-12-01
Set NFS grace time correctly before test

acerv:
Added NFS_GRACE_TIME which controls both grace and lease time.
Also NFS grace/lease time is not correctly set inside tests. Right way of doing it is:
- to set NFSV4LEASETIME inside /etc/sysconfig/nfs
- to set grace-time inside /etc/nfs.conf.local

VRs:
- [SLE](http://10.100.12.105/tests/1462)
- [TW](http://10.100.12.105/tests/1461)


pevik:
@czerw @SeroSun FYI

pevik (approved):
LGTM, it leads to great speedup. 

nit: I'd put PR description also into commit message (helps when reading commits offline).

czerw (approved):

PR #13791, opened 2021-12-01
Remove python test files after running the tests

Avinesh:
Signed-off-by: Avinesh Kumar <akumar@suse.de>

Removing the test files after test execution is finished.

- Related ticket: https://progress.opensuse.org/issues/103350
- Verification run: 
  [SLE15-SP4](https://openqa.suse.de/tests/7767276#)
  [SLE15-SP3](http://d462.qam.suse.de/tests/59)

### DIFF
--- a/tests/console/python_scientific.pm
+++ b/tests/console/python_scientific.pm
@@ -48,12 +48,18 @@ sub run {
 
 sub post_fail_hook {
     my $self = shift;
+    $self->cleanup();
     $self->SUPER::post_fail_hook;
 }
 
 sub post_run_hook {
     my $self = shift;
+    $self->cleanup();
     $self->SUPER::post_run_hook;
+}
+
+sub cleanup {
+    script_run('rm -f python3-numpy-test.py python3-scipy-test.py');
 }
 
 1;


tests/console/python_scientific.pm
```diff
@@ -33,6 +33,7 @@ sub run_python_script {
             record_soft_failure("$failmsg");
         }
     }
+    script_run("rm '$script'");
```
grisu48, line 36: I believe this should go into the `post_run_hooks` and the `post_fail_hooks`, otherwise the file remains present, also if the test run fails.

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/sshd.pm#L221 and https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/sshd.pm#L227 as an example.

Avinesh: @grisu48 , Thanks for reviewing and pointing to the correct way of implement this.
Please check the updated commit.

asdil12 (approved):

PR #13794, opened 2021-12-01
BCI: run each test environment as a different test module

jlausuch:
This refactor allows loading in runtime each test into a new
test module to improve visibility and code structure.

- Related ticket: https://progress.opensuse.org/issues/103326
- VRs: [link](http://fromm.arch.suse.de/tests/overview?distri=sle&version=15-SP3&build=17.8.38&groupid=41)


jlausuch:
How to remove Xunit results after all modules executed:
image
as they are already shown in the modules themselves.

tests/containers/bci_test.pm
```diff
@@ -10,92 +10,56 @@
...
 #
 #   This module is used to test BCI repository and BCI container images.
-#   It makes the call to tox to run the different test environments and
-#   parses the resulting JUnit logs at the end.
+#   It makes the call to tox to run a specific test environment.
 # Maintainer: qa-c team <qa-c@suse.de>
 
 use Mojo::Base qw(consoletest);
-use XML::LibXML;
+use Time::HiRes qw(clock_gettime CLOCK_MONOTONIC);
```
cfconrad, line 17: Not sure why we use `Time::HiRes`, we only display seconds unit in the end.
a hint:
`use Time::HiRes qw(time)` overwrite your current `time()` function which return then a float like `<sec>.<ms>`

pdostal:
image

This looks great, Jose! I have just one question:

What does `all` mean and where do I see on which image the test runs?

jlausuch:
There are 2 major problems with this PR:
- when a module fails, it breaks the console and the following modules crash:
http://fromm.arch.suse.de/tests/5281#

- also, when it works, it prints the junit results twice: http://fromm.arch.suse.de/tests/5280
and doesn't print the normal console output frames any more...

I already spent too much time in trying to fix this with no result, so I'm closing it as this PR is just pure aesthetics.

jlausuch:
> image
> 
> This looks great, Jose! I have just one question:
> 
> What does `all` mean and where do I see on which image the test runs?

https://dancermak.io.suse.de/bci-tests/testdoc.html#module-tests.test_all

PR #13796, opened 2021-12-02
Implement pam faillock test into openQA

rfan1:
- Related ticket: https://progress.opensuse.org/issues/102990
- Needles: n/a
- Verification run: 
https://openqa.suse.de/tests/7772350 -Full pam test run on SLE x86-64
- Sub-test VRs:
 https://openqa.suse.de/tests/7766999           - SLE s390x
 https://openqa.suse.de/tests/7767003           - SLE aarch64
 https://openqa.opensuse.org/tests/2065727 - TW x86_64

I didn't run VR on ppc64le, but I think it can pass the tests on it

### DIFF
--- a/tests/security/pam/pam_faillock.pm
+++ b/tests/security/pam/pam_faillock.pm
@@ -1,0 +1,97 @@
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary: PAM tests for faillock, the uesr login can be locked
+#          if reaches the limit number of authentication failures;
+#          we can unlock it as well with root user.
+# Maintainer: rfan1 <richard.fan@suse.com>
+# Tags: poo#102990 tc#1769824
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use base 'consoletest';
+use utils 'zypper_call';
+use version_utils 'is_sle';
+
+sub run {
+    select_console('root-console');
+    zypper_call('in expect');
+
+    my $user_name = 'pamtest';
+    my $user_pw = $testapi::password;
+    my $bad_pw = 'badpassword';
+    assert_script_run("useradd -m $user_name");
+    assert_script_run("echo $user_name:$user_pw | chpasswd");
+
+    # Package change log check on SLE: jsc#sle-20638
+    validate_script_output('rpm -q pam --changelog', sub { m/jsc#sle-20638/ }) if (is_sle);
+
+    # Basic function test, lock&unlock
+    # Modify the pam configuration files
+    my $pam_config = <<EOF;
+auth       required        pam_faillock.so preauth deny=3 unlock_time=600
+auth       required        pam_faillock.so authfail deny=3 unlock_time=600
+account    required        pam_faillock.so
+EOF
+    assert_script_run('cp /etc/pam.d/common-auth /etc/pam.d/common-auth.back');
+    assert_script_run('cp /etc/pam.d/common-password /etc/pam.d/common-password.back');
+    assert_script_run("echo '$pam_config' >> /etc/pam.d/common-auth");
+    assert_script_run("echo '$pam_config' >> /etc/pam.d/common-password");
+
+    # After 3 failed login, the user will be locked
+    assert_script_run(
+        "expect -c '
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$bad_pw\\r\" }
+    }
+    expect {
+        {denied} { exp_continue }
+        {assword} { send \"$bad_pw\\r\" }
+    }
+    expect {
+        {denied} { exp_continue }
+        {assword} { send \"$bad_pw\\r\" }
+    }
+    expect {
+        {denied} { exp_continue }
+        {assword} { send \"$bad_pw\\r\" }
+    }
+    expect -nocase {denied} { close; wait }'"
+    );
+
+    assert_script_run(
+        "expect -c '
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect -nocase {The account is locked due to 3 failed logins} { close; wait }'"
+    );
+
+    # Check the syslog can record the lock operation
+    assert_script_run("journalctl -a | grep 'pam_faillock.*$user_name account temporarily locked'");
+
+    # Unlock the user, then we can login to the host
+    assert_script_run("faillock --user $user_name --reset");
+    assert_script_run(
+        "expect -c '
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect {$user_name\@\$HOSAME} { send \"exit\\r\" }'"
+    );
+
+    # Clean up
+    assert_script_run('mv /etc/pam.d/common-auth.back /etc/pam.d/common-auth');
+    assert_script_run('mv /etc/pam.d/common-password.back /etc/pam.d/common-password');
+    assert_script_run("userdel -r $user_name");
+}
+
+1;
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -2449,6 +2449,7 @@ sub load_security_tests_pam {
     loadtest "security/pam/pam_su";
     loadtest "security/pam/pam_config";
     loadtest "security/pam/pam_mount";
+    loadtest "security/pam/pam_faillock";
 }
 
 sub load_security_tests_create_swtpm_hdd {


tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
+# Copyright 2020 SUSE LLC
```
jouyingbin, line 1: typo, s/2020/2021

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    # Modify the pam configuration files
+    my $pam_config = <<EOF;
+auth       required        pam_faillock.so preauth deny=3 unlock_time=600
+auth       required        pam_faillock.so authfail deny=3 unlock_time=600
+account    required        pam_faillock.so
+EOF
+    assert_script_run("cp /etc/pam.d/common-auth /etc/pam.d/common-auth.back");
+    assert_script_run("cp /etc/pam.d/common-password /etc/pam.d/common-password.back");
+    assert_script_run("echo '$pam_config' >> /etc/pam.d/common-auth");
+    assert_script_run("echo '$pam_config' >> /etc/pam.d/common-password");
```
lilyeyes, line 41: I recommend you use variables to define files, such as  /etc/pam.d/common-auth, /etc/pam.d/common-password

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect -nocase {The account is locked due to 3 failed logins} { close; wait }'"
+    );
+
+    # Check the syslog can recorded the lock operation
+    assert_script_run("journalctl -a | grep 'pam_faillock.*$user_name account temporarily locked'");
+
+    # Unlock the user, then we can login to the host
+    assert_script_run(" faillock --user $user_name --reset");
```
lilyeyes, line 80: Delete the "blank space" before "faillock" please.

rfan1: Will do

tests/security/pam/pam_faillock.pm
rfan1: I have the opposite option here, we don't need to set a variable for system configuration file, we used more lines and get less benefit, and make code not friendly to be read. :)

If it is not a must one, I will keep current logic.

tests/security/pam/pam_faillock.pm
rfan1: Got it, revised

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+        "expect -c '
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect -nocase {The account is locked due to 3 failed logins} { close; wait }'"
+    );
+
+    # Check the syslog can recorded the lock operation
```
Amrysliu, line 76: remove 'can' or change to 'can record'

tests/security/pam/pam_faillock.pm
lilyeyes: Sure, I understood what you mean, but defining a variable is really a good way, such as if file name changed in the future you only need to revise the definition line to update. Besides if the variable's name is well named it is also readable too :).
Anyway, it is optional you can “keep current logic". ：）

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect {$user_name\@\$HOSAME} { send \"exit\\r\" }'"
+    );
+
+    # Clean up
+    assert_script_run("mv /etc/pam.d/common-auth.back /etc/pam.d/common-auth");
+    assert_script_run("mv /etc/pam.d/common-password.back /etc/pam.d/common-password");
```
lilyeyes, line 93: Please check the usage of " and " for coding sytle, e.g., s/"/''/ in lines 92+93 and others

tests/security/pam/pam_faillock.pm
rfan1: done

Amrysliu (approved):
lgtm

tests/security/pam/pam_faillock.pm
rfan1: Good suggestion, done

lilyeyes (approved):
lgtm.

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+
+sub run {
+    select_console('root-console');
+    zypper_call('in expect');
+
+    my $user_name = 'pamtest';
+    my $user_pw = $testapi::password;
+    my $bad_pw = 'badpassword';
+    assert_script_run("useradd -m $user_name");
+    assert_script_run("echo $user_name:$user_pw | chpasswd");
```
jouyingbin, line 26: Just check the VR, 
https://openqa.suse.de/tests/7772350#step/pam_faillock/6
This is the expected result, right ?

rfan1: Cool finding, we should get this warning message due to current pam/password policy, however, the passwd can still be set there.
You are correct, it is by design

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    }
+    expect -nocase {The account is locked due to 3 failed logins} { close; wait }'"
+    );
+
+    # Check the syslog can record the lock operation
+    assert_script_run("journalctl -a | grep 'pam_faillock.*$user_name account temporarily locked'");
+
+    # Unlock the user, then we can login to the host
+    assert_script_run("faillock --user $user_name --reset");
+    assert_script_run(
```
jouyingbin, line 81: Looks like the check login method is used repeatedly, will you like to like a small sub func()  to recall it?

tests/security/pam/pam_faillock.pm
jouyingbin: got it :)

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    # After 3 failed login, the user will be locked
+    assert_script_run(
+        "expect -c '
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$bad_pw\\r\" }
+    }
+    expect {
+        {denied} { exp_continue }
```
jouyingbin, line 52: I am not sure if it is suitable to also write a small sub func() to recall the **$bad_pw** usage. But it is optional.

tests/security/pam/pam_faillock.pm
```diff
@@ -0,0 +1,97 @@
...
+    spawn ssh $user_name\@localhost
+    expect {
+        {continue} { send \"yes\\r\"; exp_continue }
+    {assword} { send \"$user_pw\\r\" }
+    }
+    expect {$user_name\@\$HOSAME} { send \"exit\\r\" }'"
+    );
+
+    # Clean up
+    assert_script_run('mv /etc/pam.d/common-auth.back /etc/pam.d/common-auth');
```
jouyingbin, line 92: Similar comment with Lily, if it's defined as a variable, you can call the variable here, and if the path is changed in the future, you just need to modify one time from the original definition. But it's optional.

tests/security/pam/pam_faillock.pm
rfan1: Actually, I wanted to do that, but it take more effort to run it out with expect script. so keep it in current style.

BTW, in cc audit test code, it uses similar style as well.

tests/security/pam/pam_faillock.pm
rfan1: The system configuration files will seldom change, IMO, it is not suitable for define a new variable for a "constant"

tests/security/pam/pam_faillock.pm
rfan1: Actually, I wanted to do that, but it take more effort to run it out with expect script. so keep it in current style.

BTW, in cc audit test code, it uses similar style as well.

tests/security/pam/pam_faillock.pm
rfan1: Actually, I wanted to do that, but it is not an easy way to implement within expect script.

tests/security/pam/pam_faillock.pm
jouyingbin: okay. no problem, thanks.

tests/security/pam/pam_faillock.pm
jouyingbin: I think so, and let's keep it. :)

jouyingbin (approved):
LGTM, thanks for the explanation.

PR #13798, opened 2021-12-02
 Fix GCE removing key_id/key_secret is unnecessary

ilausuch:
These variables are used by EC2 but not GCE.

- Related ticket:https://progress.opensuse.org/issues/103410
- Verification run: https://openqa.suse.de/tests/7772553


jlausuch:
3 more VRs:
https://openqa.suse.de/tests/7772674
https://openqa.suse.de/tests/7772675
https://openqa.suse.de/tests/7772676

Will merge when VRs are done and green.

PR #13800, opened 2021-12-02
lib/lynis/lynistest: Don't force a return "ok" if a warning was found

Vogtinator:
If the lynis result is different, but the current result contains a warning
or a particular DBus check, it forcibly returned "ok" and skipped actually
looking for soft and hard fails. This lead to random false positives.

Example: https://openqa.opensuse.org/tests/2065411#step/4_%5B+%5D_Boot_and_services/5

Just remove the code, so that failures can actually be seen.

CC @lilyeyes 

- Verification runs:
True negative (using old baseline): http://10.160.67.86/tests/1156
True positive: https://openqa.opensuse.org/tests/2067019

### DIFF
--- a/lib/lynis/lynistest.pm
+++ b/lib/lynis/lynistest.pm
@@ -259,14 +259,6 @@ sub compare_lynis_section_content {
         }
     }
 
-    # Filter out "System_Tools": "[2C- Starting dbus policy check...[28C"
-    # As this will likely change rather frequently and we have good mechanisms
-    # on our side to tackle this
-    if (grep(/Starting dbus policy check/, @section_current)) {
-        $result = "ok";
-        return $result;
-    }
-
     # If a new section then only check the current
     record_info("CHECK", "Section contents NOT the same then check \"Current\" only");
     for my $s_lynis (@lynis_ok) {
@@ -283,29 +275,6 @@ sub compare_lynis_section_content {
         $s_new = "\\[.*$s_lynis.*\\]";
         $ret = grep(/$s_new/, @section_current);
         if ($ret) {
-            # Filter out some exceptions allowed:
-            # "Boot_and_services": "[4C- Checking for password protection[23C [ WARNING ]"
-            # "Name services": "[4C- Checking /etc/hosts (hostname)[25C [ SUGGESTION ]"
-            # "Kernel: "[4CCPU support: No PAE or NoeXecute supported[15C [ NONE ]"
-            # "Initializing_program": "[2C- Program update status... [32C [ WARNING ]"
-            # "Networking": "[[4C- Minimal of 2 responsive nameservers^[[20C [ WARNING ]"
-            # "Ports and packages": "Using Zypper to find vulnerable packages[17C [ NONE ]"
-            my @exceptions = (
-                "Checking for password protection.*WARNING.*",
-                "Checking /etc/hosts .*hostname.*SUGGESTION.*",
-                "CPU support: No PAE or NoeXecute supported.*NONE.*",
-                "Program update status.*WARNING.*",
-                "Minimal of 2 responsive nameservers.*WARNING.*",
-                "Using Zypper to find vulnerable packages.*NONE.*",
-                "Test .* had a long execution: .* seconds"
-            );
-            for my $exception (@exceptions) {
-                if (grep(/$exception/, @section_current)) {
-                    $result = "ok";
-                    return $result;
-                }
-            }
-
             $result = "softfail";
             record_soft_failure("poo#91383, found $ret [ $s_lynis ] in current output");
         }


lilyeyes:
For these minor issues I aim to mark as "softfail" after developer approved the baseline is PASS.
(The approved baseline contains some "suggestions/warnings/exposed", so there are exception is code)

lilyeyes:
My suggestion is if TW wants these "softfails" changed to "fails" I don't have objection but please add product version check control, such as if is_tumble() code will not failed and not return "OK".

lilyeyes:
My suggestion is if TW wants these "softfails" changed to "fails" I don't have objection but please add product version check control, such as if is_tumble() code will not failed and not return "OK".

Vogtinator:
> My suggestion is if TW wants these "softfails" changed to "fails" I don't have objection but please add product version check control, such as if is_tumble() code will not failed and not return "OK".

The issue is that before this PR, even if the test has an actual fail, it would show up as passing. I don't think that is desirable for any product.

lilyeyes:
> > My suggestion is if TW wants these "softfails" changed to "fails" I don't have objection but please add product version check control, such as if is_tumble() code will not failed and not return "OK".
> 
> The issue is that before this PR, even if the test has an actual fail, it would show up as passing. I don't think that is desirable for any product.

You are right. so we have to check every softfail for sure when doing openQA review. 
If you want to keep your changes (without product version control) could you please offer the verification runs for sles? (I am asking because the baselines for sles and TW are different the changes may introduce issues)
FYI: search "lynis" for all the test cases: http://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=70.1&groupid=268

Vogtinator:
> If you want to keep your changes (without product version control) could you please offer the verification runs for sles? (I am asking because the baselines for sles and TW are different the changes may introduce issues)
FYI: search "lynis" for all the test cases: http://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=70.1&groupid=268

I don't have a setup for running SLE tests. Can you start some VRs?

lilyeyes:
> > If you want to keep your changes (without product version control) could you please offer the verification runs for sles? (I am asking because the baselines for sles and TW are different the changes may introduce issues)
> > FYI: search "lynis" for all the test cases: http://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=70.1&groupid=268
> 
> I don't have a setup for running SLE tests. Can you start some VRs?

Sure, please wait for hours.

lilyeyes:
I have triggered jobs on x86_86 and s390x for "gnome" mode and "textmode":
x86_64 VRs:
  lynis_gnome: https://openqa.suse.de/tests/7810981
  lynis_textmode: https://openqa.suse.de/tests/7810979

s390x VRs:
  lynis_gnome: https://openqa.suse.de/tests/7810983
  lynis_textmode: https://openqa.suse.de/tests/7810984

You can try to trigger on other arches (ppc64 hmc backend and aarch64) as following: 

```
# /usr/share/openqa/script/client isos post --host http://openqa.nue.suse.com _GROUP_ID=268 TEST=lynis_textmode DISTRI=sle VERSION=15-SP4  ARCH=x86_64  BUILD=70.1 FLAVOR=Online _SKIP_CHAINED_DEPS=1 --apikey $your-key --apisecret $your-pw CASEDIR=https://github.com/Vogtinator/os-autoinst-distri-opensuse.git#truenegative

# /usr/share/openqa/script/client isos post --host http://openqa.nue.suse.com _GROUP_ID=268 TEST=lynis_gnome DISTRI=sle VERSION=15-SP4  ARCH=x86_64  BUILD=70.1 FLAVOR=Online _SKIP_CHAINED_DEPS=1 --apikey $your-key --apisecret $your-pw CASEDIR=https://github.com/Vogtinator/os-autoinst-distri-opensuse.git#truenegative
```

lilyeyes:
You can see in the VRs (e.g., https://openqa.suse.de/tests/7810981#step/1_[+]_Initializing_program/10) the changes introduce 
 unnecessary soft failures but they are no harm and permitted via the code of `"Filter out some exceptions allowed"`.

The parsing of the outputs is really had so I chose this way, you may figure out a better way to parse the outputs but the big efforts are big IMO.

Vogtinator:
> You can see in the VRs (e.g., https://openqa.suse.de/tests/7810981#step/1_[+]_Initializing_program/10) the changes introduce
unnecessary soft failures but they are no harm and permitted via the code of "Filter out some exceptions allowed".

The issue is that they DO harm because they will also softfail if there's a critical issue found by lynis. That is IMO not acceptable.

lilyeyes:
> 
Yes, if an unknow or new softfail found it might be a critical issue, so we have to check the softfail for sure.
I remember test module is marked as "softfail" only when test found non-critical results in the lynis-outputs such as "[ WARNING,EXPOSED,NONE,SUGGESTION ]", see bellow code fyi.

Btw, the exceptions in code are known and acceptable issue.

```
# .../lib> vi lynis/lynistest.pm
# Lynix test status mappping with openQA
our @lynis_ok = split(/,/, get_var("LYNIS_OK", "OK,DONE,YES"));
our @lynis_fail = split(/,/, get_var("LYNIS_ERROR", "ERROR,WEAK,UNSAFE"));
our @lynis_softfail = split(/,/, get_var("LYNIS_WARNING", "WARNING,EXPOSED,NONE,SUGGESTION"));
```

Vogtinator:
> Yes, if an unknow or new softfail found it might be a critical issue, so we have to check the softfail for sure.

I think there's a misunderstanding.

If the lynis check sees some unexpected `ERROR` or `UNSAFE` in the output which weren't there before, it sets the test result to "fail".

However, later on if it also sees "Test had a long execution time" in the log, it sets the status to "softfail" despite there being a critical error already.

lilyeyes:
> > Yes, if an unknow or new softfail found it might be a critical issue, so we have to check the softfail for sure.
> 
> I think there's a misunderstanding.
> 
> If the lynis check sees some unexpected `ERROR` or `UNSAFE` in the output which weren't there before, it sets the test result to "fail".
> 
Yes. Agree with you. after compared with baseline if new "ERROR,WEAK,UNSAFE" founded test case should be set "fail".
If not test case should be enhanced.

> However, if it also sees "Test had a long execution time" in the log, it sets the status to "ok" despite there being a critical error already.
The current parsing of the outputs of lynis might be not very accurate we can file poo to fix/enhance if the critical "error" is new and code is not able to find out.
To be honest, the output parsing code is really hard to so accurate so need manual check also currently (I do this for sles openQA review).
But delete all the exceptions is not a good idea if you add a product version checking for TW it could be great.

lib/lynis/lynistest.pm
```diff
@@ -259,14 +259,6 @@ sub compare_lynis_section_content {
...
     }
 
-    # Filter out "System_Tools": "[2C- Starting dbus policy check...[28C"
-    # As this will likely change rather frequently and we have good mechanisms
-    # on our side to tackle this
-    if (grep(/Starting dbus policy check/, @section_current)) {
-        $result = "ok";
-        return $result;
-    }
-
```
lilyeyes, line 269: This dbus exception is approved by developer otherwise test case will fail randomly we have to pay efforts to review when fails.

lib/lynis/lynistest.pm
```diff
@@ -283,29 +275,6 @@ sub compare_lynis_section_content {
...
-                "Using Zypper to find vulnerable packages.*NONE.*",
-                "Test .* had a long execution: .* seconds"
-            );
-            for my $exception (@exceptions) {
-                if (grep(/$exception/, @section_current)) {
-                    $result = "ok";
-                    return $result;
-                }
-            }
-
```
lilyeyes, line 308: These warnings are OK too.

DimStar77:
> However, if it also sees "Test had a long execution time" in the log, it sets the status to "ok" despite there being a critical error already.

This seems to be the most critical part causing confusion.

As soon as 'Test had a long execution time", all the other "fail" being reported by the test are ignored and the test always marled "ok"

PR #13802, opened 2021-12-02
Implement perform installation with YuiRestClient

jknphy:
- Related ticket: [poo#98958](https://progress.opensuse.org/issues/98958)
- Verification run: [SLE](https://openqa.suse.de/tests/overview?build=jknphy%2Fos-autoinst-distri-opensuse%23wait_install_libyui&version=15-SP4&distri=sle) | [openSUSE](https://openqa.opensuse.org/tests/overview?distri=opensuse&build=jknphy%2Fos-autoinst-distri-opensuse%23wait_install_libyui)

### DIFF
--- a/lib/YuiRestClient/Widget/ProgressBar.pm
+++ b/lib/YuiRestClient/Widget/ProgressBar.pm
@@ -1,0 +1,65 @@
+# SUSE's openQA tests
+
+package YuiRestClient::Widget::ProgressBar;
+
+use strict;
+use warnings;
+
+use parent 'YuiRestClient::Widget::Base';
+
+sub value {
+    my ($self) = @_;
+    return $self->property('value');
+}
+
+sub max_value {
+    my ($self) = @_;
+    return $self->property('max_value');
+}
+
+1;
+
+__END__
+
+=encoding utf8
+
+=head1 NAME
+
+YuiRestClient::Widget::ProgressBar - handle progress bars
+
+=head1 COPYRIGHT
+
+Copyright © 2021 SUSE LLC
+
+SPDX-License-Identifier: FSFAP
+
+=head1 AUTHORS
+
+QE YaST <qa-sle-yast@suse.de>
+
+=head1 SYNOPSIS
+
+$self->{pba_total_packages}->value();
+$self->{pba_total_packages}->max_value();
+
+=head1 DESCRIPTION
+
+=head2 Overview
+
+Class to represent a progress bar. It can be YProgressBar. 
+
+    {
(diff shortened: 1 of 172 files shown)


lib/YuiRestClient/Widget/ProgressBar.pm
```diff
@@ -0,0 +1,65 @@
...
+
+=encoding utf8
+
+=head1 NAME
+
+YuiRestClient::Widget::ProgressBar - handle progress bars
+
+=head1 COPYRIGHT
+
+Copyright © 2020 SUSE LLC
```
ge0r, line 32: I m just nitpicking here: 2021

tests/installation/performing_installation/perform_installation.pm
```diff
@@ -0,0 +1,21 @@
...
+# for slower architectures.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use testapi 'get_var';
+
+sub run {
+    my $timeout = 2750 * get_var('TIMEOUT_SCALE', 1);
```
ge0r, line 17: I ve also seen elsewere this usage of integer numbers directly multiplying the TIMEOUT_SCALE but maybe it could be more explanatory if `2750` would be declared beforehand as a variable or constant, naming it something like `default_install_duration`?
If not, at least a comment along the lines of 2750 seconds is the typical installation duration in our setup.
It would be good to not have this constant as an obscure (for a non initiated reader) magic number.

ge0r (commented):
Good job introducing the ProgressBar widget and replacing await_install successfully on all those jobs! Just a couple of small notes from my side.

tests/installation/performing_installation/perform_installation.pm
jknphy: It is a timeout, it is not the expected time of duration, that would require further analysis (you know, like some expected time average plus some margin or something like that, with some record message when it reached the expected value for further investigation...) and as we discussed offline we are not so concern about performance to apply this changes, we should report it to the proper team if needed. So it is a magic number indeed, like any other timeout long enough to be safe, it is so obscure that in fact it should be 5500 seconds. I'm adding the comment (I had it there for a time and I deleted) but only for the reason that I wasn't sure if we would fix our client on time to represent the right value before this is merged. Anyway good point, we might not set there 5500 for a while and only we know why is that without a comment.

jknphy: https://progress.opensuse.org/issues/103548

PR #13804, opened 2021-12-02
publiccloud: Add analysis of FIO test results against previous runs

asmorodskyi:


jlausuch:
This is gold. You are making this test much more meaningful, so good work! 
I just have some questions/thoughts.

- The condition to trigger a failure is `diff(this_week, last_week) \ last_week > 10`. Why have you chose 10% threshold?

- What happens when we don't have builds in previous week? -> skip analysis?

- What if we have a lot of builds last week and just 1 or 2 this week? 

- Would it be more realistic comparing previous X builds' mean? For instance, take the mean from previous 5 or 10 builds and compare it against the current build.

asmorodskyi:
>     * The condition to trigger a failure is `diff(this_week, last_week) \ last_week > 10`. Why have you chose 10% threshold?

no reason . I am open to any other number . I just needed to choose something so I choose 10 :) 

> 
>     * What happens when we don't have builds in previous week? -> skip analysis?

[yes](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13804/files#diff-71d5940ceca3c88d8d1ab4cfea14234cadabbd47649208edcaa73c6fe67a6e1cR172)  . if there is nothing from one week before OR there is nothing from last week ( which actually not possible because we have current run :) ) than we just display record_info that there will be no analysis because there is not enough data .
> 
>     * What if we have a lot of builds last week and just 1 or 2 this week?

it will still work but we  just have higher chances to fail due to some glitch in single run . 

> 
>     * Would it be more realistic comparing previous X builds' mean? For instance, take the mean from previous 5 or 10 builds and compare it against the current build.

maybe you right , I will change queries accordingly . thanks for idea :+1:

asmorodskyi:
> @asmorodskyi can you show how does this look?

not now , maybe after I will switch logic from "analyze based on results from last 2 weeks " to "analyze based on last x runs " as @jlausuch suggested this would be easier . Tricky part that FIO tests are running only on "Latest ..." job groups and SLE 15 SP3 was released quite long time ago and SLE 15 SP4 does not yet have [testable images](https://progress.opensuse.org/issues/102966) so I simply don't have data on which I can prove that it is working . But certainly I will figure out something before asking to merge it ...

pdostal:
I'm sorry to open it now but why we do this in openQA? Will it at least be a separate test run or will it happen at the end of the `fio` test scenario? I'd definitely keep this analysis in separate test run. Why isn't Grafana enough?

asmorodskyi:
> I'm sorry to open it now but why we do this in openQA?

just to keep things clear - under "this" you mean analysis which I basically adding with this PR , right ?

> Will it at least be a separate test run or will it happen at the end of the fio test scenario?

if it will be separate test than current test will be ALWAYS green ( current state ) and what the point of the test which does not check anything ?

> Why isn't Grafana enough? 

yes you can do same in Grafana . but : 
1. this will mean that every day reviewers will need to check two places for potential failures - openQA AND Grafana . You can setup email or something else notification system but still it will be not the same as having everything in openQA 
2. Even if we somehow accept looking into Grafana additionally my question about validity of the test which never fail remain 
3. Implementing **same** logic in Grafana might be tricky . Not saying not possible but tricky : 
-  We need to trigger such check based on **event** of entering new data not sure if it is possible and never done something like this 
- We need to properly handle concurrency - e.g. when several jobs pushing data into DB and triggered check identify anomaly we need properly identify what dataset caused this . From isolated Grafana POV it is not so straightforward AFAIK 
-  ... and I am sure there will be more challenges   if you start this road :)

asmorodskyi:
> > I'm sorry to open it now but why we do this in openQA?
> 
> just to keep things clear - under "this" you mean analysis which I basically adding with this PR , right ?
> 
> > Will it at least be a separate test run or will it happen at the end of the fio test scenario?
> 
> if it will be separate test than current test will be ALWAYS green ( current state ) and what the point of the test which does not check anything ?
> 
> > Why isn't Grafana enough?
> 
> yes you can do same in Grafana . but :
> 
>     1. this will mean that every day reviewers will need to check two places for potential failures - openQA AND Grafana . You can setup email or something else notification system but still it will be not the same as having everything in openQA
> 
>     2. Even if we somehow accept looking into Grafana additionally my question about validity of the test which never fail remain
> 
>     3. Implementing **same** logic in Grafana might be tricky . Not saying not possible but tricky :
> 
> 
>     * We need to trigger such check based on **event** of entering new data not sure if it is possible and never done something like this
> 
>     * We need to properly handle concurrency - e.g. when several jobs pushing data into DB and triggered check identify anomaly we need properly identify what dataset caused this . From isolated Grafana POV it is not so straightforward AFAIK
> 
>     * ... and I am sure there will be more challenges   if you start this road :)

and please don't get this as "defense mode" :) I am truly open to that idea and will have no regrets on wipe out this PR :) but I truly have big concerns about it . **IF** you manage to resolve them I am on your side :)

pdostal:
Thank you for good answers Anton.

My concern is, that a test usually TESTS something. But this is just fetching the database and failing if the database returns ugly results.

When this analysis fails does it mean that the current build is broken?

asmorodskyi:
- fix some stupid error which I did before VRs
- add more record_info messages to improve visibility 
here is first passed VR http://autobot.qa.suse.de/tests/2583
TODO : 
- add URL generation to "die in case of anomaly detected " so reviewer can easily dive into problem investigation in Grafana ( as agreed with Pavel ) 
- do run where real analysis will happen to see that approach actually works ( might require some dummy data generation or we would need to wait :) ) 
- switch from "2 weeks" approach to "X builds" approach ( as agreed with Jose )

grisu48 (commented):
LGTM, but one comment that should be discussed

tests/publiccloud/storage_perf.pm
```diff
@@ -120,6 +154,33 @@ sub run {
...
+            );
+            if (db_has_data(%influx_read_args, timeframe => $this_week_timeframe) && db_has_data(%influx_read_args, timeframe => $previous_week_timeframe)) {
+                foreach my $load_type (keys %$values) {
+                    $influx_read_args{load_type} = $load_type;
+                    my $current_load_type = sprintf(" load_type=%s, scenario=%s, Flavor=%s, Version=%s", $influx_read_args{load_type}, $influx_read_args{scenario}, $influx_read_args{os_flavor}, $influx_read_args{os_version});
+                    my $this_week = get_median_from_db(%influx_read_args, timeframe => $this_week_timeframe);
+                    my $previous_week = get_median_from_db(%influx_read_args, timeframe => $previous_week_timeframe);
+                    if (defined($previous_week) && defined($this_week) && $previous_week != 0 && $this_week != 0) {
+                        record_info('ANALYZE', "Checking " . $current_load_type);
+                        if ((abs($previous_week - $this_week) / $previous_week) * 100 > 10) {
```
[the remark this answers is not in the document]
cfconrad: Isn't it that it could be also 10% less then before ?

cfconrad (approved):
LGTM,
maybe some more info in the git commit message!

tests/publiccloud/storage_perf.pm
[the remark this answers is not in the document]
grisu48:
```suggestion
                        if (abs(1.0 - $this_week/$previous_week) > 0.1) {    # Deviations of more than 10% are considered an anomaly
```

But the original suggestion is also fine. @asmorodskyi pick the variant you like more, both are correct for the cases we are looking at.

tests/publiccloud/storage_perf.pm
```diff
@@ -17,6 +17,38 @@ use Mojo::JSON;
...
+
+    my $json_res = influxdb_read_data($args{url}, $args{db}, $query);
+    unless (defined($json_res->{results}->[0]->{series})) {
+        record_info('Info', sprintf("No results for load_type=%s, scenario=%s, Flavor=%s, Version=%s for timeframe=[%s]\n", $args{load_type}, $args{scenario}, $args{os_flavor}, $args{os_version}, $args{timeframe}));
+        return undef;
+    }
+
+    my $series = $json_res->{results}->[0]->{series};
+    my @values = @{$series}[0]->{values};
+    return $values[0][0][1];
```
pdostal, line 39: Can you a bit comment this? Maybe with example json snippet? So one understands what `$values[0][0][1]` mean?

tests/publiccloud/storage_perf.pm
```diff
@@ -120,6 +154,33 @@ sub run {
                 values => $values
             };
             $data = influxdb_push_data($url, 'publiccloud', $data);
+            my %influx_read_args = (
```
pdostal, line 157: I'm first of all sorry Anton that I'm late with such request so feel free to ignore it but didn't you think about creating separate test module for this analyzing code?

tests/publiccloud/storage_perf.pm
```diff
@@ -120,6 +154,33 @@ sub run {
...
+                os_version => $tags->{os_version}
+            );
+            if (db_has_data(%influx_read_args, timeframe => $this_week_timeframe) && db_has_data(%influx_read_args, timeframe => $previous_week_timeframe)) {
+                foreach my $load_type (keys %$values) {
+                    $influx_read_args{load_type} = $load_type;
+                    my $current_load_type = sprintf(" load_type=%s, scenario=%s, Flavor=%s, Version=%s", $influx_read_args{load_type}, $influx_read_args{scenario}, $influx_read_args{os_flavor}, $influx_read_args{os_version});
+                    my $this_week = get_median_from_db(%influx_read_args, timeframe => $this_week_timeframe);
+                    my $previous_week = get_median_from_db(%influx_read_args, timeframe => $previous_week_timeframe);
+                    if (defined($previous_week) && defined($this_week) && $previous_week != 0 && $this_week != 0) {
+                        record_info('ANALYZE', "Checking " . $current_load_type);
```
pdostal, line 171:
```suggestion
                        record_info('ANALYZE', "Checking " . $current_load_type);
                        record_info('This week', "Median for this week: " . $this_week);
                        record_info('Previous week', "Median for previous week: " . $previous_week);
```

pdostal (commented):
This looks already so great! I have only few minor suggestions:

tests/publiccloud/storage_perf.pm
asmorodskyi: no I can not say that you are too late , so from this perspective I don't see any issue in your request . And I think you at least partially right . I think it will be right idea to "pack"  this logic into dedicated function e.g. `analyse_previous_results` ( feel free to suggest name this one was generated within 5 seconds :) ) .
But I have concerns regarding separate module . I am using `$values` hash and I don't want to create copy of it in this separate module .
So do you think wrap this logic into a function will be good enough ? Or you want to continue conversation about separate module ? ( in such a case I am happy to hear ideas about `$values`  hash )

tests/publiccloud/storage_perf.pm
cfconrad: +1 to make some comment with example json here.
I would also guard this, so the arrays do really have this content. 
Other wise you are face the trap of https://perldoc.perl.org/perlglossary#autovivification ,but I think in this case it doesn't matter.

```bash
:~> perl -e 'use Mojo::Base -strict; use Data::Dumper; my @foo; say Dumper(\@foo); say ($foo[0][0][1] // "undef"); say Dumper(\@foo);'
$VAR1 = [];

undef
$VAR1 = [
          [
            []
          ]
        ];
```

tests/publiccloud/storage_perf.pm
pdostal: Separate subroutine is OK and the name is also nice!

jlausuch:
Looks very promising! Looking forward to some VRs :) 

I see your code deals with situation when "there are no previous builds to compare" with the `else` 
```
record_info('INFO', "Not enough data for " . $href->{name} . "\n");
```

What happens when there are `X < 5` builds? I imagine the condition `if (db_has_data(%influx_read_args)` still returns `true` and it compares against the first X available builds. Right?

Would it make sense to make the limit of `5` a bit more flexible? For example:
```
my $last_records = ' limit ' . get_var('PUBLIC_CLOUD_FIO_LIMIT', '5'); # or whatever variable name
```

tests/publiccloud/storage_perf.pm
jlausuch: yeah, not bad idea of separate method outside of `run` :) 
I also think separate module is not necessary.

jlausuch:
First time I see this :D
```
 tests/publiccloud/storage_perf.pm: Code structure is deeply nested at line 172, column 21.  Consider refactoring.  (Severity: 4)
tests/publiccloud/storage_perf.pm: Code structure is deeply nested at line 174, column 25.  Consider refactoring.  (Severity: 4)
```

asmorodskyi:
> What happens when there are `X < 5` builds? I imagine the condition `if (db_has_data(%influx_read_args)` still returns `true` and it compares against the first X available builds. Right?

[No](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13804/files#diff-71d5940ceca3c88d8d1ab4cfea14234cadabbd47649208edcaa73c6fe67a6e1cR50)  you will get true only when there is more than 10 rows in DB for certain scenario/version/flavor combination ( it comparing 5 rows with last 5 ) . 

```
my $last_records = ' limit ' . get_var('PUBLIC_CLOUD_FIO_LIMIT', '5'); # or whatever variable name
```

yeah sure we can do something like this . But I personally don't see why we would need it . When there is not enough data test will simply skip this check because there not enough data and I prefer test to do so . In any case lower than defined value more likely to give us false alarms . 

And yes maybe 5+5 is not the best option and we will need to tune this value later (most probably in bigger side ) but I don't see myself cloning the job to see if it still fails in 4+4 , 6+6 or 7+7 combinations . This is not the point of this test . The point of this test to invoke alarm when there is a chance of performance regression . And when I will see the failure I will go to http://openqa-perf.qa.suse.de to deeper data analysis

jlausuch:
> > What happens when there are `X < 5` builds? I imagine the condition `if (db_has_data(%influx_read_args)` still returns `true` and it compares against the first X available builds. Right?
> 
> [No](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13804/files#diff-71d5940ceca3c88d8d1ab4cfea14234cadabbd47649208edcaa73c6fe67a6e1cR50) you will get true only when there is more than 10 rows in DB for certain scenario/version/flavor combination ( it comparing 5 rows with last 5 ) .

Ok, then I got it wrong. Yes, that makes it more robust..  So, we need history of 10 previous builds (assuming all builds pushed proper data). I guess it should be ok, I've checked the history of 15-SP3 and we had ~20 builds between Public Beta and GMC. Thanks!

jlausuch:
Btw, you should replace `median` by `mean` everywhere. They are 2 different concepts and it can bring to confusion.

tests/publiccloud/storage_perf.pm
```diff
@@ -120,6 +156,33 @@ sub run {
...
+            );
+            if (db_has_data(%influx_read_args) {
+                foreach my $load_type (keys %$values) {
+                    $influx_read_args{load_type} = $load_type;
+                    my $current_load_type = sprintf(" load_type=%s, scenario=%s, Flavor=%s, Version=%s", $influx_read_args{load_type}, $influx_read_args{scenario}, $influx_read_args{os_flavor}, $influx_read_args{os_version});
+                    my $last_records_median = get_median_from_db(%influx_read_args, limit => $last_records);
+                    my $previous_records_median = get_median_from_db(%influx_read_args, limit => $previous_records);
+                    if (defined($previous_records_median) && defined($last_records_median) && $previous_records_median != 0 && $last_records_median != 0) {
+                        record_info('ANALYZE', "Checking " . $current_load_type);
+                        if ((abs($previous_records_median - $last_records_median) / $previous_records_median) * 100 > 10) {
```
jlausuch, line 174: I guess it's ok to leave the 10% tolerance for all the pieces of data, but think we will need to re-adjust depending on which data we are analysing. 
Maybe <10% is ok for throughput, but not for latency (just to give an example without proper reasoning). I guess we will learn and improve it after getting the first results :)

tests/publiccloud/storage_perf.pm
asmorodskyi: done

tests/publiccloud/storage_perf.pm
asmorodskyi: done

tests/publiccloud/storage_perf.pm
asmorodskyi: I can not fully follow your suggestions due to some changes . But I think I follow them as much as I can :)

tests/publiccloud/storage_perf.pm
asmorodskyi: yeah I would keep this for future improvements . no matter what we will do **now** it will be random approach which would need to be improved when we will get more data

asmorodskyi:
So from my POV I implemented all suggestions from everyone . running last minute checks and PR is ready to be merged

asmorodskyi:
passed VR - http://autobot.qa.suse.de/tests/2591# 

( @foursixnine  with real data )

asmorodskyi:
latest VR - http://autobot.qa.suse.de/tests/2605

grisu48 (approved):

grisu48 (approved):

tests/publiccloud/storage_perf.pm
```diff
@@ -17,6 +17,102 @@ use Mojo::JSON;
...
+    #    [{
+    #      'statement_id' => 0,
+    #      'series' => [{
+    #        'values' => [[ '1970-01-01T00:00:00Z', 0 ]],
+    #        'name' => 'storage',
+    #        'columns' => ['time','median_write_throughput']
+    #                  }]
+    #     }]
+    #};
+    # we want to return **value** of "median_write_throughput" (from example above)
```
jlausuch, line 55:
```suggestion
    # we want to return **value** of "mean_write_throughput" (from example above)
```

PR #13809, opened 2021-12-03
pam: Copy template config form /usr/etc if not existing in /etc yet

DimStar77:
pam moved its distro config files to /usr/etc/security, with the admin being expected to modify his copies in /etc. Prepare the test to do just that

- Related ticket: https://progress.opensuse.org/issues/103470
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/2076883

### DIFF
--- a/data/pam_test/pam.sh
+++ b/data/pam_test/pam.sh
@@ -15,18 +15,25 @@ ROOT_PW='nots3cr3t'
 setup() {
     # the dir /etc/pam.d/pam_test is static in the tool pam-test
     cp ./pam_test /etc/pam.d/pam_test
-    cp /etc/security/access.conf /etc/security/access.conf.bak
-    cp /etc/security/time.conf /etc/security/time.conf.bak
-    cp /etc/security/group.conf /etc/security/group.conf.bak
-    cp /etc/security/limits.conf /etc/security/limits.conf.bak
+    mkdir -p /etc/security
+    for file in access.conf time.conf group.conf limits.conf; do
+      if [ ! -f /etc/security/$file ]; then
+         cp /usr/etc/security/$file /etc/security/$file
+      else
+         cp /etc/security/$file /etc/security/$file.bak
+      fi
+    done
 }
 
 teardown() {
     rm -f /etc/pam.d/pam_test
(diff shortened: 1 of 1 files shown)


foursixnine (approved):
I guess this would be compatible with older releases

DimStar77:
> I guess this would be compatible with older releases

should well be, as I only copy files from /usr/etc if they do not exist in /etc

Vogtinator:
The intention of the backup and restore is to keep the system unmodified after the test completed, FWICT. This way it doesn't quite achive that anymore, because the `/usr/etc/` files were copied to `/etc`. No idea whether that can be an issue in practice, probably not?

data/pam_test/pam.sh
```diff
@@ -15,9 +15,14 @@ ROOT_PW='nots3cr3t'
 setup() {
     # the dir /etc/pam.d/pam_test is static in the tool pam-test
     cp ./pam_test /etc/pam.d/pam_test
+    test ! -d /etc/security && mkdir /etc/security
+    test ! -f /etc/security/access.conf && cp /usr/etc/security/access.cof /etc/security/access.conf
```
Vogtinator, line 19: Typo

Vogtinator (changes requested):
There's a typo.

What about just making a backup if those files exist and for restoring just delete the config file if there's no backup? That's how I handled it for openscap in 1c1eba73a0f797cef606bf8e410d36e9019079e7. No idea how the test behaves with that.

data/pam_test/pam.sh
DimStar77: typo fixed

DimStar77:
> The intention of the backup and restore is to keep the system unmodified after the test completed, FWICT. This way it doesn't quite achive that anymore, because the `/usr/etc/` files were copied to `/etc`. No idea whether that can be an issue in practice, probably not?

as they are 1:1 copies, the files in /etc should just take precedence over the ones in /usr/etc, but being equal, the result should be close (enough) to equal (granted, not perfect)

Vogtinator (approved):

PR #13810, opened 2021-12-03
nproc_limits: cater for pam's default config files in /usr/etc

DimStar77:
- Related ticket: related to https://progress.opensuse.org/issues/103470
- Needles: N/A
- Verification run: https://openqa.opensuse.org/t2069528

### DIFF
--- a/tests/security/nproc_limits.pm
+++ b/tests/security/nproc_limits.pm
@@ -17,6 +17,7 @@ sub run {
     $self->select_serial_terminal;
 
     # Check the limits.conf config file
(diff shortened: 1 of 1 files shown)


foursixnine (approved):

tests/security/nproc_limits.pm
```diff
@@ -17,6 +17,7 @@ sub run {
     $self->select_serial_terminal;
 
     # Check the limits.conf config file
+    script_run('test ! -f /etc/security/limits.conf && install /usr/etc/security/limits.conf -t /etc/security');
```
Vogtinator, line 20: Not fully correct. If `/etc/security/limts.conf` exists, it ignores `/usr/etc/security/limits.conf.d/`

PR #13811, opened 2021-12-03
Disable online repos for openSUSE Tumbleweed

rakoenig:
Disable the use of online repos by declining the popup.

- Related ticket: https://progress.opensuse.org/issues/102188
- Needles: not needed because using libyui
- Verification run: https://openqa.opensuse.org/tests/overview?build=rakoenig_disable_online_repos

### DIFF
--- a/lib/Installation/Popups/YesNoPopup.pm
+++ b/lib/Installation/Popups/YesNoPopup.pm
@@ -22,7 +22,7 @@ sub new {
 sub init {
     my $self = shift;
     $self->{btn_yes} = $self->{app}->button({id => 'yes'});
-    $self->{btn_no} = $self->{app}->button({id => 'no'});
+    $self->{btn_no} = $self->{app}->button({id => qr/^(no|no_button)$/});
     $self->{lbl_warning} = $self->{app}->label({type => 'YLabel'});
     return $self;
 }
--- a/schedule/yast/opensuse/btrfs/btrfs+warnings.yaml
+++ b/schedule/yast/opensuse/btrfs/btrfs+warnings.yaml
@@ -15,7 +15,7 @@ schedule:
   - installation/setup_libyui
   - '{{access_beta_distribution}}'
   - installation/licensing/accept_license
-  - installation/online_repos
+  - installation/online_repos/disable_online_repos
   - installation/installation_mode
   - installation/logpackages
   - installation/system_role/select_role_desktop_with_KDE_plasma
--- a/schedule/yast/opensuse/btrfs/btrfs.yaml
+++ b/schedule/yast/opensuse/btrfs/btrfs.yaml
@@ -14,7 +14,7 @@ schedule:
   - installation/setup_libyui
   - '{{access_beta_distribution}}'
   - installation/licensing/accept_license
-  - installation/online_repos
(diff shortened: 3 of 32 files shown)


lib/Installation/Popups/YesNoPopup.pm
```diff
@@ -22,7 +22,9 @@ sub new {
 sub init {
     my $self = shift;
     $self->{btn_yes} = $self->{app}->button({id => 'yes'});
-    $self->{btn_no} = $self->{app}->button({id => 'no'});
+    # The id of the "No" button varies depending on the YaST version.
+    # Therefore we use a regex
+    $self->{btn_no} = $self->{app}->button({id => qr/no|no_button/});
```
ge0r, line 27: Nice one!

OleksandrOrlov: > Nice one!

Agree:)

OleksandrOrlov: @rakoenig Does your VR already includes this code? Just wonder if it works correctly.

rakoenig: Yes, the code ist in the VR. The late changes were only affecting 3 YAML files that hat a conflict when I was rebasing to the current master.

jknphy: I would prefer to use `label` here, which is valid for both products and doesn't get the whole json with a regex (regex should be used when there is not other way), It would not need a comment like that (I really prefer not read comments when there is not need and for us is obvious that products are different) and also I think this would match any string after `no*`, so if the screen would include some `no_whatever` will also match it, unlikely but confusing, when label would be really clear. wdyt?

jknphy:
Taking advantage that you are doing verification, could you please take care of removing from schedules `installation_mode` as well where it is not doing anything, as suggested in the ticket?

jknphy:
Regarding the commit title, the test already exists, so we should state that we implement it with YuiRestClient/libyui-rest-api or similar.

jknphy:
I would suggest to run at least one verification in textmode and one for other architectures.

lib/Installation/Popups/YesNoPopup.pm
```diff
@@ -22,7 +22,7 @@ sub new {
 sub init {
     my $self = shift;
     $self->{btn_yes} = $self->{app}->button({id => 'yes'});
-    $self->{btn_no} = $self->{app}->button({id => 'no'});
+    $self->{btn_no} = $self->{app}->button({id => qr/^no$|^no_button$/});
```
jknphy, line 25: if you still like this idea, this is more clear to read: `^(no|no_button)$

jknphy:
also we need some VR for some job where the dialog was already used before.

lib/Installation/Popups/YesNoPopup.pm
rakoenig: I stick with the regex.

lib/Installation/Popups/YesNoPopup.pm
rakoenig: Changed that and restarted all VRs.

PR #13816, opened 2021-12-07
Add NVIDIA compute to the list of expected modules

OleksandrOrlov:
on x64 and aarch64 the NVIDIA compute module is added. The commit adds
the module to the list of expected data in installer_extended test
suite.

- Verification run: https://openqa.suse.de/tests/7802236#step/verify_module_selection/1


rakoenig (approved):
LGTM

PR #13818, opened 2021-12-07
ltp: Add option to install selected parts from git

czerw:
Fix poo#103608: LTP package is built on daily basis. It can happen
that kernel modules are built against different kernel, than is in
tested snapshot. Variable LTP_GIT_URL allows to install
specific parts from git directly on SUT after was LTP installed from
package.

- Related ticket: https://progress.opensuse.org/issues/103608
- Needles: 
- Verification run: 
install: https://openqa.suse.de/tests/7802590

before: https://openqa.suse.de/tests/7794772#step/insmod01_sh/8
after: https://openqa.suse.de/tests/7802601#step/insmod01_sh/8

### DIFF
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -169,10 +169,9 @@ sub install_build_dependencies {
     }
 }
 
-sub install_from_git {
+sub prepare_ltp_git {
     my $url = get_var('LTP_GIT_URL', 'https://github.com/linux-test-project/ltp');
     my $rel = get_var('LTP_RELEASE');
-    my $timeout = (is_aarch64 || is_s390x) ? 7200 : 1440;
     my $prefix = get_ltproot();
     my $configure = "./configure --with-open-posix-testsuite --with-realtime-testsuite --prefix=$prefix";
     my $extra_flags = get_var('LTP_EXTRA_CONF_FLAGS', '');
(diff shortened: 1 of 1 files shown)


tests/kernel/install_ltp.pm
```diff
@@ -344,6 +366,10 @@ sub run {
     else {
         add_ltp_repo;
         install_from_repo();
+        if (get_var("LTP_INSTALL_SELECTED")) {
```
mdoucha, line 369: LGTM, but I think there's no need to add a new variable. It'd be better to check `LTP_GIT_URL` instead.

mdoucha: Ideally, it'd be better to build the modules from LTP source package for version stability reasons but this is good enough for now.

czerw: Ok changed to `LTP_GIT_URL`. This problem is only for for tests using daily ltp package builds, so fixed version is not needed.

mdoucha (approved):
LGTM

tests/kernel/install_ltp.pm
```diff
@@ -186,6 +185,29 @@ sub install_from_git {
...
+    my @paths = qw(commands/insmod
+      kernel/firmware
+      kernel/device-drivers
+      kernel/syscalls/delete_module
+      kernel/syscalls/finit_module
+      kernel/syscalls/init_module);
+
+    assert_script_run('pushd testcases');
+    foreach (@paths) {
+        assert_script_run("pushd $_ && make && make install && popd", timeout => 600);
```
pevik, line 201: nit: Instead `pushd` and `popd` I'd prefer `cd DIR` and `cd -`, but that's really personal (not asking for removing it).

pevik (approved):
LGTM, thanks!

PR #13819, opened 2021-12-07
Remove disk_activation from textmode@pvm

OleksandrOrlov:
The disk_activation test module was mistakenly added to the schedule, so
the commit removes it.

- Verification run: https://openqa.suse.de/tests/7802628


ge0r (approved):
Well spotted!

PR #13821, opened 2021-12-08
x11/remote_desktop/xrdp_client: Disable H.264/AVC by default

Vogtinator:
We can't ship FreeRDP with H.264 support out of the box, but Remmina uses that
as default. After authentication, Remmina shows a message that a non-AVC color
depth needs to be chosen, so do that by writing into the settings file.

- Verification run: https://openqa.opensuse.org/tests/2076144

### DIFF
--- a/tests/x11/remote_desktop/xrdp_client.pm
+++ b/tests/x11/remote_desktop/xrdp_client.pm
@@ -29,8 +29,11 @@ sub run {
     # Disable Remmina news before launch Remmina
     x11_start_program('xterm');
     my $pref_dir = '~/.config/remmina';
-    assert_script_run "mkdir $pref_dir";
+    assert_script_run "mkdir -p $pref_dir";
     assert_script_run 'echo -e "[remmina_news]\\nperiodic_rmnews_last_get=$(date +%s)" >> ' . $pref_dir . '/remmina.pref';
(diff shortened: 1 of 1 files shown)


DimStar77 (approved):

GraceWang571:
Nice and quick fix.
One question about the warning message after authentication https://openqa.opensuse.org/tests/2075300#step/xrdp_client/46
Is there a button not able to show on the right side of the message?

Vogtinator:
> Nice and quick fix. One question about the warning message after authentication https://openqa.opensuse.org/tests/2075300#step/xrdp_client/46 Is there a button not able to show on the right side of the message?

Yeah, the window is too wide for the 1024x768px screen. There's only the button to "Close" the message.

GraceWang571:
> Yeah, the window is too wide for the 1024x768px screen. There's only the button to "Close" the message.

Well, IMHO this is actually a bug. https://bugzilla.opensuse.org/show_bug.cgi?id=1193746 was filed to track it.

Vogtinator:
> > Yeah, the window is too wide for the 1024x768px screen. There's only the button to "Close" the message.
> 
> Well, IMHO this is actually a bug. https://bugzilla.opensuse.org/show_bug.cgi?id=1193746 was filed to track it.

Agreed, albeit a minor one. Thanks!

PR #13822, opened 2021-12-08
kernel/install_ltp: Use new repo name convention on Leap 15.4+

Vogtinator:
The "openSUSE_Leap_" part is gone from the name on 15.4+.
On 15.3+, all arches are available, so drop the conditions there.

- Verification run: https://openqa.opensuse.org/tests/2076246

### DIFF
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -229,18 +229,19 @@ sub add_ltp_repo {
         }
 
         # ltp for leap15.2 is available only x86_64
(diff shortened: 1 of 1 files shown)


pevik:
Thanks!

mloviska:
Seems like the convention has changed again https://download.opensuse.org/repositories/benchmark:/ltp:/devel/openSUSE_Leap_15.4/

mloviska:
> 
> Should be changed back - "15.4" is the global default name

Good, thanks for the input.

PR #13825, opened 2021-12-08
SLE Micro: soft fail bsc#1191897

jlausuch:
Agreed with devs that this will be fixed by end of January.
Meanwhile, we shall soft-fail the test.
https://bugzilla.suse.com/show_bug.cgi?id=1191897

VR: https://openqa.suse.de/tests/7811443


dzedro (approved):

jlausuch:
I have created https://progress.opensuse.org/issues/103754 so we don't forget to revert this when bugs are fixed.

PR #13827, opened 2021-12-08
SLE Micro: soft fail bsc#1191863

jlausuch:
Agreed with devs that this will be fixed by end of January.
Meanwhile, we shall soft-fail the test.
https://bugzilla.suse.com/show_bug.cgi?id=1191863

VR: https://openqa.suse.de/tests/7812215
NOTE: This VR fails because it also needs https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13825

### DIFF
--- a/tests/transactional/trup_smoke.pm
+++ b/tests/transactional/trup_smoke.pm
@@ -12,7 +12,8 @@ use strict;
 use warnings;
 use testapi;
 use transactional;
-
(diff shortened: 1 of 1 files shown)


dzedro (approved):

jlausuch:
I have created https://progress.opensuse.org/issues/103754 so we don't forget to revert this when bugs are fixed.

PR #13828, opened 2021-12-09
Fix can't upload logs in yast2_migration,start_install and zypper migration

hjluo:
In zypper_migration, yast2_migration and start_install, when something breaks
the upgrade process, such as conflicts, system errors, we need to get the yast2log
or zypper.log to quickly file bugs or tickets.

- Related ticket: https://progress.opensuse.org/issues/92875
- Needles: N/A
- Verification run: 
zypper_migraion:

https://openqa.nue.suse.com/tests/7843071/file/zypper_migration-zypper.log 
https://openqa.nue.suse.com/tests/7850224/file/zypper_migration-zypper.log

yast2_migration:
https://openqa.nue.suse.com/tests/7810102 
https://openqa.nue.suse.com/tests/7810071 
start_install:
https://openqa.nue.suse.com/t7850096 has log "https://openqa.nue.suse.com/tests/7850096/file/start_install-y2logs.tar.bz2"
https://openqa.nue.suse.com/t7850097 has log "https://openqa.nue.suse.com/tests/7850097/file/start_install-y2logs.tar.bz2"

### DIFF
--- a/tests/migration/online_migration/zypper_migration.pm
+++ b/tests/migration/online_migration/zypper_migration.pm
@@ -165,10 +165,11 @@ sub run {
 
 sub post_fail_hook {
     my $self = shift;
-    $self->SUPER::post_fail_hook;
     $self->select_serial_terminal;
     script_run("pkill zypper");
+    upload_logs '/var/log/zypper.log';
     $self->upload_solvertestcase_logs();
+    $self->SUPER::post_fail_hook;
 }
 
 1;
--- a/lib/y2_base.pm
+++ b/lib/y2_base.pm
@@ -110,7 +110,7 @@ sub save_system_logs {
 sub save_upload_y2logs {
     my ($self, %args) = @_;
 
-    return if (get_var('NOLOGS'));
+    return if (get_var('NOLOGS') || get_var('Y2LOGS_UPLOADED'));
     $args{suffix} //= '';
 
     # Do not test/recover network if collect from installation system, as it won't work anyway with current approach
--- a/tests/migration/online_migration/yast2_migration.pm
+++ b/tests/migration/online_migration/yast2_migration.pm
@@ -288,6 +288,8 @@ sub test_flags {
 sub post_fail_hook {
     my ($self) = @_;
     select_console 'log-console';
+    $self->save_upload_y2logs;
+    set_var('Y2LOGS_UPLOADED', 1);
     $self->save_and_upload_log('journalctl -b -o short-precise', '/tmp/journal.log', {screenshot => 1});
     $self->export_logs_desktop;
     $self->SUPER::post_fail_hook;


tests/installation/start_install.pm
```diff
@@ -134,4 +134,10 @@ sub run {
     }
 }
 
+sub post_fail_hook {
+    my $self = shift;
+    $self->save_upload_y2logs if check_var('UPGRADE', '1');
+    $self->SUPER::post_fail_hook;
```
lemon-suse, line 140: This will lead to upload two times y2log.

hjluo: Done, set_var('Y2LOGS-UPLOADED',1) to fixt this.

tests/migration/online_migration/zypper_migration.pm
```diff
@@ -165,10 +165,11 @@ sub run {
 
 sub post_fail_hook {
     my $self = shift;
-    $self->SUPER::post_fail_hook;
     $self->select_serial_terminal;
+    upload_logs '/var/log/zypper.log';
```
lemon-suse, line 169: I think move this line after kill zypper will avoid zypper process still write to zypper.log.

hjluo: will try,  but in this case, the log will not change. even we killed zypper. then next zypper command will complain some previous zypper process was running and can't work. cause
it writes the process ID in some files, but I can't find it.

tests/installation/start_install.pm
```diff
@@ -134,4 +134,18 @@ sub run {
     }
 }
 
+sub post_fail_hook {
+    my $self = shift;
+    if (check_var('UPGRADE', '1')) {
+        select_console 'log-console';
+        save_screenshot;
```
rfan1, line 141: Any reason save_screenshot here?

tests/installation/start_install.pm
```diff
@@ -134,4 +134,18 @@ sub run {
     }
 }
 
+sub post_fail_hook {
+    my $self = shift;
+    if (check_var('UPGRADE', '1')) {
```
lemon-suse, line 139: This will cause migration test can't get full log, I think you can add y2log upload here and overload the SUPER::post_fail_hook.

rfan1:
@hjluo 
How about abs: Enhance upload y2log logic in yast2_migration,start_install when experienced package conflicts

hjluo:
> @hjluo How about abs: Enhance upload y2log logic in yast2_migration,start_install when experienced package conflicts

Done

tinawang123:
Enhance upload y2log logic in yast2_migration,start_install when experienced package conflicts
I think it not only supports package conflicts problem, It also supports any error happened during the yast2_migration and start_install

tests/installation/start_install.pm
hjluo: In some cases, the curl upload will fail, saving pictures here will help us to find the reason
quickly.

hjluo:
> Enhance upload y2log logic in yast2_migration,start_install when experienced package conflicts I think it not only supports package conflicts problem, It also supports any error happened during the yast2_migration and start_install

Changed. thanks

tests/installation/start_install.pm
hjluo: this is what we need for investigation. before this with default SUPER::post_fail_hook
we can't get anything. see https://openqa.nue.suse.com/tests/7843085#downloads

chcao:
For offline media cases, cases may fail at upgrade_select, will save and upload y2log for these cases in this PR please?

hjluo:
> For offline media cases, cases may fail at upgrade_select, will save and upload y2log for these cases in this PR, please?

Yes,  the fail hook function will save the save_y2logs and upload them.

tests/installation/start_install.pm
lemon-suse: If this is just a workaround, please just upload log needed then  $self->SUPER::post_fail_hook; and add mutex setting to avoid upload log twice, else need investigate why can't upload any load with default post_fail_hook.

lemon-suse:
I suggest to simply the title to 'Fix can't upload logs in yast2_migration,start_install and zypper migration' , do not need say the detail failure, in fact, failures will happen not only these you listed.

hjluo:
> I suggest to simply the title to 'Fix can't upload logs in yast2_migration,start_install and zypper migration' , do not need say the detail failure, in fact, failures will happen not only these you listed.

Done

tests/installation/start_install.pm
```diff
@@ -134,4 +134,17 @@ sub run {
...
+        select_console 'log-console';
+        save_screenshot;
+        script_run "save_y2logs /tmp/y2logs.tar.bz2";
+        upload_logs "/tmp/y2logs.tar.bz2";
```
lemon-suse, line 143: I think this need 'set_var('Y2LOGS_UPLOADED', 1);' to avoid duplicate upload.

hjluo: already added at line 145:  set_var('Y2LOGS_UPLOADED', 1);

lemon-suse: ok

okurz (changes requested):
This is a rather complicated approach and should not be needed because earlier test modules already check for conflicts. Take a look for example in https://openqa.nue.suse.com/tests/7850096#step/resolve_dependency_issues/2 where a needle does not correctly match. If it would have matched then manual conflict resolution would have been applied there. I suggest to update resolve_dependency_issues to fail the test module unless "no conflicts" have been found

hjluo:
> This is a rather complicated approach and should not be needed because earlier test modules already check for conflicts. Take a look for example in https://openqa.nue.suse.com/tests/7850096#step/resolve_dependency_issues/2 where a needle does not correctly match. If it would have matched then manual conflict resolution would have been applied there. I suggest to update resolve_dependency_issues to fail the test module unless "no conflicts" have been found

The PR is to collect the  start_install-y2logs.tar.bz2 here to know which package conflict/errors 
which we needed to file a bug/ticket, in our migration, we can't accept manual intervention. 
that's it.

okurz:
> The PR is to collect the start_install-y2logs.tar.bz2 here to know which package conflict/errors which we needed to file a bug/ticket, in our migration, we can't accept manual intervention. that's it.

That's not the point. You don't need to collect yast logs from start_install when it should have failed already in before

hjluo:
> > The PR is to collect the start_install-y2logs.tar.bz2 here to know which package conflict/errors which we needed to file a bug/ticket, in our migration, we can't accept manual intervention. that's it.
> 
> That's not the point. You don't need to collect yast logs from start_install when it should have failed already in before

OK. I've removed it from the start_install.pm, could you please take a look again? thanks for your
help.

okurz (approved):
It's ok. I still recommend that you follow the suggestion about the resolve conflict module to make the needle mismatch obvious with a test failure

hjluo:
> It's ok. I still recommend that you follow the suggestion about the resolve conflict module to make the needle mismatch obvious with a test failure

OK. I'd file ticket later for that, and thank you very much for your review.

lemon-suse (approved):

PR #13829, opened 2021-12-09
remove gnome x11 patterns in sles12sp4PV_PRG.xml

tonyyuan1:
12sp4PV_PRG.xml includes gnome-basic and x11 patterns which makes guest installation longer than other guests. This caused waitfor_guests to fail. These patterns are not needed in guest installation.  Remove them from installation  

- Related ticket: https://progress.opensuse.org/issues/103647
- Needles: no
- Verification run: 
- http://openqa.qam.suse.cz/tests/32718#step/waitfor_guests/1
- Full xen tests on sles12sp4: http://openqa.qam.suse.cz/tests/overview?distri=sle&groupid=161&version=12-SP4&build=%3Atony%3Apoo103647

### DIFF
--- a/data/autoyast_xen/sles12sp4PV_PRG.xml
+++ b/data/autoyast_xen/sles12sp4PV_PRG.xml
@@ -51,51 +51,8 @@
     <image_installation config:type="boolean">false</image_installation>
   </deploy_image>
   <firewall>
-    <FW_ALLOW_FW_BROADCAST_DMZ>no</FW_ALLOW_FW_BROADCAST_DMZ>
-    <FW_ALLOW_FW_BROADCAST_EXT>no</FW_ALLOW_FW_BROADCAST_EXT>
-    <FW_ALLOW_FW_BROADCAST_INT>no</FW_ALLOW_FW_BROADCAST_INT>
-    <FW_BOOT_FULL_INIT>no</FW_BOOT_FULL_INIT>
-    <FW_CONFIGURATIONS_DMZ>sshd</FW_CONFIGURATIONS_DMZ>
-    <FW_CONFIGURATIONS_EXT>hamsta sshd</FW_CONFIGURATIONS_EXT>
-    <FW_CONFIGURATIONS_INT>sshd</FW_CONFIGURATIONS_INT>
-    <FW_DEV_DMZ/>
-    <FW_DEV_EXT/>
-    <FW_DEV_INT/>
-    <FW_FORWARD_ALWAYS_INOUT_DEV/>
-    <FW_FORWARD_MASQ/>
-    <FW_IGNORE_FW_BROADCAST_DMZ>no</FW_IGNORE_FW_BROADCAST_DMZ>
-    <FW_IGNORE_FW_BROADCAST_EXT>yes</FW_IGNORE_FW_BROADCAST_EXT>
-    <FW_IGNORE_FW_BROADCAST_INT>no</FW_IGNORE_FW_BROADCAST_INT>
-    <FW_IPSEC_TRUST>no</FW_IPSEC_TRUST>
-    <FW_LOAD_MODULES>nf_conntrack_netbios_ns</FW_LOAD_MODULES>
-    <FW_LOG_ACCEPT_ALL>no</FW_LOG_ACCEPT_ALL>
-    <FW_LOG_ACCEPT_CRIT>yes</FW_LOG_ACCEPT_CRIT>
-    <FW_LOG_DROP_ALL>no</FW_LOG_DROP_ALL>
-    <FW_LOG_DROP_CRIT>yes</FW_LOG_DROP_CRIT>
-    <FW_MASQUERADE>no</FW_MASQUERADE>
-    <FW_PROTECT_FROM_INT>no</FW_PROTECT_FROM_INT>
-    <FW_ROUTE>no</FW_ROUTE>
-    <FW_SERVICES_ACCEPT_DMZ/>
-    <FW_SERVICES_ACCEPT_EXT/>
-    <FW_SERVICES_ACCEPT_INT/>
-    <FW_SERVICES_ACCEPT_RELATED_DMZ/>
-    <FW_SERVICES_ACCEPT_RELATED_EXT/>
-    <FW_SERVICES_ACCEPT_RELATED_INT/>
-    <FW_SERVICES_DMZ_IP/>
-    <FW_SERVICES_DMZ_RPC/>
-    <FW_SERVICES_DMZ_TCP/>
-    <FW_SERVICES_DMZ_UDP/>
-    <FW_SERVICES_EXT_IP/>
-    <FW_SERVICES_EXT_RPC/>
-    <FW_SERVICES_EXT_TCP/>
-    <FW_SERVICES_EXT_UDP/>
-    <FW_SERVICES_INT_IP/>
-    <FW_SERVICES_INT_RPC/>
-    <FW_SERVICES_INT_TCP/>
-    <FW_SERVICES_INT_UDP/>
-    <FW_STOP_KEEP_ROUTING_STATE>no</FW_STOP_KEEP_ROUTING_STATE>
-    <enable_firewall config:type="boolean">true</enable_firewall>
-    <start_firewall config:type="boolean">true</start_firewall>
+    <enable_firewall config:type="boolean">false</enable_firewall>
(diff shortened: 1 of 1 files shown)


data/autoyast_xen/sles12sp4PV_PRG.xml
```diff
@@ -1094,36 +1051,13 @@ DefaultPolicy default
...
+      <disable config:type="list">
+      <service>SuSEfirewall2</service>
+      </disable>
       <enable config:type="list">
-        <service>apparmor</service>
-        <service>btrfsmaintenance-refresh</service>
-        <service>cron</service>
-        <service>display-manager</service>
-        <service>haveged</service>
-        <service>irqbalance</service>
```
grisu48, line 1106: `irqbalance` we probably still need for the xen test runs, no?

data/autoyast_xen/sles12sp4PV_PRG.xml
```diff
@@ -1094,36 +1051,13 @@ DefaultPolicy default
...
-        <service>ntpd</service>
-        <service>postfix</service>
-        <service>purge-kernels</service>
-        <service>rollback</service>
-        <service>rsyslog</service>
-        <service>smartd</service>
         <service>sshd</service>
-        <service>SuSEfirewall2</service>
-        <service>SuSEfirewall2_init</service>
-        <service>wicked</service>
```
grisu48, line 1119: Isn't `wicked` required for the networking?

data/autoyast_xen/sles12sp4PV_PRG.xml
```diff
@@ -1132,55 +1066,12 @@ DefaultPolicy default
     <install_recommended config:type="boolean">true</install_recommended>
     <instsource/>
     <packages config:type="list">
-      <package>sudo</package>
```
grisu48, line 1135: I believe we still need some of the basic packages. I would leave at least the following in here:

* `sudo`
* `binutils`
* `iputils`

In addition I think changes on the packages list requires a full VR.

grisu48 (commented):
Some remarks that should be discussed. In addition I'd recommend to do a full VR if we are reducing the number of base packages, as there can be unintended side-effects that break the test runs.

Overall a good PR!

data/autoyast_xen/sles12sp4PV_PRG.xml
pdostal: Yes :D We need all of those.

alice-suse:
I agree with Felix and Pavel. If gnome x11 patterns need to be removed, suggest to do as limited autoyast code change as possible to avoid potential other side effects. And a full validation job is better -- from qam-xen-install until qam-xen-snapshots.

tonyyuan1:
> Some remarks that should be discussed. In addition I'd recommend to do a full VR if we are reducing the number of base packages, as there can be unintended side-effects that break the test runs.
> 
> Overall a good PR!
> 
> How did you generated the new autoyast profile? Did you create a new one or just edit the existing one?

I edited the existing one.

data/autoyast_xen/sles12sp4PV_PRG.xml
tonyyuan1: Yast have default values for all services. The above services are impotent for OS.  Yast will enable them by default. We don't need to explicitly enable these services in autoyast profile. It's something like: let installer decided what 's best for the OS installation and configuration.

data/autoyast_xen/sles12sp4PV_PRG.xml
tonyyuan1: It's required and enabled. We don't need to explicitly enable it in autoyast profile

data/autoyast_xen/sles12sp4PV_PRG.xml
tonyyuan1: Yes, these are basic packages. However, we already specify minimal and base patterns in profile. These packages are included in patterns so they will be installed. We don't need to explicitly specify them again in packages list. 

Yes, a full VR makes sense.

tonyyuan1:
> I agree with Felix and Pavel. If gnome x11 patterns need to be removed, suggest to do as limited autoyast code change as possible to avoid potential other side effects. And a full validation job is better -- from qam-xen-install until qam-xen-snapshots.

Will do that.

data/autoyast_xen/sles12sp4PV_PRG.xml
grisu48: Ahhh, I understand, now it makes sense. If the VR is passing, then this is good!

grisu48 (approved):
LGTM, if the VRs are passing. Here I'd recomment to do a full kvm and xen test run on one SLES version, then we can merge.

tonyyuan1:
Full xen tests is done on sles12sp4 hypervisor, The link is VR part.  Since this is a PV guest profile only, we don't need run kvm tests:). 

Thanks!

alice-suse:
Great @tonyyuan1 !

pdostal:
Why is this only for 12-SP4 PV XEN? I don't understand.

pdostal:
> Full xen tests is done on sles12sp4 hypervisor, The link is VR part. Since this is a PV guest profile only, we don't need run kvm tests:).
This is not true as we may need to switch to HVM.

And also please note that the same is true for [KVM](http://openqa.qam.suse.cz/tests/33177#step/virtmanager_rm_devices/61).

alice-suse:
> > Full xen tests is done on sles12sp4 hypervisor, The link is VR part. Since this is a PV guest profile only, we don't need run kvm tests:).
> > This is not true as we may need to switch to HVM.
> 
> And also please note that the same is true for [KVM](http://openqa.qam.suse.cz/tests/33177#step/virtmanager_rm_devices/61).

@pdostal  @tonyyuan1 
Oh, did I merge too early? Sorry for that. I thought all comments were given in the first round review and earlier merge would make it work sooner.

Please continue discussion, and if needed for HVM guest also, please open a new PR.

tonyyuan1:
> ~Why is this only for 12-SP4 PV XEN? I don't understand.~ Now I understand why this is for 12-SP4 only. It would tho still be nice to keep 12-SP4HVM guest up to date.

That's ok. Better do it in a new PR as this's merged already.

tonyyuan1:
> > Full xen tests is done on sles12sp4 hypervisor, The link is VR part. Since this is a PV guest profile only, we don't need run kvm tests:).
> > This is not true as we may need to switch to HVM.
> 
> And also please note that the same is true for [KVM](http://openqa.qam.suse.cz/tests/33177#step/virtmanager_rm_devices/61).

Since virt-manager tests require GUI, I prefer not to change any current guest profiles( keeping gome pattern for sles12sp4PV). 
To to make it simple, we can use plan B:  increase wait time from 1800 to 3000. 

 "$sleep_delay = 1800 if (is_xen_host); in tests/virtualization/universal/waitfor_guests.pm

Does anyone agree with me?

pdostal:
@tonyyuan1 the `virt-manager` runs on the host so the purpose of this PR is good. You just forgot about KVM.
With sles12sp4HVM you're right that we don't use it. We just used to keep it in sync in case we need it.

I in general like this PR because you're right that we don't need desktop on guests but some packages (which we manually added intentionaly (f.e. `sudo` because of [this error](http://openqa.qam.suse.cz/tests/32471#step/patch_guests/95) ) ) would be nice to keep.

tonyyuan1:
@pdostal your link to the kvm test made me look closer at those guest screenshots (needles). :)  The screenshot is appreciably graphical gdm login screen for 12sp4KVM guest.  
http://openqa.qam.suse.cz/tests/33177#step/virtmanager_add_devices/81

The same goes for the xen test I ran.   Look at 12sp4HVM guest.  http://openqa.qam.suse.cz/tests/33187#step/virtmanager_offon/150

If  gnome had not been installed in these two guests the virtmanager test might have broken, right?

That's why I changed my mind and want to chose plan B.

pdostal:
> If gnome had not been installed in these two guests the virtmanager test might have broken, right?

No. There is another needle for console login screen.

tonyyuan1:
> > If gnome had not been installed in these two guests the virtmanager test might have broken, right?
> 
> No. There is another needle for console login screen.

Good, still stay with Plan A. I will update 12sp4KVM and HVM profiles in another PR.  I think we can let this PR go.

PR #13831, opened 2021-12-09
Validate partition via parted in xfs@s390_zVM

OleksandrOrlov:
After the schedules for xfs s390x_svirt and s390x_zVM were separated,
the validate_partition_table_via_blkd was used for zVM test, that is
incorrect. It should use parted for that validation.

The commit fixes the issue by scheduling the proper test module.

- Verification run: https://openqa.suse.de/tests/7815953


ge0r (approved):
Nice, thanks for the fix!

PR #13833, opened 2021-12-09
Mark console/journal_check as non-fatal

Vogtinator:
Unrecognized errors in the journal are no reason to abort the test, or roll back
if snapshots are supported.

- Verification run: https://openqa.opensuse.org/tests/2077754


jlausuch:
Agree!
I guess the reason  `fatal=>0` was not defined before is because normally this test is run at the end. But this makes total sense.

jlausuch (approved):

PR #13844, opened 2021-12-10
Add automation test for RMT behind proxy

lemon-suse:
We need add automation test for RMT behind proxy instead of manual test.

- Related ticket: https://progress.opensuse.org/issues/101337
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/7820330#


tinawang123:
I think you should add description to how to use this file.
As it need configure proxy server first.
It's better to add how to configure for other people.

tests/console/rmt/rmt_proxy.pm
```diff
@@ -0,0 +1,53 @@
...
+    select_console 'root-console';
+    record_info('RMT server setup', 'Start to setup a rmt server');
+    rmt_wizard();
+    # sync from SCC
+    rmt_sync;
+
+    # Set RMT behind a proxy (with NTLM Auth), and ensure RMT basic function works.
+    record_info('Set proxy', 'Set RMT behind a proxy (with NTLM Auth)');
+    # set the proxy setting to rmt.conf
+    my $proxy_url = 'http:\/\/' . get_required_var('PROXY_SERVER');
```
coolgw, line 28: What kind of configuration needed for this PROXY_SERVER?  
I suppose this case will be a multi machine setup?

tests/console/rmt/rmt_proxy.pm
```diff
@@ -0,0 +1,53 @@
...
+    my $rmt_conf = '/etc/rmt.conf';
+    my $rmt_conf_bk = '/etc/rmt.conf.bk';
+
+    # backup the rmt conf file
+    assert_script_run("cp $rmt_conf $rmt_conf_bk");
+
+    # update the rmt conf file with the proxy setting
+    assert_script_run("sed -i 's/proxy: /proxy: $proxy_url/g' $rmt_conf");
+    assert_script_run("sed -i 's/proxy_auth: /proxy_auth: ntlm/g' $rmt_conf");
```
coolgw, line 37: what's the user or domain for ntlm auth?

tests/console/rmt/rmt_proxy.pm
```diff
@@ -0,0 +1,53 @@
...
+    assert_script_run("systemctl restart rmt-server");
+
+    # verify the basic function of the rmt server
+    rmt_sync;
+    assert_script_run("rmt-cli product enable SLES/12.5/x86_64");
```
coolgw, line 42: Is there any other way to confirm rmt still work instead of hardcode some product name?

chcao:
For rmt testing, we test the rmt dev version. So in the test setting, need to add DEV_PATH setting and then run a verification case.

tests/console/rmt/rmt_proxy.pm
lemon-suse: I don't think we must run multi-machine for this test, it just need a proxy server to support proxy service is enough.

tests/console/rmt/rmt_proxy.pm
lemon-suse: Maybe you know other method to do this? In fact this is the automation for manual test, just follow the requirement from manual test steps.

coolgw: Is there any simple rmt-cli command to just confirm connection with scc ? For example check the connection status of scc, not enable some specific product. We should try best avoid this hardcode

tests/console/rmt/rmt_proxy.pm
coolgw: If other team use this module then we will face issue since no one know what PROXY_SERVER is and how configure it etc..

lemon-suse:
> For rmt testing, we test the rmt dev version. So in the test setting, need to add DEV_PATH setting and then run a verification case.

Yes, the new verification log with DEV_PATH : https://openqa.nue.suse.com/tests/7821564#details

jouyingbin:
Would you please also provide the VR of rmt_proxy_server?

lemon-suse:
> Would you please also provide the VR of rmt_proxy_server?

Yes, I just update the VR for both client and server in PR description. Thanks.

tests/console/rmt/rmt_proxy.pm
```diff
@@ -0,0 +1,71 @@
...
+    # backup the rmt conf file
+    assert_script_run("cp $rmt_conf $rmt_conf_bk");
+
+    # update the rmt conf file with the proxy setting
+    assert_script_run("sed -i 's/proxy: /proxy: $proxy_url/g' $rmt_conf");
+    assert_script_run("sed -i 's/proxy_auth: /proxy_auth: ntlm/g' $rmt_conf");
+    assert_script_run("systemctl restart rmt-server");
+
+    # verify the basic function of the rmt server
+    rmt_sync;
```
coolgw, line 58: Base https://openqa.nue.suse.com/tests/7849886#step/rmt_proxy/40, i suppose rmt-cli sync command already can verify the connection?
Use assert_script instead of rmt_sync to execute the command.

tests/console/rmt/rmt_proxy.pm
lemon-suse: I haven't found method to use enable to verify but avoid the hardcode, and the verification through rmt_cli enable is required from Chenzi.

tests/console/rmt/rmt_proxy.pm
lemon-suse: Thanks, I updated code to use assert_script_run instead.

Verification run:  https://openqa.nue.suse.com/tests/7850193#step/rmt_proxy/40  proxy client
                         https://openqa.nue.suse.com/tests/7850192#details

tests/console/rmt/rmt_proxy.pm
coolgw: I suspect this ntlm not used at all, seems tinyproxy do not support ntlm auth.

tests/console/rmt/rmt_proxy.pm
coolgw: If we still need check product enable operation , i suggest we can get the product name by output of rmt-cli instead of hardcode.

tests/console/rmt/rmt_proxy.pm
lemon-suse: Updated code to support multi machine test.

tests/console/rmt/rmt_proxy.pm
lemon-suse: I need to investigate how to setup the NTLM server, so hang the PR for now.

rfan1: @lemon-suse 
Please help add my request into your considerration:)

[SUSE-JIRA] (SLE-22181) QA: Support NTLM authentication in the Unified Installer

coolgw (approved):

tests/console/rmt/rmt_proxy.pm
lemon-suse: Sorry, since this PR only set on SUT while not the proxy server, so I think it won't help you for the Jira feature.

PR #13848, opened 2021-12-10
Add more xfstest variables for publiccloud

grisu48:
Add the `XFS_MKFS_OPTIONS` and `XFS_TESTS_REFLINK` variables, which allows to configure custom
mkfs arguments for the xfstests. This is required to limit the reflink options for some tests to an isolated test case, instead of applying this options globally.

- Related ticket: https://progress.opensuse.org/issues/60179
- Verification run: http://duck-norris.qam.suse.de/tests/7940

### DIFF
--- a/tests/publiccloud/xfsprepare.pm
+++ b/tests/publiccloud/xfsprepare.pm
@@ -39,8 +39,11 @@ sub install_xfstests {
     ensure_user_exists("daemon", 2);
     # Create test users (See https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git/tree/README)
     assert_script_run("useradd -mU fsgqa");    # Create home directory (-m) and 'fsgqa' group for the user (-U) as well
-    script_run("useradd 123456-fsgqa");    # script_run because only required by few tests and there is a chance that users starting with digits won't work
     assert_script_run("useradd fsgqa2");
+
+    # The following is only required by few tests and there is a chance that users starting with digits won't work
+    my $fsgqa_123456 = script_run("useradd 123456-fsgqa") == 0;
+    record_info("123456-fsgqa", "error creating 123456-fsgqa user.\nSome tests will not be able to run", result => 'softfail') unless ($fsgqa_123456);
 }
 
 # Format the additional disk and mount it
@@ -50,11 +53,11 @@ sub partition_disk {
     assert_script_run("parted $device --script -- mklabel gpt");
     assert_script_run("parted -s -a min $device mkpart primary 1MB 50%");
     assert_script_run("parted -s -a min $device mkpart primary 50% 100%");
+    # Note: Each test run creates a new xfs filesystem and mounts it to the given mount point (See create_config),
+    # so we don't need to mount the new devices here. We create a new filesystem to ensure, that this is safe to do
     assert_script_run("mkfs.xfs -L xfstests ${device}1");
     assert_script_run("mkfs.xfs -L scratch ${device}2");
     assert_script_run("mkdir -p $mnt_xfs $mnt_scratch");
-    assert_script_run("mount ${device}1 $mnt_xfs");
-    assert_script_run("mount ${device}2 $mnt_scratch");
 }
 
 # Create configuration files required for the xfstests suite
@@ -69,8 +72,11 @@ sub create_config {
     assert_script_run("echo 'export SCRATCH_MNT=$mnt_scratch' >> $CONFIG_FILE");
     assert_script_run("echo 'export TEST_DEV=${device}1' >> $CONFIG_FILE");
     assert_script_run("echo 'export SCRATCH_DEV=${device}2' >> $CONFIG_FILE");
-    # Ensure reflink is enabled (required for several tests)
-    assert_script_run("echo 'MKFS_OPTIONS=\"-m reflink=1\"' >> $CONFIG_FILE");
+    # Add optional mkfs options
+    my $mkfs_options = get_var('XFS_MKFS_OPTIONS', '');
+    $mkfs_options .= " -m reflink=1" if (get_var('XFS_TESTS_REFLINK', 0) == 1);
+    $mkfs_options =~ s/^\s+|\s+$//g;    # trim string (left and right)
+    assert_script_run("echo 'MKFS_OPTIONS=\"$mkfs_options\"' >> $CONFIG_FILE") unless ($mkfs_options eq '');
 }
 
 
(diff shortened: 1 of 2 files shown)


tests/publiccloud/xfsprepare.pm
```diff
@@ -39,7 +39,9 @@ sub install_xfstests {
     ensure_user_exists("daemon", 2);
     # Create test users (See https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git/tree/README)
     assert_script_run("useradd -mU fsgqa");    # Create home directory (-m) and 'fsgqa' group for the user (-U) as well
-    script_run("useradd 123456-fsgqa");    # script_run because only required by few tests and there is a chance that users starting with digits won't work
+
+    # The following is only required by few tests and there is a chance that users starting with digits won't work
+    record_info("123456-fsgqa", "error creating 123456-fsgqa user.\nSome tests will not be able to run") if (script_run("useradd 123456-fsgqa") != 0);
```
pdostal, line 44: I somehow don't like this format where you actually do (create?) something important. I'd do something like:
```suggestion
    script_run("useradd 123456-fsgqa");
    record_info("123456-fsgqa", "error creating 123456-fsgqa user.\nSome tests will not be able to run") if (script_run("cat /etc/passwd | grep 123456-fsgqa") != 0);
```

pdostal (approved):
I like how you commented those removed lines.

I've just one small suggestion but no hard feelings.

tests/publiccloud/xfsprepare.pm
ilausuch: I agree but I think will be better to use assert_script_run in line 44

`assert_script_run("useradd 123456-fsgqa")`

tests/publiccloud/xfsprepare.pm
```diff
@@ -69,8 +71,10 @@ sub create_config {
     assert_script_run("echo 'export SCRATCH_MNT=$mnt_scratch' >> $CONFIG_FILE");
     assert_script_run("echo 'export TEST_DEV=${device}1' >> $CONFIG_FILE");
     assert_script_run("echo 'export SCRATCH_DEV=${device}2' >> $CONFIG_FILE");
-    # Ensure reflink is enabled (required for several tests)
-    assert_script_run("echo 'MKFS_OPTIONS=\"-m reflink=1\"' >> $CONFIG_FILE");
+    # Add optional mkfs options
+    my $mkfs_options = get_var('XFS_MKFS_OPTIONS', '');
+    $mkfs_options .= "-m reflink=1" if (get_var('XFS_TESTS_REFLINK', 0) == 1);
```
ilausuch, line 76: Maybe  with an extra space in case XFS_MKFS_OPTIONS cames without it?

`$mkfs_options .= " -m reflink=1"`

ilausuch (changes requested):

tests/publiccloud/xfsprepare.pm
grisu48: We can't assert this, as this fails on our SLES images. I still leave it there, as it is recommended to have this user, but one can see it as a note, that we are not able to run some tests due to this limitation.

tests/publiccloud/xfsprepare.pm
grisu48: Good catch! Updating it right away ....

tests/publiccloud/xfsprepare.pm
grisu48: Please check again, I've updated the corresponding lines.

ilausuch (approved):

ilausuch:
Good job

tests/publiccloud/xfsprepare.pm
ilausuch: understood, thanks for the explanation

tests/publiccloud/xfsprepare.pm
```diff
@@ -39,8 +39,11 @@ sub install_xfstests {
     ensure_user_exists("daemon", 2);
     # Create test users (See https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git/tree/README)
     assert_script_run("useradd -mU fsgqa");    # Create home directory (-m) and 'fsgqa' group for the user (-U) as well
-    script_run("useradd 123456-fsgqa");    # script_run because only required by few tests and there is a chance that users starting with digits won't work
     assert_script_run("useradd fsgqa2");
+
+    # The following is only required by few tests and there is a chance that users starting with digits won't work
+    my $fsgqa_123456 = script_run("useradd 123456-fsgqa") == 0;
+    record_info("123456-fsgqa", "error creating 123456-fsgqa user.\nSome tests will not be able to run") unless ($fsgqa_123456);
```
asmorodskyi, line 46:
```suggestion
    record_info("123456-fsgqa", "error creating 123456-fsgqa user.\nSome tests will not be able to run", result => 'softfail') unless ($fsgqa_123456);
```

asmorodskyi (approved):
LGTM

tests/publiccloud/xfsprepare.pm
grisu48: You're welcome :slightly_smiling_face:

tests/publiccloud/xfsprepare.pm
grisu48: I don't know, but since this will always fail and it is kinda known, I'm against putting a softfailure there. And yes, I know that this is the perfect situation FOR a softfailure.

The reason I'm against, is that there is nothing we gain from this softfailure. This issue will probably never being solved, and it is a known weekness of the xfstests, that is even present in the xfstests README.

WDYT?

asmorodskyi: First of all I already approved this PR , so this is not something on which I really insists so it is up to you to take it or ignore it. 
Nevertheless : 
1. AFAIK `reuslt => 'softfail'`  is not the same as calling `record_softfailure` . While the last one will gives you yellow bubble and yellow module the first will just draw yellow square . And I find this good trade off between just "normal" record_info and actually doing softfail with `record_softfailure`. 
2. Isn't soft failed state is exactly about such situation ?  When you have some problem which have no clear future ?
3. You stating that "some tests will not be able" can you show me how this actually look like ? Is there in further failures or tests are just silently skipped ? 
4. If it is so common that you don't want to draw extra attention to it maybe we can remove this record_info ?

grisu48: 1. That is nice, now I want to try it out :slightly_smiling_face: 
2. Yes but I also find it kinda .... disappointing :wink: 
3. It's actually only one that I could find: [generic/381](https://openqa.suse.de/tests/7811616#step/1_test_result/81) and yes, it's skipped.
4. Same as two: softfailures for something that is predictably never being solved is kinda unsettling. This is however only my personal taste and can differ from person to person

Gonna try your approach, if this does not alter the overall test result I find it a very good suggestion!

grisu48: http://duck-norris.qam.suse.de/tests/7941#step/xfsprepare/63 - I love this suggestion, this is amazing :+1: :+1: :+1:

tests/publiccloud/xfsprepare.pm
pdostal: 👍

PR #13854, opened 2021-12-13
Update the whole smart card package chain

rfan1:
All smart card related packages are updated to upstream
version, we need check all of them are up to date with
basic service check.

- Related ticket: https://progress.opensuse.org/issues/103751
- Needles: n/a
- Verification run: 
https://openqa.suse.de/tests/7819992            -SLE
https://openqa.opensuse.org/tests/2082709 -TW

I only tested on x8_64 platform, but I have confidence to pass it on all platforms

### DIFF
--- a/tests/security/smartcard/version_check.pm
+++ b/tests/security/smartcard/version_check.pm
@@ -1,0 +1,49 @@
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary: Update the whole smart card package chain in SLE-15-SP4
+# Maintainer: rfan1 <richard.fan@suse.com>
+# Tags: poo#103751, tc#1769856
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+use version_utils 'is_sle';
+use registration 'add_suseconnect_product';
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+    add_suseconnect_product('PackageHub', undef, undef, undef, 300, 1) if is_sle;
+
+    # Version check
+    my $pkg_list = {
+        'pam_pkcs11' => '0.6.10',
+        'pcsc-asekey' => '3.7',
+        'pcsc-ccid' => '1.4.36',
+        'pcsc-cyberjack' => '3.99.5final.SP14',
+        'pcsc-gempc' => '1.0.8',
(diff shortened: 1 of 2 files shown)


tests/security/smartcard/version_check.pm
```diff
@@ -0,0 +1,48 @@
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary:  Update the whole smart card package chain in SLE-15-SP4
```
lilyeyes, line 4: s/Summary:  U/Summary: U/

lilyeyes (approved):

lilyeyes (approved):
LGTM.

tests/security/smartcard/version_check.pm
```diff
@@ -0,0 +1,48 @@
...
+        opensc => '0.22.0',
+        'pcsc-lite' => '1.9.4',
+        'libp11-3' => '0.4.11',
+        'pcsc-tools' => '1.5.8',
+        'pkcs11-helper' => '1.25.1'
+    };
+    zypper_call("in " . join(' ', keys %$pkg_list));
+    foreach my $pkg (keys %$pkg_list) {
+        my $current_ver = script_output("rpm -q --qf '%{version}\n' $pkg");
+        if ($current_ver lt $pkg_list->{$pkg}) {
```
Amrysliu, line 37: If the package is not install, what's the expect result?

rfan1: Nice comment!

I have the same concern at 1st, however, I think zypper call can report error there, am I right?

jouyingbin: similar comments with Xiaojing's. If the package is not in the repo and unable to install, it will report the package not found, and will it also be feasible to compare the version check? I am not quite sure.

rfan1: Hmm, IMO, whenever the package in the list is not able to installed with any reason. test should fail there. [am I correct].
in that case, we can file a bug to track the issue.

jouyingbin: okay, I agree with you. I think zypper will report some message directly.

jouyingbin (approved):
LGTM

ShawnHa0 (approved):

ShawnHa0:
LGTM

PR #13858, opened 2021-12-13
sshd: to set mode 700 for .ssh directory

DimStar77:
the ~/.ssh directory should alwats be mode 700


Vogtinator (changes requested):
While `.ssh` should indeed be mode `0700`, `0755` is also fine.

Any umask which causes `.ssh` to be created with a mode which OpenSSH complains about is undesirable though, so IMO `mkdir .ssh` must work.

DimStar77:
Much more going on - there might be even a patch on pam responsible for the wrong behavior in first place. in fact, we just found a side effect that was probably never meant to happen like this. closing

PR #13864, opened 2021-12-14
Remove workaround for poo#88247

lemon-suse:
We don't need to record soft failure any more since we changed test suite to verify password with new added user 'test' while not the default user.

- Related ticket: https://progress.opensuse.org/issues/103821
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/7849900#step/install_service/126

### DIFF
--- a/lib/services/users.pm
+++ b/lib/services/users.pm
@@ -178,7 +178,6 @@ sub full_users_check {
         logout_and_login;
         # for poo#88247, it is hard to deal with the authorization of bernhard in
         # following migration process, we have to restore current user's password.
-        record_soft_failure("poo#88247, it is hard to deal with the authorization of bernhard in following migration process, we have to restore current users password");
         restore_passwd;
     }
     else {


lib/services/users.pm
```diff
@@ -178,7 +178,8 @@ sub full_users_check {
         logout_and_login;
         # for poo#88247, it is hard to deal with the authorization of bernhard in
         # following migration process, we have to restore current user's password.
-        record_soft_failure("poo#88247, it is hard to deal with the authorization of bernhard in following migration process, we have to restore current users password");
+	# We don't need to record softfail here any more since we changed test
```
tinawang123, line 181: As you have deleted soft failure at here, it is very strange for other to read this comment

lemon-suse: Ok, updated.

tinawang123 (approved):
LGTM

lib/services/users.pm
chcao: After removing the record_soft_failure of poo#88247, there is no info about poo#88247 any longer, so need to remove its' comments together?

lemon-suse: I think reserve this comments for poo#88247 to explain why need restore password here is reasonable:
   restore_passwd;

rfan1 (approved):
LGTM

PR #13866, opened 2021-12-14
Fix double interval while using wait_until

OleksandrOrlov:
The interval was doubled while using wait_until function. It happened
because http requests already wrapped with wait_until function. In case
wait_until is used to wait for some events happened to widgets, the
interval that is specified in this wait_until become summarized with the
interval, that is used in WidgetController to send http requests.

The commit makes wait until to be dependent on time, but not on counter.
So, if it is set to wait two seconds, it will check current time, and
wait for two seconds from it, instead of just using counter increment.

Another change that helped to avoid double timeout is removal of "sleep"
from http_get and http_post. That was hard coded to wait 1 second for
each HTTP request.

- Related ticket: https://progress.opensuse.org/issues/103548
- VRs: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=70.1_oorlov&groupid=96

### DIFF
--- a/lib/YuiRestClient/Http/WidgetController.pm
+++ b/lib/YuiRestClient/Http/WidgetController.pm
@@ -4,7 +4,6 @@ package YuiRestClient::Http::WidgetController;
 use strict;
 use warnings;
 
-use YuiRestClient::Logger;
 use YuiRestClient::Wait;
 use YuiRestClient::Http::HttpClient;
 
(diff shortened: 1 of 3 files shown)


lib/YuiRestClient/Http/WidgetController.pm
```diff
@@ -50,11 +49,17 @@ sub find {
         params => $args
     );
 
-    YuiRestClient::Logger->get_instance()->debug('Finding widget by url: ' . $uri);
+    YuiRestClient::Logger->get_instance()->debug("Finding widget by url: $uri");
```
ge0r, line 52: why is it that use YuiRestClient::Logger;` is not required any more?

OleksandrOrlov: I removed it by mistake. The question now, how it worked without the import:)

jknphy: I guess non of those `use`s are needed because there are in the same namespace, but it is more clear that way.

lib/YuiRestClient/Wait.pm
```diff
@@ -12,10 +12,12 @@ sub wait_until {
     $args{message} //= '';
 
     die "No object passed to the method" unless $args{object};
-
-    my $counter = abs(int($args{timeout} / $args{interval}));
     my $result;
-    while (--$counter >= 0) {
+    # if timeout is 0, just execute the passed expression once
+    return eval { $args{object}->() } if $args{timeout} == 0;
```
jknphy, line 17: do we need in our framework timeout=0, perhaps it is more obvious for the user to have an error which say "don't use timeout 0, instead use just the function as it is'. I don't know if you borrow that concept from openQA or other framework, might not make sense, and we will just have one `eval` in the function. wdyt?

jknphy:
Applied timeout checking real time in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14275

PR #13874, opened 2021-12-15
Remove product selection from schedule for s390x

OleksandrOrlov:
s390x started having only one product on Product Selection screen on
Full medium.

The commit removes select_SLES testing module from the schedules for
s390x.

- Verification run: https://openqa.suse.de/tests/7858935


ge0r (approved):
LGTM

PR #13876, opened 2021-12-16
Install SUSEConnect when system role is common criteria

Amrysliu:
In the last build [74.1], when system role is common criteria, the
package SUSEConnect is not installed. This caused all CC test cases
failed. So we need to install this package as a workaround and need to
confirm if it is by design.

Related: https://progress.opensuse.org/issues/104070

Verify run: https://openqa.suse.de/tests/7862983

### DIFF
--- a/tests/security/cc/cc_audit_test_setup.pm
+++ b/tests/security/cc/cc_audit_test_setup.pm
@@ -5,7 +5,7 @@
 #
 # Summary: Setup 'audit-test' test environment of a system with needed packages
 # Maintainer: llzhao <llzhao@suse.com>
-# Tags: poo#93441
+# Tags: poo#93441, poo#104070
 
 use base 'consoletest';
 use strict;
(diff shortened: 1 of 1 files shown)


jouyingbin (approved):
LGTM

tests/security/cc/cc_audit_test_setup.pm
```diff
@@ -23,6 +23,9 @@ sub run {
 
     select_console 'root-console';
 
+    zypper_call('in SUSEConnect');
+    record_soft_failure('bsc#1193782', 'SUSEConnect is not installed when system role is common criteria');
```
lilyeyes, line 27: I think if `which SUSEConnect2` failed we can do `record_soft_failure` in case if `bsc#1193782` is fix and test results still show unsuitable softfailure.

Amrysliu: good idea

lilyeyes (approved):
lgtm.

rfan1:
test case -> test cases

tests/security/cc/cc_audit_test_setup.pm
```diff
@@ -23,6 +23,10 @@ sub run {
 
     select_console 'root-console';
 
+    if (script_run('which SUSEConnect') != 0) {
```
rfan1, line 26: I would suggest use rpm -ql SUSEConnect rather than which here, but not a big deal

Amrysliu: I prefer 'which', seems like easier.

rfan1 (approved):
LGTM

tinawang123 (approved):
LGTM

PR #13879, opened 2021-12-16
xfstests: Group file change name in upstream

SeroSun:
In xfstests upstream, there is a group file rename from group to group.list. And group file create during build.[1] Then we update name to adapt this change.

[1] xfstests git tag: 17f6ac8eccb4d3f974aa74350b63e186c092ea16
    fstests: automatically generate group files
    
    Now that we've moved the group membership details into the test case
    files themselves, automatically generate the group files during build.
    The autogenerated files are named "group.list" instead of "group" to
    avoid conflicts between generated and (stale) SCM files as everyone
    rebases.

- Verification run: 
- Before this PR group tests not enabled only triggered 1 test: https://openqa.suse.de/tests/7806579
- After this PR group enabled a lot of tests: https://openqa.suse.de/tests/7863592

### DIFF
--- a/tests/xfstests/run.pm
+++ b/tests/xfstests/run.pm
@@ -184,7 +184,7 @@ sub exclude_grouplist {
     foreach my $group_name (@GROUPLIST) {
         next if ($group_name !~ /^\!/);
         $group_name = substr($group_name, 1);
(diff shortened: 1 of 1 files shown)


lansuse:
LGTM

PR #13885, opened 2021-12-16
installation/await_install: Bump timeout for powepc

foursixnine:
Installation is taking a bit longer nowdays, and while I'm not sure that it's due to network or some other issue, those extra 10 minutes should help, as most of the times... looks like await install is just shy of minutes before finishing the installation

https://openqa.suse.de/tests/7856741#step/await_install/69

VR: https://openqa.suse.de/t7867595

See also https://progress.opensuse.org/issues/104106

### DIFF
--- a/tests/installation/await_install.pm
+++ b/tests/installation/await_install.pm
@@ -105,7 +105,9 @@ sub _set_timeout {
 sub run {
     my $self = shift;
     # NET isos are slow to install
-    my $timeout = is_s390x ? 2400 : 2000;
+    # If this timeout needs to be bumped again, we might be having a bigger network problem
+    # or a peformance problem on the installer
+    my $timeout = (is_s390x || is_ppc64le) ? 2400 : 2000;
 
     # workaround for yast popups and
     # detect "Wrong Digest" error to end test earlier
@@ -130,7 +132,7 @@ sub run {
 
     my $mouse_x = 1;
     while (1) {
-        die 'timeout hit on during await_install' if $timeout <= 0;
+        die "timeout ($timeout) hit on during await_install" if $timeout <= 0;
         my $ret = check_screen \@tags, 30;
         $timeout -= 30;
         diag("left total await_install timeout: $timeout");


dzedro (approved):

okurz (changes requested):
s/powepc/PowerPC/ or ppc or something. The problem is https://progress.opensuse.org/issues/102882 which is really specific to one rack in a server room used by the OSD infrastructure. I suggest to use TIMEOUT_SCALE in the worker specific config, not change the test code

punkioudi (approved):
lgtm!

foursixnine:
> s/powepc/PowerPC/ or ppc or something. The problem is https://progress.opensuse.org/issues/102882 which is really specific to one rack in a server room used by the OSD infrastructure. I suggest to use TIMEOUT_SCALE in the worker specific config, not change the test code

I will still move forward, as the test normally takes 12m, and the extra timeout should cover for future problems in the network even after https://gitlab.suse.de/openqa/salt-pillars-openqa/-/merge_requests/375, or for the time it is reverted.

okurz:
You could have at least fixed the typo :,( I don't think it's a good style to merge despite someone requesting changes.

foursixnine:
> You could have at least fixed the typo :
Which typo? :) I hope you're seeing it now and not during your first review

okurz:
The typo I mentioned in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13885#pullrequestreview-834975929

> s/powepc/PowerPC/ or ppc or something.

Seems like you overlooked my first message. That's ok, mistakes happen. But even more so this is why I suggest to not merge despite changes requested.

foursixnine:
> The typo I mentioned in [#13885 (review)](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13885#pullrequestreview-834975929)
> 
> > s/powepc/PowerPC/ or ppc or something.
> 
> Seems like you overlooked my first message. That's ok, mistakes happen. But even more so this is why I suggest to not merge despite changes requested.

Ah, wasn't really clear what you meant, the second part of that comment diluted the possible meaning of it.

wrt merging despite changes requested, let's say that there's a calculated risk (if any) to simply waiting for a bit longer, instead of just waiting for the workaround.

git history will show how that there's a reason for that in the future, and that would be a better early warning in the future (including to check for typos before merging :).

okurz:
sure, not the end of the world :)

PR #13886, opened 2021-12-16
Merge podman and docker image based tests into one module

b10n1k:
Leverage `TestArgs` to pass an attribute which defines which container engine should be initialized for each case.
Because each engine has separate function which load the module, each `TestArgs` is also separate instance which means
that the modules works also when they both scheduled in the same job.

- Related ticket: https://progress.opensuse.org/issues/101210
- Verification run:

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -41,11 +41,15 @@ sub is_ubuntu_host {
 }
 
 sub load_image_tests_podman {
-    loadtest 'containers/podman_image';
+    my $engine_args = OpenQA::Test::RunArgs->new();
+    $engine_args->{podman} = 1;
+    loadtest 'containers/image_test', name => 'podman_images', run_args => $engine_args;
 }
 
 sub load_image_tests_docker {
-    loadtest 'containers/docker_image';
+    my $engine_args = OpenQA::Test::RunArgs->new();
+    $engine_args->{docker} = 1;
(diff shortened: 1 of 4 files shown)


lib/main_containers.pm
```diff
@@ -55,9 +59,11 @@ sub load_image_tests_docker {
 
 sub load_host_tests_podman {
     if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+")) {
+        my $engine_args = OpenQA::Test::RunArgs->new();
+        $engine_args->{podman} = 1;
```
jlausuch, line 63: Looks like a good solution, but why you give it a value `1`? 
Could we do something like 
`engine_args->{runtime} = 'docker'` ?

tests/containers/image_test.pm
```diff
@@ -0,0 +1,44 @@
...
+use containers::urls 'get_suse_container_urls';
+use testapi;
+use utils;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal();
+    my $engine = $self->containers_factory($self->{run_args});
+
+    scc_apply_docker_image_credentials() if (get_var('SCC_DOCKER_IMAGE'));
```
jlausuch, line 24: My memory fails, can you remind me what is `SCC_DOCKER_IMAGE` for?

jlausuch: And, shouldn't we do this only for docker test?

PR #13895, opened 2021-12-20
Adjust the openssl hash check return message

jouyingbin:
Description:
Currently openssl just check disabled for fips|disabled for FIPS message, it leads to miss check the error which mentioned in bsc#1193859, thus tying to adjust the openssl hash check return message to "`$hash is not a known digest`" to avoid the defect.

- Related ticket: https://progress.opensuse.org/issues/104184
- Needles: NA
- Verification run: 
  (SP4) https://openqa.suse.de/tests/7893275#step/openssl_fips_hash/17
  (SP3) https://openqa.suse.de/tests/7893278#step/openssl_fips_hash/35
  (Currently there is a bug to track https://bugzilla.suse.com/show_bug.cgi?id=1193859 the error)

### DIFF
--- a/tests/fips/openssl/openssl_fips_hash.pm
+++ b/tests/fips/openssl/openssl_fips_hash.pm
@@ -1,6 +1,6 @@
 # openssl fips test
 #
-# Copyright 2016-2020 SUSE LLC
+# Copyright 2016-2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 #
 # Test description: In fips mode, openssl only works with the FIPS
@@ -13,7 +13,7 @@
 #          the cases to verify opessl hash, cipher, or public key algorithms
 #
 # Maintainer: Ben Chou <bchou@suse.com>
-# Tags: poo#44834, poo#64649, poo#64842
+# Tags: poo#44834, poo#64649, poo#64842, poo#104184
 
 use base "consoletest";
 use testapi;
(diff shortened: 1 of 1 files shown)


rfan1:
Can you please mention the poo number in test module file?

tests/fips/openssl/openssl_fips_hash.pm
```diff
@@ -40,10 +40,14 @@ sub run {
     # Remove md2 and sha, and add rmd160 and md5-sha1 from invalid hash check in fips mode
     my @invalid_hash = ("md4", "md5", "mdc2", "rmd160", "ripemd160", "whirlpool", "md5-sha1");
     for my $hash (@invalid_hash) {
-        validate_script_output "openssl dgst -$hash $tmp_file 2>&1 || true", sub { m/disabled for fips|disabled for FIPS|unknown option|Unknown digest|dgst: Unrecognized flag/ };
+        validate_script_output "openssl dgst -$hash $tmp_file 2>&1 || true", sub { m/$hash is not a known digest|unknown option|Unknown digest|dgst: Unrecognized flag/ };
```
rfan1, line 43: Ben
One quick quesion, why "d/Disabled for fips" is removed

jouyingbin: @rfan1 Because if we just check **disabled for fips|disabled for FIPS|** , it will miss check the bug https://bugzilla.suse.com/show_bug.cgi?id=1193859. It will still bypass the error
`routines:EVP_DigestInit_ex:disabled for FIPS:crypto/evp/digest.c:135:`
Confirm with Marcus, we would better to check the message as i.g. "md5 is not a known digest" instead of the a bit ugly message.

lilyeyes (approved):
lgtm.

rfan1 (approved):

tests/fips/openssl/openssl_fips_hash.pm
dzedro: If I understand Marcus correctly it's ugly but intended. https://bugzilla.suse.com/show_bug.cgi?id=1193859#c1

jouyingbin: @dzedro Yes, and Marcus mentioned that "perhaps the message can be adjusted to be similar to the one in sp2/sp3.", is that right? 
or do you think we need to add them back? Actually, I think it can be fixed in SP4.

dzedro: @jouyingbin Right now, it's failing everywhere :roll_eyes: 
I have seen the poo# and that it was discussed with Marcus,
but I don't understand how should the correct behavior look like.
I would love to have the previous regex, as there is bug for the issue.

dzedro: @jouyingbin Guess I now understand what behavior is expected, please review https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13935

PR #13899, opened 2021-12-20
thunderbird: New GUI beviour

dzedro:
Failures are happening due to https://bugzilla.suse.com/show_bug.cgi?id=1191916

- Related ticket: https://progress.opensuse.org/issues/101042
- Verification run:
https://openqa.suse.de/tests/7950480 15 SP3
https://openqa.suse.de/tests/7950499 15 SP3
https://openqa.suse.de/tests/7950506 15 SP4

### DIFF
--- a/lib/thunderbird_common.pm
+++ b/lib/thunderbird_common.pm
@@ -207,14 +207,21 @@ sub tb_send_message {
     wait_still_screen(2, 4);
 
     if ($hostname eq 'client') {
-        assert_and_click "thunderbird_attachment_reminder" if check_screen('thunderbird_attachment_reminder', 2);
-        assert_and_click "thunderbird_SSL_error_security_exception";
-        assert_and_click "thunderbird_confirm_security_exception";
-        assert_and_click 'thunderbird_maximized_send-message';
-        assert_and_click "thunderbird_get-messages" unless check_var('SLE_PRODUCT', 'sled');
+        while (1) {
+            my @tags = qw(thunderbird_attachment_reminder thunderbird_SSL_error_security_exception thunderbird_confirm_security_exception thunderbird_maximized_send-message thunderbird_cancel thunderbird_get-messages);
+            wait_still_screen(5, 10);
+            assert_screen(\@tags);
+            click_lastmatch;
+            last if match_has_tag('thunderbird_get-messages');
+        }
     }
     else {
-        assert_screen 'thunderbird_sent-folder-appeared', 90;
+        while (1) {
+            wait_still_screen(5, 10);
+            assert_screen [qw(thunderbird_sent-folder-appeared thunderbird_cancel)];
+            click_lastmatch if match_has_tag('thunderbird_cancel');
+            last if match_has_tag('thunderbird_sent-folder-appeared');
+        }
     }
 
     return $mail_subject;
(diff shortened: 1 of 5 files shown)


lib/thunderbird_common.pm
```diff
@@ -42,61 +42,154 @@ sub tb_setup_account {
...
-        wait_still_screen(3);
-        assert_and_click 'thunderbird_SSL_done_config' unless check_screen('thunderbird_confirm_security_exception');
-        assert_and_click "thunderbird_confirm_security_exception";
-        assert_and_click "thunderbird_skip-system-integration";
-        assert_and_click "thunderbird_get-messages";
+        # If use multimachine, select correct needles to configure thunderbird.
+        if ($hostname eq 'client') {
+            send_key 'end';    # go to the bottom to see whole manual configuration
+            if (check_screen 'thunderbird_in-hostname-start-with-dot') {
+                record_info 'bsc#1191866';
```
b10n1k, line 85: why not a softfail here?

dzedro: Did you see the bug ?!

lib/thunderbird_common.pm
```diff
@@ -29,7 +29,7 @@ C<$proto> can be C<pop> or C<imap>.
 C<$account> can be C<internal_account_A> or C<internal_account_B> or C<internal_account_C> or C<internal_account_D>.
 =cut
 sub tb_setup_account {
-    my $hostname = get_var('HOSTNAME');
+    my $hostname = get_var('HOSTNAME') // '';
```
b10n1k, line 32: nitpicking: it would be better with `//=` as it checks if `get_var('HOSTNAME')` is also defined and not only true
either `my $hostname = get_var('HOSTNAME') // undef;` or `my $hostname = get_var('HOSTNAME') //= '';` wdyt?

lib/thunderbird_common.pm
b10n1k: i did now. So it s invalid. there is not clear definition of usage form MOV but i guess it is fine in this case.

lib/thunderbird_common.pm
dzedro: You don't understand the issue `Fixes 'Use of uninitialized value $hostname in string eq at ...'`
The issue is that later in code is `$hostname` used and because it's `undef` it returns the error.

lib/thunderbird_common.pm
dzedro: It's common sense. There is no reason to spam bug because of soft fail.
I could just not add it at all, but it;s better than any `# comment` to understand why it is done.

lib/thunderbird_common.pm
b10n1k: what i understand from this is that `my $hostname = get_var('HOSTNAME') // undef;` is not suitable here. that s fine. And i know that it doesnt make big difference. i just think it is better with _assingment-or_ because is equal in writing with `my $hostname = defined get_var('HOSTNAME') ? get_var('HOSTNAME') : '';`

b10n1k: feel free to resolve it if you dont want to this. it was just a suggestion.

dzedro: You just complicate things, I just added empty value to the variable if it does not exist.
There are variants of this tests, thus the variable does not need to be always defined.
Maybe it could be done differently and better. This simply just does fix the issue.

dzedro:
The https://bugzilla.suse.com/show_bug.cgi?id=1191916 was tricky, but looks like it works well ... soft fail is on needle.

PR #13900, opened 2021-12-21
Increase the timeout for running trustedprograms tests

Amrysliu:
Sometimes the CC test 'trustedprograms' will fail in timeout due to the
aarch64 workers' performance in OSD. So we need to increate the waiting
time to avoid this issue.

Related: https://progress.opensuse.org/issues/103739
Verify run: https://openqa.suse.de/tests/7886357#

### DIFF
--- a/tests/security/cc/trustedprograms.pm
+++ b/tests/security/cc/trustedprograms.pm
@@ -42,7 +42,7 @@ sub run {
 
     select_console 'root-console';
 
-    run_testcase('trustedprograms', (make => 1, timeout => 900));
+    run_testcase('trustedprograms', (make => 1, timeout => 1200));
 
     # Compare current test results with baseline
     my $result = compare_run_log('trustedprograms');


lilyeyes:
I did not found the timeout issue in openQA links offered in poo ticket.
Anyway you fix works and lgtm.

lilyeyes (approved):
lgtm.

tests/security/cc/trustedprograms.pm
```diff
@@ -42,7 +42,7 @@ sub run {
 
     select_console 'root-console';
 
-    run_testcase('trustedprograms', (make => 1, timeout => 900));
+    run_testcase('trustedprograms', (make => 1, timeout => 960));
```
rfan1, line 45: @Amrysliu 
I would suggest add more timeout e.g. 1200 for aarch64 platform, and set timeout=900 for others

Amrysliu:
> I did not find the timeout issue in openQA links offered in poo ticket. Anyway you fix works and lgtm.

Here is an timeout failure: https://openqa.suse.de/tests/7882864#step/trustedprograms/38
It only happened in last build. Maybe because it failed in others issue before.

tests/security/cc/trustedprograms.pm
Amrysliu: According to https://openqa.suse.de/tests/7882864#step/trustedprograms/38, the command almost completely run. So I think increasing some seconds should work. Different timeout in different platforms make it complex. So I prefer to keep that simply.

rfan1: I agree with you on this part, however, it is an known issue that we have lower performance on arm workers, increase only 1m can't make sure we can cover "more", I mean we should use more buffer here :)

rfan1 (approved):
Thanks, LGTM now

tinawang123 (approved):
LGTM

jouyingbin (commented):
LGTM

PR #13903, opened 2021-12-21
Guests install mirror

tbaev:
- Related ticket: https://progress.opensuse.org/issues/102596
- Needles: N/A
- Verification run: [SLE-SP2-XEN](http://openqa.qam.suse.cz/tests/34333) [SLE-SP1-XEN](http://openqa.qam.suse.cz/tests/34344) [SLE-15-KVM](http://openqa.qam.suse.cz/tests/34355) [SLE-12-SP3](http://openqa.qam.suse.cz/tests/34365)

### DIFF
--- a/lib/virt_autotest/common.pm
+++ b/lib/virt_autotest/common.pm
@@ -32,7 +32,7 @@ if (get_var("REGRESSION", '') =~ /xen/) {
             macaddress => '52:54:00:78:73:a1',
             ip => '192.168.122.106',
             distro => 'SLE_12_SP3',
-            location => 'http://mirror.suse.cz/install/SLP/SLE-12-SP3-Server-GM/x86_64/DVD1/',
+            location => 'http://schnell.suse.de/SLE12/SLE-12-SP3-Server-GM/x86_64/DVD1/',
             linuxrc => 'ifcfg="eth0=192.168.122.106/24,192.168.122.1,192.168.122.1"',
         },
         sles12sp3HVM => {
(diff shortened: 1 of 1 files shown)


lib/virt_autotest/common.pm
```diff
@@ -32,7 +32,7 @@ if (get_var("REGRESSION", '') =~ /xen/) {
             macaddress => '52:54:00:78:73:a1',
             ip => '192.168.122.106',
             distro => 'SLE_12_SP3',
-            location => 'http://mirror.suse.cz/install/SLP/SLE-12-SP3-Server-GM/x86_64/DVD1/',
+            location => 'http://schnell.suse.de/SLE12/SLE-12-SP3-Server-GM/x86_64/DVD1/',
```
grisu48, line 35: I'd be very careful. schnell.suse.de is in Nüremberg, so the whole network traffic needs to go over the NUE<->PRG VPN. mirror.suse.cz is in Prague and there was a reason why we sticked to that machine @pdostal

lib/virt_autotest/common.pm
```diff
@@ -52,7 +52,7 @@ if (get_var("REGRESSION", '') =~ /xen/) {
             macaddress => '52:54:00:78:73:a3',
             ip => '192.168.122.102',
             distro => 'SLE_15',
-            location => 'http://mirror.suse.cz/install/SLP/SLE-15-Installer-LATEST/x86_64/DVD1/',
+            location => 'http://dist.nue.suse.com/install/SLP/SLE-15-Installer-LATEST/x86_64/DVD1/',
```
grisu48, line 55: Don't know for sure where dist.nue.suse.com is located, but it sounds very much also like NUE.

grisu48 (commented):
I have concerns.

lib/virt_autotest/common.pm
tbaev: We are using it only to start the installation, after that the autoyast profile will use the repos form mirror.suse.cz. 
Sending from NUE<->RPG a few hundred megabytes, don't think it will cause issue / bottleneck?

grisu48: If it isn't an issue, why switch it then?

alice-suse: Regarding Felix's concern, maybe we can get some conclusion from Teodor's existing validation results? Teodor, how many jobs did you validate for your PR? It seems all are run on openqa.qam.suse.cz. Do you ever meet problems that Felix worries? If jobs number is not enough, maybe we can try more before merging the PR?

tbaev: I have run 4 jobs, but we have agreed with Felix to discuss it after he is back from vacation.

tbaev: After discussing with Felix we agree that this is not a good solution. I will close the PR

PR #13905, opened 2021-12-22
Quit packagekit before deregister system

lemon-suse:
We need quit packagekit before deregister system to make sure it won't block to de-register system via SUSEConnect.
 
- Related ticket: https://progress.opensuse.org/issues/104238
- Needles: N/A
- Verification run: http://openqa.nue.suse.com/tests/7893303#step/scc_deregistration/11  SLES15SP2


tinawang123 (approved):
LGTM

PR #13908, opened 2021-12-22
WIP: Simplify ImageMagick test

BillAnastasiadis:
This pr is a suggestion and aims to simplify the imagemagick tests and make them less prone to failure in the process by eliminating needles.

- Related ticket: https://progress.opensuse.org/issues/102359
- Verification run: http://10.161.229.209/tests/442 (example)


punkioudi:
Bill, could you explain a bit more in detail the solution that you are proposing here? In my understanding, it doesn't have the same coverage as before, right?

BillAnastasiadis:
> Bill, could you explain a bit more in detail the solution that you are proposing here? In my understanding, it doesn't have the same coverage as before, right?

No, according to the ticket and subsequent discussions in the dailys, the idea I got is that the test should be reduced to some transformations on images and practical testing. The proposal has to do with reducing needles by comparing images with direct msre comparissons, and I have incorporated basic functions like crop, resize, reduce quality, logo creation, alpha with letter creation, blurring images, image identification, png-gif-jpg comparisson etc.

The list of comparissons is open, if we decided we could even create the same amount of images as previously but with needle-less msre comparisons. But I think that the IM tests creating to many images to test was one of the problems. That's why I wait for suggestions, otherwise I will spend the next couple of days adding more comparissons to the test (but still considerably less than the og test).

tests/x11/ImageMagick.pm
```diff
@@ -26,86 +26,55 @@ use testapi;
...
+    # Prepare to run tests
+    assert_script_run "wget --quiet " . data_url('imagemagick/im_files.tar.gz') . " -O im_files.tar.gz";
+    assert_script_run "tar -xf im_files.tar.gz";
+    assert_script_run "cd im_files";
+    # Test image identification
+    my $idtf = script_output "identify test.png";
+    die "Identify failed" unless index($idtf, "test.png PNG 300x388 300x388+0+0 8-bit sRGB") != -1;
+    # Create image transformations to compare with pretransformed test images
+    my @test_imgs = (
+        "convert -fill Snow -background Green3 -strokewidth 2 -stroke Green4 -font Roboto -pointsize 256 -density 90 -size 1800x350 label:susetestlogo testlogo_t.png",
```
dzedro, line 54: Sounds like good idea, but what is all this image transformation about ?
I would expect just comparing of images.

BillAnastasiadis: Thanks for checking this out Josef, the transformations are crops, resizings, etc of some base images, which then get compared to pre-transformed images with the same transformations, to evaluate weather the transformations work as expected.

dzedro: in `imagemagick/bg_script.sh` there is done resize and specific transformation for "each" image.
But here is same transformation done for all images ?
It does not look to me like same test of convert and comparing result with reference.

tests/x11/ImageMagick.pm
```diff
@@ -26,86 +26,55 @@ use testapi;
 use utils;
 
 
+sub cmpr {
+    my ($og, $cp, $ttl) = @_;
+    my $cmp = script_output("compare -metric rmse $og $cp nullim 2>&1", proceed_on_failure => 1);
+    my ($mse) = $cmp =~ m/(?<=\().*(?=[^(]*\))/g;
```
foursixnine, line 32: maybe you  could use  -format "%[distortion]\n" and change the output to info, to avoid the regex or check the -format option to see if there's a good way to only show the normalized value (the one inside the parentheses)

tests/x11/ImageMagick.pm
```diff
@@ -26,86 +26,55 @@ use testapi;
 use utils;
 
 
+sub cmpr {
+    my ($og, $cp, $ttl) = @_;
+    my $cmp = script_output("compare -metric rmse $og $cp nullim 2>&1", proceed_on_failure => 1);
```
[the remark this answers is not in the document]
BillAnastasiadis: This was proposed by Josef, for reasons I haven't fully investigated (probably exit code of compare together with `script_output`s behaviour) the test fails without it...

tests/x11/ImageMagick.pm
BillAnastasiadis: It doesn't need to transform the images, because they are created with the same exact commands and from the same base images as the reference images so they are guaranteed to be the same (at least if everything is working fine).

tests/x11/ImageMagick.pm
```diff
@@ -26,86 +26,55 @@ use testapi;
 use utils;
 
 
+sub cmpr {
```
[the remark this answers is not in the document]
BillAnastasiadis: The vowels have returned to their rightful place!

tests/x11/ImageMagick.pm
BillAnastasiadis: Thank you for the advice! `%[distortion]` worked like a charm.

dzedro:
Please update VR, it's 26 day old with hidden fail ? http://10.161.229.209/tests/401#step/ImageMagick/46

dzedro:
I would not call it 'simplify' but 'degradation'.
Now it is testing more than 200 imagemagick features.
I\m pretty sure it's more precise than with this simplifying  change.
With this change it will test less in few squashed commands ?
This is like simplifying rolex to fake rolex.
If I would be Romanos I would be pissed.
The test does need maintenance or improvement for sure,
but I don't like this way, at least describe and name it properly.
That sayd I'm out, don't care what will happen with this PR! :v:

BillAnastasiadis:
> I would not call it 'simplify' but 'degradation'. Now it is testing more than 200 imagemagick features. I\m pretty sure it's more precise than with this simplifying change. With this change it will test less in few squashed commands ? This is like simplifying rolex to fake rolex. If I would be Romanos I would be pissed. The test does need maintenance or improvement for sure, but I don't like this way, at least describe and name it properly. That sayd I'm out, don't care what will happen with this PR! v

Josef, the idea from the very begining was to simplify the test, and we had discussed it on our daily where there was the suggestion of running only a few select transformations, and get rid of the countless needles.

That said, I'm not sure what your suggestion is.

If we want to test ALL transformations in the previous test, just without needles, it is entirely possible.
If we want to test only a few transformations, with needles, it is also entirely possible.
If you suggest to keep the entire test as it is and see no problem with it, you should write on the ticket and suggest that we close it. It's not my ticket or proposal, and it's not my call to shoot it down.

Feel free to bring it up on the next daily. If you disagreed with the ticket in the first place, you should have brought it up and talked with Santiago when he proposed the changes before I waste my time on it.

foursixnine:
> I would not call it 'simplify' but 'degradation'. Now it is testing more than 200 imagemagick features. I\m pretty sure it's more precise than with this simplifying change. With this change it will test less in few squashed commands ? This is like simplifying rolex to fake rolex. If I would be Romanos I would be pissed. The test does need maintenance or improvement for sure, but I don't like this way, at least describe and name it properly. That sayd I'm out, don't care what will happen with this PR! v

@dzedro About the loss of coverage, you could have expressed that concern more clearly on your previous to last review~ when the PR was reviewed. That said, I can understand the frustration that this might cause.

In general, a simplification means that there will be cut corners, initial intention was and still is to reduce the amount of needles needed for this test every now and then. If you can suggest a better way, I'd love to hear/read.

> The test does need maintenance or improvement for sure, but I don't like this way, at least describe and name it properly.

Can you suggest a common ground where the ammount of needles is reduced, but also the test remains with the same coverage as is there atm?.

@BillAnastasiadis one thing I'm taking from this, and looking at the test code, I wonder how much of a bad idea would be to show source+generated image (with no assert_screen) just to be able to see the same thing the test is doing.

dzedro:
> > I would not call it 'simplify' but 'degradation'. Now it is testing more than 200 imagemagick features. I\m pretty sure it's more precise than with this simplifying change. With this change it will test less in few squashed commands ? This is like simplifying rolex to fake rolex. If I would be Romanos I would be pissed. The test does need maintenance or improvement for sure, but I don't like this way, at least describe and name it properly. That sayd I'm out, don't care what will happen with this PR! v
> 
> @dzedro About the loss of coverage, you could have expressed that concern more clearly on your previous to last review~ when the PR was reviewed. That said, I can understand the frustration that this might cause.
> 
> In general, a simplification means that there will be cut corners, initial intention was and still is to reduce the amount of needles needed for this test every now and then. If you can suggest a better way, I'd love to hear/read.
> 
> > The test does need maintenance or improvement for sure, but I don't like this way, at least describe and name it properly.
> 
> Can you suggest a common ground where the ammount of needles is reduced, but also the test remains with the same coverage as is there atm?.
> 
> @BillAnastasiadis one thing I'm taking from this, and looking at the test code, I wonder how much of a bad idea would be to show source+generated image (with no assert_screen) just to be able to see the same thing the test is doing.

I was not sure so I asked. Now I just wrote how I see it.

> in imagemagick/bg_script.sh there is done resize and specific transformation for "each" image.
But here is same transformation done for all images ?
It does not look to me like same test of convert and comparing result with reference.

dzedro:
I'm not frustrated :) I don't mind, but at least name the change properly.

dzedro:
> If we want to test ALL transformations in the previous test, just without needles, it is entirely possible.

That would be "simplify" or no needles ...
But now it's not at all.

PR #13911, opened 2021-12-22
ai_ml: Use test repos when tested on top of JeOS

ggardet:
to avoid mismatch between packages on JeOS image and packages from repos.
- Verification run: jeos_extra_tests_ai_ml@RPi4: https://openqa.opensuse.org/t2099715


okurz (approved):
You are not stating "Why" so I can't know if the change is right but it looks like it would not break anything else at least :)

ggardet:
> You are not stating "Why" so I can't know if the change is right but it looks like it would not break anything else at least :)

Fixed.

PR #13914, opened 2021-12-22
Public Cloud: soft failure for az --version bug

jlausuch:

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1192671
- VR: https://openqa.suse.de/tests/7895652#

### DIFF
--- a/tests/publiccloud/azure_cli.pm
+++ b/tests/publiccloud/azure_cli.pm
@@ -27,7 +27,8 @@ sub run {
         add_suseconnect_product 'sle-module-public-cloud';
         zypper_call('in azure-cli jq python3-susepubliccloudinfo');
     }
-    assert_script_run('az --version');
+    record_soft_failure('bsc#1192671') if (script_run('az --version', die => 0) != 0);
+    record_soft_failure('bsc#1192671') if (script_run('az --help', die => 0) != 0);
 
     set_var 'PUBLIC_CLOUD_PROVIDER' => 'AZURE';
     my $provider = $self->provider_factory();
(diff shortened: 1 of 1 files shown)


tests/publiccloud/azure_cli.pm
```diff
@@ -27,7 +27,8 @@ sub run {
         add_suseconnect_product 'sle-module-public-cloud';
         zypper_call('in azure-cli jq python3-susepubliccloudinfo');
     }
-    assert_script_run('az --version');
+    record_soft_failure('bsc#1192671') if (script_run('az --version', die => 0) != 0);
+    record_soft_failure('bsc#1192671') if (script_run('az --help', die => 0) != 0);
```
jlausuch, line 31: in case you wonder why I'm adding `--help` is because it also returns !=0
https://bugzilla.suse.com/show_bug.cgi?id=1105223

asmorodskyi:
not saying that this version is wrong but please check how same problem solved in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/prepare_tools.pm#L90

jlausuch:
> not saying that this version is wrong but please check how same problem solved in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/prepare_tools.pm#L90

Since this is a cli specific test, which is supposed to be executed every day I think we should raise a failure/soft-failure if this simple command fails. So, I would think that record_soft_failure is preferred in this case.

jlausuch:
I'm closing the PR, I just learned that azure -v will return always a value !=0, so with this code I'm not really checking that the issue is the bug mentioned...

PR #13921, opened 2021-12-23
docs: Explain _until_screen_change return value

kalikiana:
This came up as a question in chat, and might be useful for others.

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -871,6 +871,9 @@ againg if the screen has not changed within C<$wait_change> seconds after
 the C<assert_and_click>. Returns the number of attempts made.
 C<$wait_change> defaults to 2 (seconds) and C<$repeat> defaults to 3.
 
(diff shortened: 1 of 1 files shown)


lib/utils.pm
```diff
@@ -871,6 +871,9 @@ againg if the screen has not changed within C<$wait_change> seconds after
 the C<assert_and_click>. Returns the number of attempts made.
 C<$wait_change> defaults to 2 (seconds) and C<$repeat> defaults to 3.
 
+You can check if the screen changed by using an explicit repeat and comparing it
+to the returned number of attempts. If the value is lower it means the screen didn't change.
```
foursixnine, line 875: Meaning the screen couldn't change during the first round? This is not what i understand from reading the code...

kalikiana: Yeah, it should of course *equal*.

foursixnine (approved):

PR #13927, opened 2021-12-27
Add snmp test

asdil12:
Ticket: https://progress.opensuse.org/issues/99633
http://kazhua.qa.suse.de/tests/52

### DIFF
--- a/tests/network/snmp.pm
+++ b/tests/network/snmp.pm
@@ -1,0 +1,70 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Package: net-snmp
+# Summary: Test smtp server and client tools.
+# Maintainer: QE Core <qe-core@suse.de>
+
+use base 'opensusebasetest';
+use warnings;
+use strict;
+use testapi;
+use utils;
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+
+    zypper_call 'in net-snmp';
+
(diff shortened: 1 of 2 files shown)


foursixnine:
@asdil12 could you add a test for the access control stuff? (`community`?) basically checking that certain certain clients cannot access a resource within the network? (say expect an access denied if the client request doesn't come from a different ip that the host has.

asdil12:
All I could think of adding would be:

```
$ snmpget -v2c -c badcommunity localhost sysLocation.0
Timeout: No Response from localhost.
```

and

```
$ snmpset -v3 -u bootstrap -l authPriv -a MD5 -x DES -A wrong_temp_password -X temp_password localhost sysLocation.0 s Earth
snmpset: Authentication failure (incorrect password, community or key) (Sub-id not found: (top) -> sysLocation)
```

asdil12:
I added the two checks that test using an invalid community string for auth (in version 2c) and an invalid password (in version 3).

tests/network/snmp.pm
```diff
@@ -0,0 +1,70 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Package: net-snmp
+# Summary: Test smtp server and client tools.
+# Maintainer: Dominik Heidler <dheidler@suse.de>
```
foursixnine, line 8: Shouldn't this be qe-core?

foursixnine (approved):
Aside from the Maintainer field, all looks good (And even that is not that big of a deal)

PR #13928, opened 2021-12-28
Enhance package version upgrade check logic

rfan1:
We have already had a good function to compare package version,
which is much better than simple string comparing. so enhance
our test code for the existing test modules

- Related ticket: https://progress.opensuse.org/issues/104355
- Needles: n/a
- Verification run: 
https://openqa.suse.de/tests/7918770
https://openqa.suse.de/tests/7918771
https://openqa.suse.de/tests/7918772
https://openqa.suse.de/tests/7918773

Fail the case rather than record_soft_failure:
https://openqa.suse.de/tests/7918501#step/version_check/14

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -11,7 +11,7 @@ use warnings;
 use testapi qw(is_serial_terminal :DEFAULT);
 use lockapi 'mutex_wait';
 use mm_network;
-use version_utils qw(is_microos is_leap is_sle is_sle12_hdd_in_upgrade is_storage_ng is_jeos);
+use version_utils qw(is_microos is_leap is_sle is_sle12_hdd_in_upgrade is_storage_ng is_jeos package_version_cmp);
 use Utils::Architectures;
 use Utils::Systemd qw(systemctl disable_and_stop_service);
 use Utils::Backends;
@@ -90,6 +90,7 @@ our @EXPORT = qw(
   permit_root_ssh
   permit_root_ssh_in_sol
   cleanup_disk_space
+  package_upgrade_check
 );
 
 =head1 SYNOPSIS
@@ -2071,6 +2072,35 @@ sub cleanup_disk_space {
 
     # set the snapshot number to 5-10
     assert_script_run('snapper -croot set-config NUMBER_LIMIT=5-10');
+}
+
+
+=head2 package_upgrade_check
+    package_upgrade_check();
+
+This function is used for checking if the package
+is upgraded to the required version
+
+Sample config of parameter of the function below:
+my $pkg_list = {ibmtss => '1.6.0'};
+here, 'libmtss' is the package name, and '1.6.0'
+is target version which needs to be upgrated to
+
+=cut
+
+sub package_upgrade_check {
+    my ($pkg_list, $fail_flag) = @_;
+    foreach my $pkg_name (keys %$pkg_list) {
+        my $current_ver = script_output("rpm -q --qf '%{version}\n' $pkg_name");
+        record_info("Package $pkg_name version", "Current version is $current_ver, target version is $pkg_list->{$pkg_name}");
+        next if (package_version_cmp($current_ver, $pkg_list->{$pkg_name}) >= 0);
+        if ($fail_flag) {
+            die "Error: package $pkg_name is not upgraded yet, please check with developer";
+        }
+        else {
+            record_soft_failure "Warning: package $pkg_name is not upgraded yet";
+        }
+    }
 }
 
 1;
--- a/tests/security/ibmtss/version_check.pm
+++ b/tests/security/ibmtss/version_check.pm
@@ -10,17 +10,16 @@ use base 'opensusebasetest';
 use strict;
 use warnings;
 use testapi;
-use utils 'zypper_call';
+use utils qw(zypper_call package_upgrade_check);
 
 sub run {
     my $self = shift;
     $self->select_serial_terminal;
 
-    zypper_call('in ibmtss');
-    my $tagert_ver = '1.6.0';
-    my $current_ver = script_output("rpm -q --qf '%{version}\n' ibmtss");
-    record_info("Current ibmtss package version is $current_ver, target version is $tagert_ver");
-    die 'The package version is not updated yet, please check with developer' if ($current_ver lt $tagert_ver);
+    # Version check
+    my $pkg_list = {ibmtss => '1.6.0'};
+    zypper_call("in " . join(' ', keys %$pkg_list));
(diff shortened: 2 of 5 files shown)


lilyeyes:
I remember you would like to add this function into lib/utils.pm, am I correct?

rfan1:
> I remember you would like to add this function into lib/utils.pm, am I correct?

Good comment, I am not sure it is good enough to add it, to, any suggestion?

lilyeyes:
> > I remember you would like to add this function into lib/utils.pm, am I correct?
> 
> Good comment, I am not sure it is good enough to add it, to, any suggestion?

It is good.

tests/security/ibmtss/version_check.pm
```diff
@@ -11,16 +11,16 @@ use strict;
...
 
-    zypper_call('in ibmtss');
-    my $tagert_ver = '1.6.0';
-    my $current_ver = script_output("rpm -q --qf '%{version}\n' ibmtss");
-    record_info("Current ibmtss package version is $current_ver, target version is $tagert_ver");
-    die 'The package version is not updated yet, please check with developer' if ($current_ver lt $tagert_ver);
+    # Version check
+    my $pkg_list = {ibmtss => '1.6.0'};
+    zypper_call("in " . join(' ', keys %$pkg_list));
+    package_upgrade_check($pkg_list);
```
Amrysliu, line 23: According to the previous expected result, when package version is less than the target version, the test will exit with fail. But now it only record a soft failure.  fail should be better?

rfan1: Softfail should be better then exit the test here, most of time we will still have function test after pkg version check. then we may have more test coverage in this case.
As we discussed, currently, we don't have many requirements for failing the test there if pkg version is not updated yet. so I didn't add a new parameter to set it

rfan1:
> > > I remember you would like to add this function into lib/utils.pm, am I correct?
> > 
> > 
> > Good comment, I am not sure it is good enough to add it, to, any suggestion?
> 
> It is good enough IMO.
Done

lilyeyes (approved):
lgtm

lib/utils.pm
```diff
@@ -2073,4 +2074,27 @@ sub cleanup_disk_space {
...
 }
 
+
+=head2 package_upgrade_check
+    package_upgrade_check();
+
+This function is used for checking if the package
+is upgraded to the required version.
+
+Sample config of paramter of the function below:
```
jouyingbin, line 2084: typo: paramter

rfan1: Thanks, done

jouyingbin:
Looks like `zypper_call("in " . join(' ', keys %$pkg_list));` is necessary before  `package_upgrade_check($pkg_list);` in every cases.
Why not merge  `zypper_call("in " . join(' ', keys %$pkg_list));` into the package_upgrade_check() function?

chcao (approved):
LGTM

rfan1:
> Looks like `zypper_call("in " . join(' ', keys %$pkg_list));` is necessary before `package_upgrade_check($pkg_list);` in every cases. Why not merge `zypper_call("in " . join(' ', keys %$pkg_list));` into the package_upgrade_check() function?

Good comments, my original plan was adding zypper install into this function. however, it was not a common way to install the packages for all scenarios [for some packages, they are installed by default]. it may introduce "not needed steps" there.

As discussed with Xiaojing, I moved out the zypper installation command to fit more scenarios

jouyingbin:
> > Looks like `zypper_call("in " . join(' ', keys %$pkg_list));` is necessary before `package_upgrade_check($pkg_list);` in every cases. Why not merge `zypper_call("in " . join(' ', keys %$pkg_list));` into the package_upgrade_check() function?
> 
> Good comments, my original plan was adding zypper install into this function. however, it was not a common way to install the packages for all scenarios [for some packages, they are installed by default]. it may introduce "not needed steps" there.
> 
> As discussed with Xiaojing, I moved out the zypper installation command to fit more scenarios

OKay I got it, it means the zipper call install will be a option, if we met some package which are installed by default, we can skip it like this,
```
   my $pkg_list = {ibmtss => '1.6.0'};
   package_upgrade_check($pkg_list);
```
Agree with you.
LGTM.

jouyingbin (approved):

lib/utils.pm
```diff
@@ -2073,4 +2074,32 @@ sub cleanup_disk_space {
...
+
+Sample config of parameter of the function below:
+my $pkg_list = {ibmtss => '1.6.0'};
+here, 'libmtss' is the package name, and '1.6.0'
+is target version which needs to be upgrated to
+
+=cut
+
+sub package_upgrade_check {
+    my ($pkg_list, $fail_flag) = @_;
```
lilyeyes, line 2092: No default value for $fail_flag.

rfan1: If we don't define a value, it should be null, correct? then we will not need to check it.

Please share me some light if any suggest code changes?

lilyeyes: > If we don't define a value, it should be null, correct? then we will not need to check it.
> 
> Please share me some light if any suggest code changes?

You are correct, the code is OK too.

rfan1: Thanks Lili, then let me keep the current logic

lib/utils.pm
```diff
@@ -2073,4 +2074,32 @@ sub cleanup_disk_space {
...
+        record_info("Package $pkg_name version", "Current version is $current_ver, target version is $pkg_list->{$pkg_name}");
+        if ($fail_flag) {
+            die "Error: package $pkg_name is not upgraded yet" if (package_version_cmp($current_ver, $pkg_list->{$pkg_name}) < 0);
+        }
+        else {
+            record_soft_failure "Warning: package $pkg_name is not upgraded yet" if (package_version_cmp($current_ver, $pkg_list->{$pkg_name}) < 0);
+        }
+    }
+}
+
```
Amrysliu, line 2104: suggest to change:
```
        my $current_ver = script_output("rpm -q --qf '%{version}\n' $pkg_name");
        record_info("Package $pkg_name version", "Current version is $current_ver, target version is $pkg_list->{$pkg_name}");
        return undef if (package_version_cmp($current_ver, $pkg_list->{$pkg_name}) >= 0);
        if ($fail_flag) {
            die "Error: package $pkg_name is not upgraded yet";
        } else {
              record_soft_failure "Warning: package $pkg_name is not upgraded yet";
        }
```

rfan1: Nice comments!

Amrysliu (approved):

PR #13929, opened 2021-12-28
xfstests: Shorter label to stable under stress

SeroSun:
xfstests in openqa has a strategy to kill test when it timeout to avoid hanging case. But recently it kill some unnecessary case especially in case their label not match because of under heavy stress in serial. This PR try to enhance this scenario.
- Shorter heartbeat label "heartbeat" to "h", done label "done" to "d"
- Match only first two words in label when checking test not finish to minimal the checking point to make test not stop before timeout

- Verification run: 
http://10.67.133.133/tests/111

And use xfstests_btrfs-btrfs-151-999 to have a try
before: https://openqa.suse.de/tests/7911438
after: https://openqa.suse.de/tests/7916192


lansuse:
LGTM

PR #13930, opened 2021-12-28
wicked: Improve error handling in testsuite

cfconrad:
* Reboot sut if new wicked version was installed
  * Use `assert_script_run()` in `reset_wicked()`
* Check wicked* journals for errors after each tests and before reboot.

- Verification run:
  - https://openqa.suse.de/t7916530 - wicked_basic_sut@cfconrad/os-autoinst-distri-opensuse#pr_reboot_wicked_check_logs => passed
  -https://openqa.suse.de/t7916528 - wicked_basic_sut@cfconrad/os-autoinst-distri-opensuse#pr_reboot_wicked_check_logs => passed

### DIFF
--- a/lib/serial_terminal.pm
+++ b/lib/serial_terminal.pm
@@ -28,6 +28,9 @@ BEGIN {
       serial_term_prompt
       upload_file
     );
+    our @EXPORT_OK = qw(
+      reboot
+    );
 }
 
 our $serial_term_prompt;
@@ -230,6 +233,28 @@ sub upload_file {
     system('mkdir -p ulogs/') == 0 or die('Failed to create ulogs/ directory');
     system(sprintf("cp '%s' '%s'", $tmpfilename, $dst)) == 0
       or die("Failed to finally copy file from '$tmpfilename' to '$dst'");
+}
+
+
+sub reboot {
+    my (%args) = @_;
+    $args{console} //= testapi::current_console;
+    $args{reboot_cmd} //= 'systemctl reboot';
+    $args{timeout} //= 300;
+    my $check_file = '/dev/shm/openqa-reboot-check-' . random_string(8);
+
+    bmwqemu::log_call(%args);
+    die('Only root-virtio-terminal is supported') unless $args{console} eq 'root-virtio-terminal';
+
+    assert_script_run("touch '$check_file'");
+
+    record_info('REBOOT', "cmd: " . $args{reboot_cmd});
+    script_run($args{reboot_cmd}, timeout => 0);
+    reset_consoles;
+    wait_serial('[lL]ogin:', timeout => $args{timeout});
+    select_console($args{console});
+
+    assert_script_run("! test -e '$check_file'");
 }
 
 1;
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -24,7 +24,7 @@ use strict;
 use warnings;
 use Utils::Architectures;
 
-use constant WICKED_DATA_DIR => '/tmp/wicked/data';
+use constant WICKED_DATA_DIR => '/root/wicked/data';
 
 =head2 wicked_command
 
@@ -99,21 +99,21 @@ sub assert_wicked_state {
 sub reset_wicked {
     my $self = @_;
     # Remove any config file and leave the system clean to start tests
-    script_run('find /etc/sysconfig/network/ -name "ifcfg-*" | grep -v "ifcfg-lo" | xargs rm');
-    script_run('rm -f route');
+    assert_script_run('find /etc/sysconfig/network/ -name "ifcfg-*" -not -name "ifcfg-lo" -exec rm {} \;');
+    assert_script_run('find /etc/sysconfig/network/ -name "routes" -o -name "ifroute-*" -exec rm {} \;');
 
     # Remove any previous manual ip configuration
     my $iface = iface();
-    script_run("ip a flush dev $iface");
-    script_run('ip r flush all');
-    script_run("ip link set dev $iface down");
+    assert_script_run("ip a flush dev $iface");
+    assert_script_run('ip r flush all');
+    assert_script_run("ip link set dev $iface down");
 
     file_content_replace("/etc/sysconfig/network/config", "^NETCONFIG_DNS_STATIC_SERVERS=.*" => " ");
     assert_script_run("netconfig -f update");
 
     # Restart services
-    script_run('rcwickedd restart');
-    script_run('rcwicked restart');
+    assert_script_run('rcwickedd restart');
+    assert_script_run('rcwicked restart');
 }
 
 
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -16,6 +16,7 @@ use testapi qw(is_serial_terminal :DEFAULT);
 use serial_terminal;
 use Carp;
 use Mojo::File 'path';
+use Mojo::Util 'trim';
 use Regexp::Common 'net';
 use File::Basename;
 use version_utils 'check_version';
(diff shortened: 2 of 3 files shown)


lib/serial_terminal.pm
```diff
@@ -17,6 +17,7 @@ use Mojo::Util qw(b64_encode b64_decode sha1_sum trim);
 use Mojo::File 'path';
 use File::Basename;
 use File::Temp 'tempfile';
+use Time::HiRes;
```
asmorodskyi, line 20: hm maybe this needs to go here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/cpanfile ?

cfconrad: It's a standard library.

cfconrad: and already used:
```
os-autoinst-distri-opensuse> grn "use Time::HiRes" | wc -l
27
```

lib/serial_terminal.pm
```diff
@@ -232,4 +236,37 @@ sub upload_file {
       or die("Failed to finally copy file from '$tmpfilename' to '$dst'");
 }
 
+
+sub reboot {
+    my (%args) = @_;
+    $args{check} //= 1;
```
asmorodskyi, line 242:
```suggestion
    $args{record_uptime} //= 1;
```

lib/serial_terminal.pm
```diff
@@ -232,4 +236,37 @@ sub upload_file {
...
+    $args{console} //= testapi::current_console;
+    $args{reboot_cmd} //= 'systemctl reboot';
+    $args{timeout} //= 300;
+    $args{minimum_boot_time} = 3;
+
+    bmwqemu::log_call(%args);
+
+    my $uptime_start;
```
asmorodskyi, line 250: it is just IMHO so feel free to ignore this , but `_start` confuses me here and looks unrelated :
```suggestion
    my $uptime;
```

lib/serial_terminal.pm
asmorodskyi: ok I haven't read till the end , so ignore this suggestion :)

lib/serial_terminal.pm
```diff
@@ -232,4 +236,37 @@ sub upload_file {
...
+    die('Only root-virtio-terminal is supported') unless $args{console} eq 'root-virtio-terminal';
+    $uptime_start = script_output(q(cat /proc/uptime | awk '{print $1}')) if ($args{check});
+
+    record_info('REBOOT', "cmd: " . $args{reboot_cmd} . ($uptime_start ? "\nuptime: $uptime_start sec" : ''));
+    script_run($args{reboot_cmd}, timeout => 0);
+    reset_consoles;
+    wait_serial('[lL]ogin:', timeout => $args{timeout});
+    select_console($args{console});
+
+    my $duration = time() - $time_start;
```
asmorodskyi, line 262: please move this line before selecting console , so we are sure that we not hanging somewhere in openQA side

lib/serial_terminal.pm
```diff
@@ -232,4 +236,37 @@ sub upload_file {
...
+    if ($args{check}) {
+        my $uptime_end = script_output(q(cat /proc/uptime | awk '{print $1}'));
```
asmorodskyi, line 264: I am not sure why you involving uptime here , you can simply measure difference between `time()` and `$time_start`

cfconrad: because I will make sure, that the uptime is lower then before. This is for me the indicator, that the reboot really took place.
I didn't found something else like a boot counter or so.

lib/serial_terminal.pm
cfconrad: see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13930#discussion_r775924340 it is really a validation, that the reboot took place and not just a `CTRL+D` happen.

lib/serial_terminal.pm
asmorodskyi: so you can simply check that `$uptime_end < $uptime_start ` ?

asmorodskyi: and in such form I think this check should be unconditional , because if there was no uptime reset than 100% there was no reboot

lib/wickedbase.pm
```diff
@@ -737,6 +738,51 @@ sub record_console_test_result {
...
+    my $default_exclude = 'wickedd=process \d+ has not exited yet; now doing a blocking waitpid';
+    $default_exclude .= ',wickedd-dhcp6=Link-local IPv6 address is marked duplicate:';
+    $default_exclude .= ',wickedd-nanny=: device has been deleted';
+    $default_exclude .= ',wickedd-dhcp4=unable to confirm lease';
+    $default_exclude .= ',wickedd-nanny=: call to org.opensuse.Network.Interface.waitLinkUp\(\) failed: General failure';
+    $default_exclude .= ',wickedd-nanny=: call to org.opensuse.Network.Interface.waitLinkUp\(\) failed: Object does not support requested method';
+    $default_exclude .= ',wickedd-nanny=: failed to bring up device, still continuing';
+    $default_exclude .= ',wickedd=error retrieving tap attribute from sysfs';
+
+    my $exclude_var = get_var(WICKED_CHECK_LOG_EXCLUDE => $default_exclude);
```
asmorodskyi, line 753: we need to add this variables to `variables.md`

lib/serial_terminal.pm
cfconrad: ok, then lets do it unconditional. But a check of $uptime_end < $uptime_start will not work if the reboot takes longer then the $uptime_start was. Thats why I add the duration on top of $uptime_start

lib/wickedbase.pm
cfconrad: Agree but within this PR? I mean there is none of the `WICKED_*` variables in there...

lib/serial_terminal.pm
asmorodskyi: it does not matter how long reboot lasts . Let's image we have system which was running for 5 minutes . Than in [line 254](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13930/files#diff-711effe1cfae054c02b71f62a23c7a765a8aabe3f2ae80e3557e0cc0c01b27f0R254) you will get 300 . After that you trigger reboot , it could take 10 minutes or even 30 minutes you will still have 300 in variable `$uptime_start` , right ? :)  
Later in [line 264](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13930/files#diff-711effe1cfae054c02b71f62a23c7a765a8aabe3f2ae80e3557e0cc0c01b27f0R264) you will get basically time which system needed to boot which suppose to be relatively the same as before . 
The only possible race condition here that if you call this function just after system was boot and you haven't done anything else before calling it , and also there should be some "luck" with worker overloaded only during second boot. 
But I would really leave this corner case out of scope

asmorodskyi: in other words uptime is **relative** value counted from system **start** and be designed can not be influenced by duration of reboot

lib/wickedbase.pm
asmorodskyi: ok

lib/wickedbase.pm
```diff
@@ -737,6 +738,51 @@ sub record_console_test_result {
...
+    my $exclude_test_var = get_var('WICKED_CHECK_LOG_EXCLUDE_' . $self->{name}, '');
```
asmorodskyi, line 754: I am against this variable , I think it will create too much mess . let's start from just single global variable.  Also do you have real examples where you really need it ?  Also in such a case I would prefer to switch to json and still have it as single list but with some additional field which controls whether it should be applied unconditionally or only to some certain cases

lib/wickedbase.pm
```diff
@@ -737,6 +738,51 @@ sub record_console_test_result {
...
+            record_info('LOG-ERROR', $msg, result => 'fail');
+            $self->result('fail') if get_var(WICKED_CHECK_LOG_FAIL => 0) && $self->{name} ne 'before_test';
+        }
+    }
+}
+
+sub reboot {
+    my ($self) = @_;
+    $self->check_logs();
+    serial_terminal::reboot();
```
asmorodskyi, line 783: I think it make sense to do this check also **after** reboot ( but keeping before reboot check too )

lib/serial_terminal.pm
cfconrad: Yes exact the reason why I need to measure the duration as well. Just imagin that your call `reboot()` twice, there is quite a change to hit the case that `$uptime_end` is bigger the `$uptime_start` because of some timeout or what ever in bootup process...

lib/wickedbase.pm
cfconrad: I would keep it but ok... Because it is a nice way to instantly react/debug on unexpected things..

lib/wickedbase.pm
cfconrad: ok

lib/serial_terminal.pm
asmorodskyi: new idea - creation of folder in `/dev/shm` and check it's existence after reboot

lib/wickedbase.pm
asmorodskyi: ok

cfconrad: will keep it!

lib/serial_terminal.pm
cfconrad: thx for pointing this out!

cfconrad:
VR's with latest code:
https://openqa.suse.de/t7936628 
https://openqa.suse.de/t7936626

lib/serial_terminal.pm
cfconrad, line 20: left over...

cfconrad: removed

PR #13932, opened 2021-12-29
Fix remove test user failed on s390x

lemon-suse:
We need logout the test user before remove the test user account. 

- Related ticket: https://progress.opensuse.org/issues/104436
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/7916053#   # online 15SP3
                            https://openqa.nue.suse.com/tests/7916182#step/check_upgraded_service/80  # online 15SP2

### DIFF
--- a/lib/services/users.pm
+++ b/lib/services/users.pm
@@ -190,6 +190,7 @@ sub full_users_check {
         # so we have to change this test to logout and login new user.
         if (is_s390x) {
             logout_and_login($newUser, $pwd4newUser);
+            handle_logout;
         }
         else {
             switch_users;


lib/services/users.pm
```diff
@@ -190,6 +190,7 @@ sub full_users_check {
         # so we have to change this test to logout and login new user.
         if (is_s390x) {
             logout_and_login($newUser, $pwd4newUser);
+            handle_logout;
```
coolgw, line 193: can put 390 logic into func switch_users?

lemon-suse: In fact, we can move this into switch_users, but logout_and_login is not the same meaning of switch users, so it is better to keep the logic outside and judge by is_s390x.

coolgw: Then i suppose logout and remove user should all belong cleanup action?
And this action(logout and remove user) should same on all platform.

lemon-suse: It is another topic, cleanup need to remove user and remove user need logout the user, while we don't know which position failed called the cleanup, so cleanup will be a bit more complicated to think more to enhance while not just combine to use a simple combination function.

coolgw: The reason you add logout here is used for remove user, so why this is another topic?

PR #13933, opened 2021-12-29
Add version_switch_origin_system module to get system version

tinawang123:
Add version_switch_origin_system module to get system version.
We need use correct system version.

- Related ticket: https://progress.opensuse.org/issues/104409
- Verification run: https://openqa.suse.de/tests/7918562#step/patch_sle/59

### DIFF
--- a/schedule/install/offline_spvm_install.yaml
+++ b/schedule/install/offline_spvm_install.yaml
@@ -1,6 +1,8 @@
 name:          offline_spvm_install
 description:    >
     This is prepare install task before migration.
+vars:
+  ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
 schedule:
   - installation/bootloader
   - installation/welcome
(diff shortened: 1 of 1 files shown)


lemon-suse:
It is not needed to add a module, just set UPGRADE_TARGET_RELEASED_VERSION=1 and UPGRADE_TARGET_VERSION=15-SP2 will fix this issue.

schedule/install/offline_spvm_install.yaml
```diff
@@ -28,6 +30,7 @@ schedule:
   - console/system_prepare
   - console/hostname
   - console/force_scheduled_tasks
+  - migration/version_switch_origin_system
```
rfan1, line 33: I don't understand why you use this function in spvm, any operations done there?

tinawang123:
Close this PR, change to use setting to fix this problem:
UPGRADE_TARGET_RELEASED_VERSION=1 
UPGRADE_TARGET_VERSION=15-SP2

PR #13936, opened 2021-12-29
publiccloud : Die early in case issues with SSH key

asmorodskyi:


lib/publiccloud/instance.pm
```diff
@@ -215,8 +214,14 @@ sub wait_for_ssh
...
             $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
         }
-        elsif ($self->run_ssh_command(cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"', proceed_on_failure => 1, quiet => 1, username => $args{username}) =~ m/Reached target.*/) {
-            return $duration;
+        else {
+            my $output = $self->run_ssh_command(cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"', proceed_on_failure => 1, username => $args{username});
+            if ($output =~ m/Reached target.*/) {
+                return $duration;
+            }
+            elsif ($output =~ m/Permission denied (publickey).*/) {
```
grisu48, line 222: I'd move this check to https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L86 to apply it everywhere we use `run_ssh_command` instead of only here.

grisu48 (commented):
One comment, the rest looks good.

lib/publiccloud/instance.pm
asmorodskyi: good idea , I will follow up

asmorodskyi: ok after checking the code I change my mind . I think current place is better than what you suggesting . There are two reasons for this . 
1. This check does not make sense to do on every call. This failure never appear in the middle of the test . It was always the case that something is broken from very first call. If setup done correctly we can not get this error 
2. Moving this code into `run_ssh_command` assume injecting this logic into this code : 
```
  if ($args{timeout} == 0) {
        # Run the command and don't wait for it - no output nor returncode here
        script_run($ssh_cmd, %args);
    } elsif ($rc_only) {
        # Increase the hard timeout for script_run, otherwise our 'timeout $args{timeout} ...' has no effect
        $args{timeout} += 2;
        # Pipe both the standard and error output serial for debug purposes
        $ssh_cmd .= " >/dev/$serialdev 2>&1";
        # Run the command and return only the returncode here
        my $ret = script_run($ssh_cmd, %args, quiet => 0);
        die("Timeout on $ssh_cmd") unless (defined($ret));
        return $ret;
    } else {
        # Run the command, wait for it and return the output
        return script_output($ssh_cmd, %args);
    }
```
ofc this is possible but I just don't see how it can be done nicely because we have 3 way of execution so we will need to duplicate such check 3 times or we need to extract return code processing out which IMO is bad idea . 

Taking into account this 2 point I vote for keeping this check where I put it

grisu48 (approved):
LGTM!

tjyrinki:
This is WIP maybe, or ready for another review? The subject mentions only one of the commits.

asmorodskyi:
I mess up my branch while rebasing :( cherry-picking changes and opening new PR with the same content ...

asmorodskyi:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14076

PR #13938, opened 2021-12-30
Add extra wait_quit_zypper after s390x boot up

coolgw:
Add extra wait_quit_zypper after s390x boot up

- Related ticket: https://progress.opensuse.org/issues/104439
- Needles: na
- Verification run: https://openqa.suse.de/tests/7926824
NOTE: can not easy reproduce this issue with my fix after around 20 runs, current VR only show no side effect for exist test.

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1349,6 +1349,33 @@ sub _handle_login_not_found {
     die "unknown error, system couldn't boot. Detailed bootup log:\n$error_details";
 }
 
+=head2 _handle_firewall
+
+ _handle_firewall();
+
+Internal helper function used by C<reconnect_mgmt_console>.
+
+=cut
+sub _handle_firewall {
+    select_console 'root-console';
+    return if script_run("iptables -S | grep 'A input_ext.*tcp.*dport 59.*-j ACCEPT'", 30) == 0;
+    wait_quit_zypper;
+    my $ret;
+    my $max_num = 3;
+    # Sometimes wait_quit_zypper still not wait enough timeout
+    for (1 .. $max_num) {
+        $ret = zypper_call("in susefirewall2-to-firewalld", exitcode => [0, 7]);
+        if ($ret == 7) {
+            record_info('The ZYPP library is still locked, wait more seconds');
+            wait_quit_zypper;
+        } else {
+            last;
+        }
+    }
+
+    susefirewall2_to_firewalld();
+}
+
 =head2 reconnect_mgmt_console
 
  reconnect_mgmt_console([timeout => $timeout]);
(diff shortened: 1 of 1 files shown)


lib/utils.pm
```diff
@@ -1407,7 +1407,19 @@ sub reconnect_mgmt_console {
...
-                    zypper_call("in susefirewall2-to-firewalld");
+                    my $ret;
+                    my $max_num = 3;
+                    # Sometimes wait_quit_zypper still not wait enough timeout
+                    for (my $i = 1; $i < $max_num + 1; $i = $i + 1) {
+                        $ret = zypper_call("in susefirewall2-to-firewalld", exitcode => [0, 7]);
+                        if ($ret == 7) {
+                            record_info('The ZYPP library is still locked, wait more seconds');
+                            wait_quit_zypper;
+                            next;
```
hjluo, line 1418: No need to next here.

coolgw: thansk! fixed.

rfan1:
sed -i s/s390/s390x/g

lib/utils.pm
```diff
@@ -1349,6 +1349,33 @@ sub _handle_login_not_found {
...
+ _handle_firewall();
+
+Internal helper function used by C<reconnect_mgmt_console>.
+
+=cut
+sub _handle_firewall {
+    select_console 'root-console';
+    if (script_run("iptables -S | grep 'A input_ext.*tcp.*dport 59.*-j ACCEPT'", 30) != 0) {
+        wait_quit_zypper;
+        my $ret;
```
rfan1, line 1363: Is it a bug on s390x?
why locked there if no active daemon?

coolgw: not bug, this is caused by firewall upgrade

lemon-suse (approved):

lib/utils.pm
```diff
@@ -1349,6 +1349,33 @@ sub _handle_login_not_found {
...
+            $ret = zypper_call("in susefirewall2-to-firewalld", exitcode => [0, 7]);
+            if ($ret == 7) {
+                record_info('The ZYPP library is still locked, wait more seconds');
+                wait_quit_zypper;
+            } else {
+                last;
+            }
+        }
+        susefirewall2_to_firewalld();
+    }
```
okurz, line 1376:
```suggestion
    return if script_run("iptables -S | grep 'A input_ext.*tcp.*dport 59.*-j ACCEPT'", 30) == 0;
    wait_quit_zypper
    my $ret;
    my $max_num = 3;
    # Sometimes wait_quit_zypper still not wait enough timeout
    for (1 .. $max_num) {
        $ret = zypper_call("in susefirewall2-to-firewalld", exitcode => [0, 7]);
        if ($ret == 7) {
            record_info('The ZYPP library is still locked, wait more seconds');
            wait_quit_zypper;
        } else {
            last;
        }
    }
    susefirewall2_to_firewalld();
```

rfan1 (approved):

rfan1:
One more comment,  when return code is 7:

           The ZYPP library is locked, e.g. packagekit is running.

Can we also add it into our consideration?

coolgw:
> One more comment, when return code is 7:
> 
> ```
>        The ZYPP library is locked, e.g. packagekit is running.
> ```
> 
> Can we also add it into our consideration?

Base our failed case log this is not caused by packagekit.

PR #13939, opened 2021-12-31
xfstests: Mark skip test as pass and enrich summary info

SeroSun:
To make pass rate comparable between beta1 and other milestone. Change to mark skip to skip tests instead of softfail. And skip cases could also been distinguished. Because only fail and skip case could show in result.
Also enrich summary a little bit with some status information.

- Verification run: https://openqa.suse.de/tests/7939131

### DIFF
--- a/tests/xfstests/generate_report.pm
+++ b/tests/xfstests/generate_report.pm
@@ -61,7 +61,7 @@ sub analyze_result {
             my $test_time = $3;
             (my $test_path = $test_name) =~ s/-/\//;
             $test_num += 1;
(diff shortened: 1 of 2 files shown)


lansuse:
There are only two “skip" and the others are ”pass" in your VR. So the result should be "pass" but got "failed"

SeroSun:
> There are only two “skip" and the others are ”pass" in your VR. So the result should be "pass" but got "failed"

Fixed, let me mark skip tes result as skip to have a try.

lansuse:
LGTM

PR #13940, opened 2021-12-31
Fix the failures in CC syscall test case

Amrysliu:
For syscall test, the test cases that need to be run are not the same
in different arches. So we need to compare the result with the correct
one.

Relate: https://progress.opensuse.org/issues/104235
Verify run: 
         https://openqa.suse.de/tests/7929245#  [aarch64]
         https://openqa.suse.de/tests/7929244#  [x86_64]
         https://openqa.suse.de/tests/7929262#   [s390x]

### DIFF
--- a/lib/audit_test.pm
+++ b/lib/audit_test.pm
@@ -14,6 +14,7 @@ use strict;
 use warnings;
 use testapi;
 use utils;
+use Utils::Architectures;
 use Mojo::File 'path';
 use Mojo::Util 'trim';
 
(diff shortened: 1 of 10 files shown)


lilyeyes:
Why renamed cases' name? e.g., s/fail_safe/fail-safe/

lilyeyes:
I think there may be a MR accordingly, is it?

lib/audit_test.pm
```diff
@@ -64,12 +65,22 @@ sub run_testcase {
...
+
+    # For syscall test case, different arch has different baseline file,
+    # so we need to upload the correct file
+    if ($testcase eq 'syscalls') {
+        $baseline_file = $baseline_file . '.' . get_var('ARCH');
+    }
+    else {
+        $baseline_file = "baseline_run.log";
+    }
+    upload_logs("$baseline_file", (log_name => "$testcase-baseline_run.log"));
```
rfan1, line 83: $baseline_file = $baseline_file . '.' . get_var('ARCH') if ($testcase eq 'syscalls') ;
can work as well right?

lilyeyes (approved):
LGTM after discussed offline as all my Qs were anwered.

rfan1:
LGTM

jouyingbin:
@Amrysliu would you also attach the MR link in the description as there are many changes in MR related to this PR? But it's up to you.

jouyingbin (approved):
LGTM

PR #13941, opened 2021-12-31
Calculate delay and wait for services

lpalovsky:
With newest updates, there is a delay introduced to service start 
after fencing. Current test does not account for that and fails 
because cluster services are not starting. This PR introduces 
mechanism to calculate this delay and waits before starting services.

- Related ticket: https://jira.suse.com/browse/TEAM-5386
- Verification run:
  - Offline migration: https://mordor.suse.cz/tests/2183#dependencies
  - 2node cluster: https://mordor.suse.cz/tests/2145#dependencies
  - 3node cluster: https://mordor.suse.cz/tests/2148#dependencies
  - diskless SBD: https://mordor.suse.cz/tests/2187#dependencies


ricardobranco777 (approved):
LGTM

PR #13860, opened 2021-12-13
Fix disk usage check in validate_btrfs

grisu48:
The _sanity_test_btrfs routine in `validate_btrfs` check if the disk is
not full, but omits some corner cases. This commit changes the behaviour
to accept all disk filling values below 90%.

- Related ticket: https://progress.opensuse.org/issues/102377
- Verification run: [Tumbleweed](https://openqa.opensuse.org/t2083483) | [SLES15-SP3](https://openqa.suse.de/t7845664) | [SLES15](https://openqa.suse.de/t7845665) | [SLES12-SP5](https://openqa.suse.de/t7845680)

PR #13943, opened 2022-01-03
validate_btrfs: fix regex to make 9% a valid usage level

DimStar77:
Fix issue introduced with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13860

- Related ticket: https://progress.opensuse.org/issues/104559
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/2119621

### DIFF
--- a/tests/containers/validate_btrfs.pm
+++ b/tests/containers/validate_btrfs.pm
@@ -39,7 +39,7 @@ sub _sanity_test_btrfs {
     assert_script_run "btrfs fi df $dev_path/btrfs/";
     assert_script_run "ls -td $dev_path/btrfs/subvolumes/* | head -n 1 > $btrfs_head";
     # Ensure the var partition has at least 10% free space
-    validate_script_output "df -h | grep var", sub { m/\/dev\/x?[v,s]d[a-z].+ [0-8][0-9]?%/ };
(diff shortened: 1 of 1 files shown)


asmorodskyi (approved):
LGTM

tests/containers/validate_btrfs.pm
```diff
@@ -39,7 +39,7 @@ sub _sanity_test_btrfs {
     assert_script_run "btrfs fi df $dev_path/btrfs/";
     assert_script_run "ls -td $dev_path/btrfs/subvolumes/* | head -n 1 > $btrfs_head";
     # Ensure the var partition has at least 10% free space
-    validate_script_output "df -h | grep var", sub { m/\/dev\/x?[v,s]d[a-z].+ [0-8][0-9]?%/ };
+    validate_script_output "df -h | grep var", sub { m/\/dev\/x?[v,s]d[a-z].+ [1-8]?[0-9]%/ };
```
asmorodskyi, line 42: I think initial assumption was that values lower than 10 will have format `09` which not the case after all ...

DimStar77: I we prefer, I can write it as `[0-8]?[0-9]%`, which would cater for 09%; but in any case, the important thing is to have the `?` after the first digit, not the 2nd

PR #13948, opened 2022-01-04
Turn off screen saver to avoid lock screen

lemon-suse:
We need turn off screen saver to avoid lock screen, the setting for running gdm work for all arches and we need set it early so that the setting will take effect.

- Related ticket: https://progress.opensuse.org/issues/104074
- Needles: N/A
- Verification run: 
  http://openqa.nue.suse.com/tests/7922030#step/install_service/299  # aarch64 sles15sp3
  http://openqa.nue.suse.com/tests/7922033#step/check_upgraded_service/68 # x86_64 sles15sp1
  https://openqa.nue.suse.com/tests/7939098#step/install_service/197  # ppc64le sles15sp2
  https://openqa.nue.suse.com/tests/7939091#live                                 # s390x 15sp3

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -1,4 +1,4 @@
-# Copyright 2019-2021 SUSE LLC
+# Copyright 2019-2022 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package x11utils;
@@ -29,6 +29,7 @@ our @EXPORT = qw(
   turn_off_plasma_screenlocker
   turn_off_gnome_screensaver
   turn_off_gnome_screensaver_for_gdm
+  turn_off_gnome_screensaver_for_running_gdm
   turn_off_gnome_suspend
   turn_off_gnome_show_banner
   untick_welcome_on_next_startup
@@ -406,6 +407,18 @@ a command prompt, for example an xterm window.
 =cut
 sub turn_off_gnome_screensaver_for_gdm {
     script_run 'sudo -u gdm dbus-launch gsettings set org.gnome.desktop.session idle-delay 0';
+}
+
+=head2 turn_off_gnome_screensaver_for_running_gdm
+
+turn_off_gnome_screensaver_for_running_gdm()
+
+Disable screensaver in gnome for running gdm. The function should be run under root. To be called
+from a command prompt, for example an xterm window.
+
+=cut
+sub turn_off_gnome_screensaver_for_running_gdm {
+    script_run 'su gdm -s /bin/bash -c "DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u gdm)/bus gsettings set org.gnome.desktop.session idle-delay 0"';
 }
 
 =head2 turn_off_gnome_suspend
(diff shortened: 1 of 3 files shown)


lib/x11utils.pm
```diff
@@ -408,6 +409,18 @@ sub turn_off_gnome_screensaver_for_gdm {
...
+=head2 turn_off_gnome_screensaver_for_running_gdm
+
+turn_off_gnome_screensaver_for_running_gdm()
+
+Disable screensaver in gnome for running gdm. The function should be run under root. To be called
+from a command prompt, for example an xterm window.
+
+=cut
+sub turn_off_gnome_screensaver_for_running_gdm {
+    script_run 'su gdm -s /bin/bash -c "DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u gdm)/bus gsettings set org.gnome.desktop.session idle-delay 0"';
```
coolgw, line 421: why not use assert_script_run?

lib/x11utils.pm
```diff
@@ -408,6 +409,18 @@ sub turn_off_gnome_screensaver_for_gdm {
...
 
+=head2 turn_off_gnome_screensaver_for_running_gdm
+
+turn_off_gnome_screensaver_for_running_gdm()
+
+Disable screensaver in gnome for running gdm. The function should be run under root. To be called
+from a command prompt, for example an xterm window.
+
+=cut
+sub turn_off_gnome_screensaver_for_running_gdm {
```
coolgw, line 420: Why not put logic into one function turn_off_gnome_screensaver_for_gdm?

lib/x11utils.pm
tinawang123, line 421: How do user know  when to use 'turn_off_gnome_screensaver_for_gdm' when to use 'turn_off_gnome_screensaver_for_running_gdm'?

lemon-suse: This is just follow the code style in x11utils, I think use script_run may want to make the non-zero return value won't block later test. Sometimes return non-zero doesn't means the setting not take effect.

lib/x11utils.pm
lemon-suse: This is different functions, the new one is for running gdm, this suggestion comes from desktop team.

lib/x11utils.pm
lemon-suse: They are different functions, just as function name, one is for running gdm, one is not. As my understanding, running gdm means the gnome started.

lib/x11utils.pm
```diff
@@ -29,6 +29,7 @@ our @EXPORT = qw(
   turn_off_plasma_screenlocker
   turn_off_gnome_screensaver
   turn_off_gnome_screensaver_for_gdm
+  turn_off_gnome_screensaver_for_running_gdm
```
rfan1, line 32: Any difference between these 2 functions? we have qemu/svirt/zvm/powerVM backend, do we cover all?

lemon-suse: I don't think this is related with backends, just as the different name of the functions, turn_off_gnome_screensaver_for_running_gdm is for 'running' gdm.

zhaocongjia:
LGTM. This new function was suggested by me and it works as expected.

coolgw (approved):

hjluo:
LGTM

PR #13950, opened 2022-01-04
Workaround zypper can't create solverTestCase.tar.bz2

hjluo:
In resolve_dependency_issues, when we hit conflicts and the zypper
--debug-solver will fail for 'zypper: command not found. We can upload the
yast2log for further investigation.

- Related ticket: https://progress.opensuse.org/issues/104497
- Needles: N/A
- Verification run:  
   https://openqa.nue.suse.com/tests/7939141/file/resolve_dependency_issues-y2logs.tar.bz2
  https://openqa.nue.suse.com/tests/7942463/file/resolve_dependency_issues-y2logs.tar.bz2
  https://openqa.nue.suse.com/tests/7946974/file/resolve_dependency_issues-y2logs.tar.bz2
  https://openqa.nue.suse.com/tests/7947344/file/resolve_dependency_issues-y2logs.tar.bz2

### DIFF
--- a/tests/installation/resolve_dependency_issues.pm
+++ b/tests/installation/resolve_dependency_issues.pm
@@ -35,6 +35,10 @@ sub post_fail_hook {
     my $self = shift;
     select_console 'root-console';
     $self->upload_solvertestcase_logs();
+    # workaround to get the y2logs.tar.bz2 at early stage
+    script_run "save_y2logs /tmp/y2logs.tar.bz2";
+    upload_logs "/tmp/y2logs.tar.bz2";
+    set_var('Y2LOGS_UPLOADED', 1);
     $self->SUPER::post_fail_hook;
 }
 
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -514,7 +514,9 @@ Upload C</tmp/solverTestCase.tar.bz2>.
 
 =cut
 sub upload_solvertestcase_logs {
-    script_run("zypper -n patch --debug-solver --with-interactive -l");
+    my $ret = script_run("zypper -n patch --debug-solver --with-interactive -l");
+    # if zypper was not found, we just skip upload solverTestCase.tar.bz2
+    return if $ret != 0;
     script_run("tar -cvjf /tmp/solverTestCase.tar.bz2 /var/log/zypper.solverTestCase/*");
     upload_logs "/tmp/solverTestCase.tar.bz2 ";
 }


tests/installation/resolve_dependency_issues.pm
```diff
@@ -33,8 +33,12 @@ sub run {
 
 sub post_fail_hook {
     my $self = shift;
-    select_console 'root-console';
-    $self->upload_solvertestcase_logs();
```
lemon-suse, line 37: You shouldn't remove the cmd to upload solvertestcase, this will cause can't get the solvertestcase in the future in resolve_dependency_issues.

tests/installation/resolve_dependency_issues.pm
```diff
@@ -33,8 +33,12 @@ sub run {
...
+    select_console 'log-console';
+    if (check_var('UPGRADE', '1')) {
```
rfan1, line 37: Hajian, 
If my understanding is correct, we may not be able to know when the issue happens, so why not uploading this file by default?

hjluo: zypper is not available when we exit from iso installation/update

tests/installation/resolve_dependency_issues.pm
hjluo: In this case, when exiting from iso install/update, zypper is not available in single-user mode.

lemon-suse: I know this, so just add condition to avoid use zypper during installation while not simply remove it.

lemon-suse (approved):

tests/installation/resolve_dependency_issues.pm
```diff
@@ -34,7 +34,13 @@ sub run {
 sub post_fail_hook {
     my $self = shift;
     select_console 'root-console';
-    $self->upload_solvertestcase_logs();
+    if (check_var('UPGRADE', '1')) {
```
coolgw, line 37: Maybe we can check error of zypper command feedback instead of check UPGRADE?

hjluo: the yast2 installation failed to wait for manual intervention  https://openqa.nue.suse.com/tests/7919448#step/resolve_dependency_issues/29
and the solvertest case failed for 'zypper not found'
https://openqa.nue.suse.com/tests/7919448#step/resolve_dependency_issues/37

tests/installation/resolve_dependency_issues.pm
hjluo: done. thanks

chcao (approved):
LGTM

tests/installation/resolve_dependency_issues.pm
```diff
@@ -34,6 +34,11 @@ sub run {
 sub post_fail_hook {
     my $self = shift;
     select_console 'root-console';
+    if (check_var('UPGRADE', '1')) {
```
coolgw, line 37: I think use UPGRADE as condition is not good, since this also can happen during fresh installation case.
We can check zypper command return code if needed.
Also this logic can be hide into upload_solvertestcase_logs in my view

tinawang123 (approved):
LGTM

coolgw (approved):

lemon-suse (approved):

tests/installation/resolve_dependency_issues.pm
hjluo: changed

PR #13951, opened 2022-01-04
Do not stop system reboot timeout on xen-hvm

ge0r:
Stopping system reboot timeout is needed only for svirt-xen-pv

- Related ticket: https://progress.opensuse.org/issues/104580
- Needles: No needles
- Verification runs: https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=ge0r%2Fos-autoinst-distri-opensuse%23no-ok-stop-popup

### DIFF
--- a/schedule/yast/textmode/textmode@svirt-xen.yaml
+++ b/schedule/yast/textmode/textmode@svirt-xen.yaml
@@ -25,7 +25,7 @@ schedule:
   - installation/launch_installation
   - installation/confirm_installation
   - installation/performing_installation/perform_installation
-  - installation/performing_installation/stop_timeout_system_reboot_now
+  - '{{stop_reboot_timeout}}'
   - installation/logs_from_installation_system
   - installation/reboot_after_installation
   - installation/first_boot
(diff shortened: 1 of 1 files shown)


schedule/yast/textmode/textmode@svirt-xen.yaml
```diff
@@ -40,6 +40,11 @@ schedule:
   - console/textinfo
   - console/orphaned_packages_check
   - console/consoletest_finish
+conditional_schedule:
+  stop_reboot_timeout:
+    MACHINE:
```
jknphy, line 45: please use here `VIRSH_VMM_TYPE  linux` we don't trust much in MACHINE for conditional schedule (might change name in the future), also it is a setting that belong more to job groups, so even if it is less clear that linux corresponds to svirt-xen-pv is the common approach, otherwise if you disagree then it is better to have two files for clarity.

ge0r: good point, I went with `MACHINE` for the straightforwardness of the svirt-xen-pv setting but in the end MACHINE names are not prone to staying the same for ever

ge0r: I restarted the verification runs

jknphy (approved):

PR #13954, opened 2022-01-04
Stabilize 'podman search' command test

pdostal:
- Related issue: containers/podman#12738

- Failure: https://openqa.suse.de/tests/7937932#step/podman/94
- Verification run: http://pdostal-server.suse.cz/tests/13002

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -82,7 +82,7 @@ sub basic_container_tests {
     my $tumbleweed = "registry.opensuse.org/opensuse/tumbleweed";
 
     # Test search feature
-    validate_script_output("$runtime search --no-trunc tumbleweed", sub { m/Official openSUSE Tumbleweed images/ });
+    validate_script_output("$runtime search --no-trunc --format \"table {{.Name}} {{.Description}}\" tumbleweed", sub { m/Official openSUSE Tumbleweed images/ });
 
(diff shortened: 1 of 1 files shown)


lib/containers/utils.pm
```diff
@@ -82,7 +82,7 @@ sub basic_container_tests {
     my $tumbleweed = "registry.opensuse.org/opensuse/tumbleweed";
 
     # Test search feature
-    validate_script_output("$runtime search --no-trunc tumbleweed", sub { m/Official openSUSE Tumbleweed images/ });
+    validate_script_output("$runtime search --no-trunc --format \"table {{.Name}} {{.Description}}\" tumbleweed | grep \"docker.io/opensuse/tumbleweed\"", sub { m/Official openSUSE Tumbleweed images/ });
```
asmorodskyi, line 85: I vote for removing grep . It suppose to work without grep . If for some reason "docker.io/opensuse/tumbleweed" will be missing I think it might help to see what is actual output . with this grep we won't see it . If you afraid to catch "Official openSUSE Tumbleweed images" not from "docker.io/opensuse/tumbleweed" you may extend validation string instead

grisu48: Is the `grep` within a `validate_script_output` really necessary?

pdostal: Done.

grisu48:
LGTM

PR #13955, opened 2022-01-04
Shorten 'rpm -q --changelog' command output

pdostal:
- Error: https://openqa.suse.de/tests/7937937#step/rpm/105
- Verification run: http://pdostal-server.suse.cz/tests/13003#step/rpm/53

### DIFF
--- a/tests/console/rpm.pm
+++ b/tests/console/rpm.pm
@@ -69,7 +69,7 @@ sub run {
         # Get detailed information for a package
         assert_script_run("rpm -qi $installed_pkg");
         # Read changelog of a package
-        assert_script_run("rpm -q --changelog $installed_pkg", 90);
+        assert_script_run("rpm -q --changelog $installed_pkg | tail -n 40", 90);
         # List what a package provides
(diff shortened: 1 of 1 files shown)


tests/console/rpm.pm
```diff
@@ -69,7 +69,7 @@ sub run {
         # Get detailed information for a package
         assert_script_run("rpm -qi $installed_pkg");
         # Read changelog of a package
-        assert_script_run("rpm -q --changelog $installed_pkg", 90);
+        assert_script_run("rpm -q --changelog $installed_pkg", 180);
```
asmorodskyi, line 72: I the problem is that it taking too long to pass log over a network . so 180 will help until there will be bigger network problem so it will timeout even with 180 :) what if instead we will do something like this ? : 
```suggestion
        assert_script_run("rpm -q --changelog $installed_pkg | tail -n 40", 90);
```
I don't think it will harm testing coverage

pdostal: This is good idea! VR running.

pdostal:
VR refreshed.

PR #13959, opened 2022-01-05
Add product extensions via url: http & ftp

jknphy:
- Related ticket: [poo#102158](https://progress.opensuse.org/issues/102158)
- Job group MR: https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/389
- Verification run: [SLE](https://openqa.suse.de/tests/overview?build=jknphy%2Fos-autoinst-distri-opensuse%23add_addon_product_url_ftp&distri=sle&version=15-SP4)


schedule/yast/addon_extensions_http_ftp/addon_extensions_http_ftp.yaml
```diff
@@ -13,7 +12,9 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_selection/skip_module_selection
-  - installation/addon_products_sle
+  - installation/add_on_product/add_add_on_http_repo_url
+  - installation/add_on_product_installation/add_add_on_ftp_repo_url
```
ge0r, line 16: Why aren't these two modules in the same directory (`add_on_product_installation`)?

jknphy: they are in different screens, you can check the video, it will depend if it is is already some addon added.

ge0r (approved):
Nice work!

tests/console/validate_addon_extensions_repos_http_ftp.pm
```diff
@@ -0,0 +1,46 @@
...
+#
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use repo_tools 'validate_repo_properties';
+
+sub run {
```
OleksandrOrlov, line 17: Everything looks great for me in this PR. The only thing that could be changed in my opinion is this test module. I guess we can split it into two test modules: validating ftp and validating http to make them more atomic and clear or what is validated where. WDYT?

jknphy: I think I didn't do it because I'm not sure how will go the atomicity with validations, because validation can be randomly different even with the same installation due to different settings left by packages or combination of virtualization + arch. However makes sense in this case, thanks for spotting it. Restarting VRs.

PR #13966, opened 2022-01-06
install_ltp: Fix repeated installation from git

pevik:
It fixes problem during git clone:

```
fatal: destination path 'ltp' already exists and is not an empty directory.
```

Error occurs when debugging LTP with reinstalling LTP from git for job which also run LTP tests, i.e. jobs with INSTALL_LTP=git and LTP_COMMAND_FILE=foo when qcow2 image was based on git installation (including LTP_GIT_URL).

While it'd be possible to try to add required remote (if different than presented) and reset git to it, let's use the simplest way, although it takes a bit longer to reclone git again.

Fixes: 26892d4e2 ("ltp: Add option to install selected parts from git")

Verification run: 
- http://quasar.suse.cz/tests/8062

Verify for regression of other methods
- install_ltp installation from git http://quasar.suse.cz/tests/8063
- install_ltp installation from repo http://quasar.suse.cz/tests/8066 http://quasar.suse.cz/tests/8068

### DIFF
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -178,9 +178,10 @@ sub prepare_ltp_git {
 
     $rel = "-b $rel" if ($rel);
 
-    my $ret = script_run("git clone -q --depth 1 $url $rel", timeout => 360);
+    script_run('rm -rf ltp');
+    my $ret = script_run("git clone -q --depth 1 $url $rel ltp", timeout => 360);
     if (!defined($ret) || $ret) {
-        assert_script_run("git clone -q $url $rel", timeout => 360);
+        assert_script_run("git clone -q $url $rel ltp", timeout => 360);
     }
     assert_script_run 'cd ltp';
     assert_script_run 'make autotools';


pevik:
@mdoucha @czerw FYI

czerw:
Wouldn't be better just skip selected parts if directory already exists?  LTP was already compiled from source on SUT, so we dont't have to repeat it again.  I would just put dir check and extend if condition on https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/cef95ffdfe1720c79a5cb42024dc09e7cc1d0356/tests/kernel/install_ltp.pm#L382

My view is that rebuilding is not needed and PR in current form would just increased complexity.

pevik:
> Wouldn't be better just skip selected parts if directory already exists?

That would effectively disable reusing that qcow for compiling different git branch/fork/etc. Sure, there is workaround to run install_ltp and then that job, but why not fix it?

czerw:
> > Wouldn't be better just skip selected parts if directory already exists?
> 
> That would effectively disable reusing that qcow for compiling different git branch / fork / updated master. Sure, there is workaround to run install_ltp and then that job, but why not fix it?

Or just add `rm -rf` before clone? To make sure we have clear env.

pevik:
Code is wrong, because it does not handle different remote.

> Or just add `rm -rf` before clone? To make sure we have clear env.

That was my first version. Because that is the simplest way, let's use it.

czerw (approved):
LGTM

mdoucha (approved):
LGTM

PR #13967, opened 2022-01-07
In case the tip screen showup slow, add 5s to wait it

tinawang123:
In case the tip screen showup slow, add 5s to wait it

- Related ticket: https://progress.opensuse.org/issues/104538
- Verification run: https://openqa.suse.de/tests/7946525#step/ooffice/4

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -974,7 +974,7 @@ sub libreoffice_start_program {
     my %start_program_args;
     $start_program_args{timeout} = 100 if get_var('LIVECD') && check_var('MACHINE', 'uefi-usb');
     x11_start_program($program, %start_program_args);
-    if (match_has_tag('ooffice-tip-of-the-day')) {
+    if (check_screen('ooffice-tip-of-the-day', 5)) {
         # Unselect "_S_how tips on startup", select "_O_k"
         send_key "alt-s";
         send_key "alt-o";


rfan1:
Suggest commit change:
Add wait time to handle tip screen show up slowly

Others LGTM

chcao (approved):
LGTM

lib/x11test.pm
```diff
@@ -974,7 +974,7 @@ sub libreoffice_start_program {
     my %start_program_args;
     $start_program_args{timeout} = 100 if get_var('LIVECD') && check_var('MACHINE', 'uefi-usb');
     x11_start_program($program, %start_program_args);
-    if (match_has_tag('ooffice-tip-of-the-day')) {
```
lemon-suse, line 977: Maybe just add wait_screen_still 5; will fix this unstable issue.

lemon-suse:
Add Grace to review since this is test code of desktop group.

lib/x11test.pm
tinawang123: check_screen just change on line with wait time. I think it's ok.

lemon-suse: In fact, match_has_tag and check_screen both work for this part, add wait before catch the needle is the key point for this issue. Just wait_screen_still save the original code design. Anyway, not big concern from my side.

GraceWang571 (approved):
LGTM

lemon-suse (approved):

PR #13974, opened 2022-01-07
svirt: Skip asset upload if os-autoinst implements it

kalikiana:
See: https://progress.opensuse.org/issues/104520

### DIFF
--- a/tests/shutdown/svirt_upload_assets.pm
+++ b/tests/shutdown/svirt_upload_assets.pm
@@ -11,6 +11,7 @@ use strict;
 use warnings;
 use testapi;
 use version_utils 'is_vmware';
+use backend::svirt ();
 
 sub extract_assets {
     my ($args) = @_;
@@ -38,6 +39,8 @@ sub extract_assets {
 }
 
 sub run {
+    # Skip if the os-autoinst backend implements this feature
+    return 1 if defined &backend::svirt::do_extract_assets;
     # Not implemented on VMware
     return 1 if is_vmware;
     # connect to VIRSH_HOSTNAME screen and upload asset from there


tests/shutdown/svirt_upload_assets.pm
```diff
@@ -38,6 +39,8 @@ sub extract_assets {
 }
 
 sub run {
+    # Skip if the os-autoinst backend implements this feature
+    return 1 if backend::svirt->can('do_extract_assets');
```
perlpunk, line 43: As pointed out by @Martchus [here](https://github.com/os-autoinst/os-autoinst/pull/1895#discussion_r783264756), this will return true because it's implemented n the baseclass and `can()` will check parent classes.
`if defined &backend::svirt::do_extract_assets` should work instead.

tests/shutdown/svirt_upload_assets.pm
```diff
@@ -38,6 +39,8 @@ sub extract_assets {
 }
 
 sub run {
+    # Skip if the os-autoinst backend implements this feature
+    return 1 if defined &backend::svirt::do_extract_assets;
```
kalikiana, line 43: I assume this should stay **as-is** and **not** follow the `can` approach from #1895, because this is specifically about wether _svirt_ implements the method.

perlpunk: #1895 ?

Yes.

perlpunk (approved):

tests/shutdown/svirt_upload_assets.pm
kalikiana: Sorry, linking between repos on GitHub sucks. Of course I meant https://github.com/os-autoinst/os-autoinst/pull/1895#pullrequestreview-861472088

okurz (approved):

okurz (changes requested):
Please don't rely on backend internals, that's not part of a public API

tests/shutdown/svirt_upload_assets.pm
okurz, line 43: how about checking for the API version number from os-autoinst OpenQA/Isotovideo/Interface.pm

okurz (approved):
Ok, after discussing I am fine either way. It's messy so let's hope we will soon remove the code from here completely.

PR #13975, opened 2022-01-07
Save yaml file

pevik:
LTP_KNOWN_ISSUES support for o3 (poo#104613)

Verification run:
- http://quasar.suse.cz/tests/8098#step/ustat01/12

Verify no regression introduced on SLES:
- http://quasar.suse.cz/tests/8099#step/ustat01/12


pevik:
@mdoucha @czerw it might be better to put `set_var('LTP_KNOWN_ISSUES, ...` into `lib/main_ltp.pm`. WDYT?

lib/Parsers/Convert.pm
```diff
@@ -0,0 +1,34 @@
...
+
+package Parsers::Convert;
+
+use base Exporter;
+
+use strict;
+use warnings;
+
+use JSON;
+use YAML::Any;
```
perlpunk, line 14: Is there a specific reason to use YAML::Any here, but for Parsers::YAML to use YAML::PP?
I think it would make more sense to be consistent.
The parsers YAML::Any supports behave all differently, and it doesn't support YAML::PP. Depending on what is installed you could get very different behaviour.
I don't recommend YAML::Any.
YAML::PP is the perl YAML module that is closest to the spec.

lib/LTP/WhiteList.pm
```diff
@@ -129,6 +140,10 @@ sub find_whitelist_testsuite {
     return undef unless defined($path) and -e $path;
 
     my $content = path($path)->slurp;
+    my (undef, undef, $suffix) = fileparse($path, '\.[^\.]*');
+    if ($suffix ne '.json') {
+        $content = yaml2json($content);
+    }
     my $issues = Mojo::JSON::decode_json($content);
```
mdoucha, line 147: This PR is needlessly compilcated. Just save the raw data to `ulogs` in `download_whitelist()` and add an `if/else` above:

    if ($suffix eq '.json') {
      $issues = decode_json($content);
    } elsif ($suffix eq '.yaml') {
      $issues = $yp->load_string($content);
    } else {
      die 'Invalid format';
    }

The data structure is the same in both YAML and JSON.

mdoucha (changes requested):

lib/LTP/WhiteList.pm
```diff
@@ -129,6 +140,10 @@ sub find_whitelist_testsuite {
     return undef unless defined($path) and -e $path;
 
     my $content = path($path)->slurp;
+    my (undef, undef, $suffix) = fileparse($path, '\.[^\.]*');
+    if ($suffix ne '.json') {
+        $content = yaml2json($content);
```
perlpunk, line 145: Is there a reason to load native data from YAML, then convert it to a JSON string, and then decode this string to native data again two lines later?
Wouldn't `$content = YAML::PP::Load(decode_utf8 $content)` or even `$content = YAML::PP::LoadFile($path)` be enough?

lib/LTP/WhiteList.pm
perlpunk: Just utf8 decoding missing.
`$issues = $yp->load_string(decode_utf8 $content);`
or
`$yp->load_file($path)` will do the decoding automatically

t/09_parsers.t
```diff
@@ -0,0 +1,40 @@
...
+---
+syscalls:
+  ustat01:
+  - message: ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208
+    product: opensuse-Tumbleweed
+    retval: ^1$
+EOM
+
+subtest 'yaml' => sub {
+    ok save_yaml_file($valid_yaml, 'valid.yaml'), 'YAML file can be saved with save_yaml_file()';
```
perlpunk, line 29: For my taste this test wouldn't be sufficient.
It only tests that the function doesn't crash and returns a truthy value.
I would check that the file exists and has the expected content.

lib/LTP/WhiteList.pm
pevik: No, again :). Thx!

czerw:
I would prefer to have data  converted to json in openSUSE CI like we do on Gitlab for SLE to avoid this manipulation/conversion in openQA.

pevik:
But that would required to commit it, which I'd prefer not.

lib/LTP/WhiteList.pm
```diff
@@ -33,14 +37,17 @@ sub download_whitelist {
...
     my $basename = $path =~ s#.*/([^/]+)#$1#r;
-    save_tmp_file($basename, $res->body);
-    set_var('LTP_KNOWN_ISSUES', hashed_string($basename));
+
     mkdir('ulogs') if (!-d 'ulogs');
-    bmwqemu::save_json_file($res->json, "ulogs/$basename");
+    save_tmp_file($basename, $res->body);
+    copy(hashed_string($basename), "ulogs/$basename");
+
+    set_var('LTP_KNOWN_ISSUES_LOCAL', "ulogs/$basename");
```
mdoucha, line 50: Reading whitelist data from the `ulogs` copy makes the `save_tmp_file()` unnecessary. But I believe that the temp file was created because the `ulogs` copy might get accidentally overwritten if one of the test modules uploads a log file with the same filename.

lib/LTP/WhiteList.pm
```diff
@@ -33,14 +37,17 @@ sub download_whitelist {
     my $res = Mojo::UserAgent->new->get($path)->result;
     unless ($res->is_success) {
         record_info("File not downloaded!", $res->message, result => 'softfail');
-        set_var('LTP_KNOWN_ISSUES', undef);
+        set_var('LTP_KNOWN_ISSUES_LOCAL', undef);
```
mdoucha, line 40: This probably needs more thought. If the local file already exists, do we really want to not use it?

lib/LTP/WhiteList.pm
```diff
@@ -124,12 +131,22 @@ sub is_test_disabled {
...
     my $content = path($path)->slurp;
-    my $issues = Mojo::JSON::decode_json($content);
+    my $suffix = $path =~ s#.*\.([^.]+)#$1#r;
+    if ($suffix eq 'json') {
+        $issues = Mojo::JSON::decode_json($content);
+    } elsif ($suffix eq 'yaml') {
+        my $yp = YAML::PP->new;
+        $issues = $yp->load_string(decode_utf8 $content);
+    } else {
+        die 'Invalid format';
```
mdoucha, line 147: I hoped you'd write a bit more descriptive error message instead of copying my example string verbatim.

mdoucha (changes requested):

lib/main_common.pm
```diff
@@ -174,6 +174,8 @@ sub setup_env {
         set_var('INSTLANG', 'en_US');
     }
 
+    set_var('LTP_KNOWN_ISSUES', 'https://raw.githubusercontent.com/openSUSE/kernel-qe/main/ltp_known_issues.yaml') if is_opensuse and !get_var('LTP_KNOWN_ISSUES');
```
okurz, line 177:
```suggestion
    set_var('LTP_KNOWN_ISSUES', get_var('LTP_KNOWN_ISSUES', 'https://raw.githubusercontent.com/openSUSE/kernel-qe/main/ltp_known_issues.yaml')) if is_opensuse;
```

okurz (commented):
just a minor suggestion. Rest looks fine to me

lib/main_common.pm
mdoucha: The original will fix empty string. Your suggested version will not.

pevik: That could lead to setting variable to the same value and redundant `bmwqemu::save_vars()` call. Thus, if you don't mind I'd prefer the original code.

Vogtinator:
My only concern is that this makes test runs less reproducible, but the used file is uploaded as asset to this can be avoided.

So the approach itself looks good to me.

lib/LTP/WhiteList.pm
```diff
@@ -12,10 +12,14 @@ use warnings;
 use testapi;
 use bmwqemu;
 use bugzilla;
+use Encode;
 use Exporter;
+use File::Basename 'fileparse';
```
perlpunk, line 17: I don't see this function used anywhere...

lib/LTP/WhiteList.pm
pevik: OK, I'll just create file directly:

```
mkdir('ulogs') if (!-d 'ulogs');
my $file = "ulogs/$basename";
open(my $fh, '>', $file) or die "Could not open file '$file' $!";
print $fh $res->body;
close $fh;

```

lib/LTP/WhiteList.pm
pevik: Sorry, fixed.

asmorodskyi:
I am not sure why we need to introduce separate variable . In my understanding we can use `LTP_KNOWN_ISSUES` in a same way we use it internally , just for o3 point it into different target . what I am missing ?

mdoucha:
> Another question why we introducing YAML format if we already using JSON internally . I don't care actually which one will be used but I think we should choose one. If we saying that YAML is better fine for me but let's move internal list into YAML than too

We can't read the YAML file straight from internal repo due to GitLab stability issues. And rewriting the CI that converts YAML to JSON would be more work than just supporting both in OpenQA.

lib/LTP/WhiteList.pm
mdoucha: I'm talking about setting `LTP_KNOWN_ISSUES_LOCAL` to `"ulogs/$basename"` instead of `hashed_string($basename)`. The file writes are fine.

lib/LTP/WhiteList.pm
pevik: But `download_whitelist()` is called only once for particular job and pool directory is cleared after job. Thus it shouldn't be local file if it wasn't possible to download it before. What am I missing?

lib/LTP/WhiteList.pm
```diff
@@ -124,12 +130,27 @@ sub is_test_disabled {
...
+    my $suffix = $path =~ s#.*\.([^.]+)#$1#r;
+
+    if (!$suffix) {
+        die "Using '$suffix', but only JSON and YAML are supported";
+    }
+
+    if ($suffix eq 'json') {
+        $issues = Mojo::JSON::decode_json($content);
+    } elsif ($suffix eq 'yaml') {
+        my $yp = YAML::PP->new;
```
cfconrad, line 148:
```suggestion
        my $yp = YAML::PP->new(schema => [qw/ + Merge /]);
```
I think merge keys are a really cool feature in yaml and I would enable them.

pevik: Correct me if I'm wrong, but YAML Merge Key draft (https://yaml.org/type/merge.html) is not supported on JSON which we also support. Not sure whether `yamllint` transformation which we use for osd would handle that.

cfconrad: Hm, I don't know where `yamllint` is used here sorry! I only know, that the group yaml files of osd do support merge keys as well.
And only for illustration, I don't say that it is more readable, it could be used like this:
```yaml
---

.anchors:
    - &tw { product: opensuse:Tumbleweed }

syscalls:
    ustat01: &blub
    - <<: *tw
      retval: ^1$
      message: ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208
    ustat02:
      *blub

ima:
    boot_ltp:
    - <<: *tw
      message: Wrong lsm kernel parameter breaks boot. bsc#1194332
```
But in any case, we could add this later if really needed :)

pevik: I'll test it tomorrow.
@mdoucha WDYT?

pevik:
> My only concern is that this makes test runs less reproducible, but the used file is uploaded as asset to this can be avoided.
> 
> So the approach itself looks good to me.

Thx for having look. It should help the reviewers. Also the only difference is soft failure (yellow) with explanation vs. hard failure (red).

lib/LTP/WhiteList.pm
pevik: FYI it'd fail for `yamllint`:
```
$ yamllint ltp_known_issues.yaml 
ltp_known_issues.yaml
  3:12      error    too many spaces inside braces  (braces)
  3:41      error    too many spaces inside braces  (braces)
```
Maybe we need some switch.
Also conversion we use in osd creates a bit different structure (I'd need to test it whether `{"<<":{"product":"opensuse:Tumbleweed"}` would work):
```
$ perl -MJSON -MYAML::XS -E 'local $/; say to_json(Load(<>));' ltp_known_issues.yaml
# current
{"syscalls":{"ustat01":[{"product":"opensuse:Tumbleweed","message":"ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208","retval":"^1$"}],"ustat02":[{"retval":"^1$","product":"opensuse:Tumbleweed","message":"ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208"}]}} # old
# merge keys
{"syscalls":{"ustat02":[{"<<":{"product":"opensuse:Tumbleweed"},"message":"ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208","retval":"^1$"}],"ustat01":[{"retval":"^1$","message":"ustat() is known to fail with EINVAL on Btrfs. Downstream fixes didn't make it to upstream and were removed. bsc#1194208","<<":{"product":"opensuse:Tumbleweed"}}]},".anchors":[{"product":"opensuse:Tumbleweed"}]}
```

perlpunk: YAML::XS doesn't implement merge keys. YAML::PP is the only perl module that can do this.

perlpunk: yamllint doesn't care about merge keys. it complains about spaces inside braces. it's a linter :)

pevik: Ah, thx (I was blind). We could switch to `YAML::PP` in in gitlab CI, but I'd postpone it for separate PR.

cfconrad: for convenience I use `yamlpp-load-dump` for testing like:
```bash
echo '---
.blub: &blub
  foo: 'bla'

.bla2: &bla2
  foo: 'shubidu'

test2:
  <<: [*bla2, *blub]
' | yamlpp-load-dump --preserve 0 --merge
```

czerw:
> But that would required to commit it, which I'd prefer not.
 I don't see any issue with committed json generated from yaml in repo, its simple straightforward solution vs increased complexity in test code.

lib/main_common.pm
czerw: But what if we don't want to use ltp known issues at all? It will not be possible with this  line.

czerw:
> @mdoucha @czerw it might be better to put `set_var('LTP_KNOWN_ISSUES, ...` into `lib/main_ltp.pm`. WDYT?

definitely no

czerw (approved):

czerw:
All clear for me after call with @mdoucha @pevik . LGTM

asmorodskyi (approved):
I still think that it is better to drop JSON and fully switch to YAML everywhere . If you will do it in this PR we can merge https://gitlab.suse.de/qa-maintenance/metadata/-/merge_requests/567 and fully switch to YAML . But it don't want to block you if you think different . other than that LGTM

lib/LTP/WhiteList.pm
mdoucha: If merge keys are supported in the parser package, there's no harm from enabling them. But I'll leave the decision up to you.

lib/LTP/WhiteList.pm
```diff
@@ -124,12 +131,22 @@ sub is_test_disabled {
...
     my $content = path($path)->slurp;
-    my $issues = Mojo::JSON::decode_json($content);
+
+    if (substr($content, 0, 1) eq '{') {
+        $issues = Mojo::JSON::decode_json($content);
+    } elsif (substr($content, 0, 3) eq '---') {
+        my $yp = YAML::PP->new;
+        $issues = $yp->load_string(decode_utf8 $content);
+    } else {
+        die "Only valid JSON and YAML are supported";
```
mdoucha, line 147: This error message will stil look completely incomprehensible to anyone who runs into it. This is better:

    die "LTP_KNOWN_ISSUES must link to valid JSON or YAML file";

lib/LTP/WhiteList.pm
mdoucha: BTW, if you use `substr()` to detect format like that and some text editor inserts the Unicode byte order mark (0xFEFF) at the beginning of the file, which is completely invisible to the user and still valid, you'll end up with in the error branch of the format check. It's better to use regex and allow whitespace at the beginning.

lib/LTP/WhiteList.pm
```diff
@@ -124,12 +131,22 @@ sub is_test_disabled {
...
     return undef unless defined($path) and -e $path;
 
     my $content = path($path)->slurp;
-    my $issues = Mojo::JSON::decode_json($content);
+
+	if ($content =~ m/^\x{FEFF}*[^:ascii:]*\{/) {
+        $issues = Mojo::JSON::decode_json($content);
+	} elsif ($content =~ m/^\x{FEFF}*[^:ascii:]*---/) {
+        my $yp = YAML::PP->new(schema => [qw/ + Merge /]);
+        $issues = $yp->load_string(decode_utf8 $content);
```
perlpunk, line 145: If this is to check whether the input is JSON or YAML and cannot be determined from the file extension, then reading everything with YAML::PP might be the better way, as JSON is a subset of YAML.
There are however two things that YAML::PP doesn't support yet:
* Byte Order Marks
* JSON Surrogate Pairs

pevik: In the end I chose this solution, i.e. parse everything with `YAML::PP`, but before remove BOM.
I tested it various crafted files (BOM, spaces):
https://raw.githubusercontent.com/pevik/kernel-qe/bom/bom-space.json
https://raw.githubusercontent.com/pevik/kernel-qe/bom/bom-space-no-header.yaml

pevik:
@mdoucha @perlpunk can you have a quick look before I merge?

perlpunk:
The YAML part looks good to me, I can't say anything about the other stuff :)

mdoucha (approved):
LGTM

PR #13980, opened 2022-01-10
Add missing module for required installation step

b10n1k:
15SP4 asks to set up the root password before continue. The yaml scheduler is missing condition
which will schedule the `user_settings_root` module.

- Verification run: http://aquarius.suse.cz/tests/7690


schlad:
I'm confused. Haven't you remove that yesterday?

schlad:
> I'm confused. Haven't you remove that yesterday?

ah I see; that was also adding and it was for a node.

schlad (approved):

PR #13981, opened 2022-01-10
Update installation_node_dev yaml scheduler adjusting to 15sp4

b10n1k:
i havent found the scheduler in any other job but i added a conditional module
for the 15sp4. some other minor adjustement too, as the `installation/hostname_inst`
doesnt look to be needed.

- Verification run: http://aquarius.suse.cz/tests/7693


schlad:
If I recall well, installation/hostname_inst was helping to use this for running multimachine tests; so long as we don't really use it for that cases, we indeed might be good removing this module.

schlad (approved):

PR #13985, opened 2022-01-11
Containers: unschedule rootless podman on 15-SP1

jlausuch:
This feature requires at least kernel 4.18.X, and we have
4.12.14 in SLE 15 SP1.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1182874#c13


lib/main_containers.pm
```diff
@@ -61,7 +61,7 @@ sub load_host_tests_podman {
         loadtest 'containers/podman_3rd_party_images';
         loadtest 'containers/podman_firewall';
         loadtest 'containers/buildah';
-        loadtest 'containers/rootless_podman';
+        loadtest 'containers/rootless_podman' unless is_sle("15-sp1");    # https://github.com/containers/podman/issues/5732#issuecomment-610222293
```
dcermak, line 64: Would this get scheduled for 15 GA?

jlausuch: no, because the above condition:
`if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+")) {`

None of the podman tests are executed for 15GA and below.

dcermak (approved):

pdostal (approved):

PR #13988, opened 2022-01-11
Select disk for encryption in cryptlvm_iscsi

OleksandrOrlov:
The commit adds selection of the first disk in Guided Setup Wizard
during encrypt lvm scenario on iscsi, as the step is appeared on guided
setup, but was missed in test suite.

- Related ticket: https://progress.opensuse.org/issues/103718
- Verification run: https://openqa.suse.de/tests/7962344


rakoenig (approved):
LGTM.

PR #13990, opened 2022-01-11
Wait for Installation Settings screen to be loaded

OleksandrOrlov:
validate_ssh_service_enabled test module checked that Installation
Screen is loaded by ensuring that widget with "id=proposal" is shown on
the screen. Then 'assert_true' is called immediately and it failed,
because the content of 'proposal' widget was not loaded.

The commit ensures that the the content is loaded before validating that
the port is opened.

- Related issue: https://progress.opensuse.org/issues/104703
- Verification run: https://openqa.suse.de/tests/7965573

### DIFF
--- a/lib/Installation/InstallationSettings/InstallationSettingsController.pm
+++ b/lib/Installation/InstallationSettings/InstallationSettingsController.pm
@@ -29,7 +29,16 @@ sub init {
 sub get_installation_settings_page {
     my ($self) = @_;
     die "Installation Settings Page is not displayed" unless $self->{InstallationSettingsPage}->is_shown();
+    die "Overview content on Installation Settings Page is not loaded completely" unless $self->{InstallationSettingsPage}->is_loaded_completely();
     return $self->{InstallationSettingsPage};
+}
+
+sub wait_for_overview_content_to_be_loaded {
+    my ($self) = @_;
+    YuiRestClient::Wait::wait_until(object => sub {
+            my $overview_content = $self->get_installation_settings_page()->get_overview_content();
+            return ($overview_content =~ m/SSH port will be/);
+    }, timeout => 60, message => "Overview content is not loaded.");
 }
 
 sub enable_ssh_service {
--- a/lib/Installation/InstallationSettings/InstallationSettingsPage.pm
+++ b/lib/Installation/InstallationSettings/InstallationSettingsPage.pm
@@ -57,6 +57,18 @@ sub is_shown {
     return $self->{txt_overview}->exist();
 }
 
+sub is_loaded_completely {
+    my ($self) = @_;
+    my $result;
+    eval {
+        $result = YuiRestClient::Wait::wait_until(object => sub {
+                my $overview_content = $self->get_overview_content();
+                return ($overview_content =~ m/SSH port will be/);
+        }, timeout => 60, message => "Overview content is not loaded.");
+    };
+    $result ? 1 : 0;
+}
+
 sub open_ssh_port {
     my ($self) = @_;
     YuiRestClient::Wait::wait_until(object => sub {


tests/installation/installation_settings/validate_ssh_service_enabled.pm
```diff
@@ -14,6 +14,7 @@ use Test::Assert 'assert_true';
 
 sub run {
     my $installation_settings = $testapi::distri->get_installation_settings();
+    $installation_settings->wait_for_overview_content_to_be_loaded();
```
jknphy, line 17: what about to run this method after the action that we know cause a refreshment, so basically we could do it internally after changing the service or the port, the test doesn't know about refreshments let's say, but we can go with this if you want it is a good workaround without having the solution in server side, wdyt?

OleksandrOrlov: But in this case we don't do changes that cause the refreshment on Installation Setting. It's just initial page load when navigating from previous screen to Installation Settings.

jknphy: any reason why not to put it in `get_overview_content` ? At least I would have put it there :) but yeah, just to gathering feedback from you as I'm not sure about this and I had the same situation in my previous PRs. We can merge either way.

OleksandrOrlov: The reason is that if we will need to create a test that checks, that the page has placeholders when it is loading, we'll not be able to do so. For sure that may never be the case, but we'll keep single responsibility principle which is also keep our solution more consistent.

jknphy: ahm I was referring to `is_shown()` which means that if the screen is loaded, I guess that method should be more complicated in this particular case, so we can break the single principle because we don't have the proper server solution. For the rest of screen we don't need to do that because there is not refreshment. With this approach we always have to add this new method to test (which we are not supposed to change much). Does it convince you? :)

ge0r (approved):
LGTM! I also would argue that a wait for the content to load should not be visible to the test but it makes sense to have the option to test the page before the overview has been appeared.

lib/Installation/InstallationSettings/InstallationSettingsController.pm
```diff
@@ -29,9 +29,18 @@ sub init {
 sub get_installation_settings_page {
     my ($self) = @_;
     die "Installation Settings Page is not displayed" unless $self->{InstallationSettingsPage}->is_shown();
+    die "Overview content on Installation Settings Page is not loaded completely" unless $self->{InstallationSettingsPage}->is_loaded_completely();
```
jknphy, line 32: Can we merge these two lines? are not the same thing? is_shown, is_loaded are the same to me, but check my comment below.

lib/Installation/InstallationSettings/InstallationSettingsController.pm
```diff
@@ -29,9 +29,18 @@ sub init {
...
     return $self->{InstallationSettingsPage};
 }
 
+sub wait_for_overview_content_to_be_loaded {
```
jknphy, line 36: is this still in use, or you forgot to delete it?

lib/Installation/InstallationSettings/InstallationSettingsPage.pm
```diff
@@ -57,6 +57,18 @@ sub is_shown {
     return $self->{txt_overview}->exist();
 }
 
+sub is_loaded_completely {
+    my ($self) = @_;
+    my $result;
+    eval {
```
jknphy, line 63: so you silent the method with the `eval` I think this was a no-go, only should be used in the client itself, other contributors claimed it was bad practice in Perl, it doesn't behave so well as try/catch in other language (some modern languages even avoid try/catch as a build-in feature of the language).

I think it would be easier just to use the `is_shown` and add to `wait_until` a non-failure feature so we do the hack at the inner level, the only problem is the error if we die or not to use it for what I can see ... wdyt?

jknphy:
Given that https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14022 is still not done, in order to not block it for more time, let's merge it. Anyway the approach with the `eval` was also used in the past.

PR #14003, opened 2022-01-13
zfs: Update filesystem repo addresses to current state

ggardet:
- Verification run: 15.4: https://openqa.opensuse.org/t2134565
The test fails at a later point because zfs fails to build on Leap 15.4 for x86_64


grisu48 (approved):
LGTM! Thanks for the fix :+1:

PR #14006, opened 2022-01-13
Move sys_param_check testkit location to a setting

alvarocarvajald:
Currently the `sys_param_check.pm` test module downloads the external robotfw testkit from a server hard-coded in the module. This commit moves this hard-coded value to the **SYS_PARAM_CHECK_TEST** setting, leaving the current value as the default.

- Related ticket: N/A
- Needles: N/A
- Verification runs: [12-SP5](http://mango.qa.suse.de/tests/4784), [15-SP3](http://mango.qa.suse.de/tests/4782) & [15-SP4](http://mango.qa.suse.de/tests/4786). Failure in 15-SP4 is known ( [bsc#1191396](https://bugzilla.suse.com/show_bug.cgi?id=1191396) ) and unrelated to this PR.

### DIFF
--- a/tests/sles4sap/sys_param_check.pm
+++ b/tests/sles4sap/sys_param_check.pm
@@ -6,6 +6,8 @@
 # Summary: Check the sles4sap "from scratch" settings (without saptune/sapconf) using the robot framework.
 #          This test is configured to be used with a 2 GB RAM system.
 #          Some values depend of the hardware configuration.
+#          External robotfw testkit is supplied in the SYS_PARAM_CHECK_TEST setting. Defaults to
+#          qa-css-hq.qa.suse.de/robot.tar.gz.
 # Maintainer: Julien Adamek <jadamek@suse.com>
 
 use base "sles4sap";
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM. Thanks!

bs-suse:
Maybe the $testkit var should be in quotes to make sure it will still work in case there are characters like spaces in the string that would otherwise break the call?

alvarocarvajald:
> Maybe the $testkit var should be in quotes to make sure it will still work in case there are characters like spaces in the string that would otherwise break the call?

Great catch. Will add the quotes and test again. Thanks.

alvarocarvajald:
> Maybe the $testkit var should be in quotes to make sure it will still work in case there are characters like spaces in the string that would otherwise break the call?

I added the quotes around `$testkit`. Here are the verification runs:

12-SP5: http://mango.qa.suse.de/tests/4789#step/sys_param_check/3
15-SP3: http://mango.qa.suse.de/tests/4787#step/sys_param_check/3
15-SP4: http://mango.qa.suse.de/tests/4788#step/sys_param_check/3

PR #14009, opened 2022-01-13
Add a module to import users with restapi in installer

JRivrain:
Replaces the module users_import by one that uses libyui-rest-api.

- Related ticket: https://progress.opensuse.org/issues/104791

- Verification run: http://waaa-amazing.suse.cz/tests/16127
https://openqa.suse.de/tests/overview?build=JRivrain/os-autoinst-distri-opensuse%2314009

### DIFF
--- a/lib/Installation/LocalUser/LocalUserController.pm
+++ b/lib/Installation/LocalUser/LocalUserController.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2021 SUSE LLC
+# Copyright 2021-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: The class introduces business actions for Local user dialog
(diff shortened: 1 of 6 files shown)


lib/Installation/Navigation/NavigationBase.pm
```diff
@@ -32,7 +32,7 @@ sub is_shown {
 sub press_next {
     my ($self) = @_;
     YuiRestClient::Wait::wait_until(object => sub {
-            return $self->{btn_next}->is_enabled();
+            return $self->{btn_next}->exist();
```
JRivrain, line 35: This does not work, not sure why.

jknphy (commented):
Good job Jonathan, very clean code.
I would only have added a couple of things but they are not strictly needed:
- Separate the new test in two tests, it is more visible in openQA, we have done in similar fashion with other tests (the only exception is when you pass twice by the same screen). It is similar what happens with expert partitioner test (that it is hard to see what is failing when if fails).
- Perhaps to remove the old test (and from main.pm as well), probably is not used anymore, very yast related, perhaps in functional?

PR #14014, opened 2022-01-14
Collect y2logs when error during installation occurs

ge0r:
When the 'An error occured during the installation.' OK popup occurs, sending return (and as a result hitting the OK button) will result in all ttys logging out.
This results in post_fail_hooks failing and logs not being saved, because any console accessed from that point onward has already been logged out.
So sending return should be avoided in the case of the installation error OK popup, in order for the post_fail_hooks to not fail.

- Related ticket: https://progress.opensuse.org/issues/102662
- Needles: No needles
- Verification run: http://10.163.41.154/tests/3846#downloads (the logs have been uploaded, including y2logs)

Notes: 
* [video](https://w3.suse.de/~ggkioulis/return_key_logs_out.mkv) where hitting return on installation error popup logs out all ttys
* this PR depends on [Do not die when uploading the widgets json](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14022#pullrequestreview-855366170)

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1502,7 +1502,11 @@ sub show_tasks_in_blocked_state {
         send_key 'alt-sysrq-w';
         # info will be sent to serial tty
         wait_serial(qr/sysrq\s*:\s+show\s+blocked\s+state/i);
-        send_key 'ret';    # ensure clean shell prompt
+
+        # If the 'An error occured during the installation.' OK popup has popped up,
+        # do not press the 'return' key, because it will result in all ttys logging out.
+        send_key 'ret' unless (check_screen 'linuxrc-install-fail');
+
     }
 }
 


JRivrain:
Why not just save_y2logs unconditionally ? It's a post_fail_hook, so it should try to gather logs anytime something failed no ?

lib/y2_installbase.pm
```diff
@@ -550,7 +550,12 @@ sub save_remote_upload_y2logs {
 sub post_fail_hook {
     my $self = shift;
 
-    if (check_var("REMOTE_CONTROLLER", "ssh") || check_var("REMOTE_CONTROLLER", "vnc")) {
+    if (check_screen 'error_during_installation') {
```
jknphy, line 553: it would be more clear something like `linuxrc-installation-error`, but inside of `get_to_console` looks like it is already named as `linuxrc-install-fail`

lib/y2_installbase.pm
```diff
@@ -550,7 +550,12 @@ sub save_remote_upload_y2logs {
...
+        # In case of error during installation collect y2logs right away
+        get_to_console;
+        $self->save_upload_y2logs;
```
jknphy, line 556: but this flow only save y2logs, we should get the rest of the stuff as well, right? does it work moving `get_to_console()` before SUPER?

jknphy:
good discovery!

PR #14016, opened 2022-01-15
wicked: Fix wlan testsuite with custom RPM package

cfconrad:
The wicked::wlan::before_test() calls wicked::wlan::prepare_phys() which create
the network namespace and load the hwsim module. But these are not boot
save, so a reboot remove these settings...

- Verification run: http://openqa.wicked.suse.de/t46777

### DIFF
--- a/lib/wicked/wlan.pm
+++ b/lib/wicked/wlan.pm
@@ -162,16 +162,14 @@ sub stop_dhcp_server {
     assert_script_run(sprintf('test -e %s && kill $(cat %s) || true', $pidfile, $pidfile));
 }
 
-sub before_test {
+sub prepare_sut {
     my $self = shift // wicked::wlan->new();
-    $self->prepare_packages();
     $self->prepare_phys();
     $self->prepare_freeradius_server();
     $self->adopt_apparmor();
 }
 
 sub prepare_packages {
-    my $self = shift;
     if (is_sle()) {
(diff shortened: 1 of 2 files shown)


tests/wicked/before_test.pm
```diff
@@ -133,16 +133,16 @@ sub run {
             zypper_ar($repo_url . generate_version('_') . '/', name => 'wicked_maintainers', no_gpg_check => 1, priority => 60);
             $package_list .= ' ndisc6';
         }
-        if (check_var('WICKED', 'wlan')) {
-            wicked::wlan::before_test();
-        }
+        wicked::wlan::before_test() if (check_var('WICKED', 'wlan'));
```
asmorodskyi, line 136:
```suggestion
        wicked::wlan::prepare_wlan_packages() if (check_var('WICKED', 'wlan'));
```

asmorodskyi: with deletion of `before_test` function you don't need it anymore

lib/wicked/wlan.pm
```diff
@@ -165,6 +165,10 @@ sub stop_dhcp_server {
 sub before_test {
     my $self = shift // wicked::wlan->new();
     $self->prepare_packages();
+}
+
+sub before_test_after_reboot {
```
asmorodskyi, line 170:
```suggestion
sub after_reboot {
```

this before/after mix is a bit confusing

tests/wicked/before_test.pm
cfconrad: ok

PR #14017, opened 2022-01-16
Add configuration for tidyall

okurz:
A complete initial run when the cache is empty takes 1m34s for me. The old
implementation of tidy takes 1m30s so the cache generation can take roughly 4s.
However any subsequent run is very quick. A no-op run takes only 1.5s.

This is a preparation for an upcoming planned change in os-autoinst using
tidyall as well which will automatically allow the use in
os-autoinst-distri-opensuse. For this see
https://github.com/os-autoinst/os-autoinst/pull/1795

Related progress issue: https://progress.opensuse.org/issues/99678

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/*.html
 docs/**/*.html
 /script/yaml_generator/env/
 __pycache__/
+.tidyall.d/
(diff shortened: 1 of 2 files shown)


tjyrinki (approved):
Looks and sounds like welcome.

okurz:
Awesome, thank you!

PR #14020, opened 2022-01-17
Fix upgrade_select for new popup dialog

jknphy:
Fix shortcut according to [new popup developed](https://github.com/yast/yast-storage-ng/pull/1271). Provide branching for SLE and Leap as we don't know exactly when this change will apply there, so it can be edit there.

- Related tickets: [poo#104926](https://progress.opensuse.org/issues/104926) 
- Verification run: [openSUSE](https://openqa.opensuse.org/tests/overview?distri=opensuse&build=jknphy%2Fos-autoinst-distri-opensuse%23fix_upgrade_select)

### DIFF
--- a/tests/installation/upgrade_select.pm
+++ b/tests/installation/upgrade_select.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright 2009-2013 Bernhard M. Wiedemann
-# Copyright 2012-2020 SUSE LLC
+# Copyright 2012-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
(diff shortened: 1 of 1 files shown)


tests/installation/upgrade_select.pm
```diff
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright 2009-2013 Bernhard M. Wiedemann
-# Copyright 2012-2020 SUSE LLC
+# Copyright 2012-2021 SUSE LLC
```
DimStar77, line 4: We are actually already in 2022 :)

DimStar77 (approved):

PR #14027, opened 2022-01-17
script_run(): Fail on timeout by default

mdoucha:
OpenQA API has added a new script_run() parameter to control how the function handles command timeout. Set the default timeout behavior to fail the test module.

This can obviously break things. I'll announce this PR via QE mailing list and leave it open for 3 weeks so that everybody has time to fix their test code.

- Related ticket: N/A
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/7985939 (cloned from https://openqa.suse.de/tests/7969815)

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -36,6 +36,18 @@ use testapi qw(send_key %cmd assert_screen check_screen check_var click_lastmatc
   wait_still_screen wait_screen_change get_required_var diag);
 
 
+=head2 new
+
+Class constructor
+=cut
+sub new {
+    my ($class) = @_;
+    my $self = $class->SUPER::new(@_);
+
+    $self->{script_run_die_on_timeout} = 1;
+    return $self;
+}
+
 =head2 handle_password_prompt
 
 Types the password in a password prompt


mdoucha:
The deprecation warnings in autoinst-log.txt have disappeared in the verification run as expected. The PR is ready for merge as is (3 weeks from now).

mdoucha:
Note: This is a follow-up to https://github.com/os-autoinst/os-autoinst/pull/1807

asmorodskyi:
generally setting default to 1 is right thing to do . I just not getting why not simply do this in https://github.com/os-autoinst/os-autoinst/blob/master/distribution.pm#L18 ?

foursixnine:
@mdoucha mind keeping the default behavior on `post_fail_hooks`?
```
$testapi::distri->{script_run_die_on_timeout}  = -1
```

mdoucha:
> @mdoucha mind keeping the default behavior on `post_fail_hooks` sub post_fail_hook

Is it worth it? The kind of issues which the old behavior would "solve" need a different approach anyway so this PR will just highlight the post-fail hook parts which are fragile and need fixing.

b10n1k:
> @mdoucha mind keeping the default behavior on `post_fail_hooks` sub post_fail_hook
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/9804236698a92a4aa2dbe9fe2217afb48c36c75c/lib/opensusebasetest.pm#L1303
> 
> ?
> ```
> $testapi::distri->{script_run_die_on_timeout}  = -1
> ```

is this necessary? if it has to change to not fail, it should be 0, right? In any case, it would be easier to override the `post_fail_hooks` in another subclass or in some module individual. Nevertheless i rather think how to solve this later.

lib/susedistribution.pm
```diff
@@ -36,6 +36,18 @@ use testapi qw(send_key %cmd assert_screen check_screen check_var click_lastmatc
...
 
+=head2 new
+
+Class constructor
+=cut
+sub new {
+    my ($class) = @_;
+    my $self = $class->SUPER::new(@_);
+
+    $self->{script_run_die_on_timeout} = 1;
```
coolgw, line 47: If i understand correct , should we set this default to 0 to keep original behavior and avoid change legacy code.

tinawang123: Yes, I think so. It's better keep original behavior. If anyone need this, he/she can set to '1'

mdoucha: The only sane default is to throw error when the `script_run()` command times out. There's no point running another command on the SUT when the console is still blocked by the previous one. If you have some fancy error handling that'll deal with the timeout, the correct way to preserve the old behavior is this: `script_run(.., die_on_timeout => 0);`

dzedro:
> This can obviously break things. I'll announce this PR via QE mailing list and leave it open for 3 weeks so that everybody has time to fix their test code.

If this is merged before script_run is fixed then **everyhting** will be broken.
Just lib/utils uses 26 unfixed script_runs.
Letting everybody know won't help if at least one is not working on openQA anymore etc.

```
# grep "\sscript_run[ (]" lib/utils.pm 
    return if script_run 'showconsolefont';
        $ret = script_run("zypper -n $command $printer; ( exit \${PIPESTATUS[0]} )", $timeout);
            script_run('tac /var/log/zypper.log | grep -F -m1 -B10000 "Hi, me zypper" | tac | grep \'\(SolverRequester.cc\|THROW\|CAUGHT\)\' > /tmp/z104.txt');
            script_run('tac /var/log/zypper.log | grep -F -m1 -B10000 "Hi, me zypper" | tac | grep \'Exception.cc\' > /tmp/zlog.txt');
    my $ret = script_run("ssh $remote 'sudo zypper -n patch --with-interactive -l'", 1500);
    $ret = script_run("ssh $remote 'sudo zypper -n patch --with-interactive -l'", 6000);
    script_run("export ZYPP_LOCK_TIMEOUT='$timeout'");
    script_run("systemctl mask packagekit; systemctl stop packagekit; while pgrep packagekitd; do sleep 1; done");
    script_run("systemctl mask packagekit; systemctl stop packagekit; while pgrep packagekitd; do sleep 1; done");
        script_run "cat /run/initramfs/rdsosreport.txt > /dev/$serialdev";
        script_run "echo \"\n--------------Beginning of journalctl--------------\n\" > /dev/$serialdev";
        script_run "journalctl --no-pager -o short-precise > /dev/$serialdev";
        script_run "journalctl --no-pager -o short-precise > /dev/$serialdev";
over reboots. Used to ensure that testapi calls like script_run work for the
    return if script_run("iptables -S | grep 'A input_ext.*tcp.*dport 59.*-j ACCEPT'", 30) == 0;
        # timeout for script_run must be larger than for the 'timeout ...' command
        $ret = script_run($exec, ($timeout + 3));
        script_run("(script -qe -a /dev/null -c \'", 0);
        script_run($cmd, 0);
        script_run("\'; echo $endmark\$?) |& tee /dev/$serialdev", 0);
    script_run("cat $filename") if $debug;
    script_run("(zypper patch --with-interactive -l;echo ZYPPER-DONE) | tee /dev/$serialdev", 0);
    script_run('iptables -S', timeout => $timeout);
        my $results = script_run("grep 'PermitRootLogin yes' /etc/ssh/sshd_config");
however, assert_script_run and script_run is not stable in sole console
    my $ret = script_run("snapper --help | grep disable-used-space");
```

Vogtinator:
> If this is merged before script_run is fixed then **everyhting** will be broken

How?

dzedro:
> > If this is merged before script_run is fixed then **everyhting** will be broken
> 
> How?

Sorry, ignore me, didn't read the code, thought something else is going to change. :grimacing:

mdoucha:
I've announced the changes 3 weeks ago and sent a reminder last week. Let's merge it then.

michaelgrifalconi:
Hello @mdoucha , I guess wrappers for script_run(), like [script_retry()](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/0777a7cbe2f9f716e0a059dfa602d2279481d766/lib/utils.pm#L1589-L1608) have no way to switch back to the old behavior as long as they do not pass over the argument.

Any suggestion on how to handle the timeout here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/krb5.pm#L26 ? Thanks

cfconrad:
@michaelgrifalconi I think we mix here something which doesn't belong to each other.
`script_retry()` just don't have a explicit handling for `script_run()` timeout behavior. And if `script_run()` times out in `script_retry()`, it just execute the command again, depending on retry. From my point of view this was just wrong and is now fixed with the default behavior of `die_on_timeout`.

The reason why `script_retry()` works is the `timeout` command which is placed in front of the command. So we always get back the shell. Maybe this `$timeout +3 ` is not enough as `timeout` needs some more time to kill your command? Or there is a command which cannot be killed by `timeout`?!

Do you encounter problems, could you link to them?

cfconrad:
Ok sorry, didn't realized the link to your PR just commented there: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14221#issuecomment-1034970700

And if you wanna have the old/broken behavior, then you need to change `script_retry()` and add the `die_on_timeout => 0` inside there.

PR #14030, opened 2022-01-18
Create IRQ balance test for xen guest

Julie-CAO:
The test aims to check if the NIC IRQs are balanced in xen guests.

test scenario:

1. set up fv & pv guests with at least 4 vcpu core.
2. the irqbalance service should be enabled in guests by default. if not, start the service.
3. check if the IRQs are balanced in guest.
4. try to bind all NIC IRQs to cpu0. There are difference between sLE15SP4 and priorer versions, binding to CPU0 is not expected to be successful with irqbalance.service enabled on SLE15SP4, while on  SLE15SP3/SLE12SP5, it was successful. Perfoming the irqbalance once again make the CPU affinities distributed again(be reverted to their original values).
5. Use ping instead of downloading a file to generate NIC IRQs. ping is easire and more stable, and it the recommended way in official document(https://www.kernel.org/doc/Documentation/IRQ-affinity.txt).
6. check if the new increased IRQs are balanced in guest. because the IRQ list are accumulated, the IRQs after downloads minus the IRQs before downloads should be distributed among all CPUs, ie. the IRQs on each CPU core should not be zero.

Related ticket: https://progress.opensuse.org/issues/93459
Verification run:
[uefi-gi-guest_developing-on-host_sles15sp3-xen](http://openqa.nue.suse.com/tests/8009060)
[gi-guest_sles15sp3-on-host_developing-xen](http://openqa.nue.suse.com/tests/8009059)
[gi-guest_sles12sp5-on-host_developing-xen](http://openqa.nue.suse.com/tests/8009061)

### DIFF
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -23,7 +23,7 @@ use Net::IP qw(:PROC);
 use File::Basename;
 use Utils::Architectures;
 
-our @EXPORT = qw(is_vmware_virtualization is_hyperv_virtualization is_fv_guest is_pv_guest is_guest_ballooned is_xen_host is_kvm_host
+our @EXPORT = qw(is_vmware_virtualization is_hyperv_virtualization is_fv_guest is_pv_guest guest_is_sle is_guest_ballooned is_xen_host is_kvm_host
   check_host check_guest print_cmd_output_to_file ssh_setup ssh_copy_id create_guest import_guest install_default_packages
   upload_y2logs ensure_default_net_is_active ensure_guest_started ensure_online add_guest_to_hosts restart_libvirtd remove_additional_disks
   remove_additional_nic collect_virt_system_logs shutdown_guests wait_guest_online start_guests is_guest_online wait_guests_shutdown
@@ -63,6 +63,23 @@ sub is_pv_guest {
     my $guest = shift;
     return $guest =~ /\bpv\b/;
 }
+
+#Check if guest is SLE with optional filter for:
+#Version: <=12-sp3 =12-sp1 >11-sp1 >=15 15+ (>=15 and 15+ are equivalent)
+#usage: guest_is_sle($guest_name, '<=12-sp2')
+sub guest_is_sle {
+    my $guest_name = lc shift;
+    my $query = shift;
+
+    return 0 unless $guest_name =~ /sle/;
+    return 1 unless $query;
+
+    # Version check
+    $guest_name =~ /sles-*(\d{2})(?:-*sp(\d))?/;
+    my $version = $2 eq '' ? "$1-sp0" : "$1-sp$2";
+    return check_version($query, $version, qr/\d{2}(?:-sp\d)?/);
+}
+
 
 #return 1 if max_mem > memory in vm configuration file in libvirt
 sub is_guest_ballooned {
@@ -105,7 +122,9 @@ sub check_guest {
 }
 
 #ammend the output of the command to an existing log file
-#passing guest name or an remote IP as the 3rd parameter if running command in a remote machine
+#$machine=<guest_ip> to pass guest name or an remote IP if running command in a remote machine
+#default $machine is the current SUT, ie. the host.
+#make sure '$file' is present in current SUT(host), it wastes time to check the file in each call.
 #only support simple bash command so far, eg. '|' is not supported in $cmd.
 sub print_cmd_output_to_file {
     my ($cmd, $file, $machine) = @_;
--- a/tests/virt_autotest/xen_guest_irqbalance.pm
+++ b/tests/virt_autotest/xen_guest_irqbalance.pm
@@ -1,0 +1,218 @@
+# SUSE's openQA tests
+
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary: xen domU irqbalance test
+# Maintainer: Julie CAO <JCao@suse.com>
+
+use base "virt_feature_test_base";
+use strict;
+use warnings;
+use testapi;
+use version_utils qw(is_sle);
+use set_config_as_glue;
+use virt_autotest::common;
+use virt_autotest::utils qw(is_kvm_host guest_is_sle wait_guest_online print_cmd_output_to_file);
+use virt_utils qw(upload_virt_logs remove_vm restore_downloaded_guests);
+
+our $vm_xml_save_dir = "/tmp/download_vm_xml";
+
+sub run_test {
+    my $self = shift;
+
+    return if is_kvm_host;
+
+    save_original_guests();
+
+    foreach my $guest (keys %virt_autotest::common::guests) {
+
+        #irqbalance test run on sle12sp3+ guest
+        next if guest_is_sle($guest, '<=12-sp2');
+
+        record_info("Test $guest");
+        prepare_guest_for_irqbalance($guest);
+
+        # with irqbalance.service is enabled, SMP affinity will be distributed among CPUs automatically even if IRQs are bound to specified CPUs in SLE15SP4
+        assert_script_run "ssh root\@$guest \"systemctl stop irqbalance.service\"" if script_run("ssh root\@$guest \"systemctl status irqbalance\"") eq 0;
+
+        my $nproc = script_output("ssh root\@$guest 'nproc'");
+
+        # Bind all NIC IRQs to cpu0 in guest
+        my @nic_irqs_id = split('\n', script_output("ssh root\@$guest \"grep vif /proc/interrupts | cut -d ':' -f1 | sed 's/^ *//'\""));
+        foreach my $irq_id (@nic_irqs_id) {
+            assert_script_run("ssh root\@$guest \"echo 1 > /proc/irq/$irq_id/smp_affinity\"");
+        }
+        my @affinities_with_binding_cpu0 = get_irq_affinities_from_guest($guest, @nic_irqs_id);
+        foreach (@affinities_with_binding_cpu0) {
+            die "Failed to bind NIC IRQs to CPU0" if $_ ne 1;
+        }
+
+        # Start the irqbalance.service again
+        assert_script_run("ssh root\@$guest \"systemctl start irqbalance\"");
+
+        # Check if the SMP affinities are distributed among CPUs
+        my @affinities_with_irqbalance = get_irq_affinities_from_guest($guest, @nic_irqs_id);
+        # for SLE15SP3 and prior, IRQs are distributed on one cpu core ramdomly.
+        # So it is correct if some of them may happen to run on cpu0
+        # but it would not be correct if all IRQs were bound to cpu0.
+        if (guest_is_sle($guest, '<=15-sp3')) {
+            my $multiply = 1;
+            $multiply *= $_ foreach (@affinities_with_irqbalance);
+            die "SMP affinities were all bound to CPU0 with irqbalance.service enabled" if $multiply == 1;
+        }
+        # for SLE15SP4 (or newer release), IRQs are specified on all CPU cores
+        # fg. affinity value is '0xf' for machine with 4 CPU cores
+        else {
+            my $default_affinity = script_output("ssh root\@$guest \"cat /proc/irq/default_smp_affinity\"");
+            foreach (@affinities_with_irqbalance) {
+                die "The NIC IRQs did not follow the default_smp_affinity with irqbalance enabled." if $_ ne $default_affinity;
+            }
+        }
+
+        # Get the network interrupts distribution over CPUs in guest prior to network loads
+        my @initial_total_irqs_on_cpu = get_nic_irqs_distribution_from_guest($guest, $nproc);
+
+        # Generate network load in guest
+        my $ping_url = "download.opensuse.org";
+        assert_script_run("ssh root\@$guest \"ping -f -c 500 $ping_url\"");
+
+        # re-caculate the network interrupts distribution
+        # the increased NIC IRQs should be distributed in CPU cores in balance
+        my @total_irqs_on_cpu_after_network_download = get_nic_irqs_distribution_from_guest($guest, $nproc);
+        my @increased_irqs_on_cpu;
+        for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
+            $increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id];
+            die("The value of network interrupts from download for CPU" . $cpu_id . " is 0") if $increased_irqs_on_cpu[$cpu_id] == 0;
+        }
+        record_info("NIC IRQs distribution of 'ping $ping_url'", "@increased_irqs_on_cpu");
+    }
+
+    restore_xml_changed_guests();
+
+}
+
+#save the guest configuration files into a folder
+sub save_original_guests {
+    assert_script_run "mkdir -p $vm_xml_save_dir" unless script_run("ls $vm_xml_save_dir") == 0;
+    my $changed_xml_dir = "$vm_xml_save_dir/changed_xml";
+    script_run("[ -d $changed_xml_dir ] && rm -rf $changed_xml_dir/*");
+    script_run("mkdir -p $changed_xml_dir");
+    foreach my $guest (keys %virt_autotest::common::guests) {
+        unless (script_run("ls $vm_xml_save_dir/$guest.xml") == 0) {
+            assert_script_run "virsh dumpxml --inactive $guest > $vm_xml_save_dir/$guest.xml";
+        }
+    }
+}
+
+#restore guest from the configuration files in a folder
+sub restore_original_guests {
+    foreach my $guest (keys %virt_autotest::common::guests) {
+        remove_vm($guest);
+        if (script_run("ls $vm_xml_save_dir/$guest.xml") == 0) {
+            restore_downloaded_guests($guest, $vm_xml_save_dir);
+        }
+        else {
+            record_soft_failure "Fail to restore $guest!";
+        }
+    }
+}
+
+#restore guest which xml configuration files were changed in prepare_guest_for_irqbalance()
+sub restore_xml_changed_guests {
+    my $changed_xml_dir = "$vm_xml_save_dir/changed_xml";
+    my @changed_guests = split('\n', script_output("ls -1 $changed_xml_dir | cut -d '.' -f1"));
+    foreach my $guest (@changed_guests) {
+        remove_vm($guest);
+        restore_downloaded_guests($guest, $changed_xml_dir);
+    }
+}
+
+#set up guest test environment to run irqbalance test
+sub prepare_guest_for_irqbalance {
+    my $vm_name = shift;
+
+    #4 or more vcpu is needed
+    my $nproc = script_output "virsh vcpucount --config --current $vm_name";
+    if ($nproc < 4) {
+        my $changed_xml_dir = "$vm_xml_save_dir/changed_xml";
+        assert_script_run "virsh dumpxml --inactive $vm_name > $vm_name.xml";
+        #save for restore the changed guests after test
+        assert_script_run "cp $vm_name.xml $changed_xml_dir/$vm_name.xml";
+        script_run "virsh destroy $vm_name";
+        assert_script_run "xmlstarlet edit -L -d /domain/vcpu $vm_name.xml";
+        assert_script_run "xmlstarlet edit -L \\
+                               -s /domain -t elem -n vcpu -v 4 \\
+			       -s /domain/vcpu -t attr -n placement -v static \\
+			       $vm_name.xml";
+        script_run "virsh undefine $vm_name" unless (script_run "virsh undefine --nvram $vm_name") == 0;
+        assert_script_run " ! virsh list --all | grep $vm_name";
+        assert_script_run "virsh define $vm_name.xml";
+        assert_script_run "virsh start $vm_name";
+    }
+
+    wait_guest_online($vm_name);
+    assert_script_run "ssh root\@$vm_name \"zypper in -y irqbalance\"" unless script_run("ssh root\@$vm_name \"rpm -q irqbalance\"") eq 0;
+
+}
+
+# return the total NIC IRQs on each CPU core from guest
+sub get_nic_irqs_distribution_from_guest {
+    my ($vm_name, $cpu_account) = @_;
+
+    my @total_irqs_on_cpu = ();
+    my $irq_output = script_output("ssh root\@$vm_name \"grep vif0 /proc/interrupts\"");
+    record_info("NIC IRQs distribution", $irq_output);
+    for (my $cpu_id = 0; $cpu_id < $cpu_account; $cpu_id++) {
+        my @irqs_on_one_cpu = split('\n', script_output("ssh root\@$vm_name \"awk '/vif0/{ print \\\$(($cpu_id+2)) }' /proc/interrupts\""));
+        # sum of IRQs on one cpu core
+        for (my $i = 0; $i <= $#irqs_on_one_cpu; $i++) {
+            $total_irqs_on_cpu[$cpu_id] += $irqs_on_one_cpu[$i];
+        }
+    }
+
+    return @total_irqs_on_cpu;
+}
+
+# it is more stable to get the affinities after a little time than just after it is written
+# especially with irqbalance is enabled, so a function is wrapped
+sub get_irq_affinities_from_guest {
+    my ($vm_name, @irqs) = @_;
+
+    my $cat_affinity_files_cmdline = "ssh root\@$vm_name \"cd /proc/irq && cat";
+    foreach my $irq_id (@irqs) {
+        $cat_affinity_files_cmdline .= " $irq_id/smp_affinity";
+    }
+    $cat_affinity_files_cmdline .= "\"";
+    return split('\n', script_output($cat_affinity_files_cmdline, 60));
+}
+
+sub post_fail_hook {
+    my $self = shift;
+
+    diag("Module xen_guest_irqbalance post fail hook starts.");
+    my $log_dir = "/tmp/irqbalance";
+    foreach my $guest (keys %virt_autotest::common::guests) {
+        my $log_file = $log_dir . "/$guest" . "_irqbalance_debug";
+        script_run("[ -d $log_dir ] && rm -rf $log_dir/*");
+        script_run("mkdir -p $log_dir && touch $log_file");
+        print_cmd_output_to_file("rpm -q irqbalance", $log_file, $guest);
+        print_cmd_output_to_file("systemctl status irqbalance", $log_file, $guest);
+        print_cmd_output_to_file("grep vif0 /proc/interrupts", $log_file, $guest);
+        print_cmd_output_to_file("cat /proc/irq/default_smp_affinity", $log_file, $guest);
+        print_cmd_output_to_file("cat /proc/irq/*/smp_affinity", $log_file, $guest);
+        #skip post the output of 'irqbalance --debug' because it takes too long(more than 15 minutes)
+    }
+    upload_virt_logs($log_dir, "irqbalance_debug");
+    $self->SUPER::post_fail_hook;
+    restore_original_guests();
+
+}
+
+sub test_flags {
+    #continue subsequent test in the case test restored
+    return {fatal => 0};
+}
+
+1;
+
--- a/products/sle/main.pm
+++ b/products/sle/main.pm
@@ -629,6 +629,7 @@ sub load_virt_feature_tests {
         loadtest "virt_autotest/libvirt_routed_virtual_network";
         loadtest "virt_autotest/libvirt_isolated_virtual_network";
     }
+    loadtest "virt_autotest/xen_guest_irqbalance" if get_var("ENABLE_XEN_GUEST_IRQBALANCE");
     loadtest "virt_autotest/sriov_network_card_pci_passthrough" if get_var("ENABLE_SRIOV_NETWORK_CARD_PCI_PASSTHROUGH");
     if (get_var('ENABLE_HOTPLUGGING')) {
         loadtest 'virtualization/universal/hotplugging_guest_preparation';
--- a/tests/virt_autotest/virt_utils.pm
+++ b/tests/virt_autotest/virt_utils.pm
@@ -491,10 +491,11 @@ sub download_guest_assets {
 #Start the guest from the downloaded vm xml and vm disk file
 sub restore_downloaded_guests {
     my ($guest, $vm_xml_dir) = @_;
-    record_info("Guest", "$guest");
+    record_info("Guest restored", "$guest");
     my $vm_xml = "$vm_xml_dir/$guest.xml";
     assert_script_run("virsh define $vm_xml", 30);
 }
+
 
 sub is_installed_equal_upgrade_major_release {
     #get the version that the host is installed to


Julie-CAO:
@alice-suse @waynechen55 @guoxuguang @nanzhg @pdostal and others, welcome review. The PR is [WIP] status as it has only been verified on SLE15SP4 and not on MU, but I'd like it be reviewed earlier in case of any problems on test design or capitability.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,144 @@
+# SUSE's openQA tests
+
+# Copyright 2020 SUSE LLC
```
guoxuguang, line 3: Should be # Copyright 2022 SUSE LLC

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,144 @@
...
+        script_run "virsh undefine $vm_name" unless (script_run "virsh undefine --nvram $vm_name") == 0;
+        assert_script_run " ! virsh list --all | grep $vm_name";
+        assert_script_run "virsh define $vm_name.xml";
+        assert_script_run "virsh start $vm_name";
+    }
+    script_run("virsh dumpxml $vm_name | grep vcpu");  #julie debug
+
+    wait_guest_online($vm_name);
+    unless (script_run("ssh root\@$vm_name \"systemctl status irqbalance\"") == 0) {  #julie: if not running, rc still be 0?
+        assert_script_run "ssh root\@$vm_name \"zypper in irqbalance\"";
```
guoxuguang, line 105: Would better to add -y option. 
`assert_script_run "ssh root\@$vm_name \"zypper in -y irqbalance\""; `

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,144 @@
...
+        record_info("IRQs of downloads distribution", "@increased_irqs_on_cpu");
+    }
+
+}
+
+#set up guest test environment to run irqbalance test
+sub prepare_guest_for_irqbalance {
+    my $vm_name = shift;
+
+    #4 or more vcpu is needed
```
guoxuguang, line 85: why do not use with hotplug vpu here. 
For example:
`virsh setvcpus --domain $guest --count 4 --live`

Julie-CAO: if it was not a guest with cpu=2/4, cpu hotplug would not work.

lib/virt_autotest/utils.pm
```diff
@@ -64,6 +64,23 @@ sub is_pv_guest {
...
+#usage: guest_is_sle($guest_name, '<=12-sp2')
+sub guest_is_sle {
+    my $guest_name = lc shift;
+    my $query = shift;
+
+    return 0 unless $guest_name =~ /sle/;
+    return 1 unless $query;
+
+    # Version check
+    $guest_name =~ /sles-*(\d{2})(?:-*sp(\d))?/;
```
alice-suse, line 78: The guest naming rules are different in MU, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/virt_autotest/common.pm.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+        unless (script_run("ls $downloaded_xml_dir/$guest.xml") == 0) {
+            assert_script_run "mkdir -p $downloaded_xml_dir";
+            assert_script_run "virsh dumpxml --inactive $guest > $downloaded_xml_dir/$guest.xml";
+        }
+    }
+
+    foreach my $guest (keys %virt_autotest::common::guests) {
+
+        #irqbalance test run on sle12sp3+ guest
+        return if guest_is_sle($guest, '<=12-sp2');
```
alice-suse, line 34: Shall the 'return' be 'continue'?

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+
+}
+
+#set up guest test environment to run irqbalance test
+sub prepare_guest_for_irqbalance {
+    my $vm_name = shift;
+
+    #4 or more vcpu is needed
+    my $nproc = script_output "virsh vcpucount $vm_name | grep current | grep config | awk '{ print \$3 }'";
+    if ($nproc < 4) {
```
alice-suse, line 97: IMHO, if nproc > 1, it will be able to test irqbalance. Why do we need it to be 4? Besides, if there is strong reason for this, 'virsh set-vcpu xx 4 --live ' will be easier to satisfy this without worrying about changing the guest config(need restore after irq test)?

alice-suse: After second thought on this, I would recommend to keep the guests with their original configured vcpu number if it is bigger than 1. The benefit is that we can cover different vcpu numbers and save the cost of changing guests and recovering them after test.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+
+        #irqbalance test run on sle12sp3+ guest
+        return if guest_is_sle($guest, '<=12-sp2');
+
+        record_info("Test $guest");
+        prepare_guest_for_irqbalance($guest);
+
+        my $nproc = script_output("ssh root\@$guest 'nproc'");
+
+        #Check if network interrupts are distributed over CPUs in guest initally as irqbalance is enabled by default
```
alice-suse, line 41: Can we add a check about  irqbalance service status? This can help if the default service state changes or by some bug the service can not be enabled automatically.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+        my @initial_total_irqs_on_cpu = get_nic_irqs_distribution_from_guest($guest, $nproc);
+        for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
+            die("The value of network interrupts for CPU" . $cpu_id . " is 0") if ($initial_total_irqs_on_cpu[$cpu_id] == 0);
```
alice-suse, line 44: On extreme conditions this won't be true, eg vcpu numbers is big while interrupts are lower than vcpu number. Suggest to skip this check.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+        # Bind all NIC IRQs to cpu0 in guest
+        my @nic_irqs_id = split('\n', script_output("ssh root\@$guest \"grep vif /proc/interrupts | cut -d ':' -f1 | sed 's/^ *//'\""));
+        my $cat_affinity_files_cmdline = "ssh root\@$guest \"cd /proc/irq && cat";
+        foreach my $irq_id (@nic_irqs_id) {
+            assert_script_run("ssh root\@$guest \"echo 1 > /proc/irq/$irq_id/smp_affinity\"");
+            $cat_affinity_files_cmdline .= " $irq_id/smp_affinity";
+        }
+        $cat_affinity_files_cmdline .= "\"";
+        # with irqbalance.service is enabled, SMP affinity will be auto balanced even if IRQs are bound to specified CPUs in SLE15SP4
+        # while in prior releases, such as SLE15SP3 or SLE12SP5, IRQs can be bound successfully until another irqbalance is run again
```
alice-suse, line 56: what does it mean by "another irqbalance is run again"? Do you mean "restart irqbalance service"?

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: irqbalance can be test with vcpu=2, but it is not a good test condition. because cpu0 always be the most-used one with/without irqbalance, it is hard to determine if the IRQs distributed in all CPUs. I think vcpu=4 or more are much better, but it can't be more then the IRQs generated by the NIC packets.

Regarding ''virsh set-vcpu xx 4 --live ' ', Leon has raised this issue as well. I have concerns about it even if I know it is easier:

1. ''virsh set-vcpu xx 4 --live ' only works with vm with <vcpu current=2>4</vcpu>. for VMs with other vpu configurations, the test will fail. while my test applies to any VMs.
2. cpu hotplug is another test. Involving another test to irqbalance test may introduce unnecessary troubles. the bug from vcpu test may block irqbalance test; besides, when we discribe the reproduce step in the case of an IRQ balance bug, ''virsh set-vcpu xx 4 --live' has to be listed, it is not a necessary step but it may introduce extra effort or explaination. I'd like a test simpler and purer.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: we have had that check in sub prepare_guest_for_irqbalance().

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+                # So it is correct if some of them may happen to run on cpu0
+                # but it would not be correct if all IRQs were bound to cpu0 successfully.
+                if (guest_is_sle($guest, '<=15-sp3')) {
+                    record_soft_failure("One of the NIC IRQs was bound to cpu0, which may be not expected.");
+                }
+                # for SLE15SP4 (or newer release), IRQs are specified on all CPU cores
+                # fg. affinity value is '0xf' for machine with 4 CPU cores
+                else {
+                    die "It is not correct that one of the NIC IRQs was bound to cpu0 successfully with irqbalance enabled.";
+                }
```
alice-suse, line 71: Sorry, I am confused here. What's the logic here and what are the expected results?

lib/virt_autotest/utils.pm
Julie-CAO: Yes, I consider the MU guest names when I created the rule. I verified the function can determine 'sles12sp3PV', 'sles15PV', 'sles11sp4PVx32' and 'sles-15-SP4-64'. But more tests are still needed on MU.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+
+        # Generate xen guest network load
+        assert_script_run("ssh root\@$guest \"ping -f -c 500 download.opensuse.org\"");
```
alice-suse, line 76: Is 500 big enough to ensure all vcpus can get interrupts always?

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: If we skip this check, we will missing the check if irqbalance works by default. because after NIC pings, ''irqbalance --oneshot' is excuted in guests. one feasible way is to generate more NIC irqs before we do this check, what do you think?

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,159 @@
...
+    foreach my $guest (keys %virt_autotest::common::guests) {
+        script_run("ssh root\@$_ \"irqbalance -d > o\"");
+        my $log_file = $log_dir . "/$guest" . "_irqbalance_debug";
+        print_cmd_output_to_file("grep vif0 /proc/interrupts", $log_file, $guest);
+        print_cmd_output_to_file("cat /proc/irq/smp_affinity", $log_file, $guest);
+        print_cmd_output_to_file("irqbalance --debug", $log_file, $guest);
+    }
+    upload_virt_logs($log_dir, "irqbalance_debug");
+    $self->SUPER::post_fail_hook;
+    restore_original_guests();
```
alice-suse, line 149: This line is needed for successful call also.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: the service or the tool is run again, fg. "irqbalance --oneshot". I just change the comment to "until irqbalance is executed again" to make it more clear.

alice-suse:
@Julie-CAO Thanks for the quick PR.
> The test aims to check if the NIC IRQs are balanced in xen guests.
> 
> test scenario:
> 
> 1. set up fv & pv guests with at least 4 vcpu core.

We'd better keep the diversity of vcpus, eg 2, 4, 8, just as the guests are. Reasons given in code comments. 

> 2. the irqbalance service should be enabled in guests by default. if not, start the service.

Nice. I did not get it in code. Do I miss it ?

> 3. check if the IRQs are balanced in guest.
> 4. try to bind all NIC IRQs to cpu0. There are difference between sLE15SP4 and priorer versions, binding to CPU0 is not expected to be successful with irqbalance.service enabled on SLE15SP4

Shall we disable irqbalance before binding, then start irqbalance again, and verifies the binding and distribution? So does <=15sp4 guest.

>, while on  SLE15SP3/SLE12SP5, it was successful. Perfoming the irqbalance once again make the CPU affinities distributed again(be reverted to their original values).
> 5. Use ping instead of downloading a file to generate NIC IRQs. ping is easire and more stable, and it the recommended way in official document(https://www.kernel.org/doc/Documentation/IRQ-affinity.txt).

Nice catch!


> 6. check if the new increased IRQs are balanced in guest. because the IRQ list are accumulated, the IRQs after downloads minus the IRQs before downloads should be distributed among all CPUs, ie. the IRQs on each CPU core should not be zero.
>

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: the actural IRQs of the 500 pings  on 4 CPUs are "330 169 1097 1204", so I think it is ok for 4-100 CPUs.

Julie-CAO: the actural IRQs of the 500 pings  on 4 CPUs are "330 169 1097 1204", so I think it is ok for 4-100 CPUs.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: IMHO, the latter designed test flow can verify if irqbalance works when the service is enabled. Here if we want to have "default" tested, maybe checking if the service is running is enough? BTW, this script is better to be designed open/workable for any guest that comes to be tested, so it can face many different guest configuration&situations, possible to even have the service disabled by autoyast during guest installation.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: I see, reasonable too. Lucky that most guests are configured to have >=4 vcpus. 
BTW, since this guest vcpu confiig change is only done for guest with <4 vcpu. Then when multiple guests are tested with a mixure of > and < 4 vcpus, can restore_original_guests survive?

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: fg. the affinities with 4 cpus with irqbalance may be(take 4 NIC IRQs as an example):
before binding: 
```
'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3' for 15sp4(https://openqa.nue.suse.com/tests/8009060#step/xen_guest_irqbalance/136); 
'cpu1', 'cpu3', 'cpu0', 'cpu2' for 15sp3(https://openqa.nue.suse.com/tests/8009059#step/xen_guest_irqbalance/282); 
'cpu8', 'cpu2', 'cpu1', 'cpu3' for 12sp5(https://openqa.nue.suse.com/tests/8009061#step/xen_guest_irqbalance/288); 
```

after binding to cpu0 WITHOUT 'irqbalance --oneshot' run:
```
'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3', 'cpu0 cpu1 cpu2 cpu3' for 15sp4; 
'cpu0', 'cpu0', 'cpu0', 'cpu0' for 15sp3; 
'cpu0', 'cpu0', 'cpu0', 'cpu0' for 12sp5
```

after binding to cpu0 WITH 'irqbalance --oneshot' run:
```
the same with the values before binding

**Comment by Julie-CAO** [comment]
> @Julie-CAO Thanks for the quick PR. Good job! I only have some minor comments.
> 
> > The test aims to check if the NIC IRQs are balanced in xen guests.
> > test scenario:
> > 
> > 1. set up fv & pv guests with at least 4 vcpu core.
> 
> We'd better keep the diversity of vcpus, eg 2, 4, 8, just as the guests are. Reasons given in code comments.
> 

4, 8 are kept(actually 4+), only 2 are not tested. the reason is given in the reply to your code comments.

> > 2. the irqbalance service should be enabled in guests by default. if not, start the service.
> 
> Nice. I did not get it in code. Do I miss it ?
> 

It is carried out in a sub function.

> > 3. check if the IRQs are balanced in guest.
> > 4. try to bind all NIC IRQs to cpu0. There are difference between sLE15SP4 and priorer versions, binding to CPU0 is not expected to be successful with irqbalance.service enabled on SLE15SP4
> 
> Shall we disable irqbalance before binding, then start irqbalance again, and verifies the binding and distribution? So does <=15sp4 guest.

what's the purpose that we disable irqbalance before binding? to make sure binding is successful? the affinity values tell it as the test does.

File: tests/virt_autotest/xen_guest_irqbalance.pm (hunk 6 above)
Comment (Julie-CAO) [reply tests/virt_autotest/xen_guest_irqbalance.pm:97]: Yes, the vm xml files are saved before irqbalance tests. Finally whatever the test fail or succeed, the VMs are re-defined with their original xml files.

**Comment by alice-suse** [comment]
> It is carried out in a sub function.
>

Yep, see it now.


> > > 4. try to bind all NIC IRQs to cpu0. There are difference between sLE15SP4 and priorer versions, binding to CPU0 is not expected to be successful with irqbalance.service enabled on SLE15SP4
> > 
> > 
> > Shall we disable irqbalance before binding, then start irqbalance again, and verifies the binding and distribution? So does <=15sp4 guest.
> 
> what's the purpose that we disable irqbalance before binding? to make sure binding is successful? the affinity values tell it as the test does.

Yes, the purpose is to make sure binding is successful. You should take the suggestion from xen dev Jan, right? If I understand correctly, the key of his suggestion is that only when binding  succeeds, but after irqblance starts, the interrupts are distributed again which obeys the bindings, can show that irqbalance works.

File: tests/virt_autotest/xen_guest_irqbalance.pm (hunk 8 above)
Comment (Julie-CAO) [reply tests/virt_autotest/xen_guest_irqbalance.pm:44]: no, the check after NIC binding is to check the irqbalance tool, 'irqbalance --oneshot'. the tool and service may not be totally equal, but the function should be equal. 

irqbalance.service is set to running before testing is run in my test in sub prepare_guest_for_irqbalance(). so all guests will be ok for irqbalance test.
```
assert_script_run "ssh root\@$vm_name \"zypper in -y irqbalance\"" unless script_run("ssh root\@$vm_name \"rpm -q irqbalance\"") eq 0;
assert_script_run "ssh root\@$vm_name \"systemctl start irqbalance.service\"" unless script_run("ssh root\@$vm_name \"systemctl status irqbalance\"") eq 0;
```

File: tests/virt_autotest/xen_guest_irqbalance.pm (hunk 11 above)
Comment (alice-suse) [reply tests/virt_autotest/xen_guest_irqbalance.pm:76]: okay, we can tune later if it is not enough

**Comment by Julie-CAO** [comment]
> Yes, the purpose is to make sure binding is successful. You should take the suggestion from xen dev Jan, right? If I understand correctly, the key of his suggestion is that only when binding succeeds, but after irqblance starts, the interrupts are distributed again which obeys the bindings, can show that irqbalance works(rather than that the kernel distributes the interrupts).

I did not notice that kernel distribute the IRQs as well. then lets adjust the scenario a little:

- set up fv & pv guests with at least 4 vcpu core.
- the irqbalance service should be enabled in guests by default. if not, start the service.
- stop irqbalance.service.
- bind all NIC IRQs to cpu0 & check if binding is successful by read the smp affinities(all the values should be '1'(cpu0)).
- start irqbalance.service again.
- check smp affinities are changed to be distributed among CPUs.
- check if the IRQs are balanced in guest & get the total IRQs on each CPUs.
- ping a URL to generate NIC IRQs.
- check if the new increased IRQs are balanced in guest.

**Comment by alice-suse** [comment]
> > Yes, the purpose is to make sure binding is successful. You should take the suggestion from xen dev Jan, right? If I understand correctly, the key of his suggestion is that only when binding succeeds, but after irqblance starts, the interrupts are distributed again which obeys the bindings, can show that irqbalance works(rather than that the kernel distributes the interrupts).
> 
> I did not notice that kernel distribute the IRQs as well. then lets adjust the scenario a little:
> 
> * set up fv & pv guests with at least 4 vcpu core and irqbalance installed.
> * stop irqbalance.service in guest if it is running.
> * bind all NIC IRQs to cpu0 & check if binding is successful by reading the smp affinities(all the values should be '1'(cpu0)) in guest.
> * start irqbalance.service again.
> * check smp affinities are changed to be distributed among CPUs.

One minor suggestion for implementing this: check if multiply of  int(smp_affinity value) of all vcpus !=1, it can indicate the distribution?



> * get the total IRQs on each CPUs.
> * ping a URL to generate NIC IRQs.
> * get the total IRQs on each CPUs, check if the new increased IRQs are balanced in guest.

**Comment by Julie-CAO** [comment]
> > * check smp affinities are changed to be distributed among CPUs.
> 
> One minor suggestion for implementing this: check if multiply of int(smp_affinity value) of all vcpus !=1, it can indicate the distribution?

Cool! Not perfect but rather easier.

**Comment by alice-suse** [comment]
@tonyyuan1  @tbaev  Hello guys, this irqbalance test will also be added into MU xen tests soon. Please help to review and be synced with the design and code!

File: lib/virt_autotest/utils.pm [hunk 13]
```diff
@@ -64,6 +64,23 @@ sub is_pv_guest {
...
+    my $version = $2 eq '' ? "$1-sp0" : "$1-sp$2";
+    return check_version($query, $version, qr/\d{2}(?:-sp\d)?/);
+}
```
waynechen55, line 81: sle does not mean sles. If you mean sles, then use sles ?

waynechen55: This functions looks like check_sles_guest_version to me. And I think the most reliable way to check guest version is get its os_release file and checking from there. I think you can consider using check_os_release in version_utils.pm which supports fetching os_release through ssh.

waynechen55: The function name does not reflect what it does. It looks like check-sles-guest-version.

waynechen55: I do not think      $guest_name =~ /sles-*(\d{2})(?:-*sp(\d))?/;    is the optimal way to obtain guest version. Consider using check_os_release in version_utils.pm which supports fetching os_releae file and checking version from it. I think os_relase file is the most reliable source to get version info.

lib/virt_autotest/utils.pm
```diff
@@ -64,6 +64,23 @@ sub is_pv_guest {
...
 
+#Check if guest is SLE with optional filter for:
+#Version: <=12-sp3 =12-sp1 >11-sp1 >=15 15+ (>=15 and 15+ are equivalent)
+#usage: guest_is_sle($guest_name, '<=12-sp2')
+sub guest_is_sle {
+    my $guest_name = lc shift;
+    my $query = shift;
+
+    return 0 unless $guest_name =~ /sle/;
+    return 1 unless $query;
```
waynechen55, line 75: Do returned 0 and 1 represent errors ? If yes, consider choose other values, I do not think they are suitable, especially 0.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+
+    return if is_kvm_host;
+    #save original guest configuration file in case of restore in post_fail_hook()
+    my $downloaded_xml_dir = "/tmp/download_vm_xml";
+    foreach my $guest (keys %virt_autotest::common::guests) {
+        unless (script_run("ls $downloaded_xml_dir/$guest.xml") == 0) {
+            assert_script_run "mkdir -p $downloaded_xml_dir";
+            assert_script_run "virsh dumpxml --inactive $guest > $downloaded_xml_dir/$guest.xml";
+        }
+    }
```
waynechen55, line 30: I think you can make up a new function to perform 'save orginal guest configs'.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+    restore_original_guests();
+
+}
+
+#set up guest test environment to run irqbalance test
+sub prepare_guest_for_irqbalance {
+    my $vm_name = shift;
+
+    #4 or more vcpu is needed
+    my $nproc = script_output "virsh vcpucount $vm_name | grep current | grep config | awk '{ print \$3 }'";
```
waynechen55, line 103: Why not use virsh vcpucount --config --current --domain

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+    record_info("NIC IRQs distribution", $irq_output);
+    for (my $cpu_id = 0; $cpu_id < $cpu_account; $cpu_id++) {
+        my @irqs_on_one_cpu = split('\n', script_output("ssh root\@$vm_name \"awk '/vif0/{ print \\\$(($cpu_id+2)) }' /proc/interrupts\""));
+        # sum of IRQs on one cpu core
+        for (my $i = 0; $i <= $#irqs_on_one_cpu; $i++) {
+            $total_irqs_on_cpu[$cpu_id] += $irqs_on_one_cpu[$i];
+        }
+    }
+
+    return @total_irqs_on_cpu;
```
waynechen55, line 138: I do not think pass array directly is good. Passing reference to it is better.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+}
```
waynechen55, line 139: Does /proc/irq/default_smp_affinity  have any impact on your test

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+        record_info("Test $guest");
+        prepare_guest_for_irqbalance($guest);
+
+        # with irqbalance.service is enabled, SMP affinity will be distributed among CPUs automatically even if IRQs are bound to specified CPUs in SLE15SP4
+        assert_script_run "ssh root\@$guest \"systemctl stop irqbalance.service\"" if script_run("ssh root\@$guest \"systemctl status irqbalance\"") eq 0;
+
+        my $nproc = script_output("ssh root\@$guest 'nproc'");
+
+        # Bind all NIC IRQs to cpu0 in guest
+        my @nic_irqs_id = split('\n', script_output("ssh root\@$guest \"grep vif /proc/interrupts | cut -d ':' -f1 | sed 's/^ *//'\""));
```
waynechen55, line 46: I think grep vif /proc/interrupts | cut -d ' ' -f2  | cut -d ':' -f1 can also solve your problem.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+        foreach my $irq_id (@nic_irqs_id) {
+            assert_script_run("ssh root\@$guest \"echo 1 > /proc/irq/$irq_id/smp_affinity\"");
+        }
+        my @affinities_with_binding_cpu0 = get_irq_affinities_from_guest($guest, @nic_irqs_id);
```
waynechen55, line 50: I do not think passing array directly to function is good. Consider passing its reference.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+        for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
+            $increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id];
+            die("The value of network interrupts from download for CPU" . $cpu_id . " is 0") if $increased_irqs_on_cpu[$cpu_id] == 0;
+        }
+        record_info("IRQs of downloads distribution", "@increased_irqs_on_cpu");
+    }
+
+    restore_original_guests();
+
+}
```
waynechen55, line 96: I think your workflow can be organized better, for example, do everything in the order of set-do-check and group relevant code into respective functions.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+        my @affinities_with_irqbalance = get_irq_affinities_from_guest($guest, @nic_irqs_id);
+        # for SLE15SP3 and prior, IRQs are distributed on one cpu core ramdomly.
+        # So it is correct if some of them may happen to run on cpu0
+        # but it would not be correct if all IRQs were bound to cpu0.
+        if (guest_is_sle($guest, '<=15-sp3')) {
+            my $multiply = 1;
+            $multiply *= $_ foreach (@affinities_with_irqbalance);
+	    print "julie: \$multiply=$multiply of affinities for $guest after irqbalance.\n";
+            die "SMP affinities were all bound to CPU0 with irqbalance.service enabled" if $multiply == 1;
+        }
```
waynechen55, line 68: Why multiply 1

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+        # for SLE15SP4 (or newer release), IRQs are specified on all CPU cores
+        # fg. affinity value is '0xf' for machine with 4 CPU cores
+        else {
+            foreach (@affinities_with_irqbalance) {
+                die "It is not correct that one of the NIC IRQs was bound to cpu0 with irqbalance enabled." if $_ eq 1;
```
waynechen55, line 73: Why checking $_ eq 1 ? It seems that any values other than 0xf are invalid. You'd better check $_ ne 0xf ?

lib/virt_autotest/utils.pm
Julie-CAO: I copied the idea from 'use version_utils qw(is_sle)`. the function 'is_sle' is always used this way:
`if (is_opensuse || (is_sle && is_sut_reg)) {`
`*** if is_sle('<=12-sp4')`
rc=1/0 works this way.

lib/virt_autotest/utils.pm
Julie-CAO: the widely used lib function, is_sle(), works this way, so I just follow the convention to make it well accepted.

Getting the version from guest name instead of checking inside guest is not so accurate but it has many advantages:

- no need to start guest
- quite easi and quicker

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: Yes, it is called in another place as well. I'd better wrap it up.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: I am not sure if there is always a space ahead of the IRQ_numbers.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: The array is quite small. Only about 8 elements in it. but I'll consider to use reference in my later design.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: to avoid all the numbers are 1. The ideal result is that the number of value = 1 should not exceed 1/<cpu_num>*<nic_irq_number>, but it is not absolute. Using multiply = 1 can only ensure that not all are 1

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: I don't think so. the value of /proc/irq/default_smp_affinity varies in SLE releases, but the effect if IRQ balance is similiar. So I guess different balancing rules are used in SLE releases, and the default_smp_affinity is not same accordingly.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,192 @@
...
+
+    return @total_irqs_on_cpu;
+}
+
+# it is more stable to get the affinities after a little time than just after it is written
+# especially with irqbalance is enabled, so a function is wrapped
+sub get_irq_affinities_from_guest {
+    my ($vm_name, @irqs) = @_;
+
+    my $cat_affinity_files_cmdline = "ssh root\@$vm_name \"cd /proc/irq && cat";
```
guoxuguang, line 146: why do not get smp_affinity file via cat tool directly
For example:
`"ssh root\@$vm_name \"cat /proc/irq/"; `

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: vaule=1 means the IRQ is still bound to CPU0, the irqbalance does not work as expected. The judgement is enough.

the balancing rule varies in different SLES. As far as I know, it is usually 'f' in 15SP4, but it can be '1', '2', '00001000' or other values in 15SP3/12SP5. I am afraid that the value is not definitly be 'f' in future release.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: I aim to get the file content of /proc/irq/<NIC_IRQ_NUMBER>/smp_affinity. so the actural command finally is gotten run is `"ssh root@guest_name \"cd /proc/irq && cat 83/smp_affinit 84/smp_affinity 85/smp_affinity ... 92/smp_affinity \""`

tests/virt_autotest/xen_guest_irqbalance.pm
waynechen55: I think 'multiply' just a way to help you get result from multiplying elements in @affinities_with_irqbalance. What you really detect is the mutiplying result of all elements of @affinities_with_irqbalance.

Julie-CAO: a little confused. Do you think 'multiply' is ok or do you have some idea?

waynechen55: I mean what you really want is the mutiplying result of all elements of @affinities_with_irqbalance, you need variable multiple=1 to do that. Right ?

Julie-CAO: yes, exactly.

tests/virt_autotest/xen_guest_irqbalance.pm
waynechen55: If irqbalance works, I think the affinity value should not have 0 in place where is a corresponding cpu. Affinity value should distributed across all cpus. But you also mentioned sles of certan version does not give it a "f" even with irqbalance in place.

tests/virt_autotest/xen_guest_irqbalance.pm
waynechen55: Do you think it is beneficial to output value of default_smp_affinity as well in your test ?

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: yes, affinity distribute to all cpu in sle15sp4, ie. value 'f' which means "cpu0 cpu1 cup2 cup3'. but in older release, the affinity picks up one cpu ramdomly(appears to be), so the value may be '1'(cpu0), '2'(cpu1), '4'(cpu3), '000200'(cpu10)
'

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: yes. agree.

Julie-CAO: the content of the default_smp_affinity from all SLE release is 'f' in the case of 4 cpu

Julie-CAO: done

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: Have wrapped some functions per comments, such as save_original_guests() and restore_original_guests() to make the scenario clearer. The remained blocks are smaller and should not be called twice. Thanks.

Julie-CAO:
verified again:
[gi-guest_sles15sp3-on-host_developing-xen](https://openqa.nue.suse.com/tests/8026293)
[15sp4 + 15sp3 + 12sp5 guest in local openqa](http://10.67.129.54/tests/168)
[failing case to verify post_fail_hook](http://10.67.129.54/tests/169)

If no further comments, the pr is ready for merge.

tonyyuan1:
Pretty cool automation!  
I had a quick look and just one question:
If this automation is going to test maintenance updates, it need add update repos and install updates (in host or sometimes in guest). Where is this automation doing this?

Julie-CAO:
> Pretty cool automation! I had a quick look and just one question: If this automation is going to test maintenance updates, it need add update repos and install updates (in host or sometimes in guest). Where is this automation doing this?

Thanks, Tony. the test is run based on the existing environment, ie. the host and all guests are ready with updates for MU tests, just like the hotplugging or storage tests. The irqbalance test is expected to follow the virtual network tests.

Julie-CAO:
@tonyyuan1, FYI, the irqbalance test will be deployed in MU in our roadmap 7. I will be responsible for this job.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: @Julie-CAO  Sorry for the late comment, I just managed to open github. According to some online material, the relationship between the affinity value and the cpu number is that if we turn the affinity value into binary number, which bit has 1 means which cpu has irq balanced. So if in 15sp4, all cpus are bound, then the affinity value should equal to nproc bits of 1 in binary way. We can use more precise checking here. HDYT?

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,201 @@
...
+use virt_utils qw(upload_virt_logs remove_vm restore_downloaded_guests);
+
+our $vm_xml_save_dir = "/tmp/download_vm_xml";
+
+sub run_test {
+    my $self = shift;
+
+    return if is_kvm_host;
+
+    save_original_guests();
```
alice-suse, line 26: @Julie-CAO  If we only change the guest xml setting  for those with nproc<4, then we can do save_original_guests and restore_original_guests for these changed only. This will be more efficient, right? In qam tests, this will be very preferred, because they have much more guests(~10) in a single job. HDYT?

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,201 @@
...
+            foreach (@affinities_with_irqbalance) {
+                die "It is not correct that one of the NIC IRQs was bound to cpu0 with irqbalance enabled." if $_ eq 1;
+            }
+        }
+
+        # Get the network interrupts distribution over CPUs in guest prior to network loads
+        my @initial_total_irqs_on_cpu = get_nic_irqs_distribution_from_guest($guest, $nproc);
+
+        # Generate network load in guest
+        assert_script_run("ssh root\@$guest \"ping -f -c 500 download.opensuse.org\"");
```
waynechen55, line 76: I suggest you 8.8.8.8 which is always online.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -0,0 +1,201 @@
...
+
+        # re-caculate the network interrupts distribution
+        # the increased NIC IRQs should be distributed in CPU cores in balance
+        my @total_irqs_on_cpu_after_network_download = get_nic_irqs_distribution_from_guest($guest, $nproc);
+        my @increased_irqs_on_cpu;
+        for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
+            $increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id];
+            die("The value of network interrupts from download for CPU" . $cpu_id . " is 0") if $increased_irqs_on_cpu[$cpu_id] == 0;
+        }
```
waynechen55, line 85: I do not think this is accurate or even close to accurate under certain circumstances. I am afraid there is continuous traffic between SUT and worker, so it is not very meaningful to do this judgement "if $increased_irqs_on_cpu[$cpu_id] == 0". I think you'd better give it a larger number based on the volume of traffic generated and do comparison like "if $increased_irqs_on_cpu[$cpu_id] < 50" ?

waynechen55: To be more accurate, I think there might be interrupt queues per device, so you can assign different applications to different queues to separate interrupts.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: yes, that's true. I need create another function to handle the changed guests.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: ping '8.8.8.8' will generate network packets as well? Let me try

Julie-CAO: I did a search, '8.8.8.8' is a free DNS server provided by Google cloud in 2009. Is it a more fit than download.opensuse.org? I need think it over and welcome anyone's suggestions.

alice-suse: download.opensuse.org is meant to be always online too

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: sounds complex, I'd rather do some reseach on the value.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: I did a search, '8.8.8.8' is a free DNS server provided by Google cloud in 2009. Is it a more fit than download.opensuse.org? I need think it over and welcome anyone's suggestions.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: This is irqbalance test. IMHO, it does not matter from which application or where the NIC traffic are from. What matters is that the network traffic is balanced on all cpus. Once the triggered network traffic is large enough, if irqbalance is working well, statiscally, all cpus have large enough possibility( quite close to 99.9...% ) to get at least one network packet handled. That's what we want to verify to ensure irqbalance works -- packets are being distributed.

waynechen55: To be honest, "if $increased_irqs_on_cpu[$cpu_id] == 0" does not reflect the real situation, especially for cpu0, because it always gets traffic.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: the online document I have found are all talking about the 15sp4 situations. if 15sp5/6 or sle-next apply to this rule as well, '1111...' would be the best checking value. We can make this as the convence since you and Wayne both think it is worthy. Let me check '1111...' for 15sp4+ guest, although it is not too easy.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: I read through the test process again, and think about above comment, I think:
1) comparing value with 0 is correct, and not weak.
why? 
$increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id]

total_irqs_on_cpu_after_network_download is got after " bind irq to cpu0 -> restart irqbalance service in guest", and at this time, if the balance service is working, all interrupts are being distributed (not binded on cpu0).
It does not matter whether the interrupts comes from the ping that we intended to trigger, or the backgorund interrupts on test bed. What matters is that these interrupts are being distributed and will be counted in our later check of the diff. I guess this is where you kind of get lost or misunderstand.
Why we need to trigger large enough ping is that we do not know how much test bed background interrupts are here, and we are afraid that it is not enough to LET ALL CPU HAVE AT LEAST ONE INTERRUPT. It is possible to have (100, 200,300,0) with different balancing algorithm. So only when ping+background interrupts on test bed is large enough, statistically, we can ensure that each cpu has at least 1 interrupt.
So $increased_irqs_on_cpu[$cpu_id] == 0 check is not weak.

2) you made a good point about "fair balancing".
Why it is not included in the design is that there are different balancing algorithms, the "fair" is different in each different algorithm. Our goal of this automation is to check the irq is balancing, rather than validating that each balancing algorithm is behaving correctly(50% fair, or 90% fair). So we do not include this check point.

tests/virt_autotest/xen_guest_irqbalance.pm
waynechen55: You can have your own decision.

tests/virt_autotest/xen_guest_irqbalance.pm
waynechen55: I always understand the test logic. As I said, there is no test isolation at all and no deep dive into irqbalance at all either, so the checkpoint is quite weak from this point of view. Actually what is being checked here is not quite relevant to what is generated as long as there is enough volume, so it is really weak. I said it is weak without criticism. @Julie-CAO can have her own decision if she think more delicate checking is not needed.

alice-suse: Well, I did not intend to reinforce my opinions on anyone. What I was doing was to explain the test design and double confirm that at least within our team, we do not see obvious flaws here(actually we also invited the dev team to review the design and they agreed). Based on your reply, it seems there is still gap. @Julie-CAO  @waynechen55  How about we three has a short call for this? We need to deliver the right thing. Discussion is a good way to achieve this.

alice-suse: we agreed that this design can meet our test requirement, so let's keep this simple design.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: Compare to the default_smp_affinity instead of 1 for 15sp4+ guest, which is not perfect too, but it is better.

Julie-CAO:
verification run passed again in my local openqa server. Pr is ready for merge.
[a test case for successful run](http://10.67.129.102/tests/6#step/xen_guest_irqbalance/112)
[a test case for failed run](http://10.67.129.102/tests/5)

alice-suse:
@Julie-CAO  Thanks for the great work! LGTM. Let's wait if others have last comments, otherwise I will merge it at EOB today.

PR #14036, opened 2022-01-18
Add module for docker service check for migration scenarios

pdostal:
- Related ticket: [poo#104775](https://progress.opensuse.org/issues/104775)
- Verification run: [Migration-from-15-SP1-to-15-SP3](http://pdostal-server.suse.cz/tests/13040)

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -21,7 +21,7 @@ use version_utils;
 use Mojo::Util 'trim';
 
 our @EXPORT = qw(test_seccomp basic_container_tests get_vars can_build_sle_base
-  get_docker_version check_runtime_version container_ip registry_url);
+  get_docker_version check_runtime_version container_ip container_route registry_url);
 
 sub test_seccomp {
     my $no_seccomp = script_run('docker info | tee /tmp/docker_info.txt | grep seccomp');
@@ -56,6 +56,13 @@ sub container_ip {
     my $ip = script_output "$runtime inspect $container --format='{{.NetworkSettings.IPAddress}}'";
     record_info "container IP", "$ip";
     return $ip;
+}
+
+sub container_route {
+    my ($container, $runtime) = @_;
+    my $route = script_output "$runtime inspect $container --format='{{.NetworkSettings.Gateway}}'";
+    record_info "container route", "$route";
+    return $route;
 }
 
 sub registry_url {
@@ -131,7 +138,6 @@ sub basic_container_tests {
     my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
     die('zypper inside container timed out') if (!defined($ret));
     if ($ret != 0) {
-        record_info('poo#40958 - curl install failure, try with force-resolution.');
         my $output = script_output("$runtime container exec $container_name zypper in --force-resolution -y -n curl", 600);
         die('error: curl not installed in the container') unless (($output =~ m/Installing: curl.*done/) || ($output =~ m/\'curl\' .* already installed/));
     }
--- a/lib/services/docker.pm
+++ b/lib/services/docker.pm
@@ -1,0 +1,51 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Package for cups service tests
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+package services::docker;
+use base 'opensusebasetest';
+use testapi;
+use utils;
+use strict;
+use warnings;
+use containers::docker;
+use containers::utils qw(basic_container_tests registry_url);
+use containers::container_images 'build_and_run_image';
+
+my $service_type = 'Systemd';
+
+# check service is running and enabled
+sub check_service {
+    common_service_action 'docker', $service_type, 'is-enabled';
+    common_service_action 'docker', $service_type, 'is-active';
+}
+
+# check docker service before and after migration
+# stage is 'before' or 'after' system migration.
+sub full_docker_check {
+    my (%hash) = @_;
+    my ($stage, $type) = ($hash{stage}, $hash{service_type});
+    $service_type = $type;
+    my $engine = containers::docker->new();
+    $engine->init() if ($stage eq 'before');
+
+    check_service();
+
+    # Test the connectivity of Docker containers
+    $engine->check_containers_connectivity();
+
+    # Run basic runtime tests
+    basic_container_tests(runtime => $engine->runtime);
+
+    # Clean container
+    $engine->cleanup_system_host();
+
+    check_service();
+}
+
+1;
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -76,8 +76,17 @@ sub install_docker_when_needed {
                 # We may run openSUSE with DISTRI=sle and openSUSE does not have SUSEConnect
                 activate_containers_module if $host_os =~ 'sles';
 
+                # Temporarly enable LTSS product on LTSS systems where it is not present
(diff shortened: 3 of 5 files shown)


lib/services/docker.pm
```diff
@@ -0,0 +1,53 @@
...
+sub check_service {
+    common_service_action 'docker', $service_type, 'is-enabled';
+    common_service_action 'docker', $service_type, 'is-active';
+}
+
+# check docker service before and after migration
+# stage is 'before' or 'after' system migration.
+sub full_docker_check {
+    my (%hash) = @_;
+    my ($stage, $type) = ($hash{stage}, $hash{service_type});
```
lemon-suse, line 32: The $stage hasn't been used, means do the same test steps before and after migration, right?

pdostal: The `$stage` is used in line 35:
```perl
$engine->init() if ($stage eq 'before');
```

lemon-suse: Ok, so the only different is before migration need init the engine. Thanks.

lib/services/docker.pm
```diff
@@ -0,0 +1,53 @@
...
+    $service_type = $type;
+    my $engine = containers::docker->new();
+    $engine->init() if ($stage eq 'before');
```
b10n1k, line 35: there is a duplication here with line following from https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/edbae20345112c136837ad3ea49ea03f8820d8fe/lib/service_check.pm#L283 and below.

```
zypper_call "in $srv_pkg_name";
common_service_action($srv_proc_name, $service_type, 'enable');
common_service_action($srv_proc_name, $service_type, 'start');
common_service_action($srv_proc_name, $service_type, 'is-active');
```
IIUC `full_docker_check` will be executed before those lines (https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/edbae20345112c136837ad3ea49ea03f8820d8fe/lib/service_check.pm#L280) and pretty much will do the same things.
- install docker
- will check if is-active

pdostal: Well I would like to keep my changes to minimum. I'm still trying to solve earlier solution so let's wait for verification run.

b10n1k: i dont think it makes sense. if `$engine` is not initialized the rest of the commands will fail later. i cant say what the solution would be here and i cant say more without dig into the code and the test. But i think my observation is obvious.

lib/services/docker.pm
```diff
@@ -0,0 +1,53 @@
...
+
+    check_service();
+
+    # Test the connectivity of Docker containers
+    $engine->check_containers_connectivity();
+
+    # Run basic runtime tests
+    basic_container_tests(runtime => $engine->runtime);
+    # Build an image from Dockerfile and run it
+    build_and_run_image(runtime => $engine, dockerfile => 'Dockerfile.python3', base => registry_url('python', '3'));
```
b10n1k, line 45: based on the ticket description _There is NO need to test docker features_. i think it would be better to replace this with new things here. For instance, firewall checks, docker permission, filesystem checks or simple api calls like `docker images` etc

pdostal: @jlausuch @lemon-suse WDYT?

lib/services/docker.pm
pdostal: Yes, we need `$engine` in both stages but we install and prepare docker only once.

lib/services/docker.pm
jlausuch: That's a very good point.

I wrote the ticket thinking that migration tests don't actually test the feature itself, just the service, but I might be wrong.

I see some tests in https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/lib/services
that have a very basic checks in `check_function`, but some of them don't even implement this function...

Your code is definitely testing quite a lot compared to other migration tests, so IMO, I would leave only here `basic_container_tests` and remove `build_and_run_image`.

pdostal: Done.

lib/containers/utils.pm
```diff
@@ -127,19 +134,23 @@ sub basic_container_tests {
...
-    my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
-    die('zypper inside container timed out') if (!defined($ret));
-    if ($ret != 0) {
-        record_info('poo#40958 - curl install failure, try with force-resolution.');
-        my $output = script_output("$runtime container exec $container_name zypper in --force-resolution -y -n curl", 600);
-        die('error: curl not installed in the container') unless (($output =~ m/Installing: curl.*done/) || ($output =~ m/\'curl\' .* already installed/));
-    }
-    assert_script_run("$runtime container commit $container_name tw:saved", 240);
+    if ($runtime == 'docker' && is_sle('=15-SP1')) {
+        record_info('bsc#1190670', 'bsc#1190670 - docker blocks clone3 syscall');
```
jlausuch, line 138: `record_soft_failure` better?

lib/containers/utils.pm
```diff
@@ -127,19 +134,23 @@ sub basic_container_tests {
...
+        assert_script_run("$runtime container commit $container_name tw:saved", 240);
+    } else {
+        # Containers' state can be saved to a docker image
+        my $ret = script_run("$runtime container exec $container_name zypper -n in curl", 600);
+        die('zypper inside container timed out') if (!defined($ret));
+        if ($ret != 0) {
+            my $output = script_output("$runtime container exec $container_name zypper in --force-resolution -y -n curl", 600);
```
ilausuch, line 145: Why not using --force-resolution in line 142?

pdostal: As it's not usually needed. But I'm not changing that.

b10n1k:
@pdostal revise the VR. i found too many warnings and other strange things. i dont know if i miss somethings but things like that looks hidden or will go unnoticed:
- http://pdostal-server.suse.cz/tests/13030#step/install_service/25 
- http://pdostal-server.suse.cz/tests/13030#step/install_service/38
- http://pdostal-server.suse.cz/tests/13030#step/install_service/54
- http://pdostal-server.suse.cz/tests/13030#step/install_service/67
- http://pdostal-server.suse.cz/tests/13030#step/install_service/74 - shouldnt this be `docker images ls`?
- http://pdostal-server.suse.cz/tests/13030/logfile?filename=autoinst-log.txt - Use of uninitialized value + eval traces

pdostal:
I will look at the `iptables` issue - the rest is not related to this task.

pdostal:
@lemon-suse can you please review? For properly installed Docker on LTSS system, the LTSS product has to be enabled, so I:
 1) Check if the Docker is not available and if not, then:
 2) Check if LTSS registration code if available and if LTSS producf is not installed on the SUT, then:
 3) Enable the LTSS product, install Docker and disable the LTSS product (which I suppose is needed for migration)
 
 Please correct me if I am wrong.

pdostal:
New [verification run](http://pdostal-server.suse.cz/tests/13040) is available.

lemon-suse (approved):

lemon-suse:
> @lemon-suse can you please review? For properly installed Docker on LTSS system, the LTSS product has to be enabled, so I:
> 
> 1. Check if the Docker is not available and if not, then:
> 2. Check if LTSS registration code if available and if LTSS producf is not installed on the SUT, then:
> 3. Enable the LTSS product, install Docker and disable the LTSS product (which I suppose is needed for migration)
> 
> Please correct me if I am wrong.

Yes, I think there is no problem for logic. In fact, we will enable ltss on patch_sle and disable it after update on base system. Install_service will be after patch_sle, so need do the operation for ltss itself. Thanks.

jlausuch (approved):

PR #14047, opened 2022-01-19
Adjust minor log issues on ibtest

b10n1k:
The changes in this PR removes some warnings in the autoinst-log.txt.
For example `Use of uninitialized value` when _record_info_ is missing
description or when it compares with undef.


lib/utils.pm
```diff
@@ -632,7 +632,7 @@ Examples:
 sub zypper_ar {
     my ($url, %args) = @_;
     my $name = $args{name} // '';
-    my $priority = $args{priority} // undef;
+    my $priority = $args{priority} // '';
```
pevik, line 635: Why this? What it improves?


Also it changes meaning of:
```
my $prioarg = defined($priority) && !is_sle('<=12') ? "-p $priority" : "";
```
It'd have to be changed to
```
my $prioarg = $priority && !is_sle('<=12') ? "-p $priority" : "";
```
But I'd prefer it to stay, if it's only a style change.
Also it's an unrelated change, it should be changed in separate commit (in case it needs to be reverted).

pevik: OK, I see it got broken in a1badaffb8c1c571015e9d055abfac39e4c22d4c. Could you please change it as
```
-    my $cmd_mr = "mr -p $priority $url";
+    my $cmd_mr = "mr $prioarg $url";
```
Put it into separate commit and add `Fixes: a1badaffb ("utils::zypper_ar(): Fix repo priority on SLE-11")` into commit message.

tests/kernel/ibtests.pm
```diff
@@ -67,8 +67,8 @@ sub ibtest_master {
     $args = $args . '--in-vm ' if get_var('IBTEST_IN_VM');
     $args = $args . '--no-mad ' if get_var('IBTEST_NO_MAD');
 
-    if ($phase != '') {
-        $args = $args . "--phase $phase";
+    if ($phase ne '') {
+        $args = $args . "--phase $phase ";
```
pevik, line 71: This is obviously correct.

tests/installation/handle_reboot.pm
```diff
@@ -20,7 +20,7 @@ use utils qw(reconnect_mgmt_console);
 
 sub run {
     if (is_remote_backend) {
-        record_info 'Reconnect mgmt console';
+        record_info 'Remote', 'Reconnect mgmt console';
```
pevik, line 23: Is there some warning in original version? If not and if whole string `Reconnect mgmt console` is visible without clicking in original version then I'd use the original version.

b10n1k: the signature of record_info takes $title and $description. Without you see `Use of uninitialized value $_[0] in join or string at...` i am not sure if title is the right one but the description is also visible hovering over the frame.

lib/utils.pm
b10n1k: *this change is ready for review*

pevik (approved):
LGTM, thanks!

pevik:
Just a note about `Fixes: a1badaf ("utils::zypper_ar(): Fix repo priority on SLE-11")` - that's not supposed to be on the first line (i.e. "subject").
It's a git tag, see example:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v5.16&id=fffb5323780786c81ba005f8b8603d4a558aad28

So there should be normal description on the first line, but later in the body.

PR #14051, opened 2022-01-19
Remove user_import, and show what is done in SUT

JRivrain:
- Related ticket: https://progress.opensuse.org/issues/104791

### Commit 69693e5:

this module was replaced by installation/authentication/import_users.

### Commit 05e551a:

Make use of is_shown to show what is done in SUT.

**Context:**

Within the Yast group, we implemented a standard that goes as follows:
We use libyui-rest-api instead of restapi, more adapted to Yast modules, and a framework (ui-framework-documentation.md) that is inspired by POM.
Each action performed on objects (eg buttons) in the SUT, are defined in a page object. Then business actions (eg create a user) are defined as methods in a controller, then the methods are called by the module.It works very well for Yast modules, reducing the maintenance quite drastically: no needles, no spaghetti code...

**Problem:**

But there is one thing that makes me sad: we no longer see what the module is doing, unless it fails (in which case we get logs and a screenshot). While not strictly necessary, I think it is nice to be able to visualize immediately what the module does without having to dig in the code. And that can be especially useful when we use very high-level methods in test module such as:
```
sub run {
    $testapi::distri->get_local_user()->import_existing_users();
}
```
In which case we'd have to dig in libraries to understand what it does, when in the past, with testapi, you would just have to look at pictures.
**That can be specially useful for developers, to be able to visualize the steps to reproduce a bug.**

**One possible solution is already here:**

To make sure we are on the right page, we implemented for each page a method called is_shown. Then in the controller, we have methods like "get_user_creation_page" which calls the "is_shown" method, and each method from the page is using "get_user_creation_page" hence each time calling is_shown. This would be in my opinion a good place to show "what is shown".

VR: http://waaa-amazing.suse.cz/tests/16132#step/import_users/1


jknphy:
what about my first comment in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14009#pullrequestreview-853925752 shouldn't be much effort to split it, wdyt?

OleksandrOrlov:
I like the solution, as it now shows each step in openQA, that makes the review even more clear. I guess we can apply this solution for other pages as well in future.

jknphy:
LGTM now that the code is more atomic and the framework fully developed.
Only thing to take into account is when the timeout is long to produce an insane number of screenshots:
```
sub wait_installation_popup {
    my ($self, $timeout) = @_;
            return $self->{AbstractOKPopup}->is_shown();
        }, timeout => $timeout,
        message => 'System reboot popup did not appear');
}
```

JRivrain:
> LGTM now that the code is more atomic and the framework fully developed. Only thing is when we use `wait_until` and take into account when the timeout is long enough to produce an insane number of screenshots:
> 
> ```
> sub wait_installation_popup {
>     my ($self, $timeout) = @_;
>     YuiRestClient::Wait::wait_until(object => sub {
>             return $self->{AbstractOKPopup}->is_shown();
>         }, timeout => $timeout,
>         message => 'System reboot popup did not appear');
> }
> ```
> 
> But for not waiting actions should help.

I did what we talked about, here is how it looks http://waaa-amazing.suse.cz/tests/16136#step/perform_installation/2

ge0r (approved):
Looks good!

PR #14053, opened 2022-01-20
Exclude instance pattern for public cloud test with all patterns outside of public cloud images

lemon-suse:
As the discussion result with public cloud team, we need exclude instance pattern for public cloud module test with all patterns outside of public cloud images.

Related ticket: https://progress.opensuse.org/issues/105094
Verification run:
    http://openqa.nue.suse.com/tests/8020244#step/patch_sle/120 (sles12sp4 x86_64)
    https://openqa.nue.suse.com/tests/8039475#step/patch_sle/139 (sles12sp4 s390x)
    https://openqa.nue.suse.com/tests/8039392#step/patch_sle/117 (sles15sp1 ppc64le)
    http://openqa.nue.suse.com/tests/8039540#step/zypper_patch/22 (sles15sp1 aarch64)

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1,4 +1,4 @@
-# Copyright 2015-2021 SUSE LLC
+# Copyright 2015-2022 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package utils;
@@ -11,7 +11,7 @@ use warnings;
 use testapi qw(is_serial_terminal :DEFAULT);
 use lockapi 'mutex_wait';
 use mm_network;
-use version_utils qw(is_microos is_leap is_sle is_sle12_hdd_in_upgrade is_storage_ng is_jeos package_version_cmp);
+use version_utils qw(is_microos is_leap is_public_cloud is_sle is_sle12_hdd_in_upgrade is_storage_ng is_jeos package_version_cmp);
 use Utils::Architectures;
 use Utils::Systemd qw(systemctl disable_and_stop_service);
 use Utils::Backends;
@@ -1960,10 +1960,14 @@ sub install_patterns {
     for my $pt (@pt_list) {
         # if pattern is set default, skip
         next if ($pt =~ /default/);
+        # For Public cloud module test we need skip Instance pattern if outside of public cloud images.
+        next if (($pt =~ /Instance/) && !is_public_cloud);
         # Cloud patterns are conflict by each other, only install cloud pattern from single vender.
         if ($pt =~ /$pcm_list/) {
             next unless $pcm == 0;
-            $pt .= '*';
+            # For Public cloud module test we need install 'Tools' but not 'Instance' pattern if outside of public cloud images.
+            next if (($pt !~ /Tools/) && !is_public_cloud);
+            $pt .= '*' if (is_public_cloud);
             $pcm = 1;
         }
         # Only one CFEngine pattern can be installed
@@ -1980,6 +1984,8 @@ sub install_patterns {
         if (($pt =~ /sap_server/) && is_sle('=11-SP4')) {
             next;
         }
+        # For Public cloud module test we need install 'Tools' but not 'Instance' pattern if outside of public cloud images.
+        next if (($pt =~ /OpenStack/) && ($pt !~ /Tools/) && !is_public_cloud);
         # if pattern is common-criteria and PATTERNS is all, skip, poo#73645
         next if (($pt =~ /common-criteria/) && check_var('PATTERNS', 'all'));
         zypper_call("in -t pattern $pt", timeout => 1800);


chcao:
http://openqa.nue.suse.com/tests/8004130#
Your verify case is not all pattern case. And need to run some online cases.

lib/utils.pm
```diff
@@ -1960,10 +1960,14 @@ sub install_patterns {
     for my $pt (@pt_list) {
         # if pattern is set default, skip
         next if ($pt =~ /default/);
+        # For Public cloud module test we need skip Instance pattern if outside of public cloud images.
+        next if (($pt =~ /Instance/) && check_var('PUBLIC_CLOUD_IMG', '1'));
```
coolgw, line 1964: 1) I have no idea just match /Instance/ key word is enough or not, should we using more strict match pattern?

2) I think PUBLIC_CLOUD_IMG =1 means you are using public cloud img?

3) think we can check qcow name to decide it is public cloud img or not.  @jlausuch what's your suggestion?

jlausuch: I leave it to @asmorodskyi to answer as he is more involved than me in this discussion.

lemon-suse:
> @lemon-suse the discussion regarding this was not over and I planned to contact you in person to clarify some things regarding this but didn't had a chance . Let's have a sync call before moving forward with this

Sure, please tell me your opinion about this. In fact, I have wait one day for the suggestions after sent the mail. Please send me the appointment or talk with me in slack or just reply in the mail.

lib/utils.pm
lemon-suse: 1. I think Instance is enough for now, while if you have better solution it is welcome. :)
2. Yes, it is easy to check by the setting. Anyway, if the image name is special and easy to check, I can accept it.

lib/utils.pm
```diff
@@ -1980,6 +1984,8 @@ sub install_patterns {
         if (($pt =~ /sap_server/) && is_sle('=11-SP4')) {
             next;
         }
+        # For Public cloud module test we need install 'Tools' but not 'Instance' pattern if outside of public cloud images.
+        next if (($pt =~ /OpenStack/) && ($pt !~ /Tools/) && check_var('PUBLIC_CLOUD_IMG', '1'));
```
asmorodskyi, line 1988: can you give examples of strings which you do want to match and examples of strings which you want to skip ? I think it will look more clear and will work better as single regex ...

asmorodskyi:
> @lemon-suse the discussion regarding this was not over and I planned to contact you in person to clarify some things regarding this but didn't had a chance . Let's have a sync call before moving forward with this

this was clarified in a sync call . I just misunderstood this PR intention . So please ignore this

asmorodskyi:
So you want to skip in such cases : 
-  `next if (($pt =~ /Instance/) && check_var('PUBLIC_CLOUD_IMG', '1'));`
-  `next if (($pt !~ /Tools/) && check_var('PUBLIC_CLOUD_IMG', '1'));`
-  `next if (($pt =~ /OpenStack/) && ($pt !~ /Tools/) && check_var('PUBLIC_CLOUD_IMG', '1'));`

this conditions are spread in different places of the loop . I think it will look better if you will create single regex which will cover all 3 cases . I can help you with that if you share usual content of `@pt_list` array . I assume it is always more or less the same ?

lib/utils.pm
lemon-suse: I have updated the code to judge public cloud environment through is_public_cloud, please review again. Thanks.

lib/utils.pm
lemon-suse: There are 4 patterns for openstack:
openstack
openstack-instance-init
openstack-instance-tools
openstack-tools

For non public images env, we only install 'openstack-tools', here need skip 'openstack'.

lemon-suse:
> So you want to skip in such cases :
> 
>     * `next if (($pt =~ /Instance/) && check_var('PUBLIC_CLOUD_IMG', '1'));`
> 
>     * `next if (($pt !~ /Tools/) && check_var('PUBLIC_CLOUD_IMG', '1'));`
> 
>     * `next if (($pt =~ /OpenStack/) && ($pt !~ /Tools/) && check_var('PUBLIC_CLOUD_IMG', '1'));`
> 
> 
> this conditions are spread in different places of the loop . I think it will look better if you will create single regex which will cover all 3 cases . I can help you with that if you share usual content of `@pt_list` array . I assume it is always more or less the same ?

In fact the public patterns should be split to two groups:
1. Amazon-Web-Services|Google-Cloud-Platform|Microsoft-Azure, we need select 1/3 from them to avoid conflict.
2. openstack
Each has 4 patterns, ex:
openstack
openstack-instance-init
openstack-instance-tools
openstack-tools
For non public cloud env, we only install xxx-tools.
In function install_patterns, I skip instance pattern firstly. Then skip xxx in existing pcm-list operation, skip 'openstack' in a special condition. 
I can't give you the content of pt_list, while you can have a look on the job.  https://openqa.nue.suse.com/tests/8004105#step/patch_sle/113
Welcome for any suggestions.

lemon-suse:
> @lemon-suse please use `version_utils::is_public_cloud` when checking for Public Cloud environment.

Thanks, I have updated code with it.

asmorodskyi (approved):
LGTM

asmorodskyi:
>    For non public cloud env, we only install xxx-tools.
>    In function install_patterns, I skip instance pattern firstly. Then skip xxx in existing pcm-list operation, skip 'openstack' in a special condition.
>    I can't give you the content of pt_list, while you can have a look on the job.  https://openqa.nue.suse.com/tests/8004105#step/patch_sle/113
>    Welcome for any suggestions.


ok , I see . it is getting too complicated . let's keep it as it is . feel free to merge as it is

PR #14061, opened 2022-01-20
Fix nodejs_openssl test

michaelgrifalconi:
Fix nodejs tests:
- unschedule one test (that run on ibs/obs anyway)
- install additional package to other test

- Related ticket: https://progress.opensuse.org/issues/105148
- Needles: no needles
- Verification run: 

15-SP3 x86 https://openqa.suse.de/tests/8005995#
12-sp5 x86 https://openqa.suse.de/tests/8005999#live 
12-sp5 aarch https://openqa.suse.de/tests/8006000

### DIFF
--- a/data/console/test_openssl_nodejs.sh
+++ b/data/console/test_openssl_nodejs.sh
@@ -86,7 +86,7 @@ main(){
   echo "OS_VERSION: $OS_VERSION"
 
   # Install dependencies to apply source patches and run tests
-  zypper -n in quilt rpm-build
+  zypper -n in quilt rpm-build  libopenssl1_1-hmac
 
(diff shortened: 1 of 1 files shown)


tjyrinki (approved):
Looks good, thank you a lot for fixing!

punkioudi (approved):
Lgtm! 
One small thing, how the `This should be skipped also on future versions` could be tracked?

michaelgrifalconi:
Very good question @punkioudi , right now the script can only handle a `full version -> test file` record to skip. This means on the next update we will need to add another line. Let's see how often it happens and if it makes sense to change something in the code!

PR #14071, opened 2022-01-21
Don't change system role if it's already selected

dzedro:
- Fail: https://openqa.suse.de/tests/8010789
- Verification run:
https://openqa.suse.de/tests/overview?distri=sle&build=system_role SLE
https://openqa.opensuse.org/tests/overview?build=system_role openSUSE

### DIFF
--- a/tests/installation/system_role.pm
+++ b/tests/installation/system_role.pm
@@ -70,11 +70,8 @@ sub assert_system_role {
 }
 
 sub run {
-    if (is_sle('=12-SP5') && !is_x86_64) {
+    if (is_sle('<15') && !is_x86_64) {
         record_info("Skip screen", "System Role screen is displayed only for x86_64 in SLE-12-SP5 due to it has more than one role available");
-    }
-    elsif (is_aarch64 && is_sle('>=12-SP3') && is_sle('<15')) {
-        record_info("Skip screen", "System Role screen is  not displayed on aarch64 between 12SP3 and 12SP5");
     }
     else {
         assert_system_role;
--- a/tests/installation/system_role.pm
+++ b/tests/installation/system_role.pm
@@ -42,9 +42,6 @@ sub change_system_role {
         send_key 'alt-' . $role_hotkey{$system_role};
         assert_screen "system-role-$system_role-selected";
     }
-    # every system role other than default will end up in textmode for SLE
-    # But can be minimalx/lxde/xfce
-    set_var('DESKTOP', 'textmode') unless is_opensuse;
 }
 
 sub assert_system_role {
--- a/tests/installation/system_role.pm
+++ b/tests/installation/system_role.pm
@@ -34,13 +34,8 @@ sub change_system_role {
             send_key_until_needlematch "system-role-$system_role-selected", 'spc';    # enable role
         }
         else {
-            if (!check_screen("system-role-$system_role-selected")) {
-                # scroll down in case role is not visible straight away
-                send_key_until_needlematch "system-role-$system_role", 'down', 5, 1;
-                assert_and_click "system-role-$system_role";
-                # after role is selected, we get top of the list shown, so might need to scroll again
-                send_key_until_needlematch "system-role-$system_role-selected", 'down', 5, 1;
-            }
+            assert_and_click "system-role-$system_role";
+            assert_and_click "system-role-$system_role-selected";
         }
     }
     else {
--- a/tests/installation/system_role.pm
+++ b/tests/installation/system_role.pm
@@ -52,14 +52,14 @@ sub assert_system_role {
     # Asserting screen with preselected role
     # Proper default role assertion will be addressed in poo#37504
     # Product might or might not have default selected
-    if (is_opensuse) {
-        assert_screen('before-role-selection', 180);
-        change_system_role(get_var('SYSTEM_ROLE', get_var('DESKTOP')));
+    assert_screen('before-role-selection', 180);
+    if (is_opensuse || (get_var('SYSTEM_ROLE') && !check_var('SYSTEM_ROLE', 'default'))) {
+        # on opensuse is system role 1:1 with DESKTOP, on SLE change system role if it's defined
+        my $system_role = is_opensuse ? get_var('SYSTEM_ROLE', get_var('DESKTOP')) : get_var('SYSTEM_ROLE');
+        change_system_role($system_role) unless check_screen("system-role-$system_role-selected");
     }
-    else {
-        assert_screen('system-role-default-system', 180);
-        my $system_role = get_var('SYSTEM_ROLE', 'default');
-        change_system_role($system_role) if (get_var('SYSTEM_ROLE') && !check_var('SYSTEM_ROLE', 'default'));
+    elsif (check_var('SYSTEM_ROLE', 'default') || !get_var('SYSTEM_ROLE')) {
+        record_info('Default', 'SYSTEM_ROLE is default or not defined, same result');
     }
     send_key $cmd{next};
 }


dzedro:
The SYSTEM_ROLE default is ambiguous. It' can be anything based on with what modules is system registered. e.g.
15 SP4 now is installed with default moduels without DM gnome, so default is textmode
When Dsktop modules is added but the DESKTOP is textmode without specific PATTERNS then default will be fist option Gnome

jknphy:
I think that default it means that there is not selection on the screen, so it should work if you just set `SYSTEM_ROLE =default`.

tests/installation/system_role.pm
```diff
@@ -64,7 +64,7 @@ sub assert_system_role {
     else {
         assert_screen('system-role-default-system', 180);
         my $system_role = get_var('SYSTEM_ROLE', 'default');
-        change_system_role($system_role) if (get_var('SYSTEM_ROLE') && !check_var('SYSTEM_ROLE', 'default'));
+        change_system_role($system_role) if (get_var('SYSTEM_ROLE') && !check_var('SYSTEM_ROLE', 'default') && !check_screen("system-role-$system_role-selected"));
```
foursixnine, line 67: Hmm this might mean that https://github.com/os-autoinst/os-autoinst/blob/a9af388073889ea5a28f3613d4b40fdfcdee265d/testapi.pm#L1415 might be broken somewhere down the line.

@okurz do you have an idea of what might have caused https://openqa.suse.de/tests/8010789#step/system_role/2?

foursixnine (approved):
I don't mind the extra check.

jknphy (approved):

tests/installation/system_role.pm
foursixnine: @dzedro Line33 is the culprit of the problem, it should first do `if !check_screen("system-role-$system_role-selected"`

foursixnine (changes requested):

tests/installation/system_role.pm
dzedro: Yo umean 37, but not for videomode text ? `check_var('VIDEOMODE', 'text')`

dzedro:
> 

I guess you are no using it on 15SP4+

tests/installation/system_role.pm
foursixnine: Yeah, for the `VIDEOMODE=text` the extra check has to be added too

dzedro:
> I think that default it means that there is not selection on the screen, so it should work if you just set `SYSTEM_ROLE =default`. The one failing has `SYSTEM_ROLE=textmode`. But I guess if you want to make it work in both ways because you need to env variable later, then your change should be fine (note: we are not using this module anymore in YaST group).

I guess you don't use this modules on 15SP4+

dzedro:
btw instead of `assert_screen 'system-role-default-system'` and many needles nobody care about, there could be just save_screen

tests/installation/system_role.pm
okurz: You should be able to just use `match_has_tag` instead of `check_screen` if the needles used in the above `assert_screen` include the tag you look for. Keep in mind that needles can have multiple tags.

dzedro: But they must not e.g. VR https://openqa.suse.de/tests/8013200/modules/system_role/steps/2/edit

foursixnine:
> I think that default it means that there is not selection on the screen, so it should work if you just set `SYSTEM_ROLE =default`. The one failing has `SYSTEM_ROLE=textmode`. But I guess if you want to make it work in both ways because you need to env variable later, then your change should be fine (note: we are not using this module anymore in YaST group).

See: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14071#pullrequestreview-859463485

> The SYSTEM_ROLE default is ambiguous. It' can be anything based on with what modules is system registered. e.g. 15 SP4 now is installed with default moduels without DM gnome, so default is textmode. https://openqa.suse.de/tests/8010791#step/system_role/1

Somebody actually added SYSTEM_ROLE=textmode between the last and this build, which triggered the call to  `change_system_role` which triggered the bug that you see at: https://openqa.suse.de/tests/8010789#step/system_role/1

_Default system was selected, but a missing check screen* due to differences beetween types of installations, made openQA what it was supposed to do: keep searching for the one it was looking for_
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/cae6400/tests/installation/system_role.pm#L31



> When Desktop module is added but the DESKTOP is textmode without specific PATTERNS then default will be fist option Gnome. https://openqa.suse.de/tests/8010805#step/system_role/1 Not sure what's the purpose of SYSTEM_ROLE _default_

Default means, default according to the selected products (https://progress.opensuse.org/issues/37504), so we should have: default-gnome, default-textmode and so on needles.

dzedro:
I added it, what do you mean with bug, it is not bug!
I added it to fix `create_hdd_textmode`
https://openqa.suse.de/tests/7998054 -> https://openqa.suse.de/tests/8017919
Expecting 'multi-user', but getting 'graphical' is normal. https://openqa.suse.de/tests/7998054#step/installation_overview/6
You can change system target (DM) via system role, patterns or manually, but expecting multi-user with **default** system role gnome is wrong.

jknphy:
> multi-user

that failure I guess it because it has DESKTOP=textmode, https://openqa.suse.de/tests/7998054#step/installation_overview/6 right? because it is a textmode installation but the installed system is graphical, installation is controlled by videmode.

dzedro:
> > multi-user
> 
> that failure I guess it because it has DESKTOP=textmode, https://openqa.suse.de/tests/7998054#step/installation_overview/6 right? because it is a textmode installation but the installed system is graphical, installation is controlled by videmode.

The failure happened because there is no system role or patterns which would set the textmode.
After I defined system role, another test failed because systom role test is broken.
videomode does only define if installation is ncurses or gui. https://openqa.suse.de/tests/7998054#step/bootloader/9

foursixnine (approved):
Thanks @dzedro!

dzedro:
Done.
Added VRs
It can happen there will be failures due to missing needles, but I will take care of that.
It can also happen removed `set_var` will cause failures, but I could not find tests where that could happen. :crossed_fingers: 
If that happens then I will add the set_var for that specific case.

btw the failure which triggered this PR can be avoided by just removing DESKTOP -> https://dzedro.suse.cz/tests/20613
The `check_default_target` and interaction between variables DESKTOP, SYSTEM_ROLE & PATTERNS do require improvement.

tests/installation/system_role.pm
```diff
@@ -34,48 +34,37 @@ sub change_system_role {
...
 sub assert_system_role {
     # Still initializing the system at this point, can take some time
     # Asserting screen with preselected role
     # Proper default role assertion will be addressed in poo#37504
     # Product might or might not have default selected
-    if (is_opensuse) {
-        assert_screen('before-role-selection', 180);
-        change_system_role(get_var('SYSTEM_ROLE', get_var('DESKTOP')));
+    assert_screen('before-role-selection', 180);
+    if (is_opensuse || (get_var('SYSTEM_ROLE') && !check_var('SYSTEM_ROLE', 'default'))) {
```
b10n1k, line 53: are you sure you need `get_var('SYSTEM_ROLE')`. seems like `check_var` already does this.

tests/installation/system_role.pm
```diff
@@ -34,48 +34,37 @@ sub change_system_role {
...
+        # on opensuse is system role 1:1 with DESKTOP, on SLE change system role if it's defined
+        my $system_role = is_opensuse ? get_var('SYSTEM_ROLE', get_var('DESKTOP')) : get_var('SYSTEM_ROLE');
```
b10n1k, line 55: let me know if this looks more clear to you
`is_sle ? get_var('SYSTEM_ROLE') : get_var('DESKTOP');`
iiuc this is the same thing. wdyt?

tests/installation/system_role.pm
dzedro: You miss the `!`

tests/installation/system_role.pm
dzedro: It's not same.

tests/installation/system_role.pm
dzedro: @b10n1k You are right, I didn't have to use `get_var`. Sorry didn't read it good enough.

PR #14081, opened 2022-01-22
Fix installation_overview because of BSC1167736 for openSUSE Leap on s390x

skriesch:
The test-suite textmode-server is failing for openSUSE Leap 15.3 and openSUSE Leap 15.4 on s390x because of the SUSE internal bugreport BSC1167736.
This fix in the test should fix this issue by adding textmode-server equal as textmode to the expected_target.

- Related ticket: https://progress.opensuse.org/issues/81685
- Verification run: https://openqa.opensuse.org/t2150725

### DIFF
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -67,7 +67,7 @@ sub check_default_target {
     return if (!get_var('DESKTOP') || get_var('PATTERNS'));
 
     # Set expectations
-    my $expected_target = check_var('DESKTOP', 'textmode*') ? "multi-user" : "graphical";
+    my $expected_target = check_var('DESKTOP', 'textmode' | 'textmode-server' ) ? "multi-user" : "graphical";
 
     $self->validate_default_target($expected_target);
 }
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -67,7 +67,7 @@ sub check_default_target {
     return if (!get_var('DESKTOP') || get_var('PATTERNS'));
 
     # Set expectations
-    my $expected_target = check_var('DESKTOP', 'textmode') ? "multi-user" : "graphical";
+    my $expected_target = check_var('DESKTOP', 'textmode*') ? "multi-user" : "graphical";
 
     $self->validate_default_target($expected_target);
 }
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -67,7 +67,7 @@ sub check_default_target {
     return if (!get_var('DESKTOP') || get_var('PATTERNS'));
 
     # Set expectations
-    my $expected_target = check_var('DESKTOP', 'textmode' | 'textmode-server' ) ? "multi-user" : "graphical";
+    my $expected_target = check_var('DESKTOP', 'textmode' | 'textmode-server') ? "multi-user" : "graphical";
 
     $self->validate_default_target($expected_target);
 }
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -65,7 +65,6 @@ sub check_default_target {
         get_var('REMOTE_CONTROLLER') || (get_var('BACKEND', '') =~ /spvm|pvm_hmc|ipmi/));
     # exclude non-desktop environment and scenarios with edition of package selection (bsc#1167736)
     return if (!get_var('DESKTOP') || get_var('PATTERNS'));
-    return if (get_var 'BSC1167736');
 
     # Set expectations
     my $expected_target = check_var('DESKTOP', 'textmode') ? "multi-user" : "graphical";


tests/installation/installation_overview.pm
```diff
@@ -65,10 +65,9 @@ sub check_default_target {
         get_var('REMOTE_CONTROLLER') || (get_var('BACKEND', '') =~ /spvm|pvm_hmc|ipmi/));
     # exclude non-desktop environment and scenarios with edition of package selection (bsc#1167736)
     return if (!get_var('DESKTOP') || get_var('PATTERNS'));
-    return if (get_var 'BSC1167736');
 
     # Set expectations
-    my $expected_target = check_var('DESKTOP', 'textmode') ? "multi-user" : "graphical";
+    my $expected_target = check_var('DESKTOP', 'textmode*') ? "multi-user" : "graphical";
```
okurz, line 70: How is the `*` supposed to work? I don't understand what you want to achieve with that. AFAIK there is no globbing or regex support in `check_var` so you would compare against a literal string `textmode*`. I hope nobody used that in test settings. Maybe you want

```suggestion
    my $expected_target = get_var('DESKTOP', '') =~ qr/textmode/ ? "multi-user" : "graphical";
```

skriesch: textmode and textmode-server should be used in this case. That is a regex, that both are applied.

okurz: Right. My proposal would work for that. Does `check_var('DESKTOP', 'textmode*')` actually work?

skriesch:
@os-autoinst/tests-maintainer Can you merge, please?

tests/installation/installation_overview.pm
skriesch: Watch my verification run https://openqa.opensuse.org/t2150725 . That is textmode-server, which has been failing before.

okurz:
@os-autoinst/tests-maintainer please see the above change

tests/installation/installation_overview.pm
okurz: ok, let's try

okurz (approved):

skriesch:
Can you also add me to the group, please? It seems, that I can not notify with @os-autoinst/tests-maintainer without these permissions...

Vogtinator (changes requested):
I don't think this is correct. The linked test is called `textmode-server`, but has `DESKTOP=textmode`

skriesch:
That was Sergio's suggestion:[ Progress](https://progress.opensuse.org/issues/81685#note-15)

Vogtinator:
That it works in the verification run is an accident.

bsc#1167736 is a bug report that an installation over VNC enables `graphical.target` by default. It's closed as WONTFIX.
This happens in the textmode-server test scenario.

`check_var('DESKTOP', 'textmode*')` does not match with `DESKTOP=textmode`, and so it expects `graphical.target` to be the default, which matches because of that bug.

It will fail in all textmode cases without that bug.

skriesch:
Then I want to replace the star with a pipe and these 2 options...

Vogtinator:
The correct fix is to set `BSC1167736=yes` in Leap, just like it is for Tumbleweed already.

okurz:
> Can you also add me to the group, please? It seems, that I can not notify with @os-autoinst/tests-maintainer without these permissions...

Done

skriesch:
I tried it with BSC1167736=yes in the Leap config. I did a rerun of the test with this setting (you can see that in openQA continuously). That did not help...

Vogtinator:
> I tried it with BSC1167736=yes in the Leap config. I did a rerun of the test with this setting (you can see that in openQA continuously). That did not help...

Please link that failed test run and I'll have a look

skriesch:
https://openqa.opensuse.org/tests/2150724#step/installation_overview/7

Vogtinator:
> https://openqa.opensuse.org/tests/2150724#step/installation_overview/7

That doesn't have the variable set, it would be visible on the "Settings" page otherwise. I assume you changed that in the configuration on o3 and then retriggered the test? That won't have any effect as those settings are only used for new builds.

Vogtinator:
```
> openqa-clone-job --within-instance https://openqa.opensuse.org --skip-chained-deps 2150724 BSC1167736=yes
Created job #2156943: opensuse-15.4-DVD-s390x-Build140.1-textmode-server@s390x-zVM-vswitch-l2 -> https://openqa.opensuse.org/t2156943
```

This test has the variable set and should pass

skriesch:
And I did it on this way in openQA:
```
  s390x:
    opensuse-Leap-15.4-DVD-s390x:
      - textmode:
          priority: 50
          settings:
            EXTRABOOTPARAMS: hvc_iucv=8
      - textmode-server:
          testsuite: null
          settings:
            DESKTOP: textmode
            EXTRABOOTPARAMS: hvc_iucv=8
            FILESYSTEM: xfs
            BSC1167736: yes
```

Vogtinator:
> And I did it on this way in openQA:
> 
> ```
>   s390x:
>     opensuse-Leap-15.4-DVD-s390x:
>       - textmode:
>           priority: 50
>           settings:
>             EXTRABOOTPARAMS: hvc_iucv=8
>       - textmode-server:
>           testsuite: null
>           settings:
>             DESKTOP: textmode
>             EXTRABOOTPARAMS: hvc_iucv=8
>             FILESYSTEM: xfs
>             BSC1167736: yes
> ```

Changes there do not affect existing jobs or clones/restarts of those. You'll have to clone them and set the variable explicitly or reschedule the product.

skriesch:
I expected that a rerun is using the new existing configuration.

okurz (changes requested):
`check_var('DESKTOP', 'textmode' | 'textmode-server')` seems again like wishful thinking and inventing an API that does not yet exist :)

skriesch:
@Vogtinator You are right that the configuration is working. My expectations were wrong, that a re-run would use the new configuration. Then I close this PR.

Lessons learned...

foursixnine:
See also #14071

PR #14082, opened 2022-01-23
Introduce nginx test module

pdostal:
- Verification run: [SLE 15-SP3](http://pdostal-server.suse.cz/tests/13066) [Tumbleweed](http://pdostal-server.suse.cz/tests/13067)


lib/services/nginx.pm
```diff
@@ -0,0 +1,94 @@
...
+# check service is running and enabled
+sub check_service {
+    common_service_action 'nginx', $service_type, 'is-enabled';
+    common_service_action 'nginx', $service_type, 'is-active';
+}
+
+sub check_function {
+    my $grep = ' | grep "Hello from nginx"';
+    # nginx does not respond right away after service restart
+    #script_retry('curl http://localhost/' . $grep, delay => 3, retry => 3);
```
okurz, line 63: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

data/console/nginx_vhost.conf
```diff
@@ -0,0 +1,30 @@
+server {
+  listen 80;
```
grisu48, line 2:
```suggestion
  listen 80 default_server;
```

lib/services/nginx.pm
```diff
@@ -0,0 +1,94 @@
...
+    assert_script_run('curl http://localhost/');
```
grisu48, line 64: I'd add the canary `grep` here as well?

```suggestion
    assert_script_run('curl http://localhost/');  # Check if basic connection is working
    assert_script_run('curl http://localhost/' . $grep); # check for canary
```

grisu48 (commented):
Good job! Some minor comments, overall a solid nice work :+1:

data/console/nginx_vhost.conf
```diff
@@ -0,0 +1,30 @@
...
+  # IPv6 on port 80 is disabled
```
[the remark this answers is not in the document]
pdostal: Yes, regarding my testing port 80 is disabled in `nginx.conf` by default.

data/console/nginx_vhost.conf
pdostal: Why `default_server`? There is already one entry in `nginx.conf`.

pdostal: image
In the `/etc/nginx/nginx.conf` there is:
```
    server {
        listen       80;
        server_name  localhost;
```

data/console/nginx_vhost.conf
```diff
@@ -0,0 +1,30 @@
+server {
```
grisu48, line 1: See the following suggestion for the `nginx_vhost.conf`, which includes IPv6 on the `default_server` so that also `curl -6 http://localhost` would work.

```suggestion
server {
```
server {
  listen 80 default_server;
  listen [::]:80 default_server;

  server_name localhost;

  root /srv/www/htdocs;
  index index.html;

  location / {
    try_files $uri $uri/ =404;
  }
}

server {
  listen 80;
  listen [::]:80;
  listen 81;
  listen [::]:81;

  server_name vhost;

  return 301 https://$host$request_uri;
}

server {
  listen 443 ssl http2;
  listen [::]:443 ssl ipv6only=on http2;
  listen 444 ssl http2;
  listen [::]:444 ssl ipv6only=on http2;

  server_name vhost;

  root /srv/www/htdocs;
  index index.html;

  location / {
    try_files $uri $uri/ =404;
  }

  ssl_certificate /etc/nginx/ssl/cert.pem;
  ssl_certificate_key /etc/nginx/ssl/key.pem;
}
```

**Review by grisu48 — CHANGES_REQUESTED** [review]
Please see my suggestion for the `nginx_vhost.conf` which make the usage of IPv6 on port 80 possible.

File: data/console/nginx_vhost.conf (hunk 5 above)
Comment (grisu48) [reply data/console/nginx_vhost.conf:1]: See http://pasta.qam.suse.de/WyszsucM for quick copy&paste

File: data/console/nginx_vhost.conf (hunk 2 above)
Comment (pdostal) [reply data/console/nginx_vhost.conf:2]: Solved by defining the `default_server` as @grisu48 suggested.

**Review by grisu48 — APPROVED** [review]
LGTM :rocket:

**Comment by punkioudi** [comment]
Nice work Pavel! Could you also add the verification runs for the other sle versions that `nginx.pm` is scheduled?

**Review by okurz — CHANGES_REQUESTED** [review]
See https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/4933896099?check_suite_focus=true

**Comment by pdostal** [comment]
> See https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/4933896099?check_suite_focus=true

I saw that but I don't know what that means. Can you help?

**Comment by pdostal** [comment]
> > > See https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/4933896099?check_suite_focus=true
> > 
> > 
> > I saw that but I don't know what that means. Can you help?
> 
> It means that the test module that you add will be executed as part of the "default schedule" which might also be the one used for openSUSE+SLE staging tests. I don't know about that because I lost track with the thousand of yaml schedule files that people prefer :) If it is intended to include the nginx module in that schedule then adjust accordingly the reference schedule file that is used to compare against.

I schedule this test is consoletests (which I believe is used in openSUSE) and in extratests2 (which is used in SLE maintenance). I don't know what 'reference schedule file' means.

**Comment by okurz** [comment]
> > > > See https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/4933896099?check_suite_focus=true
> > > 
> > > 
> > > I saw that but I don't know what that means. Can you help?
> > 
> > 
> > It means that the test module that you add will be executed as part of the "default schedule" which might also be the one used for openSUSE+SLE staging tests. I don't know about that because I lost track with the thousand of yaml schedule files that people prefer :) If it is intended to include the nginx module in that schedule then adjust accordingly the reference schedule file that is used to compare against.
> 
> I schedule this test is consoletests (which I believe is used in openSUSE) and in extratests2 (which is used in SLE maintenance).

Yes, I know. But then I recommend to also provide according verification runs for staging tests.

> I don't know what 'reference schedule file' means.

That would be https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/t/data/test_schedule.out

**Comment by pdostal** [comment]
> > I don't know what 'reference schedule file' means.
> 
> That would be https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/t/data/test_schedule.out

I added `tests/console/nginx.pm` to that file but the result still looks the same.
Are there some examples or documentation about this CI? I still don't understand how it works.

**Comment by okurz** [comment]
> > > I don't know what 'reference schedule file' means.
> > 
> > 
> > That would be https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/t/data/test_schedule.out
> 
> I added `tests/console/nginx.pm` to that file but the result still looks the same. Are there some examples or documentation about this CI? I still don't understand how it works.

The test "isotovideo / CI: Running isotovideo test" is triggered from https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/.github/workflows/ci-isotovideo.yml which in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/.github/workflows/ci-isotovideo.yml#L12 calls the line `make test-isotovideo` so we take a look into the Makefile of the repository. In https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/Makefile#L160 one can see the make rule.  So to manually reproduce you can simply call `make test-isotovideo`. In the Makefile we can see that the make rule is actually just a simple call to `tools/test_isotovideo`. And what that script does can be seen in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tools/test_isotovideo#L5 which is basically two lines: Execute a run of isotovideo from a container and call `diff` to compare the evaluated schedule against the reference from the file that we mentioned earlier.

https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/4949811915?check_suite_focus=true#step:3:92 shows that there is a diff showing references to nginx module in both directions. This means that you likely added nginx but not in the right order. So if you add nginx in exactly the order as you added to the schedule files then the test should pass which I suggest to check locally to not need to waste time waiting for CI runs in github actions.

**Comment by okurz** [comment]
Good, the test passes :) Now please include verification runs of staging tests, e.g. https://openqa.opensuse.org/tests/2158334

**Comment by pdostal** [comment]
I just started [verification run of staging tests](http://pdostal-server.suse.cz/tests/13151) but it's still running.

**Comment by pdostal** [comment]
@okurz the module has been unscheduled for `is_opensuse && is_staging` as the package is not available in this scenario.

**Review by okurz — APPROVED** [review]
Good. So this is why we do this testing :)

**Review by grisu48 — APPROVED** [review]

**Comment by DimStar77** [comment]
https://openqa.opensuse.org/tests/2167877#step/nginx/48

nginx test failiing on TW

**Comment by DimStar77** [comment]
> https://openqa.opensuse.org/tests/2167877#step/nginx/48
> 
> nginx test failiing on TW

In the VR posted in comment 0, nginx was tested very early in the schedule (even before prepare_test_data); in the variant merged and now active though, it is close to the end, way past apache - no wonder they fight for port 80:

```
Jan 31 11:35:39 susetest nginx[19056]: nginx: [emerg] bind() to 0.0.0.0:80 failed (98: Address already in use)
```

(in the VR, apache failed for that reason)

punkioudi:
It fails also in all sle versions : https://progress.opensuse.org/issues/105786
eg https://openqa.suse.de/tests/8080202#step/nginx/75

Should we probably revert it?

DimStar77:
> It fails also in all sle versions : https://progress.opensuse.org/issues/105786 eg https://openqa.suse.de/tests/8080202#step/nginx/75
> 
> Should we probably revert it?

As there is basically no success rate, I'd say it's the best. Back to the drawing board

PR #14083, opened 2022-01-24
Remove the record_soft_failure about the type missing in firefox_nss

jouyingbin:
As poo#77143 is already fixed and hardly reproduced in aarch64
thus removing the soft-fail track from firefox_nss

- Related ticket: https://progress.opensuse.org/issues/105343
- Needles: NA
- Verification run: https://openqa.suse.de/t8020283 (aarch64)

### DIFF
--- a/tests/fips/mozilla_nss/firefox_nss.pm
+++ b/tests/fips/mozilla_nss/firefox_nss.pm
@@ -10,7 +10,7 @@
 #
 # Maintainer: Ben Chou <bchou@suse.com>
 # Tag: poo#47018, poo#58079, poo#71458, poo#77140, poo#77143,
-#      poo#80754, poo#104314, poo#104989
+#      poo#80754, poo#104314, poo#104989, poo#105343
 
 use base "x11test";
 use strict;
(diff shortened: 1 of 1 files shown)


rfan1 (approved):
LGTM

Amrysliu (approved):

tests/fips/mozilla_nss/firefox_nss.pm
```diff
@@ -176,7 +176,7 @@ sub run {
         # Retype password again once the password missed input
         # The problem frequently happaned in aarch64
         if (check_screen("firefox-password-typefield-miss")) {
-            record_soft_failure "Firefox password is missing to input, see poo#77143";
+            record_info("aarch64 type_missing", "Firefox password is missing to input, please refer to bsc#1179749 & poo#105343");
```
lilyeyes, line 179: remove poo#77143 is ok
but added bsc#1179749 & poo#105343 is useless too, as the bug is fixed and the poo will be resolved too.

jouyingbin: yes, because it is just a reference if the issue happened again randomly next time. It will have a record to check which bug or poo were reported before. is that ok?

lilyeyes (approved):
lgtm

tests/fips/mozilla_nss/firefox_nss.pm
lilyeyes: Sure, it is ok.

PR #14088, opened 2022-01-24
Fix erratic behavior of setting up the network on windows

bs-suse:
This fix uses the windows from wsl/boot_windows as base.
Due to the issue that the network adapter is sometimes (possibly
depending on the worker setup?) named "Ethernet" and sometimes
"Ethernet 2" there is the need to first find the name and use that
to actually set up the network.
This fix implements that.
For that, powershell cmdlets are used instead of netsh.

- Related issue: https://openqa.suse.de/tests/7924149#step/windows_network_setup/5
- Needles: https://gitlab.suse.de/bschmidt/os-autoinst-needles-sles/-/merge_requests/2
- Verification run: http://keks.qa.suse.de/tests/386#step/windows_network_setup/21

### DIFF
--- a/tests/x11/remote_desktop/windows_network_setup.pm
+++ b/tests/x11/remote_desktop/windows_network_setup.pm
@@ -7,33 +7,34 @@
 # Maintainer: GraceWang <gwang@suse.com>
 # Tags: tc#1610388
 
-use strict;
-use warnings;
-use base 'x11test';
+use Mojo::Base qw(windowsbasetest);
 use testapi;
-use lockapi;
+
+sub approve_network_popup {
+    # if there is the networks popup that ask's if it's ok to be discoverable, approve it
+    wait_still_screen stilltime => 5, timeout => 10;
+    assert_screen([qw(networks-popup-be-discoverable no-network-discover-popup)], 60);
+    if (match_has_tag 'networks-popup-be-discoverable') {
+        assert_and_click 'network-discover-yes';
+        wait_screen_change(sub { send_key 'ret' }, 10);
+    }
+}
 
 sub run {
     my $self = shift;
 
-    send_key "super-x";
-    assert_and_click "windows-powershell-admin";
-    assert_and_click "windows-powershell-yes";
-
-    assert_and_click "powershell-started", timeout => 120;
-    type_string "netsh interface ip set address name=Ethernet static 10.0.2.18 255.255.255.0 10.0.2.2";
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
LGTM. Thanks!

tests/x11/remote_desktop/windows_network_setup.pm
```diff
@@ -1,39 +1,41 @@
 # SUSE's openQA tests
 #
-# Copyright 2018-2021 SUSE LLC
+# Copyright 2018-2022 SUSE LLC
```
okurz, line 3: please keep the old date. See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md?plain=1#L49 for this

tests/x11/remote_desktop/windows_network_setup.pm
```diff
@@ -1,39 +1,41 @@
...
-use warnings;
-use base 'x11test';
+use Mojo::Base qw(windowsbasetest);
 use testapi;
-use lockapi;
+use version_utils qw(is_sle is_opensuse);
+
+sub approve_network_popup {
+    # if there is the networks popup that ask's if it's ok to be discoverable, approve it
+    wait_still_screen stilltime => 5, timeout => 10;
```
okurz, line 16: There should not be a need for a `wait_still_screen` as the subsequent `assert_screen` should already cover the waiting period with the sufficient timeout

okurz (changes requested):
* Please phrase the git commit message subject lines starting with a capital letter
* Squash the commits
* See inline comments

tests/x11/remote_desktop/windows_network_setup.pm
okurz: Then you might need to reconsider your needles that you use in the next call. Just think how a human tester would conduct this and we should realize we shouldn't need this command here :)

okurz (approved):

DimStar77:
Is that related to the new failures on https://openqa.opensuse.org/tests/2156917#step/windows_network_setup/20 ?

okurz:
You are right. https://openqa.opensuse.org/tests/2156549#investigation very much makes it look like it. I forgot to ask for openSUSE verification. I am not that good of a reviewer for os-autoinst-distri-opensuse anymore, it seems ;) I will propose a revert.

PR #14091, opened 2022-01-24
wicked: Fix cleanup of old config files in `wickedbase::reset_wicked()`

cfconrad:
Simply check with:
```bash
cd $(mktemp -d)
touch routes ifroute-eth0
find .  -name "routes" -o -name "ifroute-*"  -print
# compared to:
find . \( -name "routes" -o -name "ifroute-*" \) -print
```

### DIFF
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -100,8 +100,8 @@ sub assert_wicked_state {
 sub reset_wicked {
     my $self = @_;
     # Remove any config file and leave the system clean to start tests
-    assert_script_run('find /etc/sysconfig/network/ -name "ifcfg-*" -not -name "ifcfg-lo" -exec rm {} \;');
-    assert_script_run('find /etc/sysconfig/network/ -name "routes" -o -name "ifroute-*" -exec rm {} \;');
+    assert_script_run('find /etc/sysconfig/network/ \( -name "ifcfg-*" -not -name "ifcfg-lo" \) -exec rm {} \;');
+    assert_script_run('find /etc/sysconfig/network/ \( -name "routes" -o -name "ifroute-*" \) -exec rm {} \;');
 
     # Remove any previous manual ip configuration
     my $iface = iface();


lib/wickedbase.pm
```diff
@@ -100,8 +100,8 @@ sub assert_wicked_state {
 sub reset_wicked {
     my $self = @_;
     # Remove any config file and leave the system clean to start tests
-    assert_script_run('find /etc/sysconfig/network/ -name "ifcfg-*" -not -name "ifcfg-lo" -exec rm {} \;');
-    assert_script_run('find /etc/sysconfig/network/ -name "routes" -o -name "ifroute-*" -exec rm {} \;');
+    assert_script_run('find /etc/sysconfig/network/ \( -name "ifcfg-*" -not -name "ifcfg-lo" \) -exec rm {} \;');
```
asmorodskyi, line 103: can elaborate what is going on here ? why you encapsulate just two *arguments* into brackets ? when I try to run this command in bash I am getting : 
```
find /etc/sysconfig/network/ ( -name "ifcfg-*" -not -name "ifcfg-lo" ) -exec echo {};
bash: syntax error near unexpected token `('
```

from what I know `(` is creating subshell . what the point to create a subshell for `-name "ifcfg-*" -not -name "ifcfg-lo" ` ?

asmorodskyi:
please provide VR

lib/wickedbase.pm
cfconrad: you need to escape the `(` so it goes into `find` and is not interpreted by the shell...
from `man find`
```
   OPERATORS
       Listed in order of decreasing precedence:

       ( expr )
              Force precedence.  Since parentheses are special to  the  shell,
              you  will  normally need to quote them.  Many of the examples in
              this manual page use backslashes  for  this  purpose:  `\(...\)'
              instead of `(...)'.
```

cfconrad:
VR: https://openqa.suse.de/tests/8037348

PR #14092, opened 2022-01-24
 wicked: Add test t17_ifrename

cfconrad:
Bug: https://bugzilla.suse.com/show_bug.cgi?id=1194392

- Verification run: http://openqa.wicked.suse.de/tests/47837

### DIFF
--- a/tests/wicked/startandstop/sut/t17_ifrename.pm
+++ b/tests/wicked/startandstop/sut/t17_ifrename.pm
@@ -1,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: wicked
+# Summary: Rename an interface with udev and call ifreload/ifup
+# Maintainer: Anton Smorodskyi <asmorodskyi@suse.com>
+#             Clemens Famulla-Conrad <cfamullaconrad@suse.de>
+#             qa-c <qa-c@suse.de>
+
+use Mojo::Base 'wickedbase';
+use testapi;
+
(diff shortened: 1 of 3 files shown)


tests/wicked/startandstop/sut/t17_ifrename.pm
```diff
@@ -0,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: wicked
+# Summary: Rename an interface with udev and call ifreload/ifup
+# Maintainer: Anton Smorodskyi <asmorodskyi@suse.com>
+#             Jose Lausuch <jalausuch@suse.com>
```
jlausuch, line 9: feel free to remove me :D 
Or just add qac ML

asmorodskyi (changes requested):
disagree with changes related to check_logs . Don't see why we need to add flow with `$code` and `$cursor` inside check_logs function it is unrelated to function purposes

cfconrad:
The advantage is, that you actually can tell what call cause the log error as you can check logs during the testmodule and not just at the end, where you do not know what actually cause it and you need to compare timecodes or so to determine the actual root cause.

cfconrad:
any news?

cfconrad:
or suggestion, like you would add the script within 4 testmodules...

asmorodskyi:
> or suggestion, like you would add the script within 4 testmodules...

no I can not suggest anything :( so I think we have no other choice than follow your idea . sorry that I was holding it so long without good reason :(

PR #14097, opened 2022-01-25
[WIP]SLES Installation with RMT slp discovery

hjluo:
We need to add installation tests with the RMT SLP discovery function.

- Related ticket: https://progress.opensuse.org/issues/101322
- Needles: 
  8.44 local
  1)  local-registration-server: bootloader-20220125
  2) untrusted-ca-cert: welcome-20220125

- Verification run: 
  15sp3
  http://10.161.8.44/tests/147
  http://10.161.8.44/tests/148


tests/installation/bootloader.pm
```diff
@@ -54,7 +54,7 @@ sub run {
     push @params, bootmenu_network_source;
     push @params, bootmenu_remote_target;
     push @params, specific_bootmenu_params;
-    push @params, registration_bootloader_params(utils::VERY_SLOW_TYPING_SPEED);
+    push @params, registration_bootloader_params(utils::VERY_SLOW_TYPING_SPEED) unless get_var('SLP_RMT_INSTALL');
```
lemon-suse, line 57: Why need exclude it for 'SLP_RMT_INSTALL'?

hjluo: If we add https://scc.suse.com it will not pop up the screen for the
slp discovery.  http://10.161.8.44/tests/165#step/welcome/3

hjluo: If not skip it will add https://scc.suse.com and cause the installer to connect directly
 to scc.suse.com. It will not scan for the slp and will not pop up the screen for the
slp discovery.  http://10.161.8.44/tests/165#step/welcome/3

lib/registration.pm
```diff
@@ -791,6 +791,12 @@ sub fill_in_reg_server {
         $regcode = $prdcode if ($prdcode);
     }
 
+    # rmt slp discovery, the smt server address will auto filled
```
lemon-suse, line 794: Add comments here that is to avoid the impact to other openqa test, we will run SLP test on local environment.

lemon-suse:
Please run some installation test in function job group to avoid regression.

rfan1:
@hjluo if you want to implement this tests into openqa, please consider the network configuration for both rmt server and your installing VM. at the same time, I think running this on multi-machine setup should be a better solution.

hjluo:
> @hjluo if you want to implement these tests into openqa, please consider the network configuration for both rmt server and your installing VM. at the same time, I think running this on a multi-machine setup should be a better solution.

yes, first we want to make ti multi-machine, but we need run on 8.44 local env and it doesn't 
support muti-machine. and a fixed VM on 8.44 is also an acceptable solution by far.  most 
concern is that we don't want so much effort for just a simple test{test the pop-up menu of slp}.
thanks a lot.

lib/registration.pm
hjluo: done.

tinawang123 (approved):
LGTM

lemon-suse (approved):

PR #14100, opened 2022-01-25
Implement RPi hardware tests for SLE JeOS

asdil12:
SLE doesn't do DHCP and SSH with a default pw (like openSUSE does)
so this test needs to rely on a preprocessed image.
This image is generated by openQA using qemu backend
via `create_hdd_rpi.yaml`.

The SUT RPi is connected to an as sd card multiplexer
which is also connected to the worker via USB, so the worker
can flash that image onto the sdcard and then switch
the sdcard back to the SUT.
The SUT RPi is power cycled from the worker using a wifi power plug.

https://progress.opensuse.org/issues/105852

### DIFF
--- a/data/generalhw_scripts/power_on_off_shelly.sh
+++ b/data/generalhw_scripts/power_on_off_shelly.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# This script can operate shelly wifi switch plugs
+# See https://shelly-api-docs.shelly.cloud/gen1/#shelly-plug-plugs-relay-0
+
+# Check number of args
+if [ "$#" -ne 2 ] ; then
+	echo "Invalid number of arguments."
+	echo "Usage: $0 IP/HOSTNAME on|off"
+	exit 1
+fi
+
+shelly_ip=$1
+state=$2
+
+res=$(curl -s "http://${shelly_ip}/relay/0?turn=${state}")
+if ! echo "$res" | grep -q "ison" ; then
+	exit 1
+fi
--- a/tests/jeos/prepare_firstboot.pm
+++ b/tests/jeos/prepare_firstboot.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Enable jeos-firstboot as required by openQA testsuite
@@ -11,7 +11,7 @@ use warnings;
 use base "opensusebasetest";
 use testapi;
 use utils 'zypper_call';
-use version_utils 'is_leap';
+use version_utils qw(is_leap is_sle);
 use Utils::Backends;
 
 sub run {
@@ -27,8 +27,12 @@ sub run {
         # Run jeos-firstboot manually and do not reboot as we use SSH
         $reboot_for_jeos_firstboot = 0;
 
-        # Handle default credentials for ssh login
-        $testapi::password = $default_password;
+        if (!is_sle()) {
+            # Handle default credentials for ssh login
+            # On SLE we use an image preprocessed by openQA where the default
+            # $testapi::password was set
+            $testapi::password = $default_password;
+        }
         # 'root-ssh' console will wait for SUT to be reachable from ssh
         select_console('root-ssh');
     }
--- a/tests/jeos/rpi_bluetooth.pm
+++ b/tests/jeos/rpi_bluetooth.pm
@@ -1,0 +1,44 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Check RPi bluetooth: Do a scan and check if we see
+#          the openQA-worker device.
+# Maintainer: qe-core team <qe-core@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+# This requires a discoverable bluetooth device in range:
+#===========================
+#/etc/bluetooth/main.conf:
+#[Policy]
+#AutoEnable=true
+#[General]
+#DiscoverableTimeout = 0
+#Name = openQA-worker
+#===========================
+
+sub run {
+    my ($self) = @_;
+
+    zypper_call 'in bluez';
+    systemctl 'start bluetooth';
+    systemctl 'status bluetooth';
+    assert_script_run 'rfkill list';
+    assert_script_run 'bluetoothctl show';
+    assert_script_run '(echo "power on"; sleep 5; echo "scan on"; sleep 30; echo "devices") | bluetoothctl | tee /dev/stderr | grep openQA-worker';
+    assert_script_run 'bluetoothctl show';
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    $self->save_and_upload_log('dmesg', 'dmesg.log');
+    $self->save_and_upload_log('journalctl -b', 'journal.log');
+}
+
+1;
--- a/tests/jeos/rpi_wifi.pm
+++ b/tests/jeos/rpi_wifi.pm
@@ -1,0 +1,52 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Check RPi Wifi: Connect to openQA-worker wifi and do a ping.
+#          PSK is set via RPI_WIFI_PSK and ping target via RPI_WIFI_WORKER_IP.
+# Maintainer: qe-core team <qe-core@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    my ($self) = @_;
+
+    assert_script_run 'ip l';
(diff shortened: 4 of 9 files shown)


ggardet:
I doubt we have `qcow2` images targeting RPi.

asdil12:
@okurz I already have exchanged some mails with @ggardet but the sle version needs some adoptions:

* sle jeos runs firstboot on boot via hdmi/usb and I can't interface with that (opensuse tw jeos doesn't do this and simply does dhcp and listens on ssh)
* I have a different model of wifi plug
* I *plan* to use a real serial cable instead of using ssh-serial

My approach on addressing the first issue is publishing an image from a `create_hdd` job in openQA that already completed firstboot and use that one then: https://openqa.suse.de/tests/8026617

data/generalhw_scripts/power_on_off_shelly.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This script can operate shelly wifi switch plugs
+# See https://shelly-api-docs.shelly.cloud/gen1/#shelly-plug-plugs-relay-0
+
+shelly_ip=$1
+state=$2
+
```
ggardet, line 8: I would check the number of args `$#` here to fail early if needed.

asdil12:
This should now be in a state that could be merged already.
VR: http://10.131.0.32/tests/43

The documentation for worker configuration is maintained here:
https://confluence.suse.com/display/~dheidler/Hardware+Automation

ggardet:
> This should now be in a state that could be merged already. VR: http://10.131.0.32/tests/43
> 
> The documentation for worker configuration is maintained here: https://confluence.suse.com/display/~dheidler/Hardware+Automation

Unfortunately, none of the links above are publicly accessible.

WiFi and Bluetooth tests are good additions which I will add for openSUSE Tumbleweed at some point.

tests/jeos/rpi_wifi.pm
```diff
@@ -0,0 +1,39 @@
+# SUSE's openQA tests
+#
+# Copyright 2015-2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Check RPi Wifi
```
ggardet, line 6: Could you give more details here, please?
Especially since it does not try to connect to any WiFi network.

asdil12: The idea is that the worker host (also a rpi) will span up a software wifi ap that the shelly powerplugs and the sut during this test module can connect to.
This is not yet implemented in the test as I'm busy setting up an actual rpi based worker right now (I was using my laptop as a worker before) so that I have a working wifi to connect to.

tests/jeos/rpi_bluetooth.pm
```diff
@@ -0,0 +1,45 @@
+# SUSE's openQA tests
+#
+# Copyright 2015-2021 SUSE LLC
```
okurz, line 3:
```suggestion
# Copyright SUSE LLC
```

see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md?plain=1#L49

The same goes for other files added or changed in this pull request

tests/jeos/prepare_firstboot.pm
```diff
@@ -11,7 +11,7 @@ use warnings;
 use base "opensusebasetest";
 use testapi;
 use utils 'zypper_call';
-use version_utils 'is_leap';
+use version_utils;
```
okurz, line 14: please continue to use explicit imports. So likely you need the following here

```suggestion
use version_utils qw(is_leap is_sle);
```

tests/jeos/rpi_bluetooth.pm
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Check RPi Wifi
+# Maintainer: qsf-u team <qa-core@suse.de>
```
okurz, line 7: I guess that's outdated and should say "SUSE QE-Core"?

tests/jeos/rpi_bluetooth.pm
```diff
@@ -0,0 +1,45 @@
...
+
+sub run {
+    my ($self) = @_;
+
+    zypper_call "in bluez";
+    systemctl "start bluetooth";
+    systemctl "status bluetooth";
+    assert_script_run "rfkill list";
+    assert_script_run "bluetoothctl show";
+    assert_script_run '(echo "power on"; sleep 5; echo "scan on"; sleep 15; echo "devices") | bluetoothctl | tee /dev/stderr | grep openQA-worker';
```
okurz, line 35: You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

tests/jeos/rpi_wifi.pm
```diff
@@ -0,0 +1,39 @@
...
+    assert_script_run 'ip l';
+    assert_script_run 'rfkill list';
+    assert_script_run 'ip l set wlan0 up';
+    # This would be a more modern approach but
+    # iw is not installed by default:
+    # iw dev wlan0 scan | grep -e "^\(BSS\|\sSSID:\)"
+    assert_script_run 'iwlist wlan0 scan | grep -e "^\s*\(Cell\|Frequency\|Quality\|ESSID\)"';
+    assert_script_run 'ip l set wlan0 down';
+
+    #TODO: Actually connect to openQA-worker wifi
```
okurz, line 28: Don't add TODOs as likely this will stay here forever. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit or put it into tickets which are more suitable for tracking tasks

tests/jeos/rpi_wifi.pm
```diff
@@ -0,0 +1,39 @@
...
+
+    select_console 'root-console';
+    enter_cmd "jeos-config raspberrywifi";
+    assert_screen 'jeos-rpiwifi-select';
+    send_key 'esc';
+    assert_screen 'jeos-rpiwifi-really-quit';
+    send_key 'y';
+    enter_cmd "clear";
```
okurz, line 36: You are using a mixed style of single and double quotes. I recommend to consistently use single quotes except where you want to evaluate variables

tests/jeos/rpi_wifi.pm
```diff
@@ -0,0 +1,39 @@
...
+
+# Summary: Check RPi Wifi
+# Maintainer: qsf-u team <qe-core@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+use Utils::Architectures;
+use Utils::Backends;
```
okurz, line 14: I don't think you use these anywhere

okurz (changes requested):
please see inline comments.

Also, the commit message should have much more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a ticket, a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

tests/jeos/rpi_bluetooth.pm
asdil12: I know you don't like `sleep()` and I understand why, but this doesn't help if the SUT didn't get enough time to scan for bluetooth devices.

okurz (commented):
Code looks good. Now please provide details about your testing approach and results in the commit message and description

asdil12:
The full wifi connection is now covered as well.

asdil12:
VR for wifi: http://10.131.0.32/tests/60

tests/jeos/rpi_wifi.pm
ggardet: Could you be more verbose than "Check RPi WiFi" please?
Also, please add a note about required vars: `RPI_WIFI_PSK` and `RPI_WIFI_WORKER_IP` AFAICS.

asdil12: done

ggardet (approved):
I cannot check the VR as it is not public, but code looks good.

okurz (commented):
I can not access https://10.131.0.32/tests/60 as well. Please provide more details about the testing procedure in the git commit message

asdil12:
@okurz http://10.163.27.222/tests/60 should work for you

okurz (approved):
Thank you, looks good so far I can see

PR #14105, opened 2022-01-26
Use get_var to replace hardcode password

XGWang0:
Replace hard code password variable by passing from web front-end

- Related ticket: N/A
- Needles: N/A
- Verification run: N/A

### DIFF
--- a/lib/proxymode.pm
+++ b/lib/proxymode.pm
@@ -16,8 +16,8 @@ our $SLAVE_SERIALDEV = 'proxyserial';
 
 sub switch_power {
     my ($ipmi_machine, $ipmi_user, $ipmi_pass, $ipmi_status) = @_;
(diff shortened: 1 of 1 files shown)


asdil12:
Do you have a ticket this relates to or an example testrun where it can be seen that the required vars are set?

Also maybe setting `ADMIN` as default would be useful:
```perl
get_var('IPMI_PASSWORD', 'ADMIN');

XGWang0:
> Do you have a ticket this relates to or an example testrun where it can be seen that the required vars are set?
> 
> Also maybe setting `ADMIN` as default would be useful:
> 
> ```perl
> get_var('IPMI_PASSWORD', 'ADMIN');
> ```

We don't want to use 'ADMIN' as default parameters is to avoid to leak security related info.

alice-suse:
LGTM

PR #14110, opened 2022-01-26
Fail on empty repositories in publiccloud

grisu48:
Fail if there are no repositories defined in register_system step on
publiccloud. This is necessary to catch registration errors much
earlier.

- Related ticket: https://progress.opensuse.org/issues/105419
- Verification run: [15-SP3 Azure BYOS](http://duck-norris.qam.suse.de/t8048) | [15-SP3 ARM BYOS](http://duck-norris.qam.suse.de/t8049) | [15-SP3 GCE](http://duck-norris.qam.suse.de/t8050) | [12-SP5 EC2](http://duck-norris.qam.suse.de/t8051)


pdostal (approved):
VR looks good :+1:

PR #14111, opened 2022-01-26
Simplify container diff check

grisu48:
Remove unnecessary container diff check to allow empty container updates
as well.

- Related ticket: https://progress.opensuse.org/issues/105067
- Verification run: https://openqa.opensuse.org/tests/2161400 and https://openqa.suse.de/tests/8041978

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -17,7 +17,6 @@ use utils;
 use strict;
 use warnings;
 use version_utils;
-use version;
 use Utils::Architectures;
 use containers::utils;
 use containers::common qw(test_container_image is_unreleased_sle);
@@ -214,27 +213,6 @@ sub test_zypper_on_container {
     $runtime->remove_container('refreshed');
     $runtime->run_container($image, name => "refreshed-image", cmd => "zypper -nv ref", timeout => 300, retry => 3, delay => 60);
     record_info "The End", "zypper test completed";
-}
-
-sub ensure_container_rpm_updates {
-    my $diff_file = shift;
-    my $regex2match = qr/^-(?<package>[^\s]+)\s+(\d+\.?\d+?)+-(?P<update_version>\d+\.?\d+?\.\d+\b).*\s+(\d+\.?\d+?)+-(?P<stable_version>\d+\.?\d+?\.\d+\b)/;
-    my $regex2zerorpm = qr/^Version differences: None$/;
-    my $context = script_output "cat $diff_file";
-    open(my $data, '<', \$context) or die "problem with $diff_file argument", $!;
-    while (my $line = <$data>) {
-        if ($line =~ $regex2match) {
-            # Use of Dotted-Decimal-Versions. Do not remove 'v' prefix
-            my $updated_v = version->parse("v$+{update_version}");
-            my $stable_v = version->parse("v$+{stable_version}");
-            record_info("checking... $+{package}", "$+{package} $stable_v to $updated_v");
-            die "$+{package} $stable_v is not updated to $updated_v" unless ($updated_v > $stable_v);
-        } elsif ($line =~ $regex2zerorpm) {
-            record_info("No Update found", "no updates found between rpm versions");
-            last;
-        }
-    }
-    close($data);
 }
 
 sub exec_on_container {
--- a/tests/containers/container_diff.pm
+++ b/tests/containers/container_diff.pm
@@ -21,16 +21,15 @@ sub run {
 
     zypper_call("install container-diff") if (script_run("which container-diff") != 0);
 
-    my ($untested_images, $released_images) = get_suse_container_urls();
+    my ($testing_images, $released_images) = get_suse_container_urls();
     # container-diff
-    for my $i (@{$untested_images}) {
-        my $image_file = $untested_images->[$i] =~ s/\/|:/-/gr;
+    for my $i (@{$testing_images}) {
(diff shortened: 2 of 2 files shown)


Vogtinator (changes requested):
Container updates can be empty

tests/containers/container_diff.pm
```diff
@@ -21,16 +21,39 @@ sub run {
...
+        #
+        # These entries have been added to registry.opensuse.org/opensuse/tumbleweed: None
+        #
+        # These entries have been deleted from registry.opensuse.org/opensuse/tumbleweed: None
+        #
+        # These entries have been changed between registry.opensuse.org/opensuse/tumbleweed and registry.opensuse.org/opensuse/tumbleweed: None
+        record_info("Empty container update (no changes)") if (script_run("grep -v 'FILE' $container_diff_results", 300) != 0);
+
+        # Ensure there are no pending updates in the container
+        assert_script_run("docker run -t --rm $testing_images->[$i] /bin/bash -c 'zypper ref && zypper -n up -D | grep \"Nothing to do.\"'", timeout => 300);
```
Vogtinator, line 56: IMO this should be done in a different module, alongside the other zypper tests

tests/containers/container_diff.pm
```diff
@@ -21,16 +21,39 @@ sub run {
...
+        my $image_file = $testing_images->[$i] =~ s/\/|:/-/gr;
         my $container_diff_results = "/tmp/container-diff-$image_file.txt";
-        assert_script_run("docker pull $untested_images->[$i]", 360);
+        assert_script_run("docker pull $testing_images->[$i]", 360);
         assert_script_run("docker pull $released_images->[$i]", 360);
-        assert_script_run("container-diff diff daemon://$untested_images->[$i] daemon://$released_images->[$i] --type=rpm --type=file --type=size > $container_diff_results", 300);
+        assert_script_run("container-diff diff daemon://$testing_images->[$i] daemon://$released_images->[$i] --type=rpm --type=file --type=size > $container_diff_results", 300);
         upload_logs("$container_diff_results");
-        ensure_container_rpm_updates("$container_diff_results");
+        # Check if the new container is not an empty update
```
Vogtinator, line 33: This doesn't really do anything which isn't already done just above

lib/containers/container_images.pm
```diff
@@ -216,27 +216,6 @@ sub test_zypper_on_container {
...
-sub ensure_container_rpm_updates {
-    my $diff_file = shift;
-    my $regex2match = qr/^-(?<package>[^\s]+)\s+(\d+\.?\d+?)+-(?P<update_version>\d+\.?\d+?\.\d+\b).*\s+(\d+\.?\d+?)+-(?P<stable_version>\d+\.?\d+?\.\d+\b)/;
-    my $regex2zerorpm = qr/^Version differences: None$/;
-    my $context = script_output "cat $diff_file";
-    open(my $data, '<', \$context) or die "problem with $diff_file argument", $!;
-    while (my $line = <$data>) {
-        if ($line =~ $regex2match) {
-            # Use of Dotted-Decimal-Versions. Do not remove 'v' prefix
-            my $updated_v = version->parse("v$+{update_version}");
```
Vogtinator, line 228: Can probably remove the version import at the top now

tests/containers/container_diff.pm
grisu48: @jlausuch You have a better overview if this check should be moved elsewhere, I was only fixing the occurring failure. Move it or leave it?

tests/containers/container_diff.pm
grisu48: It prints an empty container update later on as info box (See line 53). However, if @jlausuch agrees that this check is not needed we can remove this whole block.

pdostal (approved):

jlausuch:
After some thinking, my opinion is that we should remove this module from image tests, because there is no point to check differences if the new container image build doesn't have any differences (which could be the case).

So, since `container_diff` is a package provided by containers module, why not making this test as "feature" test, being `container_diff` the main FUT (Feature Under Test, or however it's called), and not focusing in comparing _released images_ vs _totest images_? This way, we would treat it as just another feature from containers module, like docker, podman, etc.

If you all agree, we could disable this test from container image jobs and re-work it a bit. Let me know your opinions.

Vogtinator:
Currently the output from `container-diff` is helpful as a reference to see what changed if something broke in the image. So I'd still like it used in container image tests.

Vogtinator:
Currently the output from `container-diff` is helpful as a reference to see what changed if something broke in the image. So I'd still like it used in container image tests.

pdostal:
I like this PR as it is now - simplified but still useful for our images.

jlausuch:
Ok, that makes sense, for troubleshooting purposes it's always good to have this info.

Anyway, I still think this package should be "also" tested as part of container host tests. But this is for another ticket/PR :)

tests/containers/container_diff.pm
jlausuch: Yes, this looks more like pure "image test" ensuring that a fresh built image is up to date completely and not a container-diff thingy.. I would remove it from here.

grisu48: Removed it

tests/containers/container_diff.pm
grisu48: Removed it

grisu48:
Okok, I add it again.

Vogtinator (approved):

PR #14115, opened 2022-01-27
Fix userdel failed after user logout

lemon-suse:
We need wait a bit more time for the user related process to quit even after logout the user.

- Related ticket: https://progress.opensuse.org/issues/105402
- Needles: N/A
- Verification run: 
   https://openqa.nue.suse.com/tests/8039454#step/check_upgraded_service/78

### DIFF
--- a/lib/services/users.pm
+++ b/lib/services/users.pm
@@ -134,6 +134,8 @@ sub restore_passwd {
 
 # remove test user
 sub remove_test_user {
+    # We need wait a bit more time for the user related process to quit even after logout the user.
+    assert_script_run('for ((i=5; i>0; i--)) do if (! ps -e -u test > /dev/null 2>&1); then break; else sleep 3; fi done');
     assert_script_run("userdel -r $newUser");
 }
 


lilyeyes (approved):
I think "sleep 1" is better but LGTM.

lib/services/users.pm
```diff
@@ -134,6 +134,8 @@ sub restore_passwd {
 
 # remove test user
 sub remove_test_user {
+    # We need wait a bit more time for the user related process to quit even after logout the user.
+    assert_script_run('for ((i=5; i>0; i--)) do if (! ps -e -u test > /dev/null); then break; else sleep 3; continue; fi done');
     assert_script_run("userdel -r $newUser");
```
lilyeyes, line 139: May be "userdel -rf $testuser" can work well too.

lib/services/users.pm
```diff
@@ -134,6 +134,8 @@ sub restore_passwd {
 
 # remove test user
 sub remove_test_user {
+    # We need wait a bit more time for the user related process to quit even after logout the user.
+    assert_script_run('for ((i=5; i>0; i--)) do if (! ps -e -u test > /dev/null); then break; else sleep 3; continue; fi done');
```
coolgw, line 138: for ((i=5; i>0; i--)) do if (! ps -e -u test >/dev/null 2>&1); then break; else sleep 3; fi done

lemon-suse: Thanks, updated code.

lib/services/users.pm
lemon-suse: For us, on x11, we need wait the process quit while not force kill it, else it may cause unexpected failure such as console issue.

coolgw (approved):

lemon-suse:
> I think "sleep 1" is better but LGTM.

Maybe, I just don't want to loop this too much. :)

PR #14117, opened 2022-01-27
 Distinguish different baseline files from arches in CC audit-test

Amrysliu:
For now in test cases 'syscall' and 'misc', the baseline files are not
the same in different arches. So we need to distinguish them.

Related: https://progress.opensuse.org/issues/105352
Verify run: https://openqa.suse.de/tests/8040938#
     for syscalls: https://openqa.suse.de/tests/8040206#

### DIFF
--- a/lib/audit_test.pm
+++ b/lib/audit_test.pm
@@ -73,12 +73,10 @@ sub upload_audit_test_logs {
     my ($testcase) = @_;
     upload_logs("$current_file");
 
-    # For syscall test case, different arch has different baseline file,
-    # so we need to upload the correct file
-    if ($testcase eq 'syscalls') {
-        $baseline_file = $baseline_file . '.' . get_var('ARCH');
-    }
-    else {
+    # Base line file is ARCH specific in some cases,
+    # so we need to upload the correct file for each platform.
+    $baseline_file = 'baseline_run.log.' . get_var('ARCH');
+    if (script_run("test -e $baseline_file") != 0) {
         $baseline_file = "baseline_run.log";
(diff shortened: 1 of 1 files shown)


lib/audit_test.pm
```diff
@@ -73,12 +73,10 @@ sub upload_audit_test_logs {
     my ($testcase) = @_;
     upload_logs("$current_file");
 
-    # For syscall test case, different arch has different baseline file,
+    # For some test cases, different arch has different baseline file,
     # so we need to upload the correct file
```
rfan1, line 77: Base line file is ARCH specific, so we need upload the correct file the each platform

lilyeyes (approved):
LGTM.

lib/audit_test.pm
```diff
@@ -73,12 +73,10 @@ sub upload_audit_test_logs {
...
-    if ($testcase eq 'syscalls') {
```
lilyeyes, line 78: Since you touched syscalls could you please offer the verification runs for syscalls.

lilyeyes (approved):

jouyingbin:
How about amending the subject to 
**Distinguish different baseline files from arches in CC audit-test**

lib/audit_test.pm
Amrysliu: I did the verification runs for syscalls, but didn't paste the result.  Pasted it in the description.

rfan1 (approved):
LGTM

jouyingbin (approved):
LGTM

PR #14118, opened 2022-01-27
Disable soft fail record for 390 auth popup case

coolgw:
Disable soft fail record for 390 auth popup case, base feedback on bsc#1192992, if version < 15sp4, the auth popup window is normal behavior.

- Related ticket: https://progress.opensuse.org/issues/105178
- Needles: na
- Verification run: http://openqa.suse.de/tests/8039597#step/first_boot/6

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -193,7 +193,12 @@ sub handle_additional_polkit_windows {
         # for S390x testing, since they are not using qemu built-in vnc, it is
         # expected that polkit authentication window can open for first time login.
         # see bsc#1177446 for more information.
-        record_soft_failure 'bsc#1192992 - multiple authentication due to repositories refresh on s390x';
+        # Base latest feedback of bsc#1192992,authentication should never open if is_sle  >= 15SP4
+        if (is_sle('>=15-sp4')) {
+            record_soft_failure 'bsc#1192992 - authentication should never open if is_sle >= 15SP4';
+        } else {
+            record_info('authentication open for first time login');
+        }
         wait_still_screen(5);
         my $counter = 5;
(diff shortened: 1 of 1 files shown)


dzedro:
You should pick < or > :)

lemon-suse:
As my understanding , the description should be updated from 'version <=15sp4' to 'version < 15sp4'

lemon-suse (approved):

dzedro:
> As my understanding , the description should be updated from 'version <=15sp4' to 'version < 15sp4'

Yes, please change it to older versions than 15 SP4, this failure is still happening on older versions. `is_sle('<15-sp4')`

dzedro:
It didn't even come to the `handle_additional_polkit_windows` because authentication-required-user-settings didn't repeat

coolgw:
> As my understanding , the description should be updated from 'version <=15sp4' to 'version < 15sp4'

yes  updated

coolgw:
> > As my understanding , the description should be updated from 'version <=15sp4' to 'version < 15sp4'
> 
> Yes, please change the code to older versions than 15 SP4, this failure is still happening on older versions. `is_sle('<15-sp4')` I don't understand how there is no soft fail in VR with `is_sle('>=15-sp4')`
[GW]: Base software feedback, if is_sle(>=15sp4), it should never popup auth windows，if is_sle(<15sp4) then auth window should pop up for first login.
> 
> https://openqa.suse.de/tests/8040928 15 SP2
[GW]: it's really strange why record_soft not print out


>

dzedro (approved):
As it's expected on older versions I would not use record_soft_fail on the counter, it will just spam the bug.

PR #14119, opened 2022-01-27
Use name & device from nmcli * show output

dzedro:
- Related ticket: https://progress.opensuse.org/issues/105295
- Verification run:
TW
https://openqa.opensuse.org/tests/overview?build=nmcli
https://dzedro.suse.cz/tests/overview?build=20220126
15 SP3
https://dzedro.suse.cz/tests/20675

### DIFF
--- a/tests/network/setup_multimachine.pm
+++ b/tests/network/setup_multimachine.pm
@@ -23,8 +23,12 @@ sub is_networkmanager {
 sub run {
     my ($self) = @_;
     my $hostname = get_var('HOSTNAME');
+    my ($nm_id, $device);
     select_console 'root-console';
-    my $nm_id = is_sle('15-sp4+') ? 'eth0' : 'Wired connection 1';
+    if (is_networkmanager) {
+        $nm_id = script_output('nmcli -t -f NAME c');
+        $device = script_output('nmcli -t -f DEVICE c');
+    }
 
     # Do not use external DNS for our internal hostnames
     assert_script_run('echo "10.0.2.101 server master" >> /etc/hosts');
@@ -39,7 +43,7 @@ sub run {
         setup_static_mm_network('10.0.2.101/24');
 
         if (is_networkmanager) {
-            assert_script_run "nmcli connection modify '$nm_id' ifname 'eth0' ip4 '10.0.2.101/24' gw4 10.0.2.2 ipv4.method manual ";
+            assert_script_run "nmcli connection modify '$nm_id' ifname '$device' ip4 '10.0.2.101/24' gw4 10.0.2.2 ipv4.method manual ";
             assert_script_run "nmcli connection down '$nm_id'";
             assert_script_run "nmcli connection up '$nm_id'";
         }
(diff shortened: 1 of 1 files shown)


Vogtinator:
Why not fix `setup_static_mm_network` instead?

tests/network/setup_multimachine.pm
```diff
@@ -24,7 +24,11 @@ sub run {
     my ($self) = @_;
     my $hostname = get_var('HOSTNAME');
     select_console 'root-console';
-    my $nm_id = is_sle('15-sp4+') ? 'eth0' : 'Wired connection 1';
+    # delimiter is 2 spaces, because name can contain space, between columns are at least 2 spaces
+    if (is_networkmanager) {
+        my $nm_id = script_output(q(nmcli connection show|awk -F'  ' '/ethernet/ {print$1}'));
+        my $device = script_output(q(nmcli connection show|awk -F '  ' '/ethernet/ {print$4}'));
```
DimStar77, line 30: At least this one does not work on my connection name, which does happen to have a space in the name

```
> nmcli connection show|awk -F "  "  '/ethernet/ {print $4}'

```

what DOES work though is:

```

> nmcli connection show| awk  '/ethernet/ {print $NF}'
enp0s25
```

(NF being number of fields - and in this case we care for the last field, so this makes things easy)

Vogtinator:
By using `-t` combined with a list it can be parsed more easily:

```
> nmcli -t -f DEVICE,UUID c
ens4:da3a315c-4591-3892-b61d-9555cab60a5f
ens3:436be304-6e5b-319b-83fb-64cc2b09882d
> nmcli -t -f DEVICE,UUID c | awk -F: '$1 == "ens3" { print $2 }'
436be304-6e5b-319b-83fb-64cc2b09882d
```

tests/network/setup_multimachine.pm
dzedro: This works, it can be done differently when it does not work.
Problem is not `device` but `name` on SLED.
image

dzedro:
> By using `-t` combined with a list of fields it can be parsed more easily:
> 
> ```
> > nmcli -t -f DEVICE,UUID c
> ens4:da3a315c-4591-3892-b61d-9555cab60a5f
> ens3:436be304-6e5b-319b-83fb-64cc2b09882d
> > nmcli -t -f DEVICE,UUID c | awk -F: '$1 == "ens3" { print $2 }'
> 436be304-6e5b-319b-83fb-64cc2b09882d
> ```

cool, will check it out and probably use it.

tests/network/setup_multimachine.pm
dzedro: I will use arguments Fabian mentioned.

dzedro:
> Why not fix `setup_static_mm_network` instead?

networkmanager was added here, it can/should be part of setup_static_mm_network.
First I just want to fix the issue with device name, then it can be moved.

dzedro:
I guess this is fine, but there is some issue with DNS, but I don't think it's related to MM setup or this PR.
Is it some ingra issue ? I don't see difference between https://openqa.opensuse.org/tests/2161490 & https://openqa.opensuse.org/tests/2161472

Vogtinator:
> I guess this is fine, but there is some issue with DNS, but I don't think it's related to MM setup or this PR. Is it some infra issue ? I don't see difference between https://openqa.opensuse.org/tests/2161490 & https://openqa.opensuse.org/tests/2161472

Looks like DNS is not set up at all. The first test doesn't need it, but the second fails due to that.

dzedro:
I rerun VRs and they passed, next week I will add or create new PR to move the NM configuration.

Vogtinator (approved):

dzedro:
jfyi, I'm working on the move, part of it will be also fix for DNS issue. Will do it withing the same poo#

Vogtinator:
I filed https://bugzilla.opensuse.org/show_bug.cgi?id=1196606 for the DNS issue.

FWICT, it can be worked around by either using NM to set the static DNS as well (`ipv4.dns`) or manually removing the old DHCP lease.

PR #14120, opened 2022-01-27
SLEM5.2: SelfInstall kick off

mloviska:
SLEM5.2 introduces new type of image deployment in form of
selfinstallable iso. Image deployment is done via [KIWI](https://github.com/OSInside/kiwi/tree/master/dracut/modules.d)

- [Include the new SLE Micro Selfinstall ISO images](https://progress.opensuse.org/issues/105025)


products/sle-micro/main.pm
```diff
@@ -9,13 +9,20 @@ BEGIN {
...
 use main_common;
+use version_utils ':VERSION';
 
 init_main();
 
 my $distri = testapi::get_required_var('CASEDIR') . '/lib/susedistribution.pm';
 require $distri;
 testapi::set_distribution(susedistribution->new());
 
+if (!is_jeos) {
```
jlausuch, line 20: why?

mloviska: Maybe this is not strictly needed, but we also do not load jeos needles at all.

jlausuch:
Looks promising!

mloviska:
Managed to make this work on legacy boot. https://openqa.suse.de/tests/8076510#

Unfortunately, it seems that there is something wrong in the way openQA handles uefi vars that requires more investigation

products/sle-micro/main.pm
```diff
@@ -9,14 +9,14 @@ BEGIN {
 use utils;
 use testapi;
 use main_common;
+use version_utils ':VERSION';
```
jlausuch, line 12: leftover?

lib/version_utils.pm
```diff
@@ -232,6 +233,14 @@ sub is_sle_micro {
     return check_var('DISTRI', 'sle-micro');
 }
 
+=head2 is_selfinstall
+
+Check if SLEM is in flavor of self installable iso
+=cut
+sub is_selfinstall {
+    return is_sle_micro && (get_var('FLAVOR') =~ /selfinstall/i);
```
jlausuch, line 241: shall we make this more generic and omit `is_sle_micro` ?

products/sle-micro/main.pm
mloviska: yup, it definitely, thanks :)

lib/version_utils.pm
jlausuch: I would go generic, there is no other flavor like this so far and I don't think there are plans. Anyway, if there is something else besides sle micro, it's a valid function for it as well (e.g. microOS?)

tests/transactional/host_config.pm
```diff
@@ -14,15 +14,18 @@ use strict;
...
 
     # GRUB Configuration
     my $disable_grub_timeout = get_var('DISABLE_GRUB_TIMEOUT');
     my $extrabootparams = get_var('EXTRABOOTPARAMS');
-    assert_script_run("sed -i 's/GRUB_TIMEOUT=.*/GRUB_TIMEOUT=-1/' /etc/default/grub") if $disable_grub_timeout;
-    assert_script_run("sed -i 's/^GRUB_CMDLINE_LINUX_DEFAULT=\"[^\"]*/& $extrabootparams/' /etc/default/grub") if $extrabootparams;
+    change_grub_config('=\"[^\"]*', "& $extrabootparams", 'GRUB_CMDLINE_LINUX_DEFAULT') if $extrabootparams;
+    change_grub_config('=.*', '=-1', 'GRUB_TIMEOUT') if $disable_grub_timeout;
+    change_grub_config('=.*', '=1024x768', 'GRUB_GFXMODE');
+    change_grub_config('^#', '', 'GRUB_GFXMODE');
```
mloviska, line 28: I thought that setting gfx mode will be helpful and aligned the behaviour with the installation. It was wrong, grub2 resolution is not full screen in DVD installations. I can either remove it, or we can keep it as it looks nicer and yet it is a valid test. But no objections if we go for dropping it.

mloviska:
VRs:

* https://openqa.suse.de/tests/8105058#
* https://openqa.suse.de/tests/8105087#

mloviska:
> Can I merge this and go first with legacy and worry about uefi later on?

Sure, I just need your opinion regarding GFX payload. :) https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14120#discussion_r799384847

jlausuch (approved):

PR #14125, opened 2022-01-27
wpa_supplicant: Poll connection state on WPA_PSK

cfconrad:
This is a fix for https://bugzilla.suse.com/show_bug.cgi?id=1194811 as
connection to WPA_PSK takes slightly longer on ppc64le.

Replace fix sleep's with polling for state.

Change the hostapd.conf to use only one phy configuration, as it
isn't supported to use different phy settings via multi-bss.

Fix test if network.service isn't wicked.

- Verification run: https://openqa.opensuse.org/t2162340

### DIFF
--- a/data/wpa_supplicant/hostapd.conf
+++ b/data/wpa_supplicant/hostapd.conf
@@ -1,23 +1,23 @@
 interface=wlan0
 driver=nl80211
 country_code=DE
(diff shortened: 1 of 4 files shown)


DimStar77:
Seems ok to me

PR #14130, opened 2022-01-28
Implement SSH host key handling based on libyui

rakoenig:
Testsuites are mirgrated from main.pm based execution to YAML based schedule.

- Related ticket: https://progress.opensuse.org/issues/102317
- Needles: not needed
- Verification run: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=rakoenig_ssh_keys&groupid=96

### DIFF
--- a/tests/installation/ssh_host_key_settings/verify_ssh_key_imported.pm
+++ b/tests/installation/ssh_host_key_settings/verify_ssh_key_imported.pm
@@ -1,0 +1,31 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: verify that SSH host key got imported.
+#
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    x11_start_program('xterm -geometry 150x45+5+5', target_match => 'xterm');
+    become_root;
+    script_run 'cd /etc/ssh';
+    enter_cmd "cat /etc/ssh/ssh_host_key | tee /dev/$serialdev";
+    wait_serial "SSHHOSTKEYFILE", 5 || die "/etc/ssh/ssh_host_key content doesn't match";
+    enter_cmd "cat /etc/ssh/ssh_host_key.pub | tee /dev/$serialdev";
+    wait_serial "SSHHOSTPUBKEYFILE", 5 || die "/etc/ssh/ssh_host_key.pub content doesn't match";
+    script_run "md5sum * | tee /dev/$serialdev";
+    enter_cmd "killall xterm";
+}
+
+sub test_flags {
+    return {fatal => 1};
+}
+
+1;
--- a/tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
+++ b/tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
@@ -1,0 +1,31 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: verify that SSH host key dit not got imported.
+#
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    x11_start_program('xterm -geometry 150x45+5+5', target_match => 'xterm');
+    become_root;
+    script_run 'cd /etc/ssh';
+    enter_cmd "cat /etc/ssh/ssh_host_key | tee /dev/$serialdev";
+    wait_serial "SSHHOSTKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key content does match";
+    enter_cmd "cat /etc/ssh/ssh_host_key.pub | tee /dev/$serialdev";
+    wait_serial "SSHHOSTPUBKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key.pub content does match";
+    script_run "md5sum * | tee /dev/$serialdev";
+    enter_cmd "killall xterm";
+}
+
+sub test_flags {
+    return {fatal => 1};
+}
+
+1;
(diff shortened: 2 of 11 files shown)


tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
```diff
@@ -0,0 +1,31 @@
...
+use testapi;
+
+sub run {
+    x11_start_program('xterm -geometry 150x45+5+5', target_match => 'xterm');
+    become_root;
+    script_run 'cd /etc/ssh';
+    enter_cmd "cat /etc/ssh/ssh_host_key | tee /dev/$serialdev";
+    wait_serial "SSHHOSTKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key content does match";
+    enter_cmd "cat /etc/ssh/ssh_host_key.pub | tee /dev/$serialdev";
+    wait_serial "SSHHOSTPUBKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key.pub content does match";
```
ge0r, line 22: that's a neat trick and pretty original for checking string to be honest! I would suggest `assert_matches` to die in case of string mismatch, in combination with a script_output("cat /etc/ssh/ssh_host_key") . However your solution is so cool that I m fine with it as well : D

tests/installation/ssh_host_key_settings/verify_ssh_key_imported.pm
```diff
@@ -0,0 +1,31 @@
...
+use testapi;
+
+sub run {
+    x11_start_program('xterm -geometry 150x45+5+5', target_match => 'xterm');
+    become_root;
+    script_run 'cd /etc/ssh';
+    enter_cmd "cat /etc/ssh/ssh_host_key | tee /dev/$serialdev";
+    wait_serial "SSHHOSTKEYFILE", 5 || die "/etc/ssh/ssh_host_key content doesn't match";
+    enter_cmd "cat /etc/ssh/ssh_host_key.pub | tee /dev/$serialdev";
+    wait_serial "SSHHOSTPUBKEYFILE", 5 || die "/etc/ssh/ssh_host_key.pub content doesn't match";
```
ge0r, line 22: same `assert_matches` recommendation here for string checking

tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
```diff
@@ -0,0 +1,31 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    x11_start_program('xterm -geometry 150x45+5+5', target_match => 'xterm');
+    become_root;
```
ge0r, line 17: You could use `start_root_shell_in_xterm();` from x11utils.pm

ge0r (commented):
Great PR and I found your way of verifying of string checking really original!

tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
```diff
@@ -0,0 +1,31 @@
...
+    script_run 'cd /etc/ssh';
+    enter_cmd "cat /etc/ssh/ssh_host_key | tee /dev/$serialdev";
+    wait_serial "SSHHOSTKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key content does match";
+    enter_cmd "cat /etc/ssh/ssh_host_key.pub | tee /dev/$serialdev";
+    wait_serial "SSHHOSTPUBKEYFILE", 5, 1 || die "/etc/ssh/ssh_host_key.pub content does match";
+    script_run "md5sum * | tee /dev/$serialdev";
```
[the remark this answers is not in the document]
rakoenig: I know, I was also puzzled first. You need to see this in the context of the test suite. There is a test at the beginning that sets the files for the SSH host keys. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/c8223c263e0b18c8bafd7fe7345c69d3ae9dbf2e/tests/x11/ssh_key_check.pm#L16
Since we can't verify the existence of real SSH keys the test module just writes plain text in those files and this checks for that plain text.

tests/installation/ssh_host_key_settings/verify_ssh_key_not_imported.pm
rakoenig: Don't praise me, praise the ones that created https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/c8223c263e0b18c8bafd7fe7345c69d3ae9dbf2e/tests/x11/ssh_key_verify.pm

My contribution here was just to split up that test module into two test modules, so that we're no longer depending on variables like  SSH_KEY_IMPORT or SSH_KEY_DO_NOT_IMPORT. So I didn't look at the internals, just split up the two branches.

ge0r (approved):

jknphy:
is there any difference in this scenario if we don't import the configuration? https://openqa.suse.de/tests/8043828#step/import_ssh_host_keys/3 I was about to propose to change the name to reflect the two actions in the screen, but maybe is that the configuration part is not even needed, could please check that? perhaps we could save a step.

rakoenig:
> Is there any difference in this scenario if we don't import the configuration? https://openqa.suse.de/tests/8043828#step/import_ssh_host_keys/3 I was about to propose to change the name to reflect the two actions in the screen, but maybe is that the configuration part is not even needed, could please check that? perhaps we could save one step.

I checked the scenarios: 
Previous to import_ssh_keys or do_not_import_ssh_keys the Installation Settings screen says

SSH Host key will be copied from SUSE Linux Enterprise Server 12 SP2

So if we don't want to import we have to deselect this. 
When importing then our code also selects the "Import SSH Configuration" checkbox which is not selected by default.

jknphy:
yes, my question is if we don't select that `Import SSH Configuration` if the scenario passes? looks like we are doing something at the beginning only related with keys so I don't know what is about that ssh configuration, perhaps it is not even needed. make sense?

rakoenig:
> yes, my question is if we don't select that `Import SSH Configuration` if the scenario passes? looks like we are doing something at the beginning only related with keys so I don't know what is about that ssh configuration, perhaps it is not even needed. make sense?

I removed the line that sets the "import config" checkbox and the test still passes.

PR #14131, opened 2022-01-28
Create a generic k8s test for all the providers

ilausuch:
We don't need 3 different tests because once we use kubectl all
the providers should work in the same work. This PR converts the
EKS kubernetes test to a generic one

- Related ticket: https://progress.opensuse.org/issues/105696
- Verification run: http://copland.arch.suse.de/tests/751


tests/containers/run_container_in_k8s.pm
```diff
@@ -22,7 +22,8 @@ sub run {
 
     my $cmd = '"cat", "/etc/os-release"';
 
-    my $provider = $self->provider_factory(service => 'EKS');
+    my $service = $self->get_k8s_service_name();
+    my $provider = $self->provider_factory(service => $service);
     $self->{provider} = $provider;
```
mloviska, line 27: Is there a reason why not use either `$provider` or `$self->{provider}` only?

ilausuch: Good point, I am moving this part to the parent (k8sbasetest)

pdostal (approved):

tests/containers/run_container_in_k8s.pm
ilausuch: Done

mloviska (approved):
Looks cool!

PR #14134, opened 2022-01-28
ovs: Remove count as the ping is killed

dzedro:
It can happen ping is done sooner than 20 packets are collected and then will tcpdump fail

- Fail: https://dzedro.suse.cz/tests/20693#step/ovs_client/48
- Verification run: https://dzedro.suse.cz/tests/20696


asdil12:
The VR seems to fail.

dzedro:
There is another issue and I forgot I opened this PR :grimacing:

PR #14135, opened 2022-01-29
Upload basic logs prior to the full problem detection logs

DimStar77:
basic_logs contains loadavg - we want to get this 'as soon as possible'
to the point where the error actually occurred. If we get this only
after collecting all other logs, a few minutes have passed, which has
a great chance to make the numbers appear much lower than what is
interesting for us.

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -492,9 +492,8 @@ sub export_logs {
     save_screenshot;
     show_oom_info;
     $self->remount_tmp_if_ro;
-    $self->problem_detection;
-
     $self->export_logs_basic;
+    $self->problem_detection;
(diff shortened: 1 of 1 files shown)


okurz (commented):
s/prpoblem/problem/

DimStar77:
> s/prpoblem/problem/

was fixed (but not commented, so likely not seen)

okurz (approved):
Well, I just saw one tiny problem, doesn't mean I should be the only reviewer. That's why I also just commented and haven't even asked for changes.

Your explanation makes sense. I have checked the git log and never was the order of execution changed or explicitly explained why it must stay how it was in before so I guess it's ok what you propose :)

PR #14137, opened 2022-01-31
Increase timeout for zypper ref

grisu48:
Increase the duration for zypper ref to avoid timeout issues and add two
retries to account for network issues.

- Related ticket: https://progress.opensuse.org/issues/105419
- Verification run: http://duck-norris.qam.suse.de/tests/8076#step/cleanup_qam_testrepos/6

### DIFF
--- a/tests/console/cleanup_qam_testrepos.pm
+++ b/tests/console/cleanup_qam_testrepos.pm
@@ -30,7 +30,7 @@ sub get_repo_aliases() {
 
 sub run {
     my $self = shift;
-    return if (script_run('zypper -n ref') == 0);    # If all repos are OK we don't have to do anything
+    return if (script_retry('zypper -n ref', timeout => (is_public_cloud ? 1200 : 300), retry => 3, delay => 30) == 0); # If all repos are OK we don't have to do anything
     my $behav = get_var('QAM_TESTREPO_FAIL') // "fail";
     # Check all repositories if they are valid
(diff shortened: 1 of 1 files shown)


jlausuch (approved):
Ok, but to be in the safe place, you could even do `timeout => 1200` :D

pdostal (approved):
Idea: What about increasing the timeout for public cloud jobs only?

tests/console/cleanup_qam_testrepos.pm
```diff
@@ -30,7 +30,7 @@ sub get_repo_aliases() {
 
 sub run {
     my $self = shift;
-    return if (script_run('zypper -n ref') == 0);    # If all repos are OK we don't have to do anything
+    return if (script_retry('zypper -n ref', timeout => (1200 ? is_public_cloud : 300), retry => 3, delay => 30) == 0); # If all repos are OK we don't have to do anything
```
jlausuch, line 33: lovely ternary operator, thanks! :) 
Interesting way to use it, btw. I would have done `(is_public_cloud ? 1200 : 300)`

grisu48: I'm apparently in dire need of some proper coffee today :coffee:

pdostal: I'd also do `if if_public_cloud then 1200 else 300` :smile:

pdostal (approved):
:rocket:

PR #14138, opened 2022-01-31
Zypper migration: make target regex more flexible

jlausuch:
In SLE Micro, recently the string has changed from

   `1 | SUSE Linux Enterprise ...`

to

   `1 |SUSE Linux Enterprise ...`

The regex expects a space between `|` and `SUSE`, so it fails in some
cases. This commit makes it more flexible, and it will work with
or without space.

- Related ticket: https://openqa.suse.de/tests/8075506
- Verification run: https://openqa.suse.de/tests/8075519#step/zypper_migration/1

### DIFF
--- a/tests/migration/online_migration/zypper_migration.pm
+++ b/tests/migration/online_migration/zypper_migration.pm
@@ -64,7 +64,7 @@ sub run {
         if ($out =~ $zypper_migration_target) {
             my $version = get_var("VERSION");
             $version =~ s/-/ /;
-            if ($out =~ /(\d+)\s+\|\s+SUSE Linux Enterprise.*?$version/m) {
+            if ($out =~ /(\d+)\s+\|\s?SUSE Linux Enterprise.*?$version/m) {
                 send_key "$1";
             }
             else {


tests/migration/online_migration/zypper_migration.pm
```diff
@@ -64,7 +64,7 @@ sub run {
         if ($out =~ $zypper_migration_target) {
             my $version = get_var("VERSION");
             $version =~ s/-/ /;
-            if ($out =~ /(\d+)\s+\|\s+SUSE Linux Enterprise.*?$version/m) {
+            if ($out =~ /(\d+)\s+\|.*SUSE Linux Enterprise.*?$version/m) {
```
mloviska, line 67: WDYT about `\s?`. That seems safer than going greedy

jlausuch:
@tinawang123

jlausuch:
Ready.

jlausuch:
I forgot, It's new year's in China :) 
So, I'll just merge and wait for feedback from our Chinese colleagues when they are back.

lemon-suse:
LGTM

PR #14142, opened 2022-01-31
serial_terminal: Relogin to workaround virtio console race

pevik:
Quite often there is a race on o3 during virtio console login
(about 30-50% change in Build 20220129 and few previous builds)"

https://openqa.opensuse.org/tests/2165897

Try to enter username again to workaround it.

Suggested-by: Richard Palethorpe <rpalethorpe@suse.com>
Signed-off-by: Petr Vorel <pvorel@suse.cz>

Verification run:
- https://openqa.opensuse.org/tests/2170315#step/boot_ltp/9 (relogin)
- https://openqa.opensuse.org/tests/2170314#step/boot_ltp/9 (relogin not needed)
- https://openqa.suse.de/tests/8080911#step/boot_ltp/9 (sles => relogin not needed)

### DIFF
--- a/lib/serial_terminal.pm
+++ b/lib/serial_terminal.pm
@@ -107,18 +107,27 @@ sub login {
     # Eat stale buffer contents, otherwise the code below may get confused
     # after reboot and start typing the username before the console is actually
     # ready to accept it
-    wait_serial(qr/login:\s*$/i, timeout => 5, quiet => 1);
+    wait_serial(qr/login:\s*$/i, timeout => 3, quiet => 1);
     # newline nudges the guest to display the login prompt, if this behaviour
     # changes then remove it
     send_key 'ret';
     die 'Failed to wait for login prompt' unless wait_serial(qr/login:\s*$/i);
     enter_cmd("$user");
+
+    my $re = qr/$user/i;
+    if (!wait_serial($re, timeout => 3)) {
+        record_info('RELOGIN', 'Need to retry login to workaround virtio console race', result => 'softfail');
+        enter_cmd("$user");
+        die 'Failed to wait for password prompt' unless wait_serial($re, timeout => 3);
+    }
+
     if (length $testapi::password) {
-        die 'Failed to wait for password prompt' unless wait_serial(qr/Password:\s*$/i);
+        die 'Failed to wait for password prompt' unless wait_serial(qr/Password:\s*$/i, timeout => 30);
         type_password;
         send_key 'ret';
     }
     die 'Failed to confirm that login was successful' unless wait_serial(qr/$escseq* \w+:~\s\# $escseq* \s*$/x);
+
     # Some (older) versions of bash don't take changes to the terminal during runtime into account. Re-exec it.
     enter_cmd('export TERM=dumb; stty cols 2048; exec $SHELL');
     die 'Failed to confirm that shell re-exec was successful' unless wait_serial(qr/$escseq* \w+:~\s\# $escseq* \s*$/x);


pevik:
@richiejp @mdoucha @cfconrad @jlausuch @Vogtinator FYI

grisu48:
Adding myself, so I get notified once this is merged. Please just ignore this comment.

Vogtinator:
> Adding myself, so I get notified once this is merged. Please just ignore this comment.

You can just click the "Subscribe" button at the right.

pdostal (approved):

Vogtinator:
Please link to a test run where the login failed.

richiejp:
Thanks, however I guess this will also break at some point. It kind of looks like it is aborting the login. 

Anyway this looks good as a workaround for now.

richiejp:
https://openqa.opensuse.org/tests/2165897 a fail example

Vogtinator:
> https://openqa.opensuse.org/tests/2165897 a fail example

The reason for the prompt appearing twice is that the IP address assignment changed, and `issue-generator` updated `/etc/issue`.

pevik:
While I think it's good to have fixed boot I wonder if we should report it to LABS.

BTW there is big timeout in boot_ltp caused by relogin, but that's probably just timeout on not working virtio:

working tests:
https://openqa.opensuse.org/tests/2167223 (default grub item 20s)
https://openqa.suse.de/tests/8076534 (ltp_ima - selects non-default grub item 50s)
https://openqa.suse.de/tests/8076485 (ltp_ima - selects non-default grub item 40s)

workarounded tests:
https://openqa.opensuse.org/tests/2167227 (default grub item **1m 52s**)
https://openqa.opensuse.org/tests/2167224 (ltp_ima - selects non-default grub item **1m 52s**)

pevik:
> The reason for the prompt appearing twice is that the IP address assignment changed, and issue-generator updated /etc/issue.

Yep, I noticed ` ens4: 10.0.2.15 fec0::3103:f4d5:7903:9f2e`, but I saw that in both broken and working tests, thus it should not affect the login

Working:
https://openqa.opensuse.org/tests/2166052/file/serial0.txt
```
Welcome to openSUSE Tumbleweed 20220129 - Kernel 5.16.2-1-default (ttyS0).

ens4: 10.0.2.15 fe80::5054:ff:fe12:3456

susetest login: 

Welcome to openSUSE Tumbleweed 20220129 - Kernel 5.16.2-1-default (ttyS0).

ens4: 10.0.2.15 fec0::b51e:f7bd:f0af:3fe8

susetest login: OpenQA::run_ltp.pm: Starting FSX032
OpenQA::run_ltp.pm: Starting FSX033
```
With race:
https://openqa.opensuse.org/tests/2165852/file/serial0.txt

```
Welcome to openSUSE Tumbleweed 20220129 - Kernel 5.16.2-1-default (ttyS0).

ens4: 10.0.2.15 fec0::a051:9b3:5a29:e516

susetest login: 
```

lib/serial_terminal.pm
```diff
@@ -114,7 +114,12 @@ sub login {
     die 'Failed to wait for login prompt' unless wait_serial(qr/login:\s*$/i);
     enter_cmd("$user");
     if (length $testapi::password) {
-        die 'Failed to wait for password prompt' unless wait_serial(qr/Password:\s*$/i);
+        if (!wait_serial(qr/Password:\s*$/i)) {
```
mdoucha, line 117: I'd recommend 4 changes:
- Set timeout to 5 seconds or less (default `wait_serial` timeout is 90 seconds)
- Wait for username to be echoed back instead of password prompt
- Move the workaround out of the parent condition so that it handles even cases where login is done without password
- Record softfail if the first login attempt fails

pevik:
> I don't really have a better idea either. Human users are affected in the same way.

Interesting. Is it documented somewhere?

lib/serial_terminal.pm
mdoucha: Alternatively, you could just try to insert 1 second sleep before `enter_cmd("$user");` and see whether the race can still happen.

richiejp:
> > > I don't really have a better idea either. Human users are affected in the same way.
> > 
> > 
> > Interesting. Is it documented somewhere?
> 
> Don't think so. It's visible if you boot a VM yourself, you'll see the login prompt getting replaced.
> 
> What I'm surprised by is that the typed in username is lost completely. If getty actually gets restarted that might be unavoidable. If the input was kept, then it should continue after the refresh as if nothing had happened...

hhmm, yes if you type `sleep 5\n echo foo\n`, then it should also run echo when sleep finishes. However I found that bytes would randomly get dropped so we had to start waiting for commands to fully finish. It wasn't clear to me if this were caused by bash or the kernel TTY/PTY code. Tracing TTY code would reveal the answer.

pdostal:
This pull request seems to be fine on [Public Cloud](http://pdostal-server.suse.cz/tests/13171#).

lib/serial_terminal.pm
```diff
@@ -107,18 +107,27 @@ sub login {
...
+    my $re = qr/$user/i;
+    if (!wait_serial($re, timeout => 3)) {
+        record_info('RELOGIN', 'Need to retry login to workaround virtio console race', result => 'softfail');
+        enter_cmd("$user");
+        die 'Failed to wait for password prompt' unless wait_serial($re, timeout => 3);
+    }
+
     if (length $testapi::password) {
-        die 'Failed to wait for password prompt' unless wait_serial(qr/Password:\s*$/i);
+        die 'Failed to wait for password prompt' unless wait_serial(qr/Password:\s*$/i, timeout => 3);
```
mdoucha, line 125: I think that 3 second timeout might be too short when the worker is under heavy load. Showing the password prompt is a lot more work than just echoing input back and I've been in situations where the password prompt took 10+ seconds to show up on my home laptop.

pevik: Good point, let's change it to 30s for safety.

mdoucha (approved):
LGTM

PR #14143, opened 2022-01-31
Use podman to upload images to provider registries

ilausuch:
The idea is to use podman instead of docker

- Related ticket: https://progress.opensuse.org/issues/98550
- Verification run: 
- http://copland.arch.suse.de/tests/765


jlausuch (commented):
LGTM

jlausuch:
> Is the `podman` installed by default?

looks like https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/prepare_tools.pm#L74

PR #14096, opened 2022-01-25
Disable root ssh login for cc setup

rfan1:
Based on CC test requirement, we may need disable
root ssh login, so introduce this new test module

this change may cause some test failures which need ssh access the 
test machine. so only commit this test module only,  we may need more 
code changes to fit this new design.

- Related ticket: https://progress.opensuse.org/issues/99096
- Needles: n/a
- Verification run: http://openqa.suse.de/tests/8025298#step/trustedprograms/46 
[The failure is by design, and I didn't change the current test logic]

PR #14144, opened 2022-01-31
detect_unused_modules: Fix module detection

pevik:
Both the original code in 5cdb32775 and attempt to fix it 407cfa0d9 were missing required quotes. Thus after 407cfa0d9 we got error:

```
tools/detect_unused_modules: 40: [: tests/containers/docker.pm: unexpected operator
tools/detect_unused_modules: 40: [: tests/containers/podman.pm: unexpected operator
tools/detect_unused_modules: 40: [: tests/containers/docker.pm: unexpected operator
...
```
not only on CI (we use ubuntu-latest, which has dash as the default shell which is more strict) but also on Tumbleweed with bash as the default shell. It's clear that people blindly rely on CI being green (don't run tests locally, don't look at the log output in CI when changing things).

Moreover 407cfa0d9 added redundant 'tests/' (directory already included in $MODULES, when called with -a), therefore nothing was checked due invalid path 'tests/tests/...', because that was needed for -m MODULES usage.

Therefore make 'tests/' mandatory, i.e. use relative path to test module (well, both 'tests/' and .pm extension are then removed) and thus put it into -m in Makefile.

Also remove duplicity and call tools/detect_unused_modules only once. This speedup things a lot.

NOTE: whole script has a messy design and it's hard to debug, it'd deserve redesign.

Verification run:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/5038697673?check_suite_focus=true


pevik:
@pdostal @rwx788 FYI

tools/detect_unused_modules
```diff
@@ -36,10 +36,12 @@ while [ "$#" -gt 0 ]; do
 done
 
 for test in $MODULES ; do
+    file="`git ls-files *$test*`"
```
pdostal, line 39: Why is this here?

pevik: Sorry, fixed (I noticed yesterday, but forget to push).

pevik: I also wonder whether this filter you tried to add in 407cfa0d9 shouldn't be changed in Makefile. It already filters tests:

```
test-unused-modules-changed:                                            
    @echo "[make] Unused modules check called over modified/new files only. For a full run use make test-unused-modules-full"
    tools/detect_unused_modules -m `git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/'`
    tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master products/* | grep -oP "^-.*loadtest.*[\"']\K[^\"'].+(?=[\"'])"`
    tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
                                                                        
.PHONY: test-unused-modules                                             
test-unused-modules:                                                    
    tools/detect_unused_modules -a
```
IMHO the script should allow to test locally whatever module user wants to test. Thus I'd limit it with `| grep` in `Makefile`.

pdostal: Well from my understanding there are no test modules outside of `tests/` but please feel free to improve it.

In my case, I deleted existing test module and the CI was failing. That's why I introduced 407cfa0.

pdostal (approved):

tools/detect_unused_modules
pevik: > But with your change testing `products/*` and `schedule/*` is now not working. Either it should be fixed => we need to revert 407cfa0) or it does not work (not possible to fix it) => we should remove these lines from Makefile.
@pdostal please should we remove these line from `Makefile`as they are filtered out since 407cfa0?
tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master products/* | grep -oP "^-.*loadtest.*```
[\"']\K[^\"'].+(?=[\"'])"`
    tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
```
Or should we revert 407cfa0 and fix the issue differently?

Comment (pevik) [reply tools/detect_unused_modules]: Thx for info. Could you please post the problematic PR / commit?
Also `-a` finds modules with `tests/`, but the others not. That must be fixed.

**Comment by pevik** [comment]
@pdostal FYU rewritten (updated PR description and commit message).

**Review by okurz — APPROVED** [review]

File: tools/detect_unused_modules (hunk 1 above)
Comment (pdostal) [reply tools/detect_unused_modules]: My 407cfa0 commit was follow up to 5cdb327.
I created #13302 as requested in #13246.

So it was because of not removing existing test module but renaming it.

Comment (pevik) [reply tools/detect_unused_modules]: Thanks for info. I tested that dd379ea6bbe591f143f971f9e4f6005dfe996c9e with reverted your change 407cfa0d98c2fa5edcae5222c8e9f9cda6fa4b02 is working with my version (and obviously failing with the version from dd379ea6bbe591f143f971f9e4f6005dfe996c9e, which also tries to test string from licence, 
`/home`, `/root`) => going to update commit message and merge.
As the problem was actually also on Tumbleweed, it document that people blindly rely on CI being green (don't run tests locally, don't look at the log output in CI when changing things).

File: Makefile [hunk 2]
```diff
@@ -131,9 +131,12 @@ perlcritic: tools/lib/
...
 	@echo "[make] Unused modules check called over modified/new files only. For a full run use make test-unused-modules-full"
-	tools/detect_unused_modules -m `git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/'`
-	tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master products/* | grep -oP "^-.*loadtest.*[\"']\K[^\"'].+(?=[\"'])"`
-	tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
+	f=$$(mktemp)
+	git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/' > $$f
+	git --no-pager diff --unified=0 origin/master products/* | sed -n "s~^-.*loadtest\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" >> $$f
+	git --no-pager diff --unified=0 origin/master schedule/* | sed -n "s~^-\s\+-\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" | grep -v '{{' >> $$f
+	sort -u $$f -o $$f
+	tools/detect_unused_modules -m $$(cat $$f)
```
okurz, line 139: In the latest step this became quite unreadable. Also I don't know if the temporary file is ensured to be pruned on failure. If you really need this then please put it into tools/detect_unused_modules instead of multi-level escaping in Makefile. Also your original "Add required quotes" is by far exceeded now. Please provide an according git commit message which explains the necessity

okurz (changes requested):

Makefile
```diff
@@ -131,12 +131,15 @@ perlcritic: tools/lib/
...
-	tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
-
-.PHONY: test-unused-modules
-test-unused-modules:
+	export f=$$(mktemp); \
+	git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/' > $$f; \
+	git --no-pager diff --unified=0 origin/master products/* | sed -n "s~^-.*loadtest\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" >> $$f; \
+	git --no-pager diff --unified=0 origin/master schedule/* | sed -n "s~^-\s\+-\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" | grep -v '{{' >> $$f; \
+	sort -u $$f -o $$f; \
+	tools/detect_unused_modules -m $$(cat $$f)
```
okurz, line 139: How about

```suggestion
	tools/detect_unused_modules -m $((git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/' && \
	git --no-pager diff --unified=0 origin/master products/* | sed -n "s~^-.*loadtest\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" && \
	git --no-pager diff --unified=0 origin/master schedule/* | sed -n "s~^-\s\+-\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" | grep -v '{{') \
	| sort -u)
```

not needing a temporary file (possibly tiny bit faster even)

pevik: My main concern is that pipes are hard to debug (you error in one big line) and frankly speaking openqa CI shell scripts are really very hard to read already. That's why I slightly prefer using the file.

Also the syntax is wrong: `$((` must be ended by `))` (understand it was just an example and I also had a problem to use `$(( .. ))` I needed to replace it with ` `` `).

`&&` mean if first `git grep` finds nothing, it exits 1 and thus other searches aren't run, thus we'd have to use `;` (I wanted to have single run, because I don't want to have files detected 2x or even 3x :).)

BTW I worried if it's wise to post input to the script, due max line limits.
I tested it on dd379ea6bbe591f143f971f9e4f6005dfe996c9e (MR#13246) which has 1644 lines. Putting that into single line it's 61904 chars. But max line is still higher:
```
$ getconf ARG_MAX 
2097152
```

pevik:
FYI list of unused tests:
@SergioAtSUSE yaml which uses systemd tests was removed in b363700f6 (#12202).
@rfan1 @lilyeyes cc_disable_root_ssh.pm was added in 807d423838a4059759b189030e494e2abd1982e8 (#14096) but looks to be never used.
@richiejp tests/kernel/write_random.pm
@jknphy tests/installation/system_role/select_role_SLES_with_GNOME.pm

Can some of this be removed or are they somewhere used (even locally)?

Makefile
okurz: So then the use of pipe is no problem :) you can also use the read command if you prefer. I just think handling a file is overkill. Consider my suggestion to move to the shell script instead

Makefile
```diff
@@ -131,12 +131,14 @@ perlcritic: tools/lib/
...
-	tools/detect_unused_modules -m `git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/'`
-	tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master products/* | grep -oP "^-.*loadtest.*[\"']\K[^\"'].+(?=[\"'])"`
-	tools/detect_unused_modules -m `git --no-pager diff --unified=0 origin/master schedule/* | grep -oP "^-\s+- [\"']?\K.*(?=[\"']?)" | grep -v '{{'`
-
-.PHONY: test-unused-modules
-test-unused-modules:
+	tools/detect_unused_modules -m ` (\
+	git --no-pager diff --name-only --diff-filter=d origin/master | grep '^tests/*' | grep -v '^tests/test_pods/'; \
+	git --no-pager diff --unified=0 origin/master products/* | sed -n "s~^-.*loadtest\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p"; \
+	git --no-pager diff --unified=0 origin/master schedule/* | sed -n "s~^-\s\+-\s\+\([\"']\)\([^\"']\+\)\1.*~tests/\2.pm~p" | grep -v '{{' ) \
```
okurz, line 137: I still wonder why it's necessary to make the regex so complex now

okurz (commented):
Just delete all not scheduled files. Anyone can recover from git when needed.

richiejp:
@pevik yeah, feel free to delete write_random

jknphy:
@pevik yes, you can delete `select_role_SLES_with_GNOME.pm` thanks.

Makefile
pevik: Not sure how to make it simpler. IMHO sed does not allow to use `\1` in `[^]`. Also if we join these two together it'd be even more complex.

tblume:
> @SergioAtSUSE @tblume Updated test found these unused modules. Are they really unused, or does something need to be whitelisted?
> 
> ```
> tests/systemd_testsuite/prepare_systemd_and_testsuite.pm module is not used in any schedule
> tests/systemd_testsuite/test_16_extend_timeout.pm module is not used in any schedule
> tests/systemd_testsuite/test_17_udev_wants.pm module is not used in any schedule
> tests/systemd_testsuite/test_18_failureaction.pm module is not used in any schedule
> tests/systemd_testsuite/test_19_delegate.pm module is not used in any schedule
> tests/systemd_testsuite/test_20_mainpidgames.pm module is not used in any schedule
> tests/systemd_testsuite/test_21_sysusers.pm module is not used in any schedule
> tests/systemd_testsuite/test_23_type_exec.pm module is not used in any schedule
> tests/systemd_testsuite/test_24_unit_tests.pm module is not used in any schedule
> tests/systemd_testsuite/test_25_import.pm module is not used in any schedule
> tests/systemd_testsuite/test_26_setenv.pm module is not used in any schedule
> tests/systemd_testsuite/test_27_stdoutfile.pm module is not used in any schedule
> tests/systemd_testsuite/test_28_percentj_wantedby.pm module is not used in any schedule
> tests/systemd_testsuite/test_29_udev_id_renaming.pm module is not used in any schedule
> tests/systemd_testsuite/test_30_onclockchange.pm module is not used in any schedule
> tests/systemd_testsuite/test_31_device_enumeration.pm module is not used in any schedule
> tests/systemd_testsuite/test_32_oompolicy.pm module is not used in any schedule
> tests/systemd_testsuite/test_33_clean_unit.pm module is not used in any schedule
> tests/systemd_testsuite/test_34_dynamicusermigrate.pm module is not used in any schedule
> tests/systemd_testsuite/test_36_numapolicy.pm module is not used in any schedule
> tests/systemd_testsuite/test_37_runtimedirectorypreserve.pm module is not used in any schedule
> tests/systemd_testsuite/test_39_execreload.pm module is not used in any schedule
> tests/systemd_testsuite/test_40_exec_command_ex.pm module is not used in any schedule
> tests/systemd_testsuite/test_41_oneshot_restart.pm module is not used in any schedule
> tests/systemd_testsuite/test_42_execstoppost.pm module is not used in any schedule
> tests/systemd_testsuite/test_43_privateuser_unpriv.pm module is not used in any schedule
> tests/systemd_testsuite/test_44_log_namespace.pm module is not used in any schedule
> tests/systemd_testsuite/test_47_issue_14566.pm module is not used in any schedule
> ```

We are still using those modules on our team internal openQA machine for systemd tests before submitting a new version.

If you want to get rid of them in the official openQA git, I guess I will have to fork.
Are there other openQA forks that could still be updated with the official openQA git?

Makefile
pevik: I removed `mktemp` yesterday. If you mean to move these `git --no-pager diff` to script, I could do that (having .e.g `-c` for finding modules with these 3 git diff and keep `-m` to provide extra module). But code would be just hidden in the script - IMHO not much difference. And ubuntu on GitHub Actions has even higher limit (4194304), we don't have to worry about the limit for passing to the script.

pevik:
> We are still using those modules on our team internal openQA machine for systemd tests before submitting a new version.
> 
> If you want to get rid of them in the official openQA git, I guess I will have to fork. Are there other openQA forks that could still be updated with the official openQA git?

Couldn't you just add yaml schedule, which you'd use for triggering? That way it'd keep the script happy (it also searches `schedule/`). If you don't want create yaml, we can whitelist them (no need to fork). But having schedule is preferred.

tblume:
> > We are still using those modules on our team internal openQA machine for systemd tests before submitting a new version.
> > If you want to get rid of them in the official openQA git, I guess I will have to fork. Are there other openQA forks that could still be updated with the official openQA git?
> 
> @tblume @SergioAtSUSE Couldn't you just add yaml schedule, which you'd use for triggering? That way it'd keep the script happy (it also searches `schedule/`). If you don't want create yaml, we can whitelist them (no need to fork). But having schedule is preferred.

Ah ok, I thought:

schedule/systemd/suse_patches-systemd_testsuite.yaml

is already in the official git.
If not, I will push it there, thanks!

(btw. Sergio has left the company, so @SergioAtSUSE isn't valid anymore)

pevik:
@tblume thx for info. Please ping me in PR where you add `schedule/systemd/suse_patches-systemd_testsuite.yaml`. I'll rebase this PR after that.

tblume:
> @tblume thx for info. Please ping me in PR where you add `schedule/systemd/suse_patches-systemd_testsuite.yaml`. I'll rebase this PR after that.

Hier is my pr:

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14167

but it fails a check :(

Makefile
okurz: Well, why change the regex from its former place at all? What problems did you encounter?

okurz: ok, cool

okurz (approved):

pevik:
@okurz removed removal of systemd tests so that I can merge it today (hope @tblume fixes https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14167).

lilyeyes:
> cc_disable_root_ssh.pm

Thanks for telling, @rfan1 will response this as he is the owner.

pevik:
@rfan1 I'm sorry, I already merged the deletion. If you need it, please submit it and add a corresponding yaml with it.
@lilyeyes thanks for info!

rfan1:
> @rfan1 I'm sorry, I already merged the deletion. If you need it, please submit it and add a corresponding yaml with it. @lilyeyes thanks for info!

@pevik, thanks for the info, sorry my late reply due to my vacation, I agree with you to remove the unused module there.
will file a new PR once we have final decision to disable root ssh on s390x cc setup.

PR #14148, opened 2022-02-01
SLE Micro 5.1: soft fail bsc#1195060

jlausuch:
Need to match the following strings:
`Failed to start Load kdump kernel early on startup`
`Failed to start Load kdump kernel and initrd`
`Load kdump kernel early on startup`
`Load kdump kernel and initrd`

Soft failure, because it blocks maintenance bot.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1195060
- VR:

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -228,6 +228,13 @@
         },
         "type": "bug"
     },
+    "bsc#1195060": {
+        "description": "Load kdump kernel (early on startup|and initrd)",
+        "products": {
+            "sle-micro": [ "5.1" ]
+        },
+        "type": "bug"
+    },
     "apci-bridge": {
         "description": "fail to add MMCONFIG information, can't access extended PCI configuration space under this bridge|PCI: System does not support PCI",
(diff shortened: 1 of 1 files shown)


Vogtinator:
Does it really need to match the lines without `Failed to` as well?

jlausuch:
> Does it really need to match the lines without `Failed to` as well?

yes, otherwise you get this https://openqa.suse.de/tests/8081312#step/journal_check/34

Vogtinator:
> > Does it really need to match the lines without `Failed to` as well?
> 
> yes, otherwise you get this https://openqa.suse.de/tests/8081312#step/journal_check/34

That's coming from the block here: https://github.com/jlausuch/os-autoinst-distri-opensuse/blob/e5ba965523fa76b888e49aab47da3783d8f53aed/tests/console/journal_check.pm#L88

FWICT that doesn't use the `bug_refs.json` file at all, so it would still fail.

jlausuch:
Test unscheduled until the bug is addressed.

PR #14151, opened 2022-02-01
Add wait_for_purge_kernels to avoid RPM lock

dzedro:
Wait for purge-kernels to finish

- Related ticket: https://progress.opensuse.org/issues/103956 https://bugzilla.suse.com/show_bug.cgi?id=1194693
- Verification run:
https://dzedro.suse.cz/tests/20749#step/register_without_ltss/4
https://openqa.suse.de/tests/8082962
https://openqa.suse.de/tests/8082959

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -9,7 +9,7 @@ use strict;
 use warnings;
 use testapi;
 use Utils::Architectures;
-use utils qw(addon_decline_license assert_screen_with_soft_timeout zypper_call systemctl handle_untrusted_gpg_key quit_packagekit script_retry);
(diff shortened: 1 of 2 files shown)


Vogtinator:
Is it necessary to stop the service? IMO openQA should not interfere there, but instead just wait until it's complete.

dzedro:
> Is it necessary to stop the service? IMO openQA should not interfere there, but instead just wait until it's complete.

First I did just wait, then I decided to be like `quit_packagekit`.
I don't mind, I guess just waiting is safer as the kernel is purged and RPM lock will not happen later after reboot.
 https://openqa.suse.de/tests/8080810#step/register_without_ltss/6

PR #14156, opened 2022-02-02
Check partitioning summary using Mojo::DOM

ge0r:
Until now the partitioning summary was tested using a needle, in order to check that all mount points of the encrypted disk partition of the previous installation were deleted.
This check is now done by parsing the partitioning summary using Mojo::DOM

- Related ticket: https://progress.opensuse.org/issues/102131
- Needles: No needles
- Verification runs: https://openqa.suse.de/tests/overview?build=ge0r%2Fos-autoinst-distri-opensuse%23parse_volumes&distri=sle&version=15-SP4

### DIFF
--- a/schedule/yast/encryption/activate_encrypted_volume+import_users.yaml
+++ b/schedule/yast/encryption/activate_encrypted_volume+import_users.yaml
@@ -17,7 +17,7 @@ schedule:
   - installation/system_probing/activate_encrypted_volume
   - installation/add_on_product/skip_install_addons
   - installation/system_role/accept_selected_role_SLES_with_GNOME
-  - installation/partitioning/encrypt_lvm_reuse_existing
+  - installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted
   - installation/partitioning/accept_proposed_layout
   - installation/clock_and_timezone/accept_timezone_configuration
   - installation/authentication/import_users
@@ -31,3 +31,10 @@ schedule:
   - installation/reboot_after_installation
   - installation/grub_test
   - installation/first_boot
+test_data:
+  partitioning_deletion_entries:
+    - 'Delete btrfs on /dev/system/root'
+    - 'Delete swap on /dev/system/swap'
+    - 'Delete xfs on /dev/system/home'
+    - 'Delete volume group system'
+    - 'Delete partition'
--- a/test_data/yast/encryption/activate_encrypted_volume.yaml
+++ b/test_data/yast/encryption/activate_encrypted_volume.yaml
@@ -6,3 +6,9 @@ device_status:
     cipher: 'aes-xts-plain64'
     key_location: 'dm-crypt'
     mode: 'read/write'
+partitioning_deletion_entries:
+  - 'Delete btrfs on /dev/system/root'
+  - 'Delete swap on /dev/system/swap'
+  - 'Delete xfs on /dev/system/home'
+  - 'Delete volume group system'
+  - 'Delete partition'
--- a/tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
+++ b/tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
@@ -1,0 +1,39 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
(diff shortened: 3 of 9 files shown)


tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
```diff
@@ -0,0 +1,43 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The test module reuses the partition that was created with LVM on
```
jknphy, line 6: We could update the summary and delete all the comment inside `run` (it is just my personal taste that it is distracting to have a comment which is obvious). Saying that we will use two list it is clear that the size should be same or that the module would die if any error. I would also improve the error in line 35 to avoid any comment on top, something like "expected entry '$foo' not found in corresponding block '$bar' from Suggested Partitioner summary.
(optional, really matter of taste)

tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
```diff
@@ -0,0 +1,43 @@
...
+# previous installation and verifies that it is shown in the partitioning list.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use parent 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
```
jknphy, line 13: is it needed? for the flags maybe?

tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
```diff
@@ -0,0 +1,43 @@
...
+
+sub run {
+    my $test_data = get_test_suite_data();
+    my @expected_strings = @{$test_data->{partitioning_deletion_entries}};
+
+    my $partitioner = $testapi::distri->get_suggested_partitioning();
+    my $text = $partitioner->get_partitioning_changes_summary();
+    my @deletion_entries = Mojo::DOM->new($text)->find('b')->map('text')->each;
+
+
```
jknphy, line 26: extra line break.

tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
```diff
@@ -0,0 +1,43 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
```
jknphy, line 3: no need for this line with the year afik for new tests, check some other new recent tests.

schedule/yast/encryption/activate_encrypted_volume+import_users.yaml
```diff
@@ -31,3 +31,5 @@ schedule:
   - installation/reboot_after_installation
   - installation/grub_test
   - installation/first_boot
+test_data:
+  <<: !include test_data/yast/encryption/activate_encrypted_volume.yaml
```
jknphy, line 35: we are dragging here two more keys beside the one we need. to make it easy when we will be refactor this part it is better if we just repeat the test data in the test suite, it is more clear and visible.

jknphy (commented):
It is looking good.

tests/installation/partitioning/suggested_partitioning/verify_decrypted_partition_deleted.pm
ge0r: hmm good point

ge0r:
I will address the progress issue's second action point (Merge test suites activate_encrypted_volume activate_encrypted_volume+import_users) in a new PR.

test_data/yast/encryption/activate_encrypted_volume.yaml
```diff
@@ -6,3 +6,9 @@ device_status:
     cipher: 'aes-xts-plain64'
     key_location: 'dm-crypt'
     mode: 'read/write'
+partitioning_deletion_entries:
```
jknphy, line 9: let's delete this file then.

jknphy:
LGTM

test_data/yast/encryption/activate_encrypted_volume.yaml
ge0r: still used by schedule/yast/encryption/activate_encrypted_volume.yaml  so we can keep it for now

PR #14160, opened 2022-02-02
lib/repo_tools: Use new repo name convention for Leap 15.4+

Vogtinator:
Since 15.3+, the new convention (and default in OBS) is to just use the version
number to indicate that it's usable for both Leap and SLE.

Currently used by:
* https://download.opensuse.org/repositories/home:/asmorodskyi/
* http://download.opensuse.org/repositories/benchmark:/ltp:/devel/

### DIFF
--- a/lib/repo_tools.pm
+++ b/lib/repo_tools.pm
@@ -462,7 +462,9 @@ sub generate_version {
     my $dist = get_required_var('DISTRI');
     my $version = get_required_var('VERSION');
     $separator //= '_';
-    if (is_sle) {
+    if (is_leap(">=15.4")) {
+        return $version;
+    } elsif (is_sle) {
         $dist = 'SLE';
         $version =~ s/-/$separator/;
     } elsif (is_tumbleweed) {


asmorodskyi:
@cfconrad  this will affect wicked tests

Vogtinator:
The first repo doesn't have a repo for 15.4 yet, so this won't break anything, so it's just the LTP project missing 15.4 which blocks this.

Vogtinator:
> > The first repo doesn't have a repo for 15.4 yet, so this won't break anything, so it's just the LTP project missing 15.4 which blocks this.
> 
> you right about my repo it is not an issue . But generally this issue is not only about OBS and projects over there . `generate_version` function used in wicked related test code . and this tests run in isolated openQA instance using some projects from IBS . and they also running against Leap versions ( all supported ) and this change most probably break all this setup.

Do those tests already use Leap 15.4?

asmorodskyi:
@cfconrad  must know :)

cfconrad:
No Leap 15.4 is not enabled till now. And I don't see a problem with this change. We will change the names then if needed.

asmorodskyi (approved):
LGTM

Vogtinator:
@asmorodskyi I'm still waiting for the repos of `benchmark:ltp:devel` to get fixed

PR #14161, opened 2022-02-02
Remove the un-necessary module updload_assets

JRivrain:
We have that module scheduled in a few places right after
logs_from_installation_system, collecting again the same logs.

Ticket: https://progress.opensuse.org/issues/104832


tests/console/upload_asset.pm
```diff
@@ -1,27 +0,0 @@
-# SUSE's openQA tests
```
JRivrain, line 1: I never paid attention to this module, it made me discover the function upload_asset which is actually pretty useful. I wonder if we should keep that module, but schedule it only when we need to get logs, instead of using VNC. But then we'd need to document it. And I think it would be nicer to use a variable in such case, instead of test data.

rakoenig (approved):
LGTM

PR #14162, opened 2022-02-03
Substitute SAP repo to LTSS

dzedro:
SAP is ESPOS so updates are going into Update LTSS repo
e.g. SLES_SAP_15_x86_64 -> SLES_15-LTSS_x86_64

- Related ticket: https://progress.opensuse.org/issues/105843
- Verification run:
https://openqa.suse.de/tests/8090946 15 SP1 x86_64
https://openqa.suse.de/tests/8090945 15 SP1 ppc64le
https://openqa.suse.de/tests/8090951 15 x86_64

### DIFF
--- a/tests/qam-updinstall/update_install.pm
+++ b/tests/qam-updinstall/update_install.pm
@@ -95,6 +95,8 @@ sub run {
     # Extract module name from repo url.
     my @modules = split(/,/, $repos);
     foreach (@modules) {
+        # substitue SLES_SAP for LTSS repo at this point is SAP ESPOS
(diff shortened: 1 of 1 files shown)


a-kpappas:
I think we should talk to MaintSec Tools first. This could be a submission/ibs error, because there is an ESPOS channel.[

a-kpappas:
Security said they will be retracting the submission, as it was done in error.

msmeissn:
it was not an error, submission was fine... see slack thread.

dzedro:
I would merge this change, until new API call is used ?

PR #14163, opened 2022-02-03
nvme_checks: Do not use 'nmve show-regs' in smoke-test scenario

igaw:
'nvme show-regs' accesses the raw registers via mapping IO space to
userspace. Today most distros enable CONFIG_IO_STRICT_DEVMEM which
prevents this operation. This can be overwritten by the kernel command
line option 'iomem=relaxed'. But this is not recommended for
production systems.

Hence using 'nvme show-regs' in a smoke-test for production images is
not particular useful. There are other nvme-cli commands which qualify
for smoke-tests but not this one. Remove it.

Signed-off-by: Daniel Wagner <dwagner@suse.de>

### DIFF
--- a/tests/console/nvme_checks.pm
+++ b/tests/console/nvme_checks.pm
@@ -56,9 +56,6 @@ sub _exercise_nvme_commands_and_validate_output {
     # lbads = LBA Data Size, in terms of a power of two, so LBA data size is 2⁹ = 512 bytes
     # rp = Relative Performance, the lower the better
     validate_script_output("nvme id-ns /dev/$nvm_test_data->{nvm_disk} --namespace-id=$nvm_test_data->{nvm_ns}", sub { m/lbaf\s+0\s+:\s+ms:0\s+lbads:9\s+rp:0\s\(in use\)/ });
-    # show registers
-    my $ret_code = script_run("nvme show-regs -H /dev/$nvm_test_data->{nvm_disk}");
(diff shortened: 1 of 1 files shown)


DimStar77:
This is related to https://bugzilla.opensuse.org/show_bug.cgi?id=1189869

igaw:
Any chance to move this forward?

igaw:
Rebased on the current HEAD. Any chance to get this merged?

Vogtinator:
Currently we set `iomem=relaxed` in the openQA test and we could keep that. Is it worth it to test `show-regs` at all or should we just drop that completely?

igaw:
There is nothing really gained in keeping the `show-regs` IMO. This info is interesting for hardware and driver developers to verify the raw registers. For the OS itself it doesn't matter. What we could add are some more higher level API tests, e.g. 'nvme list-ctrl' which output is based on these registers.

Vogtinator (approved):

Vogtinator:
I dropped `iomem=relaxed` from the test now. For reference (if we need it again), it's `GRUB_KERNEL_OPTION_APPEND="iomem=relaxed"` in the `nvme` test suite.

igaw:
Thanks. When I find some time I'll try to figure how out to add a `nvme list-ctrl` test. Obviously we want to parse the output and verify if the information in there is plausible. Also I see there is `_exercise_nvme_commands_and_validate_output` which screen scraps the output. This could be improved by using the JSON output format.

PR #14175, opened 2022-02-04
Allow using root-ssh console with GENERAL_HW_SOL_CMD

asdil12:
This is to allow using a real usb-uart serial adapter for hardware raspberry pi tests.

Previously root-ssh would only work with sshserial console
which is an ssh connection and a fifo file emulating a serial dev.

Now a root-ssh console can be used with an SOL or (USB-)Serial
connection.

The advantage here is that we get the complete boot log of the SUT.

Ticket: https://progress.opensuse.org/issues/105852
VR: http://10.163.27.222/tests/75

image


ggardet:
I am not sure `root-ssh` is a good name if it is for a serial connection.
Also, I already used: `GENERAL_HW_SOL_CMD = get_sol_over_SSH.sh` and `GENERAL_HW_SOL_CMD  = get_sol_ttyUSB0.sh` successfully in the past, so, I am not sure what problem you are trying to solve. Maybe it was used for some serial connection only?

asdil12:
The issue I currently have is that the test fails if I set `GENERAL_HW_SOL_CMD` because then the `root-ssh` console is not created. But I want to use that console. The test modules are referring to it and it is also the only thing that makes sence.

Why I want to set `GENERAL_HW_SOL_CMD` at all? Because otherwise there would be no way to tell what happened if the system doesn't come up far enough to bring up network and listen on ssh.
I want the current `root-ssh` console but still have serial log. And the `root-ssh` console seems to work fine with a real serial console instead of `sshserial` as a return channel.

asdil12:
But I just realized that I can use `root-ssh` with the `sshserial` device and still set `GENERAL_HW_SOL_CMD` for the boot log.
I will update the PR after some testing.

ggardet:
Could you update the commit message and PR accordingly with your latest push, please?

asdil12:
done

ggardet (approved):
LGTM

PR #14176, opened 2022-02-05
Adjust ibtest to run against science:hpc repo

b10n1k:
install SCIENCE_HPC_REPO with higher priority if it is defined. For now i use the SLE_15_SP3 repo which is available. TODO: fix this when SLE_15_SP4 is published
    Add the IBTEST_TIMEOUT. This like likely do not make any big difference but the default is kinda strict as the /ibtest/ might exceed this limit.
    Set it to 7200 has enough time until the test reach its own limit of the OpenQA jobs.
    
    In addition, the assert_script_run causes issues as it is waiting too long for the serial return code and i am sure that the test doesnt really need to be asserted in this case.
    
- Verification run: https://openqa.suse.de/tests/8106250


tests/kernel/ibtests.pm
```diff
@@ -6,11 +6,9 @@
 # Package: git-core twopence-shell-client bc iputils python
 # Summary: run InfiniBand test suite hpc-testing
 #
-# Maintainer: Michael Moese <mmoese@suse.de>, Nick Singer <nsinger@suse.de>
+# Maintainer: Michael Moese <mmoese@suse.de>, Nick Singer <nsinger@suse.de>, ybonatakis <ybonatakis@suse.com>
 
-use base 'opensusebasetest';
-use strict;
```
frankenmichl, line 12: Is there a reason to not use strict and warnings anymore?

schedule/kernel/ibtest-master.yaml
```diff
@@ -12,6 +12,7 @@ vars:
     IBTEST_IP1: 10.162.2.93
     IBTEST_IP2: 10.162.2.94
     IBTEST_ROLE: IBTEST_MASTER
+    IBTEST_TIMEOUT: 7200
```
frankenmichl, line 15: Is there a reason for this long timeout? The machine will be blocked for other jobs in this case, and a successful run should never even take an hour, does it?

tests/kernel/ibtests.pm
```diff
@@ -87,7 +87,7 @@ sub ibtest_master {
     # wait until the two machines under test are ready setting up their local things
     assert_script_run('cd hpc-testing');
     barrier_wait('IBTEST_BEGIN');
-    assert_script_run("./ib-test.sh $args $master $slave", $timeout);
+    script_run("./ib-test.sh $args $master $slave", $timeout);
```
frankenmichl, line 90: Why would you want to change this? In case this call fails, the test should die and execute the post fail hook, shouldn't it?

tests/kernel/ibtests.pm
b10n1k: https://docs.mojolicious.org/Mojo/Base#toc says it all. no need with _Mojo::Base_

schedule/kernel/ibtest-master.yaml
b10n1k: likely it wont. but i thought that this would be save. the 7200 exceeds the openQA timeout. So either it will run within the ibtest test or it will timeout from the openQA limits. in case that other tests added or the ibtest exceeds the 3200 (as before) it will not fail before the openqa tell it so.  i thought that this is save and i dont think has significant impact.

tests/kernel/ibtests.pm
b10n1k: i dont know exactly whats going on but the ibtest destructs the system or something and VPN connection dies and the `assert_script_run` never gets the return code. in case that the call fails, the test will die in the next command or it will be logged!

schedule/kernel/ibtest-master.yaml
frankenmichl: The machines are shared with other jobs as well, so we should try to have as little waste of time as possible. Unless there is no need for a timeout of two hours, I would propose to stick with less.

b10n1k: i removed the timeout and the results are https://openqa.suse.de/tests/8115897/logfile?filename=autoinst-log.txt. Adding back the timeout the test is able to finish https://openqa.suse.de/tests/8116064.

tests/kernel/ibtests.pm
frankenmichl: I think we can do this. But I think we can remove some then obsolete duplicate code from the post_fail_hook, I think.

schedule/kernel/ibtest-master.yaml
b10n1k: https://openqa.suse.de/tests/8120524#settings (w/o ibtest_timeout and opnmpi3).
So now i am confused. But take this out of this PR. i will remove the /IBTEST_TIMEOUT/ and you can resolve the issue.

tests/kernel/ibtests.pm
b10n1k: Done. i run new VR

b10n1k: *this request is ready for review*

schedule/kernel/ibtest-master.yaml
b10n1k: *this request is ready for review*

frankenmichl (approved):

PR #14181, opened 2022-02-07
kdump: Increase timeout for final kdump configuration

czerw:
Fix poo#106098: Timeout 90s is not enough during yast configuration, we
need to set it to same value 180s like it is in command line setup.

- Related ticket: https://progress.opensuse.org/issues/106098
- Needles: none
- Verification run:  https://openqa.suse.de/tests/8114327


czerw:
@dzedro could you please check?

dzedro (approved):

PR #14182, opened 2022-02-07
PC: Measure basic networking and record time of important steps

pdostal:
* We need to better see how long some important commands run.
 * It is hard to measure the network with unregistered product - this is just first attempt.

- Related ticket: [bsc#1195382](https://bugzilla.suse.com/show_bug.cgi?id=1195382) [poo#105969](https://progress.opensuse.org/issues/105969)
- Verification run: TBD

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -324,4 +324,21 @@ sub get_state
     return $self->provider->get_state_from_instance($self, @_);
 }
 
+=head2 network_speed_test
+
+    network_speed_test();
+
+Test the network speed.
+=cut
+sub network_speed_test() {
+    my ($self, %args) = @_;
+    # Curl stats output format
+    my $write_out = 'time_namelookup:\t%{time_namelookup} s\ntime_connect:\t\t%{time_connect} s\ntime_appconnect:\t%{time_appconnect} s\ntime_pretransfer:\t%{time_pretransfer} s\ntime_redirect:\t\t%{time_redirect} s\ntime_starttransfer:\t%{time_starttransfer} s\ntime_total:\t\t%{time_total} s\n';
+    # PC RMT server domain name
+    my $rmt_host = "smt-" . lc(get_required_var('PUBLIC_CLOUD_PROVIDER')) . ".susecloud.net";
+    $self->run_ssh_command(cmd => "grep '$rmt_host' /etc/hosts", proceed_on_failure => 1);
+    record_info("ping 1.1.1.1", $self->run_ssh_command(cmd => "ping -c30 1.1.1.1", proceed_on_failure => 1, timeout => 600));
+    record_info("curl $rmt_host", $self->run_ssh_command(cmd => "curl -w '$write_out' -o /dev/null -v https://$rmt_host/", proceed_on_failure => 1));
+}
+
 1;
--- a/tests/publiccloud/patch_and_reboot.pm
+++ b/tests/publiccloud/patch_and_reboot.pm
@@ -22,8 +22,14 @@ sub run {
 
     my $remote = $args->{my_instance}->username . '@' . $args->{my_instance}->public_ip;
 
-    $args->{my_instance}->retry_ssh_command(cmd => "sudo zypper -n ref", timeout => 240, retry => 6);
+    my $cmd_time = time();
+    my $ref_timeout = check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE') ? 3600 : 240;
+    $args->{my_instance}->retry_ssh_command(cmd => "sudo zypper -n ref", timeout => $ref_timeout, retry => 6, delay => 60);
+    record_info('zypper ref time', 'The command zypper -n ref took ' . (time() - $cmd_time) . ' seconds.');
+    record_soft_failure('bsc#1195382', 'bsc#1195382 - Considerable decrease of zypper performance and increase of registration times') if ((time() - $cmd_time) > 240);
+
     ssh_fully_patch_system($remote);
+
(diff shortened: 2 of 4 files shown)


lib/publiccloud/instance.pm
```diff
@@ -324,4 +324,21 @@ sub get_state
...
+Test the network speed.
+=cut
+sub network_speed_test() {
+    my ($self, %args) = @_;
+    # Curl stats output format
+    my $write_out = 'time_namelookup:\t%{time_namelookup} s\ntime_connect:\t\t%{time_connect} s\ntime_appconnect:\t%{time_appconnect} s\ntime_pretransfer:\t%{time_pretransfer} s\ntime_redirect:\t\t%{time_redirect} s\ntime_starttransfer:\t%{time_starttransfer} s\ntime_total:\t\t%{time_total} s\n';
+    # PC RMT server domain name
+    my $rmt_host = "smt-" . lc(get_required_var('PUBLIC_CLOUD_PROVIDER')) . ".susecloud.net";
+    $self->run_ssh_command(cmd => "grep '$rmt_host' /etc/hosts", proceed_on_failure => 1);
+    $self->run_ssh_command(cmd => "ping -c30 1.1.1.1", proceed_on_failure => 1, timeout => 600);
```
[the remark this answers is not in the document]
pdostal: Done.

tests/publiccloud/patch_and_reboot.pm
```diff
@@ -22,8 +22,13 @@ sub run {
 
     my $remote = $args->{my_instance}->username . '@' . $args->{my_instance}->public_ip;
 
-    $args->{my_instance}->retry_ssh_command(cmd => "sudo zypper -n ref", timeout => 240, retry => 6);
+    my $cmd_time = time();
+    $args->{my_instance}->retry_ssh_command(cmd => "sudo zypper -n ref", timeout => (check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE') ? 3600 : 240), retry => 6, delay => 60);
```
asmorodskyi, line 26: please move out timeout calculation . it is too much for one line . hard to read and understand what is going on . e.g. : 
```
my $timeout = check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE') ? 3600 : 240;
$args->{my_instance}->retry_ssh_command(cmd => "sudo zypper -n ref", timeout => $timeout, retry => 6, delay => 60);
```

lib/publiccloud/instance.pm
```diff
@@ -324,4 +324,21 @@ sub get_state
     return $self->provider->get_state_from_instance($self, @_);
 }
 
+=head2 network_speed_test
+
+    network_speed_test();
+
+Test the network speed.
+=cut
+sub network_speed_test() {
```
asmorodskyi, line 333: nice function :+1:

asmorodskyi:
other than small correction - LGTM

asmorodskyi (changes requested):
despite the fact that I fully agree with this code and don't have any concerns about it from programmatic POV  I am blocking this PR due to let's say "organizational " POV . with this PR  we introducing non-standard behavior of VMs within all CSP againsts internal CSP infra . Such behavior may be treated as attack by some automatic monitoring tools as well as ops staff of CSPs as well as ops staff of SUSE . So before making this live I would like to discuss this with PubCloud team ...

lib/publiccloud/instance.pm
pdostal: Thank you

lib/publiccloud/instance.pm
```diff
@@ -324,4 +324,21 @@ sub get_state
...
+    my ($self, %args) = @_;
+    # Curl stats output format
+    my $write_out = 'time_namelookup:\t%{time_namelookup} s\ntime_connect:\t\t%{time_connect} s\ntime_appconnect:\t%{time_appconnect} s\ntime_pretransfer:\t%{time_pretransfer} s\ntime_redirect:\t\t%{time_redirect} s\ntime_starttransfer:\t%{time_starttransfer} s\ntime_total:\t\t%{time_total} s\n';
+    # PC RMT server domain name
+    my $rmt_host = "smt-" . lc(get_required_var('PUBLIC_CLOUD_PROVIDER')) . ".susecloud.net";
+    $self->run_ssh_command(cmd => "grep '$rmt_host' /etc/hosts", proceed_on_failure => 1);
```
jlausuch, line 339: why is the smt server supposed to be in `/etc/hosts`?

pdostal: Because of `registercloudguest` - see [SUSE-Enceladus/cloud-regionsrv-client](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest)

jlausuch:
Looks promising :)

I wonder if we should use that new function to be a "blocking" step in our tests. At least for BYOS images.. If the SMT server is not reachable, does it make sense to proceed with the job execution if registration will fail...

pdostal:
> Looks promising :)
> 
> I wonder if we should use that new function to be a "blocking" step in our tests. At least for BYOS images.. If the SMT server is not reachable, does it make sense to proceed with the job execution if registration will fail?

When the registration fails this changes won't have any effect.

I increased the `zypper ref` timeout and added soft fail  so we are notified about such cases. ([bsc#1195382](https://bugzilla.suse.com/show_bug.cgi?id=1195382))
Other than that I'm not aware of any other timeouts in today's run - everything else are registration errors. ([bsc#1194986](https://bugzilla.suse.com/show_bug.cgi?id=1194986))

PR #14191, opened 2022-02-07
Ignore bug 1186742 in journal check

jlausuch:
The bug has been resolved as UPSTREAM.

https://bugzilla.suse.com/show_bug.cgi?id=1186742

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -16,14 +16,6 @@
         },
         "type": "bug"
     },
-    "bsc#1186742": {
-        "description": "Directory '.*.hdmi' with parent 'vc4-hdmi.*' already present",
-        "products": {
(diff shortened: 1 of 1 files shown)


mloviska:
Even though it is handled in upstream, I would leave it as a bug.

jlausuch:
> Even though it is handled in upstream, I would leave it as a bug.

but then the auto review will keep posting messages in bugzilla to a closed ticket.. not sure how to handle it.

mloviska:
> > Even though it is handled in upstream, I would leave it as a bug.
> 
> but then the auto review will keep posting messages in bugzilla to a closed ticket.. not sure how to handle it.

I haven't thought about this. So no objections LGTM

Vogtinator:
IMO as long as it impacts the product, openQA needs to show it.

PR #14194, opened 2022-02-08
ncurses: run 'dialog' using enter_cmd instead of script_run 

DimStar77:
The test starts dialog, an interactive command line tool. As the command
awaits user interaction, script_run is not returning and thus running
into a timeout, which is newly an error.

- Related ticket: https://progress.opensuse.org/issues/106176
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/2178924

### DIFF
--- a/tests/console/ncurses.pm
+++ b/tests/console/ncurses.pm
@@ -26,7 +26,7 @@ sub run {
     # Try to draw a bold red line
     script_run('echo "$(tput smacs;tput setaf 1;tput bold)lqqqqqqqqqqqqqqk$(tput rmacs;tput sgr0)"');
     # Try a simple yes/no dialog
-    script_run 'dialog --yesno "test for boo#1054448" 3 20 | tee boo_1054448.tee';
+    enter_cmd 'dialog --yesno "test for boo#1054448" 3 20 | tee boo_1054448.tee';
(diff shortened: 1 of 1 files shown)


DimStar77:
Fix for fallout from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14027

tests/console/ncurses.pm
```diff
@@ -26,7 +26,7 @@ sub run {
     # Try to draw a bold red line
     script_run('echo "$(tput smacs;tput setaf 1;tput bold)lqqqqqqqqqqqqqqk$(tput rmacs;tput sgr0)"');
     # Try a simple yes/no dialog
-    script_run 'dialog --yesno "test for boo#1054448" 3 20 | tee boo_1054448.tee';
+    script_run 'dialog --yesno "test for boo#1054448" 3 20 | tee boo_1054448.tee', die_on_timeout => 0;
```
mdoucha, line 29: Using `script_run()` here is wrong. The dialog needs to be executed using `enter_cmd()` instead.

mdoucha (changes requested):

tests/console/ncurses.pm
DimStar77: Fixed

mdoucha (approved):
LGTM

PR #14201, opened 2022-02-08
WIP Remove 'vim' out of installation console-tests

Zaoliang:
see https://progress.opensuse.org/issues/95724
verification:
https://openqa.opensuse.org/tests/2178941#step/vim

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1205,7 +1205,6 @@ sub load_consoletests {
         loadtest "console/yast2_i";
         loadtest "console/yast2_bootloader";
     }
-    loadtest "console/vim" if is_opensuse || is_sle('<15') || !get_var('PATTERNS') || check_var_array('PATTERNS', 'enhanced_base');
 # textmode install comes without firewall by default atm on openSUSE. For virtualization server xen and kvm is disabled by default: https://fate.suse.com/324207
     if ((is_sle || !check_var("DESKTOP", "textmode")) && !is_krypton_argon && !is_virtualization_server) {
         loadtest "console/firewall_enabled";


okurz:
@DimStar77

okurz (changes requested):
Please explain in your git commit message how the coverage for vim is ensured if it's not included in the installation tests anymore. Do staging tests still execute the vim test module?

Zaoliang:
the idea is about put vim into extra_test_misc. vim should not be tested within installation in general. And I don't know why for staging this module is required. 
At moment isotovideo failed, investigation...

okurz:
> the idea is about put vim into extra_test_misc. vim should not be tested within installation in general.

I know what the ticket says but the git commit messages should be self-contained to be able to understand the motivation of a change.

> And I don't know why for staging this module is required

Well, if you don't know then don't change it. You change the schedule so it could be that the staging schedule does not include the module anymore.

DimStar77:
> the idea is about put vim into extra_test_misc. vim should not be tested within installation in general. And I don't know why for staging this module is required. At moment isotovideo failed, investigation...

Staging, in general, is 'to ensure things do never break'; this implies that vim needs to be tested in staging; otherwise, we can't see side-effects impacting vim until they are checked in into a product, which in turn needlessly delays product/snapshot releases

Martchus:
Note that tests are likely failing because the test module is still referenced here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/651f604045f15a4eb4d42964a921e733bca28b7b/t/data/test_schedule.out#L38

okurz:
> Note that tests are likely failing because the test module is still referenced here:
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/651f604045f15a4eb4d42964a921e733bca28b7b/t/data/test_schedule.out#L38

IMHO as it should be. And that's also the discussion regarding staging tests

Zaoliang:
will find another way to handle this issue.

PR #14202, opened 2022-02-08
Remove soft-failure for bsc#1192941 as it won't be fixed

pdostal:
- Related ticket: [bsc#1192941](https://bugzilla.suse.com/show_bug.cgi?id=1192941)
- Verification run: [SLE 15-SP3 host_docker](http://pdostal-server.suse.cz/tests/13288)

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -201,11 +201,11 @@ sub test_zypper_on_container {
     die 'Argument $image not provided!' unless $image;
     die 'Argument $runtime not provided!' unless $runtime;
 
-    # Check for bsc#1192941, which affects the docker runtime only
+    # Check for bsc#1192941, which affects the docker runtime only - this is just check not softfailure as this won't be fixed.
     # bsc#1192941 states that the entrypoint of a derived image is set in a way, that program arguments are not passed anymore
     # we run 'zypper ls' on a container, and if we detect the zypper usage message, we know that the 'ls' parameter was ignored
-    if ($runtime->runtime eq 'docker') {
-        record_soft_failure("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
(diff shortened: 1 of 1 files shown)


jlausuch (changes requested):
I'm blocking this PR since `zypper-docker` is still under support. If that ceases, it will happen for new SPs, but we should still test it on older SPs (as far as I can tell)

jlausuch (commented):
After looking at the bug report, since it's marked as RESOLVED/UPSTREAM, yes, I agree to remove a soft-failure of something that might never be solved. In this case, we could do a record_info instead.

lib/containers/container_images.pm
```diff
@@ -201,11 +201,11 @@ sub test_zypper_on_container {
     die 'Argument $image not provided!' unless $image;
     die 'Argument $runtime not provided!' unless $runtime;
 
-    # Check for bsc#1192941, which affects the docker runtime only
+    # Check for bsc#1192941, which affects the docker runtime only - this is just check not softfailure as this won't be fixed.
     # bsc#1192941 states that the entrypoint of a derived image is set in a way, that program arguments are not passed anymore
     # we run 'zypper ls' on a container, and if we detect the zypper usage message, we know that the 'ls' parameter was ignored
     if ($runtime->runtime eq 'docker') {
-        record_soft_failure("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
+        record_info("bsc#1192941 zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
```
jlausuch, line 208:
```suggestion
        record_info("bsc#1192941",  "zypper-docker entrypoint confuses program arguments") if (script_run("docker run --rm -ti $image zypper ls | grep 'Usage:'") == 0);
```

jlausuch (approved):
LGTM

PR #14203, opened 2022-02-09
Fix brtfs_qgroups

michaelgrifalconi:
do not fail on timeout to match old behavior, before https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14027

This is a quick fix to not break the test. Will keep the ticket open and we can think on a proper solution to avoid the timeout later on.

- Related ticket: https://progress.opensuse.org/issues/106320
- Needles: no needles
- Verification run:  https://openqa.suse.de/t8124373

### DIFF
--- a/tests/console/btrfs_qgroups.pm
+++ b/tests/console/btrfs_qgroups.pm
@@ -101,7 +101,7 @@ sub run {
         record_soft_failure 'File overwrite test: bsc#1113042 - btrfs is not informed to commit transaction';
     }
     # write some more times to the same file to be sure
-    if (script_run("for c in {1..38}; do $write_chunk; done")) {
+    if (script_run("for c in {1..38}; do $write_chunk; done", die_on_timeout => 0)) {
         record_soft_failure 'File overwrite test: bsc#1113042 - btrfs is not informed to commit transaction';
     }
     assert_script_run 'sync';


tests/console/btrfs_qgroups.pm
```diff
@@ -101,7 +101,7 @@ sub run {
         record_soft_failure 'File overwrite test: bsc#1113042 - btrfs is not informed to commit transaction';
     }
     # write some more times to the same file to be sure
-    if (script_run("for c in {1..38}; do $write_chunk; done")) {
+    if (script_run("for c in {1..38}; do $write_chunk; done", die_on_timeout => 0)) {
         record_soft_failure 'File overwrite test: bsc#1113042 - btrfs is not informed to commit transaction';
     }
     assert_script_run 'sync';
```
foursixnine, line 107: I wonder if it would make sense to have a bigger timeout here, in case btrfs decides to be slow. Also, do I understand correctly that this is basically unconditionally going to happen a lot of times, and always report the soft failure? :/

michaelgrifalconi: short term goal here is to replicate the behavior before the merge of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14027 and make the test green.

I would keep the poo open and we can think about a real fix afterwards, but right now it is a blocker for  update approvals

foursixnine (approved):

PR #14207, opened 2022-02-09
installation/enable_selinux: Adapt to YaST UI changes

Vogtinator:
Tumbleweed and presumably 15-SP4/15.4 have a separate combo box for selecting
between SELinux and AppApparmor now which needs to be interacted with first.
The module is only used on SLE Micro and TW as of now, so add that condition
only for now.

- Related ticket: https://progress.opensuse.org/issues/104541
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/762
- Verification run: http://10.160.67.85/tests/1170#step/enable_selinux/16

### DIFF
--- a/tests/installation/enable_selinux.pm
+++ b/tests/installation/enable_selinux.pm
@@ -10,6 +10,7 @@ use strict;
 use warnings;
 use base 'y2_installbase';
 use testapi;
+use version_utils 'is_microos';
 
 sub run {
     my $textmode = check_var('VIDEOMODE', 'text');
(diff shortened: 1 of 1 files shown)


DimStar77 (approved):

tests/installation/enable_selinux.pm
```diff
@@ -28,13 +29,26 @@ sub run {
         send_key 'ret';
     }
 
-    send_key 'alt-m';
-    send_key_until_needlematch 'security-selinux-enforcing', 'down';
-    send_key 'ret' if $textmode;
+    if (is_microos('<5.3')) {
```
jlausuch, line 32: What is this condition doing? We don't have any microOS with version 5.3. 
If this is to be applied to SLE Micro, it should be `is_sle_micro('<5.3')`

jlausuch:
SLE Micro fails in SELinux enablement step https://openqa.suse.de/tests/8131304#step/enable_selinux/46

PR #14208, opened 2022-02-09
Prepare support for generalhw with video stream

ggardet:
to not use local VNC with remote ssh connection.

This has been lightly tested locally with `os-autoinst` from https://github.com/os-autoinst/os-autoinst/pull/1898

### DIFF
--- a/lib/Utils/Backends.pm
+++ b/lib/Utils/Backends.pm
@@ -106,7 +106,7 @@ Returns true if the current instance is using ttys
 =cut
 
 sub has_ttys {
-    return ((get_var('BACKEND', '') !~ /ipmi|s390x|spvm|pvm_hmc/) && !get_var('S390_ZKVM') && !(check_var('BACKEND', 'generalhw') && !defined(get_var('GENERAL_HW_VNC_IP'))) && !get_var('PUBLIC_CLOUD'));
+    return ((get_var('BACKEND', '') !~ /ipmi|s390x|spvm|pvm_hmc/) && !get_var('S390_ZKVM') && !(check_var('BACKEND', 'generalhw') && !defined(get_var('GENERAL_HW_VNC_IP')) && !defined(get_var('GENERAL_HW_VIDEO_STREAM_URL'))) && !get_var('PUBLIC_CLOUD'));
 }
 
 =head2 has_serial_over_ssh
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -459,7 +459,7 @@ sub init_consoles {
     }
 
     if ((get_var('BACKEND', '') =~ /qemu|ikvm/
-            || (get_var('BACKEND', '') =~ /generalhw/ && get_var('GENERAL_HW_VNC_IP'))
+            || (get_var('BACKEND', '') =~ /generalhw/ && (get_var('GENERAL_HW_VNC_IP') || get_var('GENERAL_HW_VIDEO_STREAM_URL')))
             || is_svirt_except_s390x))
     {
         $self->add_console('install-shell', 'tty-console', {tty => 2});
@@ -497,7 +497,7 @@ sub init_consoles {
     }
 
     # Use ssh consoles on generalhw, without VNC connection
-    if (get_var('BACKEND', '') =~ /generalhw/ && !defined(get_var('GENERAL_HW_VNC_IP'))) {
+    if (get_var('BACKEND', '') =~ /generalhw/ && !defined(get_var('GENERAL_HW_VNC_IP')) && !defined(get_var('GENERAL_HW_VIDEO_STREAM_URL'))) {
         my $hostname = get_required_var('SUT_IP');
         $self->add_console(
             'root-ssh',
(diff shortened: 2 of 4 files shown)


data/generalhw_scripts/gadget-control.py
```diff
@@ -0,0 +1,396 @@
+#!/usr/bin/python3
+#
+# Documentation: https://www.kernel.org/doc/Documentation/usb/gadget_hid.txt
+# Based on https://gist.github.com/marmarek/5c44ffeb2f36e106b4d34e7e8780c208
```
marmarek, line 4: This got several fixes/improvements since previous posting. I've just updated the gist.

marmarek: You couldn't share [this line](https://github.com/os-autoinst/os-autoinst/pull/1898/files#diff-457f23a16e93b308d1b29031db3d25ab7f7b5bc10b63541230aa51891523b9cdR88) then: 
But I'm fine with duplicating `mouse_button` function for that. @Martchus any preference? You originally requested to de-duplicate this part.

marmarek: Context: VNC and USB HID have different bits order there, see https://gist.github.com/marmarek/5c44ffeb2f36e106b4d34e7e8780c208/revisions#diff-f92d3a59d8f9df9a7d6415db914f4403016a1bd6388444b25ba57db889633444R345-R353

ggardet: Does it mean that VNC uses the current ordering and this is not an internal arbitrary convention for `os-autoinst`?

marmarek: Yes.

Martchus: `mouse_button` looks already de-duplicated on https://github.com/os-autoinst/os-autoinst/pull/1898. Am I missing something? In general I'd say de-duplicating is always a good idea but keeping backwards compatibility is more important. Your code doesn't look like it breaks backwards compatibility, though.

marmarek: It's about undoing this deduplication, to use different bits order in the button bitmask. Currently I need to [re-arrange them when sending to USB HID](https://gist.github.com/marmarek/5c44ffeb2f36e106b4d34e7e8780c208/revisions#diff-f92d3a59d8f9df9a7d6415db914f4403016a1bd6388444b25ba57db889633444R345-R353).

Martchus: I'm fine with it either way. If you like to undo the de-duplication to avoid the re-arrangement then go for it.

ggardet: Workaround removed from this script, since it will be fixed/handled in `os-autoinst` directly.

data/generalhw_scripts/gadget-control.py
```diff
@@ -0,0 +1,403 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--storage', default=None, action='store')
+    parser.add_argument('--is-cdrom', default=False, action='store_true')
+    parser.add_argument('--storage-only', default=False, action='store_true')
+    args = parser.parse_args()
+
+    with Gadget(storage_path=args.storage, cdrom=args.is_cdrom, hid=not args.storage_only) as gadget:
+        for line in fileinput.input(files=[]):
```
asdil12, line 399: Why not just something like this?
```suggestion
        for line in sys.stdin:
```

asdil12:
Which hardware do you plan to be used here?
I guess the system running the python script will need usb gadget support - do you have something in particular in mind?
And what hardware are you planning for video capture? I suspect that it is more than a simple USB-HDMI capture stick.

ggardet:
> Which hardware do you plan to be used here? I guess the system running the python script will need usb gadget support - do you have something in particular in mind?

Yes, the RPi zero 2 W is in my mind, but quite difficult to order atm...
For now, I use a [SabreLite](https://en.opensuse.org/HCL:SABRE_Lite), but it is an armv7 board.
I think @marmarek uses a RPi4.

> And what hardware are you planning for video capture? I suspect that it is more than a simple USB-HDMI capture stick.

I use a simple USB-HDMI capture stick. I think @marmarek uses a RPi4 with HDMI-to-CSI-2.

asdil12:
So `GENERAL_HW_VIDEO_STREAM_URL` can have the value `/dev/video0`?

ggardet:
> So `GENERAL_HW_VIDEO_STREAM_URL` can have the value `/dev/video0`?

Yes, I used it this week :)

marmarek:
> I think @marmarek uses a RPi4 with HDMI-to-CSI-2.

Yes, exactly.

ggardet:
> > I think @marmarek uses a RPi4 with HDMI-to-CSI-2.
> 
> Yes, exactly.

Which OS is running on this RPi4?

marmarek:
> Which OS is running on this RPi4?

openSUSE Tumbleweed

asdil12:
And this RPi4 is running the openQA-worker?
But not the sabrelite/zero-2w?

marmarek:
> And this RPi4 is running the openQA-worker?

Yes, it is.

ggardet:
> > Which OS is running on this RPi4?
> 
> openSUSE Tumbleweed

Could you tell me how you set-up your HDMI-to-CSI-2 please? Last time I tried on my RPi4, I did not succeed.

marmarek:
> Could you tell me how you set-up your HDMI-to-CSI-2 please? Last time I tried on my RPi4, I did not succeed.

I had several issues with that indeed. In the end I gave up vanilla kernel and - used the one from raspberry.

ggardet:
> > Could you tell me how you set-up your HDMI-to-CSI-2 please? Last time I tried on my RPi4, I did not succeed.
> 
> I had several issues with that indeed. In the end I gave up vanilla kernel and used the one from raspberry.

Ok. What is weird is that camera on CSI-2 just work, but not the HDMI input.

ggardet:
> > Ok. What is weird is that camera on CSI-2 just work, but not the HDMI input.
> 
> Yes, and even HDMI input device is detected according to dmesg, but never shows up in `/dev`.

Yes, there is a mix between firmware supported devices and kernel supported devices... 
Let me reference https://en.opensuse.org/HCL:Raspberry_Pi_Camera_Modules just in case someone manages to get HDMI input working at some point.

PR #14209, opened 2022-02-09
Revert upload_asset deletion

jknphy:
detect_yast2_failures is using the asset which is uploaded.
Please, check https://openqa.suse.de/tests/8126491#dependencies

- Related ticket: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14161

### DIFF
--- a/schedule/yast/raid/raid0_sle_gpt.yaml
+++ b/schedule/yast/raid/raid0_sle_gpt.yaml
@@ -26,6 +26,7 @@ schedule:
   - installation/confirm_installation
   - installation/performing_installation/perform_installation
   - installation/logs_from_installation_system
+  - console/upload_asset
   - installation/reboot_after_installation
   - installation/grub_test
   - installation/first_boot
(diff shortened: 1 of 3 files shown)


JRivrain:
I have the impression the only thing we need to change here is ASSET_1=logs_from_installation_system instead of ASSET_1=compressed_logs, as they both contain the same logs. Checking.

jknphy:
> I have the impression the only thing we need to change here is ASSET_1=logs_from_installation_system instead of ASSET_1=compressed_logs, as they both contain the same logs. Checking.

I think it is the only way to pass info among jobs is as assets, so we need to create the asset in the parent job and get it in the other one...afir

JRivrain:
Merging, it is not enough to just change ASSET_1 as  logs_from_installation_system-y2logs.tar.bz2 are not placed under assets/other. Then we will think of a better hack that does not involve collecting the same logs twice...

PR #14215, opened 2022-02-09
Simplify the way we retrieve logs

JRivrain:
We have been collecting two times the same log files for a while,
In order to upload those afterwards for detect_yast2_failures.
This change avoids re-collecting logs that we already just collected,
and simplifies what may have been sligthly over-engineered code.

- Related ticket: https://progress.opensuse.org/issues/104832
VR: http://waaa-amazing.suse.cz/tests/16173
Needs MR to be merged


tests/console/upload_asset.pm
```diff
@@ -18,9 +18,7 @@ use scheduler "get_test_suite_data";
 
 sub run {
     select_console 'root-console';
-    my $files_to_upload = get_test_suite_data()->{asset_files};
-    assert_script_run("tar -cvf asset_files.tar $files_to_upload", fail_message => "Failed to compress file(s)");
```
JRivrain, line 22: It is visible that the module "upload_asset" was meant to be generic, hence the usage of test data. However in practice, this module is used in only one place, and is redundant as it collects again the files that were already collected in the previous module "logs_from_installation_system". 
This change uses the "strict expectation" approach, we could also keep using the "no test data in modules" approach: just change test data, and only change this line to upload existing logs instead of creating a new archive.

PR #14222, opened 2022-02-10
Containers: Allow running both runtimes in the same job

ilausuch:
Container runntime variable could be have both runtimes at the same
time

CONTAINER_RUNTIME=docker,podman

- Related ticket: https://progress.opensuse.org/issues/101451
- Verification run: 
https://openqa.suse.de/tests/8140544 (Podman)
https://openqa.suse.de/tests/8140546 (Docker)
https://openqa.suse.de/tests/8140547 (Podman & Docker)

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -100,22 +100,31 @@ sub load_host_tests_load_containerd_nerdctl {
 
 sub load_container_tests {
     my $runtime = get_required_var('CONTAINER_RUNTIME');
+
     if (get_var('BOOT_HDD_IMAGE')) {
         loadtest 'installation/bootloader_zkvm' if is_s390x;
         loadtest 'boot/boot_to_desktop' unless is_jeos;
     }
 
     if (is_container_image_test()) {
-        # Container Image tests
+        # Container Image tests common
         loadtest 'containers/host_configuration' unless (is_expanded_support_host || is_ubuntu_host || is_jeos);
-        load_image_tests_podman() if ($runtime =~ 'podman');
-        load_image_tests_docker() if ($runtime =~ 'docker');
-    } else {
-        # Container Host tests
-        load_host_tests_podman() if ($runtime =~ 'podman');
-        load_host_tests_docker() if ($runtime =~ 'docker');
-        load_host_tests_load_containerd_crictl() if ($runtime =~ 'containerd_crictl');
-        load_host_tests_load_containerd_nerdctl() if ($runtime =~ 'containerd_nerdctl');
     }
+
+    foreach (split(',\s*', $runtime)) {
+        if (is_container_image_test()) {
+            # Container Image tests
+            load_image_tests_podman() if (/podman/i);
+            load_image_tests_docker() if (/docker/i);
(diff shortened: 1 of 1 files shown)


lib/main_containers.pm
```diff
@@ -100,22 +100,34 @@ sub load_host_tests_load_containerd_nerdctl {
 
 sub load_container_tests {
     my $runtime = get_required_var('CONTAINER_RUNTIME');
+    $runtime = uc $runtime;
+    $runtime =~ tr/ //ds;
+    my @runtimes = split(',', $runtime);
```
pdostal, line 105: Can you reduce this to one line?

lib/main_containers.pm
```diff
@@ -100,22 +100,33 @@ sub load_host_tests_load_containerd_nerdctl {
 
 sub load_container_tests {
     my $runtime = get_required_var('CONTAINER_RUNTIME');
+    $runtime =~ tr/ //ds;
+    my @runtimes = split(',', uc $runtime);
+
```
mloviska, line 105:
```
use Data::Dumper;
my $r1 = "a,b, c,d,      e";
foreach (split(',\s*', $r1)) {
        print "token .... " . Dumper($_);
}
```

Works nice too

```
token .... $VAR1 = 'a';
token .... $VAR1 = 'b';
token .... $VAR1 = 'c';
token .... $VAR1 = 'd';
token .... $VAR1 = 'e';
```

I admit you would filter out all whitespaces :)

lib/main_containers.pm
```diff
@@ -100,22 +100,33 @@ sub load_host_tests_load_containerd_nerdctl {
...
-        load_host_tests_docker() if ($runtime =~ 'docker');
-        load_host_tests_load_containerd_crictl() if ($runtime =~ 'containerd_crictl');
-        load_host_tests_load_containerd_nerdctl() if ($runtime =~ 'containerd_nerdctl');
     }
+
+    foreach (@runtimes) {
+        if (is_container_image_test()) {
+            # Container Image tests
+            load_image_tests_podman() if ($_ =~ 'podman');
+            load_image_tests_docker() if ($_ =~ 'docker');
```
mloviska, line 120: I would prefer to go either string comparison `eq` or match regex literals (maybe even case-insensitive)

If we go the second way, we can also omit `$_`.

```
... if (/podman/i);
``` 

Should work

lib/main_containers.pm
ilausuch: Now this is implemented, thanks both

lib/main_containers.pm
ilausuch: This is implemented too

lib/main_containers.pm
ilausuch: solved with Martin suggestions. Thanks

mloviska (approved):

jlausuch:
Sorry I'm late in the game again, but LGTM :)

Small minor renaming for a function: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14233

PR #14228, opened 2022-02-11
Fix rhythmbox test

michaelgrifalconi:
need to open a shell to check for dependency. Was working before due to
previous behavior of script_run timeout.

- Related ticket: https://progress.opensuse.org/issues/106636
- Needles:no needles
- Verification run: https://openqa.suse.de/tests/8178095


punkioudi (approved):
lgmt!

PR #14229, opened 2022-02-11
spec: Fix dependency requirement on os-autoinst-devel

okurz:
The tools called from Makefile, e.g. "tidy", rely not only on
os-autoinst to execute but actually the development environment which
also pulls in perl-Code-TidyAll.


okurz:
@foursixnine @grisu48 I think you most recently ran into the problem that perl-Code-TidyAll is not automatically provided. This should help assuming that you also have an updated os-autoinst-distri-opensuse-deps package installed.

foursixnine:
> @foursixnine @grisu48 I think you most recently ran into the problem that perl-Code-TidyAll is not automatically provided. This should help assuming that you also have an updated os-autoinst-distri-opensuse-deps package installed.

Also assuming that I use the packages on a given machine :)

PR #14232, opened 2022-02-11
Improve error msg when not using libyui-rest-api

ge0r:
When a module that uses libyui-restapi failing on a testsuite that has not set up libyui-rest-api, the resulting error is not very
descriptive.
We improve error message so that reviewers know the exact reason, when such a module fails.

- Related ticket: https://progress.opensuse.org/issues/103917
- Needles: No needles
- Verification run: https://openqa.opensuse.org/tests/2184318#step/disable_online_repos/2

### DIFF
--- a/lib/YuiRestClient.pm
+++ b/lib/YuiRestClient.pm
@@ -25,6 +25,8 @@ my $host;
 
 sub get_app {
     my (%args) = @_;
+    die "No environment for libyui REST API has been configured.",
+      " Please schedule test module `setup_libyui.pm` in previous steps." unless YuiRestClient::is_libyui_rest_api();
     $app = init_app(%args) unless $app;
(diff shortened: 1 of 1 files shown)


jknphy (approved):

jknphy:
We missed something important here, when `YUI_REST_API` variable is not set the error is really confusing, because it fails in `setup_libyui.pm` telling you that you are not setting that one xD. Could we add also that info?

ge0r:
> We missed something important here, when `YUI_REST_API` variable is not set the error is really confusing, because it fails in `setup_libyui.pm` telling you that you are not setting that one xD. Could we add also that info?

followup PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14266

PR #14234, opened 2022-02-11
PublicCloud: increase timeout for activating PackageHub

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/106669
- vR: http://openqa.suse.de/t8142091

### DIFF
--- a/tests/publiccloud/xfsprepare.pm
+++ b/tests/publiccloud/xfsprepare.pm
@@ -14,6 +14,7 @@ use Mojo::Base 'consoletest';
 use testapi;
 use version_utils qw(is_sle);
 use registration qw(add_suseconnect_product get_addon_fullname);
+use publiccloud::utils qw(is_azure);
 use utils;
 
 # xfstests configuration file, as required by xfstests/run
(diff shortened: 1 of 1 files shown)


tests/publiccloud/xfsprepare.pm
```diff
@@ -47,7 +47,7 @@ sub ensure_user_exists {
 sub install_xfstests {
     my ($repo) = @_;
     zypper_ar($repo, name => "filesystems");    # Add filesystem repository, which contains the xfstests
-    add_suseconnect_product(get_addon_fullname('phub')) if is_sle;    # packagehub is required for dbench (required for e.g. generic/241)
+    add_suseconnect_product(get_addon_fullname('phub'), undef, undef, undef, 300, 1) if is_sle; # packagehub is required for dbench (required for e.g. generic/241)
```
asmorodskyi, line 50: for now we have confirmed problems only for Azure . let's keep it this way . please add logic which will increase timeout only for Azure

jlausuch: This is not changing the behavior or the flow, it only increases the timeout. I don't think it harms to keep it for all cases.

asmorodskyi: currently we have ticket for handling timeout for Azure only . we not working anyhow on handling timeouts for Amazon and GCP . The harm which I see is that when problem popup in GCP or Amazon we won't be able to detect this because timeout was already increased for Amazon and GCP .

tests/publiccloud/xfsprepare.pm
```diff
@@ -47,7 +47,14 @@ sub ensure_user_exists {
 sub install_xfstests {
     my ($repo) = @_;
     zypper_ar($repo, name => "filesystems");    # Add filesystem repository, which contains the xfstests
-    add_suseconnect_product(get_addon_fullname('phub')) if is_sle;    # packagehub is required for dbench (required for e.g. generic/241)
+    if (is_sle) {
+        # packagehub is required for dbench (required for e.g. generic/241)
+        if (check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE')) {
```
asmorodskyi, line 52:
```suggestion
        if (is_azure()) {
```

jlausuch:
VR: https://openqa.suse.de/tests/8163462

jlausuch:
I can't run VRs now cause the original jobs for Azure are also failing due to empty repos...

asmorodskyi:
> I can't run VRs now cause the original jobs for Azure are also failing due to empty repos...

change is pretty straightforward so I think we can merge

PR #14238, opened 2022-02-14
[virt_autotest] fix up timeout failure from s390x update_package test

guoxuguang:
According to the new change from script_run() ,  die_on_timeout=> 1 as the default option. Refer to the latest osd s390x results, figure out this default option do not work very well for s390x during call script_run(). 
Use with die_on_timeout=> 0, keep the old behavior, as the workaround for s390x during call script_run(). refer to [poo#106765](https://progress.opensuse.org/issues/106765) for more details. 
- Related ticket: https://progress.opensuse.org/issues/106765
- Verification run: 
s390x
[gi-guest_developing-on-host_developing-kvm](https://openqa.suse.de/tests/8186575#)
[gi-guest_sles12sp5-on-host_developing-kvm](http://openqa.suse.de/t8189434)
[gi-guest_sles15sp3-on-host_developing-kvm](http://openqa.suse.de/t8189435)
[gi-guest_developing-on-host_sles12sp5-kvm](http://openqa.suse.de/t8189432)
[gi-guest_developing-on-host_sles15sp3-kvm](http://openqa.suse.de/t8189433)
[prj4_guest_upgrade_sles12sp5_on_sles12sp5-kvm](https://openqa.suse.de/tests/8189604#)
[prj4_guest_upgrade_sles15sp3_on_sles15sp3-kvm](https://openqa.suse.de/tests/8186598#)
x86
[gi-guest_developing-on-host_developing-kvm](https://openqa.suse.de/tests/8206014#)
aarch64
[prj4_guest_upgrade_sles12sp5_on_sles12sp5-kvm](https://openqa.suse.de/tests/8189608#)

### DIFF
--- a/tests/virt_autotest/guest_upgrade_run.pm
+++ b/tests/virt_autotest/guest_upgrade_run.pm
@@ -108,7 +108,12 @@ sub run {
     my $self = shift;
     my $timeout = get_var('MAX_TEST_TIME', '36000') + 10;
     my $upload_log_name = 'guest-upgrade-logs';
-    script_run("echo \"Debug info: max_test_time is $timeout\"");
+    if (is_s390x) {
+        #ues die_on_timeout=> 0 as workaround for s390x test during call script_run, refer to poo#106765
+        script_run("echo \"Debug info: max_test_time is $timeout\"", die_on_timeout => 0);
+    } else {
+        script_run("echo \"Debug info: max_test_time is $timeout\"");
+    }
     # Modify source configuration file sources.* of virtauto-data pkg on host
     # to use openqa daily build installer repo and module repo for guests,
     # and it will be copied into guests to be used during guest upgrade test
@@ -116,7 +121,8 @@ sub run {
 
     $self->{'package_name'} = 'Guest Upgrade Test';
     if (is_s390x) {
-        script_run "[ -d /var/log/qa/ctcs2 ] && rm -r /var/log/qa/ctcs2 ; [ -d /var/lib/libvirt/images/prj4_guest_upgrade ] && rm -r /var/lib/libvirt/images/prj4_guest_upgrade /tmp/full_guest_upgrade_test-* /tmp/kill_zypper_procs-* /tmp/update-guest-*";
+        #ues die_on_timeout=> 0 as workaround for s390x test during call script_run, refer to poo#106765
+        script_run "[ -d /var/log/qa/ctcs2 ] && rm -r /var/log/qa/ctcs2 ; [ -d /var/lib/libvirt/images/prj4_guest_upgrade ] && rm -r /var/lib/libvirt/images/prj4_guest_upgrade /tmp/full_guest_upgrade_test-* /tmp/kill_zypper_procs-* /tmp/update-guest-*", die_on_timeout => 0;
     }
     else {
         $self->execute_script_run("[ -d /var/log/qa/ctcs2 ] && rm -r /var/log/qa/ctcs2 ; [ -d /tmp/prj4_guest_upgrade ] && rm -r /tmp/prj4_guest_upgrade", 30);
@@ -126,7 +132,8 @@ sub run {
     if (is_s390x) {
         #upload s390x_guest_upgrade_test.log
         upload_asset("/tmp/s390x_guest_upgrade_test.log", 1, 1);
-        script_run "rm -rf /tmp/s390x_guest_upgrade_test.log";
+        #ues die_on_timeout=> 0 as workaround for s390x test during call script_run, refer to poo#106765
+        script_run "rm -rf /tmp/s390x_guest_upgrade_test.log", die_on_timeout => 0;
     }
 
 }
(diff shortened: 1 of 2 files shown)


guoxuguang:
@alice-suse @Julie-CAO @waynechen55 @nanzhg if have time, please help me review this PR. thanks.

tests/virt_autotest/guest_upgrade_run.pm
```diff
@@ -108,15 +108,21 @@ sub run {
     my $self = shift;
     my $timeout = get_var('MAX_TEST_TIME', '36000') + 10;
     my $upload_log_name = 'guest-upgrade-logs';
-    script_run("echo \"Debug info: max_test_time is $timeout\"");
+    if (is_s390x) {
+        #ues die_on_timeout=> 0 as workaround for s390x test during call script_run, refer to poo#106765
+        script_run("echo \"Debug info: max_test_time is $timeout\"", die_on_timeout => 0);
+    } else {
+        script_run("echo \"Debug info: max_test_time is $timeout\"");
+    }
```
alice-suse, line 116: Can here, both s390x and non-s390x use the same script_run with die_on_timeout =>0?

tests/virt_autotest/update_package.pm
```diff
@@ -61,9 +61,16 @@ sub run {
...
+        #ues die_on_timeout=> 0 as workaround for s390x test during call script_run, refer to poo#106765
+        my %args = ();
+        if (is_s390x) {
+            $args{die_on_timeout} = 0;
+        } else {
+            $args{die_on_timeout} = 1;
+        }
+        script_run "sed -i 's/#dnssec-validation auto;/dnssec-validation no;/g' /etc/named.conf", %args;
+        script_run "grep 'dnssec-validation' /etc/named.conf", %args;
+        script_run "systemctl restart named", %args;
```
alice-suse, line 73: Similar here, shall s390 and non-s390 both use die => 0?

tests/virt_autotest/guest_upgrade_run.pm
waynechen55: @guoxuguang  I think you'd better raise this issue to openQA project or let them know. It seems that the change does not take this usage on s390x into consideration which means you have to use die_on_timeout => 0. Now you made the change which branches off s390x. I did not say branching off here is really bad here, but it does force us to make the same change to other arches here to some extent because it looks not very pleasant and leads that way.  Very interesting, we end up having to use die_on_timeout => 0 because of API change. I do not think this is what this API change initially would like to see.  
Maybe you use type_string here instead of script_run ? So we can avoid change to other tests ?

tests/virt_autotest/update_package.pm
waynechen55: @guoxuguang  I think you'd better raise this issue to openQA project or let them know. It seems that the change does not take this usage on s390x into consideration which means you have to use die_on_timeout => 0. Now you made the change which branches off s390x. I did not say branching off here is really bad here, but it does force us to make the same change to other arches here to some extent because it looks not very pleasant and leads that way.  Very interesting, we end up having to use die_on_timeout => 0 because of API change. I do not think this is what this API change initially would like to see.  
Maybe you use type_string here instead of script_run ? So we can avoid change to other tests ?

tests/virt_autotest/guest_upgrade_run.pm
guoxuguang: Good point, use with `die_on_timeout =>0` as the simple way. Moreover, `die_on_timeout =>1` means that trigger command and finished this command in expected time. IMHO, use with `die_on_timeout =>1` is so useful for our osd test. But, find out this die_on_timeout =>1 do not work with s390x test. So, we'd better to keep with `die_on_timeout =>1` for non-s390x.

tests/virt_autotest/update_package.pm
guoxuguang: Same the above comment. we'd better to keep with die_on_timeout =>1 for non-s390x.

guoxuguang: > @guoxuguang I think you'd better raise this issue to openQA project or let them know. It seems that the change does not take this usage on s390x into consideration which means you have to use die_on_timeout => 0. Now you made the change which branches off s390x. I did not say branching off here is really bad here, but it does force us to make the same change to other arches here to some extent because it looks not very pleasant and leads that way. Very interesting, we end up having to use die_on_timeout => 0 because of API change. I do not think this is what this API change initially would like to see. API option should be optional to use instead have to use. Maybe you use type_string or other script_xxxx here instead of script_run ? So we can avoid change to other tests ?

Sure, great idea. I need to create a ticket to openQA tools team to let them know that maybe this new change does not take care this usage on s390x. 
Also, I had confirmed that `type_string()` still does no work for svirt backend. So far, just  `virt_utils::lpar_cmd()` and `script_run() `work with svirt backend.  yes, I will try to confirm the other API work with svirt backend. thanks so much for your great help and suggestion.

tests/virt_autotest/guest_upgrade_run.pm
Julie-CAO: I agree with Wayne. it is more acceptable that script_run() has different behavior between s390 with ipmi/qemu backend.

guoxuguang: yes, this PR just only the workaround for all s390x test. 

Also, open a ticket [poo#107164](https://progress.opensuse.org/issues/107164) to openQA tools team now.

alice-suse:
@waynechen55  @Julie-CAO Do you agree to have this PR merged as it is now?

Julie-CAO:
> @waynechen55 @Julie-CAO Do you agree to have this PR merged as it is now?

I agree. and revert it if the script_run() is considered for s390.

PR #14240, opened 2022-02-14
Increase robot timeout and use serial terminal

dzedro:
- Fail: https://openqa.suse.de/tests/8155655#step/sys_param_check/51
- Verification run: https://dzedro.suse.cz/tests/20940


ricardobranco777 (approved):
LGTM. Thanks!

PR #14241, opened 2022-02-14
Add feature to YAML scheduler to reuse env paths

jknphy:
Add feature to YAML scheduler to reuse env flows and avoid repetition.
In principle should not need conditional schedule as there are defaults defined for each product.
Allows to have only one test suite once all the whimsical differences among test suites has been resolved.
Current mechanism using a list of test modules remains untouched.

- Related ticket: [poo#105043](https://progress.opensuse.org/issues/105043)
- Job group MR: [mr#396](https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/396)
- Verification run: [SLE-15-SP4](https://openqa.suse.de/tests/overview?distri=sle&build=jknphy%2Fos-autoinst-distri-opensuse%23poc_enh_yaml_sch&version=15-SP4) | [Tumbleweed](https://openqa.opensuse.org/tests/overview?build=jknphy%2Fos-autoinst-distri-opensuse%23poc_enh_yaml_sch&distri=opensuse&version=Tumbleweed)

Note: More tickets will be created in order to: improve CI in gitlab to check for existence of default+flows files), unit tests, utility script to generate schedules based on default+flows, documentation, etc.


JRivrain:
Looks nice ! I regret though that we do not know what default are used by just reading the yaml, we have to check the variable in Openqa for that. I'd suggest to define that variable directly in the YAML itself, that way would be more similar to using eg "use base" in perl, more readable I think, in particular for whom is not familiar with our environment.
Also, maybe we could have common defaults for sle and TW, as we do for the framework, in order to apply the factory-first logic ? Though this second point may be difficult if schedules differ too much. Or at least a SLE default directory, so that when a new SLE version/SP arrives, we will not need to copy all the schedule files to a new directory, but just a few overrides if necessary.

jknphy:
Thanks for the feedback @JRivrain . The problem with using the include in the yaml instead of loading programatically from a setting defined in Job Group yaml, is that if you have more granularity, like for example 4 or 5 different test modules for bootloader (like it was before unifying to one) then you would need to create 4 or 5 different yaml schedules for a test suite. That is an example of worst case but currently you can find the same when instead of default is used remote.yaml for example, more and more version of the same yaml for the same test suite are needed. Let me know if you have any thoughts how to solve that ... I agree with you that is no so transparent.

IMO the factory first should be applied to the Test Case, and what is a test case? do some action and then validate expected result vs current result. The defaults are the superfluous part, the part that is not unique, which support the navigation to the point where you need it. 
You also touched an interesting point, what happen for SP5 or even SP3? One thing that seems that we did right is to freeze this schedules in maintenance, because we avoid breaks, so not mixing configuration among products seems to be a good thing at the end and not a patch, like we thought at the beginning. Those defaults are really different already with different defaults, different screens specially in openSUSE as you well pointed, and it will be even less consistent when they grow, so we would not be able to distinguish which is applicable where. But regarding SLE I think you are right, I can rename folder to not be SP specific like:

`schedule/yast/defaults/sle/` and later on we will create `schedule/yast/defaults/sle_15-sp5/` where we override with the previous folder for the changes. But what about SP3? we might want to do the same when we will take over... I guess we will need to check and move generic sle to SP3 and see that all is working. make sense?

jknphy:
@rakoenig @ge0r @JRivrain please take a look.

PR #14242, opened 2022-02-14
Retry 'host' command to acquire IP address

lpalovsky:
There are sporadic failures on different aarch64 based cluster tests 
caused by command 'host' not being able to get an IP address on first 
try. This PR makes the command being re-tried 2 times before failing.

- Related ticket: https://jira.suse.com/browse/TEAM-5203
- Verification run: 
- https://mordor.suse.cz/tests/2783#step/drbd_passive/3
- https://mordor.suse.cz/tests/2780#step/drbd_passive/3

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -204,8 +204,15 @@ Returns 0 on failure.
 
 sub get_ip {
     my $node_hostname = shift;
-    my $node_ip = get_var('USE_SUPPORT_SERVER') ? script_output "host -t A $node_hostname" :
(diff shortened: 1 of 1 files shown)


alvarocarvajald (approved):
Aww. I liked it better with `? :` instead of the if-else-block, but it does LGTM. :)

dzedro (approved):

lpalovsky:
> 
@alvarocarvajald  I like the shortened if-else too, I was just afraid the code will be less readable with all the stuff. I can change it back if it is fine like that :D

bs-github:
I think both `if else / ? :` would be fine.

ricardobranco777 (approved):
LGTM

PR #14254, opened 2022-02-15
Extend supportconfig timeout

lpalovsky:
HA cluster tests tend to fail on 'check_logs' module because of 
'supportconfig' command takes longer than usual and times out. This PR 
extends command timeout from 180s to 300s.

- Verification run:
https://openqa.suse.de/tests/8173326#dependencies
https://openqa.suse.de/tests/8173329#dependencies


ricardobranco777 (approved):
LGTM

PR #14266, opened 2022-02-16
Improve libyui rest api not configured msg

ge0r:

- Related ticket: No ticket
- Needles: No needles
- Verification run: pending


lib/YuiRestClient.pm
```diff
@@ -25,8 +25,8 @@ my $host;
 
 sub get_app {
     my (%args) = @_;
-    die "No environment for libyui REST API has been configured.",
-      " Please schedule test module `setup_libyui.pm` in previous steps." unless YuiRestClient::is_libyui_rest_api();
+    die "This module is using LibYUI REST API but there is no such configuration in place. ",
```
jknphy, line 28: the previous message was correct and more clear in this specific point when reached after setup_libyui (although I would prefer not to mention specific test modules names, just setup libyui, for example, in case name would change). 

I think the problem is that we need to control the error in setup_libyui.pm, there is not point to get the app in `run` if the variable is not set, there is where we can provide a specific message "do you forget to set this variable? blablab". wdyt?

ge0r: I agree, I thought that yesterday on the discussion we had decided to keep things simple and just alter the message, my bad.

tests/installation/setup_libyui.pm
```diff
@@ -15,6 +15,7 @@ use testapi;
 use YuiRestClient;
 
 sub run {
+    die 'Module setup_libyui requires YUI_REST_API variable to be set to 1.', unless check_var('YUI_REST_API', '1');
```
jknphy, line 18: For maintenance of the code I would still recommend to not include the exact name of the module (but minor thing). os-autoinst and other software use some question, like "did you forget to ...blablab?" But this is just a suggestion, we can keep it like that if you like it more.

`YUI_REST_API` is not the API version afir, so it is just a boolean, so in Perl by default we never check with `1`. We could put there `3` but the goal is not to verify our test data or test variable here but give a good hint. In most of the places we just use get_var. Also you mentioned offline to use `get_required_var()` That also should be pretty clear for any openQA tester.

And the most important part, we still need to provide error in the get_app for the original error in the progress ticket, where this module is not schedule.

ge0r: > For maintenance of the code I would still recommend to not include the exact name of the module (but minor thing). os-autoinst and other software use some question, like "did you forget to ...blablab?" But this is just a suggestion, we can keep it like that if you like it more.

You make a good point on the topic of exact same name of the module, I removed it.

> `YUI_REST_API` is not the API version afir, so it is just a boolean, so in Perl by default we never check with `1`. We could put there `3` but the goal is not to verify our test data or test variable here but give a good hint. In most of the places we just use get_var. Also you mentioned offline to use `get_required_var()` That also should be pretty clear for any openQA tester. But as it only croaks this https://github.com/os-autoinst/os-autoinst/blob/dfa0cab76ed7b14166870f9dcfcf7ce13dd81ce1/testapi.pm#L763 It doesn't admit to pass it a message to tell the user if they forget to set the variable should be even better IMO.
> 
> And the most important part, we still need to provide error in the `get_app` for the original error in the progress ticket, where this module is not schedule. I would recommend to show a couple of verification run with both possible errors.

This was part of the previous (merged) PR, so the get_app error is still addressed as it should. I have attached a verification run for this case as well, for clarity

jknphy: > > For maintenance of the code I would still recommend to not include the exact name of the module (but minor thing). os-autoinst and other software use some question, like "did you forget to ...blablab?" But this is just a suggestion, we can keep it like that if you like it more.
> 
> You make a good point on the topic of exact same name of the module, I removed it.
> 
> > `YUI_REST_API` is not the API version afir, so it is just a boolean, so in Perl by default we never check with `1`. We could put there `3` but the goal is not to verify our test data or test variable here but give a good hint. In most of the places we just use get_var.
> 
> I presumed that in the sake of consistency we should require the value of 1, as this is the value of `YUI_REST_API` everywhere, but better err on the side of caution on this, as at some point it could be set to different values.
> 
> > And the most important part, we still need to provide error in the `get_app` for the original error in the progress ticket, where this module is not schedule. I would recommend to show a couple of verification run with both possible errors.
> 
> This was part of the previous (merged) PR, so the get_app error is still addressed as it should. I have attached a verification run for this case as well, for clarity

my bad, I need to get used to have multiple PRs, it is easier one PR with multiple commits for reviewer. I forgot what was already merged :)

jknphy (approved):
Even if it was simple, it was tricky to find the right messages.
Thanks for the time for searching the best solution.

PR #14231, opened 2022-02-11
[WIP] Merge podman_image and docker_image test

ilausuch:
Merge both tests in a single test

- Related ticket: https://progress.opensuse.org/issues/101210
- Verification run:

PR #14268, opened 2022-02-16
Fix container schedule for MicroOS

jlausuch:
This fixes the regression caused by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14231
I proposed to run some VRs in O3, but forgot to propose MicroOS which follows its own schedule file... which IMO we should change for this things not to happen. It would be nice that ANY container job (in OSD/O3, SLES, TW, Leap, SLE Micro, MicroOS, etc) uses the same schedule `main_containers.pm`.

### DIFF
--- a/products/microos/main.pm
+++ b/products/microos/main.pm
@@ -63,9 +63,12 @@ sub load_feature_tests {
         loadtest 'console/kubeadm';
     }
     elsif (check_var 'SYSTEM_ROLE', 'container-host') {
+        my $args = OpenQA::Test::RunArgs->new();
+        my $runtime = get_required_var('CONTAINER_RUNTIME');
+        $args->{runtime} = $runtime;
         loadtest 'microos/toolbox';
         loadtest 'containers/podman';
-        loadtest 'containers/podman_image';
+        loadtest('containers/image', run_args => $args, name => "image_$runtime");
     }
 }
 


products/microos/main.pm
```diff
@@ -63,9 +63,11 @@ sub load_feature_tests {
         loadtest 'console/kubeadm';
     }
     elsif (check_var 'SYSTEM_ROLE', 'container-host') {
+        my $args = OpenQA::Test::RunArgs->new();
+        $args->{runtime} = 'podman';
```
DimStar77, line 67: in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14231 you used `CONTAINER_RUNTIME` setting, which seems to be also set on the microos failures I looked at. should we use them here too?

Alternatively, forcing podman sounds ok, as it's just the same as was done before (podman_image was hardcoded)

DimStar77 (approved):

products/microos/main.pm
jlausuch: yes, better. Done.

DimStar77 (approved):

jlausuch:
VR: https://openqa.opensuse.org/tests/2192242

pdostal:
The `main_containers::load_image_test()` could be exported and reused.

jlausuch:
> The `main_containers::load_image_test()` could be exported and reused.

not that easy, we need to do some adaptations as MicroOS tests run host test and image tests with `untested_images=1` ...

PR #14269, opened 2022-02-17
Fix script_run timeout in patch_sle

lemon-suse:
We expect script_run() return undef on command timeout instead of raising an exception, so add argument die_on_timeout => 0.

- Related ticket: https://progress.opensuse.org/issues/106817
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/8183347#step/patch_sle/16

### DIFF
--- a/lib/migration.pm
+++ b/lib/migration.pm
@@ -1,4 +1,4 @@
-# Copyright 2017-2021 SUSE LLC
+# Copyright 2017-2022 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package migration;
@@ -47,7 +47,7 @@ sub setup_sle {
 
     # Enable Y2DEBUG for error debugging
     enter_cmd "echo 'export Y2DEBUG=1' >> /etc/bash.bashrc.local";
-    script_run "source /etc/bash.bashrc.local";
+    script_run("source /etc/bash.bashrc.local", die_on_timeout => 0);
 }
 
 sub setup_migration {


lib/migration.pm
```diff
@@ -47,7 +47,7 @@ sub setup_sle {
 
     # Enable Y2DEBUG for error debugging
     enter_cmd "echo 'export Y2DEBUG=1' >> /etc/bash.bashrc.local";
-    script_run "source /etc/bash.bashrc.local";
```
coolgw, line 50: base [open.qa doc](http://open.qa/api/testapi/) i am confusing now.
"With die_on_timeout -1 (default) a warning will be printed to log."   
The default -1 will not  throw an exception.

lemon-suse: From the email for the implement of script_run, 'If your OpenQA code requires that script_run() must return undef on
command timeout instead of raising an exception, you need to add a new
named argument to preserve the old behavior:
script_run(..., die_on_timeout => 0);'
If the doc has say this clear, I think it should be updated. Since on real log, it will fail for timeout with default setting.

tinawang123: With die_on_timeout -1 (default) a warning will be printed to log. To avoid this warning, set it explicit to 0. With die_on_timeout equal to 1, this command throw an exception, if timeout expires.
-1 just print a log not throw an exception. Why should we change this?

rfan1:
I suggest use asser_script_run ...

lib/migration.pm
lemon-suse: I think the http://open.qa/api/testapi/#_script_run is old, please don't think base it, just on the recent PR for script_run https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14027  and check the related mail named 'script_run() API change in OpenQA'.

lemon-suse:
> I suggest use assert_script_run ...

In fact, we use script_run originally is for not care the results whether pass or fail. WDYT?

rfan1:
> > I suggest use assert_script_run ...
> 
> In fact, we use script_run originally is for not care the results whether pass or fail. WDYT?

Yup, then no problem. LGTM

rfan1 (approved):
LGTM

PR #14274, opened 2022-02-17
wicked: Install systemd-coredump only if available

cfconrad:
We do not have systemd-coredump available on SLE-12 qam test runs,
so just skip this functionality.

- Ticked: https://progress.opensuse.org/issues/106963#note-7

- Verification run: https://openqa.suse.de/tests/8185777

### DIFF
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -859,7 +859,7 @@ sub coredumpctl_has_debug {
 
 sub prepare_coredump {
     my $self = shift;
-    zypper_call('--quiet in systemd-coredump');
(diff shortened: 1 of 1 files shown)


cfconrad:
plz wait with merge, I wonder why there is no `systemd-coredump` in general but docu say there should be
 * https://scc.suse.com/packages?name=SUSE%20Linux%20Enterprise%20Server&version=12.5&arch=x86_64&query=systemd-coredump&module=
 * https://www.suse.com/support/kb/doc/?id=000018634

cfconrad:
VR's with latest change: 
  * https://openqa.suse.de/tests/8185917
  * https://openqa.suse.de/t8185922

jlausuch (approved):
LGTM. Safest option :)

PR #14276, opened 2022-02-17
Containers: don't use script_retry when creating containers with same name

jlausuch:
When the first time fails, it will try to create a new container using the
same name, and it fails with
```
You have to remove that container to be able to reuse that name.:
that name is already in use
```

Example of failure: https://openqa.opensuse.org/tests/2193730#step/image_podman/130

### DIFF
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -197,6 +197,7 @@ sub test_opensuse_based_image {
 
 sub test_zypper_on_container {
     my ($runtime, $image) = @_;
+    record_info('zypper tests', 'Basic zypper commands in the container.');
 
     die 'Argument $image not provided!' unless $image;
     die 'Argument $runtime not provided!' unless $runtime;
@@ -208,11 +209,11 @@ sub test_zypper_on_container {
         record_info('bsc#1192941', 'bsc#1192941 - zypper-docker entrypoint confuses program arguments');
     }
     validate_script_output("$runtime run -i --entrypoint '' $image zypper lr -s", sub { m/.*Alias.*Name.*Enabled.*GPG.*Refresh.*Service/ }, timeout => 180);
-    script_retry("$runtime run -i --name 'refreshed' --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
+    assert_script_run("$runtime run -t -d --name 'refreshed' --entrypoint '' $image bash", timeout => 300);
+    script_retry("$runtime exec refreshed zypper -nv ref", timeout => 600, retry => 3, delay => 60);
     assert_script_run("$runtime commit refreshed refreshed-image", timeout => 120);
     assert_script_run("$runtime rm -f refreshed");
-    script_retry("$runtime run -i --name 'refreshed-image' --rm --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
-    record_info "The End", "zypper test completed";
+    script_retry("$runtime run -i --rm --entrypoint '' refreshed-image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
 }
 
 sub exec_on_container {


Vogtinator:
I suppose this would work if it did `podman rm` as cleanup between attempts?

Vogtinator:
> And adding --rm?
> 
> ```
> --rm                             Automatically remove the container when it exits
> ```

That also removes it on success, but the test module wants to use it later on.

ilausuch:
> > And adding --rm?
> > ```
> > --rm                             Automatically remove the container when it exits
> > ```
> 
> That also removes it on success, but the test module wants to use it later on.

adding -d with some silly loop in background?

lib/containers/container_images.pm
```diff
@@ -208,10 +208,10 @@ sub test_zypper_on_container {
         record_info('bsc#1192941', 'bsc#1192941 - zypper-docker entrypoint confuses program arguments');
     }
     validate_script_output("$runtime run -i --entrypoint '' $image zypper lr -s", sub { m/.*Alias.*Name.*Enabled.*GPG.*Refresh.*Service/ }, timeout => 180);
-    script_retry("$runtime run -i --name 'refreshed' --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
+    assert_script_run("$runtime run -i --name 'refreshed' --entrypoint '' $image zypper -nv ref", timeout => 600);
     assert_script_run("$runtime commit refreshed refreshed-image", timeout => 120);
     assert_script_run("$runtime rm -f refreshed");
-    script_retry("$runtime run -i --name 'refreshed-image' --rm --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
+    assert_script_run("$runtime run -i --name 'refreshed-image' --rm --entrypoint '' $image zypper -nv ref", timeout => 600);
```
jlausuch, line 214: I think this should be 
```
$runtime run -i --rm  --entrypoint '' localhost/refreshed-image zypper -nv ref
```
Otherwise, we are just repeating the same command as in line 211...  I guess, what we want to do is to check if `zypper ref` is "faster" in the refreshsed-container...

jlausuch:
See inline comment.. I think we are repeating the same command twice...

jlausuch:
VR: http://fromm.arch.suse.de/tests/5629

jlausuch:
Latest code creates the container and the uses script retry to run `zypper ref` on that running container. This way we ensure we don't create a new container every time with the same name in the script_retry.

lib/containers/container_images.pm
```diff
@@ -208,11 +209,11 @@ sub test_zypper_on_container {
...
     }
     validate_script_output("$runtime run -i --entrypoint '' $image zypper lr -s", sub { m/.*Alias.*Name.*Enabled.*GPG.*Refresh.*Service/ }, timeout => 180);
-    script_retry("$runtime run -i --name 'refreshed' --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
+    assert_script_run("$runtime run -t -d --name 'refreshed' --entrypoint '' $image bash", timeout => 300);
+    script_retry("$runtime exec refreshed zypper -nv ref", timeout => 600, retry => 3, delay => 60);
     assert_script_run("$runtime commit refreshed refreshed-image", timeout => 120);
     assert_script_run("$runtime rm -f refreshed");
-    script_retry("$runtime run -i --name 'refreshed-image' --rm --entrypoint '' $image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
-    record_info "The End", "zypper test completed";
+    script_retry("$runtime run -i --rm --entrypoint '' refreshed-image zypper -nv ref", timeout => 300, retry => 3, delay => 60);
```
pdostal, line 216: Why don't we use `$runtime exec` over here as well?

lib/containers/container_images.pm
pdostal: We check if the `zypper` is available also in derivative image.

lib/containers/container_images.pm
jlausuch: What I understand from this flow is:
1) Run zypper ref in container and commit the state of the container to a new local image `refreshed-image`.
2) Run another container from that committed local image

For 1) we CAN'T use script_retry because we force the `--name` in the command. If the command fails or timesout, the container might have been created and the next retry will fail. So, that's why I opted for creating the container and leave it running, and then retry ONLY the `zypper ref` command. Then, commit.

For 2) we dont' need to provide a `--name` to the container, because we create it from a local image and don't care about the container name. Docker/podman will create a random name. So, we can use `script_retry` directly there.

pdostal (approved):
All good, this is just perfect :+1:

mloviska (approved):

PR #14282, opened 2022-02-18
Provide automatic review checklist based on changed paths

okurz:
See
https://github.com/os-autoinst/os-autoinst/pull/1958#issuecomment-1044231841
for an example in os-autoinst.

### DIFF
--- a/.github/checklist.yml
+++ b/.github/checklist.yml
@@ -1,0 +1,7 @@
+paths:
+  "lib/**.pm":
+    - Consider adding or extending unit tests in t/
+  "t/data/test_schedule.out":
+    - Ensure openSUSE/SLE release managers are ok with the change in staging schedule
+  "tests/**.pm":
+    - Consider conducting at least one [verification run](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#preparing-a-new-pull-request)
--- a/.github/workflows/checklist.yml
+++ b/.github/workflows/checklist.yml
@@ -1,0 +1,15 @@
+---
+# yamllint disable-line rule:truthy
+on: [pull_request_target]
+jobs:
+  checklist_job:
+    runs-on: ubuntu-latest
+    name: Checklist job
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Checklist
+        uses: wyozi/contextual-qa-checklist-action@master
+        with:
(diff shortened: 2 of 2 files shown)


.github/checklist.yml
```diff
@@ -0,0 +1,5 @@
+paths:
+  "lib/**.pm":
+    - Consider adding or extending unit tests in t/
```
foursixnine, line 3: I like where your heart is at, not sure if it's the  right way though.

foursixnine (changes requested):
Since we're at it, please add one for tests/ for verification runs, we can disable that in the future if it becomes too noisy.

I'd also consider another one for the data directory, asking for a comment on the source

.github/checklist.yml
okurz: Well, it says "consider", not "you *must*" :)

foursixnine: > Well, it says "consider", not "you _must_" :)

That's the le's experiment part

okurz:
> Since we're at it, please add one for tests/ for verification runs, we can disable that in the future if it becomes too noisy.

I did that with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14282/files#diff-33df71e78183e3bfe7b5c4ace99dd003696f4472982d2b1eee59af81ac7879d5R6

> I'd also consider another one for the data directory, asking for a comment on the source

Sorry, wasn't sure what you mean there. I have not added that yet. I suggest we do it later. Let's collect feedback with what is suggested for now, ok?

foursixnine:
> > Since we're at it, please add one for tests/ for verification runs, we can disable that in the future if it becomes too noisy.
> 
> I did that with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14282/files#diff-33df71e78183e3bfe7b5c4ace99dd003696f4472982d2b1eee59af81ac7879d5R6
> 
> > I'd also consider another one for the data directory, asking for a comment on the source
> 
> Sorry, wasn't sure what you mean there. I have not added that yet. I suggest we do it later. Let's collect feedback with what is suggested for now, ok?

I meant things added in the Data directory, but it isn't something critical anyway :)

foursixnine (approved):

okurz:
I have to assume that everybody else either gives implicit approval or does not care. So I merge after I have at least received the approval by foursixnine. A more explicit message would be helpful though.

PR #14283, opened 2022-02-18
[Proposal] Refactor add_suseconnect_product with retries

jlausuch:
This method is used in multiple places, and we have observed that
it fails randomly due to network issues (timeouts, SCC not reachable,
etc). In most cases, restarting the test helps (unless there is a
major outage that tages long to be fixed).

This code does a similar thing what script_retry does, but with a
bit of more logic and info messages.

Before:
We do 1 time `SUSEConnect -p` and ONLY retry 1 more time if PH.

After:
We do X retries (default 3) for ANY product.
And keep the same fail message if BETA=1 and PH at the end.

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -174,25 +174,29 @@ sub add_suseconnect_product {
     $version //= '${VERSION_ID}';
     $arch //= '${CPU}';
     $params //= '';
-    $retry //= 0;    # run SUSEConnect a 2nd time to workaround the gpg error due to missing repo key on 1st run
+    $retry //= 3;    # run SUSEConnect $retry times before failing
+    $timeout //= 300;
 
     # some modules on sle12 use major version e.g. containers module
     my $major_version = '$(echo ${VERSION_ID}|cut -c1-2)';
     $version = $major_version if $name eq 'sle-module-containers' && is_sle('<15');
+    record_info('SCC product', "Activating product $name");
 
-    my $result = script_run("SUSEConnect -p $name/$version/$arch $params", $timeout);
-    if ($result != 0 && $retry) {
-        if ($name =~ /PackageHub/) {
-            record_soft_failure 'bsc#1124318 - Fail to get module repo metadata - running the command again as a workaround';
-
+    my $try_cnt = 0;
+    while ($try_cnt++ < $retry) {
+        eval { assert_script_run("SUSEConnect -p $name/$version/$arch $params", timeout => $timeout); };
+        if ($@) {
+            record_info('retry', "SUSEConnect failed to activate the module $name. Retrying...");
+            sleep 60 * $try_cnt;    # we wait a bit longer for each retry
         }
-
-        my $fail_message = ($name =~ /PackageHub/ && check_var('BETA', '1')) ? "PackageHub installation might fail in early development" : undef;
-        assert_script_run("SUSEConnect -p $name/$version/$arch $params", $timeout, $fail_message);
-
-    } elsif ($result && !$retry) {
-        die "SUSEConnect failed activating module $name with exit code (see log output): $result.";
+        else {
+            return 1;
+        }
     }
+    if ($name =~ /PackageHub/ && check_var('BETA', '1')) {
+        record_info('INFO', 'PackageHub installation might fail in early development');
+    }
+    die "SUSEConnect failed activating module $name after $$retry retries.";
 }
 
 =head2 ssh_add_suseconnect_product


foursixnine (approved):
I'm unsure if waiting 60 seconds will be enough or waiting a bit longer would be a better option (say 120s), but overall waiting makes much more sense than constantly retrying specially when the network is loaded.

jlausuch:
> I'm unsure if waiting 60 seconds will be enough or waiting a bit longer would be a better option (say 120s), but overall waiting makes much more sense than constantly retrying specially when the network is loaded.

Yes, maybe 1min is too short. I would even go for 5 min wait :D
Let's see what others think.

mloviska:
> > I'm unsure if waiting 60 seconds will be enough or waiting a bit longer would be a better option (say 120s), but overall waiting makes much more sense than constantly retrying specially when the network is loaded.
> 
> Yes, maybe 1min is too short. I would even go for 5 min wait :D Let's see what others think.

Or increase the value each iteration.

jlausuch:
Some early VRs: (search for `SCC product`)
https://openqa.suse.de/tests/8190066
https://openqa.suse.de/tests/8190065
https://openqa.suse.de/tests/8190011

lib/registration.pm
```diff
@@ -174,25 +174,31 @@ sub add_suseconnect_product {
...
+            sleep 60;
         }
-
-        my $fail_message = ($name =~ /PackageHub/ && check_var('BETA', '1')) ? "PackageHub installation might fail in early development" : undef;
-        assert_script_run("SUSEConnect -p $name/$version/$arch $params", $timeout, $fail_message);
-
-    } elsif ($result && !$retry) {
-        die "SUSEConnect failed activating module $name with exit code (see log output): $result.";
+        else {
+            return 1;
```
foursixnine, line 195: I´d add an info window here (you know, thinking about the reviewers) that there has been too many trials, so aborting.

pdostal: Isn't the line 201 enough?

pdostal (approved):

ilausuch (approved):

lib/registration.pm
jlausuch: No, no, but this line is to return if the command succeded.. something like returning 0 in bash, but in perl we need to return "true" if ok. This condition will hit when `$@` is `undef`, which means that there are no errors from the `eval` section.

foursixnine: @jlausuch exactly, I need more waay more coffee today xD

jlausuch:
> > > I'm unsure if waiting 60 seconds will be enough or waiting a bit longer would be a better option (say 120s), but overall waiting makes much more sense than constantly retrying specially when the network is loaded.
> > 
> > 
> > Yes, maybe 1min is too short. I would even go for 5 min wait :D Let's see what others think.
> 
> Or increase the value each iteration.

not bad idea :)
iteration_num * 60? e.g. 1) 60  2) 120  3) 180 ...

tjyrinki (approved):
If we can't get SCC or network to be 100% fault-free, this would be absolutely a great improvement.

jlausuch:
> If we can't get SCC or network to be 100% fault-free, this would be absolutely a great improvement.

Thanks! :)
I still have some hopes we will reach that fault-free level... but for now, we need to settle for some workaround.

jlausuch:
Shall we force default the `$timeout` in this method to e.g. 300s instead of the default by `assert_script_run` (90s)?

asmorodskyi:
I think enough people approve it already so I can merge

okurz:
> > If we can't get SCC or network to be 100% fault-free, this would be absolutely a great improvement.
> 
> Thanks! :) I still have some hopes we will reach that fault-free level... but for now, we need to settle for some workaround.

With this change a lot of motivation to improve will be gone. Did any of you raise the problems to the SCC team? I bet collaborating together can achieve great things by finding out where things might not be 100% fault-free yet. I am sure that a much higher stability can be achieved but only if issues are reported.

jlausuch:
> > > If we can't get SCC or network to be 100% fault-free, this would be absolutely a great improvement.
> > 
> > 
> > Thanks! :) I still have some hopes we will reach that fault-free level... but for now, we need to settle for some workaround.
> 
> With this change a lot of motivation to improve will be gone. Did any of you raise the problems to the SCC team? I bet collaborating together can achieve great things by finding out where things might not be 100% fault-free yet. I am sure that a much higher stability can be achieved but only if issues are reported.

I think this situation hasn't changed for long time... And.. are you sure the problem is in SCC network? It would be multiple places... random slowdowns in our openQA worker infrastructure as well, who knows? if we are to blame SCC we need proof for that.

okurz:
> I think this situation hasn't changed for long time... And.. are you sure the problem is in SCC network? It would be multiple places... random slowdowns in our openQA worker infrastructure as well, who knows? if we are to blame SCC we need proof for that.

I did not suggest to blame anyone :) I am suggesting to collaborate on the investigation so that we can identify which component might be the culprit.

jlausuch:
It seems it doesn't work as expected: https://openqa.suse.de/tests/8252341#step/suseconnect_scc/42

jlausuch:
Followu-up: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14402

PR #14291, opened 2022-02-20
Reset timeout for migraiton group test cases and upload y2log in post failed hook

coolgw:
Reset timeout for migraiton group test cases and upload y2log in post failed hook

- Related ticket: https://progress.opensuse.org/issues/106092
- Needles: na
- Verification run: http://openqa.suse.de/t8189592
The error only happen during heavy load of worker, so VR can not show the actual failed situation.


okurz:
Your git commit messages included two typos in a single line. Please be a bit more careful with proper messages and use proper pair-review, i.e. await approval from others.

PR #14295, opened 2022-02-21
Validate output retry

jlausuch:
Allow running validate_script_output with given number of retries,
similar to script_retry. It's basically the same function but
giving the chance to run the command $retry times with $delay
waiting time between them. This might be useful for commands
that are not 100% stable due to network issues and we need to try
2 or 3 times until it works.

VR: https://openqa.suse.de/tests/8203396#

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -86,6 +86,7 @@ our @EXPORT = qw(
   install_patterns
   common_service_action
   script_output_retry
+  validate_script_output_retry
   get_secureboot_status
   assert_secureboot_status
   susefirewall2_to_firewalld
@@ -1663,6 +1664,55 @@ sub script_output_retry {
     die("Waiting for Godot: $cmd") if $die;
 }
 
+
+=head2 validate_script_output_retry
+
+ validate_script_output_retry($cmd, $check, [retry => $retry], [delay => $delay], [timeout => $timeout]);
+
+Repeat command until validate_script_output succeeds or die. Return the output of the command on success.
+
+C<$retry> refers to the number of retries and defaults to C<10>.
+
+C<$delay> is the time between retries and defaults to C<30>.
+
+If the command doesn't succeed after C<$retry> retries,
+this function will die.
+
+Example:
+
+ validate_script_output_retry('ping -c1 -W1 machine', m/1 packets transmitted/, retry => 5, delay => 60);
+
+=cut
+sub validate_script_output_retry {
+    my ($cmd, $check, %args) = @_;
+    $args{retry} //= 10;
+    $args{delay} //= 30;
+    $args{timeout} //= 90;
+    $args{proceed_on_failure} //= 1;
+    my $retry = delete $args{retry};
+    my $delay = delete $args{delay};
+    my $timeout = delete $args{timeout};
+    my $ret;
+    my $exec = "timeout --foreground --kill-after 5s ${timeout}s";
+    # Exclamation mark needs to be moved before the timeout command, if present
+    if (substr($cmd, 0, 1) eq "!") {
+        $cmd = substr($cmd, 1);
+        $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
+        $exec = "! $exec";
+    }
+    $exec = "$exec $cmd";
+    $timeout += 8;    # timeout for script_run must be larger than for the 'timeout ...' command
+
+    for (1 .. $retry) {
+        eval { $ret = validate_script_output($exec, $check, $timeout, %args); };
+        return $ret if (defined($ret));
+        record_info("Retry", "validate_script_output failed or timed out. Retrying...");
+        sleep $delay;
+    }
+    die("Can't validate output after $retry retries");
+}
+
+
 =head2 script_run_interactive
 
  script_run_interactive($cmd, $prompt, $timeout);
@@ -1764,12 +1814,12 @@ sub create_btrfs_subvolume {
     assert_script_run("rm -fr /tmp/arm64-efi/");
 }
 
-=head2 create_raid_loop_device 
-    
+=head2 create_raid_loop_device
+
  create_raid_loop_device([raid_type => $raid_type], [device_num => $device_num], [file_size => $file_size]);
 
 Create a raid array over loop devices.
-Raid type is C<$raid_type>, using C<$device_num> number of loop device, 
+Raid type is C<$raid_type>, using C<$device_num> number of loop device,
 with the size of each device being C<$file_size> megabytes.
 
 Example to create a RAID C<5> array over C<3> loop devices, C<200> Mb each:
--- a/lib/containers/container_images.pm
+++ b/lib/containers/container_images.pm
@@ -172,8 +172,8 @@ sub test_opensuse_based_image {
                 # we set --entrypoint specifically to the zypper plugin to avoid bsc#1192941
                 my $plugin = '/usr/lib/zypp/plugins/services/container-suseconnect-zypp';
                 validate_script_output("$runtime run --entrypoint $plugin -i $image -v", sub { m/container-suseconnect version .*/ }, timeout => 180);
-                validate_script_output("$runtime run --entrypoint $plugin -i $image lp", sub { m/.*All available products.*/ }, timeout => 180);
-                validate_script_output("$runtime run --entrypoint $plugin -i $image lm", sub { m/.*All available modules.*/ }, timeout => 180);
+                validate_script_output_retry("$runtime run --entrypoint $plugin -i $image lp", sub { m/.*All available products.*/ }, retry => 5, delay => 60, timeout => 300);
+                validate_script_output_retry("$runtime run --entrypoint $plugin -i $image lm", sub { m/.*All available modules.*/ }, retry => 5, delay => 60, timeout => 300);
             }
         } else {
             record_info "non-SLE host", "This host ($host_id) does not support zypper service";
--- a/t/05_utils_auxiliary.t
+++ b/t/05_utils_auxiliary.t
@@ -33,4 +33,25 @@ subtest 'script_retry' => sub {
     dies_ok { script_retry('sleep 10', retry => 1, delay => 0, timeout => 1) } 'script_retry(sleep) is expected to die';
 };
 
+
+subtest 'validate_script_output_retry' => sub {
+    my $module = Test::MockModule->new('testapi');
+
+    my $basetest = Test::MockModule->new('basetest');
+    $basetest->noop('record_resultfile');
+    $autotest::current_test = new basetest;
+
+    $module->mock('script_output', sub { 'foo' });
+    lives_ok { validate_script_output_retry('echo foo', qr/foo/, retry => 2) } 'Do not throw exception';
+    throws_ok { validate_script_output_retry('echo foo', qr/bar/, retry => 2, delay => 0) } qr/validate output/, 'Exception thrown';
+
+    my @results;
+    $module->mock('script_output', sub { shift @results });
+
+    @results = qw(1 2 3 foo);
+    lives_ok { validate_script_output_retry('echo foo', qr/foo/, retry => 4, delay => 0) } 'Success on 4th retry';
+    @results = qw(1 2 3 foo);
+    throws_ok { validate_script_output_retry('echo foo', qr/foo/, retry => 3, delay => 0) } qr/validate output/, 'Not enough retries';
+};
+
 done_testing;


lib/utils.pm
```diff
@@ -1658,6 +1659,44 @@ sub script_output_retry {
...
+=cut
+sub validate_script_output_retry {
+    my ($cmd, $check, %args) = @_;
+    my $retry = $args{retry} // 10;
+    my $delay = $args{delay} // 30;
+    my $timeout = $args{timeout} // 90;
+
+    for (1 .. $retry) {
+        eval { validate_script_output($cmd, $check, timeout => $timeout); };
+        if ($@) {
```
asmorodskyi, line 1689: I think it is crucial here to figure out WHY exactly it failed . We don't want to retry when script simply failed because validation failed . ONLY good reason to fail is due to timeout

grisu48 (approved):
LGTM!

lib/utils.pm
jlausuch: Very good point... any ideas how?

cfconrad: I also think in case of timeout, simply retry will cause a console mess, as you are not able to judge if you got the prompt again.
in `wicked::wlan::retry` we do it this way: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/wicked/wlan.pm#L98
With `wicked::wlan:retry()` can can actually call every testapi console commands like: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/wicked/wlan.pm#L282 so you do not need to write a wrapper for all of these commands.


Or you go this `timeout $cmd` approach from other `retry()` functions. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1604

But in general, I think it is better to check some preconditions and then the real command should work. Simply retry might actually hide some errors.

jlausuch: ok, I'll go `timeout $cmd` way. Thanks! What I'm not sure is how to differenciate that the eval failed due to command timeout (to retry) or due to output not validating (to avoid retrying).

foursixnine: I might be mistaken, but `$@` here would have the exception that's thrown by the `croak` in the `wait_serial`  that has the timeout in `distribution::script_output`:

```perl
use Carp;
use feature 'say';
eval {
# warn user (from perspective of caller)
carp "string trimmed to 80 chars";
 
# die of errors (from perspective of caller)
croak "We're outta here!";
};

if ($@){
	say $@;
	say "oops" if $@ =~ /outta/;
}
```

foursixnine (commented):
I think it would be a good idea also, to consider adding it to the `testapi` as it might be a better place to ensure that further changes, won't break the retries

lib/utils.pm
cfconrad: ACK. But it isn't nice, cause your code will fail, once someone change `validated_script_output()`

```suggestion
        return 0 unless defined($@);
        die($@) if ($@ ne 'validate_script_output');
        
        record(...); sleep xxx;
```

lib/utils.pm
```diff
@@ -1658,6 +1659,44 @@ sub script_output_retry {
...
+
+=cut
+sub validate_script_output_retry {
+    my ($cmd, $check, %args) = @_;
+    my $retry = $args{retry} // 10;
+    my $delay = $args{delay} // 30;
+    my $timeout = $args{timeout} // 90;
+
+    for (1 .. $retry) {
+        eval { validate_script_output($cmd, $check, timeout => $timeout); };
```
cfconrad, line 1688: the function `validate_script_output()` allow to pass arguments to the `script_output()` call, e.g. `proceed_on_failure => 1`

```suggestion
    my ($cmd, $check, %args) = @_;
    my %args = compat_args(
        {
            retry => 10,
            delay => 30,
            timeout => 90
        }, [], @_);

    my $retry = delete $args{retry};
    my $delay = delete $args{delay};

    for (1 .. $retry) {
        eval { validate_script_output($cmd, $check, %args); };
```

michaelgrifalconi (approved):
I would argue that every helper function that run 'stuff' on a machine under test should have an optional argument for retry (default 0), instead of creating a new wrapper function around it.

In any case, I like the concept and I am happy with anything that helps to make some tests more stable :) Thanks for proposing it!

lib/utils.pm
jlausuch: For some reason, `$@` is always defined, whatever happens inside the eval (succeeds or fails).

foursixnine: @jlausuch it's meant to: [form perldoc](https://perldoc.perl.org/perlvar#%24%40)

> If no error occurs, eval sets $@ to the empty string.

jlausuch:
> I think it would be a good idea also, to consider adding it to the `testapi` as it might be a better place to ensure that further changes, won't break the retries

For now I would like to keep it here. There is also `script_retry()` and `script_output_retry()` in this file. I would image a new PR in os-autoinst moving all those things at the same time.

lib/utils.pm
cfconrad: thx @foursixnine wasn't aware of that detail !

asmorodskyi (approved):
LGTM

grisu48:
Please also add a unit test to https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/t/05_utils_auxiliary.t

lib/utils.pm
```diff
@@ -1658,6 +1659,43 @@ sub script_output_retry {
...
+    my %args = testapi::compat_args({retry => 10, delay => 30, timeout => 90}, [], @_);
+    my $retry = delete $args{retry};
+    my $delay = delete $args{delay};
+
+    my $exec = "timeout " . ($args{timeout} - 3) . " $cmd";
+    for (1 .. $retry) {
+        my $ret;
+        eval { $ret = validate_script_output($exec, $check, %args); };
+        return $ret if (defined($ret));
+        die($@) if ($@ =~ 'output not validating at');
```
cfconrad, line 1692: Where does this string `output not validating at` comes from? I only found this https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1287 without `at`.

And I'm not sure about this. Because for me it looks like you only wanna retry on timeout. And for me `timeout` is actually an undefined state without some recover code.

lib/utils.pm
```diff
@@ -1658,6 +1659,43 @@ sub script_output_retry {
...
+ validate_script_output_retry('ping -c1 -W1 machine', m/1 packets transmitted/, retry => 5, delay => 60);
+
+=cut
+sub validate_script_output_retry {
+    my ($cmd, $check) = splice(@_, 0, 2);
+    my %args = testapi::compat_args({retry => 10, delay => 30, timeout => 90}, [], @_);
+    my $retry = delete $args{retry};
+    my $delay = delete $args{delay};
+
+    my $exec = "timeout " . ($args{timeout} - 3) . " $cmd";
```
cfconrad, line 1687: with timeout approach, you also need to care about `!` like https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1605

You are doing the same as `script_retry()` but `script_retry()` is adding the timeout to the final function, while you are reducing the time of the command to be executed, see: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1613

lib/utils.pm
jlausuch: when the regex is not matched and `validate_script_output` failed, `$@` contains the whole error log `output not validating at /usr/lib/os-autoinst/testapi.pm line XYZ`  
But I will remove the `at`. 


Yes, as suggested by @asmorodskyi , it doesn't make sense to retry again and again if `validate_script_output` fails because the output doesn't match and not because of timeout. I don't know what's the best option... 
In my experience, in some cases `validate_script_output` failed because of unknown error (random network issue that returns !=0)  https://openqa.suse.de/tests/8189613#step/image_docker/318
Here the command fails and makes the test fail, not due to command timeout, but due to some random exit code. So, I will remove this condition.

lib/utils.pm
jlausuch: ok, I'll add that check too.

lib/utils.pm
jlausuch: ok, but in that case, the return error is not `output not validating`... so It should be ok to keep this condition in the end...

pdostal:
> I think it would be a good idea also, to consider adding it to the `testapi` as it might be a better place to ensure that further changes, won't break the retries

In that case it should be done together with `script_retry`.

pdostal (approved):

jlausuch:
VR: https://openqa.suse.de/tests/8209549#

lib/utils.pm
jlausuch: yes, these corner cases can happen. Thanks for the suggestion!

lib/utils.pm
```diff
@@ -1658,6 +1659,52 @@ sub script_output_retry {
...
+
+    $args{timeout} += 3;    # timeout for script_run must be larger than for the 'timeout ...' command
+    for (1 .. $retry) {
+        eval { $ret = validate_script_output($exec, $check, %args, procced_on_failure => 1); };
+        return $ret if (defined($ret));
+        die("Can't validate output after $retry retries") if $retry == $_;
+        record_info("Retry", "validate_script_output timed out. Retrying...");
+        sleep $delay;
+    }
+    return $ret;
```
mloviska, line 1705: I think the code will never reach this `return` as the loop handles success or fail results.
In such a case, you might move the `die` statement outside the loop, as it means that the loop exhausted all retries.

jlausuch: good point, will do so :)

jlausuch:
Latest VR: https://openqa.suse.de/tests/8213651
Luckily, this VR makes use of those retries, 
[0] retries: https://openqa.suse.de/tests/8213651#step/image_docker/736
[1] retry: https://openqa.suse.de/tests/8213651#step/image_docker/176
[2] retries: https://openqa.suse.de/tests/8213651#step/image_docker/1532

jlausuch:
> unit t

I'm sorry, but after some attempts, I really failed mocking `script_output` and make the the UT run. So, I'd rather leave it for when I have more time to invest in it.

jlausuch:
Thanks @cfconrad for helping out with UT!

cfconrad (approved):
LGTM

jlausuch:
Running VRs now...

lib/utils.pm
```diff
@@ -1658,6 +1659,53 @@ sub script_output_retry {
...
+    my $exec = "timeout $args{timeout} $cmd";
+    # Exclamation mark needs to be moved before the timeout command, if present
+    if (substr($cmd, 0, 1) eq "!") {
+        $cmd = substr($cmd, 1);
+        $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
+        $exec = "! timeout $args{timeout} $cmd";
+    }
+
+    $args{timeout} += 3;    # timeout for script_run must be larger than for the 'timeout ...' command
+    $args{procced_on_failure} = 1;
```
cfconrad, line 1699:
```suggestion
    $args{proceed_on_failure} = 1;
```
But I would let this up to the caller! SOrry I over seen this on my review!

jlausuch: You don't need to be sorry, it's my fault I type too quick :D 
Good that you spotted it, thank you!

lib/utils.pm
```diff
@@ -1663,6 +1664,54 @@ sub script_output_retry {
...
+    # Exclamation mark needs to be moved before the timeout command, if present
+    if (substr($cmd, 0, 1) eq "!") {
+        $cmd = substr($cmd, 1);
+        $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
+        $exec = "! timeout $args{timeout} $cmd";
+    }
+
+    $args{timeout} += 3;    # timeout for script_run must be larger than for the 'timeout ...' command
+    for (1 .. $retry) {
+        eval { $ret = validate_script_output($exec, $check, %args); };
```
cfconrad, line 1706:
```suggestion
        my $timeout = delete $args{timeout};
        eval { $ret = validate_script_output($exec, $check, $timeout, %args); };
```
I think this is the workaround for the failing unit-test. It's because of https://github.com/os-autoinst/os-autoinst/pull/1974 and you are passing a hash, where the order of the elements is random. 
As you set `proceed_on_failure` in that hash and this key isn't handled properly in `validate_script_output` you sometimes fail...

jlausuch:
Another failure today that could be "workarounded" by this PR: https://openqa.suse.de/tests/8242917#step/image_docker/701

VR: https://openqa.suse.de/tests/8245849

foursixnine (approved):

jlausuch:
I did ran some VRs and there is still some strange behavior only in podman tests:
[15](https://openqa.suse.de/tests/overview?distri=sle&version=15&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry)
[15-SP1](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP1&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry)
[15-SP2](https://openqa.suse.de/tests/overview?distri=sle&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry&version=15-SP2)
[15-SP3](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry)

lib/utils.pm
```diff
@@ -1663,6 +1664,55 @@ sub script_output_retry {
...
+    my $exec = "timeout $args{timeout} $cmd";
+    # Exclamation mark needs to be moved before the timeout command, if present
+    if (substr($cmd, 0, 1) eq "!") {
+        $cmd = substr($cmd, 1);
+        $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
+        $exec = "! timeout $args{timeout} $cmd";
+    }
+
+    $args{timeout} += 3;    # timeout for script_run must be larger than for the 'timeout ...' command
+    my $timeout = delete $args{timeout};
```
cfconrad, line 1705: As you pointend me to 
the output looks like:
```
# cat > /tmp/scriptIzNvi.sh << 'EOT_IzNvi'; echo IzNvi-$?-
> timeout 300 podman run --entrypoint /usr/lib/zypp/plugins/services/container-suseconnect-zypp -i registry.suse.com/suse/sles12sp3 lp
> EOT_IzNvi
IzNvi-0-
# echo IzNvi; bash -oe pipefail /tmp/scriptIzNvi.sh ; echo SCRIPT_FINISHEDIzNvi-$?-
IzNvi
cat > /tmp/scriptIzNvi.sh << 'EOT_IzNvi'; echo IzNvi-$?-
timeout 300 podman run --entrypoint /usr/lib/zypp/plugins/services/container-suseconnect-zypp -i registry.suse.com/suse/sles12sp3 lp
EOT_IzNvi
echo IzNvi; bash -oe pipefail /tmp/scriptIzNvi.sh ; echo SCRIPT_FINISHEDIzNvi-$?-
cat > /tmp/scriptIzNvi.sh << 'EOT_IzNvi'; echo IzNvi-$?-
timeout 300 podman run --entrypoint /usr/lib/zypp/plugins/services/container-suseconnect-zypp -i registry.suse.com/suse/sles12sp3 lp
EOT_IzNvi
echo IzNvi; bash -oe pipefail /tmp/scriptIzNvi.sh ; echo SCRIPT_FINISHEDIzNvi-$?-
cat > /tmp/scriptIzNvi.sh << 'EOT_IzNvi'; echo IzNvi-$?-
timeout 300 podman run --entrypoint /usr/lib/zypp/plugins/services/container-suseconnect-zypp -i registry.suse.com/suse/sles12sp3 lp
EOT_IzNvi
echo IzNvi; bash -oe pipefail /tmp/scriptIzNvi.sh ; echo SCRIPT_FINISHEDIzNvi-$?-
cat > /tmp/scriptIzNvi.sh << 'EOT_IzNvi'; echo IzNvi-$?-
timeout 300 podman run --entrypoint /usr/lib/zypp/plugins/services/container-suseconnect-zypp -i registry.suse.com/suse/sles12sp3 lp
EOT_IzNvi
echo IzNvi; bash -oe pipefail /tmp/scriptIzNvi.sh ; echo SCRIPT_FINISHEDIzNvi-$?-
```
Which for me looks like podman somehow do not stop on `TERM` signal.

My idea is to add `--kill-after=DURATION` to your command, like:
```suggestion
    my $delay = delete $args{delay};
    my $timeout = delete $args{timeout};
    my $ret;

    my $exec = "timeout --kill-after 5s ${timeout}s";
    # Exclamation mark needs to be moved before the timeout command, if present
    if (substr($cmd, 0, 1) eq "!") {
        $cmd = substr($cmd, 1);
        $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
        $exec = "! $exec";
    }
    $exec = "$exec $cmd"

    $timeout += 8;    # timeout for script_run must be larger than for the 'timeout ...' command
```

jlausuch:
Latest state works... for some reason podman output was not showing up if `timeout` was set before it, at least in openQA. I found out that the flag `--foreground` solves this problem. 
- [15](https://openqa.suse.de/tests/overview?distri=sle&version=15&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry)
- [15-SP2](https://openqa.suse.de/tests/overview?distri=sle&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry&version=15-SP2)
- [15-SP3](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=jlausuch%2Fos-autoinst-distri-opensuse%23validate_output_retry)

Many thanks to @cfconrad who helped a lot with the code.

PR #14297, opened 2022-02-21
slem5.1: Ensure booting RT kernel

mloviska:
Slem5.1 comes with 2 kernels pre-installed in case of *Default-RT-Image*
flavor. Unfortunately, grub2 to is set to boot from the first record in
grub menu which kernel-default. Removal of kernel-default cannot be done
without removing kernel-rt at the same time.

- Related ticket: [SLE Micro RT - Boot RT kernel instead of default one](https://progress.opensuse.org/issues/102723)
- Verification runs: http://kepler.suse.cz/tests/13788#step/host_config/12
  * [microos-Tumbleweed-MicroOS-Image-Kubic-kubeadm-x86_64-Build20220219-kubeadm@64bit-2G ](http://kepler.suse.cz/tests/13791)
  * [microos-Tumbleweed-MicroOS-Image-ContainerHost-x86_64-Build20220219-container-host@64bit](http://kepler.suse.cz/tests/13790)
  * [ microos-Tumbleweed-MicroOS-Image-x86_64-Build20220219-microos@64bit](http://kepler.suse.cz/tests/13789)

### DIFF
--- a/lib/microos.pm
+++ b/lib/microos.pm
@@ -11,7 +11,7 @@ use Exporter;
 use strict;
 use warnings;
 use testapi;
-use version_utils qw(is_microos is_selfinstall);
+use version_utils qw(is_microos is_selfinstall is_rt);
 use power_action_utils 'power_action';
 use Utils::Architectures qw(is_aarch64);
 
@@ -43,8 +43,7 @@ sub microos_reboot {
     # No grub bootloader on xen-pv
     # grub2 needle is unreliable (stalls during timeout) - poo#28648
     assert_screen 'grub2', 300;
-    send_key('ret') if match_has_tag('grub2');
-
+    send_key('ret') if get_var('DISABLE_GRUB_TIMEOUT');
     microos_login;
 }
 
--- a/lib/version_utils.pm
+++ b/lib/version_utils.pm
@@ -347,7 +347,7 @@ sub is_sles4sap_standard {
 Returns true if called on a real time system
 =cut
 sub is_rt {
-    return check_var('SLE_PRODUCT', 'rt');
+    return (check_var('SLE_PRODUCT', 'rt') || get_var('FLAVOR') =~ /rt/i);
 }
 
 =head2 is_hpc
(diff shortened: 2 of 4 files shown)


lib/microos.pm
```diff
@@ -43,8 +43,7 @@ sub microos_reboot {
     # No grub bootloader on xen-pv
     # grub2 needle is unreliable (stalls during timeout) - poo#28648
     assert_screen 'grub2', 300;
-    send_key('ret') if match_has_tag('grub2');
-
+    send_key('ret') if get_var('DISABLE_GRUB_TIMEOUT');
```
jlausuch, line 46: this can impact some places... this will need some VRs in several places of O3 and OSD.

mloviska:
> Looks very good... Do we need to do this for all versions or only for 5.1?
Only 5.1 as 5.2 does not come with other kernel than kernel-rt
http://kepler.suse.cz/tests/13765#step/rt_is_realtime/47

jlausuch:
> > Looks very good... Do we need to do this for all versions or only for 5.1?
> 
> Only 5.1 as 5.2 does not come with other kernel than kernel-rt http://kepler.suse.cz/tests/13765#step/rt_is_realtime/47

but then, we need some sort of condition in this line, don't we?
`change_grub_config('=0', '="1>2"', 'GRUB_DEFAULT') if is_rt;`

mloviska:
> > > Looks very good... Do we need to do this for all versions or only for 5.1?
> > 
> > 
> > Only 5.1 as 5.2 does not come with other kernel than kernel-rt http://kepler.suse.cz/tests/13765#step/rt_is_realtime/47
> 
> but then, we need some sort of condition in this line, don't we? `change_grub_config('=0', '="1>2"', 'GRUB_DEFAULT') if is_rt;`

Yup, it was there, I have just removed it by accident while playing with the code. :man_facepalming:

jlausuch (approved):

mloviska:
> https://openqa.opensuse.org/tests/2201361#investigation
> 
> This seems to be a fallout from this change

Thanks for the report, checking.

jlausuch:
This is because those jobs don't have `DISABLE_GRUB_TIMEOUT` set, so I would go back to `send_key('ret') if match_has_tag('grub2');`

PR #14298, opened 2022-02-21
Fix textmode key shortcuts in product selection

dzedro:
I could not find any textmode sles4sap tests on older versions than 15-SP4

- Fail: https://openqa.suse.de/tests/8195354#step/accept_license/2 https://openqa.suse.de/tests/8195293#step/accept_license/1
- Verification run:
https://openqa.suse.de/tests/8202819#step/welcome/6 x86_64
https://openqa.suse.de/tests/8202821#step/welcome/6 ppc64le

### DIFF
--- a/tests/installation/welcome.pm
+++ b/tests/installation/welcome.pm
@@ -65,8 +65,8 @@ sub get_product_shortcuts {
             : is_aarch64() ? 's'
             : 'i',
             sled => 'x',
-            sles4sap => is_ppc64le() ? 'i'
-            : (is_sle('15-SP2+') && is_x86_64() && !is_quarterly_iso()) ? 't'
+            sles4sap => is_ppc64le() ? 'u'
+            : (is_sle('15-SP2+') && is_x86_64() && !is_quarterly_iso()) ? 'i'
             : 'p',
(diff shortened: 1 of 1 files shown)


ricardobranco777:
The problem with this PR is that we'd have it to revert it later when they add SUMA to the product list, like we did with: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13861

See the git history for tests/installation/welcome.pm

ricardobranco777:
> @ricardobranco777 I did see the back and forth with this shortcuts, but the openQA links are deleted. Where are the SUMA tests ? I will just add SUMA or whatever is different ...

We don't have the SUMA tests, just that when SUMA is offered in the product list, we'd have to revert.

dzedro:
Ah ok I get it now. When there is SUMA, I would probably split needle `select-product-sles4sap` to with and without SUMA, with corresponding product_shortcuts

ricardobranco777:
> Ah ok I get it now. When there is SUMA, I would probably split needle `select-product-sles4sap` to with and without SUMA, with corresponding product_shortcuts

Let's try that approach then.

LGTM.

PR #14305, opened 2022-02-21
Fix Podman & Docker merge in public cloud main

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/107179
- Verification run: http://openqa.suse.de/tests/8205048

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -18,15 +18,23 @@ our @EXPORT = qw(
   load_publiccloud_tests
 );
 
+sub load_3rd_party_image_test {
+    my ($runtime) = @_;
+    my $args = OpenQA::Test::RunArgs->new();
+    $args->{runtime} = $runtime;
+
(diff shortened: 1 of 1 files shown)


jlausuch:
Maybe we should reuse `main_containers.pm` instead of duplicating code in `main_publiccloud.pm`. 
But maybe something to think for another PR, let's fix this first.

jlausuch (approved):

lib/main_publiccloud.pm
```diff
@@ -18,15 +18,23 @@ our @EXPORT = qw(
   load_publiccloud_tests
 );
 
+sub load_3rd_party_image_test {
```
pdostal, line 21: Well why we just don't export that subroutine from `main_containers`? :disappointed:

pdostal:
> Maybe we should reuse main_containers.pm instead of duplicating code in main_publiccloud.pm.
Something to think for another PR, let's fix this first.
I will do that.

PR #14307, opened 2022-02-22
Add specific python installation into openvswitch_ssl

jouyingbin:
Add specific python installation into openvswitch_ssl

python-base and libpython needs to be installed specifically since
dependency is missing as python3 by default

- Related ticket: https://progress.opensuse.org/issues/107134
- Needles: NA
- Verification run: https://openqa.suse.de/t8206911
  The fail would be a new bug: https://openqa.suse.de/tests/8206911#step/openvswitch_ssl/43

### DIFF
--- a/tests/console/openvswitch_ssl.pm
+++ b/tests/console/openvswitch_ssl.pm
@@ -1,13 +1,13 @@
 # SUSE's openQA tests
 #
-# Copyright 2017-2020 SUSE LLC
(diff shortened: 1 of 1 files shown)


lilyeyes (commented):
LGTM

lilyeyes (approved):

Amrysliu (approved):

rfan1 (approved):
LGTM

PR #14308, opened 2022-02-22
PC CLI AWS Wait until the instance is running

pdostal:
- Related ticket: [poo#107176](https://progress.opensuse.org/issues/107176)
- Sporadic failure: [OSD](https://openqa.suse.de/tests/8202287#step/aws_cli/127)
- Verification run: [pdostal-server](http://pdostal-server.suse.cz/tests/13377)

### DIFF
--- a/tests/publiccloud/aws_cli.pm
+++ b/tests/publiccloud/aws_cli.pm
@@ -49,6 +49,9 @@ sub run {
     assert_script_run("aws ec2 describe-instances --filters 'Name=tag:openqa-cli-test-tag,Values=$job_id'", 90);
     my $instance_id = script_output("aws ec2 describe-instances --filters 'Name=tag:openqa-cli-test-tag,Values=$job_id' --output=text --query 'Reservations[*].Instances[*].InstanceId'", 90);
 
+    # Wait until the instance is really running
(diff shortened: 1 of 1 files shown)


tests/publiccloud/aws_cli.pm
```diff
@@ -49,6 +49,9 @@ sub run {
     assert_script_run("aws ec2 describe-instances --filters 'Name=tag:openqa-cli-test-tag,Values=$job_id'", 90);
     my $instance_id = script_output("aws ec2 describe-instances --filters 'Name=tag:openqa-cli-test-tag,Values=$job_id' --output=text --query 'Reservations[*].Instances[*].InstanceId'", 90);
 
+    # Wait until the instance is really running
+    script_retry("aws ec2 describe-instances --instance-ids $instance_id --query 'Reservations[*].Instances[*].State.Name' --output text | grep 'running'", 90, delay => 15, retry => 12);
```
asmorodskyi, line 53: isn't that something what we usually soling with `wait_for_ssh` ?

pdostal: I find this more elegant actually :) And the `wait_for_ssh` is a bit more complex than I want.

pdostal: Also the answer to your question is no as the problem is not while waiting for SSH but for waiting for AWS to assign public IP to the instance. Normally this is done by our Terraform but here I test the CLI tools where it isn't automated.

PR #14309, opened 2022-02-22
Export the main_containers::load_3rd_party_image_test() and use it in main_publiccloud

pdostal:
- Related ticket: [poo#107179](https://progress.opensuse.org/issues/107179)
- Hotfix: #14305
- Verification run: [pdostal-server](http://pdostal-server.suse.cz/tests/13378)

I would like to wait with this until we have some successful runs with the hotflix applied.

### DIFF
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -13,18 +13,11 @@ use version_utils;
 use main_common qw(loadtest);
 use testapi qw(check_var get_var);
 use Utils::Architectures qw(is_aarch64);
+use main_containers qw(load_3rd_party_image_test);
 
 our @EXPORT = qw(
   load_publiccloud_tests
 );
-
-sub load_3rd_party_image_test {
-    my ($runtime) = @_;
-    my $args = OpenQA::Test::RunArgs->new();
-    $args->{runtime} = $runtime;
-
-    loadtest('containers/third_party_images', run_args => $args, name => $runtime . "_3rd_party_images");
-}
 
 sub load_podman_tests() {
     loadtest 'containers/podman';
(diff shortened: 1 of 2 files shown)


pdostal:
@ilausuch hello Ivan, thank you very much for the hotfix. Please don't take this PR bad. I just wanted to suggest another solution.

jlausuch (approved):

lib/main_publiccloud.pm
```diff
@@ -13,19 +13,12 @@ use version_utils;
...
 use Utils::Architectures qw(is_aarch64);
+use main_containers qw(load_3rd_party_image_test);
 
 our @EXPORT = qw(
   load_publiccloud_tests
 );
 
-sub load_3rd_party_image_test {
-    my ($runtime) = @_;
-    my $args = OpenQA::Test::RunArgs->new();
```
grisu48, line 23: Why do we generate a new RunArgs instance here? It might work (for now) but this is wrong! Every test met get the same instance of RunArgs, otherwise it's purpose is defied.

I know this is not this PR, but AFAICS this is wrong.

grisu48 (changes requested):
There is something fishy.

lib/main_publiccloud.pm
jlausuch: Yeah, I was probably too rushed merging this... sorry. Wanted to get a quick fix and made it worse :) 
Thanks Pavel for taking care.

pdostal: I agree. I created [poo#107230](https://progress.opensuse.org/issues/107230)

ilausuch (approved):

grisu48 (approved):
https://progress.opensuse.org/issues/107230 covers my objection, and this one is used as a hotfix for an ongoing issue.

Going to approve now.

PR #14312, opened 2022-02-22
Switch deprecated config item for haproxy

lpalovsky:
Haproxy rest fails because of deprecated config item 'reqadd'. This 
option is replaced by 'http-request add-header'. Option 'reqadd' is 
not recommended to use since haproxy 1.5.

- Verification run: https://mordor.suse.cz/tests/2854#dependencies


ricardobranco777 (approved):
LGTM. Thanks!

bs-suse:
LGTM

PR #14314, opened 2022-02-22
Prevent change of Apparmor selection when is presented in Patterns

b10n1k:
the Installer of 15SP4 requires Apparmor pattern activation by default.
Remove of the unnecessary variable /LINUX_SECURITY_MODULE/. and enable /set_linux_security_to_none/ only when it is missing from the patterns.
In such way, the /set_linux_security_to_none/ will not run when the Apparmor is already selected and there wouldnt be a problem with the installation.

- Verification run: http://aquarius.suse.cz/tests/7933#

### DIFF
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -101,7 +101,8 @@ sub run {
                 assert_screen 'inst-xen-pattern';
             }
         }
-        set_linux_security_to_none if (check_var('LINUX_SECURITY_MODULE', 'none') && is_sle('>=15-SP4'));
+     # the Installer of 15SP4 requires Apparmor pattern activation by default. If Apparmor not presented in PATTERNS select None for Major Linux Security Module
+        set_linux_security_to_none if (is_sle('>=15-SP4') && !(get_var('PATTERNS') =~ 'default|all|apparmor'));
         ensure_ssh_unblocked;
         $self->check_default_target();
     }


czerw:
Which failure are you trying to fix?

czerw:
Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.

b10n1k:
> Which failure are you trying to fix?

The problem is the following:
- i added apparmor in the PATTERNS. this should not deselect this pattern from the installation and the problem was resolved.
- then the https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14301 merged and it doesnt take in account the previous setup. 

The `set_linux_security_to_none` should run at least only when the installation fails which is when the apparmor is not in PATTERNS, which means that it is going to show a conflict

b10n1k:
> Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.

i added the apparmor. i had no clear requirements if it must excluded or not. But in any case the condition needs to get updated and take in account the PATTERNS

czerw:
> > Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.
> 
> i added the apparmor. i had no clear requirements if it must excluded or not. But in any case the condition needs to get updated and take in account the PATTERNS

Please do verification with `base,minimal`  only.

b10n1k:
> > > Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.
> > 
> > 
> > i added the apparmor. i had no clear requirements if it must excluded or not. But in any case the condition needs to get updated and take in account the PATTERNS
> 
> Please do also verification with `base,minimal` only.

> > Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.
> 
> 
> i added the apparmor. i had no clear requirements if it must excluded or not. But in any case the condition needs to get updated and take in account the PATTERNS

Please do also verification with `base,minimal` only.

> > > Verification was done with `PATTERN=base,minimal,apparmor` but we are interesed in `base,minimal` only.
> > 
> > 
> > i added the apparmor. i had no clear requirements if it must excluded or not. But in any case the condition needs to get updated and take in account the PATTERNS
> 
> Please do also verification with `base,minimal` only.

http://aquarius.suse.cz/tests/7936

czerw (approved):
LGTM, but needs another from @OleksandrOrlov 

Thanks for verification with `base, minimal` only.

OleksandrOrlov:
@b10n1k With this solution we should also be aware, that when `PATTERNS=all` and `PATTERNS=default`, apparmor will be selected as well. Please update the code and provide the VRs for the cases.

b10n1k:
> @b10n1k With this solution we should also be aware, that when `PATTERNS=all` and `PATTERNS=default`, apparmor will be selected as well. Please update the code and provide the VRs for the cases.

http://aquarius.suse.cz/tests/7939 *changes ready for review*

OleksandrOrlov:
Hey @b10n1k, I've reverted the PR, because it failed some jobs in Job Groups that we didn't consider with you. In case we want to apply the solution we should carefully check all the possible test suites, e.g. 
https://openqa.suse.de/tests/8213345
https://openqa.suse.de/tests/8213413
https://openqa.suse.de/tests/8213417 
and so on.

So, please don't remove `LINUX_SECURITY_MODULE` from Test Suites settings as it is still needed.

PR #14316, opened 2022-02-22
Build a kernel on sle15sp4

b10n1k:
This PR corrects steps required to build a kernel on sle15sp4.

The keys for the signing are not available, so there are two solutions.
One is to recreate them and set the proper value on /CONFIG_MODULE_SIG_KEY/ or disable the feature.
I think for this test case it is enough to go with the later.

I dont know how this used to work before but we need to copy the kernel to the boot and update the bootloader before the reboot.

### DIFF
--- a/tests/kernel/build_git_kernel.pm
+++ b/tests/kernel/build_git_kernel.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2019-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: git-core ncurses-devel gcc flex bison libelf-devel libopenssl-devel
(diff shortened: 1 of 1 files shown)


frankenmichl:
Have you also tested this with 15sp3?

b10n1k:
> Have you also tested this with 15sp3?

No. and i dont expect it working on 15sp3. Mainly because [this line](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14316/files#diff-cbc483e42f4025decd2a59a1acffe479264010ada7c2e2c6b12844b1faf26ef7R38). Can you point me to a 15sp3 test which uses this module?

frankenmichl:
I don't think we have one right now. I just wanted to ask if we should make this conditional, or if we just should not bother.

b10n1k:
> I don't think we have one right now. I just wanted to ask if we should make this conditional, or if we just should not bother.

I had considered that but i ignored that case as i didnt find usage. We can update the module once we find it needs to

frankenmichl (approved):
LGTM

PR #14327, opened 2022-02-22
Use script_retry for SUSEConnect -r command

jlausuch:
- Verification runs: 
https://openqa.suse.de/tests/8213565#step/suseconnect_scc/8
https://openqa.suse.de/tests/8213689#step/suseconnect_scc/8
https://openqa.suse.de/tests/8213690#step/suseconnect_scc/5
https://openqa.suse.de/tests/8213691#step/suseconnect_scc/17

### DIFF
--- a/tests/console/suseconnect_scc.pm
+++ b/tests/console/suseconnect_scc.pm
@@ -9,7 +9,7 @@
 
 use Mojo::Base qw(consoletest);
 use testapi;
-use utils qw(zypper_call);
+use utils qw(zypper_call script_retry);
 use version_utils qw(is_sle is_jeos is_sle_micro);
 use registration qw(register_addons_cmd verify_scc investigate_log_empty_license);
 
@@ -31,7 +31,7 @@ sub run {
     if ((is_jeos || is_sle_micro) && script_run(q(SUSEConnect --status-text | grep -i 'not registered'))) {
         die 'System has been already registered!';
     }
-    assert_script_run $cmd;
+    script_retry("$cmd", expect => 0, retry => 5, delay => 60, timeout => 300);
     # Check available extenstions (only present in sle)
     assert_script_run q[SUSEConnect --list-extensions];
     if (is_sle) {
--- a/tests/publiccloud/img_proof.pm
+++ b/tests/publiccloud/img_proof.pm
@@ -123,6 +123,9 @@ sub run {
         results_dir => 'img_proof_results'
     );
 
+    # IP address of instance can change during img_proof (because of hard-reboot)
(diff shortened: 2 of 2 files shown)


tests/console/suseconnect_scc.pm
```diff
@@ -31,7 +31,7 @@ sub run {
     if ((is_jeos || is_sle_micro) && script_run(q(SUSEConnect --status-text | grep -i 'not registered'))) {
         die 'System has been already registered!';
     }
-    assert_script_run $cmd;
+    script_retry("$cmd", expect => 0, retry => 5, delay => 60, timeout => 300);
```
jknphy, line 34: Good idea to improve that, some thoughts:

original timeout by default in `assert_script_run` was 90.
`expect` is not needed, it is default but you can be explicit there if you want.

Regarding the worst scenario, so 25 minutes of timeout + 5 minutes of delays, 30 minutes. It is still in range to get some benefit depending on the job that is restarted I guess, some jobs takes hours. But checking your VRs, I would suggest to be a bit more conservative, for example keep it `timeout => 90` and increase from there if we need it, in your VR you schedule the module at the beginning, plus most of your VR are around 30', so 90*5, 7.5 minutes...maybe that is too short to give a benefit unless you have other evidences where might be needed, so ok, perhaps timeout=150. But of course there is a benefit of not doing manually, just referring to the worst case where more job would be queued retrying. wdyt?

jknphy: on the other hand we schedule this module at the end (after one hour) :) so it will be a time-saver: https://openqa.suse.de/tests/8206576#step/suseconnect_scc/1

jknphy (approved):

tests/console/suseconnect_scc.pm
jlausuch: That's very good input, thank you :) 
Ok, I will remove the expect. 

So, I raised to 300 just to say one number above 90, where I found some timeouts, e.g. https://openqa.suse.de/tests/8193554#step/suseconnect_scc/8

So, since 300 is probably too much, I would suggest (2 * 90), which would be ~12min in the worst scenario, wdyt?

jknphy: sounds good to me.

jlausuch:
https://openqa.suse.de/tests/8216764#step/suseconnect_scc/8
https://openqa.suse.de/tests/8216765#step/suseconnect_scc/8
https://openqa.suse.de/tests/8216766#step/suseconnect_scc/5
https://openqa.suse.de/tests/8216767#step/suseconnect_scc/17

jlausuch:
Followu-up: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14402

PR #14330, opened 2022-02-23
PC: storage_perf: Improve analysis

pdostal:
Our storage performance test detects if the mean value of last and previous tests differs by more than 10%.
 * The problem is that the output in case of error was too quiet: `Anomaly occurred` so I improved the output 
 * Other problem is that it fails with first anomaly so the measurement didn't finish so now the module dies at the end.

- Related ticket: [poo#107104](https://progress.opensuse.org/issues/107104)
- Verification run: [GCE](http://pdostal-server.suse.cz/tests/13437), [GCE-BYOS](http://pdostal-server.suse.cz/tests/13436)

### DIFF
--- a/tests/publiccloud/storage_perf.pm
+++ b/tests/publiccloud/storage_perf.pm
@@ -89,6 +89,7 @@ sub db_has_data {
 =cut
 sub analyze_previous_series {
     my ($args, $load_types) = @_;
+    my $result = 0;
 
     foreach my $load_type (@$load_types) {
         $args->{load_type} = $load_type;
@@ -97,21 +98,29 @@ sub analyze_previous_series {
         my $last_records_mean = get_mean_from_db($args, " limit 5");
         # getting mean for 5 runs which were before last 5 ( from 6 to 10th)
         my $previous_records_mean = get_mean_from_db($args, " limit 5 offset 5");
+
+        my $generic_message = sprintf("See previous message and http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor});
+
         # we do analysis only if we get back some non-zero values
         if (defined($previous_records_mean) && defined($last_records_mean) && $previous_records_mean != 0 && $last_records_mean != 0) {
-            record_info('ANALYZE', "Checking " . $current_load_type);
-            # this formula detect if mean values differs more than 10%
-            if ((abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100 > 10) {
-                die sprintf("Anomaly occurred in http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor});
+            my $diff_percents = (abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100;
+            my $analyze = sprintf("Analyzing: %s\n", $current_load_type);
+            $analyze .= "Comparing mean values of previous 5 entries and last 5 entries.\n";
+            $analyze .= sprintf("Mean of previous 5 entries: %.2f Mean of last 5 entries: %.2f\n", $previous_records_mean, $last_records_mean);
+            $analyze .= sprintf("The difference is: %.2f%%.", $diff_percents);
+            record_info('ANALYZE', $analyze);
+            # This detects if mean values differs more than 10%
+            if ($diff_percents > 10) {
+                record_soft_failure("Deviation occurred. $generic_message");
+                $result = 1;
+            } else {
+                record_info('PASS', "The data looks good. $generic_message");
             }
-            else {
-                record_info('PASS', sprintf("For %s. \n previous mean: %d \n last mean: %d", $current_load_type, $previous_records_mean, $last_records_mean));
-            }
-        }
-        else {
-            record_info('N/A', sprintf("Got previous mean - %d and last mean - %d so anaysis not possible for %s", $previous_records_mean, $last_records_mean, $current_load_type));
+        } else {
+            record_info('N/A', "Analysis not possible $generic_message");
         }
     }
+    return $result;
 }
 
 
@@ -232,7 +241,8 @@ sub run {
             if (db_has_data(%influx_read_args)) {
                 # we will do anaysis for same load types which we just pushed to db
                 my @load_types = keys %$values;
-                analyze_previous_series(\%influx_read_args, \@load_types);
+                # Change the test module result to 'fail' if deviation in analyze_previous_series() occures
+                $self->{result} = 'fail' if analyze_previous_series(\%influx_read_args, \@load_types) == 1;
             }
             else {
                 record_info('NO DATA', "We need at least 10 test results to analyze " . $href->{name} . "\n");


tests/publiccloud/storage_perf.pm
```diff
@@ -14,6 +14,7 @@ use utils;
 use db_utils;
 use Mojo::JSON;
 use publiccloud::utils qw(is_byos registercloudguest);
+use Data::Dumper;
```
ilausuch, line 17: I guess this is only for development purposes

asmorodskyi, line 17: I think this was for debugging purposes :)

tests/publiccloud/storage_perf.pm
```diff
@@ -99,19 +100,20 @@ sub analyze_previous_series {
         my $previous_records_mean = get_mean_from_db($args, " limit 5 offset 5");
         # we do analysis only if we get back some non-zero values
         if (defined($previous_records_mean) && defined($last_records_mean) && $previous_records_mean != 0 && $last_records_mean != 0) {
-            record_info('ANALYZE', "Checking " . $current_load_type);
-            # this formula detect if mean values differs more than 10%
-            if ((abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100 > 10) {
-                die sprintf("Anomaly occurred in http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor});
+            my $diff_percents = (abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100;
```
ilausuch, line 103: I like more https://en.wikipedia.org/wiki/Moving_average

ilausuch: but it's ok

tests/publiccloud/storage_perf.pm
```diff
@@ -89,6 +89,7 @@ sub db_has_data {
 =cut
 sub analyze_previous_series {
     my ($args, $load_types) = @_;
+    my $return = 0;
```
asmorodskyi, line 92:
```suggestion
    my $rezult = 0;
```

`return $return;` - looks a bit odd :)

ilausuch (approved):

tests/publiccloud/storage_perf.pm
asmorodskyi: this discussion anyway goes a bit beyond what Pavel is trying to achieve here . @ilausuch feel free to create a ticket if you still thinks that it make sense to do any changes in this direction.

tests/publiccloud/storage_perf.pm
```diff
@@ -99,19 +100,20 @@ sub analyze_previous_series {
         my $previous_records_mean = get_mean_from_db($args, " limit 5 offset 5");
         # we do analysis only if we get back some non-zero values
         if (defined($previous_records_mean) && defined($last_records_mean) && $previous_records_mean != 0 && $last_records_mean != 0) {
-            record_info('ANALYZE', "Checking " . $current_load_type);
-            # this formula detect if mean values differs more than 10%
-            if ((abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100 > 10) {
-                die sprintf("Anomaly occurred in http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor});
+            my $diff_percents = (abs($previous_records_mean - $last_records_mean) / $previous_records_mean) * 100;
+            record_info('ANALYZE', sprintf("Checking %s:\nFormula: (abs(\$previous_records_mean - \$last_records_mean) / \$previous_records_mean) * 100 = (abs(%s - %s) / %s) * 100 = %s%", $current_load_type, $previous_records_mean, $last_records_mean, $previous_records_mean, $diff_percents));
```
grisu48: IMHO this is way too verbose. We don't need to print out the whole formula. The resulting values and the important values so that you could calculate it yourself should be enough:
```suggestion
            record_info('ANALYZE', "Comparing mean value of 5 datasets\Current dataset: $last_records_mean\nPrevious dataset: $previous_records_mean\nDiff: $diff_percents %");
```

This results in a infobox like 

    Comparing mean value of 5 datasets
    Current dataset: 5
    Previous dataset: 4
    Diff: 20%

tests/publiccloud/storage_perf.pm
```diff
@@ -99,19 +100,20 @@ sub analyze_previous_series {
...
+            # This detects if mean values differs more than 10%
+            if ($diff_percents > 10) {
+                record_soft_failure(sprintf("Anomaly occurred in http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor}));
```
grisu48, line 107: One important information is missing in all of those output: What value is causing the deviation? Is it the load? CPU time? Disk storage? Number of bananas on the system?

Also: Can we please not call this anomaly but a deviation please? :slightly_smiling_face:

tests/publiccloud/storage_perf.pm
```diff
@@ -99,19 +100,20 @@ sub analyze_previous_series {
...
+                $return = 1;
+            } else {
+                record_info('PASS', sprintf("Anomaly occurred in http://openqa-perf.qa.suse.de/d/hoRc37HWz/storage-performance?orgId=1&var-os_flavor=%s&var-os_version=%s", $args->{os_flavor}, $args->{os_flavor}));
```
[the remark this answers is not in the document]
grisu48: I would like to rephrase this. The terms "previous mean" and "last mean" are not understandable unless you dig into the code.
```suggestion
                record_info('PASS', sprintf("For %s\nComparing average of dataset over 5 points\nReference dataset: %d \nCurrent dataset: %d", $current_load_type, $previous_records_mean, $last_records_mean));
```

WDYT? The aim is to make this understandable for people without the need to dive into the code.

grisu48 (commented):
Some comments

tests/publiccloud/storage_perf.pm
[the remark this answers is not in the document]
pdostal: Is it 'dataset'? What about 'record'?

tests/publiccloud/storage_perf.pm
pdostal: Why '5' datasets? But yes, I vote for leaving the `Diff` there.

tests/publiccloud/storage_perf.pm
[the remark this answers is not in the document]
asmorodskyi: generally I agree with rephrasing but I would keep URL to grafana ...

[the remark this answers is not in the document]
asmorodskyi: "Comparing average of dataset over 5 points" - well without actually knowing what is going on just by this sentence I would not be able to understand something , but because my version is NOT better I am fine to keep yours :) 

"Reference dataset/Current dataset" - is more confusing and this part needs to be changed . It is not "dataset" per ce it is mean of it . So I think "Reference mean" is more accurate

tests/publiccloud/storage_perf.pm
asmorodskyi: > One important information is missing in all of those output: What value is causing the deviation? Is it the load? CPU time? Disk storage? Number of bananas on the system?

Well I would expect from a person looking into **storage_perf* test to be aware that he/she reviewing results of storage performance . As Pavel correctly mention just one record_info before this one we giving some hints . I don't think it is necessary to repeat that again . 

> Also: Can we please not call this anomaly but a deviation please? 

:+1:  yeah your deviation is certainly better than anomaly .

asmorodskyi (approved):
LGTM

grisu48 (approved):

PR #14333, opened 2022-02-23
Fix Major Linux Security Module conflicts

b10n1k:
With the previous commit[0] some tests started failing because even if the PATTERNS do not include apparmor the module is already selected and the flow doesnt effect the installation overview window.
Add a needle check should avoid to go into `set_linux_security_to_none` in such cases.

[0] https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14314

- Needles:  installation_overview-apparmor-unselected-20220223.json
- Verification run: 

- [textmode_ha](https://openqa.suse.de/tests/8218192)
- [offline_slehpc15sp3_pscc_basesys-desk-dev-hpc-python2-srv-wsm_def_full_ms](https://openqa.suse.de/tests/8218193)
- [with apparmor selected](http://aquarius.suse.cz/tests/7944)
- [with apparmor not selected](http://aquarius.suse.cz/tests/7943)

### DIFF
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -101,7 +101,9 @@ sub run {
                 assert_screen 'inst-xen-pattern';
             }
         }
-        set_linux_security_to_none if (check_var('LINUX_SECURITY_MODULE', 'none') && is_sle('>=15-SP4'));
+        # the Installer of 15SP4 requires Apparmor pattern activation by default. If Apparmor not presented in PATTERNS and needle matches
+        # select None for Major Linux Security Module.
+        set_linux_security_to_none if (is_sle('>=15-SP4') && check_screen("apparmor-not-selected") && !(get_var('PATTERNS') =~ 'default|all|apparmor'));
         ensure_ssh_unblocked;
         $self->check_default_target();
     }


lpalovsky:
Executed previously failed HA tests:
https://openqa.suse.de/tests/8221924#
https://openqa.suse.de/tests/8218311#

Both tests PASSED beyond problematic module.

From HA perspective LGTM. 
Thanks.

OleksandrOrlov:
@b10n1k Thank you a lot for your input and for your attempt to make the condition to be logical. But I'm not sure that we should apply this solution as I don't see any advantages in it comparing to #14301.

But I see disadvantages: now you allowed the code to be so smart, that depending on a needle ( `check_screen("apparmor-not-selected")` ), it makes a decision to change the test flow. It was already historically proven that it makes tests fragile and unpredictable.

I think that the solution in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14301 is more straightforward and allows to control where "None" will be selected or not. So, I suggest to keep it.

b10n1k:
> @b10n1k Thank you a lot for your input and for your attempt to make the condition to be logical. But I'm not sure that we should apply this solution as I don't see any advantages in it comparing to #14301.
> 
> But I see disadvantages: now you allowed the code to be so smart, that depending on a needle ( `check_screen("apparmor-not-selected")` ), it makes a decision to change the test flow. It was already historically proven that it makes tests fragile and unpredictable.
> 
> I think that the solution in #14301 is more straightforward and allows to control where "None" will be selected or not. So, I suggest to keep it.

But the solution on #14301 has the disadvantage of variable dependency for a case that it shouldnt. Having `apparmor` in PATTERNS and the variable cause unexpected behavior. Between a test variable and a needle i think the second is preferable, because mitigates the surprises. It is also more complex to define where the test variable should apply in the first glance. So i expect arbitrary use of it and it is really bad idea. The condition from the other side is quite straight forward without those bad surprises.

OleksandrOrlov:
> But the solution on #14301 has the disadvantage of variable dependency for a case that it shouldnt. Having `apparmor` in PATTERNS and the variable is very likely to cause unexpected behavior. 

It shouldn't and it is valid steps that a user can do. A user can select `apparmor` pattern and then select "None" in Linux Security. 

Test module should not be so smart to decide what to select depending on a needle. It causes a lot of unclear issues and confuses on review. 

Also, this `installation_overview` module reaches the end of its life. In YaST job group it is already changed to more proper module and it will be applied in the close future to other job groups, so any solution which will be used in `installation_overview` is temporary and we shouldn't spend a lot of time on it, especially because it is already solved and proved that it is working.

b10n1k:
> > But the solution on #14301 has the disadvantage of variable dependency for a case that it shouldnt. Having `apparmor` in PATTERNS and the variable is very likely to cause unexpected behavior.
> 
> It shouldn't and it is valid steps that a user can do. A user can select `apparmor` pattern and then select "None" in Linux Security.
> 
User yes but we have an existing variable for the selection of the modules which overlaps with the new one.
 
> Test module should not be so smart to decide what to select depending on a needle. It causes a lot of unclear issues and confuses on review.
> 
i dont know what are you talking about here and i cant recall those /unclear issues/ and confusions on review. 

> Also, this `installation_overview` module reaches the end of its life. In YaST job group it is already changed to more proper module and it will be applied in the close future to other job groups, so any solution which will be used in `installation_overview` is temporary and we shouldn't spend a lot of time on it, especially because it is already solved and proved that it is working.

This sounds like an assumption to me as i dont know any plan to adopt the yast way in any near future

czerw:
Ok for me, but main approval needs to come from yast team.

PR #14344, opened 2022-02-24
Comment upload_log function due to file is too big

XGWang0:
Not use upload_log to upload file to openqa

- Related ticket: N/A
- Needles: N/A
- Verification run: http://openqa.qa2.suse.asia/tests/45706

### DIFF
--- a/tests/cpu_bugs/mitigation_perf.pm
+++ b/tests/cpu_bugs/mitigation_perf.pm
@@ -174,7 +174,7 @@ sub run {
         }
         sleep(30);
     }
-    upload_logs("/root/$logfile");
+    #upload_logs("/root/$logfile");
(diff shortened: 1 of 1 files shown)


alice-suse:
@XGWang0  You are actually doing comment, rather than uncomment. Please correct commit msg. Besides, skip uploading necessary log just because it is too big, is it a good way? Of course up to your test needs to decide, I just give a hint.

XGWang0:
> @XGWang0 You are actually doing comment, rather than uncomment. Please correct commit msg. Besides, skip uploading necessary log just because it is too big, is it a good way? Of course up to your test needs to decide, I just give a hint.

@alice-suse  This is temporary solution, i will consider better way to enhance it. thanks for your reminder.

museQi:
LGTM

PR #14345, opened 2022-02-24
x11/multi_users_dm: Exit xterm before logout

Vogtinator:
Otherwise it's visible behind the logout overlay and needles don't match.

e.g. https://openqa.opensuse.org/tests/2207205#step/multi_users_dm/32

Verification run: http://10.160.67.86/tests/1176


asdil12 (approved):
LGTM

PR #14346, opened 2022-02-24
x11/vlc: Avoid unnecessary delays

Vogtinator:
send_key_until_needlematch checks first before sending the key, which adds a
delay of the specified timeout. Avoid that by explicitly sending the key
before calling it.

Also skip `wait_still_screen` while a video is playing.

- Verification run: http://10.160.67.86/tests/1180

### DIFF
--- a/tests/x11/vlc.pm
+++ b/tests/x11/vlc.pm
@@ -15,7 +15,9 @@ sub run {
     x11_start_program('vlc --no-autoscale', target_match => 'vlc-first-time-wizard');
     assert_and_click "vlc-first-time-wizard";
     assert_screen "vlc-main-window";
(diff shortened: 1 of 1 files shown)


okurz:
> send_key_until_needlematch checks first before sending the key, which adds a delay of the specified timeout.

That's bad. Can you propose a PR to change that? That would be awesome!

okurz (approved):

Vogtinator:
> > send_key_until_needlematch checks first before sending the key, which adds a delay of the specified timeout.
> 
> That's bad. Can you propose a PR to change that? That would be awesome!

I assumed that it is intentional, and changing that would break tests. It would for instance break vlc after this PR is merged...

PR #14355, opened 2022-02-25
Add debugging information for zypper ref

grisu48:
Adds debugging information as requested in bsc#1196462:

* Timestamp when adding a new product
* Print /etc/hosts so that the smt server IP is logged

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1196462
- Verification run: http://duck-norris.qam.suse.de/t8247 http://duck-norris.qam.suse.de/t8248 (passing relevant steps)

### DIFF
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -50,11 +50,20 @@ sub select_host_console {
     set_var('TUNNELED', $tunneled) if $tunneled;
 }
 
+# Get the current UTC timestamp as YYYY/mm/dd HH:MM:SS
+sub utc_timestamp {
+    my @weekday = ("Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat");
+    my ($sec, $min, $hour, $day, $mon, $year, $wday, $yday, $isdst) = gmtime(time);
+    $year = $year + 1900;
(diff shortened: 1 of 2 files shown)


lib/publiccloud/utils.pm
```diff
@@ -54,7 +54,8 @@ sub register_addon {
     my ($remote, $addon) = @_;
     my $arch = get_var('PUBLIC_CLOUD_ARCH') // "x86_64";
     $arch = "aarch64" if ($arch eq "arm64");
-    record_info($addon, "Going to register '$addon' addon");
+    my $timestamp = script_output("date -u +%d/%m/%Y%t%H:%M:%S");
```
pdostal, line 57:
```perl
use DateTime;
```
```suggestion
    my $dt = DateTime->now;
    my $timestamp = sprintf("%s. %s. %s:%s", $dt->day, $dt->month, $dt->hour, $dt->minute);
```

pdostal (approved):
I have one improvement :smile:

lib/publiccloud/utils.pm
grisu48: That means would be taking the local time of the worker, which is not guaranteed to be UTC. I don't like this.

grisu48: I changed it to use `gmtime`.

PR #14357, opened 2022-02-25
console/gpg: Use RSA keys explicitly and fix test for success

Vogtinator:
The test is meant to check with different key lengths, but those are only
used with non-ECC keys. So use RSA explicitly. Ideally there'll be a separate test for ECC keys in the future.

With ECC as default, key generation fails anyway because of
https://dev.gnupg.org/T5444.

Additionally, fix the `wait_serial` which failed with recent versions because the message isn't always printed.

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1193133
- Verification run: http://10.160.67.86/tests/1188

### DIFF
--- a/tests/console/gpg.pm
+++ b/tests/console/gpg.pm
@@ -45,9 +45,9 @@ sub gpg_test {
         # Preparing a config file for gpg --batch option
         assert_script_run(
             "echo \"\$(cat <<EOF
-Key-Type: default
+Key-Type: RSA
 Key-Length: $key_size
-Subkey-Type: default
+Subkey-Type: RSA
 Subkey-Length: $key_size
 Name-Real: $username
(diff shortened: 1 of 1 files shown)


czerw (approved):
LGTM, but i thought that my name was already removed from that test

Vogtinator:
> LGTM, but i thought that my name was already removed from that test

`# Maintainer: Petr Cervinka <pcervinka@suse.com>, Ben Chou <bchou@suse.com>`, and I couldn't add Ben as reviewer.

czerw:
@foursixnine could you please check review as well as it runs in functional group?

foursixnine:
@czerw Thanks, looks good, https://progress.opensuse.org/issues/101358 should be taken into account too

PR #14358, opened 2022-02-25
Increase timeout on yast2 firewall add service

dzedro:
- Fail: https://openqa.suse.de/tests/8228466#step/setup/27
- Verification run: https://dzedro.suse.cz/tests/21030


dzedro:
I will try to debug why is this command slower. First this happened on my instance I thought I have something wrong, but now it is happening on new SAP workers.

PR #14359, opened 2022-02-25
Increase timeout for slurm command

b10n1k:
This solves timeouts on the underline test `hpc_BETA_slurm_*` , especially on aarch64 arch.

Removes also some unused libraries.

- Verification run: 
https://openqa.suse.de/tests/8228776
https://openqa.suse.de/tests/8228771

### DIFF
--- a/tests/hpc/slurm_master.pm
+++ b/tests/hpc/slurm_master.pm
@@ -1,22 +1,18 @@
 # SUSE's openQA tests
 #
-# Copyright 2019-2020 SUSE LLC
+# Copyright 2019-2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Slurm master node
 #    This test is setting up slurm master node and runs tests depending
 #    on the slurm cluster configuration
-# Maintainer: Sebastian Chlad <sebastian.chlad@suse.com>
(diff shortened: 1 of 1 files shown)


tests/hpc/slurm_master.pm
```diff
@@ -1,22 +1,18 @@
...
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Slurm master node
 #    This test is setting up slurm master node and runs tests depending
 #    on the slurm cluster configuration
-# Maintainer: Sebastian Chlad <sebastian.chlad@suse.com>
+# Maintainer: Sebastian Chlad <sebastian.chlad@suse.com>, Yiannis Bonatakis <ybonatakis@suse.com>
 
-use base 'hpcbase';
+use Mojo::Base 'hpcbase';
```
czerw, line 11: I'm curious about base change, why it is needed?

b10n1k: it s a small trick which inherits some modules like `strict, warnings`. So we can omit to add them to each file. https://docs.mojolicious.org/Mojo/Base

b10n1k: Let me add also a link to OpenQA for future reference. https://github.com/os-autoinst/openQA/pull/4473/files Seems that the tooling team has discovered the same trick

PR #14360, opened 2022-02-25
Add needle cleanup subroutine for sle-micro

mloviska:
Ticket: [Include the new SLE Micro Selfinstall ISO images](https://progress.opensuse.org/issues/105025)
- Verification run: http://kepler.suse.cz/tests/14230#


jlausuch:
This is good, unfortunately, not all needles are properly labelled... I wonder if we could simple unregister ALL and create our own needles for slem with specific tag.

PR #14361, opened 2022-02-26
Add y2start.log output for bootloader_s390 (poo#107659)

skriesch:
The test suite textmode-server for s390x is failing in the test
bootloader_s390. There is no log output. Further information is
written to /var/log/y2start.log. This extension will be used for
the output of the log information to fix the bug.

- Related ticket: https://progress.opensuse.org/issues/107659

### DIFF
--- a/tests/installation/bootloader_s390.pm
+++ b/tests/installation/bootloader_s390.pm
@@ -297,6 +297,8 @@ sub run {
 
     my $exception = $@;
 
+    # add y2start/log output if exception is happening
+    die join("\n", '#', `cat /var/log/YaST2/y2start.log`) if $exception;
     die join("\n", '#' x 67, $exception, '#' x 67) if $exception;
 
     # activate console so we can call wait_serial later


tests/installation/bootloader_s390.pm
```diff
@@ -298,7 +298,7 @@ sub run {
     my $exception = $@;
 
     # Add y2start.log for output
-    die join("\n", '##########', `ls -l /var/log/ 2>&1 && cat /var/log/YaST2/vncserver.log`) if $exception;
+    die join("\n", '##########', `ls -l /var/log/YaST2/ 2>&1 && cat /var/log/YaST2/vncserver.log`) if $exception;
```
b10n1k: i would suggest the same format as L302. `'#' x <multiplier>` if you dont mind

skriesch: Thank you for the suggestion. I had to debug, why the log files did not exist for a while. I have fixed that now.

PR #14371, opened 2022-02-28
console/journalctl: Update has_default_persistent_journal for Leap 15.4

Vogtinator:
With 15.4, systemd-logger became a noop, and such even the AArch64 JeOS
uses the SLE default of not having journal persistence by default.

So only Tumbleweed and 15.3 AArch64 JeOS remain.

- Verification run: https://openqa.opensuse.org/tests/2216222

### DIFF
--- a/tests/console/journalctl.pm
+++ b/tests/console/journalctl.pm
@@ -33,9 +33,9 @@ use constant {
     SEALING_DELAY => 10
 };
 
-# by default persistent journal is set for opensuse, except of leap 15.3+
+# Tumbleweed uses a persistent journal, Leap 15.3+ (except 15.3 AArch64 JeOS) inherits SLE's default
 sub has_default_persistent_journal {
-    return is_opensuse && (!is_leap('15.3+') || check_var('FLAVOR', 'JeOS-for-AArch64'));
+    return is_tumbleweed || (is_leap('=15.3') && check_var('FLAVOR', 'JeOS-for-AArch64'));
 }
 
 # If the daemon is stopped uncleanly, or if the files are found to be corrupted, they are renamed using the ".journal~" suffix


tests/console/journalctl.pm
```diff
@@ -33,9 +33,9 @@ use constant {
     SEALING_DELAY => 10
 };
 
-# by default persistent journal is set for opensuse, except of leap 15.3+
+# Tumbleweed uses a persistent journal, Leap 15.3+ (except 15.3 AArch64 JeOS) inherits SLE's default
 sub has_default_persistent_journal {
-    return is_opensuse && (!is_leap('15.3+') || check_var('FLAVOR', 'JeOS-for-AArch64'));
+    return is_tumbleweed || (is_leap('15.3') && check_var('FLAVOR', 'JeOS-for-AArch64'));
```
mloviska, line 38: I guess you need to pass '=15.3'

Vogtinator: Yep, I fell into that trap again. It started the post_fail_hook which just wasted time...

ggardet (approved):

Vogtinator:
> Code changes work as expected. Should `rsyslog` be pre-installed as it is in `SLE`, or we have another exception in place ?

Good point, that should be checked as well. Having a volatile journal without `rsyslog` is also broken.

Vogtinator:
BTW:

The journalctl module has this:
`script_run("journalctl --boot=-1 | grep 'no persistent journal was found'") or die "Persistent journal should not be enabled by default on SLE!\n";`
I think that doesn't quite work as intended. In the case of a persistent journal, journalctl succeeds and grep fails -> exit code 1, in the case of no persistent journal, journalctl fails and grep succeeds (but I think that also fails, because of stderr) -> also exit code 1

> > Code changes work as expected. Should `rsyslog` be pre-installed as it is in `SLE`, or we have another exception in place ?
> 
> Good point, that should be checked as well. Having a volatile journal without `rsyslog` is also broken.

I filed https://bugzilla.opensuse.org/show_bug.cgi?id=1196615

mloviska:
> BTW:
> 
> The journalctl module has this: `script_run("journalctl --boot=-1 | grep 'no persistent journal was found'") or die "Persistent journal should not be enabled by default on SLE!\n";` I think that doesn't quite work as intended. In the case of a persistent journal, journalctl succeeds and grep fails -> exit code 1, in the case of no persistent journal, journalctl fails and grep succeeds (but I think that also fails, because of stderr) -> also exit code 1
> 
> > > Code changes work as expected. Should `rsyslog` be pre-installed as it is in `SLE`, or we have another exception in place ?
> > 
> > 
> > Good point, that should be checked as well. Having a volatile journal without `rsyslog` is also broken.
> 
> I filed https://bugzilla.opensuse.org/show_bug.cgi?id=1196615

Thanks! I will create a patch for persistent journal check.

PR #14374, opened 2022-03-01
xfstests: fix mkfs options not enable after a reboot

SeroSun:
Previous options use export and it only enable before a reboot, if it's wrote in configure file will not only enable when run xfstests but also after reboot happends during tests.

- Verification run: 
Just check if following VR bring the correct mkfs options or not.
https://openqa.suse.de/t8247996 -> this test will fail by a crash issue.
https://openqa.suse.de/t8247997 -> this will pass in the end


lansuse:
LGTM

PR #14379, opened 2022-03-01
Drop useless greps of extensions

dzedro:
I first saved the extensions output in file and did grep just the module without colors and each hardcoded character.
But then I realized this grep have no purpose. It will just grep the module does not matter in what state it is.

- Fail: https://openqa.suse.de/tests/8248277#step/suseconnect_scc/20
- Verification run: https://openqa.suse.de/tests/8248413#step/suseconnect_scc/18

### DIFF
--- a/tests/console/suseconnect_scc.pm
+++ b/tests/console/suseconnect_scc.pm
@@ -37,12 +37,8 @@ sub run {
     script_retry("$cmd", retry => 5, delay => 60, timeout => 180);
     # Check available extenstions (only present in sle)
     assert_script_run q[SUSEConnect --list-extensions];
-    if (is_sle) {
-        # What has been activated by default
-        assert_script_run q[SUSEConnect --list-extensions | grep -e '\(Activated\)'];
-        assert_script_run 'SUSEConnect --list-extensions | grep "$(echo -en \'    \e\[1mServer Applications Module\')"' unless is_sle('=12-sp5');
-        assert_script_run 'SUSEConnect --list-extensions | grep "$(echo -en \'    \e\[1mWeb and Scripting Module\')"';
-    }
+    # What has been activated by default
+    assert_script_run q[SUSEConnect --list-extensions | grep -e '\(Activated\)'] if is_sle;
 
     # add modules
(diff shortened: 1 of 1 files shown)


tests/console/suseconnect_scc.pm
```diff
@@ -37,12 +37,6 @@ sub run {
     script_retry("$cmd", retry => 5, delay => 60, timeout => 180);
     # Check available extenstions (only present in sle)
     assert_script_run q[SUSEConnect --list-extensions];
-    if (is_sle) {
-        # What has been activated by default
-        assert_script_run q[SUSEConnect --list-extensions | grep -e '\(Activated\)'];
```
jlausuch, line 42: Would it make sense to at least keep this check? This way we know that there is at least 1 module activated..

dzedro: I was thinking about it, but why ? On line 26 is first done registration.
This will print that there is something registered ? Cool, we can see that without the grep.

dzedro: But I don't mind, I can leave it there.

jlausuch: yeah, kind of obvious check :) 
Ok to remove it. Added other reviewers just in case.

jlausuch (approved):

tests/console/suseconnect_scc.pm
mloviska: It actually helps to see what is activated by default. Maybe the below ones are questionable. Nevertheless, instead of removal, I would vote for skipping the last 2 checks in case you don't expect those modules to be present in your product.

jknphy: afir that might be there just because doesn't fit in the screen the whole list.

dzedro: Sure, but it's useless anyway. Registration is done just step before.
This is like over-explaining buttons in elevator.

dzedro: There it is.

PR #14380, opened 2022-03-01
lib/main_common: Drop ristretto from the Rescue-CD

Vogtinator:
On TW it's no longer on there for some time now and this also reached 15.4
meanwhile.

- Verification runs:
  * 15.4: http://10.160.67.86/t1196
  * 15.3: http://10.160.67.86/t1197


Vogtinator:
VRs passed

PR #14386, opened 2022-03-01
Fix volatile journal check

mloviska:
1. persistent journal, journalctl succeeds and grep fails-> exit code 1
2. no persistent journal, journalctl fails and grep succeeds -> exit code 1

Issue spot by @Vogtinator, thanks!

- Verification runs: 
* [tw - simulated error](http://kepler.suse.cz/tests/14263#step/journalctl/34)
* [leap15.4 - not matching expected string](http://kepler.suse.cz/tests/14269#step/journalctl/34)
* [opensuse-15.4-JeOS-for-kvm-and-xen-x86_64-Build6.3-jeos-extra@64bit_virtio-2G](http://kepler.suse.cz/tests/14271#step/journalctl/30)
* [sle-15-SP3-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220301-1-jeos-extratest@uefi-virtio-vga](http://kepler.suse.cz/tests/14270#step/journalctl/18)
* [tw](http://kepler.suse.cz/tests/14273)
* [opensuse-15.3-JeOS-for-kvm-and-xen-x86_64-Build9.379-jeos-extra@64bit_virtio-2](http://kepler.suse.cz/tests/14272#step/journalctl/30)

### DIFF
--- a/tests/console/journalctl.pm
+++ b/tests/console/journalctl.pm
@@ -156,7 +156,7 @@ sub run {
             assert_script_run "rpm -q --conflicts systemd-logger | tee -a /dev/$serialdev | grep syslog";
         }
     } else {
-        script_run("journalctl --boot=-1 | grep 'no persistent journal was found'") or die "Persistent journal should not be enabled by default on SLE!\n";
+        validate_script_output('journalctl --no-pager --boot=-1', qr/no persistent journal was found/i) unless is_sle('<15');
         assert_script_run "mkdir -p ${\ PERSISTENT_LOG_DIR }";
         assert_script_run "systemd-tmpfiles --create --prefix ${\ PERSISTENT_LOG_DIR }";
         # test for installed rsyslog and for imuxsock existance


jlausuch (approved):

grisu48 (approved):
LGTM!

mloviska:
A lost comment -> https://github.com/mloviska/os-autoinst-distri-opensuse/commit/45e87d641159ca2dcdc40b8f1b36eadc11fa5a24#commitcomment-67722692

Vogtinator (approved):
Ideally it also checks the exit code, but that's not easily done with the current API

mloviska:
> Ideally it also checks the exit code, but that's not easily done with the current API

It does if journaclt returns non-zero code. http://kepler.suse.cz/tests/14288#step/journalctl/36

Vogtinator:
> > Ideally it also checks the exit code, but that's not easily done with the current API
> 
> It does when journactl returns non-zero code. http://kepler.suse.cz/tests/14288#step/journalctl/36

`journalctl --no-pager --boot=-1` actually succeeds in the case of no persistent journal? That's weird.

Vogtinator:
> > > > Ideally it also checks the exit code, but that's not easily done with the current API
> > > 
> > > 
> > > It does when journactl returns non-zero code. http://kepler.suse.cz/tests/14288#step/journalctl/36
> > 
> > 
> > `journalctl --no-pager --boot=-1` actually succeeds in the case of no persistent journal? That's weird.
> 
> Yes, seems like `Specifying boot ID or boot offset has no effect, no persistent journal was found.` is a valid return in case volatile journal is set. However, querying non-existing bootid while persistent setting is active yields an error.

Then this really just worked by accident at all!

`journalctl --boot=-1 | grep 'no persistent journal was found'` would succeed, and `script_run` return 0 as exit code. That's false-y, so the `die` would run.

However, `journalctl` writes the message to stderr, which is not captured by `grep`, so `grep` actually fails.

mloviska:
> > > > > Ideally it also checks the exit code, but that's not easily done with the current API
> > > > 
> > > > 
> > > > It does when journactl returns non-zero code. http://kepler.suse.cz/tests/14288#step/journalctl/36
> > > 
> > > 
> > > `journalctl --no-pager --boot=-1` actually succeeds in the case of no persistent journal? That's weird.
> > 
> > 
> > Yes, seems like `Specifying boot ID or boot offset has no effect, no persistent journal was found.` is a valid return in case volatile journal is set. However, querying non-existing bootid while persistent setting is active yields an error.
> 
> Then this really just worked by accident at all!
> 
> `journalctl --boot=-1 | grep 'no persistent journal was found'` would succeed, and `script_run` return 0 as exit code. That's false-y, so the `die` would run.
> 
> However, `journalctl` writes the message to stderr, which is not captured by `grep`, so `grep` actually fails.

Exactly! Thanks that you have pointed it out yesterday. Lemme run at least 12sp4 test run ( cloning in progress) to verify the behaviour.

mloviska:
> > > > > Ideally it also checks the exit code, but that's not easily done with the current API
> > > > 
> > > > 
> > > > It does when journactl returns non-zero code. http://kepler.suse.cz/tests/14288#step/journalctl/36
> > > 
> > > 
> > > `journalctl --no-pager --boot=-1` actually succeeds in the case of no persistent journal? That's weird.
> > 
> > 
> > Yes, seems like `Specifying boot ID or boot offset has no effect, no persistent journal was found.` is a valid return in case volatile journal is set. However, querying non-existing bootid while persistent setting is active yields an error.
> 
> Then this really just worked by accident at all!
> 
> `journalctl --boot=-1 | grep 'no persistent journal was found'` would succeed, and `script_run` return 0 as exit code. That's false-y, so the `die` would run.
> 
> However, `journalctl` writes the message to stderr, which is not captured by `grep`, so `grep` actually fails.

Seems like older systemd version do not have the `no persistent...` message http://kepler.suse.cz/tests/14298#step/journalctl/34

mloviska:
> > > > > > Ideally it also checks the exit code, but that's not easily done with the current API
> 
> Seems like older systemd version do not have the `no persistent...` message http://kepler.suse.cz/tests/14298#step/journalctl/34

It is only in sle15+

* [12sp5](http://kepler.suse.cz/tests/14302#step/journalctl/34)
* [12sp4](http://kepler.suse.cz/tests/14297#step/journalctl/34)
* [15](http://kepler.suse.cz/tests/14301#step/journalctl/35)

PR #14392, opened 2022-03-02
Don't install unsupported or unavailable packages

dzedro:
https://openqa.suse.de/tests/8254161
https://openqa.suse.de/tests/8254131

This looks like bug https://openqa.suse.de/tests/8254141#step/zypper_in/26
but that is not reason to break everything else!

- Related ticket: https://progress.opensuse.org/issues/105948
- Verification run: https://openqa.suse.de/tests/8251592


dzedro:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14396

PR #14366, opened 2022-02-28
Test for zypper single transaction

asdil12:
See: https://lists.opensuse.org/archives/list/zypp-devel@lists.opensuse.org/thread/YEFSZZY7FRGJPGPWKNCUT3UXQWVENOCL/
Ticket: https://progress.opensuse.org/issues/105948

Verification:
sle: http://kazhua.qa.suse.de/tests/64
tw: http://kazhua.qa.suse.de/tests/63

PR #14393, opened 2022-03-02
Fix zypper_in test for maintainanc

aylen384:
Fix tests like https://openqa.suse.de/tests/8253225#step/zypper_in/26

Verification: http://kazhua.qa.suse.de/tests/65

PR #14394, opened 2022-03-02
Revert "Test for zypper single transaction"

DimStar77:
Reverts os-autoinst/os-autoinst-distri-opensuse#14366

Causing stagings to block as discussed on #14366


asdil12:
There is already https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14393

PR #14397, opened 2022-03-02
Don't run salt on live media

Vogtinator:
The `is_staging` condition was added in 2017, probably no longer necessary.

Not running it on live media saves RAM, which makes the rest of the test faster and more reliable.

Verification runs:
- Leap 15.4 Xfce Live: http://10.160.67.86/tests/1200

No VR for SLE stagings, they all use the YAML schedule.

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1195,8 +1195,8 @@ sub load_consoletests {
         loadtest "console/puppet";
     }
     # salt in SLE is only available for SLE12 ASMM or SLES15 and variants of
-    # SLES but not SLED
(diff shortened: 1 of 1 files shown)


okurz (commented):
As long as we have the code for "is_staging" and as long as we have products/sle/main.pm I suggest to keep the "is_staging" calls where it makes sense. Here IMHO it still makes sense so I suggest to keep it.

Vogtinator:
> Here IMHO it still makes sense so I suggest to keep it.

FWICT `salt` is available in stagings, so IMO it doesn't make sense.

okurz (approved):
Well, ok then

PR #14400, opened 2022-03-02
Containers: Implement CONTAINER_IMAGES_TO_TEST variable

pdostal:
- Related ticket: [poo#107776](https://progress.opensuse.org/issues/107776)
- Verification run: [with](http://pdostal-server.suse.cz/tests/13458) and [without](http://pdostal-server.suse.cz/tests/13459) the variable

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -14,11 +14,13 @@ use Exporter;
 use strict;
 use warnings;
 use testapi;
+use Mojo::Util 'trim';
 use Utils::Architectures;
 use version_utils qw(is_sle is_opensuse is_tumbleweed is_leap is_microos is_sle_micro is_released);
 
 our @EXPORT = qw(
   get_opensuse_registry_prefix
+  get_urls_from_var
   get_suse_container_urls
   get_3rd_party_images
 );
@@ -258,6 +260,14 @@ our %images_uri = (
 sub supports_image_arch {
     my ($distri, $version, $arch) = @_;
     (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
+}
+
+# Returns a tuple of image urls.
+# If empty, no images available.
+sub get_urls_from_var {
+    my $var = shift;
+    my @urls = split(/,/, trim(get_var($var)));
+    return (\@urls);
 }
 
 # Returns a tuple of image urls and their matching released "stable" counterpart.
--- a/tests/containers/image.pm
+++ b/tests/containers/image.pm
@@ -13,7 +13,7 @@ use testapi;
 use utils;
 use containers::common;
 use containers::container_images;
-use containers::urls 'get_suse_container_urls';
+use containers::urls qw(get_suse_container_urls get_urls_from_var);
 
 sub run {
     my ($self, $args) = @_;
@@ -27,14 +27,19 @@ sub run {
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGE_VERSIONS
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
-        my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
-        my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
+        my $images_to_test;
+        # Get list of images from CONTAINER_IMAGES_TO_TEST or use the default
+        unless ($images_to_test = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')) {
+            my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
+            $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
+        }
+
         for my $iname (@{$images_to_test}) {
             record_info "IMAGE", "Testing image: $iname";
             test_container_image(image => $iname, runtime => $engine);
             test_rpm_db_backend(image => $iname, runtime => $engine);
             my $beta = $version eq get_var('VERSION') ? get_var('BETA', 0) : 0;
-            test_opensuse_based_image(image => $iname, runtime => $engine, version => $version, beta => $beta);
+            test_opensuse_based_image(image => $iname, runtime => $engine, version => $version, beta => $beta) unless ($iname =~ /bci/);
         }
     }
     scc_restore_docker_image_credentials() if ($runtime eq 'docker');
--- a/tests/containers/push_container_image_to_acr.pm
+++ b/tests/containers/push_container_image_to_acr.pm
@@ -9,7 +9,7 @@
 
 use Mojo::Base 'publiccloud::basetest';
 use testapi;
-use containers::urls 'get_suse_container_urls';
+use containers::urls qw(get_suse_container_urls get_urls_from_var);
 
 sub run {
     my ($self, $args) = @_;
@@ -17,8 +17,13 @@ sub run {
     $self->select_serial_terminal;
 
     my $provider = $self->provider_factory(service => 'ACR');
-    my ($untested_images, $released_images) = get_suse_container_urls();
-    my $image = $untested_images->[0];
+
+    my $image;
+    unless ($image = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')->[0]) {
+        # Get list of images from CONTAINER_IMAGES_TO_TEST or use the default
+        my ($untested_images, $released_images) = get_suse_container_urls();
+        $image = $untested_images->[0];
+    }
     my $tag = $provider->get_default_tag();
 
     record_info('Pull', "Pulling $image");
--- a/tests/containers/push_container_image_to_ecr.pm
+++ b/tests/containers/push_container_image_to_ecr.pm
@@ -9,19 +9,22 @@
 
 use Mojo::Base 'publiccloud::basetest';
 use testapi;
-use containers::urls 'get_suse_container_urls';
+use containers::urls qw(get_suse_container_urls get_urls_from_var);
 
 sub run {
     my ($self, $args) = @_;
 
     $self->select_serial_terminal;
 
-    my ($untested_images, $released_images) = get_suse_container_urls();
-
     my $provider = $self->provider_factory(service => 'ECR');
     $self->{provider} = $provider;
 
-    my $image = $untested_images->[0];
+    my $image;
+    unless ($image = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')->[0]) {
+        # Get list of images from CONTAINER_IMAGES_TO_TEST or use the default
+        my ($untested_images, $released_images) = get_suse_container_urls();
+        $image = $untested_images->[0];
+    }
     my $tag = $provider->get_default_tag();
     $self->{tag} = $tag;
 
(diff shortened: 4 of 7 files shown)


variables.md
```diff
@@ -195,7 +195,8 @@ QAM_MINIMAL | string | "full" or "small" | Full is adding patterns x11, gnome-ba
 RAIDLEVEL | integer | | Define raid level to be configured. Possible values: 0,1,5,6,10.
 REBOOT_TIMEOUT | integer | 0 | Set and handle reboot timeout available in YaST installer. 0 disables the timeout and needs explicit reboot confirmation.
 REGISTRY | string | docker.io | Registry to pull third-party container images from
-CONTAINER_IMAGE_VERSIONS | string | | List of comma-separated versions from `get_suse_container_urls()` to test in `tests/containers/*_image.pm`
+CONTAINER_IMAGE_VERSIONS | string | | List of comma-separated versions from `get_suse_container_urls()`
+CONTAINER_IMAGES_TO_TEST | string | | The URL of a specific container image to test.
```
jlausuch, line 199: My initial idea of this variable was for single image instead of array, as you describe it here, so it would be `CONTAINER_IMAGE_TO_TEST`.
But in the code `get_urls_from_var()` you are giving the possibility to provide an array, which is not a bad idea either. If we go the "array" way, we should also mention it here, e.g. `List of comma-separated URLs of container images to be tested`

pdostal: I decided to go this way because of if `tests/containers/image.pm` we have:
```perl
for my $iname (@{$images_to_test}) {
```
So even if I keep `CONTAINER_IMAGE_TO_TEST` I'd need to translate it to array.

pdostal: Fixed.

pdostal:
@ilausuch can you please also take a look? I did verification runs [with](http://pdostal-server.suse.cz/tests/13461) and [without](http://pdostal-server.suse.cz/tests/13460) the variable on EKS as well.

jlausuch:
I had a quick thought that after this PR we can even get rid of  `[urls.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/containers/urls.pm)`  along with `CONTAINER_IMAGE_VERSIONS`, `CONTAINERS_UNTESTED_IMAGES` vars that are a bit confusing. This way, we could just use this new variable which would include full URL of the image to be tested.

But looking at the test matrix where we are testing a lot of images on single Host, the value of the variable would be too long and difficult to read... e.g. 
`CONTAINER_IMAGES_TO_TEST=registry.suse.de/suse/sle-15/update/cr/totest/images/suse/sle15:15.0,registry.suse.de/suse/sle-15-sp2/update/cr/totest/images/suse/sle15:15.2,registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3, etc etc`.

Ideas? Other thoughts?

dcermak:
> But looking at the test matrix where we are testing a lot of images on single Host, the value of the variable would be too long and difficult to read... e.g. `CONTAINER_IMAGES_TO_TEST=registry.suse.de/suse/sle-15/update/cr/totest/images/suse/sle15:15.0,registry.suse.de/suse/sle-15-sp2/update/cr/totest/images/suse/sle15:15.2,registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3, etc etc`.
> 
> Ideas? Other thoughts?

Oh boy, this is exactly the problem that I am facing with the python based tests as well: how do you run the tests on a varying number of images with potentially different urls. I don't have a great solution either, but I think it'll come down to having a json file with all the images, each of them having an alias and then an environment variable where you specify all images by alias.

Would something like that work here?

lib/containers/urls.pm
```diff
@@ -260,6 +261,13 @@ sub supports_image_arch {
     (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
 
+# Returns a tuple of image urls.
+# If empty, no images available.
+sub get_urls_from_var {
+    my @urls = split(/,/, get_var('CONTAINER_IMAGES_TO_TEST'));
```
ilausuch, line 267: I think will be nice some extra code here to avoid human mistakes, such as extra spaces,..

jlausuch:
> > But looking at the test matrix where we are testing a lot of images on single Host, the value of the variable would be too long and difficult to read... e.g. `CONTAINER_IMAGES_TO_TEST=registry.suse.de/suse/sle-15/update/cr/totest/images/suse/sle15:15.0,registry.suse.de/suse/sle-15-sp2/update/cr/totest/images/suse/sle15:15.2,registry.suse.de/suse/sle-15-sp3/update/cr/totest/images/suse/sle15:15.3, etc etc`.
> > Ideas? Other thoughts?
> 
> Oh boy, this is exactly the problem that I am facing with the python based tests as well: how do you run the tests on a varying number of images with potentially different urls. I don't have a great solution either, but I think it'll come down to having a json file with all the images, each of them having an alias and then an environment variable where you specify all images by alias.
> 
> Would something like that work here?

This would be perfect, some centric place (outside of openQA) where we can query for urls in our tests... I note it down as an idea with potential implementation, thanks! :)

jlausuch (approved):
LGTM

lib/containers/urls.pm
pdostal: Done.

lib/containers/urls.pm
```diff
@@ -260,6 +262,13 @@ sub supports_image_arch {
     (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
 
+# Returns a tuple of image urls.
+# If empty, no images available.
+sub get_urls_from_var {
```
grisu48, line 267: This name is too unspecific. Or rename `get_urls_from_var` to something more specific e.g. `get_container_image_urls` or pass the var as argument, e.g.

```suggestion
# Return tuple of image urls from the specified variable
sub get_urls_from_var {
    my $var = $_[0];
    my @urls = split(/,/, trim(get_var($var, "")));
    s{^\s+|\s+$}{}g foreach @urls;         # Trim each element in the array
    return (\@urls);
}
```

tests/containers/image.pm
```diff
@@ -27,14 +27,18 @@ sub run {
     # We may test either one specific image VERSION or comma-separated CONTAINER_IMAGE_VERSIONS
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
-        my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
-        my $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
+        my $images_to_test;
+        unless ($images_to_test = get_urls_from_var()) {
```
grisu48, line 31: This block is very difficult to understand. Please add a comment that explains, that this checks for the list to be empty.

```suggestion
        unless ($images_to_test = get_urls_from_var("CONTAINER_IMAGES_TO_TEST")) {
            # If the returned list is empty, get the default images
```

Even better for the understanding would be to check if the variable is present, e.g.

```suggestion
        if (get_var('CONTAINERS_UNTESTED_IMAGES', '') ne '') {
            $images_to_test = get_urls_from_var();
        } else {
             my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
             $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
        }
```

tests/containers/push_container_image_to_acr.pm
```diff
@@ -9,16 +9,20 @@
...
     my ($self, $args) = @_;
 
     $self->select_serial_terminal;
 
     my $provider = $self->provider_factory(service => 'ACR');
-    my ($untested_images, $released_images) = get_suse_container_urls();
-    my $image = $untested_images->[0];
+
+    my $image;
+    unless ($image = get_urls_from_var()->[0]) {
```
grisu48, line 22: Same as previous. This `unless` block is hard to understand and the name `get_urls_from_var` requires explicit knowledge of the function to get an idea what happens.

lib/containers/urls.pm
pdostal: I don't agree. There is already `get_suse_container_urls` so I find `get_urls_from_var` as the specific option.

pdostal: I added the comment. Please check again.

tests/containers/image.pm
pdostal: I added the comment as I don't understand the code you suggested. Please check again.

tests/containers/push_container_image_to_acr.pm
pdostal: I added the variable as a parameter and also added a comment. Can you check again?

pdostal:
All verification runs were restarted and succeeded.

tests/containers/push_container_image_to_acr.pm
```diff
@@ -9,16 +9,21 @@
...
     my ($self, $args) = @_;
 
     $self->select_serial_terminal;
 
     my $provider = $self->provider_factory(service => 'ACR');
-    my ($untested_images, $released_images) = get_suse_container_urls();
-    my $image = $untested_images->[0];
+
+    my $image;
+    unless ($image = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')->[0]) {
```
grisu48, line 22: Seeing the exact same routine in 3 different files kinda motivates to extract a subroutine for it, WDYT?

tests/containers/push_container_image_to_ecr.pm
```diff
@@ -9,19 +9,22 @@
...
     $self->select_serial_terminal;
 
-    my ($untested_images, $released_images) = get_suse_container_urls();
-
     my $provider = $self->provider_factory(service => 'ECR');
     $self->{provider} = $provider;
 
-    my $image = $untested_images->[0];
+    my $image;
+    unless ($image = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')->[0]) {
```
grisu48, line 23: Seeing the exact same routine in 3 different files kinda motivates to extract a subroutine for it, WDYT?

tests/containers/push_container_image_to_gcr.pm
```diff
@@ -9,16 +9,21 @@
...
     my ($self, $args) = @_;
 
     $self->select_serial_terminal;
 
     my $provider = $self->provider_factory(service => 'GCR');
-    my ($untested_images, $released_images) = get_suse_container_urls();
-    my $image = $untested_images->[0];
+
+    my $image;
+    unless ($image = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')->[0]) {
```
grisu48, line 22: Seeing the exact same routine in 3 different files kinda motivates to extract a subroutine for it, WDYT?

tests/containers/push_container_image_to_acr.pm
pdostal: Yes, the subroutine is called `get_urls_from_var` and already exists.

tests/containers/push_container_image_to_ecr.pm
pdostal: Yes, the subroutine is called `get_urls_from_var` and already exists.

tests/containers/push_container_image_to_gcr.pm
pdostal: Yes, the subroutine is called `get_urls_from_var` and already exists.

grisu48 (approved):

PR #14402, opened 2022-03-02
Handle command retry in register_addons_cmd

jlausuch:
The code was hardcoding the number of retries to 1, but it would
be more meaningful that that parameter is given by the test.

Failure: https://openqa.suse.de/tests/8254899#step/suseconnect_scc/50
VR: https://openqa.suse.de/tests/8255243

### DIFF
--- a/tests/console/suseconnect_scc.pm
+++ b/tests/console/suseconnect_scc.pm
@@ -24,6 +24,7 @@ sub run {
     # please unset in job def *SCC_URL* if not required
     my $fake_scc = get_var 'SCC_URL', '';
     $cmd .= ' --url ' . $fake_scc if $fake_scc;
+    my $retries = 5;    # number of retries to run SUSEConnect commands
 
 
     $self->select_serial_terminal;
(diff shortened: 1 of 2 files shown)


mloviska (approved):

pdostal (approved):
@asmorodskyi we use `add_suseconnect_product` also in `tests/publiccloud/` do you think that we should add this `retry` parameter there as well?

jknphy (approved):

asmorodskyi:
> @asmorodskyi we use `add_suseconnect_product` also in `tests/publiccloud/` do you think that we should add this `retry` parameter there as well?

the changes done directly into the function so my understanding it will already increase amount of attempts to 3 (default value) , right ? so let's see how it will affect us and then discuss if we need to tune it up or not

tests/console/suseconnect_scc.pm
```diff
@@ -41,7 +41,7 @@ sub run {
     assert_script_run q[SUSEConnect --list-extensions | grep -e '\(Activated\)'] if is_sle;
 
     # add modules
-    register_addons_cmd if $scc_addons;
+    register_addons_cmd($scc_addons, 5) if $scc_addons;
```
asmorodskyi, line 44: there is also retry in line 37 , maybe we can extract amount of retries to a variable ?

asmorodskyi (approved):
LGTM , small suggestion is not a must feel free to ignore it

tests/console/suseconnect_scc.pm
jlausuch: yes, good point

jlausuch:
VR https://openqa.suse.de/tests/8255471

PR #14407, opened 2022-03-02
Improve softreboot to work on TUNNELED instances

grisu48:
Improve the softreboot for publiccloud, so it works on TUNELLED test
runs without the need of the user to re-establish the tunnel afterwards.

- Verification run: [SLE 15-SP4 GCE](http://duck-norris.qam.suse.de/tests/8319) | [15-SP3 Azure containers](http://duck-norris.qam.suse.de/tests/8316) | [15-SP3 GCE imgproof](http://duck-norris.qam.suse.de/tests/8317) | [15-SP2 Staging images EC2 ltp](http://duck-norris.qam.suse.de/tests/8312) | [12-SP5 GCE-BYOS imgproof](http://duck-norris.qam.suse.de/tests/8318) | [12-SP5 EC2-BYOS ltp-syscalls](http://duck-norris.qam.suse.de/tests/8320)

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -14,6 +14,7 @@ use Mojo::Base -base;
 use Mojo::Util 'trim';
 use File::Basename;
 use publiccloud::utils;
+use publiccloud::ssh_interactive;
 use version_utils;
 
 use constant SSH_TIMEOUT => 90;
@@ -273,9 +274,17 @@ sub softreboot
 
     my $duration;
 
+    # On TUNNELED test runs, ensure we are not on the publiccloud instance
+    my $prev_console = current_console();
+    # We only need to re-establish the ssh tunnel, if we are in a TUNNELED test run, and if the tunnel is already initialized
+    my $tunneled = get_var('TUNNELED', 0) && get_var("_SSH_TUNNELS_INITIALIZED", 0);
+    if ($tunneled) {
+        select_host_console(force => 1);
+        ssh_interactive_leave();
+    }
+
     $self->run_ssh_command(cmd => 'sudo shutdown -r +1');
-    # skip the one minute waiting
-    sleep 60;
+    sleep 60;    # wait for the +1 in the previous command
     my $start_time = time();
 
     # wait till ssh disappear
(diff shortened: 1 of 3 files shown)


lib/publiccloud/instance.pm
```diff
@@ -273,9 +274,17 @@ sub softreboot
 
     my $duration;
 
+    # On TUNNELED test runs, ensure we are not on the publiccloud instance
+    my $prev_console = current_console();
+    # Need to store in variable, because the var can be altered
+    my $tunneled = get_var('TUNNELED', 0) && get_var("_SSH_TUNNELS_INITIALIZED", 0);
```
pdostal, line 280: I don't understand what the `&&` means here. Can you elaborate on what possible scenarios can happen over here?

pdostal (approved):
This looks good :smile:

lib/publiccloud/instance.pm
jlausuch: I guess to make sure that we have reached the point that the tunnel is established: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/utils.pm#L44

pdostal: Then let's use `testapi::check_var()`? See the [testapi docummentation](http://open.qa/api/testapi/#_check_var).

lib/publiccloud/ssh_interactive.pm
```diff
@@ -18,7 +18,16 @@ our @ISA = qw(Exporter);
...
+    # Establish the ssh interarctive tunnel to the publiccloud instance.
+    # Optional arguments: 'force => 1' - reestablish tunnel, also if already established.
+
+    my $instance = shift;
+    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @_);
+    $args{force} //= 0;
+
+    # $serialdev should be always set from os-autoinst/testapi.pm
+    die '$serialdev is not set' if (!defined $serialdev || $serialdev eq '');
+    return if ($args{force} != 1 && get_var('_SSH_TUNNELS_INITIALIZED', 0) == 1);
```
pdostal, line 30:
```suggestion
    return if ($args{force} != 1 && check_var('_SSH_TUNNELS_INITIALIZED', 1));
```

lib/publiccloud/instance.pm
grisu48: We only need to re-establish the ssh tunnel after rebooting, if we are in a TUNNELED test run and if the ssh tunnel has been established before.

I also updated the comment accordingly.

lib/publiccloud/ssh_interactive.pm
grisu48: Apart from style, is there any other reason for doing this change? I honestly don't like `check_var` and try to avoid that whenever possible.

pdostal: Yes, this is of course minor.

pdostal (approved):

lib/publiccloud/instance.pm
pdostal: That's better :smile: Thanks

PR #14408, opened 2022-03-02
Adapt yast2_kdump after UI change

JRivrain:
We are missing ret key in https://openqa.suse.de/tests/8216714#step/yast2_kdump/20
It seems like there was some UI change that makes it necessary to hit enter twice: https://openqa.suse.de/tests/8257437#step/yast2_kdump/19 we enter 320, which is more than the maximum allowed value. So Yast changes it to the max allowed value, 287, after we type enter, but then we need to type enter again. Another option would be to set $memory_kdump to 287 only for aarch64, but that might cause problems again if the value changes for some reason. Typing enter twice does not hurt in this case, as we see on ppc.

- Related ticket: https://progress.opensuse.org/issues/107314

VR 
aarch64 https://openqa.suse.de/tests/8263831
ppc https://openqa.suse.de/tests/8264596

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -169,7 +169,7 @@ sub activate_kdump {
     if (is_ppc64le || is_aarch64) {
         send_key('alt-y');
         type_string $memory_kdump;
-        send_key('ret');
+        wait_screen_change(sub { send_key 'ret' }, 10) for (1 .. 2);
         record_soft_failure 'default kdump memory size is too small for ppc64le and aarch64, see bsc#1161421';
         $expect_restart_info = 1;
     }


lib/kdump_utils.pm
```diff
@@ -169,7 +169,7 @@ sub activate_kdump {
     if (is_ppc64le || is_aarch64) {
         send_key('alt-y');
         type_string $memory_kdump;
-        send_key('ret');
+        send_key_until_needlematch('yast2-kdump-restart-info', 'ret');
```
jknphy, line 172: I would suggest to avoid having the possibility to trigger two `returns` in case the needle doesn't match for some reason...it might press the next OK accidentally, deleting the field before typing on it might make the trick, for example.

JRivrain: It seems like there was some UI change that makes it necessary to hit enter twice: https://openqa.suse.de/tests/8257437#step/yast2_kdump/19 we enter 320, which is more than the maximum allowed value. So Yast changes it to the max allowed value, 287, after we type enter, but then we need to type enter again. Another option would be to set $memory_kdump to 287 only for aarch64, but that might cause problems again if the value changes for some reason. Typing enter twice does not hurt in this case.

rakoenig (approved):
LGTM

jknphy:
arent' we missing something here? suggested bigger than max? https://openqa.suse.de/tests/8207242#step/yast2_kdump/16
max here 900? https://openqa.suse.de/tests/8179659#step/yast2_kdump/15 Although there is a bug open for a while...

With current code, without checking the needle in the screen, if the value is in range in next build (because looks pretty random this) would we do some extra ret?

JRivrain:
> arent' we missing something here? suggested bigger than max? https://openqa.suse.de/tests/8207242#step/yast2_kdump/16 max here 900? https://openqa.suse.de/tests/8179659#step/yast2_kdump/15 Although there is a bug open for a while...
> 
> With current code, without checking the needle in the screen, if the value is in range in next build (because looks pretty random this and could be the case, right?) wouldn't we do some extra ret?

The value is defined like this `my $memory_kdump = $memory_total >= 2048 ? 1024 : 320;`. There is why 320 https://progress.opensuse.org/issues/63772#note-9.
As we see on ppc, hitting "ret" twice  does not harm.

JRivrain:
Set back to [WIP], I want to make sure the workaround is still actually needed. Maybe the bug was fixed and nobody noticed.

JRivrain:
> Set back to [WIP], I want to make sure the workaround is still actually needed. Maybe the bug was fixed and nobody noticed.

looks like it may still be needed https://openqa.suse.de/tests/8265553 however I think we should investigate further as some there were recent changes in kdump scripts (https://build.suse.de/package/view_file/SUSE:SLE-15-SP4:GA/kdump/kdump.changes?expand=1) that may have fixed it

jknphy:
> > arent' we missing something here? suggested bigger than max? https://openqa.suse.de/tests/8207242#step/yast2_kdump/16 max here 900? https://openqa.suse.de/tests/8179659#step/yast2_kdump/15 Although there is a bug open for a while...
> > With current code, without checking the needle in the screen, if the value is in range in next build (because looks pretty random this and could be the case, right?) wouldn't we do some extra ret?
> 
> The value is defined like this `my $memory_kdump = $memory_total >= 2048 ? 1024 : 320;`. There is why 320 https://progress.opensuse.org/issues/63772#note-9. As we see on ppc, hitting "ret" twice does not harm.

I agree to merge it as it solve the problem, but the fact that in one build for another arch hitting twice `ret` doesn't harm does not ensure that when the value is in range in a different arch and build, the second `ret` it is not needed and might go to the next screen depending on the execution speed.

JRivrain:
> > > arent' we missing something here? suggested bigger than max? https://openqa.suse.de/tests/8207242#step/yast2_kdump/16 max here 900? https://openqa.suse.de/tests/8179659#step/yast2_kdump/15 Although there is a bug open for a while...
> > > With current code, without checking the needle in the screen, if the value is in range in next build (because looks pretty random this and could be the case, right?) wouldn't we do some extra ret?
> > 
> > 
> > The value is defined like this `my $memory_kdump = $memory_total >= 2048 ? 1024 : 320;`. There is why 320 https://progress.opensuse.org/issues/63772#note-9. As we see on ppc, hitting "ret" twice does not harm.
> 
> I agree to merge it as it solve the problem, but the fact that in one build for another arch hitting twice `ret` doesn't harm does not ensure that when the value is in range in a different arch and build, the second `ret` it is not needed and might go to the next screen depending on the execution speed.

This workaround is limited to aarch64 ans ppc64, and we can see that on ppc it does not cause trouble, even though the second return is not needed here. I created a follow-up ticket https://progress.opensuse.org/issues/107815

PR #14413, opened 2022-03-03
Add three test cases to Audit test module

ShawnHa0:
Add auditd, auditctl, aulast to Audit module.
Auditd: Controlling audit daemon as root by systemctl to verify it can
work
Auditctl: Controlling the audit system using auditctl
Aulast: Verify the aualst can print a list of the last logged-in users

- Related ticket: https://progress.opensuse.org/issues/81772
- Needles: N/A
- Verification run: 
  - o3: https://openqa.opensuse.org/tests/2243169#
  - osd: https://openqa.suse.de/tests/8323123#

### DIFF
--- a/tests/security/audit/auditctl.pm
+++ b/tests/security/audit/auditctl.pm
@@ -1,0 +1,62 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-Later
+#
+# Summary: Controlling the Audit system using auditctl
+# Maintainer: llzhao <llzhao@suse.com>, shawnhao <weixuan.hao@suse.com>
+# Tags: poo#81772, tc#1768551
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub run {
+    my $audit_rules = '/etc/audit/rules.d/audit.rules';
+    my $audit_log = '/var/log/audit/audit.log';
+    select_console 'root-console';
+
+    # Disable audit
+    validate_script_output('auditctl -e 0', sub { m/enabled 0/ });
+
+    # Make sure auditctl indeed disabled auditd
+    assert_script_run("echo '' > $audit_log");
+    systemctl('restart apparmor');
+    my $ret = script_run("tail -1 $audit_log | grep SERVICE_START");
+    if ($ret == 0) {
+        # $ret == 0, report error since there should be no audit logs related to apparmor restart
+        record_info('Error: ', 'unexpected audit log recorded', result => 'fail');
+    } else {
+        # $ret == 1 is the expected return value
+        record_info('Checked: ', 'auditd temporarily disabled as expected');
+    }
+
+    # Check audit status
+    validate_script_output('auditctl -s', sub { m/enabled 0/ });
+
+    # Enable audit
+    validate_script_output('auditctl -e 1', sub { m/enabled 1/ });
+
+    # Make sure auditctl indeed enabled auditd
+    assert_script_run("echo '' > $audit_log");
+    systemctl('restart apparmor');
+    $ret = script_run("tail -1 $audit_log | grep SERVICE_START");
+    if ($ret == 0) {
+        # $ret == 0 is the expected return value showing that auditd is enabled now
+        record_info('Checked: ', 'auditd temporarily enabled as expected');
+    } else {
+        # $ret == 1, report error that apparmor log has not been recorded
+        record_info('Error: ', 'auditd failed to record apparmor logs', result => 'fail');
+    }
+
+    # Check audit status again
+    validate_script_output('auditctl -s', sub { m/enabled 1/ });
+
+    # List all rules by default
+    validate_script_output('auditctl -l', sub { m/-a never,task/ });
+
+    # Double check the audit rule file
+    validate_script_output("cat $audit_rules", sub { m/-a task,never/ });
+}
+
+1;
--- a/tests/security/audit/auditd.pm
+++ b/tests/security/audit/auditd.pm
@@ -1,0 +1,38 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-Later
+#
+# Summary: Controlling the Auditd daemon as root by "systemctl" to verify it can work
+# Maintainer: llzhao <llzhao@suse.com>, shawnhao <weixuan.hao@suse.com>
+# Tags: poo#81772, tc#1768549
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub run {
+    select_console 'root-console';
+
+    # Install packages if they are not installed by default
+    if (script_run('rpm -q audit')) {
+        zypper_call('in audit libaudit1');
+    }
+
+    # Check auditd status by default on a clean new VM
+    systemctl('is-active auditd');
+
+    # Stop auditd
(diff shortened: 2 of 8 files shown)


tests/security/audit/auditd.pm
```diff
@@ -0,0 +1,36 @@
...
+use utils;
+
+sub run {
+    select_console 'root-console';
+
+    # Make sure audit packages are installed
+    zypper_call('in audit libaudit1');
+
+    # Check auditd status by default on a clean new VM
+    assert_script_run('systemctl status --no-pager auditd', sub { m/Active: active./ });
```
rfan1, line 21: I suggest using  systemctl('is-active auditd');

tests/security/audit/auditctl.pm
```diff
@@ -0,0 +1,37 @@
...
+    validate_script_output('auditctl -s', sub { m/enabled 0/ });
+
+    # Enable audit
+    validate_script_output('auditctl -e 1', sub { m/enabled 1/ });
+
+    # Check audit status again
+    validate_script_output('auditctl -s', sub { m/enabled 1/ });
+
+    # List all rules by default
+    validate_script_output('auditctl -l', sub { m/-a never,task/ });
```
rfan1, line 31: Shawn,

IMO, you should add more checking here. e.g.

1. after disabling the auditd, you will not able to get any auditd logs/record
2. after enabling the auditd, you are able to get auditd logs/record.
Any way,  I think cc tests have covered this part, you can double check that. 

You only check the command return value, but no function check, I would suggest you add it

tests/security/audit/auditd.pm
```diff
@@ -0,0 +1,36 @@
...
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub run {
+    select_console 'root-console';
+
+    # Make sure audit packages are installed
+    zypper_call('in audit libaudit1');
```
rfan1, line 18: Please double confirm the packages are installed by default. 
I remembered they are installed, so we don't need install them any more [not 100% sure]

tests/security/audit/auditd.pm
```diff
@@ -0,0 +1,36 @@
...
+
+    # Check auditd status by default on a clean new VM
+    assert_script_run('systemctl status --no-pager auditd', sub { m/Active: active./ });
+
+    # Stop auditd
+    assert_script_run('systemctl stop auditd');
```
rfan1, line 24: We have already available function:
use utils 'systemctl'

YOu can use it :)

tests/security/audit/aulast.pm
```diff
@@ -0,0 +1,67 @@
...
+    assert_script_run("cp $audit_log $testfile");
+
+    # Run aulast -f testfile again to print last logged-in user
+    validate_script_output("aulast -f $testfile", sub { m/$user/ });
+
+    # Report bad logins
+    assert_script_run('aulast --bad');
+
+    # Let test user login localhost but fail it
+    assert_script_run(
```
rfan1, line 56: In your expect script, even no error seen in VR, I would suggest you use more strong code:

ssh -v -o StrictHostKeyChecking=no localhost

tests/security/audit/auditd.pm
rfan1: zypper_call('audit') if script_run('rpm -q audit');

tests/security/audit/auditctl.pm
ShawnHa0: Good point!
I'm gonna check the function by:
1. stop auditd and clear audit.log
2. start some service (e.g. apparmor)
3. check audit.log, should be empty
4. start auditd and clear audit.log, repeat steps above

tests/security/audit/autrace.pm
```diff
@@ -28,10 +28,6 @@ sub run {
 
     select_console 'root-console';
 
-    # Install related packages and start audit service
-    zypper_call('in audit libaudit1');
-    assert_script_run('systemctl start auditd');
-
     # Use autrace to trace an individual process, output will be logged to audit log
     script_run('autrace /bin/ls /tmp');
     record_info('autrace_output: ', 'autrace will report error here as expected');
```
lilyeyes, line 33: It depends the audit related config file, so please check the return value of `script_run('autrace /bin/ls /tmp');` if failed you can do "record_info", if passed everything is OK.

lilyeyes (approved):
Generally looks good to me.

jouyingbin:
@ShawnHa0 I just see there is still CI check fail, are you still modifying the PR?

ShawnHa0:
> @ShawnHa0 I just see there is still CI check fail, are you still modifying the PR?

Yes, I'm modifying the codes as the suggestions.

tests/security/audit/auditd.pm
ShawnHa0: Changed.

tests/security/audit/auditd.pm
ShawnHa0: Added package checking.

tests/security/audit/autrace.pm
ShawnHa0: Added record_info for both pass and fail scenarios. It will return error if autrace returns with unexpected reports here.

rfan1 (approved):

jouyingbin:
minor comment:
you miss " **,** " between poo and tc from the Tags.
Others LGTM.

tests/security/audit/auditctl.pm
```diff
@@ -0,0 +1,58 @@
...
+
+    # Disable audit
+    validate_script_output('auditctl -e 0', sub { m/enabled 0/ });
+
+    # Clear audit log, generate audit reports and these should not be recorded
+    assert_script_run("echo -n '' > $audit_log");
+    systemctl('stop apparmor');
+    systemctl('start apparmor');
+    if ([-s $audit_log]) {
+        record_info('Auditd disabled: ', 'No audit reports have been recorded');
```
Amrysliu, line 27: `-s  File has nonzero size (returns size in bytes).`
the message in record_info is opposite.

Amrysliu: `[]` is not mandatory.

ShawnHa0: Here audit log has been cleared so recorded information of audit log should be empty (or more precisely 'null' since I'm using `echo -n` here). So I think it is corret that `-s audit_log` should fail here and return 1 and it suggests the log is indeed empty.

ShawnHa0:
> minor comment: you miss " **,** " between poo and tc from the Tags. Others LGTM.

Thanks! Commas added to all previous cases too.

jouyingbin (approved):

ShawnHa0:
This pr is under further research and investigating now upon behavior and anticipated result of 'auditctl' now.

lilyeyes (approved):
The new changes look good to me. Thanks.

tests/security/audit/auditd.pm
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    # Check auditd status by default on a clean new VM
+    systemctl('is-active auditd');
+
+    # Stop auditd
+    systemctl('stop auditd');
+
+    # Check auditd status
+    script_run('systemctl status --no-pager auditd', sub { m/Active: inactive./ });
```
Amrysliu, line 29: why not `validate_script_output`? I see you want to check the output, right?

ShawnHa0: updated

Amrysliu (approved):

PR #14416, opened 2022-03-03
Hpc best practices

b10n1k:
- Use of Mojo::Base https://docs.mojolicious.org/Mojo/Base
      strict and warnings are inherited by this Class
- Use kernel maillist for maintainer                                                                                                                                                                                                 
                                                                                                                                                                                                                                             
With this PR all the HPC modules introduce the `signatures` flag which is                                                                                                                                                                    
supported by `Mojo::Base`. The signature declares lexical variables that are                                                                                                                                                                 
in scope for the block                                                                                                                                                                                                                       
and every subroutine invocation, first is checked for its signature.


- Verification run: 
https://openqa.suse.de/tests/8266698
https://openqa.suse.de/tests/8266713#dependencies

### DIFF
--- a/lib/hpc/cluster.pm
+++ b/lib/hpc/cluster.pm
@@ -7,7 +7,7 @@
 # Maintainer: Kernel QE <kernel-qa@suse.de>
 
 package hpc::cluster;
-use Mojo::Base 'hpcbase';
+use Mojo::Base 'hpcbase', -signatures;
 use testapi;
 use utils;
 
@@ -15,8 +15,7 @@ our @EXPORT = qw(
   provision_cluster
 );
 
-sub provision_cluster {
-    my ($self) = @_;
+sub provision_cluster ($self) {
     my $config = << "EOF";
 sed -i '/^DHCLIENT_SET_HOSTNAME.*/c\\DHCLIENT_SET_HOSTNAME="no"' /etc/sysconfig/network/dhcp
 EOF
--- a/lib/hpc/cluster.pm
+++ b/lib/hpc/cluster.pm
@@ -4,12 +4,10 @@
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Base module for HPC cluster provisioning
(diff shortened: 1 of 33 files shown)


lib/hpc/cluster.pm
```diff
@@ -1,24 +1,21 @@
 # SUSE's openQA tests
 #
-# Copyright 2020 SUSE LLC
+# Copyright 2020-2022 SUSE LLC
```
okurz, line 3: don't update copyright years with ranges, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md?plain=1#L49

tests/hpc/ganglia_client.pm
```diff
@@ -1,23 +1,18 @@
 # SUSE's openQA tests
 #
-# Copyright 2018 SUSE LLC
+# Copyright 2018-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Ganglia Test - client
 #   Acts as client node, which publishes data to the server via gmetric command
-# Maintainer: soulofdestiny <mgriessmeier@suse.com>
-# Tags: https://fate.suse.com/323979
```
okurz, line 9: I suggest to keep that. Even if fate is not actively used anymore it still acts as a reference explaining why we have test modules at all

okurz (changes requested):

lib/hpc/cluster.pm
b10n1k: *this change is ready for review*

tests/hpc/ganglia_client.pm
b10n1k: *this change is ready for review*

okurz (approved):
Please await review from former and updated test module maintainers

schlad:
nice cleanup! thx @b10n1k

dzedro:
https://openqa.suse.de/tests/8284597

PR #14419, opened 2022-03-03
Fix race condition in open_ssh_port

JRivrain:
We verified that the port is open before it's actually opened in some
cases (hyperv). This fixes it. 
Also now takes screenshots in overview page.

Ticket: https://progress.opensuse.org/issues/104226
VR: https://openqa.suse.de/tests/overview?build=JRivrain/os-autoinst-distri-opensuse%2314419

### DIFF
--- a/lib/Installation/InstallationSettings/InstallationSettingsPage.pm
+++ b/lib/Installation/InstallationSettings/InstallationSettingsPage.pm
@@ -32,18 +32,15 @@ sub get_overview_content {
     return $self->{rct_overview}->text();
 }
 
-sub enable_ssh_service {
-    my ($self) = @_;
-    YuiRestClient::Wait::wait_until(object => sub {
-            $self->{rct_overview}->activate_link('security--enable_sshd');
-            return $self->is_ssh_service_enabled;
-    }, message => "'SSH service will be enabled' message is not shown, though SSH service is expected to be enabled.");
-}
-
 sub is_ssh_service_enabled {
     my ($self) = @_;
     my $overview_content = $self->get_overview_content();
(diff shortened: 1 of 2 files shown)


lib/Installation/InstallationSettings/InstallationSettingsPage.pm
```diff
@@ -34,10 +35,7 @@ sub get_overview_content {
 
 sub enable_ssh_service {
     my ($self) = @_;
-    YuiRestClient::Wait::wait_until(object => sub {
-            $self->{rct_overview}->activate_link('security--enable_sshd');
-            return $self->is_ssh_service_enabled;
-    }, message => "'SSH service will be enabled' message is not shown, though SSH service is expected to be enabled.");
+    $self->{rct_overview}->activate_link('security--enable_sshd');
```
jknphy, line 38: I'm not sure about this, I remember to have seen a huge time of load after changing the link, so deleting the `wait_until` here might potentially break many jobs that run slowly.
I also don't see the point of the decoupling at this level, the check of the action should be in the test level, but was already designed wrong this part from the first day.
We probably just need to put outside of `wait_until` the activation of the link and increase the timeout to avoid this https://openqa.suse.de/tests/8035199#step/open_ssh_port/1 It looks like only waited 40 seconds, https://openqa.suse.de/tests/8035199/logfile?filename=yui-log.txt make sense?

jknphy: discussed offline, that we have new method to fully wait for loaded page but wasn't fully integrated, therefore the confusion.

jknphy (approved):

PR #14421, opened 2022-03-04
To make sure the firefox menu can show up

tinawang123:
To make sure the firefox menu can show up

- Related ticket: https://progress.opensuse.org/issues/103533
- Verification run: 
http://openqa.suse.de/t8262402
http://openqa.suse.de/tests/8270332#step/firefox/10
opensuse: https://openqa.opensuse.org/tests/2230818#

### DIFF
--- a/tests/x11/firefox.pm
+++ b/tests/x11/firefox.pm
@@ -23,8 +23,8 @@ sub run() {
 
     $self->start_firefox;
     wait_still_screen;
-    send_key_until_needlematch('firefox-top-bar-highlighted', 'alt', 4, 10);
-    send_key('h');
+    send_key('alt');
+    send_key_until_needlematch('firefox-top-bar-highlighted', 'alt-h', 5, 10);
     wait_still_screen;
     assert_screen('firefox-help-menu');
     send_key_until_needlematch('test-firefox-3', 'a', 9, 6);


hjluo:
LGTM

tests/x11/firefox.pm
```diff
@@ -23,8 +23,9 @@ sub run() {
 
     $self->start_firefox;
     wait_still_screen;
-    send_key_until_needlematch('firefox-top-bar-highlighted', 'alt', 4, 10);
-    send_key('h');
+    send_key('alt');
```
rfan1, line 26: Yutao,
Can you please add some comments here why you change the hot key?
It can provide more clear info to reader :)

tests/x11/firefox.pm
```diff
@@ -23,8 +23,9 @@ sub run() {
...
+    send_key_until_needlematch('firefox-top-bar-highlighted', 'alt-h', 4, 10);
```
lemon-suse, line 27: Maybe we can change 4->5 and remove the next line send_key('alt-h');

tests/x11/firefox.pm
tinawang123: I just change back. Original code is 'alt'

lemon-suse (approved):

okurz:
Just revert it. You can't verify a fix for a sporadic issue with just a single verification run. I suggest to use https://progress.opensuse.org/projects/openqatests/wiki/Wiki#Statistical-investigation for both the analysis of the original issue and fail rate as well as any attempts of trying to improve the situation

PR #14422, opened 2022-03-04
Enable irqbalance test in MU tests

Julie-CAO:
test scenario:

- ensure guests with at least 4 vcpu core and irqbalance installed.
- stop irqbalance.service in guest if it is running.
- bind all NIC IRQs to cpu0 & check if binding is successful by reading the smp affinities(all the values should be '1'(cpu0)) in guest.
- start irqbalance.service again.
- check smp affinities are changed to be distributed among CPUs.
- get the total interrupts on each CPUs.
- ping multiple URLs to generate NIC interrupts on different IRQs.
- get the total interrupts on each CPUs, check if the new increased interrupts are distributed among CPUs

Related ticket: https://progress.opensuse.org/issues/92815
Verification run:
[multiple guests in my own openqa server](http://10.67.129.102/tests/119)
[15sp3 guests in 15sp4 host in OSD](https://openqa.nue.suse.com/tests/8279264)
[MU tests with 15sp3 xen host](http://openqa.qam.suse.cz/tests/37776)
[MU tests with 12sp5 xen host](http://openqa.qam.suse.cz/tests/38110) 
ignore this test: [MU tests with 15sp2 xen host](http://openqa.qam.suse.cz/tests/37862)   15sp2 host install fail due to LTSS code

@alice-suse @waynechen55 @guoxuguang @nanzhg @tbaev @tonyyuan1 @pdostal  Welcome review!

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -2763,6 +2763,10 @@ sub load_hypervisor_tests {
         loadtest "virt_autotest/libvirt_isolated_virtual_network";
     }
 
+    if (check_var('VIRT_PART', 'irqbalance')) {
+        loadtest "virt_autotest/xen_guest_irqbalance";
+    }
+
     if (check_var('VIRT_PART', 'snapshots')) {
         loadtest "virt_autotest/virsh_internal_snapshot";
         loadtest "virt_autotest/virsh_external_snapshot";
--- a/tests/virt_autotest/xen_guest_irqbalance.pm
+++ b/tests/virt_autotest/xen_guest_irqbalance.pm
@@ -30,6 +30,10 @@ sub run_test {
 
         #irqbalance test run on sle12sp3+ guest
         next if guest_is_sle($guest, '<=12-sp2');
+        if (guest_is_sle($guest, '=15-sp0')) {
+            record_soft_failure("Skip the test as fix for SLE15 was not requested by customer in bsc#1178477.");
+            next;
+        }
 
         record_info("Test $guest");
         prepare_guest_for_irqbalance($guest);
--- a/tests/virt_autotest/xen_guest_irqbalance.pm
+++ b/tests/virt_autotest/xen_guest_irqbalance.pm
@@ -10,6 +10,7 @@ use base "virt_feature_test_base";
 use strict;
 use warnings;
 use testapi;
+use utils 'script_retry';
 use version_utils qw(is_sle);
 use set_config_as_glue;
 use virt_autotest::common;
@@ -39,7 +40,7 @@ sub run_test {
         my $nproc = script_output("ssh root\@$guest 'nproc'");
 
         # Bind all NIC IRQs to cpu0 in guest
-        my @nic_irqs_id = split('\n', script_output("ssh root\@$guest \"grep vif /proc/interrupts | cut -d ':' -f1 | sed 's/^ *//'\""));
+        my @nic_irqs_id = split('\n', script_output("ssh root\@$guest \"grep -e vif -e eth /proc/interrupts | cut -d ':' -f1 | sed 's/^ *//'\""));
         foreach my $irq_id (@nic_irqs_id) {
             assert_script_run("ssh root\@$guest \"echo 1 > /proc/irq/$irq_id/smp_affinity\"");
         }
@@ -48,8 +49,9 @@ sub run_test {
             die "Failed to bind NIC IRQs to CPU0" if $_ ne 1;
         }
 
-        # Start the irqbalance.service again
+        # Start irqbalance.service again
         assert_script_run("ssh root\@$guest \"systemctl start irqbalance\"");
+        sleep 30;    #need some time to take effect in some cases
 
         # Check if the SMP affinities are distributed among CPUs
         my @affinities_with_irqbalance = get_irq_affinities_from_guest($guest, @nic_irqs_id);
@@ -74,8 +76,7 @@ sub run_test {
         my @initial_total_irqs_on_cpu = get_nic_irqs_distribution_from_guest($guest, $nproc);
 
         # Generate network load in guest
-        my $ping_url = "download.opensuse.org";
-        assert_script_run("ssh root\@$guest \"ping -f -c 500 $ping_url\"");
+        generate_vif_interrupts_in_guest($guest);
 
         # re-caculate the network interrupts distribution
         # the increased NIC IRQs should be distributed in CPU cores in balance
@@ -83,9 +84,13 @@ sub run_test {
         my @increased_irqs_on_cpu;
         for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
             $increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id];
-            die("The value of network interrupts from download for CPU" . $cpu_id . " is 0") if $increased_irqs_on_cpu[$cpu_id] == 0;
+            #at least a few interrupts on each cpu core
+            if ($increased_irqs_on_cpu[$cpu_id] < 10) {
+                #Please look into the soft failure to identify if it is a product bug or temporary lack of network load coverage
+                record_soft_failure("IRQ are not balanced as the vif interrupts for CPU" . $cpu_id . " is " . $increased_irqs_on_cpu[$cpu_id]);
+            }
         }
-        record_info("NIC IRQs distribution of 'ping $ping_url'", "@increased_irqs_on_cpu");
+        record_info("NIC IRQs distribution on $nproc cpu cores", "@increased_irqs_on_cpu");
     }
 
     restore_xml_changed_guests();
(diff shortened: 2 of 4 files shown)


lib/main_common.pm
```diff
@@ -2763,6 +2763,10 @@ sub load_hypervisor_tests {
         loadtest "virt_autotest/libvirt_isolated_virtual_network";
     }
 
+    if (check_var('VIRT_PART', 'irqbalance')) {
```
alice-suse, line 2766: This should be enabled for xen only, right?

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -156,15 +160,30 @@ sub prepare_guest_for_irqbalance {
...
+sub generate_vif_interrupts_in_guest {
+    my $vm_name = shift;
+
+    #interrupts are handled by different vif IRQs from different sources
+    #and always the same IRQ handle the interrupts from same sources
+    my @ping_url_list = qw(download.opensuse.org 8.8.8.8 xen.org opensuse.org libvirt.org suse.com www.perl.org);
+    my $gateway = script_output("ssh root\@$vm_name \"ip r | grep 'default via' | cut -d ' ' -f3\"");
+    push @ping_url_list, $gateway;
+    foreach my $url (@ping_url_list) {
+        script_run("ssh root\@$vm_name \"ping -f -c 20 $url\"");
```
alice-suse, line 173: Suggest to let the ping run in background with a "&" at the end of the command, so that mulitile ping can go parallelly and save test time.

lib/main_common.pm
Julie-CAO: the qam-xen-irqbalance test suite is only listed in SLE-xxx Virt XEN Incidents job group. And the irqbalance test will return if it is not xen host. I did not add the check in module scheduling in case the test suite is added to kvm job group accidently.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: it does not take too much time for each ping, only 0.1s(https://openqa.nue.suse.com/tests/8279264#step/xen_guest_irqbalance/323). if it runs in backgroud, the result will not be shown in screenshots, sometimes the screenshots are helpful for test debugging.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -74,18 +75,21 @@ sub run_test {
...
         # the increased NIC IRQs should be distributed in CPU cores in balance
         my @total_irqs_on_cpu_after_network_download = get_nic_irqs_distribution_from_guest($guest, $nproc);
         my @increased_irqs_on_cpu;
         for (my $cpu_id = 0; $cpu_id < $nproc; $cpu_id++) {
             $increased_irqs_on_cpu[$cpu_id] = $total_irqs_on_cpu_after_network_download[$cpu_id] - $initial_total_irqs_on_cpu[$cpu_id];
-            die("The value of network interrupts from download for CPU" . $cpu_id . " is 0") if $increased_irqs_on_cpu[$cpu_id] == 0;
+            #at least a few interrupts on each cpu core
+            if ($increased_irqs_on_cpu[$cpu_id] < 10) {
+                # julie debug: will change 'record_soft_failure' to 'die' before pr is merge
+                record_soft_failure("IRQ are not balanced as the vif interrupts for CPU" . $cpu_id . " is " . $increased_irqs_on_cpu[$cpu_id]);
```
Julie-CAO, line 89: 3 guests are found IRQ are not balanced in the verfification test, http://openqa.qam.suse.cz/tests/37776. they are sle12sp4HVM, sle15HVM, sle15sp1PV. So here, do we abort test once a problematic guest is found here or continue test? @all @alice-suse

alice-suse: I would suggest to continue test with record_soft_failure, qam reviewers always review every soft-failure every time.

Besides, regarding the irqbalance result on 12sp4/15/15sp1, we need to identify if this is product bug or this is expected behavior. They are older guests, and may have different behavior. Do they ever pass? What happens if you do not bind all interrupts to cpu0 and directly trigger network workload with irqbalance started? Are they balanced?

Julie-CAO: > I would suggest to continue test with record_soft_failure, qam reviewers always review every soft-failure every time.

Ok, I'll keep the current soft_failure.

> Besides, regarding the irqbalance result on 12sp4/15/15sp1, we need to identify if this is product bug or this is expected behavior. They are older guests, and may have different behavior. Do they ever pass? What happens if you do not bind all interrupts to cpu0 and directly trigger network workload with irqbalance started? Are they balanced?

No, they have never passed. other newer guests are irq balanced with latest updated packages only.

if I would not bind all interrupts to cpu0, it most likely all IRQ still were assigned on CPU0, because there was zero interrupts on other CPUs initially. ie. no balance at all regardless what I did.

Julie-CAO: Or because they are LTSS guests, we don't update them to latest LTSS repo?

tonyyuan1:
virt-xml might be easier in automation than xmlstarlet.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: In MU regular test flow, the guests should have been updated to latest LTSS. But here in your validation job, I do not see its parent jobs in "Dependency". Do you skip the step? But this should be able to be validated by checking the guest inside repo and the irqbalance rpm version. I checked again the bug [1178477](https://bugzilla.suse.com/show_bug.cgi?id=1178477), it seems the fix was submitted to 15SP1 LTSS already, but not sure whether it has been RELEASED to customer's LTSS repo. If the guest's installed irqbalance rpm >=   irqbalance-1.4.0, it should be with the fix.

```
SUSE-RU-2021:1192-1: An update that has two recommended fixes can now be installed.

Category: recommended (important)
Bug References: 1178477,1183405
CVE References: 
JIRA References: 
Sources used:
SUSE Linux Enterprise Server 15-SP1-LTSS (src):    irqbalance-1.4.0-7.6.1
```

Julie-CAO: > In MU regular test flow, the guests should have been updated to latest LTSS. But here in your validation job, I do not see its parent jobs in "Dependency". Do you skip the step? 

the parent job is listed in 'dependency', please check http://openqa.qam.suse.cz/tests/37776#dependencies

> But this should be able to be validated by checking the guest inside repo and the irqbalance rpm version. I checked again the bug 1178477, it seems the fix was submitted to 15SP1 LTSS already, but not sure whether it has been RELEASED to customer's LTSS repo. If the guest's installed irqbalance rpm >= irqbalance-1.4.0, it should be with the fix.

in 15sp1 guest, 
irqbalance-1.4.0-7.3.1.x86_64(http://openqa.qam.suse.cz/tests/37776#step/xen_guest_irqbalance/716)
so the fix has not gone into our guest yet.

Julie-CAO: In the latest run http://openqa.qam.suse.cz/tests/38110, only sle15HVM failed due to IRQ assigned on CPU0 only, sle15sp1PV & sle12sp4HVM passed.
irqbalance-1.4.0-7.6.1.x86_64 in 15sp1 now.

alice-suse:
@tonyyuan1  @tbaev  Please help review if the 12sp4 autoyast change is what you want.

tbaev:
` data/autoyast_xen/sles12sp4HVM_PRG.xml` looks good, but `data/autoyast_xen/sles12sp4PV_PRG.xml `, is not fixed. Maybe you can make the same changes to it too.

Julie-CAO:
> virt-xml might be easier in automation than xmlstarlet.

Yes,  change to use "virt-xml $vm_name --edit --vcpus vcpus=4,maxvcpus=4", which is much shorter. Thanks.

Julie-CAO:
> ` data/autoyast_xen/sles12sp4HVM_PRG.xml` looks good, but `data/autoyast_xen/sles12sp4PV_PRG.xml `, is not fixed. Maybe you can make the same changes to it too.

Is sles12sp4PV_PRG.xml same with sles12sp4HVM_PRG.xml? If yes, I can add it in this PR. If not, let Tony add it by himself :-) because I don't know much about MU guest autoyast files.

Julie-CAO:
The verification run for the last commit: http://openqa.qam.suse.cz/tests/38113
If there is no more comment, PR is ready for merge.

tests/virt_autotest/xen_guest_irqbalance.pm
alice-suse: @Julie-CAO If this is final status that the fix for 15GA is not yet in update repo, in the final test code for MU,  will it be better to, at the beginning of this test, add a check for irqbalance version, if below the fix version, skip the test and  add the bug info in the record_soft_fail info?

tonyyuan1:
> > ` data/autoyast_xen/sles12sp4HVM_PRG.xml` looks good, but `data/autoyast_xen/sles12sp4PV_PRG.xml `, is not fixed. Maybe you can make the same changes to it too.
> 
> Is sles12sp4PV_PRG.xml same with sles12sp4HVM_PRG.xml? If yes, I can add it in this PR. If not, let Tony add it by himself :-) because I don't know much about MU guest autoyast files.
 sles12sp4PV_PRG is a little different. I will add it.

tests/virt_autotest/xen_guest_irqbalance.pm
Julie-CAO: Hi @alice-suse , I did not find the package listed in the bug for SLE15. https://bugzilla.suse.com/show_bug.cgi?id=1178477. So does it mean no fix goes into SLE15 update repo?

alice-suse: @Julie-CAO Likely, but better double confirm in the bug.

Julie-CAO: there will be no fix in SLE15 as the Olaf said no customer requests in the bug. So I skip SLE15 guest in the new push, the verification run passed, http://openqa.qam.suse.cz/tests/38164.

tests/virt_autotest/xen_guest_irqbalance.pm
```diff
@@ -28,7 +29,7 @@ sub run_test {
     foreach my $guest (keys %virt_autotest::common::guests) {
 
         #irqbalance test run on sle12sp3+ guest
-        next if guest_is_sle($guest, '<=12-sp2');
+        next if (guest_is_sle($guest, '<=12-sp2') || guest_is_sle($guest, '=15-sp0'));
```
alice-suse, line 32: Can you add a record_soft_fail or record_info for 15sp0 with bug id and reason to skip?

Julie-CAO: Ok, alreay added. please see the new push: http://openqa.qam.suse.cz/tests/38168#step/xen_guest_irqbalance/2584

alice-suse:
LGTM.

PR #14435, opened 2022-03-08
Fix xrdp_server failure on GNOME41

GraceWang571:
The behavior of "restart" on GNOME40+ is different from previous
We need to handle them separately. 

- Related ticket: https://progress.opensuse.org/issues/107305
- Needles: N/A
- Verification run: 
- https://openqa.suse.de/tests/8286827#details


zhaocongjia:
LGTM.

PR #14437, opened 2022-03-08
Use retry command for SUSEConnect --list-extensions

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/107062#note-32

- Verification runs: 

https://openqa.suse.de/tests/8287612
https://openqa.suse.de/tests/8287611
https://openqa.suse.de/tests/8287610

### DIFF
--- a/tests/console/suseconnect_scc.pm
+++ b/tests/console/suseconnect_scc.pm
@@ -9,7 +9,7 @@
 
 use Mojo::Base qw(consoletest);
 use testapi;
-use utils qw(zypper_call script_retry);
+use utils qw(zypper_call script_retry script_output_retry);
 use version_utils qw(is_sle is_jeos is_sle_micro);
 use registration qw(register_addons_cmd verify_scc investigate_log_empty_license);
 
@@ -25,6 +25,7 @@ sub run {
     my $fake_scc = get_var 'SCC_URL', '';
     $cmd .= ' --url ' . $fake_scc if $fake_scc;
     my $retries = 5;    # number of retries to run SUSEConnect commands
+    my $delay = 60;    # time between retries to run SUSEConnect commands
 
 
     $self->select_serial_terminal;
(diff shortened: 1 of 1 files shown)


tests/console/suseconnect_scc.pm
```diff
@@ -35,11 +36,13 @@ sub run {
...
+    script_retry("$cmd", retry => $retries, delay => $delay, timeout => 180);
     # Check available extenstions (only present in sle)
-    assert_script_run q[SUSEConnect --list-extensions];
-    # What has been activated by default
-    assert_script_run q[SUSEConnect --list-extensions | grep -e '\(Activated\)'] if is_sle;
+    my $extensions = script_output_retry("SUSEConnect --list-extensions", retry => $retries, delay => $delay);
+    record_info('Extensions', $extensions);
+
+    # Make sure there is at least 1 module "Activated".
+    die("None of the modules are Activated") if ($extensions !~ m/Activated/ && is_sle);
```
jlausuch, line 45: I just found out this is useles... the command outputs:
```
(Not available) The module/extension is not enabled on your RMT/SMT
(Activated)     The module/extension is activated on your system
```
so it will be always true, not sure why this check was there before....

mloviska (approved):

tests/console/suseconnect_scc.pm
mloviska: You would need to dig to the history. I am not sure, before there was a needle assertion there and as the tty ouput was not always completely seen on the screen it get you a nice overview what was activate.

mloviska: On the second thought there might be a regression introduce when SUSEConnect won't display the "activated" string there. Thus I would not personally remove it.

dzedro (approved):

tests/console/suseconnect_scc.pm
jlausuch: ok, since this PR is about retry only, I will leave the logic it was there before.

jknphy: yes, it is useless due to the legend at the bottom shows what you mentioned, so always there is one hit.
we might need here something like `my $number = () = $extensions =~ /Activated/; die "None of the modules are Activated" if (is_sle() && $number <  2);` wdyt?

jlausuch: Yeah, but looking at the VR https://openqa.suse.de/tests/8287612#step/suseconnect_scc/18,
there are no modules active, which should be also fine, shouldn't it?

foursixnine (approved):

jlausuch:
Merging, feel free to remove the logic to check for `Activated` string.

PR #14438, opened 2022-03-08
Exclude nginx from consoletest for SLED

Zaoliang:
need to unschedule nginx from SLED. It should not be tested on SLED at all.
see https://progress.opensuse.org/issues/106673
verification run:
http://10.162.30.85/tests/4621 (SLES consoletests with nginx)
http://10.162.30.85/tests/4620 (SLED consoletests without nginx)


asdil12 (changes requested):
I think it would be better to use the `is_desktop()` function to check for SLED.

Zaoliang:
okay, thanks. updated now.

asdil12 (approved):

PR #14439, opened 2022-03-08
Fix mrsh_master from typo in the signature of Mojo::Base

b10n1k:
The signature of the `Mojo::Base` needs a comma before the `-signatures` flag.


czerw (approved):

b10n1k:
@martinsmarcelo this will solve https://openqa.suse.de/tests/8284670. Let me know if there is any over file which have been slipped out

PR #14440, opened 2022-03-08
K8s test uses CONTAINER_IMAGES_TO_TEST variable

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/107776
- Verification run: 
- http://copland.arch.suse.de/tests/1013 (without variable)
- http://copland.arch.suse.de/tests/1015 (with variable)


ilausuch:
Not necessary

PR #14451, opened 2022-03-09
[WIP]Repair ssh public key file to avoid invalid format problem

tinawang123:
To make sure the ssh public key file can be used. Sometimes, load
key got invalid format error. We can repair the public key before
load it. In case, the public key cannot be used. Add public RSA 
key to make sure ssh without password works fine

- Related ticket: https://progress.opensuse.org/issues/92332
- Verification run:http://openqa.suse.de/tests/8296533#step/sshxterm/5
  http://openqa.suse.de/tests/8296532#step/sshxterm/4

### DIFF
--- a/tests/x11/sshxterm.pm
+++ b/tests/x11/sshxterm.pm
@@ -23,6 +23,8 @@ sub run {
     my ($self) = @_;
     mouse_hide(1);
     x11_start_program('xterm');
+    # Repair public key to make sure ssh without password works fine
+    enter_cmd("ssh-keygen -f ~/.ssh/id_rsa -y > ~/.ssh/id_rsa.pub");
     enter_cmd("ssh -o StrictHostKeyChecking=no -XC root\@localhost xterm");
     assert_screen "ssh-second-xterm";
     $self->set_standard_prompt();


tests/x11/sshxterm.pm
```diff
@@ -23,6 +23,7 @@ sub run {
     my ($self) = @_;
     mouse_hide(1);
     x11_start_program('xterm');
+    enter_cmd("ssh-keygen -f ~/.ssh/id_rsa -y > ~/.ssh/id_rsa.pub");
```
lemon-suse, line 26: I think you'd better to add comments for the reason of re-create the public key here.

rfan1:
@tinawang123 
I suggest your git commit msg:
Add public RSA key to make sure ssh without password works fine

lemon-suse:
Sometimes, load key got invalid format error. 
- maybe here is a potential bug. Besides, do you confirm if got invalid format error then the test issue reported?
We can repair the public key before
load it. In case, the public key cannot be used.
- I think you can update it as 'We can re-create the public key before load it in case can't use it.

tests/x11/sshxterm.pm
coolgw: Maybe an product issue after migration?

rfan1 (approved):

PR #14434, opened 2022-03-07
Refactor get_container_test_image

grisu48:
Remove duplicate handling of getting the container test images and
create a library function for this. In addition this renames the used variable to
`CONTAINER_IMAGE_TO_TEST` and changes the behaviour to assuming a single
image to be tested.

- Related ticket: https://progress.opensuse.org/issues/107776
- Follow-up of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14400
- Verification run: [AKS](http://duck-norris.qam.suse.de/tests/8342) | [EKS](http://duck-norris.qam.suse.de/tests/8337) | [podman](https://openqa.suse.de/tests/8286612)

PR #14452, opened 2022-03-09
Containers: Fix image URL logic

pdostal:
* Rename and fix `containers::urls::get_container_url_from_var()` (previously `get_urls_from_var()`)
 * Rename `CONTAINER_IMAGES_TO_TEST` to `CONTAINER_IMAGE_TO_TEST`
 * Introduce `containers::urls::get_container_url_to_test()`
   * Usable in test modules where only single image is expected.
   * Not usable in `tests/containers/image.pm` where multiple images of various versions are expected.

- Related ticket: [poo#107776](https://progress.opensuse.org/issues/107776)
- Follow-up of: #14400
- Parallel attempt: #14434 
- Verification run: [EKS](http://pdostal-server.suse.cz/tests/13764), [EKS with specified image](http://pdostal-server.suse.cz/tests/13765), [Docker with specified image](http://pdostal-server.suse.cz/tests/13766), [Docker](http://pdostal-server.suse.cz/tests/13767), [Docker with multiple SLE image versions](http://pdostal-server.suse.cz/tests/13753)

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -20,7 +20,8 @@ use version_utils qw(is_sle is_opensuse is_tumbleweed is_leap is_microos is_sle_
 
 our @EXPORT = qw(
   get_opensuse_registry_prefix
-  get_urls_from_var
+  get_container_image_to_test
+  get_container_url_from_var
   get_suse_container_urls
   get_3rd_party_images
 );
@@ -262,12 +263,12 @@ sub supports_image_arch {
     (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
 
-# Returns a tuple of image urls.
-# If empty, no images available.
-sub get_urls_from_var {
+# Returns an array with the contents of the given variable or undef, if the var is not set
+sub get_container_url_from_var {
     my $var = shift;
-    my @urls = split(/,/, trim(get_var($var)));
-    return (\@urls);
+    my @urls = (trim(get_var($var)));
+    # Return undef if empty string
+    return ($urls[0]) ? (\@urls) : undef;
 }
 
 # Returns a tuple of image urls and their matching released "stable" counterpart.
@@ -292,6 +293,16 @@ sub get_suse_container_urls {
 
     return (\@untested_images, \@released_images);
 }
+
+# Get single image from CONTAINER_IMAGE_TO_TEST or untested image from get_suse_container_urls()
+sub get_container_image_to_test {
+    my $images_to_test;
+    unless ($images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
+        ($images_to_test, undef) = get_suse_container_urls();
+    }
+    return $images_to_test->[0];
+}
+
 
 sub get_3rd_party_images {
     my $ex_reg = get_var('REGISTRY', 'docker.io');
--- a/tests/containers/image.pm
+++ b/tests/containers/image.pm
@@ -13,7 +13,7 @@ use testapi;
 use utils;
 use containers::common;
 use containers::container_images;
-use containers::urls qw(get_suse_container_urls get_urls_from_var);
+use containers::urls qw(get_suse_container_urls get_container_url_from_var);
 
 sub run {
     my ($self, $args) = @_;
@@ -28,8 +28,8 @@ sub run {
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
         my $images_to_test;
-        # Get list of images from CONTAINER_IMAGES_TO_TEST or use the default
-        unless ($images_to_test = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')) {
+        # Get array of single image from CONTAINER_IMAGES_TO_TEST or from get_suse_container_urls()
+        unless ($images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
             my ($untested_images, $released_images) = get_suse_container_urls(version => $version);
             $images_to_test = check_var('CONTAINERS_UNTESTED_IMAGES', '1') ? $untested_images : $released_images;
         }
--- a/tests/containers/push_container_image_to_acr.pm
+++ b/tests/containers/push_container_image_to_acr.pm
@@ -9,7 +9,7 @@
 
 use Mojo::Base 'publiccloud::basetest';
 use testapi;
-use containers::urls qw(get_suse_container_urls get_urls_from_var);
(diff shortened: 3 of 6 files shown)


lib/containers/urls.pm
```diff
@@ -293,6 +294,16 @@ sub get_suse_container_urls {
     return (\@untested_images, \@released_images);
 }
 
+# Get single image from CONTAINER_IMAGE_TO_TEST or untested image from get_suse_container_urls()
+sub get_container_url_to_test {
```
grisu48, line 298:
```suggestion
sub get_container_url_to_test {
sub get_container_image_to_test {
```

Maybe we could align the naming of the function with the naming of the variable? WDYT?

lib/containers/urls.pm
```diff
@@ -262,12 +263,12 @@ sub supports_image_arch {
     (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
 
-# Returns a tuple of image urls.
+# Returns a tuple of single image url.
```
grisu48, line 266:
```suggestion
# Returns an array with the contents of the given variable or undef, if the var is not set
```
Sorry, I'm nitpicking here :slightly_smiling_face: 

AFAIK perl has no tuples, also it does not return empty but undef if the var is not set. I made a suggestion how this can be addressed.

grisu48 (commented):
Good work Pavel! I have two suggestions for improvement but overall this makes more sense now.

ilausuch (approved):
It's a ready good job.

lib/containers/urls.pm
pdostal: Nice, thank you :smile:

b10n1k: more specifically a `ref to array`. one question. @pdostal  Why dont you use perlpod format?

lib/containers/urls.pm
```diff
@@ -262,12 +263,11 @@ sub supports_image_arch {
...
-# Returns a tuple of image urls.
-# If empty, no images available.
-sub get_urls_from_var {
+# Returns an array with the contents of the given variable or undef, if the var is not set
+sub get_container_url_from_var {
     my $var = shift;
-    my @urls = split(/,/, trim(get_var($var)));
-    return (\@urls);
+    my @urls = (trim(get_var($var)));
+    return (length(@urls[0]) > 0) ? (\@urls) : undef;
```
b10n1k, line 270: `s/length(@urls[0])/$urls[0]/` should do the same, and it is more perlist way

lib/containers/urls.pm
```diff
@@ -293,6 +293,16 @@ sub get_suse_container_urls {
     return (\@untested_images, \@released_images);
 }
 
+# Get single image from CONTAINER_IMAGE_TO_TEST or untested image from get_suse_container_urls()
+sub get_container_image_to_test {
+    my $images_to_test;
+    unless ($images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
```
b10n1k, line 299:
```suggestion
    unless (my $images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
```

tests/containers/image.pm
```diff
@@ -28,8 +28,8 @@ sub run {
     my $versions = get_var('CONTAINER_IMAGE_VERSIONS', get_required_var('VERSION'));
     for my $version (split(/,/, $versions)) {
         my $images_to_test;
-        # Get list of images from CONTAINER_IMAGES_TO_TEST or use the default
-        unless ($images_to_test = get_urls_from_var('CONTAINER_IMAGES_TO_TEST')) {
+        # Get single image from CONTAINER_IMAGES_TO_TEST or array from get_suse_container_urls()
```
b10n1k, line 31: i know that this is likely after @grisu48 request but i think it is good idea to point out that this is a ref to array.

b10n1k (changes requested):
Some small improvements if you dont mind

lib/containers/urls.pm
```diff
@@ -293,6 +293,16 @@ sub get_suse_container_urls {
     return (\@untested_images, \@released_images);
 }
 
+# Get single image from CONTAINER_IMAGE_TO_TEST or untested image from get_suse_container_urls()
+sub get_container_image_to_test {
+    my $images_to_test;
```
jlausuch, line 298: Could we do the following?
1) Remove `get_container_url_from_var` method
2) Write this method like this

```
my $image = get_var('CONTAINER_IMAGE_TO_TEST');
return $image if ($image);
($image, undef) = get_suse_container_urls();
return $image->[0]; # or similar, you can use different var if you want
```

3) For image.pm see below

tests/containers/image.pm
```diff
@@ -28,8 +28,8 @@ sub run {
...
+        unless ($images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
```
jlausuch, line 32: Following the previous comment, this would turn into:
```
unless (get_container_image_to_test()) {
```

lib/containers/urls.pm
pdostal: Yes, can you please create a ticket out of this and close?

tests/containers/image.pm
pdostal: No. This does not make sense. Let me please handle this in separate pull request and please close.

lib/containers/urls.pm
pdostal: Done.

lib/containers/urls.pm
pdostal: No, this is not possible as we work with `$images_to_test` either way. I tested your suggestion tho [here](http://pdostal-server.suse.cz/tests/13776) tho.

tests/containers/image.pm
pdostal: Done.

lib/containers/urls.pm
b10n1k: shit man. i missed that reused later. ignore

tests/containers/image.pm
pdostal: Or let me rephrase it: It think that this condition can be avoided completely.

lib/containers/urls.pm
pdostal: I think [this ticket](https://progress.opensuse.org/issues/108058) is suitable.

jlausuch (approved):
Ok to fix the current issue that we are not testing any image.
We have https://progress.opensuse.org/issues/108058 to continue with this topic and have more time to review/suggest.

pdostal:
All the verification runs have been restarted and are still working properly.

b10n1k (approved):

PR #14454, opened 2022-03-09
JeOS: help troubleshoot btrfs_autocompletion issues

jlausuch:
There are some not-so-sporadic issues in this module for JeOS tests
that fail when doing zypper install bash-completion with the error:
System management is locked by the application with pid XYZ...

It's difficult to reproduce the issue as it appears randomly.

Printing the list of processes before running zypper might help
to spot the issue when it happens again.

- Related ticket: https://progress.opensuse.org/issues/108064
- Verification run: https://openqa.suse.de/tests/8297079#step/btrfs_autocompletion/4


mloviska (approved):
It would be nice to upload `zypper.log` in `post_fail_hook` too

PR #14464, opened 2022-03-10
Upload zypper.log in zypper_in post fail hook

asdil12:
http://kazhua.qa.suse.de/tests/92#downloads

### DIFF
--- a/tests/console/zypper_in.pm
+++ b/tests/console/zypper_in.pm
@@ -75,4 +75,12 @@ sub run {
     }
 }
 
+
+sub post_fail_hook {
+    my $self = shift;
+
+    upload_logs "/var/log/zypper.log";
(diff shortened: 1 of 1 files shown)


foursixnine (approved):

tests/console/zypper_in.pm
```diff
@@ -75,4 +75,12 @@ sub run {
     }
 }
 
+
+sub post_fail_hook {
+    my $self = shift;
+
+    upload_logs "/var/log/zypper.log";
```
foursixnine, line 82: maybe also add ` /var/log/history`

asdil12: not sure what this should contain - but it doesn't exist on my local tumbleweed system

PR #14466, opened 2022-03-10
Containers: Implement get_image_url()

pdostal:
Implement `get_image_url()` instead of `get_container_image_to_test()`, `get_container_url_from_var()` and `get_suse_container_urls()`. We have a list of SUSE and openSUSE released and unreleased container images but we always need only one image url to work with.

- Related ticket: [poo#108058](https://progress.opensuse.org/issues/108058)
- Verification run: In further comments

### DIFF
--- a/lib/containers/urls.pm
+++ b/lib/containers/urls.pm
@@ -14,16 +14,13 @@ use Exporter;
 use strict;
 use warnings;
 use testapi;
-use Mojo::Util 'trim';
 use Utils::Architectures;
 use version_utils qw(is_sle is_opensuse is_tumbleweed is_leap is_microos is_sle_micro is_released);
 
 our @EXPORT = qw(
   get_opensuse_registry_prefix
-  get_container_image_to_test
-  get_container_url_from_var
-  get_suse_container_urls
   get_3rd_party_images
+  get_image_uri
 );
 
 # Returns a string which should be prepended to every pull from registry.opensuse.org.
@@ -51,7 +48,7 @@ sub get_opensuse_registry_prefix {
     }
 }
 
-our %images_uri = (
+our %images_list = (
     sle => {
         '12-SP3' => {
             released => sub { 'registry.suse.com/suse/sles12sp3' },
@@ -260,49 +257,8 @@ our %images_uri = (
 
 sub supports_image_arch {
     my ($distri, $version, $arch) = @_;
-    (grep { $_ eq $arch } @{$images_uri{$distri}{$version}{available_arch}}) ? 1 : 0;
+    (grep { $_ eq $arch } @{$images_list{$distri}{$version}{available_arch}}) ? 1 : 0;
 }
-
-# Returns an array with the contents of the given variable or undef, if the var is not set
-sub get_container_url_from_var {
-    my $var = shift;
-    my @urls = (trim(get_var($var)));
-    # Return undef if empty string
-    return ($urls[0]) ? (\@urls) : undef;
-}
-
-# Returns a tuple of image urls and their matching released "stable" counterpart.
-# If empty, no images available.
-sub get_suse_container_urls {
-    my %args = (
-        version => get_required_var('VERSION'),
-        arch => get_required_var('ARCH'),
-        distri => get_required_var('DISTRI'),
-        @_
-    );
-    my @untested_images = ();
-    my @released_images = ();
-
-    $args{version} =~ s/^Staging:(?<letter>.)$/Tumbleweed/ if is_tumbleweed || is_microos("Tumbleweed");
-    if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
-        push @untested_images, $images_uri{$args{distri}}{$args{version}}{totest}->($args{arch});
-        push @released_images, $images_uri{$args{distri}}{$args{version}}{released}->($args{arch});
-    } else {
-        die("Unknown combination of distro/arch.");
-    }
-
-    return (\@untested_images, \@released_images);
-}
-
-# Get single image from CONTAINER_IMAGE_TO_TEST or untested image from get_suse_container_urls()
-sub get_container_image_to_test {
-    my $images_to_test;
-    unless ($images_to_test = get_container_url_from_var('CONTAINER_IMAGE_TO_TEST')) {
-        ($images_to_test, undef) = get_suse_container_urls();
-    }
-    return $images_to_test->[0];
-}
-
 
 sub get_3rd_party_images {
     my $ex_reg = get_var('REGISTRY', 'docker.io');
@@ -340,6 +296,31 @@ sub get_3rd_party_images {
     ) unless (is_aarch64 || check_var('PUBLIC_CLOUD_ARCH', 'arm64'));
 
     return (\@images);
+}
+
+# Return URI of the container image to be tested. It will:
+# - Be configured by the CONTAINER_IMAGE_TO_TEST variable.
+# - Be configured by the distri, version and arch parameter.
+# - Be determined by the running OS.
+# - Die otherwise.
+sub get_image_uri {
+    my (%args) = @_;
+    $args{version} //= get_required_var('VERSION');
+    $args{arch} //= get_required_var('ARCH');
+    $args{distri} //= get_required_var('DISTRI');
+    $args{released} //= !get_var('CONTAINERS_UNTESTED_IMAGES');
+
+    $args{version} =~ s/^Staging:(?<letter>.)$/Tumbleweed/ if is_tumbleweed || is_microos("Tumbleweed");
+
+    my $url = get_var('CONTAINER_IMAGE_TO_TEST');
+    return $url if ($url);
+
+    my $type = $args{released} ? 'released' : 'totest';
+    if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
+        return $images_list{$args{distri}}{$args{version}}{$type}->($args{arch});
+    }
+
+    die "Cannot find container image for ($args{distri},$args{version},$args{arch}) or missing CONTAINER_IMAGE_TO_TEST variable.";
 }
 
 1;
--- a/tests/containers/container_diff.pm
+++ b/tests/containers/container_diff.pm
@@ -12,7 +12,7 @@ use testapi;
 use utils;
 use containers::common;
 use containers::container_images;
-use containers::urls 'get_suse_container_urls';
+use containers::urls 'get_image_uri';
 
 sub run {
     my ($self) = @_;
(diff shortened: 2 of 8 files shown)


lib/containers/urls.pm
```diff
@@ -342,4 +299,29 @@ sub get_3rd_party_images {
     return (\@images);
 }
 
+# Return url of specified image as a string.
+# Specify released => 0 for unreleased image.
+# Die when nothing to return.
+sub get_container_url {
```
jlausuch, line 305: I would prefer to have a name that implies an `image` and not a `container`.
e.g. `get_image_url` or `get_container_image_url`

jlausuch:
Looks like it goes into the right direction. I left some suggestions. 
I guess you still have in mind to modify all the tests using `get_suse_container_urls` by this new method.

tests/containers/image.pm
```diff
@@ -27,20 +27,13 @@ sub run {
...
-        }
-
-        for my $iname (@{$images_to_test}) {
-            record_info "IMAGE", "Testing image: $iname";
-            test_container_image(image => $iname, runtime => $engine);
-            test_rpm_db_backend(image => $iname, runtime => $engine);
-            my $beta = $version eq get_var('VERSION') ? get_var('BETA', 0) : 0;
-            test_opensuse_based_image(image => $iname, runtime => $engine, version => $version, beta => $beta) unless ($iname =~ /bci/);
-        }
+        my $image = get_image_url();
```
b10n1k, line 30: is this module something which is scheduled in both `released`, `unreleased`? With current approach you will have always `released=1`. i think this breaks for instance https://openqa.suse.de/tests/8312796#step/image_docker/46

lib/containers/urls.pm
```diff
@@ -342,4 +299,28 @@ sub get_3rd_party_images {
...
+
+    $args{version} =~ s/^Staging:(?<letter>.)$/Tumbleweed/ if is_tumbleweed || is_microos("Tumbleweed");
+
+    my $url = trim(get_var('CONTAINER_IMAGE_TO_TEST'));
+    return $url unless $url;
+
+    if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
+        return $images_list{$args{distri}}{$args{version}}{released}->($args{arch}) if ($args{released});
+        return $images_list{$args{distri}}{$args{version}}{totest}->($args{arch});
+    }
```
b10n1k, line 321:
```suggestion
   my $repo = ($args{released}) ? 'released' : totest;
   return $images_list{$args{distri}}{$args{version}}{$repo}->($args{arch});  if supports_image_arch($args{distri}, $args{version}, $args{arch});
```

b10n1k (changes requested):
i think `$args{released}` will cause some issues in the modules which is used in both _released_ and _totest_. One option is to make use of the already existing variables `CONTAINER_IMAGE_TYPE` or `CONTAINERS_UNTESTED_IMAGES` i guess

lib/containers/urls.pm
pdostal: Sorry, I will not do that. It just decreases readability as the lines would be too long.

tests/containers/image.pm
pdostal: Good catch, thank you :smile:

tests/containers/container_diff.pm
```diff
@@ -21,16 +21,15 @@ sub run {
 
     zypper_call("install container-diff") if (script_run("which container-diff") != 0);
 
-    my ($testing_images, $released_images) = get_suse_container_urls();
+    my $unreleased_image = get_image_url(unreleased => 0);
```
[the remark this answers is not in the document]
pdostal: Good catch, thank you :smile:

lib/containers/urls.pm
b10n1k: Sorry i reopen it for further discussion. its a small thing but how does it decreases readability? the main issue here is that you have pretty much the same return twice and i cant see any good reason.

pdostal: Readability is the reason.

pdostal: I am not sure I understand. Do you have another suggestion?

b10n1k: it depends what do you consider /readability/

jlausuch: Agree, that 1 liner is nice but difficult to understand at a first glance...  maybe... 
```
my $type = $args{released} ? 'released' : 'totest';
if supports_image_arch($args{distri}, $args{version}, $args{arch}) {
    return $images_list{$args{distri}}{$args{version}}{$type}->($args{arch})
}
```

jlausuch:
> i think `$args{released}` will cause some issues in the modules which is used in both _released_ and _totest_. One option is to make use of the already existing variables `CONTAINER_IMAGE_TYPE` or `CONTAINERS_UNTESTED_IMAGES` i guess

Exactly, `CONTAINER_IMAGE_TYPE` doesn't exist. `CONTAINERS_UNTESTED_IMAGES=1` means that we should use the `totest` flag.

lib/containers/urls.pm
```diff
@@ -342,4 +299,27 @@ sub get_3rd_party_images {
...
 }
 
+# Return url of specified image as a string. Die when nothing to return.
+# Specify released => 0 or CONTAINERS_UNTESTED_IMAGES=1 for unreleased image.
+sub get_image_url {
+    my (%args) = @_;
+    $args{version} //= get_required_var('VERSION');
+    $args{arch} //= get_required_var('ARCH');
+    $args{distri} //= get_required_var('DISTRI');
+    $args{released} //= get_var('CONTAINERS_UNTESTED_IMAGES') ? 0 : 1;
```
jlausuch, line 309:
```suggestion
    $args{released} //= ! get_var('CONTAINERS_UNTESTED_IMAGES');
```

lib/containers/urls.pm
```diff
@@ -342,4 +299,27 @@ sub get_3rd_party_images {
...
+
+    my $url = trim(get_var('CONTAINER_IMAGE_TO_TEST'));
+    return $url if ($url);
+
+    if (supports_image_arch($args{distri}, $args{version}, $args{arch})) {
+        return $images_list{$args{distri}}{$args{version}}{released}->($args{arch}) if ($args{released});
+        return $images_list{$args{distri}}{$args{version}}{totest}->($args{arch});
+    }
+
+    ($args{die}) ? die("Unknown combination of distro/arch or missing CONTAINER_IMAGE_TO_TEST variable.") : return undef;
```
jlausuch, line 322: It's time to change this error message. To me, when I hit this error in openQA, it was difficult to understand why and I needed to dig into the code. So, for someone who is not used to this code, it's a bit tricky to understand. 

Also, ternary operators are awesome, but when the line is too long, it makes the code a bit unreadable.
So, either you use if-else clause, or store the message in a variable.
Suggestion:
```
my $error_msg = "Cannot find container image for $args{distri} $args{version} $args{arch} or missing CONTAINER_IMAGE_TO_TEST variable";

($args{die}) ? die($error_msg ) : return undef;
```

lib/containers/urls.pm
```diff
@@ -342,4 +299,27 @@ sub get_3rd_party_images {
...
+    my (%args) = @_;
+    $args{version} //= get_required_var('VERSION');
+    $args{arch} //= get_required_var('ARCH');
+    $args{distri} //= get_required_var('DISTRI');
+    $args{released} //= get_var('CONTAINERS_UNTESTED_IMAGES') ? 0 : 1;
+    $args{die} //= 1;    # 0 => do not die
+
+    $args{version} =~ s/^Staging:(?<letter>.)$/Tumbleweed/ if is_tumbleweed || is_microos("Tumbleweed");
+
+    my $url = trim(get_var('CONTAINER_IMAGE_TO_TEST'));
```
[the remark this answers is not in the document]
pdostal: The `trim()` is there to trim spaces at the beginning or end of the variable.
E.g. if you specify `CONTAINER_IMAGE_TO_TEST="registry.suse.de/sle:15.3 "` it will *trim* the last space.

This was suggested by @ilausuch - Ivan, can you help here? :smile:

lib/containers/urls.pm
pdostal: I like this too, thank you @jlausuch !

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Spaces don't harm. But the trim will just simply fail if the variable is empty, which is wrong approach.
If we want to do checks, I would do it in the right way, using `URI` package from perl, e.g. https://stackoverflow.com/questions/2599047/how-can-i-validate-a-website-url-in-perl
But that's optional, that can be improved later, I would simple go for the code I suggested.

b10n1k (approved):

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: I see your point. But I wonder how is then possible that my example worker in all the verification runs both with and without this variable.

Please re-check.

lib/containers/urls.pm
```diff
@@ -342,4 +298,26 @@ sub get_3rd_party_images {
     return (\@images);
 }
 
+# Return url of specified image as a string. Die when nothing to return.
```
grisu48, line 301:
```suggestion
# Return URI of the container image to be tested. This can be either configured via the CONTAINER_IMAGE_TO_TEST variable or it will try to determinte the same image as the running OS. Dies if the image could not be determined
```

WDYT?

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: Yes. I removed the `trim` - is it resolved?

tests/containers/container_diff.pm
```diff
@@ -21,16 +21,15 @@ sub run {
...
-        my $image_file = $testing_images->[$i] =~ s/\/|:/-/gr;
-        my $container_diff_results = "/tmp/container-diff-$image_file.txt";
-        assert_script_run("docker pull $testing_images->[$i]", 360);
-        assert_script_run("docker pull $released_images->[$i]", 360);
-        assert_script_run("container-diff diff daemon://$testing_images->[$i] daemon://$released_images->[$i] --type=rpm --type=file --type=size > $container_diff_results", 300);
-        upload_logs("$container_diff_results");
-    }
+    my $image_file = $unreleased_image =~ s/\/|:/-/gr;
+    my $container_diff_results = "/tmp/container-diff-$image_file.txt";
+    assert_script_run("docker pull $unreleased_image", 360);
```
grisu48, line 29: `docker pull` smells like a good candidate for `script_retry` ....

```suggestion
    script_retry("docker pull $unreleased_image", timeout => 360, retries => 3, delay => 120);
```

Same for the next line

pdostal:
After I get some more approvals I'd like to do more verification runs.

grisu48 (commented):
Good job Pavel! I added some suggestions, but feel free to ignore them if not applicable.

lib/containers/urls.pm
jlausuch: There are 3 cases:

- Provide `CONTAINER_IMAGE_TO_TEST` variable with URI
- Give it as input parameter `$args{version}` which is what we do to test our big matrix.
- Determine the same image as the running OS.

tests/containers/container_diff.pm
pdostal: Yes, let's leave this for separate pull request please.

lib/containers/urls.pm
pdostal:
```
# Return URI of the container image to be tested. This will be:
# - Configured by the CONTAINER_IMAGE_TO_TEST variable.
# - Configured by the distri, version and arch parameter.
# - Determined by the running OS.
# - Terminated otherwise.
```

lib/containers/urls.pm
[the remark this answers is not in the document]
pdostal: I implemented Ivan's solution, thank you!
```
$url = get_var('CONTAINER_IMAGE_TO_TEST');
return trim($url) if ($url)
```

lib/containers/urls.pm
pdostal: Please re-check.

lib/containers/urls.pm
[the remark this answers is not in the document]
jlausuch: Sorry for insisting, but I still don't understand why we need trim here :) 
Just to remove any spaces in the variable is not a good reason, we would need to do this for every variable, which doesn't make sense.

grisu48 (approved):
LGTM

jlausuch (commented):
Looks very good 👍 
Can you please run more VRs?

tests/containers/rootless_podman.pm
```diff
@@ -58,8 +57,6 @@ sub run {
     }
 
     my $image = 'registry.opensuse.org/opensuse/tumbleweed:latest';
-
-    my ($untested_images, $released_images) = get_suse_container_urls();
```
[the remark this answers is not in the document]
pdostal: :smile: thank you

pdostal:
Verification runs: [15-SP3 on AKS](http://pdostal-server.suse.cz/tests/13876), [15-SP3-docker](http://pdostal-server.suse.cz/tests/13879), [15-SP3-podman](http://pdostal-server.suse.cz/tests/13880), [Tumbleweed-podman](http://pdostal-server.suse.cz/tests/13875)

ilausuch (approved):

ilausuch:
Mice work!

PR #14468, opened 2022-03-11
Fix connection reset by peer

tonyyuan1:
When installing some Guest VMs via PXE at the same time some installer got error "connection reset by peer". This will fix the issue. See more detail in ticket.

- Related ticket: https://progress.opensuse.org/issues/101256
- Needles: no
- Verification run: 
- sles12sp5: http://openqa.qam.suse.cz/tests/38105
- sles15sp2: http://openqa.qam.suse.cz/tests/38106
- sles15sp3: http://openqa.qam.suse.cz/tests/38085

### DIFF
--- a/tests/virtualization/universal/prepare_guests.pm
+++ b/tests/virtualization/universal/prepare_guests.pm
@@ -48,6 +48,8 @@ sub run {
 
     # Disable bash monitoring, so the output of completed background jobs doesn't confuse openQA
     script_run("set +m");
+    # Fix poo#101256, connection reset by peer
+    assert_script_run 'iptables -I INPUT 1 -m conntrack --ctstate INVALID -j DROP';
 
     # Install or import defined guests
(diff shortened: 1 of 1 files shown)


tests/virtualization/universal/prepare_guests.pm
```diff
@@ -48,6 +48,8 @@ sub run {
 
     # Disable bash monitoring, so the output of completed background jobs doesn't confuse openQA
     script_run("set +m");
+    # Fix poo#101256, connection reset by peer
+    assert_script_run 'iptables -I INPUT 1 -m conntrack --ctstate INVALID -j DROP';
```
alice-suse, line 52: This line drops INVALID packets, right? To be honest, I would be very surprised to see this can solve the problem. But I am not expert on this. Since you are so confident that this can solve the problem, let's see validation jobs when the network problem reproduces again.

PR #14470, opened 2022-03-11
Fix the failure when installing rmt-server success

chcao:
The added workaround for bsc#1195759 has a problem that when a 
case installing rmt-server package successfully, the test would go 'else'
and die with 'zypper in rmt-server failed'. Our case is using the dev
version of rmt-server, which is higher than released version. The soft
failure bsc#1195759 is not existing on our case, but it is die. This
ticket is going to fix this problem.

- Related ticket: https://progress.opensuse.org/issues/107914
- Needles: N/A
- Verification run: 
migration team case:
https://openqa.nue.suse.com/tests/8381157#step/rmt_feature/15 (this case has no problem with installing rmt-server)
https://openqa.nue.suse.com/tests/8381159#step/rmt_import/9 (this case has soft failure bsc#1195759)
qam team case:
https://openqa.nue.suse.com/tests/8381160#step/rmt_feature/16 (this case has soft failure bsc#1195759)

### DIFF
--- a/lib/repo_tools.pm
+++ b/lib/repo_tools.pm
@@ -251,15 +251,10 @@ sub rmt_wizard {
     my $setup_console = current_console();
 
     # install RMT and mariadb
-    eval {
-        zypper_call 'in rmt-server', log => 'zypper.log';
-    };
-    if ($@ && script_run('grep -E "rmt-server-config.*scriptlet failed" /tmp/zypper.log') == 0) {
+    my $ret = zypper_call('in rmt-server', exitcode => [0, 107], log => 'zypper.log');
+    if (($ret == 107) && (script_run('grep -E "rmt-server-config.*scriptlet failed" /tmp/zypper.log') == 0)) {
         record_soft_failure 'bsc#1195759';
         zypper_call 'in rmt-server';
-    }
-    else {
-        die 'zypper in rmt-server failed';
     }
     zypper_call 'in mariadb';
 


lib/repo_tools.pm
```diff
@@ -258,9 +258,6 @@ sub rmt_wizard {
         record_soft_failure 'bsc#1195759';
         zypper_call 'in rmt-server';
     }
-    else {
```
lemon-suse, line 261: I think you can't remove it directly. This workaround is for rmt case without DEV_PATH, for DEV_PATH you can zypper in rmt_server instead. So there should be two branches.

chcao:
@dzedro Would you please help to review this pr? Thanks.

lib/repo_tools.pm
dzedro: I would not remove the else because then other failure than `scriptlet fail` will pass.
But I don't know how to suppress the notification. I hope the bug is fixed soon and the workaround can be removed.

dzedro (approved):

lib/repo_tools.pm
lemon-suse: In fact, I think there are two status for different zypper version, on DEV_PATH there is no bug found, without DEV_PATH, it will fail with return code of 107 and waning message in log.
So why not use the return value of zypper_call as the condition of the workaround such as 
if ($ret == 107)
    softfailure

tinawang123 (approved):
LGTM

lib/repo_tools.pm
coolgw: How about add extra elif (get_var("DEV_PATH")) instead of remove the else? If i understand correctly.

dzedro: Return code 107 could be used.

chcao: Using return code 107, please help to review, thanks.

lemon-suse (approved):

lib/repo_tools.pm
lemon-suse: Please run VRs with latest code.

PR #14441, opened 2022-03-08
Adapt `mm_network::setup_static_mm_network()` for Network Manager

pdostal:
Adapt `mm_network::setup_static_mm_network()` for Network Manager

- Tickets: [poo#105295](https://progress.opensuse.org/issues/105295), [bsc#1196606](https://bugzilla.opensuse.org/show_bug.cgi?id=1196606)
- Verification run: [openvpn-client@SLE15-SP3](http://pdostal-server.suse.cz/tests/13806), [wireguard-client@Tumbleweed](http://pdostal-server.suse.cz/tests/13804)

PR #14471, opened 2022-03-11
Fix configure_static_dns call as the order of parameters changed

pdostal:
I checked `configure_static_dns()` and `configure_static_ip()` and this is the only case where is the `silent` parameter misplaced.

- Previous request: #14441
- Verification run: [wicked](https://openqa.suse.de/tests/8313033)

### DIFF
--- a/lib/network_utils.pm
+++ b/lib/network_utils.pm
@@ -37,7 +37,7 @@ sub setup_static_network {
     $args{ip} //= '10.0.2.15';
     $args{gw} //= testapi::host_ip();
     $args{silent} //= 0;
-    configure_static_dns(get_host_resolv_conf(), $args{silent});
+    configure_static_dns(get_host_resolv_conf(), undef, $args{silent});
     assert_script_run('echo default ' . $args{gw} . ' - - > /etc/sysconfig/network/routes');
     my $iface = iface();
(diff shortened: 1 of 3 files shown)


lib/network_utils.pm
```diff
@@ -37,7 +37,7 @@ sub setup_static_network {
     $args{ip} //= '10.0.2.15';
     $args{gw} //= testapi::host_ip();
     $args{silent} //= 0;
-    configure_static_dns(get_host_resolv_conf(), $args{silent});
+    configure_static_dns(get_host_resolv_conf(), undef, $args{silent});
```
jlausuch, line 40: Just for the record why the previous PR was reverted, is because it was missing this change.

jlausuch (approved):

DimStar77:
https://openqa.opensuse.org/tests/2240838#step/before_test/20

?!?

DimStar77:
Also: https://openqa.opensuse.org/tests/2240848#step/xrdp_server/27

jlausuch:
This was actually a PR that was merged before and we reverted since it caused some regression. We came up with this PR but I see it also causes some other regression. I'm gonna revert again and we'll work on it on Monday making sure that that regression is covered. Sorry for the inconvenience.

PR #14473, opened 2022-03-11
Add support for icelake in the mitigations function test.

museQi:
Change the code of taa.pm and mitigations.pm to support the icelake machine in the mitigaitions function test.

- Verification run: http://openqa.qa2.suse.asia/group_overview/44

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1269,8 +1269,7 @@ sub ensure_serialdev_permissions {
     else {
         # when serial getty is started, it changes the group of serialdev from dialout to tty (but doesn't change it back when stopped)
         # let's make sure that both will work
-        # based on bsc#1195620, let's restore file permission to '620'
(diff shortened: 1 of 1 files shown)


tests/cpu_bugs/mitigations.pm
```diff
@@ -43,6 +43,12 @@ my %mitigations_list =
     },
     cmdline => ["auto,nosmt", "off", "auto"],
   );
+#Add icelake information
```
alice-suse, line 46: Suggested comment format is like this "# Your comment"

alice-suse:
Suggest to involve @XGWang0  to review since he is familiar with your code requirement.

Besides, regarding the commits, there are two, and one is the merge commit. Generally we do not do 'git merge', but to use 'git rebase master' to let your commit do on top of latest master code, and for small fix, it is better to be done in 1 commit.

alice-suse:
Suggest to involve @XGWang0  to review since he is familiar with your code requirement.

Besides, regarding the commits, there are two, and one is the merge commit. Generally we do not do 'git merge', but to use 'git rebase master' to let your commit do on top of latest master code, and for small fix, it is better to be done in 1 commit.

PR #14475, opened 2022-03-11
xfstests: fix boot fail when enable kdump in spvm

SeroSun:
enable_kdump run in ppc64le machin fails in booting (e.g. https://openqa.suse.de/tests/8306291#step/enable_kdump/36). This PR is to fix this issue.

- Verification run: https://openqa.suse.de/t8314983


lansuse:
LGTM

PR #14481, opened 2022-03-14
Wait for reset-rollback-registration to finish

mloviska:
The script is called on first boot after a snapper rollback in order to
verify that after a rollback the correct products are registered at SCC.

Tests were proceeding further after rollback and that can lead to
problems with lock described in [Test fails in btrfs_autocompletion -
System management is locked by the application with pid 1658
(zypper).](https://progress.opensuse.org/issues/108064).

- Verification runs: 
  * [sle-12-SP5-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220314-1-jeos-filesystem@uefi-virtio-vga](http://kepler.suse.cz/tests/14501#live)
  * [sle-12-SP5-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220314-1-jeos-filesystem@uefi-virtio-vga](http://kepler.suse.cz/tests/14494#step/snapper_jeos_cli/105)

### DIFF
--- a/tests/console/snapper_jeos_cli.pm
+++ b/tests/console/snapper_jeos_cli.pm
@@ -15,8 +15,6 @@ use strict;
 use warnings;
 use power_action_utils qw(power_action);
 
-
-
 sub check_package
 {
     my ($not_installed, $pkgname, $check_path) = @_;
@@ -47,6 +45,15 @@ sub rollback_and_reboot {
     }
     select_console('root-console');
     assert_script_run("snapper list");
+    # check whether SUSEConnect --rollback is running executed by rollback-reset-registration
+    # this might cause a system management lock by zypper
+    for (my $runs = 0; $runs < 5; $runs++) {
+        if (script_run('test -f /var/lib/rollback/check-registration') == 1) {
+            return 1;
+        }
+        record_info('ps', script_output('ps -ef'));
+        sleep 20;
+    }
 }
 
 sub run {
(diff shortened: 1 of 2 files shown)


tests/console/snapper_jeos_cli.pm
```diff
@@ -47,6 +45,15 @@ sub rollback_and_reboot {
     }
     select_console('root-console');
     assert_script_run("snapper list");
+    # check whether SUSEConnect --rollback is running executed by rollback-reset-registration
+    # this might cause a system management lock by zypper
+    for (my $runs = 0; $runs < 5; $runs++) {
+        if (script_run('test -f /var/lib/rollback/check-registration') == 1) {
```
jlausuch, line 51: Will this be enough? Shouldn't we also check for zypper process?
In your VR, I see zypper running  in the ps output:
http://kepler.suse.cz/tests/14494#step/snapper_jeos_cli/104

jlausuch: Ah ok, got it. If that file doesn't exist, the rollback is supposed to be completed and no more actions will be performed.

mloviska: https://github.com/openSUSE/rollback-helper/blob/master/sbin/rollback-reset-registration#L30 

I think it should be, as the file only exists while `SUSEConnect --rollback` hasn't successfully finished.

PR #14474, opened 2022-03-11
containers: restart firewalld after installing docker

DimStar77:
- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1196801
- Needles: N/A
- Verification runs:
* [jeos-container_host@64bit_virtio-2G](https://openqa.opensuse.org/t2239248)
* [jeos-container_image@64bit_virtio-2G](https://openqa.opensuse.org/t2239250)

PR #14483, opened 2022-03-14
containers: restart firewall if enabled before

jlausuch:
This fixes a regression caused by #14474

VRs: 
[opensuse-Tumbleweed-JeOS-for-kvm-and-xen-x86_64](https://openqa.opensuse.org/tests/2243604)
[opensuse-Tumbleweed-JeOS-for-kvm-and-xen-x86_64](https://openqa.opensuse.org/tests/2243605)
[sle-15-SP3-AZURE-BYOS-Updates](https://openqa.suse.de/tests/8325426)

### DIFF
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -86,8 +86,10 @@ sub install_docker_when_needed {
                 # docker package can be installed
                 zypper_call('in docker', timeout => 300);
 
-                # Restart firewalld if enabled before. Ensure docker can properly interact
-                systemctl 'try-restart firewalld';
+                # Restart firewalld if enabled before. Ensure docker can properly interact (boo#1196801)
+                if (script_run('systemctl is-active firewalld') == 0) {
+                    systemctl 'try-restart firewalld';
+                }
 
                 remove_suseconnect_product('SLES-LTSS') if ($ltss_needed);
             }


lib/containers/common.pm
```diff
@@ -86,8 +86,10 @@ sub install_docker_when_needed {
                 # docker package can be installed
                 zypper_call('in docker', timeout => 300);
 
-                # Restart firewalld if enabled before. Ensure docker can properly interact
-                systemctl 'try-restart firewalld';
+                # Restart firewalld if enabled before. Ensure docker can properly interact (boo#1196801)
+                if (script_run('systemctl is-active firewalld') == 0) {
```
dzedro, line 90: This will not work if firewalld does not exist, rather check if firewalld is present.

jlausuch: Why not? It's a `script_run`, not `assert_script_run`.
See also https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/services/sshd.pm#L55

dzedro: True, it will work.
It's just weird, it is failing now because there is no firewalld. https://openqa.suse.de/tests/8325047#step/docker_runc/180

jlausuch:
More regressions today https://openqa.suse.de/tests/8326717#step/docker/49
I am merging this as I expect more failed tests.

PR #14488, opened 2022-03-15
[WIP] Merge docker & podman libs into container_runtime

ilausuch:
Docker and podman container runtime libraries
can be merged in only one common library helping
to simplify the maintenance.

- Related ticket: https://progress.opensuse.org/issues/101454
- Verification run: 
- http://copland.arch.suse.de/tests/1063
- http://copland.arch.suse.de/tests/1064


jlausuch (changes requested):
The idea was to merge together:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/containers/docker.pm
and
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/containers/podman.pm

which share the same test logic and code.

PR #14489, opened 2022-03-15
Do whole patch process separately for each patch

dzedro:
- Related ticket: https://progress.opensuse.org/issues/95863
- Verification run:
https://dzedro.suse.cz/tests/21327 Conflicts
https://dzedro.suse.cz/tests/21328 New package
WIP Multiple patches


foursixnine:
looks like tidy needs to be ran `make tidy-full` I think? on the other hand, I wonder if for a separate PR we could write unit tests, so that we have a clear definition of what's being tested...

tests/qam-updinstall/update_install.pm
```diff
@@ -92,15 +97,17 @@ sub run {
     my ($self) = @_;
     my $incident_id = get_required_var('INCIDENT_ID');
     my $repos = get_required_var('INCIDENT_REPO');
+    my %installable;    #Binaries already released that can already be installed.
+    my @new_binaries;    #Binaries introduced by the update that will be installed after the repos are added.
```
dzedro, line 101: Tidy was failing because of this, which is right now wrong comment, I think ...
I spit on tidy and hope I will not have to use it ever again :pray:

dzedro:
This code will create prepare.log zypper.log for each patch, at end it will merge them into one which is expected.
@a-kpappas @baierjan Is this OK ? I don't know how the testsuite template parse the log.

baierjan:
> This code will create prepare.log zypper.log for each patch, at end it will merge them into one which is expected. @a-kpappas @baierjan Is this OK ? I don't know how the testsuite template parse the log.

Currently, template generator is looking for `/tests/<id>/file/update_install-zypper.log`

dzedro:
> > This code will create prepare.log zypper.log for each patch, at end it will merge them into one which is expected. @a-kpappas @baierjan Is this OK ? I don't know how the testsuite template parse the log.
> 
> Currently, template generator is looking for `/tests/<id>/file/update_install-zypper.log`

Logs will be there, but I'm not sure if multiple installations instead one installation will work fine
 e.g. here are two patches thus two installations in update_install-zypper.log, is that ok ?
https://openqa.suse.de/tests/9031785/file/update_install-zypper.log

baierjan:
> Logs will be there, but I'm not sure if multiple installations instead of one installation will work fine ? e.g. here are two patches thus two installations in update_install-zypper.log, is that ok ? https://openqa.suse.de/tests/9031785/file/update_install-zypper.log

As far as I know, there is no parsing involved. The generator just downloads and stores the log for (manual) evaluation.

dzedro:
> > Logs will be there, but I'm not sure if multiple installations instead of one installation will work fine ? e.g. here are two patches thus two installations in update_install-zypper.log, is that ok ? https://openqa.suse.de/tests/9031785/file/update_install-zypper.log
> 
> As far as I know, there is no parsing involved. The generator just downloads and stores the log for (manual) evaluation.

Ah ok, I thought it's parsing something from the log, then good, Thank you!

foursixnine (approved):
for the most part looks good to me

foursixnine:
@dzedro I'll add the do not merge until you call it done then, at end of week I'll ping again :)

foursixnine (approved):

tests/qam-updinstall/update_install.pm
```diff
@@ -122,117 +140,146 @@ sub run {
...
-    my @installable_l3 = grep { $bins{$_}->{supportstatus} eq 'l3' } keys %installable;
-    # If not all l3 released binaries are in the conflict binaries, fail.
-    print "\nInstallable L3 @installable_l3\n";
-    for my $package (@installable_l3) {
-        my $hit = first { $package eq $_ } @conflict_names;
-        unless ($hit) {
-            record_info "Error", "$package is l3 but does not exist in the patch. The update may have been misconfigured";
+        # Install released version of installable binaries.
+        # Make sure on SLE 15+ zyppper 1.14+ with '--force-resolution --solver-focus Update' patched binaries are installed
+        my $solver_focus = $zypper_version >= 14 ? '--force-resolution --solver-focus Update ' : '';
```
b10n1k, line 216: Can [check_version](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/version_utils.pm#L178) be used here? And using a regex to match 1.14 instead of merely 14?

dzedro: I don't know. This works

dzedro:
Cloning conflicting job with `WORKAROUND_PREINSTALL_CONFLICT=1` to workaround the conflict.
https://openqa.suse.de/tests/overview?build=%3A25111%3Agcc10&distri=sle&groupid=282

Added rpm-ndb conflict
https://openqa.suse.de/tests/overview?build=%3A25112%3Arpm&distri=sle&groupid=439

dzedro:
:game_die:

foursixnine:
> game_die

PR #14494, opened 2022-03-16
console/consoletest_finish: Close PackageKit error notification

Vogtinator:
`system_prepare` stops packagekitd, which causes an error notification. This
needs to be closed to not be in the way of needle matches.

- Needles: Created during the VR on o3
- Verification run: https://openqa.opensuse.org/tests/2246439


Vogtinator:
I'll merge in a few hours if no objections. (Normally I'd do that tomorrow, but that's a Friday)

PR #14498, opened 2022-03-16
Merge docker & podman tests

ilausuch:
In order to simplify maintenance we want to merge
both tests in one because there are a lot of
common source code lines.

- Related ticket: https://progress.opensuse.org/issues/101454
- Verification run:
- http://copland.arch.suse.de/tests/1080 (docker)
- http://copland.arch.suse.de/tests/1079 (podman)
- http://copland.arch.suse.de/tests/1087 (public cloud)
- http://openqa.suse.de/tests/8340025 (microos)
- http://openqa.suse.de/tests/8340473 (jeos) (running)

With the collaboration of @m-dati


tests/containers/container_runtime.pm
```diff
@@ -27,25 +27,27 @@
 use Mojo::Base 'containers::basetest';
 use testapi;
 use utils;
+use registration;
```
pdostal, line 30: Why is this here?

tests/containers/container_runtime.pm
```diff
@@ -56,4 +58,15 @@ sub run {
...
 }
 
+sub post_fail_hook {
+    my ($self) = @_;
+    if ($self->{runtime} eq 'podman') {
+        select_console 'log-console';
+        script_run "podman version | tee /dev/$serialdev";
+        script_run "podman info --debug | tee /dev/$serialdev";
+        $self->SUPER::post_fail_hook;
+    }
```
pdostal, line 68:
```suggestion
    }
    $self->SUPER::post_fail_hook;
```

pdostal (approved):

tests/containers/container_runtime.pm
ilausuch: Good catch. solved

tests/containers/container_runtime.pm
ilausuch: yes should be removed, was in the old podman.pm.
Solved

jlausuch:
This will need a change too:
```
.//products/microos/main.pm:70:        loadtest 'containers/podman';
.//lib/main_publiccloud.pm:23:    loadtest 'containers/podman';
.//lib/main_publiccloud.pm:28:    loadtest 'containers/docker';
```

I haven't seen any podman|docker load test in yaml schedule. Almost all container tests in OSD and O3 are using `main_containers.pm`. We might need to improve that with `microos/main.pm` and `main_publiccloud.pm` in the future.

jlausuch:
Looks good!

Maybe it would be a good time to rename "runtime" to "engine", because podman and docker are not actually "runtimes" per see as we have been calling them for long time. Some examples of runtimes are `containerd`, `crio`, etc..
What do you think?

ilausuch:
I agree

ilausuch:
renamed

jlausuch (approved):

mloviska (approved):
Me gusta!

PR #14501, opened 2022-03-17
Add setting to support qcow generation after migration for regression test

lemon-suse:
We need add setting in YAML file of regression job group to set the branch for qcow generation using condition-schedule. Set QCOW_GENERATION=1 then generate qcow after migration, QCOW_GENERATION=0 then continue run regression test after migration.

- Related ticket: https://progress.opensuse.org/issues/108263
- Needles: N/A
- Verification run: 
  x86_64:
  http://openqa.nue.suse.com/t8327814  offline, generate qcow
  http://openqa.nue.suse.com/t8327610  offline, no qcow generated
  http://openqa.nue.suse.com/tests/8334569#  online, generate qcow
  http://openqa.nue.suse.com/tests/8334234     online, no qcow generated
  s390x:
  http://openqa.nue.suse.com/tests/8334235  offline,generate qcow
  http://openqa.nue.suse.com/tests/8334237 offline no qcow
  http://openqa.nue.suse.com/tests/8334238#  online , generate qcow
  http://openqa.nue.suse.com/tests/8334160#  online, no qcow
  ppc64le:
  http://openqa.nue.suse.com/tests/8334240   offline, qcow
  http://openqa.nue.suse.com/tests/8334261  offline, no qcow
  http://openqa.nue.suse.com/tests/8334262  online, qcow
  http://openqa.nue.suse.com/tests/8334263  online, no qcow
  aarch64:
  http://openqa.nue.suse.com/tests/8334264 offline, qcow
  http://openqa.nue.suse.com/tests/8334265  offline, no qcow
  http://openqa.nue.suse.com/tests/8334266  online, qcow
  http://openqa.nue.suse.com/tests/8334327  online , no qcow

### DIFF
--- a/schedule/migration/aarch64_regression_test_offline.yaml
+++ b/schedule/migration/aarch64_regression_test_offline.yaml
@@ -6,6 +6,7 @@ description:    |
   #REGRESSION_TEST: '1' means regression test part1 include console test. '2'
   #means regression test part2 include x11 test.
   #REGRESSION_389DS: '1' means load 389ds test.
+  #QCOW_GENERATION: '1' means publish qcow after migration; '0' means no need.
 vars:
   DESKTOP: 'gnome'
   BOOT_HDD_IMAGE: 1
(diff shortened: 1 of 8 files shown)


coolgw (approved):

chcao:
From the VR, the generated qcow's names are not unique. The cases with  QCOW_GENERATION=1 should generate their own qcows.

chcao:
And before the qcows are published, the qcows are de-registered and cleanup. Why? If need to use the qcow to do debug, I think we should keep the status that it is after migration.

lemon-suse:
> From the VR, the generated qcow's names are not unique. The cases with QCOW_GENERATION=1 should generate their own qcows.

This is just VR to prove there is no problem to generate qcow when set QCOW_GENERATION=1, when user try to use it he should set own qcow name.

lemon-suse:
> And before the qcows are published, the qcows are de-registered and cleanup. Why? If need to use the qcow to do debug, I think we should keep the status that it is after migration.

No, we can use such de-registered qcow directly on openQA, if guys want to do manual test then register by themself, so de-register system can cover the two scenarios.

hjluo:
LGTM

lemon-suse:
> And before the qcows are published, the qcows are de-registered and cleanup. Why? If need to use the qcow to do debug, I think we should keep the status that it is after migration.

After some thought, I think you are right, we should not de-register system for the qcow, I can't see the use case, I will update code for it.

PR #14509, opened 2022-03-17
Public Cloud: Fix terraform for Azure

pdostal:
After we updated terraform we have problems destroying Azure instances.

- Related ticket: [poo#108518](https://progress.opensuse.org/issues/108518)
- Failed job: [OSD](https://openqa.suse.de/tests/8340177)
- Verification run: [15-SP3](http://pdostal-server.suse.cz/tests/13865) [15-SP4](http://pdostal-server.suse.cz/tests/13866#)

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -11,6 +11,7 @@ package publiccloud::provider;
 use testapi qw(is_serial_terminal :DEFAULT);
 use Mojo::Base -base;
 use publiccloud::instance;
+use publiccloud::utils 'is_azure';
 use Carp;
 use List::Util qw(max);
 use Data::Dumper;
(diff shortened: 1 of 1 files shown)


asmorodskyi (changes requested):
it is not gonna work for product runs , we not always using "offer" sometimes we use images manually uploaded and than we specifying image id

grisu48 (approved):
LGTM! But the VRs are still missing

pdostal:
> it is not gonna work for product runs , we not always using "offer" sometimes we use images manually uploaded and than we specifying image id

```terraform
offer     = var.image_id != "" ? "" : var.offer
sku       = var.image_id != "" ? "" : var.sku
```
This will in my opinion be `""` for uploaded image.

ricardobranco777 (approved):
LGTM

PR #14510, opened 2022-03-17
Introduce FIRST_BOOT_CONFIG variable for MicroOS

Vogtinator:
Allows new scenarios which don't use the combustion+ignition combination.

- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/764
- Verification runs:
  * Variable unset, as before: http://10.160.67.86/tests/1226 (still running)
  * Using jeos-firstboot: http://10.160.67.86/tests/1225 (still running)

~~I'm not sure whether the `WIZARD` variable is the best approach here. Ideally we'd also test `ignition` and `combustion` separately, so we might want something like `MICROOS_CONFIG` with values like `combustion`, `combustion+ignition`, `jeos-firstboot` instead.~~  Using `FIRST_BOOT_CONFIG` now.

### DIFF
--- a/variables.md
+++ b/variables.md
@@ -68,7 +68,6 @@ EXTRABOOTPARAMS_DELETE_NEEDLE_TARGET | string | | If specified, go back with the
 EXTRATEST | boolean | false | Enables execution of extra tests, see `load_extra_tests`
 FIRST_BOOT_CONFIG | string | combustion+ignition | The method used for initial configuration of MicroOS images. Possible values are: `combustion`, `ignition`, `combustion+ignition` and `wizard`. For ignition/combustion, the job needs to have a matching HDD attached.
 FLAVOR | string | | Defines flavor of the product under test, e.g. `staging-.-DVD`, `Krypton`, `Argon`, `Gnome-Live`, `DVD`, `Rescue-CD`, etc.
-SALT_FORMULAS_PATH | string | | Used to point to a tarball with relative path to [/data/yast2](https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/data/yast2) which contains all the needed files (top.sls, form.yml, ...) to support provisioning with Salt masterless mode.
 FULLURL | string | | Full url to the factory repo. Is relevant for openSUSE only.
 FULL_LVM_ENCRYPT | boolean | false | Enables/indicates encryption using lvm. boot partition may or not be encrypted, depending on the product default behavior.
 FUNCTION | string | | Specifies SUT's role for MM test suites. E.g. Used to determine which SUT acts as target/server and initiator/client for iscsi test suite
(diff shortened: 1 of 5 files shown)


mloviska:
> It's also a bit weird how to boot MicroOS disk images. The usual approach of using `installation/bootloader_uefi` doesn't work because of missing needles (`grub2-bootloader` vs. `grub2`/`bootloader` tags) and missing code for the grub config cursor moves and `microos/disk_boot` doesn't support this either because it expects a console at the end.

IMO, it would be nice to replace `microos/disk_boot` by `installation/bootloader_uefi` in the future.
As the use case is only for QEMU for both and `disk_boot` lacks support to inject kernel parameters.

mloviska (approved):
LGTM! Is it planned to use different initial setup tools for SLE/Leap as well ?

Vogtinator:
> LGTM! Is it planned to use different initial setup tools for SLE/Leap as well ?

Yep, but no idea when.

jlausuch:
That looks nice!

I would actually vote for `MICROOS_CONFIG` or something more generic `FIRST_BOOT_CONFIG`.

Vogtinator:
Ok, I'll try that.

variables.md
```diff
@@ -66,8 +66,8 @@ EXTRABOOTPARAMS_BOOT_LOCAL | string | | Boot options applied during the boot pro
 EXTRABOOTPARAMS_DELETE_CHARACTERS | string | | Characters to delete from boot prompt.
 EXTRABOOTPARAMS_DELETE_NEEDLE_TARGET | string | | If specified, go back with the cursor until this needle is matched to delete characters from there. Needs EXTRABOOTPARAMS_BOOT_LOCAL and should be combined with EXTRABOOTPARAMS_DELETE_CHARACTERS.
 EXTRATEST | boolean | false | Enables execution of extra tests, see `load_extra_tests`
+FIRST_BOOT_CONFIG | string | combustion+ignition | The method used for initial configuration of MicroOS images (combustion/ignition/wizard). Needs to have a matching HDD attached.
```
jlausuch, line 69:
```suggestion
FIRST_BOOT_CONFIG | string | combustion+ignition | The method used for initial configuration of MicroOS images. Possible values are: combustion | ignition | combustion+ignition | wizard`. For ignition/combustion, the job needs to have a matching HDD attached.
```

jlausuch: and I know I suggested it, but what about `FIRST_BOOT_MODE` ?

jlausuch:
```suggestion
FIRST_BOOT_CONFIG | string | combustion+ignition | The method used for initial configuration of MicroOS images. Possible values are: "combustion" "ignition" "combustion+ignition" "wizard". For ignition/combustion, the job needs to have a matching HDD attached.
```

Vogtinator: Changed

jlausuch (approved):

mloviska (approved):

Vogtinator:
I added new microos-{combustion,ignition,wizard} tests to TW

PR #14513, opened 2022-03-17
lib/main_common: Run systemd_wo_udev after repositories are set up

Vogtinator:
The module uses zypper and relies on repositories being configured correctly,
but that only happened later.

- Verification run: https://openqa.opensuse.org/tests/2249048


mloviska:
oh, Thanks!

PR #10474, opened 2020-06-08
Stabilize KDE live installation on aarch64

OleksandrOrlov:
The PR adds the required steps to stabilize Tumbleweed installation using Live CD on aarch64.

Changes introduced:
1. live_installation test module sporadically failed on aarch64 because when clicking on "Installation" icon on Desktop, the mouse button was hold too long and popup appeared instead of opening the application. The PR changes the click with `x11_start_program`, which also allows to open the app but without relying on click;
2. await_install failed because black screen appeared after some time without activity. The PR just adjusts the existing approach with mouse moving from time to time. As shown from 4 verification runs, this is effective and fixes the issue;
3. Legacy `sub handle_livecd_screenlock` seems to be not effective with the current product, but it is still left (maybe temporary), just to not broke existing tests that potentially may use it (e.g. on x64). Though, the way how `check_screen` is used here was changed. As previously it used array in function arguments instead of arrayref, that is required by function definition;
4. Also, `QEMURAM=4096` `QEMUCPUS=4` `TIMEOUT_SCALE=4` were added to Test Suite settings.

- Related ticket: https://progress.opensuse.org/issues/66355
- Verification runs:
https://openqa.opensuse.org/tests/1293429
https://openqa.opensuse.org/tests/1293430
https://openqa.opensuse.org/tests/1293431
https://openqa.opensuse.org/tests/1293432

PR #14514, opened 2022-03-17
installation/await_install: Remove dead code for screenlock handling

Vogtinator:
7e79403f937b removed the relevant tags from @tags, so the screenlock condition
inside the loop is only ever reached if a different tag already matched
check_screen. In that case the two check_screen calls actually cause a delay,
which can cause the module to miss stopping the reboot countdown in time.

- Verification run: http://10.160.67.86/tests/1229

### DIFF
--- a/tests/installation/await_install.pm
+++ b/tests/installation/await_install.pm
@@ -20,17 +20,6 @@
 #     - Send 'alt-d', check for needle 'ERROR-removing-package-details'
 #     - Send 'alt-i', check for needle 'WARNING-ignoring-package-failure'
 #     - Send 'alt-o'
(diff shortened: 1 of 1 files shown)


okurz (approved):
Yes, that's what we tried to tell in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10474#discussion_r437971702 and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10474#pullrequestreview-427846430 (and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/10474#issuecomment-641858993 ). Good that you take this into your own hands to at least repair that flaw.

PR #14515, opened 2022-03-17
Fix failing ssh-keyscan in boottime

grisu48:
Fix the failing ssh-keyscan command in the boottime test runs.

- Related ticket: https://progress.opensuse.org/issues/108320
- Verification run: http://duck-norris.qam.suse.de/t8360

### DIFF
--- a/tests/publiccloud/boottime.pm
+++ b/tests/publiccloud/boottime.pm
@@ -297,6 +297,7 @@ sub measure_timings {
     }
 
     $ret->{analyze}->{ssh_access} = $instance->wait_for_ssh(timeout => 300);
+    assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
 
     my ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
     $ret->{analyze}->{$_} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -257,9 +257,11 @@ sub create_instances {
     my @vms = $self->terraform_apply(%args);
     foreach my $instance (@vms) {
         record_info("INSTANCE $instance->{instance_id}", Dumper($instance));
-        $instance->wait_for_ssh() if ($args{check_connectivity});
-        # Install server's ssh publicckeys to prevent authenticity interactions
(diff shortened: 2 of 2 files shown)


pdostal (approved):
:rocket:

asmorodskyi (changes requested):
correct me if I wrong but from what I know currently create_instances is called with check_connectivity=> 1 in all cases without any exceptions . 
so from this perspective having :
```
 $instance->wait_for_ssh() if ($args{check_connectivity});
```

is 100% the same as having :

```
if ($args{check_connectivity}) {
            $instance->wait_for_ssh();
        }
```

also I wonder why we want to skip ssh-keyscan if someone will call create_instances with check_connectivity => 0 ?

grisu48:
For the `boottime` test, where we measure the time until ssh is reachable it is crucial that we don't wait for ssh in `create_instance`. Waiting for ssh there is however a necessary requirement for `ssh-keyscan` and the reason why they don't go together.

jlausuch (approved):
Makes sense!

asmorodskyi:
> If `check_connectivity =>0` then the routine returns while the instance is booting. So `ssh-keyscan` cannot be executed in this routine, because we are not waiting for `ssh` anymore. And that's exactly the purpose of `check_connectivity => 0`.

ok I get this part , but avoiding running of `ssh-keyscan` in case of `check_connectivity` means that in such a case we will face "fingerprint have changed " issues which `ssh-keyscan` suppose to solve , right ?

tests/publiccloud/boottime.pm
```diff
@@ -297,6 +297,7 @@ sub measure_timings {
     }
 
     $ret->{analyze}->{ssh_access} = $instance->wait_for_ssh(timeout => 300);
+    assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
```
asmorodskyi, line 300: exactly ! :)

asmorodskyi (approved):
LGTM

tests/publiccloud/boottime.pm
grisu48: I re-added it and don't wanna hear now any comments about duplication here *pointing at the last PR I made here* :wink:

asmorodskyi: my bad , I was not realizing what price we were paying here :)

PR #14519, opened 2022-03-18
Openstack jeos

mloviska:

- Related ticket: [[JeOS OpenStack] Enable jeos-containers](https://progress.opensuse.org/issues/107452)
- Verification run: [openstack-containers](http://kepler.suse.cz/tests/14628)


lib/LTP/utils.pm
```diff
@@ -49,7 +49,7 @@ sub loadtest_runltp {
 
 sub shutdown_ltp {
     loadtest_kernel('proc_sys_dump') if get_var('PROC_SYS_DUMP');
-    loadtest_kernel('shutdown_ltp', @_);
+    loadtest_kernel('shutdown_ltp', @_) unless is_openstack;
```
jlausuch, line 52: Can you add a comment why we don't execute this module in openstack?

lib/opensusebasetest.pm
```diff
@@ -1394,6 +1394,19 @@ sub post_fail_hook {
     send_key 'spc';
 
     $self->export_logs;
+
+    if (is_public_cloud() || is_openstack) {
```
jlausuch, line 1398: For consistency:
`if (is_public_cloud() || is_openstack()) {`
or
`if (is_public_cloud || is_openstack) {`

lib/consoletest.pm
```diff
@@ -8,7 +8,7 @@ use strict;
 use warnings;
 use testapi;
 use known_bugs;
-use version_utils 'is_public_cloud';
+use version_utils qw(is_public_cloud is_openstack);
```
jlausuch, line 11: leftover?

lib/main_common.pm
```diff
@@ -568,47 +570,69 @@ sub load_system_role_tests {
...
-        }
-        if (is_sle) {
-            loadtest "console/suseconnect_scc";
-            loadtest "jeos/efi_tid" if (get_var('UEFI') && is_sle('=12-sp5'));
-        }
+        loadtest "jeos/prepare_openstack", run_args => $args;
+        loadtest 'publiccloud/ssh_interactive_start', run_args => $args;
+    }
+
+    loadtest "jeos/record_machine_id";
```
jlausuch, line 587: I wonder if we can reuse somehow `load_jeos_tests` here, as both methods are quite similar at this stage. Of course, we would need to do things like `load_boot_tests unless is_openstack` and similar in other places... not sure if it's worth it or not..

lib/main_containers.pm
```diff
@@ -87,18 +87,18 @@ sub load_host_tests_podman {
         load_container_engine_test('podman');
         load_image_test('podman');
         load_3rd_party_image_test('podman');
-        loadtest 'containers/podman_firewall';
+        loadtest 'containers/podman_firewall' unless is_openstack;
```
jlausuch, line 90: why?

lib/main_containers.pm
```diff
@@ -87,18 +87,18 @@ sub load_host_tests_podman {
...
         loadtest 'containers/buildah' unless is_sle_micro;
         # https://github.com/containers/podman/issues/5732#issuecomment-610222293
-        loadtest 'containers/rootless_podman' unless (is_sle('=15-sp1'));
+        loadtest 'containers/rootless_podman' unless (is_sle('=15-sp1') || is_openstack);
```
jlausuch, line 93: I guess it's challenging, as it switches root/user consoles :)

lib/main_containers.pm
```diff
@@ -87,18 +87,18 @@ sub load_host_tests_podman {
...
     }
 }
 
 sub load_host_tests_docker {
     load_container_engine_test('docker');
     load_image_test('docker');
     load_3rd_party_image_test('docker');
-    loadtest 'containers/docker_firewall';
+    loadtest 'containers/docker_firewall' unless is_openstack;
```
jlausuch, line 101: same question as for podman_firewall. What is the issue of running this module in OpenStack? Maybe a comment would help.

lib/version_utils.pm
```diff
@@ -652,6 +653,16 @@ sub is_public_cloud {
     return get_var('PUBLIC_CLOUD');
 }
 
+=head2 is_openstack
```
jlausuch, line 656: After reading this method, I wonder if we should name it `is_jeos_openstack`... wdyt?

lib/publiccloud/openstack.pm
```diff
@@ -54,12 +56,16 @@ sub create_keypair {
 
 sub delete_keypair {
     my $self = shift;
-    my $name = shift || $self->ssh_key_name;
+    return unless $self->{ssh_key_name};
```
jlausuch, line 59: good catch

jlausuch (commented):
This is an amazing work!

lib/main_containers.pm
mloviska: openstack image does not have firewalld

lib/main_containers.pm
mloviska: Oh, this should work by now. Let me check again! Thanks!

lib/main_containers.pm
pdostal: And the connectivity is tested anyways.

pdostal (approved):
This is very good!

lib/LTP/utils.pm
mloviska: It is the same as for https://openqa.suse.de/tests/8363826. The test case failed due to it shutdown the instance, but openqa was still running a VM. Thus it has died. 
Nevertheless I will drop this part and focus on LTP in dedicated PR. Thanks!

lib/main_common.pm
mloviska: Yes, we can however then the function becomes hard to read. I fancy wrapper load function as it is used in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/main_publiccloud.pm#L35. The thing is to keep it somehow concise and our tests need to end with `ssh_interactive_end` that will be loaded with exactly the same `run_args` object as `ssh_interactive_start'`

lib/main_containers.pm
mloviska: It is not executed in [sle15sp4](https://openqa.suse.de/tests/8343826) as well. Maybe a problem with reboots ? http://kepler.suse.cz/tests/14643#step/rootless_podman/19

lib/version_utils.pm
mloviska: yeah, we can do that, OTOH I am not sure whether we can run an ISO installation in openstack. Thus I have kinda thought it is implicit

lib/main_containers.pm
jlausuch: Ok!

lib/main_common.pm
jlausuch: ok, let's keep it easier to read and not pollute legacy function which is used for openSUSE as well. Thanks!

lib/main_containers.pm
jlausuch: Decided to not run it, same as for PC tests.

lib/consoletest.pm
mloviska: Nope, I have just forgotten to add `return` right after we come back from the super post_fail_hook. Not sure whether the log console part is needed in case of PC/OpenStack jobs. Wdyt ?

jlausuch: I don't think it's needed, everything is happening remotely.. Local logs won't help much as we are booting the PC tools image which is not the SUT.

lib/version_utils.pm
jlausuch: Should be fine, but PC is also producing OpenStack images, e.g. `SLES15-SP4-CHOST-BYOS.aarch64-0.9.7-OpenStack-Build2.20.qcow2`

jlausuch (approved):

lib/version_utils.pm
mloviska: Is that anything we might need to test in the future ? If so then it kinda makes sense to keep it or not ?

jlausuch: > Is that anything we might need to test in the future ?
I really hope not :D 

 If so then it kinda makes sense to keep it or not ?
Merged :)

PR #14522, opened 2022-03-18
[WIP] Extend clamd service startup timeout to 30min

pablo-herranz:
The time provided for service *clamd.service* to start was not enough in *ppc64le* machines, so it has been extended.

- Related ticket: https://progress.opensuse.org/issues/108082
- Needles: *not needed*
- Verification runs:

| | | | | |
| - | - | - | - | - |
| **SLE15-SP4** | [x86_64](https://openqa.suse.de/t8351556) | [aarch64](https://openqa.suse.de/t8351557) | [ppc64le](https://openqa.suse.de/t8351555) | [s390x](https://openqa.suse.de/t8351558) |
| **Tumbleweed** | [x86_64](https://openqa.opensuse.org/t2251217) | [aarch64](https://openqa.opensuse.org/t2251209) | [ppc64le](https://openqa.opensuse.org/t2251210) | |


tests/console/clamav.pm
```diff
@@ -28,7 +28,7 @@ sub scan_and_parse {
     my $cmd = shift;
     my $log_file = "$cmd.log";
 
-    script_run "$cmd -i --log=$log_file eicar_test_files", 300;
+    script_run "$cmd -i --log=$log_file eicar_test_files", 1800;
```
punkioudi, line 31: why did you change this timeout? As far as I can see from the progress ticket the fail happens when you try to start clamd service.

tests/console/clamav.pm
```diff
@@ -74,14 +74,12 @@ sub run {
 
     # Clamd start timeout sometimes. The default systemd timeout is 90s,
     # override it with a longer duration in runtime.
-    my $runtime_dir = '/run/systemd/system/clamd.service.d';
-    assert_script_run "mkdir -p $runtime_dir";
-    assert_script_run "echo -e \'[Service]\\nTimeoutSec=400\' > $runtime_dir/override.conf";
+    assert_script_run("sed -i '/;User/i TimeoutSec=30min' /usr/lib/systemd/system/clamd.service");
```
punkioudi, line 77: how did you conclude to this 30min timeout?

tests/console/clamav.pm
pablo-herranz: Yes but when I made my tests, it ended up that this gave a timeout too. So I went ahead and fixed all the timeouts.

tests/console/clamav.pm
pablo-herranz: Tried incrementing by 10min until I reached a sufficient time.

tests/console/clamav.pm
punkioudi: what was the percentage of failure here? still I would suggest that you try a smaller timeout than 1800s

tests/console/clamav.pm
punkioudi: I think that 30min is too big timeout for starting a service. It seems a bit buggy to me

pablo-herranz:
After further research, the issue can be solved by increasing the RAM amount of `ppc64le` machines without need of changing the timeouts.

PR #14531, opened 2022-03-21
The EFI secret key for hibernation is only supported on x86_64

mloviska:
On other architectures with enabled SecureBoot the hibernation should be
locked down. More information can be found in
https://bugzilla.suse.com/show_bug.cgi?id=1187591

- Verification run: [sle-15-SP4-JeOS-for-kvm-and-xen-aarch64-Build1.53-jeos-filesystem@aarch64](https://openqa.suse.de/t8362696)

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -156,14 +156,6 @@
         },
         "type": "bug"
     },
-    "bsc#1187591": {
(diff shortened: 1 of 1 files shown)


jlausuch (approved):

mloviska:
> If it's not available for other architectures, the error message should not be shown at all IMO.

Thanks for the comment in the bug. I agree, later we can simply remove this reference if the behaviour changes.

PR #14533, opened 2022-03-21
lib/containers/container_images: Increase timeout for zypper in systemd

Vogtinator:
It has to refresh the repos first and then download and install the packages,
which can take longer than the default timeout of 90s. Use 300s like above.

- Verification run: https://openqa.opensuse.org/tests/2256620


Vogtinator:
Or should this also use `script_retry`? IMO that shouldn't be necessary and it wouldn't really have the same effect, as there's no state preserved in the containers.

mloviska (approved):
I think it is better to increase the TO rather than replacing for `script_retry`

DimStar77 (approved):

PR #14538, opened 2022-03-21
Containers: Improve and deduplicate containers::common::check_containers_connectivity()

pdostal:
* The `check_containers_connectivity()` for `podman` and `docker` basically works the same - so I deduplicated it.
 * All three checks now executed firstly from the host and then from container. This should improve debugging.
 * When `dig` is available or installable we test the outside connectivity firstly without name resolution.

- Verification run: [podman@Tumbleweed](https://openqa.opensuse.org/tests/2262070), [docker@SLE12-SP5](http://pdostal-server.suse.cz/tests/13982), [podman@SLE Micro 5.2](https://openqa.suse.de/tests/8385286)

### DIFF
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -10,12 +10,12 @@ use warnings;
 use testapi;
 use registration;
 use utils qw(zypper_call systemctl file_content_replace script_retry script_output_retry);
-use version_utils qw(is_sle is_leap is_microos is_sle_micro is_opensuse is_jeos is_public_cloud get_os_release check_version);
-use containers::utils qw(can_build_sle_base registry_url);
+use version_utils qw(is_sle is_leap is_microos is_sle_micro is_opensuse is_jeos is_public_cloud get_os_release check_version is_transactional);
+use containers::utils qw(can_build_sle_base registry_url container_ip container_route);
 
 our @EXPORT = qw(is_unreleased_sle install_podman_when_needed install_docker_when_needed install_containerd_when_needed
-  test_container_runtime test_container_image scc_apply_docker_image_credentials
-  scc_restore_docker_image_credentials install_buildah_when_needed test_rpm_db_backend activate_containers_module);
+  test_container_runtime test_container_image scc_apply_docker_image_credentials scc_restore_docker_image_credentials
+  install_buildah_when_needed test_rpm_db_backend activate_containers_module check_containers_connectivity);
 
 
 sub is_unreleased_sle {
@@ -261,6 +261,37 @@ sub test_rpm_db_backend {
     if ($host_distri eq 'opensuse-tumbleweed' || ($host_distri eq 'sles' && check_version('>=15-SP3', "$running_version-SP$sp", qr/\d{2}(?:-sp\d)?/))) {
         validate_script_output "$runtime run $image rpm --eval %_db_backend", sub { m/ndb/ };
     }
+}
+
+sub check_containers_connectivity {
+    my $runtime = shift;
+    record_info "connectivity", "Checking that containers can connect to the host, to each other and outside of the host";
+    my $container_name = 'sut_container';
+
+    # Run container in the background (sleep for 30d because infinite is not supported by sleep in busybox)
+    assert_script_run "$runtime pull " . registry_url('alpine');
+    assert_script_run "$runtime run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
+    my $container_ip = container_ip $container_name, $runtime;
+
+    # Connectivity to host check
+    my $container_route = container_route($container_name, $runtime);
+    assert_script_run "ping -c3 " . $container_route;
+    assert_script_run "$runtime run --rm " . registry_url('alpine') . " ping -c3 " . $container_route;
+
+    # Cross-container connectivity check
+    assert_script_run "ping -c3 " . $container_ip;
+    assert_script_run "$runtime run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outside IP connectivity check
+    script_retry "ping -c3 8.8.8.8", retry => 3, delay => 120;
+    script_retry "$runtime run --rm " . registry_url('alpine') . " ping -c3 8.8.8.8", retry => 3, delay => 120;
+
+    # Outside IP+DNS connectivity check
+    script_retry "ping -c3 google.com", retry => 3, delay => 120;
+    script_retry "$runtime run --rm " . registry_url('alpine') . " ping -c3 google.com", retry => 3, delay => 120;
+
+    # Kill the container running on background
+    assert_script_run "$runtime kill $container_name";
 }
 
 1;
--- a/tests/containers/docker_firewall.pm
+++ b/tests/containers/docker_firewall.pm
@@ -12,6 +12,7 @@ use testapi;
 use utils 'script_retry';
 use version_utils qw(is_sle is_leap);
 use containers::utils qw(registry_url container_ip);
+use containers::common 'check_containers_connectivity';
 use Utils::Systemd 'systemctl';
 
 my $stop_firewall = 0;
@@ -61,7 +62,7 @@ sub run {
     script_retry "docker ps -q | wc -l | grep 0", delay => 5, retry => 12;
 
     # Test the connectivity of Docker containers
-    $engine->check_containers_connectivity();
+    check_containers_connectivity($engine);
 
     # Stop the firewall if it was started by this test module
     if ($stop_firewall == 1) {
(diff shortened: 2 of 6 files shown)


tests/containers/podman_firewall.pm
```diff
@@ -32,24 +33,24 @@ sub run {
     }
 
     # cni-podman0 interface is created when running the first container
-    assert_script_run "podman pull " . registry_url('alpine');
-    assert_script_run "podman run --rm " . registry_url('alpine');
+    assert_script_run "$podman pull " . registry_url('alpine');
```
jlausuch, line 36: Does this work? This is now an object which doesn't return any string `podman` or anything like this. The reason to have this object is to :
1) install & config registries
2) optionally use the subroutines defined in the Class

So you can do things like `$podman->methodXYZ()` but you can't use it as replacement for `podman` as a string.

Either you keep `podman pull` or you do `$podman->_engine_assert_script_run('pull XYZ')`.

jlausuch (changes requested):
Looks good to save some code, but I have some concerns about some code, please check, I might be wrong.

lib/containers/common.pm
```diff
@@ -262,4 +262,39 @@ sub test_rpm_db_backend {
...
+    # Connectivity to host check
+    my $container_route = container_route($container_name, $runtime);
+    assert_script_run "ping -c3 " . $container_route;
+    assert_script_run "$runtime run --rm " . registry_url('alpine') . " ping -c3 " . $container_route;
+
+    # Cross-container connectivity check
+    assert_script_run "ping -c3 " . $container_ip;
+    assert_script_run "$runtime run --rm " . registry_url('alpine') . " ping -c3 " . $container_ip;
+
+    # Outisde IP connectivity check
```
b10n1k, line 284: typo

lib/containers/common.pm
```diff
@@ -262,4 +262,39 @@ sub test_rpm_db_backend {
...
 }
 
+sub check_containers_connectivity {
+    my $runtime = shift;
+    record_info "connectivity", "Checking that containers can connect to the host, to each other and outside of the host";
+    my $container_name = 'sut_container';
+
+    # Run container in the background
+    assert_script_run "$runtime pull " . registry_url('alpine');
+    assert_script_run "$runtime run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
```
b10n1k, line 272: just making sure. did you really meant 30 days? :)

lib/containers/common.pm
```diff
@@ -262,4 +262,39 @@ sub test_rpm_db_backend {
...
+    zypper_call('in bind-utils') if (!is_transactional() && script_run("which dig") != 0);
+    if (script_run("which dig") == 0) {
+        my $remote_ip = script_output 'dig +short google.com';
+        assert_script_run "wget $remote_ip";
+        assert_script_run "$runtime run --rm " . registry_url('alpine') . " wget $remote_ip";
+    }
+
+    # Outside IP+DNS connectivity check
+    assert_script_run "wget google.com";
+    assert_script_run "$runtime run --rm " . registry_url('alpine') . " wget google.com";
```
b10n1k, line 294: Some observations:
- looks like a if..else thing here
- Why do you need the first `assert_script_run`? the host is tested already in this point anyway

tests/containers/podman_firewall.pm
pdostal: This works and this is valid. I will not use those `_engine` subroutines as they are slowly getting deprecated.

lib/containers/common.pm
pdostal: Yes as the `infinite` value does not work on `sleep` from `busybox`. :smile:

lib/containers/common.pm
pdostal: It's not `if..else` - I need to check that the connectivity (including DNS) works. To be sure that the problem is container-specific and not host-related, I decided to run the command firstly from the host and then from the container.

lib/containers/common.pm
grisu48: Please add a corresponding comment e.g. `30d because infinite is not supported by sleep in busybox`

lib/containers/common.pm
b10n1k: IMO this doesnt make sense. But i wont object if others are good with it. thanks @pdostal

jlausuch: I agree with both of you. I would actually compare somehow both outputs, to make sure that it doesn't return some 50X error or anything weird.

I would approach this differently, instead of `wget`, I would use `nslookup`, since we want to check DNS resolution only.
What about something like this?

```
my $server1 = script_output 'nslookup google.com | grep Server';
my $server2 = script_output 'podman run --rm alpine nslookup google.com | grep Server';
die 'some error message' if ($server1 ne $server2);
```

tests/containers/podman_firewall.pm
jlausuch: ok, but using the podman object like this looks strange... since this module is called `podman_something`, you can just use `podman` as it was before.

lib/containers/common.pm
pdostal: Well but `nslookup` as well as `dig` are in `bind-utils` package sometimes not installed by default neither in (open)SUSE nor in Alpine container. That's why I use `wget` - it's everywhere.

tests/containers/podman_firewall.pm
pdostal: OK. I'll just pet it back :smile:.

But when you have `$engine` you don't have to do `$engine->engine`.

lib/containers/common.pm
jlausuch: `nslookup` is in Alpine. The commands I posted were tested locally.

tests/containers/docker_firewall.pm
```diff
@@ -61,7 +62,7 @@ sub run {
     script_retry "docker ps -q | wc -l | grep 0", delay => 5, retry => 12;
 
     # Test the connectivity of Docker containers
-    $engine->check_containers_connectivity();
+    check_containers_connectivity($engine);
```
jlausuch, line 65:
```suggestion
    check_containers_connectivity($engine->runtime);
```
maybe?

pdostal: It's not really necessary. Please see #13408

jlausuch: Yes, sorry, I forgot about that magic :)

lib/containers/common.pm
jlausuch: ah ok, nslookup not available in SLE Micro... damn... ok to use wget. But still, the idea of comparing both htmls would be interesting. Not sure how though...

pdostal: @jlausuch what do you think now? For comparing the output we probably need more stable website plus we use HTTP anyways in the other tests (f.e. `zypper` uses it).

jlausuch: Yes! And even  `ping` is actually more likely to be installed in any host and container than `wget` :D 
Thanks!

jlausuch (approved):

lib/containers/common.pm
```diff
@@ -262,4 +262,35 @@ sub test_rpm_db_backend {
...
+    my $container_name = 'sut_container';
+
+    # Run container in the background (sleep for 30d because infinite is not supported by sleep in busybox)
+    assert_script_run "$runtime pull " . registry_url('alpine');
+    assert_script_run "$runtime run -id --rm --name $container_name -p 1234:1234 " . registry_url('alpine') . " sleep 30d";
+    my $container_ip = container_ip $container_name, $runtime;
+
+    # Connectivity to host check
+    my $container_route = container_route($container_name, $runtime);
+    assert_script_run "ping -c3 " . $container_route;
```
grisu48, line 277: Please do a `script_retry` on all ping commands with at `retry => 3, delay => 120`, otherwise we have to do this in a month or so anyways.

I would do with for ALL `ping` commands, external and internal. Experience has shown us that not doing so always leads to pain.

grisu48 (changes requested):
In general a LGTM, but I would like to prevent the typical network timeout failures by using `script_retry` on all network test parts.

lib/containers/common.pm
jlausuch: or just `ping -c1000`  :)

pdostal: That would prolong the test for hours.

jlausuch: > That would prolong the test for hours 😄

yep, sorry, it was a joke :p

grisu48 (approved):
LGTM :rocket:

grisu48:
Something's still fishy with the latest VR

jlausuch (approved):

jlausuch:
Can you please run more VRs with latest code?

PR #14542, opened 2022-03-21
Containers: Get rid of several RunArgs initiations

pdostal:
This is attempt to deduplicate the `OpenQA::Test::RunArgs->new()` objects in `lib/main_containers.pm`.

- Verification runs (all passing): [SLE15-SP4-containerd](http://pdostal-server.suse.cz/tests/14043), [SLE15-SP4-docker](http://pdostal-server.suse.cz/tests/14044), [Tumbleweed-podman](http://pdostal-server.suse.cz/tests/14045), [SLE15-SP3-bci-python](http://pdostal-server.suse.cz/tests/14053), [SLE15-SP3-image-on-SLE15-SP3-host-podman](http://pdostal-server.suse.cz/tests/14052), [MicroOS-containerhost](http://pdostal-server.suse.cz/tests/14046), [SLE15-SP3-EC2](http://pdostal-server.suse.cz/tests/14145), [SLE15-SP4-JeOS-containers](http://pdostal-server.suse.cz/tests/14051), [SLE-Micro-5.2-containers](http://pdostal-server.suse.cz/tests/14054), [Tumbleweed-image-on-Centos](http://pdostal-server.suse.cz/tests/14055), [Tumbleweed-image-on-Ubuntu](http://pdostal-server.suse.cz/tests/14056), [Tumbleweed-containers_image](http://pdostal-server.suse.cz/tests/14057), [Tumbleweed-containers_host_podman](http://pdostal-server.suse.cz/tests/14147), [Tumbleweed-containers_host_docker](http://pdostal-server.suse.cz/tests/14059)

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -46,34 +46,28 @@ sub is_ubuntu_host {
 }
 
 sub load_image_test {
-    my ($runtime, $run_args) = @_;
-    $run_args->{runtime} = $runtime;
-
-    loadtest('containers/image', run_args => $run_args, name => "image_$runtime");
+    my ($run_args) = @_;
+    loadtest('containers/image', run_args => $run_args, name => 'image_' . $run_args->{runtime});
 }
 
 sub load_3rd_party_image_test {
-    my ($runtime, $run_args) = @_;
-    $run_args->{runtime} = $runtime;
-
-    loadtest('containers/third_party_images', run_args => $run_args, name => $runtime . "_3rd_party_images");
+    my ($run_args) = @_;
+    loadtest('containers/third_party_images', run_args => $run_args, name => $run_args->{runtime} . '_3rd_party_images');
 }
 
 sub load_container_engine_test {
-    my ($runtime, $run_args) = @_;
-    $run_args->{runtime} = $runtime;
-
-    loadtest('containers/container_engine', run_args => $run_args, name => $runtime);
+    my ($run_args) = @_;
+    loadtest('containers/container_engine', run_args => $run_args, name => $run_args->{runtime});
 }
 
 sub load_image_tests_podman {
     my ($run_args) = @_;
-    load_image_test('podman', $run_args);
+    load_image_test($run_args);
 }
 
 sub load_image_tests_docker {
     my ($run_args) = @_;
-    load_image_test('docker', $run_args);
+    load_image_test($run_args);
     # container_diff package is not avaiable for <=15 in aarch64
     # Also, we don't want to run it on 3rd party hosts
     unless ((is_sle("<=15") and is_aarch64) || get_var('CONTAINERS_NO_SUSE_OS')) {
@@ -85,9 +79,9 @@ sub load_host_tests_podman {
     my ($run_args) = @_;
     if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+") || is_sle_micro || is_microos) {
         # podman package is only available as of 15-SP1
-        load_container_engine_test('podman', $run_args);
-        load_image_test('podman', $run_args);
-        load_3rd_party_image_test('podman', $run_args);
+        load_container_engine_test($run_args);
+        load_image_test($run_args);
+        load_3rd_party_image_test($run_args);
         # Firewall is not installed in JeOS OpenStack and MicroOS
         loadtest 'containers/podman_firewall' unless (is_openstack || is_microos);
         # Buildah is not available in SLE Micro and MicroOS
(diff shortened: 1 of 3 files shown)


pdostal:
Hello @ilausuch can you please take a look as well?

lib/main_containers.pm
```diff
@@ -141,6 +141,7 @@ sub load_container_tests {
     }
 
     foreach (split(',\s*', $runtime)) {
+        my $run_args = OpenQA::Test::RunArgs->new();
```
grisu48, line 144: Does this really need to happen for each new runtime or should this be before the `foreach` loop?

lib/main_containers.pm
```diff
@@ -149,14 +150,14 @@ sub load_container_tests {
             }
             else {
                 # Common openQA image tests
-                load_image_tests_podman() if (/podman/i);
-                load_image_tests_docker() if (/docker/i);
+                load_image_tests_podman($run_args) if (/podman/i);
```
jlausuch, line 153: I don't get why we need to pass it to this method, which doesn't need it.

lib/main_containers.pm
b10n1k: i would say yes. As i remember the design(i am not sure what has changed since then) each instance has some properties. if the two instances share the same `RunArgs` might cause some conflicts. it was depended what are those properties and how are used in general. Having said that, you can have one single `RunArgs` but then it might need changes in other places too.

lib/main_containers.pm
```diff
@@ -95,9 +95,10 @@ sub load_host_tests_podman {
 }
 
 sub load_host_tests_docker {
-    load_container_engine_test('docker');
-    load_image_test('docker');
-    load_3rd_party_image_test('docker');
+    my ($run_args) = @_;
+    load_container_engine_test('docker', $run_args);
```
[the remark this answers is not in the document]
pdostal: Please take a look how the `load_container_engine_test` subroutine is defined. It passes `run_args` parameter to `loadtest`.

lib/main_containers.pm
pdostal: The `load_image_tests_podman` subroutine calls `load_image_test` subroutine which again passes `run_args` parameter to `loadtest`.

lib/main_containers.pm
pdostal: Maybe Felix is right as we still do `$run_args->{runtime} = $runtime` in `load_image_test`, `load_3rd_party_image_test` and in `load_container_engine_test`.

grisu48: AFAIK the `$run_args` are used for publiccloud specific arguments, so those should not change per container runtime.

lib/main_containers.pm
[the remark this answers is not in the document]
grisu48: `$run_args` is needed for the publiccloud test runs, which are using the same routines as the "normal" test runs.
Before we were creating a new instance of the `$run_args` in each subroutine, which is wrong because the important variables like instance are then not shared between them.

This PR moves the creation of `$run_args` to one place and then passes those variables to each test run, which was creating a new instance before, thus violating the singleton pattern.

lib/main_containers.pm
grisu48: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14542#discussion_r831873707

lib/main_containers.pm
pdostal: They are also used to pass the runtime to container tests. I will try to move this line out of the `foreach` loop.

lib/main_containers.pm
```diff
@@ -45,35 +45,34 @@ sub is_ubuntu_host {
...
+    $run_args->{runtime} = $runtime;
 
-    loadtest('containers/container_engine', run_args => $args, name => $runtime);
+    loadtest('containers/container_engine', run_args => $run_args, name => $runtime);
 }
 
 sub load_image_tests_podman {
-    load_image_test('podman');
+    my ($run_args) = @_;
+    load_image_test('podman', $run_args);
```
jlausuch, line 70: would it make sense to do this here:
```
$run_args->{runtime} = 'podman';
load_image_test($run_args);
```

and then 
```
sub load_image_test {
    my ($run_args) = @_;
    loadtest('containers/image', run_args => $run_args, name => "image_$runtime");
}
```

WDYT?

lib/main_containers.pm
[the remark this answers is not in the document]
jlausuch: ok, thanks for clarification.

lib/main_containers.pm
jlausuch: I would also move this out of the loop. We are setting the runtime argument accordingly in each function.

lib/main_containers.pm
pdostal: I did what you suggested as separate commit. Please re-check @jlausuch

jlausuch (commented):
LGTM. Waiting for VRs.

grisu48 (approved):
Also LGTM, however I'd expect some fallout due to widespread changes. I'm ok with it but we need to keep a good eye on the test runs after merging.

ilausuch (approved):

jlausuch (approved):

PR #14545, opened 2022-03-22
Restart vconsole service to avoid font problem in x11 service check

tinawang123:
Restart vconsole service to avoid font problem in x11 service check

- Related ticket:https://progress.opensuse.org/issues/108623
- Verification run: http://openqa.suse.de/tests/8367498#step/install_service/108

### DIFF
--- a/lib/service_check.pm
+++ b/lib/service_check.pm
@@ -263,7 +263,8 @@ sub install_services {
     }
     # On ppc64le, sometime the console font will be distorted into pseudo graphics characters.
     # we need to reset the console font. As it impacted all the console services, added this command to bashrc file
-    systemctl('restart systemd-vconsole-setup.service') if is_ppc64le;
+    assert_script_run('echo /usr/lib/systemd/systemd-vconsole-setup >> /etc/bash.bashrc.local') if is_ppc64le;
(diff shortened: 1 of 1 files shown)


hjluo:
LGTM

lib/service_check.pm
```diff
@@ -263,7 +263,8 @@ sub install_services {
     }
     # On ppc64le, sometime the console font will be distorted into pseudo graphics characters.
     # we need to reset the console font. As it impacted all the console services, added this command to bashrc file
-    systemctl('restart systemd-vconsole-setup.service') if is_ppc64le;
+    assert_script_run('echo /usr/lib/systemd/systemd-vconsole-setup >> /etc/bash.bashrc.local') if is_ppc64le;
+    assert_script_run '. /etc/bash.bashrc.local';
```
coolgw, line 267: what's the enable status of systemd-vconsole-setup.service? It should work is service is enabled after system reboot, otherwise it should be a bug.

tinawang123: OK, I will try

tinawang123: Cannot work will report one bug

lemon-suse (approved):

PR #14546, opened 2022-03-22
Code enhance for new micro-arch add

XGWang0:
Code enhance for new micro-arch add

- Verification run: http://openqa.qa2.suse.asia/group_overview/44


XGWang0:
This is code style verification PR,  remove it after done

PR #14550, opened 2022-03-22
Separate publiccloud variables

grisu48:
Separate publiccloud-related variables from the sea of variables to
increase the overview.

### DIFF
--- a/variables.md
+++ b/variables.md
@@ -6,6 +6,10 @@ Please, find [os-autoinst backend variables](https://github.com/os-autoinst/os-a
 
 NOTE: This list is not complete and may contain outdated info. If you face such a case, please, create pull request with required changes.
 
+For a better overview some domain-specific values have been moved to their own section:
+
+* [Publiccloud](#publiccloud-specific-variables)
+
 Variable        | Type      | Default value | Details
 ---             | ---       | ---           | ---
 AARCH64_MTE_SUPPORTED | boolean | false     | Set to 1 if your machine supports Memory Tagging Extension (MTE)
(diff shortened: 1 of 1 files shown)


variables.md
```diff
@@ -260,3 +199,75 @@ LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are bootin
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
 PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in CSP
 PREPARE_TEST_DATA_TIMEOUT | integer | 300 | Download assets in the prepare_test_data module timeout
+
+### Publiccloud specific variables
+
+The following variables are relevant for publiccloud related jobs. Keep in mind that variables that start with `_SECRET` are secret variables, visible only to the job while running but otherwise not exposed and e.g. not present in cloned jobs.
```
pdostal, line 205: I don't understand the 'not present in cloned jobs' - maybe
```suggestion
The following variables are relevant for publiccloud related jobs. Keep in mind that variables that start with `_SECRET` are secret variables, visible only to the job while running but otherwise not exposed and e.g. not present in Settings and variables.json.
```

jlausuch: `visible only to the job while running` ? Really? in settings tab? then we have a problem... they should be really hidden ALL the times.

grisu48: If you clone a job, then the `_SECRET` variables are not present in the cloned one, right?

grisu48: Thanks for the input. Please read again, I updated this paragraph to make it clearer that those variables are

* Accessible by the job
* Invisible from the webui
* Not present in cloned jobs

I hope this makes it better now.

okurz (approved):

pdostal (approved):

variables.md
jlausuch: Much better now :)

asmorodskyi:
wow I didn't imagine that list is sooo huge 0_o

PR #14555, opened 2022-03-22
Tweak is_leap: GNOME-Next can falsy trigger to be 'is_leap' with GNOME 42.0

DimStar77:
GNOME 42.0 is out. While testing it as part of GNOME-Next, we have seen very
obscure failures, which were not the case up to the release candidate

Turns out we managed to confuse openQA with VERSION=42.0 and various
places checking is_leap('<15.1') and the like. With 42.rc is_leap did not
yet interpret this as Leap based, but 42.0 matches close enough that we are
considered Leap now.

- Related ticket: Non filed - debugged directly by scratching my head
- Needles: N/A
- Verification run: https://openqa.opensuse.org/t2258899 (important step is passing over consoletest_finsih)

### DIFF
--- a/lib/version_utils.pm
+++ b/lib/version_utils.pm
@@ -272,6 +272,9 @@ sub is_leap {
     # Leap and its stagings
     return 0 unless check_var('DISTRI', 'opensuse');
     return 0 unless $version =~ /^\d{2,}\.\d/ || $version =~ /^Jump/;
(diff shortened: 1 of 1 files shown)


foursixnine (approved):

Vogtinator:
You could also just do it like Argon/Krypton and use the distro's version as `VERSION`. I don't think the test code cares about the GNOME version at all.

PR #14405, opened 2022-03-02
Test for zypper single transaction

asdil12:
See: https://lists.opensuse.org/archives/list/zypp-devel@lists.opensuse.org/thread/YEFSZZY7FRGJPGPWKNCUT3UXQWVENOCL/
Ticket: https://progress.opensuse.org/issues/105948

PR #14557, opened 2022-03-22
Don't run ZYPP_SINGLE_RPMTRANS tests on OFFLINE_SUT

asdil12:
Without network access we can't download the RPMs

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14405#discussion_r832043713
https://openqa.suse.de/t8369805


ge0r (approved):
Thanks!

PR #14558, opened 2022-03-22
Increase timeout checks for SUSEConnect --rollback

mloviska:
- Related ticket: https://progress.opensuse.org/issues/108064
- Verification run: [sle-15-SP2-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220322-1-jeos-filesystem@svirt-xen-hvm](http://kepler.suse.cz/tests/14636#step/snapper_jeos_cli/58)

PR #14561, opened 2022-03-23
Wait longer for rollback registration checks

mloviska:
Follow up for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14558


jlausuch (commented):
LGTM

PR #14568, opened 2022-03-24
Fix SUSEConnect timeout on system_prepare

lemon-suse:
Add die_on_timeout=0 to make script_run won't report fail when timeout. 

- Related ticket: https://progress.opensuse.org/issues/106954
- Needles: N/A
- Verification run:  
  http://openqa.nue.suse.com/tests/8372921#step/system_prepare/21
  https://openqa.nue.suse.com/tests/8372921#step/system_prepare/21
  http://openqa.nue.suse.com/tests/8372671#step/system_prepare/21
  https://openqa.nue.suse.com/tests/8372921#step/system_prepare/21
  https://openqa.nue.suse.com/tests/8372921#step/system_prepare/21

### DIFF
--- a/lib/services/registered_addons.pm
+++ b/lib/services/registered_addons.pm
@@ -32,7 +32,7 @@ sub check_registered_system {
     $pro = 'SLE_' . $pro if ($pro eq 'HPC');
     suseconnect_ls($pro);
     my $ver = $system =~ s/\-SP/./r;
-    script_run("SUSEConnect -s | grep " . $ver);
+    script_run("SUSEConnect -s | grep " . $ver, die_on_timeout => 0);
 }
 
 sub check_registered_addons {


tinawang123 (approved):
LGTM

lib/services/registered_addons.pm
```diff
@@ -32,7 +32,7 @@ sub check_registered_system {
     $pro = 'SLE_' . $pro if ($pro eq 'HPC');
     suseconnect_ls($pro);
     my $ver = $system =~ s/\-SP/./r;
-    script_run("SUSEConnect -s | grep " . $ver);
+    script_run("SUSEConnect -s | grep " . $ver, die_on_timeout => 0);
```
coolgw, line 35: I suppose this should be assert_script_run? 
Add this parameter can fix ticket or not?

lemon-suse: We use script_run originally will take this run as not important test, that's means we don't care the return value of this line and don't want to stop here and hope to continue run other test. Add die_on_timeout => 0 just make script_run to work as original expectation.

chcao (approved):
LGTM

PR #14571, opened 2022-03-24
Clear screen after ncurses to prevent cluttering

asdil12:
Prevent something like https://openqa.suse.de/tests/8383000#step/rpi_wifi/27

### DIFF
--- a/tests/jeos/rpi_wifi.pm
+++ b/tests/jeos/rpi_wifi.pm
@@ -40,6 +40,8 @@ sub run {
     enter_cmd(get_required_var('RPI_WIFI_PSK'));
 
     assert_screen 'text-logged-in-root';
+    script_run 'clear';
+
     assert_script_run 'ip a';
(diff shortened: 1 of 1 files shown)


tests/jeos/rpi_wifi.pm
```diff
@@ -40,6 +40,8 @@ sub run {
     enter_cmd(get_required_var('RPI_WIFI_PSK'));
 
     assert_screen 'text-logged-in-root';
+    script_run 'clear';
```
okurz, line 43: That should be better `enter_cmd`

asdil12: why not check if it returned?

okurz: You are ignoring the result code anyway, you wouldn't loose anything

asdil12: I ignore the result code but not the fact that it returned.
This prevents a race condition as it really waits until then before the next command is typed.

PR #14573, opened 2022-03-24
Add leap-micro product

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/107872
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/765

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -58,6 +58,7 @@
         "description": "Failed unmounting \/\\S+\\.|-- Reboot --|pam_systemd.*Failed to release session",
         "products": {
             "sle-micro": [ "5.0", "5.1", "5.2" ],
+            "leap-micro": [ "5.2" ],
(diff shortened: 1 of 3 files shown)


Vogtinator:
Any VRs? I don't think adding the link on its own has much value.

jlausuch:
> Any VRs? I don't think adding the link on its own has much value.

It's a start :)
See https://openqa.opensuse.org/tests/2262025/logfile?filename=worker-log.txt

Not sure how to add a VR without creating that link manually in O3...

jlausuch:
I have some VRs for the image, which should be enough for this PR:
http://fromm.arch.suse.de/tests/overview?distri=leap-micro&version=5.2&build=_&groupid=53

I even created a bug for RT: https://bugzilla.opensuse.org/show_bug.cgi?id=1197666

There are more issues that will be tackled later on, but let's first enable this new project.

Vogtinator (approved):

PR #14575, opened 2022-03-24
Fix opensuse_welcome in upgrade_Leap

punkioudi:
- Related ticket: https://progress.opensuse.org/issues/107620
- Verification run: pending


tests/installation/opensuse_welcome.pm
```diff
@@ -15,7 +15,20 @@ use utils;
 use x11utils 'handle_welcome_screen';
 
 sub run {
-    handle_welcome_screen;
+    my $upgrade = get_var('UPGRADE');
+    if ($upgrade == 1) {
+        send_key 'esc';
```
Vogtinator, line 20: What does this escape do?

lib/x11utils.pm
```diff
@@ -511,7 +511,7 @@ Also handle workarounds when needed.
 =cut
 sub handle_welcome_screen {
     my (%args) = @_;
-    assert_screen([qw(opensuse-welcome opensuse-welcome-boo1169203 opensuse-welcome-gnome40-activities)], $args{timeout});
+    assert_screen([qw(opensuse-welcome opensuse-welcome-gnome40-activities)], $args{timeout});
     send_key 'esc' if match_has_tag('opensuse-welcome-gnome40-activities');
     workaround_broken_opensuse_welcome_window() if match_has_tag("opensuse-welcome-boo1169203");
```
Vogtinator, line 516: You removed `opensuse-welcome-boo1169203` above, so this is dead code

tests/installation/opensuse_welcome.pm
```diff
@@ -15,7 +15,20 @@ use utils;
...
+        x11_start_program('xterm');
+        my $config = script_run("ls .config/autostart | grep org.opensuse.opensuse_welcome.desktop");
```
Vogtinator, line 22: Sounds like `test -f` is more appropriate

tests/installation/opensuse_welcome.pm
punkioudi: I'm working on that, in order to avoid this situation : https://openqa.opensuse.org/tests/2264087#step/opensuse_welcome/4

Vogtinator:
What this test does is check whether opensuse-welcome was already disabled in `~/.config`. IMO it's better to have openQA check that what the system state is is actually as expected, i.e. opensuse-welcome is visible after new installations of 15.3+ and upgrades of < 15.3 but not anywhere else?

tests/installation/opensuse_welcome.pm
punkioudi: Or both actually, sth like:
`test -f .config/autostart` -> file is there
`ls .config/autostart | grep org.opensuse.opensuse_welcome.desktop` -> org.opensuse.opensuse_welcome.desktop is not mispelled for example

Vogtinator: I meant `test -f ~/.config/autostart/org.opensuse.opensuse_welcome.desktop`

tests/installation/opensuse_welcome.pm
Vogtinator: That's for GNOME activities handling, the `opensuse-welcome-gnome40-activities` tag is used to detect that and there's also `handle_gnome_activities`.

punkioudi:
> What this test does is check whether opensuse-welcome was already disabled in `~/.config`. IMO it's better to have openQA check that what the system state is is actually as expected, i.e. opensuse-welcome is visible after new installations of 15.3+ and upgrades of < 15.3 but not anywhere else?

So we had a discussion with @foursixnine and decided to go with this implementation (I haven't yet added descriptions etc), for covering possible changes in the future and not just comparing the different leap versions.

lib/x11utils.pm
punkioudi: Ah sorry I didn't understand you. My intention here was to avoid this wrong soft-fail imo : https://openqa.opensuse.org/tests/2255543#step/opensuse_welcome/1

Vogtinator: In the upgrade case it's definitely wrong, but is the referenced bug actually gone?

Vogtinator:
> > What this test does is check whether opensuse-welcome was already disabled in `~/.config`. IMO it's better to have openQA check that what the system state is is actually as expected, i.e. opensuse-welcome is visible after new installations of 15.3+ and upgrades of < 15.3 but not anywhere else?
> 
> So we had a discussion with @foursixnine and decided to go with this implementation (I haven't yet added descriptions etc), for covering possible changes in the future and not just comparing the different leap versions.

Ok.

Vogtinator:
Actually, if you don't care that the opensuse-welcome state is verified by openQA, this can be done in a simpler way by calling `handle_welcome_screen` only if it appears. This makes working around boo#1169203 impossible as it's not actually known whether it should show up, but that's also true for the PR currently.

```
if (get_var('UPGRADE')) {
    my @tags = qw(generic-desktop opensuse-welcome);
    push(@tags, qw(gnome-activities opensuse-welcome-gnome40-activities)) if check_var('DESKTOP', 'gnome');
    assert_screen \@tags;
    if (match_has_tag('opensuse-welcome') || match_has_tag('opensuse-welcome-gnome40-activities')) {
        handle_welcome_screen;
    }
}
```

Vogtinator:
> Actually, if you don't care that the opensuse-welcome state is verified by openQA, this can be done in a simpler way by calling `handle_welcome_screen` only if it appears. This makes working around boo#1169203 impossible as it's not actually known whether it should show up, but that's also true for the PR currently.
> 
> ```
> if (get_var('UPGRADE')) {
>     my @tags = qw(generic-desktop opensuse-welcome);
>     push(@tags, qw(gnome-activities opensuse-welcome-gnome40-activities)) if check_var('DESKTOP', 'gnome');
>     assert_screen \@tags;
>     if (match_has_tag('opensuse-welcome') || match_has_tag('opensuse-welcome-gnome40-activities')) {
>         handle_welcome_screen;
>     }
> }
> ```

I'm not sure whether there's the possibility of a race condition here that `generic-desktop` is matched by accident because opensuse-welcome hasn't started yet. FWICT it shows up in `first_boot` reliably, but that might just be luck.

lib/x11utils.pm
punkioudi: As the bug issue has never been reported again since October of 2020, I don't find the reason that we should keep this workaround in the code.

punkioudi:
> > Actually, if you don't care that the opensuse-welcome state is verified by openQA, this can be done in a simpler way by calling `handle_welcome_screen` only if it appears. This makes working around boo#1169203 impossible as it's not actually known whether it should show up, but that's also true for the PR currently.
> > ```
> > if (get_var('UPGRADE')) {
> >     my @tags = qw(generic-desktop opensuse-welcome);
> >     push(@tags, qw(gnome-activities opensuse-welcome-gnome40-activities)) if check_var('DESKTOP', 'gnome');
> >     assert_screen \@tags;
> >     if (match_has_tag('opensuse-welcome') || match_has_tag('opensuse-welcome-gnome40-activities')) {
> >         handle_welcome_screen;
> >     }
> > }
> > ```
> 
> I'm not sure whether there's the possibility of a race condition here that `generic-desktop` is matched by accident because opensuse-welcome hasn't started yet. FWICT it shows up in `first_boot` reliably, but that might just be luck.

Tbh, I don't know either. In my mind, checking the files is more robust than checking the needles, but I can do some runs also with your proposal.

lib/x11utils.pm
Vogtinator: Agreed. Can you remove all the relevant code, preferably in a separate commit?

Vogtinator:
> > > Actually, if you don't care that the opensuse-welcome state is verified by openQA, this can be done in a simpler way by calling `handle_welcome_screen` only if it appears. This makes working around boo#1169203 impossible as it's not actually known whether it should show up, but that's also true for the PR currently.
> > > ```
> > > if (get_var('UPGRADE')) {
> > >     my @tags = qw(generic-desktop opensuse-welcome);
> > >     push(@tags, qw(gnome-activities opensuse-welcome-gnome40-activities)) if check_var('DESKTOP', 'gnome');
> > >     assert_screen \@tags;
> > >     if (match_has_tag('opensuse-welcome') || match_has_tag('opensuse-welcome-gnome40-activities')) {
> > >         handle_welcome_screen;
> > >     }
> > > }
> > > ```
> > 
> > 
> > I'm not sure whether there's the possibility of a race condition here that `generic-desktop` is matched by accident because opensuse-welcome hasn't started yet. FWICT it shows up in `first_boot` reliably, but that might just be luck.
> 
> Tbh, I don't know either. In my mind, checking the files is more robust than checking the needles, but I can do some runs also with your proposal.

I don't mind either. I'm wondering how reliable it is to start xterm that way without having a match for `generic-desktop` or `opensuse-welcome`. That it needs GNOME activities handling makes it a bit awkward already.

foursixnine (approved):

Vogtinator (approved):

DimStar77:
https://openqa.opensuse.org/tests/2284558

That seems to do more than just 'UPGRADE=1' case.. in normal cases, opensuse-welcome is simply a NOP module now and results in opensuse-welcome not being closed.

Vogtinator:
Indeed, the `else` case is missing...

punkioudi:
I am really sorry for that :( Thanks @DimStar77 for identifying and reverting it quickly.
I will push the correct code with the else condition tomorrow morning.

PR #14576, opened 2022-03-24
MicroOS: enable docker testing

jlausuch:


lib/main_containers.pm
```diff
@@ -95,27 +95,31 @@ sub load_host_tests_podman {
...
+        # PackageHub is not available in SLE Micro
         loadtest 'containers/registry' if is_x86_64;
         loadtest 'containers/docker_compose';
     }
     # works currently only for x86_64, more are coming (poo#103977)
     # Expected to work for all but JeOS on 15sp4 after
     # https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13860
     # Disabled on svirt backends (VMWare, Hyper-V and XEN) as the device name might be different than vdX
-    loadtest 'containers/validate_btrfs' if (is_x86_64 and is_qemu and !is_openstack);
+    if ((is_x86_64 && is_qemu) && !(is_openstack || is_sle_micro)) {
```
Vogtinator, line 120: Why?

mloviska:
[systemd-coredump](https://openqa.suse.de/tests/8386573#step/coredump_collect/3) is not pre-installed either. Maybe we can skip this module too

lib/main_containers.pm
jlausuch: It requires an additional disk, we already have HDD_1 and HDD_2 for SLEM/MicroOS.. We are already running this test in general TW/SLE container host tests... I don't think it's worth it here.

mloviska: The additional disk can be solved with `NUMDISKS=3`, unfortunately the test has hardcoded it will use the second one (vdb). I guess, we can work on that later if there is an interest to include this module in MicroOS/SLEM testing. WDYT?

jlausuch: Yes, that was another reason I decided to omit it for now (it was never run anyways).

mloviska: I guess usage of this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/btrfs_test.pm#L22 would help in the future to make the test more versatile

products/microos/main.pm
```diff
@@ -68,12 +68,21 @@ sub load_feature_tests {
         loadtest 'console/kubeadm';
     }
     elsif (check_var 'SYSTEM_ROLE', 'container-host') {
-        my $args = OpenQA::Test::RunArgs->new();
-        my $runtime = get_required_var('CONTAINER_RUNTIME');
-        $args->{runtime} = $runtime;
+        my $args_podman = OpenQA::Test::RunArgs->new();
+        my $args_docker = OpenQA::Test::RunArgs->new();
+        $args_podman->{runtime} = 'podman';
+        $args_docker->{runtime} = 'docker';
```
pdostal, line 74: I don't understand this. Is it even related? Why are you getting rid of `CONTAINER_RUNTIME` variable?

products/microos/main.pm
```diff
@@ -68,12 +68,21 @@ sub load_feature_tests {
...
         loadtest 'microos/toolbox';
+
+        # Podman tests
         load_container_engine_test('podman');
-        loadtest('containers/image', run_args => $args, name => "image_$runtime");
+        loadtest('containers/rootless_podman');
+        loadtest('containers/image', run_args => $args_podman, name => 'image_podman');
+
+        # Docker tests
+        loadtest('microos/install_docker');
```
pdostal, line 83: Please get rid of the test module and simply modify `containers::common::install_docker_when_needed`.

jlausuch: Since it's a transactional-server specific with a reboot, I would like to leave it as an individual module which serves as a test itself, to see if it's possible to install and fail otherwise.

products/microos/main.pm
jlausuch: For MicroOS, it's a bit special. I wanted to keep it simple and run podman and docker tests in one shot regardless of that variable. We are not calling main_containers.pm in this schedule, that variable is not needed. 
But if it's confusin, I can add it back and we can do `CONTAINER_RUNTIME=podman,docker`

products/microos/main.pm
pdostal: We are already doing transaction server specific actions in the btrfs test module. There is simple task: Install docker. Now there will be two places installing docker and noone will know which of those is used where. Also when your module or any module after your module fails the SUT will restore snapshot where the docker is not yet available. This is already getting messy and too complicated. Please do not create duplicates.

products/microos/main.pm
pdostal: I would be for `CONTAINER_RUNTIME=podman,docker`. Can we also have a ticket where we move container-related logic from microos to containers?

PR #14577, opened 2022-03-24
Implement boost module test installed with spack

b10n1k:
The steps are very similar to `mpi_*` modules, with the difference of
    installing different packages which mainly differ the install path and the
    use of environment modules.
    
    The main aim in this PR is to initialize Spack testing which installs and
    loads boost module and run a hello_world. Testing includes other spack
    features like uninstall and unload modules.
    
    The steps in the module has been testing only on sle15sp4. The plan is to
    merge those modules with the mpi_*. So the modules will subject to some
    refactoring to provide a better design and integrate more modules to test.
    
    Also the jobs uses 1G memory. Note that in the manual testing this reproduced
    some issues, and as such if the jobs start showing defects during the
    compilation by spack, it is recommended to increase QEMURAM to 4.

- Verification run: http://aquarius.suse.cz/tests/9289

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -219,4 +219,40 @@ sub prepare_user_and_group {
     assert_script_run('useradd -u 7777 -g 7777 slurm');
 }
 
+=head2 prepare_spack_env
+
+  prepare_spack_env($mpi)
+
+After install spack and HPC C<mpi> required packages, prepares env
+variables. The HPC packages (*-gnu-hpc) use an installation path that is separate
+from the rest and can be exported via a network file system.
+
+After C<prepare_spack_env> run, C<spack> should be ready to build entire tool stack,
+downloading and installing all bits required for whatever package or compiler.
+=cut
+sub prepare_spack_env {
+    my ($self, $mpi) = @_;
+    $mpi //= 'mpich';
+    zypper_call "in spack $mpi-gnu-hpc $mpi-gnu-hpc-devel";
+    $self->relogin_root;
+    assert_script_run 'source /etc/profile.d/lmod.sh';
+    assert_script_run 'module load gnu $mpi';
+    assert_script_run '/usr/lib/spack/run-find-external.sh';
+    assert_script_run 'source /usr/share/spack/setup-env.sh';
+}
+
+=head2 uninstall_spack_module
+
+  uninstall_spack_module($module)
+
+Unload and uninstall C<module> from spack stack
+=cut
+sub uninstall_spack_module {
+    my ($self, $module) = @_;
+    die 'uninstall_spack_module requires a module name' unless $module;
+    assert_script_run("spack unload $module");
+    script_run('module av');
+    assert_script_run("spack uninstall -y $module");
+    assert_script_run("spack find $module | grep 'No package matches the query'");
+}
 1;
--- a/tests/hpc/spack_master.pm
+++ b/tests/hpc/spack_master.pm
@@ -1,0 +1,82 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Summary: HPC_Module: Test Spack package installation and features
+#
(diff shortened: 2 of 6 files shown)


tests/hpc/spack_master.pm
```diff
@@ -0,0 +1,82 @@
...
+    $self->check_nodes_availability();
+
+    record_info('INFO', script_output('cat /proc/cpuinfo'));
+
+    my $hostname = get_var('HOSTNAME', 'susetest');
+    record_info "hostname", "$hostname";
+    assert_script_run "hostnamectl status|grep $hostname";
+
+    assert_script_run("wget --quiet " . data_url("hpc/$mpi_c") . " -O /tmp/$mpi_c");
+    assert_script_run("$mpi_compiler /tmp/$mpi_c -o /tmp/$mpi_bin -l boost_mpi -I \${BOOST_ROOT}/include/ -L \${BOOST_ROOT}/lib 2>&1 > /tmp/make.out");
```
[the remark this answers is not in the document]
b10n1k: And this is exactly what is happening. from http://aquarius.suse.cz/tests/9289/logfile?filename=serial_terminal.txt
```
# mpic++ /tmp/sample_boost.cpp -o /tmp/mpi_bin -l boost_mpi -I ${BOOST_ROOT}/include/ -L ${BOOST_ROOT}/lib 2>&1 > /tmp/make.out; echo iNKeY-$?-
iNKeY-0-
```
and then 
```
# mpirun /tmp/mpi_bin; echo 4FtSC-$?-
I am process 0 of 1.
4FtSC-0-
# mpirun -n 2 --host master-node00,slave-node00,slave-node01 /tmp/mpi_bin; echo qExF5-$?-
I am process 0 of 2.
I am process 1 of 2.
qExF5-0-
```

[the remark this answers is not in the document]
schlad: ah yeah the $mpi_c is a bit misleading then but yeah I see now:
wget --quiet http://10.0.2.2:20023/ChQwxllU3VllqEqT/data/hpc/sample_boost.cpp -O /tmp/sample_boost.cpp; echo 0bpZ8-$?-

schlad (approved):

lib/hpcbase.pm
```diff
@@ -219,4 +219,40 @@ sub prepare_user_and_group {
...
+
+After C<prepare_spack_env> run, C<spack> should be ready to build entire tool stack,
+downloading and installing all bits required for whatever package or compiler.
+=cut
+sub prepare_spack_env {
+    my ($self, $mpi) = @_;
+    $mpi //= 'mpich';
+    zypper_call "in spack $mpi-gnu-hpc $mpi-gnu-hpc-devel";
+    $self->relogin_root;
+    assert_script_run 'source /etc/profile.d/lmod.sh';
```
e4t, line 238: After a relogin you should not have to source /etc/profile.d/lmod.sh. If you do, something would be wrong.

tests/hpc/spack_master.pm
```diff
@@ -0,0 +1,82 @@
...
+
+    ## distribute the binary
+    foreach (@cluster_nodes) {
+        assert_script_run("scp -o StrictHostKeyChecking=no /tmp/$mpi_bin root\@$_\:/tmp/$mpi_bin");
```
e4t, line 47: While this should work, I'm not a great fan of this. The directory where this is located should really be shared.

e4t (commented):
It looks fine. I still think a bit of time should be invested to get some code in place to set up a common infrastructure - like an NFS-exported home on one node. This can be reused for other use cases. There is no immediate need to do this now, though.

lib/hpcbase.pm
b10n1k: should this be before relogin?

tests/hpc/spack_master.pm
b10n1k: i am planning to implement this in the near future. Which will simplify things for the compute nodes too as we discussed

PR #14578, opened 2022-03-24
wicked: Add LTSS for autoyast SLE-12

cfconrad:
- Verification run: http://openqa.wicked.suse.de/t53304


asmorodskyi:
what about aarch64 ?

asmorodskyi:
ah I assume we don't have LTSS for aarch64 ...

cfconrad:
> ah I assume we don't have LTSS for aarch64 ...

I had only x86_64 in my focus as we do not run aarch64 in wicked-ci... thus cannot add something more valuable for it, sorry!

PR #14581, opened 2022-03-25
Extract 'clamav' from 'extra_tests_textmode' to meet the RAM req

pablo-herranz:
As the vendor states, `clamav` needs at least 2GB of RAM to work smooth. To avoid interference and overload the **openQA**, the test is extracted from its original location and executed on its own dedicated test suite.

- Related ticket: https://progress.opensuse.org/issues/108082
- Needles:
- Verification runs:

|Distro |x86_64 |aarch64 |ppc64le |s390x |
|:---------------|:----------:|:-----------:|:-----------:|:---------:|
|Tumbleweed  |[:heavy_check_mark:](https://openqa.opensuse.org/t2264780) |[:heavy_check_mark:](https://openqa.opensuse.org/t2264782) |- |- |
|Leap 15.4   |[:heavy_check_mark:](https://openqa.opensuse.org/t2264779) |- |- |- |
|SLE 15 SP4  |[:heavy_check_mark:](https://openqa.suse.de/t8393542) |[:heavy_check_mark:](https://openqa.suse.de/t8393543) |[:heavy_check_mark:](https://openqa.suse.de/t8393545) |- |

### DIFF
--- a/schedule/functional/extra_tests_clamav.yaml
+++ b/schedule/functional/extra_tests_clamav.yaml
@@ -1,0 +1,10 @@
+name: extra_tests_clamav
+description:    >
+    Maintainer: QE Core.
+    Extract 'clamav' module from 'extra_tests_textmode' to meet the RAM
+    requirements.
+schedule:
+    - installation/bootloader_start
+    - boot/boot_to_desktop
+    - console/clamav
+    - console/coredump_collect
--- a/schedule/functional/extra_tests_textmode.yaml
+++ b/schedule/functional/extra_tests_textmode.yaml
@@ -135,7 +135,6 @@ schedule:
     - console/redis
     - console/rsync
     - console/rust
-    - console/clamav
     - console/shells
     - console/sudo
     - console/dstat


schedule/functional/extra_tests_clamav.yaml
```diff
@@ -0,0 +1,11 @@
...
+description:    >
+    Maintainer: QE Core.
+    Extract 'clamav' module from 'extra_tests_textmode' to meet the RAM
+    requirements.
+schedule:
+    - installation/bootloader_start
+    - boot/boot_to_desktop
+    - console/consoletest_setup
+    - console/clamav
+    - console/coredump_collect
```
Avinesh, line 11: Just curious to know why `coredump_collect` test is also included?

pablo-herranz: I thought it was necessary in the end of all testsuites :sweat_smile:

pablo-herranz:
> 

By now, I'm sticking to solve the issue, which happened in `SLE15 SP4 - ppc64le`, but my intention is to create a new separate ticket to solve it in maintenance too.

schedule/functional/extra_tests_clamav.yaml
punkioudi: I think that you should keep the coredump_collect test, here :) It can probably give some coredumps that are not visible during the test.

punkioudi: On the other hand, I don't see the need of `consoletest_setup`, especially if it has already run in a previous test.

pablo-herranz: Seems legit :)
https://openqa.suse.de/tests/8393452

punkioudi (approved):
Lgtm! If you can add the VRs with the new changes and then I'll merge :)

PR #14582, opened 2022-03-25
Add delay for ssh

jesusbv:


pdostal (changes requested):
This in my opinion does not make any sense as there is already `wait_for_ssh` subroutine.

asmorodskyi:
> This in my opinion does not make any sense as there is already `wait_for_ssh` subroutine.

exactly . and if you have timeouts you need just increase waiting time in `wait_for_ssh`

jesusbv:
> > This in my opinion does not make any sense as there is already `wait_for_ssh` subroutine.
> 
> exactly . and if you have timeouts you need just increase waiting time in `wait_for_ssh`

Ack, feel free to close the PR
I was told it would be a good improvement for the ssh softreboot

lib/publiccloud/instance.pm
```diff
@@ -287,6 +288,7 @@ sub softreboot
     sleep 60;    # wait for the +1 in the previous command
     my $start_time = time();
 
+    sleep($args{delay});
```
grisu48, line 291: This is difficult. This routine is also used for measuring of boottime (see the line before), so adding an additional delay here might interfere with the boottime test. We could update line 288 though and state, that if delay is set, this subroutine might not return the accurate boottime.

PR #14589, opened 2022-03-28
Containers Image tests: refactor host configuration

jlausuch:
I would like to enable Host config module in all hosts to perform the following actions:
- System update: so we test the latest state of the hosts as   users would do
- Install docker|podman if they are not pre-installed.
- in SLE, retart firewall and docker service. For some reason, there were some failures with containers not reaching internet. After some debugging, restarting the services was the solution.

Example of issues:
https://openqa.suse.de/tests/8400211
https://openqa.suse.de/tests/8400221
https://openqa.suse.de/tests/8400216
Even though the issue is not apparent in the test, after some analysis, the problem is that the containers can't reach internet.

- Verification run: 
[bci-python on all hosts](http://fromm.arch.suse.de/tests/overview?distri=sle&build=13.30_python-3.9-image&version=15-SP3)
[base image on all hosts](http://fromm.arch.suse.de/tests/overview?build=17.11.11&version=15-SP3&distri=sle)

### DIFF
--- a/tests/containers/host_configuration.pm
+++ b/tests/containers/host_configuration.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2020-2021 SUSE LLC
+# Copyright 2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: Setup system which will host containers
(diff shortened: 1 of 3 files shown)


tests/containers/host_configuration.pm
```diff
@@ -12,26 +12,53 @@
...
+        } elsif ($host_distri eq 'rhel') {
+            assert_script_run("yum update -y", timeout => $update_timeout);
+        }
+    }
+
+    # Make sure we can access internet and DNS works
+    script_retry('ping -c 3 www.google.com');
+
+    # Install engines in case they are not installed
+    install_docker_when_needed($host_distri) if ($engine =~ 'docker');
```
ilausuch, line 48: using -eq instead of =~? only a suggestion :)

ilausuch (approved):

tests/containers/host_configuration.pm
jlausuch: no, because `$engine` can take as value `docker,podman`

ilausuch: Oh yes! thank you

pdostal:
> Install docker|podman if they are not pre-installed.
> in SLE, retart firewall and docker service. For some reason, there were some failures with containers not reaching internet. After some debugging, restarting the services was the solution.

Those two things are done in `containers::basetest::containers_factory`. Please keep it there.

jlausuch:
> > Install docker|podman if they are not pre-installed.
> > in SLE, retart firewall and docker service. For some reason, there were some failures with containers not reaching internet. After some debugging, restarting the services was the solution.
> 
> Those two things are done in `containers::basetest::containers_factory`. Please keep it there.

Yes and no. Some hosts come with docker and podman pre-installed, some others don't.
The `init` method of docker will call `install_docker_when_needed` which will skip the firewall restart part if docker is already installed.

What I want to make sure is to restart the firewall after system update in SLE cases where I have observed that error.

pdostal (approved):

tests/containers/host_configuration.pm
```diff
@@ -12,26 +12,53 @@
...
-        disable_and_stop_service(opensusebasetest::firewall, ignore_failure => 1);
     }
     else {
-        # Re-running dhclient on RHEL is confusing the routing tables
-        assert_script_run "dhclient -v" unless get_var("HDD_1") =~ /rhel/;
-        $interface = script_output q@ip r s default | head -1 | awk '{printf $5}'@;
-        validate_script_output "ip a s '$interface'", sub { m/((\d{1,3}\.){3}\d{1,3}\/\d{1,2})/ };
-        assert_script_run "curl http://ca.suse.de/certificates/ca/SUSE_Trust_Root.crt -o /etc/ssl/certs/SUSE_Trust_Root.crt" if is_sle();
+        if ($host_distri eq 'ubuntu') {
+            assert_script_run("dhclient -v");
```
grisu48, line 34: If we know why, please add a comment why we need to run this here and also for CentOS.

grisu48 (commented):
Small comment, otherwise LGTM

tests/containers/host_configuration.pm
jlausuch: Done. It has some history, and maybe not needed any more...  I would like to try again if it works without this and remove it if so, but in separate PR.

grisu48 (approved):

DimStar77:
Fallouts on Tumbleweed (collecting; probably more to come)

* https://openqa.opensuse.org/tests/2270207

jlausuch:
> Fallouts on Tumbleweed (collecting; probably more to come)
> 
>     * https://openqa.opensuse.org/tests/2270207
> 
>     * https://openqa.opensuse.org/tests/2269975
> 
>     * https://openqa.opensuse.org/tests/2269976

on it!

PR #14590, opened 2022-03-28
Add helm test for Tumbleweed

ilausuch:
This new test checks the main operations of helm, and deploys
an apache on EKS

- Related ticket: https://progress.opensuse.org/issues/108572
- Verification run: http://copland.arch.suse.de/tests/1276 (Tumbleweed)
- http://copland.arch.suse.de/tests/1275 (Leap 15.3)

### DIFF
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -1,0 +1,71 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Test deploy a helm chart
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use Mojo::Base 'publiccloud::basetest';
+use testapi;
+use utils qw(zypper_call script_retry);
+use version_utils qw(is_sle);
+use registration qw(add_suseconnect_product get_addon_fullname);
+use mmapi 'get_current_job_id';
+
+sub run {
+    my ($self) = @_;
+
+    $self->select_serial_terminal;
+    $self->{deployment_name} = "apache-" . get_current_job_id();
+    my $chart = "bitnami/apache";
+
+    zypper_call("in -y jq aws-cli helm kubernetes-client");
+
+    # Check helm version
+    record_info('Helm version', script_output("helm version"));
+
+    # Connection to AWS
+    my $provider = $self->provider_factory(service => 'EKS');
+    $self->{provider} = $provider;
+
+    # Access to Cluster
+    record_info('Kubectl version', script_output("kubectl version"));
+    assert_script_run("cat \$HOME/.kube/config");
+    assert_script_run("kubectl get nodes");
+
+    # Add repo, search and show values
+    assert_script_run("helm repo add bitnami https://charts.bitnami.com/bitnami");
+    assert_script_run("helm repo update");
+    assert_script_run("helm search repo apache");
+    assert_script_run("helm show all $chart");
+
+    # Install apache
+    assert_script_run("helm install $self->{deployment_name} $chart");
+    assert_script_run("helm list");
+
+    # Wait for deployment
+    assert_script_run("kubectl rollout status deployment/$self->{deployment_name}");
+    assert_script_run("kubectl describe deployment/$self->{deployment_name}");
+    assert_script_run("kubectl get pods | grep $self->{deployment_name}");
+    assert_script_run("kubectl describe services/$self->{deployment_name}");
+
+    # Test
+    enter_cmd("kubectl port-forward  services/$self->{deployment_name} 10001:80 &");
+    script_retry("curl http://localhost:10001", delay => 30, retry => 5);
+
+    # Destroy the chart
+    assert_script_run("helm delete $self->{deployment_name}");
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    script_run("helm delete $self->{deployment_name}");
+}
+
+sub test_flags {
+    return {fatal => 1, milestone => 1};
+}
+
+1;
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -126,6 +126,12 @@ sub load_host_tests_containerd_nerdctl {
     loadtest 'containers/containerd_nerdctl';
 }
 
+sub load_host_tests_helm() {
+    if (is_tumbleweed || is_leap) {
+        loadtest "containers/helm";
+    }
+}
+
 sub load_container_tests {
     my $args = OpenQA::Test::RunArgs->new();
     my $runtime = get_required_var('CONTAINER_RUNTIME');
@@ -159,6 +165,7 @@ sub load_container_tests {
             load_host_tests_docker() if (/docker/i);
             load_host_tests_containerd_crictl() if (/containerd_crictl/i);
             load_host_tests_containerd_nerdctl() if (/containerd_nerdctl/i);
+            load_host_tests_helm() if (/helm/i);
         }
     }
 


pdostal (approved):
This is so easy to read, I like it, it is cool! :+1:

tests/containers/helm.pm
```diff
@@ -0,0 +1,87 @@
...
+
+use Mojo::Base 'publiccloud::basetest';
+use testapi;
+use utils qw(zypper_call);
+use version_utils qw(is_sle);
+use registration qw(add_suseconnect_product get_addon_fullname);
+
+sub install_kubectl {
+    my $version = get_var('PUBLICCLOUD_KUBECTL_VERSION', 'v1.22.2');
+    assert_script_run("curl -LO https://dl.k8s.io/release/$version/bin/linux/amd64/kubectl");
```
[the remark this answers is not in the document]
ilausuch: mmm, yesss maybe is only on sle we don't have it, in other tests we are doing this... I have to do the version for sle too, so I will save this for sle and if we have kubectl on tumbleweed let's use it. Thanks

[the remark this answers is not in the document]
ilausuch: And thanks Pavel :)

ilausuch:
I fixed the installation of kubectl 
And I changed the assert_script_run for checking the connectivity with apache. Now I am using script_retry that allows to do several retries until the service is ready

tests/containers/helm.pm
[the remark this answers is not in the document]
ilausuch: done

ilausuch:
I checked that the same code works with Leap too

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+    my ($self) = @_;
+
+    $self->select_serial_terminal;
+    $self->{deployment_name} = "apache";
+    my $chart = "bitnami/apache";
+
+    zypper_call("in -y jq aws-cli helm podman kubernetes-client");
+
+    # Check helm version
+    assert_script_run("helm version");
```
jlausuch, line 26: I guess this line is redundant as it's executed below.

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+    record_info('Helm version', script_output("helm version"));
+
+    # Connexion to AWS
```
jlausuch, line 29:
```suggestion
    # Connection to AWS
```

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+    my $provider = $self->provider_factory(service => 'EKS');
+    $self->{provider} = $provider;
+
+    # Access to Cluster
+    record_info('Kubectl version', script_output("kubectl version"));
+    assert_script_run("cat \$HOME/.kube/config");
+    assert_script_run("kubectl get nodes");
+
+    # Ensure is deleted
```
jlausuch, line 38: just curious, why deleting something that hasn't been created before?

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+
+    # Wait for deployment
+    assert_script_run("kubectl rollout status deployment/$self->{deployment_name}");
+    assert_script_run("kubectl describe deployment/$self->{deployment_name}");
+    assert_script_run("kubectl get pods | grep $self->{deployment_name}");
+    assert_script_run("kubectl describe services/$self->{deployment_name}");
+
+    # Test
+    enter_cmd("kubectl port-forward  services/$self->{deployment_name} 10001:80 &");
+    script_retry("curl http://localhost:10001", delay => 5, retry => 3);
```
jlausuch, line 58: Good idea. Since we have recently expected some network slowdown, maybe we could increase the retries and the delay. Something like `delay => 30, retry => 5`? Just to be in the safe side.

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    assert_script_run("helm delete $self->{deployment_name}");
+}
+
+sub post_run_hook {
+    my ($self) = @_;
+    assert_script_run("helm delete $self->{deployment_name}");
```
jlausuch, line 68: I would move this to the test flow instead of the post hook, as it's kind of part of the package test.

tests/containers/helm.pm
```diff
@@ -0,0 +1,75 @@
...
+    assert_script_run("kubectl describe services/$self->{deployment_name}");
+
+    # Test
+    enter_cmd("kubectl port-forward  services/$self->{deployment_name} 10001:80 &");
+    script_retry("curl http://localhost:10001", delay => 5, retry => 3);
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    assert_script_run("helm delete $self->{deployment_name}");
```
jlausuch, line 63:
```suggestion
    script_run("helm delete $self->{deployment_name}");
```
Just to not fail the test here, in case the test fails before creating the heml chart.

schedule/containers/helm.yaml
```diff
@@ -0,0 +1,7 @@
+---
+name: helm
+description: |
+  helm test
+schedule:
```
jlausuch, line 5: after a chat, agreed to use `main_containers.pm` instead of yaml schedule.

tests/containers/helm.pm
ilausuch: If the previous step of deletion has failed then this test never will start and manually we'll have to clean up the k8s. 
But there is an other solution, and that is to set up a new name every time, this allows to have some tests in parallel. Changing that

jlausuch: This is dangerous, what if 2 tests run at the same time?

tests/containers/helm.pm
ilausuch: Ok, but only for the post hook, we need to keep it for the post_fail_hook

tests/containers/helm.pm
ilausuch: mmm I show that in a post_fail_hook, this line is executed inside a post_fail_hook https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f02af1d6b443591159e95e0ca0ab330ec2f7b308/lib/publiccloud/eks.pm#L33 should we remove the assert_script_run from there too?

jlausuch: Yes, probably `script_run` fits better in these cases... I'm not 100% sure how post_fail hooks treat failures.... 
Anyway `eks.pm` can be changed in a different PR.

tests/containers/helm.pm
jlausuch: yes, the post fail is fine as it is

tests/containers/helm.pm
ilausuch: done

tests/containers/helm.pm
ilausuch: done

tests/containers/helm.pm
ilausuch: now, no problem

tests/containers/helm.pm
ilausuch: done

tests/containers/helm.pm
ilausuch: done

tests/containers/helm.pm
ilausuch: done

schedule/containers/helm.yaml
ilausuch: done

ilausuch:
I change the name of the deployment and now is dynamic
```$self->{deployment_name} = "apache-" . get_current_job_id();```

File: tests/containers/helm.pm (hunk 4 above)
Comment (jlausuch) [reply tests/containers/helm.pm:38]: if we are generating an unique deployment name, do we still need this?

File: tests/containers/helm.pm [hunk 9]
```diff
@@ -0,0 +1,73 @@
...
+    record_info('Kubectl version', script_output("kubectl version"));
+    assert_script_run("cat \$HOME/.kube/config");
+    assert_script_run("kubectl get nodes");
+
+    # Ensure is deleted
+    script_run("helm delete $self->{deployment_name}");
+
+    # Add repo, search and show values
+    assert_script_run("helm repo add bitnami https://charts.bitnami.com/bitnami");
+    assert_script_run("helm search repo apache");
```
mloviska, line 43: Just out of curiosity, should not the test try `helm repo update` before search and install ?

ilausuch: yes, we can add this. Thanks

ilausuch: done and testing

jlausuch (approved):

PR #14584, opened 2022-03-28
Add the CPU model information in fips_setup

jouyingbin:
- Related ticket: https://progress.opensuse.org/issues/108647
- Needles: NA
- Verification run:
intel : https://openqa.suse.de/t8417738 
amd : https://openqa.suse.de/tests/8417739#
note: Both intel/amd show **QEMU Virtual CPU version 2.5+**, I think this PR is only suitable for Baremetal testing.

PR #14592, opened 2022-03-28
Revert "Add the CPU model information in fips_setup"

lilyeyes:
Reverts os-autoinst/os-autoinst-distri-opensuse#14584


jouyingbin (commented):
LGTM and under discussion.

PR #14601, opened 2022-03-29
Increase timeout for dig command

rakoenig:
- on aarch64 the timeout was too short for this platfrom which  resulted in failed tests.
- Problems with scripts terminated with SIGTERM could be resolved by setting QEMURAM to 4096

- Related ticket: https://progress.opensuse.org/issues/107326
- Needles: 
- Verification run: https://openqa.suse.de/tests/8437442


jknphy:
could you please add the QEMURAM in the corresponding yaml?

PR #14604, opened 2022-03-29
Fix failing wpa_supplicant test

grisu48:
Attempt to fix the failing wpa_supplicant test on Tumbleweed by introducing two changes:

* Don't fail wpa_supplicant if an error happens during the cleanup routine (e.g. killing an already terminated process)
* The test now passes a cookie to openQA if and only if the test passes. We ignore the return code of the script and await the cookie to mark the test as successful.

- Related ticket: https://progress.opensuse.org/issues/106676
- Verification run: https://duck-norris.qam.suse.de/tests/8478#step/wpa_supplicant/38 (wpa_supplicant passes, failing in other steps)

### DIFF
--- a/data/wpa_supplicant/wpa_supplicant_test.sh
+++ b/data/wpa_supplicant/wpa_supplicant_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -ex
 
 function prepare() {
 	modprobe mac80211_hwsim radios=2
@@ -18,7 +18,7 @@ function cleanup() {
 	rm -f wifi_scan.txt networks.txt status.txt hostapd.com dnsmasq.log dhclinet.log dhclient.pid dhclient.lease
 	rm -f /etc/sysconfig/network/ifcfg-wlan1
 	wpa_cli -i wlan1 terminate >/dev/null 2>/dev/null
-	ip netns pids wifi_master | xargs kill
+	ip netns pids wifi_master | xargs kill 2> /dev/null || true        # Don't display error messages, as they are misleading
 	ip netns del wifi_master
 }
 
@@ -33,6 +33,7 @@ function GREP {
 }
 
 ## ==== Prepare environment ================================================= ##
+set -o pipefail
 trap cleanup EXIT
 set -e
 prepare
@@ -116,9 +117,12 @@ if systemctl is-active wickedd >& /dev/null; then
     echo "Restarting wicked ... "
     systemctl restart wicked
     sleep 20
-    systemctl status wicked
+    systemctl status wicked | cat
     echo "starting wicked ... "
-    wicked --debug all ifup wlan1 --timeout 60 > wicked.log 2>&1
+    # wicked ifup might terminate with setup-in-progress (rc=162), so check the link manually
+    wicked --debug ifup wlan1 --timeout 60 > wicked.log 2>&1 || true
+    sleep 30
+    wicked ifstatus wlan1 | grep 'link' | grep 'state up\|state device-up'
 else
     echo "use dhclient"
     touch dhclient.lease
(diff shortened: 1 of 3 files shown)


data/wpa_supplicant/wpa_supplicant_test.sh
```diff
@@ -116,9 +116,12 @@ if systemctl is-active wickedd >& /dev/null; then
     echo "Restarting wicked ... "
     systemctl restart wicked
     sleep 20
-    systemctl status wicked
+    systemctl status wicked | cat
```
asmorodskyi, line 119: why we are ok if wicked has failed state here ?

grisu48: Good catch, I need to set `pipefail` as well. The reason for this is that otherwise the script blocks when executed outside of openQA

grisu48: I've added `set -o pipefail`, should be ok now. Please mark as resolved if you agree

asmorodskyi:
but I would wait for @cfconrad  before merge

data/wpa_supplicant/wpa_supplicant_test.sh
```diff
@@ -116,9 +117,12 @@ if systemctl is-active wickedd >& /dev/null; then
...
     echo "starting wicked ... "
-    wicked --debug all ifup wlan1 --timeout 60 > wicked.log 2>&1
+    # wicked ifup might terminate with setup-in-progress (rc=162), so check the link manually
+    wicked --debug ifup wlan1 --timeout 60 > wicked.log 2>&1 || true
+    sleep 30
+    wicked ifstatus wlan1 | grep 'link' | grep 'state up\|state device-up'
```
cfconrad, line 125: I think we should judge from wicked perspective.

```suggestion
    wicked --log-level debug --debug all ifup wlan1 --timeout 90 2>&1 | tee wicked.log
```

cfconrad (commented):
General agree, I only would change the `wicked ifup` part as suggested, wdyt?

data/wpa_supplicant/wpa_supplicant_test.sh
grisu48: Please see the comment just above this command:

    # wicked ifup might terminate with setup-in-progress (rc=162), so check the link manually

This is one of the reasons why this test occasionally fails. AFAICS `wicked` terminates prematurely with rc=162, but the link is setup correctly. In this state the test fails because we have `-e` set.

cfconrad: yes, thats why I increased from 60 to 90 the timeout as you also do this `60 + sleep 30`. 
And I actually wonder why this failing. Even 60 seconds should be enough in all circumstances here.

grisu48: In my test runs it never waited for even the 60 seconds but terminated after just a couple of them, so increasing the timeout alone is not sufficient.

cfconrad: oh, that is something I didn't expect.  I need to check it then lets first go with your change to unblock the tests. 
Wicked wlan is tested in wicked-testsuite https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/tests/wicked/wlan/sut

cfconrad (approved):

data/wpa_supplicant/wpa_supplicant_test.sh
cfconrad: I was to fast... now I realized your words :)

```bash
wicked --debug ifup wlan1 --timeout 60
Bad debug facility "ifup"
wicked [options] <command> ...
```
This is what I get with your command. I guess it should be `--debug all` ...

ostruzina:
Hi Felix, we're seeing failures in maintenance jobs after this change since today: https://openqa.suse.de/tests/8470105#step/wpa_supplicant/29

PR #14613, opened 2022-03-31
Record whilelist normally whether has it set or not

lemon-suse:
We don't need mark red when white list hasn't been set.

- Related ticket: https://progress.opensuse.org/issues/105199
- Needles: N/A
- Verification run: http://openqa.nue.suse.com/tests/8434843#step/orphaned_packages_check/10

### DIFF
--- a/tests/console/orphaned_packages_check.pm
+++ b/tests/console/orphaned_packages_check.pm
@@ -44,8 +44,7 @@ sub compare_orphans_lists {
     # Summary
     record_info('Detected Orphans', to_string @{$args{zypper_orphans}});
     record_info('Orphans whitelisted',
-        $args{whitelist} // 'No orphans whitelisted within the test suite',
-        result => $args{whitelist} ? 'ok' : 'fail'
+        $args{whitelist} // 'No orphans whitelisted within the test suite'
     );
     record_info('Missing',
         @missed_orphans ? to_string @missed_orphans : 'None',


tinawang123 (approved):
LGTM

coolgw:
VR still how "Test died: There have been unexpected orphans detected!" Should a bug or script need adapt?

hjluo:
LGTM

lemon-suse:
> VR still how "Test died: There have been unexpected orphans detected!" Should a bug or script need adapt?

It's a bug for orphans detected, bsc#1190635

chcao (approved):
LGTM

lemon-suse:
> @lemon-suse You shouldn't ask me but the author of [9a01e92#diff-ac379a7bcf47a075f8748337554865ba83eb183961a0e01c0b810e9659bfb7fcR38](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/9a01e928302089bc97430af7814b7657dbaebe7d#diff-ac379a7bcf47a075f8748337554865ba83eb183961a0e01c0b810e9659bfb7fcR38) :)

Ok, thanks. I have requested Martin Loviska to review it. Please remove the change request then. Thanks.

mloviska:
It is only `record_info` so the initial purpose was to just visualize the different state as it does not really affect the result of the test case. But sure, we can remove it.

lemon-suse:
> It is only `record_info` so the initial purpose was to just visualize the different state as it does not really affect the result of the test case. But sure, we can remove it.

Ok, thanks for the review.

okurz (approved):

PR #14615, opened 2022-03-31
Vagrant: give boxes up to two minutes to shut down

DimStar77:
Boxes fired up using virtualbox are not the fastest of kinds when run inside qemu
To start up a box, we allow up to 1200s already. To shut down, the timeout of 30s
is too optimistic.

According the logs of e.g. https://openqa.opensuse.org/tests/2273451#step/tumbleweed/58
we can see that in …/tumbleweed/77 the serial finally received the result from the
halt command; this was approx with 1 minute delay

- Related ticket: https://progress.opensuse.org/issues/105271 (unspecific)
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/2274039 [running]


DimStar77:
Closing in favor of the even more complete https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14616

PR #14624, opened 2022-04-01
Incorrect behavior in crash test

lpalovsky:
Crash test does not follow expected result described in result file. 
Currently the test accepts node being fenced as correct behavior in 
'pacemaker-kill' test. This PR adds an error message in case fencing 
occured or node was rebooted for some reason.

- Related ticket: https://jira.suse.com/browse/TEAM-5875
- Verification run: 
  https://mordor.suse.cz/tests/3450#step/ha_cluster_crash_test/65 (test pass)
  https://mordor.suse.cz/tests/3446#step/ha_cluster_crash_test/85 (error message after node rebooted)

### DIFF
--- a/tests/ha/ha_cluster_crash_test.pm
+++ b/tests/ha/ha_cluster_crash_test.pm
@@ -45,17 +45,21 @@ sub run {
     # List of things to check
     my @checks = qw(kill-sbd kill-corosync kill-pacemakerd split-brain-iptables);
     my $preflight_start_time = time;
-    my $fencing_start_delay = calculate_sbd_start_delay();
+    # Start delay after fencing to prevent node joining cluster too quickly
+    my $start_delay_after_fencing = calculate_sbd_start_delay();
 
     # Loop on each check
     foreach my $check (@checks) {
-        # Execute the command
-        my $cmd = "crm cluster crash_test --${check} --force";
-        record_info("${check}", "Executing ${cmd}");
-        my $cmd_fails = script_run "${cmd}";
-        #record_info('ERROR', "Failure while executing '$cmd'", result => 'fail') unless (defined $cmd_fails and $cmd_fails == 0);
+        # reset fencing status form previous loop
+        $node_was_fenced = 0;
 
-        # All the commands leads to a reboot of the node
+        # Execute the command
+        my $cmd = "crm cluster crash_test --$check --force";
+        record_info($check, "Executing $cmd");
+        my $cmd_fails = script_run($cmd, die_on_timeout => 0);
+
+        # Killing pacemaker should result in service restart
+        # All remaining commands lead to a reboot of the node
         my $loop_count = bmwqemu::scale_timeout(15);    # Wait 1 minute (15*4) maximum, can be scaled with SCALE_TIMEOUT
         while (1) {
             last if ($loop_count-- <= 0);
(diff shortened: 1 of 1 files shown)


tests/ha/ha_cluster_crash_test.pm
```diff
@@ -68,12 +71,28 @@ sub run {
             }
             sleep 4;
         }
+
+        if ($check eq "kill-pacemakerd") {
+            if ($node_was_fenced == 1) {
```
ricardobranco777, line 76:
```suggestion
            if ($node_was_fenced) {
```

tests/ha/ha_cluster_crash_test.pm
```diff
@@ -68,12 +71,28 @@ sub run {
...
+                record_info('ERROR', "Failure while executing '$cmd'. Killing pacemaker should not trigger fencing", result => 'fail');
+                # Wait for fencing delay and resources to start
+                sleep $fencing_start_delay;
+                wait_until_resources_started();
```
ricardobranco777, line 80:
```suggestion
                wait_until_resources_started(timeout => $fencing_start_delay);
```

tests/ha/ha_cluster_crash_test.pm
ricardobranco777: There are other instances in the code where we have this redundant boolean check.

tests/ha/ha_cluster_crash_test.pm
ricardobranco777: Perhaps we could try `wait_until_resources_started()` didn't work for me with the `timeout` parameter.

ricardobranco777 (approved):
Just small suggestions. Otherwise LGTM

tests/ha/ha_cluster_crash_test.pm
lpalovsky: the 'sleep $fencing_start_delay; ' is a delay which happens after node is fenced and boots up. The resources are not started immediately but after this time in case the node boots quicker that cluster finishes fencing procedure. It's a dead time and 'wait_after_resources_started' should begin afterwards.

I see that 'wait_after_resources_started' has default timeout of 120s but I can do a test run for that as well. 
I realized I can move this section to the grub check loop so the code is not duplicated. 

I will do another VR and ping you later.

tests/ha/ha_cluster_crash_test.pm
lpalovsky: Thanks, I have changed all of them.

tests/ha/ha_cluster_crash_test.pm
```diff
@@ -45,35 +45,52 @@ sub run {
     # List of things to check
     my @checks = qw(kill-sbd kill-corosync kill-pacemakerd split-brain-iptables);
     my $preflight_start_time = time;
-    my $fencing_start_delay = calculate_sbd_start_delay();
+    #
```
ricardobranco777, line 48: Void comment?

tests/ha/ha_cluster_crash_test.pm
```diff
@@ -45,35 +45,52 @@ sub run {
...
     foreach my $check (@checks) {
+        # reset fencing status form previous loop
+        $node_was_fenced = 0;
+
         # Execute the command
         my $cmd = "crm cluster crash_test --${check} --force";
         record_info("${check}", "Executing ${cmd}");
-        my $cmd_fails = script_run "${cmd}";
-        #record_info('ERROR', "Failure while executing '$cmd'", result => 'fail') unless (defined $cmd_fails and $cmd_fails == 0);
+        my $cmd_fails = script_run("${cmd}", die_on_timeout => 0);
```
ricardobranco777, line 59:
```suggestion
        my $cmd_fails = script_run($cmd, die_on_timeout => 0);
```
This is not Bash ;)  Other instances of `"${cmd}"` could be simplified.

ricardobranco777 (approved):
Small suggestions. Otherwise LGTM

tests/ha/ha_cluster_crash_test.pm
lpalovsky: Oh did I delete the comment somehow ? Just added it back.

tests/ha/ha_cluster_crash_test.pm
lpalovsky: You are right, butI am innocent in this one, it was part of the old code :D
I removed it from other places as well.

tests/ha/ha_cluster_crash_test.pm
```diff
@@ -45,35 +45,52 @@ sub run {
...
+        # reset fencing status form previous loop
+        $node_was_fenced = 0;
+
         # Execute the command
-        my $cmd = "crm cluster crash_test --${check} --force";
-        record_info("${check}", "Executing ${cmd}");
-        my $cmd_fails = script_run "${cmd}";
-        #record_info('ERROR', "Failure while executing '$cmd'", result => 'fail') unless (defined $cmd_fails and $cmd_fails == 0);
+        my $cmd = "crm cluster crash_test --$check --force";
+        record_info("$check", "Executing $cmd");
```
ricardobranco777, line 58:
```suggestion
        record_info($check, "Executing $cmd");
```

michelepagot (approved):
ok

PR #14625, opened 2022-04-01
check for nvidia repo's gpg key for SLED install

zhaocongjia:
The installation fails for SLED, after Nvidia repo is enabled in SCC recently. Now there is a window for Nvidia reop's gpg key.

This window appeared in different position before, there are code to handle that by import this key.

Now this window appears at the end of test `scc_registration`, so I will adjust the test case to check for this screen and trust this key.

- Related ticket: https://progress.opensuse.org/issues/109346
- Needles: None
- Verification run: https://openqa.suse.de/tests/8453667#step/scc_registration/12

### DIFF
--- a/tests/installation/scc_registration.pm
+++ b/tests/installation/scc_registration.pm
@@ -13,7 +13,7 @@ use warnings;
 use parent "y2_installbase";
 
 use testapi;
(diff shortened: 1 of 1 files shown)


GraceWang571:
LGTM

PR #14630, opened 2022-04-01
Making transactional_update module fatal

Avinesh:
Missing curl package in sle15.4 transactional server role turned out to be a product bug which is being fixed 
(https://bugzilla.suse.com/show_bug.cgi?id=1197913)

As part of this commit, I am making the transactional_update module fatal, so no further related test are run unnecessarily.  

- Related ticket: https://progress.opensuse.org/issues/107356
- Verification run: https://openqa.opensuse.org/tests/2276517#

### DIFF
--- a/tests/transactional/transactional_update.pm
+++ b/tests/transactional/transactional_update.pm
@@ -118,7 +118,10 @@ sub run {
 }
 
 sub test_flags {
-    return {no_rollback => 1};
+    return {
+        fatal => 1,
+        no_rollback => 1
+    };
 }
 
 1;


jlausuch:
I would like to revert this PR. This module is used in a lot of tests in SLE Micro and MicroOS. In most cases we would like to continue with further tests instead of stopping the test which I think are necessary to provide feedback about the rest of the tests which are independent from transactional update.

@Vogtinator

jlausuch:
I accidentaly delete my previous comment. I was saying that I would like to revert this PR. This module is used in multiple places in SLE Micro and MicroOS and executing the rest of the tests if this fails is still valid.
@Vogtinator

Avinesh:
> 

Okay, in that case I am reverting this commit. I will discuss with @foursixnine if we need to handle it some other way.

Vogtinator:
`fatal => 1, no_rollback => 1` is a weird combination anyway, it doesn't really make sense.

PR #14631, opened 2022-04-01
Fix zgrep output for release_notes_from

rakoenig:
- PR 14536 introduced zgrep for checking the release notes. As a side
  effect when theree are more than one log file that matches the glob
  pattern the output will be prefixed by the filenmae. This moves the
  substring that the needle wants to match to the right and then we
  get matches around 70% only. Test then fails.
- [failed test release_notes_from_rpm](https://openqa.suse.de/tests/8444887#step/release_notes_from_rpm/4)
- As a bugfix we use parameter -h now too, this surpresses the output
  of filenames for matches.

- Related ticket: no ticket
- Needles: no needles
- Verification run: https://openqa.suse.de/tests/8453295#

### DIFF
--- a/tests/installation/release_notes_from_rpm.pm
+++ b/tests/installation/release_notes_from_rpm.pm
@@ -15,7 +15,7 @@ sub run {
     assert_screen('release-notes-button');
     send_key('ctrl-shift-alt-x');
     assert_screen('yast-xterm');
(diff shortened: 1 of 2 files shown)


jknphy:
LGTM, but could you please add to the first comment the link to the test you try to fix (when there is not ticked linked and it is a quick fix it helps to know what we fix).

ge0r (approved):
LGTM!

PR #14632, opened 2022-04-01
Postpone snapper cron job execution

mloviska:
`snapper-0.5` that is present in sle12sp5 comes with cron jobs
`/etc/cron.{daily,hourly}/suse.de-snapper` to do the regular snapshot cleanup.
Jobs might be triggered by a single schedule of run-crons `-*/15 * * * *   root  test -x /usr/lib/cron/run-crons &&
/usr/lib/cron/run-crons >/dev/null 2>&1` and according to timestamps of
`/var/spool/cron/lastrun/cron.{daily,hourly,monthly,weekly}` files.

- Related ticket: [[sporadic] test fails in snapper_cleanup - snapper cleanup number - Config is in use.](https://progress.opensuse.org/issues/103299)
- Verification run: [sle-12-SP5-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220401-1-jeos-filesystem@64bit-virtio-vga](http://kepler.suse.cz/tests/16053)

### DIFF
--- a/lib/btrfs_test.pm
+++ b/lib/btrfs_test.pm
@@ -80,6 +80,18 @@ sub snapper_nodbus_restore {
     }
 }
 
+=head2 cron_mock_lastrun
+snapper-0.5 and older is using cron jobs in order to schedule and execute cleanup routines.
+Script /usr/lib/cron/run-crons looks into /etc/cron.{hourly,daily,weekly,monthly} for jobs
+to be executed. The info about last run is stored in /var/spool/cron/lastrun
+By updating the lastrun files timestamps, we make sure those routines won't be executed 
+while tests are running. 
+=cut
+sub cron_mock_lastrun {
+    assert_script_run 'touch /var/spool/cron/lastrun/cron.{hourly,daily,weekly,monthly}';
+    assert_script_run 'ls -al /var/spool/cron/lastrun/cron.{hourly,daily,weekly,monthly}';
+}
+
 sub post_fail_hook {
     my ($self) = shift;
     select_console('log-console');
--- a/tests/console/snapper_cleanup.pm
+++ b/tests/console/snapper_cleanup.pm
@@ -23,6 +23,7 @@ use warnings;
 use testapi;
 use utils 'clear_console';
 use List::Util qw(max min);
+use version_utils qw(is_sle);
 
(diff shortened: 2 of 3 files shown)


tests/console/snapper_cleanup.pm
```diff
@@ -71,6 +72,11 @@ sub run {
     my $self = shift;
     $self->select_serial_terminal;
 
+    if (is_sle('<15')) {
+        assert_script_run 'touch /var/spool/cron/lastrun/cron.{hourly,daily,weekly,monthly}';
```
jlausuch, line 76: So, those files are not created by default and we need them to be there?

jlausuch:
https://openqa.suse.de/tests/8461380#step/snapper_cleanup/159

tests/console/snapper_cleanup.pm
mloviska: Or to just update the timestamps of the files. In https://build.suse.de/package/view_file/SUSE:SLE-12-SP5:Update/cronie/run-crons?expand=1 L#128 and then L#223

mloviska:
Maybe we need the same for [snapper_create](http://kepler.suse.cz/tests/16075#step/snapper_create/157)

lib/btrfs_test.pm
```diff
@@ -80,6 +80,16 @@ sub snapper_nodbus_restore {
     }
 }
 
+=head2 cron_mock_lastrun
+snapper-0.5 and older is using cron jobs in order to schedule and execute cleanup routines.
+Script /usr/lib/cron/run-crons looks into /etc/cron.{hourly,daily,weekly,monthly} for jobs
+to be executed. The info about last run is stored in /var/spool/cron/lastrun
```
jlausuch, line 86: Even if it's obvious, you can add also:
`By updating the lastrun files timestamps, we make sure those routines won't be executed while tests are running`
or something like this, if I get it right.

mloviska: updated, resolving

mloviska:
VR after applied comments: http://kepler.suse.cz/tests/16333#live

jlausuch (approved):
LGTM. Waiting for VR.

PR #14635, opened 2022-04-02
Split host and guest install, Adding host autoyast install option

tonyyuan1:
This PR will do:

1.  Splitting the host and guest  installation with yaml schedule files.
2.  Adding autoyast host installation option, see two templates host_12.xml.ep, host_15.xml.ep for detail.   GUI installation option and autoyast can be switched with variable: HOST_INSTALL_AUTOYAST
3.  autoyast configures host to comply with SUSE  "virtualization best practices" during installation
4. guest  installation test module supports  "rerun".

Autoyast  is SUSE linux native automated installation.  It does not only automate OS installation but also automate OS system configuration.  Unlike GUI installation modules in OpenQA , Autoyast install just need 3 test modules.  The two test modules (prepare_profile.pm, installation.pm)are developed and maintained by autoyast QA.  They are stable and rarely changed in upstream. The whole autoyast install uses much  fewer test modules and needles than GUI installation automation.

- Related ticket: https://progress.opensuse.org/issues/81094
- Needles: no
- Verification run: 
- sles15sp2-xen autoyast: 
http://openqa.qam.suse.cz/tests/overview?groupid=157&build=%3Asplit%3Axen%3Ariver&version=15-SP2&distri=sle
http://openqa.qam.suse.cz/tests/39404#dependencies
http://openqa.qam.suse.cz/tests/39404#

- sles15sp2-xen gui install: 
http://openqa.qam.suse.cz/tests/overview?build=%3A23269%3A&distri=sle&version=15-SP2&groupid=157

sles12sp5-xen autoyast:
http://openqa.qam.suse.cz/tests/overview?distri=sle&version=12-SP5&build=%3Asplit%3Axen%3Ainara&groupid=160

sles12sp4-kvm autoyast:
http://openqa.qam.suse.cz/tests/overview?distri=sle&version=12-SP4&build=%3Asplit%3Akvm%3Amisanthrope&groupid=142

sles15sp3 gui install:
http://openqa.qam.suse.cz/tests/overview?distri=sle&version=15-SP3&build=%3Atbaev%3Akaylee&groupid=156

Autoyast and gui installation have also been run with other sles15sp3, sp2, sp1 and sles15, sles12sp5, sp4. It might not be necessary post them all here.

### DIFF
--- a/data/virtualization/autoyast/host_12.xml.ep
+++ b/data/virtualization/autoyast/host_12.xml.ep
@@ -1,0 +1,324 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <bootloader>
+    <global>
+      <activate>true</activate>
+      <boot_mbr>true</boot_mbr>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --unit=1 --speed=115200 --parity=no</serial>
+      <terminal>serial</terminal>
+      <timeout config:type="integer">10</timeout>
+      <trusted_grub>false</trusted_grub>
+      <xen_append>splash=silent quiet console=<%= $get_var->('SERIALDEV') %>,115200</xen_append>
+      <xen_kernel_append>dom0_max_vcpus=4 dom0_mem=8192M,max:8192M loglvl=all guest_loglvl=all</xen_kernel_append>
+    </global>
+    <loader_type>grub2</loader_type>
+  </bootloader>
+  <general>
+    <ask-list config:type="list"/>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+    </mode>
+    <proposals config:type="list"/>
+    <signature-handling>
+      <accept_file_without_checksum config:type="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
+      <accept_unknown_gpg_key config:type="boolean">true</accept_unknown_gpg_key>
+      <accept_unsigned_file config:type="boolean">true</accept_unsigned_file>
+      <accept_verification_failed config:type="boolean">true</accept_verification_failed>
+      <import_gpg_key config:type="boolean">true</import_gpg_key>
+    </signature-handling>
+    <storage>
+      <partition_alignment config:type="symbol">align_optimal</partition_alignment>
+      <start_multipath config:type="boolean">false</start_multipath>
+    </storage>
+  </general>
+  <kdump>
+    <add_crash_kernel config:type="boolean">true</add_crash_kernel>
+    <crash_kernel config:type="list">
+      <listentry>72M,low</listentry>
+      <listentry>256M,high</listentry>
+    </crash_kernel>
+    <crash_xen_kernel>201M\&lt;4G</crash_xen_kernel>
+  </kdump>
+  <suse_register>
+    <do_registration config:type="boolean">true</do_registration>
+    <email><%= $get_var->('SCC_EMAIL') %></email>
+    <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+    <install_updates config:type="boolean">true</install_updates>
+    % if (keys %$addons) {
+    <addons config:type="list">
+      % while (my ($key, $addon) = each (%$addons)) {
+      <addon>
+        <name><%= $addon->{name} %></name>
+        <version><%= $addon->{version} %></version>
+        <arch><%= $addon->{arch} %></arch>
+        <release_type/>
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sles')) {
+        <reg_code><%= $get_var->('SCC_REGCODE_WE') %></reg_code>
+        % }
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        % }
+        % if ($key eq 'rt') {
+        <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+        % }
+        % if ($key eq 'ltss') {
+        <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
+        % }
+      </addon>
+      % }
+    </addons>
+    %}
+  </suse_register>
+  <add-on>
+    <add_on_products config:type="list">
+      % for my $repo (@$repos) {
+      <listentry>
+        <media_url><%= $repo %></media_url>
+      </listentry>
+      % }
+    </add_on_products>
+  </add-on>
+  <firewall>
+    <enable_firewall config:type="boolean">false</enable_firewall>
+    <start_firewall config:type="boolean">false</start_firewall>
+  </firewall>
+  <keyboard>
+    <keymap>english-us</keymap>
+  </keyboard>
+  <language>
+    <language>en_US</language>
+    <languages/>
+  </language>
+  <login_settings/>
+  <networking>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+    <managed config:type="boolean">false</managed>
+    <interfaces config:type="list">
+      <interface>
+        <device>br0</device>
+        <bootproto>dhcp</bootproto>
+        <bridge>yes</bridge>
+        <bridge_forwarddelay>0</bridge_forwarddelay>
+        <bridge_ports>eth0 eth1</bridge_ports>
+        <bridge_stp>off</bridge_stp>
+        <startmode>auto</startmode>
+      </interface>
+      <interface>
+        <device>eth0</device>
+        <bootproto>none</bootproto>
+        <startmode>hotplug</startmode>
+      </interface>
+      <interface>
+        <device>eth1</device>
+        <bootproto>none</bootproto>
+        <startmode>hotplug</startmode>
+      </interface>
+    </interfaces>
+  </networking>
+  <partitioning config:type="list">
+    <drive>
+      <disklabel>msdos</disklabel>
+      <enable_snapshots config:type="boolean">false</enable_snapshots>
+      <initialize config:type="boolean">true</initialize>
+      <partitions config:type="list">
+        <partition>
+          <create config:type="boolean">true</create>
+          <crypt_fs config:type="boolean">false</crypt_fs>
+          <filesystem config:type="symbol">btrfs</filesystem>
+          <format config:type="boolean">true</format>
+          <fstopt>defaults</fstopt>
+          <loop_fs config:type="boolean">false</loop_fs>
+          <mount>/</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">131</partition_id>
+          <partition_nr config:type="integer">1</partition_nr>
+          <partition_type>primary</partition_type>
+          <resize config:type="boolean">false</resize>
+          <size>400G</size>
+        </partition>
+        <partition>
+          <create config:type="boolean">true</create>
+          <crypt_fs config:type="boolean">false</crypt_fs>
+          <filesystem config:type="symbol">swap</filesystem>
+          <format config:type="boolean">true</format>
+          <fstopt>defaults</fstopt>
+          <loop_fs config:type="boolean">false</loop_fs>
+          <mount>swap</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">130</partition_id>
+          <partition_nr config:type="integer">2</partition_nr>
+          <partition_type>primary</partition_type>
+          <resize config:type="boolean">false</resize>
+          <size>16G</size>
+        </partition>
+      </partitions>
+      <type config:type="symbol">CT_DISK</type>
+      <use>all</use>
+    </drive>
+  </partitioning>
+  <report>
+    <errors>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">0</timeout>
+    </errors>
+    <messages>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </messages>
+    <warnings>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </warnings>
+    <yesno_messages>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </yesno_messages>
+  </report>
+  <services-manager>
+    <default_target>multi-user</default_target>
+    <services>
+      <disable config:type="list"/>
+      <enable config:type="list">
+        <service>sshd</service>
+      </enable>
+    </services>
+  </services-manager>
+  <software>
+    <packages config:type="list">
+      <package>dhcp-client</package>
+    </packages>
+    <products config:type="list">
+      <product><%= uc $get_var->('SLE_PRODUCT') %></product>
+    </products>
+    <patterns config:type="list">
+      % for my $pattern (@$patterns) {
+      <pattern><%= $pattern %></pattern>
+      % }
+    </patterns>
+  </software>
+  <ssh_import>
+    <copy_config config:type="boolean">false</copy_config>
+    <import config:type="boolean">false</import>
+  </ssh_import>
+  <timezone>
+    <hwclock>UTC</hwclock>
+    <timezone>UTC</timezone>
+  </timezone>
+  <ntp-client>
+    <ntp_policy>auto</ntp_policy>
+    <start_at_boot config:type="boolean">true</start_at_boot>
+   <restricts config:type="list">
+     <restrict>
+       <options>kod nomodify notrap nopeer noquery</options>
+       <target>default</target>
+     </restrict>
+     <restrict>
+       <target>127.0.0.1</target>
+     </restrict>
+     <restrict>
+       <target>::1</target>
+     </restrict>
+   </restricts>
+   <peers config:type="list">
+     <peer>
+       <address>0.europe.pool.ntp.org</address>
+       <options>iburst</options>
+       <type>server</type>
+     </peer>
+     <peer>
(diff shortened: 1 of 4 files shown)


tbaev (approved):
LGTM, great PR Tony. :)

data/virtualization/autoyast/host_12.xml.ep
```diff
@@ -0,0 +1,323 @@
...
+    <global>
+      <activate>true</activate>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --unit=1 --speed=115200 --parity=no</serial>
+      <terminal>serial</terminal>
+      <timeout config:type="integer">10</timeout>
+      <trusted_grub>false</trusted_grub>
+      <xen_append>splash=silent quiet console=ttyS1,115200</xen_append>
```
alice-suse, line 14: Some worker needs ttyS0 while some needs ttyS1, right?

data/virtualization/autoyast/host_12.xml.ep
```diff
@@ -0,0 +1,323 @@
...
+        % }
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        % }
+        % if ($key eq 'rt') {
+        <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+        % }
+        % if ($key eq 'ltss') {
+        <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
+        % }
```
alice-suse, line 70: From an incident job http://openqa.qam.suse.cz/tests/36483#settings, I get below.

SCC_ADDONS = base,sdk,serverapp,desktop

Which if branch does it go to?

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --unit=1 --speed=115200 --parity=no</serial>
+      <timeout config:type="integer">10</timeout>
+      <trusted_grub>false</trusted_grub>
+      <xen_append>splash=silent quiet console=ttyS1,115200</xen_append>
+      % if ($check_var->('VERSION', '15')) {
+      <terminal>serial</terminal>
+      <xen_kernel_append>dom0_max_vcpus=4 dom0_mem=8192M,max:8192M vga=gfx-1024x768x16 loglvl=all guest_loglvl=all</xen_kernel_append>
+      % } else {
```
alice-suse, line 55: What version is the "else branch" supposed to fit for? version>15?
I assume that host_12.xml.ep is for sle12spx products and host_15.xml.ep is for sle15spx products, right?

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+        <address>europe.pool.ntp.org</address>
+        <iburst config:type="boolean">true</iburst>
+        <offline config:type="boolean">false</offline>
+      </ntp_server>
+    </ntp_servers>
+    <ntp_sync>systemd</ntp_sync>
+  </ntp-client>
+  <partitioning config:type="list">
+    <drive>
+      <!--device>/dev/sda</device-->
```
alice-suse, line 130: Why is it commented off? Without it, which device is the partition created on?

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+          <create config:type="boolean">true</create>
+          <create_subvolumes config:type="boolean">true</create_subvolumes>
+          <filesystem config:type="symbol">btrfs</filesystem>
+          <format config:type="boolean">true</format>
+          <mount>/</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">131</partition_id>
+          <partition_nr config:type="integer">1</partition_nr>
+          <resize config:type="boolean">false</resize>
+          <size>400G</size>
```
alice-suse, line 145: Do all machines have one single disk with size >400G? Besides, there is no mount point given. Why not assign it to /?

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+        <partition>
+          <create config:type="boolean">true</create>
+          <filesystem config:type="symbol">swap</filesystem>
+          <format config:type="boolean">true</format>
+          <mount>swap</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">130</partition_id>
+          <partition_nr config:type="integer">2</partition_nr>
+          <resize config:type="boolean">false</resize>
+          <size>20G</size>
```
alice-suse, line 156: Will it be better to make swap size fitting 2^n? eg 16G or 32G?

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+      <on_demand config:type="list"/>
+    </services>
+  </services-manager>
+  <software>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <instsource/>
+    <packages config:type="list">
+      <package>dhcp-client</package>
+    % if ($check_var->('SYSTEM_ROLE', 'kvm')) {
+      <package>qemu-kvm</package>
```
alice-suse, line 203: Why does it need to be explicitly given? It shall be included in kvm_server pattern. KVM server is better to be installed with kvm_server and kvm_tools pattern, while xen server needs xen_server and xen_tools pattern, then all packages in the patterns will be installed accordingly.

data/virtualization/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,305 @@
...
+        <filename>disable_dom0_balloon.sh</filename>
+        <source><![CDATA[
+if [ -e /etc/xen/xl.conf ]; then sed -i '/autoballoon=/ s/.*/autoballoon="off"/' /etc/xen/xl.conf; fi
+]]>
+        </source>
+      </script>
+   % } 
+    </post-scripts>
+  </scripts>
+</profile>
```
alice-suse, line 305: The two files host_12 | host_15.xml.ep have a lot in common. Why not making them in one? What's the considerations?

tests/virtualization/universal/prepare_guests.pm
```diff
@@ -20,6 +20,12 @@ sub run {
     my $self = shift;
     # Use serial terminal, unless defined otherwise. The unless will go away once we are certain this is stable
     $self->select_serial_terminal unless get_var('_VIRT_SERIAL_TERMINAL', 1) == 0;
+    # Clean up VMs if any. This makes it possable to reinstall VMs.
+    systemctl("restart libvirtd");
+    assert_script_run('for i in $(virsh list --name|grep sles);do virsh shutdown $i;done');
```
alice-suse, line 25: "Virsh destroy" can force the guest off like physically power off and is faster than 'virsh shutdown'. IMHO, 'virsh shutdown' can be removed.

tests/virtualization/universal/prepare_guests.pm
```diff
@@ -20,6 +20,12 @@ sub run {
...
+    sleep 30;
+    assert_script_run('for i in $(virsh list --name|grep sles);do virsh destroy $i;done');
+    assert_script_run('for i in $(virsh list --name --inactive); do virsh undefine $i --remove-all-storage;done');
```
alice-suse, line 28: How about the windows guest?

alice-suse: @tonyyuan1  @tbaev  I guess this code block is the part that you say can reject the guest installation retry ticket? If yes, IMHO, it is not the same. 
This code block ensures that each time you rerun the prepare_guests.pm, all guests are cleaned and reinstalled freshly. This is beneficial and good when only 1~2 guests are tested at the same time, but it is not efficient enough when multiple guests are included in one test. Eg, you test 6 guests, 5 passed and 1 failed in installation. If using this code block, the 5 guests have to be recreated again.
However that guest retry ticket is that DURING the prepare_guests.pm step, if one guest fails to be installed, it retries multiple times by cleaning up this guest only -> reinstall guest only until it is successful or fail after max retries. It will be useful for above 6 guest example. So, IMHO, the guest retry ticket should be reopened again. HDYT?

alice-suse:
@tonyyuan1 Thanks for the big PR. It involves an unplanned task of doing host installation by autoyast. Glad to see that there is a switch setting to enable/disable it, which give us room to try and stablize the new installation. This is a fundamental  step and involves a lot of details which are critical to compatiblity on all physical test machines, product list, hypervisor types. I tried my best to review it deeply in details that I know, but can't promise all are covered. So have to rely on your experimental tests and future live incident validation jobs to consolidate the code.

data/virtualization/autoyast/host_12.xml.ep
tonyyuan1: The workers just use the value of SERIALDEV.  
Xen Kernel will dump more information to serial port during boot if console=ttySx is appended here. We can add a console=ttyS0,115200 or remove "console=ttyS1" here.

data/virtualization/autoyast/host_12.xml.ep
tonyyuan1: line51-57.  Check prepare_ay_file in prepare_profile.pm for detail.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: There is  a lot of  syntax change between autoyast for sle12 and sle15 so we have to create two separate templates.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: autoyast can detect and find out the right device. AMD machines does not use /dev/sda. it's specified it will break AMD installation.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: Different machines have different disk size. The minimum size is 450G so assign 400G to root partition and 20G  to swap. See line 140 for / mount point

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: That's work for small physical memory. 2^n is too big space for large memory. We give 20G to swap in case system is out of memory which I believe, never going to happen as our machines memory are all big.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: There is much syntax change between sle12 autoyast  and sle115.  We have to have two templates.

tests/virtualization/universal/prepare_guests.pm
tonyyuan1: Yes, you are right. "Virsh destroy" is sufficient.

tests/virtualization/universal/prepare_guests.pm
tonyyuan1: Which ticket is rejected?

data/virtualization/autoyast/host_12.xml.ep
alice-suse: It seems to bring some misunderstanding. Let me explain again. If I understand correctly, the xen_append in autoyast will affect the grub.cfg that is generated after host installation. However, the console setting about ttySx should be directly related to which physical com port the serial console is binded to. For all MU machines, most machines uses com0 => ttyS0, while the new amd machines use com1 => ttyS1. So to let all machines console setting to be correct, the xen_append here should be dynamically set rather than hard code to ttyS0 or ttyS1, right?  You can also refer to http://openqa.suse.de/tests/8162489/modules/logs_from_installation_system/steps/1/src line 59 for complete xen serial console setup.

data/virtualization/autoyast/host_15.xml.ep
alice-suse: It makes sense to have two templates for sle12 and 15. Would you please answer the above question "What version is the "else branch" supposed to fit for? version>15?"?

data/virtualization/autoyast/host_12.xml.ep
alice-suse: Sorry tony, this won't help. I read the code. But I am afraid these addons would not match any of the branch, and end up with no reg_code. I assume the logic of this code block should be adding <addon> iteratively for the listed SCC_ADDONS? Did you check the generated autoyast file?

data/virtualization/autoyast/host_15.xml.ep
alice-suse: ok

data/virtualization/autoyast/host_15.xml.ep
alice-suse: Well, some have 16G which is not that big IMHO. Anyway, we set swap just as any user would do. And for best practice, usually the size is better to be the exponential number of 2, like 2G, 4G,8G, etc. If you choose 20G, better to slightly adjust it to be 16G.

data/virtualization/autoyast/host_15.xml.ep
alice-suse: ok, make sense.

tests/virtualization/universal/prepare_guests.pm
alice-suse: https://progress.opensuse.org/issues/102984

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: qemu-kvm command is obsolete.  qemu-kvm package is not in kvm_server pattern since sles15.

tests/virtualization/universal/prepare_guests.pm
tonyyuan1: > How about the windows guest?

windows guest is imported. It just removes the downloaded image.

tonyyuan1: By the way, I have never seen any windows guest running in the test.

data/virtualization/autoyast/host_12.xml.ep
tonyyuan1: I do hate code split in pieces and spreading into different files. The readers have to locate them and put them together so as to read them as whole. 

It's implemented by expand_addons() in lib/autoyast.pm. expand_addons constructs  hash.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: yes, for all versions >15

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: yes, 16G is ok.

tests/virtualization/universal/prepare_guests.pm
tonyyuan1: > https://progress.opensuse.org/issues/102984
Below is what I understand, 
There is ~30 mins waiting time limit. The all guest install must be complete within 30 mins. The failure must be found as early as possible.  The second/third retry must be started as early as possible so that second/third installation could finish within remaining time before timeout. 

Currently, we reinstall all if any guest install fails. Teodor is exploring a better solution, https://progress.opensuse.org/issues/109328

Alice, as  your ticket exposes the issue we need to resolve. I would think that we could keep it open to track this issue. Maybe associate the two tickets. It's up to your two' decision :)

alice-suse: Okay, given that neither this PR(cleanup all guest when test rerun) nor https://progress.opensuse.org/issues/109328 (fail earlier when guest install fails) actually do guest installation retry, I'd suggest to reopen https://progress.opensuse.org/issues/102984. @tbaev HDYT? Reasonable?

data/virtualization/autoyast/host_15.xml.ep
alice-suse: Oh, ok. If it is obsolete, what tests of MU use qemu-kvm? Is it NECESSARY to keep using obsolete command? Will it be better to obsolete it and use the new one(I think it is qemu-system-x86_64)?

data/virtualization/autoyast/host_15.xml.ep
alice-suse: Interesting... how is it known exactly the setting for something like sle16, which is not available yet?

data/virtualization/autoyast/host_12.xml.ep
alice-suse: I read it ;-). Your validation job proves my guess in http://openqa.qam.suse.cz/tests/39404/file/autoyast-host_15.xml. ALthough the addons won't all match into if else and have reg_code, but it will be fine because the top reg_code also works for  addons like "base,sdk,serverapp,desktop". 
Mark it as resolved.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: qemu-system-x86_64 is default. qemu-kvm should not be used. 
When I was testing live migration, a test was broken if qemu-kvm was missing. I had to install qemu-kvm as workround. I can't remember the detail of it now. 

I will remove qemu-kvm to see what's going to happen with this test.

data/virtualization/autoyast/host_15.xml.ep
tonyyuan1: The logic here means:
Set grub2 to use "serial" and "console" in terminal tag for all of sle versions but 15GA.  15GA does support using both of them. 
This autoyast template is just for sle15. sle16 isn't covered.

data/virtualization/autoyast/host_12.xml.ep
tonyyuan1: Check if the new change is "dynamic" :)

data/virtualization/autoyast/host_15.xml.ep
alice-suse: Aha, okay. $check_var->('VERSION', '15') stands for 15GA.

tests/virtualization/universal/prepare_guests.pm
alice-suse: > By the way, I have never seen any windows guest running in the test.

windows test is run after qam-kvm|xen-install, eg http://openqa.qam.suse.cz/tests/36488#dependencies

tbaev: I agree, I will link the two tickets together (109328 and 102984)

alice-suse (approved):
LGTM

PR #14636, opened 2022-04-02
Public Cloud: exclude validate_btrfs tests

jlausuch:
This test can't be executed in PC jobs as we the filesystem is XFS
and we lack second disk.
Fixing a small regression by #14542

Example of failed test: https://openqa.suse.de/tests/8457925


jlausuch:
@pdostal ^

pdostal (approved):

PR #14647, opened 2022-04-04
Containers: add conditions for Leap Micro 

jlausuch:
[MicroOS](https://openqa.opensuse.org/tests/2283217)
[Leap Micro](https://openqa.opensuse.org/tests/2283218)  (error should be fixed by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14648)
[SLE Micro](https://openqa.suse.de/tests/8470872)
[SLE 12-SP5](http://openqa.suse.de/t8470873)
[SLE 15-SP1](http://openqa.suse.de/t8470874)

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -77,8 +77,8 @@ sub load_image_tests_docker {
 
 sub load_host_tests_podman {
     my ($run_args) = @_;
-    if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+") || is_sle_micro || is_microos) {
-        # podman package is only available as of 15-SP1
+    # podman package is only available as of 15-SP1
+    unless (is_sle("<15-sp1")) {
         load_container_engine_test($run_args);
         # In Public Cloud we don't have internal resources
         load_image_test($run_args) unless is_public_cloud;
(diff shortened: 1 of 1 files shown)


pdostal (approved):

jlausuch:
> I'm against this change: Functional disadvantages by implicitly assuming that we will never run container test on transactional system for aesthetic reasons only, is not something that I do not find a good practice. Our future selves might be unhappy ....

Ok, you convinced me :)

lib/main_containers.pm
```diff
@@ -77,7 +77,7 @@ sub load_image_tests_docker {
 
 sub load_host_tests_podman {
     my ($run_args) = @_;
-    if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+") || is_sle_micro || is_microos) {
+    if (is_leap('15.1+') || is_tumbleweed || is_sle("15-sp1+") || is_sle_micro || is_microos || is_leap_micro) {
```
mloviska, line 80: Hm...Doesn't it already mean run everywhere except of sle <= 15 ?

jlausuch: done, and new VRs

grisu48 (approved):
LGTM!

PR #14655, opened 2022-04-05
Public Cloud: Introduce registercloudguest test

pdostal:
There are two parts:
 * `publiccloud::instance::check_guestregister()` executed right after the instance boots up.
 * `publiccloud/registercloudguest` test module deregistering, reregistering and force registering the instance.
 * `publiccloud/register_system` ➡️ `publiccloud/register_addons` register only the addons **(new change)**.

This test follows [SUSE-Enceladus/cloud-regionsrv-client/integration_test-process.txt](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt)

- Related ticket: [poo#109304](https://progress.opensuse.org/issues/109304)
- Verification run: [12-SP4 GCE BYOS containers](http://pdostal-server.suse.cz/tests/14361), [12-SP5 EC2 img_proof](http://pdostal-server.suse.cz/tests/14359), [15-SP1 Azure BYOS gen2 consoletests](http://pdostal-server.suse.cz/tests/14360), [15-SP3 EC2 consoletests](http://pdostal-server.suse.cz/tests/14362), [15-SP3 EC2 BYOS ARM containers](http://pdostal-server.suse.cz/tests/14354), [15-SP4 Azure BYOS gen2 fio](http://pdostal-server.suse.cz/tests/14348)

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -203,8 +203,58 @@ sub wait_for_guestregister
         }
         sleep 1;
     }
+
     $self->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
     die('guestregister didn\'t end in expected timeout=' . $args{timeout});
+}
+
+=head2 check_guestregister
+
+    check_guestregister();
+
+Check that the registration state of newly created guest match our requirements.
+https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt
+=cut
+sub check_guestregister {
+    my ($self) = @_;
+
+    if (is_byos) {
+        if ($self->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
+            die 'The BYOS instance should be unregistered and report "Warning: No repositories defined.".';
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
+            die('guestregister.service is not disabled');
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
+            die('/etc/zypp/credentials.d/ is not empty');
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) != 0) {
+            die('/var/log/cloudregister is not empty');
+        }
+    } else {
+        if ($self->run_ssh_command(cmd => 'sudo zypper lr | wc -l', proceed_on_failure => 1, timeout => 360) < 5) {
+            die 'The list of zypper repositories is too short.';
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
+            die('guestregister.service is not enabled');
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) == 0) {
+            die('/etc/zypp/credentials.d/ is empty');
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) == 0) {
+            die('/var/log/cloudregister is empty');
+        }
+    }
+
+    if (is_azure() && $self->run_ssh_command(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer', proceed_on_failure => 1) !~ /enabled/) {
+        die('regionsrv-enabler-azure.timer is not enabled');
+    }
 }
 
 =head2 wait_for_ssh
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -19,7 +19,7 @@ use version_utils qw(is_sle is_public_cloud);
 use publiccloud::ssh_interactive;
 use registration;
 
-our @EXPORT = qw(select_host_console is_byos is_ondemand is_ec2 is_azure is_gce registercloudguest register_addon register_openstack);
+our @EXPORT = qw(select_host_console is_byos is_ondemand is_ec2 is_azure is_gce registercloudguest register_addon deregister_addon register_openstack);
 
 # Select console on the test host, if force is set, the interactive session will
 # be destroyed. If called in TUNNELED environment, this function die.
@@ -77,6 +77,27 @@ sub register_addon {
         ssh_add_suseconnect_product($remote, get_addon_fullname($addon), undef, $arch, '', $timeout, $retries, $delay);
     }
     record_info('SUSEConnect time', 'The command SUSEConnect -r ' . get_addon_fullname($addon) . ' took ' . (time() - $cmd_time) . ' seconds.');
+}
+
+sub deregister_addon {
+    my ($remote, $addon) = @_;
+    my $arch = get_var('PUBLIC_CLOUD_ARCH', 'x86_64');
+    $arch = "aarch64" if ($arch eq "arm64");
+    my $timestamp = utc_timestamp();
+    record_info($addon, "Going to deregister '$addon' addon\nUTC: $timestamp");
+    my $cmd_time = time();
+    my ($timeout, $retries, $delay) = (300, 3, 120);
+    if ($addon =~ /ltss/) {
+        # ssh_remove_suseconnect_product($remote, $name, $version, $arch, $params, $timeout, $retries, $delay)
+        ssh_remove_suseconnect_product($remote, get_addon_fullname($addon), '${VERSION_ID}', $arch, "-r " . get_required_var('SCC_REGCODE_LTSS'), $timeout, $retries, $delay);
+    } elsif (is_sle('<15') && $addon =~ /tcm|wsm|contm|asmm|pcm/) {
+        ssh_remove_suseconnect_product($remote, get_addon_fullname($addon), '`echo ${VERSION} | cut -d- -f1`', $arch, '', $timeout, $retries, $delay);
+    } elsif (is_sle('<15') && $addon =~ /sdk|we/) {
+        ssh_remove_suseconnect_product($remote, get_addon_fullname($addon), '${VERSION_ID}', $arch, '', $timeout, $retries, $delay);
+    } else {
+        ssh_remove_suseconnect_product($remote, get_addon_fullname($addon), undef, $arch, '', $timeout, $retries, $delay);
+    }
+    record_info('SUSEConnect time', 'The command SUSEConnect -d ' . get_addon_fullname($addon) . ' took ' . (time() - $cmd_time) . ' seconds.');
 }
 
 sub registercloudguest {
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -18,6 +18,7 @@ our @EXPORT = qw(
   add_suseconnect_product
   ssh_add_suseconnect_product
   remove_suseconnect_product
+  ssh_remove_suseconnect_product
   cleanup_registration
   register_product
   assert_registration_screen_present
@@ -231,6 +232,22 @@ sub remove_suseconnect_product {
     $arch //= get_required_var('ARCH');
     $params //= '';
     script_retry("SUSEConnect -d -p $name/$version/$arch $params", retry => 5, delay => 60, timeout => 180);
+}
+
+=head2 ssh_remove_suseconnect_product
+
+    ssh_remove_suseconnect_product($name, [$version, [$arch, [$params]]]);
+
+Wrapper for SUSEConnect -d $name over ssh.
+=cut
+sub ssh_remove_suseconnect_product {
+    my ($remote, $name, $version, $arch, $params) = @_;
+    assert_script_run "sftp $remote:/etc/os-release /tmp/os-release";
+    assert_script_run 'source /tmp/os-release';
+    $version //= scc_version();
+    $arch //= get_required_var('arch');
+    $params //= '';
+    script_retry("ssh $remote sudo SUSEConnect -d -p $name/$version/$arch $params", retry => 5, delay => 60, timeout => 180);
 }
 
 =head2 cleanup_registration
--- a/tests/publiccloud/img_proof.pm
+++ b/tests/publiccloud/img_proof.pm
@@ -112,13 +112,14 @@ sub run {
     } else {
         $provider = $self->provider_factory();
         $instance = $provider->create_instance();
+
+        $instance->wait_for_guestregister() if is_ondemand();
+        $instance->check_guestregister();
     }
     if ($tests eq "default") {
         $tests = $img_proof_tests->{$flavor};
         die("Missing img_proof tests for $flavor - plz change img_proof.pm") unless $tests;
     }
-
-    $instance->wait_for_guestregister() if is_ondemand();
 
     my $img_proof = $provider->img_proof(
         instance => $instance,
--- a/tests/publiccloud/registercloudguest.pm
+++ b/tests/publiccloud/registercloudguest.pm
@@ -27,8 +27,7 @@ sub run {
     # Create $instance to make the code easier to read
     my $instance = $args->{my_instance};
 
-    my $regcode = (is_byos()) ? get_required_var('SCC_REGCODE') : undef;
-    my $regcode_param = ($regcode) ? "-r $regcode" : '';
+    my $regcode_param = (is_byos()) ? "-r " . get_required_var('SCC_REGCODE') : '';
 
     select_host_console();    # select console on the host, not the PC instance
 
--- a/tests/publiccloud/registercloudguest.pm
+++ b/tests/publiccloud/registercloudguest.pm
@@ -1,0 +1,79 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: cloud-regionsrv-client
+# Summary: Test system (re)registration
+# https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt
+#
+# Maintainer: <qa-c@suse.de>
+
+use Mojo::Base 'publiccloud::basetest';
+use version_utils;
+use registration;
+use warnings;
+use testapi;
+use strict;
+use utils;
+use publiccloud::utils;
+
+sub run {
+    my ($self, $args) = @_;
+
+    # Preserve args for post_fail_hook
+    $self->{instance} = $args->{my_instance};
+
+    # Create $instance to make the code easier to read
+    my $instance = $args->{my_instance};
+
+    my $regcode = (is_byos()) ? get_required_var('SCC_REGCODE') : undef;
+    my $regcode_param = ($regcode) ? "-r $regcode" : '';
+
+    select_host_console();    # select console on the host, not the PC instance
+
+    # Test re-registration. Assuming the system has been registered before
+    # The `sudo SUSEConnect -d` is not supported on BYOS and should fail.
+    $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d') if (is_byos());
+    $instance->run_ssh_command(cmd => 'sudo registercloudguest --clean');
+    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600, proceed_on_failure => 1) > 2) {
+        die('The list of zypper repositories is not empty.');
+    }
+    if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l', proceed_on_failure => 1) != 0) {
+        die('Directory /etc/zypp/credentials.d/ is not empty.');
+    }
+
+    # The SUSEConnect registration should still work on BYOS
+    if (is_byos()) {
+        $instance->run_ssh_command(cmd => 'sudo SUSEConnect --version');
+        $instance->run_ssh_command(cmd => "sudo SUSEConnect $regcode_param");
+        $instance->run_ssh_command(cmd => 'sudo registercloudguest --clean');
+    }
+
+    $instance->run_ssh_command(cmd => "sudo registercloudguest $regcode_param");
+    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
+        die('The list of zypper repositories is empty.');
+    }
+    if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
+        die('Directory /etc/zypp/credentials.d/ is empty.');
+    }
+
+    $instance->run_ssh_command(cmd => "sudo registercloudguest $regcode_param --force-new");
+    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
+        die('The list of zypper repositories is empty.');
+    }
+    if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
+        die('Directory /etc/zypp/credentials.d/ is empty.');
+    }
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    $self->{instance}->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
+    if (is_azure()) {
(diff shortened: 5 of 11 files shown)


pdostal:
The pull request is *blocked* due to [bsc#1198080](https://bugzilla.suse.com/show_bug.cgi?id=1198080). Although please feel free to review.

lib/publiccloud/instance.pm
```diff
@@ -203,10 +203,57 @@ sub wait_for_guestregister
...
+
+Check that the registration state of newly created guest match our requirements.
+https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt
+=cut
+sub check_guestregister
+{
+    my ($self) = @_;
+
+    if (is_byos) {
+        $self->run_ssh_command(cmd => '! sudo zypper lr', timeout => 6000);
```
jlausuch, line 223: why is this needed? If we want to make sure repos are not defined, why don't we use the same approach as below?
```
if ($self->run_ssh_command(cmd => '! sudo zypper lr', timeout => 6000, proceed_on_failure => 1, quiet => 1) !~ /no repos defined/) {
```
ofc, not that output message :)

lib/publiccloud/utils.pm
```diff
@@ -79,15 +79,36 @@ sub register_addon {
     record_info('SUSEConnect time', 'The command SUSEConnect -r ' . get_addon_fullname($addon) . ' took ' . (time() - $cmd_time) . ' seconds.');
 }
 
+sub deregister_addon {
+    my ($remote, $addon) = @_;
+    my $arch = get_var('PUBLIC_CLOUD_ARCH') // "x86_64";
```
jlausuch, line 84:
```suggestion
    my $arch = get_var('PUBLIC_CLOUD_ARCH', 'x86_64');
```

lib/publiccloud/utils.pm
```diff
@@ -79,15 +79,36 @@ sub register_addon {
...
-    my $regcode = get_required_var('SCC_REGCODE');
+    my $regcode = (is_byos()) ? '-r ' . get_required_var('SCC_REGCODE') : '';
     my $remote = $instance->username . '@' . $instance->public_ip;
-    # not all images currently have registercloudguest pre-installed .
+    # not all BYOS images currently have registercloudguest pre-installed .
     # in such a case,we need to regsiter against SCC and install registercloudguest with all needed dependencies and then
     # unregister and re-register with registercloudguest
     if ($instance->run_ssh_command(cmd => "sudo which registercloudguest > /dev/null; echo \"registercloudguest\$?\" ", proceed_on_failure => 1) =~ m/registercloudguest1/) {
-        $instance->retry_ssh_command(cmd => "sudo SUSEConnect -r $regcode", timeout => 420, retry => 3, delay => 120);
+        $instance->retry_ssh_command(cmd => "sudo SUSEConnect $regcode", timeout => 420, retry => 3, delay => 120);
```
jlausuch, line 111: if is on-demand image, it will execute only `sudo SUSEConnect`. What's the point?

lib/registration.pm
```diff
@@ -223,14 +224,30 @@ sub ssh_add_suseconnect_product {
...
+wrapper for suseconnect -d $name.
 =cut
 sub remove_suseconnect_product {
     my ($name, $version, $arch, $params) = @_;
     $version //= scc_version();
-    $arch //= get_required_var('ARCH');
+    $arch //= get_required_var('arch');
     $params //= '';
-    script_retry("SUSEConnect -d -p $name/$version/$arch $params", retry => 5, delay => 60, timeout => 180);
+    script_retry("suseconnect -d -p $name/$version/$arch $params", retry => 5, delay => 60, timeout => 180);
```
jlausuch, line 234: I've never seen this command in lower case, why do you change it?

lib/publiccloud/instance.pm
pdostal: In [SUSE-Enceladus/cloud-regionsrv-client/integration_test-process.txt](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt) is required both.

lib/publiccloud/utils.pm
pdostal: Good catch :wink:

lib/publiccloud/utils.pm
```diff
@@ -79,15 +79,36 @@ sub register_addon {
...
+    } else {
+        ssh_remove_suseconnect_product($remote, get_addon_fullname($addon), undef, $arch, '', $timeout, $retries, $delay);
+    }
+    record_info('SUSEConnect time', 'The command SUSEConnect -d ' . get_addon_fullname($addon) . ' took ' . (time() - $cmd_time) . ' seconds.');
+}
+
 sub registercloudguest {
     my ($instance) = @_;
-    my $regcode = get_required_var('SCC_REGCODE');
+    my $regcode = (is_byos()) ? '-r ' . get_required_var('SCC_REGCODE') : '';
```
grisu48, line 105: I would expect a variable `$regcode` to only contain the REGCODE and not an additional argument for a command that follows later. I would move the `-r` to `SUSEConnect` and `registercloudguest` do here simply

```suggestion
    my $regcode = (is_byos()) ? get_required_var('SCC_REGCODE') : '';
```

lib/registration.pm
```diff
@@ -223,16 +224,32 @@ sub ssh_add_suseconnect_product {
...
     remove_suseconnect_product($name, [$version, [$arch, [$params]]]);
 
-Wrapper for SUSEConnect -d $name.
+wrapper for suseconnect -d $name.
 =cut
 sub remove_suseconnect_product {
     my ($name, $version, $arch, $params) = @_;
     $version //= scc_version();
-    $arch //= get_required_var('ARCH');
+    $arch //= get_required_var('arch');
```
grisu48, line 232: Why is this now lowercase?

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,61 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: cloud-regionsrv-client
+# Summary: Test system (re)registration
```
grisu48, line 7: I would add here the reference to https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt or do you think that doesn't make sense? e.g.

    # Summary: Test system registration
    #          See https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt for the agreed test procedure

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,61 @@
...
+
+sub run {
+    my ($self, $args) = @_;
+
+    # Preserve args for post_fail_hook
+    $self->{instance} = $args->{my_instance};
+
+    # Create $instance to make the code easier to read
+    my $instance = $args->{my_instance};
+    my $regcode = (is_byos()) ? '-r ' . get_required_var('SCC_REGCODE') : '';
```
grisu48, line 28: Same as above: A variable named `$regcode` should only contain the REGCODE and not additional variables unless strictly necessary.

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,61 @@
...
+
+    $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d', quiet => 0) if (is_byos());
+    $instance->run_ssh_command(cmd => 'sudo registercloudguest --clean', quiet => 0);
+    $instance->run_ssh_command(cmd => '! sudo zypper lr', timeout => 600);
+    $instance->run_ssh_command(cmd => '! sudo ls /etc/zypp/credentials.d/*', quiet => 0);
+    if (is_byos()) {
+        $instance->run_ssh_command(cmd => 'sudo SUSEConnect --version');
+        $instance->run_ssh_command(cmd => "sudo SUSEConnect $regcode");
+        my $remote = $args->{my_instance}->username . '@' . $args->{my_instance}->public_ip;
+        #register_addon($remote, 'pcm');
```
grisu48, line 40: Leftover?

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,61 @@
...
+    # Preserve args for post_fail_hook
+    $self->{instance} = $args->{my_instance};
+
+    # Create $instance to make the code easier to read
+    my $instance = $args->{my_instance};
+    my $regcode = (is_byos()) ? '-r ' . get_required_var('SCC_REGCODE') : '';
+
+    select_host_console();    # select console on the host, not the PC instance
+
+    $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d', quiet => 0) if (is_byos());
```
grisu48, line 32: Please add a comment here stating 1) why we are doing this and 2) what the pre-requirements are. e.g.

```suggestion
    # Test the re-registration. This test assumes the registration has been tested before
    $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d', quiet => 0) if (is_byos());
```

grisu48 (commented):
Good work Pavel! I have left some comments, but feel that we are on a good track. Keep it up!

lib/publiccloud/instance.pm
pdostal: I used `rc_only` but it did not really work for me. I'll try once more.

tests/publiccloud/registercloudguest.pm
pdostal: This is commented because of the [bsc#1198080](https://bugzilla.suse.com/show_bug.cgi?id=1198080)

asmorodskyi: soft-failure than ?

lib/publiccloud/instance.pm
```diff
@@ -203,10 +203,57 @@ sub wait_for_guestregister
...
+
+        if ($self->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1, quiet => 1) !~ /disabled/) {
```
asmorodskyi, line 225: looks like something which needs `validate_ssh_command_output` ? if you know what I mean ?

tests/publiccloud/boottime.pm
```diff
@@ -271,6 +271,7 @@ sub do_systemd_analyze {
 
     # Wait for guest register, before calling syastemd-analyze
     $instance->wait_for_guestregister();
+    $instance->check_guestregister();
```
asmorodskyi, line 274: I think this test should be exception . check one line below . following measurement does not make sense with line which you adding here ...

lib/publiccloud/utils.pm
```diff
@@ -79,15 +79,36 @@ sub register_addon {
...
     my $remote = $instance->username . '@' . $instance->public_ip;
-    # not all images currently have registercloudguest pre-installed .
+    # not all BYOS images currently have registercloudguest pre-installed .
     # in such a case,we need to regsiter against SCC and install registercloudguest with all needed dependencies and then
     # unregister and re-register with registercloudguest
-    if ($instance->run_ssh_command(cmd => "sudo which registercloudguest > /dev/null; echo \"registercloudguest\$?\" ", proceed_on_failure => 1) =~ m/registercloudguest1/) {
-        $instance->retry_ssh_command(cmd => "sudo SUSEConnect -r $regcode", timeout => 420, retry => 3, delay => 120);
+    if (is_byos() && $instance->run_ssh_command(cmd => "sudo which registercloudguest > /dev/null; echo \"registercloudguest\$?\" ", proceed_on_failure => 1) =~ m/registercloudguest1/) {
```
asmorodskyi, line 110: if you check current master all calls of `registercloudguest` function come with condition `if(is_byos())` or `if(!is_on_demand())` . while generally I like idea to "move" this `if` into function and don't have it "outside" . I would call this cleaner approach . But currently I see two problems : 
1.  Current PR is not "refactoring of registercloudguest function" . and the fact that you mix this change with addition of new test make it harder to review for both the test and the refactoring. It would be just easier to tackle both if they come split. And I am speaking not just two different commits but really to separate PRs 
2. While you introduce `is_byos`  internally in function you haven't drop duplicated checks which we have on every call of this function. I would say it is overkill to do this check two times .

asmorodskyi: Regarding 1st point there is one this which could fully drop it - necessity of having them in one PR . If I overlook it please point me to it ...

tests/publiccloud/registercloudguest.pm
pdostal: The test dies in `registration::ssh_remove_suseconnect_product`. It is too complicated to softfailure there. I will remove it for now and mention it in the progress ticket.

lib/publiccloud/instance.pm
pdostal: I don't. I would first of all split and refactor `run_ssh_command`.

tests/publiccloud/registercloudguest.pm
pdostal: Can you please re-check now?

lib/publiccloud/utils.pm
pdostal: The `registercloudguest` is now as it was before. Good point, thank you

lib/publiccloud/instance.pm
asmorodskyi: I mean something like this https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1245 but instead of `script_output` call you doing `run_ssh_command` . And what you have on your mind when you saying about refactor `run_ssh_command` ?

lib/publiccloud/instance.pm
```diff
@@ -65,7 +65,7 @@ sub run_ssh_command {
 
     my $ssh_cmd = sprintf('ssh %s "%s@%s" -- %s', $args{ssh_opts}, $args{username}, $self->public_ip, $cmd);
     $ssh_cmd = "timeout $timeout $ssh_cmd" if ($timeout > 0);
-    record_info('SSH CMD', $ssh_cmd);
+    record_info('SSH CMD', $ssh_cmd) unless ($rc_only);
```
asmorodskyi, line 68: why we don't want to show command for case when you just need return code ? We already have `quiet` option maybe better to use it here instead ?

lib/publiccloud/utils.pm
```diff
@@ -79,6 +79,27 @@ sub register_addon {
...
 }
 
+sub deregister_addon {
+    my ($remote, $addon) = @_;
+    my $arch = get_var('PUBLIC_CLOUD_ARCH', 'x86_64');
+    $arch = "aarch64" if ($arch eq "arm64");
+    my $timestamp = utc_timestamp();
+    record_info($addon, "Going to deregister '$addon' addon\nUTC: $timestamp");
+    my $cmd_time = time();
+    # ssh_remove_suseconnect_product($remote, $name, $version, $arch, $params, $timeout, $retries, $delay)
```
asmorodskyi, line 89: leftover ?

lib/registration.pm
```diff
@@ -223,7 +224,7 @@ sub ssh_add_suseconnect_product {
 
     remove_suseconnect_product($name, [$version, [$arch, [$params]]]);
 
-Wrapper for SUSEConnect -d $name.
+wrapper for suseconnect -d $name.
```
asmorodskyi, line 227: what the purpose of this change ?

lib/registration.pm
```diff
@@ -233,6 +234,22 @@ sub remove_suseconnect_product {
...
 
+=head2 ssh_remove_suseconnect_product
+
+    ssh_remove_suseconnect_product($name, [$version, [$arch, [$params]]]);
+
+wrapper for suseconnect -d $name over ssh.
+=cut
+sub ssh_remove_suseconnect_product {
+    my ($remote, $name, $version, $arch, $params) = @_;
+    assert_script_run "sftp $remote:/etc/os-release /tmp/os-release";
```
asmorodskyi, line 245: what about our lovely  "fingerprint have changed " error , I assume we will start getting it here in some cases ?

lib/registration.pm
asmorodskyi: this question remain unanswered , can you please clarify ?

tests/publiccloud/img_proof.pm
```diff
@@ -112,14 +112,15 @@ sub run {
     } else {
         $provider = $self->provider_factory();
         $instance = $provider->create_instance();
+
+        $instance->wait_for_guestregister();
```
asmorodskyi, line 116: by this change you fully remove `wait_for_guestregister()` for all QAM runs , don't think that this was idea , right ?
also check will be only introduced for product runs

lib/publiccloud/instance.pm
pdostal: It's probably for separate discussion. Do you think that I should do this in this pull request?

lib/publiccloud/utils.pm
pdostal: No. It's valid comment. The `registration:: ssh_remove_suseconnect_product()` has too many parameters.

lib/registration.pm
pdostal: There is no purpose, it's a mistake caused by my new keyboard. I'm sorry

lib/registration.pm
pdostal: I didn't notice it lately. Plus it's in my opinion separate problem to solve.

tests/publiccloud/img_proof.pm
pdostal: Please re-check

lib/publiccloud/instance.pm
pdostal: I discarded this change. It should not be needed.

lib/registration.pm
pdostal: Yes, in `vim`, if you select some text, you can press `y` to yank it or `u` to make it lowercase. I simply didn't notice I pressed `u` instead of `y`.

lib/publiccloud/instance.pm
asmorodskyi: +1 , this is just thoughts for the future . resolving the thread

lib/publiccloud/instance.pm
pdostal: @grisu48 @asmorodskyi this is not true. See `testapi::script_output` [documentation](http://open.qa/api/testapi/#_script_output):
> Executing script inside SUT with bash -eox (in case of serial console with bash -eo) and directs stdout (not' stderr!) > to the serial console and returns the output if' the script exits with 0. Otherwise the test is set to failed.

lib/publiccloud/utils.pm
asmorodskyi: ah I see now , can you please move it two lines down so it would be clear to what it belongs

lib/registration.pm
asmorodskyi: ok

tests/publiccloud/img_proof.pm
```diff
@@ -112,14 +112,15 @@ sub run {
     } else {
         $provider = $self->provider_factory();
         $instance = $provider->create_instance();
+
+        $instance->wait_for_guestregister() if is_ondemand();
```
asmorodskyi, line 116: now you moved it back into "not QAM" branch , any particular reason for this ?

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,62 @@
...
+    # Create $instance to make the code easier to read
+    my $instance = $args->{my_instance};
+    my $regcode = get_required_var('SCC_REGCODE') if (is_byos());
+    my $regcode_param = ($regcode) ? "-r $regcode" : '';
+
+    select_host_console();    # select console on the host, not the PC instance
+
+    # Test re-registration. Assuming the system has been registered before
+    $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d', quiet => 0) if (is_byos());
+    $instance->run_ssh_command(cmd => 'sudo registercloudguest --clean', quiet => 0);
```
asmorodskyi, line 36: calling `registercloudguest --clean` just after `SUSEConnect -d` was called , I am not sure that this is good idea . Why you would do de-register two time in a row ?

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,62 @@
...
+    $instance->run_ssh_command(cmd => '! sudo zypper lr', timeout => 600);
```
asmorodskyi, line 37:
```suggestion
    $instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600);
```

and then checking for amount of lines is safer approach . in case of current implementation (with '!' ) you will also get green bubble in case of any errors from zypper e.g. broken repo configs

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,62 @@
...
+    $instance->run_ssh_command(cmd => '! sudo ls /etc/zypp/credentials.d/*', quiet => 0);
```
asmorodskyi, line 38: same here , better count lines

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,62 @@
...
+    if (is_byos()) {
+        $instance->run_ssh_command(cmd => 'sudo SUSEConnect --version');
+        $instance->run_ssh_command(cmd => "sudo SUSEConnect $regcode_param");
+        my $remote = $args->{my_instance}->username . '@' . $args->{my_instance}->public_ip;
```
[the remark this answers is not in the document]
pdostal: We don't. It's another leftover.

tests/publiccloud/img_proof.pm
pdostal: Yes. In latest we create and register the instance directly in `img_proof` test module. But in QAM we have separate modules for that.

tests/publiccloud/registercloudguest.pm
pdostal: I test that `SUSEConnect -d` fails. See the [documentation](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt):
> - SUSEConnect -d
>   + produce error message

tests/publiccloud/registercloudguest.pm
pdostal: Done.

tests/publiccloud/registercloudguest.pm
pdostal: Done.

lib/publiccloud/utils.pm
pdostal: Done.

asmorodskyi (approved):
LGTM

asmorodskyi:
I will not merge because I am not sure if @grisu48  resolve all his concerns about this PR.

lib/publiccloud/instance.pm
grisu48: Thanks for the clarification, then your code is correct :+1:

lib/publiccloud/instance.pm
```diff
@@ -203,10 +203,60 @@ sub wait_for_guestregister
...
+
+        if ($self->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1, quiet => 1) != 0) {
+            die('/etc/zypp/credentials.d/ is not empty');
+        }
+
+        if ($self->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1, quiet => 1) != 0) {
+            die('/var/log/cloudregister is not empty');
+        }
+    } else {
+        if ($self->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 360) < 5) {
```
grisu48, line 238: Forgotten `quiet`? You can ignore this comment if this is only a matter of style without functional implications.

grisu48 (approved):
LGTM!

lib/publiccloud/instance.pm
pdostal: TBH I'd put `quiet => 0` everywhere. That would make the review easier but also the code even less readable. WDYT?

grisu48: Yes, this is the only command where it is missing. I was wondering if that was on purpose of simply forgotten.

grisu48 (approved):

tests/publiccloud/registercloudguest.pm
```diff
@@ -0,0 +1,79 @@
...
+    my ($self, $args) = @_;
+
+    # Preserve args for post_fail_hook
+    $self->{instance} = $args->{my_instance};
+
+    # Create $instance to make the code easier to read
+    my $instance = $args->{my_instance};
+
+    my $regcode = (is_byos()) ? get_required_var('SCC_REGCODE') : undef;
+    my $regcode_param = ($regcode) ? "-r $regcode" : '';
```
asmorodskyi, line 31:
```suggestion
    my $regcode_param = (is_byos()) ? "-r " . get_required_var('SCC_REGCODE')  : '';
```

if I haven't miss something variable `$regcode` used only here , so it can be dropped

asmorodskyi (approved):
LGTM ( you may ignore my single comment :) )

pdostal:
Re-verified Anton's requested change on [PAYG](http://pdostal-server.suse.cz/tests/14363) and [BYOS](http://pdostal-server.suse.cz/tests/14364).

PR #14656, opened 2022-04-06
Enlarge counter value to match expected snapshot

chcao:
For some regression test modules, it need to install some packages
by "zypper" before performing regression test, where snapshots are
created. Now need to press more times 'down' key to find the snapshot
of before migration. This PR is going to enlarge counter value to match
expected snapshot.

- Related ticket: https://progress.opensuse.org/issues/109283
- Needles: N/A
- Verification run:
http://openqa.nue.suse.com/tests/8459275#step/grub_test_snapshot/62
http://openqa.nue.suse.com/tests/8459278#step/grub_test_snapshot/59

### DIFF
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -273,11 +273,11 @@ sub boot_into_snapshot {
     }
     # in upgrade/migration scenario, we want to boot from snapshot 1 before migration.
     if ((get_var('UPGRADE') && !get_var('ONLINE_MIGRATION', 0)) || get_var('ZDUP')) {
-        send_key_until_needlematch('snap-before-update', 'down', 40, 5);
+        send_key_until_needlematch('snap-before-update', 'down', 60, 5);
         save_screenshot;
     }
     # in an online migration
-    send_key_until_needlematch('snap-before-migration', 'down', 40, 5) if (get_var('ONLINE_MIGRATION'));
+    send_key_until_needlematch('snap-before-migration', 'down', 60, 5) if (get_var('ONLINE_MIGRATION'));
     save_screenshot;
     send_key 'ret';
     # avoid timeout for booting to HDD


lemon-suse (approved):

lemon-suse:
For the PR description, what do you think change as following?

For regression test,  it need to install more packages via 'zypper' in patch_sle for full addons and modules registered, 
where more shapshots are created. Now need to send more times 'down' key to find the snapshot of before migration, so enlarge counter value to resolve this problem.

lemon-suse:
For PR title, suggest to change to 'Enlarge counter value to match expected snapshot'

chcao:
> For the PR description, what do you think change as following?
> 
> For regression test, it need to install more packages via 'zypper' in patch_sle for full addons and modules registered, where more shapshots are created. Now need to send more times 'down' key to find the snapshot of before migration, so enlarge counter value to resolve this problem.

No, the snapshots are not created in patch_sle, they are created after migration, you can see the snapshots' named as SLES15SP4. If the snapshots are created in patch_sle, their names are SLES15SPx（x<4）. I checked some modules, for example, gnome-music, it did 'zypper in gnome-music' at the beginning of gnome-music testing.

chcao:
> For PR title, suggest to change to 'Enlarge counter value to match expected snapshot'

Thanks, done.

lemon-suse:
> 



> > For the PR description, what do you think change as following?
> > For regression test, it need to install more packages via 'zypper' in patch_sle for full addons and modules registered, where more shapshots are created. Now need to send more times 'down' key to find the snapshot of before migration, so enlarge counter value to resolve this problem.
> 
> No, the snapshots are not created in patch_sle, they are created after migration, you can see the snapshots' named as SLES15SP4. If the snapshots are created in patch_sle, their names are SLES15SPx（x<4）. I checked some modules, for example, gnome-music, it did 'zypper in gnome-music' at the beginning of gnome-music testing.

As my understanding, more addons & modules registered more related packages will be installed, so this should be the root cause of so many snapshot created. Anyway, you can keep your description since I haven't check each snapshot.

For the second sentence, I suggest to use 'send' instead of 'press'.

PR #14658, opened 2022-04-06
LTP testing on OpenStack JeOS

mloviska:
- motivation: []eOS OpenStack] Enable LTP tests](https://progress.opensuse.org/issues/107455)

As part of the OpenStack JeOS epic, we need to execute LTP tests against
JeOS openstack image. As of now it would be enough to stress *containers, syscalls and cve*

- Verification runs: 
  * [openstack-commands@64bit](http://kepler.suse.cz/tests/16353)
  * [openstack-cve@64bit](http://kepler.suse.cz/tests/16349)
  * [openstack-syscalls@64bit](http://kepler.suse.cz/tests/16357#)
  * [openstack-containers@64bit](http://kepler.suse.cz/tests/16362#)

DISCLAIMER: I haven't used any LTP_EXCLUDES in VRs

### DIFF
--- a/data/publiccloud/log_instance.sh
+++ b/data/publiccloud/log_instance.sh
@@ -121,6 +121,22 @@ azure_stop_log()
     fi
 }
 
+openstack_start_log()
+{
+    read_unique_counter
+    set +e
+    openstack server start "$INSTANCE_ID" > "${OUTPUT_DIR}/$CNT""_boot_log_start.txt" 2>&1
+    set -e
+}
+
+openstack_stop_log()
+{
+    read_unique_counter
+    set +e
+    openstack server stop "$INSTANCE_ID" > "${OUTPUT_DIR}/$CNT""_boot_log_start.txt" 2>&1
+    set -e
+}
+
 read_unique_counter()
 {
     CNT=$(printf "%03d" "$(cat "$CNT_FILE" 2> /dev/null)")
@@ -163,6 +179,9 @@ case $PROVIDER in
         ;;
     GCE)
         gce_"${COMMAND}"_log
+        ;;
+    OPENSTACK)
+        openstack_"${COMMAND}"_log
         ;;
     *)
         echo "Unknown provider $PROVIDER given";
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -19,7 +19,7 @@ use version_utils qw(is_sle is_public_cloud);
 use publiccloud::ssh_interactive;
 use registration;
 
-our @EXPORT = qw(select_host_console is_byos is_ondemand is_ec2 is_azure is_gce registercloudguest register_addon);
+our @EXPORT = qw(select_host_console is_byos is_ondemand is_ec2 is_azure is_gce registercloudguest register_addon register_openstack);
 
 # Select console on the test host, if force is set, the interactive session will
 # be destroyed. If called in TUNNELED environment, this function die.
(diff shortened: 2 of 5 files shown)


lib/publiccloud/utils.pm
```diff
@@ -113,6 +113,17 @@ sub registercloudguest {
     record_info('registercloudguest time', 'The command registercloudguest took ' . (time() - $cmd_time) . ' seconds.');
 }
 
+sub register_openstack {
```
jlausuch, line 116: Shall we move this method to `lib/publiccloud/openstack.pm`?

jlausuch (approved):
This is awesome! Well done!

lib/publiccloud/utils.pm
mloviska: Oh, I haven't thought about that. Nice point. The only thing that pops to my mind is that we might have mixed the purpose of the that library. IMHO, `lib/publiccloud/openstack.pm` is more about provisioning routines than runtime tasks. OTOH, no objects to move it.

jlausuch: yes, also good point. Let's keep it as it is :)

data/publiccloud/log_instance.sh
```diff
@@ -121,6 +121,22 @@ azure_stop_log()
...
+    set +e
+    openstack server start "$INSTANCE_ID" > "${OUTPUT_DIR}/$CNT""_boot_log_start.txt" 2>&1
+    set -e
+}
+
+openstack_stop_log()
+{
+    read_unique_counter
+    set +e
+    openstack server start "$INSTANCE_ID" > "${OUTPUT_DIR}/$CNT""_boot_log_start.txt" 2>&1
```
b10n1k, line 136: was it meant to be `openstack server stop "$INSTANCE_ID" > "${OUTPUT_DIR}/$CNT""_boot_log_stop.txt" 2>&1`?

mloviska: Yup, as it is for others too

tests/publiccloud/run_ltp.pm
```diff
@@ -106,6 +107,7 @@ sub run {
     assert_script_run('chmod +x log_instance.sh');
 
     registercloudguest($instance) if (is_byos() && !$qam);
+    register_openstack($instance) if (is_openstack);
```
b10n1k, line 110: nitpicking: s/if (is_openstack);/if is_openstack;/

mloviska: fixed! Ty

data/publiccloud/log_instance.sh
b10n1k: just to clarify. i am not sure what do you mean /as it is for others too/. the command is exactly the same as in the `openstack_start_log`. in the `openstack_stop_log` you use `start` instead of `stop`, `pause` or whatever meant to be.

grisu48 (commented):
Nice addition to our workflow! If this continues at some point we probably need to rename `publiccloud` -> `hybridcloud` but that's for at some point in the future then.

good work!

data/publiccloud/log_instance.sh
mloviska: ah, overlooked the server command. ctrl+c & crtl+v, thanks!

mloviska:
> Nice addition to our workflow! If this continues at some point we probably need to rename `publiccloud` -> `hybridcloud` but that's for at some point in the future then.
> 
Yup, that might happen. Let's see what replacement of SOC we will get, if any.
> good work!

Thanks!

PR #8839, opened 2019-11-05
Add pam automated tests

badboywj:
This module is a suite of tests for PAM, and outputs
the TAP format result. It depends on "BATS"(a TAP-compliant
testing framework for Bash) and "pam-test"(to test a PAM stack
for authentication, https://github.com/pbrezina/pam-test)

- Related ticket: https://progress.opensuse.org/issues/51686
- Needles: no
- Verification run:
     - SLE12SP2: http://10.67.19.91/tests/867 :heavy_check_mark: 
     - SLE12SP3: http://10.67.19.91/tests/868 :heavy_check_mark: 
     - SLE12SP4: http://10.67.19.91/tests/869 :heavy_check_mark:
     - SLE12SP5: http://10.67.19.91/tests/871 :heavy_check_mark:
     - SLE15      : http://10.67.19.91/tests/870 :heavy_check_mark: 
     - SLE15SP1: http://10.67.19.91/tests/872 :heavy_check_mark:
     - Tumbleweed: http://10.67.19.91/tests/874 :heavy_check_mark:

PR #14663, opened 2022-04-06
Enable PAM test module for leap>=15.3

asdil12:
Ticket: https://progress.opensuse.org/issues/107434
Verification: http://kazhua.qa.suse.de/tests/96#

### DIFF
--- a/tests/console/pam.pm
+++ b/tests/console/pam.pm
@@ -29,7 +29,7 @@ sub run {
     my $self = shift;
     $self->select_serial_terminal;
 
-    die "This test module is not enabled for openSUSE Leap yet" if is_leap();
+    die "This test module is not enabled for openSUSE Leap yet" if is_leap('<15.3');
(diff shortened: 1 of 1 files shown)


tests/console/pam.pm
```diff
@@ -29,7 +29,7 @@ sub run {
     my $self = shift;
     $self->select_serial_terminal;
 
-    die "This test module is not enabled for openSUSE Leap yet" if is_leap();
+    die "This test module is not enabled for openSUSE Leap yet" if is_leap('<15.4');
```
foursixnine, line 32: I actually wonder, what happens with leap 15.3. I don't think it's gonna be the end of the world. (although past me, seems to not have said anything about that... looking at the review of #8839

asdil12: 15.3 works as well:
http://kazhua.qa.suse.de/tests/97

foursixnine (approved):

PR #14664, opened 2022-04-06
Fix sssd_389ds_functional

michaelgrifalconi:
Install awk and use a larger shm 

- Related ticket: https://progress.opensuse.org/issues/109473
- Needles: N/A
- Verification run: 
  - https://openqa.suse.de/tests/8476212#
  - https://openqa.suse.de/tests/8476275

### DIFF
--- a/tests/console/sssd_389ds_functional.pm
+++ b/tests/console/sssd_389ds_functional.pm
@@ -53,8 +53,8 @@ sub run {
     assert_script_run("mkdir /tmp/sssd && cd /tmp/sssd");
     assert_script_run("curl " . "--remote-name-all " . data_url("sssd/398-ds/{user_389.ldif,access.ldif,Dockerfile_$docker,instance_389.inf}"));
     assert_script_run(qq($docker build -t ds389_image --build-arg tag="$tag" --build-arg pkgs="$pkgs" -f Dockerfile_$docker .), timeout => 600);
-    assert_script_run("$docker run -itd --name ds389_container --hostname ldapserver --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro --restart=always ds389_image") if ($docker eq "docker");
-    assert_script_run("$docker run -itd --name ds389_container --hostname ldapserver ds389_image") if ($docker eq "podman");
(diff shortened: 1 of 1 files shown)


Avinesh (approved):
Looks fine!

punkioudi:
Hi Michael, my question here is if this base image used here (i think it's 15sp3, right?) is supposed to have awk. If it should, then what I suggest is to open a bug and then soft-fail it by installing awk :)

tests/console/sssd_389ds_functional.pm
```diff
@@ -53,8 +53,8 @@ sub run {
     assert_script_run("mkdir /tmp/sssd && cd /tmp/sssd");
     assert_script_run("curl " . "--remote-name-all " . data_url("sssd/398-ds/{user_389.ldif,access.ldif,Dockerfile_$docker,instance_389.inf}"));
     assert_script_run(qq($docker build -t ds389_image --build-arg tag="$tag" --build-arg pkgs="$pkgs" -f Dockerfile_$docker .), timeout => 600);
-    assert_script_run("$docker run -itd --name ds389_container --hostname ldapserver --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro --restart=always ds389_image") if ($docker eq "docker");
-    assert_script_run("$docker run -itd --name ds389_container --hostname ldapserver ds389_image") if ($docker eq "podman");
+    assert_script_run("$docker run -itd --shm-size=256m --name ds389_container --hostname ldapserver --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro --restart=always ds389_image") if ($docker eq "docker");
+    assert_script_run("$docker run -itd --shm-size=256m --name ds389_container --hostname ldapserver ds389_image") if ($docker eq "podman");
```
foursixnine, line 57: We need to run this test with a bit more ram.

foursixnine:
> Hi Michael, my question here is if this base image used here (i think it's 15sp3, right?) is supposed to have awk. If it should, then what I suggest is to open a bug and then soft-fail it by installing awk :)

@dcermak wdyt?

dcermak:
> > Hi Michael, my question here is if this base image used here (i think it's 15sp3, right?) is supposed to have awk. If it should, then what I suggest is to open a bug and then soft-fail it by installing awk :)
> 
> @dcermak wdyt?

The image is supposed to be small, it not having awk is certainly not a bug but a feature :wink:

punkioudi (approved):
Looks cool!

PR #14665, opened 2022-04-06
gcc: fix JIT library dependency name

marxin:
Install libgccjit-devel-gcc package and do not hardcode
an exact version of the package.

@grisu48


marxin:
The test currently fails here:
https://openqa.opensuse.org/tests/2284026/modules/gcc/steps/1/src

DimStar77:
VR triggered: https://openqa.opensuse.org/tests/2284204 (JeOS case, it's slightly shorter)

PR #14670, opened 2022-04-07
Extend time for susefirewall2-to-firewalld command

tinawang123:
Extend time for susefirewall2-to-firewalld command

- Related ticket: https://progress.opensuse.org/issues/109524
- Verification run: http://openqa.suse.de/t8475027

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -2118,8 +2118,7 @@ sub assert_secureboot_status {
 }
 
 sub susefirewall2_to_firewalld {
-    my $timeout = 360;
-    $timeout = 1200 if is_aarch64;
+    my $timeout = (is_aarch64 || is_ppc64le) ? '1200' : '360'; 
     assert_script_run('susefirewall2-to-firewalld -c', timeout => $timeout);
     assert_script_run('firewall-cmd --permanent --zone=external --add-service=vnc-server', timeout => 60);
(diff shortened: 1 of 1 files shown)


hjluo:
LGTM

lib/utils.pm
```diff
@@ -2118,8 +2118,7 @@ sub assert_secureboot_status {
 }
 
 sub susefirewall2_to_firewalld {
-    my $timeout = 360;
-    $timeout = 1200 if is_aarch64;
+    my $timeout = 1200;
```
lemon-suse, line 2121: How long need to wait on power? If don't need 1200s, could we set it as $timeout = xxx if is_power?

coolgw, line 2121: Your VR show 360s needed on power, what's timeout for other platform? such as 390?

tinawang123: Not sure, just run many cases it will fail. I think we should use platform at here.

tinawang123: I don't think it need so many condition to set timeout. As if not need 1200s, when it finished, it will go on next code.

rfan1, line 2121: my $timeout = is_aarch64 ? '1200' : '360';

rfan1 (approved):
LGTM

okurz (changes requested):
Please fix the git commit author. It says "openQA web UI"

PR #14675, opened 2022-04-07
Increase timeout to list vmcores

mloviska:
[Failure](https://openqa.suse.de/tests/8476055#step/kdump_and_crash/66)
occures often on hyperv test runs.

- Verification run [sle-15-SP3-JeOS-for-MS-HyperV-QR-x86_64-Build4.10.3-jeos-extratest@svirt-hyperv](https://openqa.suse.de/t8482094)

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -351,7 +351,7 @@ sub check_function {
 
     if ($args{test_type} eq 'function') {
         # Check, that vmcore exists, otherwise fail
-        assert_script_run('ls -lah /var/crash/*/vmcore');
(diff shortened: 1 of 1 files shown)


lib/kdump_utils.pm
```diff
@@ -351,7 +351,7 @@ sub check_function {
 
     if ($args{test_type} eq 'function') {
         # Check, that vmcore exists, otherwise fail
-        assert_script_run('ls -lah /var/crash/*/vmcore');
+        assert_script_run('ls -lah /var/crash/*/vmcore', 120);
```
jlausuch, line 354: default is 90s, maybe to be more in the safe side, shall we double the default? so `180`?

PR #14681, opened 2022-04-08
Update scsi id boot from slof for ppc64le

tinawang123:
The scsi device id changed from 6 to 7. So we need to updated the scsi boot id from 6 to 7.
- Related ticket:https://progress.opensuse.org/issues/109692
- Verification run: https://openqa.suse.de/tests/8490219#
   https://openqa.suse.de/tests/8490248


tinawang123:
@okurz   Please take a look. Not sure why it changed

okurz (dismissed):

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

Also, why adding 2 in case of virtio-console is not obvious to me. Is this valid for all product versions? Instead of applying such magic number please consider a more robust automatic resolution of the necessary device

tinawang123:
@okurz  At grub shell, we cannot use command to get scsi id. Refer case: https://openqa.suse.de/tests/8502556#step/boot_to_desktop/6. This blocked all of my powerKVM cases.

rfan1:
@tinawang123 

I would suggest you find the RC why the disk id is changed before merging this request

1. Is it caused by new qemu backend device changes?
2. Is it caused by new ISO image [new kernel dirvers loaded process]?
3. Can we do some more re-try logic to make sure it can boot the "right" disk device before fail cases there? [e.g. try scsi6, 7, and 8 
]

tinawang123:
> @tinawang123
> 
> I would suggest you find the RC why the disk id is changed before merging this request
> 
>     1. Is it caused by new qemu backend device changes?
> 
>     2. Is it caused by new ISO image [new kernel dirvers loaded process]?
> 
>     3. Can we do some more re-try logic to make sure it can boot the "right" disk device before fail cases there? [e.g. try scsi6, 7, and 8
>        ]

We can investigate those situation, but now it blocked all my powerKVM cases. I need fix it first.  I can add soft fail at here for workaround now.

lemon-suse:
In fact, it is clearly insert a device which caused the scsi id increated from 6->7.
The device info on 119.1 https://openqa.suse.de/tests/8450987/logfile?filename=serial0.txt
Populating /pci@800000020000000
                     00 0000 (D) : 1234 1111    qemu vga
                     00 0800 (D) : 8086 2668    multimedia-device*
                     00 1000 (D) : 1af4 1000    virtio [ net ]
                     00 1800 (D) : 1033 0194    serial bus [ usb-xhci ]
                     00 2000 (D) : 1af4 1003    virtio [ serial ]
                     00 2800 (D) : 1af4 1004    virtio [ scsi ]
The device info on 124.1: https://openqa.suse.de/tests/8486442/logfile?filename=serial0.txt
Populating /pci@800000020000000
                     00 0000 (D) : 1234 1111    qemu vga
                     00 0800 (D) : 8086 2668    multimedia-device*
                     00 1000 (D) : 1af4 1000    virtio [ net ]
                     00 1800 (D) : 1af4 1005    legacy-device*
                     00 2000 (D) : 1033 0194    serial bus [ usb-xhci ]
                     00 2800 (D) : 1af4 1003    virtio [ serial ]
                     00 3000 (D) : 1af4 1004    virtio [ scsi ]

So the new inserted device is ' 00 1800 (D) : 1af4 1005    legacy-device*'.

Compared the openQA cmd, I found since 124.1 it added a new device, '-device virtio-rng-pci,rng=rng0', I think it is this device inserted caused scsi id need change from 6 to 7.

foursixnine:
> Use setting "QEMU_VIRTIO_RNG=0 " can workaround this problem. But we should fix this one with correct scsi ID.

I Guess https://progress.opensuse.org/issues/109980#note-4 may be related, but indeed @tinawang123 it should help for now.

> In fact, it is clearly insert a device which caused the scsi id increated from 6->7. The device info on 119.1 
> ...
> Compared the openQA cmd, I found since 124.1 it added a new device, '-device virtio-rng-pci,rng=rng0', I think it is this device inserted caused scsi id need change from 6 to 7.

Is it possible that the qemu versions are different and are causing the problem? but also there's the change in the bootloader (see progress ticket)

coolgw:
@foursixnine  @okurz  Is there any smart way detect scsi id in grub mode?

okurz:
> @foursixnine @okurz Is there any smart way detect scsi id in grub mode?

I don't know any specific way. I suggest you research how qemu behaves and what should be expected from that.

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
         if (match_has_tag 'inst-slof') {
             diag 'specifying local disk for boot from slof';
             my $slof = 5;
-            $slof += 1 if get_var('VIRTIO_CONSOLE');
+            if (get_var('VIRTIO_CONSOLE') && is_ppc64le) {
+                type_string_very_slow "devalias";
+                send_key 'ret';
+                save_screenshot;
+                my $file = './serial0';
```
rfan1, line 257: Can you explain more on this file? As I know it is defined in qemu backend
` "-chardev ringbuf,id=serial0,logfile=serial0,logappend=on -serial chardev:serial0"`

My question is that, does this configuration have something to do with VIRTIO_CONSOLE?

2nd question, if we are in the same directory name as serial0 file, can we just set
`my $file = serial0`

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
...
+                open (FILE, $file);
+                my @content = <FILE>;
+                close FILE;
+                my $out = 0;
+                foreach my $line (@content) {
+                     if ($line =~ /^disk/) {
```
coolgw, line 263: what's diff between "disk" and "scsi" line?

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
...
+                         $line =~ s/\0|\r|\n//g;
+                         my @num = split(/scsi@/, $line);
+                         $out = $num[1];
```
coolgw, line 266: maybe we need break loop if we get the num

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
         if (match_has_tag 'inst-slof') {
             diag 'specifying local disk for boot from slof';
             my $slof = 5;
-            $slof += 1 if get_var('VIRTIO_CONSOLE');
+            if (get_var('VIRTIO_CONSOLE') && is_ppc64le) {
```
lemon-suse, line 253: Don't need is_ppc64le since this is in the condition of 'if (get_var('OFW'))', OFW is for ppc64le.

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
...
+                my $out = 0;
+                foreach my $line (@content) {
+                     if ($line =~ /^disk/) {
+                         $line =~ s/\0|\r|\n//g;
+                         my @num = split(/scsi@/, $line);
+                         $out = $num[1];
+                     }
+                }
+                $slof = $out;
+            }
```
rfan1, line 270: I think it is a cool method to find the scsi id here, however minor comment here [may suitable for big file :)]

open(FILE, $file) || die"cannot open the file: $!\n";
while (<FILE>){
       print;
}
close FILE;

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,24 @@ sub boot_local_disk {
         if (match_has_tag 'inst-slof') {
             diag 'specifying local disk for boot from slof';
             my $slof = 5;
-            $slof += 1 if get_var('VIRTIO_CONSOLE');
+            if (get_var('VIRTIO_CONSOLE') && is_ppc64le) {
+                type_string_very_slow "devalias";
```
lemon-suse, line 254: Not sure whether here support '\n' as return, if support then don't need send key 'ret' seperately.

lib/bootloader_setup.pm
lemon-suse: Agree, besides in my opinion to find the id in so long a file as serial0 will need long time, so I think it is a successful solution while maybe not the best solution. :)

lib/bootloader_setup.pm
tinawang123: disk : /pci@800000020000000/scsi@7

foursixnine: @tinawang123 one open question for a possible follow up on this PR, is: what happens when the system has more than one iscsi device? :)... but that's for later I guess

lib/bootloader_setup.pm
```diff
@@ -249,7 +250,28 @@ sub boot_local_disk {
...
+                save_screenshot;
+		#my $file = './serial0';
+		#open (FILE, $file);
+		#my @content = <FILE>;
+		#close FILE;
+		#my $out = 0;
+		#foreach my $line (@content) {
+		#     if ($line =~ /^disk/) {
+		#         $line =~ s/\0|\r|\n//g;
+		#         my @num = split(/scsi@/, $line);
```
coolgw, line 265: suggest using following one line get $num
my $num = ($line=~ m/scsi@(\d+)/);

coolgw (approved):

lib/bootloader_setup.pm
```diff
@@ -227,6 +228,22 @@ sub boot_grub_item {
...
+sub get_scsi_id {
+    my $sid;
+    type_string_very_slow "devalias";
+    send_key 'ret';
+    save_screenshot;
+    my $file = './serial0';
+    open(my $fh, '<', "./serial0");
+    while (my $line = <$fh>) {
+        if ($line =~ /^disk/) {
+            $sid = ($line =~ m/scsi@(\d+)/)[0];
```
hjluo, line 240: open FILE, '<', $file_name or die $!; this is the standard to handle error
if ($line =~ /^disk/) 
should match start with 'disk : /pci' and have scsi in the string
once get sid, we should check if it's a number or not

lib/bootloader_setup.pm
```diff
@@ -227,6 +228,22 @@ sub boot_grub_item {
     grub_key_enter;
 }
 
+sub get_scsi_id {
+    my $sid;
+    type_string_very_slow "devalias";
+    send_key 'ret';
+    save_screenshot;
+    my $file = './serial0';
+    open(my $fh, '<', "./serial0");
```
foursixnine, line 237: As pointed out by @hjluo 
```suggestion
    open(my $fh, '<', "./serial0") or die $!;
```

lib/bootloader_setup.pm
```diff
@@ -227,6 +228,22 @@ sub boot_grub_item {
...
+    while (my $line = <$fh>) {
+        if ($line =~ /^disk/) {
+            $sid = ($line =~ m/scsi@(\d+)/)[0];
+            last;
+        }
```
foursixnine, line 242: Maybe I'm reading the code wrong, but we want scsi and not disks right?

```suggestion
       next if ($line =~ /^disk/);
       if ($line =~ /^scsi/) {
           $sid = ($line =~ m/scsi@(\d+)/)[0];
           last if $id;
        }
    }
```

foursixnine (changes requested):

lib/bootloader_setup.pm
foursixnine: @tinawang123 ahhh! I need more coffee :)

rfan1 (approved):

lib/bootloader_setup.pm
hjluo: LGTM

hjluo (approved):
LGTM

coolgw:
@okurz  could you help check again?

okurz (commented):
looks nice, thank you

PR #14683, opened 2022-04-08
Migrate create_hdd_gnome in Functional group to use AutoYaST

ge0r:
- Related ticket: https://progress.opensuse.org/issues/107674
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/overview?version=15-SP4&build=ge0r%2Fos-autoinst-distri-opensuse%23autoyast_create_hdd_gnome&distri=sle

Related MRs:
 * https://gitlab.suse.de/qsf-u/qa-sle-functional-userspace/-/merge_requests/151
 * https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/404

## Synopsis

We have several openQA testsuites that are performing installation in order to publish an hdd (.qcow) image.
They do this by going through all the steps of the installer.
These testsuites can however be adapted to use autoyast in order to perform a simpler and faster installation.
The following steps make up an example workflow that migrates such a testsuite (`create_hdd_gnome`) to using autoyast, in order to create an hdd image.

For this workflow we will be using autoyast profiles in their compact version. This makes them half in size and less complex to go through.

## Example workflow
There are two parts in migrating the `create_hdd_gnome` testsuite to use autoyast:
1. Acquire the autoyast profile from the testsuite's installed system.
2. Convert the autoyast profile to a generic one that can be used by testsuite runs.
3. Adapt the testsuite to use autoyast for the installation.

These two parts are broken down in smaller steps. 
Let's check them out.

##### 1. Acquire the (compact) autoyast profile of `create_hdd_gnome`:

* Modify `tests/autoyast/clone.pm` so that it generates a compact autoyast profile.
    We do this by passing `modules target=compact` as argument to the `yast2_console_exec` function, as shown [here](https://github.com/ge0r/os-autoinst-distri-opensuse/blob/clone-create_hdd-autoyast/tests/autoyast/clone.pm#L21).
* Schedule the `clone.pm` module after the system has been installed in the `create_hdd_gnome` testsuite.
    For instance [here](https://github.com/ge0r/os-autoinst-distri-opensuse/blob/clone-create_hdd-autoyast/lib/main_common.pm#L1943) it is placed right after the `system_prepare` module.
* For each available architecture, run the `create_hdd_gnome` testsuite with the aforementioned changes. The `clone.pm` module will fail when uploading a second autoyast profile, but this is expected. Our autoyast profile `clone-autoinst.xml` will be now available in the *Logs & Assets* tab.

##### 2. Convert the autoyast profile to a generic one:
In order to use the autoyast profile in testsuite runs we need to do some modifications:
* Sections `add_on` and `append` need to be removed.
* The text in section `reg_code` contains the registration code for the equivalent product and needs to be replaced with `{{SCC_REGCODE}}` or equivalent variable (e.g. `{{SCC_REGCODE_HA}}` or `{{SCC_REGCODE_WE}}`).
* The text in section `reg_server` needs to be replaced with `{{SCC_URL}}`.
* The text in section `arch` needs to be replaced with `{{ARCH}}`.
* The text in section `version` needs to be replaced with `{{VERSION}}`.
* The text in section `ipaddr` (if the section exists), needs to be replaced with `{{HostIP}}`.
* In case of Z/VM (setting BACKEND is s390x), remove all listentry channels except for the one used in installation (most likely 0.0.0150) and also remove the `hosts_entry` section that contains ip. FInally remove `hostname` sections.

Alternatively you could use [this script to do that](https://github.com/ge0r/autoyast-profile-converter) but depending on the profile, you might need to do some additional changes.

##### 3. Adapt the testsuite to use autoyast for the installation:
* Add the modified autoyast profile e.g. to `data/autoyast_sle15/create_hdd`.
* Add a new yaml schedule for the testsuite that will create the hdd. The yaml should set the `AUTOYAST` parameter to the afforementioned autoyast profile, and contain the `autoyast/installation` module to perform the autoyast installation.
* Create a new testsuite definition in openqa.suse.de, similar to the one of the testsuite you are replacing. For example I created `autoyast_create_hdd_gnome` as a replacement to the `create_hdd_gnome` testuite.
* Add this new testsuite to your job group.

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_gnome_aarch64.xml
+++ b/data/autoyast_sle15/create_hdd/create_hdd_gnome_aarch64.xml
@@ -1,0 +1,421 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <bootloader t="map">
+    <global t="map">
+      <cpu_mitigations>auto</cpu_mitigations>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <secure_boot>true</secure_boot>
+      <terminal>gfxterm</terminal>
+      <timeout t="integer">-1</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>vga=gfx-1024x768x16</xen_kernel_append>
+    </global>
+    <loader_type>grub2-efi</loader_type>
+  </bootloader>
+  <firewall t="map">
+    <default_zone>public</default_zone>
+    <enable_firewall t="boolean">true</enable_firewall>
+    <log_denied_packets>off</log_denied_packets>
+    <start_firewall t="boolean">true</start_firewall>
+    <zones t="list">
+      <zone t="map">
+        <description>For use in public areas. You do not trust the other computers on networks to not harm your computer. Only selected incoming connections are accepted.</description>
+        <interfaces t="list">
+          <interface>eth0</interface>
+        </interfaces>
+        <masquerade t="boolean">false</masquerade>
+        <name>public</name>
+        <ports t="list"/>
+        <protocols t="list"/>
+        <services t="list">
+          <service>dhcpv6-client</service>
+        </services>
+        <short>Public</short>
+        <target>default</target>
+      </zone>
+    </zones>
+  </firewall>
+  <general t="map">
+    <mode t="map">
+      <confirm t="boolean">false</confirm>
+    </mode>
+  </general>
+  <groups t="list">
+    <group t="map">
+      <gid>100</gid>
+      <groupname>users</groupname>
+      <userlist/>
+    </group>
+  </groups>
+  <host t="map">
+    <hosts t="list">
+      <hosts_entry t="map">
+        <host_address>127.0.0.1</host_address>
+        <names t="list">
+          <name>localhost</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>::1</host_address>
+        <names t="list">
+          <name>localhost ipv6-localhost ipv6-loopback</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>fe00::0</host_address>
+        <names t="list">
+          <name>ipv6-localnet</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff00::0</host_address>
+        <names t="list">
+          <name>ipv6-mcastprefix</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff02::1</host_address>
+        <names t="list">
+          <name>ipv6-allnodes</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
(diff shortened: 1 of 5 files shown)


b10n1k:
@ge0r Epeidi ola ta xmls fainontai ta idia, des an se endiaferei na kaneis kati paromoio https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/containers/autoyast_containers.xml.ep

data/autoyast_sle15/create_hdd/create_hdd_gnome_aarch64.xml
```diff
@@ -0,0 +1,421 @@
...
+        <min/>
+        <warn/>
+      </password_settings>
+      <shell>/bin/bash</shell>
+      <uid>0</uid>
+      <user_password>$6$mhonb/3TIuegSE4m$l/1CZdAxYWvV8KwIWwQxXxo5JOh9IGOu11VMnd2XHuJGQzXnN.rN6LGv5189751zOYm/ommDfJtp3UzK300QI0</user_password>
+      <username>root</username>
+    </user>
+  </users>
+</profile>
```
jknphy, line 421: you can configure your editor (vscode?) with `Insert Final Newline`.

schedule/functional/autoyast_create_hdd/autoyast_create_hdd_gnome.yaml
```diff
@@ -0,0 +1,17 @@
+name: autoyast_create_hdd_gnome
+description:    >
+  Placeholder
```
jknphy, line 3: missing description, just to not forget :) when you will remove the draft.

jknphy:
@ge0r  Nice documentation for other teams to follow, very clear.
As we agreed let's not improve the profiles with ERB or other things for now, we can do it in the future.
Is it ready to merge except the part for hyperv? if so, I would recommend to merge it also adding here the MR for job groups for functional group. In a different PR we can deal with the svirt-xen ones. wdyt?

ge0r:
> @ge0r Nice documentation for other teams to follow, very clear. As we agreed let's not improve the profiles with ERB or other things for now, we can do it in the future. Is it ready to merge except the part for svirt-xen? if so, I would recommend to merge it also adding here the MR for job groups for functional group. In a different PR we can deal with the svirt-xen ones. wdyt?

I agree. I ll bring this up to date and remove the draft

jknphy:
In OSD we point to x86_64 for this new test suite `autoyast_create_hdd_gnome`.

ge0r:
> In OSD we point to x86_64 for this new test suite `autoyast_create_hdd_gnome`. Do you have a list of all the test suite where we need to update `START_AFTER_TEST` (some of them will be changed directly in openQA, but other needs to be changed in corresponding repos)?

Yes but those will be given as a guidance in the progress ticket, as handover to the functional team

jknphy:
Agreed offline to connect only the jobs in YaST group unconnecting them from existing create_hdd_image in Functional job group and find some different name for the published qcow2.

schedule/functional/autoyast_create_hdd/autoyast_create_hdd_gnome.yaml
```diff
@@ -0,0 +1,23 @@
+name: autoyast_create_hdd_gnome
+description:    >
+  Test performs autoyast installation to generate qcow images used by dependent functional tests.
+conditional_schedule:
+  svirt_upload_assets:
+    ARCH:
+      's390x':
```
jknphy, line 7: no need for quotes.

PR #12483, opened 2021-05-06
ipxe: Set EFI options for pxe bootdev

czerw:
Fix poo#92203: If previous job/test changed boot order, new
installation job didn't boot from network. Ipmitool command must
contain options=efiboot to correctly set EFI boot order.

- Related ticket: https://progress.opensuse.org/issues/92203
- Needles: none
- Verification run: 
aarch64 efi: https://openqa.suse.de/tests/5961311
x86_64 legacy: https://openqa.suse.de/tests/5961310

PR #14684, opened 2022-04-08
[WIP]Support x86_64 UEFI pxe boot

Julie-CAO:
The new DELL machines, vh080/081, support UEFI boot only.
https://suse.slack.com/archives/C02C4BMR0PR/p1646038242478279

- Related ticket: https://progress.opensuse.org/issues/109157
- Needles: https://openqa.nue.suse.com/tests/8485398/modules/boot_from_pxe/steps/3/edit
- Verification run: 
[sp4_host_x86_uefi_local](http://10.67.129.102/tests/180) PASS
[sp4_host_Install_x86_non_uefi_OSD](https://openqa.suse.de/tests/8568432)  PASS
[sp3_kvm_host_install_aarch64_local_openqa](http://10.67.129.102/tests/189)  boot module pass but login fail not due to the PR
[sp4_host_Install_aarch64_OSD](https://openqa.suse.de/tests/8568694)  PASS
[12sp5host_install_aarch64_OSD](http://openqa.suse.de/tests/8573709)  PASS

@alice-suse @guoxuguang @waynechen55 @nanzhg

### DIFF
--- a/lib/ipmi_backend_utils.pm
+++ b/lib/ipmi_backend_utils.pm
@@ -264,6 +264,21 @@ sub adjust_for_ipmi_xen {
     }
 }
 
+# Set boot from pxe via ipmitool
+sub set_pxe_boot {
+    while (1) {
+        my $stdout = ipmitool('chassis bootparam get 5');
+        if ($stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/) {
+            record_info("Current boot device is PXE", "Current boot device is $stdout");
+            last;
+        }
+        diag "setting boot device to pxe";
+        my $options = get_var('IPXE_UEFI') ? 'options=efiboot' : '';
+        ipmitool("chassis bootdev pxe ${options}");
+        sleep(3);
+    }
+}
+
 sub set_pxe_efiboot {
     my ($root_prefix) = @_;
     $root_prefix = "/" if (!defined $root_prefix) || ($root_prefix eq "");
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -1172,7 +1172,7 @@ sub wait_boot {
         $self->handle_pxeboot(bootloader_time => $bootloader_time, pxemenu => 'pxe-custom-kernel', pxeselect => 'pxe-custom-kernel-selected');
     }
     else {
-        assert_screen([qw(virttest-pxe-menu qa-net-selection prague-pxe-menu pxe-menu)], 600) if (uses_qa_net_hardware() || get_var("PXEBOOT"));
+        assert_screen([qw(virttest-pxe-menu qa-net-selection qa-net-selection-uefi prague-pxe-menu pxe-menu)], 600) if (uses_qa_net_hardware() || get_var("PXEBOOT"));
         $self->handle_grub(bootloader_time => $bootloader_time, in_grub => $in_grub);
         # part of soft failure bsc#1118456
         if (get_var('UEFI') && is_hyperv) {
--- a/tests/boot/boot_from_pxe.pm
+++ b/tests/boot/boot_from_pxe.pm
@@ -19,6 +19,7 @@ use utils qw(type_string_slow enter_cmd_slow);
 use Utils::Backends;
 use Utils::Architectures;
 use version_utils qw(is_upgrade is_sle);
+use ipmi_backend_utils 'set_pxe_boot';
 
 sub run {
     my ($image_path, $image_name, $cmdline);
@@ -31,6 +32,8 @@ sub run {
         resume_vm();
     }
 
+    set_pxe_boot if get_var('UEFI_PXE_BOOT');
+
     if (is_ipmi) {
         if (is_remote_backend && is_aarch64 && get_var('IPMI_HW') eq 'thunderx') {
             select_console 'sol', await_console => 1;
@@ -41,10 +44,10 @@ sub run {
             select_console 'sol', await_console => 0;
         }
     }
-    if (!check_screen([qw(virttest-pxe-menu qa-net-selection prague-pxe-menu pxe-menu)], 600)) {    # nocheck: old code, should be updated
+    if (!check_screen([qw(virttest-pxe-menu qa-net-selection qa-net-selection-uefi prague-pxe-menu pxe-menu)], 600)) {    # nocheck: old code, should be updated
         ipmi_backend_utils::ipmitool 'chassis power reset';
     }
-    assert_screen([qw(virttest-pxe-menu qa-net-selection prague-pxe-menu pxe-menu)], 600);
+    assert_screen([qw(virttest-pxe-menu qa-net-selection qa-net-selection-uefi prague-pxe-menu pxe-menu)], 600);
 
     # boot bare-metal/IPMI machine
     if (is_ipmi && get_var('BOOT_IPMI_SYSTEM')) {
@@ -60,7 +63,10 @@ sub run {
 
         $image_path = get_var("HOST_IMG_URL");
     }
-    elsif (match_has_tag("qa-net-selection")) {
+    # the tags, 'qa-net-selection' & 'qa-net-selection-uefi', are expected to be used
+    # in the needles for UEFI boot menu in QA-NET, though they are treated differently here
+    # because 'qa-net-selection' is widely used
+    elsif (match_has_tag("qa-net-selection") or match_has_tag("qa-net-selection-uefi")) {
         if (check_var("INSTALL_TO_OTHERS", 1)) {
             $image_name = get_var("REPO_0_TO_INSTALL");
         }
@@ -72,7 +78,7 @@ sub run {
         $openqa_url = 'http://' . $openqa_url unless $openqa_url =~ /http:\/\//;
         my $repo = $openqa_url . "/assets/repo/${image_name}";
         my $key_used = '';
-        if (is_remote_backend && is_aarch64 && is_supported_suse_domain) {
+        if ((is_remote_backend && is_aarch64 && is_supported_suse_domain) or match_has_tag("qa-net-selection-uefi")) {
             $key_used = 'c';
             send_key 'down';
         }
@@ -93,6 +99,7 @@ sub run {
             my $path_prefix = "auto/openqa/repo";
             $path_prefix = "/mnt/openqa/repo" if (!is_orthos_machine);
             my $path = "${path_prefix}/${image_name}/boot/${arch}";
+            $path .= "/loader" if is_x86_64 && !is_orthos_machine;
             $image_path = "linux $path/linux install=$repo";
         }
 
@@ -187,7 +194,11 @@ sub run {
             if (match_has_tag("orthos-grub-boot-linux") or match_has_tag("qa-net-grub-boot-linux")) {
                 my $image_name = eval { check_var("INSTALL_TO_OTHERS", 1) ? get_var("REPO_0_TO_INSTALL") : get_var("REPO_0") };
                 my $args = "initrd auto/openqa/repo/${image_name}/boot/${arch}/initrd";
-                $args = "initrd /mnt/openqa/repo/${image_name}/boot/${arch}/initrd" if (!is_orthos_machine);
+                if (!is_orthos_machine) {
+                    $args = "initrd /mnt/openqa/repo/${image_name}/boot/${arch}";
+                    $args .= "/loader" if is_x86_64;
+                    $args .= "/initrd";
+                }
                 type_string_slow $args;
                 send_key 'ret';
                 #Detect orthos-grub-boot-initrd and qa-net-grub-boot-initrd for aarch64 in orthos and openQA networks respectively
--- a/tests/installation/ipxe_install.pm
+++ b/tests/installation/ipxe_install.pm
@@ -44,17 +44,6 @@ sub poweron_host {
     }
 }
 
-sub set_pxe_boot {
-    while (1) {
-        my $stdout = ipmitool('chassis bootparam get 5');
-        last if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/;
-        diag "setting boot device to pxe";
-        my $options = get_var('IPXE_UEFI') ? 'options=efiboot' : '';
-        ipmitool("chassis bootdev pxe ${options}");
-        sleep(3);
-    }
-}
-
 sub set_bootscript {
     my $host = get_required_var('SUT_IP');
     my $ip = inet_ntoa(inet_aton($host));
--- a/tests/installation/logs_from_installation_system.pm
+++ b/tests/installation/logs_from_installation_system.pm
@@ -71,7 +71,6 @@ sub run {
             set_grub_on_vh('/mnt', '', 'xen') if (get_var('XEN') || check_var('HOST_HYPERVISOR', 'xen'));
             set_grub_on_vh('/mnt', '', 'kvm') if (check_var('HOST_HYPERVISOR', 'kvm') || check_var('SYSTEM_ROLE', 'kvm'));
             adjust_for_ipmi_xen('/mnt') if (get_var('REGRESSION') && (get_var('XEN') || check_var('HOST_HYPERVISOR', 'xen')));
-            set_pxe_efiboot('/mnt') if is_aarch64;
         }
     }
     else {


alice-suse:
I do not have specific suggestions. But since this PR changes many conditions for pxe boot commandline, I would suggest more validation jobs and invite more people to review, especially for those tests out of VT or testers, which/who may be impacted.

Julie-CAO:
> I do not have specific suggestions. But since this PR changes many conditions for pxe boot commandline, I would suggest more validation jobs and invite more people to review, especially for those tests out of VT or testers, which/who may be impacted.

It is expected to involve x86/arm tests with ipmi backend. but in case of some side effect, I'll post in testing channel for wider audience.

rfan1:
@Julie-CAO

I have one minor comment here, for x86 uefi mode, Do you have any test module to check your SUT is UEFI boot?

Julie-CAO:
> @Julie-CAO
> 
> I have one minor comment here, for x86 uefi mode, Do you have any test module to check your SUT is UEFI boot?

No. only when the UEFI grub menu shows up and is found out by the needle match, the machine will boot up based on the UEFI menu.

waynechen55:
This implementation means both x86_64 and aarch64 use the same needle which means the same screenshot/matched area and the name tag. Different screenshot/matched area but with the same tag does not guarantee both x86_64 and aarch64 test run pass as usual. This can happen if you there is needle mismatch and you re-captured a new needle for x86_64 which ends up two different needles with the same tag, so test run may not match the desired one for both arch. I suggest you use different tag names, for example:
```
if (match_has_tag(['this one', 'that one'])
```

waynechen55:
I think you can make following changes to make your implementation better
````
assert_screen([qw(virttest-pxe-menu qa-net-selection qa-net-selection-uefi prague-pxe-menu pxe-menu)], 600);

elsif (match_has_tag("qa-net-selection") or match_has_tag("qa-net-selection-uefi")) {
```

waynechen55:
I think you can make following changes to make your implementation better

assert_screen([qw(virttest-pxe-menu qa-net-selection qa-net-selection-uefi prague-pxe-menu pxe-menu)], 600);

elsif (match_has_tag("qa-net-selection") or match_has_tag("qa-net-selection-uefi")) {

        if (is_remote_backend && is_aarch64 && is_supported_suse_domain) or match_has_tag("qa-net-selection-uefi") {

```

**Comment by mdoucha** [comment]
> No. only when the UEFI grub menu shows up and is found out by the needle match, the machine will boot up based on the UEFI menu.

Jobs which are supposed to boot from IPXE in UEFI mode must have `IPXE_UEFI=1` in job settings.

**Comment by Julie-CAO** [comment]
> > No. only when the UEFI grub menu shows up and is found out by the needle match, the machine will boot up based on the UEFI menu.
> 
> Jobs which are supposed to boot from IPXE in UEFI mode must have `IPXE_UEFI=1` in job settings.

The machines will boot from 'PXE' instead of 'IPXE', so it is a new bootup method for x86_64 machines.

**Comment by mdoucha** [comment]
I've tested the PR on Prague PXE setup: http://openqa.qam.suse.cz/tests/40781
Don't mind the failure at the end, the fix for that is in PR #14688
LGTM.

**Comment by Julie-CAO** [comment]
@mdoucha and @czerw, I moved the set_pxe_boot function from ipxe_install.pm to lib/ as we need it as well. Could you help review? and Could you provide tests that I can verify in case it blocks your tests with ipxe bootup?

**Comment by mdoucha** [comment]
> @mdoucha and @czerw, I moved the set_pxe_boot function from ipxe_install.pm to lib/ as we need it as well. Could you help review? and Could you provide tests that I can verify in case it blocks your tests with ipxe bootup?

Verification run booted successfully without UEFI: http://openqa.qam.suse.cz/tests/41369

But please don't forget to add the `qa-net-selection-uefi` needle tag to `opensusebasetest::wait_boot()` right before the call to `handle_grub()`, otherwise UEFI IPMI jobs will fail to reboot from disk.

**Comment by czerw** [comment]
Please do verification on this job, it is aarch64 and already is UEFI: https://openqa.suse.de/tests/8569132

**Comment by Julie-CAO** [comment]
> Verification run booted successfully without UEFI: http://openqa.qam.suse.cz/tests/41369

Thanks for the verification.

> But please don't forget to add the `qa-net-selection-uefi` needle tag to `opensusebasetest::wait_boot()` right before the call to `handle_grub()`, otherwise UEFI IPMI jobs will fail to reboot from disk.

Just added it. thanks.

**Comment by Julie-CAO** [comment]
> Please do verification on this job, it is aarch64 and already is UEFI: https://openqa.suse.de/tests/8569132

verification run passed. https://openqa.suse.de/tests/8580043

**Comment by Julie-CAO** [comment]
Verification runs passed. PR is ready. Welcome review & merge.
@alice-suse @guoxuguang @waynechen55 @nanzhg

File: lib/ipmi_backend_utils.pm [hunk 1]
```diff
@@ -262,6 +262,18 @@ sub adjust_for_ipmi_xen {
     }
 }
 
+# Set boot from pxe via ipmitool
+sub set_pxe_boot {
```
alice-suse, line 266: This function has similarity with the following function "sub set_pxe_efiboot" . The latter seems to handle efiboot with efibootmgr and call handle multiple NIC pxe boot situation. I am wondering is it possible to merge/reuse that function? 
Besides, regarding set_pxe_boot function, can it survive on machines with two nic pxe boot items? And how about it on ipmi machines with different vendors, eg supermicro, dell, etc.

waynechen55:
After this change, uefi host always boot from local disk directly except for boot_from_pxe module. This means assert pxe menu is only feasible in boot_from_pxe module. Correct ? @Julie-CAO

lib/ipmi_backend_utils.pm
```diff
@@ -262,6 +262,18 @@ sub adjust_for_ipmi_xen {
...
+    while (1) {
+        my $stdout = ipmitool('chassis bootparam get 5');
+        last if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/;
+        diag "setting boot device to pxe";
```
waynechen55, line 270: I think you'd better use record-info here.

lib/ipmi_backend_utils.pm
```diff
@@ -262,6 +262,18 @@ sub adjust_for_ipmi_xen {
...
+        my $options = get_var('IPXE_UEFI') ? 'options=efiboot' : '';
+        ipmitool("chassis bootdev pxe ${options}");
+        sleep(3);
```
waynechen55, line 273: Another record_info is better:
```
record_info("boot device is pxe ${options}", "This is pxe ${options} boot");
```

waynechen55, line 273: Do double check after  ipmitool("chassis bootdev pxe ${options}") instead of sleep 3 ? For example,
```
ipmitool(sol info)
```

lib/ipmi_backend_utils.pm
```diff
@@ -262,6 +262,18 @@ sub adjust_for_ipmi_xen {
     }
 }
 
+# Set boot from pxe via ipmitool
+sub set_pxe_boot {
+    while (1) {
+        my $stdout = ipmitool('chassis bootparam get 5');
+        last if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/;
```
waynechen55, line 269: I think
```
if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/ {
record_info("current boot device is pxe", "Current boot device is $stdout");
last;
}

File: lib/ipmi_backend_utils.pm (hunk 1 above)
Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:266]: set_pxe_efiboot can only be run on a machine installed OS, I guess. while set_pxe_boot control the BIOS directly. It can be run anytime, especially before boot up to install. They are used in different situations.

set_pxe_efiboot just configure BIOS via ipmitool. If the NIC boot order is unable to be configured by BIOS, the functioin can do nothing either, I think.

**Comment by Julie-CAO** [comment]
> After this change, uefi host always boot from local disk directly except for boot_from_pxe module. This means assert pxe menu is only feasible in boot_from_pxe module. Correct ? @Julie-CAO

Only where set_pxe_boot/set_pxe_efiboot is called, the UEFI machines will boot from PXE.

**Comment by waynechen55** [comment]
> > After this change, uefi host always boot from local disk directly except for boot_from_pxe module. This means assert pxe menu is only feasible in boot_from_pxe module. Correct ? @Julie-CAO
> 
> Only where set_pxe_boot/set_pxe_efiboot is called, the UEFI machines will boot from PXE.

Ye

> > After this change, uefi host always boot from local disk directly except for boot_from_pxe module. This means assert pxe menu is only feasible in boot_from_pxe module. Correct ? @Julie-CAO
> 
> Only where set_pxe_boot/set_pxe_efiboot is called, the UEFI machines will boot from PXE.

Yes. This basically means only boot_from_pxe can boot from pxe based on current test flow, but LGTM.

File: lib/ipmi_backend_utils.pm (hunk 1 above)
Comment (b10n1k) [reply lib/ipmi_backend_utils.pm:266]: I wonder whether the tests will work with replacing `boot_from_pxe` with `ipxe_install`. this block is exactly the same as https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/installation/ipxe_install.pm#L47. i am not sure what do you try to do then. If we treat the repo like this the maintenance and the squads collaboration becomes even harder. is it possible for a better approach?

File: lib/ipmi_backend_utils.pm (hunk 3 above)
Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:273]: It is the following 'ipmitool('chassis bootparam get 5');'  in next loop that does the check like your ipmitool(sol info). ipmitool("chassis bootdev pxe ${options}") does take a little time.

File: lib/ipmi_backend_utils.pm (hunk 2 above)
Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:270]: Yes, record_info does help. I did not notice whether the machine boots from pxe when I once run the test with it.

File: lib/ipmi_backend_utils.pm (hunk 1 above)
Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:266]: The test is supposed to work with both boot_from_pxe and ipxe_install with small adaptation

If ipxe_install would be used,  we need a http server which allows HTTP POST(like writer permission) similiar as:
`"IPXE_HTTPSERVER"="http://baremetal-support.qa.suse.de:8080"`

And config it in OSD/Beijing dhcp server the way like:
```
host unarmed-1  { hardware ethernet 18:C0:4D:86:23:A4; fixed-address 10.162.3.85; option host-name "unarmed-1"; filename "kernelqa/ipxe_aarch64.efi"; } 
```

And anything else?

@alice-suse, you may have some idea which one is more suitable to our virtualization tests.

File: lib/ipmi_backend_utils.pm (hunk 3 above)
Comment (waynechen55) [reply lib/ipmi_backend_utils.pm:273]: OK.

File: lib/ipmi_backend_utils.pm (hunk 2 above)
Comment (waynechen55) [reply lib/ipmi_backend_utils.pm:270]: Will you add record_info ?

Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:270]: yes, doing verificatioin run before push

**Comment by guoxuguang** [comment]
> > @Julie-CAO
> > I have one minor comment here, for x86 uefi mode, Do you have any test module to check your SUT is UEFI boot?
> 
> No. only when the UEFI grub menu shows up and is found out by the needle match, the machine will boot up based on the UEFI menu.

IMHO, would better to add `UEFI_BOOT_ONLY=1 `for SUT support UEFI boot only as osd worker config.  Such as vh080/081 here.
just only the needle match that is not enough.

File: lib/ipmi_backend_utils.pm (hunk 1 above)
Comment (Julie-CAO) [reply lib/ipmi_backend_utils.pm:266]: After discussion, we have a plan to move our test machines to use iPXE install, not limited to UEFI machines. We create a ticket to track this, poo#110133.

Hi @b10n1k, where is the http server or who owns it, "http://baremetal-support.qa.suse.de:8080"? Could our virtualization tests use it directly? our machines are in OSD and Beijing:
```
10.162.2.xx  //in OSD, including x86_64 & aarch
10.67.131.xx  //in Beijing VLAN
```

**Comment by Julie-CAO** [comment]
> > > @Julie-CAO
> > > I have one minor comment here, for x86 uefi mode, Do you have any test module to check your SUT is UEFI boot?
> > 
> > 
> > No. only when the UEFI grub menu shows up and is found out by the needle match, the machine will boot up based on the UEFI menu.
> 
> IMHO, would better to add `UEFI_BOOT_ONLY=1 `for SUT support UEFI boot only as osd worker config. Such as vh080/081 here. just only the needle match that is not enough.

In the latest commit, UEFI_PXE_BOOT is used.

**Comment by Julie-CAO** [comment]
All verification runs passed. PR is ready for merge.

File: lib/ipmi_backend_utils.pm (hunk 1 above)
Comment (alice-suse) [reply lib/ipmi_backend_utils.pm:266]: Let me add some info here to avoid confusion. Yes, we have plan to move to ipxe install, but NOT NOW, because the beijing ipxe server is not ready for this yet. And julie has tasks blocked by missing this PR. So let's make the task continue with PR merged firstly, then we will switch to ipxe when all conditions are ready.

File: lib/ipmi_backend_utils.pm [hunk 5]
```diff
@@ -262,6 +262,21 @@ sub adjust_for_ipmi_xen {
     }
 }
 
+# Set boot from pxe via ipmitool
+sub set_pxe_boot {
+    while (1) {
```
alice-suse, line 267: Can this while 1 be replaced with while $timer < $time_out_value? In extreme conditions, eg ipmi nic is plugged off or ipmi is not functioning, this while 1 will take hours to let the job stop after max_job_time.

lib/ipmi_backend_utils.pm
```diff
@@ -262,6 +262,21 @@ sub adjust_for_ipmi_xen {
...
+        my $stdout = ipmitool('chassis bootparam get 5');
+        if ($stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/) {
+            record_info("Current boot device is PXE", "Current boot device is $stdout");
+            last;
+        }
+        diag "setting boot device to pxe";
+        my $options = get_var('IPXE_UEFI') ? 'options=efiboot' : '';
```
alice-suse, line 274: This check is necessary for all UEFI types of booting? Then "options=efiboot"  should be set to options when either IPXE_UEFI or  UEFI_PXE_BOOT is non-empty?

lib/ipmi_backend_utils.pm
Julie-CAO: init_backend before run any test is expected to run from openqa side, so if there was IPMI connection problem, any test module are not supposed to be run.

lib/ipmi_backend_utils.pm
Julie-CAO: No, I don't think so. I don't know what 'options=efiboot' do yet. when boot_from_pxe.pm is called for our uefi machines, 'options=efiboot' is not required. Is it for accessing **i**PXE menu? @b10n1k @mdoucha @czerw may know about it.

lib/ipmi_backend_utils.pm
alice-suse: Yes, I thought about that. It could be possible that the infra problem(ipmi especially network which does not very rarely happen) starts after that. While 1 will be a nightmare with that.

lib/ipmi_backend_utils.pm
alice-suse: Aha, okay. Then current check makes sense.

czerw: @Julie-CAO see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12483  it will not set boot order via ipmi tool without it.

lib/ipmi_backend_utils.pm
czerw: @Julie-CAO  see https://confluence.suse.com/display/qasle/Baremetal+Support+Service it is server managed by kernel team

czerw:
> All verification runs passed. PR is ready for merge.

Could you please point me on verification of our kernel aarch64 uefi installation?

Julie-CAO:
> Could you please point me on verification of our kernel aarch64 uefi installation?

this one, https://openqa.suse.de/tests/8580043,  in my previous comment https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14684#issuecomment-1103376901

but it seems there is still some code change need to be done from my side. I'll re-verify it after final change.

Julie-CAO:
I have to hang up this PR as there is not resources to verify. It will be continued or be replaced with using ipxe_install. I'll get back to you before the pr got merge.

PR #14686, opened 2022-04-08
k3s uses scripts for installation

ilausuch:
Now we are downloading k3s binary and installing that manually
but there are some scripts provided to be used
https://rancher.com/docs/k3s/latest/en/installation/install-options/

- Related ticket: https://progress.opensuse.org/issues/109295
- Verification run: http://copland.arch.suse.de/tests/1408

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -21,11 +21,9 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm);
 
 sub install_k3s {
-    my $k3s_dowload_url = get_required_var("K3S_DOWNLOAD_URL");
-    assert_script_run("curl -LO $k3s_dowload_url");
-    assert_script_run("install -c -m 744 ./k3s /usr/local/bin/k3s");
-    enter_cmd("k3s server 2>&1 | tee k3s.log &");
-    script_retry("grep 'Wrote kubeconfig /etc/rancher/k3s/k3s.yaml' k3s.log", delay => 20, retry => 10);
+    assert_script_run("curl -sfL https://get.k3s.io | sh -");
+    enter_cmd("k3s server &");
+    script_retry("test -e /etc/rancher/k3s/k3s.yaml", delay => 20, retry => 10);
     assert_script_run("k3s kubectl get node");
     script_run("mkdir \$HOME/.kube");
     script_run("rm \$HOME/.kube/config");
@@ -33,10 +31,8 @@ sub install_k3s {
 }
 
 sub uninstall_k3s {
-    my $pid = script_output("ps | grep k3s-server | head -n 1 | cut -d ' ' -f1");
-    script_run("kill $pid");
-    script_run("rm \$HOME/.kube/config");
-    script_run("rm /usr/local/bin/k3s");
+    assert_script_run("rm \$HOME/.kube/config");
+    assert_script_run("/usr/local/bin/k3s-uninstall.sh");
 }
 
 sub install_kubectl {
--- a/tests/containers/helm_k3s.pm
+++ b/tests/containers/helm_k3s.pm
@@ -24,7 +24,6 @@ sub run {
     my ($self) = @_;
 
     $self->select_serial_terminal;
-    $self->{rancher_name} = "rancher-" . get_current_job_id();
     $self->{deployment_name} = "apache-" . get_current_job_id();
     my $chart = "bitnami/apache";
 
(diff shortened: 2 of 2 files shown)


pdostal (approved):

lib/containers/k8s.pm
```diff
@@ -21,22 +21,18 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm);
 
 sub install_k3s {
-    my $k3s_dowload_url = get_required_var("K3S_DOWNLOAD_URL");
-    assert_script_run("curl -LO $k3s_dowload_url");
-    assert_script_run("install -c -m 744 ./k3s /usr/local/bin/k3s");
+    assert_script_run("curl -sfL https://get.k3s.io | sh -");
     enter_cmd("k3s server 2>&1 | tee k3s.log &");
-    script_retry("grep 'Wrote kubeconfig /etc/rancher/k3s/k3s.yaml' k3s.log", delay => 20, retry => 10);
+    script_retry("cat /etc/rancher/k3s/k3s.yaml", delay => 20, retry => 10);
```
b10n1k, line 26: does the `script_retry` still make sense here? it seems to me that `record_info`, `script_output` is better fit. wdyt?

lib/containers/k8s.pm
```diff
@@ -21,22 +21,18 @@ use registration qw(add_suseconnect_product get_addon_fullname);
...
     script_run("rm \$HOME/.kube/config");
     assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml \$HOME/.kube/config");
 }
 
 sub uninstall_k3s {
-    my $pid = script_output("ps | grep k3s-server | head -n 1 | cut -d ' ' -f1");
-    script_run("kill $pid");
     script_run("rm \$HOME/.kube/config");
-    script_run("rm /usr/local/bin/k3s");
+    script_run("/usr/local/bin/k3s-uninstall.sh");
```
b10n1k, line 35: why not `assert_script_run` here?

jlausuch: yes, agree, and just in case, with `timeout => xyz` provided `>90` in case it takes more than 90s to remove stuff.

lib/containers/k8s.pm
ilausuch: the idea is not to print the configuration, is to ensure that the configuration is created. This is replacing the grep because we cannot access to this log (or at least I didn't found) once this config yaml is created the cluster is accessible. It's a barrier to continue with the test

lib/containers/k8s.pm
ilausuch: This is an interesting question, this is inside of a post_fail_hook and based on my  experience we don't assert_script_runs on post_fail_hook. I am going to wait other experienced mates about that

pdostal: I vote for `assert_script_run` as well.

ilausuch:
Thanks for your reviews

lib/containers/k8s.pm
ilausuch: Thanks, now are `assert_script_run`

lib/containers/k8s.pm
ilausuch: I kept this without changes

ilausuch: mmm I am going to check something... my words maybe are not accurate.. checking

jlausuch: ok, I get it now, thanks. But anyway, the `assert_script_run` in line 24 will wait until the command finishes. Question is: when `curl -sfL https://get.k3s.io | sh -` is completed, is there some waiting time until the yaml is created? or is it created by that command at some point?

ilausuch: The `curl...` line only download the binaries and scripts and install them. 
`k3s server` executes the kuberentes, this is the reason has to be executed in background

b10n1k: > the idea is not to print the configuration, is to ensure that the configuration is created. This is replacing the grep because we cannot access to this log (or at least I didn't found) once this config yaml is created the cluster is accessible. It's a barrier to continue with the test

Not sure about the obstacle but it sounds like you could use `test -e` for just check the existence of the file. just an idea. i am not sure what the problem is exactly

ilausuch: mmm yes, good point, thanks

ilausuch: re-testing

ilausuch: test passed

jlausuch (approved):

PR #14688, opened 2022-04-08
wait_boot: Check for PXE menu on machines with PXE boot

mdoucha:
On IPMI machines, GRUB menu does not show up at all and `handle_grub()` returns immediately. This may result in login prompt needle matching too early during reboot, even before the machine actually starts rebooting. The test will then try to activate a console and fail because the system is still booting. Insert an alternative needle check for PXE boot menu where applicable to prevent matching stale login prompt.

Example of failure: http://openqa.qam.suse.cz/tests/40307#step/install_ltp/2

- Related ticket: N/A
- Needles: N/A
- Verification runs:
  - With PXE: http://openqa.qam.suse.cz/tests/40615#step/update_kernel/41
  - Without PXE: https://openqa.suse.de/tests/8503763

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -1147,6 +1147,7 @@ sub wait_boot {
         $self->handle_pxeboot(bootloader_time => $bootloader_time, pxemenu => 'pxe-custom-kernel', pxeselect => 'pxe-custom-kernel-selected');
     }
     else {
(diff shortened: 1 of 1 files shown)


mdoucha:
Both verification runs have passed, this PR is ready for review and merge.

dzedro (approved):

mdoucha:
BTW, this PR is in conflict with PR #14684. If that PR gets merged first, I'll have to add `qa-net-selection-uefi` to the list of needles here.

b10n1k (approved):

b10n1k:
thanks @mdoucha. this will solve some hpc tests as well which had some sporadic failures. So merged

PR #14701, opened 2022-04-11
klp: Remove livepatch git revision check

pevik:
livepatch git revision check was removed from uname output, thus remove the check from test. Check for git-hash/(lp|kGraft) was kept.

Link: bsc#1196281

Verification run:
- http://quasar.suse.cz/tests/8965 (SLES 15.4)
- http://quasar.suse.cz/tests/8961 (SLES 15)
- http://quasar.suse.cz/tests/8964 (SLES 12.5)
NOTE: openSUSE does not use this test

### DIFF
--- a/lib/klp.pm
+++ b/lib/klp.pm
@@ -351,23 +351,30 @@ sub verify_klp_pkg_patch_is_active {
     # livepatch git revision matches the one from the package
     # description.
     my $output = script_output("uname -v");
-    my $uname_v_gitrev;
+    my $uname_pattern = '\([a-z0-9]+/(?:lp|kGraft)\)$';
+    my $uses_gitrev = is_sle('<15-sp4');
+
     chomp($output);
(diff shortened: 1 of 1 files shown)


pevik:
@nicstange can you please review?
@mdoucha @czerw FYI

mdoucha:
The git revision still exists in SLE-15SP3 and earlier. I've asked in the BSC whether that's going to change. If this is just a 15-SP4 thing, then we should make the git revision check conditional instead of removing it completely.

mdoucha:
Please mark this as `WIP:` until we get a reply and use `CASEDIR=...` as a workaround for 15-SP4 jobs.

mdoucha (changes requested):
Got a reply in Bugzilla. The livepatch uname change affects only SLE-15SP4 and later. Please keep the check as is for SLE-15SP3 and earlier.

mdoucha (approved):
LGTM

PR #14704, opened 2022-04-12
Update openladp configration file

coolgw:
Since openladp version udpate trigger test case failed, so we need adapt the configuration file also.

- Related ticket: https://progress.opensuse.org/issues/109422
- Needles: na
- Verification run: https://openqa.opensuse.org/tests/2292456
  NOTE: currently the 389 tools need a fix so you can see the case failed, once the fix is ready the case should pass. Detail can     refer ticket's comments.

### DIFF
--- a/tests/migration/openldap_to_389ds.pm
+++ b/tests/migration/openldap_to_389ds.pm
@@ -52,7 +52,15 @@ sub run {
 
     # We need start openldap to kick out date base file which stored in directory
     assert_script_run "mkdir /tmp/ldap-sssdtest";
+    if (is_tumbleweed) {
+        assert_script_run "sed -i -e '/cachesize/d' ./slapd.conf";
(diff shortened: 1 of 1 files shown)


tests/migration/openldap_to_389ds.pm
```diff
@@ -52,7 +52,17 @@ sub run {
 
     # We need start openldap to kick out date base file which stored in directory
     assert_script_run "mkdir /tmp/ldap-sssdtest";
+    if (is_tumbleweed) {
+        assert_script_run "sed -i -e '/cachesize/d' ./slapd.conf";
+        assert_script_run "sed -i -e 's/hdb/mdb/g' ./slapd.conf";
+        # for openqa debug
+        assert_script_run(qq(echo "PermitRootLogin yes" > /etc/ssh/sshd_config));
```
DimStar77, line 59: you could call `permit_root_ssh` (from utils.pm)

PR #14706, opened 2022-04-12
Unify helm_k3s and helm_eks in one file

ilausuch:
Unify the two files in one to optimize the maintenance, the code is
mostly the same

- Related ticket: https://progress.opensuse.org/issues/109295
- Verification run:
http://fromm.arch.suse.de/tests/6896 (SLE)
https://openqa.opensuse.org/tests/2339620 (opensuse pendent)

NOTE: the variables: PUBLIC_CLOUD_K8S_CLUSTER, PUBLIC_CLOUD_PROVIDER, PUBLIC_CLOUD_SERVICE shouldn't be defined on the test because the test now should be transversal to all the public cloud providers


lib/main_containers.pm
```diff
@@ -131,12 +136,16 @@ sub load_host_tests_containerd_nerdctl {
...
 }
 
-sub load_host_tests_helm() {
+sub load_host_tests_helm {
+    my ($run_args) = @_;
     if (is_sle('15-sp3+')) {
-        loadtest "containers/helm_eks";
-        loadtest "containers/helm_k3s";
+        $run_args->{k8s_env} = "k3s";
+        load_helm_test($run_args);
```
jlausuch, line 143: Maybe we can skip `$run_args` for calling the test module. So, something like this:

```
load_helm_test(); # k3s as default if no arguments
load_helm_test(k8s_env => 'eks');
```

Then, in the `run` method of `helm.pm` you can do something like:
```
sub run {
    my ($self, %args) = @_;
    $args{k8s_env} //= 'k3s'; # use k3s as default
    
    my $is_k3s = $args{k8s_env} == 'k3s';
    my $is_eks = $args{k8s_env} == 'eks';
    ....
```

ilausuch:
Before you ask about this line
`script_run("ln -s /usr/bin/kubectl /usr/local/bin/kubectl");`

It is because something weird happens. Take a look to these lines. 
```
# whereis kubectl; echo rOcq1-$?-
kubectl: /usr/bin/kubectl /usr/share/man/man1/kubectl.1.gz
rOcq1-0-
# kubectl describe deployment/apache-1427; echo QKsU8-$?-
bash: /usr/local/bin/kubectl: No such file or directory
```

pdostal:
I found this in your verification run:
```
# ln -s /usr/bin/kubectl /usr/local/bin/kubectl; echo G5f6k-$?-
ln: failed to create symbolic link '/usr/local/bin/kubectl': File exists
```

ilausuch:
> I found this in your verification run:
> 
> ```
> # ln -s /usr/bin/kubectl /usr/local/bin/kubectl; echo G5f6k-$?-
> ln: failed to create symbolic link '/usr/local/bin/kubectl': File exists
> ```

Yes, this is the reason is not assert script, but could be more elegant, I agree..

lib/main_containers.pm
ilausuch: I applied a mix solution

pdostal:
Why is this `/usr/local/bin/kubectl` sometimes missing in the first place? Isn't it product bug?

ilausuch:
Ok, I found several problems in the main_containers, I am fixing that before to be sure about this question

tests/containers/helm.pm
```diff
@@ -16,21 +16,43 @@
...
+    record_info("Kubernetes environment", $run_args->{k8s_env});
+    record_info("Chart name", $chart);
+
+    install_k3s() if $is_k3s;
+    install_kubectl() if $is_eks;
     install_helm();
 
+    if ($is_eks) {
+        add_suseconnect_product(get_addon_fullname('pcm')) if (is_sle);
+        zypper_call("in -y jq aws-cli");
```
jlausuch, line 45:
```suggestion
        zypper_call("in jq aws-cli");
```
`zypper_call` already includes `-n` option.

tests/containers/helm.pm
```diff
@@ -16,21 +16,43 @@
...
-    install_kubectl();
+    record_info("Kubernetes environment", $run_args->{k8s_env});
+    record_info("Chart name", $chart);
+
+    install_k3s() if $is_k3s;
+    install_kubectl() if $is_eks;
     install_helm();
 
+    if ($is_eks) {
+        add_suseconnect_product(get_addon_fullname('pcm')) if (is_sle);
```
jlausuch, line 44:
```suggestion
        add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
```

ilausuch: done

tests/containers/helm.pm
ilausuch: done

lib/main_containers.pm
```diff
@@ -60,6 +60,14 @@ sub load_container_engine_test {
     loadtest('containers/container_engine', run_args => $run_args, name => $run_args->{runtime});
 }
 
+sub load_helm_test {
+    my ($k8s_env) = @_;
+    my $run_args = OpenQA::Test::RunArgs->new();
```
jlausuch, line 65: do you need to create a new `run_args`? can't we pass the var from `load_host_tests_helm` to this method ?

lib/main_containers.pm
```diff
@@ -131,12 +139,13 @@ sub load_host_tests_containerd_nerdctl {
...
 }
 
-sub load_host_tests_helm() {
+sub load_host_tests_helm {
+    my ($run_args) = @_;
     if (is_sle('15-sp3+')) {
-        loadtest "containers/helm_eks";
-        loadtest "containers/helm_k3s";
+        load_helm_test('eks');
+        load_helm_test('k3s');
```
pdostal, line 146: Please follow the rest of the file and do not redefine the `$run_args`.
```suggestion
        $run_args->{k8s_env} = 'eks';
        load_helm_test($run_args);
        $run_args->{k8s_env} = 'k3s';
        load_helm_test($run_args);
```

tests/containers/helm.pm
```diff
@@ -42,24 +64,26 @@ sub run {
     assert_script_run("helm list");
 
     # Wait for deployment
-    script_retry("kubectl rollout status deployment/$self->{deployment_name}", delay => 30, retry => 5);
+    script_retry("kubectl rollout status deployment/$self->{deployment_name}", delay => 30, retry => 10);
+    assert_script_run("whereis kubectl");
```
pdostal, line 68: What is this for? It's not even the first `kubectl` command.

tests/containers/helm.pm
```diff
@@ -42,24 +64,26 @@ sub run {
...
     assert_script_run("helm delete $self->{deployment_name}");
-    uninstall_k3s();
+    uninstall_k3s() if $is_k3s;
 }
 
 sub post_fail_hook {
     my ($self) = @_;
     script_run("helm delete $self->{deployment_name}");
-    uninstall_k3s();
+    uninstall_k3s() if $self->{is_k3s};
```
mloviska, line 86: Why we need to remove `k3s` only? Sorry, if the question has been already answered. It just caught my attention.

tests/containers/helm.pm
```diff
@@ -16,21 +16,43 @@
...
     $self->{deployment_name} = "apache-" . get_current_job_id();
     my $chart = "bitnami/apache";
 
-    install_k3s();
-    install_kubectl();
+    record_info("Kubernetes environment", $run_args->{k8s_env});
+    record_info("Chart name", $chart);
+
+    install_k3s() if $is_k3s;
+    install_kubectl() if $is_eks;
```
mloviska, line 40: can this be moved below under L#43? It would be nice to have as much as `eks` stuff under this condition

ilausuch:
After a meeting I am re-writting this test with a new approach, current code is under construction but seems that the tests work well http://fromm.arch.suse.de/tests/6686, https://openqa.opensuse.org/tests/2326418

tests/containers/helm.pm
ilausuch: we uninstall k3s when k3s is installed previously. This step is not necessary but clean up a little the VM

tests/containers/helm.pm
ilausuch: This has changed

tests/containers/helm.pm
ilausuch: Developing stuff, removed

lib/main_containers.pm
ilausuch: New approach

lib/main_containers.pm
ilausuch: New approach

tests/containers/helm.pm
```diff
@@ -16,19 +16,58 @@
...
+use version_utils qw(is_sle);
 use mmapi 'get_current_job_id';
+use registration qw(add_suseconnect_product get_addon_fullname);
 use containers::k8s;
 
 sub run {
-    my ($self) = @_;
+    my ($self, $run_args) = @_;
+
+    my $provider_name = get_var('PUBLIC_CLOUD_SERVICE', 'k3s');
```
grisu48, line 28:
```suggestion
    my $provider_name = get_required_var('PUBLIC_CLOUD_SERVICE');
```

I would make the choice of the kubernetes flavor here mandatory to avoid confusion or is this too verbose?

ilausuch: K3s is not a public cloud service, the idea is that if you don't specify this then we don't use a public cloud kubernetes to test the helm. This is the reason to not use required

pdostal: I'd add a comment before this line. Maybe something like:
```
# We either test various cloud clusters or local k3s
```

pdostal (approved):
Awesome 🌞

ilausuch:
I squashed all commits in only one

tests/containers/helm.pm
```diff
@@ -26,8 +26,9 @@ sub run {
     my ($self, $run_args) = @_;
 
     # We either test various cloud clusters or local k3s
-    my $k8s_backend = get_required_var('HELM_K8S_BACKEND_2');
+    my $k8s_backend = $run_args->{backends}->[0];
```
mloviska, line 29: I was thinking to do the `shift` over here :) Instead of `main_containers` as you need to preserve the backends until test runtime

ilausuch: Works, thanks!

tests/containers/helm.pm
ilausuch: done, thanks!

pdostal (approved):
You may probably run the `tools/tidy` to make the CI happy. Good job 🚀

tests/containers/helm.pm
```diff
@@ -0,0 +1,123 @@
...
+        my $provider;
+
+        # Configure the public cloud kubernetes
+        if ($k8s_backend eq "EKS") {
+            add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
+            zypper_call("in jq aws-cli");
+
+            $provider = publiccloud::eks->new(
+                key_id => get_var('PUBLIC_CLOUD_KEY_ID'),
+                key_secret => get_var('PUBLIC_CLOUD_KEY_SECRET'),
```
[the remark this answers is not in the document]
asmorodskyi: this is not used anymore after my PR was merged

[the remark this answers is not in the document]
ilausuch: I have to fix that, this is the old fashion way, so.. it's incorrect. Thanks to point on that!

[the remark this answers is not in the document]
ilausuch: Fixed

.tidyallrc
```diff
@@ -1,3 +1,3 @@
 [PerlTidy]
 select = {lib,products,script,tests,t}/**/*.{pl,pm,t}
-argv = --pro=$ROOT/.perltidyrc
+argv = --pro=$ROOT/.perltidyrc -g
```
jlausuch, line 3: I guess you can't change this as it will affect everyone.

lib/main_containers.pm
```diff
@@ -131,15 +136,21 @@ sub load_host_tests_containerd_nerdctl {
...
-        loadtest "containers/helm_k3s";
+        $backends = get_var("HELM_K8S_BACKEND","K3S");
+    } else {
+        die("Helm backend not supported on this host");
     }
-    else {
-        die("Helm test not supported on this host");
+
+    foreach (split(',\s*', $backends)) {
+        push @{$run_args->{backends}}, $_;
```
jlausuch, line 152: So, for my own understanding, you are creating a new array `@{$run_args->{backends}}` and then pushing the backends in this array? So, for SLE case the array would be
`@{$run_args->{backends}} = ('EKS' ,'AKS', 'K3S')` ?

But how is this handled by the test later on? 
`my $k8s_backend = shift(@{$run_args->{backends}});`

.tidyallrc
ilausuch: yes, only was for testing purposes now

lib/main_containers.pm
```diff
@@ -131,15 +136,21 @@ sub load_host_tests_containerd_nerdctl {
...
+    my ($run_args) = @_;
+    my $backends = undef;
+
     if (is_sle('15-sp3+')) {
-        loadtest "containers/helm_eks";
-        loadtest "containers/helm_k3s";
+        $backends = get_var("HELM_K8S_BACKEND","EKS,AKS,K3S");
     } elsif (is_opensuse) {
-        loadtest "containers/helm_k3s";
+        $backends = get_var("HELM_K8S_BACKEND","K3S");
```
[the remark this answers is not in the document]
ilausuch: should be this the problem

jlausuch (approved):

PR #14707, opened 2022-04-13
Fix check_nodejs_common not works on MEDIA_UPGRADE migration

hjluo:
We need to exclude the check_nodejs_common check on MEDIA_UPGRADE
migration cases. They're all disabled at that stage.

- Related ticket: https://progress.opensuse.org/issues/109680 
- Needles: N/A
- Verification run:
  media: 
  https://openqa.nue.suse.com/tests/8523389  
  https://openqa.nue.suse.com/tests/8527158  
 
  normal:
  https://openqa.nue.suse.com/t8528233

  installed:
  http://openqa.nue.suse.com/tests/8528530#step/check_system_info#2/31


tinawang123 (approved):
LGTM

chcao (approved):
LGTM

tests/console/check_system_info.pm
```diff
@@ -91,11 +91,13 @@ sub check_nodejs_common {
     record_info('SLE-21783', 'check nodejs-common');
     zypper_call('in nodejs-common | tee /tmp/install_nodejs.log');
 
-    assert_script_run('grep -E "Installing: nodejs16" /tmp/install_nodejs.log');
-    if ((script_output(q(zypper se -i nodejs | sed -n '/nodejs16 /p' | awk '{print $1}')) !~ /i/) || \
-        (script_output(q(zypper se -i nodejs | sed -n '/nodejs-common/p' | awk '{print $1}')) !~ /i\+/)) {
-        die "Expected package is not installed";
+    my $ret = script_run(q(grep "'nodejs-common' is already installed." /tmp/install_nodejs.log));
+    if ($ret != 0) {
+        assert_script_run('grep -E "Installing: nodejs16" /tmp/install_nodejs.log');
```
lemon-suse, line 96: If nodejs-common hasn't installed, on what condition that the nodejs16 will be in the install log?

tests/console/check_system_info.pm
```diff
@@ -91,11 +91,13 @@ sub check_nodejs_common {
...
     }
+    my $ret_s1 = script_output(q(zypper se -i nodejs | sed -n '/nodejs16 /p' | awk '{print $1}'));
+    my $ret_s2 = script_output(q(zypper se -i nodejs | sed -n '/nodejs-common/p' | awk '{print $1}'));
+    die "Expected package is not installed" if ('i' ne "$ret_s1" || 'i+' ne "$ret_s2");
```
lemon-suse, line 100: It is hard to say which is better to compare with the old code.

tests/console/check_system_info.pm
hjluo: https://openqa.nue.suse.com/tests/8538473#step/check_system_info/29
there's real install action

tests/console/check_system_info.pm
hjluo: and the old code can't work
https://openqa.nue.suse.com/tests/8502837#step/check_system_info/39

tests/console/check_system_info.pm
lemon-suse: I think only assert_script_run('grep -E "Installing: nodejs16" /tmp/install_nodejs.log'); is enough.
Since we just check the dependency relationship with nodejs-common, just after cmd to install nodejs-common. WDYT?

hjluo: Do you mean to remove the check for "i" and "i+"?  if so I'm Ok with it.

hjluo: changed the code as we discussed and now run VR again.

lemon-suse (approved):

PR #14712, opened 2022-04-13
Kick off for BCI test repo

mloviska:
- motivation: [Create a new module for BCI Repository testing](https://progress.opensuse.org/issues/109827)

Use `podman` in order to pull a BCI specified container (`CONTAINER_IMAGE_TO_TEST`).
Check whether it has default added repositories, list available
patterns, check orphans and
then replace the default repository with `BCI_TEST_REPO` and run the
same tests.

- Verification run: [sle-15-SP3-BCI-Updates-x86_64-Build12.4_golang](http://kepler.suse.cz/tests/16463#)

### DIFF
--- a/tests/containers/bci_repo.pm
+++ b/tests/containers/bci_repo.pm
@@ -1,0 +1,77 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: bci-repo test
+#   SUSE Linux Enterprise Base Container Images (SLE BCI)
+#   provides truly open, flexible and secure container images and application
+#   development tools for immediate use by developers and integrators without
+#   the lock-in imposed by alternative offerings.
+#
+#   This module is used to test BCI repository.
+#   We use podman as container engine of reference.
+#   Podman is assumed to be installed on the host
+#   (e.g. run containers/host_configuration.pm before this test case).
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use Mojo::Base qw(consoletest);
+use testapi;
+use utils;
+
+my $patterns = [
+    [qw(Amazon_Web_Services Amazon_Web_Services_Instance_Init Amazon_Web_Services_Instance_Tools Amazon_Web_Services_Tools)],
+    [qw(Google_Cloud_Platform Google_Cloud_Platform_Instance_Init Google_Cloud_Platform_Instance_Tools Google_Cloud_Platform_Tools)],
+    [qw(Microsoft_Azure Microsoft_Azure_Instance_Init Microsoft_Azure_Instance_Tools Microsoft_Azure_Tools)],
+    [qw(OpenStack OpenStack_Instance_Init OpenStack_Instance_Tools OpenStack_Tools)],
+    [qw(apparmor base devel_basis documentation enhanced_base fips ofed sw_management)]
+];
+
+sub container_exec {
+    my $container = shift;
+    my $cmd = shift;
+    assert_script_run(qq[podman exec $container /bin/sh -c '$cmd'], @_);
+}
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    my $image = get_required_var('CONTAINER_IMAGE_TO_TEST');
+    my $bci_repo = get_required_var('REPO_BCI');
+
+    # download and start a BCI container
+    record_info('IMAGE', $image);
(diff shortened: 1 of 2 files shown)


jlausuch:
Nice!
I was thinking that maybe instead of using `my $image = get_required_var('CONTAINER_IMAGE_TO_TEST');` we could let the test run these checks in several images, not single one. 
This way, we could do a loop for the images where we want to test the repo (e.g. micro, minimal, etc).

tests/containers/bci_repo.pm
```diff
@@ -0,0 +1,63 @@
...
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: bci-repo test
+#   SUSE Linux Enterprise Base Container Images (SLE BCI)
+#   provides truly open, flexible and secure container images and application
+#   development tools for immediate use by developers and integrators without
+#   the lock-in imposed by alternative offerings.
+#
+#   This module is used to test BCI repository.
```
jlausuch, line 12: Could you add some more info, like:
# We use podman as container engine of reference.
# Podman is assumed to be installed on the host (we can discuss this btw).
# ...

mloviska:
> Nice! I was thinking that maybe instead of using `my $image = get_required_var('CONTAINER_IMAGE_TO_TEST');` we could let the test run these checks in several images, not single one. This way, we could do a loop for the images where we want to test the repo (e.g. micro, minimal, etc).

I was thinking about that as well, but we will also increase a chance of failure while pulling an image as we cannot rely on our network performance.

tests/containers/bci_repo.pm
```diff
@@ -0,0 +1,63 @@
...
+    container_exec($container, "zypper -q -s 11 pa --orphaned | tee -a repo.org", timeout => 600);
+    # remove container-suseconnect, in order to install the Cloud patterns
+    container_exec($container, 'zypper -n rm container-suseconnect', timeout => 180);
+    # remove SLE_BCI repo pointing to official update servers and add SUT repo
+    container_exec($container, 'zypper rr 1');
+    container_exec($container, "zypper ar $bci_repo BCI_TEST");
+    container_exec($container, 'zypper ref', timeout => 180);
+    container_exec($container, 'zypper lr -d');
+    container_exec($container, 'zypper search -t pattern', timeout => 180);
+    container_exec($container, "zypper -n in -t pattern $patterns", timeout => 600);
```
jlausuch, line 53: This caused some issues to me, I couldn't install all patterns at once... There were conflicting packages for the different public cloud provider patterns, so I needed to install them separately. Maybe a loop for each pattern and installing it in fresh container?

tests/containers/bci_repo.pm
mloviska: oh, that is a big comment that I cannot miss :D

tests/containers/bci_repo.pm
mloviska: It should install patterns according to specified variable as it is mentioned in the description of the ticked. Thus in this case it has pulled the group with `Google_Cloud_Platform,Google_Cloud_Platform_Instance_Init,Google_Cloud_Platform_Instance_Tools,Google_Cloud_Platform_Tools`.

jlausuch:
> > Nice! I was thinking that maybe instead of using `my $image = get_required_var('CONTAINER_IMAGE_TO_TEST');` we could let the test run these checks in several images, not single one. This way, we could do a loop for the images where we want to test the repo (e.g. micro, minimal, etc).
> 
> I was thinking about that as well, but we will also increase a chance of failure while pulling an image as we cannot rely on our network performance.

`script_retry` :D

mloviska:
> > > Nice! I was thinking that maybe instead of using `my $image = get_required_var('CONTAINER_IMAGE_TO_TEST');` we could let the test run these checks in several images, not single one. This way, we could do a loop for the images where we want to test the repo (e.g. micro, minimal, etc).
> > 
> > 
> > I was thinking about that as well, but we will also increase a chance of failure while pulling an image as we cannot rely on our network performance.
> 
> `script_retry` :D

It is already there :D

tests/containers/bci_repo.pm
jlausuch: Yes, and I probably forgot to update the ticket with the outcome of my manual testing... 
btw, the zypper command should return failure if it fails, in the [VR](http://kepler.suse.cz/tests/16450#step/bci_repo/55) I see
`System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down`

jlausuch (approved):
LGTM

PR #14713, opened 2022-04-13
journal_check: Failed to start Load kdump kernel and initrd

DimStar77:
The error appears sometimes, but not always
https://bugzilla.opensuse.org/show_bug.cgi?id=1196335

(as there is no reaction, the bug seems to not be worthy to be flagged
in stagings over and over)

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1196335
- Needles: N/A
- Verification run: Not Done

### DIFF
--- a/data/journal_check/bug_refs.json
+++ b/data/journal_check/bug_refs.json
@@ -334,5 +334,12 @@
             "sle": ["15-SP2", "15-SP3", "15-SP4"]
         },
         "type": "ignore"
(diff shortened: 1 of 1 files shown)


foursixnine:
@mloviska are you aware of this bug?

mloviska:
> @mloviska are you aware of this bug?

Now yes, thanks! :)

PR #14715, opened 2022-04-14
Enhance gpg testcase.

wangqianzou:
Kill gpg-agent service when executing gpg2 comand in case gpg-agent
does NOT see your environment variable: LIBGCRYPT_FORCE_FIPS_MODE=1.

- Related ticket: https://progress.opensuse.org/issues/109596
- Needles: N/A
- Verification run:https://openqa.suse.de/tests/8559539#step/gpg/21

### DIFF
--- a/tests/console/gpg.pm
+++ b/tests/console/gpg.pm
@@ -57,6 +57,18 @@ EOF
         );
         assert_script_run("cat $egg_file");
 
+        # The gpg binary does not call libgcrypt directly to generate keys.
+        # gpg instead calls into a running "gpg-agent"
+        # which is launched during GNOME session start to let it create the key.
+        # This gpg-agent does NOT see your environment variable,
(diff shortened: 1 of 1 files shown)


rfan1:
@wangqianzou Can you please provide both kernel and env mode test results here?

jouyingbin:
@wangqianzou 
1.  Commit message title can be more clear if possible
2. There is no need to add "." in the end
3. typo: comand
4. will you also provide the kernel mode VR 
5. If there is a regression in kernel mode, perhaps you can add if **ENV** condition while running "gpgconf --kill gpg-agent"

wangqianzou:
> * the kernel mode VR
> 
> * If there is a regression in kernel mode, perhaps you can add if **ENV** condition while running "gpgconf --kill gpg-agent"

PR #14718, opened 2022-04-14
Bump timeout on select_patterns waiting change

JRivrain:
On slower machines 10 seconds is sometimes not enough to probe for screen change in select_patterns. Trying with 20.

- Related ticket: https://progress.opensuse.org/issues/108977
VR: https://openqa.suse.de/tests/8554059#next_previous

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -175,7 +175,9 @@ Being in the "select pattern" screen, performs steps needed to move the
 highlight cursor one item down.
 =cut
 sub move_down {
-    my $ret = wait_screen_change { send_key 'down' };
+    my $ret = wait_screen_change(sub {
+            send_key 'down';
+    }, 20);
     workaround_bsc1189550() if (!$workaround_bsc1189550_done && is_sle('>=15-sp3'));
     last if (!$ret);    # down didn't change the screen, so exit here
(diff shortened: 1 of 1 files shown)


JRivrain:
> Great PR! Please pay attention to the following items before merging:
> 
> **Files matching `lib/**.pm`:**
> 
>     * [ ]  Consider adding or extending unit tests in t/
> 
> 
> This is an automatically generated QA checklist based on modified files

What is this ?

jknphy:
When I tried during the time where there were networks issues, it didn't help much (sorry I don't have the VR that I did), I mentioned here: https://progress.opensuse.org/issues/108977#note-3
Perhaps we can just leave it as it is. wdyt?

JRivrain:
> When I tried during the time where there were networks issues, it didn't help much (sorry I don't have the VR that I did), I mentioned here: https://progress.opensuse.org/issues/108977#note-3 Perhaps we can just leave it as it is. wdyt?

It always fails in hyperv, (when it ever reaches that point...) let's see if it helps https://openqa.suse.de/tests/8580980#next_previous, but yes, obviously hyperv workers are weak anyway.

JRivrain:
Ok, does not help https://openqa.suse.de/tests/8580984#step/select_patterns/18

PR #14723, opened 2022-04-18
Need not load check_system_info in migration regression test

lemon-suse:
We already have the check_system_info after post_migration, so we don't need load the check_system_info in regression test any more.

- Related ticket: https://progress.opensuse.org/issues/109996
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/8559441#
  daily jobs to avoid regression:
  https://openqa.nue.suse.com/tests/8559477
  https://openqa.nue.suse.com/tests/8559478

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1192,7 +1192,6 @@ sub load_consoletests {
         loadtest "console/x_vt";
     }
     loadtest "console/zypper_lr";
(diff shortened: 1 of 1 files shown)


hjluo:
LGTM

coolgw (approved):

PR #14726, opened 2022-04-19
Remove unnecessary scripts in the spack setup

b10n1k:
Remove `source /etc/profile.d/lmod.sh` which based on the feedbacks
it should not be sourced after a relogin. If do, something would be wrong.
Also `/usr/lib/spack/run-find-external.sh` should be run from the spack
installation. So it doesnt need to repeat it.


b10n1k:
https://openqa.suse.de/tests/8574241#
https://openqa.suse.de/tests/8574241#

PR #14732, opened 2022-04-19
Refactor LTP whitelist module into a class

mdoucha:
Reloading the LTP known issues file multiple times is somewhat expensive and causes performance issues during LTP job initialization because every test is checked against skip entries in the file. Refactor the whitelist module into a class so that the whitelist data can be queried multiple times after being loaded once.

- Related ticket: N/A
- Needles: N/A
- Verification runs:
  - ltp_kernel_misc https://openqa.suse.de/tests/8576977 (`block_dev` and `tpci` tests will not be scheduled)
  - ltp_net_rpc_tests https://openqa.suse.de/tests/8576978
  - ltp_net_rpc_tests without `LTP_KNOWN_ISSUES` https://openqa.suse.de/tests/8576979
  - ltp_syscalls https://openqa.suse.de/tests/8576980 (compare boot_ltp run time with original job: https://openqa.suse.de/tests/8554772)

### DIFF
--- a/lib/LTP/WhiteList.pm
+++ b/lib/LTP/WhiteList.pm
@@ -6,7 +6,6 @@
 
 package LTP::WhiteList;
 
-use base Exporter;
 use strict;
 use warnings;
 use testapi;
@@ -19,23 +18,28 @@ use Mojo::UserAgent;
 use Mojo::File 'path';
 use YAML::PP;
 
-our @EXPORT = qw(
-  download_whitelist
-  find_whitelist_testsuite
-  find_whitelist_entry
-  override_known_failures
-  is_test_disabled
-  list_skipped_tests
-);
+sub new {
+    my $class = shift;
+    my $self = bless({}, $class);
+    my $path = get_var('LTP_KNOWN_ISSUES_LOCAL');
 
-sub download_whitelist {
+    if (!defined($path) && get_var('LTP_KNOWN_ISSUES')) {
+        $path = _download_whitelist();
+    }
+
+    $self->{whitelist} = _load_whitelist_file($path) if $path;
+    $self->{whitelist} ||= {};
+    return $self;
+}
+
+sub _download_whitelist {
     my $path = get_var('LTP_KNOWN_ISSUES');
     return undef unless defined($path);
 
     my $res = Mojo::UserAgent->new->get($path)->result;
     unless ($res->is_success) {
         record_info("File not downloaded!", $res->message, result => 'softfail');
-        set_var('LTP_KNOWN_ISSUES_LOCAL', undef);
+        set_var('LTP_KNOWN_ISSUES_LOCAL', '');
         return;
     }
 
@@ -47,12 +51,13 @@ sub download_whitelist {
     copy($lfile, "ulogs/$basename");
 
     set_var('LTP_KNOWN_ISSUES_LOCAL', $lfile);
+    return $lfile;
 }
 
 sub find_whitelist_entry {
-    my ($env, $suite, $test) = @_;
+    my ($self, $env, $suite, $test) = @_;
 
-    $suite = find_whitelist_testsuite($suite);
+    $suite = $self->{whitelist}->{$suite};
     return undef unless ($suite);
 
     my @issues;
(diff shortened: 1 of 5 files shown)


tests/publiccloud/run_ltp.pm
```diff
@@ -61,13 +61,12 @@ sub upload_ltp_logs
 
         my $ltp_log = Mojo::JSON::decode_json($log_file->slurp());
         my $parser = OpenQA::Parser::Format::LTP->new()->load($log_file->to_string);
+        my %ltp_log_results = map { $_->{test_fqn} => $_->{test} } @{$ltp_log->{results}};
+        my $whitelist = LTP::WhiteList->new();
 
-        if (find_whitelist_testsuite($ltp_testsuite)) {
-            my %ltp_log_results = map { $_->{test_fqn} => $_->{test} } @{$ltp_log->{results}};
-            for my $result (@{$parser->results()}) {
-                if (override_known_failures($self, {%{$self->{ltp_env}}, retval => $ltp_log_results{$result->{test_fqn}}->{retval}}, $ltp_testsuite, $result->{test_fqn})) {
```
mdoucha, line 68: BTW, this part is not checking whether the test actually failed before applying whitelist. It may end up turning `pass` into `softfail`. It should be fixed in a separate PR.

asmorodskyi:
http://autobot.qa.suse.de/tests/3116
http://autobot.qa.suse.de/tests/3117
http://autobot.qa.suse.de/tests/3118

this is Public Cloud SLE 15 SP4 . SLE 12 SP5 will come later ( solving issues unrelated to this PR... )

lib/LTP/WhiteList.pm
```diff
@@ -19,23 +18,28 @@ use Mojo::UserAgent;
...
+
+sub _download_whitelist {
     my $path = get_var('LTP_KNOWN_ISSUES');
     return undef unless defined($path);
 
     my $res = Mojo::UserAgent->new->get($path)->result;
     unless ($res->is_success) {
         record_info("File not downloaded!", $res->message, result => 'softfail');
-        set_var('LTP_KNOWN_ISSUES_LOCAL', undef);
+        set_var('LTP_KNOWN_ISSUES_LOCAL', '');
```
b10n1k, line 42: i dont think you need this statement anymore. Where is it used?

lib/LTP/WhiteList.pm
```diff
@@ -19,23 +18,28 @@ use Mojo::UserAgent;
...
+
+    if (!defined($path) && get_var('LTP_KNOWN_ISSUES')) {
+        $path = _download_whitelist();
+    }
 
-sub download_whitelist {
+    $self->{whitelist} = _load_whitelist_file($path) if $path;
+    $self->{whitelist} ||= {};
+    return $self;
+}
```
b10n1k, line 33: i was thinking that with Mojo::Base, you could avoid some boilerplate. WDYT?

lib/LTP/WhiteList.pm
```diff
@@ -19,23 +18,28 @@ use Mojo::UserAgent;
...
-  find_whitelist_testsuite
-  find_whitelist_entry
-  override_known_failures
-  is_test_disabled
-  list_skipped_tests
-);
+sub new {
+    my $class = shift;
+    my $self = bless({}, $class);
+    my $path = get_var('LTP_KNOWN_ISSUES_LOCAL');
```
asmorodskyi, line 24: I wonder what the point to create this class when this variable is not defined ? Shouldn't we simply fail here ?

asmorodskyi:
http://autobot.qa.suse.de/tests/3123
http://autobot.qa.suse.de/tests/3122
http://autobot.qa.suse.de/tests/3121

PubCloud SLE 12 SP5 GCE/EC2/Azure LTP syscalls

lib/LTP/WhiteList.pm
mdoucha: It's used in `LTP::WhiteList::new()`:
- `undef` => download the whitelist file
- empty string => do nothing
- non-empty value => just load the already downloaded file

lib/LTP/WhiteList.pm
mdoucha: Can you be more specific?

lib/LTP/WhiteList.pm
mdoucha: No, we should not fail here. Running LTP tests without any whitelist file is perfectly valid.

mdoucha: Also, if `LTP_KNOWN_ISSUES_LOCAL` is not defined, it just means that this is the first ever class instance in the job and the file needs to be downloaded.

pevik (approved):
LGTM, thanks a lot for debugging and fixing the issue.

czerw (approved):

asmorodskyi (approved):
LGTM

lib/LTP/WhiteList.pm
mdoucha: After clarifying this a bit on IRC, I'm not going to use any of those special features. It'd add nothing but unnecessary complexity.

PR #14733, opened 2022-04-19
Perform different checks if HANA volumes are mounted via NFS

alvarocarvajald:
The `sles4sap/hana_test` test module currently expects the HANA installation to have been performed on local disks by either `sles4sap/hana_install` or `sles4sap/wizard_hana_install`. This limits the scenarios on which the module can be used to test a HANA installation.

This PR adds code to `sles4sap/hana_test` so it is able to determine when the HANA volumes are mounted over NFS, and perform checks different than the LVM commands on the `vg_hana` volume group as it currently does.

Additionally, a new **SKIP_HANADB_QUERY** setting is introduced to skip the **hdbsql** checks, for those scenarios where the HANA _SYSTEM_ password is not available.

- Related ticket: N/A
- Needles: N/A
- Verification run: http://mango.qa.suse.de/tests/4811

### DIFF
--- a/tests/sles4sap/hana_test.pm
+++ b/tests/sles4sap/hana_test.pm
@@ -53,10 +53,18 @@ sub run {
     # NOTE: done here because AutoYaST part is not HANA specific
     $self->upload_hana_install_log if get_var('AUTOYAST');
 
-    # Check the memory/disk configuration
+    # Check the memory/disk configuration. Determine first whether HANA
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM

mpagot (approved):

lpalovsky:
LGTM

lpalovsky (approved):

PR #14735, opened 2022-04-20
Add auxiliary functions for pc instance

grisu48:
Introduce some auxiliary functions for executing commands on a
publiccloud instance. The names follow the default testapi conventions
to remove some confusion with the almighty 'run_ssh_command' subroutine.

- Related ticket: https://progress.opensuse.org/issues/110119
- Verification run: [GCE](https://duck-norris.qam.suse.de/tests/10523)

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -16,6 +16,7 @@ use File::Basename;
 use publiccloud::utils;
 use publiccloud::ssh_interactive qw(ssh_interactive_tunnel ssh_interactive_leave);
 use version_utils;
+use utils;
 
 use constant SSH_TIMEOUT => 90;
 
@@ -103,6 +104,8 @@ With C<<ssh_opts=>'...'>> you can overwrite all default ops which are in
 C<<$instance->ssh_opts>>.
 Use argument C<username> to specify a different username then
 C<<$instance->username()>>.
+
+This function is deprecated. Please use ssh_script_retry instead.
 =cut
 
 sub retry_ssh_command {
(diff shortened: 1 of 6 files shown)


grisu48:
Before further processing I'm asking for feedback if this is a good idea and if we should proceed to use those more explanatory subroutines in our test runs.

lib/publiccloud/instance.pm
```diff
@@ -119,6 +122,90 @@ sub retry_ssh_command {
...
+
+Runs a command C<cmd> via ssh on the publiccloud instance and returns the return code.
+=cut
+sub ssh_script_run {
+    my $self = shift;
+    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @_);
+    my $ssh_cmd = $self->_prepare_ssh_cmd(%args);
+    delete($args{cmd});
+    delete($args{ssh_opts});
+    delete($args{username});
```
pdostal, line 156: What about `$self->instance->username`?

grisu48: I guess the purpose was to being able to pass an arbitrary username (e.g. `root`) but I'm not aware of it being anywhere used.
In the end it's best to stay close to the original `run_ssh_command` implementation, that's why this is still there.

lib/publiccloud/instance.pm
```diff
@@ -101,6 +102,8 @@ With C<<ssh_opts=>'...'>> you can overwrite all default ops which are in
 C<<$instance->ssh_opts>>.
 Use argument C<username> to specify a different username then
 C<<$instance->username()>>.
+
+This function is deprecated. Please use ssh_script_retry instead.
```
asmorodskyi, line 106: I would also add `record_info` into function call with same message

lib/publiccloud/instance.pm
```diff
@@ -119,6 +122,90 @@ sub retry_ssh_command {
...
+    my ($self, %args) = @_;
+    die('No command defined') unless ($args{cmd});
+    $args{ssh_opts} //= $self->ssh_opts() . " -i '" . $self->ssh_key . "'";
+    $args{username} //= $self->username();
+    $args{timeout} //= SSH_TIMEOUT;
+
+    my $cmd = $args{cmd};
+    unless ($args{no_quote}) {
+        $cmd =~ s/'/\'/g;    # Espace ' character
+        $cmd = "\$'$cmd'";
```
asmorodskyi, line 136: sorry but my perl/bash knowledge not allowing me to understand purpose of this line , can you elaborate ?

lib/publiccloud/instance.pm
asmorodskyi: it is very useful flow which I have used several times . For example SLEM pubcloud images has hardcoded user `suse` and current master allowing simple fire test with  `PUBLIC_CLOUD_USER` to perfectly handle this

asmorodskyi:
nice ! I like how it is done !

lib/publiccloud/instance.pm
grisu48: You mean to add a "warning" to the test run, when a deprecated function is used?

Yes makes sense, but I would do this as a second step, not within this PR. I would kickstart it here, and then replace them step by step to limit the possible fallout.

asmorodskyi: yeah sure this will work to , but I actually don't any harm from adding this info box already with this PR .

grisu48:
Currently blocked as stated in https://progress.opensuse.org/issues/110119

asmorodskyi:
this one was just forgotten , we may resume work on it

grisu48:
Reviving PR. I would not merge it before next week though so that we can immediately react to potential issues with it.

pdostal (approved):
This looks clear to me :+1:

asmorodskyi (approved):
code = LGTM , but I have concerns regarding VR : 
1. It is 4 months old during this period of time we have merged a lot of code so I would re-run fresh VR 
2. For some reason I see only GCE VR does it make sense ?

grisu48:
> code = LGTM , but I have concerns regarding VR :
> 
>     1. It is 4 months old during this period of time we have merged a lot of code so I would re-run fresh VR
> 
>     2. For some reason I see only GCE VR does it make sense ?

1. Nothing changed in the code since the last VR, so I didn't saw the need for running a new one - But here it is: https://duck-norris.qam.suse.de/tests/10523#live
2. I'm only changing generic methods which apply to all CSP providers, so IMHO testing one of them should be enough. On request I can schedule some more.

lib/publiccloud/instance.pm
```diff
@@ -122,6 +125,94 @@ sub retry_ssh_command {
...
+    }
+
+    my $ssh_cmd = sprintf('ssh %s "%s@%s" -- %s', $args{ssh_opts}, $args{username}, $self->public_ip, $cmd);
+    #$ssh_cmd = "timeout $args{timeout} $ssh_cmd" if ($args{timeout} > 0);
```
jlausuch, line 143: leftover?

PR #14730, opened 2022-04-19
Add terraform variables

grisu48:
Add missing terraform variables.

PR #14736, opened 2022-04-20
Add azure-nfs test

grisu48:
Adds the new azure_nfs test for testing the internal Azure NFS file share.
This test requires also adding a new terraform profile that includes the
creation of an Azure storage account.

- Related ticket: https://progress.opensuse.org/issues/101189
- Verification run: [Azure-NFS](https://duck-norris.qam.suse.de/tests/8807) | [GCE](http://duck-norris.qam.suse.de/tests/8808) (ensure no side effects are there)

### DIFF
--- a/data/publiccloud/terraform/azure_nfstest.tf
+++ b/data/publiccloud/terraform/azure_nfstest.tf
@@ -1,0 +1,258 @@
+terraform {
+  required_providers {
+    azurerm = {
+      version = ">= 3.2.0"
+      source = "hashicorp/azurerm"
+    }
+    random = {
+      version = "= 3.1.0"
+      source = "hashicorp/random"
+    }
+  }
+}
+
+provider "azurerm" {
+    features {}
+}
+
+## ---- variables ----------------------------------------------------------- ##
+
+## general variables
+
+variable "name" {
+    default = "openqa-vm"
+}
+
+variable "tags" {
+    type = map(string)
+    default = {}
+}
+
+variable "region" {
+    default = "westeurope"
+}
+
+## vm-instance variables
+
+variable "instance_count" {
+    default = "1"
+}
+variable "type" {
+    default = "Standard_A2_v2"
+}
+
+variable "image_id" {
+    default = ""
+}
+
+variable "offer" {
+    default=""
+}
+
+variable "sku" {
+    default="gen1"
+}
+
+## ---- data ---------------------------------------------------------------- ##
+
+// IP address of the client
+data "http" "myip" {
+  url = "http://ipv4.icanhazip.com"
+}
+
+## -------------------------------------------------------------------------- ##
+
+resource "random_id" "service" {
+    keepers = {
+        name = var.name
+    }
+    byte_length = 8
+}
+
+## resource group
+
+resource "azurerm_resource_group" "openqa-group" {
+    name     = "${var.name}-${element(random_id.service.*.hex, 0)}"
+    location = var.region
+
+    tags = merge({
+            openqa_created_by = var.name
+            openqa_created_date = timestamp()
+            openqa_created_id = element(random_id.service.*.hex, 0)
+        }, var.tags)
+}
+
+## virtual network
+
+resource "azurerm_virtual_network" "openqa-network" {
+    name                = "${azurerm_resource_group.openqa-group.name}-vnet"
+    address_space       = ["192.168.0.0/16"]
+    location            = var.region
+    resource_group_name = azurerm_resource_group.openqa-group.name
+}
+
+resource "azurerm_subnet" "openqa-subnet" {
+    name                 = "${azurerm_resource_group.openqa-group.name}-subnet"
+    resource_group_name  = azurerm_resource_group.openqa-group.name
+    virtual_network_name = azurerm_virtual_network.openqa-network.name
+    address_prefixes     = ["192.168.1.0/24"]
+    service_endpoints    = ["Microsoft.Storage"]
+}
+
+resource "azurerm_public_ip" "openqa-publicip" {
+    name                         = "${azurerm_resource_group.openqa-group.name}-public-ip"
+    location                     = var.region
+    resource_group_name          = azurerm_resource_group.openqa-group.name
+    allocation_method            = "Dynamic"
+    count                        = var.instance_count
+}
+
+resource "azurerm_network_security_group" "openqa-nsg" {
+    name                = "${azurerm_resource_group.openqa-group.name}-nsg"
+    location            = var.region
+    resource_group_name = azurerm_resource_group.openqa-group.name
+
+    security_rule {
+        name                       = "SSH"
+        priority                   = 1001
+        direction                  = "Inbound"
+        access                     = "Allow"
+        protocol                   = "Tcp"
+        source_port_range          = "*"
+        destination_port_range     = "22"
+        source_address_prefix      = "*"
+        destination_address_prefix = "*"
+    }
+}
+
+resource "azurerm_subnet_network_security_group_association" "openqa-net-sec-association" {
+    subnet_id                   = azurerm_subnet.openqa-subnet.id
+    network_security_group_id   = azurerm_network_security_group.openqa-nsg.id
+}
+
+resource "azurerm_network_interface" "openqa-nic" {
+    name                      = "${azurerm_resource_group.openqa-group.name}-nic"
+    location                  = var.region
+    resource_group_name       = azurerm_resource_group.openqa-group.name
+    count                     = var.instance_count
+
+    ip_configuration {
+        name                          = "${element(random_id.service.*.hex, count.index)}-nic-config"
+        subnet_id                     = azurerm_subnet.openqa-subnet.id
+        private_ip_address_allocation = "Dynamic"
+        public_ip_address_id          = element(azurerm_public_ip.openqa-publicip.*.id, count.index)
+    }
+}
+
+## storage and NFS share
+
+resource "azurerm_storage_account" "openqa-group" {
+  name                      = "storage${element(random_id.service.*.hex, 0)}"
+  resource_group_name       = azurerm_resource_group.openqa-group.name
+  location                  = azurerm_resource_group.openqa-group.location
+  account_tier              = "Premium"    # Required for NFS share
+  account_replication_type  = "LRS"
+  account_kind              = "FileStorage"
+  enable_https_traffic_only = false
+}
+
+resource "azurerm_storage_account_network_rules" "openqa-group" {
+  depends_on = [azurerm_storage_share.openqa-group]
+
+  storage_account_id         = azurerm_storage_account.openqa-group.id
+  default_action             = "Deny"
+  virtual_network_subnet_ids = [azurerm_subnet.openqa-subnet.id]
+  // AZURE LIMITATION: After setting Deny, we need to allow this host otherwise we cannot do changes or delete the resources
+  ip_rules = [chomp(data.http.myip.body)]
+  
+  private_link_access {
+    endpoint_resource_id     = azurerm_subnet.openqa-subnet.id
+  }
+}
+
+
+resource "azurerm_storage_share" "openqa-group" {
+  name                 = "nfsdata"
+  storage_account_name = azurerm_storage_account.openqa-group.name
+  quota                = 100
+  enabled_protocol     = "NFS"
+}
+
+## virtual machine
+
+resource "azurerm_image" "image" {
+    name                      = "${azurerm_resource_group.openqa-group.name}-disk1"
+    location                  = var.region
+    resource_group_name       = azurerm_resource_group.openqa-group.name
+    count = var.image_id != "" ? 1 : 0
+
+    os_disk {
+        os_type = "Linux"
+        os_state = "Generalized"
+        blob_uri = "https://openqa.blob.core.windows.net/sle-images/${var.image_id}"
+        size_gb = 30
+    }
+}
+
+resource "azurerm_linux_virtual_machine" "openqa-vm" {
+  name                = "${var.name}-${element(random_id.service.*.hex, count.index)}"
+  resource_group_name = azurerm_resource_group.openqa-group.name
+  location            = var.region
+  size                = var.type
+  computer_name       = "${var.name}-${element(random_id.service.*.hex, count.index)}"
+  admin_username      = "azureuser"
+  disable_password_authentication = true
+
+  count                 = var.instance_count
+
+  network_interface_ids = [azurerm_network_interface.openqa-nic[count.index].id]
+
+  tags = merge({
+          openqa_created_by = var.name
+          openqa_created_date = timestamp()
+          openqa_created_id = element(random_id.service.*.hex, count.index)
+      }, var.tags)
+
+  admin_ssh_key {
+    username   = "azureuser"
+    public_key = file("~/.ssh/id_rsa.pub")
+  }
+
+  os_disk {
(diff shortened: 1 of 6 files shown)


tests/publiccloud/azure_nfs.pm
```diff
@@ -0,0 +1,84 @@
...
+    # * Check if the permissiosn are correctly set
+    # * Check if the default user (azureuser) can not read the file
+    # * Check if hard links are working
+    # * Check if soft links are working
+    assert_script_run("cd $share");
+    assert_script_run("curl -v -o Big_Buck_Bunny_8_seconds_bird_clip.ogv " . data_url('Big_Buck_Bunny_8_seconds_bird_clip.ogv'));
+    assert_script_run("sync");
+    assert_script_run("chmod 0400 Big_Buck_Bunny_8_seconds_bird_clip.ogv");
+    assert_script_run("chown root:root Big_Buck_Bunny_8_seconds_bird_clip.ogv");
+    validate_script_output("ls -al Big_Buck_Bunny_8_seconds_bird_clip.ogv", sub { m/^-r--------.*/ });
```
mloviska, line 30: `stat -c "%a" Big_Buck_Bunny_8_seconds_bird_clip.ogv` should return permissions in octal format. That might be considered as more readable

tests/publiccloud/azure_nfs.pm
```diff
@@ -0,0 +1,84 @@
...
+    assert_script_run("! sudo -u azureuser cat Big_Buck_Bunny_8_seconds_bird_clip.ogv >/dev/null");
+    # Check file integrity via md5sums
+    assert_script_run("md5sum Big_Buck_Bunny_8_seconds_bird_clip.ogv > Big_Buck_Bunny_8_seconds_bird_clip.ogv.md5");
+    assert_script_run("mv Big_Buck_Bunny_8_seconds_bird_clip.ogv Big_Buck_Bunny_8_seconds_bird_clip.ogv.orig");
+    # Check hard link
+    assert_script_run("ln Big_Buck_Bunny_8_seconds_bird_clip.ogv.orig Big_Buck_Bunny_8_seconds_bird_clip.ogv");
+    assert_script_run("md5sum -c Big_Buck_Bunny_8_seconds_bird_clip.ogv.md5");
+    assert_script_run("rm Big_Buck_Bunny_8_seconds_bird_clip.ogv");
+    assert_script_run("! stat Big_Buck_Bunny_8_seconds_bird_clip.ogv");
```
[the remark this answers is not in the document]
grisu48: Good point, I will adapt the test accordingly.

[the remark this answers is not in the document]
grisu48: Done, please review again and mark the issue as resolved if you agree

grisu48:
> Part of `data/publiccloud/terraform/azure_nfstest.tf` is a copy of `data/publiccloud/terraform/azure.tf`. What if you add a condition there or something similar so we keep the configuration in one file?
> 
> But otherwise it is nice pull request smile

Because the network configuration is incompatible. There are then too many conditionals, starting from the different NIC to a different subnet, to NSG, to the storage account. It gets too messy.

pdostal (approved):
This is awesome 🥇

tests/publiccloud/azure_nfs.pm
```diff
@@ -0,0 +1,89 @@
...
+    my $instance = $args->{my_instance};
+    my $resource_id = $instance->resource_id;
+    record_info("resource_id", $resource_id);
+
+    # Create mount point and canary to check if remote nfs is mounted
+    assert_script_run('mkdir -p /mount/nfsdata');
+    assert_script_run("echo 'This is the local directory' > /mount/nfsdata/local");
+    # This is the location for the NFS file share created in the azure_nfstest.tf terraform profile
+    my $share = "storage${resource_id}.file.core.windows.net:/storage${resource_id}/nfsdata";
+    script_retry("mount -t nfs \"$share\" /mount/nfsdata -o vers=4,minorversion=1,sec=sys", retry => 3, delay => 120);
```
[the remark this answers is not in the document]
grisu48: IT IS AZURE.

data/publiccloud/terraform/azure_nfstest.tf
```diff
@@ -0,0 +1,258 @@
...
+
+resource "azurerm_storage_account" "openqa-group" {
+  name                      = "storage${element(random_id.service.*.hex, 0)}"
+  resource_group_name       = azurerm_resource_group.openqa-group.name
+  location                  = azurerm_resource_group.openqa-group.location
+  account_tier              = "Premium"    # Required for NFS share
+  account_replication_type  = "LRS"
+  account_kind              = "FileStorage"
+  enable_https_traffic_only = false
+  #allow_blob_public_access  = "false"
```
asmorodskyi, line 157: unless we have really good reason we should try to avoid commented code

variables.md
```diff
@@ -198,6 +198,7 @@ ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in th
 LINUXRC_BOOT | boolean | true | To be used only in scenarios where we are booting an installed system from the installer medium (for example, a DVD) with the menu option "Boot Linux System" (not "boot From Hard Disk"). This option uses linuxrc.
 ZYPPER_WHITELISTED_ORPHANS | string | empty | Whitelist expected orphaned packages, do not fail if any are found. Upgrade scenarios are expecting orphans by default. Used by console/orphaned_packages_check.pm
 PUBLIC_CLOUD_CONTAINER_IMAGES_REPO | string | | The Container images repository in CSP
+PUBLIC_CLOUD_TERRAFORM_FILE | string | | Custom terraform file for publiccloud test runs
```
asmorodskyi, line 201: looks like you forgot about https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14730

lib/publiccloud/provider.pm
```diff
@@ -318,7 +318,12 @@ sub terraform_prepare_env {
         assert_script_run('curl ' . data_url("publiccloud/terraform/sap/$file.tfvars") . ' -o ' . TERRAFORM_DIR . "/$cloud_name/terraform.tfvars");
     }
     else {
-        assert_script_run('curl ' . data_url("publiccloud/terraform/$file.tf") . ' -o ' . TERRAFORM_DIR . '/plan.tf');
+        if (get_var('PUBLIC_CLOUD_TERRAFORM_FILE')) {
```
asmorodskyi, line 321: I am not sure that we actually need this flexibility , usually we exposing something into openQA variable to have ability to clone job with different value for this variable and have different behavior at the end . I don't such possibility for this certain case . If you will try clone Azure NFS test with "normal " `azure.tf` it will not gonna work , right ? Also you can not set here any random file because it should be merged into `data` folder . 
I would recommend to keep "hardcoded" approach as it was before . So this whole block will be : 
```
my $tf_file = "publiccloud/terraform/" . (get_var('PUBLIC_CLOUD_AZURE_NFS_TEST') ? 'azure_nfstest.tf' : "$file.tf" );

assert_script_run('curl ' . data_url($tf_file) . ' -o ' . TERRAFORM_DIR . '/plan.tf');
```

tests/publiccloud/azure_nfs.pm
```diff
@@ -0,0 +1,89 @@
...
+}
+
+sub run {
+    my ($self, $args) = @_;
+    my $instance = $args->{my_instance};
+    my $resource_id = $instance->resource_id;
+    record_info("resource_id", $resource_id);
+
+    # Create mount point and canary to check if remote nfs is mounted
+    assert_script_run('mkdir -p /mount/nfsdata');
```
asmorodskyi, line 61: I think it make sense to create variable with mount point value

lib/publiccloud/provider.pm
grisu48: I believe the ability to inject a custom terraform profile is something that we anyway want to have in the long run. So instead of every time adding another hardcoded conditional, why not provide a custom terraform variable and hit this problem once and forever?

In general I appreciate the simple solution over the more complicated one, but I think having the ability to inject a custom terraform profile into our workflow is more than motivated to have in the longer run. Especially because now we have a use case for using a "non-default" terraform profile.

I believe in the long run it will be simpler to give the flexibility to choose a non-default profile at the level of openQA variables, instead of adding additional conditionals there every time we need to. It might be slimmer now, but then you add another, and another, and another, .... If you need to add a conditional once and you cannot exclude that more will come, then I prefer a strategy that allows us to avoid adding hardcoded conditionals at all.

WDYT?

lib/publiccloud/provider.pm
```diff
@@ -426,20 +431,23 @@ sub terraform_apply {
...
             push @{$vms}, @{$output->{$vm_type . '_name'}->{value}};
             push @{$ips}, @{$output->{$vm_type . '_public_ip'}->{value}};
         }
-    }
-    else {
+    } else {
         $vms = $output->{vm_name}->{value};
         $ips = $output->{public_ip}->{value};
+        # Extract resource ID from output. The output looks like the following:
+        $resource_id = $output->{resource_id}->{value};
```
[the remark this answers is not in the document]
grisu48: I will add this as requested but I don't like this.

Once again: We are adding a hardcoded conditional check with no real value. If the user does not provide this resource in the terraform file, then one would expect that the value is not set or empty. This is the case, and the [GCE](http://duck-norris.qam.suse.de/t8625) test runs shows that there are no side effects.

On the other hand, if we add another terraform profile where we also need the resource_id we then have to change this line in `provider.pm` as well. And probably a engineer gets some gray hair to figure out, why `resource_id` is not set in `instance`, although it is defined in the new terraform profile.

Adding hardcoded checks like this on components that should work with any terraform profiles is something that I don't like because they are a tripping hazard.

tests/publiccloud/azure_nfs.pm
grisu48: Please elaborate. Do you mean a class variable, a openQA variable or a variable that will be passed to the subroutines? And what value would that bring or which use case are you thinking of? :thinking:

variables.md
grisu48: Damn, yes! :sweat_smile:

data/publiccloud/terraform/azure_nfstest.tf
```diff
@@ -0,0 +1,258 @@
...
+    resource_group_name = azurerm_linux_virtual_machine.openqa-vm.0.resource_group_name
+    count               = var.instance_count
+}
+
+output "public_ip" {
+    value = data.azurerm_public_ip.openqa-publicip.*.ip_address
+}
+output "resource_id" {
+    value = "${element(random_id.service.*.hex, 0)}"
+}
```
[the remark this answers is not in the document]
grisu48: done

tests/publiccloud/azure_nfs.pm
```diff
@@ -0,0 +1,89 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Test Azure NFS File shares
```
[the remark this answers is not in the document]
grisu48: I've added a more detailed summary. Please mark as resolved if you agree with my additions.

lib/publiccloud/provider.pm
asmorodskyi: looks like you miss my main point . If we will merge your PR as is right now , we don't have any flexibility. You won't be able to run any tests except Azure NFS with profile which you have created and you won't be able to run Azure NFS test with any other profile than you have created .

tests/publiccloud/azure_nfs.pm
asmorodskyi: just perl variable , use case more clean code where you not repeating same string several times . But if you like to keep it as is it is fine for me

lib/publiccloud/provider.pm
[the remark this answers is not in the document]
asmorodskyi: IMO flexibility which you trying to introduce here is artificial . Our code from several different perspectives is tightly coupled with terraform profiles which we are using . And I don't believe in bright future where we will have code which can run with ANY random terraform profile . Even if I am missing something and it is really possible I am not sure that I really want that future . There is some contract between test code and terraform profile and it is good thing because it gives me some confidence in environment which we have . 
Azure NFS is *first* time our standard profile was not enough to cover our needs . So now we have only 2 profiles for one provider and 1 profile per provider for 2 others . If we will start getting more and more than we can discuss options how to deal with that , but IMO it is too early discussion as for today ...

lib/publiccloud/provider.pm
grisu48: For this test case perhaps not, but for any future test runs, which require a different terraform profile.
Sure, we cannot predict if we ever need this flexibility. But I know that we need to add the ability to run a different terraform profile now and I don't see any good reason, why we should not go half a step further and give ourselves the ability to run any given terraform profile that we want.

While it does not allow us to run any terraform profile, it gives us the chance to run any **compatible** terraform profile (i.e. profiles that work with our test cases). If we then need to add another terraform profile, there is a high chance we don't need to add another conditional to this code segment but can just configure it via the introduced variable.

***

TLDR: This flexibility is cheap and hard-coded conditionals don't give us any benefit either. I don't see why this flexibility should hurt us, while I oppose hard-coded conditionals unless justified by simplicity or by necessity. Those two boundary conditions are IMHO not fulfilled here, as this PR proofs.

WDYT?

asmorodskyi: NOTE: We getting into "programming style" kind of discussion , where is no correct answer but multiply and you can find supporters for each. So I want underline that it is just my personal opinion and I accepting the fact there might be different.

> For this test case perhaps not, but for any future test runs, which require a different terraform profile.

I always against such approach . The functionality which might be used at some point of time should be implemented at this point of time not earlier. Since my statement is fully abstract it is hard to prove it or deny it :). But I can just tell you that I have seen too many  cases where such code hit me back ...

> This flexibility is cheap 

I disagree with this statement from 2 different perspectives : 
1. Most probably you taking into account only "implementation" effort , but you also need to sum up time spent by each engineer who will need to trigger Public Cloud test and spent some extra time looking into code/documentation for yet-another-variable , and plus time when bot-ng/openQA/qac-openqa-yaml will mess up this variable with whatever random reason and someone will need to figure why test is failing and I can ensure you that investigation will be not obvious . And just to remind you as for now this price is for NOTHING as we don't have anything we can take back for it 
2. "Everything is measured in comparison ". You saying that flexibility is cheap and I am saying that it is cheap as effort needed to add same flexibility later when it will actually have a valid use case

grisu48: I was referring to THIS particular flexibility and here my argument still holds :slightly_smiling_face: 

The effort you are referring needs to be invested in all cases, so this point is not valid. Weather you would add an additional conditional or you would simply set the `PUBLIC_CLOUD_TERRAFORM_FILE` variable, any engineer would still need to invest time into openQA-specifics. At no point there was any statement that this would not be the case.

But: Changing a variable is relatively simple, adding a conditional with corresponding VRs and review is expensive (see this thread).

Ad 2) I never made a generalization and absolutely agree with your previous statement, that we should not introduce too much flexibility in places where it is probably never being used. So please also see my code suggestion as a specific case.

grisu48: Issue resolved in discussion, closing thread to avoid unnecessary wall-of-text

lib/publiccloud/provider.pm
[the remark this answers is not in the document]
grisu48: Issue resolved in discussion, closing thread to avoid unnecessary wall-of-text.

PR #14743, opened 2022-04-20
Improve MM test mru-install-multipath-remote

dzedro:
-use verbose (not quiet) in pxe setup, it's better to see what is happening
-increase boot ready_time for PXE custom kernel, it can take some time
-increase installtion timeout when USE_SUPPORT_SERVER_PXE_CUSTOMKERNEL
-drop unnecessary mutexes, not needed because support server can do all the
 configuration without waiting for client
-use default timeout (30) for assert_screen initboot_ESC_prompt

- Related ticket: https://progress.opensuse.org/issues/94264
- Verification run: https://openqa.suse.de/tests/overview?build=pxe&distri=sle, multiple runs see `Next & previous results`

### DIFF
--- a/tests/boot/boot_to_desktop.pm
+++ b/tests/boot/boot_to_desktop.pm
@@ -24,6 +24,7 @@ sub run {
     # We have tests that boot from HDD and wait for DVD boot menu's timeout, so
     # the timeout here must cover it. UEFI DVD adds some 60 seconds on top.
     my $timeout = get_var('UEFI') ? 140 : 80;
+    my $ready_time = get_var('USE_SUPPORT_SERVER_PXE_CUSTOMKERNEL') ? 900 : 500;
     # Increase timeout on ipmi bare metal backend, firmware initialization takes
     # a lot of time
     $timeout += 300 if is_ipmi;
@@ -48,7 +49,7 @@ sub run {
         wait_serial('Welcome to SUSE Linux', $timeout) || die "System did not boot in $timeout seconds.";
     }
     else {
-        $self->wait_boot(bootloader_time => $timeout, nologin => $nologin);
+        $self->wait_boot(bootloader_time => $timeout, nologin => $nologin, ready_time => $ready_time);
     }
 }
 
--- a/data/supportserver/pxe/setup_pxe.sh
+++ b/data/supportserver/pxe/setup_pxe.sh
@@ -3,7 +3,7 @@
 myname="$(basename "$0")"
 kernel_dflt="client/vmlinuz"
 kernelargs_dflt=\
-"initrd=client/initrd splash=off video=1024x768-16 plymouth.ignore-serial-consoles console=ttyS0 console=tty quiet mitigations=auto"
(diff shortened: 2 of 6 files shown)


tests/boot/boot_to_desktop.pm
```diff
@@ -24,6 +24,7 @@ sub run {
     # We have tests that boot from HDD and wait for DVD boot menu's timeout, so
     # the timeout here must cover it. UEFI DVD adds some 60 seconds on top.
     my $timeout = get_var('UEFI') ? 140 : 80;
+    my $ready_time = get_var('USE_SUPPORT_SERVER_PXE_CUSTOMKERNEL') ? 900 : 500;
     # Increase timeout on ipmi bare metal backend, firmware initialization takes
```
foursixnine, line 28: Maybe we want to wait a bit more (twice as much?), for those times when the network is loaded (which is kinda proved by how many boot_to_desktop failed across).

https://openqa.suse.de/tests/8576312#step/boot_to_desktop/10

tests/boot/boot_to_desktop.pm
```diff
@@ -48,7 +49,7 @@ sub run {
         wait_serial('Welcome to SUSE Linux', $timeout) || die "System did not boot in $timeout seconds.";
     }
     else {
-        $self->wait_boot(bootloader_time => $timeout, nologin => $nologin);
+        $self->wait_boot(bootloader_time => $timeout, nologin => $nologin, ready_time => $ready_time);
```
foursixnine, line 52: Same story, https://openqa.suse.de/tests/8574262#step/await_install/98 let's wait a bit longer

foursixnine:
Aside from those two timeouts, looks pretty good!.

tests/boot/boot_to_desktop.pm
dzedro: No, because longer timeout will end with black screen. Test you added used 500 timeout.

tests/boot/boot_to_desktop.pm
dzedro: C'mon, this test is not even related. I don't mind using more time at all, but now it's nearly doubled.
btw many of previous failures happened before timeout increase.

dzedro:
I think main trouble maker is 12 SP3, but this should be soon (30 jun) EOS.

tests/boot/boot_to_desktop.pm
foursixnine: @dzedro I imagined you standing next to Biden:

foursixnine (approved):

tests/boot/boot_to_desktop.pm
dzedro: I do change, run tests if something pop-up like this timeouts I add another change and run tests multiple times again.
Like last page tests should be relevant.

dzedro:
Clones of Todays failures:
Increased timeout for `initboot_ESC_prompt` due to https://openqa.suse.de/tests/8586414
https://openqa.suse.de/tests/8586528
https://openqa.suse.de/tests/8586416

PR #14745, opened 2022-04-20
Fix failing virt_install

grisu48:
Add requires `--osinfo` parameter to the virt-install step of the test.

- Related ticket: https://progress.opensuse.org/issues/107893
- Verification run: https://duck-norris.qam.suse.de/tests/8627


grisu48:
One LGTM for a rather simple fix. Going to merge now.

PR #14747, opened 2022-04-20
JeOS: get info from booted image and store it in DB

jlausuch:
Related request: https://jira.suse.com/browse/TEAM-5808

[15-SP4 kvm](https://openqa.suse.de/tests/8585956#step/image_info/43)
[15-SP4 kvm aarch64](https://openqa.suse.de/tests/8585963#step/image_info/43)
[15-SP4 VMWare](https://openqa.suse.de/tests/8585964#step/image_info/24)
[15-SP4 RPi](https://openqa.suse.de/tests/8585965#step/image_info/24)
[15-SP4 OpenStack](https://openqa.suse.de/tests/8585966#step/image_info/31)
[Tumbleweed](https://openqa.opensuse.org/tests/2307295#step/image_info/43)

This is how the DB looks like:
```
> select * from size
name: size
time                arch    build    distri   flavor                   image                                                                     url                                    value           version
----                ----    -----    ------   ------                   -----                                                                     ---                                    -----           -------
1650525816199643116 x86_64  1.111    sle      JeOS-for-kvm-and-xen     SLES15-SP4-Minimal-VM.x86_64-kvm-and-xen-Build1.111.qcow2                 https://openqa.suse.de/tests/8585956   282.875         15-SP4
1650526511909937628 x86_64  1.111    sle      JeOS-for-OpenStack-Cloud SLES15-SP4-Minimal-VM.x86_64-OpenStack-Cloud-Build1.111.qcow2             https://openqa.suse.de/tests/8585966   275.22314453125 15-SP4
1650526694405757912 aarch64 1.111    sle      JeOS-for-kvm-and-xen     SLES15-SP4-Minimal-VM.aarch64-kvm-Build1.111.qcow2                        https://openqa.suse.de/tests/8585963   267.625         15-SP4
1650526775085662009 x86_64  1.111    sle      JeOS-for-VMware          SLES15-SP4-Minimal-VM.x86_64-VMware-Build1.111.vmdk                       https://openqa.suse.de/tests/8585964   730.0625        15-SP4
1650526795698940617 x86_64  1.111    sle      JeOS-for-MS-HyperV       SLES15-SP4-Minimal-VM.x86_64-MS-HyperV-Build1.111.vhdx                    https://openqa.suse.de/tests/8585967   856             15-SP4
1650526856595254480 aarch64 1.111    sle      JeOS-for-RaspberryPi     SLES15-SP4-Minimal-Image.aarch64-RaspberryPi-Build1.111.raw               https://openqa.suse.de/tests/8585965   2244            15-SP4
1650537481410555479 x86_64  20220419 opensuse JeOS-for-kvm-and-xen     openSUSE-Tumbleweed-JeOS.x86_64-15.1.0-kvm-and-xen-Snapshot20220419.qcow2 https://openqa.opensuse.org/tests/6563 211.25          Tumbleweed
```

### DIFF
--- a/tests/jeos/image_info.pm
+++ b/tests/jeos/image_info.pm
@@ -1,0 +1,127 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: This module is to collect information about JeOS/Minimal-VM
+#          information, like image size, directory sizes, package list,
+#          etc. and push it to an InfluxDB to be displayed in Grafana.
+#          This module doesn't do any interaction with the booted system.
+#          All the operations are done on worker level.
+#          This module is not intended to be a limiting factor to fail a
+#          job, specially pushing data to the DB, therefore no `die`
+#          operations are used.
+# Maintainer: QA-C <qa-c@suse.de>
+
+use Mojo::Base qw(opensusebasetest);
+use testapi;
+use version_utils qw(is_sle is_opensuse is_openstack);
+use mmapi qw(get_current_job_id);
+use Cwd;
+
+sub ls {
+    my $output = `ls -l`;
+    record_info('ls', $output);
+}
+
+sub run {
+    my $self = shift;
+    my $download_url;
+    my $hdd;
+    my $image;
+    my $size;
+
+    my $dir = getcwd;
+    record_info('pwd', $dir);    # For debug purposes
+
+    ls;    # debug
+    if (is_openstack) {
+        # In OpenStack JeOS tests, we boot a different HDD which serves as
+        # a jumphost with the needed CLI tools to upload the image and create
+        # the VM in the remote openStack environment.
+        # The JeOS HDD is not copied in the openQA pool directory, we need to
+        # download using PUBLIC_CLOUD_IMAGE_LOCATION variable
+        my $img_url = get_required_var('PUBLIC_CLOUD_IMAGE_LOCATION');
+        record_info('URL', "Downloading $img_url ...");
+        system("curl -O $img_url");
+        ($hdd) = $img_url =~ /([^\/]+)$/;
+    } else {
+        $hdd = get_required_var('HDD_1');
+    }
+
+    record_info('HDD', $hdd);
+    ls;
+
+    if ($hdd =~ /\.xz/) {
+        # We want to monitor the size of uncompressed images.
+        my $cmd = "nice ionice unxz -k $hdd -c > hdd_uncompressed";
+        record_info('unxz', "Extracting compressed file to get its size.\n$cmd");
+        system($cmd);
+        $size = -s 'hdd_uncompressed';
+        ls;    # debug
+        system("rm hdd_uncompressed");
+        ($image = $hdd) =~ s/\.xz//;
+    } else {
+        $image = $hdd;
+        $size = -s $hdd;
+    }
+    my $size_mb = $size / 1024 / 1024;
+    record_info('Image', "Image: $image\nSize: $size_mb");
+
+    # Record this value in the InfluxDB
+    # We only allow this on certain conditions:
+    #  - The job must be executed from OSD or O3
+    #  - The job must contain INFLUXDB_SERVER and _SECRET_INFLUXDB_USER and _SECRET_INFLUXDB_PWD
+    #  - The job shouldn't be a verification run => CASEDIR variable
+    my $influxdb_server = get_var('INFLUXDB_SERVER');
+    my $influxdb_user = get_var('_SECRET_INFLUXDB_USER');
+    my $influxdb_pwd = get_var('_SECRET_INFLUXDB_PWD');
+    #return unless (get_var('CASEDIR') || $influxdb_server || $influxdb_user || $influxdb_pwd); #this should remain before merging
+    return unless ($influxdb_server || $influxdb_user || $influxdb_pwd);    # remove CASEDIR condition to allow running VR before merging
+
+    my $job_url;
+    my $openqa_host = get_required_var('OPENQA_HOSTNAME');
+    if ($openqa_host =~ /openqa1-opensuse|openqa.opensuse.org/) {    # O3 hostname
+        $job_url = 'https://openqa.opensuse.org/tests/' . get_current_job_id();
+    }
+    elsif ($openqa_host =~ /openqa.suse.de/) {    # OSD hostname
+        $job_url = 'https://openqa.suse.de/tests/' . get_current_job_id();
+    }
+
+    return unless ($job_url);
+
+    # Check if the data for this image has been already published before.
+    # This will avoid publishing the same data if a job is restarted.
+    system("echo '-u $influxdb_user:$influxdb_pwd' > $dir/influxdb_conf");
+
+    my $query = 'curl -K ' . $dir . '/influxdb_conf -i -G "http://' . $influxdb_server . ':8086/query?db=data"';
+    $query .= ' --data-urlencode "q=SELECT \"value\" FROM \"image_size\" WHERE \"image\" = \'' . $image . '\'"';
+    my $query_output = `$query 2>&1`;
+    record_info('DB check', "Check if this value exists in the DB.\nquery = $query\nresult = \n$query_output");
+    # Successful query should return 'HTTP/1.1 200 OK'
+    # Empty records will return  '{"results":[{"statement_id":0}]}'
+    # Existing records will return '{"results":[{"statement_id":0,"series":[{"name":"size","columns":["time","value"],"values"' ...
+    if ($query_output !~ /200 OK/) {
+        record_soft_failure("poo#110221 - There has been a problem with the query.");
+    } elsif ($query_output =~ /values/) {
+        record_info('Skip', "Skip pushing the data to the DB. There is already a DB record for this image.");
+    } else {
+        $query = 'curl -K influxdb_conf -i -XPOST "http://' . $influxdb_server . ':8086/write?db=data"';
+        $query .= ' --data-binary "image_size';
+        $query .= ',distri=' . get_required_var('DISTRI');
+        $query .= ',version=' . get_required_var('VERSION');
+        $query .= ',arch=' . get_required_var('ARCH');
+        $query .= ',flavor=' . get_required_var('FLAVOR');
+        $query .= ',build=' . get_required_var('BUILD');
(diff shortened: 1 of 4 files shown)


tests/jeos/image_info.pm
```diff
@@ -0,0 +1,97 @@
...
+    # Check url exists
+    if (script_output("curl -I -k $asset_url") !~ /HTTP.*200/) {
+        record_soft_failure("Asset not found in $asset_url");
+        return;
+    }
+    (my $image = $asset) =~ s/\.xz//;    # in case the image is compressed
+    record_info('Image', $image);
+
+    # We shall skip this module for verification runs with special build numbers
+    if (get_required_var('BUILD') =~ /os-autoinst-distri-opensuse/) {
```
Vogtinator, line 55: I do verification runs using `openqa-clone-job`, so only `CASEDIR` and `_GROUP` are different. Something similar appears to be the case for automatic investigation jobs like https://openqa.opensuse.org/tests/2307416#settings. I suggest to check for `/` in `BUILD` or any non-default `CASEDIR`, but the latter might also have false positives in case an explicit version is given.

Or maybe just hardcode the `_GROUP` numbers for production? Not sure whether that is available in test modules at all.

tests/jeos/image_info.pm
```diff
@@ -0,0 +1,97 @@
...
+        record_info('Uncompress', 'We need to get the size of the uncompressed image.');
+        assert_script_run("nice ionice unxz -k -f $asset", timeout => 300);
+    }
+    my $bytes = script_output('stat -c%s ' . $image);
+    my $size_mb = $bytes / 1024 / 1024;
+    record_info('Size', "The size of the image $image is $size_mb MB");
+
+    # Check if the data for this image has been already published before.
+    # This will avoid publishing the same data if a job is restarted.
+    my $query = 'curl -i -G "http://35.158.113.219:8086/query?db=data" --data-urlencode';
```
Vogtinator, line 72: Shouldn't this also use `INFLUX_SERVER`?

jlausuch:
> Currently the test actually downloads the image (and the uncompressed data, if applicable) onto the SUT, which is a bit weird.
> 
> The test module is run on the host, which already has the compressed (and uncompressed, but possibly in use) image available. That should simplify the code a bit, like `isosize.pm`:
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1c411867e3f00a3bad368e76da86b9963e8ac510/tests/installation/isosize.pm#L18-L19

This is a very good hint. I have done some tests using that magic with `-s $hdd` and it works.
See https://openqa.suse.de/tests/8591093#step/image_info/1
which obtains `191986252` from the HDD (including .xz), so:
```
$ stat -c%s SLES15-SP4-Minimal-VM.x86_64-VMware-Build1.111.vmdk.xz
191986252
$ stat -c%s SLES15-SP4-Minimal-VM.x86_64-VMware-Build1.111.vmdk   
765526016
```

However, I don't know how to get the size of the uncompressed image which is booted... I have tried to do:
```
my $hdd = get_var('HDD_1'); 
(my $uncompresed_hdd = $hdd) =~ s/\.xz//;
 my $size = -s $uncompresed_hdd; 
 ```
 but it doesn't return anything.

**Comment by jlausuch** [comment]
> I suggest a new column `size_type` so that if the method for size determination changes in the future, that's still comparable or visible. Could also be used to express multiple types of size (compressed image, used blocks, btrfs data size, ...)

Ok. I can start with adding `size_type=raw|compressed` or similar. For other type of measurements, like file system information I would create another table in the DB.

**Comment by Vogtinator** [comment]
> However, I don't know how to get the size of the uncompressed image which is booted... I have tried to do:
> 
> ```
> my $hdd = get_var('HDD_1'); 
> (my $uncompresed_hdd = $hdd) =~ s/\.xz//;
>  my $size = -s $uncompresed_hdd; 
> ```
> 
> but it doesn't return anything.

Question is where `HDD_1` points to. The shared cache directory or the worker pool?

At least for `.raw.xz` in the RPi case, the uncompression is only in the worker pool IIRC.

So you might have to run `unxz` on the worker.

**Comment by jlausuch** [comment]
I have done some changes taking into account all your comments.
Now the test is not executed on the host itself, but on the worker running the commands on the openQA host/worker side directly.

The only peculiarity is in OpenStack tests, where the image which is booted in the remote OpenStack env is not synced to the worker pool directory, only the HDD that we boot (public cloud tools helper VM).

As @Vogtinator suggested, we could consider this test similar to `isosize`, but being this a bit more complete having the possibility to inject data to some DB.

I have also added some security to the DB (before it was exposed without user/pwd).

**Comment by jlausuch** [comment]
Amazing... workers in O3 can't do `curl` locally... 
https://openqa.opensuse.org/tests/2308935/logfile?filename=autoinst-log.txt

```
[2022-04-22T18:12:11.053568+02:00] [debug] <<< testapi::record_info(title="Push", output="Pushing the image size to the DB.\ncurl -i -XPOST \"[http://35.158.113.219:8086/write?db=data&u=openqa_user&p=openqa_pwd\"](http://35.158.113.219:8086/write?db=data&u=xyz&p=xyz\) --data-binary \"image_size,distri=opensuse,version=Tumbleweed,arch=x86_64,flavor=JeOS-for-kvm-and-xen,build=20220420,image=openSUSE-Tumbleweed-JeOS.x86_64-15.1.0-kvm-and-xen-Snapshot20220420.qcow2,job_url=https://openqa.opensuse.org/tests/2308935,type=uncompressed value=211.25\"", result="ok")
sh: /usr/bin/curl: Permission denied
```

**Comment by Vogtinator** [comment]
apparmor...

**Comment by jlausuch** [comment]
> apparmor...

any ideas how to workaround this?

**Comment by Vogtinator** [comment]
Using perl code directly instead of calling curl should work.

**Comment by jlausuch** [comment]
> Using perl code directly instead of calling curl should work.

I wish... 
```
[2022-04-22T19:25:25.899073+01:00] [warn] !!! autotest::loadtest: error on tests/jeos/image_info.pm: Can't locate WWW/Curl.pm in @INC (you may need to install the WWW::Curl module) 
```

File: tests/jeos/image_info.pm [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+    ls;    # debug
+    if (is_openstack) {
+        # In OpenStack JeOS tests, we boot a different HDD which serves as
+        # a jumphost with the needed CLI tools to upload the image and create
+        # the VM in the remote openStack environment.
+        # The JeOS HDD is not copied in the openQA pool directory, we need to
+        # download using PUBLIC_CLOUD_IMAGE_LOCATION variable
+        my $img_url = get_required_var('PUBLIC_CLOUD_IMAGE_LOCATION');
+        record_info('URL', "Downloading $img_url ...");
+        system("curl -O $img_url");
```
pdostal, line 46: Will the downloaded file be removed after the test exits?

tests/jeos/image_info.pm
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+        $query = 'curl -K influxdb_conf -i -XPOST "http://' . $influxdb_server . ':8086/write?db=data"';
+        $query .= ' --data-binary "image_size';
+        $query .= ',distri=' . get_required_var('DISTRI');
+        $query .= ',version=' . get_required_var('VERSION');
+        $query .= ',arch=' . get_required_var('ARCH');
+        $query .= ',flavor=' . get_required_var('FLAVOR');
+        $query .= ',build=' . get_required_var('BUILD');
+        $query .= ',image=' . $image;
+        $query .= ',job_url=' . $job_url;
```
pdostal, line 117: Why do we need this?

tests/jeos/image_info.pm
jlausuch: This is downloaded in the pool directly, so it will be removed entirely after test is finished, but I could do it explicitly here.

tests/jeos/image_info.pm
jlausuch: It would be good to have the information of which job posted the value. This is just metadata, so the more info we have, the easier to troubleshoot.

Vogtinator:
What's the current plan? Port this over to libcurl, then merge #14767, merge #14767 then try libcurl or stay as-is using curl and allow calling curl from test code in the apparmor profile?

tests/jeos/image_info.pm
grisu48: The `CASEDIR` condition alone is not sufficient. I occasionally clone jobs to my own instance and we don't want those test runs to contaminate the database.

Vogtinator:
> > What's the current plan? Port this over to libcurl, then merge #14767, merge #14767 then try libcurl or stay as-is using curl and allow calling curl from test code in the apparmor profile?
> 
> I would prefer adding curl to apparmor profile as it is right now. This way, we have the exact shell command we can copy paste for manual debugging. I have manged to hide the credentials in a file and call it from curl command, so it's safer now.

Ok, that's something for https://github.com/os-autoinst/openQA/blob/master/profiles/apparmor.d/usr.share.openqa.script.worker

jlausuch:
> > > What's the current plan? Port this over to libcurl, then merge #14767, merge #14767 then try libcurl or stay as-is using curl and allow calling curl from test code in the apparmor profile?
> > 
> > 
> > I would prefer adding curl to apparmor profile as it is right now. This way, we have the exact shell command we can copy paste for manual debugging. I have manged to hide the credentials in a file and call it from curl command, so it's safer now.
> 
> Ok, that's something for https://github.com/os-autoinst/openQA/blob/master/profiles/apparmor.d/usr.share.openqa.script.worker

Yes. Let's move forward with this, enable the DB-push only in OSD until I figure out how to modify the apparmor profile accordingly.

mloviska:
This super nice, but I think we will have problems with `svirt` since only 2 images are different, could try a VR for `vmware` and `hyperv`, please?

jlausuch:
> This super nice, but I think we will have problems with `svirt` since only 2 images are different, could you try a VR for `vmware` and `hyperv`, please?

VMWare is covered https://openqa.suse.de/tests/8625990#step/image_info/9
I thought I ran one for hyperv. Doing it now.

jlausuch:
> This super nice, but I think we will have problems with `svirt` since only 2 images are different, could you try a VR for `vmware` and `hyperv`, please?

and Hyper-V: https://openqa.suse.de/tests/8652528#step/image_info/7

jlausuch:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14803

PR #14749, opened 2022-04-21
Fix gnote failure on GNOME42

GraceWang571:
gnote won't be installed by default on SLE start from 15SP4, so we need to install it before testing.
GNOTE42 has a new UI, so we need update test cases to make them work well on GNOTE42.

- Related ticket: https://progress.opensuse.org/issues/108887
- Needles: already added when debugging on osd and o3
- Verification run: 
- SLE https://openqa.suse.de/tests/8580413
- TW https://openqa.opensuse.org/tests/2306329


zhaocongjia:
LGTM.

PR #14750, opened 2022-04-21
Increase timeout for slower architectures

rakoenig:
This commit increases the timeout for the AddOnProductPage to 240 seconds. This avoids that tests using this page time out on slower plattforms like aarch64.

- Related ticket: https://progress.opensuse.org/issues/108758
- Needles: -
- Verification run: https://openqa.suse.de/tests/8585932#


lib/Installation/AddOnProduct/AddOnProductPage.pm
```diff
@@ -9,6 +9,7 @@
 
 package Installation::AddOnProduct::AddOnProductPage;
 use parent 'Installation::Navigation::NavigationBase';
+use YuiRestClient;
```
jknphy, line 12: not needed.

lib/Installation/AddOnProduct/AddOnProductPage.pm
```diff
@@ -22,7 +23,7 @@ sub init {
 
 sub is_shown {
     my ($self) = @_;
-    return $self->{rdb_specify_url}->exist();
+    return $self->{rdb_specify_url}->exist({timeout => 240});
```
jknphy, line 26: let's add a simple comment here about why we do this mentioning the System Probing screen, wdyt?

jknphy (commented):
LGTM, quite more simple than the initial approach.
Could you please, provide more specific commit title? `Increase timeout for slower architectures` might sound as a heavy change for a git-log-commit reader.

lib/Installation/AddOnProduct/AddOnProductPage.pm
rakoenig: removed

lib/Installation/AddOnProduct/AddOnProductPage.pm
rakoenig: comment added

rakoenig: comment added

PR #14752, opened 2022-04-21
Add query for SIE on s390x

grisu48:
Check for SIE (start-interpretive execution) support on the CPU, which
is a required feature for native kvm on s390x. Fail the test if SIE is
not a listed feature of the CPU.

- Related ticket: https://progress.opensuse.org/issues/110155
- Verification run: https://openqa.suse.de/t8585957 (passing) | https://openqa.suse.de/t8585957 (failing, CPU does not support SIE)

### DIFF
--- a/tests/qemu/kvm.pm
+++ b/tests/qemu/kvm.pm
@@ -36,6 +36,8 @@ sub run {
         }
     }
     elsif (is_s390x) {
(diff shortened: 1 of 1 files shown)


asdil12 (approved):
LGTM

jlausuch:
Thanks!

PR #14756, opened 2022-04-21
WIP: Mark newly uploaded GCE images as SEV compatible

grisu48:
Newly uploaded GCE are capable of confidential compute
(SEV, Secure Encrypted Virtualization) and should be marked as such by
default.

- Related ticket: https://progress.opensuse.org/issues/110088
- Verification run: TBD


grisu48:
SEV is available since Kernel Linux 4.16 and SLES 15-SP3 (see e.g. https://documentation.suse.com/sles/15-SP3/pdf/article-amd-sev_color_en.pdf). Since we are using `upload_img` only for Product-QA I believe adding this new feature by default should be sane.

This marks also the image as capable of running confidential compute instances, so also if this would be applied to an older image version, unless the user tries to dedicatedly boot that image in confidential compute mode, nothing changes. It can be seen as "removing a warning sign" but with no functional implications.

The default values can still be overwritten by setting the `PUBLIC_CLOUD_GCE_UPLOAD_GUEST_FEATURES` variable in case of issues.

PR #14759, opened 2022-04-22
Redesign check_system_info 

lemon-suse:
The original design for check_system_info is to check addons and modules register status before and after migration, but some other test added to the module later that's is not expected, I move migration feature related test which only check on target system to regression test, move nodejs to service check.

- Related ticket: https://progress.opensuse.org/issues/109998
- Needles: N/A
- Verification run: 
  x86:
http://openqa.nue.suse.com/tests/8585985#live  offline 15SP3
https://openqa.nue.suse.com/tests/8625216#step/install_service/350  online 15SP3

s390x:
http://openqa.nue.suse.com/tests/8585689#live offline 15SP1
http://openqa.nue.suse.com/tests/8585690#live online 15SP3

ppc64le:
http://openqa.nue.suse.com/tests/8585691#live offline 15SP3
http://openqa.nue.suse.com/tests/8585692#live online 15SP2

aarch64:
http://openqa.nue.suse.com/tests/8585693#live offline 15SP2
http://openqa.nue.suse.com/tests/8585694#live online 15SP3

### DIFF
--- a/lib/services/nodejs.pm
+++ b/lib/services/nodejs.pm
@@ -1,0 +1,67 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Package for nodejs update tests
+#
+# Maintainer: Lemon Li <leli@suse.com>
+
+package services::nodejs;
+use base 'opensusebasetest';
+use testapi;
+use utils;
+use strict;
+use warnings;
+
+my $service_type = 'Systemd';
+
+# check nodejs package installed as expected
+sub check_installed {
+    assert_script_run('grep -E "Installing: nodejs16" /tmp/install_nodejs.log');
+    script_run('zypper se -i nodejs | tee /tmp/search_nodejs', timeout => 60);
+    my $ret_s1 = script_output(q(grep package /tmp/search_nodejs | grep nodejs16 | awk -F\| '{print $1}'));
+    my $ret_s2 = script_output(q(grep package /tmp/search_nodejs | grep nodejs-common| awk -F\| '{print $1}'));
+    die "Expected package is not installed. ret_s1=$ret_s1' and ret_s2=$ret_s2'" if ('i' ne "$ret_s1" || 'i+' ne "$ret_s2");
+}
+
+# remove the nodejs and related log after test
+sub clean_nodejs {
+    zypper_call("rm nodejs-common");
+    assert_script_run('rm /tmp/install_nodejs.log');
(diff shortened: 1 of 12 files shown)


tests/console/check_migration_features.pm
```diff
@@ -0,0 +1,78 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Package: SUSEConnect
+# Summary: Verify milestone version and display some info.
+# Check product info before and after migration
```
tinawang123, line 6: Need update those summary description.

lib/services/nodejs.pm
```diff
@@ -0,0 +1,62 @@
...
+# stage is 'before' or 'after' system migration.
+sub full_nodejs_check {
+    my (%hash) = @_;
+    my ($stage, $type) = ($hash{stage}, $hash{service_type});
+    $service_type = $type;
+    if ($stage eq 'before') {
+        check_nodejs_common();
+    }
+    check_installed();
+    check_nodejs_common();
```
hjluo, line 58: if the stage is before then check_installed will run 3 times with this logic.
I'd suggest to change check_nodejs_common as 
check_nodejs_common($stage) like
{
 $stage = @_;
 if ($stage eq 'before') {
   do the before related action
  } else {
    do the after migration related actions
    }
}

tests/console/check_migration_features.pm
lemon-suse: Updated, thanks.

lib/services/nodejs.pm
lemon-suse: Yes, updated code. Please have a look.

tinawang123 (approved):
LGTM

tests/console/check_migration_features.pm
```diff
@@ -0,0 +1,77 @@
...
+
+use base "consoletest";
+use strict;
+use warnings;
+use testapi;
+use Utils::Architectures;
+use utils;
+use version_utils;
+use registration;
+use List::MoreUtils 'uniq';
```
hjluo, line 16: use registration;
use List::MoreUtils 'uniq';

these 2 modules maybe not used here

lemon-suse: Updated.

lib/services/nodejs.pm
hjluo: LGTM

hjluo:
LGTM

chcao (approved):
LGTM

coolgw (approved):

PR #14760, opened 2022-04-22
Add mandatory packages for ipmi2qemu

XGWang0:
Add mandatory packages

- Verification run: http://openqa.qa2.suse.asia/tests/47519


museQi:
LGTM

PR #14764, opened 2022-04-22
Fix failing samba test runs for Tumbleweed

grisu48:
net share does not list the username when listing the shares available
on the server. This commit checks for the presence of the default share names
instead of the username.

- Related ticket: https://progress.opensuse.org/issues/110049
- Verification run: https://duck-norris.qam.suse.de/tests/8804


asdil12 (approved):
LGTM

PR #14766, opened 2022-04-22
Add AutoYaST in YaST Maintenance Updates

JRivrain:
Replace mru_install with an autoyast install for yast maintenance.

Currently mau-extratests-yast2ui and mau-extratests-yast2ui-textmode
are depending on mru-install-with-addons. We want to use a ligthter, and 
easier-to-maintain (for us) job as a parent job. 

for now this is limited to sle 15 sp3.

- Related ticket: https://progress.opensuse.org/issues/109214
- Verification run: 
mru_install
http://waaa-amazing.suse.cz/tests/16803
http://waaa-amazing.suse.cz/tests/16802
dependencies
http://waaa-amazing.suse.cz/tests/16805
http://waaa-amazing.suse.cz/tests/16804

Without autoyast/yaml and new parameter http://waaa-amazing.suse.cz/tests/16800

Gitlab MR https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/253/diffs

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
+++ b/data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
@@ -1,0 +1,219 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+    <suse_register>
+        <do_registration config:type="boolean">true</do_registration>
+        <email><%= $get_var->('SCC_EMAIL') %></email>
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        <install_updates config:type="boolean">true</install_updates>
+        % if (keys %$addons) {
+        <addons config:type="list">
+            % while (my ($key, $addon) = each (%$addons)) {
+            <addon>
+                <name><%= $addon->{name} %></name>
+                <version><%= $addon->{version} %></version>
+                <arch><%= $addon->{arch} %></arch>
+                % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sles')) {
+                <reg_code><%= $get_var->('SCC_REGCODE_WE') %></reg_code>
+                % }
+                % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+                <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+                % }
+                % if ($key eq 'rt') {
+                <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+                % }
+                % if ($key eq 'ltss') {
+                <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
+                % }
+            </addon>
+            % }
+        </addons>
+        %}
+    </suse_register>
+    <add-on>
+        <add_on_products config:type="list">
+            % for my $repo (@$repos) {
+            % my ($repo_id, $addon) = $repo =~ (/(\d+)\/(.*)\//);
+            <listentry>
+                <media_url><%= $repo %></media_url>
+		        <name><%= $addon . "_" . $repo_id %></name>
+		        <alias><%= $addon . "_" . $repo_id %></alias>
+            </listentry>
+            % }
+        </add_on_products>
+    </add-on>
+    <bootloader>
+        <global>
+            <timeout config:type="integer">45</timeout>
+        </global>
+    </bootloader>
+    <general>
+        <mode>
+            <confirm config:type="boolean">false</confirm>
+        </mode>
+        <signature-handling>
+            <accept_unsigned_file config:type="boolean">true</accept_unsigned_file>
+            <accept_unknown_gpg_key config:type="boolean">true</accept_unknown_gpg_key>
+            <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
+            <import_gpg_key config:type="boolean">true</import_gpg_key>
+        </signature-handling>
+    </general>
+    <partitioning config:type="list">
+        <drive>
+            <initialize config:type="boolean">true</initialize>
+            <use>all</use>
+        </drive>
+    </partitioning>
+    <networking>
+        <interfaces config:type="list">
+            <interface>
+                <bootproto>dhcp</bootproto>
+                <device>eth0</device>
+                <dhclient_set_default_route>yes</dhclient_set_default_route>
+                <startmode>auto</startmode>
+            </interface>
+        </interfaces>
+    </networking>
+    <firewall>
+        <enable_firewall config:type="boolean">true</enable_firewall>
+        <start_firewall config:type="boolean">true</start_firewall>
+        <zones config:type="list">
+            <zone>
+                <name>public</name>
+                <interfaces config:type="list">
+                    <interface>eth0</interface>
+                </interfaces>
+                <services config:type="list">
+                    <service>ssh</service>
+                </services>
+                <ports config:type="list">
+                    <port>22/tcp</port>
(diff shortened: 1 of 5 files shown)


schedule/yast/sle_15_sp3/create-hdd-y2-maintenance-with-addons.yaml
```diff
@@ -0,0 +1,23 @@
+---
+name: create-hdd-y2-maintenance-with-addons
```
jknphy: I would suggest to use `_` it is easier to select text in most editors and remove the `y2`. So something like `autoyast_create_hdd_mru_all_patterns` and `autoyast_create_hdd_mru_minimal_patterns`. wdyt?

jknphy: ahm ok it is only one, so just we should change `create_*` by `autoyast_create_hdd_maintenance.`

tests/installation/prepare_maintenance_env.pm
```diff
@@ -0,0 +1,37 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: Prepare test repositories for children jobs.
+# - Disable Nvidia repos that may have been automatically added as part of autoinstall.
```
b10n1k, line 5: i am not sure if this is accurate. Can you confirm?

data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
```diff
@@ -0,0 +1,220 @@
...
+                % }
+            </addon>
+            % }
+        </addons>
+        %}
+    </suse_register>
+    <add-on>
+        <add_on_products config:type="list">
+            % for my $repo (@$repos) {
+                % my $addon = (split '_', $repo)[2];
```
jknphy, line 36: I would recommend to go with something more resistant like:
`my ($num, $name) = $repo =~ (/(\d+)\/(.*)\//);`
Index needs to be avoided as much as possible, url could change and we just need a number and the following string where we don't need to cut the arch. output of that would be for example:
```
23978
SUSE_Updates_SLE-Module-Containers_15-SP3_x86_64
```

tests/installation/prepare_maintenance_env.pm
```diff
@@ -0,0 +1,37 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: Prepare test repositories for children jobs.
```
jknphy, line 4: other teams seems to have its own `patch_and_reboot.pm`, we don't need to invent new name and the reboot should be in the same file, because `tests/shutdown/reboot.pm` as I mentioned to you offline sounds very generic, each arch does things on its way and contributors will complain or would not know exact use case of this.
The trend now is to have each squad their own file, in the future there are some plan to organize the repo, so that is yet another reason.

jknphy: existing summary description was more clear:
```
# Summary: QA Automation: patch the system before running the test
#          This is to test Test Updates
```
Test files do not care about children jobs, that is an openQA setting/relationship.
Subsequent tests could be run right after the installation and not in child jobs for example.

tests/installation/prepare_maintenance_env.pm
jknphy: apparently this was done twice, so in the autoyast profile might suffice:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a8142f8da6745b5b09e2d77369f4ae436ac4ab1c/tests/qa_automation/patch_and_reboot.pm#L38
but take into account that this file was copy/pasted from https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_qam/15_installation.xml.ep
My initial suggestion was to clone the compact profile and put on top the templating, but it was done reusing that file above, so in this case we can optimize things later if something is not needed.

tests/installation/prepare_maintenance_env.pm
```diff
@@ -0,0 +1,37 @@
...
+use warnings;
+use utils;
+use testapi;
+
+sub run {
+    my $self = shift;
+    $self->select_serial_terminal;
+    quit_packagekit unless check_var('DESKTOP', 'textmode');
+    # Disable gpg check for all maintenance repositories.
+    zypper_call('--gpg-auto-import-keys ref');
```
jknphy, line 22: we should keep the extra params and the comment:
```
    # refresh repositories, inf 106 is accepted because repositories with test
    # can be removed before test start
    zypper_call('ref', timeout => 1400, exitcode => [0, 106]);
```

tests/shutdown/reboot.pm
```diff
@@ -0,0 +1,24 @@
...
+
+use strict;
+use warnings;
+use base "opensusebasetest";
+use testapi;
+use power_action_utils 'power_action';
+
+sub run {
+    my $self = shift;
+    power_action('reboot', textmode => 1);
```
jknphy, line 16: we should keep the comment as well:
`# DESKTOP can be gnome, but patch is happening in shell, thus always force reboot in shell`

tests/installation/prepare_maintenance_env.pm
```diff
@@ -0,0 +1,37 @@
...
+    return {fatal => 1};
+}
+
+sub post_run_hook {
+    assert_script_run("rpm -ql --changelog kernel-default > /tmp/kernel_changelog.log");
+    assert_script_run("zypper lr -u >/tmp/repos_list.txt");
+    upload_logs('/tmp/kernel_changelog.log');
+    upload_logs('/tmp/repos_list.txt');
+}
+
```
jknphy, line 36: no flag `fatal => 1` but as suggested we we can just have one file.

tests/installation/prepare_maintenance_env.pm
```diff
@@ -0,0 +1,37 @@
...
+# - Disable Nvidia repos that may have been automatically added as part of autoinstall.
+# - Import maintenance repos gpg keys
+# - Run zypper patch
+#
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use utils;
```
jknphy, line 14: utils is pretty huge, we should just import the functions needed.

jknphy (commented):
It is looking good, I think we don't gain much creating yet another file for path_and_reboot, we just need to pass a parameter to `fully_patch_system`, honestly I would hack it, passing it some new? setting. It will be less mess that trying to put all in libraries when the trend is the opposite.

tests/installation/prepare_maintenance_env.pm
JRivrain: Good catch, I actually removed the line since it's being removed in the autoyast. I'll remove it.

data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
JRivrain: That will do, thanks !

lib/qam.pm
```diff
@@ -84,9 +84,15 @@ sub add_test_repositories {
...
+        # we still need to disable gpg check for all repositories.
+        zypper_call('--gpg-auto-import-keys ref', timeout => 1400, exitcode => [0, 106]);
+    } else {
+        for my $var (@repos) {
+            zypper_call("--no-gpg-checks ar -f $gpg -n 'TEST_$counter' $var 'TEST_$counter'");
+            $counter++;
+        }
     }
 
     if (is_sle('=12-SP2')) {
```
JRivrain, line 98: In the end, I modified the lib as we may need the piece of code below, this way we can apply this change in every maintained product, the oldest being sle12-sp2, as far as I could see.

jknphy:
LGTM, besides that minor change, we need to update the MR as well before merging.

PR #14767, opened 2022-04-22
Add WWW::Curl as a perl dependency

jlausuch:
In order to run some operations on the worker directly, like collecting some information of the HDD which is booted and store this information is a remote DB, curl is needed. Using `WWW::Curl` instead of running `curl` in a shell offers some advantages, like setting some options before doing the call, getting the return code and the output, etc.

Also, this would be needed by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14747


jlausuch:
I decided to modify app armor instead.

PR #14772, opened 2022-04-24
xfstests: Add feature to inject code for debugging

SeroSun:
Introduce new parameter INJECT_INFO to add 1 line code in a subtest. INJECT_INFO format in "SUBTEST_NAME INJECT_LINE_NUMBER INJECT_CODE".
e.g. When need to make a xfs_metadump before xfs_repair, assume xfs_repair in line 35, then you could configure like this:
INJECT_INFO=“xfs/xxx:35:‘_scratch_xfs_metadump /opt/log/xfs/xxx-metadumpfile’”

- Related ticket: https://progress.opensuse.org/issues/56237
- Verification run: 
- 1. inject a line in xfs/220 and check if its log has metadump after test or not https://10.67.133.102/tests/630
- 2. xfs test inject nothing https://10.67.133.102/tests/614

### DIFF
--- a/data/xfstests/wrapper.sh
+++ b/data/xfstests/wrapper.sh
@@ -2,11 +2,14 @@
 XFSTESTS_DIR='/opt/xfstests'
 PROG="$XFSTESTS_DIR/check"
 SCRIPT_DIR=$(realpath $(dirname "$0"))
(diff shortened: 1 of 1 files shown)


lansuse:
LGTM

PR #14774, opened 2022-04-25
Workaround for bsc#1189543, before de-register system need register python2

tinawang123:
Workaround for bsc#1189543, before de-register system need register
python2. Then de-register the system it can work.

- Related ticket: https://progress.opensuse.org/issues/110214
- Verification run: https://openqa.nue.suse.com/tests/8627882#

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -867,6 +867,11 @@ sub scc_deregistration {
         quit_packagekit;
         wait_for_purge_kernels;
         assert_script_run('SUSEConnect --version');
+        if (get_var('UPGRADE_TARGET_VERSION') == '15-SP3') {
+            # Workaround for bsc#1189543, need register python2 before de-register system
+            record_soft_failure 'bsc#1189543 - Stale python2 module blocks de-registration after system migration';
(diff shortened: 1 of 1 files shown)


lib/registration.pm
```diff
@@ -867,6 +867,11 @@ sub scc_deregistration {
         quit_packagekit;
         wait_for_purge_kernels;
         assert_script_run('SUSEConnect --version');
+        if (get_var('UPGRADE_TARGET_VERSION') == '15-SP3') {
+            # https://bugzilla.suse.com/show_bug.cgi?id=1189543
```
lemon-suse, line 871: I don't think comments bug here is a must(code already have it), while it is better to comments why and how we workaround the bug.

lib/registration.pm
```diff
@@ -867,6 +867,11 @@ sub scc_deregistration {
...
+            record_soft_failure 'bsc#1189543';
```
lemon-suse, line 872: I think add the bug title in the record_soft_failure will help others to know the bug info easily.

lemon-suse (approved):

PR #14779, opened 2022-04-25
Add rootless flash script for generalhw backend

asdil12:
- Verification run: https://openqa.suse.de/tests/8629979

This requires some udev rules to set the permissions of the `/dev/sg*` and `/dev/sd*` devices of the usb-sd-mux devices:
```
ACTION=="add", SUBSYSTEM=="scsi_generic", KERNEL=="sg[0-9]", ATTRS{manufacturer}=="Linux Automation GmbH", ATTRS{product}=="usb-sd-mux*", OWNER="_openqa-worker"
ACTION=="add", SUBSYSTEM=="block", KERNEL=="sd*", ATTRS{manufacturer}=="Linux Automation GmbH", ATTRS{product}=="usb-sd-mux*", OWNER="_openqa-worker"
```

Also a recent (>=0.2.1) version of `python3-usbsdmux` is required.

### DIFF
--- a/data/generalhw_scripts/flash_sd_rootless.sh
+++ b/data/generalhw_scripts/flash_sd_rootless.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# Script to be used with: 'usbsdmux' software, with the following hardware:
+#  USB-SD-mux from https://shop.linux-automation.com/index.php?route=product/product&product_id=50
+
+# Print commands and abort script on failure
+set -ex
+
+echo "Flash script start...";
+
+# Check number of args
+if [ "$#" -ne 4 ]; then
+    echo "Please provide <destination>, <device serial>, <image to flash> and <hdd size> (ignored)."
+    exit 1;
+fi
+
+# Get destination: /storage/
+destination_folder=$1
+# Get device serial (check 'sd-mux-ctrl --list', or usb-sd-mux/id-<SERIAL>)
+device_serial=$2
+# Get image to flash
+image_to_flash=$3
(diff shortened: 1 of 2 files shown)


ggardet:
This look good to me, but I think the udev rules should be packaged properly. Maybe in a `os-autoinst` sub package?

asdil12:
> This look good to me, but I think the udev rules should be packaged properly. Maybe in a `os-autoinst` sub package? Or in a `usbsdmux` subpackage?

It wouldn't really fit for the `usbsdmux` package as it would be openQA specific.
And I'm not sure about `os-autoinst`.
There is so much manual config required anyway, so I will simply add a documentation for the worker setup for rpi hw tests:

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14779/commits/dc081593317ce9236215f282126f933314dd6d1a?short_path=38451fb#diff-38451fb0e86ffc9226ef27aac4e81b72d6ccd7ce14daf958cc5bb8acae362259

ggardet:
Regarding the docs, could you explicitly write that your setup is for _osd_, not _o3_, please? Because the current o3 setup is slightly different.

asdil12:
> Regarding the docs, could you explicitly write that your setup is for _osd_, not _o3_, please? Because the current o3 setup is slightly different.

What are the differences?

- different NFS share (or do you pull the stuff using rsync?)
- no wifi or bluetooth setup
- no hw serial connection
- use of the older `flash_sd.sh` script

Is that everything?

PR #14780, opened 2022-04-25
publiccloud : limit running of registercloudguest only to console tests

asmorodskyi:
VRs : 
from : http://autobot.qa.suse.de/tests/3183
till :  http://autobot.qa.suse.de/tests/3192
Note : only failing test is http://autobot.qa.suse.de/tests/3187 and it is perfect catch of https://bugzilla.suse.com/show_bug.cgi?id=1198815

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -209,60 +209,6 @@ sub wait_for_guestregister
     die('guestregister didn\'t end in expected timeout=' . $args{timeout});
 }
 
-=head2 check_guestregister
-
-    check_guestregister();
-
-Check that the registration state of newly created guest match our requirements.
-https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt
-=cut
-sub check_guestregister {
-    my ($self) = @_;
-
-    if (is_chost()) {
-        record_info('SKIP CHECK', 'Not checking guestregister as it is not installed on CHOST images');
-        return 0;
-    }
-
-    if (is_byos) {
-        if ($self->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
-            die 'The BYOS instance should be unregistered and report "Warning: No repositories defined.".';
-        }
-
-        if ($self->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
-            die('guestregister.service is not disabled');
-        }
-
-        if ($self->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
-            die('/etc/zypp/credentials.d/ is not empty');
-        }
-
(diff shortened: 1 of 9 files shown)


lib/publiccloud/instance.pm
```diff
@@ -209,55 +209,6 @@ sub wait_for_guestregister
...
 }
 
-=head2 check_guestregister
-
-    check_guestregister();
-
-Check that the registration state of newly created guest match our requirements.
-https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/integration_test-process.txt
-=cut
-sub check_guestregister {
```
pdostal, line 219: Please also remove it from exports.

asmorodskyi:
first VR: http://autobot.qa.suse.de/tests/3153 but I will run way more before merge ...

lib/publiccloud/instance.pm
pdostal: Ah yes, sorry

tests/publiccloud/registercloudguest.pm
```diff
@@ -75,4 +119,11 @@ sub post_fail_hook {
     $self->SUPER::post_fail_hook;
 }
 
+sub test_flags {
+    return {
+        fatal => 0,
+        milestone => 0
+    };
```
pdostal, line 126:
```suggestion
    return { fatal => 0 };
```
I think the `milestone` is `0` by default. See [documentation](https://open.qa/docs/#_how_to_write_tests).

pdostal (approved):
Thank you Anton, this looks nice 🚀

lib/publiccloud/utils.pm
```diff
@@ -172,4 +184,8 @@ sub is_gce() {
     return is_public_cloud && check_var('PUBLIC_CLOUD_PROVIDER', 'GCE');
 }
 
+sub is_chost() {
```
grisu48, line 187: I would recommend to use the full name instead of the abbreviation here to reduce the number of potentially unknown abbreviations

```suggestion
sub is_container_host() {
```

Pro: Learns new members and other people what `chost` means  
Con: Longer to write

WDYT?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -31,9 +32,53 @@ sub run {
 
     select_host_console();    # select console on the host, not the PC instance
 
+    registercloudguest($instance) if is_chost();
+
+    if (is_byos()) {
+        # We need to skip all BYOS checks for CHOST BYOS images because of custom registration we doing above.
```
grisu48, line 38: Please also note that CHOST images come only as BYOS, e.g.

```suggestion
        # Container-Host images (CHOST) come only as BYOS - Here We skip all checks because of a custom registration procedure (see above)
```

grisu48 (approved):
LGTM! Two style suggestions, feel free to pick the ones you agree with.

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: this is very important point which will change NOT only comment . thanks Felix :+1:

PR #14782, opened 2022-04-25
WIP: Add wait_still_screen before product selection in welcome.pm

rakoenig:
- in an ssh-X scenario on ppc64le the assert_and_click
  won't work if there are still popus aroud

- Related ticket: https://progress.opensuse.org/issues/106514
- Needles: -
- Verification run: https://openqa.suse.de/tests/8630910

### DIFF
--- a/tests/installation/welcome.pm
+++ b/tests/installation/welcome.pm
@@ -193,6 +193,14 @@ sub run {
         }
         else {
             assert_and_click('before-select-product-' . $product);
+            if (!check_screen('select-product-' . $product)) {
+                if (check_var('VIDEOMODE', 'ssh-x') && is_ppc64le) {
+                    # on ssh-X we need to wait for some popups to go away on ppc64le
+                    record_info('Note:', 'X-Forwarding via SSH is very slow on ppc64le. Retrying selection.');
+                    wait_still_screen(stilltime => 20, timeout => 100);
+                    assert_and_click('before-select-product-' . $product);
+                }
+            }
         }
         assert_screen('select-product-' . $product);
     }


okurz:
I suggest to solve this with proper needles instead of waiting such long time

rakoenig:
> I suggest to solve this with proper needles instead of waiting such long time

Oliver, this is not a needle problem. The problem is that the needle tagged with "welcome-before-product-$product" perfectly matches. Its an [assert_and_click](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f1daa33268b4e7870af27de579b6322b7e87f80b/tests/installation/welcome.pm#L195) instruction.

So the needle matches and then we produce a mouse click. Unfortunately YaST is still showing popups with "Downloading..." some assets. So the assert_**and_click** now does not click on the radio button in the product selection but rather klicks on the popup. Which has no effect on the product selection, so the next needle that checks if the product is really selected fails. 

This happens only in ssh-x scenarios, other seem to be fast enough that we don't suffer from this "race condition". That's why I wrapped the wait_still_screen with so that it only gets executed when VIDEOMODE is 'ssh-x'.

okurz:
> > I suggest to solve this with proper needles instead of waiting such long time
> 
> Oliver, this is not a needle problem. The problem is that the needle tagged with "welcome-before-product-$product" perfectly matches. Its an [assert_and_click](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f1daa33268b4e7870af27de579b6322b7e87f80b/tests/installation/welcome.pm#L195) instruction.
> 
> So the needle matches and then we produce a mouse click. Unfortunately YaST is still showing popups with "Downloading..." some assets. So the assert_**and_click** now does not click on the radio button in the product selection but rather klicks on the popup. Which has no effect on the product selection, so the next needle that checks if the product is really selected fails.

Right, so the existing needles are not strict enough. Because they shouldn't match as long as a popup is shown. Or rather a needle that should detect the popups and handle accordingly. Think of this how you as human operator would do it: If you see the screen that you want to reach but there is a popup in front, you wouldn't just wait another minute before doing something but you would closely read what is on the screen "is it just the welcome screen or is it the welcome screen hidden by a popup" and act accordingly. And this is how openQA tests should handle it just as well.

> This happens only in ssh-x scenarios, other seem to be fast enough that we don't suffer from this "race condition". That's why I wrapped the wait_still_screen with so that it only gets executed when VIDEOMODE is 'ssh-x'.

Yes, but it's still a race condition and race conditions can not be avoided by sleeping, you can just better hide errors but it could still be that in some other scenarios tests fail, just more seldomly, and it might even still fail in ssh-x if something is really slow.

You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

rakoenig:
> Right, so the existing needles are not strict enough. Because they shouldn't match as long as a popup is shown.

Nope. The problem is that "assert_and_click" is not atomic. The screen matches because no popup is shown. The assert part is satisfied and proceeds to the "and_click". Meanwhile because we're on a slow system a popup appears. So the click is performed, but hits the popup window.

> Think of this how you as human operator would do it

Yes, as a human operator I would also think in the first moment "Yeah, screen is here, so let me click on the radio button". Then "Oops there is a popup". Lets wait. And then I see other popus and since I have no idea how many popups there will be my human approach would be "let me wait until no mor popups show up". Isn't that what wait_still_screen is supposed to do? 

Of course I could also file a bug against YaST, because even when using the YuiRestClient (tried that in another setup) I can send "press UI widget" events to the server and get a "200 OK" response while YaST is throwing away that event. And probably everyone will say, no bug, just wait until there are no popups. ;-)

> It is important to know the exact condition that the code should wait for and check for that instead of delaying program execution for an arbitrary amount of time.

Ok, the exact condition is "no more popups popping up". How are you detecting this when there is no popup that says "hey, popups are done now". For this problem I see 2 approaches:

- I wait for a reasonable amount of time and if no more "noise" is seen on the screen I assume that I can proceed.
- I could also try a brute-force "proceed", so lets do the assert_and_click as it is coded in the test. And then if the next needle that checks if my product is selected fails (as it does because I cklicked on the popup) then I say "oops, cklick didn't hit the radio button, lets try again". And how many iterations shall I do for that? And keep in mind that maybe even sending the key "$cmd{next}" might not work as long as popups are around. 

For me the first solution looks more reasonable than trying the brute force approach. And I guess the code would be also  easier to understand than  the other approach.

jknphy:
I confirmed the behaviour with YaST developers and I got:
`popups are more visible in this case because everything is slower` and `The Online medium contains just an empty repository, the repository files are very small, normally it takes just a fraction of second so the popup is not visible.`
So basically we have pairs of popups 'scanning product ...xyz' and 'downloading ...xyz' which in a normal installation looks like one and in a slower installation are visible one by one.

In other words, it is unpredictable in this case, so `wait_still_screen` it is probably the best we can use, because we don't even know the number of popups that will appear :)

On the other hand we should avoid to waste time when this bad design doesn't hit us in a slower environment (yes it is a bad design, developers also are trying to move to a better UI that doesn't abuse of popups).
There is a service request (still unreplied) and a bug for this particular slow environment (which is not really a slow arch, it is due to a network problem) https://sd.suse.com/servicedesk/customer/portal/1/SD-84371 https://bugzilla.suse.com/show_bug.cgi?id=1178172#c15

What we could simply do here is narrow the issue even more an do a softfailure, we can assume that any ssh-x scenario run in OSD in powerVM should be affected (if there is any other, because it is a very specific test case).
@rakoenig could you please add the soft-failure pointing to a new progress ticket which we will use exclusively to share the info from the service request, not the one you are working on, because in this case I don't think it makes much sense to use the bug, also it could be an option, or point to both) and narrow the condition in the code to only wait in PowerVM. Otherwise with the current code the infra problem is not visible and we would wait unnecessary time for ssh-x scenario in s390x as Oliver pointed.

I added to the bug the infra ticket for cross reference when someone would look at the soft-failure.

okurz:
The above sounds like a sensible solution. I have seen many well designed applications that make it unambiguously clear when things are still processed in the background. In such cases one would be able to see for example from a message on the screen "processing …" or a different text in the buttons if there is a possibility of popups still appearing at a later time.
I would also be ok to use `wait_still_screen` with a clear indication that this is a workaround for a bug (or such bad design choice) in the software when one is in the ambiguous situation that it is not clear if popups can still happen. Yet another solution could be to not use `assert_and_click` – yes, it's not atomic – and instead use a loop using `assert_screen`, `match_has_tag` and `click_lastmatch` with multiple retries.

tests/installation/welcome.pm
```diff
@@ -192,6 +192,11 @@ sub run {
             send_key 'alt-' . $hotkey{$product};
         }
         else {
+            if (check_var('VIDEOMODE', 'ssh-x') && is_ppc64le) {
+                # on ssh-X we need to wait for some popups to go away on ppc64le
+                record_soft_failure('poo#110470: X-Forwarding via SSH is very slow on ppc64le');
```
okurz, line 197: This unfortunately records a soft failure without checking for the actual error condition. Also I missed that you wanted to point to a progress ticket, not a bug in bugzilla, which should be avoided. Please only reference product issues with soft failure

tests/installation/welcome.pm
```diff
@@ -193,6 +193,14 @@ sub run {
         }
         else {
             assert_and_click('before-select-product-' . $product);
+            if (!check_screen('select-product-' . $product)) {
+                if (check_var('VIDEOMODE', 'ssh-x') && is_ppc64le) {
```
okurz, line 197: I don't think this can work. `check_screen` by default has a 0s timeout so this would only match in cases the screen is already showing the expected screen. Why do you need two nested if's and not just one statement

```suggestion
            if (!check_screen('select-product-' . $product) && check_var('VIDEOMODE', 'ssh-x') && is_ppc64le) {
```

?

This wouldn't solve the above problem though. In that case you could still add a reasonable timeout to the `check_screen` code

@coolo do you have a good idea how in this case we could avoid `check_screen` and `wait_still_screen` altogether but still work around the badly designed YaST installer that can at any time interfer with a popup that is preventing the previous click on the product selection to be properly caught?

okurz (changes requested):
Next to my inline comment there are also typos in the git commit "popus aroud"

jknphy:
Hi @okurz I already agreed with @rakoenig to stop working on this. Probably he was just trying some code and forgot to put the WIP, we plan to use your feedback and apply it with libyui REST. According the previous comment, those pairs of popup are not predicable so we cannot check for them leading us to the only apparent solution which is to retry as you mentioned. The retry mechanism it is much more straightforward to implement with REST calls than with needles, and more maintenable.
Anyway this strange effect of popup is only visible as result of the network issue, so even passing this screen, there are other screens where there are also problems, so we don't gain much here moving forward.
@rakoenig feel free to close the PR so it is clear that we don't touch welcome.pm unnecesarily.

tests/installation/welcome.pm
coolo: Well, as you already wrote: without the app behaving (showing its progress properly) I sense this is what a human would do too: wait until it stops acting strangely - and a little more.

But file a bug about it and let's hope for the best. That won't stop the QA code from working around it until we stop installing old products.

rakoenig:
I'm going to close this PR now as @jknphy suggested and I will encounter new obstacles by trying to implement this with libyui. ;-)

PR #14785, opened 2022-04-26
Improve containers::common::activate_containers_module subroutine

pdostal:
Parse the `SUSEConnect -s` output and check for the containers module status.

- Related ticket: https://progress.opensuse.org/issues/110331
- Related issue: https://openqa.suse.de/tests/8634857
- Verification run: [SLE12-SP3](http://pdostal-server.suse.cz/tests/14534), [SLE15-SP3-GCE-BYOS](http://pdostal-server.suse.cz/tests/14532)

### DIFF
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -13,6 +13,7 @@ use version_utils;
 use utils qw(zypper_call systemctl file_content_replace script_retry script_output_retry);
 use containers::utils qw(can_build_sle_base registry_url container_ip container_route);
 use transactional qw(trup_call check_reboot_changes);
+use Mojo::JSON;
 
 our @EXPORT = qw(is_unreleased_sle install_podman_when_needed install_docker_when_needed install_containerd_when_needed
   test_container_runtime test_container_image scc_apply_docker_image_credentials scc_restore_docker_image_credentials
@@ -25,13 +26,18 @@ sub is_unreleased_sle {
 }
 
 sub activate_containers_module {
-    my ($running_version, $sp, $host_distri) = get_os_release;
-    my $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
-    if ($suseconnect !~ m/Containers/) {
-        $running_version eq '12' ? add_suseconnect_product("sle-module-containers", 12) : add_suseconnect_product("sle-module-containers");
-        $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
+    my $json = Mojo::JSON::decode_json(script_output_retry('SUSEConnect -s', timeout => 240, retry => 3, delay => 60));
+    my ($not_registered, $version, $arch);
+    foreach (@$json) {
+        if ($_->{identifier} =~ 'sle-module-containers' && $_->{status} =~ 'Not Registered') {
+            $not_registered = 1;
+            $version = $_->{version};
+            $arch = $_->{arch};
+            last;
+        }
     }
-    record_info('SUSEConnect', $suseconnect);
+    add_suseconnect_product("sle-module-containers", $version, $arch) if ($not_registered);
+    record_info('SUSEConnect', script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60));
 }
 
 


lib/containers/common.pm
```diff
@@ -25,13 +26,14 @@ sub is_unreleased_sle {
 }
 
 sub activate_containers_module {
-    my ($running_version, $sp, $host_distri) = get_os_release;
-    my $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
-    if ($suseconnect !~ m/Containers/) {
-        $running_version eq '12' ? add_suseconnect_product("sle-module-containers", 12) : add_suseconnect_product("sle-module-containers");
-        $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
+    my $running_version = shift;
```
grisu48, line 29: This change will break our test runs. e.g. for https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/containers/common.pm#L48 where the argument is not passed.

lib/containers/common.pm
```diff
@@ -25,13 +26,14 @@ sub is_unreleased_sle {
...
+    my $json = Mojo::JSON::decode_json(script_output_retry('SUSEConnect -s', timeout => 240, retry => 3, delay => 60));
+    my $registered = 0;    # We expect that containers module is not registered
+    foreach (@$json) {
+        $registered = 1 if ($_->{identifier} =~ 'container' && $_->{status} =~ 'Registered');
     }
-    record_info('SUSEConnect', $suseconnect);
+    ($registered && $running_version eq '12') ? add_suseconnect_product("sle-module-containers", 12) : add_suseconnect_product("sle-module-containers");
```
grisu48, line 35: Shouldn't it be
```suggestion
    (!($registered) && $running_version eq '12') ? add_suseconnect_product("sle-module-containers", 12) : add_suseconnect_product("sle-module-containers");
```
In words: We add the container product only, if it is not yet registered, right?

grisu48 (changes requested):
I believe one of the parts would break out test runs. Please have a look at my comments

lib/containers/common.pm
jlausuch: yes, look at https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/containers/host_configuration.pm#L23
where this test is executed on 3rd party hosts as well.

I would keep same arguments to avoid regressions, and just adapt the code of the suseconnect output parsing.

pdostal: Please re-check.

lib/containers/common.pm
pdostal: I changed the variable name to `$not_registered`.

lib/containers/common.pm
```diff
@@ -25,13 +26,17 @@ sub is_unreleased_sle {
...
 sub activate_containers_module {
-    my ($running_version, $sp, $host_distri) = get_os_release;
-    my $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
-    if ($suseconnect !~ m/Containers/) {
-        $running_version eq '12' ? add_suseconnect_product("sle-module-containers", 12) : add_suseconnect_product("sle-module-containers");
-        $suseconnect = script_output_retry('SUSEConnect --status-text', timeout => 240, retry => 3, delay => 60);
+    my $json = Mojo::JSON::decode_json(script_output_retry('SUSEConnect -s', timeout => 240, retry => 3, delay => 60));
+    my ($not_registered, $version, $arch);
+    foreach (@$json) {
+        if ($_->{identifier} =~ 'container' && $_->{status} =~ 'Not Registered') {
```
grisu48, line 32: I'm nitpicking but IMHO it's better to do two things here:

1. Check for the explicit module name `sle-module-containers`, same as in line 38 where it is also used explicitly
2. Instead of checking for "Not Registered", check for not "Registered" - this check is broader as it would recognise any new status that is not explicitly "Registered" e.g. "Expired"
```suggestion
        if ($_->{identifier} =~ 'sle-module-containers' && $_->{status} !~ 'Registered') {
```

lib/containers/common.pm
```diff
@@ -25,13 +26,17 @@ sub is_unreleased_sle {
...
+            $not_registered = 1;
+            $version = $_->{version};
+            $arch = $_->{arch};
+        }
```
grisu48, line 36: I would add a `last` here, since at this point there is no need to further process the other json elements.

grisu48 (approved):
LGTM now, however I still have two suggestions to improve. Feel free to take the ones you like, but it should already work like this.

lib/containers/common.pm
asmorodskyi: > Check for the explicit module name sle-module-containers, same as in line 38 where it is also used explicitly

+1 

> Instead of checking for "Not Registered", check for not "Registered" - this check is broader as it would recognise any new status that is not explicitly "Registered" e.g. "Expired"

not sure here , if we will grep just for `Registered` , `Not Registered` will also match right ?

jlausuch (approved):
This looks very nice!

PR #14797, opened 2022-04-27
publiccloud: Hide sensitive information from all kind of logs in openQA

asmorodskyi:


foursixnine:
@asmorodskyi do you have a ticket for this? So I can link it to another?

asmorodskyi:
> @asmorodskyi do you have a ticket for this? So I can link it to another?

https://progress.opensuse.org/issues/108506

lib/publiccloud/utils.pm
```diff
@@ -172,4 +172,12 @@ sub is_gce() {
     return is_public_cloud && check_var('PUBLIC_CLOUD_PROVIDER', 'GCE');
 }
 
+sub define_secret_variable {
```
asmorodskyi, line 175: @foursixnine  this function might be useful for broader audience , I wonder if it make sense to move it into `lib/utils` or maybe some other place which you have in your mind ...

foursixnine: @asmorodskyi I think that it could be moved to it's own namespace (say `Utils::Security::Commands` not to pollute `lib/utils`...  but that would be for later in the future, so I guess for now here is ok, once I find a place good enough, I'll come back to this.

Pinging @punkioudi and @michaelgrifalconi as they are working on the leaking story atm

ilausuch (approved):

lib/publiccloud/utils.pm
```diff
@@ -188,4 +202,12 @@ sub is_container_host() {
     return is_public_cloud && get_var('FLAVOR') =~ 'CHOST';
 }
 
+sub define_secret_variable {
+    my ($var_name, $var_value) = @_;
+    script_run("set -a");
+    script_run("read -sp \"enter value: \" $var_name", 0);
+    type_password($var_value . "\n");
```
pdostal, line 209: This is smart 😆 👍

asmorodskyi: surprisingly credits goes to MS documentation :) I was just smart enough to find this and read :)

PR #14798, opened 2022-04-27
Remove deprecated code

punkioudi:
There is a part dedicated in an older version of libreoffice, which is
not anymore applicable as the testsuite runs only in sle 15-sp3 with newer
version of libreoffice.

- Related ticket: https://progress.opensuse.org/issues/110356
- Needles: N/A
- Verification run: [sle15sp3](https://openqa.suse.de/tests/8647285#step/libreoffice_pyuno_bridge_no_evolution_dep/14)


foursixnine (approved):
I ❤️ Pyuno

PR #14800, opened 2022-04-28
Deactivate python2 module when failure happened

lemon-suse:
We need cleanup environment to deactivate pyhonn2 when issue happened in case it will block later test.

- Related ticket: https://progress.opensuse.org/issues/110302
- Needles: N/A
- Verification run: http://openqa.nue.suse.com/tests/8645523#step/yast2_migration/14


coolgw (approved):

tinawang123 (approved):
LGTM

PR #14809, opened 2022-04-29
xfstests: Add raw dump function via dd

lansuse:
xfstests: Add raw dump function via dd

In cases where the primary fs superblock is totally corrupted(such as xfs_metadump not work)
we want to have a raw dump from the device

- Related ticket: https://progress.opensuse.org/issues/56243
- Needles: N/A
- Verification run: 
http://10.67.133.102/tests/642#step/run/154 (btrfs)
http://10.67.133.102/tests/641#step/run/64 (xfs)

### DIFF
--- a/tests/xfstests/run.pm
+++ b/tests/xfstests/run.pm
@@ -324,6 +324,13 @@ sub dump_btrfs_img {
     script_run($cmd);
 }
 
+# Log: Raw dump from SCRATCH_DEV via dd
+sub raw_dump {
+    my ($category, $num) = @_;
+    my $dev = get_var('XFSTESTS_SCRATCH_DEV') ? get_var('XFSTESTS_SCRATCH_DEV') : (split(/ /, get_var("XFSTESTS_SCRATCH_DEV_POOL")))[0];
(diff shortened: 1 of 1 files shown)


tests/xfstests/run.pm
```diff
@@ -451,6 +458,7 @@ sub run {
                 collect_fs_status($category, $num);
                 if (get_var('BTRFS_DUMP', 0) && (check_var 'XFSTESTS', 'btrfs')) { dump_btrfs_img($category, $num); }
             }
+            if (get_var('RAW_DUMP', 0)) { raw_dump($category, $num); }
```
SeroSun, line 461: I recommend this line move to the if structure when`$status =~ /FAILED/`. Only rawdump when test fail, I see all pass tests has this rawdump.

SeroSun (commented):
512kb looks enough for rawdump, and I suggest only fail case has this rawdump, commands in line

tests/xfstests/run.pm
SeroSun: Considering this has parameter to control it's OK to leave this out of if structure.

SeroSun:
LGTM

PR #14810, opened 2022-04-29
Snapper can be running before mock function executes

mloviska:
- Related ticket: [[sporadic] test fails in snapper_cleanup - snapper cleanup number - Config is in use.](https://progress.opensuse.org/issues/103299)
- Verification run: [sle-12-SP5-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220428-1-jeos-filesystem@64bit-virtio-vga](http://kepler.suse.cz/tests/16749#step/snapper_create/8)

### DIFF
--- a/lib/btrfs_test.pm
+++ b/lib/btrfs_test.pm
@@ -88,6 +88,14 @@ By updating the lastrun files timestamps, we make sure those routines won't be e
 while tests are running. 
 =cut
 sub cron_mock_lastrun {
+    my $tries = 5;
+    while (script_run(q/ps aux | grep '[s]napper'/) == 0 && $tries > 0) {
+        sleep 30;
+        bmwqemu::diag('Snapper is running in the background...');
+        $tries--;
+    }
+    $tries or bmwqemu::diag('Snapper might be still running in the background');
+
     assert_script_run 'touch /var/spool/cron/lastrun/cron.{hourly,daily,weekly,monthly}';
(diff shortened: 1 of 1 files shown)


lib/btrfs_test.pm
```diff
@@ -88,6 +88,14 @@ By updating the lastrun files timestamps, we make sure those routines won't be e
 while tests are running. 
 =cut
 sub cron_mock_lastrun {
+    my $tries = 5;
+    while (script_run(q/ps aux | grep '[s]napper'/) == 0 && $tries > 0) {
+        sleep 30;
+        bmwqemu::diag('Snapper is running in the background...');
+        $tries--;
+    }
+    $tries or bmwqemu::diag('Snapper might be still running in the background');
```
jlausuch, line 97: This is a bit weird. Let's say, the first iteration the snapper is not running, so `tries=5` and you are still showing the message. You only hide this message in the case we have the 5 iterations in the loop where `tries=0`...

mloviska: It is vice versa. When `$tries` evaluates as falsy the message appears. 
Some something like
```
bmwqemu::diag('Snapper might be still running in the background') unless $tries;
```

jlausuch: you are right, I shouldn't do reviews on Saturday :D

jlausuch (approved):

PR #14813, opened 2022-05-02
BCI: adaptations to run 15-SP4 images

jlausuch:
According to https://github.com/SUSE/BCI-tests/blob/main/bci_tester/data.py#L54
we need to provide these variables to make sure we are testing the right images. 
Currently, `15.3` (15-SP3) by default, but this is needed to enable 15-SP4 images.

- Related ticket: https://progress.opensuse.org/issues/110304
- VR: https://openqa.suse.de/tests/8672500#


pdostal (approved):

dcermak (approved):
LGTM!

PR #14815, opened 2022-05-03
JeOS: rework condition for pushing data to InfluxDB

jlausuch:
CASEDIR variable is always set, so this condition was always
returning and never executing the code afterwards.

OSD -> `CASEDIR=sle`
O3 -> `CASEDIR=opensuse`

- Related tickets: 
    - https://jira.suse.com/browse/TEAM-5808
    - https://progress.opensuse.org/issues/110224

### DIFF
--- a/tests/jeos/image_info.pm
+++ b/tests/jeos/image_info.pm
@@ -76,7 +76,9 @@ sub run {
     my $influxdb_server = get_var('INFLUXDB_SERVER');
     my $influxdb_user = get_var('_SECRET_INFLUXDB_USER');
     my $influxdb_pwd = get_var('_SECRET_INFLUXDB_PWD');
-    return if (get_var('CASEDIR') || !($influxdb_server || $influxdb_user || $influxdb_pwd));
+
+    return if (get_var('CASEDIR') !~ m/^sle$|^opensuse$/);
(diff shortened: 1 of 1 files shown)


tests/jeos/image_info.pm
```diff
@@ -76,7 +76,9 @@ sub run {
     my $influxdb_server = get_var('INFLUXDB_SERVER');
     my $influxdb_user = get_var('_SECRET_INFLUXDB_USER');
     my $influxdb_pwd = get_var('_SECRET_INFLUXDB_PWD');
-    return if (get_var('CASEDIR') || !($influxdb_server || $influxdb_user || $influxdb_pwd));
+
+    return if (get_var('CASEDIR') =~ m/^sle$|^opensuse$/);
```
mloviska, line 80: Shouldn't the code continue as long as `CASEDIR=sle|opensuse`?

jlausuch: damn, yes!

jlausuch: the problem is that I can't run VRs :D

mloviska: :D let's do a VR on site. If there will be an issue, another PR would fix it

PR #14831, opened 2022-05-04
Migrate ssh-X testsuite to libyui

rakoenig:
- This commit migrates the ssh-X testsuite from a needle based approach
  to libyui. So we don't care if YaST windows are fullscreen or not.

Note that the VR for ppc64le still fails because when libyui sends the "press next" event in install_SLES there are still popups showing up sporadically on the screen, so this event won't arrive at YaST, even if the YUIRestServer sends an "200 OK" response back. Then in the next step accept license will complain that the required page will not appear. 

- Related ticket: https://progress.opensuse.org/issues/106514
- Needles: 
- Verification run: [Verification runs](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=rakoenig_fix_ensure_fullscreen&groupid=96)

### DIFF
--- a/schedule/yast/ssh-x.yaml
+++ b/schedule/yast/ssh-x.yaml
@@ -3,30 +3,38 @@ description:    >
   Conduct an installation using ssh with X-Forwarding.
   Might only be effective for zVM and powerVM
 vars:
+  YUI_REST_API: 1
   DESKTOP: textmode
-  PATTERNS: minimal,base
   VIDEOMODE: ssh-x
 schedule:
   - installation/bootloader_start
-  - installation/welcome
-  - installation/accept_license
-  # Required on zVM
+  - installation/setup_libyui
+  - installation/access_beta_distribution
+  - installation/ensure_installer_fullscreen
+  - installation/product_selection/install_SLES
+  - installation/licensing/accept_license
+    # Required on zVM
   - '{{disk_activation}}'
-  - installation/scc_registration
-  - installation/addon_products_sle
-  - installation/system_role
-  - installation/partitioning
-  - installation/partitioning_finish
-  - installation/installer_timezone
-  - installation/user_settings
-  - installation/user_settings_root
-  - installation/select_patterns
-  - installation/installation_overview
-  - installation/disable_grub_timeout
-  - installation/start_install
-  - installation/await_install
+  - installation/registration/register_via_scc
+  - installation/module_registration/skip_module_registration
+  - installation/add_on_product/skip_install_addons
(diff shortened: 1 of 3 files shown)


jknphy:
Could you please create a test module to perform the full screen action to be scheduled before `install_SLES`. As a user I would prefer to see full screen, in theory it should work in s390x after the fix you did, right?

rakoenig:
> Could you please create a test module to perform the full screen action to be scheduled before `install_SLES`. As a user I would prefer to see full screen, in theory it should work in s390x after the fix you did, right?

Made a module to go fullscreen and restarted the VR on S390, seems to work.

schedule/yast/ssh-x.yaml
```diff
@@ -3,30 +3,39 @@ description:    >
   Conduct an installation using ssh with X-Forwarding.
   Might only be effective for zVM and powerVM
 vars:
+  YUI_REST_API: 1
   DESKTOP: textmode
   PATTERNS: minimal,base
```
jknphy, line 8: this setting is not needed.

tests/installation/yast_fullscreen.pm
```diff
@@ -0,0 +1,20 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Enlarge the YaST window for fullscreen in ssh-X test.
```
jknphy, line 6: let's name it `ensure_installer_fullscreen` wdyt? similar than the method name, this functionality is not very YaSTish' IMO :) . In the future someone could pass specific string to search for screens. Also it is not clear which yast screen in particular as there are popups and more windows.

tests/installation/yast_fullscreen.pm
```diff
@@ -0,0 +1,20 @@
...
+
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use testapi;
```
jknphy, line 13: not needed.

jknphy (commented):
LGTM besides those minor things, let's remove the WIP.
good work in general reviving the ssh-x scenario!

tests/installation/yast_fullscreen.pm
rakoenig: Tell this to the CI. ;-) 
Started without this and got complaints...

jknphy: mmm perhaps the difference between `use parent` and `use base`, strange, you're right.

PR #14853, opened 2022-05-09
Public Cloud: Remove YAML schedule (excluding HA_SAP)

pdostal:
We don't use `schedule/` any more as we have `lib/main_publiccloud.pm`. The only tests where `YAML_SCHEDULE` variable is used are HA-SAP tests.
 - **Merge with: [qac/qac-openqa-yaml!794](https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/794)** 
- Verification run: [upload_img](http://pdostal-server.suse.cz/tests/14704), [smoketests](http://pdostal-server.suse.cz/tests/14703)


grisu48 (approved):
LGTM! After merging we need to keep a close eye on our test runs however. Issues with scheduling are almost always present.

Needs to be merged after https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/794

PR #14856, opened 2022-05-09
publiccloud : add registration in post_fail_hook to increase chances that system is actually registered

asmorodskyi:
At the end of check_registercloudguest test module we expect the SUT to be registered.
If this test module fails the system may stay unregistered which breaks the following modules.

I suggest to force-register the system in post_fail_hook subroutine of this test module so the test run can continue.


pdostal (approved):
👌

asmorodskyi:
VR: http://autobot.qa.suse.de/tests/3216

PR #14864, opened 2022-05-10
Ignore to install gun9/10/11 packages refer bsc#1194917

tinawang123:
Ignore to install gun9/10/11 packages refer bsc#1194917. Because there is an error which was introduced thru a maintenance update when versions of gnu-compilers-hpc for older gcc versions were erroneously released for SLE-15-SP3.

- Related ticket: https://progress.opensuse.org/issues/110154
- Verification run: https://openqa.suse.de/tests/8702813#step/install_service/19

### DIFF
--- a/lib/services/hpcpackage_remain.pm
+++ b/lib/services/hpcpackage_remain.pm
@@ -37,7 +37,8 @@ sub list_pkg {
 sub install_pkg {
     my $version = get_var('HDDVERSION');
     # Install all 'library wrappers' (packages matching -hpc, not having a version number with '_' after their name)
-    my @pkginstall = split('\n', script_output q[zypper search -r SLE-Module-HPC] . $version . q[-Pool -r SLE-Module-HPC] . $version . q[-Updates | cut -d '|' -f 2 | sed -e 's/ *//g' | grep -E '.*-hpc.*' | grep -vE 'system|module|suse' |  grep -vE '.*_[[:digit:]]+_[[:digit:]]+.*gnu|.*_[[:digit:]]+_[[:digit:]]+.*-hpc' | grep -vE '.*-static$' | grep -vE '.*hpc-macros.*'], proceed_on_failure => 1, timeout => 180);
+    record_soft_failure('bsc#1194917', "openQA test fails : nothing provides 'gcc9' needed by the to be installed gnu9-compilers-hpc-devel-1.4-3.14.3.noarch");
+    my @pkginstall = split('\n', script_output q[zypper search -r SLE-Module-HPC] . $version . q[-Pool -r SLE-Module-HPC] . $version . q[-Updates | cut -d '|' -f 2 | sed -e 's/ *//g' | grep -E '.*-hpc.*' | grep -vE 'system|module|suse' | grep -vE 'gnu9|gnu10|gnu11' |  grep -vE '.*_[[:digit:]]+_[[:digit:]]+.*gnu|.*_[[:digit:]]+_[[:digit:]]+.*-hpc' | grep -vE '.*-static$' | grep -vE '.*hpc-macros.*'], proceed_on_failure => 1, timeout => 180);
     # on x86 zypper will print out the Shell debug information, we need exclude it.
     @pkginstall = grep { !/LMOD_SH_DBG_ON/ } @pkginstall;
     zypper_call("in " . join(' ', @pkginstall), timeout => 1800);


lemon-suse (approved):

lemon-suse:
LGTM, just a minor comments, update the PR title to 'Ignore to install gun9/10/11-* packages refer bsc#1194917', same for the  PR description and git commit title and description.

lemon-suse:
> LGTM, just a minor comments, update the PR title to 'Ignore to install gun9/10/11-* packages refer bsc#1194917', same for the PR description and git commit title and description.

Besides, if you can write the reason in description that why we need to do this will be better.

lib/services/hpcpackage_remain.pm
```diff
@@ -37,7 +37,7 @@ sub list_pkg {
 sub install_pkg {
     my $version = get_var('HDDVERSION');
     # Install all 'library wrappers' (packages matching -hpc, not having a version number with '_' after their name)
-    my @pkginstall = split('\n', script_output q[zypper search -r SLE-Module-HPC] . $version . q[-Pool -r SLE-Module-HPC] . $version . q[-Updates | cut -d '|' -f 2 | sed -e 's/ *//g' | grep -E '.*-hpc.*' | grep -vE 'system|module|suse' |  grep -vE '.*_[[:digit:]]+_[[:digit:]]+.*gnu|.*_[[:digit:]]+_[[:digit:]]+.*-hpc' | grep -vE '.*-static$' | grep -vE '.*hpc-macros.*'], proceed_on_failure => 1, timeout => 180);
+    my @pkginstall = split('\n', script_output q[zypper search -r SLE-Module-HPC] . $version . q[-Pool -r SLE-Module-HPC] . $version . q[-Updates | cut -d '|' -f 2 | sed -e 's/ *//g' | grep -E '.*-hpc.*' | grep -vE 'system|module|suse' | grep -vE 'gnu9|gnu10|gnu11' |  grep -vE '.*_[[:digit:]]+_[[:digit:]]+.*gnu|.*_[[:digit:]]+_[[:digit:]]+.*-hpc' | grep -vE '.*-static$' | grep -vE '.*hpc-macros.*'], proceed_on_failure => 1, timeout => 180);
```
coolgw, line 40: Q 1
"On SP4 we will only provide gnu11-*."  my understanding we need keep gnu11-*?

Q 2
If we need remove all "versioned" then we should using pattern
gnu[[:digit:]]+-compilers-hpc etc.

tinawang123: Only need remove gun9/10/11 others don't need remove.
It just installed those packages on sles15sp3 not related with sles15sp4

coolgw: Base Egbert's feedback, there seems some conflict, "On SP4 we will only provide gnu11-*"  vs "You should drop any versioned compilers"

tinawang123: At here, we only compared the installed packages before and after migration. As comment#3, sles15sp3 dropped gun9/10/11, so we don't need install it on sles15sp3.

b10n1k: why [vr](https://openqa.suse.de/tests/8710278/logfile?filename=autoinst-log.txt) still installs `gnu10-compilers-hpc*` and `gnu11-compilers-hpc*?

tinawang123: > why [vr](https://openqa.suse.de/tests/8710278/logfile?filename=autoinst-log.txt) still installs `gnu10-compilers-hpc*` and `gnu11-compilers-hpc*`?

Updated VR case.

e4t: If we were to keep any 'versioned' compiler packages, the tests will have to be adjusted for every SP - sometimes in the middle during a release cycle, as gnu<N>-compilers-hpc* may be replaced by gnu<N+1>-compilers-hpc* at any time during the cycle. 
But the problem here arose because maintenance released the wrong packages - despite the fact, these packages obviously were not able to satisfy all their dependencies.
The way you create the package list - even without this patch - actually does make sense - provided the data you rely on is ok and has been verified. There is something else that you may not catch with this method: there may be multiple packages that conflict with each other and cannot installed simultaneously.
Unfortunately we had this twice recently that multiple packages got released which had undetected dependency issues: flavors of gnu-compilers-hpc (this issue) as well as trilinos. I believe both issues happened thu a single maintenance incident:   https://smelt.suse.de/incident/22623/. Unfortunately, they were neither caught by the installcheck procedure done by the maintenance team nor thru explicit package installation done by maintenance QA.
I believe maintenance has done steps to remove the offending packages from the release, so this problem should go away fairly soon.
So for this SP we will get by with the fix as proposed here - as a quick solution. Going forward and in the hope such issues won't repeat itself, it's probably fine to drop this change again.

lib/services/hpcpackage_remain.pm
```diff
@@ -37,7 +37,8 @@ sub list_pkg {
 sub install_pkg {
     my $version = get_var('HDDVERSION');
     # Install all 'library wrappers' (packages matching -hpc, not having a version number with '_' after their name)
-    my @pkginstall = split('\n', script_output q[zypper search -r SLE-Module-HPC] . $version . q[-Pool -r SLE-Module-HPC] . $version . q[-Updates | cut -d '|' -f 2 | sed -e 's/ *//g' | grep -E '.*-hpc.*' | grep -vE 'system|module|suse' |  grep -vE '.*_[[:digit:]]+_[[:digit:]]+.*gnu|.*_[[:digit:]]+_[[:digit:]]+.*-hpc' | grep -vE '.*-static$' | grep -vE '.*hpc-macros.*'], proceed_on_failure => 1, timeout => 180);
+    record_soft_failure('bsc#1194917', 'Ignore to install gun9/10/11 packages refer bsc#1194917');
```
lemon-suse, line 40: Normally the record_soft_failure should be 'bug ID' , 'Bug title'. You can move the comments before this line if needed.

PR #14868, opened 2022-05-10
Remove soft failure bgo#731320/boo#1047262

rfan1:
- Related ticket: https://progress.opensuse.org/issues/110821
- Needles: n/a
- Verification run: n/a
No function impact there.

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -261,8 +261,9 @@ sub handle_login {
     my $mykey = check_var('DESKTOP', 'gnome') ? 'esc' : 'shift';
     send_key_until_needlematch('displaymanager', $mykey, 30, 3);
     if (get_var('ROOTONLY')) {
+        # we now use this tag to support login as root
         if (check_screen 'displaymanager-username-notlisted', 10) {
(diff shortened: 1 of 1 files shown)


lib/x11utils.pm
```diff
@@ -262,7 +262,7 @@ sub handle_login {
     send_key_until_needlematch('displaymanager', $mykey, 30, 3);
     if (get_var('ROOTONLY')) {
         if (check_screen 'displaymanager-username-notlisted', 10) {
-            record_soft_failure 'bgo#731320/boo#1047262 "not listed" Login screen for root user is not intuitive';
+            record_info 'bgo#731320/boo#1047262 "not listed" Login screen for root user is not intuitive';
```
lilyeyes, line 265: Should we delete these related codes since the bugs are fixed?

lilyeyes (approved):
LGTM.

lib/x11utils.pm
zhaocongjia: I suggest to keep the old bug reference and add additional info: "we now use this tag to support login as root user".

PR #14877, opened 2022-05-11
BCI: some time after console switch to prevent timeouts

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/110866

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -18,6 +18,7 @@ use Mojo::Base qw(consoletest);
 use XML::LibXML;
 use testapi;
 use File::Basename;
+use Utils::Architectures qw(is_ppc64le);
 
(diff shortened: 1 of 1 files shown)


tests/containers/bci_test.pm
```diff
@@ -92,6 +92,8 @@ sub run {
 
     record_info('Run', "Starting the tests for the following environments:\n$test_envs");
     my $bci_dir = fileparse($bci_tests_repo, qr/\.[^.]*/);
+    # sometimes ppc64le are still running type_string 'cd' and fail to run cd BCI-tests
```
grisu48, line 95: This comment suggest that the sleep should be in the next line. So either the comment is misleading or the sleep needs to be one line down.

grisu48 (approved):

jlausuch:
This doesn't help much... we need to find the root cause of this issue...

jlausuch:
This doesn't help much... we need to find the root cause of this issue...

PR #14879, opened 2022-05-11
Workaround sshxterm was not shown

hjluo:
Sometimes in the kdump_and_crash service check, we'd hit
bsc#1199326. This cause the ~/bernhard/.ssh/id_rsa zero-sized. 
we need to recreate the ssh key in this scenario.

- Related ticket: https://progress.opensuse.org/issues/110250  
- Needles: N/A
- Verification run: 
 https://openqa.nue.suse.com/t8733618
 https://openqa.nue.suse.com/t8733730
 https://openqa.nue.suse.com/t8733731
 https://openqa.nue.suse.com/t8733732


lib/kdump_utils.pm
```diff
@@ -379,6 +379,28 @@ sub check_function {
     }
 }
 
+# for bsc#1199326, we need to check if ~/bernhard/.ssh/id_rsa was not
+# affected after kdump_and_crash.
+sub check_ssh_files {
+    # if file ~bernhard/.ssh/id_rsa missing or zero size
+    # we need to recreate the ssh key.
+    my $ret1 = script_run('! test -e ~bernhard/.ssh/id_rsa');
+    my $ret2 = script_run('! test -s ~bernhard/.ssh/id_rsa');
```
lemon-suse, line 388: I think the '-s' is enough for file existing and with content, no need to use '-e' any more. WDYT?

lib/kdump_utils.pm
```diff
@@ -397,6 +419,10 @@ sub full_kdump_check {
     if ($stage ne 'before') {
         ensure_serialdev_permissions;
     }
+    # We need to check bsc#1199326 after kdump_and_crash
+    if ($stage eq 'after') {
+        check_ssh_files();
```
lemon-suse, line 424: I have an idea about the solution: 
Maybe we can save this ssh file before migration and restore it after migration if needed. WDYT?

lib/kdump_utils.pm
hjluo: it's include 2 kind of scenarios. 1 file not existed, the other is file existed but with zero size (null)

lib/kdump_utils.pm
hjluo: that's a little bit complicated when this bug happens the .ssh dir sometimes was removed.
so a freshly created ssh key is the same as a backup and most important is it wil 100% work.

lib/kdump_utils.pm
lemon-suse: The condition is '!' for 'file existed and with content', so I think the ''! test -s ~bernhard/.ssh/id_rsa' is enough for the two scenarios.

lib/kdump_utils.pm
lemon-suse: In fact, I think the fresh created process is complicated. Anyway, this definitely work, NP.

b10n1k:
none of the VR hits the workaround. Do you have any link to an actual resolved case?

hjluo:
> none of the VR hits the workaround. Do you have any link to an actual resolved case?

In the VR:
https://openqa.nue.suse.com/tests/8749956#step/check_upgraded_service/105
https://openqa.nue.suse.com/tests/8749957#step/check_upgraded_service/101

b10n1k (approved):
+1

lemon-suse (approved):

PR #14843, opened 2022-05-06
Check related hmac packages when FIPS installation

wangqianzou:
Progress issue: https://progress.opensuse.org/issues/110707
Check if related hmac packages are installed when installing FIPS.

- Related ticket: https://progress.opensuse.org/issues/110707
- Needles:  None
- Verification run
   -Env mode test:  https://openqa.suse.de/tests/8702841#step/fips_setup/76
   -Kernel mode test: http://openqa.suse.de/tests/8702834#step/fips_setup/73

PR #14880, opened 2022-05-11
Revert "Check related hmac packages when FIPS installation"

lilyeyes:
Reverts os-autoinst/os-autoinst-distri-opensuse#14843


wangqianzou (approved):
LGTM

PR #14883, opened 2022-05-11
Add add_update_test_repo to create_hdd_transactional_server

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/110253
- Needles: -
- Verification run: [SLE](https://openqa.suse.de/tests/overview?distri=sle&version=&build=maintenance_update_test_repo&groupid=96)

### DIFF
--- a/schedule/qam/common/create_hdd_transactional_server.yaml
+++ b/schedule/qam/common/create_hdd_transactional_server.yaml
@@ -5,22 +5,16 @@ description: >
     root filesystem to provide atomic, automatic updates of a
     system without interfering with the running system.
 vars:
-    SCC_ADDONS: tsm
+    DEPENDENCY_RESOLVER_FLAG: 1
+    DESKTOP: textmode
+    PATTERNS: base,minimal
     HDDSIZEGB: 40
-conditional_schedule:
-    kernel_cmd_parameters:
-        BACKEND:
-            pvm_hmc:
-            - installation/edit_optional_kernel_cmd_parameters
-    shutdown_svirt:
-        BACKEND:
-            svirt:
-            - shutdown/svirt_upload_assets
 schedule:
     - installation/bootloader_start
     - installation/welcome
     - installation/accept_license
     - installation/scc_registration
+    - installation/add_update_test_repo
     - installation/addon_products_sle
     - installation/system_role
     - installation/partitioning
@@ -29,8 +23,8 @@ schedule:
     - installation/user_settings
     - installation/user_settings_root
     - installation/resolve_dependency_issues
+    - installation/select_patterns
     - installation/installation_overview
-    - "{{kernel_cmd_parameters}}"
     - installation/disable_grub_timeout
     - installation/start_install
     - installation/await_install
@@ -44,4 +38,3 @@ schedule:
     - shutdown/grub_set_bootargs
     - shutdown/cleanup_before_shutdown
     - shutdown/shutdown
-    - "{{shutdown_svirt}}"


jknphy:
could you also run it for SLE-15-SP2 and remove the WIP if ready?

schedule/qam/common/create_hdd_transactional_server.yaml
```diff
@@ -11,16 +11,17 @@ conditional_schedule:
     kernel_cmd_parameters:
         BACKEND:
             pvm_hmc:
-            - installation/edit_optional_kernel_cmd_parameters
+                - installation/edit_optional_kernel_cmd_parameters
```
jknphy, line 14: apparently we don't run for anything different than x86_64, so we could remove `conditional_schedule`.

jknphy:
According to what we talked offline, we just need here to add test module `- installation/add_update_test_repo` and remove var `SCC_ADDONS: tsm`. There is not need to update the needle-based module to handle empty repos. Ideally we could provide a VR adding the right value for `SCC_ADDONS` from Medium, but this is optional because it might fail due to internal settings of variables in `main.pm`, so I'm also fine if we just merge here without VR, anyway the test was neglected for a long time not testing any updates, so we can wait for next build to see how it goes and act accordingly.

rakoenig:
> According to what we talked offline, we just need here to add test module `- installation/add_update_test_repo` and remove var `SCC_ADDONS: tsm`. 

Bad news about this: I had a VR with `SCC_ADDONS:` tsm removed. Turned out that both on [SLE-15SP2](https://openqa.suse.de/tests/8778736) and on [SLE15-SP3](https://openqa.suse.de/tests/8779160) the test suite then fails in `first_boot`. After restoring the setting it worked again. 

Regarding the patch for the test module: Can we avoid running the test without repos? The test for SLE15-SP2 was last run [6 hours ago](https://openqa.suse.de/tests/8786625), so if I merge the modified YAML with `add_update_test_repo` in place and no patch then we will hit the problem soon again. Only solution for this would be to move the YAML out of `schedule/qam/common` and place one with  `add_update_test_repo` for SP3 and one without for SP2 and then change the job group settings so that the jobs get scheduled using a YAML that doesn't end up in known problems. From my perspective it looks much easier to apply the patch to the test module than doing all this extra precautions to avoid problems.

jknphy:
tbh Idk if for a transactional update scenario makes sense a desktop but as it looks like this updates are tested all the same time I guess that just adding `DESKTOP=gnome` should avoid the problem with the login which is expecting textmode.
the part about changing the needle-based module I don't get it, I mean, what we need to do is to add all the repos, an scenario without them doesn't make any sense and it is how it is currently tested. So if we add the repos and fails we file a bug or even consider to remove the test suite temporarily, so we don't have to make it green, just ensure that we test the updates in this scenario (which never was the case).

rakoenig:
Finally achieved the holy grail of having a successful VR on SLE15-SP3. So PR is ready to merge.

jknphy:
> Finally achieved the holy grail of having a successful VR on SLE15-SP3. So PR is ready to merge.

you mean SP2, right? it looks practically done, good job investigating here!
before merging, could you confirm that you tested SP2 removing patterns as well, I could only find this VR:
https://openqa.suse.de/tests/8781468

rakoenig:
SP2 is working now, but needs a [MR for the job group](https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/257) in Gitlab.

jknphy:
We need the corresponding add-on for transactional to be added: `SCC_ADDONS: %SCC_ADDONS,tsm`
You are not installing a transactional server, you would easily notice that cloning its children:
https://openqa.suse.de/tests/8825075
https://openqa.suse.de/tests/8825076

PR #14860, opened 2022-05-09
Remove workaround for bsc#1161421

jknphy:
- Related ticket: https://progress.opensuse.org/issues/107815
- Verification run: [yast2_ncurses_textmode](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=jknphy%2Fos-autoinst-distri-opensuse%23remove_kdump_workaround)

PR #14887, opened 2022-05-12
Fix workaround bsc#1161421

jknphy:
Adjust  #14860 
- Related ticket: https://progress.opensuse.org/issues/107815
- Verification run: [sle](https://openqa.suse.de/tests/overview?distri=sle&build=jknphy%2Fos-autoinst-distri-opensuse%23fix_kdump&version=15-SP4)

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -141,8 +141,12 @@ sub handle_warning_install_os_prober {
 
 # use yast2 kdump to enable the kdump service
 sub activate_kdump {
+    my (%args) = @_;
+    # increase kdump memory when bsc#1161421 applies
+    my $increase_kdump_memory = $args{increase_kdump_memory} // 1;
     # restart info will appear only when change has been done
     my $expect_restart_info = 0;
+
     # get kdump memory size bsc#1161421
     my $memory_total = script_output('kdumptool  calibrate | awk \'/Total:/ {print $2}\'');
     my $memory_kdump = $memory_total >= 2048 ? 1024 : 320;
@@ -167,7 +171,7 @@ sub activate_kdump {
     }
     # ppcl64e and aarch64 needs increased kdump memory bsc#1161421
     if (is_ppc64le || is_aarch64) {
-        if (is_sle('<15-sp4')) {
+        if ($increase_kdump_memory) {
             send_key('alt-y');
             type_string $memory_kdump;
             wait_screen_change(sub { send_key 'ret' }, 10) for (1 .. 2);
--- a/tests/console/yast2_kdump.pm
+++ b/tests/console/yast2_kdump.pm
@@ -27,7 +27,7 @@ sub run {
     zypper_call('in yast2-kdump');
 
     # Kdump configuration with YaST module
-    kdump_utils::activate_kdump;
+    kdump_utils::activate_kdump(increase_kdump_memory => 0);
 
     # check service (without restarting)
     systemctl('is-enabled kdump');


jknphy:
hi @czerw I remember why I have to put that outside which looked a bit weird, but do you know why is different behavior between this two:
https://openqa.suse.de/tests/8737686#step/kdump_and_crash/38
https://openqa.suse.de/tests/8743111#step/yast2_kdump/16

both are enabled at the beginning but only one has the popup to reboot, perhaps I'm missing something obvious here? :)

czerw:
I don't think it is a bug, test should detect either need to confirm reboot or just continue without it. yast could detect different memory and it could lead for new memory value (both tests have different memory setup).

jknphy:
> I don't think it is a bug, test should detect either need to confirm reboot or just continue without it. yast could detect different memory and it could lead for new memory value (both tests have different memory setup).

ahm ok, it is that, right, thanks for the hint Petr, 4GB vs 1GB.

Nevertheless the tests should be predictable to save us time in their maintenance, selecting behavior based on needles is very fragile. This is old code we try to do better since a time ago, I remember we inherited this test suite a time ago but didn't have the chance to improve it yet.

Current code is very deceiving because types 2 'ret' one for enter the value and another for accept the screen, one line for two very different things, and then there is a `alt-ok` at the end that in these two archs serves for the ok button of the popup but in other cases just for the ok of the main screen. Definitely there is some gap for improvement here. 

I will create a ticked for us to take a look, perhaps this module could be a good candidate for libyui REST API and a better design. In the meantime.

I added another 'hack' that might work for all the cases, it is ugly, but after many tries and not finding what is going on due to lack of video, I came up with that, it looks like sometimes some ret is lost and sometimes does what is supposed to do depending on there is typing or not...

czerw:
https://openqa.suse.de/tests/8750817 still fails, but there is  no rush for solution

czerw:
Hmm looks like  that some scenarios still needs memory workaround https://openqa.suse.de/tests/8763993#step/kdump_and_crash/69  i will double check it more.

lib/kdump_utils.pm
```diff
@@ -172,8 +172,12 @@ sub activate_kdump {
             type_string $memory_kdump;
             wait_screen_change(sub { send_key 'ret' }, 10) for (1 .. 2);
             record_soft_failure 'default kdump memory size is too small for ppc64le and aarch64, see bsc#1161421';
+            $expect_restart_info = 1;
+        }
+        else {
+            wait_screen_change(sub { send_key 'ret' }, 10);
+            $expect_restart_info = 1 unless check_var('QEMURAM', '4096');
```
b10n1k, line 179: this can go after the first if (outside the if..else) apparently.

b10n1k (changes requested):
if i am not wrong the `$expect_restart_info` doesnt need to used twice. In additional, add a proper  commit message explaining why you do what you do, please

lib/kdump_utils.pm
jknphy: I thought the same and my initial code was like that, but apparently one of the 'ret' is used sometimes and sometimes not. Anyway the purpose of this PR is to remove the soft-failure if can be removed (Petr is checking). The code has two clear issues, (1) it utilizes the same logic to handle two different test condition actually using/processing 1 or 2 'ret's depending on the case (2) the workaround doesn't behave like a plugin, when you remove it, stop working.

jknphy:
> Hmm looks like that some scenarios still needs memory workaround https://openqa.suse.de/tests/8763993#step/kdump_and_crash/69 i will double check it more.

let me know, because from the point of view of YaST scope, given that we actually don't try to crash the system for real and just focus on the configuration, I could make it conditional to the memory size the upper block instead of using the product, in other words for yast2_kdump might not make sense to display the workaround in any product.

czerw:
> let me know, because from the point of view of YaST scope, given that we actually don't try to crash the system for real and just focus on the configuration, I could make it conditional to the memory size the upper block instead of using the product, in other words for yast2_kdump might not make sense to display the workaround in any product.


I think you are right, just please remove the QEMURAM condition,we can merge it. I will migrate all our tests to cli kdump version (including maintenance jobs).

czerw (approved):

jknphy:
> > let me know, because from the point of view of YaST scope, given that we actually don't try to crash the system for real and just focus on the configuration, I could make it conditional to the memory size the upper block instead of using the product, in other words for yast2_kdump might not make sense to display the workaround in any product.
> 
> I think you are right, just please remove the QEMURAM condition,we can merge it. I will migrate all our tests to cli kdump version (including maintenance jobs).

This is the cleanest solution I could find for this approach, basically the workaround will be done always except when we signal it from the test module which does not do any crash `yast2_kdump`.

PR #14900, opened 2022-05-13
Configure nfs sharing between hpc master and slave nodes

b10n1k:
Trying to simulate a real-world hpc setup, where the compute nodes are bare
machines which merely used to perform computation. This also makes the jobs a
bit faster as we do not need to provision those nodes with hpc packages. The
only compromise here is some required dependencies which hpc modules install
and are required to run mpi binaries.
    
To do so we need to export the location of the binary of which will run to test
mpi and the location of the installed HPC modules.

> The same needs to apply to spack. But expected to be slightly different situation there

- Related ticket: https://progress.opensuse.org/issues/109584

### DIFF
--- a/lib/hpc/formatter.pm
+++ b/lib/hpc/formatter.pm
@@ -15,7 +15,7 @@ has mpirun => sub {
     my ($self) = shift;
     my $mpi = get_required_var('MPI');
     $self->mpirun("mpirun");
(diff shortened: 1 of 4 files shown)


schlad:
this look really good!

schlad (approved):

PR #14905, opened 2022-05-13
WIP: Add firefox html5 selenium test

BillAnastasiadis:
*still WIP*
This is a POC exploring ways to use python-selenium in openqa for specific tests. The test runs a python script which uses selenium to access https://html5test.opensuse.org/ and check a number of values and scores, which are then parsed in perl to determine success/failure of the test.

The test also uploads the geckodriver log and the full test result log.

(limits and checks are arbitrarily selected for the sake of demonstration)

- Related ticket: https://progress.opensuse.org/issues/109896
- Verification runs:
example of successful run: http://10.161.229.209/tests/640
example of failed run (enforced higher score expectations): http://10.161.229.209/tests/645#step/firefox_html5/31

### DIFF
--- a/tests/x11/firefox/firefox_html5.pm
+++ b/tests/x11/firefox/firefox_html5.pm
@@ -34,22 +34,12 @@ sub run {
     assert_script_run "export PATH=\"\$(pwd):\$PATH\"";
     assert_script_run "wget --quiet " . data_url('selenium/selenium_html5.py') . " -O selenium_html5.py";
 
-    # run selenium script, parse results and upload logs
-    assert_script_run "python3 selenium_html5.py >&1 | tee selenium_output.txt ", timeout => 120;
-    my $res = script_output("cat selenium_output.txt");
-    my @errs = ();
-    foreach my $line (split(/\n/, $res)) {
-        if (index($line, 'FAIL') != -1) {
-            my ($match) = $line =~ /\[\[ ( (?:(?!\]\]).)* ) \]\]/x;
-            push(@errs, $match);
-        }
-    }
-    upload_logs('geckodriver.log', log_name => 'html5-geckodriver-log.txt');
+    # run selenium tests
+    script_run 'python3 selenium_html5.py >&1 | tee selenium_output.txt', timeout => 120;
+    # upload results and logs and conclude test
+    upload_logs('geckodriver.log', log_name => 'html5-geckodriver-log.txt', failok => 1);
     upload_logs('selenium_output.txt', log_name => 'html5-selenium-results.txt');
+    assert_script_run 'EXP="TESTS OK" ; FAIL="$(tail -n 1 selenium_output.txt)" ; [ "$EXP" == "$FAIL" ]', fail_message => 'Test failed: check "html5-selenium-results" log for details.';
     assert_script_run 'cd .. && rm -rf temp_selenium';
-    if (@errs) {
-        my $errors = join(" | ", @errs);
-        die "$errors";
-    }
 }
 1;
--- a/tests/x11/firefox/firefox_html5.pm
+++ b/tests/x11/firefox/firefox_html5.pm
@@ -5,41 +5,51 @@
 # SPDX-License-Identifier: FSFAP
 
 # Package: MozillaFirefox
-# Summary: Case#1479221: Firefox: HTML5 Video
-# - Launch xterm, kill firefox, cleanup previous firefox configuration, launch
-# firefox
-# - Open "youtube.com/html5" and check result
-# - Open "youtube.com/watch?v=Z4j5rJQMdOU" and check result
-# - Exit firefox
-# Maintainer: wnereiz <wnereiz@gmail.com>
+# Summary: Case#1479221: Firefox: HTML5
+# - Launch xterm, install python and selenium
+# - Open opensuse html5 test page
+# - Access various page elements and check results
+# - Close
+# Maintainer: vanastasiadis <vasilios.anastasiadis@suse.com>
 
 use strict;
 use warnings;
 use base "x11test";
 use testapi;
+use utils;
 
 sub run {
     my ($self) = @_;
-    $self->start_firefox_with_profile;
+    select_console "x11";
+    x11_start_program('xterm');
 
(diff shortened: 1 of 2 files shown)


BillAnastasiadis:
@foursixnine the script is - for the time being - in `data`, exactly where the ticket said it shouldn't be :p

The question here is: ImageMagick also makes use of a (bash) script, which also resides in data. If you take this into account, it makes sense that possible future selenium scripts would reside in a dedicated folder `data` too.

On the other hand, the scripts could also reside in the dedicated `scripts` repo of `os-autoinst`. No personal preference between the two. I'm open to suggestions.

tests/x11/firefox/firefox_html5.pm
```diff
@@ -5,41 +5,51 @@
...
+    zypper_call("in python3");
+    enter_cmd "exit";
+    assert_script_run("pip3 install selenium");
+    assert_script_run "mkdir temp_selenium && cd temp_selenium";
+    assert_script_run "wget https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz && tar -xf geckodriver-v0.30.0-linux64.tar.gz";
+    assert_script_run "export PATH=\"\$(pwd):\$PATH\"";
+    assert_script_run "wget --quiet " . data_url('selenium/selenium_html5.py') . " -O selenium_html5.py";
+
+    # run selenium script, parse results and upload logs
+    assert_script_run "python3 selenium_html5.py >&1 | tee selenium_output.txt ", timeout => 120;
```
b10n1k, line 38: the problem with this is that the `assert_script_run` will always pass, as it returns the exit code from the second part of the pipe. you could use `script_output` i guess. i havent found a better way to solve this myself. wdyt?

tests/x11/firefox/firefox_html5.pm
```diff
@@ -5,41 +5,51 @@
...
+    my $res = script_output("cat selenium_output.txt");
+    my @errs = ();
+    foreach my $line (split(/\n/, $res)) {
+        if (index($line, 'FAIL') != -1) {
+            my ($match) = $line =~ /\[\[ ( (?:(?!\]\]).)* ) \]\]/x;
+            push(@errs, $match);
```
foursixnine, line 44: Maybe we can look into pytest? another idea is to rather evaluate the exit code of the script, anything that is a failure, is more than `0`.

tests/x11/firefox/firefox_html5.pm
foursixnine: make the script write the output to a file, so if something is off with the script, a failure is a failure. On top of that either pytest or TAP for python should be good enough (https://pypi.org/project/pytest-tap/)

tests/x11/firefox/firefox_html5.pm
```diff
@@ -5,41 +5,51 @@
...
-            wait_still_screen(2);
-            next;
+    # prepare python + selenium
+    become_root;
+    quit_packagekit;
+    zypper_call("in python3");
+    enter_cmd "exit";
+    assert_script_run("pip3 install selenium");
+    assert_script_run "mkdir temp_selenium && cd temp_selenium";
+    assert_script_run "wget https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz && tar -xf geckodriver-v0.30.0-linux64.tar.gz";
```
foursixnine, line 33: I wonder if this can be packaged or if it's packaged already (Not for us atm, more of a rethoric question)

tests/x11/firefox/firefox_html5.pm
```diff
@@ -5,41 +5,51 @@
...
+    foreach my $line (split(/\n/, $res)) {
+        if (index($line, 'FAIL') != -1) {
+            my ($match) = $line =~ /\[\[ ( (?:(?!\]\]).)* ) \]\]/x;
+            push(@errs, $match);
         }
-        last;
     }
-    send_key_until_needlematch('firefox-testvideo', 'spc', 30, 5);
-    $self->exit_firefox;
+    upload_logs('geckodriver.log', log_name => 'html5-geckodriver-log.txt');
```
foursixnine, line 47: What does the geckodriver-log have that could be of interest for us? any errors or things whre it should be parsed or so?

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: True, that's why I write the first command output to the `output.txt` file and later read from that to make sure everything went fine. I do need to make the checks a bit more thorough though - thanks!

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: I'd say that since the dirs and scripts used by different tests won't be the same (probably - unless we standardize the way selenium is used), the only "generic" part of this is installing selenium, which is a single command

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: I thought that it would be useful to have in case of driver errors that lead to abortion of the test. But I don't see a reason to parse - if the test fails, the reviewer can take a look at the log and determine what exactly the problem is, if it's driver related

tests/x11/firefox/firefox_html5.pm
mloviska: One can always set `pipefail` in bash

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: Ultimately, I went with neither options, but rather checking the output file in the end. I *could* just let the python script fail and fail the test, but then the output file would not exist and be uploaded for the user to check details at their leisure. Pytest felt wrong for this purpose, see the related ticket for more details on that.

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: The test can be made to work in various ways (pipefail and pipestatus don't seem to work correctly though, don't know if it's openqa related or what is the exact problem) but it is not needed, as there are stuff to be done after the selenium script runs anyway, so aborting on failure was not the intended action.

PR #14912, opened 2022-05-16
control_center: install yast2-vpn on Leap before testing

nilxam:
On Leap, yast2-vpn isn't installed in the default.

- Related failrun: https://openqa.opensuse.org/tests/2350620#step/yast2_control_center/38
- Verification run: https://openqa.opensuse.org/tests/2351431#step/yast2_control_center/61

### DIFF
--- a/tests/yast2_gui/yast2_control_center.pm
+++ b/tests/yast2_gui/yast2_control_center.pm
@@ -325,10 +325,10 @@ sub run {
         # see bsc#1062331, sound is not added to the yast2 pattern
         ensure_installed 'yast2-boot-server yast2-sound';
     }
-    elsif (is_tumbleweed) {
+    elsif (is_tumbleweed || is_leap('>15.3')) {
         record_soft_failure('bsc#1182125', "yast2-online-update-frontend is not pre-installed on TW");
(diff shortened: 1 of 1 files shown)


tests/yast2_gui/yast2_control_center.pm
```diff
@@ -325,10 +325,10 @@ sub run {
         # see bsc#1062331, sound is not added to the yast2 pattern
         ensure_installed 'yast2-boot-server yast2-sound';
     }
-    elsif (is_tumbleweed) {
+    elsif (is_tumbleweed || is_leap('>15.3')) {
```
foursixnine, line 328:
```suggestion
    elsif (!is_sle || is_leap('>15.3')) {
```

foursixnine (approved):
Just a niptick, but looks good in general

tests/yast2_gui/yast2_control_center.pm
nilxam: `!is_sle` implies `is_leap` is true, but I'm not sure we have the same issue on `Leap < 15.4`, we will see how it is on 15.3 QU3 Respin, if we had the same case on QU3, we can just leave it as `elif (!is_sle)`

PR #14916, opened 2022-05-17
Fix update polkit cause GDM restart

lemon-suse:
Update polkit cause GDM restart which cause system reboot fail, we need add select_console after zypper patch to make test robust.

- Related ticket: https://progress.opensuse.org/issues/97310
- Needles: N/A
- Verification run: https://openqa.nue.suse.com/tests/8770153#step/zypper_patch/11

### DIFF
--- a/tests/migration/online_migration/zypper_patch.pm
+++ b/tests/migration/online_migration/zypper_patch.pm
@@ -26,6 +26,9 @@ sub run {
     add_test_repositories;
     set_zypp_single_rpmtrans;
     fully_patch_system;
+    # Sometimes update package 'polkit' will cause GDM restart, so after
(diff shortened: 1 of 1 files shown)


chcao:
The original failed case in the openqa ticket is a offline case, it failed in 'patch_sle', and since build79.1, the issue is gone.
Why the code modified here is in 'zypper_patch.pm'?

hjluo (approved):
LGTM

lemon-suse:
> The original failed case in the openqa ticket is a offline case, it failed in 'patch_sle', and since build79.1, the issue is gone. Why the code modified here is in 'zypper_patch.pm'?
This issue happened both on offline and online migration, last PR fixed  issue on offline migration test but missed online migration test.

PR #14933, opened 2022-05-19
gnuhealth: Adjust to version 4.0 setup dialogs

DimStar77:
Gnuhealt 4.x has a new 'configure languages' dialog

We
 * assert on the screen present
 * Load the default language (alt-l)
 * Continue to the next dialog (alt-n)

- Related ticket: https://progress.opensuse.org/issues/108770
- Needles: Done as part of VR
- Verification run: https://openqa.opensuse.org/t2357235

### DIFF
--- a/tests/gnuhealth/gnuhealth_client_first_time.pm
+++ b/tests/gnuhealth/gnuhealth_client_first_time.pm
@@ -21,6 +21,13 @@ sub run {
     enter_cmd "susetesting";
     assert_screen "$gnuhealth-module_configuration_wizard_start";
     send_key $gnuhealth_34 ? 'ret' : 'alt-o';
+    # gnuhealth 4.0 introduced a new language configuration screen; gnuhealth 4.x exists since leap 15.4
+    if (!is_leap('<15.4')) {
+        assert_screen "$gnuhealth-configure-language";
+        send_key 'alt-l';
+        assert_screen "$gnuhealth-module_configuration_wizard-next_step";
+        send_key 'alt-n';
+    }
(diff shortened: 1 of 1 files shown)


tests/gnuhealth/gnuhealth_client_first_time.pm
```diff
@@ -10,17 +10,25 @@ use base 'x11test';
 use strict;
 use warnings;
 use testapi;
-use version_utils 'is_leap';
+use version_utils qw(is_leap is_tumbleweed);
 
 sub run {
     my $gnuhealth = get_var('GNUHEALTH_CLIENT', 'gnuhealth-client');
     my $gnuhealth_34 = is_leap('<15.2');
+    my $gnuhealth_4 = is_tumbleweed() || is_leap('>=15.4');
```
okurz, line 18: Our tests should always consider openSUSE Tumbleweed as the default product without a version – you should like that ;). That means any differing behaviour should have an explicit exclude rule for the "older products" and potentially an explanation why the behaviour should differ. You can try to negate the logic check.

okurz (approved):

tests/gnuhealth/gnuhealth_client_first_time.pm
DimStar77: I was/am aware of that.. with gnuhealth being tracked only in Leap and TW, the alternative was

```
if (!is_leap('<15.4')) {

}
```

which looks rather silly to me - anyway: PR has been updated now

coogor:
Thanks for picking this up Dominique!

PR #14936, opened 2022-05-20
Fix registry path for UBI images

jlausuch:
The images are not BETA any more, but they forgot to update the
documentation, which still points to beta...

https://progress.opensuse.org/issues/111359


grisu48 (approved):
LGTM

PR #14937, opened 2022-05-20
JeOS: small fix for image_info

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/111362
VR:
https://openqa.opensuse.org/tests/2359265 (RPi)
https://openqa.opensuse.org/tests/2359274 (aarch64)
https://openqa.opensuse.org/tests/2359267 (x86_64)
https://openqa.suse.de/tests/8793664 (openstack)

### DIFF
--- a/tests/jeos/image_info.pm
+++ b/tests/jeos/image_info.pm
@@ -39,7 +39,7 @@ sub check_db {
     }
 
     # CASEDIR var is always set, and if when no specified, the value is "sle" for OSD and "opensuse" for O3
-    return if (get_required_var('CASEDIR') !~ m/^sle$|^opensuse$/);
+    return 0 if (get_required_var('CASEDIR') !~ m/^sle$|^opensuse$/);
 
     my $openqa_host = get_required_var('OPENQA_HOSTNAME');
     if ($openqa_host =~ /openqa1-opensuse|openqa.opensuse.org/) {    # O3 hostname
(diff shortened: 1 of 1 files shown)


ggardet (approved):
LGTM. Thanks!

tests/jeos/image_info.pm
```diff
@@ -177,8 +177,9 @@ sub run {
     }
 
     # Get the size of different file types
-    unless (is_openstack) {
-        my $btrfs_summary = script_output('btrfs filesystem df --mbytes --si /');
+    # This step applies to BTRFS images. For simplicity, others images will be skipped.
+    my $btrfs_summary = script_output('btrfs filesystem df --mbytes --si /', proceed_on_failure => 1);
+    if ($btrfs_summary !~ /ERROR/) {
```
mloviska, line 182: Not sure whether `btrfsprogs` are installed on OpenStack. If the package is not there, you will get `command not found` instead of `ERROR` ( that comes from `btrfsprogs`, right? )

tests/jeos/image_info.pm
```diff
@@ -39,7 +39,7 @@ sub check_db {
     }
 
     # CASEDIR var is always set, and if when no specified, the value is "sle" for OSD and "opensuse" for O3
-    return if (get_required_var('CASEDIR') !~ m/^sle$|^opensuse$/);
+    return 0 if (get_required_var('CASEDIR') !~ m/^sle$|^opensuse$/);
```
mloviska, line 42: Oh, returning `undef` was a problem before?

tests/jeos/image_info.pm
jlausuch: You are super right https://openqa.suse.de/tests/8793555#step/image_info/8

tests/jeos/image_info.pm
jlausuch: Maybe not, but there is `return 0` in all other conditions, just to be precise, and just `return 1` if everything is ok. Some sort of "return True/False" for python :)

tests/jeos/image_info.pm
jlausuch: Amended

PR #14939, opened 2022-05-20
Quickfix: Disable UBI9 on openSUSE

grisu48:
UBI9 images require CPU support x86-64-v2, which we don't have at the moment.

Verification run: https://openqa.opensuse.org/tests/2359269


jlausuch (changes requested):
wait for it, we might do cpu=host, let's hang on a bit.

jlausuch (approved):
Ok, agreed with @DimStar77 to have a quick fix to not block staging and further investigate next week. It would be still good to test this in openSUSE, but we might need to change some config (which will take more time).

PR #14940, opened 2022-05-20
Select serial console to get full CLI cmd output for debug

lemon-suse:
Select serial console to get full CLI cmd output for debug for service check, and update complicated script on SUT to simple perl script. The purpose is 'the entire command line and the output to stdout and stderr should be made available for debugging purposes'.

- Related ticket: https://progress.opensuse.org/issues/105388
- Needles: N/A
- Verification run:  http://openqa.suse.de/tests/8793403#step/install_service/38

### DIFF
--- a/tests/installation/install_service.pm
+++ b/tests/installation/install_service.pm
@@ -16,7 +16,12 @@ use main_common 'is_desktop';
 
 sub run {
 
-    select_console 'root-console';
+    if (get_var('SEL_SERIAL_CONSOLE')) {
+        opensusebasetest::select_serial_terminal();
+    }
+    else {
+        select_console 'root-console';
+    }
 
     install_services($default_services)
       if is_sle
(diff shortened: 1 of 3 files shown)


tests/installation/install_service.pm
```diff
@@ -16,7 +16,12 @@ use main_common 'is_desktop';
 
 sub run {
 
-    select_console 'root-console';
+    if (get_var('SEL_SERIAL_CONSOLE')) {
```
coolgw, line 19: I am not like introduce new setting, can hpc directly use serial console?

lemon-suse: Although this requirement comes from HPC, but I do want to select serial console for whole service check through the switch of 'SEL_SERIAL_CONSOLE' when needed.

lilyeyes (approved):
LGTM

PR #14943, opened 2022-05-20
Fix audit2allow test on 15-SP3

paolostivanin:
On 15-SP3, we need to use `$original_audit` instead of the `$audit_log_short`. 

Verification runs:
- 15-SP3:
  - aarch64: https://openqa.suse.de/tests/8796815
  - s390x: https://openqa.suse.de/tests/8796814
  - x86_64: https://openqa.suse.de/tests/8796659
- 15-SP2:
  - aarch64: https://openqa.suse.de/tests/8796816
  - s390x: https://openqa.suse.de/tests/8796817
  - x86_64: https://openqa.suse.de/tests/8796818

### DIFF
--- a/tests/security/selinux/audit2allow.pm
+++ b/tests/security/selinux/audit2allow.pm
@@ -21,6 +21,9 @@ sub run {
     my $original_audit = "/var/log/audit/audit.log";
     my $audit_log = "/var/log/audit/audit.txt";
     my $audit_log_short = "/var/log/audit/audit.short.txt";
+    # On 15-SP3 the first 500 lines do not contain the needed messages, therefore we
+    # have to use the full log when testing audit2allow.
+    my $audit_log_test = is_sle('=15-SP3') ? "$original_audit" : "$audit_log_short";
 
     $self->select_serial_terminal;
 
@@ -32,7 +35,7 @@ sub run {
     validate_script_output("audit2allow -i $audit_log", sub { m/allow\ .*_t\ .*;.*/sx });
     assert_script_run("tail -n 500 $audit_log > $audit_log_short");
     validate_script_output(
-        "audit2allow -w -i $audit_log_short",
+        "audit2allow -w -i $audit_log_test",
         sub {
             m/
 	    type=.*AVC.*denied.*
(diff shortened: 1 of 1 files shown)


rfan1:
@paolostivanin LGTM
But I have one quick question here, do you have any bug/poo ticket filed? why we have different logic to check the result for SLES15SP3 only?

tests/security/selinux/audit2allow.pm
```diff
@@ -45,8 +46,9 @@ sub run {
     upload_logs($audit_log_short);
 
     # create an SELinux module, make this policy package active, check the new added module
+    $cmd = "cat " . (is_sle('=15-SP3') ? "$original_audit" : "$audit_log_short") . "| audit2allow -M $test_module";
```
lilyeyes, line 49: Add one `" "` (blank) before `"|"` is better.

lilyeyes (approved):
LGTM.

lilyeyes:
NOTE: https://openqa.suse.de/tests/8796814 failed, maybe can be fixed by rerun.

tests/security/selinux/audit2allow.pm
Amrysliu, line 49: How about set a variable for audit log, e.g. `$audit_log = is_sle('=15-SP3') ? "$original_audit" : "$audit_log_short"`, then use this variable in following code.

paolostivanin: good point :) I'll do this

paolostivanin: good point, I'll do this :+1:

paolostivanin:
@rfan1 no, I didn't open any bug report. I don't think this is worth a bug, because the "short" audit log is taken (first 500 lines) from the original log. It seems that on openqa the short version doesn't contain the needed messages, but full one does.

rfan1:
> @rfan1 no, I didn't open any bug report. I don't think this is worth a bug, because the "short" audit log is taken (first 500 lines) from the original log. It seems that on openqa the short version doesn't contain the needed messages, but full one does.

Got, that makes sense, then I would suggest you adding some comments there, but not a big deal.

rfan1 (approved):

paolostivanin:
new verification runs:
- 15-SP3:
  - x86_64: https://openqa.suse.de/tests/8806804
  - aarch64: https://openqa.suse.de/tests/8806805
- 15-SP2:
  - x86_64: https://openqa.suse.de/tests/8806807
  - aarch64: https://openqa.suse.de/tests/8806806

PR #14957, opened 2022-05-24
Fix unstable firefox download test

michaelgrifalconi:
Use leap 15.3 images instead of 15.4 beacause the latter is updated
often and during the update the file is not available for download

- Related ticket: https://progress.opensuse.org/issues/111207
- Needles: no needles
- Verification run: https://openqa.suse.de/tests/8816235


Avinesh (approved):
Looks good!

PR #14961, opened 2022-05-25
Fix tss home dir creation on 15-SP2

paolostivanin:
On 15-SP2, the package `system-user-tss` is not available, therefore we have to manually create the home directory.

Verification runs:
- 15-SP3: https://openqa.suse.de/tests/8820675
- 15-SP2: https://openqa.suse.de/tests/8820673

### DIFF
--- a/tests/security/tpm2/tpm2_env_setup.pm
+++ b/tests/security/tpm2/tpm2_env_setup.pm
@@ -34,8 +34,16 @@ sub run {
     # by default in newer releases, we can workaroud it by install this package
     # manually
     if (script_run("ls $tss_home") != 0) {
-        record_soft_failure("bsc#1193305, the home directory of user $tss_user is not created, install 'system-user-tss' as a workaround");
-        zypper_call("in system-user-tss");
+        # The package is only present on SLE >= 15.sp3, so for the previous versions
+        # we have to create the user manually.
+        if (is_sle('>=15-SP3')) {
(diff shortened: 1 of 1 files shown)


tests/security/tpm2/tpm2_env_setup.pm
```diff
@@ -34,8 +34,16 @@ sub run {
...
     if (script_run("ls $tss_home") != 0) {
-        record_soft_failure("bsc#1193305, the home directory of user $tss_user is not created, install 'system-user-tss' as a workaround");
-        zypper_call("in system-user-tss");
+        # The package is only present on SLE >= 15.sp3, so for the previous versions
+        # we have to create the user manually.
+        if (!is_sle('<15-SP3')) {
+            record_soft_failure("bsc#1193305, the home directory of user $tss_user is not created, install 'system-user-tss' as a workaround");
+            zypper_call("in system-user-tss");
+        } else {
+            record_soft_failure("Due to bsc#1193305, the home directory of user $tss_user must be manually created.");
```
rfan1, line 43: Paolo, 
a minor comments, since we don't need to install the package on tw, we can set 
if (is_sle('>15-SP3'))

at the same time, I would suggest use record_info instead of record_soft_failure there since dev marked it as invalid :)

paolostivanin:
new vr runs:
* https://openqa.suse.de/tests/8820733
* https://openqa.suse.de/tests/8820734

rfan1 (approved):
LGTM

PR #14967, opened 2022-05-25
Provide user-virtio-terminal console option

b10n1k:
The `select_serial_terminal` picks by default `root-virtio-terminal` but the
subroutine provides parameter to select a console as normal user.

Somehow this found misconfigured. The condition was like that
`$console = $root ? 'root-virtio-terminal' : 'virtio-terminal';`. There was
two main issues. One was that the propor console was not added during the
`init_consoles` and the second was that the variable was not a valid option.

The first is solved easily by adding another line to call `add_console`.
In the second case the correct name is `user-virtio-terminal` instead
`virtio-terminal`. the underline variable is evaluated later against a regex
to assign some other variables, one of them is *$user* and another is *type*
which each are required for the user _login_ and the console _initialization_.

- Related ticket: https://progress.opensuse.org/issues/88193
- Verification run: http://aquarius.suse.cz/tests/10113

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -1229,8 +1229,11 @@ sub remount_tmp_if_ro {
 
  select_serial_terminal($root);
 
-Select most suitable text console with root user. The choice is made by
-BACKEND and other variables.
+Select most suitable text console. The optional parameter C<root> controls
+whether the console will have root privileges or not. Passing any value that
+evaluates to true will select a root console (default). Passing any value that
+evaluates to false will select unprivileged user console.
+The choice is made by BACKEND and other variables.
 
 Purpose of this wrapper is to avoid if/else conditions when selecting console.
 
@@ -1239,7 +1242,7 @@ default when parameter not specified) or prefer non-root user if available.
 
 Variables affecting behavior:
 C<VIRTIO_CONSOLE>=0 disables virtio console (use {root,user}-console instead
-of the default {root-,}virtio-terminal)
+of the default {root-,user-}virtio-terminal)
 NOTE: virtio console is enabled by default (C<VIRTIO_CONSOLE>=1).
 For ppc64le it requires to call prepare_serial_console() to before first use
 (used in console/system_prepare and shutdown/cleanup_before_shutdown modules)
@@ -1269,7 +1272,7 @@ sub select_serial_terminal {
         if (check_var('VIRTIO_CONSOLE', 0)) {
             $console = $root ? 'root-console' : 'user-console';
         } else {
-            $console = $root ? 'root-virtio-terminal' : 'virtio-terminal';
+            $console = $root ? 'root-virtio-terminal' : 'user-virtio-terminal';
         }
     } elsif (get_var('SUT_IP')) {
         $console = $root ? 'root-serial-ssh' : 'user-serial-ssh';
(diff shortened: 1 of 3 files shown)


lib/opensusebasetest.pm
```diff
@@ -1229,8 +1229,9 @@ sub remount_tmp_if_ro {
 
  select_serial_terminal($root);
 
-Select most suitable text console with root user. The choice is made by
-BACKEND and other variables.
+Select most suitable text console. By default the C<root> is used. To switch to a
```
kalikiana, line 1232:
```suggestion
Select most suitable text console. By default C<root> is used. To switch to a
```

lib/opensusebasetest.pm
```diff
@@ -1229,8 +1229,9 @@ sub remount_tmp_if_ro {
...
+no-root user (usually this is C<testapi::username>, the C<select_serial_terminal> can
+accept a scalar 0 value. The choice is made by BACKEND and other variables.
```
kalikiana, line 1234: What is `a scalar 0 value`?

b10n1k: an int 0

kalikiana: I still don't know what this tells me here.

At a guess you could be meaning this: `C<select_serial_terminal> accepts a number corresponding to the user ID`

b10n1k: What i try to document here is that `select_serial_terminal()` by default will give you a root terminal invoking the `root-virtio-terminal`. if it is called as `select_serial_terminal(0)` (note that 0 in perl is a scalar type) this shifts the selection to `user-virtio-terminal` which will give you a user's console. 
Let me try to submit a slightly different phrasing and we continue there.

b10n1k: *This change is ready for review*

b10n1k: please resolve and continue to https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/838b65095d71b4eea67010e0aac2a09a66ce9c40

lib/opensusebasetest.pm
```diff
@@ -1229,8 +1229,11 @@ sub remount_tmp_if_ro {
 
  select_serial_terminal($root);
 
-Select most suitable text console with root user. The choice is made by
-BACKEND and other variables.
+Select most suitable text console. C<root> parameter determines the login user
+on the console. By default, when it is invoked without arguments, it will give
+you a root console. To switch to a no-root user (usually this is
+C<testapi::username>), the C<select_serial_terminal> can accept a false scalar.
```
b10n1k, line 1235: @kalikiana ^^^

mdoucha:
```suggestion
Select most suitable text console. The optional parameter C<root> controls
whether the console will have root privileges or not. Passing any value that
evaluates to true will select a root console (default). Passing any value that
evaluates to false will select unprivileged user console.
```

mdoucha (approved):
Aside from the docs wording discussed above, LGTM.

lib/opensusebasetest.pm
b10n1k: *this change is ready for review*

kalikiana (approved):

PR #14979, opened 2022-05-27
Record soft failure when executing usr_sbin_smbd on 15-SP3

paolostivanin:
Add a soft failure when executing the workaround on 15-SP3, so when the fix from upstream is backported we won't forget to remove it.

poo: https://progress.opensuse.org/issues/111659


paolostivanin:
Verification runs:
- 15-SP2: https://openqa.suse.de/tests/8830671#step/usr_sbin_smbd/1
- 15-SP3: https://openqa.suse.de/tests/8830726#step/usr_sbin_smbd/1

PR #14971, opened 2022-05-26
Display status of rollback service

mloviska:
`SUSEConnect --rollback` might failed and the error is display in the
status of oneshot `rollback` service. If the `SUSEConnect` call fails in
`/usr/sbin/rollback-reset-registration`, `/var/lib/rollback/check-registration`
file is not removed.

- Related ticket: https://progress.opensuse.org/issues/111635
- Verification run: 
  * [sle-15-SP3-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220525-1-jeos-filesystem@64bit-virtio-vga](http://kepler.suse.cz/tests/17229#)
  * [sle-12-SP5-JeOS-for-kvm-and-xen-Updates-x86_64-Build20220525-1-jeos-filesystem@64bit-virtio-vga](http://kepler.suse.cz/tests/17227#)

PR #14980, opened 2022-05-27
Replace systemctl with script_run

jlausuch:
systemctl status fails if the process is not running.
bug introduced in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14971 

Failure: https://openqa.suse.de/tests/8828359#step/snapper_jeos_cli/98
VR: http://openqa.suse.de/t8830691


mloviska:
Thanks, nice catch! Could you please add `--lines --full` so we can see the full output?

jlausuch:
re-run https://openqa.suse.de/tests/8830751

PR #14990, opened 2022-05-30
Modify spack modules to use nfs sharing

b10n1k:
Same as the other mpi modules. Try to simulate a HPC setup as close to the
reality. The so-called modules should be able to run mpi code without
installing HPC specific modules. Although with spack this is a bit tricky and
we have actual to install the spack package itself. But later we can avoid to
run the compilation of the spack module we need, as we share them from the
management node(spack_master).

I set a job variable in the beginning of the execution which helps to avoid
edit the job variables manual and is used to distinguish when the regular
mpi modules or the spack is used. When is set the common funtions between
those two will choose what to share. Which, in spack case, is the /opt/spack
directory

- Related ticket: https://progress.opensuse.org/issues/109584
- Verification run: 
https://openqa.suse.de/tests/overview?version=15-SP4&build=b10n1k%2Fos-autoinst-distri-opensuse%2314990&distri=sle


b10n1k:
tagged as WIP as i encounter some destruction on aarch64. investigating

schlad (approved):

PR #14991, opened 2022-05-30
sles4sap: Fix missing Exporter import

dzedro:
Functions should be exported, but they are not without Exporter module


alvarocarvajald (approved):
LGTM

PR #14994, opened 2022-05-30
Add AutoConfig for Firefox to disable pop-ups on start

baierjan:
I adapted the code from https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/87 to create the AutoConfig file before the first Firefox launch.

For now, it is only used in the _x11/firefox_ test. Other can easily reuse the code by adding `$self->prepare_firefox_autoconfig;
` before launching the Firefox for the first time.

- Related ticket: https://progress.opensuse.org/issues/111416
- Verification run: http://polaris.suse.cz/tests/2546#step/firefox/9

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -16,7 +16,7 @@ use Config::Tiny;
 use Utils::Architectures;
 use utils;
 use version_utils qw(is_sle is_leap is_tumbleweed);
-use x11utils qw(select_user_gnome handle_gnome_activities);
+use x11utils qw(select_user_gnome start_root_shell_in_xterm handle_gnome_activities);
 use POSIX 'strftime';
 use mm_network;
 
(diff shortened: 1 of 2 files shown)


lib/x11test.pm
```diff
@@ -516,6 +516,36 @@ sub setup_mail_account {
...
+pref("browser.aboutwelcome.enabled", false);
+pref("browser.startup.upgradeDialog.enabled", false);
+pref("privacy.restrict3rdpartystorage.rollout.enabledByDefault", false);
+EOF
+});
+
+    save_screenshot;
+
+    # Close root shell
+    send_key "ctrl-d", 1;
```
okurz, line 544: Please avoid the positional parameter, use the syntax from https://github.com/os-autoinst/os-autoinst/blob/45a9ae5bd271e0415b8ef0835398ae0ffbac3e31/testapi.pm#L1343 or better don't use any "waiting" at all but explicit screen checks which in general can be faster and more stable, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/33e81b324e3079ca3ce3550578fdc7650e915569/tests/x11/sshxterm.pm#L32

okurz (changes requested):

okurz:
Writing this config file should not influence any older firefox ESR versions but would be cool if you could still test on Tumbleweed *and* Leap 15.3 or an even older SLE version. Should be trivial using openqa-clone-custom-git-refspec

baierjan:
> Writing this config file should not influence any older firefox ESR versions but would be cool if you could still test on Tumbleweed _and_ Leap 15.3 or an even older SLE version. Should be trivial using openqa-clone-custom-git-refspec

That would be
* another test (Gnome this time) for Tumbleweed: https://openqa.opensuse.org/t2395519
* for Leap 15.3: https://openqa.opensuse.org/t2395515
* for 12-SP3: https://openqa.suse.de/t8844740

Tests are running...

okurz (approved):

PR #14996, opened 2022-05-31
xfstests: Fix incorrect 'power_action' call

okurz:
Related progress issue: https://progress.opensuse.org/issues/99663

### DIFF
--- a/tests/xfstests/run.pm
+++ b/tests/xfstests/run.pm
@@ -471,7 +471,7 @@ sub run {
         };
         # If SUT didn't reboot for some reason, force reset
         if ($@) {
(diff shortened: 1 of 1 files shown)


okurz:
Updated the commit message with more details. I don't see the benefit of a VR as we likely do not run into this code anyway.

b10n1k (approved):
Approved. even without VR :). and thanks for the commit update. A bit better. Link was inconvenienced because 1) you had to read all of it instead a brief and quick read 2) the link contains link to other ticket/PR etc. So thanks again

perlpunk (approved):

PR #14998, opened 2022-05-31
Fix incorrect uses of 'record_soft_failure'

okurz:
record_soft_failure only takes one argument

Related progress issue: https://progress.opensuse.org/issues/99663


kalikiana (approved):
Looks like this might have been confused with `record_info` which actually takes **title** and **output** arguments, meaning we'll now see the whole description which was silently failing before. Nice!

perlpunk (approved):

PR #14999, opened 2022-05-31
schedule: only run yast2_rmt on SLE/Leap >= 15.1

DimStar77:
yast2_rmt no longer exists on Tumbleweed

- Related ticket: https://progress.opensuse.org/issues/111479
- Needles: N/A
- Verification run: NONE

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1458,7 +1458,7 @@ sub load_extra_tests_y2uitest_ncurses {
         return;
     }
     # start extra yast console tests (self-contained only) from here
-    loadtest "console/yast2_rmt" unless (is_sle('<15-sp1') || is_leap('<15.0'));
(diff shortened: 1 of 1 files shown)


okurz (approved):
oh geez. If it was removed from Tumbleweed then the approach is correct. I wonder though if the package is meant to be obsoleted.

DimStar77:
> oh geez. If it was removed from Tumbleweed then the approach is correct. I wonder though if the package is meant to be obsoleted.

```
r17 | dimstar_suse | 2022-05-22 18:27:19 | 60fe8257dd6cbdf16e133b8ec2c63442 | 1.3.3 | rq975588

[botdel] Package has had build problems for &gt;= 6 weeks
```

At least nobody cared enough to fix it in all that time

PR #15001, opened 2022-05-31
Podman pods: cleanup system host

jlausuch:
When we don't run podman play kube --down the next test module,
podman_firewall complains that there are containers running:
https://openqa.suse.de/tests/8850484#step/podman_firewall/115


pdostal:
Thank you Jose! Out of all the container tests we have firewall test is the only one expecting to start with no containers running… Bad luck.

PR #15007, opened 2022-06-01
Enable libyui-rest-api in create_hdd_yast_maintenance

JRivrain:
See https://progress.opensuse.org/issues/111497
Once this is merged, children jobs will have libyui-rest-api installed.

VRs:
http://waaa-amazing.suse.cz/tests/16959
http://waaa-amazing.suse.cz/tests/16960

Verified Qcow image. it contains the expected packages.

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
+++ b/data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
@@ -88,6 +88,7 @@
                 </services>
                 <ports config:type="list">
                     <port>22/tcp</port>
+                    <port>30000-50000/tcp</port>
                 </ports>
             </zone>
         </zones>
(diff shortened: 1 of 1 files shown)


data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep
```diff
@@ -210,6 +210,15 @@
 cd /etc/zypp/repos.d
 sed -i 's/enabled=1/enabled=0/' $(ls|grep -i nvidia)
 zypper lr
+
+# Regarding rpm lock see Q9@ https://documentation.suse.com/de-de/sles/15-SP1/html/SLES-all/app-ay-faq.html
+# quote : "during the post-script phase, YaST still has an exclusive lock on the RPM database."
+# the package cannot be installed the autoyast way, because of the changing package name (eg libyui-rest-api15)
+# only zypper allows to install "by capability".
+mv /var/run/zypp.pid /var/run/zypp.sav
+zypper -n in libyui-rest-api
```
jknphy, line 219: we will need a a couple of conditions with an 'or' in the middle with ERB to only run it for SLE-15-SP3 and SLE-15-SP4 (to be soon settled in maintenance). There is not libyui-rest-api for older products.
We are assuming that we will reuse the same profile even for SLE-15-SP2, that is why i wanted to schedule later (not sure how you took it so soon, maybe my bad :) but thanks for taking a look so soon anyway. Could you please add them, otherwise, we just need to mention this in the follow-up tickets for creating AutoYaST images in older products.

JRivrain: Condition added.

jknphy: we can do regex! that is even better! :)

jknphy:
could you also take care of the firewall adjustments? I guess it can be done on the profile itself instead of running test module.

PR #15009, opened 2022-06-02
Use cli for kdump check

hjluo:
We need to change kdump test from yast to cli, which will avoid
many unexpect yast2 needle matching errors.

And we can specify CRASH_MEMORY variable in your job group schedule
per architecture

- Related ticket:  https://progress.opensuse.org/issues/111938 
- Needles: N/A
- Verification run: 
  http://openqa.nue.suse.com/t8897362
  http://openqa.nue.suse.com/t8897363

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -346,7 +346,7 @@ sub check_function {
         reconnect_mgmt_console if is_pvm;
     }
     else {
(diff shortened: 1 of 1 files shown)


czerw (approved):

lilyeyes:
Do we need the optional "CRASH_MEMORY" parameter in "Settings"?

lilyeyes (approved):
lgtm.

PR #15012, opened 2022-06-03
Public Cloud: bugfix for cleanup introduced by 068d7dd

jlausuch:
We need to skip the vault call in the cleanup if we are using the new
credentials system

- Related ticket: https://progress.opensuse.org/issues/108509
Failure: https://openqa.suse.de/tests/8867283#step/ssh_interactive_end/12


jlausuch:
@asmorodskyi @pdostal be aware. I'm merging as this is a small fix and makes the tests fail.

PR #15021, opened 2022-06-07
SLE Micro 5.3: white list some of the journal errors

jlausuch:


Vogtinator (approved):
I wonder whether there should be syntax to have some of the ignores product-agnostic

mloviska:
> I wonder whether there should be syntax to have some of the ignores product-agnostic

Make sense, but is there any reason why not to ignore all of the `type: ignore`.

jlausuch:
> I wonder whether there should be syntax to have some of the ignores product-agnostic

yes, that's a good idea to avoid doing this again and again for new products...

jlausuch:
> > I wonder whether there should be syntax to have some of the ignores product-agnostic
> 
> Make sense, but is there any reason why not to ignore all of the `type: ignore`.

I haven't bumped into them yet, slem5.3 is still under high changes, and they might show up later. Some of the tests still fail to boot. So, expect a second round when tests get a bit more stable in the regular modules and start executing journal_check.

Vogtinator:
IMO allowing to omit the `products` key completely could be an intuitive way to express that.

> > I wonder whether there should be syntax to have some of the ignores product-agnostic
> 
> Make sense, but is there any reason why not to ignore all of the `type: ignore`.

Some of the ignores are IMO a bit too unspecific and could hide real errors, but arguably that's something to be fixed in a different way. If we assume that all `ignores` are knowingly produced by the test code, then it could be done unconditionally.

PR #15023, opened 2022-06-07
Use normal user to run HPC relative tasks on mpi jobs

b10n1k:
Once the `user-virtio-console` resolved we can now switch easily to the normal
user for all the mpi implementations. As per `select_serial_terminal` was
called earlier to get a root console, we have to kill that session, because
the implementation doesnt taking care of this scenario and fails in a login
check.
During the implementation i encounter some timeouts and such i bump up the
particular parameter in specific invocations. However not sure whether it was
a sporadic issue on my local OpenQA or something else.
Finally i modified the `generate_and_distribute_ssh` to make copies only on
the compute nodes.

To run MPI binaries on the compute nodes as a regular user the management node
has to export the `/opt/spack` and `$HOME/spack` in additional of the HPC
libraries.

Initialy i thought that sharing the whole `$HOME` could work but during the
testing that failed. Likely because it is override previous steps that we have
performed, for instance the ssh key configuration. Anyhow, the functions which
setup the NFS were slightly modified to accept a hash ref with the absolute
paths to export. That impacted also the *mpi_master* ans *mpi_slave* modules.

Due to the need to run some commands as root, there are several switches
between the root and the regural user, which look ungly and need to improve in the future.

- Related ticket: https://progress.opensuse.org/issues/111206

- Verification run: 
https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2315023

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -143,11 +143,17 @@ sub distribute_slurm_conf {
 
 =head2 generate_and_distribute_ssh
 
-Generates and distributes ssh keys across all cluster nodes
+     generate_and_distribute_ssh($user)
+
+Generates and distributes ssh keys across compute nodes. C<user> by default is set
+to B<root> user unless another value is passed to the parameters.
+C<user> is used to determine the user on the remote machine where the ssh_id will
+be copied.
 
 =cut
 sub generate_and_distribute_ssh {
     my ($self, $user) = @_;
+    $user //= 'root';
     my @slave_nodes = slave_node_names();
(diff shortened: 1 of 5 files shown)


schlad:
'relative tasks' - what's that?

b10n1k:
> 'relative tasks' - what's that?

Reading this again i got confused too :) But what it means is that we have mpi modules and then mpi tests with spack.

schlad (approved):

PR #15027, opened 2022-06-08
Quotes added in tox for BCI_IMAGE_MARKER expression

m-dati:
In the **tox** command the -k options for **pytest** allows to select specific test based on a string stored in BCI_IMAGE_MARKER, 

`tox -e <test> -- -k $BCI_IMAGE_MARKER`

But also boolean expressions with and/or/not, enclosed in quotes, can be used for more specific selections, as BCI_IMAGE_MARKER="<test_group> and <test_name> or <string>", so that when expanded the command becomes:

`tox -e <test> -- -k "<test_group> and <test_name> or <string>"`

This PR is to add quotes around that parameter, otherwise the boolean term are not aggregated for the -k option.

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -67,7 +67,7 @@ sub run_tox_cmd {
     my $bci_reruns = get_var('BCI_RERUNS', 3);
     my $bci_reruns_delay = get_var('BCI_RERUNS_DELAY', 10);
     my $cmd = "tox -e $env -- -n auto";
-    $cmd .= " -k $bci_marker" if $bci_marker;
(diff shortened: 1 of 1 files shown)


m-dati:
Here is a test run that PASS, also showing the selection of a single test:
BCI_IMAGE_MARKER = bci/python_3.10 and webserver_1 or (bci/python:3.10 and all)
http://marvin5.arch.suse.cz/tests/273#step/bci_test/24
http://marvin5.arch.suse.cz/tests/273#step/python/1

jlausuch (approved):
LGTM!

PR #15031, opened 2022-06-08
chromium: Use pre-commit input instead of waiting

kalikiana:
- Use --allow-pre-commit-input to ensure key input is processed even
if rendering is still on-going, see
https://bugs.chromium.org/p/chromium/issues/detail?id=987626 for details
This means we don't need to check for needles to see if the urlbar is
active. We just type blind and wait for when the website is loaded.
- Use default typing speed because of the previous point.

See: https://progress.opensuse.org/issues/109737

### DIFF
--- a/tests/x11/chromium.pm
+++ b/tests/x11/chromium.pm
@@ -12,13 +12,9 @@ use Mojo::Base 'x11test', -signatures;
 use testapi;
 use utils;
 
-# Prevent lost characters due to temporary unresponsiveness of chromium address bar.
-# It looks like chromium becomes unresponsive after typing about 3
-# characters, likely due to some auto-completion feature.
-# See https://progress.opensuse.org/issues/109737 for details
 sub type_address ($string) {
-    type_string substr($string, 0, 10), @_, max_interval => utils::SLOW_TYPING_SPEED;
-    enter_cmd substr($string, 10), @_;
+    send_key 'ctrl-l';    # select text in address bar
+    enter_cmd($string);
 }
 
 sub run {
@@ -27,25 +23,16 @@ sub run {
     ensure_installed 'chromium';
 
     # avoid async keyring popups
-    x11_start_program('chromium --password-store=basic', target_match => 'chromium-main-window', match_timeout => 50);
+    # allow key input before rendering is done, see poo#109737 for details
+    x11_start_program('chromium --password-store=basic --allow-pre-commit-input', target_match => 'chromium-main-window', match_timeout => 50);
 
-    wait_screen_change { send_key 'esc' };    # get rid of popup (or abort loading)
-    send_key 'ctrl-l';    # select text in address bar
-
-    # Additional waiting to prevent unready address bar
-    # https://progress.opensuse.org/issues/36304
-    assert_screen 'chromium-highlighted-urlbar';
-    type_address('chrome://version ');
+    type_address('chrome://version');
     assert_screen 'chromium-about';
 
-    send_key 'ctrl-l';
-    assert_screen 'chromium-highlighted-urlbar';
     type_address('https://html5test.opensuse.org');
     assert_screen 'chromium-html-test', 90;
 
     # check a site with different ssl configuration (boo#1144625)
-    send_key 'ctrl-l';
-    assert_screen 'chromium-highlighted-urlbar';
     type_address('https://upload.wikimedia.org/wikipedia/commons/d/d0/OpenSUSE_Logo.svg');
     assert_screen 'chromium-opensuse-logo', 90;
     send_key 'alt-f4';


perlpunk:
I think the right link to the verificaion run might be https://openqa.opensuse.org/tests/2406610#step/chromium/19

tests/x11/chromium.pm
```diff
@@ -12,41 +12,29 @@ use Mojo::Base 'x11test', -signatures;
...
-
-    wait_screen_change { send_key 'esc' };    # get rid of popup (or abort loading)
-    send_key 'ctrl-l';    # select text in address bar
-
-    # Additional waiting to prevent unready address bar
-    # https://progress.opensuse.org/issues/36304
-    assert_screen 'chromium-highlighted-urlbar';
-    type_address('chrome://version ');
+    launch_with_url('chrome://version');
     assert_screen 'chromium-about';
```
okurz, line 29: Using `$target_match` within the `x11_start_program` and then `assert_screen` is redundant. Better pass on a parameter `$target_match` as well:

```suggestion
    launch_with_url('chrome://version', target_match => 'chromium-about');
```

I would likely use only named parameters and make it

```suggestion
    launch(url => 'chrome://version', target_match => 'chromium-about');
```

okurz (changes requested):

tests/x11/chromium.pm
kalikiana: Actually the url is the important part here, the rest we can pass through directly

tests/x11/chromium.pm
```diff
@@ -12,43 +12,32 @@ use Mojo::Base 'x11test', -signatures;
...
-    assert_screen 'chromium-highlighted-urlbar';
-    type_address('https://upload.wikimedia.org/wikipedia/commons/d/d0/OpenSUSE_Logo.svg');
-    assert_screen 'chromium-opensuse-logo', 90;
+    launch_with_url('https://upload.wikimedia.org/wikipedia/commons/d/d0/OpenSUSE_Logo.svg',
+        target_match => 'chromium-opensuse-logo');
+
+    # the version URL has to be typed since it is ignored on the command-line
+    send_key 'ctrl-l';    # select text in address bar
+    enter_cmd('chrome://version');
+    assert_screen('chromium-about');
```
okurz, line 38: so you are actually going back to typing faster again but hope that it works reliably because you opened chromium in before or why should this work? I suggest to trigger multiple hundreds of jobs for verification. If this does not work in a stable manner I suggest to try use `assert_and_click` as alternative

okurz (changes requested):

tests/x11/chromium.pm
kalikiana: No. Because of --allow-pre-commit-input

okurz: oh, cool. Where have you found that? Feel welcome to share a bit more in the git commit message :) Still then please at least *one* updated verification run for the happy-path :)

kalikiana: https://openqa.opensuse.org/tests/2408670#step/chromium/24

Vogtinator:
I haven't actually seen any recent chromium fails, but maybe I was just lucky...

IMO this goes against the "what a user would do" guideline, but if there are too many failures without this this PR looks good.

perlpunk (approved):

tests/x11/chromium.pm
```diff
@@ -27,27 +23,19 @@ sub run {
     ensure_installed 'chromium';
 
     # avoid async keyring popups
-    x11_start_program('chromium --password-store=basic', target_match => 'chromium-main-window', match_timeout => 50);
-
-    wait_screen_change { send_key 'esc' };    # get rid of popup (or abort loading)
-    send_key 'ctrl-l';    # select text in address bar
+    # allow key input before rendering is done, see poo#109737 for details
+    x11_start_program('chromium --password-store=basic --allow-pre-commit-input', target_match => 'chromium-main-window');
```
okurz, line 27: why delete the `match_timeout`? I don't recall what the default match timeout is but I assume it was there for a purpose

tests/x11/chromium.pm
```diff
@@ -27,27 +23,19 @@ sub run {
...
     type_address('https://html5test.opensuse.org');
-    assert_screen 'chromium-html-test', 90;
+    assert_screen('chromium-html-test');
 
     # check a site with different ssl configuration (boo#1144625)
-    send_key 'ctrl-l';
-    assert_screen 'chromium-highlighted-urlbar';
     type_address('https://upload.wikimedia.org/wikipedia/commons/d/d0/OpenSUSE_Logo.svg');
-    assert_screen 'chromium-opensuse-logo', 90;
+    assert_screen('chromium-opensuse-logo');
```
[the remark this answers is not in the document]
kalikiana: Originally I deleted the timeout when changing to rely on x11_start_program because it defaults to 90s. The default here is 30s but it still seems to work fine so I figured if we can reduce it that's even better

tests/x11/chromium.pm
kalikiana: It's 90s

tests/x11/chromium.pm
```diff
@@ -27,27 +23,19 @@ sub run {
...
+
+    # the version URL has to be typed since it is ignored on the command-line
```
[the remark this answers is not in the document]
kalikiana: It only makes sense here because the special URL can't be opened via the command-line. But you can argue it's irrelevant since we type everything again

okurz (approved):
We can accept without multiple verification runs but then please monitor production runs for stability in the upcoming days

PR #15033, opened 2022-06-09
Migrate EKS test to new account

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/111983
- Verification run: http://marvin5.arch.suse.cz/tests/285

### DIFF
--- a/lib/publiccloud/eks.pm
+++ b/lib/publiccloud/eks.pm
@@ -19,8 +19,10 @@ sub init {
     my ($self, %args) = @_;
     $self->SUPER::init("EKS");
 
-    my $cluster = get_var("PUBLIC_CLOUD_K8S_CLUSTER", "qe-c-testing-2");
-    assert_script_run("aws eks update-kubeconfig --name $cluster", 120);
+    my $cluster = get_var("PUBLIC_CLOUD_K8S_CLUSTER", "qe-c-openqa");
+    my $region = get_var("PUBLIC_CLOUD_ZONE", "eu-central-1");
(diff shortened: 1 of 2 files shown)


lib/publiccloud/eks.pm
```diff
@@ -19,8 +19,10 @@ sub init {
     my ($self, %args) = @_;
     $self->SUPER::init("EKS");
 
-    my $cluster = get_var("PUBLIC_CLOUD_K8S_CLUSTER", "qe-c-testing-2");
-    assert_script_run("aws eks update-kubeconfig --name $cluster", 120);
+    my $cluster = get_var("PUBLIC_CLOUD_K8S_CLUSTER", "qe-c-openqa");
+    my $region = get_var("PUBLIC_CLOUD_ZONE", "eu-central-1");
+    assert_script_run("aws eks update-kubeconfig --name $cluster --region $region", 120);
+    assert_script_run("kubectl get nodes");
```
jlausuch, line 25: should we validate the output somehow? like the output really returns some nodes?

ilausuch: This is an early check. We care here is to be sure that kubectl can connect to the cluster, and with assert_script_run this is covered. But also, the result of this operation is random (we could have here interference results from other tests). For this reason is not necessary and inconvenient to check the results

jlausuch (approved):

asmorodskyi (approved):
LGTM

pdostal:
🎉

PR #15035, opened 2022-06-09
Migrate helm tets to new accounts

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/111917
- Verification run: http://marvin5.arch.suse.cz/tests/342#

### DIFF
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -49,18 +49,16 @@ sub run {
         # Configure the public cloud kubernetes
         if ($k8s_backend eq "EKS") {
             add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
-            zypper_call("in jq aws-cli");
+            zypper_call("in jq aws-cli", timeout => 300);
 
-            $provider = publiccloud::eks->new(
-                region => get_var('PUBLIC_CLOUD_REGION', 'eu-central-1')
-            );
+            $provider = publiccloud::eks->new();
         }
         elsif ($k8s_backend eq 'AKS') {
             add_suseconnect_product(get_addon_fullname('pcm'),
                 (is_sle('=12-sp5') ? '12' : undef));
             add_suseconnect_product(get_addon_fullname('phub'))
               if is_sle('=12-sp5');
-            zypper_call('in jq azure-cli');
+            zypper_call('in jq azure-cli', timeout => 300);
 
             $provider = publiccloud::aks->new();
         }


tests/containers/helm.pm
```diff
@@ -44,18 +44,21 @@ sub run {
     else {
         install_kubectl();
 
+        my $base = get_required_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE");
```
pdostal, line 47: It don't think it makes sense to create yet another variable. I think that `PUBLIC_CLOUD_CREDENTIALS_URL` is enough.

pdostal (changes requested):

tests/containers/helm.pm
ilausuch: yes, I agree. changed

pdostal: There are still remains…

ilausuch:
New solution works

tests/containers/helm.pm
```diff
@@ -44,18 +44,22 @@ sub run {
...
+        set_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE", get_required_var("PUBLIC_CLOUD_CREDENTIALS_URL")) unless get_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE");
+        my $base = get_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE");
         my $provider;
 
         # Configure the public cloud kubernetes
         if ($k8s_backend eq "EKS") {
             add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
             zypper_call("in jq aws-cli");
 
+            set_var("PUBLIC_CLOUD_CREDENTIALS_URL", $base . 'aws.json');
```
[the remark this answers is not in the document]
ilausuch: This is what we did https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f13d49c793889a04f9d2763bb86531cfe48cf4a5/lib/main_containers.pm#L145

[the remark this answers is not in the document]
jlausuch: Yes, this is a limitation we didn't think before. Some jobs might use the 3 providers at the same time (in this case Helm tets). Splitting the job into 3 jobs would be just overkill. Instead, we need to find a solution for to allow this flexibility we had with Vault.

Maybe, for Helm tests, we could assume that `PUBLIC_CLOUD_CREDENTIALS_URL` does not contain the last part of the url, and we can simply do:

```
my $base = get_required_var('PUBLIC_CLOUD_CREDENTIALS_URL');
set_var("PUBLIC_CLOUD_CREDENTIALS_URL", $base . 'azure.json');
```

tests/containers/helm.pm
```diff
@@ -44,18 +44,22 @@ sub run {
     else {
         install_kubectl();
 
+        set_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE", get_required_var("PUBLIC_CLOUD_CREDENTIALS_URL")) unless get_var("PUBLIC_CLOUD_CREDENTIALS_URL_BASE");
```
grisu48, line 47: This might confuse us in the future.

Let's assume you work on a test case and forgot to set the variables. openQA will in this case complain about the missing variable `PUBLIC_CLOUD_CREDENTIALS_URL`, however it should complain about the missing `PUBLIC_CLOUD_CREDENTIALS_URL_BASE` variable instead, as this is the primary reference.

Is there any good reason for overtaking one variable with another one? This increases the risk of a confused poor QA engineer in the future.

grisu48: I would simply remove this section unless necessary. Overtaking variables adds unnecessary redundancy which makes our lives often more difficult for no real advantage IMHO.

jlausuch: There has been a [long discussion](https://suse.slack.com/archives/C02CGKBCGT1/p1654855363080599) about this, even 2 closed PRs. I agree with you, I don't like this hack, and we need to agree on a solution for these cases and make it generic. There was somewhere proposed to re-use the idea of namespaces we have in Vault. But we need agreement before merging this.

jlausuch (approved):

asmorodskyi (approved):
LGTM

pdostal (approved):
🚀

PR #15040, opened 2022-06-10
Use Public AWS ECR registry

jlausuch:
The VM where the current registry service is running today will be
decommissioned soon as we are migrating the Public Cloud accounts.

There is solution using the public ECR Gallery which basically contains
the same images of DockerHub.
https://gallery.ecr.aws

There are some limitations when pulling images with docker, like
toomanyrequests: Rate exceeded
but with using retries, we can workaround the problem.

This solution is desired as we don't need to maintain any registry
ourselves which has caused some issues in the past.

https://progress.opensuse.org/issues/98415

[Tumbleweed](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_new_registry2)
[SLE](https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=jlausuch%2Fos-autoinst-distri-opensuse%23containers_new_registry2)

### DIFF
--- a/lib/containers/docker.pm
+++ b/lib/containers/docker.pm
@@ -26,9 +26,7 @@ sub configure_insecure_registries {
     my $registry = registry_url();
     # The debug output is messing with terminal in migration tests
     my $debug = (get_var('UPGRADE')) ? 'false' : 'true';
-    # Allow our internal 'insecure' registry
-    assert_script_run(
-'echo "{ \"debug\": ' . $debug . ', \"insecure-registries\" : [\"localhost:5000\", \"registry.suse.de\", \"' . $registry . '\"] }" > /etc/docker/daemon.json');
+    assert_script_run('echo "{ \"debug\": ' . $debug . ', \"insecure-registries\" : [\"localhost:5000\", \"registry.suse.de\"] }" > /etc/docker/daemon.json');
     assert_script_run('cat /etc/docker/daemon.json');
     systemctl('restart docker');
     record_info "setup $self->runtime", "deamon.json ready";
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -136,14 +136,14 @@ sub basic_container_tests {
 
     #   - pull minimalistic alpine image of declared version using tag
     #   - https://store.docker.com/images/alpine
-    assert_script_run("$runtime image pull $alpine", timeout => 300);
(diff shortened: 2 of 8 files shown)


lib/containers/utils.pm
```diff
@@ -83,7 +83,7 @@ sub registry_url {
 sub runtime_smoke_tests {
     my %args = @_;
     my $runtime = $args{runtime};
-    my $image = $args{image} // registry_url('alpine', '3.6');
+    my $image = $args{image} // registry_url('alpine', 'latest');
```
Vogtinator, line 86: Is this intentional?

lib/containers/docker.pm
```diff
@@ -26,9 +26,7 @@ sub configure_insecure_registries {
     my $registry = registry_url();
```
m-dati, line 26: Is this `registry` variable no more used?

lib/containers/utils.pm
jlausuch: Yes, 3.6 is not available and it's ancient. Why not using just latest?

lib/containers/docker.pm
jlausuch: yes, I've seen it in some places, I didn't want to touch it.

lib/containers/utils.pm
Vogtinator: For consistent test results.

IMO this should be in a separate commit at least.

jlausuch: done

jlausuch:
I have ran an experiment of running 500 times the same job to check the reliability of using ECR public registry... 
https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=ecr_registry_test

Around 20% of the tests failed with 
`Error response from daemon: toomanyrequests: Rate exceeded`

I know 500 jobs in less than 3 hours is a bit extreme... but this proves that this public registry makes our tests prone to sporadic failures which we can do much about, as we are already doing `script_retry` for the image pull commands...  

I will do the same experiment with the former approach of creating the registry in a VM using `registry:2` https://hub.docker.com/_/registry

jlausuch:
I have created another VM with a public registry in our new Public Cloud accounts. It doesn't have pull rate limits and I consider it as the most important feature for our tests to be more stable.

PR #15045, opened 2022-06-10
Public Cloud: use different variables for credentials URL

jlausuch:
This way, we can use more than one provider in a single job.

There is another proposal here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15044
It uses single variable but hardcodes the suffix for each provider, but I think hardcoding the url suffix makes things more complicated, as we will need different suffixes for each account. So, I vote for the approach presented here where we have 3 different variables, one per provider.

### DIFF
--- a/lib/publiccloud/gcp_client.pm
+++ b/lib/publiccloud/gcp_client.pm
@@ -36,8 +36,8 @@ has vault => undef;
 sub init {
     my ($self) = @_;
     # For now we support Vault and the credentials-microservice. Vault will be removed after a certain transition period
-    if (get_var('PUBLIC_CLOUD_CREDENTIALS_URL')) {
-        my $data = get_credentials(CREDENTIALS_FILE);
+    if (my $url = get_var('PUBLIC_CLOUD_CREDENTIALS_URL_GCP')) {
+        my $data = get_credentials($url, CREDENTIALS_FILE);
         $self->project_id($data->{project_id});
         $self->account($data->{client_id});
     } else {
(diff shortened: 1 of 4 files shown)


ilausuch (approved):

ilausuch:
I agree that this version is better than the other, more flexible

lib/publiccloud/gcp_client.pm
```diff
@@ -36,8 +36,8 @@ has vault => undef;
 sub init {
     my ($self) = @_;
     # For now we support Vault and the credentials-microservice. Vault will be removed after a certain transition period
-    if (get_var('PUBLIC_CLOUD_CREDENTIALS_URL')) {
-        my $data = get_credentials(CREDENTIALS_FILE);
+    if (my $url = get_var('PUBLIC_CLOUD_CREDENTIALS_URL_GCP')) {
+        my $data = get_credentials($url, CREDENTIALS_FILE);
```
asmorodskyi, line 40:
```suggestion
        my $data = get_credentials($url . 'GCE.json', CREDENTIALS_FILE);
```

with keeping single variable

asmorodskyi: same applies to other clients

jlausuch: Yes, this is what I proposed in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15044
And in the description of this PR you'll see why I don't like that :)
What about SAP or any other account using this code? They won't use `GCE.json` but something else..

jlausuch:
Agreed to use namespaces for each account and hardcode the provider suffix, e.g. for AWS:
```
my $url = get_required_var('PUBLIC_CLOUD_CREDENTIALS_URL') . '/' . get_required_var('PUBLIC_CLOUD_NAMESPACE') . '/aws.json'; 
```

PR #15019, opened 2022-06-07
Trento test first implementation.

mpagot:
Two schedule provided:

- **trento_jumphost** : designed to create the Jumphost qcow2 file.
   The test will check the presence of all needed tools and eventually
   add them (az, helm, podman). The cypress image is pulled to be ready
   available. For the moment the test is expected to run against publiccloud_tools_xxx.qcow2

```
openqa-clone-job --from https://openqa.suse.de/tests/8736011 --skip-chained-deps  YAML_SCHEDULE=schedule/sles4sap/trento/trento_jumphost.yml HDD_1=publiccloud_tools_0034.qcow2  PUBLISH_HDD_1=trento_tools.qcow2
```
Execution example http://1c242.qa.suse.de/tests/167  and http://1c242.qa.suse.de/tests/292#

 - **trento** : test main scheduling. Test take care to deploy Trento on
   Azure and perform some basic test on the web front-end.

Execution example: http://1c242.qa.suse.de/tests/192

PR #15049, opened 2022-06-10
Move all common code about Trento tests to a lib

mpagot:
Better cypress.io integration by starting tests one by one
fail hook of the deploy_trento takes care to record all cluster logs

- Verification run: 
--- trento_jumphost http://1c242.qa.suse.de/tests/399  -  http://1c242.qa.suse.de/tests/439
--- trento http://1c242.qa.suse.de/tests/400  -  http://1c242.qa.suse.de/tests/441

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -4,8 +4,27 @@
 # SPDX-License-Identifier: FSFAP
 #
 # Summary: Functions for Trento tests
+# Maintainer: QE-SAP <qe-sap@suse.de>
 
 ## no critic (RequireFilenameMatchesPackage);
+
+=encoding utf8
+
+=head1 NAME
+
+Trento test lib
+
+=head1 COPYRIGHT
+
+Copyright 2017-2020 SUSE LLC
+SPDX-License-Identifier: FSFAP
+
+=head1 AUTHORS
+
+QE SAP <qe-sap@suse.de>
+
+=cut
+
 package trento;
 
 use strict;
@@ -13,11 +32,10 @@ use warnings;
 use testapi;
 use mmapi 'get_current_job_id';
 use utils qw(script_retry);
-use Carp qw(croak);
+use File::Basename qw(basename);
 use Exporter 'import';
 
 our @EXPORT = qw(
-  TRENTO_AZ_PREFIX
   get_resource_group
   get_vm_name
   get_acr_name
@@ -25,58 +43,65 @@ our @EXPORT = qw(
   VM_USER
   SSH_KEY
   cypress_install_container
-  PODMAN_PULL_LOG
   CYPRESS_LOG_DIR
+  PODMAN_PULL_LOG
   cypress_exec
   cypress_test_exec
   cypress_log_upload
 );
 
-=head1 SYNOPSIS
-Package with common methods and default values for Trento tests
+# Exported constants
+use constant VM_USER => 'cloudadmin';
+use constant SSH_KEY => '/root/.ssh/id_rsa';
+use constant CYPRESS_LOG_DIR => '/root/result';
+use constant PODMAN_PULL_LOG => '/tmp/podman_pull.log';
 
-=cut
-
+# Lib internal constants
 use constant TRENTO_AZ_PREFIX => 'openqa-trento';
-# Parameter 'registry_name' must conform to the following pattern: '^[a-zA-Z0-9]*$'.
+# Parameter 'registry_name' must conform to
+# the following pattern: '^[a-zA-Z0-9]*$'.
+# Azure does not support dash or underscore in ACR name
 use constant TRENTO_AZ_ACR_PREFIX => 'openqatrentoacr';
+use constant CYPRESS_IMAGE_TAG => 'goofy';
 
+=head1 DESCRIPTION 
 
-=head2 get_resource_group
+Package with common methods and default or constant  values for Trento tests
+
+=head2 Methods
+
+=head3 get_resource_group
+
 Return a string to be used as cloud resource group.
 It contains the JobId
-
 =cut
 sub get_resource_group {
     my $job_id = get_current_job_id();
     return TRENTO_AZ_PREFIX . "-rg-$job_id";
 }
 
+=head3 get_vm_name
 
-=head2 get_vm_name
 Return a string to be used as cloud VM name.
 It contains the JobId
-
 =cut
 sub get_vm_name {
     my $job_id = get_current_job_id();
     return TRENTO_AZ_PREFIX . "-vm-$job_id";
 }
 
+=head3 get_acr_name
 
-=head2 get_acr_name
 Return a string to be used as cloud ACR name.
 It contains the JobId
-
 =cut
 sub get_acr_name {
     return TRENTO_AZ_ACR_PREFIX . get_current_job_id();
 }
 
+=head3 az_get_vm_ip
 
-=head2 az_get_vm_ip
-Return the running VM IP
-
+Return the running VM public IP
 =cut
 sub az_get_vm_ip {
     my $az_cmd = 'az vm show -d ' .
@@ -87,10 +112,9 @@ sub az_get_vm_ip {
     return script_output($az_cmd, 180);
 }
 
+=head3 az_delete_group
 
-=head2 az_delete_group
-Delere resource group and so all its content
-
+Delete the resource group associated to this JobID and all its content
 =cut
 sub az_delete_group {
     script_run('echo "Delete all resources"');
@@ -100,27 +124,27 @@ sub az_delete_group {
     script_retry($az_cmd, timeout => 600, retry => 5, delay => 60);
 }
 
+=head3 az_vm_ssh_cmd
 
-use constant VM_USER => 'cloudadmin';
-use constant SSH_KEY => '/root/.ssh/id_rsa';
+Compose I<ssh> command for remote execution on the VM machine.
+The function optionally accept the VM public IP.
+If not provided, the IP is calculated on the fly with an I<az> query,
+take care that is a time consuming I<az> query.
 
+=over 2
 
-=head2 az_vm_ssh_cmd
-Compose ssh command for remote ssh execution on the VM machine
-The function optionally accept the vm ip.
-If it is not provided it is calculated on the fly with an az query,
-take care that is a time consuming az query.
+=item B<CMD_ARG> - String of the command to be executed remotely
+
+=item B<VM_IP_ARG> - Public IP of the remote machine where to execute the command
+
+=back
 =cut
 sub az_vm_ssh_cmd {
     my ($self, $cmd_arg, $vm_ip_arg) = @_;
-    my $vm_ip;
     record_info('REMOTE', "cmd:$cmd_arg");
-    if (defined $vm_ip_arg) {
-        $vm_ip = $vm_ip_arg;
-    }
-    else {
-        $vm_ip = az_get_vm_ip();
-    }
+
+    # undef comparision operator
+    my $vm_ip = $vm_ip_arg // az_get_vm_ip();
     return 'ssh' .
       ' -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR' .
       ' -i ' . SSH_KEY . ' ' .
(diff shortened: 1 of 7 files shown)


dzedro:
I had similar issue and should run `make tidy-full`, don't know if it will help, I didn't finish the PR.

mpagot:
@dzedro  thanks, tested `make tidy-full` locally. Unfortunately does not seems to help. It produce no result and does not change any file. I also give a try to various `tools/tidy` script invocation.

perlpunk:
I tried this PR locally and also did not get any changes from `make tidy-full`.
Unfortunately the output is not very informative. It would be good to know which version of Perl::Tidy it uses.

perlpunk:
Ah. I just rebased the PR on the current master and now I get changes.
So try to rebase it =)

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+    my $cypress_run_cmd = 'podman run ' .
+      '-it --name goofy ' .
+      "-v " . CYPRESS_LOG_DIR . ":/results " .
+      "-v $cypress_test_dir:/e2e -w /e2e " .
+      '-e "DEBUG=cypress:*" ' .
+      '--entrypoint=\'[' .
+      '"/bin/sh", "-c", ' .
+      ' "/usr/local/bin/cypress ' . $cmd .
+      ' 2>/results/cypress_' . $log_prefix . '_log.txt"' .
+      ']\' ' . $container_name .
```
alvarocarvajald, line 257: Be careful with the terminology. If names in podman are the same as in docker, then `$container_name` is actually the **Image Name** and what you are passing to the command via `--name` is the **Container Name**. (I could be wrong though ... haven't used podman yet)

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+Execute a cypress test 
+
+=cut
+sub cypress_test_exec {
+    my ($self, $cypress_test_dir, $test_tag, $timeout, $failok) = @_;
+    my $ret = 0;
+
+    my $test_file_list = script_output("find $cypress_test_dir/cypress/integration/$test_tag -type f -iname \"*.js\"");
+
+    my @arr;
```
alvarocarvajald, line 282: `@arr` is not used in the function.

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+    for (split(/\n/, $test_file_list)) {
+        my $test_filename = +(split /\//, $_)[-1];
```
alvarocarvajald, line 284: Suggestion:
```
use File::Basename qw(basename);
...
my $test_filename = basename($_);
```

tests/sles4sap/trento/test_trento_deploy.pm
```diff
@@ -8,33 +8,49 @@ use Mojo::Base 'publiccloud::basetest';
...
+            my $pod_name = (split /\s/, $row)[0];
+            my $trento_ver_cmd = $self->az_vm_ssh_cmd("kubectl exec --stdin $pod_name -- /app/bin/trento version", $machine_ip);
+            script_run($trento_ver_cmd, 180);
+        }
+    }
 
     # test if the web page is reachable on http
-    assert_script_run("curl -k  http://" . $machine_ip . "/");
+    my $trento_url = 'http://' . $machine_ip . '/';
+    script_run('curl --version');
```
alvarocarvajald, line 38: Is this really needed?

alvarocarvajald (commented):
Very nice PR. It addresses some of my questions from #14951 & #15019.

Thanks for adding POD documentation to `lib/trento.pm`. Once this is merged, it will be available in https://os-autoinst.github.io/os-autoinst-distri-opensuse/.

Did you check with `perldoc -U ./lib/trento.pm` or with `pod2man ./lib/trento.pm | nroff -man | less` for the formatting of the documentation? Documentation is there, and seems good enough in the output of `perldoc -U`, but if you compare it to the other libraries (for example `lib/utils.pm` or `lib/sles4sap.pm`) you'll see some differences. I suppose when it's rendered for the web page, it'll look better, but I am adding the comment here just in case you want to re-format for command line.

I have added more comments inline.

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+=head2 k8s_logs
+Get all relevan tinfo out from the cluster
+=cut
+sub k8s_logs {
+    my $self = shift;
+    my (@pods_list) = @_;
+    my $machine_ip = az_get_vm_ip;
+
+    if ($machine_ip ne "") {
+        my $kubectl_pods = script_output($self->az_vm_ssh_cmd('kubectl get pods', $machine_ip), 180);
```
mpagot, line 141: code to be improved as it could result in teardown to fails http://1c242.qa.suse.de/tests/383#step/deploy_trento/28  fails as bash: kubectl: command not found  due to fact that teardown code try to collect cluster logs but the cluster is not there yet.

mpagot:
> BTW, is this error expected?
> 
> http://1c242.qa.suse.de/tests/400#step/deploy_trento/52

Error is at http://1c242.qa.suse.de/tests/400#step/deploy_trento/50
```
[E:1][R] Trento Server start takes more then 3 minutes
         or not running at all. Exiting.
```

This failure does not depends on this perl test code

FYI @jankohoutek

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+  cypress_log_upload
+);
+
+=head1 SYNOPSIS
+Package with common methods and default values for Trento tests
+
+=cut
+
+use constant TRENTO_AZ_PREFIX => 'openqa-trento';
+# Parameter 'registry_name' must conform to the following pattern: '^[a-zA-Z0-9]*$'.
```
dzedro, line 41: I don't understand this comment. :thinking:

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+use constant TRENTO_AZ_ACR_PREFIX => 'openqatrentoacr';
+
+
```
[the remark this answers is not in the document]
mpagot: Probably inspired by https://peps.python.org/pep-0008/#blank-lines
Any suggestion for perl code in this context?

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+    my $ret = 0;
+
+    record_info('CY EXEC', 'Cypress exec:' . $cmd);
+    my $cypress_ver = get_var('TRENTO_CYPRESS_VERSION', '4.4.0');
+    my $container_name = "docker.io/cypress/included:$cypress_ver";
+    # container is executed with --name to easy the log retrieve
+    # to do so we need to rm eventually present container with that name
+    script_run('podman rm goofy || echo "No Goofy to delete"');
+    my $cypress_run_cmd = 'podman run ' .
+      '-it --name goofy ' .
```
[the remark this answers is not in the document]
mpagot: I'll create a constant but probably not a variable. It is really just a constant and I do not see any additional value to let the "user" to configure it.

lib/trento.pm
```diff
@@ -0,0 +1,300 @@
...
+    if ($ret != 0) {
+        record_soft_failure("Cypress exit code:$ret at $log_prefix");
+        # look for SIGTERM
+        script_run('podman logs -t goofy');
+        $self->result("fail");
+    }
+    if ($failok) {
+        $ret = 0;
+    }
+    croak "Cypress exec error at '$cmd'" unless ($ret == 0);
```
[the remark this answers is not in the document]
mpagot: not sure but it was after reading this definition : https://www.perlmonks.org/?node_id=685452 
what do you think?

tests/sles4sap/trento/test_trento_deploy.pm
mpagot: yes, I like to keep track of it. I was not able to  use a better curl arguments combination only available in a newer version. So I feel that I like to keep track of the running curl version.

lib/trento.pm
mpagot: prefix is like **openqa-trento** with a dash in the middle. Most of the names are composed from this one. It is readable thanks to the dash.
Unfortunately the Azure ACR service does not support dash or underscore in the names, so only letter and number, that result in a less readable string like the one for **TRENTO_AZ_ACR_PREFIX**. The comment is a justification about the not so readable name. It is due to an Azure requirement.

lib/trento.pm
```diff
@@ -0,0 +1,371 @@
...
+use warnings;
+use testapi;
+use mmapi 'get_current_job_id';
+use utils qw(script_retry);
+use File::Basename qw(basename);
+use Carp qw(croak);
+use Exporter 'import';
+
+our @EXPORT = qw(
+  TRENTO_AZ_PREFIX
```
dzedro, line 40: You don't need to export constant or ? I think it should be not done.
https://perldoc.perl.org/Exporter#What-Not-to-Export

lib/trento.pm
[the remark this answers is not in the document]
alvarocarvajald: > Probably inspired by https://peps.python.org/pep-0008/#blank-lines Any suggestion for perl code in this context?

Oh, nice. Did not know about that convention for python. I am not aware of any sub convention for Perl, but looking some libraries at a glance inside os-autoinst-distri-opensuse I don't see anything similar.

lib/trento.pm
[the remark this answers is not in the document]
alvarocarvajald: IMO, looking at that perlmonks.org article, this is a case of the call failing due to an issue within the method and not due to how it was called, so I would go with `die`, and that way you don't need to do `use Carp qw(croak)` either.

alvarocarvajald:
> > Great PR! Please pay attention to the following items before merging:
> > **Files matching `lib/**.pm`:**
> > 
> > * [ ]   Consider adding or extending unit tests in t/
> > 
> > This is an automatically generated QA checklist based on modified files
> 
> Please @alvarocarvajald suggest to me some ideas about possible trento.pm unit tests please

Usually with libraries, one tries to write tests for all the methods defined in the library, but since this library has methods that perform installation and execution of podman images and containers, I am not really sure how to proceed. One idea is to look at `t/08-publiccloud.t` for some examples. It seems some of the tests there deal only with settings which is something you can do for the get* methods in `lib/trento.pm`

lib/trento.pm
dzedro: I thought you don't need the constant out of `lib/trento.pm`
It's not explicitly bad, but it should be not done. You can do it if you like I think it's not first case of exported "variable/constant" :man_shrugging:

dzedro: btw define constants on one place, there is another definition of VM_USER at line 123

perlpunk: if the constant is not needed in a different file, don't export it.
And if you export something, prefer `@EXPORT_OK` over `@EXPORT` to only import on request.

lib/trento.pm
```diff
@@ -0,0 +1,369 @@
...
+    assert_script_run('podman --version');
+    assert_script_run('podman info --debug');
+    assert_script_run('podman ps');
+    assert_script_run('podman images');
+    assert_script_run('df -h');
+}
+
+use constant PODMAN_PULL_LOG => '/tmp/podman_pull.log';
+
+=head3 crypress_install_container
```
alvarocarvajald, line 218: s/crypress/cypress/

lib/trento.pm
```diff
@@ -0,0 +1,369 @@
...
+=item B<VM_IP_ARG> - Public IP of the remote machine where to execute the command
+
+=back
+=cut
+sub k8s_logs {
+    my $self = shift;
+    my (@pods_list) = @_;
+    my $machine_ip = az_get_vm_ip;
+
+    if ($machine_ip) {
```
alvarocarvajald, line 171: I think you can re-write this as `return unless ($machine_ip);` since there is no else in this if-block and you would remove a full 4-spaces indentation below as well as some brackets. It would improve readibility IMO.

lib/trento.pm
```diff
@@ -0,0 +1,369 @@
...
+        # execute the test: force $failok=1 to keep the execution going.
+        # Any cypress test failure will be reported during the XUnit parsing
+        $self->cypress_exec($cypress_test_dir, $test_cmd, $timeout, $test_tag, 1);
+
+        # parsing the results
+        my $find_cmd = 'find ' . $self->CYPRESS_LOG_DIR . ' -type f -iname "' . $test_result . '"';
+        parse_extra_log("XUnit", $_) for split(/\n/, script_output($find_cmd));
+
+        # upload all logs at once
+        $self->cypress_log_upload(('.txt', '.mp4'));
```
perlpunk, line 365: double braces
```suggestion
        $self->cypress_log_upload('.txt', '.mp4');
```

tests/sles4sap/trento/deploy_trento.pm
```diff
@@ -8,68 +8,74 @@ use Mojo::Base 'publiccloud::basetest';
...
+    upload_logs($deploy_script_log);
 }
 
 sub post_fail_hook {
     my ($self) = @_;
-    my $job_id = get_current_job_id();
-    my $resource_group = TRENTO_AZ_PREFIX . "-rg-$job_id";
 
-    assert_script_run("az group delete --resource-group $resource_group --yes", 180);
+    $self->k8s_logs(('web', 'runner'));
```
perlpunk, line 76:
```suggestion
    $self->k8s_logs(qw(web runner));
```

tests/sles4sap/trento/setup_jumphost.pm
```diff
@@ -78,8 +48,8 @@ sub run {
     assert_script_run("git clone $gitlab_clone_cmd . | tee " . GITLAB_CLONE_LOG);
 
     # Cypress.io installation
-    my $cypress_ver = get_var('TRENTO_CYPRESS_VERSION', '3.4.0');
-    cypress_install_container($cypress_ver);
+    my $cypress_ver = get_var('TRENTO_CYPRESS_VERSION', '4.4.0');
```
perlpunk, line 51: That's a general suggestion:
```suggestion
    my $cypress_ver = get_var(TRENTO_CYPRESS_VERSION => '4.4.0');
```
If you use the "fat comma" `=>`, then you don't need to quote the string before.

tests/sles4sap/trento/test_trento_deploy.pm
```diff
@@ -8,33 +8,49 @@ use Mojo::Base 'publiccloud::basetest';
...
+        if ($row =~ m/trento-server-web/) {
+            my $pod_name = (split /\s/, $row)[0];
+            my $trento_ver_cmd = $self->az_vm_ssh_cmd("kubectl exec --stdin $pod_name -- /app/bin/trento version", $machine_ip);
+            script_run($trento_ver_cmd, 180);
+        }
+    }
 
     # test if the web page is reachable on http
-    assert_script_run("curl -k  http://" . $machine_ip . "/");
+    my $trento_url = 'http://' . $machine_ip . '/';
```
perlpunk, line 37: Also a general suggestion:
```suggestion
    my $trento_url = "http://$machine_ip/";
```
I personally find interpolation more readable in many cases, where it's possible.
For longer concatenations I can also recommend to use [sprintf](https://perldoc.perl.org/functions/sprintf)

lib/trento.pm
mpagot: There are some test files like **deploy_trento.pm** that are using many symbols from lib/trento.pm

Does not changing the overall list of exported symbols from EXPORT to EXPORT_OK result in a long and so difficult to maintain list of symbols in the test files?

Does it make sense to have two export lists, one using EXPORT and the other one using EXPORT_OK?

alvarocarvajald (approved):
LGTM

dzedro (approved):

lib/trento.pm
```diff
@@ -0,0 +1,366 @@
...
+
+=item B<TIMEOUT> - Integer used as timeout for the cypress command execution
+
+=item B<LOG_PREFIX> - String of the command to be executed remotely
+
+=item B<FAILOK> - Integer boolean value. 0:test marked as failure if the podman/cypress
+return not 0 exit code. 1:all not 0 podman/cypress exit code are ignored. SoftFail reported.
+
+=back
+=cut
```
perlpunk, line 284: I didn't see this earlier, but always leave a newline between pod and code. Some pod renderers will see this as malformed.

alvarocarvajald: I didn't know about that, but I've always instinctively left empty lines surrounding all pod command as in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/hacluster.pm or https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/sles4sap.pm (my main contributions to pod documentation in this repo).

@mpagot it seems it would be a good idea to apply Tina's suggestion even though most libraries within os-autoinst-distri-opensuse are not following it (examples: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm, https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/opensusebasetest.pm or https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm).

mpagot: I play a little bit with it locally using `pod2html` exactly about this topic: "Where new line is needed"
For example new line is strictly needed after =head and =index

Ignoring for a second what is strictly needed for a proper rendering, and just focusing of the generic code readability: leaving one empty like between pod and code is maybe fine but somehow collide (personal opinion) with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15049#discussion_r895862427  (leaving 2 new lines between each method).

alvarocarvajald:
I will merge this after Michele's last comment and the fact that some existing libraries in the repository do not leave an empty line between pod and code. I guess we should create a progress ticket to track the reformatting of the pod documentation in the libraries to apply Tina's suggestion.

alvarocarvajald:
> I will merge this after Michele's last comment and the fact that some existing libraries in the repository do not leave an empty line between pod and code. I guess we should create a progress ticket to track the reformatting of the pod documentation in the libraries to apply Tina's suggestion.

I created https://progress.opensuse.org/issues/112544 to follow up on this.

PR #15052, opened 2022-06-10
Add workaround for race condition between agetty and logind

mloviska:
As described in https://github.com/systemd/systemd/issues/23660 or [Bug 1185098 - [Build 15.82]
openQA test fails in snapper_undochange - logind fails after return back from rescue](https://bugzilla.suse.com/show_bug.cgi?id=1185098)
we need to add workaround that ensures `tty2` will be available for login.
Issue occurs only in systems running sle15sp2 or below.

VRs:
    - [workaround case](http://kepler.suse.cz/tests/17406#step/snapper_undochange/55)
    - [expected](http://kepler.suse.cz/tests/17405#step/snapper_undochange/42)

### DIFF
--- a/lib/btrfs_test.pm
+++ b/lib/btrfs_test.pm
@@ -10,6 +10,8 @@ use warnings;
 use testapi;
 use utils 'get_root_console_tty';
 use Exporter 'import';
+use version_utils qw(is_sle);
+use Utils::Systemd qw(systemctl);
 
 our @EXPORT_OK = qw(set_playground_disk cleanup_partition_table);
 
@@ -72,6 +74,17 @@ sub snapper_nodbus_restore {
     die 'DBus service should be inactive, but it is active' if ($ret == 0);
     script_run('systemctl default', timeout => 600, die_on_timeout => 0);
     my $tty = get_root_console_tty;
+
+    if (is_sle('<15-SP3') && !defined(my $match = check_screen("tty$tty-selected", 120))) {
+        record_soft_failure("bsc#1185098 - logind fails after return back from rescue");
+        select_console('log-console');
+        if (script_run('systemctl is-active getty@tty2.service', die_on_timeout => 1) == 3) {
+            systemctl('start getty@tty2');
+            reset_consoles;
+        }
+        select_console('root-console');
+    }
+
     assert_screen "tty$tty-selected", 600;
     reset_consoles;
     select_console 'root-console';


lib/btrfs_test.pm
```diff
@@ -72,6 +74,17 @@ sub snapper_nodbus_restore {
     die 'DBus service should be inactive, but it is active' if ($ret == 0);
     script_run('systemctl default', timeout => 600, die_on_timeout => 0);
     my $tty = get_root_console_tty;
+
+    if (is_sle('<15-SP3') && !defined(my $match = check_screen("tty$tty-selected", 120))) {
+        record_soft_failure("bsc#1185098 - [Build 15.82] openQA test fails in snapper_undochange - logind fails after return back from rescue");
```
jlausuch, line 79: nitpicking, no need to provide full title of the bug, but feel free to reject :)
```suggestion
        record_soft_failure("bsc#1185098 - logind fails after return back from rescue");
```

lib/btrfs_test.pm
```diff
@@ -72,6 +74,17 @@ sub snapper_nodbus_restore {
...
+        select_console('log-console');
+        if (script_run('systemctl is-active getty@tty2.service') == 3) {
```
jlausuch, line 81: We should have the best practice of `script_run` together with `die_on_timeout` flag.
https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L1010

```suggestion
        if (script_run('systemctl is-active getty@tty2.service', die_on_timeout => 1) == 3) {
```

lib/btrfs_test.pm
mloviska: Nah, I that was just my lazy copy&paste. Updated :)

lib/btrfs_test.pm
mloviska: Nice catch, I always forget about this option. Thanks!

mloviska: http://kepler.suse.cz/tests/17407
http://kepler.suse.cz/tests/17408

lib/btrfs_test.pm
```diff
@@ -72,6 +74,17 @@ sub snapper_nodbus_restore {
     die 'DBus service should be inactive, but it is active' if ($ret == 0);
     script_run('systemctl default', timeout => 600, die_on_timeout => 0);
     my $tty = get_root_console_tty;
+
+    if (is_sle('<15-SP3') && !defined(my $match = check_screen("tty$tty-selected", 120))) {
+        record_soft_failure("bsc#1185098 - logind fails after return back from rescue");
```
jlausuch, line 79: Also, Oli had a valid point about WON'T FIX bugs, this seems to be one of them. By doing soft-failure, our jobs will be always soft-failed and will keep posting comments to the ticket, I wonder if we should just use record_info...

mloviska: That is a valid point, we can change that when there will be some response from the upstream. As of now, there are no comments, so maybe they will consider fixing it.

jlausuch: Makes sense.

jlausuch (approved):

PR #15060, opened 2022-06-13
Public Cloud: use base URL and namespace variables

jlausuch:
To allow flexibility when we want to run modules using different providers in the same job, this approach makes this possible by hardcoding the sufix of the URL in the provider and use the var `PUBLIC_CLOUD_CREDENTIALS_URL` for the base url and `PUBLIC_CLOUD_NAMESPACE` to differentiate between different accounts.

### DIFF
--- a/lib/publiccloud/gcp_client.pm
+++ b/lib/publiccloud/gcp_client.pm
@@ -37,7 +37,7 @@ sub init {
     my ($self) = @_;
     # For now we support Vault and the credentials-microservice. Vault will be removed after a certain transition period
     if (get_var('PUBLIC_CLOUD_CREDENTIALS_URL')) {
-        my $data = get_credentials(CREDENTIALS_FILE);
+        my $data = get_credentials('gce.json', CREDENTIALS_FILE);
         $self->project_id($data->{project_id});
         $self->account($data->{client_id});
     } else {
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -217,10 +217,13 @@ sub define_secret_variable {
 # Get credentials from the Public Cloud micro service, which requires user
 # and password. The resulting json will be stored in a file.
 sub get_credentials {
-    my ($output_json) = @_;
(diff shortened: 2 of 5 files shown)


lib/publiccloud/gcp_client.pm
```diff
@@ -37,7 +37,7 @@ sub init {
     my ($self) = @_;
     # For now we support Vault and the credentials-microservice. Vault will be removed after a certain transition period
     if (get_var('PUBLIC_CLOUD_CREDENTIALS_URL')) {
-        my $data = get_credentials(CREDENTIALS_FILE);
+        my $data = get_credentials('gce.json', CREDENTIALS_FILE);
```
pdostal, line 40: Why is this `CREDENTIALS_FILE` missing on the other clients?

pdostal (approved):

lib/publiccloud/gcp_client.pm
jlausuch: Because it's not needed. Only Google needs a phyisical file to read the credentials from, the others are use environment variables.

ilausuch (approved):

grisu48 (approved):
LGTM

lib/publiccloud/utils.pm
```diff
@@ -217,10 +217,13 @@ sub define_secret_variable {
 # Get credentials from the Public Cloud micro service, which requires user
 # and password. The resulting json will be stored in a file.
 sub get_credentials {
-    my ($output_json) = @_;
-    my $url = get_required_var('PUBLIC_CLOUD_CREDENTIALS_URL');
+    my ($url_sufix, $output_json) = @_;
+    my $base_url = get_required_var('PUBLIC_CLOUD_CREDENTIALS_URL');
+    my $namespace = get_required_var('PUBLIC_CLOUD_NAMESPACE');
```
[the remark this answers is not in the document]
jlausuch: Yes, actually `PUBLIC_CLOUD_CREDENTIALS_URL` was not there. Both added.

PR #15063, opened 2022-06-13
Azure CLI: make bsc#1192671 soft

jlausuch:
We are not running this test at all, and the only issue that the CLI
has is this bug. The rest could be run.
Also, by enabling the test again, we will be able to verify a possible
fix for it.

VR: https://openqa.suse.de/tests/8950812#


grisu48 (approved):
It is sad that we have to do this, but packaging azure-cli is known to be difficult.

LGTM.

jlausuch:
Changed to Draft, as the next module fails when enabling Azure cli module... what the heck...

PR #15064, opened 2022-06-13
Fix usage of certain methods in test repo

foursixnine:
- Containers, Public Cloud: Get rid of multiple arguments in record_soft_failure
- Fix wrong usages of signatures
- Fix usage of certain methods

PR #15068, opened 2022-06-13
Followup to #15064

foursixnine:
- Fix wrong usage of check_var
VR: https://openqa.suse.de/tests/8946426#

### DIFF
--- a/data/filesystem/autoyast_filesystem_withouthome.xml.ep
+++ b/data/filesystem/autoyast_filesystem_withouthome.xml.ep
@@ -6,7 +6,7 @@
     <email/>
     <reg_code>{{SCC_REGCODE}}</reg_code>
     <install_updates config:type="boolean">true</install_updates>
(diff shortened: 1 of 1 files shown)


data/filesystem/autoyast_filesystem_withouthome.xml.ep
```diff
@@ -6,7 +6,7 @@
     <email/>
     <reg_code>{{SCC_REGCODE}}</reg_code>
     <install_updates config:type="boolean">true</install_updates>
-    % unless ($check_var->('FLAVOR', '') =~ /-Updates$|-Incidents/) {
+    % unless ($get_var->('FLAVOR') =~ /-Updates$|-Incidents/) {
```
okurz, line 9: Actually it's still good to use `get_var($variable, $default)` so here it would be `$get_var->('FLAVOR', '')` to prevent a warning about comparing undef by regex in case flavor is not defined.

PR #15070, opened 2022-06-13
Revert "Workaround license insert disc issue"

mdoucha:
This reverts commit a6ffe9569133236e848a9ea0a6a3a63424c4bbde.

Remove old workaround for bsc#1153326. The issue has been fixed in YaST a long time ago and the workaround needles have last matched 3 years ago.

If the issue appears again in the future, the tests should fail so that reviewers can file a new bug report. The needle check is done twice during installation and it adds unnecessary 30 second delay each time.

- Related ticket: https://progress.opensuse.org/issues/112343
- Needles: N/A
- Verification run: https://openqa.suse.de/t8947255

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -107,10 +107,6 @@ sub is_module {
 sub accept_addons_license {
     my (@scc_addons) = @_;
 
-    # Trap the 'missing license file on media' issue
-    # Needle is configured as a workaround, so a soft-fail will be shown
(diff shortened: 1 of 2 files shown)


okurz (approved):
Thank you a lot. Unfortunately people keep coming back to introduce `check_screen` when they shouldn't.

dzedro:
> Thank you a lot. Unfortunately people keep coming back to introduce `check_screen` when they shouldn't.

It's ~3 years old commit ...

PR #15071, opened 2022-06-13
Run slurm configuration merely against compute nodes

b10n1k:
The operations which try to copy the ssh keys and the slurm configuration
consider all the slave(compute) nodes including the master(management) one.
However i think this is not necessary. In this commit i change this to run
merely against the compute nodes.

For some reason the behavior of those operations started failing, when the
where trying to copy data from master to itself. I didnt dig into it but this
shouldnt be like that anyway IMO. Anyway merging this seems to resolve the
problem.

- Verification run:

before:  https://openqa.suse.de/tests/8946387#
after: http://aquarius.suse.cz/tests/10944#

https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2315071&distri=sle&version=15-SP2

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -117,26 +117,26 @@ sub cluster_names {
 
 =head2 distribute_munge_key
 
-Distributes munge kyes across all cluster nodes
+Distributes munge keys across all compute nodes of the cluster
 
 =cut
 sub distribute_munge_key {
     my ($self) = @_;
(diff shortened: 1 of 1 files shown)


b10n1k:
This solve a number of slurm suites. i am checking the rest atm

b10n1k:
i am going to merge this and open a progress ticket for the other issues

PR #15074, opened 2022-06-14
Fix duplicate docker.io registry

grisu48:
docker.io will not be added to registries.conf if defined via the
REGISTRY variable because it's already present there with conflicting
insecure settings. This makes test runs succeed, also if no REGISTRY is
defined.

- Related ticket: https://progress.opensuse.org/issues/112370
- Verification run: https://duck-norris.qam.suse.de/t8935

### DIFF
--- a/lib/containers/podman.pm
+++ b/lib/containers/podman.pm
@@ -23,11 +23,11 @@ sub init {
 
 sub configure_insecure_registries {
     my ($self) = shift;
-    my $registry = registry_url();
 
     assert_script_run "curl " . data_url('containers/registries.conf') . " -o /etc/containers/registries.conf";
     assert_script_run "chmod 644 /etc/containers/registries.conf";
-    file_content_replace("/etc/containers/registries.conf", REGISTRY => $registry);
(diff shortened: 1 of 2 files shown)


lib/containers/podman.pm
```diff
@@ -27,7 +27,8 @@ sub configure_insecure_registries {
 
     assert_script_run "curl " . data_url('containers/registries.conf') . " -o /etc/containers/registries.conf";
     assert_script_run "chmod 644 /etc/containers/registries.conf";
-    file_content_replace("/etc/containers/registries.conf", REGISTRY => $registry);
+    # Add custom registry only if it is not docker.io, which is already present in registries.conf and docker.io is the default return value for get_var('REGISTRY')
```
jlausuch, line 30:
```suggestion
    # Add custom registry only if it is not docker.io, which is already present in registries.conf and docker.io is the default return value for registry_url();
```

grisu48: The whole comment got replaced in the meantime, following your previous comment

jlausuch (approved):

PR #15077, opened 2022-06-14
Migrate Kubernetes GKE tests to new accounts

ilausuch:
Fix Google Cluster client for new account

- Related ticket:  https://progress.opensuse.org/issues/111983
- Verification run: http://marvin5.arch.suse.cz/tests/361

### DIFF
--- a/lib/publiccloud/gcp_client.pm
+++ b/lib/publiccloud/gcp_client.pm
@@ -44,7 +44,7 @@ sub init {
         $self->vault(publiccloud::vault->new());
         $self->create_credentials_file();
     }
-    assert_script_run('source ~/.bashrc');
+    assert_script_run("export GOOGLE_APPLICATION_CREDENTIALS=" . CREDENTIALS_FILE);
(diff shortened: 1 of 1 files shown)


pdostal (approved):

grisu48:
If someone else is also wondering why it is `GOOGLE_APPLICATION_CREDENTIALS`, see https://cloud.google.com/docs/authentication/getting-started

lib/publiccloud/gcp_client.pm
```diff
@@ -44,7 +44,7 @@ sub init {
         $self->vault(publiccloud::vault->new());
         $self->create_credentials_file();
     }
-    assert_script_run('source ~/.bashrc');
+    assert_script_run("export GOOGLE_APPLICATION_CREDENTIALS=" . CREDENTIALS_FILE);
```
grisu48, line 47: Not for this PR but this constant should be renamed to `$CREDENTIALS_FILE`, as it is otherwise considered a Bareword and this would fail if we would use `strict subs` (which we should).

grisu48 (approved):
LGTM!

jlausuch:
Draft. Can't merge yet

ilausuch:
Not necessary

PR #15080, opened 2022-06-14
Draft: Add local rmt registration option and popup

BillAnastasiadis:
This pr provides a new module to enable users to register using local rmt server instead of scc. It also handles the security popup that shows up when using https, or dies if it doesn't show up (i.e. if using http instead of https)

- Related ticket: https://progress.opensuse.org/issues/112058
- Verification run: https://openqa.suse.de/tests/8953433


tests/installation/registration/register_via_rmt.pm
```diff
@@ -0,0 +1,23 @@
...
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    $testapi::distri->get_registration()->register_via_rmt({
+            server => get_var('RMT_SERVER')});
```
jknphy, line 18: we can harcode it, it is exposed everywhere, progress ticket, bugs, etc, beside that we avoid to use settings as much as possible, otherwise the setting needs to be documented in `variables.md`, but in this case doesn't worth it.

lib/Installation/Registration/RegistrationPage.pm
```diff
@@ -26,6 +26,8 @@ sub init {
     $self->{rdb_skip_registration} = $self->{app}->radiobutton({id => 'skip_registration'});
     $self->{txb_email} = $self->{app}->textbox({id => 'email'});
     $self->{txb_reg_code} = $self->{app}->textbox({id => 'reg_code'});
+    $self->{rdb_rmt_server_registration} = $self->{app}->radiobutton({id => 'register_local'});
```
jknphy, line 29: what about `rdb_rmt_server`, it is a bit shorter and rmt is all above registration, should we could omit it.

jknphy (commented):
Initial round of review. It is looking good, just some minor things I saw.
You didn't use regurl pointing to that server, we could catch the issue before, but I think it is fine, you did it in the place where the bug happened :)
We need to perform the full installation and validate those repos after installation. And once that is done optionally we could try the default+flow approach for scheduling.

lib/Installation/Registration/RegistrationPage.pm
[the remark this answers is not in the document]
BillAnastasiadis: Done! Since I'm here, why not also get rid of code duplication

BillAnastasiadis: Done!

tests/installation/registration/register_via_rmt.pm
BillAnastasiadis: I can do that, but since this will be it's own testsuite, maybe it makes more sense to put it in a dedicated YAML file for that testcase instead of hardcoding it in the code? Mixing data and logic that much just doesn't sit very well with me. 
  
like this one for example https://github.com/BillAnastasiadis/os-autoinst-distri-opensuse/blob/cb0480bb229e030731d52fd8460cf2d121249477/schedule/yast/select_modules_and_patterns/select_modules_and_patterns.yaml# (but for the new testsuite, `register_with_rmt` or whatever it will be named)

Just wondering. If you don't like the idea, I'll just hardcode it and be done with it :p

jknphy: the problem is that there is a lack of control on those variables and it is better to not contribute to that mess. In order to not hardcode it the best solution is using testdata, you have seen probably in many test that approach, but in the past we noticed that it was unnecessary for this kind of things that are sort of constants. We also agreed in the past that on the layer Test doesn't hurt, at the level of Page or Controller would be a problem and not the best code style. If you please could hardcode it for now, perhaps all these ideas would make more sense to you in a while.

BillAnastasiadis:
> could you please also push the yaml file? (we cannot result that select_module_and_patterns, we need a new one) we aim here for a default installation:
> 
>     * we don't need to enter the guided setup: https://openqa.suse.de/tests/9004526#step/select_guided_setup/1
> 
>     * this step for home is not needed either: https://openqa.suse.de/tests/9004526#step/do_not_propose_separate_home/1
> 
>     * no pattern selection required.
> 
> 
> basically we shouldn't add anything else post-installation than the check of the URIs coming from this server (with `zypper lr --uri`), if you need test data for those uris perhaps instead of having as openQA setting you actually would need the server as test data. The idea I have from a time is that for steps on the installation is ok to hardcode stuff, but for the validation because there is more combinations, we can use test data, right now we have a mix of both approaches.

Sounds good. Let me also add repo validation and remove all unneeded post-installation steps, and we can look at it.

tests/installation/registration/register_via_rmt.pm
BillAnastasiadis: Maybe the testdata approach makes sense, containing the server and all the repos to be validated (as there is already a `validate_repos` function that uses testdata). I'm going for that and we can evaluate it. Thanks for the suggestions!

jknphy: ok, let's keep it then, thanks for your input @JRivrain, @BillAnastasiadis could you please document the variable in `variables.md`.

BillAnastasiadis: Ok, will do

coolgw: There exist an setting already used in current migration test cases, SMT_URL, this setting used to flag SMT or RMT url. 
But this setting seems not mentioned in the [variables.md]

I list one smt test case and one rmt test case for example:
https://openqa.suse.de/tests/8777824#step/yast2_migration/1
https://openqa.suse.de/tests/8776608#settings

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - installation/logs_from_installation_system
+  - installation/performing_installation/confirm_reboot
+  - installation/handle_reboot
+  - installation/first_boot
+  - console/system_prepare
+  - console/consoletest_setup
+  - console/validate_repos
+test_data:
+  repos:
+    - uri: /repo/SUSE/Products/SLE-Module-Basesystem/15-SP4/x86_64/product/?credentials=Basesystem_Module_x86_64
```
coolgw, line 39: Put data contain "x86_64" keyword in here maybe not a good idea, how to handle it if we start load this yaml on other platform such as 390?
I suggest just using "zypper lr" and check output with key word such as SLE-Module-Basesystem etc in code directly.
But i have no idea is feasible for libyui case or not, this is new thing to me :)

BillAnastasiadis: That's a valid point. I'll modify it so it gets the version and arch from the test vars.

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,48 @@
+---
+name: rmt_server_install
+description: |
+  Perform an installation enabling some modules and selecting some
```
jknphy, line 4: could we improve this description and describe the test case, installation using registration server RMT trusting and importing blablab and validation of ... for example.

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,48 @@
...
+schedule:
+  - installation/bootloader_start
+  - installation/setup_libyui
+  - installation/product_selection/install_SLES
+  - installation/licensing/accept_license
+  - installation/registration/register_via_rmt
+  - installation/module_registration/register_nonconflicting_modules
+  - installation/add_on_product/skip_install_addons
+  - installation/system_role/accept_selected_role_SLES_with_GNOME
+  - installation/partitioning/select_guided_setup
```
jknphy, line 17: we can remove all these test modules until `accept_proposed_layout` we don't need to change defaults or pass through the guided setup.

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - installation/partitioning/guided_setup/accept_default_part_scheme
+  - installation/partitioning/guided_setup/do_not_propose_separate_home
+  - installation/partitioning/accept_proposed_layout
+  - installation/clock_and_timezone/accept_timezone_configuration
+  - installation/authentication/use_same_password_for_root
+  - installation/authentication/default_user_simple_pwd
+  - installation/select_patterns
```
jknphy, line 24: we can remove this module as well, no pattern selection is needed.

tests/console/validate_repos.pm
```diff
@@ -19,16 +19,35 @@ sub run {
...
-        my $filter = $repo->{filter} ? $repo->{$repo->{filter}} : undef;
-        validate_repo_properties({
-                Filter => $filter,
-                Alias => $repo->{alias},
-                Name => $repo->{name},
-                URI => $repo->{uri},
-                Enabled => $repo->{enabled},
-                Autorefresh => $repo->{autorefresh}
-        });
+    if (check_var("RMT_SERVER", "default") || !get_var("RMT_SERVER")) {
```
jknphy, line 22: for sure we need to separate this in two test modules, we try to avoid this `if/else` at the end someone else would add yet another complicated condition to this line. I also not familiar with value 'default', not sure what you wanted to express there :)
do you know that you can use in test data expansion as well?
%ARCH% would it help to have just one module?

jknphy (commented):
The code looks almost done, only a some suggestions for splitting parts. Good job.
I just learnt yesterday that this test case would not work in Product validation when it is not a Milestone build. RMT server is only synchronized pulling new updates of packages for builds that are marked as milestones, so in any normal snapshots we will be testing packages from the previous milestone from that build. Migration team solves that triggering manually this kind of tests only for milestones, but for now it is fine we run it all the time, we will do some separation later.

tests/console/validate_repos.pm
BillAnastasiadis: I too find the creation of a new validation module a better idea. A few lines of similar code are better than hard to read code. Will move it to each own module.

jknphy: all that you use here seems to be just openQA settings `URI => join('', /$rmt_uri/, /$repo->{uri}/, /$ver/, /$sep/, /$arch/, /$repo->{suffix}/, /$ver/)` so there is not need to have this complex concatenation, we can add interpolation with %var% in `test_data` in yaml, then one test module is enough for the uri, right?

schedule/yast/rmt_server_install/rmt_server_install.yaml
jknphy: not inclusive, you still need to accept the proposed layout :)

tests/console/validate_repos.pm
BillAnastasiadis: I did not know that, in that case yes it should be enough. Let me rewrite the test with no new module and see if it works!

BillAnastasiadis: @jknphy I did follow your advise, however a change is needed in the `validate_repos` string for this to work. I suspect the existance of `/?` in the URI string causes this. Please see the change in the `validate_repos` pm, it was the only way I could get this to work.

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,40 @@
...
+schedule:
+  - installation/bootloader_start
+  - installation/setup_libyui
+  - installation/product_selection/install_SLES
+  - installation/licensing/accept_license
+  - installation/registration/register_via_rmt
+  - installation/module_registration/register_nonconflicting_modules
+  - installation/add_on_product/skip_install_addons
+  - installation/system_role/accept_selected_role_SLES_with_GNOME
+  - installation/partitioning
```
jknphy, line 17: these 2 modules are needle-based modules, you just need this sequence:
```
   - installation/system_role/accept_selected_role_SLES_with_GNOME 
   - installation/clock_and_timezone/accept_timezone_configuration
```

tests/console/validate_repos.pm
```diff
@@ -25,7 +25,7 @@ sub run {
                 Filter => $filter,
                 Alias => $repo->{alias},
                 Name => $repo->{name},
-                URI => $repo->{uri},
+                URI => join('', /$repo->{uri}/),
```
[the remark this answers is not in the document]
BillAnastasiadis: a) Yes, we do. `$repo->{uri}` and `/$repo->{uri}/` both fail (and they fail while they show that the string is actually the same), this join is the only way I could get it to work. I sincerely have no idea why.

b) Will do!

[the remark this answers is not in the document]
jknphy: a) probably because you are stringifying it internally with Perl Regex, but if you escape the `?` with `\?` should be fine, but just guessing :)

jknphy:
Could you please create MR in to setup this test suite?
https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/blob/master/JobGroups/test_suites.yaml
https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/blob/master/JobGroups/SLE_15/x86_64.yaml

tests/installation/registration/rmt_registration_trust_import.pm
```diff
@@ -0,0 +1,20 @@
...
+
+# Summary: Register the system via local RMT in the installer with server url
+# and enabling update repositories.
+#
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
```
jknphy, line 14: you don't need to import the lib when using `$testapi`.

tests/console/validate_repos.pm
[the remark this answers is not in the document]
BillAnastasiadis: Tried escaping first, but I'll retry every possible solution (since it's Friday evening and the rest is done) and re-confirm.

tests/installation/registration/rmt_registration_trust_import.pm
BillAnastasiadis: Is that true? I copied the imports based on already existing `register_via_scc.pm` - should I create a second commit and remove it from there as well?

schedule/yast/rmt_server_install/rmt_server_install.yaml
BillAnastasiadis: Done!

tests/installation/registration/register_via_rmt.pm
```diff
@@ -0,0 +1,23 @@
...
+    $testapi::distri->get_registration()->trust_and_import();
```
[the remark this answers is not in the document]
BillAnastasiadis: Done!

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - console/system_prepare
+  - console/consoletest_setup
+  - console/validate_repos
+test_data:
+  repos:
+    - uri: '%RMT_SERVER%/repo/SUSE/Products/SLE-Module-Basesystem/%VERSION%/%ARCH%/product/?credentials=Basesystem_Module_%VERSION%'
+    - uri: '%RMT_SERVER%/repo/SUSE/Updates/SLE-Module-Basesystem/%VERSION%/%ARCH%/update/?credentials=Basesystem_Module_%VERSION%'
+    - uri: '%RMT_SERVER%/repo/SUSE/Products/SLE-Module-Containers/%VERSION%/%ARCH%/product/?credentials=Containers_Module_%VERSION%'
+    - uri: '%RMT_SERVER%/repo/SUSE/Products/SLE-Module-Desktop-Applications/%VERSION%/%ARCH%/product/?credentials=Desktop_Applications_Module_%VERSION%'
+    - uri: '%RMT_SERVER%/repo/SUSE/Products/SLE-Module-Server-Applications/%VERSION%/%ARCH%//product/?credentials=Server_Applications_Module_%VERSION%'
```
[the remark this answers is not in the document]
BillAnastasiadis: True, fixed anyway!

schedule/yast/rmt_server_install/rmt_server_install.yaml
BillAnastasiadis: Done!

jknphy:
could you please reference to the MR in your first comment to have all together there?

tests/console/validate_repos.pm
```diff
@@ -13,19 +13,22 @@ use base "consoletest";
...
 use repo_tools 'validate_repo_properties';
 use scheduler 'get_test_suite_data';
+use utils;
 
 sub run {
     my $test_data = get_test_suite_data();
 
     select_console 'root-console';
 
+    zypper_call('lr --uri');
```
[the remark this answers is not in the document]
BillAnastasiadis: Of course. Done - now running a VR with `script_output`, in a few mins when it ends I will update the description

schedule/yast/rmt_server_install/rmt_server_install.yaml
```diff
@@ -0,0 +1,47 @@
+---
+name:           rmt_server_install
+description:    >
+  Perform an installation using rmt server for registration,
+  handling the subsequent trust&import popup and validating
+  repo urls after installation. his test case would not work
```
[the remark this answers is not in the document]
BillAnastasiadis: Done!

BillAnastasiadis:
> could you please reference to the MR in your first comment to have all together there?

Done!

tests/console/validate_repos.pm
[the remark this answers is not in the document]
BillAnastasiadis: I am confused as well. Feel free to test/try stuff as well, I was hoping someone more experienced with perl might have a better idea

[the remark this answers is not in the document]
jknphy: we need to debug, it is not obvious there several tools involved, feel free to create a ticket to research that.

jknphy (approved):

PR #15082, opened 2022-06-15
Use new Azure storage account

grisu48:
In runs that use the new Azure credentials also use the new storage
account name.

- Related ticket: https://progress.opensuse.org/issues/112481
- Verification run: https://duck-norris.qam.suse.de/tests/8936#

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -17,7 +17,8 @@ use utils qw(script_output_retry);
 use publiccloud::azure_client;
 
 has resource_group => 'openqa-upload';
-has storage_account => 'openqa';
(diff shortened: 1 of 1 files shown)


pdostal (approved):

jlausuch (approved):
We do have a variable for Google `PUBLIC_CLOUD_GOOGLE_STORAGE`. Not sure why we don't do the same for the others, this way it would be easier to fix in variables instead of hardcode the value.

grisu48:
> We do have a variable for Google `PUBLIC_CLOUD_GOOGLE_STORAGE`. Not sure why we don't do the same for the others, this way it would be easier to fix in variables instead of hardcode the value.

True, I would leave this however outside of the scope of this quickfix.

PR #15089, opened 2022-06-15
Introduce PUBLIC_CLOUD_STORAGE_ACCOUNT variable

grisu48:
Introduce the `PUBLIC_CLOUD_STORAGE_ACCOUNT` variable to allow
customization of the CSP storage account being used.

- Related ticket: https://progress.opensuse.org/issues/112481
- Verification run: [SLE Micro](https://duck-norris.qam.suse.de/tests/8947) | [Old credential handling](https://duck-norris.qam.suse.de/tests/8963#dependencies) | [New credentials](https://duck-norris.qam.suse.de/tests/8961#dependencies)

### DIFF
--- a/data/publiccloud/terraform/azure.tf
+++ b/data/publiccloud/terraform/azure.tf
@@ -56,6 +56,10 @@ variable "create-extra-disk" {
 }
 
 variable "storage-account" {
+    # Note: Don't delete the default value!!!
+    # Not all of our `terraform destroy` calls pass this variable and neither is it necessary.
+    # However removing the default value might cause `terraform destroy` to fail in corner cases,
+    # resulting effectively in leaking resources due to failed cleanups.
     default="openqa"
 }
 
@@ -152,7 +156,7 @@ resource "azurerm_image" "image" {
     os_disk {
         os_type = "Linux"
         os_state = "Generalized"
-        blob_uri = "https://openqa.blob.core.windows.net/sle-images/${var.image_id}"
+        blob_uri = "https://${var.storage-account}.blob.core.windows.net/sle-images/${var.image_id}"
         size_gb = 30
     }
 }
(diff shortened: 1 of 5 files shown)


grisu48:
Before merging I need to update some job settings in https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/843

lib/publiccloud/provider.pm
```diff
@@ -387,6 +387,12 @@ sub terraform_apply {
             $cmd .= sprintf(q(-var '%s=%s' ), $key, escape_single_quote($value));
         }
         $cmd .= "-var 'image_id=" . $image . "' " if ($image);
+        if (is_azure) {
```
jlausuch, line 390: Why is this condition needed?  In `terraform_destroy` you are only doing:
`$cmd .= " -var 'storage-account=$storage_account'" if ($storage_account);` without checking if it's azure or not.

grisu48: Because passing the `storage_account` variable is only valid for the Azure Terraform profile because only there the variable is present.

grisu48: But it is there already, see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/e8e4e17ad27f7ee4482c71d326fec96923ddd544/lib/publiccloud/provider.pm#L498 or what am I missing?

jlausuch: sorry, I mean `if (is_azure) {` but it is there, I needed to expand the code to see it (facepalm)

jlausuch (approved):

data/publiccloud/terraform/azure.tf
```diff
@@ -56,6 +56,10 @@ variable "create-extra-disk" {
 }
 
 variable "storage-account" {
+    # Note: Don't delete the default value!!!
+    # Not all of our `terraform destroy` calls pass this variable and neither is it necessary.
+    # However removing the default value might cause `terraform destroy` to fail in corner cases,
+    # resulting effectively in leaking resources due to failed cleanups.
     default="openqa"
```
asmorodskyi, line 63: and why do we keep `openqa` as default keeping in mind that we already switched majority of our tests to run in new account ?

jlausuch: incidents still run in old ones.

PR #15091, opened 2022-06-15
Attempt to fix typing issues in xterm

mloviska:
As it is described in [[leap][qe-core] test fails in xterm - mistyping test string](https://progress.opensuse.org/issues/111752)
there are several typing issues observed in `xterm` usage. Either the
test string is not typed correctly or the focus of the `xterm` window is
lost.


okurz:
An issue that would *not* be fixed by the change in `enter_test_text` is https://openqa.opensuse.org/tests/2417087#step/updates_packagekit_gpk/9 which also happens in xterm but in the test module `updates_packagekit_gpk`. Tricky but I don't think it's beneficial to type everything slower. How about `xset r off` as a generic setup command as soon as logging into the X session to disable key repeat? This prevents key repeat issues not missing keys, depending on what issues in particular you found and want to fix

Vogtinator:
IIRC there's an older bug report for this and it's a GNOME specific window manager bug about focus.

IMO this should be treated as a DE-specific softfail at least.

mloviska:
> IIRC there's an older bug report for this and it's a GNOME specific window manager bug about focus.
> 
> IMO this should be treated as a DE-specific softfail at least.

Yes, I think the focus issue happens in GNOME only. Would you mind to share the bug report, please?

lib/susedistribution.pm
```diff
@@ -293,6 +293,8 @@ sub x11_start_program {
     for (1 .. 3) {
         push @target, check_var('DESKTOP', 'kde') ? 'desktop-runner-plasma-suggestions' : 'desktop-runner-border';
         assert_screen([@target], $args{match_timeout}, no_wait => $args{match_no_wait});
+        # ensure that app is focused
+        click_lastmatch(mousehide => 1);
```
okurz, line 297: based on the comment by @Vogtinator I suggest

```suggestion
        click_lastmatch(mousehide => 1) if check_var('DESKTOP', 'gnome');
```

okurz:
@mloviska Your PR brought me to an idea: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15095 . Might be helpful.

mloviska:
> @mloviska Your PR brought me to an idea: #15095 . Might be helpful.

Alright, let's try it out

okurz (changes requested):
As explained in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15091#issuecomment-1156543572 I do *not* think typing the editor test slow in general is a good idea. It would only fix some problems but not all. The next logical consequence would be to type *everything* slower which is certainly not what we want to do. You could just keep the "focus"-problem separately and we can merge that.

mloviska:
Still [a problem](https://openqa.suse.de/tests/9064865#step/xterm/6) in spite of merging  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15095

okurz:
> Still [a problem](https://openqa.suse.de/tests/9064865#step/xterm/6) in spite of merging #15095

That was expected. https://openqa.suse.de/tests/9064865#step/xterm/6 shows missing characters, not repeated characters so https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15095 could not have fixed it. I would be happy if you continue your work here but not simply typing slower in all cases.

okurz:
> It might occur in any other usage of `xterm` before this module, as `setup` precedes `xterm` module.

If you happen to find cases of xterm being called before "setup" please include this test module there as well

> Alright, so what do you suggest except changing the pace of typing?

Investigate if the system is responsive enough to actually accept commands, make sure there is enough free memory, only apply slow typing workarounds on machines , backends, architectures that need it, try the reproduce the problem on another OS, e.g. TinyCoreOS from os-autoinst, debug if the keys get lost on the level of the driver, VM, qemu, Linux Kernel, GUI Frameworks, etc.

okurz (approved):

okurz:
> VRs related to focus:
> 
> [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467085) [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467086) [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467087) …

I suggest you use https://progress.opensuse.org/projects/openqatests/wiki/Wiki#Statistical-investigation for nicer test overviews.

mloviska:
> > VRs related to focus:
> > [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467085) [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467086) [opensuse-15.4-DVD-Backports-Incidents-x86_64-Build20220712-5-gnome@uefi](https://openqa.opensuse.org/t2467087) …
> 
> I suggest you use https://progress.opensuse.org/projects/openqatests/wiki/Wiki#Statistical-investigation for nicer test overviews.

Yeah, thanks. I forgot to add different `build` value in `openqa-clone-custom-git-refspec` calls, it kinda does not matter as of now.

DimStar77:
This change is really bad, as it adds 'random clicks' after every single program start

https://openqa.opensuse.org/tests/2468012#step/user_gui_login/3 - we do a program start for 'session quit' (which is, arguably not a program), check for the logout dialog to be present and 'next step would be click logout'; but with this change, the click happens implicit now

https://openqa.opensuse.org/tests/2468012#step/gnome_control_center/8 - here we just happen to click in the top menu as the needles were never prepared to have click areas respected.

PR #15075, opened 2022-06-14
Check if org.opensuse.reference label contains CONTAINER_IMAGE_BUILD

pdostal:
- Related ticket: [poo#100916](https://progress.opensuse.org/issues/100916)
- Verification run: [SLE15-SP3-podman](http://pdostal-server.suse.cz/tests/87), [Ubuntu-22.04-docker](http://pdostal-server.suse.cz/tests/92), [RES8.5-docker](http://pdostal-server.suse.cz/tests/90) (unrelated error)

PR #15093, opened 2022-06-15
Register phub addon because of jq on SLE12-SP5

pdostal:
* Follows: #15075 
- Related ticket: [poo#112541](https://progress.opensuse.org/issues/112541)
- Verification run: http://pdostal-server.suse.cz/tests/102

### DIFF
--- a/tests/containers/bci_prepare.pm
+++ b/tests/containers/bci_prepare.pm
@@ -41,6 +41,8 @@ sub packages_to_install {
         my $version = "$version.$sp";
         push @packages, 'python3-devel';
         if ($version eq "12.5") {
+            # PackageHub is needed for jq
+            script_retry("SUSEConnect -p PackageHub/12.5/$arch", delay => 60, retry => 3);
             push @packages, 'python36-pip';
         } elsif ($version eq '15.0') {
             # On SLES15 go needs to be installed from packagehub. On later SLES it comes from the SDK module


tests/containers/bci_prepare.pm
```diff
@@ -41,6 +41,8 @@ sub packages_to_install {
         my $version = "$version.$sp";
         push @packages, 'python3-devel';
         if ($version eq "12.5") {
+            # PackageHub is needed for jq
+            script_retry("SUSEConnect -p PackageHub/12.5/$arch", delay => 60, retry => 3);
```
grisu48, line 45: Why not use the provided standard subroutines?

```suggestion
            add_suseconnect_product(get_addon_fullname('phub'));
```

tests/containers/bci_prepare.pm
```diff
@@ -41,6 +41,8 @@ sub packages_to_install {
         my $version = "$version.$sp";
         push @packages, 'python3-devel';
         if ($version eq "12.5") {
```
grisu48, line 43: What about 12-SP4? Pretty sure if `jq` is missing on 12-SP5 then 12-SP4 (and 12-SP3) will also complain.

jlausuch: yeah, I had to check myself till which version we are using in the tests :)

tests/containers/bci_prepare.pm
pdostal: Open the VR and see the `BUILD` and `VERSION`.

tests/containers/bci_prepare.pm
pdostal: We don't run this on 12-SP4. Or do we?

tests/containers/bci_prepare.pm
grisu48: I see. Don't mind me resolving the issue here as well.

grisu48 (approved):

PR #15098, opened 2022-06-15
ALP: create new product "alp" with its main.pm

jlausuch:
This ALP PoC in openQA can have its own product directory,
similar to microos, leap-micro or sle-micro.
For these jobs, we will be using DISTRI=alp, which means
new way of scheduling tests.
I am proposing to create this main.pm which will be the main
entry point for scheduling tests for ALP. Most of them are
the same as for MicroOS/Leap-Micro, and I might come up with
a common schedule file file for ALL these transactional-server
based products, but for now, I would prefer to keep it with
single main schedule.

Related ticket: https://progress.opensuse.org/issues/112409

### DIFF
--- a/products/alp/main.pm
+++ b/products/alp/main.pm
@@ -1,0 +1,72 @@
+use strict;
+use warnings;
+use needle;
+use File::Basename;
+use scheduler 'load_yaml_schedule';
+BEGIN {
+    unshift @INC, dirname(__FILE__) . '/../../lib';
+}
+use utils;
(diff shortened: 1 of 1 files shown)


okurz (approved):

mloviska (approved):

foursixnine (approved):
I would shy away from using the old style of scheduling tests, but if you're already having the thought on your mind, then I have no objections :)

PR #15103, opened 2022-06-16
[WIP]Set flag to run install_service

hjluo:
For the new regression tests, we need to set INSTALL_ONLY=1 to
generate the qcow file and then run the regression tests. and
We want to load install_service in the installation part.

- Related ticket: https://progress.opensuse.org/issues/110875
- Needles: N/A
- Verification run:  
https://openqa.nue.suse.com/tests/8965952 with FORCE_INSTALL_SERVICE=1
https://openqa.nue.suse.com/t8965961 w/o FORCE_INSTALL_SERVICE

### DIFF
--- a/products/sle/main.pm
+++ b/products/sle/main.pm
@@ -477,7 +477,7 @@ sub load_online_migration_tests {
     if (is_sle && (get_var('FLAVOR') =~ /Migration/) && (get_var('SCC_ADDONS') !~ /ha/) && !is_sles4sap && (is_upgrade || get_var('MEDIA_UPGRADE'))) {
         loadtest "console/check_system_info";
     }
-    loadtest 'installation/install_service' if (is_sle && !is_desktop && !get_var('INSTALLONLY'));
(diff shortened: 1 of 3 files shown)


tinawang123:
I think install_service only configured at yaml file. Do we still need add this at here?

hjluo:
> I think install_service only configured at yaml file. Do we still need add this at here?

Now this PR was hold to WIP, cause We are now use yaml to avoid set INSTALL_ONLY=1. thanks

products/sle/main.pm
```diff
@@ -477,7 +477,7 @@ sub load_online_migration_tests {
     if (is_sle && (get_var('FLAVOR') =~ /Migration/) && (get_var('SCC_ADDONS') !~ /ha/) && !is_sles4sap && (is_upgrade || get_var('MEDIA_UPGRADE'))) {
         loadtest "console/check_system_info";
     }
-    loadtest 'installation/install_service' if (is_sle && !is_desktop && !get_var('INSTALLONLY'));
+    loadtest 'installation/install_service' if (get_var('FORCE_INSTALL_SERVICE') || (is_sle && !is_desktop && !get_var('INSTALLONLY')));
```
lemon-suse, line 480: It is better to use yaml to do such specific progress.

PR #15104, opened 2022-06-17
xfstests: modify the upload log wait loop

SeroSun:
Using background_script_run to replace previous loop sleep way to wait the file complete is more safety.

- Verification run: 
https://openqa.opensuse.org/tests/2419805
https://openqa.suse.de/tests/8972022


lansuse:
LGTM

PR #15106, opened 2022-06-17
console/ping: Add more logging

pevik:
systemd-sysctl.service does some problems to reading sysctl configuration from aaa_base, thus log systemd version.

And obviously log also ping and iputils.

- Verification run: http://quasar.suse.cz/tests/126#step/ping/27


tests/console/ping.pm
```diff
@@ -18,14 +18,14 @@ sub run {
...
-    record_info('aaa_base', script_output('rpm -qi aaa_base'));
     record_info('KERNEL VERSION', script_output('uname -a'));
-    record_info('sysctl', script_output('sysctl -V'));
-    record_info('procps', script_output('rpm -qi procps'));
+    record_info('net.ipv4.ping_group_range', script_output('sysctl net.ipv4.ping_group_range'));
+    record_info('ping', script_output('ping -V'));
 
     my $kernel_pkg = is_jeos ? 'kernel-default-base' : 'kernel-default';
-    record_info('KERNEL DEFAULT PKG', script_output("rpm -qi $kernel_pkg", proceed_on_failure => 1));
+    foreach my $pkg ($kernel_pkg, 'aaa_base', 'iputils', 'procps', 'sysctl', 'systemd') {
```
dzedro, line 26: sysctl is part of procps, threre is no sysctl package

pevik: Thx! Originally it was `sysctl -V`, but that's redundant, let's remove it.

PR #15109, opened 2022-06-17
Azure CLI: make bsc#1192671 soft

jlausuch:
We are not running this test at all, and the only issue that the CLI has is this bug. The rest could be run.
Also, by enabling the test again, we will be able to verify a possible fix for it.

VR: https://openqa.suse.de/tests/8972597


pdostal (approved):
🚀

PR #15110, opened 2022-06-17
Optimize installation/add_update_test_repo

mdoucha:
Handle optional screens and pop-ups while adding custom repository URLs based on needle matches instead of needlessly waiting for optional screens which will not show up.

- Related ticket: https://progress.opensuse.org/issues/112343
- Needles: N/A
- Verification runs:
  - SLE-12SP5: https://openqa.suse.de/tests/10323009
  - SLE-15SP4: https://openqa.suse.de/tests/10323008

### DIFF
--- a/lib/qam.pm
+++ b/lib/qam.pm
@@ -12,7 +12,7 @@ use base "Exporter";
 use Exporter;
 
 use testapi;
-use utils qw(zypper_call);
+use utils qw(zypper_call handle_screen);
 use JSON;
 use List::Util qw(max);
 use version_utils 'is_sle';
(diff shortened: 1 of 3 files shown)


dzedro:
Looks pretty complicated, isn't `check_screen` with `match_has_tag` doing the same ? e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/installation/await_install.pm#L115

mdoucha:
`handle_screen` has a few benefits over using `match_has_tag()` directly:
- You don't have to make a long if/else code noodle to pick the right handler. If you have predefine handler functions elsewhere, `handle_screen` needs just one line per handler.
- You don't have to wrap the whole thing in a loop if you need to handle multiple optional screens.
- You can pass a partially constructed handler hash another function that'll add more handlers and finish the task. For example this'll be useful for handling GPG key popups which need to be handled in multiple places during installation with different follow-up screens.

dzedro:
Please also document the return / exit behavior I think it's important part of the handler_map.

lib/qam.pm
```diff
@@ -140,15 +140,26 @@ sub advance_installer_window {
...
-        }
+            return 0;
+        };
     }
-    unless (check_screen "$screenName", 60) {
-        my $key = check_screen('cannot-access-installation-media') ? "alt-y" : "$cmd{next}";
-        send_key_until_needlematch $screenName, $key, 5, 60;
+
+    unless (handle_screen([keys %handlers], \%handlers, assert => 0, timeout => 60)) {
+        send_key_until_needlematch $screenName, $cmd{next}, 5, 60;
```
dzedro, line 162: I think this will be never executed as handler will exit only when $exit is true.

mdoucha: `handle_screen()` exits when:
- handler function returns true (line 162 will be skipped)
- the internal loop runs out of retries, in this case after 3 or 4 iterations, depending on `$build` (line 162 will be executed)
- none of the needles match, since `assert` is 0 (line 162 will be executed)

dzedro (approved):

mdoucha:
Rebased on current master and updated verification run. This PR is ready for merge.

mdoucha:
Rebased again. Verification runs updated. This PR has been ready for merge for 6 months already.

kalikiana (approved):
Looks sensible to me, and it's well-documented.

I was unsure about the name on first glance, but it makes sense.

PR #15115, opened 2022-06-19
Test adaptations for ALP

jlausuch:
This is combination of several commits:
- Create new util function is_alp
- Execute filesystem_ro test only for transactional ALP
- Execute enable_selinux on ALP always
- Execute enable_selinux on ALP always
- Add some bugs in ALP to journal whitelist

Related ticket: https://progress.opensuse.org/issues/112427

VR:
- https://openqa.opensuse.org/tests/2425554
- https://openqa.opensuse.org/tests/2425553

### DIFF
--- a/lib/version_utils.pm
+++ b/lib/version_utils.pm
@@ -21,6 +21,7 @@ use constant {
           is_microos
           is_leap_micro
           is_sle_micro
+          is_alp
           is_selfinstall
           is_gnome_next
           is_jeos
@@ -257,6 +258,21 @@ sub is_sle_micro {
     return check_version($query, $version, qr/\d{1,}\.\d/);
 }
 
+=head2 is_alp
+
+Check if distribution is ALP
+=cut
+sub is_alp {
+    my $query = shift;
+    my $version = shift // get_var('VERSION');
+
+    return 0 unless check_var('DISTRI', 'alp');
+    return 1 unless $query;
+
+    # Version check
+    return check_version($query, $version, qr/\d{1,}\.\d/);
+}
+
 =head2 is_selfinstall
 
 Check if SLEM is in flavor of self installable iso
@@ -311,6 +327,7 @@ sub is_opensuse {
     return 1 if check_var('DISTRI', 'opensuse');
     return 1 if check_var('DISTRI', 'microos');
     return 1 if check_var('DISTRI', 'leap-micro');
+    return 1 if check_var('DISTRI', 'alp');
     return 0;
 }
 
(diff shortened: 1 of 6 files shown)


data/microos/utt-alp.repo
```diff
@@ -0,0 +1,7 @@
+[openSUSE_Factory_utt]
+name=openSUSE:Factory:utt
+type=rpm-md
+baseurl=https://download.opensuse.org/repositories/openSUSE:/Leap:/Micro:/5.1:/Update/standard/
```
mloviska, line 4: Not this one `https://download.opensuse.org/repositories/devel:/LEO/standard/` ?

jlausuch: No, because 2 things:
- This repo doesn't include UTT packages
- This repo addition is not for all the updates, it's meant ony for UTT packages. If I add the repo you suggest, it will "override" the packages that we have synced from local in openQA.

Maybe we can ask someone to make some opensuse generic repo in OBS for UTT only... or rename this Leap Micro 5.1 to something more generic we can use in ALP and Leap Micro... I only found this one which works...  
Tumbleweed already contains UTT packages in the Factory repo...

jlausuch: Already asked, we might be using the same we use for Leap Micro, waiting now.

jlausuch: Already asked, we might be using the same we use for Leap Micro, waiting now.

tests/transactional/host_config.pm
```diff
@@ -30,8 +32,15 @@ sub run {
         assert_script_run('transactional-update grub.cfg');
         process_reboot(trigger => 1);
     }
-
-    # Placeholder for other configurations we might need (not related to grub)
+    if (is_alp) {
+        record_info('Packages', 'Install needed packages to run the tests');
+        if (is_transactional) {
+            trup_call 'pkg install tar', timeout => 300;
```
kalikiana, line 38: Minor point, but I'd suggest to use braces consistently
```suggestion
            trup_call('pkg install tar', timeout => 300);
```

lib/version_utils.pm
```diff
@@ -257,6 +258,21 @@ sub is_sle_micro {
     return check_version($query, $version, qr/\d{1,}\.\d/);
 }
 
+=head2 is_alp
+
+Check if distribution is ALP
+=cut
+sub is_alp {
+    my $query = shift;
+    my $version = shift // get_var('VERSION');
```
foursixnine, line 267:
```suggestion
    my ($query, $version) = @_;
    $version //= get_var('VERSION');
```

foursixnine (commented):
🥳

lib/version_utils.pm
jlausuch: Yes, that's probably how I would do it, but I wanted to keep the same style of the other similar methods..

PR #15123, opened 2022-06-21
Restart cups service after configuration changes

mloviska:
As long as cups daemon is running prior to this test case, starting the
service becomes a noop therefore configuration changes are not applied.

Warning message: `lpadmin: Raw queues are deprecated and will stop
working in a future version of CUPS.` is not handled in this PR as it is
harmless in the current version of `cups`.

- Verification runs: 
  - [opensuse-15.4-DVD-x86_64-Build243.2-extra_tests_textmode@64bit](http://kepler.suse.cz/tests/17513#step/cups/64)
  - [Build20220619 of opensuse-Tumbleweed-DVD.x86_64](http://kepler.suse.cz/tests/17514#step/cups/64)

### DIFF
--- a/lib/services/cups.pm
+++ b/lib/services/cups.pm
@@ -21,22 +21,26 @@ sub install_service {
 }
 
 sub config_service {
+    my $conf = '/etc/cups/cups-files.conf';
     if ($service_type eq 'Systemd') {
-        script_run 'echo FileDevice Yes >> /etc/cups/cups-files.conf';
+        script_run "echo FileDevice Yes >> $conf";
         validate_script_output 'cupsd -t', sub { m/is OK/ };
     } else {
+        $conf = '/etc/cups/cupsd.conf';
         # Allow new printers to be added using device URIs "file:/filename"
-        script_run 'echo FileDevice Yes >> /etc/cups/cupsd.conf';
+        script_run "echo FileDevice Yes >> $conf";
         assert_script_run 'cupsd';
     }
+
+    record_info('cups-files', script_output("cat $conf"));
 }
 
 sub enable_service {
     common_service_action 'cups', $service_type, 'enable';
 }
 
-sub start_service {
-    common_service_action 'cups', $service_type, 'start';
+sub restart_service {
+    common_service_action 'cups', $service_type, 'restart';
 }
 
 # check service is running and enabled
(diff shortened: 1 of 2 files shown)


lib/services/cups.pm
```diff
@@ -29,6 +29,8 @@ sub config_service {
         script_run 'echo FileDevice Yes >> /etc/cups/cupsd.conf';
         assert_script_run 'cupsd';
     }
+
+    record_info('cups-files', script_output('cat /etc/cups/cups-files.conf'));
```
Vogtinator, line 33: Does this file matter (or even exist) in the non-systemd case?

tests/console/cups.pm
```diff
@@ -35,6 +35,7 @@ sub run {
     services::cups::config_service();
     services::cups::enable_service();
     services::cups::start_service();
```
Vogtinator, line 37: This means that if cups wasn't running previously, it's started and then immediately restarted. That's a waste of time and space and might actually break it if it gets forcibly terminated during initial startup somehow.

So I suggest to just restart instead of start.

mloviska: Right, that might be a problem. Fixed, thanks!

lib/services/cups.pm
mloviska: Oh, I have assumed they are the same. Fixed, thanks!

Vogtinator:
There's another instance of `start_service` in `full_cups_check`, maybe that should be replaced by `restart_service` too?

mloviska:
> There's another instance of `start_service` in `full_cups_check`, maybe that should be replaced by `restart_service` too?

Sure, why not, done!

punkioudi (approved):
Nice! Could you also add the VRs with the latest changes?

mloviska:
> Nice! Could you also add the VRs with the latest changes?

I have updated links in the description. Adding here full schedule VRs:

* http://kepler.suse.cz/tests/17512#step/cups/52
* http://kepler.suse.cz/tests/17511#step/cups/49

lib/services/cups.pm
```diff
@@ -39,6 +43,10 @@ sub start_service {
     common_service_action 'cups', $service_type, 'start';
```
Vogtinator, line 43: Can be dropped now

mloviska: Done

PR #15127, opened 2022-06-21
publicloud: switch from legacy defaults to new accounts default

asmorodskyi:
VR : http://autobot.qa.suse.de/tests/3225

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -17,8 +17,6 @@ use utils qw(script_output_retry);
 use publiccloud::azure_client;
 
 has resource_group => 'openqa-upload';
-# Note: Remove the default 'openqa' once the deprecated account will be removed.
-has storage_account => get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'openqa');
 has container => 'sle-images';
 has lease_id => undef;
 has vault => undef;
@@ -77,9 +75,9 @@ sub find_img {
 }
 
 sub get_storage_account_keys {
-    my ($self, %args) = @_;
+    my ($self, $storage_account) = @_;
     my $output = script_output("az storage account keys list --resource-group "
-          . $self->resource_group . " --account-name " . $self->storage_account);
+          . $self->resource_group . " --account-name " . $storage_account);
     my $json = decode_azure_json($output);
     my $key = undef;
     if (@{$json} > 0) {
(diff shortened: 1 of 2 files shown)


lib/publiccloud/azure.pm
```diff
@@ -17,8 +17,6 @@ use utils qw(script_output_retry);
 use publiccloud::azure_client;
 
 has resource_group => 'openqa-upload';
-# Note: Remove the default 'openqa' once the deprecated account will be removed.
-has storage_account => get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'openqa');
```
grisu48, line 21: Why do we remove this setting here now and not just the default value? This was present before and I think many of the other code changes are thereby unnecessary.

lib/publiccloud/azure.pm
```diff
@@ -114,18 +111,19 @@ sub upload_img {
     my ($img_name) = $file =~ /([^\/]+)$/;
     $img_name =~ s/\.vhdfixed/.vhd/;
     my $disk_name = $img_name;
+    my $storage_account = get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'eisleqaopenqa');
```
grisu48, line 114: Weren't you the one that was against using a default value for `PUBLIC_CLOUD_STORAGE_ACCOUNT` here, if it is already in the terraform profile? :slightly_smiling_face:

grisu48 (commented):
I believe is we keep the new default value and `has storage_account` variable here, then many of the subsequent changes are not needed?

lib/publiccloud/azure.pm
cfconrad: I would also need more info for this change. Maybe the `$storage_account` should be a optional parameter?

asmorodskyi:
> I believe is we keep the new default value and `has storage_account` variable here, then many of the subsequent changes are not needed?

generally yes

lib/publiccloud/azure.pm
cfconrad: but then, it would not make to have the same values in tf and here or?

If it is upload specific, I would just add a parameter of `upload_img()` and pass it.

lib/publiccloud/azure.pm
cfconrad: A other thing. Does this actually work and should not be 
```perl
has storage_account => sub { get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'openqa')};
```

lib/publiccloud/azure.pm
cfconrad: And use a variable which is like `PUBLIC_CLOUD_IMG_UPLOAD_AZURE_STORAGE_ACCOUNT`

lib/publiccloud/azure.pm
asmorodskyi: now we migrated to new account and because storage account names suppose to be unique globally in whole Azure we could not create storage account with `openqa` name anymore this is why we picked up new name

lib/publiccloud/azure.pm
asmorodskyi: > but then, it would not make to have the same values in tf and here or?

well yes and no . 

Yes it **would not** make sense to have same variable for this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L392 and this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/azure.pm#L21 because from test code perspective this 2 have no relations to each other 

No it **would** make sense to have same variable because from Azure perspective we speaking about exactly same thing ( https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview )

ilausuch (approved):

PR #15128, opened 2022-06-21
helm GKE added and gcloud_install() defined

m-dati:
GKE publiccloud added in containers/helm.pm charts and _gcloud_install()_ defined in lib/utils.pm

- Related ticket: _[Test Helm command with a GKE kubernetes environment](https://progress.opensuse.org/issues/112247)_
- Needles: none
- Verification run: http://marvin5.arch.suse.cz/tests/440

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -94,7 +94,6 @@ our @EXPORT = qw(
   permit_root_ssh_in_sol
   cleanup_disk_space
   package_upgrade_check
-  gcloud_install
 );
 
 =head1 SYNOPSIS
@@ -2232,33 +2231,5 @@ sub package_upgrade_check {
         }
     }
 }
-
-=head2 gcloud_install
-    gcloud_install();
-
-This function is used to install the gcloud CLI 
-for the GKE Google Cloud.
-
-From $url we get the full package and install in $dir local folder,
-that is supposed to be a subdir of the root user home /root.
-The Defaults are availeble for a simple call without parameters.
-
-=cut
-
-sub gcloud_install {
-    my %args = @_;
-    my $url = $args{url} || 'sdk.cloud.google.com';
-    my $dir = $args{dir} || 'google-cloud-sdk';
-    my $timeout = $args{timeout} || 700;
-
-    zypper_call("in curl tar gzip", $timeout);
-    assert_script_run("export CLOUDSDK_CORE_DISABLE_PROMPTS=1");
-    assert_script_run("curl $url | bash", $timeout);
-
-    assert_script_run("echo . /root/$dir/completion.bash.inc >> ~/.bashrc");
-    assert_script_run("echo . /root/$dir/path.bash.inc >> ~/.bashrc");
-    record_info('GCE', script_output('source ~/.bashrc && gcloud version'));
-}
-
 
 1;
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -2251,7 +2251,7 @@ sub gcloud_install {
     my $dir = $args{dir} || 'google-cloud-sdk';
     my $timeout = $args{timeout} || 700;
 
-    assert_script_run("zypper -n in curl tar gzip", $timeout);
+    zypper_call("in curl tar gzip", $timeout);
     assert_script_run("export CLOUDSDK_CORE_DISABLE_PROMPTS=1");
     assert_script_run("curl $url | bash", $timeout);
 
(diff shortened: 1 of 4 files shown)


lib/utils.pm
```diff
@@ -2232,4 +2233,32 @@ sub package_upgrade_check {
...
+    my $timeout = $args{timeout} || 700;
+
+    assert_script_run("zypper -n in curl tar gzip", $timeout);
+    assert_script_run("export CLOUDSDK_CORE_DISABLE_PROMPTS=1");
+    assert_script_run("curl $url | bash", $timeout);
+
+    assert_script_run("echo . /root/$dir/completion.bash.inc >> ~/.bashrc");
+    assert_script_run("echo . /root/$dir/path.bash.inc >> ~/.bashrc");
+    record_info('GCE', script_output('source ~/.bashrc && gcloud version'));
+}
```
grisu48, line 2261: We also need to reload the `.bashrc` otherwise the newly added include files are not loaded

```suggestion
    assert_script_run('. ~/.bash_profile');
}
```

grisu48 (commented):
One addition is needed, otherwise LGTM

tests/containers/helm.pm
```diff
@@ -63,7 +64,15 @@ sub run {
             $provider = publiccloud::aks->new();
         }
         elsif ($k8s_backend eq 'GKE') {
-            die('GKE is not implemented yet');
+            add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
+            gcloud_install();
+
+            # package needed by init():
+            (is_sle('=15-SP4')) ? 
+               assert_script_run('zypper in -y chrony') :
```
jlausuch, line 72: aren't these packages installed by default?

lib/utils.pm
m-dati: > We also need to reload the `.bashrc` otherwise the newly added include files are not loaded

`.bashrc` is already loaded in line 2260. Consider that a `.bash_profile` is executed only at _login_ of a shell

tests/containers/helm.pm
m-dati: No, not by default.
In the test 440 [log](http://marvin5.arch.suse.cz/tests/440/logfile?filename=serial_terminal.txt) if you search  the string `-y ntp` you will see then **ntp** (needed for ntpdate check) not present yet:
```
The following 2 NEW packages are going to be installed:
  libdns_sd ntp
```

jlausuch: and why do we need ntp and chrony for google?

lib/utils.pm
grisu48: IMHO this will not work - Look at http://marvin5.arch.suse.cz/tests/440#step/helm_GKE/46 - if you run `script_output` or `script_run` it is wrapped in some kind of helper script that is executed. This means, that the `source ~./bashrc && gcloud version` will be executed in a new bash, so it has no effect on the current bash.

Only `assert_script_run` runs the commands on the bare bash, `script_output` does some internal magic so you will not have the intended effect.

***

This is very cryptic to understand, but look at http://marvin5.arch.suse.cz/tests/440#step/helm_GKE/46:

    "echo GR72W; bash -oe pipefail /tmp/scriptGR72W.sh ; echo SCRIPT_FINISHEDGR72W-\$?-"

Your command `source ~/.bashrc && gcloud version` is wrapped into `/tmp/scriptGR72W.sh`, so it will be only present in that script, not in the actual bash running here.

m-dati: done

lib/utils.pm
```diff
@@ -2232,4 +2233,32 @@ sub package_upgrade_check {
     }
 }
 
+=head2 gcloud_install
```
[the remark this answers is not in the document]
m-dati: done

m-dati:
here is the **successful** VR of last branch status, with all Helm GKE, EKS, ASK, K3S: http://marvin5.arch.suse.cz/tests/482

tests/containers/helm.pm
```diff
@@ -63,7 +65,15 @@ sub run {
             $provider = publiccloud::aks->new();
         }
         elsif ($k8s_backend eq 'GKE') {
-            die('GKE is not implemented yet');
+            add_suseconnect_product(get_addon_fullname('pcm')) if is_sle;
+            gcloud_install();
+
+            # package needed by init():
+            (is_sle('=15-SP4'))
```
asmorodskyi, line 72:
```
my $pkgs = is_sle('=15-SP4') ? "in chrony" : "in ntp" ;
zypper_call($pkgs, timeout => 300);
```

your version will work ofc but this one looks better

m-dati:
successful VR of the full changes: http://marvin5.arch.suse.cz/tests/567

jlausuch (approved):

PR #15129, opened 2022-06-21
Fix failing download_repos with SKIP_MU=1

grisu48:
When SKIP_MU=1 is set, the `~/repos` directory is not created but
required in other test runs. This commit ensures, this directory is
created regardless if the actual download is skipped or not.

- Verification run: [without SKIP_MU=1](https://duck-norris.qam.suse.de/tests/8974#step/download_repos/95) and [with SKIP_MU=1](https://duck-norris.qam.suse.de/tests/8975#step/download_repos/13)

### DIFF
--- a/tests/publiccloud/download_repos.pm
+++ b/tests/publiccloud/download_repos.pm
@@ -27,12 +27,11 @@ sub run {
     my ($self, $args) = @_;
     select_host_console();    # select console on the host, not the PC instance
 
-    # Skip maintenance updates. This is useful for debug runs
-    my $skip_mu = get_var('PUBLIC_CLOUD_SKIP_MU', 0);
 
-    # Trigger to skip the download to speed up verification runs
-    if ($skip_mu) {
+    if (get_var('PUBLIC_CLOUD_SKIP_MU')) {
+        # Skip maintenance updates. This is useful for debug runs
         record_info('Skip download', 'Skipping maintenance update download (triggered by setting)');
+        return;
     } else {
         # Skip if we already downloaded the repos
         if (get_repo_status() == 1) {


pablo-herranz (approved):

tests/publiccloud/download_repos.pm
```diff
@@ -27,20 +27,19 @@ sub run {
     my ($self, $args) = @_;
     select_host_console();    # select console on the host, not the PC instance
 
-    # Skip maintenance updates. This is useful for debug runs
-    my $skip_mu = get_var('PUBLIC_CLOUD_SKIP_MU', 0);
+    assert_script_run("mkdir ~/repos");
```
asmorodskyi, line 30: why do we need to create folder when we skipping ?

grisu48: Because other tests e.g. `transfer_repos` depend on the existence of this folder.

asmorodskyi: > Because other tests e.g. `transfer_repos` depend on the existence of this folder.

and it is already correctly handled https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/transfer_repos.pm#L28

tests/publiccloud/download_repos.pm
```diff
@@ -27,20 +27,19 @@ sub run {
...
 
-    # Trigger to skip the download to speed up verification runs
-    if ($skip_mu) {
+    if (get_var('PUBLIC_CLOUD_SKIP_MU', 0)) {
```
jlausuch, line 32:
```suggestion
    if (get_var('PUBLIC_CLOUD_SKIP_MU')) {
```

asmorodskyi: well IMHO there is no difference , I mean I don't mind but just out of curiosity what reasoning do you have  ? :)

jlausuch: no diff, but the default value is not needed for the condition.

asmorodskyi (approved):
LGTM

asmorodskyi:
please fix pipeline

pdostal (approved):

pdostal:
🚀

PR #15138, opened 2022-06-22
lib/x11utils: ensure_unlocked_screen: Avoid waiting after screen change

Vogtinator:
Currently it presses Esc to show the lock screen, waits for the lock screen
to appear (through wait_screen_change) and then it does a wait_still_screen
which can take long enough for the lock screen to disappear again.

Avoid the latter by skipping straight to the next match after a screen change
was detected.

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1200615
- Verification run: https://openqa.opensuse.org/tests/2430136

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -178,6 +178,7 @@ sub ensure_unlocked_desktop {
                 send_key 'esc';    # end screenlock
                 diag("Screen lock present");
             };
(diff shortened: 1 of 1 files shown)


ggardet (approved):
LGTM. Thanks!

Vogtinator:
Merged. let's see how well this works in practice.

PR #15139, opened 2022-06-22
Leave an empty line between pod and code to all the files under lib

b10n1k:
Following the recommendation all the perl POD documentation should have a
empty line before the beginning of the code as a good practice. In this PR all
the files which found under lib directory and its subdirectories are adding a
new line from all the cases which matched `=cut\nsub`

- Related ticket: https://progress.opensuse.org/issues/112544

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,10 @@ test-compile: check-links
 test-compile-changed: os-autoinst/
 	export PERL5LIB=${PERL5LIB_} ; for f in `git diff --name-only | grep '.pm'` ; do perl -c $$f 2>&1 | grep -v " OK$$" && exit 2; done ; true
 
+.PHONY: test_pod_whitespace_rule
+test_pod_whitespace_rule:
+	tools/check_pod_whitespace_rule
+
 .PHONY: test-yaml-valid
 test-yaml-valid:
 	$(eval YAMLS=$(shell sh -c "git ls-files schedule/ test_data/ | grep '\\.ya\?ml$$'"))
@@ -107,7 +111,7 @@ test-spec:
 	tools/update_spec --check
 
 .PHONY: test-static
-test-static: tidy-check test-yaml-valid test-modules-in-yaml-schedule test-merge test-dry test-no-wait_idle test-deleted-renamed-referenced-files test-unused-modules-changed test-soft_failure-no-reference test-spec test-invalid-syntax test-code-style test-metadata
+test-static: tidy-check test-yaml-valid test-modules-in-yaml-schedule test-merge test-dry test-no-wait_idle test-deleted-renamed-referenced-files test-unused-modules-changed test-soft_failure-no-reference test-spec test-invalid-syntax test-code-style test-metadata test_pod_whitespace_rule
 
 .PHONY: test
 ifeq ($(TESTS),compile)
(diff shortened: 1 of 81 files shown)


mpagot:
Why not to try to add the script that you used somewhere in the tools folder and call it from the Makefile?
If we agree about this comment style, why not to try to automatically "enforce" it (similar of what we are doing with perl tidy)?

tools/check_pod_whitespace_rule
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -ex
+<< 'heredoc_pod_whitespace_rule'
+As a convention the POD and the code needs to have an empty line in between.
+The script checks all the files rather the commit files in the lib directory.
+heredoc_pod_whitespace_rule
+
+
+grep -Pzo "=cut\nsub" -rn lib || exit 0
```
mpagot, line 8: Improvement idea: looping through all files one by one and test one by one could provide a better error report to end user also indicating which file is wrong.

mpagot: does it make sense to add this test within the existing https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tools/check_code_style script?

b10n1k: > Improvement idea: looping through all files one by one and test one by one could provide a better error report to end user also indicating which file is wrong.

i think you are right about the better reporting however IMO it is a really simple test. you just need to validate that the rule is not violated. the output in the current check will print all the files which do not have the whitespace. So i doubt that the one by one approach will make a big difference. In the other side, it will make the script to run a bit longer.

b10n1k: > does it make sense to add this test within the existing https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tools/check_code_style script?

it does. i will do so if it will be requested. Do you want to move the check there?

Makefile
```diff
@@ -153,6 +153,10 @@ test-soft_failure-no-reference:
 test-invalid-syntax:
 	tools/check_invalid_syntax
 
+.PHONY: test-pod-whitespace_rule
+test-code-style:
```
okurz, line 157: this should be part of the target "test-static"

tools/check_pod_whitespace_rule
okurz: I suggest to actually use `git grep` which can be much faster. This is also what we use in other cases.

okurz (changes requested):

tools/check_pod_whitespace_rule
b10n1k: The thing is that the `git grep` didnt give me the results when i tried. I am not sure why but it doesnt seem to work with those options. Can you check by yourself please? What do i miss?

Makefile
b10n1k: Done *this change is ready for review*

tools/check_pod_whitespace_rule
b10n1k: > > does it make sense to add this test within the existing https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tools/check_code_style script?

As i moved the check in the `test-static` i guess there is no anymore a reason to discuss this.

okurz: I assume your check relies on some special behaviour to grep for the newline. That's fine, just leave it as is

okurz (approved):

mpagot (approved):

PR #15150, opened 2022-06-23
Trento deploy from unreleased version

mpagot:
New variables added to allow the user to specify a registry and a
version for each deployment part: chart, runner and web.

- Verification run: http://1c242.qa.suse.de/tests/504/  http://1c242.qa.suse.de/tests/588

### DIFF
--- a/tests/sles4sap/trento/deploy_trento.pm
+++ b/tests/sles4sap/trento/deploy_trento.pm
@@ -20,13 +20,14 @@ sub run {
     my $machine_name = $self->get_vm_name;
     my $acr_name = $self->get_acr_name;
 
-    enter_cmd "cd /root/test";
+    my $basedir = '/root/test';
+    enter_cmd "cd $basedir";
 
+    my $script_run = ' ./';
     # Run the Trento deployment
     my $vm_image = get_var(TRENTO_VM_IMAGE => 'SUSE:sles-sap-15-sp3-byos:gen2:latest');
     my $deploy_script_log = 'script_00.040.txt';
-    my $deploy_script_run = './';
-    my $cmd_00_040 = 'set -o pipefail ; ' . $deploy_script_run . "00.040-trento_vm_server_deploy_azure.sh " .
+    my $cmd_00_040 = 'set -o pipefail ; ' . $script_run . "00.040-trento_vm_server_deploy_azure.sh " .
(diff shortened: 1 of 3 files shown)


tests/sles4sap/trento/deploy_trento.pm
```diff
@@ -20,13 +20,14 @@ sub run {
     my $machine_name = $self->get_vm_name;
     my $acr_name = $self->get_acr_name;
 
-    enter_cmd "cd /root/test";
+    my $cp = '/root/test';
+    enter_cmd "cd $cp";
 
+    my $sr = ' ./';
```
alvarocarvajald, line 26: Any reason to switch from `$deploy_script_run` to `$sr`? IMO it was clearer before.

tests/sles4sap/trento/deploy_trento.pm
```diff
@@ -37,14 +38,48 @@ sub run {
     upload_logs($deploy_script_log);
 
     my $trento_registry_chart = get_var(TRENTO_REGISTRY_CHART => 'registry.suse.com/trento/trento-server');
+    my $cfg_json = 'config_images_gen.json';
+    my @imgs = qw(WEB RUNNER);
+    if (get_var("TRENTO_REGISTRY_IMAGE_$imgs[0]") || get_var("TRENTO_REGISTRY_IMAGE_$imgs[1]")) {
```
alvarocarvajald, line 43: I don't understand the need for this if conditional here, when you check for those settings again on L51.

mpagot: There are multiple TRENTO_REGISTRY_IMAGE_*** variables, all the code block from L43 is executed if at least one of them exist, al the other tests are dedicated to one particular TRENTO_REGISTRY_IMAGE_*** 

I can do something like
```
my $unreleased_trento = get_var("TRENTO_REGISTRY_IMAGE_$imgs[0]") || get_var("TRENTO_REGISTRY_IMAGE_$imgs[1]");

if ($unreleased_trento) {
```

alvarocarvajald: Disregard my comment. I see now that you have some command executions between L62-L64

alvarocarvajald (approved):
LGTM

PR #15154, opened 2022-06-24
Redesign the migration regression cases

chcao:
We redesign the migration regression cases that spliting a long case into
several cases: after migration, publsh a qcow, and then do regression
tests with the qcow; for the regression tests, we use INCLUDE_MODULES
setting to load the regression test modules.

- Related ticket: https://progress.opensuse.org/issues/112397
- Needles: N/A
- Verification run: 
https://openqa.suse.de/tests/9009427 (migration and publish qcow)
https://openqa.suse.de/tests/9013971 (regression test of check_upgraded_service)
https://openqa.suse.de/tests/9009432 (regression test of ImageMagick)
https://openqa.suse.de/tests/9009435 (regression test of console)
https://openqa.suse.de/tests/9009434 (regression test of gnome)
https://openqa.suse.de/tests/9009428 (regression test of we)
https://openqa.suse.de/tests/9009431 (regression test of wireshark)
https://openqa.suse.de/tests/9009429 (regression test of openldap_to_389ds)
https://openqa.suse.de/tests/9009430 (regression test of migration_features)
https://openqa.suse.de/tests/9009433 (regression test of tomcat)


lib/service_check.pm
```diff
@@ -334,7 +334,7 @@ sub check_services {
         my $srv_proc_name = $service->{$s}->{srv_proc_name};
         my $support_ver = $service->{$s}->{support_ver};
         my $service_type = 'Systemd';
-        next unless (($service->{$s}->{before_migration} eq 'PASS') && _is_applicable($srv_pkg_name));
```
coolgw, line 337: suggest put comments here why why we remove this.

lemon-suse (changes requested):
I think this need more investigation. 
1. Regression module have module dependency, how to solve it? If need more module combination for the migration qcow, need investigate whether the redesign is valuable.
2. Need investigate whether to split to two jobs is valuable to do it.
3. I don't think change the logic of service check is needed. If you want to change it for a specific change please add setting to control.
 From my opinion, we just need split the regression test into more small group is enough, do not need so big change

lib/service_check.pm
chcao: OK, I add the reason to the commit and the description, thanks.

chcao:
> I think this need more investigation.
> 
>     1. Regression module have module dependency, how to solve it? If need more module combination for the migration qcow, need investigate whether the redesign is valuable.
If any modules are needed by regression tests, add the module in the migration case would solve it. BTW, the uploaded qcow is registered, so the environment is the same as before.
> 
>     2. Need investigate whether to split to two jobs is valuable to do it.
This is what I did recently, splitting to several jobs would save the time that running the regression tests one by one. 
> 
>     3. I don't think change the logic of service check is needed. If you want to change it for a specific change please add setting to control.
I didn't change the logic of service check. I just put the checking of the upgraded services into the second case.
>        From my opinion, we just need split the regression test into more small group is enough, do not need so big change

lib/service_check.pm
coolgw: We need add comments into the code, that's can make our life more easier.

lemon-suse:
The PR seems works while base on the solution will case lost test coverage, and it's benefit can be realize with debug method. I will reserve my opinion, PO & SM can decide whether to use this solution.

lemon-suse (approved):

PR #15158, opened 2022-06-24
Add a test suite for ncurses installation in maintenance.

JRivrain:
We would like to have an interactive installation with ncurses in MU
environment.

VR
restapi http://waaa-amazing.suse.cz/tests/17278
needle-based http://waaa-amazing.suse.cz/tests/17199


JRivrain (commented):
I am keeping this as WIP for now. I am not sure about the pertinence of the change: the installer is GM, no matter what we install on the disk. There is the auto-update of the installer itself though, but it does not depend on packages from MU.

PR #15160, opened 2022-06-26
Containers: create new schedule to update a host

jlausuch:
https://progress.opensuse.org/issues/112886
VRs: http://marvin5.arch.suse.cz/group_overview/1

### DIFF
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -155,8 +155,21 @@ sub load_host_tests_helm {
     }
 }
 
+sub update_host {
+    # Method used to update the non-sle hosts, booting
+    # the existing qcow2 and publish a new qcow2
+    loadtest 'boot/boot_to_desktop';
+    loadtest 'containers/update_host';
+    loadtest 'shutdown/shutdown';
+}
+
 sub load_container_tests {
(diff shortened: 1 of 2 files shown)


mloviska:
I think we should do a reboot as new kernel could be installed.

pdostal:
I think that this will not fix the console mismatch on RHEL. You need to reboot.

mloviska:
> > I think we should do a reboot as new kernel could be installed.
> 
> You mean to check it boots ok?

Ah, sorry! I thought that the jobs are just a PoC and the module will be added to the existing test flow. :facepalm:

mloviska (approved):

pdostal (approved):
> > I think that this will not fix the console mismatch on RHEL. You need to reboot.
> 
> Just bare in mind this job is to be executed individually, not before the tests. This is to keep the qcow2's updated. But I think it's a good idea to add a reboot in this module to check that they boot ok after update. will do so.

Everything is OK then 🚀

jlausuch:
I added a reboot to make sure the hosts boot properly.
Green VRs: http://marvin5.arch.suse.cz/group_overview/1

PR #15165, opened 2022-06-29
Force resolution of policycoreutils-python-utils

dzedro:
Software packages depend on each other in various ways. Packages usually require
or recommend other packages, but they can also conflict with them. Packages may
support specific hardware or language settings. Zypper uses a dependency solver
to find out which packages need to be installed to satisfy the user’s request.

In Focus on Update mode the solver focuses on updating the requested package
and all its dependencies as much as possible. Beware, installing a single package
in this mode may easily lead to a mini system update.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1200649
- Verification run: https://openqa.suse.de/tests/9048309


dzedro:
Wow, I forgot this existed. Is it still valid ? :innocent:

PR #15167, opened 2022-06-30
Containers: The k3s version can be controlled by variable

ilausuch:
The k3s version can be controlled by variable

- Related ticket: https://progress.opensuse.org/issues/113114
- Verification run: http://marvin5.arch.suse.cz/tests/680


lib/containers/k8s.pm
```diff
@@ -46,6 +47,7 @@ sub install_kubectl {
 
 sub install_helm {
     zypper_call("in helm");
+    assert_script_run("helm version");
```
jlausuch, line 50:
```suggestion
    record_info('helm', script_output("helm version"));
```

ilausuch: done

jlausuch (approved):
LGTM

PR #15172, opened 2022-07-04
BCI: enable metadata tests

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/112757
- VR: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=jlausuch%2Fos-autoinst-distri-opensuse%23bci_metadata

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -33,13 +33,15 @@ sub parse_logs {
     my $root = $dom->createElement('testsuites');
     $dom->setDocumentElement($root);
 
-    record_info('Files', script_output('ls -lh'));
+    my $result_files = script_output('ls | grep junit');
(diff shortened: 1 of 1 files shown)


tests/containers/bci_test.pm
```diff
@@ -33,13 +33,15 @@ sub parse_logs {
     my $root = $dom->createElement('testsuites');
     $dom->setDocumentElement($root);
 
-    record_info('Files', script_output('ls -lh'));
+    my $result_files = script_output('ls | grep junit');
```
dcermak, line 36: These files will only appear *after* you ran the tests.

jlausuch: Yes, and this method is called at the very end after all the tox commands.

PR #15174, opened 2022-07-05
Add kubectl test run

grisu48:
Add a test run for testing `kubectl` for Tumbleweed. This PR adds the test code. The modifications to the job groups on O3 will be done after merging.

It also adds new variables to the `install_k3s` library function that are required for this test run to function properly.

- Related ticket: https://progress.opensuse.org/issues/112457
- Verification run: [TW kubectl](https://duck-norris.qam.suse.de/tests/9020) | [TW helm](https://duck-norris.qam.suse.de/tests/9021)

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -21,27 +21,37 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm install_oc);
 
 sub install_k3s {
-    my $k3s_version = get_var("CONTAINERS_K3S_VERSION", "v1.23.6+k3s1");
-    record_info('k3s', $k3s_version);
-    assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=$k3s_version sh -");
-    enter_cmd("k3s server &");
+  # Apply additional options. For more information see https://rancher.com/docs/k3s/latest/en/installation/install-options/#options-for-installation-with-script
+    my $k3s_version = get_var("CONTAINERS_K3S_VERSION");
+    if ($k3s_version) {
+        record_info('k3s', $k3s_version);
+        assert_script_run("export INSTALL_K3S_VERSION=$k3s_version");
+    }
+    assert_script_run("export INSTALL_K3S_SYMLINK=" . get_var('K3S_SYMLINK')) if (get_var('K3S_SYMLINK'));
+    assert_script_run("export INSTALL_K3S_BIN_DIR=" . get_var('K3S_BIN_DIR')) if (get_var('K3S_BIN_DIR'));
+    assert_script_run("export INSTALL_K3S_CHANNEL=" . get_var('K3S_CHANNEL')) if (get_var('K3S_CHANNEL'));
+    assert_script_run("curl -sfL https://get.k3s.io | sh -");
+    # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
+    sleep(20);    # Wait one iteration interval before checking because the server needs some time to boot-up
     script_retry("test -e /etc/rancher/k3s/k3s.yaml", delay => 20, retry => 10);
+    assert_script_run('systemctl is-active k3s');
     assert_script_run("k3s kubectl get node");
-    script_run("mkdir \$HOME/.kube");
-    script_run("rm \$HOME/.kube/config");
-    assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml \$HOME/.kube/config");
+    script_run("mkdir -p ~/.kube");
+    script_run("rm -f ~/.kube/config");
+    assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml ~/.kube/config");
 }
 
 sub uninstall_k3s {
-    assert_script_run("rm \$HOME/.kube/config");
+    assert_script_run("rm -f ~/.kube/config");
     assert_script_run("/usr/local/bin/k3s-uninstall.sh");
 }
 
 sub install_kubectl {
     if (is_sle) {
+        # kubectl is in the container module
+        add_suseconnect_product(get_addon_fullname('contm'));
         zypper_call("in kubernetes1.18-client");
-    }
-    else {
+    } else {
         zypper_call("in kubernetes-client");
     }
 }
--- a/tests/containers/kubectl.pm
+++ b/tests/containers/kubectl.pm
@@ -1,0 +1,136 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Test the kubectl utility
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use base 'consoletest';
+use testapi;
+use utils;
+use version_utils;
+use publiccloud::utils;
+use containers::k8s;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    install_kubectl();
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
+
+    # Prepare the webserver testdata
+    assert_script_run('mkdir -p /srv/www/kubectl');
+    assert_script_run("echo 'I am groot' > /srv/www/kubectl/index.html");
+
+    # Configure CSP/k3s. Only one CSP at a time is possible due to the conflicting configuration in .kube/config
+    assert_script_run('mkdir -p ~/.kube');
+    my $provider = get_var("KUBECTL_CLUSTER", "k3s");
+    if ($provider eq "k3s") {
+        install_k3s();
+
+        # Ensure k3s has not installed its own kubectl
+        assert_script_run("! stat /usr/local/bin/kubectl");
+        validate_script_output("whereis kubectl", sub { $_ !~ m/\/usr\/local\/bin/ });
+    } else {
+        die "Invalid KUBECTL_CLUSTER defined";
+    }
+
+    # Function to check if the output either has no resources or has at least two lines
+    # Two lines because one line is the header and the following lines are the data rows
+    validate_script_output("kubectl cluster-info", qr/Kubernetes control plane/);
+    record_info('Test get', 'State tests (get commands)');
+    # Check for default namespaces to be present
+    validate_script_output("kubectl get namespaces", qr/default/);
+    validate_script_output("kubectl get namespaces", qr/system/);
+    # Check service output
+    validate_script_output("kubectl get service --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
+    # At least the coredns pod must be present
+    validate_script_output("kubectl get pods --all-namespaces", sub { $_ =~ m/coredns/ });
+    # There must be at least one endpoint present
+    validate_script_output("kubectl get endpoints", sub { split(/\n/, $_) > 1; });
+    # There must be at least the coredns deployment present
+    validate_script_output("kubectl get deployments --all-namespaces", sub { $_ =~ m/coredns/ });
+    # Check if more than 1 replicasets are present
+    validate_script_output("kubectl get replicasets --all-namespaces --no-headers", sub { split(/\n/, $_) > 1; });
+    # Check ingresses, those are empty by default
+    validate_script_output("kubectl get ingresses --all-namespaces --no-headers", sub { $_ =~ m/No resources found/ });
+
+    ## Test the configuration
+    record_info('Test config', 'Configugration tests');
+    assert_script_run("kubectl config view");
+    validate_script_output("kubectl config view -o jsonpath={.clusters[].name}", qr/default/);
+    validate_script_output("kubectl config view -o jsonpath={.contexts[].name}", qr/default/);
+
+    ## Test the context configuration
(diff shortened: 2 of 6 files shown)


lib/containers/k8s.pm
```diff
@@ -21,19 +21,28 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm install_oc);
 
 sub install_k3s {
-    my $k3s_version = get_var("CONTAINERS_K3S_VERSION", "v1.23.6+k3s1");
-    record_info('k3s', $k3s_version);
-    assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=$k3s_version sh -");
-    enter_cmd("k3s server &");
```
grisu48, line 27: Note: The install script starts a k3s-server by default, unless `INSTALL_K3S_SKIP_START` is set to true. See https://rancher.com/docs/k3s/latest/en/installation/install-options/#options-for-installation-with-script for more details.

The same comment is also present in line 34 now.

lib/containers/k8s.pm
```diff
@@ -21,19 +21,28 @@ use registration qw(add_suseconnect_product get_addon_fullname);
...
+    assert_script_run("export INSTALL_K3S_CHANNEL=" . get_var('K3S_CHANNEL')) if (get_var('K3S_CHANNEL'));
+    assert_script_run("curl -sfL https://get.k3s.io | sh -");
+    # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
+    sleep(20);    # Wait one iteration interval before checking because the server needs some time to boot-up
     script_retry("test -e /etc/rancher/k3s/k3s.yaml", delay => 20, retry => 10);
+    assert_script_run('systemctl is-active k3s');
     assert_script_run("k3s kubectl get node");
-    script_run("mkdir \$HOME/.kube");
-    script_run("rm \$HOME/.kube/config");
+    assert_script_run("mkdir -p ~/.kube");
```
ilausuch, line 39: I  think this one and the following cannot be asset_script_run... we don't know if the directory exists previously or not. The same for the config file. Please run the helm tests to check it.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,69 @@
...
+use registration qw(cleanup_registration register_product add_suseconnect_product get_addon_fullname remove_suseconnect_product);
+use containers::k8s;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    # kubectl is in the container module (on SLES)
+    add_suseconnect_product(get_addon_fullname('contm')) if (is_sle);
+    zypper_call("-n in kubernetes-client");
```
ilausuch, line 25: Should we use the function install_kubectl that from k8s library?

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,69 @@
...
+
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
```
ilausuch, line 28: If we are going to use the install_kubectl will be interesting to move this line inside the k8s library after installing to have this benefit in all tests that are using this.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,69 @@
...
+        validate_script_output("whereis kubectl", sub { $_ !~ m/\/usr\/local\/bin/ });
+    } else {
+        die "Invalid KUBECTL_CLUSTER defined";
+    }
+
+    # Function to check if the output either has no resources or has at least two lines
+    # Two lines because one line is the header and the following lines are the data rows
+    validate_script_output("kubectl cluster-info", qr/Kubernetes control plane/);
+    # Check for default namespaces to be present
+    validate_script_output("kubectl get namespaces", qr/default/);
```
ilausuch, line 49: mm not sure that in all providers (e.g. AKS) this namespace will be created by default

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,69 @@
...
+    add_suseconnect_product(get_addon_fullname('contm')) if (is_sle);
+    zypper_call("-n in kubernetes-client");
+
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
+
+    # Configure CSP/k3s. Only one CSP at a time is possible due to the conflicting configuration in .kube/config
+    assert_script_run('mkdir -p ~/.kube');
+    my $provider = get_var("KUBECTL_CLUSTER", "k3s");
+    if ($provider eq "AKS") {
```
ilausuch, line 33: Not sure we could use AKS after seeing the checks (e.g. check if there are not ingresses). Maybe for this test should be removed, or the checks changes. In public kubernetes we could have concurrent tests that could create these resources and the test will fail. An other option is to create a namespace and check these things inside the namespace. Deleting the namespace at the end

ilausuch (changes requested):

ilausuch:
Nice test! only some suggestions and concerns

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,69 @@
...
+    # At least the coredns pod must be present
+    validate_script_output("kubectl get pods --all-namespaces", sub { $_ =~ m/coredns/ });
+    # There must be at least one endpoint present
+    validate_script_output("kubectl get endpoints", sub { split(/\n/, $_) > 1; });
+    # There must be at least the coredns deployment present
+    validate_script_output("kubectl get deployments --all-namespaces", sub { $_ =~ m/coredns/ });
+    # Check replicasets. For now allow no replicatests as well.
+    validate_script_output("kubectl get replicasets --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
+    # Check ingresses, those are likely to be empty.
+    validate_script_output("kubectl get ingresses --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
```
ilausuch, line 62: Should be interesting add other kubectl commands here: describe, get with json output, get with filtering, get with selecting output data. Also secrets and configmaps...

lib/containers/k8s.pm
grisu48: `mkdir -p` also returns 0 if the directory already exists, and `rm -f` also returns 0 if the file does not exists. But if you don't like it we can also use `script_run` here.

tests/containers/kubectl.pm
grisu48: Good point, I will move it there and also add the container module addition to the library function

tests/containers/kubectl.pm
grisu48: Yes, I know, when adding more providers we will need to tweak those checks accordingly. I don't know a better way here than to start with Tumbleweed, and then add provider by provider with the corresponding changes to the test cases. WDYT?

tests/containers/kubectl.pm
grisu48: I'm a bit lost here - where is the conflict?

We for sure need to adjust some of the checks later, but AFAICS the whole `kubectl` test run is read-only and does not create nor delete nor modify any resources?

tests/containers/kubectl.pm
grisu48: Indeed, the problem for those is that we need well defined objects that we then check for and we don't have them (yet).

tests/containers/kubectl.pm
grisu48: I think this output is only interesting in the test run where we test `kubectl`. In other test cases this spams the webui with unnecessary `record_info` boxes

tests/containers/kubectl.pm
grisu48: Wait, I think I know what you mean. This was meant as a template for future extensions. I commented it out and added a note to mark it as such. This is not (yet) intended to be used within the test code.

tests/containers/kubectl.pm
ilausuch: ok

tests/containers/kubectl.pm
ilausuch: Answering the previous question, maybe we could cover K3s now adding the others cloud providers in future PR. in this way creating your own namespace is not going to affect anything. I suggest to create the namespace adding the job id or some similar, to avoid overlapping on tests

ilausuch (approved):

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,68 @@
...
+    install_kubectl();
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
+
+    # Configure CSP/k3s. Only one CSP at a time is possible due to the conflicting configuration in .kube/config
+    assert_script_run('mkdir -p ~/.kube');
+    my $provider = get_var("KUBECTL_CLUSTER", "k3s");
+    # Template for the extension to CSPs
+    # Currently this test only runs with the 'k3s' provider. The following is to help us to extend the test case for each CSP.
+    #if ($provider eq "AKS") {
```
jlausuch, line 31: I would remove this code, the `else` statement is not even finished `els` :)

jlausuch (commented):
Looks quite good, although I am wondering why we are only testing `get` methods. 
Shouldn't we exercise kubectl cli more? 
Looking at https://kubernetes.io/docs/reference/kubectl/cheatsheet/ there is plenty of possibilities, I would say that we at least should try to `apply` a manifest to create a simple cluster using our cluster provider (`k3s` in this case).  

Also, how is this test triggered? We will need to set the variable `CONTAINER_RUNTIME` to hit `main_containers.pm` schedule, right? What would be the value of that variable?

tests/containers/kubectl.pm
grisu48: Ok, but the `els` is intentional, because that transforms into an `elsif` with the if from the next line :pineapple:

jlausuch:
> > Looks quite good, although I am wondering why we are only testing `get` methods. Shouldn't we exercise kubectl cli more? Looking at https://kubernetes.io/docs/reference/kubectl/cheatsheet/ there is plenty of possibilities, I would say that we at least should try to `apply` a manifest to create a simple cluster using our cluster provider (`k3s` in this case).
> 
> We can do this on the local `k3s` cluster. I have reservations to do this against one of our publiccloud clusters, as we would otherwise establish race condition checks and make sure that leftovers get cleaned up. If we cover this only for a local cluster, this should test `kubectl` sufficiently.

To me, testing kubectl on k3s only would be sufficient to cover the test case, this way, we can enable it in O3 and OSD having parallel coverage.  Testing it on Public Cloud providers in OSD later on is an extra, a "nice to have" but not mandatory. But I wouldn't lose coverage just because of this...
In any case, if there are problems if we try to enable it in AKS/EKS/GKE, we can always add the needed conditions when that happens. 

> 
> > Also, how is this test triggered? We will need to set the variable `CONTAINER_RUNTIME` to hit `main_containers.pm` schedule, right? What would be the value of that variable?
> 
> Look at the change in `lib/main_containers.pm`. We only need to specify the `KUBECTL_CLUSTER` variable and as stated in the description, the inclusion of this as a new job in O3 is the next step.

Yes and no. That only points to which condition you want to execute inside `main_containers.pm`, but in order to use `main_containers.pm`, the method [is_container_test()](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/main_containers.pm#L30) is called:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/products/sle/main.pm#L680
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/products/opensuse/main.pm#L269
And this method checks for `return get_var('CONTAINER_RUNTIME', 0);`

Your [VR](https://duck-norris.qam.suse.de/tests/9020) works because you are using `YAML_SCHEDULE=schedule/functional/extra_tests_textmode.yaml` which is a bit weird to me... 

I understand this method `is_container_test` is now self-explanatory, as we shouldn't need to explicitly give CONTAINER_RUNTIME for tests like helm or kubectl, but we can maybe fix this in another PR, but the fact is that you need this variable.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,98 @@
...
+    validate_script_output("kubectl get endpoints", sub { split(/\n/, $_) > 1; });
+    # There must be at least the coredns deployment present
+    validate_script_output("kubectl get deployments --all-namespaces", sub { $_ =~ m/coredns/ });
+    # Check replicasets. For now allow no replicatests as well.
+    validate_script_output("kubectl get replicasets --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
+    # Check ingresses, those are likely to be empty.
+    validate_script_output("kubectl get ingresses --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
+
+    # More extended tests where we modify the cluster change only happens on the local cluster, where those changes are ephermal
+    if ($provider eq "k3s") {
```
jlausuch, line 59: Shall we even move it out of the condition? For now we won't enable any tests in EKS/AKS/GKE, and if we do before we have proper clean-up in place (pcw) we can add the condition.

jlausuch (commented):
Would you mind adding some `record_info` for the groups of CLI commands?
e.g. (suggestion)
`record_info('Test 1', 'Get commands');`
or
`record_info('Test 2', 'Config commands');`
or
`record_info('Test 3', 'Create jobs/pods');`

PR #15176, opened 2022-07-05
Change softfailure to info for s390x shutdown

rakoenig:
Backend does not implement is_shutdown so there is not much chance to detect a shutdown from the GUI. Nevertheless we can replace the softfailure with just an info.

- Related ticket: https://progress.opensuse.org/issues/110437
- Needles: -
- Verification run: https://openqa.suse.de/tests/9072280#

### DIFF
--- a/lib/power_action_utils.pm
+++ b/lib/power_action_utils.pm
@@ -277,7 +277,7 @@ sub power_action {
         }
         elsif ($action eq 'poweroff') {
             if (is_backend_s390x) {
-                record_soft_failure('poo#58127 - Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.');
+                record_info('poo#58127', 'Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.');
                 select_console 'root-console';
                 enter_cmd "$action";
             }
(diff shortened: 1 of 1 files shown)


ge0r (approved):
LGTM!

ge0r:
> @ge0r have you overlooked my review or did you think it was not important or already resolved (which it wasn't)?

I did not overlook it, which is why I did not merge the PR, it did not make sense for me to repeat the commit message point since it was already addressed, so I marked it as ok as far as the implementation went, with a suggestion from my side.

okurz:
But my review included multiple points and only one was addressed. I would prefer if a review is taken seriously and not disregarded, especially not without a comment, e.g. "sorry, I need to merge this because whole production is down today and we urgently need this fixed and the git commit author has gone to vacation and nobody else can fix the typo in the comment that you flagged".

jknphy:
Hi @rakoenig as we agreed a time ago, after our initial investigation of the issue by a couple of team members in our squad we didn't find what was the problem, so we will be creating the `record_info` with a ticket for the tools teams's backlog. We should not reference the progress ticket 'Resolved', which by they way we use 'Close' in our flow.

Other than that, regarding the feedback, I didn't see the previous version of the PR, but now looks sort of ok, I would only had added 'shutdown from GUI' to title/commit to make it more clear probably. It is a good suggestion from Oliver to use some automatic tool, as I can see that we sometimes forget to match PR title and PR commit, please consider that.

PR #15173, opened 2022-07-04
chromium: Type each character as the screen changes

kalikiana:
See: https://progress.opensuse.org/issues/109737

PR #15180, opened 2022-07-06
chromium: Ensure each character is typed in the address bar without waiting too long

okurz:
Needs https://github.com/os-autoinst/os-autoinst/pull/2109

Related progress issue: https://progress.opensuse.org/issues/109737


okurz:
@Martchus @kalikiana we would still profit from this, don't we?

Martchus:
No, it is the same as https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15173 which has already been merged and the `no_wait` is not required here (as it is always used for typing with `wait_screen_change` anyways).

PR #15183, opened 2022-07-07
Enhance VT performance run to easy to use

XGWang0:
1.Show detail info for screen name; 
2. kill screen session when repeatedly trigger

- Verification run: http://openqa.qa2.suse.asia/tests/49664#


XGWang0:
@alice-suse @museQi Please help to review and merge, thanks

PR #15185, opened 2022-07-07
Use yast2 registration instead of SUSEConnect

ge0r:
Instead of using the SUSEConnect command line tool to cleanup the registration and also checking registration status, we should use the YaST registration module option 'Register again'.

- Related ticket: https://progress.opensuse.org/issues/112151
- Needles: -
- Verification run: https://openqa.suse.de/tests/overview?build=ge0r%2Fos-autoinst-distri-opensuse%23poo-112151&distri=sle

### DIFF
--- a/tests/console/yast2_registration.pm
+++ b/tests/console/yast2_registration.pm
@@ -1,21 +1,21 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2019-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: yast2-registration
-# Summary: check if an unregistered system can be registered and if
+# Summary: check if a registered system can be re-registered and if
 #          enabling and disabling extensions correctly work.
 # - Install yast2-registration
-# - Cleanup registration (SUSEConnect --cleanup)
 # - Launch yast2 registration
+# - Re-register system
 # - Fill email and registration code (using system variables)
 # - Enable web and scripting modules
 # - Accept license agreement
 # - Accept install summary
 # - Accept automatic changes
 # - Wait for installation report
-# - Check registration status by checking output of "SUSEConnect --status-text"
+# - Check registration status
 # Maintainer: Paolo Stivanin <pstivanin@suse.com>
 
 use base "opensusebasetest";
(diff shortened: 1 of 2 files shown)


tests/console/yast2_registration.pm
```diff
@@ -61,15 +61,16 @@ sub run {
...
+    assert_screen 'yast2_registration-overview';
+    send_key "alt-e";
+    assert_screen 'yast2_registration-registration-page-registered';
     register_system_and_add_extension;
     wait_serial("$module_name-0", 200) || die "'yast2 $module_name' didn't finish";
 
-    assert_script_run "SUSEConnect --status-text |grep -A3 -E 'SUSE Linux Enterprise Server|Web and Scripting Module' | grep -qE '^\\s+Registered'";
+    # Check via YaST if the system is already registered
+    $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'registration');
+    assert_screen 'yast2_registration-overview';
```
coolgw, line 73: Before this change the SUSEConnect check the register status of module such as web and scripting, but now ONLY check the yast register status? I feel loss some check points compare with old implementation.

ge0r: So I have tried checking registration state of individual modules and, although the idea seems good, checking the list of modules with needles can get flaky given the list of products that this is supposed to run on. For instance SLES 12-SP5 has a completely different module listing than SLES 15-SP4 as can be seen [here](https://openqa.suse.de/tests/9094062#step/yast2_registration/21).
So although we do lose this specific coverage compared to the old command line check, for the sake of stability and consistency of the test on all products, I would suggest that we do not check individual module registration with needles.

coolgw: Yes,  check modules status using yast needle will trigger even more trouble compare with command.
OK for me if you think loss coverage can be acceptable.

jknphy:
let's add label `qe-yast` to filter better our PR. wdyt? By name doesn't work anymore, because team members in rotation have PRs unrelated with YaST scope.

tests/console/yast2_registration.pm
jknphy: let's see what is the less evil, if the needling in product in maintenance which should be sort of stable, or checking with command SUSEConnect when there are network glitches.

jknphy (approved):

PR #15190, opened 2022-07-07
Update lynis

paolostivanin:
On 15-SP4, the test was failing because it found `ERROR|WEAK|UNSAFE` even though the baseline file had these items in it.
This happens when the `$baseline` and `$current` file have one or more differences, because in such case only the `$current` file is being checked and, since there are `ERROR|WEAK|UNSAFE` items, the test fails.
This PR add some exceptions when checking for `lynis_fail` based on what is already present in the baseline file.

VRs:
- 15-SP3: 
  - https://openqa.suse.de/tests/9085821
  - https://openqa.suse.de/tests/9085822
- 15-SP4:
  - https://openqa.suse.de/tests/9085820
  - https://openqa.suse.de/tests/9085809
  - https://openqa.suse.de/tests/9085810

### DIFF
--- a/lib/lynis/lynistest.pm
+++ b/lib/lynis/lynistest.pm
@@ -12,6 +12,7 @@ use warnings;
 use testapi;
 use utils;
 use autotest;
+use version_utils 'is_sle';
 
 use base 'consoletest';
 use LTP::TestInfo 'testinfo';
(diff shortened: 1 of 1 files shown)


rfan1:
CI check failed.

paolostivanin:
The CI fails, but I don't get why:

```
 ~/git_projs/os-autoinst-distri-opensuse [fix-lynis-data L|✔] 
08:26 $ ./tools/tidy --check
✔ ~/git_projs/os-autoinst-distri-opensuse [fix-lynis-data L|✔] 
08:26 $ 
```

lib/lynis/lynistest.pm
```diff
@@ -205,6 +206,52 @@ sub parse_lynis_section_content {
     return @section_content;
 }
 
+sub check_exceptions {
```
rfan1, line 209: LGTM, only minor comment here,
could you please add some description why so many 'exceptions' here?

rfan1 (approved):
LGTM

lib/lynis/lynistest.pm
paolostivanin: sure thing!

jouyingbin (commented):
LGTM. :)

PR #15194, opened 2022-07-08
Test realhw hdmi output

asdil12:
https://progress.opensuse.org/issues/111398
Verification: https://openqa.suse.de/tests/10049809#step/prepare_firstboot/2
Requires: https://github.com/os-autoinst/os-autoinst/pull/2220

This is as of now only enbled for the RPi4 SUT (via `worker.ini`) as only that one is equipped with an HDMI to USB grabber connected to the worker.


ggardet (approved):
LGTM, but I cannot check the verification run.

Martchus (commented):
Looks generally good. The commit message could have at least the ticket reference as well.

PR #15195, opened 2022-07-08
Update repos and meta before invoke yast

b10n1k:
`deactivate_kdump_cli` breaks as once the YaST run for first time will prompt
to trust signing key from Basesystem-Update module. The simpliest solution is
to run `zypper ref` before `yast`. Other solutions would be to go through
needling or import the key directly with `rpm --import`. However i found
zypper convenience because i do not need to care about the specifics of the key.

- Related ticket: https://progress.opensuse.org/issues/113351
- Verification run:
sle-15-SP4-Server-DVD-Incidents-x86_64-Build:24623:python3-mau-sles-sys-param-check@64bit-2gbram -> https://openqa.suse.de/t9094057

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -221,6 +221,8 @@ sub activate_kdump_cli {
 
 # Deactivate kdump using yast command line interface
 sub deactivate_kdump_cli {
+    # Solution to poo113351. Avoid to use needles to solve this case.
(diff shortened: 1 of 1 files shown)


alvarocarvajald (approved):
LGTM

okurz:
merging this as I stumbled over https://openqa.suse.de/tests/9093730#comments where this PR was linked to so hoping this helps.

PR #15198, opened 2022-07-11
[wip]Enhance wait_countdown_stop logic to retry more

rfan1:
The system will auto-reboot after installation
within 10s, current logic is send 'alt-s' key
to stop it every 3s. set it to 1s then we can
retry more times on low performance setups.

- Related ticket: https://progress.opensuse.org/issues/113396
- Verification run:  [run the same job for 5 times]
http://openqa.suse.de/t9101633
http://openqa.suse.de/t9101634
http://openqa.suse.de/t9101635
http://openqa.suse.de/t9101636
http://openqa.suse.de/t9101637

### DIFF
--- a/tests/installation/await_install.pm
+++ b/tests/installation/await_install.pm
@@ -194,7 +194,7 @@ sub run {
         # A single changing digit is only a minor change, overide default
         # similarity level considered a screen change
         my $minor_change_similarity = 55;
-        while ($counter-- and wait_countdown_stop(3, $minor_change_similarity)) {
+        while ($counter-- and wait_countdown_stop(1, $minor_change_similarity)) {
             record_info('workaround', "While trying to stop countdown we saw a screen change, retrying up to $counter times more");
         }
     }


rfan1:
@jknphy 
Can I ask for your kindly help to review my PR from Yast team?

jknphy:
I would suggest to ping who created this method, in YaST squad we work with libyui REST mainly so we normally don't touch these modules. IMO it looks a bit of an overkill to be sending key each second, we might find some strange situation. I would recommend to add a condition to change this timeout only in those cased needed, not for all. There is also a way to setup the timeout iirc via bootloader param to give a fixed behavior, that would change the default flow of installation, but as this logic works for other non-slow arch that might be a solution as well.

rfan1:
> I would suggest to ping who created this method, in YaST squad we work with libyui REST mainly so we normally don't touch these modules. IMO it looks a bit of an overkill to be sending key each second, we might find some strange situation. I would recommend to add a condition to change this timeout only in those cased needed, not for all. There is also a way to setup the timeout iirc via bootloader param to give a fixed behavior, that would change the default flow of installation, but as this logic works for other non-slow arch that might be a solution as well.

Thanks much expert! let me try to find a way to change the reboot timeout there.

jknphy:
> > I would suggest to ping who created this method, in YaST squad we work with libyui REST mainly so we normally don't touch these modules. IMO it looks a bit of an overkill to be sending key each second, we might find some strange situation. I would recommend to add a condition to change this timeout only in those cased needed, not for all. There is also a way to setup the timeout iirc via bootloader param to give a fixed behavior, that would change the default flow of installation, but as this logic works for other non-slow arch that might be a solution as well.
> 
> Thanks much expert! let me try to find a way to change the reboot timeout there.

@rakoenig do you remember how you did it last time?

rakoenig:
> @rakoenig do you remember how you did it last time?

There are 2 variables that come to my mind.

- TIMEOUT_SCALE was made for slwoer architectures so that we can extend timeouts there.
- REBOOT_TIMEOUT could be set to 0, so a confirmation would be needed to reboot, maybe this is a better approach than trying to send "stop" commands. WDYT?

foursixnine:
> > @rakoenig do you remember how you did it last time?
> 
> There are 2 variables that come to my mind.
> 
>     * TIMEOUT_SCALE was made for slwoer architectures so that we can extend timeouts there.
> 
>     * REBOOT_TIMEOUT could be set to 0, so a confirmation would be needed to reboot, maybe this is a better approach than trying to send "stop" commands. WDYT?

I was about to suggest this `reboot_timeout=0`, since I think in almost all cases with the old installer, we collect logs

rfan1:
> > @rakoenig do you remember how you did it last time?
> 
> There are 2 variables that come to my mind.
> 
> * TIMEOUT_SCALE was made for slwoer architectures so that we can extend timeouts there.
> * REBOOT_TIMEOUT could be set to 0, so a confirmation would be needed to reboot, maybe this is a better approach than trying to send "stop" commands. WDYT?

It is SLES15, I am not sure it can support this linuxrc parameter,
`lib/bootloader_setup.pm:504:    push @params, 'reboot_timeout=' . get_var('REBOOT_TIMEOUT', 0) unless (is_leap('<15.2') || is_sle('<15-SP2'));
`

rfan1:
> TIMEOUT_SCALE

foursixnine:
> > > @rakoenig do you remember how you did it last time?
> > 
> > 
> > There are 2 variables that come to my mind.
> > 
> > * TIMEOUT_SCALE was made for slwoer architectures so that we can extend timeouts there.
> > * REBOOT_TIMEOUT could be set to 0, so a confirmation would be needed to reboot, maybe this is a better approach than trying to send "stop" commands. WDYT?
> 
> It is SLES15, I am not sure it can support this linuxrc parameter, `lib/bootloader_setup.pm:504: push @params, 'reboot_timeout=' . get_var('REBOOT_TIMEOUT', 0) unless (is_leap('<15.2') || is_sle('<15-SP2')); `

@jknphy @rakoenig is libyui available for 15 and 15-sp1?

jknphy:
> am not sure it can support this linuxrc parameter,

libyui REST is available from 15-SP3/15.3 and above.

rfan1:
> 
'TIMEOUT_SCALE'
Seems we didn't reach the timeout value, I have checked 20+ builds, and it can complete the installation process everytime

jknphy:
I don't know how timeout_scale can help, basically you want to catch the popup, so you need to check faster.
I would identify exactly the product and the arch and will change the timeout only for that with a comment why you would not do in a different way.

rfan1:
> I don't know how timeout_scale can help, basically you want to catch the popup, so you need to check faster. I would identify exactly the product and the arch and will change the timeout only for that with a comment why you would not do in a different way.

Thanks much! we have to stop the auto_reboot process within 10 seconds. ideally,  we can send more "alt-s“ keys to stop it on aarch platform [the only affected platform so far]

foursixnine:
@rfan1 Let's try something different, can you do the same runs, excluding the `[logs_from_installation_system](https://openqa.suse.de/tests/9101633/modules/logs_from_installation_system/steps/1/src)` module? schedule 100 and let's wait for the results :)

Might not be ideal, but it could solve the problem... since checking every second will most likely increase the load on host system if multiple builds are running at the same time

rfan1:
> @rfan1 Let's try something different, can you do the same runs, excluding the `[logs_from_installation_system](https://openqa.suse.de/tests/9101633/modules/logs_from_installation_system/steps/1/src)` module? schedule 100 and let's wait for the results :)
> 
> Might not be ideal, but it could solve the problem... since checking every second will most likely increase the load on host system if multiple builds are running at the same time

Nice comment, Let me try!
However, it might miss some steps like "save and upload y2logs"

PR #15199, opened 2022-07-11
Fix the SAP zypper migration

hjluo:
In SAP zypper migration for 15sp4, It has no zypper-migration-plugin
package. We need to skip the check before 'zypper migration'.

- Related ticket: https://progress.opensuse.org/issues/113465
- Needles: N/A
- Verification run: 
  https://openqa.nue.suse.com/t9109274
  regression:
 https://openqa.nue.suse.com/t9109275


tinawang123 (approved):
LGTM

coolgw (approved):

PR #15204, opened 2022-07-11
Fix security_audit on 12-SP5

paolostivanin:
https://openqa.suse.de/tests/9103321

### DIFF
--- a/tests/security/audit/ausearch.pm
+++ b/tests/security/audit/ausearch.pm
@@ -33,7 +33,8 @@ sub run {
 
     # Extract pid from output log
     script_run("tail -1 $tmp_output > $tmp_backup");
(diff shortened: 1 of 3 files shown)


tests/security/audit/ausearch.pm
```diff
@@ -33,7 +33,8 @@ sub run {
 
     # Extract pid from output log
     script_run("tail -1 $tmp_output > $tmp_backup");
-    script_run("cat $tmp_backup | cut -d '=' -f 9 > $tmp_output");
+    my $cut_index = is_sle('<=12-SP5') ? 7 : 9;
```
jouyingbin, line 36: I'm not sure the 7 and 9 are different here, but I think it's OK after checking.

jouyingbin (approved):
LGTM

PR #15205, opened 2022-07-11
Add TRENTO_VERSION variable

mpagot:
Add TRENTO_VERSION variable to set the expected version tag on the web
interface.Add some documentation for all the TRENTO variables.

- Verification run: http://1c242.qa.suse.de/tests/641  The key point there is in http://1c242.qa.suse.de/tests/641#settings where TRENTO_VERSION is visible and http://1c242.qa.suse.de/tests/641/file/test_trento_web-cypress.env.json that has the version to from that variable. This file is the expectation file provided as input to the  cypress test engine

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -203,6 +203,7 @@ sub k8s_logs {
             }
         }
     }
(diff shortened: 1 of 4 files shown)


ricardobranco777 (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #15207, opened 2022-07-12
Improve error message when using libyui REST below SLE-15-SP3

skliu:
Improve  libyui REST error message. 

- Related ticket: https://progress.opensuse.org/issues/112832
- Verification run: http://d218.qam.suse.de/tests/42

### DIFF
--- a/tests/installation/setup_libyui.pm
+++ b/tests/installation/setup_libyui.pm
@@ -1,4 +1,4 @@
-# Copyright 2020-2021 SUSE LLC
+# Copyright 2020-2022 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 # Summary: Module to set up the environment for using libyui REST API with the
@@ -13,8 +13,10 @@ use Utils::Backends;
 use Utils::Architectures;
 use testapi;
 use YuiRestClient;
+use version_utils 'is_sle';
 
 sub run {
+    die 'Leap 15.2 and below and SLE 15-SP2 and below do not have libyui-REST, exit now.' if (is_sle('<=15-SP2') || is_leap('<=15.2'));
     die 'Module requires YUI_REST_API variable to be set.', unless get_var('YUI_REST_API');
     my $app = YuiRestClient::get_app(installation => 1, timeout => 120, interval => 1);
     my $port = $app->get_port();


jknphy:
could you please do a couple of things:
- squash both commits
- shorten the title of the PR and the commit message, instead of `when attempting to use` just `when using` would be ok.
I don't remember exactly the rules in the repo for the length of this text, but git normally indicate it when creating a commit (I'm not following strictly but I always try to simplify keeping still meaningful).

jknphy (commented):
I gave some review in the morning but forgot to push "Submit review".

tests/installation/setup_libyui.pm
```diff
@@ -13,8 +13,10 @@ use Utils::Backends;
 use Utils::Architectures;
 use testapi;
 use YuiRestClient;
+use version_utils 'is_sle';
```
jknphy, line 16: update date in file copyright, please.

tests/installation/setup_libyui.pm
```diff
@@ -13,9 +13,10 @@ use Utils::Backends;
...
 
 sub run {
-    die 'Module requires YUI_REST_API variable to be set.', unless get_var('YUI_REST_API');
+    die 'Leap 15.2 and below and SLE 15-SP2 and below do not have libyui-rest, exit now.' if (is_sle('<=15-SP2') || is_leap('<=15.2') || get_var('YUI_REST_API'));
```
ge0r, line 19: we should not die if `get_var('YUI_REST_API')` returns true.
Maybe negate the condition?

skliu: done

ge0r: Now that I think about it, in case the YUI_REST_API variable is not there, the message `Leap 15.2 and below and SLE 15-SP2 and below do not have libyui-rest, exit now.` is not very helpful to the user.
I think that it would be better to have it as you originally had it, first checking for the `YUI_REST_API` variable, and then checking the version. At least this is how I think about it.

skliu: How about this: 'The SUT version is below Leap 15.2 or SLE 15-SP2, or the variable YUI_REST_API is not set, exit now.'

ge0r: For me this is more complicated, it should give an error that corresponds to each case. Still let's see what others say

jknphy: 2 `die`s is preferable for sure.

skliu: re-committed

DimStar77:
fail

`# Test died: Undefined subroutine &setup_libyui::is_leap called at opensuse/tests/installation/setup_libyui.pm line 19.`

https://openqa.opensuse.org/tests/2470499#step/setup_libyui/2

jknphy:
thanks for the revert! we missed that that lib was delete in the last moment.

PR #15213, opened 2022-07-13
Schedule SLE Micro toolbox container tests

jlausuch:
Includes:
  - Automation for SLEM tests which publish updated SLEM HDD
  - toolbox,pm test updated to use the toolbox image under test

Related ticket: https://progress.opensuse.org/issues/113036
VRs:
- HDD creation with system update: [slem 5.1](https://openqa.suse.de/tests/overview?distri=sle-micro&version=5.1&build=GM&groupid=377) [slem 5.2](https://openqa.suse.de/tests/overview?distri=sle-micro&version=5.2&build=GM&groupid=377)
- Toolbox test with specific image: [slem 5.1](https://openqa.suse.de/tests/9124243) [slem 5.2](https://openqa.suse.de/tests/9124242)

### DIFF
--- a/tests/microos/toolbox.pm
+++ b/tests/microos/toolbox.pm
@@ -39,6 +39,20 @@ sub run {
     select_console 'root-console';
     $self->create_user;
 
+    my $toolbox_image_to_test = get_var('CONTAINER_IMAGE_TO_TEST');
+
+    if ($toolbox_image_to_test) {
+        # We need to extract the registry from the full image uri, e.g.
+        # From registry.suse.de/suse/sle-15-sp3/update/products/microos51/update/cr/images/suse/sle-micro/5.1/toolbox:latest
+        # registry = registry.suse.de
+        # $image = suse/sle-15-sp3/update/products/microos51/update/cr/images/suse/sle-micro/5.1/toolbox:latest
+        (my $registry = $toolbox_image_to_test) =~ s/\/.*//;
+        (my $image = $toolbox_image_to_test) =~ s/^[^\/]*\///;
+        assert_script_run "echo REGISTRY=$registry > /etc/toolboxrc";
+        assert_script_run "echo IMAGE=$image >> /etc/toolboxrc";
+        record_info('toolboxrc', script_output('cat /etc/toolboxrc'));
+    }
+
     # Display help
     assert_script_run 'toolbox -h';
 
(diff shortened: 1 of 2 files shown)


tests/microos/toolbox.pm
```diff
@@ -39,6 +39,16 @@ sub run {
     select_console 'root-console';
     $self->create_user;
 
+    my $toolbox_image_to_test = get_var('CONTAINER_IMAGE_TO_TEST');
+
+    if ($toolbox_image_to_test) {
+        (my $registry = $toolbox_image_to_test) =~ s/\/.*//;
```
grisu48, line 45: A small comment describing what those cryptic regexes are would be very helpful for our future selves as regexes are typically write-only :wink:

tests/microos/toolbox.pm
```diff
@@ -76,25 +86,28 @@ sub run {
...
 
-    record_info 'Test', 'Pulling toolbox image from different registry';
-    # Switch default registries for openSUSE MicroOS and SLE Micro
-    if (is_sle_micro) {
-        assert_script_run 'echo -e "REGISTRY=registry.opensuse.org\nIMAGE=opensuse/toolbox" > ~/.toolboxrc';
-        validate_script_output 'toolbox -r cat /etc/os-release', sub { m/opensuse/ }, timeout => 180;
-    } else {
-        assert_script_run 'echo -e "REGISTRY=registry.suse.com\nIMAGE=suse/sle-micro/5.0/toolbox" > ~/.toolboxrc';
-        validate_script_output 'toolbox -r cat /etc/os-release', sub { m/sles/ }, timeout => 180;
+    unless ($toolbox_image_to_test) {
```
grisu48, line 89: Why are we limiting the toolbox tests to the cases when we have a custom toolbox URI specified? Is toolbox not being tested by default also for SLEM?

grisu48 (commented):
Very good work! I have two questions, but then LGTM

tests/microos/toolbox.pm
jlausuch: Yes, this module is executed in regular SLEM container jobs, but using the default image which comes from registry.suse.com.

I put this `unless` here to skip this checks which are testing functionality of `toolboxrc` using different parameters, which is what we need to test the toolbox image under test..

tests/microos/toolbox.pm
jlausuch: done

tests/microos/toolbox.pm
grisu48: Ahh, I see. The actual tests are coming later. Thnx for the clarification!

mloviska (approved):

PR #15215, opened 2022-07-13
Add kubectl test run

grisu48:
Add a test for `kubectl` on Tumbleweed. This PR adds the test code. The modifications to the job groups on O3 will be done after merging.

It also adds new variables to the `install_k3s` library function that are required for this test run to function properly.

Related ticket: https://progress.opensuse.org/issues/112457  
Verification run: [TW kubectl](https://duck-norris.qam.suse.de/tests/10059) | [TW helm](https://duck-norris.qam.suse.de/tests/10055)

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -21,27 +21,42 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm install_oc);
 
 sub install_k3s {
-    my $k3s_version = get_var("CONTAINERS_K3S_VERSION", "v1.23.6+k3s1");
-    record_info('k3s', $k3s_version);
-    assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=$k3s_version sh -");
-    enter_cmd("k3s server &");
+  # Apply additional options. For more information see https://rancher.com/docs/k3s/latest/en/installation/install-options/#options-for-installation-with-script
+    my $k3s_version = get_var("CONTAINERS_K3S_VERSION");
+    if ($k3s_version) {
+        record_info('k3s', $k3s_version);
+        assert_script_run("export INSTALL_K3S_VERSION=$k3s_version");
+    }
+    assert_script_run("export INSTALL_K3S_SYMLINK=" . get_var('K3S_SYMLINK')) if (get_var('K3S_SYMLINK'));
+    assert_script_run("export INSTALL_K3S_BIN_DIR=" . get_var('K3S_BIN_DIR')) if (get_var('K3S_BIN_DIR'));
+    assert_script_run("export INSTALL_K3S_CHANNEL=" . get_var('K3S_CHANNEL')) if (get_var('K3S_CHANNEL'));
+    assert_script_run("curl -sfL https://get.k3s.io | sh -");
+    # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
+    sleep(20);    # Wait one iteration interval before checking because the server needs some time to boot-up
     script_retry("test -e /etc/rancher/k3s/k3s.yaml", delay => 20, retry => 10);
+    assert_script_run('systemctl is-active k3s');
     assert_script_run("k3s kubectl get node");
-    script_run("mkdir \$HOME/.kube");
-    script_run("rm \$HOME/.kube/config");
-    assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml \$HOME/.kube/config");
+    script_run("mkdir -p ~/.kube");
+    script_run("rm -f ~/.kube/config");
+    assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml ~/.kube/config");
 }
 
 sub uninstall_k3s {
-    assert_script_run("rm \$HOME/.kube/config");
+    assert_script_run("rm -f ~/.kube/config");
     assert_script_run("/usr/local/bin/k3s-uninstall.sh");
 }
 
 sub install_kubectl {
     if (is_sle) {
-        zypper_call("in kubernetes1.18-client");
-    }
-    else {
+        # kubectl is in the container module
+        add_suseconnect_product(get_addon_fullname('contm'));
+        # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
+        if (is_sle(">15-SP1")) {
+            zypper_call("in kubernetes1.18-client");
+        } else {
+            zypper_call("in kubernetes-client");
+        }
+    } else {
         zypper_call("in kubernetes-client");
     }
 }
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -200,13 +200,13 @@ sub load_container_tests {
         } elsif (get_var('REPO_BCI')) {
             loadtest 'containers/host_configuration';
             loadtest 'containers/bci_repo';
-        }
-        else {
+        } else {
             # Container Host tests
             load_host_tests_podman($run_args) if (/podman/i);
             load_host_tests_docker($run_args) if (/docker/i);
             load_host_tests_containerd_crictl() if (/containerd_crictl/i);
             load_host_tests_containerd_nerdctl() if (/containerd_nerdctl/i);
+            loadtest('containers/kubectl') if (/kubectl/i);
             load_host_tests_helm($run_args) if (/helm/i);
         }
     }
--- a/tests/containers/kubectl.pm
+++ b/tests/containers/kubectl.pm
@@ -1,0 +1,140 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Test the kubectl utility
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+use version_utils;
+use publiccloud::utils;
+use containers::k8s;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    install_kubectl();
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
+
+    # Prepare the webserver testdata
+    assert_script_run('mkdir -p /srv/www/kubectl');
+    assert_script_run("echo 'I am Groot' > /srv/www/kubectl/index.html");
+
+    # Configure CSP/k3s. Only one CSP at a time is possible due to the conflicting configuration in .kube/config
+    assert_script_run('mkdir -p ~/.kube');
+    my $provider = get_var("KUBECTL_CLUSTER", "k3s");
+    if ($provider eq "k3s") {
+        install_k3s();
+
+        # Ensure k3s has not installed its own kubectl
+        assert_script_run("! stat /usr/local/bin/kubectl");
+        validate_script_output("whereis kubectl", sub { $_ !~ m/\/usr\/local\/bin/ });
+    } else {
+        die "Invalid or unsupported KUBECTL_CLUSTER";
+    }
+
+    # Function to check if the output either has no resources or has at least two lines
+    # Two lines because one line is the header and the following lines are the data rows
+    validate_script_output("kubectl cluster-info", qr/Kubernetes control plane/);
+    record_info('Testing: get/smoketest', 'State tests (get commands)');
+    # Check for default namespaces to be present
+    validate_script_output("kubectl get namespaces", qr/default/);
+    validate_script_output("kubectl get namespaces", qr/system/);
+    # Check service output
+    validate_script_output("kubectl get service --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
+    # At least the coredns pod must be present
+    validate_script_output("kubectl get pods --all-namespaces", sub { $_ =~ m/coredns/ });
+    # There must be at least one endpoint present
+    validate_script_output("kubectl get endpoints", sub { split(/\n/, $_) > 1; });
+    # There must be at least the coredns deployment present
+    validate_script_output("kubectl get deployments --all-namespaces", sub { $_ =~ m/coredns/ });
+    # Check if more than 1 replicasets are present
+    validate_script_output("kubectl get replicasets --all-namespaces --no-headers", sub { split(/\n/, $_) > 1; });
+    # Check ingresses, those are empty by default
+    validate_script_output("kubectl get ingresses --all-namespaces --no-headers", sub { $_ =~ m/No resources found/ });
+
+    ## Test the configuration
+    record_info('Testing: config', 'Configugration tests');
+    assert_script_run("kubectl config view");
+    validate_script_output("kubectl config view -o jsonpath={.clusters[].name}", qr/default/);
+    validate_script_output("kubectl config view -o jsonpath={.contexts[].name}", qr/default/);
+
+    ## Test the context configuration
+    assert_script_run("kubectl config get-contexts");
+    validate_script_output("kubectl config get-contexts", sub { $_ =~ m/default/ });
+    validate_script_output("kubectl config current-context", sub { $_ =~ m/default/ });
+    assert_script_run("kubectl config use-context default");
+    assert_script_run("kubectl config set-context --current --namespace=default");
+    assert_script_run("kubectl config set-context localhost --user=root --namespace=default");
+    assert_script_run("kubectl config use-context localhost");
+    assert_script_run("kubectl config use-context default");
+    assert_script_run("kubectl config unset contexts.localhost");
+
+    ## Test jobs
+    record_info('Testing: jobs and pods', 'job and pod test runs');
+    assert_script_run('kubectl create job sayhello --image=busybox:1.28 -- echo "Hello World"');
+    assert_script_run('kubectl create job gimme-date --image=busybox -- date');
+    validate_script_output("kubectl get jobs --no-headers", qr/sayhello/);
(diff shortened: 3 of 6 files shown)


grisu48:
This PR is a reboot of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15174 to start from a fresh state after considerable changes have been applied. All open remarks have been tried to incorporate, so that we can start fresh without looking back.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,136 @@
...
+    # At least the coredns pod must be present
+    validate_script_output("kubectl get pods --all-namespaces", sub { $_ =~ m/coredns/ });
+    # There must be at least one endpoint present
+    validate_script_output("kubectl get endpoints", sub { split(/\n/, $_) > 1; });
+    # There must be at least the coredns deployment present
+    validate_script_output("kubectl get deployments --all-namespaces", sub { $_ =~ m/coredns/ });
+    # Check if more than 1 replicasets are present
+    validate_script_output("kubectl get replicasets --all-namespaces --no-headers", sub { split(/\n/, $_) > 1; });
+    # Check ingresses, those are empty by default
+    validate_script_output("kubectl get ingresses --all-namespaces --no-headers", sub { $_ =~ m/No resources found/ });
```
pdostal, line 60: Again, what do we gain by checking there is no ingress?

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,136 @@
...
+    assert_script_run('kubectl scale --replicas=5 deployments/nginx-deployment');
+    assert_script_run('kubectl wait deployment nginx-deployment --for condition=available --timeout=300s', timeout => 330);
+    validate_script_output('kubectl describe deployments/nginx-deployment | grep Replicas', sub { $_ =~ m/5 desired.*5 total/ });
+    # Scale in
+    assert_script_run('kubectl scale --replicas=2 deployments/nginx-deployment');
+    assert_script_run('kubectl wait deployment nginx-deployment --for condition=available --timeout=300s', timeout => 330);
+    validate_script_output('kubectl describe deployments/nginx-deployment | grep Replicas', sub { $_ =~ m/2 desired.*2 total/ });
+    # Test the port-forwarding and the webserver
+    my $pid = background_script_run('kubectl port-forward deploy/nginx-deployment 8008:80');
+    sleep(10);    # I don't see a better way to wait for the forward to be ready :-(
```
pdostal, line 107: You may do `script_retry('curl` as it ends with `0` only when the forwarding works.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,136 @@
...
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+
+    # Try to collect as much information about kubernetes as possible
+    script_run('kubectl describe deployments');
+    script_run('kubectl describe services');
+    script_run('kubectl describe pods');
+}
```
pdostal, line 134: Also delete what you created
```
script_run('kubectl delete -f service.yml');
script_run('kubectl delete -f deployment.yml');
```

pdostal (commented):
Nice work @grisu48 !

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,136 @@
...
+
+    # Function to check if the output either has no resources or has at least two lines
+    # Two lines because one line is the header and the following lines are the data rows
+    validate_script_output("kubectl cluster-info", qr/Kubernetes control plane/);
+    record_info('Test get', 'State tests (get commands)');
+    # Check for default namespaces to be present
+    validate_script_output("kubectl get namespaces", qr/default/);
+    validate_script_output("kubectl get namespaces", qr/system/);
+    # Check service output
+    validate_script_output("kubectl get service --all-namespaces", sub { $_ =~ m/No resources found/ || split(/\n/, $_) > 1; });
```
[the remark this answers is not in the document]
grisu48: This acts as a smoketest for the get method, so I'd say yes, it is useful. Not necessary but useful.

tests/containers/kubectl.pm
grisu48: Same as above - perliminary smoke test to allow to fail fast.

tests/containers/kubectl.pm
grisu48: I tried this at first, but it is not really better:

```
sleep(10); # wait one iteration to not waste resources
script_retry('curl ... ');
```

Arguably this is better because script_retry tries multiple times in case it is not yet ready. The 10 seconds are rather generous, so missing that timeframe should never be an issue. So, do we really gain something from switching to `script_retry` here? WDYT?

pdostal: I am used to try to avoid `sleep()` - that's it. From [documentation](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style):
> Avoid sleep(): Do not use sleep() or simulate a sleep-like functionality. This introduces the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You can use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)). sleep() should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. We are waiting because some condition changes some time ... keyword(s) is/are some time. It is important to know the exact condition that the code should wait for and check for that instead of delaying program execution for an arbitrary amount of time. For openQA tests in general just try to do what you as a human user would also do. You would not look on your watch waiting for exactly 100s before you look at the screen again, right? :)

tests/containers/kubectl.pm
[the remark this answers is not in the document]
pdostal: We can keep this for now. But I have a feeling that in some k8s distributions there will already be one or two services present.

tests/containers/kubectl.pm
grisu48: You are right from a principle standpoint, but there is no rule without exception.

The problem that you have in this particular case, is that `sleep` wastes 10 seconds of running test time, which in principle should be unnecessary. In the lack of a better alternative (I.e. a way to wait for this process to complete), it is IMHO still the most economic way of solving the problem at hand.

Here I see no efficient way of ensuring that the background process has completed. `script_retry` **in this case** does not make the situation anyhow better, as it would still wait between failed attempts. You have the exact same situation as with sleep: You wait for 10 seconds, which is in principle not necessary.

In short: `script_retry` **in this case** is just a `sleep` in disguise. I don't see how this would improve the situation at all.

tests/containers/kubectl.pm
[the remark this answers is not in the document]
grisu48: Probably yes. When (and if) we add other kubernetes distributions there is a chance that we have to change a couple of things. That's why we have the `KUBECTL_CLUSTER` variable to allow for such modifications without too much of a fuzz.

Also, if it becomes an issue later, we can still remove this line later.

lib/containers/k8s.pm
```diff
@@ -21,27 +21,43 @@ use registration qw(add_suseconnect_product get_addon_fullname);
...
     assert_script_run("/usr/local/bin/k3s-uninstall.sh");
 }
 
 sub install_kubectl {
     if (is_sle) {
-        zypper_call("in kubernetes1.18-client");
-    }
-    else {
+        # kubectl is in the container module
+        add_suseconnect_product(get_addon_fullname('contm'));
```
ilausuch, line 52: Is it necessary to add the module now? this worked until now

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,139 @@
...
+    my ($self) = @_;
+
+    # Try to collect as much information about kubernetes as possible
+    script_run('kubectl describe deployments');
+    script_run('kubectl describe services');
+    script_run('kubectl describe pods');
+    # Cleanup
+    script_run('kubectl delete -f service.yml');
+    script_run('kubectl delete -f deployment.yml');
+}
```
ilausuch, line 137: I guess we are not deleting k3s because is not necessary. The VM will be destroyed.

lib/containers/k8s.pm
grisu48: If you looks closely, this was also done before.

grisu48: Wait, my bad, this came from somewhere else. Let me check ....

grisu48: Sorry, I don't know where this came from, it should be there. Removing it now.

lib/containers/k8s.pm
```diff
@@ -21,27 +21,40 @@ use registration qw(add_suseconnect_product get_addon_fullname);
...
 }
 
 sub install_kubectl {
     if (is_sle) {
-        zypper_call("in kubernetes1.18-client");
-    }
-    else {
+        # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
+        if (is_sle(">15-SP1")) {
+            zypper_call("in kubernetes1.18-client");
```
jlausuch, line 53: would it make sense to add here `add_suseconnect_product("sle-module-containers")` ?

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,139 @@
...
+    validate_script_output("kubectl config current-context", sub { $_ =~ m/default/ });
+    assert_script_run("kubectl config use-context default");
+    assert_script_run("kubectl config set-context --current --namespace=default");
+    assert_script_run("kubectl config set-context localhost --user=root --namespace=default");
+    assert_script_run("kubectl config use-context localhost");
+    assert_script_run("kubectl config use-context default");
+    assert_script_run("kubectl config unset contexts.localhost");
+
+    ## Test jobs
+    record_info('Test 3', 'job and pod test runs');
```
jlausuch, line 80: You name this "test 3" but the other tests don't have a number. Either use numbers everywhere or just use here "Test jobs" to be more consistent.

tests/containers/kubectl.pm
jlausuch: We need to implement new argument `sleep_first => 1`  in `script_retry`  :D :D :D

jlausuch: Btw, you use `script_retry` here https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15215/files#diff-301489d174f96744185348e9a3aec18c327ec62bb49fded2463b5a662b731ffcR36

lib/containers/k8s.pm
```diff
@@ -21,27 +21,40 @@ use registration qw(add_suseconnect_product get_addon_fullname);
 our @EXPORT = qw(install_k3s uninstall_k3s install_kubectl install_helm install_oc);
 
 sub install_k3s {
-    my $k3s_version = get_var("CONTAINERS_K3S_VERSION", "v1.23.6+k3s1");
-    record_info('k3s', $k3s_version);
-    assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=$k3s_version sh -");
-    enter_cmd("k3s server &");
+  # Apply additional options. For more information see https://rancher.com/docs/k3s/latest/en/installation/install-options/#options-for-installation-with-script
+    my $k3s_version = get_var("CONTAINERS_K3S_VERSION");
```
jlausuch, line 25: No more hardcoded default version?

lib/containers/k8s.pm
jlausuch: It doesn't harm to have it. Maybe it was working before because the module was enabled by the parent job .. but having it here should be more clear what we are testing.

tests/containers/kubectl.pm
```diff
@@ -0,0 +1,139 @@
...
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    install_kubectl();
+    # Record kubectl version and check if the tool itself is healthy
+    record_info("kubectl", script_output("kubectl version --client --output=json"));
+
+    # Prepare the webserver testdata
+    assert_script_run('mkdir -p /srv/www/kubectl');
+    assert_script_run("echo 'I am groot' > /srv/www/kubectl/index.html");
```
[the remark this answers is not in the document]
grisu48: Good catch! :wink:

lib/main_containers.pm
```diff
@@ -186,7 +186,11 @@ sub load_container_tests {
     foreach (split(',\s*', $runtime)) {
         my $run_args = OpenQA::Test::RunArgs->new();
         $run_args->{runtime} = $_;
-        if (is_container_image_test()) {
+
+        if ($runtime eq 'kubectl') {
```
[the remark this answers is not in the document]
grisu48: Are you sure this makes sense? I moved it to the beginning, because I though that choosing the runtime tests have precedence over images tests on a runtime.

lib/containers/k8s.pm
grisu48: Handled in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15215#discussion_r921146563

lib/containers/k8s.pm
grisu48: See the block before. Unless an explicit version is stated via `CONTAINERS_K3S_VERSION`, there is no hardcoded version used. We can change this at any time simply by setting the variable accordingly.

grisu48:
The pipeline is complaining:

```
tests/containers/kubectl.pm: Missing strict or warnings at line 1, column 1.  The strict and warnings pragmas are important to avoid common pitfalls and deprecated/experimental functionality. Make sure each script or module contains "use strict; use warnings;" or a module that does this for you.  (Severity: 4)
tests/containers/kubectl.pm: Missing strict or warnings at line 1, column 1.  The strict and warnings pragmas are important to avoid common pitfalls and deprecated/experimental functionality. Make sure each script or module contains "use strict; use warnings;" or a module that does this for you.  (Severity: 4)
```

Both are present though, it's a bit picky today :wink:

lib/main_containers.pm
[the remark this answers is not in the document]
jlausuch: yes, it's a container host, and therefore it should be in the same place as the container hosts :)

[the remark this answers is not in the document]
jlausuch: The preference is set by these conditions and according variables. That's why we have that block for host tests, so let's try to put everything together.

tests/containers/kubectl.pm
jlausuch: Yes, it was done like this to have a proper "wait for things to be ready with timeout". I actually like it, it's a loop with timeout in single line. In this case, are you sure that 10 seconds is enough in ALL the corner cases that can happen in openQA?

lib/main_containers.pm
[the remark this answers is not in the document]
grisu48: Ahhh, now I see what you mean. Yes that makes sense. Done.

tests/containers/kubectl.pm
grisu48: > In this case, are you sure that 10 seconds is enough in ALL the corner cases that can happen in openQA?

IMHO it is reasonable to assume that 10 seconds is enough to set a simply firewall NAT rule. And you know very well the answer to all questions that contain "All corner cases" and "openQA" in the same sentence :wink: 

I'll change it as it really doesn't matter which way we do it. There is no functional difference in a healthy test run and weird ones would otherwise probably die.

lib/main_containers.pm
```diff
@@ -186,6 +186,8 @@ sub load_container_tests {
     foreach (split(',\s*', $runtime)) {
         my $run_args = OpenQA::Test::RunArgs->new();
         $run_args->{runtime} = $_;
+
+
```
jlausuch, line 190: intentional?

lib/main_containers.pm
```diff
@@ -200,13 +202,13 @@ sub load_container_tests {
...
             loadtest 'containers/bci_repo';
-        }
-        else {
+        } else {
             # Container Host tests
             load_host_tests_podman($run_args) if (/podman/i);
             load_host_tests_docker($run_args) if (/docker/i);
             load_host_tests_containerd_crictl() if (/containerd_crictl/i);
             load_host_tests_containerd_nerdctl() if (/containerd_nerdctl/i);
+            loadtest('containers/kubectl') if ($runtime eq 'kubectl');
```
jlausuch, line 211: This will work cause we will enable a single job for kubectl. But in this loop, we breakdown `$runtime` into its elements separated by coma. So, in case we decide to run kubectl and helm (for instance) tests together, this test won't be scheduled. So, I would propose:

`loadtest('containers/kubectl') if if (/kubectl/i);`

variables.md
```diff
@@ -37,7 +37,8 @@ CHECK_RELEASENOTES | boolean | false | Loads `installation/releasenotes` test mo
 CHECKSUM_* | string | | SHA256 checksum of the * medium. E.g. CHECKSUM_ISO_1 for ISO_1.
 CHECKSUM_FAILED | string | | Variable is set if checksum of installation medium fails to visualize error in the test module and not just put this information in the autoinst log file.
 CLUSTER_TYPES | string | false | Set the type of cluster that have to be analyzed (example: "drbd hana"). This variable belongs to PUBLIC_CLOUD_.
-CONTAINER_RUNTIME | string | | Container runtime to be used, e.g.  `docker`, `podman`, or both `podman,docker`.
+CONTAINER_RUNTIME | string | | Container runtime to be used, e.g.  `docker`, `podman`, or both `podman,docker`. In addition `kubectl` is used for kubernetes client test runs.
```
jlausuch, line 40:
```suggestion
CONTAINER_RUNTIME | string | | Container runtime to be used, e.g.  `docker`, `podman`, or both `podman,docker`. In addition, it is also used for other container tests, like  `kubectl`, `helm`, etc.
```

lib/main_containers.pm
grisu48: nope.

jlausuch (approved):
LGTM! Good job!

PR #15220, opened 2022-07-14
Public Cloud: Add ability to redownload maintenance repositories

pdostal:
Currently, when we clone a test and modify the list of issues it gets ignored as repositories were downloaded by parent.

- Verification run: [PUBLIC_CLOUD_REDOWNLOAD_MU=](https://pdostal-server.suse.cz/tests/175), [PUBLIC_CLOUD_REDOWNLOAD_MU=1](https://pdostal-server.suse.cz/tests/174)

### DIFF
--- a/tests/publiccloud/download_repos.pm
+++ b/tests/publiccloud/download_repos.pm
@@ -33,6 +33,10 @@ sub run {
         record_info('Skip download', 'Skipping maintenance update download (triggered by setting)');
         return;
     } else {
+        # Remove the ~/repos so they can be redownloaded
+        if (get_var('PUBLIC_CLOUD_REDOWNLOAD_MU')) {
+            script_run("rm -rf ~/repos");
+        }
         # Skip if we already downloaded the repos
         if (get_repo_status() == 1) {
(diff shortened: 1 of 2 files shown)


tests/publiccloud/download_repos.pm
```diff
@@ -33,6 +33,10 @@ sub run {
         record_info('Skip download', 'Skipping maintenance update download (triggered by setting)');
         return;
     } else {
+        # Remove the ~/repos so they can be redownloaded
+        if (get_var('PUBLIC_CLOUD_REDOWNLOAD_MU')) {
+            script_run("rm -rf ~/repos");
```
jlausuch, line 38: why not assert?

pdostal: I was not 100% but only 99% sure that the `rm -rf` exits `0` if the target does not exist.

jlausuch: `-f` forces `0` result if the dir/file doesn't exist:
```
cat nothing
cat: nothing: No such file or directory

rm -f nothing
echo $?
0
```

PR #15221, opened 2022-07-14
WIP: Upload azure images to region if specified

grisu48:
Adds the ability to create images in a custom region defined by
`PUBLIC_CLOUD_REGION`.

This commit allows to create disk images in regions that vary from the
location of the storage account. By default images are created in the
same region as the storage account. This is necessary, as we need 
to schedule some test runs in other regions than the storage account,
because certain instance types are not available there.

- Related ticket: https://progress.opensuse.org/issues/112400
- Verification run: https://duck-norris.qam.suse.de/tests/10034


grisu48:
This will not work, as the container and blobs are bound to the storage account's region.

PR #15228, opened 2022-07-15
Use our own testsite in firefox html5 test

asdil12:
https://progress.opensuse.org/issues/110665


asdil12:
https://openqa.suse.de/tests/9132731

tests/x11/firefox/firefox_html5.pm
```diff
@@ -22,24 +21,8 @@ sub run {
...
-        }
-        elsif (match_has_tag('firefox-youtube-signin')) {
-            assert_and_click('firefox-youtube-signin');
-            wait_still_screen(2);
-            next;
-        }
-        last;
-    }
-    send_key_until_needlematch('firefox-testvideo', 'spc', 30, 5);
+    $self->firefox_open_url('os-autoinst.github.io/os-autoinst-distri-opensuse/testwebsites/html5_video');
```
mloviska, line 24: Is this URL valid ?

asdil12: It should be valid after this PR got merged and the GitHub action ran through.

asdil12:
Note that before merging I need to fix the order of the test nodules to something that makes more sense.

tests/x11/firefox/firefox_html5.pm
foursixnine: @asdil12 can you add a follow up ticket  to have that running in an opensuse.org resource?

asdil12: the python webserver idea is not so bad - I will give it a try

PR #15229, opened 2022-07-15
lib/main_common: Run autoyast/wicked only on SLE+Leap

Vogtinator:
Tumbleweed uses NetworkManager by default, where this module is useless.

- Verification run: https://openqa.opensuse.org/tests/2470568#live


DimStar77 (approved):
LGTM

PR #1000, opened 2016-02-01
Add reset_consoles because of multiple reboot in test

dzedro:
tested: http://10.100.98.90/tests/283

PR #15231, opened 2022-07-15
ping: Test as non-root

pevik:
ping as non-root to detect missing capabilities on systems where RAW socket is used.

Use sudo for testing because virtio console supports proper non-root user login.

Add sanity check for localhost and ::1 to verify any ping command is working.

Also remove year in copyright (new policy).

Verification run:
- http://quasar.suse.cz/tests/314
- http://quasar.suse.cz/tests/307
- http://quasar.suse.cz/tests/317

### DIFF
--- a/tests/console/ping.pm
+++ b/tests/console/ping.pm
@@ -1,11 +1,15 @@
 # SUSE's ping tests in openQA
 #
-# Copyright 2016-2021 SUSE LLC
+# Copyright SUSE LLC
 # SPDX-License-Identifier: FSFAP
 #
-# Summary: Very basic ping test.
-# Tests pinging site-local IPv6, which had problems on ICMP datagram socket,
-# there were also problems with sysctl setup (bsc#1200617).
+# Summary: Very basic ping tests.
+
+# Tests pinging as user:
+# * localhost (sanity checks for both RAW sockets used on older SLES/openSUSE
+#   and newer ICMP datagram socket)
+# * site-local IPv6, which had problems on ICMP datagram socket,
+#   there were also problems with sysctl setup (bsc#1200617).
 #
 # Maintainer: Petr Vorel <pvorel@suse.cz>
 
@@ -18,8 +22,7 @@ sub run {
     my ($self) = @_;
 
     $self->select_serial_terminal;
-
-    zypper_call('in iputils libcap-progs');
+    zypper_call('in iputils libcap-progs sudo');
 
     record_info('KERNEL VERSION', script_output('uname -a'));
     record_info('net.ipv4.ping_group_range', script_output('sysctl net.ipv4.ping_group_range'));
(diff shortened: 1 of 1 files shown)


dzedro (approved):

dzedro:
How is the `pkill -u rootcat` related to this PR ?

tests/console/ping.pm
```diff
@@ -35,13 +34,23 @@ sub run {
...
+    my $cmd = "$sudo ping -c2 ::1";
+    record_info('ping6 ::1');
+    my $rc = script_run($cmd);
+    $self->result("fail") if ($rc);
+
+    $cmd = "$sudo ping6 -c2 $addr%$ifname";
     record_info('ping %');
-    assert_script_run($cmd);
+    $rc = script_run($cmd);
+    $self->result("fail") if ($rc);
```
czerw, line 49: why not `assert_script_run` to handle return value?

pevik:
> How is mentioned `pkill -u rootcat` related to this PR ?

Workaround to get non-root user, suggested by @b10n1k in 03376dcc8e191c51b7a13b8927321a001e75c3de.

czerw:
> > How is mentioned `pkill -u rootcat` related to this PR ?
> 
> Workaround to get non-root user, suggested by @b10n1k in [03376dc](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/03376dcc8e191c51b7a13b8927321a001e75c3de).

`enter_cmd` should have been used there instead of `type_string` to avoid mis-typing like rootcat is.

tests/console/ping.pm
pevik: +1, used for all but the last commit which needs evaluation for soft failure.

pevik:
> > > How is mentioned `pkill -u rootcat` related to this PR ?
> > 
> > 
> > Workaround to get non-root user, suggested by @b10n1k in [03376dc](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/03376dcc8e191c51b7a13b8927321a001e75c3de).
> 
> `enter_cmd` should have been used there instead of `type_string` to avoid mis-typing like rootcat is.

+1, but I prefer using `sudo`, because this will work also for svirt backend (s390x) which is IMHO also broken. Once I fix non-root user for both virtio and svirt I' use non-root login via `select_console()`.

czerw:
> +1, but I prefer using `sudo`, because this will work also for svirt backend (s390x) which is IMHO also broken. Once I fix non-root user for both virtio and svirt I' use non-root login via `select_console()`.

`sudo` solution is perfectly fine...

pevik:
> `sudo` solution is perfectly fine...

Maybe that is the reason, why fixing virtio console for non-root was not a priority for anybody :).

czerw:
Please remove reference to HPC in commit messasage, it is just confusing. I would put there  that we need sudo to use as non-root user to be able properly handle return failures for soft-failures.

czerw (approved):
LGTM for code, just please update commit msg

tests/console/ping.pm
```diff
@@ -35,13 +34,23 @@ sub run {
...
+    # test as non-root user
+    my $sudo = 'sudo -u \#1000';
+    record_info('id non-root', script_output("$sudo id", proceed_on_failure => 1));
+
+    my $cmd = "$sudo ping -c2 ::1";
+    record_info('ping6 ::1');
+    my $rc = script_run($cmd);
+    $self->result("fail") if ($rc);
+
+    $cmd = "$sudo ping6 -c2 $addr%$ifname";
```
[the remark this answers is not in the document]
pevik: Thx for a pointer, but I have the opposite (running as root command as normal user: `sudo -u \#1000`). I could backport this to `script_sudo()`, but I'll probably avoid more investment into openQA :).
Also, is `script_sudo()` intended for virtio/svirt console? Or is it for VNC?

pevik:
> Please remove reference to HPC in commit message, it is just confusing. I would put there that we need sudo to use as non-root user to be able properly handle return failures for soft-failures.

No, we need `sudo` because proper testing requires to run as non-root user. Handle return failures for soft-failures is different thing. I was also verbose in the commit message to be able to find other uses of this (i.e. HPC) once I get to fixing this issue, thus I'd prefer to keep it. IMHO it's quite clear: 
> ping as *non-root* to detect missing capabilities on systems where RAW socket is used.
> ....
> Use sudo for testing until virtio console supports proper non-root user login ...

czerw:
> No, we need `sudo` because proper testing requires to run as non-root user. Handle return failures for soft-failures is different thing. I was also verbose in the commit message to be able to find other uses of this (i.e. HPC) once I get to fixing this issue, 

You don't have to refer to HPC commit to defend usage of sudo. I'm just saying that referring that commit is confusing and we don't have to point that others do (and in this case even in sub-optimal way).

pevik:
OK, HPC removed from commit message and description, going to merge it. Thx for a review!

PR #15234, opened 2022-07-18
Check whether testfile is a binary file

paolostivanin:
As per bsc#1200893, using `file` to check if a file is encrypted is neither correct nor reliable. This commit replaces `file` with a perl check.

VR: https://openqa.opensuse.org/tests/2471554

### DIFF
--- a/tests/security/ecryptfs/ecryptfs.pm
+++ b/tests/security/ecryptfs/ecryptfs.pm
@@ -45,8 +45,9 @@ sub ecryptfs_mount {
     validate_script_output('mount | grep -m 1 ecryptfs', sub { m/\/root\/private/ });
     # The testfile should be readable and writable
     assert_script_run('cd private && touch testfile && echo Hello > testfile && grep Hello testfile && cd ..');
-    # The testfile should be an encrypted file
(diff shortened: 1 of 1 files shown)


rfan1:
@paolostivanin 
Can you please do VR on SLE as well? http://openqa.suse.de/tests/8945617
Only one concern is that 'perl' is install by default :)

punkioudi:
> @paolostivanin Can you please do VR on SLE as well? http://openqa.suse.de/tests/8945617 Only one concern is that 'perl' is install by default :)

Hm, as far as I can see this testsuite runs only in product testing, would it make sense to add it also in maintenance runs?

paolostivanin:
@rfan1  VR for SLE 15-SP3: https://openqa.suse.de/tests/9158353

paolostivanin:
> Hm, as far as I can see this testsuite runs only in product testing, would it make sense to add it also in maintenance runs?

many tests are still missing from maintenance. IMHO, I think we should open a dedicated poo ticket and decided which test suites to add to each codestream and what kind of prio they have.

jouyingbin (commented):
Thanks Paolo also mentioned that green also on 15sp4 https://openqa.suse.de/tests/9158972#

PR #15238, opened 2022-07-18
Use new sssd log option

asdil12:
Ticket: https://progress.opensuse.org/issues/104674


asdil12:
https://openqa.opensuse.org/t2472269

rfan1 (approved):

PR #15244, opened 2022-07-19
rt: Add cleanup and shutdow to prepare migration

czerw:
We need to publish hdd after migration. Image have to be cleaned and
system have to be shutdown.

- Related ticket: none
- Verification run: http://10.100.12.105/tests/1704


b10n1k (approved):
+1

PR #15245, opened 2022-07-19
Introduce new secrets scan for pr and commits

michaelgrifalconi:
Goal is to try out trufflehog to scan our pull requests (or direct commits) for unwanted secrets and learn from it. 
It should not be a mandatory check for now, to not disrupt current work.

More info in the poo# linked.

- Related ticket: https://progress.opensuse.org/issues/112190
- Needles: none
- Verification run: none


foursixnine:
@michaelgrifalconi how does it look like when somebody creates a PR or commits code that has a secret?

.github/workflows/trufflehog-commit.yml
```diff
@@ -0,0 +1,18 @@
+name: Leaked Secrets Scan on Commits
+on:
+  push:
+    branches: [ "main" ]
```
alvarocarvajald, line 4: Is this a literal branch name? IIRC os-autoinst-distri-opensuse does not have a main branch, but master.

baierjan:
Is there a reason to not have that in one workflow with
```
on:
  push:
    branches: [ "main" ]
  pull_request:
```
or did I miss another difference in the job definitions?

michaelgrifalconi:
> Is there a reason to not have that in one workflow with
> 
> ```
> on:
>   push:
>     branches: [ "main" ]
>   pull_request:
> ```
> 
> or did I miss another difference in the job definitions?

I see no reasons to have 2 jobs other than keeping the scope PR/commit separate. I can change it to a single one since they would be exactly the same

Vogtinator:
My main concern is that GitHub actions have full access to the repo, so this effectively grants everyone with write access to the `trufflehog` repo also write access to this one. Please at least pin the action to the specific SHA1 hash of the commit.

asmorodskyi (approved):
LGTM

foursixnine:
> 

We will need to add likely both detectors (at least the SCC one)

michaelgrifalconi:
@Vogtinator had a really good point, thanks! I was not aware of this risk, putting WIP to make sure it's not getting merged until we address the issue. 
The general discussion can still continue of course

foursixnine:
> Does this thing also detect openQA api tokens as well as SCC credentials? And is there some way to tell it to ignore certain entries? It would be good to have that documented somewhere.

I think the next two tasks of this epic would be to write the detector for openQA and for SCC

michaelgrifalconi:
We pass no tokens as argument and the action SHA1 is hardcoded now so we should not have surprises.

PR #15246, opened 2022-07-19
Rename ltp package to follow upstream changes

asmorodskyi:


jlausuch (approved):

pdostal (approved):

grisu48 (approved):

tests/publiccloud/run_ltp.pm
```diff
@@ -86,6 +86,7 @@ sub run {
     my $ltp_repo = get_var('LTP_REPO', 'https://download.opensuse.org/repositories/benchmark:/ltp:/stable/' . generate_version("_") . '/');
     my $provider;
     my $instance;
+    my $ltp_package = 'ltp-stable';
```
czerw, line 89: We use `LTP_PKG` variable  https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1a35043d7dac67f2f376d78ac4da1c78173196f3/tests/kernel/install_ltp.pm#L276  or default.

b10n1k, line 89: just a small thing. I dont see any good reason to define the package here. Would it make sense to have it public for this module? in such case the `gen_ltp_env` wouldnt need to change. wdyt?

asmorodskyi: this tests are running for 2-3 years and during this period there was never the case that I need to use different package here. On other side we have sooo many variables so each another variable especially when there is 0 chances  to actually use this variable

b10n1k (approved):
thanks @asmorodskyi.

PR #15250, opened 2022-07-19
wicked: Check only logs created during test

cfconrad:
We should not check older logs. Otherwise we would get
a log error in all subsequence tests showing the same issue, but they
actually occured only in the first run/boot.

- Verification run: https://openqa.suse.de/t9173655

### DIFF
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -858,17 +858,22 @@ sub skip_check_logs_on_post_run {
     shift->{skip_check_logs_on_post_run} = 1;
 }
 
+sub get_log_cursor {
+    my $cursor = script_output(q(journalctl -o export -n 1 | tr -dc '\n|[[:print:]]' |  grep __CURSOR));
+    ($cursor) = ($cursor =~ /^__CURSOR=(.*)$/m);
+    return $cursor;
+}
+
 sub check_logs {
     my $self = shift;
     my $code = shift;
-    my $cursor = '';
+    my $cursor = $self->{pre_run_log_cursor} // '';
 
     if (ref($code) eq 'CODE') {
-        $cursor = script_output(q(journalctl -o export -n 1 | tr -dc '\n|[[:print:]]' |  grep __CURSOR));
-        ($cursor) = ($cursor =~ /^__CURSOR=(.*)$/m);
-        $cursor = "-c '$cursor'";
+        $cursor = $self->get_log_cursor();
         $code->();
     }
+    $cursor = "-c '$cursor'" if (length($cursor) > 0);
     my @units = qw(wickedd-nanny wickedd-dhcp4 wickedd-dhcp6 wicked wickedd);
     my $default_exclude = 'wickedd=process \d+ has not exited yet; now doing a blocking waitpid';
     $default_exclude .= ',wickedd-dhcp6=Link-local IPv6 address is marked duplicate:';
@@ -995,6 +1000,7 @@ sub pre_run_hook {
         set_var('WICKED_TCPDUMP_PID', script_output('echo $CHECK_TCPDUMP_PID'));
     }
     $self->upload_wicked_logs('pre');
+    $self->{pre_run_log_cursor} = $self->get_log_cursor() if ($self->{name} ne 'before_test');
     $self->SUPER::pre_run_hook;
     $self->do_barrier('pre_run');
 }


lib/wickedbase.pm
```diff
@@ -995,6 +1000,7 @@ sub pre_run_hook {
         set_var('WICKED_TCPDUMP_PID', script_output('echo $CHECK_TCPDUMP_PID'));
     }
     $self->upload_wicked_logs('pre');
+    $self->{pre_run_check_logs_cursor} = $self->check_logs_cursor() if ($self->{name} ne 'before_test');
```
jlausuch, line 1003: Instead of condition by looking in which module it is, wouldn't it be better to check if logs exists ?

lib/wickedbase.pm
```diff
@@ -858,17 +858,22 @@ sub skip_check_logs_on_post_run {
     shift->{skip_check_logs_on_post_run} = 1;
 }
 
+sub check_logs_cursor {
```
asmorodskyi, line 861:
```suggestion
sub get_cursor_from_logs {
```

lib/wickedbase.pm
```diff
@@ -858,17 +858,22 @@ sub skip_check_logs_on_post_run {
...
     my $self = shift;
     my $code = shift;
-    my $cursor = '';
+    my $cursor = $self->{pre_run_check_logs_cursor} // '';
 
     if (ref($code) eq 'CODE') {
-        $cursor = script_output(q(journalctl -o export -n 1 | tr -dc '\n|[[:print:]]' |  grep __CURSOR));
-        ($cursor) = ($cursor =~ /^__CURSOR=(.*)$/m);
-        $cursor = "-c '$cursor'";
+        $cursor = $self->check_logs_cursor();
```
asmorodskyi, line 873: can you elaborate why we need to redefine cursor second time in this case ?

asmorodskyi: and why we can not just do here : 
```suggestion
        $self->{pre_run_check_logs_cursor} = $self->check_logs_cursor();
```

this would made all code a bit clearer

lib/wickedbase.pm
```diff
@@ -858,17 +858,22 @@ sub skip_check_logs_on_post_run {
...
         $code->();
     }
+    $cursor = "-c '$cursor'" if (length($cursor) > 0);
```
asmorodskyi, line 876: I would drop it from here and just add `$cmd .= "-c '" . $self->{pre_run_check_logs_cursor} . "'"` in place where it is actually used  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15250/files#diff-e5a344b1982337ed8b1fcce6c2ec13224675b3db7733fcce7acb66c82a8e8d71R895

lib/wickedbase.pm
cfconrad: ack, I rename it to `get_log_cursor()`

lib/wickedbase.pm
cfconrad: The thing is, that we do not wanna limit the logs check on the first boot. Because we would like to see also the logs during boot.
This is actually questionable but I think it makes sense, WDYT?

lib/wickedbase.pm
cfconrad: It's for calls like this: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/wickedbase.pm#L838
which is used here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/wicked/startandstop/sut/t17_ifrename.pm#L24 and what we wanna get a log error specific for each `run_test_shell_script()` call, so we can actually map on which call the error occurred.

cfconrad: Maybe I misunderstood you. It would work to set the `$self->{pre_run_log_cursor}` variable direclty. But then we would have two places where it is set, which makes the code harder to read and we would loose the functionality where we wanna check logs specific on  a call **and** at the end for the full runtime of the test-module.

cfconrad: If we would do it, then I would actually also set the `skip_check_logs_on_post_run()` if `check_logs()` is called with `CODE`. So we actually forbid the checking of full test-module logs once someone call `run_test_shell_script()`.

asmorodskyi: ehm maybe I am wrong but `__CURSOR=` is set only once , no ? `get_log_cursor` is just getting it from journal so no matter *when * you will call `get_log_cursor` you will always get same place

lib/wickedbase.pm
asmorodskyi: I briefly checked `run_test_shell_script` and I failed to find where set of `__CURSOR` happens there so I don't see relation between `run_test_shell_script` and setting of `__CURSOR`

lib/wickedbase.pm
cfconrad: no, its like a timestamp and if you pass this to a subsequent `journalctl` call you will get all message from that "timestamp" to `now`.

asmorodskyi (approved):
LGTM

PR #15251, opened 2022-07-19
Create autoyast profiles and schedulers for HPC

b10n1k:
ay profiles are generated with the clone.pm module and were adjusted in order
to remove unnecessary elements and modified to run in each sle15 version.
I assume that the sle12 might be a bit different.
 
After created and tested 4 different xmls i merged the ones of the same arch.
So each arch should work for either the textmode or gnome_textmode. I used the
SSC_ADDONS variable which in the `*_gnome_textmode` seems to be a unique
identifier. But not sure if that the optimal approach. Instead we could use
another job variable, with best candidate `DESKTOP`. But this is not defined in
the test and for now i just didnt touch anything in the settings to make it work.

- Related ticket: https://progress.opensuse.org/issues/113593
- Verification run: 

https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=b10n1k%2Fos-autoinst-distri-opensuse%23hpc_autoyast
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=b10n1k%2Fos-autoinst-distri-opensuse%2315251

### DIFF
--- a/data/autoyast_sle15/hpc/create_hdd_textmode_aarch64.xml.ep
+++ b/data/autoyast_sle15/hpc/create_hdd_textmode_aarch64.xml.ep
@@ -1,0 +1,507 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <suse_register t="map">
+    <addons t="list">
+      <addon t="map">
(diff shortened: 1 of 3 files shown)


czerw:
Does it add maintenance repos?

czerw:
I cant find latest verification run,  i can find only  https://openqa.suse.de/tests/overview?version=15-SP4&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2315251 which fails on aarch64.

czerw (approved):

PR #15252, opened 2022-07-19
Prevent kernel taint during livepatch installation

mdoucha:
Livepatch installation runs LTP tests in the background to create some system load while the livepatch is being applied. A few syscalls tests load custom kernel modules which taints the running kernel and causes the livepatch installation to fail. Skip those tests during livepatch setup.

- Related ticket: N/A
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/9173921


mdoucha:
Verification run passed, this PR is ready for review and merge.

PR #15253, opened 2022-07-20
Support installing TW with ipxe menu in O3

Julie-CAO:
In O3, only iPXE is supported to install Tumbleweed. Meanwhile, dnsmasq + http in O3 does not allow asign an seperate iPXE bootloader for an machine in its VLAN. As a consequence, the iPXE menu has to be kept working in general so maybe more people can make use of it in the future if needed.

A static iPXE menu is supported in this PR. For virtualization tests in O3, typing 't' to enter the TW entry to start install. Check screens a few times to ensure the install process is ongoing.

- Related ticket: https://progress.opensuse.org/issues/92380 & https://progress.opensuse.org/issues/113246
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/780
- Verification run: http://10.67.129.102/tests/303 (the iPXE menu is not shown due to ipmi sol console issues of my HP machine)

Welcome comments and suggestions. 

@alice-suse @guoxuguang @waynechen55 @nanzhg @tbaev @tonyyuan1 @lemon-suse @RoyCai7 @mdoucha @czerw

### DIFF
--- a/tests/installation/ipxe_install.pm
+++ b/tests/installation/ipxe_install.pm
@@ -14,7 +14,7 @@ use utils;
 use testapi;
 use bmwqemu;
 use ipmi_backend_utils;
-use version_utils 'is_upgrade';
+use version_utils qw(is_upgrade is_tumbleweed);
 use bootloader_setup 'prepare_disks';
 use Utils::Architectures;
 
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
     diag "$response->{status} $response->{reason}\n";
 }
 
+
 sub run {
     my $self = shift;
 
     poweroff_host;
 
-    #virtualization tests use a static ipxe configuration file
-    set_bootscript unless get_var('VIRT_AUTOTEST');
+    #virtualization tests use a static ipxe configuration file in O3
+    set_bootscript unless get_var('VIRT_AUTOTEST') && is_tumbleweed;
 
     set_pxe_boot;
 
     poweron_host;
 
+    select_console 'sol', await_console => 0;
+
+    #use ipxe bootloader in O3
+    #it is static menu and choose the TW entry to start installation
+    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
+        ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
+        assert_screen('o3-ipxe-menu', 300);
+        send_key 't';
+        assert_screen([qw(load-linux-kernel load-initrd)], 240);
+        #loading initrd spend much time(fg. 10-15 minutes to Beijing SUT)
+        assert_screen([qw(start-tw-install)], 360);
+        assert_screen([qw(network-config-created loading-installation-system)], 60);
+        wait_still_screen(stilltime => 30, similarity_level => 60, timeout => 120);
+        save_screenshot;
+    }
+
     # when we don't use autoyast, we need to also load the right test modules to perform the remote installation
     if (get_var('AUTOYAST')) {
-        select_console 'sol', await_console => 0;
         # make sure to wait for a while befor changing the boot device again, in order to not change it too early
         sleep 120;
     } else {
-        select_console 'sol', await_console => 0;
         my $ssh_vnc_wait_time = 1500;
+        #for TW virtualization test, 15 minutes is enough to load installation system, 75 minutes is too long
+        $ssh_vnc_wait_time = 300 if get_var('VIRT_AUTOTEST') && is_tumbleweed;
         my $ssh_vnc_tag = eval { check_var('VIDEOMODE', 'text') ? 'sshd' : 'vnc' } . '-server-started';
         my @tags = ($ssh_vnc_tag);
         if (check_screen(\@tags, $ssh_vnc_wait_time)) {
--- a/schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
+++ b/schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
@@ -1,4 +1,4 @@
-name:           virt-guest-installation
+name:           tumbleweed-virt-guest-install-tests
 description:    >
     Maintainer: jcao@suse.com
     Virtualization guest installation tests including extended feature tests schedule
@@ -16,7 +16,7 @@ conditional_schedule:
     install_and_bootup:
         DO_NOT_INSTALL_HOST:
             0:
-                - '{{bootup}}'
+                - installation/ipxe_install
                 - installation/welcome
                 - installation/online_repos
                 - installation/installation_mode
@@ -65,9 +65,3 @@ conditional_schedule:
         ENABLE_STORAGE:
             1:
                 - virtualization/universal/storage
-    bootup:
-        IPXE:
-            1:
-                - installation/ipxe_install
-            0:
-                - boot/boot_from_pxe


alice-suse:
LGTM, but needs to fix the checking issue.

czerw:
I'm on vacation :) @frankenmichl  please have a look

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
+    #use ipxe bootloader in O3
+    #it is static menu and choose the TW entry to start installation
+    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
+        ipmi_backend_utils::ipmitool 'chassis power reset' if !check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 300);
+        assert_screen('o3-ipxe-menu', 300) if match_has_tag('ipxe-boot-failure');
+        send_key 't';
+        assert_screen([qw(load-linux-kernel load-initrd)], 240);
+        #loading initrd spend much time(fg. 10-15 minutes to Beijing SUT)
+        assert_screen([qw(start-tw-install)], 360);
+        assert_screen([qw(network-config-created loading-installation-system)], 60);
```
waynechen55, line 158: Sequence of check_screen and assert_screen without any interaction. I doubt it can function very well.

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
     # when we don't use autoyast, we need to also load the right test modules to perform the remote installation
     if (get_var('AUTOYAST')) {
-        select_console 'sol', await_console => 0;
         # make sure to wait for a while befor changing the boot device again, in order to not change it too early
         sleep 120;
     } else {
-        select_console 'sol', await_console => 0;
         my $ssh_vnc_wait_time = 1500;
+        #for TW virtualization test, 15 minutes is enough to load installation system, 75 minutes is too long
+        $ssh_vnc_wait_time = 300 if get_var('VIRT_AUTOTEST') && is_tumbleweed;
```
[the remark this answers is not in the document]
guoxuguang: refer to the comment, there would be set 75x60 as $ssh_vnc_wait_time.

tests/installation/ipxe_install.pm
Julie-CAO: It takes longer to download kernel files from downloads.opensuse.org than SUSE internal network. These asserts are intend to make sure the process is ongoing, and detect failure ASAP.

tests/installation/ipxe_install.pm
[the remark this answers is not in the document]
Julie-CAO: The TIMEOUT_SCALE used to scale the timeout is always set to 3 by default. So in our tests, when we set 30s as timeout in openqa api, actually it spends 30X3=90 seconds. As a consequence, 300s here is 15 minutes actually.

Julie-CAO:
> I'm on vacation :) @frankenmichl please have a look

When will you or @frankenmichl come back from vacation?

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
 
     set_pxe_boot;
 
     poweron_host;
 
+    select_console 'sol', await_console => 0;
+
+    #use ipxe bootloader in O3
+    #it is static menu and choose the TW entry to start installation
+    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
```
waynechen55, line 151: Why need get_var('VIRT_AUTOTEST') here ? Does virt makes any difference when comes to ipxe install ? Is this because virt installs baremetal host via ipmi ?

Julie-CAO: the ipxe_install.pm is used in both OSD and O3 by kernel team, virt team and etc. The ipxe menu is only used in O3 by virt team. kernel team uses the ipxe install in a runtime way with ipmi backend, fg. post the install command based in different SUT.

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
+    set_bootscript unless get_var('VIRT_AUTOTEST') && is_tumbleweed;
 
     set_pxe_boot;
 
     poweron_host;
 
+    select_console 'sol', await_console => 0;
+
+    #use ipxe bootloader in O3
+    #it is static menu and choose the TW entry to start installation
```
waynechen55, line 150: iPXE module already supports "boot scrip customization", why virt test use static menu ?

waynechen55: I think the power of iPXE lies in "customized boot script" and "iPXE command line", but virt test looks backward and use static menu instead.

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
+    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
+        ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
+        assert_screen('o3-ipxe-menu', 300) if match_has_tag('ipxe-boot-failure');
```
waynechen55, line 153: These two lines:
```
ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
assert_screen('o3-ipxe-menu', 300) if match_has_tag('ipxe-boot-failure');
```
I think there is possibility that 'o3-ipxe-menu' will not be detected, for example, check_screen fails. So if this is case, the test just press 't' blindly ?

tests/installation/ipxe_install.pm
Julie-CAO: It is because the special infra restriction in O3. There is no independent dhcp server and htttp server. The static iPXE menu is a tradeoff.

tests/installation/ipxe_install.pm
Julie-CAO: 'o3-ipxe-menu' is expected although in a few cases that ipmi console is messy. As the cases are not normal, I should remove the if condition.

waynechen55: Then detecting 'ipxe-boot-failure' becomes meaningless ?

tests/installation/ipxe_install.pm
waynechen55: What about 
```
if (use static menu or similar)
```
? 
Because what is being done in the if is using static menu compared to customized bootscript.

tests/installation/ipxe_install.pm
```diff
@@ -131,26 +131,43 @@ END_BOOTSCRIPT
...
 
+    select_console 'sol', await_console => 0;
+
+    #use ipxe bootloader in O3
+    #it is static menu and choose the TW entry to start installation
+    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
+        ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
+        assert_screen('o3-ipxe-menu', 300);
+        send_key 't';
+        assert_screen([qw(load-linux-kernel load-initrd)], 240);
```
waynechen55, line 155: Asserting 'load-linux-kernel' and  'load-initrd' at the same time does not make too much sense, because these two can not appear on the screen at the same time. They appear in order, firstly 'kernel' then 'initrd'. So assert_screen will always  return on detection of 'kernel'. It nearly has no chance of finding 'initrd'. Suggest
```
assert_screen('load-initrd', 240);
```
or 
```
assert_screen('load-kernel', 240);
assert_screen('load-initrd', 240);
```
or
skip detecting these two needles, but you can record the screen by using check_screen ?

tests/installation/ipxe_install.pm
Julie-CAO: Static menu may vary in different places, thus the key or precedure to kick off a boot would be different. Also, introducing an extra setting or variable is not neccessary.

tests/installation/ipxe_install.pm
Julie-CAO: It was removed in my lasted commit.

tests/installation/ipxe_install.pm
Julie-CAO: In actural, becuase of ipmi console problem or something else, the 'load-kernel' or 'load-initrd' are not always occur at the same time, f.g. http://10.67.129.102/tests/306#step/ipxe_install/4, in this case only 'load-initrd' is present. there was some cases only 'load-kernel' occured. So the test continues with either of them.

waynechen55: I think
```
assert_screen('load-kernel-and-initrd', 240);
```
is better.
But if you think this will cause instability, then I am fine with current code.

Julie-CAO:
> > I'm on vacation :) @frankenmichl please have a look
> 
> Hi @czerw, When will you or @frankenmichl come back from vacation?

Hi @frankenmichl, could you help review? The test is waiting for being deployed in O3.

alice-suse:
> > > I'm on vacation :) @frankenmichl please have a look
> > 
> > 
> > Hi @czerw, When will you or @frankenmichl come back from vacation?
> 
> Hi @frankenmichl, could you help review? The test is waiting for being deployed in O3.

@czerw  Hello Petr, is it possible to find someone else in kernel team to review the PR? There seems no response for half a month.

frankenmichl (approved):

alice-suse:
Merge it to see our tumbleweed VT test live in O3.

PR #15254, opened 2022-07-20
Fix autoyast user configuration for sle rt

b10n1k:
This will fix the login after autoyast installatiion for sle RT. The login
fails because it takes the hashed string as the actual password.

- Verification run: http://aquarius.suse.cz/tests/11358
https://openqa.suse.de/t9177828

### DIFF
--- a/data/autoyast_sle15/autoyast_rt_textmode.xml
+++ b/data/autoyast_sle15/autoyast_rt_textmode.xml
@@ -128,7 +128,7 @@
   </networking>
   <users config:type="list">
     <user>
-      <encrypted config:type="boolean">false</encrypted>
+      <encrypted config:type="boolean">true</encrypted>
       <fullname>Bernhard M. Wiedemann</fullname>
       <gid>100</gid>
       <home>/home/bernhard</home>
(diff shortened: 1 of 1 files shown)


data/autoyast_sle15/autoyast_rt_textmode.xml
```diff
@@ -1,163 +1,163 @@
...
+      <home>/home/bernhard</home>
+      <password_settings>
+        <inact>-1</inact>
+        <max>99999</max>
+        <min>0</min>
+        <warn>7</warn>
+      </password_settings>
+      <shell>/bin/bash</shell>
+      <uid>1000</uid>
+      <user_password>$6$vYbbuJ9WMriFxGHY$gQ7shLw9ZBsRcPgo6/8KmfDvQ/lCqxW8/WnMoLCoWGdHO6Touush1nhegYfdBbXRpsQuy/FTZZeg7gQL50IbA</user_password>
```
czerw, line 143: maybe this one could be fixed too

b10n1k: i could test it but i think this is correct. the only problem was that `encrypted` was set false.

PR #15260, opened 2022-07-21
Update preparation of pr's (but really update github pages workflow)

foursixnine:
I mostly want to test that explicitly changing workflow permissions, won't break anything, while avoiding privilege escalation - just in case - 

Screenshot 2022-07-21 at 13 49 52

follow up to: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15245


foursixnine:
Self merging for now, let's  see what breaks

PR #15261, opened 2022-07-21
Double confirm smt-mirror can pass

rfan1:
Due to bsc#1201738, enhance current logic to double check
the log file to make sure smt-mirror passes without error

- Related ticket: https://progress.opensuse.org/issues/114403
- Verification run:
http://openqa.suse.de/t9207330
http://openqa.suse.de/t9207331
http://openqa.suse.de/t9207332

### DIFF
--- a/tests/smt/smt_server_install.pm
+++ b/tests/smt/smt_server_install.pm
@@ -160,10 +160,16 @@ sub run {
     validate_script_output "smt-repos -o", sub { m/SLES12-SP5-Updates/ };
     validate_script_output "smt-repos -o", sub { m/SLES12-SP5-Pool/ };
 
-    assert_script_run "smt-mirror", 16000;
+    assert_script_run "smt-mirror --logfile /var/log/smt/smt-mirror.log", 16000;
 
     assert_script_run "df -h";
     save_screenshot;
+
+    #We need double confirm the mirroring procedure succeeds without any error,
+    #please refer to bsc#1201738 for more detail information
+    assert_script_run "sync";
+    assert_script_run "tail -2 /var/log/smt/smt-mirror.log | grep 'Errors:                   : 0'";
+
     select_console "x11";
 }
 


foursixnine (dismissed):

foursixnine:
@rfan1 can you provide a VR where the error is properly detected?, I'm not sure about the regexp

rfan1:
> @rfan1 can you provide a VR where the error is properly detected?, I'm not sure about the regexp

http://openqa.suse.de/t9207330
http://openqa.suse.de/t9207331
http://openqa.suse.de/t9207332

foursixnine:
@rfan1 can you provide a failing case?

rfan1:
> @rfan1 can you provide a failing case?

http://openqa.suse.de/tests/9171525#step/system_prepare/5

But, I don't have any job can explain without "sync" cmd, the job failed one. however, I think it makes sense we issue "sync" command

rfan1:
> > @rfan1 can you provide a failing case?
> 
> http://openqa.suse.de/tests/9171525#step/system_prepare/5
> 
> But, I don't have any job can explain without "sync" cmd, the job failed one. however, I think it makes sense we issue "sync" command

Let me try to set the disk size to 40GB and trigger a failed job then, hold on the merge process

a-kpappas:
It's not really a regex, more of a string match. It may fail in the future but it will fail loudly and we can attempt a regex then. If @foursixnine doesn't still object I think we should merge.

a-kpappas (dismissed):

rfan1:
> > > @rfan1 can you provide a failing case?
> 
> Let me try to set the disk size to 40GB and trigger a faild job then, hold on the merge process

http://openqa.suse.de/tests/9208875#step/smt_server_install/57

BTW, I did some modification and executed the "df -h" command prior to log check, then we can check the disk full error easily

foursixnine:
@rfan1 Do we want to continue working on this?

PR #15264, opened 2022-07-21
Save logs directly on the worker for offline upload via ulogs

Zaoliang:
see poo# https://progress.opensuse.org/issues/109792
sub routine: save_ulog - save logs directly on the worker for offline upload via ulog
see verification test run for start_systemd_testkit_offline

### DIFF
--- a/lib/Utils/Logging.pm
+++ b/lib/Utils/Logging.pm
@@ -1,0 +1,41 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+=head1 Utils::Logging
+
+C<Utils::Logging> - Save logs directly on the worker for offline upload via ulogs 
+
+=cut
+
+package Utils::Logging;
+
+use base 'Exporter';
+use Exporter;
+use strict;
+use warnings;
+use testapi 'script_output';
+use Mojo::File qw(path);
+
+our @EXPORT = qw(save_ulog);
+
+=head2
+
+save_ulog($out $filename);
+
+Creates a file from a string, the file is then saved in the ulogs directory of the worker running isotovideo. 
+This is particularily useful when the SUT has no network connection.
+
+example: 
+
+$out = script_output('journalctl --no-pager -axb -o short-precise');
+$filename = "my-test.log";
+
+=cut
+
+sub save_ulog {
+    my ($out, $filename) = @_;
+    mkdir('ulogs') if (!-d 'ulogs');
+    path("ulogs/$filename")->spurt($out);    # save the logs to the ulogs directory on the worker directly
+}
+
+1;
--- a/tests/console/start_systemd_testkit_offline.pm
+++ b/tests/console/start_systemd_testkit_offline.pm
@@ -21,6 +21,7 @@ use Mojo::JSON qw(encode_json);
 use version_utils qw(is_sle);
 use strict;
 use warnings;
+use Utils::Logging;
 
 my $log = '/tmp/systemd_run.log';
 my $testdir = '/usr/lib/test/external/';
@@ -122,6 +123,8 @@ sub upload_systemdlib_tests_logs {
     my ($self) = @_;
     my $out = script_output('journalctl --no-pager -axb -o short-precise');
     record_info("JOURNAL", "$out");
+    my $filename = "start_external_testkit_offline.log";
+    save_ulog($out, $filename);
 }
 
 1;


foursixnine (commented):
I would adjust the PR and commit title to say something like:

"Save logs directly on the worker for offline upload via ulogs"

foursixnine:
@Zaoliang can you provide a VR?

foursixnine:
@Zaoliang one example of logs we would like to have uploaded this way, is this one: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15294

lib/Utils/Logging.pm
```diff
@@ -0,0 +1,24 @@
...
+# see https://progress.opensuse.org/issues/109792
+
+package Utils::Logging;
+
+use base 'Exporter';
+use Exporter;
+use strict;
+use warnings;
+use testapi 'script_output';
+use Mojo::File;
```
foursixnine, line 14:
```suggestion
use Mojo::File qw(path);
```

Zaoliang: yes, already found this was wrong, thx

Zaoliang:
VR:
finally works :)
http://openqa.suse.de/tests/9248311#step/shutdown/13

lib/Utils/Logging.pm
```diff
@@ -0,0 +1,25 @@
...
+use testapi 'script_output';
+use Mojo::File qw(path);
+
+our @EXPORT = qw(log2serial);
+
+sub log2serial {
+    my $data = script_output("journalctl -k | gzip | base64 -w0");
+    my $logname = "temporary-filename.log.gz";
+    my $path = path('/tmp/ulogs')->make_path;
+    path("$path/$logname")->spurt($data);    # save the logs to the ulogs directory on the worker directly
```
foursixnine, line 22: The error you were getting, in the previous runs, was because the ulogs directory was not created

```suggestion
    mkdir('ulogs') if (!-d 'ulogs');
    path("ulogs/$logname")->spurt($data);    # save the logs to the ulogs directory on the worker directly
```

foursixnine (commented):
It works, but the actual file, is never uploaded... See my comment, I think that should do the trick

lib/Utils/Logging.pm
Zaoliang: yes, this is working now, thanks!

see VR: http://openqa.suse.de/tests/9252081

I checked the logs and it looks good:
cat temporary-filename.log.gz | base64 -di | zcat

lib/Utils/Logging.pm
```diff
@@ -0,0 +1,25 @@
...
+use strict;
+use warnings;
+use testapi 'script_output';
+use Mojo::File qw(path);
+
+our @EXPORT = qw(log2serial);
+
+sub log2serial {
+    my $data = script_output("journalctl -k | gzip | base64 -w0");
+    my $logname = "temporary-filename.log.gz";
```
foursixnine, line 20: since the upload of the logs works out of the box, modify logs2serial, so that it recieves 2 parameters, one passing the content of the file, and a second one passing the filename...

and let's call it here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f34f7221e0bc2f44c2202ab1c851b6b7ae1936e9/tests/console/start_systemd_testkit_offline.pm#L123

so that these logs, are uploaded here: https://openqa.suse.de/tests/9247366#step/start_systemd_testkit_offline/60

lib/Utils/Logging.pm
```diff
@@ -0,0 +1,25 @@
...
+use base 'Exporter';
+use Exporter;
+use strict;
+use warnings;
+use testapi 'script_output';
+use Mojo::File qw(path);
+
+our @EXPORT = qw(log2serial);
+
+sub log2serial {
```
foursixnine, line 18:
```suggestion
our @EXPORT = qw(save_ulog);

sub save_ulog {
```

tests/console/start_systemd_testkit_offline.pm
```diff
@@ -118,10 +119,4 @@ sub parse_results_from_output {
     assert_script_run "echo '$json' > /tmp/$results_file";
 }
 
-sub upload_systemdlib_tests_logs {
-    my ($self) = @_;
-    my $out = script_output('journalctl --no-pager -axb -o short-precise');
-    record_info("JOURNAL", "$out");
-}
```
foursixnine, line 125: This is actually still needed, you will need to add: 

```
my $filename= "my-amazing-test";
log2serial($out, $filename);
```

lib/Utils/Logging.pm
Zaoliang: changed now, thanks

tests/console/start_systemd_testkit_offline.pm
Zaoliang: yes, updated now, thanks!

Zaoliang:
VR:
https://openqa.suse.de/tests/9261956#downloads

lib/Utils/Logging.pm
```diff
@@ -0,0 +1,37 @@
...
+use Mojo::File qw(path);
+
+our @EXPORT = qw(save_ulog);
+
+=head2
+
+save_ulog($out $filename);
+
+example: 
+$out = script_output('journalctl --no-pager -axb -o short-precise');
```
foursixnine, line 26:
```suggestion

Creates a file from a sting, that is then saved in the ulogs directory of the worker running isotovideo, this is particularily useful when the SUT has no network connection.

example: 
$out = script_output('journalctl --no-pager -axb -o short-precise');
```

foursixnine (approved):
One last niptick with the POD documentation, but other than that, looks great! :D

lib/Utils/Logging.pm
Zaoliang: sure, this sounds good, thanks!

PR #15268, opened 2022-07-22
Make sure that autoyast2 is installed before clone_system

b10n1k:
Call `zypper in` on /autoyast2/ package which provides the clone_system yast
feature. This makes the module autonomous and self-contained. In case the
package is already installed, this addition would not cause any destructions
and the flow will continue as expected. I added a timeout because i encounter
some issues on aarch64. no needed for x86_64.


ge0r (approved):
Makes sense, thanks!

PR #15270, opened 2022-07-22
Enable running docker and podman in the same job for BCI tests

jlausuch:
Each BCI run will produce some log files, e.g. junit_base.xml.
But if we run the same command twice (using different engine)
those files will be overwritten.
With this PR, after each BCI test run, those files will be renamed
with the container engine.
I have createda third module to parse all the logs generated by the
tests.

https://progress.opensuse.org/issues/113501

VRs:
[base image 15-SP4](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=27.11.4_sles15-image&groupid=398)
[lang image](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=7.19_golang-1.18-image&groupid=398)


tests/containers/bci_logs.pm
```diff
@@ -0,0 +1,62 @@
...
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    assert_script_run("cd /root/BCI-tests");
+    my $dom = XML::LibXML::Document->new('1.0', 'utf-8');
+    my $root = $dom->createElement('testsuites');
+    $dom->setDocumentElement($root);
+    my $result_files = script_output('ls | grep junit');
```
pdostal, line 28:
```suggestion
    my $result_files = script_output('find /root/BCI-tests -type f -name "*.junit"');
```
With this, you can also avoid `cd`.

tests/containers/bci_prepare.pm
```diff
@@ -98,15 +99,17 @@ sub run {
...
     record_info('Clone', "Clone BCI tests repository: $bci_tests_repo");
     my $branch = $bci_tests_branch ? "-b $bci_tests_branch" : '';
-    assert_script_run("git clone $branch -q --depth 1 $bci_tests_repo");
+    assert_script_run("git clone $branch -q --depth 1 $bci_tests_repo /root/BCI-tests");
 
     # Pull the image in advance
     if (my $image = get_var('CONTAINER_IMAGE_TO_TEST')) {
         record_info('IMAGE', $image);
+        # If $engines are multiple (e.g. CONTAINER_RUNTIME=podman,docker), we just pick one of them for this check
+        my $engine = ($engines =~ /podman/) ? 'podman' : 'docker';
```
ilausuch, line 112: Would be better?
```
if ($engine =~ /podman/) { ... }
else if($engine =~ /docker/) {...}
else No valid engine

ilausuch (approved):

pdostal (approved):
This is 💯 nice to see!

tests/containers/bci_prepare.pm
jlausuch: ok, why not

m-dati: if **$engines**="podman,docker", where will it span both values in order to repeat the test for each engine?

jlausuch: I have updated it. This module is executed only 1 time, I just pick one of the engines that are in the variable.

tests/containers/bci_logs.pm
jlausuch: I am renaming those files after bci_test.pm module, but the suggestion is still valid, thank you.

tests/containers/bci_prepare.pm
m-dati: But _docker_ set the pulled repo in a different place than _podman_, so we should repeat the $engine pull for each type

jlausuch: Right, but this code block is just to verify the build tag of the image and report also the size to some database. The BCI-tests will pull the image themselves when needed.

tests/containers/bci_logs.pm
jlausuch: Updated with both suggestions. `find /root/BCI-tests -type f -name "junit_*.xml"`

jlausuch:
Re-running VRs: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=27.11.4_sles15-image&groupid=398

asmorodskyi (approved):
LGTM

PR #15275, opened 2022-07-25
Make sure that autoyast2 is installed before clone_system

b10n1k:
Call zypper in on /autoyast2/ package which provides the clone_system yast
feature. This makes the module autonomous and self-contained. In case the
package is already installed, this addition would not cause any destructions
and the flow will continue as expected. I added a timeout because i encounter
some issues on aarch64. no needed for x86_64.

The missing package seems to affect only SLE.

- Verification run: http://aquarius.suse.cz/tests/11366
[latest](http://aquarius.suse.cz/tests/11413)

### DIFF
--- a/tests/autoyast/clone.pm
+++ b/tests/autoyast/clone.pm
@@ -13,10 +13,12 @@ use strict;
 use warnings;
 use parent 'y2_module_consoletest';
 use testapi;
+use utils qw(zypper_call);
 
 sub run {
     my $self = shift;
     assert_script_run 'rm -f /root/autoinst.xml';
+    zypper_call('in autoyast2', 300);
     my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'clone_system', yast2_opts => '--ncurses');
     if (check_screen 'autoyast2-install-accept', 10) {
         send_key 'alt-i';    # confirm package installation


ge0r (commented):
Shouldn't you also add the line `zypper_call('in autoyast2')` given that the previous commit that had this line got reverted?
Also from your verification run link, it does not look like this branch's changes have been applied to the run's [tests/autoyast/clone.pm](http://aquarius.suse.cz/tests/11366/modules/clone/steps/1/src.txt)

b10n1k:
> Shouldn't you also add the line `zypper_call('in autoyast2')` given that the previous commit that had this line got reverted?
> Also from your verification run link, it does not look like this branch's changes have been applied to the run's [tests/autoyast/clone.pm](http://aquarius.suse.cz/tests/11366/modules/clone/steps/1/src.txt)

ofc and i feel stupid. I though it was there somehow. 

> Shouldn't you also add the line `zypper_call('in autoyast2')` given that the previous commit that had this line got reverted? Also from your verification run link, it does not look like this branch's changes have been applied to the run's [tests/autoyast/clone.pm](http://aquarius.suse.cz/tests/11366/modules/clone/steps/1/src.txt)

ofc. i forgot to rebase from master so it took only this changed. fixed

ge0r:
Thanks! Could you also update the VR link just to be sure?

ge0r (approved):
Nice thanks!

PR #15277, opened 2022-07-25
FIx GUI shutdown issue on s390x backend zVM and zKVM

Zaoliang:
See https://progress.opensuse.org/issues/114439 for details.
With PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15176 we expect the failure
of GUI shutdown because record_soft_failure got removed and no fix was in place.

Verification run:
https://openqa.suse.de/tests/9225020#step/shutdown

### DIFF
--- a/lib/power_action_utils.pm
+++ b/lib/power_action_utils.pm
@@ -276,10 +276,17 @@ sub power_action {
             reboot_x11;
         }
         elsif ($action eq 'poweroff') {
-            if (is_backend_s390x) {
-                record_info('poo#114439', 'Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.');
-                select_console 'root-console';
-                enter_cmd "$action";
+            if ((is_s390x || is_backend_s390x) && check_var("DESKTOP", "gnome")) {
+                send_key "ctrl-alt-delete";
+                assert_screen 'logoutdialog', 15;
+                assert_and_click 'gnome-shell_shutdown_btn';
+                if (get_var("SHUTDOWN_NEEDS_AUTH")) {
+                    assert_screen 'shutdown-auth';
+                    type_password;
+                    # we need to take care of test module scheduled after shutdown like 'svirt_upload_assets'
+                    send_key "ret" && return 1 if is_backend_s390x;
+                    send_key "ret" if is_s390x;
+                }
             }
             else {
                 poweroff_x11;


lib/power_action_utils.pm
```diff
@@ -277,12 +277,23 @@ sub power_action {
...
+                    assert_screen 'logoutdialog', 15;
+                    assert_and_click 'gnome-shell_shutdown_btn';
+                    if (get_var("SHUTDOWN_NEEDS_AUTH")) {
+                        assert_screen 'shutdown-auth';
+                        type_password;
+                        # we need to kill all open ssh connections before the system shuts down
+                        prepare_system_shutdown;
+                        send_key "ret";
+                    }
+                }
```
rfan1, line 291: How about
```
if (check_var("DESKTOP", "gnome")) {
    poweroff_x11;
}
```
?

Zaoliang: tried already this is not working, zVM is different then zKVM, thanks for your input. Now I have changed this and will update with verification run :)

lib/power_action_utils.pm
```diff
@@ -276,14 +276,21 @@ sub power_action {
...
+            }
+            else {
                 record_info('poo#114439', 'Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.');
                 select_console 'root-console';
                 enter_cmd "$action";
             }
-            else {
-                poweroff_x11;
-            }
```
rfan1, line 286: @Zaoliang 
I noticed that you remove 'poweroff_x11' totally. I am wondering if it will impact the power_off action for other platforms [x86_64 aarch64 power s390x_zkvm etc]

Zaoliang: let me check ;)

Zaoliang: zkvm is not impacted, this is only related to zVM, gui or console,  see 'else' after 'if (is_backend_s390x && check_var("DESKTOP", "gnome"))' 
I've checked already tha kvm is fine, will provide latest test run anyway, thanks!

Zaoliang: I implemented actually the replacement of poweroff_x11 because it doesn't work for gui shutdown on zVM.

rfan1: https://openqa.suse.de/tests/8752474 take this job as an example on x86_64 platform.
I think it may change the current logic to shutdown the VM.

Zaoliang: https://openqa.suse.de/tests/9230121 shows that zkvm test is successful.

Zaoliang:
just for sure that this PR has no impact on zkvm :)

see https://openqa.suse.de/tests/9230121

lib/power_action_utils.pm
```diff
@@ -276,14 +276,21 @@ sub power_action {
             reboot_x11;
         }
         elsif ($action eq 'poweroff') {
-            if (is_backend_s390x) {
```
rfan1, line 279: I would suggest:
            if (is_backend_s390x) {
                if (check_var("DESKTOP", "gnome")) {
                    send_key "ctrl-alt-delete";
                    assert_screen 'logoutdialog', 15;
                    assert_and_click 'gnome-shell_shutdown_btn';
                    if (get_var("SHUTDOWN_NEEDS_AUTH")) {
                        assert_screen 'shutdown-auth';
                        type_password;
                        send_key "ret" && return 1;
                    }
                }
                else {
                    record_info('poo#114439', 'Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.');
                    enter_cmd "$action";
            }
            else {
                poweroff_x11;
            }

Zaoliang: no, I'm afraid to say your suggestion is not correct
because all other cases are covered like:
  # Shutdown takes longer than 60 seconds on SLE12 SP4 and SLE 15
    if (is_sle('12+') && check_var('DESKTOP', 'gnome') && ($action eq 'poweroff')) {
        $soft_fail_data = {bugref => 'bsc#1055462', soft_timeout => 60, timeout => $shutdown_timeout *= 3};

 Please see my verification test run for zkvm provided above.

rfan1: hmm, However, I checked the VR 
https://openqa.suse.de/tests/9230121#step/shutdown/10, I got 
`poo#114439 Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.

Do we need this workaround on s390x + zKVM? 

Anyway, please go ahead, actually, I am not familiar with this part, just wondering why you remove 'poweroff_x11' for all platforms.
`

rfan1: hmm, However, I checked the VR 
https://openqa.suse.de/tests/9230121#step/shutdown/10, I got 
`poo#114439 Temporary workaround, because shutdown module is marked as failed on s390x backend when shutting down from GUI.

Do we need this workaround on s390x + zKVM? 

Anyway, please go ahead, actually, I am not familiar with this part, just wondering why you remove 'poweroff_x11' for all platforms.

Zaoliang: I checked this again and I think now I can even to try fix both now, I wondered as well that record_info about 'poo#114439' refers to zkvm.  Actually the issue was only zVM, for for safety's sake, I just keep poweroff_x11 for the moment. I will check further for potenial issue.

Zaoliang:
New verification runs:
https://openqa.suse.de/tests/9232492 (zVM)
https://openqa.suse.de/tests/9232491 (zkvm)
https://openqa.suse.de/tests/9232490 (x86_64)

rfan1 (approved):
LGTM

rfan1:
@Zaoliang 
One minor comment here, since you have changed the logic for zvm and zkvm, I would suggest you add more VRs for SLES12Px  and SLE15SPx on s390x platform. then we can make sure the the code is strong enough :)

foursixnine:
Waiting for the VR's suggested by @rfan1

Zaoliang:
further verification run:
sles 15 sp3 
https://openqa.suse.de/tests/9242630 (ISO was replaced: SLE-15-SP3-Online-s390x-GM-Media1.iso )

Zaoliang:
looks good as well for sles 12 sp5 on zVM:
http://openqa.suse.de/tests/9242536

Zaoliang:
> Waiting for the VR's suggested by @rfan1

done, please check, thanks!

Zaoliang:
@foursixnine can you please check and approve or merge this? thanks!

PR #15278, opened 2022-07-25
Delete drawboard function from WebSocketsTest

rfan1:
Due to poo#114532, in the sake of stability, that we can safely
delete the function that tests the drawboard example.
We already cover the rest of Tomcat's WebSocket examples,
and, in reality, this drawboard example is not really tested
with multiple connections as it would ideally be the case.

At the same time, increase the timeout value for some tomcat
test modules to reduce the impact by performance.

- Related ticket: https://progress.opensuse.org/issues/114532
- Needles:n/a
- Verification run: http://openqa.suse.de/tests/9208849

### DIFF
--- a/lib/Tomcat/JspTest.pm
+++ b/lib/Tomcat/JspTest.pm
@@ -16,7 +16,7 @@ use Tomcat::Utils;
 use version_utils 'is_sle';
 
 # allow a 60 second timeout for asserting needles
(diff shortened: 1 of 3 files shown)


Zaoliang:
you can remove from commit msg:

a-kpappas (approved):

PR #15284, opened 2022-07-27
Add "FULLRUN_MONITOR" to support for monitoring the status of hosts

badboywj:
After enabling "FULLRUN_MONITOR" we can let openQA to monitor the status of hosts, and openQA instance will wait until the testing finish. So openQA instances finish after our testings finish.

**Note**: if you plan to enable "FULLRUN_MONITOR", it is best to work with "MAX_JOB_TIME=<**seconds**>" (**seconds** depends on the time that your testing takes), for example, I will set
"MAX_JOB_TIME=86400" when running our testing.

- Verification run: http://10.67.129.4/tests/50035
- Related ticket: N/A
- Needles: N/A

### DIFF
--- a/tests/kernel_performance/full_run.pm
+++ b/tests/kernel_performance/full_run.pm
@@ -15,6 +15,7 @@ use warnings;
 use testapi;
 use base 'y2_installbase';
 use File::Basename;
+use LWP::Simple;
 
 sub full_run {
     my $time_out //= 6;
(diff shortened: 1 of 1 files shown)


Dawei-Pang:
The code is OK for me.
@JoyceNa @alice-suse would you please help to review? Thanks!

JoyceNa:
LGTM

PR #15287, opened 2022-07-27
kubectl: Fix performance issue when scaling in

pdostal:
- Failure: [openqa.opensuse.org](https://openqa.opensuse.org/tests/2484785)
- Verification run: [openqa.opensuse.org](https://openqa.opensuse.org/tests/2485535)

### DIFF
--- a/tests/containers/kubectl.pm
+++ b/tests/containers/kubectl.pm
@@ -102,6 +102,8 @@ sub run {
     validate_script_output('kubectl describe deployments/nginx-deployment | grep Replicas', sub { $_ =~ m/5 desired.*5 total/ });
     # Scale in
     assert_script_run('kubectl scale --replicas=2 deployments/nginx-deployment');
+    # When scaling healthy cluster in the deployment condition is always available even when more than specified amount of replicas still exist
+    validate_script_output_retry('kubectl get events | grep ScalingReplicaSet | tail -n1', qr/Scaled down replica set nginx-deployment-.*to 2/, retry => 6, delay => 20, timeout => 10);
     assert_script_run('kubectl wait deployment nginx-deployment --for condition=available --timeout=300s', timeout => 330);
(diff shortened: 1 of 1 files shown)


tests/containers/kubectl.pm
```diff
@@ -102,6 +102,8 @@ sub run {
     validate_script_output('kubectl describe deployments/nginx-deployment | grep Replicas', sub { $_ =~ m/5 desired.*5 total/ });
     # Scale in
     assert_script_run('kubectl scale --replicas=2 deployments/nginx-deployment');
+    # When scaling healthy cluster in the deployment condition is always available even when more than specified amount of replicas still exist
+    validate_script_output('kubectl get events | grep ScalingReplicaSet | tail -n1', sub { $_ =~ m/Scaled down replica set nginx-deployment-.*to 2/ });
     assert_script_run('kubectl wait deployment nginx-deployment --for condition=available --timeout=300s', timeout => 330);
     validate_script_output('kubectl describe deployments/nginx-deployment | grep Replicas', sub { $_ =~ m/2 desired.*2 total/ });
```
ilausuch, line 108: I think line 108 "kubectl desctibe | grep Replicas" still is incorrect and could generate the same error again. Yes we'll have 2 desired, but maybe not 2 in total because is a scaling down and "kubectl wait" is not doing what we think here because is not blocking until the pods are destroyed.

ilausuch (changes requested):

tests/containers/kubectl.pm
pdostal: @ilausuch please check now.

jlausuch (approved):

ilausuch (approved):

tests/containers/kubectl.pm
ilausuch: Thank you for your explanation (in a meeting) :)

PR #15291, opened 2022-07-28
Add addon_dud with restapi in installer

tinawang123:
Replaces the module dud_addon by one that uses libyui-rest-api.

- Related ticket: https://progress.opensuse.org/issues/113734
- Verification run: https://openqa.suse.de/tests/9235799#step/addon_dud/1

### DIFF
--- a/lib/Installation/ProductSelection/AdditionalProductPage.pm
+++ b/lib/Installation/ProductSelection/AdditionalProductPage.pm
@@ -1,0 +1,38 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Additional Prodcuts page
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::ProductSelection::AdditionalProductPage;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my ($self) = @_;
+    $self->{btn_add} = $self->{app}->button({id => 'ok'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{btn_add}->exist();
+}
+
+sub press_add {
+    my ($self) = @_;
+    return $self->{btn_add}->click();
+}
+
+1;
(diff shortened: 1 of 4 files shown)


lib/Installation/ProductSelection/AdditionalProductPage.pm
```diff
@@ -0,0 +1,38 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Additional Prodcuts page
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package Installation::ProductSelection::AdditionalProductPage;
+use parent 'Installation::Navigation::NavigationBase';
```
coolgw, line 10: My view it should belong lib/Installation/SystemProbing class since https://openqa.suse.de/tests/9231215#step/dud_addon/1 show background page is system probing.

lib/Installation/ProductSelection/AdditionalProductPage.pm
```diff
@@ -0,0 +1,38 @@
...
+
+sub init {
+    my ($self) = @_;
+    $self->{btn_add} = $self->{app}->button({id => 'ok'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{btn_add}->exist();
```
coolgw, line 30: I suppose we can add more condition to confirm this page exist not only check ok button?

lib/Installation/ProductSelection/AdditionalProductPage.pm
tinawang123: It is only for additional products page. Not related with background page.

BillAnastasiadis:
Please create a new file instead of replacing `dud_addon.pm`, to have both libyui and non-libyui ways to add products. Then, replace `dud_addon` with your new file in [schedule/yast/dud_development_tools.yaml](https://openqa.suse.de/tests/9231215/settings/schedule/yast/dud_development_tools.yaml).

lib/Installation/ProductSelection/AdditionalProductPage.pm
BillAnastasiadis: I guess it's fine, this is how the `is_shown()` function is implemented in other pages as well.

BillAnastasiadis:
LGTM!

lib/Installation/ProductSelection/AdditionalProductPage.pm
tinawang123: As this is new page, the background page don't have ok button. So use this can ensure the 'additional products' page pop up.

BillAnastasiadis: BTW: Instead of writing it as a page, since it's a popup, you could also write it as a popup (essentially, it's just a page that just has "popup" in the file title - see the `lib/Installation/Popups` folder).

However it's not important, and you've done a great work here, so I'm not asking to change it - it looks good to me as it is.

jknphy (commented):
Please, address this changes in follow-up PRs.

PR #15292, opened 2022-07-28
installation/bootloader_uefi: Avoid sleeping during the grub timeout

Vogtinator:
This sleep makes it much more likely to miss the grub menu with its default
10s timeout. The reason why this was introduced is not clear, remove it.

Before: https://openqa.opensuse.org/tests/2485325 - >80% failure rate

After: https://openqa.opensuse.org/tests/2485579 - https://openqa.opensuse.org/tests/2485584 - all green


mloviska (approved):
LGTM, very nice catch! Thanks!

Vogtinator:
Let's see whether this breaks some unexpected scenario...

PR #15294, opened 2022-07-28
Run list-timers before shutdown

michaelgrifalconi:
Run list-timers before shutdown and print the result to the console

- Related ticket: https://progress.opensuse.org/issues/114433
- Needles: none
- Verification run:
  - https://openqa.suse.de/t9231576
  - https://openqa.suse.de/tests/9231577
  - https://openqa.suse.de/tests/9231578

### DIFF
--- a/tests/shutdown/shutdown.pm
+++ b/tests/shutdown/shutdown.pm
@@ -13,6 +13,9 @@ use power_action_utils 'power_action';
 use utils;
 
 sub run {
+    my $self = shift;
+    select_console('root-console');
+    script_run('systemctl list-timers --all');
     power_action('poweroff');
 }
 


tests/shutdown/shutdown.pm
```diff
@@ -13,6 +13,9 @@ use power_action_utils 'power_action';
 use utils;
 
 sub run {
+    my $self = shift;
+    select_console('root-console');
+    script_run('systemctl list-timers --all');
```
b10n1k, line 18: i would recommend to replace this with `$self->save_and_upload_log('systemctl list-timers --all', '/tmp/all_timers.log', {screenshot => 1});` . that will store a file in the *Logs and Assets* wdyt?

michaelgrifalconi: Hi, upload won't work because `shutdown` is run after `cleanup_before_shutdown` that takes down the network. See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/0385b0e789210c0ac7d6af10b9c9bf3a0c32331e/tests/shutdown/cleanup_before_shutdown.pm#L57-L62
and https://openqa.suse.de/tests/9230015#step/shutdown/6

I guess it should be enough to print it since it's a rather small output and easy to read, but open to ideas

mloviska: Seems like this breaks the test https://openqa.suse.de/tests/9239946#step/shutdown/2.
I would recommend to use https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/Utils/Systemd.pm#L58 instead of `script_run`

PR #15299, opened 2022-07-29
Add test to add maintenance test repos

BillAnastasiadis:
Create a test module to add maintenance test repos using libyui instead of needles, the libyui equivalent of [add_update_test_repo.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/installation/add_update_test_repo.pm).

- Related ticket: https://openqa.suse.de/tests/9406283
- Verification run: https://openqa.suse.de/tests/9418773

gitlab mr: https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/369

### DIFF
--- a/schedule/yast/qam-yast_self_update_libyui.yaml
+++ b/schedule/yast/qam-yast_self_update_libyui.yaml
@@ -1,0 +1,37 @@
+---
+name: qam-yast_self_update_libyui
+description: installation using self_update as boot parameter
+vars:
+  YUI_REST_API: 1
+schedule:
+  - installation/bootloader_start
+  - installation/setup_libyui
+  - installation/product_selection/install_SLES
+  - installation/validate_self_update
+  - installation/licensing/accept_license
+  - installation/registration/register_via_scc
+  - installation/module_registration/register_extensions_and_modules
+  - installation/add_on_product/add_maintenance_repos
+  - installation/addon_products_sle
+  - installation/partitioning
+  - installation/partitioning_finish
+  - installation/installer_timezone
+  - installation/hostname_inst
+  - installation/user_settings
+  - installation/user_settings_root
+  - installation/resolve_dependency_issues
+  - installation/installation_overview
+  - installation/disable_grub_timeout
+  - installation/start_install
+  - installation/await_install
+  - installation/logs_from_installation_system
+  - installation/reboot_after_installation
+  - installation/grub_test
+  - installation/first_boot
+  - '{{efi}}'
+conditional_schedule:
+  efi:
+    MACHINE:
+      uefi:
+        - console/consoletest_setup
+        - console/verify_efi_mok
--- a/tests/installation/add_on_product/add_maintenance_repos.pm
+++ b/tests/installation/add_on_product/add_maintenance_repos.pm
@@ -1,0 +1,43 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Add new add-ons using maintenance test repo URLs.
+#
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+sub create_repo_name {
+    my ($repo_id, $addon) = $_[0] =~ (/(\d+)\/(.*)\//);
+    my $name = $addon . "_" . $repo_id;
+    return $name;
+}
+
+sub run {
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
+
+    my $maintrepo = shift @repos;
+    my $reponame = create_repo_name($maintrepo);
+    $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
+    save_screenshot;
+    $testapi::distri->get_repository_url()->add_repo({url => $maintrepo, name => $reponame});
+    save_screenshot;
+
+    for my $repo (@repos) {
+        $testapi::distri->get_add_on_product_installation()->add_add_on_product();
+        $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
+        my $name = create_repo_name($repo);
+        $testapi::distri->get_repository_url()->add_repo({url => $repo, name => $name});
+        save_screenshot;
+    }
+    $testapi::distri->get_add_on_product_installation()->accept_add_on_products();
+}
+
+1;


tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+sub run {
+    set_var('MAINT_TEST_REPO', get_var('INCIDENT_REPO')) if get_var('INCIDENT_REPO');
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
+
+    while (defined(my $maintrepo = shift @repos)) {
+        $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
+        $testapi::distri->get_repository_url()->add_repo({url => $maintrepo});
+        if (@repos) {
```
b10n1k, line 24: i cant see the reason for this /if/ condition. The results would be identical with or without. right?

coolgw:
> If you want to see all the steps you can do what I suggest in this ticket https://progress.opensuse.org/issues/107824, basically, make a screenshot every time you use is_shown, or maybe you can find a better hack. LGTM.

If too much screen shot happen, maybe we can add another setting for control the screen show or not

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: I believe it's a remnant from an earlier version where I pressed next as a last step in the loop, instead of outside. I removed it and running a VR to make sure all is ok. Thanks!

BillAnastasiadis: @b10n1k my bad - this is actually needed. We want to add another repo only if there's another repo to add. If `repos` is depleted, we want to move to the next page and not add another one.

b10n1k: maybe add this small detail as a comment? it looks a bit weird in first glance but its up to you and your team. i am gonna resolve the conversation for now

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi 'get_var';
+
+sub run {
+    set_var('MAINT_TEST_REPO', get_var('INCIDENT_REPO')) if get_var('INCIDENT_REPO');
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
```
ilmanzo, line 17: is setting the var 'MAINT_TEST_REPO' needed outside the code ?
Otherwise maybe concept can be expressed in a concise way: use "foo" if available OR "bar" if not 

```perl
my @repos = split(/,/, get_var('INCIDENT_REPO') || get_var('MAINT_TEST_REPO'));
```

and get rid of the double `get_var ` at 16

BillAnastasiadis: True, but this is a direct replication of the code of the original file (see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/installation/add_update_test_repo.pm) - and I'm not taking the risk of changing it :p

jknphy:
> > If you want to see all the steps you can do what I suggest in this ticket https://progress.opensuse.org/issues/107824, basically, make a screenshot every time you use is_shown, or maybe you can find a better hack. LGTM.
> 
> If too much screen shot happen, maybe we can add another setting for control the screen show or not

I have the same concern when making workable this ticket, we should revisit and perhaps try to find some code structure for step, otherwise we will depend on settings to control something that could be out of control... to have it in one screenshot looks good to me for now.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
+
+    while (defined(my $maintrepo = shift @repos)) {
```
jknphy, line 21: why this safety is required? why not just iterate `@repos`, `sub map_incidents_to_repo` should be responsible to give us a good mapping without undef values IMO, maybe already does...and relay in settings that we can assume always exist. I would simplify the code, specially in the test layer things should as simple as possible.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+        $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
```
jknphy, line 22: another screenshot after this line would help as well.

tests/installation/add_maintenance_repos.pm
jknphy, line 24: this is not needed, just iterate the array please.
btw, no need for screenshots here, it would be an overkill.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+        $testapi::distri->get_repository_url()->add_repo({url => $maintrepo});
```
jknphy, line 23: we are missing `name` parameter, as we wanted to do something similar to what we have in AutoYaST, it is more clear to have the repos with names in the installed system than with some hashes that are created when no name is introduced.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: It is needed, because in the case this is the last item (meaning `@repos` now returns false due to last shift) the code must not press the button to add a new repo, but instead press next and continue with the installation.

This cannot take place at the begining of the loop (as it must not take place during the first iteration), so it needs to be conditional.

There are other ways to do it (like having the first iteration outside of the loop) but it leads to lengthier and even more confusing code. On the other hand, this is also a replication of the current behavior of the (non-libyui) module.

jknphy: the code should reflect what is happening which is ... (no surprise, a bad UI design, but it is what it is). Therefore I would expect to see:
- Add-on-product page: click checkbox 'I would like to ...'
- (optional) I think there is not need to click on "specify url' radiobutton as it is  selected by default.
- Repository-url page: enter name and url and next
- Add-on-product-installation-page: click on add button
- (optional) add-on-product page: looks like the radiobutton is still selected by default and now we don't have "I would like to ...". Therefore this step should be "accept current options method" so basically just next behind the scene.

So yes, it would be more helpful to shift the first element and treat it differently as the UI is also different and then do the loops with N-1. It exactly how UI does. It would mimic better the behavior we see.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: This module reads the maintenance repos as they are given in the relevant variables, meaning the repos separated by comas, no name. This is also the case for the original module. How would we insert names there? We would have to change the variable structure and way they are read.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,31 @@
...
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi 'get_var';
+
+sub run {
+    set_var('MAINT_TEST_REPO', get_var('INCIDENT_REPO')) if get_var('INCIDENT_REPO');
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
```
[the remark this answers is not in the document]
BillAnastasiadis: Yes, that sounds like a good idea.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: True, but does it really help with anything? I mean, it's something that we would (probably) have to change again if we run single incidents, and it doesn't have any difference when not in single incidents. Plus, if someone decides to run single incidents with libyui, the can :p 

But I'll remove it for now.

schedule/qam/common/qam-yast-self-update.yaml
```diff
@@ -3,11 +3,11 @@ name: qam-yast_self_update
 description: self_update via INSTALLER repo & efi test on UEFI
 schedule:
 - installation/bootloader_start
+- installation/setup_libyui
 - installation/welcome
 - installation/scc_registration
-- installation/add_update_test_repo
+- installation/add_maintenance_repos
```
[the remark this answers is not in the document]
BillAnastasiadis: Yes, this is here just for the ver runs (I run them on osd). Will remove before merge, leaving only the module.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: Agreed, that kind of conditional problem (first iteration outside vs conditional action inside the loop) is always annoying. I almost always chose the `less code` option by default, even if it means one extra conditional for every loop iteration, but that's just habit.

I re-wrote it removing the loop and using foreach, running a ver run and update the pr: https://openqa.suse.de/tests/9386975#live

BillAnastasiadis: Done

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: I did what we discussed in the other comment, so this is done as well

schedule/qam/common/qam-yast-self-update.yaml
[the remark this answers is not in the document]
jknphy: we need to apply at least to on job for real (typical acceptance criteria in agile of doing small parts but always touch production even if it is in a minimal way, you don't need to apply everywhere for example), we need to create an additional schedule for SLE-15-SP(3,4} and reflect that in MR in development group for MUs.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,32 @@
...
+use testapi 'get_var';
+
+sub run {
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
+
+    my $maintrepo = shift @repos;
+    $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
+    $testapi::distri->get_repository_url()->add_repo({url => $maintrepo});
```
jknphy, line 22: we still need to add the name of the repo.
see suggestion: https://progress.opensuse.org/issues/113731
we did for AutoYaST because it might help to investigate and it is cheap to type it with libyui-rest-api.

schedule/yast/qam-yast_self_update.yaml
```diff
@@ -11,9 +11,8 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_registration/register_extensions_and_modules
-  - installation/add_update_test_repo
+  - installation/add_maintenance_repos
   - installation/addon_products_sle
```
jknphy, line 15: we don't need this test module anymore, right? or maybe you left it because following test modules are not libyui-rest-api and something failed? if so, it is ok, in other ticket we could just move all to libyui.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: Sure, adding a name for the repo is super easy... but where would the repo names be coming from? I mean, `MAINT_TEST_REPO` does not include repo names, only urls. Can you point me to the way it's done in autoyast?

jknphy: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15392#discussion_r954548934

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,37 @@
...
+    save_screenshot;
+
+    foreach (@repos) {
+        $testapi::distri->get_add_on_product_installation()->add_add_on_product();
+        $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
+        my $repo = $_;
```
jknphy, line 28: this is not usually done like that, you can just `for my $repo (@repos)`.
always avoid to declare variable, because Perl is thought to be more expressive (except when it is too cryptic, of course).

tests/installation/add_maintenance_repos.pm
jknphy: now first repo doesn't have name. feel free to extract the function on top.

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,37 @@
...
+        my ($repo_id, $addon) = $repo =~ (/(\d+)\/(.*)\//);
+        my $name = $addon . "_" . $repo_id;
+        $testapi::distri->get_repository_url()->add_repo({url => $repo, name => $name});
+        save_screenshot;
```
jknphy, line 32: I think we just need an screenshot after line 21 in the whole file, otherwise when there will be 40 repos, it is too much. wdyt?

schedule/yast/qam-yast_self_update.yaml
```diff
@@ -11,9 +11,8 @@ schedule:
   - installation/licensing/accept_license
```
jknphy, line 11: please create a new file to be used only in a copy of this test suite in MUs development as discussed offline.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: I'm thinking, we could show a screenshot for each repo. This way one can check repo-specific stuff, should they want to. Even check names, too. Plus, it will resemble the previous needle based behavior (only this time we will strictly get 1 screenshot per repo).

If you think we don't need that information, then I think that even no screenshot at all is fine. But I can insert one at line 19. Should I do that?

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: done

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: done!

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,42 @@
+# SUSE's openQA tests
```
jknphy, line 1: we always use a folder to distinguish libyui for non-libyui (not written anywhere, sorry, you didn't have to know that). I guess it should belong to `tests/installation/add_on_product`

tests/installation/add_maintenance_repos.pm
```diff
@@ -0,0 +1,42 @@
...
+}
+
+sub run {
+    my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+
+    $testapi::distri->get_add_on_product()->confirm_like_additional_add_on();
+
+    my $maintrepo = shift @repos;
+    my $reponame = name_repo($maintrepo);
+    $testapi::distri->get_add_on_product()->accept_current_media_type_selection();
```
jknphy, line 28: could we please add a screenshot after this line to see every single screen in the test? just for completeness.

tests/installation/add_maintenance_repos.pm
jknphy: no, it looks very neat as you designed! good choice!

jknphy (commented):
LGTM, just a few minor details.

jknphy:
Could you please add the MR in the first comment and add the link to this PR in the first comment of your MR, we use this trick to navigate back and forth when reviewing/demo.

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: That's good to know, it may not be ideal but there's some logic to differentiate between libyui and non-libyui. Done!

tests/installation/add_maintenance_repos.pm
BillAnastasiadis: Sure, done!

PR #15307, opened 2022-08-01
Fix isotovideo test by removing colors

foursixnine:
https://github.com/os-autoinst/os-autoinst/pull/2134 added colors to line endings, which make them in fact, different to what the expected output we have.

First test is to see if it works. Altered diff command too to get more info

https://progress.opensuse.org/issues/114869

### DIFF
--- a/.github/workflows/ci-isotovideo.yml
+++ b/.github/workflows/ci-isotovideo.yml
@@ -10,3 +10,6 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run isotovideo tests
       run: make test-isotovideo
+    - name: Error in isotovideo
+      if: ${{ failure() }}
+      run: echo 'isotovideo failed, something modified the default schedule or the output; run locally to find out more.' && exit 1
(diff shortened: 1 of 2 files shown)


foursixnine:
Now we even have a nice message: https://github.com/os-autoinst/os-autoinst-distri-opensuse/runs/7610502431?check_suite_focus=true

perlpunk (approved):

jlausuch (approved):
Thanks for fixing it!

perlpunk:
Sorry for breaking it /o\

foursixnine:
> Sorry for breaking it /o\

We can sing: _Oops it's broken again..._

.github/workflows/ci-isotovideo.yml
```diff
@@ -10,3 +10,6 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run isotovideo tests
       run: make test-isotovideo
+    - name: Error in isotovideo
+      if: ${{ failure() }}
+      run: echo 'isotovideo failed, something modified the default schedule or the output; run locally to find out more.' && exit 1
```
b10n1k, line 15: Correct me if i am wrong but why the message cant be in the same line with `make test-isotovideo`?

foursixnine: Nothing to correct, I guess the message can be in the same place, since the second run also executes both statements.

PR #15308, opened 2022-08-01
Revert "Run list-timers before shutdown"

sysrich:
Reverts os-autoinst/os-autoinst-distri-opensuse#15294

This breaks Factory

# Test died: command 'systemctl list-timers --all' timed out at /usr/lib/os-autoinst/testapi.pm line 1008.

eg. https://openqa.opensuse.org/tests/2490184#step/shutdown/4


foursixnine:
@michaelgrifalconi perhaps adding the no-pager option should fix it

PR #15304, opened 2022-08-01
Avoid kubectl installation if it is already installed

ilausuch:
[type description here, PLEASE, REMOVE THIS LINE, PLACEHOLDER, BEFORE SUBMITTING THIS PULL REQUEST]

- Related ticket: https://progress.opensuse.org/issues/xyz
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/xyz
- Verification run: openqa.mypersonalinstance.de/tests/xyz#step/module/x

PR #15310, opened 2022-08-01
Fix kubectl in public cloud tools image

pdostal:
Problems:
1) The `bitnami/kubectl` container image expect `/.kube/config` instead of `~/.kube/config`
2) The `~/` and `/tmp` from the hosts need to be accessible from the container
3) The container runs user with ID `1001` so there were permission issues

- Verification run: [pdostal-server.suse.cz](https://pdostal-server.suse.cz/tests/348)
- Work initiated by: @ilausuch in #15304

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -47,18 +47,23 @@ sub uninstall_k3s {
 }
 
 sub install_kubectl {
-    if (is_sle) {
-        # kubectl is in the container module
-        add_suseconnect_product(get_addon_fullname('contm'));
-        # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
-        if (is_sle(">15-SP1")) {
-            zypper_call("in kubernetes1.18-client");
+    if (script_run("which kubectl") != 0) {
+        if (is_sle) {
+            # kubectl is in the container module
+            add_suseconnect_product(get_addon_fullname('contm'));
+            # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
+            if (is_sle(">15-SP1")) {
+                zypper_call("in kubernetes1.18-client");
+            } else {
+                zypper_call("in kubernetes-client");
+            }
         } else {
             zypper_call("in kubernetes-client");
         }
     } else {
-        zypper_call("in kubernetes-client");
+        record_info('kubectl preinstalled', 'The kubectl package is already installed.');
     }
+    record_info('kubectl version', script_output('kubectl version', proceed_on_failure => 1));
 }
 
 sub install_helm {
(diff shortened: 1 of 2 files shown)


lib/containers/k8s.pm
```diff
@@ -47,17 +47,21 @@ sub uninstall_k3s {
...
+                zypper_call("in kubernetes1.18-client");
+            } else {
+                zypper_call("in kubernetes-client");
+            }
         } else {
             zypper_call("in kubernetes-client");
         }
     } else {
-        zypper_call("in kubernetes-client");
+        record_info('kubectl already', 'The kubectl package is already installed.');
```
jlausuch, line 64:
```suggestion
        record_info('kubectl', 'The kubectl package is already installed.');
```
`kubectl already` alone sounds a bit weird :) 

Also, what about printing the version in another record_info ?

pdostal: Done.

lib/containers/k8s.pm
```diff
@@ -47,17 +47,22 @@ sub uninstall_k3s {
...
+            } else {
+                zypper_call("in kubernetes-client");
+            }
         } else {
             zypper_call("in kubernetes-client");
         }
     } else {
-        zypper_call("in kubernetes-client");
+        record_info('kubectl preinstalled', 'The kubectl package is already installed.');
+        record_info('kubectl version', script_output('kubectl version', proceed_on_failure => 1));
```
jlausuch, line 65: but I also meant in both cases, when it's installed and when not. It would good to see the version output also when installing it. Maybe you can add it at the end, outside of the if-else.

pdostal: Oh yes, I'm sorry, it was late :smile:

Done now.

jlausuch (approved):
LGTM

lib/containers/k8s.pm
```diff
@@ -47,18 +47,23 @@ sub uninstall_k3s {
...
+                zypper_call("in kubernetes-client");
+            }
         } else {
             zypper_call("in kubernetes-client");
         }
     } else {
-        zypper_call("in kubernetes-client");
+        record_info('kubectl preinstalled', 'The kubectl package is already installed.');
     }
+    record_info('kubectl version', script_output('kubectl version', proceed_on_failure => 1));
```
asmorodskyi, line 66: why do we proceed on failure here ?

pdostal: Because `kubectl version` exits with `1` when no server is defined.

asmorodskyi: ok

asmorodskyi (approved):
LGTM

ilausuch (approved):
Really nice work. For future changes will be nice to have the kubectl_version as a optional settings variable

PR #15313, opened 2022-08-02
Containers: Create our own helm chart

pdostal:
This adds our own helm chart. The chart right now deploys only a very simple job.
To test the 'helm repository' feature we would need to get a helm repository.

So please take this as a first alpha over which we should discuss how to proceed.

- Related ticket: [poo#112517](https://progress.opensuse.org/issues/112517)
- Verification run: [pdostal-server.suse.cz](https://pdostal-server.suse.cz/tests/416)
- Issues found: k3s-io/k3s#5946

### DIFF
--- a/data/containers/helm-test/templates/job.yaml
+++ b/data/containers/helm-test/templates/job.yaml
@@ -1,0 +1,21 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}-job"
+spec:
+  template:
+    spec:
+      containers:
+      - image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.imagePullPolicy }}
+        name: helm-test-{{ .Values.job_id }}
+        command: [ "cat", "/etc/os-release" ]
+        resources:
+          limits:
+            cpu: "0.5"
+            memory: 256Mi
+          requests:
+            cpu: "0.1"
+            memory: 128Mi
+      restartPolicy: Never
+  backoffLimit: 4
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -25,6 +25,7 @@ use publiccloud::utils qw(gcloud_install);
 
 sub run {
     my ($self, $run_args) = @_;
+    my $job_id = get_current_job_id();
 
     # We either test various cloud clusters or local k3s
     my $k8s_backend = shift(@{$run_args->{backends}});
@@ -35,7 +36,6 @@ sub run {
     $self->{is_k3s} = $is_k3s;
 
     $self->select_serial_terminal;
-    $self->{deployment_name} = "apache-" . get_current_job_id();
     my $chart = "bitnami/apache";
 
     record_info("Chart name", $chart);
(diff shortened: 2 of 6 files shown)


data/containers/helm-test/templates/job.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: "{{ .Release.Name }}-job"
+spec:
+  template:
+    spec:
+      containers:
+      - image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.imagePullPolicy }}
+        name: helm-test-{{ .Values.job_id }}
+        command: [ "cat", "/etc/os-release" ]
+        resources:
```
ilausuch, line 13: Should be nice to have the limits on variables with a default value

data/containers/helm-test/templates/job.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}-job"
```
ilausuch, line 4: I am thinking that for each deployment in testing we should create a namespace. Then we can destroy the namespace

tests/containers/helm.pm
```diff
@@ -83,6 +84,17 @@ sub run {
...
+    assert_script_run('curl --create-dir -vo ~/helm-test/Chart.yaml '.data_url('containers/helm-test/').'Chart.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/values.yaml '.data_url('containers/helm-test/').'values.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/templates/job.yaml '.data_url('containers/helm-test/templates/').'job.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/templates/NOTES.txt '.data_url('containers/helm-test/templates/').'NOTES.txt');
+
+    assert_script_run("helm install helm-test-$job_id ~/helm-test/ --values ~/helm-test/values.yaml --set job_id=$job_id");
+    assert_script_run("helm list");
+    my $pod = script_output('kubectl get pods -o name --no-headers=true | grep helm-test-$job_id');
+    script_retry("kubectl logs $pod | grep 'SUSE'", retry => 5, delay => 10);
+    assert_script_run("helm uninstall helm-test-$job_id");
```
ilausuch, line 96: I think the uninstall should be added in case the test fails

ilausuch (changes requested):
Good job. But I would like to comment some ideas to improve this a little more

data/containers/helm-test/templates/job.yaml
pdostal: I was thinking actually the same. Let me implement it.

data/containers/helm-test/templates/job.yaml
pdostal: Do you think that this is needed? The variables are defined just in another file.

pdostal:
I think so. I think we should get a helm repository. But as this is in clouds it would need to be public.

data/containers/helm-test/templates/job.yaml
pdostal: Done.

tests/containers/helm.pm
pdostal: Done.

ilausuch (approved):

tests/containers/helm.pm
```diff
@@ -83,41 +83,43 @@ sub run {
 
     install_helm();
 
+    assert_script_run('curl --create-dir -vo ~/helm-test/Chart.yaml ' . data_url('containers/helm-test/') . 'Chart.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/values.yaml ' . data_url('containers/helm-test/') . 'values.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/templates/job.yaml ' . data_url('containers/helm-test/templates/') . 'job.yaml');
+    assert_script_run('curl --create-dir -vo ~/helm-test/templates/NOTES.txt ' . data_url('containers/helm-test/templates/') . 'NOTES.txt');
+
+    assert_script_run("kubectl create namespace helm-ns-$job_id");
```
ilausuch, line 91: Perfect!

tests/containers/helm.pm
```diff
@@ -83,41 +83,43 @@ sub run {
...
-    assert_script_run("helm delete $self->{deployment_name}");
     uninstall_k3s() if $is_k3s;
 }
 
 sub post_fail_hook {
     my ($self) = @_;
-    script_run("helm delete $self->{deployment_name}");
+    my $job_id = get_current_job_id();
+
+    assert_script_run("helm uninstall helm-test-$job_id");
```
grisu48, line 118: `assert_script_run` in the `post_fail_hook` is asking for trouble. I'd use `script_run` to avoid failures and incomprehensible error messages here.

pdostal: And it happened 🎉

grisu48 (approved):

PR #15315, opened 2022-08-03
install_ltp: Add missing prepare_ltp_env for sle < 12

pevik:
Fixes: fe8b2f909 ("Allow installing and running LTP tests without reboot in between")


pevik:
@mdoucha As SLE11 is EOL, maybe it'd be better to get rid of `is_sle('<12')` conditions? There are quite a lot of them. WDYT?

mdoucha:
We still have one live SLE-11SP3 codestream. Maybe next year.

pevik:
Thx for info. In that case we need this fix for SLE11 when testing network tests via git (corner case but should be tested).

mdoucha:
I doubt we'll ever use this code path on SLE-11SP3 but it doesn't hurt to fix it.

PR #15324, opened 2022-08-04
Skip ha/drbd_passive test when not in TWO_NODES scenario

alvarocarvajald:
Currently the 3 cluster node scenario tested for Maintenance jobs is [always scheduling](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/schedule/ha/qam/common/qam_cluster_node.yaml#L60-L61) the `ha/drbd_passive` test module, even when this particular test module is not designed to work in cluster scenarios with more than 2 nodes.

As a result, whenever a MU job is triggered with a package that requires a drbd test, the 3 node scenario job will fail as it is skipped in the third node before any of the `barrier_wait()` calls. This means the other nodes, remain blocked in a `barrier_wait()` call until they reach **MAX_JOB_TIME** and fail the whole scenario.

One possible solution would be to create a different schedule for the 3 cluster node scenario, but this implies a PR to this repo, as well as a change to the job settings.

This PR instead modifies `ha/drbd_passive` so it is skipped when running in scenarios with the setting **TWO_NODES=no**. It will also skip the test module `ha/filesystem` that is usually scheduled right after `ha/drbd_passive`, as there would be no block device on which to test the filesystem creation.

- Related ticket: https://progress.opensuse.org/issues/114727
- Needles: N/A
- Failing job: https://openqa.suse.de/tests/9247409#step/drbd_passive/5
- Verification runs:
+ QAM jobs (3 nodes): [node1](http://mango.qa.suse.de/tests/4871), [node 2](http://mango.qa.suse.de/tests/4872), [node 3](http://mango.qa.suse.de/tests/4873) & [support server](http://mango.qa.suse.de/tests/4870)
+ QAM jobs (2 nodes): [node 1](http://mango.qa.suse.de/tests/4875), [node 2](http://mango.qa.suse.de/tests/4876), [client](http://mango.qa.suse.de/tests/4877) & [support server](http://mango.qa.suse.de/tests/4874)
+ Alpha Cluster on Milestone Build Validation: [node 1](http://mango.qa.suse.de/tests/4860), [node 2](http://mango.qa.suse.de/tests/4861) & [support server](http://mango.qa.suse.de/tests/4859)

P.S.: I also include here a small optimization in `ha/vg` test module to avoid having multiple calls to `hacluster::read_tag` which sends commands to SUT.

### DIFF
--- a/tests/ha/filesystem.pm
+++ b/tests/ha/filesystem.pm
@@ -17,7 +17,8 @@ use hacluster;
 
 sub run {
     # Exit of this module if 'tag=drbd_passive' and if we are in a maintenance update not related to drbd
-    return 1 if (read_tag eq 'drbd_passive' and is_not_maintenance_update('drbd'));
+    my $tag = read_tag;
+    return 1 if (($tag eq 'drbd_passive' and is_not_maintenance_update('drbd')) or $tag eq 'skip_fs_test');
 
     my $cluster_name = get_cluster_name;
     my $node = get_hostname;
@@ -28,16 +29,16 @@ sub run {
     my $fs_opts = '-F -N 16';    # Force the filesystem creation and allows 16 nodes
 
     # This Filesystem test can be called multiple time
-    if (read_tag eq 'cluster_md') {
+    if ($tag eq 'cluster_md') {
         $resource = 'cluster_md';
     }
-    elsif (read_tag eq 'drbd_passive') {
+    elsif ($tag eq 'drbd_passive') {
         $resource = 'drbd_passive';
         $fs_lun = '/dev/drbd_passive' if is_node(1);
         $fs_type = 'xfs';
         $fs_opts = '-f';
     }
-    elsif (read_tag eq 'drbd_active') {
+    elsif ($tag eq 'drbd_active') {
         $resource = 'drbd_active';
     }
     else {
--- a/tests/ha/drbd_passive.pm
+++ b/tests/ha/drbd_passive.pm
@@ -44,7 +44,11 @@ sub run {
 
     # At this time, we only test DRBD on a 2 nodes cluster
     # And if the cluster has more than 2 nodes, we only use the first 2 nodes
-    return if (!is_node(1) && !is_node(2));
+    if ((!is_node(1) && !is_node(2)) || check_var('TWO_NODES', 'no')) {
+        write_tag('skip_fs_test');
+        record_info 'Skipped - Scenario', 'Test skipped because this job is not running in a two nodes scenario';
+        return;
+    }
 
     # Wait until DRBD test is initialized
     barrier_wait("DRBD_INIT_$cluster_name");
--- a/tests/ha/vg.pm
+++ b/tests/ha/vg.pm
@@ -26,20 +26,21 @@ sub run {
     my $vg_luns = undef;
 
     # This test can be called multiple time
-    if (read_tag eq 'cluster_md') {
+    my $tag = read_tag;
+    if ($tag eq 'cluster_md') {
         $resource = 'cluster_md';
         $vg_luns = '/dev/md*' if is_node(1);
 
         # Use a named RAID in SLE15
         $vg_luns = "/dev/md/$resource" if (is_sle('15+') && is_node(1));
     }
-    elsif (read_tag eq 'drbd_passive') {
+    elsif ($tag eq 'drbd_passive') {
         $resource = 'drbd_passive';
         $vg_luns = "/dev/$resource" if is_node(1);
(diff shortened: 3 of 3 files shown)


emiura:
"One possible solution would be to create a different schedule for the 3 cluster node scenario, but this implies a PR to this repo, as well as a change to the job settings."

emiura:
One possible solution would be to create a different schedule for the 3 cluster node scenario, but this implies a PR to this repo, as well as a change to the job settings. - don't we need a PR to add the "TWO_NODES=no" anyway? Anyway, if it passes your vr, lgtm.

alvarocarvajald:
> One possible solution would be to create a different schedule for the 3 cluster node scenario, but this implies a PR to this repo, as well as a change to the job settings. - don't we need a PR to add the "TWO_NODES=no" anyway? Anyway, if it passes your vr, lgtm.

It is already there:

- https://gitlab.suse.de/qa-css/openqa_ha_sap/-/blob/master/openQA/yaml_job_groups/maintenance/maintenance_sle15_sp4_ha.yml#L176
- https://gitlab.suse.de/qa-css/openqa_ha_sap/-/blob/master/openQA/yaml_job_groups/maintenance/maintenance_sle15_sp3_ha.yml#L176

If memory serves, **TWO_NODE=no** is used by other test modules and added quite a while ago. I am just simply re-using the same setting for this particular test module.

In fact, I think L47 on `tests/ha/drbd_passive.pm` which reads:
```
    if ((!is_node(1) && !is_node(2)) || check_var('TWO_NODES', 'no')) {
```
Could be re-written safely as only:
```
    if (check_var('TWO_NODES', 'no')) {
```
But I am leaving the node checks just in case there is already a working test using `ha/drbd_passive` with more than 2 nodes and that has not set **TWO_NODES=no**.

tests/ha/filesystem.pm
```diff
@@ -28,18 +29,22 @@ sub run {
...
         $resource = 'drbd_passive';
         $fs_lun = '/dev/drbd_passive' if is_node(1);
         $fs_type = 'xfs';
         $fs_opts = '-f';
     }
-    elsif (read_tag eq 'drbd_active') {
+    elsif ($tag eq 'drbd_active') {
         $resource = 'drbd_active';
     }
+    elsif ($tag eq 'skip_fs_test') {
```
dzedro, line 44: Isn't this duplicate of line 21 ?
And if it should be at begging because tag drbd_passive would be evaluated before this skip thus not skip.

alvarocarvajald: > Isn't this duplicate of line 21 ?

Not exactly the same. L21 would skip the test when _not_ on drbd Maintenance Updates, but continue it if testing a drbd MU, such as in https://openqa.suse.de/tests/9247409#step/drbd_passive/5

L44 on the other hand is skipping the test on scenarios where the test would not work (such as in 3 node) per L47-L51 in `ha/drbd_passive`.

> And if it should be at begging because tag drbd_passive would be evaluated before this skip thus not skip.

Not a problem. $tag would be one or the other, and never both: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/hacluster.pm#L458-L480

(This  probably could've been implemented with `get_var()` and `set_var()` instead, but no idea why Loic went with a local file in the cluster nodes for these tags. Could be a candidate for a re-write in the future)

dzedro: But instead of two places where the test will return there could be just one at the beginning of the if statement. Otherwise fine.

alvarocarvajald: Ah, got it. Let me move both exit conditions to the same line and test.

alvarocarvajald: New verification runs:

- QAM jobs (3 nodes): [node1](http://mango.qa.suse.de/tests/4886), [node 2](http://mango.qa.suse.de/tests/4887), [node 3](http://mango.qa.suse.de/tests/4888) & [support server](http://mango.qa.suse.de/tests/4885)
- QAM jobs (2 nodes): [node 1](http://mango.qa.suse.de/tests/4882), [node 2](http://mango.qa.suse.de/tests/4883), [client](http://mango.qa.suse.de/tests/4884) & [support server](http://mango.qa.suse.de/tests/4881)
- Alpha Cluster on Milestone Build Validation: [node 1](http://mango.qa.suse.de/tests/4879), [node 2](http://mango.qa.suse.de/tests/4880) & [support server](http://mango.qa.suse.de/tests/4878)

dzedro: Merge this PR, you are writing `skip_fs_test` tag in first commit

PR #15325, opened 2022-08-05
Fix timeout issue when starting nfs_server

rfan1:
- Related ticket: https://progress.opensuse.org/issues/114854
- Needles: n/a
- Verification run: http://openqa.suse.de/tests/9261717

### DIFF
--- a/lib/nfs_common.pm
+++ b/lib/nfs_common.pm
@@ -248,7 +248,6 @@ sub start_service {
     clear_console;
 
     # Server is up and running, client can use it now!
-    script_run "( journalctl -fu nfs-server -o short-precise > /dev/$serialdev & )";
     check_nfs_ready($rw, $ro);
 }
 
--- a/tests/console/yast2_nfs_server.pm
+++ b/tests/console/yast2_nfs_server.pm
@@ -74,6 +74,8 @@ sub run {
     assert_script_run 'umount /mnt';
 
     wait_for_children;
+
+    $self->save_and_upload_log("journalctl --no-pager -u nfs-server -o short-precise", "journal_nfs_server.log");
 }
 
 1;


pdostal (approved):
🚀

lib/nfs_common.pm
```diff
@@ -247,8 +247,14 @@ sub start_service {
     # Back on the console
     clear_console;
 
+    # Duo to poo#114875, the below journalctl command may take several time to
+    # get return code in serial terminal on low performance setups. so enhance
+    # the logic not to check the return code and add some sleep here to make
+    # the test more stable.
+    enter_cmd "( journalctl -fu nfs-server -o short-precise > /dev/$serialdev & )";
+    sleep 3;
```
foursixnine, line 255: You can increase the timeout if you want to still get the journal log, and not have it failing due to slow performance, but let's not add a `sleep` unless it's absolutely necessary

```suggestion
    enter_cmd "( journalctl -fu nfs-server -o short-precise > /dev/$serialdev )";
```

foursixnine (changes requested):

lib/nfs_common.pm
rfan1: Thanks @foursixnine, As per my test result `sleep 3` on aarch64 platform should be added here, since it took some time to flush the output to serial terminal. 

As I mentioned in the ticket 114854, Even I increased the script_run timeout to 90s, I can't fix the issue 

`( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ); echo qBERp-$?-
zypper -n in mariadb | cat; ( exit ${PIPESTATUS[0]} ); echo 7UoWa-$?-# ( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ); echo qBERp-$?-

qBERp-0-

I can suppose that serial is busy there and ( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ) may take some time to get return.

Let me try to use "enter_cmd" instead of "script_run"`

Later, `enter_cmd` can get better result.  however without `sleep 3`, similar issue can still be seen.  

Let me try to add `sleep 3` on aarch64 platform only and see

rfan1: Thanks @foursixnine, As per my test result `sleep 3` on aarch64 platform should be added here, since it took some time to flush the output to serial terminal. 

As I mentioned in the ticket 114854, Even I increased the script_run timeout to 90s, I can't fix the issue 

`( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ); echo qBERp-$?-
zypper -n in mariadb | cat; ( exit ${PIPESTATUS[0]} ); echo 7UoWa-$?-# ( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ); echo qBERp-$?-

qBERp-0-

I can suppose that serial is busy there and ( journalctl -fu nfs-server -o short-precise > /dev/ttyAMA0 & ) may take some time to get return.

Let me try to use "enter_cmd" instead of "script_run"`

Later, `enter_cmd` can get better result.  however without `sleep 3`, similar issue can still be seen.  

Let me try to add `sleep 3` on aarch64 platform only and see

rfan1: @foursixnine http://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=rfan_nfs_server_entercmd&test=qam-minima_%2Bbase_RFAN
Test passed on all platforms.

BTW, `enter_cmd "( journalctl -fu nfs-server -o short-precise > /dev/$serialdev )";` seems not work since it can't get return code if we don't run it in for-ground.

rfan1:
> @foursixnine is right. Let's fix it properly please. You may find handy `utils::script_retry()`

Good point, I wanted to use this function as well, let me try

lib/nfs_common.pm
```diff
@@ -247,8 +247,13 @@ sub start_service {
     # Back on the console
     clear_console;
 
+    # Duo to poo#114875, the below journalctl command may take several time to
+    # get return code in serial terminal on low performance setups. so enhance
+    # the logic not to check the return code and add some sleep here to make
+    # the test more stable.
+    background_script_run "( journalctl -fu nfs-server -o short-precise > /dev/$serialdev )";
+
```
foursixnine, line 255: I'm starting to think that this is meant to only get some logs, and nothing is actually depending on this call... (i.e, there's no wait_serial  that is looking for any string that might come out of the journal)... Perhaps we can simply remove it, and use `save_upload_logs` instead?

rfan1: I can't agree with you any more! ：）
Delete this line will impact the function IMO, only one concern is `short-precise` may give more detail logs there.
@pdostal, what is your sugestion?

rfan1:
> nfs-server

https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&test=qam-minima_base_RFAN

> > @foursixnine is right. Let's fix it properly please. You may find handy `utils::script_retry()`
> 
> Good point, I wanted to use this function as well, let me try [but it seems not easy to fix the performance issue unless we increase the timeout value to huge value :)] BTW, let me try with `background_script_run`

https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&test=qam-minima_base_RFAN

lib/nfs_common.pm
rfan1: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&test=qam-minima_base_RFAN

rfan1: @foursixnine @pdostal 
https://openqa.suse.de/tests/9291896#

tests/console/yast2_nfs_server.pm
```diff
@@ -19,6 +19,7 @@
 use strict;
 use warnings;
 
+use base "opensusebasetest";
```
pdostal, line 22: This is duplicity.

pdostal (approved):

rfan1:
@foursixnine can you please help review again?

foursixnine (approved):

PR #15326, opened 2022-08-05
Add new test for validation of ZFCP multipath configuration

ilmanzo:
After [#poo105440](https://progress.opensuse.org/issues/105440) that enabled a new multipath topology on S390x, this test validates the updated infrastructure

- Related ticket: https://progress.opensuse.org/issues/105449
- VR: https://openqa.suse.de/tests/9315250#step/validate_zfcp_multipath/7

### DIFF
--- a/schedule/yast/zfcp.yaml
+++ b/schedule/yast/zfcp.yaml
@@ -38,6 +38,7 @@ schedule:
   - installation/handle_reboot
   - installation/first_boot
   - console/validate_zfcp
+  - console/validate_zfcp_multipath
     #  Unscheduling due to poo#99267
     #  - console/validate_multipath
 test_data:
--- a/tests/console/validate_zfcp_multipath.pm
+++ b/tests/console/validate_zfcp_multipath.pm
@@ -1,0 +1,60 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Test module to validate zfcp multipath on z/VM testing infrastructure.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "opensusebasetest";
+use strict;
+use warnings;
+use testapi;
+use Test::Assert ':all';
+
+# Verify that there are two Host Bus adapters Attached (HBA), and that the corresponding channels are listed
+sub verify_host_bus_adapters_attached {
+    my @expected_channels = @_;
+    my $actual_channels = script_output "ls -ld /sys/class/fc_host/*";
+    # number of lines in the output should be equal to number of channels
+    assert_equals scalar(@expected_channels), scalar(split /\n/, $actual_channels), "Invalid number of channels";
+    for my $expected (@expected_channels) {
+        assert_matches qr/$expected/, $actual_channels, "Channel not found: $expected";
+    }
+}
+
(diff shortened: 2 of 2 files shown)


jknphy:
the goal is not validate the infrastructure but to know if the installer handle properly zfcp+multipath, it might be ok to do this early validation during installation but the validation should go in the installed system IMO or I am missing something...?

ilmanzo:
> 

as the installation does not make any use of multipath block devices, and given there's already the ZFCP validation in place, we can surely move this test after the reboot step ...

tests/installation/validate_zfcp_multipath.pm
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+#verify that there are 8 SCSI block devices listed (as a result of having two adapters, two hard disks and two paths to each disk, 2^3 = 8)
+#verify that LUN 0x4001403200000000 corresponds to the 214GB disk, and LUN 0x4001405000000000 to the 42.9GB disk
+sub verify_plain_scsi_block_devices() {
+    my @all_scsi_list = split /\n/, script_output "lsscsi -xxgst";
+    # must have a total of 8 devices listed
+    assert_equals 8, @all_scsi_list, "Expected 8 SCSI block devices";
+    assert_number_scsi_block_devices(4, "0x4001405000000000", "42.9GB", @all_scsi_list);
```
jknphy, line 40: you don't need an extra function to do this, as there are few items is efficient to traverse the list twice and also more maintainable. Please check this snippet:
```my @all_scsi_list = split /\n/, $line;
my @small_disk = grep {/0x4001405000000000.*42.9GB/} @all_scsi_list;
my @big_disk = grep {/0x4001403200000000.*214GB/} @all_scsi_list; 
print scalar @small_disk . "\n";
print scalar @big_disk . "\n";
 ```

File: tests/installation/validate_zfcp_multipath.pm [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+    assert_number_scsi_block_devices(4, "0x4001403200000000", "214GB", @all_scsi_list);
+}
+
+#verify that the multipath device with WWID 36005076307ffd3b30000000000000132 corresponds to the 200G disk and the multipath device with WWID 36005076307ffd3b30000000000000150 to the 40G disk
+#verify that there are 4 block devices listed for each virtual multipath device, and those corresponds to the right LUN (1077035009 and 1079001089)
+sub verify_multipath_block_devices() {
+    my @mpath = split /\n/, script_output "multipath -l";
+    assert_wwid_size_lun("36005076307ffd3b30000000000000150", "size=40G", "1079001089", @mpath);
```
jknphy, line 48: at this point the most simple option is to (1) divide in two blocks with regex or grep or other mechanism the output, (2.1) check the size for each part, (2.2) check that 4 devices with that number are added and finally (2.3) compare those 4 devices (sda..sdg) for first one and (sdb...sdf) for second one are the ones expected for the right lun.

Point 2.3 is the one I don't have so clear, it is always fixed sda to disk 200G?
But IMO it is hard to maintain the code you created in `assert_wwid_size_lun` I would recommend to play a bit more with regexp and grep/map mechanism in Perl and perhaps clarify 2.3 with @ge0r If everything is constant we could declare some hash structure with all the information related or use test data, probably better the later.

jknphy:
LGTM, good job, looks very clear to dissect problems now.
waiting for VR succeded to merge (please, set to WIP if you want me to stop the merge due to you are performing some late minute changes or something, and remove the WIP when you are finished)

PR #15327, opened 2022-08-05
tests/containers/docker_compose: Install docker-compose-switch on TW

Vogtinator:
The docker-compose binary is offered by a different package now.

- Related ticket: https://progress.opensuse.org/issues/114986
- Verification run: https://openqa.opensuse.org/tests/2498281

### DIFF
--- a/tests/containers/docker_compose.pm
+++ b/tests/containers/docker_compose.pm
@@ -21,7 +21,7 @@ use Mojo::Base 'containers::basetest';
 use testapi;
 use registration;
 use utils;
-use version_utils qw(is_sle);
+use version_utils qw(is_leap is_sle);
 use containers::common;
 use publiccloud::utils 'is_ondemand';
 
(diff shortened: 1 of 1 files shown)


DimStar77:
Short term the right thing - longer term we might want to test v2 directly (without translation glue v1 -> v2)

DimStar77 (approved):

tests/containers/docker_compose.pm
```diff
@@ -37,7 +37,13 @@ sub run {
 
     record_info 'Test #1', 'Test: Installation';
 
-    my $ret = zypper_call "in docker-compose", exitcode => [0, 4];
+    my $ret;
+    if (is_sle || is_leap) {
```
jlausuch, line 41: maybe to avoid potential errors in other platforms, I would have done 
`if is_tumbleweed` -> install switch.
`else`,  leave default.

Vogtinator: `is_tunbleweed` is something that should be avoided (see the code style policy) and IMO it does make sense here. At some point (15.5?) Leap and SLE will also get the change, then you just need to add a version number.

PR #15334, opened 2022-08-08
Workaround lost windows focus in GNOME

mloviska:
Bug report: https://bugzilla.opensuse.org/show_bug.cgi?id=1202103

Opening windows from desktop runner might open windows without a focus.
That causes typing failures in `xterm` for instance.

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -298,7 +298,16 @@ sub x11_start_program {
     my @target = ref $args{target_match} eq 'ARRAY' ? @{$args{target_match}} : $args{target_match};
     for (1 .. 3) {
         push @target, check_var('DESKTOP', 'kde') ? 'desktop-runner-plasma-suggestions' : 'desktop-runner-border';
-        assert_screen([@target], $args{match_timeout}, no_wait => $args{match_no_wait});
+        my $match = assert_screen([@target], $args{match_timeout}, no_wait => $args{match_no_wait});
+        # workaround for lost focus in GNOME
+        # bsc#1202103 - [Build 20220803-2] GNOME: lost window focus
+        if ($program eq 'xterm') {
+            foreach my $property (@{$match->{needle}->{properties}}) {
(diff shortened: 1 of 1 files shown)


lib/susedistribution.pm
```diff
@@ -298,7 +298,16 @@ sub x11_start_program {
     my @target = ref $args{target_match} eq 'ARRAY' ? @{$args{target_match}} : $args{target_match};
     for (1 .. 3) {
         push @target, check_var('DESKTOP', 'kde') ? 'desktop-runner-plasma-suggestions' : 'desktop-runner-border';
-        assert_screen([@target], $args{match_timeout}, no_wait => $args{match_no_wait});
+        my $match = assert_screen([@target], $args{match_timeout}, no_wait => $args{match_no_wait});
+        # workaround for lost focus in GNOME
+        # bsc#1202103 - [Build 20220803-2] GNOME: lost window focus
+        if ($program eq 'xterm') {
+            foreach my $property (@{$match->{needle}->{properties}}) {
```
mloviska, line 305: Is it possible to use `$match->has_property('workaround')` here? I would need to serialize the object first?

okurz: Sorry, I don't know that. At best we can avoid this. How about

```
            click_lastmatch() if match_has_tag('unfocussed');
```

As you can add multiple tags to needles. This way you also don't need the unusual handling of the assert_screen return value

mloviska: Yup, that is another solution, adding new elements (`xterm-unfocussed`) into `@target`. I do not really have any preference over here, but it seems fine just to add a workaround property for GNOME's xterm needle.

okurz: Well, if you don't have a preference may I please ask you to implement my suggestion as that relies less on internals but only the stable API and should therefore be more reliable

okurz (changes requested):
See latest suggestions

PR #15338, opened 2022-08-09
Enable container module for ALP

pdostal:
- Related ticket: [poo#114935](https://progress.opensuse.org/issues/114935)
- Related pull request os-autoinst/opensuse-jobgroups#185
- Related fix: thkukuk/hello-kubic#1
- Verification run: [pdostal-server.suse.cz/t449](https://pdostal-server.suse.cz/tests/452)

### DIFF
--- a/lib/containers/common.pm
+++ b/lib/containers/common.pm
@@ -254,7 +254,7 @@ sub test_container_image {
     # Images from custom registry are listed with the '$registry/library/'
     $image =~ s/^docker\.io\/library\///;
 
-    my $smoketest = qq[/bin/sh -c '/bin/uname -r; /bin/echo "Heartbeat from $image";ps'];
+    my $smoketest = qq[/bin/sh -c '/bin/uname -r; /bin/echo "Heartbeat from $image"'];
 
     $runtime->pull($image, timeout => 420);
     $runtime->check_image_in_host($image);
--- a/data/containers/Dockerfile
+++ b/data/containers/Dockerfile
@@ -3,6 +3,7 @@
 FROM baseimage_var
 
 RUN zypper -n in apache2 && zypper clean -a
+RUN echo 'ServerName AppacheInContainer' >> /etc/apache2/httpd.conf
 
 # set a directory for the app
 WORKDIR /srv/www/htdocs/
--- a/data/containers/www.py
+++ b/data/containers/www.py
@@ -19,6 +19,8 @@ class TestPassRequestHandler(BaseHTTPRequestHandler):
         self.send_response(200)
         self.end_headers()
         self.wfile.write(CONTENT.encode("UTF-8"))
+    def log_request(self, format, *args):
+        return
 
 
 # Main program routine: Setup and run the webserver
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -187,7 +187,7 @@ sub basic_container_tests {
     assert_script_run("$runtime container commit $container_name tw:saved", 240);
 
     # Network is working inside of the containers
-    my $output = script_output("$runtime container run tw:saved curl -I google.de");
+    my $output = script_output("$runtime container run tw:saved curl -sI google.de");
     die("network is not working inside of the container tw:saved") unless ($output =~ m{Location: http://www\.google\.de/});
 
     # Using an init process as PID 1
--- a/lib/main_containers.pm
+++ b/lib/main_containers.pm
@@ -86,13 +86,13 @@ sub load_host_tests_podman {
     unless (is_sle("<15-sp1")) {
         load_container_engine_test($run_args);
         # In Public Cloud we don't have internal resources
-        load_image_test($run_args) unless is_public_cloud;
+        load_image_test($run_args) unless is_public_cloud || is_alp;
         load_3rd_party_image_test($run_args);
         loadtest 'containers/podman_pods';
         # Firewall is not installed in JeOS OpenStack, MicroOS and Public Cloud images
-        loadtest 'containers/podman_firewall' unless (is_public_cloud || is_openstack || is_microos);
+        loadtest 'containers/podman_firewall' unless (is_public_cloud || is_openstack || is_microos || is_alp);
         # Buildah is not available in SLE Micro and MicroOS
-        loadtest 'containers/buildah' unless (is_sle_micro || is_microos || is_leap_micro);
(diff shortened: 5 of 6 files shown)


lib/containers/common.pm
```diff
@@ -254,7 +254,7 @@ sub test_container_image {
     # Images from custom registry are listed with the '$registry/library/'
     $image =~ s/^docker\.io\/library\///;
 
-    my $smoketest = qq[/bin/sh -c '/bin/uname -r; /bin/echo "Heartbeat from $image";ps'];
+    my $smoketest = qq[/bin/sh -c '/bin/uname -r; /bin/echo "Heartbeat from $image"'];
```
grisu48, line 257: Why remove the `ps` command?

data/journal_check/bug_refs.json
```diff
@@ -388,5 +388,26 @@
...
+        },
+        "type": "ignore"
+    },
+    "containers/hello-kubic-pod-kubic-started": {
+        "description": "hello-kubic-pod.*Hello-Kubic.*started",
+        "products": {
+            "alp":  ["0.1"]
+        },
+        "type": "ignore"
     }
```
mloviska, line 412: Are all of these messages warnings or even errors ?!

lib/containers/common.pm
mloviska: IIRC, the `ps` call was added there to double-check that the whole chain of commands is passed correctly into the container. Not really necessary, but nice to have from my point of view and mainly the call was buggy before so it was an easy review check.

pdostal: Because of an error in `journal_check.pm`:
```
Aug 08 15:59:40.737886 localhost.localdomain testing[4827]: /bin/sh: 1: ps: not found
```

data/journal_check/bug_refs.json
pdostal: No, they are `info` messages from containers:
```
Aug 08 15:58:51.733399 localhost.localdomain myapp[3562]: 10.88.0.1 - - [08/Aug/2022 15:58:51] "GET / HTTP/1.1" 200 -
```
and
```
Aug 08 16:00:57.878292 localhost.localdomain hello-kubic-pod-0-hello-kubic[7137]: 2022/08/08 16:00:57 Hello-Kubic 1.4 started
```
and
```
Aug 08 16:04:14.817806 localhost.localdomain myapp[9633]: AH00558: httpd-prefork: Could not reliably determine the server's fully qualified domain name, using 10.0.2.100. Set the 'ServerName' directive globally to suppress this message
```

lib/containers/common.pm
grisu48: Since @mloviska stated that it's not necessary anymore, I'm OK with removing it. Thanks for the clarification! :+1:

grisu48 (approved):
LGTM!

data/journal_check/bug_refs.json
mloviska: That is weird, as the `journalctl --no-pager --quiet -p err -o short-precise` should print errors or higher messages. I guess that should be fixed somewhere if they are info messages.

pdostal: Hmmm, yes, that's weird but where? They are all from podman container so it will be probably related. But how? … :(

pdostal: @dcermak can you take a look please?

dcermak: @pdostal I really need some context here, I must confess that I have my troubles finding out what the issue here is.

pdostal: @dcermak we are checking journal for error logs and some messages from podman appear there. But it does not make any sense to me why `Hello-Kubic 1.4 started` is treater as error message.

pdostal: It seems to me like some containers are outputting everything to stderr.

pdostal:
```bash
# podman run registry.opensuse.org/opensuse/leap:latest sh -c 'echo test345-OUT > /dev/stdout'
test345-OUT
# podman run registry.opensuse.org/opensuse/leap:latest sh -c 'echo test345-ERR > /dev/stderr'
test345-ERR
# journalctl --no-pager -b --quiet -p err -o short-precise | grep 'test345'
Aug 09 19:16:56.342123 podman pensive_euclid[5017]: test345-ERR
# journalctl --no-pager -b --quiet -o short-precise | grep 'test345'
Aug 09 19:16:47.501007 podman wonderful_bardeen[4823]: test345-OUT
Aug 09 19:16:56.342123 podman pensive_euclid[5017]: test345-ERR
```

pdostal: So the HTTP 200 entry is from our testing Python app and it's printed by `http.server` to stderr by default:
```
$ cat www.py
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
# Summary: Simple http server replying a static page with CONTENT
#          By default it listens on *:80

from http.server import HTTPServer, BaseHTTPRequestHandler
import os

# Content of the static http page
CONTENT = "<html>The test shall pass</html>"


class TestPassRequestHandler(BaseHTTPRequestHandler):
    """
    Actual webserver request handler. It returns CONTENT on a GET request
    """

    def do_GET(self):
        self.send_response(200)
        self.end_headers()
        self.wfile.write(CONTENT.encode("UTF-8"))
    #def log_request(self, format, *args):
    #    return


# Main program routine: Setup and run the webserver
if __name__ == "__main__":
    print("TEST\n")
    # If defined, take the PORT and ADDR environment variables
    port, addr = int(os.environ.get("PORT", 8080)), os.environ.get("ADDR", "0.0.0.0")
    print("http serving on %s:%d" % (addr, port))
    httpd = HTTPServer((addr, port), TestPassRequestHandler)
    httpd.serve_forever()
pavel@pdostal-server:/data/openqa/tests/opensuse/data/containers alp_containers_enable'$ python3 www.py 1> /dev/null
127.0.0.1 - - [10/Aug/2022 08:17:01] "GET / HTTP/1.1" 200 -
```

I now think that there is no error in journal, neither in podman or conmon.
I will try to fix those messages one by one so the `bug_refs.json` will remain unchanged.

pdostal: I opened thkukuk/hello-kubic#1 for the `hello-kubic` stderr message.

mloviska (approved):

asdil12 (approved):

PR #15339, opened 2022-08-09
update_kernel: Check for retracted kernel packages

mdoucha:
Retracted kernels should not receive any new livepatches. However, if a livepatch incident get accidentally created for a retracted kernel, installation test job will fail with rather ambiguous message that "patch is not needed". Check for retracted kernels during livepatch installation and fail with a clear error message.

- Related ticket: N/A
- Needles: N/A
- Verification run:
  - SLE-12SP4 (success): https://openqa.suse.de/tests/9293335
  - SLE-15SP3 (success): https://openqa.suse.de/tests/9293334
  - unfortunately, there are currently no open incidents for retracted kernels to test the failure code path

### DIFF
--- a/tests/kernel/update_kernel.pm
+++ b/tests/kernel/update_kernel.pm
@@ -261,10 +261,8 @@ sub prepare_kgraft {
 
     fully_patch_system;
 
-    my $verlist = zypper_search(q(-s -x kernel-default));
-    my $kernel_version = right_kversion($verlist, $incident_klp_pkg);
-    $verlist = zypper_search(q(-s -x kernel-source));
-    my $src_version = right_kversion($verlist, $incident_klp_pkg);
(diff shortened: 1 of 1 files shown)


mdoucha:
SLE-12SP4 verification run failed because the livepatch was already released but the relevant part of the test module finished successfully.

pevik (approved):
LGTM. Could you please add this description from PR also to the commit message? (we never know if we stay with github forever)

b10n1k (approved):
+1

mdoucha:
> LGTM. Could you please add this description from PR also to the commit message? (we never know if we stay with github forever)

What I wrote to the PR description should be obvious to anyone who needs to look at update_kernel.pm in the future. I need to explain the problem to PR reviewers who may know nothing about livepatch testing but other than that, the code is straightforward enough.

PR #15342, opened 2022-08-09
Add workaround for 12SP5 livepatch installation issue

mdoucha:
SLE-12SP5 livepatches for kernels without Retbleed mitigations fail to install due to dependency issue with crash-kmp-default package. Install an older version of that package to work around the issue.

- Related ticket: N/A
- Needles: N/A
- Verification run:
  - 12-SP4: https://openqa.suse.de/tests/9293920
  - 12-SP5: https://openqa.suse.de/tests/9293896
  - 15GA: https://openqa.suse.de/tests/9293921
  - 15-SP4: https://openqa.suse.de/tests/9293922

### DIFF
--- a/tests/kernel/update_kernel.pm
+++ b/tests/kernel/update_kernel.pm
@@ -221,6 +221,10 @@ sub install_lock_kernel {
         $package .= '-' . ($packver{$package} // $kernel_version);
     }
 
+    # Workaround for kgraft installation issue due to Retbleed mitigations
+    push @packages, 'crash-kmp-default-7.2.1_k4.12.14_122.124'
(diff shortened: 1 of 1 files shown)


b10n1k (commented):
would be helpful to add a record_info or softfail as a reminder of the workaround?

mdoucha:
Reminder is not necessary. The workaround will be needed until September 2023 but it's not a problem if it stays there until 12SP5 goes out of support.

12SP4 verification run failed as expected due to the incident being already released but the relevant part of the test code passed in all verification runs. This PR is ready for merge.

PR #14342, opened 2022-02-24
Resize as workaround for YaST content not loading

ge0r:
Due to bsc#1191112, sometimes YaST module screens do not load their content.
A workaround to this is to maximize and unmaximize the YaST window via keybinds, in order to force refresh the content.

- Related ticket: https://progress.opensuse.org/issues/105046
- Needles: No needles
- Verification runs: 
[yast2_security before](https://openqa.suse.de/tests/7976962#step/yast2_security/16) | [yast2_security after](https://openqa.suse.de/tests/8218453#step/yast2_security/16)
[addon_products_via_SCC_yast2 before](https://openqa.suse.de/tests/8208270#step/addon_products_via_SCC_yast2/21) | [addon_products_via_SCC_yast2 after](https://openqa.suse.de/tests/8218090#step/addon_products_via_SCC_yast2/21)
[yast2_hostnames before](https://openqa.suse.de/tests/8250026#step/yast2_hostnames/12) | [yast2_hostnames after](https://openqa.suse.de/tests/8250339#)

also all three `iscsi` multimachine jobs:
[iscsi_client_normal_auth_backstore_fileio](https://openqa.suse.de/tests/8222152)
[iscsi_client_normal_auth_backstore_hdd](https://openqa.suse.de/tests/8222511)
[iscsi_client_normal_auth_backstore_lvm](https://openqa.suse.de/tests/8223121)

PR #15343, opened 2022-08-10
Apply workaround for yast content not being loaded for MU sp4

skliu:
Apply workaround for yast content not being loaded for MU sp4

- Related ticket: https://progress.opensuse.org/issues/112577
- Needles: -
- Verification run: http://d218.qam.suse.de/tests/156

### DIFF
--- a/tests/update/zypper_ref_auto_import_keys.pm
+++ b/tests/update/zypper_ref_auto_import_keys.pm
@@ -1,0 +1,22 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: zypper
+# Summary: auto import gpg keys
+# Auto import gpg keys, useful when we re-launch the test suite after the gpg key for maintenance repositories have expired (eg next day after initial run).
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_consoletest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub run {
+    select_console 'root-console';
+    zypper_call('--gpg-auto-import-keys ref');
+}
+
+1;
--- a/tests/yast2_gui/yast2_bootloader.pm
+++ b/tests/yast2_gui/yast2_bootloader.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright 2009-2013 Bernhard M. Wiedemann
-# Copyright 2012-2021 SUSE LLC
+# Copyright 2012-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: yast2-bootloader
@@ -45,8 +45,10 @@ sub run {
     type_string '16';
 
     #	default boot section
+    for (1 .. 2) { send_key 'alt-f10' }
     assert_and_click 'yast2-bootloader_default-boot-section';
-    assert_screen 'yast2-bootloader_default-boot-section_tw';
+    record_soft_failure('bsc#1191112 - Resizing window as workaround for YaST content not loading');
+    send_key_until_needlematch('yast2-bootloader_default-boot-section_tw', 'alt-f10', 9, 2);
     send_key 'esc';    # Close drop down
 
     #	proctect boot loader with password
--- a/schedule/yast/maintenance/mau_extratests_yast2ui_restapi.yaml
+++ b/schedule/yast/maintenance/mau_extratests_yast2ui_restapi.yaml
@@ -10,6 +10,7 @@ schedule:
   - console/system_prepare
   - console/prepare_test_data
   - console/consoletest_setup
+  - update/zypper_ref_auto_import_keys
   - yast2_gui/yast2_bootloader
   - yast2_gui/yast2_security
   - yast2_gui/yast2_keyboard
--- a/tests/yast2_gui/yast2_instserver.pm
+++ b/tests/yast2_gui/yast2_instserver.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: nfs-client yast2-instserver lftp xinetd vsftpd openslp-server yast2-nfs-server nfs-client apache2
@@ -110,7 +110,8 @@ sub test_http_instserver {
     wait_still_screen 2, 2;
     send_key_and_wait("alt-n", 2);
     send_key_and_wait("alt-a", 2);
-    assert_screen('yast2-instserver-repository-conf');
+    record_soft_failure('bsc#1191112 - Resizing window as workaround for YaST content not loading');
+    send_key_until_needlematch('yast2-instserver-repository-conf', 'alt-f10', 9, 2);
     send_key_and_wait("alt-p", 2);
     type_string "instserver";
     wait_still_screen 2, 2;
@@ -121,7 +122,7 @@ sub test_http_instserver {
     send_key_until_needlematch("yast2-instserver_sr0dev", "down", 3);
     send_key_and_wait("alt-n", 2);
     send_key_and_wait("alt-o", 2);
-    assert_screen([qw(yast2-instserver-ui yast2-instserver-change-media)], 300);
+    send_key_until_needlematch([qw(yast2-instserver-ui yast2-instserver-change-media)], 'alt-f10', 21, 30);
     # skip "insert next cd" on SLE 12.x
     send_key_and_wait("alt-s", 2) if is_sle("<=12-SP5") && match_has_tag('yast2-instserver-change-media');
     assert_screen('yast2-instserver-ui');
@@ -168,5 +169,5 @@ sub run {
     clean_env;
 
 }
-1;
 
+1;


tests/yast2_gui/yast2_bootloader.pm
```diff
@@ -38,15 +38,23 @@ sub run {
 
     #	bootloader options and set probe foreign OS, timeout
     assert_and_click 'yast2-bootloader_bootloader-options';
+
+    wait_still_screen 10;
+    send_key 'alt-t';    # Import untrusted GnuPG key -- workaround
```
jknphy, line 43: not needed if we change the priority of the child job or if that doesn't work creating a new test module to zypper ref with the param to auto accept gpg keys.

tests/yast2_gui/yast2_bootloader.pm
```diff
@@ -38,15 +38,23 @@ sub run {
...
+    wait_still_screen 10;
+
     assert_screen 'yast2-bootloader_bootloader-options-switched';
     send_key 'alt-b';
     wait_still_screen 3;
     send_key 'alt-t';
     type_string '16';
 
     #	default boot section
+    for (1 .. 2) { send_key 'alt-f10' }
```
jknphy, line 53: I don't think this loop is needed, I haven't seen in previous occurrences of the workaround, could you check please?

tests/yast2_gui/yast2_bootloader.pm
```diff
@@ -38,15 +38,23 @@ sub run {
...
     assert_and_click 'yast2-bootloader_default-boot-section';
-    assert_screen 'yast2-bootloader_default-boot-section_tw';
+    record_soft_failure('bsc#1191112 - Resizing window as workaround for YaST content not loading');
+    send_key_until_needlematch('yast2-bootloader_default-boot-section_tw', 'alt-f10', 9, 2);
+    #assert_screen 'yast2-bootloader_default-boot-section_tw';
```
jknphy, line 57: unintended commented code.

tests/yast2_gui/yast2_bootloader.pm
```diff
@@ -38,15 +38,23 @@ sub run {
...
+
     assert_screen 'yast2-bootloader_bootloader-options-switched';
     send_key 'alt-b';
     wait_still_screen 3;
     send_key 'alt-t';
     type_string '16';
 
     #	default boot section
+    for (1 .. 2) { send_key 'alt-f10' }
     assert_and_click 'yast2-bootloader_default-boot-section';
```
jknphy, line 54: we could do a `check_screen($tag)` to see if it is visible this area and if not run the workaround.
right now we are clicking before make it visible due to the workaround, it would make sense to only click after, right?

jknphy:
@skliu any clue about what is this [failure](http://d218.qam.suse.de/tests/100#step/yast2_instserver/21) about at the end of the VR?

tests/yast2_gui/yast2_bootloader.pm
```diff
@@ -38,15 +38,23 @@ sub run {
...
-    assert_screen 'yast2-bootloader_default-boot-section_tw';
+    record_soft_failure('bsc#1191112 - Resizing window as workaround for YaST content not loading');
+    send_key_until_needlematch('yast2-bootloader_default-boot-section_tw', 'alt-f10', 9, 2);
```
JRivrain, line 56: The problem with sending alt-f10 is that we en-up with fullscreen as here http://d218.qam.suse.de/tests/100#step/yast2_instserver/15. Not the end of the workd, but then we may need to maintain two sets of needles, one for full screen and one for windowed.

JRivrain: Then it creates random failures as we saw here https://openqa.suse.de/tests/9289934#step/iscsi_client/77. Not sure how to tackle this.

tests/yast2_gui/yast2_bootloader.pm
skliu: here it is: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14342/files#diff-e5ae008bc24ec64aac7d7f80f7671324b746ffe329c8fd6c8e5dba662f942ae7

tests/yast2_gui/yast2_bootloader.pm
skliu: Can we send alt-f10 10 times?

tests/yast2_gui/yast2_bootloader.pm
skliu: added a new module: zypper_ref

skliu:
> @skliu any clue about what is this [failure](http://d218.qam.suse.de/tests/100#step/yast2_instserver/21) about at the end of the VR?

I think it's due to the time out.

tests/yast2_gui/yast2_bootloader.pm
skliu: "it would make sense to only click after, right?"---sorry, I cannot get it. I added the  `for` loop because the window is not fully loaded, this caused the next line `assert_and_click` failed

tests/yast2_gui/yast2_bootloader.pm
jknphy: you need to pass zypper ref the parameter to accept the gpg keys, for which I would not touch this module used everywhere.

skliu: I've added `--gpg-auto-import-keys` in zypper_ref.pm

jknphy:
I see too much troubles with this workaround, now that we have libyui, we could switch to 
yast2_gui/bootloader/bootcode_options
which covers only the first tab, but we can plan for covering a bit more.
For the instserver have you investigated the error? did you file a bug for it?

tests/yast2_gui/yast2_bootloader.pm
jknphy: it is not reflected in your code changes...

skliu: Do you mean I should use `zypper_call('--gpg-auto-import-keys ref')` in the existing module?

tests/yast2_gui/yast2_bootloader.pm
ge0r: it is needed to always send an even number of `'alt-f10'` because the needle will match the unmaximized version of the yast screen, but not the maximized

tests/yast2_gui/yast2_bootloader.pm
ge0r: @jknphy it sends alt-f10 ten times, not nine, despite what the argument says, check https://progress.opensuse.org/issues/107749

tests/yast2_gui/yast2_bootloader.pm
jknphy: better create a new module `zypper_ref_auto_import_keys.pm` so we could use it regardless the order of the test modules in the test suite when this happen sporadically. But you didn't try to adjust yet the priority, I think that was the option number 1 we decided instead of this.

skliu: I changed the priority to 20.

tests/yast2_gui/yast2_bootloader.pm
jknphy: I guess we are not planning to do the workaround in a better way just to apply it, never mind :)

tests/yast2_gui/yast2_bootloader.pm
jknphy: great, thanks! please always paste the link for referenced PRs/MRs
https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/354/

tests/yast2_gui/yast2_bootloader.pm
JRivrain: Ah ok,  so I suspect this happens just before when we are using `for (1 .. 2) { send_key 'alt-f10' },` no ? I guess the two send_key are sent too fast, then the window stays full screen. I would suggest to put the same send_key_until_needlematch also there.

jknphy:
@skliu @JRivrain  @ge0r  Do we have a solid solution already for bootloader, if not, please @skliu apply `yast2_gui/bootloader/bootcode_options` instead of needle-based module. And then merge this PR.
If we don't have a clear idea for the second issue, file a ticket with current findings, looks less important that bootloader module, we can live without it for a while.

JRivrain:
> > @skliu any clue about what is this [failure](http://d218.qam.suse.de/tests/100#step/yast2_instserver/21) about at the end of the VR?
> 
> I think it's due to the time out.

I had mentioned it in https://progress.opensuse.org/issues/112589: there are more issues with this module.

tests/yast2_gui/yast2_bootloader.pm
JRivrain: I think we should also use send_key_until_needlematch('yast2-bootloader_default-boot-section_tw', 'alt-f10', 9, 2);
as otherwise we randomly end-up fullscreen (or is there some reason not to ?)

jknphy:
@JRivrain  we just need to fix the first error, correct, but we need a ticket with clear info to tackle the second error as part of the outcome of this task, that ticket that I rejected didn't have that info.

@skliu are we missing something before merging? I remembered we talked offline about creating new module to be schedule before this to do zypper ref and autoimport the keys. Are you planning to add that?

skliu:
> @JRivrain we just need to fix the first error, correct, but we need a ticket with clear info to tackle the second error as part of the outcome of this task, that ticket that I rejected didn't have that info (including links to our observations could also help for next time)
> 
> @skliu are we missing something before merging? I remembered we talked offline about creating new module to be schedule before this to do zypper ref and autoimport the keys. Are you planning to add that?

only the coping CD problem. http://d218.qam.suse.de/tests/131#step/yast2_instserver/43

tests/yast2_gui/yast2_bootloader.pm
JRivrain: > Ah ok, so I suspect this happens just before when we are using `for (1 .. 2) { send_key 'alt-f10' },` no ? I guess the two send_key are sent too fast, then the window stays full screen. I would suggest to put the same send_key_until_needlematch also there.

I see that here https://openqa.nue.suse.com/tests/9386473#step/installation/59, we en-up full screen while using send_key_until_needlematch, so I was wrong, resolving conversation.

tests/yast2_gui/yast2_bootloader.pm
JRivrain: > I think here we should also use `send_key_until_needlematch('yast2-bootloader_default-boot-section_tw', 'alt-f10', 9, 2);` as otherwise we randomly end-up fullscreen, see my other comment just below. Or is there some reason not to ?

I see that here https://openqa.nue.suse.com/tests/9386473#step/installation/59, we end-up full screen even if using send_key_until_needlematch, so I was wrong assuming it can be because of `for (1 .. 2)` , resolving conversation.

tests/yast2_gui/yast2_bootloader.pm
JRivrain: This being said, it means the code works, just those fullscreen needles should be removed, and we should create more specific ones that are not fullscreen.

tests/yast2_gui/yast2_bootloader.pm
skliu: code changed, but there is a failure. http://d218.qam.suse.de/tests/157#step/yast2_bootloader/28

tests/yast2_gui/yast2_bootloader.pm
skliu: I added send_key_until_needlematch, so I think it's fixed.

skliu: > This being said, it means the code works, just those fullscreen needles should be removed, and we should create more specific ones that are not fullscreen.

ok, I will revert the code.

JRivrain: I am saying that the code works, why do you want to revert it ?

skliu:
> For the instserver have you investigated the error? did you file a bug for it?

Yes, I did. the first error is due to the time out of coping CD contents, and the rest errors are all due to the failure of sending "alt-f", so "yast2 instserver" window is not disappeared, all the errors happens randomly. BTW: where can I file a bug? on https://bugzilla.suse.com/ ?

skliu:
> @JRivrain we just need to fix the first error, correct, but we need a ticket with clear info to tackle the second error as part of the outcome of this task, that ticket that I rejected didn't have that info (including links to our observations could also help for next time)
> 
> @skliu are we missing something before merging? I remembered we talked offline about creating new module to be schedule before this to do zypper ref and autoimport the keys. Are you planning to add that?

the module is created.

tests/update/zypper_ref_auto_import_keys.pm
```diff
@@ -0,0 +1,21 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: zypper
+# Summary: auto import gpg keys
+# Maintainer: Shukui Liu <skliu@suse.com>
```
JRivrain, line 8: Nowadays we use `# Maintainer: QE YaST <qa-sle-yast@suse.de>`, as all the team is responsible for the code. 
I would also complete the summary with more explanation: "Auto import gpg keys, useful when we re-launch the test suite after the gpg key for maintenance repositories have expired (eg next day after initial run)."

skliu: fixed.

skliu:
> I see too much troubles with this workaround, now that we have libyui, we could switch to [yast2_gui/bootloader/bootcode_options](https://openqa.suse.de/tests/8753210#step/bootcode_options/1) which covers only the first tab, but we can plan for covering a bit more. For the instserver have you investigated the error? did you file a bug for it?

I created a progress ticket for it. https://progress.opensuse.org/issues/115748

jknphy:
LGTM

tests/yast2_gui/yast2_bootloader.pm
JRivrain: Sorry, I had closed this conversation with the following comment:
_I see that here https://openqa.nue.suse.com/tests/9386473#step/installation/59, we end-up full screen even if using send_key_until_needlematch, so I was wrong assuming it can be because of for (1 .. 2) , resolving conversation._
By that I meant you did not have to change that part of the code. , when we click "resolve conversation", it means no longer action is necessary.

Now if you would want to use send_key_until_needlematch here to be more consistent, you should use this:
`record_soft_failure('bsc#1191112 - Resizing window as workaround for YaST content not loading');
send_key_until_needlematch('yast2-bootloader_bootloader-options', 'alt-f10', 9, 2)`,
Sorry for not being enough precise.

PR #15351, opened 2022-08-10
Use *_TEST_REPOS variable instead of *_TEST_{ISSUES,TEMPLATE}

mimi1vx:
- Related ticket: https://progress.opensuse.org/issues/111710
- Needles: not needed
- Verification run: http://quasar.suse.cz/tests/547

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -102,6 +102,7 @@ our @EXPORT = qw(
   load_zdup_tests
   logcurrentenv
   map_incidents_to_repo
+  join_incidents_to_repo
   need_clear_repos
   noupdatestep_is_applicable
   opensuse_welcome_applicable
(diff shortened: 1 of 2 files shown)


lib/main_common.pm
```diff
@@ -794,6 +795,24 @@ sub map_incidents_to_repo {
...
+        next unless $incidents->{$k};
+        for my $i (split(/,/, $incidents->{k})) {
+            if ($i) {
+                push @repos, $i;
+            }
+        }
+    }
+
+    my $ret = join(',', @repos);
+    $ret =~ s/^,//s;
```
perlpunk, line 812: With the current code there can't be a comma at the beginning, I would say, so this can be removed and the three lines shortened to
```perl
return join ',', @repos;
```

PR #15353, opened 2022-08-10
Containerized PC tools image update.

m-dati:
Related to the already merged PC tools update, same ticket, this PR contains the code used to create the container and update the virtualenv that produced the files with the new tools versions.

Check the procedure in the README file.

- Related ticket: https://progress.opensuse.org/issues/106658
- Needles: none
- Verification run: see in https://github.com/m-dati/os-autoinst-distri-opensuse/tree/master/data/publiccloud/venv/*.txt, the new txt files produced following the README procedure.


tools/pctools/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+a) set the directory containing those files:
+    
+`% cd /var/lib/openqa/share/tests/opensuse/tools/pctools`
+
+b) build the container using the current Dockerfile for BCI 15SP3 with python3:6 and assign a *name* for reference:
+
+`% podman build . -t <name>` 
+
+c) run the container sharing the workdir with the SHARDIR defined in the Dockerfile, without other commands, so that the CMD instruction by dafault will copy here the files with the new versions:
```
asmorodskyi, line 37:
```suggestion
c) run the container sharing the workdir with the SHARDIR defined in the Dockerfile, without other commands, so that the CMD instruction by default will copy here the files with the new versions:
```

asmorodskyi (approved):
LGTM , but I would like someone else to check it before merge

tools/pctools/README.md
```diff
@@ -0,0 +1,41 @@
+PUBLIC CLOUD TOOLS IMAGE
+========================
+
+A Public Cloud Tools image is used for specific tests for azure, amazon, openstack, ec2 and google CSPs.
```
grisu48, line 4: It is a good habit to always establish the context of the repository to allow readers without prior background information to get a grasp what's going on here, e.g.


```suggestion
The Public Cloud Tools image is a modified SLES image with a pre-installed and configured set of various cloud related tools intended for testing of Cloud Service Providers (CSP) within openQA. It include various tools including command-line interfaces for Azure, Amazon, OpenStack and Google Compute Platform (GCP). This Dockerfile helps to generate the automatically generate the requirements for those utilities.
```

It is also a courtesy to the reader to always introduce abbreviations when used for the first time (or refer to a glossary).

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
...
+# --- 
+# $1 CLI Cloud type: aws, azure, ec2, openstack
+# $2 resources output file
+# $3,... : install modules
+# ---
+
+echo $0 $*
+
+# parameters check
+[ $# -lt 3 ] && exit 1
```
grisu48, line 16: Might be a good idea to exit with an error message, WDYT?

```suggestion
[ $# -lt 3 ] && (echo "not enough arguments"; exit 1)
```

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# COPY and RUN this script in a container image, to:
+# create a python virtualenv named as $1,
```
grisu48, line 3:
```suggestion
# create a python virtualenv named as test_$1,
```

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
grisu48, line 1: Perhaps it's a good idea to fail the bash script in case of errors in the commands?

```suggestion
#!/bin/bash -e
```

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
...
+# Versions in venv
+python3 --version
+pip --version
+
+# install the modules
+shift 2 # remove 2 used param
+pip install $@
+
+# install result
+[ $? -ne 0 ] && exit 2
```
grisu48, line 38: Not needed if you use the shebang with `/bin/bash -e`

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
...
+# collect the resulting modules and versions
+pip freeze > $FILE; 
+
+#deact. venv
+deactivate
+
+# copy file in parent dir
+cp $FILE .. 
+
+cd ..
```
grisu48, line 49: This has no effect because after this the new bash is anyways terminated (bash script start in a new `bash` and don't affect the outside environment)

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
...
+# ---
+
+echo $0 $*
+
+# parameters check
+[ $# -lt 3 ] && exit 1
+
+# create venv
+CLOUD=$1
+FILE=$2
```
grisu48, line 20: Perhaps giving the resources output file a more descriptive variable name might increase the readability?

```suggestion
OUTPUT=$2
```

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# install the modules
+shift 2 # remove 2 used param
+pip install $@
+
+# install result
+[ $? -ne 0 ] && exit 2
+
+# collect the resulting modules and versions
+pip freeze > $FILE;
```
grisu48, line 41:
```suggestion
pip freeze > ../$FILE; 
```

Then you don't need to copy the file over later on, but I leave it to you if you like this or not :-)

tools/pctools/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+# this dir will be a shared volume at runtime
+ARG SHARDIR=/home/tmp
+
+# script to install
+ARG VENV=./venv_inst.sh
+
+# venv dir and pip modules to install:
+# AWS
+ARG DIRAWS=aws 
+ARG OUTAWS="resources_$DIRAWS.txt"
```
grisu48, line 16: Why is here another variable (`$DIRAWS`) being used?

This gets a bit too convoluted. We are using an `ARG`, which contains a bash variable, which is passed as an argument to a bash script ...  
Is this level of complexity really necessary? Following the flow is difficult here.

grisu48 (commented):
Good job! I've added some improvement suggestions, feel free to take them or comment on them if they are not applicable :slightly_smiling_face:

tools/pctools/venv_inst.sh
m-dati: thanks, indeed I already did so, but the main process of the command RUN resulted still ok and the build doesn't stop, because in this way the eventual error occurs in a sub-shell (...) but the parent shell is ok. 
I should convert the command using a normal 'if ... then'.

tools/pctools/venv_inst.sh
m-dati: if you mean the command `cd ..`, I agree :+1:

tools/pctools/venv_inst.sh
m-dati: I'd prefer to create the file in the specific folder dedicated to that cloud type so that in the final container you have the correct status set for each type. 
Indeed, then that copy '..' is only needed to collect all the files in a folder that will be shared at runtime, from where we can get all the files our needs.

tools/pctools/Dockerfile
m-dati: I was using the variable to create a filename with a proper common template for the copy phase. The same variable is ised  to create a venv dir. But I agree, I can change somethings and reduce parameters use

m-dati:
Please find here the log of a build/run execution after last update: http://pasta.qam.suse.de/mY8zjiei

tools/pctools/venv_inst.sh
grisu48: Up to you, both approaches work just fine :slightly_smiling_face:

tools/pctools/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# EC2
+ARG ENVEC2=ec2 
+ARG PIPEC2="ec2imgutils"
+
+# Openstack
+ARG ENVOPS=opstk 
+ARG PIPOPS="openstacksdk python-openstackclient typing-extensions importlib-metadata importlib-resources colorama"
+
+# create the internal container's environment
+RUN cat /etc/os-release; mkdir $WKDIR; pip install virtualenv; echo Pre-virtualenv; python3 --version; pip --version
```
grisu48, line 29: I would concatenate the commands with `&&` otherwise we continue to run this also if there are failures.  
If here .eg. `pip install virtualenv` fails, the `RUN` will still succeed.

Simply replacing all `;` with `&&` should do the job

tools/pctools/venv_inst.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# copy file in parent dir
+cp $RESOURCE_FILE .. 
+
+# file exists and size>0
+if [ ! -s ../$RESOURCE_FILE ]
+then
+    echo "Resource file $(pwd)/../$RESOURCE_FILE error"
+    exit 2
+fi
```
grisu48, line 58: Missing newline at the end

grisu48 (approved):
Two nitpicks, otherwise LGTM :rocket:

grisu48 (approved):

PR #15357, opened 2022-08-11
Public Cloud: Enable the boottime test in maintenance aggregates

pdostal:
- Related ticket: [poo#113252](https://progress.opensuse.org/issues/113252)
- Verification run: [pdostal-server.suse.cz/t475](https://pdostal-server.suse.cz/tests/475)


lib/publiccloud/instance.pm
```diff
@@ -250,6 +251,7 @@ sub wait_for_ssh
     }
 
     # Debug output: We have ocasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
```
grisu48, line 253:
```suggestion
    # Debug output: We have occasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
```

I know it's outside of this PR but this typo kinda bugs me ...

lib/publiccloud/instance.pm
```diff
@@ -227,6 +227,7 @@ sub wait_for_ssh
     $args{timeout} //= 600;
     $args{proceed_on_failure} //= 0;
     $args{username} //= $self->username();
+    $args{public_ip} //= $self->public_ip();
```
grisu48, line 230: Why is this needed for enabling the boottime test?

lib/publiccloud/instance.pm
```diff
@@ -250,6 +251,7 @@ sub wait_for_ssh
     }
 
     # Debug output: We have ocasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
+    script_run("ssh -v $args{username}\@$args{public_ip} true", timeout => 360);
```
grisu48, line 254: Why?

lib/publiccloud/instance.pm
pdostal: It's not.

pdostal: It's not.

lib/publiccloud/instance.pm
pdostal: Yes, but we see why it will fail. There is `-v`.

lib/publiccloud/instance.pm
grisu48: I see, it's needed for the ssh command above

lib/publiccloud/instance.pm
grisu48: I see, OK

grisu48 (approved):

lib/publiccloud/instance.pm
```diff
@@ -249,7 +255,8 @@ sub wait_for_ssh
         sleep 1;
     }
 
-    # Debug output: We have ocasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
+    # Debug output: We have occasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
```
asmorodskyi, line 258: this comment belong to one line below

asmorodskyi (changes requested):
this code just blindly adding current boottime test into QAM , this is not enough we need to skip https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/boottime.pm#L420 . We don't want to fail in this test , just record boot time into DB

asmorodskyi:
also we need to do something here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/boottime.pm#L362 . I know that we don't have that in QAM but we can store some image version ( image release date ? )

jlausuch:
Is the boottime test executed after installing updates?

pdostal:
> Is the boottime test executed after installing updates?

Not yet. There will be another pull request for it.

PR #15358, opened 2022-08-11
tests/installation/boot_into_snapshot: Wait until the snapshot menu is shown

Vogtinator:
Otherwise it will try to select the entry before the menu is visible.

- Verification run: https://openqa.opensuse.org/tests/2508793


DimStar77 (approved):
LGTM

PR #15361, opened 2022-08-12
Fix some more send_key_until_needlematch races

Vogtinator:
`send_key_until_needlematch` expects that it can get straight to work, but the
previous keypress might not be processed yet.

- Verification runs:
change_desktop: http://10.168.4.168/tests/1238
grub2_test: http://10.168.4.168/tests/1240

### DIFF
--- a/tests/installation/change_desktop.pm
+++ b/tests/installation/change_desktop.pm
@@ -43,7 +43,10 @@ sub change_desktop {
         }
         send_key 'ret';
     }
-    send_key_until_needlematch 'patterns-list-selected', 'tab', 10, 2;
+    # Give it some time to process the 'ret'
+    if (!check_screen('patterns-list-selected', 10)) {
+        send_key_until_needlematch 'patterns-list-selected', 'tab', 10, 2;
+    }
 
     if (is_sle('<=12-SP1') && get_var("REGRESSION", '') =~ /xen|kvm|qemu/) {
         assert_and_click 'gnome_logo';
--- a/tests/boot/grub2_test.pm
+++ b/tests/boot/grub2_test.pm
@@ -29,7 +29,7 @@ sub reboot {
 
 sub edit_cmdline {
     send_key 'e';
-    for (1 .. 13) { send_key 'down'; }
+    for (1 .. 12) { send_key 'down'; }
     send_key_until_needlematch 'grub2-edit-linux-line', 'down';
     send_key 'end';
 }


tests/installation/change_desktop.pm
```diff
@@ -43,7 +43,10 @@ sub change_desktop {
         }
         send_key 'ret';
     }
-    send_key_until_needlematch 'patterns-list-selected', 'tab', 10, 2;
+    # Give it some time to process the 'ret'
+    if (!check_screen('patterns-list-selected', 10)) {
+        send_key_until_needlematch 'patterns-list-selected', 'tab', 10, 2;
+    }
```
okurz, line 49: If this is really needed then it sounds like going back to the old behaviour of send_key_until_needlematch is better. Can you think of a better way avoiding `check_screen` with non-zero timeout? You are saying "fix races" but I think you know just as well that introducing such time-dependant checks is *introducing* race conditions

okurz (changes requested):

tests/installation/change_desktop.pm
Vogtinator: Well, `send_key_until_needlematch` itself is racy. If the key processing takes longer than the given `check_screen` timeout, it wouldn't work either.

I can think of a better way, but that's only something for next week. Instead of the `check_screen`, adding another `assert_screen 'pattern_selector';`, but that requires adding various new needles.

okurz: Sure, it was racy in before. Just saying that you are not really "fixing" the race condition, it's just shifted. The assert_screen way sounds good. Would be great if you could to this, doesn't have to be rushed. New explicit needles are often better and even saves time because checking an additional needle takes like 0.5s when `check_screen` needs to wait the full 10s.

okurz (approved):
See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15361/files#r944234962

tests/installation/change_desktop.pm
Vogtinator: > Sure, it was racy in before. Just saying that you are not really "fixing" the race condition, it's just shifted.

Right, the PR title is a bit misleading, but at least the commit messages should be accurate.

> The assert_screen way sounds good. Would be great if you could to this, doesn't have to be rushed. New explicit needles are often better and even saves time because checking an additional needle takes like 0.5s when check_screen needs to wait the full 10s.

Yep. I'll make a draft PR for that.

Vogtinator: I had a better idea: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15362

PR #15362, opened 2022-08-12
tests/installation/change_desktop: Replace delay with wait_screen_change

Vogtinator:
Instead of doing a check_screen which effectively just waits for the timeout in
the relevant case, use wait_screen_change to assure the 'ret' was processed.

Unlike a new `assert_screen`, this doesn't need new needles and is probably even quicker.

- Verification run: http://10.168.4.168/tests/1241

### DIFF
--- a/tests/installation/change_desktop.pm
+++ b/tests/installation/change_desktop.pm
@@ -41,12 +41,10 @@ sub change_desktop {
         for (1 .. 4) {
             wait_screen_change { send_key 'up'; };
         }
-        send_key 'ret';
+        wait_screen_change { send_key 'ret'; };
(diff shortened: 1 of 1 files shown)


okurz (approved):
Hm, I think an `assert_screen` call is always safer and most likely also faster than all the other calls but ok

Vogtinator:
> Hm, I think an assert_screen call is always safer and most likely also faster than all the other calls but ok

Probably safer, yes. I'd guess that calculating a diff is quicker than an `assert_screen` though, especially if there are multiple needles involved or it's not a 100% match in the exact pixel position.

PR #15370, opened 2022-08-16
Check default selection for yast2_keyboard.pm

coolgw:
We need check screen before sent down key on yast2_keyboard layout setting.

- Related ticket: https://progress.opensuse.org/issues/115220
- Needles: na
- Verification run: http://10.162.2.137/tests/374#step/yast2_keyboard/9

### DIFF
--- a/tests/yast2_gui/yast2_keyboard.pm
+++ b/tests/yast2_gui/yast2_keyboard.pm
@@ -45,7 +45,7 @@ sub run {
     send_key "g";
 
     # 2. Switch keymap from us to german
-    send_key_until_needlematch("yast2_keyboard-layout-german", "down");
+    send_key_until_needlematch("yast2_keyboard-layout-german", "down") unless check_screen("yast2_keyboard-layout-german", 30);
     wait_screen_change { send_key $accept_keybind };
     assert_screen "generic-desktop", timeout => 90;
 


jknphy:
good catch, so the test could only work when the 'g' key is lost, terrible design, hope we have time for this test modules in the future, they were done many years ago in the worst possible.
could you address ci error?

tests/yast2_gui/yast2_keyboard.pm
```diff
@@ -45,7 +45,7 @@ sub run {
     send_key "g";
 
     # 2. Switch keymap from us to german
-    send_key_until_needlematch("yast2_keyboard-layout-german", "down");
+    send_key_until_needlematch("yast2_keyboard-layout-german", "down") unless check_screen("yast2_keyboard-layout-german", 30);
```
ge0r, line 48: I m curious, could we avoid having this `unless` (which can be confusing to someone who is not aware of the problem), by just making the `yast2_keyboard-layout-german` needle wider to include the empty space on the right?
It will then only match the plain german version and the code will stay as it was.

jknphy: it is worse than that, we send `alt-g` before that, so we are supposed to be on the item, and then we always iterate +1 for what the test is designed :(
so to now overthink much I think it is acceptable. sometimes the 'alt-g' doesn't reach the SUT and there is not problem, it starts from item 1 to the german item we want.

ge0r (approved):

coolgw:
> good catch, so the test could only work when the 'g' key is lost, terrible design, hope we have time for this test modules in the future, they were done many years ago in the worst possible. could you address ci error (perhaps rebasing?)

I have no idea how to fix this CI, the error seems no relation with my change. I suggest just SKIP this CI error

jknphy:
other PRs do not have that problem, please, ensure your branch is up-to-date, rebasing for example.

PR #15372, opened 2022-08-16
tests/network/firewalld_policy_objects: Contact the right openQA host

Vogtinator:
The client and server VMs are possibly on other openQA workers than the
firewall, but they need to access their actual worker through 10.0.2.2 for
asset download and upload. Add an explicit route for that and rely on IP
rewriting done by OpenVSwitch.

Hopefully this fixes the remaining random firewalld test failures.

https://progress.opensuse.org/issues/115313#note-3

Depends on https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15371

Verification runs:

https://openqa.opensuse.org/tests/2517890
https://openqa.opensuse.org/tests/2517926

### DIFF
--- a/lib/mm_network.pm
+++ b/lib/mm_network.pm
@@ -231,6 +231,8 @@ sub restart_networking {
     if ($is_nm) {
         assert_script_run 'nmcli networking off';
         assert_script_run 'nmcli networking on';
+        # Wait until the connections are configured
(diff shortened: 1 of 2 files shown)


michaelgrifalconi (approved):
Not feeling expert enough to give you a real assessment, but at least it makes sense to me and I agree with your idea!

PR #15373, opened 2022-08-16
wicked: verify installation of wicked

cfconrad:
If WICKED_REPO is set, we will install wicked and
wicked-service from it. But this wasn't checked and thus we simple continue
but the version under test is something different!

- Verification run: 
  - http://openqa.wicked.suse.de/tests/76401#step/before_test/22 (successfull)
  - https://openqa.suse.de/t9342239 (error case)

### DIFF
--- a/tests/wicked/before_test.pm
+++ b/tests/wicked/before_test.pm
@@ -115,12 +115,12 @@ sub run {
             zypper_call("in --from $repo_id $resolv_options --force -y --force-resolution  wicked wicked-service", log => 'zypper_in_wicked.log');
             my ($zypper_in_output) = script_output('cat /tmp/zypper_in_wicked.log');
             my @installed_packages;
-            for my $reg (('The following \d+ packages? (are|is) going to be upgraded:',
-                    'The following NEW packages? (are|is) going to be installed:',
-                    'The following \d+ packages? (are|is) going to be reinstalled:')) {
-                push(@installed_packages, split(/\s+/, $+{packages})) if ($zypper_in_output =~ m/(?s)($reg(?<packages>.*?))(?:\r*\n){2}/);
-            }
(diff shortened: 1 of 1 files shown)


cfconrad:
VR with latest version: http://openqa.wicked.suse.de/tests/78064

tests/wicked/before_test.pm
```diff
@@ -115,12 +115,12 @@ sub run {
             zypper_call("in --from $repo_id $resolv_options --force -y --force-resolution  wicked wicked-service", log => 'zypper_in_wicked.log');
             my ($zypper_in_output) = script_output('cat /tmp/zypper_in_wicked.log');
             my @installed_packages;
-            for my $reg (('The following \d+ packages? (are|is) going to be upgraded:',
-                    'The following NEW packages? (are|is) going to be installed:',
-                    'The following \d+ packages? (are|is) going to be reinstalled:')) {
-                push(@installed_packages, split(/\s+/, $+{packages})) if ($zypper_in_output =~ m/(?s)($reg(?<packages>.*?))(?:\r*\n){2}/);
-            }
+            my $reg = 'The following (\d+|NEW) packages? (are|is) going to be (installed|reinstalled|upgraded|downgraded):';
```
asmorodskyi, line 118: maybe `reinstalled` should go into special flow and we should raise a warning ? I would say that if we simply reinstall ( meaning we installing exactly same version which is in image than something is wrong **outside** test

cfconrad: That's a valid point. But from CI perspective I would say we do not care as long as we get in the end the version which was passed with `WICKED_REPO`.
Remember that also for automated CI there is this extra check with `WICKED_COMMIT_SHA` which validate the changelog containing the right version.

For me, I would keep it as it is and not adding extra condition here.

PR #15376, opened 2022-08-17
wsl/prepare_wsl_feature: Close the serial port before installing the appx

Vogtinator:
As part of the appx installation, the Tumbleweed initial setup is started
at the end. This opens a window which conflicts with the "run_in_powershell"
command for closing the serial port. As the serial port isn't needed for
appx installation, just close it earlier already.

- Related ticket: https://progress.opensuse.org/issues/111473#note-49 (yes, it's the right one)
- Verification run: https://openqa.opensuse.org/tests/2520424#live


mloviska:
LGTM

PR #15381, opened 2022-08-19
Give more time to match needles on exit_firefox

michaelgrifalconi:
Give more time to match needles on exit_firefox, to avoid sending too many alt-f4 and then failing because we can't find the open terminal

- Related ticket: https://progress.opensuse.org/issues/115472
- Needles: none
- Verification run:
  - reproducing issue: https://openqa.suse.de/tests/9358395#details
  - test fix: 15 sp4 https://openqa.suse.de/tests/9358396
  - test fix: 12sp2 https://openqa.suse.de/tests/9358840
  - test fix: 15ga  https://openqa.suse.de/tests/9358841
  - test fix: 15sp2 https://openqa.suse.de/tests/9358842

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -717,7 +717,7 @@ sub exit_firefox_common {
     # Exit
     send_key 'ctrl-q';
     wait_still_screen 1, 2;
-    send_key_until_needlematch([qw(firefox-save-and-quit xterm-left-open xterm-without-focus)], "alt-f4", 3, 30);
+    send_key_until_needlematch([qw(firefox-save-and-quit xterm-left-open xterm-without-focus)], "alt-f4", 6, 30);
     if (match_has_tag 'firefox-save-and-quit') {
         # confirm "save&quit"
         send_key "ret";
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -716,14 +716,14 @@ sub firefox_preferences {
 sub exit_firefox_common {
     # Exit
     send_key 'ctrl-q';
-    wait_still_screen 1, 2;
+    wait_still_screen 3, 6;
     send_key_until_needlematch([qw(firefox-save-and-quit xterm-left-open xterm-without-focus)], "alt-f4", 6, 30);
     if (match_has_tag 'firefox-save-and-quit') {
         # confirm "save&quit"
         send_key "ret";
     }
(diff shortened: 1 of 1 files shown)


lib/x11test.pm
```diff
@@ -717,7 +717,7 @@ sub exit_firefox_common {
     # Exit
     send_key 'ctrl-q';
     wait_still_screen 1, 2;
-    send_key_until_needlematch([qw(firefox-save-and-quit xterm-left-open xterm-without-focus)], "alt-f4", 3, 30);
+    send_key_until_needlematch([qw(firefox-save-and-quit xterm-left-open xterm-without-focus)], "alt-f4", 6, 30);
```
rfan1, line 720: @michaelgrifalconi 
Seems a timing or performance issue here, I am wondering if we can add some timeout here
`wait_still_screen 1, 2;`

BTW, I would suggest do more VRs on other SLE versions which hit the same issue as well :)

michaelgrifalconi: Hello, started more validation runs.

I don't think a wait_still_screen at the beginning can help much, the issue is between multiple runs of send_key_until_needlematch.
It will send a alt-f4 key initially, wait for it to close firefox and if firefox is not fast enough to close, send another alt-f4 and then both firefox and xterm close and it fails. Increasing the time between sending these alt-f4 keys is the only relevant factor IMHO

rfan1: Well, the reason why I add the comments is that, I try to do some manual tests on my local setup.
`ctrl-q` key can close the firfox page, then I don't need to send `alt-f4` to force closing it.

Anyway, I agree with you, your changes make more sense!

michaelgrifalconi:
Issue is still happening in some tries, looking more into it

lib/x11test.pm
michaelgrifalconi: Right! the ctrl-q is the one supposed to close it, I missed that sorry! Maybe giving more time could help! I will try to increase it as well.

michaelgrifalconi:
Tried some statistical investigation with the fix, since the issue is sporadic: https://openqa.suse.de/tests/overview?build=mgrifalconi-firefoxdev

Does not look too bad.

rfan1 (approved):
LGTM

PR #15383, opened 2022-08-19
Fix security/yast2-bootloader

paolostivanin:
Install `yast2-bootloader` before calling it, since it's not installed by default.

- Verification run: https://openqa.suse.de/tests/9358993

### DIFF
--- a/tests/security/aarch64_secure_boot/yast2_bootloader.pm
+++ b/tests/security/aarch64_secure_boot/yast2_bootloader.pm
@@ -23,6 +23,14 @@ sub run {
     # ROM, it will change the bootloader file from shim.efi to
     # grubaa64.efi
     select_console("root-console");
+
+    # On the maintenance job group, we need to install yast2-bootloader, since
+    # it's not present in the created ISO.
+    my $parent_job = get_var('START_AFTER_TEST');
+    if ($parent_job eq "mru-install-minimal-with-addons_security_uefi") {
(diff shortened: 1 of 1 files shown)


punkioudi:
Looking from the last GMC tests: https://openqa.suse.de/tests/8752426, the yast2-bootloader module was pre-installed.
Could it be that it is a bug?

paolostivanin:
I guess that is caused because we are using 2 different jobs to create the iso. On the main jobgroup, we are using `create_hdd_*` , while on the maint jobgroup we are using `mru-install-*`.

punkioudi:
Yes, now I see that in maintenance it uses the qcow2 from a minimal installation, where it unselects the yast utilities.
My concern though is that for 15-SP5, this addition is not needed, so either you can add it under condition, either this test can use a qcow image with the yast utilities as default

paolostivanin:
Didn't think about sp5, good point Anna! I'll make a condition on the current test :)

punkioudi (approved):
Cool, lgtm! When the VR passes, we can merge it :)

paolostivanin:
sure, thanks :smile: I'll post the results as soon as they're done (should be about 30m)

paolostivanin:
15-sp3: https://openqa.suse.de/tests/9359330
15-sp4: https://openqa.suse.de/tests/9359331

jouyingbin (commented):
LGTM. sorry for being late.

PR #15385, opened 2022-08-19
Remove stderr from json output

lpalovsky:
OpenQA PC 'upload_image' tests are failing because of some Azure tool 
commands are reading stderr info json output files. This PR redirects 
the stderr into /dev/null to sppress those messages.

- Needles: N/A
- Failing test: http://10.100.96.70/tests/5838#step/upload_image/104
- VR: http://mordor.suse.cz/tests/4179#

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -76,8 +76,8 @@ sub find_img {
 
 sub get_storage_account_keys {
     my ($self, $storage_account) = @_;
-    my $output = script_output("az storage account keys list --resource-group "
-          . $self->resource_group . " --account-name " . $storage_account);
+    my $cmd = "az storage account keys list --resource-group " . $self->resource_group . " --account-name " . $storage_account;
+    my $output = script_output("echo \"\n $cmd\" > /dev/$serialdev; " . $cmd . " 2>/dev/$serialdev");
     my $json = decode_azure_json($output);
     my $key = undef;
     if (@{$json} > 0) {


lib/publiccloud/azure_client.pm
```diff
@@ -40,7 +40,7 @@ sub init {
 
 sub az_login {
     my ($self) = @_;
-    my $login_cmd = sprintf(q(while ! az login --service-principal -u $ARM_CLIENT_ID -p $ARM_CLIENT_SECRET -t $ARM_TENANT_ID; do sleep 10; done),
```
Vogtinator, line 43: The output here isn't read, so stderr could be preserved?

lpalovsky: Ah yes you are correct. Sorry, my mistake. I missed there is another message in the log file. Will revert this change and re-run the VR.

lpalovsky: Fixed and VR passed as well.

pdostal:
I suggest something like:
```perl
my $cmd = "az storage account keys list --resource-group " . $self->resource_group . " --account-name " . $storage_account;
assert_script_run($cmd);
my $output = script_output($cmd . " 2>/dev/null");
```

Vogtinator (approved):

Vogtinator:
> ```perl
> script_output
> ```

That runs the command twice. `script_output` has a `proceed_on_failure` parameter which can be set to `0` to error out if the command failed.

pdostal:
> > I suggest something like:
> 
> > 
> 
> > ```perl
> 
> > my $cmd = "az storage account keys list --resource-group " . $self->resource_group . " --account-name " . $storage_account;
> 
> > assert_script_run($cmd);
> 
> > my $output = script_output($cmd . " 2>/dev/null");
> 
> > ```
> 
> 
> 
> That runs the command twice. `script_output` has a `proceed_on_failure` parameter which can be set to `0` to error out if the command failed.

My point is that it is pitty yo loose the stderr as it can be useful for debug in case of a failure.

Vogtinator:
> > > I suggest something like:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > my $cmd = "az storage account keys list --resource-group " . $self->resource_group . " --account-name " . $storage_account;
> > 
> > 
> > > assert_script_run($cmd);
> > 
> > 
> > > my $output = script_output($cmd . " 2>/dev/null");
> > 
> > 
> > > 
> > 
> > 
> > That runs the command twice. `script_output` has a `proceed_on_failure` parameter which can be set to `0` to error out if the command failed.
> 
> My point is that it is pitty yo loose the stderr as it can be useful for debug in case of a failure.

Then instead of `2>/dev/null`, `2>/dev/$serialdev`? As long as the serial terminal is used that should work.

pdostal:
> Then instead of `2>/dev/null`, `2>/dev/$serialdev`? As long as the serial terminal is used that should work.

Yes, that should do the trick :+1:

lpalovsky:
> > Then instead of `2>/dev/null`, `2>/dev/$serialdev`? As long as the serial terminal is used that should work.
> 
> Yes, that should do the trick +1

Oh, cool. thanks. Going to try that one.

lpalovsky:
> Then instead of `2>/dev/null`, `2>/dev/$serialdev`? As long as the serial terminal is used that should work.

Tried that one but the output went into ttyS0 instead of serial_terminal and it is kinda out of place:
https://mordor.suse.cz/tests/4185/logfile?filename=serial0.txt

I am thinking about either somehow filter invalid data from string or if there is a possibility for json_decode to ignore malformed lines.

Vogtinator:
> > Then instead of `2>/dev/null`, `2>/dev/$serialdev`? As long as the serial terminal is used that should work.
> 
> Tried that one but the output went into ttyS0 instead of serial_terminal and it is kinda out of place: https://mordor.suse.cz/tests/4185/logfile?filename=serial0.txt

Yeah, but that's the best we can do currently.

> I am thinking about either somehow filter invalid data from string or if there is a possibility for json_decode to ignore malformed lines.

Not really. The root cause is still that stdout and stderr get mixed. Except for keeping them separate the entire time there's no correct way to handle that.

lib/publiccloud/azure.pm
```diff
@@ -76,8 +76,8 @@ sub find_img {
 
 sub get_storage_account_keys {
     my ($self, $storage_account) = @_;
-    my $output = script_output("az storage account keys list --resource-group "
-          . $self->resource_group . " --account-name " . $storage_account);
+    my $cmd = "az storage account keys list --resource-group " . $self->resource_group . " --account-name " . $storage_account;
+    my $output = script_output( "echo \"\n $cmd\" > /dev/$serialdev; " .$cmd . " 2>/dev/$serialdev");
```
pdostal, line 80: Why not just
```suggestion
    my $output = script_output($cmd . " 2> /dev/$serialdev");
```

lpalovsky: I added the echo so the serial console shows the commans itself as well, otherwise there would be only the error message without any context:
https://mordor.suse.cz/tests/4185/logfile?filename=serial0.txt
You would have the error message but no information where it comes from.

PR #15390, opened 2022-08-22
Add unit tests for YAML schedule with reusable flows

rakoenig:
### Links to Progress, etc:

- Related ticket: https://progress.opensuse.org/issues/110236
- Needles: -
- Verification run: -

### Positve tests
- Supply YAML_SCHEDULE_DEFAULT, one YAML_SCHEDULE_FLOWS entry and individual schedule, check if expected schedule is created (2 tests)
-  Supply YAML_SCHEDULE_DEFAULT, two YAML_SCHEDULE_FLOWS entries and individual_schedule, check if the last flow "wins" by writing the final schedule. (2 tests)

### Negative tests
- Supply invalid YAML_SCHEDULE_DEFAULT and check if scheduler dies
- Supply wron flow name in YAML_SCHEDULE_FLOWS and check if scheduler dies

### DIFF
--- a/t/02_yaml_schedule.t
+++ b/t/02_yaml_schedule.t
@@ -87,4 +87,32 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
     ok $modules[3] eq 'bar/test3', "Basic scheduling";
 };
 
+subtest 'merge default schedules with flows and individual schedules' => sub {
+    use scheduler;
+    use testapi 'set_var';
+
+    # positive unit tests
+
+    my $schedule = $ypp->load_file(dirname(__FILE__) . '/data/test_schedule_based_on_default.yaml');
+    my @modules;
+    set_var('YAML_SCHEDULE_DEFAULT', '/t/data/flows/default.yaml');
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a');
+    @modules = scheduler::parse_schedule($schedule);
+    my @expected = ('flow_a_foo', 'flow_a_bar', 'individual_step_3');
+    ok(eq_array(\@modules, \@expected), "Merge default with flows and schedule");
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_b');
+    @modules = scheduler::parse_schedule($schedule);
+    @expected = ('flow_b_foo', 'flow_b_bar', 'individual_step_3');
+    ok(eq_array(\@modules, \@expected), "Merge default with flows and schedule");
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a,flow_b');    # flow_b should win
+    @modules = scheduler::parse_schedule($schedule);
+    ok(eq_array(\@modules, \@expected), "Merge default with flows and schedule");
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_b,flow_a');    # flow_a should win
+    @modules = scheduler::parse_schedule($schedule);
+    @expected = ('flow_a_foo', 'flow_a_bar', 'individual_step_3');
+    ok(eq_array(\@modules, \@expected), "Merge default with flows and schedule");
+
+
+};
+
 done_testing;
--- a/t/data/flows/default.yaml
+++ b/t/data/flows/default.yaml
@@ -1,0 +1,7 @@
+---
+# minimal default file
+phase_1:
+  - step_1
+phase_2: []
+phase_3:
+  - step_3
--- a/t/data/flows/flow_a.yaml
+++ b/t/data/flows/flow_a.yaml
@@ -1,0 +1,6 @@
+---
+# example flow a
+phase_1:
+  - flow_a_foo
+phase_2: 
+  - flow_a_bar
--- a/t/data/flows/flow_b.yaml
+++ b/t/data/flows/flow_b.yaml
@@ -1,0 +1,6 @@
+---
+# example flow b
+phase_1: 
+  - flow_b_foo
+phase_2:
+  - flow_b_bar
--- a/t/data/test_schedule_based_on_default.yaml
+++ b/t/data/test_schedule_based_on_default.yaml
@@ -1,0 +1,6 @@
+---
+name: test_data_mini_default_flows
+description: Minimal file to merge with default and flows.
+schedule:
+  phase_3:
+    - individual_step_3


jknphy:
typo in the commit `s/mergeing/merging/`
also in the title, `Add unit testing...` or `Add unit tests...`

t/data/flows/default_broken.yaml
```diff
@@ -0,0 +1,7 @@
+---
+# minimal default file with an intentionally broken 3rd step
+phase_1:
+  - step_1
+phase_2: []
+phase_3
```
jknphy, line 6: this is not a valid case, it is not valid yaml, the problem is that we don't check with yamllint the flows, we should ...

t/02_yaml_schedule.t
```diff
@@ -87,4 +87,44 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
...
+    ok $modules[0] eq 'flow_a_foo', "Merge default with flows and individual";
+    ok $modules[1] eq 'flow_a_bar', "Merge default with flows and individual";
+    ok $modules[2] eq 'individual_step_3', "Merge default with flows and individual";
+
+    # negative unit tests
+    set_var('YAML_SCHEDULE_DEFAULT', '/t/data/flows/default_broken.yaml');
+    dies_ok { scheduler::parse_schedule($schedule) } 'Expected to see an error here';
+
+    set_var('YAML_SCHEDULE_DEFAULT', '/t/data/flows/default.yaml');
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a,nonexistent_flow');
```
jknphy, line 125: same here, this is not a negative case, it shouldn't happen, we should check existence of files, and in fact we do, we added that to our ci.
let me think about negative cases, if there is any...

t/data/test_data_mini_default_flows.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: test_data_mini_default_flows
```
jknphy, line 2: `---`
for consistency.

jknphy:
just noticed a little bug from previous work, that when no flows are provided (which is fine, because they are optional) the script through us an error: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/script/helper_default_based_schedule.pl#L42

t/02_yaml_schedule.t
```diff
@@ -87,4 +87,44 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
...
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a');
+    @modules = scheduler::parse_schedule($schedule);
+    ok $modules[0] eq 'flow_a_foo', "Merge default with flows and individual";
+    ok $modules[1] eq 'flow_a_bar', "Merge default with flows and individual";
+    ok $modules[2] eq 'individual_step_3', "Merge default with flows and individual";
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_b');
+    @modules = scheduler::parse_schedule($schedule);
+    ok $modules[0] eq 'flow_b_foo', "Merge default with flows and individual";
+    ok $modules[1] eq 'flow_b_bar', "Merge default with flows and individual";
+    ok $modules[2] eq 'individual_step_3', "Merge default with flows and individual";
```
jknphy, line 108: this is more like "key present in default and schedule but not present in flows is merged" I'm not sure "individual" can be understood correctly, as you are referring to 'individual' schedule, but I guess we should refer to the keys. wdyt?

t/02_yaml_schedule.t
```diff
@@ -87,4 +87,44 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
     ok $modules[3] eq 'bar/test3', "Basic scheduling";
 };
 
+subtest 'merge default schedules with flows and individual schedules' => sub {
```
jknphy, line 90: /schedules/schedule/
/schedules/schedule/

t/02_yaml_schedule.t
```diff
@@ -87,4 +87,44 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
...
+    use testapi 'set_var';
+
+    # positive unit tests
+
+    my $schedule = $ypp->load_file(dirname(__FILE__) . '/data/test_data_mini_default_flows.yaml');
+    my @modules;
+    set_var('YAML_SCHEDULE_DEFAULT', '/t/data/flows/default.yaml');
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a');
+    @modules = scheduler::parse_schedule($schedule);
+    ok $modules[0] eq 'flow_a_foo', "Merge default with flows and individual";
```
jknphy, line 101: this framework has options to compare arrays, we don't need to go one by one :)
https://perldoc.perl.org/Test::More#eq_array

jknphy (commented):
I found another bug by modifying one flow having a key in default which is cleared with a flow but then is overwritten in the schedule:
```
---
# example flow a
phase_1:
  - []
 ```
 It is expected to find bug doing unit testing :)
 
Result:
```
---
- []
- flow_a_bar
- individual_step_3
```

File: t/02_yaml_schedule.t (hunk 4 above)
Comment (rakoenig) [reply t/02_yaml_schedule.t:108]: My point of view for individual is that we have practically 3 layers in this scheduler:

- Default: This is where the keys that you can use are defined. The scheduler is only iterating though the keys found in the default schedule, so if I mistype a key in the other layers this wil be ignored.
- Flows: This is the first level of substituion. Usually defined on TestSuite or JobGroup level.
- Individual: This is what you really "create" when designing a test. Default and Flows are just like Lego bricks that you put together and get a standard framework, but here you fine tune for what you need to test.

**Comment by rakoenig** [comment]
> I found another bug by modifying one flow having a key in default which is cleared with a flow but then is overwritten in the schedule:
> 
> ```
> ---
> # example flow a
> phase_1:
>   - []
> ```
> 
> It is expected to find bug doing unit testing :)
> 
> Result:
> 
> ```
> ---
> - []
> - flow_a_bar
> - individual_step_3
> ```
> 
> it might need some additional check for empty array, could you check? https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/script/helper_default_based_schedule.pl#L60


Shouldn't clearing a key look like this?

```
---
# example flow a
phase_1: []
---
```

If I put this in the helper script then the key is cleared and does not show up in the output.

**Comment by rakoenig** [comment]
> just noticed a little bug from previous work, that when no flows are provided (which is fine, because they are optional) the script through us an error: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/script/helper_default_based_schedule.pl#L42
> Please add a positive case for it.

The error can be avoided by setting `YAML_SCHEDULE_DEFAULT="" `

**Comment by jknphy** [comment]
> > I found another bug by modifying one flow having a key in default which is cleared with a flow but then is overwritten in the schedule:
> > ```
> > ---
> > # example flow a
> > phase_1:
> >   - []
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is expected to find bug doing unit testing :)
> > Result:
> > ```
> > ---
> > - []
> > - flow_a_bar
> > - individual_step_3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it might need some additional check for empty array, could you check? https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/script/helper_default_based_schedule.pl#L60
> 
> Shouldn't clearing a key look like this?
> 
> ```
> ---
> # example flow a
> phase_1: []
> ---
> ```
> 
> If I put this in the helper script then the key is cleared and does not show up in the output.

you're right I messed up with list and keys, but the example is the same, it doesn't show in the output, which is surprising, the schedule should win somehow, but it makes no sense to combine with that flow probably, but we should reflect it as unit testing as weird behavior, right?

**Comment by jknphy** [comment]
> > just noticed a little bug from previous work, that when no flows are provided (which is fine, because they are optional) the script through us an error: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/script/helper_default_based_schedule.pl#L42
> > Please add a positive case for it.
> 
> The error can be avoided by setting `YAML_SCHEDULE_DEFAULT="" `

you can use default file and schedule file, and not use flows, that doesn't solve that or I didn't get your point.

File: t/02_yaml_schedule.t (hunk 4 above)
Comment (jknphy) [reply t/02_yaml_schedule.t:108]: I got what you mean, but it is not documented, and everybody knows this file as 'schedule' or 'YAML schedule' due to previous work, so if we refers to them as default, flows and schedule files is more simple.

**Comment by rakoenig** [comment]
> you can use default file and schedule file, and not use flows, that doesn't solve that or I didn't get your point.
> I still want to set the default.

I thought you're refering to the fact that when you call the helper script without setting YAML_SCHEDULE_FLOWS to an empty string it throws an error. That is the difference to the scheduler, that uses get_var with the default of an empty string. But of course I can put this in the helper script to check for an undefined environment variable.

**Comment by rakoenig** [comment]
> YAML_SCHEDULE_DEFAULT=""

That should have read YAML_SCHEDULE_FLOWS. Mea culpa .

**Comment by jknphy** [comment]
> > YAML_SCHEDULE_DEFAULT=""
> 
> That should have read YAML_SCHEDULE_FLOWS. Mea culpa .

no worries, thanks for taking a look, I'm mixing issues with unit testing and that script, but it is ok, all is connected.

File: t/02_yaml_schedule.t (hunk 4 above)
Comment (rakoenig) [reply t/02_yaml_schedule.t:108]: > I got what you mean, but it is not documented, and everybody knows this file as 'schedule' or 'YAML schedule' due to previous work, so if we refers to them as default, flows and schedule files is more simple.

Ok, let me rework the code for the unit tests and the script, then we see.

File: t/02_yaml_schedule.t [hunk 7]
```diff
@@ -87,4 +87,32 @@ subtest 'parse_yaml_test_schedule_recursive_conditional' => sub {
...
+
+    # positive unit tests
+
+    my $schedule = $ypp->load_file(dirname(__FILE__) . '/data/test_schedule_based_on_default.yaml');
+    my @modules;
+    set_var('YAML_SCHEDULE_DEFAULT', '/t/data/flows/default.yaml');
+    set_var('YAML_SCHEDULE_FLOWS', 'flow_a');
+    @modules = scheduler::parse_schedule($schedule);
+    my @expected = ('flow_a_foo', 'flow_a_bar', 'individual_step_3');
+    ok(eq_array(\@modules, \@expected), "Nerge default with flows and schedule");
```
jknphy, line 102: s/Nerge/Merge/

PR #15395, opened 2022-08-23
Change /usr/etc to /usr/lib for TW

punkioudi:
Related progress ticket: https://progress.opensuse.org/issues/115364

- Related ticket: https://progress.opensuse.org/issues/115364
- Verification run: [Tumbleweed](https://openqa.opensuse.org/tests/2538861) | [sle15sp4](https://openqa.suse.de/tests/9371744) | [Leap 15.4](https://openqa.opensuse.org/tests/2538862)

### DIFF
--- a/tests/security/pam/pam_login.pm
+++ b/tests/security/pam/pam_login.pm
@@ -9,6 +9,7 @@ use base 'opensusebasetest';
 use strict;
 use warnings;
 use testapi;
+use version_utils;
 
 sub run {
     my $self = shift;
@@ -17,15 +18,20 @@ sub run {
     # Define the user and password, which are already configured in previous milestone
     my $user = 'suse';
     my $passwd = 'susetesting';
+    my $pam_login_tw = '';
 
     # Modify the login/sshd files to set the PAM authentication
     my $deny_user_file = '/etc/deniedusers';
     my $pam_sshd = '/etc/pam.d/sshd';
-    my $pam_login = '/etc/pam.d/login';
     my $pam_sshd_bak = '/tmp/sshd_bak';
     my $pam_login_bak = '/tmp/login_bak';
     my $pam_sshd_tw = '/usr/etc/pam.d/sshd';
-    my $pam_login_tw = '/usr/etc/pam.d/login';
+    my $pam_login = '/etc/pam.d/login';
+    if (is_sle || is_leap) {
+        $pam_login_tw = '/usr/etc/pam.d/login';
+    } else {
+        $pam_login_tw = '/usr/lib/pam.d/login';
+    }
     my $ret_sshd = script_run("[[ -e $pam_sshd ]]");
     my $ret_login = script_run("[[ -e $pam_login ]]");
     assert_script_run "echo $user > $deny_user_file";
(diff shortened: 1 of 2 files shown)


tests/security/pam/pam_login.pm
```diff
@@ -17,15 +18,22 @@ sub run {
...
 
     # Modify the login/sshd files to set the PAM authentication
     my $deny_user_file = '/etc/deniedusers';
     my $pam_sshd = '/etc/pam.d/sshd';
-    my $pam_login = '/etc/pam.d/login';
     my $pam_sshd_bak = '/tmp/sshd_bak';
     my $pam_login_bak = '/tmp/login_bak';
     my $pam_sshd_tw = '/usr/etc/pam.d/sshd';
-    my $pam_login_tw = '/usr/etc/pam.d/login';
+    if (is_sle) {
```
DimStar77, line 30: And Leap ?

tests/security/pam/pam_login.pm
```diff
@@ -17,15 +18,22 @@ sub run {
...
-    my $pam_login = '/etc/pam.d/login';
     my $pam_sshd_bak = '/tmp/sshd_bak';
     my $pam_login_bak = '/tmp/login_bak';
     my $pam_sshd_tw = '/usr/etc/pam.d/sshd';
-    my $pam_login_tw = '/usr/etc/pam.d/login';
+    if (is_sle || is_leap) {
+        $pam_login = '/etc/pam.d/login';
+        $pam_login_tw = '/usr/lib/pam.d/login';
+    } else {
+        $pam_login = '/lib/pam.d/login';
```
DimStar77, line 34: That has a very negative side effect...

this code is used further down in:

```
my $ret_login = script_run("[[ -e $pam_login ]]");
…
    if ($ret_login != 0) {
        script_run "cp $pam_login_tw $pam_login";
    }
…
```

So this now copies a file from `/usr/lib/pam.d` to `/lib/pam.d`, which is, on TW, even the same. 
Even worse, that file is further down the line being modified.

This is definitively not what an admin is supposed to do; the admin is supposed to instanciate his template in /etc/pam.d and modify it there.

DimStar77: https://openqa.opensuse.org/tests/2536618#step/pam_login/23 here we can see it modifying a file in /lib

tests/security/pam/pam_su.pm
```diff
@@ -23,14 +24,25 @@ sub run {
...
+    my $sul_file = '';
+    my $su_file_tw = '';
+    my $sul_file_tw = '';
+    if (is_sle || is_leap) {
+        $su_file = '/etc/pam.d/su';
+        $sul_file = '/etc/pam.d/su-l';
+        $su_file_tw = '/usr/etc/pam.d/su';
+        $sul_file_tw = '/usr/etc/pam.d/su-l';
+    } else {
+        $su_file = '/lib/pam.d/su';
```
DimStar77, line 37: same issue here - we modify direcly in lib.

the copy job does not fail as /lib == /usr/lib, so the check further down if the file exists returns true.

The admin modified files are to remain in /etc/pam.d (so copy from /usr/lib/pam.d to /etc/pam.d)

tests/security/pam/pam_login.pm
punkioudi: Thanks for the comment, yes it makes sense, I will fix it!

DimStar77 (approved):

paolostivanin:
LGTM

jouyingbin (commented):
LGTM :-)

PR #15396, opened 2022-08-23
Revert workarounds on usr_sbin_smbd since fix has been released.

paolostivanin:
Since the gvfs fix has been released around mid August, we can now remove the workarounds (and the softfail).
- revert commit 7672f21020bf07261b65323521d938276dfd74c7
- revert commit 889a6d84ffdded60498667a6c3ff17f1a866e620

Verification runs:
- 15-SP3: https://openqa.suse.de/tests/9371238#step/usr_sbin_smbd/1
- 15-SP4: https://openqa.suse.de/tests/9371239#step/usr_sbin_smbd/1


tjyrinki (approved):
Very nice that the problem was fixed! And we will catch it again if it regresses.

PR #15399, opened 2022-08-23
LXQt: use sddm as default configured display manager

DimStar77:
- Related ticket: https://progress.opensuse.org/issues/115385
- Needles: 
- Verification run: https://openqa.opensuse.org/tests/2538863#live

### DIFF
--- a/tests/x11/lxqt_reconfigure_openqa.pm
+++ b/tests/x11/lxqt_reconfigure_openqa.pm
@@ -20,7 +20,7 @@ sub run {
 
     # LXQt uses sddm as window manager, which has the user preselected
     set_var('DISPLAYMANAGER', 'sddm');
-    # sddm asks straoght for PW with only one user; there is no need to type the username
+    # sddm asks straight for PW with only one user; there is no need to type the username
     set_var('DM_NEEDS_USERNAME', 0);
 
     $self->result('ok');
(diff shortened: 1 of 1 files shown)


DimStar77:
This was changed in TW with https://build.opensuse.org/request/show/995113

The Leap tests don't get that far anyway at the moment and I suspect strongly to get this also into Leap 15.5, which is why I did not add special handling for versions (i.e. all versions tested are the same)

DimStar77:
Actually, Leap 15.5 already caught up to this: https://build.opensuse.org/request/show/996454

tests/x11/lxqt_reconfigure_openqa.pm
```diff
@@ -18,9 +18,9 @@ sub run {
 
     set_var("DESKTOP", "lxqt");
 
-    # LXQt uses lightdm as window manager, which has the user preselected
-    set_var('DISPLAYMANAGER', 'lightdm');
-    # LightDM has the user in a drop-down preselected; there is no need to type the username
+    # LXQt uses sddm as window manager, which has the user preselected
+    set_var('DISPLAYMANAGER', 'sddm');
+    # sddm asks straoght for PW with only one user; there is no need to type the username
```
Vogtinator, line 23:
```suggestion
    # sddm asks straight for PW with only one user; there is no need to type the username
```

Vogtinator (approved):
Leap maintenance doesn't test LXQt either, so that should be fine.

+1 once the typo is fixed.

PR #15402, opened 2022-08-23
Fix SUSEConnect failure by adding product directly

Zaoliang:
because 15.4 used for SUSEConnect, python310_version_check.pm failed.
see ticket: https://progress.opensuse.org/issues/115661
VR: https://openqa.suse.de/tests/9431427#step/python310_version_check

### DIFF
--- a/tests/console/python310_version_check.pm
+++ b/tests/console/python310_version_check.pm
@@ -17,6 +17,7 @@ use warnings;
 use testapi;
 use version_utils;
 use utils "zypper_call";
+use registration "add_suseconnect_product";
 
 sub run {
     my $self = shift;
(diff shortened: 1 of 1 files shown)


tests/console/python310_version_check.pm
```diff
@@ -29,7 +29,7 @@ sub run {
             die("Python default version differs from 3.6");
         }
         my $arch = get_var('ARCH');
-        assert_script_run("SUSEConnect -p sle-module-python3/15.4/$arch");
+        assert_script_run("SUSEConnect -p sle-module-python3/15.5/$arch");
```
foursixnine, line 32: it's better to use: `add_suseconnect_product('sle-module-python3')`

foursixnine (changes requested):

tests/console/python310_version_check.pm
Zaoliang: yes, this makes sense, thanks!

See new VR: https://openqa.suse.de/tests/9431427#step/python310_version_check

foursixnine (approved):

PR #15404, opened 2022-08-23
Bump $count in calls to send_key_until_needlematch

ilmanzo:
The function `send_key_until_needlematch()` had an off-by-one error.
This error has been [recently fixed](https://github.com/os-autoinst/os-autoinst/pull/2151), so we need to increment +1 the `$count` argument (3rd positional) in all test suites calling it.

- Related ticket: https://progress.opensuse.org/issues/115619
- Verification Runs:
  - http://d505.qam.suse.de/tests/109
  - http://d505.qam.suse.de/tests/116
  - http://d505.qam.suse.de/tests/117
  - http://d505.qam.suse.de/tests/123
  - http://d505.qam.suse.de/tests/124
  - http://d505.qam.suse.de/tests/138
  - http://d505.qam.suse.de/tests/144

### DIFF
--- a/lib/Installation/Partitioner/FormattingOptionsPage.pm
+++ b/lib/Installation/Partitioner/FormattingOptionsPage.pm
@@ -79,7 +79,7 @@ sub select_partition_id {
     }
     # poo#35134 Sporadic synchronization failure resulted in incorrect choice of partition type
     # add partition screen was not refreshing fast enough
-    send_key_until_needlematch($partition_id_needle, 'up', 20, 5);
+    send_key_until_needlematch($partition_id_needle, 'up', 21, 5);
 }
 
 sub select_filesystem {
@@ -88,7 +88,7 @@ sub select_filesystem {
     assert_screen(FORMATTING_OPTIONS_PAGE);
     send_key($self->{filesystem_shortcut});
     send_key 'end', wait_screen_change => 1;
-    send_key_until_needlematch((sprintf FILESYSTEM_TYPE, $filesystem), 'up', 20, 5);
+    send_key_until_needlematch((sprintf FILESYSTEM_TYPE, $filesystem), 'up', 21, 5);
 }
 
 # Mounting Options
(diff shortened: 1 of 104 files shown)


jknphy (approved):

Martchus:
Have all of these calls really been made under the assumption that send_key_until_needlematch works like it did *before* https://github.com/os-autoinst/os-autoinst/pull/2151? To me that PR seems to "fix" something in the sense that most people would have expected that behavior anyways. However, if that is not true, I'm wondering about the usefulness of https://github.com/os-autoinst/os-autoinst/pull/2151.

ilmanzo:
> Have all of these calls really been made under the assumption that send_key_until_needlematch works like it did _before_ [os-autoinst/os-autoinst#2151](https://github.com/os-autoinst/os-autoinst/pull/2151)? To me that PR seems to "fix" something in the sense that most people would have expected that behavior anyways. However, if that is not true, I'm wondering about the usefulness of [os-autoinst/os-autoinst#2151](https://github.com/os-autoinst/os-autoinst/pull/2151).

I guess nobody can't tell if ALL the calls are made with some expectation or the other. Most of them are likely, where you can see some key are sent 9 times or 19 times [example1](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/97ddbd31beb063aa45cf19cdc98ee00325e1df30/tests/x11/firefox.pm#L32), [example2](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/97ddbd31beb063aa45cf19cdc98ee00325e1df30/tests/yast2_gui/yast2_control_center.pm#L208) when an EVEN number of keys are required.

Starting motivation of https://github.com/os-autoinst/os-autoinst/pull/2151 and [related ticket](https://progress.opensuse.org/issues/107749) is to make the testapi function behave like user expects, e.g. you ask to send a key up to 3 times, it sends the key up to 3 times, and not 4. Once we got that fixed, we should update accordingly the calling code. What do you think?

Martchus:
Ok, and more a decision of test writers anyways.

PR #15406, opened 2022-08-23
Implement hpc modules to test dolly package in a cluster

b10n1k:
Implement hpc modules to test dolly package in a cluster
    
Create a couple of files and copy them to other nodes using dolly. For more
info take a look at the documentation
https://documentation.suse.com/sle-hpc/15-SP3/html/hpc-guide/cha-dolly.html#
Many options are not supported so we go with the very basic scenario.
    
Some other approach/test cases might be to do the same with a config file.

- Related ticket: https://progress.opensuse.org/issues/113605
- Verification run: 
[master](http://aquarius.suse.cz/tests/12093)
[slave](http://aquarius.suse.cz/tests/12094)

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -57,6 +57,11 @@ sub destroy_test_barriers {
         barrier_destroy('SLAVE_MRLOGIN_STARTED');
         barrier_destroy('MRSH_MASTER_DONE');
     }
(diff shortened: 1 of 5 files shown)


schlad (approved):

czerw (approved):
LGTM

PR #15413, opened 2022-08-25
Increase wget timeout for larger images

grisu48:
To account for the increased size of Azure images (30GB) we need to
increase the timeout for `wget` downloads, which can take now more than an hour.

- Related ticket: https://progress.opensuse.org/issues/112883
- Verification run: https://duck-norris.qam.suse.de/tests/10487 and https://duck-norris.qam.suse.de/tests/10488

### DIFF
--- a/tests/publiccloud/upload_image.pm
+++ b/tests/publiccloud/upload_image.pm
@@ -19,7 +19,8 @@ use publiccloud::openstack;
 
 sub run {
     my ($self) = @_;
-    $self->select_serial_terminal;
+    # Better use the root-console here so that the download progress can be monitored in openQA
+    select_console("root-console");
 
     my $provider = $self->provider_factory();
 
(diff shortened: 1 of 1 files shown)


tests/publiccloud/upload_image.pm
```diff
@@ -31,14 +32,16 @@ sub run {
...
 
-    # Download the given image. Check for 404 errors and make them better visible
-    my $cmd = "wget --no-check-certificate $img_url -O $img_name";
-    my $rc = script_run("$cmd 2>download.txt", timeout => 60 * 10);
+    # Download the given image via wget. Note that by default wget retries 20 times before giving up
+    my $cmd = "wget --no-check-certificate --retry-connrefused --retry-on-host-error $img_url -O $img_name";
+    # A generious timeout is required because downloading up to 30 GB (Azure images) can take more than an hour.
+    my $rc = script_run("$cmd 2>&1 | tee download.txt", timeout => 120 * 60);
     if ($rc != 0) {
-        upload_logs("download.txt");
```
mloviska, line 38: Is this useless? IMO better than dumping the output in `record_info`

grisu48: The output containts 99% only the download progress, so IMHO this is completely useless to upload.
I might need to adjust the `script_output` as well, though :thinking: Will continue on Monday.

tests/publiccloud/upload_image.pm
```diff
@@ -31,14 +32,16 @@ sub run {
...
-        script_run("cat download.txt");
-        die "404 - Image not found" if (script_run("grep '404' download.txt") == 0);
-        die "command '$cmd' failed with rc=$rc";
+        # Check for 404 errors and make them better visible
+        my $output = script_output("cat download.txt");
+        record_info("wget failed with status code $rc", "$cmd\n\n$output");
```
pdostal, line 42:
```suggestion
        record_info("wget finished with status code $rc", "$cmd\n\n$output");
```

tests/publiccloud/upload_image.pm
```diff
@@ -31,14 +32,16 @@ sub run {
...
+        die "404 - Image not found" if ($output =~ "404");
+        die "wget failed with rc=$rc";
```
pdostal, line 44:
```suggestion
        die "wget failed with $rc return code.";
```

pdostal (approved):

tests/publiccloud/upload_image.pm
grisu48: Since this is in the `if ($rc != 0)` branch I would much rather keep it at failed.

tests/publiccloud/upload_image.pm
grisu48: ok

tests/publiccloud/upload_image.pm
grisu48: I added the `upload_logs` again and cropped the output in record_info to the last 10 lines of that log to avoid a HUGE wall of text.

grisu48:
New VR scheduled.

pdostal (approved):

PR #15414, opened 2022-08-25
Record soft failure if language isn't tranlated

rakoenig:
### Links to tickets, needles and VRs

- Related ticket: https://progress.opensuse.org/issues/115358
- Needles: -
- Verification run: [SLE 15 SP5](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=softfail_license_translations&groupid=96) (note that the modified test softfails, but the whole test suite fails later)

### Additional info 

- in case of not validating EULA translations we now just record a soft
  failure instead of letting the test die.
- other problems like wrong preselected default language or
  langauage is not available will still lead to errors.
- So in the beginning of the product lifecycle we won't fail
  anymore, just record soft failures. At the end, close to GM,
   those soft failures should not happen anymore.

### DIFF
--- a/tests/installation/licensing/verify_license_translations.pm
+++ b/tests/installation/licensing/verify_license_translations.pm
@@ -55,7 +55,7 @@ sub run {
         $license_agreement->select_language($translation->{language});
         $license_agreement_info = $license_agreement->collect_current_license_agreement_info();
         if ($license_agreement_info->{text} !~ /$translation->{text}/) {
-            $errors .= "EULA content for the language: '$translation->{language}' didn't validate. Please, see autoints-log for the detailed content of EULA\n";
+            record_soft_failure("EULA content for the language: '$translation->{language}' didn't validate. See bsc#1203004 for details.\n");
             diag("EULA validation failed:\nExpected:\n$translation->{text}\nActual:\n$license_agreement_info->{text}\n\n");
         }
     }


ilmanzo:
LGTM

tests/installation/licensing/verify_license_translations.pm
```diff
@@ -55,7 +55,7 @@ sub run {
         $license_agreement->select_language($translation->{language});
         $license_agreement_info = $license_agreement->collect_current_license_agreement_info();
         if ($license_agreement_info->{text} !~ /$translation->{text}/) {
-            $errors .= "EULA content for the language: '$translation->{language}' didn't validate. Please, see autoints-log for the detailed content of EULA\n";
+            record_soft_failure("EULA content for the language: '$translation->{language}' didn't validate. Please, see autoints-log for the detailed content of EULA\n");
```
b10n1k, line 58: i wonder if you should check for BETA in this case as IIUC what you want is to prevent false negatives in this stage of the project. Do we expect those translations to be ready in BETA? Or we are not sure?

b10n1k (commented):
My only concern is that the softfail will overseen (marked as softfail) even when the localisation is expected, but it might not be noticed then. Wdyt?

tests/installation/licensing/verify_license_translations.pm
rakoenig: I raised this question in our daily and got assured, that nothing will be overseen in the reviews.

b10n1k: So the problem is that there is no way to automate the logic or what? As a QA I would say, do not trust anyone, an assurance wont do it. And it is not exactly justified. But its fine i guess if you want to oversee it for now. As minimum requirement i would add a comment in the softfail for the reviewer, for instance something like "This should not softfail in GA" or "Do not oversee this after BETA" or something. WDYT?

b10n1k (approved):
thanks @rakoenig for addition.

tests/installation/licensing/verify_license_translations.pm
```diff
@@ -55,7 +55,7 @@ sub run {
         $license_agreement->select_language($translation->{language});
         $license_agreement_info = $license_agreement->collect_current_license_agreement_info();
         if ($license_agreement_info->{text} !~ /$translation->{text}/) {
-            $errors .= "EULA content for the language: '$translation->{language}' didn't validate. Please, see autoints-log for the detailed content of EULA\n";
+            record_soft_failure("EULA content for the language: '$translation->{language}' didn't validate. This should not happen after BETA!\n");
```
jknphy, line 58: this is more complicated than that, there are 3 stages for the eula:
https://confluence.suse.com/display/SUSELinuxEnterpriseServer15SP5/SLE15+SP5+Quality
and in practice I'm been requested some of them until GMC
We don't have that granularity in this test, but it sort of work because we match with a small amount of text.
Therefore it might not match after BETA, and the idea it is not having failing this scenario and been able to adjust the workaround with the missing parts.
- But we still to have a bug open with the language missing stating that for now is expected the Beta EULAS for them.
- @rakoenig what about to create a new field in the test data `available: yes` | `available: no` so we still can track what fails and what we accept which is not there yet?

As we don't have the granularity to distinguish between beta uelas and final eulas in this test, I think it would be more useful to point to the bug that this additional message, and only record soft-failure for those two cases, but if in next build is missing another one, it should fail, we need to adjust it and go to the bug and comment the issue.

rakoenig: @jknphy creating new fields in the test data which is part of the yaml would mean, that we need to update this whenever a new translation is available. That looks a bit weird to me.

jknphy: I see what you meant, so we will create a record sof-failure for each one missing, that is fine, perhaps we don't need any inclusion list, just replace 'This should not happen after BETA!' by the a NEW bug reference, because according to that confluence link is not that clear that after beta you have the translations for everything. makes sense to you know?

PR #15415, opened 2022-08-25
Limit swap space to 2GB in pc_tools image

grisu48:
Limit the swap partition to 2GB instead of using half of the available
disk space (currently 50GB) for swap.

- Related ticket: https://progress.opensuse.org/issues/112883
- Verification run: https://duck-norris.qam.suse.de/t10481

### DIFF
--- a/data/autoyast_sle15/pc_tools.xml
+++ b/data/autoyast_sle15/pc_tools.xml
@@ -116,13 +116,14 @@
       <partitions config:type="list">
         <partition>
           <mountby config:type="symbol">device</mountby>
-          <filesystem config:type="symbol">swap</filesystem>
-          <mount>swap</mount>
+          <filesystem config:type="symbol">ext4</filesystem>
+          <mount>/</mount>
         </partition>
         <partition>
           <mountby config:type="symbol">device</mountby>
-          <filesystem config:type="symbol">ext4</filesystem>
-          <mount>/</mount>
+          <filesystem config:type="symbol">swap</filesystem>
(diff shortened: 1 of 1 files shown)


grisu48:
The resulting layout looks like the following

screen

data/autoyast_sle15/pc_tools.xml
```diff
@@ -116,13 +116,14 @@
...
+          <filesystem config:type="symbol">ext4</filesystem>
+          <mount>/</mount>
         </partition>
         <partition>
           <mountby config:type="symbol">device</mountby>
-          <filesystem config:type="symbol">ext4</filesystem>
-          <mount>/</mount>
+          <filesystem config:type="symbol">swap</filesystem>
+          <mount>swap</mount>
+          <size>2G</size>
```
pdostal, line 126: I would get rid of the swap completely.

pdostal (approved):
That's good that you are shrinking the swap so I'm OK to merge this but I'd shrink the swap to zero.

data/autoyast_sle15/pc_tools.xml
grisu48: Good idea. I would still proceed with this PR to first reduce the swap space and then remove it, so we can easier roll back if needed.

Let's do two steps instead of one.

grisu48: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15417

PR #15418, opened 2022-08-26
wait for the screen to be still after tab key press

bs-suse:
To avoid pressing the _right_ key before the previous _tab_ has been taken effect wait for a still screen for 2 seconds.

- Related: https://github.com/os-autoinst/os-autoinst/pull/2123#issuecomment-1219366078
- Verification run: https://openqa.suse.de/tests/9387550#step/releasenotes/3


alvarocarvajald (approved):
LGTM

ricardobranco777 (approved):
LGTM

PR #15419, opened 2022-08-26
Optionally get Trento deployment scripts from a release tag

mpagot:
Add lib function to take care to get or just update the Trento
deployment script accordingly to the presence of TRENTO_DEPLOY_VER
variable.
Add lib function to get the public IP of the deployed Trento web app: it
also allow specifying the IP using a variable, meaning that the test has
to be performed against an instance of Trento deployed outside openQA.
Lib function to get the Trento web admin password.
Manage the version of the cypress/included docker image in a more
centralized way.
Remove most of the configuration variables from the schedule yaml file
as managed at JobGroup level or using default values in the test code.
Improved deployment scripts log management, especially in case of
failure.

- Verification run: http://1c242.qa.suse.de/tests/898 --> http://1c242.qa.suse.de/tests/903

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -33,18 +33,23 @@ use testapi;
 use mmapi 'get_current_job_id';
 use utils qw(script_retry);
 use File::Basename qw(basename);
+use Mojo::JSON qw(decode_json);
+
 use Exporter 'import';
 
 our @EXPORT = qw(
+  get_trento_deployment
   get_resource_group
   get_vm_name
   get_acr_name
-  az_get_vm_ip
+  get_trento_ip
+  get_trento_password
   VM_USER
   SSH_KEY
   cypress_install_container
   CYPRESS_LOG_DIR
   PODMAN_PULL_LOG
+  cypress_version
   cypress_exec
   cypress_test_exec
   cypress_log_upload
@@ -63,12 +68,96 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
 # Azure does not support dash or underscore in ACR name
 use constant TRENTO_AZ_ACR_PREFIX => 'openqatrentoacr';
 use constant CYPRESS_IMAGE_TAG => 'goofy';
+use constant CYPRESS_IMAGE => 'docker.io/cypress/included';
 
 =head1 DESCRIPTION 
 
-Package with common methods and default or constant  values for Trento tests
+Package with common methods and default or
+constant values for Trento tests
 
 =head2 Methods
+=cut
+
+=head3 get_trento_deployment
+
+Get the set of scripts for the Trento deployment
+=cut
+
+sub get_trento_deployment {
+    my ($self, $work_dir) = @_;
+
+    enter_cmd "cd $work_dir";
+    script_run 'read -s GITLAB_TOKEN', 0;
+    type_password get_var(TRENTO_GITLAB_TOKEN => get_required_var('_SECRET_TRENTO_GITLAB_TOKEN')) . "\n";
+
+    # Script from a release
+    if (get_var('TRENTO_DEPLOY_VER')) {
+
+        # Clean up whatever is present in the indicated working folder
+        # of the qcow2 image, to avoid confusion
+        enter_cmd 'rm -rf *';
+
+        # Get some 'coordinates' from the test settings
+        my $ver = get_var('TRENTO_DEPLOY_VER');
+        my $gitlab_repo = get_var(TRENTO_GITLAB_REPO => 'gitlab.suse.de/qa-css/trento');
+        my @gitlab_url = split('/', $gitlab_repo);
+        my $gitlab_namespace = $gitlab_url[-2];
+        my $gitlab_project = $gitlab_url[-1];
+
+        assert_script_run('echo "PRIVATE-TOKEN: ${GITLAB_TOKEN}" > gitlab_conf');
+        my $curl_cmd = 'curl -s ' .
+          '-H @gitlab_conf';
+        my $gitlab_api = 'https://gitlab.suse.de/api/v4/projects';
+
+        # Get the ID of the requested project
+        my $repo_id_cmd = $curl_cmd .
+          " \"$gitlab_api/$gitlab_namespace%2F$gitlab_project\"";
+        my $repo_output = decode_json(script_output($repo_id_cmd));
+        my $repo_id = $repo_output->{id};
+
+        my $repo_url = "$gitlab_api/$repo_id";
+        my $rel_api_url = "$repo_url/releases/v$ver";
+
+        # Get the file name of the release archive
+        my $ver_artifact_cmd = "basename \$($curl_cmd $rel_api_url " .
+          "| jq -r '.assets.sources[]|select(.format == \"tar.gz\")" .
+          "|.url')";
+        my $ver_artifact = script_output($ver_artifact_cmd);
+
+        # Get the commit ID of the release
+        my $commit_id_cmd = "$curl_cmd $rel_api_url " .
+          "| jq -r .commit.id";
+        my $commit_id = script_output($commit_id_cmd);
+
+        # Download the release archive
+        my $download_api_url = "$repo_url/repository/archive.tar.gz?sha=$commit_id";
+        my $download_cmd = $curl_cmd .
+          " \"$download_api_url\"" .
+          " --output $ver_artifact ";
+        assert_script_run($download_cmd);
+
+        # Extract and test file presence
+        my $tar_cmd = "tar xvf $ver_artifact --strip-components=1";
+        assert_script_run($tar_cmd);
+        enter_cmd 'ls -lai';
+    }
+
+    # Script from Gitlab
+    else {
+        my $git_branch = get_var(TRENTO_GITLAB_BRANCH => 'master');
+
+        # Test that the token in worker.ini and the one in the qcow2 match
+        my $qcow_token_cmd = 'git config --get remote.origin.url' .
+          '|cut -d@ -f1|cut -d: -f3';
+        assert_script_run "QCOW_TOKEN=\"\$($qcow_token_cmd)\"";
+        my $different_tokens = script_run '[[ "$GITLAB_TOKEN" == "$QCOW_TOKEN" ]]';
+        die "Invalid gitlab token" if ($different_tokens);
+
+        # Switch branch and get latest
+        assert_script_run("git checkout $git_branch");
+        assert_script_run("git pull origin $git_branch");
+    }
+}
 
 =head3 get_resource_group
 
@@ -102,18 +191,44 @@ sub get_acr_name {
     return TRENTO_AZ_ACR_PREFIX . get_current_job_id();
 }
 
-=head3 az_get_vm_ip
+=head3 get_trento_ip
 
 Return the running VM public IP
 =cut
 
-sub az_get_vm_ip {
-    my $az_cmd = 'az vm show -d ' .
-      '-g ' . get_resource_group() . ' ' .
-      '-n ' . get_vm_name() . ' ' .
-      '--query "publicIps" ' .
-      '-o tsv';
+sub get_trento_ip {
+    my $machine_ip = get_var('TRENTO_EXT_DEPLOY_IP');
+    return $machine_ip if ($machine_ip);
+    my $az_cmd = 'az vm show -d' .
+      ' -g ' . get_resource_group() .
+      ' -n ' . get_vm_name() .
+      ' --query "publicIps"' .
+      ' -o tsv';
     return script_output($az_cmd, 180);
+}
+
+=head3 get_trento_password
+
+Return the password for the Trento WebUI
+=cut
+
+sub get_trento_password {
+    my $self = shift;
+    my $trento_web_password;
+
+    if (get_var('TRENTO_EXT_DEPLOY_IP')) {
+        $trento_web_password = get_required_var('TRENTO_WEB_PASSWORD');
+    }
+    else {
+        my $machine_ip = get_trento_ip;
+        record_info("TRENTO IP", $machine_ip);
+        my $trento_web_password_cmd = $self->az_vm_ssh_cmd(
+            'kubectl get secret trento-server-web-secret' .
+              " -o jsonpath='{.data.ADMIN_PASSWORD}'" .
+              '|base64 --decode', $machine_ip);
+        $trento_web_password = script_output($trento_web_password_cmd);
+    }
+    return $trento_web_password;
 }
 
 =head3 az_delete_group
@@ -149,12 +264,12 @@ sub az_vm_ssh_cmd {
     my ($self, $cmd_arg, $vm_ip_arg) = @_;
     record_info('REMOTE', "cmd:$cmd_arg");
 
-    # undef comparision operator
-    my $vm_ip = $vm_ip_arg // az_get_vm_ip();
+    # Undef comparison operator
+    $vm_ip_arg //= get_trento_ip();
     return 'ssh' .
       ' -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR' .
       ' -i ' . SSH_KEY . ' ' .
-      VM_USER . '@' . $vm_ip .
+      VM_USER . '@' . $vm_ip_arg .
       ' -- ' . $cmd_arg;
 }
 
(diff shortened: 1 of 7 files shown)


lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+=head3 get_trento_deployment
+
+Get the set of scripts for the Trento deployment
+=cut
+
+sub get_trento_deployment {
+    my ($self, $work_dir) = @_;
+
+    enter_cmd "cd ${work_dir}";
+    my $gitlab_token = get_var(TRENTO_GITLAB_TOKEN => get_required_var('_SECRET_TRENTO_GITLAB_TOKEN'));
```
dzedro, line 88: What is the difference between `TRENTO_GITLAB_TOKEN` and `_SECRET_TRENTO_GITLAB_TOKEN` ?

mpagot: It was a way to let the user specify its own token, but as the repo is fixed, it makes no more sense. Could be removed.

tests/sles4sap/trento/deploy_trento.pm
```diff
@@ -25,8 +28,10 @@ sub run {
 
     my $script_run = ' ./';
     # Run the Trento deployment
+
+    record_info('00.040');
```
dzedro, line 32: What does this `00.040` mean ?

tests/sles4sap/trento/test_trento_web.pm
```diff
@@ -17,22 +17,17 @@ sub run {
...
 
     my $cypress_test_dir = "/root/test/test";
     enter_cmd "cd " . $cypress_test_dir;
     my $cypress_env_cmd = './cypress.env.py' .
       " -u http://$machine_ip" .
       " -p $trento_web_password" .
       ' -f Premium';
-    if (get_var("TRENTO_VERSION")) {
-        $cypress_env_cmd .= ' --trento-version ' . get_var("TRENTO_VERSION");
+    if (get_var('TRENTO_VERSION')) {
```
dzedro, line 29: I know you just changed the quotes, but `TRENTO_VERSION` does not make sense to me, the trento version does not require variable to find out what trento version is running.

tests/sles4sap/trento/deploy_trento.pm
mpagot: It is the ID of the deployment script, it is part of the name of the deployment script

dzedro: Thank you

lib/trento.pm
dzedro: You can change it later.

dzedro (approved):

tests/sles4sap/trento/test_trento_web.pm
mpagot: This line of code is in charge to tell to the cypress engine (used to test the web interface of Trento) which string to expect about the version on the web interface.
This variable usage can be improved here and in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15419/files#diff-9616afe1f4b58be89b042ba489a6f886f4ff68c43e9c1cdb613ccd885078a5a7L43

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        $repo_id = '7182';
+
+        my $repo_url = "${gitlab_api}/${repo_id}";
+        my $rel_api_url = "${repo_url}/releases/v${ver}";
+
+        # Get the file name of the release archive
+        my $ver_artifact_cmd = $curl_cmd .
+          " ${rel_api_url} " .
+          "| jq -r '.assets.sources[]|select(.format == \"tar.gz\")" .
+          "|.url'|rev|cut -d/ -f1|rev";
```
alvarocarvajald, line 123: Neat trick with the double `rev` commands, but you could do the same with a shorter shell pipeline and `File::Basename::basename` and even drop the `jq` and use `decode_json` as well.

If wanting to keep this in the shell and not in the Perl code, there is also `basename`:
```
$ echo /home/michelepa/trento/my_directory | rev | cut -d/ -f1 | rev
my_directory
$ basename /home/michelepa/trento/my_directory
my_directory
$ basename $(echo /home/michelepa/trento/my_directory)
my_directory
$ curl -s --header "PRIVATE-TOKEN: ****"  https://gitlab.url/api/v4/projects/7182/releases/v0.2.0 | jq -r '.assets.sources[]|select(.format == "tar.gz")|.url'|rev|cut -d/ -f1|rev
trento-v0.2.0.tar.gz
$ basename $(curl -s --header "PRIVATE-TOKEN: ****"  https://gitlab.url/api/v4/projects/7182/releases/v0.2.0 | jq -r '.assets.sources[]|select(.format == "tar.gz")|.url')
trento-v0.2.0.tar.gz
```

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        my $ver_artifact = script_output($ver_artifact_cmd);
+
+        # Get the commit ID of the release
+        my $commit_id_cmd = $curl_cmd .
+          " ${rel_api_url} " .
+          "| jq -r .commit.id";
+        my $commit_id = script_output($commit_id_cmd);
+
+        # Download the release archive
+        my $download_api_url = "${repo_url}/repository/archive.tar.gz?sha=${commit_id}";
```
alvarocarvajald, line 133: Curly brackets not needed.

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        my $download_cmd = $curl_cmd .
+          " \"${download_api_url}\"" .
+          " --output ${ver_artifact} ";
```
alvarocarvajald, line 136: Curly brackets not needed.

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        assert_script_run($download_cmd);
+
+        # Extract and test file presence
+        my $tar_cmd = "tar xvf ${ver_artifact} --strip-components=1";
```
alvarocarvajald, line 140: Curly brackets not needed.

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        assert_script_run($tar_cmd);
+        script_run('ls -lai');
```
alvarocarvajald, line 142: If not checking the return code, this can be replaced with an `enter_cmd`

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+
+        # Test that the token in worker.ini and the one in the qcow2 match
+        my $qcow_token_cmd = 'git config --get remote.origin.url' .
+          '|cut -d@ -f1|cut -d: -f3';
+        my $qcow_token = script_output($qcow_token_cmd);
+        die "Invalid gitlab token" unless $gitlab_token eq $qcow_token;
+
+        # Switch branch and get latest
+        assert_script_run("git checkout " . $git_branch);
+        assert_script_run("git pull origin " . $git_branch);
```
alvarocarvajald, line 157: Won't this make repeatability of the test complicated? If you run it once and then restart it, there's no guarantee that you're running the same code from the trento repo.

lib/trento.pm
```diff
@@ -102,18 +190,49 @@ sub get_acr_name {
...
+    }
+    else {
+        my $az_cmd = 'az vm show -d ' .
+          '-g ' . get_resource_group() . ' ' .
+          '-n ' . get_vm_name() . ' ' .
+          '--query "publicIps" ' .
+          '-o tsv';
+        $machine_ip = script_output($az_cmd, 180);
+    }
+    return $machine_ip;
```
alvarocarvajald, line 211: Suggestion:
```
my $machine_ip = get_var('TRENTO_EXT_DEPLOY_IP');
return $machine_ip if ($machine_ip);
my $az_cmd = 'az vm show -d ' .
          '-g ' . get_resource_group() . ' ' .
          '-n ' . get_vm_name() . ' ' .
          '--query "publicIps" ' .
          '-o tsv';
return script_output($az_cmd, 180);
```

lib/trento.pm
```diff
@@ -274,6 +392,20 @@ sub cypress_log_upload {
...
+Return the cypress.io version to use.
+It could be the default one or one fixed by the user using TRENTO_CYPRESS_VERSION
+
+=cut
+
+sub cypress_version {
+    my $self = shift;
+    my $cypress_ver = get_var(TRENTO_CYPRESS_VERSION => '9.6.1');
+
+    return $cypress_ver;
```
alvarocarvajald, line 404: Suggestion:
```
return get_var(TRENTO_CYPRESS_VERSION => '9.6.1');
```

tests/sles4sap/trento/deploy_trento.pm
alvarocarvajald, line 32: You are using `00.040` in several places below. Perhaps you can assign it in a variable and then when you need to change it, you will only need to change it in only one place.

tests/sles4sap/trento/deploy_trento.pm
```diff
@@ -82,15 +88,16 @@ sub run {
...
+    my $machine_ip = $self->get_trento_ip;
     my $acr_server = script_output("az acr list -g $resource_group --query \"[0].loginServer\" -o tsv");
     my $acr_username = script_output("az acr credential show -n $acr_name --query username -o tsv");
     my $acr_secret = script_output("az acr credential show -n $acr_name --query 'passwords[0].value' -o tsv");
 
     # Check what registry has been created by  trento_acr_azure_cmd
     assert_script_run("az acr repository list -n $acr_name");
 
-    $deploy_script_log = 'script_1.010.log.txt';
+    record_info('01.010');
```
alvarocarvajald, line 99: Same comment as with `00.040`

tests/sles4sap/trento/init_jumphost.pm
```diff
@@ -8,16 +8,14 @@ use Mojo::Base 'publiccloud::basetest';
...
 sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
     # Get the code for the Trento deployment
-    enter_cmd 'cd ${HOME}/test';
-    my $git_branch = get_var('TRENTO_GITLAB_BRANCH', 'master');
-    assert_script_run("git checkout " . $git_branch);
-    assert_script_run("git pull origin " . $git_branch);
+    $self->get_trento_deployment('${HOME}/test');
```
alvarocarvajald, line 18: Curly brackets are not needed.

tests/sles4sap/trento/setup_jumphost.pm
```diff
@@ -35,10 +35,12 @@ sub run {
...
 
     # The usage of a variable with a different name is to
-    # be able to overwrite the token when triggering manually.
+    # be able to overwrite the token when manually triggering
+    # the setup_jumphost test.
     #
-    # Note: this test is mostly used to create a HDD image; part of it is this cloned repo.
-    # The key is part of the cloned repo itself so TRENTO_GITLAB_TOKEN for the moment
+    # Note: this test is mostly only used to create
+    # a HDD image; part of this test adn the qcow2 image is this cloned repo.
```
alvarocarvajald, line 42: s/adn/and/

alvarocarvajald (commented):
Besides the in-line comments, I think you may need to do something to avoid printing the private token in the test results. As it is, it is good the information can be easily accessed in the results if needing to debug, and these tests as well as our gitlab instance should be internal only, but I think that eventually we may want to avoid printing the token in the results.

lib/trento.pm
```diff
@@ -63,12 +66,97 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        my $curl_cmd = 'curl -s ' .
+          "--header \"PRIVATE-TOKEN: ${gitlab_token}\" ";
+        my $gitlab_api = 'https://gitlab.suse.de/api/v4/projects';
+
+        # Get the ID of the requested project
+        my $repo_id_cmd = $curl_cmd .
+          " \"${gitlab_api}/${gitlab_namespace}%2F${gitlab_project}\"" .
+          " | jq -r '.id'";
+        my $repo_id = script_output($repo_id_cmd);
+        # hotfix: harcoded here as not able to strip color codes added also if using -r
```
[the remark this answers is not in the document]
mpagot: echo trick tested with no luck. 
The `-r` option for the **jq** command should force the output to be a raw string, and it works for all the others ...

lib/trento.pm
mpagot: yes, and for this reason I created TRENTO_DEPLOY_VER
On the other hand, the 'pull' approach help a lot during the deployment script and test development, and allow to use custom deployment script when debugging failures

lib/trento.pm
```diff
@@ -63,12 +68,95 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+Get the set of scripts for the Trento deployment
+=cut
+
+sub get_trento_deployment {
+    my ($self, $work_dir) = @_;
+
+    enter_cmd "cd $work_dir";
+    my $gitlab_token = get_var(TRENTO_GITLAB_TOKEN => get_required_var('_SECRET_TRENTO_GITLAB_TOKEN'));
+    script_run 'read -s GITLAB_TOKEN', 0;
+    type_password get_var(TRENTO_GITLAB_TOKEN => get_required_var('_SECRET_TRENTO_GITLAB_TOKEN')) . "\n";
```
mpagot, line 91: Secret management for the TRENTO_DEPLOY_VER part, L89 can be removed as figuring out how to use the env variable in the else

lib/trento.pm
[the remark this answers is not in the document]
mpagot: Solved using Mojo::JSON::decode_json

lib/trento.pm
```diff
@@ -63,12 +68,95 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
 # Azure does not support dash or underscore in ACR name
 use constant TRENTO_AZ_ACR_PREFIX => 'openqatrentoacr';
 use constant CYPRESS_IMAGE_TAG => 'goofy';
+use constant CYPRESS_IMAGE => 'docker.io/cypress/included';
 
 =head1 DESCRIPTION 
 
 Package with common methods and default or constant  values for Trento tests
```
lilyeyes, line 75: One redundant "space" between "constant" and "values".

lib/trento.pm
lilyeyes: Why not use `type_password $gitlab_token ...` instead of `type_password get_var(TRENTO_GITLAB_TOKEN => ...` in line 91?

mpagot: I like to try to remove line 89 at all in some next commits. Problem is how to manage the secret in the else from line 146

lib/trento.pm
```diff
@@ -63,12 +68,95 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+          " --output $ver_artifact ";
+        assert_script_run($download_cmd);
+
+        # Extract and test file presence
+        my $tar_cmd = "tar xvf $ver_artifact --strip-components=1";
+        assert_script_run($tar_cmd);
+        enter_cmd 'ls -lai';
+    }
+
+    # script from gitlab
```
lilyeyes, line 145: For coding style consistent: `s/# script/# Script/` in line 145

lib/trento.pm
```diff
@@ -149,8 +263,8 @@ sub az_vm_ssh_cmd {
     my ($self, $cmd_arg, $vm_ip_arg) = @_;
     record_info('REMOTE', "cmd:$cmd_arg");
 
-    # undef comparision operator
-    my $vm_ip = $vm_ip_arg // az_get_vm_ip();
+    # undef comparison operator
```
lilyeyes, line 266: s/# undef/# Unfef/

lib/trento.pm
alvarocarvajald: Seems like in the block below L146 the only thing done with the token is to check whether it is the same as the one supplied. As such, one option would be to get rid of the check, and assume that the token stored in the qcow2 is the correct one. The other option would be to move the check to the shell, as you have the token already stored in a shell environment variable; for example you could check with:
```
assert_script_run '[[ "$GITLAB_TOKEN" == "$(git config --get remote.origin.url | cut -d@ -f1|cut -d: -f3)" ]]'
```
If they are the same, the above command will return 0 and pass, otherwise it will return 1 and `assert_script_run()` will fail the test.

lib/trento.pm
```diff
@@ -63,12 +68,96 @@ use constant TRENTO_AZ_PREFIX => 'openqa-trento';
...
+        # Get some 'coordinates' from the test settings
+        my $ver = get_var('TRENTO_DEPLOY_VER');
+        my $gitlab_repo = get_var(TRENTO_GITLAB_REPO => 'gitlab.suse.de/qa-css/trento');
+        my @gitlab_url = split('/', $gitlab_repo);
+        my $gitlab_namespace = $gitlab_url[-2];
+        my $gitlab_project = $gitlab_url[-1];
+
+        assert_script_run('echo "PRIVATE-TOKEN: ${GITLAB_TOKEN}" > gitlab_conf');
+        my $curl_cmd = 'curl -s ' .
+          '-H @gitlab_conf';
```
alvarocarvajald, line 110: There is not really a reason to have here a concatenation operation between two strings if both are constants, right? You could directly do `my $curl_cmd = 'curl -s -H @gitlab_conf';`

lib/trento.pm
```diff
@@ -149,8 +264,8 @@ sub az_vm_ssh_cmd {
     my ($self, $cmd_arg, $vm_ip_arg) = @_;
     record_info('REMOTE', "cmd:$cmd_arg");
 
-    # undef comparision operator
-    my $vm_ip = $vm_ip_arg // az_get_vm_ip();
+    # Undef comparison operator
+    my $vm_ip = $vm_ip_arg // get_trento_ip();
```
alvarocarvajald, line 268: You could also do:
```
$vm_ip_arg //= get_trento_ip();
```
And that way you avoid having 2 variables (`$vm_ip` and `$vm_ip_arg`) which basically have the same info.

alvarocarvajald (commented):
LGTM.

Just a couple of comments, but nothing major.

lib/trento.pm
mpagot: I like to split them to facilitate the code reader to understand different sections of the command

alvarocarvajald (approved):

PR #15421, opened 2022-08-26
Wait for still screen after pressing ok 

ilmanzo:
yast2_security test module fills in login settings and press ok, then immediately launch another YaST modules, while the system is still applying the configuration. 

This change introduces a small delay so the screen has time to stabilize before opening the next configuration item.

- Verification run: https://openqa.suse.de/tests/9392980#

### DIFF
--- a/tests/yast2_gui/yast2_security.pm
+++ b/tests/yast2_gui/yast2_security.pm
@@ -50,6 +50,7 @@ sub run {
     wait_still_screen 1;
     wait_screen_change { type_string "5" };
     wait_screen_change { send_key "alt-o" };
+    wait_still_screen 3;
 
     # Check previously set values + Miscellaneous Settings
     y2_module_guitest::launch_yast2_module_x11("security", match_timeout => 120);


tests/yast2_gui/yast2_security.pm
```diff
@@ -50,6 +50,7 @@ sub run {
     wait_still_screen 1;
     wait_screen_change { type_string "5" };
     wait_screen_change { send_key "alt-o" };
+    wait_still_screen;
```
jknphy, line 53: default stilltime is 7s, could we narrow it a bit?

jknphy:
for the command: s/Yast2 security/yast2 security/
for the team: YaST
It took for me a time to write it well afir :)

tests/yast2_gui/yast2_security.pm
ilmanzo: yes I left the default to be sure, let me run another job with half ( 3 sec ? ) and see

PR #15425, opened 2022-08-29
syscall_thrasher.pm: retry on trash failure

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/115754 and https://progress.opensuse.org/issues/118360
- Verification run: https://openqa.suse.de/tests/9743022

### DIFF
--- a/tests/security/atsec/syscall_thrasher.pm
+++ b/tests/security/atsec/syscall_thrasher.pm
@@ -38,7 +38,10 @@ sub run {
 
     # Run the test
     assert_script_run("cd $test_dir");
-    assert_script_run("./$exe_file", timeout => 1200);
+
+    # Since this randomly fails, we retry 10 times each time adding a delay before failing and exiting.
+    script_retry("./$exe_file", retry => 10, timeout => 1200, delay => 60);
+
     assert_script_run('ls');
 }
 


punkioudi:
Thanks Paolo for the PR! :)
Did you try to increase ram/cpu from the settings and see the behavior? If again nothing changes, could you try and use the `script_retry` function? It includes timeout,retry and delay for cases like this :)

paolostivanin:
Hi @punkioudi , thanks for the review :)
To answer your points:
1. no, I didn't try to increase the cpu/ram. I know I can increase the RAM with QEMURAM, but is there a way to increase also the # of vcpus?
2. I was not aware of `script_retry`, and I also can't find it here https://open.qa/api/testapi . Is this an official api?

Thanks :smile:

punkioudi:
> Hi @punkioudi , thanks for the review :) To answer your points:
> 
>     1. no, I didn't try to increase the cpu/ram. I know I can increase the RAM with QEMURAM, but is there a way to increase also the # of vcpus?
> 
>     2. I was not aware of `script_retry`, and I also can't find it here https://open.qa/api/testapi . Is this an official api?
> 
> 
> Thanks smile

1. For the cpu you can add in the settings QEMUCPUS (you can try for example for here QEMUCPUS=4)
2. Well, I am not sure exactly why it is not in the official page of osd, probably mainly because people were split on if we should use it or not. In general, the workarounds with sleep should be avoided and `script_retry` was made for these edge cases that we really need to give some time in order for the test to pass (and of course when there is no product bug). You can see many examples in the tests folder and imho it's better that we all use a common workaround function for these kind of cases :) 

I hope this helps :D

paolostivanin:
cool Anna, thanks a lot :smile: 
So I'll keep this on WIP and change the resources first. If we don't see any future failure, I'll close this. Otherwise, we'll go with the retry workaround.

PR #15431, opened 2022-08-29
Validate output from pdsh

b10n1k:
Test doesnt validate whether the file contains any context or not. So this
patch perform a minimal validation that indicates that the file is not
empty. I dont add checks for the actual context as i am not of its impoirtance
in this case

- Related ticket: https://progress.opensuse.org/issues/115928
- Verification run: http://aquarius.suse.cz/tests/12102

### DIFF
--- a/tests/hpc/pdsh_slave.pm
+++ b/tests/hpc/pdsh_slave.pm
@@ -36,6 +36,7 @@ sub run ($self) {
     $self->switch_user('nobody');
     my $genders_plugin = get_var('PDSH_GENDER_TEST') ? '-g type=genders-test' : '';
     assert_script_run("pdsh -R mrsh $genders_plugin -w $server_hostname ls / &> /tmp/pdsh.log");
+    assert_script_run("test -s /tmp/pdsh.log");
     upload_logs '/tmp/pdsh.log';
     barrier_wait("PDSH_SLAVE_DONE");
 }


tests/hpc/pdsh_slave.pm
```diff
@@ -36,6 +36,7 @@ sub run ($self) {
     $self->switch_user('nobody');
     my $genders_plugin = get_var('PDSH_GENDER_TEST') ? '-g type=genders-test' : '';
     assert_script_run("pdsh -R mrsh $genders_plugin -w $server_hostname ls / &> /tmp/pdsh.log");
+    assert_script_run("test -s /tmp/pdsh.log");
```
schlad, line 39: ah OK, I did not think of using it this way for checking the output file.
Would it make sense to do the same for the MPI tests? Some time back those produced a file with 'Hello world' printed from an MPI basic application to a  file, so perhaps this test -s trick would be good there too?

schlad (approved):

tests/hpc/pdsh_slave.pm
b10n1k: sure. i review other modules as well and i will do so in small chucks. To answer more specific i would have to check how the behavior is MPI again. Also this approach is very minimalistic. We might think a better solution later on

schlad: yeah definitely; specifically for the MPI jobs parsing of the output would be definitely better idea. On my TODO list for the MPI there was an idea of running way more sophisticated/real-life MPI application and there the parsing of the output would be imo crucial.

PR #15434, opened 2022-08-29
ui-framework-documentation.md: explain pros and cons of older vs newer approach

ilmanzo:
Rephrase and explain the motivations behind the adoption of Page-Object design pattern and declarative scheduling vs the legacy test style

- Related ticket: https://progress.opensuse.org/issues/106936


jknphy:
Very nice! thanks a lot for this update, very well written!

PR #15436, opened 2022-08-29
Add library for RPI

pevik:
ATM containing only single method.

Verification run:
* **https://openqa.opensuse.org/tests/2584395 opensuse-Tumbleweed-JeOS-for-RPi-aarch64-Buildpevik_os-autoinst-distri-opensuse_lib-rpi-jeos-ltp-ima@pevik_os-autoinst-distri-opensuse_lib-rpi@RPi4**
NOTE: error on https://openqa.opensuse.org/tests/2584395#step/ima_tpm/8 is bug found by LTP test, which is visible due this change (good we found a problem in OS => no need to hide that)

Verifications non-RPI tests still work:
* https://openqa.opensuse.org/tests/2580422 opensuse-Tumbleweed-DVD-x86_64-Buildpevik_os-autoinst-distri-opensuse_lib-rpi-security_swtpm@pevik_os-autoinst-distri-opensuse_lib-rpi@64bit
* https://openqa.opensuse.org/tests/2580421 opensuse-Tumbleweed-DVD-x86_64-Buildpevik_os-autoinst-distri-opensuse_lib-rpi-security_swtpm_uefi@pevik_os-autoinst-distri-opensuse_lib-rpi@64bit

@STARRY-S do we run on o3?
@ggardet FYI

### DIFF
--- a/lib/rpi.pm
+++ b/lib/rpi.pm
@@ -1,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+package rpi;
+
+use warnings;
+use strict;
+
+use Exporter 'import';
+use testapi;
+use power_action_utils 'power_action';
+
+our @EXPORT = 'enable_tpm_slb9670';
+
+# Enable TPM on Raspberry Pi
+# https://en.opensuse.org/HCL:Raspberry_Pi3_TPM
+sub enable_tpm_slb9670 {
+    my $module = "tpm_tis_spi";
+
+    assert_script_run('echo -e "dtparam=spi=on\ndtoverlay=tpm-slb9670" >> /boot/efi/extraconfig.txt');
+    assert_script_run('cat /boot/efi/extraconfig.txt');
+    assert_script_run("echo '$module' > /etc/modules-load.d/tpm.conf");
+    power_action('reboot', textmode => 1);
+
+    # Add some timeout to wait for reboot
+    sleep(60);
+
+    # Restore SSH connection
+    reset_consoles;
+    select_console('root-ssh');
+
+    record_info('RPi TPM', script_output("dmesg | grep '$module.*2.0 TPM.*device-id.*rev-id'"));
+}
+
+1;
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -23,6 +23,7 @@ use version_utils qw(is_jeos is_opensuse is_released is_sle is_leap is_tumblewee
 use Utils::Architectures;
 use Utils::Systemd qw(systemctl disable_and_stop_service);
 use LTP::utils;
+use rpi 'enable_tpm_slb9670';
 
 sub add_we_repo_if_available {
     # opensuse doesn't have extensions
@@ -356,6 +357,8 @@ sub run {
     if (!get_var('KGRAFT') && !get_var('LTP_BAREMETAL') && !is_jeos) {
         $self->wait_boot;
     }
+
+    enable_tpm_slb9670 if (get_var('MACHINE') =~ /RPi/);
 
     $self->select_serial_terminal;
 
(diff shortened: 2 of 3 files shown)


lib/rpi.pm
```diff
@@ -0,0 +1,36 @@
...
+    power_action('reboot', textmode => 1);
+
+    # Add some timeout to wait for reboot
+    sleep(60);
+
+    # Restore SSH connection
+    reset_consoles;
+    select_console('root-ssh');
+
+    assert_script_run('dmesg | grep tpm');
```
ggardet, line 32: Maybe you could be more specific than only _tpm_? To ensure this is really the _slb9670_.

pevik: I copy pasted this code, but indeed this is needed. Because for QEMU and RPI without this setup there is info from IMA which would be also valid for the original grep:

`ima: No TPM chip found, activating TPM-bypass!`

STARRY-S:
Hi, I do not know the current status of this test case on O3 after I finished my internship, I've asked @rfan1 for help.
Thanks for your PR.

pevik:
@rfan1 FYI

ggardet (approved):
LGTM. Thanks!

lib/rpi.pm
```diff
@@ -0,0 +1,37 @@
...
+sub enable_tpm_slb9670 {
+    my $module = "tpm_tis_spi";
+
+    assert_script_run('echo -e "dtparam=spi=on\ndtoverlay=tpm-slb9670" >> /boot/efi/extraconfig.txt');
+    assert_script_run('cat /boot/efi/extraconfig.txt');
+    assert_script_run("echo '$module' > /etc/modules-load.d/tpm.conf");
+    power_action('reboot', textmode => 1);
+
+    # Add some timeout to wait for reboot
+    sleep(60);
```
b10n1k, line 28: i know that this is moved here and it is not coming from you but i am wondering if it would be better to replace it with `wait_serial`? wdyt?

pevik: I can try that.

mimi1vx: not wait serial but something like `wait_boot` should be here

rfan1:
> @rfan1 FYI

Thanks for the kindly reminder, @paolostivanin @tjyrinki can you please help review and update?

tjyrinki (approved):
Looks like a good replacement to me, although I see there are other reviews waiting for changes still.

paolostivanin:
LGTM

pevik:
FYI why there is no further update: jobs with '@RPi4' are scheduled but not run for nearly one day (https://openqa.opensuse.org/tests/2586266).

ggardet:
> FYI why there is no further update: jobs with '@rpi4' are scheduled but not run for nearly one day (https://openqa.opensuse.org/tests/2586266).

The server was down. It is up and running again.

lib/rpi.pm
pevik: Well, since `wait_boot` is skipped for JeOS

https://github.com/pevik/os-autoinst-distri-opensuse/blob/5cbccd4f49840ef36cf07bff474c03c83ad90b61/tests/kernel/install_ltp.pm#L358-L360

and none of these helped:
```
$self->wait_boot(nologin => 1, bootloader_time => 60);
$self->wait_boot;
```
Because `wait_boot()` uses `assert_screen()`

https://github.com/pevik/os-autoinst-distri-opensuse/blob/5cbccd4f49840ef36cf07bff474c03c83ad90b61/lib/opensusebasetest.pm#L1175-L1187. I suppose we need some support for RPI in openQA, but that's out of my scope.

ggardet: I think the `sleep` is the only option here since it is a headless system, reached via ssh only.

pevik: Given that I suggest to merge. @b10n1k @mimi1vx @tjyrinki ?

b10n1k: let s move on

b10n1k (approved):
+1

pevik:
Fix was needed: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/18352

PR #15446, opened 2022-08-31
Fix systemtap test in 12-SP3

DeepthiYV:
- Related ticket: https://progress.opensuse.org/issues/115994
- Needles: No
- Verification run:
   https://openqa.suse.de/tests/9419005#


foursixnine (approved):
@DeepthiYV can you create a follow up ticket to extract LTSS management for customers?

We would be basically refactoring LTSS Setup so that it only lives in a single place.

PR #15453, opened 2022-08-31
Fix InfiniBand tests

frankenmichl:
Make sure we have less flaky InfiniBand tests and fix the timeouts issue
before actual test execution (due to missing handle_reboot module).

Verification run: http://baremetal-support.qa.suse.de/tests/2302 and http://baremetal-support.qa.suse.de/tests/2303

### DIFF
--- a/tests/kernel/ibtests.pm
+++ b/tests/kernel/ibtests.pm
@@ -77,12 +77,12 @@ sub ibtest_master {
 
 
     # pull in the testsuite
-    assert_script_run("git clone $hpc_testing --branch $hpc_testing_branch");
+    assert_script_run("git clone $hpc_testing --branch $hpc_testing_branch", timeout => $timeout);
 
     # wait until the two machines under test are ready setting up their local things
     assert_script_run('cd hpc-testing');
     barrier_wait('IBTEST_BEGIN');
-    script_run("./ib-test.sh $args $master $slave", $timeout);
+    script_run("./ib-test.sh $args $master $slave", timeout => $timeout);
     script_run('tr -cd \'\11\12\15\40-\176\' < results/TEST-ib-test.xml > /tmp/results.xml');
     parse_extra_log('XUnit', '/tmp/results.xml');
 
--- a/tests/kernel/ibtests_prepare.pm
+++ b/tests/kernel/ibtests_prepare.pm
@@ -17,19 +17,6 @@ use lockapi;
 use version_utils;
 use mmapi;
 
-
-sub permit_root_login {
-    my $self = shift;
-
-    select_console('sol', await_console => 1);
-    type_string "root";
-    wait_screen_change { type_string "\n" };
-    type_password;
-    wait_screen_change { type_string "\n" };
-    permit_root_ssh_in_sol;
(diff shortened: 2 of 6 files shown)


frankenmichl:
http://baremetal-support.qa.suse.de/tests/2308 and http://baremetal-support.qa.suse.de/tests/2309 are the latest verification runs.

pevik (approved):
I'd say timeout `0` looks to me quite unusual (why needed even for `clear`), but in the worst case it will just take long time to test fail.

tests/kernel/ibtests_prepare.pm
```diff
@@ -40,9 +27,9 @@ sub run {
     my $packages_master = $packages . " git-core twopence-shell-client bc";
 
 
-    $self->permit_root_login if is_ipmi;
-
     $self->select_serial_terminal;
+    permit_root_ssh;
```
b10n1k, line 31: i took a look to the utils.pm. I would expect `permit_root_ssh_in_sol` here instead of `permit_root_ssh`. those two looks to do quite the same in a different way. Would it make sense to delete `permit_root_ssh_in_sol`(or merge)? However that sounds like another PR/commit. I put it here as a note.

tests/kernel/ibtests.pm
```diff
@@ -127,7 +127,7 @@ sub post_fail_hook {
     my $role = get_required_var('IBTEST_ROLE');
 
     if ($role eq 'IBTEST_MASTER') {
-        script_run('tr -cd \'\11\12\15\40-\176\' < results/TEST-ib-test.xml > /tmp/results.xml');
+        script_run('tr -cd \'\11\12\15\40-\176\' < results/TEST-ib-test.xml > /tmp/results.xml', 0);
```
b10n1k, line 130: have you forgotten `timeout =>` before 0

tests/kernel/ibtests.pm
```diff
@@ -77,13 +77,13 @@ sub ibtest_master {
 
 
     # pull in the testsuite
-    assert_script_run("git clone $hpc_testing --branch $hpc_testing_branch");
+    assert_script_run("git clone $hpc_testing --branch $hpc_testing_branch", timeout => 3600);
```
b10n1k, line 80: i guess you can reuse `$timeout` here (its default value is 3600)

b10n1k (changes requested):
just for the record. Why do you need timeout=>0 for all the `script_run`?

tests/kernel/ibtests_prepare.pm
frankenmichl: Right, at least for now permit_root_ssh_in_sol should be fine, too. I changed that and give it a try.

tests/kernel/ibtests.pm
frankenmichl: I removed all timeouts of 0

b10n1k (approved):
thanks for the changes

pevik:
@frankenmichl Could you please update also the commit message? It still mentions `permit_root_ssh`.

pevik (approved):
Thx!

PR #15456, opened 2022-09-01
Move trufflehogs checks to its own branch

foursixnine:
We know already the problem with Trufflehog, so let's wait for resolution before doing anything else.


michaelgrifalconi (approved):

foursixnine:
@michaelgrifalconi we can keep an eye on: https://github.com/trufflesecurity/trufflehog/issues/666

PR #15457, opened 2022-09-01
Import WE registration codes during installation

tinawang123:
Import WE registration codes during installation

- Related ticket: https://progress.opensuse.org/issues/115367
- Verification run: https://openqa.suse.de/tests/9426489#


tests/autoyast/installation.pm
```diff
@@ -303,6 +304,9 @@ sub run {
         elsif (match_has_tag 'expired-gpg-key') {
             send_key 'alt-y';
         }
+        elsif (match_has_tag 'registration-codes') {
+            send_key 'alt-n';
```
jknphy, line 308: this screen https://openqa.suse.de/tests/9427059#step/installation/6
will not be here when the NVIDIA repo will exist:
https://openqa.suse.de/tests/8753030#step/installation/5
We should process all this 'workaround' flow above in the same `elseif` I mean and you don't need to add it to `@needles` it is expected that one thing is after the other in this case.

tests/autoyast/installation.pm
```diff
@@ -289,9 +289,10 @@ sub run {
         }
         elsif (match_has_tag('nvidia-validation-failed')) {
             # nvidia repositories are unstable and really not needed for anything
-            record_info("nVidia", "nVidia repository is broken");
+            record_info("NVIDIA", "NVIDIA repository is broken");
             send_key 'alt-y';
```
jknphy, line 293: there is only OK button, I don't think we need here to press 'Yes'.

tests/autoyast/clone.pm
```diff
@@ -18,6 +18,8 @@ use utils qw(zypper_call);
 sub run {
     my $self = shift;
     assert_script_run 'rm -f /root/autoinst.xml';
+    assert_script_run 'cd /etc/zypp/repos.d';
+    assert_script_run "sed -i 's/enabled=1/enabled=0/' \$(ls|grep -i nvidia)";
```
jknphy, line 22: does it fail when cloning? can I see the run?
we don't plan to use the system afterwards, so if a repo that doesn't exist is not there is ok.

tinawang123: Failed at clone test module: https://openqa.suse.de/tests/9426602#step/clone/3

tests/autoyast/installation.pm
tinawang123: I'm not sure why it press 'Yes' at here. Maybe other case need this one. It was written by others.

jknphy: but the needle didn't match in OSD so we are supposed to do whatever we want here, nobody is using it (I hope :) )
`alt-y` probably was there because there was a different dialog with a `yes`.

tests/autoyast/clone.pm
jknphy: but then we should do this in the AutoYaST profile as you were doing at the begining? otherwise any module schedule would fail unless that code is there. is it possible do the post_script in autoyast after having clicking ok on the error dialog?

tinawang123: I did use post_script, it also doesn't work. You can see: https://openqa.suse.de/tests/9426602/file/autoyast_home_encrypted.xml
I think it cannot disable NVIDIA at first, then clone it will include this NVIDIA repo.

jknphy: we should give a soft-failure here pointing to a bug (no answer from developers yet, so better create the bug). But we should soft-fails always, only if nvidia is found, that `ls|grep -i nvidia` should serve us as condition, could you please reshape it a bit, then should be ready to merge IMO.

tinawang123: Use 'init-script' can fix this problem.

PR #15462, opened 2022-09-01
Replace wrong assert_script_run function usage

BillAnastasiadis:
In `verify_btrfs_quota.pm`, `assert_script_run` was used to get the output of 2 commands. That function does not collect the output, but the test did not catch that and passed. Here, it was replaced with `script_output` which collects command output.

- Related ticket: https://progress.opensuse.org/issues/115985
- Verification run: https://openqa.suse.de/tests/9427431#step/verify_btrfs_quota/2

See the difference (under `Result`) with the previous, wrong case: https://openqa.suse.de/tests/9313739#step/verify_btrfs_quota/2

### DIFF
--- a/tests/autoyast/verify_btrfs_quota.pm
+++ b/tests/autoyast/verify_btrfs_quota.pm
@@ -19,9 +19,9 @@ sub run {
     foreach my $disk (@{$test_data->{disks}}) {
         foreach my $partition (@{$disk->{partitions}}) {
             my $mount_point = $partition->{mounting_options}->{mount_point};
-            my $output_btrfs_subvolumes = assert_script_run("btrfs subvolume list $mount_point");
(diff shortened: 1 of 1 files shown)


ge0r:
I don't get the VR difference, don't the two last jobs have the same result in [verify_btrfs_quota](https://openqa.suse.de/tests/9418132/modules/verify_btrfs_quota/steps/1/src) `Result: rdoJR-0-`?

ge0r:
> I don't get the VR difference, don't the two last jobs have the same result in [verify_btrfs_quota](https://openqa.suse.de/tests/9418132/modules/verify_btrfs_quota/steps/1/src) `Result: rdoJR-0-`?

Ok I missed the link above... LGTM!

ge0r (approved):

pevik:
Thx for fixing, It'd be worth to have way to detect this wrong usage in CI.

PR #15469, opened 2022-09-02
Add a test suite for migration over maintenance updates

JRivrain:
See https://progress.opensuse.org/issues/113647
related MR https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/364

There is an issue that system gets de-registered the next day in the afternoon, and we cannot register it again. I tried many different hacks to work around it but none works. I don't think it is worth the effort as anyway, the test suite works otherwise. Just like all other maintenance update test suites, this one cannot be re-launched too long after the build.

VRs
https://openqa.suse.de/tests/9475427

### DIFF
--- a/lib/migration.pm
+++ b/lib/migration.pm
@@ -29,6 +29,7 @@ our @EXPORT = qw(
   reset_consoles_tty
   set_scc_proxy_url
   set_zypp_single_rpmtrans
+  remove_dropped_modules_packages
 );
 
 sub setup_sle {
@@ -129,6 +130,17 @@ sub deregister_dropped_modules {
         @all_addons = grep { $_ ne $name } @all_addons;
     }
     set_var('SCC_ADDONS', join(',', @all_addons));
+}
+
+# This function removes the packages specified with variable DROPPED_MODULES instead of un-registering the modules
+# before migrating: it is enough to just remove the -release package to avoid warnings like "the module
+# cannot be updated".
+sub remove_dropped_modules_packages {
+    my $droplist = get_var('DROPPED_MODULES', '');
+    for my $name (split(/,/, $droplist)) {
+        my $release_package = get_addon_fullname($name) . "-release";
+        zypper_call("rm $release_package");
+    }
 }
 
 # Disable installation repos before online migration
--- a/schedule/yast/maintenance/autoupgrade_sle15sp3+1_scc_we_sdk_basesys_wsm_srv_desk_lgm_contm_pcm_def_full.yaml
+++ b/schedule/yast/maintenance/autoupgrade_sle15sp3+1_scc_we_sdk_basesys_wsm_srv_desk_lgm_contm_pcm_def_full.yaml
@@ -1,0 +1,26 @@
+---
+name: autoupgrade_sle15sp3+1_scc_we_sdk_basesys_wsm_srv_desk_lgm_contm_pcm_def_full
+description: >
+  Performs a migration between installed system with unreleased maintenance updates installed, and next release, eg sle15 sp3>sp4
+schedule:
+  - autoyast/prepare_profile
+  - migration/version_switch_origin_system
+  - boot/boot_to_desktop
+  - console/check_system_info
+  - migration/remove_dropped_modules_packages
+  - migration/reboot_to_upgrade
+  - migration/version_switch_upgrade_target
+  - installation/bootloader
+  - autoyast/installation
+  - autoyast/console
+  - autoyast/login
+  - autoyast/wicked
+  - autoyast/repos
+  - autoyast/clone
+  - autoyast/logs
+  - autoyast/autoyast_reboot
+  - installation/grub_test
+  - installation/first_boot
+  - boot/grub_test_snapshot
+  - migration/version_switch_origin_system
+  - boot/snapper_rollback
--- a/tests/migration/remove_dropped_modules_packages.pm
+++ b/tests/migration/remove_dropped_modules_packages.pm
@@ -1,0 +1,22 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Register again is system is un-registered on SCC side, then
+# remove -release packages from dropped modules before installation to avoid related warnings.
+
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use migration;    #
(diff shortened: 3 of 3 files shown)


lib/registration.pm
```diff
@@ -231,6 +231,12 @@ sub remove_suseconnect_product {
     $version //= scc_version();
     $arch //= get_required_var('ARCH');
     $params //= '';
+    # De-registration fails when module was already de-registered the same day on
+    # the same qcow image (eg if we re-launch the test suite several times...)
+    # so we need to register it again if not registered (if = unless as bash returns 1 = false)
+    if (script_run("SUSEConnect --status-text |grep -a3 python2 |grep Registered")) {
+        script_run("SUSEConnect -p $name/$version/$arch $params");
```
jknphy, line 238: this library is widely used, not sure about to change the behavior for all calls...
probably we should do this from outside having a function that give us the status (if doesn't exist in this lib?)
it should be `add_suseconnect_product` so we could still make use of the retries I guess.

schedule/yast/maintenance/autoyast_migration_on_maintenance_updates.yaml
```diff
@@ -0,0 +1,26 @@
+---
+name: autoyast_migration_on_maintenance_updates
+description: >
+  Perform a migration between installed system with maintenance updates installed, and next release, eg sle15 sp3>sp4
```
jknphy, line 4: s/Perform/Performs/
s/maintenance updates/unreleased maintenance updates/ (I would like to do this distinction)

jknphy:
could you please reference the MR in first comment as usual?

lib/registration.pm
JRivrain: You're right, this should not be here, this was kinda POC to see what we can do, I should move this elsewhere. I think we will not use zypper_patch and create our own module here.

jknphy: @lemon-suse any suggestion where to move this?

lemon-suse: What do you think to add condition to position where invoking this function? 
Ex:
remove_suseconnect_product if module registered

If considered it is hard to judge the status of module, just for special ones then can add a setting to control it as a workaround.

schedule/yast/maintenance/autoyast_migration_on_maintenance_updates.yaml
JRivrain: Perform versus performs: both are correct, the symbol ":" introducing an explanation means we are continuing a sentence, and normally don't need an uppercase letter next, eg "(the object of the module is [to]) perfom" vs "(this) performs".
Then I agree for the next point :)

jknphy: makes sense, I have seen both ways, but never sure if was really ok, thanks!

tests/migration/prepare_addons_for_migration.pm
```diff
@@ -0,0 +1,27 @@
...
+use warnings;
+use testapi;
+use migration 'remove_dropped_module_packages';
+use registration 'register_addons_cmd';
+
+
+sub run {
+    unless (script_run("SUSEConnect --status-text |grep -a3 basesystem |grep 'Not Registered'")) {
+        assert_script_run("SUSEConnect --cleanup");
+        assert_script_run("SUSEConnect -p sle-module-basesystem/15.3/x86_64");
```
jknphy, line 21: we should use the library to avoid writing code for specific arch.

tests/migration/prepare_addons_for_migration.pm
```diff
@@ -0,0 +1,27 @@
...
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use migration 'remove_dropped_module_packages';
+use registration 'register_addons_cmd';
+
+
+sub run {
+    unless (script_run("SUSEConnect --status-text |grep -a3 basesystem |grep 'Not Registered'")) {
```
jknphy, line 19: - `unless` is only used in certain programming languages with the premise the condition is easy to read, in general some function or variable, but this piped command is not the case.

- this is to check if something is not register first then register it, I guess, we should use some lib function for it, is the only way to check for the basesystem or is there some more generic way?

lib/migration.pm
```diff
@@ -131,6 +132,16 @@ sub deregister_dropped_modules {
     set_var('SCC_ADDONS', join(',', @all_addons));
 }
 
+# We don't always need to un-register a module before migrating, it is enough to just remove the -release package
+# to avoid warnings that the module cannot be updated.
```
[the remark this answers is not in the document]
JRivrain: That is what the comment says, we only need to remove the package... unlike what the already existing code does. it's much more easy to remove the package and does not alter the registration status.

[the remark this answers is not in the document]
JRivrain: I will update description to make it more clear

lib/migration.pm
```diff
@@ -131,6 +132,16 @@ sub deregister_dropped_modules {
...
+sub remove_dropped_module_packages {
```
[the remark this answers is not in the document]
JRivrain: modified

tests/migration/prepare_addons_for_migration.pm
JRivrain: I gave up on this, whatever we do, at some point the system is de-registered entirely and the registration code expires, there is no workaround unless we change that behgavior on SCC server's side: more or less after 4pm no maintenance job from previous day can be re-launched. I removed the workaround as it does not work anyway.

tests/migration/prepare_addons_for_migration.pm
JRivrain: see previous comment, this was more of an experiment, I then wrote it better in a lib, but anyway it fails because of scc, so I dropped that altogether, the only solution is to make sure we don't launch the test after 4pm (more or less).

jknphy:
@JRivrain what would be missing, only to update the MR with the new schedule? once we see it in dev group we can check if something was missing.

tests/migration/prepare_addons_for_migration.pm
```diff
@@ -0,0 +1,20 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Make sure all addons are registered then remove -release packages from dropped modules before installation
```
jknphy, line 6: let's update this then.

jknphy:
Let's merge in development and we can continue figuring out additional issues you found.

PR #15429, opened 2022-08-29
Containers: Fix kubectl test

pdostal:
- Related ticket: [poo#114457](https://progress.opensuse.org/issues/114457)
- Verification run: [Tumbleweed](https://openqa.opensuse.org/tests/2591351), [SLE15-SP1 x86_64](https://openqa.suse.de/tests/9418701), [SLE15-SP3 aarch64](https://openqa.suse.de/tests/9418637),

PR #15468, opened 2022-09-02
Fix K3S helm test

ilausuch:
In some cases, k3s needs some time to create a pod from the job.
In this commit, the test waits some time after creating the helm

- Related ticket: https://progress.opensuse.org/issues/116161
- Verification run: https://openqa.opensuse.org/tests/2621989

PR #15472, opened 2022-09-04
Fix slow k3s start

pdostal:
The `k3` is starting for quite a some time so we need to check that at least one node of the k8s cluster is `Ready` before we continue with the testing.

```
kubectl get node
NAME       STATUS     ROLES    AGE   VERSION
susetest   NotReady   <none>   0s    v1.24.4+k3s1

k3s kubectl get node
NAME       STATUS   ROLES                  AGE   VERSION
susetest   Ready    control-plane,master   15s   v1.24.4+k3s1
```

As this is quite small fix I think that one reviewer should be enough to merge this.

- Related request: #15429
- Verification run: [pdostal-server.suse.cz/t535](https://pdostal-server.suse.cz/tests/535)

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -14,7 +14,7 @@ use Exporter;
 use strict;
 use warnings;
 use testapi;
-use utils qw(zypper_call script_retry file_content_replace);
+use utils qw(zypper_call script_retry file_content_replace validate_script_output_retry);
 use Utils::Systemd qw(systemctl);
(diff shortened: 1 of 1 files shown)


jlausuch:
There was this fix as well from @ilausuch https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15468
Which one should we merge?

jlausuch (approved):
Both approaches are valid and complementary, merging.

PR #15473, opened 2022-09-04
Add work around & soft fail for poo#116191

alvarocarvajald:
SLES+HA migrations from 15-SP3 to 15-SP4 & 15-SP5 on ppc64le fail on the `ha/check_after_reboot` test module because the UUID configured for the `cluster_md` MD RAID device in `/etc/mdadm.conf` in the qcow2 images from the 15-SP3 installations is different than the actual block devices UUIDs after the migration. This commit works around the issue and soft fail the test when the issue is detected.

- Related ticket: https://progress.opensuse.org/issues/116191
- Needles: N/A
- Verification runs: [support server](https://openqa.suse.de/tests/9458942), [node 1](https://openqa.suse.de/tests/9458943), [node 2](https://openqa.suse.de/tests/9458941)
- Verification run (regression): Alpha Cluster: [node 1](http://mango.qa.suse.de/tests/4945), [node 2](http://mango.qa.suse.de/tests/4946) & [support server](http://mango.qa.suse.de/tests/4946); 15-SP3 to 15-SP5 Migrations on x86_64: [node 1](http://mango.qa.suse.de/tests/4948), [node 2](http://mango.qa.suse.de/tests/4949) & [support server](http://mango.qa.suse.de/tests/4947)

### DIFF
--- a/schedule/ha/bv/migration/migration_online_sles_ha.yaml
+++ b/schedule/ha/bv/migration/migration_online_sles_ha.yaml
@@ -26,6 +26,7 @@ schedule:
   - migration/version_switch_origin_system
   - '{{bootloader}}'
   - migration/online_migration/online_migration_setup
+  - ha/stop_pacemaker
   - migration/online_migration/register_system
   - migration/online_migration/zypper_patch
   - '{{remove_ltss}}'
--- a/tests/ha/check_after_reboot.pm
+++ b/tests/ha/check_after_reboot.pm
@@ -16,7 +16,7 @@ use Utils::Architectures;
 use lockapi;
 use hacluster;
 use version_utils qw(is_sles4sap);
-use utils qw(systemctl);
+use utils qw(systemctl file_content_replace);
 
 sub run {
     my $cluster_name = get_cluster_name;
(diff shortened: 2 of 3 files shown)


lilyeyes:
I only checked "stop_pacemaker" related code changes, it looks good to me.
For other code changes I don't have knowledge on this yet but generally looks good to me.

bs-github:
I like this very much!
LGTM

lpalovsky (approved):
LGTM

ricardobranco777 (approved):
LGTM

alvarocarvajald:
> I only checked "stop_pacemaker" related code changes, it looks good to me. For other code changes I don't have knowledge on this yet but generally looks good to me.

This is only intended to stop pacemaker during migrations, as migration jobs do not run in Multi-Machine mode, hence cluster nodes do not have access to the support server, and cluster is not working. Intention is to get rid of those pacemaker failures journal messages from reaching the console and causing a needle match failure during the test (such as in [this case](http://mango.qa.suse.de/tests/4912#step/register_system/4)). I added it only in the online migrations schedule for now, as I ran into it during the verification runs, but it may be necessary to add it to the other migration schedules. 

Module should not be used in migration verification jobs as we need pacemaker active in those.

PR #15475, opened 2022-09-05
Add is_sle_micro for open ssh checking

waynechen55:
* **Add** is_sle_micro for open ssh checking

### DIFF
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -25,7 +25,7 @@ sub ensure_ssh_unblocked {
 
         # ssh section is not shown up directly in text mode. Navigate into
         # installation overview frame and hitting down button to get there.
-        if (check_var('VIDEOMODE', 'text')) {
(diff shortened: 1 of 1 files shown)


waynechen55:
@punkioudi @alice-suse  If you think this is helpful, you can just merge this directly. Anyway anyone can solve problem by any other ways. Thanks.

punkioudi:
Hi Wayne, thanks for the PR! 
I would suggest that you add some verification runs before we merge, as this module is being used in different places in osd :)

waynechen55:
> Hi Wayne, thanks for the PR! I would suggest that you add some verification runs before we merge, as this module is being used in different places in osd :)

Of course, I thought you can just apply this for emergency. 
Anyway I did two verification runs. I think it is enough. I added is_sle_micro, so:
It still works for sle micro [verification run 1](http://10.67.131.12/tests/849)
And now it works for sles as well because it is not sle micro [verification run 2](http://10.67.131.12/tests/851)

punkioudi (approved):
lgtm!

PR #15478, opened 2022-09-05
Remove workaround

paolostivanin:
Since the fix has been released, we can now remove the workaround (bsc#1200649).

Verification runs:
- 15-SP4: https://openqa.suse.de/tests/9458270
- 15-SP3: https://openqa.suse.de/tests/9458269


punkioudi (approved):
Nice!

PR #15480, opened 2022-09-06
Add function to get qe-sap-deployment code in the test

mpagot:
Create a library for the qe-sap-deployment project management
Add a Trento test step about deploying a SAP Landscape
Change the Trento scheduling to use this new test step

- Verification run with git clone of the qe-sap-deployment repo: http://openqaworker15.qa.suse.cz/tests/20849
- Verification run with QESAPDEPLOY_VER=0.2.0: http://openqaworker15.qa.suse.cz/tests/20850

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -1,0 +1,97 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Functions to use qe-sap-deployment project
+# Maintainer: QE-SAP <qe-sap@suse.de>
+
+## no critic (RequireFilenameMatchesPackage);
+
+=encoding utf8
+
+=head1 NAME
+
+qe-sap-deployment test lib
+
+=head1 COPYRIGHT
+
+Copyright 2022 SUSE LLC
+SPDX-License-Identifier: FSFAP
+
+=head1 AUTHORS
+
+QE SAP <qe-sap@suse.de>
+
+=cut
+
+package qesapdeployment;
+
+use strict;
+use warnings;
+use testapi;
+use Exporter 'import';
+
+# Constants
+use constant DEPLOYMENT_DIR => get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment');
+use constant QESAP_GIT_CLONE_LOG => '/tmp/git_clone.log';
+
+my @log_files = ();
+
+our @EXPORT = qw(
+  qesap_create_folder_tree
+  qesap_get_deployment_code
+);
+
+
+=head1 DESCRIPTION
+
+Package with common methods and default or constant  values for qe-sap-deployment
+
+=head2 Methods
+=head3 qesap_create_folder_tree
+
+Create all needed folders
+
+=cut
+
+sub qesap_create_folder_tree {
+    assert_script_run('mkdir -p ' . DEPLOYMENT_DIR, quiet => 1);
+}
+
+
+=head3 qesap_get_deployment_code
+
+Get the qe-sap-deployment code
+=cut
+
+sub qesap_get_deployment_code {
+    record_info("QESAP repo", "Preparing qe-sap-deployment repository");
+
+    my $git_repo = get_var(QESAPDEPLOY_GITHUB_REPO => 'github.com/SUSE/qe-sap-deployment');
+    enter_cmd "cd " . DEPLOYMENT_DIR;
+
+    # Script from a release
+    if (get_var('QESAPDEPLOY_VER')) {
+        my $ver_artifact = 'v' . get_var('QESAPDEPLOY_VER') . '.tar.gz';
+
+        my $curl_cmd = "curl -v -L https://$git_repo/archive/refs/tags/$ver_artifact -o$ver_artifact";
+        assert_script_run("set -o pipefail ; $curl_cmd | tee " . QESAP_GIT_CLONE_LOG, quiet => 1);
+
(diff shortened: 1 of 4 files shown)


tests/sles4sap/trento/cluster_configure.pm
```diff
@@ -0,0 +1,52 @@
...
+
+    # tfvars file
+    my $qesap_provider = $self->qesap_translate_provider_name();
+
+    $self->qesap_configure_tfvar($qesap_provider,
+        $provider->provider_client->region,
+        get_required_var('QESAPDEPLOY_OS_VER'));
+
+    # variables.sh file
+    $self->qesap_configure_variables($qesap_provider,
```
alvarocarvajald, line 35: Same with `qesap_configure_variables()`

tests/sles4sap/trento/cluster_configure.pm
```diff
@@ -0,0 +1,52 @@
...
+        get_required_var('SCC_REGCODE_SLES4SAP'));
+
+    # Ansible blob file
+    $self->qesap_configure_hanamedia(get_var('QESAPDEPLOY_SAPCAR'),
```
alvarocarvajald, line 39: Same with `qesap_configure_hanamedia()`

tests/sles4sap/trento/cluster_configure.pm
```diff
@@ -0,0 +1,52 @@
...
+        get_var('QESAPDEPLOY_IMDB_SERVER'),
+        get_var('QESAPDEPLOY_IMDB_CLIENT'));
+}
+
+sub post_fail_hook {
+    my ($self) = shift;
+    $self->select_serial_terminal;
+    enter_cmd 'cd  ~/test/qe-sap-deployment';
+    $self->qesap_upload_logs('', 1);
```
alvarocarvajald, line 48: Same with `qesap_upload_logs()`

tests/sles4sap/trento/cluster_configure.pm
```diff
@@ -0,0 +1,52 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: Configuration steps for qe-sap-deployment
+# Maintainer: QE-SAP <qe-sap@suse.de>, Michele Pagot <michele.pagot@suse.com>
+
+use strict;
+use warnings;
+use Mojo::Base 'publiccloud::basetest';
+use testapi;
+use base 'qesapdeployment';
```
alvarocarvajald, line 11: Having this module inheriting methods from 2 base classes worries me a bit. And from looking at the current code in the PR, I don't see a need for `qesapdeployment` being a base class. None of its methods do anything with `$self` and the only variable stored within the class is `@qesapdeployment::logs` which is used once. Perhaps with the missing methods it'll be more clear, but as things are, I would suggest changing this line to `use 'qesapdeployment';` instead.

alvarocarvajald (changes requested):
There are a couple of typos in the commit message:
```
Add function to get qe-sap-deployment code in the test

Create a library for the qe-sap-deployment project maangement
Add a Trento test step about deploying a SAP Landscape
Chenge the Trento scheduling to use this new test step
```
i.e.:
```
s/maangement/management/
s/Chenge/Change/
```
Also, the verification run does not match what I see in the code. There are some `cp` and `sed` commands which are not in the code. Perhaps they are in the missing functions?

lpalovsky (approved):
LGTM

tests/sles4sap/trento/cluster_configure.pm
```diff
@@ -0,0 +1,28 @@
...
+use testapi;
+use qesapdeployment;
+
+sub run {
+    my ($self) = @_;
+    $self->select_serial_terminal;
+
+    # Get the code for the qe-sap-deployment
+    $self->qesap_create_folder_tree();
+    $self->qesap_get_deployment_code();
```
alvarocarvajald, line 19: Now that `qesapdeployment` is not a base class, there's no need to reference its methods with `$self->`

lib/qesapdeployment.pm
```diff
@@ -0,0 +1,97 @@
...
+        enter_cmd 'ls -lai';
+    }
+    else {
+        # Get the code for the qe-sap-deployment by cloning its repository
+        assert_script_run('git config --global http.sslVerify false', quiet => 1) if get_var('QESAPDEPLOY_GIT_NO_VERIFY');
+        my $git_branch = get_var('QESAPDEPLOY_GITHUB_BRANCH', 'main');
+
+        my $git_clone_cmd = 'git clone --depth 1 --branch ' . $git_branch . ' https://' . $git_repo . ' ' . DEPLOYMENT_DIR;
+        push(@log_files, QESAP_GIT_CLONE_LOG);
+        assert_script_run("set -o pipefail ; $git_clone_cmd | tee " . QESAP_GIT_CLONE_LOG, quiet => 1);
```
alvarocarvajald, line 92: You are saving the logs and not really doing anything with them neither in the if-block nor in the else-block. Perhaps you could add a `upload_logs QESAP_GIT_CLONE_LOG, failok => 1` at the end of the function so it saves the output of the commands in the job in case the reviewer wants to take a look at it.

alvarocarvajald (approved):
LGTM.

VR for `git clone` option does not have much details as all commands are running with `quiet => 1`, but I suppose we would see more in case of failures.

However, I do have 2 comments regarding the logs and method referencing below.

lib/qesapdeployment.pm
mpagot: About it, this is next step where I like to move https://github.com/mpagot/os-autoinst-distri-opensuse/commit/c92f8e3041251fdb94c7e0aacc340b7242fb81fc

PR #15483, opened 2022-09-07
Replace qa-apac-i with qe-security

paolostivanin:
https://progress.opensuse.org/issues/116185


tjyrinki (approved):
Seems good! This will also be a test whether the permissions on the repositories are set right..

PR #15485, opened 2022-09-07
Refactor powerman module

b10n1k:
This starting as an investigation for the weird last validation from `powerman -c`. it turned out that this was not an issue as the device used on the test it doesnt supported. So the question was if it should be removed or improved in one or the other way.

So as part of that improvement, the module checks for the capabilities of the divice given by the specification files. In additional, the validation of the commands are replaced by `validate_script_output` and some info added to the OpenQA UI.

https://openqa.suse.de/tests/overview?build=b10n1k%2Fos-autoinst-distri-opensuse%2315485&version=15-SP5&distri=sle

### DIFF
--- a/tests/hpc/powerman.pm
+++ b/tests/hpc/powerman.pm
@@ -16,44 +16,60 @@ use testapi;
 use utils;
 use susedistribution;
 
+
+our $cfg_file = "/etc/powerman/powerman.conf";
+
+our $dev_caps = sub {
+    my $dev = shift;
+    my %dat;
+    my $dev_file_cnt = script_output("cat /etc/powerman/${dev}.dev");
+    ($dat{$1} = 1) while $dev_file_cnt =~ m/script (.*)\W\S/g;
+    return \%dat;
+};
+
 sub run ($self) {
     # install powerman
     zypper_call('in powerman');
-
+    my $powerman_dev = 'bashfun';
     # Adapt config
-    my $cfg_file = "/etc/powerman/powerman.conf";
     my $hostname = script_output('hostname');
     assert_script_run(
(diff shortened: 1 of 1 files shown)


tests/hpc/powerman.pm
```diff
@@ -16,44 +16,60 @@ use testapi;
...
     # check if target can be turned off
     assert_script_run("powerman -0 \$(hostname)");
-    script_run("powerman -Q \$(hostname) | tee /dev/$serialdev", 0);
-    wait_serial(/^off:.*$hostname.*/);
+    validate_script_output("powerman -Q \$(hostname)", sub { /off:\s+$hostname.*/ });
 
     # check if command can be handled by power control device
-    script_run("powerman -c \$(hostname) | tee /dev/$serialdev", 0);
-    wait_serial(/.*cannot be handled by power control device.*/);
+    # This is depended by the device/driver type.
```
schlad, line 65: depended on

schlad:
Ideally the git commit message would be amended as there are some grammar mistakes. Sorry for saying :-)
For instance: "This starting". Preferably the git commit would be written in a passive form as that is the standard.

The change itself is nice however small. I think, once we merge your changes, the test will be easier to maintain and any failure in the future should be easier to investigate. So I really like this PR.

schlad (changes requested):

tests/hpc/powerman.pm
schlad: actually it should say something like: This depends on the device/driver type.

b10n1k: *this change is ready for review*

PR #15488, opened 2022-09-08
Install kdumptool where needed and fix synchronization failure in yast kdump module

lemon-suse:
Need install kdump which is not dependency of yast2-kdump module in SLE15-SP5 and is no longer installed when yast2-kdump is installed. And reboot requirement warning message need to be deal with on s390x. Use multi-tag to deal with the different behavior for reboot popup message.

- Related ticket: https://progress.opensuse.org/issues/115622
- Needles:  yast2-kdump-no-restart-info:  already merged to OSD.
- Verification run: 
   https://openqa.nue.suse.com/tests/9514874#  x86_64
   https://openqa.nue.suse.com/tests/9514876#   s390x
   http://openqa.nue.suse.com/tests/9514822#    aarch64  failed for bsc#1202738
   http://openqa.nue.suse.com/tests/9514821#    ppc64le

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -192,13 +192,15 @@ sub activate_kdump {
         $expect_restart_info = 1;
     }
     send_key('alt-o');
+    # Expect yast2-kdump-restart-info on s390x
+    $expect_restart_info = 1 if (is_s390x && is_sle('15-SP5+'));
     if ($expect_restart_info == 1) {
-        my @tags = qw(yast2-kdump-restart-info os-prober-warning);
+        my @tags = qw(yast2-kdump-restart-info os-prober-warning yast2-kdump-no-restart-info);
         do {
             assert_screen(\@tags, timeout => 180);
             handle_warning_install_os_prober() if match_has_tag('os-prober-warning');
-        } until (match_has_tag('yast2-kdump-restart-info'));
-        send_key('alt-o');
+        } until (match_has_tag('yast2-kdump-restart-info') || match_has_tag('yast2-kdump-no-restart-info'));
+        send_key('alt-o') if match_has_tag('yast2-kdump-restart-info');
     }
     wait_serial("$module_name-0", 240) || die "'yast2 kdump' didn't finish";
 }
--- a/tests/console/yast2_kdump.pm
+++ b/tests/console/yast2_kdump.pm
@@ -17,6 +17,7 @@ use registration;
 use scheduler 'get_test_suite_data';
 use testapi;
 use utils;
+use version_utils 'is_sle';
 
 sub run {
     select_console('root-console');
@@ -24,6 +25,7 @@ sub run {
     # install kdump by adding additional modules
     add_suseconnect_product('sle-module-desktop-applications');
     add_suseconnect_product('sle-module-development-tools');
+    zypper_call('in kdump') if is_sle('15-SP5+');
     zypper_call('in yast2-kdump');
 
     # Kdump configuration with YaST module


lib/kdump_utils.pm
```diff
@@ -192,6 +192,8 @@ sub activate_kdump {
         $expect_restart_info = 1;
     }
     send_key('alt-o');
+    # Expect yast2-kdump-restart-info on s390x
+    $expect_restart_info = 1 if (is_s390x && is_sle('15-SP5+'));
```
coolgw, line 196: base https://openqa.nue.suse.com/tests/9474977#step/yast2_kdump/18 i found s390 default kdump is enable, so nothing is changed, but a kdump restart page popup after push OK button, i suppose this is an issue of s390?

lemon-suse: I'm not sure this is a feature or a bug, maybe we can ask a kernel expert about this or just file a bug and wait developer to confirm?

coolgw: I think this is an bug if nothing updated by yast from end user view.

ge0r:
Hi, thanks for addressing the issue. Is it possible to also add a verification run for `ppc64le` ?

lemon-suse:
> Hi, thanks for addressing the issue. Is it possible to also add a verification run for `ppc64le` ?

Ok, I just noticed it is ppc64-spvm has some environment issue while not ppc64le, I will run VR on ppc64le.

lemon-suse:
Updated code and all VRs.

lib/kdump_utils.pm
ge0r: I would suggest to open a bug for s390x and use `record_soft_failure` mentioning the bug here. We can delete the `record_soft_failure` line afterwards if it is confirmed to be intended instead.

lib/kdump_utils.pm
```diff
@@ -192,13 +192,22 @@ sub activate_kdump {
         $expect_restart_info = 1;
     }
     send_key('alt-o');
+    #if (is_s390x || is_ppc64le) {
+    #    assert_screen([qw(yast2-kdump-restart-info yast2-kdump-no-restart-info)]);
+    #    if (match_has_tag('yast2-kdump-restart-info')) {
+    #        send_key('alt-o');
+    #	}
+    #	$expect_restart_info = 0;
+    #}
```
ge0r, line 201: I think this commented out section could be deleted

ge0r (commented):
Thanks for taking a look into this!

lib/kdump_utils.pm
lemon-suse: Thanks for comments.
I got the info from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14887#issuecomment-1124851781 that this different behavior is not a bug. And in fact I have filed a bug for s390 (https://bugzilla.suse.com/show_bug.cgi?id=1203237) but no response till now.
So I changed code to suit it for different behaviors on test side.

ge0r: That's great, thanks for the info!
All good then, I would maybe suggest to add a `record_soft_failure('bsc#1203237 - s390x openQA test fails in yast2_kdump - reboot page popup unexpected')` just to keep track of this adaptation in the code

ge0r (approved):

lib/kdump_utils.pm
jknphy: YaST makes some calculation and depends on that the popup appears o not:
yast2_ncurses_textmode
it depends of course of settings of the job.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14887#issuecomment-1124851781

jknphy:
The code became more and more conditional, we will address this in: https://progress.opensuse.org/issues/111016 using libyui REST. @lemon-suse thanks for the fix.
For what I can see in Kernel job group they move to cli or at least there is not more use of this test module.
Let's merge and see if we account for all the differences.

rfan1:
@lemon-suse 
Could you please fix your typo in git commit message?

jknphy:
> @lemon-suse Could you please fix your typo in git commit message?

sorry, too late, I didn't notice that before merging, now to change history I don't think is a good idea.

lemon-suse:
> @lemon-suse Could you please fix your typo in git commit message?

It should be for my keyboard issue, sometimes type no response, sometimes more. Anyway later I will keep an eye on such typo, and later use my new laptop should avoid such issue. Thanks.

PR #15494, opened 2022-09-09
Unschedule regression test modules unrelated with migration

chcao:
This pr is to unschedule the test modules unrelated with migration in regression job group. So I update the schedule yaml files of migration regression jobs.

- Related ticket: https://progress.opensuse.org/issues/115487
- Related mr: https://gitlab.suse.de/coolgw/wegao-test/-/merge_requests/139
- Needles: N/A
- Verification run: 
offline:
x86_64: https://openqa.suse.de/tests/9554935#
s390x: https://openqa.suse.de/tests/9555425#
ppc64le: https://openqa.suse.de/tests/9548988#
online:
aarch64: https://openqa.suse.de/tests/9548989#
x86_64: https://openqa.suse.de/tests/9548990#

### DIFF
--- a/schedule/migration/aarch64_regression_test_offline.yaml
+++ b/schedule/migration/aarch64_regression_test_offline.yaml
@@ -3,9 +3,7 @@ description:    |
   This is for aarch64 offline regression test.
   #REGRESSION_SERVICE: '1' means support service check test. '0' means doesn't
   #support service check test, normally set '0' for package media test.
-  #REGRESSION_TEST: '1' means regression test part1 include console test. '2'
-  #means regression test part2 include x11 test.
-  #REGRESSION_389DS: '1' means load 389ds test.
+  #REGRESSION_TEST: '1' means load regression test modules after migration.
   #QCOW_GENERATION: '1' means publish qcow after migration; '0' means no need.
 vars:
   DESKTOP: 'gnome'
@@ -46,20 +44,21 @@ conditional_schedule:
         - console/system_prepare
         - console/check_os_release
         - console/check_system_info
-        - console/check_migration_features
+        - '{{check_migration_features}}'
         - console/check_network
         - console/system_state
         - console/prepare_test_data
         - console/consoletest_setup
         - '{{check_upgraded_service}}'
-        - '{{openldap_to_389ds}}'
         - '{{regression_tests}}'
-        - boot/grub_test_snapshot
-        - migration/version_switch_origin_system
-        - boot/snapper_rollback
+        - '{{rollback_after_migration}}'
       1:
         - shutdown/cleanup_before_shutdown
         - shutdown/shutdown
+  check_migration_features:
+    MIGRATION_FEATURES:
+      1:
+        - console/check_migration_features
   check_upgraded_service:
     REGRESSION_SERVICE:
       1:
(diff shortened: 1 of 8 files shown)


schedule/migration/aarch64_regression_test_offline.yaml
```diff
@@ -46,20 +44,23 @@ conditional_schedule:
...
         - console/check_os_release
         - console/check_system_info
-        - console/check_migration_features
+        - '{{check_migration_features}}'
         - console/check_network
         - console/system_state
         - console/prepare_test_data
         - console/consoletest_setup
         - '{{check_upgraded_service}}'
-        - '{{openldap_to_389ds}}'
```
ge0r, line 55: From what I understood from the progress ticket, this should not be removed:  `modules like migration/openldap_to_389ds or console/orphaned_packages_check or basic checks of the migrated system should remain there`

ge0r: Also applies for the other yamls where `openldap_to_389ds` is removed.

lemon-suse:
Before the review, we'd better make it clear for the meaning of 'related migration'.
To me it is just strong or weak related, or directly related or non-directly related, while no unrelated. Most regression test mainly focus on the functionality verification test after migration. 
Maybe here the related with migration means only directly involved migration flow?
If remove most of them, I don't think it can be called regression group. Anyway, just make it more clear. :)

schedule/migration/aarch64_regression_test_offline.yaml
chcao: openldap_to_389ds is only tested on x86_64, for the other arches, it is not supported and I checked the cases' setting, no case set OPENLDAP_TO_389DS=1. So I remove it from the other three arches' yamls directly.

jknphy:
> Before the review, we'd better make it clear for the meaning of 'related migration'. To me it is just strong or weak related, or directly related or non-directly related, while no unrelated. Most regression test mainly focus on the functionality verification test after migration. Maybe here the related with migration means only directly involved migration flow? If remove most of them, I don't think it can be called regression group. Anyway, just make it more clear. :)

the idea is to include two kind of regression test, (1) the basic ones, generic, system regressions, the most simple ones, the ones not problematic to review (le't say it like that), and also (2) other test modules checking for specific regression related with migration.

schedule/migration/aarch64_regression_test_offline.yaml
```diff
@@ -72,57 +73,15 @@ conditional_schedule:
...
-        - console/ncurses
-        - console/yast2_lan
-        - console/curl_https
-        - console/salt
-        - console/glibc_tunables
-        - console/zypper_in
-        - console/zypper_log
-        - console/yast2_i
-        - console/yast2_bootloader
-        - console/firewall_enabled
```
jknphy, line 104: firewall check seems simple enough to be included, wdyt?

schedule/migration/aarch64_regression_test_online.yaml
```diff
@@ -60,57 +55,15 @@ conditional_schedule:
...
-        - console/firewall_enabled
-        - console/krb5
-        - console/sshd
-        - console/ssh_cleanup
-        - console/mtab
+        - console/check_system_info
         - console/zypper_lifecycle
         - console/orphaned_packages_check
-        - console/kdump_and_crash
-        - console/consoletest_finish
```
jknphy, line 100: this was included in the previous test suite but not in this one, did you miss it or it is for some reason?

schedule/migration/aarch64_regression_test_online.yaml
```diff
@@ -60,57 +55,15 @@ conditional_schedule:
...
-  openldap_to_389ds:
-    REGRESSION_389DS:
-      1:
-        - migration/openldap_to_389ds
   regression_tests:
     REGRESSION_TEST:
       1:
         - locale/keymap_or_locale
-        - console/supportutils
-        - console/force_scheduled_tasks
```
jknphy, line 72: `force_scheduled_tasks` is a basic one to ensure the load is ok, sometimes the system is too busy to actually do anything and it is recommended to schedule it.

schedule/migration/ppc64le_regression_test_online.yaml
```diff
@@ -64,57 +59,15 @@ conditional_schedule:
...
-        - console/sshd
-        - console/ssh_cleanup
-        - console/mtab
         - console/zypper_lifecycle
         - console/orphaned_packages_check
-        - console/kdump_and_crash
-        - console/consoletest_finish
-      2:
         - console/consoletest_finish
-        - x11/desktop_runner
```
jknphy, line 107: I think instead of deleting all the x11 test cases, we could at least keep this one, that is simple enough.
`xterm` optionally could also be included, the rest are using those two somehow, wdyt? `setup` perhaps is needed I don't know what that does, perhaps is needed before `xterm`.

lemon-suse:
> > Before the review, we'd better make it clear for the meaning of 'related migration'. To me it is just strong or weak related, or directly related or non-directly related, while no unrelated. Most regression test mainly focus on the functionality verification test after migration. Maybe here the related with migration means only directly involved migration flow? If remove most of them, I don't think it can be called regression group. Anyway, just make it more clear. :)
> 
> the idea is to include only two kind of regression test modules, (1) the basic ones, generic, system regressions, the most simple ones, the ones not problematic to review (le't say it like that), and also (2) other test modules checking for specific regression related with migration.

That's make sense for me, thanks. Then I will involve the next review.

lemon-suse:
> > Before the review, we'd better make it clear for the meaning of 'related migration'. To me it is just strong or weak related, or directly related or non-directly related, while no unrelated. Most regression test mainly focus on the functionality verification test after migration. Maybe here the related with migration means only directly involved migration flow? If remove most of them, I don't think it can be called regression group. Anyway, just make it more clear. :)
> 
> the idea is to include only two kind of regression test modules, (1) the basic ones, generic, system regressions, the most simple ones, the ones not problematic to review (le't say it like that), and also (2) other test modules checking for specific regression related with migration.

That's make sense for me, thanks. Then I will involve the next review.

PR #15496, opened 2022-09-09
Create equivalent yast2_firewall test module based on libyui-rest-api

coolgw:
Using libyui to check yast firewall module 

- Related ticket: https://progress.opensuse.org/issues/113492
- Needles: na
- Latest VR: 
https://openqa.opensuse.org/tests/2844742#details
https://openqa.opensuse.org/tests/2846745
https://openqa.suse.de/tests/9843828#details

### DIFF
--- a/lib/YaST/Firewall/FirewallController.pm
+++ b/lib/YaST/Firewall/FirewallController.pm
@@ -1,0 +1,156 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces actions Firewall Settings Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::FirewallController;
+use strict;
+use warnings;
+use testapi;
+use YaST::Firewall::MainPage;
+use YaST::Firewall::StartUpPage;
+use YaST::Firewall::InterfacesPage;
+use YaST::Firewall::ZonesPage;
+use YaST::Firewall::ZonePage;
+use YaST::Firewall::ServicesPage;
+use YaST::Firewall::PortsPage;
+
+use YuiRestClient;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {}, $class;
+    return $self->init($args);
+}
+
+sub init {
+    my ($self, $args) = @_;
+    $self->{MainPage} = YaST::Firewall::MainPage->new({app => YuiRestClient::get_app()});
+    $self->{StartUpPage} = YaST::Firewall::StartUpPage->new({app => YuiRestClient::get_app()});
+    $self->{InterfacesPage} = YaST::Firewall::InterfacesPage->new({app => YuiRestClient::get_app()});
+    $self->{ZonesPage} = YaST::Firewall::ZonesPage->new({app => YuiRestClient::get_app()});
+    $self->{ZonePage} = YaST::Firewall::ZonePage->new({app => YuiRestClient::get_app()});
+    $self->{ServicesPage} = YaST::Firewall::ServicesPage->new({app => YuiRestClient::get_app()});
+    $self->{PortsPage} = YaST::Firewall::PortsPage->new({app => YuiRestClient::get_app()});
+    return $self;
+}
+
+sub get_main_page {
+    my ($self) = @_;
+    die 'Main Firewall Page is not displayed' unless $self->{MainPage}->is_shown();
+    return $self->{MainPage};
+}
+
+sub get_start_up_page {
+    my ($self) = @_;
+    die 'StartUp Pane is not displayed' unless $self->{StartUpPage}->is_shown();
+    return $self->{StartUpPage};
+}
+
+sub get_interfaces_page {
+    my ($self) = @_;
+    die 'Interfaces Pane is not displayed' unless $self->{InterfacesPage}->is_shown();
+    return $self->{InterfacesPage};
+}
+
+sub get_zones_page {
+    my ($self) = @_;
+    die 'Zones Pane is not displayed' unless $self->{ZonesPage}->is_shown();
+    return $self->{ZonesPage};
+}
+
+sub get_zone_page {
+    my ($self) = @_;
+    die 'Zone Pane is not displayed' unless $self->{ZonePage}->is_shown();
+    return $self->{ZonePage};
+}
+
+sub get_services_page {
+    my ($self) = @_;
+    die 'Services Pane is not displayed' unless $self->{ServicesPage}->is_shown();
+    return $self->{ServicesPage};
+}
+
+sub get_ports_page {
+    my ($self) = @_;
+    die 'Ports Pane is not displayed' unless $self->{PortsPage}->is_shown();
+    return $self->{PortsPage};
+}
+
+sub select_start_up_page {
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_start_up_page();
+}
+
+sub select_interfaces_page {
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_interfaces_page();
+}
+
+sub select_zones_page {
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_zones_page();
+}
+
+sub select_zone_page {
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_zone_page($zone);
+}
+
+sub start_firewall {
+    my ($self) = @_;
+    $self->get_start_up_page()->start_firewall();
+}
+
+sub stop_firewall {
+    my ($self) = @_;
+    $self->get_start_up_page()->stop_firewall();
+}
+
+sub accept_change {
+    my ($self) = @_;
+    $self->get_main_page()->press_accept();
+}
+
+sub set_default_zone {
+    my ($self, $zone) = @_;
+    $self->get_zones_page()->select_zone($zone);
+    save_screenshot;
+    $self->get_zones_page()->set_default_zone();
+}
+
+sub set_interface_zone {
+    my ($self, $device, $zone) = @_;
+    $self->get_interfaces_page()->set_interface_zone($device, $zone);
+}
+
+sub switch_ports_tab {
+    my ($self) = @_;
+    $self->get_zone_page()->switch_ports_tab();
+}
+
+sub switch_services_tab {
+    my ($self) = @_;
+    $self->get_zone_page()->switch_services_tab();
+}
+
+sub add_service {
+    my ($self, $zone, $service) = @_;
+    $self->switch_services_tab();
+    save_screenshot;
+    $self->get_services_page()->select_service($zone, $service);
+    save_screenshot;
+    $self->get_services_page()->add_service();
+}
+
+sub add_tcp_port {
+    my ($self, $port) = @_;
+    $self->switch_ports_tab();
+    save_screenshot;
+    $self->get_ports_page()->set_tcp_port($port);
+}
+
+1;
--- a/lib/YaST/Firewall/MainPage.pm
+++ b/lib/YaST/Firewall/MainPage.pm
@@ -1,0 +1,58 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Firewall Main Page(Include left view tree menu and bottom button).
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::MainPage;
+use strict;
+use warnings;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my $self = shift;
+    $self->{tre_overview} = $self->{app}->tree({id => "\"Y2Firewall::Widgets::OverviewTree\""});
+    $self->{btn_accept} = $self->{app}->button({id => 'next'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{tre_overview}->exist();
+}
+
+sub select_start_up_page {
+    my ($self) = @_;
+    $self->{tre_overview}->select("Start-Up");
+}
+
+sub select_interfaces_page {
+    my ($self) = @_;
+    $self->{tre_overview}->select("Interfaces");
+}
+
+sub select_zones_page {
+    my ($self) = @_;
+    $self->{tre_overview}->select("Zones");
+}
+
+sub press_accept {
+    my ($self) = @_;
+    $self->{btn_accept}->click();
+}
+
+sub select_zone_page {
+    my ($self, $zone) = @_;
+    $self->{tre_overview}->select("Zones|$zone");
+}
+
+1;
--- a/lib/YaST/Firewall/StartUpPage.pm
+++ b/lib/YaST/Firewall/StartUpPage.pm
@@ -1,0 +1,43 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Firewall Startup Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::StartUpPage;
+use strict;
+use warnings;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my $self = shift;
+    $self->{lbl_status} = $self->{app}->label({id => 'service_widget_status'});
+    $self->{cmb_set_firewall_state} = $self->{app}->combobox({id => 'service_widget_action'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{lbl_status}->exist();
+}
+
+sub start_firewall {
+    my ($self) = @_;
+    $self->{cmb_set_firewall_state}->select("Start");
+}
+
+sub stop_firewall {
+    my ($self) = @_;
+    $self->{cmb_set_firewall_state}->select("Stop");
+}
+
+1;
--- a/lib/YaST/Firewall/ZonePage.pm
+++ b/lib/YaST/Firewall/ZonePage.pm
@@ -1,0 +1,44 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Firewall Zone Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::ZonePage;
+use strict;
+use warnings;
+
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my $self = shift;
+    $self->{tab_services_ports} = $self->{app}->tab({id => "\"CWM::DumbTabPager\""});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{tab_services_ports}->exist();
+}
+
+# Using "Ports" instead of "&Ports" as tab select function parameter since YuiRestClient::Widget::Base will
+# using sanitize function remove "&" before compare the string
+sub switch_ports_tab {
+    my ($self) = @_;
+    $self->{tab_services_ports}->select("Ports");
+}
+
+sub switch_services_tab {
+    my ($self) = @_;
+    $self->{tab_services_ports}->select("Services");
+}
+
+1;
--- a/schedule/yast/opensuse/yast2_gui/yast2_gui.yaml
+++ b/schedule/yast/opensuse/yast2_gui/yast2_gui.yaml
@@ -22,13 +22,18 @@ schedule:
   - yast2_gui/yast2_bootloader
   - "{{bootcode_options}}"
   - yast2_gui/yast2_datetime
-  - yast2_gui/yast2_firewall
+  - yast2_gui/yast2_firewall_stop_service
+  - yast2_gui/yast2_firewall_start_service
+  - yast2_gui/yast2_firewall_set_default_zone
+  - yast2_gui/yast2_firewall_set_interface
+  - yast2_gui/yast2_firewall_set_service_port
   - yast2_gui/yast2_hostnames
   - yast2_gui/yast2_lang
   - yast2_gui/yast2_network_settings
   - yast2_gui/yast2_software_management
   - yast2_gui/yast2_users
   - yast2_gui/yast2_security
+
 conditional_schedule:
   bootcode_options:
     ARCH:
--- a/schedule/yast/yast2_gui/yast2_gui_sle.yaml
+++ b/schedule/yast/yast2_gui/yast2_gui_sle.yaml
@@ -29,7 +29,12 @@ conditional_schedule:
         - yast2_gui/yast2_expert_partitioner
         - yast2_gui/yast2_software_management
         - yast2_gui/yast2_security
-        - yast2_gui/yast2_firewall
+        - yast2_gui/yast2_firewall_stop_service
+        - yast2_gui/yast2_firewall_start_service
+        - yast2_gui/yast2_firewall_set_default_zone_prepare
+        - yast2_gui/yast2_firewall_set_default_zone
+        - yast2_gui/yast2_firewall_set_interface
+        - yast2_gui/yast2_firewall_set_service_port
         - yast2_gui/yast2_bootloader
         - yast2_gui/bootloader/bootcode_options
         - x11/yast2_snapper
--- a/test_data/yast/yast2_gui.yaml
+++ b/test_data/yast/yast2_gui.yaml
@@ -28,3 +28,5 @@ bootcode_applied_params:
 bootcode_device:
   default: /dev/vda
   write_to_partition: /dev/vda2
+port: 30000-50000
+zone: public
--- a/tests/yast2_gui/yast2_firewall_set_default_zone.pm
+++ b/tests/yast2_gui/yast2_firewall_set_default_zone.pm
@@ -1,0 +1,40 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: YaST2 Firewall UI test checks verious configurations and settings of firewall
+# Make sure yast2 firewall can opened properly. Configurations can be changed and written correctly.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_guitest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
+use YaST::Module;
+
+sub run {
+    my $self = shift;
+    my $iface = iface();
+
+    $self->select_serial_terminal();
+    validate_script_output("firewall-cmd --get-default-zone", sub { m/public/ }, proceed_on_failure => 1);
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    $testapi::distri->get_firewall()->select_interfaces_page();
+    save_screenshot;
+    $testapi::distri->get_firewall()->select_zones_page();
+    save_screenshot;
+    $testapi::distri->get_firewall()->set_default_zone("trusted");
+    save_screenshot;
+    $testapi::distri->get_firewall()->accept_change();
+    assert_screen 'generic-desktop';
+    $self->select_serial_terminal();
+    validate_script_output("firewall-cmd --list-interfaces --zone=trusted", sub { m/$iface/ }, proceed_on_failure => 1);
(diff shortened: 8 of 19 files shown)


lib/Installation/ModuleRegistration/ModuleRegistrationInstallationReportPage.pm
```diff
@@ -0,0 +1,27 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act with page that ask for module extra registration code
```
jknphy, line 6: is this accidentally added from another branch?

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,68 @@
...
+
+sub is_shown {
+    my ($self) = @_;
+    my $is_shown = $self->{overview_tree}->exist();
+    save_screenshot if $is_shown;
+    return $is_shown;
+
+}
+
+sub select_start_up_page {
```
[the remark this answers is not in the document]
jknphy: the hyphens can be abbreviated removing them, s/start_up/startup/ to distinguish them from different words. I have seen this in other naming conventions, we  don't have it written anywhere :) but just in case you want to shorten it.

lib/YaST/Firewall/StartUpPage.pm
```diff
@@ -0,0 +1,47 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for YaST module
+# System Settings Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::StartUpPage;
```
jknphy, line 10: this should be encapsulated in ServiceConfigurationPage, as we will be reusing this outside /Firewall/ in other YaST modules.

tests/yast2_gui/yast2_firewall.pm
```diff
@@ -18,6 +18,7 @@ use version_utils qw(is_sle is_leap is_tumbleweed);
 use yast2_shortcuts '%fw';
 use network_utils 'iface';
 use yast2_widget_utils 'change_service_configuration';
+use YaST::Module;
```
jknphy, line 21: do we need a new test? are we sure this yast2_firewall is not used where we don't have libyui-rest-api? We don't need logic for old firewall technology when using libyui.

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,68 @@
...
+    save_screenshot;
+}
+
+sub select_zones_page {
+    my ($self) = @_;
+    $self->{overview_tree}->select("Zones");
+    save_screenshot;
+}
+
+sub press_accept {
```
jknphy, line 56: you mixed here this idea of main page with the specific page.
We probably need to extract here the bottom buttons and the tree and call it for example 'BasePage' . Because 'main' doesn't have that meaning.
then the specific pages will inherit this basepage.

jknphy (commented):
Initial round of review.

lib/YaST/Firewall/MainPage.pm
coolgw: The other page do not need any action of main page's button(accept button etc..), so from my view no need BasePage in our case

lib/YaST/Firewall/StartUpPage.pm
coolgw: The module name is consistent with yast UI, so you will map it 1-to-1 more easier if you need do more development.
From outside we should call function $testapi::distri->get_firewall()->start_firewall() etc.  this should not impact end user's usage.

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,68 @@
...
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for YaST module
+# System Settings Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::MainPage;
+use strict;
+use warnings;
+use testapi;
```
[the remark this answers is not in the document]
coolgw: I have rewrite the is_shown related function.

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,68 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for YaST module
+# System Settings Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::MainPage;
```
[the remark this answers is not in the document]
coolgw: This main page include left navigation tree and the  bottom button(accept, abort, etc..)

tests/yast2_gui/yast2_firewall.pm
coolgw: new test case created

jknphy:
setting to WIP (let's do that always, and remove the WIP when we think is ready) other optioni is to use label `draft`, but it is more clear the wip IMO.

tests/yast2_gui/yast2_firewall_libyui.pm
```diff
@@ -0,0 +1,207 @@
...
+    assert_screen "yast2_firewall_summary", 30;
+    send_key "alt-f";
+}
+
+sub verify_service_stopped {
+
+    record_info('Start-Up', "Managing the firewalld service: Stop");
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    assert_screen 'yast2_firewall_start-up';
```
jknphy, line 78: we should not need this needle at all, just check that an element of the page where we land exists.

tests/yast2_gui/yast2_firewall_libyui.pm
```diff
@@ -0,0 +1,207 @@
...
+
+    my $interfaces = $args{interfaces} //= 'no_interfaces';
+    my $default = $args{default} //= 'no_default';
+    my $menu_selected = $args{menu_selected} //= 0;
+    $testapi::distri->get_firewall()->get_zones_page();
+    save_screenshot;
+    if ($testapi::distri->get_firewall()->verify_zone($args{name}, $interfaces, $default) != 1) {
+        die("Verify zone failed! " . $args{name} . " " . $args{device} . " " . $interfaces . " " . $default);
+    } else {
+        record_info('verify zone ok! ', $args{name} . " " . $args{device} . " " . $interfaces . " " . $default);
```
jknphy, line 135: original test module abuses of using `record_info`, we should consider later when it makes sense because actually doesn't help much when there are too many, a well designed test doesn't need much use of that or even none.

jknphy (commented):
It is taking shape, half a way done and looking good!, you are setting a good example here about how to create automation with libyui-rest-api in the installed system, next modules will be easier to do based on this one. Let's discuss design decision offline if you have doubts, other topics can be already addressed.

tests/yast2_gui/yast2_firewall_libyui.pm
```diff
@@ -0,0 +1,207 @@
...
+    $testapi::distri->get_firewall()->select_zone_page($args{zone});
+    save_screenshot;
+    $testapi::distri->get_firewall()->add_service($args{zone}, $args{service});
+    $testapi::distri->get_firewall()->add_tcp_port($args{port});
+}
+
+sub configure_firewalld {
+
+    record_info('Interface/Zones ', "Verify zone info changing default zone when interface assigned to default zone");
+    my $iface = iface;
```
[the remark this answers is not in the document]
coolgw: iface is a function, use $iface store return value of function iface is efficient way.
Otherwise you need call function again and again in later code.

lib/YaST/Firewall/ZonePage.pm
```diff
@@ -0,0 +1,48 @@
...
+sub switch_ports_tab {
+    my ($self) = @_;
+    $self->{tab_services_ports}->select("Ports");
+    save_screenshot;
+}
+
+sub switch_services_tab {
+    my ($self) = @_;
+    $self->{tab_services_ports}->select("Services");
+    save_screenshot;
```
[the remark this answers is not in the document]
coolgw: I agree not use save_screenshot in page level, but in ControlPage level i still insist use some save_screenshot since controlPage also need warp some action, then we have more simple / clean api for test modules.
So i plan move save_screen to control level instead of remove totally within Page&ControlPage

lib/YaST/Firewall/MainPage.pm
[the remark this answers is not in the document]
coolgw: In order to keep one style in whole code, i think keep "page" is good.

lib/YaST/Firewall/ZonePage.pm
```diff
@@ -0,0 +1,48 @@
...
+    $self->{tab_services_ports} = $self->{app}->tab({id => "\"CWM::DumbTabPager\""});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{tab_services_ports}->exist();
+}
+
+# Using "Ports" instead of "&Ports" as tab select function parameter since YuiRestClient::Widget::Base will
```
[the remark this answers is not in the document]
coolgw: This is very exactly specific example to show this tricky part which hidden in libyui api, so i think keep it here as a notification.

tests/yast2_gui/yast2_firewall_libyui.pm
coolgw: Yes, it like enable verbose log, but it help to narrow down issue once fail happen.

tests/yast2_gui/yast2_firewall_libyui.pm
```diff
@@ -0,0 +1,207 @@
...
+sub configure_zone {
+    my (%args) = @_;
+
+    $testapi::distri->get_firewall()->select_zone_page($args{zone});
+    save_screenshot;
+    $testapi::distri->get_firewall()->add_service($args{zone}, $args{service});
+    $testapi::distri->get_firewall()->add_tcp_port($args{port});
+}
+
+sub configure_firewalld {
```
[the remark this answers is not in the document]
coolgw: Current control layer already wrap some operation to avoid too much detail shown in test code, it's not easy thing to balance between flexible api vs simple easy api.

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,60 @@
...
```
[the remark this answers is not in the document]
coolgw: Extra info "Summary: Handles Firewall Main Page(Include left view tree menu and bottom button)"  to explain what's this page contain.

tests/yast2_gui/yast2_firewall_libyui.pm
```diff
@@ -0,0 +1,140 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: yast2-firewall yast2-http-server apache2 apache2-prefork firewalld
```
jknphy, line 6: Package line, we don't use it. The description I think it is borrowed from old test case.

lib/YaST/Firewall/FirewallController.pm
```diff
@@ -0,0 +1,179 @@
...
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces actions Firewall Settings Page.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+package YaST::Firewall::FirewallController;
+use strict;
+use warnings;
+use testapi;
```
jknphy, line 12: we don't need `testapi`.

tests/yast2_gui/yast2_firewall_stop_service.pm
```diff
@@ -0,0 +1,31 @@
...
+# Summary: YaST2 Firewall UI test checks verious configurations and settings of firewall
+# Make sure yast2 firewall can stop properly. Configurations can be changed and written correctly.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_guitest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
```
jknphy, line 15: not needed.

tests/yast2_gui/yast2_firewall_start_service.pm
```diff
@@ -0,0 +1,31 @@
...
+# Summary: YaST2 Firewall UI test checks verious configurations and settings of firewall
+# Make sure yast2 firewall can start properly. Configurations can be changed and written correctly.
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_guitest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
```
jknphy, line 15: not needed.
I would recommend only import what we need with `use <lib> qw(a b c)`  to avoid the over-use of other libraries in libyui-rest-api test.

tests/yast2_gui/yast2_firewall.pm
```diff
@@ -206,3 +206,4 @@ sub run {
 }
 
 1;
+
```
jknphy, line 209: accidental line break.

jknphy (commented):
Looking better, now we just need to break `configure_firewalld` as we discussed offline avoiding UI validations and favoring console validations (will required some new ones as we agreed).

lib/YaST/Firewall/MainPage.pm
```diff
@@ -0,0 +1,58 @@
...
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my $self = shift;
+    $self->{tre_overview_tree} = $self->{app}->tree({id => "\"Y2Firewall::Widgets::OverviewTree\""});
```
jknphy, line 23: s/tre_overview_tree/tre_overview/

lib/YaST/Firewall/FirewallController.pm
coolgw: I need testapi to do capture screen since in Controller level i wrap some operation together let api more simple to use in test.pm.
such as line 175.

lib/YaST/Firewall/FirewallController.pm
```diff
@@ -0,0 +1,179 @@
...
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_zones_page();
+}
+
+sub select_zone_page {
+    my ($self, $zone) = @_;
+    $self->get_main_page()->select_zone_page($zone);
+}
+
+sub start_firewall {
```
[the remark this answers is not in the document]
coolgw: start/stop firewall belong same page, start_up_page, i have no idea how to reuse for other services.

lib/YaST/Firewall/FirewallController.pm
```diff
@@ -0,0 +1,179 @@
...
+    my ($self) = @_;
+    $self->get_start_up_page->start_firewall();
+}
+
+sub stop_firewall {
+    my ($self) = @_;
+    $self->get_start_up_page->stop_firewall();
+}
+
+sub accept_change {
```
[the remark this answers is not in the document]
coolgw: This is the name function for accept button, not means we only do one change or many change.

lib/YaST/Firewall/MainPage.pm
coolgw: The Widgets name is "OverviewTree"
so base our rule we use extra prefix "tre" + "overview_tree"(name of widgets)

jknphy: It doesn't say how to name the second part, I think it is open because, we don't use the id, we just make up a name that is reasonable, so we don't call it `tre_y2firewall_widgets_overviewtree`, that is because those id contains strange strings which don't have much use as name of variable, so in this case with `tree_overview` we don't repeat the 'tree' part.

lib/YaST/Firewall/FirewallController.pm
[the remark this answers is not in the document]
jknphy: that will be done in the future, but for now, just put this logic in a new page.

tests/yast2_gui/yast2_firewall_start_service.pm
```diff
@@ -0,0 +1,30 @@
...
+
+sub run {
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    $testapi::distri->get_firewall()->start_firewall();
+    save_screenshot;
+    $testapi::distri->get_firewall()->accept_change();
+    assert_screen 'generic-desktop';
+
+    my $self = shift;
```
jknphy, line 25: this goes always in the first line to unpack the parameter, Perl doesn't have signature but it is more clear to have them at the beginning by convention.

lib/YaST/Firewall/FirewallController.pm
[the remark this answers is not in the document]
coolgw: You mean create another control page for start_firewall/stop_firewall?
Current design is to avoid a lot of xxx_control page, so merge all business operation into one control page.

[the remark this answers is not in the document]
jknphy: sorry, I missed it was already created, this function is the wrapper, is fine.

tests/yast2_gui/yast2_firewall_stop_service.pm
```diff
@@ -0,0 +1,32 @@
...
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+
+    $testapi::distri->get_firewall()->stop_firewall();
+    save_screenshot;
+    $testapi::distri->get_firewall()->accept_change();
+    assert_screen 'generic-desktop';
+
+    $self->select_serial_terminal();
+    validate_script_output("firewall-cmd --state", sub { m/91mnot running/ }, proceed_on_failure => 1);
```
jknphy, line 29: typo?

tests/yast2_gui/yast2_firewall_settings.pm
```diff
@@ -0,0 +1,124 @@
+# SUSE's openQA tests
```
jknphy, line 1: for this test we need to break it down in different parts, avoid validation on the UI, and perform those validation with command line tools.At the moment it is impossible to know what this does without looking at the screenshots. The goal is to have something simple like when start/stop the service, we can open and close the module as needed.

jknphy: I was checking and the original test might not make sense for this part, because it is changing the interface and default zone to trusted but then right afterwards is checking in the UI in which zone is the interface, but yast module do not work like that, do not refresh with previous action without re-opening the module.
Therefore changing the interface to trusted with libyui and verify with command line could be a test module, we exclude going to interface and change the zone (because it is probably messing up with this, you need to confirm manually).
The other test module is for adding service and ports to a particular zone and its cmd verification.
Both modules should be fine for tw and sle.

tests/yast2_gui/yast2_firewall_stop_service.pm
coolgw: Not typo, the serial console will feedback special characters which used to describe font color :)

tests/yast2_gui/yast2_firewall_settings.pm
```diff
@@ -0,0 +1,114 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_guitest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
+use YaST::Module;
+use version_utils "is_sle";
```
jknphy, line 17: seems that the only part missing is to break this module in several ones, avoiding verification in UI and making it really simple as start/stop ones.

coolgw: https://openqa.opensuse.org/tests/2809501  TW
https://openqa.suse.de/t9742845  SLE

tests/yast2_gui/yast2_firewall_set_service_port.pm
```diff
@@ -0,0 +1,49 @@
...
+    save_screenshot;
+    $testapi::distri->get_firewall()->add_tcp_port($args{port});
+    save_screenshot;
+}
+
+sub run {
+    my $self = shift;
+    my $iface = iface();
+
+    record_info('Zones', "Configure zone adding service and port");
```
jknphy, line 38: let's remove all those `record_info` with this atomicity are not needed, we can just refer to the description of the test.

tests/yast2_gui/yast2_firewall_set_service_port.pm
```diff
@@ -0,0 +1,49 @@
...
+use network_utils 'iface';
+use YaST::Module;
+use version_utils "is_sle";
+
+sub verify_interface {
+    my (%args) = @_;
+    validate_script_output("firewall-cmd --list-interfaces --zone=$args{zone}", sub { m/$args{device}/ }, proceed_on_failure => 1);
+}
+
+sub configure_zone {
```
jknphy, line 24: let's avoid to create additional methods in the test layer, that is another bad choice of the design of test in general, if they are small and simple enough we don't need them, anyway here we use it just once.

tests/yast2_gui/yast2_firewall_set_service_port.pm
```diff
@@ -0,0 +1,49 @@
...
+# Maintainer: QE YaST <qa-sle-yast@suse.de>
+
+use base "y2_module_guitest";
+use strict;
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
+use YaST::Module;
+use version_utils "is_sle";
```
jknphy, line 17: not needed. could you please check `utils` I don't think it is needed, the best approach for libyui-rest-api is always to import exactly what you need, easier for review and we avoid those strange mix between needle-based test approach and new approach.

tests/yast2_gui/yast2_firewall_set_interface.pm
```diff
@@ -0,0 +1,46 @@
...
+    save_screenshot;
+}
+
+sub run {
+    my $self = shift;
+    my $iface = iface();
+
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    change_interface_zone(device => $iface, zone => 'public');
```
jknphy, line 38: same comment than below, let's not use methods, and have the whole code here, it is simple enough, and then we avoid to contributors to try to create a library with those or something more strange.

tests/yast2_gui/yast2_firewall_set_default_zone.pm
```diff
@@ -0,0 +1,71 @@
...
+    my $iface = iface();
+
+    $self->select_serial_terminal();
+    verify_default_zone(zone => "public");
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    $testapi::distri->get_firewall()->select_interfaces_page();
+    save_screenshot;
+    # TW set default zone with iface together but sle not, so we need update iface's zone to trusted zone firstly
+    if (is_sle()) {
```
jknphy, line 56: we don't use conditions in the test layer, let's create a new test module for tumbleweed that we schedule before this one to change the interface, so later when we move the default, all is fine.

jknphy (commented):
a few details more.

jknphy:
I think once we solve the latest batch of comments we can squash commits here and update first comment with both verifications.

tests/yast2_gui/yast2_firewall_set_default_zone.pm
coolgw: I do not think make extra module is good idea since this part of code is combine with current specific test check logic(will set trusted zone to default and check the result) .

jknphy: then you need to create two different modules checking two different results, so the thing with libyui-rest-api is that we don't use `if/else` as they have snow ball effect, and other contributors keep adding conditions and the code will end unreadable. For that seems to me more simple to add this extra part as a new test module and then validation are the same, at the end in the test you want to have interface and zone aligned.

schedule/yast/opensuse/yast2_gui/yast2_gui.yaml
```diff
@@ -15,20 +15,27 @@ schedule:
   - console/prepare_test_data
   - console/consoletest_setup
   - console/hostname
+  - console/install_packages_simple
```
jknphy, line 18: why we need this two modules here if they are done in the parent:
https://openqa.opensuse.org/tests/2843930#step/install_packages_simple/1
same for sle:
https://openqa.suse.de/tests/9825529#step/install_packages_simple/1

schedule/yast/opensuse/yast2_gui/yast2_gui.yaml
```diff
@@ -15,20 +15,27 @@ schedule:
...
+  - console/firewalld_add_port
   - console/setup_libyui_running_system
+  - yast2_gui/yast2_firewall_stop_service
```
jknphy, line 21: let's not change the order where we schedule firewall tests, we might affect some other tests.

tests/yast2_gui/yast2_firewall_set_default_zone.pm
```diff
@@ -0,0 +1,42 @@
...
+
+sub run {
+    my $self = shift;
+    my $iface = iface();
+
+    $self->select_serial_terminal();
+    validate_script_output("firewall-cmd --get-default-zone", sub { m/public/ }, proceed_on_failure => 1);
+    select_console 'x11', await_console => 0;
+    YaST::Module::open(module => 'firewall', ui => 'qt');
+    $testapi::distri->get_firewall()->select_interfaces_page();
```
jknphy, line 26: from line 26 to line 34 we don't need any blank line, all in the same block, it is the core of the test, we could also apply to rest of the test modules.

jknphy (commented):
great job! just add a few minors things and ready to merge.

coolgw:
tw VR: https://openqa.opensuse.org/tests/2844641#step/yast2_firewall_set_interface/16


sle VR: https://openqa.suse.de/tests/9843511#step/yast2_firewall_set_default_zone_prepare/16
block by an issue: https://bugzilla.suse.com/show_bug.cgi?id=1204893

coolgw:
Latest result:
https://openqa.opensuse.org/tests/2844742#details
https://openqa.suse.de/tests/9843828#details

jknphy:
> Latest result: https://openqa.opensuse.org/tests/2844742#details https://openqa.suse.de/tests/9843828#details

let's always use the first comment in the PR to update the links, it is more handy, at least we have been following that for years and makes harder to miss it.

schedule/yast/opensuse/yast2_gui/yast2_gui.yaml
```diff
@@ -15,20 +15,26 @@ schedule:
   - console/prepare_test_data
   - console/consoletest_setup
   - console/hostname
+  - console/firewalld_add_port
```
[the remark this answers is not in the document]
coolgw: yes, base result of https://openqa.opensuse.org/tests/2844742#step/firewalld_add_port/2 it show port already opened.

coolgw:
> > Latest result: https://openqa.opensuse.org/tests/2844742#details https://openqa.suse.de/tests/9843828#details
> 
> let's always use the first comment in the PR to update the links, it is more handy, at least we have been following that for years and makes harder to miss it.

udpated

schedule/yast/yast2_gui/yast2_gui_sle.yaml
```diff
@@ -23,13 +23,19 @@ conditional_schedule:
   yast2_modules:
     ARCH:
       x86_64:
+        - console/firewalld_add_port
```
jknphy, line 26: same here.

test_data/yast/yast2_gui.yaml
```diff
@@ -28,3 +28,7 @@ bootcode_applied_params:
 bootcode_device:
   default: /dev/vda
   write_to_partition: /dev/vda2
+install_packages:
```
jknphy, line 31: we don't need this anymore I guess.

tests/yast2_gui/yast2_firewall_set_default_zone_prepare.pm
```diff
@@ -0,0 +1,35 @@
...
+use warnings;
+use testapi;
+use utils;
+use network_utils 'iface';
+use YaST::Module;
+
+sub run {
+    my $self = shift;
+    my $iface = iface();
+    my %setting = (device => $iface, zone => 'trusted');
```
jknphy, line 21: not sure why to use a hash, you already need a variable above for $iface, so just adding another one with $zone, but you can also hardcode in the calls and only have $iface, not zone.

jknphy, line 21: s/setting/settings/

jknphy (approved):

DimStar77:
https://openqa.opensuse.org/tests/2850773#step/yast2_firewall_stop_service/8

Something seems to have gone wrong with VRs?

coolgw:
> https://openqa.opensuse.org/tests/2850773#step/yast2_firewall_stop_service/8
> 
> Something seems to have gone wrong with VRs?

strange, latest VR show good result:
https://openqa.opensuse.org/tests/2848570

jknphy:
opensusebasetest only import:
`use serial_terminal 'get_login_message';` but not sure how it works in your VR then...

coolgw:
fix by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15836

PR #15499, opened 2022-09-09
Add workaround for bsc#1184679 on wizard_hana_install

alvarocarvajald:
Module `sles4sap/wizard_hand_install` is currently failing on some specific bare metal Dell servers on `BACKEND=ipmi` with a crash on the Gnome session that takes the SUT back to the displaymanager login screen. Since the related bug hasn't been active in a while, this adds a workaround and a softfail.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1184679
- Needles: N/A
- Verification run: [ipmi](https://openqa.suse.de/t9485609), [x86_64 gnome](http://mango.qa.suse.de/tests/4963), [x86_64 textmode](http://mango.qa.suse.de/tests/4966), [ppc64le gnome](https://openqa.suse.de/t9485611), [ppc64le textmode](https://openqa.suse.de/t9485614)

### DIFF
--- a/tests/sles4sap/wizard_hana_install.pm
+++ b/tests/sles4sap/wizard_hana_install.pm
@@ -78,12 +78,16 @@ sub run {
     type_password $sles4sap::instance_password;
     wait_screen_change { send_key 'tab' };
     type_password $sles4sap::instance_password;
-    if (is_sle('<15-SP4')) {
-        wait_screen_change { send_key $cmd{ok} };
-    } else {
-        wait_screen_change { send_key $cmd{next} };
+    wait_screen_change { send_key is_sle('<15-SP4') ? $cmd{ok} : $cmd{next} };
+    while (1) {
+        assert_screen ['sap-wizard-profile-ready', 'displaymanager', 'dm-nousers'], 300;
+        if (match_has_tag 'sap-wizard-profile-ready') { last; }
+        else {
+            record_soft_failure 'bsc#1184679 - Crash of Gnome session during wizard_hana_install';
+            # Gnome session crashed. Need to log in again
+            $self->handle_displaymanager_login(ready_time => 60);
+        }
     }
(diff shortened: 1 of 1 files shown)


alvarocarvajald:
Labeling this as WIP until the VR on ipmi passes.

bs-suse:
LGTM

alvarocarvajald:
> * Verification run: [ipmi](https://openqa.suse.de/t9485609)

Seems like it's not that easy to recover from the gnome crash.

lilyeyes:
The VR is still failed on ipmi.

alvarocarvajald:
> The VR is still failed on ipmi.

Yes, seems like there's no way around this issue. I even attempted getting into the developer mode to see if there is a way to recover, and by the time of the crash, even the wizard is gone.

Some more VRs on IPMI:

https://openqa.suse.de/tests/9496774
https://openqa.suse.de/tests/9497814

I am trying one last time, but if there's no way to recover, I don't think it makes sense to merge this.

alvarocarvajald:
It is not possible to reproduce the issue when connecting directly to the console via the management console, and on osd's developer mode VNC connection is taken up by the worker so direct connection from a VNC client to the server is not possible, and VNC to the worker is not working (connection refused for some reason). I am afraid we would need to temporarily disable the worker, boot the bare metal system, enable VNC and attempt the test manually _just_ to confirm whether a workaround is even possible.

alvarocarvajald:
Attempting to handle the `oh no, something has gone wrong` screen.

VR: https://openqa.suse.de/tests/9503042

alvarocarvajald:
Closing this.

Workaround is not possible as it seems there is no way to reestablish a valid X session after the crash, and probably even if this is accomplished, installation process windows would probably be not present in the new session, which will cause the test module to also fail. Seems best way to track the failure in this scenario is with the current test code.

PR #15501, opened 2022-09-11
publiccloud : Move AHB test into maintenance and latest exec branches

asmorodskyi:


asmorodskyi:
VR : http://autobot.qa.suse.de/tests/3318

grisu48:
What is AHB? Also the maintenance VR is broken.

asmorodskyi:
> What is AHB? 

Azure Hybrid Benefit https://progress.opensuse.org/issues/114793

> Also the maintenance VR is broken.

this PR is exclusively about `main_publiccloud.pm` so the fact we got right set of modules is sufficient ( and we did get right one ) . Also failure is due missing updates which is expected because test was triggered manually so don't have all need variables

grisu48 (approved):

PR #15502, opened 2022-09-11
publiccloud : switch to python39

asmorodskyi:
https://progress.opensuse.org/issues/116440

### DIFF
--- a/tests/publiccloud/prepare_tools.pm
+++ b/tests/publiccloud/prepare_tools.pm
@@ -33,8 +33,8 @@ sub install_in_venv {
     assert_script_run(sprintf('curl -f -v %s/data/publiccloud/venv/%s.txt > /tmp/%s.txt', autoinst_url(), $binary, $binary)) if defined($args{requirements});
 
     my $venv = '/root/.venv_' . $binary;
-    assert_script_run("virtualenv '$venv'");
-    assert_script_run(". '$venv/bin/activate'");
+    assert_script_run("python3.9 -m venv $venv");
+    assert_script_run("source '$venv/bin/activate'");
     my $what_to_install = defined($args{requirements}) ? sprintf('-r /tmp/%s.txt', $binary) : $args{pip_packages};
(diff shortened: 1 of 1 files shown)


asmorodskyi:
VRs : 
- image creation job http://autobot.qa.suse.de/tests/3316
usage of created image on Azure : 
- http://autobot.qa.suse.de/tests/3319
- http://autobot.qa.suse.de/tests/3321

pdostal (approved):

grisu48:
I need a bit more context. Why do we need this? Is there a ticket for this?

Why do we switch to python 3.9 and not 3.10 (I guess because 3.10 is not yet available on 15-SP3?) What about the other python packages, that remain on 3.6, namely `python3-img-proof python3-img-proof-tests`?

asmorodskyi:
> I need a bit more context. Why do we need this? Is there a ticket for this?

it is based on https://progress.opensuse.org/issues/115361 . But initial problem raised in ticket concluded as one-time issue so I rejected ticket, nevertheless I decided to get rid of this python 3.6 warnings

> Why do we switch to python 3.9 and not 3.10 (I guess because 3.10 is not yet available on 15-SP3?) 

I am not sure I haven't checked  , I just didn't want to switch to latest version due to stability concerns 


> What about the other python packages, that remain on 3.6, namely `python3-img-proof python3-img-proof-tests`?

they don't have any other version so we don't have much options here . Only thing which we can do is to drop packaging and switch to clone repos and make things work "manually". I think it will work but I keep it as last resort when we get something bigger than just warnings ...

ilausuch (approved):

grisu48:
OK for now to get rid of the warning.

However this is getting messy. We are having now two different python versions, are using pip and virtual environments. The approach "it works for now" is unlikely to give us stability. As of now this looks like a hot potato that we just keep patching.

Perhaps it would be a good idea to switch to 15-SP4 and when doing so explore options to unify the workflow but that's beyond the scope of this PR.

grisu48 (approved):

PR #15506, opened 2022-09-12
[Trento][HANA SR] - Qesap deployment library

lpalovsky:
This PR is extending qesapdeployment library needed for Trento and 
HANAsr projects. The library provides functions to prepare the 
environment and deploy public cloud infrastructure using 'qesap.py' 
deployment script.

- Related ticket: https://jira.suse.com/browse/TEAM-6642
- Needles: N/A
- Verification run:
  GCP: https://mordor.suse.cz/tests/4590#
  Azure: https://mordor.suse.cz/tests/4581
  
  * VRs failures are unrelated to the testmodule, but to ansible configuration.

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -12,16 +12,16 @@
 
 =head1 NAME
 
-qe-sap-deployment test lib
+    qe-sap-deployment test lib
 
 =head1 COPYRIGHT
 
-Copyright 2022 SUSE LLC
-SPDX-License-Identifier: FSFAP
+    Copyright 2022 SUSE LLC
+    SPDX-License-Identifier: FSFAP
 
 =head1 AUTHORS
 
-QE SAP <qe-sap@suse.de>
+    QE SAP <qe-sap@suse.de>
 
 =cut
 
@@ -29,13 +29,13 @@ package qesapdeployment;
 
 use strict;
 use warnings;
+use utils 'file_content_replace';
 use testapi;
 use Exporter 'import';
+use Data::Dumper;
 
-# Constants
-use constant DEPLOYMENT_DIR => get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment');
-use constant QESAP_GIT_CLONE_LOG => '/tmp/git_clone.log';
-use constant PIP_INSTALL_LOG => '/tmp/pip_install.log';
+
+my @log_files = ();
 
 # Terraform requirement
 #  terraform/azure/infrastructure.tf  "azurerm_storage_account" "mytfstorageacc"
@@ -43,29 +43,44 @@ use constant PIP_INSTALL_LOG => '/tmp/pip_install.log';
 # and must be between 3 and 24 characters long
 use constant QESAPDEPLOY_PREFIX => 'qesapdep';
 
-my @log_files = ();
-
 our @EXPORT = qw(
   qesap_create_folder_tree
   qesap_pip_install
   qesap_upload_logs
   qesap_get_deployment_code
+  qesap_prepare_env
+  qesap_execute
+  qesap_yaml_replace
 );
-
 
 =head1 DESCRIPTION
 
-Package with common methods and default or constant  values for qe-sap-deployment
-
+    Package with common methods and default or constant  values for qe-sap-deployment
 =head2 Methods
+
+
+=head3 qesap_get_file_paths
+
+    Returns a hash containing file paths for config files
+=cut
+
+sub qesap_get_file_paths {
+    my %paths;
+    $paths{qesap_conf_filename} = get_required_var('QESAP_CONFIG_FILE');
+    $paths{deployment_dir} = get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment');
+    $paths{terraform_dir} = get_var('PUBLIC_CLOUD_TERRAFORM_DIR', $paths{deployment_dir} . '/terraform/');
+    $paths{qesap_conf_trgt} = $paths{deployment_dir} . "/scripts/qesap/" . $paths{qesap_conf_filename};
+    return (%paths);
+}
+
 =head3 qesap_create_folder_tree
 
-Create all needed folders
-
+    Create all needed folders
 =cut
 
 sub qesap_create_folder_tree {
-    assert_script_run('mkdir -p ' . DEPLOYMENT_DIR, quiet => 1);
+    my %paths = qesap_get_file_paths();
+    assert_script_run("mkdir -p $paths{deployment_dir}", quiet => 1);
 }
 
 =head3 qesap_pip_install
@@ -77,51 +92,56 @@ sub qesap_create_folder_tree {
 sub qesap_pip_install {
     enter_cmd 'pip config --site set global.progress_bar off';
     my $pip_ints_cmd = 'pip install --no-color --no-cache-dir ';
+    my $pip_install_log = '/tmp/pip_install.txt';
+    my %paths = qesap_get_file_paths();
+
     # Hack to fix an installation conflict. Someone install PyYAML 6.0 and awscli needs an older one
-    push(@log_files, PIP_INSTALL_LOG);
-    assert_script_run($pip_ints_cmd . 'awscli==1.19.48 | tee ' . PIP_INSTALL_LOG, 180);
-    assert_script_run($pip_ints_cmd . '-r ' . DEPLOYMENT_DIR . '/requirements.txt | tee -a ' . PIP_INSTALL_LOG, 180);
+    push(@log_files, $pip_install_log);
+    record_info("QESAP repo", "Installing pip requirements");
+    assert_script_run(join(" ", $pip_ints_cmd, 'awscli==1.19.48 | tee', $pip_install_log), 180);
+    assert_script_run(join(" ", $pip_ints_cmd, '-r', $paths{deployment_dir} . '/requirements.txt | tee -a', $pip_install_log), 180);
 }
 
 =head3 qesap_upload_logs
 
-    collect and upload logs (pip, qesap, tfvars, config.yaml)
+    qesap_upload_logs([failok=1])
 
-=over 2
+    Collect and upload logs present in @log_files.
+    failok - continue even in case upload fails
 
-=item B<QESAPREPO> - String path of the local clone of qe-sap-deployment
-
-=item B<FAILOK> - used as failok for the upload_logs
-
-=back
 =cut
 
 sub qesap_upload_logs {
-    my ($self, $failok) = @_;
+    my (%args) = @_;
+    my $failok = $args{failok};
     record_info("Uploading logfiles", join("\n", @log_files));
     for my $file (@log_files) {
         upload_logs($file, failok => $failok);
     }
+    # Remove already uploaded files from arrays
+    @log_files = ();
 }
-
 
 =head3 qesap_get_deployment_code
 
-Get the qe-sap-deployment code
+    Get the qe-sap-deployment code
 =cut
 
 sub qesap_get_deployment_code {
-    record_info("QESAP repo", "Preparing qe-sap-deployment repository");
-
     my $git_repo = get_var(QESAPDEPLOY_GITHUB_REPO => 'github.com/SUSE/qe-sap-deployment');
-    enter_cmd "cd " . DEPLOYMENT_DIR;
+    my $qesap_git_clone_log = '/tmp/git_clone.txt';
+    my %paths = qesap_get_file_paths();
+
+    record_info("QESAP repo", "Preparing qe-sap-deployment repository");
+    qesap_create_folder_tree();
+    enter_cmd "cd " . $paths{deployment_dir};
 
     # Script from a release
     if (get_var('QESAPDEPLOY_VER')) {
         my $ver_artifact = 'v' . get_var('QESAPDEPLOY_VER') . '.tar.gz';
 
         my $curl_cmd = "curl -v -L https://$git_repo/archive/refs/tags/$ver_artifact -o$ver_artifact";
-        assert_script_run("set -o pipefail ; $curl_cmd | tee " . QESAP_GIT_CLONE_LOG, quiet => 1);
+        assert_script_run("set -o pipefail ; $curl_cmd | tee " . $qesap_git_clone_log, quiet => 1);
 
         my $tar_cmd = "tar xvf $ver_artifact --strip-components=1";
         assert_script_run($tar_cmd);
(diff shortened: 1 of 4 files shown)


lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+    for my $variable (keys %variables) {
+        $replaced_variables{"%" . $variable . "%"} = $variables{$variable};
+    }
+
+    file_content_replace(QESAP_CONF_TRGT, %replaced_variables);
+}
+
+=head3 qesap_execute
+
+    execute qesap glue script commands
```
mpagot, line 142: I propose to add an example about how to call it

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+sub qesap_prepare_env {
+    my $pip_cmd = "pip install --no-color --no-cache-dir --progress-bar off ";
+    my $tfvars_template = get_var('QESAP_TFVARS_TEMPLATE');
+    my $curl = "curl -v -L ";
+
+    qesap_get_deployment_code();
+
+    record_info("QESAP repo", "Installing pip requirements");
+    assert_script_run($pip_cmd . " -r " . DEPLOYMENT_DIR . "/requirements.txt > /root/pip_requirements.log");
+    push(@log_files, "/root/pip_requirements.log");
```
mpagot, line 176: I suggest to push before to call

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+    # Copy tfvars template file if defined in parameters
+    if (get_var('QESAP_TFVARS_TEMPLATE')) {
+        record_info("QESAP tfvars template", "Preparing terraform template: \n" . $tfvars_template);
+        assert_script_run('cd ' . TERRAFORM_DIR, quiet => 1);
+        assert_script_run('cp ' . $tfvars_template . ' terraform.tfvars.template');
+    }
+
+    record_info("QESAP yaml", "Preparing yaml config file");
+    assert_script_run($curl . data_url(QESAP_CONF_SRC) . ' -o ' . QESAP_CONF_TRGT);
+    qesap_yaml_replace();
```
alvarocarvajald, line 195: If this working? I thought `qesap_yaml_replace()` received at least one argument.

lib/qesapdeployment.pm
```diff
@@ -91,7 +102,103 @@ sub qesap_get_deployment_code {
...
+
+    Replaces yaml config file variables with the ones defined in openqa.
+    Openqa variables need to be added as 'openqa_variables' inside %args as following:
+    <OPENQA_VARIABLE_UPPERCASE> = "<VARIABLE_VALUE>"
+
+=cut
+
+sub qesap_yaml_replace {
+    my (%run_args) = @_;
+    my %variables = $run_args{openqa_variables};
```
alvarocarvajald, line 134: Isn't it better to pass a list of settings to replace and iterate with `get_var()` instead of duplicating these that are already defined in `%bmwqemu::vars`? I'd like to understand why they are needed in `%run_args` since all vars should be available to all test modules all the time.

lib/qesapdeployment.pm
alvarocarvajald: Agree if `post_fail_hook` will try to upload all logs defined in `@log_files`; otherwise, it makes no difference.

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
         push(@log_files, QESAP_GIT_CLONE_LOG);
         assert_script_run("set -o pipefail ; $git_clone_cmd | tee " . QESAP_GIT_CLONE_LOG, quiet => 1);
     }
+    assert_script_run("ln -s " . TERRAFORM_DIR . "/aws " . TERRAFORM_DIR . "/ec2");
```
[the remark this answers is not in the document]
lpalovsky: This is a problem with terraform directory naming in qe-sap-deployment project and what is expected by PC terraform library. 
In qesap-deployment project uses 'aws' but openqa library works with 'ec2'.
The least intrusive way for both to be working is a symlink. 

Google is fine since terraform directories are the same. If you use 'PUBLIC_CLOUD_PROVIDER': GCP there is no need for a symlink.

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+}
+
+=head3 qesap_upload_logs
+
+    Collect and upload logs present in @log_files.
+=cut
+sub qesap_upload_logs {
```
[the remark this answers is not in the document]
lpalovsky: Thanks. Fixed.

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+    @log_files = ();
+}
+
+=head3 qesap_yaml_replace
+
+    Replaces yaml config file variables with the ones defined in openqa.
+    Openqa variables need to be added as 'openqa_variables' inside %args as following:
+    <OPENQA_VARIABLE_UPPERCASE> = "<VARIABLE_VALUE>"
+
+=cut
```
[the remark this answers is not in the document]
lpalovsky: Fixed

lib/qesapdeployment.pm
mpagot: >  iterate with get_var() instead 

not necessary all of them come from get_var

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+    for my $file (@log_files) {
+        upload_logs($file, failok => 1);
+    }
+    # Remove uploaded files from arrays
+    @log_files = ();
+}
+
+=head3 qesap_yaml_replace
+
+    Replaces yaml config file variables with the ones defined in openqa.
```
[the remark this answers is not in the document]
lpalovsky: What about this:
```
Replaces yaml config file variables with parameters defined by OpenQA testode, yaml template or yaml schedule.
```

lib/qesapdeployment.pm
alvarocarvajald: Yes, but it's duplicating memory without it being necessary AFAICS. All settings are in memory, available to all modules in  `%bmwqemu::vars`, then there some of these are duplicated in `%run_args{openqa_variables}` after calling `qesap_get_variables()`, and this copy it's the one used here. The question is why do we need to duplicate these, when the ones in `%bmwqemu::vars` are accessible to all modules via `get_var()`. If the duplication is introduced only to have a subset with which to iterate in the code of `qesap_yaml_replace()`, then the same can be accomplished with a list of the settings to replace and `get_var()`

lib/qesapdeployment.pm
lpalovsky: Yep. Function is taking the hash of variables from $run_args{openqa_variables} so it is up to the test code to take care of it. But thinking about it now... there is no failure defined if the hash is empty or not defined at all.

lib/qesapdeployment.pm
```diff
@@ -91,6 +102,95 @@ sub qesap_get_deployment_code {
...
+}
+
+=head3 qesap_upload_logs
+
+    Collect and upload logs present in @log_files.
+=cut
+sub qesap_upload_logs {
+    record_info("Uploading logfiles", join("\n", @log_files));
+    for my $file (@log_files) {
+        upload_logs($file, failok => 1);
```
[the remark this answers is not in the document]
lpalovsky: You mean that qesap_upload_logs would not be called from the library itself but only by testcode?

lib/qesapdeployment.pm
lpalovsky: So the idea is that library will only append log files into array but the qesap_upload_logs(); will be called from outside of the library? byt testcode, fail hooks etc...

lib/qesapdeployment.pm
lpalovsky: > Isn't it better to pass a list of settings to replace and iterate with `get_var()` instead of duplicating these that are already defined in `%bmwqemu::vars`? I'd like to understand why they are needed in `%run_args` since all vars should be available to all test modules all the time.

Yeah I have done it like that after suggestion from @mpagot that the variables should be handled by testcode for each project separately.
With passing only a list there is a problem with defining default values like:

```
    $variables{FENCING_MECHANISM} = get_var("FENCING_MECHANISM", "sbd");
```

Not the best example but I saw in the old code parameters defined like:

```
my $sle_version = get_var('FORCED_DEPLOY_REPO_VERSION') ? get_var('FORCED_DEPLOY_REPO_VERSION') : get_var('VERSION');
```
But the point is not to be limited only by output form get_var()

lib/qesapdeployment.pm
mpagot: I suggest to do 
```
push(@log_files, "/root/pip_requirements.log");
assert_script_run($pip_cmd . " -r " . DEPLOYMENT_DIR . "/requirements.txt > /root/pip_requirements.log");
```

in place of 
```
assert_script_run($pip_cmd . " -r " . DEPLOYMENT_DIR . "/requirements.txt > /root/pip_requirements.log");
push(@log_files, "/root/pip_requirements.log");
```

so that in case of assert_script_run failure, you keep to have the log

lib/qesapdeployment.pm
mpagot: Is it feasible to have an interface like:

```
# Here we are in th etest code
my variables_for_this_test = here a map with all needed variables and values
qesap_yaml_replace(%variables_for_this_test);
```

alvarocarvajald: I think this is what Lumir is providing with `qesap_get_variables()`

lib/qesapdeployment.pm
lpalovsky: Ah ok... I understand now. 
Fixed.

lib/qesapdeployment.pm
lpalovsky: Correct. Everything needed is just to define variables in $run_args{openqa_variables}.
I have added an example how to use it as well.

lib/qesapdeployment.pm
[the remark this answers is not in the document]
mpagot: this improvement can be added later

lib/qesapdeployment.pm
```diff
@@ -91,7 +103,115 @@ sub qesap_get_deployment_code {
...
+
+=head3 qesap_yaml_replace
+
+    Replaces yaml config file variables with parameters defined by OpenQA testode, yaml template or yaml schedule.
+    Openqa variables need to be added as a hash with key/value pair inside %run_args{openqa_variables}.
+    Example:
+        my %variables;
+        $variables{HANA_SAR} = get_required_var("HANA_SAR");
+        $variables{HANA_CLIENT_SAR} = get_required_var("HANA_CLIENT_SAR");
+        $run_args{openqa_variables} = %variables;
```
mpagot, line 133: is line 133 of this example needed?

mpagot (approved):
LGTM

lib/qesapdeployment.pm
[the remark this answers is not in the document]
lpalovsky: OK, got it. I added 'failok' variable.
I am thinking about adding another one 'log_files' which will allow to add another list of variables from outside of lib module. 
But that can wait for now.

lib/qesapdeployment.pm
lpalovsky: Oh thanks. Good catch... that was a remainder of the old code.
Removed.

lib/qesapdeployment.pm
```diff
@@ -91,7 +101,122 @@ sub qesap_get_deployment_code {
...
+    For variables example see 'qesap_yaml_replace'
+=cut
+
+sub qesap_prepare_env {
+    my (%args) = @_;
+    my $variables = $args{openqa_variables};
+    my $pip_cmd = "pip install --no-color --no-cache-dir --progress-bar off ";
+    my $tfvars_template = get_var('QESAP_TFVARS_TEMPLATE');
+    my $curl = "curl -v -L ";
+
```
mpagot, line 198: maybe here you like to call qesap_create_folder_tree ?

mpagot (approved):
LGTM

lib/qesapdeployment.pm
```diff
@@ -136,7 +143,107 @@ sub qesap_get_deployment_code {
...
+    Execute qesap glue script commands. Check project documentation for available options:
+    https://github.com/SUSE/qe-sap-deployment
+=cut
+
+sub qesap_execute {
+    my (%args) = @_;
+    my $verbose = $args{verbose} ? "--verbose" : "";
+    my $qesap_cmd = DEPLOYMENT_DIR . "/scripts/qesap/qesap.py " . $verbose . " -c " . QESAP_CONF_TRGT . " -b " . DEPLOYMENT_DIR;
+    my $command = defined($args{cmd}) ? $args{cmd} : die("QESAP command to execute undefined");
+    my $exec_log = "/root/qesap_exec_" . $command . ".log.txt";
```
mpagot, line 191: other code part are using /tmp/

lib/qesapdeployment.pm
```diff
@@ -136,7 +143,107 @@ sub qesap_get_deployment_code {
...
+        assert_script_run('cd ' . TERRAFORM_DIR . PROVIDER, quiet => 1);
+        assert_script_run('cp ' . $tfvars_template . ' terraform.tfvars.template');
+    }
+
+    record_info("QESAP yaml", "Preparing yaml config file");
+    assert_script_run($curl . data_url(QESAP_CONF_SRC) . ' -o ' . QESAP_CONF_TRGT);
+    qesap_yaml_replace(openqa_variables => $variables);
+
+    record_info("QESAP conf", "Generating tfvars file");
+    qesap_execute(cmd => 'configure', quiet => 1);
```
mpagot, line 244: `quiet` parameter is not handled by qesap_execute

lib/qesapdeployment.pm
```diff
@@ -136,7 +144,104 @@ sub qesap_get_deployment_code {
...
+    qesap_execute(cmd => $qesap_script_cmd [, verbose => 1] );
+
+    Execute qesap glue script commands. Check project documentation for available options:
+    https://github.com/SUSE/qe-sap-deployment
+=cut
+
+sub qesap_execute {
+    my (%args) = @_;
+    my $verbose = $args{verbose} ? "--verbose" : "";
+    my $qesap_cmd = DEPLOYMENT_DIR . "/scripts/qesap/qesap.py " . $verbose . " -c " . QESAP_CONF_TRGT . " -b " . DEPLOYMENT_DIR;
```
alvarocarvajald, line 190: Suggestion:
```
my $qesap_cmd = join(' ', DEPLOYMENT_DIR, '/scripts/qesap/qesap.py', $verbose, '-c', QESAP_CONF_TRGT, '-b', DEPLOYMENT_DIR);
```

lib/qesapdeployment.pm
```diff
@@ -136,7 +144,104 @@ sub qesap_get_deployment_code {
...
+    my $command = defined($args{cmd}) ? $args{cmd} : die("QESAP command to execute undefined");
```
alvarocarvajald, line 191: Suggestion:
```
die 'QESAP command to execute undefined' unless $args{cmd};
```
And then replace `$command` for `$args{cmd}` in lines 192, 195, 198 & 201.

BTW, `defined` checks for the value to be defined within the hash, so it returns true if said value is 0 or the empty string. As I don't suppose either of those can be a command, you can drop `defined`.

lib/qesapdeployment.pm
```diff
@@ -136,7 +144,104 @@ sub qesap_get_deployment_code {
...
+    my $exec_log = "/tmp/qesap_exec_" . $command . ".log.txt";
+    push(@log_files, $exec_log);
+
+    record_info('QESAP exec', 'Executing: \n' . $qesap_cmd . " " . $command);
+
+    if (defined($args{timeout})) {
```
alvarocarvajald, line 197: Same suggestion about `defined`: use it here if a timeout of 0 is expected, otherwise drop the `defined`.

lib/qesapdeployment.pm
lpalovsky: OK, moved all logs from /root/ to /tmp

lib/qesapdeployment.pm
```diff
@@ -79,35 +88,34 @@ sub qesap_pip_install {
...
+    Collect and upload logs present in @log_files.
+    failok - continue even in case upload fails
 
-=back
 =cut
 
 sub qesap_upload_logs {
-    my ($self, $failok) = @_;
+    my (%args) = @_;
+    my $failok = $args{failok} ? $args{failok} : 0;
```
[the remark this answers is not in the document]
lpalovsky: Ah ok... I thought perl would complain I am referencing to hash key that does not exist. 
Will remove that one.

lib/qesapdeployment.pm
```diff
@@ -50,18 +57,20 @@ our @EXPORT = qw(
   qesap_pip_install
   qesap_upload_logs
   qesap_get_deployment_code
+  qesap_prepare_env
+  qesap_execute
+  qesap_yaml_replace
 );
```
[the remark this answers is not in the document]
lpalovsky: OK

lib/qesapdeployment.pm
lpalovsky: OK thanks. Like previously I thought perl would not like it if for example $args{something} was not specified from calling function and I would point to this key.

mpagot (approved):
LGTM

lib/qesapdeployment.pm
lpalovsky: Hmm... that is a good point. But that might be only needed if not using git clone. 
Git clone does create the directory by itself so somewhere in qesap_get_deployment_code.

mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM

mpagot (approved):
LGTM

PR #15508, opened 2022-09-12
Migrate create_hdd_multipath_lvm_textmode to autoyast

Avinesh:
Signed-off-by: Avinesh Kumar <akumar@suse.de>

Autoyast profiles and yaml schedule for creating image with multipath and lvm.

- Related ticket: https://progress.opensuse.org/issues/113596
- Verification runs:
  Jobs which create the hdd images and dependent jobs -
x86_64: https://openqa.suse.de/tests/9484677 => https://openqa.suse.de/tests/9496865#
aarch64: https://openqa.suse.de/tests/9484680 => https://openqa.suse.de/tests/9496871#
ppc64le: https://openqa.suse.de/tests/9485083 => https://openqa.suse.de/tests/9496866#

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
+++ b/data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
@@ -1,0 +1,402 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <bootloader t="map">
+    <global t="map">
+      <cpu_mitigations>auto</cpu_mitigations>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <secure_boot>true</secure_boot>
+      <terminal>gfxterm</terminal>
+      <timeout t="integer">-1</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>vga=gfx-1024x768x16</xen_kernel_append>
+    </global>
+    <loader_type>grub2-efi</loader_type>
+  </bootloader>
+  <firewall t="map">
+    <default_zone>public</default_zone>
+    <enable_firewall t="boolean">true</enable_firewall>
+    <log_denied_packets>off</log_denied_packets>
+    <start_firewall t="boolean">true</start_firewall>
+    <zones t="list">
+      <zone t="map">
+        <description>For use in public areas. You do not trust the other computers on networks to not harm your computer. Only selected incoming connections are accepted.</description>
+        <interfaces t="list">
+          <interface>eth0</interface>
+        </interfaces>
+        <masquerade t="boolean">false</masquerade>
+        <name>public</name>
+        <ports t="list"/>
+        <protocols t="list"/>
+        <services t="list">
+          <service>dhcpv6-client</service>
+        </services>
+        <short>Public</short>
+        <target>default</target>
+      </zone>
+    </zones>
+  </firewall>
+  <general t="map">
+    <mode t="map">
+      <confirm t="boolean">false</confirm>
+    </mode>
+    <storage t="map">
+      <start_multipath t="boolean">true</start_multipath>
+    </storage>
+  </general>
+  <groups t="list">
+    <group t="map">
+      <gid>100</gid>
+      <groupname>users</groupname>
+      <userlist/>
+    </group>
+  </groups>
+  <host t="map">
+    <hosts t="list">
+      <hosts_entry t="map">
+        <host_address>127.0.0.1</host_address>
+        <names t="list">
+          <name>localhost</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>::1</host_address>
+        <names t="list">
+          <name>localhost ipv6-localhost ipv6-loopback</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>fe00::0</host_address>
+        <names t="list">
+          <name>ipv6-localnet</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff00::0</host_address>
+        <names t="list">
+          <name>ipv6-mcastprefix</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff02::1</host_address>
+        <names t="list">
+          <name>ipv6-allnodes</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff02::2</host_address>
+        <names t="list">
+          <name>ipv6-allrouters</name>
+        </names>
+      </hosts_entry>
+      <hosts_entry t="map">
+        <host_address>ff02::3</host_address>
+        <names t="list">
+          <name>ipv6-allhosts</name>
+        </names>
+      </hosts_entry>
+    </hosts>
+  </host>
+  <networking t="map">
+    <dhcp_options t="map">
+      <dhclient_client_id/>
+      <dhclient_hostname_option>AUTO</dhclient_hostname_option>
+    </dhcp_options>
+    <dns t="map">
+      <dhcp_hostname t="boolean">false</dhcp_hostname>
+      <hostname>susetest</hostname>
+      <resolv_conf_policy>auto</resolv_conf_policy>
+    </dns>
+    <interfaces t="list">
+      <interface t="map">
+        <bootproto>dhcp</bootproto>
+        <name>eth0</name>
+        <startmode>auto</startmode>
+        <zone>public</zone>
+      </interface>
+    </interfaces>
+    <ipv6 t="boolean">true</ipv6>
+    <keep_install_network t="boolean">true</keep_install_network>
+    <managed t="boolean">false</managed>
+    <net-udev t="list">
+      <rule t="map">
+        <name>eth0</name>
+        <rule>KERNELS</rule>
+        <value>0000:00:03.0</value>
+      </rule>
+    </net-udev>
+    <routing t="map">
+      <ipv4_forward t="boolean">false</ipv4_forward>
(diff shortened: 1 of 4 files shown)


ge0r (approved):
LGTM!

data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
```diff
@@ -0,0 +1,402 @@
...
+          <create_subvolumes t="boolean">true</create_subvolumes>
+          <filesystem t="symbol">btrfs</filesystem>
+          <format t="boolean">false</format>
+          <lv_name>root</lv_name>
+          <mount>/</mount>
+          <mountby t="symbol">device</mountby>
+          <pool t="boolean">false</pool>
+          <quotas t="boolean">false</quotas>
+          <resize t="boolean">false</resize>
+          <size>10599006208</size>
```
b10n1k, line 156: i would rather have the size in percentage. Such as when the image size is changing the installation can adapt. wdyt?

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
```diff
@@ -0,0 +1,25 @@
+name: multipath
```
b10n1k, line 1: s/multipath/multipath_textmode/

b10n1k: or s/multipath/multipath_lvm_textmode/

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
```diff
@@ -0,0 +1,25 @@
...
+    Prepare image with multipath and lvm enabled using autoyast installation for kernel testing.
+
+vars:
+    DESKTOP: textmode
+    MULTIPATH: 1
+    LVM: 1
+    HDDMODEL: scsi-hd
+    SEPARATE_HOME: 0
+    INSTALLONLY: 1
+    AUTOYAST: autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_%ARCH%.xml
```
b10n1k, line 13: i am not sure how you plan to setup the test. in case this is a new testsuite that wouldnt make sense here. in additional mixing variables doesnt look good. for instance `HDDMODEL` should come from machine variables. Can you shed some light why do we need the `vars` section in the yaml?

b10n1k (changes requested):
in additional to the comments, can you consider to merge the xmls? that will simplify future maintainance

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
czerw: that is correct, `HDDMODEL` needs to be defined at test suite level to have impact, before is vm created (it's late in yaml), but it doesn't do any harm here.    Ir was always like this and we didn't consider to create new machine definition for it, so it is fine, like it is now.

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
Avinesh: Thank you, updating this.

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
Avinesh: Thank you @b10n1k and @czerw for the explanation. I think I just based this on the existing yaml - `create_hdd_multipath_lvm_extmode.yaml`
Please confirm if I should remove the vars other than AUTOYAST.

Avinesh:
> in additional to the comments, can you consider to merge the xmls? that will simplify future maintainance

@b10n1k , @ge0r , Do we have any documentation or example commit which can help with merging the xmls for all the archs?

data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
b10n1k: An example is https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/hpc/create_hdd_textmode_x86_64.xml.ep#L350. 
If you really need documentation i have found https://documentation.suse.com/sles/15-SP2/html/SLES-all/cha-configuration-installation-options.html#ay-partition-configuration

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
b10n1k: i am not sure what @czerw says.
However my emphasis is what reasoning the use of the variables in the yaml here. i still dont understand if we have a good explanation having the variables here. At least all of them. there are different approaches to schedule the job and i guess the main difference is the AUTOYAST. That can be defined here (depends!). But the rest of the variables are already defined in the `create_hdd_multipath_lvm_textmode`.

Here is some of the flows:
1. create a new testsuite
2. override the `create_hdd_multipath_lvm_textmode` in the jobgroup

if you do the 1 all those variables need to be in testsuite. but you kinda duplicate entries in the db
if you go with 2 all the variables except AUTOYAST do not need to be here as you can reuse the `create_hdd_multipath_lvm_textmode`. (or set them there but this is just matter of personal taste). 

Do you see my objection or do we need to have a chat?

b10n1k:
> > in additional to the comments, can you consider to merge the xmls? that will simplify future maintainance
> 
> @b10n1k , @ge0r , Do we have any documentation or example commit which can help with merging the xmls for all the archs?

 https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/hpc/create_hdd_textmode_x86_64.xml.ep

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
Avinesh: I think we are following the option 1 of creating new testsuites for autoyast based installation. And I think we can remove the existing testsuites once we have moved to the new ones completely. @czerw, please correct me if I am wrong.
So all these variables except AUTOYAST should be moved to new testsuite settings in osd?

Avinesh:
> > > in additional to the comments, can you consider to merge the xmls? that will simplify future maintainance
> > 
> > 
> > @b10n1k , @ge0r , Do we have any documentation or example commit which can help with merging the xmls for all the archs?
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/hpc/create_hdd_textmode_x86_64.xml.ep

Oh, I was thinking you are asking to merge all three archs' xmls into one, but the idea is to merge 
[multipath xml](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/create_hdd/create_hdd_multipath_textmode_x86_64.xml), and multipath+lvm xml in this PR (and then multipath+lvm+encrypt also) for each archs separately.

Avinesh:
> > > in additional to the comments, can you consider to merge the xmls? that will simplify future maintainance
> > 
> > 
> > @b10n1k , @ge0r , Do we have any documentation or example commit which can help with merging the xmls for all the archs?
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/hpc/create_hdd_textmode_x86_64.xml.ep

Thank you @b10n1k for sharing a sample patch (in email) showing how to combine the different arch xmls into single one.
But now that I understand it, I would say it is better to keep them separate as adding the arch based conditional logic (like - `% if ($check_var->('ARCH', 'ppc64le'))` in many places in xml brings in unnecessary complexities.
And anyway, we should first commit these profiles as separate files just to have complete git log history, and if we decide to combine, we should that in subsequent commits.
@czerw , please provide your inputs.

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
czerw: @b10n1k  i don't understand you too :)

czerw (approved):

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
b10n1k: As per the current situation we have `autoyast_create_hdd_multipath_lvm_encrypt_textmode` `autoyast_create_hdd_multipath_lvm_textmode` and `autoyast_create_hdd_multipath_textmode` (new test suites). 

What you can do is to
1. either remove the `DESKTOP=textmode` from there to not get confused.
2. move all the variables into the new test suites

data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
b10n1k: lets close this as per offline agreement. fix that in another round

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
Avinesh: @b10n1k , I removed the vars from the yaml in updated commit and added them in the test suite in osd.

b10n1k: @Avinesh thanks. and sorry for making this that long :)

b10n1k (approved):
i still have some suggestions but let finish with that now and do the rest later. thanks @Avinesh for your patient :)

schedule/kernel/autoyast_create_hdd_multipath_lvm_textmode.yaml
Avinesh: Thank you @b10n1k for all the suggestions.

data/autoyast_sle15/create_hdd/create_hdd_multipath_lvm_textmode_aarch64.xml
Avinesh: yes, I'll update it in the next commit.

PR #15509, opened 2022-09-12
Make test module to export results in xml

b10n1k:
Refactor test module to export the results in xml and make use of the xml representation in OpenQA.

- Verification run: http://aquarius.suse.cz/tests/12251

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -17,6 +17,7 @@ use Utils::Systemd qw(systemctl disable_and_stop_service);
 use Utils::Backends;
 use Mojo::UserAgent;
 use zypper qw(wait_quit_zypper);
+use Storable qw(dclone);
 
 our @EXPORT = qw(
   generate_results
@@ -96,6 +97,8 @@ our @EXPORT = qw(
   permit_root_ssh_in_sol
   cleanup_disk_space
   package_upgrade_check
+  test_case
+  @all_tests_results
 );
 
 =head1 SYNOPSIS
@@ -2413,7 +2416,6 @@ sub pars_results {
     # and if so, make sure the whole testsuite will fail
     my $fail_check = 0;
     for my $i (@test) {
-        record_info "i", $i;
         if ($i->{result} eq 'FAIL') {
             $fail_check++;
         }
@@ -2438,6 +2440,23 @@ sub pars_results {
         script_run("echo \"</testcase>\" >> $xmlfile");
     }
     script_run("echo \"</testsuite>\" >> $xmlfile");
+}
+
+our @all_tests_results;
+
+=head2 test_case
+    test_case($name, $description, $result);
+
+C<test_case> can produce a data_structure which C<pars_results> can utilize.
+Using C<test_case> in an OpenQA module you are able to /name/ and describe
+the whole test as subtasks, in a XUnit format.
+
+=cut
+
+sub test_case {
+    my ($name, $description, $result) = @_;
+    my %results = generate_results($name, $description, $result);
+    push(@all_tests_results, dclone(\%results));
 }
 
 1;
--- a/tests/hpc/powerman.pm
+++ b/tests/hpc/powerman.pm
@@ -16,8 +16,8 @@ use testapi;
 use utils;
 use susedistribution;
 
-
 our $cfg_file = "/etc/powerman/powerman.conf";
+our $file = 'tmpresults.xml';
 
 our $dev_caps = sub {
     my $dev = shift;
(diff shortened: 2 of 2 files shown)


foursixnine (commented):
You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...

b10n1k:
> You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...

Do you have a link to take a look? thing is that this approach is already used in another HPC module. i just try to abstracted to reuse it wherever

foursixnine:
> > You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...
> 
> Do you have a link to take a look? thing is that this approach is already used in another HPC module. i just try to abstracted to reuse it wherever

[    parse_extra_log(TAP => $tap_results);](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/afff37a3fc28da190070e15f8f481b0baf2c665b/tests/console/sqlite3.pm#L42)

and here's an example of the tap output: 

file: https://openqa.suse.de/tests/9601745/file/sqlite3-results.tap, is represented like: https://openqa.suse.de/tests/9601745 (see at the end)

b10n1k:
> > > You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...
> > 
> > 
> > Do you have a link to take a look? thing is that this approach is already used in another HPC module. i just try to abstracted to reuse it wherever
> 
> [ parse_extra_log(TAP => $tap_results);](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/afff37a3fc28da190070e15f8f481b0baf2c665b/tests/console/sqlite3.pm#L42)
> 
> and here's an example of the tap output:
> 
> file: https://openqa.suse.de/tests/9601745/file/sqlite3-results.tap, is represented like: https://openqa.suse.de/tests/9601745 (see at the end)...
> 
> File is fairly simple: https://testanything.org
> 
> ```
> 1..4
> ok 1 - Input file opened
> not ok 2 - First line of the input valid
> ok 3 - Read the rest of the file
> not ok 4 - Summarized correctly # TODO Not written yet
> ```

i took a look. From my understanding this will require to implement a different design of the module to produce the TAP output. I think this is different concept here. What we need is the openqa results which it would be represent even without TAP format. But thanks for alternative idea

foursixnine:
> > > > You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...
> > > 
> > > 
> > > Do you have a link to take a look? thing is that this approach is already used in another HPC module. i just try to abstracted to reuse it wherever
> > 
> > 
> > [ parse_extra_log(TAP => $tap_results);](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/afff37a3fc28da190070e15f8f481b0baf2c665b/tests/console/sqlite3.pm#L42)
> > and here's an example of the tap output:
> > file: https://openqa.suse.de/tests/9601745/file/sqlite3-results.tap, is represented like: https://openqa.suse.de/tests/9601745 (see at the end)...
> > File is fairly simple: https://testanything.org
> > ```
> > 1..4
> > ok 1 - Input file opened
> > not ok 2 - First line of the input valid
> > ok 3 - Read the rest of the file
> > not ok 4 - Summarized correctly # TODO Not written yet
> > ```
> 
> i took a look. From my understanding this will require to implement a different design of the module to produce the TAP output. I think this is different concept here. What we need is the openqa results which it would be represent even without TAP format. But thanks for alternative idea

Your last changes look much better :)

b10n1k:
> > > > > You could also use TAP, like it's done in the SQLite tests, which would be much easier to parse and generate...
> > > > 
> > > > 
> > > > Do you have a link to take a look? thing is that this approach is already used in another HPC module. i just try to abstracted to reuse it wherever
> > > 
> > > 
> > > [ parse_extra_log(TAP => $tap_results);](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/afff37a3fc28da190070e15f8f481b0baf2c665b/tests/console/sqlite3.pm#L42)
> > > and here's an example of the tap output:
> > > file: https://openqa.suse.de/tests/9601745/file/sqlite3-results.tap, is represented like: https://openqa.suse.de/tests/9601745 (see at the end)...
> > > File is fairly simple: https://testanything.org
> > > ```
> > > 1..4
> > > ok 1 - Input file opened
> > > not ok 2 - First line of the input valid
> > > ok 3 - Read the rest of the file
> > > not ok 4 - Summarized correctly # TODO Not written yet
> > > ```
> > 
> > 
> > i took a look. From my understanding this will require to implement a different design of the module to produce the TAP output. I think this is different concept here. What we need is the openqa results which it would be represent even without TAP format. But thanks for alternative idea
> 
> Your last changes look much better :)

Thats why it was a drafted. i was waiting another PR with related changes before finish with it. Thanks again @foursixnine

schlad (approved):
This looks imho very good; It is nicely abstracting some steps.

PR #15510, opened 2022-09-13
Changed to use 'skip_install_addons' test module

tinawang123:
Changed to use 'skip_install_addons' test module

- Related ticket: https://progress.opensuse.org/issues/113734
- Verification run: https://openqa.suse.de/tests/9501954#


jknphy:
Missing part:
- Agreed offline that we will tackle the timeout issue with additional function in corresponding controller.

jknphy:
> on slow

you're right, I merged that change some time ago, I didn't notice it sort of breaks the design.
so imagine that in the future we will have a aarch64 interactive installation in maintenance, I'm pretty sure we will need a setting with some default value and an a multiplier from test layer, otherwise will would hardcode the setting in the page level and will no be very transparent or we will put some super huge timeout that would hide bug.
https://progress.opensuse.org/issues/116521

PR #15514, opened 2022-09-13
Disable kernel messages on console and wrap terraform container into bash -e

grisu48:
This PR contains two commits:

1. Disable kernel messages on the console for publiccloud test runs
2. Wrap terraform container into bash -e so that it terminates with non-success error code on errors.

- Verification run: [image creation](https://duck-norris.qam.suse.de/tests/10707) | [img-proof](https://duck-norris.qam.suse.de/tests/10712) (using the new image) | [fixed console](https://duck-norris.qam.suse.de/tests/10713#step/prepare_instance/117) fails in unrelated later step

### DIFF
--- a/tests/publiccloud/prepare_tools.pm
+++ b/tests/publiccloud/prepare_tools.pm
@@ -114,11 +114,12 @@ sub run {
     my $terraform_version = '1.1.7';
     # Terraform in a container
     my $terraform_wrapper = <<EOT;
-#!/bin/sh
-podman run -v /root/:/root/ --rm --env-host=true -w=\$PWD docker.io/hashicorp/terraform:$terraform_version \$@
+#!/bin/bash -e
+podman run --rm -w=\$PWD -v /root/:/root/ --env-host=true docker.io/hashicorp/terraform:$terraform_version \$@
 EOT
 
-    create_script_file('terraform', '/usr/bin/terraform', $terraform_wrapper);
+    create_script_file('terraform', '/usr/local/bin/terraform', $terraform_wrapper);
+    validate_script_output("terraform -version", qr/$terraform_version/);
     record_info('Terraform', script_output('terraform -version'));
 
     # Kubectl in a container
--- a/tests/publiccloud/prepare_instance.pm
+++ b/tests/publiccloud/prepare_instance.pm
@@ -61,6 +61,11 @@ sub run {
 
     select_host_console();    # select console on the host, not the PC instance
 
+    # Prevent kernel messages of the helper VM to contaminate the serial console
+    # this is needed on our ext4-based helper VM to avoid kernel messages from overlayfs
+    # to confuse openQA
+    assert_script_run("dmesg -n emerg");
+
     my $additional_disk_size = get_var('PUBLIC_CLOUD_HDD2_SIZE', 0);
     my $additional_disk_type = get_var('PUBLIC_CLOUD_HDD2_TYPE', '');    # Optional variable, also if PUBLIC_CLOUD_HDD2_SIZE is set
 


asmorodskyi (changes requested):
please elaborate what benefits we will get from BTRFS on qcow image which is destroyed after every run ?

grisu48:
> please elaborate what benefits we will get from BTRFS on qcow image which is destroyed after every run ?

We don't have to use the overlay driver, which is responsible for the error message that cause all the fuzz. See https://progress.opensuse.org/issues/116332#note-8

lib/bootloader_setup.pm
```diff
@@ -353,7 +353,7 @@ sub get_bootmenu_console_params {
 
     # See bsc#1011815, last console set as boot parameter is linked to /dev/console
     # and doesn't work if set to serial device. Don't want this on some backends.
-    push @params, "console=tty" unless (get_var('BACKEND', '') =~ /ipmi|spvm|pvm_hmc/);
+    push @params, "console=tty" unless (get_var('BACKEND', '') =~ /ipmi|spvm|pvm_hmc/ || check_var('PUBLIC_CLOUD', 1));
```
asmorodskyi, line 356: from what I understand just this change without changing FS suppose to fix the problem . or am I missing something ?

grisu48: It should, but the main underlying issue remains. I don't see any downsides of switching to the default filesystem for SLES, which should solve the issue where it actually occurs.

asmorodskyi:
all in all changing FS type just to mute some message coming from one app it is overkill there are plenty of options to handle it in more gentle manner. BTRFS is well-known problem generator generally and in openQA specifically . Just ask tools team about rebalancing service and how much headache it brought . So I want to hear some really , really good reasons for switching to BTRFS . And "there is one app which pollute dmesg with messages " is not good one

lib/bootloader_setup.pm
pdostal:
```suggestion
    push @params, "console=tty" unless (get_var('BACKEND', '') =~ /ipmi|spvm|pvm_hmc/ || is_publiccloud);
```

pdostal (approved):
Awesome pull request from awesome Felix! :duck:
Congrats man, I was watching your fight in the progress ticket 💪

grisu48:
> all in all changing FS type just to mute some message coming from one app it is overkill there are plenty of options to handle it in more gentle manner. BTRFS is well-known problem generator generally and in openQA specifically . Just ask tools team about rebalancing service and how much headache it brought . So I want to hear some really , really good reasons for switching to BTRFS . And "there is one app which pollute dmesg with messages " is not good one

btrfs is the default filesystem for SLES. `mru-install-minimal-with-addons` uses it and is the basis for countless openQA test runs that run on a daily basis. I really don't see where those concerns are coming from ...

Let's avoid further discussions that don't bring us anywhere. It is clear that we will not be able to switch to btrfs, so I will try to solve the issue by using `ext4`. Marking PR as draft.

asmorodskyi (approved):
LGTM

asmorodskyi (approved):
LGTM

pdostal (approved):

tests/publiccloud/prepare_tools.pm
```diff
@@ -114,11 +114,12 @@ sub run {
     my $terraform_version = '1.1.7';
     # Terraform in a container
     my $terraform_wrapper = <<EOT;
-#!/bin/sh
-podman run -v /root/:/root/ --rm --env-host=true -w=\$PWD docker.io/hashicorp/terraform:$terraform_version \$@
+#!/bin/bash -e
+podman run --rm -w=\$PWD -v /root/:/root/ --env-host=true docker.io/hashicorp/terraform:$terraform_version \$@
```
jlausuch, line 118: have you only changed the order of the params here?

jlausuch (approved):

tests/publiccloud/prepare_tools.pm
pdostal: Felix switched from `sh` to `bash` and used the `-e` parameter.

PR #15515, opened 2022-09-13
Fix development tools repository uri

ge0r:
The only provider of jing package, which is required for autoyast profile validation, is the Development Tools module. In staging this module's repository is not as previously defined (ending with -CURRENT-Media1) so we need to use the downloads.suse.de GA repo of the module instead, to provide the jing package.

- Related ticket: No ticket, fails in [Staging](https://openqa.suse.de/tests/9497734#step/yast2_clone_system/89)
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/9502564#


b10n1k (approved):
+1

PR #15526, opened 2022-09-14
Move xmlresult specific subs outside of the slurm test module

b10n1k:
Doing so, and move them to the lib we can reuse the subroutines in other modules as well. This is the first step of restructuring the slurm module. For now the subroutines do not consider complicated scheduling. That means that it hasnt tested in a scenario where two modules want to extract test results into an xml file. But with current changes I drive the functionality in this direction. For instance the *pars_results* can define a different xmlfile to store test results per module.

- Verification run: http://aquarius.suse.cz/tests/12499


schlad (approved):

schlad:
I approve it now despite this being a draft! 
This is really nice direction, so we shall carry on! thx @b10n1k !

lib/utils.pm
```diff
@@ -2349,4 +2351,89 @@ sub package_upgrade_check {
...
+        result => _validate_result($result)
+    );
+    return %results;
+}
+
+=head2 pars_results
+    pars_results();
+
+Takes C<test> as an argument. C<test> is an array of hashes which contain the
+test results. They usually those are generated by C<generate_results>. Those
```
Avinesh, line 2400: nit: ~those~

lib/utils.pm
```diff
@@ -2349,4 +2351,89 @@ sub package_upgrade_check {
...
+        }
+    }
+
+    if ($fail_check > 0) {
+        script_run(qq{echo "<testsuite name='$testsuite' errors='1'>" >> $xmlfile});
+    } else {
+        script_run(qq{echo "<testsuite name='$testsuite'>" >> $xmlfile});
+    }
+
+    # pars all results and provide expected xml file
```
Avinesh, line 2424: nit: parse

I know it's nothing, but may be we should use `parse` everywhere in here.

Avinesh (approved):
@b10n1k , this looks good, except couple of cosmetic changes I mentioned. But feel free to merge as it is also.

czerw (approved):

lib/utils.pm
b10n1k: *ready for review* edited only this one. But i think you are right. But let me do it in a separate PR as this would have to touch likely other parts

lib/utils.pm
b10n1k: thanks. *ready for review*

PR #15529, opened 2022-09-14
qesapdeployment lib function for build.sh configuration

mpagot:
More functions to generate all the configuration files needed as running the qe-sap-deployment using its `build.sh`

- Verification run: http://openqaworker15.qa.suse.cz/tests/32374

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -44,7 +44,6 @@ my @log_files = ();
 use constant QESAPDEPLOY_PREFIX => 'qesapdep';
 
 our @EXPORT = qw(
-  qesap_create_folder_tree
   qesap_pip_install
   qesap_upload_logs
   qesap_get_deployment_code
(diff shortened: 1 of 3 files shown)


alvarocarvajald (commented):
Code LGTM.

I don't like that this is introducing an alternate approach to `qesap.py` as defined in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/qesapdeployment.pm#L175-L187 and in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/qesapdeployment.pm#L198-L218, but I guess the reason for this is that `qesap.py` is probably not handling yet these scenarios. My suggestion would be to first expand `qesap.py` so it covers these scenarios, and then extend `lib/qesapdeployment.pm` with whatever changes are required by `qesap.py` instead of this intermediate step.

Having said that however, I will not be opposed to merging this if there is a pressing issue that this intends to solve.

PR #15531, opened 2022-09-14
Remove deprecated probing in iscsi_client.

JRivrain:
In iscsi_client.pm we are testing if iscsiuio.socket is activated. It used to be activated by default, but that was un actual bug: this socket is only useful if a Qlogic HBA is connected to the system but if it isn't,
the activation of the socket delays the system startup for no reason. So it was decided to no longer activate that socket unless such hardware is detected.
See details here https://bugzilla.suse.com/show_bug.cgi?id=1203062 and here https://github.com/yast/yast-iscsi-client/pull/116. It does not look like we can easily emulate such hardware in Qemu. In older products, the socket will likely still be activated but not in 15.5 and later, so it looks like a reasonable option to just stop probing that socket.

VR https://openqa.suse.de/tests/9519817

### DIFF
--- a/tests/iscsi/iscsi_client.pm
+++ b/tests/iscsi/iscsi_client.pm
@@ -114,7 +114,6 @@ sub run {
     record_info 'Systemd', 'Verify status of iscsi services and sockets';
     systemctl("is-active iscsid.service");
     systemctl("is-active iscsid.socket");
-    systemctl("is-active iscsiuio.socket");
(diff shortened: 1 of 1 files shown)


ge0r (approved):
I agree with this!
Btw, I know I might sound a bit paranoid, but could we have a VR run of an iscsi job pair with the change?

JRivrain:
> I agree with this! Btw, I know I might sound a bit paranoid, but could we have a VR run of an iscsi job pair with the change?

VR added

PR #15532, opened 2022-09-15
Bci xfail soft 1

m-dati:
PR to set in opeqa as soft-failures the XFAIL tests resulting from pytest of containers BCI-tests in python

- Related ticket: https://progress.opensuse.org/issues/113510
- Verification run: https://openqa.suse.de/tests/9520968

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -29,9 +29,13 @@ sub run_tox_cmd {
     my $bci_timeout = get_var('BCI_TIMEOUT', 1200);
     my $bci_reruns = get_var('BCI_RERUNS', 3);
     my $bci_reruns_delay = get_var('BCI_RERUNS_DELAY', 10);
-    my $cmd = "tox -e $env -- -rx -n auto";
+    my $tox_out = "tox_output.txt";
+    my $cmd = "tox -e $env -- -rxX -n auto";
     $cmd .= " -k \"$bci_marker\"" if $bci_marker;
     $cmd .= " --reruns $bci_reruns --reruns-delay $bci_reruns_delay";
+    $cmd .= "| tee $tox_out";
+    # Cut the tox log from the header onward and filter the text
+    my $cmd_xf = "awk '/short test summary info/{f=1}f' $tox_out | grep XFAIL";
     record_info("tox", "Running command: $cmd");
     my $ret = script_run("timeout $bci_timeout $cmd", timeout => ($bci_timeout + 3));
     if ($ret == 124) {
@@ -39,10 +43,14 @@ sub run_tox_cmd {
         record_soft_failure("The command <tox -e $env> timed out.");
         $error_count += 1;
     } elsif ($ret != 0) {
-        record_soft_failure("The command <tox -e $env> failed.");
+        record_soft_failure("The command <tox -e $env> failed $ret.");
         $error_count += 1;
     } else {
         record_info('PASSED');
+    }
+    my $ret_xf = script_run("$cmd_xf", timeout => ($bci_timeout + 3));
+    if ($ret_xf == 0) {
+        record_soft_failure("The command <tox -e $env> has XFAIL.");
     }
     # Rename resulting junit file because it will be overwritten if we run
     # the same tox command later with another container engine. This way,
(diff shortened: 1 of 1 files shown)


m-dati:
This update contains in the bci_test.pm containers module some changes in order to mark as openqa soft-failures the error of python code marked ad XFAIL in pytest env.

The BCI tests are written in python and tested with 'pytest' using 'tox'. The openqa container tests start with the Perl module bci_test.pm that run in a bash session the tox command composed with proper parameters based on the test settings.

The pytest exit codes are 5 specific values but where xfail occurrence is not visible, that is i.e., if only xfail occurr in the test execution, we have no evidence of it and zero is returned as all pass.

Therefore the simplest way to detect xfail in the openqa bci_test.pm module was,after tox ran with the pytest '-rxX' option added, to save the output in a log and execute a pytest logs inspection searching for XFAIL string and manage the result in the Perl module, in order to assign a Soft Failure.

So, to verify the validity of this method, I also cloned some tests using a temporary modifyed branch of the BCI test with additional XFAIL tests in test_all.py, that have to appear in the log and proof that behaviour.

tests/containers/bci_test.pm
```diff
@@ -29,21 +29,29 @@ sub run_tox_cmd {
...
     my $bci_reruns = get_var('BCI_RERUNS', 3);
     my $bci_reruns_delay = get_var('BCI_RERUNS_DELAY', 10);
-    my $cmd = "tox -e $env -- -n auto";
+    my $tox_out = "tox_output.txt";
+    my $cmd = "tox -e $env -- -rxX -n auto";
     $cmd .= " -k \"$bci_marker\"" if $bci_marker;
     $cmd .= " --reruns $bci_reruns --reruns-delay $bci_reruns_delay";
+    $cmd .= "| tee $tox_out";
+    # Cut the tox log from the header onward and filter the text
+    my $cmd_xf = "awk '/short test summary info/{f=1}f' $tox_out | grep XFAIL";
```
grisu48, line 38: Please move this variable declaration down to where it is needed. It is weird to have the variable being declared somewhere and then being used somewhere else.

tests/containers/bci_test.pm
```diff
@@ -29,21 +29,29 @@ sub run_tox_cmd {
...
     } elsif ($ret != 0) {
-        record_soft_failure("The command <tox -e $env> failed.");
+        record_soft_failure("The command <tox -e $env> failed $ret.");
         $error_count += 1;
     } else {
         record_info('PASSED');
     }
+    my $ret_xf = script_run("$cmd_xf", timeout => ($bci_timeout + 3));
+    if ($ret_xf == 0) {
+        record_soft_failure("The command <tox -e $env> has XFAIL.");
```
grisu48, line 53:
```suggestion
        record_soft_failure("The command <tox -e $env> has softfailures") if ($ret_xf == 0);
```

This would also get rid of the whole if block. Just as a suggestion, feel free to ignore if you don't like it.

grisu48 (commented):
Two comments, but it looks in general good to me.

tests/containers/bci_test.pm
m-dati: thanks, indeed moving the if at end is better.

grisu48 (approved):

tests/containers/bci_test.pm
```diff
@@ -29,21 +29,27 @@ sub run_tox_cmd {
...
     record_info("tox", "Running command: $cmd");
     my $ret = script_run("timeout $bci_timeout $cmd", timeout => ($bci_timeout + 3));
     if ($ret == 124) {
         # man timeout: If  the command times out, and --preserve-status is not set, then exit with status 124.
         record_soft_failure("The command <tox -e $env> timed out.");
         $error_count += 1;
     } elsif ($ret != 0) {
-        record_soft_failure("The command <tox -e $env> failed.");
+        record_soft_failure("The command <tox -e $env> failed $ret.");
         $error_count += 1;
```
m-dati, line 45: A kind question to reviewers: actually all the BCI errors are marked as` Soft Failure` also the effective failures and timeouts: should we remove this status for such cases, instead ?

grisu48: Without an example I cannot make a real judgement here. How would it look?

m-dati: I propose to skip this last question and move it to a possible next discussion, keeping here the only poo requirement as target of this PR.

PR #15533, opened 2022-09-15
WIP: Fix capitalization issue in addon check

BillAnastasiadis:
While validating registered addons after migration, the test fails on `IBM-Power-Tools` and `IBM-Power-Advance-Toolchain`.

`IBM-Power-Tools` was failing due to capitalization, and passes when performing a case-insensitive string-matching. But `IBM-Power-Advance-Toolchain` is actually not registered, due to https://bugzilla.suse.com/show_bug.cgi?id=1202419. The proposed action is to softfail in case of IBM modules, since there are problems with the `-release` packages for them.

- Related ticket: https://progress.opensuse.org/issues/115697
- Verification run: https://openqa.suse.de/tests/9511344

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -44,6 +44,7 @@ our @EXPORT = qw(
   %ADDONS_REGCODE
   @SLE15_ADDONS_WITHOUT_LICENSE
   @SLE12_MODULES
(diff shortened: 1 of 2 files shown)


jknphy:
I lost track of this, we will rethink it in the future.

PR #15534, opened 2022-09-15
openscap: multiple fixes

paolostivanin:
poo: https://progress.opensuse.org/issues/116368

* as suggested on https://bugzilla.suse.com/show_bug.cgi?id=1203408#c8 , use two tags instead of a sed expression
* on opensuse, manually add the 15.5 data while we wait for an update

Verification runs:
- x86_64: https://openqa.suse.de/tests/9523918
- aarch64: https://openqa.suse.de/tests/9524048
- leap aarch64: https://openqa.opensuse.org/tests/2675105
- leap x86_64: https://openqa.opensuse.org/tests/2675220

### DIFF
--- a/tests/security/openscap/oscap_setup.pm
+++ b/tests/security/openscap/oscap_setup.pm
@@ -19,11 +19,6 @@ sub run {
 
     oscap_get_test_file("oval.xml");
     oscap_get_test_file("xccdf.xml");
-
-    # xccdf.xml is for SLE, the CPE is different on openSUSE
-    if (is_opensuse) {
-        assert_script_run "sed -i 's#cpe:/o:suse#cpe:/o:opensuse#g' xccdf.xml";
-    }
 }
 
 sub test_flags {
--- a/data/openscap/xccdf.xml
+++ b/data/openscap/xccdf.xml
@@ -3,6 +3,7 @@
   <status date="2017-10-09">draft</status>
   <title>Hardening SUSE Linux Enterprise</title>
   <platform idref="cpe:/o:suse"/>
+  <platform idref="cpe:/o:opensuse"/>
   <version>1</version>
 
   <Profile id="standard">
--- a/tests/security/openscap/oscap_remediating_offline.pm
+++ b/tests/security/openscap/oscap_remediating_offline.pm
@@ -38,7 +38,7 @@ sub run {
     ensure_generated_file($xccdf_result);
     prepare_remediate_validation;
 
-    validate_script_output "oscap xccdf remediate --results $remediate_result $xccdf_result", sub { $remediate_match };
+    validate_script_output_retry("oscap xccdf remediate --results $remediate_result $xccdf_result", sub { $remediate_match }, retry => 5, delay => 10);
     validate_result($remediate_result, $remediate_result_match);
 
     # Verify the remediate action result


data/openscap/oscap-dict.patch
```diff
@@ -0,0 +1,14 @@
+--- /usr/share/openscap/cpe/openscap-cpe-dict.xml	2022-05-07 21:11:24.000000000 -0400
++++ /usr/share/openscap/cpe/openscap-cpe-dict.xml	2022-09-15 07:30:17.067140391 -0400
+@@ -65,6 +65,10 @@
+             <title xml:lang="en-us">openSUSE Leap 15.3</title>
+             <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.opensuse:def:153</check>
+       </cpe-item>
++      <cpe-item name="cpe:/o:opensuse:leap:15.5">
++            <title xml:lang="en-us">openSUSE Leap 15.5</title>
++            <check system="http://oval.mitre.org/XMLSchema/oval-definitions-5" href="openscap-cpe-oval.xml">oval:org.open-scap.cpe.opensuse:def:155</check>
++      </cpe-item>
```
punkioudi, line 10: Hm, if I understand correctly, the lines that you add manually for leap 15.5 should be fixed in a future update, right? 
If yes, I disagree on doing this workaround. We are in product testing and these failures, because of a bug should be visible :)

data/openscap/oscap-oval.patch
```diff
@@ -0,0 +1,45 @@
...
+             <rpminfo_state id="oval:org.open-scap.cpe.opensuse:ste:153" version="1" xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux">
+                   <version operation="pattern match">^15.3$</version>
+             </rpminfo_state>
++            <rpminfo_state id="oval:org.open-scap.cpe.opensuse:ste:155" version="1" xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux">
++                  <version operation="pattern match">^15.5$</version>
++            </rpminfo_state>
+             <rpminfo_state id="oval:org.open-scap.cpe.opensuse:ste:9999" version="1" xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#linux">
+               <!-- matching for timestamp -->
+               <version operation="pattern match">^\d{8}$</version>
+
```
punkioudi, line 45: Same here, I suppose

tests/security/openscap/oscap_setup.pm
```diff
@@ -15,14 +15,15 @@ use version_utils 'is_opensuse';
...
     oscap_get_test_file("oval.xml");
     oscap_get_test_file("xccdf.xml");
-
-    # xccdf.xml is for SLE, the CPE is different on openSUSE
     if (is_opensuse) {
-        assert_script_run "sed -i 's#cpe:/o:suse#cpe:/o:opensuse#g' xccdf.xml";
+        oscap_get_test_file("oscap-dict.patch");
+        oscap_get_test_file("oscap-oval.patch");
+        assert_script_run("patch -u -b /usr/share/openscap/cpe/openscap-cpe-dict.xml -i oscap-dict.patch");
+        assert_script_run("patch -u -b /usr/share/openscap/cpe/openscap-cpe-oval.xml -i oscap-oval.patch");
```
punkioudi, line 26: Same here, let it fail for leap 15.5 until it is fixed

jouyingbin (approved):
LGTM now if everything VR is updated. Thanks. :)

PR #15535, opened 2022-09-15
Remove mariadb_srv from schedule

Zaoliang:
since mariadb is not available in container repo, so we cannot schedue it on ALP for now.


Zaoliang:
http://10.162.30.85/tests/5020#step/mariadb_srv/32

PR #15545, opened 2022-09-16
Remove cockpit_service from schedule

Zaoliang:
because cockpit_service failed in new build:
http://10.162.30.85/tests/5020#step/cockpit_service/ For database testsuite better not to include failing tests


Zaoliang:
http://10.162.30.85/tests/5024:
yaml schedule itself looks good.

PR #15559, opened 2022-09-20
JeOS: Handle WiFi interface in JeOS firstboot, if shown

ggardet:
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/commit/95f12905385467ec9883b95482443339e421ae69
- Verification run: https://openqa.opensuse.org/tests/2703648#step/firstrun/12


asdil12:
PR looks fine but where is that needle used? It doesn't appear in the VR as the firstboot test module answers "no" to the question if wifi should be configured.
Are you planning to use the needle later with `tests/jeos/rpi_wifi.pm`?

asdil12 (approved):
LGTM

PR #15560, opened 2022-09-20
ALP: fix cockpit modules to be installed

jlausuch:
- `cockpit-wicked` is not available in ALP.
- `cockpit-bridge` and `cockpit-system` can be installed in ALP .

https://build.opensuse.org/package/view_file/SUSE:ALP/000product/ALP-ftp-POOL-aarch64.kiwi?expand=1

VR: https://openqa.opensuse.org/tests/2703946#

### DIFF
--- a/tests/microos/cockpit_service.pm
+++ b/tests/microos/cockpit_service.pm
@@ -13,7 +13,7 @@ use testapi;
 use transactional;
 use utils qw(systemctl);
 use mm_network qw(is_networkmanager);
-use version_utils qw(is_microos is_sle_micro is_leap_micro);
+use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp);
 
 sub run {
     my ($self) = @_;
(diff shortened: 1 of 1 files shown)


mloviska (approved):

jknphy:
The right way in ALP seems to be installing pattern `alp_cockpit` via transactional script, which should contain everything available at the moment.

foursixnine:
> The right way in ALP seems to be installing pattern `alp_cockpit` via transactional script, which should contain everything available at the moment.

@jlausuch I'm with Joaquin here... it should be the pattern, wdyt?

jlausuch:
> The right way in ALP seems to be installing pattern `alp_cockpit` via transactional script, which should contain everything available at the moment.

Not only for ALP, but for all distris, all of them have that cockpit pattern, but we wanted to make sure those packages are present and installable according to the kiwi file.

jlausuch:
Make we can install the pattern, and after reboot, check that all expected packages are installed... WDYT?

jknphy:
> Maybe we can install the pattern, and after reboot, check that all expected packages are installed... WDYT?

yes, it would be more transparent that way, even when checking the patterns and die to say that doesn't match expected kiwi file or something like that.

jlausuch:
@foursixnine @jknphy I tried in SLE Micro, MicroOS and ALP. After installing the pattern `patterns-microos-cockpit` or `patterns-alp-cockpit` (for ALP), not ALL cockpit modules get installed... Actually, ALP comes with `alp_cockpit` pattern installed, but only few cockpit modules installed... so I guess we can't follow that approach anywhere...

jlausuch:
I went through all products/versions and I came up with this new proposal. Installing the patterns do not install ALL packages in all the cases... so I think we should try to install all available packages for each product/version.

ggardet:
> I went through all products/versions and I came up with this new proposal. Installing the patterns do not install ALL packages in all the cases... so I think we should try to install all available packages for each product/version.

Maybe we should just fix the patterns? I guess the user expects to have all the requirements installed when he installs the pattern.

foursixnine:
>  Maybe we should just fix the patterns?

I was thinking the same, this looks more like a bug than a by-design thing

mloviska:
IMO, installing all individual packages that come with each `sle-micro/leap-micro` version creates a test maintenance burden. Also, I believe that the package list of `leap-micro` should correspond with `sle-micro`

jlausuch:
> IMO, installing all individual packages that come with each `sle-micro/leap-micro` version creates a test maintenance burden. 

That's true, but isn't it better that we find the bug first of "something can't be installed" than users find it out?
In this example, I just found a bug that cockpit-pcp can't be installed on transactional systems:
https://openqa.opensuse.org/tests/2711293#step/journal_check/22

jlausuch:
> > Maybe we should just fix the patterns?
> 
> I was thinking the same, this looks more like a bug than a by-design thing

Ok, I'll then close this PR as it has diverged from the original purpose. I'll create bug respective bug reports and when they are fixed we can modify the test to install the patterns and check the packages.

PR #15562, opened 2022-09-20
Implement new module to test apptainer

b10n1k:
Apptainer is missing testing in both SLE and OpenSUSE but the package is available only in OpenSUSE. The module tests some basic functionalities of the tool to pull, build and run containers.

One of the things left out is the `capability` command which seems to not work properly at the moment.

- Verification run: 
- https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2315562
- https://openqa.opensuse.org/tests/overview?distri=microos&version=Tumbleweed&build=b10n1k%2Fos-autoinst-distri-opensuse%2315562

### DIFF
--- a/data/containers/apptainer_container.def
+++ b/data/containers/apptainer_container.def
@@ -1,0 +1,17 @@
+BootStrap: zypper
+MirrorURL: http://download.opensuse.org/distribution/openSUSE-stable/repo/oss/
+Include: zypper
+
+%post
+    zypper ref
+    zypper -n in cowsay
+
+%environment
+    export LC_ALL=C
+    export PATH=/usr/games:$PATH
+
+%runscript
+    date | cowsay "I love Containers!!"
+
+%labels
+    Author iob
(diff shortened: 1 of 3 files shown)


data/containers/apptainer_container.def
```diff
@@ -0,0 +1,17 @@
...
+MirrorURL: http://download.opensuse.org/distribution/openSUSE-stable/repo/oss/
+Include: zypper
+
+%post
+    zypper ref
+    zypper -n in cowsay
+
+%environment
+    export LC_ALL=C
+        export PATH=/usr/games:$PATH
```
mimi1vx, line 11: why is here indented ?

b10n1k: no reason. fixed

czerw:
Why is  JeOS-for-kvm-and-xen included in verification runs? I don't see  apptainer scheduled there.

b10n1k:
> Why is JeOS-for-kvm-and-xen included in verification runs? I don't see apptainer scheduled there.

the apptainer has been scheduled only against TW. I run other VERSIONS because of the container's scheduler, just to make sure that nothing unsurprised would happen. Meaning, that not seeing `apptainer` in `JeOS-for-kvm-and-xen` is the right/expected thing

czerw (approved):

PR #15564, opened 2022-09-20
Add support for transactional systems to LTP

czerw:
Adaptible Linux Platform and SLE Micro use transactional approach for
system management. LTP installation and configuration had to be adapted.

The goal is to cover installation, syscalls and cve for ALP September prototype. Network setup is disabled, as we don't need at the moment. It will be explored/covered for ALP December release. 

- Related ticket: none
- Needles: none
- Verification runs

#####  ALP
- install_ltp: http://10.100.12.105/tests/1822
- run_ltp: http://10.100.12.105/tests/1803
- install_ltp with command file : http://10.100.12.105/tests/1828#

##### SLE Micro
- install_ltp: http://10.100.12.105/tests/1819
- run_ltp http://10.100.12.105/tests/1818

##### SLES
- https://openqa.suse.de/tests/9556110

`PUBLISH_HDD_1` variable was added to experimental runs of `slem_installation_default` and `alp_default`.

### DIFF
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2016-2020 SUSE LLC
+# Copyright 2016-2022 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 #
 # Summary: This module installs the LTP (Linux Test Project) and then reboots.
@@ -19,7 +19,7 @@ use bootloader_setup qw(add_custom_grub_entries add_grub_cmdline_settings);
 use power_action_utils 'power_action';
 use repo_tools 'add_qa_head_repo';
 use upload_system_log;
-use version_utils qw(is_jeos is_opensuse is_released is_sle is_leap is_tumbleweed is_rt);
+use version_utils qw(is_jeos is_opensuse is_released is_sle is_leap is_tumbleweed is_rt is_transactional is_alp);
 use Utils::Architectures;
 use Utils::Systemd qw(systemctl disable_and_stop_service);
 use LTP::utils;
@@ -245,7 +245,7 @@ sub add_ltp_repo {
     my $repo = get_var('LTP_REPOSITORY');
 
     if (!$repo) {
-        if (is_sle) {
+        if (is_sle || is_transactional) {
             add_qa_head_repo;
             return;
         }
@@ -286,11 +286,16 @@ sub get_default_pkg {
 sub install_from_repo {
     my @pkgs = split(/\s* \s*/, get_var('LTP_PKG', get_default_pkg));
 
-    zypper_call("in --recommends " . join(' ', @pkgs));
+    if (is_transactional) {
+        assert_script_run("transactional-update -n -c pkg install " . join(' ', @pkgs), 90);
+    } else {
+        zypper_call("in --recommends " . join(' ', @pkgs));
+    }
 
+    my $run_cmd = is_transactional ? 'transactional-update -c -d --quiet run' : '';
     for my $pkg (@pkgs) {
         my $want_32bit = $pkg =~ m/32bit/;
-        record_info("LTP pkg: $pkg", script_output("rpm -qi $pkg | tee "
+        record_info("LTP pkg: $pkg", script_output("$run_cmd rpm -qi $pkg | tee "
                   . get_ltp_version_file($want_32bit)));
         assert_script_run "find " . get_ltproot($want_32bit) .
           q(/testcases/bin/openposix/conformance/interfaces/ -name '*.run-test' > )
@@ -400,7 +405,11 @@ sub run {
 
     log_versions 1;
 
-    zypper_call('in efivar') if is_sle('12+') || is_opensuse;
+    if (is_alp) {
+        assert_script_run("transactional-update -n -c pkg install efivar", 90);
+    } else {
+        zypper_call('in efivar') if is_sle('12+') || is_opensuse;
+    }
 
     $grub_param .= ' console=hvc0' if (get_var('ARCH') eq 'ppc64le');
     $grub_param .= ' console=ttysclp0' if (get_var('ARCH') eq 's390x');
@@ -408,8 +417,8 @@ sub run {
         add_grub_cmdline_settings($grub_param, update_grub => 1);
     }
 
-    add_custom_grub_entries if (is_sle('12+') || is_opensuse) && !is_jeos;
-    setup_network;
+    add_custom_grub_entries if (is_sle('12+') || is_opensuse || is_transactional) && !is_jeos;
+    setup_network unless is_transactional;
 
     # we don't run LVM tests in 32bit, thus not generating the runtest file
     # for 32 bit packages
@@ -422,7 +431,7 @@ sub run {
 
     # boot_ltp will schedule the tests and shutdown_ltp if there is a command
     # file
-    if (get_var('LTP_INSTALL_REBOOT')) {
+    if (get_var('LTP_INSTALL_REBOOT') || (is_transactional && $cmd_file)) {
         power_action('reboot', textmode => 1) unless is_jeos;
         loadtest_kernel 'boot_ltp';
     } elsif ($cmd_file) {
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -15,7 +15,7 @@ use Time::HiRes 'sleep';
 use testapi;
 use Utils::Architectures;
 use utils;
-use version_utils qw(is_microos is_sle_micro is_jeos is_leap is_sle is_tumbleweed is_selfinstall);
+use version_utils qw(is_microos is_sle_micro is_jeos is_leap is_sle is_tumbleweed is_selfinstall is_alp is_transactional);
 use mm_network;
 use Utils::Backends;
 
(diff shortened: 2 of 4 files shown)


mloviska:
Looks very nice, good job!

czerw:
@mdoucha @pevik please check too

tests/kernel/install_ltp.pm
```diff
@@ -400,16 +405,20 @@ sub run {
...
     $grub_param .= ' console=hvc0' if (get_var('ARCH') eq 'ppc64le');
     $grub_param .= ' console=ttysclp0' if (get_var('ARCH') eq 's390x');
     if (!is_sle('<12') && defined $grub_param) {
         add_grub_cmdline_settings($grub_param, update_grub => 1);
     }
 
-    add_custom_grub_entries if (is_sle('12+') || is_opensuse) && !is_jeos;
-    setup_network;
+    add_custom_grub_entries if (is_sle('12+') || is_opensuse || is_transactional) && !is_jeos;
+    setup_network unless is_transactional;
```
pevik, line 421: LGTM. The only question is why not run network setup for transactional. Would it be too complicated or testing not relevant? Anyway, we need to remember not run network tests for transactional.

pevik (approved):

foursixnine (commented):
I guess after all the transactional updates changes... there's the reboot happening again when `LTP_INSTALL_REBOOT` is set right?... if there's the chance that system will not be rebooted, then an explicit reboot will be needed too...

tests/kernel/install_ltp.pm
czerw: I should mentioned it into PR, it is more complicated at the moment for September prototype which will just cover cve and syscalls. There will be follow up to extend testing.

czerw:
> I guess after all the transactional updates changes... there's the reboot happening again when `LTP_INSTALL_REBOOT` is set right?... if there's the chance that system will not be rebooted, then an explicit reboot will be needed too...

We don't do any reboot related to transactional changes now. All added transactional changes (installation, modifications) continue with previous transaction and system is just shutdown at the end.

tests/kernel/install_ltp.pm
```diff
@@ -286,11 +286,16 @@ sub get_default_pkg {
...
+        assert_script_run("transactional-update -n -c pkg install" . join(' ', @pkgs), 90);
+    } else {
+        zypper_call("in --recommends " . join(' ', @pkgs));
+    }
 
+    my $run_cmd = is_transactional ? 'transactional-update -c -d --quiet run' : '';
     for my $pkg (@pkgs) {
         my $want_32bit = $pkg =~ m/32bit/;
-        record_info("LTP pkg: $pkg", script_output("rpm -qi $pkg | tee "
+        record_info("LTP pkg: $pkg", script_output("$run_cmd rpm -qi $pkg | tee "
```
mdoucha, line 298: This `rpm` command should be a read-only query. Does it need to be wrapped in `transactional-update`?

mdoucha:
> We don't do any reboot related to transactional changes now. All added transactional changes (installation, modifications) continue with previous transaction and system is just shutdown at the end.

@foursixnine has a point though. The reboot condition at the end should be changed to:

    if (get_var('LTP_INSTALL_REBOOT') || (is_transactional && $cmd_file)) {

czerw:
> @foursixnine has a point though. The reboot condition at the end should be changed to:
> 
> ```
> if (get_var('LTP_INSTALL_REBOOT') || (is_transactional && $cmd_file)) {
> ```

OK, i didn't think about as we don't use it at all (at least what i'm aware)

czerw:
Code updated with additional reboot in case of command file is supplied. It can schedule installation and test in one job now see  http://10.100.12.105/tests/1828#

tests/kernel/install_ltp.pm
```diff
@@ -286,11 +286,16 @@ sub get_default_pkg {
 sub install_from_repo {
     my @pkgs = split(/\s* \s*/, get_var('LTP_PKG', get_default_pkg));
 
-    zypper_call("in --recommends " . join(' ', @pkgs));
+    if (is_transactional) {
+        assert_script_run("transactional-update -n -c pkg install " . join(' ', @pkgs), 90);
+    } else {
+        zypper_call("in --recommends " . join(' ', @pkgs));
```
mdoucha, line 292: It'd be nice to add a helper function to the test lib that'll handle this if/else internally so that you can call just something like `install_pkg(@pkgs);` in the test.

czerw: We had it in liberty, but this should have been done already, not sure why it wasn't done already. @jlausuch is there any plan for it?

jlausuch: Not that I know... Maybe it would be nice to have it, but I would leave it for another PR.

tests/kernel/install_ltp.pm
```diff
@@ -286,11 +286,16 @@ sub get_default_pkg {
 sub install_from_repo {
     my @pkgs = split(/\s* \s*/, get_var('LTP_PKG', get_default_pkg));
 
-    zypper_call("in --recommends " . join(' ', @pkgs));
+    if (is_transactional) {
+        assert_script_run("transactional-update -n -c pkg install " . join(' ', @pkgs), 90);
```
jlausuch, line 290: Can't it just be `"transactional-update -n -c pkg install @pkgs"` ?

czerw: I think the main concern is one universal function to cover installation for both types and don't use if-else logic in the test code on different places.

jlausuch:
This looks very promising! Looking forward to see LTP tests running on transactional products family :) 

Maybe, a hint about how transactional-update commands are used. There is this library 
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/transactional.pm

where you can use `trup_call` and wrap the command as argument. Also, for reboots, we normally use `check_reboot_changes` function.

For example: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/transactional/transactional_update.pm#L111
```
    trup_call "--continue pkg install" . rpmver('optional');
    check_reboot_changes;`
```

jlausuch:
sorry, pressed close button instead of comment...

czerw:
> but even the command file is not supplied, you still use transactional-update to install some stuff, right? that will always need a reboot.

No, we don't. We just install and shutdown.

czerw:
> where you can use `trup_call` and wrap the command as argument. Also, for reboots, we normally use `check_reboot_changes` function.

Thanks for `trup_call` hint, will use it.  But library doesn't contain any function which would transparently install packages independently on system type.

tests/kernel/install_ltp.pm
czerw: We could also have wrapper for `transactional-update -c -d --quiet run` inside `script_run`, but as you say, should be in different PR.

jlausuch:
> > but even the command file is not supplied, you still use transactional-update to install some stuff, right? that will always need a reboot.
> 
> When a command file is not supplied, the VM will shut down and export disk image for other jobs that'll run tests after fresh boot.

Ah ok,sure, if no more actions are needed after pkg installation, no need for reboot. It could be still done, to make sure the system boots with all those changes and not discover it in the children jobs. But I leave that up to you.

czerw:
> > where you can use `trup_call` and wrap the command as argument. Also, for reboots, we normally use `check_reboot_changes` function.
> 
> Thanks for `trup_call` hint, will use it. But library doesn't contain any function which would transparently install packages independently on system type.

@jlausuch  I will definitely not use `trup_call`, it needs `root-console`. It will fail  http://10.100.12.105/tests/1835#step/install_ltp/52, if  it is used with `$self->select_serial_terminal`. I checked all existing code and all `trup_call` are in `root-console` or is even set just before command.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1217a95d463f5e8792b04e0e7b2dd9f2c2486cb5/tests/console/java.pm#L55-L56

jlausuch:
> > > where you can use `trup_call` and wrap the command as argument. Also, for reboots, we normally use `check_reboot_changes` function.
> > 
> > 
> > Thanks for `trup_call` hint, will use it. But library doesn't contain any function which would transparently install packages independently on system type.
> 
> @jlausuch I will definitely not use `trup_call`, it needs `root-console`. It will fail http://10.100.12.105/tests/1835#step/install_ltp/52, if it is used with `$self->select_serial_terminal`. I checked all existing code and all `trup_call` are in `root-console` or is even set just before command.
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1217a95d463f5e8792b04e0e7b2dd9f2c2486cb5/tests/console/java.pm#L55-L56

ok, right, yes it uses root-console... I didn't think about this limitation of using it here.. sorry.

mloviska (approved):

jlausuch (approved):

czerw:
> ok, right, yes it uses root-console... I didn't think about this limitation of using it here.. sorry.

No problem, i will create some progress tickets to discuss it further

czerw:
@jlausuch @foursixnine @mdoucha i created https://progress.opensuse.org/issues/117028 based on findings during review.

foursixnine:
> @jlausuch @foursixnine @mdoucha i created https://progress.opensuse.org/issues/117028 based on findings during review.

Thanks! I have a couple of thoughts on how the package management is done... the SLL approach is exactly part of what I was thinking

PR #15565, opened 2022-09-20
Fix xml notation for HPC profiles

b10n1k:
`t` is a new notation which is supported by sle15sp3+ so using it for older versions makes the installation complain.
In addition I have fixed package conflict between those version regarding python

- Related ticket: https://progress.opensuse.org/issues/116866
- Verification run: only 15-SP2 but i think its good to enough
 https://openqa.suse.de/tests/overview?distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2315565&version=15-SP2


czerw:
Please upper case hpc to HPC in commit msg and PR.

b10n1k:
i encounter a new error on aarch64 (gnome) but that might be different thingy. https://openqa.suse.de/tests/9551313#step/installation/63

> Please upper case hpc to HPC in commit msg and PR.

DONE

czerw (approved):

PR #15567, opened 2022-09-21
Fix timeout issue of assert_script_run

RoyCai7:
POO:https://progress.opensuse.org/issues/115451
& symbol sometimes can effect return value of assert_script_run, expecially parallel jobs.
IMO: the best way to overcome it is to remove special character &.

- Related ticket: https://progress.opensuse.org/issues/115451
- Needles: None
- Parallel verification run with sle15sp5 on build 21.1:
      -- http://openqa.suse.de/tests/9562372#step/add_serial_console/9
      -- http://openqa.suse.de/tests/9562373#step/add_serial_console/9

### DIFF
--- a/tests/installation/add_serial_console.pm
+++ b/tests/installation/add_serial_console.pm
@@ -17,10 +17,10 @@ sub run {
     my $device = 'sda';
     # Find installed OS root device, mount it, and chroot into it
     assert_script_run("for part in /dev/${device}[1-4]; do
-            mount \$part /mnt &> /dev/null &&
+            mount \$part /mnt &&
             test -e /mnt/etc/default/grub &&
                 break ||
-                umount /mnt &> /dev/null
+                umount /mnt
         done");
     assert_script_run('mount -o bind /dev /mnt/dev');
     assert_script_run('mount -o bind /proc /mnt/proc');


Julie-CAO:
Since there is no output for the shell script to run, I think it is ok to remove '/dev/null'. but we still need @nanzhg 's input to make it sure. 
@RoyCai7 , Is there a verification run for vmware 7.0?

RoyCai7:
> Since there is no output for the shell script to run, I think it is ok to remove '/dev/null'. but we still need @nanzhg 's input to make it sure. 
> 
> @RoyCai7 , Is there a verification run for vmware 7.0?

Will add vmware test

nanzhg:
> > Since there is no output for the shell script to run, I think it is ok to remove '/dev/null'. but we still need @nanzhg 's input to make it sure. @RoyCai7 , Is there a verification run for vmware 7.0?
> 
> Passed alll hyperv jobs which use module add_serial_console at the same time. http://openqa.suse.de/t9586713 http://openqa.suse.de/t9586712 http://openqa.suse.de/t9586714 http://openqa.suse.de/t9586715 http://openqa.suse.de/t9586716 http://openqa.suse.de/t9586717 Please ignore first_boot failure. Once VMWARE jobs with add_serial_console is ready(still have needle mismatch issue), then test it.

We discussed this issue before, it looks good to me for this change.

alice-suse (approved):

PR #15570, opened 2022-09-21
Test trento-agent installation on cluster nodes

mpagot:
Create a test to install the agents on all the nodes of the SAP Landscape cluster.
Add proper trento and qesap destroy in more post fail hooks.

qesapdeploy lib:
 - new api to get the location of the generated Ansible inventory
 - longer timeout for the python requirements installation
 - setting rename DEPLOYMENT_DIR --> QESAP_DEPLOYMENT_DIR
 - git clone log fix
 - api change in qesap_prepare_env
  
 trento lib:
 - new deploy_vm function for 00.040
 - new install_agent function
 - expose k8s_logs
 - deploy_qesap to save the inventory Unit test for qesapdeploy and trento lib
Improve documentation for the TRENTO settings

- Verification run: http://openqaworker15.qa.suse.cz/tests/35612 http://openqaworker15.qa.suse.cz/tests/38602

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -51,6 +51,7 @@ our @EXPORT = qw(
   qesap_configure_hanamedia
   qesap_sh_deploy
   qesap_sh_destroy
+  qesap_get_inventory
   qesap_prepare_env
   qesap_execute
   qesap_yaml_replace
@@ -70,7 +71,7 @@ our @EXPORT = qw(
 sub qesap_get_file_paths {
     my %paths;
     $paths{qesap_conf_filename} = get_required_var('QESAP_CONFIG_FILE');
-    $paths{deployment_dir} = get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment');
+    $paths{deployment_dir} = get_var('QESAP_DEPLOYMENT_DIR', get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment'));
     $paths{terraform_dir} = get_var('PUBLIC_CLOUD_TERRAFORM_DIR', $paths{deployment_dir} . '/terraform/');
     $paths{qesap_conf_trgt} = $paths{deployment_dir} . "/scripts/qesap/" . $paths{qesap_conf_filename};
     return (%paths);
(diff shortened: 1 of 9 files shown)


lib/qesapdeployment.pm
```diff
@@ -379,7 +393,7 @@ sub qesap_execute {
 sub qesap_prepare_env {
     my (%args) = @_;
     my $variables = $args{openqa_variables};
-    my $provider = lc get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $provider = $args{provider};
```
lpalovsky, line 396: Maybe we can convert the provider name to lowercase here?
Just to make it more tolerable against incorrect input.

lib/trento.pm
```diff
@@ -220,13 +224,42 @@ sub config_cluster {
...
+}
+
 =head3 deploy_qesap
 
 Deploy a SAP Landscape using a previously configured qe-sap-deployment
 =cut
 
 sub deploy_qesap {
     qesap_sh_deploy(SSH_KEY);
+    my $inventory = qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER'));
```
dzedro, line 260: Why you replace get_required_var('PUBLIC_CLOUD_PROVIDER') for argument in lib/qesapdeployment.pm and then use get_required_var('PUBLIC_CLOUD_PROVIDER') here ?
Also the argument is not used anywhere in this PR, it's just confusing for me.

mpagot: I like the idea to not use any PUBLIC_CLOUD variables directly in the qesapdeploy library, as the qesapdeploy library is not part of any publiccloud library. For this reason, I moved one layer up this dependency. 
trento is not part of PC too so the same is valid for trento lib too, but I decided to move one step at the time.
The next step is to move get_required_var('PUBLIC_CLOUD_PROVIDER') one layer up, in the test

lib/qesapdeployment.pm
mpagot: Agree, 
Generally speaking, this change is an API change. This PR is mostly for Trento testing, and Trento is not using qesap_prepare_env yet. I decided to revert this change at all.

dzedro (approved):

lpalovsky (approved):
LGTM

PR #15574, opened 2022-09-22
publiccloud : replace typing of real text with env variables

asmorodskyi:


lib/publiccloud/provider.pm
```diff
@@ -189,8 +189,8 @@ sub run_img_proof {
     $cmd .= '--no-cleanup ';
     $cmd .= '--collect-vm-info ';
     $cmd .= '--service-account-file "' . $args{credentials_file} . '" ' if ($args{credentials_file});
-    $cmd .= "--access-key-id '" . $args{key_id} . "' " if ($args{key_id});
-    $cmd .= "--secret-access-key '" . $args{key_secret} . "' " if ($args{key_secret});
+    #TODO: this if is just dirty hack which needs to be replaced with something more saine ASAP.
+    $cmd .= "--access-key-id \$AWS_ACCESS_KEY_ID --secret-access-key \$AWS_SECRET_ACCESS_KEY" if(is_ec2());
```
mimi1vx, line 193: small 
`$cmd .= '--access-key-id $AWS_ACCESS_KEY_ID --secret-access-key $AWS_SECRET_ACCESS_KEY' if (is_ec2)`

mimi1vx: + hated perltidy ..

mimi1vx (approved):

lib/publiccloud/provider.pm
asmorodskyi: ok I will try that

lib/publiccloud/provider.pm
```diff
@@ -189,8 +189,8 @@ sub run_img_proof {
     $cmd .= '--no-cleanup ';
     $cmd .= '--collect-vm-info ';
     $cmd .= '--service-account-file "' . $args{credentials_file} . '" ' if ($args{credentials_file});
-    $cmd .= "--access-key-id '" . $args{key_id} . "' " if ($args{key_id});
-    $cmd .= "--secret-access-key '" . $args{key_secret} . "' " if ($args{key_secret});
+    #TODO: this if is just dirty hack which needs to be replaced with something more saine ASAP.
```
grisu48, line 192:
```suggestion
    #TODO: this if is just dirty hack which needs to be replaced with something more sane ASAP.
```

asmorodskyi: fixed

PR #15578, opened 2022-09-22
Enable trufflehog (again) 

foursixnine:
https://github.com/trufflesecurity/trufflehog/issues/666 has been resolved, so we can enable TruffleHog again

After merging this, we should not have trouble anymore with TruffleHog again: https://github.com/foursixnine/os-autoinst-distri-opensuse/actions/runs/3107033239/jobs/5034650536


foursixnine:
Let's see what breaks

PR #15579, opened 2022-09-23
Add a timeout for the 15-sp4 failure ipsec_client

punkioudi:
- Related ticket: https://progress.opensuse.org/issues/117208
- Needles: N/A 
- Verification run: [ipsec_server](https://openqa.suse.de/tests/9602089) | [ipsec_client](https://openqa.suse.de/tests/9602088)

### DIFF
--- a/tests/security/cc/ipsec/ipsec_client.pm
+++ b/tests/security/cc/ipsec/ipsec_client.pm
@@ -42,6 +42,8 @@ sub run {
     # Test the IPSec connection
     assert_script_run('ipsec start');
     assert_script_run('stime=$(date +\'%H:%M:%S\')');
+    record_soft_failure("poo#117208 - The addition of the sleep command is a temporary workaround");
+    sleep 60;
     assert_script_run('ipsec up ikev2suse');
 
     # Test the IPSec connection
--- a/tests/fips/mozilla_nss/firefox_nss.pm
+++ b/tests/fips/mozilla_nss/firefox_nss.pm
@@ -191,7 +191,6 @@ sub run {
 
     # Firefox Preferences
     firefox_preferences;
-
     # Search "Certificates" section
     search_certificates;
 


paolostivanin (approved):

tests/security/cc/ipsec/ipsec_client.pm
```diff
@@ -42,6 +42,7 @@ sub run {
     # Test the IPSec connection
     assert_script_run('ipsec start');
     assert_script_run('stime=$(date +\'%H:%M:%S\')');
+    sleep 60;
```
foursixnine, line 45: is there a way to not rely on `sleep`, but rather check something either in the logs or poll if a device is created? I can imagine that on heavy-loaded hosts, this could fail...

foursixnine (changes requested):

tests/security/cc/ipsec/ipsec_client.pm
punkioudi: > is there a way to not rely on `sleep`, but rather check something either in the logs or poll if a device is created? I can imagine that on heavy-loaded hosts, this could fail...

Hi Santi, we are doing it really temporarily because it fails all week in maintenance. Next week we will push another workaround, which avoids sleep :)

foursixnine: make it a soft fail with a reference to the ticket and I'm good :D

foursixnine: > record_info("The addition of the sleep command is a temporary workaround.", "It will be replaced soon :)");

still a reference to a  ticket is missing... (a soft-failure would be better)

punkioudi: done!

foursixnine (approved):

PR #15582, opened 2022-09-23
Improve wait_still_screen so it better handles slow connection

pdostal:
- Related issue: https://openqa.suse.de/tests/9588218#step/firefox_headers/12
- Verification run:


lib/x11test.pm
```diff
@@ -704,7 +704,8 @@ sub firefox_open_url {
         }
     }
     enter_cmd_slow "$url";
-    wait_still_screen 2, 4;
+    # Wait until screen is still for 5s or give up after 30s
+    wait_still_screen 5, 30;
```
okurz, line 708: This means that on every call *at least* 5s are spent. Wouldn't it be better to try to get rid of this call completely? The previous statement says `enter_cmd_slow "$url"`, how about using the same approach as in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/x11/chromium.pm#L20 ?

okurz (changes requested):

lib/x11test.pm
pdostal: This is not about entering wrong command but about loading the page.
Previously the `wait_still_screen` got triggered about two seconds and as the page was not yet loaded the `send_key_until_needlematch` sent `F5` which aborted the new page loading and reloaded the previous page. - That's what I'm trying to solve.

I agree that there might be better approach so I'm open to other suggestions.

Martchus: When it is about loading the page I'm wondering why it isn't possible to create a needle for it.

okurz: Yes, sounds like an explicit `assert_screen` would be much better.

pdostal: The `chrome` and `chromium` are single small tests but we have 36 different scenarios where we call `firefox_open_url()` subroutine with different URLs.

Martchus: I understand. It would of course be possible to let the user of the `firefox_open_url()` function specify a needle for the page but of course doing this for 36 different places is a lot of effort.

If `wait_still_screen` returns while a slowly loading page is actually still loading, doesn't that mean the loading indication is wrongfully considered "still"? Maybe increasing the similarity level would help than? It would actually make sense that Firefox's loading indication still yields a high similarity considering it is only a very small portion of the screen that changes. To check that you could have a look at the logs which should contain the similarity level when the screen is considered "still".

pdostal: @Martchus @okurz can you please check now?

Martchus (approved):
In your latest verification run the `wait_still_screen` calls always result in `last detected similarity is 1000000`. That means the screen is now definitely still when the call returns (and it does not sometimes already return on a quite low similarity like around 52 as in https://openqa.suse.de/tests/9602421). So that seems like an improvement, indeed.

pdostal:
I think that another problem was the `wait_still_screen` timeout of 4 seconds. Is there something else I can do?

Martchus:
> I think that another problem was the `wait_still_screen` timeout of 4 seconds.

That's true. In your verification run it sometimes took more than 4 seconds for the screen to be "still" (although I'm not 100 % sure whether all of those occurrences were about loading Firefox pages).

> Is there something else I can do?

I wouldn't know considering creating tons of needles for the specific pages is likely too much effort.

lib/x11test.pm
okurz: The use of similarity level helps a bit but I still suggest to think about the use of a proper `assert_screen`. Look at what https://openqa.opensuse.org/tests/2758259#step/firefox/12 is doing trying to detect the "reload" button. Only if it shows the page is considered to be fully loaded. Wouldn't that work? As alternative if we need call specific matches compare to the use of `target_match` in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/susedistribution.pm#L298 . Wherever `x11_start_program` is used an assert_screen call is defined with specific needles needed.

okurz (changes requested):
see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15582#discussion_r983458645

lib/x11test.pm
Martchus: The reload button might not work if the page loads further data via AJAX/WebSockets but you could give it a try.

pdostal: > The use of similarity level helps a bit but I still suggest to think about the use of a proper assert_screen. Look at what https://openqa.opensuse.org/tests/2758259#step/firefox/12 is doing trying to detect the "reload" button. Only if it shows the page is considered to be fully loaded. Wouldn't that work? 

@okurz the problem is:
1) Typing URL and sending return
2) Waiting 4 seconds for still screen
3) Checking if the reload button is present and if it is not, then sending F5

But when the page is loading and user sends F5 then **previous** page is reloaded.

I don't really know what else should I do.

Martchus: I don't think @okurz 's suggestion was to ever send F5. The idea was just to *look* at the reload button (via a needle) to check whether the page is still loading (or actually to assert that it is not loading anymore). However, as mentioned before I'm not sure whether that's a reliable method in all cases.

pdostal: Thank you @Martchus, I think that this might work.

Only one issue I'm still dealing with is that sometimes the reload button is [greyed out](https://pdostal-server.suse.cz/tests/981#step/firefox_extcontent/10) and I don't understand why but the page seems to be loaded so we may just ignore that.

PR #15583, opened 2022-09-23
Adding new python runltp-ng for LTP tests

m-dati:

- Related ticket: https://progress.opensuse.org/issues/116668
- Needles: none
- Verification run: TBD

### DIFF
--- a/tests/publiccloud/run_ltp.pm
+++ b/tests/publiccloud/run_ltp.pm
@@ -24,6 +24,9 @@ use version_utils;
 
 our $root_dir = '/root';
 
+# LTP runtime can be 'python' (new runner) or the default old 'perl' runner
+my $ltp_runtime = get_var('LTP_RUNTIME', 'perl');
+
 sub get_ltp_rpm
 {
     my ($url) = @_;
@@ -48,10 +51,18 @@ sub instance_log_args
 sub upload_ltp_logs
 {
     my ($self) = @_;
-    my $log_file = Mojo::File::path('ulogs/result.json');
+    my $log_file;
     my $ltp_testsuite = get_required_var('LTP_COMMAND_FILE');
+    if ($ltp_runtime =~ /^python/i) {
+        $log_file = Mojo::File::path('ulogs/result.json');
+        upload_logs("$root_dir/result.json", log_name => $log_file->basename, failok => 1);
+    } else {
+        # default ltp runtime: perl
+        $log_file = Mojo::File::path('ulogs/ltp_log.json');
+        upload_logs("$root_dir/ltp_log.raw", log_name => 'ltp_log.raw', failok => 1);
+        upload_logs("$root_dir/ltp_log.json", log_name => $log_file->basename, failok => 1);
 
-    upload_logs("$root_dir/result.json", log_name => $log_file->basename, failok => 1);
+    }
 
     return unless -e $log_file->to_string;
 
@@ -84,8 +95,11 @@ sub run {
     my ($self, $args) = @_;
     my $arch = check_var('PUBLIC_CLOUD_ARCH', 'arm64') ? 'aarch64' : 'x86_64';
     my $ltp_repo = get_var('LTP_REPO', 'https://download.opensuse.org/repositories/benchmark:/ltp:/stable/' . generate_version("_") . '/');
+
     my $provider;
     my $instance;
+
+    record_info('LTP RUNTIME', $ltp_runtime);
 
     select_host_console();
 
@@ -129,8 +143,12 @@ sub run {
         }
     }
 
-    my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/acerv/runltp-ng.git");
-    my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "ssh");
+    my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/metan-ucw/runltp-ng.git");
+    my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "master");
+    record_info('LTP CLONE REPO', "Repo: " . $runltp_ng_repo . "\nBranch: " . $runltp_ng_branch);
+    # Temporary code for ltp_runtime option check, until transition to python will be completed.
+    die('Not valid python LTP_RUN_NG_REPO provided')
+      if ($ltp_runtime =~ /^python/i and $runltp_ng_repo =~ /\/metan-ucw\//i);
     assert_script_run("git clone -q --single-branch -b $runltp_ng_branch --depth 1 $runltp_ng_repo");
     $instance->run_ssh_command(cmd => 'sudo CREATE_ENTRIES=1 ' . get_ltproot() . '/IDcheck.sh', timeout => 300);
     record_info('Kernel info', $instance->run_ssh_command(cmd => q(rpm -qa 'kernel*' --qf '%{NAME}\n' | sort | uniq | xargs rpm -qi)));
@@ -140,27 +158,44 @@ sub run {
 
     assert_script_run($log_start_cmd);
 
-    zypper_call("in -y python3-paramiko python3-scp");
+    my $cmd;
 
-    my $ltp_suite_timeout=1200;
+    if ($ltp_runtime =~ /^python/i) {
+        zypper_call("in -y python3-paramiko python3-scp");
 
-    my $sut_opt= '--sut=ssh';
-    $sut_opt .= ':user=' . $instance->username;
-    $sut_opt .= ':sudo=1';
-    $sut_opt .= ':key_file=' . $instance->ssh_key;
-    $sut_opt .= ':host=' . $instance->public_ip;
-    $sut_opt .= ':reset_command=\'' . $reset_cmd . '\'';
-    $sut_opt .= ':hostkey_policy=missing';
-    $sut_opt .= ':known_hosts=/dev/null';
+        my $sut = ':user=' . $instance->username;
+        $sut .= ':sudo=1';
+        $sut .= ':key_file=' . $instance->ssh_key;
+        $sut .= ':host=' . $instance->public_ip;
+        $sut .= ':reset_command=\'' . $reset_cmd . '\'';
+        $sut .= ':hostkey_policy=missing';
+        $sut .= ':known_hosts=/dev/null';
 
-    my $cmd = 'python3 runltp-ng/runltp-ng ';
-    $cmd .= "--json-report=$root_dir/result.json ";
-    $cmd .= '--verbose ';
-    $cmd .= '--suite-timeout=' . $ltp_suite_timeout . ' ';
-    $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
-    $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
-    $cmd .= $sut_opt;
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--sut=ssh' . $sut . ' ';
+    } elsif ($ltp_runtime =~ /^perl$/i) {
+        # default ltp runtime: perl
+        my $backend .= ':user=' . $instance->username;
+        $backend .= ':key_file=' . $instance->ssh_key;
+        $backend .= ':host=' . $instance->public_ip;
+        $backend .= ':reset_command=\'' . $reset_cmd . '\'';
+        $backend .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
 
+        $cmd = 'perl -I runltp-ng runltp-ng/runltp-ng ';
+        $cmd .= '--logname=ltp_log --verbose ';
+        $cmd .= '--timeout=1200 ';
+        $cmd .= '--run ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--backend=ssh' . $backend . ' ';
+        $cmd .= '--json_filter=openqa ';
+    } else {
+        die('Not supported LTP_RUNTIME provided');
+    }
+    record_info('LTP START', 'Command launch');
     assert_script_run($cmd, timeout => get_var('LTP_TIMEOUT', 30 * 60));
 }
 
@@ -216,6 +251,11 @@ This regex is used to exclude tests from command file.
 =head2 LTP_REPO
 
 The repo which will be added and is used to install LTP package.
+
+=head2 LTP_RUNTIME
+
+The language and related environment that will be used to run LTP package.
+See more info in variables.md in the home of this repo.
 
 =head2 LTP_KNOWN_ISSUES
 
--- a/tests/publiccloud/run_ltp.pm
+++ b/tests/publiccloud/run_ltp.pm
@@ -24,8 +24,8 @@ use version_utils;
 
 our $root_dir = '/root';
 
-# LTP runtime can be 'python' (new runner) or the default old 'perl' runner
-my $ltp_runtime = get_var('LTP_RUNTIME', 'perl');
+# LTP runtime can be 1 for 'python' (new runner) or 0 for the default old 'perl' runner
+my $ltp_runtime = get_var('LTP_RUNTIME_SWITCH', 0);
 
 sub get_ltp_rpm
 {
@@ -53,7 +53,7 @@ sub upload_ltp_logs
     my ($self) = @_;
     my $log_file;
     my $ltp_testsuite = get_required_var('LTP_COMMAND_FILE');
-    if ($ltp_runtime =~ /^python/i) {
+    if ($ltp_runtime) {
         $log_file = Mojo::File::path('ulogs/result.json');
         upload_logs("$root_dir/result.json", log_name => $log_file->basename, failok => 1);
     } else {
@@ -61,7 +61,6 @@ sub upload_ltp_logs
         $log_file = Mojo::File::path('ulogs/ltp_log.json');
         upload_logs("$root_dir/ltp_log.raw", log_name => 'ltp_log.raw', failok => 1);
         upload_logs("$root_dir/ltp_log.json", log_name => $log_file->basename, failok => 1);
-
     }
 
     return unless -e $log_file->to_string;
@@ -99,7 +98,7 @@ sub run {
     my $provider;
     my $instance;
 
-    record_info('LTP RUNTIME', $ltp_runtime);
+    record_info('LTP RUNTIME', ($ltp_runtime) ? 'python' : 'perl');
 
     select_host_console();
 
@@ -148,7 +147,7 @@ sub run {
     record_info('LTP CLONE REPO', "Repo: " . $runltp_ng_repo . "\nBranch: " . $runltp_ng_branch);
     # Temporary code for ltp_runtime option check, until transition to python will be completed.
     die('Not valid python LTP_RUN_NG_REPO provided')
-      if ($ltp_runtime =~ /^python/i and $runltp_ng_repo =~ /\/metan-ucw\//i);
+      if ($ltp_runtime and $runltp_ng_repo =~ /\/metan-ucw\//i);
     assert_script_run("git clone -q --single-branch -b $runltp_ng_branch --depth 1 $runltp_ng_repo");
     $instance->run_ssh_command(cmd => 'sudo CREATE_ENTRIES=1 ' . get_ltproot() . '/IDcheck.sh', timeout => 300);
     record_info('Kernel info', $instance->run_ssh_command(cmd => q(rpm -qa 'kernel*' --qf '%{NAME}\n' | sort | uniq | xargs rpm -qi)));
@@ -160,7 +159,7 @@ sub run {
 
     my $cmd;
 
-    if ($ltp_runtime =~ /^python/i) {
+    if ($ltp_runtime) {
         zypper_call("in -y python3-paramiko python3-scp");
 
         my $sut = ':user=' . $instance->username;
@@ -177,7 +176,7 @@ sub run {
         $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
         $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
         $cmd .= '--sut=ssh' . $sut . ' ';
-    } elsif ($ltp_runtime =~ /^perl$/i) {
+    } else {
         # default ltp runtime: perl
         my $backend .= ':user=' . $instance->username;
         $backend .= ':key_file=' . $instance->ssh_key;
@@ -192,8 +191,6 @@ sub run {
         $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
         $cmd .= '--backend=ssh' . $backend . ' ';
         $cmd .= '--json_filter=openqa ';
-    } else {
-        die('Not supported LTP_RUNTIME provided');
     }
     record_info('LTP START', 'Command launch');
     assert_script_run($cmd, timeout => get_var('LTP_TIMEOUT', 30 * 60));
@@ -252,9 +249,9 @@ This regex is used to exclude tests from command file.
 
 The repo which will be added and is used to install LTP package.
 
-=head2 LTP_RUNTIME
+=head2 LTP_RUNTIME_SWITCH
 
-The language and related environment that will be used to run LTP package.
+Select the language and related environment that will be used to run LTP package.
 See more info in variables.md in the home of this repo.
 
 =head2 LTP_KNOWN_ISSUES
--- a/tests/publiccloud/run_ltp.pm
+++ b/tests/publiccloud/run_ltp.pm
@@ -51,7 +51,6 @@ sub upload_ltp_logs
     my $log_file = Mojo::File::path('ulogs/result.json');
     my $ltp_testsuite = get_required_var('LTP_COMMAND_FILE');
 
-#    upload_logs("$root_dir/ltp_log.raw", log_name => 'ltp_log.raw', failok => 1);
     upload_logs("$root_dir/result.json", log_name => $log_file->basename, failok => 1);
 
     return unless -e $log_file->to_string;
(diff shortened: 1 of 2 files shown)


tests/publiccloud/run_ltp.pm
```diff
@@ -130,8 +130,8 @@ sub run {
         }
     }
 
-    my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/metan-ucw/runltp-ng.git");
-    my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "master");
+    my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/acerv/runltp-ng.git");
+    my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "ssh");
```
asmorodskyi, line 134: this raise a warning for me , why we not using master ? is ssh feature not fully ready to use ? I think I would need to talk about that with kernel-qa before fully switching to this approach ...

tests/publiccloud/run_ltp.pm
```diff
@@ -48,11 +48,11 @@ sub instance_log_args
 sub upload_ltp_logs
 {
     my ($self) = @_;
-    my $log_file = Mojo::File::path('ulogs/ltp_log.json');
+    my $log_file = Mojo::File::path('ulogs/result.json');
     my $ltp_testsuite = get_required_var('LTP_COMMAND_FILE');
 
-    upload_logs("$root_dir/ltp_log.raw", log_name => 'ltp_log.raw', failok => 1);
-    upload_logs("$root_dir/ltp_log.json", log_name => $log_file->basename, failok => 1);
+#    upload_logs("$root_dir/ltp_log.raw", log_name => 'ltp_log.raw', failok => 1);
```
asmorodskyi, line 54: what about raw result ? they don't have it anymore ?

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +141,20 @@ sub run {
...
-    $cmd .= '--timeout=1200 ';
-    $cmd .= '--run ' . get_required_var('LTP_COMMAND_FILE') . ' ';
-    $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
-    $cmd .= '--backend=ssh';
+    zypper_call("in -y python3-paramiko python3-scp");
+ 
+    my $cmd = 'python3 runltp-ng/runltp-ng ';
+    $cmd .= "--json-report=$root_dir/result.json ";
+    $cmd .= '--verbose ';
+    $cmd .= '--suite-timeout=2400 ';
```
asmorodskyi, line 149: why double the timeout ?

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +141,20 @@ sub run {
...
+    $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+    $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+    $cmd .= '--sut=ssh';
     $cmd .= ':user=' . $instance->username;
+    $cmd .= ':sudo=1';
     $cmd .= ':key_file=' . $instance->ssh_key;
     $cmd .= ':host=' . $instance->public_ip;
     $cmd .= ':reset_command=\'' . $reset_cmd . '\'';
-    $cmd .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
```
asmorodskyi, line 154: why we not passing this ssh options anymore ? I think ssh tunnel without them may get into trouble when host will change host signature ( and this happens a lot in public cloud )

tests/publiccloud/run_ltp.pm
asmorodskyi: @pevik , @acerv can you please comment on this

tests/publiccloud/run_ltp.pm
m-dati: Please consider that I executed this below test on openqa (clone of [9562977](https://openqa.suse.de/tests/9562977)) that even **timed out with this 2400 timeout** against the old one 1200: 
https://openqa.suse.de/tests/9588743

I'd even suggest to apply thi change:
```
    my $timeout = get_var('LTP_TIMEOUT', 30 * 60);
```
then:
```
    my $cmd = 'python3 runltp-ng/runltp-ng ';
...
    $cmd .= '--suite-timeout=' . $timeout . ' ';
...

```

tests/publiccloud/run_ltp.pm
acerv: @asmorodskyi the feature is under development for a main reason: runltp-ng will be part of the LTP upstream. That means we are going to test a simple version, plain python 3.6+ without external dependency. When things are getting more stable and fixed, SSH support will be introduced. For now, SSH stays outside master to wait for LTP upstream test

tests/publiccloud/run_ltp.pm
acerv: raw results were really only a stdout redirection. You can use tee command and bring stdout if you need it

tests/publiccloud/run_ltp.pm
acerv: `ssh` command is no longer used, so SSH options are not available
https://github.com/acerv/runltp-ng/blob/945196ca4b6814ce7efdb04afcf9c0be022d4ea5/ltp/ssh.py#L87

But you can open an issue request and provide some new SSH supports to add. We can add more specific options to the ssh SUT option.

asmorodskyi: we do need to pass this "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" and it is NO GO for us before it won't be implemented ( failure rate will be too high ... )

asmorodskyi (changes requested):
Due to no ability to pass ssh options to the tunnel we putting this on-hold

tests/publiccloud/run_ltp.pm
acerv: "StrictHostKeyChecking=no" can be implemented here https://github.com/acerv/runltp-ng/blob/945196ca4b6814ce7efdb04afcf9c0be022d4ea5/ltp/ssh.py#L89

Using:

`self._client.set_missing_host_key_policy(MissingHostKeyPolicy())`

This can be implemented adding an option to set host key policy.

pevik:
@asmorodskyi thanks a lot for your time to test!

tests/publiccloud/run_ltp.pm
asmorodskyi: @m-dati please remove this line

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +141,27 @@ sub run {
...
-    $cmd .= '--run ' . get_required_var('LTP_COMMAND_FILE') . ' ';
-    $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
-    $cmd .= '--backend=ssh';
-    $cmd .= ':user=' . $instance->username;
-    $cmd .= ':key_file=' . $instance->ssh_key;
-    $cmd .= ':host=' . $instance->public_ip;
-    $cmd .= ':reset_command=\'' . $reset_cmd . '\'';
-    $cmd .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
-    $cmd .= '--json_filter=openqa ';
+    zypper_call("in -y python3-paramiko python3-scp");
```
asmorodskyi, line 144: maybe you can add this to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15583/files#diff-49293c308cb7b8e8f6bbfc7f14a6fa90507cd690aa9d5ab41c1f60f96e4cb5ecL116 ?

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +141,27 @@ sub run {
...
+
+    my $timeout=2400;
```
asmorodskyi, line 146: 1. can also try to fallback to initial 1200 as it was before ?
2. not sure if it make sense to keep it as variable . but if you think it is than we need to give it more concrete name ... e.g. "$ltp_suite_timeout"

m-dati: Actually this variable is to follow up discussion on it, to make update easy and also because it seems to depend on the sub-tests to run

m-dati: updated

tests/publiccloud/run_ltp.pm
```diff
@@ -48,11 +51,18 @@ sub instance_log_args
 sub upload_ltp_logs
 {
     my ($self) = @_;
-    my $log_file = Mojo::File::path('ulogs/ltp_log.json');
+    my $log_file = '';
```
asmorodskyi, line 54: it is a bit confusing to init something which suppose to be an object with empty string . Can you just drop this line ?

tests/publiccloud/run_ltp.pm
```diff
@@ -24,6 +24,9 @@ use version_utils;
 
 our $root_dir = '/root';
 
+# LTP runtime set to 'python' or default is 'perl'
+my $ltp_runtime = get_var('LTP_RUNTIME', 'Perl');
```
asmorodskyi, line 28:
```suggestion
my $ltp_runtime = get_var('LTP_RUNTIME', 'perl');
```

tests/publiccloud/run_ltp.pm
```diff
@@ -129,9 +142,19 @@ sub run {
             set_var('LTP_COMMAND_EXCLUDE', $exclude);
         }
     }
+   
+    my $runltp_ng_repo = '';
+    my $runltp_ng_branch = '';
 
-    my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/metan-ucw/runltp-ng.git");
-    my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "master");
+    if ( $ltp_runtime =~ /^python/i ) {
```
asmorodskyi, line 149: I would not touch this lines when you trigger test with `LTP_RUNTIME=python`  you may just add also `LTP_RUN_NG_REPO=https://github.com/acerv/runltp-ng.git` and `LTP_RUN_NG_BRANCH=ssh` , having logic which set defaults to variables based on other variables it is too much

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +158,48 @@ sub run {
...
+        $sut .= ':key_file=' . $instance->ssh_key;
+        $sut .= ':host=' . $instance->public_ip;
+        $sut .= ':reset_command=\'' . $reset_cmd . '\'';
+        $sut .= ':hostkey_policy=missing';
+        $sut .= ':known_hosts=/dev/null';
+
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        # $cmd .= '--suite-timeout=' . $ltp_suite_timeout . ' ';
```
asmorodskyi, line 180: please remove commented code

tests/publiccloud/run_ltp.pm
m-dati: Please, is this way also acceptable? https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/258b96b13606dfaf9801f3d33f985b128bff7fa6/tests/publiccloud/run_ltp.pm#L146-L147

m-dati: I removed the switch from here, keeping the original code: as clarified, for repo different from default we'll set the parameters at clone_job run

m-dati:
Started 2 proof tests after last update:

- PERL: (default)
clone_job.pl --from openqa.suse.de --host openqa.suse.de --skip-chained-deps --skip-download 9562977 TEST=pc_ltp_cve@runltp-ng#MD23 CASEDIR=https://github.com/m-dati/os-autoinst-distri-opensuse.git#runltp-ng BUILD=runltp_py_AC1 _GROUP_ID=447 LTP_RUNTIME=perl
    * job [#9659635](https://openqa.suse.de/t9659635)

- PYTHON:
clone_job.pl --from openqa.suse.de --host openqa.suse.de --skip-chained-deps --skip-download 9562977 TEST=pc_ltp_cve@runltp-ng#MD24 CASEDIR=https://github.com/m-dati/os-autoinst-distri-opensuse.git#runltp-ng BUILD=runltp_py_AC1 _GROUP_ID=447 LTP_RUNTIME=python LTP_RUN_NG_BRANCH=ssh LTP_RUN_NG_REPO=https://github.com/acerv/runltp-ng.git
    * job [#9659633](https://openqa.suse.de/t9659633)

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
-    $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
-    $cmd .= '--backend=ssh';
-    $cmd .= ':user=' . $instance->username;
-    $cmd .= ':key_file=' . $instance->ssh_key;
-    $cmd .= ':host=' . $instance->public_ip;
-    $cmd .= ':reset_command=\'' . $reset_cmd . '\'';
-    $cmd .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
-    $cmd .= '--json_filter=openqa ';
-    assert_script_run($cmd, timeout => get_var('LTP_TIMEOUT', 30 * 60));
+    my $ltp_timeout = get_var('LTP_TIMEOUT', 30 * 60);
```
asmorodskyi, line 159: this variable is used in only [one place ](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15583/files#diff-49293c308cb7b8e8f6bbfc7f14a6fa90507cd690aa9d5ab41c1f60f96e4cb5ecR199), I think we can drop this variable and just directly call `get_var('LTP_TIMEOUT', 30 * 60)` in place where variable used now

tests/publiccloud/run_ltp.pm
```diff
@@ -24,6 +24,9 @@ use version_utils;
 
 our $root_dir = '/root';
 
+# LTP runtime set to 'python' or default is 'perl'
```
grisu48, line 27:
```suggestion
# LTP runtime can be 'python' (new runner) or the default old 'perl' runner
```

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
-    $cmd .= '--backend=ssh';
-    $cmd .= ':user=' . $instance->username;
-    $cmd .= ':key_file=' . $instance->ssh_key;
-    $cmd .= ':host=' . $instance->public_ip;
-    $cmd .= ':reset_command=\'' . $reset_cmd . '\'';
-    $cmd .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
-    $cmd .= '--json_filter=openqa ';
-    assert_script_run($cmd, timeout => get_var('LTP_TIMEOUT', 30 * 60));
+    my $ltp_timeout = get_var('LTP_TIMEOUT', 30 * 60);
+    my $ltp_suite_timeout = 1200;
```
asmorodskyi, line 160: first of all as it was clarified with @acerv  this is not "suite timeout" but "test timeout" . and also since this variable used in single place we can just do `$cmd .= '--timeout=1200 ';` ( as it was before ) and no need to introduce variable for this

tests/publiccloud/run_ltp.pm
```diff
@@ -85,9 +95,12 @@ sub run {
     my ($self, $args) = @_;
     my $arch = check_var('PUBLIC_CLOUD_ARCH', 'arm64') ? 'aarch64' : 'x86_64';
     my $ltp_repo = get_var('LTP_REPO', 'https://download.opensuse.org/repositories/benchmark:/ltp:/stable/' . generate_version("_") . '/');
+    
     my $provider;
     my $instance;
 
+    record_info('LTP RUNTIME', $ltp_runtime);
```
grisu48, line 102: The `LTP_RUNTIME` setting can be directly seen in the SETTINGS tab, I would say that this is not really adding some new information anywhere.

tests/publiccloud/run_ltp.pm
m-dati: ok

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
+        $cmd = 'perl -I runltp-ng runltp-ng/runltp-ng ';
+        $cmd .= '--logname=ltp_log --verbose ';
+        $cmd .= '--timeout=' . $ltp_suite_timeout . ' ';
+        $cmd .= '--run ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--exclude \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--backend=ssh' . $backend . ' ';
+        $cmd .= '--json_filter=openqa ';
+    }
+    record_info('LTP START TEST', $cmd);
+    assert_script_run($cmd, timeout => $ltp_timeout + 3);
```
asmorodskyi, line 199: I think we don't need "+3" here

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
-    $cmd .= '--json_filter=openqa ';
-    assert_script_run($cmd, timeout => get_var('LTP_TIMEOUT', 30 * 60));
+    my $ltp_timeout = get_var('LTP_TIMEOUT', 30 * 60);
+    my $ltp_suite_timeout = 1200;
+    my $cmd;
+  
+    if ( $ltp_runtime =~ /^python/i ) {
+        zypper_call("in -y python3-paramiko python3-scp");
+
+        my $sut= '';
```
asmorodskyi, line 166:
```suggestion
        my $sut= ':user=' . $instance->username;
```

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
+
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--sut=ssh' . $sut . ' ';
+    } else {
+        # default ltp runtime: perl
+        my $backend= '';
```
asmorodskyi, line 183:
```suggestion
        my $backend= ':user=' . $instance->username;
```

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
+        $sut .= ':hostkey_policy=missing';
+        $sut .= ':known_hosts=/dev/null';
+
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--sut=ssh' . $sut . ' ';
+    } else {
```
grisu48, line 181:
```suggestion
    } elsif ( $ltp_runtime eq 'perl') {
```

This is prone to use errors when extending the test cases. One should only act on the acceptable input and reject other methods to avoid user confusion.

Here I could just e.g. create a test run with `LTP_RUNTIME=pythn` and it would happily run the perl codebase, while it should fail and tell me that I have a typo in my test settings.

tests/publiccloud/run_ltp.pm
m-dati: That setting was a choice for uniform programming, so that each needed parameter line has the same format, but I will apply the proposed change.

tests/publiccloud/run_ltp.pm
```diff
@@ -129,9 +142,11 @@ sub run {
             set_var('LTP_COMMAND_EXCLUDE', $exclude);
         }
     }
+    
+    my $runltp_ng_repo   = get_var( "LTP_RUN_NG_REPO", "https://github.com/metan-ucw/runltp-ng.git" );
+    my $runltp_ng_branch = get_var( "LTP_RUN_NG_BRANCH", "master" );
+    record_info( 'LTP CLONE REPO',  "Repo: " . $runltp_ng_repo . "Branch: " . $runltp_ng_branch  );
```
grisu48, line 148:
```suggestion
    record_info( 'LTP CLONE REPO',  "Repo: " . $runltp_ng_repo . "\nBranch: " . $runltp_ng_branch  );
```

tests/publiccloud/run_ltp.pm
m-dati: That setting was a choice for uniform programming, so that each needed parameter line has the same format, but I will apply the proposed change.

tests/publiccloud/run_ltp.pm
```diff
@@ -141,19 +156,47 @@ sub run {
...
+        $sut .= ':reset_command=\'' . $reset_cmd . '\'';
+        $sut .= ':hostkey_policy=missing';
+        $sut .= ':known_hosts=/dev/null';
+
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--sut=ssh' . $sut . ' ';
```
grisu48, line 180:
```suggestion
        $cmd .= '--sut=ssh ' . $sut . ' ';
```

Missing space?

grisu48 (commented):
Nice job! I 've added some suggestions and would also add the new variables to `variables.md` in the root directory of the repository.

tests/publiccloud/run_ltp.pm
asmorodskyi: I can agree on that but in such a case we also need 
```
} else {
die ('Not supported LTP_RUNTIME provided'); 
}
```

tests/publiccloud/run_ltp.pm
asmorodskyi: hm but we have VRs which proving that it works like that :thinking:  how is that possible ?

tests/publiccloud/run_ltp.pm
m-dati: It is to help the log analysis without switching context and it has been useful to check the correct application of the new  parameter keeping the attention on the execution log, marked with the expected runtime.

tests/publiccloud/run_ltp.pm
m-dati: Sorry for my reply, but the switching context should be temporary, while default is the only perl one, until we consolidate a stable python context, that will remain the only one, removing all these options.

tests/publiccloud/run_ltp.pm
asmorodskyi: I think space is not expected here

tests/publiccloud/run_ltp.pm
```diff
@@ -24,6 +24,11 @@ use version_utils;
 
 our $root_dir = '/root';
 
+# LTP runtime can be 'python' (new runner) or the default old 'perl' runner
+my $ltp_runtime = get_var('LTP_RUNTIME','perl');
+
+die ('Not supported LTP_RUNTIME provided') unless ( $ltp_runtime =~ /^(python|perl)/i );
```
m-dati, line 30: This to allow 2 only values as runtime, for https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15583#discussion_r987826773

tests/publiccloud/run_ltp.pm
m-dati: please check https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15583#discussion_r988028961 if ok

m-dati:
- PERL new VR after last commit, in place of [job #9663835](https://openqa.suse.de/t9663835):
    - https://openqa.suse.de/tests/9665070#

tests/publiccloud/run_ltp.pm
grisu48: While it works this is suboptimal style because you now have two places where you check for those variables.

Assuming you would add a third option tomorrow, e.g. `bash`, you would need to add it now it to places: here and in the if block. If you just add it to the if block, you only have one place to update and it is clear that you have to do it.

tests/publiccloud/run_ltp.pm
grisu48: I see, yes that makes sense.:

tests/publiccloud/run_ltp.pm
m-dati: I agree with you if **one** only switch case is present, while I don't think that more than 2 options can happen, otherwise it means to have 3 or more contemporary different _languages_ in LTP projects.
My consideration was that without this initialization, in the case more future switch case in the code should be needed, additional 'elsif' means a more test to add instead of simple 'else' and more complexity. 
Anyway I will update the code as by you suggestion.

asmorodskyi (approved):

tests/publiccloud/run_ltp.pm
```diff
@@ -132,6 +145,9 @@ sub run {
 
     my $runltp_ng_repo = get_var("LTP_RUN_NG_REPO", "https://github.com/metan-ucw/runltp-ng.git");
     my $runltp_ng_branch = get_var("LTP_RUN_NG_BRANCH", "master");
+    record_info('LTP CLONE REPO', "Repo: " . $runltp_ng_repo . "\nBranch: " . $runltp_ng_branch);
+    die('Not valid python LTP_RUN_NG_REPO provided')
+      if ($ltp_runtime =~ /^python/i and $runltp_ng_repo =~ /\/metan-ucw\//i);
```
m-dati, line 150: I added this check to prevent omission of the REPO value in case of python test, where record_info reports the values helpful for debugging of crash case too.

m-dati:
Please check note https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/cdc4a19604fbf6acaaeedc58b968ff5a9373ce7d#r85997806

tests/publiccloud/run_ltp.pm
grisu48: Please add a comment stating that this is temporary and this will be removed when the transition is completed.

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +157,44 @@ sub run {
...
-    $cmd .= '--backend=ssh';
-    $cmd .= ':user=' . $instance->username;
-    $cmd .= ':key_file=' . $instance->ssh_key;
-    $cmd .= ':host=' . $instance->public_ip;
-    $cmd .= ':reset_command=\'' . $reset_cmd . '\'';
-    $cmd .= ':ssh_opts=\'-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\' ';
-    $cmd .= '--json_filter=openqa ';
+    my $cmd;
+
+    if ($ltp_runtime =~ /^python/i) {
```
grisu48, line 162:
```suggestion
    if ($ltp_runtime =~ /^python$/i) {
```

tests/publiccloud/run_ltp.pm
```diff
@@ -141,18 +157,44 @@ sub run {
...
+        $sut .= ':hostkey_policy=missing';
+        $sut .= ':known_hosts=/dev/null';
+
+        $cmd = 'python3 runltp-ng/runltp-ng ';
+        $cmd .= "--json-report=$root_dir/result.json ";
+        $cmd .= '--verbose ';
+        $cmd .= '--run-suite ' . get_required_var('LTP_COMMAND_FILE') . ' ';
+        $cmd .= '--skip-tests \'' . get_var('LTP_COMMAND_EXCLUDE') . '\' ' if get_var('LTP_COMMAND_EXCLUDE');
+        $cmd .= '--sut=ssh' . $sut . ' ';
+    } elsif ($ltp_runtime eq 'perl') {
```
grisu48, line 179: Let's stay consistent. Either use everywhere the case-insensitive regex matching or nowhere. I don't see why we should use a regex here, it adds mental effort when trying to understand it and the only advantage that we have is that the matching is case insensitive.

Maurizio, please pick your side :slightly_smiling_face: Either everywhere perl regex matching or nowhere in this if ... elsif ... block

variables.md
```diff
@@ -110,6 +110,7 @@ LTP_KNOWN_ISSUES | string | | Used to specify a url for a json file with well kn
 LTP_REPO | string | | The repo which will be added and is used to install LTP package.
 LTP_RUN_NG_BRANCH | string | | Define the branch of the LTP_RUN_NG_REPO.
 LTP_RUN_NG_REPO | string | | Define the runltp-ng repo to be used. Default in publiccloud/run_ltp.pm is the upstream master branch from https://github.com/metan-ucw/runltp-ng.git.
+LTP_RUNTIME | string | | Values: **python** or **perl**, default empty = perl. <p>This parameter is used to run proof and consolidation tests for the new python version of the LTP runltp-ng repository, allowing to switch between the original Perl and the new Python repo versions.</p><p>NOTE: when **python** value set, also needed to set: *LTP_RUN_NG_REPO='https://github.com/acerv/runltp-ng.git'* and *LTP_RUN_NG_BRANCH='ssh'*.</p>
```
grisu48, line 113: Line 28 in `run_ltp.pm` suggests that the default value is not empty but `perl`. Also - please don't use html tags in markdown.

tests/publiccloud/run_ltp.pm
m-dati: I avoided to close the right part $, leaving any version tags to still select the same option, like: python3.

m-dati:
I here propose a last change: set boolean LTP_RUNTIME=0 or 1: if 1 it is Python and 0 or undef is perl, similar to other PUBLIC_CLOUD parameters

m-dati:
Here the new VR after last commit:
- PERL:
    - https://openqa.suse.de/t9674944 - LTP_RUNTIME_SWITCH undefined
    - https://openqa.suse.de/t9675303 - LTP_RUNTIME_SWITCH=0
- PYTH:
    - https://openqa.suse.de/t9675299 - LTP_RUNTIME_SWITCH=1, 
         - LTP_RUN_NG_BRANCH=ssh, LTP_RUN_NG_REPO=https://github.com/acerv/runltp-ng.git

tests/publiccloud/run_ltp.pm
```diff
@@ -209,6 +249,11 @@ This regex is used to exclude tests from command file.
 
 The repo which will be added and is used to install LTP package.
 
+=head2 LTP_RUNTIME
```
asmorodskyi, line 252: you forgot to update variable name 
```suggestion
=head2 LTP_RUNTIME_SWITCH
```

PR #15589, opened 2022-09-26
Remove test_repo_setup from maintenance runs

punkioudi:
The `test_repo_setup` module is not needed in maintenance runs. For enabling fips tests in 15-SP4, this module has to be excluded.

- Related ticket: https://progress.opensuse.org/issues/116956
- Needles: N/A
- Verification run: [15-SP4](https://openqa.suse.de/tests/9602455#step/fips_setup/91) | [15-sp5](https://openqa.suse.de/tests/9602613#step/fips_setup/91)

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -2214,7 +2214,8 @@ sub load_applicationstests {
 
 sub load_security_console_prepare {
     loadtest "console/consoletest_setup";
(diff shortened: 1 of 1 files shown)


jouyingbin (approved):
as I remember there is some test needed to import this module, but in most cases is not necessary. LGTM :-)

paolostivanin:
LGTM

PR #15592, opened 2022-09-26
WIP fix redis cli for validate_script_output('redis-cli get foo', sub { m/(nil)/ })

Zaoliang:
need to use svirt remote worker on OSD

### DIFF
--- a/tests/console/redis.pm
+++ b/tests/console/redis.pm
@@ -37,7 +37,7 @@ sub run {
     validate_script_output('redis-cli get foo', sub { m/bar/ });
     validate_script_output('redis-cli pfselftest', sub { m/OK/ });
     validate_script_output('redis-cli flushdb', sub { m/OK/ });
-    validate_script_output('redis-cli get foo', sub { m/(nil)/ });
+    validate_script_output('redis-cli get foo', sub { chomp; m/(nil)/ });
 
     assert_script_run 'curl -O ' . data_url('console/movies.redis');
     assert_script_run('redis-cli -h localhost -p 6379 < ./movies.redis');


schedule/functional/redis.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
pdostal, line 1: This is not necessary. Just use `INCLUDE_MODULES=bootloader_start,boot_to_desktop,redis` when you're running your verifications ;)

tests/console/redis.pm
```diff
@@ -37,7 +38,7 @@ sub run {
     validate_script_output('redis-cli get foo', sub { m/bar/ });
     validate_script_output('redis-cli pfselftest', sub { m/OK/ });
     validate_script_output('redis-cli flushdb', sub { m/OK/ });
-    validate_script_output('redis-cli get foo', sub { m/(nil)/ });
+    validate_script_output('redis-cli get foo', sub { m/(nil)/ }) if (!is_svirt);
```
pdostal, line 41: This is not needed. The output is just not `nil` but `\n` so it can be done better with some perl magic :smile:

pdostal: Maybe this can help: https://perlmaven.com/check-if-string-is-empty-or-has-only-spaces-in-perl

pdostal (changes requested):

schedule/functional/redis.yaml
Zaoliang: yes, thanks!

tests/console/redis.pm
Zaoliang: I still don't understand this, this is working on	qemu_x86_64. Do you mean to use sub { m/(nil) \n/ } ?
thanks!

tests/console/redis.pm
```diff
@@ -37,7 +37,7 @@ sub run {
     validate_script_output('redis-cli get foo', sub { m/bar/ });
     validate_script_output('redis-cli pfselftest', sub { m/OK/ });
     validate_script_output('redis-cli flushdb', sub { m/OK/ });
-    validate_script_output('redis-cli get foo', sub { m/(nil)/ });
+    validate_script_output('redis-cli get foo', sub { m/(nil\n)/ });
```
foursixnine, line 40:
```suggestion
    validate_script_output('redis-cli get foo', sub { chomp; m/(nil)/ });
```

Zaoliang: https://openqa.suse.de/tests/9635022#step/redis/21 still not working, thanks!

foursixnine:
@Zaoliang you can close this one now. Happens that flushdb was deleting all keys, so just checking that the key didn't exist in the store, was good enough

Zaoliang:
close this pr because of #15630, thanks for your feedback and input!

tests/console/redis.pm
pdostal: Something like that. In qemu backend it is `(null)` but in `svirt` it's probably `\n`

PR #15595, opened 2022-09-27
Enable sqlite3 and postgresql_server tests on ALP

DeepthiYV:
- Related ticket:https://progress.opensuse.org/issues/116057
- Needles: No
- Verification run: 
   [alp-0.1-kvm-x86_64](http://deepthi-openqa.qam.suse.de/tests/4014#step/sqlite3/1)

### DIFF
--- a/tests/console/postgresql_server.pm
+++ b/tests/console/postgresql_server.pm
@@ -20,13 +20,20 @@ use testapi;
 use utils 'zypper_call';
 use apachetest qw(setup_pgsqldb destroy_pgsqldb test_pgsql postgresql_cleanup);
 use Utils::Systemd 'systemctl';
+use version_utils qw(is_transactional);
+use transactional qw(trup_call check_reboot_changes);
 
 sub run {
     my $self = shift;
-    $self->select_serial_terminal;
-
     # install the postgresql server package
-    zypper_call "in postgresql-server sudo";
+    if (is_transactional) {
+        select_console 'root-console';
+        trup_call("pkg install postgresql-server sudo");
+        check_reboot_changes;
+    } else {
+        $self->select_serial_terminal;
+        zypper_call "in postgresql-server sudo";
+    }
 
     # start the postgresql service
     systemctl 'start postgresql.service', timeout => 200;
--- a/tests/console/sqlite3.pm
+++ b/tests/console/sqlite3.pm
@@ -19,13 +19,19 @@ use strict;
 use warnings;
 use testapi;
 use utils 'zypper_call';
-
+use version_utils qw(is_transactional);
+use transactional qw(trup_call check_reboot_changes);
 
 sub run {
     my ($self) = @_;
-    $self->select_serial_terminal;
-
-    zypper_call('install sqlite3 expect perl');
+    if (is_transactional) {
+        select_console 'root-console';
+        trup_call("pkg install sqlite3 expect perl");
+        check_reboot_changes;
+    } else {
+        $self->select_serial_terminal;
+        zypper_call('install sqlite3 expect perl');
+    }
 
     my $archive = "sqlite3-tests.data";
     assert_script_run(


okurz (changes requested):
I think it would be better to extract that concept of `just install package` elsewhere and reuse in test modules. On the other hand, a reboot in every single test module sounds quite expensive. Isn't the idea of ALP to run the workload in containers? That would apply here in particular for databases

DeepthiYV:
> I think it would be better to extract that concept of `just install package` elsewhere and reuse in test modules. On the other hand, a reboot in every single test module sounds quite expensive. Isn't the idea of ALP to run the workload in containers? That would apply here in particular for databases

I think installation of the package is handled in single function `trup_call` and as suggested in the ticket - https://progress.opensuse.org/issues/115196#note-20, I just called the trup_call  function in root-console for the transactional update. I will take a note on the reboot after every module installation.
As I understood we are doing smoke test of ALP and so we are making changes. @foursixnine, please correct me if I'm wrong

foursixnine:
> > I think it would be better to extract that concept of `just install package` elsewhere and reuse in test modules. On the other hand, a reboot in every single test module sounds quite expensive. Isn't the idea of ALP to run the workload in containers? That would apply here in particular for databases
> 
> I think installation of the package is handled in single function `trup_call` and as suggested in the ticket - https://progress.opensuse.org/issues/115196#note-20, I just called the trup_call function in root-console for the transactional update. I will take a note on the reboot after every module installation. As I understood we are doing smoke test of ALP and so we are making changes. @foursixnine, please correct me if I'm wrong

Indeed, the intent has been even shared on the irc channel: https://matrix.to/#/!lCErltlTrMqnzHmKMf:opensuse.org/$4eGIlHWmO6VACZvQpw4sjFOYPw5T2JEU1nGqR7mKs9M?via=libera.chat&via=opensuse.org&via=matrix.org

This will be even more valid after the switch away from TW for containers. As for the boot and reboot of the system, there's little to do/discuss for now

okurz:
> As for the boot and reboot of the system, there's little to do/discuss for now

what do you mean by that?

foursixnine:
> > As for the boot and reboot of the system, there's little to do/discuss for now
> 
> what do you mean by that?

for now we can rely on rebooting the system 2 or 3 times per test, and likely finding a solution for [115196#note-20](https://progress.opensuse.org/issues/115196#note-20) should also help to reduce how expensive transactional systems would be

okurz:
> > > As for the boot and reboot of the system, there's little to do/discuss for now
> > 
> > 
> > what do you mean by that?
> 
> for now we can rely on rebooting the system 2 or 3 times per test, and likely finding a solution for [115196#note-20](https://progress.opensuse.org/issues/115196#note-20) should also help to reduce how expensive transactional systems would be

Yes, we could. But it is quite likely that as soon as code like this is in people add to the mess and the chance to clean it up and do it properly never comes. So I would say do it properly from the beginning to find a concept how to test multiple services on transactional servers though I am arguing that this isn't even required and we should use containers right away from the beginning.

okurz (changes requested):
> > > As for the boot and reboot of the system, there's little to do/discuss for now
> > 
> > 
> > what do you mean by that?
> 
> for now we can rely on rebooting the system 2 or 3 times per test, and likely finding a solution for [115196#note-20](https://progress.opensuse.org/issues/115196#note-20) should also help to reduce how expensive transactional systems would be

Yes, we could. But it is quite likely that as soon as code like this is in people add to the mess and the chance to clean it up and do it properly never comes. So I would say do it properly from the beginning to find a concept how to test multiple services on transactional servers though I am arguing that this isn't even required and we should use containers right away from the beginning.

foursixnine:
@DeepthiYV I think for now, we can close this PR and come back later, once these workloads are already available

PR #15597, opened 2022-09-27
Force tty on ssh commands

grisu48:
Force the usage of tty (`-t`) for ssh commands on publiccloud instances, which is necessary for e.g. sudo and should not have any negative effects on other commands.

- Related ticket: https://progress.opensuse.org/issues/117271
- Verification run: https://duck-norris.qam.suse.de/tests/10862 | https://duck-norris.qam.suse.de/tests/10863

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -139,7 +139,7 @@ sub _prepare_ssh_cmd {
         $cmd = "\$'$cmd'";
     }
 
-    my $ssh_cmd = sprintf('ssh %s "%s@%s" -- %s', $args{ssh_opts}, $args{username}, $self->public_ip, $cmd);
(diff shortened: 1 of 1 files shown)


lib/publiccloud/instance.pm
```diff
@@ -388,7 +388,7 @@ sub softreboot
         ssh_interactive_leave();
     }
 
-    $self->run_ssh_command(cmd => 'sudo shutdown -r +1');
+    $self->ssh_script_run(cmd => 'sudo shutdown -r +1');
```
asmorodskyi, line 391: just to keep things clear - this is unrelated , right ? ( it is fine just for general understanding )

mimi1vx (approved):

lib/publiccloud/instance.pm
grisu48: This is required so that we use the modified `_prepare_ssh_cmd` subroutine

PR #15599, opened 2022-09-27
jeos/firstrun: Handle WiFi on Leap 15.5 as well

Vogtinator:
15.5 also includes jeos-firstboot-rpiwifi now.

Failure: https://openqa.opensuse.org/tests/2741790#step/firstrun/12

- Verification run: https://openqa.opensuse.org/tests/2749361

### DIFF
--- a/tests/jeos/firstrun.pm
+++ b/tests/jeos/firstrun.pm
@@ -124,7 +124,7 @@ sub run {
         send_key 'ret';
     }
 
-    if (is_generalhw && (is_tumbleweed || is_sle || is_sle_micro)) {
+    if (is_generalhw && !is_leap("<15.5")) {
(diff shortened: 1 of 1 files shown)


ggardet (approved):
LGTM. 
Wondering if we could just use `is_leap("<15.5")`?

Vogtinator:
> LGTM. Wondering if we could just use `is_leap("<15.5")`?

For now I was a bit cautious as I'm not entirely sure where it's used, but that should work as well. Should I change it?

ggardet:
> > LGTM. Wondering if we could just use `is_leap("<15.5")`?
> 
> For now I was a bit cautious as I'm not entirely sure where it's used, but that should work as well. Should I change it?

Let's try it.

PR #15601, opened 2022-09-27
Wait for remnant zypper tasks

grisu48:
Wait for remaining zypper tasks before proceeding with bci prepare steps. This is required because sometimes after reboot a `zypp-main` task is still doing some background tasks.

- Related ticket: https://progress.opensuse.org/issues/117286
- Verification run: https://duck-norris.qam.suse.de/tests/10870

### DIFF
--- a/tests/containers/bci_prepare.pm
+++ b/tests/containers/bci_prepare.pm
@@ -67,6 +67,9 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
+    # Wait for any zypper tasks in the background to finish
+    assert_script_run('while pgrep -f zypp; do sleep 1; done', timeout => 300);
+
     # CONTAINER_RUNTIME can be "docker", "podman" or both "podman,docker"
     my $engines = get_required_var('CONTAINER_RUNTIME');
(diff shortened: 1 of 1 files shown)


mloviska (approved):

tests/containers/bci_prepare.pm
```diff
@@ -67,6 +67,9 @@ sub run {
     my ($self) = @_;
     $self->select_serial_terminal;
 
+    # Wait for any zypper tasks in the background to finish
+    assert_script_run('while true; do sleep 1; if ! ps | grep zypp; then break; fi done', timeout => 300);
```
Vogtinator, line 71: Won't work for two reasons:

1. `ps` only prints processes in the current session, you want something like `ps -ef` instead
2. `ps -ef | grep foo` would find itself, leading to an endless loop

I suggest

```suggestion
    assert_script_run('while pgrep -f zypp; do sleep 1; done', timeout => 300);
```

Vogtinator (changes requested):

tests/containers/bci_prepare.pm
grisu48: Thanks, PR amended, restarting verification run.

Vogtinator (approved):

PR #15603, opened 2022-09-27
Implement automation test for nvidia in pubcloud

b10n1k:
Test open source nvidia drivers on GCE. Test jobs require
`PUBLIC_CLOUD_TERRAFORM_FILE=publiccloud/terraform/nvidia_gce.tf` and
`PUBLIC_CLOUD_NVIDIA=1`. With the given terraform the provider enables an
instance with *Nvidia Tesla T4 GPU*. Maybe i should name the tf accordantly as
other GPU can be used according the provider's GPU support

- Related ticket: https://progress.opensuse.org/issues/117316
- Verification run: http://aquarius.suse.cz/tests/13450

### DIFF
--- a/data/publiccloud/terraform/gce.tf
+++ b/data/publiccloud/terraform/gce.tf
@@ -78,6 +78,12 @@ variable "enable_confidential_vm" {
 	default=false
 }
 
+variable "gpu" {
+  description = "Enable and configure node GPUs"
+
+  default = false
+}
+
 resource "random_id" "service" {
     count = var.instance_count
     keepers = {
@@ -91,7 +97,12 @@ resource "google_compute_instance" "openqa" {
     name                         = "${var.name}-${element(random_id.service.*.hex, count.index)}"
     machine_type                 = var.type
     zone                         = var.region
-    
+
+    guest_accelerator {
+      type = "nvidia-tesla-t4"
+      count = var.gpu ? 1 : 0
+    }
+
     confidential_instance_config {
     	enable_confidential_compute = var.enable_confidential_vm ? true : false
     }
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -14,6 +14,7 @@ use main_common qw(loadtest);
 use testapi qw(check_var get_var);
 use Utils::Architectures qw(is_aarch64);
 use main_containers qw(load_container_tests);
+require bmwqemu;
 
 our @EXPORT = qw(
   load_publiccloud_tests
@@ -58,7 +59,11 @@ sub load_maintenance_publiccloud_tests {
             loadtest "publiccloud/az_l8s_nvme" if (get_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'Standard_L(8|16|32|64)s_v2');
         } elsif (get_var('PUBLIC_CLOUD_AZURE_NFS_TEST')) {
             loadtest("publiccloud/azure_nfs", run_args => $args);
+        } elsif (check_var('PUBLIC_CLOUD_NVIDIA', 1)) {
+            die "ConfigError: Either the provider is not supported or SLE version is old!\n" unless (check_var('PUBLIC_CLOUD_PROVIDER', 'GCE') && is_sle('15-SP4+'));
+            loadtest "publiccloud/nvidia", run_args => $args;
         }
+
         loadtest("publiccloud/ssh_interactive_end", run_args => $args);
     }
 }
@@ -82,6 +87,16 @@ sub load_publiccloud_consoletests {
     loadtest 'console/libgcrypt' unless is_sle '=12-SP4';
 }
 
+my $should_use_runargs = sub {
+    my @public_cloud_variables = qw(
+      PUBLIC_CLOUD_CONSOLE_TESTS
+      PUBLIC_CLOUD_CONTAINERS
+      PUBLIC_CLOUD_SMOKETEST
+      PUBLIC_CLOUD_AZURE_NFS_TEST
+      PUBLIC_CLOUD_NVIDIA);
+    return grep { exists $bmwqemu::vars{$_} } @public_cloud_variables;
+};
+
 sub load_latest_publiccloud_tests {
     if (get_var('PUBLIC_CLOUD_IMG_PROOF_TESTS')) {
         loadtest "publiccloud/img_proof";
@@ -101,7 +116,7 @@ sub load_latest_publiccloud_tests {
     elsif (get_var('PUBLIC_CLOUD_FIO')) {
         loadtest 'publiccloud/storage_perf';
     }
-    elsif (get_var('PUBLIC_CLOUD_CONSOLE_TESTS') || get_var('PUBLIC_CLOUD_CONTAINERS') || get_var('PUBLIC_CLOUD_SMOKETEST') || get_var('PUBLIC_CLOUD_AZURE_NFS_TEST')) {
+    elsif (&$should_use_runargs()) {
         my $args = OpenQA::Test::RunArgs->new();
         loadtest "publiccloud/prepare_instance", run_args => $args;
         if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
@@ -113,6 +128,10 @@ sub load_latest_publiccloud_tests {
         loadtest "publiccloud/ssh_interactive_start", run_args => $args;
         if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
             load_publiccloud_consoletests($args);
+        }
+        elsif (check_var('PUBLIC_CLOUD_NVIDIA', 1)) {
+            die "ConfigError: The provider is not supported\n" unless (check_var('PUBLIC_CLOUD_PROVIDER', 'GCE') && is_sle('15-SP4+'));
+            loadtest "publiccloud/nvidia", run_args => $args;
         }
         elsif (get_var('PUBLIC_CLOUD_CONTAINERS')) {
             load_container_tests();
--- a/tests/publiccloud/nvidia.pm
+++ b/tests/publiccloud/nvidia.pm
@@ -1,0 +1,76 @@
+# SUSE's openQA tests
+#
+# Copyright @ SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Opensource Nvidia test
+#          Test the opensourced nvidia drivers (SLE15-SP4+)
+# Maintainer: ybonatakis <ybonatakis@suse.com>
+
+#use Mojo::Base 'consoletest';
+use Mojo::Base 'publiccloud::basetest';
+use registration;
+use testapi;
+use utils;
+use publiccloud::utils;
+
+sub run {
+    my ($self, $args) = @_;
+    $self->{provider} = $args->{my_provider};
+
+    script_run("cat /etc/os-release");
+    zypper_call('--gpg-auto-import-keys addrepo -p 90 ' . get_required_var('NVIDIA_REPO') . ' nvidia_repo');
+    zypper_call '--gpg-auto-import-keys ref';
+    zypper_call("in nvidia-open-gfxG06-kmp-default ", quiet => 1);
+    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 600));
+
+    validate_script_output("hwinfo --gfxcard", sub { /nVidia.*Tesla T4/mg });    # depends on terraform setup
+    assert_script_run("LD_LIBRARY_PATH=/usr/lib/kernel-firmware-nvidia-gsp /usr/lib/kernel-firmware-nvidia-gsp/nvidia-smi --query");
+    assert_script_run("SUSEConnect --status-text", 300);
+}
+
+1;
+
+=head1 Discussion
+
+Test module to run Opensource Nvidia test on publiccloud with Turing or Ampere architecture GPUs.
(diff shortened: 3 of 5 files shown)


grisu48 (commented):
Congrats @b10n1k to this new test run! I've got two minor comments, overall nice work!

Also it's nice to see that someone else is going to make good use of the `PUBLIC_CLOUD_TERRAFORM_FILE` variable :slightly_smiling_face:

data/publiccloud/terraform/nvidia_gce.tf
```diff
@@ -0,0 +1,173 @@
...
+
+resource "google_compute_instance" "openqa" {
+    count                        = var.instance_count
+    name                         = "${var.name}-${element(random_id.service.*.hex, count.index)}"
+    machine_type                 = var.type
+    zone                         = var.region
+
+    guest_accelerator {
+      type = "nvidia-tesla-t4"
+      count = 1
```
asmorodskyi, line 97: I did a diff of this file in comparison to "regular" gce.tf . this guest_accelerator is the only block which differs . I don't think it make sense to keep it as separate file . 

Just add this module into `gce.tf` with this change : 

```suggestion
      count = var.guest_accelerator ? 1 : 0
```

also add variable : 
```
variable "guest_accelerator" {
    default=false
}
```

lib/main_publiccloud.pm
```diff
@@ -58,7 +58,10 @@ sub load_maintenance_publiccloud_tests {
             loadtest "publiccloud/az_l8s_nvme" if (get_var('PUBLIC_CLOUD_INSTANCE_TYPE') =~ 'Standard_L(8|16|32|64)s_v2');
         } elsif (get_var('PUBLIC_CLOUD_AZURE_NFS_TEST')) {
             loadtest("publiccloud/azure_nfs", run_args => $args);
+        } elsif (check_var('PUBLIC_CLOUD_NVIDIA', 1)) {
+            loadtest "publiccloud/nvidia", run_args => $args if (check_var('PUBLIC_CLOUD_PROVIDER', 'GCE') && is_sle('15-SP4+'));
```
asmorodskyi, line 62: so if you will define variable PUBLIC_CLOUD_NVIDIA but will set PUBLIC_CLOUD_PROVIDER=Azure it will boot create instance and than simply shutdown ? I would suggest to move this checks into module and fail test there

```suggestion
            loadtest "publiccloud/nvidia", run_args => $args;
```

asmorodskyi:
```
elsif (get_var('PUBLIC_CLOUD_CONSOLE_TESTS') || get_var('PUBLIC_CLOUD_CONTAINERS') || get_var('PUBLIC_CLOUD_SMOKETEST') || get_var('PUBLIC_CLOUD_AZURE_NFS_TEST')) {
```

this is from `load_latest_publiccloud_tests` ( see line 107 in main_publiccloud.pm )  it wraps around your Nvidia block , you need to tweak it to be able to run tests on latest . You would notice this if you will do VR for latest tests . I think it is needed here

tests/publiccloud/nvidia.pm
```diff
@@ -0,0 +1,63 @@
+# SUSE's openQA tests
+#
+# Copyright @ SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Opensource Nvidia test
+#          Drivers are only supported on SLE15SP4+
+# Maintainer: ybonatakis <ybonatakis@suse.com>
+
+use Mojo::Base 'consoletest';
```
asmorodskyi, line 10:
```suggestion
use Mojo::Base 'publiccloud::basetest';
```

this way you will get auto-cleanup logic

tests/publiccloud/nvidia.pm
```diff
@@ -0,0 +1,63 @@
...
+use registration;
+use testapi;
+use utils;
+use publiccloud::utils;
+
+sub run {
+    my ($self, $args) = @_;
+    $self->{provider} = $args->{my_provider};
+
+    script_run("hostname -f");
```
asmorodskyi, line 20: is this necessary ?

tests/publiccloud/nvidia.pm
```diff
@@ -0,0 +1,63 @@
...
+    script_run("cat /etc/os-release");
+    # https://download.opensuse.org/repositories/X11:/Drivers:/Video/SLE_15_SP4/ X11:Drivers:Video
```
asmorodskyi, line 22: why we need this comment ?

lib/main_publiccloud.pm
b10n1k: No. First test is not designed or tested with Azure. And for now i think it doesnt matter (but maybe i will need to confirm). The think is that no all providers provide the GPU. The `is_sle('15-SP4+')` is required as the driver is not supported in older versions.

data/publiccloud/terraform/nvidia_gce.tf
b10n1k: i need to test this. i will come back to you

lib/main_publiccloud.pm
asmorodskyi: yes I know that it is not designed for Azure ( or Amazon )  . and main point of this `if` to handle the case when someone mistakenly try to run this test against Azure (or Amazon ) if I get it right . And my point here that I don't like how you handle such case . In such case you will silently skip one module and that's it . I think better to : 
1. Not do such checks in main_publiccloud.pm
2. Move checks **inside** nvidia.pm
3. not just skip but die

tests/publiccloud/nvidia.pm
b10n1k: We dont. It's leftover

tests/publiccloud/nvidia.pm
b10n1k: No really. leftover from the initial tests i was running. i will remove it

lib/main_publiccloud.pm
b10n1k: After offline discussion, i update the code with the following approach. Make the test die during the scheduling. With this we prevent any execution with the wrong requirements. http://aquarius.suse.cz/tests/13481

image

lib/main_publiccloud.pm
```diff
@@ -114,6 +117,9 @@ sub load_latest_publiccloud_tests {
         if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
             load_publiccloud_consoletests($args);
         }
+        elsif (check_var('PUBLIC_CLOUD_NVIDIA', 1)) {
+            loadtest "publiccloud/nvidia", run_args => $args if (check_var('PUBLIC_CLOUD_PROVIDER', 'GCE') && is_sle('15-SP4+'));
```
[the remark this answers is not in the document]
b10n1k: same here as https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15603#discussion_r1000526515

tests/publiccloud/nvidia.pm
b10n1k: *ready for review*

tests/publiccloud/nvidia.pm
b10n1k: *ready for review*

lib/main_publiccloud.pm
b10n1k: *ready for review*

lib/main_publiccloud.pm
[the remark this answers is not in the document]
b10n1k: *ready for review*

data/publiccloud/terraform/nvidia_gce.tf
b10n1k: *ready for review*

tests/publiccloud/nvidia.pm
```diff
@@ -0,0 +1,63 @@
+# SUSE's openQA tests
+#
+# Copyright @ SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Opensource Nvidia test
+#          Drivers are only supported on SLE15SP4+
```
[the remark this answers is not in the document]
b10n1k: *ready for review*

data/publiccloud/terraform/gce.tf
```diff
@@ -78,6 +78,16 @@ variable "enable_confidential_vm" {
 	default=false
 }
 
+variable "gpu" {
```
asmorodskyi, line 81: If you will just tweak it to :
```
guest_accelerator {
      type = "nvidia-tesla-t4"
      count = var.gpu ? 1 : 0
}
``` 

and 

```
variable "gpu" {
 default=false
}
```
it will gives you same result but will be much easy to read and understand what is going on

asmorodskyi: you don't need to reset type in case count is 0

lib/main_publiccloud.pm
```diff
@@ -114,6 +129,10 @@ sub load_latest_publiccloud_tests {
         if (get_var('PUBLIC_CLOUD_CONSOLE_TESTS')) {
             load_publiccloud_consoletests($args);
         }
+        elsif (check_var('PUBLIC_CLOUD_NVIDIA', 1)) {
+            die "ConfigError: The provider is not supported\n" unless (check_var('PUBLIC_CLOUD_PROVIDER', 'GCE'));
```
asmorodskyi, line 133: no one can stop you from loading latest tests with SLE 12 SP5 :)  yes normally it should not happen but this is not how we decide if we need to check for something or not

data/publiccloud/terraform/gce.tf
b10n1k: > If you will just tweak it to :
> 
> ```
> guest_accelerator {
>       type = "nvidia-tesla-t4"
>       count = var.gpu ? 1 : 0
> }
> ```
> 
it didnt work. As is, it is a much better approach IMO.

> and
> 
> ```
> variable "gpu" {
>  default=false
> }
> ```
> 
> it will gives you same result but will be much easy to read and understand what is going on

lib/main_publiccloud.pm
b10n1k: Please remind me. `load_latest_publiccloud_tests` vs `SLE 12 SP5 `, dont see the connection. My guess is that you ask me to include version check! am i right?

asmorodskyi: yeah I am saying that check should be the same as for maintenance tests

data/publiccloud/terraform/gce.tf
asmorodskyi: looks like we have two different discussions here : 
1. It didn't work - can you point me to failed job where you trying to use this approach so I can see the problem ?
2. Can you elaborate what is "better" means to you here ? You don't win anything if type would be null or have some value when you not using this section at all

asmorodskyi: also please provide a VR of your code when gpu is not set , I have doubts about 
```
        for_each = var.gpu[*]
            content {
                type  = guest_accelerator.value.type
                count = guest_accelerator.value.count
            }
```

when gpu is null that it won't throw exception ...

b10n1k: > looks like we have two different discussions here :
> 
> 1. It didn't work - can you point me to failed job where you trying to use this approach so I can see the problem ?

http://aquarius.suse.cz/tests/13543#step/prepare_instance/71

> 2. Can you elaborate what is "better" means to you here ? You don't win anything if type would be null or have some value when you not using this section at all

IIUC with `null` as a default, terraform ignores the resource. if resource's value is not `null` then it uses it. In the other had a boolean `false` needs that conditional logic.

b10n1k: > also please provide a VR of your code when gpu is not set , I have doubts about
> 
> ```
>         for_each = var.gpu[*]
>             content {
>                 type  = guest_accelerator.value.type
>                 count = guest_accelerator.value.count
>             }
> ```
> 
> when gpu is null that it won't throw exception ...

http://aquarius.suse.cz/tests/13490#step/prepare_instance/56

asmorodskyi: > > looks like we have two different discussions here :
> > 
> > 1. It didn't work - can you point me to failed job where you trying to use this approach so I can see the problem ?
> 
> http://aquarius.suse.cz/tests/13543#step/prepare_instance/71

`-var 'gpu={"count":1,"type":"nvidia-tesla-t4"}'`  it is not what you suppose to pass when  

```
guest_accelerator {
      type = "nvidia-tesla-t4"
      count = var.gpu ? 1 : 0
}
```
you need to pass true/false than everything will work as expected !  So this run can not prove that "it didn't work" because you defined boolean variable and trying to pass dict there 

> IIUC with `null` as a default, terraform ignores the resource. if resource's value is not `null` then it uses it. In the other had a boolean `false` needs that conditional logic.
```
guest_accelerator {
      type = "nvidia-tesla-t4"
      count = 0
}
```

this WILL be ignored no matter that type has value

lib/main_publiccloud.pm
b10n1k: I am not sure if this is even a think here but i add the version check. *ready for review*

data/publiccloud/terraform/gce.tf
b10n1k: *ready for review*

b10n1k: for the record, now the disadvantage is that the type is hardcoded. My previous approach it was given the agility to keep the terraform file the same and pass the variable as you like.

asmorodskyi (approved):
LGTM

PR #15608, opened 2022-09-28
xfstests: use alt-sysrq-b when system got hang

SeroSun:
As before when reboot not work we change to use reset, but most of case a reset still can't reboot successful. Here change to use magic sysrq key to a force reboot.

- Verification run: https://openqa.suse.de/t9622898 reboot successful here(https://openqa.suse.de/tests/9622898#step/run/1961), and log file upload successfully in log page(https://openqa.suse.de/tests/9622898/file/run-opt_logs.tar.gz)
- Compare with previous result https://openqa.suse.de/tests/9565012#step/run/1964 when btrfs/262 hang the system
- Add one more line with fatal => 0, now it could show logs when crash happens. https://openqa.suse.de/tests/9627636

### DIFF
--- a/tests/xfstests/run.pm
+++ b/tests/xfstests/run.pm
@@ -25,7 +25,7 @@ use File::Basename;
 use testapi;
 use utils;
 use Utils::Backends 'is_pvm';
-use power_action_utils 'power_action';
+use power_action_utils qw(power_action prepare_system_shutdown);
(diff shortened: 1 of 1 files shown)


lansuse:
LGTM

schlad (approved):

czerw (approved):

PR #15619, opened 2022-09-29
qesap-py, allow appending optional arguments

lpalovsky:
qesapdeployment.pm does not currently accept additional arguments for 
commands used by qesap.py. This PR adds additional "cmd_options" into 
qesapdeployment/qesap_execute sub and appends it to the final command.

- Needles: N/A
- Verification run:
  GCP: https://mordor.suse.cz/tests/4771
  Azure: https://mordor.suse.cz/tests/4772#

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -337,7 +337,11 @@ sub qesap_sh_destroy {
 
 =head3 qesap_execute
 
-    qesap_execute(cmd => $qesap_script_cmd [, verbose => 1] );
+    qesap_execute(cmd => $qesap_script_cmd [, verbose => 1, cmd_options => $cmd_options] );
+    cmd_options - allows to append additional qesap.py commans arguments like "qesap.py terraform -d"
+        Example:
+        qesap_execute(cmd => 'terraform', cmd => '-d') will result in:
+        qesap.py terraform -d
 
     Execute qesap glue script commands. Check project documentation for available options:
     https://github.com/SUSE/qe-sap-deployment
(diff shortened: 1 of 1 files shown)


lib/qesapdeployment.pm
```diff
@@ -349,19 +350,20 @@ sub qesap_execute {
...
-    else {
-        assert_script_run(join(" ", $qesap_cmd, $args{cmd}, "2>&1 | tee -a", $exec_log));
-    }
+    my $qesap_cmd = join(" ", $paths{deployment_dir} . "/scripts/qesap/qesap.py",
+        $verbose,
+        "-c", $paths{qesap_conf_trgt},
+        "-b", $paths{deployment_dir},
+        $args{cmd},
+        $args{cmd_options},
+        "2>&1 | tee -a",
```
dzedro, line 360: Just tip, `|&` can be used instead `2>&1 |`

dzedro (approved):

lib/qesapdeployment.pm
```diff
@@ -337,8 +337,9 @@ sub qesap_sh_destroy {
 
 =head3 qesap_execute
 
-    qesap_execute(cmd => $qesap_script_cmd [, verbose => 1] );
+    qesap_execute(cmd => $qesap_script_cmd [, verbose => 1, cmd_options => $cmd_options] );
 
+    cmd_options - allows to append additional qesap.py commans arguments like "qesap.py terraform -d"
```
mpagot, line 342: Does could be the explanation and the example like

> ... like : 
```
qesap_execute(cmd => 'terraform', cmd => '-d')
```
result in 
```
qesap.py terraform -d
```

mpagot (approved):
LGTM

lib/qesapdeployment.pm
lpalovsky: Cool it works:
https://mordor.suse.cz/tests/4935/#
Test failed but command was executed correctly:
```
# /root/qe_sap_deployment/scripts/qesap/qesap.py --verbose -c /root/qe_sap_deployment/scripts/qesap/hanasr_azure.yaml -b /root/qe_sap_deployment ansible  |& tee -a /tmp/qesap_exec_ansible.log.txt; echo RMEEv-$?-
```

lib/qesapdeployment.pm
lpalovsky: Yep. That is a good idea. 
Fixed and pushed.

PR #15622, opened 2022-09-30
Move generic k8s tools to k8s library from k8sbasetest

ilausuch:
In order to use these tools without having a test with k8sbasetest as base. The best option is to move these tools to k8s library

- Related ticket: https://progress.opensuse.org/issues/117004
- Verification run: https://openqa.suse.de/tests/9646619

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -14,13 +14,17 @@ use Exporter;
 use strict;
 use warnings;
 use testapi;
-use utils qw(zypper_call script_retry file_content_replace validate_script_output_retry);
+use utils qw(zypper_call script_retry file_content_replace validate_script_output_retry random_string);
 use Utils::Systemd qw(systemctl);
 use containers::utils 'registry_url';
 use version_utils qw(is_sle);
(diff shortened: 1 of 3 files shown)


lib/containers/k8s.pm
```diff
@@ -92,4 +112,47 @@ sub install_oc {
...
+    upload_logs($path, failok => 1);
+
+    assert_script_run("kubectl apply -f $path");
+}
+
+=head2 find_pods
+Find pods using kubectl queries
+=cut
+
+sub wait_for_job_complete {
```
mloviska, line 135: I guess, it would be less ambiguous to express that the function waits for a `kube` job/pod to finished

ilausuch: good point, but it is inside k8s lib, is necessary to specify? wait_for_k8s_job_complete, for instance?

mloviska: I think that it would be handy, mainly for people that are often `greping` test repo for function names as me. :D

ilausuch: Fixed then

PR #15630, opened 2022-10-03
Fix redis-cli get foo nil output

pdostal:
As the serial console is handled differently on svirt backends it is not possible to check for `nil`.

- Related ticket: [poo#116773](https://progress.opensuse.org/issues/116773)
- Previous attempt: #15592
- Verification run: [svirt](https://openqa.suse.de/tests/9645893#step/redis/15), [qemu](https://openqa.suse.de/tests/9645894#step/redis/18)

### DIFF
--- a/tests/console/redis.pm
+++ b/tests/console/redis.pm
@@ -37,7 +37,7 @@ sub run {
     validate_script_output('redis-cli get foo', sub { m/bar/ });
     validate_script_output('redis-cli pfselftest', sub { m/OK/ });
     validate_script_output('redis-cli flushdb', sub { m/OK/ });
-    validate_script_output('redis-cli get foo', sub { m/(nil)/ });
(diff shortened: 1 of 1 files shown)


tests/console/redis.pm
```diff
@@ -37,7 +37,7 @@ sub run {
     validate_script_output('redis-cli get foo', sub { m/bar/ });
     validate_script_output('redis-cli pfselftest', sub { m/OK/ });
     validate_script_output('redis-cli flushdb', sub { m/OK/ });
-    validate_script_output('redis-cli get foo', sub { m/(nil)/ });
+    validate_script_output('redis-cli get foo', sub { !m/bar/ });
```
foursixnine, line 40: I'm not a fan of relying on differences in behavior, but I guess it isn't the end of the world :)

foursixnine (approved):

PR #15631, opened 2022-10-03
yast2_users: send enter before clearing console

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/117163
- Verification run: https://openqa.suse.de/tests/9646051


jouyingbin (approved):
LGTM, and send 1 "ret" is not working sometimes, right?

paolostivanin:
yep! I found it to be more stable this way. Anyway, this workaround will be removed once nokogiri will be rebuilt.

PR #15638, opened 2022-10-04
Expand kernel mismatch message

BillAnastasiadis:
Expands the error message when the kernel doesn't match the kernel-default-devel version, which usually happens briefly when there's a new kernel, to better inform testers.

- Related ticket: https://progress.opensuse.org/issues/116860
- Verification run: (tbd - but it's just a string expansion, so everything is expected to be ok)


Avinesh (approved):
looks fine!

PR #15643, opened 2022-10-05
WIP: Re-establish ssh tunnel after instance reboot

grisu48:
Re-establishes the ssh tunnel after a publiccloud instance is being rebooted using the instance->softreboot method. This is necessary for the slem on publiccloud test runs, where we need to do many reboots due to the transactional updates.

This commit partly enabled SLEM test runs on Publiccloud by allowing transactional-update to reboot the instance. It does not address other issues for SLEM on Publiccloud, which will be addressed later.

- Verification runs: TBD


tests/publiccloud/registration.pm
```diff
@@ -25,11 +25,20 @@ sub run {
...
     select_host_console();    # select console on the host, not the PC instance
 
-    registercloudguest($args->{my_instance});
-    register_addons_in_pc($args->{my_instance});
+
+    if (is_slem_on_pc) {
+        # SLEM on PublicCloud is still using SUSEConnect
+        $instance->ssh_assert_script_run("sudo SUSEConnect -r " . get_required_var('SCC_REGCODE'));
+        $instance->ssh_assert_script_run("sudo zypper lr");
+        return;
```
pdostal, line 37:
```suggestion
```

tests/publiccloud/prepare_tools.pm
```diff
@@ -7,14 +7,14 @@
...
-# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>, qa-c team <qa-c@suse.de>
+# Maintainer: qa-c team <qa-c@suse.de>
 
 use base "opensusebasetest";
 use strict;
 use warnings;
 use testapi;
 use utils;
-use registration 'add_suseconnect_product';
+use registration;
```
pdostal, line 17: Why? Listing the subroutines helps keeping track of them.

tests/publiccloud/prepare_tools.pm
```diff
@@ -71,7 +71,8 @@ sub run {
     ensure_ca_certificates_suse_installed();
 
     # Install prerequesite packages test
-    zypper_call('-q in python39-pip python39-devel python3-img-proof python3-img-proof-tests podman docker jq rsync');
+    add_suseconnect_product(get_addon_fullname('phub')) if is_sle();
+    zypper_call('-q in python39-pip python39-devel python3-img-proof python3-img-proof-tests podman docker jq rsync autossh');
```
pdostal, line 75: Maybe off topic but why do we use both `podman` and `docker` at the same time?

lib/transactional.pm
```diff
@@ -76,6 +78,17 @@ sub process_reboot {
 
     handle_first_grub if ($args{automated_rollback});
 
+    # On tunneled instances we need to reboot the attached guest, not the helper VM
+    if (get_var("TUNNELED")) {
+        if (get_var("PUBLIC_CLOUD")) {
```
pdostal, line 83:
```suggestion
        if (version_utils::is_public_cloud()) {
```

lib/transactional.pm
```diff
@@ -76,6 +78,17 @@ sub process_reboot {
 
     handle_first_grub if ($args{automated_rollback});
 
+    # On tunneled instances we need to reboot the attached guest, not the helper VM
+    if (get_var("TUNNELED")) {
```
pdostal, line 82: Didn't we have subroutine for this? (If not, maybe it will make sense?)

lib/transactional.pm
```diff
@@ -156,14 +169,19 @@ sub trup_call {
     my ($cmd, %args) = @_;
     $args{timeout} //= 180;
     $args{exit_code} //= 0;
+    my $tunneled = get_var("TUNNELED") && testapi::is_serial_terminal();    # Don't pipe stuff to serialterminal in TUNNELED mode
```
pdostal, line 172: Why? It should work.

lib/publiccloud/utils.pm
```diff
@@ -176,6 +177,10 @@ sub is_container_host() {
     return is_public_cloud && get_var('FLAVOR') =~ 'CHOST';
 }
 
+sub is_slem_on_pc() {
+    return is_public_cloud && get_var('FLAVOR') =~ 'Micro';
+}
+
```
pdostal, line 183: I think we use this only in public cloud scenarios anyways, so maybe `version_utils::is_sle_micro()` would be enough?

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,93 @@
...
 use base Exporter;
 use testapi;
 use Utils::Backends qw(set_sshserial_dev unset_sshserial_dev);
 use strict;
 use warnings;
+use distribution;
 
 our @EXPORT = qw(ssh_interactive_tunnel ssh_interactive_leave select_host_console);
 
+sub active_console {
```
pdostal, line 21:
```suggestion
sub activate_console {
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,93 @@
...
-        cmd => "'rm -rf /dev/sshserial; mkfifo -m a=rwx /dev/sshserial; tail -fn +1 /dev/sshserial' 2>&1 | tee /dev/$serialdev; clear",
-        timeout => 0,
-        no_quote => 1,
-        ssh_opts => "-yt -R $upload_port:$upload_host:$upload_port",
-        username => 'root',
-    );
-    sleep 3;
-    save_screenshot;
+    # Pipe the output of the device fifo to the local serial terminal
+    background_script_run("while true; do ssh sut 'rm -f /dev/sshserial && mkfifo -m a=rwx /dev/sshserial && tail -fn +1 /dev/sshserial' 2>&1 >/dev/$serialdev; sleep 5; done");
```
pdostal, line 54: You're missing the port forwarding.

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,93 @@
...
+    # give the ssh connection some time to settle
+    sleep 10;
```
pdostal, line 56:
```suggestion
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,93 @@
...
-    script_run("test -p /dev/sshserial && exit", timeout => 0);
+    my $prev_console = current_console();
+    select_console('tunnel-console') unless ($prev_console =~ /tunnel-console/);
+
+    # Because the terminal is busy, it tends to swallow some keypresses.
+    # send the termination signal multiple times to ensure it goes through.
+    enter_cmd('kill %1');    # Don't use script_run here, because the serial device is still set to /dev/sshserial
+    enter_cmd('kill %1');
+    enter_cmd('kill %1');
+    save_screenshot;    # Keep for some time until we are sure, this works properly
```
pdostal, line 84: This does make any sense in serial terminal.

lib/publiccloud/instances.pm
```diff
@@ -0,0 +1,26 @@
...
+# Summary: Base class for the public cloud namespace
+#
+# Maintainer: qa-c team <qa-c@suse.de>
+
+package publiccloud::instances;
+use testapi;
+use strict;
+use warnings;
+
+our @instances;    # Package variable containing all instanciated instances for global access without RunArgs
```
pdostal, line 15: WOW this sounds super cool!

pdostal (commented):
Removed my review as Felix asked for it.

grisu48:
Starting a set of extensive test runs. It might take some time and changes until this is ready for review

grisu48:
After some fixes, I'm starting now a bunch of test runs.

lib/publiccloud/instance.pm
```diff
@@ -379,12 +380,11 @@ sub softreboot
 
     my $duration;
 
-    # On TUNNELED test runs, ensure we are not on the publiccloud instance
     my $prev_console = current_console();
-    # We only need to re-establish the ssh tunnel, if we are in a TUNNELED test run, and if the tunnel is already initialized
-    my $tunneled = get_var('TUNNELED', 0) && get_var("_SSH_TUNNELS_INITIALIZED", 0);
+    # On TUNNELED test runs, we need to re-establish the tunnel
+    my $tunneled = get_var('TUNNELED', 0) && check_var("_SSH_TUNNELS_INITIALIZED", 1);
```
pdostal, line 385: What about creating `is_tunneled()`?

lib/publiccloud/provider.pm
```diff
@@ -490,6 +491,7 @@ sub terraform_apply {
     }
 
     # Return an ARRAY of objects 'instance'
+    publiccloud::instances::set_instances(@instances);
```
pdostal, line 494: Just export the subroutines.

```suggestion
    set_instances(@instances);
```

pdostal: Maybe @cfconrad could be interested in this.

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
 #
-# Maintainer: Pavel Dostal <pdostal@suse.cz>
+# Maintainer: qa-c@suse.de
 
 package publiccloud::ssh_interactive;
+use testapi qw(is_serial_terminal);
 use base Exporter;
 use testapi;
 use Utils::Backends qw(set_sshserial_dev unset_sshserial_dev);
+use utils qw(type_string_very_slow);
```
pdostal, line 15: This is not used over here.
```suggestion
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
 use strict;
 use warnings;
+use distribution;
```
pdostal, line 18: Please list the subroutines so one know what and from which module is used over here
```suggestion
use distribution qw(set_sshserial_dev establish_tunnel_console);
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
-        cmd => "'rm -rf /dev/sshserial; mkfifo -m a=rwx /dev/sshserial; tail -fn +1 /dev/sshserial' 2>&1 | tee /dev/$serialdev; clear",
-        timeout => 0,
-        no_quote => 1,
-        ssh_opts => "-yt -R $upload_port:$upload_host:$upload_port",
-        username => 'root',
-    );
-    sleep 3;
-    save_screenshot;
+    # Pipe the output of the device fifo to the local serial terminal
+    background_script_run("while true; do ssh sut -R '$upload_port:$upload_host:$upload_port' 'rm -f /dev/sshserial && mkfifo -m a=rwx /dev/sshserial && tail -fn +1 /dev/sshserial' 2>&1 >/dev/$serialdev; sleep 5; done");
```
pdostal, line 56: You forgot about `-yt`
```
$ man ssh
-y  Send log information using the syslog(3) system module.  By default this information is sent to stderr.
-t  Force pseudo-terminal allocation.  This can be used to execute arbitrary screen-based programs on a remote machine, which can be very useful, e.g. when implementing menu services.
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
+    # give the ssh connection some time to settle
+    sleep 10;
```
pdostal, line 58:
```suggestion
```
Why this is needed? It was not before. Let's not use `sleep` unless absolutely needed.

lib/transactional.pm
pdostal, line 83:
```suggestion
        if (is_public_cloud) {
```

lib/transactional.pm
```diff
@@ -156,14 +169,19 @@ sub trup_call {
...
 
     # Always wait for rollback.service to be finished before triggering manually transactional-update
     ensure_rollback_service_not_running();
 
-    my $script = "transactional-update $cmd > /dev/$serialdev";
+    my $script = "transactional-update $cmd";
+    $script .= " > /dev/$serialdev" unless ($tunneled);
```
pdostal, line 178: This should work in tunnelled mode as in tunnelled mode the `/dev/$serialdev` should point to `/dev/sshserial`

tests/publiccloud/prepare_tools.pm
```diff
@@ -133,6 +132,10 @@ EOT
     # Remove persistent net rules, necessary to boot the x86_64 image in the aarch64 test runs
     assert_script_run('rm /etc/udev/rules.d/70-persistent-net.rules');
 
+    # Add marker file for PC tools image
+    my $PUBLISH_HDD_1 = get_var("PUBLISH_HDD_1", "");
+    assert_script_run("echo -e 'PC tools image\\nHDD: $PUBLISH_HDD_1' > /root/pc_tools_image.txt");
```
pdostal, line 137: What is this for?

tests/publiccloud/ssh_interactive_start.pm
```diff
@@ -19,27 +19,14 @@ sub run {
...
-            #   it doesn't interfere with tests as it isn't piped to /dev/sshserial
-            script_run('ssh -E /var/tmp/ssh_sut.log -vt sut', timeout => 0);
-        }
-    }
+    # Initialize ssh tunnel for the serial device, if not yet happened
+    ssh_interactive_tunnel($args->{my_instance}) if (get_var('_SSH_TUNNELS_INITIALIZED', 0) == 0);
     die("expect ssh serial") unless (get_var('SERIALDEV') =~ /ssh/);
+    # The serial terminal needs to be activated manually, as it requires the $self argument
+    $self->select_serial_terminal();
+    enter_cmd('ssh -t sut');
```
pdostal, line 27: The `select_serial_terminal` subroutine does not required `$self`.

pdostal (commented):
This is great work! I love the `lib/publiccloud/instances.pm` idea which should allow us in the future to run out of `run_args`!

lib/publiccloud/instance.pm
```diff
@@ -1,11 +1,11 @@
 # SUSE's openQA tests
 #
-# Copyright 2018 SUSE LLC
+# Copyright 2022 SUSE LLC
```
asmorodskyi, line 3:
```suggestion
# Copyright 2018-2022 SUSE LLC
```

lib/publiccloud/instance.pm
asmorodskyi, line 385: does not make a big difference , just asking out of curiosity so don't take this as request to change something : 
why you replace only one get_var with check_var ? for me they have exactly same reasons to stay as get_var and exactly same reasons to be replaced with check_var ?

lib/publiccloud/instances.pm
```diff
@@ -0,0 +1,26 @@
...
+
+sub set_instances {
+    @instances = @_;
+}
+
+sub get_instance {
+    return undef if (scalar @instances) < 1;
+    return $instances[0];
```
asmorodskyi, line 23: this is a bit strange to me . We saying "ok we have collection of something" but getter method will give you only first item . Generally speaking I treat everything regarding "collection of instances" as dead code because we actually don't have cases where we create more than one entity . And I wish to drop this for a long time . But if you decided to define dedicated collection of instance getter suppose to give access to any element in collection

lib/publiccloud/provider.pm
```diff
@@ -490,6 +491,7 @@ sub terraform_apply {
     }
 
     # Return an ARRAY of objects 'instance'
```
asmorodskyi, line 493: comment should stay above return . Or you may just delete it because it is stating the obvious so does not hold any real information ...

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2022 SUSE LLC
```
asmorodskyi, line 3:
```suggestion
# Copyright 2019-2022 SUSE LLC
```

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
 use Utils::Backends qw(set_sshserial_dev unset_sshserial_dev);
+use utils qw(type_string_very_slow);
 use strict;
 use warnings;
+use distribution;
 
 our @EXPORT = qw(ssh_interactive_tunnel ssh_interactive_leave select_host_console);
 
+sub establish_tunnel_console {
+    # Helper call to activate a console and establish the ssh connection therein
```
asmorodskyi, line 23: looks like comment about whole functions so maybe we need to move it one line above ?

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
+    my $console = $_[0];
```
asmorodskyi, line 24:
```suggestion
    my ($console) = @_;
```
I am really bad with all this special perl vars , but isn't that equal to what I wrote ? If yes I would prefer this more common version which is used commonly in this repo

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
+
+    select_console("$console");
+    # Note: Don't use script_run here! The serial terminal is set to /dev/sshserial, so every script_run will time out
+    type_string("\n~.\n", max_interval => 1);    # ensure no previous ssh connection is present
+    enter_cmd("clear");
+    enter_cmd('ssh -t sut');
+    # give the ssh connection some time to settle
+    sleep 5;
```
asmorodskyi, line 32: maybe from the beginning we will move this into var ? Something telling me that move than one commit in future will tweak this line :)

tests/publiccloud/prepare_tools.pm
```diff
@@ -133,6 +132,10 @@ EOT
     # Remove persistent net rules, necessary to boot the x86_64 image in the aarch64 test runs
     assert_script_run('rm /etc/udev/rules.d/70-persistent-net.rules');
 
+    # Add marker file for PC tools image
+    my $PUBLISH_HDD_1 = get_var("PUBLISH_HDD_1", "");
```
asmorodskyi, line 136: I would go for `get_required_var` here . if `prepare_tools` running in job which does not have `PUBLISH_HDD_1` var something going wrong

tests/publiccloud/ssh_interactive_start.pm
```diff
@@ -1,12 +1,12 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2022 SUSE LLC
```
asmorodskyi, line 3:
```suggestion
# Copyright  2019-2022 SUSE LLC
```

asmorodskyi:
this looks pretty huge , I haven't understand all changes so I think we will need "live review" session for this

pdostal:
> this looks pretty huge , I haven't understand all changes so I think we will need "live review" session for this

Can I also attend please?

lib/publiccloud/ssh_interactive.pm
grisu48: As stated in the comment above, this is helpful to let ssh establish the connection to the guest system.
If we don't do this, it might happen that the following commands are being typed in too hastily.

Do you have a better idea on how to synchronize this? We need to wait for the ssh connection to be established and I didn't came up with an elegant way of doing this.

tests/publiccloud/ssh_interactive_start.pm
grisu48: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/opensusebasetest.pm#L1295

I might not require the argument right now, but the function signature is meant to receive `$self` and that's also how it is used in most test cases.

I might not be needed in the current implementation, but relying on that assumption is dangerous.

lib/publiccloud/instance.pm
grisu48: ok

lib/publiccloud/ssh_interactive.pm
grisu48: ok

lib/transactional.pm
grisu48: This change should be not part of this PR. I'll remove it, this is not intended to be here.

lib/publiccloud/instance.pm
grisu48: This probably happened while development, no particular reason and I'll change it back.

lib/publiccloud/instances.pm
grisu48: I integrated this singleton variable in the current workflow, which allows for multiple instances. Don't blame this PR, it simply moves the existing variable handling into a global singleton that can be used also from other tests.

A refactoring of the underlying handling is IMHO way beyond the scope of this PR.

tests/publiccloud/prepare_tools.pm
grisu48: On test runs I don't set this variable, and a failure of this kind at the end of the test run is more annoying than helpful. I'd say this check is not really useful, but if you want to have it, then it should be at the beginning of the test module, not at the end.

lib/publiccloud/provider.pm
asmorodskyi: I did , you insert `publiccloud::instances::set_instances(@instances);` between comment and the line which it belongs too . So from now on reader will assume that comment corresponds to `publiccloud::instances::set_instances(@instances);` which does not make any sense

lib/publiccloud/instances.pm
asmorodskyi: ok

tests/publiccloud/prepare_tools.pm
asmorodskyi: yes please move this to the beginning of the module

grisu48: done.

lib/publiccloud/provider.pm
cfconrad: Thx for pointing. Interesting point to use singleton here for test data!

lib/publiccloud/ssh_interactive.pm
pdostal: How about we remove it? You may assert the prompt from the remote server instead.

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,97 @@
...
-        cmd => "'rm -rf /dev/sshserial; mkfifo -m a=rwx /dev/sshserial; tail -fn +1 /dev/sshserial' 2>&1 | tee /dev/$serialdev; clear",
-        timeout => 0,
-        no_quote => 1,
-        ssh_opts => "-yt -R $upload_port:$upload_host:$upload_port",
-        username => 'root',
-    );
-    sleep 3;
-    save_screenshot;
+    # Pipe the output of the device fifo to the local serial terminal
+    background_script_run("while true; do ssh sut -yt -R '$upload_port:$upload_host:$upload_port' 'rm -f /dev/sshserial && mkfifo -m a=rwx /dev/sshserial && tail -fn +1 /dev/sshserial' 2>&1 >/dev/$serialdev; sleep 5; done");
```
pdostal, line 55:
```suggestion
    background_script_run("while true; do ssh sut -yt -R '$upload_port:$upload_host:$upload_port' 'rm -f /dev/sshserial && mkfifo -m a=rwx /dev/sshserial && tail -fn +1 /dev/sshserial' 2>&1 >/dev/$serialdev; clear");
```

lib/publiccloud/ssh_interactive.pm
pdostal: Just run `ssh sut true` so you're sure that after such command the SSH is initiated.

pdostal: Or `ssh sut test -p /dev/sshserial`

tests/publiccloud/ssh_interactive_start.pm
pdostal: OK. But it would be worth improvement in separate pull request.

lib/publiccloud/instances.pm
pdostal: Can we use this in the future to get rid of `$run_args`?

lib/publiccloud/ssh_interactive.pm
pdostal: Or use something like `script_retry('true');` in order to wait until you get your return code which means that you're logged in the instance.

grisu48: We don't get anything from doing so. `sleep` itself is not evil. One should not use sleep for synchronization problems but that's not the case here.

The intention is to give ssh some time to establish the connection before proceeding. It takes typically 1-2 seconds, so a 5 second sleep here is more than enough for most typical cases.

Also: We can't use any `script_run`, `script_retry`, `assert_script_run` or anything at this point because the tunnel is not yet established.

lib/publiccloud/ssh_interactive.pm
grisu48: The intention of the `sleep` here is to prevent hogging the machine in case this loop ends up in an infinite loop. Without it, this would consume 100% CPU in case something goes wrong.

lib/publiccloud/ssh_interactive.pm
grisu48: The tunnel-terminal tends to swallow characters while the tunnel is in place, every command in this terminal is very unreliable. This will not work.

pdostal: I don't think that this is the case here. The previous SSH command is way longer and works reliable.

pdostal: Maybe your experience initiates in your verification runs without `-y` where the previous SSH was messing your console up?

lib/publiccloud/ssh_interactive.pm
pdostal: Oh, I did not notice the `while` - please get rid of it. It was not necessary before.

pdostal (changes requested):
We need to get rid of sleeps and loops first.

lib/publiccloud/ssh_interactive.pm
grisu48: No this is on the level of the terminal, see e.g. https://duck-norris.qam.suse.de/tests/10956#step/ssh_interactive_end/2 or https://duck-norris.qam.suse.de/tests/10956#step/ssh_interactive_end/2

While the command is active, the terminal swallows characters sometimes. Not the ssh tunnel, the serial terminal.

lib/publiccloud/ssh_interactive.pm
grisu48: It is now because the tunnel will be re-established after reboots. That is the purpose of the whole PR.

lib/publiccloud/ssh_interactive.pm
pdostal: But why? Maybe `-f` could help? But also because of the `while` loop you basically cannot kill that `ssh` background process because of it will be started again in 5 seconds.

lib/publiccloud/ssh_interactive.pm
grisu48: Good idea! Will add it right away.

lib/publiccloud/ssh_interactive.pm
grisu48: I asked and the response was basically "this is somewhat of a known issue and we don't know yet what to do about it". Preliminary investigations didn't gave me any results, my hypothesis is that the serial console doesn't like the piping stuff into `/dev/ttyS0` somehow but I don't have found the real cause nor a way to avoid this. And I really tried.

lib/publiccloud/instances.pm
asmorodskyi: AFAIK , it is not possible . TBH I am not sure how this code works . From what I know each module is executed separately so no matter what code structures you are creating in one module other one won't be able to reach them because after each module context is **destroyed**. This is why currently we passing everything through args of loadtest

lib/publiccloud/ssh_interactive.pm
asmorodskyi: the trap here is that you sleeping on worker to wait for something which happening inside VM . And I can say you with big confidence as witness of dozens of similar stories when initialization which in normal circumstances takes 1-2 seconds can lasts up to one minute on super overloaded worker . 
BUT I agree to deal "till first tweak" :)

b10n1k: There is also `ssh -q -o BatchMode=yes sut` which can query the machine and check if it is up or not. would that work if you wrap it with `script_retry`?

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
+
 sub ssh_interactive_tunnel {
     # Establish the ssh interarctive tunnel to the publiccloud instance.
     # Optional arguments: 'force => 1' - reestablish tunnel, also if already established.
 
     my $instance = shift;
-    my %args = testapi::compat_args({cmd => undef}, ['cmd'], @_);
+    my %args = testapi::compat_args({force => 0, reconnect => 0}, ['force', 'reconnect'], @_);
     $args{force} //= 0;
+    $args{reconnect} //= 0;
```
b10n1k, line 41: i dont think you need those two lines. `testapi::compat_args` should return the ```$VAR1 = 'force';
$VAR2 = 0;
$VAR3 = 'reconnect';
$VAR4 = 1;```

Also the perlpod looks need update

lib/publiccloud/ssh_interactive.pm
b10n1k: again `ssh -q -o BatchMode=yes sut`?

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
 
 sub ssh_interactive_leave {
-    # Check if the SSH tunnel is still up and leave the SSH interactive session
-    script_run("test -p /dev/sshserial && exit", timeout => 0);
-
-    # Restore the environment to not use the SSH tunnel for upload/download from the worker
-    #set_var('SUT_HOSTNAME',          testapi::host_ip());
-    set_var('AUTOINST_URL_HOSTNAME', testapi::host_ip());
+    my $prev_console = current_console();
+    # Switch to the local console, terminte the ssh tunnel and set the serial device back to ttyS0
```
b10n1k, line 79: typo on `terminte`

lib/publiccloud/ssh_interactive.pm
```diff
@@ -1,67 +1,98 @@
...
-    $testapi::distri->set_standard_prompt('root');
+    # Terminate the ssh tunnel.
+    # It can happen that the serial console drops some characters thereby we might need to try this multiple times
+    my $retries = 8;
+    while ($retries-- > 0) {
+        type_string('kill %1', max_interval => 1, lf => 1);    # Slow typing reduces the chance of characters being dropped
+        last if (script_run('true', die_on_timeout => 0) == 0);
+    }
+    die "tunnel-console is not terminated" if ($retries <= 0);
+    assert_script_run("true", fail_message => "tunnel-console is not functional");    # additional health check
```
b10n1k, line 92: i dont understand why do you need this after $retries. So `while` can return true but still fail in the next attempt?

tests/publiccloud/prepare_instance.pm
```diff
@@ -83,7 +84,7 @@ sub run {
     $instance->network_speed_test();
 
     # ssh-tunnel settings
-    prepare_ssh_tunnel($instance) if (get_var('TUNNELED'));
+    prepare_ssh_tunnel($instance) if (is_tunneled);
```
b10n1k, line 87: nit: `if is_tunneled;` or `if is_tunneled()`;

tests/publiccloud/ssh_interactive_start.pm
```diff
@@ -1,45 +1,33 @@
...
 use Mojo::Base 'consoletest';
 use testapi;
 use utils;
 use publiccloud::ssh_interactive "ssh_interactive_tunnel";
+use version_utils;
 
 sub run {
     my ($self, $args) = @_;
-    die "tunnel-console requires the TUNELLED=1 setting" unless (get_var("TUNNELED"));
+    die "tunnel-console requires the TUNELLED=1 setting" unless (is_tunneled);
```
b10n1k, line 19: same nitpicking here

lib/publiccloud/ssh_interactive.pm
grisu48: Again: We can't use any `script_run`, `script_retry`, `assert_script_run` or anything at this point because the tunnel is not yet established.

pdostal (approved):

lib/publiccloud/ssh_interactive.pm
grisu48: Not strictly necessary, if should already work with the ones before.

Still because there is a (low but still) risk that the console is not functional at this point, the check is required as a last barrier to really ensure, everything works as expected.

lib/publiccloud/ssh_interactive.pm
grisu48: ok, done.

lib/publiccloud/ssh_interactive.pm
grisu48: ok

lib/publiccloud/ssh_interactive.pm
grisu48: @b10n1k that would not help us here at all. We need the ssh tunnel itself, not a script. Also we know already that the instance is up and ready, that is handled elsewhere.

I don't see how the ssh BatchMode could help here in any way.

lib/publiccloud/ssh_interactive.pm
grisu48: I changed the order, the first true is actually not needed. Please recheck.

grisu48:
The test run with journalctl and enabled reboot passes: https://duck-norris.qam.suse.de/tests/11012

I updated the latest comments and re-scheduled all test runs.

tests/publiccloud/ssh_interactive_start.pm
pdostal: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15725

asmorodskyi (approved):
LGTM

lib/publiccloud/ssh_interactive.pm
```diff
@@ -72,25 +101,24 @@ sub ssh_interactive_leave {
 sub select_host_console {
     my (%args) = @_;
     $args{force} //= 0;
-    my $tunneled = get_var('TUNNELED');
+    my $tunneled = is_tunneled;
+    my $tunnels_initialized = $tunneled && get_var('_SSH_TUNNELS_INITIALIZED');
```
mimi1vx, line 105: why this var ?  and also why `$tunneled` if there `is_tunneled`

tests/publiccloud/upload_image.pm
```diff
@@ -33,7 +33,7 @@ sub run {
     }
 
     # Download the given image via wget. Note that by default wget retries 20 times before giving up
-    my $cmd = "wget --no-check-certificate --retry-connrefused --retry-on-host-error $img_url -O $img_name";
+    my $cmd = "wget -q --no-check-certificate --retry-connrefused --retry-on-host-error $img_url -O $img_name";
```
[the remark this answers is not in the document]
grisu48: This runs on our PC helper instance which is SLES.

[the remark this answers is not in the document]
grisu48: No but we control the image ourselves and it is only one for every test run. Because we can control the tools we need here we don't need to limit ourselves.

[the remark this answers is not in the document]
mimi1vx: yy, i understand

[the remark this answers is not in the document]
grisu48: The tools image is created [here](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=0045&groupid=276) and is used in all `upload_img` jobs. All of those jobs use the same HDD, regardless of the SLES version because it's just a helper VM.

PR #15655, opened 2022-10-06
Move Trento test to qesap.py

mpagot:
Move Trento test SAP deployment from build.sh to qesap.py. Minor qesapdeploy lib API change to not directly use PUBLICCLOUD settings.

- Verification run: http://openqaworker15.qa.suse.cz/tests/49008 http://openqaworker15.qa.suse.cz/tests/49518 http://openqaworker15.qa.suse.cz/tests/49519

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -43,6 +43,7 @@ my @log_files = ();
 use constant QESAPDEPLOY_PREFIX => 'qesapdep';
 
 our @EXPORT = qw(
+  qesap_create_folder_tree
   qesap_pip_install
   qesap_upload_logs
   qesap_get_deployment_code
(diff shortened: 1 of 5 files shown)


lib/qesapdeployment.pm
```diff
@@ -353,7 +354,8 @@ sub qesap_execute {
 
     my $verbose = $args{verbose} ? "--verbose" : "";
     my %paths = qesap_get_file_paths();
-    my $exec_log = "/tmp/qesap_exec_" . $args{cmd} . ".log.txt";
+    my $exec_log = "/tmp/qesap_exec_$args{cmd}_$args{cmd_options}.log.txt";
+    $exec_log =~ s/[-\s]+/_/g;
```
lpalovsky, line 358: Thanks for the log name fix !
And you even included the cmd option :+1:

lpalovsky (approved):
LGTM :+1:

lib/qesapdeployment.pm
mpagot: It is about fixing the problem of
```
qesap_execute(cmd => 'terraform', verbose => 1, timeout => 1200);
```
and
```
qesap_execute(cmd => 'terraform', cmd_options => '-d', verbose => 1, timeout => 1200);
```
Producing a log file with the same name. 
My proposed solution works but it is not optimal for commands that are not using **cmd_options**, like http://openqaworker15.qa.suse.cz/tests/49519/logfile?filename=cluster_configure-qesap_exec_configure_.log.txt (do you see, the underscore at the end)

jankohoutek:
LGTM

jankohoutek (commented):
LGTM

PR #15657, opened 2022-10-06
Fix mpi failure on permission issues with ping

b10n1k:
Two commits:
* Make handle_password_prompt to consider serial_terminal
* Remove assertion on check_nodes_availability of HPC library

Verification run: http://aquarius.suse.cz/tests/13115

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -272,7 +272,7 @@ sub check_nodes_availability {
     my ($self) = @_;
     my @cluster_nodes = cluster_names();
     foreach (@cluster_nodes) {
-        assert_script_run("ping -c 3 $_");
+        script_sudo("ping -i 5 -c 3 $_");
     }
 }
 
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -31,7 +31,7 @@ Base class implementation of distribution class necessary for testapi
 
 # don't import script_run - it will overwrite script_run from distribution and create a recursion
 use testapi qw(send_key %cmd assert_screen check_screen check_var click_lastmatch get_var save_screenshot
-  match_has_tag set_var type_password type_string enter_cmd wait_serial $serialdev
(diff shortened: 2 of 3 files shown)


okurz (approved):
Both commits look sound to me. Thank you for the attention to detail in the commit messages. I am wondering a bit how scalable that approach with if (is_serial_terminal)/else will be. Also, please mind the suggestion about lib/ changes is. Would be good to add unit tests

lib/hpcbase.pm
```diff
@@ -272,7 +272,7 @@ sub check_nodes_availability {
     my ($self) = @_;
     my @cluster_nodes = cluster_names();
     foreach (@cluster_nodes) {
-        assert_script_run("ping -c 3 $_");
+        script_sudo("ping -c 3 $_");
```
foursixnine, line 275: What was the failure here, that you moved to `script_sudo` instead of `assert_script_run`? you could wait longer between pings, by using `ping -i $interval_in_seconds -c $count $_`, mind the default timeout for assert_script_run.

Looking at [this](http://aquarius.suse.cz/tests/13115#step/mpi_master/64), looks like you're relying on a behavior to avoid the test failing (not intentionally) and not even checking the output from script_sudo... maybe for what you want to do, `script_retry` is a better choice.

foursixnine (changes requested):
With the code as is, `check_nodes_availability` could not do anything, and you will get the same result

lib/hpcbase.pm
b10n1k: Let me ping you tomorrow about that. i will need to take another look myself too. I did use `-i` in the beginning of this PR but i remove it at the end. But i am not sure i am thinking clear right now. But i feel your input is on the right direction. Thank you

b10n1k: i created https://github.com/os-autoinst/os-autoinst/pull/2191 for the failure of `assert_script_sudo`.

b10n1k:
The problem which it was trying to fix on HPC module is actually a bug. So i will close this and create a new one for the serial console

PR #15663, opened 2022-10-07
Workaround SelfInstall kexec issue on UEFI runs

mloviska:
As currently used dracut module (`kiwi-dump-reboot-system.sh`) attempts to `kexec` signed kernel with KEXEC_LOAD` syscall, we need to add workaround for [bsc#1203896](https://bugzilla.suse.com/show_bug.cgi?id=1203896).

The image was successfully copied on the drive, but as the first boot options stays set to CD, before rebooting after failed `kexec` tests need to eject the CD so the SUT boots from HDD.

This PR also adds a verification module that after re-installing the bootloader, `sle|opensuse` efi boot record used instead of generic boot option.

- ticket: [Add new job for Self-Install images with UEFI machine](https://progress.opensuse.org/issues/117130)

#### Verification runs

* [slem5.2](http://kepler.suse.cz/tests/19349#step/host_config/1)
* [slem5.3](http://kepler.suse.cz/tests/19348#step/host_config/1)

### DIFF
--- a/tests/microos/efiboot_check.pm
+++ b/tests/microos/efiboot_check.pm
@@ -1,0 +1,30 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Basic smoke test to verify SUSE/openSUSE efiboot
+# Maintainer: qa-c team <qa-c@suse.de>
+
+use Mojo::Base qw(opensusebasetest);
+use testapi;
+use transactional qw(trup_call check_reboot_changes);
+use version_utils qw(is_sle_micro);
+use utils qw(assert_secureboot_status);
+
+sub run {
+    select_console('root-console');
+
+    trup_call('bootloader');
+    check_reboot_changes;
+
+    my $efiboot = script_output('efibootmgr -v', proceed_on_failure => 1);
+
+    unless ($efiboot == /^Boot\w{4}\*\s(opensuse|sle)-secureboot/) {
+        die 'System has booted from unexpected efi boot record!';
+    }
+
+    validate_script_output('mokutil --sb-state', sub { m/secureboot enabled/i });
+}
+
+1;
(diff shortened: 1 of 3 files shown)


tests/microos/efiboot_check.pm
```diff
@@ -0,0 +1,30 @@
...
+    trup_call('bootloader');
+    check_reboot_changes;
+
+    my $efiboot = script_output('efibootmgr -v', proceed_on_failure => 1);
+
+    unless ($efiboot == /^Boot\w{4}\*\s(opensuse|sle)-secureboot/) {
+        die 'System has booted from unexpected efi boot record!';
+    }
+
+    validate_script_output('mokutil --sb-state', sub { m/secureboot enabled/i });
```
grisu48, line 27: Do you think it makes sense to introduce a `SECURE_BOOT` variable that allows this test module to check for the efi boot, but not secure boot? Something like e.g.

```suggestion
    validate_script_output('mokutil --sb-state', sub { m/secureboot enabled/i }) unless (get_var("SECURE_BOOT", "1") == 0);
```

grisu48 (approved):
LGTM with one suggestion

tests/microos/efiboot_check.pm
mloviska: would make sense for https://openqa.opensuse.org/tests/2834312#step/efiboot_check/27, but sure what is expected there.
@ggardet any idea?

ggardet: Not sure what is the default for self-install, but IMO, SecureBoot should be enabled by default.

mloviska: for sle-micro it is enabled, let me try to make a run for leap-micro 5.3 on x86_64

mloviska: https://openqa.opensuse.org/tests/2866763#step/efiboot_check/26

ggardet: Looks like you found a bug then. ;)

mloviska: > Looks like you found a bug then. ;)

Yay! :smile:  Going to report it then. Thanks

mloviska: https://bugzilla.opensuse.org/show_bug.cgi?id=1205299

pdostal (approved):

PR #15664, opened 2022-10-07
Fix keylime config path

ilmanzo:
New [keylime](https://keylime.dev/) version has changed configuration file folder from /etc to /etc/keylime/ , so we need to adjust config files path for tests

- Related ticket: https://progress.opensuse.org/issues/117499
- Verification run: https://openqa.opensuse.org/tests/2784174


paolostivanin:
Since we execute this test also on 15-SP5, could you please check that this change works also there? Thanks!

ilmanzo:
> Since we execute this test also on 15-SP5, could you please check that this change works also there? Thanks!

good catch, I'll add a verification run for 15SP5

paolostivanin:
LGTM

Vogtinator (approved):

PR #15671, opened 2022-10-10
Adjust timeout in kdump_and_crash for aarch64

mloviska:
JeOS aarch64 runs are often failing in reboot after triggering crash or while waiting for `crash` utility to finish its execution.

- ticket: [[aarch64] Adjust timeouts in kdump_and_crash](https://progress.opensuse.org/issues/117796)

##### Verification runs:

https://openqa.suse.de/tests/9715332

### DIFF
--- a/lib/kdump_utils.pm
+++ b/lib/kdump_utils.pm
@@ -343,6 +343,7 @@ sub configure_service {
 sub check_function {
     my %args = @_;
     $args{test_type} //= '';
+    my $boot_timeout = is_aarch64 || is_hyperv ? 240 : undef;
 
     my $self = y2_module_consoletest->new();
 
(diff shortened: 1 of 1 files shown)


lib/kdump_utils.pm
```diff
@@ -365,15 +366,15 @@ sub check_function {
...
+    # Wait for system's reboot; more time for Hyper-V / aarch64 as it's slow.
+    $self->wait_boot(bootloader_time => $boot_timeout);
     select_console 'root-console';
 
     assert_script_run 'find /var/crash/';
 
     if ($args{test_type} eq 'function') {
         # Check, that vmcore exists, otherwise fail
-        assert_script_run('ls -lah /var/crash/*/vmcore', 180);
+        assert_script_run('ls -lah /var/crash/*/vmcore', is_aarch64 ? 240 : 180);
```
grisu48, line 377: In principle ok, but why not simply increase the timeout to 240 everywhere?

Pro: It's simpler and nobody will ask in a year: "why the heck is this different on aarch64?"  
Cons: We might loose 60 seconds of test time in a very rare corner case where the test run is anyways doomed.

```suggestion
        assert_script_run('ls -lah /var/crash/*/vmcore', 240);
```

grisu48 (approved):
Approved with one tiny comment.

grisu48 (approved):

PR #15675, opened 2022-10-11
Replace select_patterns when unselecting all patterns

tinawang123:
Replace select_patterns when unselecting all patterns

- Related ticket: https://progress.opensuse.org/issues/108293
- Verification run: https://openqa.suse.de/tests/9757048#

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -255,6 +255,41 @@ sub select_all_patterns_by_menu {
     assert_screen 'inst-overview';
 }
 
+=head2 select_not_install_any_pattern 
+
+    select_not_install_any_pattern() 
+
+Being in the "select pattern" screen, performs steps to not install any
+patterns.
+=cut
+
+sub select_not_install_any_pattern {
+    my ($self) = @_;
+
+    # Ensure mouse on certain pattern then right click
+    assert_and_click("minimal-system", button => 'right');
+    assert_screen 'selection-menu';
+    # select action on all patterns
+    wait_screen_change { send_key 'a'; };
+    # confirm do not install
+    assert_and_click 'all-do-not-install';
+    save_screenshot;
+}
+
+=head2 select_visible_unselected_patterns
+
+    select_visible_unselected_patterns([@patterns])
+
+Being in the "select pattern" screen, performs steps to select visible
+patterns.
+=cut
+
+sub select_visible_unselected_patterns {
+    my ($self, $patterns) = @_;
+
+    assert_and_click("$_-pattern") for ($patterns->@*);
+}
+
 =head2 deselect_pattern
 
     deselect_pattern();
--- a/tests/installation/select_only_visible_patterns_from_top.pm
+++ b/tests/installation/select_only_visible_patterns_from_top.pm
@@ -1,0 +1,33 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Select given PATTERNS
+#    You can pass
+#    PATTERNS=minimal,base or
+#
+#    For this you need to have needles that provide base-pattern,
+#    minimal-pattern...
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use utils 'type_string_slow';
+
+sub run {
+    my ($self) = @_;
+    $self->go_to_patterns();
+    # Remove default patterns first
+    $self->select_not_install_any_pattern();
+    # go to the top of the list before looking for the pattern
+    send_key "home";
+    my @patterns = grep($_, split(/,/, get_required_var('PATTERNS')));
+    # Select visible patterns
+    $self->select_visible_unselected_patterns([@patterns]);
+    $self->accept_changes();
+}
+
+1;
--- a/schedule/yast/minimal+base/minimal+base@yast.yaml
+++ b/schedule/yast/minimal+base/minimal+base@yast.yaml
@@ -22,7 +22,7 @@ schedule:
   - installation/clock_and_timezone/accept_timezone_configuration
   - installation/authentication/use_same_password_for_root
   - installation/authentication/default_user_simple_pwd
-  - installation/select_patterns
+  - installation/select_only_visible_patterns_from_top
   - installation/bootloader_settings/disable_boot_menu_timeout
   - installation/security/select_security_module_none
   - installation/launch_installation


coolgw:
In your VR PATTERNS = enhanced_base， but actually select base + enhanced_base?

tinawang123:
> In your VR PATTERNS = enhanced_base， but actually select base + enhanced_base?

It use PATTERNS=base,enhanced_base.  You can check https://openqa.suse.de/tests/9696311/file/vars.json this file.

tinawang123:
> Checking your code, I think the first part is reusable for others, we could call it `select_not_install_any_pattern` but for this hack to select something that you know is visible a priori, we can have another function for the iterative part, because we know they are visible, so `select_current_visible_patterns`, for example, and then you have inside the assert and the click of each one. make sense?

Updated the functions.  And I cannot use assert_and_click, it cannot select the patterns.

lib/y2_installbase.pm
```diff
@@ -207,6 +207,9 @@ sub process_patterns {
             $self->select_all_patterns_by_menu();
             $self->deselect_pattern() if get_var('EXCLUDE_PATTERNS');
         }
+        elsif (get_var('PATTERNS', '') !~ /default|all/) {
```
jknphy, line 210: we should create a new test module, this logic here is not clear why and this is used extensively, we need to narrow where we place this IMO. Besides, it is not transparent that the patterns needs to be visible.
A new test modules `select_visible_patterns` with `$self->process_visible_patterns();` would be much more clear IMO.

jknphy: Additionally we need to apply this change to one test suite at least, but it would be ideal to apply to all the test suites in YaST group that needs this (I think they are are few afir, but could be handle in differen ticket).

lib/y2_installbase.pm
```diff
@@ -255,6 +258,55 @@ sub select_all_patterns_by_menu {
...
+    select_current_visible_patterns()
+
+Being in the "select pattern" screen, performs steps to select visible
+patterns.
+=cut
+
+sub select_current_visible_patterns {
+    my ($self) = @_;
+    # Remove default patterns first
+    select_not_install_any_pattern();
```
jknphy, line 293: I don't think we need to couple with this code here, you can put outside in the new module suggested.

For this method we just need to create a new needle with [this](https://openqa.suse.de/tests/9680761#step/select_patterns/17) and [this](https://openqa.suse.de/tests/9680761#step/select_patterns/22) and do `assert_and_click`.
The tag for the needle shouldn't be `on-pattern` we need something different, like `pattern-$pattern`
Makes sense? Because this method looks complicated clicking always on home, going to the first item, which is not the idea, user could navigate to the end and select there two visible patterns. Besides I would use the existing iterative approach `process_patterns` for this instead. So my proposal is to just click both items directly with new code.

tinawang123: As at on-pattern, it already use 'pattern-$pattern' tag, please check: https://openqa.suse.de/tests/9696311#step/select_patterns/16 tag: pattern-base.   To separate my assert_and_click pattern from on-pattern. I created tag named '$pattern-pattern'

lib/y2_installbase.pm
tinawang123: Updated

lib/y2_installbase.pm
```diff
@@ -189,6 +189,25 @@ sub move_down {
...
+    process_visible_patterns();
+
+Decides to select specific patterns based on setting PATTERNS
+Possible values for PATTERNS
+  PATTERNS=minimal,base
+=cut
+
+sub process_visible_patterns {
+    my ($self) = @_;
+    if (get_required_var('PATTERNS')) {
```
jknphy, line 203: `get_required_var` dies if doesn't find the setting, you don't need an `if`
but you don't use it in the function, better follow suggestions below.

tests/installation/select_visible_patterns.pm
```diff
@@ -0,0 +1,27 @@
...
+#    minimal-pattern...
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use utils 'type_string_slow';
+
+sub run {
```
jknphy, line 20: we could call this test module `select_only_visible_patterns_from_top` so it is clear that we clean the list and navigate from top. Same for the method name: `process_only_visible_patterns_from_top`.

tests/installation/select_visible_patterns.pm
```diff
@@ -0,0 +1,27 @@
...
+    my ($self) = @_;
+    $self->go_to_patterns();
+    $self->process_visible_patterns();
```
jknphy, line 23: you don't need existing function `process_visible_patterns` or as suggested `process_only_visible_patterns_from_top`, we can just have here both methods:
`select_not_install_any_pattern` and `select_current_visible_patterns` I think it is more clear, and only one of those actually have parameters.

lib/y2_installbase.pm
```diff
@@ -255,6 +274,50 @@ sub select_all_patterns_by_menu {
...
+
+    select_current_visible_patterns()
+
+Being in the "select pattern" screen, performs steps to select visible
+patterns.
+=cut
+
+sub select_current_visible_patterns {
+    my ($self) = @_;
+    my @patterns = split(/,/, get_var('PATTERNS', ''));
```
jknphy, line 308: we don't need the default value `' '`, we could use `get_required_var`.

lib/y2_installbase.pm
```diff
@@ -255,6 +274,50 @@ sub select_all_patterns_by_menu {
...
+    # remove empty elements
+    @patterns = grep { $_ ne '' } @patterns;
```
jknphy, line 310: we should do this outside the function in the test module, I think better with your approach [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15722/files#diff-edbe5e5e59e46c6bbb5beae3b3116a94517b2e181048a157072dab57f72219a8R17)

lib/y2_installbase.pm
```diff
@@ -255,6 +274,50 @@ sub select_all_patterns_by_menu {
...
+
+    for my $pattern (@patterns) {
+        # go to the top of the list before looking for the pattern
+        send_key "home";
```
jknphy, line 314: I would put this line at the level of test module, besides you don't need to do it for each iteration.

lib/y2_installbase.pm
```diff
@@ -255,6 +274,50 @@ sub select_all_patterns_by_menu {
...
+        # move the list of patterns down until the current pattern is found
+        send_key_until_needlematch "$pattern-pattern", "down", 80;
```
jknphy, line 316: why we cannot avoid iteration and just do `assert_and_click` ? that was the whole point to have a different needle, because the element is not selected, and the iteration one matches only with selected, because otherwise we could iterate the list with existing code without creating anything new...

lib/y2_installbase.pm
tinawang123: When I want to search one pattern, I need put the pattern list part to the first page.
If the pattern cannot be found at first page, I need send down key to move the page to the next.
So for each pattern, I need put the key to 'home' first.

tinawang123: When I want to search one pattern, I need put the pattern list part to the first page.
If the pattern cannot be found at first page, I need send down key to move the page to the next.
So for each pattern, I need put the key to 'home' first.

lib/y2_installbase.pm
tinawang123: The purpose to do this iteration with down is change to next page for pattern list. As, the pattern list has very long list, it cannot show in page. If some patterns cannot be found at first page, we need this down key to move to next page to get the expected pattern.

tinawang123: If the pattern can be found at first page, It will not send down key.

lib/y2_installbase.pm
jknphy: putting it out we can do the same from the bottom in another test module.
I still don't get why you need 'down', the pattern should be always found (unselected), because this method only works for visible patterns.

lib/y2_installbase.pm
jknphy: we always expect the item to be found and first page, that is why we call it 'visible', the users of that method should know their exact location, that is why I think that we can put 'alt-home' or 'alt-end' at the level of test, so the user can use this shortcut (it is a bit of a hack, but if you have always the same patterns, works super fast).

tests/installation/select_only_visible_patterns_from_top.pm
```diff
@@ -0,0 +1,33 @@
...
+#
+#    For this you need to have needles that provide base-pattern,
+#    minimal-pattern...
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use utils 'type_string_slow';
```
jknphy, line 18: this library is not used.

PR #15678, opened 2022-10-11
Adjust from susedistribution which do not work with serial terminal

b10n1k:
Redirection to /dev/<serialdev> doesnt need when the test modules use serial terminal. This causes a problem to the assertions of the commands' markers of the openqa scripts. One such case is `script_sudo`

Also `assert_screen` doesnt make sense on serial_terminal. `handle_password_prompt` fails when test runs on serial_terminal for such reason. Make it work in either case.

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -31,7 +31,7 @@ Base class implementation of distribution class necessary for testapi
 
 # don't import script_run - it will overwrite script_run from distribution and create a recursion
 use testapi qw(send_key %cmd assert_screen check_screen check_var click_lastmatch get_var save_screenshot
-  match_has_tag set_var type_password type_string enter_cmd wait_serial $serialdev
+  match_has_tag set_var type_password type_string enter_cmd wait_serial $serialdev is_serial_terminal
   mouse_hide send_key_until_needlematch record_info record_soft_failure
   wait_still_screen wait_screen_change get_required_var diag);
 
@@ -59,7 +59,11 @@ sub handle_password_prompt {
     $console //= '';
 
     return if get_var("LIVETEST") || get_var('LIVECD');
-    assert_screen("password-prompt", 60);
+    if (is_serial_terminal()) {
+        wait_serial(qr/Password:\s*$/i, timeout => 30);
+    } else {
+        assert_screen("password-prompt", 60);
+    }
     if ($console eq 'hyperv-intermediary') {
         type_string get_required_var('VIRSH_GUEST_PASSWORD');
     }
@@ -347,7 +351,8 @@ sub script_sudo {
 
     my $str = time;
     if ($wait > 0) {
-        $prog = "$prog; echo $str-\$?- > /dev/$testapi::serialdev" unless $prog eq 'bash';
+        $prog .= "; echo $str-\$?-" unless $prog eq 'bash';
+        $prog .= " > /dev/$testapi::serialdev" unless is_serial_terminal();
     }
     enter_cmd "clear";    # poo#13710
     enter_cmd "su -c \'$prog\'", max_interval => 125;
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -358,7 +358,7 @@ sub script_sudo {
     enter_cmd "su -c \'$prog\'", max_interval => 125;
     handle_password_prompt unless ($testapi::username eq 'root');
     if ($wait > 0) {
-        if ($prog eq 'bash') {
+        if ($prog =~ /^bash/) {
             return wait_still_screen(4, 8);
         }
         else {
--- a/t/10_susedistribution.t
+++ b/t/10_susedistribution.t
@@ -1,0 +1,75 @@
+#!/usr/bin/perl
+# Copyright @ SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use testapi;
+use Test::MockModule;
+use susedistribution;
+
+my $testapi_mocked = Test::MockModule->new('testapi');
+$testapi_mocked->mock(type_string => sub { 'randompass' });
+my $susedistri = susedistribution->new();
+
+my $suse_mocked = Test::MockModule->new('susedistribution');
+$suse_mocked->noop(qw(send_key enter_cmd));
+# Increase counter for each wait_serial invocation
+my $wait_serial_hit = 0;
+$suse_mocked->mock(wait_serial => sub { $wait_serial_hit++; return 'Password:' });
+# Increase counter for each assert_screen invocation
+my $assert_screen_called = 0;
+$suse_mocked->mock(assert_screen => sub { $assert_screen_called++; return 1 });
+# Increase counter for each wait_still_screen invocation
+my $wait_still_screen_called = 0;
+$suse_mocked->mock(wait_still_screen => sub { $wait_still_screen_called++; return 1 });
+
+subtest 'handle_password_prompt on serial terminal checks wait_serial' => sub {
+    set_var('VIRTIO_CONSOLE', '1');
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => 1);
+    ok susedistribution::handle_password_prompt(), 'password prompt is handled';
+    is $wait_serial_hit, 1, 'wait_serial is called';
+    is $assert_screen_called, 0, 'assert screen is not called';
+};
+
+subtest 'handle_password_prompt assert screen on when not serial_terminal' => sub {
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => 0);
+    ok susedistribution::handle_password_prompt(), 'password prompt handled';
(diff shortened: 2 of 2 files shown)


okurz (approved):

b10n1k:
@perlpunk @okurz please review added unit tests

lib/susedistribution.pm
```diff
@@ -347,7 +351,8 @@ sub script_sudo {
 
     my $str = time;
     if ($wait > 0) {
-        $prog = "$prog; echo $str-\$?- > /dev/$testapi::serialdev" unless $prog eq 'bash';
+        $prog = "$prog; echo $str-\$?-" unless $prog eq 'bash';
+        $prog = "$prog > /dev/$testapi::serialdev" unless is_serial_terminal();
```
perlpunk, line 355: Perl has a convenient `.=` operator:
```suggestion
        $prog .= "; echo $str-\$?-" unless $prog eq 'bash';
        $prog .= " > /dev/$testapi::serialdev" unless is_serial_terminal();
```

t/10_susedistribution.t
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/perl
+# Copyright @ SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use testapi;
+use Test::MockModule;
+use Test::MockObject;
```
perlpunk, line 8: This module isn't used, so I would remove it:
```suggestion
```

t/10_susedistribution.t
```diff
@@ -0,0 +1,52 @@
...
+    ok susedistribution::handle_password_prompt();
+};
+
+subtest 'handle_password_prompt assert screen on when not serial_terminal' => sub {
+    set_var('BACKEND', 'qemu');
+    my $testapi_mocked = Test::MockModule->new('testapi');
+    $testapi_mocked->mock('is_serial_terminal' => sub { return 0 });
+    $testapi_mocked->mock('type_string' => sub { return 'randompass' });
+    $testapi_mocked->noop('send_key', 'assert_screen');
+    require susedistribution;
```
perlpunk, line 28: The problem with requiring `susedistribution` a second time is, that it is basically a no-op.
So changing the functions `is_serial_terminal` etc. after the first `require` will not use the new ones.

But as the test is passing, that means, the test is not sufficient enough.
You are only checking the return value of `handle_password_prompt`, but I think it should check more details.

I will play with it a bit and come back to you.

t/10_susedistribution.t
```diff
@@ -0,0 +1,52 @@
...
+
+subtest 'handle_password_prompt on serial terminal checks wait_serial' => sub {
+    set_var('VIRTIO_CONSOLE', '1');
+    set_var('BACKEND', 'qemu');
+    my $testapi_mocked = Test::MockModule->new('testapi');
+    $testapi_mocked->mock('is_serial_terminal' => sub { return 1 });
+    $testapi_mocked->mock('type_string' => sub { return 'randompass' });
+    $testapi_mocked->noop('send_key', 'assert_screen');
+    $testapi_mocked->mock('wait_serial' => sub { return 'Password:' });
+    require susedistribution;
```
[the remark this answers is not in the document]
b10n1k: The truth is that I was provided with this solution as the `use susedistribution;` seemed to not working. @Tina explained this to me on /#discuss-perl/. 

>  with this code the mocking will happen first, and then susedistribution will get the mocked functions instead

and also suggested to mock `susedistribution`.
But the tests passed i didnt bother to study them further.

lib/susedistribution.pm
b10n1k: *ready for review*

t/10_susedistribution.t
b10n1k: *ready for review*

t/10_susedistribution.t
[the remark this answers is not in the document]
perlpunk: I'm not `@tina` but @perlpunk.

t/10_susedistribution.t
```diff
@@ -0,0 +1,52 @@
...
+use Test::Most;
+use testapi;
+use Test::MockModule;
+use Test::MockObject;
+
+subtest 'handle_password_prompt on serial terminal checks wait_serial' => sub {
+    set_var('VIRTIO_CONSOLE', '1');
+    set_var('BACKEND', 'qemu');
+    my $testapi_mocked = Test::MockModule->new('testapi');
+    $testapi_mocked->mock('is_serial_terminal' => sub { return 1 });
```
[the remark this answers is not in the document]
b10n1k: *ready for review*

t/10_susedistribution.t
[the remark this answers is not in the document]
perlpunk: I suggested to use require in your example because it only had a single test.
I'm now rewriting the test with mocking susedistribution, which was my second suggestion.
please stand by

t/10_susedistribution.t
perlpunk: Here is my suggestion: https://gist.github.com/perlpunk/8576af61249bdbba51900821170c47f4

b10n1k: interesting approach
I did some minor changes without modifying the main logic
*ready for review*

t/10_susedistribution.t
```diff
@@ -0,0 +1,58 @@
...
+    is $wait_serial_hit, 1;
+    is $assert_screen_called, 1;
+};
+
+subtest 'script_sudo works when is not serial_terminal and wait>0' => sub {
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => sub { return 0 });
+    local $testapi::serialdev = 23;
+    local $testapi::username = 'me';
+    ok susedistribution::script_sudo('echo foo', wait => 10);
```
perlpunk, line 44: Like we discussed, the usage of the function `susedistribution::script_sudo` seems to be wrong (note that it is different from `testapi::script_sudo`):
```suggestion
    ok (susedistribution->script_sudo('echo foo', 10));
```
And the same change below.

perlpunk (approved):

t/10_susedistribution.t
```diff
@@ -0,0 +1,75 @@
...
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => sub { 0 });
+    ok susedistribution::handle_password_prompt();
+    is $wait_serial_hit, 1;
+    is $assert_screen_called, 1;
+};
+
+subtest 'script_sudo works when is not serial_terminal and wait>0' => sub {
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => sub { return 0 });
```
Martchus, line 45: Lines like this could be written as
```suggestion
    $suse_mocked->mock(is_serial_terminal => 0);
```

t/10_susedistribution.t
```diff
@@ -0,0 +1,75 @@
...
+my $wait_still_screen_called = 0;
+$suse_mocked->mock(wait_still_screen => sub { $wait_still_screen_called++; return 1 });
+
+subtest 'handle_password_prompt on serial terminal checks wait_serial' => sub {
+    set_var('VIRTIO_CONSOLE', '1');
+    set_var('BACKEND', 'qemu');
+    $suse_mocked->mock('is_serial_terminal' => sub { 1 });
+    ok susedistribution::handle_password_prompt();
+    is $wait_serial_hit, 1;
+    is $assert_screen_called, 0;
```
Martchus, line 32: It would make sense to add test descriptions here and in other subtests, e.g.:

```suggestion
    ok susedistribution::handle_password_prompt(), 'password prompt handled';
    is $wait_serial_hit, 1, 'wait_serial called';
    is $assert_screen_called, 0, 'assert screen not called';
```

t/10_susedistribution.t
b10n1k: Can we resolve this or it is still stands as a request. I have come up with different approach.

t/10_susedistribution.t
b10n1k: i thought i had changed that. checking

b10n1k: Somehow i revert that part. i have changed it. I will submit in the morning

t/10_susedistribution.t
b10n1k: Done. will submi in the morning

t/10_susedistribution.t
b10n1k: Done *ready for review*

t/10_susedistribution.t
b10n1k: Done *ready for review*

Martchus (approved):

okurz (approved):
Awesome to see that you added a unit test +1

b10n1k:
> Awesome to see that you added a unit test +1

Honestly I consider it must. it even found issues in other places. and thanks to @perlpunk for the help

Vogtinator:
This broke at least two tests:

https://progress.opensuse.org/issues/118645
https://progress.opensuse.org/issues/118648

PR #15680, opened 2022-10-11
return value for qesap_execute

lpalovsky:
Currently, qesap_exec (lib/qesapdeployment) function handles command 
execution failures which does not allow calling function to make the 
decision. This PR makes execution done via script_run with returning 
RC only.

- Related ticket: https://jira.suse.com/browse/TEAM-6822
- Needles: N/A
- Verification run: https://mordor.suse.cz/tests/5059

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -346,6 +346,7 @@ sub qesap_sh_destroy {
 
     Execute qesap glue script commands. Check project documentation for available options:
     https://github.com/SUSE/qe-sap-deployment
+    Test only returns execution result, failure has to be handled by calling method.
 =cut
 
 sub qesap_execute {
@@ -368,8 +369,9 @@ sub qesap_execute {
 
     push(@log_files, $exec_log);
     record_info('QESAP exec', "Executing: \n$qesap_cmd");
-    assert_script_run($qesap_cmd, timeout => $args{timeout});
+    my $exec_rc = script_run($qesap_cmd, timeout => $args{timeout});
     qesap_upload_logs();
+    return $exec_rc;
 }
 
 =head3 qesap_get_inventory
@@ -395,6 +397,7 @@ sub qesap_get_inventory {
     - generates config files with qesap script
 
     For variables example see 'qesap_yaml_replace'
+    Returns only result, failure handling has to be done by calling method.
 =cut
 
 sub qesap_prepare_env {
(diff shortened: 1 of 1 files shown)


lib/qesapdeployment.pm
```diff
@@ -368,8 +369,9 @@ sub qesap_execute {
 
     push(@log_files, $exec_log);
     record_info('QESAP exec', "Executing: \n$qesap_cmd");
-    assert_script_run($qesap_cmd, timeout => $args{timeout});
+    my $exec_rc = script_run($qesap_cmd, timeout => $args{timeout});
     qesap_upload_logs();
+    return ($exec_rc);
```
mpagot, line 374: Why parenthesis ?

lib/qesapdeployment.pm
```diff
@@ -423,8 +426,9 @@ sub qesap_prepare_env {
 
     record_info("QESAP conf", "Generating tfvars file");
     push(@log_files, $paths{terraform_dir} . '/' . $provider . "/terraform.tfvars");
-    qesap_execute(cmd => 'configure', verbose => 1);
+    my $exec_rc = qesap_execute(cmd => 'configure', verbose => 1);
```
mpagot, line 429: I propose not to change API for this, I propose to directly manage exec_rc within qesap_prepare_env.
Do we have a use case to allowing configuration error?

lib/qesapdeployment.pm
mpagot: API change to be managed in caller. Here some of them:
- https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/sles4sap/qesapdeployment/deploy.pm
- https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/sles4sap/qesapdeployment/destroy.pm
- https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a58baa00222ae7c99ded4a82589d0ba4cdf07496/lib/trento.pm#L288


I'm fine about it

mpagot:
This PR is potentially a good candidate to also add a test in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/t/09_qesapdeployment.t

lib/qesapdeployment.pm
lpalovsky: Sorry... it was supposed to be without space return($exec_rc). I removed the parentheses completely.

lib/qesapdeployment.pm
lpalovsky: Just to have uniform behavior for all commands and ability to handle the error by calling module instead of library. 
But I don't need that one at least atm. 
Will make it fail directly in the lib.

mpagot (approved):
LGTM

dzedro (approved):

PR #15683, opened 2022-10-11
New testsuite for testing Kernel Live Patching in SLE Micro

pablo-herranz:
- Related ticket: https://progress.opensuse.org/issues/98234
- Needles: 
- Verification run: https://openqa.suse.de/t9725261

### DIFF
--- a/tests/kernel/qa_test_klp.pm
+++ b/tests/kernel/qa_test_klp.pm
@@ -16,7 +16,8 @@ use testapi;
 use serial_terminal 'select_serial_terminal';
 use utils;
 use registration;
-use version_utils 'is_sle';
+use version_utils qw(is_sle);
+use Wrappers::packages;
 
 sub run {
     if (get_var('AZURE')) {
@@ -25,7 +26,7 @@ sub run {
     }
 
     if (script_run('[ -d /lib/modules/$(uname -r)/build ]') != 0) {
-        zypper_call('in -l kernel-devel');
+        install_package('kernel-devel');
     }
 
     my $git_repo = get_required_var('QA_TEST_KLP_REPO');
@@ -35,8 +36,7 @@ sub run {
     (is_sle(">12-sp1") || !is_sle) ? select_serial_terminal() : select_console('root-console');
 
     add_suseconnect_product("sle-sdk") if (is_sle('<12-SP5'));
-    zypper_call('in -l autoconf automake gcc git make');
-
+    install_package('autoconf automake gcc git make');
     assert_script_run('git config --global http.sslVerify false');
     assert_script_run('git clone ' . $git_repo);
     assert_script_run("cd $dir && ./run.sh", 2760);
--- a/tests/microos/libzypp_config.pm
+++ b/tests/microos/libzypp_config.pm
@@ -11,8 +11,7 @@ use base "consoletest";
 use strict;
 use warnings;
 use testapi;
-use version_utils qw(is_jeos);
-use serial_terminal qw(select_serial_terminal);
+use version_utils qw(is_jeos is_transactional);
 
 sub run {
     select_serial_terminal();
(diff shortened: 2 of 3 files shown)


tests/kernel/qa_test_klp.pm
```diff
@@ -26,7 +27,11 @@ sub run {
     }
 
     if (script_run('[ -d /lib/modules/$(uname -r)/build ]') != 0) {
-        zypper_call('in -l kernel-devel');
+        if (is_transactional) {
+            trup_call('pkg in -l kernel-devel');
```
mdoucha, line 31: It'd be better to add a helper function like `install_package()` that'll do this if/else internally.

tests/kernel/qa_test_klp.pm
```diff
@@ -36,8 +41,11 @@ sub run {
     (is_sle(">12-sp1") || !is_sle) ? $self->select_serial_terminal() : select_console('root-console');
 
     add_suseconnect_product("sle-sdk") if (is_sle('<12-SP5'));
-    zypper_call('in -l autoconf automake gcc git make');
-
+    if (is_transactional) {
+        trup_call('in -l autoconf automake gcc git make');
```
mdoucha, line 45: Shouldn't the argument start with `pkg in`?

mdoucha (changes requested):

tests/kernel/qa_test_klp.pm
pablo-herranz: You're right, I'll fix it

tests/kernel/qa_test_klp.pm
pablo-herranz: Yes, sounds like a good idea!

mdoucha:
The `install_package()` function will be useful for other tests as well so it'd be better to put it somewhere in `lib/`. Preferably in a separate commit.

tests/kernel/qa_test_klp.pm
```diff
@@ -15,7 +15,17 @@ use base 'opensusebasetest';
...
+use transactional 'trup_call';
+
+sub install_package {
+    my $command = $_[0];
+    if (is_transactional) {
+        trup_call('pkg in -l ' . $command);
+    } else {
+        zypper_call('in -l ' . $command);
+    }
+}
```
foursixnine, line 28: as @mdoucha is suggesting, please move this to `lib/Wrappers/Packages.pm` (or something that isn't just dumping it on `lib/`

tests/microos/libzypp_config.pm
```diff
@@ -11,15 +11,19 @@ use base "opensusebasetest";
...
     shift->select_serial_terminal();
     unless (check_var('FLAVOR', 'JeOS-for-AArch64') || check_var('FLAVOR', 'JeOS-for-RPi')) {
         assert_script_run 'egrep -x "^solver.onlyRequires ?= ?true" /etc/zypp/zypp.conf';
         assert_script_run 'egrep -x "^rpm.install.excludedocs ?= ?yes" /etc/zypp/zypp.conf';
     }
-    assert_script_run sprintf('egrep -x "^multiversion ?=%s" /etc/zypp/zypp.conf', is_jeos ? ' ?provides:multiversion\(kernel\)' : '');
+    if (is_jeos || is_transactional) {
+        assert_script_run "echo 'multiversion = provides:multiversion(kernel)' >> /etc/zypp/zypp.conf";
+        assert_script_run "echo 'multiversion.kernels = latest' >> /etc/zypp/zypp.conf";
+        assert_script_run "echo \"LIVEPATCH_KERNEL='always'\" >> /etc/sysconfig/livepatching";
```
foursixnine, line 25: Maybe:

```suggestion
        assert_script_run 'echo LIVEPATCH_KERNEL="always" >> /etc/sysconfig/livepatching';
```

foursixnine (changes requested):

tests/microos/libzypp_config.pm
pablo-herranz: Changed :)

tests/kernel/qa_test_klp.pm
pablo-herranz: I've made the changes, let's see if this works this way :crossed_fingers:

pdostal:
> The `install_package()` function will be useful for other tests as well so it'd be better to put it somewhere in `lib/`. Preferably in a separate commit.

@mdoucha maybe even separate pull request?

mdoucha:
> > The `install_package()` function will be useful for other tests as well so it'd be better to put it somewhere in `lib/`. Preferably in a separate commit.
> 
> @mdoucha maybe even separate pull request?

If the separate PR gets merged faster than this whole thing, then sure, why not.

czerw (approved):

jlausuch:
This is a bit tricky, you can't just run `qa_test_klp.pm` as it won't find the needed packages `autoconf automake gcc git make. I would suggest to close this one and let kernel qa team take over as they might have better ideas how to handle live patching on sle micro.

PR #15685, opened 2022-10-11
Remove /var/lib/wicked/*.xml also on svirt backend

pdostal:
- Related ticket: [poo#116770](https://progress.opensuse.org/issues/116770)
- Verification run: [create_hdd_textmode](https://openqa.suse.de/tests/9716078), [extra_tests_textmode_mod_desktop](https://openqa.suse.de/tests/9715835), [extra_tests_textmode](https://openqa.suse.de/tests/9716177), [yast2_ncurses_textmode](https://openqa.suse.de/tests/9716178)

### DIFF
--- a/tests/shutdown/cleanup_before_shutdown.pm
+++ b/tests/shutdown/cleanup_before_shutdown.pm
@@ -9,7 +9,8 @@
 # - if DEBUG_SHUTDOWN is set, then collect detailed logs to investigate shutdown issues
 #   and redirect them to serial console
 # - if KEEP_PERSISTENT_NET_RULES is set, 70-persistent-net.rules will not be deleted on backend with image support
-# - dhcp cleanup on qemu backend (stop network and wickedd, remove xml files from /var/lib/wicked)
+# - Clean /var/lib/wicked - remove the DUID so it gets regenerated and forget DHCP leases
+#     ( There might be multiple machines at the same time originating from the same HDD )
 # - if DESKTOP is set, then set 'ForwardToConsole=yes', 'MaxLevelConsole=debug' and 'TTYPath=/dev/$serialdev'
 #   in /etc/systemd/journald.conf and restart systemd-journalctl
 # Maintainer: Oleksandr Orlov <oorlov@suse.de>
@@ -50,8 +51,8 @@ END_SCRIPT
 
     # Proceed with dhcp cleanup on qemu backend only.
     # Cleanup is made, because if same hdd image used in multimachine scenario
-    # on several nodes, the dhcp clients use same id and cause conflicts on dhcpd server.
-    if (is_qemu) {
+    # on several nodes, the dhcp clients use same DUID and cause conflicts on dhcpd server.
+    if (is_qemu || is_svirt_except_s390x) {
         my $network_status = script_output('systemctl status network');
         # Do dhcp cleanup for wicked
         if ($network_status =~ /wicked/) {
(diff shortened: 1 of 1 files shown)


tests/shutdown/cleanup_before_shutdown.pm
```diff
@@ -50,16 +50,16 @@ END_SCRIPT
 
     # Proceed with dhcp cleanup on qemu backend only.
     # Cleanup is made, because if same hdd image used in multimachine scenario
-    # on several nodes, the dhcp clients use same id and cause conflicts on dhcpd server.
-    if (is_qemu) {
+    # on several nodes, the dhcp clients use same DUID and cause conflicts on dhcpd server.
+    if (is_qemu || is_svirt) {
```
rfan1, line 54: If we don't need this change on s390x, how about `is_svirt_except_s390x`

rfan1:
I suggest adding some comments in line 12.

tests/shutdown/cleanup_before_shutdown.pm
pdostal: Does it make any difference on s390? Can you give me example test run I can verify on?

tests/shutdown/cleanup_before_shutdown.pm
```diff
@@ -50,16 +50,16 @@ END_SCRIPT
...
         my $network_status = script_output('systemctl status network');
         # Do dhcp cleanup for wicked
         if ($network_status =~ /wicked/) {
             systemctl 'stop network.service';
             systemctl 'stop wickedd.service';
             assert_script_run('ls /var/lib/wicked/');
             save_screenshot;
-            script_run('rm -f /var/lib/wicked/*.xml');
+            script_run('rm -f /var/lib/wicked/duid.xml');
```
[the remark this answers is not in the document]
pdostal: This is probably good point. Let's try to remove also `lease-*.xml` - verification run: https://openqa.suse.de/tests/9715937

tests/shutdown/cleanup_before_shutdown.pm
rfan1: for 390x, we use static ip address  defined in https://gitlab.suse.de/openqa/salt-pillars-openqa/-/blob/master/openqa/workerconf.sls

tests/shutdown/cleanup_before_shutdown.pm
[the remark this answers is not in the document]
rfan1: Cool, one more comment, to make sure the code can work perfectly, can you please start 1 -2 more jobs which use the same qcow2 image in parallel?

tests/shutdown/cleanup_before_shutdown.pm
rfan1: Anyway, I don't think use `is_svirt` can lead to any issue, we can use this workaround for s390x as well.

pdostal: That's why we do `rm -f` so if those files are missing we will just proceed. Can you give me example test run I can verify on?

tests/shutdown/cleanup_before_shutdown.pm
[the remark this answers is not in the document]
pdostal: Good idea. I will.

tests/shutdown/cleanup_before_shutdown.pm
rfan1: https://openqa.suse.de/tests/9678944#dependencies

PR #15691, opened 2022-10-12
openvswitch_ssl.pm: update pox to latest version

paolostivanin:
Updating pox to latest master allows us to use Python3 and, thus, fixing the test execution on 15-SP5, where Py2 is no longer available.
Since we ship the same Python3 `3.6.15` to all SLES 15 codestreams, there's no need for us to have a separate use case for Python2.

- Related ticket: https://progress.opensuse.org/issues/115778
- Verification runs:
  - x86_64: https://openqa.suse.de/tests/9716776#step/openvswitch_ssl/1
  - s390x: https://openqa.suse.de/tests/9716778#step/openvswitch_ssl/1


punkioudi (approved):
awesome!

jouyingbin (commented):
Great, LGTM.

PR #15693, opened 2022-10-13
Select yast development pattern only

lemon-suse:
To reduce the execution time for select yast development only,  go to the end of pattern list to select yast development pattern directly.

- Related ticket: https://progress.opensuse.org/issues/108380
- Needles: N/A
- Verification run: 
  https://openqa.nue.suse.com/tests/9720359#step/select_patterns/12   PATTERNS=default,yast-development
  To compare and avoid regression:
  http://openqa.nue.suse.com/tests/9720383#step/select_patterns/127   PATTERNS=default,yast-development,base
  http://openqa.nue.suse.com/tests/9720384#step/select_patterns/13     PATTERNS=all


ge0r (approved):

jknphy:
@lemon-suse could we apply solution in  #15675 here ?
instead of a hack and hard to understand code in that PR we find useful to create specific test module that does ONE thing.
There, creating new test module `tests/installation/select_visible_patterns_from_bottom.pm` (notice I remomed 'only' from existing name) it should do, you just need to create similar function that that one, but without the clean up of the whole list.

The result should be clean and simple. We can refactor those in the future and send the key to locate by test data, but for now two test modules fits well.wdyt?

lemon-suse:
> @lemon-suse could we apply solution in #15675 here ? instead of a hack and hard to understand code in that PR we find useful to create specific test module that does ONE thing. There, creating new test module `tests/installation/select_visible_patterns_from_bottom.pm` (notice I remomed 'only' from existing name to reference that there is not clean up of all patterns) it should do, you just need to create similar function that that one, but without the clean up of the whole list.
> 
> For patterns we will not need `PATTERNS | default,yast-development` just 'yast-development' I guess.
> 
> The result should be clean and simple. We can refactor those in the future and send the key to locate by test data, but for now two test modules fits well.wdyt?

Ok, updated. The new VR: http://openqa.nue.suse.com/tests/9775751#step/select_visible_patterns_from_bottom/12

tests/installation/select_visible_patterns_from_bottom.pm
```diff
@@ -0,0 +1,32 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Select only YaST Development pattern
```
jknphy, line 6: without the 'only' and better say that we select a pattern from the bottom.

tests/installation/select_visible_patterns_from_bottom.pm
```diff
@@ -0,0 +1,32 @@
...
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    my ($self) = @_;
+    $self->go_to_patterns();
+    my @patterns = grep($_, split(/,/, get_required_var('PATTERNS')));
+    # specific for yast development at the end of list
+    if (get_var('PATTERNS') =~ /yast-development/) {
```
jknphy, line 22: this is not dynamic, it is fixed to that pattern specific and needle, but you got the idea :) thanks for making the change reshaping your initial idea.

Could you please replace it with `$self->select_visible_unselected_patterns([@patterns]);` it should work out the box IMO.

Does it work without the tab?

lemon-suse: I can't replace the assert_and_click("pattern-yast-development"); with $self->select_visible_unselected_patterns([@patterns]); because the function use a 'assert_and_click("$_-pattern")' while the existing tag is pattern_yast-development, the prefix is different.

new VR: http://openqa.nue.suse.com/tests/9776383#step/select_visible_patterns_from_bottom/10

jknphy: yes, you need to create a new needle for this, we are not using existing needles.

lemon-suse: Updated code and updated VR. I think it still need create a MR to update cases to make PATTERN=yast development, while how about the scope?  All cases with PATTERNS = default,yast-development ? Or to be easy, just remove 'default' from @patterns in code.

JRivrain: "Remove 'default' from @patterns in code" makes sense, as this module is very specific, and "default" is not an actual pattern. I'd say go ahead. LGTM.

JRivrain (approved):

tests/installation/select_visible_patterns_from_bottom.pm
lemon-suse: Updated code to remove 'default' from @patterns and update the VR also. Please review it again, thanks.

tests/installation/select_visible_patterns_from_bottom.pm
```diff
@@ -0,0 +1,30 @@
...
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    my ($self) = @_;
+    $self->go_to_patterns();
+    my @patterns = grep($_, split(/,/, get_required_var('PATTERNS')));
+    # delete special 'default' key from the check
+    @patterns = grep { $_ ne 'default' } @patterns;
```
jknphy, line 22: we don't need this line, we should remove it from settings via MR.

tests/installation/select_visible_patterns_from_bottom.pm
jknphy: see https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15693/files#r1009059205

tests/installation/select_visible_patterns_from_bottom.pm
JRivrain: Technically yes, however someone checking the settings can interpret that "PATTERNS=yast_develpoment" may mean *only* yast development, and not default patterns, no ?

jknphy: our way is to check the name of the test module, if we pass as settings or test data is just by our own convenience, we have `tests/installation/select_only_visible_patterns_from_top.pm` for disambiguation` with `only` in the name.

PR #15695, opened 2022-10-13
Install libvirt package

tbaev:
Install libvirt package when testing libvirt.

- Related ticket: https://progress.opensuse.org/issues/116407
- Needles: N/A
- Verification run: [KVM](http://openqa.qam.suse.cz/tests/47083#) and [XEN](http://openqa.qam.suse.cz/tests/47089#)

### DIFF
--- a/tests/virtualization/universal/prepare_guests.pm
+++ b/tests/virtualization/universal/prepare_guests.pm
@@ -91,6 +91,11 @@ sub run {
 
     # Ensure additional package is installed
     zypper_call '-t in libvirt-client iputils nmap supportutils';
+
+    # If testing libvirt package specificly
+    if ('libvirt' eq get_var("UPDATE_PACKAGE")) {
+        zypper_call '-t in libvirt';
+    }
     if (script_run("virsh net-list --all | grep default") != 0) {
         assert_script_run "curl " . data_url("virt_autotest/default_network.xml") . " -o ~/default_network.xml";
         assert_script_run "virsh net-define --file ~/default_network.xml";


RoyCai7:
I just checked your VRs, their UPDATE_PACKAGE are kernel-default.
It should be libvirt, please check again, or it cannot trigger your code.
Just like: http://openqa.qam.suse.cz/tests/47089#step/prepare_guests/13

tests/virtualization/universal/prepare_guests.pm
```diff
@@ -91,6 +91,11 @@ sub run {
 
     # Ensure additional package is installed
     zypper_call '-t in libvirt-client iputils nmap supportutils';
+
+    # If testing libvirt package specificly
+    if ('libvirt' eq get_var("UPDATE_PACKAGE")) {
+        zypper_call '-t in libvirt';
```
alice-suse, line 97: This seems the easiest way to make automation pass. But is it meanfingful in real customer scenarios to check if this package is updated to incident repo? Just as developer explained in the bug, `libvirt` package is optional to be installed when setting up kvm host. It is not required in package dependency. So in real host, `libvirt` package is likely not installed either. What's really meaningful for libvirt related incidents, I am afraid, is to check libvirt-damon-xx and other  required packages for virtualization host. HDYT?

tbaev: As discussed in the poo#116407, libvirt is "meta" package that installs more dependencies then the standard xen/kvm patterns from YaST. I think when a new version of the libvirt package is release it makes sens to install it "fully" when testing even that will be "bloated" for production server.

alice-suse: I read the ticket poo#116407 fully. I am afraid this solution is not fully discussed and agreed. See https://progress.opensuse.org/issues/116407#note-20.

tbaev:
This PR will not be needed as discuss in [ticket](https://progress.opensuse.org/issues/116407).

PR #15698, opened 2022-10-13
Restart systemd-journald before checking it's status

Zaoliang:
on svirt-xen-hvm we have issue for checking status of systemd-journald.
add restart command before checking it's status.
see https://progress.opensuse.org/issues/118198

VR: https://openqa.suse.de/tests/9725275


Zaoliang:
https://openqa.suse.de/tests/9725374 works fine for systemd-nspawn@64bit

PR #15699, opened 2022-10-13
Replace serial_terminal on apptainer module with select_console

b10n1k:
This is a workaround to speed up the test as the serial terminal causes some delays due to failing wait_serial check. The `serial_term_prompt` doesnt match with the apptainer interactive terminal.

- Verification run: http://aquarius.suse.cz/tests/13387

### DIFF
--- a/tests/containers/apptainer.pm
+++ b/tests/containers/apptainer.pm
@@ -14,7 +14,9 @@ use containers::utils qw(registry_url);
 
 sub run {
     my ($self) = @_;
-    $self->select_serial_terminal;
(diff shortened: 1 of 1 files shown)


mdoucha:
You can simply change the apptainer prompt using `utils::set_serial_prompt($new_prompt)` and then change it back after to the original value when you exit apptainer shell. If apptainer shell does not honor `$PS1`, just set `utils::$serial_term_prompt` directly.

mdoucha:
Alternatively, you can also use `validate_script_output("apptainer exec $instance 'cat /etc/os-release && echo testdone'")` to avoid prompt juggling entirely.

PR #15706, opened 2022-10-13
Firewalld is disabled for VMWare image

mloviska:
VMWare minimal-vm image image just received cloud init support.
Therefore the firewall has been disabled by default as it is in
Openstack image.

- ticket: [[vmware] test failures due to CI supported add to vmware image](https://progress.opensuse.org/issues/118624)
- VR: http://kepler.suse.cz/tests/19418#step/firewall_enabled/4


grisu48 (approved):
LGTM

PR #15707, opened 2022-10-13
wicked: disable journal RateLimit

cfconrad:
We faced the problem, that journal skiped messages because of RateLimit hits. The message in journal was:
   systemd-journald[355]: Suppressed 1239 messages from /system.slice/wickedd.service

- Verification run: http://openqa.wicked.suse.de/tests/84110


asmorodskyi:
you need to tidy up and than it will be merged

PR #15709, opened 2022-10-13
wicked: Add firewall extension test

cfconrad:
Related bugs: bsc#1189560 bsc#1201053
PR: https://github.com/openSUSE/wicked/pull/933

- Verification run: http://openqa.wicked.suse.de/tests/84114

### DIFF
--- a/tests/wicked/before_test.pm
+++ b/tests/wicked/before_test.pm
@@ -138,6 +138,13 @@ sub run {
             zypper_ar($repo_url . generate_version('_') . '/', name => 'wicked_maintainers', no_gpg_check => 1, priority => 60);
             $package_list .= ' ndisc6';
         }
+        if (check_var('WICKED', 'startandstop')) {
+            # No firewalld on sles 12-SP5 (bsc#1180116)
+            if (is_sle('>12-SP5')) {
+                zypper_call('-q in firewalld', timeout => 400);
+                systemctl('disable --now firewalld');
+            }
+        }
         wicked::wlan::prepare_packages() if (check_var('WICKED', 'wlan'));
 
         $package_list .= ' openvswitch iputils';
--- a/tests/wicked/startandstop/sut/t18_firewall_extension.pm
+++ b/tests/wicked/startandstop/sut/t18_firewall_extension.pm
@@ -1,0 +1,41 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: wicked
+# Summary: Verify firewalld zone settings after ifup/ifdown/ifreload
+# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>
+
+use Mojo::Base 'wickedbase';
+use testapi;
+use Utils::Systemd;
+use version_utils 'is_sle';
+
+has wicked_version => '>=0.6.70';
+
+sub run {
+    my ($self, $ctx) = @_;
+    my $ifc = $ctx->iface();
+
+    return if ($self->skip_by_wicked_version());
+
+    # No firewalld on sles 12-SP5 (bsc#1180116)
+    if (is_sle('<=12-SP5')) {
+        $self->result('skip');
+        return;
+    }
+
+    systemctl('enable --now firewalld');
+
+    $self->get_from_data('wicked/test-ext-firewall.sh', '/tmp/test-ext-firewall.sh', executable => 1);
+    $self->run_test_shell_script("ext-firewall $ifc", "time /tmp/test-ext-firewall.sh '$ifc'");
+
+    $self->skip_check_logs_on_post_run();
+}
+
+sub test_flags {
+    return {always_rollback => 1};
+}
+
+1;
--- a/data/wicked/test-ext-firewall.sh
+++ b/data/wicked/test-ext-firewall.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+
+ifc=${1:-eth1}
+zone1=zone-test-1
+zone2=zone-test-2
+default_zone=$(firewall-cmd --get-default-zone)
+ifcfg="/etc/sysconfig/network/ifcfg-$ifc"
+
+function error {
+    local msg=${1:? Error message is mandatory}
+    local code=${2:-127}
+
+    echo "ERROR: $msg" 1>&2
+    exit "$code"
+}
+
+function fw_zone_list_ifc {
+    local zone=${1:? Function need zone as parameter}
+    local permanent=${2}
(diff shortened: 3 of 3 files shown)


tests/wicked/before_test.pm
```diff
@@ -138,6 +138,13 @@ sub run {
             zypper_ar($repo_url . generate_version('_') . '/', name => 'wicked_maintainers', no_gpg_check => 1, priority => 60);
             $package_list .= ' ndisc6';
         }
+        if (check_var('WICKED', 'startandstop')) {
+            # No firewalld on sles 12-SP5 (bsc#1180116)
+            if (!is_sle('<=12-SP5')) {
```
asmorodskyi, line 143:
```suggestion
            if (is_sle('>12-SP5')) {
```
it is exactly the same but it is always easier to read and understand true expressions ( have read this somewhere :) )

tests/wicked/before_test.pm
```diff
@@ -138,6 +138,13 @@ sub run {
...
+                zypper_call('-q in firewalld', timeout => 400);
```
asmorodskyi, line 144: hm , AFAIK firewalld suppose to be installed by default on all SLE 15

asmorodskyi: yes you messing up here with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15709/files#diff-4444ed2221a3fad7639c99baa4ed295548f6638dd2cda4daf842a1307a7e536cR39

tests/wicked/startandstop/sut/t18_firewall_extension.pm
```diff
@@ -0,0 +1,42 @@
...
+has wicked_version => '>=0.6.70';
+
+sub run {
+    my ($self, $ctx) = @_;
+    my $ifc = $ctx->iface();
+
+    return if ($self->skip_by_wicked_version());
+
+    # No firewalld on sles 12-SP5 (bsc#1180116)
+    if (is_sle('<=12-SP5')) {
```
asmorodskyi, line 25:
```suggestion
    if (is_sle('<12-SP5')) {
```
otherwise SLE 12-SP5 will be in

tests/wicked/before_test.pm
cfconrad: don't understand the link!

We use autoyast to install and it looks like there is no firewalld by default see
 * tumbelweed: http://openqa.wicked.suse.de/tests/84114/logfile?filename=serial_terminal.txt
 * sle-15-sp4: http://openqa.wicked.suse.de/tests/84117/logfile?filename=serial_terminal.txt

tests/wicked/startandstop/sut/t18_firewall_extension.pm
cfconrad: Thx!

cfconrad: Ah but `is_sle('<=12-SP5')` is what we want here, cause we will skipp everything lower and including `12-SP5`.

cfconrad: What we could do is
```perl
if (!is_sle('>12-SP5')) {
```

tests/wicked/before_test.pm
asmorodskyi: just look into `before_test.pm` in line 39 . the point is that when you getting into line 144 there is already logic which trying to handle firewalld . so at least there is no point to try to stop it because if it is installed it is already stopped

tests/wicked/before_test.pm
cfconrad: ack

tests/wicked/before_test.pm
asmorodskyi: maybe you can this install request into module itself ? because currently you installed firewalld even for `wicked_version < 0.6.70`  and module will not run after all

cfconrad: Now I get it!

Or we simply install firewalld via autoyast and deactivate it always?

I mean what we need is firewall deactivated for all tests. But installing it during tests is hard, as we would need to setup network and we have that install logic only in `before` test.

tests/wicked/startandstop/sut/t18_firewall_extension.pm
asmorodskyi: ah right , my bad

tests/wicked/before_test.pm
asmorodskyi: isn't it will fail if package is already installed ?

asmorodskyi: > sle-15-sp4: http://openqa.wicked.suse.de/tests/84117/logfile?filename=serial_terminal.txt

this is NOT SLE 15 SP4 but Leap !  For SLE 15 SP* firewalld is default firewall which should be present by default and if it is not presented it is a bug and blindly installing it is hiding a bug . Regarding Leap / TW yes you might need to install it but you need to use proper `is_leap` expression , regarding TW it will be more complicated because there is no `is_tw` the only way to figure out that it TW is check that it is NOT something else which would be ugly

cfconrad: Sorry for not pasting the right link:
  * SLE 15-SP4 http://openqa.wicked.suse.de/tests/84118/logfile?filename=serial_terminal.txt
  * all my tests: http://openqa.wicked.suse.de/tests/overview?build=firewall_ext
But it looks like TW and LEAP.

Also installing something which is already there, do not produce a errcode:
```bash
> zypper search firewalld
Loading repository data...
Reading installed packages...

S  | Name                        | Summary                                                             | Type
---+-----------------------------+---------------------------------------------------------------------+-----------
i+ | firewalld                   | A firewall daemon with D-Bus interface providing a dynamic firewall | package
   | firewalld                   | A firewall daemon with D-Bus interface providing a dynamic firewall | srcpackage
i  | firewalld-lang              | Translations for package firewalld                                  | package
   | firewalld-prometheus-config | Firewalld configuration file for Prometheus                         | package
   | firewalld-rpcbind-helper    | Tool for static port assignment of NFSv3, ypserv, ypbind services   | package
   | keylime-firewalld           | Firewalld service file for keylime                                  | package
   | susefirewall2-to-firewalld  | Basic SuSEfirewall2 to FirewallD migration script                   | package

> sudo zypper -n -q in firewalld
[sudo] password for root: 

> echo $?
0
```

If it hide a error, when we blindly install the package ... yes it does. But I think this is not the focus of these tests?!

asmorodskyi: ok , I found what I missed initially . Firewalld will be installed by default when you using default installation path . But this test runs against something created by autoyast profile anyway so it is kinda expected that we don't have firewalld here

asmorodskyi:
you need to fix pipeline and it will be merged

PR #15714, opened 2022-10-14
Increase timeout for aarch64

ilmanzo:
After observing flaky behavior, we to introduce an extra timeout to compensate for aarch64 worker performance issue.

- Related ticket: https://progress.opensuse.org/issues/115769
- Verification run: https://openqa.suse.de/t9731591


paolostivanin:
LGTM

PR #15717, opened 2022-10-17
Create hdd via autoyast for leap

rfan1:
- Related ticket: https://progress.opensuse.org/issues/116006
- Needles: Added via openQA webUI
- Verification run: https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.4&build=20221024-1&groupid=39


rfan1:
Please help review/merge https://github.com/os-autoinst/opensuse-jobgroups/pull/218 at the same time.

a-kpappas:
This failure https://openqa.opensuse.org/tests/2829804#step/journalctl/103 seems weird. Are you sure that the changed test doesn´t mess with the assumptions of this test?

rfan1:
> This failure https://openqa.opensuse.org/tests/2829804#step/journalctl/103 seems weird. Are you sure that the changed scenario doesn´t mess with the assumptions of this test?

Good comment! I noticed this error as well, so far, it seems to be the only one case which is not the same as in production job group.

However, I think it should be a sporadic issue since as I can see it passed in previous builds as well.

Let me increase the memory size to 2048. and I will file another ticket to track this random failure.

foursixnine:
> > > This failure https://openqa.opensuse.org/tests/2829804#step/journalctl/103 seems weird. Are you sure that the changed scenario doesn´t mess with the assumptions of this test?
> > 
> > 
> > Good comment! I noticed this error as well, so far, it seems to be the only one case which is not the same as in production job group.
> > However, I think it should be a sporadic issue since as I can see it passed in previous builds as well.
> > Let me increase the memory size to 2048. and I will file another ticket to track this random failure.
> 
> @a-kpappas I have updated the test suite setting with QEMURAM=2048, can you please help review again? Anyway, I don't think the new issue is caused by my code changes :)

Smells like a product bug

rfan1:
> > > > This failure https://openqa.opensuse.org/tests/2829804#step/journalctl/103 seems weird. Are you sure that the changed scenario doesn´t mess with the assumptions of this test?
> > > 
> > > 
> > > Good comment! I noticed this error as well, so far, it seems to be the only one case which is not the same as in production job group.
> > > However, I think it should be a sporadic issue since as I can see it passed in previous builds as well.
> > > Let me increase the memory size to 2048. and I will file another ticket to track this random failure.
> > 
> > 
> > @a-kpappas I have updated the test suite setting with QEMURAM=2048, can you please help review again? Anyway, I don't think the new issue is caused by my code changes :)
> 
> Smells like a product bug

I agree with you, it is a sporadic issue and might be a product bug.

Vogtinator:
This broke `keymap_or_locale_x11` due to the added call to `clear_console` with no explanation: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15831

What's the `leap_default_repos` module for?

rfan1:
> This broke `keymap_or_locale_x11` due to the added call to `clear_console` with no explanation: #15831
> 
> What's the `leap_default_repos` module for?

Thank you @Vogtinator!
The reason why I added this line was due to failure https://openqa.opensuse.org/tests/2849128#step/keymap_or_locale_x11/2.   there was "guest disabled display" there. I tried to use `clear_console` to reset the vnc console. sorry for the impact.

for `leap_default_repos`,  I used this module to copy the repo files which from installed system via interactive installation.
since for autoyast installation, I have trouble to add the repos to the system and don't enable them.

rfan1:
> > This broke `keymap_or_locale_x11` due to the added call to `clear_console` with no explanation: #15831
> > What's the `leap_default_repos` module for?
> 
> Thank you @Vogtinator! The reason why I added this line was due to failure https://openqa.opensuse.org/tests/2849128#step/keymap_or_locale_x11/2. there was "guest disabled display" there. I tried to use `clear_console` to reset the vnc console. sorry for the impact.
> 
> for `leap_default_repos`, I used this module to copy the repo files which from installed system via interactive installation. since for autoyast installation, I have trouble to add the repos to the system and don't enable them.

I may need add a new needle to match `guest-disabled-display` rather than `clear_console`
    if (match_has_tag('guest-disabled-display')) {
            wait_screen_change(sub {
                    send_key 'shift';
            }, 10);

Vogtinator:
> The reason why I added this line was due to failure https://openqa.opensuse.org/tests/2849128#step/keymap_or_locale_x11/2. there was "guest disabled display" there. I tried to use clear_console to reset the vnc console. sorry for the impact.

There actually is not, that's just a needle that did not match. What actually happened is that it's showing tty2.

> for leap_default_repos, I used this module to copy the repo files which from installed system via interactive installation.
since for autoyast installation, I have trouble to add the repos to the system and don't enable them.

That sounds like a severe product bug if that does not work out of the box.

rfan1:
> > The reason why I added this line was due to failure https://openqa.opensuse.org/tests/2849128#step/keymap_or_locale_x11/2. there was "guest disabled display" there. I tried to use clear_console to reset the vnc console. sorry for the impact.
> 
> There actually is not, that's just a needle that did not match. What actually happened is that it's showing tty2.

Thanks again! let me try to fix the test failures once hitting the issue once more.
> 
> > for leap_default_repos, I used this module to copy the repo files which from installed system via interactive installation.
> > since for autoyast installation, I have trouble to add the repos to the system and don't enable them.
> 
> That sounds like a severe product bug if that does not work out of the box.

I raised the question in help_yast channel as well, and got some feedback that for now, we may not able to add some repos and set them as `disabled` in autoyast xml configuration file like: 

`<enabled t="boolean">false</enabled>
`

PR #15723, opened 2022-10-18
Install curl if not present when running script_output 

tinawang123:
Install curl if not present when running script_output with content fetched through HTTP

- Related ticket: https://progress.opensuse.org/issues/116854
- Verification run: http://openqa.suse.de/t9725451
http://openqa.suse.de/tests/9750431#


b10n1k (changes requested):
Please fix commit title and add proper description. Additionally, i wonder if this should fixed in `os-autoinst/distribution.pm` instead once and for good.

jknphy:
> Please fix commit title and add proper description. Additionally, i wonder if this should fixed in `os-autoinst/distribution.pm` instead once and for good.

- [ ] 

> Please fix commit title and add proper description. Additionally, i wonder if this should fixed in `os-autoinst/distribution.pm` instead once and for good.

It is

PR #15684, opened 2022-10-11
Introduce Ansible console test

pdostal:
This test is single machine but uses **SSH** and **SUDO** as those are default.

- Related ticket: [poo#116935](https://progress.opensuse.org/issues/116935)
- Discovered bug: [bsc#1204182](https://bugzilla.opensuse.org/show_bug.cgi?id=1204182)
- Verification run: [Tumbleweed](https://pdostal-server.suse.cz/tests/2632), [Leap 15.4](https://pdostal-server.suse.cz/tests/2605), [Leap 15.3](https://pdostal-server.suse.cz/tests/2633), [SLE 15-SP4](https://pdostal-server.suse.cz/tests/2742), [SLE 15-SP3](https://pdostal-server.suse.cz/tests/2747)

PR #15725, opened 2022-10-18
Get rid of $self in opensusebasetest::select_serial_console

pdostal:
The `$self` parameter in `opensusebasetest::select_serial_console` is not used in the subroutine. That brings only confusion.

- Related threads: [#15684](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15684#discussion_r996648918), [#15643](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15643#discussion_r991080723)
- Verification run: [publiccloud_consoletests@64bit](https://pdostal-server.suse.cz/tests/2775), [publiccloud_img_proof@64bit](https://pdostal-server.suse.cz/tests/2776), [podman_tests@64bit](https://pdostal-server.suse.cz/tests/2768), [autoyast_filesystem_withouthome_yast@64bit](https://pdostal-server.suse.cz/tests/2772), [mau-apparmor@64bit](https://pdostal-server.suse.cz/tests/2773), [qam-wicked_basic_ref@64bit](https://pdostal-server.suse.cz/tests/2771), [containerd_tests@64bit](https://pdostal-server.suse.cz/tests/2769)

The key change is in `lib/serial_terminal.pm`
Please **review** and/or **suggest verification runs**.


b10n1k:
@os-autoinst/tests-maintainer

lib/opensusebasetest.pm
```diff
@@ -1292,7 +1292,7 @@ use root-ssh console directly.
 =cut
 
 sub select_serial_terminal {
-    my ($self, $root) = @_;
+    my ($root) = @_;
```
okurz, line 1295: If you want to prevent confusion you could use signatures with a mandatory argument but as the parameter "root" is optional this won't easily work. But if you want to make this not a member method I recommend to also move it out of the class so that it's clear that this is a free function, not a member method.

okurz (changes requested):

b10n1k:
in my opinion `opensusebasetest` is defined as a class and removing the `$self` from its subroutines breaks the OO theory. The subroutines of the Class should point to the Object which they come from. It s basic OO concept. I would like to see what other have to say.

lib/opensusebasetest.pm
cfconrad, line 1295: Maybe we should allow the call as static function and as member method?

```suggestion
    my ($self, $root) = @_
    $root = $self if (ref($root) ne 'opensusebasetest');
```

pdostal: Do you think that moving this to `utils` would make a sense?

pdostal: This would solve the issue I have in [#15684](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15684#discussion_r996648918).

mdoucha: Yes, `utils` would be a good place. But then this PR also needs to refactor all calls to `select_serial_terminal()` throughout the codebase. Preferably in the same commit.

mdoucha:
> in my opinion `opensusebasetest` is defined as a class and removing the `$self` from its subroutines breaks the OO theory. The subroutines of the Class should point to the Object which they come from. It s basic OO concept. I would like to see what other have to say.

There's no reason why `select_serial_terminal()` needs to be a method of `opensusebasetest`. It can just as well be a stand-alone library function.

pevik:
> in my opinion `opensusebasetest` is defined as a class and removing the `$self` from its subroutines breaks the OO theory. The subroutines of the Class should point to the Object which they come from. It s basic OO concept.

There are static methods in OOP
https://en.wikipedia.org/wiki/Method_(computer_programming)#Static_methods

Static methods in perl
http://www.netalive.org/tinkering/serious-perl/#oop_static_methods
https://stackoverflow.com/questions/3845722/how-should-i-define-static-subroutines-in-perl/3845813#3845813

pevik:
> There's no reason why `select_serial_terminal()` needs to be a method of `opensusebasetest`. It can just as well be a stand-alone library function.

The reason I created `select_serial_terminal()` as `opensusebasetest` method was to be available everywhere without any extra `use`. But sure, no problem to put it to `lib/utils.pm` or anywhere else.

mdoucha:
> The reason I created `select_serial_terminal()` as `opensusebasetest` method was to be available everywhere without any extra `use`. But sure, no problem to put it to `lib/utils.pm` or anywhere else.

The drawback of that is that it complicates use from other library functions. And there are quite a few places in `lib` where `select_serial_terminal()` is used.

pevik:
> The drawback of that is that it complicates use from other library functions. And there are quite a few places in `lib` where `select_serial_terminal()` is used.

+1

@pdostal Right, than thanks for improving/fixing it.

okurz (approved):
ok, convinced :)

pdostal:
Does it mean that the current solution is OK? I was actually in favour of the solution from @cfconrad but I wanted to proceed here after I manage to merge another pull request.

okurz:
> Does it mean that the current solution is OK? I was actually in favour of the solution from @cfconrad but I wanted to proceed here after I manage to merge another pull request.

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15725#discussion_r997866784 looks a bit unusual to me. I just approved the change of code as-is

cfconrad:
> Does it mean that the current solution is OK? I was actually in favour of the solution from @cfconrad but I wanted to proceed here after I manage to merge another pull request.

My solution was meant in terms of backward compatibility and minimal change. 

But I would prefer to extend this PR and change all 550 calls to `select_serial_terminal()` to not passing the object. It's more work but I think it is worth to take this extra `sed`. As the code base in the end is more clean.

Moving this function to a different place sound ok, what about `lib/serial_terminal.pm`.

okurz (approved):

lib/opensusebasetest.pm
pdostal: Please check the current implementation.

pdostal: Thank you Clemens for interesting solution but as you also [suggested](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15725#issuecomment-1283585825) I proceed otherwise.

grisu48 (approved):
No concerns from the QEC related part of the PR, IMHO it should work. I can not comment on the test runs outside of this though.

asmorodskyi (approved):
LGTM . I like this idea !

pevik (approved):
Thanks!

lib/serial_terminal.pm
```diff
@@ -6,7 +6,9 @@ package serial_terminal;
 use 5.018;
 use warnings;
 use testapi;
+use serial_terminal 'select_serial_terminal';
```
pevik, line 9: @pdostal Hm import itself? Please remove this.

pevik (changes requested):
`use serial_terminal 'select_serial_terminal'` in `lib/serial_terminal.pm` is IMHO wrong or at least useless.

lib/serial_terminal.pm
```diff
@@ -287,4 +291,84 @@ sub reboot {
...
+profile or update grub setup manually with add_grub_cmdline_settings()).
+
+On ikvm|ipmi|spvm|pvm_hmc it's expected, that use_ssh_serial_console() has been called
+(done via activate_console()) therefore SERIALDEV has been set and we can
+use root-ssh console directly.
+=cut
+
+sub select_serial_terminal {
+    my ($root) = @_;
+    $root //= 1;
```
pevik, line 333: Maybe instead of these 2 lines just single line:
```
my $root = shift // 1;
```

pdostal: Thank you!

pevik (approved):
LGTM now. I tested few LTP tests on intel, works well.

tests/console/add_phub_extension.pm
```diff
@@ -10,11 +10,12 @@ use base "consoletest";
 use strict;
 use warnings;
 use testapi;
+use serial_terminal 'select_serial_terminal';
 use registration qw(add_suseconnect_product get_addon_fullname);
 
 sub run {
     my $self = shift;
```
b10n1k, line 17: This can be removed now.

b10n1k (changes requested):
`$self` is defined in many places but remains unused. Also as a side note, maybe it is a good chance to introduce some tests.

b10n1k:
> > `$self` is defined in many places but remains unused. Also as a side note, maybe it is a good chance to introduce some tests.
> 
> @b10n1k Given this PR already touches 458, I would not overcomplicate this PR (perfect is the enemy of good, also steps by steps evolution is easier for everybody than revolution).
> 
> BTW exactly this type of request "you could also do this and that and of course, don't forget about the tests" always put me off to do contribution to openQA (test framework is horrible, I'm not the only one who complains that contributing tests means write more code than for the feature).

I would not disagree neither i would agree. it s back and forth discussion. But i told to @pdostal to let me know how he plan to proceed and i will approve in any case. But as a reviewer i feel i should point those things out

pdostal:
This is only about `select_serial_terminal()` and `select_user_serial_terminal` Yianni. But thank you for the suggestion - I will do that.

pevik:
BTW `select_user_serial_terminal()` (i.e. using non-root) is not working well and should be fixed. I had a discussion with @dzedro some time ago and planned to fix it, but in the end just used `sudo -u \#1000`. That's what should be fixed.

pdostal:
> BTW `select_user_serial_terminal()` (i.e. using non-root) is not working well and should be fixed. I had a discussion with @dzedro some time ago and planned to fix it, but in the end just used `sudo -u \#1000`. That's what should be fixed.

Yes, I run into that issue on `tests/console/ansible.pm` - as there is only one serial terminal you need to 'run_cmd 'exit'; reset_consoles();' so it will relog. Also `reset_consoles` resets all consoles so when you `select_console 'root-console'` the next time it's searching for login prompt instead of searching for bash prompt.

lib/Tomcat/ModjkTest.pm
```diff
@@ -33,7 +34,7 @@ sub mod_jk_setup() {
 # Connection from apache2 to tomcat: Functionality test
 sub func_conn_apache2_tomcat() {
     my $self = shift;
```
pevik, line 36: Looks like this one is not needed.

tests/console/cpio.pm
```diff
@@ -11,12 +11,12 @@ use base "consoletest";
 use strict;
 use warnings;
 use testapi;
+use serial_terminal 'select_serial_terminal';
 use utils;
 
 sub run {
 
-    my $self = shift;
```
pevik, line 18: Could you please delete here blank line?

pevik (approved):
You've done amazing work, thank you. I haven't found any test which would have `$self` which is not needed.

b10n1k (approved):
looks good.

pdostal:
Thank you @pevik, I'm not afraid of tests having unused `$self` but tests using undefined `$self` so let's see on Monday in Production :-)

a-kpappas:
Since this has been approved by so many people, I am merging this relatively early on Monday with time to fix things should they come up.

mloviska:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15769

PR #15728, opened 2022-10-18
console/snapper_used_space: Use "none" disable btrfs compression

Vogtinator:
In newer kernel versions, the meaning of "" changed in a backwards-incompatible way. To disable compression, "none" has to be used instead. This also works with older versions of the kernel and btrfsprogs.

- Related tickets: https://bugzilla.opensuse.org/show_bug.cgi?id=1195224, https://jira.suse.com/browse/PED-63
- Verification run: https://openqa.opensuse.org/tests/2813217

### DIFF
--- a/tests/console/snapper_used_space.pm
+++ b/tests/console/snapper_used_space.pm
@@ -19,7 +19,7 @@ use testapi;
 
 use constant COLUMN_FILTER => "awk -F '|' '{print \$1  \$6}'";    # Filter by columns: # and Used Space
 use constant SUBVOLUME_FILTER => "tail -n4 | sed -n 2,3p | cut -d ' ' -f2";    # Subvolume IDs
-use constant CREATE_BIG_FILE => "touch /big-data && btrfs prop set /big-data compression '' && dd if=/dev/zero of=/big-data bs=1M count=1024";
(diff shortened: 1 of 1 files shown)


pevik:
@Vogtinator not sure why I was pinged, I don't maintain this. But it LGTM.

Adding a context for a reviewer: btrfsprogs change included in v5.18
https://git.kernel.org/pub/scm/linux/kernel/git/kdave/btrfs-progs.git/commit/?id=7781d1a2daa6564c3988dc268eb0680e057027f5

> Currently, if user specifies value 'no' or 'none' on the command line,
> it gets translated to an empty value that is passed to kernel. There was
> a change in kernel 5.14 done by commit 5548c8c6f55b ("btrfs: props:
> change how empty value is interpreted") that changes the behaviour
> in that case.
> 
> The empty value is supposed to mean 'the default value' for any
> property. For compression there is a need to distinguish resetting the
> value and also setting the NOCOMPRESS property. The translation to empty
> value makes that impossible.

pevik (approved):

Vogtinator:
> @Vogtinator not sure why I was pinged, I don't maintain this. But it LGTM.

GitHub suggested you as reviewer and it didn't seem that unlikely to me...
Thanks for the review!

PR #15731, opened 2022-10-18
Revert "hostname: Restart network on MM jobs with supportserver"

dzedro:
Reverts os-autoinst/os-autoinst-distri-opensuse#15729

Have to revert because of this issue.
https://openqa.suse.de/tests/9753120#step/hostname/130
https://openqa.suse.de/tests/9753618#step/hostname/130
Nodes are even pingable but they can't connect via ssh.


dzedro:
image

PR #15733, opened 2022-10-19
Add Leap 15.4 to default persistent journal check

grisu48:
Update the check for the default persistent journal to work with 15.4.

- Verification run: https://duck-norris.qam.suse.de/tests/11026#step/journalctl/91


mloviska (approved):

mloviska (commented):
Oh, taking back. We would break https://openqa.opensuse.org/tests/2808654#step/journalctl/1

grisu48:
This does not make any sense.

PR #15735, opened 2022-10-19
WIP Remove use testapi 'script_output'

Zaoliang:
with revert of PR#15713 we can correct it.
also see poo#117622 for more details


Zaoliang:
got a mistake

PR #15737, opened 2022-10-19
Remove script_output from use line

Zaoliang:
with revert of PR#15713 we can correct this minor error.
See poo#117622 for more details.


Zaoliang:
https://openqa.suse.de/tests/9760290 looks good

foursixnine (approved):

PR #15739, opened 2022-10-19
Create new slem5.3 AY profile

mloviska:
Due to changes in the assets names and product name, and also because of
NM replacing `wicked`, we have to introduce a new profile.
The `<networking>` section is a workaround for
https://bugzilla.suse.com/show_bug.cgi?id=1203934

- Verification run: [sle-micro-5.3-DVD-Updates-x86_64-Build20221018-1-slem_installation_autoyast@64bit](http://kepler.suse.cz/tests/19436#)

### DIFF
--- a/data/autoyast_sle15/autoyast_sle-micro53_updates.xml.ep
+++ b/data/autoyast_sle15/autoyast_sle-micro53_updates.xml.ep
@@ -9,9 +9,9 @@
   <add-on>
     <add_on_products config:type="list">
       <listentry>
-        <name>SUSE-MicroOS-<%= $get_var->('VERSION') %>-Updates</name>
-        <alias>SUSE-MicroOS-<%= $get_var->('VERSION') %>-Updates</alias>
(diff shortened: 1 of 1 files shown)


data/autoyast_sle15/autoyast_sle-micro_updates.xml.ep
```diff
@@ -5,7 +5,6 @@
     <do_registration config:type="boolean">true</do_registration>
     <reg_code>{{SCC_REGCODE}}</reg_code>
     <install_updates config:type="boolean">true</install_updates>
-    <reg_server>{{SCC_URL}}</reg_server>
```
grisu48, line 8: Why is this being removed? Is the registration now being done after the installation?

mloviska: no, SCC_URL is not used in updates testing as we cannot use proxy SCC (developemnt thingy). Mainly the variable is not even defined for update mediums.

mloviska: http://kepler.suse.cz/tests/19436#step/bootloader_start/9 see `regurl`

grisu48 (approved):

PR #15742, opened 2022-10-19
wicked/wlan: Skip auto keymgmt with SAE on wpa_supplicant <2.10

cfconrad:
wpa_supplicant <2.10 does not send SAE in DBus capabilites. Thus wicked will not auto select SAE, which is in my point of view correct.

- Verification run: http://openqa.wicked.suse.de/tests/84944#step/t15_wpa3_personal/10

### DIFF
--- a/lib/wicked/wlan.pm
+++ b/lib/wicked/wlan.pm
@@ -14,8 +14,10 @@ use version_utils qw(is_sle);
 use repo_tools qw(add_qa_head_repo generate_version);
 use utils qw(zypper_call);
 use testapi;
+use version_utils 'check_version';
 
 has wicked_version => undef;
+has wpa_supplicant_version => undef;
 has need_key_mgmt => undef;
 has eap_user => 'tester';
 has eap_password => 'test1234';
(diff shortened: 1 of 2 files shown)


lib/wicked/wlan.pm
```diff
@@ -413,7 +442,8 @@ sub run {
         for my $ifcfg_wlan (__as_config_array($self->ifcfg_wlan())) {
             $self->hostapd_start($hostapd_conf->{config});
 
-            if ($self->check_wicked_version($ifcfg_wlan->{wicked_version})) {
+            if ($self->check_wicked_version($ifcfg_wlan->{wicked_version}) &&
```
asmorodskyi, line 445: and why we need to do same check two times ?

cfconrad: The first one is for "test-global" version specification. The second one for each config.

As example:
If a full tests can be skipped duo to version check: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/wicked/wlan/sut/t07_multi_networks_wpa_psk.pm#L23
or only a single config out of a test: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/wicked/wlan/sut/t03_wpa_psk.pm#L94

asmorodskyi: nice !

PR #15749, opened 2022-10-20
WIP: Fix instance reboot on 15-SP1 and 15-SP2

grisu48:
The shutdown command was passed in a way that 15-SP1 and 15-SP2 could not work with. In addition we fail now if the shutdown command is not handled properly.

- Related ticket: https://progress.opensuse.org/issues/119098
- Verification run: https://duck-norris.qam.suse.de/tests/11053#step/patch_and_reboot/26


grisu48:
VRs are running

ilausuch (approved):

PR #15665, opened 2022-10-08
Add a test suite to create install image for HA

JRivrain:
See https://progress.opensuse.org/issues/116536

gitlab MR https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/400

The new profile is based on https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/data/autoyast_sle15/create_hdd/create_hdd_maintenance.xml.ep, but simplified, as the cloned profile triggered a bug.

VR https://openqa.suse.de/tests/9693308#

PR #15750, opened 2022-10-20
Modify Autoyast profile for HA

JRivrain:
See https://progress.opensuse.org/issues/116536 and https://progress.opensuse.org/issues/116533. In
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15673, it was decided to go for a scripted approach. Therefore it makes sense to use the same profile for those two very similar product versions. This change also brings the following modifications:
 - The profile is placed under autoyast_sle12, since this is a sle12 installation.
 - Some lines inherited from a different profile and useless in this case have been removed.

VRs 
12-sp4 https://openqa.suse.de/tests/overview?distri=sle&version=12-SP4&build=JRivrain%2Fos-autoinst-distri-opensuse%2315750
12-sp5 https://openqa.suse.de/tests/overview?version=12-SP5&build=JRivrain%2Fos-autoinst-distri-opensuse%2315750&distri=sle

MR: https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/409


data/autoyast_sle12/create_hdd/create_hdd_ha_sles.xml.ep
```diff
@@ -305,25 +305,66 @@
         <products config:type="list">
             <product><%= uc $get_var->('SLE_PRODUCT') %></product>
         </products>
-        <!-- Workaround for bsc#1202234: [addon]-release packages are missing after autoyast installation. -->
         <packages config:type="list">
-        % foreach (values %$addons) {
+            <!-- Workaround for bsc#1202234: [addon]-release packages are missing after autoyast installation. -->
```
jknphy, line 309: not sure yet if this would work for sle 15 ha, we could just delete the sp from the name if works for both sle 12.

data/autoyast_sle12/create_hdd/create_hdd_ha_sles.xml.ep
```diff
@@ -305,25 +305,66 @@
...
+            % foreach (values %$addons) {
             <package><%= lc($_->{name}) %>-release</package>
-        % }
+            % }
+            <!-- end of workaround -->
+            <package>sles-release</package>
```
jknphy, line 314: I cannot tell for the commit or any message in the PR if this is some intermediate test (that can be signal in many ways, as we trigger review even if it is WIP).

Anyway this patterns and services are very specific, we don't need to include them in sp4, we could could just create another profile, if they would be a few, perhaps with some `if` would do, but doesn't look like it worth it to have all in the same file (but perhaps this is just a test as I mentioned above).

jknphy:
btw as it was mentioned before many many times the goal is not optimize the AutoYaST profile, cloning will just do the work, optimize it can be done later. For what we can see regarding the squad experience with AutoYaST, seems that takes long to do it sometimes, with many 'try and error', still if you can make it in a timely manner is fine, just think that to have an AutoYaST for this task is just an intermediate step, that we could do with interactive installation, the main goal is to have an standalone migration.

JRivrain:
> btw as it was mentioned before many many times the goal is not optimize the AutoYaST profile, cloning will just do the work, optimize it can be done later. For what we can see regarding the squad experience with AutoYaST, seems that takes long to do it sometimes, with many 'try and error', still if you can make it in a timely manner is fine, just think that to have an AutoYaST for this task is just an intermediate step, that we could do with interactive installation, the main goal is to have an standalone migration.

Using a cloned profile was what I tried first, but then I hit a bug as I had cloned it from the old image from ha migration. So I decided to craft a minimalistic profile in [previous PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15665). 
Meanwhile, https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15673 was merged, using a scripting approach similar to what we have in other maintenance group's jobs.
Now, I could use a cloned profile (~1400 lines) as we determined that the bug does not occur with a current sle12sp5 image, but I prefer to re-use what @lemon-suse did, since it was already merged, also works for me (See VRs), and for the sake of consistency.

jknphy:
> > btw as it was mentioned before many many times the goal is not optimize the AutoYaST profile, cloning will just do the work, optimize it can be done later. For what we can see regarding the squad experience with AutoYaST, seems that takes long to do it sometimes, with many 'try and error', still if you can make it in a timely manner is fine, just think that to have an AutoYaST for this task is just an intermediate step, that we could do with interactive installation, the main goal is to have an standalone migration.
> 
> Using a cloned profile was what I tried first, but then I hit a bug as I had cloned it from the old image from ha migration. So I decided to craft a minimalistic profile in [previous PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15665). Meanwhile, #15673 was merged, using a scripting approach similar to what we have in other maintenance group's jobs. Now, I could use a cloned profile (~1400 lines) as we determined that the bug does not occur with a current sle12sp5 image, but I prefer to re-use what @lemon-suse did, since it was already merged, also works for me (See VRs), and for the sake of consistency.

could you please remove all the patterns and services not needed that YaST will include by default for each product? It is too specific, make me hesitate for merging it tbh...easy to hit bugs.... that is why I would prefer you reuse the same profile but create another file with all this patterns and service to investigate later if worth it. wdyt?

JRivrain:
> could you please remove all the patterns and services not needed that YaST will include by default for this product? the profile is too specific to try to be generalistic with scripting, make me hesitate for merging it tbh...easy to hit bugs.... that is why I would prefer you reuse the same profile but create another file with all this patterns and service to investigate later if worth it. wdyt?

Sure, that was more like a proposal, I thought not including those patterns/packages/services might impact future children jobs.

data/autoyast_sle12/create_hdd/create_hdd_ha_sles.xml.ep
JRivrain: That profile is in autoyast_sle12 it's not for sle 15, effectively it seems to be missing mapping in registration.pm for sle 15, but we can make adjustements in the future if needed

chcao:
The pr is good to me. Just one problem that there's a issue on the VR of 12SP5 on s390x, would you like to take a look at it?
https://openqa.suse.de/tests/overview?version=12-SP5&build=JRivrain%2Fos-autoinst-distri-opensuse%2315750&distri=sle

rakoenig (approved):

data/autoyast_sle12/create_hdd/create_hdd_ha_sles.xml.ep
```diff
@@ -305,21 +305,16 @@
...
+            % foreach (values %$addons) {
             <package><%= lc($_->{name}) %>-release</package>
-        % }
+            % }
+            <!-- end of workaround -->
         </packages>
-        <!-- end of workaround -->
     </software>
     <services-manager>
-        % if ($check_var->('DESKTOP', 'gnome')) {
```
lemon-suse, line 317: I don't know why remove these lines, I prefer the profile should support both 'gnome' and 'textmode'.

JRivrain: The test suite we are replacing is only in text mode, we neve install gnome. Of course it is not a big deal to keep it, but I tried to remove anything we don't need to make the profile as simple as possible.

lemon-suse: Ok, then go ahead. I just think a different way from you, I want to make the profile can support more test to avoid other effort to modify or create profile for other requirement. :)

JRivrain: In Yast job group we've been working in this manner for the past 2-3 years: avoid conditions in code when possible, and instead make tests as specific as possible. It makes code more readable and maintainable, but of course we have more files in the end. We chose to do that because we were spending hours trying to understand bugs in tests, due to too many conditions in code, like "if is_sle 12+ BUT not sle15 OR ARCH=s390x" etc...

JRivrain:
> The pr is good to me. Just one problem that there's a issue on the VR of 12SP5 on s390x, would you like to take a look at it? https://openqa.suse.de/tests/overview?version=12-SP5&build=JRivrain%2Fos-autoinst-distri-opensuse%2315750&distri=sle

Looks like a bug to me, or a problem with registration server: I could not find any other HA job s390x for sle 12-sp5, so I suspect it's a problem on server side that there is just no registration for that product. The regcode for base system is correct and works in other places. I propose to merge since it will got to develpoment group for now, then we can create a ticket to tackle this issue.

chcao (approved):

lemon-suse (approved):

PR #15751, opened 2022-10-21
Add ha autoyast installation profile for sle15

chcao:
This pr is to add HA autoyast installation profile for SLES15SPx, for each arch, I add one profile.

- Related ticket: https://progress.opensuse.org/issues/116509; https://progress.opensuse.org/issues/116557
- Needles: n/a
- Verification run: 
autoyast installation and publish qcows cases:
15sp2: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=25.1&groupid=251
15sp3: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=25.1&groupid=251
15sp4: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=25.1&groupid=251
migration cases:
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=25.1&groupid=251
(see the ha offline cases in this group, and all the ppc64le cases)

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
+++ b/data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
@@ -1,0 +1,326 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <bootloader config:type="map">
+    <global config:type="map">
+      <cpu_mitigations>auto</cpu_mitigations>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <secure_boot>true</secure_boot>
+      <terminal>gfxterm</terminal>
+      <timeout config:type="integer">-1</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>vga=gfx-1024x768x16</xen_kernel_append>
+    </global>
+    <loader_type>grub2-efi</loader_type>
+  </bootloader>
+  <firewall config:type="map">
+    <default_zone>public</default_zone>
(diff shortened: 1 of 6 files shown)


data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
```diff
@@ -0,0 +1,326 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
lemon-suse, line 1: This profile should support part of SLE15+ while not or SLE12+, so suggest to add 'sle15' at least in the names.

jknphy: I think we made a mistake here @lemon-suse there was another folder with sle12, so we didn't need the prefix, like in this case :)

schedule/yast/autoyast/autoyast_ha.yaml
```diff
@@ -0,0 +1,45 @@
+---
+name: create_hdd_ha
+description: >
+  HA autoYaST installation and publish qcow for ha migration standalone
+  cases.
+vars:
+  AUTOYAST: autoyast_sle15/create_hdd/create_hdd_ha_%ARCH%.xml.ep
+  AUTOYAST_PREPARE_PROFILE: '1'
+  DESKTOP: textmode
+  SYSTEM_ROLE: HA_node
```
jknphy, line 10: probably we don't need this for auto-installation.

jknphy:
I can see that there was used conditional scheduling, but I think for those modules to reboot the system is ok for now, we will tackle that in the future. Let's merge all and see result next week in dev group. Good job!

data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
lemon-suse: Yes, you are right, the folder is sle15. While in actual using for the profiles in the folder, we haven't distinguish much.(That is, even a case for SLE12+ we will use profile in the folder also.)

PR #15756, opened 2022-10-21
Apply keyboard based workaround for bsc1189550

ge0r:
Sending "End" and "Home" no longer works for mitigating the scrolling bar issue described in [bsc#1189550](https://bugzilla.suse.com/show_bug.cgi?id=1189550).
A new workaround is to press down twice and the press up, forcing the scroll bar to adjust on the proper item.
The issue can be seen [here](https://openqa.suse.de/tests/9807690#step/select_patterns/53) where the Gnome Desktop Environment pattern is not unselected, because the bar has not scrolled down to make it visible, causing the test to fail at the final module. This was mitigated by an old workaround that no longer works.

- Related ticket: No
- Needles: No needles
- Verification run:  [with workaround](https://openqa.suse.de/tests/9808175#) | [without workaround](https://openqa.suse.de/tests/9808170#)

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -17,8 +17,6 @@ use lockapi;
 use mmapi;
 use Test::Assert 'assert_equals';
 
-my $workaround_bsc1189550_done;
-
 =head1 y2_installbase
 
 C<y2_installbase> - Base class for Yast installer related functionality
@@ -183,9 +181,11 @@ highlight cursor one item down.
 =cut
 
 sub move_down {
+    # Sending 'down' twice, followed by 'up', scrolls to the intended item.
+    wait_screen_change { send_key 'down' } if is_sle('>=15-sp3');
     my $ret = wait_screen_change { send_key 'down' };
-    workaround_bsc1189550() if (!$workaround_bsc1189550_done && is_sle('>=15-sp3'));
     last if (!$ret);    # down didn't change the screen, so exit here
+    wait_screen_change { send_key 'up' } if is_sle('>=15-sp3');
     check12qtbug if check_var('VERSION', '12');
 }
 
@@ -623,12 +623,6 @@ sub post_fail_hook {
         # Collect yast2 installer  strace and gbd debug output if is still running
         $self->save_strace_gdb_output;
     }
-}
-
-sub workaround_bsc1189550 {
-    wait_screen_change { send_key 'end' };
-    wait_screen_change { send_key 'home' };
-    $workaround_bsc1189550_done = 1;
 }
 
 # All steps in the installation are 'fatal'.


lib/y2_installbase.pm
```diff
@@ -183,9 +183,12 @@ highlight cursor one item down.
 =cut
 
 sub move_down {
+    # Sending 'down' twice, followed by 'up', scrolls to the intended item.
+    wait_screen_change { send_key 'down' };
     my $ret = wait_screen_change { send_key 'down' };
-    workaround_bsc1189550() if (!$workaround_bsc1189550_done && is_sle('>=15-sp3'));
     last if (!$ret);    # down didn't change the screen, so exit here
+    wait_screen_change { send_key 'up' };
+    workaround_bsc1189550() if (!$workaround_bsc1189550_done && is_sle('>=15-sp3'));
```
jknphy, line 191: we should not do it for each step, it might cause issues in slower architectures:
we should try to see if previous workaround in single place still works, see previous commit:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13155/files
otherwise find a way to do the trick after certain number of items.

ge0r: You are doing the same steps no? after every move down, you move one more down and up

jknphy: yes, the difference is that is encapsulated in the function that only does it once at the beginning, with the first down.

jknphy: if we don't find anything better and this change needs to be merge to unblock further testing, I would suggest to change line 191 removing variable `$workaround_bsc1189550_done` and replacing the workaround.

jknphy: without video I cannot tell, but it looks like the current workaround might not work anymore because the alt-end and alt-home are lost when they are sent if the selection is on a category:
https://openqa.suse.de/tests/9806816#step/select_patterns/11 (just a theory)
because this up-down trick should be equivalent to that.

ge0r: I am a bit confused on what to do, your second comment now says something else, but anyway, I think I can mark this as resolved, I deleted the old workaround completely and it should not run for older products.

lib/y2_installbase.pm
```diff
@@ -183,9 +183,11 @@ highlight cursor one item down.
 =cut
 
 sub move_down {
+    # Sending 'down' twice, followed by 'up', scrolls to the intended item.
+    wait_screen_change { send_key 'down' } if is_sle('>=15-sp3');
     my $ret = wait_screen_change { send_key 'down' };
-    workaround_bsc1189550() if (!$workaround_bsc1189550_done && is_sle('>=15-sp3'));
     last if (!$ret);    # down didn't change the screen, so exit here
+    wait_screen_change { send_key 'up' } if is_sle('>=15ps-3');
```
jknphy, line 190: typo, 'sp3'

ge0r:
> In general before merging this, I would try to find if alt-end and alt-home was really sent, as it should be equivalent to this, but not running each time for each item, just once at the beginning to recalculate the scroll length. The problem with this workaround is that in slower arch it is too much UI changes, up and down, easy that something would not work plus the huge amount of time to run now this module.

There is not much time difference between running the workaround and not running it.
Actually the difference is negligible, `10m 49s` for the one without! the workaround and `10m 48s` for the one with the workaround. I would argue that it makes no sense to improve it, trying to make it smart to not run for the first 12 items, there is no discernible difference.

punkioudi:
Also the previous failing test now passes: https://openqa.suse.de/tests/9808354#step/first_boot/1 , thanks George for the workaround!

punkioudi (approved):
lgtm!

jknphy:
> one wi

yes, but biggest concern is not the time, it is slower architectures.

jknphy:
> > In general before merging this, I would try to find if alt-end and alt-home was really sent, as it should be equivalent to this, but not running each time for each item, just once at the beginning to recalculate the scroll length. The problem with this workaround is that in slower arch it is too much UI changes, up and down, easy that something would not work plus the huge amount of time to run now this module.
> 
> There is not much time difference between running the workaround and not running it. Actually the difference is negligible, `10m 49s` for the one without! the workaround and `10m 48s` for the one with the workaround. I would argue that it makes no sense to improve it, trying to make it smart to not run for the first 12 items, there is no discernible difference.


yes, but biggest concern is not the time, it is slower architectures, that sometimes keys do not reach the SUT.
the time might be a concern for debugging, let's mention in the bug that the workaround doesn't work anymore.

PR #15761, opened 2022-10-24
Split register_extensions_and_modules to handle separately additional registration

tinawang123:
* Split register_extensions_and_modules to handle separately additional registration
* Replaced '- installation/installation_overview' and '- installation/disable_grub_timeout' to '- installation/security/select_security_module_none'. In order to use libyui test module.
* Remove useless code at 'tests/installation/select_only_visible_patterns_from_top.pm'. As I added this test module in my last PR.

- Related ticket: https://progress.opensuse.org/issues/118993
- Verification run: https://openqa.nue.suse.com/tests/9869593
  https://openqa.nue.suse.com/tests/9869587
  https://openqa.nue.suse.com/tests/9869588#

### DIFF
--- a/schedule/yast/sle/flows/default_sle15sp4.yaml
+++ b/schedule/yast/sle/flows/default_sle15sp4.yaml
@@ -1,49 +1,45 @@
 ---
-name: yast-mru-install-minimal-with-addons
-vars:
-  PATTERNS: base,minimal
-  YUI_REST_API: 1
-schedule:
+# Default ordered sequence of steps for sle15sp4 with maintenance updates
+bootloader:
   - installation/bootloader_start
+setup_libyui:
   - installation/setup_libyui
-  - '{{installsles}}'
+product_selection:
+  - installation/product_selection/install_SLES
+license_agreement:
   - installation/licensing/accept_license
+registration:
   - installation/registration/register_via_scc
+extension_module_selection:
   - installation/module_registration/register_extensions_and_modules
+add_on_product:
   - installation/add_on_product/add_maintenance_repos
   - installation/addon_products_sle
+suggested_partitioning:
   - installation/partitioning/accept_proposed_layout
+clock_and_timezone:
   - installation/clock_and_timezone/accept_timezone_configuration
+local_user:
   - installation/authentication/use_same_password_for_root
   - installation/authentication/default_user_simple_pwd
-  - installation/select_patterns
-  - '{{validate}}'
+software:
+  - installation/select_only_visible_patterns_from_top
+default_systemd_target: []
+booting:
   - installation/bootloader_settings/disable_boot_menu_timeout
-  - installation/installation_overview
-  - installation/disable_grub_timeout
+none_security:
+  - installation/security/select_security_module_none
+installation_settings:
   - installation/launch_installation
+installation:
   - installation/confirm_installation
   - installation/performing_installation/perform_installation
-  - '{{stoptimeout}}'
+stop_timeout_system_reboot: []
+installation_logs:
   - installation/logs_from_installation_system
+confirm_reboot:
   - installation/performing_installation/confirm_reboot
+grub:
   - installation/grub_test
+first_login:
   - installation/first_boot
-conditional_schedule:
-  installsles:
-    ARCH:
-      x86_64:
-        - installation/product_selection/install_SLES
-      aarch64:
-        - installation/product_selection/install_SLES
-  validate:
-    ARCH:
-      x86_64:
-        - installation/installation_settings/validate_default_target
-  stoptimeout:
-    ARCH:
-      aarch64:
-        - installation/performing_installation/stop_timeout_system_reboot_now
-      s390x:
-        - installation/performing_installation/stop_timeout_system_reboot_now
-...
--- a/schedule/yast/maintenance/lvm_thin_provisioning_dev.yaml
+++ b/schedule/yast/maintenance/lvm_thin_provisioning_dev.yaml
@@ -10,6 +10,8 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_registration/register_extensions_and_modules
+  - installation/module_registration/add_additional_regcodes
+  - installation/module_registration/import_untrusted_gnpupg_key
   - installation/add_on_product_installation/add_additional_products
   - installation/add_on_product/skip_install_addons
   - installation/system_role/accept_selected_role_SLES_with_GNOME
--- a/schedule/yast/maintenance/ncurses_interactive_installation_dev.yaml
+++ b/schedule/yast/maintenance/ncurses_interactive_installation_dev.yaml
@@ -12,6 +12,8 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_registration/register_extensions_and_modules
+  - installation/module_registration/add_additional_regcodes
+  - installation/module_registration/import_untrusted_gnpupg_key
   - installation/add_on_product_installation/add_additional_products
   - installation/add_on_product/skip_install_addons
   - installation/system_role/select_role_text_mode
--- a/schedule/yast/maintenance/qam-yast_self_update_dev.yaml
+++ b/schedule/yast/maintenance/qam-yast_self_update_dev.yaml
@@ -11,6 +11,8 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_registration/register_extensions_and_modules
+  - installation/module_registration/add_additional_regcodes
+  - installation/module_registration/import_untrusted_gnpupg_key
   - installation/add_on_product/add_maintenance_repos
   - installation/addon_products_sle
   - installation/partitioning/accept_proposed_layout
--- a/schedule/yast/maintenance/yast-mru-install_dev.yaml
+++ b/schedule/yast/maintenance/yast-mru-install_dev.yaml
@@ -10,6 +10,8 @@ schedule:
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
   - installation/module_registration/register_extensions_and_modules
+  - installation/module_registration/add_additional_regcodes
+  - installation/module_registration/import_untrusted_gnpupg_key
   - installation/add_on_product_installation/add_additional_products
   - installation/add_on_product/skip_install_addons
   - installation/system_role/accept_selected_role_SLES_with_GNOME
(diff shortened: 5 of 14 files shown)


JRivrain (commented):
That seems to work.
Another solution would be to create two modules, one called for example "register_extension_and_modules" containing this 
`    my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
    $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
    save_screenshot;`
the second named for example "add_separate_registration_code" containing 
`   # when some module (e.g. workstation extension) requires registration, provide separate code
    my $timeout = 60 * get_var('TIMEOUT_SCALE', 1);
    my $regcode = get_var('SCC_REGCODE_WE');
    $testapi::distri->get_module_regcode()->add_separate_registration_code($regcode, $timeout);
    save_screenshot;
    # confirm to trust the untrusted GPG key
    $testapi::distri->get_module_regcode()->trust_gnupg_key();`

Then, schedule the second one only where it is necessary.

schedule/yast/maintenance/yast-mru-install-minimal-with-addons.yaml
```diff
@@ -1,26 +1,25 @@
...
   - installation/addon_products_sle
   - installation/partitioning/accept_proposed_layout
   - installation/clock_and_timezone/accept_timezone_configuration
   - installation/authentication/use_same_password_for_root
   - installation/authentication/default_user_simple_pwd
-  - installation/select_patterns
+  - installation/select_only_visible_patterns_from_top
   - '{{validate}}'
   - installation/bootloader_settings/disable_boot_menu_timeout
-  - installation/installation_overview
```
JRivrain, line 22: This change looks not directly related, would be good to mention it in the commit.

tinawang123: Changed to use libyui test modules. 
'- installation/installation_overview' and '- installation/disable_grub_timeout' don't use libyui

tinawang123:
> That seems to work fine. Another solution would be to create two modules, one called for example "register_extension_and_modules" containing this ` my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS'))); $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]); save_screenshot;` the second named for example "add_separate_registration_code" containing ` # when some module (e.g. workstation extension) requires registration, provide separate code my $timeout = 60 * get_var('TIMEOUT_SCALE', 1); my $regcode = get_var('SCC_REGCODE_WE'); $testapi::distri->get_module_regcode()->add_separate_registration_code($regcode, $timeout); save_screenshot; # confirm to trust the untrusted GPG key $testapi::distri->get_module_regcode()->trust_gnupg_key();`
> 
> Then, schedule the second one only where it is necessary. I think it's probably what @jknphy had in mind when he changed the title of the ticket to "Split register_extensions_and_modules to handle separately additional registration".

If use this solution, It will impact our original test cases. I need updated all test cases' yaml file which use 'register_extensions_and_modules'. I think it may impact a lot.

JRivrain:
> If use this solution, It will impact our original test cases. I need updated all test cases' yaml file which use 'register_extensions_and_modules'. I think it may impact a lot.

Here is the whole list:
yast/maintenance/lvm_thin_provisioning_dev.yaml
yast/maintenance/ncurses_interactive_installation_dev.yaml
yast/maintenance/qam-yast_self_update_dev.yaml
yast/maintenance/yast-mru-install-minimal-with-addons.yaml
yast/maintenance/yast-mru-install_dev.yaml
yast/qam-yast_self_update.yaml
yast/qam-yast_self_update_libyui.yaml

Can be all modified at once with this 
```
cd /var/lib/openqa/share/tests/sle/schedule
for i in `grep -r register_extensions_and_modules |cut -d: -f1`; do sed -ie 's:- installation/module_registration/register_extensions_and_modules:- installation/module_registration/register_extensions_and_modules\n  - installation/module_registration/add_separate_registration_code:g' $i; done
```
This is what I use when there is a need to change many schedules at once, some may prefer some python script for example.

Just a suggestion though, we can go with your solution for now, it also works.

JRivrain:
LGTM, but I would prefer to have more explanations in the commit, as there are more things in the change than what is described.

tinawang123:
> LGTM, but I would prefer to have more explanations in the commit, as there are more things in the change than what is described.

Updated description.

rakoenig (approved):
LGTM

jknphy:
The ticket specifically mention to split the existing test module, please proceed as agreed.
Another point, avoid to use conditional schedule from now on and create separate files as needed, we will improve the situation with yaml based on default files in the future creating different default files for each architecture.

There is still a simpler approach which doesn't require so many schedule files but it is less transparent, where we don't need to split the module, we can just grep for the codes that we now on advance that they will need additional registration code and loop for typing them.

For this screens we rely on setting value to add the list of addon, which might make sense, given that ltss and other values, like python2 or python3 version are changing and then we don't worry among products, to have separate test module for registration might be more clear, like it was described in the ticket, but then that will not be dynamic, although that part is actually not expected to be dynamic anyway, so both solution I think are fine, but your solution not changing existing module I think will not be useful for future cases where imagine, there are two registration codes to be added for example.

tests/installation/module_registration/register_module_regpage.pm
```diff
@@ -0,0 +1,41 @@
...
+
+# Summary: After register Application module
+#          in "Extension and Module Selection" dialog
+#          has register page
+# Maintainer: QA SLE YaST team <qa-sle-yast@suse.de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use registration;
```
jknphy, line 14: let's import only the method used here.

tests/installation/module_registration/register_module_regpage.pm
```diff
@@ -0,0 +1,41 @@
...
+    'sle-we' => get_var('SCC_REGCODE_WE'),
+    'sle-module-live-patching' => get_var('SCC_REGCODE_LIVE'),
+    'sle-live-patching' => get_var('SCC_REGCODE_LIVE'),
+    'SLES-LTSS' => get_var('SCC_REGCODE_LTSS'),
+    'SUSE-Linux-Enterprise-RT' => get_var('SCC_REGCODE_RT'),
+    ESPOS => get_var('SCC_REGCODE_ESPOS'),
+);
+
+sub run {
+    my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
```
jknphy, line 29: you just need after this line a `grep /we|ha| etc./ @scc_addons` which will give you a list empty if the job doesn't have it or a list with the elements that requires additional regcode. Then you don't need to get the full name at all, your hash can have as keys the addons keys, `we, ha, etc.`

tests/installation/module_registration/register_module_regpage.pm
```diff
@@ -0,0 +1,41 @@
...
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use registration;
+use testapi qw(save_screenshot get_var get_required_var);
+
+our %ADDONS_REGCODE = (
+    'sle-ha' => get_var('SCC_REGCODE_HA'),
+    'sle-ha-geo' => get_var('SCC_REGCODE_GEO'),
```
jknphy, line 19: I'm not sure if it is necessary at this point to gather all the possible regcodes, we are using libyui-rest-api only for sle-15-sp{3,4}, perhaps it is better to not give false hopes that this would work and we can do this later on demand.  we just need 'we' for now? right? a hash with one for now?

schedule/yast/maintenance/yast_mru_install_minimal_with_addons_aarch64.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
jknphy, line 1: `*_aarch64_dev.yaml` we should keep that naming before deploying, it would help to distinguish them.

schedule/yast/maintenance/yast_mru_install_minimal_with_addons_aarch64.yaml
```diff
@@ -0,0 +1,11 @@
...
+name: yast_mru_install_minimal_with_addons_aarch64
+vars:
+  PATTERNS: base,enhanced_base
+  YUI_REST_API: 1
+schedule:
+  product_selection:
+    - installation/product_selection/install_SLES
```
jknphy, line 8: I don't get this line, yaml schedule based on default we can apply later, I would suggest just not use conditional and have 3 different files reflecting the differences with a list of test modules.
https://progress.opensuse.org/issues/119617

tests/installation/module_registration/register_module_regpage.pm
tinawang123: Do we need live patching to do register at here?

schedule/yast/maintenance/yast_mru_install_minimal_with_addons_aarch64.yaml
tinawang123: As s390x yaml file don't need this test module.
At default yaml file, I set product_selection: []

JRivrain:
> The ticket specifically mention to split the existing test module, please proceed as agreed. 

Well, the module was splitted. This statement can be interpreted in various ways, the proposed code works and is a possible (and correct) interpretation of what is asked in the ticket, just not what as expected.

jknphy:
> > The ticket specifically mention to split the existing test module, please proceed as agreed.
> 
> Well, the module was splitted in two modules. This statement can be interpreted in various ways, the proposed code worked and is a possible (and correct) interpretation of what is asked in the ticket, just not what was expected.

literally one part of the existing module was extracted in another module, splitting the existing one is to break it into pieces, which was not the case, the existing one remained exactly the same when I did this comment, an additional module was created excluding the part of the registration. btw this was discussed offline with the author to avoid missunderstanding, but I don't find better verb than 'split'. You cannot split in two modules if the original one remains the same, right?

schedule/yast/sle/flows/default_sle15sp4.yaml
```diff
@@ -1,49 +1,45 @@
...
-  - installation/select_patterns
-  - '{{validate}}'
+software:
+  - installation/select_only_visible_patterns_from_top
+default_systemd_target: []
+booting:
   - installation/bootloader_settings/disable_boot_menu_timeout
-  - installation/installation_overview
-  - installation/disable_grub_timeout
+none_security:
```
jknphy, line 30: just `security` and for `software` I don't think we need to be so specific, in a default file we should include default patterns, so this change should be at the level of the schedule, but good for now, let's deal with that in follo-up ticket.

PR #15762, opened 2022-10-24
Increase ansible-galaxy install timeout in ansible test module

pdostal:
This is quick fix for the following failure:
- Related failures: [ansible-galaxy install](https://openqa.suse.de/tests/9798421), [ansible-playbook](https://openqa.suse.de/tests/9798377#step/ansible/102)
- Verification run: TBD


pdostal:
We also need to extend the `MAX_JOB_TIME` from `2h` to `3h` for the test suite `mau-extratests2`.

PR #15767, opened 2022-10-24
Make ssh_interactive_leave more resilient

grisu48:
Prevent a test failure in case the serial console swallows character while probing true. This is necessary because script_run would make the test fail on a timeout, instead of retrying.

- Related failure: https://openqa.suse.de/tests/9796056#step/ssh_interactive_end/4
- Verification run: https://duck-norris.qam.suse.de/tests/11069

### DIFF
--- a/lib/publiccloud/ssh_interactive.pm
+++ b/lib/publiccloud/ssh_interactive.pm
@@ -88,7 +88,7 @@ sub ssh_interactive_leave {
         send_key 'ctrl-c';
         send_key 'ctrl-c';
         send_key 'ret';
-        last if (script_run("true") == 0);
+        last if (script_run("true", timeout => 5, die_on_timeout => 0) == 0);
         sleep 5;    # some cool down after a failed attempt
     }
     die "tunnel-console is not functional" if ($retries <= 0);


ilausuch (approved):

lib/publiccloud/ssh_interactive.pm
```diff
@@ -88,7 +88,11 @@ sub ssh_interactive_leave {
         send_key 'ctrl-c';
         send_key 'ctrl-c';
         send_key 'ret';
-        last if (script_run("true") == 0);
+        # Catch exception and retry. This is necessary to catch the "command 'true' timed out" in case the serial console is busy.
+        eval {
+            last if (script_run("true") == 0);
```
asmorodskyi, line 93: maybe we need to set also smaller waiting time in script_run ?

asmorodskyi: also script_run has  `die_on_timeout` parameter exactly for such cases see https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L940

grisu48: We can. What would be a good one?

grisu48: > also script_run has `die_on_timeout` parameter exactly for such cases see https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm#L940

`With C<die_on_timeout> -1 (default) a warning will be printed to log.` - this does not explain the failures we are seeing: https://openqa.suse.de/tests/9796056#step/ssh_interactive_end/4 but we can try. Using `eval` here is anyways very ugly.

grisu48: amended and [VR](https://duck-norris.qam.suse.de/tests/11068#live) restarted

asmorodskyi (approved):
LGTM, feel free to merge when VRs will pass

PR #15771, opened 2022-10-24
Call Trento support scripts and collect the logs

mpagot:
Call trento_support and dump_scenario_from_k8 scripts on the Trento VM  as part of the k8s_logs openQA lib function

- Verification run: https://openqaworker15.qa.suse.cz/tests/58103 https://openqaworker15.qa.suse.cz/tests/58152

Here trento_support called remotely: https://openqaworker15.qa.suse.cz/tests/58103#step/deploy_trento/121 It runs with an error but produce the archive in any case:

```
Error 1 during 'base' collection
COMPRESSING OUTPUT
```

Here dump_scenario called remotely https://openqaworker15.qa.suse.cz/tests/58103#step/deploy_trento/139

It works fine but, due to the moment of the test where it is called, it produces a valid but empty json

Result artifacts are like:
- https://openqaworker15.qa.suse.cz/tests/58103/file/deploy_trento-2022-10-25_0920_support.txt.tar.gz
- https://openqaworker15.qa.suse.cz/tests/58103/file/deploy_trento-discarded_discovery_events.json

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -66,6 +66,7 @@ our @EXPORT = qw(
   cypress_test_exec
   cypress_log_upload
   k8s_logs
+  trento_support
 );
 
 # Exported constants
@@ -169,7 +170,7 @@ sub get_trento_deployment {
 
         # Download the release archive
         my $download_api_url = "$repo_url/repository/archive.tar.gz?sha=$commit_id";
-        my $download_cmd = $curl_cmd .
+        my $download_cmd = $curl_cmd . ' -f' .
           " \"$download_api_url\"" .
           " --output $ver_artifact ";
         assert_script_run($download_cmd);
@@ -283,7 +284,7 @@ Deploy a SAP Landscape using a previously configured qe-sap-deployment
 sub deploy_qesap {
     my $ret = qesap_execute(cmd => 'terraform', verbose => 1, timeout => 1800);
     die "'qesap.py terraform' return: $ret" if ($ret);
-    $ret = qesap_execute(cmd => 'ansible', verbose => 1, timeout => 1800);
+    $ret = qesap_execute(cmd => 'ansible', verbose => 1, timeout => 3600);
     die "'qesap.py ansible' return: $ret" if ($ret);
     my $inventory = qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER'));
     upload_logs($inventory);
@@ -297,7 +298,7 @@ Destroy the qe-sap-deployment SAP Landscape
 sub destroy_qesap {
     my $ret = qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300);
     die "'qesap.py ansible -d' return: $ret" if ($ret);
-    $ret = qesap_execute(cmd => 'terraform', cmd_options => '-d', verbose => 1, timeout => 300);
+    $ret = qesap_execute(cmd => 'terraform', cmd_options => '-d', verbose => 1, timeout => 3600);
     die "'qesap.py terraform -d' return: $ret" if ($ret);
 }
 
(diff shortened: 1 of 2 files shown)


jankohoutek:
LGTM

jankohoutek (approved):
LGTM

rtorrero (approved):
Not an expert in perl by any means, but looks good :+1: . Only thing is that I noticed a few commented-put lines that you may what to remove?

lib/trento.pm
```diff
@@ -518,6 +519,44 @@ sub k8s_logs {
...
+      SSH_KEY, VM_USER, $machine_ip;
+    my $cmd = join(' ',
+        'sudo',
+        'KUBECONFIG=/etc/rancher/k3s/k3s.yaml', 'PATH=${PATH}:/home/' . VM_USER . '/bin/',
+        './trento-support.sh',
+        '--output', 'file-tgz',
+        '--collect', 'all');
+    script_run(az_vm_ssh_cmd($cmd, $machine_ip));
+    script_run("$scp_cmd:'*.tar.gz' $log_dir");
+    my @files = split("\n", script_output('ls -1 ' . $log_dir . '*.tar.gz | cat'));
```
alvarocarvajald, line 549: Is `| cat` really necessary here? I thought it was not needed due to `ls -1`

lib/trento.pm
```diff
@@ -518,6 +519,44 @@ sub k8s_logs {
...
+    foreach my $this_file (@files) {
+        upload_logs($this_file, failok => 1);
+    }
+
+    script_run(az_vm_ssh_cmd('./dump_scenario_from_k8.sh -n goofy', $machine_ip));
```
alvarocarvajald, line 554: `goofy` is hard coded here. Shouldn't you use `CYPRESS_IMAGE_TAG()` insead?

lib/trento.pm
```diff
@@ -518,6 +519,44 @@ sub k8s_logs {
...
+    script_run("$scp_cmd:'scenarios/goofy/*.json' $log_dir");
+    @files = split("\n", script_output('ls -1 ' . $log_dir . '*.json | cat'));
+    foreach my $this_file (@files) {
+        upload_logs($this_file, failok => 1);
+    }
```
alvarocarvajald, line 559: This `for` block is exactly the same as the one above. I would vote to not have it repeated.

Since you're already using a list with `@files` you can do:
```
my @files = split("\n", script_output('ls -1 ' . $log_dir . '*.tar.gz'));
push @files, split("\n", script_output('ls -1 ' . $log_dir . '*.json'));
foreach my $this_file (@files) {
    upload_logs($this_file, failok => 1);
}
```
or better yet:
```
foreach my $this_file (split("\n", script_output("ls -1 $log_dir *.tar.gz *.json")))
    upload_logs($this_file, failok => 1);
}
```
Fewer calls to the API are always better.

lib/trento.pm
```diff
@@ -518,6 +519,44 @@ sub k8s_logs {
...
+    script_run(az_vm_ssh_cmd($cmd, $machine_ip));
+    script_run("$scp_cmd:'*.tar.gz' $log_dir");
+    my @files = split("\n", script_output('ls -1 ' . $log_dir . '*.tar.gz | cat'));
+    foreach my $this_file (@files) {
+        upload_logs($this_file, failok => 1);
+    }
+
+    script_run(az_vm_ssh_cmd('./dump_scenario_from_k8.sh -n goofy', $machine_ip));
+    script_run("$scp_cmd:'scenarios/goofy/*.json' $log_dir");
+    @files = split("\n", script_output('ls -1 ' . $log_dir . '*.json | cat'));
```
alvarocarvajald, line 556: Same question about the `| cat`

lib/trento.pm
mpagot: Inspired by https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/ff698892d36ba624a7e616363b5dc599f219aae6/lib/opensusebasetest.pm#L1283

I'm testing without  cat

alvarocarvajald: Weird. I would usually add a `| cat` to an `ls` command if I want to treat the output with a file per line, i.e.:
```
acarvajal@linux-mkji:~/tmp> ls
hola  mundo  robot
acarvajal@linux-mkji:~/tmp> ls | cat
hola
mundo
robot
```
But `ls -1` already does that AFAICT:
```
acarvajal@linux-mkji:~/tmp> ls -1
hola
mundo
robot
```
The man page even says:
```
-1     list one file per line.  Avoid '\n' with -q or -b
```

lib/trento.pm
alvarocarvajald: Not needed. I thought it was the same thing.

alvarocarvajald (approved):
LGTM

PR #15720, opened 2022-10-17
Fix mau-extratests2 yaml schedule for SLE 15-SP4

pdostal:
Schedule **15-SP4** -specific tests from **15-SP3** in `mau-extratests2.yaml`

- Verification run: [pdostal-server.suse.cz/t2746](https://pdostal-server.suse.cz/tests/2746)

PR #15779, opened 2022-10-25
Revert ansible and valgrind from 15-sp4

michaelgrifalconi:
Revert part of #15720 since these tests do not work on 15-sp4 and are blocking update approval.

- Related ticket: https://progress.opensuse.org/issues/119278 and https://progress.opensuse.org/issues/119377
- Needles: none
- Verification run: none


a-kpappas:
AFAICT these tests have been failing only on some architectures? If that's the case, they need to be unscheduled only on those, and for the sle versions they are failing at.

PR #15781, opened 2022-10-25
Fix incompability of AY on 15sp3

b10n1k:
No idea why but those conditions makes the autoyast valid and the installation continues. As a quick fix for incidents, i havent test the profile against other versions. i just kept it _untouched_ for anything else

- Related ticket: no ticket but this is a quick fix for incidents https://openqa.suse.de/tests/9808348#step/installation/4
- Verification run: http://aquarius.suse.cz/tests/13580


b10n1k:
@Avinesh FYI

PR #15784, opened 2022-10-26
[HA/SLES4SAP] Deployment tests for qesapdeployment project

lpalovsky:
Introduces HA/SLES4SAP tests and library for PC infra deployment and 
cleanup using qe-sap-deployment project. Has ability to execute 
ansible and terraform deployment in separate steps with behavior 
controlled via OpenQA variables. Cleanup is done with separate test 
module and included in post run/fail hooks.

- Related ticket: https://jira.suse.com/browse/TEAM-6852
- Needles: none
- Verification run: 
Single VM, no cluster: https://mordor.suse.cz/tests/5506
HANA Cluster:  https://mordor.suse.cz/tests/5505
With HANASR test: https://mordor.suse.cz/tests/5512#
  - HanaSR test fails due to unrelated test code issue. However test is able to interact and execute commands on SUT.

### DIFF
--- a/data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
+++ b/data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
@@ -1,0 +1,38 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Generic yaml template for use with qe-sap-deployment project: https://github.com/SUSE/qe-sap-deployment
+#   Settings are meant to be controlled via OpenQA variables and managed by test:
+#   tests/sles4sap/publiccloud/qesap_ansible.pm
+
+provider: 'azure'
+apiver: 2
+terraform:
+  provider: 'azure'
+  variables:
+    # GENERAL VARIABLES #
+    az_region: '%PUBLIC_CLOUD_REGION%'
+    admin_user: 'cloudadmin'
+    public_key : '~/.ssh/id_rsa.pub'
+    private_key: '~/.ssh/id_rsa'
+    sles4sap_uri: 'https://%STORAGE_ACCOUNT_NAME%.blob.core.windows.net/sle-images/%SLE_IMAGE%'
+    hana_os_major_version: '%HANA_OS_MAJOR_VERSION%'
+    storage_account_name: '%STORAGE_ACCOUNT_NAME%'
+    storage_account_key: '%STORAGE_ACCOUNT_KEY%'
+
+    # HANA
+    hana_count: '%NODE_COUNT%'
+    hana_ha_enabled: '%HA_CLUSTER%'
+    vm_size: '%PUBLIC_CLOUD_INSTANCE_TYPE%'
+    hana_cluster_fencing_mechanism: '%FENCING_MECHANISM%'
+
+ansible:
+  hana_urls:
+    - '%HANA_SAR%'
+    - '%HANA_CLIENT_SAR%'
+    - '%HANA_SAPCAR%'
+  destroy:
+    - deregister.yaml
+
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -30,6 +30,7 @@ package qesapdeployment;
 use strict;
 use warnings;
 use Carp qw(croak);
+use YAML::PP;
 use utils 'file_content_replace';
 use testapi;
 use Exporter 'import';
@@ -55,6 +56,7 @@ our @EXPORT = qw(
   qesap_execute
   qesap_yaml_replace
   qesap_ansible_cmd
+  qesap_create_ansible_section
 );
 
 =head1 DESCRIPTION
@@ -73,7 +75,8 @@ sub qesap_get_file_paths {
     $paths{qesap_conf_filename} = get_required_var('QESAP_CONFIG_FILE');
     $paths{deployment_dir} = get_var('QESAP_DEPLOYMENT_DIR', get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment'));
     $paths{terraform_dir} = get_var('PUBLIC_CLOUD_TERRAFORM_DIR', $paths{deployment_dir} . '/terraform');
-    $paths{qesap_conf_trgt} = $paths{deployment_dir} . "/scripts/qesap/" . $paths{qesap_conf_filename};
+    $paths{qesap_conf_trgt} = $paths{deployment_dir} . '/scripts/qesap/' . $paths{qesap_conf_filename};
+    $paths{qesap_conf_src} = data_url('sles4sap/qe_sap_deployment/' . $paths{qesap_conf_filename});
     return (%paths);
 }
 
@@ -85,6 +88,58 @@ sub qesap_get_file_paths {
 sub qesap_create_folder_tree {
     my %paths = qesap_get_file_paths();
     assert_script_run("mkdir -p $paths{deployment_dir}", quiet => 1);
+}
+
+=head3 qesap_get_variables
+
+    Scans yaml config for '%OPENQA_VARIABLE%' placeholders and searches for values in OpenQA defined variables.
+    Returns hash with openqa variable key/value pairs.
+=cut
+
+sub qesap_get_variables {
+    my %paths = qesap_get_file_paths();
+    my $yaml_file = $paths{'qesap_conf_src'};
+    my %variables;
+    my $grep_cmd = "grep -v '#' | grep -oE %[A-Z0-9_]*% | sed s/%//g";
+    my $cmd = join(" ", "curl -s -fL", $yaml_file, "|", $grep_cmd);
+
+    for my $variable (split(" ", script_output($cmd))) {
+        $variables{$variable} = get_required_var($variable);
+    }
+    return \%variables;
+}
+
+=head3 qesap_create_ansible_section
+
+    Writes "ansible" section into yaml config file.
+    $args{ansible_section} defines section(key) name.
+    $args{section_content} defines content of names section.
+        Example:
+            @playbook_list = ("pre-cluster.yaml", "cluster_sbd_prep.yaml");
+            qesap_create_ansible_section(ansible_section=>'create', section_content=>\@playbook_list);
+
+=cut
+
+sub qesap_create_ansible_section {
+    my (%args) = @_;
+    my $ypp = YAML::PP->new;
+    my $section = $args{ansible_section} // 'no_section_provided';
+    my $content = $args{section_content} // {};
+    my %paths = qesap_get_file_paths();
+    my $yaml_config_path = $paths{qesap_conf_trgt};
+
+    assert_script_run("test -e $yaml_config_path", fail_message => "Yaml config file '$yaml_config_path' does not exist.");
+
+    my $raw_file = script_output("cat $yaml_config_path");
+    my $yaml_data = $ypp->load_string($raw_file);
+
+    $yaml_data->{ansible}{$section} = $content;
+
+    # write into file
+    my $yaml_dumped = $ypp->dump_string($yaml_data);
+    save_tmp_file($paths{qesap_conf_filename}, $yaml_dumped);
+    assert_script_run('curl -v -fL ' . autoinst_url . "/files/" . $paths{qesap_conf_filename} . ' -o ' . $paths{qesap_conf_trgt});
+    return;
 }
 
 =head3 qesap_pip_install
@@ -292,17 +347,20 @@ sub qesap_get_terraform_dir {
 
 sub qesap_prepare_env {
     my (%args) = @_;
-    my $variables = $args{openqa_variables};
+    my $variables = $args{openqa_variables} ? $args{openqa_variables} : qesap_get_variables();
     my $provider = $args{provider};
     my %paths = qesap_get_file_paths();
-    my $qesap_conf_src = "sles4sap/qe_sap_deployment/" . $paths{qesap_conf_filename};
 
-    qesap_create_folder_tree();
-    qesap_get_deployment_code();
-    qesap_pip_install();
+    # Option to skip straight to configuration
+    unless ($args{only_configure}) {
+        qesap_create_folder_tree();
+        qesap_get_deployment_code();
+        qesap_pip_install();
 
-    record_info("QESAP yaml", "Preparing yaml config file");
-    assert_script_run('curl -v -fL ' . data_url($qesap_conf_src) . ' -o ' . $paths{qesap_conf_trgt});
+        record_info("QESAP yaml", "Preparing yaml config file");
+        assert_script_run('curl -v -fL ' . $paths{qesap_conf_src} . ' -o ' . $paths{qesap_conf_trgt});
+    }
+
     qesap_yaml_replace(openqa_variables => $variables);
     push(@log_files, $paths{qesap_conf_trgt});
 
@@ -313,7 +371,7 @@ sub qesap_prepare_env {
     my $exec_rc = qesap_execute(cmd => 'configure', verbose => 1);
     push(@log_files, $hana_vars) if (script_run("test -e $hana_vars") == 0);
     qesap_upload_logs(failok => 1);
-    die if $exec_rc != 0;
+    die("Qesap deployment returned non zero value during 'configure' phase.") if $exec_rc;
     return;
 }
 
--- a/lib/sles4sap_publiccloud_basetest.pm
+++ b/lib/sles4sap_publiccloud_basetest.pm
@@ -1,0 +1,71 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Base test providing cleanup, post fail and post run hooks for tests using qe-sap-deployment project.
+# https://github.com/SUSE/qe-sap-deployment
+
+package sles4sap_publiccloud_basetest;
+
+use Mojo::Base 'publiccloud::basetest';
+use strict;
+use warnings FATAL => 'all';
+use Exporter 'import';
+use testapi;
+use qesapdeployment;
+
+our @EXPORT = qw(cleanup);
+
+
+sub cleanup {
+    my ($self) = @_;
+    # Do not run destroy if already executed
+    return if ($self->{cleanup_called});
+    $self->{cleanup_called} = 1;
+
+    for my $command ("ansible", "terraform") {
+
+        # Skip cleanup if ansible inventory is not present (deployment could not have been done without it)
+        my $inventory_check_cmd = join(" ", ("test", "-f", qesap_get_inventory()));
+        next if script_run($inventory_check_cmd) == 0;
+
+        record_info("Cleanup", "Executing $command cleanup");
+        # 3 attempts for both terraform and ansible cleanup
+        for (1 .. 3) {
+            my $cleanup_cmd_rc = qesap_execute(verbose => "--verbose", cmd => $command, cmd_options => "-d", timeout => 1200);
+            if ($cleanup_cmd_rc == 0) {
+                diag(ucfirst($command) . " cleanup attempt # $_ PASSED.");
+                record_info("Clean $command", ucfirst($command) . " cleanup PASSED.");
+                last;
+            }
+            else {
+                diag(ucfirst($command) . " cleanup attempt # $_ FAILED.");
+                sleep 10;
+            }
+            record_info("Cleanup FAILED", "Cleanup $command FAILED", result => "fail") if $_ == 3 && $cleanup_cmd_rc;
+            $self->{result} = "fail" if $_ == 3 && $cleanup_cmd_rc;
+        }
+    }
+    record_info("Cleanup finished");
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    if (get_var("PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE")) {
+        diag("Skip post fail", "Variable 'PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE' defined.");
+        return;
+    }
+    $self->cleanup();
+}
+
+sub post_run_hook {
+    my ($self) = @_;
+    if ($self->test_flags()->{publiccloud_multi_module} or get_var("PUBLIC_CLOUD_NO_CLEANUP")) {
+        diag("Skip post run", "Skipping post run hook. \n Variable 'PUBLIC_CLOUD_NO_CLEANUP' defined or test_flag 'publiccloud_multi_module' active");
+        return;
+    }
+    $self->cleanup();
+}
+
+1;
--- a/schedule/sles4sap/publiccloud_hanasr.yml
+++ b/schedule/sles4sap/publiccloud_hanasr.yml
@@ -1,0 +1,15 @@
+---
+name: publiccloud_hanasr
+description: |
+  Run HANASR tests on public cloud instances using qe-sap-deployment project:
+    - deploy
+    - test
+    - destroy
+vars:
+  PUBLIC_CLOUD_RESOURCE_GROUP: 'HanaSR'
+  TEST_CONTEXT: 'OpenQA::Test::RunArgs'
+schedule:
+  - boot/boot_to_desktop
+  - sles4sap/publiccloud/qesap_terraform
+  - sles4sap/publiccloud/qesap_ansible
+  - sles4sap/publiccloud/qesap_cleanup
--- a/t/09_qesapdeployment.t
+++ b/t/09_qesapdeployment.t
@@ -8,6 +8,10 @@ use List::Util qw(any);
 use testapi 'set_var';
 use qesapdeployment;
 set_var('QESAP_CONFIG_FILE', 'MARLIN');
+set_var('BACKEND', 'qemu');
+set_var('AUTOINST_URL_HOSTNAME', 'praisethesun');
+set_var('QEMUPORT', 9000);
+set_var('JOBTOKEN', "solaire");
 
 subtest '[qesap_get_inventory] upper case' => sub {
     my $inventory_path = qesap_get_inventory('NEMO');
--- a/tests/sles4sap/publiccloud/qesap_ansible.pm
+++ b/tests/sles4sap/publiccloud/qesap_ansible.pm
@@ -1,0 +1,56 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Execute ansible deployment using qe-sap-deployment project.
+# https://github.com/SUSE/qe-sap-deployment
+
+use base 'sles4sap_publiccloud_basetest';
+use strict;
+use warnings;
+use testapi;
+use Mojo::File 'path';
+use publiccloud::utils;
+use sles4sap_publiccloud;
+use qesapdeployment;
+use serial_terminal 'select_serial_terminal';
+
+sub test_flags {
+    return {fatal => 1, publiccloud_multi_module => 1};
+}
+
+sub run {
+    my ($self, $run_args) = @_;
+    select_serial_terminal;
+    my $ha_enabled = get_required_var("HA_CLUSTER") =~ /false|0/i ? 0 : 1;
+    my $instances = $run_args->{instances};
+
+    die("Ansible deploymend FAILED. Check 'qesap*' logs for details.") if qesap_execute(cmd => 'ansible', timeout => 3600, verbose => 1) > 0;
+    record_info("FINISHED", "Ansible deployment process finished successfully.");
+    return unless $ha_enabled;
+
+    # In canse of Hana/HA deployment, detect initial primary and secondary site location
+    record_info("HANA chk", "Ansible deployment process finished successfully.");
+    foreach my $instance (@$instances) {
+        $self->{my_instance} = $instance;
+        my $instance_id = $instance->{'instance_id'};
+        # Skip instances without HANA db
+        next if ($instance_id !~ m/vmhana/);
+        $self->wait_for_sync();
+
+        # Define initial state for both sites
+        # Site A is always PROMOTED (Master node) after deployment
+        my $resource_output = $self->run_cmd(cmd => "crm status full", quiet => 1); record_info("crm out", $resource_output);
+        my $master_node = $self->get_promoted_hostname();
+        $run_args->{site_a} = $instance if ($instance_id eq $master_node);
+        $run_args->{site_b} = $instance if ($instance_id ne $master_node);
+        record_info("Instances:", "Detected HANA instances:
+        Site A: $run_args->{site_a}{instance_id}
+        Site B: $run_args->{site_b}{instance_id}") if ($instance_id eq $master_node);
+    }
+
+    return 1;
+}
+
+1;
--- a/tests/sles4sap/publiccloud/qesap_cleanup.pm
+++ b/tests/sles4sap/publiccloud/qesap_cleanup.pm
@@ -1,0 +1,22 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Cleanup test meant to be used with multimodule setup with qe-sap-deployment project.
+# https://github.com/SUSE/qe-sap-deployment
+
+use base 'sles4sap_publiccloud_basetest';
+use sles4sap_publiccloud_basetest;
+use strict;
+use warnings FATAL => 'all';
+use testapi;
+
+
+sub run {
+    my ($self) = @_;
+    return if get_var("PUBLIC_CLOUD_NO_CLEANUP");
+    $self->cleanup();
+}
+
+1;
--- a/tests/sles4sap/publiccloud/qesap_terraform.pm
+++ b/tests/sles4sap/publiccloud/qesap_terraform.pm
@@ -1,0 +1,152 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Deploy public cloud infrastructure using terraform using qe-sap-deployment project.
+# https://github.com/SUSE/qe-sap-deployment
+
+# Available OpenQA parameters:
+# HA_CLUSTER - Enables HA/Hana cluser scenario
+# NODE_COUNT - number of nodes to deploy. Needs to be >1 for cluster usage.
+# PUBLIC_CLOUD_INSTANCE_TYPE - VM size, sets terraform 'vm_size' parameter
+# USE_SAPCONF - (true/false) set 'false' to use saptune
+# HANA_OS_MAJOR_VERSION - sets 'hana_os_major_version' terraform parameter - default is taken from 'VERSION'
+# FENCING_MECHANISM - (sbd/native) choose fencing mechanism
+# QESAP_SCC_NO_REGISTER - define variable in openqa to skip SCC registration via ANSIBLE
+# HANA_MEDIA - Hana install media directory
+# _HANA_MASTER_PW (mandatory) - Hana master PW (secret)
+# INSTANCE_SID - SAP Sid
+# INSTANCE_ID - SAP instance id
+
+
+use base 'sles4sap_publiccloud_basetest';
+use strict;
+use warnings;
+use testapi;
+use Mojo::File 'path';
+use publiccloud::utils;
+use qesapdeployment;
+use sles4sap_publiccloud;
+use serial_terminal 'select_serial_terminal';
+
+our $ha_enabled = set_var_output("HA_CLUSTER", "0") =~ /false|0/i ? 0 : 1;
+
+sub test_flags {
+    return {fatal => 1, publiccloud_multi_module => 1};
+}
+
+=head2 set_var_output
+    Check if requested openQA variable is defined and returns it's current value.
+    If the variable is not defined, it will set up the default value provided and returns it.
+    $variable - variable to check against OpenQA settings
+    $default - default value to set in case the variable is missing.
+=cut
+
+sub set_var_output {
+    my ($variable, $default) = @_;
+    set_var($variable, get_var($variable, $default));
+    return (get_var($variable));
+}
+
+=head2 create_ansible_playbook_list
+
+    Detects HANA/HA scenario from openQA variables and creates "ansible: create:" section in config.yaml file.
+
+=cut
+
+sub create_playbook_section {
+    # Cluster related setup
+    my @playbook_list;
+    my @hana_playbook_list = (
+        "pre-cluster.yaml",
+        "sap-hana-preconfigure.yaml -e use_sapconf=" . set_var_output("USE_SAPCONF", "true"),
+        "cluster_sbd_prep.yaml",
+        "sap-hana-storage.yaml",
+        "sap-hana-download-media.yaml",
+        "sap-hana-install.yaml",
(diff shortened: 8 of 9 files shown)


lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    record_info("Cleaning up terraform infrastructure");
+    my $terraform_cleanup_rc = qesap_execute(verbose => "--verbose", cmd => "terraform", cmd_options => "-d", timeout => 1200);
+    force_soft_failure("Terraform destroy failed.") if $terraform_cleanup_rc != 0;
+    record_info("Cleanup finished");
+}
+
+sub post_fail_hook {
+    my ($self,) = @_;
```
dzedro, line 39: ,

lpalovsky: Fixed

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
...
+sub cleanup {
+    my ($self) = @_;
+    # Do not run destroy if already executed
+    return if ($self->{cleanup_called});
+    $self->{cleanup_called} = 1;
+    my $inventory_check_cmd = join(" ", ("test", "-f", qesap_get_inventory()));
+    if (script_run($inventory_check_cmd) == 0) {
+        record_info("Ansible cleanup");
+        my $ansible_cleanup_rc = qesap_execute(verbose => "--verbose", cmd => "ansible", cmd_options => "-d", timeout => 1200);
+        force_soft_failure("Ansible destroy failed.") if $ansible_cleanup_rc != 0;
```
dzedro, line 29: What is the point of this soft fail ?
I would rather do cleanup again.
Nobody is looking at soft fails if there is no reason.

lpalovsky: Yeah... that is true :/
The idea was that if ansible cleanup fails for some reason, the test would continue with terraform cleanup and destroy at least infrastructure/resources.
But what if cleanup fails for the second time again? Should the whole test fail as well? 

I see now I have the same thing with terraform destroy fail. That is not a good idea since nobody would cleanup the resources manually.

dzedro: I would make test fail, it will be better at begging.
We will see that there are some issues even after retry and debug it.
Otherwise there will be left resources. :money_mouth_face:

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+
+    Check certain openQA variables and set default values if not existing.
+    Detects HANA/HA scenario, sets values accordingly and checks for possible conflicts.
+    Default scenario is plain 1 node deployment without HANA and HA stack.
+    Enable 'QESAP_HANA_HA_ENABLED' and adjust 'NODE_COUNT' accordingly for HA/HANA scenario.
+
+=cut
+
+sub set_default_values {
+    set_var_output("QESAP_HANA_HA_ENABLED", "false");
```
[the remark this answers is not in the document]
lpalovsky: You mean specific ones or all of them ? 
Yeah HA and related variables make sense... I'll make them mandatory:
QESAP_HANA_HA_ENABLED
NODE_COUNT
QESAP_VM_SIZE

What about:
USE_SAPCONF - I think there is a plan to set default value in qesap deployment project

Do Hana related vars like INSTANCE_SID, INSTANCE_ID, HANA_MEDIA be mandatory?
no idea what SAP_DOMAIN is suposed to be  :/

lib/sles4sap_publiccloud_basetest.pm
lpalovsky: There is one thing I am afraid about ... the test will fail multiple times on ansible cleanup, and won't proceed to terraform destroy.
A person reviewing the test will work on the issue but will not think about going and manually do the cleanup. 
Is it possible to have:
- ansible cleanup will fail
- test will be set as failed, but will continue to terraform destroy, so the resources are forced to be cleaned up?

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
```diff
@@ -0,0 +1,39 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Generic yaml template for use with qe-sap-deployment project.
+#   Settings are meant to be controlled via OpenQA variables and managed by test:
+#   tests/sles4sap/publiccloud/qesap_ansible.pm
+# https://github.com/SUSE/qe-sap-deployment
```
mpagot, line 9: What is purpose to add the link to the qe-sap project here? Maybe a better place is line 6

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
```diff
@@ -0,0 +1,39 @@
...
+    vm_size: "%QESAP_VM_SIZE%"
+    hana_cluster_fencing_mechanism: "%FENCING_MECHANISM%"
+
+ansible:
+  hana_urls:
+    - "%HANA_SAR%"
+    - "%HANA_CLIENT_SAR%"
+    - "%HANA_SAPCAR%"
+  destroy:
+    - deregister.yaml
```
mpagot, line 38: So this YAML is special as the _configure:_ playbooks list is missing. Right?

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
dzedro: I would say all, IMO it's better and cleaner to have "one" place where test settings are defined.
Some settings which are same for all e.g. Azure tests can be defined for whole group or provider. `pc_defaults` or `azure_defaults`
SAP_DOMAIN ? Me neither.

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: Yep. It's a bit quirky to add playbook (or any other) list into config using the %SOMETHING% method.
So I am doing it directly in YAML. Ideally it would be to do the whole thing like that and there will be no need for yaml template. But having deployment is priority so we can think about it in the future.

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
     assert_script_run("mkdir -p $paths{deployment_dir}", quiet => 1);
 }
 
+=head3 qesap_get_variables
+
+    Scans yaml config for '%OPENQA_VARIABLE%' placeholders and searches for values in OpenQA defined variables.
+    Returns hash with openqa variable key/value pairs.
```
mpagot, line 97: Is it mostly intended to be used as input for **qesap_prepare_env**? Are there any other use cases for it?
If it is the only purpose, maybe we can change it to a private function and change  **qesap_prepare_env** to something like

```
sub qesap_prepare_env {
    my (%args) = @_;
    my $variables
    if($args{openqa_variables}) {
        $variables = $args{openqa_variables};
    }
    else {
        $variables = qesap_get_variables();
    }
```

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: No particular reason. I can put it on line ... maybe 7?  Line 6 will become quite long if I put the link in there. 
But in general... I don't have preference. It's just the 120ch line python habit :D

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
...
+=cut
+
+sub qesap_get_variables {
+    my %paths = qesap_get_file_paths();
+    my $yaml_file = $paths{'qesap_conf_src'};
+    my %variables;
+    my $grep_cmd = "grep -v '#' | grep -oE %[A-Z0-9_]*% | sed s/%//g";
+    my $cmd = join(" ", "curl -s -L", $yaml_file, "|", $grep_cmd);
```
mpagot, line 105: Please change to **curl -s -fL** by adding the **--fail** argument. It results in culr to clearly fails if the $yaml_file is not there. We recently applied the same fix in all other curl cmd on this lib file

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
...
+    die("Yaml config file '$yaml_config_path' does not exist.") if assert_script_run("test $yaml_config_path") > 0;
+    my $raw_file = script_output("cat $yaml_config_path");
+    my $yaml_data = $ypp->load_string($raw_file);
+
+    $yaml_data->{ansible}{$section} = $content;
+
+    # write into file
+    my $yaml_dumped = $ypp->dump_string($yaml_data);
+    save_tmp_file($paths{qesap_conf_filename}, $yaml_dumped);
+    assert_script_run('curl -v -L ' . autoinst_url . "/files/" . $paths{qesap_conf_filename} . ' -o ' . $paths{qesap_conf_trgt});
```
mpagot, line 140: **-fL**

lib/sles4sap_publiccloud_basetest.pm
dzedro: I don't know how often tend ansible destroy to fail.
We should focus to make sure the resource is destroyed.
Like retry (or third retry) could be done with terraform destroy ?

lpalovsky: Yep that makes sense. 
So something like 3x retry for both ansible and terraform and fail afterwards?

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
lpalovsky: Ok then. Will make everything madatory.

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
...
+    my $ypp = YAML::PP->new;
+    my $section = $args{ansible_section};
+    my $content = $args{section_content};
+    my %paths = qesap_get_file_paths();
+    my $yaml_config_path = $paths{qesap_conf_trgt};
+    die("Yaml config file '$yaml_config_path' does not exist.") if assert_script_run("test $yaml_config_path") > 0;
+    my $raw_file = script_output("cat $yaml_config_path");
+    my $yaml_data = $ypp->load_string($raw_file);
+
+    $yaml_data->{ansible}{$section} = $content;
```
mpagot, line 135: Generic consideration (no need any code change in this PR): this function introduces a openQA test code dependency with the internal conf.yaml of qe-sap-deployment structure. Now this function and the code of the caller 

```
my $section = $args{ansible_section};
```

guess on the fact that the YAML has an **ansible:** section and the caller guess that the **ansible:** has a **create:** (or other) section in it intended to be a sequence of playbooks. Before this PR the dependency was only in the files in  data/sles4sap/qe_sap_deployment and the gluescript uses **apiver:**
 in them to try to keep any dependency as transparent as possible.

lib/qesapdeployment.pm
```diff
@@ -295,14 +349,17 @@ sub qesap_prepare_env {
...
-    qesap_get_deployment_code();
-    qesap_pip_install();
+    # Option to skip straight to configuration
+    unless ($args{only_configure}) {
+        qesap_create_folder_tree();
+        qesap_get_deployment_code();
+        qesap_pip_install();
+
+        record_info("QESAP yaml", "Preparing yaml config file");
+        assert_script_run('curl -v -L ' . $paths{qesap_conf_src} . ' -o ' . $paths{qesap_conf_trgt});
```
mpagot, line 360: **-f** (it was there in the original code, any specific reason not to have it?)

lib/qesapdeployment.pm
```diff
@@ -295,14 +349,17 @@ sub qesap_prepare_env {
     my $variables = $args{openqa_variables};
     my $provider = $args{provider};
     my %paths = qesap_get_file_paths();
-    my $qesap_conf_src = "sles4sap/qe_sap_deployment/" . $paths{qesap_conf_filename};
 
-    qesap_create_folder_tree();
-    qesap_get_deployment_code();
-    qesap_pip_install();
+    # Option to skip straight to configuration
+    unless ($args{only_configure}) {
```
mpagot, line 354: Not for this PR, but I propose to move

```
 qesap_create_folder_tree();
        qesap_get_deployment_code();
        qesap_pip_install();
```

out from this function and leave to the test to call them

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+# Maintainer: QE-SAP <qe-sap@suse.de>
+# Summary: Base test providing cleanup, post fail and post run hooks for qe-sap-deployment project.
```
mpagot, line 6: for tests using qe-sap-deployment project.

lib/qesapdeployment.pm
lpalovsky: Yep that could work, I will make it like that. 
I just didn't want to force this mechanism into other functions.

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
```
mpagot, line 3: why not to have something in the filename to refer to qe-sap-deployment?

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
...
+use qesapdeployment;
+
+our @EXPORT = qw(cleanup);
+
+sub cleanup {
+    my ($self) = @_;
+    # Do not run destroy if already executed
+    return if ($self->{cleanup_called});
+    $self->{cleanup_called} = 1;
+    my $inventory_check_cmd = join(" ", ("test", "-f", qesap_get_inventory()));
```
mpagot, line 25: Is the idea here to uses the existence of the inventory as sort of internal state?

lib/sles4sap_publiccloud_basetest.pm
lpalovsky: Depends if the intention is to have sles4sap specific basetest or it should be a general one for all qesap deployments.
qesap_publiccloud_basetest.pm ?

lib/sles4sap_publiccloud_lib.pm
```diff
@@ -0,0 +1,148 @@
...
+
+our @EXPORT = qw(
+  run_cmd
+  get_promoted_hostname
+  get_hana_topology
+  wait_for_sync
+);
+
+# Global variables
+our $crm_mon_cmd = 'crm_mon -R -r -n -N -1';
```
mpagot, line 25: You can 

```
use hacluster '$crm_mon_cmd';
```

lib/sles4sap_publiccloud_basetest.pm
dzedro: Sure :+1:

lib/sles4sap_publiccloud_basetest.pm
lpalovsky: It's a check if ansible destroy does even need to be called. 
If there is no inventory preset, the ansible was most probably not executed. The intention is to avoid "ansible destroy" returing some strage value and causing test failures.
There is probably a better test for that but we can think about it later.

lib/qesapdeployment.pm
lpalovsky: Ah sorry... just a honest mistake while replacing the qesap_conf_src variable. Will fix it.

lib/qesapdeployment.pm
lpalovsky: Thanks for suggestion. Added.

lib/qesapdeployment.pm
lpalovsky: Fixed

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: Appended to L6

lib/sles4sap_publiccloud_basetest.pm
mpagot: I like the idea, no better tests come to my mind too

lib/sles4sap_publiccloud_basetest.pm
alvarocarvajald: > There is one thing I am afraid about ... the test will fail multiple times on ansible cleanup, and won't proceed to terraform destroy.

Shouldn't `lib/publiccloud/provider` call the `cleanup()` method on failures which issues a `terraform destroy`?

schedule/sles4sap/publiccloud_hanasr.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: publiccloud_hanasr
+description: |
+  Run HANASR tests on public cloud instances using ha-sap-terraform-deployments tool:
```
mpagot, line 4: is it reference to ha-sap-terraform-deployments right here?

lib/sles4sap_publiccloud_basetest.pm
lpalovsky: Yes, but we need the ansible destroy as well so it for example de-registers host at the end. 
This one also uses the qe-sap-deployment projects glue script instead of plain commands.

t/09_qesapdeployment.t
```diff
@@ -8,6 +8,10 @@ use List::Util qw(any);
 use testapi 'set_var';
 use qesapdeployment;
 set_var('QESAP_CONFIG_FILE', 'MARLIN');
+set_var('BACKEND', 'qemu');
+set_var('AUTOINST_URL_HOSTNAME', 'praisethesun');
+set_var('QEMUPORT', 9000);
+set_var('JOBTOKEN', "solaire");
```
mpagot, line 14: Fine for this PR, needs to be changed in the future. Without them the error was:

```
#   Failed test 'No tests run for subtest "[qesap_get_inventory] upper case"'
#   at t/09_qesapdeployment.t line 16.
Could not retrieve required variable WORKER_HOSTNAME at /__w/os-autoinst-distri-opensuse/os-autoinst-distri-opensuse/lib/qesapdeployment.pm line 80.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 2.
t/09_qesapdeployment.t ..... 
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/2 subtests 
```

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
alvarocarvajald: I don't think settings such as USE_SAPCONF or QESAP_HANA_HA_ENABLED should be mandatory. It's clutter for tests that don't use sapconf or that do not start a cluster. Agree on the other settings though.

schedule/sles4sap/publiccloud_hanasr.yml
lpalovsky: Oh yes... thanks. Good catch. The schedule is quite old. Fixed.

lib/sles4sap_publiccloud_lib.pm
alvarocarvajald: +1

schedule/sles4sap/publiccloud_hanasr.yml
alvarocarvajald: +1. Should be qe-sap-deployments, right?

t/09_qesapdeployment.t
lpalovsky: Ah yes... the thing is if you do the data_url mock function, you will need it almost for all test existing there. 
The other this is that tests like:
'[qesap_get_inventory] upper case'
'[qesap_get_inventory] lower case'
Will kind of only test if perl lc function is working. Do we really test stuff like that? Especially if VR is required for each PR which should cover things like that. 
If you use variables, you will at least check if the functions are behaving correctly with testapi.

tests/sles4sap/publiccloud/qesap_ansible.pm
```diff
@@ -0,0 +1,56 @@
...
+use serial_terminal 'select_serial_terminal';
+
+sub test_flags {
+    return {fatal => 1, publiccloud_multi_module => 1};
+}
+
+sub run {
+    my ($self, $run_args) = @_;
+    select_serial_terminal;
+    my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /False|false|0/ ? 0 : 1;
```
mpagot, line 26: Is this new QESAP_ to be considered a generic qesapdeploy lib setting? Maybe not as far as it is in the test code and not in qesapdeployment?

If it is we can think about documenting it in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/0836e03e6d49b791d5986b597374992febe58d58/variables.md?plain=1#L244

If not, at least for Trento, I decided to add a prefix to them **HANASR_QESAP_HANA_HA_ENABLED**

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+    set_var_output("NODE_COUNT", 1);
+    set_var_output("QESAP_VM_SIZE", "Standard_E4s_v3");
```
mpagot, line 66: QESAP_  same as before

mpagot (approved):
LGTM

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
dzedro: Default settings like e.g. USE_SAPCONF don't need to be explicitly defined, they should have default value.

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+    set_var_output("HANA_OS_MAJOR_VERSION", (split("-", get_var("VERSION")))[0]);
+
+    # Cluster related setup
+    my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /False|false|0/ ? 0 : 1;
```
alvarocarvajald, line 70: You can use the `i` modifier in regexp to check independent of case. For example:
```
my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /false|0/i ? 0 : 1;
```

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+
+=head2 create_ansible_playbook_list
+
+    Detects HANA/HA scenario from openQA variables and creates "ansible: create:" section in config.yaml file.
+
+=cut
+
+sub create_hana_vars_section {
+    # Cluster related setup
+    my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /False|false|0/ ? 0 : 1;
```
alvarocarvajald, line 128: Same suggestion regarding `i` modifier to regexps.

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+    $default - default value to set in case the variable is missing.
+
+=cut
+
+sub set_var_output {
+    my ($variable, $default) = @_;
+    return (get_var($variable)) if get_var($variable);
+    set_var($variable, $default);
+    return ($default);
+}
```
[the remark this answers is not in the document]
lpalovsky: I think we talked about it. I wanted to avoid setting variables even if they are already set with the same value. But if it is not an issue, I will use the shorter version.

lib/qesapdeployment.pm
lpalovsky: You cannot avoid this dependency. Yaml files are very static and I am not sure how would you like to have granular control over qesap settings from openQA side. Remeber there will be a lot of scenarios and having multiple almost identical yaml files is not ideal either. %VAR% method is really limited. There is a problem if you introduce a simple list like adding list of hana media or list of playbooks.
Maybe we should stop changing yaml format constantly? I see this as much bigger problem already.

tests/sles4sap/publiccloud/qesap_ansible.pm
lpalovsky: I will change it to HA_CLUSTER which is already standard variable for sles4sap/ha tests

tests/sles4sap/publiccloud/qesap_terraform.pm
lpalovsky: Will change variable to PUBLIC_CLOUD_INSTANCE_TYPE which is already used one for this purpose. Less confusion for people familiar with existing PC tests and no need to change them in existing job group settings.

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
...
+            @playbook_list = ("pre-cluster.yaml", "cluster_sbd_prep.yaml");
+            qesap_create_ansible_section(ansible_section=>'create', section_content=>\@playbook_list);
+
+=cut
+
+sub qesap_create_ansible_section {
+    my (%args) = @_;
+    my $ypp = YAML::PP->new;
+    my $section = $args{ansible_section};
+    my $content = $args{section_content};
```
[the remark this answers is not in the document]
lpalovsky: Thanks. Will try it out.

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
```diff
@@ -0,0 +1,38 @@
...
+#   tests/sles4sap/publiccloud/qesap_ansible.pm
+
+provider: 'azure'
+apiver: 2
+terraform:
+  provider: 'azure'
+  variables:
+    # GENERAL VARIABLES #
+    az_region: "%PUBLIC_CLOUD_REGION%"
+    admin_user: "cloudadmin"
```
lilyeyes, line 17: Coding style consistent: using 'cloudadmin' instead of "cloudadmin"

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
```diff
@@ -0,0 +1,38 @@
...
+    public_key : "~/.ssh/id_rsa.pub"
+    private_key: "~/.ssh/id_rsa"
```
lilyeyes, line 19: Coding style consistent: using `'` instead of `"`

lib/qesapdeployment.pm
```diff
@@ -74,6 +77,7 @@ sub qesap_get_file_paths {
     $paths{deployment_dir} = get_var('QESAP_DEPLOYMENT_DIR', get_var('DEPLOYMENT_DIR', '/root/qe-sap-deployment'));
     $paths{terraform_dir} = get_var('PUBLIC_CLOUD_TERRAFORM_DIR', $paths{deployment_dir} . '/terraform');
     $paths{qesap_conf_trgt} = $paths{deployment_dir} . "/scripts/qesap/" . $paths{qesap_conf_filename};
+    $paths{qesap_conf_src} = data_url("sles4sap/qe_sap_deployment/" . $paths{qesap_conf_filename});
```
lilyeyes, line 80: Coding style consistent: using ' instead of "

lilyeyes (commented):
I really do not familiar with this code changes, but only found some coding style consistent: 
It is better using `'` instead of `"` in some related lines.

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
dzedro: I kind of agree, but we should focus on more important things.

lib/qesapdeployment.pm
lpalovsky: Thanks. Will be fixed in next push

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: Thanks that is no problem. I just did a mass replace and will be fixed in next push.

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: Will be changed in next push

lpalovsky:
New VRs after changes:
SIngle node deploy: https://mordor.suse.cz/tests/5536
2 nodes without cluster: https://mordor.suse.cz/tests/5539
Cluster with 2 nodes: https://mordor.suse.cz/tests/5538

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
lpalovsky: Fixed as we discussed on slack channel:
- Variable name changes so they match existing variables from qemu tests:
  -  QESAP_HANA_HA_ENABLED -> HA_CLUSTER
  - QESAP_VM_SIZE -> PUBLIC_CLOUD_INSTANCE_TYPE
- Default behavior is deployment of one node and no HA cluster 
- HA_CLUSTER - enables cluster deployment
-  USE_SAPCONF - default = true
- Mandatory variables: HANA_MEDIA _HANA_MASTER_PW INSTANCE_SID INSTANCE_ID SCC_REGCODE_SLES4SAP
- NODE_COUNT needs to be specified if ha cluster is enabled

Please let me know if I am missing something.

lib/qesapdeployment.pm
lpalovsky: Change applied:
```
    my $variables = $args{openqa_variables} ? $args{openqa_variables} : qesap_get_variables();
```

lib/sles4sap_publiccloud_basetest.pm
lpalovsky: Changed

lib/sles4sap_publiccloud_lib.pm
lpalovsky: Yep, good idea. Changed with last push.
@alvarocarvajald I changed the filename as well to lib/sles4sap_publiccloud.pm :)

tests/sles4sap/publiccloud/qesap_ansible.pm
lpalovsky: Fix applied

lib/qesapdeployment.pm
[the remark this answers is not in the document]
lpalovsky: Works and change is pushed.

lib/qesapdeployment.pm
```diff
@@ -313,7 +370,7 @@ sub qesap_prepare_env {
     my $exec_rc = qesap_execute(cmd => 'configure', verbose => 1);
     push(@log_files, $hana_vars) if (script_run("test -e $hana_vars") == 0);
     qesap_upload_logs(failok => 1);
-    die if $exec_rc != 0;
+    die if $exec_rc > 0;
```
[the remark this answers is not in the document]
lpalovsky: I remember having some issue with that so I used it as a stupid workaround. Then I have forgotten to look at it again.  
I have reverted it back to original. If it fails for some reason again I will just deal with it in another PR

tests/sles4sap/publiccloud/qesap_ansible.pm
```diff
@@ -0,0 +1,56 @@
...
+        $run_args->{site_b} = $instance if ($instance_id ne $master_node);
+        record_info("Instances:", "Detected HANA instances:
+        Site A: $run_args->{site_a}{instance_id}
+        Site B: $run_args->{site_b}{instance_id}") if ($instance_id eq $master_node);
+    }
+
+    return 1;
+}
+
+1;
```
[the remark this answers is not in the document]
lpalovsky: If I understand correctly, this is added in qesap_terraform.pm into $run_args->{instances}

```
*** 
my @instances = $provider->create_instances(check_connectivity => 0);
***
$self->{instances} = $run_args->{instances} = \@instances_export;
```

tests/sles4sap/publiccloud/qesap_cleanup.pm
```diff
@@ -0,0 +1,22 @@
...
+use testapi;
+
+
+sub run {
+    my ($self) = @_;
+    return if get_var("PUBLIC_CLOUD_NO_CLEANUP");
+    $self->cleanup();
+}
+
+1;
```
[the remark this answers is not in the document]
lpalovsky: Cleanup routine does not need this since now it is triggered directly by " qesap.py terraform -d ". It does not use any PC modules.

tests/sles4sap/publiccloud/qesap_terraform.pm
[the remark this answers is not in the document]
lpalovsky: Done and pushed.

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+    set_var_output("QESAP_HANA_HA_ENABLED", "false");
+    set_var_output("NODE_COUNT", 1);
+    set_var_output("QESAP_VM_SIZE", "Standard_E4s_v3");
+    set_var_output("HANA_OS_MAJOR_VERSION", (split("-", get_var("VERSION")))[0]);
+
+    # Cluster related setup
+    my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /False|false|0/ ? 0 : 1;
+
+    # set fencing to native if no ha cluster is set to avoid deploying ISCSI server
+    set_var("FENCING_MECHANISM", "native") if $ha_enabled == 1;
```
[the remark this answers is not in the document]
lpalovsky: Ah yes... it should have... this is why I had some issues deploying without iscsi. I thought it is related to deployment project. :/
Thanks. 
Fixed.

tests/sles4sap/publiccloud/qesap_terraform.pm
lpalovsky: Thanks. Did not know that one.
Fixed and pushed.

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -0,0 +1,180 @@
...
+
+=head2 create_ansible_playbook_list
+
+    Detects HANA/HA scenario from openQA variables and creates "ansible: create:" section in config.yaml file.
+
+=cut
+
+sub create_playbook_section {
+    # Cluster related setup
+    my $ha_enabled = get_var("QESAP_HANA_HA_ENABLED") =~ /False|false|0/ ? 0 : 1;
```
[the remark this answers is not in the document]
lpalovsky: Fixed and pushed.

tests/sles4sap/publiccloud/qesap_terraform.pm
lpalovsky: Fixed

tests/sles4sap/publiccloud/qesap_terraform.pm
lpalovsky: Done

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
lpalovsky: And fixed.

lib/qesapdeployment.pm
```diff
@@ -87,6 +91,56 @@ sub qesap_create_folder_tree {
...
+=cut
+
+sub qesap_create_ansible_section {
+    my (%args) = @_;
+    my $ypp = YAML::PP->new;
+    my $section = $args{ansible_section};
+    my $content = $args{section_content};
+    my %paths = qesap_get_file_paths();
+    my $yaml_config_path = $paths{qesap_conf_trgt};
+    die("Yaml config file '$yaml_config_path' does not exist.") if assert_script_run("test $yaml_config_path") > 0;
```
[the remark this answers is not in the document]
lpalovsky: Aaah... yeah you are right. I have forgotten asser_scrip_run has a fail message. I just changed it to pure assert_script_run with fail message defined..

tests/sles4sap/publiccloud/qesap_ansible.pm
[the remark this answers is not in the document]
alvarocarvajald: Now I'm not sure. Have you seen `tests/sles4sap/publiccloud/qesap_ansible.pm` failing and the job successfully calling on `terraform destroy`? I thought either `$self->{provider}` or `$self->{instance}` was required for that.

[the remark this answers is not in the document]
lpalovsky: Yes this works fine. In this case "terraform destroy" is not called via Perl module. Instead glue script is executed with "qesap.py ... terraform - d". So $self-{provider} is not needed here. The cleanup is executed outside of OpenQA just via assert_script_run.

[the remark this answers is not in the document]
alvarocarvajald: Cool. Then disregard my comment.

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -0,0 +1,56 @@
...
+    if (script_run($inventory_check_cmd) == 0) {
+        record_info("Ansible cleanup");
+        my $ansible_cleanup_rc = qesap_execute(verbose => "--verbose", cmd => "ansible", cmd_options => "-d", timeout => 1200);
+        force_soft_failure("Ansible destroy failed.") if $ansible_cleanup_rc != 0;
+    }
+
+    record_info("Cleaning up terraform infrastructure");
+    my $terraform_cleanup_rc = qesap_execute(verbose => "--verbose", cmd => "terraform", cmd_options => "-d", timeout => 1200);
+    force_soft_failure("Terraform destroy failed.") if $terraform_cleanup_rc != 0;
```
alvarocarvajald, line 34: Same thing: `if $terraform_cleanup_rc;` would be enough.

alvarocarvajald (approved):
I left some comments on some small things, but in general LGTM

lib/qesapdeployment.pm
```diff
@@ -313,7 +371,7 @@ sub qesap_prepare_env {
     my $exec_rc = qesap_execute(cmd => 'configure', verbose => 1);
     push(@log_files, $hana_vars) if (script_run("test -e $hana_vars") == 0);
     qesap_upload_logs(failok => 1);
-    die if $exec_rc != 0;
+    die("Qesap deployment returned non zero value during 'configure' phase.") if $exec_rc != 0;
```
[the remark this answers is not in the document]
lpalovsky: Ah yes, thank you... I am forgetting about the reversed true/false logic between perl and return codes.
Fixed

tests/sles4sap/publiccloud/qesap_ansible.pm
```diff
@@ -0,0 +1,56 @@
...
+    return {fatal => 1, publiccloud_multi_module => 1};
+}
+
+sub run {
+    my ($self, $run_args) = @_;
+    select_serial_terminal;
+    my $ha_enabled = get_required_var("HA_CLUSTER") =~ /false|0/i ? 0 : 1;
+    my $instances = $run_args->{instances};
+
+    die if qesap_execute(cmd => 'ansible', timeout => 3600, verbose => 1) > 0;
```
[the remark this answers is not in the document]
lpalovsky: Added: "die("Ansible deploymend FAILED. Check 'qesap*' logs for details.")"

tests/sles4sap/publiccloud/qesap_ansible.pm
```diff
@@ -0,0 +1,56 @@
...
+    record_info("FINISHED", "Ansible deployment process finished successfully.");
+    return unless $ha_enabled == 1;
```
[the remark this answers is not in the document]
lpalovsky: Fixed

PR #15755, opened 2022-10-21
Adjust ansible test module for ALP and SLEM and get rid of user serial terminal

pdostal:
* Enable the Ansible test module for ALP, MicroOS and JeOS.
 * Get rid of user serial terminal as it's hacky and unstable on s390x.

- Follows after: #15684
- Related ticket: [poo#116935](https://progress.opensuse.org/issues/116935)
- Discovered bug: [bsc#1204544](https://bugzilla.suse.com/show_bug.cgi?id=1204544)
- Verification run:
  - opensuse-Tumbleweed-DVD.x86_64 [extra_tests_textmode@64bit](https://pdostal-server.suse.cz/tests/2864)
  - jeos-extratest@uefi-virtio-vga [jeos-extratest@uefi-virtio-vga](https://pdostal-server.suse.cz/tests/2866)
  - microos@64bit [microos@64bit](https://pdostal-server.suse.cz/tests/2867)
  - mau-extratests2@64bit [mau-extratests2@64bit](https://pdostal-server.suse.cz/tests/2870)
  - alp_default@64bit [alp_default@64bit](https://pdostal-server.suse.cz/tests/2871)

This currently does not test the `zypper` module on transactional systems - that will be added later.

PR #15401, opened 2022-08-23
SLE Micro 5.3 wizard install tests

pablo-herranz:
Add a new job to test the *'new'* installation wizard that pops up when no *ignition* neither *combustion* disk is present.

- Related ticket: https://progress.opensuse.org/issues/114938
- Needles: *not yet*
- Verification runs: 
  - [slem_image_default with ignition](https://openqa.suse.de/t9378781)
  - [slem_containers with ignition](https://openqa.suse.de/t9378779)
  - [slem_image_default with wizard](https://openqa.suse.de/t9378385)
  - [slem_containers with wizard](https://openqa.suse.de/t9378247)

PR #15790, opened 2022-10-27
toolbox: create user only if it does not already exist

DimStar77:
In MicroOS tests it is well possible that some other test running
before toolbox already created the user, which would then result
in failure to re-create it once again.

- Related ticket: https://openqa.opensuse.org/tests/2837897#step/toolbox/5
- Needles: N/A
- Verification run: N/A


DimStar77:
2nd issue identified after https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15755

Vogtinator:
I wonder what `if (!check_var('FIRST_BOOT_CONFIG', 'wizard'))` is/was about. With `jeos-firstboot`, the user isn't created either, so this should just fail. I tried to find a test to check why it works anyway but couldn't actually find any job with the wizard which had the `toolbox` module included.

CC @pablo-herranz

pablo-herranz:
> I wonder what `if (!check_var('FIRST_BOOT_CONFIG', 'wizard'))` is/was about. With `jeos-firstboot`, the user isn't created either, so this should just fail. I tried to find a test to check why it works anyway but couldn't actually find any job with the wizard which had the `toolbox` module included.
> 
> CC @pablo-herranz

This was added in [this PR](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15401) to re-use JeOS installation wizard in SLE Micro. The toolbox tests failed if the wizard had been run before because the user already existed, so I skipped user creation in that case.

Vogtinator:
> The toolbox tests failed if the wizard had been run before because the user already existed, so I skipped user creation in that case.

I see why now. It's not `jeos-firstboot` which creates the user, but `jeos/firstrun.pm` which does:

```
    # Create user account, if image doesn't already contain user
    # (which is the case for SLE images that were already prepared by openQA)
    if (script_run("getent passwd $username") != 0) {
        assert_script_run "useradd -m $username -c '$realname'";
        assert_script_run "echo $username:$password | chpasswd";
    }
```

sounds like some refactoring is in order to deduplicate that.

Vogtinator (approved):

pdostal:
Yes, the user is also not present in current ALP test runs.

Thank you @DimStar77 for fixing this.

PR #15795, opened 2022-10-27
Install yast2-migration package when creating hdd for HA on ppc64le

badboywj:
yast2-migration package is needed to do HA online migration testing.

- Related ticket: https://progress.opensuse.org/issues/117238
- Verification run: https://openqa.suse.de/tests/9822941, https://openqa.suse.de/tests/9822942


coolgw:
LGTM if this change not impact other cases which also use this file since this ONLY needed by powerVM

coolgw (approved):

chcao:
> LGTM if this change not impact other cases which also use this file since this ONLY needed by powerVM

The profile is for ppc64le, so it won't impact other arches' cases.

chcao (approved):

PR #15798, opened 2022-10-28
New wsl_gui pattern added to the repos

pablo-herranz:
By adding the new `wsl_gui` pattern to the repos, also a new screen has been added to the YaST2 installation, causing all the tests to fail. This fix will take that screen into account and also, mark the option to install that package only in tests that will register into SCC.

- Related ticket: https://progress.opensuse.org/issues/110140
- Needles:
  - https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/blob/master/wsl-gui-pattern-20221028.png
- Verification runs: :heavy_check_mark: 
  - [Win11 SLE15-SP3 register](https://openqa.suse.de/tests/9855037)
  - [Win11 SLE15-SP3 skip register](https://openqa.suse.de/tests/9855038)
  - [Win10-BIOS SLE15-SP4 register](https://openqa.suse.de/tests/9855034)
  - [Win10-BIOS SLE15-SP4 skip register](https://openqa.suse.de/tests/9855036)
  - [Win10-UEFI Tumbleweed](https://openqa.opensuse.org/tests/2849191)

### DIFF
--- a/tests/wsl/firstrun.pm
+++ b/tests/wsl/firstrun.pm
@@ -108,6 +108,11 @@ sub run {
         assert_screen 'local-user-credentials';
         enter_user_details([$realname, undef, $password, $password]);
         send_key 'alt-n';
+        # wsl-gui pattern installation
+        if (is_sle('>15-SP3')) {
+            assert_screen 'wsl-gui-pattern';
+            send_key 'alt-n';
+        }
         # Registration
         is_sle && register_via_scc();
         # And done!


mloviska:
I think this requires a new function to determine which registration we wanna use, ideally parametrized with openQA test variable. 
Now we have these options:

1) SLES w/ GUI
2) SLES w/o GUI
3) SLED w/ GUI
4) SLED w/o GUI

mloviska:
Also we should keep in mind the old flow. I guess SP3 wont get this feature at all
https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=2.137&groupid=326

mloviska:
For product specification, we already have a variable in use `SLE_PRODUCT`

pablo-herranz:
I've added some control for checking that screen only if 15-SP4 or greater. Also if `SLE_PRODUCT='SLED'` is provided, the SLED version will be selected in that screen.

tests/wsl/firstrun.pm
```diff
@@ -108,6 +108,15 @@ sub run {
         assert_screen 'local-user-credentials';
         enter_user_details([$realname, undef, $password, $password]);
         send_key 'alt-n';
+        # wsl-gui pattern installation
+        if (is_sle('>15-SP3')) {
+            assert_screen 'wsl-gui-pattern';
+            send_key 'alt-u' if (check_var('SLE_PRODUCT', 'SLED'));    # Install SLED instead of SLES
+            save_screenshot;    # Save screenshot to verify the checkbox was marked properly
+            send_key 'alt-i' if (is_sut_reg);    # wsl-gui requires registration
```
mloviska, line 116: I still think we should be able to test both options. We should be able to install `textmode` WSL (what we have now) and also `WSL_G`. The thing is that in order to install `WSL_G` I assume we need to have repositories enabled, therefore if we have no registration case, it won't be possible.

mloviska: I would also like to keep the logic of main simple and have a function for each screen as it is now.

okurz:
This seems to have caused a regression, see https://openqa.suse.de/tests/9861643#step/firstrun/13 where you expect the gui pattern selection but first the system needs to be registered.

pablo-herranz:
> This seems to have caused a regression, see https://openqa.suse.de/tests/9861643#step/firstrun/13 where you expect the gui pattern selection but first the system needs to be registered.

Yes, I'm about to fix it in #15829

PR #15801, opened 2022-10-31
[alp][kvm] Add alp kvm workload container setup code

alice-suse:
For alp virtualization test, this PR serves to setup kvm workload container. In details, it mainly provides:
1) a utility lib for kvm workload container
2) a file to do the kvm container setup
3) a schedule file for alp kvm core tests(to be extended with more supported)
4) a nat virtual network to be used in guest installation test

- Related ticket: https://progress.opensuse.org/issues/114664
- Needles: https://github.com/os-autoinst/os-autoinst-needles-opensuse/pull/788
- Verification run(updated based on latest push 20221111): 
  PASS job: http://10.67.129.185/tests/366
  FAIL job(post_fail_hook verification with explicit die at the end): http://10.67.129.185/tests/367

### DIFF
--- a/lib/alp_workloads/kvm_workload_utils.pm
+++ b/lib/alp_workloads/kvm_workload_utils.pm
@@ -1,0 +1,209 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: This file provides the utility functions for kvm workload container.
+# Maintainer: alice <xlai@suse.com>, qe-virt@suse.de
+
+package alp_workloads::kvm_workload_utils;
+
+use base Exporter;
+use Exporter;
+use strict;
+use warnings;
+use testapi;
+use Utils::Architectures;
+use Utils::Backends qw(use_ssh_serial_console);
+use ipmi_backend_utils;
+use utils;
+use virt_autotest::virtual_network_utils qw(clean_all_virt_networks);
+
+our @EXPORT = qw(
+  set_ct_engine
+  set_kvm_container_image
+  pull_kvm_container_image
+  config_host_and_kvm_container
+  cleanup_host_and_kvm_container
+  start_kvm_container
+  enter_kvm_container_sh
+  exit_kvm_container
+  install_tools_within_kvm_container
+  setup_services_within_kvm_container
+  collect_kvm_container_setup_logs
+  setup_kvm_container_from_scratch
+  clean_and_resetup_kvm_container
+);
+
+# Container engine: can be podman(default), docker(not tested)
+our $ct_engine = "podman";
+
+my $DEFAULT_KVM_IMAGE = "registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/kvm:latest";
+our $kvm_image = $DEFAULT_KVM_IMAGE;
+
+sub set_ct_engine {
+    $ct_engine = shift;
+
+    die "Container engine can only be set to podman or docker!" if ($ct_engine ne "podman" && $ct_engine ne "docker");
+}
+
+sub set_kvm_container_image {
+    my $_image = shift;
+
+    # Allow customization in case of special test needs
+    if ($_image ne $DEFAULT_KVM_IMAGE) {
+        $kvm_image = $_image;
+        record_info("Change kvm container image from default!", "Now test is using kvm image: $kvm_image.");
+    }
+}
+
+sub pull_kvm_container_image {
+    # Ensure large enough time is given to download
+    # TODO: uncomment once automation is done
+    # assert_script_run("$ct_engine pull $kvm_image", 1800/get_var('TIMEOUT_SCALE', 1));
+    record_info("Download kvm container image successfully.", script_output("$ct_engine images -a --digests"));
+}
+
+sub config_host_and_kvm_container {
+    # Copy contents from kvm container to host to set it up
+    assert_script_run("clear; $ct_engine container runlabel install $kvm_image");
+    save_screenshot;
+
+    # Replace /etc/kvm-container.conf IMAGE value with $kvm_image if non-default
+    if ($kvm_image ne $DEFAULT_KVM_IMAGE) {
+        assert_script_run("cp /etc/kvm-container.conf /etc/kvm-container.conf.orig");
+        assert_script_run(q@sed -i 's#^\s*IMAGE=.*$#IMAGE=@ . $kvm_image . q@#g' /etc/kvm-container.conf@);
+        assert_script_run("grep ^IMAGE= /etc/kvm-container.conf");
+        save_screenshot;
+        record_info("Change /etc/kvm-container.conf default IMAGE!", "New value is IMAGE=$kvm_image.");
+    }
+
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.orig");
+    assert_script_run(q@sed -i 's#podman create#podman create --volume /dev/sshserial:/dev/sshserial #g' /usr/local/bin/kvm-container-manage.sh@);
+    script_run("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.orig");
+    save_screenshot;
+    record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.");
+    assert_script_run("kvm-container-manage.sh create");
+    save_screenshot;
+    record_info("Create KVM container successfully!", script_output("$ct_engine ps --all"));
+}
+
+sub cleanup_host_and_kvm_container {
+    assert_script_run("if $ct_engine ps | grep libvirtd; then kvm-container-manage.sh stop;fi");
+    assert_script_run("if $ct_engine ps --all | grep libvirtd; then kvm-container-manage.sh rm;fi");
+    assert_script_run("if which kvm-container-manage.sh;then kvm-container-manage.sh uninstall;fi");
+    validate_script_output("$ct_engine ps --all", sub { $_ !~ 'libvirtd' });
+    save_screenshot;
+    record_info("Remove KVM container and related files from host successfully!");
+}
+
+sub start_kvm_container {
+    assert_script_run("$ct_engine start libvirtd");
+    save_screenshot;
+    assert_script_run("$ct_engine ps | grep -i libvirtd | grep Up");
+    save_screenshot;
+    record_info("Start kvm container successfully!");
+}
+
+sub enter_kvm_container_sh {
+    enter_cmd("clear;$ct_engine exec -ti libvirtd bash", 15);    # interactive mode, so no return immediately
+    save_screenshot;
+    assert_screen('in-libvirtd-container-bash');
+    assert_script_run("echo Hello from libvirtd container");
+    save_screenshot;
+
+    assert_script_run("ip a");
+    save_screenshot;
+    assert_script_run("ip route show all");
+    save_screenshot;
+    assert_script_run("virsh list --all");
+    record_info('Please check if it is from within libvirtd container');
+}
+
+sub exit_kvm_container {
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+    enter_cmd("exit");
+    wait_still_screen 3;
+    save_screenshot;
+    assert_screen('back-to-host-shell');
+}
+
+sub install_tools_within_kvm_container {
+    my $_tools_to_install = shift;
+
+    $_tools_to_install //= "openssh hostname gawk supportutils";
+
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+    zypper_call("in $_tools_to_install", 120);
+    save_screenshot;
+    record_info("Install tools within VT container successfully!");
+}
+
+sub setup_services_within_kvm_container {
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+
+    # Stop all pre-start virtual network
+    clean_all_virt_networks;
+    # Setup nat virtual network as needed
+    my $_vnet_loc = "/etc/libvirt/qemu/networks/alp-nat-vnet.xml";
+    assert_script_run("curl " . data_url("virt_autotest/virtual_network_configs/alp-nat-vnet.xml") . " -o $_vnet_loc");
+    assert_script_run("virsh net-define $_vnet_loc");
+    assert_script_run("virsh net-autostart test-virt-net");
+    assert_script_run('virsh net-start test-virt-net');
+    save_screenshot;
+    record_info("NAT virtual network test-virt-net is set successfully for automation test.", script_output("ip a; ip route show all"));
+}
+
+sub setup_kvm_container_from_scratch {
+    pull_kvm_container_image;
+    config_host_and_kvm_container;
+    start_kvm_container;
+    enter_kvm_container_sh;
+    install_tools_within_kvm_container;
+    setup_services_within_kvm_container;
+}
+
+sub clean_and_resetup_kvm_container {
+    cleanup_host_and_kvm_container;
+    setup_kvm_container_from_scratch;
+}
+
+sub collect_kvm_container_setup_logs {
+    reset_consoles;
+    use_ssh_serial_console;
+
+    my $_kvm_container_log_file = "/tmp/kvm_container.log";
+    my @_cmds = (
+        "$ct_engine ps --all",
+        "$ct_engine images -a --digests",
+        "$ct_engine inspect libvirtd",
+        "$ct_engine logs libvirtd"
+    );
+
+    script_run("rm $_kvm_container_log_file");
+    foreach my $_cmd (@_cmds) {
+        script_run("echo \"Executing $_cmd:\" >>$_kvm_container_log_file");
+        script_run("$_cmd >> $_kvm_container_log_file 2>&1");
+        script_run(qq@echo "\n\n" >> @ . $_kvm_container_log_file);
+    }
+    save_screenshot;
+
+    my @_logs_to_upload = (
+        $_kvm_container_log_file,
+        "/etc/kvm-container.conf",
+        "/usr/local/bin/kvm-container-manage.sh",
+        "/etc/libvirt/qemu/networks/alp-nat-vnet.xml",
+        "/usr/share/containers/containers.conf"
+    );
+    foreach my $_file (@_logs_to_upload) {
+        upload_logs("$_file", failok => 1);
+    }
+    save_screenshot;
+}
+
+1;
+
--- a/lib/virt_autotest/virtual_network_utils.pm
+++ b/lib/virt_autotest/virtual_network_utils.pm
@@ -32,7 +32,7 @@ our @EXPORT
   = qw(download_network_cfg prepare_network restore_standalone destroy_standalone restart_network
   restore_guests restore_network destroy_vir_network restore_libvirt_default enable_libvirt_log pload_debug_log
   check_guest_status check_guest_module check_guest_ip save_guest_ip test_network_interface hosts_backup
-  hosts_restore get_free_mem get_active_pool_and_available_space);
+  hosts_restore get_free_mem get_active_pool_and_available_space clean_all_virt_networks);
 
 sub check_guest_ip {
     my ($guest, %args) = @_;
@@ -329,6 +329,19 @@ sub get_active_pool_and_available_space {
     # default available pool unit as GiB
     $available_size = ($pool_unit eq "TiB") ? int($available_size * 1024) : int($available_size);
     return ($active_pool, $available_size);
+}
+
+sub clean_all_virt_networks {
+    my $_virt_networks = script_output("virsh net-list --name --all", 30, type_command => 0, proceed_on_failure => 0);
+
+    foreach my $vnet (split(/\n+/, $_virt_networks)) {
+        my $_br = script_output(q@virsh net-dumpxml @ . $vnet . q@|grep -o "bridge name=[^\s]*" | sed  's#bridge name=##'@, type_command => 0, proceed_on_failure => 0);
+        script_run("virsh net-destroy $vnet");
+        assert_script_run("virsh net-undefine $vnet");
+        assert_script_run("if ip a|grep $_br;then ip link del $_br;fi");
+        save_screenshot;
+    }
+    record_info("All existing virtual networks: \n$_virt_networks \nhave been destroy and undefined.", script_output("ip a; ip route show all"));
 }
 
 1;
--- a/schedule/virt_autotest/alp-virt-core-tests.yaml
+++ b/schedule/virt_autotest/alp-virt-core-tests.yaml
@@ -1,0 +1,9 @@
+name:           alp-virt-core-tests
+description:    >
+    Maintainer: xlai@suse.com, qe-virt@suse.de
+    ALP virtualization guest installation and virt feature tests schedule
+schedule:
+    - virt_autotest/login_console
+    - virt_autotest/setup_kvm_container
+
+# To be extended
--- a/data/virt_autotest/virtual_network_configs/alp-nat-vnet.xml
+++ b/data/virt_autotest/virtual_network_configs/alp-nat-vnet.xml
@@ -1,0 +1,20 @@
+<network>
+  <!--name can't have vnet keyword, to not conflict with virtual network test-->
+  <name>test-virt-net</name>
+  <uuid>cd1d0029-4ff9-4a24-9a3d-9f0877cb8477</uuid>
+  <forward mode='nat'>
+    <nat>
+      <port start='1024' end='65535'/>
+    </nat>
+  </forward>
+  <bridge name='br123' stp='on' delay='0'/>
+  <mac address='52:54:00:ea:a8:7e'/>
+  <ip address='192.168.123.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.123.2' end='192.168.123.254'>
+        <lease expiry='12' unit='hours'/>
+      </range>
+    </dhcp>
+  </ip>
+</network>
+
--- a/tests/virt_autotest/setup_kvm_container.pm
+++ b/tests/virt_autotest/setup_kvm_container.pm
@@ -1,0 +1,37 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: This file handles the setup for kvm workload container for virtualization test.
+# Maintainer: alice <xlai@suse.com>, qe-virt@suse.de
+
+package setup_kvm_container;
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use alp_workloads::kvm_workload_utils qw(set_kvm_container_image clean_and_resetup_kvm_container collect_kvm_container_setup_logs);
+
+sub run {
+    my $self = shift;
+
+    if (get_var('KVM_WORKLOAD_IMAGE', '')) {
+        set_kvm_container_image(get_var('KVM_WORKLOAD_IMAGE'));
+    }
+
+    clean_and_resetup_kvm_container;
+}
+
+sub test_flags {
+    return {fatal => 1};
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+
+    collect_kvm_container_setup_logs;
+}
+
+1;
+


lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+  setup_services_within_kvm_container
+  collect_kvm_container_setup_logs
+  setup_kvm_container_from_scratch
+  clean_and_resetup_kvm_container
+);
+
+# Container engine: can be podman(default), docker(not tested)
+our $ct_engine = "podman";
+
+my $DEFAULT_KVM_IMAGE = "registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/kvm:latest";
```
waynechen55, line 41: It seems that there is a published one:
```
registry.opensuse.org/suse/alp/workloads/publish/tumbleweed_containerfiles/suse/alp/workloads/kvm:latest
```
Why not use this one ?

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    die "Container engine can only be set to podman or docker!" if ($ct_engine ne "podman" && $ct_engine ne "docker");
+}
+
+sub set_kvm_image {
+    my $_image = shift;
+
+    # Allow customization in case of special test needs
+    if ($_image ne $DEFAULT_KVM_IMAGE) {
+        $kvm_image = $_image;
+        record_info("The kvm container image has been changed from default!", "Now test is using kvm image: $kvm_image.");
```
waynechen55, line 56: Remove "The" to make text shorter and more info exposed at the first glimpse.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    assert_script_run(q@sed -i 's#podman create#podman create --volume /dev/sshserial:/dev/sshserial #g' /usr/local/bin/kvm-container-manage.sh@);
+    script_run("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    save_screenshot;
+    record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.");
+    assert_script_run("kvm-container-manage.sh create");
+    save_screenshot;
+    record_info("KVM container is created!");
+}
```
waynechen55, line 91: 1. I am not sure whether this is the official way to create and run container, namely by using some built-in scripts. These scripts might be experimental. And we lost freedom to customize podman command line arguments which means less flexibility, especially some special settings needs to be used.
2. Maybe APPLIANCE_MIRROR, APPLIANCE and DOMAIN need to customize under certain circumstances ?

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    assert_script_run(q@sed -i 's#podman create#podman create --volume /dev/sshserial:/dev/sshserial #g' /usr/local/bin/kvm-container-manage.sh@);
+    script_run("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    save_screenshot;
+    record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.");
+    assert_script_run("kvm-container-manage.sh create");
+    save_screenshot;
+    record_info("KVM container is created!");
```
waynechen55, line 90: I think all your parameters replacement work can be wrapped up into a separate function which has named arguments, not only for clear view but also relieve effort on us in the future. Otherwise, I expect this block will continue to grow with more and more "replacement work".

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: This published one is for releasing prototypes(september, december prototypes), at least for now. The  one given in code is the rolling release, like daily build, which is used in O3 ALP job gorup https://openqa.opensuse.org/group_overview/100. By default, the code aims to test daily build. So set it as default.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+}
+
+sub cleanup_host_and_kvm_container {
+    assert_script_run("if $ct_engine ps | grep libvirtd; then kvm-container-manage.sh stop;fi");
+    assert_script_run("if $ct_engine ps --all | grep libvirtd; then kvm-container-manage.sh rm;fi");
+    assert_script_run("if which kvm-container-manage.sh;then kvm-container-manage.sh uninstall;fi");
```
waynechen55, line 96: I think currently all your assert_script_run related to "container/vm" operations depend on some built-in scripts, so does its return value. So whether it fails or not depends on the accuracy of scripts instead of the actual commands executed, which looks to me a little bit unreliable unless those scripts are built with "testing activity" in mind. But to my best knowledge, they are just "a demo".

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: > 1. I am not sure whether this is the official way to create and run container, namely by using some built-in scripts. These scripts might be experimental. And we lost freedom to customize podman command line arguments which means less flexibility, especially some special settings needs to be used.

Good question! I have same concern months ago. But in the past several months' progress in alp project for multiple containers -- cockpit, firewall, yast, this is becoming widely used way. And such usage is given in suse's alp official document, which means suse is recommending users to use this way(or default way). That's why finally , in our automation, I carefully choose this way. 

> 2. Maybe APPLIANCE_MIRROR, APPLIANCE and DOMAIN need to customize under certain circumstances ? Also cpu and memory which are directly related to vm creation. Currently all settings are closely coupled together.

The concern on this is valid for now, even for future. The kvm container's script for vm installation is limited for DEMO purpose only, which is not flexible at all. So in alp automation test, we won't use it. Instead , we extend your openqa guest installation to test GI in ALP.

So no need to touch APPLIANCE_MIRROR, APPLIANCE and DOMAIN .

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I understand your concern. When carefully choosing this way, it is based on the judgement that it is quite likely that for kvm container creation, it will stay this way. And I do not see the need that we will customize kvm container a lot. Otherwise, in automation, direct kvm container creation command will be used with extreme flexibility. But here, we need to test the "default"/recommended kvm container.
In short, I predict that for this function, there will be only these two major change blocks, which is not big, and can well fit into the function's name. I would not prefer to wrap up into smaller functions. But if you insist, I can do that.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+
+sub pull_kvm_container {
+    # Ensure large enough time is given to download
+    # TODO: uncomment once automation is done
+    # assert_script_run("$ct_engine pull $kvm_image", 1800/get_var('TIMEOUT_SCALE', 1));
+    record_info("Kvm container image is successfully downloaded.", "Image: $kvm_image.");
+
+}
+
+sub config_host_and_kvm_container {
```
waynechen55, line 68: I think ```create_kvm_container``` is enough for the name

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    assert_script_run("$ct_engine start libvirtd");
+    save_screenshot;
+    assert_script_run("$ct_engine ps | grep -i libvirtd | grep Up");
+    save_screenshot;
+}
+
+sub enter_kvm_container_sh {
+    enter_cmd("clear;$ct_engine exec -ti libvirtd bash", 15);    # interactive mode, so no return immediately
+    save_screenshot;
+    assert_screen('in-libvirtd-container-bash');
```
waynechen55, line 111: It this ```in-libvirtd-container-bash``` a valid command ?

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: This is true. I am afraid this is what we have to tolerate for proof of concept phase produts' automation. Such unreliable things for now, are just what we need to test(eg kvm-container-manage.sh uninstall/install to install and uninstall kvm container) and what we have to rely on for test. Maybe fragile, but if it fails, it is the moment that bugs are likely found.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    assert_script_run("echo Hello from libvirtd container");
+    save_screenshot;
+
+    assert_script_run("ip a");
```
waynechen55, line 115: You can also print ```ip route show all``` out.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: There is more than only kvm configuration and creation(actually podman create only kind of 'define' the container). The host is massively setup for the kvm container by "$ct_engine container runlabel install $kvm_image".

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: It is in assert_screen, just a needle tag name.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    my $_tools_to_install = shift;
+
+    $_tools_to_install = "curl openssh hostname gawk supportutils" if (!$_tools_to_install);
+
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+    zypper_call("in $_tools_to_install", 120);
+    save_screenshot;
+    record_info("Successfully installed test needed tools within VT container.");
+
```
waynechen55, line 141: I am a little bit concerned with there is no ping (or other useful commands) in kvm container. It might not be necessary or is not going to be used at all if you only manipulate vm from on host instead of inside container, especially you only use "host network".

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    wait_still_screen 3;
+    save_screenshot;
+    assert_screen('back-to-host-shell');
+
+}
+
+sub install_tools_within_kvm_container {
+    my $_tools_to_install = shift;
+
+    $_tools_to_install = "curl openssh hostname gawk supportutils" if (!$_tools_to_install);
```
waynechen55, line 134: I think
```
$_tools_to_install //= "curl openssh hostname gawk supportutils";
```
is better

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    # Stop all pre-start virtual network
+    clean_all_virt_networks;
+    # Setup nat virtual network as needed
+    my $_vnet_loc = "/etc/libvirt/qemu/networks/alp-nat-vnet.xml";
+    assert_script_run("curl " . data_url("virt_autotest/alp-nat-vnet.xml") . " -o $_vnet_loc");
+    assert_script_run("virsh net-define $_vnet_loc");
+    assert_script_run("virsh net-autostart test-virt-net");
+    assert_script_run('virsh net-start test-virt-net');
+    save_screenshot;
+    record_info("Successfully setup nat virtual network test-virt-net for automation test.", script_output("ip a"));
```
waynechen55, line 157: Please also print ```ip route show all``` out.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+    zypper_call("in $_tools_to_install", 120);
+    save_screenshot;
+    record_info("Successfully installed test needed tools within VT container.");
+
+}
+
+sub setup_services_within_kvm_container {
```
waynechen55, line 144: How about make this function more flexible by introducing  argument like $network_file or something ? I think some high-level settings for ALP test can be test suite level settings, for example, container engine, kvm container network, network inside kvm container and etc. Not too much.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    assert_script_run('virsh net-start test-virt-net');
+    save_screenshot;
+    record_info("Successfully setup nat virtual network test-virt-net for automation test.", script_output("ip a"));
+}
+
+sub setup_kvm_container_from_scratch {
+    pull_kvm_container;
+    config_host_and_kvm_container;
+    start_kvm_container;
+    enter_kvm_container_sh;
```
waynechen55, line 164: I think the name ```enter_kvm_container``` is enough.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+        script_run(qq@echo "\n\n" >> @ . $_kvm_container_log_file);
+    }
+    save_screenshot;
+    upload_logs("$_kvm_container_log_file");
+
+    upload_logs("/etc/kvm-container.conf");
+    upload_logs("/usr/local/bin/kvm-container-manage.sh");
+    upload_logs("/etc/libvirt/qemu/networks/alp-nat-vnet.xml");
+    save_screenshot;
+}
```
waynechen55, line 199: Regarding "libvirtd container" log, I think you can create a systemd unit for "kvm container" by "podman generate systed", so "libvirtd container" can be controlled as well on host by systemctl. There will be service logs for this "libvirtd container" on host, can you can also make "kvm container" start up on system boot/reboot/power cycle by "systemctl enable". Very convenient and handy.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+        script_run("echo \"Executing $cmd:\" >>$_kvm_container_log_file");
+        script_run("$cmd >> $_kvm_container_log_file 2>&1");
+        script_run(qq@echo "\n\n" >> @ . $_kvm_container_log_file);
+    }
+    save_screenshot;
+    upload_logs("$_kvm_container_log_file");
+
+    upload_logs("/etc/kvm-container.conf");
+    upload_logs("/usr/local/bin/kvm-container-manage.sh");
+    upload_logs("/etc/libvirt/qemu/networks/alp-nat-vnet.xml");
```
waynechen55, line 197: There are some other files in 
```
/usr/share/containers
```
needs to pay attention.

lib/virt_autotest/virtual_network_utils.pm
```diff
@@ -331,4 +331,17 @@ sub get_active_pool_and_available_space {
...
 }
 
+sub clean_all_virt_networks {
+    my $_virt_networks = script_output("virsh net-list --name --all", 30, type_command => 0, proceed_on_failure => 0);
+
+    foreach my $vnet (split(/\n+/, $_virt_networks)) {
+        my $_br = script_output(q@virsh net-dumpxml @ . $vnet . q@|grep -o "bridge name=[^\s]*" | sed  's#bridge name=##'@, type_command => 0, proceed_on_failure => 0);
+        script_run("virsh net-destroy $vnet");
+        assert_script_run("virsh net-undefine $vnet");
+        assert_script_run("if ip a|grep $_br;then ip link del $_br;fi");
```
waynechen55, line 341: I think you can
```
ip addr flush
ip link down
ip link del
```
The sequence looks to me more elegant.

schedule/virt_autotest/alp-virt-core-tests.yaml
```diff
@@ -0,0 +1,9 @@
+name:           alp-virt-core-tests
+description:    >
+    Maintainer: xlai@suse.com, qe-virt@suse.de
+    ALP virtualization guest installation and virt feature tests schedule
+schedule:
+    - virt_autotest/login_console
+    - virt_autotest/setup_kvm_container
```
waynechen55, line 7: Will this test module setup_kvm_container work on SLES/SLE-Micro/Tumbleweed/ALP host ?

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: Add some checkpoints after each "script" can alleviate this problem, for example, check container kvm is created after kvm-container-manage.sh create.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    if ($kvm_image ne $DEFAULT_KVM_IMAGE) {
+        assert_script_run("cp /etc/kvm-container.conf /etc/kvm-container.conf.org");
+        assert_script_run(q@sed -i 's#^\s*IMAGE=.*$#IMAGE=@ . $kvm_image . q@#g' /etc/kvm-container.conf@);
+        assert_script_run("grep ^IMAGE= /etc/kvm-container.conf");
+        save_screenshot;
+        record_info("/etc/kvm-container.conf default IMAGE is changed!", "New value is IMAGE=$kvm_image.");
+    }
+
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
```
waynechen55, line 83: I think this a typo. It should be "orig" instead of "org".

waynechen55: We have a backup_file function in lib/virt_autotest/utils.pm. You can use it directly.

data/virt_autotest/alp-nat-vnet.xml
```diff
@@ -0,0 +1,20 @@
...
+  <mac address='52:54:00:ea:a8:7e'/>
+  <ip address='192.168.123.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.123.2' end='192.168.123.254'>
+        <lease expiry='12' unit='hours'/>
+      </range>
+    </dhcp>
+  </ip>
+</network>
+
```
waynechen55, line 20: Please put all your configs in a ALP folder which is better.

waynechen55:
Because ALP is going to be a transactional os, I think you should do verification run on tumbleweed transactional and sle micro to have a try.

data/virt_autotest/alp-nat-vnet.xml
alice-suse: I considered this, but it is put here because it is more reasonable -- this config although used by alp, but it is  open to be used by other product's test too, after openqa guest installation code is modified to support other network types like virtual network.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I think more confidence can be given to their demo code. Although it is early phase code, but Dev side has been using such way to do their devel test. Besides, based on my use experience, the return code for each call is reliable. Although sometimes they fail/return fail code, but it is real fail. Then it is enough. Extra check can be saved.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: ok, will add it.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Do not worry. The pkg list is likely to be extended later based on need. Now within automation, nmap is used for vm network connectivity check. If ping is needed, will add later.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: It is already provided within kvm container repo as `https://build.opensuse.org/package/view_file/SUSE:ALP:Workloads/kvm-container/libvirtd.service?expand=1`. Will likely to be recommended to users by default(dev's plan). But now it is  a little buggy. So will add that later in our automation test code.

lib/virt_autotest/virtual_network_utils.pm
alice-suse: Maybe more elegant, but more unnecessary? The last command directly works...

schedule/virt_autotest/alp-virt-core-tests.yaml
alice-suse: Based on project test need, only ALP requies it.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: # ls /usr/share/containers/
containers.conf  mounts.conf  oci  seccomp.json  selinux

Which ones do you find extraly useful?

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: ok, will add it

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I prefer current way -- more explicit I think

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Good point, but currently it lacks enough project requirement and user stories for these needs, eg container engine,there is no clear idea on this on PM side, not given any attention on antoine's side yet, and for container network, antoine expressed that host will be used as the only choice which is used in our automation now. But I am open to refine it when these are required and when it is clear for how.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Only one line for one file backup... maybe current way is more efficient. Will use the suggested way once more backup is needed.

data/virt_autotest/alp-nat-vnet.xml
waynechen55: Then you should use other names instead of alp-nat-vnet.xml. You can use just some general names like vm-nat-vnet.xml or other names. And please be aware that when a same category of things keep growing, you should move them to a separate folder. I think networking config files belong to this category. That will be too late if we do not starting do so at some point.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: I think "do not rely on anything, do what QA should do". If you consider relying solely on script is enough, I do not want to insist on adding additional checkpoint. I just think there is no enough and explicit checkpoints both in script and test module which looks not good to me by and large.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: containers.conf I think.

data/virt_autotest/alp-nat-vnet.xml
alice-suse: I thought about general names like  vm-nat-vnet.xm. In automation, it is open to use other virtual network config. If someone picks other config, and name the file specially, then how would we interpret this vm-nat-vnet.xm? So I name it alp-nat-vnet.xml, which means it is the customized vnet config file for alp, although it is capable to be reused in other products like sles. Make sense? 
Besides, in reality, I think we won't have too much such virtual network config files, at most several. And Leon's several [network].xml stay static like that(won't grow for years). Do you think it is necessary to add a sub-dir for several files? Personally I prefer not. 

But if you insist on either item, I can change.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I understand your concern. Maybe let's use it this way for a while to see if it's enough. If not, will add necessary additional checks on fragile parts.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Will add it then.

alice-suse:
> Because ALP is going to be a transactional os, I think you should do verification run on tumbleweed transactional and sle micro to have a try.

ALP will be both transactional and non-transactional.

There is no project requirement to run it on tumbleweed. It is for ALP only.

Currently given validation jobs are run on sle micro(host), because ALP does not have image for host yet, only KVM VM qcow2. However the method for kvm container creation and vm creation have been validated in nested virtualization environment which has ALP VM as L1 host during ALP september prototype test. No worries.

data/virt_autotest/alp-nat-vnet.xml
waynechen55: I do not insist on the names. I think you should put your config files into a separate folder. Just like I said above, starting take actions to organize them better before it is too late. Putting everything in top folder is not very good. For example, a virtual_network_config folder or other names you prefer. We should start getting into habit of organizing files better. Think about it.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+  collect_kvm_container_setup_logs
+  setup_kvm_container_from_scratch
+  clean_and_resetup_kvm_container
+);
+
+# Container engine: can be podman(default), docker(not tested)
+our $ct_engine = "podman";
+
+my $DEFAULT_KVM_IMAGE = "registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/kvm:latest";
+our $kvm_image = $DEFAULT_KVM_IMAGE;
```
waynechen55, line 42: Do you need "our" for $kvm_image variable ?

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+  install_tools_within_kvm_container
+  setup_services_within_kvm_container
+  collect_kvm_container_setup_logs
+  setup_kvm_container_from_scratch
+  clean_and_resetup_kvm_container
+);
+
+# Container engine: can be podman(default), docker(not tested)
+our $ct_engine = "podman";
+
```
waynechen55, line 40: Remove this empty line.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55, line 41: I am not sure whether registry is going to stay the same. If not, I think you also need a variable like "$container_image_registry" ? Then $DEFAULT_KVM_IMAGE = $container_image_registry . "xxxx".

data/virt_autotest/alp-nat-vnet.xml
alice-suse: Okay, will create separate folder for virtual_network_config. This is good direction anyway.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I think so, in case any other need for it elsewhere.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+my $DEFAULT_KVM_IMAGE = "registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/kvm:latest";
+our $kvm_image = $DEFAULT_KVM_IMAGE;
+
+sub set_ct_engine {
+    $ct_engine = shift;
```
waynechen55, line 45: I just found your function definition is not consistent. For example,
```
sub set_ct_engine {
    $ct_engine = shift;
```
and
```
sub set_kvm_image {
    my $_image = shift;
```

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I do not see the need now or near future. Besides, the registry info is already binded in kvm_image.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: where is not consistent? I can't tell ...

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+
+    die "Container engine can only be set to podman or docker!" if ($ct_engine ne "podman" && $ct_engine ne "docker");
+}
+
+sub set_kvm_image {
```
waynechen55, line 50: I think
```
set_kvm_container_image
```
is better.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    my $_image = shift;
+
+    # Allow customization in case of special test needs
+    if ($_image ne $DEFAULT_KVM_IMAGE) {
+        $kvm_image = $_image;
+        record_info("The kvm container image has been changed from default!", "Now test is using kvm image: $kvm_image.");
+    }
+}
+
+sub pull_kvm_container {
```
waynechen55, line 60: I think
```
pull_kvm_container_image
```
is better.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: Nothing about set_ct_engine function. Compared with this:
```
sub check_guest_ip {
    my ($guest, %args) = @_;
```
variables with or without underscore.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: Generally speaking, I think your function names should differentiate "image" from "container".

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: My mistake. They are in different module. Ignore this comment.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55, line 96: I do not understand this
```
if which kvm-container-manage.sh
```

waynechen55:
I think kvm container setup and config functions can be made more generic. A rough example:
```
sub setup_container {
}

sub setup_kvm_container {
    setup_container(libvirtd)
}
```
It does not introduce complexity. And it can help lay the ground for all our future container testing.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Will use your suggested name.

I had a double check, set_kvm_image and this function are the only two functions related with image. After updating this one, all will be fine.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: if kvm-container-manage.sh exists. 
This cleanup function can be run on whatever environment, kvm container deployed, half-cleaned, or not exist. This is to ensure assert_script_run returns correct value.

alice-suse:
> I think kvm container setup and config functions can be made more generic. A rough example:
> 
> ```
> sub setup_container {
> }
> 
> sub setup_kvm_container {
>     setup_container(libvirtd)
> }
> ```
> 
> It does not introduce complexity. And it can help lay the ground for all our future container testing.

Virtualization only has one container, kvm. No need for virt then. If doing so, the intended usage of setup_container will be for general alp workload container. So it's one more layer abstraction. To make it actually useful, this requires knowledge of how other alp workload containers can be setup and extract general things. But I do not have effort to do it now. And  I do not think it can be well defined at so early alp phase which may change the usage later. I'd recommend to do that later.

lib/alp_workloads/kvm_workload_utils.pm
waynechen55, line 90: ···
record_info("KVM container is created!", script_output("podman image ls"));
···

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+    save_screenshot;
+    record_info("KVM container is created!");
+}
+
+sub cleanup_host_and_kvm_container {
+    assert_script_run("if $ct_engine ps | grep libvirtd; then kvm-container-manage.sh stop;fi");
+    assert_script_run("if $ct_engine ps --all | grep libvirtd; then kvm-container-manage.sh rm;fi");
+    assert_script_run("if which kvm-container-manage.sh;then kvm-container-manage.sh uninstall;fi");
+    save_screenshot;
+    record_info("KVM container is removed and container related files are removed from host");
```
waynechen55, line 98:
```
record_info("KVM container is removed and container related files are removed from host", script_output("podman image ls;podman containers list"));
```

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,201 @@
...
+        save_screenshot;
+        record_info("/etc/kvm-container.conf default IMAGE is changed!", "New value is IMAGE=$kvm_image.");
+    }
+
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    assert_script_run(q@sed -i 's#podman create#podman create --volume /dev/sshserial:/dev/sshserial #g' /usr/local/bin/kvm-container-manage.sh@);
+    script_run("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org");
+    save_screenshot;
+    record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.");
```
waynechen55, line 87:
```
record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.", script_ouput("cat kvm-container-image.sh"));
```

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Debug info already in http://10.67.129.185/tests/326#step/setup_kvm_container/8. I do not like over-checking.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Debug info in http://10.67.129.185/tests/326#step/setup_kvm_container/32.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: The diff in above line is enough for quick check.  kvm-container-image.sh is uploaded once test fails. Too much  info here and there, is likely to be ignored, and will slow down test when accumulated.

waynechen55: If you like diff, I think this way is better:
```
record_info("kvm-container-manage.sh has been changed to share sol console needed device with kvm container.", cript_output("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.org"));
    save_screenshot;);
```
This is more obvious.

data/virt_autotest/alp-nat-vnet.xml
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Keep it for better classification of data.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Not convinced, keep it current way.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: > exactly there is no checking at all actually before adding script_output

NO, assert_script_run("kvm-container-manage.sh create"), already asserts. Script_output is just output, unless human checks, it does not assert anything. Usually in test, no one would check such record_info outputs, unless test fails. So we should rely on such assert_script_run to automatically checks, rather than human involvement in check. 
But here, I will add the script_output of `$ct_engine ps --all` to show all containers, since it is not given in earlier save_screenshot of the executed script.
BTW, script_run and save_screen , with record_info following that, can show what commands are executed, the output, and notification(record_info) to reviewers, which is usually better than relying only on record_info, in which way, it would not tell you the executed command and usually introduces extra command execution. I only pick this way when the previous script has long outputs which in screenshot, people can not get full output.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Will add assert_script_run("!($ct_engine ps --all | grep libvirtd)") to confirm libvirtd container is removed.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: done

lib/alp_workloads/kvm_workload_utils.pm
waynechen55: Adding some sort of output from container engine looks better to test itself now.

alice-suse:
Will push code once validation is done, now it is blocked by bj lab network issue.

alice-suse:
Verification is done based on latest code. Links updated in PR description(top).

alice-suse:
CI check fails at pulling validation container, which is network issue. Will reschedule CI test later.

alice-suse:
@Julie-CAO  @nanzhg @guoxuguang @tbaev @martinsmarcelo @RoyCai7 Hello all, I do not see your comments ;-), if you are interested to give it a look, feel free to do so before EOB next Monday. Otherwise I plan to merge it since it's so carefully reviewed by wayne.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+        assert_script_run(q@sed -i 's#^\s*IMAGE=.*$#IMAGE=@ . $kvm_image . q@#g' /etc/kvm-container.conf@);
+        assert_script_run("grep ^IMAGE= /etc/kvm-container.conf");
+        save_screenshot;
+        record_info("Change /etc/kvm-container.conf default IMAGE!", "New value is IMAGE=$kvm_image.");
+    }
+
+    # Share /dev/sshserial from host to kvm container for let openqa root-ssh console works
+    assert_script_run("cp /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.orig");
+    assert_script_run(q@sed -i 's#podman create#podman create --volume /dev/sshserial:/dev/sshserial #g' /usr/local/bin/kvm-container-manage.sh@);
+    script_run("diff /usr/local/bin/kvm-container-manage.sh /usr/local/bin/kvm-container-manage.sh.orig");
```
guoxuguang, line 84: Is there the debug info from /usr/local/bin/kvm-container-manage.sh

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+    assert_script_run("if $ct_engine ps | grep libvirtd; then kvm-container-manage.sh stop;fi");
+    assert_script_run("if $ct_engine ps --all | grep libvirtd; then kvm-container-manage.sh rm;fi");
+    assert_script_run("if which kvm-container-manage.sh;then kvm-container-manage.sh uninstall;fi");
+    validate_script_output("$ct_engine ps --all", sub { $_ !~ 'libvirtd' });
+    save_screenshot;
+    record_info("Remove KVM container and related files from host successfully!");
+}
+
+sub start_kvm_container {
+    assert_script_run("$ct_engine start libvirtd");
```
guoxuguang, line 102: what is the default status for libvirtd.service  for ALP?  Before start  libvirtd.service on ALP, we'd better to check the status of  libvirtd.service, imho.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+    assert_script_run("ip a");
+    save_screenshot;
+    assert_script_run("ip route show all");
+    save_screenshot;
+    assert_script_run("virsh list --all");
+    record_info('Please check if it is from within libvirtd container');
+}
+
+sub exit_kvm_container {
+    assert_script_run("clear");
```
guoxuguang, line 125: Quick question, why use with `assert_script_run("clear"); ` here?

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: I do not quite understand the latter part of the question. But for this file, as you can see it is possible that the file is updated to customized container image. So it is uploaded for debug if test fails. This file is  used for managing kvm container.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: There is no libvirtd service in ALP kvm container.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Assert_screen is used to ensure/check whether the container exit/entrance is successful or not. Clear can help clean up screens first, which ensures that the latter screenshot check the needed areas, rather than earlier screen output.

lib/virt_autotest/virtual_network_utils.pm
```diff
@@ -331,4 +331,17 @@ sub get_active_pool_and_available_space {
...
+        save_screenshot;
+    }
+    record_info("All existing virtual networks: \n$_virt_networks \nhave been destroy and undefined.", script_output("ip a"));
```
waynechen55, line 344:
```
record_info("All existing virtual networks: \n$_virt_networks \nhave been destroy and undefined.", script_output("ip addr show all; ip route show all"));
```
is better I think.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+    record_info("NAT virtual network test-virt-net is set successfully for automation test.", script_output("ip a; ip route show all"));
+}
+
+sub setup_kvm_container_from_scratch {
+    pull_kvm_container_image;
+    config_host_and_kvm_container;
+    start_kvm_container;
+    enter_kvm_container_sh;
+    install_tools_within_kvm_container;
+    setup_services_within_kvm_container;
```
waynechen55, line 167:
```
install_tools_in_kvm_container;
setup_services_in_kvm_container;
```
is enough.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+    record_info('Please check if it is from within libvirtd container');
+}
+
+sub exit_kvm_container {
+    assert_script_run("clear");
+    assert_screen('in-libvirtd-container-bash');
+    enter_cmd("exit");
+    wait_still_screen 3;
+    save_screenshot;
+    assert_screen('back-to-host-shell');
```
waynechen55, line 130: I think you can use linux command
```
which podman
```
to know whether inside or outside of container. Save the effort to capture a needle.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -0,0 +1,209 @@
...
+    save_screenshot;
+    record_info("Start kvm container successfully!");
+}
+
+sub enter_kvm_container_sh {
+    enter_cmd("clear;$ct_engine exec -ti libvirtd bash", 15);    # interactive mode, so no return immediately
+    save_screenshot;
+    assert_screen('in-libvirtd-container-bash');
+    assert_script_run("echo Hello from libvirtd container");
+    save_screenshot;
```
waynechen55, line 114: I think you can use linux command
```
which podman
```
to know whether inside or outside of container. Save the effort to capture a needle.

alice-suse: No, podman can be installed from within container.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: same here.

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Of course not.

lib/virt_autotest/virtual_network_utils.pm
alice-suse: Can be added.

Julie-CAO:
> @Julie-CAO @nanzhg @guoxuguang @tbaev @martinsmarcelo @RoyCai7 Hello all, I do not see your comments ;-), if you are interested to give it a look, feel free to do so before EOB next Monday. Otherwise I plan to merge it since it's so carefully reviewed by wayne.

Hi @alice-suse I viewed the PR and I am ok with it.

alice-suse:
@guoxuguang I do not see your further reply to open conversations by you. I take it as agree, and mark them resolved.

guoxuguang:
@alice-suse LGTM sorry for my late reply.

PR #15802, opened 2022-10-31
Adapt s390x-zVM-vswitch-l2 cases with install_SLES

hjluo:
We need to add install_SLES to make the s390x-zVM test case to work.

- Related ticket: https://progress.opensuse.org/issues/119587
- Needles: N/A
- Verification run: 
  http://openqa.nue.suse.com/t9842445
  http://openqa.nue.suse.com/t9842446


ge0r (approved):
LGTM

PR #15803, opened 2022-10-31
Return true value in cleanup function

mimi1vx:
If is post_fai_hook running on test module without cleanup override it now logs warning about failed self->cleanup(). This change fixes this -> only real fails are logged.


mloviska (approved):

grisu48 (commented):
LGTM! One typo in the PR description (post_fai_hook) and I would still do a VR despite the change being trivial because it touches the `basetest.pm`.

ilausuch (approved):

grisu48 (approved):

grisu48:
The `post_fai_hook` typo is still there

PR #15806, opened 2022-10-31
Install *-hpc-gnu packages on HPC compute nodes

b10n1k:
Some time ago I implemented a NFS sharing between management and compute
nodes. This had the advantage that we had to install the binaries just once
and make them available immediately to the other nodes. Only requirement was
that the compute nodes were needed to install some dependencies. It turned out
that this might not an optimal solution for those binaries as the dependencies
are likely to cause a hell among the various products. As it happens with the
latest builds on sle15sp5 after some changes on openmpi.

- Verification run: http://aquarius.suse.cz/tests/13747

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -383,11 +383,21 @@ sub uninstall_spack_module {
 This function is used to select dependencies packages which are required to be installed
 on HPC compute nodes in order to run code against particular C<mpi> implementation.
 C<get_compute_nodes_deps> returns an array of packages
+
+=head2 CAVEATS
+
+Obsolete function. Not in use since sle15sp5
+Used to install dependencies of the HPC modules when the binaries were shared
+through NFS. Changes in openmpi breaks this on SLE15SP5. Need to get updated to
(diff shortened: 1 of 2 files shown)


lib/hpcbase.pm
```diff
@@ -380,6 +380,8 @@ sub uninstall_spack_module {
 
   get_compute_nodes_deps($mpi)
 
+Obsolete function. Not in use since sle15sp5
```
Avinesh, line 383: only comment I have is -
why not remove the function itself if is not being called in any test module.

b10n1k: i was not sure if i want to remove it because it works for <sle15sp5 and i was thinking it might be useful somehow if i fix it later on.

Avinesh: okay, may be we should add this info also along with obsolete function.

Avinesh (approved):

lib/hpcbase.pm
b10n1k: updated

PR #15807, opened 2022-10-31
Expect display manager to appear earlier on boot 

ge0r:
Sometimes, on ppc64le architecture with qemu backend, the grub entries will not bounce back and we can land directly to the dispay manager. In order to avoid the occasional failure we must anticipate display manager to appear earlier when running `boot_to_desktop` on ppc64le.

- Related ticket: https://progress.opensuse.org/issues/118336
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=ge0r%2Fos-autoinst-distri-opensuse%23fix-boot-to-desktop-ppc64le-verification 
Verification run link contains testsuites with both textmode and gnome. 
Success rate is 100% for a run of 50 jobs for each testsuite (see Next & Previous results).
Without this fix the success rate is around 54%.

### DIFF
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -269,6 +269,13 @@ sub boot_local_disk {
         }
         my @tags = qw(inst-slof grub2);
         push @tags, 'encrypted-disk-password-prompt' if (get_var('ENCRYPT'));
+
+        # Workaround for poo#118336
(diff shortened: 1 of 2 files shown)


rakoenig (approved):
LGTM.

coolgw (approved):

PR #15812, opened 2022-10-31
Fix script_sudo condition from 739deb2a4b

b10n1k:
The refactoring of the condition which took place in 739deb2a4b should have
include both statements as it was stated in first place, and not pass
redirection to serial device.

- Verification run: http://aquarius.suse.cz/tests/13762 
http://aquarius.suse.cz/tests/13761


Vogtinator (commented):
+1

PR #15815, opened 2022-11-01
Use 'run_args' in Public Cloud instead of custom provider

mimi1vx:
This removes need to define in PC tests:
`$self->{provider} = $args->{my_provider};    # required for cleanup`

VR GCE:
- http://quasar.suse.cz/tests/1214
- http://quasar.suse.cz/tests/1213
- registration -> http://quasar.suse.cz/tests/1215

VR Azure:
- registration -> http://quasar.suse.cz/tests/1220
- http://quasar.suse.cz/tests/1238
- http://quasar.suse.cz/tests/1239

VR EC2:
- registration -> http://quasar.suse.cz/tests/1221
- http://quasar.suse.cz/tests/1223
- http://quasar.suse.cz/tests/1240

### DIFF
--- a/lib/publiccloud/basetest.pm
+++ b/lib/publiccloud/basetest.pm
@@ -120,8 +120,8 @@ sub _cleanup {
     return if ($flags->{publiccloud_multi_module} && !($self->{result} eq 'fail' && $flags->{fatal}));
     # 2. Job should have PUBLIC_CLOUD_NO_CLEANUP defined and job should have result = 'fail'
     return if ($self->{result} eq 'fail' && get_var('PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE'));
-    if ($self->{provider}) {
-        eval { $self->{provider}->cleanup(); } or bmwqemu::fctwarn("provider::cleanup() failed -- $@");
+    if ($self->{run_args} && $self->{run_args}->{my_provider}) {
+        eval { $self->{run_args}->{my_provider}->cleanup(); } or bmwqemu::fctwarn("provider::cleanup() failed -- $@");
     }
 }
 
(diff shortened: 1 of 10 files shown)


grisu48 (commented):
While the changes look OK to me, I have a hard time to be sure that this will work like this for all changed tests. Can you extend the VRs so that we are sure, that the cleanup routines are working there? Best would be one VR per changed test module, but if you find a better way I'm open for a smaller set of VRs.

The provided ones are not sufficient IMHO to show that nothing breaks here.

mimi1vx:
@grisu48 I think this set is representative:
- `instance_overview` .. simple module without any cleanup override and with full rest of hook
- 'run_ltp` - full hook + custom cleanup
-  `registration` - partial hook, but rewritten custom `post_fail_hook` to use standard `cleanup` override

grisu48:
> @grisu48 I think this set is representative:
> 
>     * `instance_overview` .. simple module without any cleanup override and with full rest of hook
> 
>     * `run_ltp` - full hook + custom cleanup
> 
>     * `registration` - partial hook, but rewritten custom `post_fail_hook` to use standard `cleanup` override
> 
> 
> And this changes in reality are valid only for tests with flag `fatal => 1` and without `publiccloud_multi_module => 1`
> 
> but you're right , we need add VR also for other providers ...... so I'll prepare some for Azure and EC2

Thanks for the elaboration. Yes, I agree, this should be exhaustive testing. If you would not mind adding 1-2 test runs from either Azure or EC2 that we also check if the provider handling works for another platform, then this can fly :rocket:

grisu48 (approved):

pdostal:
This nicely simplifies the code, cool 👍

PR #15823, opened 2022-11-02
PublicCloud - don't run full cleanup when result is undef

mimi1vx:
VR:

with publiccloud_module only : http://quasar.suse.cz/tests/1236#

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -12,6 +12,7 @@ use testapi qw(is_serial_terminal :DEFAULT);
 use Mojo::Base -base;
 use publiccloud::instance;
 use publiccloud::instances;
+use publiccloud::ssh_interactive 'select_host_console';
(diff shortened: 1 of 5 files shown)


lib/publiccloud/basetest.pm
```diff
@@ -113,6 +113,7 @@ sub _cleanup {
 
     eval { $self->cleanup(); } or bmwqemu::fctwarn("self::cleanup() failed -- $@");
 
+    return if !($self->{result});    # skip when result is undef
```
asmorodskyi, line 116: I vote for dropping this line with `publiccloud_multi_module` all should be fine without it

asmorodskyi: I agree that bigger refactoring of this logic is needed and I agree that current approach is messy but this "bigger refactoring" won't contain this line either ...

grisu48 (approved):

mimi1vx:
...

PR #15828, opened 2022-11-03
disable GRUB_TIMEOUT in grub_test_snapshot.pm

badboywj:
**Desciption**:
grub_test_snapshot.pm failed in grub boot menu because of timeout like this https://openqa.suse.de/tests/9759181#step/grub_test_snapshot/82, so I disable GRUB_TIMEOUT in the module to fix the issue.

- Related ticket:
> https://progress.opensuse.org/issues/119329

- Verification run:
> https://openqa.suse.de/tests/9863029
> https://openqa.suse.de/tests/9865632
> https://openqa.suse.de/tests/9869445
> https://openqa.suse.de/tests/9869566

### DIFF
--- a/tests/boot/grub_test_snapshot.pm
+++ b/tests/boot/grub_test_snapshot.pm
@@ -13,13 +13,16 @@ use base 'opensusebasetest';
 use testapi;
 use power_action_utils 'power_action';
 use utils qw(workaround_type_encrypted_passphrase reconnect_mgmt_console);
-use bootloader_setup qw(stop_grub_timeout boot_into_snapshot);
+use bootloader_setup qw(stop_grub_timeout boot_into_snapshot change_grub_config);
 use Utils::Backends 'is_pvm';
+use Utils::Architectures qw(is_aarch64);
 
 sub run {
     my $self = shift;
 
     select_console 'root-console';
+    # disable GRUB_TIMEOUT to avoid this module failed since grub menu timeout
+    change_grub_config('=.*', '=-1', 'GRUB_TIMEOUT', '', 1) if (is_aarch64);
     power_action('reboot', keepconsole => 1, textmode => 1);
     reset_consoles;
     reconnect_mgmt_console if is_pvm;


tests/boot/grub_test_snapshot.pm
```diff
@@ -13,13 +13,15 @@ use base 'opensusebasetest';
...
-use bootloader_setup qw(stop_grub_timeout boot_into_snapshot);
+use bootloader_setup qw(stop_grub_timeout boot_into_snapshot change_grub_config);
 use Utils::Backends 'is_pvm';
 
 sub run {
     my $self = shift;
 
     select_console 'root-console';
+    # disable GRUB_TIMEOUT to avoid this module failed since grub menu timeout
+    change_grub_config('=.*', '=-1', 'GRUB_TIMEOUT', '', 1);
```
coolgw, line 24: if only happen on aarch64, we can add "if arch==aarch64."

rakoenig: I agree with @coolgw, on the other hand would it do any harm if we disable the timeout on all architectures?
If we don't probe for a specific architecture and disable grub unconditionally, then we should think if the line with `stop_grub_timeout ` later in the test module is still needed.

badboywj: yes, you are right, I agree with you, I have improved the code.

tests/boot/grub_test_snapshot.pm
```diff
@@ -13,13 +13,16 @@ use base 'opensusebasetest';
...
+use bootloader_setup qw(stop_grub_timeout boot_into_snapshot change_grub_config);
 use Utils::Backends 'is_pvm';
+use Utils::Architectures qw(is_aarch64);
 
 sub run {
     my $self = shift;
 
     select_console 'root-console';
+    # disable GRUB_TIMEOUT to avoid this module failed since grub menu timeout
+    change_grub_config('=.*', '=-1', 'GRUB_TIMEOUT', '', 1) if (is_aarch64);
```
ge0r, line 25: There is no need to add the fourth argument `''` since it is already set to this in the [change_grub_config definition](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/bootloader_setup.pm#L1224). Still it doesn't hurt either.

ge0r (approved):
Looks good!

coolgw (approved):

PR #15839, opened 2022-11-04
Select last item on select_patterns

ge0r:
Until now the workaround did not select the last item because the up button was pressed last. With this new fix we will only send up when the screen did not change after we have typed the second down, which means that we are on the last item. This will result with the last item being selected.

- Related ticket: https://progress.opensuse.org/issues/119584, related to [this failure](https://openqa.suse.de/tests/9842779#step/select_patterns/125).
- Needles: No
- Verification run: https://openqa.suse.de/tests/9936092#step/select_patterns/127

related MR: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1630

### DIFF
--- a/lib/y2_installbase.pm
+++ b/lib/y2_installbase.pm
@@ -181,11 +181,15 @@ highlight cursor one item down.
 =cut
 
 sub move_down {
-    # Sending 'down' twice, followed by 'up', scrolls to the intended item.
-    wait_screen_change { send_key 'down' } if is_sle('>=15-sp3');
     my $ret = wait_screen_change { send_key 'down' };
     last if (!$ret);    # down didn't change the screen, so exit here
-    wait_screen_change { send_key 'up' } if is_sle('>=15-sp3');
+    if (is_sle('>=15-sp3')) {
+        # Sending 'down' twice, followed by 'up', scrolls to the intended item.
+        $ret = wait_screen_change { send_key 'down' };
+        # If screen did not change with the second down we are on the last item.
+        # In that case do not press up so that it remains selected.
+        wait_screen_change { send_key 'up' } if $ret;
+    }
     check12qtbug if check_var('VERSION', '12');
 }
 


lib/y2_installbase.pm
```diff
@@ -182,10 +182,14 @@ highlight cursor one item down.
...
 sub move_down {
     # Sending 'down' twice, followed by 'up', scrolls to the intended item.
-    wait_screen_change { send_key 'down' } if is_sle('>=15-sp3');
     my $ret = wait_screen_change { send_key 'down' };
     last if (!$ret);    # down didn't change the screen, so exit here
-    wait_screen_change { send_key 'up' } if is_sle('>=15-sp3');
+    if (is_sle('>=15-sp3')) {
+        my $ret2 = wait_screen_change { send_key 'down' };
+        # If screen did not change with the second down we are on the last item.
+        # In that case do not press up so that it remenains selected.
```
jknphy, line 190: s/remenains/remains/

lib/y2_installbase.pm
```diff
@@ -182,10 +182,14 @@ highlight cursor one item down.
...
+        wait_screen_change { send_key 'up' } if !$ret2;
```
jknphy, line 191: `if $ret2` ?

lib/y2_installbase.pm
```diff
@@ -182,10 +182,14 @@ highlight cursor one item down.
 
 sub move_down {
     # Sending 'down' twice, followed by 'up', scrolls to the intended item.
```
jknphy, line 184: that comment now is orphan, should be above the second down.

lib/y2_installbase.pm
ge0r: I don't think so, if the screen changed, we send the key up, otherwise the screen did not change, so we are in the last item.

ge0r: ah actually yes, this is what I meat to write but did not in the code

ge0r:
It turns out that the issue was this:
A second tag in the form of `pattern-$name_of_pattern` needs to be there, at the needle json. Once the needle that was created by the team was update with this secondary needle, the yast-development pattern is set up for selection and then is selected.

lib/y2_installbase.pm
```diff
@@ -181,11 +181,15 @@ highlight cursor one item down.
...
 
 sub move_down {
-    # Sending 'down' twice, followed by 'up', scrolls to the intended item.
-    wait_screen_change { send_key 'down' } if is_sle('>=15-sp3');
     my $ret = wait_screen_change { send_key 'down' };
     last if (!$ret);    # down didn't change the screen, so exit here
-    wait_screen_change { send_key 'up' } if is_sle('>=15-sp3');
+    if (is_sle('>=15-sp3')) {
+        # Sending 'down' twice, followed by 'up', scrolls to the intended item.
+        my $ret2 = wait_screen_change { send_key 'down' };
```
jknphy, line 188: we can reuse `$ret`.

jknphy:
LGTM

PR #15840, opened 2022-11-04
Add a test suite to create a hdd for sap migration

JRivrain:
See https://progress.opensuse.org/issues/119683

For sle12 we need a specific medium for SAP, it is not an addon like in sle15. This profile was cloned from qam-create_hdd_sles4sap_gnome, then all useless (default) parts were removed, otherwise the file had about 1800 lines.

VRs available here (flavor=SAP-DVD-Updates)
https://openqa.suse.de/tests/overview?distri=sle&version=12-SP5&build=20221117-1&groupid=446
https://openqa.suse.de/tests/overview?distri=sle&version=12-SP4&build=20221117-1&groupid=446

MR https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/417/diffs

### DIFF
--- a/data/autoyast_sle12/create_hdd/create_hdd_sap_sles.xml.ep
+++ b/data/autoyast_sle12/create_hdd/create_hdd_sap_sles.xml.ep
@@ -1,0 +1,296 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <suse_register>
+  <do_registration config:type="boolean">true</do_registration>
+  <reg_code>{{SCC_REGCODE}}</reg_code>
+    <install_updates config:type="boolean">true</install_updates>
+  </suse_register>
+  <bootloader>
+    <device_map config:type="list">
+      <device_map_entry>
+        <firmware>hd0</firmware>
+        <linux>/dev/vda</linux>
+      </device_map_entry>
+    </device_map>
+    <global>
+      <timeout config:type="integer">1</timeout>
+      <activate>true</activate>
+      <timeout config:type="integer">-1</timeout>
+      <boot_boot>false</boot_boot>
+      <xen_append>fadump= crashkernel=201M</xen_append>
+      <xen_kernel_append><![CDATA[vga=gfx-1024x768x16 vga=gfx-1024x768x16 crashkernel=201M<4G]]></xen_kernel_append>
+    </global>
+    <loader_type>grub2</loader_type>
+  </bootloader>
+  <deploy_image>
+    <image_installation config:type="boolean">false</image_installation>
+  </deploy_image>
+  <firewall>
+    <FW_ALLOW_FW_BROADCAST_DMZ>no</FW_ALLOW_FW_BROADCAST_DMZ>
+    <FW_ALLOW_FW_BROADCAST_EXT>no</FW_ALLOW_FW_BROADCAST_EXT>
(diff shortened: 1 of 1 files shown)


schedule/yast/maintenance/autoyast_create_hdd_maintenance_nopatch.yaml
```diff
@@ -0,0 +1,21 @@
+---
+name: autoyast_create_hdd_maintenance_nopatch
```
lemon-suse, line 2: For later migration test, we'd better de-registered system before publish qcow. You can use
schedule/yast/maintenance/autoyast_create_hdd_maintenance_deregister.yaml, or update it for your need.

JRivrain: The system is already not registered. If we omit the "register" section in the profile, then it's not register, maybe we can try it also for other product, and then remove the de-register module everywhere we use autoyast, if that works. I suspect it would not work for sle15 online medium, but works here.

JRivrain: susetest:~ # SUSEConnect --status-text
Installed Products:
------------------------------------------

  SUSE Linux Enterprise Server for SAP Applications 12 SP4
  (SLES_SAP/12.4/x86_64)

  Not Registered

lemon-suse (approved):

schedule/yast/maintenance/autoyast_create_hdd_maintenance_nopatch.yaml
JRivrain: I tried to use first the main create_hdd_ha.xml.ep as we discussed, but that was not working, SAP requires different registration codes and partitioning, so it would require a lot of conditions in the profile.

coolgw, line 2: i suppose we do not need this nopatch since we use autoyast_create_hdd_maintenance_deregister.yaml ?

coolgw:
for VR 12sp4 https://openqa.suse.de/tests/9920694(Still use nopatch schedule yaml), we need new one(with de-register) VR for 12sp4.

JRivrain:
> for VR 12sp4 https://openqa.suse.de/tests/9920694(Still use nopatch schedule yaml), we need new one(with de-register) VR for 12sp4.

VRs updated

coolgw:
Base ticket description, we also need VR for ppc64le.

schedule/yast/maintenance/autoyast_create_hdd_maintenance_nopatch.yaml
JRivrain: In the end, the partitioning and registration could be the same, but it's a different installation medium and requires specific packages to be installed, so I kept the specific profile.

data/autoyast_sle12/create_hdd/create_hdd_sap_sles.xml.ep
```diff
@@ -0,0 +1,296 @@
...
+    <group>100</group>
+    <groups/>
+    <home>/home</home>
+    <inactive>-1</inactive>
+    <no_groups config:type="boolean">true</no_groups>
+    <shell>/bin/bash</shell>
+    <skel>/etc/skel</skel>
+    <umask>022</umask>
+  </user_defaults>
+  <users config:type="list">
```
jknphy, line 276: and the admin?
https://openqa.suse.de/tests/10027979#step/boot_to_desktop/4

jknphy:
Let's merge PR and MR and let's address both points (1) admin and (2) test suite name if needed in different PR/MR.

PR #15846, opened 2022-11-07
Ansible: Do not install python3-yamllint on transactional systems and run trup_call in non-interactive mode

pdostal:
- Related pull request: #15847, #15814
- Verification run: [jeos-extratest](https://pdostal-server.suse.cz/tests/3107), [microos](https://pdostal-server.suse.cz/tests/3109), [alp_default](https://pdostal-server.suse.cz/tests/3108), [mau-extratests2](https://pdostal-server.suse.cz/tests/3105), [extra_tests_textmode](https://pdostal-server.suse.cz/tests/3103)

PR #15814, opened 2022-11-01
Amend ansible tests for ALP

grisu48:
Skip the ansible-test module on ALP since some dependencies are not available.

- Related ticket: https://progress.opensuse.org/issues/119086
- Verification run: [ALP](https://duck-norris.qam.suse.de/tests/11078#step/ansible/218) | [SLES15-SP4](https://duck-norris.qam.suse.de/t11079) | [MicroOS](https://duck-norris.qam.suse.de/t11087) | [Tumbleweed](https://duck-norris.qam.suse.de/t11081)

Note: The journalcheck issue on ALP is unrelated and happens also if [ansible is excluded](https://duck-norris.qam.suse.de/tests/11082#step/journal_check/21)

PR #15847, opened 2022-11-07
Allow non-interactive mode for trup_call commands

pdostal:
## What the pull request does?
 * Do not rely on `root-console` when `transactional::trup_call` command starts with `-n parameter`.
 * Switch to `root-console` in the beginning of `transactional::process_reboot` and back at the end.

## How will it help?
when we work serial terminal, we currently have to switch to `root-console`:
```perl
trup_call 'pkg install ed';
check_reboot_changes;
select_serial_terminal;
```
After this patch we will be able to do just continue in non-interactive mode:
```perl
trup_call '-n pkg install ed';
check_reboot_changes;
```

- Related pull request: #15846, #15814
- Verification run: [MicroOS](https://pdostal-server.suse.cz/tests/2916), [SLE Micro](https://pdostal-server.suse.cz/tests/2917)

### DIFF
--- a/lib/transactional.pm
+++ b/lib/transactional.pm
@@ -74,6 +74,10 @@ sub process_reboot {
     $args{automated_rollback} //= 0;
     $args{expected_grub} //= 1;
 
+    # Switch to root-console as we need VNC to check for grub and for login prompt
+    my $prev_console = current_console();
+    select_console 'root-console', await_console => 0;
+
     handle_first_grub if ($args{automated_rollback});
 
     if (is_microos || is_sle_micro && !is_s390x) {
@@ -102,6 +106,9 @@ sub process_reboot {
         select_console 'root-console';
         assert_script_run 'clear';
     }
+
+    # Switch to the previous console
+    select_console $prev_console;
 }
 
 # Reboot if there's a diff between the current FS and the new snapshot
@@ -165,15 +172,16 @@ sub trup_call {
     $script .= "; echo trup-\$?- | tee -a /dev/$serialdev" unless $cmd =~ /reboot / && $args{exit_code} == 0;
     script_run $script, 0;
     if ($cmd =~ /pkg |ptf /) {
-        if (wait_serial "Continue?") {
+        if ($cmd =~ /(^|\s)-\w*n\w* pkg/) {
+            record_info 'non-interactive', 'The transactional-update command is in non-interactive mode';
+        } elsif (wait_serial "Continue?") {
             send_key "ret";
             # Abort update of broken package
             if ($cmd =~ /\bup(date)?\b/ && $args{exit_code} == 1) {
                 die 'Abort dialog not shown' unless wait_serial('Abort');
                 send_key 'ret';
             }
-        }
-        else {
+        } else {
             die "Confirmation dialog not shown";
         }
     }


lib/transactional.pm
```diff
@@ -172,8 +179,9 @@ sub trup_call {
                 die 'Abort dialog not shown' unless wait_serial('Abort');
                 send_key 'ret';
             }
-        }
-        else {
+        } elsif ($cmd =~ /^-n/) {
```
Vogtinator, line 182: This should come before the `wait_serial` to not waste 30s waiting for something which doesn't happen. Especially with an automatic reboot at the end that might cause openQA to miss the reboot.

mloviska, line 182: This puts some restrictions on the caller as it is now required to use non-interactive as first argument

lib/transactional.pm
```diff
@@ -102,6 +106,9 @@ sub process_reboot {
         select_console 'root-console';
         assert_script_run 'clear';
     }
+
+    # Switch to the previous console
+    ($prev_console) ? select_console($prev_console) : die('The $prev_console is not defined.');
```
mloviska, line 111: Is this necessary? Is there any situation that `current_console` returns a falsy value?

pdostal: It's not. I'll get rid of it.

lib/transactional.pm
pdostal: Good point!

pdostal: From the `transactional-update --help` output:
```bash
# transactional-update --help
...
Syntax: transactional-update [option...] [general-command...] [package-command]
        transactional-update [option...] standalone-command
…
Options:
--interactive, -i          Use interactive mode for package command
--non-interactive, -n      Use non-interactive mode for package command
```

My new regular expression:
```perl
$cmd =~ /-\w*n\w* pkg/
```

Samples I was testing on:
```bash
-n pkg install q # Valid
-q -n pkg install q # Valid
-qn pkg install q # Valid
-nq pkg install q # Valid
reboot -nq pkg install q # Valid
pkg -n install -q # Not valid
-q pkg -n install q # Not valid
```

lib/transactional.pm
```diff
@@ -165,15 +172,16 @@ sub trup_call {
     $script .= "; echo trup-\$?- | tee -a /dev/$serialdev" unless $cmd =~ /reboot / && $args{exit_code} == 0;
     script_run $script, 0;
     if ($cmd =~ /pkg |ptf /) {
-        if (wait_serial "Continue?") {
+        if ($cmd =~ /(^|\s)-\w*n\w* pkg/) {
+            record_info 'non-interactive', 'The transactional-update command is in non-interactive mode';
+        } elsif (wait_serial "Continue?") {
```
foursixnine, line 177: I wish we could add a unit test for this kind of thing by extracting it.

foursixnine (approved):

lib/transactional.pm
pdostal: I don't understand :/

foursixnine: > I don't understand :/

I mean, if we had the condition, in a sort routine, we could easily test the code, like we do for `version_check` for instance but we can think about this in the following weeks (I'm just making a note here)

pdostal:
@mloviska @Vogtinator please re-check. This should be now ready to merge.

mloviska (approved):

PR #15849, opened 2022-11-07
tests/console/sudo: Test the default configuration

Vogtinator:
Test sudo without changing the configuration. The expected behaviour is different on public cloud.

- Related ticket: https://progress.opensuse.org/issues/120019
- Verification runs:
Broken sudo: https://openqa.opensuse.org/tests/2859181#live
Working sudo: https://openqa.opensuse.org/tests/2859138#live

Is my interpretation of public cloud behaviour correct?

### DIFF
--- a/tests/console/sudo.pm
+++ b/tests/console/sudo.pm
@@ -29,11 +29,11 @@ sub sudo_with_pw {
     my $password = $args{password} //= $testapi::password;
     assert_script_run 'sudo -K';
     if ($command =~ /sudo -i|sudo -s|sudo su/) {
-        enter_cmd "expect -c 'spawn $command;expect \"password\";send \"$password\\r\";interact'";
+        enter_cmd "expect -c 'spawn $command;expect \"password\" {send \"$password\\r\";interact'} default {exit 1}";
         sleep 2;
     }
     else {
-        assert_script_run("expect -c '${env}spawn $command;expect \"password\";send \"$password\\r\";interact'$grep", timeout => $args{timeout});
+        assert_script_run("expect -c '${env}spawn $command;expect \"password\" {send \"$password\\r\";interact} default {exit 1}'$grep", timeout => $args{timeout});
     }
 }
 
@@ -47,6 +47,16 @@ sub run {
     my $test_password = 'Sud0_t3st';
     select_console 'root-console';
     zypper_call 'in sudo expect';
+    select_console 'user-console';
+    # Verify that sudo works with the default configuration.
+    if (is_public_cloud) {
+        # No defaults targetpw (public cloud) -> asks for user PW
+        assert_script_run("expect -c 'spawn sudo id -un;expect \"password for $testapi::username\" {send \"$testapi::password\\r\";interact} default {exit 1}' | grep ^root");
+    } else {
+        # Defaults targetpw -> asks for root PW
+        assert_script_run("expect -c 'spawn sudo id -un;expect \"password for root\" {send \"$testapi::password\\r\";interact} default {exit 1}' | grep ^root");
+    }
+    select_console 'root-console';
     # Prepare a file with content '1' for later IO redirection test
     assert_script_run 'echo 1 >/run/openqa_sudo_test';
     # prepare sudoers and test user


tests/console/sudo.pm
```diff
@@ -47,6 +47,16 @@ sub run {
     my $test_password = 'Sud0_t3st';
     select_console 'root-console';
     zypper_call 'in sudo expect';
+    select_console 'user-console';
+    # Verify that sudo works with the default configuration.
+    if (is_public_cloud) {
+        # No defaults targetpw (public cloud) -> asks for user PW
+        assert_script_run("expect -c 'spawn sudo id -un;expect \"password for $testapi::username\" {send \"$testapi::password\\r\";interact} default {exit 1}' | grep ^root");
+    } else {
```
foursixnine, line 55: If I got it right... isn't this the actual condition that will land on TW eventually? cc @DimStar77

Vogtinator: Eventually maybe, but not now and not outside of TW.

DimStar77: > If I got it right... isn't this the actual condition that will land on TW eventually? cc @DimStar77

Indeed, eventually, this might be the case. But for now, we want to protect our users from this implementation that is half-baked and definitively not ready

pdostal (approved):

mloviska:
Seems like this has broken PubCloud jobs https://openqa.suse.de/tests/9902644#step/sudo/7

Vogtinator:
> Seems like this has broken PubCloud jobs https://openqa.suse.de/tests/9902644#step/sudo/7

I'm not surprised, the change for that was only a guess. What's mismatching exactly?

mloviska:
> > Seems like this has broken PubCloud jobs https://openqa.suse.de/tests/9902644#step/sudo/7
> 
> I'm not surprised, the change for that was only a guess. What's mismatching exactly?

I have just tested in locally, https://openqa.suse.de/tests/9902586#step/sudo/7, IMO it should expect "password for root" or not?

Vogtinator:
> > > Seems like this has broken PubCloud jobs https://openqa.suse.de/tests/9902644#step/sudo/7
> > 
> > 
> > I'm not surprised, the change for that was only a guess. What's mismatching exactly?
> 
> I have just tested it locally, https://openqa.suse.de/tests/9902586#step/sudo/7, IMO it should expect "password for root" or not?

According to the code later on, root does not have a password set, so that wouldn't work.

mloviska:
> > I have just tested it locally, https://openqa.suse.de/tests/9902586#step/sudo/7, IMO it should expect "password for root" or not?
> 
> According to the code later on, root does not have a password set, so that wouldn't work.

I think it does have. I have put `sleep` just before the condition and run some manual tests.

screenshot-2022-11-09-14:00:25
screenshot-2022-11-09-14:05:58

Vogtinator:
That is against those lines:

```
    # on publiccloud the root password is not yet set
    # note: due to security reasons, the root password must be reset afterwards
```

Security issue in the images?

mloviska:
> That is against those lines:
> 
> ```
>     # on publiccloud the root password is not yet set
>     # note: due to security reasons, the root password must be reset afterwards
> ```
> 
> Security issue in the images or test code?

Test code: https://openqa.suse.de/tests/9902678#step/prepare_instance/160

Vogtinator:
If the instance is publicly reachable that sounds like an issue indeed.

mloviska:
> If the instance is publicly reachable that sounds like an issue indeed.

Not ideal, but the instance will be destroyed when the test finishes, regardless of the result.

Vogtinator:
I don't think that gets us much further, we need to figure out a fix anyway. Either remove the `is_public_cloud` code in the test because the assumption that the root password is not set is meanwhile wrong, or make sure to not set a root password.

mloviska:
> I don't think that gets us much further, we need to figure out a fix anyway. Either remove the `is_public_cloud` code in the test because the assumption that the root password is not set is meanwhile wrong, or make sure to not set a root password.

Without root password, we won't be able to run console tests as of now.Thus, I vote for removal of the condition.

Vogtinator:
Ok. Can you do a PR? I wouldn't be able to do a VR anyway.

mloviska:
Sure

PR #15856, opened 2022-11-08
Fix regex for health-checker

mloviska:
- failure: [ALP](https://openqa.opensuse.org/tests/2842805#step/journal_check/20)
- Verification run: [alp-0.1-kvm-x86_64](https://openqa.opensuse.org/t2868588)

### DIFF
--- a/tests/console/journal_check.pm
+++ b/tests/console/journal_check.pm
@@ -98,7 +98,7 @@ sub run {
             my $service = $1;
             my $failed_service_output = script_output("systemctl status $service -l || true");
             foreach my $bsc (@matched_bugs) {
-                if ($failed_service_output =~ /$bug_pattern->{$bsc}->{description}/) {
+                if ($failed_service_output =~ $bug_pattern->{$bsc}->{description}) {
                     record_info('Softfail', "Service: $service failed due to $bsc\n$failed_service_output", result => 'softfail');
                     next SRV;
                 }


Vogtinator:
What's actually broken here? For Tumbleweed the text is identical and it matches the regex just fine (https://openqa.opensuse.org/tests/2867851).

mloviska:
> What's actually broken here? For Tumbleweed the text is identical and it matches the regex just fine (https://openqa.opensuse.org/tests/2867851).

It didn't match in ALP unless escaped https://openqa.opensuse.org/tests/2861065#step/journal_check/20. 

Also according to [regex101.com](https://regex101.com/r/YPBmnq/1) it doesn't work :shrug:

mloviska:
> Right, the issue is a conflict with the delimiters:
> 
> ```
> if ($line =~ /$bug_pattern->{$bug}->{description}/)
> ```
> 
> Maybe it works better like this?
> 
> ```
> if ($line =~ $bug_pattern->{$bug}->{description})
> ```

good point, let me try. Thanks

mloviska:
> Right, the issue is a conflict with the delimiters:
> 
> ```
> if ($line =~ /$bug_pattern->{$bug}->{description}/)
> ```
> 
> Maybe it works better like this?
> 
> ```
> if ($line =~ $bug_pattern->{$bug}->{description})
> ```

I guess it works
http://kepler.suse.cz/tests/19536#step/journal_check/20

Vogtinator (approved):

PR #15858, opened 2022-11-08
Temporary disable qesapdeploy lib unit testing

mpagot:
Temporary disable them in order to present and discuss with other involved developers about their usage, purpose and implementation.

https://jira.suse.com/browse/TEAM-6926


alvarocarvajald (changes requested):
I'm against this.

PR #15860, opened 2022-11-08
Install gnu-hpc packages on HPC compute nodes

b10n1k:
Due to the `openmpi` changes I refactored the spack test to use the gnu-hpc packages. As for now the spack test suite uses *mpich* so the test suite is not affected directly. However, with this refactoring, we remove the use of `get_compute_nodes_deps` which is now deprecated.
The only binaries shared with NFS is the directory with the code to run and `prepare_spack_env` takes care of the installation of the required libraries

- Related ticket: https://progress.opensuse.org/issues/119983
- Verification run: https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2315860

### DIFF
--- a/tests/hpc/spack_master.pm
+++ b/tests/hpc/spack_master.pm
@@ -20,23 +20,12 @@ sub run ($self) {
     my $mpi_bin = 'mpi_bin';
     my @cluster_nodes = $self->cluster_names();
     my $cluster_nodes = join(',', @cluster_nodes);
+    my %exports_path = (bin => '/home/bernhard/bin');
+    $self->setup_nfs_server(\%exports_path);
     $self->prepare_spack_env($mpi);
-    my %exports_path = (
-        bin => '/home/bernhard/bin',
-        spack => '/home/bernhard/spack',
-        hpc_lib => '/usr/lib/hpc',
-        spack_lib => '/opt/spack'
-    );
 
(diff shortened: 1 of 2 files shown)


schlad:
preferably the git commits aren't written in a form "I did".
The change is obvious but I don't get why we need this; I can only ready: "Due to the openmpi changes" which does not explain much.

tests/hpc/spack_master.pm
```diff
@@ -20,23 +20,12 @@ sub run ($self) {
     my $mpi_bin = 'mpi_bin';
     my @cluster_nodes = $self->cluster_names();
     my $cluster_nodes = join(',', @cluster_nodes);
+    my %exports_path = (bin => '/home/bernhard/bin');
```
schlad, line 23: can't we just do my $exports_path = '/home/bernhard/bin'; ? 
I'm not sure but I don't see why not :-)

b10n1k: The only reason is because i want to keep the loop and add directory on request. If i change that i will have to touch other parts too. and when/if something else need to be shared it is gonna be back and forth. As it it is, you need only to add/remove paths in the list

schlad (approved):

PR #15866, opened 2022-11-09
x11/network/NM_wpa2_enterprise: Do not click "Connect" in the second time

zhaocongjia:
In the second time we connect to the wifi, there is no need to click "Connect" again -- we already typed password, so the system will auto-connect in the background.

- Related ticket: https://progress.opensuse.org/issues/117031
- Needles: None
- Verification run:
Leap15.4: https://openqa.opensuse.org/tests/2861035#step/NM_wpa2_enterprise/34
TW: https://openqa.opensuse.org/tests/2861036

### DIFF
--- a/tests/x11/network/NM_wpa2_enterprise.pm
+++ b/tests/x11/network/NM_wpa2_enterprise.pm
@@ -28,8 +28,8 @@ sub run {
     enter_cmd "exit";
     enter_cmd "exit";
 
-    # connect again to see if NM has a "connection" after we disabled v4 and v6
-    $self->connect_to_network;
+    # wait for auto reconnect to see if NM has a "connection" after we disabled v4 and v6
+    wait_still_screen;
     assert_screen [qw(network_manager-network_connected network_manager-wrong_card_selected)];
     if (match_has_tag 'network_manager-wrong_card_selected') {
         record_soft_failure 'boo#1079320';


tests/x11/network/NM_wpa2_enterprise.pm
```diff
@@ -36,6 +38,7 @@ sub run {
         assert_and_click 'network_manager-wrong_card_selected';
         assert_screen 'network_manager-network_connected';
     }
+    assert_and_click 'network_manager-popup-cancel';
```
rfan1, line 41: Hello @zhaocongjia 
One minor comment here, can you please add some description here as well?
then everyone can know why we don't need to click connect button any more.

LGTM

rfan1 (approved):

GraceWang571 (commented):
Sorry for failing to quote the code. I mean Line88 of the [NM_wpa2_enterprise.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15866/files#diff-d3861e76837f357f12deb45b710026235a87932772aada987ba38c42c205027f)
This could be simplified. Use `type_password;` would be enough.

I know this is not introduced in this PR, it would be nice if this can be enhanced together with this PR.

tests/x11/network/NM_wpa2_enterprise.pm
GraceWang571: I saw there's no popup window for gnome43 but only a needle with tag "network_manager-popup-cancel" there which seems not real. Can we handle this more elegantly?
https://openqa.opensuse.org/tests/2861036#step/NM_wpa2_enterprise/34

zhaocongjia: I think current way handles both GNOME43 and older GNOME versions.  Otherwise there will be `check_screen`.

GraceWang571: Yeah, I know it works for all GNOME versions. But it looks weird to assert_and_click a "fake" needle.

zhaocongjia: Hi, I have update code logic to remove the need of `network_manager-popup-cancel` -- now we simply `wait_still_screen` to wait for NetworkManager to auto reconnect.

https://openqa.opensuse.org/tests/2865773
https://openqa.opensuse.org/tests/2865779

GraceWang571: Great idea. LGTM.

PR #15868, opened 2022-11-09
wicked: Delay ifup of t08_setup_second_card

cfconrad:
OVS is blocking the port in the beginning sometimes. This cause trouble in this test, as the first DHCP-Request times out and wicked is using the last lease (which is something from the installation).

Now we simply wait 30s to give OVS some time. Most properly this is something related to RSTP configured on the OVS bridge.

Reference: [bsc#1204604](https://bugzilla.suse.com/show_bug.cgi?id=1204604)

- Verification run: https://openqa.opensuse.org/tests/2863400

### DIFF
--- a/tests/wicked/basic/sut/t08_setup_second_card.pm
+++ b/tests/wicked/basic/sut/t08_setup_second_card.pm
@@ -17,6 +17,9 @@ use lockapi;
 
 sub run {
     my ($self, $ctx) = @_;
+
+    sleep(30);    # OVS on a worker is slow sometimes to change and we haven't found better way how to handle it
+
     my $cfg_ifc1 = '/etc/sysconfig/network/ifcfg-' . $ctx->iface();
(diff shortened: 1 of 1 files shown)


ggardet:
It still fails on aarch64: https://openqa.opensuse.org/tests/2863290#step/t08_setup_second_card/12

cfconrad:
I don't see the commit of this PR in the https://openqa.opensuse.org/tests/2863290#investigation tab, maybe this fix was just not deployed yet. I let it run once more with CASEDIR, lets see... https://openqa.opensuse.org/t2864275

ggardet:
> I don't see the commit of this PR in the https://openqa.opensuse.org/tests/2863290#investigation tab, maybe this fix was just not deployed yet. I let it run once more with CASEDIR, lets see... https://openqa.opensuse.org/t2864275

Still failing. Same for x86_64: http://bugzilla.opensuse.org/show_bug.cgi?id=1204604#c3

PR #15869, opened 2022-11-09
transactional/rebootmgr: Adjust to 2.0 and do some cleanup

Vogtinator:
- Related ticket: https://progress.opensuse.org/issues/119482
- Verification run: http://10.168.4.168/tests/1245, http://10.168.4.168/tests/1246


mloviska (approved):
LGTM, it would be nice to check whether `$duration` is `undef` since the default value has been removed and it is up to the caller to remember that it is mandatory from now on

Vogtinator:
> LGTM, it would be nice to check whether `$duration` is `undef` since the default value has been removed and it is up to the caller to remember that it is mandatory from now on

IMO not necessary because then the command is invalid and the cause for that is fairly obvious. There aren't any other callers left.

PR #15871, opened 2022-11-09
Add retry on refused connection for firefox_html5

BillAnastasiadis:
`firefox_html5` test invokes python to get an http server running and then curls that server. Sporadically, curl fails because the server is not up by the time the curl is called.

The curl command used `retry`, but retry does not work with refused connections. It is now replaced with `retry_connrefused` and the delay is set to 1 second. That way it will pass for relatively small delays (if the server comes up after 10 seconds for example) but not bigger ones (if it takes 20 seconds).

That should fix the sporadic issue.

- Related ticket: https://progress.opensuse.org/issues/120061
- Verification run: 
 10 second delay: SUCCESS (http://10.161.225.197/tests/654#step/firefox_html5/12)
 20 second delay: FAILURE (http://10.161.225.197/tests/653#step/firefox_html5/12)


tests/x11/firefox/firefox_html5.pm
```diff
@@ -25,7 +25,7 @@ sub run {
     x11_start_program('xterm');
     script_run('cd ~/data/testwebsites');
     enter_cmd('python3 -m http.server 48080 &');
-    assert_script_run 'curl --connect-timeout 5 --max-time 10 --retry 5 --retry-delay 0 --retry-max-time 40 http://localhost:48080/';
+    assert_script_run 'curl --connect-timeout 5 --max-time 10 --retry-connrefused 5 --retry-delay 1 --retry-max-time 40 http://localhost:48080/';
```
foursixnine, line 28: Can you add a comment, that if it fails on in longer than 20secs, we need to reevaluate how the python webserver is being started?

pdostal: Uh :O I did not know that `curl` implements `retry` internally.

pdostal (approved):

tests/x11/firefox/firefox_html5.pm
rfan1: I have a minor comment, can you please add a port check with `script_retry`?
ss -tnlp | grep 48080

IMO, if the port is up, then curl should work :)

tests/x11/firefox/firefox_html5.pm
```diff
@@ -25,7 +25,7 @@ sub run {
     x11_start_program('xterm');
     script_run('cd ~/data/testwebsites');
     enter_cmd('python3 -m http.server 48080 &');
```
rfan1, line 27: Another comment, how about `background_script_run('python3 -m http.server 48080')` ?

tests/x11/firefox/firefox_html5.pm
BillAnastasiadis: That's a valid point, @rfan1, but apart from @pdostal 's comment, I think that if the server is not up in 10-15 seconds for whatever reason it's a good idea to fail and the reason will be visible in the results. At this point I think we should leave this as it is. If this ends up happening often, we can re-evaluate.

BillAnastasiadis: @foursixnine I added a comment to explain that the server is given 10-15 seconds to come up. Exact time is not available, since this is dependent on the reason of failure and the time it takes for curl to fail, which is variant. Curl here tries to connect the first time for 5 seconds, then retries 5 times, with 1 second between tries, but how long it will take each try to fail is variant and reason-dependent.

It could potentially not fail even in 20 seconds (if curl waited for all the 5 second period each time) but here it won't, because curl fails quickly with "connection refused".

Now that I think about it, I will remove all exact time numbers from the comment.

foursixnine (approved):

rfan1 (approved):

pdostal:
@foursixnine I think that this was not ready yet.

PR #15873, opened 2022-11-09
Increase timeout for booting to local disk

ge0r:
The default timeout occasionaly causes a stall on test runs, because the display manager screen takes longer to appear. Increasing the timeout mitigates this.

- Related ticket: https://progress.opensuse.org/issues/118336, related to [this failure](https://openqa.suse.de/tests/9889455#step/snapper_rollback/22)
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/9907879#step/snapper_rollback/24 (passed 3 consecutive times)


rakoenig (approved):
LGTM

PR #15876, opened 2022-11-09
PubCloud: root password set by tests

mloviska:
Follow up for PR#15849
Password for root is configured in `publiccloud/prepare_instance.pm` in order to run console tests.

- ticket: [test fails in sudo](https://progress.opensuse.org/issues/119860)
- Verification run: 
  * [sle-15-SP1-AZURE-BYOS-Updates-x86_64-Build20221108-1-publiccloud_consoletests@64bit](http://kepler.suse.cz/tests/19507#step/sudo/5)
  * [sle-15-SP4-AZURE-BYOS-Updates-x86_64-Build20221108-1-publiccloud_consoletests@64bit](http://kepler.suse.cz/tests/19508#live)

### DIFF
--- a/tests/console/sudo.pm
+++ b/tests/console/sudo.pm
@@ -19,7 +19,7 @@ use strict;
 use warnings;
 use testapi;
 use utils 'zypper_call';
-use version_utils qw(is_sle is_public_cloud);
+use version_utils qw(is_sle);
 
 sub sudo_with_pw {
     my ($command, %args) = @_;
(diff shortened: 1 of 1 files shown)


mloviska:
> If this is the right fix, then the following lines should be removed too:
> 
> ```
>     # on publiccloud the root password is not yet set
>     # note: due to security reasons, the root password must be reset afterwards
>     my $password = $testapi::password;
>     assert_script_run("echo -e '$password\n$password' | passwd root") if is_public_cloud;
> ```
> 
> ```
>     # remove root password on publiccloud again
>     assert_script_run("passwd root --lock") if is_public_cloud;
> ```

Done

asmorodskyi:
does it mean that from now on `root` user will have password on PubCloud images ?

Vogtinator:
> does it mean that from now on `root` user will have password on PubCloud images ?

It already had one apparently

mloviska:
> does it mean that from now on `root` user will have password on PubCloud images ?

Isn't it done in order for console tests?
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/prepare_instance.pm#L33

As I understand it, image proof verifies, that root password is unset by default.

asmorodskyi (approved):
LGTM

PR #15881, opened 2022-11-09
Trento test, HANA primary stop

mpagot:
Add new test module to stop Primary HANA and check how Trento UI detect it.
Main OS user, in the deployed cluster, is controlled by the test Fix in 'Trento deployment from git' about branches not available Cypress to support Agent version expectation
Better teardown for clusters, in case of failure.

- Verification run:


tests/sles4sap/trento/cluster_deploy.pm
```diff
@@ -28,6 +28,11 @@ sub run {
         '-a', get_vnet($cluster_rg));
     record_info('NET PEERING');
     assert_script_run($cmd, 360);
+
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $inventory = qesap_get_inventory($prov);
+
+    qesap_ansible_cmd(cmd => 'crm status', provider => $prov, filter => $_), for ('vmhana01', 'vmhana02');
```
alvarocarvajald, line 35: There's an extra `,` there just before the `for`

alvarocarvajald (approved):
LGTM.

tests/sles4sap/trento/cluster_deploy.pm
mpagot: is it also **crm status** supposed to have the same output an all node?

lpalovsky (approved):
LGTM

tests/sles4sap/trento/cluster_deploy.pm
alvarocarvajald: Yes, the output should be the same in all cluster nodes.

jankohoutek (commented):
LGTM

PR #15882, opened 2022-11-10
Change regression powerKVM cases to use powerVM

tinawang123:
Change regression powerKVM cases to use powerVM, and change prepare job to use autoyast

- Related ticket: https://progress.opensuse.org/issues/116962
- Verification run: https://openqa.suse.de/tests/10048752
   https://openqa.suse.de/tests/10048911#
   https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=50.1&groupid=466
   All others powerVM case VR
- Related MR: https://gitlab.suse.de/coolgw/wegao-test/-/merge_requests/186


lib/service_check.pm
```diff
@@ -337,7 +337,7 @@ sub check_services {
         my $srv_proc_name = $service->{$s}->{srv_proc_name};
         my $support_ver = $service->{$s}->{support_ver};
         my $service_type = 'Systemd';
-        next unless (($service->{$s}->{before_migration} eq 'PASS') && _is_applicable($srv_pkg_name));
+        next unless _is_applicable($srv_pkg_name);
```
lemon-suse, line 340: I suggestion use a switch here for chained jobs.

schedule/migration/offline_spvm_Upgrade.yaml
```diff
@@ -20,9 +20,53 @@ schedule:
   - installation/handle_reboot
   - installation/first_boot
   - migration/post_upgrade
-  - console/system_prepare
-  - console/consoletest_setup
-  - console/zypper_lr
-  - console/check_system_info
-  - boot/grub_test_snapshot
-  - boot/snapper_rollback
+  - '{{qcow_generation}}'
```
lemon-suse, line 23: Just don't know that spvm already support qcow? Or just used for judgement of regression test? If so it is better to use other setting, such as REGRESSION_TEST.

data/autoyast_sle15/autoyast_sle_powervm_pre.xml.ep
```diff
@@ -0,0 +1,181 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jknphy, line 1: let's remove this `_pre` from the name, in theory is a auto-installation than can be used for anything.

schedule/migration/autoyast_spvm_regression_test_offline_pre.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
jknphy, line 1: is not spvm, we are using hmc for PowerVM.
Same comment, seems a generic installation. we using in YaST group @pvm at the end, but you could use for example: `autoyast_sle_spvm.yaml` or something like that, `offline` is only related with the linked test suite.

the other schedule below for auto-installation is exactly the same, we should just have one, and then directly chain the test suite that does rest of changes.

schedule/migration/online_spvm_Upgrade.yaml
```diff
@@ -0,0 +1,81 @@
+name:   online_spvm_Upgrade
```
jknphy, line 1: I would prefer to have all in lower letter, I have seen many issues with this in the past:
s/Upgrade/upgrade/ but if breaks current nomenclature, leave it like that.

lib/service_check.pm
coolgw: Add comments why is_spvm need check.

lemon-suse: I agree with coolgw, please add some comments for the reason to add powervm as an exception.

lemon-suse: Please run a VR not on powerVM to prove this change won't affect other test.

schedule/migration/offline_spvm_Upgrade.yaml
jknphy: we shouldn't use spvm backend (only some Kernel tests use it and we have limited resources), we should use the hmc one (that is what we refer with PowerVM), which doesn't support snapshots.

tinawang123: I have changed file name to autoyast_regression_test_offline_pre@pvm.yaml  and autoyast_regression_test_online_pre@pvm.yaml
Please check again.

lemon-suse (approved):

jknphy:
could you please point in first comment to the rest of VR and MR?

PR #15883, opened 2022-11-10
Set hostname to 'localhost' if hostname boot option is not used

rfan1:
https://progress.opensuse.org/issues/117424

If hostname boot option is not used:
linuxrc doesn't write /etc/install.inf::Hostname, so it is not present. linuxrc writes /etc/install.inf::SetHostnameUsed=0. linuxrc sets the transient hostname to install.

- Related ticket: https://progress.opensuse.org/issues/117424
- Verification run: https://openqa.suse.de/tests/9916003


pdostal (changes requested):
The commit message (and pull request description) are misleading as this code does not set the hostname but only checks it.

rfan1:
> The commit message (and pull request description) are misleading as this code does not set the hostname but only checks it.

Thanks @pdostal , any suggestion?

pdostal:
Instead of:
> Set hostname to 'localhost' if hostname boot option is not used

You can say:
> Check that hostname is 'localhost' if hostname boot option is not used on SLE15-SP2+

rfan1:
> Instead of:
> 
> > Set hostname to 'localhost' if hostname boot option is not used
> 
> You can say:
> 
> > Check that hostname is 'localhost' if hostname boot option is not used on SLE15-SP2+

Fixed

pdostal (approved):

rfan1:
@pdostal @foursixnine 
Could you please help merge it?

PR #15885, opened 2022-11-10
installation/boot_encrypt: New grub2 is in more of Tumbleweed now

Vogtinator:
Ugly :-/

- Verification run: https://openqa.opensuse.org/t2866162


DimStar77 (approved):
I hope we got all the exceptions right; as timeprogresses, the various is_* will receive version constraints, but otherwise stay. So probably the best we can do

PR #15887, opened 2022-11-10
Update mpirun parameters to prevent warnings when in Ethernet

b10n1k:
With recent updates the `openmpi` throws warnings when mpirun runs. That should occur when the Ethernet is used instead of the Infiniband. AFAIK the detection of one or the other is not easy task and even enabling a whole bunch of log it's not easy to be sure what is really being used

- Related ticket: https://progress.opensuse.org/issues/119758
- Verification run: http://aquarius.suse.cz/tests/14047#step/mpi_master/120 vs https://openqa.suse.de/tests/9916014#step/mpi_master/120

VR with second commit:
- http://aquarius.suse.cz/tests/14280#step/mpi_master/134
- http://aquarius.suse.cz/tests/14271#step/mpi_master/3
- http://aquarius.suse.cz/tests/14284

### DIFF
--- a/lib/hpc/formatter.pm
+++ b/lib/hpc/formatter.pm
@@ -10,6 +10,7 @@ package hpc::formatter;
 use Mojo::Base 'hpcbase', -signatures;
 use hpc::utils;
 use testapi qw(get_var get_required_var);
+use version_utils qw(is_sle);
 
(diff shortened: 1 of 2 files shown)


Avinesh (approved):
Looks fine!

PR #15889, opened 2022-11-10
Apply workaround to force screen refresh on YaST QT window

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/117949
- All workarounds for screen refresh problems that used alt-f10 are now replaced with the shift-f3 workaround.
- The workaround code is outsourced into a library module and called with the needle(s) that we are looking for. The workaround is only applied when necessary (needle would not match) and then accompanied by a soft failure. So soft failures are only recorded when the screen refresh problem occurs.
- Version matiching is now is_sle(''>=15-SP4') because problem seems to occur on SLE 15-SP5 too.

- Verification run: https://openqa.suse.de/tests/10032585 (remaining errors are related to firewall problems)

### DIFF
--- a/lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
+++ b/lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
@@ -12,6 +12,8 @@ use strict;
 use warnings;
 use testapi;
 use parent 'YaST::NetworkSettings::NetworkCardSetup::NetworkCardSetupWizard';
+use YaST::workarounds;
+use version_utils qw(is_sle);
 
 use constant {
     NETWORK_CARD_SETUP => 'yast2_lan_network_card_setup',
@@ -35,6 +37,7 @@ sub select_tab {
 
 sub select_bond_slave_in_list {
     assert_screen(BOND_SLAVES_TAB);
+    apply_workaround_bsc1204176(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED) if (is_sle('>=15-SP4'));
     assert_and_click(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED);
 }
 
--- a/lib/YaST/NetworkSettings/OverviewTab.pm
+++ b/lib/YaST/NetworkSettings/OverviewTab.pm
@@ -11,6 +11,8 @@ package YaST::NetworkSettings::OverviewTab;
 use strict;
 use warnings;
 use testapi;
+use YaST::workarounds;
+use version_utils qw(is_sle);
 
 use constant {
     OVERVIEW_TAB => 'yast2_lan_overview_tab_selected',
@@ -27,16 +29,19 @@ sub new {
 }
 
 sub press_add {
+    apply_workaround_bsc1204176(OVERVIEW_TAB) if (is_sle('>=15-SP4'));
     assert_screen(OVERVIEW_TAB);
     send_key('alt-a');
 }
 
 sub press_edit {
+    apply_workaround_bsc1204176(OVERVIEW_TAB) if (is_sle('>=15-SP4'));
     assert_screen(OVERVIEW_TAB);
     send_key('alt-i');
 }
 
 sub press_delete {
+    apply_workaround_bsc1204176(OVERVIEW_TAB) if (is_sle('>=15-SP4'));
     assert_screen(OVERVIEW_TAB);
     send_key('alt-t');
 }
@@ -65,9 +70,8 @@ sub select_device {
 }
 
 sub press_ok {
-    assert_screen(OVERVIEW_TAB);
+    apply_workaround_bsc1204176(OVERVIEW_TAB) if (is_sle('>=15-SP4'));
     send_key('alt-o');
 }
-
 
 1;
--- a/lib/YaST/workarounds.pm
+++ b/lib/YaST/workarounds.pm
@@ -1,0 +1,51 @@
+# Copyright 2015-2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package YaST::workarounds;
+
+
+use strict;
+use warnings;
+use Exporter 'import';
+use testapi;
+use version_utils;
+
+
+our @EXPORT = qw(apply_workaround_bsc1204176);
+
+=head1 Workarounds for known issues
+
+=head2 apply_workaround_bsc1204176 ($mustmatch, [,[$timeout] | timeout => $timeout] ):
+
+Workaround for the screen refresh issue.
+
+First checks if we need to apply the workaround, if the needle matches 
+already no workaround is required.
+
+If the workaround is needed we record a soft failure and apply a 'shift-f3' 
+and 'esc' sequence that should fix the problem
+
+If the problem still persists (we saw one occasion in the VRs) then we retry
+with maximazing and shrinking the screen twice by sending 'alt-f10' two times. 
+
+=cut
+
+sub apply_workaround_bsc1204176 {
+    my ($mustmatch) = shift;
+    my $timeout;
+    $timeout = shift if (@_ % 2);
+    my %args = (timeout => $timeout // 0, @_);
+    if (!check_screen($mustmatch, %args)) {
+        record_soft_failure('bsc#1204176 - Resizing window as workaround for YaST content not loading');
+        send_key('shift-f3', wait_screen_change => 1);
+        send_key('esc', wait_screen_change => 1);
+        # in some verification tests this didn't work, so let's check
+        if (!check_screen($mustmatch)) {
+            record_info('Retry', "shift-f3 workaround did not solve the problem");
+            send_key('alt-f10', wait_screen_change => 1);
+            send_key('alt-f10', wait_screen_change => 1);
+        }
+    }
+}
+
+1;
--- a/tests/yast2_gui/yast2_security.pm
+++ b/tests/yast2_gui/yast2_security.pm
@@ -22,6 +22,7 @@ use strict;
 use warnings;
 use testapi;
 use version_utils qw(is_sle);
+use YaST::workarounds;
 
 sub run {
     select_console "x11";
@@ -40,10 +41,7 @@ sub run {
     # Check previously set values + Login Settings
     y2_module_guitest::launch_yast2_module_x11("security", match_timeout => 120);
     assert_and_click "yast2_security-pwd-settings";
-    if (is_sle('=15-SP4')) {
-        record_soft_failure('bsc#1204176 - Resizing window as workaround for YaST content not loading');
-        send_key_until_needlematch('yast2_security-check-min-pwd-len-and-exp-days', 'alt-f10', 10, 2);
-    }
+    apply_workaround_bsc1204176('yast2_security-check-min-pwd-len-and-exp-days') if (is_sle('>=15-SP4'));
     assert_screen "yast2_security-check-min-pwd-len-and-exp-days";
     assert_and_click "yast2_security-login-settings";
     send_key "alt-d";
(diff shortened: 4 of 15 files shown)


lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,27 @@
+# Copyright 2015-2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package YaST::workarounds;
+
+
```
jknphy, line 6: extra break line

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,27 @@
...
+use strict;
+use warnings;
+use Exporter;
+use Mojo::Base 'Exporter', -signatures;
+use testapi;
+use version_utils;
+
+
```
jknphy, line 14: extra break line.

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,27 @@
...
+our @EXPORT = qw(workaround_screen_refresh);
+
+sub workaround_screen_refresh {
+    my ($needle) = shift;
+    if (!check_screen($needle, 180) && (is_sle('>=15-SP4'))) {
```
jknphy, line 19: condition should go outside, otherwise you cannot give a right name to the method and less you call it `check_and_apply_blablab`, it is more clear for read to see when something is applied from the call.
besides you should use the complex check on the right of the && operation to save time.

jknphy: normally workaround have in the name of the method the one that they apply, otherwise you need to dive into the code to figure out: `apply_workaround_bsc1204176`. There might be different refreshment issues so doesn't worth to include it in the name.

jknphy: check screen with 180? even in the documentation states that that is not recommended.
could you please comment the algorithm of the workaround, reading the code I don't get why we do it, perhaps in the bug or trying manually would make sense, but it is not clear from the code.

jknphy (commented):
first round of review.

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,27 @@
...
+        record_soft_failure('bsc#1204176 - Resizing window as workaround for YaST content not loading');
+        send_key('shift-f3', wait_screen_change => 1);
+        send_key('esc');
+        send_key_until_needlematch($needle, 'esc', 10, 2);
```
jknphy, line 23: with a wait_screen_change should do, it is better to avoid send the same key so many times, we might find ourselves in a complete different screen...

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,27 @@
+# Copyright 2015-2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package YaST::workarounds;
+
+
+use strict;
+use warnings;
+use Exporter;
+use Mojo::Base 'Exporter', -signatures;
```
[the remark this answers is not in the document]
rakoenig: In theory I totally agree with you. I remember trying out some settings with the Exporter, but nothing worked. But ok, let me give it a try, I guess that `use Exporter 'import'` I didn't try so far.

lib/YaST/workarounds.pm
rakoenig: The moment I did that lib I was replacing an assert screen that waited for 180. But maybe a better approach is to submit a timeout parameter that defaults to something much less and in the cases where it will take longer I can set up the timeout explicitly.

lib/YaST/workarounds.pm
rakoenig: The extra `send_key('esc')` is there for a reason: I figured out that "send_key_until_needlematch" does not send a key at all if the needle matches already. So its more a "while_not_needlematch_send_key" function than what the name suggests. Nevertheless, I will try it out. The VRs still show that sometimes the workaround is applied, but does not work.

jknphy: I think you don't need that command, just  wrap it as usual: `wait_screen_change { send_key 'esc'};`

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,52 @@
+# Copyright 2015-2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package YaST::workarounds;
+
+
+use strict;
+use warnings;
+use Exporter 'import';
+# use Mojo::Base 'Exporter', -signatures;
```
jknphy, line 10: let's drop this line then.

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,52 @@
...
+sub apply_workaround_bsc1204176 {
+    my ($needle) = shift;
+    my $timeout;
+    $timeout = shift if (@_ % 2);
+    my %args = (timeout => $timeout // 0, @_);
+    if ((is_sle('>=15-SP4') && !check_screen($needle, %args))) {
+        record_soft_failure('bsc#1204176 - Resizing window as workaround for YaST content not loading');
+        send_key('shift-f3', wait_screen_change => 1);
+        send_key('esc', wait_screen_change => 1);
+        # in some verifcation tests this didn't work, so let's check
```
jknphy, line 43: s/verifcation/verification/

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,52 @@
...
+
+If the problem still persists (we saw one occasion in the VRs) then we retry
+with maximazing and shrinking the screen twice by sending 'alt-f10' two times. 
+
+=cut
+
+sub apply_workaround_bsc1204176 {
+    my ($needle) = shift;
+    my $timeout;
+    $timeout = shift if (@_ % 2);
```
jknphy, line 36: one line: `$my timeout = ...`

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,51 @@
...
```
jknphy, line 36: never will be 2, we only have 1 optional parameter.

lib/YaST/NetworkSettings/OverviewTab.pm
```diff
@@ -65,9 +69,8 @@ sub select_device {
 }
 
 sub press_ok {
-    assert_screen(OVERVIEW_TAB);
+    apply_workaround_bsc1204176(OVERVIEW_TAB) if (is_sle('>=15-SP4'));
```
jknphy, line 72: might be missing tidy? two spaces here...strange.

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,50 @@
...
+and 'esc' sequence that should fix the problem
+
+If the problem still persists (we saw one occasion in the VRs) then we retry
+with maximazing and shrinking the screen twice by sending 'alt-f10' two times. 
+
+=cut
+
+sub apply_workaround_bsc1204176 {
+    my ($needle) = shift;
+    my $timeout = shift if (@_ % 1);
```
jknphy, line 35: this is always 0, it is the reminder of a division by 1.

lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
```diff
@@ -35,6 +37,7 @@ sub select_tab {
 
 sub select_bond_slave_in_list {
     assert_screen(BOND_SLAVES_TAB);
+    apply_workaround_bsc1204176(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED) if (is_sle('>=15-SP4'));
```
jknphy, line 40: (optional) you don't need the surrounding parenthesis, just:
`is_sle('>=15-SP4')`

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,51 @@
...
+
+If the workaround is needed we record a soft failure and apply a 'shift-f3' 
+and 'esc' sequence that should fix the problem
+
+If the problem still persists (we saw one occasion in the VRs) then we retry
+with maximazing and shrinking the screen twice by sending 'alt-f10' two times. 
+
+=cut
+
+sub apply_workaround_bsc1204176 {
```
jknphy, line 33: The code from check_screen can be reused, it takes into account 1 or more optional parameters, this is why I think it happens:

```
fn([qw(needle1 needle2)], timeout => 5); # call with named param

my ($needle) = shift; #unpack needle or array ref of needles
my $timeout = shift if (@_ % 2); # @_ contains a list `qw(timeout 5)` so module operation is `0` and `$timeout` is not initialized
my %args = (timeout => $timeout // 0, @_); # %args is (timeout => 0, timeout => 5) where the second key/value overrides the first

fn('needle',5); # call with positional param

my ($needle) = shift; #unpack needle or array ref of needles
my $timeout = shift if (@_ % 2); # the remainder is 1 so `$timeout` is set and `@_` is empty.
my %args = (timeout => $timeout // 0, @_); # $timeout is set and we got %args = (timeout => 5, ()) and the empty list doesn't append anything as it is not a reference.
```

Conclusion: Perl is full of magic!

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,51 @@
...
+use Exporter 'import';
+use testapi;
+use version_utils;
+
+
+our @EXPORT = qw(apply_workaround_bsc1204176);
+
+=head1 Workarounds for known issues
+
+=head2 apply_workaround_bsc1204176 ($needle, [,[$timeout] | timeout => $timeout] ):
```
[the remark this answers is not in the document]
rakoenig: fixed

lib/YaST/workarounds.pm
```diff
@@ -0,0 +1,51 @@
...
+
+If the problem still persists (we saw one occasion in the VRs) then we retry
+with maximazing and shrinking the screen twice by sending 'alt-f10' two times. 
+
+=cut
+
+sub apply_workaround_bsc1204176 {
+    my ($mustmatch) = shift;
+    my $timeout;
+    $timeout = shift if (@_);
```
jknphy, line 36:
```suggestion
    $timeout = shift if (@_ % 2);
```

jknphy: otherwise we might be appending parameters that are not pairs, failing for an odd number.

DimStar77:
This seems to have resulted in https://openqa.opensuse.org/tests/2919731#step/yast2_security/28

Looking at the investigation jobs on that test, we see:
* New snapshot, New test => fail
* New snapshot, Old testtest => success (unlikely to be a product issue)
* Old snapshot, New test => fail
* Old snapshot, Old test => successful

tests/yast2_gui/yast2_security.pm
```diff
@@ -53,11 +51,8 @@ sub run {
     wait_still_screen 3;
 
     # Check previously set values + Miscellaneous Settings
-    y2_module_guitest::launch_yast2_module_x11("security", match_timeout => 120);
-    if (is_sle('=15-SP4')) {
-        record_soft_failure('bsc#1204176 - Resizing window as workaround for YaST content not loading');
-        send_key_until_needlematch('yast2_security-login-settings', 'alt-f10', 10, 2);
-    }
+    y2_module_guitest::launch_yast2_module_x11("security", match_timeout => 120) if (is_sle('>=15-SP4'));
+    apply_workaround_bsc1204176('yast2_security-login-settings');
```
jknphy, line 55: yep, we missed this `if`.

DimStar77: you mean the `if` should have been on the `apply_workaround` line, not the `y2_module_guitest::…`, right?

rakoenig: Yes, the if should be trigger apply_workaround_bsc1204176 only for SLE '>=15-SP4'. 
Made a fix, VR is currently running. PR will be made when VR is successful.

rakoenig: VR failed. The problem was that we didn't "miss" the check, it was accidentially at the end of the line above, so calling yast2_module_x11 never happened on Tumbleweed. 
Fixed that, restarting VR.

DimStar77: which is what I wrote above - but maybe not clear enough. sorry for that

PR #15891, opened 2022-11-10
Pass the ARCH variable to the BCI test environment

ilausuch:
We need to pass the ARCH to avoid the execution of the BCI test of cassandra in case of ppc

- Related ticket: https://progress.opensuse.org/issues/119344
- Verification run: http://openqa.suse.de/tests/9921160

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -86,6 +86,7 @@ sub run {
     my $bci_devel_repo = get_var('BCI_DEVEL_REPO');
     my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
     my $version = get_required_var('VERSION');
+    my $arch = get_required_var('ARCH');
     my $test_envs = get_required_var('BCI_TEST_ENVS');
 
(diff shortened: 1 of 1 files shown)


m-dati (approved):

mloviska:
I guess it might be straightforward, but yet I have got a silly question, sorry. How exporting the variable to `env` will prevent scheduling of these tests? Or there is some follow up PR connected to this one ?

grisu48 (changes requested):
Wait, we don't need this. More text incoming

ilausuch:
This PR is not needed. Thanks @grisu48

mloviska:
> Wait, we don't need this.
> 
> You can probably use `auto_container.connection.system_info.arch` in the BCI tests itself, as e.g. in https://github.com/SUSE/BCI-tests/blob/main/tests/test_init.py#L42

Just to clarify the situation, doesn't it imply that the openQA test will be still triggered, but nothing will be executed by `bci_test`? Or it will skip just `tests_openjdk.*` tests?

PR #15897, opened 2022-11-11
Add path to utilities located in /usr/sbin

mloviska:
As [ [Build 20221110-1] Defaults secure_path is commented in /etc/sudoers for images v20221108 and newer ](https://bugzilla.suse.com/show_bug.cgi?id=1205325) has hit all new PC images and we cannot block the updates, I would propose to call some of our utils with absolute path.

VRs:

* [sle-15-SP1-EC2-BYOS-Updates-x86_64-Build20221110-1-publiccloud_consoletests@64bit ](http://kepler.suse.cz/tests/19529#)
* [sle-15-SP4-GCE-BYOS-Updates](http://kepler.suse.cz/tests/19531#)
* [sle-15-SP2-EC2-BYOS-Updates-x86_64](http://kepler.suse.cz/tests/19533)

### DIFF
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -92,6 +92,7 @@ sub registercloudguest {
     my ($instance) = @_;
     my $regcode = get_required_var('SCC_REGCODE');
     my $remote = $instance->username . '@' . $instance->public_ip;
+    my $path = is_sle('>15') && is_sle('<15-SP3') ? '/usr/sbin/' : '';
     # not all images currently have registercloudguest pre-installed .
     # in such a case,we need to regsiter against SCC and install registercloudguest with all needed dependencies and then
     # unregister and re-register with registercloudguest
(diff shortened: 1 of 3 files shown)


mloviska:
The other option would be to uncomment `Defaults secure_path` in `/etc/sudoers`

grisu48 (commented):
While the workaround looks good to me, I wonder if we actually SHOULD do this. IMHO this is an issue with the `/etc/sudoers` and we should not correct for it, unless explicitly stated. And in such case we probably also should only do it for the affected versions.

grisu48 (approved):
LGTM if the devs approve this workaround.

asmorodskyi (approved):
LGTM

mloviska:
> And in such case we probably also should only do it for the affected versions.
If other images are to be refreshed will hit it everywhere I suppose. The part with uncommenting would solve the problem as it would be a `noop` in case the line is already uncommented

ilausuch (approved):

PR #15900, opened 2022-11-14
Extend installation of chrony to 15-SP5

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/120234
- Verification run: http://openqa.suse.de/tests/9952683

### DIFF
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -70,7 +70,7 @@ sub run {
             gcloud_install();
 
             # package needed by init():
-            my $pkg = is_sle('=15-SP4') ? "in chrony" : "in ntp";
+            my $pkg = is_sle('>15-SP4') ? "in chrony" : "in ntp";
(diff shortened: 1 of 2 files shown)


pdostal (approved):
Cool

clanig:
Hi Ivan, thank you for the update.
Could you please check if chrony should still be installed for SP4? As far as I see, that code would not just include SP5, but exclude SP4?

tests/containers/helm.pm
```diff
@@ -70,7 +70,7 @@ sub run {
             gcloud_install();
 
             # package needed by init():
-            my $pkg = is_sle('=15-SP4') ? "in chrony" : "in ntp";
+            my $pkg = is_sle('>15-SP4') ? "in chrony" : "in ntp";
```
mloviska, line 73: hm...`chrony` should be used on all sle15 systems

PR #15904, opened 2022-11-14
publiccloud : drop flags which does not make sense

asmorodskyi:
ssh_interactive_end is designed to run at the end of test suite
currently only thing which module is doing is cleanup which makes
use of fatal flag unappropriate, publiccloud_multi_module is designed
to allow other modules to run AFTER this one, but as was stated above
it is not the case for ssh_interactive_end


mloviska (approved):
Makes sense to me!

asmorodskyi:
TUNNELED: 0 - http://autobot.qa.suse.de/tests/3495

PR #15915, opened 2022-11-15
Add ALP prepare run

grisu48:
Adds a preparation step for the ALP. Currently it only sets the staging image repository.

- Related ticket: https://progress.opensuse.org/issues/119086
- Verification run: https://duck-norris.qam.suse.de/tests/11109 (failure in journalctl expected)

### DIFF
--- a/tests/transactional/host_config.pm
+++ b/tests/transactional/host_config.pm
@@ -30,6 +30,8 @@ sub run {
     if (is_alp) {
         change_grub_config('=.*', '=1024x768', 'GRUB_GFXMODE=');
         zypper_call('mr -e ALP-Build');
+
+        add_staging_repos() if (get_var("STAGING"));
     }
 
     if (!$keep_grub_timeout or $extrabootparams) {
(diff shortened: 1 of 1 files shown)


mloviska:
I think it would be better to hook it into host_config, so the host configuration is one place rather than scattered around. Other than that, LGTM

tests/transactional/host_config.pm
```diff
@@ -30,6 +30,8 @@ sub run {
     if (is_alp) {
         change_grub_config('=.*', '=1024x768', 'GRUB_GFXMODE=');
         zypper_call('mr -e ALP-Build');
+
+        add_staging_repos() if (get_var("STAGING"));
```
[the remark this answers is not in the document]
grisu48: See line 30

tests/transactional/host_config.pm
```diff
@@ -40,6 +42,17 @@ sub run {
     }
 }
 
+
+sub add_staging_repos {
+    my $stage = get_required_var("STAGING");
+
+    # Add staging repository with increased priority over default ones
+    my $repo = "https://download.opensuse.org/repositories/SUSE:/ALP:/Staging:/$stage/images/repo/ALP-0.1-x86_64-Media1/";
```
pdostal, line 50: Does that mean that there are 'normal' repositories and additional 'staging repository'?

I mean, I had a problem in the past that there was no new build and the old build's repositories were gone which blocked me.

Will that be stabilised?

pdostal (approved):

tests/transactional/host_config.pm
grisu48: I don't know about the problems in the past, but this repository was requested to be added in https://progress.opensuse.org/issues/119086#note-16. This is the expected flow for ALP staging as of now.

fcrozat: @pdostal ATM, ALP doesn't do registration, so no proxy-scc mecanism can't be used. We hardcoded the "PUBLISH" repository (ie milestone repository) as default repository in the ALP image to ensure people didn't use the development repository by mistake. But for Staging, we need to use the associated staging repository instead of the PUBLISH one.

tests/transactional/host_config.pm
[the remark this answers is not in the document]
mloviska: ah, missed that. thanks!

PR #15920, opened 2022-11-16
Install opens-iscsi package in iscsi_client

ge0r:
Package open-iscsi is no longer a dependency for yast2-iscsi-client. This results in a popup, when the yast2 iscsi client  module starts that asks for open-iscsi to be installed. This is resolved by just installing open-iscsi before launching the yast2 module.

- Related ticket: https://progress.opensuse.org/issues/120013
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/9973939# (Fails on a new spot now)


rakoenig (approved):
LGTM

PR #15932, opened 2022-11-17
Ansible remote command with returned output

mpagot:
It is sort of a script_output version for Ansible commands

- Verification run: https://openqaworker15.qa.suse.cz/tests/68519 In particular the string for this box https://openqaworker15.qa.suse.cz/tests/68519#step/test_cluster/33 is obtained as output of this new function

https://openqaworker15.qa.suse.cz/tests/68541

### DIFF
--- a/data/sles4sap/script_output.yaml
+++ b/data/sles4sap/script_output.yaml
@@ -1,0 +1,15 @@
+---
+- hosts: all
+  vars:
+    cmd: ''
+    out_path: '/tmp/ansible_script_output/'
+    out_file: 'testout.txt'
+  tasks:
+  - name: 'Run cmd'
+    shell: "{{ cmd }} > /tmp/{{ out_file }}"
+    register: crm_status_files
+  - name: fetch
+    fetch:
+      flat: yes
+      src: "/tmp/{{ out_file }}"
+      dest: "{{ out_path }}"
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -57,6 +57,7 @@ our @EXPORT = qw(
   qesap_execute
   qesap_yaml_replace
   qesap_ansible_cmd
+  qesap_ansible_script_output
   qesap_create_ansible_section
 );
 
@@ -102,7 +103,7 @@ sub qesap_get_variables {
     my $yaml_file = $paths{'qesap_conf_src'};
     my %variables;
     my $grep_cmd = "grep -v '#' | grep -oE %[A-Z0-9_]*% | sed s/%//g";
-    my $cmd = join(" ", "curl -s -fL", $yaml_file, "|", $grep_cmd);
+    my $cmd = join(' ', 'curl -s -fL', $yaml_file, '|', $grep_cmd);
 
     for my $variable (split(" ", script_output($cmd))) {
         $variables{$variable} = get_required_var($variable);
@@ -158,7 +159,7 @@ sub qesap_pip_install {
 
     push(@log_files, $pip_install_log);
     record_info("QESAP repo", "Installing pip requirements");
-    assert_script_run(join(" ", $pip_ints_cmd, '-r', $paths{deployment_dir} . '/requirements.txt | tee -a', $pip_install_log), 360);
+    assert_script_run(join(' ', $pip_ints_cmd, '-r', $paths{deployment_dir} . '/requirements.txt | tee -a', $pip_install_log), 360);
 }
 
 =head3 qesap_upload_logs
@@ -273,13 +274,13 @@ sub qesap_execute {
     $exec_log .= '.log.txt';
     $exec_log =~ s/[-\s]+/_/g;
 
-    my $qesap_cmd = join(" ", "python3.9", $paths{deployment_dir} . "/scripts/qesap/qesap.py",
+    my $qesap_cmd = join(' ', 'python3.9', $paths{deployment_dir} . '/scripts/qesap/qesap.py',
         $verbose,
-        "-c", $paths{qesap_conf_trgt},
-        "-b", $paths{deployment_dir},
+        '-c', $paths{qesap_conf_trgt},
+        '-b', $paths{deployment_dir},
         $args{cmd},
         $args{cmd_options},
-        "|& tee -a",
+        '|& tee -a',
         $exec_log
     );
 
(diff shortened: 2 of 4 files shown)


lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+=back
+=cut
+
+sub qesap_ansible_script_output {
+    my (%args) = @_;
+    croak 'Missing mandatory cmd argument' unless $args{cmd};
+    croak 'Missing mandatory host argument' unless $args{host};
+    my $inventory = qesap_get_inventory($args{provider});
+
+    if (script_run('test -e script_output.yaml') != 0) {
```
alvarocarvajald, line 449: No need to check against `!= 0` .. `if (script_run 'test -e script_output.yaml') {` is enough.

data/sles4sap/script_output.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- hosts: all
+  remote_user: cloudadmin
```
alvarocarvajald, line 3: I'm wondering if the remote_user can be also an argument of the function to make this more general. I suppose it can be overwritten via the CLI. And if it's more general, we could even move it out of `data/sles4sap/` ... just thinking out loud, not requesting a change here.

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+=head3 qesap_ansible_script_output
+
+    Use Ansible to run a command remotely and get the stdout.
+    Command is executed with elevated privileges
+
+    qesap_ansible_script_output(cmd => 'crm status', provider => 'aws', host => 'vmhana01');
+
+    It uses playbook data/sles4sap/script_output.yaml
+
+    1. ansible-playbook run the playbook, remote cmd ar paramenter
```
alvarocarvajald, line 427: ar? Not sure whether it should be `remote cmd or parameter` or `remote cmd and parameter`. There's also a typo in parame**n**ter.

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+    2. the playbook executes the command and redirects the output to file, both remotely
+    3. the playbook download the file locally
+    4. the file is read and sored to be returned to the caller
```
alvarocarvajald, line 430: s/sored/stored/ ... although looking at the code I would say:
```
4. the file is read and its content is returned to the caller
```

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+
+    if (script_run('test -e script_output.yaml') != 0) {
+        my $cmd = join(' ',
+            'curl', '-v', '-fL',
+            data_url('sles4sap/script_output.yaml'),
+            '-o', 'script_output.yaml');
+        assert_script_run($cmd);
+    }
+
+    # ansible-playbook -l vmhana01 -vvvv -i ~/p_repo/qe-sap-deployment/terraform/azure/inventory.yaml script_output.yaml -e "cmd='crm status'"
```
alvarocarvajald, line 457: Dangling comment.

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+    my $ansible_cmd = join(' ',
+        'ansible-playbook',
+        '-l', $args{host},
+        '-vvvv',
+        '-i', $inventory,
+        'script_output.yaml',
```
alvarocarvajald, line 463: You use `script_output.yaml` in several places in the function. I'd suggest declaring it in a variable at the start of the function and that way if you ever need to rename the file, it would be enough to change it in the one place.

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+        '-e', "\"cmd='$args{cmd}'\"");
+    script_run('rm /tmp/ansible_script_output/testout.txt || echo "Nothing to clean"');
```
alvarocarvajald, line 465: Should this be a `script_run` if you don't care about its return value?

I'd go simply with:
```
enter_cmd 'rm -f /tmp/ansible_script_output/testout.txt';
```

Also, same suggestion regarding `/tmp/ansible_script_output/testout.txt`: you use it already 3 times in the function. You can define it in a variable at the start of the function and then if you ever need to change it, it would be enough to change it in the one place.

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,59 @@ sub qesap_ansible_cmd {
...
+    assert_script_run($ansible_cmd);
+    my $output = script_output('cat /tmp/ansible_script_output/testout.txt');
+    script_run('rm /tmp/ansible_script_output/testout.txt');
+    return $output;
```
alvarocarvajald, line 469: Suggestion for the 3 lines above:
```
return script_output('cat /tmp/ansible_script_output/testout.txt && rm /tmp/ansible_script_output/testout.txt');
```
Or if you use a variable for `/tmp/ansible_script_output/testout.txt`:
```
return script_output("cat $ansible_out && rm $ansible_out');
```

alvarocarvajald (changes requested):

lib/qesapdeployment.pm
mpagot: was intentional but I'm ok to remove it. Was to help the reader to figure out how the command composition is working

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,73 @@ sub qesap_ansible_cmd {
...
+            '-o', $pb);
+        assert_script_run($cmd);
+    }
+
+    my @ansible_cmd = ('ansible-playbook', '-vvvv', $pb);
+    push @ansible_cmd, ('-l', $args{host}, '-i', $inventory);
+    push @ansible_cmd, ('-u', $args{user});
+    if ($args{root}) {
+        push @ansible_cmd, ('-b', '--become-user', 'root');
+    }
```
alvarocarvajald, line 474:
```
push(@ansible_cmd, '-b', '-- become-user', 'root') if ($args{root});
```

lib/qesapdeployment.pm
```diff
@@ -413,4 +414,73 @@ sub qesap_ansible_cmd {
...
+    push @ansible_cmd, ('-e', "\"cmd='$args{cmd}'\"",
```
alvarocarvajald, line 475: You can use `qq` here:
```
push @ansible_cmd, ('-e', qq("cmd='$args{cmd}'"),
```

alvarocarvajald (approved):
A couple of small things, but LGTM

PR #15933, opened 2022-11-17
Expect ip instead of FQDN in ibft validation

ge0r:
After a recent change in the worker salt states, the variable `WORKER_HOSTNAME` will always return the FQDN.
Since in ibft test the expected ip address was initialized using the `WORKER_HOSTNAME` variable, this means that currently there is a difference between the expected ip value and the actual ip value. We mitigate this by acquiring the expected ip based on the FQDN.

- Related ticket: https://progress.opensuse.org/issues/120079
- Needles: No needles
- Verification run:  https://openqa.suse.de/tests/9987191#step/ibft/57

### DIFF
--- a/tests/installation/validation/ibft.pm
+++ b/tests/installation/validation/ibft.pm
@@ -14,6 +14,7 @@
 use base "opensusebasetest";
 use strict;
 use warnings;
+use Socket;
 use testapi;
 use Test::Assert 'assert_equals';
 use version_utils qw(is_sle);
@@ -35,7 +36,6 @@ my $ibft_expected = {
         target_name => get_required_var('NBF'),
         flags => 3,
         port => 3260,
-        ip_addr => get_required_var('WORKER_HOSTNAME'),
         chap_type => 0
     },
     acpi_header => {
@@ -112,6 +112,9 @@ sub run {
     my $reg_rx = qr/rxdata_octets:\s+\d+/;
     # Requires NICTYPE=user and backend/qemu.pm code to run
     $ibft_expected->{ethernet}->{mac} = get_required_var('NICMAC');
+
+    my $fqdn = testapi::get_required_var('WORKER_HOSTNAME');
+    $ibft_expected->{target}->{ip_addr} = inet_ntoa(inet_aton($fqdn));
 
     select_console 'root-console';
     # find iscsi drive


coolgw:
@okurz   do you have any idea https://openqa.suse.de/tests/9980914#  fail reason? is there any change in openqa server side?

tests/installation/validation/ibft.pm
```diff
@@ -113,6 +112,9 @@ sub run {
     # Requires NICTYPE=user and backend/qemu.pm code to run
     $ibft_expected->{ethernet}->{mac} = get_required_var('NICMAC');
 
+    my $fqdn = testapi::get_required_var('WORKER_HOSTNAME');
+    $ibft_expected->{target}->{ip_addr} = script_output("dig +short $fqdn");
```
okurz, line 116: this might work but WORKER_HOSTNAME allows to still use an IP address for which dig will not return any result. Also, why go the extra mile to pass a command to the SUT and then back? Here you don't actually want to test the behaviour of "dig", do you?

okurz (changes requested):

okurz:
> @okurz do you have any idea https://openqa.suse.de/tests/9980914# fail reason?

No idea. I just see that the command fails, nothing else.

> is there any change in openqa server side?

Yes, always :) Just depends on "since when". I suggest to check the "Investigation information" as well as look into openqa-investigate results. For the example you mentioned as this is a personally triggered job that information does not exist but it might exist for the production job template you used

coolgw:
> > @okurz do you have any idea https://openqa.suse.de/tests/9980914# fail reason?
> 
> No idea. I just see that the command fails, nothing else.
> 
> > is there any change in openqa server side?
> 
> Yes, always :) Just depends on "since when". I suggest to check the "Investigation information" as well as look into openqa-investigate results. For the example you mentioned as this is a personally triggered job that information does not exist but it might exist for the production job template you used

https://openqa.suse.de/tests/9974171#step/iscsi_configuration/11 is error case which i know, what i can see is from build 29.1, the issue happen from time to time.  So some network env related issue?

okurz:
> https://openqa.suse.de/tests/9974171#step/iscsi_configuration/11 is error case which i know, what i can see is from build 29.1, the issue happen from time to time. So some network env related issue?

That job is an "openqa-investigate" job itself, not a "production job" for which we would have investigation info and openqa-investigate jobs. Following the URL in OPENQA_INVESTIGATE_ORIGIN in https://openqa.suse.de/tests/9974171#settings we can find https://openqa.suse.de/tests/9970137. https://openqa.suse.de/tests/9970137#next_previous shows that the same error already appeared 29 days in https://openqa.suse.de/tests/9759546#step/iscsi_configuration/12 which was before the work on https://progress.opensuse.org/issues/119443 was started (22 days ago) so not related to a "network zone migration". And I am not aware of other generic network issues at that time. So I suspect that this is not a specific network infrastructure problem but a general test design problem and should be handled accordingly with making the test more resilient.

tests/installation/validation/ibft.pm
coolgw: @okurz 
I agree we need check WORKER_HOSTNAME is fqdn or ip address before do fqdn to ip action, but if you think use dig command on SUT is not good idea, then what's your suggestion to get map relation between WORKER_HOSTNAME & ip?

okurz: > Can grains['fqdn'][0] [here](https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/766/diffs#61f80b79fa50900862d72f5ea55b805b8ca14fb8_154_153) contain an IP address?

No. But os-autoinst-distri-opensuse shouldn't code against how the infrastructure is managed

> but if you think use dig command on SUT is not good idea, then what's your suggestion to get map relation between WORKER_HOSTNAME & ip?

you can still use `dig` but why call it within the test VM and not just on the worker?

How about

```suggestion
use Socket;
…
    $ibft_expected->{target}->{ip_addr} = inet_ntoa(inet_aton($fqdn));
```

I tried it out with:

```
> print(inet_ntoa(inet_aton('openqa.suse.de')))
10.160.0.207
> print(inet_ntoa(inet_aton('10.160.0.207')))
10.160.0.207
```

ge0r: Thank you for the suggestion Oliver! I understand your reasoning, this works great

tests/installation/iscsi_configuration.pm
```diff
@@ -31,7 +31,7 @@ sub run {
     assert_screen 'iscsi-ibft';
     send_key 'alt-o';    # OK
     assert_screen 'disk-activation-iscsi';
-    wipe_iscsi_disk;    # At this point should be mounted and can proceed to erase it
+    # wipe_iscsi_disk;    # At this point should be mounted and can proceed to erase it
```
okurz, line 34: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

okurz (changes requested):

tests/installation/iscsi_configuration.pm
ge0r: This is a separate commit, to be removed before I remove the WIP label. I assumed since this PR is still WIP with no VR it would not be an issue

ge0r: Makes sense though, thanks for bringing it up

okurz (approved):

coolgw (approved):

PR #15935, opened 2022-11-17
Add Win11 case to Windows Update

pablo-herranz:
The idea here is to update Windows right after the installation, and then run the update process in a separate job who will be pre-requisite to all WSL ones.

- Related ticket: https://progress.opensuse.org/issues/117577
- Needles: *so many...*
- Verification runs:
#### Install and update
  - https://openqa.suse.de/tests/9987150
  - https://openqa.suse.de/tests/9987246
  - https://openqa.suse.de/tests/10004268
#### Just update
  - https://openqa.suse.de/tests/9987552
  - https://openqa.suse.de/tests/9987603
  - https://openqa.suse.de/tests/10013051

### DIFF
--- a/tests/wsl/install/ms_win_firstboot.pm
+++ b/tests/wsl/install/ms_win_firstboot.pm
@@ -157,7 +157,8 @@ sub run {
     );
 
     # poweroff
-    $self->reboot_or_shutdown;
+    $self->reboot_or_shutdown(1);
+    $self->wait_boot_windows;
 }
 
 1;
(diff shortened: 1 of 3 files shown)


tests/wsl/install/ms_win_firstboot.pm
```diff
@@ -157,7 +157,8 @@ sub run {
     );
 
     # poweroff
-    $self->reboot_or_shutdown;
+    $self->reboot_or_shutdown(1);
```
grisu48, line 160: No for this PR, but one might consider a refactoring where we are creating `reboot` and `shutdown` functions in windowsbasetest, so that we don't have to do such cryptic calls.

Note for my future self: The 1 means that it's a reboot and not a shutdown

```
sub reboot_or_shutdown {
    my ($self, $is_reboot) = @_;
```

grisu48 (approved):

pdostal:
> * Needles: _so many..._

OK :smile:

PR #15940, opened 2022-11-18
Collect AVC generated by each test module

mloviska:
Search for AVC denials after execution of each test module that runs in
sle|leap-micro, MicroOS or ALP test suites.
Eventually as `journal_check` is scheduled at the end test suites,
upload content of `audit` log directory for further investigation if
required.

- ticket: [Upload and check the content of audit.log](https://progress.opensuse.org/issues/120070)

#### VRs:

* [Build20221121-1 of sle-micro-5.2-DVD-Updates.x86_64](http://kepler.suse.cz/tests/19633#live)
* [sle-micro-5.3-DVD-Updates-x86_64-Build20221121-1](http://kepler.suse.cz/tests/19632#)
* [sle-micro-5.1-DVD-Updates-x86_64-Build20221121-1](http://kepler.suse.cz/tests/19631#live)
* [microos-Tumbleweed-MicroOS-Image-x86_64-Build20221116-microos@64bit](http://kepler.suse.cz/tests/19630#)
* [sle-micro-5.3-MicroOS-Image-Updates-x86_64-Build20221117-1-slem_selinux@64bit](http://kepler.suse.cz/tests/19627#step/cockpit_service/81)
* [sle-micro-5.2-MicroOS-Image-Updates-x86_64-Build20221120-1-slem_selinux@64bit](http://kepler.suse.cz/tests/19629#step/disk_boot/19)
* [sle-micro-5.1-MicroOS-Image-Updates-x86_64-Build20221120-1-slem_selinux@64bit](http://kepler.suse.cz/tests/19628#step/install_updates/43)
* [alp-0.1-kvm-x86_64-Build21.3-alp_default@uefi](https://openqa.opensuse.org/tests/2886374#step/cockpit_service/84)

### DIFF
--- a/lib/consoletest.pm
+++ b/lib/consoletest.pm
@@ -11,6 +11,11 @@ use known_bugs;
 use version_utils qw(is_public_cloud is_openstack);
 use utils;
 
+my %avc_record = (
+    start => 0,
+    end => undef
+);
+
 =head1 consoletest
 
 C<consoletest> - Base class for all console tests
@@ -29,6 +34,7 @@ sub post_run_hook {
     # start next test in home directory
     enter_cmd "cd";
 
+    record_avc_selinux_alerts();
     # clear screen to make screen content ready for next test
     $self->clear_and_verify_console;
 }
@@ -41,6 +47,7 @@ Method executed when run() finishes and the module has result => 'fail'
 
 sub post_fail_hook {
     my ($self) = @_;
+    record_avc_selinux_alerts();
     $self->SUPER::post_fail_hook;
     # at this point the instance is shutdown
     return if (is_public_cloud() || is_openstack());
@@ -49,6 +56,31 @@ sub post_fail_hook {
     $self->export_logs_basic;
     # Export extra log after failure for further check gdm issue 1127317, also poo#45236 used for tracking action on Openqa
     $self->export_logs_desktop;
+}
+
+=head2 record_avc_selinux_alerts
+
+List AVCs that have been recorded during a runtime of a test module that executes this function
+
+=cut
+
+sub record_avc_selinux_alerts {
+    if ((current_console() !~ /root|log/) || (script_run('test -f /var/log/audit/audit.log') != 0)) {
+        return;
+    }
+
+    my @logged = split(/\n/, script_output('ausearch -m avc -r', proceed_on_failure => 1));
+
+    # no new messages are registered
+    if ($avc_record{end} <= $avc_record{start}) {
+        return;
+    }
+
+    $avc_record{end} = scalar @logged - 1;
+    my @avc = @logged[$avc_record{start} .. $avc_record{end}];
+    $avc_record{start} = $avc_record{end} + 1;
+
+    record_info('AVC', join("\n", @avc));
 }
 
 =head2 use_wicked_network_manager
--- a/tests/console/journal_check.pm
+++ b/tests/console/journal_check.pm
@@ -107,6 +107,13 @@ sub run {
             $failed = 1;
         }
     }
+
+    # upload all content of audit directory
+    if (script_run('test -d /var/log/audit/') == 0) {
+        assert_script_run('tar cvf /tmp/audit.tar  /var/log/audit/*');
+        upload_logs('/tmp/audit.tar');
+    }
+
     $self->result('fail') if $failed;
 }
 
--- a/tests/microos/cockpit_service.pm
+++ b/tests/microos/cockpit_service.pm
@@ -6,7 +6,7 @@
 # Summary: Basic check for cockpit service
 # Maintainer: qa-c team <qa-c@suse.de>
 
-use base "opensusebasetest";
+use base "consoletest";
 use strict;
 use warnings;
 use testapi;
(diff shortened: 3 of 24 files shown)


mloviska:
Hello @grisu48, @pdostal, @Vogtinator! Would you find some time for review, please? We need to start to collect the AVC for SELinux devs.

lib/consoletest.pm
```diff
@@ -51,6 +58,31 @@ sub post_fail_hook {
     $self->export_logs_desktop;
 }
 
+=head2 get_avc_selinux_alerts
+
+List AVCs that have been recorded during a runtime of a test module that executes this function
+
+=cut
+
+sub get_avc_selinux_alerts {
```
grisu48, line 67: The name `get_avc_selinux_alerts` suggests that something will be returned, which is not the case when messages are retrieved from the AVC.

Given how I **think** this subroutine was intended, I would rename it to `record_selinux_alerts` and replace the `return undef` with simple `return`, as the result is anyways not used anywhere.

lib/consoletest.pm
```diff
@@ -11,6 +11,11 @@ use known_bugs;
 use version_utils qw(is_public_cloud is_openstack);
 use utils;
 
+my %avc_record = (
+    start => 0,
```
Vogtinator, line 15: Is this shared between all instances of `consoletest`? If not, all tests will start at 0

Vogtinator: Ok - perl is still an alien language to me

lib/consoletest.pm
```diff
@@ -51,6 +58,31 @@ sub post_fail_hook {
...
+List AVCs that have been recorded during a runtime of a test module that executes this function
+
+=cut
+
+sub get_avc_selinux_alerts {
+    if ((current_console() !~ /root|log/) || (script_run('test -f /var/log/audit/audit.log') != 0)) {
+        return undef;
+    }
+
+    my @logged = split(/\n/, script_output('ausearch -m avc -r', proceed_on_failure => 1));
```
Vogtinator, line 72: You could do `ausearch -m avc -r | tail -n +${avc_record{start}}` here to skip the messages earlier in the flow.

lib/consoletest.pm
```diff
@@ -51,6 +58,31 @@ sub post_fail_hook {
...
+    $avc_record{end} = scalar @logged - 1;
```
Vogtinator, line 73: If `ausearch` fails or returns less messages than expected, end can go backwards. I suggest moving the assignment past the check below.

Vogtinator: Ideally this would not skip based on line numbers but timestamps

lib/consoletest.pm
mloviska: That would indeed work, but I still find it useful to see the whole output of the `ausearch` if something would go wrong. e.i. a bug will be introduced to the utility itself etc.

lib/consoletest.pm
mloviska: Something like extracting from the last line `msg=audit(1669121672.441:16)` and whatever that comes after ?! Isn't the first float `1669121672.441` a date information? I am thinking whether the output would not get mixed up if the `chrony` would make a time correction somewhere in the middle of the test. On slower machines, we can see here and there that chrony/ntp was doing some time corrections.

Vogtinator: I'd hope that the logs are still in order, so by omitting everything until the first timestamp match you'll never miss anything new.

mloviska: Ok, let's try it out. I have used array indexes in order to return a slice at the end.

pdostal (approved):

coolo:
merge?

mloviska:
> merge?

I am trying to apply Fabian's suggestion from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15940#discussion_r1030563194. But if you are in a hurry, maybe this can be done as an improvement in another ticket.

coolo:
I'm not exactly in a hurry, but as I understand PED-2493, this is just the start

mloviska:
> I'm not exactly in a hurry, but as I understand PED-2493, this is just the start

Alright, let's merge it, as further updates will be necessary anyway.

Vogtinator:
> > I'm not exactly in a hurry, but as I understand PED-2493, this is just the start
> 
> Alright, let's merge it, as further updates will be necessary anyway.

No objections from my side FWIW.

lib/consoletest.pm
```diff
@@ -51,6 +58,31 @@ sub post_fail_hook {
...
+
+sub record_avc_selinux_alerts {
+    if ((current_console() !~ /root|log/) || (script_run('test -f /var/log/audit/audit.log') != 0)) {
+        return;
+    }
+
+    my @logged = split(/\n/, script_output('ausearch -m avc -r', proceed_on_failure => 1));
+
+    # no new messages are registered
+    if ($avc_record{end} <= $avc_record{start}) {
```
Vogtinator, line 75: This looks like it's missing some code in between. Or should this be

`if (scalar @logged <= $avc_record{start}) {`

mloviska: damn, forgot to check whether the list `@logged` contains anything reasonable data. Thanks

mloviska: But your suggestion should be enough anyway.

mloviska: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15975

PR #15942, opened 2022-11-18
Skip registry installations on unsupported SLE versions

asmorodskyi:


Vogtinator:
Is there any more info (ticket, bug report, etc.)?

asmorodskyi:
> Is there any more info (ticket, bug report, etc.)?

is this one is enough https://packagehub.suse.com/support/ ? 
I came with this PR after working on this bug https://bugzilla.suse.com/show_bug.cgi?id=1205044 but it is actually unrelated . Just using common sense - installing docker container registry from unsupported package on SLE which is in LTSS what we are proving ? If the only impediment is absence of progress issue for reference I will happily create one . 
Also initially I wanted to omit this only for runs in Public Cloud flavors but after getting support from @ilausuch  on this topic I decided to exclude it unconditionally . If I will resistance from you or anyone else I will tweak to exclude only in Public Cloud flavors ...

Vogtinator:
> is this one is enough https://packagehub.suse.com/support/ ?

Which part?

I'm mostly wondering why this should be removed. Either it's

a) working and we could keep it
b) not working and you can link to a failure
c) not working and we don't test it

or even
d) working and we don't test it

asmorodskyi:
> > is this one is enough https://packagehub.suse.com/support/ ?
> 
> Which part?

the part where SUSE stating that there is no support for packages installed from PackageHub

> 
> I'm mostly wondering why this should be removed. 

Because I don't see good reason to run scenario where we installing  container registry from unsupported package on SLE version which in LTSS . And "it is working " is not good one for me 

 
>  b) not working and you can link to a failure 

currently it is not working for SLE 12 SP4 / SP5 in Public Cloud see https://bugzilla.suse.com/show_bug.cgi?id=1205044 .

Vogtinator:
> the part where SUSE stating that there is no support for packages installed from PackageHub

No L3 support. That's a huge difference. This argument applies equally to supported versions of SLE.

> currently it is not working for SLE 12 SP4 / SP5 in Public Cloud see https://bugzilla.suse.com/show_bug.cgi?id=1205044 .

Looks like a SLE 12 specific issue. Maybe skip it on SLE 12?

asmorodskyi:
> > the part where SUSE stating that there is no support for packages installed from PackageHub
> 
> No L3 support. That's a huge difference. This argument applies equally to supported versions of SLE.
> 
> > currently it is not working for SLE 12 SP4 / SP5 in Public Cloud see https://bugzilla.suse.com/show_bug.cgi?id=1205044 .
> 
> Looks like a SLE 12 specific issue. Maybe skip it on SLE 12?

Done . Also  here is VR http://autobot.qa.suse.de/tests/3503#

ilausuch (approved):
I talked with Anton and I agree whit this idea.

asmorodskyi:
> But SLES 12-SP5 is still supported.

supported by what ?

asmorodskyi:
quote from @msmeissn : 
" ... fwiw distribution-registry was only added in sles15 sp4, not sure why this testcase tries it on earlier releases ... as far as I see it was not even on packagbehub"
" e.g the registry test should only be scheduled on sles15 sp4+ and leap 15.4 "

pdostal:
> supported by what ?

Supported, e.g. not in LTSS yet.

> " e.g the registry test should only be scheduled on sles15 sp4+ and leap 15.4 "

OK.

pdostal (approved):

PR #15951, opened 2022-11-21
Increase boot time for svirt backend

Zaoliang:
svirt banckend needs much longer for reboot.
see https://progress.opensuse.org/issues/120747

VR:
https://openqa.suse.de/tests/10005042#next_previous
https://openqa.suse.de/tests/10011722
https://openqa.suse.de/tests/10011724
https://openqa.suse.de/tests/10011724

### DIFF
--- a/tests/console/journalctl.pm
+++ b/tests/console/journalctl.pm
@@ -24,7 +24,7 @@ use testapi;
 use serial_terminal 'select_serial_terminal';
 use utils qw(zypper_call script_retry systemctl);
 use version_utils qw(is_opensuse is_tumbleweed is_sle is_public_cloud is_leap);
-use Utils::Backends qw(is_hyperv);
+use Utils::Backends qw(is_hyperv is_svirt_except_s390x);
 use Utils::Architectures;
 use power_action_utils qw(power_action);
 use publiccloud::instances;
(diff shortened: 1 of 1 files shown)


Zaoliang:
https://openqa.opensuse.org/tests/2893802 (tumbleweed)

tests/console/journalctl.pm
```diff
@@ -87,8 +87,10 @@ sub reboot {
         my $instance = publiccloud::instances::get_instance();
         $instance->softreboot();
     } else {
+        # svirt backend needs much longer for reboot
+        my $timeout = (is_svirt_except_s390x && is_sle) ? 600 : 300;
```
pdostal, line 91: Why `is_sle`?

Zaoliang: becasue at moment we have only issue for sle, don't want extend timeout for opensuse

Zaoliang:
> 

https://openqa.suse.de/tests/10005042#next_previous retriggered test failed at that time, if this is only driver issue which means setup issue and it is working now, then we don't need this workaround for extend timeou, but we need to capture this driver issue and need a record_softfail :)

Zaoliang:
I am not sure that is about a driver issue here, if the xen host told us this issue, then we cannot fix on our side. I have 100 tests run and didn't see this.

rfan1:
> I am not sure that is about a driver issue here, if the xen host told us this issue, then we cannot fix on our side. I have 100 tests run and didn't see this.

I agree with you we can't reproduce the issue easily, then I checked the latest job https://openqa.suse.de/tests/10005042/logfile?filename=autoinst-log.txt
The VM took ~2 minutes to boot up. so I would suggest you hold on merging this PR and wait for next openqa job :)

tests/console/journalctl.pm
pdostal: That is not valid reason. It may fail also in openSUSE in the future. Just increase the timeout globally.

PR #15965, opened 2022-11-23
Leave 'post_fail_hook' early if 'NOLOGS' is set

rfan1:
https://progress.opensuse.org/issues/120729

For some reason, somebody might not want to collect logs, but still run the post_fail_hooks, and we only have that for opensusebasetest but not for subclasses of it
This will help to run the post_fail_hooks of specific test modules that are also calling $self->SUPER::post_fail_hook where collecting those extra logs isn't really necessary

- Related ticket: https://progress.opensuse.org/issues/120729
- Verification run: n/a

### DIFF
--- a/lib/apparmortest.pm
+++ b/lib/apparmortest.pm
@@ -802,6 +802,7 @@ Run post_fail_hook and upload audit logs
 sub post_fail_hook {
     my ($self) = shift;
 
+    return if get_var('NOLOGS');
     # Exit x11 and turn to console in case
     send_key("alt-f4");
     select_console("root-console");
--- a/lib/bootbasetest.pm
+++ b/lib/bootbasetest.pm
@@ -27,6 +27,7 @@ sub post_fail_hook {
 
     # crosscheck for text login on tty1
     select_console 'root-console';
+    return if get_var('NOLOGS');
     # call parent's post fail hook
     $self->SUPER::post_fail_hook;
 
--- a/lib/y2_base.pm
+++ b/lib/y2_base.pm
@@ -161,6 +161,7 @@ sub upload_widgets_json {
 
 sub post_fail_hook {
     my $self = shift;
+    return if get_var('NOLOGS');
     upload_widgets_json();
     $self->SUPER::post_fail_hook;
 }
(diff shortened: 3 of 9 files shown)


Zaoliang:
LGTM

pdostal (dismissed):
This is big change. I would definitely inform wider group of people about this change.

rfan1:
@foursixnine 
Could you please involve more experts to review this PR?

lib/apparmortest.pm
```diff
@@ -802,6 +802,8 @@ Run post_fail_hook and upload audit logs
 sub post_fail_hook {
     my ($self) = shift;
 
+    return if (get_var('NOLOGS'));
+
```
okurz, line 806: Please use just `return if get_var('NOLOGS')` without the additional pair of parentheses as there is only a single argument.

lib/bootbasetest.pm
```diff
@@ -7,6 +7,7 @@ use warnings;
 
 sub post_fail_hook {
     my ($self) = @_;
+
```
okurz, line 10: unrelated change

lib/y2_base.pm
```diff
@@ -160,6 +160,9 @@ sub upload_widgets_json {
 
 sub post_fail_hook {
     my $self = shift;
+
+    return if (get_var('NOLOGS'));
+
```
okurz, line 165: I consider this too verbose. Just leave out the blank lines please. This applies for all changes

okurz (changes requested):

lib/apparmortest.pm
rfan1: Thanks @okurz 
fixed as your kindly suggestion.

foursixnine (approved):
Looks good for me, @os-autoinst/tests-maintainer any comments or objections?

alvarocarvajald (approved):
LGTM

jlausuch (approved):

dzedro (approved):

ge0r (approved):
LGTM

lemon-suse:
It's a good idea and save time for debug. If we can find easy method to apply this to overwritten post_fail_hook, then that will be perfect. :)

Just please update the PR description and remove things from template if you don't need them any more.

okurz (approved):

PR #15969, opened 2022-11-24
Add milestone for 'add_phub_extension'

rfan1:
https://progress.opensuse.org/issues/120393

Add milestone flag to save setup into the snapshot which 'packagehub' is activated, then all dependent jobs can re-use it

- Related ticket: https://progress.opensuse.org/issues/120393
- Verification run: https://openqa.suse.de/tests/10027279 [we can see even test `python_flake8` failed, the test `vmstat` can   roll back to new snapshot and pass]

### DIFF
--- a/tests/console/add_phub_extension.pm
+++ b/tests/console/add_phub_extension.pm
@@ -19,4 +19,10 @@ sub run {
     add_suseconnect_product(get_addon_fullname('phub'));
 }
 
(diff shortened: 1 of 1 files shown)


Avinesh (approved):
looks good.

DeepthiYV (commented):
LGTM!

PR #15977, opened 2022-11-24
Move logging subroutines from opensusebasetest

pdostal:
Move logs related subroutines which do not make sense in `opensusebasetest` to `Utils::Loging` and some not related subroutines to `utils`. Some subroutines removed because of the `post_fail_hook` inheritance.

That is required for deduplication of the code in order to log exporting speedup in upcoming commits.

- Related ticket: [poo#120735](https://progress.opensuse.org/issues/120735)
- Verification run: [extra_tests_gnome@64bit](https://pdostal-server.suse.cz/tests/3357), [extra_tests_textmode@64bit](https://pdostal-server.suse.cz/tests/3358), [ha_qdevice_node1@64bit](https://pdostal-server.suse.cz/tests/3382), [qam-wicked_basic_sut@64bit](https://pdostal-server.suse.cz/tests/3366), [publiccloud_containers@64bit](https://pdostal-server.suse.cz/tests/3370), [kubectl_tests@64bit](https://pdostal-server.suse.cz/tests/3408)

### DIFF
--- a/lib/Utils/Logging.pm
+++ b/lib/Utils/Logging.pm
@@ -1,4 +1,4 @@
-# Copyright 2022 SUSE LLC
+# Copyright SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 =head1 Utils::Logging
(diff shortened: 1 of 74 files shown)


okurz (approved):

b10n1k (changes requested):
I dont get it. I read the poo and i dont see how this is related to the ticket. Ticket description mentions *Decrease typing on `post_fail_hooks` log collection* (Commit title and description needs to be fixed too btw). is this a RFC or actual PR?

First question is why do you need that (it doesnt decrease typing, does it)?
Second why do you move the files into a new module and not into `Utils::Logging::`?

pdostal:
> First question is why do you need that (it doesnt decrease typing, does it)?

Before I start to work on speeding the `post_fail_hook` up I want to clean it.

> Second why do you move the files into a new module and not into `Utils::Logging::`?

Good idea! It's moved now.

@b10n1k please re-review.

b10n1k (changes requested):
It looks better IMO. However i found some lines removed which seems to make sense in some cases, i am not sure in others. And it is hard to know the why in each case as this is a massive commit. One of the cases i am talking about is on lib/locale.pm where `export_logs_basic` is removed. I assume that this is because of post_fail_hook but still the later calls `export_logs` which is a little different process. I wonder if this is necessary on that first stage.
In addition, i would like to ask again to fix commit message and elaborate the description with some details. That will help the review as well.

pdostal:
Yes, I removed `export_logs_basic ` from `lib/locale.pm` as it is based on `opensusebasetest` which calls `export_logs` which then calls `export_logs_basic` so it makes no sense to call it twice. I am going to clean this later.

I've updated the commit message, the first verification run is finished and more are in process.

foursixnine (approved):

PR #15981, opened 2022-11-25
Bump timeout of openldap test, seems to be not enough under load

tjyrinki:
- Related ticket: https://progress.opensuse.org/issues/
- Verification run: https://openqa.suse.de/tests/10038608

Even though it doesn't solve the problem of the day, it sounds like this could be an useful change since sometimes the test would look like failing already earlier, because of the timeout


foursixnine:
This is also https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16003

PR #15983, opened 2022-11-25
Create & apply default schedule for svirt-xen-pv backend

rakoenig:
Create a default file for the new scheduler that can be used for tests on x86_64 svirt-xen-pv backend. 

- Related ticket: https://progress.opensuse.org/issues/119890 
  (this is just part 1 for the svirt-xen-pv backend)
- Default file created
- Adapting test suite YAML files that use the svirt-xen-pv backend
- Releated MR for the job group definition: https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/445
- Needles: 
- [Verification runs](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=rakoenig%2Fos-autoinst-distri-opensuse%23default_x86_64_svirt_xen_pv&groupid=96)

Still some tests to adapt, but I start here to collect successful verification runs.


schedule/yast/lvm_raid1/lvm+raid1_sle15_default.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
jknphy, line 1: we don´t need to call them `_default.yaml` but use the architecture/backend used.
Later we will refactor to have only 1 file without extension.
For the extension, better going with underscore i.e.: `_arch_backend.yaml`

schedule/yast/lvm_raid1/lvm+raid1_sle15_default.yaml
```diff
@@ -0,0 +1,18 @@
...
+name:           lvm+raid1_default
+description:    >
+  Validation of partitioning for raid1 on lvm
+  Installation of RAID1 using expert partitioner.
+# Note: Test data for different backends need to be
```
jknphy, line 6: you don't need this, this will be done later, now you can just focus on the arch/backend on your ticket duplicating or referencing same test data used for the existing yaml that will be untouched for other archs.

schedule/yast/sle/flows/default@svirt-xen-pv.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - installation/performing_installation/stop_timeout_system_reboot_now
+installation_logs:
+  - installation/logs_from_installation_system
+virsh_boot_hdd:
+  - installation/performing_installation/update_virsh_config_to_boot_from_hd
+confirm_reboot:
+  - installation/performing_installation/confirm_reboot
+reconnect_svirt:
+  - installation/performing_installation/reconnect_after_reboot
+grub: []
```
jknphy, line 55: not needed, will never be overwritten.

schedule/yast/sle/flows/default@svirt-xen-pv.yaml
```diff
@@ -0,0 +1,59 @@
+---
```
jknphy, line 1: I know that is a mess between the mutiple ways to write it spread in the repo, with `@` with `-` or with `_`, but at least for the default files, let's follow the same pattern, so `default_svirt_xen_pv.yaml`.
When this would be more mature I think we should a ticket to make all naming consistent, but it is fine for now.

schedule/yast/sle/guided_xfs/guided_xfs.yaml
```diff
@@ -2,9 +2,18 @@
 name:           guided_xfs
 description:    >
   Guided Partitioning installation with xfs filesystem.
+# Note: Test data for different backends need to be
```
jknphy, line 5: we don't need this comment anywhere, this will probably be obvious and documented when we try to have a single test suite in the next iterations.
It is not easy to say that this would work fine with other arch/backends, unless you want to provide extra verification, the easiest way is to create a new file for now rather than editing this one. The plan is to analyze test suite per test suite to have one yaml file and then we will be providing VR for each arch/backend to see that is consistent, even with openSUSE.

schedule/yast/sle/guided_xfs/guided_xfs.yaml
```diff
@@ -2,9 +2,18 @@
...
 schedule:
+  extension_module_selection:
+  - installation/module_registration/register_module_desktop
+  system_role:
+  - installation/system_role/accept_selected_role_SLES_with_GNOME
+  guided_hard_disks:
+    - installation/partitioning/select_guided_setup
+    - installation/partitioning/guided_setup/accept_default_part_scheme
   guided_filesystem:
     - installation/partitioning/guided_setup/select_filesystem_option_xfs
```
jknphy, line 18: we can just group this one with the previous key under `guided_partitioning`

schedule/yast/sle/guided_ext4/guided_ext4_svirt_xen.yaml
```diff
@@ -6,6 +6,12 @@ vars:
   FILESYSTEM: ext4
   YUI_REST_API: 1
 schedule:
+  extension_module_selection:
+    - installation/module_registration/register_module_desktop
+  system_role:
+    - installation/system_role/accept_selected_role_SLES_with_GNOME
+  guided_open:
```
jknphy, line 13: let's group here as well as suggested.

schedule/yast/lvm_raid1/lvm+raid1_sle15_svirt-xen-pv.yaml
```diff
@@ -8,31 +8,11 @@ vars:
   LVM: 1
```
jknphy, line 8: do we need LVM setting? perhaps we can drop it.

schedule/yast/minimal+base/minimal+base@yast-xen-pv.yaml
```diff
@@ -11,47 +11,29 @@ vars:
...
-  - console/zypper_ref
-  - console/ncurses
-  - console/glibc_sanity
-  - update/zypper_up
-  - console/zypper_lifecycle
-  - console/orphaned_packages_check
-  - console/validate_installed_patterns
-  - console/consoletest_finish
+  software:
+    - installation/select_patterns
```
jknphy, line 15: we now we have this super-fast test module: https://openqa.suse.de/tests/10030133#step/select_only_visible_patterns_from_top/2

schedule/yast/minimal+base/minimal+base@yast-xen-pv.yaml
```diff
@@ -11,47 +11,29 @@ vars:
...
+  default_systemd_target:
+    - installation/installation_settings/validate_default_target
+  installation_settings:
+    - installation/security/select_security_module_none
```
jknphy, line 19: this should be under `security`

schedule/yast/minimal+base/minimal+base@yast-xen-pv.yaml
```diff
@@ -11,47 +11,29 @@ vars:
...
+    - installation/launch_installation
+  system_preparation:
+    - console/system_prepare
+    - console/prepare_test_data
+    - console/consoletest_setup
+    - console/force_scheduled_tasks
+    - console/installation_snapshots
```
jknphy, line 26: from this one I would move all the test module to `system_validation`, perhaps `zypper_lr` and `zypper_ref` could be considered preparation, but to not change the order of the test module we can deal with that later.

schedule/yast/msdos/msdos@svirt-xen-pv.yaml
```diff
@@ -1,35 +1,17 @@
...
-  - installation/first_boot
-  - console/validate_partition_table_via_parted
-  - console/validate_blockdevices
-  - console/validate_free_space
-  - console/validate_read_write
+  suggested_partitioning:
+    - installation/partitioning/msdos_partition_table
+  default_systemd_target:
+    - installation/installation_settings/validate_default_target
+  installation_logs: []
```
jknphy, line 12: I think there is not reason to not do this, probably we missed in some point, it is good we do this iteration to notice.

schedule/yast/textmode/textmode_default.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
jknphy, line 1: let's provide postfix name `_svirt_xen.yaml` (or the variation you prefer for now :) ) instead of default.

schedule/yast/textmode/textmode_default.yaml
```diff
@@ -0,0 +1,31 @@
...
+description: |
+  Installation in textmode which selects System Role 'Text Mode'.
+vars:
+  YUI_REST_API: 1
+schedule:
+  system_preparation:
+    - console/system_prepare
+    - console/prepare_test_data
+    - console/consoletest_setup
+  default_systemd_target:
```
jknphy, line 12: this should be before `system_preparation` to follow the order in the screens

jknphy:
could you please reflect in the PR title and commit that this is the first batch for example including " - Part 1", I would also suggest to say that we are not only creating the default but also adapting the test suites to use it.

schedule/yast/textmode/textmode_default.yaml
jknphy: why there are two files for textmode?

rakoenig: Renamed my file to `textmode_svirt_xen_pv.yaml`, the `textmode_svirt_xen` file is the old scheduler and maybe still needed for svirt-xen-hvm backend. I will see when I work on svirt-xen-hvm.

PR #15984, opened 2022-11-25
Remove duplicate word "testing" from title in file raspberry_pi_hardware_testing_setup.md

asdil12:


DimStar77:
Imho the fix is proper - but the commit message is not helpful when you see this in the commit history

asdil12:
Updated commit message

PR #15996, opened 2022-11-29
xfstests: increase generate_result download log timeout

SeroSun:
s390x tests recently frequently fail in the download status log part. Increase download log timeout to avoid this kind of fails. http://openqa.suse.de/tests/10028309#step/generate_report/2

- Verification run: http://openqa.suse.de/tests/10049376


SeroSun:
The root cause is a known issue poo#120261, closing this PR

PR #15997, opened 2022-11-29
PublicCloud - split post_fail_hook and post_run_hook logic

mimi1vx:
https://progress.opensuse.org/issues/119761

This change adds new method destroy to basetest and to overrrides

Now `cleanup` - code which clean up after test module , used by post_run_hook -> for upload extra logs or change state of SUT after test module run

`destroy` starts after fatal fail or on test with `last` flag destroys public cloud instace, it can be overrided (added code) in testmodule itself

New test_flag `last` used by last module in testsuite, now `run_ltp.pm` and `ssh_interactive_end` for start destroy in post_run_hook

- VR's:
  - http://quasar.suse.cz/tests/1348
  - http://quasar.suse.cz/tests/1349
  - http://quasar.suse.cz/tests/1353

### DIFF
--- a/lib/publiccloud/aws_client.pm
+++ b/lib/publiccloud/aws_client.pm
@@ -74,6 +74,10 @@ sub configure_podman {
           . " | podman login --username AWS --password-stdin $full_name_prefix");
 }
 
+sub destroy {
+    my ($self) = @_;
+}
+
 sub cleanup {
     my ($self) = @_;
 }
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -283,10 +283,21 @@ sub parse_instance_id
 }
 
 =head2 cleanup
-This method is called called after each test on failure or success to revoke the credentials
+This method is called after each test module to cleanup
 =cut
 
 sub cleanup {
+    my ($self) = @_;
+    $self->SUPER::cleanup();
+    $self->provider_client->cleanup();
+}
+
+
+=head2 destroy
+This method is called called after each test on failure or success to revoke the credentials
+=cut
+
+sub destroy {
     my ($self, $args) = @_;
     select_host_console(force => 1);
 
@@ -295,8 +306,8 @@ sub cleanup {
     script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt");
     upload_logs("bootlog.txt", failok => 1);
 
-    $self->SUPER::cleanup();
-    $self->provider_client->cleanup();
+    $self->SUPER::destroy();
+    $self->provider_client->destroy();
 }
 
 1;
--- a/lib/publiccloud/basetest.pm
+++ b/lib/publiccloud/basetest.pm
@@ -103,23 +103,33 @@ sub provider_factory {
 
 sub cleanup {
     # to be overridden by tests
+    # non-destructive cleanup for test module started by post_run_hook
+    # primary for log collection | restore clean state after test module
+    # can and be called after each module run
     return 1;
+}
+
+sub destroy {
+    # to be overridden by tests
+    # destructive routine in post_fail_hook | last module of test
+    return 1;
+}
+
+sub _destroy {
+    my ($self) = @_;
+    # run module overriden destroy
+    eval { $self->destroy(); } or bmwqemu::fctwarn("self::destroy() failed -- $@");
+    # run provider destroy
+    if ($self->{run_args} && $self->{run_args}->{my_provider}) {
+        eval { $self->{run_args}->{my_provider}->destroy($self->{run_args}); } or bmwqemu::fctwarn("provider::destroy() failed -- $@");
+    }
 }
 
 sub _cleanup {
     my ($self) = @_;
-    die("Cleanup called twice!") if ($self->{cleanup_called});
-    $self->{cleanup_called} = 1;
-
+    # run module overriden cleanup
     eval { $self->cleanup(); } or bmwqemu::fctwarn("self::cleanup() failed -- $@");
-
-    my $flags = $self->test_flags();
-    # currently we have two cases when cleanup of image will be skipped:
-    # 1. Calling module needs to have publiccloud_multi_module => 1 test flag
-    # and not have fatal => 1. Job should not have result = 'fail'
-    return if ($flags->{publiccloud_multi_module} && !($self->{result} eq 'fail' && $flags->{fatal}));
-    # 2. Job should have PUBLIC_CLOUD_NO_CLEANUP defined and job should have result = 'fail'
-    return if ($self->{result} eq 'fail' && get_var('PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE'));
+    # run provider cleanup
     if ($self->{run_args} && $self->{run_args}->{my_provider}) {
         eval { $self->{run_args}->{my_provider}->cleanup($self->{run_args}); } or bmwqemu::fctwarn("provider::cleanup() failed -- $@");
     }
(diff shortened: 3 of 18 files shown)


lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,54 @@ sub provider_factory {
...
+    # non-destructive cleanup for test module started by post_run_hook
+    # primary for log collection | restore clean state after test module
+    # can and be called after each module run
     return 1;
 }
 
-sub _cleanup {
+sub destroy {
+    # to be overridden by tests
+    # destrictive routine in post_fail_hook | last module of test
```
pdostal, line 114:
```suggestion
    # destructive routine in post_fail_hook | last module of test
```

lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,54 @@ sub provider_factory {
...
+    # run module overriden cleanup
     eval { $self->cleanup(); } or bmwqemu::fctwarn("self::cleanup() failed -- $@");
-
-    my $flags = $self->test_flags();
-    # currently we have two cases when cleanup of image will be skipped:
-    # 1. Calling module needs to have publiccloud_multi_module => 1 test flag
-    # and not have fatal => 1. Job should not have result = 'fail'
-    return if ($flags->{publiccloud_multi_module} && !($self->{result} eq 'fail' && $flags->{fatal}));
-    # 2. Job should have PUBLIC_CLOUD_NO_CLEANUP defined and job should have result = 'fail'
-    return if ($self->{result} eq 'fail' && get_var('PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE'));
```
pdostal, line 122: Isn't the variable `PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE` now obsolete? If so, we should also remove it from `variables.md`.

pdostal (approved):
This looks good to me

lib/publiccloud/aws_client.pm
```diff
@@ -74,6 +74,10 @@ sub configure_podman {
           . " | podman login --username AWS --password-stdin $full_name_prefix");
 }
 
+sub destroy {
```
asmorodskyi, line 77: all `*_client` classes ( aws_client / azure_client / gce_client ) has  `cleanup` method empty . And it does make sense because `*_clients` classes are not responsible for any resources , so I don't see why we actually need them . So instead of adding one more phantom method which no one ever call can you please drop "cleanup" method below , together with useless `$self->provider_client->cleanup();` calls in provider's children ?

lib/publiccloud/azure.pm
```diff
@@ -283,10 +283,21 @@ sub parse_instance_id
...
 
 =head2 cleanup
-This method is called called after each test on failure or success to revoke the credentials
+This method is called after each test module to cleanup
 =cut
 
 sub cleanup {
+    my ($self) = @_;
+    $self->SUPER::cleanup();
+    $self->provider_client->cleanup();
```
asmorodskyi, line 292: and if you will follow my suggestion about  dropping `$self->provider_client->cleanup();`  you may also drop this implementation of "cleanup" because overloading method in child class just to call parent implementation does not make any sense

lib/publiccloud/azure.pm
```diff
@@ -283,10 +283,21 @@ sub parse_instance_id
...
+}
+
+
+=head2 destroy
+This method is called called after each test on failure or success to revoke the credentials
```
grisu48, line 297:
```suggestion
This method is called called after each test run (not module) on failure or success to revoke the credentials
```

I'd make it more explicit, that this is executed once per test run and not once per test module.

lib/publiccloud/azure.pm
```diff
@@ -295,8 +306,8 @@ sub cleanup {
     script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt");
     upload_logs("bootlog.txt", failok => 1);
 
-    $self->SUPER::cleanup();
-    $self->provider_client->cleanup();
+    $self->SUPER::destroy();
```
grisu48, line 309: According to the PR description, the `destroy` got introduced to decouple crucial cleanup steps (i.e. `terraform destroy`) from optional cleanup steps like collection and upload of logs.

But this part here couples them again, which omits the purpose of the PR. If `destroy` is also collecting all logs then I don't get why we should introduce a second cleanup routine.

lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,56 @@ sub provider_factory {
...
+sub _cleanup {
+    my ($self) = @_;
+    # run module overriden cleanup
     eval { $self->cleanup(); } or bmwqemu::fctwarn("self::cleanup() failed -- $@");
-
-    my $flags = $self->test_flags();
-    # currently we have two cases when cleanup of image will be skipped:
-    # 1. Calling module needs to have publiccloud_multi_module => 1 test flag
-    # and not have fatal => 1. Job should not have result = 'fail'
-    return if ($flags->{publiccloud_multi_module} && !($self->{result} eq 'fail' && $flags->{fatal}));
```
asmorodskyi, line 120: so what is happening here is rename of `publiccloud_multi_module` into `last` generally I don't mind , but : 
1. we defining this flag in quite some modules , you need to rename it in all this places . otherwise behavior will change in this tests 
2. what about "execute_cleanup" instead of "last" ?  IMHO is more verbose than just "last" : 
```
sub test_flags {
    return {
        fatal => 1,
        milestone => 0,
        execute_pubcloud_cleanup => 1
    };
}
```

lib/publiccloud/basetest.pm
asmorodskyi: why we dropping it ? how you would stop openQA from deleting VM in the cloud from now on ?

lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,56 @@ sub provider_factory {
 
 sub cleanup {
     # to be overridden by tests
+    # non-destructive cleanup for test module started by post_run_hook
```
grisu48, line 106: I'm very much confused by this. What is the purpose of a non-destructive cleanup after each test module? Where is this or will this used?

lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,56 @@ sub provider_factory {
...
+
+sub _destroy {
     my ($self) = @_;
-    die("Cleanup called twice!") if ($self->{cleanup_called});
-    $self->{cleanup_called} = 1;
+    # run module overriden destroy
+    eval { $self->destroy(); } or bmwqemu::fctwarn("self::destroy() failed -- $@");
+    # run provider destroy
+    if ($self->{run_args} && $self->{run_args}->{my_provider}) {
+        eval { $self->{run_args}->{my_provider}->destroy($self->{run_args}); } or bmwqemu::fctwarn("provider::destroy() failed -- $@");
```
asmorodskyi, line 124:
```suggestion
        eval { $self->{run_args}->{my_provider}->destroy($self->{run_args}); } or log::fctwarn("provider::destroy() failed -- $@");
```

see https://github.com/os-autoinst/os-autoinst/blob/master/bmwqemu.pm#L56

grisu48 (commented):
The code looks OK but I don't see why we are introducing a new cleanup routine. I don't see the need for it based on this PR.

The referenced [poo#119761](https://progress.opensuse.org/issues/119761) is about improving the test logic so that `fatal` errors will be caught better - Isn't there a way how to achive this without introducing another cleanup routine? I find this a bit confusing.

lib/publiccloud/basetest.pm
```diff
@@ -103,36 +103,56 @@ sub provider_factory {
...
+    # destructive routine in post_fail_hook | last module of test
+    return 1;
+}
+
+sub _destroy {
     my ($self) = @_;
-    die("Cleanup called twice!") if ($self->{cleanup_called});
-    $self->{cleanup_called} = 1;
+    # run module overriden destroy
+    eval { $self->destroy(); } or bmwqemu::fctwarn("self::destroy() failed -- $@");
```
asmorodskyi, line 121: I just briefly checked implementation of `destroy` in `azure` . I see that in this implementation there is also call to `SUPER::destroy()` which make it to be called two times ( first here , because `$self->destroy();` is calling parent implementation , and second time below in line 124 )

lib/publiccloud/basetest.pm
asmorodskyi: ok I see , you just moved variable into other place .

lib/publiccloud/azure.pm
asmorodskyi: there is no "revoke of credentials " anywhere in our code

pdostal: called called

lib/publiccloud/ec2.pm
```diff
@@ -177,6 +177,10 @@ sub img_proof {
 }
 
 sub cleanup {
+    my ($self) = @_;
```
asmorodskyi, line 180: for `gce::cleanup()` we are calling `SUPER::cleanup()` and here we don't any good reason for this ? generally I am against having this empty calls . let's keep only pieces which actually doing something . IMO it will be frustrating for person who will try to read and understand this code to jump between several files just to figure out that there is only empty stubs calling other empty stubs ...

PR #16012, opened 2022-11-30
Implement userspace livepatching tests

mdoucha:
New test based on ltp_openposix (ulp_openposix):
1. Install `libpulp` and `libpulp-tools`
2. Start all openposix tests in background (paused)
3. Livepatch all openposix test processes
4. Unpause all openposix test processes and check results as usual

New multithread livepatching test (ulp_threads):
1. Start test process with given number of threads
2. Apply livepatches
3. Send exit signal to test process
4. Check result

When the livepatch package applies to multiple glibc target versions, the above tests will be scheduled multiple times in the job, once for each target version.

New job settings:
- `LIBC_LIVEPATCH` - Enables the above test scenario if set to 1
- `ULP_THREAD_COUNT` - Number of threads to create in ulp_threads
- `ULP_THREAD_SLEEP` - Sleep length after each thread loop iteration in ulp_threads

----

- Needles: N/A
- Verification run: Pending (waiting for related LTP changes)

### DIFF
--- a/lib/LTP/utils.pm
+++ b/lib/LTP/utils.pm
@@ -231,7 +231,8 @@ sub read_runfile {
 }
 
 sub schedule_tests {
-    my ($cmd_file) = @_;
+    my ($cmd_file, $suffix) = @_;
+    $suffix //= '';
 
     my $test_result_export = {
         format => 'result_array:v2',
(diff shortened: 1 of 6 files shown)


lib/main_ltp.pm
```diff
@@ -42,9 +42,17 @@ sub load_kernel_tests {
...
+        if (get_var('LIBC_LIVEPATCH')) {
+            die 'LTP_COMMAND_FILE and LIBC_LIVEPATCH are mutually exclusive'
+              if get_var('LTP_COMMAND_FILE');
+            loadtest_kernel 'ulp_openposix';
+        }
+
         # If there is a command file then install_ltp schedules boot_ltp which
         # will schedule shutdown
-        shutdown_ltp() unless get_var('LTP_COMMAND_FILE');
+        shutdown_ltp()
```
b10n1k, line 54: Maybe update comment above as well?

lib/LTP/utils.pm
```diff
@@ -273,24 +274,39 @@ sub schedule_tests {
...
+      unless get_var('LIBC_LIVEPATCH');
 }
 
 sub parse_openposix_runfile {
     my ($name, $cmds, $cmd_pattern, $cmd_exclude, $test_result_export, $suffix) = @_;
+    my $ulp_test = get_var('LIBC_LIVEPATCH', 0);
     my $whitelist = LTP::WhiteList->new();
 
+    if ($ulp_test) {
+        assert_script_run('export LD_PRELOAD=/usr/lib64/libpulp.so.0');
```
b10n1k, line 294: this seems it can be one liner

b10n1k (commented):
it looks like you use new variables. Can we update the variables.pm too?

lib/main_ltp.pm
mdoucha: Updated the comment and added the variables to variables.md.

lib/LTP/utils.pm
mdoucha: Fixed. Originally there was one more debug command and I forgot to reformat the condition after removing it.

b10n1k (approved):

pevik (approved):
Looks good, thank you!

pevik:
@czerw FYI

czerw (approved):
LGTM

PR #16019, opened 2022-12-01
Increase timeout in upload logs check

ge0r:
Occasionally the default time out is not enough when checking if we can upload logs, and, while the ping command returns successfully, the script_run command times out. We increase the timeout to 60 seconds to mitigate this sporadic issue.

- Related ticket: https://progress.opensuse.org/issues/120429
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/10064769#step/logs_from_installation_system/4

### DIFF
--- a/lib/network_utils.pm
+++ b/lib/network_utils.pm
@@ -14,6 +14,7 @@ package network_utils;
 
 use base Exporter;
 use Exporter;
+use Socket;
 
 use strict;
 use warnings;
@@ -72,7 +73,8 @@ Returns if can ping worker host gateway
 sub can_upload_logs {
     my ($gw) = @_;
     $gw ||= testapi::host_ip();
-    return (script_run('ping -c 1 ' . $gw) == 0);
+    my $gw_ip = inet_ntoa(inet_aton($gw));
+    return (script_run('ping -c 1 -W 120 ' . $gw_ip, timeout => 120) == 0);
 }
 
 


rakoenig (dismissed):
LGTM

jknphy:
I would prefer to have the WIP on the title, it is easy to check to review from our filter:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pulls?q=is%3Aopen+is%3Apr+label%3Aqe-yam+sort%3Acreated-asc
is ok for you? (at least while we have this long list of PRs :) to check)

lib/network_utils.pm
```diff
@@ -72,7 +72,7 @@ Returns if can ping worker host gateway
 sub can_upload_logs {
     my ($gw) = @_;
     $gw ||= testapi::host_ip();
-    return (script_run('ping -c 1 ' . $gw) == 0);
+    return (script_run('ping -c 3 ' . $gw, 60) == 0);
```
coolgw, line 75: I suggest increase timeout for both script_run and ping but keep "-c 1"
script_run('time ping -c 1 -W 120' . $gw , timeout => 120)

ge0r: the problem is that no matter how big the timeout is, it will always sometimes fail. That is because when ping returns, sometimes the return code is not captured by openQA. When that happens, even if the ping returns successfully in 5 seconds, openQA will just wait for 120 seconds to say timed out.

I must say that we probably hit a bug that goes deep in the os-autoinst tooling with this one.

coolgw: ok, let's give a try in openqa.
Will be more confidence if we have more validation result.

coolgw (approved):

lib/network_utils.pm
jknphy: but wait a second, isn't this the same thing that we have fixed in a couple of places?
that fqdn seems to misbehave, we should use the ip of the worker.

ge0r: > but wait a second, isn't this the same thing that we have fixed in a couple of places? that fqdn seems to misbehave, we should use the ip of the worker.

This issue is unrelated to the other one (where fqdn was not misbehaving but it was that the ibft test needed an ip instead of fqdn, but there was recently a change where WORKER_HOSTNAME was no longer providing ip but fqdn). 

The problem here is that some times `script_run` times out (not the return code being a wrong value). It sometimes times out because it does not catch the return code, even though ping has returned. Ping returns fairly quickly, return code is not caught and script_run times out when the timeout duration is expired.

jknphy: as you described looks like some problem in os-autoinst.. a couple of questions:
(1) could you try with IP if we could find better rate of success for this unknown issue of script_run (if doesn't need to resolve anything, perhaps it is faster and the serial is read faster, just a theory...).
(2) does it help to increase the timeout or not? I don't get from the first comment if there is any improvement.

ge0r: On the second point, increasing timeout does not help, since the ping command returns quite quickly either way.
On the first point, I can try with IP, but I am convinced this is unrelated to speed of the command or how fast the serial output is read, it is just that, when it fails, it is never read, although the command has returned and we are on the next line on the command prompt.

jknphy: thanks, if you can take a look at least we discard that.
So if the timeout increase doesn't help do we need to merge this PR?

ge0r:
@coolgw unfortunately the suggestion with the `-W` parameter did not work, [here](https://openqa.suse.de/tests/10172061#step/logs_from_installation_system/4) is the failure (after 3 successful runs)

jknphy:
we could retry the command:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1688
and put some comment that for SLE 12 SP{4-5} there is a problem with the serial with a reference to a progress ticket for tools team, have you checked with them what could be?

ge0r:
> we could retry the command: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/utils.pm#L1688 and put some comment that for SLE 12 SP{4-5} there is a problem with the serial with a reference to a progress ticket for tools team, have you checked with them what could be?

I have already tried with the `script_retry` command, we still get time out failures.

lib/network_utils.pm
ge0r: I tried pinging IP instead of fqdn, still manages to fail some times.

ge0r:
Closing PR. I will work on a new solution that will address directly the return-code-not-captured issue.

PR #16024, opened 2022-12-02
x11/remote_desktop: use assert_and_click to close vncviewer

zhaocongjia:
In recent build, vncviewer is updated to version 1.12.0, some key shortcuts are changed.  So to make the test more robust and compatible with both old and new version, it is better to use "assert_and_click" instead of "send_key" to close the vncviewer.

- Related ticket: https://progress.opensuse.org/issues/121297
- Needles: already added
- Verification run: 
https://openqa.suse.de/tests/10070042#step/onetime_vncsession_xvnc_tigervnc/30
https://openqa.suse.de/tests/10070045#step/persistent_vncsession_xvnc/24
opensuse does not contain such tests.


GraceWang571 (approved):
LGTM

PR #15978, opened 2022-11-25
Skip 'extra_tests_phub' test if packagehub is not ready

rfan1:
- Related ticket:https://progress.opensuse.org/issues/120879
- Verification run: https://openqa.suse.de/tests/10041939
[Please ignore the failed cases, they are known issues]

PR #16027, opened 2022-12-02
Move django to phub tests

BillAnastasiadis:
python3-django and various dependencies are under packagehub, so this pr moves the django test to packagehub and removes it from 15sp5 webserver tests. It also adds the `is_phub_ready` check to  the django test, which is introduced [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/15978).

- Related ticket: https://progress.opensuse.org/issues/120369
- Verification run: -

### DIFF
--- a/schedule/qam/common/mau-webserver.yaml
+++ b/schedule/qam/common/mau-webserver.yaml
@@ -53,7 +53,6 @@ conditional_schedule:
         - console/django
         - console/nginx
       15-SP5:
-        - console/django
         - console/nginx
       tumbleweed:
         - console/django
--- a/schedule/functional/extra_tests_textmode_phub.yaml
+++ b/schedule/functional/extra_tests_textmode_phub.yaml
@@ -8,6 +8,7 @@ schedule:
     - console/add_phub_extension
     - console/python_scientific
     - console/python_flake8
+    - console/django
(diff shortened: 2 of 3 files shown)


schedule/qam/common/mau-webserver.yaml
```diff
@@ -53,7 +53,6 @@ conditional_schedule:
         - console/django
         - console/nginx
       15-SP5:
-        - console/django
```
foursixnine, line 56: does this mean it is unscheduled from mau-webserver?

foursixnine (approved):

schedule/qam/common/mau-webserver.yaml
BillAnastasiadis: I think it's fine to do this like it is for now, but if @foursixnine agrees we could do that

BillAnastasiadis: I think it's fine to do this like it is for now, but if @foursixnine agrees we could do that

foursixnine: @BillAnastasiadis, @rfan1 I trust the team here :), on the bright side is a single schedule to mantain if we also use it for TW and Leap

foursixnine (approved):

PR #16029, opened 2022-12-02
Add record-soft-failure and re-try if loading initrd error occurs on pvm

Zaoliang:
we have sporadic netwwork issue on power machine 'grenache', mnt directory cannot be accessed,
see poo#120570.
reset lpar netboot depends on variuos errors and re-try to load kernel for booting up
VRs:
https://openqa.suse.de/tests/10159676
https://openqa.suse.de/tests/10164018#step/bootloader/25 (softfail)

### DIFF
--- a/lib/bootloader_pvm.pm
+++ b/lib/bootloader_pvm.pm
@@ -65,7 +65,8 @@ sub get_into_net_boot {
     assert_screen 'pvm-bootmenu-boot-exit';
 
     enter_cmd "1";
-    assert_screen ["pvm-grub", "novalink-failed-first-boot"];
+    # download file from tftp server takes sometimes longer, so test could die before matching above needles
+    assert_screen ["pvm-grub", "novalink-failed-first-boot"], 90;
 }
 
 =head2 prepare_pvm_installation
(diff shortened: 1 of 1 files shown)


lib/bootloader_pvm.pm
```diff
@@ -112,13 +100,43 @@ sub prepare_pvm_installation {
...
+    }
+    enter_netboot_parameters;
+    # we have sporadic netwwork issue on power machine 'grenache', mnt directory cannot be accessed, see poo#120570
+    # record_soft_failure if error occurs while loading initrd
+    assert_screen([qw(pvm-grub-command-line-fresh-prompt load-initrd-error)], 180, no_wait => 1);    # initrd is downloaded while waiting
+    if (match_has_tag 'load-initrd-error') {
+        record_info 'cannot load initrd, sporadic issue, see poo#120570. Re-trying now...';
+        enter_cmd_slow "exit";
+        enter_cmd_slow "reset-all";
+        send_key '1';
```
foursixnine, line 129: Maybe you also want to `assert_screen "pvm-firmware-prompt"` and even better, since the steps seem to be the same, you could make a new function, and use it in both places, since the code is (almost) the same

foursixnine (commented):
Great catch!

lib/bootloader_pvm.pm
Zaoliang: yes, I will change this now, thanks!

foursixnine:
Thanks a lot @Zaoliang!

Zaoliang:
> Thanks a lot @Zaoliang!

thanks for review and suggestion! :)

PR #16031, opened 2022-12-02
Fix usage of postgres user for pg_ctl commands

michaelgrifalconi:
Fix usage of postgres user for pg_ctl commands

Before postgres15, there was no error on running psql commands as root. 
Now it started to complain and showed an issue on the switch to the postgres user already in place. The user did not have a shell so it  would fallback to root without error.

- Related ticket: https://progress.opensuse.org/issues/121216
- Needles: none
- Verification run: 
  - 15 SP3 https://openqa.suse.de/tests/10071697 full schedule https://openqa.suse.de/tests/10071701
  - 15 SP2 https://openqa.suse.de/tests/10071698
  - 15 GA  https://openqa.suse.de/tests/10071699
  - 12 SP4 https://openqa.suse.de/tests/10071700

### DIFF
--- a/lib/apachetest.pm
+++ b/lib/apachetest.pm
@@ -237,6 +237,7 @@ Set up a postgres database and configure for:
 sub test_pgsql {
     # configuration so that PHP can access PostgreSQL
     # setup password
+    assert_script_run 'pushd /tmp';
     enter_cmd "sudo -u postgres psql postgres";
     wait_still_screen(1);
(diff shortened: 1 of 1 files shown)


foursixnine:
Can you give it a try for Tumbleweed too?

michaelgrifalconi:
here it is @foursixnine https://openqa.opensuse.org/tests/2922383# (mysql was already failing)

PR #16032, opened 2022-12-02
Unify schedules for Micro and ALP products

jlausuch:
The following products share similar schedules:
- MicroOS
- SLE Micro
- Leap Micro
- ALP

However, there is a mix between yaml schedules and main.pm files.

The idea of this commit is to unify all the schedules into a single
file with the needed conditions and get rid of all the yaml schedules.

Related ticket: https://progress.opensuse.org/issues/120928
Verification runs:
- [MicroOS](https://openqa.opensuse.org/tests/overview?version=Tumbleweed&distri=microos&build=jlausuch%2Fos-autoinst-distri-opensuse%23schedule_micro_alp) 
- [ALP](https://openqa.opensuse.org/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23schedule_micro_alp&version=0.1&distri=alp)
- [SLE Micro](https://openqa.suse.de/tests/overview?build=jlausuch%2Fos-autoinst-distri-opensuse%23schedule_micro_alp&version=5.4&distri=sle-micro)

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -1,0 +1,276 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: module loader for MicroOS, SLE Micro, Leap Micro and ALP.
+# Maintainer: qa-c@suse.de
+
+package main_micro_alp;
+use strict;
+use warnings;
+use base 'Exporter';
+use Exporter;
+use main_common;
+use main_containers qw(load_container_tests is_container_test);
+use testapi qw(check_var get_required_var get_var set_var);
+use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp is_staging is_released is_transactional is_rt);
+use utils;
+use Utils::Architectures;
+use Utils::Backends;
+
+sub is_image {
+    return get_required_var('FLAVOR') =~ /image|default|kvm/i;
+}
+
+sub is_dvd {
+    return get_required_var('FLAVOR') =~ /dvd/i;
+}
+
+sub is_regproxy_required {
+    # For now only the Kubic kubeadm test needs a registry proxy.
+    # docker_image and podman_image pull with the full path themselves.
+    return check_var('SYSTEM_ROLE', 'kubeadm');
+}
+
+sub load_config_tests {
+    if (is_microos) {
+        loadtest 'update/zypper_clear_repos';
+        loadtest 'console/zypper_ar';
+        loadtest 'console/zypper_ref';
+    }
+    loadtest 'transactional/tdup' if get_var('TDUP');
+    loadtest 'transactional/host_config' unless is_dvd;
+    loadtest 'rt/rt_is_realtime' if is_rt;
+    loadtest 'transactional/enable_selinux' if (get_var('ENABLE_SELINUX') && is_image);
+    loadtest 'console/suseconnect_scc' if (is_sle_micro && get_var('SCC_REGISTER') && !is_dvd);
+    loadtest 'transactional/install_updates' if (is_sle_micro && is_released);
+}
+
+sub load_boot_from_disk_tests {
+    if (is_s390x()) {
+        loadtest 'installation/bootloader_start';
+        loadtest 'boot/boot_to_desktop';
+    } else {
+        if (check_var('FIRST_BOOT_CONFIG', 'wizard')) {
+            loadtest 'jeos/firstrun';
+        } else {
+            loadtest 'microos/disk_boot';
+        }
+    }
+    loadtest 'installation/system_workarounds' if (is_aarch64 && is_microos);
+    replace_opensuse_repos_tests if is_repo_replacement_required;
+}
+
+sub load_boot_from_dvd_tests {
+    if (is_s390x) {
+        loadtest 'installation/bootloader_start';
+    } elsif (get_var('UEFI')) {
+        loadtest 'installation/bootloader_uefi';
+    } else {
+        loadtest 'installation/bootloader';
+    }
+}
+
+sub load_installation_tests {
+    if (check_var('HDDSIZEGB', '10')) {
+        # boo#load_installation_tests
+        # undefined method "safe_copy" for nil:NilClass
+        # YaST2 crashes if disk is too small for a viable proposal
+        loadtest 'installation/welcome';
+        loadtest 'installation/installation_mode';
+        loadtest 'installation/logpackages';
+        loadtest 'installation/system_role';
+        loadtest 'installation/ntp_config_settings';
+        loadtest 'installation/user_settings_root';
+        loadtest 'installation/installation_overview';
+    }
+    else {
+        loadtest 'installation/welcome';
+        loadtest 'installation/disk_activation' if is_zvm;
+        loadtest 'installation/scc_registration' if is_sle_micro;
+        if (is_microos) {
+            loadtest 'installation/online_repos';
+            loadtest 'installation/installation_mode';
+            loadtest 'installation/logpackages';
+            loadtest 'installation/system_role';
+        }
+
+        loadtest 'installation/ntp_config_settings';
+        loadtest 'installation/user_settings_root';
+        loadtest 'installation/resolve_dependency_issues';
+        loadtest 'installation/select_patterns' if get_var('PATTERNS');
+        loadtest 'installation/installation_overview';
+        loadtest 'installation/disable_grub_timeout';
+        loadtest 'installation/enable_selinux' if get_var('ENABLE_SELINUX');
+        loadtest 'installation/start_install';
+        loadtest 'installation/await_install';
+        loadtest 'installation/logs_from_installation_system';
+        loadtest 'installation/reboot_after_installation';
+        if (is_s390x) {
+            loadtest 'boot/reconnect_mgmt_console';
+            loadtest 'installation/first_boot';
+        } else {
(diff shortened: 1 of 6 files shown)


mimi1vx (approved):

lib/main_transactional.pm
```diff
@@ -0,0 +1,283 @@
...
+    }
+    loadtest 'installation/system_workarounds' if (is_aarch64 && is_microos);
+    replace_opensuse_repos_tests if is_repo_replacement_required;
+}
+
+sub load_boot_from_dvd_tests {
+    if (is_s390x) {
+        loadtest 'installation/bootloader_start';
+    } elsif (get_var("UEFI")) {
+        loadtest 'installation/bootloader_uefi';
```
mloviska, line 77: I would rather prefer to make this one default. For instance, it can handle `EXTRABOOTPARAMS` which can be useful to inject some grub parameters. And the module works fine for x86_64 uefi, bios or aarch64

lib/main_transactional.pm
```diff
@@ -0,0 +1,283 @@
...
+    loadtest 'microos/libzypp_config';
+    loadtest 'microos/image_checks' if is_image;
+    loadtest 'microos/one_line_checks';
+    loadtest 'microos/services_enabled';
+    # MicroOS -old images use wicked, but cockpit-wicked is no longer supported in TW
+    loadtest 'microos/cockpit_service' unless is_staging || (is_microos('Tumbleweed') && get_var('HDD_1') =~ /-old/);
+    # Staging has no access to repos and the MicroOS-DVD does not contain ansible
+    # Ansible test needs Packagehub in SLE and it can't be enabled in SLEM
+    loadtest 'console/ansible' unless (is_staging || is_sle_micro);
+    loadtest 'console/kubeadm' if (check_var('SYSTEM_ROLE', 'kubeadm'));
```
mloviska, line 190: Do we run this test somewhere? I thought it is intended for Kubic or something that runs k8s.

lib/main_transactional.pm
```diff
@@ -0,0 +1,283 @@
...
+    loadtest 'transactional/transactional_update';
+    loadtest 'transactional/rebootmgr';
+    loadtest 'transactional/health_check';
+}
+
+
+sub load_network_tests {
+    loadtest 'microos/networking';
+    loadtest 'microos/libzypp_config';
+    loadtest 'console/firewalld';
```
mloviska, line 206: sle-micro does not come with configured firewall, I think the same applies for ALP. Doesn't it fail? I haven't noticed any enclosed VR with this module

jlausuch: You are right, but this section is only executed in ALP tests.

lib/main_transactional.pm
jlausuch: I wasn't sure if it's executed somewhere in O3, in some dev group or something, so I left it there just in case as it was in main.pm of `products/microos/main.pm`.

lib/main_transactional.pm
mloviska: We do it in JeOS and it works fine. :)

lib/main_transactional.pm
mloviska: Ah, now I see it there. Ok

mloviska: But maybe a comment would be nice here, as I guess there will be changes regarding firewalld in ALP.

lib/main_transactional.pm
mloviska: Yeah, I have not found it myself. Let's see whether someone would have more insight. Otherwise, I would vote for removal as now it is a good opportunity to also clean up the schedule a bit.

lib/main_transactional.pm
```diff
@@ -0,0 +1,283 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: module loader for Transactional product family:
```
Vogtinator, line 6: A bit misleading because it doesn't cover transactional server...

Calling it `lib/main_micro.pm` is a bit misleading as well though.

grisu48:
```suggestion
# Summary: module loader for various transactional products:
```

Perhaps?

products/sle-micro/main.pm
```diff
@@ -74,15 +56,6 @@ if (is_updates_test_repo && !get_var('MAINT_TEST_REPO')) {
...
-if (is_container_test) {
-    load_boot_from_disk_tests();
-    load_container_tests();
-}
-
-if (is_kernel_test()) {
-    load_kernel_tests();
-    return 1;
-}
+load_tests;
```
grisu48, line 59:
```suggestion
load_tests; # see main_transactional.pm
```

Let's be nice to our future selves - I would do the same in the other `.pm` files.

grisu48 (approved):
It's difficult to see what the outcome is, but I cannot find any obvious mistakes, so LGTM!

Two suggestions though.

lib/main_transactional.pm
jlausuch: Yes, both can be misleading.. `main_micro_alp.pm` ?

products/sle-micro/main.pm
jlausuch: I actually tried with `main_transactional::load_tests` but for some reason it didn't work...

mloviska: > I actually tried with `main_transactional::load_tests` but for some reason it didn't work...

I was about to suggest that, meh

lib/main_transactional.pm
```diff
@@ -0,0 +1,283 @@
...
+    loadtest 'console/suseconnect_scc' if (is_sle_micro && get_var('SCC_REGISTER') && !is_dvd);
+    loadtest 'transactional/install_updates' if (is_sle_micro && is_released);
+}
+
+sub load_boot_from_disk_tests {
+    if (is_s390x) {
+        loadtest 'installation/bootloader_start';
+        loadtest 'boot/boot_to_desktop';
+    } else {
+        if (check_var('FIRST_BOOT_CONFIG', 'wizard')) {
```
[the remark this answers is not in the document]
jlausuch: why? we need the jeos first boot loader here

products/sle-micro/main.pm
jlausuch: https://openqa.opensuse.org/tests/2928784
`Undefined subroutine &main_micro_alp::load_tests called at os-autoinst-distri-opensuse/products/alp/main.pm line 20.`

lib/main_transactional.pm
jlausuch: if you don't mind, I'll leave it here, otherwise I need to remove `tests/console/kubeadm.pm` to get green pipeline:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/actions/runs/3619223707/jobs/6100012477

products/sle-micro/main.pm
jlausuch: Done. The reason was obvious, it's not in the same directory, but if I import it, I can use `main_transactional::load_tests` and no need for comment.

lib/main_transactional.pm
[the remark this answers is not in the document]
mloviska: yes ofc, it should run before jeos-firstboot. The reason is that sometimes there are problems with jeos-firstboot as https://openqa.suse.de/tests/10081900#step/firstrun/6.

It would be nice to have the option to handle EXTRABOOTPARAMS in case it is needed. I am testing it now, but apparently it needs a bit of code modifications. Hence, we can leave it for now and do it in separate PR

lib/main_transactional.pm
jlausuch: bad idea :(
https://openqa.suse.de/tests/10082654

jlausuch:
Can we move on with this?

lib/main_micro_alp.pm
```diff
@@ -0,0 +1,280 @@
...
+use main_containers qw(load_container_tests is_container_test);
+use testapi qw(check_var get_required_var get_var set_var);
+use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp is_staging is_released is_transactional is_rt);
+use utils;
+use Utils::Architectures;
+use Utils::Backends;
+
+our @EXPORT = qw(
+  load_tests
+);
```
foursixnine, line 24: Since you're doing static calls, you can even remove this, [along with all the exporter stuff](https://gist.github.com/foursixnine/5f001c5f924e077a713383cfdd9dd0ba) and it will still work (don't believe me 100% though)

jlausuch: it works, thanks!

foursixnine (approved):

jlausuch:
@mloviska @Vogtinator ok to merge?

mloviska:
> @mloviska @Vogtinator ok to merge?

No objections, I thought you wish to solve the ansible problem within this PR though

jlausuch:
> > @mloviska @Vogtinator ok to merge?
> 
> No objections, I thought you wish to solve the ansible problem within this PR though

I left it as it was scheduled before. Current condition is
`loadtest 'console/ansible' unless (is_staging || is_sle_micro);`

Let's investigate and discuss further conditions to run it, but I think it's ok for another PR.

mloviska (approved):

jlausuch:
Merging. If there is more feedback it can be handled in new PR.

DimStar77:
https://openqa.opensuse.org/tests/2949474#

Looks like this PR changed the schedule, which makes the tests break

PR #16020, opened 2022-12-02
Run YaST with ncurses in yast2_migration module in PowerVM

badboywj:
make sure run yast2_migration with ncurses in PowerVM

- Related ticket: https://progress.opensuse.org/issues/119950
- Verification run: https://openqa.suse.de/tests/10070229#step/yast2_migration/2, https://openqa.suse.de/tests/10070231#step/yast2_migration/1

PR #16035, opened 2022-12-05
Run yast ui test with ncurse for pvm backend

coolgw:
Run yast ui test with ncurse for pvm backend

- Related ticket: https://progress.opensuse.org/issues/110779
- Needles: na
- Verification run: https://openqa.suse.de/tests/10081723

### DIFF
--- a/lib/nfs_common.pm
+++ b/lib/nfs_common.pm
@@ -7,6 +7,7 @@ use utils qw(systemctl file_content_replace clear_console zypper_call clear_cons
 use Utils::Systemd 'disable_and_stop_service';
 use mm_network;
 use version_utils;
+use Utils::Backends 'is_pvm_hmc';
 
 
 our @ISA = qw(Exporter);
@@ -214,8 +215,8 @@ sub config_service {
     try_nfsv2();
 
     prepare_exports($rw, $ro);
-
-    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'nfs-server');
+    my $y2_opts = is_pvm_hmc() ? "--ncurses" : "";
+    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'nfs-server', yast2_opts => $y2_opts);
 
     yast2_server_initial();
 
--- a/lib/y2lan_restart_common.pm
+++ b/lib/y2lan_restart_common.pm
@@ -20,6 +20,7 @@ use Exporter 'import';
 use testapi;
 use utils 'systemctl';
 use version_utils qw(is_sle is_leap);
+use Utils::Backends 'is_pvm_hmc';
 use y2_module_basetest qw(accept_warning_network_manager_default is_network_manager_default);
 use y2_module_consoletest;
 use Test::Assert ':all';
@@ -398,7 +399,9 @@ sub change_ipforward {
 
 =head2 open_yast2_lan
 
- open_yast2_lan();
+ open_yast2_lan([$ui]);
+
+C<$ui> ncurses/qt will used for set yast ui mode
 
 Open yast2 lan, run handle_dhcp_popup() and install and check firewalld
 
@@ -407,9 +410,16 @@ If network is controlled by Networkmanager, don't change any network settings.
 =cut
 
 sub open_yast2_lan {
-    my $is_nm = !script_run('systemctl is-active NetworkManager');    # Revert boolean because of bash vs perl's return code.
+    my %options = @_;
+    my $is_nm
+      = !script_run('systemctl is-active NetworkManager');    # Revert boolean because of bash vs perl's return code.
+    my $y2_opts
+      = ($options{ui} =~ /ncurses/)
+      ? "--" . $options{ui}
+      : "";
+    $y2_opts = "--ncurses" if is_pvm_hmc();
 
-    $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'lan');
+    $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'lan', yast2_opts => $y2_opts);
 
     if ($is_nm) {
         handle_Networkmanager_controlled;    # don't change any settings
--- a/tests/console/yast2_bootloader.pm
+++ b/tests/console/yast2_bootloader.pm
@@ -18,7 +18,7 @@ use warnings;
 use testapi;
 use Utils::Architectures;
 use utils;
-use Utils::Backends 'is_hyperv';
+use Utils::Backends qw(is_hyperv is_pvm_hmc);
 
 sub run {
     my $self = shift;
@@ -26,8 +26,8 @@ sub run {
 
     # make sure yast2 bootloader module is installed
     zypper_call 'in yast2-bootloader';
-
-    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'bootloader');
+    my $y2_opts = is_pvm_hmc() ? "--ncurses" : "";
+    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'bootloader', yast2_opts => $y2_opts);
 
     # YaST2 prompts user to install missing packages found during storage probing.
     # Otherwise YaST2 shows bootloader settings options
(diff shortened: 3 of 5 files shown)


lib/y2lan_restart_common.pm
```diff
@@ -407,9 +408,16 @@ If network is controlled by Networkmanager, don't change any network settings.
 =cut
 
 sub open_yast2_lan {
-    my $is_nm = !script_run('systemctl is-active NetworkManager');    # Revert boolean because of bash vs perl's return code.
-
-    $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'lan');
+    my %options = @_;
```
lemon-suse, line 411: Why need the parameter for open_yast2_lan? Just judge the PowerVM inside the function and set correct $y2_opts is ok.

jknphy:
I think it doesn't make sense to do this hack each time before calling the function, could we consider to do it inside the function?

lib/nfs_common.pm
```diff
@@ -214,8 +215,8 @@ sub config_service {
     try_nfsv2();
 
     prepare_exports($rw, $ro);
-
-    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'nfs-server');
+    my $y2_opts = is_pvm_hmc() ? "--ncurses" : "";
```
badboywj, line 218: I think it is more simple without "()"

lib/y2lan_restart_common.pm
badboywj: I suggest to add some comments about how to use this function because of the new argument, this make users clear.

lib/y2lan_restart_common.pm
```diff
@@ -407,9 +408,16 @@ If network is controlled by Networkmanager, don't change any network settings.
...
+    my $is_nm
+      = !script_run('systemctl is-active NetworkManager');    # Revert boolean because of bash vs perl's return code.
+    my $y2_opts
+      = ($options{ui} =~ /ncurses/)
+      ? "--" . $options{ui}
+      : "";
+    $y2_opts = "--ncurses" if is_pvm_hmc();
```
badboywj, line 418: same to the above, maybe it is better without "()"

tests/console/yast2_bootloader.pm
```diff
@@ -18,16 +18,16 @@ use warnings;
...
 
 sub run {
     my $self = shift;
     select_console 'root-console';
 
     # make sure yast2 bootloader module is installed
     zypper_call 'in yast2-bootloader';
-
-    my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'bootloader');
+    my $y2_opts = is_pvm_hmc() ? "--ncurses": "";
```
badboywj, line 29: same to the above, maybe it is better without "()"

tests/console/yast2_lan.pm
```diff
@@ -38,7 +39,8 @@ sub run {
     script_run('ls -alF /etc/sysconfig/network/');
     save_screenshot;
 
-    my $opened = open_yast2_lan();
+    my $y2_opts = is_pvm_hmc() ? "ncurses": "";
```
badboywj, line 42: same to the above, maybe it is better without "()"

badboywj: and this line needs tidying

lib/nfs_common.pm
coolgw: This is on purpose, it will show clearly function call.

lib/y2lan_restart_common.pm
coolgw: 1.open_yast2_lan already used in many place in current code, so i prefer extend parameter have ability to handle future change.

2.Comments already added for new argument.

lib/y2lan_restart_common.pm
coolgw: see above feedback.

tests/console/yast2_lan.pm
coolgw: tidy fixed, thanks!

coolgw:
> > I think it doesn't make sense to do this hack each time before calling the function, could we consider to do it inside the function?
> 
> @coolgw could be this code inside the method `yast2_console_exec` and not all the time outside or is there any troubles with that? if it simplifies the code we could borrow this: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16020/files#diff-283b909b6e798ec5f92a947c75b4e9ef2cc6805d36e0cdcfadab2b8fbd0a80a0R155

yast2_console_exec is common function call and it's original design contain a %args can be used for flag ui mode, i do not want yast2_console_exec function contain a hidden hard code logic such as put ncurse mod if is_pvm.  I think my implementation has more flexibility and fit original design, the special logic should put on test level not within function level, function level logic should base on parameter.
@ge0r   @rakoenig what's your opinion ?

jknphy:
> > > I think it doesn't make sense to do this hack each time before calling the function, could we consider to do it inside the function?
> > 
> > 
> > @coolgw could be this code inside the method `yast2_console_exec` and not all the time outside or is there any troubles with that? if it simplifies the code we could borrow this: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16020/files#diff-283b909b6e798ec5f92a947c75b4e9ef2cc6805d36e0cdcfadab2b8fbd0a80a0R155
> 
> yast2_console_exec is common function call and it's original design contain a %args can be used for flag ui mode, i do not want yast2_console_exec function contain a hidden hard code logic such as put ncurse mod if is_pvm. I think my implementation has more flexibility and fit original design, the special logic should put on test level not within function level, function level logic should base on parameter. @ge0r @rakoenig what's your opinion ?

it is a hack, because we cannot test PowerVM in a different way, so single point of hack it is better than a hack all over the place, and if someone will face this in the future, needs to put this code again or ask. What I'm not sure if is_pvm would be enough, we need to check if we put inside that is the `pvm_hmc` one, because for `spvm` backend it might be possible to run some desktop stuff, I don't know that, we barely use it, only Kernel squad, I didn't see any desktop in the installed system.

jknphy:
> > > > > I think it doesn't make sense to do this hack each time before calling the function, could we consider to do it inside the function?
> > > > 
> > > > 
> > > > @coolgw could be this code inside the method `yast2_console_exec` and not all the time outside or is there any troubles with that? if it simplifies the code we could borrow this: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16020/files#diff-283b909b6e798ec5f92a947c75b4e9ef2cc6805d36e0cdcfadab2b8fbd0a80a0R155
> > > 
> > > 
> > > yast2_console_exec is common function call and it's original design contain a %args can be used for flag ui mode, i do not want yast2_console_exec function contain a hidden hard code logic such as put ncurse mod if is_pvm. I think my implementation has more flexibility and fit original design, the special logic should put on test level not within function level, function level logic should base on parameter. @ge0r @rakoenig what's your opinion ?
> > 
> > 
> > it is a hack, because we cannot test PowerVM in a different way, so single point of hack it is better than a hack all over the place, and if someone will face this in the future, needs to put this code again or ask. What I'm not sure if is_pvm would be enough, we need to check if we put inside that is the `pvm_hmc` one, because for `spvm` backend it might be possible to run some desktop stuff, I don't know that, we barely use it, only Kernel squad, I didn't see any desktop in the installed system. Another point is that the fact that the method has some parameters, doesn't mean that you can give a default ones given some special conditions, like it is this case. For the name of the function you always expect ncurses because you are running in the console, but unless you want to do ssh-x in PowerVM you should set a proper default.
> 
> I do not think add extra parameter for yast2_console_exec is a hack, it's normal way to extend function and make it start support more feature, so no hack all over the place, it should give correct parameter for function if needed. In future if you need add more backend or change backend then you need modify on test module level, not function level, the low api test function level should touch as less as possible. I do not like some hidden logic under the function api, and add condition is_xxx is very popular in os-autoinst-distri-opensuse test module level. Further more, in function level i do not add any default one or give some special conditions, default logic is as same as before and new logic ONLY happen if you add new parameter, i do not call that special conditions, it's new logic support for new parameter.

I'm referring that is a hack the fact that we need to adapt the test with that condition, because that env actually has desktop but we don't test it like that, so it is somehow our workaround. I'm fine if you prefer to do it like that, anyway these are legacy methods.

PR #16038, opened 2022-12-05
Simplify stunnel test 

ilmanzo:
Since real test target is stunnel, with the upcoming wayland default desktop vnc is a bit too shaky choice; so we'd replace vnc with a netcat message exchange from client to server

- Related ticket: https://progress.opensuse.org/issues/121030
- 15SP4 Verification runs:
    https://openqa.suse.de/tests/10082094           
    https://openqa.suse.de/tests/10082095              
- 15SP5:
    https://openqa.suse.de/tests/10082119          
    https://openqa.suse.de/tests/10082120

### DIFF
--- a/tests/fips/stunnel.pm
+++ b/tests/fips/stunnel.pm
@@ -15,12 +15,13 @@ use lockapi;
 use utils;
 use mm_tests;
 use mmapi 'wait_for_children';
+use version_utils 'package_version_cmp';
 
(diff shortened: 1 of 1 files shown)


paolostivanin:
LGTM

punkioudi (approved):
I think they are missing from the VRs ker_mode from 15-SP4 and env_mode from 15-SP5. Could you also add them? Other than that, lgtm!

PR #16041, opened 2022-12-05
Fail when 'az vm create' command fails

pdostal:
- Verification run: [azure_cli@64bit](https://pdostal-server.suse.cz/tests/3417)

### DIFF
--- a/tests/publiccloud/azure_cli.pm
+++ b/tests/publiccloud/azure_cli.pm
@@ -54,7 +54,7 @@ sub run {
     # VM creation
     my $vm_create = "az vm create --resource-group $resource_group --name $machine_name --public-ip-sku Standard --tags '$tags'";
     $vm_create .= " --image $image_name --size Standard_B1ms --admin-username azureuser --ssh-key-values ~/.ssh/id_rsa.pub";
-    my $output = script_output($vm_create, timeout => 600, proceed_on_failure => 1);
+    my $output = script_output($vm_create, timeout => 600);
     if ($output =~ /ValidationError.*object has no attribute/) {
         record_soft_failure('bsc#1191482 - Failed to start/stop vms with azure cli');
         return;


tests/publiccloud/azure_cli.pm
```diff
@@ -54,10 +54,14 @@ sub run {
     # VM creation
     my $vm_create = "az vm create --resource-group $resource_group --name $machine_name --public-ip-sku Standard --tags '$tags'";
     $vm_create .= " --image $image_name --size Standard_B1ms --admin-username azureuser --ssh-key-values ~/.ssh/id_rsa.pub";
-    my $output = script_output($vm_create, timeout => 600, proceed_on_failure => 1);
-    if ($output =~ /ValidationError.*object has no attribute/) {
-        record_soft_failure('bsc#1191482 - Failed to start/stop vms with azure cli');
-        return;
+    if (script_run($vm_create) != 0, timeout => 600, die_on_timeout => 1) {
```
grisu48, line 57: Why not simply use `assert_script_run`? :thinking: Would save us this if block entirely with the same effect.

asmorodskyi: agree with Felix

tests/publiccloud/azure_cli.pm
```diff
@@ -54,11 +54,7 @@ sub run {
     # VM creation
     my $vm_create = "az vm create --resource-group $resource_group --name $machine_name --public-ip-sku Standard --tags '$tags'";
     $vm_create .= " --image $image_name --size Standard_B1ms --admin-username azureuser --ssh-key-values ~/.ssh/id_rsa.pub";
-    my $output = script_output($vm_create, timeout => 600, proceed_on_failure => 1);
-    if ($output =~ /ValidationError.*object has no attribute/) {
-        record_soft_failure('bsc#1191482 - Failed to start/stop vms with azure cli');
-        return;
-    }
+    assert_script_run($vm_create, timeout => 600);
```
grisu48, line 57: Sorry, but I would have even a better idea: Why not simply remove the `proceed_on_failure => 1`? Then the next check can also remain intact.

```suggestion
    my $output = script_output($vm_create, timeout => 600);
```

grisu48 (approved):

PR #16043, opened 2022-12-05
Verify secure boot successfully in kernel v6.0

ge0r:
In kernel v6.0, the /sys/firmware/efi/vars interface has been removed.
Instead we are to use the contents of /sys/firmware/efi/efivars to verify whether secure boot has been enabled or not.

- Related ticket: https://progress.opensuse.org/issues/120100
- Needles: No needles
- Verification run: https://openqa.suse.de/tests/10171482#details

### DIFF
--- a/tests/console/verify_secure_boot.pm
+++ b/tests/console/verify_secure_boot.pm
@@ -14,16 +14,24 @@ use testapi;
 use scheduler 'get_test_suite_data';
 use Test::Assert ':all';
 
+# The minimum number of octal digits in the efivars file should be 5,
+# because the fifth octal digit shows whether Secure Boot is enabled (1) or disabled (0)
+use constant MIN_OCTALS => 5;
+
 sub run {
     my $test_data = get_test_suite_data();
+
     select_console 'root-console';
-    record_info("Check file", "Check if file /sys/firmware/efi/vars/SecureBoot-*/data exists");
-    assert_script_run("ls /sys/firmware/efi/vars/SecureBoot-*/data");
+    record_info("Check file", "Check if file /sys/firmware/efi/efivars/SecureBoot-* exists");
+    # From v6.0 kernel and onwards, the SecureBoot file resides in /sys/firmware/efi/efivars/
+    assert_script_run("ls /sys/firmware/efi/efivars/SecureBoot-*");
     record_info("Check secure boot", "Check if secure boot option is set as expected");
-    my $secure_boot = script_output("od -An -t u1 /sys/firmware/efi/vars/SecureBoot-*/data") ? 'enabled' : 'disabled';
+    my $octal_str = script_output("od -An -t u1 /sys/firmware/efi/efivars/SecureBoot-*");
+    my @octal_array = split(/\s+/, $octal_str);
+
+    die 'Unexpected values in Secure Boot file' unless 0 + @octal_array >= MIN_OCTALS && $octal_array[MIN_OCTALS - 1] =~ /0|1/;
+    my $secure_boot = $octal_array[MIN_OCTALS - 1] ? 'enabled' : 'disabled';
     assert_equals($test_data->{secure_boot}, $secure_boot, "The secure boot option is not $test_data->{secure_boot}");
 }
 
 1;
-
-


jknphy:
The idea to use the file itself instead of external tool I think it is a good one, it is a more basic check, the only problem is that we need more low-level code not so easy to maintain for most of the cases, but I would keep it for this case.
I would like to hear from other reviewers:
- suggestion about how to avoid separate regex and split by regex directly.
- Unify existing assertion with new introduced `die`
- How to remove that if/else and make it more elegant.

ge0r:
> The idea to use the file itself instead of external tool I think it is a good one, it is a more basic check, the only problem is that we need more low-level code not so easy to maintain for most of the cases, but I would keep it for this case. I would like to hear from other reviewers:
> 
>     * suggestion about how to avoid separate regex and split by regex directly.
Why would you we want to split by regex directly? This would mean a more cluttered line and it would be not so easy to determine which was the point of failure.

>     * Unify existing assertion with new introduced `die`
I have a different opinion here as well. The assertion is related to whether our test has passed or failed, based on our expected result. However the die is related to the mechanism to extract the data to come to that result. When test execution reaches that die it means that something went wrong with the logic or setup of the test, not the expected result. Mixing them both in a single assert would, in my opinion, make a possible failure harder to debug.

coolgw:
If mokutil --sb-state can directly show the status of efi then i prefer use this command, more simple.

man page of mokutil
       --sb-state
              Show SecureBoot State

jknphy:
> > The idea to use the file itself instead of external tool I think it is a good one, it is a more basic check, the only problem is that we need more low-level code not so easy to maintain for most of the cases, but I would keep it for this case. I would like to hear from other reviewers:
> > ```
> > * suggestion about how to avoid separate regex and split by regex directly.
> > ```
> 
> Why would you we want to split by regex directly? This would mean a more cluttered line and it would be not so easy to determine which was the point of failure.
> 
> > ```
> > * Unify existing assertion with new introduced `die`
> > ```
> 
> I have a different opinion here as well. The assertion is related to whether our test has passed or failed, based on our expected result. However the die is related to the mechanism to extract the data to come to that result. When test execution reaches that die it means that something went wrong with the logic or setup of the test, not the expected result. Mixing them both in a single assert would, in my opinion, make a possible failure harder to debug.

(1)  we are removing several spaces and converting them in one and then split them, right? we could just separate by a multiple space regex in one action, something like `@octal_digits = split(/ +/, $octal);` what value adds to have two lines or am I missing some nuance here?
(2) if it is not 1 or 0 the parsing might have work and have value 6, we actually could have an index exception in that case, that is another reason to avoid this manipulations for the good of the maintenance of the code. What you might want to do is something like:
```
die 'parsing error 5th element blbalbl unless $octals[4];
die 'parsing values not in range' unless $octals[4] =~ /1|2/;
```
That this the thing, if you make the code a bit better I like the solution to check the most native thing the file, but if keep it like that I would prefer to use by far `mokutil --sb-state` which is very simple and recommended in the bug. Anyway we are relying here still in `od`.
(3) is already mentioned with example above, in Perl you do this kind of style of programming, there is not need for some much brackets, many things in-line are really easy to understand instead of if/else/ifelse, etc.

ge0r:
> > > The idea to use the file itself instead of external tool I think it is a good one, it is a more basic check, the only problem is that we need more low-level code not so easy to maintain for most of the cases, but I would keep it for this case. I would like to hear from other reviewers:
> > > ```
> > > * suggestion about how to avoid separate regex and split by regex directly.
> > > ```
> > 
> > 
> > Why would you we want to split by regex directly? This would mean a more cluttered line and it would be not so easy to determine which was the point of failure.
> > > ```
> > > * Unify existing assertion with new introduced `die`
> > > ```
> > 
> > 
> > I have a different opinion here as well. The assertion is related to whether our test has passed or failed, based on our expected result. However the die is related to the mechanism to extract the data to come to that result. When test execution reaches that die it means that something went wrong with the logic or setup of the test, not the expected result. Mixing them both in a single assert would, in my opinion, make a possible failure harder to debug.
> 
> (1) we are removing several spaces and converting them in one and then split them, right? we could just separate by a multiple space regex in one action, something like `@octal_digits = split(/ +/, $octal);` what value adds to have two lines or am I missing some nuance here? (2) if it is not 1 or 0 the parsing might have work and have value 6, we actually could have an index exception in that case, that is another reason to avoid this manipulations for the good of the maintenance of the code. What you might want to do is something like:

> 
> ```
> die 'parsing error 5th element blbalbl unless $octals[4];
> die 'parsing values not in range' unless $octals[4] =~ /1|2/;
> ```
> 
> That this the thing, if you make the code a bit better I like the solution to check the most native thing the file, but if keep it like that I would prefer to use by far `mokutil --sb-state` which is very simple and recommended in the bug. Anyway we are relying here still in `od`. (3) is already mentioned with example above, in Perl you do this kind of style of programming, there is not need for some much brackets, many things in-line are really easy to understand instead of if/else/ifelse, etc.



> > > The idea to use the file itself instead of external tool I think it is a good one, it is a more basic check, the only problem is that we need more low-level code not so easy to maintain for most of the cases, but I would keep it for this case. I would like to hear from other reviewers:
> > > ```
> > > * suggestion about how to avoid separate regex and split by regex directly.
> > > ```
> > 
> > 
> > Why would you we want to split by regex directly? This would mean a more cluttered line and it would be not so easy to determine which was the point of failure.
> > > ```
> > > * Unify existing assertion with new introduced `die`
> > > ```
> > 
> > 
> > I have a different opinion here as well. The assertion is related to whether our test has passed or failed, based on our expected result. However the die is related to the mechanism to extract the data to come to that result. When test execution reaches that die it means that something went wrong with the logic or setup of the test, not the expected result. Mixing them both in a single assert would, in my opinion, make a possible failure harder to debug.
> 
> (1) we are removing several spaces and converting them in one and then split them, right? we could just separate by a multiple space regex in one action, something like `@octal_digits = split(/ +/, $octal);` what value adds to have two lines or am I missing some nuance here?

On this point I agree, it makes sense as a simplification

 (2) if it is not 1 or 0 the parsing might have work and have value 6, we actually could have an index exception in that case, that is another reason to avoid this manipulations for the good of the maintenance of the code. What you might want to do is something like:
> 
> ```
> die 'parsing error 5th element blbalbl unless $octals[4];
> die 'parsing values not in range' unless $octals[4] =~ /1|2/;
> ```

I also agree with this, I was arguing against unifying it with the assert

> That this the thing, if you make the code a bit better I like the solution to check the most native thing the file, but if keep it like that I would prefer to use by far `mokutil --sb-state` which is very simple and recommended in the bug. Anyway we are relying here still in `od`. (3) is already mentioned with example above, in Perl you do this kind of style of programming, there is not need for some much brackets, many things in-line are really easy to understand instead of if/else/ifelse, etc.

I would argue that it is better to avoid using the mokutil, the tool is not available by default and needs to be installed, adding an extra layer of instability

jknphy:
another option to re-work a bit this code is simply check the length and get the last element, so no index, all good :)

jknphy:
yep, I agree to not install stuff, that also breaks now and then.

ge0r:
> If mokutil --sb-state can directly show the status of efi then i prefer use this command, more simple.
> 
> man page of mokutil --sb-state Show SecureBoot State

I think, because mokutil is not part of the system and needs to be installed, that it will make the test less robust to add this dependency for the check

badboywj:
I have a stupid question since I notice that "efivars" derectory replaces "vars" in kernel 6.0 in the description, can the openQA module "verify_secure_boot.pm" run with the kernel version lower than 6? if yes, we need to consider the scenario.

ge0r:
> verify_secure_boot.pm

That is a valid point. But since we are only running this module in the YaST group, on the latest product version, as seen [here](https://openqa.suse.de/tests/overview?modules=verify_secure_boot), there is currently no need to draw the distinction.
Still I think it makes sense to add a comment for clarity. Thanks!

tests/console/verify_secure_boot.pm
```diff
@@ -17,13 +17,17 @@ use Test::Assert ':all';
...
+    record_info("Check file", "Check if file /sys/firmware/efi/efivars/SecureBoot-* exists");
+    # From v6.0 kernel and onwards, the SecureBoot file resides in /sys/firmware/efi/efivars/
+    assert_script_run("ls /sys/firmware/efi/efivars/SecureBoot-*");
     record_info("Check secure boot", "Check if secure boot option is set as expected");
-    my $secure_boot = script_output("od -An -t u1 /sys/firmware/efi/vars/SecureBoot-*/data") ? 'enabled' : 'disabled';
+    my $octal_str = script_output("od -An -t u1 /sys/firmware/efi/efivars/SecureBoot-*");
+
+    # The final octal digit shows whether Secure Boot is enabled (1) or disabled (0)
+    my $secure_boot_octal = substr($octal_str, -1);
+    die 'Unexpected values in Secure Boot file' unless (length($octal_str) == 17) and ($secure_boot_octal =~ /0|1/);
```
jknphy, line 28: About line 27: still we are doing some kind of indexing, we have a library to get the first item in Perl, I don't remember the name right now, probably has the option to get the last too..
About line 28: do we know that 17 is a fixed number, is somewhere documented? perhaps those spaces between the real values can change from 3 to 2 in next version and the validation breaks. We could still split them by regex as mentioned, considering that could be 1 or 3 or 5 spaces let's say if we don't know it is quite safe like that.

tests/console/verify_secure_boot.pm
```diff
@@ -17,13 +17,17 @@ use Test::Assert ':all';
...
-    assert_script_run("ls /sys/firmware/efi/vars/SecureBoot-*/data");
+    record_info("Check file", "Check if file /sys/firmware/efi/efivars/SecureBoot-* exists");
+    # From v6.0 kernel and onwards, the SecureBoot file resides in /sys/firmware/efi/efivars/
+    assert_script_run("ls /sys/firmware/efi/efivars/SecureBoot-*");
     record_info("Check secure boot", "Check if secure boot option is set as expected");
-    my $secure_boot = script_output("od -An -t u1 /sys/firmware/efi/vars/SecureBoot-*/data") ? 'enabled' : 'disabled';
+    my $octal_str = script_output("od -An -t u1 /sys/firmware/efi/efivars/SecureBoot-*");
+
+    # The final octal digit shows whether Secure Boot is enabled (1) or disabled (0)
+    my $secure_boot_octal = substr($octal_str, -1);
```
jknphy, line 27: instead of index we could use: `last` from `List::Util`.

ge0r:
According to https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot#After_booting_the_OS the fifth digit of the efivars file shows whether secure boot is enabled or not. 
This gave me the impression that there file should contain exactly five digits. However, judging from [this post](https://bbs.archlinux.org/viewtopic.php?id=258190) it looks like the number of digits could be larger. It is just that the fifth digit will indicate the secure boot status.
Given the above I have changed the code accordingly.

jknphy:
> According to https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface/Secure_Boot#After_booting_the_OS the fifth digit of the efivars file shows whether secure boot is enabled or not. This gave me the impression that the file should contain exactly five digits. However, judging from [this post](https://bbs.archlinux.org/viewtopic.php?id=258190) it looks like the number of digits could be larger. It is just that the fifth digit will indicate the secure boot status. Given the above I have changed the code accordingly.

Very clear now, low level but maintainable :) Thanks for sharing those interesting findings!

PR #16045, opened 2022-12-06
filesystem: Add a debug para for pynfs and cthon04

SeroSun:
Add a debug parameter(DEBUG_WITH_COMMIT) for pynfs and cthon04 to debug script issue.

- Related ticket: https://progress.opensuse.org/issues/121480
- Verification run: https://openqa.opensuse.org/t2929948
- git checkout OK with given commit set in DEBUG_WITH_COMMIT https://openqa.opensuse.org/tests/2929948#step/install/27

### DIFF
--- a/tests/nfs/install.pm
+++ b/tests/nfs/install.pm
@@ -45,7 +45,8 @@ sub install_dependencies_cthon04 {
 
 sub install_testsuite {
     my $testsuite = shift;
-    if (get_var("PYNFS")) {
+    my $debug_with_commit = get_var('DEBUG_WITH_COMMIT');
+    if (get_var('PYNFS')) {
         my $url = get_var('PYNFS_GIT_URL', 'git://git.linux-nfs.org/projects/bfields/pynfs.git');
         my $rel = get_var('PYNFS_RELEASE');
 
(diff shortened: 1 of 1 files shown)


lansuse:
LGTM

tests/nfs/install.pm
```diff
@@ -45,20 +45,29 @@ sub install_dependencies_cthon04 {
 
 sub install_testsuite {
     my $testsuite = shift;
-    if (get_var("PYNFS")) {
+    my $debug_with_commit = get_var('DEBUG_WITH_COMMIT');
+    if (get_var('PYNFS')) {
         my $url = get_var('PYNFS_GIT_URL', 'git://git.linux-nfs.org/projects/bfields/pynfs.git');
         my $rel = get_var('PYNFS_RELEASE');
```
pevik, line 51: LTP happily uses `LTP_RELEASE` (equivalent of `PYNFS_RELEASE`): https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/93621d80777bee0ef0e33ec8807b4af0f51843e8/tests/kernel/install_ltp.pm#L197-L210

I'm verifying if `git clone -b ...` works with git hash as well. But even if it not works, what prevents you to put somewhere git repository with master pointing to the commit you want to test? Or tagging it, create branch...?

Also name `DEBUG_WITH_COMMIT` is a bit strange.

pevik: Well, `git clone -b hash` does not work, but I suggest to either workaround it with your own fork and `PYNFS_GIT_URL` and `PYNFS_RELEASE`) or rename variable to `PYNFS_COMMIT`.

SeroSun: This PR is used in debugging pynfs fails, a lot of fails happen at the same time as the last commit. So I add this PR to debug if something wrong with the last commit. But not a script issue in the end. It's OK to just keep this debug parameter in the personal branch because it does not use that frequently. I will close this PR.

PR #16051, opened 2022-12-06
jeos/firstrun: Skip WiFi configuration on Leap 15.4

Vogtinator:
15.4 JeOS images got jeos-firstboot-rpiwifi as well now, deal with it.

- Verification run: https://openqa.opensuse.org/tests/2932104


Vogtinator:
VR didn't finish yet but it'll definitely be broken without this

PR #16058, opened 2022-12-07
Repair save_and_upload_systemd_unit_log subroutine

pdostal:
this is fixing de0279c9365a9e7902512281eb8d28ef9927f426 (#15977)


frankenmichl (approved):
Thanks for the fix, looks good to me

PR #16061, opened 2022-12-07
Simplify default instance type for ec2

grisu48:
The default instance type evaluation doesn't work for ARM instances and is needlessly complicated. This commit changes the behaviour to always returns a `large` instance of the corresponding family (t2 for x86_64 and a1 for arm64), which simplifies the workflow considerably while increasing the running costs only by a negligible amount.

- Verification run: https://duck-norris.qe.suse.de/tests/11242

### DIFF
--- a/lib/publiccloud/ec2.pm
+++ b/lib/publiccloud/ec2.pm
@@ -37,6 +37,14 @@ sub find_img {
     return;
 }
 
+sub get_default_instance_type {
+    # Returns the default machine family type to be used, based on the public cloud architecture
+
+    my $arch = get_var("PUBLIC_CLOUD_ARCH", "x86_64");
+    return "a1.large" if ($arch eq 'arm64');
+    return "t2.large";
+}
+
 sub create_keypair {
     my ($self, $prefix, $out_file) = @_;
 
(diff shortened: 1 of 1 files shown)


pdostal (approved):

lib/publiccloud/ec2.pm
```diff
@@ -37,6 +37,14 @@ sub find_img {
     return;
 }
 
+sub get_default_instance_type {
```
mimi1vx, line 40: maybe a more specific name?  this is too generic?

lib/publiccloud/ec2.pm
```diff
@@ -126,9 +134,7 @@ sub upload_img {
     my $img_arch = get_var('PUBLIC_CLOUD_ARCH', 'x86_64');
     my $sec_group = get_var('PUBLIC_CLOUD_EC2_UPLOAD_SECGROUP');
     my $vpc_subnet = get_var('PUBLIC_CLOUD_EC2_UPLOAD_VPCSUBNET');
-    my @instance_main_type = split(/\./, get_var('PUBLIC_CLOUD_INSTANCE_TYPE'));
-    my $instance_size = get_var('PUBLIC_CLOUD_IMAGE_LOCATION') =~ /-SAP-/ ? 'large' : 'micro';
```
asmorodskyi, line 130: isn't you changing SAP related logic ? Have you confirmed this with SAP team ? @ricardobranco777 FYI

dzedro: I added this because SAP upload failed with smaller instance. But now is the instance large by default, so should be OK.

PR #16062, opened 2022-12-07
Public Cloud: Refactor the check_registercloudguest test module

pdostal:
Instead of `run_ssh_command` switch to `ssh_script_run`, `ssh_assert_script_run` and `ssh_script_output`

- Related ticket: [poo#121630](https://progress.opensuse.org/issues/121630)
- Verification run: [sle-15-SP4-AZURE-BYOS-Updates@publiccloud_consoletests@aarch64](https://pdostal-server.suse.cz/tests/3453), [sle-15-SP4-EC2-Updates@publiccloud_consoletests@aarch64](https://pdostal-server.suse.cz/tests/3452), [sle-15-SP4-EC2-BYOS-Updates@publiccloud_consoletests@aarch64](https://pdostal-server.suse.cz/tests/3451)

### DIFF
--- a/tests/publiccloud/check_registercloudguest.pm
+++ b/tests/publiccloud/check_registercloudguest.pm
@@ -39,80 +39,79 @@ sub run {
         # all BYOS related checks. So we just regestering system and going further
         registercloudguest($instance);
     } elsif (is_byos()) {
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
             die 'The BYOS instance should be unregistered and report "Warning: No repositories defined.".';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
+        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
             die('guestregister.service is not disabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
-            $instance->run_ssh_command(cmd => 'sudo ls -la /etc/zypp/credentials.d/', proceed_on_failure => 1);
-            die('/etc/zypp/credentials.d/ is not empty');
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l') != 0) {
+            $instance->ssh_script_run(cmd => 'sudo ls -la /etc/zypp/credentials.d/');
+            die("/etc/zypp/credentials.d/ is not empty:\n" . $instance->ssh_script_output(cmd => 'sudo ls -la /etc/zypp/credentials.d/'));
         }
 
-        if (is_azure() && $instance->run_ssh_command(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer', proceed_on_failure => 1) !~ /enabled/) {
+        if (is_azure() && $instance->ssh_assert_script_run(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer')) {
             die('regionsrv-enabler-azure.timer is not enabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) != 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister') != 0) {
             die('/var/log/cloudregister is not empty');
         }
-        # The `sudo SUSEConnect -d` is not supported on BYOS and should fail.
-        $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d');
+        $instance->ssh_assert_script_run(cmd => '! sudo SUSEConnect -d', fail_message => 'SUSEConnect succeeds but it is not supported should fail on BYOS');
     } else {
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', proceed_on_failure => 1, timeout => 360) < 5) {
-            record_info('zypper lr', $instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1));
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 360) < 5) {
+            record_info('zypper lr', $instance->ssh_script_output(cmd => 'sudo zypper lr'));
             die 'The list of zypper repositories is too short.';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
+        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
             die('guestregister.service is not enabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) == 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l') == 0) {
             die('/etc/zypp/credentials.d/ is empty');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) == 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister') == 0) {
             die('/var/log/cloudregister is empty');
         }
     }
 
     my $path = is_sle('>15') && is_sle('<15-SP3') ? '/usr/sbin/' : '';
-    $instance->run_ssh_command(cmd => "sudo ${path}registercloudguest --clean");
-    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600, proceed_on_failure => 1) > 2) {
+    $instance->ssh_assert_script_run(cmd => "sudo ${path}registercloudguest --clean");
+    if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 600) > 2) {
         die('The list of zypper repositories is not empty.');
     }
-    if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l', proceed_on_failure => 1) != 0) {
+    if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') != 0) {
         die('Directory /etc/zypp/credentials.d/ is not empty.');
     }
 
     # The SUSEConnect registration should still work on BYOS
     if (is_byos()) {
-        $instance->run_ssh_command(cmd => 'sudo SUSEConnect --version');
-        $instance->run_ssh_command(cmd => "sudo SUSEConnect $regcode_param");
+        $instance->ssh_assert_script_run(cmd => 'sudo SUSEConnect --version');
+        $instance->ssh_assert_script_run(cmd => "sudo SUSEConnect $regcode_param");
         # The registercloudguest tool is not yet part of 15-SP5 as the infrastructure is not yet ready for it.
-        $instance->run_ssh_command(cmd => "sudo ${path}registercloudguest --clean") if (get_var('PUBLIC_CLOUD_QAM'));
+        $instance->ssh_assert_script_run(cmd => "sudo ${path}registercloudguest --clean") if (get_var('PUBLIC_CLOUD_QAM'));
     }
 
     # The registercloudguest tool is not yet part of 15-SP5 as the infrastructure is not yet ready for it.
-    $instance->run_ssh_command(cmd => "sudo ${path}registercloudguest $regcode_param") if (get_var('PUBLIC_CLOUD_QAM'));
-    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
+    $instance->ssh_assert_script_run(cmd => "sudo ${path}registercloudguest $regcode_param") if (get_var('PUBLIC_CLOUD_QAM'));
+    if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
         die('The list of zypper repositories is empty.');
     }
-    if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
+    if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
         die('Directory /etc/zypp/credentials.d/ is empty.');
     }
 
     # The registercloudguest tool is not yet part of 15-SP5 as the infrastructure is not yet ready for it.
     if (get_var('PUBLIC_CLOUD_QAM')) {
-        $instance->run_ssh_command(cmd => "sudo ${path}registercloudguest $regcode_param --force-new");
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
+        $instance->ssh_assert_script_run(cmd => "sudo ${path}registercloudguest $regcode_param --force-new");
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 600) == 0) {
             die('The list of zypper repositories is empty.');
         }
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/* | wc -l') == 0) {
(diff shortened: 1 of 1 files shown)


tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
         # all BYOS related checks. So we just regestering system and going further
         registercloudguest($instance);
     } elsif (is_byos()) {
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
```
grisu48, line 42: Do we really want `proceed_on_failure => 1` here?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
+        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
             die('guestregister.service is not disabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
-            $instance->run_ssh_command(cmd => 'sudo ls -la /etc/zypp/credentials.d/', proceed_on_failure => 1);
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
```
grisu48, line 50: Do we really want `proceed_on_failure => 1` here? (Don't make much of a difference for THIS command here though)

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
+            $instance->ssh_script_run(cmd => 'sudo ls -la /etc/zypp/credentials.d/');
             die('/etc/zypp/credentials.d/ is not empty');
         }
 
-        if (is_azure() && $instance->run_ssh_command(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer', proceed_on_failure => 1) !~ /enabled/) {
+        if (is_azure() && $instance->ssh_script_output(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer', proceed_on_failure => 1) !~ /enabled/) {
```
grisu48, line 55: Do we really want `proceed_on_failure => 1` here?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
             die('regionsrv-enabler-azure.timer is not enabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) != 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) != 0) {
```
grisu48, line 59: Do we really want `proceed_on_failure => 1` here?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
             die('/var/log/cloudregister is not empty');
         }
         # The `sudo SUSEConnect -d` is not supported on BYOS and should fail.
-        $instance->run_ssh_command(cmd => '! sudo SUSEConnect -d');
+        $instance->ssh_assert_script_run(cmd => '! sudo SUSEConnect -d');
```
grisu48, line 63:
```suggestion
        $instance->ssh_assert_script_run(cmd => '! sudo SUSEConnect -d', fail_message => "SUSEConnect succeeds when it should fail on BYOS");
```

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
     } else {
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', proceed_on_failure => 1, timeout => 360) < 5) {
-            record_info('zypper lr', $instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1));
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', proceed_on_failure => 1, timeout => 360) < 5) {
```
grisu48, line 65: Do we really want `proceed_on_failure => 1` here? Error codes above 5 would make this check pass, but they shouldn't.

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
+            record_info('zypper lr', $instance->ssh_script_run(cmd => 'sudo zypper lr'));
             die 'The list of zypper repositories is too short.';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
+        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
```
grisu48, line 70: Do we really want `proceed_on_failure => 1` here?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
+        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) == 0) {
             die('/var/log/cloudregister is empty');
         }
     }
 
     my $path = is_sle('>15') && is_sle('<15-SP3') ? '/usr/sbin/' : '';
-    $instance->run_ssh_command(cmd => "sudo ${path}registercloudguest --clean");
-    if ($instance->run_ssh_command(cmd => 'sudo zypper lr | wc -l', timeout => 600, proceed_on_failure => 1) > 2) {
+    $instance->ssh_assert_script_run(cmd => "sudo ${path}registercloudguest --clean");
+    if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 600, proceed_on_failure => 1) > 2) {
```
grisu48, line 85: Do we really want `proceed_on_failure => 1` here? Error code 1 would make this check pass, where it shouldn't, right?

grisu48 (commented):
There are some `proceed_on_failure => 1` that might be problematic.

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: same idea for all similar questions below . won't spam PR with duplicating it

grisu48: Which `zypper` errors can occur when listing the local repositories?

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,80 @@ sub run {
...
             die 'The list of zypper repositories is too short.';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
+        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
             die('guestregister.service is not enabled');
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) == 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) == 0) {
```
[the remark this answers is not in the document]
asmorodskyi: and this *certain* case I would actually support your question

tests/publiccloud/check_registercloudguest.pm
pdostal:
```
$ zypper lr; echo $?
Warning: No repositories defined.
Use the 'zypper addrepo' command to add one or more repositories.
6
```

asmorodskyi: I have no idea , zypper may simply crash on calling list repo command just because poorly written code or some library missing , there could be billions reasons why package manager will throw error . And the point that here we want to fail only in one very certain situation . Because we not testing zypper in this certain test but we using zypper as testing tool allowing to prove that there is no repos nothing more than that

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,75 @@ sub run {
         # all BYOS related checks. So we just regestering system and going further
         registercloudguest($instance);
     } elsif (is_byos()) {
-        if ($instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1) !~ /No repositories defined/gm) {
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr') !~ /No repositories defined/gm) {
```
asmorodskyi, line 42: you did a VR for PAYG image , please clone BYOS one . I suspect that you will get unexpected behavior here ...

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,75 @@ sub run {
...
             die 'The BYOS instance should be unregistered and report "Warning: No repositories defined.".';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /disabled/) {
-            die('guestregister.service is not disabled');
-        }
+        $instance->ssh_assert_script_run(cmd => '! sudo systemctl is-enabled guestregister.service', fail_message => 'guestregister.service is not disabled');
```
asmorodskyi, line 46: again `systemctl is-enabled guestregister.service` may return non-zero exit code not only because service is disabled . systemctl crash will also be treated by this version of the test as positive

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,75 @@ sub run {
...
-            record_info('zypper lr', $instance->run_ssh_command(cmd => 'sudo zypper lr', proceed_on_failure => 1));
+        if ($instance->ssh_script_output(cmd => 'sudo zypper lr | wc -l', timeout => 360) < 5) {
+            record_info('zypper lr', $instance->ssh_script_output(cmd => 'sudo zypper lr'));
             die 'The list of zypper repositories is too short.';
         }
 
-        if ($instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/) {
-            die('guestregister.service is not enabled');
-        }
+        $instance->ssh_assert_script_run(cmd => 'sudo systemctl is-enabled guestregister.service', fail_message => 'guestregister.service is not enabled');
```
asmorodskyi, line 67: we dramatically changing test coverage and test behavior here , let's not do this

tests/publiccloud/check_registercloudguest.pm
grisu48: > ```shell
> $ zypper lr; echo $?
> Warning: No repositories defined.
> Use the 'zypper addrepo' command to add one or more repositories.
> 6
> ```

Does it make sense to continue in such a case? I though the check for empty repositories is exactly what we are doing here, no?

> I have no idea , zypper may simply crash on calling list repo command just because poorly written code or some library missing , there could be billions reasons why package manager will throw error . And the point that here we want to fail only in one very certain situation . Because we not testing zypper in this certain test but we using zypper as testing tool allowing to prove that there is no repos nothing more than that

Yes but if `zypper` is broken then there is no point in continuing because "ex falso quodlibet".

Keep in mind: This is a local zypper call without network interaction, so the chance of sporadic failures due to network issues, outdated repositories, SCC, ... does not happen here. That's why I don't see why we should make a bad compromise **at this place**. If it would be a `zypper ref` or something else, we wouldn't have this argument :slightly_smiling_face:

tests/publiccloud/check_registercloudguest.pm
pdostal: We're not changing test coverage but rather getting rid of workarounds caused by `run_ssh_command`.

asmorodskyi: `$instance->run_ssh_command(cmd => 'sudo systemctl is-enabled guestregister.service', proceed_on_failure => 1) !~ /enabled/`

test will die ONLY when `systemctl is-enabled guestregister.service`  will output string containing "enabled"

`$instance->ssh_assert_script_run(cmd => 'sudo systemctl is-enabled guestregister.service', fail_message => 'guestregister.service is not enabled');`

ANY none-zero exit code will cause test failure

P.S. I would like to avoid discussion around "how big chances that something else will happen here "

grisu48: Any other non-zero exit code would mean, that we can't say if the service is disabled and thus the state is undefined. In such cases we should fail the test run.

If you can't test a requirement, then the only safe assumption is that it's broken.

pdostal: The `systemctl is-enabled` will return non zero code only when the service is not enabled. I use this check on many tests for several years and I did not record exception so far. I also initiated this test module and the check for `"enabled"` is there as a workaround for my problems with `run_ssh_command`.

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: > If it is or if the state cannot be determined (i.e. your other non-zero exit codes) then the test should fail

exactly ! totally agree we should fail and we won't  because "! sudo systemctl is-enabled guestregister.service" will give you 0 for ANY none-zero value due to "!"

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: please look for "is-enabled" in systemctl documentation  https://www.freedesktop.org/software/systemd/man/systemctl.html

additionally to enabled/disabled service can be : 
-  masked
- static
- indirect
- transient
- bad

some of them will causing 0 exit code others none zero this is what I am calling "change of test coverage" .

and I would like to ask again - **let's not speak about "what probability of getting status transient" ? I never saw it in my whole life **  we not betting here . If program has some status we should expect that it will return it

tests/publiccloud/check_registercloudguest.pm
grisu48: Sorry my bad. Yes Anton is right here. @pdostal we need to really check or for the `1` value of for the `disabled` output, otherwise any other issues are undetected. I would leave this as close as possible to the original state

```suggestion
        if ($instance->ssh_script_output(cmd => 'sudo systemctl is-enabled guestregister.service') !~ /disabled/) die('guestregister.service is not disabled');
```

And yes, in principle this should not make a difference, but since the topic was raised ...

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: so for example "static" return 0 too  BUT static is - " The unit file **is not enabled**, and has no provisions for enabling in the [Install] unit file section."

tests/publiccloud/check_registercloudguest.pm
```diff
@@ -39,80 +39,75 @@ sub run {
...
 
-        if ($instance->run_ssh_command(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l', proceed_on_failure => 1) != 0) {
-            $instance->run_ssh_command(cmd => 'sudo ls -la /etc/zypp/credentials.d/', proceed_on_failure => 1);
+        if ($instance->ssh_script_output(cmd => 'sudo ls /etc/zypp/credentials.d/ | wc -l') != 0) {
+            $instance->ssh_script_run(cmd => 'sudo ls -la /etc/zypp/credentials.d/');
             die('/etc/zypp/credentials.d/ is not empty');
         }
 
-        if (is_azure() && $instance->run_ssh_command(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer', proceed_on_failure => 1) !~ /enabled/) {
+        if (is_azure() && $instance->ssh_sssert_cript_run(cmd => 'sudo systemctl is-enabled regionsrv-enabler-azure.timer')) {
```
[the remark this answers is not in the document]
pdostal: https://openqa.suse.de/tests/10106797#

tests/publiccloud/check_registercloudguest.pm
asmorodskyi: > I use this check on many tests for several years and I did not record exception so far.

who knows maybe you have missed some service in "static" state due to taking such approach

tests/publiccloud/check_registercloudguest.pm
pdostal: https://openqa.suse.de/tests/10106797#

tests/publiccloud/check_registercloudguest.pm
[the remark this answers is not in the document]
asmorodskyi: > https://openqa.suse.de/tests/10106797#

CASEDIR | https://github.com/pdostal/os-autoinst-distri-opensuse.git#15sp5numdisks

tests/publiccloud/check_registercloudguest.pm
pdostal: @asmorodskyi @grisu48 please let me know what I should do here.

grisu48: From my point of view: Simply remove the `proceed_on_failure => 1`, if possible.

tests/publiccloud/check_registercloudguest.pm
grisu48: I really think this one is fine as it is. If it works, everything is fine, if it fails for whatever reason the test fails. What more do we want?

asmorodskyi: so you don't care if service is enabled or it is "static" state with broken service file ?

asmorodskyi (approved):
feel free to merge ...

PR #16059, opened 2022-12-07
security/yast2_users/add_users.pm: improve test reliability

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/117163
- Verification run: https://openqa.suse.de/tests/10098065

PR #16063, opened 2022-12-07
Revert "security/yast2_users/add_users.pm: improve test reliability"

DimStar77:
Reverts os-autoinst/os-autoinst-distri-opensuse#16059

Expecting nokogiri to print out a warning to be compiled about different components (mostly libxml in my experience) is wrong. nokogiri needs to be properly rebuild.

### DIFF
--- a/tests/security/yast2_users/add_users.pm
+++ b/tests/security/yast2_users/add_users.pm
@@ -74,10 +74,17 @@ sub run {
     # If no, check the user was created successfully
     assert_screen("Yast2-Users-Add-User-Created");
     wait_screen_change { send_key "alt-o" };
-    assert_screen("yast2-user-add_xterm_nokogiri");
+    # Until bsc#1202053 is fixed, we need to send "enter" a couple of times before we can clear the console
+    send_key "ret";
+    send_key "ret";
+
+    # Enhence code for stability: avoid time racing
+    clear_console;
+    assert_screen("root-console-x11");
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
Looks more like it's just missing a needle

DimStar77:
> Looks more like it's just missing a needle

a 'workaround' needle to be precise

Vogtinator (changes requested):
On osd the test doesn't match the warning message at all - just the xterm window title. It's probably made explicitly not care about the window content, which would differ between warning shown (prompt in a different location) and not shown cases

Vogtinator:
> > Looks more like it's just missing a needle
> 
> a 'workaround' needle to be precise (on the SLE side, that resulted in the poo ticket in the first place)

You can add a needle which matches "warning not there" without a workaround tag

DimStar77:
the change I'm reverting here is the one that added the request to have a needle called 'nokogiri' which clearly shows they want nokogiri to print a warning. There was no such need - and the warning on screen is something to be addressed in the distro that sees it.

paolostivanin:
sorry for that. I'm gonna fix the code right now

DimStar77:
Checking that the terminal is back there and usable - as done before, is what the correct behaviour should be.

paolostivanin:
PR with improved fix: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16064

PR #16064, opened 2022-12-07
[WIP] security/yast2_users/add_users.pm: fix issue introduced with #16059

paolostivanin:
- Verification run: 
  - 15-SP5:
  - openSUSE TW:


tests/security/yast2_users/add_users.pm
```diff
@@ -74,11 +74,13 @@ sub run {
     # If no, check the user was created successfully
     assert_screen("Yast2-Users-Add-User-Created");
     wait_screen_change { send_key "alt-o" };
+
     # Until bsc#1202053 is fixed, we need to send "enter" a couple of times before we can clear the console
-    send_key "ret";
-    send_key "ret";
+    if (check_screen("yast2-user-add_xterm_nokogiri", 10)) {
```
DimStar77, line 79: that's a slow-down of 10s for every run where nokogiri works properly

paolostivanin: I can try to lower it to 3 and see if that's enough. What do you think?

paolostivanin: you mean to manually add a `record_soft_fail` ? It shouldn't be needed since the needle is marked as "workaround"

paolostivanin:
New VRs with 3s as timeout:
TW: https://openqa.opensuse.org/tests/2935170
15sp5: :on:

tests/security/yast2_users/add_users.pm
DimStar77: how do you guarantee that everybody creating new needles with the tag will always be marked workaround? 

one worry: what happens with check_screen("non-existing-needle") which would be the case currently on TW, as we don't have that problem and could not create a possibly matching needle for it (or you need to copy the one from SLE which we'll hopefully never use)

paolostivanin: How do you suggest to improve `check_screen`?
I could check `if tumbleweed then skip check_screen`, but this would imply that this issue will never occur on TW :thinking:

paolostivanin: good point about the softail and the needle. I'll add it!

DimStar77: I rather meant in the backend: check_screen against 0 needles has no reason to wait $timeout. it won't ever match (your code in this PR is fine)

DimStar77 (approved):

PR #16065, opened 2022-12-07
Containers: SLE15-SP5: Use client 1.23 instead of 1.18

pdostal:
This is because SLE15-SP5 comes with the new version.

- Verification run: [Tumbleweed](https://pdostal-server.suse.cz/tests/3509), [15-SP5](https://pdostal-server.suse.cz/tests/3510), [15-SP4](https://pdostal-server.suse.cz/tests/3507), [15-SP3](https://pdostal-server.suse.cz/tests/3511), [15-SP2](https://pdostal-server.suse.cz/tests/3512), [15-SP1](https://pdostal-server.suse.cz/tests/3508)

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -84,7 +84,9 @@ sub install_kubectl {
             # kubectl is in the container module
             add_suseconnect_product(get_addon_fullname('contm'));
             # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
-            if (is_sle(">15-SP1")) {
+            if (is_sle(">=15-SP3")) {
+                zypper_call("in kubernetes1.23-client");
+            } elsif (is_sle("=15-SP2")) {
                 zypper_call("in kubernetes1.18-client");
             } else {
                 zypper_call("in kubernetes-client");


grisu48 (commented):
Please also state in the PR description WHY we're using 1.23 instead of 1.18 for anyone to whom this is not obvious (like me).

pdostal:
I don't know why.

mloviska:
> I don't know why.

Because sle15sp5 comes with newer version, I guess that would be the answer what we can provide here. :)

lib/containers/k8s.pm
```diff
@@ -84,7 +84,9 @@ sub install_kubectl {
             # kubectl is in the container module
             add_suseconnect_product(get_addon_fullname('contm'));
             # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
-            if (is_sle(">15-SP1")) {
+            if (is_sle(">=15-SP5")) {
```
asmorodskyi, line 87:
```suggestion
            if (is_sle("==15-SP5")) {
```

I think it will more clear message , also I would assume for SP6 we would need some code tweaking here anyway ...

mloviska: This would lead to a syntax error. Anyway, I believe we should try to use zypper's capabilities to download the best fitting package, rather than bumping the versions for each service pack.

mloviska: I have just tested the installation in TW, `kubernetes-client` always pulled the newest one `1.24`. Do you remember why we have introduced this complex logic?

grisu48:
> I don't know why.

As Martin stated,the question is not why it is this version, but why you are doing this change. The justification "On 15-SP5 the kubectl version is now 1.23" is all we need to understand why you are making this change. Nothing more, but also nothing less.

"But isn't that clear?" No it isn't. If we're gonna look in a year from now at this specific commit in your git history, we will have no idea why this change was made. Could you tell? Maybe there were two different versions and we needed a feature from the newer one? Maybe there was only the newer version available? Hard to tell without any context description.

A commit should (if possible) contain enough context to understand why this change was introduced, also if you will look at it in 2 years from now. That's a courtesy to your reviewers and your future self :slightly_smiling_face: 

Can you please also update the commit message as PR descriptions will be only available in GitHub but not in the `git log` itself?


Sorry for being picky, but I strongly believe this is important, otherwise we're losing the battle against the always increasing entropy even faster than we anyways do ...

lib/containers/k8s.pm
grisu48: I **THINK** (!!!!) because at some point `kubernetes-client` stopped working.

pdostal: @mloviska [SLE 15-SP4](https://pdostal-server.suse.cz/tests/3464):
```
# zypper -n in kubernetes-client | cat; ( exit ${PIPESTATUS[0]} ); echo s_CXw-$?-
Refreshing service 'Basesystem_Module_15_SP4_x86_64'.
Refreshing service 'Containers_Module_15_SP4_x86_64'.
Refreshing service 'Desktop_Applications_Module_15_SP4_x86_64'.
Refreshing service 'Development_Tools_Module_15_SP4_x86_64'.
Refreshing service 'Legacy_Module_15_SP4_x86_64'.
Refreshing service 'Python_3_Module_15_SP4_x86_64'.
Refreshing service 'SUSE_Linux_Enterprise_Server_15_SP4_x86_64'.
Refreshing service 'SUSE_Linux_Enterprise_Workstation_Extension_15_SP4_x86_64'.
Refreshing service 'Server_Applications_Module_15_SP4_x86_64'.
Refreshing service 'Web_and_Scripting_Module_15_SP4_x86_64'.
Building repository 'SLE-Module-Containers15-SP4-Pool' cache [....done]
Retrieving repository 'SLE-Module-Containers15-SP4-Updates' metadata [.done]
Building repository 'SLE-Module-Containers15-SP4-Updates' cache [....done]
Loading repository data...
Reading installed packages...
'kubernetes-client' not found in package names. Trying capabilities.
No provider of 'kubernetes-client' found.
```

grisu48 (approved):

jlausuch:
The reason it's done this way is because (for some reason) the version is hardcoded in the name of the package.
If you take a look at [scc](https://scc.suse.com/packages?name=SUSE%20Linux%20Enterprise%20Server&version=15.5&arch=x86_64&query=kubernetes&module=2486)...
So, there is no package called `kubernetes-client` you can install and get the "latest" version... actually, 1.23 is also available for 15-SP4 and 15-SP3..

lib/containers/k8s.pm
```diff
@@ -84,7 +84,9 @@ sub install_kubectl {
             # kubectl is in the container module
             add_suseconnect_product(get_addon_fullname('contm'));
             # SLES-15SP2+ ships a specific kubernetes client version. Older versions hold a version-independent kubernetes-client package.
-            if (is_sle(">15-SP1")) {
+            if (is_sle("=15-SP5")) {
```
jlausuch, line 87:
```suggestion
            if (is_sle(">=15-SP3")) {
```

lib/containers/k8s.pm
```diff
@@ -84,7 +84,9 @@ sub install_kubectl {
...
+                zypper_call("in kubernetes1.23-client");
+            } elsif (is_sle(">15-SP1")) {
```
jlausuch, line 89:
```suggestion
            } elsif (is_sle("=15-SP2")) {
```

grisu48 (approved):

jlausuch (approved):

jlausuch (approved):

PR #16069, opened 2022-12-08
[WIP] ltp: Use SERIALDEV to configure serial console on XEN

czerw:
Serial console is currently hardcoded on ttys. It works on majority of available hardware, but there are exceptions which use different tty configured by `SERIALDEV` variable. 
- Related ticket: none
- Needles: 
- Verification run:


mdoucha (approved):
LGTM

PR #16070, opened 2022-12-08
Add architecture handling for GCE image upload

grisu48:
Adds a handle to include PUBLIC_CLOUD_ARCH when uploading an new image to GCE. This is required to allow the upload of custom arm64 images e.g. for SLEM on PublicCloud.

- Verification run: https://duck-norris.qe.suse.de/tests/11257#step/upload_image/32

### DIFF
--- a/lib/publiccloud/gce.pm
+++ b/lib/publiccloud/gce.pm
@@ -51,11 +51,17 @@ sub upload_img {
     # See https://cloud.google.com/sdk/gcloud/reference/compute/images/create for a list of available features
     # SEV_CAPABLE is added because all images from 15-SP2 onwards support SEV
     my $guest_os_features = get_var('PUBLIC_CLOUD_GCE_UPLOAD_GUEST_FEATURES', 'MULTI_IP_SUBNET,UEFI_COMPATIBLE,VIRTIO_SCSI_MULTIQUEUE,SEV_CAPABLE');
+    my $arch = get_var('PUBLIC_CLOUD_ARCH', '');
 
     assert_script_run("gsutil cp '$file' 'gs://$uri'", timeout => 60 * 60);
 
     my $cmd = "gcloud compute images create '$img_name' --source-uri 'gs://$uri'";
     $cmd .= " --guest-os-features '$guest_os_features'" unless (trim($guest_os_features) eq '');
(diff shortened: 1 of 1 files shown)


lib/publiccloud/gce.pm
```diff
@@ -51,11 +51,17 @@ sub upload_img {
     # See https://cloud.google.com/sdk/gcloud/reference/compute/images/create for a list of available features
     # SEV_CAPABLE is added because all images from 15-SP2 onwards support SEV
     my $guest_os_features = get_var('PUBLIC_CLOUD_GCE_UPLOAD_GUEST_FEATURES', 'MULTI_IP_SUBNET,UEFI_COMPATIBLE,VIRTIO_SCSI_MULTIQUEUE,SEV_CAPABLE');
+    my $arch = get_var('PUBLIC_CLOUD_ARCH', '');
```
pdostal, line 54: There are already places where we assume that the default architecture is `x86_64` so you can do:
```suggestion
    my $arch = get_var('PUBLIC_CLOUD_ARCH', 'x86_64');
```

grisu48: But this would change the current workflow, where we don't set the architecture.

So, except for cosmetic changes - why would this be better?

asmorodskyi: I also thought what Pavel suggesting but than I concluded exactly the same what Felix replying :)

pdostal: I don't really get how the 'workflouw' would be changed? Everywhere is by default expected `x86_64` … why is this suddenly different?

pdostal: Probably no. We can leave this for another refactoring.

But it is adding 6 lines and one condition vs. one line and zero conditions.

grisu48: > Probably no. We can leave this for another refactoring.
> 
> But it is adding 6 lines and one condition vs. one line and zero conditions.

But all other modifications are still needed, because we haven't used the `--architecture` parameter before. That's exactly why I don't see why this change would make any difference.

We haven't been using `PUBLIC_CLOUD_ARCH` in this test run at all before, it got introduced in this PR.

PR #16083, opened 2022-12-12
Switch possible parts of post_fail_hooks to serial terminal

pdostal:
- Related ticket: [poo#120735](https://progress.opensuse.org/issues/120735)
- Follow up to: de0279c (#15977) and 4bad8a0 (#16058)
- Verification run: [extra_tests_textmode@64bit](https://pdostal-server.suse.cz/tests/3516), [extra_tests_gnome@64bit](https://pdostal-server.suse.cz/tests/3515)

### DIFF
--- a/lib/Utils/Logging.pm
+++ b/lib/Utils/Logging.pm
@@ -16,6 +16,7 @@ use warnings;
 use testapi;
 use utils qw(clear_console show_oom_info remount_tmp_if_ro detect_bsc_1063638);
 use Utils::Systemd 'get_started_systemd_services';
+use serial_terminal 'select_serial_terminal';
 use Mojo::File 'path';
 
 our @EXPORT = qw(
@@ -183,8 +184,7 @@ This method will call several other log gathering methods from this class.
 =cut
 
 sub export_logs {
-    select_log_console();
-    save_screenshot();
+    select_serial_terminal();
     show_oom_info();
     remount_tmp_if_ro();
     export_logs_basic();
@@ -193,8 +193,6 @@ sub export_logs {
     # Just after the setup: let's see the network configuration
     save_and_upload_log("ip addr show", "/tmp/ip-addr-show.log");
     save_and_upload_log("cat /etc/resolv.conf", "/tmp/resolv-conf.log");
-
-    save_screenshot();
 
     export_logs_desktop();
 
@@ -336,8 +334,7 @@ Upload several KDE, GNOME, X11, GDM and SDDM related logs and configs.
 =cut
 
 sub export_logs_desktop {
-    select_log_console;
-    save_screenshot;
+    select_serial_terminal();
 
     if (check_var("DESKTOP", "kde")) {
         if (get_var('PLASMA5')) {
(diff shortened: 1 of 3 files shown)


foursixnine:
@pdostal, do you have some VR's for us? TW/Leap15.5 and JeOS? :)

pdostal:
@foursixnine done.

okurz (changes requested):

okurz (commented):
Please describe in the git commit message what the benefit of the change is, at best make it prominent in the git commit message subject line. To me it's not obvious that the advantage is

lib/Utils/Logging.pm
```diff
@@ -183,8 +184,7 @@ This method will call several other log gathering methods from this class.
 =cut
 
 sub export_logs {
-    select_log_console();
-    save_screenshot();
```
okurz, line 187: Someone put in these `save_screenshot` calls in on purpose. I am not sure it's a good idea to remove those. I suggest you show in more details in the git commit message how the output looked in before and how it's better now (if it actually is)

lib/consoletest.pm
```diff
@@ -51,13 +50,6 @@ sub post_fail_hook {
     return if get_var('NOLOGS');
     record_avc_selinux_alerts();
     $self->SUPER::post_fail_hook;
-    # at this point the instance is shutdown
-    return if (is_public_cloud() || is_openstack());
-    select_console('log-console');
-    remount_tmp_if_ro;
```
okurz, line 57: Where did you move this code? Or do you just remove it without replacement?

lib/Utils/Logging.pm
pdostal: As I switched from log console to serial terminal, do we still really need this?

lib/consoletest.pm
pdostal: This is already executed in `Utils::Logging::export_logs()` which is called from `opensusebasetest::post_fail_hook()` and so it makes no sense to be duplicated also here.

lib/Utils/Logging.pm
okurz: Well, I don't know that so you need to convince us :)

lib/consoletest.pm
okurz: Ok, cool

lib/Utils/Logging.pm
pdostal: This does not make sense. After my change we don't switch to log console so we do not have to take picture of it.

What I'm saying is that before we switched to log console, took screenshot of it and then do something with it. But now, we don't switch to it at all.

okurz: oh, yes. I see that now. Forgive me, I read it the other way around where a screenshot was taken from wherever we would come frome *before* we select the log console. So all is good.

okurz (approved):

BillAnastasiadis (approved):
LGTM

DimStar77:
Responsible for https://openqa.opensuse.org/tests/2968643#step/kate/4

Firefox is the 'real' failure, but then we end up on serial console, reload a VM snapshot state of LASTGOOD (which is on X-console), continue typing over VNC until we end up in kate does that ensure_installed, which interacts on serial and is completely consfused

PR #16084, opened 2022-12-12
ulp_openposix: Add support for testing arbitrary incidents

mdoucha:
Add support for running `ulp_openposix` on any maintenance incident. The test will first read package list from incident repos and decide how to proceed:
1) Repo contains `glibc-livepatches`: Continue as in the previous version.
2) Repo contains `libpulp0` or `libpulp-tools`: Install update, then run openposix tests livepatched by openposix-livepatches from `QA:Head` repo.
3) Repo contains none of the above: Nothing to do, exit.

- Needles: N/A
- Verification runs:
  - Normal LTP job: https://openqa.suse.de/tests/10152547
  - ULP job, case 1 (glibc-livepatches): https://openqa.suse.de/tests/10157485 (Expected to fail, the `glibc-livepatches` package intentionally breaks `calloc()`. The job uses a [workaround](https://github.com/mdoucha/os-autoinst-distri-opensuse/commit/6827d1787ad2e89c6f07344d50b49b3ccfb33763) to even allow the package to install, otherwise `zypper` would fail)
  - ULP job, case 3 (no livepatching packages): https://openqa.suse.de/tests/10157502

### DIFF
--- a/tests/kernel/ulp_openposix.pm
+++ b/tests/kernel/ulp_openposix.pm
@@ -13,13 +13,16 @@ use testapi;
 use utils;
 use serial_terminal;
 use klp;
+use qam;
 use LTP::utils;
 use OpenQA::Test::RunArgs;
 
 sub prepare_repo {
-    my ($packname) = @_;
-    my @repos = split(",", get_required_var('INCIDENT_REPO'));
+    my $incident_id = get_required_var('INCIDENT_ID');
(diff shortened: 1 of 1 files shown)


czerw (approved):
LGTM

mdoucha:
I've simplified incident package name checks to use `$$_{name} eq ...` instead of unnecessary regex.

PR #16088, opened 2022-12-13
Add kernel audit messages

grisu48:
Record the kernel audit messages before the auditd daemon was started.

- Related ticket: https://progress.opensuse.org/issues/121450
- Verification runs: [SLEM 5.4](https://duck-norris.qe.suse.de/tests/11381#step/disk_boot/15) | [MicroOS](https://duck-norris.qe.suse.de/tests/11382#step/disk_boot/17)

### DIFF
--- a/tests/microos/disk_boot.pm
+++ b/tests/microos/disk_boot.pm
@@ -13,6 +13,7 @@ use testapi;
 use version_utils qw(is_sle_micro is_leap_micro);
 use Utils::Architectures qw(is_aarch64);
 use microos "microos_login";
+use transactional "record_kernel_audit_messages";
 
 sub run {
 
@@ -28,6 +29,7 @@ sub run {
         shift->wait_boot(bootloader_time => 300);
     }
     microos_login;
+    record_kernel_audit_messages(log_upload => 1);
 }
 
 sub test_flags {
--- a/lib/transactional.pm
+++ b/lib/transactional.pm
@@ -35,6 +35,7 @@ our @EXPORT = qw(
   enter_trup_shell
   exit_trup_shell_and_reboot
   reboot_on_changes
+  record_kernel_audit_messages
 );
 
 # Download files needed for transactional update tests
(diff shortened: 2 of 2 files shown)


tests/microos/disk_boot.pm
```diff
@@ -28,6 +28,24 @@ sub run {
...
+Record the SELinux messages before the auditd daemon has been started, if present. If there are no such entries, this function has no effect.
+
+=cut
+
+sub record_kernel_audit_messages {
+    my $output = script_output("journalctl -k | grep 'audit:.*avc:'");
+    return unless ($output);    # Don't log anything if there is no output
+    record_info("Audit", "Kernel audit messages:\n\n$output");
+    # Upload the same log and don't fail on errors (supplemental material)
+    script_run("journalctl -k | grep 'audit:.*avc:' > /var/tmp/k-audit.log");
```
pdostal, line 46:
```suggestion
    script_run("journalctl -k -u audit -u avc > /var/tmp/k-audit.log");
```

grisu48: This does not work. `audit` is not a unit.

pdostal: OK, I don't have better idea then :smile:
image

pdostal (approved):

tests/microos/disk_boot.pm
grisu48: That's ok. Me neither.

mloviska:
I think we should run that after each boot

tests/microos/disk_boot.pm
```diff
@@ -28,6 +28,24 @@ sub run {
         shift->wait_boot(bootloader_time => 300);
     }
     microos_login;
+    record_kernel_audit_messages();
+}
+
+
+=head2 record_kernel_audit_messages
+
+Record the SELinux messages before the auditd daemon has been started, if present. If there are no such entries, this function has no effect.
```
jlausuch, line 37: old SLEM images don't have SELinux enabled by default, so maybe we could run this module `if get_var('ENABLE_SELINUX')` ?

mloviska: You may find the messages even there https://openqa.suse.de/tests/10157368#step/install_updates/41

grisu48: @jlausuch But we are not using any `SELINUX` variables in many test runs, e.g. the verification run.

grisu48:
> I think we should run that after each boot

I can add this to `process_reboot` in `transactional.pm. Would that include all desired places then?

jlausuch:
> > I think we should run that after each boot
> 
> I can add this to `process_reboot` in `transactional.pm`. Would that include all desired places then?

if you want to catch all reboots/logins, try better [microos_login()](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/microos.pm#L22)

mloviska (approved):

jlausuch (approved):

PR #16090, opened 2022-12-13
Add kernel-default-kgraft conflict for 12-SP3 kernel updates

dzedro:
https://suse.slack.com/archives/C02D16TCP99/p1670920439758239

- Verification run: https://openqa.suse.de/tests/10157842


dzedro:
As `kernel-default-kgraft` will be removed then this change is not required.

PR #16092, opened 2022-12-13
ltp: Create own installation job for ALP

czerw:
LTP installation was depending on default installation scenario, but was not stable enough and blocked kernel testing. LTP installation job will now boot from official ALP image and prepares system for further testing. Installation job needs additional variables NUMDISKS=2 and HDD_2=ignition.qcow2 to boot original image defined in HDD_1. ALP qemu image has kernel-default-base as default kernel. It is recommended to set KERNEL_BASE=1 to correctly show information during installation.

- Related ticket: none
- Needles: none
- Verification run: 
installation: http://10.100.12.105/tests/2013
test: http://10.100.12.105/tests/2014

Related jobgroup reconfiguration: https://github.com/os-autoinst/opensuse-jobgroups/pull/245

### DIFF
--- a/lib/main_ltp.pm
+++ b/lib/main_ltp.pm
@@ -11,8 +11,9 @@ use utils;
 use main_common qw(boot_hdd_image load_bootloader_s390x load_kernel_baremetal_tests);
 use 5.018;
 use Utils::Backends;
-use version_utils 'is_opensuse';
+use version_utils qw(is_opensuse is_alp);
 use LTP::utils qw(loadtest_kernel shutdown_ltp);
+use main_common 'loadtest';
 # FIXME: Delete the "## no critic (Strict)" line and uncomment "use warnings;"
 # use warnings;
 
@@ -30,6 +31,11 @@ sub load_kernel_tests {
     loadtest_kernel "../installation/bootloader" if is_pvm;
 
     if (get_var('INSTALL_LTP')) {
+        if (is_alp) {
+            loadtest('microos/disk_boot');
+            loadtest('transactional/host_config');
+        }
+
         if (get_var('INSTALL_KOTD')) {
             loadtest_kernel 'install_kotd';
         }
--- a/tests/kernel/install_ltp.pm
+++ b/tests/kernel/install_ltp.pm
@@ -364,7 +364,7 @@ sub run {
         die 'INSTALL_LTP must contain "git" or "repo"';
     }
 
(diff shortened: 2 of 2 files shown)


czerw:
@mdoucha @pevik @jlausuch  please check, it is more like POC for discussion, also it needs job group modification (if going to be accepted)

lib/main_ltp.pm
```diff
@@ -41,6 +42,10 @@ sub load_kernel_tests {
         if (get_var('FLAVOR', '') =~ /Incidents-Kernel/) {
             loadtest_kernel 'update_kernel';
         }
+        if (is_alp && is_qemu) {
```
jlausuch, line 45: I know we don't have this combination yet, but `is_qemu` is also applicable to the ISO, right?

jlausuch:
Nice! This might also help enabling LTP on SLE Micro.

lib/main_ltp.pm
```diff
@@ -41,6 +42,10 @@ sub load_kernel_tests {
         if (get_var('FLAVOR', '') =~ /Incidents-Kernel/) {
             loadtest_kernel 'update_kernel';
         }
+        if (is_alp) {
```
pevik, line 45: I wonder if this new code should be put above `install_kotd`. Although it's unlikely these modules will be used for Alp, we never now.

pevik (approved):
LGTM, thank you!

lib/main_ltp.pm
czerw: Moved up, but `install_kotd` will have to be updated as well for ALP, but this is not in the scope now.

lib/main_ltp.pm
```diff
@@ -30,6 +31,10 @@ sub load_kernel_tests {
     loadtest_kernel "../installation/bootloader" if is_pvm;
 
     if (get_var('INSTALL_LTP')) {
+        if (is_alp) {
+            loadtest('microos/disk_boot');
+            loadtest('transactional/host_config');
+        }
         if (get_var('INSTALL_KOTD')) {
```
mdoucha, line 38: Please add a blank line between the two `if` blocks so that it doesn't look like there should be an `else`. Otherwise LGTM.

mdoucha (changes requested):

lib/main_ltp.pm
pevik: I thought that as well, but didn't want to bother with it.

mdoucha (approved):
LGTM

PR #16094, opened 2022-12-14
Create a daily auto-installation for SLE 12 SP5 aarch64

lemon-suse:
For Regression job group we need to replace the use of this image SLES-12-SP5-aarch64-GM-SDK-allpatterns.qcow2 with a daily auto-installation qcow via autoyast.

- Related ticket: https://progress.opensuse.org/issues/120543
- Needles: N/A
- Related MR: https://gitlab.suse.de/qa-maintenance/qam-openqa-yml/-/merge_requests/438
- Verification run:  https://openqa.nue.suse.com/tests/10268659   create qcow
                                 https://openqa.suse.de/tests/10164484#step/patch_sle/183 regression test, failed for bug#1206255

### DIFF
--- a/data/autoyast_sle12/create_hdd/create_hdd_sles_regression_aarch64.xml.ep
+++ b/data/autoyast_sle12/create_hdd/create_hdd_sles_regression_aarch64.xml.ep
@@ -1,0 +1,411 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+    <suse_register>
+        <do_registration config:type="boolean">true</do_registration>
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        <install_updates config:type="boolean">true</install_updates>
+        % if (keys %$addons) {
+        <addons config:type="list">
+            % while (my ($key, $addon) = each (%$addons)) {
+            <addon>
+                <name><%= $addon->{name} %></name>
+                <version><%= $addon->{version} %></version>
+                <arch><%= $addon->{arch} %></arch>
+                % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sles')) {
+                <reg_code><%= $get_var->('SCC_REGCODE_WE') %></reg_code>
+                % }
+                % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+                <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+                % }
+                % if ($key eq 'rt') {
+                <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+                % }
+                % if ($key eq 'ltss') {
+                <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
+                % }
+		% if ($key eq 'ha') {
+                <reg_code><%= $get_var->('SCC_REGCODE_HA') %></reg_code>
+                % }
+            </addon>
+            % }
+        </addons>
+        %}
+    </suse_register>
+    % if ($check_var->('ARCH', 's390x')) {
+    <bootloader>
+      <global>
+        <append>hvc_iucv=8 TERM=dumb resume=/dev/disk/by-path/ccw-0.0.0000-part3 crashkernel=163M</append>
+        <gfxmode>auto</gfxmode>
+        <hiddenmenu>false</hiddenmenu>
+        <os_prober>false</os_prober>
+        <terminal>console</terminal>
+        <timeout config:type="integer">8</timeout>
+        <trusted_grub>false</trusted_grub>
+        <xen_append>crashkernel=163M</xen_append>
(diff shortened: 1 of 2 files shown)


data/autoyast_sle12/create_hdd/create_hdd_sles.xml.ep
```diff
@@ -0,0 +1,410 @@
...
+          <pattern>lamp_server</pattern>
+          <pattern>mail_server</pattern>
+          <pattern>ofed</pattern>
+          <pattern>printing</pattern>
+          <pattern>smt</pattern>
+          <pattern>x11</pattern>
+	  <pattern>yast2</pattern>
+        </patterns>
+	% }
+	% unless ($check_var->('PATTERNS', 'all')) {
```
JRivrain, line 343: We probably don't want gnome-basic and X11 if we install a base server. I would just put no `<pattern>` section at all, as in create_hdd_ha_sles : then the default patterns for the selected product will be selected.
OR, if you want to be able to select specific patterns you could try to use this section from qam's sle 12 profile:  https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/a395a6d9103cb2f20c112750e4ab0935c25cf393/data/autoyast_qam/12_installation.xml.ep#L322.

It looks like the only difference between this file and create_hdd_ha_sles is this patterns section (?) so I would merge the two files.

JRivrain:
There is a typo in the commit message: iva/via. Also, the commit should include more than one line, everything you put there is copied to PR's description anyway.

data/autoyast_sle12/create_hdd/create_hdd_sles.xml.ep
lemon-suse: Thanks. We need generate default/all patterns qcow, so we need the 'patterns' section and the condition. Maybe I can update the profile name to make it better to understand for regression migration test.

JRivrain: Hi, I understand we need the patterns if PATTERNS=all, but I don't nderstand why you want to install gnome and x11 on a text-based server. Then I see two problems:

- if we have PATTERNS variable empty, then we should  remove this section to let the system automatically detect the right patterns:
```
% unless ($check_var->('PATTERNS', 'all')) {
        <patterns config:type="list">
          <pattern>Minimal</pattern>
          <pattern>apparmor</pattern>
          <pattern>base</pattern>
          <pattern>documentation</pattern>
          <pattern>gnome-basic</pattern>
          <pattern>x11</pattern>
        </patterns>
	% }
```

- if we plan to do something different like "PATTERNS=base,mail_server", then this section is wrong.

In such case we better do this:
```
% if  (($get_var->(PATTERNS) and  !($check_var->('PATTERNS', 'all')) {
    <patterns config:type="list">
      % for my $pattern (@$patterns) {
      <pattern><%= $pattern %></pattern>
      % }
    </patterns>
% }
```
Or, again, we could use a normal profile that we will use only for one test suite, with no scripting/conditions at all.

lemon-suse: We have two kinds of PATTERNS (all/default), and this ticket aims to generate qcow to replace SLES-12-SP5-aarch64-GM-SDK-allpatterns.qcow2 which is gnome and not textmode, to keep the migration test so we need generate gnome qcow.
I updated the profile for PATTERNS=default to avoid misunderstanding, and so that it will get patterns automatically if no PATTERNS set. In fact, we only support PATTERNS=all or PATTERNS=default, anyway, to make the profile more strict is better. Thanks.

JRivrain: Ah ok, now I understand. Looks good like this.

JRivrain (approved):

data/autoyast_sle12/create_hdd/create_hdd_sles_regression.xml.ep
```diff
@@ -0,0 +1,411 @@
...
+            <partition_id config:type="integer">130</partition_id>
+            <partition_nr config:type="integer">3</partition_nr>
+            <resize config:type="boolean">false</resize>
+            <size>2G</size>
+          </partition>
+        </partitions>
+      </drive>
+    </partitioning>
+    % }
+    % unless ($check_var->('ARCH', 'aarch64')) {
```
jknphy, line 240: doesn´t exist normal `% else`, I found this difficult to maintain... doesn't get a proper partitioning if you just pass a few field or just the section below for aarch64?

jknphy: I have doubts how reusable is this when someone else try to use this profile for other architecture, because default and all might mean a different thing.

lemon-suse: Under this is not aarch64, it is unless. These conditions for partition already used in some Maintenance autoyast test, if have problems found we can maintain later I think.

For the 'all' or 'default' patterns, it may have difference between arches, cause this ticket is for aarch64, we can change the profile name to create_hdd_sles_regression_aarch64.xml.ep wdyt?  Later if one want to re-use this profile for other arches then create a new profile or update the profile to make it suitable for other arches.

PR #16095, opened 2022-12-14
Test NetworkManager on ALP

rfan1:
https://progress.opensuse.org/issues/121093

Set test module 'enable_selinux' as new milestone, since NetworkManager test changes the default system configuration.

- Verification run:  https://openqa.opensuse.org/tests/2972458
-Known issue:https://bugzilla.suse.com/show_bug.cgi?id=1206352


foursixnine (approved):

foursixnine:
@cfconrad @jlausuch, would you guys mind giving this a look? :)

lib/main_micro_alp.pm
```diff
@@ -170,6 +170,7 @@ sub load_remote_controller_tests {
 sub load_common_tests {
     loadtest 'console/regproxy' if is_regproxy_required;
     loadtest 'microos/networking';
+    loadtest 'microos/networkmanager';
```
jlausuch, line 173: I would move this to `load_network_tests`, which is executed only in ALP (for now).

jlausuch:
Looking good! Let's not execute in common tests, since it would run in SLE Micro, MicroOS, etc... let's focus in this PR in ALP only, and we can extend it to the others later on.

I remember there was some discussion about the usage of `script_run($cmd) == 0` is better to use it with  `die_on_timeout => 1`. If there is a timeout, it will return `undef` and make the condition valid, which we don't want.
I know your commands are not going to timeout cause they are rather simple, but let's try to apply best practices just in case.

tests/microos/networkmanager.pm
```diff
@@ -0,0 +1,130 @@
...
+#
+# Copyright 2016-2018 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Test NetworkManager
+# Maintainer: rfan1 <richard.fan@suse.com>
+
+use base "consoletest";
+use strict;
+use warnings;
```
cfconrad, line 11: Just a personal flavor, as openQA is heavily using Mojo already.
```suggestion
use Mojo::Base "consoletest";
```

tests/microos/networkmanager.pm
```diff
@@ -0,0 +1,130 @@
...
+'curl https://git.tuxfamily.org/chrony/chrony.git/plain/examples/chrony.nm-dispatcher.dhcp -o /etc/NetworkManager/dispatcher.d/chrony.nm-dispatcher.dhcp'
+    );
+    script_run(
+        'cat > /etc/chrony.d/dhcp-source.conf << EOF
+sourcedir /var/run/chrony-dhcp
+EOF
+true'
+    );
+    systemctl('restart chronyd');
+    systemctl('restart NetworkManager');
```
cfconrad, line 81: Since https://build.opensuse.org/request/show/1038940 it should be there in ALP without extra work.
Better do not install it via this hook, so we test the `chrony` package here as well.
```suggestion

```

tests/microos/networkmanager.pm
```diff
@@ -0,0 +1,130 @@
...
+    die 'wrong DNS-Manager is currently used for dnsmasq' if ($RcManager !~ /symlink/ || $mode !~ /dnsmasq/);
+    ping_check;
+    # systemd-resolved
+    assert_script_run('rm -rf /etc/NetworkManager/conf.d/00-use-dnsmasq.conf');
+    trup_call('pkg install systemd-network');
+    check_reboot_changes;
+    if (script_run('systemctl enable --now systemd-resolved.service') != 0) {
+        record_soft_failure('bsc#1206352  ALP fails to enable service "systemd-resolved.service"');
+        return;
+    }
```
cfconrad, line 113: Thats, cool! But maybe systemd-resolvd isn't something we will use in ALP. @foursixnine is this question already raised somewhere?

cfconrad (changes requested):
Nice done!
Only the thing with chrony should work without extra setup. And I think we should not tweek the test, but test if the pkg provide what we expect, thx!

cfconrad:
> I remember there was some discussion about the usage of `script_run($cmd) == 0` is better to use it with `die_on_timeout => 1`. If there is a timeout, it will return `undef` and make the condition valid, which we don't want.
> I know your commands are not going to timeout cause they are rather simple, but let's try to apply best practices just in case.

Since https://github.com/os-autoinst/os-autoinst/pull/1917 and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14027 for distri opensuse it should be `1` by default. AFIK there is no need to explicit set it to `1`.

tests/microos/networkmanager.pm
foursixnine: @cfconrad Stefan is aware of it, I'll add it to the questions too just to be safe

lib/main_micro_alp.pm
rfan1: sure, will do

tests/microos/networkmanager.pm
```diff
@@ -0,0 +1,115 @@
...
+    systemctl('is-active NetworkManager');
+    assert_script_run('grep "Generated by NetworkManager" /etc/resolv.conf');
+    assert_script_run qq(grep "nameserver $dns_server" /etc/resolv.conf);
+    # DNS-Manager check
+    record_info('default dns bind config');
+    ($RcManager, $mode) = dns_mgr();
+    die 'wrong DNS-Manager is currently used for default' if ($RcManager !~ /symlink/ || $mode !~ /default/);
+    ping_check;
+    record_info('chronyd service check');
+    assert_script_run('chronyc -n sources');
```
cfconrad, line 68: I don't see a check here. I would expect something like `chronyc -n sources | grep $THE_EXPECTED_NTP_IP`.
I also do not see a "local" ntp IP here: https://openqa.opensuse.org/tests/2972458#step/networkmanager/43 maybe the DHCP server you are using doesn't provide something?
But lets take this with a extra step, or?

cfconrad (approved):
LGTM

the NTP check should go in a other PR, or?

tests/microos/networkmanager.pm
rfan1: Good comments, I don't cover NTP check in this test module, just checking the `chronyc` daemon is running there.

Because with current qemu user net configuration, it might not able to provide the NTP server role.

PR #16097, opened 2022-12-14
Fix CI: work around https://github.com/actions/checkout/issues/766

Vogtinator:
I don't know why this isn't fixed at a GitHub level yet, looks rather major.


DimStar77 (approved):
lgtm

PR #16099, opened 2022-12-14
check_failures_in_journal(): Upload system logs when needed

mdoucha:
Reporting detected errors is nice, but it's really hard to debug them when there's no info about **what error** it actually was.

- Related ticket: N/A
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/10526689

### DIFF
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -164,6 +164,12 @@ sub check_failures_in_journal {
         else {
             record_soft_failure("Found new failures: Fake bsc#5555(by PR rule)\n" . $failures . "This is an unknown failure which need to be investigated!");
         }
+
+        my $logfile = "/tmp/journalctl-$machine.log";
+
+        script_run("rm -f $logfile");
+        print_cmd_output_to_file('journalctl -b', $logfile, $machine);
+        upload_logs($logfile);
     }
     return $failures;
 }


mdoucha:
@Julie-CAO Please review.

lib/virt_autotest/utils.pm
```diff
@@ -140,6 +140,8 @@ sub check_failures_in_journal {
         else {
             record_soft_failure("Found new failures: Fake bsc#5555(by PR rule)\n" . $failures . "This is an unknown failure which need to be investigated!");
         }
+
+        upload_system_logs;
```
waynechen55, line 144: I think calling ```upload_system_logs``` here directly is not very good, because ```check_failures_in_journal``` does not serve the logs uploading purpose by design. And ```upload_system_logs``` does not do as expected when it comes to virtual machine. I think you can use it more wisely and also simply as below:
```
upload_system_logs if (check_failures_in_journal returns some failures);
```
This does not only server your purpose, but also keep everything tidy and clear.  Calling ```upload_system_logs``` inside introduces coupling and, at the same time, make the original function very difficult to use and to be called by other functions, especially when there are repeated calls.

waynechen55: Healthy checking let user know when "the error message" firstly occurred, which is the most important. For example, you firstly spot a call trace starts from a certain test module. Then you can postpone logs collecting at the end of the whole test suite by using test module:
```
tests/virt_autotest/validate_system_health.pm
```
This is the original design. ```check_host_health``` and ```check_guest_health``` are in base module which means ```check_failures_in_journal``` is also frequently called at the beginning of each test module chained together. It is not very meaningful for the test to upload the same kind of logs for each module from my POV.

mdoucha: No. Always upload logs *when you check them*. The test can fail long before it reaches `validate_system_health` and then you'll have nothing to debug with, especially when the error is hard to reproduce.

lib/virt_autotest/utils.pm
```diff
@@ -156,6 +156,7 @@ sub check_host_health {
         return 'pass';
     }
     record_info("Unhealthy host", $failures, result => 'fail');
+    upload_system_logs;
```
Julie-CAO, line 159: I can understand that logs are uploaded when there are failures, but for our virtualization tests, it is a different case. The host is checked before each test module, once 'core dump'/'call trace' are found in journal files on the machine, they will be reported in each check in all test modules. fg. https://openqa.suse.de/tests/10188966
In this case, we have 18 times check, it would be meaningless and unrealistic to upload similiar logs 18 times. 

Considering our case, could you come up with a better solution to this problem?

mdoucha: I don't see a problem with having too many logs. Logfiles uploaded by different test modules are prefixed by the test module name so you can immediately see which logfile was uploaded at which point in the job. Then you can quickly review them by looking at the first one and the last one to see whether it's all the same error or different test modules produced additional errors during the job. And then you can check the logfiles in between as needed.

mdoucha: If you insist on collecting logs only for new errors, you can use the `journalctl --cursor-file=...` option to only read journal entries that had not been checked before. However, that is outside the scope of this PR.

waynechen55: As I said in [this comment](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16099#discussion_r1048411471), you can also use it in your way and does not break any others.

Julie-CAO: I think `journalctl --cursor-file=` may be a better solution for both needs. fg.
```
sub check_failures_in_journal {
...
my $cmd = "journalctl --cursor-file=FILE | grep '$warn'";
```
then test only fails once with the first spot of failures. 'FILE' can be removed in the test where you'd like have an entire search.

waynechen55: Using cursor is also option. I also agree if this can be done well.

Julie-CAO:
@mdoucha , I had a PR in progress during your holiday, sorry for not telling you in advance. https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16169

I am on vacation leave today. I'll check how we figure out our two PRs tomorrow.

mdoucha:
It turns out that `--cursor-file=...` is not supported on SLE-12SP5 so the log check has to use `--cursor=...` instead and keep the cursor strings in a variable.

I've rewritten the log upload to work for both host and guests. The last remaining question is whether the `validate_system_health.pm` test module should reset cursor to report all errors, including those already reported by earlier test modules, or report only new errors.

New verification run: https://openqa.suse.de/tests/10275854

Julie-CAO:
Since you had a good solution to compromise for both side, I am going to give up my PR, https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16169.  I did not notice that --cursor-file is not supported on sle12sp5.

The only issue is that `$log = script_output($cmd);` takes 5~15 seconds and output very long journals in autoinst-log.txt(fg. in your verification run, https://openqa.suse.de/tests/10275854/logfile?filename=autoinst-log.txt). Can we just use something like
```
    $cmd .= " -e '$_'" foreach (@warnings);
    $cmd = "ssh root\@$machine " . "\"$cmd\"" if $machine ne 'localhost';
```
to search once with a cursor and without using script_output() ?

About `validate_system_health.pm`, I'll handle it myself with your reset_log_cursor() after the PR is merged. Thanks for considering of it.

mdoucha:
> The only issue is that `$log = script_output($cmd);` takes 5~15 seconds and output very long journals in autoinst-log.txt(fg. in your verification run, https://openqa.suse.de/tests/10275854/logfile?filename=autoinst-log.txt). Can we just use something like

The performance penalty of `script_output()` is unavoidable because we have to read the cursor string either way. Most of the performance cost is spent on typing the command into VNC console because unlike plain `scritp_run()`, the `script_output()` function needs to make several typing steps. The best way to speed that up is to switch to serial terminal. I see the extra output in the logs as a benefit for test failure analysis but feel free to implement your suggestion in your upcoming PR.

> About `validate_system_health.pm`, I'll handle it myself with your reset_log_cursor() after the PR is merged. Thanks for considering of it.

Thanks.

Julie-CAO:
> > The only issue is that `$log = script_output($cmd);` takes 5~15 seconds and output very long journals in autoinst-log.txt(fg. in your verification run, https://openqa.suse.de/tests/10275854/logfile?filename=autoinst-log.txt). Can we just use something like
> 
> The performance penalty of `script_output()` is unavoidable because we have to read the cursor string either way. Most of the performance cost is spent on typing the command into VNC console because unlike plain `scritp_run()`, the `script_output()` function needs to make several typing steps. The best way to speed that up is to switch to serial terminal. I see the extra output in the logs as a benefit for test failure analysis but feel free to implement your suggestion in your upcoming PR.
> 

The journal files are uploaded when there are failures, so the output in autoinst-log.txt is not neccessary. Since you are not objected, I'll try to stop using script_output() in my upcoming PR.

Then the PR is ok to me. @waynechen55 , do you have any comments?

mdoucha:
Rebased on latest master, verification run updated.

foursixnine (approved):

Julie-CAO:
Hi @mdoucha, some of our virtualization tests failed when running commands "journalctl --show-cursor  ...". Despite the it says "curl: could not resolve host grenache-1.openqa.suse.de", acturally previous commands did run into this problem. And if I commentted out the function call to check_failures_in_journal(), test passed.

I am unable to find out the root cause. I suspect "my $log = script_output($cmd);" which might not always work well in OSD, no idea if it is an ipmi backend sol console problem, or the lab relocation problem. Quite a few tests were blocked. fg.

https://openqa.suse.de/tests/10716194#step/guest_migration_src/11
http://10.67.129.218/tests/335#step/libvirt_virtual_network_init/12
https://openqa.suse.de/tests/10716192#step/guest_migration_src/12

Could we revert this PR or could you fix or get around of this blocker?

FYI. @waynechen55 @alice-suse @guoxuguang

Julie-CAO:
> > Hi @mdoucha, some of our virtualization tests failed when running commands "journalctl --show-cursor ...". Despite the it says "curl: could not resolve host grenache-1.openqa.suse.de", acturally previous commands did run into this problem. And if I commentted out the function call to check_failures_in_journal(), test passed.
> 
> The `curl: could not resolve host grenache-1.openqa.suse.de` errors in migration tests appears to be a network failure. Both jobs with this failure have multiple `curl` uploads to the same hostname without any issues: https://openqa.suse.de/tests/10716192#step/update_package/45 https://openqa.suse.de/tests/10716194#step/update_package/48
> 
> Restarted job uploaded the journalctl log successfully: https://openqa.suse.de/tests/10719597#step/guest_migration_src/9

I missed 'not' in my previous words "acturally previous commands did run into this problem." I meant there are quite some uploads/downloads between the SUT and the worker(grenache-1), while "journalctl --show-cursor ..." failed the most times. The network connection or traffic is likely a reason, and the large output of the journal command would make it worse. We are going to make it more stable from our side.

> 
> > http://10.67.129.218/tests/335#step/libvirt_virtual_network_init/12
> 
> The ALP failure appears to be a test issue. The `journalctl` command is called from inside a container where it probably cannot access the system log. You need to either call the command from outside the container, or disable the health check here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/virt_feature_test_base.pm#L59

This is not ALP test, it is SLE. Anyway, thanks.

mdoucha:
> > > http://10.67.129.218/tests/335#step/libvirt_virtual_network_init/12
> > 
> > The ALP failure appears to be a test issue. The `journalctl` command is called from inside a container where it probably cannot access the system log. You need to either call the command from outside the container, or disable the health check here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/virt_feature_test_base.pm#L59
> 
> This is not ALP test, it is SLE. Anyway, thanks.

The scheduled product says ALP. `DISTRI` says ALP. `KVM_WORKLOAD_IMAGE` says ALP. `YAML_SCHEDULE` says ALP. The console welcome message says ALP. Are you sure that http://10.67.129.218/tests/335 is SLE?

mdoucha:
> I missed 'not' in my previous words "acturally previous commands did run into this problem." I meant there are quite some uploads/downloads between the SUT and the worker(grenache-1), while "journalctl --show-cursor ..." failed the most times. The network connection or traffic is likely a reason, and the large output of the journal command would make it worse. We are going to make it more stable from our side.

If the error message says `could not resolve hostname`, then the log file size does not matter, the actual file upload did not even start.

Julie-CAO:
> The scheduled product says ALP. `DISTRI` says ALP. `KVM_WORKLOAD_IMAGE` says ALP. `YAML_SCHEDULE` says ALP. The console welcome message says ALP. Are you sure that http://10.67.129.218/tests/335 is SLE?

You are right, it is ALP. I did not notice it is a new test.

Julie-CAO:
> > I missed 'not' in my previous words "acturally previous commands did run into this problem." I meant there are quite some uploads/downloads between the SUT and the worker(grenache-1), while "journalctl --show-cursor ..." failed the most times. The network connection or traffic is likely a reason, and the large output of the journal command would make it worse. We are going to make it more stable from our side.
> 
> If the error message says `could not resolve host`, then the log file size does not matter, the actual file upload did not even start.

script_output() does not work as well as script_run() as it needs http access. We are trying with setting "type_command=> 1" in script_output() api to avoid unstable network. Let's cross fingers..

Julie-CAO:
with `type_command=> 1` in script_output() works. It does not connect worker's seriel console, it types commands on SUT's serial console directly the way script_run() does. I'll add it to the test script.

Julie-CAO:
> You could also switch to serial terminal using `select_serial_terminal` from `lib/serial_terminal.pm`. That might speed up the test by a factor of 10. But even with type_command, the test will likely just fail on the next `upload_logs()` call. `script_run()` is not usable because you can't get the journalctl cursor token out of it.

Oh, no, thanks. The serial terminal is also having problem after the machines relocation. refer to https://progress.opensuse.org/issues/120651.

Julie-CAO:
> > Oh, no, thanks. The serial terminal is also having problem after the machines relocation. refer to https://progress.opensuse.org/issues/120651.
> 
> The ticket is about VNC console. This is serial terminal: https://openqa.suse.de/tests/10754703

Is there a fix in your github repo to run this test? It is only a branch in your test, I can't see the changes directly like a PR in your test.

PR #16100, opened 2022-12-14
Extend timeout for booting up on aarch64

Zaoliang:
we have sometimes timeout issue while system is booting up on aarch64. 
see https://progress.opensuse.org/issues/120513
extend timeout to prevent this failure.
VR:
http://openqa.suse.de/tests/10165744#next_previous
https://openqa.opensuse.org/tests/2926919 (ppc64le)
https://openqa.opensuse.org/tests/2926919 (x86_64)


rfan1 (approved):
LGTM

PR #16102, opened 2022-12-14
Use serial terminal in transactional tests

mdoucha:
Allow running `trup_call()` on serial terminals and make the commands non-interactive by default. Tests which validate interactive commands must call the function with `interactive => 1`.

This PR fixes a testing issue on s390x where bogus Btrfs messages triggered by `snapper` pollute test command output and confuse the test.

- Related ticket: N/A
- Needles: N/A
- Verification runs:
  - x86_64: https://openqa.suse.de/tests/10272686
  - s390x: https://openqa.suse.de/tests/10269051
  - aarch64: https://openqa.suse.de/tests/10269050

### DIFF
--- a/lib/transactional.pm
+++ b/lib/transactional.pm
@@ -14,7 +14,7 @@ use Exporter;
 
 use strict;
 use warnings;
-use testapi;
+use testapi qw(is_serial_terminal :DEFAULT);
 use utils;
 use Carp;
 use microos 'microos_reboot';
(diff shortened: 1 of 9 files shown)


mdoucha:
> Since no tests seem to be using the interactive install failure check, force all commands to be non-interactive to make the code simpler.

Hmm, looks like I've misunderstood `trup_call "cleanup up", exit_code => 1;`... The non-interactive tests still checks the exit code but the `Abort, retry, ignore?` prompt will not be checked. Do we need the prompt check?

Vogtinator:
> > Since no tests seem to be using the interactive install failure check, force all commands to be non-interactive to make the code simpler.
> 
> Hmm, looks like I've misunderstood `trup_call "cleanup up", exit_code => 1;`... The non-interactive tests still checks the exit code but the `Abort, retry, ignore?` prompt will not be checked. Do we need the prompt check?

In non-interactive mode the prompt should still be shown, just without the input handling. So that could still be verified.

mdoucha:
> In non-interactive mode the prompt should still be shown, just without the input handling. So that could still be verified.

Yes, but that would require emulating `script_run()` inside `trup_call()` like in the original version which I'd rather avoid.

Vogtinator:
> > In non-interactive mode the prompt should still be shown, just without the input handling. So that could still be verified.
> 
> Yes, but that would require emulating `script_run()` inside `trup_call()` like in the original version which I'd rather avoid.

You could use `script_output` in `transactional_update.pm` directly.

mdoucha:
> You could use `script_output` in `transactional_update.pm` directly.

Given the performance penalty in VNC console and test output readability issues, I'd rather implement generic support for interactive console scripts. Something that would be used like this:

    script_run_interactive('some --command');
    type_string("foo\n");
    wait_serial(...);
    type_string("bar\n");
    wait_serial(...);
    finish_interactive();

pdostal (approved):
This is super cool!

grisu48 (approved):

czerw (approved):

mloviska (approved):
Since the dialogue is not going to be tested any more, I guess we should come up with a specific interactive test, in order not to lose this check. WDYT @grisu48, @Vogtinator?

mdoucha:
> Since the dialogue is not going to be tested any more, I guess we should come up with a specific interactive test, in order not to lose this check. WDYT @grisu48, @Vogtinator?

I will rewrite this PR to allow interactive `trup_call()` commands.

mdoucha:
I've implemented a pair of helper functions for testing interactive console commands. `trup_call()` now uses them if you call it with `interactive => 1` and handles the Continue/Abort prompts from zypper. I've also removed redundant `-n` parameters from some existing `trup_call()` uses.

One of the install commands in `transactional_update` module now runs again in interactive mode but there's no test for the failure case.

mdoucha:
All verification runs have passed. Please review again and merge.

Interactive install test: https://openqa.suse.de/tests/10272686#step/transactional_update/158

grisu48:
VR for SLEM 5.3 on GCE: https://duck-norris.qe.suse.de/tests/11627

grisu48:
4 approvals and no new comments or reviews. I'm proceeding to merge this PR. Thank you Martin, this is great! :slightly_smiling_face:

grisu48:
> 4 approvals but all of them before the PR went through a major rewrite...

Ah yes, the approvals are not being removed by a force pull. I forgot, my bad, sorry.

PR #16116, opened 2022-12-16
Add AWS configuration when running on EC2 and wait for ssh

alvarocarvajald:
Regression tests of [qe-sap-deployments](https://github.com/SUSE/qe-sap-deployment) in openQA are failing on [EC2](http://openqaworker15.qa.suse.cz/tests/104272#) because the ansible commands over SSH are called before the cloud infrastructure is ready for SSH connections and because of a missing `~/.aws/config` file referenced by the ansible playbooks.

This PR fixes both issues by adding new methods into `lib/qesapdeployment.pm` to create the configuration files required by AWS/EC2 and to get the public IP addresses from the terraform deployment so the module can wait until SSH is ready.

- Related ticket: N/A
- Needles: N/A
- Verification run: [Azure SLES for SAP 15-SP3](http://openqaworker15.qa.suse.cz/t104320), [AWS/EC2 SLES for SAP 15-SP3](http://openqaworker15.qa.suse.cz/t104328), [Azure SLES for SAP 12-SP5](http://openqaworker15.qa.suse.cz/t104322)

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -30,12 +30,12 @@ package qesapdeployment;
 use strict;
 use warnings;
 use Carp qw(croak);
+use Mojo::JSON qw(decode_json);
 use YAML::PP;
-use utils 'file_content_replace';
+use utils qw(file_content_replace zypper_call);
+use publiccloud::utils qw(get_credentials);
 use testapi;
 use Exporter 'import';
-use utils 'zypper_call';
-
 
 my @log_files = ();
 
@@ -59,11 +59,15 @@ our @EXPORT = qw(
   qesap_ansible_cmd
   qesap_ansible_script_output
   qesap_create_ansible_section
+  qesap_create_aws_credentials
+  qesap_create_aws_config
+  qesap_remote_hana_public_ips
 );
 
 =head1 DESCRIPTION
 
     Package with common methods and default or constant  values for qe-sap-deployment
+
 =head2 Methods
 
 
@@ -370,6 +374,13 @@ sub qesap_prepare_env {
     push(@log_files, "$paths{deployment_dir}/ansible/playbooks/vars/hana_media.yaml");
     my $hana_vars = "$paths{deployment_dir}/ansible/playbooks/vars/hana_vars.yaml";
     my $exec_rc = qesap_execute(cmd => 'configure', verbose => 1);
+
+    if (check_var('PUBLIC_CLOUD_PROVIDER', 'EC2')) {
+        my $data = get_credentials('aws.json');
+        qesap_create_aws_config();
+        qesap_create_aws_credentials($data->{access_key_id}, $data->{secret_access_key});
+    }
+
     push(@log_files, $hana_vars) if (script_run("test -e $hana_vars") == 0);
     qesap_upload_logs(failok => 1);
     die("Qesap deployment returned non zero value during 'configure' phase.") if $exec_rc;
@@ -478,6 +489,49 @@ sub qesap_ansible_script_output {
     my $output = script_output("cat $local_tmp");
     enter_cmd "rm $local_tmp";
     return $output;
+}
+
+=head3 qesap_create_aws_credentials
+
+    Creates a AWS credentials file as required by QE-SAP Terraform deployment code.
+=cut
+
+sub qesap_create_aws_credentials {
+    my ($key, $secret) = @_;
+    my %paths = qesap_get_file_paths();
+    my $credfile = script_output q|awk -F\" '/aws_credentials/ {print $2}' | . $paths{qesap_conf_trgt};
+    save_tmp_file('credentials', "[default]\naws_access_key_id = $key\naws_secret_access_key = $secret\n");
+    assert_script_run 'mkdir -p ~/.aws';
+    assert_script_run 'curl ' . autoinst_url . "/files/credentials -o $credfile";
+    assert_script_run "cp $credfile ~/.aws/credentials";
+}
+
+=head3 qesap_create_aws_config
+
+    Creates a AWS config file in ~/.aws as required by the QE-SAP Terraform & Ansible deployment code.
+=cut
+
+sub qesap_create_aws_config {
+    my %paths = qesap_get_file_paths();
+    my $region = script_output q|awk -F\" '/aws_region/ {print $2}' | . $paths{qesap_conf_trgt};
+    $region = get_required_var('PUBLIC_CLOUD_REGION') if ($region =~ /^%.+%$/);
+    save_tmp_file('config', "[default]\nregion = $region\n");
+    assert_script_run 'mkdir -p ~/.aws';
+    assert_script_run 'curl ' . autoinst_url . "/files/config -o ~/.aws/config";
+}
+
+=head3 qesap_remote_hana_public_ips
+
+    Return a list of the public IP addresses of the systems deployed by qesapdeployment, as reported
+    by C<terraform output>. Needs to run after C<qesap_execute(cmd => 'terraform');> call.
+
+=cut
+
+sub qesap_remote_hana_public_ips {
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $tfdir = qesap_get_terraform_dir($prov);
+    my $data = decode_json(script_output "terraform -chdir=$tfdir output -json");
+    return @{$data->{hana_public_ip}->{value}};
 }
 
 1;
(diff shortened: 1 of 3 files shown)


lpalovsky (approved):
LGTM

lib/qesapdeployment.pm
```diff
@@ -480,4 +485,50 @@ sub qesap_ansible_script_output {
...
 
+=head3 qesap_create_aws_credentials
+
+    Creates a AWS credentials file as required by QE-SAP Terraform deployment code.
+=cut
+
+sub qesap_create_aws_credentials {
+    my ($key, $secret) = @_;
+    my %paths = qesap_get_file_paths();
+    my $credfile = script_output q|awk -F\" '/aws_credentials/ {print $2}' | . $paths{qesap_conf_trgt};
```
mpagot, line 496: is this to create a new path, starting from an existing one from the "paths catalogue" in **qesap_get_file_paths** ? If it is so, why not just to add the new path within **qesap_get_file_paths** in a more explicit way?

lib/qesapdeployment.pm
```diff
@@ -480,4 +485,50 @@ sub qesap_ansible_script_output {
...
+    save_tmp_file('credentials', join(' ', "[default]\naws_access_key_id", '=', $key,
```
mpagot, line 497: I propose to change it in `join("\n"` and code each file line for each join element. It is to keep the code more readable.

lib/qesapdeployment.pm
```diff
@@ -480,4 +485,50 @@ sub qesap_ansible_script_output {
...
+}
+
+=head3 qesap_create_aws_config
+
+    Creates a AWS config file in ~/.aws as required by the QE-SAP Terraform & Ansible deployment code.
+=cut
+
+sub qesap_create_aws_config {
+    my %paths = qesap_get_file_paths();
+    my $region = script_output q|awk -F\" '/aws_region/ {print $2}' | . $paths{qesap_conf_trgt};
```
mpagot, line 511: I propose to name the variable: `$region_file`

mpagot, line 511: Seems that all tests are able to get the region, why not to change qesap_create_aws_config to directly accept the region as argument?

lib/qesapdeployment.pm
```diff
@@ -480,4 +485,50 @@ sub qesap_ansible_script_output {
...
+=head3 qesap_remote_hana_public_ips
+
+    Return a list of the public IP addresses of the systems deployed by qesapdeployment, as reported
+    by C<terraform output>. Needs to run after C<qesap_execute(cmd => 'terraform');> call.
+
+=cut
+
+sub qesap_remote_hana_public_ips {
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    $prov = uc($prov) eq 'EC2' ? 'aws' : (uc($prov) eq 'GCE' ? 'gcp' : $prov);
```
mpagot, line 526: Is it needed? We have https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/ab18a6421250ba5f7f6a860c02961b273c415747/lib/qesapdeployment.pm#L221 to avoid the translation.

t/09_qesapdeployment.t
```diff
@@ -406,4 +406,74 @@ subtest '[qesap_ansible_script_output] root' => sub {
...
+    ok((any { qr/eu-central-1/ } @calls), 'AWS Region matches');
+    is $contents[0], 'config', "AWS config file: config is the expected value and got $contents[0]";
+    like $contents[1], qr/region = eu-central-1/, "Expected region eu-central-1 is in the config file";
+};
+
+subtest '[qesap_remote_hana_public_ips]' => sub {
+    my $qesap = Test::MockModule->new('qesapdeployment', no_auto => 1);
+    my @calls;
+
+    $qesap->redefine(get_required_var => sub { return 'EC2'; });
```
mpagot, line 466: you do not need to do it, just call `set_var` (at least it is the approach used in other UT in this repo)

Take care to also set to undef before the end of the test 
```
set_var('YOUR_VAR', undef);
```

mpagot (approved):
LGTM

lib/qesapdeployment.pm
alvarocarvajald: > is this to create a new path, starting from an existing one from the "paths catalogue" in **qesap_get_file_paths** ? 

No. This is to create the credentials file in the path specified by the the YAML configuration file. 

> If it is so, why not just to add the new path within **qesap_get_file_paths** in a more explicit way?

Because this way ensures that if you run the test with a YAML configuration pointing to a location different than any default or anything specified on the settings, then the file will be created where both terraform and ansible are looking for it as the YAML configuration feeds the values for the ansible playbooks and the terraform configuration.

lib/qesapdeployment.pm
alvarocarvajald: That way the file will lack a newline at the end, which I guess it could be non-essential. I can change it to:
```
save_tmp_file('credentials', join("[default]\naws_access_key_id = $key\naws_secret_access_key = $secret\n"));
```
If you feel it's more readable.

lib/qesapdeployment.pm
alvarocarvajald: Same issue as with the credentials file: tests can have one region on a setting and another in the YAML configuration file. With this I'm just making sure that the `~/.aws/config` file generated by the method has the one that's actually used by terraform and ansible. If we decide to always replace the region in the YAML configuration file with whatever's in the **PUBLIC_CLOUD_REGION** setting, then we can change the above line to:
```
my $region = get_required_var('PUBLIC_CLOUD_REGION');
```

lib/qesapdeployment.pm
alvarocarvajald: Oh, forgot about that. I'll drop the line.

t/09_qesapdeployment.t
alvarocarvajald: Will change it.

lib/qesapdeployment.pm
mpagot: Probably not

lib/qesapdeployment.pm
alvarocarvajald: Taking this into consideration, I'm adding an additional check in the function.

mpagot: we does not agree to always replace REGION in the yaml with exactly  PUBLIC_CLOUD_REGION, but we always replace it with something. We always have REGION controlled by the test code. But maybe I get your point: "As the only source of truth for the qe-sap-deployment is the config.yaml, and as the qe-sap-deployment expect some aws files to be in a specific disk place openQA test code should also use conf.yaml as source of truth to feel the gap and create the files." If it is so, maybe we can think to move it in the qesap glue script, the write, or at least the check, of the needed files.

tests/sles4sap/qesapdeployment/deploy.pm
```diff
@@ -7,14 +7,44 @@
...
+        else {
+            return $duration;
+        }
+        sleep 5;
+    }
+    die 'Timed out while waiting for ssh to be available in the CSP instances';
+}
+
 sub run {
+    if (check_var('PUBLIC_CLOUD_PROVIDER', 'EC2')) {
```
mpagot, line 36: can we think about moving lines 36-39 in qesap_prepare_env of lib/qesapdeploy.pm ?

lib/qesapdeployment.pm
alvarocarvajald: I changed it anyway. Liked it more after the fact :P

mpagot (approved):

lib/qesapdeployment.pm
alvarocarvajald: This did not work: https://openqaworker15.qa.suse.cz/tests/105671#step/deploy/21

I'm adding the assignment with the ternary operator again.

alvarocarvajald: Found a way to remove the line. Done.

tests/sles4sap/qesapdeployment/deploy.pm
alvarocarvajald: Done!

Verification runs:

https://openqaworker15.qa.suse.cz/tests/105697
https://openqaworker15.qa.suse.cz/tests/105698
https://openqaworker15.qa.suse.cz/tests/105699

alvarocarvajald:
I think I'm done with this PR. Feel free to merge if no more changes are required.

lib/qesapdeployment.pm
mpagot: Does it make sense to add a comment in the code like...?

```
# get the region from the conf.yaml or from a job setting if not present there
```

Is it the intent of this two lines?

PR #16118, opened 2022-12-19
Fix publiccloud/check_registercloudguest /var/log/cloudregister check

pdostal:
The `/var/log/cloudregister` on BYOS must by empty or missing but this check is broken after 25a3a7e (#16062).

- Related ticket: [poo#122134](https://progress.opensuse.org/issues/122134)
- Failing example: [openqa.suse.de/t10205814](https://openqa.suse.de/tests/10205814#step/check_registercloudguest/46)
- Verification run:


tests/publiccloud/check_registercloudguest.pm
```diff
@@ -56,7 +56,7 @@ sub run {
             die('regionsrv-enabler-azure.timer is not enabled');
         }
 
-        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister') != 0) {
+        if ($instance->ssh_script_output(cmd => 'sudo stat --printf="%s" /var/log/cloudregister', proceed_on_failure => 1) != 0) {
```
grisu48, line 59: This is very nasty but it seems to work. What about the following? `test -s` returns 0 if the file exists and is nonempty. Like this we don't rely on the output of a command that results in an error.

```suggestion
        if ($instance->ssh_script_run(cmd => 'sudo test -s /var/log/cloudregister') == 0) {
```

pdostal: Sounds cool

asmorodskyi (approved):
LGTM

grisu48 (approved):

PR #16125, opened 2022-12-20
Enable the typos utility Github action

pdostal:
There is `_typos.toml` file where I correct all the false-positives and excluude all the third-party of autogenerated files.

Link to the project: [crate-ci/typos](https://github.com/crate-ci/typos)

### DIFF
--- a/tools/detect_unused_modules
+++ b/tools/detect_unused_modules
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-usage="Detect test modules usage in the schedules, inluding main.pm and yaml schedules.
+usage="Detect test modules usage in the schedules, including main.pm and yaml schedules.
 
 Usage: $(basename "$0") { -a | -m MODULES } | -h
 
--- a/tools/detect_unused_modules
+++ b/tools/detect_unused_modules
@@ -38,7 +38,7 @@ done
 for test in $MODULES ; do
     # Don't check non-existing modules it as it was either renamed or removed
     # also check only for tests in tests/ directory
-    git ls-files "*$test*" | grep -q ^tests/ || continue
+    git ls-files "*$test*" | grep -qE '^tests/.*\.pm' || continue
 
     echo "check $test" >&2
 
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,0 +1,13 @@
+name: Typos
+on: [push, pull_request]
+
+jobs:
+  run:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v2
+
+    - name: Check spelling of the entire repository
+      uses: crate-ci/typos@master
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ Redmine project. Look for tickets with [easy] or [easy-hack] tags.
 
 ## How to get this repository working
 
-Upon setting up a new openQA instance, it's also necessary to install some aditional dependencies that are inherent to this repository,
+Upon setting up a new openQA instance, it's also necessary to install some additional dependencies that are inherent to this repository,
 for which there are two ways:
 
 * In case you're using cpanm (with or without local::lib, or others), from within the working copy: call `cpanm -n --mirror http://no.where/ --installdeps . `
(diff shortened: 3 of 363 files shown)


lib/publiccloud/k8sbasetest.pm
```diff
@@ -51,7 +51,7 @@ sub get_k8s_service_name {
         return "GKE";
     }
     elsif ($provider eq 'AZURE') {
-        return "AKS";
```
olblak, line 54: While I don't know the context here, it seems to me that AKS wasn't a typo

data/console/movies.redis
```diff
@@ -92,7 +92,7 @@ HSET "movie:92" title "Honeytrap" genre "Crime" votes 325 rating 5.6 release_yea
 HSET "movie:93" title "X+Y" genre "Drama" votes 19609 rating 7.2 release_year 2014 
 HSET "movie:94" title "Creep" genre "Horror" votes 14696 rating 6.2 release_year 2014 
 HSET "movie:95" title "The Drop" genre "Crime" votes 109038 rating 7.1 release_year 2014 
-HSET "movie:96" title "Ich seh ich seh" genre "Drama" votes 22916 rating 6.7 release_year 2014 
+HSET "movie:96" title "Ich she ich she" genre "Drama" votes 22916 rating 6.7 release_year 2014
```
okurz, line 95: I suggest to also exclude this file

data/containers/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # baseimage_var is a placeholder of your base image
-# use base variable when you call test_containered_app
+# use base variable when you call test_contained_app
```
okurz, line 2: wrong

data/hpc/installcheck_hpc_module.sh
```diff
@@ -14,7 +14,7 @@ ftp://dist.suse.de/ibs/SUSE/Products/SLE-Module-Web-Scripting/12/${arch}/product
 base="ftp://openqa.suse.de/SLE-${version}-SERVER-POOL-${arch}-Media1-CURRENT/repodata/*-primary.xml.gz"
 sdk="ftp://openqa.suse.de/SLE-${version}-SDK-POOL-${arch}-Media1-CURRENT/repodata/*-primary.xml.gz"
 
-checkd=$(mktemp check-XXXXX)
+checked=$(mktemp check-XXXXX)
```
okurz, line 17: wrong

data/ldap/test.ldif
```diff
@@ -5,7 +5,7 @@ member: cn=Puca Almeida,ou=Green Corporation,ou=People,dc=green,dc=com
 member: cn=Fredo Almeida,ou=LAB Technical,ou=People,dc=green,dc=com
 member: cn=Otavio Augustus,ou=Green Corporation,ou=People,dc=green,dc=com
 member: cn=Pedro de Lara,ou=Green Corporation,ou=People,dc=green,dc=com
-member: cn=Gugu Liberato,ou=LAB Technical,ou=People,dc=green,dc=com
+member: cn=Gugu Liberation,ou=LAB Technical,ou=People,dc=green,dc=com
```
okurz, line 8: wrong

lib/publiccloud/k8sbasetest.pm
pdostal: Hello @olblak, true. I'm still working on this. There will be lot of such cases.

okurz (changes requested):
good idea but too many wrong changes right now

tools/detect_unused_modules
```diff
@@ -38,7 +38,7 @@ done
 for test in $MODULES ; do
     # Don't check non-existing modules it as it was either renamed or removed
     # also check only for tests in tests/ directory
-    git ls-files "*$test*" | grep -q ^tests/ || continue
+    git ls-files "*$test*" | grep -qE '^tests/.*\.pm' || continue
```
okurz, line 41: this looks like an unwanted change

okurz (changes requested):
After skimming over the content I found one issue https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16125#discussion_r1059915718 which seems to be a conflict resolution gone wrong.

In general I like the approach to fix and prevent spelling mistakes. However in general I prefer if we use tooling which is less reliant on github actions for two reasons:
1. Contributors should be able to detect and fix spelling mistakes *before* they open a pull request
2. We should prevent a "vendor lock-in" which is making any switch of project repository or CI harder

Of course one can still call a spell checker as part of github actions but it would be preferred to just call an application directly which also contributors can call instead of just a github action

baierjan (changes requested):
Indeed, it is a good idea. After all, it was also able to find and fix one real bug.

On the other hand, it still makes a lot of false-positives some of them pretty dangerous with serious impact. I tried to point some of the controversial suggestions.

pdostal:
crate-ci/typos#642, crate-ci/typos#643, crate-ci/typos#316

d3flex (changes requested):

PR #16126, opened 2022-12-20
Add HanaSR master tests

lpalovsky:
Extends HANA SR coverage with modues and libraries for testing primary 
database (stop, crash, kill).

- Related ticket: https://jira.suse.com/browse/TEAM-7079
- Needles: N/A
- Verification run: http://openqaworker15.qa.suse.cz/tests/107087

### DIFF
--- a/lib/sles4sap_publiccloud.pm
+++ b/lib/sles4sap_publiccloud.pm
@@ -29,9 +29,14 @@ our @EXPORT = qw(
   get_promoted_hostname
   is_hana_resource_running
   stop_hana
+  start_hana
+  check_takeover
   get_replication_info
   is_hana_online
   get_hana_topology
+  enable_replication
+  cleanup_resource
+  get_promoted_instance
   wait_for_sync
   wait_for_pacemaker
   cloud_file_content_replace
@@ -45,7 +50,6 @@ our @EXPORT = qw(
     All commands are executed through C<sudo>.
     If 'runas' defined, command will be executed as specified user,
     otherwise it will be executed as root.
-
 =cut
 
 sub run_cmd {
@@ -128,7 +132,6 @@ sub get_hana_topology {
     is_hana_online([timeout => 120, wait_for_start => 'false']);
 
     Check if hana DB is online. Define 'wait_for_start' to wait for DB to start.
-
 =cut
 
 sub is_hana_online {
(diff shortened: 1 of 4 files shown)


lib/sles4sap_publiccloud.pm
```diff
@@ -215,6 +218,97 @@ sub stop_hana {
...
 
+=head2 start_hana
+    start_hana([timeout => 60]);
+
+    Start HANA DB using "HDB start" command
+=cut
+
+sub start_hana {
+    my ($self) = @_;
+    $self->run_cmd(cmd => "HDB start", runas => get_var("SAP_SIDADM"));
```
mpagot, line 229: Does `SAP_SIDADM` needs to be configurable? I think that it is hardcoded somewhere in qe-sap-deployment Ansible code. So we need to take care to have a job setting to match to an Ansible hardcoded value in qe-sap-deployment. Do we need to make it configurable in qe-sap-deployment (like a conf.yaml variable) too?

lib/sles4sap_publiccloud.pm
```diff
@@ -215,6 +218,97 @@ sub stop_hana {
...
+}
+
+=head2 cleanup_resource
+    cleanup_resource([timeout => 60]);
+
+    Cleanup rsource 'msl_SAPHana_*', wait for DB start automaticlly.
```
mpagot, line 235: `rsource` typo

schedule/sles4sap/publiccloud_hanasr.yml
```diff
@@ -12,6 +12,6 @@ schedule:
   - boot/boot_to_desktop
   - sles4sap/publiccloud/qesap_terraform
   - sles4sap/publiccloud/qesap_ansible
-  # - sles4sap/publiccloud/hana_sr_schedule_primary_tests
+  - sles4sap/publiccloud/hana_sr_schedule_primary_tests
   - sles4sap/publiccloud/hana_sr_schedule_replica_tests
   - sles4sap/publiccloud/hana_sr_schedule_cleanup
```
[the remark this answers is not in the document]
lpalovsky: No. "hana_sr_takeover" is scheduled by "hana_sr_schedule_primary_tests" .  The purpose is scheduling one module multiple times with different parameters avoiding having multiple modules with very similar code.

lib/sles4sap_publiccloud.pm
lpalovsky: In this case "SAP_SIDADM" is defined internally inside openQA deployment module and is required for the testcode, not the deployment project. 
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/7dd0fc351b578477941cc11b0b656fa25caf9410/tests/sles4sap/publiccloud/qesap_terraform.pm#L98

But I should change it to get_required_var just in case if the deployment was done in some other way. 
Or just change it to "lc(get_var("INSTANCE_SID") . "adm")"

lib/sles4sap_publiccloud.pm
lpalovsky: Thanks. Fixed

dzedro (approved):

lib/sles4sap_publiccloud.pm
```diff
@@ -211,8 +214,99 @@ sub stop_hana {
         return ();
     }
     else {
-        $self->run_cmd(cmd => $cmd, runas => get_var("SAP_SIDADM"), timeout => $timeout);
+        $self->run_cmd(cmd => $cmd, runas => get_required_var("SAP_SIDADM"), timeout => $timeout);
+    }
+}
+
+=head2 start_hana
```
mpagot, line 221: Who is calling this new function?

lpalovsky: Yep you are right. This one is not used anymore since "crm resource cleanup" starts the hana instance.

PR #16127, opened 2022-12-20
Set ncurses option for spvm yast tests

sofiasyria:
Ppc test fails due to neglecting spvm machines to be specifies to use ncurses.
https://openqa.suse.de/tests/10197089
https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/21ec86201becbcbe92c1d3bf44c2242b1f8e9255

### DIFF
--- a/lib/y2lan_restart_common.pm
+++ b/lib/y2lan_restart_common.pm
@@ -20,7 +20,7 @@ use Exporter 'import';
 use testapi;
 use utils 'systemctl';
 use version_utils qw(is_sle is_leap);
-use Utils::Backends 'is_pvm_hmc';
+use Utils::Backends qw(is_pvm_hmc is_spvm);
 use y2_module_basetest qw(accept_warning_network_manager_default is_network_manager_default);
 use y2_module_consoletest;
 use Test::Assert ':all';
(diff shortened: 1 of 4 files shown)


lib/y2lan_restart_common.pm
```diff
@@ -417,7 +417,7 @@ sub open_yast2_lan {
       = ($options{ui} =~ /ncurses/)
       ? "--" . $options{ui}
       : "";
-    $y2_opts = "--ncurses" if is_pvm_hmc();
+    $y2_opts = "--ncurses" if (is_pvm_hmc() or is_spvm());
```
jknphy, line 420: we usually use `||` instead of `or` for the mess that involve to use it as it has lower precedence, probably fine for this case but we could change it in case this condition would grow somehow and commas would added on the right...seems to have issues with that i particular (but not expert on this :) )

jknphy:
btw, not forget to set the label for `qe-yam` doing it for you.

lib/y2lan_restart_common.pm
sofiasyria: Ahh... I'm starting to remember ..hehe, done

PR #16133, opened 2022-12-21
Public Cloud: SSHD: AllowTcpForwarding in Hardened Azure images

pdostal:
- Another solution: #16160 (more preferred)
- Related merge requests: [qa-maintenance/metadata!750](https://gitlab.suse.de/qa-maintenance/metadata/-/merge_requests/750), [qac/qac-openqa-yaml!1049](https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1049)
- Verification run: [pdostal-server.suse.cz](https://pdostal-server.suse.cz/tests/overview?build=&version=15-SP4&distri=sle&groupid=1) (The link contains also LTP, NFS and XFS tests but those will not be scheduled yet.)

Merge together with the other linked pull requests.

### DIFF
--- a/lib/publiccloud/ssh_interactive.pm
+++ b/lib/publiccloud/ssh_interactive.pm
@@ -144,6 +144,7 @@ sub prepare_ssh_tunnel {
 
     # Permit root passwordless login over SSH
     $instance->ssh_assert_script_run('sudo sed -i "s/PermitRootLogin no/PermitRootLogin prohibit-password/g" /etc/ssh/sshd_config');
+    $instance->ssh_assert_script_run('sudo sed -iE "/^AllowTcpForwarding/c\AllowTcpForwarding yes" /etc/ssh/sshd_config') if (get_var('FLAVOR') =~ 'EC2' && get_var('FLAVOR') =~ 'Hardened');
     $instance->ssh_assert_script_run('sudo systemctl reload sshd');
 
     # Copy SSH settings for remote root


lib/publiccloud/ssh_interactive.pm
```diff
@@ -144,6 +144,7 @@ sub prepare_ssh_tunnel {
 
     # Permit root passwordless login over SSH
     $instance->ssh_assert_script_run('sudo sed -i "s/PermitRootLogin no/PermitRootLogin prohibit-password/g" /etc/ssh/sshd_config');
+    $instance->ssh_assert_script_run('sudo sed -i "s/#AllowTcpForwarding yes/AllowTcpForwarding yes/g" /etc/ssh/sshd_config');
```
grisu48, line 147: I would recommend to match for lines containing anything `AllowTcpForwarding ` and replace the whole line, e.g.

```suggestion
    $instance->ssh_assert_script_run('sudo sed -i "/AllowTcpForwarding /c\AllowTcpForwarding yes" /etc/ssh/sshd_config');
```

jlausuch: I agree. Shouldn't it be `'s/.*AllowTcpForwarding.*/AllowTcpForwarding yes/'` ?

grisu48: Agreed to move forward and do the improvement later.

grisu48 (approved):

lib/publiccloud/ssh_interactive.pm
asmorodskyi, line 147: can we limit this to hardended images only ?

grisu48: I've updated my suggestion. This now matches comment and non-comment values alike, while it does not touch lines, that don't start with `AllowTcpForwarding` or `#AllowTcpForwarding`.

This should make the test runs a bit more stable, as we are not relying on the exact match.

grisu48: Why? This is a necessary requirement for our test runs everywhere and the usage of `sed` only alters this if needed.

asmorodskyi: because currently this requirement is coming from image . and without this change being merged when image for whatever reason change this our test will fail and investigation of this failure will end up on product bug. If we will accept this change as is than such change in image will pass through our tests not noticed

asmorodskyi: let's keep in mind that we speaking about *test*  code not some production grade server app . So "more stable" code often translated into "test which not catching errors" . I think here exactly this case . Only hardended images suppose to have this line commented all other images suppose to have it un-commented without test code involvement

grisu48: Yes, that makes sense - Let's limit the change to hardened images only!

lib/publiccloud/ssh_interactive.pm
```diff
@@ -144,6 +145,7 @@ sub prepare_ssh_tunnel {
 
     # Permit root passwordless login over SSH
     $instance->ssh_assert_script_run('sudo sed -i "s/PermitRootLogin no/PermitRootLogin prohibit-password/g" /etc/ssh/sshd_config');
+    $instance->ssh_assert_script_run('sudo sed -iE "/^[\#]\?AllowTcpForwarding/c\AllowTcpForwarding yes\" /etc/ssh/sshd_config') if (is_ec2() && is_hardened());
```
asmorodskyi, line 148: why we doing this only for *Amazon* hardended images ? is this setting is set by default for Azure and Google but not for Amazon ?

grisu48 (approved):

lib/publiccloud/ssh_interactive.pm
pdostal: True.

lib/publiccloud/ssh_interactive.pm
```diff
@@ -144,6 +144,7 @@ sub prepare_ssh_tunnel {
 
     # Permit root passwordless login over SSH
     $instance->ssh_assert_script_run('sudo sed -i "s/PermitRootLogin no/PermitRootLogin prohibit-password/g" /etc/ssh/sshd_config');
+    $instance->ssh_assert_script_run('sudo sed -iE "/^AllowTcpForwarding/c\AllowTcpForwarding yes" /etc/ssh/sshd_config') if (get_var('FLAVOR') =~ 'EC2' && get_var('FLAVOR') =~ 'Hardened');
```
asmorodskyi, line 147:
```suggestion
    $instance->ssh_assert_script_run('sudo sed -iE "/^AllowTcpForwarding/c\AllowTcpForwarding yes" /etc/ssh/sshd_config') if (is_ec2() && is_hardended());
```

asmorodskyi:
other than that - LGTM
P.S. I like regex , it took some googling to understand it which makes it even more attractive :)

lib/publiccloud/ssh_interactive.pm
grisu48: We don't have `is_hardended()` anymore after Pavel encountered some nasty Perl errors.

asmorodskyi: ok

asmorodskyi: let's just merge it :)

pdostal: No. It should be done for all hardened images, not just for EC2.
There is also better solution in #16160

PR #16135, opened 2022-12-21
Add zypper ref before installing cockpit

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/122029

VR: https://openqa.opensuse.org/tests/2968916

### DIFF
--- a/tests/microos/cockpit_service.pm
+++ b/tests/microos/cockpit_service.pm
@@ -49,6 +49,8 @@ sub run {
 
     if (@pkgs) {
         record_info('TEST', 'Installing Cockpit\'s Modules...');
+        # In ALP, we need to refresh the metadata. poo#122029
+        assert_script_run('zypper ref') if is_alp;
         trup_call("pkg install @pkgs", timeout => 480);
         check_reboot_changes;
     }


jlausuch:
@ggardet FYI, just trying an idea from @okurz to see if metadata refresh helps with this random installation issue.

tests/microos/cockpit_service.pm
```diff
@@ -49,6 +49,7 @@ sub run {
 
     if (@pkgs) {
         record_info('TEST', 'Installing Cockpit\'s Modules...');
+        assert_script_run('zypper ref');
```
okurz, line 52: I suggest to at least add a comment here that the explicit zypper ref might help to work around some refresh problems. *If* it helps then I would still regard it as a product regression and should be reported as such.

jlausuch:
I've run some VRs running only cockpit module and all are green. Maybe we can give it a shot and see how this behaves during next builds. 
https://openqa.opensuse.org/tests/2968953
https://openqa.opensuse.org/tests/2968975
https://openqa.opensuse.org/tests/2968976
https://openqa.opensuse.org/tests/2968977
https://openqa.opensuse.org/tests/2968978

ggardet (approved):
LGTM. Thanks!

jknphy:
could be needed for ALP same configuration that is than for JeOS in previous step?
https://openqa.opensuse.org/tests/2956621/modules/libzypp_config/steps/1/src

jlausuch:
> could be needed to exclude for ALP same configuration that is excluded for JeOS aarch in previous step? https://openqa.opensuse.org/tests/2956621/modules/libzypp_config/steps/1/src

Could you explain why? In any case, I guess that would be for a different PR.

grisu48 (approved):

jlausuch:
Merging. Let's see how this goes.

jknphy:
> > could be needed to exclude for ALP same configuration that is excluded for JeOS aarch in previous step? https://openqa.opensuse.org/tests/2956621/modules/libzypp_config/steps/1/src
> 
> Could you explain why? In any case, I guess that would be for a different PR.

it was just a grep, wrong clue, just looking at previous steps related with zypper, I'm wondering if we should do zypper ref after enable the repo at the very beginning, but yes, for another PR :)

PR #16136, opened 2022-12-21
Change InfluxDB to v2

grisu48:
Change the handling of InfluxDB requests to use version 2 for the publiccloud metric database.

- Related ticket: https://progress.opensuse.org/issues/118936
- Verification runs: [Azure boottime](https://duck-norris.qe.suse.de/tests/11479) | [EC2 Boottime](https://duck-norris.qe.suse.de/tests/11465) | [GCE Boottime](https://duck-norris.qe.suse.de/tests/11453) | [Azure fio](https://duck-norris.qe.suse.de/tests/11470) | [EC2 fio](https://duck-norris.qe.suse.de/tests/11473) | [GCE fio](https://duck-norris.qe.suse.de/tests/11474)

Failures are due to performance deviations outside of the scope of this PR. Only the [GCE fio](https://duck-norris.qe.suse.de/tests/11474) test run compares old values.

### DIFF
--- a/lib/db_utils.pm
+++ b/lib/db_utils.pm
@@ -40,10 +40,11 @@ sub build_influx_query {
 
 =head2 influxdb_push_data
 
-    influxdb_push_data($url, $db, $data [, quiet => 1])
+    influxdb_push_data($url, $db, $org, $token, $data [, quiet => 1])
 
 Builds an influx-db query and write it to the given database specified with
-C<url> and C<db> for the Influx DB name.
+C<url>, C<org> and C<db> for the Influx DB organization and database name.
+C<token> is used for user authentication.
 C<data> is a hash containing the table name in Influx DB, the tags
 and the values to plot.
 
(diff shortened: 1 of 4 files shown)


lib/db_utils.pm
```diff
@@ -69,24 +75,35 @@ sub influxdb_push_data {
...
+    $mojo_url->query(orgID => $org, db => $db, q => $query);
+    my $res = $ua->post($mojo_url => {Authorization => "Token $token", Accept => "application/json", 'Content-type' => "application/json"})->res;
+    if ($res->code != 200) {
         die sprintf("Failed to get data from InfluxDB. \n Response code : %s \n Message: %s \n", $res->code, $res->message);
+    } else {
+        my $json = $res->json;
+        my @results = $json->{results};
+        # Check if we got an error object as reply
+        my $result = $results[0][0];
+        if (defined($result->{error})) {
```
asmorodskyi, line 102: is it possible to get HTTP=200 but error inside ?

grisu48: Yes that happens. Don't ask me why, but it's rather common.

tests/publiccloud/boottime.pm
```diff
@@ -348,6 +348,9 @@ sub store_in_db {
     my ($self, $results) = @_;
     my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
     return unless ($url);
+    my $db = get_var('PUBLIC_CLOUD_PERF_DB', 'perf');
+    my $token = get_required_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
```
asmorodskyi, line 352: maybe we want to make this variables start with `_SECRET` ? and keep them in `workers.ini` as we doing to other public cloud creds ?

lib/db_utils.pm
asmorodskyi: ok I won't ask *you* why :)

tests/publiccloud/boottime.pm
grisu48: I don't think this token is worth to keep a secret. Worst thing people can do is to push wrong metrics to our database.

asmorodskyi (approved):
LGTM

tests/publiccloud/boottime.pm
grisu48: The underscore has a different purpose: When cloning the job, you don't have those variables in the clones job on your instance, so we don't spam the database with development runs.

PR #16139, opened 2022-12-21
Revert "Switch possible parts of post_fail_hooks to serial terminal"

DimStar77:
Reverts os-autoinst/os-autoinst-distri-opensuse#16083

There are structural issues: whenever a test triggers the new fail hook, triggering a restore of LASTGOOD, subsequent tests are not clear on the console state

### DIFF
--- a/lib/Utils/Logging.pm
+++ b/lib/Utils/Logging.pm
@@ -16,7 +16,6 @@ use warnings;
 use testapi;
 use utils qw(clear_console show_oom_info remount_tmp_if_ro detect_bsc_1063638);
 use Utils::Systemd 'get_started_systemd_services';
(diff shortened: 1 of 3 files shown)


Vogtinator:
> There are structural issues: whenever a test triggers the new fail hook, triggering a restore of LASTGOOD, subsequent tests are not clear on the console state

To add some more detail: It looks like `autotest::selected_console` is not restored to the snapshot state, so that claims the serial terminal is selected while the VNC console is actually active. That means `is_serial_terminal` returns true and `assert_script_run` breaks.

Vogtinator (approved):

okurz (approved):

PR #16143, opened 2022-12-22
Switch source of SUSE CA to http

grisu48:
https requests on ca.suse.de redirect now to the confluence page, so that we need to use http for the SUSE CA certificate.

- Verification run: https://duck-norris.qe.suse.de/t11582


grisu48:
Issue resolved on the side of infra. This PR is not needed anymore.

PR #16145, opened 2022-12-22
main_ltp: Fix kernel-live-patching on s390x z/VM on o3

pevik:
This is needed for kernel-live-patching tests on z/VM (BACKEND=s390x => use bootloader_s390 instead of bootloader_zkvm), which unlike LTP tests on z/VM does not use install_ltp.pm, but only boot_ltp.pm.

NOTE: with this fix z/VM tests (both LTP and kernel-live-patching) do *not* need to use LTP_BAREMETAL=1.

Verification run:
- https://openqa.opensuse.org/tests/2971882 (Tumbleweed kernel-live-patching@s390x-zVM-vswitch-l2)
- https://openqa.suse.de/tests/overview?build=kernel-live-patching-s390x (check for a regression)
- https://openqa.opensuse.org/tests/overview?build=kernel-live-patching-s390x

### DIFF
--- a/lib/LTP/utils.pm
+++ b/lib/LTP/utils.pm
@@ -87,8 +87,10 @@ sub log_versions {
     script_run("rpm -qi $kernel_pkg > $kernel_pkg_log 2>&1");
     upload_logs($kernel_pkg_log, failok => 1);
 
-    script_run(get_ltproot . "/ver_linux > $ver_linux_log 2>&1");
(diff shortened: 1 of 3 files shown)


frankenmichl (approved):
looks fine to me

czerw (approved):

mdoucha (approved):

PR #16147, opened 2022-12-22
LTP: Run ver_linux only for LTP tests

pevik:
While this has been fixes for shutdown_ltp.pm in 36e798e4e, it needs to be fixed also for boot_ltp.pm, which is used also by kernel-live-patching tests (QA_TEST_KLP_REPO variable).

Verification run: TODO


pevik:
Moved to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16145.

PR #16148, opened 2022-12-23
Update Yam Mailing List in source code

chcao:
The squad name and mailing list are changed, update the Yam mailing list in source code.

- Related ticket: https://progress.opensuse.org/issues/122053
- Needles: n/a
- Verification run: n/a


lib/Distribution/Opensuse/Leap/15.pm
```diff
@@ -8,7 +8,7 @@
 # It follows latest SLE 15 and only if it behaves different in Leap 15 then it
 # should be overriden here.
 #
-# Maintainer: QE YaST <qa-sle-yast@suse.de>
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam@suse.de>
```
coolgw, line 11: why not just use simple key work  "QE Yam team" instead of "QE YaST and Migration (QE Yam)"

chcao: In ticket, it advises to use "QE YaST and Migration (QE Yam)", I think in case of others don't know what does "QE Yam" mean. Using "QE YaST and Migration (QE Yam)" makes it more clear.

coolgw (approved):

jknphy:
we need to squash commits before merging.

PR #16155, opened 2022-12-27
Add SemVer as a dependency to .spec file

BillAnastasiadis:
Semver is added in the spec file, and the updated spec is included.

- Related ticket: https://progress.opensuse.org/issues/120918


rfan1 (approved):
LGTM

PR #16167, opened 2022-12-30
Improve quality of Photofinish scenarios collection

mpagot:
Remove duplicated scenarios upload
Only generate scenarios after the cluster to Trento peering Upload scenarios archives with a more recognizable filename

- Ticket [CFSA-1692](https://jira.suse.com/browse/CFSA-1692)
Verification run: 
1. http://openqaworker15.qa.suse.cz/tests/118214
2. http://openqaworker15.qa.suse.cz/tests/118215
3. Latest code http://openqaworker15.qa.suse.cz/tests/121955 http://openqaworker15.qa.suse.cz/tests/121956

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -66,6 +66,7 @@ our @EXPORT = qw(
   k8s_logs
   k8s_test
   trento_support
+  trento_collect_scenarios
   trento_api_key
 );
 
@@ -708,9 +709,7 @@ and upload the logs
 =cut
 
 sub trento_support {
-    my ($scenario) = @_;
     my $machine_ip = get_trento_ip();
-
     return unless ($machine_ip);
 
     my $log_dir = 'remote_logs/';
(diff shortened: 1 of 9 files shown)


rtorrero (commented):
From what I see, nothing to add on my side :yes:
PS: the `PATATINE` just made me hungry :smile:

lib/trento.pm
```diff
@@ -726,16 +725,40 @@ sub trento_support {
     script_run(az_vm_ssh_cmd($cmd, $machine_ip));
     script_run("$scp_cmd:'*.tar.gz' $log_dir");
 
+    foreach my $this_file (split("\n", script_output('ls -1 ' . $log_dir . '*.tar.gz'))) {
+        upload_logs($this_file, failok => 1);
+    }
+    script_run("rm $log_dir *.tar.gz");
```
alvarocarvajald, line 731: I think this will fail as you cannot remove a directory with `rm`. It should be either:
```
script_run("rm $log_dir*.tar.gz");
```
... without the blank space, or:
```
script_run("rm $log_dir" . '*.tar.gz');
```
... which I don't like as it's almost the concatenation of 2 constants, or:
```
script_run('rm ' . $log_dir . '*.tar.gz');
```
... which I don't like as it's the concatenation of 2 constants and one scalar which can be placed between double quotes..

You can also probably do:
```
script_run("rm $log_dir/*.tar.gz");
```
... which IMO looks better.

Also, since return value it's not important, this `rm` can probably be a `rm -f` and this `script_run()` can be replaced with an `enter_cmd()`.

mpagot: Here an execution with the fixed `rm` http://openqaworker15.qa.suse.cz/tests/121956#step/cluster_deploy/95 (there's one extra slash that I can remove)

alvarocarvajald (approved):
LGTM

PR #16171, opened 2023-01-03
ping:Select console at the start of test

mloviska:
User `bernhard` does not have rights to execute `sysctl`.

Failure: https://openqa.suse.de/tests/10266147#step/ping/4
VR: http://kepler.suse.cz/tests/19781#step/ping/4


pevik:
I'm sorry, I didn't notice this, otherwise I would not have created: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16172.

PR #16178, opened 2023-01-04
Add Azure Compute Galleries for Arm64 image upload

grisu48:
For Arm64 image uploads, we need to use Azure Compute Galleries. This commit implements the handling for Arm64 image uploads, while preserves the workflow for all other ones.

- Verification run: [Uploading Arm64 image](https://duck-norris.qe.suse.de/tests/11639) | [Upload x86_64 image](https://duck-norris.qe.suse.de/tests/11640)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -19,6 +19,7 @@ use publiccloud::ssh_interactive 'select_host_console';
 
 has resource_group => 'openqa-upload';
 has container => 'sle-images';
+has image_gallery => 'test_image_gallery';
 has lease_id => undef;
 
 sub init {
@@ -100,6 +101,21 @@ sub create_resources {
     record_info('INFO', 'Create storage container ' . $self->container);
     assert_script_run('az storage container create --account-name ' . $storage_account
           . ' --name ' . $self->container, $timeout);
+    # Image gallery for Arm64 images
+    assert_script_run('az sig create --resource-group ' . $self->resource_group . ' --gallery-name "' . $self->image_gallery . '" --description "openQA upload Gallery"', timeout => 300);
+}
+
+sub calc_img_version {
+    # Build the image Version for upload to the Compute Gallery.
+    # The expected format is 'X.Y.Z'.
+    # We assemble the img version by the PUBLIC_CLOUD_BUILD (Format: 'X.Y') and the digits of PUBLIC_CLOUD_KIWI_BUILD, formatted as integer
+
+    my $build = get_required_var('PUBLIC_CLOUD_BUILD');
+    # Take only the digits of PUBLIC_CLOUD_KIWI_BUILD and convert to an int, to remove leading zeros
+    my $kiwi = get_required_var('PUBLIC_CLOUD_BUILD_KIWI');
+    $kiwi =~ s/\.//g;
+    $kiwi = int($kiwi);
+    return "$build.$kiwi";
 }
 
 sub upload_img {
@@ -124,16 +140,70 @@ sub upload_img {
 
     my $key = $self->get_storage_account_keys($storage_account);
 
-    assert_script_run('az storage blob upload --max-connections 4 --account-name '
-          . $storage_account . ' --account-key ' . $key . ' --container-name ' . $self->container
-          . ' --type page --file ' . $file . ' --name ' . $img_name, timeout => 60 * 60 * 2);
-    assert_script_run('az disk create --resource-group ' . $self->resource_group . ' --name ' . $disk_name
-          . ' --source https://' . $storage_account . '.blob.core.windows.net/' . $self->container . '/' . $img_name
-          . ' --hyper-v-generation=' . $sku . ' --architecture=' . $arch);
 
-    assert_script_run('az image create --resource-group ' . $self->resource_group . ' --name ' . $img_name
-          . ' --os-type Linux --hyper-v-generation=' . $sku . ' --source=' . $disk_name);
 
+    # Check if blob already exists
+    my $container = $self->container;
+    my $blobs = script_output("az storage blob list --container-name '$container' --account-name '$storage_account' | jq '.[].name'");
+    $blobs =~ s/^\s+|\s+$//g;    # trim
+    my @blobs = split(/\n/, $blobs);
+    if (grep(/$img_name/, @blobs)) {
+        record_info('blob', "Blob already exists, omitting upload\nExisting blobs:\n$blobs");
+    } else {
+        record_info("blobs", $blobs);
+        # Note: VM images need to be a page blob type
+        assert_script_run('az storage blob upload --max-connections 4 --account-name '
+              . $storage_account . ' --account-key ' . $key . ' --container-name ' . $self->container
+              . ' --type page --file ' . $file . ' --name ' . $img_name, timeout => 60 * 60 * 2);
+    }
+
+    if ($arch eq 'Arm64') {
+        # For Arm64 images we need to use the image galleries
+        my $publisher = get_var("PUBLIC_CLOUD_AZURE_PUBLISHER", "qe-c");
+        my $offer = get_required_var("PUBLIC_CLOUD_AZURE_OFFER");
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        $sku = get_required_var("PUBLIC_CLOUD_AZURE_SKU");
+        ## For the Azure Compute Gallery, multiple target regions are supported.
+        # This is necessary, because the image version upload needs to happen once for all regions, for which we want to
+        # execute test runs. For reasons of being concise we re-use the existing variable PUBLIC_CLOUD_REGION, but here
+        # it can contain a comma-separated list of all regions, in which the uploaded image should be available
+        my $target_regions = get_var("PUBLIC_CLOUD_REGION", "westeurope");
+        $target_regions =~ s/,/ /g;    # CLI expects spaces as separation, not commas
+        my $hyperv = $sku =~ 'gen1' ? 'V1' : 'V2';
+        my $subscription = $self->provider_client->subscription;
+        my $sa_url = "/subscriptions/$subscription/resourceGroups/imageGroups/providers/Microsoft.Storage/storageAccounts/$storage_account";
+        my $version = calc_img_version();
+
+        my $resource_group = $self->resource_group;
+        my $gallery = $self->image_gallery;
+
+        ## Create image definition. This image definition can then be used by addressing it with it's
+        ## /subscription/.../resourceGroups/openqa-upload/providers/Microsoft.Compute/galleries/...
+        ## link.
+        ## 1. Ensure the image definition in the Azure Compute Gallery exists
+        ## 2. Create a new image version for that definition. Use the link to the uploaded blob to create this version
+
+        # Print image definitions as a help to debug possible conflicting definitions
+        my $images = script_output("az sig image-definition list -g '$resource_group' -r '$gallery'");
+        record_info("img-def", "Existing image definitions:\n$images");
+
+        # Note: Repetitive calls are do not fail
+        assert_script_run("az sig image-definition create --resource-group '$resource_group' --gallery-name '$gallery' " .
+              "--gallery-image-definition '$definition' --os-type Linux --publisher '$publisher' --offer '$offer' --sku '$sku' " .
+              "--architecture '$arch' --hyper-v-generation '$hyperv' --os-state 'Generalized'", timeout => 300);
+        assert_script_run("az sig image-version create --resource-group '$resource_group' --gallery-name '$gallery' " .
+              "--gallery-image-definition '$definition' --gallery-image-version '$version' --os-vhd-storage-account '$sa_url' " .
+              "--os-vhd-uri https://$storage_account.blob.core.windows.net/$container/$img_name --target-regions $target_regions", timeout => 60 * 30);
+    } else {
+        # Create disk from blob
+        assert_script_run('az disk create --resource-group ' . $self->resource_group . ' --name ' . $disk_name
+              . ' --source https://' . $storage_account . '.blob.core.windows.net/' . $self->container . '/' . $img_name
+              . ' --hyper-v-generation=' . $sku . ' --architecture=' . $arch);
+        # Create image from disk
+        assert_script_run('az image create --resource-group ' . $self->resource_group . ' --name ' . $img_name
+              . ' --os-type Linux --hyper-v-generation=' . $sku . ' --source=' . $disk_name);
+    }
     return $img_name;
 }
 
--- a/variables.md
+++ b/variables.md
@@ -265,6 +265,7 @@ PUBLIC_CLOUD_AZ_API_VERSION | string | "2021-02-01" | For Azure, it is the API v
 PUBLIC_CLOUD_HDD2_SIZE | integer | "" | If set, the instance will have an additional disk with the given capacity in GB
 PUBLIC_CLOUD_HDD2_TYPE | string | "" | If PUBLIC_CLOUD_ADDITIONAL_DISK_SIZE is set, this defines the additional disk type (optional). The required value depends on the cloud service provider.
 PUBLIC_CLOUD_ARCH | string | "x86_64" | The architecture of created VM.
+PUBLIC_CLOUD_AZURE_PUBLISHER | string | "SUSE" | Specific to Azure. Allows to define the used publisher, if it should not be "SUSE"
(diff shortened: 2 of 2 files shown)


lib/publiccloud/azure.pm
```diff
@@ -100,6 +105,8 @@ sub create_resources {
     record_info('INFO', 'Create storage container ' . $self->container);
     assert_script_run('az storage container create --account-name ' . $storage_account
           . ' --name ' . $self->container, $timeout);
+    # Image gallery for Arm64 images
+    assert_script_run('az sig create --resource-group ' . $self->resource_group . ' --gallery-name "' . $self->image_gallery . '" --description "openQA upload Gallery"', timeout => 300);
```
asmorodskyi, line 109: is it necessary to do this also for "x86_64" calls ? ( let's do this conditionally to arch )

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
-    assert_script_run('az disk create --resource-group ' . $self->resource_group . ' --name ' . $disk_name
-          . ' --source https://' . $storage_account . '.blob.core.windows.net/' . $self->container . '/' . $img_name
-          . ' --hyper-v-generation=' . $sku . ' --architecture=' . $arch);
 
-    assert_script_run('az image create --resource-group ' . $self->resource_group . ' --name ' . $img_name
-          . ' --os-type Linux --hyper-v-generation=' . $sku . ' --source=' . $disk_name);
 
+    # Check if blob already exists
+    my $container = $self->container;
+    my $blobs = script_output("az storage blob list --container-name '$container' --account-name '$storage_account' | jq '.[].name'");
```
asmorodskyi, line 138: we already doing this here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/upload_image.pm#L31 

it is not expected that you getting into this function when image is already there . I am fine to **move** this check into the function but I find it not effective to do this two times

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
+    } else {
+        # Upload blob to Azure. Keep in mind that VM images need to be a page blob type
+        assert_script_run('az storage blob upload --max-connections 4 --account-name '
+              . $storage_account . ' --account-key ' . $key . ' --container-name ' . $self->container
+              . ' --type page --file ' . $file . ' --name ' . $img_name, timeout => 60 * 60 * 2);
+    }
+
+    if ($arch eq 'Arm64') {
+        # For Arm64 images we need to use the image galleries
+        my $publisher = get_var("PUBLIC_CLOUD_AZURE_PUBLISHER", "qe-c");
```
asmorodskyi, line 153: please add into variables.md

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
+        my $offer = get_required_var("PUBLIC_CLOUD_AZURE_OFFER");
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
```
asmorodskyi, line 156: please add into variables.md

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
+        $sku = get_required_var("PUBLIC_CLOUD_AZURE_SKU");
+        my $target_regions = get_var("PUBLIC_CLOUD_REGION", "westeurope");
+        $target_regions =~ s/,/ /;    # CLI expects spaces as separation, not commas
+        my $hyperv = $sku =~ 'gen1' ? 'V1' : 'V2';
+        my $subscription = get_subscription();
+        my $sa_url = "/subscriptions/$subscription/resourceGroups/imageGroups/providers/Microsoft.Storage/storageAccounts/$storage_account";
+        my $version = get_required_var('PUBLIC_CLOUD_BUILD');
```
asmorodskyi, line 163: this value is very common , what will happen when such value already exists ? are we ok with this behavior ? maybe better use date here ? or at least combination of KIWI build together with this value

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
+    if ($arch eq 'Arm64') {
+        # For Arm64 images we need to use the image galleries
+        my $publisher = get_var("PUBLIC_CLOUD_AZURE_PUBLISHER", "qe-c");
+        my $offer = get_required_var("PUBLIC_CLOUD_AZURE_OFFER");
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        $sku = get_required_var("PUBLIC_CLOUD_AZURE_SKU");
+        my $target_regions = get_var("PUBLIC_CLOUD_REGION", "westeurope");
+        $target_regions =~ s/,/ /;    # CLI expects spaces as separation, not commas
+        my $hyperv = $sku =~ 'gen1' ? 'V1' : 'V2';
```
asmorodskyi, line 160: @pdostal  I remember you were doing something in same direction recently but can not catch up what exactly ...
do you see any code duplication here ?

lib/publiccloud/azure.pm
```diff
@@ -124,16 +131,61 @@ sub upload_img {
...
+    }
+
+    if ($arch eq 'Arm64') {
+        # For Arm64 images we need to use the image galleries
+        my $publisher = get_var("PUBLIC_CLOUD_AZURE_PUBLISHER", "qe-c");
+        my $offer = get_required_var("PUBLIC_CLOUD_AZURE_OFFER");
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        $sku = get_required_var("PUBLIC_CLOUD_AZURE_SKU");
+        my $target_regions = get_var("PUBLIC_CLOUD_REGION", "westeurope");
```
asmorodskyi, line 158: why plural and comma ? I don't remember that we need/support upload to several regions ?

lib/publiccloud/azure.pm
grisu48: This is the routine for the creation of the persistent objects. This is architecture independent.

lib/publiccloud/azure.pm
grisu48: `find_img` checks for existing disk images. In the Arm64 case we are not creating disk images, so that check doesn't work there.

Also checking for a already existing blob is useful in all cases, e.g. when the upload might be interrupted.

There are no downsides in doing this check for all images, and it is necessary for Arm64 blob, where we are not creating disk images.

lib/publiccloud/azure.pm
grisu48: Azure expects the version to be a three digit value `1.2.3`, so it would be difficult to implement it in any other way.

lib/publiccloud/azure.pm
grisu48: Please see also line 157, where we need the `sku` variable. If at all, the usage of `sku` in line 121 is wrong, because it is not the sku but the hyper-v version based on the sku.

lib/publiccloud/azure.pm
grisu48: It's not yet needed, but it is supported by the tooling. And I don't see any downsides in allowing this from the side of `upload_img`, or do you?

grisu48: Also, if we ever want to test Arm64 images on other regions as well, we would need to extend the list here. This code allows this already, so all we would need to do, is to change the variables.

lib/publiccloud/azure.pm
pdostal: I agree with Felix The condition on line 121 is not 100% right and $sku = get_required_var("PUBLIC_CLOUD_AZURE_SKU"); looks like better approach to me.

lib/publiccloud/azure.pm
pdostal: What about this?
```suggestion
        my $build = get_required_var('PUBLIC_CLOUD_BUILD');
```

grisu48: In this context we are using the build number as image gallery version - See line 179:

    az sig image-version create ... --gallery-image-version '$version ...

I think `$version` is the better variable name, because it underlines the usage/intention and not the origin.

lib/publiccloud/azure.pm
asmorodskyi: > find_img checks for existing disk images. In the Arm64 case we are not creating disk images, so that check doesn't work there.

first of all for Arm64 images currently we running some code which does not work . the fact that you see green bubble does not change this statement . You running code which does not make sense and ALWAYS return you same result. This is bad practice and this is downside. 

> Checking for a already existing blob is anyway useful in all cases, e.g. when the subsequent creation of a disk is interrupted or fails.

in case of  "x86"  `find_img` will return you nothing in such case so again this line will always pass for "x86" case 

what if you will move  this check inside `if ($arch eq 'Arm64') {` and add condition to "find_img" call so we will skip it for "Arm64" ?

lib/publiccloud/azure.pm
asmorodskyi: if you will put several regions into that variable I assume you will get several failures in another places . So it may be supported by **Azure** but I have doubts about our test code . Please provide VR which proves me wrong or revert

lib/publiccloud/azure.pm
asmorodskyi: I would be nice if you bring order in this variable naming including line 121 . also it would be better from readability PoV all activity around $sku/$hyperv will happen in one place . But feel free to ignore this request ...

lib/publiccloud/azure.pm
asmorodskyi: ok , than "DD.MM.YY" ?

pdostal (approved):
🚀 🚀 🚀

lib/publiccloud/azure.pm
pdostal: But the `$self->resource_group` and `$self->image_gallery` are not often changing. Is it OK if this will be called multiple times with the same parameters?

lib/publiccloud/azure.pm
pdostal: > First of all for Arm64 images currently we running some code which does not work .

Not true. Can you @asmorodskyi point out where we already upload Arm64 images?

> 
> > Checking for a already existing blob is anyway useful in all cases, e.g. when the subsequent creation of a disk is interrupted or fails.
> 
> in case of "x86" `find_img` will return you nothing in such case so again this line will always pass for "x86" case
> 
> what if you will move this check inside `if ($arch eq 'Arm64') {` and add condition to "find_img" call so we will skip it for "Arm64" ?

And? The line does change anything. It just prints the blocks.

lib/publiccloud/azure.pm
asmorodskyi: I would say that currently this function is ALWAYS skipped . We calling it only when 'openqa_upload' resource group is not exists https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/azure.pm#L123 . 
But I  see that it is called https://duck-norris.qe.suse.de/tests/11625#step/upload_image/71 0_o . @grisu48  what am I missing here ?

lib/publiccloud/azure.pm
```diff
@@ -89,6 +90,10 @@ sub get_storage_account_keys {
     return $key;
 }
 
+sub get_subscription {
```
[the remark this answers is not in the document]
grisu48: Possibly, but since our credentials are only bound to one single subscription this is not happening.

In the case of having multiple subscriptions it's likely that we are having bigger issues, because then also `terraform` and the other `az` commands will get confused or simply use the default one.

lib/publiccloud/azure.pm
grisu48: > first of all for Arm64 images currently we running some code which does not work . the fact that you see green bubble does not change this statement . You running code which does not make sense and ALWAYS return you same result. This is bad practice and this is downside.

We have a passing VR and I used the uploaded image to create a test VM. I fail to see how this would be possible, if the upload process would not work.

Please point me to the place which doesn't make any sense because I'm completely lost here.

lib/publiccloud/azure.pm
grisu48: We can do that, but then we cannot have multiple builds on the same day, which is arguably equally bad. However we could use the digits of the kiwi build number as the third version number in here, would that make more sense to you?

lib/publiccloud/azure.pm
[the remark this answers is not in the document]
pdostal: OK, we can keep it as it is so when something will change we will at least notice early :D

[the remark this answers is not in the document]
asmorodskyi: we do have subscription already , please use [$self->provider_client->init();](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/azure_client.pm#L16) instead of introduction of new function

lib/publiccloud/azure.pm
asmorodskyi: ok, resolved

lib/publiccloud/azure.pm
asmorodskyi: currently every KIWI build reset count of PUBLIC_CLOUD_BUILD , so : 
1.  PUBLIC_CLOUD_KIWI_BUILD=0.9.0 PUBLIC_CLOUD_BUILD= 1.00
2.  PUBLIC_CLOUD_KIWI_BUILD=0.9.0 PUBLIC_CLOUD_BUILD= 1.01
3.  PUBLIC_CLOUD_KIWI_BUILD=0.9.0 PUBLIC_CLOUD_BUILD= 1.02
4.  PUBLIC_CLOUD_KIWI_BUILD=0.9.1 PUBLIC_CLOUD_BUILD= 1.00
5.  PUBLIC_CLOUD_KIWI_BUILD=0.9.1 PUBLIC_CLOUD_BUILD= 1.01

so while my idea will make several images during one day overwrite each other , current approach also have some scenarios when it may happen .

lib/publiccloud/azure.pm
grisu48: Here is a passing test run with `PUBLIC_CLOUD_REGION=westeurope,northeurope,westus,eastus2`: https://duck-norris.qe.suse.de/tests/11633#step/upload_image/70

And here is a screenshot from the Azure portal, showing the image version in the defined target regions

image

As far as I can see this works nicely.

lib/publiccloud/azure.pm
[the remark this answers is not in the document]
grisu48: That method relies on having the `PUBLIC_CLOUD_AZURE_SUBSCRIPTION_ID` variable, which I much rather would avoid because it's unnecessary.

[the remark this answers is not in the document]
asmorodskyi: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/azure_client.pm#L34  if this variable undefined you won't be able to continue ANY test anyway , so you MAY rely on the fact that this variable is ALWAYS available

lib/publiccloud/azure.pm
asmorodskyi: https://duck-norris.qe.suse.de/tests/11633#step/upload_image/65 but upload didn't happened with this job

lib/publiccloud/azure.pm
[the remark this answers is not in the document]
grisu48: No it isn't: https://openqa.suse.de/tests/10221030#settings. And [here](https://openqa.suse.de/tests/10221421#settings) and [here](https://openqa.suse.de/tests/10221033#settings) and [here](https://openqa.suse.de/tests/10221035#settings)

lib/publiccloud/azure.pm
asmorodskyi: resolved

lib/publiccloud/azure.pm
asmorodskyi: resolved

lib/publiccloud/azure.pm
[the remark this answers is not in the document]
grisu48: Switched to use `$self->provider_client->subscription`, which works, because the `subscription` variable is later on set via the credentials file.

lib/publiccloud/azure.pm
grisu48: Done by introducing the `calc_img_version` subroutine.

PR #16183, opened 2023-01-05
xfstests: set higher download log timeout

SeroSun:
xfstests in s390x fails by downloading log timeout in curl. Details in related ticket. This PR simply adds timeout from default 120s to 300s.

- Related ticket: https://progress.opensuse.org/issues/122539
- Verification run: https://openqa.suse.de/t10275671

### DIFF
--- a/tests/xfstests/generate_report.pm
+++ b/tests/xfstests/generate_report.pm
@@ -107,7 +107,7 @@ sub run {
     sleep 5;
 
     # Reload uploaded status log back to file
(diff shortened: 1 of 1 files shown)


lansuse:
LGTM

b10n1k:
just noticed that the VR still fails... is that right?

SeroSun:
@b10n1k it's failed by some known traffic blocking issue, which is under fixing by IT security team, but the timeout needs to be bigger than 130s is a conclusion in the ticket comments https://progress.opensuse.org/issues/122539#note-5

PR #16189, opened 2023-01-06
Add softfail for 'gnome_music' test due to known issue

rfan1:
https://progress.opensuse.org/issues/122725
https://bugzilla.suse.com/show_bug.cgi?id=1206793

- Related ticket: https://progress.opensuse.org/issues/122725
- Verification run: 
https://openqa.opensuse.org/tests/3017576 -Leap
https://openqa.opensuse.org/tests/3017732 -TW

### DIFF
--- a/tests/x11/gnome_music.pm
+++ b/tests/x11/gnome_music.pm
@@ -10,13 +10,28 @@
 # - Close gnome-music
 # Maintainer: Max Lin <mlin@suse.com>
 
-use base "x11test";
+use base "opensusebasetest";
+use version_utils qw(is_sle is_leap);
 use strict;
 use warnings;
 use testapi;
 use utils;
 
+sub check_bsc1206793 {
+    # Due to bsc#1206793, the test may fail with package version at '41.1-150400.3.3.1'
+    if (is_sle || is_leap) {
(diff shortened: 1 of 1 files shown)


tests/x11/gnome_music.pm
```diff
@@ -10,13 +10,23 @@
...
+use base "opensusebasetest";
 use strict;
 use warnings;
 use testapi;
 use utils;
 
+sub check_bsc1206793 {
+    # Due to bsc#1206793, the test may fail with package version at '41.1-150400.3.3.1'
+    select_console 'root-console';
+    zypper_call('in gnome-music');
```
GraceWang571, line 22: IMHO, this is duplicate with the `if script_output('rpm -q gnome-music') =~ '41.1-150400.3.3.1'`.
The aim of `sub check_bsc1206793` is to get the gnome-music version and then decide if we should record a soft failure, am I right?

rfan1: Yes, we have seen the issue with gnome-music pkg at this version, so add a workaround here to "softfail" this test. and I will revert this commit once the bug fix is checked in.

GraceWang571:
LGTM

PR #16191, opened 2023-01-09
Skip zpool scrub in case the pool is currently resilvering

rfan1:
https://progress.opensuse.org/issues/122770

- Related ticket: https://progress.opensuse.org/issues/122770
- Verification run: [VRs](https://openqa.opensuse.org/tests/3026529#next_previous)

### DIFF
--- a/tests/console/zfs.pm
+++ b/tests/console/zfs.pm
@@ -102,7 +102,9 @@ sub cleanup {
 
 sub scrub {
     my $pool = shift;
-    assert_script_run("zpool scrub $pool", timeout => 180);
+    # We may skip scrub operation in case the zpool is currently resilvering.
+    my $in_progress = script_run("zpool status $pool | grep -e 'scan:.*in progress'") == 0;
+    assert_script_run("zpool scrub $pool", timeout => 180) if (!$in_progress);
     # Wait for scrub to finish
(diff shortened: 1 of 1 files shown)


rfan1:
@grisu48 May I ask for your kindly help to review this PR?

tests/console/zfs.pm
```diff
@@ -102,7 +102,8 @@ sub cleanup {
 
 sub scrub {
     my $pool = shift;
-    assert_script_run("zpool scrub $pool", timeout => 180);
+    # We may need retry several times in case the zpool is currently resilvering.
+    script_retry("zpool scrub $pool", delay => 60, retry => 3, timeout => 180);
```
grisu48, line 106: This is suboptimal, because like this we miss all other error cases. Here it might be better to check if the pool is already being scrubbed and in such case don't start a new scrub

```suggestion
    my $in_progress = script_run("zpool status $pool | grep -e 'scan:.*in progress'") == 0;
    assert_script_run("zpool scrub $pool", timeout => 180) if (!$in_progress);
```

grisu48 (commented):
One suggestion, thank you for looking into this!

tests/console/zfs.pm
rfan1: Thanks much!

grisu48 (approved):
LGTM!

PR #16192, opened 2023-01-09
Revisit wrong settings for HA migrations

tinawang123:
Revisit wrong settings for HA migrations

- Related ticket: https://progress.opensuse.org/issues/120444
- Verification run: https://openqa.suse.de/tests/10291139
  http://openqa.suse.de/t10307585
https://openqa.suse.de/tests/10307721
aarch 15sp3
 http://openqa.suse.de/t10307586
 https://openqa.suse.de/tests/10307632
s390x 15sp2

### DIFF
--- a/data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
+++ b/data/autoyast_sle15/create_hdd/create_hdd_ha_aarch64.xml.ep
@@ -215,7 +215,7 @@
     <enabled config:type="boolean">false</enabled>
   </proxy>
   <services-manager config:type="map">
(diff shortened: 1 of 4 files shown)


b10n1k (commented):
i wonder if those xmls are used on other arches! if so it would be reasonable to run a few more VR. What do you think?

rakoenig:
LGTM

tinawang123:
> i wonder if those xmls are used on other arches! if so it would be reasonable to run a few more VR. What do you think?

Updated, Thanks

b10n1k (approved):
thanks for VRs

PR #16193, opened 2023-01-09
Revert "Add softfail for 'gnome_music' test due to known issue"

GraceWang571:
Reverts os-autoinst/os-autoinst-distri-opensuse#16189


rfan1:
Thanks Grace, the original logic has some problem, it will return '1' always for leap and sle.

PR #16200, opened 2023-01-09
Always do a rollback after zfs tests

grisu48:
The zfs test is rather intrusive on the system. To prevent disturbances with other test runs, always do a rollback after the test run.

- Related ticket: https://progress.opensuse.org/issues/122773
- Verification run: https://duck-norris.qe.suse.de/tests/11696

### DIFF
--- a/tests/console/zfs.pm
+++ b/tests/console/zfs.pm
@@ -294,4 +294,9 @@ sub post_run_hook {
     cleanup();
 }
 
+sub test_flags {
+    # Test is rather intrusive, so to be sure to not disturb the rest of the test run, do a rollback
+    return {always_rollback => 1};
+}
+
 1;


rfan1 (approved):
LGTM

rfan1:
Hello @grisu48,
Can you please have a try to 'fail' the zfs test and see we can still hit the problem mentioned in https://progress.opensuse.org/issues/122773?

grisu48:
> Hello @grisu48, Can you please have a try to 'fail' the zfs test and see we can still hit the problem mentioned in https://progress.opensuse.org/issues/122773?
> 
> As the original issue was caused by initial failure at zfs test, then rollback operation failed.

I saw that and am still puzzled why the rollback failed. However it [works nicely in the verification run](https://duck-norris.qe.suse.de/tests/11696#step/cifs/1)

I will schedule a bunch of test runs, see if I can reproduce the issue and report back here.

grisu48:
@rfan1 I run 20 test runs ([11697](http://duck-norris.qe.suse.de/tests/11697) to [11716](http://duck-norris.qe.suse.de/tests/11716)) and none of them had issues in doing the rollback.

I suggest to proceed with the PR and keep an eye on the test runs to see, if the issue is resolved or keeps popping up. I would not know at this point what we could improve further.

rfan1:
> @rfan1 I run 20 test runs ([11697](http://duck-norris.qe.suse.de/tests/11697) to [11716](http://duck-norris.qe.suse.de/tests/11716)) and none of them had issues in doing the rollback.
> 
> I suggest to proceed with the PR and keep an eye on the test runs to see, if the issue is resolved or keeps popping up. I would not know at this point what we could improve further.

I agree with you, please go ahead to merge this PR, then we can file a new ticket if the issue happens again.

PR #16203, opened 2023-01-10
Remove 'systemd-resolvd' test coverage on ALP

rfan1:
https://progress.opensuse.org/issues/122833

- Verification run: https://openqa.opensuse.org/tests/3028515

### DIFF
--- a/tests/microos/networkmanager.pm
+++ b/tests/microos/networkmanager.pm
@@ -10,6 +10,7 @@ use Mojo::Base "consoletest";
 use testapi;
 use utils;
 use transactional;
+use version_utils 'is_alp';
 
 # the below network confiration is used if we use qemu user net
 my $dhcp_server = '10.0.2.2';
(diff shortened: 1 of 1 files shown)


tests/microos/networkmanager.pm
```diff
@@ -88,23 +88,6 @@ true'
     ($RcManager, $mode) = dns_mgr();
     die 'wrong DNS-Manager is currently used for dnsmasq' if ($RcManager !~ /symlink/ || $mode !~ /dnsmasq/);
     ping_check;
-    # systemd-resolved
```
grisu48, line 91: If we are skipping this on ALP, I would also encapsulate it in an `is_ALP` if-block to avoid reduced coverage on SLE Micro.

Please guard the whole block within an `if (!is_ALP)` block.

grisu48 (changes requested):
The tests we're modifying is in `tests/microos`, so I think we should not do overall changes that apply to ALP only.

tests/microos/networkmanager.pm
rfan1: Good comments, let me do

PR #16220, opened 2023-01-12
WIP Add wait_still_screen before checking tomcat status

Zaoliang:
sometimes it takes much longer to complete start progress. 
see poo#120834


Zaoliang:
We cannot fix or work around this performance issue at moment, we need a better solution like re-write the test module.

PR #16222, opened 2023-01-12
Fix regexp to detect key issue in wait_for_ssh

mpagot:
The API need to detect messages like

```
someuser@1.2.3.4: Permission denied (publickey).
```

Add a parenthesis escape in the regexp about it.

- Verification run:  https://openqa.suse.de/tests/10312736


asmorodskyi (approved):
LGTM

pdostal (approved):

m-dati (approved):
LGTM
Also, much simpler, another way could be: `=~ m/Permission denied .publickey./`

PR #16228, opened 2023-01-13
Move zfs test run behind cifs and samba

grisu48:
Due to snapshot restore issues tracked in poo#109929, the zfs tests are moved behind the cifs tests, so that snapshot restoration is not causing any issues for now.

- Related ticket: https://progress.opensuse.org/issues/122773

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -1861,9 +1861,10 @@ sub load_extra_tests_filesystem {
     }
     loadtest 'console/snapper_used_space' if (is_sle('15-SP1+') || (is_opensuse && !is_leap('<15.1')));
     loadtest "console/udisks2" unless (is_sle('<=15-SP2') || get_var('VIRSH_VMM_FAMILY') =~ /xen/);
-    loadtest "console/zfs" if (is_leap(">=15.1") && is_x86_64 && !is_jeos);
     loadtest "network/cifs" if (is_sle('>=15-sp3') || is_opensuse);
     loadtest "network/samba/server" if (is_sle('>=15-sp3') || is_opensuse);
+    # Note: Until the snapshot restoration has been fixed (poo#109929), zfs should be the last test run
+    loadtest "console/zfs" if (is_leap(">=15.1") && is_x86_64 && !is_jeos);
 }
 
 sub get_wicked_tests {


rfan1:
@grisu48 
Do you mind deleting the line at zfs.pm:

sub test_flags {
    # Test is rather intrusive, so to be sure to not disturb the rest of the test run, do a rollback
    return {always_rollback => 1};
}

grisu48:
> @grisu48 Do you mind deleting the line at zfs.pm?
> 
> sub test_flags { # Test is rather intrusive, so to be sure to not disturb the rest of the test run, do a rollback return {always_rollback => 1}; }

I would not - Ultimately zfs installs kernel modules and custom repositories so it is rather intrusive and we should revert to a known good state to avoid disruptions of other tests.

Just because right now the snapshot feature is broken, does not mean that we don't want to have it.

And: In case of a failure, openQA also reverts to a snapshot, so we anyways need to deal with it.

What do you think?

lib/main_common.pm
```diff
@@ -1861,9 +1861,10 @@ sub load_extra_tests_filesystem {
     }
     loadtest 'console/snapper_used_space' if (is_sle('15-SP1+') || (is_opensuse && !is_leap('<15.1')));
     loadtest "console/udisks2" unless (is_sle('<=15-SP2') || get_var('VIRSH_VMM_FAMILY') =~ /xen/);
-    loadtest "console/zfs" if (is_leap(">=15.1") && is_x86_64 && !is_jeos);
     loadtest "network/cifs" if (is_sle('>=15-sp3') || is_opensuse);
     loadtest "network/samba/server" if (is_sle('>=15-sp3') || is_opensuse);
+    # Note: Until the snapshot restoration has been fixed (poo#109929), zfs should be the last test run
```
rfan1, line 1866: I think it is a good solution for now

grisu48: Thanks! :slightly_smiling_face:

rfan1: > I think it is a good solution for now



> > @grisu48 Do you mind deleting the line at zfs.pm?
> > sub test_flags { # Test is rather intrusive, so to be sure to not disturb the rest of the test run, do a rollback return {always_rollback => 1}; }
> 
> I would not - Ultimately zfs installs kernel modules and custom repositories so it is rather intrusive and we should revert to a known good state to avoid disruptions of other tests.
> 
> Just because right now the snapshot feature is broken, does not mean that we don't want to have it.
> 
> And: In case of a failure, openQA also reverts to a snapshot, so we anyways need to deal with it.
> 
> What do you think?

Agree with you, then LGTM

rfan1 (approved):

PR #16233, opened 2023-01-13
Introduce cloud-init verification test

mloviska:
Using `CI_DATA_FILE` the custom Cloud config data file to configure the SUT. Boolean variable `CI_VERIFICATION` loads the test verification module.
I have extended and also a bit modified the provided Cloud config file form the Jira ticket.

- ticket: [[JeOS OpenStack] Cloud-init specific test](https://progress.opensuse.org/issues/107458)
- motivation: https://jira.suse.com/browse/TEAM-6907
- Verification run: http://kepler.suse.cz/tests/19925#step/verify_cloudinit/1

### DIFF
--- a/data/jeos/g_cloudinit.cfg
+++ b/data/jeos/g_cloudinit.cfg
@@ -1,0 +1,49 @@
+#cloud-config
+# vim: syntax=yaml
+# examples are documented here https://documentation.suse.com/sles/15-SP4/html/SLES-all/article-minimal-vm.html#sec-cloud-init-config-examples
+hostname: cucaracha
+timezone: Europe/Prague
+users:
+  - default
+# New user with password and sudo
+  - name: bernhard
+    shell: /bin/bash
+    groups: users
+    # lock_passwd: Disable password login. Defaults to true
+    lock_passwd: false
+    passwd: $6$SalTsaLt$QZZq7a5g5.CHg2zW6VGhKV2NZ6WnYDmrX1P8aVvE5YZtVBk9I4UhWkmT9yegT35QYXjOWCxGWAKnFcuC84w46/
+    sudo: ALL=(ALL) NOPASSWD:ALL
+# New user with SSH keys and sudo #
+  - name: tester_ssh
+    shell: /bin/bash
+    groups: users
+    ssh_import_id: None
+    lock_passwd: true
+    sudo: ALL=(ALL) NOPASSWD:ALL
+    ssh_authorized_keys:
+      - %SSH_PUBK_RSA%
+      - %SSH_PUBK_ECDSA%
+
+# Register system
+bootcmd:
+  - SUSEConnect -r %SCC_REGCODE% %SCC_URL%
+# Register with RMT
+# runcmd:
+#   - curl http://RMT_SERVER/tools/rmt-client-setup \ --output rmt-client-setup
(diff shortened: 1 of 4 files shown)


tests/jeos/verify_cloudinit.pm
```diff
@@ -0,0 +1,102 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
```
jlausuch, line 3:
```suggestion
# Copyright 2023 SUSE LLC
```

tests/jeos/verify_cloudinit.pm
```diff
@@ -0,0 +1,102 @@
...
+            keyid => '~/.ssh/test_ci_ecdsa'
+        },
+        $testapi::username => {
+            passwd => 'P',
+            keyid => '~/.ssh/id_rsa'
+        }
+    };
+
+    select_console('root-console');
+    assert_script_run('cloud-init status --wait --long');
```
jlausuch, line 31: shall we add a check for the output of this command?

data/jeos/g_cloudinit.cfg
```diff
@@ -0,0 +1,49 @@
+#cloud-config
+# vim: syntax=yaml
+# examples are documented here https://documentation.suse.com/sles/15-SP4/html/SLES-all/article-minimal-vm.html#sec-cloud-init-config-examples
+hostname: cucaracha
```
jlausuch, line 4: cucaracha? :D

lib/main_common.pm
```diff
@@ -597,7 +597,14 @@ sub load_jeos_openstack_tests {
     } else {
         loadtest 'publiccloud/ssh_interactive_start', run_args => $args;
     }
-    #    loadtest "jeos/image_info";
+
+    if (get_var('CI_VERIFICATION')) {
```
jlausuch, line 601: `CI_` sounds like Continuous Integration to me. Shall we use `CLOUD_INIT_VERIFICATION`?

lib/main_common.pm
```diff
@@ -597,7 +597,14 @@ sub load_jeos_openstack_tests {
...
+        loadtest 'jeos/verify_cloudinit', run_args => $args;
+        loadtest("publiccloud/ssh_interactive_end", run_args => $args);
```
jlausuch, line 603: I'm confused why you add this here. If we don't wanna run things on OpenStack, we can even remove this and ssh_start, can't we?

jlausuch (commented):
This looks already very good!

lib/main_common.pm
mloviska: The test is using multiple consoles, so I have kinda found it useful to run interactive ssh. As it should run in a single test suite, we need to close the tunnel here. Or do you prefer to run some additional tests?

tests/jeos/verify_cloudinit.pm
mloviska: Returns non-zero code when the status reports a failure

lib/main_common.pm
jlausuch: But ssh_interactive_start/end is used for PublicCloud/OpenStack tests. We are not planning to run this test on those platforms, right?

jlausuch (approved):

gmoro (approved):
Looks good to me

PR #16236, opened 2023-01-16
Remove any previous java installed before installing jdk11

ilmanzo:
Test installs and check crypto features on jdk11, but there can be some newer JDK on the image; so we uninstall them before testing. 

- Related ticket: https://progress.opensuse.org/issues/116263
- Verification run: https://openqa.suse.de/t10322798


paolostivanin:
LGTM

punkioudi (approved):
lgtm!

PR #16241, opened 2023-01-17
Bump wait_serial timeout in yast2_clone_system

sofiasyria:
There have been some failures on aarch64, because of cloning taking too long. Increasing tiemout of the command is resulting in successful runs:
https://openqa.suse.de/tests/10324395#next_previous

Progress ticket: https://progress.opensuse.org/issues/122851

### DIFF
--- a/tests/console/yast2_clone_system.pm
+++ b/tests/console/yast2_clone_system.pm
@@ -13,6 +13,7 @@ use strict;
 use warnings;
 use testapi;
 use serial_terminal 'select_serial_terminal';
+use Utils::Architectures 'is_aarch64';
 use version_utils qw(is_sle is_opensuse is_staging);
 use utils 'zypper_call';
 use repo_tools 'get_repo_var_name';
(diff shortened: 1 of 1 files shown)


tests/console/yast2_clone_system.pm
```diff
@@ -25,7 +25,7 @@ sub run {
     zypper_call "in autoyast2";
 
     my $module_name = y2_module_consoletest::yast2_console_exec(yast2_module => 'clone_system');
-    wait_serial("$module_name-0", 360) || die "'yast2 clone_system' didn't finish";
+    wait_serial("$module_name-0", 1500) || die "'yast2 clone_system' didn't finish";
```
jknphy, line 28: let's do an ugly `if aarch` because if this happens in other arch like x86 we would missed some bug for sure.
I can see that normally takes 900 aprox (15 minutes), 1200 could work, right?

sofiasyria: ok, rerunning the VRs

ge0r (approved):

jknphy:
CI is stuck, could you please try to push again?
re-running checks doesn't seem to work.

PR #16242, opened 2023-01-17
Create & apply defaults for uefi backend

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/119890
- Needles: - 
- [Verification runs](https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=rakoenig%2Fos-autoinst-distri-opensuse%23default_x86_64_uefi)
- Related MR on GitlLab: https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/460


hjluo:
LGTM

JRivrain:
VRs are gone!

rakoenig:
That means we should merge much faster :,-)
Nevertheless I restarted the VRs but ran into strange problems (openqa does not want to clone my tree) for 2 of the tests.

rakoenig:
Restarted all VRs again because of a typo in the branch-spec. Now everything should work as expected.

JRivrain:
For some reason nvme was missing, I launched it, it ended up with a slightly different build name so not with the same url https://openqa.suse.de/tests/10527806

JRivrain:
Stupid question that came to me after reviewing each file one by one, and that I wish I had thought of earlier: why do we need this at all? there is no difference in schedule between uefi and non-uefi AFAIK, just a couple of variables like UEFI=1. Take any uefi test, they are absolutely identical to non-uefi, the only difference being the way  bootloader_start handles the bootloader, so no need to have specific schedule. If there is some good reason, then consider LGTM, but otherwise I think we just overcomplicated our lives.

rakoenig:
I compared the YAMLs now and yes, after merging the PR for x86_64 default vars it is almost identical. The only difference is the `yast/nvme.yaml` that got not modified in the x86_64 PR, but is in this one. So I'd suggest we remove everything here except the  newer `yast/nvme.yaml` and then adjust the MR to use the x86_64 default scheudles. WDYT?

JRivrain:
> I compared the YAMLs now and yes, after merging the PR for x86_64 default vars it is almost identical. The only difference is the `yast/nvme.yaml` that got not modified in the x86_64 PR, but is in this one. So I'd suggest we remove everything here except the newer `yast/nvme.yaml` and then adjust the MR to use the x86_64 default scheudles. WDYT?

Looks like a good idea yes :)

JRivrain:
Vrs updated and passed.

PR #16243, opened 2023-01-17
Need install yast2-iscsi-client for iscsi-client x11 test

lemon-suse:
We need install yast2-iscsi-client for iscsi-client x11 test.

- Related ticket: N/A
- Needles: N/A
- Verification run:  http://openqa.suse.de/tests/10322965#step/iscsi_client/69

### DIFF
--- a/tests/iscsi/iscsi_client.pm
+++ b/tests/iscsi/iscsi_client.pm
@@ -105,7 +105,7 @@ sub initiator_connected_targets_tab {
 
 sub run {
     prepare_xterm_and_setup_static_network(ip => $test_data->{initiator_conf}->{ip}, message => 'Configure MM network - client');
-    zypper_call("in open-iscsi");
+    zypper_call("in yast2-iscsi-client");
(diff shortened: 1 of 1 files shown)


tests/iscsi/iscsi_client.pm
```diff
@@ -106,6 +106,7 @@ sub initiator_connected_targets_tab {
 sub run {
     prepare_xterm_and_setup_static_network(ip => $test_data->{initiator_conf}->{ip}, message => 'Configure MM network - client');
     zypper_call("in open-iscsi");
+    zypper_call("in yast2-iscsi-client");
```
dzedro, line 109: You can just install `yast2-iscsi-client` it will pull in `open-iscsi`.

lemon-suse: Ok, thanks, updated. I will run a new VR for it.

jknphy: this doesn't happen anymore for new products since YaST has containerized YaST modules, the deps needs to be installed explicitly.

PR #16219, opened 2023-01-12
Fix xterm case in LEAP 15.4

DeepthiYV:
Due to the issue [bsc#1205518](https://bugzilla.opensuse.org/show_bug.cgi?id=1205518) in LEAP,  the 'xterm' test is failing. As suggested called VM reboot before the 'xterm' test.

- Related ticket: https://progress.opensuse.org/issues/122992
- Needles: NO
- Verification run:  [Leap_15.4](http://deepthi-openqa.qe.suse.de/tests/4142#step/xterm/1)

PR #16244, opened 2023-01-17
Revert "Fix xterm case in LEAP 15.4"

Vogtinator:
Reverts os-autoinst/os-autoinst-distri-opensuse#16219

This breaks most Leap 15.4 tests: https://openqa.opensuse.org/tests/overview?distri=opensuse&version=15.4&build=31.94&groupid=89

Is this actually Leap specific? IMO this should at least do `if (is_leap("=15.4") && is_gnome)`, ideally also check for this specific error condition.

Why is this workaround in the `xterm` module?

@DeepthiYV

### DIFF
--- a/tests/x11/xterm.pm
+++ b/tests/x11/xterm.pm
@@ -14,17 +14,9 @@ use base "x11test";
 use strict;
 use warnings;
 use testapi;
-use power_action_utils qw(power_action);
-use version_utils qw(is_leap);
 
 sub run {
     my ($self) = @_;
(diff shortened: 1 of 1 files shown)


DeepthiYV:
The issue https://bugzilla.opensuse.org/show_bug.cgi?id=1205518 was causing the failure of xterm test in Leap 15.4. https://openqa.opensuse.org/tests/3042798#step/xterm/1 , So added the VM reboot as a workaround.

PR #16245, opened 2023-01-17
wget return code got lost due to pipe 

mloviska:
Current image download does not warn the use that [image has not been found](http://kepler.suse.cz/tests/20005#step/upload_image/63). Reading the HTTP header we can examine whether `wget` was successful or not. For testing purposes I have tried to download `I_do_Not_Exist.qcow2` image from OSD hdd store.

- ticket: https://progress.opensuse.org/issues/121258
- VR: http://kepler.suse.cz/tests/20004#step/upload_image/65

### DIFF
--- a/tests/publiccloud/upload_image.pm
+++ b/tests/publiccloud/upload_image.pm
@@ -33,9 +33,9 @@ sub run {
     }
 
     # Download the given image via wget. Note that by default wget retries 20 times before giving up
-    my $cmd = "wget -q --no-check-certificate --retry-connrefused --retry-on-host-error $img_url -O $img_name";
+    my $cmd = "wget -q --server-response --no-check-certificate --retry-connrefused --retry-on-host-error $img_url -O $img_name";
     # A generious timeout is required because downloading up to 30 GB (Azure images) can take more than an hour.
-    my $rc = script_run("$cmd 2>&1 | tee download.txt", timeout => 120 * 60);
+    my $rc = script_run("(set -o pipefail && $cmd 2>&1 | tee download.txt)", timeout => 120 * 60);
     if ($rc != 0) {
         # Check for 404 errors and make them better visible
         upload_logs("download.txt");


tests/publiccloud/upload_image.pm
```diff
@@ -32,6 +32,12 @@ sub run {
         return;
     }
 
+    # Check if remote file exists
+    my $res = script_output("wget -q --spider -S $img_url");
+    if ($res =~ m@HTTP/1.1 404 Not Found@) {
+        die "Image was not found";
```
asmorodskyi, line 38:
```suggestion
        die "Image was not found in IBS";
```

tests/publiccloud/upload_image.pm
```diff
@@ -32,6 +32,12 @@ sub run {
         return;
     }
 
+    # Check if remote file exists
+    my $res = script_output("wget -q --spider -S $img_url");
+    if ($res =~ m@HTTP/1.1 404 Not Found@) {
```
m-dati, line 37: Also working by excluding version and slash: m/HTTP.*404 Not Found/,
but ok as it is now, LGTM.

tests/publiccloud/upload_image.pm
mloviska: I find IBS misleading, as for OpenStack we don't pull it from there by the test. Maybe I can print also the content of `$img_url` in the message

tests/publiccloud/upload_image.pm
mloviska: Yeah, I thought about that. I will update the regex. Thanks!

tests/publiccloud/upload_image.pm
asmorodskyi, line 38:
```suggestion
        die "Image was not found in $img_url";
```

mloviska:
VR: http://kepler.suse.cz/tests/20009#step/upload_image/50

ricardobranco777 (approved):
LGTM

m-dati (approved):
LGTM

PR #16248, opened 2023-01-18
WIP split snapper commands and add timeout for assert_script_run

Zaoliang:
we have sporadic issue with timeout for snapper list, it takes just time. 
split 2 commands and add timeout for 'snapper list' see https://progress.opensuse.org/issues/122557


Zaoliang:
VR: https://openqa.opensuse.org/tests/3050863 (JeOS)

Zaoliang:
VR:
http://openqa.suse.de/tests/10337183 (sles 15 sp5)

tests/console/snapper_cleanup.pm
```diff
@@ -136,7 +136,10 @@ sub run {
     # tidy up and restore default settings
     assert_script_run("snapper set-config NUMBER_LIMIT=0; snapper cleanup number; rm -fv data", 300);
     assert_script_run("snapper set-config NUMBER_LIMIT=$number_limit_pre NUMBER_MIN_AGE=$number_min_age_pre");
-    assert_script_run("snapper get-config; snapper ls");    # final report
+    assert_script_run("snapper get-config");    # final report
+        # command 'snapper ls' runs timeout or it stucks because of display used space, so we have to disable it on tumbleweed, see poo#122557
+    my $command_args = (is_tumbleweed && check_var('MACHINE', '64bit')) ? '--disable-used-space' : undef;
```
rfan1, line 141: Is it too strict with  check_var('MACHINE', '64bit')? 
We may use 64bit-2G as a machine type, then we may hit a new issue :)

tests/console/snapper_cleanup.pm
```diff
@@ -136,7 +136,10 @@ sub run {
     # tidy up and restore default settings
     assert_script_run("snapper set-config NUMBER_LIMIT=0; snapper cleanup number; rm -fv data", 300);
     assert_script_run("snapper set-config NUMBER_LIMIT=$number_limit_pre NUMBER_MIN_AGE=$number_min_age_pre");
-    assert_script_run("snapper get-config; snapper ls");    # final report
+    assert_script_run("snapper get-config");    # final report
+        # command 'snapper ls' runs timeout or it stucks because of display used space, so we have to disable it on tumbleweed, see poo#122557
```
rfan1, line 140:
```suggestion
    # command 'snapper ls' runs timeout or it stucks because of display used space, so we have to disable it on tumbleweed, see poo#122557
```

tests/console/snapper_cleanup.pm
```diff
@@ -136,7 +136,10 @@ sub run {
...
+    my $command_args = (is_tumbleweed && check_var('MACHINE', '64bit')) ? '--disable-used-space' : undef;
+    assert_script_run("snapper ls $command_args", 30);
```
grisu48, line 142: Maybe we also want to give `snapper ls` some more time?

```suggestion
    assert_script_run("snapper ls $command_args", 300);
```

grisu48 (commented):
Small suggestion, otherwise LGTM!

tests/console/snapper_cleanup.pm
Zaoliang: well, I want just limit this change to tw 64bit, I checked that tw has at moment only this test run with 64 bit qemuram 1.5. I don't care about the future :) if it has 2G then it migt be okay anyway... the strange thing it that this test module passed everywhere except on tw with very limited allocation of cpu or mem.

tests/console/snapper_cleanup.pm
Zaoliang: this is tidy correction, really weird ;)

tests/console/snapper_cleanup.pm
Zaoliang: tried already, 300 is too much, this is a large test run. without displaying used space it takes just under 1 sec, with it it takes forever on tw.

tests/console/snapper_cleanup.pm
rfan1: Do you mind checking the qemuram as well?

PR #16249, opened 2023-01-18
Retry zypper install as opensuse download is unstable laterly

Julie-CAO:
Improve test stability by following ways:

- try 5 times to zypper install kvm/xen patterns to fix this failing case: https://openqa.opensuse.org/tests/3049532#step/switch_to_ssh_and_install_hypervisor/18. 3 times tries still failed, fg. https://openqa.opensuse.org/tests/3077870.
- replace script_output() with script_run() as it always fails. fg. https://openqa.opensuse.org/tests/3069180#step/unified_guest_installation/38 & https://openqa.opensuse.org/tests/3076648#step/unified_guest_installation/100
- wait 5s after sshd restarted, because https://openqa.opensuse.org/tests/3073980
- Need not update kvm/xen packages as it is TumbleWeed.
- Make test continue with failing `dnsdomainname` command as it might
  not work on SUTs with multiple interfaces, fg. https://openqa.opensuse.org/tests/3082110#step/unified_guest_installation/336
- Try one more time login as TW fails with incorrect password sporatically in O3, fg. https://openqa.opensuse.org/tests/3089170#step/switch_to_ssh_and_install_hypervisor/5. have to verify this commit in O3 builds. Improve this change on need later.
- Mark dry run failed and continue tests as it always timeout in O3 when download kernel files from download.opensuse.org, fg. https://openqa.opensuse.org/tests/3102454#step/unified_guest_installation/296. In addition, reconnnect ssh console is added because https://openqa.opensuse.org/tests/3106501#step/unified_guest_installation/306.

Related ticket: https://progress.opensuse.org/issues/103742

Verification run: 
- TW kvm: https://openqa.opensuse.org/tests/3081080
- TW xen:  https://openqa.opensuse.org/tests/3078108
- SLE in my openqa: http://10.67.129.102/tests/473

### DIFF
--- a/lib/guest_installation_and_configuration_base.pm
+++ b/lib/guest_installation_and_configuration_base.pm
@@ -1,6 +1,6 @@
 # VIRTUAL MACHINE INSTALLATION AND CONFIGURATION BASE MODULE
 #
-# Copyright 2021 SUSE LLC
+# Copyright SUSE LLC
 # SPDX-License-Identifier: FSFAP
 #
 # Summary: This module provides framework and APIs to install virtual
@@ -253,7 +253,8 @@ sub initialize_guest_params {
     $self->{$_} //= '' foreach (keys %guest_params);
     $self->{host_ipaddr} = get_required_var('SUT_IP');
     $self->{host_name} = script_output("hostname");
-    $self->{host_domain_name} = script_output("dnsdomainname");
+    # For SUTs with multiple interfaces, `dnsdomainname` sometimes does not work
+    $self->{host_domain_name} = script_output("dnsdomainname", proceed_on_failure => 1);
     $self->{start_run} = time();
     return $self;
 }
@@ -401,9 +402,12 @@ sub prepare_non_transactional_environment {
         virt_autotest::utils::setup_rsyslog_host($common_log_folder);
         my $_packages_to_check = 'wget curl screen dnsmasq xmlstarlet yast2-schema python3 nmap';
         zypper_call("install -y $_packages_to_check");
-        my $_patterns_to_check = 'kvm_server kvm_tools';
-        $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
-        zypper_call("install -y -t pattern $_patterns_to_check");
+        # There is already the highest version for kvm/xen packages on TW
+        if (is_sle) {
+            my $_patterns_to_check = 'kvm_server kvm_tools';
+            $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
+            zypper_call("install -y -t pattern $_patterns_to_check");
+        }
     }
     return $self;
 }
@@ -1139,7 +1143,7 @@ sub config_guest_network_bridge_device {
     my $_bridge_network_in_route = shift;
     my $_bridge_device = shift;
     $_bridge_device //= $self->{guest_network_device};
-    if ((script_output("ip route show | grep -o $_bridge_device", proceed_on_failure => 1) eq '') and (script_output("ip route show | grep -o $_bridge_network_in_route", proceed_on_failure => 1) eq '')) {
+    unless ((script_run("ip route show | grep -o $_bridge_device") == 0) or (script_run("ip route show | grep -o $_bridge_network_in_route") == 0)) {
         my $_detect_active_route = '';
         my $_detect_inactive_route = '';
         if ($self->{guest_netaddr} ne 'host-default') {
--- a/lib/guest_installation_and_configuration_base.pm
+++ b/lib/guest_installation_and_configuration_base.pm
@@ -33,6 +33,7 @@ use IO::Scalar;
 use List::Util qw(first);
 use testapi;
 use utils;
+use ipmi_backend_utils qw(reconnect_when_ssh_console_broken);
 use virt_utils;
 use virt_autotest::utils;
 use virt_autotest::virtual_network_utils;
@@ -1734,15 +1735,18 @@ sub start_guest_installation {
     my $_guest_installation_dryrun_log = "$common_log_folder/$self->{guest_name}/$self->{guest_name}" . "_installation_dryrun_log_" . $_start_installation_timestamp;
     my $_guest_installation_log = "$common_log_folder/$self->{guest_name}/$self->{guest_name}" . "_installation_log_" . $_start_installation_timestamp;
     assert_script_run("touch $_guest_installation_log && chmod 777 $_guest_installation_log");
-    if (script_run("set -o pipefail;$self->{virt_install_command_line_dryrun} 2>&1 | tee -a $_guest_installation_dryrun_log", timeout => 600 / get_var('TIMEOUT_SCALE', 1)) ne 0) {
-        record_info("Guest $self->{guest_name} installation dry run failed", "The virt-install command used is $self->{virt_install_command_line_dryrun}");
-        $self->record_guest_installation_result('FAILED');
+    # Dry run always timeout when downloading initrd from download.opensuse.org in O3
+    my $ret = script_run("set -o pipefail; $self->{virt_install_command_line_dryrun} 2>&1 | tee -a $_guest_installation_dryrun_log", timeout => 600 / get_var('TIMEOUT_SCALE', 1), die_on_timeout => 0);
+    save_screenshot;
+    unless (defined(script_run('set +o pipefail', die_on_timeout => 0))) {
+        reconnect_when_ssh_console_broken;
         script_run("set +o pipefail");
-        save_screenshot;
+    }
+    if ($ret ne 0) {
+        record_info("Guest $self->{guest_name} installation dry run failed", "The virt-install command used is $self->{virt_install_command_line_dryrun}", result => 'fail');
+        $self->record_guest_installation_result('FAILED');
         return $self;
     }
-    script_run("set +o pipefail");
-    save_screenshot;
     record_info("Guest $self->{guest_name} installation dry run succeeded", "Going to install by using $self->{virt_install_command_line}");
     #Use "screen" in the most compatible way, screen -t "title (window's name)" -c "screen configuration file" -L(turn on output logging) "command to run".
     #The -Logfile option is only supported by more recent operating systems.
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -2303,7 +2303,7 @@ sub permit_root_ssh_in_sol {
     my $sshd_config_file = shift;
 
     $sshd_config_file //= "/etc/ssh/sshd_config";
-    enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
+    enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)", wait_still_screen => 5);
 }
 
 =head2 cleanup_disk_space
--- a/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
+++ b/tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
@@ -19,7 +19,7 @@ use base 'virt_autotest_base';
 use testapi;
 use ipmi_backend_utils;
 use Utils::Backends qw(use_ssh_serial_console is_remote_backend);
-use utils qw(zypper_call systemctl permit_root_ssh_in_sol);
+use utils qw(zypper_call systemctl permit_root_ssh_in_sol script_retry);
 use virt_autotest::utils qw(is_kvm_host is_xen_host);
 
 sub run {
@@ -49,7 +49,7 @@ sub run {
     die 'Need one of both to be true: is_kvm_host || is_xen_host' unless is_kvm_host || is_xen_host;
     my $hypervisor = is_kvm_host ? 'kvm' : 'xen';
     zypper_call('--gpg-auto-import-keys ref');
-    zypper_call("in -t pattern ${hypervisor}_server ${hypervisor}_tools", 1800);
+    script_retry("zypper -n in -t pattern ${hypervisor}_server ${hypervisor}_tools", timeout => 1800, retry => 5, delay => 10);
     set_grub_on_vh('', '', $hypervisor);
     systemctl 'enable libvirtd', ignore_failure => 1;
 


Julie-CAO:
@alice-suse @xguo @waynechen55 @nanzhg @tbaev @RoyCai7 @varunkojha   It's most about TumbleWeed tests but be related to Unified guest installation tests. Welcome review!

okurz:
s/laterly/lately/

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

You might want to use the more up-to-date tool https://github.com/cli/cli

At the list the ticket reference can be helpul in the git commit messages

lib/guest_installation_and_configuration_base.pm
```diff
@@ -401,9 +401,12 @@ sub prepare_non_transactional_environment {
         virt_autotest::utils::setup_rsyslog_host($common_log_folder);
         my $_packages_to_check = 'wget curl screen dnsmasq xmlstarlet yast2-schema python3 nmap';
         zypper_call("install -y $_packages_to_check");
-        my $_patterns_to_check = 'kvm_server kvm_tools';
-        $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
-        zypper_call("install -y -t pattern $_patterns_to_check");
+        # There is already the highest version for kvm/xen packages on TW
+        unless (is_tumbleweed) {
```
okurz, line 405: Our tests should always consider openSUSE Tumbleweed as the default product without a version. That means any differing behaviour should have an explicit exclude rule for the "older products" and potentially an explanation why the behaviour should differ. You can try to negate the logic check.

lib/utils.pm
```diff
@@ -2300,7 +2300,7 @@ sub permit_root_ssh_in_sol {
     my $sshd_config_file = shift;
 
     $sshd_config_file //= "/etc/ssh/sshd_config";
-    enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)");
+    enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)", wait_still_screen => 5);
```
okurz, line 2303: You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

okurz (changes requested):

Julie-CAO:
> s/laterly/lately/

Thanks. corrected! 

> What you put into the description of the pull request could also be in the commit message details. I don't know how you created your pull request but your git commit messages has _only_ the subject line while the pull request description has more details. Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage.

Just appended these github description to the commit message. 

> I can recommend the tool [hub](https://github.com/github/hub) (zypper in rubygem-hub) for easier PR creation. Also, I myself use a script [git-pr-last](https://github.com/okurz/scripts/tree/master/git-pr-last) to create a PR with proper description for these simple one-commit PRs:
> 
> ```shell
> $ cat $(which git-pr-last )
> #!/bin/sh -e
> target="${target:-"$USER"}"
> git push $target && git show --no-patch --format=%B | hub pull-request -F -
> ```
> 
> You might want to use the more up-to-date tool https://github.com/cli/cli
> 
> See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details
> 
> At the list the ticket reference can be helpul in the git commit messages

I really appreciate your recomendation. I just zypper in gh to play with the tool. It would be great if gh makes github things easier.

lib/guest_installation_and_configuration_base.pm
Julie-CAO: I changed it to `if (is_sle)`.

lib/utils.pm
Julie-CAO: Thank you for the detailed explaination. I understand 'sleep' should be avoided in our tests as we could. I explain why it had to be used here:

- The command is running in SOL console to start sshd service. It is necessary because ssh root console is not available until sshd service is restarted.
- SOL console(ipmi Serial Over Lan console) has very limited support in openQA, fg. script_output() and assert_script_run() are not supported well. It means any checks for some conditions are not easy to be implemented and even with some unexpected results.
- SOL console is not stable. We avoid run command on it as we could as possible. An extra check would make the test be at risk for failing.
- I have monitored in a few test cases, simply waiting for 5s will make sure sshd service is brought up. 

So I think 'sleep' in this PR is a smart move.

lib/guest_installation_and_configuration_base.pm
```diff
@@ -401,9 +402,12 @@ sub prepare_non_transactional_environment {
         virt_autotest::utils::setup_rsyslog_host($common_log_folder);
         my $_packages_to_check = 'wget curl screen dnsmasq xmlstarlet yast2-schema python3 nmap';
         zypper_call("install -y $_packages_to_check");
-        my $_patterns_to_check = 'kvm_server kvm_tools';
-        $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
-        zypper_call("install -y -t pattern $_patterns_to_check");
+        # There is already the highest version for kvm/xen packages on TW
+        if (is_sle) {
```
okurz, line 406: I don't think this is correct as the versions of packages or patterns isn't depending on the distribution type. How about

```suggestion
        if (is_sle('<15-SP5') || is_leap('<15.5')) {
```

Julie-CAO: No, If a fixed version would be used at here, we will have to update it when we are in SLE-next. Meanwhile on the latest SLE/Leap(fg. 15SP5 for now), a 'zypper in/update' usually does not cause troubles.

I remove it from TW because zypper failed sporatically from download.opensuse.org lately.

okurz: > No, If a fixed version would be used at here, we will have to update it when we are in SLE-next. Meanwhile on the latest SLE/Leap(fg. 15SP5 for now), a 'zypper in/update' usually does not cause troubles.
> 
> I remove it from TW because zypper failed sporatically from download.opensuse.org lately.

So you are not calling zypper on Tumbleweed because the mirroring infrastructure of openSUSE has problems? That does not sound right. Tumbleweed and Leap use the same download.opensuse.org. Then I suggest you check if the necessary version of packages is installed. If they are already there, regardless if it's Tumbleweed or Leap or SLE you can skip the installation.

Julie-CAO: Obviously TW always has the latest packages installed. Meanwhile, It would not helpful to check the versions as access to download.opensuse.org via zypper is still required.
Leap can be skipped as we have not tests for Leap yet.

okurz (approved):
Ok then

lib/utils.pm
alice-suse, line 2303: I guess there will be still chances that permit_root_ssh_in_sol fail since typing commands in sol console usually have errors.  It won't be fully solved until we switch to autoyast host installation.  We can prioritize task if it indeed is quite unstable here. HDYT?

alice-suse, line 2303: I guess there will be still chances that permit_root_ssh_in_sol fail since typing commands in sol console usually have errors.  It won't be fully solved until we switch to autoyast host installation.  We can prioritize task if it indeed is quite unstable here. Of course not in this PR, but later. HDYT?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -19,25 +19,25 @@ use base 'virt_autotest_base';
...
     }
 
     if (check_screen('text-login')) {
-        enter_cmd "root";
-        assert_screen "password-prompt";
-        type_password;
-        send_key('ret');
+        sol_login();
+        # Try one more time as TW login fails with incorrect password sporatically in O3
+        sol_login() if check_screen('sol-login-password-incorrect', 60);
```
alice-suse, line 40: Will one more time try enough? How about do it in a  while for more times?

alice-suse (commented):
It seems most unstability fixes happen on sol console when setting up host at the beginning. It makes the need to switch to autoyast host installation even higher to me.

alice-suse:
I finally get time to review the PR. Sorry for the delay.

lib/utils.pm
Julie-CAO: Good idea. It is said that this sshd configuration is required by SLE-next as well.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -49,7 +49,7 @@ sub run {
     die 'Need one of both to be true: is_kvm_host || is_xen_host' unless is_kvm_host || is_xen_host;
     my $hypervisor = is_kvm_host ? 'kvm' : 'xen';
     zypper_call('--gpg-auto-import-keys ref');
-    zypper_call("in -t pattern ${hypervisor}_server ${hypervisor}_tools", 1800);
+    script_retry("zypper -n in -t pattern ${hypervisor}_server ${hypervisor}_tools", timeout => 1800, retry => 5, delay => 10);
```
alice-suse, line 52: This may increase the possibility to pass this step of pkg installation, but the job time will be much longer, which is non-negligible sacrifice.

@okurz Hi Oliver, we are not very familiar with O3 infra setup. Can you give some help here? Is it known unstable/badly network performed when installing packages from tumbleweed repo on O3? Or are we actually not using the best way to do such operations? How do you recommend?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: Sol console is not so unstable that password has been typed incorrectly more times. I would prefer to believe that password is not set correctly during installation. But it needs more verification in product builds.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -68,6 +68,14 @@ sub run {
 
 }
 
+sub sol_login {
```
alice-suse, line 71: How about moving it to [ipmi_backend_utils.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/ipmi_backend_utils.pm)?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
alice-suse: Of course, up to the need. We can consider it if needed.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: Oliver is aware that download.opensuse.org is unstable and showed me two PRs via slack. https://suse.slack.com/archives/C02CANHLANP/p1674025447904969

From our side, what we can do is to make test more resilient by retry or other methods. 

Retry is helpful in some cases , for example we can retry zypper. But I fail to find a good way figure out the download failure in 'virt-install' command, fg. https://openqa.opensuse.org/tests/3096451#step/unified_guest_installation/296

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: Yes, good suggestion. The functionis expected to be used in yy vgpu test as well.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
okurz: Let me provide the references here directly. For example there is https://github.com/openSUSE/zypper/issues/399 which describes not the same symptoms but maybe is linked to the same root cause mentioning https://progress.opensuse.org/issues/123172. There is also https://github.com/openSUSE/zypper/issues/420.

In the meantime https://progress.opensuse.org/issues/123172 was resolved (on 2023-01-26) so I suggest if you still see problems then make sure there actually is a corresponding error report for the corresponding compoment, e.g. in https://progress.opensuse.org/projects/opensuse-admin/issues

lib/utils.pm
alice-suse: @Julie-CAO  Glad to know that you have the same idea :). Do we have ticket tracking this? Or shall we create a new one for it?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
alice-suse: @Julie-CAO Based on oliver's above suggestion, do you think we should open a ticket for it in https://progress.opensuse.org/projects/opensuse-admin/issues (project for opensuse infra issues)?

lib/utils.pm
Julie-CAO: Are you talking about a ticket to allow root ssh login via autoyast host installation? If so, there has not been this ticket yet.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: I commented in https://github.com/openSUSE/zypper/issues/399#issuecomment-1386668065. As https://progress.opensuse.org/issues/123172 was resolved (on 2023-01-26) , I will monitor a few more builds before I open a new ticket.

I plan to raise up this kind of infra issue after we fixed some unstable issues from our end firstly. The tests will show more realiable results at that time.

lib/utils.pm
alice-suse: > @Julie-CAO Glad to know that you have the same idea :). Do we have ticket tracking this? Or shall we create a new one for it?

I had a search, seems no ticket for it yet, so create https://progress.opensuse.org/issues/124149.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
alice-suse: Sounds good.

Julie-CAO: And I found things are not just on the opensuse side, in some cases, TW kernel downloads failed only in SUTs in O3 VLAN. I need more cases to confirm. I'll open a ticket to the right component when I get time to have it done.

Julie-CAO: Posted a comment on https://progress.opensuse.org/issues/123328. Will keep an eye on it and determine what to do next.

Julie-CAO:
More verification runs passed. The PR with multiple commits is ready for merge. @alice-suse 

TW xen: https://openqa.opensuse.org/tests/3108424
SLE kvm: http://openqa.suse.de/tests/10462260
SLE xen: https://openqa.suse.de/tests/10462257

alice-suse:
> More verification runs passed. The PR with multiple commits is ready for merge. @alice-suse
> 
> TW xen: https://openqa.opensuse.org/tests/3108424 SLE kvm: http://openqa.suse.de/tests/10462260 SLE xen: https://openqa.suse.de/tests/10462257

Thanks Julie.

lib/guest_installation_and_configuration_base.pm
```diff
@@ -1734,15 +1735,18 @@ sub start_guest_installation {
...
+    # Dry run always timeout when downloading initrd from download.opensuse.org in O3
+    my $ret = script_run("set -o pipefail; $self->{virt_install_command_line_dryrun} 2>&1 | tee -a $_guest_installation_dryrun_log", timeout => 600 / get_var('TIMEOUT_SCALE', 1), die_on_timeout => 0);
+    save_screenshot;
+    unless (defined(script_run('set +o pipefail', die_on_timeout => 0))) {
+        reconnect_when_ssh_console_broken;
         script_run("set +o pipefail");
-        save_screenshot;
+    }
+    if ($ret ne 0) {
+        record_info("Guest $self->{guest_name} installation dry run failed", "The virt-install command used is $self->{virt_install_command_line_dryrun}", result => 'fail');
```
alice-suse, line 1746: @Julie-CAO Hi Julie, do I understand correctly that with this commit , you'd like to  mainly add record_info for dry run failure, which in previous code logic it dies directly when timeout arrives?

Julie-CAO: I guess it was not Wayne's intention because

- usually timeout have not happened for SLE
- test is supposed to be continued as there are still other guest need to be installed
- post_fail_hook will not be called when script timeout. It means there are no logs uploaded.

So my change will make test continue and post_fail_hook will be called at the end of the guest installation module.

alice-suse: I get the point now. Thanks for explanation. Make sense to me.

alice-suse (approved):

lib/guest_installation_and_configuration_base.pm
```diff
@@ -253,7 +254,8 @@ sub initialize_guest_params {
     $self->{$_} //= '' foreach (keys %guest_params);
     $self->{host_ipaddr} = get_required_var('SUT_IP');
     $self->{host_name} = script_output("hostname");
-    $self->{host_domain_name} = script_output("dnsdomainname");
+    # For SUTs with multiple interfaces, `dnsdomainname` sometimes does not work
+    $self->{host_domain_name} = script_output("dnsdomainname", proceed_on_failure => 1);
```
waynechen55, line 258: Can I have a look on this machine ?

lib/guest_installation_and_configuration_base.pm
```diff
@@ -401,9 +403,12 @@ sub prepare_non_transactional_environment {
...
         my $_packages_to_check = 'wget curl screen dnsmasq xmlstarlet yast2-schema python3 nmap';
         zypper_call("install -y $_packages_to_check");
-        my $_patterns_to_check = 'kvm_server kvm_tools';
-        $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
-        zypper_call("install -y -t pattern $_patterns_to_check");
+        # There is already the highest version for kvm/xen packages on TW
+        if (is_sle) {
+            my $_patterns_to_check = 'kvm_server kvm_tools';
+            $_patterns_to_check = 'xen_server xen_tools' if ($self->{host_virt_type} eq 'xen');
+            zypper_call("install -y -t pattern $_patterns_to_check");
```
waynechen55, line 410: ```if （!is_tubleweed)``` better ?

lib/guest_installation_and_configuration_base.pm
```diff
@@ -1730,15 +1735,18 @@ sub start_guest_installation {
...
     my $_guest_installation_log = "$common_log_folder/$self->{guest_name}/$self->{guest_name}" . "_installation_log_" . $_start_installation_timestamp;
     assert_script_run("touch $_guest_installation_log && chmod 777 $_guest_installation_log");
-    if (script_run("set -o pipefail;$self->{virt_install_command_line_dryrun} 2>&1 | tee -a $_guest_installation_dryrun_log", timeout => 600 / get_var('TIMEOUT_SCALE', 1)) ne 0) {
-        record_info("Guest $self->{guest_name} installation dry run failed", "The virt-install command used is $self->{virt_install_command_line_dryrun}");
-        $self->record_guest_installation_result('FAILED');
+    # Dry run always timeout when downloading initrd from download.opensuse.org in O3
+    my $ret = script_run("set -o pipefail; $self->{virt_install_command_line_dryrun} 2>&1 | tee -a $_guest_installation_dryrun_log", timeout => 600 / get_var('TIMEOUT_SCALE', 1), die_on_timeout => 0);
+    save_screenshot;
+    unless (defined(script_run('set +o pipefail', die_on_timeout => 0))) {
+        reconnect_when_ssh_console_broken;
```
waynechen55, line 1742: Why add reconnect_when_ssh_console_broken here ?

lib/ipmi_backend_utils.pm
```diff
@@ -544,4 +544,12 @@ sub reconnect_when_ssh_console_broken {
     save_screenshot;
 }
 
+sub sol_login {
+    enter_cmd "root";
+    assert_screen "password-prompt";
+    type_password;
+    send_key('ret');
+    save_screenshot;
+}
```
waynechen55, line 553: This is just a wrapper function ?

Julie-CAO: yes.

lib/utils.pm
waynechen55, line 2303: Can this line be improved a bit ?
```
enter_cmd("[ `grep \"^PermitRootLogin *yes\" $sshd_config_file | wc -l` -gt 0 ] || (echo 'PermitRootLogin yes' >>$sshd_config_file; systemctl restart sshd)", wait_still_screen => 5);
```
to 
```
sed 's/a/bg'
or 
sed '/a/d' then echo /b/
```
?

lib/guest_installation_and_configuration_base.pm
Julie-CAO: because ssh console is broken when virt-install timeouts.  fg. https://openqa.opensuse.org/tests/3106501#step/unified_guest_installation/304. we have to reconnect ssh console to continue tests.

lib/guest_installation_and_configuration_base.pm
Julie-CAO: I appericiate you are ready to help, but the SUT is in O3 VLAN. Only The authorized people(based on sshkeys)  can access it from a jump server.  A easier way I can think of is to access the SUT from my laptop when we are at office, Let's talk offline. but I think we'd better let the fix in firstly. then we can try to find the root cause and fix it in no hurry. The root cause I can think of is that the two interfaces are in different VLAN, which may confuse 'dnsdomainname` at times.

lib/guest_installation_and_configuration_base.pm
waynechen55: I do not think this is quite necessary, because there is "check_root_ssh_console" in post_fail_hook in concurrent_guest_installation. If set +o pipefail fails, just let it go. So remove this reconnect_when_ssh_console_broken ?

lib/guest_installation_and_configuration_base.pm
Julie-CAO: 'is_tubleweed' does not follow the code convention about factory-first which was warned by Oliver a few times above. It should be used with limitation. Is 'is_sle' at  here correct? Are there any other products such as sle_micro or ALP required as well? If so, I have to reconsider of 'is_tumbleweed'.

lib/guest_installation_and_configuration_base.pm
waynechen55: I think you can try on your laptop when you in the office, because your laptop has connections on both ethx and wlanx. Just check the return value of dnsdomainname, because there is no domain name on your laptop.

lib/guest_installation_and_configuration_base.pm
waynechen55: 1.This function is prepare_non_transactional_environment, so only take effect on non-transactional ones. Maybe you can keep it like this at the moment, and add more checking later if necessary. 
2.If you just want to skip this part because there is already highest version of them, I think you do not need to do anything, just leave it as what it was before.

lib/guest_installation_and_configuration_base.pm
waynechen55: At the moment, I have no access to AMD ZEN3 machine which also has multiple interfaces. There is no such issue on it.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
waynechen55, line 40: I personally do not prefer a new needle like this on sol console, especially an "incorrect one". I think 
```if !check_screen("logged in as root in sole")  retry sol_login```
is better.

lib/guest_installation_and_configuration_base.pm
Julie-CAO: No, reconnect is not just for post_fail_hook. Test should be continued when ssh console is broken, because other guests are expected to be installed as well.

lib/guest_installation_and_configuration_base.pm
Julie-CAO: > 2. If you just want to skip this part because there is already highest version of them, I think you do not need to do anything, just leave it as what it was before.

because zypper still do some fresh or querry work to connect remote repo, download.opensuse.org, which has risks of timeout or failures. I am trying to avoid any unnecessary connection to download.opensuse.org as its unstable problem seems not to be fixed in short time.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: The test still experiencing password typing problem. I leave the commit in next PR, only merge other commits in the PR.

Julie-CAO:
Move one commit about sol login to next PR and leave 3 commits in this PR. Verfified again: https://openqa.opensuse.org/tests/3118974

The PR is ready to be merged when the CI checks in github passed, which has been waiting for quite some time.

Julie-CAO:
CI checks passed. @alice-suse , Could you help merge this PR?

PR #16262, opened 2023-01-19
syscall_thrasher.pm: output to file instead of video

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/123223
- Verification run: https://openqa.suse.de/tests/10348012

### DIFF
--- a/tests/security/atsec/syscall_thrasher.pm
+++ b/tests/security/atsec/syscall_thrasher.pm
@@ -13,37 +13,40 @@ use warnings;
 use testapi;
 use utils;
 
+my $log_file = '/tmp/syscalls_output.log';
+
 sub run {
     my ($self) = shift;
 
     select_console 'root-console';
 
-    # Complile thrash.c
     my $exe_file = 'thrash';
     assert_script_run('cd /usr/local/atsec');
     assert_script_run("gcc -o $exe_file thrash.c");
 
-    # Change the permission of test script because we need to run it as non-root
     assert_script_run("chmod 755 $exe_file");
 
     # The test needs to run by non-root
     select_console 'user-console';
 
-    # Prepare the test directory
     my $test_dir = 'test_syscall_thrasher';
     assert_script_run("mkdir -p $test_dir");
 
-    # Copy the executable file to test directory
     assert_script_run("cp /usr/local/atsec/$exe_file $test_dir/");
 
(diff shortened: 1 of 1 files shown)


tests/security/atsec/syscall_thrasher.pm
```diff
@@ -18,28 +18,24 @@ sub run {
...
 
-    # Copy the executable file to test directory
     assert_script_run("cp /usr/local/atsec/$exe_file $test_dir/");
 
-    # Run the test
     assert_script_run("cd $test_dir");
-    assert_script_run("./$exe_file", timeout => 1200);
-    assert_script_run('ls');
+    assert_script_run("./$exe_file >> $log_file", timeout => 900);
+    upload_logs("$log_file");
```
Vogtinator, line 38: This means that on failure it will not upload logs

paolostivanin: good point! let me fix that

punkioudi (approved):
lgtm!

tests/security/atsec/syscall_thrasher.pm
```diff
@@ -13,37 +13,38 @@ use warnings;
...
 }
 
 sub test_flags {
     return {always_rollback => 1};
 }
 
+sub post_fail_hook {
+    upload_logs("$log_file");
```
Vogtinator, line 47: I think this needs to chain to the parent function

paolostivanin: what do you mean? AFAIK, this function is triggered automatically when a test fails (unless SKIP_POST_FAIL_HOOK is set, oc)

Vogtinator: This overrides the default `post_fail_hook` of the `consoletest` base class, so it will only attempt to upload this log but not the journal and other files anymore.

Vogtinator: You need

```
sub post_fail_hook {
    my ($self) = @_;
    upload_logs("$log_file");
    $self->SUPER::post_fail_hook;
}
```

paolostivanin: interesting, thanks a lot Fabian, I didn't know that.

Vogtinator (approved):

PR #16266, opened 2023-01-20
Install Public Cloud Module before starting migration

asmorodskyi:


asmorodskyi:
VR: http://autobot.qa.suse.de/tests/3604#step/migration/144 ( failing but after provided changes so failure is unrelated )

PR #16269, opened 2023-01-20
Fix volume mount failure on kubevirt agent

nanzhg:
To resolve volume mounting failure while running datavolume_test, it needs to make a workaround for sle-micro.

https://openqa.suse.de/tests/10337166
http://10.67.129.96/tests/1981
"message": "MountVolume.SetUp failed for volume \"local-storage\" : mkdir /mnt/local-storage: read-only file system"

- Related ticket: https://progress.opensuse.org/issues/123415
- Verification run: https://openqa.suse.de/tests/10738406

### DIFF
--- a/tests/virt_autotest/kubevirt_tests_agent.pm
+++ b/tests/virt_autotest/kubevirt_tests_agent.pm
@@ -90,7 +90,7 @@ sub rke2_agent_setup {
     $self->check_service_status();
 
     # Start rke2-agent service ready
-    mutex_create('RKE2_AGENT_START_READY');
+    mutex_create('rke2_agent_start_ready');
 
     assert_script_run("mkdir -p ~/.kube; scp root\@$server_ip:/etc/rancher/rke2/rke2.yaml ~/.kube/config");
     assert_script_run("sed -i 's/127.0.0.1/$server_ip/' ~/.kube/config");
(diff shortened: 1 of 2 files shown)


tests/virt_autotest/kubevirt_tests_agent.pm
```diff
@@ -90,7 +90,7 @@ sub rke2_agent_setup {
     $self->check_service_status();
 
     # Start rke2-agent service ready
-    mutex_create('RKE2_AGENT_START_READY');
+    mutex_create('rke2_agent_start_ready');
```
waynechen55, line 93: Does lower case make any difference ?

tests/virt_autotest/kubevirt_tests_agent.pm
```diff
@@ -106,8 +106,11 @@ sub rke2_agent_setup {
     $self->check_service_status();
     assert_script_run("grep static /var/lib/kubelet/cpu_manager_state");
 
+    # Workaround for failure 'MountVolume.SetUp failed for volume "local-storage" : mkdir /mnt/local-storage: read-only file system'
+    assert_script_run('mkdir -p /root/tmp && mount -o bind /root/tmp /mnt') if (is_transactional);
```
waynechen55, line 110: Does this mean your new volume dir is /root/tmp ?

tests/virt_autotest/kubevirt_tests_server.pm
```diff
@@ -288,7 +288,10 @@ sub run_virt_tests {
     # Ensure Config::Tiny module installed
     assert_script_run('cpan install Config::Tiny <<<yes', timeout => 300) if (script_run('cpan -l <<<yes | grep Config::Tiny') == 1);
 
-    transactional::exit_trup_shell_and_reboot() if (is_transactional);
+    if (is_transactional) {
+        transactional::exit_trup_shell_and_reboot();
+        assert_script_run('mount -o bind /root/tmp /mnt');
+    }
```
waynechen55, line 294: Why there is no ```mkdir -p /root/tmp``` ?

tests/virt_autotest/kubevirt_tests_agent.pm
nanzhg: Created lock name in server side is lower case, so just want to keep consistent with the server side.

tests/virt_autotest/kubevirt_tests_agent.pm
nanzhg: Yes, some test files will be created under /mnt during test running, and the directory is read-only. So it's a workaround for sle-micro.

tests/virt_autotest/kubevirt_tests_server.pm
nanzhg: Updated.

Julie-CAO:
LGTM

alice-suse (approved):

PR #16271, opened 2023-01-20
Call show_tasks_in_blocked_state in post_fail_hook

dzedro:
- Related ticket: https://progress.opensuse.org/issues/120723
- Verification run: https://openqa.suse.de/tests/overview?build=dzedro%2Fos-autoinst-distri-opensuse%2316271&distri=sle

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1644,10 +1644,13 @@ See L<https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/sy
 sub show_tasks_in_blocked_state {
     # sending sysrqs doesn't work for svirt
     if (has_ttys) {
+        my $has_logger = script_run('test -x /usr/bin/logger') == 0;
+        script_run('logger "### Beginning of show_tasks_in_blocked_state"') if $has_logger;
         send_key 'alt-sysrq-t';
         send_key 'alt-sysrq-w';
         # info will be sent to serial tty
         wait_serial(qr/sysrq\s*:\s+show\s+blocked\s+state/i);
+        script_run('logger "### End of show_tasks_in_blocked_state"') if $has_logger;
 
         # If the 'An error occured during the installation.' OK popup has popped up,
         # do not press the 'return' key, because it will result in all ttys logging out.
--- a/lib/consoletest.pm
+++ b/lib/consoletest.pm
@@ -55,6 +55,7 @@ sub post_fail_hook {
     return if (is_public_cloud() || is_openstack());
     select_console('log-console');
     remount_tmp_if_ro;
+    show_tasks_in_blocked_state;
     export_logs_basic;
     # Export extra log after failure for further check gdm issue 1127317, also poo#45236 used for tracking action on Openqa
     export_logs_desktop;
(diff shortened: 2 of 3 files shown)


foursixnine (approved):

lib/utils.pm
```diff
@@ -1644,10 +1644,13 @@ See L<https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/sy
 sub show_tasks_in_blocked_state {
     # sending sysrqs doesn't work for svirt
     if (has_ttys) {
+        my $has_logger = script_run('test -x /usr/bin/logger') == 0;
+        script_run('logger "### Beginning of show_tasks_in_blocked_state"') if $has_logger;
```
okurz, line 1648: I would not do that. There are certain cases when such commands would also block indefinitely and the next sysrq's that would not be blocked are never executed.

dzedro: It's part of acceptance criteria, but it's not required.
I can't think of how it could "block indefinitely" ? Can you explain ?

okurz: well, if a Linux is not responsive anymore in the userspace and hence not able to accept even the input of characters on a readline then that `script_run` wouldn't be executed. Ok, it might be "executed" in the sense that openQA would just type those characters and the system just ignores those as it's not `assert_script_run` so we don't wait for any effect or response. But in contrast the magic-sysrq have the benefit of still being responsive even if everything but the kernel itself is locked up.

dzedro: If system is not responding anymore, then it does not matter sysrq will work, because you will not upload the response.

okurz: well, the sysrq output should directly go the serial port which is still directly controlled by the kernel so that's the reason why we (or me?) added the sysrq commands here in the first place

dzedro: Looking on the VRs, I don't see the alt-sysrq-t, alt-sysrq-w output in serial log.
Anyway this cosmetical message is not required,
but I don't think it will block anything because the `script_run` failure does not do that,
it will behave same as `/usr/bin/logger` is not present or not ?

foursixnine: @dzedro I think last time I looked at the PR, I noted that none of the VR's had the log output, but I forgot to mention it. (that's why I dismissed my review)...

> There are certain cases when such commands would also block indefinitely and the next sysrq's that would not be blocked are never executed.

This is premature optimization; from experience and knowledge of how soft-lockups are happening, a script_run won't hang forever, and if it does, then we deal with it when that happens (like doing the check once the first console login and using variables instead to know it is supported), we will know fast enough and should be able to react fast enough for this not to become a problem.

>  Anyway this cosmetical message is not required,

Indeed, what's important is the alt-sys-req 🚀

dzedro: @foursixnine each xterm-journal.log  and clamav-journal.log  should have it. https://openqa.suse.de/tests/10368409/file/xterm-journal.log

foursixnine (approved):

okurz (approved):

PR #16276, opened 2023-01-24
Public Cloud Azure ARM PUBLIC_CLOUD_IMAGE_URI=auto

pdostal:
When the image is already uploaded, we can simply calculate it's URI.

- Related ticket: [poo#116698](https://progress.opensuse.org/issues/116698)
- Verification run: [PUBLIC_CLOUD_IMAGE_URI=XYZ](https://pdostal-server.suse.cz/tests/3912), [PUBLIC_CLOUD_IMAGE_URI=auto](https://pdostal-server.suse.cz/tests/3911)

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -250,6 +250,31 @@ sub get_image_id {
     return $image_id;
 }
 
+=head2 get_image_uri
+
+Retrieves the CSP image uri if exists, otherwise exception is thrown.
+This is currently used specifically in Azure so the subroutine will die afterwards.
+=cut
+
+sub get_image_uri {
+    my ($self) = @_;
+    die 'The PUBLIC_CLOUD_IMAGE_URI variable makes sense only for Azure' unless is_azure;
+    my $image_uri = get_var("PUBLIC_CLOUD_IMAGE_URI");
+    if ($image_uri =~ /^auto$/mi) {
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        my $version = $self->calc_img_version();    # PUBLIC_CLOUD_BUILD PUBLIC_CLOUD_BUILD_KIWI
+        my $subscriptions = $self->provider_client->subscription;
+        my $resource_group = $self->resource_group;
+        my $image_gallery = $self->image_gallery;
+        $image_uri = "/subscriptions/$subscriptions/resourceGroups/$resource_group/providers/";
+        $image_uri .= "Microsoft.Compute/galleries/$image_gallery/images/$definition/versions/$version";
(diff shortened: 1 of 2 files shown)


lib/publiccloud/provider.pm
```diff
@@ -250,6 +250,32 @@ sub get_image_id {
...
+
+Retrieves the CSP image uri if exists, otherwise exception is thrown.
+This is currently used specifically in Azure so the subroutine will die afterwards.
+=cut
+
+sub get_image_uri {
+    my ($self) = @_;
+    die 'The PUBLIC_CLOUD_IMAGE_URI variable makes sense only for Azure' unless is_azure;
+    my $image_uri = get_var("PUBLIC_CLOUD_IMAGE_URI");
+    record_info('XXXXX', 'X'.$image_uri.'X');
```
grisu48, line 263: Leftover?

lib/publiccloud/provider.pm
```diff
@@ -250,6 +250,32 @@ sub get_image_id {
...
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        my $version = $self->calc_img_version(); # PUBLIC_CLOUD_BUILD PUBLIC_CLOUD_BUILD_KIWI
+	my $subscriptions = $self->provider_client->subscription;
+	my $resource_group = $self->resource_group;
+	my $image_gallery = $self->image_gallery;
+        $image_uri = "/subscriptions/$subscriptions/resourceGroups/$resource_group/providers/";
+        $image_uri .= "Microsoft.Compute/galleries/$image_gallery/images/$definition/versions/$version";
+        record_info 'IMAGE_URI', "Computed IMAGE_URI=$image_uri";
+    } else {
+        record_info 'IMAGE_URI', "Provided IMAGE_URI=$image_uri";
```
grisu48, line 274: If the image URI is simply taken from the variable, then I don't think that this info is useful.

grisu48 (commented):
Two small comments, otherwise LGTM!

asmorodskyi (approved):
agree with both comments from Felix other than that LGTM

lib/publiccloud/provider.pm
pdostal: I wanted to make the test consistent - so the info box is always there. Is it OK or should I remove it?

pdostal: Also we see if the image URI is Calculated or provided.

asmorodskyi: it is ok to me to keep it . This is very important variable so it might be useful to have it value displayed

grisu48: ok

grisu48 (approved):

PR #16288, opened 2023-01-25
Stabilize clock and for journalctl check

mloviska:
`chrony-wait.service` is failing often in journalctl. The service needs to run in each boot so the SUT won't start with drifted time. In order to ensure minimal time drift, we need to make sure that `waitsync` call syncs the time properly.

- ticket: [15-SP5 Beta 4 Minimal-VM times out in enabling chrony](https://progress.opensuse.org/issues/123385)
- VR: [sle-15-SP4-JeOS-for-kvm-and-xen-Updates](http://kepler.suse.cz/tests/20312)

### DIFF
--- a/tests/console/journalctl.pm
+++ b/tests/console/journalctl.pm
@@ -160,8 +160,8 @@ sub run {
         }
     } else {
         assert_script_run("systemctl start chronyd");
-        assert_script_run("chronyc waitsync", fail_message => "time synchronization failed");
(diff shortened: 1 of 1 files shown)


jlausuch (approved):

jlausuch:
`script_retry` again saving us :)

jlausuch:
meeeeh https://openqa.suse.de/tests/10376316#step/journalctl/14
Well, it's better trying some times than only 1 :D

PR #16295, opened 2023-01-26
We need to exclude embargoed incidents.

pdostal:
- Related tickets: [poo#123655](https://progress.opensuse.org/issues/123655)
- Verification run: [pdostal-server.suse.cz/t3941](https://pdostal-server.suse.cz/tests/3941)

### DIFF
--- a/tests/publiccloud/download_repos.pm
+++ b/tests/publiccloud/download_repos.pm
@@ -14,6 +14,7 @@ use testapi;
 use strict;
 use utils;
 use publiccloud::ssh_interactive "select_host_console";
+use List::MoreUtils qw(uniq);
 
 # Get the status of the update repos
 # 0 = no repo, 1 = repos already downloaded, 2 = repos downloading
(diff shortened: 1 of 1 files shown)


tests/publiccloud/download_repos.pm
```diff
@@ -55,12 +56,21 @@ sub run {
         assert_script_run('touch /tmp/repos.list.txt');
 
         my $ret = 0;
+        my $reject = "'robots.txt,*.ico,*.png,*.gif,*.css,*.js,*.htm*'";
+        my $regex = "'s390x\\/|ppc64le\\/|kernel*debuginfo*.rpm|src\\/'";
         for my $maintrepo (@repos) {
-            next if $maintrepo !~ m/^http/;
```
asmorodskyi, line 59: what about this check ? looks like we missing it now , was that intentional ?

tests/publiccloud/download_repos.pm
```diff
@@ -55,12 +56,21 @@ sub run {
...
+            my $incident = ($1) while $maintrepo =~ /\/Maintenance:\/(\d+)/g;
+            die "We did not detect incident number for URL \"$maintrepo\". We detected \"$incident\"" if $incident =~ /\d+/;
```
alvarocarvajald: Won't this kill the test when `$incident` is a number/digit?

tests/publiccloud/download_repos.pm
pdostal: Well it was intentional, because:
1) I don't think that there will be problems and if so, then:
2) We were just blindly ignoring it until now.
3) If there will be problem with it, we will notice it and will be able to solve it.

pdostal:
New passing verification run in the description.

tests/publiccloud/download_repos.pm
```diff
@@ -55,12 +56,22 @@ sub run {
...
 
         my $ret = 0;
+        my $reject = "'robots.txt,*.ico,*.png,*.gif,*.css,*.js,*.htm*'";
+        my $regex = "'s390x\\/|ppc64le\\/|kernel*debuginfo*.rpm|src\\/'";
+        my $incident;
         for my $maintrepo (@repos) {
-            next if $maintrepo !~ m/^http/;
+            $incident = $1 while $maintrepo =~ /\/Maintenance:\/(\d+)/g;
+            die "We did not detect incident number for URL \"$maintrepo\". We detected \"$incident\"" unless $incident =~ /\d+/;
+            script_retry("curl -s https://build.suse.de/attribs/SUSE:Maintenance:$incident -o /tmp/$incident.txt");
```
alvarocarvajald, line 65: I think it would be a good idea to add a `-f` here so `curl` can return a failed retval in case the URL does not return what we expect:
```
# curl -s mango.qa.suse.de/non-existent -o non-existent ; echo $?
0
# md5sum non-existent ; rm non-existent
ddecdfa0919ead549d583b866effceb0  non-existent
# curl -s -f mango.qa.suse.de/non-existent -o non-existent ; echo $?
22
# md5sum non-existent ; rm non-existent
md5sum: non-existent: No such file or directory
rm: cannot remove 'non-existent': No such file or directory
```
Specially important since L66 is assuming "non embargoed" as the default, and only skipping it if the file downloaded by curl matches with `OBS:EmbargoDate`.

pdostal: Wow, how come I did know this before? Thank you!

asmorodskyi (approved):
LGTM ! Well done Pavel !

alvarocarvajald (approved):
LGTM

PR #16298, opened 2023-01-26
Provide workaround for iscsi failure bsc#1207157

ge0r:
Iscsi MU tests for 15 SP3 are failing due to yast2-iscsi-client return with an unexpected exit code. This is an issue that has already appeared on 15 SP5, was addressed with bsc#1206132 and subsequently fixed.
On 15 SP3, while waiting for the fix to land, in order to not block maintainance updates due to failure we apply the fix manually, as workaround, which is pretty straightforward requires some changes to iscsid service file.

- Related ticket: https://progress.opensuse.org/issues/123460
- Needles: No needles
- Verification run for 15 - SP3 : https://openqa.suse.de/tests/overview?distri=sle&version=15-SP3&build=ge0r%2Fos-autoinst-distri-opensuse%23iscsi-workaround
also VR for 15 - SP4 just to prove that it is not affected by the workaround and still works : https://openqa.suse.de/tests/10384717

### DIFF
--- a/lib/YaST/workarounds.pm
+++ b/lib/YaST/workarounds.pm
@@ -8,10 +8,11 @@ use strict;
 use warnings;
 use Exporter 'import';
 use testapi;
+use utils;
 use version_utils;
 
 
-our @EXPORT = qw(apply_workaround_bsc1204176);
+our @EXPORT = qw(apply_workaround_bsc1204176 apply_workaround_bsc1207157);
 
 =head1 Workarounds for known issues
 
@@ -46,6 +47,28 @@ sub apply_workaround_bsc1204176 {
             send_key('alt-f10', wait_screen_change => 1);
         }
     }
+}
+
+=head2 apply_workaround_bsc1207157 ():
+
+Workaround for the iscsi return code issue.
+
+Records a soft failure with a reference to bsc#1207157
+
+Changes the iscsid service file to require and start after the iscsid socket.
+Then reloads systemd, in order to scan for the changed unit.
+
+=cut
+
+sub apply_workaround_bsc1207157 {
+    record_soft_failure('bsc#1207157 - openQA test fails in iscsi_client - launched then no response');
+    my $service_unit = '/usr/lib/systemd/system/iscsid.service';
+    # append the two lines at the end of the [Unit] section of the service file, as specified in bsc#1206132
+    my $cmd = q(awk -i inplace 'BEGIN {c=1}; /^$/ {if (c){print("Requires=iscsid.socket\nAfter=iscsid.socket")} c=0}; {print}' );
+    $cmd .= $service_unit;
+
+    assert_script_run($cmd);
+    systemctl('daemon-reload');
 }
 
 1;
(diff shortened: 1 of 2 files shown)


ge0r:
Hmmm I was under the impression that for SLE 15 SP5 there would be a fix available, as mentioned in the comments of [bsc#1206132](https://bugzilla.suse.com/show_bug.cgi?id=1206132#c16). Now I see that this is not the case and SLE 15 SP5 also fails in the same way.
I was wandering whether we should communicate, where has this fix been released/pushed to? There is no IBS metion, however the bug has been resolved as resolved - fixed, and the issue still occurs on SLE 15 SP5 builds.

Either way, I would say, for the sake of urgency, to review this PR as is, with the purpose of not blocking MU updates any longer (and no having to force a softfail label daily). In the meantime we could hold the discussion in the bug comments,  because if the fix is not released this workaround will not be temporary.

lib/YaST/workarounds.pm
```diff
@@ -48,4 +49,26 @@ sub apply_workaround_bsc1204176 {
...
+Changes the iscsid service file to require and start after the iscsid socket.
+Then reloads systemd, in order to scan for the changed unit.
+
+=cut
+
+sub apply_workaround_bsc1207157 {
+    record_soft_failure('bsc#1207157 - openQA test fails in iscsi_client - launched then no response');
+    my $service_unit = '/usr/lib/systemd/system/iscsid.service';
+    # append the two lines at the end of the [Unit] section of the service file, as specified in bsc#1206132
+    my $cmd = q(awk -i inplace 'BEGIN {c=1}; /^$/ {if (c){print("Requires=iscsid.socket\nAfter=iscsid.socket")} c=0}; {print}' );
```
jknphy, line 67: please, use [Config::Tiny](https://metacpan.org/pod/Config::Tiny) to modify the INI file, it will be more maintainable as we don't know when this will be fixed, this bug might stay for us for a while.

lib/YaST/workarounds.pm
```diff
@@ -48,4 +49,26 @@ sub apply_workaround_bsc1204176 {
...
+
+Records a soft failure with a reference to bsc#1207157
+
+Changes the iscsid service file to require and start after the iscsid socket.
+Then reloads systemd, in order to scan for the changed unit.
+
+=cut
+
+sub apply_workaround_bsc1207157 {
+    record_soft_failure('bsc#1207157 - openQA test fails in iscsi_client - launched then no response');
```
jknphy, line 64: we should mention the ret code here, is the no response on launching related somehow? Perhaps is wrong in the bug as well because initially we don't install the dep? Let's fix bug title and sof-failure description if it is that the case.

lib/YaST/workarounds.pm
```diff
@@ -8,10 +8,11 @@ use strict;
 use warnings;
 use Exporter 'import';
 use testapi;
+use utils;
 use version_utils;
 
 
-our @EXPORT = qw(apply_workaround_bsc1204176);
+our @EXPORT = qw(apply_workaround_bsc1204176 apply_workaround_bsc1207157);
```
jknphy, line 15: nitpick: just `qw(` at the end of this line, so the values of the list are aligned below.

jknphy (commented):
We should have another PR to apply to SP5.

PR #16301, opened 2023-01-26
SLEM 5.4: remove networking section in autoyast profile

jlausuch:
This section is not needed as yast should install NetworkManager by default.

Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1207221
VR: https://openqa.suse.de/tests/10380626


mloviska:
Does it mean we need to remove the section for sle-micro 5.3 as well?

jlausuch:
> Does it mean we need to remove the section for sle-micro 5.3 as well?

Good question. I'll try and see. 
In any case, `<managed config:type="boolean">true</managed>` is deprecated and we should use either nothing or `<backend>network_manager</backend>`

mloviska:
> > Does it mean we need to remove the section for sle-micro 5.3 as well?
> 
> Good question. I'll try and see. In any case, `<managed config:type="boolean">true</managed>` is deprecated and we should use either nothing or `<backend>network_manager</backend>`

Alright, so it can be changed, however it works for 5.3 as of now

jlausuch:
> > > Does it mean we need to remove the section for sle-micro 5.3 as well?
> > 
> > 
> > Good question. I'll try and see. In any case, `<managed config:type="boolean">true</managed>` is deprecated and we should use either nothing or `<backend>network_manager</backend>`
> 
> Alright, so it can be changed, however it works for 5.3 as of now

We dont have that section defined in 5.1 and 5.2, but it's there for 5.3.

mloviska:
> We dont have that section defined in 5.1 and 5.2, but it's there for 5.3.

Yes, those are using wicked.

jlausuch:
> > We dont have that section defined in 5.1 and 5.2, but it's there for 5.3.
> 
> Yes, those are using wicked.

ok, so for 5.3 we really need to specify it https://openqa.suse.de/tests/10380649#step/cockpit_service/20

PR #16313, opened 2023-02-02
For rmt containerd testing

tinawang123:
For rmt containerd testing

- Related ticket: https://progress.opensuse.org/issues/111734
- Verification run: https://openqa.suse.de/tests/10419801


tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    # pull in the testsuite
+    assert_script_run("wget --quiet --no-check-certificate " . $rmt_helm);
+    assert_script_run("unzip rmt-helm-main.zip");
+    my $myvalue = get_var('MY_VALUE', 'https://gitlab.suse.de/QA-APAC-I/testing/-/raw/master/data/rmtcontainer/myvalue.yaml');
+    assert_script_run("wget --no-check-certificate " . $myvalue);
+    assert_script_run('cat myvalue.yaml');
+    assert_script_run("helm install rmt ./rmt-helm-main -f myvalue.yaml");
+    assert_script_run("helm list");
+    my @out = split(' ', script_output("kubectl get pods | grep rmt-app"));
+    while (1) {
```
mloviska, line 45: I find this approach of infinite loop a bit tricky. The pod is being initialized at the time, the test is already trying to gather logs from the pod.
Maybe it would be better to be more granular and split these actions to:

1) wait while the pod successfully starts
2) query the logs whether the repos inside the container have been enabled.

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    my ($self) = @_;
+
+    select_serial_terminal;
+    systemctl 'stop firewalld';
+    install_k3s();
+    install_kubectl();
+    install_helm();
+
+    my $rmt_helm = get_var('RMTTEST_REPO', 'https://gitlab.suse.de/brunoleon/rmt-helm/-/archive/main/rmt-helm-main.zip');
+    zypper_call('in git-core', exitcode => [0, 65, 107]);
```
mloviska, line 35: What is `65` return code, please? I have not seen in the docs.
BTW, is `git-core` mandatory for the test?

tests/containers/containerd_rmt.pm
tinawang123: I want to wait time exactly. 
If only wait pod started successfully. then, we still need wait until rmt-app mirrored the expected repos.
We need get two wait times. 
Use this loop, we can get exact time.

tinawang123: The wait time is not constant, if we need mirror more repos at 'myvalue.yaml', we need wait more time. So I use loop to wait the expected information in the log file.

tests/containers/containerd_rmt.pm
tinawang123: Need install git-core first, otherwise, cannot get rmt_helm packages from gitlab.

tests/containers/containerd_rmt.pm
jlausuch: but this loop is dangerous, there is no timeout.

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    install_helm();
+
+    my $rmt_helm = get_var('RMTTEST_REPO', 'https://gitlab.suse.de/brunoleon/rmt-helm/-/archive/main/rmt-helm-main.zip');
+    zypper_call('in git-core');
+    # pull in the testsuite
+    assert_script_run("wget --quiet --no-check-certificate " . $rmt_helm);
+    assert_script_run("unzip rmt-helm-main.zip");
+    my $myvalue = get_var('MY_VALUE', 'https://gitlab.suse.de/QA-APAC-I/testing/-/raw/master/data/rmtcontainer/myvalue.yaml');
+    assert_script_run("wget --no-check-certificate " . $myvalue);
+    assert_script_run('cat myvalue.yaml');
```
jlausuch, line 41: suggestion: would be nice to record this output in the UI:
`record_info('chart', script_output('cat myvalue.yaml'));`

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    my ($self) = @_;
+
+    select_serial_terminal;
+    systemctl 'stop firewalld';
+    install_k3s();
+    install_kubectl();
+    install_helm();
+
+    my $rmt_helm = get_var('RMTTEST_REPO', 'https://gitlab.suse.de/brunoleon/rmt-helm/-/archive/main/rmt-helm-main.zip');
+    zypper_call('in git-core');
```
jlausuch, line 35: Same question as Martin, why is git-core needed? You are pulling the files with `wget`

tests/containers/containerd_rmt.pm
tinawang123: Yes, I will set timeout .

tests/containers/containerd_rmt.pm
tinawang123: You're right. I need refresh my memory. 
Maybe I debug code use this line before.  Delete it. Thanks a lot.

tests/containers/containerd_rmt.pm
tinawang123: You're right. I need refresh my memory. 
Maybe I debug code use this line before.  Delete it. Thanks a lot.

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+sub run {
+    my ($self) = @_;
+
+    select_serial_terminal;
+    systemctl 'stop firewalld';
+    install_k3s();
+    install_kubectl();
+    install_helm();
+
+    my $rmt_helm = get_var('RMTTEST_REPO', 'https://gitlab.suse.de/brunoleon/rmt-helm/-/archive/main/rmt-helm-main.zip');
```
ilausuch, line 34: I think this should be in a non personal repository

brunoleon:
The repo for rmt helm chart (and supposedly all SUSE future helm charts) is at https://github.com/SUSE/helm-charts

https://gitlab.suse.de/brunoleon/rmt-helm/ was used for dev puproses only.

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    # pull in the testsuite
+    assert_script_run("wget --quiet --no-check-certificate " . $rmt_helm);
+    assert_script_run("unzip rmt-helm-main.zip");
+    my $myvalue = get_var('MY_VALUE', 'https://gitlab.suse.de/QA-APAC-I/testing/-/raw/master/data/rmtcontainer/myvalue.yaml');
+    assert_script_run("wget --no-check-certificate " . $myvalue);
+    record_info('chart', script_output('cat myvalue.yaml'));
```
ilausuch, line 40: I am worried about the security of this. because in myvalue there is a password. In QE-C we managed the password in an other way. Ask us if you are interested

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    assert_script_run("helm install rmt ./rmt-helm-main -f myvalue.yaml");
```
ilausuch, line 41: I think is better to use the flag -wait

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    assert_script_run("helm list");
+    my @out = split(' ', script_output("kubectl get pods | grep rmt-app"));
```
ilausuch, line 43: Also, I think each helm has to be executed in different namespaces, to prevent leftovers from previous tests, or concurrent tests.

tests/containers/containerd_rmt.pm
ilausuch: Also the name of the helm deployment should be randomized to prevent concurrent executions or previous leftovers (e.g. in case helm delete didn't work) or the execution was interrupted

tests/containers/containerd_rmt.pm
```diff
@@ -0,0 +1,63 @@
...
+    my $counter = 0;
+    while ($counter++ < 50) {
+        sleep 20;
+        my $logs = script_output("kubectl logs $out[0]", proceed_on_failure => 1);
+        last if ($logs =~ /All repositories have already been enabled/);
```
ilausuch, line 48: Please use the generic function "validate_pod_log" function https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/ae15a2a66941409008d8e1e8afcb111c73bb065c/lib/containers/k8s.pm#L162

tests/containers/containerd_rmt.pm
ilausuch: Forget my comment, you are using local k3s installation. This should be mandatory on a k8s in a cloud provider

tests/containers/containerd_rmt.pm
ilausuch: The same than before, In a local k3s is not mandatory

ilausuch (changes requested):

tests/containers/containerd_rmt.pm
tinawang123: Do you have the example, how to manage the password? Thanks.

tests/containers/containerd_rmt.pm
tinawang123: If use "validate_pod_log" function, I need add 'proceed_on_failure => 1' to this function, or it will be failed: https://openqa.suse.de/tests/10448609#step/containerd_rmt/120

tinawang123: Even add 'proceed_on_failure => 1', this function still cannot work: https://openqa.suse.de/tests/10448655#step/containerd_rmt/119

tests/containers/containerd_rmt.pm
ilausuch: We have this function 
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/22d08e8f7130030e80d073b31c5b7c7f7669e1d1/lib/publiccloud/utils.pm#L181 where we are getting the credentials from a server PUBLIC_CLOUD_CREDENTIALS_URL
e.g.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/22d08e8f7130030e80d073b31c5b7c7f7669e1d1/lib/publiccloud/aws_client.pm#L36

tests/containers/containerd_rmt.pm
ilausuch: ah, ok

tests/containers/containerd_rmt.pm
ilausuch: Please contact me on the slack

jlausuch:
Any more updates here?

tests/containers/containerd_rmt.pm
tinawang123: Removed myvalue file checking to not show password at openQA

tinawang123:
> Any more updates here?

I removed password files check at openQA. Any more suggestions?

brunoleon:
@tinawang123 Could you target the proper repository as requested above ?

The code is not in my personal namespace anymore

tinawang123:
> @tinawang123 Could you target the proper repository as requested above ?
> 
> The code is not in my personal namespace anymore

Updated, Use this one :https://github.com/SUSE/helm-charts    Thanks

ilausuch (approved):

tinawang123:
@jlausuch   Any more suggestions?

tests/containers/containerd_rmt.pm
ngetahun: The chart is at https://github.com/SUSE/helm-charts/tree/main/rmt-helm

jlausuch (approved):
LGTM

PR #16323, opened 2023-02-06
Rewrite and simplify tomcat test module

Zaoliang:
we have performance in past, escpially with some funciton tests for
Servelet, JSP, Websocket. With firefox and loging into webui tomcat manager, 
it failed very often. The goal is to test basic functions without firefox webui
See https://progress.opensuse.org/issues/123388

### DIFF
--- a/lib/Tomcat/Utils.pm
+++ b/lib/Tomcat/Utils.pm
@@ -15,6 +15,7 @@ use testapi;
 use utils;
 use version_utils 'is_sle';
 use registration;
+use serial_terminal;
 
 # allow a 60 second timeout for asserting needles
 use constant TIMEOUT => 90;
@@ -49,8 +50,10 @@ EOF
     select_console('root-console');
 
     record_info('Initial Setup');
+    # we need to disable packagekit because it can block zypper sometimes later
+    quit_packagekit if is_sle;
 
-    zypper_call('in tomcat tomcat-webapps tomcat-admin-webapps');
+    zypper_call('in tomcat tomcat-webapps tomcat-admin-webapps', timeout => 300);
     assert_script_run('rpm -q tomcat');
 
     # start the tomcat daemon and check that it is running
@@ -68,20 +71,28 @@ EOF
 }
 
 
-# Access the tomcat web application manager
+# Check Servlet, JSP and Websocket, the example files can be accessed, login with admin via tomcat manager
 sub tomcat_manager_test() {
-    my ($self) = shift;
+    select_serial_terminal;
 
-    $self->firefox_open_url('localhost:8080/manager');
-    assert_screen('tomcat-manager-authentication', 180);
-    type_string('admin');
-    send_key('tab');
-    type_string('admin');
-    assert_and_click('tomcat-OK-autentication');
-    wait_still_screen(2);
-    assert_and_click('tomcat-click-save-login');
-    assert_screen('tomcat-web-application-manager', TIMEOUT);
+    # sometimes we have sporadic issue with connection to localhost for unknown reason
+    # just a short check with ping
+    assert_script_run('ping -c 4 localhost');
+
+    record_info('curl examples of Servelt, JSP and Websocket');
+    # curl Servlet examples and login with authentification
+    assert_script_run('curl --connect-timeout 20 --user admin:admin --output servelets 127.0.0.1:8080/examples/servlets', 90);
+    # curl examples of JSP and Websockets
+    assert_script_run('curl --connect-timeout 20 --output jsp localhost:8080/examples/jsp --output websocket 127.0.0.1:8080/examples/websocket', 90);
 }
 
+# Switch to desktop
+sub switch_to_desktop() {
+
+    # switch to desktop
+    if (!check_var('DESKTOP', 'textmode')) {
+        select_console('x11', await_console => 0);
+    }
+}
 
 1;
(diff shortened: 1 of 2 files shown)


Zaoliang:
VR example: https://openqa.opensuse.org/tests/3101127

Zaoliang:
VR:
https://openqa.opensuse.org/tests/3105493#next_previous (tumbleweed)
http://openqa.suse.de/tests/10451156#next_previous (sles 15 sp5)

lib/Tomcat/Utils.pm
```diff
@@ -49,8 +50,10 @@ EOF
     select_console('root-console');
 
     record_info('Initial Setup');
+    # we need to disable packagekit because it can block zypper sometimes
+    quit_packagekit;
```
DimStar77, line 54: That should not be used?

At least on TW (where I checked) we run on a generated image that disables/masks packagekit as part of system_prepare (https://openqa.opensuse.org/tests/3107267#step/system_prepare/20) - so after booting that image, the service Is masked and should not be running

lib/Tomcat/Utils.pm
```diff
@@ -68,20 +71,28 @@ EOF
...
-    send_key('tab');
-    type_string('admin');
-    assert_and_click('tomcat-OK-autentication');
-    wait_still_screen(2);
-    assert_and_click('tomcat-click-save-login');
-    assert_screen('tomcat-web-application-manager', TIMEOUT);
+    select_serial_terminal;
+
+    # sometimes we have network issue (cannot connect to localhost)
+    systemctl('restart network.service');
```
DimStar77, line 79: instead of hitting this with a big hammer, I'd rather want to know when this does not work (and why).

Maybe: first check if there is an issue to connect to localhost (should hardly ever be an issue to ping ::1 or 127.0.0.1); in case of failure, record soft_failure, restart network, retest and fail on error

lib/Tomcat/Utils.pm
```diff
@@ -68,20 +71,28 @@ EOF
...
+    assert_script_run('ping -c 4 localhost');
+
+    record_info('curl examples of Servelt, JSP and Websocket');
+    # curl Servlet examples and login with authentification
+    assert_script_run('curl --connect-timeout 20 --user admin:admin --output servelets 127.0.0.1:8080/examples/servlets', 90);
```
DimStar77, line 84: after pulling the network from under the tomcat server (above, network restart) I'm actually surprised the tomcat works reliably here (except if the service managed to have restarted thanks to systemd chaining too)

lib/Tomcat/Utils.pm
Zaoliang: I hit this on sles :)

lib/Tomcat/Utils.pm
Zaoliang: sporadic issue, really hard to determine, I am open for suggestion though.

lib/Tomcat/Utils.pm
Zaoliang: yes, my goal is avoid sporadic issue which bothers us in future ;)

lib/Tomcat/Utils.pm
```diff
@@ -68,20 +71,28 @@ EOF
...
-    type_string('admin');
-    send_key('tab');
-    type_string('admin');
-    assert_and_click('tomcat-OK-autentication');
-    wait_still_screen(2);
-    assert_and_click('tomcat-click-save-login');
-    assert_screen('tomcat-web-application-manager', TIMEOUT);
+    select_serial_terminal;
+
+    # sometimes we have sporadi issue with connection to localhost for unknown reason
```
rfan1, line 78: s/sporadi/sporadic/

lib/Tomcat/Utils.pm
```diff
@@ -68,20 +71,28 @@ EOF
...
+    # just a short check with ping
+    assert_script_run('ping -c 4 localhost');
```
rfan1, line 80: Hello Zaoliang,

May you please have a try to restart the network service if connection to localhost is lost? and then quit the test if the network can't still work.

rfan1:
@Zaoliang 
I have one more comment here, you have switched all the test steps to serial terminal, and seems we don't need to run tomcat in x11 mode, correct?

In that case, can we remove the function `switch_to_desktop` and move the test module to 'consoletest'?

lib/Tomcat/Utils.pm
Zaoliang: thanks

lib/Tomcat/Utils.pm
Zaoliang: this is dimstar already mentioned before, restarting network.service is hammer :) so I can accept this without to start it, but at least we know where is problem by checking ping localhost. So we can still fix easy and fast the kind of sporadic issue in future. At moment I cannot determine why connect to localhost doesn't work sometimes :)
I assume at moment it is working... we can leave this now.

Zaoliang:
@rfan1 well, I don't move it to consoletest at moment. We don't know if someone wants still keep it in x11 test for some extension in future. 
At moment we just make tomcat as stable as possible. There is also scenarios for MM-testing which was brought to us last year.

The problem is without 'switch to desktop', it test can fail from serial console somehow, I don't know exactly why, and shutdown triggered directly from desktop anyway which is good :)

Zaoliang:
http://openqa.suse.de/tests/10463654#next_previous all passed after latest update in MP.

rfan1:
Thanks Zaoling, LGTM

PR #16325, opened 2023-02-06
Fix typo

asmorodskyi:


m-dati (approved):
The `while` could create an infinite loop 
```
my $a=0;
my $b=1; 
$a=1 while $b == 1;
```
So, with IF it is much clean :)

jlausuch (approved):

pdostal (approved):
I see no problem here, but please do verification runs with `PUBLIC_CLOUD_REDOWNLOAD_MU=1` before merging this.

PR #16327, opened 2023-02-07
Add install_SLES to fix no license shown on s390x

lemon-suse:
We need add install_SLES to fix no license shown on s390x. 
failed job: https://openqa.nue.suse.com/tests/10434468#step/accept_license/1

- Related ticket: N/A
- Needles: N/A
- Verification run:  https://openqa.suse.de/tests/10441730#details


jknphy:
it is a bug, see https://jira.suse.com/browse/PED-2973

PR #16234, opened 2023-01-13
DISA STIG test fix for BZ1194676

vtrubovics:
STIG test fix: error code fix and additional checks for number of passed/failed rules

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1194676
- Verification run: https://openqa.suse.de/tests/10312910

PR #16332, opened 2023-02-07
Oscap stig rework

vtrubovics:
Reworked open SCAP STIG profile test with ability to add other profiles tests.

Issue link: https://progress.opensuse.org/issues/124056
- Verification run: 
aarch64 https://openqa.suse.de/tests/10443256
s390x https://openqa.suse.de/tests/10443257
x86_64 https://openqa.suse.de/tests/10443258


paolostivanin:
does this replace https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16234 ?

vtrubovics:
> does this replace #16234 ?

Yes, it replace it and extends.

lib/oscap_tests.pm
```diff
@@ -0,0 +1,348 @@
...
+    my $data = $_[1];
+    my $pattern = $_[2];
+    my @rules;
+    my $count = 0;
+
+    my @lines = split /\n|\r/, $data;
+    for my $i (0 .. $#lines) {
+        if ($lines[$i] =~ /$pattern/) {
+            $count++;
+            push(@rules, $lines[$i - 4]);
```
ilmanzo, line 119: whant happens if $i < 4 ?

lib/oscap_tests.pm
```diff
@@ -0,0 +1,348 @@
...
+    my $l_rules = $_[3];
+    my @rules;
+    my $count = 0;
+
+    my @lines = split /\n|\r/, $data;
+    my @a_rules = @$l_rules;
+
+    for my $i (0 .. $#lines) {
+        for my $j (0 .. $#a_rules) {
+            if ($lines[$i] =~ /$a_rules[$j]/ and $lines[$i + 4] =~ /$pattern/) {
```
ilmanzo, line 143: - quadratic complexity ?
- second part of the condition is constant in the inner loop 
- consider using Perl grep function

vtrubovics: It is simplest that I can do when looking for the rule name and checking it passed or failed:
Rule
	xccdf_org.ssgproject.content_rule_is_fips_mode_enabled  >> $lines[$i]
Ident
	CCE-85763-1
Result
	fail                                                                                       >>$lines[$i + 4]

vtrubovics: Did not see simpler implementation. I tried had to implement grep with pattern - looks OpenQA has issues with it...

lib/oscap_tests.pm
vtrubovics: Perl will fail, but he I rely on format of stdout:
Title
	Use Only FIPS 140-2 Validated MACs
Rule
	xccdf_org.ssgproject.content_rule_sshd_use_approved_macs_ordered_stig
Ident
	CCE-83280-8
Result
	pass
	
If [$i] exists -  [$i - 4 ] exists 100%

lib/oscap_tests.pm
ilmanzo: just to clarify: I'm talking about [Perl's grep function](https://perldoc.perl.org/functions/grep) , not the unix grep(1) command :)

lib/oscap_tests.pm
ilmanzo: the code relies on a specific output format, which can change and make it prone to unexpected behavior. Please consider adding at least a check on the array length and consider the usage of more idiomatic Perl constructs like foreach, grep. Perhaps you could also build a key,value hash with the data you expect instead of relying on fixed index offsets.

lib/oscap_tests.pm
vtrubovics: grep function you mentioned does not fit the purpose - "actually modifies the element in the original list. This is usually something to be avoided when writing clear code."

lib/oscap_tests.pm
vtrubovics: > the code relies on a specific output format, which can change and make it prone to unexpected behavior.
We always rely on output format and cannot predict when and how it will change. There are no universal analysis method.  I will add additional checks to not get out of array.

>consider the usage of more idiomatic Perl constructs like foreach, grep.
Here **for my $i (0 .. $#lines)** equals **foreach** but with ability to get needed elements of array and easy rely on output format. Also in possible format change case can do easy fix. Grep function you referring does not fit the purpose.

>Perhaps you could also build a key,value hash with the data you expect instead of relying on fixed index offsets.
Do not see the point of building full key,value hash - I do not need all this data (Simpler is better and faster), only some of it.

paolostivanin:
FYI, since a couple of days the test started failing on x86_64 (https://openqa.suse.de/group_overview/431). It seems that, all of a sudden, 16GB of RAM are no longer enough.
I'm now trying with 24GB.

paolostivanin:
@vtrubovics with latest change test is now failing with: https://openqa.suse.de/tests/10527237#step/oscap_xccdf_eval#1/9

vtrubovics:
> @vtrubovics with latest change test is now failing with: https://openqa.suse.de/tests/10527237#step/oscap_xccdf_eval#1/9

There were bug in my last fix. I did local testing and fixed incorrect comparison. Rerunning now in https://openqa.suse.de/tests/10527562
By the way - perl is tolerant in trying to read array member outside index.

tjyrinki (commented):
LGTM, some clean up will follow later on

PR #16342, opened 2023-02-09
Fix grafana login and other navigation issues

emiura:
Fixes poo#124113

VR at https://openqa.suse.de/tests/10457536


lpalovsky (approved):
LGTM

PR #16347, opened 2023-02-10
Unschedule yast2_ntpclient test from security testsuite

ilmanzo:
Rationale:
    - this test module only verifies the working of yast2 NTP, which has little to share with product security testing.
    - the actual time synchronization feature and communication protocol tests are already implemented and checked by module ntp_client
    - yast2_ntpclient module is properly tested in YaST testsuites, like [yast2_ncurses](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/schedule/yast/yast2_ncurses/yast2_ncurses_textmode.yaml) and [yast2_gnome](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/schedule/yast/yast2_ncurses/yast2_ncurses_gnome.yaml)
    - reduce complexity and resource usage

- Related ticket: https://progress.opensuse.org/issues/123442
- Verification runs: 
  - https://openqa.suse.de/tests/10458627
  - https://openqa.suse.de/tests/10458628
  - https://openqa.suse.de/tests/10462635

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -2304,7 +2304,6 @@ sub load_security_tests_crypt_tool {
     loadtest "console/clamav";
     loadtest "console/openvswitch_ssl";
     loadtest "console/ntp_client";
-    loadtest "console/yast2_ntpclient";
     loadtest "console/cups";
     loadtest "console/syslog";
 }


punkioudi:
The difference with the yast team is that here it is tested under fips, as also it was part of this ticket: https://progress.opensuse.org/issues/119764
Small idea would be to schedule yast2_ncurses or yast2_gnome under fips

ilmanzo:
The FIPS 140-2 standard ensures that cryptographic tools implement their algorithms properly. see [full standard specs](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.140-2.pdf) for details. Since we already test ntp communication with ntp_client test and yast_ntpclient module does not perform any specific cryptographic operation (it internally executes `/usr/sbin/chronyd`), see [source](https://github.com/yast/yast-ntp-client/blob/master/src/modules/NtpClient.rb), I think can be removed from security tests

punkioudi:
Could you also remove it from here? https://confluence.suse.com/pages/viewpage.action?pageId=937623759

punkioudi (approved):
lgtm!

PR #16349, opened 2023-02-10
Add Distribution provider for ALP

lemon-suse:
We need add Distribution provider for ALP.

- Related ticket:  https://progress.opensuse.org/issues/123277
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/3109492#step/podman_yast_firewall/22

### DIFF
--- a/lib/Distribution/Alp.pm
+++ b/lib/Distribution/Alp.pm
@@ -1,0 +1,16 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class represents ALP distribution and provides access to
+# its features.
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Distribution::Alp;
+use strict;
+use warnings FATAL => 'all';
+use parent 'susedistribution';
+
+1;
--- a/lib/DistributionProvider.pm
+++ b/lib/DistributionProvider.pm
@@ -20,6 +20,7 @@ use Distribution::Sle::12;
 use Distribution::Opensuse::Leap::42;
 use Distribution::Opensuse::Leap::15;
 use Distribution::Opensuse::Tumbleweed;
+use Distribution::Alp;
 
 =head2 provide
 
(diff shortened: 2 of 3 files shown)


lib/DistributionProvider.pm
```diff
@@ -20,6 +20,7 @@ use Distribution::Sle::12;
 use Distribution::Opensuse::Leap::42;
 use Distribution::Opensuse::Leap::15;
 use Distribution::Opensuse::Tumbleweed;
+se Distribution::Alp::Alp;
```
tinawang123, line 23: Maybe miss 'u'

lemon-suse: Yes, copy code miss a 'u'. Thanks.

lib/DistributionProvider.pm
```diff
@@ -20,6 +20,7 @@ use Distribution::Sle::12;
 use Distribution::Opensuse::Leap::42;
 use Distribution::Opensuse::Leap::15;
 use Distribution::Opensuse::Tumbleweed;
+use Distribution::Alp::Alp;
```
jknphy, line 23: just one Alp, we don't need a folder if there is only one file and we don't know exactly the flavors of ALP at this point.

lib/Distribution/Alp.pm
```diff
@@ -0,0 +1,16 @@
...
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Distribution::Alp;
+use strict;
+use warnings FATAL => 'all';
+use parent 'susedistribution';
+use YaST::Firewall::FirewallController;
+
+sub get_firewall {
```
jknphy, line 12: we don't need this method, neither a controller.

products/alp/main.pm
```diff
@@ -16,6 +17,7 @@ init_main();
 my $distri = testapi::get_required_var('CASEDIR') . '/lib/susedistribution.pm';
 require $distri;
 testapi::set_distribution(susedistribution->new());
+testapi::set_distribution(DistributionProvider->provide());
```
jknphy, line 20: we just need to set it once, the one 'provided' is derived class of susedistribution.

jknphy:
in title instead of 'control' I would say is 'provider'
Please specify ticket related and if there is not needles you don't need to write it.
Verification is fine to do it with your draft for firewall but put out of the commit before merging.

products/alp/main.pm
lemon-suse: updated

lib/Distribution/Alp.pm
lemon-suse: Updated

lib/Distribution/Alp.pm
```diff
@@ -0,0 +1,11 @@
+# Summary: The class represents ALP distribution and provides access to
```
jknphy, line 1: and the copyright?

lemon-suse: Updated

PR #16350, opened 2023-02-10
Deactivate PackageHub for SLE Mciro 5.4

waynechen55:
* **PackageHub** is not needed at the moment, so it would be better to deactivate it because associated PackageHub repository can not be refreshed successfully with local registration.

* **Verification runs**:
  * [slem on slem](https://openqa.suse.de/tests/10481794)
  * [sles on slem](https://openqa.suse.de/tests/10481200)

### DIFF
--- a/tests/virt_autotest/prepare_transactional_server.pm
+++ b/tests/virt_autotest/prepare_transactional_server.pm
@@ -34,7 +34,7 @@ sub prepare_in_trup_shell {
     my $self = shift;
 
     transactional::enter_trup_shell(global_options => '--drop-if-no-change');
-    $self->prepare_extensions;
+    #$self->prepare_extensions;
     $self->prepare_packages;
     $self->prepare_bootloader;
     transactional::exit_trup_shell_and_reboot();


waynechen55:
@alice-suse @Julie-CAO @guoxuguang @nanzhg @tbaev Pleas have a look.

tests/virt_autotest/prepare_transactional_server.pm
```diff
@@ -50,7 +50,7 @@ sub prepare_extensions {
     my $self = shift;
 
     #Subscribing packagehub that enables access to many useful software tools
-    virt_autotest::utils::subscribe_extensions_and_modules(reg_exts => 'PackageHub');
+    virt_autotest::utils::subscribe_extensions_and_modules(reg_exts => 'PackageHub', activate => 0);
```
alice-suse, line 53: Is it correct that the calling to virt_autotest::utils::subscribe_extensions_and_modules is not needed at all at the moment?  How about commenting this line?

waynechen55: How about just keep it and use activate => 0 ?  So I do not need to do more verification runs. I think it does no harm.

alice-suse: No big deal of course. I thought commenting off would be more efficient -- no jumping in and out of the function. But up to you to decide.

waynechen55: Done

alice-suse:
LGTM

PR #16353, opened 2023-02-11
SLE Micro: handle activation of SCC modules

jlausuch:
VR: https://openqa.suse.de/tests/10477990
The test failure is due to https://bugzilla.suse.com/show_bug.cgi?id=1208148

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -133,7 +133,7 @@ sub accept_addons_license {
 
     # In SLE 15 some modules do not have license or have the same
     # license (see bsc#1089163) and so are not be shown twice
-    push @addons_with_license, @SLE15_ADDONS_WITHOUT_LICENSE unless is_sle('15+');
+    push @addons_with_license, @SLE15_ADDONS_WITHOUT_LICENSE unless (is_sle('15+') || is_sle_micro);
     # HA and WE have licenses when calling yast2 scc
     push @addons_with_license, @SLE15_ADDONS_WITH_LICENSE_NOINSTALL if (is_sle('15+') and get_var('IN_PATCH_SLE'));
     # HA does not show EULA when doing migration to 12-SP5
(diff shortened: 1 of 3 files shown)


ricardobranco777 (approved):
LGTM

tests/installation/scc_registration.pm
```diff
@@ -24,7 +24,7 @@ sub run {
         assert_registration_screen_present();
         fill_in_registration_data();
         wait_still_screen();
-        if (is_sle('>=15-SP4') && check_screen('import-untrusted-gpg-key', 5)) {
+        if ((is_sle('>=15-SP4') || is_sle_micro) && check_screen('import-untrusted-gpg-key', 5)) {
```
asmorodskyi, line 27: for SLE we handling keys only for latest unreleased version , which means if such issue will suddenly pop-up somewhere like 15 SP2 this will be treated as a bug and fail the test . Are you sure that you want blindly accept unknown gpg keys on SLEM5.1/5.2/5.3 ?

jlausuch: Good point. I'm planning to run this in "latest", so I'll add the version check here. Thanks!

lib/registration.pm
```diff
@@ -525,6 +525,9 @@ sub process_scc_register_addons {
         # start addons/modules registration, it needs longer time if select multiple or all addons/modules
         my $counter = ADDONS_COUNT;
         my @needles = qw(import-untrusted-gpg-key nvidia-validation-failed yast_scc-pkgtoinstall yast-scc-emptypkg inst-addon contacting-registration-server refreshing-repository system-probing);
+
+        # In SLE Micro, we don't need to continue with registering any additional module.
+        return if is_sle_micro;
```
mloviska, line 530: We might still want to register livepatching or not?

asmorodskyi (approved):
LGTM

lib/registration.pm
jlausuch: I had to add this `return` because the installation was not asking for it and the test got stuck waiting for license needle...

PR #16356, opened 2023-02-13
Hide CI passed logs for yaml files checking

rfan1:
- Related ticket: https://progress.opensuse.org/issues/122923
- Verification run: n/a

### DIFF
--- a/tools/check_yaml
+++ b/tools/check_yaml
@@ -1,0 +1,14 @@
+#!/bin/bash -e
+
+success=1
+YAMLS=$(git ls-files schedule/ test_data/ | grep '.ya\?ml$' | xargs echo)
+
+if test -n "$YAMLS"; then
+    export PERL5LIB=${PERL5LIB_} ; echo "$YAMLS" | xargs tools/test_yaml_valid | grep -Ev "has valid yaml syntax|has valid schema";
+    which yamllint >/dev/null 2>&1 || (echo "Command 'yamllint' not found, can not execute YAML syntax checks" && exit 1) || success=0;
+    echo "$YAMLS" | xargs yamllint -c .yamllint || exit 1 || success=0;
+else
+    echo "No yamls modified.";
+fi
+[ $success = 1 ] && echo "check_yaml SUCCESS" && exit 0
+exit 1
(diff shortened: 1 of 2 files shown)


RoyCai7:
It looks like it works well. LGTM

tools/check_yaml
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+success=1
+YAMLS=$(git ls-files schedule/ test_data/ | grep '.ya\?ml$' | xargs echo)
+
+if test -n "$YAMLS"; then \
```
grisu48, line 6:
```suggestion
if test -n "$YAMLS"; then
```
If this is in an external bash file, then all the `\` at the end can be omitted.

Same applies to almost every line in this script.

tools/check_yaml
```diff
@@ -0,0 +1,17 @@
...
+        export PERL5LIB=${PERL5LIB_} ; echo "$YAMLS" | xargs tools/test_yaml_valid | grep -Ev "has valid yaml syntax|has valid schema";\
+        else \
+        echo "No yamls modified.";\
+fi
+if test -n "$YAMLS"; then \
```
grisu48, line 11: Isn't this check done twice and can be merged? :thinking: 

Line 6 and 11 look exactly the same to me and I don't see any changes to `$YAMLS` itself.

grisu48 (commented):
Two suggestions. Let's remove the unnecessary `\` from the script!

tools/check_yaml
rfan1: You are right, let me revise my code

tools/check_yaml
```diff
@@ -0,0 +1,17 @@
...
+        which yamllint >/dev/null 2>&1 || (echo "Command 'yamllint' not found, can not execute YAML syntax checks" && exit 1) || success=0;\
+
+        echo "$YAMLS" | xargs yamllint -c .yamllint || exit 1 || success=0;\
+fi
+[ $success = 1 ] && echo "check_yaml SUCCESS" && exit 0
+exit 1
```
b10n1k, line 17: That can be just `exit $success` right?

rfan1: hmm, for shell, we may need to `exit 1` if any failure

grisu48: Richard is right. `$success = 1` results in `exit 0`, so `exit $success` won't work.

grisu48 (approved):
LGTM!

PR #16358, opened 2023-02-13
WIP Add wait_still_screen for for needle matching

Zaoliang:
the performance with firefox is low, default setting for memory is 1024 only. try to work around this issue without changing QEMURAM see https://progress.opensuse.org/issues/124095


Zaoliang:
https://openqa.suse.de/tests/10482121#next_previous looks good now

Zaoliang:
no single failure from 100 test runs on OSD.

PR #16361, opened 2023-02-13
Reset the VMs in textmode for xen VMs for test 'user_defined_snapshot'

rfan1:
https://progress.opensuse.org/issues/123999

On xen setups, reset the VMs in x11 mode fails sometimes. however it can't be reproduced stably.
Swith the reboot operation in textmode will get better performance.

- Verification run: [VRs](http://openqa.suse.de/tests/overview?build=rfan_xen&version=15-SP5&distri=sle)

### DIFF
--- a/tests/x11/user_defined_snapshot.pm
+++ b/tests/x11/user_defined_snapshot.pm
@@ -64,7 +64,13 @@ sub run {
     wait_serial("$module_name-0", 200) || die "'yast2 $module_name' didn't finish";
     $self->{in_wait_boot} = 1;
     record_info 'Snapshot created', 'booting the system into created snapshot';
-    power_action('reboot', keepconsole => 1);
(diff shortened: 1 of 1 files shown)


dzedro (approved):
Looking on the failure, the reboot is stuck at `shutdown-auth`, maybe fixing that would be better. But if "only" xen does have this problem and it works, then it's good enough.

foursixnine:
@rfan1 is the WIP label still needed? or are we ok to merge?

rfan1:
> @rfan1 is the WIP label still needed? or are we ok to merge?

Let me ask for QE-2 members to review this PR as well :)

RoyCai7:
LGTM

alice-suse:
Thanks for involving us for review. Pity that AFAIK we (VT) do not have tests using this code piece. So we are not familiar with it. But I had a look, and it is good to me.

PR #16363, opened 2023-02-13
Move PAM testing after registration

mloviska:
PAM test requires several packages to be installed, therefore we need refreshed repos to be available.

- VR: http://kepler.suse.cz/tests/20356#step/pam/17


ricardobranco777 (approved):
LGTM

jlausuch (approved):

PR #16364, opened 2023-02-13
Refactor the post_fail_hook functionality

DeepthiYV:
- Related ticket:  https://progress.opensuse.org/issues/120735
- Needles: No
- Verification run: 
 I forcefully made the firefox test fail and results looks fine : https://openqa.opensuse.org/tests/3161410#
  Other VRs:

https://openqa.suse.de/tests/10635862
https://openqa.opensuse.org/tests/3161407
https://openqa.opensuse.org/tests/3161403
https://openqa.opensuse.org/tests/3161402#
https://openqa.opensuse.org/tests/3161404#

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -24,6 +24,7 @@ use IO::Socket::INET;
 use x11utils qw(handle_login ensure_unlocked_desktop handle_additional_polkit_windows);
 use publiccloud::ssh_interactive 'select_host_console';
 use Utils::Logging qw(save_and_upload_log tar_and_upload_log export_healthcheck_basic select_log_console upload_coredumps export_logs);
+use serial_terminal 'select_serial_terminal';
 
 # Base class for all openSUSE tests
 
@@ -981,21 +982,18 @@ base method using C<$self-E<gt>SUPER::post_fail_hook;> at the end.
 
 sub post_fail_hook {
     my ($self) = @_;
-    return if is_serial_terminal();    # unless VIRTIO_CONSOLE=0 nothing below make sense
-
-    show_tasks_in_blocked_state;
 
     return if (get_var('NOLOGS'));
 
     # Upload basic health check log
-    select_log_console;
+    select_serial_terminal();
     export_healthcheck_basic;
 
     # set by x11_start_program
     if (get_var('IN_X11_START_PROGRAM')) {
         my ($program) = get_var('IN_X11_START_PROGRAM') =~ m/(\S+)/;
         set_var('IN_X11_START_PROGRAM', undef);
-        select_log_console;
+
         my $r = script_run "which $program";
         if ($r != 0) {
             record_info("no $program", "Could not find '$program' on the system", result => 'fail');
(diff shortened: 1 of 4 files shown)


pdostal:
Hello Deepthi,

please check #16083 commit by commit. The serial console sadly still has issues in `post_fail_hook` but the rest can be done.

lib/opensusebasetest.pm
```diff
@@ -1025,19 +1021,20 @@ sub post_fail_hook {
...
 
     # In case the system is stuck in shutting down or during boot up, press
     # 'esc' just in case the plymouth splash screen is shown and we can not
     # see any interesting console logs.
-    send_key 'esc';
-    save_screenshot;
-    # the space prevents the esc from eating up the next alphanumerical
-    # character typed into the console
-    send_key 'spc';
+    unless (is_serial_terminal()) {
```
grisu48, line 1031: Do we need this block? Line 989 should ensure that we are on a serial terminal, no?

rfan1:
Can you please add some VRs on non-qemu backend tests if any? e.g. I don't think you need to do all of them, but worth to have a try

ipmi
s390x
spvm
Hyper-v
Xen
vmware

lib/opensusebasetest.pm
DeepthiYV: yeah! we do not need this block. It has never been executed. I will remove it.

lib/opensusebasetest.pm
```diff
@@ -981,29 +982,25 @@ base method using C<$self-E<gt>SUPER::post_fail_hook;> at the end.
...
     my ($self) = @_;
-    return if is_serial_terminal();    # unless VIRTIO_CONSOLE=0 nothing below make sense
-
-    show_tasks_in_blocked_state;
 
     return if (get_var('NOLOGS'));
 
     # Upload basic health check log
-    select_log_console;
+    select_serial_terminal();
```
b10n1k, line 989: i really think that you have to check whether serial_terminal is already used to avoid calling this unnecessarily. Many tests are already calling ` select_serial_terminal`. Although `select_console` does some checks, I think it would be better if you prevent to run the whole stuck when serial terminal is already set. WDYT?

pdostal: You can use `select_serial_terminal` multiple times.

lib/x11test.pm
```diff
@@ -29,12 +30,13 @@ sub post_run_hook {
...
     return if (get_var('NOLOGS'));
-    select_log_console;
+    select_serial_terminal();
     export_healthcheck_basic;
-    show_tasks_in_blocked_state;
     export_logs_basic;
     # Export extra log after failure for further check gdm issue 1127317, also poo#45236 used for tracking action on Openqa
     export_logs_desktop;
+    select_log_console;
+    show_tasks_in_blocked_state;
```
foursixnine, line 39: `show_oom_info` is missing

foursixnine (approved):

lib/x11test.pm
foursixnine: @DeepthiYV aside from this, looks pretty good!

PR #16370, opened 2023-02-14
Fix product shortcuts for 15-SP4 QU Full image as well

tjyrinki:
The 15-SP4 QU2 candidate images have added SUSE Manager products to the Full images, resulting in shortcut changes. Earlier QUs will not be further tested.

- Related ticket: https://progress.opensuse.org/issues/124206
- Verification run: https://openqa.suse.de/tests/10494658


paolostivanin:
LGTM

PR #16373, opened 2023-02-14
gpg: Install libraries with --allow-downgrade

pdostal:
Please refer to [bsc#1199312](https://bugzilla.suse.com/show_bug.cgi?id=1199312) and [bsc#1199561](https://bugzilla.suse.com/show_bug.cgi?id=1199561) for more information.

- Related ticket: [poo#122077](https://progress.opensuse.org/issues/122077)
- Verification run: [pdostal-server.suse.cz/t3957](https://pdostal-server.suse.cz/t3957)

### DIFF
--- a/tests/console/gpg.pm
+++ b/tests/console/gpg.pm
@@ -177,7 +177,14 @@ sub run {
         libgcrypt20 => '1.9.0',
         'libgcrypt20-hmac' => '1.9.0'
     };
-    zypper_call("in " . join(' ', keys %$pkg_list));
+    if (is_public_cloud() && check_var('BETA', '1')) {
+        if (zypper_call("in " . join(' ', keys %$pkg_list), exitcode => [0, 4]) == '4') {
+            record_info("Aborting", "The repositories are behind Public Cloud image. See bsc#1199312");
+            return 1;
+        }
+    } else {
+        zypper_call("in " . join(' ', keys %$pkg_list));
+    }
 
     if (is_sle('>=15-sp4')) {
         package_upgrade_check($pkg_list) unless (is_public_cloud() && check_var('BETA', '1'));


tests/console/gpg.pm
```diff
@@ -172,7 +177,7 @@ sub run {
         libgcrypt20 => '1.9.0',
         'libgcrypt20-hmac' => '1.9.0'
     };
-    zypper_call("in " . join(' ', keys %$pkg_list));
+    zypper_call("in --allow-downgrade " . join(' ', keys %$pkg_list));
```
asmorodskyi, line 180: this problem is very specific to situation which we have ONLY when we testing "Latest" when image is newer than state of the repo , in any maintenance update I would treat downgrade as a bug or at least some which should be reviewed by human , so please add check for BETA=1

mloviska, line 180: This is extremely dangerous and can cover a lot of issues, mainly with updates.

pdostal: Is `if(check_var('BETA', '1'))` OK for you? Otherwise I'm worried that there is other way for us how to check that.

mloviska: Frankly, I don't think it is a good idea. This can happen when the image is built with a newer package than it is in repo. Thus, repos are not in sync with image. Leading to project issues, and they should be raised by us. Since we are talking about the latest testing, the benefits of this change are questionable.

asmorodskyi: after more thinking about this I  am suggesting the following : 
1. remove `--allow-downgrade` 
2. catch zypper error here and in case `BETA=1` make it soft_fail and skip further test

There are two points why : 
1. As @mloviska correctly mention everywhere except Public Cloud Latest it simply should not be like that 
2. In case of Public Cloud Latest we don't have other choice actually because repos are one step behind the image ( repos for Public Beta contains latest **released** beta while image built on Latest Beta ) BUT if such situation really happens there is no point to run the test anyway because anyway **we won't cover what is on image** because we will downgrade packages so from coverage POV of Latest image there is 0 value of continue this test

asmorodskyi (approved):
LGTM

asmorodskyi:
@mloviska please merge if it is ok for you

tests/console/gpg.pm
```diff
@@ -177,7 +177,15 @@ sub run {
         libgcrypt20 => '1.9.0',
         'libgcrypt20-hmac' => '1.9.0'
     };
-    zypper_call("in " . join(' ', keys %$pkg_list));
+    my $ret;
+    if (is_public_cloud() && check_var('BETA', '1')) {
+        ## We don't continue when latest PC image is behind repos
+        $ret = zypper_call("in " . join(' ', keys %$pkg_list), exitcode => [0, 4]);
+        record_info("X", $ret);
+        exit(1) if ($ret == "4");
```
mloviska, line 185: Let's use return/die/softfail as everywhere else in the test code. 

```
Don't use exit to abort a subroutine if there's any chance that someone might want to trap whatever error happened. Use die instead, which can be trapped by an eval.
```

IIRC, tests are eventually running in `eval`, but I am not that much experienced over here.

mloviska (approved):

PR #16374, opened 2023-02-14
Reorder cs_wait_for_cluster calls in sles4sap/sap_suse_cluster_connector

alvarocarvajald:
In the `sles4sap/sap_suse_cluster_connector` test module, not all calls to the `sap_suse_cluster_connector` binary results in changes in the cluster, so calls to `wait_for_idle_cluster` are not always needed. Moreover, the big two calls (stop & start) already had calls to `wait_until_resources_started` and to `wait_until_resources_stopped` with a wait time of 1200 seconds, which should also cover the wait for idle cluster function. This commit drop calls to `wait_for_idle_cluster` introduced to `sles4sap/sap_suse_cluster_connector` in commit 7b07cb6, and adds them only before the calls to `exec_conn_cmd` which would impact the cluster itself.

### DIFF
--- a/tests/sles4sap/sap_suse_cluster_connector.pm
+++ b/tests/sles4sap/sap_suse_cluster_connector.pm
@@ -26,7 +26,6 @@ sub exec_conn_cmd {
     my $cmd = $args{cmd};
     $cmd .= " --out $args{log_file}" if ($args{log_file});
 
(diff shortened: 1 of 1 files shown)


bs-github:
yes, let's try that.

emiura:
LGTM!

lpalovsky (approved):
LGTM

PR #16376, opened 2023-02-14
Bring back install_SLES on aarch64

rakoenig:
- https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16357 removed a bit too much, on aarch64 some test suites now fail because hpc product is also avialable.

- Related ticket: 
- Needles: 
- [Verification runs](https://openqa.suse.de/tests/overview?version=15-SP5&build=rakoenig%2Fos-autoinst-distri-opensuse%23fix_aarch64_problems&distri=sle)

### DIFF
--- a/schedule/yast/btrfs/btrfs+warnings.yaml
+++ b/schedule/yast/btrfs/btrfs+warnings.yaml
@@ -14,7 +14,7 @@ schedule:
   - installation/bootloader_start
   - installation/setup_libyui
   - installation/access_beta_distribution
-  - '{{install_sles}}'
+  - installation/product_selection/install_SLES
   - installation/licensing/accept_license
   - installation/registration/register_via_scc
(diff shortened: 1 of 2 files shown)


jknphy:
Please, resolve conflicts.

rakoenig:
Conflicts came after [PR#16380](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16380) was merged. PR#16380 does practically the same, but introduces conditional YAML for aarch64 in files that are only used in aarch64, for all other architectures we have specific YAML schedules that are used in the job group. 

How shall we proceed:
a) resolve the conflict by changing back to non-conditional YAML statements
b) close this PR and forget about it

WDYT?

jknphy:
that PR do not make sense because now we have more files we don't need to do any conditional, please proceed.

PR #16379, opened 2023-02-14
opensuse_welcome: Wait longer for desktop-runner on KDE

dzedro:
It can take some time for KRunner to start, until then alt-f2 has no effect
The linked bug is pretty outdated and IMO not related to KDE

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=978027
- Verification run: https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=opensuse_welcome_bsc978027

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -200,9 +200,13 @@ sub init_desktop_runner {
 
     mouse_hide(1);
     if (!check_screen('desktop-runner', $timeout)) {
-        record_info('workaround', "desktop-runner does not show up on $hotkey, retrying up to three times (see bsc#978027)");
-        send_key 'esc';    # To avoid failing needle on missing 'alt' key - poo#20608
-        send_key_until_needlematch 'desktop-runner', $hotkey, 4, 10;
+        record_info('workaround', "desktop-runner does not show up on $hotkey, retrying");
+        # To avoid failing needle on missing 'alt' key - poo#20608
+        send_key 'esc';
+        # It can take some time until KRunner is started
+        my $key_count = is_opensuse && check_var('DESKTOP', 'kde') ? 5 : 4;
+        my $key_timeout = is_opensuse && check_var('DESKTOP', 'kde') ? 20 : 10;
+        send_key_until_needlematch 'desktop-runner', $hotkey, $key_count, $key_timeout;
     }
     wait_still_screen(2);
     for (my $retries = 10; $retries > 0; $retries--) {


lib/susedistribution.pm
```diff
@@ -202,7 +202,10 @@ sub init_desktop_runner {
     if (!check_screen('desktop-runner', $timeout)) {
         record_info('workaround', "desktop-runner does not show up on $hotkey, retrying up to three times (see bsc#978027)");
```
DimStar77, line 203: That bug ref is the one that's been linked to the failing tests; as you already stated: the softfail info does not match at all - so it should probably be improved (if even still being valid)

dzedro: ok :+1:

dzedro: Not sure if I understood the improvement right, to remove it.
I guess for KDE it's clear that KRunner is not running and for other desktop it's kind of the same.
We have to retry because of performance or system has to finish start.

DimStar77: The workaround makes sense - but the recorder softfail info with the bug report does not make sense; that part should be corrected/adjusted imho

dzedro: It's just record_info.  https://openqa.opensuse.org/tests/3118260#step/opensuse_welcome/7

DimStar77: yes - clear - but it records info referring to a bug number that is not accurate (as you stated yourself: """ The linked bug is pretty outdated and IMO not related to KDE""")

dzedro: True, but the retry is generic, specific to KDE is that KRunner has to start and it needs longer than other DMs.
I would leave the bug there as it's the "only" reference.

dzedro: ok, I removed it, it's mentioned in the commit and it's old/ not related

dzedro:
@DimStar77 I increased QEMURAM to 3072 for KDE-Live tests, QEMURAM=3072 was defined in medium types (now +QEMURAM=3072), but machine settings overwrote it so it was 2048.
 I checked why was firefox failing on KDE, the tests were super slow, all RAM used. starting firefox or just pressing alt-f1 to get menu took more than 10 seconds.
Now firefox runs below 3 minutes, before above 6 minutes.
I would wait few runs and see how this tests behave and at the end this change will be not required, I could just improve record_info. More RAM could improve, ideally fix, firefox, opensuse_welcome and grub_test in upgrade jobs.

dzedro:
KDE can still fail on desktop-runner, because KRunner was not started.

PR #16381, opened 2023-02-15
Add python311 test coverage

rfan1:
https://progress.opensuse.org/issues/123990

- Verification run: 

[SLE15SP5](http://openqa.suse.de/tests/10513708)
[TW](https://openqa.opensuse.org/tests/3121575)


tests/console/python3_new_version_check.pm
```diff
@@ -32,16 +32,19 @@ sub run {
         add_suseconnect_product('sle-module-python3');
     }
 
-    my @python310_results = split "\n", script_output("zypper se python310 | awk -F \'|\' \'/python310/ {gsub(\" \", \"\"); print \$2}\'");
-    if ((scalar(@python310_results) == 0) || ($python310_results[0] ne "python310")) {
-        die("Python 3.10 not found in the repositories\n\n");
+    # Test newer python311 in newer release
+    my $subversion = is_sle('<=15-SP4') ? 10 : 11;
```
grisu48, line 36: The main use case is for Tumbleweed, right?

Since there are multiple python versions available on TW, I would recommend to test all of them. We could introduce a variable e.g. `PYTHON_VERSIONS` and check all versions defined there.

This means we can move the python version from hardcoded values in the test code to test settings, and we can have multiple versions checked at the same time.

WDYT?

grisu48 (commented):
I would rather move the Python version to job settings, otherwise we need to constantly update the test code with every python version change in Tumbleweed.

tests/console/python3_new_version_check.pm
rfan1: That would be great, let me do

tests/console/python3_new_version_check.pm
```diff
@@ -0,0 +1,58 @@
...
+        my @python31x_results = split "\n", script_output("zypper se $python_version | awk -F \'|\' \'/$python_version/ {gsub(\" \", \"\"); print \$2}\'");
+        if ((scalar(@python31x_results) == 0) || ($python31x_results[0] ne "$python_version")) {
+            die("$python_version not found in the repositories\n\n");
+        }
+        else {
+            zypper_call("install $python_version");
+            # Running classic testing algorithm 'man_or_boy'. More info at:
+            # https://rosettacode.org/wiki/Man_or_boy_test
+            assert_script_run("curl -O " . data_url("console/man_or_boy.py"));
+            $sub_version = substr($python_version, 7);
```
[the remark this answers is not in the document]
rfan1: I agree with you, let me do

grisu48 (approved):
LGTM!

PR #16383, opened 2023-02-15
Update open source nvidia installation and registration

b10n1k:
Fix the broken nvidia test installing the correct packages. Update also other test modules in order to allow nvidia addon registration.

- Related ticket: https://progress.opensuse.org/issues/124382
- Verification run: [with registration](https://openqa.suse.de/t10504609) [without registration](https://openqa.suse.de/t10504669)

### DIFF
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -18,7 +18,7 @@ use strict;
 use warnings;
 use testapi;
 use utils;
-use version_utils qw(is_sle is_public_cloud);
+use version_utils qw(is_sle is_public_cloud get_version_id);
 use registration;
 
 our @EXPORT = qw(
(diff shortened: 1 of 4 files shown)


b10n1k:
NOTE: This goes along with https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1087

lib/publiccloud/utils.pm
```diff
@@ -107,6 +112,7 @@ sub register_addons_in_pc {
     my ($instance) = @_;
     my @addons = split(/,/, get_var('SCC_ADDONS', ''));
     my $remote = $instance->username . '@' . $instance->public_ip;
+    assert_script_run("ssh $remote sudo zypper -n --gpg-auto-import-keys ref", timeout => 300);
```
asmorodskyi, line 115:
```suggestion
    $instance->ssh_assert_script_run(cmd => "zypper -n --gpg-auto-import-keys ref", timeout => 300);
```

lib/version_utils.pm
```diff
@@ -860,7 +863,13 @@ sub get_version_id {
     $args{verid_file} //= '/etc/os-release';
 
     my $cmd = "cat $args{verid_file} | grep VERSION_ID | grep -Eo \"[[:digit:]]{1,}\\.[[:digit:]]{1,}\"";
-    $cmd = "ssh root\@$args{dst_machine} " . "$cmd" if ($args{dst_machine} ne 'localhost');
+    if ($args{dst_machine} ne 'localhost') {
```
asmorodskyi, line 866: I would introduce dedicated function instead

asmorodskyi (changes requested):
this PR needs some improvements before it can be merged see my comments above

lib/version_utils.pm
b10n1k: Why do you think that it would need a dedicated function? It would duplicate the code mostly

lib/publiccloud/utils.pm
b10n1k: *done - ready for review*

lib/version_utils.pm
asmorodskyi: ok , I thought this function is used in a lot of places but grepping the code proves me wrong . so let's leave it as is

asmorodskyi:
LGTM , but I would like at least @pdostal  to review it before we will merge it

b10n1k:
Also tell me if i need additional tests

b10n1k:
@asmorodskyi I think we need to enable this test in 15sp5 too FYI

asmorodskyi:
> @asmorodskyi I think we need to enable this test in 15sp5 too FYI

sure please PR for Job Groups repo

pdostal (approved):
All good from me

PR #16366, opened 2023-02-13
Update shortcuts for SLES 15-SP5

foursixnine:
Quck and dirty hack to try to get: https://progress.opensuse.org/issues/124347#note-3

Trying out:
openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16366 https://openqa.suse.de/tests/10478712

Created job #10483482: sle-15-SP5-Online-x86_64-Build70.1-gi-guest_developing-on-host_developing-xen@64bit-ipmi-large-mem-intel -> https://openqa.suse.de/t10483482
Created job #10483534: sle-15-SP5-Online-ppc64le-Build70.1-create_hdd_textmode@ppc64le-spvm -> https://openqa.suse.de/t10483534

PR #16384, opened 2023-02-15
Workaround for disable_grub_timeout

bs-suse:
The disable_grub_timeout module fails and blocks multiple test down the line.

Use alr-r instead of multiple right key presses.

This is a fix for an issue that should not be there. 
The multiple 'right' key presses used to work and still should, but don't.
As this issue is not reproducible manually and hard to pinpoint, I introduce this workaround fix to get the tests done that come after this module.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1208266 https://progress.opensuse.org/issues/124430
- Verification run: https://openqa.suse.de/tests/10504562#

### DIFF
--- a/tests/installation/disable_grub_timeout.pm
+++ b/tests/installation/disable_grub_timeout.pm
@@ -53,7 +53,7 @@ sub run {
     # Depending on an optional button "release notes" we need to press "tab"
     # to go to the first tab
     send_key 'tab' unless match_has_tag 'inst-bootloader-settings-first_tab_highlighted';
-    send_key_until_needlematch 'inst-bootloader-options-highlighted', 'right';
+    send_key 'alt-r';
     assert_screen 'installation-bootloader-options';
     # Select Timeout dropdown box and disable
     send_key 'alt-t';


rfan1:
Seems https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16366 did the same thing :)

tests/installation/disable_grub_timeout.pm
```diff
@@ -53,7 +53,7 @@ sub run {
     # Depending on an optional button "release notes" we need to press "tab"
     # to go to the first tab
     send_key 'tab' unless match_has_tag 'inst-bootloader-settings-first_tab_highlighted';
-    send_key_until_needlematch 'inst-bootloader-options-highlighted', 'right';
+    send_key 'alt-r';
```
foursixnine, line 56: You need to take into account that some architectures where we have different shortcuts

RoyCai7:
> Seems #16366 did the same thing :)

They are not the same thing.

foursixnine (changes requested):
This is done by #16366 already

tests/installation/disable_grub_timeout.pm
bs-github: Yes, just seen the other PR (which is better).

okurz:
closing as requested

PR #16393, opened 2023-02-15
Fix disk selection in iscsi_ibft

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/124574
- Needles: -
- [Verification run](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=rakoenig%2Fos-autoinst-distri-opensuse%23fix_iscsi_ibft)

### DIFF
--- a/schedule/yast/iscsi_ibft.yaml
+++ b/schedule/yast/iscsi_ibft.yaml
@@ -17,7 +17,9 @@ schedule:
     - installation/system_role/accept_selected_role_SLES_with_GNOME
   guided_partitioning:
     - installation/partitioning/select_guided_setup
-    - installation/partitioning_iscsi
+    - installation/partitioning/guided_setup/select_disks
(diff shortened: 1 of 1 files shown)


schedule/yast/iscsi_ibft.yaml
```diff
@@ -16,7 +16,7 @@ schedule:
   system_role:
     - installation/system_role/accept_selected_role_SLES_with_GNOME
   guided_partitioning:
-    - installation/partitioning/select_guided_setup
+    - installation/partitioning
```
jknphy, line 19: we don't want to use needle-based modules, please keep using  `select_guided_setup` and  fix the setting via MR if it is this test the one that was doing it. We should change in the future the need for a setting.

rakoenig: VR now successful with libyui based guided partitioning.

PR #16402, opened 2023-02-16
Remove softfailure because bug is fixed

asmorodskyi:


asmorodskyi:
VR: http://autobot.qa.suse.de/tests/3644

PR #16403, opened 2023-02-16
Thresholds checks switch in boottime

m-dati:
Default defined as do not check thresholds in `boottime`: checks will be executed when `PUBLIC_CLOUD_PERF_THRESH_CHECK = 1`.

- Related ticket: https://progress.opensuse.org/issues/104881#note-15
- Needles: N/D
- Verification run: will be provided soon.

### DIFF
--- a/tests/publiccloud/boottime.pm
+++ b/tests/publiccloud/boottime.pm
@@ -423,7 +423,7 @@ sub run {
 
     my $results = $self->measure_timings($args);
     $self->store_in_db($results) if (check_var('_PUBLIC_CLOUD_PERF_PUSH_DATA', 1));
-    $self->check_thresholds($results);
+    $self->check_thresholds($results) if (get_var('PUBLIC_CLOUD_PERF_THRESH_CHECK'));
 }
 
 1;
--- a/variables.md
+++ b/variables.md
@@ -301,8 +301,9 @@ PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE | boolean | false | Do not remove the instanc
 PUBLIC_CLOUD_PERF_DB_URI | string | "" | If set, the bootup times get stored in the influx database. (e.g. PUBLIC_CLOUD_PERF_DB_URI=http://publiccloud-ng.qe.suse.de:8086)
 PUBLIC_CLOUD_PERF_DB | string | "perf" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the bucket in which the performance metrics are stored
 PUBLIC_CLOUD_PERF_DB_ORG | string | "qec" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the organization in which the performance metrics are stored
-_PUBLIC_CLOUD_PERF_DB_TOKEN | string | "" | | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this required variable is the access token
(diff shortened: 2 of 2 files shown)


m-dati:
Here is a VR PASS: [test 3645](http://10.168.4.117/tests/3645#)

tests/publiccloud/boottime.pm
```diff
@@ -423,7 +423,7 @@ sub run {
 
     my $results = $self->measure_timings($args);
     $self->store_in_db($results) if (check_var('_PUBLIC_CLOUD_PERF_PUSH_DATA', 1));
-    $self->check_thresholds($results);
+    $self->check_thresholds($results) if (check_var('PUBLIC_CLOUD_PERF_THRESH_CHECK', 1));
```
ricardobranco777, line 426:
```suggestion
    $self->check_thresholds($results) if check_var('PUBLIC_CLOUD_PERF_THRESH_CHECK');
```

I believe any value that translates to True in Perl is enough.

ricardobranco777 (approved):
Just a small suggestion. LGTM anyway.

tests/publiccloud/boottime.pm
asmorodskyi: no check_var won't work this way : 

```
sub check_var {
    my ($var, $val) = @_;
    return 1 if (defined $bmwqemu::vars{$var} && $bmwqemu::vars{$var} eq $val);
    return 0;
}
```

asmorodskyi: so if $val will be undef it will always return 0

m-dati: In this case I choosed the only value ` 1 ` as explicit intention to enable the checks. But since I named it in variables.md to be boolean, indeed any value must be ok.

tests/publiccloud/boottime.pm
```diff
@@ -423,7 +423,7 @@ sub run {
 
     my $results = $self->measure_timings($args);
     $self->store_in_db($results) if (check_var('_PUBLIC_CLOUD_PERF_PUSH_DATA', 1));
-    $self->check_thresholds($results);
+    $self->check_thresholds($results) if (check_var('PUBLIC_CLOUD_PERF_THRESH_CHECK'));
```
asmorodskyi, line 426: please change it back it won't work , you need to pass 1 here

asmorodskyi (changes requested):
you need to pass value to check into check_var

tests/publiccloud/boottime.pm
m-dati: Sorry, I have read late your reply, after I changed it.

m-dati: But I could use directly **get_var**, any *not null* value is ok:
` ... if (get_var('PUBLIC_CLOUD_PERF_THRESH_CHECK'))
`

tests/publiccloud/boottime.pm
ricardobranco777: Sorry, I meant `get_var()`

m-dati:
New VR executed all PASS:  

-  test with new param set to a symbolic value 210:  
[3648](http://10.168.4.117/tests/3648) : PASS, thresholds not executed as expected

- test without new param, undef:  
[3649](http://10.168.4.117/tests/3649): PASS, thresholds *executed* and this time all checks ok all pass

PR #16405, opened 2023-02-16
Drop sound for all but SLE

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/123924
- Needles: -
- [Verification run](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=rakoenig%2Fos-autoinst-distri-opensuse%23drop_yast2_sound)

### DIFF
--- a/tests/yast2_gui/yast2_control_center.pm
+++ b/tests/yast2_gui/yast2_control_center.pm
@@ -342,7 +342,7 @@ sub run {
     start_online_update;
     start_software_repositories;
     start_printer;
-    start_sound;
+    start_sound if is_sle('<=15-sp5') or is_leap('<=15.5');
     start_sysconfig_editor;
     start_partitioner;
     start_vpn_gateway;


tests/yast2_gui/yast2_control_center.pm
```diff
@@ -342,7 +342,7 @@ sub run {
     start_online_update;
     start_software_repositories;
     start_printer;
-    start_sound;
+    start_sound if is_sle;
```
okurz, line 345: According to https://progress.opensuse.org/issues/123924#Additional-information that should be versioned, maybe

```suggestion
    start_sound if is_sle('<15-sp5');
```

okurz (changes requested):

tests/yast2_gui/yast2_control_center.pm
okurz: oh, and please check for Leap as well.

rakoenig: Trello ticket says "NOT SLE-15 SPx; we need to keep it for maintenance for that." and focusses on Tumbleweed and ALP only.

okurz: Yes, sorry, I tried to check the trello card but don't have access. Not really "open source" there, eh? ;) Well, the equivalent of "NOT SLE-15 SPx" would be:

```
start_sound unless is_sle('15-SPx');
```

which would still call `start_sound` for older versions of SLE as well as for Leap and Tumbleweed. So I am not clear on this: Is the sound module still in Tumbleweed or not?

rakoenig: The ticket was created because yast2_control_center fails in Tumbleweed since sond got dropped in YaST. But it is still there in SLE 15 SP5.

okurz: Right, that makes sense. Then it's also there in Leap 15.5 I assume. So then how about

```suggestion
    start_sound if is_sle('<=15-sp5') or is_leap('<=15.5');
```

there might be a combined statement

jknphy:
Change requested provided.

PR #16409, opened 2023-02-17
Unschedule health_check for spvm

hjluo:
We'd unschedule health_check from spvm and in the future perhaps take a look.

- Related ticket: https://progress.opensuse.org/issues/124589
- Needles: N/A
- Verification run:


jknphy:
removing WIP and merging as pvm might take a while to run.

PR #16414, opened 2023-02-17
Fix the hibernation_enabled issue on aarch64

JRivrain:
See https://progress.opensuse.org/issues/123394.
Previous builds use hibernation_disabled but now we switched to hibernation_enabled and this might be a problem, because in the default filesystem option the checkmark for "Enlarge to RAM size" for the swap partition is not set on aarch64. A question about this was answered by Josef Reidinger in the help-yast Slack channel, the default is only set for x86_64, so no hibernation is enabled on aarch. That means we cannot use the same schedule for guided_btrfs on both aarch64 and x86_64.

Other options could be to use conditional schedule, or some extra flow file for guided_btrfs.


schedule/yast/sle/guided_btrfs/guided_btrfs_aarch64.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - installation/installation_settings/validate_default_target
+  extension_module_selection:
+    - installation/module_registration/register_module_desktop
+  system_role:
+    - installation/system_role/accept_selected_role_SLES_with_GNOME
+  system_preparation:
+    - console/system_prepare
+    - console/force_scheduled_tasks
+  system_validation:
+    - console/hibernation_enabled
```
jknphy, line 20: we have to schedule `tests/console/hibernation_disabled.pm` afir

jknphy:
MR to set the schedule?

PR #16415, opened 2023-02-20
Re-write mau-qa_userspace_nfs(v4) test cases

rfan1:
https://progress.opensuse.org/issues/99636
https://progress.opensuse.org/issues/124541

- Verification run: [VR](http://openqa.suse.de/tests/overview?build=nfs_qa_run_renew&distri=sle)


dzedro:
Looks to do more than just sync, as you already looked into `/usr/share/qa/qa_test_nfs/nfs_read_write_fn.sh` it should be done IMO in https://github.com/SUSE/qa-testsuites/tree/master/tests/qa_test_nfs :smile:

rfan1:
> Looks to do more than just sync, as you already looked into `/usr/share/qa/qa_test_nfs/nfs_read_write_fn.sh` it should be done IMO in https://github.com/SUSE/qa-testsuites/tree/master/tests/qa_test_nfs smile

As we discussed, Let me re-write this test code and don't run it with qa_run.pm.

I will extracted all test scripts from the package 'qa_test_nfs' and add the workaround to the script as well

rfan1:
> > Looks to do more than just sync, as you already looked into `/usr/share/qa/qa_test_nfs/nfs_read_write_fn.sh` it should be done IMO in https://github.com/SUSE/qa-testsuites/tree/master/tests/qa_test_nfs smile
> 
> I failed to open the link above, is there any permission control there?

Now I use a simple method to re-use the shell scripts directly

tests/network/nfs.pm
```diff
@@ -0,0 +1,54 @@
...
+use utils;
+use version_utils 'is_sle';
+
+sub nfs_test {
+    my ($nfs_ver) = @_;
+    $nfs_ver //= 'nfs';
+    my @test_cases
+      = qw(nfs_start_stop nfs_mount_umount nfs_read nfs_write nfs_dontwrite nfs_usermapping_rootsquash nfs_usermapping_norootsquash nfs_usermapping_allsquash);
+    foreach my $test_case (@test_cases) {
+        record_info("$nfs_ver: $test_case Starts");
```
dzedro, line 35: I would remove the `Starts`, it makes the text just longer and harder to read
Generally the record info is at the beginning of the "operation".

dzedro (approved):

tests/network/nfs.pm
rfan1: Make sense, will do

PR #16420, opened 2023-02-20
Increase timeout for vnc_two_passwords on aarch64 because of performance issue

Zaoliang:
we still have sporadic timeout issue for closing xev or (xev not finished in time) 
see https://progress.opensuse.org/issues/119416


Zaoliang:
looks good, no single failure from 100 test runs:

https://openqa.suse.de/tests/10540126#next_previous

PR #16421, opened 2023-02-20
Improve results parsing in ha/pacemaker_cts_cluster_exerciser

alvarocarvajald:
Current version of the parser can wrongfully pick results like `test name\t failed` (see [this](https://openqa.suse.de/tests/10517143) example) which are processed later in the output due to a small issue in the `awk` command getting results from the log. This commit fixes the `awk` command while also re-writing some of the commands so the fit in under 100 characters per line.

- Related ticket: N/A
- Failing Test: https://openqa.suse.de/tests/10517143
- Needles: N/A
- Verification run: http://mango.qa.suse.de/tests/5513 (Test failure is expected, as in order to test the parser we need failures in the log)

### DIFF
--- a/tests/ha/pacemaker_cts_cluster_exerciser.pm
+++ b/tests/ha/pacemaker_cts_cluster_exerciser.pm
@@ -49,12 +49,17 @@ sub run {
 
         # Start pacemaker cts cluster exerciser
         my $cts_start_time = time;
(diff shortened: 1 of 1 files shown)


lpalovsky (approved):
LGTM

emiura:
LGTM

PR #16424, opened 2023-02-20
Remove use of create_instances in qesap_terraform module

alvarocarvajald:
Remove use of the `publiccloud::provider::create_instances` method in the `sles4sap/publiccloud/qesap_terraform` test module as many of the functions provided by that method are not required by the cloud deployment implemented by `lib/qesapdeployment`.

- Related ticket: https://jira.suse.com/browse/TEAM-7143
- Needles: N/A
- Verification run: TBA


lilyeyes:
@alvarocarvajald  
Why there is no "terraform destroy ..." related "INFO" in module: http://openqaworker15.qa.suse.cz/tests/129437#step/ssh_interactive_end/1
In module "publiccloud/ssh_interactive_end.pm" it will do instance cleanup.

Here is an example which did "cleanup": https://openqaworker15.qa.suse.cz/tests/127447#step/ssh_interactive_end/91 (there is "terraform destroy ...")

Not sure if it is introduced by your PR.

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -51,34 +54,26 @@ sub set_var_output {
...
-        push(@hana_playbook_list, "sap-hana-preconfigure.yaml -e use_sapconf=" . set_var_output("USE_SAPCONF", "true"));
-        push(@hana_playbook_list, "cluster_sbd_prep.yaml") if (check_var('FENCING_MECHANISM', 'sbd'));
-        push(@hana_playbook_list, "sap-hana-storage.yaml");
-        push(@hana_playbook_list, "sap-hana-download-media.yaml");
-        push(@hana_playbook_list, "sap-hana-install.yaml");
-        push(@hana_playbook_list, "sap-hana-system-replication.yaml");
-        push(@hana_playbook_list, "sap-hana-system-replication-hooks.yaml");
-        push(@hana_playbook_list, "sap-hana-cluster.yaml");
-
+        push @hana_playbook_list, 'pre-cluster.yaml', 'sap-hana-preconfigure.yaml -e use_sapconf=' . set_var_output('USE_SAPCONF', 'true');
```
mpagot, line 73: Personally, one playbook inclusion for line is more readable

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -95,34 +90,76 @@ sub create_hana_vars_section {
...
+        $hana_vars{primary_site} = get_var('HANA_PRIMARY_SITE', 'site_a');
+        $hana_vars{secondary_site} = get_var('HANA_SECONDARY_SITE', 'site_b');
+        set_var('SAP_SIDADM', lc(get_var('INSTANCE_SID') . 'adm'));
     }
     return (\%hana_vars);
 }
 
+=head2 create_instance_data
+
+    Create and populate a publiccloud::provider compatible memory structure with the current
```
mpagot, line 107: Why it says `memory` ?

tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -95,34 +90,76 @@ sub create_hana_vars_section {
...
+            push @instances, $instance;
+        }
+    }
+    publiccloud::instances::set_instances(@instances);
+    return \@instances;
+}
+
 sub run {
     my ($self, $run_args) = @_;
+    my $workspace = get_var('PUBLIC_CLOUD_RESOURCE_GROUP', 'qesapopenqa');
```
mpagot, line 144: Variable renaming proposal: `$workspace` --> `$resource_group`

tests/sles4sap/publiccloud/qesap_terraform.pm
mpagot: Does not make sense to move this new method to a library file, maybe an existing one? Seems a generic functionality to me.

lpalovsky (approved):
LGTM

alvarocarvajald:
> @alvarocarvajald Why there is no "terraform destroy ..." related "INFO" in module: http://openqaworker15.qa.suse.cz/tests/129437#step/ssh_interactive_end/1 In module "publiccloud/ssh_interactive_end.pm" it will do instance cleanup.
> 
> Here is an example which did "cleanup": https://openqaworker15.qa.suse.cz/tests/127447#step/ssh_interactive_end/91 (there is "terraform destroy ...")
> 
> Not sure if it is introduced by your PR.

Good catch Lili. It must have been introduced by this.

If I understand correctly, `publiccloud/ssh_interactive_end` must be accessing the existing deployment via `run_args`, and since this PR is dropping `publiccloud::create_instances()` in favor of creating the memory structure to store into `run_args` from the test module itself, it's possible the structure is not exactly the same and some compatibility was dropped.

I tagged the PR with `notready` & `WIP` while I debug.

tests/sles4sap/publiccloud/qesap_terraform.pm
alvarocarvajald: I don't think it's efficient to send so many `push` commands on the same list in succession when fewer commands accomplish the same task.

WDYT about:
```
push @hana_playbook_list,
    'pre-cluster.yaml',
    'sap-hana-preconfigure.yaml -e use_sapconf=' . set_var_output('USE_SAPCONF', 'true');
```
This way preserves the one-playbook-per-line while also not calling `push` on the same list N times consecutively.

BillAnastasiadis: I agree, and having it in multiple lines makes locating and adding/removing pushes easier

tests/sles4sap/publiccloud/qesap_terraform.pm
alvarocarvajald: > Does not make sense to move this new method to a library file, maybe an existing one? Seems a generic functionality to me.

Could be. I would propose to move it whenever we need to call this from outside the module.

mpagot: `in memory` is like `an instance of the class`, right?

tests/sles4sap/publiccloud/qesap_terraform.pm
alvarocarvajald: It's a resource group and a workspace _in Azure_, but in GCE and EC2 it's only used as a workspace.

I read this line as: "let's define a workspace; if we have a value for PUBLIC_CLOUD_RESOURCE_GROUP, we will use that as a workspace name, otherwise we will use 'qesapopenqa'", so IMO it's a better fit to call the variable `$workspace`.

I can add a comment before L144 to make this clearer.

tests/sles4sap/publiccloud/qesap_terraform.pm
alvarocarvajald: > `in memory` is like `an instance of the class`, right?

More like a list of `publiccloud::instance` class instances, but yes.

alvarocarvajald: I will re-word the documentation.

alvarocarvajald:
> @alvarocarvajald Why there is no "terraform destroy ..." related "INFO" in module: http://openqaworker15.qa.suse.cz/tests/129437#step/ssh_interactive_end/1 In module "publiccloud/ssh_interactive_end.pm" it will do instance cleanup.
> 
> Here is an example which did "cleanup": https://openqaworker15.qa.suse.cz/tests/127447#step/ssh_interactive_end/91 (there is "terraform destroy ...")
> 
> Not sure if it is introduced by your PR.

So, AFAICS, `publiccloud/ssh_interactive_end` calls `$args->{my_provider}->cleanup($args)`; in this case `$args->{my_provider` is set by `sles4sap/publiccloud/qesap_terraform` to a `publiccloud::provider` class instance, which means that `$args->{my_provider}->cleanup($args)` will call in this case `publiccloud::azure->cleanup($args)`.

`publiccloud::azure->cleanup()` calls `$self->SUPER::cleanup()` which would call `$self->terraform_destroy()`. First thing `terraform_destroy()` does is check if [terraform was applied](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L543) before [recording](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/provider.pm#L548) [this message](https://openqaworker15.qa.suse.cz/tests/127447#step/ssh_interactive_end/46).

So it seems what's missing in the new version of `qesap_terraform` is a call to `$provider->terraform_applied(1)`.

I will add this and push the changes, and then restart the VR.

mpagot (approved):
LGTM

alvarocarvajald:
@lilyeyes now saptune tests are also doing `terraform destroy`. See: https://openqaworker15.qa.suse.cz/tests/129696#step/ssh_interactive_end/51 & https://openqaworker15.qa.suse.cz/tests/129697#step/ssh_interactive_end/57

Updated the verification runs at the top.

lilyeyes (approved):
I only checked the instance create/destroy VRs it look good to me.

BillAnastasiadis (approved):
I believe everything's alright now

PR #16428, opened 2023-02-21
Ensure to press next in nis_server

tinawang123:
Ensure to press next in nis_server

- Related ticket: https://progress.opensuse.org/issues/124583
- Verification run: https://openqa.suse.de/tests/10547482#

### DIFF
--- a/tests/x11/nis_server.pm
+++ b/tests/x11/nis_server.pm
@@ -68,7 +68,7 @@ sub nis_server_configuration {
     assert_screen 'nis-server-server-maps-setup-finished';
     send_key $cmd{next};
     # NIS Server Query Hosts
-    assert_screen 'nis-server-query-hosts-setup';
+    apply_workaround_bsc1204176('nis-server-query-hosts-setup') if (is_sle('>=15-SP4'));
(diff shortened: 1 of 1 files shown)


tests/x11/nis_server.pm
```diff
@@ -68,7 +68,7 @@ sub nis_server_configuration {
     assert_screen 'nis-server-server-maps-setup-finished';
     send_key $cmd{next};
     # NIS Server Query Hosts
-    assert_screen 'nis-server-query-hosts-setup';
+    apply_workaround_bsc1204176('nis-server-query-hosts-setup') if (is_sle('>=15-SP4'));
```
lemon-suse, line 71: What will be there if VERSION less than 15-SP4?

tinawang123: It is related with bsc1204176. It doesn't have this problem except 15sp5.

lemon-suse: Ok, so this should need workaround for bsc1204176.

lemon-suse (approved):

chcao (approved):
LGTM

PR #16432, opened 2023-02-21
Run saptune (mr_test) test using autoyast install

lilyeyes:
Run saptune (mr_test) test on On-Premise x86 KVM using autoyast install on osd for new product testing 
TEAM-7344 - [AC17+AC2+AC6+AC7] Implement mr test (for new products) on x86 on-Premise

Related MR: https://gitlab.suse.de/qa-css/openqa_ha_sap/-/merge_requests/409

- Related ticket: https://jira.suse.com/browse/TEAM-7344
- Needles: NA
- Verification run (all passed): 
OSD sles4sap 15sp5: https://openqa.suse.de/tests/10596981#dependencies
OW15 sles4sap 15sp5: https://openqaworker15.qa.suse.cz/tests/130968#dependencies

### DIFF
--- a/tests/sles4sap/saptune/mr_test.pm
+++ b/tests/sles4sap/saptune/mr_test.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019-2022 SUSE LLC
+# Copyright 2019-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Summary: saptune testing with mr_test, setup mr_test environments and load mr_test
@@ -47,8 +47,8 @@ sub setup {
     select_serial_terminal;
     # Disable packagekit
     quit_packagekit;
-    # saptune is not installed by default on SLES4SAP 12 on ppc64le and in textmode profile
-    zypper_call "-n in saptune" if ((is_ppc64le() and is_sle('<15')) or check_var('DESKTOP', 'textmode'));
+    # Install saptune
+    zypper_call "-n in saptune";
     if (systemctl("-q is-active sapconf.service", ignore_failure => 1)) {
         record_soft_failure("bsc#1190787 - sapconf is not started");
         zypper_call "in sapconf";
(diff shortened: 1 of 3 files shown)


bs-suse:
I can say that the schedule and the perl module are looking good.
I'm not able to tell about autoyast_sles4sap_saptune.xml.ep. Sorry.

lilyeyes:
> I can say that the schedule and the perl module are looking good. I'm not able to tell about autoyast_sles4sap_saptune.xml.ep. Sorry.

Thanks for the review, I am not familiar it too. I just copied from "data/autoyast_sle15/autoyast_sles4sap.xml.ep" and renamed to "*_saptune.xml.ep" and deleted the "hana" related elements.

tests/sles4sap/saptune/mr_test.pm
```diff
@@ -47,8 +47,8 @@ sub setup {
     select_serial_terminal;
     # Disable packagekit
     quit_packagekit;
-    # saptune is not installed by default on SLES4SAP 12 on ppc64le and in textmode profile
-    zypper_call "-n in saptune" if ((is_ppc64le() and is_sle('<15')) or check_var('DESKTOP', 'textmode'));
```
lpalovsky, line 51: If there are still tests running for other architectures, it might be good to do a VR there just to be sure the change does not break anything.

mpagot (approved):
LGTM

tests/sles4sap/saptune/mr_test.pm
alvarocarvajald: I guess it will still work as below those lines there is a `zypper_call "-n in saptune"` without a conditional.

What I don't like about this change is that it will obscure possible bugs when the `saptune` package is not installed by default as part of the `sap_server` pattern, i.e., installing `saptune` always will guarantee the SUT has `saptune` _for_ the test, but we won't quickly see whether `saptune` was installed by the test explicitly or if it was installed as part of the SLES for SAP Applications installation.

There used to be a test scenario using the `sles4sap/patterns` and `sles4sap/saptune` test modules that would cover this, by checking that saptune gets installed as part of the pattern, but I see it disabled on 15-SP5. It's still enabled for QAM.

So TL;DR: LGTM as long as we're covering elsewhere that `saptune` gets installed by default as part of the SLES for SAP Applications product.

lilyeyes: Thanks for the comments. Good to know. @alvarocarvajald 
And here is the VR for sles4sap 12sp5: https://openqaworker15.qa.suse.cz/tests/130986#step/mr_test/17 @lpalovsky

lpalovsky (approved):
LGTM

lilyeyes:
Merge this PR since there is no more comments.

PR #16434, opened 2023-02-21
Test packagehub repo for sle micro

tonyyuan1:
Check availability of packagehub repo for sle micro

- Related ticket: https://progress.opensuse.org/issues/117418
- Needles: no
- Verification run: 
- x86_64, aarch, s390:  https://openqa.suse.de/tests/overview?build=tony.21.2_5.1_test&version=5.4&distri=sle-micro

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -252,7 +252,11 @@ sub load_tests {
         # in 10G-disk tests, we don't run more tests
         return if check_var('HDDSIZEGB', '10');
         # Stop here if we are testing only scc extensions (live, phub, ...) activation
-        return if get_var('SCC_ADDONS');
+        if (get_var('SCC_ADDONS') =~ /phub/) {
+            loadtest 'transactional/check_phub';
+            return;
+        }
+
     }
 
     load_config_tests;
--- a/tests/transactional/check_phub.pm
+++ b/tests/transactional/check_phub.pm
@@ -1,0 +1,36 @@
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Summary: Test packagehub repo on sle micro
+#
+# 1. Check if PackageHub extention is availble in sle micro
+# 2. Activate PackageHub extention
+# 3. Install a package from PackageHub repo
+#
+# Maintainer: qac <qa-c@suse.de>
+
+package check_phub;
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use transactional qw(trup_call check_reboot_changes);
+
+sub run {
+    select_console 'root-console';
+    my $out = script_output("transactional-update --quiet register --list-extensions|grep PackageHub", proceed_on_failure => 1);
+    if ($out ne "") {    #check if PackageHub extension is available
+        if ($out =~ /Activate with: transactional-update /p) {    #Check if it is activated
+            trup_call("${^POSTMATCH}");
+            check_reboot_changes;
+        }
+        trup_call("pkg install sshpass");
+        check_reboot_changes;
+        assert_script_run("rpm -q sshpass");
+    }
+    else {
+        record_info('INFO', "PackageHub Unavaible");
+    }
+}
+
+1;


waynechen55:
PackageHub repo can not be refreshed with local registration which means any subsequent operations involving "repo refreshing" will fail, because there is no local PackageHub repo on openQA or any other servers. You have to register with update.suse.com I think. And according to my limited experience, it seems PackageHub extension only available after Alpha phase.

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,40 @@
...
+    my $out = script_output("transactional-update --quiet register --list-extensions");
+    if ($out =~ m#PackageHub/(.*)/#) {
+        trup_call("register -p PackageHub/$1/" . get_var('ARCH'));
+        process_reboot(trigger => 1);
+        trup_call("pkg install sshpass");
+        process_reboot(trigger => 1);
+        validate_script_output("rpm  -qa|grep sshpass", sub { m/sshpass/ });
+    }
+    else {
+        record_info('AVC', "PackageHub Unavaible", result => 'softfail');
```
mloviska, line 31: `s/AVC/info/`, so we dont confuse this message with SELinux messages. :)

mloviska:
Just minor comments, but it looks good! Thanks Tony

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+sub post_fail_hook {
+    my ($self) = shift;
+    $self->SUPER::post_fail_hook;
```
ilausuch, line 37: Is this necessary? I am only questioning from my ignorance

lib/main_micro_alp.pm
```diff
@@ -252,7 +252,10 @@ sub load_tests {
         # in 10G-disk tests, we don't run more tests
         return if check_var('HDDSIZEGB', '10');
         # Stop here if we are testing only scc extensions (live, phub, ...) activation
-        return if get_var('SCC_ADDONS');
+        if (get_var('SCC_ADDONS')) {
```
jlausuch, line 255:
```suggestion
        if (get_var('SCC_ADDONS') =~ /phub/ && !is_released) {
```

lib/main_micro_alp.pm
```diff
@@ -252,7 +252,10 @@ sub load_tests {
...
+            loadtest 'transactional/check_phub' if (get_var('SCC_ADDONS') =~ /phub/ && !is_released);
```
jlausuch, line 256: loadtest 'transactional/check_phub';

tests/transactional/check_phub.pm
jlausuch: shouldn't we actually `die` instead of softfail? The purpose of the test is to see if we can install a package from phub

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,40 @@
...
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use version_utils 'is_sle_micro';
+use transactional;
+
+sub run {
+    select_console 'root-console';
+    my $out = script_output("transactional-update --quiet register --list-extensions");
```
jlausuch, line 22: please make a condition to register the extension if it's not registered. In this test, it will be registered already in the yast installation phase.

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,40 @@
...
+    if ($out =~ m#PackageHub/(.*)/#) {
+        trup_call("register -p PackageHub/$1/" . get_var('ARCH'));
+        process_reboot(trigger => 1);
+        trup_call("pkg install sshpass");
```
jlausuch, line 26: Can you move this out of the condition? Otherwise, this won't be run on an already phub registered system.

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,40 @@
...
+use warnings;
+use testapi;
+use version_utils 'is_sle_micro';
+use transactional;
+
+sub run {
+    select_console 'root-console';
+    my $out = script_output("transactional-update --quiet register --list-extensions");
+    if ($out =~ m#PackageHub/(.*)/#) {
+        trup_call("register -p PackageHub/$1/" . get_var('ARCH'));
```
jlausuch, line 24: Please do this step only if the phub is not registered. In the DVD test case I created, the phub extension is registered at yast installation time.

tests/transactional/check_phub.pm
mloviska: I was thinking of this as well. I guess if the test is not suitable for some version, let's just don't schedule it

tonyyuan1: This test is to check if phub extension is avaiable in repository server (proxy.scc.suse.de) provided by openQA. It's currently all-21.2.proxy.scc.suse.de. The server name might changes but we can take them as one server because of the the same scc code. 
The phub is not enabled in early developemt stage until (maybe) after Beta as @waynechen55  mentioned above. If we use "die" instead of "softfail", the test including module will always fail until phub is available.

tests/transactional/check_phub.pm
tonyyuan1: No. removed

tonyyuan1:
I changed the logic and marked some change request resolved. Feel free to unresolve them if you think they are not resolved.

tests/transactional/check_phub.pm
mloviska: sle micro is generally behind of sles schedule, so the alpha phase of phub should not be an issue over here.

tonyyuan1: Hello guys, 
Which option would you like:
1.  die,  the test will fail until phub is available in proxy.
2. record_info,  record_info('AVC') is more noticeable/highlighted than record_info('info') .  It would be easier for test reviewer to review.  

I am fine with either of option.  You are the "Latest test" reviewers.   It's your call :)

lib/main_micro_alp.pm
```diff
@@ -252,7 +252,11 @@ sub load_tests {
         # in 10G-disk tests, we don't run more tests
         return if check_var('HDDSIZEGB', '10');
         # Stop here if we are testing only scc extensions (live, phub, ...) activation
-        return if get_var('SCC_ADDONS');
+        if (get_var('SCC_ADDONS') =~ /phub/ && !is_released) {
```
jlausuch, line 255:
```suggestion
        if (get_var('SCC_ADDONS') =~ /phub/) {
```
we will run this when 5.4 is released.

tests/transactional/check_phub.pm
jlausuch: Again, Phub won't be available using SCC proxy, because it's not a module which is delivered together with SLE Micro, so it is not synchronized with SLE Micro images. So, we can only test it using real SCC registration. 

I prefer record_info, but as Martin said, `AVC` is for SELinux related messages, doesn't make sense to mix SELinux with Phub.

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,36 @@
...
+use warnings;
+use testapi;
+use transactional;
+
+sub run {
+    select_console 'root-console';
+    my $out = script_output("transactional-update --quiet register --list-extensions|grep PackageHub", proceed_on_failure => 1);
+    if ($out ne "") {    #check if PackageHub extension is available
+        if ($out =~ /Activate with: transactional-update /) {    #Check if it is activated
+            trup_call("$'");
```
mloviska, line 24: If you wish to use perlish vars, I guess this is going to be a better choice.
https://perldoc.perl.org/perlvar#$%7B%5EPOSTMATCH%7D
Firstly, it is self-explanatory and it doest no have the performance issue mentioned in https://perldoc.perl.org/perlvar#Performance-issues

tonyyuan1:
Verification tests were rerun.  https://openqa.suse.de/tests/overview?build=tony.21.2_5.1_test&version=5.4&distri=sle-micro

tests/transactional/check_phub.pm
```diff
@@ -0,0 +1,36 @@
...
+sub run {
+    select_console 'root-console';
+    my $out = script_output("transactional-update --quiet register --list-extensions|grep PackageHub", proceed_on_failure => 1);
+    if ($out ne "") {    #check if PackageHub extension is available
+        if ($out =~ /Activate with: transactional-update /p) {    #Check if it is activated
+            trup_call("${^POSTMATCH}");
+            process_reboot(trigger => 1);
+        }
+        trup_call("pkg install sshpass");
+        process_reboot(trigger => 1);
```
jlausuch, line 28: Only minor thing if you don't mind, instead of `process_reboot`, can you use `check_reboot_changes;` ? This will ensure that there is a diff between the current snapshot and previous.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/transactional.pm#L119

tonyyuan1: Yes, I can change but one thing should be noted:
The transactional-update option: "--drop-if-no-change" is enabled by default for sle micro 5.4. 
Therefore, transactional-update in "trup_call" for 5.4 will diff snapshots and discard snapshot if there were no changes to the root file system. 

There is no need to check again with "check_reboot_changes" in this test.

Do you still want to change?

mloviska: The changes are expected. Either way, it would then fail to find the package

jlausuch: Yes, in this case, we should get changes.

tonyyuan1: Changed

jlausuch (approved):
LGTM. Thanks!

tonyyuan1:
Thank you, all reviewers!

PR #16435, opened 2023-02-21
Kdump: increase timeout for aarch64

jlausuch:
Timeouts: https://openqa.suse.de/tests/10548120#step/kdump_and_crash/280 
VR: https://openqa.suse.de/tests/10549003


mloviska:
Let's try it out. Hopefully this covers the worst case scenario in openQA

PR #16439, opened 2023-02-21
Pass provider via args instead of re-creating

asmorodskyi:
VR: http://autobot.qa.suse.de/tests/3660 ( test failure unrelated to this PR and happens far away from changed code ) 
VR : http://autobot.qa.suse.de/tests/3661 ( checking that I haven't broke other upload_image tests which not passing args )


ricardobranco777 (approved):
LGTM

jlausuch (commented):
LGTM

PR #16450, opened 2023-02-22
test/container: add test user to systemd-journal

danishprakash:
add testuser to systemd-journal group to allow non-root user to access container logs via journald event driver

- https://bugzilla.suse.com/show_bug.cgi?id=1207673
- https://openqa.suse.de/tests/10556227
- https://github.com/containers/podman/issues/15159

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -21,7 +21,7 @@ use version_utils;
 use Mojo::Util 'trim';
 
 our @EXPORT = qw(test_seccomp runtime_smoke_tests basic_container_tests get_vars
-  can_build_sle_base get_docker_version check_runtime_version
+  can_build_sle_base get_docker_version get_podman_version check_runtime_version
   container_ip container_route registry_url);
 
 sub test_seccomp {
@@ -45,6 +45,10 @@ sub get_docker_version {
     my $v = script_output("docker --version");
     record_info "$v", $v =~ /(\d{2}\.\d{2})/;
     return $v =~ /(\d{2}\.\d{2})/;
+}
+
+sub get_podman_version {
+    return script_output "podman version | awk '/^Version:/ { print \$2 }'";
 }
 
 sub check_runtime_version {
@@ -86,7 +90,7 @@ sub test_update_cmd {
 
     # podman update was added to v4.3.0
     if ($runtime eq 'podman') {
-        my $version = script_output "podman version | awk '/^Version:/ { print \$2 }'";
+        my $version = get_podman_version();
         if (package_version_cmp($version, '4.3.0') <= 0) {
             record_info("SKIP", "The update command is not supported on podman $version");
             return;
--- a/tests/containers/rootless_podman.pm
+++ b/tests/containers/rootless_podman.pm
@@ -21,8 +21,8 @@ use serial_terminal 'select_serial_terminal';
 use utils;
 use containers::common;
 use containers::container_images;
-use containers::utils 'registry_url';
-use version_utils qw(is_sle is_leap is_jeos is_transactional);
+use containers::utils qw(registry_url get_podman_version);
+use version_utils qw(is_sle is_leap is_jeos is_transactional package_version_cmp);
 use power_action_utils 'power_action';
 use bootloader_setup 'add_grub_cmdline_settings';
 use Utils::Architectures;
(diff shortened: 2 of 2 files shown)


jlausuch:
I am also running it on different versions: https://openqa.suse.de/tests/overview?distri=sle&build=danishprakash%2Fos-autoinst-distri-opensuse%23danishprakash-patch-1

jlausuch (approved):

dcermak:
Do we really want to apply this non-standard config unconditionally and not only for the problematic case (podman < 4.0)?

danishprakash:
> Do we really want to apply this non-standard config unconditionally and not only for the problematic case (podman < 4.0)?

I'd say we can apply it, even If podman does the right thing >= v4.0, this change shouldn't harm us at all and would avoid flakes irrespective of which version we're testing. Besides, this shouldn't affect any functionality we're testing in this suite afaik.

mloviska:
> Do we really want to apply this non-standard config unconditionally and not only for the problematic case (podman < 4.0)?

I am also keen on the fact that we should keep it only for the problematic case. It would be easier to catch any regression if the occur in later versions

danishprakash:
> It would be easier to catch any regression if the occur in later versions

Fair enough, I think I just thought it shouldn't affect anything ideally but it should certainly be easier this way to catch regressions.

tests/containers/rootless_podman.pm
```diff
@@ -38,6 +38,16 @@ sub run {
 
     my $podman = $self->containers_factory('podman');
 
+    # add testuser to systemd-journal group to allow non-root
+    # user to access container logs via journald event driver
+    # to avoid flakes in SLE15 SP1 & SLE15 SP2 w/ Podman <=4.0.0
+    if (is_sle('=15-SP1') || is_sle('=15-SP2') || is_leap('=15.1') || is_leap('15.2')) {
```
danishprakash, line 44: @jlausuch I've added checks for Leap here too, but I'm not sure if we run these tests on Leap, can you please confirm once if this is okay?

lib/containers/utils.pm
```diff
@@ -47,6 +47,11 @@ sub get_docker_version {
     return $v =~ /(\d{2}\.\d{2})/;
 }
 
+sub get_podman_version {
+    my $v = script_output "podman version | awk '/^Version:/ { print \$2 }'";
```
jlausuch, line 51:
```suggestion
    return script_output "podman version | awk '/^Version:/ { print \$2 }'";
```

lib/containers/utils.pm
```diff
@@ -86,7 +91,7 @@ sub test_update_cmd {
 
     # podman update was added to v4.3.0
     if ($runtime eq 'podman') {
-        my $version = script_output "podman version | awk '/^Version:/ { print \$2 }'";
+        my $version = get_podman_version()
```
jlausuch, line 94:
```suggestion
        my $version = get_podman_version();
```

tests/containers/rootless_podman.pm
jlausuch: Why don't we run this condition regardless of the OS (only checking the podman version)?

jlausuch (changes requested):

tests/containers/rootless_podman.pm
danishprakash: That's because it was reported only for SLE15SP1 and 2 and since we wanted to just add the condition for the specific failing case, I added this.

jlausuch: Stil, I would do it version dependant, not OS. Cause the version in each OS can change if there are updates requested within the general support or LTSS time.

jlausuch (approved):

lib/containers/utils.pm
jlausuch: Ok, here there is something you need to change due to Perl syntax .. you can't declare a variable inside an `if`.
Please do this:
```
my $version;
if ($runtime eq 'podman') {
    $version = get_podman_version();
    if (package_version_cmp($version, '4.3.0') <= 0) {
        record_info("SKIP", "The update command is not supported on podman $version");
        return;
    }
} ....
```

or if you wanna simplify it:
```
if ($runtime eq 'podman' && package_version_cmp(get_podman_version, '4.3.0') <= 0) {
    record_info("SKIP", "The update command is not supported on podman version");
    return;
}
```

jlausuch (changes requested):
This code doesn't work: https://openqa.suse.de/tests/10576954#

tests/containers/rootless_podman.pm
```diff
@@ -21,7 +21,7 @@ use serial_terminal 'select_serial_terminal';
 use utils;
 use containers::common;
 use containers::container_images;
-use containers::utils 'registry_url';
+use containers::utils qw(registry_url get_podman_version);
 use version_utils qw(is_sle is_leap is_jeos is_transactional);
```
jlausuch, line 25:
```suggestion
use version_utils qw(is_sle is_leap is_jeos is_transactional package_version_cmp);
```

jlausuch:
VRs: 
https://openqa.suse.de/tests/10579495#
https://openqa.suse.de/tests/10579496#

jlausuch (approved):

mloviska (approved):

PR #16453, opened 2023-02-22
Improve btrfsmaintenance

grisu48:
Improve the test structure of the btrfsmaintenance test run. Reorganize the test run and improve the error output of certain commands.

- Related ticket: https://progress.opensuse.org/issues/124898
- Verification run: [Tumbleweed](https://duck-norris.qe.suse.de/tests/12098) | [15-SP4](https://duck-norris.qe.suse.de/tests/12091) | [15-SP3](https://duck-norris.qe.suse.de/tests/12092) | [15-SP2](https://duck-norris.qe.suse.de/tests/12093) | [15-SP1](https://duck-norris.qe.suse.de/tests/12095) | [12-SP5](https://duck-norris.qe.suse.de/tests/12094) | [12-SP4](https://duck-norris.qe.suse.de/tests/12096) | [12-SP3](https://duck-norris.qe.suse.de/tests/12097)

### DIFF
--- a/tests/console/btrfsmaintenance.pm
+++ b/tests/console/btrfsmaintenance.pm
@@ -20,6 +20,8 @@ use serial_terminal 'select_serial_terminal';
 use utils;
 use version_utils qw(is_leap is_sle);
 
+my $reinstall_btrfsmaintenance = 0;
+
 sub btrfs_service_unavailable {
     my $service = $_[0];
     # Check if the given btrfs service (e.g. scrub or balance) is enabled in one of the following methods:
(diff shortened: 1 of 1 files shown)


tests/console/btrfsmaintenance.pm
```diff
@@ -41,45 +43,80 @@ sub btrfs_service_unavailable {
     }
 }
 
+sub btrfs_await_scrub {
+    if (is_sle("<15-SP4")) {
+        script_retry('btrfs scrub status / | grep -e "scrub started .* and finished after"', retry => 20, delay => 30);
```
rfan1, line 48: One minor comment,

Why do we need retry many times with short `delay` time for these functions? 

Does it make sense to reduce the `retry` value and increase the "delay` time?

rfan1, line 48: One minor comment,  It seems retry many times with short `delay` time for these functions,
Does it make sense to reduce the `retry` value and increase the "delay` time?

grisu48: We could, but a delay of 30 seconds is rather generous already. If we increase this further then we also block the openQA workers unnecessarily long.

Also in the test runs this `script_retry` typically only takes 1-2 iterations to complete.

So, I would argue increasing this delay here does not bring us much more.

grisu48: I think this is a duplicate, so I'm marking it as resolved.

rfan1 (approved):

PR #16455, opened 2023-02-23
Drop use of absolute path because bsc#1205325 is fixed

asmorodskyi:


asmorodskyi:
VRs showed that bug is still reproducible moving PR to a draft until issue will be fixed

PR #16457, opened 2023-02-23
Replace softfail of bsc1204176 with poo reference

ge0r:
Bug bsc#1204176 has been marked as WONTFIX by the yast developers. The bug can be only reproduced in the openQA environment and, as such, it is now considered an openQA framework issue.
The resulting progress issue ticke is poo#124652 and it will be the new reference point for the gtk refresh issue workaround.

- Related ticket: No ticket
- Needles: No needles
- Verification runs: https://openqa.suse.de/tests/overview?build=ge0r%2Fos-autoinst-distri-opensuse%23replace-softfailure-bsc1204176&version=15-SP5&distri=sle


rakoenig:
LGTM

PR #16458, opened 2023-02-23
Introduce cleanup logic in helm test

pdostal:
1) Implement k8s old namespace cleanup
2) Assert classic namespace deletion

- Related ticket: [poo#123511](https://progress.opensuse.org/issues/123511)
- Verification run: [pdostal-server.suse.cz/t4000](https://pdostal-server.suse.cz/t4000)

### DIFF
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -102,7 +102,7 @@ sub run {
     assert_script_run("helm uninstall helm-test-$job_id");
 
     assert_script_run("kubectl config set-context --current --namespace=default");
-    script_run("kubectl delete namespace helm-ns-$job_id");
+    assert_script_run("kubectl delete namespace helm-ns-$job_id");
 
     # Add repo, search and show values
     assert_script_run(
--- a/tests/containers/helm.pm
+++ b/tests/containers/helm.pm
@@ -23,6 +23,8 @@ use mmapi 'get_current_job_id';
 use registration qw(add_suseconnect_product get_addon_fullname);
 use containers::k8s;
 use publiccloud::utils qw(gcloud_install);
+use JSON 'decode_json';
+use Date::Parse 'str2time';
 
 sub run {
     my ($self, $run_args) = @_;
@@ -109,6 +111,7 @@ sub run {
     assert_script_run("helm search repo apache");
     assert_script_run("helm show all $chart");
 
+    helm_cleanup() unless $is_k3s;
     uninstall_k3s() if $is_k3s;
 }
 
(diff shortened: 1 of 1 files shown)


tests/containers/helm.pm
```diff
@@ -109,6 +112,7 @@ sub run {
     assert_script_run("helm search repo apache");
     assert_script_run("helm show all $chart");
 
+    helm_cleanup() unless $self->{is_k3s};
```
asmorodskyi, line 115: does it make sense that one line below there is variable  with same name but not from `$self` ?

tests/containers/helm.pm
```diff
@@ -121,9 +125,31 @@ sub post_fail_hook {
     script_run("kubectl config set-context --current --namespace=default");
     script_run("kubectl delete namespace helm-ns-$job_id");
 
+    helm_cleanup() unless $self->{is_k3s};
     uninstall_k3s() if $self->{is_k3s};
 }
 
+sub helm_cleanup {
+    diag("helm_cleanup started.");
```
asmorodskyi, line 133: AFAIK diag will only print something in os-autoinst log , maybe better to do record_info instead ? ( or in addition ? )

tests/containers/helm.pm
```diff
@@ -121,9 +125,31 @@ sub post_fail_hook {
...
+    my $namespaces_output = script_output("kubectl get namespaces -o=json");
+    my $namespaces_json = decode_json($namespaces_output);
+    my @namespaces = @{$namespaces_json->{items}};
+    foreach my $item (@namespaces) {
+        next if ($item->{metadata}->{name} !~ /^helm-ns-/);
+        if ($item->{metadata}->{creationTimestamp}) {
+            my $age = time() - str2time($item->{metadata}->{creationTimestamp});
+            if ($age > 10800) { # 3 hours
+                diag("The namespace $item->{metadata}->{name} is $age seconds old. It will be removed.");
+                script_run("kubectl delete all --all -n " . $item->{metadata}->{name});
```
asmorodskyi, line 143: I am not sure that it is good idea to put something like into the test , imagine that 2/3/5 tests will execute such code in parallel we can catch some not nice bugs in k8s cluster ...

tests/containers/helm.pm
pdostal: I switched from `record_info` to `diag` to not clutter the test overview that much. those messages are anyways not necessary for the 'test' - they're just 'administrative' informations.

tests/containers/helm.pm
pdostal: Good catch

tests/containers/helm.pm
asmorodskyi: > I don't understand. Only 3 hours old namespaces will be deleted. So most of the time none. If none is deleted then the only API call this does is `kubectl get namespaces -o=json` which is pretty minimal.

containers tests are running in really different setup for different versions of SLE , please tell me how many times this code will run during one day ? 

> But there is question if this should be here or in PCW...

IMO this task is 100% match PCW and 0% match test code

ilausuch: I agree with Anton, this should be moved to PCW. We had similar discussion for jobs and the conclussions were that the PCW is the best place to do this

PR #16462, opened 2023-02-24
Pass timeout/interval when using regex to find UI element

hjluo:
When using regex to find UI element with libyui-rest-api, we need to pass timeout/interval as a parameter to the find function while don't break anything.

- Related ticket: https://progress.opensuse.org/issues/121807
- Needles: N/A
- Verification run:
  * ARM MU [install_SLE](https://openqa.suse.de/tests/10596510#step/install_SLES/2)  with 
  TIMEoUT_SCALE=1
  https://openqa.suse.de/tests/10634836/ **FAILED**
  TIMEOUT_SCALE=3
  https://openqa.suse.de/tests/10640655
  https://openqa.suse.de/tests/10640646

  * regression with  add_ha_regcode in yast daily
   https://openqa.suse.de/tests/10640647
   https://openqa.suse.de/tests/10640648
   https://openqa.suse.de/tests/10640649
   https://openqa.suse.de/tests/10640650
   https://openqa.suse.de/tests/10640651
   https://openqa.suse.de/tests/10640652
   https://openqa.suse.de/tests/10640653
   https://openqa.suse.de/tests/10640654

   * MU [lvm_thin_provisioning](https://openqa.suse.de/tests/10583641 )
     https://openqa.suse.de/tests/10635628
     https://openqa.suse.de/tests/10640658
  * regression on yast daily
     https://openqa.suse.de/tests/10614955
     https://openqa.suse.de/tests/10614956
     https://openqa.suse.de/tests/10614957
     https://openqa.suse.de/tests/10614958
     https://openqa.suse.de/tests/10614959
     https://openqa.suse.de/tests/10614960
     https://openqa.suse.de/tests/10614955

### DIFF
--- a/lib/YuiRestClient.pm
+++ b/lib/YuiRestClient.pm
@@ -49,6 +49,10 @@ sub get_port {
     return $port;
 }
 
+sub get_timeout {
+    return $app->{timeout};
+}
+
 sub set_host {
     my ($yuihost) = @_;
     $host = $yuihost;
@@ -215,6 +219,8 @@ B<get_host(%args)> - returns name or IP of the REST server.
 The %args is a boolean $installation (see parameters of C<get_app()> above.)
 
 B<get_port()> - returns port number for the rest server.
+
+B<get_timeout()> - returns the current timeout.
 
 B<init_logger> - Initializes logger instance.
 
--- a/tests/installation/product_selection/install_SLES.pm
+++ b/tests/installation/product_selection/install_SLES.pm
@@ -5,11 +5,15 @@
 # Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
 
 use base 'y2_installbase';
+use testapi 'get_var';
 use strict;
 use warnings;
 
 sub run {
-    $testapi::distri->get_product_selection()->install_product('SLES');
+    my $product_selection = $testapi::distri->get_product_selection();
+
+    $product_selection->wait_product_selection_page({timeout_scale => get_var('TIMEOUT_SCALE', 1), message => 'Product Selection page is not displayed'});
+    $product_selection->install_product('SLES');
 }
 
 1;
--- a/lib/Installation/ProductSelection/ProductSelectionController.pm
+++ b/lib/Installation/ProductSelection/ProductSelectionController.pm
@@ -32,6 +32,14 @@ sub get_product_selection_page {
     return $self->{ProductSelectionPage};
 }
 
+sub wait_product_selection_page {
+    my ($self, $args) = @_;
+    $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
+    YuiRestClient::Wait::wait_until(object => sub {
+            $self->{ProductSelectionPage}->is_shown({timeout => 0});
+    }, %$args);
+}
+
 sub get_access_beta_distribution_popup {
     my ($self) = @_;
     die "Popup for accessing Beta Distribution is not displayed" unless $self->{AccessBetaDistributionPopup}->is_shown();


rakoenig:
LGTM.

lib/Installation/ProductSelection/ProductSelectionPage.pm
```diff
@@ -34,7 +34,7 @@ sub init {
 
 sub is_shown {
     my ($self) = @_;
-    return $self->{rdb_SLES}->exist();
+    return $self->{rdb_SLES}->exist({timeout => 300, interval => 10});
```
jknphy, line 37: we don't need to have this for every single product selection, I know you use it to probe that his works, but we are probably increasing the timeout where is not needed to 300, hiding a potential bug.

hjluo: how about just 300 for Aarch64 for the others use the default(120)?

jknphy: we might have done something like this in a rush in the past, but the normal way is to put the timeout from outside, in the test layer, and pass these two parameters through the layer to the is_shown to here when needed, so we can multiply normal timeout by some number. We don't do those hacks of hide conditions in this framework " in general " like it is done over the place for testapi.

hjluo: I'd think of a way to pass the timeout from the outer to the inner part as you described.

jknphy: ok, you want to do the extra mile :)
check for example: `installation/performing_installation/perform_installation.pm`

hjluo: the is_shown was controlled by NavigationBase, and called before the  install_product. so we can't pass timeout from the outer install_SLE.pm.
so I just add TIMOUT_SCALE in  ProductSelectionPage.

lib/Installation/ProductSelection/ProductSelectionPage.pm
```diff
@@ -33,8 +34,9 @@ sub init {
 }
 
 sub is_shown {
-    my ($self) = @_;
-    return $self->{rdb_SLES}->exist();
+    my ($self, %args) = @_;
+    my $timeout = 120 * get_var('TIMEOUT_SCALE', 1);
```
jknphy, line 38: this need to be done in the level of Test layer, Page layer only cares about being a repository of ids according to the pattern, that is why most of the function are really simple, doing it here breaks the design.

hjluo: OK, I'd think about a way to fix this. Thanks.

tests/installation/product_selection/install_SLES.pm
```diff
@@ -5,11 +5,15 @@
 # Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
 
 use base 'y2_installbase';
+use testapi 'get_var';
 use strict;
 use warnings;
 
 sub run {
-    $testapi::distri->get_product_selection()->install_product('SLES');
+    my $timeout = 120 * get_var('TIMEOUT_SCALE', 1);
```
jknphy, line 13: isn't the default for libyui-rest-api 30 seconds?

tests/installation/product_selection/install_SLES.pm
```diff
@@ -5,11 +5,15 @@
...
+    my $product_selection = $testapi::distri->get_product_selection();
+    $product_selection->get_product_selection_page({timeout => $timeout, interval => 10, message => 'Product Selection page is not displayed'});
```
jknphy, line 15: how does it look the error here? we normally create a specific function for waiting, not reusing the existing one, there is a die message and and message you passed here, one of them might not be displayed, right?

tests/installation/product_selection/install_SLES.pm
hjluo: It's 120 seconds, I've seen in the log. and you can check from [**yui-log**](https://openqa.suse.de/tests/10609779/logfile?filename=yui-log.txt) it start from `[2023-03-02T10:22:27.772353+01:00] `to to

tests/installation/product_selection/install_SLES.pm
hjluo: I've added another function `wait_product_selection_page` and avoid touching the Base.pm for regression.  and updated all VRs.  please take a look again. Thanks.

tests/installation/product_selection/install_SLES.pm
jknphy: unfortunately we did that :(
please create a method `get_timeout` in `lib/YuiRestClient.pm` to match both values and not hardcode it.

jknphy:
LGTM other than that hardcoded 120, but easy fix and we can merge it.

hjluo:
> LGTM other than that hardcoded 120, but easy fix and we can merge it.

Added get_timeout and now all VRs are updated. Thanks a lot for the comments.

lib/YuiRestClient.pm
```diff
@@ -49,6 +49,10 @@ sub get_port {
     return $port;
 }
 
+sub get_timeout {
+    return ($app->{timeout}, $app->{interval});
```
jknphy, line 53: one at at time, please, it is not good practice to return two values, specially because is expected only one by the name of the function.
Besides we only need the timeout to know its current value, not the interval, which we don't modify it, and the interval depends on the test.

tests/installation/product_selection/install_SLES.pm
```diff
@@ -5,11 +5,17 @@
 # Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
 
 use base 'y2_installbase';
+use testapi 'get_var';
+use YuiRestClient 'get_timeout';
```
jknphy, line 9: actually we should pass the timeout scale as parameter to not introduce dependency with YuiRestClient (sorry I missed that...) but then we would repeat that operation in several places and we need some object to encapsulate, could you please create a ticket for that?

lib/YuiRestClient.pm
```diff
@@ -216,6 +220,8 @@ The %args is a boolean $installation (see parameters of C<get_app()> above.)
 
 B<get_port()> - returns port number for the rest server.
 
+B<get_timeout()> - returns the default timeout.
```
jknphy, line 223: returns the current timeout.

tests/installation/product_selection/install_SLES.pm
hjluo: so, how about we move the get_timeout to the lib/YuiRestClient/App.pm, we can have set_timeout and get_timeout there, then we can use the  logic 
like: 
```
my $app = YuiRestClient::get_app(timeout => 120, interval => 1);
my $timeout = $app->get_timeout();
```

jknphy: we don't need the dependency of the App in the Test Layer neither.

hjluo: OK,  then I'd create a ticket for it. thanks

jknphy:
```
$product_selection->wait_product_selection_page({timeout_scale => get_var('TIMEOUT_SCALE', 1), message => 'Product Selection page is not displayed'});
```
Then we do the multiplication in the Controller Layer, where we already have `YuiRestClient`. But this multiplication should be encapsulated in the future, you can do it like this and file the ticket just for the encapsulation ( a new object that does the operation).

hjluo: New ticket [**125561**](https://progress.opensuse.org/issues/125561) was created.

PR #16465, opened 2023-02-24
sqlite3: adhere to the SQL standard when it comes to string literals

DimStar77:
The SQL standard requires double-quotes around identifiers and single-quotes around string literals. For example:

    "this is a legal SQL column name"
    'this is an SQL string literal'

- Related ticket: https://bugzilla.opensuse.org/show_bug.cgi?id=1208636 (was thought to be a prod bug first)
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/3141813


DimStar77:
This fixes issues with sqlite 3.41 - which became stricter in enforcing the SQL standard when executed via CLI

foursixnine (approved):

PR #16470, opened 2023-02-27
lib/nfs_common: Detect if the kernel does not have NFSv2 support

Vogtinator:
- Related ticket: https://progress.opensuse.org/issues/125072
- Verification run: https://openqa.opensuse.org/tests/3147227

### DIFF
--- a/lib/nfs_common.pm
+++ b/lib/nfs_common.pm
@@ -29,6 +29,18 @@ sub server_configure_network {
 sub try_nfsv2 {
     # Try that NFSv2 is disabled by default
     systemctl 'start nfs-server';
+
+    # Make sure it exists and also print the content as info.
+    assert_script_run 'cat /proc/fs/nfsd/versions', fail_message => 'NFS versions file must exist';
+
+    # It's ok if it's not available in the kernel
+    if (script_run("grep -q '[+-]2' /proc/fs/nfsd/versions") == 1) {
+        record_info('Info', 'NFSv2 not available in the kernel');
+        systemctl 'stop nfs-server';
+        return;
+    }
+
+    # If available, make sure it's disabled by default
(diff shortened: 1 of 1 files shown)


lib/nfs_common.pm
```diff
@@ -29,6 +29,18 @@ sub server_configure_network {
 sub try_nfsv2 {
     # Try that NFSv2 is disabled by default
     systemctl 'start nfs-server';
+
+    # Make sure it exists and also print the content as info.
+    assert_script_run "cat /proc/fs/nfsd/versions";
```
okurz, line 34: you can consider using a `fail_message` which would also be a way to replace the comment with a text string with function

```suggestion
    assert_script_run 'cat /proc/fs/nfsd/versions', fail_message => 'NFS versions file must exist';
```

lib/nfs_common.pm
```diff
@@ -29,6 +29,18 @@ sub server_configure_network {
...
+
+    # It's ok if it's not available in the kernel
+    if (script_run("grep -q '[+-]2' /proc/fs/nfsd/versions") == 1) {
+        record_info("Info", "NFS 2 not available in the kernel");
```
okurz, line 38:
```suggestion
        record_info('Info', 'NFS 2 not available in the kernel');
```

lib/nfs_common.pm
Vogtinator: That's missing the "also print content as info" part

okurz: well, you get the idea

lib/nfs_common.pm
okurz: you gotta start somewhere :)

Vogtinator: Changed.

lib/nfs_common.pm
Vogtinator: Changed.

PR #16476, opened 2023-02-27
Split boottime measure call and provider instance management

m-dati:
In order to prepare the boottime measure triggerring in the provider instance creation, a refactoring is needed:
  - in sub `measure_timings()`, move the instance management code part in a _new sub_ `provider_instance()`;
  - in sub run(), change `measure_timings `input parameter to use _$instance_ only;
  - then add just before the above call, the new _sub call_

- Needles: No change
- Verification run: Latest Azure 15 SP5 test [3678](http://10.168.4.117/tests/3678#)


pdostal:
What if we reuse `tests/publiccloud/prepare_instance.pm`?

m-dati:
new VR:
  - test [3680](http://10.168.4.117/tests/3680#)

m-dati:
> What if we reuse `tests/publiccloud/prepare_instance.pm`?

We can discuss on that :)

m-dati:
New code updated
VR: P.C. LTP test with additional parameters for triggering tests: [3698](http://10.168.4.117/tests/3698)

Moved from boottime.md to instance.md:
systemd_time_to_sec
extract_analyze
extract_blame
do_systemd_analyze 
store_in_db

copied measure_timings from boottime.md to instance.md and adapted in measure_boottime and adapted to the new call from create_instances in provider.pm

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,202 @@ sub network_speed_test() {
...
+Data is then collected in an internal record, ready for storing in a DB.
+- when PUBLIC_CLOUD_PERF_COLLECT=0 boottime measurement disabled;
+- when bit-0 ON,First reboot only enabled =1;
+- when bit-1 ON, first,softreboot enabled =3 [2 too];
+- when bit-2 ON, first,hardreboot enabled =5 [4 too];
+- when all bits ON,all 3 measures enabled =7 [6 too].
+
+=cut
+
+sub measure_boottime() {
```
asmorodskyi, line 499: I would keep this function responsibility purely on collecting boottime data , in case we would like to collect data after soft/hard reboot this should look like (psedo-code just to give idea ):

```
$instance->softreboot();
measure_boottime(); 
```

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,202 @@ sub network_speed_test() {
...
+        $ret->{analyze}->{$_ . '_hard'} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
+        $ret->{blame}->{hard} = $systemd_blame;
+    }
+
+    # Do logging to openqa UI
+    $Data::Dumper::Sortkeys = 1;
+    record_info("RESULTS", Dumper($ret));
+
+    $instance->run_ssh_command(cmd => 'sudo tar -czvf /tmp/sle_cloud.tar.gz /var/log/cloudregister /var/log/cloud-init.log /var/log/cloud-init-output.log /var/log/messages /var/log/NetworkManager', proceed_on_failure => 1);
+    $instance->upload_log('/tmp/sle_cloud.tar.gz');
```
asmorodskyi, line 550: upload log also has proceed_on_failure key , we should make this code transparent to all other setup . we don't want ltp test to fail because it can not measure boot time

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,202 @@ sub network_speed_test() {
...
+    };
+    record_info("BOOT TIME", 'systemd_analyze');
+    # first deployment analysis
+    my ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
+    $ret->{analyze}->{$_} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
+    $ret->{blame}->{first} = $systemd_blame;
+
+    # Collect kernel version
+    $ret->{kernel_release} = $instance->run_ssh_command(cmd => 'uname -r');
+    $ret->{kernel_version} = $instance->run_ssh_command(cmd => 'uname -v');
```
asmorodskyi, line 523: every command call should be protected from failing everything

tests/publiccloud/boottime.pm
```diff
@@ -228,71 +228,25 @@ our $thresholds_by_flavor = {
...
-    my $output = "";
-    my @ret;
-
-    # Wait for guest register, before calling syastemd-analyze
-    $instance->wait_for_guestregister();
-    while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
-        $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
-        sleep 5;
+    if (get_var('PUBLIC_CLOUD_QAM')) {
+        $provider = $args->{my_provider};
```
asmorodskyi, line 237: this does not make any sense because we not using provider variable when it is QAM run

tests/publiccloud/boottime.pm
```diff
@@ -228,71 +228,25 @@ our $thresholds_by_flavor = {
...
+        $instance = $args->{my_instance};
+    } else {
+        $provider = $self->provider_factory();
```
asmorodskyi, line 240: this is only place where we using this variable and just in next line so you may do `$self->provider_factory()->create_instance(check_connectivity => 0);`  or keep it like this if you don't like long lines

lib/publiccloud/provider.pm
```diff
@@ -313,6 +313,9 @@ sub create_instances {
             # Install server's ssh publicckeys to prevent authenticity interactions
             assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
         }
+        # Performance data: boottime
+        my $btime = $instance->measure_boottime($instance);
```
asmorodskyi, line 317: this calls should be controlled by some variable so we can switch this off if it will start bringing troubles ...

tests/publiccloud/boottime.pm
```diff
@@ -228,71 +228,25 @@ our $thresholds_by_flavor = {
     },
 };
 
+sub provider_instance {
+    my ($self, $args) = @_;
+    my $provider;
```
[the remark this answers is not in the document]
m-dati: Sorry but I see that` $provide` it is used in the `QAM false ` branch: should I move declaration there?

tests/publiccloud/boottime.pm
m-dati: ok ,true I'll remove from this branch.

tests/publiccloud/boottime.pm
m-dati: Possibly, to keep readability of the different functions and classes,  I'd prefer:
 `my $provider =  $self->provider_factory();`

lib/publiccloud/provider.pm
m-dati: the enabling variables checks are inside those functions `measure` and `store`.
But, do you mean that those function calls shall be moved outside `create_instance` instead?

m-dati:
Please consider that for this step I keep the `boottime.pm` funciontality still working,until it will be decided to remove the module, with logic flow adapted to the new status of enabling parameters and routines.

tests/publiccloud/boottime.pm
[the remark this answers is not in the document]
asmorodskyi: yes

tests/publiccloud/boottime.pm
asmorodskyi: ok fine for me

lib/publiccloud/instance.pm
m-dati: I set the parameter `failok => 1`, as in [upload_logs](https://open.qa/api/testapi/#_upload_logs)

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,202 @@ sub network_speed_test() {
...
+        analyze => {},
+        blame => {
+            first => {}, soft => {}, hard => {}
+        },
+    };
+    record_info("BOOT TIME", 'systemd_analyze');
+    # first deployment analysis
+    my ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
+    $ret->{analyze}->{$_} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
+    $ret->{blame}->{first} = $systemd_blame;
```
asmorodskyi, line 519:
```suggestion
    $ret->{blame} = $systemd_blame;
    $ret->{type} = 'first';
```

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,202 @@ sub network_speed_test() {
...
+- when PUBLIC_CLOUD_PERF_COLLECT=0 boottime measurement disabled;
+- when bit-0 ON,First reboot only enabled =1;
+- when bit-1 ON, first,softreboot enabled =3 [2 too];
+- when bit-2 ON, first,hardreboot enabled =5 [4 too];
+- when all bits ON,all 3 measures enabled =7 [6 too].
+
+=cut
+
+sub measure_boottime() {
+    my ($self, $instance) = @_;
```
asmorodskyi, line 500:
```suggestion
    my ($self, $instance,$type) = @_;
```

lib/publiccloud/instance.pm
m-dati: ok.

**Note**: considering also the proposal to restore the original `boottime.md` until removal decision made, after this data_record will be changed here and on _Influxdb_ too, it will be _different_ from the original one.
Therefore `boottime.md` can be used for only _local display_, having care to always set  `PUBLIC_CLOUD_PERF_COLLECT` and `_PUBLIC_CLOUD_PERF_PUSH_DATA` to **false**

m-dati:
As agreed, the above changes will be applied in a separated **new branch** and a **new PR** will be submitted soon too.

lib/publiccloud/provider.pm
m-dati: As discussed, the enabling variables checks location could be ok.

tests/publiccloud/boottime.pm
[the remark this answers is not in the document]
m-dati: Considering last proposal to keep the original functionality of this module, until the integration inside _create_instance_ will be completed, possibly no change to the original code will be applied other than corrective.

tests/publiccloud/boottime.pm
m-dati: See last comment above about boottime.pm

tests/publiccloud/boottime.pm
m-dati: See last comment above about boottime.pm

PR #16477, opened 2023-02-28
Slow down the command execution speed to give Hyper-V some response time

RoyCai7:
poo#[121027](https://progress.opensuse.org/issues/121027)

- Related ticket: [121027](https://progress.opensuse.org/issues/121027)
- Needles: N/A
- Verification run:[ legacy + hyperv](http://openqa.suse.de/tests/10604267#step/grub_test/1) [UEFI + hyperv](http://openqa.suse.de/tests/10604264#step/grub_test/1)

### DIFF
--- a/lib/grub_utils.pm
+++ b/lib/grub_utils.pm
@@ -47,11 +47,11 @@ sub grub_test {
         bug_workaround_bsc1005313() unless get_var("BOOT_TO_SNAPSHOT");
     }
     else {
+        # Slow down the command 'return' execution speed to give Hyper-V some response time
+        wait_still_screen 2 if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
         # avoid timeout for booting to HDD
         send_key 'ret';
     }
-    # Avoid return key not received occasionally for hyperv-uefi guest at first boot
-    send_key 'ret' if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
 }
 
 =head2 handle_installer_medium_bootup


RoyCai7:
Monitoring some time whether not hit the same issue in future.
Already done some tests in parallel, not hit it.
http://openqa.suse.de/t10596897
http://openqa.suse.de/t10596898
http://openqa.suse.de/t10596899
http://openqa.suse.de/t10596900
http://openqa.suse.de/t10596901
http://openqa.suse.de/t10596902

lib/grub_utils.pm
```diff
@@ -47,11 +47,11 @@ sub grub_test {
         bug_workaround_bsc1005313() unless get_var("BOOT_TO_SNAPSHOT");
     }
     else {
+        # Avoid return key not received occasionally for hyperv-uefi guest at first boot
+        sleep 3 if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
```
waynechen55, line 51: I think you can detect something before taking action instead of just sleep

okurz: That's right. 
You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

RoyCai7:
@alice-suse @xguo @waynechen55 @Julie-CAO @tbaev @nanzhg @varunkojha please help to review this PR. Thanks

lib/grub_utils.pm
```diff
@@ -47,11 +47,11 @@ sub grub_test {
         bug_workaround_bsc1005313() unless get_var("BOOT_TO_SNAPSHOT");
     }
     else {
+        # Slow down the command 'return' execution speed to give Hyper-V some response time
+        wait_still_screen 2 if (check_var('VIRSH_VMM_FAMILY', 'hyperv') && get_var('UEFI'));
```
okurz, line 51: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16477#discussion_r1121176445 still applies here. `wait_still_screen` is effectively a "sleep *at least* 2s". Please reconsider to find out the *actual event* you need to wait for.

PR #16479, opened 2023-02-28
Fix License screen in s390x, which is now using different UI controls

hjluo:
The license screen for s390x is using different UI controls because now SUMA is not there.

- Related ticket: https://progress.opensuse.org/issues/124418
- Needles: N/A
- Verification run: 
  * s390x:
  https://openqa.suse.de/tests/10603872
  https://openqa.suse.de/tests/10603874
  * regression:
  https://openqa.suse.de/tests/10603875
  https://openqa.suse.de/tests/10603876
  https://openqa.suse.de/tests/10603877

### DIFF
--- a/lib/Installation/License/Sle/LicenseAgreementController.pm
+++ b/lib/Installation/License/Sle/LicenseAgreementController.pm
@@ -20,11 +20,11 @@ sub init {
     my ($self, $args) = @_;
     $self->{AcceptLicensePopup} = Installation::License::AcceptLicensePopup->new({
             app => YuiRestClient::get_app(),
-            btn_ok_filter => {id => 'ok'}});
+            btn_ok_filter => {id => qr/ok_msg|ok/}});
     $self->{LicenseAgreementPage} = Installation::License::LicenseAgreementExplicitPage->new({
             app => YuiRestClient::get_app(),
             chb_accept_license_filter => {id => '"Y2Packager::Widgets::ProductLicenseConfirmation"'},
-            cmb_language_filter => {id => '"simple_language_selection"'},
+            cmb_language_filter => {id => qr/Y2Country::Widgets::LanguageSelection|simple_language_selection/},
             rct_eula_filter => {id => '"CWM::RichText"'}});
     return $self;
 }
--- a/tests/installation/licensing/verify_license_translations.pm
+++ b/tests/installation/licensing/verify_license_translations.pm
@@ -45,14 +45,13 @@ sub run {
 
     my @available_translations = @{$license_agreement_info->{available_languages}};
     foreach my $translation (@{$test_data->{license}->{translations}}) {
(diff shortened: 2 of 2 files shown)


lib/Installation/License/Sle/LicenseAgreementController.pm
```diff
@@ -15,16 +15,17 @@ use warnings;
 use Installation::License::AcceptLicensePopup;
 use Installation::License::LicenseAgreementExplicitPage;
 use YuiRestClient;
+use Utils::Architectures 'is_s390x';
```
jknphy, line 18: you need to implement other page/controller (perhaps inheriting from current one or not). There are not conditions `if/else` when designing with libyui-rest-api in general.

rakoenig: I understand your point, but somewhere the decision which IDs are used for the widgets need to be used depending on is_s390x(). Where shall we place this? The test module invokes the Controller, so if we build up different Controllers of s390x and non-s390x then we need to put the decision in the test module and adapt more than when we decide to do the switch inside the controller.

jknphy: actually we should adapt less, I mean change existing code, but we might need more new files to do it, that is the benefit of having such a design. If we introduce conditions in the framework then the framework would start to look more like test using testapi and really hard to maintain. Let's have discussion offline about how to design it.

hjluo: I'm thinking if we can use this kind of design:
btn_ok_filter => {id => 'ok_msg|ok'} then we can find each 'id' in the find_widget, once we find one then we can return. if both can't be found then we can error out. this will make it arch-independent.

jknphy: actually that is a good option :) 
you can use regex for those filters, `id => qr/ok_msg|ok/`
it is not perfect solution, because those are different screens and with don't follow the POM strictly...but if you don't need to change anything else, is the case?

lemon-suse (approved):

tests/installation/licensing/verify_license_translations.pm
```diff
@@ -45,14 +45,13 @@ sub run {
 
     my @available_translations = @{$license_agreement_info->{available_languages}};
     foreach my $translation (@{$test_data->{license}->{translations}}) {
-        unless (first { $_ eq $translation->{language} } @available_translations)
-        {
+        my $found_lang = first { /^\Q$translation->{language}\E/ } @available_translations;
```
jknphy, line 48: why this is needed?

hjluo: We'd check if "English (UK)" and  "English (US)" with brackets in the new language combox.

PR #16485, opened 2023-02-28
Schedule trustedprograms only on SP4 and SP5

realcharmer:
The `trustedprograms` test should only be scheduled for SP4 and SP5.

- Related ticket: https://progress.opensuse.org/issues/125135


ilmanzo:
LGTM

PR #16486, opened 2023-02-28
Call the new trento authentication script

hadeskun:
Starting using the new authentication script on trento.pm file, change the name of the script and some arguments.

- Related ticket: CFSA-1726
- Verification run: here the script called https://openqaworker15.qa.suse.cz/tests/131816#step/install_agents/30 and here the output https://openqaworker15.qa.suse.cz/tests/131816#step/install_agents/34 and here the key used from the output in the next command https://openqaworker15.qa.suse.cz/tests/131816#step/install_agents/55

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -768,11 +768,12 @@ Get the api-key from the Trento installation
 
 sub trento_api_key {
     my ($wd, $basedir) = @_;
-    my $cmd = join(' ', $basedir . '/trento-server-api-key.sh',
+    my $cmd = join(' ', $basedir . 'trento_deploy/trento_deploy.py',
+        '--verbose', 'api_key',
         '-u', 'admin',
         '-p', get_trento_password(),
         '-i', get_trento_ip(),
-        '-d', $wd, '-v');
+        '-v');
     my $agent_api_key = '';
     my @lines = split(/\n/, script_output($cmd));
     foreach my $line (@lines) {
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -768,7 +768,7 @@ Get the api-key from the Trento installation
 
 sub trento_api_key {
     my ($wd, $basedir) = @_;
-    my $cmd = join(' ', $basedir . 'trento_deploy/trento_deploy.py',
+    my $cmd = join(' ', $basedir . '/trento_deploy/trento_deploy.py',
         '--verbose', 'api_key',
         '-u', 'admin',
         '-p', get_trento_password(),
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -767,7 +767,7 @@ Get the api-key from the Trento installation
 =cut
 
 sub trento_api_key {
-    my ($wd, $basedir) = @_;
+    my ($basedir) = @_;
     my $cmd = join(' ', $basedir . '/trento_deploy/trento_deploy.py',
         '--verbose', 'api_key',
         '-u', 'admin',
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -772,8 +772,7 @@ sub trento_api_key {
         '--verbose', 'api_key',
         '-u', 'admin',
         '-p', get_trento_password(),
-        '-i', get_trento_ip(),
-        '-v');
+        '-i', get_trento_ip());
     my $agent_api_key = '';
     my @lines = split(/\n/, script_output($cmd));
     foreach my $line (@lines) {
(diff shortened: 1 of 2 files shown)


lib/trento.pm
```diff
@@ -768,11 +768,12 @@ Get the api-key from the Trento installation
 
 sub trento_api_key {
     my ($wd, $basedir) = @_;
-    my $cmd = join(' ', $basedir . '/trento-server-api-key.sh',
+    my $cmd = join(' ', $basedir . 'trento_deploy/trento_deploy.py',
```
alvarocarvajald, line 771: I see in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/sles4sap/trento/install_agents.pm#L19-L23 that this is called with `$basedir = '/root/test'`, so unless we expect a directory called `/root/testtrento_deploy/...` to exist, then the line above is missing a `/` right after `$basedir`.

I was going to check how that's working in the VRs, but they have not been added yet.

lib/trento.pm
```diff
@@ -768,11 +768,12 @@ Get the api-key from the Trento installation
...
+        '--verbose', 'api_key',
         '-u', 'admin',
         '-p', get_trento_password(),
         '-i', get_trento_ip(),
-        '-d', $wd, '-v');
```
alvarocarvajald, line 775: If `$wd` is no longer needed, shouldn't it be removed as an argument?

lib/trento.pm
hadeskun: Good obervation master thx, refactored

lib/trento.pm
hadeskun: Nice comment thx master :)

lib/trento.pm
```diff
@@ -767,12 +767,13 @@ Get the api-key from the Trento installation
...
-    my ($wd, $basedir) = @_;
-    my $cmd = join(' ', $basedir . '/trento-server-api-key.sh',
+    my ($basedir) = @_;
+    my $cmd = join(' ', $basedir . '/trento_deploy/trento_deploy.py',
+        '--verbose', 'api_key',
         '-u', 'admin',
         '-p', get_trento_password(),
         '-i', get_trento_ip(),
-        '-d', $wd, '-v');
+        '-v');
```
mpagot, line 776: I think that this last `-v` is not needed too

lib/trento.pm
```diff
@@ -767,12 +767,13 @@ Get the api-key from the Trento installation
 =cut
 
 sub trento_api_key {
-    my ($wd, $basedir) = @_;
-    my $cmd = join(' ', $basedir . '/trento-server-api-key.sh',
+    my ($basedir) = @_;
```
mpagot, line 770: if you remove `$wd` from here (that is fine) you also need to remove it from the caller https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/25bec175868356be70cde4d3796c2a02fd4f4256/tests/sles4sap/trento/install_agents.pm#L23

mpagot:
You can try to add a unit testing function in *t/08_trento.pm* to help you in changing the code. As reference starting point a test that is passing with the code before your PR

```
subtest '[trento_api_key]' => sub {
    my $trento = Test::MockModule->new('trento', no_auto => 1);
    @calls = ();

    $trento->redefine(script_output => sub {
        push @calls, $_[0];
        return 'GARBAGE
GARBAGE
api_key:SUPERSECRETINGREDIENT';});
    $trento->redefine(get_trento_password => sub { return 'PIZZOCCHERI'; });
    $trento->redefine(get_trento_ip => sub { return 'BITTO'; });

    my $key = trento_api_key('/PEPERONATA', '/RATATUILLE');

    note("\n  C-->  " . join("\n  C-->  ", @calls));

    like $calls[0], qr/\/RATATUILLE\/trento-server-api-key.sh/;
    like $calls[0], qr/\-u admin/;
    like $calls[0], qr/\-p PIZZOCCHERI/;
    like $calls[0], qr/\-i BITTO/;
    is $key, "SUPERSECRETINGREDIENT";
};
```

lib/trento.pm
alvarocarvajald: It should still work, but I agree that it's better to remove it from the caller as well.

lib/trento.pm
hadeskun: thx for the comment master, will remove the argument.

mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #16494, opened 2023-02-28
Stream raw.xz images directly to the SD card

jlausuch:


asdil12:
I refactored the script to stream everything as I discovered `qemu-img dd`: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16495

PR #16495, opened 2023-02-28
RPi: Stream all formats when flashing sdcard

asdil12:
This will allow larger images that would exceed worker RAM.

### DIFF
--- a/data/generalhw_scripts/flash_sd_rootless.sh
+++ b/data/generalhw_scripts/flash_sd_rootless.sh
@@ -9,54 +9,39 @@ echo "Flash script start...";
 
 # Check number of args
 if [ "$#" -ne 4 ]; then
-    echo "Please provide <destination>, <device serial>, <image to flash> and <hdd size> (ignored)."
+    echo "Please provide <device serial>, <image to flash> and <hdd size> (ignored)."
     exit 1;
 fi
 
-# Get destination: /storage/
-destination_folder=$1
 # Get device serial (check 'sd-mux-ctrl --list', or usb-sd-mux/id-<SERIAL>)
-device_serial=$2
+device_serial=$1
 # Get image to flash
-image_to_flash=$3
+image_to_flash=$2
 # Get hdd size (ignored for SD, but passed as arg from openQA)
-hdd_size=$4
+hdd_size=$3
 
 echo "* Switch SD card to flasher"
 usbsdmux /dev/usb-sd-mux/id-$device_serial host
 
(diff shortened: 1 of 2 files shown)


data/generalhw_scripts/flash_sd_rootless.sh
```diff
@@ -25,38 +25,25 @@ hdd_size=$4
...
-	dd if=$uncompressed_full_path of=/dev/disk/by-id/usb-LinuxAut_sdmux_HS-SD_MMC_${device_serial}-0:0 oflag=sync bs=8M
+	image_to_flash_extension="${image_to_flash##*.}"
+	if [ "$image_to_flash_extension" == "qcow2" ] ; then
+		qemu-img dd -f qcow2 -O raw if=$image_to_flash of=$device_link bs=8M
+	elif [ "$image_to_flash_extension" == "xz" ] ; then
+		xzcat --threads=0 $image_to_flash | dd of=$device_link oflag=sync bs=8M
+	elif [ "$image_to_flash_extension" == "gz" ] ; then
+		zcat $image_to_flash | dd of=$device_link oflag=sync bs=8M
+	else
+		cat $image_to_flash | dd of=$device_link oflag=sync bs=8M
```
ggardet, line 45: Is there an advantage/disadvantage over  `dd if=$image_to_flash of=$device_link oflag=sync bs=8M` ?

data/generalhw_scripts/flash_sd_rootless.sh
```diff
@@ -13,7 +13,7 @@ if [ "$#" -ne 4 ]; then
     exit 1;
 fi
 
-# Get destination: /storage/
+# Get destination (tmpdir to extract image): /storage/
 destination_folder=$1
```
ggardet, line 17: `destination_folder` is not used anymore apparently, and thus, could be removed.

ggardet:
> This will allow larger images that would exceed worker RAM.

That would exceed worker **disk space**, not RAM

asdil12:
> > This will allow larger images that would exceed worker RAM.
> 
> That would exceed worker **disk space**, not RAM

On my setup the destination dir is `/dev/shm/…` to save sdcard write cycles - so that's why I run out of RAM.
But yes it will also help with disk space.

data/generalhw_scripts/flash_sd_rootless.sh
asdil12: Right. Updated script and docs.

PR #16496, opened 2023-03-01
Catch regression when sending right key when switching to tab Bootloader Options in ncurses

tinawang123:
Catch regression when sending right key when switching to tab Bootloader Options in ncurses

- Related ticket: https://progress.opensuse.org/issues/124610
- Verification run: https://openqa.suse.de/tests/11038962#step/disable_boot_menu_timeout_navigating_tabs/3
https://openqa.suse.de/tests/11041255#step/disable_boot_menu_timeout_navigating_tabs/3

### DIFF
--- a/lib/YaST/Bootloader/BootloaderOptionsTab.pm
+++ b/lib/YaST/Bootloader/BootloaderOptionsTab.pm
@@ -1,0 +1,26 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for Bootloader Options tab
+# in Boot Loader Settings.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package YaST::Bootloader::BootloaderOptionsTab;
+use strict;
+use warnings;
+use testapi;
+
+sub new {
+    my $self = shift;
+    return $self;
+}
+
+sub navigate {
+    my ($self) = @_;
+    send_key 'tab' unless match_has_tag 'inst-bootloader-settings-first_tab_highlighted';
+    send_key_until_needlematch 'inst-bootloader-options-highlighted', 'right';
+}
+
+1;
--- a/lib/YaST/Bootloader/BootloaderSettingsController.pm
+++ b/lib/YaST/Bootloader/BootloaderSettingsController.pm
@@ -13,6 +13,7 @@ use YuiRestClient;
 use YaST::Bootloader::BootCodeOptionsPage;
 use YaST::Bootloader::KernelParametersPage;
 use YaST::Bootloader::BootloaderOptionsPage;
+use YaST::Bootloader::BootloaderOptionsTab;
 
 sub new {
     my ($class, $args) = @_;
@@ -25,6 +26,7 @@ sub init {
     $self->{BootCodeOptionsPage} = YaST::Bootloader::BootCodeOptionsPage->new({app => YuiRestClient::get_app()});
     $self->{KernelParametersPage} = YaST::Bootloader::KernelParametersPage->new({app => YuiRestClient::get_app()});
     $self->{BootloaderOptionsPage} = YaST::Bootloader::BootloaderOptionsPage->new({app => YuiRestClient::get_app()});
+    $self->{BootloaderOptionsTab} = YaST::Bootloader::BootloaderOptionsTab->new({});
     return $self;
 }
 
@@ -44,6 +46,11 @@ sub get_bootloader_options_page {
     my ($self) = @_;
     die 'Bootloader Options tab is not shown' unless $self->{BootloaderOptionsPage}->is_shown();
     return $self->{BootloaderOptionsPage};
+}
+
+sub get_bootloader_options_tab {
+    my ($self) = @_;
+    return $self->{BootloaderOptionsTab};
 }
 
 sub get_current_settings {
@@ -77,6 +84,14 @@ sub write_to_partition {
     my ($self) = @_;
     $self->get_boot_code_options_page->check_write_to_partition();
     $self->get_boot_code_options_page->press_next();
+}
+
+sub disable_grub_timeout_navigating_tabs {
+    my ($self) = @_;
+    $self->get_boot_code_options_page;
+    $self->get_bootloader_options_tab->navigate();
+    $self->get_bootloader_options_page->set_grub_timeout('-1');
+    $self->get_bootloader_options_page->press_next();
 }
 
 sub disable_grub_timeout {
--- a/tests/installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs.pm
+++ b/tests/installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs.pm
@@ -1,0 +1,18 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Disable grub timeout from the Installer to catch regression of bsc#1208266
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    $testapi::distri->get_installation_settings()->access_booting_options();
+    $testapi::distri->get_bootloader_settings()->disable_grub_timeout_navigating_tabs();
+}
+
+1;
--- a/schedule/staging/textmode_install@64bit-staging.yaml
+++ b/schedule/staging/textmode_install@64bit-staging.yaml
@@ -19,7 +19,7 @@ schedule:
   - installation/clock_and_timezone/accept_timezone_configuration
   - installation/authentication/use_same_password_for_root
   - installation/authentication/default_user_simple_pwd
-  - installation/bootloader_settings/disable_boot_menu_timeout
+  - installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs
   - installation/launch_installation
   - installation/confirm_installation
   - installation/performing_installation/perform_installation
--- a/schedule/yast/textmode_installation_minimal_role/textmode_installation_minimal_role_x86_64.yaml
+++ b/schedule/yast/textmode_installation_minimal_role/textmode_installation_minimal_role_x86_64.yaml
@@ -15,6 +15,8 @@ schedule:
     - installation/add_on_product/skip_install_addons
   system_role:
     - installation/system_role/select_role_minimal
+  booting:
+    - installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs
   default_systemd_target:
     - installation/installation_settings/validate_default_target
   system_preparation:


tests/installation/bootloader_settings/disable_boot_menu_timeout_testapi.pm
```diff
@@ -0,0 +1,20 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
```
okurz, line 3:
```suggestion
# Copyright SUSE LLC
```

See our contribution hints for generic remarks regarding copyright

schedule/yast/textmode/ha_textmode_skip_registration_bsc1208266.yaml
```diff
@@ -0,0 +1,28 @@
+name: ha_textmode_skip_registration
```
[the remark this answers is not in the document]
tinawang123: For case ' ha_textmode_skip_registration ', it doesn't have default yaml file at x86_64, I need create a new one for aarch64. As I have ticket 'https://progress.opensuse.org/issues/124035' to do this task. I'd like to create default yaml file at other PR instead of this one.

tinawang123:
> Please, prepare also the staging test, we should reuse the textmode test suite that we recently added (coordinate with @rakoenig in https://progress.opensuse.org/issues/124676) Once addressed feedback and verified, we will merge the PR and resolve the ticket without enabling it in job groups for staging. We should file a ticket for the enablement (but test here that all works as expected) as the bug is still not fixed (and we are not going to block staging).

I used one staging case to check this problem: https://openqa.suse.de/tests/10658338
Only the case which uses 'disable_boot_menu_timeout_right_key' can catch the problem, others cases will not be impacted even this PR merged.

jknphy:
> > Please, prepare also the staging test, we should reuse the textmode test suite that we recently added (coordinate with @rakoenig in https://progress.opensuse.org/issues/124676) Once addressed feedback and verified, we will merge the PR and resolve the ticket without enabling it in job groups for staging. We should file a ticket for the enablement (but test here that all works as expected) as the bug is still not fixed (and we are not going to block staging).
> 
> I used one staging case to check this problem: https://openqa.suse.de/tests/10658338 Only the case which uses 'disable_boot_menu_timeout_right_key' can catch the problem, others cases will not be impacted even this PR merged.

could you please file an additional WIP PR with the change on schedule for staging, so we remember to enable? (no ticket needed I guess then).

schedule/staging/textmode_install_bsc1208266@64bit-staging.yaml
```diff
@@ -0,0 +1,29 @@
+---
```
jknphy, line 1: we don't need a new yaml schedule, we have already a textmode test suite in staging, please use that one.

tests/installation/bootloader_settings/disable_boot_menu_timeout.pm
```diff
@@ -12,6 +12,7 @@ use base 'y2_installbase';
 
 sub run {
     $testapi::distri->get_installation_settings()->access_booting_options();
+    $testapi::distri->get_bootloader_settings()->switch_to_bootloader_options_navigating_tabs();
```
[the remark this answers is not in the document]
tinawang123: Original code, tab to bootloader page and set grub timeout value in the same function 'sub disable_grub_timeout'.
I split this one function to two functions, one for switch to bootloader page, one for set grub timeout.
If we want to use right key to get bootloader page, we can reuse set grub timeout function.

schedule/staging/textmode_install_bsc1208266@64bit-staging.yaml
tinawang123: If we use this yaml file, the staging case will be failed as bug has not fix yet.

jknphy: > If we use this yaml file, the staging case will be failed as bug has not fix yet.

because we don't plan to merge the PR until the bug is fixed, it is just preparation.

tests/installation/bootloader_settings/disable_boot_menu_timeout_right_key.pm
```diff
@@ -0,0 +1,20 @@
...
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use testapi;
+
+sub run {
+    $testapi::distri->get_installation_settings()->access_booting_options();
+    send_key_until_needlematch 'inst-bootloader-options-highlighted', 'right';
```
jknphy, line 16: for this case you need to wrap this in a method and put testapi in Controller, it is the exception of course.

tests/installation/bootloader_settings/disable_boot_menu_timeout.pm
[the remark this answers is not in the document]
jknphy: then we need tow method:
`switch_to_bootloader_options` (doesn't need any prefix, because it is the only way to navigate with libyui-rest-api, going directly).
`switch_to_bootloader_options_via_tabs`

tests/installation/bootloader_settings/disable_boot_menu_timeout_right_key.pm
```diff
@@ -0,0 +1,19 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Disable grub timeout from the Installer
```
jknphy, line 6: I don't think nobody would understand what you mean with `_right_key` name, I would replace everywhere by `_navigating_tabs`. Do you want to test it with someone? wdyt?

tests/installation/bootloader_settings/disable_boot_menu_timeout_right_key.pm
```diff
@@ -0,0 +1,19 @@
...
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use testapi;
```
jknphy, line 12: you don't need to import the library with using `$testapi`.

jknphy:
> Catch regression when sending right key when switching to tab Bootloader Options in ncurses
> 
> * Related ticket: https://progress.opensuse.org/issues/124610
> * Verification run: https://openqa.suse.de/tests/10912241#step/disable_boot_menu_timeout_right_key/1

and verification for textmode installation with minimal role in YaST job group ?

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -79,6 +80,15 @@ sub write_to_partition {
     $self->get_boot_code_options_page->press_next();
 }
 
+sub disable_grub_timeout_right_key {
```
jknphy, line 83: the existence of this function using testapi worth a mention to the bug and why is there in the form of a comment? wdyt?

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -10,9 +10,11 @@ package YaST::Bootloader::BootloaderSettingsController;
 use strict;
 use warnings;
 use YuiRestClient;
+use testapi;
 use YaST::Bootloader::BootCodeOptionsPage;
 use YaST::Bootloader::KernelParametersPage;
 use YaST::Bootloader::BootloaderOptionsPage;
+use YaST::workarounds;
```
jknphy, line 17: it is not a workaround, the bug is fixed, it is just another testing path that we should cover to avoid regression, but we just need to mention that in a comment in the method.

tinawang123:
Any more comment?

tests/installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs.pm
```diff
@@ -0,0 +1,18 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Disable grub timeout from the Installer accoding bsc#1208266
```
jknphy, line 6: s/accoding/according/

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -79,6 +88,13 @@ sub write_to_partition {
     $self->get_boot_code_options_page->press_next();
 }
 
+sub disable_grub_timeout_navigating_tabs {
+    my ($self) = @_;
+    $self->get_bootloader_options_Tab->use_navigating_tabs();
```
jknphy, line 93: your POM should represent some part of the screen, not just one Tab, plus should be in lower letter.
IMO the getter should be: `get_bootloader_settings_tabs()` (please, do not forget the `()` for each function to follow the current coding style).
And the function will be just `navigate()`,
so `$self->get_bootloader_settings_tab->navigate()`.

lib/YaST/Bootloader/BootloaderOptionsTab.pm
```diff
@@ -0,0 +1,34 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The class introduces all accessing methods for Bootloader Options tab
```
jknphy, line 6: "bootloader options tab" is jut the one tab more on the right, it is BootloaderSettingsTabs the object to map.

lib/YaST/Bootloader/BootloaderOptionsTab.pm
```diff
@@ -0,0 +1,34 @@
...
+
+package YaST::Bootloader::BootloaderOptionsTab;
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+use testapi;
+
+sub init {
+    my $self = shift;
+    $self->SUPER::init();
```
jknphy, line 18: this class doesn't need to container NavigationBase or any libyui-rest-api,
to check that you are in the right screen, you just need existing method:
`get_boot_code_options_page`. I guess in the current design we would need another controller which would include `BootCodeOptionsPage` in similar way that `BootloaderSettingsController` does plus the needling code below.

tinawang123: I need this code: $self->{cmb_bootloader} = $self->{app}->combobox({id => "\"Bootloader::LoaderTypeWidget\""});
To make sure the page: https://openqa.suse.de/tests/10917817#step/disable_boot_menu_timeout_navigating_tabs/1 is shown
After this, I can use right key to get 'bootloader options' page

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -79,6 +81,14 @@ sub write_to_partition {
     $self->get_boot_code_options_page->press_next();
 }
 
+sub disable_grub_timeout_navigating_tabs {
+    my ($self) = @_;
+    $self->get_boot_code_options_page->is_shown();
+    YaST::Bootloader::BootloaderOptionsTab->navigate();
```
jknphy, line 87: you need a getter, do not access directly to object

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -79,6 +81,14 @@ sub write_to_partition {
     $self->get_boot_code_options_page->press_next();
 }
 
+sub disable_grub_timeout_navigating_tabs {
+    my ($self) = @_;
+    $self->get_boot_code_options_page->is_shown();
```
jknphy, line 86: I guess just `$self->get_boot_code_options_page()` you don't need to run `is_shown()` twice.

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -10,9 +10,11 @@ package YaST::Bootloader::BootloaderSettingsController;
 use strict;
 use warnings;
 use YuiRestClient;
+use testapi;
```
jknphy, line 13: now you can drop it.

lib/YaST/Bootloader/BootloaderOptionsTab.pm
jknphy: it should have the name suggested above.
you are missing to create some `new()` method and you can drop the `init()`

tests/installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs.pm
```diff
@@ -0,0 +1,18 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Disable grub timeout from the Installer according bsc#1208266
```
jknphy, line 6: instead of `according` perhaps would be more clear to phrase it like `to catch regression of`

tinawang123:
Any more comments?

lib/YaST/Bootloader/BootloaderOptionsTab.pm
```diff
@@ -0,0 +1,26 @@
...
+use warnings;
+use testapi;
+
+sub new {
+    my $self = shift;
+    return $self;
+}
+
+sub navigate {
+    my ($self) = @_;
```
jknphy, line 21: not needed.

lib/YaST/Bootloader/BootloaderSettingsController.pm
```diff
@@ -79,6 +86,14 @@ sub write_to_partition {
     $self->get_boot_code_options_page->press_next();
 }
 
+sub disable_grub_timeout_navigating_tabs {
+    my ($self) = @_;
+    $self->get_boot_code_options_page;
```
jknphy, line 91: `$self->get_boot_code_options_page()`

PR #16497, opened 2023-03-01
Reduce execution of compile tests by checking changed files only

rfan1:
- Related ticket: https://progress.opensuse.org/issues/124688

- Verification run: n/a
Now only changed files are checked
`export PERL5LIB=../..:os-autoinst:lib:tests/installation:tests/x11:tests/qa_automation:tests/virt_autotest:tests/cpu_bugs:tests/sles4sap/saptune:$PERL5LIB: ; for f in `git diff --name-only | grep '.pm'` ; do perl -c $f 2>&1 | grep -v " OK$" && exit 2; done ; true`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,8 @@ jobs:
             perl-version: 5.32
           - test: unit
             perl-version: 5.32
+          - test: compile-changed
+            perl-version: 5.26
           - test: compile
             perl-version: 5.26
     container:
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,8 @@ test-static: tidy-check test-yaml-valid test-modules-in-yaml-schedule test-merge
 .PHONY: test
 ifeq ($(TESTS),compile)
 test: test-compile
+else ifeq ($(TESTS),compile-changed)
+test: test-compile-changed
 else ifeq ($(TESTS),static)
 test: test-static
 else ifeq ($(TESTS),unit)


grisu48:
I'd argue that the test should check all files to prevent that we're accidentally missing out some files and thus this should not be changed. Isn't also the CI using this test? If so, this change might reduce the overall coverage and introduce the possibility of missing out some files, that should be checked.

`test-compile-changed` is an additional convenience function to be executed locally.

b10n1k (changes requested):
Should also this line be changed? -> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/bebe56bb31c8baa9f4ee7e71c9adb105ac7d399b/Makefile#L118 ?

foursixnine:
> I'd argue that the test should check all files to prevent that we're accidentally missing out some files and thus this should not be changed. Isn't also the CI using this test? If so, this change might reduce the overall coverage and introduce the possibility of missing out some files, that should be checked.
> 
> IMHO `test-compile-changed` is an additional convenience function to be executed locally only.

We can do this for PRs and keep the longer running one for the merges

okurz:
> We can do this for PRs and keep the longer running one for the merges

This introduces the risk that failures in master will not be attended for long but I think that's an acceptable risk

rfan1:
> Should also this line be changed? ->
> 
> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/bebe56bb31c8baa9f4ee7e71c9adb105ac7d399b/Makefile#L118
> ?

done

okurz (changes requested):
Wait, this looks wrong now. Please keep the `make test` rule as the complete one as is. Now you are making the use of "…-changed" inconsistent with other rules. If you want to only test changed files in pull requests do that explicitly but the default should be that everything is tested completely. Your change is *very* dangerous as *by default* only changed files would be compiled *but also* on merge to master not all files are compile tested.

rfan1:
Thanks all for the kindly feedback.

I will hold on this PR due to some risks

okurz:
You can just follow the suggestion in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16497#issuecomment-1449668278 to have the "changed" target called in GitHub actions for pull requests, the complete one otherwise. I suggest to look into the GitHub workflow file

rfan1:
> You can just follow the suggestion in [#16497 (comment)](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16497#issuecomment-1449668278) to have the "changed" target called in GitHub actions for pull requests, the complete one otherwise. I suggest to look into the GitHub workflow file

Sure, will do

rfan1:
@foursixnine Seems it is hard to define the different job for ci check for both push and pull_request, so I create a new file

rfan1:
Based on the CI check result, comile tests  at pull_request phase is `Required`
that we can skip it.
 ci / CI: Running compile tests with perl v5.26 (pull_request) Successful in 14m **Required**

We may need change the default required workflow [I don't have the permission to do this]
 https://docs.github.com/en/actions/using-workflows/required-workflows

michaelgrifalconi:
Hello @okurz , trying to resume this topic to finally close it. I see it was agreed to have both compile and compile-changed running, where only compile-changed mandatory (while now the full compile is).
Is this still fine?

If yes, I would propose to get this PR merged and then have someone with admin permission in this repo to remove 'compile' test from the Required ones to merge PRs.

okurz (approved):

michaelgrifalconi:
For the record, the force push was with no changes, to trigger github checks again, which were stuck.

PR #16500, opened 2023-03-01
Wait until time synced before listing NTP sources

ilmanzo:
NTP servers are chosen randomly in a pool, so sometimes at the start of the service, they can't be all reached. As the NTP sync should occur nevertheless, we wait for it before enumerating all the NTP sources.

- Related ticket: https://progress.opensuse.org/issues/124496
- Verification runs: 
    - https://openqa.suse.de/tests/10565433              
    - https://openqa.suse.de/tests/10576693              
    - https://openqa.suse.de/tests/10576695              
    - https://openqa.suse.de/tests/10599890


tjyrinki (approved):
Wow, that's great!

PR #16501, opened 2023-03-01
Solve the installation failue on disable_grub_graphics stage when SUT installation

museQi:
To solve the installation failue  on disable_grub_graphics stage when SUT installtion.It needs to check whether have the cover.
http://10.67.129.4/tests/56463
image

Verification run: [](http://10.67.129.4/tests/56464#[](url))

### DIFF
--- a/tests/installation/installation_overview.pm
+++ b/tests/installation/installation_overview.pm
@@ -119,6 +119,9 @@ sub run {
         # select None for Major Linux Security Module.
         set_linux_security_to_none if (is_sle('>=15-SP4') && check_screen("apparmor-not-selected") && !(get_var('PATTERNS') =~ 'default|all|apparmor'));
         ensure_ssh_unblocked;
+        until (!check_screen("install-overview-options-evaluating-pkg-selection")) {
+            save_screenshot;
+        }
         $self->check_default_target();
     }
 }


museQi:
@XGWang0 @zyuhu please help me review.

okurz (changes requested):

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

You might want to use the more up-to-date tool https://github.com/cli/cli

Please also delete the dot at the end of the git subject lines. I suggest you take a look into the git log for existing git commits to find good examples as reference.

Also, you are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

museQi:
@JoyceNa @Dawei-Pang Please help me review.

museQi:
@okurz Thanks for your advice.I have changed.

tests/installation/installation_overview.pm
```diff
@@ -119,8 +119,15 @@ sub run {
         # select None for Major Linux Security Module.
         set_linux_security_to_none if (is_sle('>=15-SP4') && check_screen("apparmor-not-selected") && !(get_var('PATTERNS') =~ 'default|all|apparmor'));
         ensure_ssh_unblocked;
-        $self->check_default_target();
+        wait_still_screen;
```
okurz, line 122: You are using `sleep()` or are simulating a sleep-like functionality with wait_still_screen which is effectively a "sleep at least 7s". This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

Please consider that this test module is used in a lot of tests and hence extra care should be taken to not introduce unnecessary delays in execution

museQi: Yes,you are right. Please help me review again.

museQi:
@okurz Please help me review again,thanks.

okurz (changes requested):

I suggest to also look into previous commits in this repository as reference. Also there should not be a dot at the end of line of the git commit message subject line.

tests/installation/installation_overview.pm
```diff
@@ -119,8 +119,11 @@ sub run {
         # select None for Major Linux Security Module.
         set_linux_security_to_none if (is_sle('>=15-SP4') && check_screen("apparmor-not-selected") && !(get_var('PATTERNS') =~ 'default|all|apparmor'));
         ensure_ssh_unblocked;
-        $self->check_default_target();
+        until (!check_screen("install-overview-options-evaluating-pkg-selection")) {
+            save_screenshot;
+        }
     }
+    $self->check_default_target();
```
okurz, line 126: I don't understand what you are trying to achieve but I am pretty sure this will break a lot of tests. I suggest you consult with the test module maintainers and also I suggest to run an extensive set of test scenarios. Without a timeout this will run into MAX_JOB_TIME if the needle tag install-overview-options-evaluating-pkg-selection is never matched. Also because `check_screen` checks once without waiting time this will be a busy-loop occupying the system and wasting CPU ressources until a match is found so also discouraged.

museQi: Hello,Oliver. Because we encountered some machines' job failure during installation.The time of evaluate packages in the installation overview phase is relatively long, which will lead to failure, so these two lines are added. See details:http://10.67.129.4/tests/56464#

museQi: I updated the commit comments.

okurz: I think you confused the github description with git commit message. The git commit message is still just one line

museQi: Thank you for giving me so much advice and knowledge. I will study the knowledge you shared about git again.
Please help me check again. I am sorry that this PR wasted your so much time.

museQi:
Only one machine always occur ,So decide to solve it locally.

PR #16502, opened 2023-03-01
[virtualization] ALP host preparations for VT test

alice-suse:
This PR mainly does 4 things in 4 commits:
    a) download vm imported disks
    b) alp host other preparations and checks to let VT test continue
    c) enhance virtual network cleanup function to survive non-persistent networks cleaning
    d) enhance unified_guest_installation to cleanup potential affecting vm disks when no vm exists but vm disks exist

- Related ticket: https://progress.opensuse.org/issues/123154

- Verification run:

    Fail on purpose job with wrong vm disk location:
    http://10.67.129.185/tests/534#step/setup_kvm_container/138

    PASS:   
    alp: http://10.67.129.185/tests/537#
    osd regression test:   https://openqa.suse.de/t10614820 , and https://openqa.suse.de/t10614891

### DIFF
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -30,7 +30,8 @@ our @EXPORT = qw(is_vmware_virtualization is_hyperv_virtualization is_fv_guest i
   print_cmd_output_to_file ssh_setup ssh_copy_id create_guest import_guest install_default_packages upload_y2logs ensure_default_net_is_active ensure_guest_started
   ensure_online add_guest_to_hosts restart_libvirtd remove_additional_disks remove_additional_nic collect_virt_system_logs shutdown_guests wait_guest_online start_guests restore_downloaded_guests save_original_guest_xmls restore_original_guests
   is_guest_online wait_guests_shutdown remove_vm setup_common_ssh_config add_alias_in_ssh_config parse_subnet_address_ipv4 backup_file manage_system_service setup_rsyslog_host
-  check_port_state subscribe_extensions_and_modules download_script download_script_and_execute is_sev_es_guest upload_virt_logs recreate_guests);
+  check_port_state subscribe_extensions_and_modules download_script download_script_and_execute is_sev_es_guest upload_virt_logs recreate_guests
+  download_vm_import_disks);
 
 # helper function: Trim string
 sub trim {
@@ -856,5 +857,44 @@ sub recreate_guests {
         script_run("virsh start $_");
     }
 }
+
+# For vms with imported disks, this function can be used to download the disk from ${OPENQA_URL}/assets/ to SUT.
+sub download_vm_import_disks {
+    my $download_dir = shift;
+
+    $download_dir //= "/var/lib/libvirt/images";
+    my $disks_to_download = get_var("VM_IMPORT_DISK_CONFIG", '');
+
+    if (!$disks_to_download) {
+        record_info "No import disk configured, skip download.";
+        return;
+    } else {
+        record_info "Going to download imported disk...";
+    }
+
+    assert_script_run("mkdir -p $download_dir");
+
+    my $BASE_URL = get_required_var("OPENQA_URL") . "/assets/";
+    foreach my $disk (split(/,/, $disks_to_download)) {
+        my $download_url = $BASE_URL . $disk;
+        die "URL is not accessible: $download_url." unless head($download_url);
+
+        $disk =~ /.*\/([^\/]+)\.([^\/\.]+)$/m;
+        my $output_name = "$1-back.$2";    # example: name.qcow2 => name-back.qcow2
+        my $cmd = "curl -L $download_url -o $download_dir/$output_name";
+        script_retry($cmd, retry => 2, delay => 5, timeout => 600, die => 1);
+        save_screenshot;
+
+        # Check if the downloaded image is good.
+        # `qemu-img check` is fast even for several GB disks.
+        # For download from corect URL, it can detect qcow2/raw format disk errors,
+        assert_script_run("qemu-img check $download_dir/$output_name");
+        save_screenshot;
+        record_info("Disk downloaded successfully for $download_url.");
+    }
+    assert_script_run("ls -latr $download_dir");
+    record_info("All imported disk download is done.");
+}
+
 
 1;
--- a/lib/virt_autotest/virtual_network_utils.pm
+++ b/lib/virt_autotest/virtual_network_utils.pm
@@ -357,10 +357,12 @@ sub clean_all_virt_networks {
     foreach my $vnet (split(/\n+/, $_virt_networks)) {
         my $_br = script_output(q@virsh net-dumpxml @ . $vnet . q@|grep -o "bridge name=[^\s]*" | sed  's#bridge name=##'@, type_command => 0, proceed_on_failure => 0);
         script_run("virsh net-destroy $vnet");
-        assert_script_run("virsh net-undefine $vnet");
+        script_run("virsh net-undefine $vnet");
         assert_script_run("if ip a|grep $_br;then ip link del $_br;fi");
         save_screenshot;
     }
+
+    die "Virtual networks are not fully cleaned!" if (script_output("virsh net-list --name --all"));
     record_info("All existing virtual networks: \n$_virt_networks \nhave been destroy and undefined.", script_output("ip a; ip route show all"));
 }
 
--- a/tests/virt_autotest/prepare_alp_host.pm
+++ b/tests/virt_autotest/prepare_alp_host.pm
@@ -1,0 +1,47 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: This file does the necessary setup and checks for alp host to do virtualization test.
+# Maintainer: alice <xlai@suse.com>, qe-virt@suse.de
+
+package prepare_alp_host;
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+use utils;
+
+sub run {
+    my $self = shift;
+
+    # Validate that kernel-default is installed and kvm module is loaded.
+    # Note: the alp host unattended installation will include kernel-default
+    #       package installation. See poo#118939 for more details.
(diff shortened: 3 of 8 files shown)


alice-suse:
This PR mainly does 4 things in 4 commits:
- poo#123154
  a) download vm imported disks
  b) alp host preparations and checks to let VT test continue
- minor enhacements
  c) enhance virtual network cleanup function to survive non-persistent networks cleaning
  d) enhance unified_guest_installation to cleanup potential affecting vm disks when no vm exists but vm disks exist

- Related ticket: https://progress.opensuse.org/issues/123154

- Verification run:
```
Fail on purpose job with wrong vm disk location:
- http://10.67.129.185/tests/534#step/setup_kvm_container/138

PASS?:   wip
- alp: http://10.67.129.185/tests/537#
- osd regression test: https://openqa.suse.de/t10614820
```

tests/virt_autotest/prepare_alp_host.pm
```diff
@@ -0,0 +1,50 @@
...
+sub run {
+    my $self = shift;
+
+    # Validate that kernel-default is installed and kvm module is loaded.
+    # Note: the alp host unattended installation will include kernel-default
+    #       package installation. See poo#118939 for more details.
+    assert_script_run(q@rpm -qa | grep kernel-default-[0-9]@);
+    save_screenshot;
+    assert_script_run(q@lsmod |grep kvm@);
+    save_screenshot;
```
Julie-CAO, line 25: No more comments about the entire PR, just duplicate code statements can be removed, such as save_screenshot() after assert_script_run() at here. assert_script_run() takes screenshot automatically, so save_screenshot() is redundent.

alice-suse: From testapi.pm, I do not see direct save_screen, maybe hiding deeper in distribution code. Anyway, if the screenshot is captured there, the save_screenshot can be saved.

lib/virt_autotest/utils.pm
```diff
@@ -857,4 +858,43 @@ sub recreate_guests {
...
+    }
+
+    assert_script_run("mkdir -p $download_dir");
+
+    my $BASE_URL = get_required_var("OPENQA_URL") . "/assets/";
+    foreach my $disk (split(/,/, $disks_to_download)) {
+        my $download_url = $BASE_URL . $disk;
+        die "URL is not accessible: $download_url." unless head($download_url);
+
+        $disk =~ /.*\/([^\/]+)\.([^\/\.]+)$/m;
```
waynechen55, line 882: If $disk is just file name itself, the following is enough:
```
$disk =~ /^(.*)\.(.*)$/m;
```

alice-suse: It looks like hdd/xxx.qcow2, or hdd/fixed/xx.xx.

lib/virt_autotest/virtual_network_utils.pm
```diff
@@ -357,10 +357,12 @@ sub clean_all_virt_networks {
     foreach my $vnet (split(/\n+/, $_virt_networks)) {
         my $_br = script_output(q@virsh net-dumpxml @ . $vnet . q@|grep -o "bridge name=[^\s]*" | sed  's#bridge name=##'@, type_command => 0, proceed_on_failure => 0);
         script_run("virsh net-destroy $vnet");
-        assert_script_run("virsh net-undefine $vnet");
+        script_run("virsh net-undefine $vnet");
         assert_script_run("if ip a|grep $_br;then ip link del $_br;fi");
         save_screenshot;
     }
+
+    die "Virtual networks are not fully cleaned!" if (script_output("virsh net-list --name --all"));
```
waynechen55, line 365: Necessary to remove "default" as well ? I do not think so.

tests/virt_autotest/prepare_alp_host.pm
```diff
@@ -0,0 +1,50 @@
...
+    record_info("KVM module is successfully loaded.");
+
+    # Download alp vm ignition config files
+    assert_script_run("mkdir -p /var/lib/libvirt/images");
+    my $cmd = "curl -L "
+      . data_url("virt_autotest/guest_unattended_installation_files/VIRT_TEST_VM_config.ign")
+      . " -o /var/lib/libvirt/images/VIRT_TEST_VM_config.ign";
+    script_retry($cmd, retry => 2, delay => 5, timeout => 60, die => 1);
+    save_screenshot;
+    assert_script_run("ls -latr /var/lib/libvirt/images/");
```
waynechen55, line 35: You should list desire file out instead of directory I think

lib/virt_autotest/virtual_network_utils.pm
alice-suse: This function is "sub clean_all_virt_netoworks".

tests/virt_autotest/prepare_alp_host.pm
alice-suse: Same effect I think. If script_retry passes, it already actually ensures the target file is there. ALP test currently starts from existing host, which can have existing data in this dir. So listing it can give additional info here .

lib/virt_autotest/virtual_network_utils.pm
waynechen55: I think you'd better to keep "default". You can remove all the others. If any following tests to run after this and they use "default", there is nearly no way to re-create it.

tests/virt_autotest/prepare_alp_host.pm
waynechen55: Listing out the folder can really give info about, but it does not guarantee what you want because it does not assert on it at all. You may think creating directory succeeded, downloading succeeded, then the file must exist, but what you do here does not match your thinking IMHO. If you only need show info, use record_info, if you want to make sure the file is there, then list that file out.  Listing out the folder only does not look like anything, it just assert the folder existence.

lib/virt_autotest/virtual_network_utils.pm
waynechen55: Additionally it looks like deleting br0 on host, although deleting "default" would not lead to failures directly, but it would be better to keep it. If "default" is gone, there is no way to know what it is afterwards, either from its name or other things. The subroutine name still apply, no need to delete "default", the built-in one.

tests/virt_autotest/prepare_alp_host.pm
alice-suse: Is it really worth the discussion effort on this minor thing?
Here the `assert_script_run("ls -latr /var/lib/libvirt/images/")` is to show the files under this dir. I care about the assert and the saved screenshot for it. Yes, record_info can achieve this, but this line does it too I think. To do it with record_info, one has to call script_output in addition to record_info, so simpler just use assert_script_run.

alice-suse: The default network on alp can be seen in http://10.67.129.185/tests/537#step/setup_kvm_container/147. So you can see, on baremetal alp host, no default bridge br0 at all. Within libvirtd container, the "default" network is sle_network, a nat based network which has nothing to do with br0 either. Even if we create br0 during host preparation but before libvirtd container is created, the br0 is not hurt at all. 
This function cleans VIRTUAL networks, rather than physical networks. So the physical networks on host are still there.
Removing the "default" virtual network won't hurt the libvirtd container connectivity either, and it will exclude any potential impact to some of our hard-coded existing code from sles. So that's why in previous PRs, this solution was taken. This logic is not introduced in this PR.
I am open to change the solution, as long as there are good reasons. But not convinced by above comments. It will need a separate ticket to handle all possible work for it to not break any existing alp test.

lib/virt_autotest/virtual_network_utils.pm
alice-suse: The default network on alp can be seen in http://10.67.129.185/tests/537#step/setup_kvm_container/147. So you can see, on baremetal alp host, no default bridge br0 at all. Within libvirtd container, the "default" network is sle_network, a nat based network which has nothing to do with br0 either. Even if we create br0 during host preparation but before libvirtd container is created, the br0 is not hurt at all. 
This function cleans VIRTUAL networks, rather than physical networks. So the physical networks on host are still there.
Removing the "default" virtual network won't hurt the libvirtd container connectivity either, and it will exclude any potential impact to some of our hard-coded existing code from sles. So that's why in previous PRs, this solution was taken. This logic is not introduced in this PR.
I am open to change the solution, as long as there are good reasons, but not convinced by above comment, sorry.  It will need a separate ticket to handle all possible work for it to not break any existing alp test.

tests/virt_autotest/prepare_alp_host.pm
alice-suse: Well, is it worth the discussion effort on this minor thing ;-) ?

The purpose of the line is to get the image dir contents. Assert should work for this ls command, although the main purpose is not to assert, but to see the saved screenshot from assert_script_run. Yes, record_info can do it too, but it needs to also call script_output. So this way is simpler I think.

I would keep it if you do not insist on change.

lib/virt_autotest/virtual_network_utils.pm
waynechen55: I know br0 has nothing to do with "default". It is metaphor. My opinion is to keep "default" one if there is one regardless of names (it may not use "default" name). I think keeping it does no harm, but if you think it may has negative impact on you test, you can take reasonable action.

tests/virt_autotest/prepare_alp_host.pm
waynechen55: I will not do things this way. But if you insist, you can keep it this way.

lib/virt_autotest/virtual_network_utils.pm
alice-suse: Thanks for the suggestion. I will keep this in mind. When it shows the need to keep "default", will create a new ticket for it.

tests/virt_autotest/prepare_alp_host.pm
alice-suse: Ok, thanks. I will keep it then. But I got your point :).

tests/virt_autotest/prepare_alp_host.pm
alice-suse: Updated.

alice-suse:
All comments are resolved. If no more review, plan to merge it at EOB today.

PR #16503, opened 2023-03-01
Unschedule logs_from_installation_system

ge0r:
Due to poo#122608, test module logs_from_installation_system occasionaly fails. This is because the exit code of the first command we run in the serial console is sometimes not captured and the command times out. 
To avoid this we unschedule logs_from_installation_system where this issue occurs.

- Related ticket: https://progress.opensuse.org/issues/125078
- Needles: No needles
- Verification runs: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=ge0r%2Fos-autoinst-distri-opensuse%23unschedule-logs

### DIFF
--- a/schedule/migration/s390x_regression_test_offline.yaml
+++ b/schedule/migration/s390x_regression_test_offline.yaml
@@ -29,7 +29,7 @@ schedule:
   - installation/disable_grub_timeout
   - installation/start_install
   - installation/await_install
-  - installation/logs_from_installation_system
+  - '{{logs_from_installation_system}}'
   - installation/reboot_after_installation
   - installation/handle_reboot
   - installation/first_boot
@@ -64,6 +64,10 @@ conditional_schedule:
     REGRESSION_SERVICE:
       1:
         - installation/install_service
+  logs_from_installation_system:
+    FLAVOR:
+      Regression-on-Migration-from-SLE15-SPx:
+        - installation/logs_from_installation_system
   regression_tests:
     REGRESSION_TEST:
       1:


jknphy:
this yaml seems that is used in 7 test suites https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml and we don't want to avoid collecting logs in all of them.
btw there is one failure not in regression group mentioned in the progress ticket.

ge0r:
> this yaml seems that is used in 7 test suites https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml and we don't want to avoid collecting logs in all of them. btw there is one failure not in regression group mentioned in the progress ticket.

I think it fails (potentially) on all of the 12-SPX flavor ones, right?
Also the failure not addressed here is because it is related to a MR, I added it to the original ticket now.

jknphy:
> > this yaml seems that is used in 7 test suites https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml and we don't want to avoid collecting logs in all of them. btw there is one failure not in regression group mentioned in the progress ticket.
> 
> I think it fails (potentially) on all of the 12-SPX flavor ones, right? Also the failure not addressed here is because it is related to a MR, I added it to the original ticket now.

The three jobs failing are in this flavors:
Migration-from-SLE12-SPx
Regression-on-Migration-from-SLE12-SPx
Migration-from-SLE12-SPx-Milestone

But I don't think it is a good idea to remove logs collection for test suites not failing, it is pretty handy when new bug appears and we need to file it quickly without trying to retrieve it manually. IMO acting here 'potentially' might not be beneficial. Even a new settings as last option could do...
have you checked `NOLOGS` ? looks like it is what we need.

ge0r:
> > > this yaml seems that is used in 7 test suites https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml and we don't want to avoid collecting logs in all of them. btw there is one failure not in regression group mentioned in the progress ticket.
> > 
> > 
> > I think it fails (potentially) on all of the 12-SPX flavor ones, right? Also the failure not addressed here is because it is related to a MR, I added it to the original ticket now.
> 
> The three jobs failing are in this flavors:
> 
> ```
> Migration-from-SLE12-SPx
> Regression-on-Migration-from-SLE12-SPx
> Migration-from-SLE12-SPx-Milestone
> ```
> 
> But I don't think it is a good idea to remove logs collection for test suites not failing, it is pretty handy when new bug appears and we need to file it quickly without trying to retrieve it manually. IMO acting here 'potentially' might not be beneficial. Even a new settings as last option could do...
> 
> Have you checked `NOLOGS` ? looks like it is what we need.

The jobs in Migration-from-SLE12-SPx and Migration-from-SLE12-SPx-Milestone
 are not affected by this change. They are related to [this MR](https://gitlab.suse.de/coolgw/wegao-test/-/merge_requests/243)
The jobs in Regression-on-Migration-from-SLE12-SPx are addressed with this PR.
The problem is not collecting logs or not, we have to avoid moving to serial console during that 12-SPx install, the first command we type has high probability of timing out because the return code will not get captured.
There is no way to know beforehand if a potentially unstable testsuite will fail or not, so in my opinion, the most safe approach is to unscheduled them from there.
I am not unscheduling them from any jobs that do not suffer from this issue.

ge0r:
It could be that I just do not understand your proposal. What do you suggest exactly?

jknphy:
Thanks for the explanation, I merged the the MR, it is not a REMOTE CONTROLLER scenario, but there is a comment so ok.
For regrestion I have the the doubt that in this [file](https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml) I can find things like this:
```yaml
s390x:
    sle-15-SP5-Regression-on-Migration-from-SLE12-SPx-s390x:
    - offline_sles12sp4_ltss_media_sdk-asmm-contm-lgm-tcm-wsm-pcm_all_full:        
             ...
             YAML_SCHEDULE: 'schedule/migration/s390x_regression_test_offline.yaml'
    - offline_sles12sp4_ltss_pscc_sdk-asmm-contm-lgm-tcm-wsm-pcm_all_full:
             ... 
             YAML_SCHEDULE: 'schedule/migration/s390x_regression_test_offline.yaml'
```
So with this change we will unschedule logs from more than one or I am missing something?

jknphy:
btw, now we are focusing on test modules (discarding entering on it thanks to your investigation) there was a simple way to do it for this one and the other 2 merged:
`EXCLUDE_MODULES` that setting should work in main.pm and with yaml schedule.

ge0r:
> btw, now we are focusing on test modules (discarding entering on it thanks to your investigation) there was a simple way to do it for this one and the other 2 merged: `EXCLUDE_MODULES` that setting should work in main.pm and with yaml schedule.

True, I could have used EXCLUDE_MODULES for the MR, I did not think of it.

ge0r:
> Thanks for the explanation, I merged the the MR, it is not a REMOTE CONTROLLER scenario, but there is a comment so ok. For regression I have the the doubt that in this [file](https://gitlab.suse.de/coolgw/wegao-test/-/blob/master/JobGroups/migration_regression.yaml) I can find things like this:
> 
> ```yaml
> s390x:
>     sle-15-SP5-Regression-on-Migration-from-SLE12-SPx-s390x:
>     - offline_sles12sp4_ltss_media_sdk-asmm-contm-lgm-tcm-wsm-pcm_all_full:        
>              ...
>              YAML_SCHEDULE: 'schedule/migration/s390x_regression_test_offline.yaml'
>     - offline_sles12sp4_ltss_pscc_sdk-asmm-contm-lgm-tcm-wsm-pcm_all_full:
>              ... 
>              YAML_SCHEDULE: 'schedule/migration/s390x_regression_test_offline.yaml'
> ```
> 
> So with this change we will unschedule logs from more than one test suite or I am missing something?

The point of this change is to prevent all s390x testsuites with flavor `Regression-on-Migration-from-SLE12-SPx ` from running the `logs_from_installation_system` module since (from my understanding) they will fail. I do not target specific jobs in that matter. If you think an approach where I target individual testsuites is preferred then I can do this via an MR instead of what I am doing here.

PR #16505, opened 2023-03-01
Split openldap_to_389ds

sofiasyria:
Since openldap2 is depricated after SLE 15SP3, we need to reform out mirgation test so that the openldap is configurated before migration.

- Needed gitlab PR: https://gitlab.suse.de/coolgw/wegao-test/-/merge_requests/250/diffs
- Related ticket: https://progress.opensuse.org/issues/119662
- VRs: 

Tumbleweed : http://falafel.suse.cz/tests/991

SLE: https://openqa.suse.de/tests/overview?version=15-SP5&build=sofiasyria%2Fos-autoinst-distri-opensuse%23ldap&distri=sle

### DIFF
--- a/tests/migration/openldap_to_389ds.pm
+++ b/tests/migration/openldap_to_389ds.pm
@@ -12,9 +12,7 @@ use serial_terminal 'select_serial_terminal';
 use strict;
 use warnings;
 use utils;
-use lockapi;
-use Utils::Systemd qw(systemctl disable_and_stop_service);
-use registration qw(add_suseconnect_product);
+use Utils::Systemd qw(systemctl);
 use version_utils;
 use Utils::Logging 'tar_and_upload_log';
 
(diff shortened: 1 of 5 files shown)


tests/migration/openldap_to_389ds.pm
```diff
@@ -23,58 +21,12 @@ sub run {
...
-    assert_script_run "ps -aux | grep slapd";
-
-    # setup sssd
-    assert_script_run "cp ./sssd.conf /etc/sssd/sssd.conf";
-    systemctl("stop sssd");
-    assert_script_run "rm -rf /var/lib/sss/db/*";
-    systemctl("restart sssd");
-    systemctl("status sssd");
 
     # Prepare data file for migration
```
coolgw, line 28: prepare data section should move to openldap_configuration

sofiasyria: thanks for taking a look. Are you referring to the password replacement bellow? That would make sense. I moved this to the part after migration because dscreate is included in 389-ds and not in openldap2

JRivrain:
LGTM but I see that two of the VR are failing. I had a quick look but did not see anything obvious. I even tried to install openldap2 on the image of the on that fails installing openldap2, and it worked for me. Do you know what' going on?

sofiasyria:
> I see the WIP is still here, I think we could merge at that point. Could be good to have another review just in case, but the change is relatively trivial and passes VRs so up to you, WDYT ?

I kept WIP till I made sure VRs pass. We could merge.

PR #16509, opened 2023-03-01
Automation Development for Guest Migration Test

waynechen55:
* **Add** one more console root-ssh-virt for virtualization testing using ipmi backend. Manual postcopy guest migration needs two commands consecutively, the latter one needs to run during execution of the former, but unfortunately the former does not return until it finishes, so the latter do not have chance to run in the same terminal on some SLES versions. A second console is needed.

* **Extend** script_retry to have $option which can be used for passing additional arguments for "timeout".

* **New** file ```lib/virt_autotest/domain_management_utils.pm``` to hold domain management subroutines, including ```construct_uri```, ```create_guest```, ```remove_guest```, ```shutdown_guest```, ```show_guest``` and ```check_guest_state```.

* **Add** ssh options ```-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no``` in ```check_failures_in_journal``` and ```check_guest_health```, get rid of "localhost" on ssh command if machine is "localhost" and ensure ```validate_script_output``` more resilient to avoid unnecessary failures.

* **YAML** file ```schedule/qam/common/virt/virt_guest_migration.yaml``` to run guest migration test suite. This can be changed later on demand.

* **Add** option ```log_token``` in ```collect_host_and_guest_logs```, so uploaded logs can use personalized names to avoid duplicates.

* **File** ```tests/virt_autotest/parallel_guest_migration_barrier.pm``` creates locks that are going to be used during test urn in advance. ```DO_GUEST_MIGRATION_DONE_[1...N]``` and ```DO_GUEST_MIGRATION_READY_[1...N]``` will created dynamically in ```guest_migration_test```.

* **The** entire test can be broken down into 4 parts, ```prepare_host```, ```prepare_guest```, ```prepare_log``` and ```guest_migration_test```:
  * **```prepare_host```** is responsible for checking architecture, operation system version, virtualization module/service, connected network segment, software packages need to be available, qemu user id, qemu/kvm/libvirt group id, setup nfs shared storage and turning off unnecessary security services on both source and destination hosts.
  * **```prepare_guest```** is responsible for getting info about guest to be tested, initializing test result at the beginning, saving/restoring guest assets, configuring guest clock/storage/serial console, gathering networking info about guest, creating network to be used by guest and populating %guest_matrix which stores info about guest mac address, ip address, type/name/mode of network to which guest connected.
  * **```prepare_host```** and **```prepare_guest```** are mainly based on [Migration Requirements](https://susedoc.github.io/doc-sle/main/single-html/SLES-virtualization/#libvirt-admin-live-migration-requirements).
  * **```prepare_log```** is responsible for enabling debugging log, checking host and guest health before test starts.
  * **```guest_migration_test```** is responsible for performing guest migration tests. Source and destination hosts execute tests in collaborative manner by employing locks ```DO_GUEST_MIGRATION_DONE_[0...N]``` and ```DO_GUEST_MIGRATION_READY_[0...N]```. 
    * **Source** does basic administration on guest before migration if ```GUSET_ADMINISTRATION``` is set followed by guest migration test with virsh or xl command. It restores and re-creates guest before next guest migration test starts by using virsh or xl command accordingly.
    * **Destination** checks migration result from source. If it is successful migration from source, setup passwordless ssh connection, testing networking accessibility and disk read/write, and basic administration after migration will be performed. At last, destination will migrate the guest back to source. Destination only needs remove guest under certain circumstances, for example, failed migration back, undefinesource is not given during migration back or etc.

* **Test** suite level settings:
  * **GUEST_LIST** specifies guest to be tested. If empty, all guests defined on host will be tested.
  * **GUEST_MIGRATION_TEST** specifies migration test to be executed. If empty, all tests in %guest_migration_matrix will be executed.
  * **SKIP_GUEST_INSTALL** indicates whether guest is already installed directly on host. 
  * **GUEST_ADMINISTRATION** indicates whether basic administration on guest is needed.
  * **INTERVAL_LOG** indicates whether collecting log for each failed migration test and cleanup log after each test.
  * **EXTERNAL_NFS_SHARE** indicates both source and destination needs to mount external standalone nfs share.
  * **REGRESSION** indicates MU incidents testing activity which may use static ip address for guest.

* **Structures** being used in test:
  * **```%guest_matrix```** stores networking info about guest. Example as below:
$VAR1 = {
          'sles-15-sp4-64-fv-xen' => {
                                       'netmode' => 'bridge',
                                       'nettype' => 'bridge',
                                       'macaddr' => '52:54:00:ef:05:09',
                                       'netname' => 'br123',
                                       'ipaddr' => '192.168.123.2',
                                       'staticip' => 'no'
                                     },
          'sles-15-sp4-64-pv-xen' => {
                                       'netmode' => 'nat',
                                       'nettype' => 'network',
                                       'macaddr' => '00:16:3e:04:5b:c2',
                                       'netname' => 'vnet_nat_virbr77',
                                       'ipaddr' => '192.168.124.103',
                                       'staticip' => 'no'
                                     }
        };
  * **%guest_network_matrix** stores info about network to be created for guest based on the type of network being used.
  * **%guest_migration_matrix** stores all migrate tests to be executed.
  * **%test_result** stores final test result for all migration tests.

* **Some** run-time settings being used by test:
  * **```LOCAL_IPADDR,LOCAL_FQDN,PEER_IPADDR and PEER_FQDN```** store ip address and FQDN of source and destination hots.
  * **```GUEST_UNDER_TEST```** stores guest to be tested.
  * **```GUEST_UNDER_TEST_MACADDR,GUEST_UNDER_TEST_IPADDR,GUEST_UNDER_TEST_NETTYPE,GUEST_UNDER_TEST_NETNAME, GUEST_UNDER_TEST_NETMODE and GUEST_UNDER_TEST_STATICIP```** are derived from ```%guest_matrix```. So destination host can have exactly the same piece of information about guest during test run.
  * **```TEST_RUN_PROGRESS```** stores current subroutine in action as progress. This is achieved by calling ```set_test_run_progress``` in subroutine. So peer test run can know state of each other, especially when one in ```post_fail_hook```, failed test run calls ```check_peer_test_run``` to get test result of peer test run by examining ```TEST_RUN_PROGRESS``` to determine whether it needs to wait for peer to finish operation after failure as well and avoid interrupting peer to finish as normal.
  * **```TEST_RUN_RESULT```** stores the final test run result.

* **```check_guest_network_config```** retrieves guest network info from its xml config and store in ```%guest_matrix``` which will be used by ```create_guest_network``` to create guest network together with ```%guest_network_matrix```. In order to be consistent and simple, the network name should take the following form:
  * **If** guest uses virtual network created by ```virsh net-define/net-created```, the network name should be like ```vnet_``` + ```nat/route/host``` + ```_other_strings``` if network uses nat, route or host bridge respectively. ```$guest_matrix{guest}{netmode}``` will be nat, route or host accordingly and ```$guest_matrix{guest}{nettype}``` will be 'network'. For guest uses default network, ```$guest_matrix{guest}{netmode}``` will be default as well.
  * **If** guest uses bridge device directly, both ```$guest_matrix{guest}{nettype}``` and ```$guest_matrix{guest}{netmode}``` will be 'bridge'. But if guest used br0 directly, ```$guest_matrix{guest}{netmode}``` will be 'host'.

* **The** detailed execution flow of this Guest Migration Test is as below:
Screenshot 2023-03-27 at 21 15 06
Screenshot 2023-03-27 at 21 16 58
Screenshot 2023-03-27 at 21 19 30

* **Verification runs:**
  * Passed xen migration 15sp5 guest from 15sp5 host to 15sp5 host with administration:
    * [source test run](http://10.163.28.134/tests/1204)
    * [destination test run](http://10.163.28.134/tests/1203)
  * Failed xen migration 15sp5 guest from 15sp5 host to 15sp5 host with administration and log for each failure:
    *  [source test run](http://10.163.28.134/tests/1206)
    * [destination test run](http://10.163.28.134/tests/1205)
  * Passed kvm migration 15sp5 guest from 15sp5 host to 15sp5 host with administration and two test failures/logs:
    * [source test run](http://10.163.28.134/tests/1222)
    * [destination test run](http://10.163.28.134/tests/1221)
  * Passed xen migration 12sp3 from 12sp3 host to 12sp3 host with administration:
    * [source test run](http://10.163.28.134/tests/1246) 
    * [destination test run](http://10.163.28.134/tests/1245)
  * Passed kvm migration 12sp3 from 12sp3 host to 12sp3 host without administration:
    * [source test run](http://10.163.28.134/tests/1266)
    * [destination test run](http://10.163.28.134/tests/1265)
  * Failed xen migration 15sp5 host to 15sp4 host:
    * [source test run](http://10.163.28.134/tests/1290)
    * [destination test run](http://10.163.28.134/tests/1289)
  * Passed xen migration 15sp4 guest from 15sp4 host to 15sp5 host without administration and logs:
    * [source test run](http://10.163.28.134/tests/1294)
    * [destination test run](http://10.163.28.134/tests/1293)
  * Passed kvm migration 15sp4 guest from 15sp4 host to 15sp5 host without administration and logs:
    * [source test run](http://10.163.28.134/tests/1322)
    * [destination test run](http://10.163.28.134/tests/1321)
  * MU Scenario - Passed 12sp3-15sp4 guest between 15sp4 kvm without administration:
    * [source test run](http://10.163.28.134/tests/1485)
    * [destination test run](http://10.163.28.134/tests/1484)
  * MU Scenario - Passed 12sp3-15sp4 guest between 15sp4 xen without administration:
    * [source test run](http://10.163.28.134/tests/1537)
    * [destination test run](http://10.163.28.134/tests/1536)
  * MU Scenario - Passed 12sp3-15sp4 guest between 12sp3 kvm without administration:
    * [source test run](http://10.163.28.134/tests/1645)
    * [destination test run](http://10.163.28.134/tests/1644)
  * MU Scenario - Passed 12sp3-15sp4 guest between 12sp3 xen without administration:
    * [source test run](http://10.163.28.134/tests/1633)
    * [destination test run](http://10.163.28.134/tests/1632)
  * [Guest Installation KVM](https://openqa.suse.de/tests/10796583)
  * [Guest Installation XEN](https://openqa.suse.de/tests/10796586)
  * [UEFI Guest Installation KVM](https://openqa.suse.de/tests/10819008)
  * [UEFI Guest Installation XEN](https://openqa.suse.de/tests/10825459)
  * [PRJ2 Verification Run](https://openqa.suse.de/tests/10853140)
  * Please refer to [this checklist](https://progress.opensuse.org/issues/124928) for full list of MU integration verification runs

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -518,6 +518,16 @@ sub init_consoles {
                 serial => 'rm -f /dev/sshserial; mkfifo /dev/sshserial; chmod 666 /dev/sshserial; while true; do cat /dev/sshserial; done',
                 gui => 1
             });
+        $self->add_console(
+            'root-ssh-virt',
+            'ssh-xterm',
+            {
+                hostname => get_required_var('SUT_IP'),
+                password => $testapi::password,
+                username => 'root',
+                serial => 'rm -f /dev/virtsshserial; mkfifo /dev/virtsshserial; chmod 666 /dev/virtsshserial; while true; do cat /dev/virtsshserial; done',
+                gui => 1
+            }) if (get_var('VIRT_AUTOTEST', '') or get_var('REGRESSION', ''));
     }
 
     # Use ssh consoles on generalhw, without VNC connection
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1815,17 +1815,18 @@ sub script_retry {
     my $retry = $args{retry} // 10;
     my $delay = $args{delay} // 30;
     my $timeout = $args{timeout} // 30;
+    my $option = $args{option} // '';
     my $die = $args{die} // 1;
     my $fail_msg = $args{fail_message} // "Waiting for Godot: $cmd";
 
     my $ret;
 
-    my $exec = "timeout $timeout $cmd";
+    my $exec = "timeout $option $timeout $cmd";
     # Exclamation mark needs to be moved before the timeout command, if present
     if (substr($cmd, 0, 1) eq "!") {
         $cmd = substr($cmd, 1);
         $cmd =~ s/^\s+//;    # left trim spaces after the exclamation mark
-        $exec = "! timeout $timeout $cmd";
+        $exec = "! timeout $option $timeout $cmd";
     }
     for (1 .. $retry) {
         # timeout for script_run must be larger than for the 'timeout ...' command
--- a/lib/virt_autotest/domain_management_utils.pm
+++ b/lib/virt_autotest/domain_management_utils.pm
@@ -1,0 +1,277 @@
+# VIRTUALIZAITON DOMAIN MANAGEMENT UTILITIES
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Domain management utilities providied by various tools,
+# for example, libvirt, xl and etc.
+#
+# Maintainer: Wayne Chen <wchen@suse.com>, qe-virt <qe-virt@suse.de>
+package virt_autotest::domain_management_utils;
+
+use strict;
+use warnings;
+use testapi;
+use virt_autotest::utils qw(is_kvm_host is_xen_host);
+use utils qw(script_retry);
+use Carp;
+
+our @EXPORT = qw(
+  construct_uri
+  create_guest
+  remove_guest
+  shutdown_guest
+  show_guest
+  check_guest_state
+);
+
+=head2 construct_uri
+
+Construct connection URI to be used with virsh command. URI is composed of many
+parts which are all supported here. Please refer to libvirt page for details:
+https://libvirt.org/uri.html#remote-uris.
+=cut
+
+sub construct_uri {
+    my (%args) = @_;
+    $args{driver} //= '';
+    $args{transport} //= 'ssh';
+    $args{user} //= '';
+    $args{host} //= 'localhost';
+    $args{port} //= '';
+    $args{path} //= 'system';
+    $args{extra} //= '';
+    $args{driver} = is_kvm_host ? "qemu" : "xen" if (!$args{driver});
+
+    my $uri = "";
+    if ($args{host} eq 'localhost') {
+        $uri = "$args{driver}:///$args{path}";
+    }
+    else {
+        $uri = $args{transport} ? "$args{driver}+$args{transport}://" : "$args{driver}://";
+        $uri .= "$args{user}@" if ($args{user});
+        $uri .= $args{host};
+        $uri .= ":$args{port}" if ($args{port});
+        $uri .= "/";
+        $uri .= $args{path} if ($args{path});
+        $uri .= "?$args{extra}" if ($args{extra});
+    }
+    return $uri;
+}
+
+=head2 create_guest
+
+Create guest by using virsh define or xl create. Main arguments are guest to be
+created, virtualization management tool to be used, whether die (1) or not (0)
+if any failures happen, directory in which guest xml and xl config are stored
+and whether start (1) or not (0) guest after defining it. This subroutine also 
+calls construct_uri to determine the desired URI to be connected if the interested 
+party is not localhost. Please refer to subroutine construct_uri for the arguments 
+related.
+=cut
+
+sub create_guest {
+    my (%args) = @_;
+    $args{guest} //= '';
+    $args{virttool} //= 'virsh';
+    $args{confdir} //= '/var/lib/libvirt/images';
+    $args{start} //= 1;
+    $args{die} //= 0;
+    $args{driver} //= '';
+    $args{transport} //= 'ssh';
+    $args{user} //= '';
+    $args{host} //= 'localhost';
+    $args{port} //= '';
+    $args{path} //= 'system';
+    $args{extra} //= '';
+    croak("Guest to be created must be given") if (!$args{guest});
+
+    my $ret = 0;
+    my $uri = "--connect=" . construct_uri(driver => $args{driver}, transport => $args{transport}, user => $args{user}, host => $args{host}, port => $args{port}, path => $args{path}, extra => $args{extra});
+    foreach my $guest (split(/ /, $args{guest})) {
+        my $temp = 1;
+        if ($args{virttool} eq 'virsh') {
+            $temp = script_run("virsh $uri define --file $args{confdir}/$guest.xml --validate");
+            $temp |= script_run("virsh $uri start $guest") if ($args{start} == 1);
+            record_info("Failed to create guest $guest from $args{confdir}/$guest.xml", "Failed to create guest $guest using virsh define/start --file $args{confdir}/$guest.xml --validate", result => 'fail') if ($temp != 0);
+        }
+        elsif ($args{virttool} eq 'xl') {
+            $temp = script_run("ls $args{confdir}/$guest.cfg") == 0 ? 0 : script_run("virsh $uri domxml-to-native --xml $args{confdir}/$guest.xml --format xen-xl > $args{confdir}/$guest.cfg");
+            $temp |= script_run("xl -vvv create $args{confdir}/$guest.cfg");
+            record_info("Guest $guest creating failed", "Failed to create guest $guest using xl -vvv create $args{confdir}/$guest.cfg", result => 'fail') if ($temp != 0);
+        }
+        $ret |= $temp;
+        save_screenshot;
+        record_info("Guest $guest config", script_output("cat $args{confdir}/$guest.xml;cat $args{confdir}/$guest.cfg", type_command => 1, proceed_on_failure => 1));
+    }
+    croak("Failed to define all guests") if ($ret != 0 and $args{die} == 1);
+    return $ret;
+}
+
+=head2 shutdown_guest
+
+Shutdown guest and verify result. Main arguments are guest to be powered off,
+virtualization management tool (virsh or xl) to be used and whether die (1) or
+not (0) if any failures happen. This subroutine also calls construct_uri to
+determine the desired URI to be connected if the interested party is not
+localhost. Please refer to subroutine construct_uri for the arguments related.
+=cut
+
+sub shutdown_guest {
+    my (%args) = @_;
+    $args{guest} //= '';
+    $args{virttool} //= 'virsh';
+    $args{die} //= 0;
+    $args{driver} //= '';
+    $args{transport} //= 'ssh';
+    $args{user} //= '';
+    $args{host} //= 'localhost';
+    $args{port} //= '';
+    $args{path} //= 'system';
+    $args{extra} //= '';
+    croak("Guest to be shut down must be given") if (!$args{guest});
+
+    my $ret = 0;
+    my $uri = "--connect=" . construct_uri(driver => $args{driver}, transport => $args{transport}, user => $args{user}, host => $args{host}, port => $args{port}, path => $args{path}, extra => $args{extra});
+    foreach my $guest (split(/ /, $args{guest})) {
+        my $temp = 1;
+        if ($args{virttool} eq 'virsh') {
+            if (script_run("virsh $uri list --state-shutoff | grep \"$guest \"") != 0) {
+                script_retry("virsh $uri shutdown $guest", retry => 12, delay => 10, die => 0);
+                if (script_retry("virsh $uri list --state-shutoff | grep \"$guest \"", retry => 12, delay => 10, die => 0) != 0) {
+                    script_run("virsh $uri destroy $guest");
+                    $temp = script_retry("virsh $uri list --state-shutoff | grep \"$guest \"", retry => 12, delay => 10, die => 0);
+                }
+                else {
+                    $temp = 0;
+                }
+            }
+            else {
+                $temp = 0;
+            }
+        }
+        elsif ($args{virttool} eq 'xl') {
+            if (script_run("xl -vvv list | grep \"$guest .*---s-- \"") != 0) {
+                script_retry("xl -vvv shutdown $guest", retry => 12, delay => 10, die => 0);
+                $temp = script_retry("xl -vvv list | grep \"$guest .*---s-- \"", retry => 12, delay => 10, die => 0);
+            }
+            else {
+                $temp = 0;
+            }
+        }
+        $ret |= $temp;
+        save_screenshot;
+        record_info("Failed to stop guest $guest", "Guest $guest can not be stopped using $args{virttool} shutdown or destroy", result => 'fail') if ($temp != 0);
+    }
+    croak("Failed to stop all guests") if ($ret != 0 and $args{die} == 1);
+    return $ret;
+}
+
+=head2 remove_guest
+
+Remove guest forcibly. Main arguments are guest to be removed, and whether die
+(1) or not (0) if any failures happen. This subroutine also calls construct_uri
+to determine the desired URI to be connected if the interested party is not
+localhost. Please refer to subroutine construct_uri for the arguments related.
+=cut
+
+sub remove_guest {
+    my (%args) = @_;
+    $args{guest} //= '';
+    $args{die} //= 0;
+    $args{driver} //= '';
+    $args{transport} //= 'ssh';
+    $args{user} //= '';
+    $args{host} //= 'localhost';
+    $args{port} //= '';
+    $args{path} //= 'system';
+    $args{extra} //= '';
+    croak("Guest to be removed must be given") if (!$args{guest});
+
+    my $ret = 0;
+    my $uri = "--connect=" . construct_uri(driver => $args{driver}, transport => $args{transport}, user => $args{user}, host => $args{host}, port => $args{port}, path => $args{path}, extra => $args{extra});
+    foreach my $guest (split(/ /, $args{guest})) {
+        my $temp = 1;
+        script_run("virsh $uri destroy $guest");
+        $temp = script_run("virsh $uri list --all | grep \"$guest \"") == 0 ? script_run("virsh $uri undefine $guest || virsh $uri undefine $guest --keep-nvram") : 0;
+        $temp |= script_run("xl -vvv destroy $guest") if (is_xen_host and script_run("xl list | grep \"$guest \"") == 0);
+        $ret |= $temp;
+        save_screenshot;
+        record_info("Guest $guest removing failed", "Failed to remove guest $guest", result => 'fail') if ($temp != 0);
+    }
+    croak("Failed to remove all guests") if ($ret != 0 and $args{die} == 1);
+    return $ret;
+}
+
+=head2 show_guest
+
+Show all guests available on host or specified ones. Main arguments are guest to
+be checked, virtualization management tool to be used (virsh or xl) and whether 
+die (1) or not (0) if any failures happen. This subroutine also calls construct_uri 
+to determine the desired URI to be connected if the interested party is not localhost. 
+Please refer to subroutine construct_uri for the arguments related.
+=cut
+
+sub show_guest {
+    my (%args) = @_;
+    $args{guest} //= '';
+    $args{virttool} //= 'virsh';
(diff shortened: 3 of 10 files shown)


waynechen55:
@alice-suse @guoxuguang @Julie-CAO @nanzhg @RoyCai7 @tbaev Please have a look.

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+# Initialize above global structures to facilitate test run
+# Check hosts meet requirements for guest migration test
+# Check guets meet requirements for guest migration test
+# Restore and create environment for guest to start
+# Check and update information about guest during test run
+# Check final test results and create junit logs
+# Synchronization between peers in pass/fail situations
+#
+# Maintainer: Wayne Chen <wchen@suse.com>, qe-virt <qe-virt@suse.de>
+package parallel_test_base;
```
alice-suse, line 22: Judging from file name, I thought it was a new base class for general parallel tests. After checking details, this package contains quite some migration specific settings. I guess it is not intended to serve as general parallel test base? Then, how about updating the package name to more specific name and updating the summary?

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+        record_info("Config $_addr SSH PubKey auth");
+        next if (script_run("timeout --kill-after=3 --signal=9 15 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$_addr ls") == 0);
+        enter_cmd("clear", wait_still_screen => 3);
+        enter_cmd("ssh-copy-id -f -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ~/.ssh/id_rsa.pub root\@$_addr", wait_still_screen => 3);
+        if ($args{host} == 1) {
+            susedistribution::handle_password_prompt();
+        }
+        else {
+            check_screen("password-prompt", 60);
+            enter_cmd("novell", wait_screen_change => 50, max_interval => 1);
```
alice-suse, line 355: For the guest password, better to not use explicit value here? Can be passed in by SECRET_XX.

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+            wait_still_screen(10);
+        }
+        my $_temp = 1;
+        $_temp = script_run("timeout --kill-after=3 --signal=9 15 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$_addr ls");
+        $_ret |= $_temp;
+        record_info("Guest $_addr SSH PubKeyAuth failed", "Can not establish ssh connection to guest $_addr using Public Key Authentication", result => 'fail') if ($_temp != 0);
```
alice-suse, line 361: Host will also go to this line based on code logic?

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+    record_info("Check host os");
+    my $_ret = 0;
+    my $_localip = get_var('LOCAL_IPADDR');
+    my $_peerip = get_var('PEER_IPADDR');
+    if (is_sle) {
+        my ($_localosver, $_localossp,) = get_os_release;
+        my ($_peerosver, $_peerossp,) = get_os_release("ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$_peerip");
+        print "LOCAL OS $_localosver-sp$_localossp PEER OS $_peerosver-sp$_peerossp\n";
+        save_screenshot;
+        unless (($_peerosver > $_localosver) or ($_peerosver == $_localosver and $_peerossp >= $_localossp)) {
```
alice-suse, line 404: This does not seem correct when $args{role} =dst ?

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+sub check_host_network {
+    my $self = shift;
+
+    record_info("Check host network");
+    my $_localgw = script_output("ip route show default | awk \'{print \$3}\'");
+    my $_localip = get_var('LOCAL_IPADDR');
+    my $_peerip = get_var('PEER_IPADDR');
+    my $_peergw = script_output("ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$_peerip ip route show default | awk \'{print \$3}\'");
+    save_screenshot;
+    croak("$_localip and $_peerip hosts are not in the same subnet") unless ($_localgw eq $_peergw);
```
alice-suse, line 457: Would you please share where this requirement comes? I do not remember any on this for migration. In real migration, I think it is possible that src and dst are on different sub-network.

waynechen55: https://susedoc.github.io/doc-sle/main/single-html/SLES-virtualization/#sec-libvirt-admin-migrate
```Host and target machine should be in the same subnet on the network, otherwise networking will not work after the migration. ```

File: lib/parallel_test_base.pm [hunk 6]
```diff
@@ -0,0 +1,1946 @@
...
+is packages to be installed.
+=cut
+
+sub check_host_package {
+    my ($self, %args) = @_;
+    $args{package} //= '';
+
+    record_info("Check host package");
+    zypper_call("--gpg-auto-import-keys ref");
+    is_kvm_host ? zypper_call("in -t pattern kvm_tools") : zypper_call("in -t pattern xen_tools");
```
alice-suse, line 472: This won't work for alp... It is ok to be extended in new PR later, but better keep in mind the difference about "host" between alp and sle, and add some if ... else or TODO comment ahead of time, to save some redesign/extend effort.

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+
+    record_info("Check host virtualization");
+    if (is_kvm_host) {
+        assert_script_run("lsmod | grep kvm");
+    }
+    elsif (is_xen_host) {
+        assert_script_run("lsmod | grep xen");
+    }
+
+    if (script_run("systemctl is-active libvirtd") != 0) {
```
alice-suse, line 434: This won't work for alp...

alice-suse (commented):
Review wip...

lib/parallel_test_base.pm
waynechen55: This is correct. Please give you example if you think this is not correct.

lib/parallel_test_base.pm
waynechen55: IF...ELSE can be added when needed.

lib/parallel_test_base.pm
waynechen55: You can skip for alp if needed later.

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1946 @@
...
+    foreach my $_ss (@_security_service) {
+        if (script_run("systemctl is-enabled $_ss") == 0) {
+            systemctl("stop $_ss");
+            systemctl("disable $_ss");
+            save_screenshot;
+        }
+    }
+
+    if (script_run("ls /etc/selinux/config") == 0) {
+        assert_script_run("sed -i -r \'s/^SELINUX=.*\$/SELINUX=disabled/g\' /etc/selinux/config");
```
alice-suse, line 603: This will require reboot to take effect...

waynechen55: Do you want to remove this line ?

lib/parallel_test_base.pm
alice-suse: Yes, that's true. My understanding is that the src and dst should be both in the network that the migrated vm is created in/used in its network config,  but it is not necessarily the default network for src and dst?  Otherwise, it won't say "otherwise networking will not work AFTER the migration" ?

lib/parallel_test_base.pm
alice-suse: If role is dst, this function will directly return the initialized _ret 0 (means supported/correct) if peer/src is 12sp3 and local/dst is 15sp4.

alice-suse: if role is dst, src host is 12sp3 and dst host is 15sp4, this should be returned 0, but here it will return 1 .

alice-suse: I see, in later usage, it seems return value is not cared, the croak or not croak actually matters. That will be ok. But better differentiate return value between croak and not croak case.

waynechen55: Return value is also OK. 0 is OK, 1 is NOTOK. The difference is whether croak.

lib/parallel_test_base.pm
waynechen55: Do you mean "the same subnet" refers to the network in which src/dst guest reside ?

lib/parallel_test_base.pm
alice-suse: If role is src, src host is 12sp3 and dst host is 15sp4, the function is returning 0. So correct. 
If role is dst, src host is 12sp3 and dst host is 15sp4, the function is returning 1 and the record_info on Line 410 is giving correct info that it is valid, so conflicting info. "No croak" should match return 0, isn't it. You can also have a unit test in perl.

Besides, from just logic, if the design is to croak for invalid scenario, then it can happen when role is either src or dst. But now in this code, croak only happens to role=src. Problematic, isn't it?

lib/parallel_test_base.pm
alice-suse: That's my understanding. As long as src and dst host can reach each other, the vm will be able to be migrated from src to dst. Then if dst has the same network that the vm xml requires in its network configuration, which also applied on src host, the vm's networking will work after migration.

lib/parallel_test_base.pm
alice-suse: Actually I was express similar things with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16509/files#r1132001780.

lib/parallel_test_base.pm
alice-suse: It seems the existing host setup in developing and MU test is ok on this, and doesn't affect VT test.  For alp, selinux will be setup during host installation too. Migration test will be integrated after those steps, so should be fine on this , I think.

lib/parallel_test_base.pm
waynechen55: No. The recod_info is correct even with your example. For dst, there is no need to croak. Because dst does not have to migration back to src if src falls behind dst. But return value has to tell the real result. There is no problem here.

lib/parallel_test_base.pm
waynechen55: I can remove this checking if you think this is not very necessary. But I think in the same subnet is reasonable.

lib/parallel_test_base.pm
waynechen55: I do not test on ALP, I can help add is_sle if you want. But I do not want to add something like is_alp or !is_alp, because I mainly do verification run on SLES.

lib/parallel_test_base.pm
waynechen55: I can still remove this I think anyway.

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1961 @@
...
+    my ($self, %args) = @_;
+    $args{guest} //= '';
+    $args{die} //= 0;
+    croak("Guest to be tested must be given") if (!$args{guest});
+
+    my $_ret = 0;
+    foreach my $_guest (split(/ /, $args{guest})) {
+        record_info("Test $_guest network");
+        my $_temp = 1;
+        $_temp = script_run("timeout --kill-after=3 --signal=9 20 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$_guest ping -c5 openqa.suse.de");
```
alice-suse, line 1624: This won't work for O3/Tumbleweed. Suggest to use separate openqa for different product..

lib/susedistribution.pm
```diff
@@ -518,6 +518,16 @@ sub init_consoles {
...
                 gui => 1
             });
+        $self->add_console(
+            'root-ssh-virt',
+            'ssh-xterm',
+            {
+                hostname => get_required_var('SUT_IP'),
+                password => $testapi::password,
+                username => 'root',
+                serial => 'rm -f /dev/sshserial; mkfifo /dev/sshserial; chmod 666 /dev/sshserial; while true; do cat /dev/sshserial; done',
```
alice-suse, line 528: In this way, the two consoles `root-ssh` and `root-ssh-virt` will share the same /dev/sshserial. It will be problematic when both consoles are actively used at the same time. Suggest to use different device name .

lib/parallel_test_base.pm
```diff
@@ -0,0 +1,1961 @@
...
+=head2 shutdown_guest
+
+Shutdown guest and verify result. Main arguments are guest to be powered off,
+virtualization management tool (virsh or xl) to be used and whether die (1) or
+not (0) if any failures happen. This subroutine also calls construct_uri to
+determine the desired URI to be connected if the interested party is not
+localhost. Please refer to subroutine construct_uri for the arguments related.
+=cut
+
+sub shutdown_guest {
```
alice-suse, line 1143: For these  general apis in this block, create_guest/define_guest/shutdown_guest/etc, before implementing them in this new module parallel_test_base.pm which serves migration/parallel mainly, is it being considered extending the existing library functions of similar functionality , or extending the existing library files by putting them there so that non-parallel jobs can use them too?

lib/parallel_test_base.pm
alice-suse: Seems I was expecting differently. Doesn't matter. Up to your detailed design and need of course.

lib/parallel_test_base.pm
alice-suse: Got your point. It is ok on my side. Adding is_sle would work to tell where alp won't  fit and remind the designer the gap for multi-product automation support. This would save too much re-design work when extending it for TW/ALP. BTW, I am afraid Oliver would remind that this repo needs to have default for tumbleweed/leap. Better to pay attention too.

lib/parallel_test_base.pm
alice-suse: I know it takes effort to implement and test the code for this. Sorry for suggesting to remove. Among our pair machines on OSD, there is one pair won't match same subnet, scooter is moved to new FC lab while the other one is in old NUE lab. It can also save some test time.

lib/parallel_test_base.pm
waynechen55: Extending existing functions might not be good enough, because those existing ones are too shabby. I will consider putting these new ones in library files if feasible (Needs fulfill test purpose for guest migration in the first place).

lib/virt_autotest/domain_management_utils.pm
```diff
@@ -0,0 +1,277 @@
...
+    $args{extra} //= '';
+    croak("Guest to be removed must be given") if (!$args{guest});
+
+    my $ret = 0;
+    my $uri = "--connect=" . construct_uri(driver => $args{driver}, transport => $args{transport}, user => $args{user}, host => $args{host}, port => $args{port}, path => $args{path}, extra => $args{extra});
+    foreach my $guest (split(/ /, $args{guest})) {
+        my $temp = 1;
+        script_run("virsh $uri destroy $guest");
+        $temp = script_run("virsh $uri list --all | grep \"$guest \"") == 0 ? script_run("virsh $uri undefine $guest || virsh $uri undefine $guest --keep-nvram") : 0;
+        $temp |= script_run("xl -vvv destroy $guest") if (script_run("xl list | grep \"$guest \"") == 0);
```
RoyCai7, line 197: if KVM, no such command xl.

waynechen55: Added is_xen_host

alice-suse:
> Ready to be merged for next stage activity @alice-suse

@RoyCai7  @tbaev @Julie-CAO Hello guys, I am going to merge it if no more comment before EOB today  :-) . If you have, feel free to give it here.

lib/utils.pm
mpagot, line 1800: Documentation is missing

PR #16520, opened 2023-03-02
Remove soft-failure bsc#1195046

skliu:
Remove soft-failure bsc#1195046

**Ticket:**

- https://progress.opensuse.org/issues/125069

**Verification run:**
- https://openqa.suse.de/tests/10609677
- https://openqa.suse.de/tests/10610216
- https://openqa.suse.de/tests/10610836
- https://openqa.suse.de/tests/10611327
- https://openqa.suse.de/tests/10614085
- https://openqa.suse.de/tests/10614158
- https://openqa.suse.de/tests/10614778
- https://openqa.suse.de/tests/10614861
- https://openqa.suse.de/tests/10614871
- https://openqa.suse.de/tests/10614928
- https://openqa.suse.de/tests/10614950
- https://openqa.suse.de/tests/10615011

### DIFF
--- a/lib/migration.pm
+++ b/lib/migration.pm
@@ -35,11 +35,6 @@ our @EXPORT = qw(
 sub setup_sle {
     select_console 'root-console';
 
-    if (is_ppc64le && is_sle('<=12-sp5')) {
-        record_soft_failure('bsc#1195046 - ncurses display a wrong checker board character');
-        systemctl('restart systemd-vconsole-setup.service');
-    }
-
     # Stop packagekitd
     if (is_sle('12+')) {
         quit_packagekit;


lib/migration.pm
```diff
@@ -36,7 +36,6 @@ sub setup_sle {
     select_console 'root-console';
 
     if (is_ppc64le && is_sle('<=12-sp5')) {
```
jknphy, line 38: could you please check if we can remove the whole condition, I think after the bug is fixed, we don't even need to restart the service. But you will need at least 10 runs, because I remember it was a race condition.

lemon-suse: So it is not a workaround for bug, just a test issue, I don't think we need remove the condition since other arches do not have this issue.

jknphy: I'm proposing remove the condition and the workaround and the soft-failure, remove all, because I believe this is already fixed.

jknphy: @skliu could you please paste in your first PR comment those 10 runs to evaluate?

skliu: Done.

PR #16522, opened 2023-03-02
Use better variables for MPI module to avoid interchange values

b10n1k:
When `openmpi` is used, because the various versions on each product we try to
get the required one using `get_mpi` function, despite the MPI job variable.
The mpi_master uses it but some operation need plain `openmpi`. Before the
implementation reused the $mpi variable but that had to interchange for one
step to the other. Using a new variable `mpi2load` it stores the value to be
used for loading the mpi module and let untouched the $mpi which can be used
in other steps later on.

### DIFF
--- a/tests/hpc/mpi_master.pm
+++ b/tests/hpc/mpi_master.pm
@@ -26,6 +26,7 @@ sub run ($self) {
     my $mpi = $self->get_mpi();
     my ($mpi_compiler, $mpi_c) = $self->get_mpi_src();
     my $mpi_bin = 'mpi_bin';
+    my $mpi2load = '';
     my @cluster_nodes = $self->cluster_names();
     my $cluster_nodes = join(',', @cluster_nodes);
     my %exports_path = (
@@ -45,7 +46,7 @@ sub run ($self) {
     type_string('pkill -u root', lf => 1) unless $user_virtio_fixed;
     select_user_serial_terminal($prompt);
     # for <15-SP2 the openmpi2 module is named simply openmpi
-    $mpi = 'openmpi' if ($mpi =~ /openmpi2|openmpi3|openmpi4/);
+    $mpi2load = 'openmpi' if ($mpi =~ /openmpi2|openmpi3|openmpi4/);
 
     barrier_wait('CLUSTER_PROVISIONED');
     record_info 'CLUSTER_PROVISIONED', strftime("\%H:\%M:\%S", localtime);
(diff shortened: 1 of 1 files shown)


tests/hpc/mpi_master.pm
```diff
@@ -151,3 +151,66 @@ sub post_fail_hook ($self) {
...
+
+=item $mpi_c
+The source code to compile and run. The source codes are located in
+ L<data|data/hpc>
+
+=item $mpi_bin
+Holds the name of the compiled source code
+
+=item $cluster_nodes
+A str represantation of all the nodes of the cluster, including master node.
```
Avinesh, line 175: s/represanation/representation

tests/hpc/mpi_master.pm
```diff
@@ -151,3 +151,66 @@ sub post_fail_hook ($self) {
...
+=item $mpi2load
+differentiates the openmpi name to be used in lmod loading. C<lmod> can load
+only one mpi. In case of openmpi2, openmpi3, openmpi4 which is stored in C<mpi>,
+it takes their place as all are found as I<openmpi>
+
+=item $hostname
+It just holds the I<hostname> to avoid recall C<get_var> again and again
+
+=item $mpirun_s
+Holds an object which implements wrappers for B<mpirun>. Implentation can be
```
Avinesh, line 196: s/Implentation/Implementation

tests/hpc/mpi_master.pm
```diff
@@ -45,7 +45,7 @@ sub run ($self) {
     type_string('pkill -u root', lf => 1) unless $user_virtio_fixed;
     select_user_serial_terminal($prompt);
     # for <15-SP2 the openmpi2 module is named simply openmpi
-    $mpi = 'openmpi' if ($mpi =~ /openmpi2|openmpi3|openmpi4/);
+    my $mpi2load = 'openmpi' if ($mpi =~ /openmpi2|openmpi3|openmpi4/);
```
Avinesh, line 48: I think variable needs to be declared before this conditional statement to pass CI check.

Avinesh (approved):
Couple of nits, otherwise good!

tests/hpc/mpi_master.pm
b10n1k: fixed

tests/hpc/mpi_master.pm
b10n1k: fixed

PR #16334, opened 2023-02-07
Include azurec cli test shell script

yarunachalam:
Include azure cli test shell script - Virtual Network test and Run Command test
Calling shell script from azure_cli.pm
Verification Run : (https://openqa.suse.de/tests/11233972#
https://openqa.suse.de/tests/11234043#settings
https://openqa.suse.de/tests/11234170#)
https://github.com/yarunachalam/os-autoinst-distri-opensuse.git#azcli

https://openqa.suse.de/tests/11265261# - Verification Run with renamed parameter
PUBLIC_CLOUD_AZURE_CLI_TEST | vmss
PUBLIC_CLOUD_TOOLS_CLI | 1

Created multiple smaller PR for remaining test scripts
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17073
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17074

PR #16528, opened 2023-03-02
Add wrapper script to create azure vm

yarunachalam:
create azure vm before calling azure_cli test

- Verification run: WIP


asmorodskyi (changes requested):
just add this code as a function to azure_cli , it should not be another test . Also we should keep possibility to run azure_cli test without this repos

yarunachalam:
as per anton's suggested merged this changes to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16334/files

Closing the ticket

PR #16533, opened 2023-03-03
WIP: Introduce services health test run

grisu48:
Adds a test run that checks for failing systemd services.

- Verification runs: [MicroOS](https://duck-norris.qe.suse.de/tests/12315)


jlausuch (changes requested):
This is already covered by `journal_check.pm` with a whitelist of bugs that can be marked as soft-fail or ignored.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/journal_check.pm#L95

grisu48:
> This is already covered by `journal_check.pm` with a whitelist of bugs that can be marked as soft-fail or ignored. https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/journal_check.pm#L95

Ah ,thnx, didn't knew that. Then this is indeed not needed.

PR #16536, opened 2023-03-03
lib/x11utils: Use Ctrl instead of Esc to show the lockscreen

Vogtinator:
Plasma 5.27 introduced that Esc turns the monitor off. With 5.27.2, it no longer causes the lockscreen UI to show, so we can't use it here anymore. Ctrl appears to work fine.

Verification runs:

KDE Live: https://openqa.opensuse.org/tests/3154656
GNOME Live: https://openqa.opensuse.org/tests/3154657
Xfce Live: https://openqa.opensuse.org/tests/3154662

Any other VRs needed? The lxde and the otherDE tests apparently don't trigger the lockscreen.


DimStar77 (approved):
Let's try - worst case we'll have to revert over the weekend and find better solutions

Vogtinator:
> Let's try - worst case we'll have to revert over the weekend and find better solutions

If necessary just an `if is_kde;` bandaid until a better key shows up...

PR #16541, opened 2023-03-06
Fix broken lockscreen on SLES

grisu48:
Fix the broken unlocking of the lockscreen on all SLES versions.

- Related ticket: https://progress.opensuse.org/issues/125414
- Verification run: [TW Gnome live](https://duck-norris.qe.suse.de/tests/12317) | [TW XFCE live](https://duck-norris.qe.suse.de/tests/12318) | [TW KDE live](https://duck-norris.qe.suse.de/tests/12316) | [12-SP5 consoletest_finish](https://duck-norris.qe.suse.de/tests/12319) | [12-SP5 shutdown](https://duck-norris.qe.suse.de/tests/12320) | [15-SP3 python_liblouis](https://duck-norris.qe.suse.de/tests/12321) (fails in different module) | [15-SP3 shutdown](https://duck-norris.qe.suse.de/tests/12320) | [15-SP5](https://duck-norris.qe.suse.de/tests/12323)

### DIFF
--- a/lib/x11utils.pm
+++ b/lib/x11utils.pm
@@ -179,7 +179,8 @@ sub ensure_unlocked_desktop {
         die 'ensure_unlocked_desktop repeated too much. Check for X-server crash.' if ($counter eq 1);    # die loop when generic-desktop not matched
         if (match_has_tag('screenlock') || match_has_tag('blackscreen')) {
             wait_screen_change {
-                send_key 'ctrl';    # end screenlock
+                # ESC of KDE turns the monitor off and CTRL does not work on older SLES versions to unlock the screen
+                send_key(is_sle("<15-SP4") ? 'esc' : 'ctrl');    # end screenlock
                 diag("Screen lock present");
             };
             next;    # Go directly to assert_screen, skip wait_still_screen (and don't collect $200)


lib/x11utils.pm
```diff
@@ -179,7 +179,7 @@ sub ensure_unlocked_desktop {
         die 'ensure_unlocked_desktop repeated too much. Check for X-server crash.' if ($counter eq 1);    # die loop when generic-desktop not matched
         if (match_has_tag('screenlock') || match_has_tag('blackscreen')) {
             wait_screen_change {
-                send_key 'ctrl';    # end screenlock
+                send_key(is_sle("<=15-SP4") ? 'esc' : 'ctrl');    # end screenlock
```
Vogtinator, line 182: Needs a comment.

Does Ctrl work on 15 SP5?

What about GNOME on Leap 15.4?

grisu48: Checking ....

Vogtinator (approved):
If that works, +1

Vogtinator:
Idea for a possible other key that might work everywhere (for a future PR): backspace! Using space, tab or return might break some lockscreens in case that is treated as input, but backspace shouldn't have any negative effects.

dzedro:
I would exclude just TW KDE where 'Esc turns the monitor off'.
Single VR of each version does not prove this random behavior.
15-SP5 VRs are missing which will use new key.

Vogtinator:
> I would exclude just TW KDE where 'Esc turns the monitor off'.

If you go that way, check for KDE on all versions please and not just TW. `ctrl` is known to work fine with them and 15.5 will have the same Plasma as TW.

grisu48:
> I would exclude just TW KDE where 'Esc turns the monitor off'. Single VR of each version does not prove this random behavior. 15-SP5 VRs are missing which will use new key.

I checked 15-SP5 manually. From 15-SP4 onwards CTRL works.

dzedro:
> > I would exclude just TW KDE where 'Esc turns the monitor off'.
> 
> If you go that way, check for KDE on all versions please and not just TW. `ctrl` is known to work fine with them and 15.5 will have the same Plasma as TW.

Ye if `ctrl` works on KDE then that is the way.

PR #16549, opened 2023-03-07
Enable FIPS testing for SLED - Core maintenance

rfan1:
https://progress.opensuse.org/issues/125195
https://bugzilla.suse.com/show_bug.cgi?id=1208951

- Verification run: 
[Firefox-with fips](https://openqa.suse.de/tests/10633356)
[Firefox-non-fips](http://openqa.suse.de/tests/10633358)
[SLED-others](http://openqa.suse.de/tests/10626751)

### DIFF
--- a/schedule/qam/15-SP4/qam-regression-other_fips.yaml
+++ b/schedule/qam/15-SP4/qam-regression-other_fips.yaml
@@ -1,0 +1,22 @@
+---
+name: qam regression other tests with fips enabled for Vendor Affirmation
+schedule:
+- boot/boot_to_desktop
+- fips/fips_setup
+- x11/window_system
+- console/consoletest_setup
+- x11/brasero/brasero_launch
+- x11/gnome_documents
+- x11/totem/totem_launch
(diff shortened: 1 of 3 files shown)


schedule/qam/15-SP4/qam-regression-other_fips.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: qam regression other tests with fips enabled for Vendor Affirmation
+schedule:
+- boot/boot_to_desktop
```
foursixnine, line 4: openjdk is missing

foursixnine: Yeah, for both (SLES and SLED)... x11 is not an issue in this case

rfan1: Let me do!

rfan1: http://openqa.suse.de/tests/10641681#

PR #16555, opened 2023-03-07
Workaround s390-tools bug on 15-SP4

dzedro:
- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1208983
- Verification run:
https://openqa.suse.de/tests/10630301#step/s390tools/11 -> https://openqa.suse.de/tests/10634243
https://openqa.suse.de/tests/10630300#step/s390tools/11 -> https://openqa.suse.de/tests/10634732

### DIFF
--- a/tests/qam-minimal/s390tools.pm
+++ b/tests/qam-minimal/s390tools.pm
@@ -13,6 +13,7 @@ use strict;
 use warnings;
 use testapi;
 use utils 'zypper_call';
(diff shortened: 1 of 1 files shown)


dzedro:
This should be reverted soon, in bug is already working fixed script, but it can take some time until the fix is committed and released. I will revert this workaround when it's released.

PR #16557, opened 2023-03-07
podman_pods: set replica count to 1 in hello-kubic.yaml

dcermak:
Since version 4.4.0, podman will limit the number of replicas to 1 when using podman play kube. Thus it makes no sense anymore to have it set to 3 as that avoids additional branches depending on the podman version. (Also, supporting replicas was never really intended upstream and thus we shouldn't have tested it anyway.)

- Related ticket: fix for https://build.opensuse.org/request/show/1068441
- Needles: Not required
- Verification run: https://openqa.opensuse.org/tests/3161257

### DIFF
--- a/tests/containers/podman_pods.pm
+++ b/tests/containers/podman_pods.pm
@@ -6,7 +6,7 @@
 # Package: podman
 # Summary: Test podman pods functionality
 # - Use data/containers/hello-kubic.yaml to run pods
-# - Confirm 3 pods are spawned
+# - Confirm 1 pod is spawned
 # - Clean up pods using hello-kubic.yaml
 # Maintainer: Richard Brown <rbrown@suse.com>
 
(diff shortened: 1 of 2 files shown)


dcermak:
The failure in the test run is unrelated to this specific issue/fix.

danishprakash:
> The failure in the test run is unrelated to this specific issue/fix.

More info here - https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16567

tests/containers/podman_pods.pm
```diff
@@ -28,13 +28,12 @@ sub run {
 
     record_info('Test', 'Confirm pods are running');
     record_info('pod ps', script_output('podman pod ps'));
-    assert_script_run('podman pod ps | grep "Running" | wc -l | grep -q 3');
+    assert_script_run('podman pod ps | grep "Running" | wc -l | grep -q 1');
```
danishprakash, line 31: Just curious how this would come out for tests that are run against podman < 4.4.0? Should this be conditional?

dcermak: It would if the replica count in the yaml was still 3, but because I wanted to avoid branches here, I changed it to 1.

dcermak: You call it clever, I call it lazy ;-)

PR #16566, opened 2023-03-08
Avoid adding openqa repo in leap

lemon-suse:
Avoid removing official repos and add openQA repos twice in aarch Leap 15.5 through set KEEP_ONLINE_REPOS=1.

- Related ticket: https://progress.opensuse.org/issues/124889
- Needles: yast2_control-center-no-update-repo, yast2_control-center_online-update_close already merged in OSD.
-  Related PR: https://github.com/os-autoinst/opensuse-jobgroups/pull/283
- Verification run:  https://openqa.opensuse.org/tests/3161815#step/yast2_control_center/39

### DIFF
--- a/tests/yast2_gui/yast2_control_center.pm
+++ b/tests/yast2_gui/yast2_control_center.pm
@@ -79,7 +79,7 @@ sub start_online_update {
             if (is_tumbleweed()) {
                 zypper_call("ar -f http://download.opensuse.org/ports/$repo_arch/update/tumbleweed repo-update");
             } else {
-                 zypper_call("ar -f http://download.opensuse.org/ports/update/$update_name repo-update");
+                zypper_call("ar -f http://download.opensuse.org/ports/update/$update_name repo-update");
             }
         }
         select_console 'x11', await_console => 0;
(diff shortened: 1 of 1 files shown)


rakoenig:
I am desperately trying to understand how your PR solves the issue. If I see the [error from the ticket](https://openqa.opensuse.org/tests/3134032#step/yast2_control_center/44) we see a warning popup showing up after invoking the Online-Update function that is not handled by openQA which leads to the failure.

When looking at [your VR](https://openqa.opensuse.org/tests/3161815#step/yast2_control_center/39) we now see another popup that gots handled by the new needle with the tag `yast2_control-center-no-update-repo`. But this does not explain why we end up here and not in that warning that we have seen 15 days ago. 

Seems like something else has changed in the environment that does not trigger the problem, but I don't see anything that prevents us from running into this in the future.

lemon-suse:
> I am desperately trying to understand how your PR solves the issue. If I see the [error from the ticket](https://openqa.opensuse.org/tests/3134032#step/yast2_control_center/44) we see a warning popup showing up after invoking the Online-Update function that is not handled by openQA which leads to the failure.
> 
> When looking at [your VR](https://openqa.opensuse.org/tests/3161815#step/yast2_control_center/39) we now see another popup that gots handled by the new needle with the tag `yast2_control-center-no-update-repo`. But this does not explain why we end up here and not in that warning that we have seen 15 days ago.
> 
> Seems like something else has changed in the environment that does not trigger the problem, but I don't see anything that prevents us from running into this in the future.

It is because I just follow the condition in the code to set KEEP_ONLINE_REPOS=1 then no update repo will be added, while this will only for the test in leap 15.5 so I have MR to update test suite.

PR #16567, opened 2023-03-08
tests/container: update grep for podman top

danishprakash:
While starting a container with --user, podman appends a /etc/passwd entry in which it uses the UID for the host/container user. Starting v4.4.1, it now replaces[1] the UID with the Username if present. Fixing the grep here to play nicely with the updated top output.

[1] - https://github.com/containers/podman/pull/17174

- Related ticket: https://build.opensuse.org/request/show/1068441
- Needles: 

### DIFF
--- a/tests/containers/rootless_podman.pm
+++ b/tests/containers/rootless_podman.pm
@@ -154,7 +154,18 @@ sub verify_userid_on_container {
     $cid = script_output "podman run -d --rm --name test2 --user 1000 $image sleep infinity";
     $cid = check_bsc1200623($cid);
     my $id = $start_id + $huser_id - 1;
-    validate_script_output "podman top $cid user huser", sub { /1000\s+${id}/ };
+
+    # podman >= v4.4.0 lists username instead of uid
+    # when using `user` descriptor with `podman top`, handle
+    # conditionally.
+    # https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16567
+    my $podman_version = get_podman_version();
+    if (package_version_cmp($podman_version, '4.4.0') >= 0) {
+        my $huser_name = script_output "whoami";
+        validate_script_output "podman top $cid user huser", sub { /${huser_name}\s+${id}/ };
+    } else {
+        validate_script_output "podman top $cid user huser", sub { /1000\s+${id}/ };
+    }
     validate_script_output "podman top $cid capeff", sub { /none/ };
 
     record_info "root with keep-id", "the default user(root) starts process with the same uid as host user";


jlausuch:
You need a check for podman version, otherwise this will fail in versions <4.4

tests/containers/rootless_podman.pm
```diff
@@ -154,7 +154,8 @@ sub verify_userid_on_container {
     $cid = script_output "podman run -d --rm --name test2 --user 1000 $image sleep infinity";
     $cid = check_bsc1200623($cid);
     my $id = $start_id + $huser_id - 1;
-    validate_script_output "podman top $cid user huser", sub { /1000\s+${id}/ };
+    my $huser_name = script_output "echo \$USERNAME";
```
dcermak, line 157: You can use `'` to prevent variable expansion in strings:
```suggestion
    my $huser_name = script_output 'echo $USERNAME';
```
However, I'd advise against relying on the value of the `USERNAME` variable, as I am not sure if that's a standard value (it's e.g. not set in the fish shell). I'd suggest to use `$(whoami)` instead.

jlausuch:
Also, you won't be able to test this until the SR is accepted and available on TW.

tests/containers/rootless_podman.pm
danishprakash: That's a good point, I've however, almost always assumed that USERNAME was as standard as UID on most systems. Updated.

dcermak:
> Also, you won't be able to test this until the SR is accepted and available on TW.

you can test against Staging:L

danishprakash:
> Also, you won't be able to test this until the SR is accepted and available on TW.

Not sure I understand that honestly. Can I not trigger a run based off my branch and then rerun the failing openQA tests? Is that possible?

tests/containers/rootless_podman.pm
dcermak: > Why'd you use command substitution there though? `$(whoami)` I don't know much perl but going by the function name, it looks like it'll return the stdout from whatever command being run as arg.

Because I'm dumb :facepalm: … `$(whoami)` makes absolutely no sense and `script_run 'whoami';` is the correct form. I guess it's time for another :coffee:

dcermak:
> Not sure I understand that honestly. Can I not trigger a run based off my branch and then rerun the failing openQA tests? Is that possible?

You can trigger the tests based of this branch. What @jlausuch meant, was that as podman 4.4. is currently not yet in Factory, your changes wouldn't be verified. But as podman 4.4 has been submitted to Factory and is currently staged in Staging:L, you can go to that staging project: https://build.opensuse.org/staging_workflows/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:L, pick one of the openQA tests pasted there by the openQA bot and clone the appropriate test from there. It should contain podman 4.4.

jlausuch:
This was merged without  approvals and even without VRs... this could break things everywhere if not verified. Please don't do this again.

I just started:
https://openqa.opensuse.org/t3162572

jlausuch:
Backwards compatibility VR (podman 4.3.1) https://openqa.opensuse.org/t3162577

tests/containers/rootless_podman.pm
```diff
@@ -154,7 +154,18 @@ sub verify_userid_on_container {
...
+
+    # podman >= v4.4.0 lists username instead of uid
+    # when using `user` descriptor with `podman top`, handle
+    # conditionally.
+    # https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16567
+    my $podman_version = get_podman_version();
+    if (package_version_cmp($podman_version, '4.4.0') >= 0) {
+        my $huser_name = script_output "whoami";
+        validate_script_output "podman top $cid user huser", sub { /${huser_name}\s+${id}/ };
+    } else {
```
m-dati, line 166: `$id` here still is _numeric_: is the` sub {...} ` here compatible ?

tests/containers/rootless_podman.pm
```diff
@@ -154,7 +154,18 @@ sub verify_userid_on_container {
...
     my $id = $start_id + $huser_id - 1;
-    validate_script_output "podman top $cid user huser", sub { /1000\s+${id}/ };
+
+    # podman >= v4.4.0 lists username instead of uid
+    # when using `user` descriptor with `podman top`, handle
+    # conditionally.
+    # https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16567
+    my $podman_version = get_podman_version();
+    if (package_version_cmp($podman_version, '4.4.0') >= 0) {
+        my $huser_name = script_output "whoami";
```
Vogtinator, line 164: Is `testapi::$username` not correct?

jlausuch: Most likely yes, as that's the default user created by default, and manually in microos. So, it could be used that way, yes.

jlausuch:
Staging https://openqa.opensuse.org/tests/3162572#step/rootless_podman/158
vs
Current Factory  https://openqa.opensuse.org/tests/3162577#step/rootless_podman/169

tests/containers/rootless_podman.pm
danishprakash: Thanks for pointing this out, I'll raise another PR with this change.

tests/containers/rootless_podman.pm
danishprakash: It _was_ working as expected before so I think, yes. But I'm not well-versed with perl so.

PR #16575, opened 2023-03-08
Add default gnome and kde badges to the readme

foursixnine:
Since https://github.com/os-autoinst/openQA/pull/5022 enables badges on o3, we can now add that to multiple places, not only to showcase but also as a visual aid, especially on PR.

zimage


grisu48 (approved):
LGTM!

PR #16577, opened 2023-03-08
Boottime measures triggered at P.C. instance creation

m-dati:
Primary Target of this PR in publiccloud is to trigger the boot time measurements collection  in the VM instance creation phase, instead of running in a separate module.
And as the collection as pushing data to InfluxDB are enabled by means of proper parmeters setting.
The _data push_ has been tested in a new created `perf_2` bucket and also in the existing one `perf` ( see param. `PUBLIC_CLOUD_PERF_DB`), successfully.

More details are in the [note-8](https://progress.opensuse.org/issues/118027#note-8) of the  poo.

- Needles: N/A
- Verification run:   
[x86_64 AZURE 3715](http://10.168.4.117/tests/3715)  
[aarch64 GCE 3716](http://10.168.4.117/tests/3716)  
[x86_64 GCE 3718](http://10.168.4.117/tests/3718)  

This PR  to replace of the [16476](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16476)

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -22,6 +22,7 @@ use utils;
 use db_utils;
 use Mojo::Util 'trim';
 use Data::Dumper;
+use mmapi qw(get_current_job_id);
 
 use constant SSH_TIMEOUT => 90;
 
@@ -555,14 +556,16 @@ sub store_boottime_db() {
     my $token = get_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
     unless ($url && $token) {
         record_info("WARN", "PUBLIC_CLOUD_PERF_DB_URI or _PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');
-        return;
+        return -1;
     }
+
     my $org = get_var('PUBLIC_CLOUD_PERF_DB_ORG', 'qec');
     my $db = get_var('PUBLIC_CLOUD_PERF_DB', 'perf');
 
     my $tags = {
         instance_type => get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE'),
-        provider => get_required_var('PUBLIC_CLOUD_PROVIDER'),
+        job_id => get_current_job_id(),
+        os_provider => get_required_var('PUBLIC_CLOUD_PROVIDER'),
         os_build => get_required_var('BUILD'),
         os_flavor => get_required_var('FLAVOR'),
         os_version => get_required_var('VERSION'),
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -518,6 +518,8 @@ sub measure_boottime() {
     record_info("BOOT TIME", 'systemd_analyze');
     # first deployment analysis
     my ($systemd_analyze, $systemd_blame) = do_systemd_analyze_time($instance);
+    return 0 unless ($systemd_analyze && $systemd_blame);
+
     $ret->{analyze}->{$_} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
     $ret->{blame} = $systemd_blame;
     $ret->{type} = $type;
@@ -556,7 +558,7 @@ sub store_boottime_db() {
     my $token = get_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
     unless ($url && $token) {
         record_info("WARN", "PUBLIC_CLOUD_PERF_DB_URI or _PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');
-        return -1;
+        return 0;
     }
 
     my $org = get_var('PUBLIC_CLOUD_PERF_DB_ORG', 'qec');
@@ -599,8 +601,10 @@ sub store_boottime_db() {
 sub systemd_time_to_second
 {
     my $str_time = trim(shift);
+
     if ($str_time !~ /^(?<check_min>(?<min>\d{1,2})\s*min\s*)?((?<sec>\d{1,2}\.\d{1,3})s|(?<ms>\d+)ms)$/) {
-        die("Unable to parse systemd time '$str_time'");
+        record_info("WARN", "Unable to parse systemd time '$str_time'", result => 'fail');
+        return -1;
     }
     my $sec = $+{sec} // $+{ms} / 1000;
     $sec += $+{min} * 60 if (defined($+{check_min}));
@@ -617,8 +621,9 @@ sub extract_analyze_time {
         $time = trim($time);
         my ($time, $type) = $time =~ /^(.+)\s*\((\w+)\)$/;
         $res->{$type} = systemd_time_to_second($time);
+        return 0 if ($res->{$type} == -1);
     }
-    map { die("Fail to detect $_ timing") unless exists($res->{$_}) } qw(kernel initrd userspace overall);
+    map { return 0 unless exists($res->{$_}) } qw(kernel initrd userspace overall);
     return $res;
 }
 
@@ -629,6 +634,7 @@ sub extract_blame_time {
         $line = trim($line);
         my ($time, $service) = $line =~ /^(.+)\s+(\S+)$/;
         $ret->{$service} = systemd_time_to_second($time);
+        return 0 if ($ret->{$service} == -1);
     }
     return $ret;
 }
@@ -641,7 +647,15 @@ sub do_systemd_analyze_time {
     my @ret;
 
     # calling syastemd-analyze time & blame
-    $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
+    while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
+        $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
+        sleep 5;
+    }
+
+    unless ($output && (time() - $start_time < $args{timeout})) {
+        record_info("WARN", "Unable to get system-analyze in $args{timeout} seconds", result => 'fail');
+        return (0, 0);
+    }
     push @ret, extract_analyze_time($output);
 
     $output = $instance->run_ssh_command(cmd => 'systemd-analyze blame', proceed_on_failure => 1);
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -260,9 +260,9 @@ sub upload_log {
     assert_script_run("test -d '$tmpdir' && rm -rf '$tmpdir'");
 }
 
-=head2 wait_for_guestregister
+=head2 wait_for_guestregister_chk
 
-    wait_for_guestregister([timeout => 300]);
+    wait_for_guestregister_chk([timeout => 300]);
 
 Run command C<systemctl is-active guestregister> on the instance in a loop and
 wait till guestregister is ready. If guestregister finish with state failed,
@@ -270,31 +270,34 @@ a soft-failure will be recorded.
 If guestregister will not finish within C<timeout> seconds, job dies.
 In case of BYOS images we checking that service is inactive and quit
 Returns the time needed to wait for the guestregister to complete.
+C<wait_for_guestregister_chk> is called inside C<create_instance()>, enabled by C<check_guestregister>
 =cut
 
-sub wait_for_guestregister {
+sub wait_for_guestregister_chk {
     my ($self, %args) = @_;
     $args{timeout} //= 300;
     my $start_time = time();
     my $last_info = 0;
+    my $log = '/var/log/cloudregister';
+    my $name = $autotest::current_test->{name} . '-cloudregister.log.txt';
 
     # Check what version of registercloudguest binary we use
     $self->run_ssh_command(cmd => "rpm -qa cloud-regionsrv-client", proceed_on_failure => 1);
-
+    record_info('CHECK', 'guestregister check');
     while (time() - $start_time < $args{timeout}) {
         my $out = $self->run_ssh_command(cmd => 'sudo systemctl is-active guestregister', proceed_on_failure => 1, quiet => 1);
         # guestregister is expected to be inactive because it runs only once
         if ($out eq 'inactive') {
-            $self->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
+            $self->upload_log($log, log_name => $name);
             return time() - $start_time;
         }
         elsif ($out eq 'failed') {
-            $self->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
+            $self->upload_log($log, log_name => $name);
             $out = $self->run_ssh_command(cmd => 'sudo systemctl status guestregister', quiet => 1);
             return time() - $start_time;
         }
         elsif ($out eq 'active') {
-            $self->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
+            $self->upload_log($log, log_name => $name);
             die "guestregister should not be active on BYOS" if (is_byos);
         }
 
@@ -305,8 +308,26 @@ sub wait_for_guestregister {
         sleep 1;
     }
 
-    $self->upload_log('/var/log/cloudregister', log_name => $autotest::current_test->{name} . '-cloudregister.log');
+    $self->upload_log($log, log_name => $name);
     die('guestregister didn\'t end in expected timeout=' . $args{timeout});
+}
+
+
+=head2 wait_for_guestregister
+
+    wait_for_guestregister([timeout => 300]);
+
+The previous functionality has been migrated into wait_for_guestregister_chk above, for logic refactoring 
+planned to run directly when publiccloud create_instances executed. 
+This function is now a temporary placeholder without effects, to quickly proof the new logic, 
+but passing the existing calls still present in many modules,reducing the changes of deleting those ones.
+After the new logic will be consilidated, this function and related calls will be removed. 
+=cut
+
+sub wait_for_guestregister {
+    my ($self, %args) = @_;
+    my $out = $self->run_ssh_command(cmd => 'sudo systemctl is-active guestregister', proceed_on_failure => 1, quiet => 1);
+    return;
 }
 
 =head2 wait_for_ssh
@@ -532,11 +553,8 @@ sub measure_boottime() {
     # Do logging to openqa UI
     $Data::Dumper::Sortkeys = 1;
     record_info("RESULTS", Dumper($ret));
-
-    my $logs = '/var/log/cloudregister /var/log/cloud-init.log /var/log/cloud-init-output.log /var/log/messages /var/log/NetworkManager';
-    $instance->run_ssh_command(cmd => 'sudo tar -czvf /tmp/sle_cloud.tar.gz ' . $logs, proceed_on_failure => 1);
-    $instance->upload_log('/tmp/sle_cloud.tar.gz', failok => 1);
-
+    my @logs = qw(cloudregister cloud-init.log cloud-init-output.log messages NetworkManager);
+    $instance->upload_log("/var/log/" . $_, log_name => 'measure_boottime_' . $_ . '.txt', failok => 1) foreach (@logs);
     return $ret;
 }
 
@@ -623,7 +641,7 @@ sub extract_analyze_time {
         $res->{$type} = systemd_time_to_second($time);
         return 0 if ($res->{$type} == -1);
     }
-    map { return 0 unless exists($res->{$_}) } qw(kernel initrd userspace overall);
+    foreach (qw(kernel initrd userspace overall)) { return 0 unless exists($res->{$_}); }
     return $res;
 }
 
@@ -646,7 +664,8 @@ sub do_systemd_analyze_time {
     my $output = "";
     my @ret;
 
-    # calling syastemd-analyze time & blame
+    # calling systemd-analyze time & blame
+    # guestregister check executed in create_instances
     while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
         $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
         sleep 5;
@@ -654,6 +673,7 @@ sub do_systemd_analyze_time {
 
     unless ($output && (time() - $start_time < $args{timeout})) {
         record_info("WARN", "Unable to get system-analyze in $args{timeout} seconds", result => 'fail');
+        # handle_boot_failure: soft exit from measurement.
         return (0, 0);
     }
     push @ret, extract_analyze_time($output);
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -498,111 +498,84 @@ sub network_speed_test() {
 Perfomrance measurement of the system Boot time. 
 Mainly used C<systemd-analyze> command for the data extraction.
 Data is then collected in an internal record, ready for storing in a DB.
-- when PUBLIC_CLOUD_PERF_COLLECT=0 boottime measurement disabled;
-- when bit-0 ON,First reboot only enabled =1;
-- when bit-1 ON, first,softreboot enabled =3 [2 too];
-- when bit-2 ON, first,hardreboot enabled =5 [4 too];
-- when all bits ON,all 3 measures enabled =7 [6 too].
+Set PUBLIC_CLOUD_PERF_COLLECT true or >0, to activate boottime measurements.
 
 =cut
 
 sub measure_boottime() {
-    my ($self, $instance) = @_;
-    my $perf = get_var('PUBLIC_CLOUD_PERF_COLLECT');
-    return 0 unless ($perf);
-
-    my $softreboot = $perf & 2;    # bit-1 check
-    my $hardreboot = $perf & 4;    # bit-2 check
+    my ($self, $instance, $type) = @_;
+    return 0 unless (get_var('PUBLIC_CLOUD_PERF_COLLECT'));
 
     my $ret = {
         kernel_release => undef,
         kernel_version => undef,
+        type => undef,
         analyze => {},
-        blame => {
-            first => {}, soft => {}, hard => {}
-        },
+        blame => {},
     };
+
     record_info("BOOT TIME", 'systemd_analyze');
     # first deployment analysis
-    my ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
+    my ($systemd_analyze, $systemd_blame) = do_systemd_analyze_time($instance);
     $ret->{analyze}->{$_} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
-    $ret->{blame}->{first} = $systemd_blame;
+    $ret->{blame} = $systemd_blame;
+    $ret->{type} = $type;
+    # $ret->{analyze}->{ssh_access} = $startup_time; # placeholder for next implementation
 
     # Collect kernel version
-    $ret->{kernel_release} = $instance->run_ssh_command(cmd => 'uname -r');
-    $ret->{kernel_version} = $instance->run_ssh_command(cmd => 'uname -v');
-
-    if ($softreboot) {
-        # Do soft reboot
-        my ($shutdown_time, $startup_time) = $instance->softreboot();
-        $ret->{analyze}->{ssh_access_soft} = $startup_time;
-
-        ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
-        $ret->{analyze}->{$_ . '_soft'} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
-        $ret->{blame}->{soft} = $systemd_blame;
-    }
-
-    if ($hardreboot) {
-        # Do hard reboot
-        $instance->stop();
-        $ret->{analyze}->{ssh_access_hard} = $instance->start();
-
-        ($systemd_analyze, $systemd_blame) = do_systemd_analyze($instance);
-        $ret->{analyze}->{$_ . '_hard'} = $systemd_analyze->{$_} foreach (keys(%{$systemd_analyze}));
-        $ret->{blame}->{hard} = $systemd_blame;
-    }
+    $ret->{kernel_release} = $instance->run_ssh_command(cmd => 'uname -r', proceed_on_failure => 1);
+    $ret->{kernel_version} = $instance->run_ssh_command(cmd => 'uname -v', proceed_on_failure => 1);
 
     # Do logging to openqa UI
     $Data::Dumper::Sortkeys = 1;
     record_info("RESULTS", Dumper($ret));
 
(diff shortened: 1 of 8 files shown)


lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,166 @@ sub network_speed_test() {
...
+Save data collected with measure_boottime in a DB;
+Mainly stored on a remote InfluxDB on a Grafana server.
+
+=cut
+
+sub store_boottime_db() {
+    my ($self, $results) = @_;
+    return unless (get_var('_PUBLIC_CLOUD_PERF_PUSH_DATA') && $results);
+
+    my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
```
asmorodskyi, line 544:
```
my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
my $token = get_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
unless ($url && $token) {
        record_info("WARN", "PUBLIC_CLOUD_PERF_DB_URI or _PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');    
}
```

If _PUBLIC_CLOUD_PERF_PUSH_DATA is defined but PUBLIC_CLOUD_PERF_DB_URI or _PUBLIC_CLOUD_PERF_DB_TOKEN is missing I would assume some misconfiguration in such case . We still don't want to fail but we can add something which will drug attention of reviewer ...

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,166 @@ sub network_speed_test() {
...
+    my $token = get_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
+    return unless ($token);
+
+    my $org = get_var('PUBLIC_CLOUD_PERF_DB_ORG', 'qec');
+    my $db = get_var('PUBLIC_CLOUD_PERF_DB', 'perf');
+
+    my $tags = {
+        instance_type => get_required_var('PUBLIC_CLOUD_INSTANCE_TYPE'),
+        os_build => get_required_var('BUILD'),
+        os_flavor => get_required_var('FLAVOR'),
```
asmorodskyi, line 556: let's also include `provider => get_required_var('PUBLIC_CLOUD_PROVIDER'),`

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,166 @@ sub network_speed_test() {
...
+
+sub do_systemd_analyze_time {
+    my ($instance, %args) = @_;
+    $args{timeout} = 120;
+    my $start_time = time();
+    my $output = "";
+    my @ret;
+
+    # Wait for guest register, before calling syastemd-analyze
+    $instance->wait_for_guestregister();
```
asmorodskyi, line 630: this should not be here , this was needed in test where this was part of the flow , here we removing all this . we just assuming that instance already up and running . if it is not than it is a problem of the code which called boot measurement w/o making sure that system is booted

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,166 @@ sub network_speed_test() {
...
+    while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
+        $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
+        sleep 5;
+    }
+
+    die("Unable to get system-analyze in $args{timeout} seconds") unless (time() - $start_time < $args{timeout});
```
asmorodskyi, line 636: same here , just remove this line

lib/publiccloud/instance.pm
```diff
@@ -477,4 +481,166 @@ sub network_speed_test() {
...
+sub do_systemd_analyze_time {
+    my ($instance, %args) = @_;
+    $args{timeout} = 120;
+    my $start_time = time();
+    my $output = "";
+    my @ret;
+
+    # Wait for guest register, before calling syastemd-analyze
+    $instance->wait_for_guestregister();
+    while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
```
asmorodskyi, line 631: no loop needed because we not waiting for anything here , we just measuring system state **when measurement was called** and other code will make sure that it is done in proper time and place

variables.md
```diff
@@ -307,7 +307,7 @@ PUBLIC_CLOUD_PERF_THRESH_CHECK | boolean | "" | If set to `1` or any not empty v
 PUBLIC_CLOUD_PREPARE_TOOLS | boolean | false | Activate prepare_tools test module by setting this variable.
 PUBLIC_CLOUD_GOOGLE_PROJECT_ID | string | "" | GCP only, used to specify the project id.
 PUBLIC_CLOUD_PROVIDER | string | "" | The type of the CSP (e.g. AZURE, EC2, GCE).
-PUBLIC_CLOUD_QAM | boolean | false | Previously used to recognize maintenance jobs - now deprecated - should be removed.
+PUBLIC_CLOUD_QAM | boolean | false | Used to select Maintenance jobs (true/1) or "Latest" jobs (false/0). The 1st prepare a cloud instance in top module before any other test, the 2nd prepare instance in other different phases. This param. also used to trigger creation or reuse of an instance.
```
asmorodskyi, line 310:
```suggestion
PUBLIC_CLOUD_QAM | boolean | false | 1 - identify job running to test maintenance updates , 0 - identifying job testing "Latest " ( in development ) . Used to control all behavioral implications which this brings
```

asmorodskyi:
generally I like it , let's address my comments and it can be merged from my POV . @pdostal WDYT ?

lib/publiccloud/instance.pm
m-dati: > We still don't want to fail but

But if we cannot write to any DB_URI, could we set it as soft failure, at least ?

asmorodskyi: no,  we may confuse a lot of test results in such case , let's keep it transparent for other tests . "keep transparent" in such case means no job status change and minimizing chances for test failure  as it will be not the main goal of any test which we starting

asmorodskyi: ah sorry , I forgot to add return ofc. it should be 
```
my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
my $token = get_var('_PUBLIC_CLOUD_PERF_DB_TOKEN');
unless ($url && $token) {
        record_info("WARN", "PUBLIC_CLOUD_PERF_DB_URI or _PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');    
       return -1;
}

File: lib/publiccloud/instance.pm (hunk 2 above)
Comment (m-dati) [reply lib/publiccloud/instance.pm:556]: I will add the ID into the data->values of InfluxDB record

File: lib/publiccloud/instance.pm (hunk 5 above)
Comment (m-dati) [reply lib/publiccloud/instance.pm]: After **loop** for Startup finished removed, I got many test rerun Failing for:
```
Bootup is not yet finished (org.freedesktop.systemd1.Manager.FinishTimestampMonotonic=0).
Please try again later.
Hint: Use 'systemctl list-jobs' to see active jobs
SCRIPT_FINISHEDRo5gS-1-
```
see: 
[3723](http://10.168.4.117/tests/3723) 
[3719](http://10.168.4.117/tests/3719) 
[3721](http://10.168.4.117/tests/3721)

It seems that the measure needs for some  wait-sync before to start, so I restore that loop and retry

**Comment by m-dati** [comment]
After **loop** for Startup finished has been removed, I got many test rerun Failing for:
```
Bootup is not yet finished (org.freedesktop.systemd1.Manager.FinishTimestampMonotonic=0).
Please try again later.
Hint: Use 'systemctl list-jobs' to see active jobs
SCRIPT_FINISHEDRo5gS-1-
```
see: 
[3723](http://10.168.4.117/tests/3723) 
[3719](http://10.168.4.117/tests/3719) 
[3721](http://10.168.4.117/tests/3721)

It seems that the measure needs for some  wait-sync before to start, so I restore that loop and retry

File: lib/publiccloud/instance.pm (hunk 5 above)
Comment (m-dati) [reply lib/publiccloud/instance.pm]: see note https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16577#issuecomment-1460886997

**Comment by m-dati** [comment]
After last update, loop waiting end boot restored and some error management added,  
the tests PASS:
[3726](http://10.168.4.117/tests/3726)
[3727](http://10.168.4.117/tests/3727)
[3728](http://10.168.4.117/tests/3728)

File: lib/publiccloud/provider.pm [hunk 7]
```diff
@@ -313,6 +313,9 @@ sub create_instances {
             # Install server's ssh publicckeys to prevent authenticity interactions
             assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
         }
+        # Performance data: boottime
```
asmorodskyi, line 316: $instance->wait_for_guestregister();

m-dati:
After last code update, these V.R. PASS;
LTP tests, with boottime measurement and data push to db `perf_2`:
- [3738](http://10.168.4.117/tests/3738) GCE Incidents	x86_64
- [3739](http://10.168.4.117/tests/3739) GCE Incidents	aarch64
- [3744](http://10.168.4.117/tests/3744) Azure Incidents	x86_64

InfluxdDB bucket has been checked too and related data are present ,see exported file:
[2023-03-15_12 35_influxdb_data](https://confluence.suse.com/download/attachments/1204486382/2023-03-15_12%2035_influxdb_data.csv)

lib/publiccloud/instance.pm
```diff
@@ -255,19 +260,26 @@ sub upload_log {
     assert_script_run("test -d '$tmpdir' && rm -rf '$tmpdir'");
 }
 
-=head2 wait_for_guestregister
+=head2 wait_for_guestregister_chk
```
pdostal, line 263: This does not belong here.

lib/publiccloud/instance.pm
```diff
@@ -255,19 +260,26 @@ sub upload_log {
...
+C<wait_for_guestregister_chk> is called inside C<create_instance()>, enabled by C<check_guestregister>
 =cut
 
 sub wait_for_guestregister {
+    my ($self, %args) = @_;
+    my $out = $self->run_ssh_command(cmd => 'sudo systemctl is-active guestregister', proceed_on_failure => 1, quiet => 1);
+    return;
+}
+
+sub wait_for_guestregister_chk {
```
pdostal, line 282: Can you documment this subroutine? I don't know what `_chk` means.

lib/publiccloud/instance.pm
```diff
@@ -477,4 +491,179 @@ sub network_speed_test() {
...
+    $ret->{kernel_release} = $instance->run_ssh_command(cmd => 'uname -r', proceed_on_failure => 1);
+    $ret->{kernel_version} = $instance->run_ssh_command(cmd => 'uname -v', proceed_on_failure => 1);
+
+    # Do logging to openqa UI
+    $Data::Dumper::Sortkeys = 1;
+    record_info("RESULTS", Dumper($ret));
+
+    my $logs = '/var/log/cloudregister /var/log/cloud-init.log /var/log/cloud-init-output.log /var/log/messages /var/log/NetworkManager';
+    $instance->run_ssh_command(cmd => 'sudo tar -czvf /tmp/sle_cloud.tar.gz ' . $logs, proceed_on_failure => 1);
+    $instance->upload_log('/tmp/sle_cloud.tar.gz', failok => 1);
```
pdostal, line 537: Please upload those files one by one. Those archives are slower to review.

lib/publiccloud/instance.pm
```diff
@@ -477,4 +491,179 @@ sub network_speed_test() {
...
+    }
+
+    unless ($output && (time() - $start_time < $args{timeout})) {
+        record_info("WARN", "Unable to get system-analyze in $args{timeout} seconds", result => 'fail');
+        # handle_boot_failure: soft exit from measurement.
+        return (0, 0);
+    }
+    push @ret, extract_analyze_time($output);
+
+    $output = $instance->run_ssh_command(cmd => 'systemd-analyze blame', proceed_on_failure => 1);
```
pdostal, line 662: This is already on line 651.

m-dati:
VR for non regression of boottime:
[3749](http://10.168.4.117/tests/3749)

lib/publiccloud/instance.pm
m-dati: Sorry, but those files go  into a **tar.gz** that we'll find into the final logs, faster to check then.

That file list was so long already in the original version of [boottime.pm](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/322959a47faef1309f3c4d8395b624eb866f5f2c/tests/publiccloud/boottime.pm#L345) ,before my changes, all together inline into that run_ssh_command. In this new version I only moved that list to the `logs` var, to improve readiness.

To upload those files one-by-one, it means to provide that same list into an additional loop or a map, with no improvement.

lib/publiccloud/instance.pm
m-dati: Yes sorry, I forgot to clarify it, we discussed this change, during our meeting last Mar 10.

For this poo 118027 about boottime measure triggered in `provider::create_instances`, we need the _boot finished_ and _guest register checked_ before measurement start.

Given that in almost all modules where `create_instances` used it is soon  followed by a call to `wait_for_guestregister`, we agreed to move this last function inside of `it`.

Now as `first phase,` to soon verify the new internal location and reduce the changes of deleting all external 'wait_for_guestregister' calls, I "renamed" it for the internal call as `wait_for_guestregister_chk` and the _old_ `wait_for_guestregister`  changed to perform a simple display data where still present as external call.

m-dati: Also  I added to `create_instance` a more input paramenter `check_guestregister`, to enable-1/disable-0 that  _internal_ function call where not needed, like in the 4 modules here also updated with `create_instance(check_guestregister => 0)`

lib/publiccloud/instance.pm
m-dati: See previous comment  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16577#discussion_r1136998017

lib/publiccloud/instance.pm
m-dati: No, the function logic collect 2 times the data, first from systemd-analyze time and push into the @ret array, then for systemd-analyze blame, and append/push to the same @ret array. 
The result is visible in the test [RESULTS](http://10.168.4.117/tests/3744#step/run_ltp/161)

m-dati:
As non-regression of `slem_basic` updated module, like in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16577#discussion_r1137448258, this VR 
[3753](http://10.168.4.117/tests/3753) PASS

lib/publiccloud/instance.pm
pdostal: Yes, that's true. You can do:
```perl
my @logs = ('/var/log/x', '/var/log/y', '/var/log/z');
upload_logs($_) each @logs;
```

m-dati: Ok thanks, or even so :
```
my @logs = qw( /var/log/x   /var/log/y   /var/log/z );
upload_logs($_) each @logs;
```
But, without a containing **tar.gz**, at end of test we have that long number of different files, also unrelated even being collected all in that same point.

pdostal: Yes, please keep the "long number of different files" part :smile:
The openQA has cleanup scripts and as long as the files are properly named it's easier to open them directly than downloading an archive, decompressing it and what not...

m-dati: > Yes, please keep the "long number of different files" part

Well, will do it, but the same then we'll have to _download_ those `*.log` named files for reading.
Openqa UI only open .txt or .json.

m-dati: Yes, please keep the "long number of different files" part

Well, I prefer the _tar_, but I will do it. Anyway the same we'll have to download those *.log named files for reading, because Openqa UI seems to only open .txt or .json files., like [that](https://openqa.suse.de/tests/10611783#downloads)

m-dati: > Yes, please keep the "long number of different files" part

Well, I'd have no problem to handle the _tar_ as a log. Anyway, then, the same we'll have to download those single *.log named files for reading, because Openqa UI seems to only open .txt or .json files., like [that](https://openqa.suse.de/tests/10611783#downloads)

lib/publiccloud/instance.pm
asmorodskyi: nevertheless you edited docstring which still belongs to `wait_for_guestregister` so you need to move this documentation below `wait_for_guestregister_chk`  . I assume Pavel meant that

m-dati: yes, will do it

m-dati:
In last instance.pm update, in addition to previous changes:
`wait_for_guestregister` moved and documented, some `upload_logs` calls updated.
VR ongoing.

lib/publiccloud/instance.pm
m-dati: done

lib/publiccloud/instance.pm
m-dati: done

lib/publiccloud/provider.pm
m-dati: `create_instance()` call's input parameter updated in those modules where it does _not_ followed by _wait_for_guestregister_ call.

lib/publiccloud/instance.pm
```diff
@@ -477,4 +482,177 @@ sub network_speed_test() {
...
+    my @ret;
+
+    # calling syastemd-analyze time & blame
+    while ($output !~ /Startup finished in/ && time() - $start_time < $args{timeout}) {
+        $output = $instance->run_ssh_command(cmd => 'systemd-analyze time', proceed_on_failure => 1);
+        sleep 5;
+    }
+
+    unless ($output && (time() - $start_time < $args{timeout})) {
+        record_info("WARN", "Unable to get system-analyze in $args{timeout} seconds", result => 'fail');
```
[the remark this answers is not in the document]
m-dati: In case of errors here, soft exit from measurement and continue:
`        return (0, 0);`

tests/jeos/prepare_openstack.pm
```diff
@@ -16,7 +16,7 @@ sub run {
     select_host_console();
 
     my $provider = $self->provider_factory();
-    my $instance = $provider->create_instance();
+    my $instance = $provider->create_instance(check_guestregister => 0);
```
asmorodskyi, line 19: nice ! I like that !

asmorodskyi: @mloviska  , @jlausuch  - FYI

tests/publiccloud/boottime.pm
```diff
@@ -308,11 +308,11 @@ sub measure_timings {
         $provider = $args->{my_provider};
     } else {
         $provider = $self->provider_factory();
-        $instance = $self->{my_instance} = $provider->create_instance(check_connectivity => 0);
+        $instance = $self->{my_instance} = $provider->create_instance();
     }
 
-    $ret->{analyze}->{ssh_access} = $instance->wait_for_ssh(timeout => 300);
-    assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
+    #$ret->{analyze}->{ssh_access} = $instance->wait_for_ssh(timeout => 300);
```
asmorodskyi, line 314: just delete this line

tests/publiccloud/boottime.pm
```diff
@@ -308,11 +308,11 @@ sub measure_timings {
...
+    #assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
```
asmorodskyi, line 315: why you commented this one ?

tests/publiccloud/slem_basic.pm
```diff
@@ -21,7 +21,7 @@ sub run {
     select_serial_terminal();
     my $provider = $self->provider_factory();
     $provider->{username} = 'suse';
-    my $instance = $self->{my_instance} = $provider->create_instance();
+    my $instance = $self->{my_instance} = $provider->create_instance(check_guestregister => 0);
```
asmorodskyi, line 24: @jlausuch  FYI

m-dati: for that, [3755](http://10.168.4.117/tests/3755) |PASS| Micro-5-3 BYOS GCE / no-regression test

tests/publiccloud/boottime.pm
m-dati: yes, I forgot to remove it

asmorodskyi (approved):
LGTM

jlausuch:
I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.

tests/publiccloud/boottime.pm
m-dati: these 2 commented lines are already executed in create_instance, infact I also added to its call the parameter to enable internal wait_ssh

lib/publiccloud/instance.pm
```diff
@@ -477,4 +503,178 @@ sub network_speed_test() {
...
+    ($str_time) = split(/\r?\n/, $str_time, 2);
+    $str_time =~ s/Startup finished in\s*//;
+    $str_time =~ s/=(.+)$/+$1 (overall)/;
+    for my $time (split(/\s*\+\s*/, $str_time)) {
+        $time = trim($time);
+        my ($time, $type) = $time =~ /^(.+)\s*\((\w+)\)$/;
+        $res->{$type} = systemd_time_to_second($time);
+        return 0 if ($res->{$type} == -1);
+    }
+    map { return 0 unless exists($res->{$_}) } qw(kernel initrd userspace overall);
```
mloviska, line 636: as on L#548, I consider the usage of `map` as rather a bit odd to me.
Those usages are rather a check, whereas `map` is mostly handy to manipulate data.
IMHO I would rather go for a classic loop.

m-dati:
> I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.

Sorry but this way I brak all the planned job needing to change many (7) modules where this check is exectued

asmorodskyi:
> I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.

why ? looking into current state of PR Maurizio tweaked just openstack , slem and sapha . In we will revert logic he will need to tweak two times more places ...

asmorodskyi:
> > I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.
> 
> Sorry, it is much more used =1 than not used =0, so this way I'd break all my planned job to reduce changes, needing to change many (7) modules where this check is exectued
> 
> abh img_proof migration prepare_inst run_ltp storage_perf boottime ...
> 
> And repeat test for each of them.
> 
> But let me know.

+1

lib/publiccloud/instance.pm
m-dati: Sorry I  don't agree: why to increase the code, instead of reducing it using languages smart solutions? `map` isn't any virus :)  and the loop increases the code complexity for a simple log file transfer.

m-dati: As more clarification, this new code is a change, in order to satisfy a request to upload each log file, instead of the previous  *.tar.gz containing all the files

mloviska: I am not against `map`, it is definitely a powerful feature, but it feels like a misuse as you could not really fully benefit from it here.
Also, code wise, loop is just more readable here as well. But that is a matter of taste ofc

asmorodskyi: I don't have strong opinion here ... I would say that I am closer to Martin's position that map is misused in this case . But also I don't see any importance in this decision . Whether it will be map or for loop does not change much . So let's use golden rule **"the one who do decides"** and leave it to @m-dati  . Deal ?

mloviska: No objects, it just got my attention, therefore I have raised the point

m-dati: Sorry, I am lerner, not sure to get the point of _map misuse_ here, also looking in [documentation](https://perldoc.perl.org/functions/map) 
> map evaluates the BLOCK ...  for each element of LIST ... and composes a list of the results ...",

anyway I will change the code this way:
```
foreach (@LIST) { BLOCK };
```

m-dati: done

tests/publiccloud/boottime.pm
m-dati: Removed.

jlausuch:
> > > I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.
> > 
> > 
> > Sorry, it is much more used =1 than not used =0, so this way I'd break all my planned job to reduce changes, needing to change many (7) modules where this check is exectued
> > abh img_proof migration prepare_inst run_ltp storage_perf boottime ...
> > And repeat test for each of them.
> > But let me know.
> 
> +1

To me is more logical that `when you need a special feature, you specify it` than `we enable a special feature by default and if it's not needed, it should be specified`. The second option might lead people to miss that such feature exist and they are using it by default when calling that method. 
But if it's too much hassle, I'm ok with current approach.

asmorodskyi:
> To me is more logical that `when you need a special feature, you specify it` than `we enable a special feature by default and if it's not needed, it should be specified`. 

Fully agree with you but I think we have a different understanding what is "special feature " in that flow :)  . For me  the core of all Public Cloud related testing code is build around Amazon, Azure and Google cloud testing . Whatever this 3 guys need should be **default behavior ** of functions in provider/instance/basetest . Openstack / SLEM  are exceptions . And this is what current state of Maurizio's PR is showing - you getting `default behavior` ( wait for guest register ) without specifying anything and when you need `special feature` ( skip waiting for guest register which is absent in Openstack / SLEM )  than you need to pass something into the function . 

P.S. SAP/HA is just one big exception so I am keeping it out of discussion :)

m-dati:
> > > > I would make `check_guestregister` 0 by default and only specify `check_guestregister => 1` where it's needed (Azure, Amazon, Google), and leave it default to 0 to the others.
> > > 
> > > 
> > > Sorry, it is much more used =1 than not used =0, so this way I'd break all my planned job to reduce changes, needing to change many (7) modules where this check is exectued
> > > abh img_proof migration prepare_inst run_ltp storage_perf boottime ...
> > > And repeat test for each of them.
> > > But let me know.
> > 
> > 
> > +1
> 
> To me is more logical that `when you need a special feature, you specify it` than `we enable a special feature by default and if it's not needed, it should be specified`. The second option might lead people to miss that such feature exist and they are using it by default when calling that method. But if it's too much hassle, I'm ok with current approach.


IMHO this _guest register check_ should not be considered  _special_ feature but usual instead, as suggested by call frequence in our test modules. 
Moreover, let we consider that in `create_instance` also `wait_for_ssh()` is enabled by default and shall be disabled by a proper input parameter set=0, so I think: could an eventual logic invertion of the register check, also imply inversion for wait_ssh too?

jlausuch:
> > To me is more logical that `when you need a special feature, you specify it` than `we enable a special feature by default and if it's not needed, it should be specified`.
> 
> Fully agree with you but I think we have a different understanding what is "special feature " in that flow :) . For me the core of all Public Cloud related testing code is build around Amazon, Azure and Google cloud testing . Whatever this 3 guys need should be **default behavior ** of functions in provider/instance/basetest . Openstack / SLEM are exceptions . And this is what current state of Maurizio's PR is showing - you getting `default behavior` ( wait for guest register ) without specifying anything and when you need `special feature` ( skip waiting for guest register which is absent in Openstack / SLEM ) than you need to pass something into the function .
> 
> P.S. SAP/HA is just one big exception so I am keeping it out of discussion :)

ok, you convinced me :)

lib/publiccloud/provider.pm
m-dati: see i.e. this parameter's use in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16577#discussion_r1140070396

lib/publiccloud/instance.pm
```diff
@@ -487,4 +513,176 @@ sub network_speed_test() {
...
+        kernel_version => undef,
+        type => undef,
+        analyze => {},
+        blame => {},
+    };
+
+    record_info("BOOT TIME", 'systemd_analyze');
+    # first deployment analysis
+    my ($systemd_analyze, $systemd_blame) = do_systemd_analyze_time($instance);
+    return 0 unless ($systemd_analyze && $systemd_blame);
```
pdostal, line 542: Do we really want to ignore `do_systemd_analyze_time` failures?

lib/publiccloud/instance.pm
```diff
@@ -487,4 +513,176 @@ sub network_speed_test() {
...
+
+    # Collect kernel version
+    $ret->{kernel_release} = $instance->run_ssh_command(cmd => 'uname -r', proceed_on_failure => 1);
+    $ret->{kernel_version} = $instance->run_ssh_command(cmd => 'uname -v', proceed_on_failure => 1);
+
+    # Do logging to openqa UI
+    $Data::Dumper::Sortkeys = 1;
+    record_info("RESULTS", Dumper($ret));
+    $instance->upload_log("/var/log/" . $_, log_name => 'measure_boottime_' . $_ . '.txt', failok => 1)
+      foreach (qw(cloudregister cloud-init.log cloud-init-output.log messages NetworkManager));
```
pdostal, line 557:
```suggestion
    my @logfiles = qw(cloudregister cloud-init.log cloud-init-output.log messages NetworkManager);
    $instance->upload_log("/var/log/$_", log_name => 'measure_boottime_$_.txt', failok => 1) foreach (@logfiles);
```

pdostal (commented):
I still have two questions, but I'm positive we can merge it soon

lib/publiccloud/instance.pm
asmorodskyi: yes. 
1. Not all of them necessarily means something what will block further testing 
2. We do ignoring them at the moment , right ? ( By not actually checking this part on every instance creation ). So there is no change from this POV

pdostal (approved):

lib/publiccloud/instance.pm
m-dati: I already did so in a previous version of mine, but recently I noted that adding a new code line and a new parameter it didn't provide any improvement in reading and performance, against this way a single line command with the foreach loop seeming much clear with list in it.

m-dati: I'll do that change back.

m-dati: Your change proposal I already did in a previous version (using push -f it isn't visible anymore) but then, I decided to not to use the @log list anymore, leaving the data into the foreach, much more clear.
But now I am going to re-apply it.

pdostal (commented):
Just do the perl tidy and we can merge.

m-dati:
Tidy done

PR #16367, opened 2023-02-13
yast2 modules removal

DimStar77:
- openssl3: Accept "Engine" and "engine" output from openssl
  yast2_ncurses: no longer test yast2-{dhcp,dns,http-server

- Related tickets:
  * https://build.opensuse.org/request/show/1063904
  * https://build.opensuse.org/request/show/1063903
  * https://build.opensuse.org/request/show/1063901
- Needles: N/A
- Verification run: N/A

PR #16580, opened 2023-03-09
Create new yaml schedule copied from yast2_ncurses.yaml for leap

lemon-suse:
In Tumbleweed job group, the yast2_http,yast2-dns-server and yast2-dhcp-server will be removed from yast2_ncurses.yaml, to reserve these yast2 test in leap job group we create a new yaml schedule copied from yast2_ncurses.yaml for leap.

- Related ticket: https://progress.opensuse.org/issues/124433
- Related PR:  https://github.com/os-autoinst/opensuse-jobgroups/pull/285
- Needles: N/A
- Verification run: N/A (same content from yast2_ncurses.yaml)

### DIFF
--- a/schedule/yast/opensuse/yast2_ncurses/yast2_ncurses_leap.yaml
+++ b/schedule/yast/opensuse/yast2_ncurses/yast2_ncurses_leap.yaml
@@ -1,0 +1,38 @@
+---
+name:           yast2_ncurses_leap
+description:    >
+  Test for yast2 UI, ncurses only. Running on created gnome
+  images which provides both text console for ncurses UI tests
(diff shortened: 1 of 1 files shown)


rakoenig:
The copied file looks good, but shouldn't we remove the broken tests from the original Tumbleweed schedule?

lemon-suse:
> The copied file looks good, but shouldn't we remove the broken tests from the original Tumbleweed schedule?

I think they will remove that in this PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16367

rakoenig:
Ok, that makes sense then.

PR #16586, opened 2023-03-10
Soft failure 1195046

skliu:
Quick fix for https://openqa.suse.de/tests/10657979

**Verification run:**

- https://openqa.suse.de/tests/10658934

### DIFF
--- a/lib/migration.pm
+++ b/lib/migration.pm
@@ -35,6 +35,11 @@ our @EXPORT = qw(
 sub setup_sle {
     select_console 'root-console';
 
+    # https://bugzilla.suse.com/show_bug.cgi?id=1205290#c3
+    if (is_ppc64le && is_sle('<12-sp5')) {
+        systemctl('restart systemd-vconsole-setup.service');
+    }
+
     # Stop packagekitd
     if (is_sle('12+')) {
         quit_packagekit;


lib/migration.pm
```diff
@@ -35,6 +35,11 @@ our @EXPORT = qw(
 sub setup_sle {
     select_console 'root-console';
 
+    if (is_ppc64le && is_sle('<=12-sp5')) {
+        record_soft_failure('bsc#1195046 - ncurses display a wrong checker board character');
```
lemon-suse, line 39: For me, I think we can see this as a test requirement while not a workaround, so please remove the softfail.

schedule/yast/autoyast/autoyast_create_hdd_sdk_gnome.yaml
```diff
@@ -20,6 +20,7 @@ schedule:
   - '{{handle_reboot}}'
   - installation/first_boot
   - console/system_prepare
+  - '{{disable_screensaver}}'
```
lemon-suse, line 23: I think you make it mixed with another PR, please clean it.

lib/migration.pm
skliu: Done.

schedule/yast/autoyast/autoyast_create_hdd_sdk_gnome.yaml
skliu: Fixed, but I don't why there are two commits, I've rebased it.

lemon-suse (approved):

lib/migration.pm
jknphy: It was solved for sle 12 sp5, please keep the workaround for sle 12 sp4 only.
https://bugzilla.suse.com/show_bug.cgi?id=1205290#c3

skliu: Done.

PR #16590, opened 2023-03-10
Add helper functions for parsing zypper output

mdoucha:
Parsing zypper output is complicated because the output format changes in different SLE versions and also depending on command line options. Add helper functions that will return a list of hashes:
- `zypper_search()` - simplified and added support for calling without `--details` option
- `zypper_repos()`
- `zypper_patches()`

----

- Related ticket: N/A
- Needles: N/A
- Verification run:
  - SLE-11SP4: https://openqa.suse.de/tests/10690294
  - SLE-12SP2: https://openqa.suse.de/tests/10690295
  - SLE-12SP5: https://openqa.suse.de/tests/10690296
  - SLE-15SP1: https://openqa.suse.de/tests/10690297
  - SLE-15SP4: https://openqa.suse.de/tests/10690298


mdoucha:
The verification run uses this special test module to call each helper function with different `zypper` parameters and dump the parsed results into YAML:
https://github.com/mdoucha/os-autoinst-distri-opensuse/blob/zypper_utils_vr/tests/selftest/zypper.pm

The YAML output in info boxes must match the columns in console output.

mimi1vx (approved):

lib/utils.pm
```diff
@@ -831,39 +834,128 @@ sub minimal_patch_system {
...
+sub zypper_search {
+    my $params = shift;
+    my %opts;
+    my @fields = ('status', 'name', 'summary', 'type');
+
+    # Set Getopt to ignore any unrecognized options
+    Getopt::Long::Configure('bundling', 'pass_through', 'permute');
+
+    # Call in array context to silence warnings about extra options and args
+    my @tmp = GetOptionsFromString($params, \%opts, 'details|s', 'verbose|v');
```
b10n1k, line 887: `@tmp` is not used, right? Do you really need the assignment?

mdoucha: As the comment above says, I need to call `GetOptionsFromString()` in array context, otherwise it'll print warnings about unknown options and leftover arguments into log. Do you have any better idea than assigning to an unused variable?

b10n1k: i read that. but i dont think you need the assignment. Sorry if it is kinda picky, i just thought to bring it up. but no problem if you want to keep it as it is

b10n1k (approved):

lib/utils.pm
mdoucha: Well, if there is no explicit switch to array context, the function will be called in scalar context and print warnings. I need either the assignment or something else that forces array context. Do you have any suggestion what that "something else" could be?

lib/utils.pm
```diff
@@ -831,39 +834,128 @@ sub minimal_patch_system {
...
+of hashes. Pass the ASCII table in C<$table> as a single string. Pass an array
+of hash keys in C<$fields>. Keys will be mapped to columns by position.
 
 =cut
 
-sub zypper_search {
-    my $params = shift;
-    my @fields = ('status', 'name', 'type', 'version', 'arch', 'repository');
+sub parse_zypper_table {
+    (my $table, my $fields) = @_;
```
Martchus, line 848:
```suggestion
    (my $table, $fields) = @_;
```

lib/utils.pm
```diff
@@ -831,39 +834,128 @@ sub minimal_patch_system {
...
+        $line =~ s/^\s*(.*?)\s*$/$1/;
         my @tokens = split /\s*\|\s*/, $line;
-        next if $#tokens < $#fields;
+        next if $#tokens < $#$fields;
         my %tmp;
 
-        for (my $i = 0; $i < scalar @fields; $i++) {
-            $tmp{$fields[$i]} = $tokens[$i];
+        for (my $i = 0; $i < scalar @$fields; $i++) {
+            $tmp{$$fields[$i]} = $tokens[$i];
```
Martchus, line 858:
```suggestion
        for (my $i = 0; $i < scalar @$fields; $i++) {
            $tmp{$fields->[$i]} = $tokens[$i];
```

Martchus (commented):
It would be nice to add basic unit tests. Maybe you can also make use of function signatures in new code. Have you considered using `--xmlout`? At least for some commands (*not* status) it provides easier to parse output.

pevik (approved):

lib/utils.pm
mdoucha: Fixed.

mdoucha:
> It would be nice to add basic unit tests. Maybe you can also make use of function signatures in new code. Have you considered using `--xmlout`? At least for some commands (_not_ status) it provides easier to parse output.

I've tried using `--xmlout` but it's too broken to be used in any of the helper functions. https://bugzilla.suse.com/show_bug.cgi?id=1209025

The only part that could use some unit tests is the `parse_zypper_table()` function but it's intentionally not exported so it cannot be called outside `utils.pm` itself. The exported helper functions depend on `zypper` behavior which can change in the future and any changes need to be verified manually.

Verification runs updated.

Martchus:
You can always mock functions in Perl. It may not be worth the effort but just as a reference, this is how it could be done without even having to load the regular test API at all: https://github.com/os-autoinst/os-autoinst-wheel-launcher/commit/23806faf84312b7ca955c75bd5e967a7824ed886

mdoucha:
What would be the point of mocking the very function you're trying to test?

Martchus:
You'd only mock the test API functions, not the functions you want to test. Of course this way you're not testing the integration with the test API and only your code. So you'd test that your code behaves correctly under certain assumptions about the test API (which can still be wrong of course).

foursixnine (approved):

lib/utils.pm
```diff
@@ -831,39 +834,128 @@ sub minimal_patch_system {
...
+=head2 parse_zypper_table
 
- zypper_search($search_params);
+ parse_zypper_table($table, $fields);
 
-Run zypper search with given command line arguments and parse the output into
-an array of hashes.
+Parse ASCII table generated by commands such as C<zypper search> into an array
+of hashes. Pass the ASCII table in C<$table> as a single string. Pass an array
+of hash keys in C<$fields>. Keys will be mapped to columns by position.
```
foursixnine, line 843: It's sad that the machine format of zypper it's only a wrapper for the text output of it :/

mdoucha: Zypper also has `--xmlout` but unfortunately it's too broken. `zypper --xmlout search -s` does not report package retractions, `zypper --xmlout patches` prints the usual ASCII table wrapped in XML tags... Even if it gets fixed, we'll probably have to wait until everything before 12SP5 and 15SP4 reaches EOL.

PR #16601, opened 2023-03-13
Switch to textmode to avoid first_boot fail

realcharmer:
Set `DESKTOP: textmode` to avoid a fail in `first_boot` module.

- Related ticket: https://progress.opensuse.org/issues/124931
- Verification run: http://emiler-openqa.qe.suse.de/tests/156


paolostivanin:
LGTM

PR #16602, opened 2023-03-13
epiphany: handle the new 'set as default browser' dialog

DimStar77:
- Related ticket: N/A (identified as part of GNOME Next live testing)
- Needles: Created on o3 during VR
- Verification run: https://openqa.opensuse.org/tests/3171206


zaitor:
+1

PR #16606, opened 2023-03-13
Use installation/bootloader_start instead of own conditional schedule

paolostivanin:
- Verification runs: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=20230313-1&groupid=431

### DIFF
--- a/schedule/security/apparmor.yaml
+++ b/schedule/security/apparmor.yaml
@@ -2,8 +2,7 @@ name: apparmor
 description:    >
     This is for the apparmor and mau-apparmor tests.
 schedule:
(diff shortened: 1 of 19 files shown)


okurz:
Thank you for cleaning up. I don't know what motivated people to the previous solution.

paolostivanin:
Thanks @okurz ! To be honest, those abominations were my fault :see_no_evil: when yesterday I saw that there was a better way to do the same thing, I immediately changed the code and opened this pr.

ilmanzo:
good job!

PR #16607, opened 2023-03-13
Filter only CRITICAL and ERROR sections

grisu48:
Filter only CRITICAL and ERROR sections, instead of word. This prevents a false positive when reporting the error module has been loaded.

- Related ticket: https://progress.opensuse.org/issues/125834
- Verification run: https://duck-norris.qe.suse.de/tests/12355


michaelgrifalconi (approved):
LGTM!

PR #16612, opened 2023-03-13
Re-write and exclude small part  because of performance issue on aarch64

Zaoliang:
we have performance issue on aarch64 with less RAM and poor host cpu.
It fails quite often for starting qt5 application and matching needles.
So skip these steps and check the package installation and compile app on aarch64.     

see https://progress.opensuse.org/issues/125873

### DIFF
--- a/tests/x11/libqt5_qtbase.pm
+++ b/tests/x11/libqt5_qtbase.pm
@@ -20,6 +20,7 @@ use utils;
 use x11utils 'ensure_unlocked_desktop';
 use version_utils 'is_sle';
 use registration qw(cleanup_registration register_product add_suseconnect_product get_addon_fullname remove_suseconnect_product);
+use Utils::Architectures 'is_aarch64';
 
 sub run {
     if (is_sle('>=15-sp4')) {
@@ -30,26 +31,25 @@ sub run {
 
     select_console('x11');
     ensure_installed("libqt5-qttools yast2-installation", timeout => 180);
+    # performance issue (see poo#125873) impacts only aarch64 workers on OSD
+    designer_qt5() unless (is_aarch64 && is_sle);
+    release_notes() unless (is_aarch64 && is_sle);
+    compile_qt5();
+}
 
-    # Test designer-qt5
-    x11_start_program('designer-qt5');
-    wait_still_screen 3;
-    assert_and_click("designer-qt5-start");
-    wait_still_screen 3;
-    assert_screen("designer-qt5-main");
+sub designer_qt5 {
+    my $timeout = '90';
+    x11_start_program("designer-qt5", target_match => "designer-qt5-start", match_timeout => $timeout);
+    assert_and_click("designer-qt5-start", timeout => $timeout);
+    assert_screen("designer-qt5-main", $timeout);
     wait_screen_change { send_key "ctrl-r" };    # run the design preview
-    assert_screen("designer-qt5-preview");
+    assert_screen("designer-qt5-preview", $timeout);
     send_key "alt-f4";    # close preview
     send_key "alt-f4";    # close program
+}
 
-    # Test release notes
-    x11_start_program('/usr/sbin/yast2 inst_release_notes', target_match => 'inst_release_notes');
-    send_key "alt-l";    # make sure it was closed ('Close' button shortcut)
-    send_key "alt-f4";    # close program
-
-    # Compile an application and run it, check that exits with 0
+sub compile_qt5 {
     ensure_installed "gcc gcc-c++ libQt5Core-devel libQt5Gui-devel libQt5Network-devel libQt5Widgets-devel", timeout => 400;
-
     x11_start_program('xterm');
     assert_script_run 'cd data';
     assert_script_run 'tar xvf libqt5-qtbase.tar.gz';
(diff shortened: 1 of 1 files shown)


Zaoliang:
https://openqa.suse.de/tests/10689186#step/libqt5_qtbase/24 shows no issue with code changes, will provide statics for more test runs later.

Zaoliang:
https://openqa.suse.de/tests/10699243 looks good on x86_64

Zaoliang:
successful test run after re-write this module:
https://openqa.suse.de/tests/10702422

Zaoliang:
VT for tumbleweed:
http://10.161.8.50/tests/14 (kde)
http://10.161.8.50/tests/15 (gnome)

tests/x11/libqt5_qtbase.pm
```diff
@@ -30,26 +31,31 @@ sub run {
 
     select_console('x11');
     ensure_installed("libqt5-qttools yast2-installation", timeout => 180);
+    # we have at moment performance issue on aarch64 OSD only, so no need to change this for tumbleweed
+    # see poo#125873 for more details
+    if (is_aarch64 && is_sle) {
+        compile_qt5();
+    }
+    elsif (is_x86_64 || is_s390x || is_ppc64le || (is_aarch64 && is_tumbleweed)) {
```
grisu48, line 39: why `elsif` and not a simple `else`?

tests/x11/libqt5_qtbase.pm
```diff
@@ -30,26 +31,31 @@ sub run {
...
+        designer_qt5();
```
grisu48, line 40: How  about changing the whole `if` block to something like

```
designer_qt5() unless (is_aarch64 && is_sle);
release_notes() unless (is_aarch64 && is_sle);
compile_qt5()
```

Would make it easier to understand and has the same effect. WDYT?

tests/x11/libqt5_qtbase.pm
```diff
@@ -30,26 +31,31 @@ sub run {
...
 
-    # Test designer-qt5
-    x11_start_program('designer-qt5');
-    wait_still_screen 3;
-    assert_and_click("designer-qt5-start");
-    wait_still_screen 3;
-    assert_screen("designer-qt5-main");
+sub designer_qt5 {
+    my $timeout = '90';
+    x11_start_program("designer-qt5", target_match => "designer-qt5-start", match_timeout => $timeout);
```
grisu48, line 48: Looking at poo#125873 and especially at https://openqa.suse.de/tests/10676638#step/libqt5_qtbase/25 I'm wondering, if increasing the timeout alone wouldn't be already enough?

tests/x11/libqt5_qtbase.pm
```diff
@@ -30,26 +31,31 @@ sub run {
 
     select_console('x11');
     ensure_installed("libqt5-qttools yast2-installation", timeout => 180);
+    # we have at moment performance issue on aarch64 OSD only, so no need to change this for tumbleweed
```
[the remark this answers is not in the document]
Zaoliang: good, will update this

tests/x11/libqt5_qtbase.pm
Zaoliang: because to be sure not break other test: for example is_aarch64 && is_tumbleweed should still run the same test as before.

Zaoliang: okay, since I already define is_aarch64 && is_sle, I can use else here, should be safe.

tests/x11/libqt5_qtbase.pm
Zaoliang: yes, I like this better :)

tests/x11/libqt5_qtbase.pm
grisu48: ok

Zaoliang:
check again with updated code changes:
https://openqa.suse.de/tests/10709152
https://openqa.suse.de/tests/10709151

tests/x11/libqt5_qtbase.pm
```diff
@@ -30,26 +31,25 @@ sub run {
 
     select_console('x11');
     ensure_installed("libqt5-qttools yast2-installation", timeout => 180);
+    # performance issue (see poo#125873) impacts only aarch64 workers on OSD
+    designer_qt5() unless (is_aarch64 && is_sle);
+    release_notes() unless (is_aarch64 && is_sle);
```
rfan1, line 36: Did we miss any test coverage on aarch64 platform with this change?

grisu48: Yes of course, the whole qt5 designer path is omitted.

If possible I would keep the `release_notes()` test. For the QT5 designer part .... I see that's a painful test run and the number of people who run qt5designer on aarch64 is probably rather low.

Let's keep the `release_notes()` though?

Zaoliang: who wants do this on aarch64 and s390x in real life?
release_notes is not relevant to openSUSE anyway, but you know I don't want drop this as long as it works...
by the way, just start and close application is too less for function test actually, but better than nothing.

grisu48: It's a smoke test for the qt bindings in YaST as such I would keep them, unless you think they are a major pain. I leave the decision to you, the PR is IMHO good as it is.

grisu48 (approved):

rfan1 (approved):
LGTM

PR #16621, opened 2023-03-14
Handle conflicts in size monitoring

mloviska:
If there is already existing record for the image, we can safely assume that all related records have been pushed before.


jlausuch (approved):
Neat!

PR #16622, opened 2023-03-14
Implement test for HPC warewulf4

b10n1k:
A new test which will run one controller node and two compute nodes without
supportserver. An important aspect of this test is the controller VM which
requires two network interfaces. The compute nodes should run on this network
once they boot.
    
- Related ticket: https://jira.suse.com/browse/PED-3033
- Needles: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1639
- Verification run: https://aquarius.suse.cz/tests/16376#

### DIFF
--- a/schedule/hpc/ww4.yaml
+++ b/schedule/hpc/ww4.yaml
@@ -1,0 +1,30 @@
+---
+name: Warewulf4
+description:    >
+    Maintainer: kernel-qa
+    Unified scheduler for warewulf4 multimachine tests
+    '{{pxe}}' runs the scheduler for compute nodes
+    '{{ww4}}' schedules the controller
+vars:
+  DESKTOP: textmode
+  BOOT_HDD_IMAGE: 1
+
+conditional_schedule:
+  bootmenu:
+    ARCH:
+      aarch64:
+  ww4:
+    HPC:
+      ww4_controller:
+        - hpc/barrier_init
+        - support_server/login
+        - hpc/before_test
+        - hpc/ww4
+  pxe:
+    HPC:
+      ww4_compute:
+        - hpc/ww4_await_pxe_install
+schedule:
+  - '{{bootmenu}}'
+  - '{{pxe}}'
+  - '{{ww4}}'
--- a/tests/hpc/ww4.pm
+++ b/tests/hpc/ww4.pm
@@ -1,0 +1,147 @@
+# SUSE's openQA tests
+#
+# Copyright SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Configure and run a warewulf4 controller
+# Maintainer: Kernel QE <kernel-qa@suse.de>
+
+use Mojo::Base qw(hpcbase hpc::utils), -signatures;
+use testapi;
+use serial_terminal qw(select_serial_terminal);
+use lockapi;
+use utils;
+use Utils::Logging 'export_logs';
+use hpc::formatter;
+use isotovideo;
+use mm_tests;
+use Utils::Logging 'save_and_upload_log';
+use POSIX 'strftime';
+
+our $file = 'tmpresults.xml';
+
+sub run ($self) {
+    select_serial_terminal();
+    configure_static_network('10.0.2.1/24');
+
+    my $user_virtio_fixed = isotovideo::get_version() >= 35;
+    my $prompt = $user_virtio_fixed ? $testapi::username . '@' . get_required_var('HOSTNAME') . ':~> ' : undef;
+
+    ensure_ca_certificates_suse_installed();
+    my $rt = zypper_call("in warewulf4");
+    test_case('Installation', 'ww4', $rt);
+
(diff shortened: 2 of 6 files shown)


schedule/hpc/ww4.yaml
```diff
@@ -0,0 +1,30 @@
+---
+name: Warwulf4
```
czerw, line 2: typo: Warwulf

tests/hpc/ww4_await_pxe_install.pm
```diff
@@ -0,0 +1,48 @@
...
+use POSIX 'strftime';
+
+sub run ($self) {
+    my $i = 0;
+
+    while ($i < 10) {
+        check_screen('pxe-start');
+        record_info "pxe0", "pxe0";
+        sleep(1);
+        send_key "ctrl-B";
```
czerw, line 25: Wouldn't be better to use `send_key_until_needlematch`? To avoid combination of `check_screen`, `sleep` and `send_key`.

tests/hpc/ww4_await_pxe_install.pm
```diff
@@ -0,0 +1,48 @@
...
+        sleep(20);
```
czerw, line 26: Direct `sleep` use is usually bad sign and should be avoided. I think multimachine barrier can be inserted before reboot and we can wait for controller.

tests/hpc/ww4_await_pxe_install.pm
```diff
@@ -0,0 +1,48 @@
...
+        save_screenshot();
+        $i++;
+    }
+    wait_still_screen 5;
+    while (1) {
+        unless (check_screen('ww4-ready')) {
+            record_info('Booting', '');
+            next;
+        }
+        last;
```
czerw, line 37: What is the purpose of `wait_still_screen` and above booting loop? `assert_screen 'ww4-ready'` would work as well, as you just wait for https://aquarius.suse.cz/tests/16377#step/ww4_await_pxe_install/41

tests/hpc/ww4.pm
```diff
@@ -0,0 +1,150 @@
...
+use mm_tests;
+use Utils::Logging 'save_and_upload_log';
+use POSIX 'strftime';
+
+our $file = 'tmpresults.xml';
+
+sub run ($self) {
+    select_serial_terminal();
+    configure_static_network('10.0.2.1/24');
+
```
mslacken, line 26: is this the network for eth0 or eth1?

tests/hpc/ww4.pm
```diff
@@ -0,0 +1,150 @@
...
+    my $user_virtio_fixed = isotovideo::get_version() >= 35;
+    my $prompt = $user_virtio_fixed ? $testapi::username . '@' . get_required_var('HOSTNAME') . ':~> ' : undef;
+
+    ensure_ca_certificates_suse_installed();
+    my $rt = zypper_call("in warewulf4");
+    test_case('Installation', 'ww4', $rt);
+
+    $rt = systemctl 'enable --now warewulfd';
+    test_case('systemd', 'ww4', $rt);
```
mslacken, line 35: start warewulfd after you edited warewulf.conf like you do in line 40 and following

mslacken (commented):
looks good to me
Having the right IP on the screen, just means that the host has booted, not that the network is setup in the right way, as /etc/issue is generated as a static file

tests/hpc/ww4.pm
b10n1k: this configures the eth0 and talks public

tests/hpc/ww4_await_pxe_install.pm
b10n1k: i can think of a proper way until now. `send_key_until_needlematch` doesnt seem to be the key tho. It couldnt handle the expected sequence of events. If `pxe-start` the send_key should be sent. Then wait a bit and reboot. The function, in the other side says, send some event and then match, which is of not help here. No? 
ofc i let it like that, but i am aware of the issue and i know that i should find a better way for this loop. but i dont have anything better at the moment and the implementation has taken longer than it was expected. Mainly i am saying, lets fix it in a follow up PR

tests/hpc/ww4_await_pxe_install.pm
b10n1k: I agree about the sleep. But i consider this a special case. Also look at my reply https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16622#discussion_r1182909512. the whole loop should be reworked in the near future.

schedule/hpc/ww4.yaml
b10n1k: done

tests/hpc/ww4_await_pxe_install.pm
b10n1k: removed

tests/hpc/ww4_await_pxe_install.pm
czerw: ok

czerw (approved):
LGTM

mslacken (approved):
Thanks, great work

PR #16624, opened 2023-03-14
Fix usage of validate_script_output

perlpunk:
I was just pointed to this code by @vtrubovics, and that it does not work as expected.

This should do what was oriinally intended.

Maybe someone familiar with this test can do a verification run?

http://open.qa/api/testapi/#_validate_script_output

validate_script_output can take a callback subroutine, where you can write your own code to match the output stored in `$_`. Alternatively it takes a regex object, which you can create with `qr//`.

If you use

    $something = 'm/regex/';

then the variable will just contain an ordinary string starting with the characters 'm' and a forward slash.

- Verification run: tbd

### DIFF
--- a/lib/openscaptest.pm
+++ b/lib/openscaptest.pm
@@ -59,7 +59,7 @@ sub validate_result {
         assert_script_run "xmllint --noout $xml_args $result_file";
     }
 
-    validate_script_output "cat $result_file", sub { $match }, timeout => 300;
+    validate_script_output "cat $result_file", $match, timeout => 300;
     upload_logs($result_file);
 }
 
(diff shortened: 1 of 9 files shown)


perlpunk:
I actually had to fix more usages of `openscaptest::_validate_result`. Will run another test

perlpunk:
https://openqa.suse.de/tests/10701002 looks good so far except https://openqa.suse.de/tests/10701002#step/oscap_generating_report/10
But that is because the regex doesn't match, specifically the `contains 2 rules` because there is only test with `contains 1...`.
I'm not the one who can fix that I think.
The last test is still  ongoing, not sure why it takes so long.

perlpunk:
Is anyone else taking this? Like I said this is a draft because I cannot fix the issue which was uncovered by this.

vtrubovics:
I will fix it later.

perlpunk:
This is still an issue. Currently the test will hide failures.

PR #16625, opened 2023-03-15
Support installing host with autoyast

Julie-CAO:
Put all commits in one PR because tests have to be verified in O3, downloading TW medium from download.opensuse.org always timeout and is not supported from O3 repo.

Autoyast profiles for KVM host has been done and for XEN host will available soon. but we need not wait for xen autoyast profiles, it will be in a seperate PR.

1. Support installing host with autoyast.  The normal path for host autoyast profiles, data/ or autoyast/ in openqa path, cannot be adoppted in O3 , because static ipxe installation command is used in O3. I commit the files at here but actually copy them manually to a location in O3, /var/lib/openqa/share/factory/other/autoyast_opensuse_kvm_sshd.xml. No installation.pm is used in O3 for the time being, it will be considered if it is implemented for SLE in the future.

2. Use sol console as it is used in other tests and it is said stable enough.

3. Determine modular libvirt by the version of libvirt package, libvirt 9.1.0. As Alice suggested, determine the version of libvirt by running command directly in SUT rather than by products and releases to prevent adaptions in the future. 

4. Provide a function to restart modular libvirt daemons for test owners to implement in their test in the future.

5. Enable debug logging for modular libvirt and output logs to seperate log files.

6. Install guests with medium from O3 repo. It is more stable to download kernel files from O3 repo than download.opensuse.org for guest. 

7.  Check guest more often as IP change after reboot. Unlike SLE15SP5 guests, the IP TW guest during installation is different with the one after guest reboot even with the identical MAC address. It took me a lot of time in this issue. I have tried configurations from both dhcp client side and server side, passing static IP to guest,  but all failed. Finally, I have to change test scripts to check guest IP more often. 

8. Remove "DISPLAY= " from yast.ssh install command. Installation with this system variable began to fail a few weeks ago. I don't find the reason.

- Related ticket: https://progress.opensuse.org/issues/124149 & https://progress.opensuse.org/issues/126092
- Verification run with the latest commits: 
[Passed TW kvm test](https://openqa.opensuse.org/tests/3284826#live) 
[Failed TW kvm test and logs uploaded](https://openqa.opensuse.org/tests/3282580)   // failed as expected and logs were uploaded in [virt_logs_all.tar.gz](https://openqa.opensuse.org/tests/3282580/file/virt_logs_all.tar.gz)
[Failed SLE 12sp5 xen host and libvirtd log uploaded](https://openqa.suse.de/tests/11109911)
[Passed SLE kvm test](https://openqa.suse.de/tests/11109919)

more verificat runs with previous commits
[uefi-gi-guest_developing-on-host_developing-kvm](http://openqa.suse.de/tests/11084741)  
[gi-guest_developing-on-host_developing-xen](https://openqa.suse.de/tests/11084731)  
[uefi-gi-guest_developing-on-host_sles15sp4-kvm](http://openqa.suse.de/t11083532)

### DIFF
--- a/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
@@ -1,0 +1,136 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <general>
+    <mode>
+      <confirm t="boolean">false</confirm>
+      <final_reboot t="boolean">true</final_reboot>
+      <second_stage t="boolean">false</second_stage>
+    </mode>
+    <signature-handling>
+      <accept_file_without_checksum t="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key t="boolean">true</accept_non_trusted_gpg_key>
+      <accept_unknown_gpg_key t="boolean">true</accept_unknown_gpg_key>
+      <accept_unsigned_file t="boolean">true</accept_unsigned_file>
+      <accept_verification_failed t="boolean">true</accept_verification_failed>
+      <import_gpg_key t="boolean">true</import_gpg_key>
+    </signature-handling>
+  </general>
+  <report>
+    <errors>
+      <show t="boolean">true</show>
+      <timeout t="integer">20</timeout>
+      <log t="boolean">true</log>
+    </errors>
+    <messages>
+      <show t="boolean">true</show>
+      <timeout t="integer">5</timeout>
+      <log t="boolean">true</log>
+    </messages>
+    <warnings>
+      <show t="boolean">true</show>
+      <timeout t="integer">10</timeout>
+      <log t="boolean">true</log>
+    </warnings>
+    <yesno_messages>
+      <show t="boolean">true</show>
+      <timeout t="integer">30</timeout>
+      <log t="boolean">true</log>
+    </yesno_messages>
+  </report>
+  <bootloader>
+    <loader_type>grub2</loader_type>
+    <global>
+      <activate>true</activate>
+      <boot_mbr>true</boot_mbr>
+      <gfxmode>auto</gfxmode>
+      <os_prober>true</os_prober>
+      <terminal>console serial</terminal>
+      <serial>serial --unit=1 --speed=115200 --word=8 --parity=no</serial>
+      <timeout t="integer">30</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>console=com2,115200 loglvl=all guest_loglvl=all sync_console</xen_kernel_append>
+      <xen_append>console=hvc0,115200 console=tty loglevel=5</xen_append>
+    </global>
+  </bootloader>
+  <partitioning t="list">
+    <drive>
+      <device>/dev/sda</device>
+      <initialize t="boolean">true</initialize>
+      <use>all</use>
+      <enable_snapshots t="boolean">false</enable_snapshots>
+    </drive>
+  </partitioning>
+  <timezone>
+    <hwclock>UTC</hwclock>
+    <timezone>America/New_York</timezone>
+  </timezone>
+  <networking>
+    <keep_install_network t="boolean">true</keep_install_network>
+    <backend>network_manager</backend>
+  </networking>
+  <software>
+    <products t="list">
+      <product>openSUSE</product>
+    </products>
+    <install_recommended t="boolean">true</install_recommended>
+    <patterns t="list">
+      <pattern>base</pattern>
+      <pattern>basesystem</pattern>
+      <pattern>enhanced_base</pattern>
+      <pattern>kvm_server</pattern>
+      <pattern>kvm_tools</pattern>
+    </patterns>
+    <packages t="list">
+      <package>autoyast2</package>
+      <package>supportutils</package>
+      <package>nmap</package>
+    </packages>
+  </software>
+  <users t="list">
+    <user>
+      <username>bernhard</username>
+      <user_password>$y$j9T$xsY9ubdI0jRF5JoSV8vvg1$03I5jIewyiCMm/EmkgRtYxvEYm83ZSJo0tl1rKP8Ad.</user_password>
+      <fullname>Bernhard M. Wiedemann</fullname>  
+      <encrypted t="boolean">true</encrypted>
+    </user>
+    <user>
+      <username>root</username>
+      <user_password>$y$j9T$Bc3oHTMRcmYxyzTlxEd7A/$jU0KChUccTna2tH9ItMRyRHC9fofibG37RGH4XkCxYD</user_password>
+      <encrypted t="boolean">true</encrypted>
+    </user>
+  </users>
+  <services-manager>
+    <default_target>multi-user</default_target>
+    <services>
+      <disable t="list">
+        <service>firewalld</service>
+      </disable>
+      <enable t="list">
+        <service>sshd</service>
+	<service>virtqemud</service>
+	<service>virtstoraged</service>
+	<service>virtnetworkd</service>
+	<service>virtnodedevd</service>
+	<service>virtsecretd</service>
+	<service>virtproxyd</service>
+      </enable>
+    </services>
+  </services-manager>
+  <scripts>
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- keep ssh session alive -->
+    <!-- /etc/ssh/sshd_config is not default installed in TW -->
+    <init-scripts t="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config.d/permit_root_login.conf"
+echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
+sshd_config_file="/etc/ssh/sshd_config.d/keep_alive.conf"
+echo -e "TCPKeepAlive yes\nClientAliveInterval 60\nClientAliveCountMax 480" > $sshd_config_file
+]]>
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
+</profile>
--- a/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
@@ -114,6 +114,7 @@
 	<service>virtnodedevd</service>
 	<service>virtsecretd</service>
 	<service>virtproxyd</service>
+	<service>virtnwfilterd</service>
       </enable>
     </services>
   </services-manager>
--- a/data/virt_autotest/virt_logs_collector.sh
+++ b/data/virt_autotest/virt_logs_collector.sh
@@ -2,6 +2,18 @@
 set -o pipefail
 shopt -s nocasematch
 
+# Get Product name
+# Product name are "SLES", "openSUSE Tumbleweed", "openSUSE Leap"
+# TBD: ALP, ...
+function get_product_name() {
+	local version_file=$1
+	if [[ -z ${version_file} ]];then
+	    version_file="/etc/os-release"
+	fi
+	local product_name=`cat /etc/os-release | grep ^NAME= | cut -d '"' -f2`
+	echo $product_name
+}
+
 #Obtain SLES release version and service pack level
 function get_sles_release() {
         local query_type=$1
@@ -339,6 +351,7 @@ function collect_extra_logs_from_host() {
 	local libvirt_qemu_log="/var/lib/libvirt/qemu"
 	local libvirt_log="/var/log/libvirt"
 	local libvirtd_log="${libvirt_log}/libvirtd.log"
+	local libvirt_daemon_logs="${libvirt_log}/*d.log"
 	local xen_log="/var/log/xen"
 	local xen_boot_log="${xen_log}/xen-boot.log"
 
@@ -360,11 +373,13 @@ function collect_extra_logs_from_host() {
               cp --parent -f -r ${xen_boot_log} ${extra_logs_folder}
               ret_result=$(( ${ret_result} | $? ))
            fi           
-        elif [[ ${release} -eq 15 ]];then
+        elif [[ ${release} -eq 15 || `get_product_name` == "openSUSE Tumbleweed" ]];then
            cp --parent -f -r ${libvirt_boot_log} ${extra_logs_folder}
            ret_result=$(( ${ret_result} | $? ))
            cp --parent -f -r ${libvirt_qemu_log} ${extra_logs_folder}
            ret_result=$(( ${ret_result} | $? ))
+	   cp --parent -f -r ${libvirt_daemon_logs} ${extra_logs_folder}
+	   ret_result=$(( ${ret_result} | $? ))
            if [[ `get_sles_hypervisor` == "XEN" ]];then
               cp --parent -f -r ${xen_boot_log} ${extra_logs_folder}
               ret_result=$(( ${ret_result} | $? ))
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -27,6 +27,7 @@ use IO::Socket::INET;
 use Carp;
 
 our @EXPORT = qw(is_vmware_virtualization is_hyperv_virtualization is_fv_guest is_pv_guest guest_is_sle is_guest_ballooned is_xen_host is_kvm_host reset_log_cursor check_failures_in_journal check_host_health check_guest_health
+  is_monolithic_libvirtd turn_on_libvirt_debugging_log
   print_cmd_output_to_file ssh_setup ssh_copy_id create_guest import_guest install_default_packages upload_y2logs ensure_default_net_is_active ensure_guest_started
   ensure_online add_guest_to_hosts restart_libvirtd remove_additional_disks remove_additional_nic collect_virt_system_logs shutdown_guests wait_guest_online start_guests restore_downloaded_guests save_original_guest_xmls restore_original_guests
   is_guest_online wait_guests_shutdown remove_vm setup_common_ssh_config add_alias_in_ssh_config parse_subnet_address_ipv4 backup_file manage_system_service setup_rsyslog_host
@@ -55,8 +56,23 @@ sub restart_libvirtd {
     else {
         systemctl("restart libvirtd", timeout => 180);
     }
+    record_info("Libvirtd has been restarted!");
     save_screenshot;
-    record_info("Debug log for libvirtd has been enabled!");
+}
+
+# Usage: restart_modular_libvirt_daemons([daemon1_name daemon2_name ...]). For example:
+# to specify daemons which will be restarted: restart_modular_libvirt_daemons(virtqemud virtstoraged ...)
+# to restart all modular daemons without any daemons passed
+sub restart_modular_libvirt_daemons {
+    my @daemons = @_ == 0 ? qw(virtqemud virtstoraged virtnetworkd virtnodedevd virtsecretd virtproxyd virtnwfilterd) : split(" ", @_);
+    if (is_alp) {
+        record_soft_failure("Restaring modular libvirt daemons has not been implemented in ALP. See poo#129086");
+    }
+    else {
+        systemctl("restart $_\{,-ro,-admin\}.socket") foreach @daemons;
+        systemctl("restart $_.service") foreach @daemons;
+    }
+    record_info("Libvirt daemons restarted", join(' ', @daemons));
 }
 
 #return 1 if it is a VMware test judging by REGRESSION variable
@@ -114,6 +130,43 @@ sub is_guest_ballooned {
 #return 1 if test is expected to run on KVM hypervisor
 sub is_kvm_host {
     return check_var("SYSTEM_ROLE", "kvm") || check_var("HOST_HYPERVISOR", "kvm") || check_var("REGRESSION", "qemu-hypervisor");
+}
+
+#retrun 1 if libvirt 9.0- is running which monolithic libvirtd is the default service
+sub is_monolithic_libvirtd {
+    unless (is_alp) {
+        return 1 if script_run('rpm -q libvirt-libs | grep -e "libs-9\.0" -e "libs-[1-8]\."') == 0;
+    }
+    return 0;
+}
+
+# For legacy libvird, set debug level logging for libvirtd services
+# For modular libvirt, do the same settings to /etc/libvirt/virt{qemu,xen,driver}d.conf.
+# virt{qemu,xen}d daemons provide the most important libvirt log(sufficient for most issues).
+# virt{driver}.d daemons is only required by specific issues, eg virtual network failures may need virtnetworkd log.
+# But our automation is better to set them to collect more logs as we could as possible.
+# Developer asked to use different log file as log_output per daemon.
+sub turn_on_libvirt_debugging_log {
+
+    my @libvirt_daemons = is_monolithic_libvirtd ? "libvirtd" : qw(virtqemud virtstoraged virtnetworkd virtnodedevd virtsecretd virtproxyd virtnwfilterd virtlockd virtlogd);
+
+    #turn on debug and log filter for libvirt services
+    #set log_level = 1 'debug'
+    #the size of libvirtd with debug level and without any filter on sles15sp3 xen is over 100G,
+    #which consumes all the disk space. Now get comfirmation from virt developers,
+    #log filter is set to store component logs with different levels.
+    foreach my $daemon (@libvirt_daemons) {
+        my $conf_file = "/etc/libvirt/$daemon.conf";
+        if (script_run("ls $conf_file") == 0) {
+            script_run "sed -i '/^[# ]*log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $conf_file";
+            script_run "sed -i '/^[# ]*log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs = \"1:file:/var/log/libvirt/$daemon.log\"%};\${x;/^\$/{s%%log_outputs = \"1:file:/var/log/libvirt/$daemon.log\"%;H};x}' $conf_file";
+            script_run "sed -i '/^[# ]*log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters = \"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters = \"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $conf_file";
+        }
+    }
+    script_run "grep -e 'log_level.*=' -e 'log_outputs.*=' -e 'log_filters.*=' /etc/libvirt/*d.conf";
+    save_screenshot;
+
+    is_monolithic_libvirtd ? restart_libvirtd : restart_modular_libvirt_daemons;
 }
 
 #return 1 if test is expected to run on XEN hypervisor
@@ -430,7 +483,8 @@ sub ensure_online {
             }
             # Check also if name resolution works - restart libvirtd if not
             if (script_run("ssh $guest ping -c 3 -w 120 $dns_host", timeout => 180) != 0) {
-                restart_libvirtd if (is_xen_host || is_kvm_host);
+                # Note: TBD for modular libvirt. See poo#129086 for detail.
+                restart_libvirtd if (is_monolithic_libvirtd and (is_xen_host || is_kvm_host));
                 die "name resolution failed for $guest" if (script_retry("ssh $guest ping -c 3 -w 120 $dns_host", delay => 1, retry => 10, timeout => 180) != 0);
             }
         }
@@ -446,7 +500,8 @@ sub upload_y2logs {
 
 sub ensure_default_net_is_active {
     if (script_run("virsh net-list --all | grep default | grep ' active'", 90) != 0) {
-        restart_libvirtd;
+        # Note: TBD for modular libvirt. See poo#129086 for detail.
+        restart_libvirtd if is_monolithic_libvirtd;
         if (script_run("virsh net-list --all | grep default | grep ' active'", 90) != 0) {
             assert_script_run "virsh net-start default";
         }
@@ -483,10 +538,11 @@ sub remove_additional_nic {
 }
 
 sub collect_virt_system_logs {
-    if (script_run("test -f /var/log/libvirt/libvirtd.log") == 0) {
-        upload_logs("/var/log/libvirt/libvirtd.log");
-    } else {
-        record_info "File /var/log/libvirt/libvirtd.log does not exist.";
+    if (script_run("test -f /var/log/libvirt/*d.log") == 0) {
+        upload_logs("/var/log/libvirt/*d.log");
+    }
+    else {
+        record_info "File /var/log/libvirt/*d.log does not exist.";
     }
 
     if (script_run("test -d /var/log/libvirt/libxl/") == 0) {
--- a/lib/virt_autotest/virtual_network_utils.pm
+++ b/lib/virt_autotest/virtual_network_utils.pm
@@ -27,7 +27,7 @@ use virt_autotest::utils;
 
 our @EXPORT
   = qw(download_network_cfg prepare_network restore_standalone destroy_standalone restart_network
-  restore_guests restore_network destroy_vir_network restore_libvirt_default enable_libvirt_log pload_debug_log
+  restore_guests restore_network destroy_vir_network restore_libvirt_default pload_debug_log
   check_guest_status check_guest_module check_guest_ip save_guest_ip test_network_interface hosts_backup
   hosts_restore get_free_mem get_active_pool_and_available_space clean_all_virt_networks setup_vm_simple_dns_with_ip
   get_guest_ip_from_vnet_with_mac update_simple_dns_for_all_vm validate_guest_status);
@@ -286,13 +286,6 @@ sub restore_libvirt_default {
         assert_script_run("virsh net-define $default_path", 60);
         assert_script_run("rm -rf $default_path");
     }
-}
-
-sub enable_libvirt_log {
-    assert_script_run qq(echo 'log_level = 1
-    log_filters="3:remote 4:event 3:json 3:rpc"
-    log_outputs="1:file:/var/log/libvirt/libvirtd.log"' >> /etc/libvirt/libvirtd.conf);
-    restart_libvirtd;
 }
 
 sub upload_debug_log {
--- a/schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
+++ b/schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
@@ -4,8 +4,6 @@ description:    >
     Virtualization guest installation tests including extended feature tests schedule
 schedule:
     - '{{install_and_bootup}}'
-    - virt_autotest/switch_to_ssh_and_install_hypervisor
-    - '{{reboot}}'
     - '{{install_guest}}'
     - virt_autotest/set_config_as_glue
     - '{{virtual_network_test}}'
@@ -17,6 +15,13 @@ conditional_schedule:
         DO_NOT_INSTALL_HOST:
             0:
                 - installation/ipxe_install
+                - '{{install_host_with_needle_matching}}'
+    install_host_with_needle_matching:
+        HOST_INSTALL_AUTOYAST:
+            1:
+                - autoyast/installation
+                - virt_autotest/login_console
+            0:
                 - installation/welcome
                 - installation/online_repos
                 - installation/installation_mode
@@ -32,9 +37,7 @@ conditional_schedule:
                 - installation/start_install
                 - installation/await_install
                 - installation/reboot_after_installation
-    reboot:
-        DO_NOT_INSTALL_HOST:
-            0:
+                - virt_autotest/switch_to_ssh_and_install_hypervisor
                 - virt_autotest/reboot_and_wait_up_normal
     install_guest:
         SKIP_GUEST_INSTALL:
--- a/tests/installation/ipxe_install.pm
+++ b/tests/installation/ipxe_install.pm
@@ -17,6 +17,7 @@ use ipmi_backend_utils;
 use version_utils qw(is_upgrade is_tumbleweed);
 use bootloader_setup 'prepare_disks';
 use Utils::Architectures;
+use virt_autotest::utils qw(is_kvm_host is_xen_host);
 
 use HTTP::Tiny;
 use IPC::Run;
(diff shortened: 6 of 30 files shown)


Julie-CAO:
@alice-suse  @waynechen55 @varunkojha  @tonyyuan1 @xguo @nanzhg @tbaev @RoyCai7  Welcome review!

lib/virt_autotest/utils.pm
```diff
@@ -116,6 +116,13 @@ sub is_kvm_host {
     return check_var("SYSTEM_ROLE", "kvm") || check_var("HOST_HYPERVISOR", "kvm") || check_var("REGRESSION", "qemu-hypervisor");
 }
 
+#retrun 1 if libvirt 9.1+ is running
+sub is_libvirt91 {
+    unless (is_alp) {
+        return 1 if script_run('rpm -q libvirt-libs | grep "libs-9.[1-9]"') == 0;
```
alice-suse, line 122: How about libvirt version 10.x?

schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
```diff
@@ -17,6 +15,12 @@ conditional_schedule:
         DO_NOT_INSTALL_HOST:
             0:
                 - installation/ipxe_install
+                - '{{install_host_with_needle_matching}}'
+    install_host_with_needle_matching:
```
alice-suse, line 19: It is not used in section `schedule`? Forgotten?

schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
```diff
@@ -17,6 +15,12 @@ conditional_schedule:
         DO_NOT_INSTALL_HOST:
```
alice-suse, line 15: if `DO_NOT_INSTALL_HOST` is 1, it will directly go to {{install_guest}}? Miss login_console?

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
```diff
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: FSFAP
 #
 # Summary: For openSUSE virtualization test only. login in console, install kvm/xen patterns if needed.
-#  - Even if you'd like to run tests without host installation(DO_NOT_INSTALL_HOST=1), this module is still necessary as login console in this module is required.
 #  - This module is added for openSUSE TW because of the difference beteen SLE and TW. Meanwile, login_console, install_package and update_package from SLE are not needed. The reasons are listed below:
```
alice-suse, line 7: Hi Julie, would you please add a comment in this file summary to let readers know it is not stable to use this file for TW test and the short main reasons?

lib/virt_autotest/utils.pm
alice-suse: BTW, how about the function name is_libvirt91_plus?

tests/virt_autotest/virt_utils.pm
```diff
@@ -30,19 +29,23 @@ our @EXPORT
...
-    #log filter is set to store component logs with different levels.
-    my $libvirtd_conf_file = "/etc/libvirt/libvirtd.conf";
-    if (!script_run "ls $libvirtd_conf_file") {
-        script_run "sed -i '/^[# ]*log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $libvirtd_conf_file";
-        script_run "sed -i '/^[# ]*log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%};\${x;/^\$/{s%%log_outputs=\"1:file:/var/log/libvirt/libvirtd.log\"%;H};x}' $libvirtd_conf_file";
-        script_run "sed -i '/^[# ]*log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters=\"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $libvirtd_conf_file";
-        script_run "grep -e log_level -e log_outputs -e log_filters $libvirtd_conf_file";
+    # libvirtd is only present in SLE serials product(SLEM and ALP)
+    # it has been replaced with modular daemons in Tumbleweed since libvirt 9.1.0
+    unless (is_libvirt91) {
```
alice-suse, line 34: Hi Julie, I see nothing done for libvirt 9.1+ for debug enablement. Nothing needs to be done?

lib/virt_autotest/utils.pm
Julie-CAO: It is easier for checking libvirt9.x. It is possible this libvirt version checking would not been required any more for SLES when libvirt goes into 10.x.

tests/virt_autotest/virt_utils.pm
Julie-CAO: I saw debugging enablement is called in unified_guest_installation.pm, so I skip it in autoyast profiles.

Julie-CAO: Oh, I get your meaning. The services have changed since libvirt 9.1, original configuration files have not taken effects any longer. I'll find new configurations in the future after understanding more about libvirt.

lib/virt_autotest/utils.pm
alice-suse: Yes, it is unknown for higher version behavior. We can change again when the code won't work again in the future by either a newer version or newer behavior. It's also ok to assume that higher version works the same with current latest version, just like you did for 9.[2-9] and other similar judgement of is_sle(15+). Up to you to decide :) .

How about function name? Will it be good to claim is_libvirt91 true for libvirt 9.2?

tonyyuan1:
Hi Julie,

Did you test "autoyast installation with sol" in TW test environment? Are you sure that it's NOT stable in your environment?  

The use of ssh console makes your installation complicated and the autoyast install process is not normally used by  customers. 

The configurations are done at second stage.  if you disable the second stage some of configurations might not be done. 
You might not be able to used some autoyast features of configurations.

```
second_stage
A regular installation of SUSE Linux Enterprise Server is performed in a single stage. The auto-installation process, however, is divided into two stages. After the installation of the basic system the system boots into the second stage where the system configuration is done. Set this option to false to disable the second stage. Setting this value is optional. The default is true.
```

Julie-CAO:
> Hi Julie,
> 
> Did you test "autoyast installation with sol" in TW test environment? Are you sure that it's NOT stable in your environment?

No, I did not "test autoyast installation with sol". The purpose of my ticket of autoyast host installation is to abondan SOL console as much as possible, so I try not to use sol console in it.

Test failed in log in with sol console due to limitation of needle matching. Meanwhile, ssh configration had to be done via sol console previously, so autoyast host installation was brought up.

> The use of ssh console makes your installation complicated and the autoyast auto-installation process is not normally used by customers.

What complicated things do you mean? I don't get it. Typing 'yast.ssh' in ssh console?

> The configurations are done at second stage. if you disable the second stage some of configurations might not be done. You might not be able to used some autoyast features of configurations.
> 
> Below is the quote of second_stage
> 
> ```
> second_stage
> A regular installation of SUSE Linux Enterprise Server is performed in a single stage. The auto-installation process, however, is divided into two stages. After the installation of the basic system the system boots into the second stage where the system configuration is done. Set this option to false to disable the second stage. Setting this value is optional. The default is true.
> ```

Actually I don't know exactly what the 2nd stage does in autoyast, I just referenced to the guest autoyast file and made it work for host. Could you give some examples about what the 2nd stage do? I'll consider of it if it is needed to install a TW host. Thanks!

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -0,0 +1,143 @@
...
+	sed -i "/^[# ]*ClientAliveInterval */{h;s/^[# ]*ClientAliveInterval *.*\$/ClientAliveInterval 60/};\${x;/^\$/{s//ClientAliveInterval 60/;H};x}" $sshd_config_file
+        sed -i "/^[# ]*ClientAliveCountMax */{h;s/^[# ]*ClientAliveCountMax *.*\$/ClientAliveCountMax 480/};\${x;/^\$/{s//ClientAliveCountMax 480/;H};x}" $sshd_config_file 
+    done
+		fi
+]]>
+        </source>
+      </script>
+    </init-scripts>
+  </scripts>
+</profile>
```
waynechen55, line 143: Can we just change the folder name to ```host_unattended_installation_files``` ?

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -0,0 +1,143 @@
...
+    <!-- permit root login, password login and pubkeys login -->
+    <!-- keep ssh session alive -->
+    <!-- /etc/ssh/sshd_config is not default installed in TW -->
+    <init-scripts t="list">
+      <script>
+        <source><![CDATA[
+sshd_config_file="/etc/ssh/sshd_config"
+if [ ! -f $sshd_config_file ]; then
+    mkdir -p `dirname $sshd_config_file`
+    echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
```
waynechen55, line 128: If I remember correctly, the sshd config file for tumbleweed is ```/usr/share/etc/ssh/sshd_config``` ?

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -0,0 +1,143 @@
...
+    <global>
+      <activate>true</activate>
+      <boot_mbr>true</boot_mbr>
+      <gfxmode>auto</gfxmode>
+      <os_prober>true</os_prober>
+      <terminal>console serial</terminal>
+      <serial>serial --unit=1 --speed=115200 --word=8 --parity=no</serial>
+      <timeout t="integer">30</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>console=com2,115200 log_lvl=all guest_loglvl=all sync_console</xen_kernel_append>
```
waynechen55, line 52: I think the correct setting should be ```loglvl=all``` instead of ```log_lvl=all```.

lib/virt_autotest/utils.pm
```diff
@@ -116,6 +116,13 @@ sub is_kvm_host {
     return check_var("SYSTEM_ROLE", "kvm") || check_var("HOST_HYPERVISOR", "kvm") || check_var("REGRESSION", "qemu-hypervisor");
 }
 
+#retrun 1 if libvirt 9.1+ is running
+sub is_libvirt91 {
```
waynechen55, line 120: I think you can put your libvirt subroutine in ```lib/services/libvirtd.pm``` and make it better.

tonyyuan1:
> What complicated things do you mean? I don't get it. Typing 'yast.ssh' in ssh console?
> 
Yes, it's one of them.  you have to use the sshd that linuxrc set up. have to use one stage and  can't reboot to installed OS system, otherwise ssh connection will break.

 

> Actually I don't know exactly what the 2nd stage does in autoyast, I just referenced to the guest autoyast file and made it work for host. Could you give some examples about what the 2nd stage do?

I see you use init-scripts, below is quote:

`init-scripts (during the first boot of the installed system, all services up and running) 
`
You don't have reboot during installation so as to keep ssh connection alive.   Did you check that init-scripts was executed ?

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: "log_lvl" is used in SLE tests, it seems we should correct it as well. Add it to backlog. thanks!

waynechen55: Your kvm profile does not need ```xen_kernel_append``` ?

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -0,0 +1,143 @@
...
+      <xen_append>console=hvc0,115200 console=tty loglevel=5</xen_append>
+    </global>
+  </bootloader>
+  <partitioning t="list">
+    <drive>
+      <device>/dev/sda</device>
```
waynechen55, line 58: Better to avoid use device name directly ?

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: I keep it in kvm profile in order that at some point it might be switched to be a xen server easily, ie. just by zypper'ing install xen patterns.

lib/virt_autotest/utils.pm
Julie-CAO: I looked at this file. it has not been used by our squard. All the libvirt functions created or used by us are located in lib/virt_autotest/* or tests/virt_autotests/virt_utils.pm. If we would consider of using lib/services/libvirtd.pm, we'd better move all libvirt related functions to it. It doesn't feel very suitable to only include this function.

lib/virt_autotest/utils.pm
Julie-CAO: Yes, I agree with is_moduler_libvirt(). And thinking of improving it a little bit and verifying on SLE.

Julie-CAO: I am thinking of 
```
sub is_legacy_libvirtd {
    unless (is_alp) {
        return 1 if script_run('rpm -q libvirt-libs | grep -e "libs-9.0" -e "libs-[1-8]"') == 0;
    }
}
```
because modular libvirtd is the tendency and libvritd will be obsolete.

schedule/virt_autotest/tumbleweed-virt-guest-install.yaml
Julie-CAO: Yes, `DO_NOT_INSTALL_HOST` is meant to run guest installation test or feature tests on an installed system, so no restart or login is required, but you have to make sure the machine is in running state and to trigger test with `IPMI_DO_NOT_RESTART_HOST=1`. It is helpful for debugging.

tests/virt_autotest/switch_to_ssh_and_install_hypervisor.pm
Julie-CAO: Yes, done locally. will push in a moment.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: No problem, to be inline with guest autoyast profiles.

Julie-CAO:
> Yes, it's one of them. you have to use the sshd that linuxrc set up. have to use one stage and can't reboot to installed OS system, otherwise ssh connection will break.

I'll take a closer look about second_stage. 

> > Actually I don't know exactly what the 2nd stage does in autoyast, I just referenced to the guest autoyast file and made it work for host. Could you give some examples about what the 2nd stage do?
> 
> I see you use init-scripts, below is quote:
> 
> `init-scripts (during the first boot of the installed system, all services up and running) ` You don't have reboot during installation so as to keep ssh connection alive. Did you check that init-scripts was executed ?

It was executed in my verification test. I'll do more check.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: There is no `/usr/share/etc/ssh/sshd_config` at present. I guess you are saying `/usr/etc/ssh/sshd_config`.  

```
~ # cat /usr/etc/ssh/sshd_config
# To modify the system-wide sshd configuration, create a "*.conf" file under
# "/etc/ssh/sshd_config.d/" which will be automatically included below.
# Don't edit this configuration file itself if possible to avoid update
# problems.
```

I once tried with creating a .conf file under "/etc/ssh/sshd_config.d/", but it did not take effects. "/etc/ssh/sshd_config" was found to be effective. Don't know why, but I will keey an eye on it.

Julie-CAO: There is no `/usr/share/etc/ssh/sshd_config` at present. I guess you are saying `/usr/etc/ssh/sshd_config`.  

```
~ # cat /usr/etc/ssh/sshd_config
# To modify the system-wide sshd configuration, create a "*.conf" file under
# "/etc/ssh/sshd_config.d/" which will be automatically included below.
# Don't edit this configuration file itself if possible to avoid update
# problems.
```

I once tried with creating a .conf file under "/etc/ssh/sshd_config.d/", but it did not take effects. "/etc/ssh/sshd_config" was found to be effective. Don't know why, but I will keey an eye on it.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: I specify /dev/sda at here to install TW on the first disk and keep the 2nd disk in case of multiple systems on the machine.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: Just tried, adding two *.conf files under "/etc/ssh/sshd_config.d/" works, permit_root_login.conf and keep_alive.conf. so "/etc/ssh/sshd_config" will not be used in TW anymore.

Julie-CAO: Just tried, adding two *.conf files under "/etc/ssh/sshd_config.d/" works, permit_root_login.conf and keep_alive.conf. so "/etc/ssh/sshd_config" will not be used in TW anymore.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
waynechen55: According to autoyast doc, better to use persistent device representation ?
```
device

    xxxxxxxxx

    You can use persistent device names via ID, like /dev/disk/by-id/ata-WDC_WD3200AAKS-75L9 or by-path, like /dev/disk/by-path/pci-0001:00:03.0-scsi-0:0:0:0.
```

Julie-CAO:
I've tried with Autoyast stage 2 in TW:
[autoyast install with 2nd stage](https://openqa.opensuse.org/tests/3274091)
[without 2nd stage](https://openqa.opensuse.org/tests/3275014)

From the system configuration side, no noticeable different has been found, and it might bring troubles for logging. Meanwhile the autoyast guide of TW said:
```
The second stage runs only if it is strictly necessary, and the second stage can be turned off completely with the second_stage parameter:
```
So I decide to skip the 2nd stage in autoyast installation. It will be easily turn on in the future if it becomes necessary at some point.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: Can the device IDs or path for different machine be identicle? I am not quite familiar with it.

In this TW case, there are 2 machines in O3, maybe more in the future or be replaced by new machines. Only a static autoyast profile can be used in O3 infra, that is, no test settings can be referenced, so is it possible to use a device ID or path? @waynechen55  @alice-suse @tonyyuan1

Julie-CAO:
Going to improve the commit about modular libvirt described in https://149.44.161.6/pages/viewpage.action?spaceKey=qasleapac2&title=%5BWIP+DRAFT%5D+Modular+libvirt.

data/virt_autotest/host_autoyast/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: Don't use disk wwn in O3 as the infra is different from OSD as we discussed in meeting today.

Julie-CAO:
The final edits to these commits are ready. Here are the verification run:

[TW kvm test with sol console, debug log enabled for modular libvirt daemons conf files in TW](https://openqa.opensuse.org/tests/3276368#step/unified_guest_installation/113)
[SLE kvm test with debug log enabled for libvirtd, and the logs are uploaded for a failed test](https://openqa.suse.de/tests/11075667#step/unified_guest_installation/62)
[uefi-gi-guest_developing-on-host_developing-kvm](http://openqa.suse.de/tests/11083486)   //running ...
[gi-guest_developing-on-host_developing-xen](https://openqa.suse.de/tests/11083492)   //running ...
[uefi-gi-guest_developing-on-host_sles15sp4-kvm](http://openqa.suse.de/t11083532)   //running ...

lib/virt_autotest/utils.pm
```diff
@@ -116,6 +117,42 @@ sub is_kvm_host {
...
+
+# For legacy libvird, set debug level logging for libvirtd services
+# For modular libvirt, do the same settings to /etc/libvirt/virt{qemu,xen,driver}d.conf.
+# virt{qemu,xen}d daemons provide the most important libvirt log(sufficient for most issues).
+# virt{driver}.d daemons is only required by specific issues, eg virtual network failures may need virtnetworkd log.
+# But our automation is better to set them to collect more logs as we could as possible.
+# Developer asked to use different log file as log_output per daemon.
+sub turn_on_libvirt_debugging_log {
+
+    my @libvirt_daemons = is_monolithic_libvirtd ? "libvirtd" : qw(virtqemud virtstoraged virtnetworkd virtnodedevd virtsecretd virtproxyd virtlockd virtlogd);
```
alice-suse, line 135: For virtlockd and virtlogd, suggest to let other virt{drv}d start/stop/restart them by dependency need(eg virtlogd is not needed by xen). That's the usage way in alp container and upstream doc too.

Besides, would you please add virtnwfilterd? That's one of the core daemons. Even if we do not have dedicated VT test for it, at least we can cover start/stop the service which is the fundamental but most important check point actually :)

lib/virt_autotest/utils.pm
```diff
@@ -116,6 +117,42 @@ sub is_kvm_host {
...
+        if (script_run("ls $conf_file") == 0) {
+            script_run "sed -i '/^[# ]*log_level *=/{h;s/^[# ]*log_level *= *[0-9].*\$/log_level = 1/};\${x;/^\$/{s//log_level = 1/;H};x}' $conf_file";
+            script_run "sed -i '/^[# ]*log_outputs *=/{h;s%^[# ]*log_outputs *=.*[0-9].*\$%log_outputs = \"1:file:/var/log/libvirt/$daemon.log\"%};\${x;/^\$/{s%%log_outputs = \"1:file:/var/log/libvirt/$daemon.log\"%;H};x}' $conf_file";
+            script_run "sed -i '/^[# ]*log_filters *=/{h;s%^[# ]*log_filters *=.*[0-9].*\$%log_filters = \"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%};\${x;/^\$/{s%%log_filters = \"1:qemu 1:libvirt 4:object 4:json 4:event 3:util 1:util.pci\"%;H};x}' $conf_file";
+        }
+    }
+    script_run "grep -e 'log_level.*=' -e 'log_outputs.*=' -e 'log_filters.*=' /etc/libvirt/*d.conf";
+    save_screenshot;
+
+    restart_libvirtd if is_monolithic_libvirtd;
```
alice-suse, line 153: For modular daemons, they also need restart.

lib/virt_autotest/utils.pm
```diff
@@ -430,7 +467,7 @@ sub ensure_online {
             }
             # Check also if name resolution works - restart libvirtd if not
             if (script_run("ssh $guest ping -c 3 -w 120 $dns_host", timeout => 180) != 0) {
-                restart_libvirtd if (is_xen_host || is_kvm_host);
+                restart_libvirtd if (is_monolithic_libvirtd and (is_xen_host || is_kvm_host));
```
alice-suse, line 470: For nearly all below libvirtd related operations restart/stop, the is_monolithic_libvirtd condition is added. Why? Modular libvirtd needs similar operations too. Do I miss any of your plans, eg implementing them in another PR?

data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -0,0 +1,136 @@
...
+    <default_target>multi-user</default_target>
+    <services>
+      <disable t="list">
+        <service>firewalld</service>
+      </disable>
+      <enable t="list">
+        <service>sshd</service>
+	<service>virtqemud</service>
+	<service>virtstoraged</service>
+	<service>virtnetworkd</service>
```
alice-suse, line 113: Suggest to add virtnwfilterd too, see reason in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16625/files#r1189422090.

alice-suse:
> The final edits to these commits are ready. Here are the verification run:
> 
> [TW kvm test with sol console, debug log enabled for modular libvirt daemons conf files in TW](https://openqa.opensuse.org/tests/3276368#step/unified_guest_installation/113) [SLE kvm test with debug log enabled for libvirtd, and the logs are uploaded for a failed test](https://openqa.suse.de/tests/11075667#step/unified_guest_installation/62) [uefi-gi-guest_developing-on-host_developing-kvm](http://openqa.suse.de/tests/11083486) //running ... [gi-guest_developing-on-host_developing-xen](https://openqa.suse.de/tests/11083492) //running ... [uefi-gi-guest_developing-on-host_sles15sp4-kvm](http://openqa.suse.de/t11083532) //running ...

Will there be failure-on-purpose jobs to verify the modular logs?

lib/virt_autotest/utils.pm
Julie-CAO: These daemons are not meant to be start/stop/restarted at here. The function is to add some settings to the configuration files of all the modular libvirt daemons. So virtlockd and virtlogd daemons theirselves are not touched.

I can add virtnwfilterd as it is required, including daemons and configuration files.

lib/virt_autotest/utils.pm
Julie-CAO: Is it realy required? In my test, "/var/log/virt*d.log" are created without restart any libvirt services as it is defined as `log_outputs = \"1:file:/var/log/libvirt/$daemon.log` in its configuration file.

Julie-CAO:
> > The final edits to these commits are ready. Here are the verification run:
> > [TW kvm test with sol console, debug log enabled for modular libvirt daemons conf files in TW](https://openqa.opensuse.org/tests/3276368#step/unified_guest_installation/113) [SLE kvm test with debug log enabled for libvirtd, and the logs are uploaded for a failed test](https://openqa.suse.de/tests/11075667#step/unified_guest_installation/62) [uefi-gi-guest_developing-on-host_developing-kvm](http://openqa.suse.de/tests/11083486) //running ... [gi-guest_developing-on-host_developing-xen](https://openqa.suse.de/tests/11083492) //running ... [uefi-gi-guest_developing-on-host_sles15sp4-kvm](http://openqa.suse.de/t11083532) //running ...
> 
> Will there be failure-on-purpose jobs to verify the modular logs?

The 2nd one in the list above.

lib/virt_autotest/utils.pm
alice-suse: In https://www.libvirt.org/kbase/debuglogs.html#modular-vs-monolithic-daemons, it gives below. So I think it should be yes, just like monolithic libvirtd. But I can't explain why in your case no restart needed...

```
find & replace, or set the appropriate [Log outputs](https://www.libvirt.org/kbase/debuglogs.html#log-outputs) and [Log filters](https://www.libvirt.org/kbase/debuglogs.html#log-filters), e.g

log_filters="3:remote 4:event 3:util.json 3:rpc 1:*"
log_outputs="1:file:/var/log/libvirt/libvirtd.log"
save and exit

restart the corresponding service/daemon 
```

lib/virt_autotest/utils.pm
alice-suse: > For virtlockd and virtlogd, suggest to let other virt{drv}d start/stop/restart them by dependency need(eg virtlogd is not needed by xen). That's the usage way in alp container and upstream doc too.

Sorry if any confusion. I gave it here because I was considering restarting libvirt daemons as needed , based on https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16625#discussion_r1189650705. So this is for below line 153, but not for setting debug logs in *.conf.

tests/virt_autotest/libvirt_isolated_virtual_network.pm
```diff
@@ -83,7 +83,7 @@ sub post_fail_hook {
     $self->SUPER::post_fail_hook;
 
     #Restart libvirtd service
-    virt_autotest::utils::restart_libvirtd();
+    virt_autotest::utils::restart_libvirtd() if is_monolithic_libvirtd;
```
alice-suse, line 86: For this libvirt restart operation, I guess the reason is that it is in post_fail_hook, so possible that libvirtd run into some problematic state, to let the destroy_virt_network handled successfully, restart libvirtd may work, to achieve a good env cleanup.

lib/virt_autotest/virtual_network_utils.pm
```diff
@@ -288,13 +288,6 @@ sub restore_libvirt_default {
     }
 }
 
-sub enable_libvirt_log {
-    assert_script_run qq(echo 'log_level = 1
-    log_filters="3:remote 4:event 3:json 3:rpc"
-    log_outputs="1:file:/var/log/libvirt/libvirtd.log"' >> /etc/libvirt/libvirtd.conf);
-    restart_libvirtd;
```
alice-suse, line 295: This libvirtd restart is to load the new config file.

Julie-CAO: I remove the entire function because I replace it with the common function turn_on_libvirt_debugging_log() in libvirt_virtual_network_init.pm.

lib/virt_autotest/utils.pm
Julie-CAO: I'll run a test to confirm whether these daemons are required to be restarted in TW.

Julie-CAO: I'll run a test to confirm whether these daemons are required to be restarted in TW.

lib/virt_autotest/utils.pm
Julie-CAO: I agree with creating a new ticket to add a new function instead of extend the current restart_libvirtd(),

tests/virt_autotest/libvirt_isolated_virtual_network.pm
Julie-CAO: I'll create a new function named restart_modular_libvirt_daemons() function which will restart any daemons specified by parameters, fg. restart_modular_libvirt_daemons(virtqemud), restart all 7 daemons without any parameters passed.

Then the test owner will decide if it should be called and how it is called in their tests in the next release.

alice-suse: Good to have default to all if nothing specified :)

alice-suse: I created a ticket https://progress.opensuse.org/issues/129086 for further modular adaption in sle.

Julie-CAO:
the latest verification run:
[failed TW kvm test and libvirt logs uploaded](https://openqa.opensuse.org/tests/3282580)

lib/virt_autotest/utils.pm
Julie-CAO: Restarting all daemons in my test. It took 40s and seems not noisy.

Julie-CAO:
The PR is ready for merge as all things are resolved.  @alice-suse @varunkojha @xguo @waynechen55  @tonyyuan1

data/virt_autotest/virt_logs_collector.sh
```diff
@@ -360,11 +373,13 @@ function collect_extra_logs_from_host() {
...
               ret_result=$(( ${ret_result} | $? ))
            fi           
-        elif [[ ${release} -eq 15 ]];then
+        elif [[ ${release} -eq 15 || `get_product_name` == "openSUSE Tumbleweed" ]];then
            cp --parent -f -r ${libvirt_boot_log} ${extra_logs_folder}
            ret_result=$(( ${ret_result} | $? ))
            cp --parent -f -r ${libvirt_qemu_log} ${extra_logs_folder}
            ret_result=$(( ${ret_result} | $? ))
+	   cp --parent -f -r ${libvirt_daemon_logs} ${extra_logs_folder}
+	   et_result=$(( ${ret_result} | $? ))
```
alice-suse, line 382: Typo? et_result => ret_result

Julie-CAO: corrected. thanks!

lib/virt_autotest/utils.pm
```diff
@@ -55,8 +56,22 @@ sub restart_libvirtd {
...
+# usage: restart_modular_libvirt_daemons([daemon1_name daemon2_name ...]). For example:
+# to specify daemons which will be restarted: restart_modular_libvirt_daemons(virtqemud virtstoraged ...)
+# to restart all modular daemons without any daemons passed
+sub restart_modular_libvirt_daemons {
+    my @daemons = @_ == 0 ? qw(virtqemud virtstoraged virtnetworkd virtnodedevd virtsecretd virtproxyd virtnwfilterd) : split(" ", @_);
+    if (is_alp) {
+        record_soft_failure("Restaring modular libvirt daemons has not been emplemented in ALP. See poo#129086");
+    }
+    else {
+        systemctl("restart $_\{,-ro,-admin\}.socket") foreach @daemons;
```
alice-suse, line 72: Would you please also add  "restart $_.service" ? Sorry if you were misled by my wip modular libvirt confluence page, which didn't differentiate clearly for switching to modular and restart of modular.

lib/virt_autotest/utils.pm
```diff
@@ -55,8 +56,22 @@ sub restart_libvirtd {
     else {
         systemctl("restart libvirtd", timeout => 180);
     }
+    record_info("Libvirtd has been restarted!");
     save_screenshot;
-    record_info("Debug log for libvirtd has been enabled!");
+}
+
+# usage: restart_modular_libvirt_daemons([daemon1_name daemon2_name ...]). For example:
```
alice-suse, line 63: usage => Usage?

lib/virt_autotest/utils.pm
```diff
@@ -55,8 +56,22 @@ sub restart_libvirtd {
...
+# to specify daemons which will be restarted: restart_modular_libvirt_daemons(virtqemud virtstoraged ...)
+# to restart all modular daemons without any daemons passed
+sub restart_modular_libvirt_daemons {
+    my @daemons = @_ == 0 ? qw(virtqemud virtstoraged virtnetworkd virtnodedevd virtsecretd virtproxyd virtnwfilterd) : split(" ", @_);
+    if (is_alp) {
+        record_soft_failure("Restaring modular libvirt daemons has not been emplemented in ALP. See poo#129086");
```
alice-suse, line 69: emplemented => implemented?

lib/virt_autotest/utils.pm
Julie-CAO: Done.

lib/virt_autotest/utils.pm
Julie-CAO: Corrected.

lib/virt_autotest/utils.pm
Julie-CAO: Thanks.

alice-suse (approved):
Thanks for the efforts. It looks good.

lib/virt_autotest/utils.pm
```diff
@@ -116,6 +132,42 @@ sub is_kvm_host {
     return check_var("SYSTEM_ROLE", "kvm") || check_var("HOST_HYPERVISOR", "kvm") || check_var("REGRESSION", "qemu-hypervisor");
 }
 
+#retrun 1 if libvirt 9.0- is running which monolithic libvirtd is the default service
+sub is_monolithic_libvirtd {
+    unless (is_alp) {
+        return 1 if script_run('rpm -q libvirt-libs | grep -e "libs-9\.0" -e "libs-[1-8]\."') == 0;
+    }
+}
```
waynechen55, line 140: It seems that this subroutine does not return explicitly for all cases. Better to return 0 or 1 for all cases ?

tests/installation/ipxe_install.pm
```diff
@@ -149,22 +150,36 @@ sub run {
...
-        wait_still_screen(stilltime => 30, similarity_level => 60, timeout => 120);
+        wait_still_screen(stilltime => 12, similarity_level => 60, timeout => 30) unless check_screen('sshd-server-started', timeout => 60);
         save_screenshot;
     }
 
     # when we don't use autoyast, we need to also load the right test modules to perform the remote installation
-    if (get_var('AUTOYAST')) {
+    if (get_var('AUTOYAST') and !get_var('VIRT_AUTOTEST')) {
         # make sure to wait for a while befor changing the boot device again, in order to not change it too early
         sleep 120;
```
waynechen55, line 180: Any method to avoid use ```sleep 120``` ?

lib/virt_autotest/utils.pm
Julie-CAO: yes, thanks.

Julie-CAO:
final verification run passed. https://openqa.opensuse.org/tests/3293388 The PR is ready for merge.

Julie-CAO:
@alice-suse since there is no new comments, could you please merge this PR?

tests/virt_autotest/login_console.pm
```diff
@@ -44,19 +45,29 @@ sub double_check_xen_role {
         die 'Double-check xen kernel failed';
     }
 
+    # for modular libvirt, virtxend is expected in "loaded: active or inactive" status.
+    # virtxend.socket seems to be always in "loaded: active" status
+    unless (is_monolithic_libvirtd) {
+        die 'virtxend.socket is not running!' unless script_run("systemctl is-active virtxend.socket") eq 0;
+    }
```
waynechen55, line 52: Actually this can be further simplified as:
```
die 'virtxend.socket is not running!' if (script_run("systemctl is-active virtxend.socket") ne 0 and !is_monolithic_libvirtd)
```

tests/virt_autotest/login_console.pm
```diff
@@ -44,19 +45,29 @@ sub double_check_xen_role {
...
-    }
-    else {
         die "KVM modules are not loaded!";
     }
+
+    # for modular libvirt, virtqemud is expected in "loaded: active or inactive" status.
+    # virtqemud.socket seems to be always in "loaded: active" status
+    unless (is_monolithic_libvirtd) {
+        die 'virtqemud.socket is not running!' unless script_run("systemctl is-active virtqemud.socket") eq 0;
+    }
```
waynechen55, line 69: This can also be simplified as:
```
die 'virtqemud.socket is not running!' if (script_run("systemctl is-active virtqemud.socket") ne 0 and !is_monolithic_libvirtd)
```

tests/virt_autotest/login_console.pm
```diff
@@ -110,7 +121,7 @@ sub login_to_console {
         }
     }
 
-    if (!check_screen([qw(grub2 grub1 prague-pxe-menu)], 210)) {
+    unless (!check_screen([qw(grub2 grub1 prague-pxe-menu)], 210) or is_tumbleweed) {
```
waynechen55, line 124: The logic here changes ? If it is SLES and check_screen fails, ipmitool power reset will be triggered previously, but now if check_screen succeeds, system will reboot ?

tests/virt_autotest/login_console.pm
Julie-CAO: It is too long for one line, so I split them into two lines.

tests/virt_autotest/login_console.pm
Julie-CAO: good catch. forgot to remove "!" when 'if' was changed to 'unless'.

Julie-CAO:
typos are corrected.

waynechen55:
> typos are corrected.

It would be better if you can go over and confirm your code. Thanks.

Julie-CAO:
> > typos are corrected.
> 
> It would be better if you can go over and confirm your code line by line. Thanks

Just verified on a SLT test and it passed, https://openqa.suse.de/tests/11153982

Julie-CAO:
> > typos are corrected.
> 
> It would be better if you can go over and confirm your code line by line. Thanks.

Thanks, I checked codes. I'll monitor recent builds closely and new PR will be submitted to fix any errors introduced by this PR.

alice-suse:
Seems all comments are resolved. Merge it. Thanks Julie.

PR #16633, opened 2023-03-15
Add a simple NVMe loopback test

frankenmichl:
Before running more complex tests that also require more time to run, we should have some simple smoke tests for nvme, especially nvme-cli and vmetcli.

### DIFF
--- a/tests/kernel/nvme.pm
+++ b/tests/kernel/nvme.pm
@@ -1,0 +1,76 @@
+# SUSE's openQA tests
+#
+# Copyright 2018-2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: nvme
+# Summary: NVMe smoke tests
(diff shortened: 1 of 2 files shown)


tests/kernel/nvme.pm
```diff
@@ -0,0 +1,76 @@
...
+
+    record_info("ERROR", "nvmetcli is not equal to upstram version: $result", result => 'fail') unless ($result eq "");
+
+    # load modules
+    foreach my $module (@nvme_modules) {
+        assert_script_run("modprobe $module");
+
+    }
+    #setup nvme loop device
+    assert_script_run "dd if=/dev/zero of=/tmp/nvme_loopback bs=1M count=4096";
```
schlad, line 52: shall we use parenthesis consistently across entire script?

schlad (approved):

PR #16634, opened 2023-03-15
Setup fips during installation in Micro

mloviska:
Install fips pattern and update kernel parameters in order to enable fips during installation time.

- ticket: [Enable FIPS during ISO installation](https://progress.opensuse.org/issues/125084)
- Needles: 
  * https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1636
- Verification runs:
  * [sle-micro 5.4](http://kepler.suse.cz/tests/20457#)
  * [microos](http://kepler.suse.cz/tests/20458)

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -99,8 +99,13 @@ sub load_installation_tests {
         loadtest 'installation/ntp_config_settings';
         loadtest 'installation/user_settings_root';
         loadtest 'installation/resolve_dependency_issues';
-        loadtest 'installation/select_patterns' if get_var('PATTERNS');
-        loadtest 'installation/installation_overview';
+        if (get_var('PATTERNS')) {
+            loadtest 'installation/select_patterns';
+            loadtest 'installation/installation_overview';
+            loadtest 'installation/edit_optional_kernel_cmd_parameters' if get_var('PATTERNS') =~ m/fips/i;
+        } else {
+            loadtest 'installation/installation_overview';
+        }
         loadtest 'installation/disable_grub_timeout';
         loadtest 'installation/enable_selinux' if get_var('ENABLE_SELINUX');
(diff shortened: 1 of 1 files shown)


lib/main_micro_alp.pm
```diff
@@ -268,7 +273,8 @@ sub load_tests {
         # in 10G-disk tests, we don't run more tests
         return if check_var('HDDSIZEGB', '10');
         # Stop here if we are testing only scc extensions (live, phub, ...) activation
-        if (get_var('SCC_ADDONS') =~ /phub/) {
+        my $is_phub = get_var('SCC_ADDONS');
+        if (defined($is_phub) && $is_phub =~ /phub/) {
```
mloviska, line 277: This is part is fixing a warning

```
[2023-03-15T14:54:43.657527+01:00] [debug] scheduling textinfo tests/console/textinfo.pm
Use of uninitialized value in pattern match (m//) at sle-micro/products/sle-micro/../../lib/main_micro_alp.pm line 272.
	main_micro_alp::load_tests() called at sle-micro/products/sle-micro/main.pm line 59
```

jlausuch:
This only installs the software, but don't we also need to run the kernel with `fips=1`?

jlausuch:
> This only installs the software, but don't we also need to run the kernel with `fips=1`?

ah ok, I see this is done here: http://kepler.suse.cz/tests/20457#step/edit_optional_kernel_cmd_parameters/24

So, I wonder why gpg test fails and it's green in OSD...

mloviska:
> So, I wonder why gpg test fails and it's green in OSD...

Apparently I have missed this check, https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/console/gpg.pm#L120

jlausuch (approved):
LGTM

PR #16641, opened 2023-03-16
Fixed Regex to extract version number

yarunachalam:
current DMS job fails latest version as "N/A". Fixed regex to extract version number
- Verification run: (https://openqa.suse.de/tests/10807091)

CASEDIR | https://github.com/yarunachalam/os-autoinst-distri-opensuse.git#regfix
-- | --

--

### DIFF
--- a/tests/publiccloud/migration.pm
+++ b/tests/publiccloud/migration.pm
@@ -34,7 +34,7 @@ sub run {
     register_addons_in_pc($instance);
     $instance->ssh_assert_script_run('sudo zypper ref; sudo zypper -n up', timeout => 300);
 
-    record_info('INFO', $target_version);
+    record_info('Target Version', $target_version);
 
     sleep 90;    # wait for a bit for zypper to be available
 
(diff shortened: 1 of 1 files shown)


tests/publiccloud/migration.pm
```diff
@@ -66,11 +66,14 @@ sub run {
     # migration finished and instance rebooted
     record_info('INFO', 'Checking the migration succeed');
 
-    my $product_version = $instance->ssh_script_run(cmd => 'sudo cat /etc/os-release');
+    my $product_version = $instance->ssh_script_run(cmd => 'sudo cat /etc/os-release | grep ^VERSION_ID');
```
asmorodskyi, line 69: this is just FYI message , why you cut it here ? we can keep full os-release content maybe it will be useful one day

tests/publiccloud/migration.pm
```diff
@@ -66,11 +66,14 @@ sub run {
...
     record_info('INFO', $product_version);
 
     my $migrated_version = 'N/A';
-    $migrated_version = $1 if $product_version =~ /^VERSION_ID="([\d\.]+)"/sm;
+    my $get_version_id_cmd = "grep '^VERSION_ID=' /etc/os-release | cut -d'=' -f2 | tr -d '\"'";
+    $migrated_version = $instance->run_ssh_command(cmd => $get_version_id_cmd);
```
asmorodskyi, line 74: no need to do a second call to get what you already have in `$product_version` variable . please stick to old approach and extract all needed info from a variable using perl .  Also you can find out right regex without running dozens of tests . just copy content of os release to local file and create small perl file which trying to grep it

tests/publiccloud/migration.pm
yarunachalam: for me it makes sense to only product version in the variable.. So, many other details or there in os-release. (we can extract with appropriate variable later.)

tests/publiccloud/migration.pm
yarunachalam: As you suggested tried with small perl file and fixed the code back in perl

yarunachalam:
> 

fixed

tests/publiccloud/migration.pm
```diff
@@ -66,11 +66,11 @@ sub run {
     # migration finished and instance rebooted
     record_info('INFO', 'Checking the migration succeed');
 
-    my $product_version = $instance->ssh_script_run(cmd => 'sudo cat /etc/os-release');
-    record_info('INFO', $product_version);
+    my $product_version = $instance->run_ssh_command(cmd => 'sudo cat /etc/os-release');
+    record_info('INFO Product Version', $product_version);
```
jlausuch, line 70: In the openQA frame, long text strings are cut. As a recommendation, I would do one of the following:
`record_info('INFO', "Product Version: $product_version");`
`record_info('Product Version', $product_version);`

jlausuch (approved):
One small suggestion, but I can merge if you want.

yarunachalam:
> we need to extract needed info using perl not doing second call

Done

jlausuch (approved):

PR #16644, opened 2023-03-17
Bump timeout in yast2_kdump_use_fadump

lemon-suse:
Spvm is not particularly slow so we could bump the timeout for all arch in the point where we expect the popup. We need wait the restart info page shown then check the page is existed.

- Related ticket: https://progress.opensuse.org/issues/125957
- Needles: N/A
- Verification run: http://openqa.nue.suse.com/tests/10716015

### DIFF
--- a/lib/YaST/RestartInfoPage.pm
+++ b/lib/YaST/RestartInfoPage.pm
@@ -28,4 +28,12 @@ sub confirm_reboot_needed {
     $self->get_restart_info_page()->{btn_ok}->click();
 }
 
+sub wait_restart_info_popup {
+    my ($self) = @_;
+
+    YuiRestClient::Wait::wait_until(object => sub {
+            $self->{btn_ok}->exist({timeout => 0});
+    }, timeout => 180, message => "btn_ok does not exist");
+}
+
 1;
--- a/tests/console/yast2_kdump_use_fadump.pm
+++ b/tests/console/yast2_kdump_use_fadump.pm
@@ -29,6 +29,7 @@ sub run {
     save_screenshot;
     $fadump->get_navigation->ok();
     save_screenshot;
+    $restartinfo->wait_restart_info_popup();
     $restartinfo->confirm_reboot_needed();
     save_screenshot;
 


lib/YaST/RestartInfoPage.pm
```diff
@@ -19,6 +19,7 @@ sub init {
 
 sub get_restart_info_page {
     my ($self) = @_;
+    $self->wait_restart_info_popup();
```
jknphy, line 22: this call needs to go in the test layer, this getter should have only one purpose not two.

lib/YaST/RestartInfoPage.pm
```diff
@@ -28,4 +28,13 @@ sub confirm_reboot_needed {
     $self->get_restart_info_page()->{btn_ok}->click();
 }
 
+sub wait_restart_info_popup {
+    my ($self) = @_;
+
+    YuiRestClient::Wait::wait_until(object => sub {
+            $self->{btn_ok}->exist({timeout => 0});
+    }, timeout => 180, message => "btn_ok is not exist");
```
jknphy, line 36: typo, 'does not'

tests/console/yast2_kdump_use_fadump.pm
```diff
@@ -29,7 +29,7 @@ sub run {
     save_screenshot;
     $fadump->get_navigation->ok();
     save_screenshot;
-    $restartinfo->confirm_reboot_needed();
+    $restartinfo->wait_restart_info_popup->get_restart_info_page->confirm_reboot_needed();
```
jknphy, line 32: Those are two calls:
```
$restartinfo->wait_restart_info_popup();
$restartinfo->confirm_reboot_needed();
```
If you chain them, you don't need the `get_restart_info_page` in the middle because you already have the $self, but `wait_restart_info_popup` should not return $self, because its purpose is just to 'wait'.

lib/YaST/RestartInfoPage.pm
lemon-suse: Updated.

lib/YaST/RestartInfoPage.pm
lemon-suse: Updated

tests/console/yast2_kdump_use_fadump.pm
lemon-suse: Updated code and VR. Thanks.

PR #16651, opened 2023-03-17
BCI: Skip metadata tests if the public image is not available

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/126158
- Failure: https://openqa.suse.de/tests/10716006#step/_root_BCI-tests_metadata_podman/6
- VRs: 
    https://openqa.suse.de/tests/10717366 (from failed test)
    https://openqa.suse.de/tests/10717367 (from already green test)

### DIFF
--- a/tests/containers/bci_prepare.pm
+++ b/tests/containers/bci_prepare.pm
@@ -134,6 +134,16 @@ sub run {
             # Note: Both lines are aligned, thus the additional space
             record_info('builds', "CONTAINER_IMAGE_BUILD:  $build\norg.opensuse.reference: $reference");
             die('Missmatch in image build number. The image build number is different than the one triggered by the container bot!') if ($reference !~ /$build$/);
+            if (check_var('BCI_TEST_METADATA', '1')) {
+                # Check if we can run Metadata tests when the variable BCI_TEST_METADATA is set.
+                # Metadata tests compare the image under tests vs the released image in the public registry,
+                # so we need to make sure the public image exists, which won't be the case for brand new images
+                my $out = script_output("$engine pull -q $reference 2>&1", timeout => 600, proceed_on_failure => 1);
+                if ($out =~ m/manifest unknown/g) {
+                    record_info("SKIP", "Skip metadata tests. Image $reference does not exist yet.");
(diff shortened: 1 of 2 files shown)


mloviska (approved):

tests/containers/bci_prepare.pm
```diff
@@ -134,6 +134,16 @@ sub run {
             # Note: Both lines are aligned, thus the additional space
             record_info('builds', "CONTAINER_IMAGE_BUILD:  $build\norg.opensuse.reference: $reference");
             die('Missmatch in image build number. The image build number is different than the one triggered by the container bot!') if ($reference !~ /$build$/);
+            if (check_var('BCI_TEST_METADATA', '1')) {
+                # Check if we can run Metadata tests when the variable BCI_TEST_METADATA is set.
+                # Metadata tests compare the image under tests vs the released image in the public registry,
+                # so we need to make sure the public image exists, which won't be the case for brand new images
```
dcermak, line 140: This is wrong. The metadata tests check the labels of the image and skipping them universally is a **bad** idea.

jlausuch: Not universally, they will be skipped for images that are not public yet. Once they are published in public repo, the check will be `true` and it will run the metadata tests.

dcermak: Yes and then we will figure out that the metadata are wrong, but the damage has been already done. What you actually want to do, is just skip the test of the reference not all of them

jlausuch: ok, I guess we can handle that condition in the bci-tests repo instead. I'll revert

dcermak: we don't have to handle it in the bci-tests repo, you can also just filter the test out via by appending `-k "not reference"` to the tox call

jlausuch: Sure, but then we might also have failing tests in the bci-tests CI when some image is not yet released but enabled. If that's ok, then I'll add that option in this code.

dcermak: If you have a good & reliable way how to do that, please send a PR!

PR #16660, opened 2023-03-20
[virt_autotest] Add ALP Host Bridge Network Test Support

guoxuguang:
Add ALP Host Bridge Network Test Support

- Related ticket: https://progress.opensuse.org/issues/122731
- Verification run: 
[use existed host bridge network interface](http://10.67.129.218/tests/395)
[setup new host bridge network interface](http://10.67.129.218/tests/393#)
[call post_fail_hook](http://10.67.129.218/tests/381)

### DIFF
--- a/data/virt_autotest/alp_host_bridge_init.sh
+++ b/data/virt_autotest/alp_host_bridge_init.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Print commands and abort script on failure
+set -ex
+#
+# PREPARATION
+# --------------------
+. /etc/os-release
+# Variables Definition
+# --------------------
+BRIDGEIF_NAME="br0"
+ACTIVE_DEVICE=$(ip a|awk -F': ' '/state UP/ {print $2}'|head -n1)
+CONNECTION_NAME=$(nmcli -f DEVICE,CONNECTION dev status| grep ${ACTIVE_DEVICE} | awk '{print $4}')
+ACTIVE_CONNECTION="Wired connection ${CONNECTION_NAME}"
+ACTIVE_UUID=$(nmcli -g UUID,DEVICE con | grep ${ACTIVE_DEVICE} | cut -d: -f1)
+IPV4_ADDRESS=$(nmcli -g IP4.ADDRESS con show "${ACTIVE_UUID}")
+IPV4_GATEWAY=$(nmcli -g IP4.GATEWAY con show "${ACTIVE_UUID}")
+IPV4_DNS_A=$(nmcli -g IP4.DNS con show "${ACTIVE_UUID}"| awk '{print $1}')
+IPV4_DNS_B=$(nmcli -g IP4.DNS con show "${ACTIVE_UUID}"| awk '{print $3}')
+
+# Function Definitions
+# --------------------
+# Add Host Bridge Network Interface via nmcli
+setup_host_bridge_network_interface() {
+    # Add bridge type
+    nmcli con add type bridge con-name ${BRIDGEIF_NAME} ifname ${BRIDGEIF_NAME} \
+         autoconnect yes ipv4.method manual ipv4.address ${IPV4_ADDRESS} \
+         ipv4.gateway ${IPV4_GATEWAY} ipv4.dns ${IPV4_DNS_A} +ipv4.dns ${IPV4_DNS_B} \
+    # Add bridge-slave type
+    nmcli con add type bridge-slave con-name ${BRIDGEIF_NAME}-slave ifname ${ACTIVE_DEVICE} master ${BRIDGEIF_NAME}
+}
+
+# Enable Host Bridge Network Interface via nmcli
+enable_host_bridge_network_interface() {
+    nmcli con up ${BRIDGEIF_NAME}
+    # refer to bsc#1208005 for more details
+    nmcli con down "${ACTIVE_CONNECTION}"
+}
+
+# ==== #
+# MAIN # Testing starts here ...
+# ==== #
+if [[ $ID =~ 'alp' ]]; then
+    echo "Confirm Test Environment: ALP"
+else
+    echo "Error: Please perform this Tool on ALP! Exit... now"
+    exit 1
+fi
+
+setup_host_bridge_network_interface
+enable_host_bridge_network_interface
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -30,8 +30,7 @@ our @EXPORT = qw(is_vmware_virtualization is_hyperv_virtualization is_fv_guest i
   print_cmd_output_to_file ssh_setup ssh_copy_id create_guest import_guest install_default_packages upload_y2logs ensure_default_net_is_active ensure_guest_started
   ensure_online add_guest_to_hosts restart_libvirtd remove_additional_disks remove_additional_nic collect_virt_system_logs shutdown_guests wait_guest_online start_guests restore_downloaded_guests save_original_guest_xmls restore_original_guests
   is_guest_online wait_guests_shutdown remove_vm setup_common_ssh_config add_alias_in_ssh_config parse_subnet_address_ipv4 backup_file manage_system_service setup_rsyslog_host
-  check_port_state subscribe_extensions_and_modules download_script download_script_and_execute is_sev_es_guest upload_virt_logs recreate_guests
-  download_vm_import_disks);
+  check_port_state subscribe_extensions_and_modules download_script download_script_and_execute is_sev_es_guest upload_virt_logs recreate_guests download_vm_import_disks enable_nm_debug check_activate_network_interface set_host_bridge_interface_with_nm upload_nm_debug_log);
 
 my %log_cursors;
 
@@ -943,5 +942,44 @@ sub download_vm_import_disks {
     record_info("All imported disk download is done.");
 }
 
+sub enable_nm_debug {
+    # Enable Network Manager Debug Log Level
+    assert_script_run("nmcli general logging level DEBUG domains ALL", 60);
+    record_info("Enable Network Manager in Debug Level successfully for automation test.");
+}
+
+sub check_activate_network_interface {
+    # Check with activate network interface as required
+    my ($network_interface, $target_name) = @_;
+    $network_interface //= "br0";
+    $target_name //= get_required_var('OPENQA_URL');
+    assert_script_run("ping -I $network_interface -c 3 $target_name", 60);
+    assert_script_run("nmcli device show $network_interface", 60);
+    save_screenshot;
+    record_info("Activate Network Interface check successfully for automation test.");
+}
+
+sub set_host_bridge_interface_with_nm {
+    # Setup Host Bridge Network Interface with nmcli(NetworkManager) as needed
+    my $_host_bridge_cfg = "/etc/NetworkManager/system-connections/br0.nmconnection";
+
+    # Change the NetworkManager log-level as DEBUG at runtime
+    enable_nm_debug;
+
+    if (script_run("[[ -f $_host_bridge_cfg ]]") != 0) {
+        my $_alp_host_bridge = "/root/alp_host_bridge_init.sh";
+        assert_script_run("curl " . data_url("virt_autotest/alp_host_bridge_init.sh") . " -o $_alp_host_bridge");
+        assert_script_run("chmod +rx $_alp_host_bridge && $_alp_host_bridge");
+        save_screenshot;
+        record_info("Host Bridge Network Interface is set successfully for automation test.", script_output("ip a; ip route show all"));
+    }
+    check_activate_network_interface;
+}
+
+sub upload_nm_debug_log {
+    script_run("journalctl -u NetworkManager.service > /tmp/NetworkManager.logs");
+    upload_virt_logs("/tmp/NetworkManager.logs", "NetworkManager-debug-logs");
+    script_run("rm -rf /tmp/NetworkManager.logs");
+}
 
 1;
--- a/tests/virt_autotest/prepare_alp_host.pm
+++ b/tests/virt_autotest/prepare_alp_host.pm
@@ -32,6 +32,9 @@ sub run {
     save_screenshot;
     assert_script_run("ls -latr /var/lib/libvirt/images/");
     record_info("Ignition file is successfully downloaded.");
+
+    # Setup Host Bridge Network Interface as required
+    virt_autotest::utils::set_host_bridge_interface_with_nm;
 }
 
 sub test_flags {
@@ -41,6 +44,8 @@ sub test_flags {
 sub post_fail_hook {
     # Let it empty now.
     # Need to let parent one called after yast team modifies the parent class to fit alp.
+    # Upload expected NetworkManager debug log
+    virt_autotest::utils::upload_nm_debug_log;
 }
 
 1;
--- a/schedule/virt_autotest/alp-virt-core-tests.yaml
+++ b/schedule/virt_autotest/alp-virt-core-tests.yaml
@@ -32,6 +32,7 @@ conditional_schedule:
         ENABLE_VIR_NET:
             1:
                 - virt_autotest/libvirt_virtual_network_init
+                - virt_autotest/libvirt_host_bridge_virtual_network
                 - virt_autotest/libvirt_nated_virtual_network
                 - virt_autotest/libvirt_routed_virtual_network
                 - virt_autotest/libvirt_isolated_virtual_network
--- a/tests/virt_autotest/libvirt_host_bridge_virtual_network.pm
+++ b/tests/virt_autotest/libvirt_host_bridge_virtual_network.pm
@@ -17,16 +17,19 @@ use strict;
 use warnings;
 use testapi;
 use utils;
-use version_utils 'is_sle';
+use version_utils qw(is_sle is_alp);
 
 our $virt_host_bridge = 'br0';
 our $based_guest_dir = 'tmp';
 sub run_test {
     my ($self) = @_;
 
-    #Prepare VM HOST SERVER Network Interface Configuration
-    #for libvirt virtual network testing
-    virt_autotest::virtual_network_utils::prepare_network($virt_host_bridge, $based_guest_dir);
+    # ALP has done this in earlier setup
+    unless (is_alp) {
+        #Prepare VM HOST SERVER Network Interface Configuration
+        #for libvirt virtual network testing
+        virt_autotest::virtual_network_utils::prepare_network($virt_host_bridge, $based_guest_dir);
+    }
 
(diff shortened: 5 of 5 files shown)


guoxuguang:
@alice-suse @Julie-CAO @waynechen55 @nanzhg @RoyCai7 @varunkojha please help me review this PR. thanks

data/virt_autotest/alp_host_bridge_init.sh
```diff
@@ -0,0 +1,56 @@
...
+    ifname ${ACTIVE_DEVICE} master ${BRIDGEIF_NAME}
+}
+
+# Enable Host Bridge Network Interface via nmcli
+enable_host_bridge_network_interface() {
+
+nmcli con up ${BRIDGEIF_NAME}
+
+# refer to bsc#1208005 for more details
+nmcli con down "${ACTIVE_CONNECTION}"
```
alice-suse, line 42: Missing indent?

data/virt_autotest/alp_host_bridge_init.sh
```diff
@@ -0,0 +1,56 @@
...
+# ==== #
+if [[ $ID =~ 'alp' ]]; then
+    echo "Confirm Test Environment: ALP"
+else
+    echo "Error: Please perform this Tool on ALP! Exit... now"
+    exit 1
+fi
+
+setup_host_bridge_network_interface
+enable_host_bridge_network_interface
```
alice-suse, line 56: Nice script!

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -23,6 +23,7 @@ use virt_autotest::utils qw(ssh_setup);
 our @EXPORT = qw(
   set_ct_engine
   set_kvm_container_image
+  set_host_bridge_interface_with_host
```
alice-suse, line 26: This function seems more related with alp host setup, not kvm container. Suggest to call this function in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/virt_autotest/prepare_alp_host.pm#L25.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -162,6 +173,7 @@ sub setup_services_within_kvm_container {
 }
 
 sub setup_kvm_container_from_scratch {
+    set_host_bridge_interface_with_host;
```
alice-suse, line 176: This function seems more related with alp host setup, not kvm container utility function. Suggest to call this function in https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/virt_autotest/prepare_alp_host.pm#L25.

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -58,6 +59,16 @@ sub set_kvm_container_image {
     }
 }
 
+sub set_host_bridge_interface_with_host {
```
alice-suse, line 62: How about moving the definition of this function to https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/virt_autotest/utils.pm?

BTW, I do not quite get the point of setting the function name to xxx_with_host? Would you please explain? If no specific reasons, how about naming it sth like `set_host_bridge_with_nm`?

data/virt_autotest/alp_host_bridge_init.sh
```diff
@@ -0,0 +1,56 @@
...
+IPV4_GATEWAY=$(nmcli -g IP4.GATEWAY con show "${ACTIVE_UUID}")
+IPV4_DNS_A=$(nmcli -g IP4.DNS con show "${ACTIVE_UUID}"| awk '{print $1}')
+IPV4_DNS_B=$(nmcli -g IP4.DNS con show "${ACTIVE_UUID}"| awk '{print $3}')
+
+# Function Definitions
+# --------------------
+# Add Host Bridge Network Interface via nmcli
+setup_host_bridge_network_interface() {
+
+nmcli con add type bridge \
```
alice-suse, line 24: Add indent here?

data/virt_autotest/alp_host_bridge_init.sh
```diff
@@ -0,0 +1,56 @@
...
+    con-name ${BRIDGEIF_NAME} ifname ${BRIDGEIF_NAME} \
+    autoconnect yes ipv4.method manual \
+    ipv4.address ${IPV4_ADDRESS} \
+    ipv4.gateway ${IPV4_GATEWAY} \
+    ipv4.dns ${IPV4_DNS_A} +ipv4.dns ${IPV4_DNS_B} \
+
+nmcli con add type bridge-slave \
```
alice-suse, line 31: So does here?

lib/alp_workloads/kvm_workload_utils.pm
```diff
@@ -58,6 +59,16 @@ sub set_kvm_container_image {
...
+    # Setup Host Bridge Network via NetworkManager as needed
+    my $_alp_host_bridge = "/root/alp_host_bridge_init.sh";
+    assert_script_run("curl " . data_url("virt_autotest/alp_host_bridge_init.sh") . " -o $_alp_host_bridge");
+    assert_script_run("chmod +rx $_alp_host_bridge && $_alp_host_bridge");
```
alice-suse, line 66: What if there is br0 already? Add the logic to skip?

alice-suse (changes requested):
Nice job!

data/virt_autotest/alp_host_bridge_init.sh
```diff
@@ -0,0 +1,56 @@
...
+}
+
+# Enable Host Bridge Network Interface via nmcli
+enable_host_bridge_network_interface() {
+
+nmcli con up ${BRIDGEIF_NAME}
+
+# refer to bsc#1208005 for more details
+nmcli con down "${ACTIVE_CONNECTION}"
+}
```
alice-suse, line 43: How about adding some  checks after bringing up the connection, eg check br0 is up with the desired ip, and ping openqa_url is successful?

lib/alp_workloads/kvm_workload_utils.pm
alice-suse: Besides, when the bridge setup fails, what logs shall we add in post_fail_hook?

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: Good point! will update code format.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: Good point! let me add check point here.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: Awesome! good idea!

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: sure, will update code format.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: Good point! we have the same idea!  will add the basic check for the existed br0 situation.

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: Good point! we have the same idea together! will add the basic check after bringing up the connection.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: well, regarding to the func `set_host_bridge_interface_with_host()`  let me do the explanation here. 
1stly, just compare both with config_host_and_kvm_container() and cleanup_host_and_kvm_container(),  `set_host_bridge_interface_with_host()` just only set up host bridge network interface on alp host without under kvm container.

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: Add indent now.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: Short update, had updated func name as `set_host_bridge_interface_with_nm()` and move it to  https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/virt_autotest/utils.pm

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: Add indent now.

lib/alp_workloads/kvm_workload_utils.pm
guoxuguang: Add the basic check for the existed br0. If so, will skip to perform alp_host_bridge_init.sh script.

guoxuguang: Short update, add new function `check_activate_network_interface()` to check with br0 network interface.

data/virt_autotest/alp_host_bridge_init.sh
guoxuguang: Short update, add new function check_activate_network_interface() to check with br0 network interface.

guoxuguang:
Short update, add the latest validation run:
[alp-virt-core-tests](http://10.67.129.218/tests/373)

alice-suse:
@guoxuguang  I will check into the updated code later. 

Now there is an important question -- what's the pass ratio/stability of this way of solution for setting up bridge( on ssh connection based console,  execute a script which will actually introduce a network outage period and thus has chance that the root-ssh console is broken, some commands return is not got)? I tried to find answer in your openqa historical jobs. But  there are not that many latest successful jobs and it's not easy to tell which failure jobs are by unready devel code, and which are by such root cause. It will be crucial to have such data before we can decide if this solution can be used with high enough pass ratio.

lib/virt_autotest/utils.pm
```diff
@@ -943,5 +942,44 @@ sub download_vm_import_disks {
...
 
+sub enable_nm_debug {
+    # Enable Network Manager Debug Log Level
+    assert_script_run("nmcli general logging level DEBUG domains ALL", 60);
+    record_info("Enable Network Manager in Debug Level successfully for automation test.");
+}
+
+sub check_activate_network_interface {
+    # Check with activate network interface as required
+    my $target = "openqa.suse.de";
```
alice-suse, line 953: We may need different target for OSD and O3 (by OPENQA_URL setting).

lib/virt_autotest/utils.pm
```diff
@@ -943,5 +942,44 @@ sub download_vm_import_disks {
...
+    my $nic = "br0";
```
alice-suse, line 954: If this func is put in utils.pm, maybe good to have $nic as an input parameter?

lib/virt_autotest/utils.pm
```diff
@@ -943,5 +942,44 @@ sub download_vm_import_disks {
...
+}
+
+sub set_host_bridge_interface_with_nm {
+    # Setup Host Bridge Network Interface with nmcli(NetworkManager) as needed
+    my $_host_bridge_cfg = "/etc/NetworkManager/system-connections/br0.nmconnection";
+
+    # Change the NetworkManager log-level as DEBUG at runtime
+    enable_nm_debug;
+
+    if (script_run("[[ -f $_host_bridge_cfg ]]") != 0) {
```
alice-suse, line 968: We can only save br0 setup if br0 is working well, right? Checking nmconnection file will hide some not-working scenarios?

lib/virt_autotest/utils.pm
guoxuguang: Wow, make sense now. will update to use with OPENQA_URL setting.

lib/virt_autotest/utils.pm
guoxuguang: Great! it is a good idea. will update to have $nic as an input parameter.

lib/virt_autotest/utils.pm
guoxuguang: Emm,  good point.  just only check with br0.nmconnection would be not enough as expected.  we should double check with br0-slave.nmconnection and br0.nmconnection files together.

alice-suse: Sorry, please ignore comment on this, `check_activate_network_interface` is called anyway , so should be fine.

alice-suse:
> @guoxuguang I will check into the updated code later.
> 
> Now there is an important question -- what's the pass ratio/stability of this way of solution for setting up bridge( on ssh connection based console, execute a script which will actually introduce a network outage period and thus has chance that the root-ssh console is broken, some commands return is not got)? I tried to find answer in your openqa historical jobs. But there are not that many latest successful jobs and it's not easy to tell which failure jobs are by unready devel code, and which are by such root cause. It will be crucial to have such data before we can decide if this solution can be used with high enough pass ratio.

@guoxuguang  Hi Leon, any update on this? Do you have time estimation needed for it? It seems we can have a quick merge soon, after we can confirm the solution stability with enough times of jobs.

lib/virt_autotest/utils.pm
guoxuguang: update to use with OPENQA_URL setting

lib/virt_autotest/utils.pm
guoxuguang: Update to use with `$network_interface` ($nic) as an input parameter

guoxuguang:
> > @guoxuguang I will check into the updated code later.
> > Now there is an important question -- what's the pass ratio/stability of this way of solution for setting up bridge( on ssh connection based console, execute a script which will actually introduce a network outage period and thus has chance that the root-ssh console is broken, some commands return is not got)? I tried to find answer in your openqa historical jobs. But there are not that many latest successful jobs and it's not easy to tell which failure jobs are by unready devel code, and which are by such root cause. It will be crucial to have such data before we can decide if this solution can be used with high enough pass ratio.
> 
> @guoxuguang Hi Leon, any update on this? Do you have time estimation needed for it? It seems we can have a quick merge soon, after we can confirm the solution stability with enough times of jobs.

yes, through the process of debugging Network Manager, I have discovered many issues that have been overlooked in the details. For example, how to debug Network Manager product issues. I have recently spent some time working on these issues and have asked our NM maintenance colleagues many times and have received a lot of positive feedback. In the meantime, the function `enable_nm_debug() `was created to enable Network Manager in debug mode while Network Manager is running statue, so that we can get more debug information about Network Manager in case of an Network Manager exception. Similarly, when an Network Manager exception is found, the newly created function `upload_nm_debug_log()` can be used to collect the Network Manager debug log as we expect - this log file has been confirmed several times with our Network Manager maintenance colleagues together. 

Through this period of research, some experience has been gained when it comes to debugging Network Manager problems. At the moment, stay with the local validation runs. If it goes well, hope that this pr will be merge next week.

guoxuguang:
WIP [setup host bridge network interface](http://10.67.129.218/tests/393#)

lib/virt_autotest/utils.pm
```diff
@@ -943,5 +942,44 @@ sub download_vm_import_disks {
...
+        my $_alp_host_bridge = "/root/alp_host_bridge_init.sh";
+        assert_script_run("curl " . data_url("virt_autotest/alp_host_bridge_init.sh") . " -o $_alp_host_bridge");
+        assert_script_run("chmod +rx $_alp_host_bridge && $_alp_host_bridge");
+        save_screenshot;
+        record_info("Host Bridge Network Interface is set successfully for automation test.", script_output("ip a; ip route show all"));
+    }
+    check_activate_network_interface;
```
waynechen55, line 976: Is this checking ```check_activate_network_interface``` reliable ? For example, what is the behavior when activating br0 fails ?

guoxuguang: As usual, it will call with the func  `post_fail_hook()` -> `upload_nm_debug_log()` to create and upload expected Network Manager debug log as expected when activating br0 fails .

guoxuguang:
> @guoxuguang I will check into the updated code later.
> 
> Now there is an important question -- what's the pass ratio/stability of this way of solution for setting up bridge( on ssh connection based console, execute a script which will actually introduce a network outage period and thus has chance that the root-ssh console is broken, some commands return is not got)? I tried to find answer in your openqa historical jobs. But there are not that many latest successful jobs and it's not easy to tell which failure jobs are by unready devel code, and which are by such root cause. It will be crucial to have such data before we can decide if this solution can be used with high enough pass ratio.

@alice-suse 
Good point,  currently ,  depend on the current way of the solution for setting up host bridge. Have performed almost 10 times fully local validation run as below:
[alp-virt-core-tests](http://10.67.129.218/tests/393) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/396) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/397) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/398) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/399) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/404) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/405) PASS
[alp-virt-core-tests](http://10.67.129.218/tests/406) PASS
[alp-virt-core-tests]() PASS
[alp-virt-core-tests]() PASS

tests/virt_autotest/prepare_alp_host.pm
```diff
@@ -41,6 +44,8 @@ sub test_flags {
 sub post_fail_hook {
     # Let it empty now.
     # Need to let parent one called after yast team modifies the parent class to fit alp.
+    # Upload expected NetworkManager debug log
+    virt_autotest::utils::upload_nm_debug_log;
```
alice-suse, line 48: @guoxuguang  Hi Leon, during our offline talk this Tuesday, you mentioned that you were considering how to upload failure logs when it fails to set up bridge for whatever reasons, eg network manager bugs. I talked with Julie to see if there can be any experience to be referred from sriov test which has chance for similar network outage during VF operations. The result is that there is NO such handling either in sriov. And we discussed the possibility to use sol console for upload -- possible but no guaranteed way because sol console does not have return code given to openqa. We two agree that if the chance for such issue is low in your validations, we can have this PR merged WITHOUT it. If in official openqa test, it turns out that it happens with possibility higher than acceptable level, you can add failure debug log in post_fail_hook via sol console then. HDYT?

guoxuguang: Totally agree on this as your mentioned.  thanks so much for your great help and suggestion! I do appreciate that.

alice-suse (approved):
LGTM. Congratulations!

alice-suse:
Seems all conversations are resolved. Merge it.

PR #1661, opened 2016-08-08
Wait longer for grub2 screen

dzedro:
poo#11908

I would ignore this tidy fail because it's wrong in this case
It would put comment line 26 on same level as comment on line 25

PR #16661, opened 2023-03-20
Qesap collect logs in case of failure

mpagot:
Collect logs from each HANA cluster node in case of failure
Change all the Ansible API in qesap lib to support failok mode
Change the script_output Ansible API in qesap also needs a change in the playbook
Move the wait_ssh from qesap regression test module to an API function in qesap
Align input validation in qesap_execute to all the other API of the qesap lib
Add tons of UT about these changes

Related ticket: TEAM-7152
Verification run: 

 - AZURE : 
 1. http://openqaworker15.qa.suse.cz/tests/138655 it is again a PASS and here again an example of log http://openqaworker15.qa.suse.cz/tests/138655/logfile?filename=test_cluster-crm_status.txt
 2. [LATEST CODE] http://openqaworker15.qa.suse.cz/tests/139091

 - GCP :  it is fail but it does not have logs. It is because the PR code is using fixed Ansible name for the two HANA hosts (like in [here](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16661/files#diff-23c988d39d71dc252bb5dce95a699047e93941078f2152ab4a6587286e9a4e47R671) ) but GCP Terraform wrongly include deployment name in them
 1. http://openqaworker15.qa.suse.cz/tests/138657
 2. http://openqaworker15.qa.suse.cz/tests/138661
 3. http://openqaworker15.qa.suse.cz/tests/138662
 2. [LATEST CODE] http://openqaworker15.qa.suse.cz/tests/139092
 
- AWS :
 1. http://openqaworker15.qa.suse.cz/tests/138654 it is a PASS one. Here an example of logs uploaded from the `run` of the test_cluster module http://openqaworker15.qa.suse.cz/tests/138654/logfile?filename=test_cluster-journalctl.txt
 2. http://openqaworker15.qa.suse.cz/tests/138658
 3. http://openqaworker15.qa.suse.cz/tests/138659
 4. http://openqaworker15.qa.suse.cz/tests/138660
 5. [LATEST CODE] http://openqaworker15.qa.suse.cz/tests/139090

### DIFF
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -411,15 +411,19 @@ sub qesap_prepare_env {
 
 =item B<FILTER> - filter hosts in the inventory
 
+=item B<FAILOK> - if not set, ansible failure result in die
+
 =back
 =cut
 
 sub qesap_ansible_cmd {
     my (%args) = @_;
     croak 'Missing mandatory cmd argument' unless $args{cmd};
-    my $inventory = qesap_get_inventory($args{provider});
     $args{user} ||= 'cloudadmin';
     $args{filter} ||= 'all';
+    $args{failok} ||= 0;
+
+    my $inventory = qesap_get_inventory($args{provider});
 
     my $ansible_cmd = join(' ',
         'ansible',
@@ -428,8 +432,14 @@ sub qesap_ansible_cmd {
         '-u', $args{user},
         '-b', '--become-user=root',
         '-a', "\"$args{cmd}\"");
-    enter_cmd("source " . QESAPDEPLOY_VENV . "/bin/activate");
-    assert_script_run($ansible_cmd);
+    assert_script_run("source " . QESAPDEPLOY_VENV . "/bin/activate");
+
+    if ($args{failok}) {
+        script_run($ansible_cmd);
+    }
+    else {
+        assert_script_run($ansible_cmd);
+    }
     enter_cmd("deactivate");
 }
 
@@ -460,6 +470,8 @@ sub qesap_ansible_cmd {
 
 =item B<ROOT> - 1 to enable remote execution with elevated user, default to 0
 
+=item B<FAILOK> - if not set, ansible failure result in die
+
 =back
 =cut
 
@@ -469,6 +481,7 @@ sub qesap_ansible_script_output {
     croak 'Missing mandatory host argument' unless $args{host};
     $args{user} ||= 'cloudadmin';
     $args{root} ||= 0;
+    $args{failok} ||= 0;
 
     my $inventory = qesap_get_inventory($args{provider});
 
@@ -494,8 +507,14 @@ sub qesap_ansible_script_output {
         '-e', "out_file='$local_file'");
 
     enter_cmd "rm $local_tmp";
-    enter_cmd("source " . QESAPDEPLOY_VENV . "/bin/activate");    # venv activate
-    assert_script_run(join(' ', @ansible_cmd));
+    assert_script_run("source " . QESAPDEPLOY_VENV . "/bin/activate");    # venv activate
+
+    if ($args{failok}) {
+        script_run(join(' ', @ansible_cmd));
+    }
+    else {
+        assert_script_run(join(' ', @ansible_cmd));
+    }
     enter_cmd("deactivate");    #venv deactivate
     my $output = script_output("cat $local_tmp");
     enter_cmd "rm $local_tmp";
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -63,6 +63,9 @@ our @EXPORT = qw(
   qesap_create_aws_credentials
   qesap_create_aws_config
   qesap_remote_hana_public_ips
+  qesap_wait_for_ssh
+  qesap_cluster_log_cmds
+  qesap_cluster_logs
 );
 
 =head1 DESCRIPTION
@@ -269,7 +272,7 @@ sub qesap_yaml_replace {
 
 sub qesap_execute {
     my (%args) = @_;
-    die 'QESAP command to execute undefined' unless $args{cmd};
+    croak 'Missing mandatory cmd argument' unless $args{cmd};
 
     my $verbose = $args{verbose} ? "--verbose" : "";
     my %paths = qesap_get_file_paths();
@@ -421,7 +424,6 @@ sub qesap_ansible_cmd {
     croak 'Missing mandatory cmd argument' unless $args{cmd};
     $args{user} ||= 'cloudadmin';
     $args{filter} ||= 'all';
-    $args{failok} ||= 0;
 
     my $inventory = qesap_get_inventory($args{provider});
 
@@ -434,15 +436,10 @@ sub qesap_ansible_cmd {
         '-a', "\"$args{cmd}\"");
     assert_script_run("source " . QESAPDEPLOY_VENV . "/bin/activate");
 
-    if ($args{failok}) {
-        script_run($ansible_cmd);
-    }
-    else {
-        assert_script_run($ansible_cmd);
-    }
+    $args{failok} ? script_run($ansible_cmd) : assert_script_run($ansible_cmd);
+
     enter_cmd("deactivate");
 }
-
 
 =head3 qesap_ansible_script_output
 
@@ -458,7 +455,10 @@ sub qesap_ansible_cmd {
     3. the playbook download the file locally
     4. the file is read and stored to be returned to the caller
 
-=over 5
+    If local_file and local_path are specified, the output is written to file, return is the full path;
+    otherwise the return is the command output as string.
+
+=over 8
 
 =item B<PROVIDER> - Cloud provider name, used to find the inventory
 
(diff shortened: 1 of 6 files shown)


lib/qesapdeployment.pm
```diff
@@ -545,4 +585,102 @@ sub qesap_remote_hana_public_ips {
...
+  Collect logs from a deployed cluster
+
+=cut
+
+sub qesap_cluster_logs {
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $inventory = qesap_get_inventory($prov);
+    if (script_run("test -e $inventory") == 0)
+    {
+        # sleep bmwqemu::scale_timeout(300)
```
alvarocarvajald, line 670: Dangling comment.

lib/qesapdeployment.pm
```diff
@@ -545,4 +585,102 @@ sub qesap_remote_hana_public_ips {
...
+        foreach my $host ('vmhana01', 'vmhana02') {
+            foreach my $cmd (qesap_cluster_log_cmds()) {
```
alvarocarvajald, line 672: Is there an advantage with having this `qesap_cluster_log_cmds` which returns a HASH vs. defining the HASH as a variable inside `qesap_cluster_logs`?

tests/sles4sap/qesapdeployment/deploy.pm
```diff
@@ -10,37 +10,22 @@ use Mojo::Base 'publiccloud::basetest';
 use testapi;
 use qesapdeployment;
 
-sub wait_for_ssh {
-    my ($host) = @_;
-    my $timeout //= bmwqemu::scale_timeout(600);
```
alvarocarvajald, line 15: See? It was using `//=` :P

tests/sles4sap/qesapdeployment/deploy.pm
```diff
@@ -10,37 +10,22 @@ use Mojo::Base 'publiccloud::basetest';
...
+    foreach my $host (@remote_ips) {
+        die 'Timed out while waiting for ssh to be available in the CSP instances' if qesap_wait_for_ssh(host => $host) == -1;
+    }
     $ret = qesap_execute(cmd => 'ansible', cmd_options => '--profile', verbose => 1, timeout => 3600);
-    die "'qesap.py ansible' return: $ret" if ($ret);
+    if ($ret)
+    {
+        qesap_cluster_logs();
+        die "'qesap.py ansible' return: $ret";
+    }
```
alvarocarvajald, line 28: Why not keep the line as `die "'qesap.py ansible' return: $ret" if ($ret);` and add the call to `qesap_cluster_logs()` in the `post_fail_hook`?

tests/sles4sap/qesapdeployment/deploy.pm
```diff
@@ -50,7 +35,11 @@ sub test_flags {
 sub post_fail_hook {
     my ($self) = shift;
     qesap_upload_logs();
-    qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300);
+    my $inventory = qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER'));
+    if (script_run("test -e $inventory") == 0)
+    {
+        qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300);
+    }
```
alvarocarvajald, line 42: Suggestion:
```
qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300) unless (script_run("test -e $inventory"));
```

tests/sles4sap/qesapdeployment/destroy.pm
```diff
@@ -14,10 +14,13 @@ use qesapdeployment;
...
 
-    my $ret = qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300);
-    die "'qesap.py ansible -d' return: $ret" if ($ret);
-    $ret = qesap_execute(cmd => 'terraform', cmd_options => '-d', verbose => 1, timeout => 1200);
-    die "'qesap.py terraform -d' return: $ret" if ($ret);
+    my $ansible_ret = qesap_execute(cmd => 'ansible', cmd_options => '-d', verbose => 1, timeout => 300);
+    if ($ansible_ret) {
+        qesap_cluster_logs();
+    }
+    my $terraform_ret = qesap_execute(cmd => 'terraform', cmd_options => '-d', verbose => 1, timeout => 1200);
```
alvarocarvajald, line 21: Oh, I like this! Old way was not calling on `terraform destroy` if the ansible cleanup failed. Good catch!

lib/qesapdeployment.pm
mpagot: yes, It is the first step to unify the code with `ha_export_logs` that I plan to split in two functions too

tests/sles4sap/qesapdeployment/deploy.pm
mpagot: generally speaking I'm trying a new idea here. It is about keep all the error handling of the expected errors in the `run` part. I like to leave only the handling of the  `unexpected` events in the `post_fail_hook` ... like an error during the log collection. I can survive if logs are not collected at a failure, but I like to try to organize the code to avoid that anything could prevent the deployment destroy to be executed. Does it make sense?

lib/qesapdeployment.pm
```diff
@@ -545,4 +585,102 @@ sub qesap_remote_hana_public_ips {
...
+            Cmd => 'lsblk -i -a',
+            Logs => ['lsblk.txt'],
+        },
+        {
+            Cmd => 'journalctl -b --no-pager -o short-precise',
+            Logs => ['journalctl.txt'],
+        },
+        {
+            Cmd => 'systemctl --no-pager --full status sbd',
+            Logs => ['sbd.txt'],
```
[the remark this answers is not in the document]
mpagot: I like this to be merged with `ha_cluster_logs, and here you see an example of a single command collecting two log files at time https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/79441e7d5b1211ddc3f4af39c4194001058cb6cd/lib/hacluster.pm#L578

lib/qesapdeployment.pm
mpagot: Removed but I like to talk about it. The point is: "should we wait for some time before to collect the logs?" Maybe we are calling this function as result of an Ansible failure, and maybe the particular  failure is only due to some timing issue like Ansible to give up to early before something happens. Do we like to collect logs straight after the failure and maybe during the state transition of the cluster? Or should we wait for a more steady state before to collect the logs?

lib/qesapdeployment.pm
```diff
@@ -545,4 +585,102 @@ sub qesap_remote_hana_public_ips {
...
+        # sleep bmwqemu::scale_timeout(300)
+        foreach my $host ('vmhana01', 'vmhana02') {
+            foreach my $cmd (qesap_cluster_log_cmds()) {
+                my $out = qesap_ansible_script_output(cmd => $cmd->{Cmd},
+                    provider => $prov,
+                    host => $host,
+                    failok => 1,
+                    root => 1,
+                    local_path => '/tmp/',
+                    local_file => @{$cmd->{Logs}}[0]);
```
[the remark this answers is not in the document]
mpagot: > First of all: why define Logs as a list if it's only using the first element?

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16661#discussion_r1144363393

> Also, as you're only using the first element, a better way to write what's above is to do

ok

> qesap_ansible_script_output expect a string

I decided to use a list of string as `ha_cluster_logs` in some circumstance needs them. On the other side `ha_cluster_logs` implement the log collection using `script_run` and `upload_logs`. It is not a problem for both of them that more than a log is generated and need to be collected. But  `qesap_ansible_script_output` has a strong requirement and limitation only to have one file, deeply related to how it is implemented. So I like to ideally define a generic data structure to describe the list of logs (I also plan to add Timeout field) but `qesap_ansible_script_output` will be only able to "partially" consume it. 

> My suggestions to fix this are either of the following 2:

Let say that I do not like the 1. I do not fully understand the second one.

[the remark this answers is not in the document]
mpagot: >  I see potential issues with this

I'm adding a unit testing to cover this one.

lib/qesapdeployment.pm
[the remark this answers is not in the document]
alvarocarvajald: Oh yes, the `hb_report` command that saves a explicit log and it's not saving the command output.

One idea would be to add the concept of command output besides logs, and you use use the output value to save the output of the command for `qesap_ansible_script_output`, and then upload as logs the output as well as everything listed in logs.

Something like this:
```
{
   Cmd => "hb_report $report_opt -E $bootstrap_log $hb_log",
   Output => 'hb_report.out',
   Logs => [$bootstrap_log, $hb_log]
}
```

lib/qesapdeployment.pm
alvarocarvajald: In general it's not a good idea to use `sleep` in the tests (I know: sometimes it cannot be avoided).

Isn't there some command we can use to know ansible is responsive before collecting the logs?

lib/qesapdeployment.pm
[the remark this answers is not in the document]
alvarocarvajald: > > qesap_ansible_script_output expect a string
> 
> I decided to use a list of string as `ha_cluster_logs` in some circumstance needs them. On the other side `ha_cluster_logs` implement the log collection using `script_run` and `upload_logs`. It is not a problem for both of them that more than a log is generated and need to be collected. But `qesap_ansible_script_output` has a strong requirement and limitation only to have one file, deeply related to how it is implemented. So I like to ideally define a generic data structure to describe the list of logs (I also plan to add Timeout field) but `qesap_ansible_script_output` will be only able to "partially" consume it.
>

Got it. Good idea.
 
> Let say that I do not like the 1. I do not fully understand the second one.

I guess if you go with [#1661 comment](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16661#discussion_r1144581857), then neither suggestion is relevant anymore.

tests/sles4sap/qesapdeployment/deploy.pm
alvarocarvajald: Yes, it does. But keep in mind that the `die` there is going to call the `post_fail_hook` ... could we end up with a test module that it's uploading the same set of logs twice? Wouldn't that be confusing?

lib/qesapdeployment.pm
```diff
@@ -545,4 +585,102 @@ sub qesap_remote_hana_public_ips {
...
+
+=item B<PORT> - port to probe, default is 22
+
+=back
+=cut
+
+sub qesap_wait_for_ssh {
+    my (%args) = @_;
+    croak 'Missing mandatory host argument' unless $args{host};
+    $args{timeout} ||= bmwqemu::scale_timeout(600);
```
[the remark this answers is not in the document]
alvarocarvajald: "Wait Forever" or "Zero Timeout" depending on the implementation I suppose; for example you can tell `check_screen` to work with zero timeout so it returns faster.

alvarocarvajald (approved):
LGTM. Just a minor recommendation which can be ignored.

PR #16669, opened 2023-03-21
kernel/bpf: Add bpftrace smoke test

richiejp:
Add a basic test for eBPF and user land tooling.

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -2551,6 +2551,7 @@ sub load_extra_tests_kernel {
     loadtest "kernel/module_build";
     loadtest "kernel/tuned";
     loadtest "kernel/fwupd" if is_sle('15+');
+    loadtest "kernel/bpftrace" if is_tumbleweed || is_sle('>=15-sp5');
 }
 
 # Scheduling set for validation of specific installation
--- a/tests/kernel/bpftrace.pm
+++ b/tests/kernel/bpftrace.pm
@@ -1,0 +1,104 @@
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Package: bpftrace
+# Summary: Compile and attach eBPF probes with bpftrace
+# Maintainer: kernel-qa@suse.de
+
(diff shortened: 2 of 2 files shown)


coolgw:
Is there any openqa run case result?

lib/main_common.pm
```diff
@@ -2557,6 +2557,7 @@ sub load_extra_tests_kernel {
     loadtest "kernel/module_build";
     loadtest "kernel/tuned";
     loadtest "kernel/fwupd" if is_sle('15+');
+    loadtest "kernel/bpftrace" if is_tumbleweed;
```
czerw, line 2560: We need to run it on 15-sp5 as well.

richiejp: Thanks, I was still looking for the ticket which had the SP

czerw:
> Maybe you can add it to extra_tests_kernel https://openqa.suse.de/tests/10719901 https://openqa.opensuse.org/tests/3183911 for TW and 15-SP5+.

Please do verification runs from above jobs by openqa-clone-custom-git-refspec.

tests/kernel/bpftrace.pm
```diff
@@ -0,0 +1,65 @@
...
+      threadsnoop.bt
+      undump.bt
+      vfscount.bt
+      vfsstat.bt
+      writeback.bt
+      xfsdist.bt
+    );
+
+    foreach my $t (@tests) {
+        script_run("timeout --preserve-status -s SIGINT 5 bpftrace $tools_dir/$t");
```
czerw, line 61: It will just run command without any return value check or  comparison with expected value. It will run, but if something fails, we will not spot it (i think tests will be all the time green).  Could you you please add one or two tests which will compare script output with some expected result?

richiejp:
I split the tests in two lists, some are asserted, some not. It seems like that more will need to be added to the unasserted list.

richiejp:
New verfication runs:
TW: https://openqa.opensuse.org/tests/3220748#
SLE: https://openqa.suse.de/tests/10896636#

pevik (approved):
Thanks!

PR #16670, opened 2023-03-21
Create softfailure for 'snapper_cleanup' test

rfan1:
We need to use the same workaround for test 'snapper_used_space' due to job dependency
https://progress.opensuse.org/issues/126260
https://bugzilla.opensuse.org/show_bug.cgi?id=1206814

- Verification run: https://openqa.opensuse.org/tests/3184468


tests/console/snapper_cleanup.pm
```diff
@@ -48,7 +48,12 @@ sub snapper_cleanup {
     for (1 .. $scratch_size_gb) { assert_script_run("$snap_create", 500); }
     assert_script_run('sync');
     script_run "echo There are `$snaps_numb` snapshots BEFORE cleanup";
-    assert_script_run("snapper cleanup number", 300);    # cleanup created snapshots
+    # cleanup created snapshots
+    # softfail the test if hitting bsc#1206814, so that the subsequent test runs are not affected
+    if ((script_run('snapper cleanup number', timeout => 300) != 0) && is_leap) {
```
grisu48, line 53: This has two issues:

1. Any issue on leap would now get marked as `bsc#1206814`
2. Any other issue on any other system would be ignored.

As a suggestion: On Leap - Check if we can use `script_output` to check, if the command was successful. If unsuccessful, check if `Failure (error.something).` is therein and then record the softfailure.  
On all other systems - use `assert_script_run`.

grisu48 (changes requested):
This change is problematic, as it would make us blind for any issues in `snapper cleanup number`.

tests/console/snapper_cleanup.pm
rfan1: You are correct! 
I will use script_output

tests/console/snapper_cleanup.pm
```diff
@@ -48,7 +48,18 @@ sub snapper_cleanup {
     for (1 .. $scratch_size_gb) { assert_script_run("$snap_create", 500); }
     assert_script_run('sync');
     script_run "echo There are `$snaps_numb` snapshots BEFORE cleanup";
-    assert_script_run("snapper cleanup number", 300);    # cleanup created snapshots
+    # cleanup created snapshots
+    # softfail the test if hitting bsc#1206814, so that the subsequent test runs are not affected
+    my $output = script_output('snapper cleanup number', 300, proceed_on_failure => 1);
```
grisu48, line 53: Careful!! `script_output` only redirects stdout, not stderr. If we're using `$output ne ''` later on and there would be an error written to stderr (which it should be) we are not detecting the error!

A quick solution would be to do the following:

```suggestion
    my $output = script_output('snapper cleanup number || echo "error detected"', 300, proceed_on_failure => 1);
```

That prints "error detected" in all error cases to stdout and we can check for this string in line 54.

grisu48 (changes requested):

tests/console/snapper_cleanup.pm
rfan1: Hmm
`
proceed_on_failure - allows to proceed with validation when $script is failing (return non-zero exit code)
`

rfan1: Based on the VRs, the codes worked fine for both passed and failed tests :)

grisu48: Ah, I got the `if` block wrong, sorry. But still, we would not detect a failure where something is written and then `snapper` dies.

tests/console/snapper_cleanup.pm
```diff
@@ -48,7 +48,18 @@ sub snapper_cleanup {
...
+    # cleanup created snapshots
+    # softfail the test if hitting bsc#1206814, so that the subsequent test runs are not affected
+    my $output = script_output('snapper cleanup number || echo "error detected"', 300, proceed_on_failure => 1);
+    if ($output =~ m/error detected/) {
+        if (($output =~ m/Failure\s+\(error\.something\)\./) && is_leap) {
+            record_soft_failure 'bsc#1206814';
+            return;
+        }
+        else {
+            die 'snapper cleanup number failed';
```
grisu48, line 60: This should not be in the `else`.

rfan1: fixed, thanks @grisu48

grisu48 (approved):

PR #16672, opened 2023-03-21
Use table for badges in Readme

asdil12:


foursixnine (approved):

foursixnine:
Sadly GH keeps caching the badges

asdil12:
I think this is something we can improve with caching headers:
https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls

PR #16676, opened 2023-03-22
Fix wrong mount point for swap

skliu:
Fix wrong mount point for swap in autoyast_home_encrypted.xml

**Ticket:**

- https://progress.opensuse.org/issues/126227

**Verification run:**

- https://openqa.suse.de/tests/10744985


rakoenig (commented):
LGTM

PR #16677, opened 2023-03-22
ulp_openposix: Fix libc version regex

mdoucha:
If the livepatch filename is libc_livepatch*.so, the old regex could match across multiple lines. Exclude parantheses which are added to the RPM filename list to prevent matching across lines.

- Needles: N/A
- Verification run: https://openqa.suse.de/tests/10745529


mdoucha:
The verification run failed as expected with the correct error message. This PR is ready for merge.

PR #16684, opened 2023-03-22
Disable IPv6 for samba/AD test run

grisu48:
Some of our workers cannot reach the AD controller via IPv6, so we need to disable IPv6 for this test run.

- Related ticket: https://progress.opensuse.org/issues/96512
- Verification run: https://openqa.suse.de/tests/10748272

### DIFF
--- a/tests/network/samba/samba_adcli.pm
+++ b/tests/network/samba/samba_adcli.pm
@@ -116,6 +116,14 @@ sub randomize_hostname {
     assert_script_run("hostnamectl set-hostname '$hostname'");
 }
 
+sub disable_ipv6 {
+    assert_script_run('sysctl -w net.ipv6.conf.all.disable_ipv6=1');
+}
+
+sub enable_ipv6 {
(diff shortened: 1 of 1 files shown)


tests/network/samba/samba_adcli.pm
```diff
@@ -125,6 +133,7 @@ sub run {
 
     samba_sssd_install();
     randomize_hostname();    # Prevent race condition with parallel test runs
+    disable_ipv6();    # AD host is not reachable via IPv6 on some of our workers
```
foursixnine, line 136: I recall this was also done for s390 too... do we have a POO ticket to reference the issue? no need to soft-fail, but ipv6 should work too..

foursixnine (approved):

tests/network/samba/samba_adcli.pm
grisu48: Now we do :slightly_smiling_face:  https://progress.opensuse.org/issues/126482

PR #16694, opened 2023-03-23
selinux on ALP is already enabled and in enforcing mode by default

tjyrinki:
These two modules were already skipped in the YAML schedule version of this.

- Related ticket: https://progress.opensuse.org/issues/126359
- Verification run: https://openqa.opensuse.org/tests/3186924


paolostivanin:
LGTM

PR #16696, opened 2023-03-24
Set ssh serial console alive and check kvm modules

Julie-CAO:
- check kvm modules to avoid the case happens again: bsc#1209596
- keep ssh serial console always alive to get test output after a long run

- Verification run: 
[host_upgrade from Wayne](https://openqa.suse.de/tests/10758433)
[guest-migration-developing-from-developing-to-developing-kvm](https://openqa.suse.de/tests/10761711)   //2.5 hours for guest migration. It should be convincing.

### DIFF
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -15,7 +15,7 @@ use testapi;
 use Utils::Architectures;
 use Utils::Backends qw(use_ssh_serial_console is_remote_backend set_ssh_console_timeout);
 use ipmi_backend_utils;
-use virt_autotest::utils qw(is_xen_host check_port_state check_host_health);
+use virt_autotest::utils qw(is_xen_host is_kvm_host check_port_state check_host_health);
 use IPC::Run;
 
 sub set_ssh_console_timeout_before_use {
@@ -23,7 +23,7 @@ sub set_ssh_console_timeout_before_use {
     select_console('root-console');
     set_ssh_console_timeout('/etc/ssh/sshd_config', '28800');
     reset_consoles;
-    select_console 'sol', await_console => 1;
+    select_console 'sol', await_console => 0;
     send_key 'ret';
     check_screen([qw(linux-login virttest-displaymanager)], 60);
     save_screenshot;
@@ -47,6 +47,16 @@ sub double_check_xen_role {
     record_info 'INFO', 'Check if start bootloader from a read-only snapshot';
     assert_script_run('touch /root/read-only.fs && rm -rf /root/read-only.fs');
     save_screenshot;
+}
+
+sub check_kvm_modules {
+    if (script_run('lsmod | grep "^kvm\b"') == 0 and script_run('lsmod | grep -e "^kvm_intel\b" -e "^kvm_amd\b"') == 0) {
+        save_screenshot;
+        record_info("KVM", "kvm and kvm_intel/amd modules are loaded");
+    }
+    else {
+        die "KVM modules are not loaded!";
+    }
 }
 
 #Explanation for parameters introduced to facilitate offline host upgrade:
@@ -191,12 +201,14 @@ sub login_to_console {
         send_key 'ret';
     }
 
-    # Set ssh console timeout for thunderx machine
-    set_ssh_console_timeout_before_use if (is_remote_backend && is_aarch64 && get_var('IPMI_HW') eq 'thunderx');
+    # Set ssh console timeout for virt tests on ipmi backend machines
+    # it will make ssh serial console alive even with long time command
+    set_ssh_console_timeout_before_use if (is_remote_backend and is_x86_64 and get_var('VIRT_AUTOTEST', ''));
     # use console based on ssh to avoid unstable ipmi
     use_ssh_serial_console;
     # double-check xen role for xen host
     double_check_xen_role if (is_xen_host and !get_var('REBOOT_AFTER_UPGRADE'));
+    check_kvm_modules if is_x86_64 and is_kvm_host and !get_var('REBOOT_AFTER_UPGRADE');
 }
 
 sub run {
--- a/lib/Utils/Backends.pm
+++ b/lib/Utils/Backends.pm
@@ -263,8 +263,15 @@ sub is_generalhw { check_var('BACKEND', 'generalhw'); }
 sub set_ssh_console_timeout {
     my ($sshd_config_file, $sshd_timeout) = @_;
     my $client_count_max = $sshd_timeout / 60;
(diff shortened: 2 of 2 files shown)


waynechen55:
I think this [verification run](https://openqa.suse.de/tests/10758433) is convincing. It passed host_upgrade_step2_run which took around 2hrs.

For guest upgrade, I do think these two test runs [1](https://openqa.suse.de/tests/10759082) and [2](https://openqa.suse.de/tests/10759525) are convincing as above, because I did not see a failure with these two machines recently (This does not mean issue comes from SUT machine). But setting this should does no harm anyway from my perspective.  I think more verification runs with guest upgrade needed.

waynechen55:
I think this [verification run](https://openqa.suse.de/tests/10758433) is convincing. It passed host_upgrade_step2_run which took around 2hrs.

For guest upgrade, I do think these two test runs [1](https://openqa.suse.de/tests/10759082) and [2](https://openqa.suse.de/tests/10759525) are convincing as above, because I did not see a failure with these two machines recently (This does not mean issue comes from SUT machine). But setting this should does no harm anyway from my perspective.  I think more verification runs with guest upgrade needed. @guoxuguang

Julie-CAO:
@alice-suse @xguo @waynechen55 @nanzhg @tbaev @RoyCai7 @varunkojha  @tonyyuan1   Welcome review!

waynechen55:
I am verifying guest upgrade with your branch

tests/virt_autotest/login_console.pm
```diff
@@ -191,12 +201,14 @@ sub login_to_console {
...
-    # Set ssh console timeout for thunderx machine
-    set_ssh_console_timeout_before_use if (is_remote_backend && is_aarch64 && get_var('IPMI_HW') eq 'thunderx');
+    # Set ssh console timeout for virt tests on ipmi backend machines
+    # it will make ssh serial console alive even with long time command
+    set_ssh_console_timeout_before_use if (is_remote_backend and is_x86_64 and get_var('VIRT_AUTOTEST', ''));
     # use console based on ssh to avoid unstable ipmi
     use_ssh_serial_console;
     # double-check xen role for xen host
     double_check_xen_role if (is_xen_host and !get_var('REBOOT_AFTER_UPGRADE'));
+    check_kvm_modules if is_kvm_host and !get_var('REBOOT_AFTER_UPGRADE');
```
Julie-CAO, line 211: Is it ok for s390 or arm tests? @guoxuguang  @waynechen55

tests/virt_autotest/login_console.pm
```diff
@@ -15,15 +15,15 @@ use testapi;
...
+use virt_autotest::utils qw(is_xen_host is_kvm_host check_port_state check_host_health);
 use IPC::Run;
 
 sub set_ssh_console_timeout_before_use {
     reset_consoles;
     select_console('root-console');
     set_ssh_console_timeout('/etc/ssh/sshd_config', '28800');
     reset_consoles;
-    select_console 'sol', await_console => 1;
+    select_console 'sol', await_console => 0;
```
alice-suse, line 26: What are the reasons for/benefit from disabling await_console?

waynechen55:
I did several verification runs for guest upgrade. I think the result is convincing.

Although [Test Run 1](https://openqa.suse.de/tests/10807287) failed, but it passed wait_serial step judging by serial0.txt and autoinst-log.txt: ```[2023-03-28T13:07:07.945335+02:00] [debug] [pid:338304] >>> testapi::wait_serial: CMD_FINISHED-519058: ok```
Also [Test Run 2](https://openqa.suse.de/tests/10807288): ```[2023-03-28T13:07:07.954561+02:00] [debug] [pid:339531] >>> testapi::wait_serial: CMD_FINISHED-469118: ok```.

tests/virt_autotest/login_console.pm
waynechen55: ARM64 does not have KVM module because KVM module is eithe intel_kvm or amd_kvm I think.

Julie-CAO: Ok, I'll skip arm and s390 machines.

tests/virt_autotest/login_console.pm
Julie-CAO: because I don't find there is any reason we have to wait for sol console on x86 at the moment. Sol console is not mandatory. @waynechen55 correct me if this is not correct.

alice-suse: This await_console ensures that the console connection is setup successfully before it returns, and it actually is quick , only backend tries to build the connection with 'ipmitool xx sol'. Without this , it is possible that when `send_key 'ret'` in the next line is executed, the console is not ready yet.

alice-suse: I just had a check in code. This function set_ssh_console_timeout_before_use is only used for arm case in login_console. Given that it has been running with await_console =>1 without problem, the sol needle should be there already. With the change in this PR to use await_console =>0, it will just loosen the condition for arm, no impact to x86 I think.

Anyway, it is not big deal. I suppose it is not the core of the this PR either. I do not insist on any option, just curious about the motivation/win for this change. I am willing to merge either of the final decision on this if there is benefit.

waynechen55: From my verification run, I did not found a similar screen that can match previously 'sol' needle for arm. I did not find a suitable area suitable for 'sol' needle either on x86_64. Test run will assert 'linux-login' on sol which serves the same purpose, I think it would be better to do not use await_console with x86 otherwise it will introduce unexpected failures during test run, especially for login_console.pm.

waynechen55:
In order increase certainty, I think you can add one more checking in set_ssh_console_time in lib/Utils/Backend.pm

tests/virt_autotest/login_console.pm
alice-suse: Agree.

Julie-CAO:
> In order to increase certainty, I think you can add one more checking in `set_ssh_console_time` in `lib/Utils/Backend.pm` like this:
> 
> ```
> if (script_run("ls $sshd_config_fle") == 0) {
>     seting keepalive cofig
> }
> else {
>     record_info
> }
> ```
> 
> @Julie-CAO

I agree.

Julie-CAO:
> In order to increase certainty, I think you can add one more checking in `set_ssh_console_time` in `lib/Utils/Backend.pm` like this:
> 
> ```
> if (script_run("ls $sshd_config_fle") == 0) {
>     seting keepalive cofig
> }
> else {
>     record_info
> }
> ```
> 
> @Julie-CAO

I agree.

Julie-CAO:
Add record_info and two more verification runs:
[gi-guest_developing-on-host_developing-kvm](https://openqa.suse.de/tests/10844485)
[uefi-gi-guest_sles12sp5-on-host_developing-xen](https://openqa.suse.de/tests/10844516#step/login_console/16)

PR is ready to merge.

PR #16700, opened 2023-03-24
ALP: move kernel tests after image boot

jlausuch:
This way, the image boot will be done in the main scheduler and the rest in the kernel scheduler.

failure: https://openqa.opensuse.org/tests/3189603#step/install_ltp/128
VR: https://openqa.opensuse.org/tests/3189672


czerw (approved):
Thank you.

mloviska (approved):

PR #16704, opened 2023-03-27
Use local file to record orphaned packages

rfan1:
https://progress.opensuse.org/issues/126578
The `script_output` functioin directs stdout (not' stderr!) to the serial console, If some un-expeted messages show up along with our test run, the test may fail sporadically due to handle all messages at the same time. so enhance the current logic to use a local file.

- Verification run: https://openqa.opensuse.org/tests/3193564


tests/console/orphaned_packages_check.pm
```diff
@@ -67,14 +63,18 @@ sub run {
...
-         grep -v "^Warning" | grep -v "\(release-DVD\|release-dvd\|openSUSE-release\|skelcd\)" |
-         awk -F \| 'NR>2 {print $3}'], proceed_on_failure => 1, timeout => 180);
+
+    script_run(
+q[zypper --quiet packages --orphaned | grep -v "^Warning" | grep -v "\(release-DVD\|release-dvd\|openSUSE-release\|skelcd\)" | awk -F '\\| ' 'NR>2 {print $3}' > /tmp/orphaned.log],
+        timeout => 180,
+        die_on_timeout => 0
+    );
+    upload_logs('/tmp/orphaned.log', (log_name => 'orphaned.log'));
+    my @orphans = split(/\n/, path('ulogs/orphaned.log')->slurp);
```
dzedro, line 73: Why not just `/tmp/orphaned.log` ?

rfan1: Good question! I believed that we can format the file within the VM at the first! but I was wrong.
the file operation is done in kvm host rather than VM.

rfan1:
> You could fix also https://openqa.opensuse.org/tests/3193564#step/orphaned_packages_check/11

Thanks much!  it needs to be fixed indeed, let me do

rfan1:
> You could fix also https://openqa.opensuse.org/tests/3193564#step/orphaned_packages_check/11



> > You could fix also https://openqa.opensuse.org/tests/3193564#step/orphaned_packages_check/11
> 
> Thanks much! it needs to be fixed indeed, let me do

https://openqa.opensuse.org/tests/3195342#step/orphaned_packages_check/13

rfan1:
[VR on TW](https://openqa.opensuse.org/tests/overview?distri=opensuse&version=Tumbleweed&build=rfan0328)

Vogtinator:
> FWICT `awk -F'\\\\| '` is even more broken than the previous `-F'\|'`.

Fixed with https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16731.

rfan1:
> FWICT `awk -F'\\\\| '` is even more broken than the previous `-F'\|'`.
> 
> Also, why not just use `select_serial_terminal` instead of `root-console`? Then `script_output` works as expected.
> 
> IMO `script_output` needs to work everywhere. If there is a case where it does not, that needs to be fixed instead of working around it.

Thanks @Vogtinator for the kindly help!

You are correct, `select_serial_terminal`  will not see the issue, since the output is directing to "serial_termial" rather than "serial0".

The reason why I don't like to use `select_serial_terminal` in our tests is https://progress.opensuse.org/issues/109929
We used to failed to rollback to a snapshot after a failed case!

Vogtinator:
> The reason why I don't like to use select_serial_terminal in our tests is https://progress.opensuse.org/issues/109929
We used to failed to rollback to a snapshot after a failed case!

Yes, but many (most?) other tests already use it, so it's something we'll have to deal with anyway.

rfan1:
I will file a new PR to use `serial_terminal`

PR #16707, opened 2023-03-27
Ensure to select bond port yast2_lan_restart_bond

skliu:
Ensure to select bond port yast2_lan_restart_bond

**Ticket:**

- https://progress.opensuse.org/issues/125999

**Verification run:**

- https://openqa.suse.de/t10825294 
- https://openqa.suse.de/t10825266
- https://openqa.suse.de/t10825278
- https://openqa.suse.de/t10825284
- https://openqa.suse.de/t10825286
- https://openqa.suse.de/t10825292
- https://openqa.suse.de/t10825296
- https://openqa.suse.de/t10825299
- https://openqa.suse.de/t10825301
- https://openqa.suse.de/t10825303
- https://openqa.suse.de/t10825340

### DIFF
--- a/lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
+++ b/lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
@@ -39,6 +39,7 @@ sub select_bond_slave_in_list {
     assert_screen(BOND_SLAVES_TAB);
     apply_workaround_poo124652(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED) if (is_sle('>=15-SP4'));
     assert_and_click(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED);
+    wait_still_screen;
 }
 
 sub select_continue_in_popup {


lib/YaST/NetworkSettings/NetworkCardSetup/BondSlavesTab.pm
```diff
@@ -38,7 +39,10 @@ sub select_tab {
 sub select_bond_slave_in_list {
     assert_screen(BOND_SLAVES_TAB);
     apply_workaround_poo124652(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED) if (is_sle('>=15-SP4'));
+
     assert_and_click(BOND_SLAVE_DEVICE_CHECKBOX_UNCHECKED);
+
+    apply_workaround_poo124652(BOND_SLAVE_DEVICE_CHECKBOX_CHECKED) if (is_sle('>=15-SP4'));
```
jknphy, line 45: have you tried manually that it is really due to refreshment glitch that the checkbox is not tick?
I think the glitch has only been seen in large area of the screen, not just small square of the checkbox, very unlikely...

skliu: No, I haven't. I agree with you that the glitch has only been seen in large area of the screen. I will check it right now. BTW: after I add my code, there is no failure any more.

jknphy: your VR should show always 5-10 run for this kind of sporadic failure, I'm afraid that introducing the workaround here where probably is not needed it is just adding waiting time, that we could could just add just in the workaround itself or in the failing test code with wait_still_screen.

skliu: I am runing the fifth time.

skliu: I used `wait_still_screen` and ran it for 10 times, only https://openqa.suse.de/tests/10825294 failed, but I think it's another problem, please check the `Next & previous results` tab.

jknphy: we are out for 1 seconds or so when opening the module and waiting for the needle in that failure I think, you could bump timeout 30 seconds there, it happens sometimes when there is a initialization screen before the module opens.

jknphy: Please, remove commented code and those constants not used to have a clear view of the point where we are.

skliu: > this is the best way to show clear VR when you need repetition:
> 
> ```
>  for i in {1..5}; do openqa-clone-custom-git-refspec http://github.com/<user>/os-autoinst-distri-opensuse/tree/<your-branch> https://openqa.suse.de/tests/<job-id> TEST=<original_name>_$i  _SKIP_POST_FAIL_HOOKS= ; done
> ```
> 
> I cannot see what you mentioned about 9/10 passed. Could you please paste them those 10 or proceed with advice above? (in that case you just need to show one link).

I pasted all the links, the first one is failed.

jknphy: thanks, seems that helped, merging.
Could you please restart the job in YaST group and label it with other failures found there.
Looks like yast2 security sporadically fails, but that can be done in another ticket for sure.

PR #16711, opened 2023-03-27
Remove workaround of 'cannot load initrd' from pvm installation

Zaoliang:
we have issue 'cannot load initrd' since quite long time. The real root cause is on qanet which has problem with exported directory mounts sometimes.

So remove previous workaround for trying to reset at GRUB so that we can save a lot of time for assert_screen and speed up the whole installation progress.

see https://progress.opensuse.org/issues/120570


Zaoliang:
VR:
https://openqa.suse.de/tests/10807336

PR #16726, opened 2023-03-28
alp: Fix schedule of kernel tests

czerw:
Basically reverts https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16700 and adds more configuration modules.

- Fail: https://openqa.opensuse.org/tests/3195539#
- Needles: none
- Verification runs:
 1. ALP Micro:
  installation: https://openqa.opensuse.org/tests/3196124
  run: https://openqa.opensuse.org/tests/3196128
 2. ALP Bedrock:
  installation: https://openqa.opensuse.org/tests/3195953
  run: https://openqa.opensuse.org/tests/3195946

### DIFF
--- a/lib/main_ltp.pm
+++ b/lib/main_ltp.pm
@@ -8,7 +8,7 @@ use base 'Exporter';
 use Exporter;
 use testapi qw(check_var get_required_var get_var);
 use utils;
-use main_common qw(boot_hdd_image load_bootloader_s390x load_kernel_baremetal_tests);
+use main_common qw(boot_hdd_image load_bootloader_s390x load_kernel_baremetal_tests replace_opensuse_repos_tests is_repo_replacement_required);
 use 5.018;
 use Utils::Backends;
 use version_utils qw(is_opensuse is_alp);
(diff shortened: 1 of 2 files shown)


lib/main_ltp.pm
```diff
@@ -32,6 +32,14 @@ sub load_kernel_tests {
     loadtest_kernel "../installation/bootloader" if is_pvm;
 
     if (get_var('INSTALL_LTP')) {
+        if (is_alp) {
+            loadtest 'microos/disk_boot';
+            loadtest 'update/zypper_clear_repos';
```
jlausuch, line 37: for these 3 tests, you can use: `replace_opensuse_repos_tests if is_repo_replacement_required;` (also need to import it from main_common)

czerw:
> Nice! Just out of curiosity, where is the actual schedule of the children jobs which only do `boot_ltp` ?

`main_ltp` loads `boot_ltp` which does dynamic scheduling using function `schedule_tests`.

lib/main_ltp.pm
czerw: Not sure, if i understand. Do you mean to add that line after zypper actions?

mloviska: `replace_opensuse_repos_tests` schedules all the modules that handle repo replacement according to test suite variables.

czerw: Thanks, updated.

PR #16731, opened 2023-03-28
console/orphaned_packages_check: Fix zypper output parsing

Vogtinator:
Previously it also tried to parse the table header, interpreting "Name" as package name. Just wait for the table to start, this also skips all warnings as a side effect.

Also don't ignore timeouts, that just leads to false negatives.

- Verification run: https://openqa.opensuse.org/tests/3196495


dzedro (approved):
I didin't see the fix for '\\|'

PR #16752, opened 2023-03-30
Configure scipy module run properly with MPI modules

b10n1k:
As for now the scipy test was configured incorrectly and was running with wrong source code. First we need to set HPC_LIB. Then the rest is on this PR.

- Related ticket: https://progress.opensuse.org/issues/126779
- Verification run: https://aquarius.suse.cz/tests/15687

### DIFF
--- a/lib/hpc/utils.pm
+++ b/lib/hpc/utils.pm
@@ -93,13 +93,17 @@ sub setup_scientific_module {
     my $mpi = get_required_var('MPI');
 
     if (get_var('HPC_LIB') eq 'scipy') {
-        zypper_call("in python3-scipy-gnu-hpc");
-        assert_script_run("env MPICC=mpicc python3 -m pip install mpi4py");
+        # $mpi-gnu-hpc-devel will be installed on compute nodes
+        # which is required to compile the mpi4py.
+        # This seems is not be shared by NFS
+        # If you have a better idea let me know.
+        zypper_call("in python3-scipy-gnu-hpc python3-devel $mpi-gnu-hpc-devel");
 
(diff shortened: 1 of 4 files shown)


lib/hpc/utils.pm
```diff
@@ -93,13 +93,17 @@ sub setup_scientific_module {
     my $mpi = get_required_var('MPI');
 
     if (get_var('HPC_LIB') eq 'scipy') {
-        zypper_call("in python3-scipy-gnu-hpc");
-        assert_script_run("env MPICC=mpicc python3 -m pip install mpi4py");
+        # $mpi-gnu-hpc-devel will be installed on compute nodes
+        # which is required to compile the mpi4py.
+        # This seems is not be shared by NFS
```
Avinesh, line 98: I don't understand what this statement means.

b10n1k: Controller and compute nodes should have the *-gnu-hpc packages installed. As per design it s supposed the controller shares `/usr/lib/hpc` via NFS, in order to avoid re-installing HPC packages on compute nodes. But compiling mpi4py, i found that i needed $mpi-gnu-hpc-devel. i would expect that those packages are found in `/usr/lib/hpc` but likely this is not the case.

Avinesh: so don't we test the NFS sharing scenario for HPC clusters?

czerw: NFS will be covered in https://progress.opensuse.org/issues/126698

Avinesh: okay, Thanks @czerw !

Avinesh (approved):
Approving as NFS scenario is being taken care in another ticket.

czerw (approved):
Thank you!

PR #16753, opened 2023-03-30
Increase timeout on spack spec command

b10n1k:
This is intent to fix the timeout on aarch64 which getting slower and slower

- Related ticket: https://progress.opensuse.org/issues/126806
- Verification run: https://openqa.suse.de/tests/10827843


b10n1k:
solved in another PR

PR #16754, opened 2023-03-31
Y2K38 Epochalypse

rfan1:
Smoke test to detect year 2038 issue

https://en.wikipedia.org/wiki/Year_2038_problem
https://progress.opensuse.org/issues/125639
https://progress.opensuse.org/issues/124736

- Verification run: 
[SLE15SP5](https://openqa.suse.de/tests/10808235#step/year_2038_detection/45)
[ALP](https://openqa.opensuse.org/tests/3197649#step/year_2038_detection/68)

**I didn't add the test to QEM, since current releases should be EOS in 2038 IMO**

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -330,6 +330,7 @@ sub load_tests {
         load_common_tests;
         load_transactional_tests unless is_zvm;
     }
+    loadtest 'console/year_2038_detection';
     load_journal_check_tests;
 }
 
--- a/tests/console/year_2038_detection.pm
+++ b/tests/console/year_2038_detection.pm
@@ -1,0 +1,80 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Monitor Year 2038 problem, but so far no fix is required due to
+# https://bugzilla.suse.com/show_bug.cgi?id=1188626
+#
+# The Year 2038 (Y2038) problem relates to representing time in many digital systems
+# as the number of seconds passed since 00:00:00 UTC on 1 January 1970 and storing it
+# as a signed 32-bit integer. Such implementations cannot encode times after
+# 03:14:07 UTC on 19 January 2038. Similar to the Y2K problem, the Year 2038 problem
+# is caused by insufficient capacity used to represent time.
+#
+# Maintainer: QE Core <qe-core@suse.de>
+
+use base 'consoletest';
+use strict;
+use warnings;
+use testapi;
+use utils;
+use serial_terminal 'select_serial_terminal';
+use transactional qw(trup_call check_reboot_changes);
+use version_utils qw(is_transactional);
+
+sub install_pkg {
+    if (is_transactional) {
+        trup_call('pkg install chrony');
+        check_reboot_changes;
+    }
+    else {
+        zypper_call('in chrony');
+    }
+}
+
+sub run {
+    select_serial_terminal;
+
+    install_pkg if (script_run('rpm -qi chrony') != 0);
+    assert_script_run 'chronyc makestep';
+    record_info('Show current date and time', script_output('date +"%Y-%m-%d"'));
+    assert_script_run('utmpdump /var/run/utmp');
+    assert_script_run('utmpdump /var/log/wtmp');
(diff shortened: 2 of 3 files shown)


tests/console/year_2038_detection.pm
```diff
@@ -0,0 +1,75 @@
...
+        assert_script_run "useradd -m $testapi::username";
+        assert_script_run "echo '$testapi::username:$testapi::password' | chpasswd";
+    }
+    ensure_serialdev_permissions;
+
+    # Switch user to check if the issue can be captured
+    select_console 'user-console';
+    my $utmp_output = script_output('utmpdump /var/run/utmp');
+    my $wtmp_output = script_output('utmpdump /var/log/wtmp');
+    record_soft_failure('bsc#1188626 uttmpdump shows incorrect year for 2038 and beyond') if ($utmp_output =~ m/1901/sx || $wtmp_output =~ m/1901/sx);
```
grisu48, line 61: Instead of checking only for `1901` I think it's better to check for the date NOT being 2038

```suggestion
    record_soft_failure('bsc#1188626 uttmpdump shows incorrect year for 2038 and beyond') if ($utmp_output !~ m/2038/ || $wtmp_output !~ m/2038/);
```

tests/console/year_2038_detection.pm
```diff
@@ -0,0 +1,75 @@
...
+    else {
+        zypper_call('in chrony');
+    }
+}
+
+sub run {
+    select_serial_terminal;
+
+    install_pkg if (script_run('rpm -qi chrony') != 0);
+    systemctl('start chronyd.service');
```
grisu48, line 40: Wouldn't a `chronyc makestep` be better here to ensure the systemtime is correct, regardless of how off it is? (Looking at RPi workers and others)

tests/console/year_2038_detection.pm
```diff
@@ -0,0 +1,75 @@
...
+    systemctl('stop chronyd.service');
+    # Set the time and date beyond a Y2038
+    assert_script_run('timedatectl set-time "2038-01-20 03:14:07"');
+    # Create user account, if image doesn't already contain user
+    # (which is the case for SLE images that were already prepared by openQA)
+    if (script_run("getent passwd $username") != 0) {
+        assert_script_run "useradd -m $testapi::username";
+        assert_script_run "echo '$testapi::username:$testapi::password' | chpasswd";
+    }
+    ensure_serialdev_permissions;
```
grisu48, line 55: Is this needed as part of the test run? :thinking:

lib/main_micro_alp.pm
```diff
@@ -330,6 +330,7 @@ sub load_tests {
         load_common_tests;
         load_transactional_tests unless is_zvm;
     }
+    loadtest 'console/year_2038_detection' if is_alp;
```
foursixnine, line 333:
```suggestion
    loadtest 'console/year_2038_detection';
```

tests/console/year_2038_detection.pm
rfan1: Good question, based on the test requirement, we need to logout and re-login after changing the date/time.
However, I failed to find any existing function/method to achieve it.

So, I decided to switch user to login the system again.
Please share me some light if you have a better solution.

rfan1: @grisu48

tests/console/year_2038_detection.pm
rfan1: You are correct! ：）

foursixnine (approved):

tests/console/year_2038_detection.pm
grisu48: That's fine but please add a comment why this is necessary

rfan1: Sure, will do

grisu48 (approved):
LGTM!

DimStar77:
Failure detected on openSUSE MicroOS:
  https://openqa.opensuse.org/tests/3208639#step/year_2038_detection/58

DimStar77:
> Failure detected on openSUSE MicroOS: https://openqa.opensuse.org/tests/3208639#step/year_2038_detection/58

The permission denied while writing on dev/ttyS0 should probably be taken serious

PR #16759, opened 2023-03-31
Ensure that version of podman is upper than 4.4.2 before testing kube

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/124601
VR:
- 15-SP4 http://openqa.suse.de/tests/10832660
- 15-SP3 http://openqa.suse.de/tests/10832455
- 15-SP2 http://openqa.suse.de/tests/10832662
- 15-SP1 http://openqa.suse.de/tests/10832661

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -22,7 +22,7 @@ use Mojo::Util 'trim';
 
 our @EXPORT = qw(test_seccomp runtime_smoke_tests basic_container_tests get_vars
   can_build_sle_base get_docker_version get_podman_version check_runtime_version
-  container_ip container_route registry_url);
+  check_min_runtime_version container_ip container_route registry_url);
 
 sub test_seccomp {
     my $no_seccomp = script_run('docker info | tee /tmp/docker_info.txt | grep seccomp');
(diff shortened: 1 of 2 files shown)


lib/containers/utils.pm
```diff
@@ -56,6 +56,13 @@ sub check_runtime_version {
     return check_version($other, $current, qr/\d{2}(?:\.\d+)/);
 }
 
+sub check_min_runtime_version() {
```
pablo-herranz, line 59: I'll choose a function name that explicitly said `podman`. Otherwise, LGTM

pablo-herranz (approved):

lib/containers/utils.pm
ilausuch: its on utils, Check the previous function, I like the idea to have a checker for versions no mater what product

pdostal (approved):

tests/containers/podman_pods.pm
```diff
@@ -42,7 +43,7 @@ sub run {
         assert_script_run('podman play kube --down hello-kubic.yaml');
     }
 
-    unless (is_sle('<=15-SP2')) {
+    unless (is_sle('<=15-SP2') || !check_min_runtime_version('4.4.2')) {
```
jlausuch, line 46: maybe we can remove the check `is_sle('<=15-SP2')` completely, so we only depend on the podman version?

ilausuch: ok

jlausuch (approved):

PR #16770, opened 2023-04-03
Make sure we save image md5 as Azure blob metadata

pdostal:
This idea has two stages:
1) Upload image md5 as blob metadata after blob is uploaded
2) Check that the md5 metadata field is available

The goal is to not upload one blob two times at the same time.

- Author: @mpagot
- Verification run: [non-existing image](https://pdostal-server.suse.cz/tests/4244), [existing image](https://pdostal-server.suse.cz/tests/4245)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -141,20 +141,35 @@ sub upload_img {
     my $key = $self->get_storage_account_keys($storage_account);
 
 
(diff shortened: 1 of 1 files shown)


asmorodskyi (approved):
LGTM

mpagot (approved):
LGTM

asmorodskyi:
fix tidy and it can be merged

PR #16771, opened 2023-04-03
Add apache2 version setting to SLES12-SP5

grisu48:
Introduce the `APACHE2_PKG` setting to select `apache` or `apache2-tls13`, as we need to test both on SLES12-SP5.

- Related ticket: https://progress.opensuse.org/issues/126146
- Verification run: [12-SP5 with apache2](https://duck-norris.qe.suse.de/tests/12480) | [12-SP5 with apache2-tls13](https://duck-norris.qe.suse.de/tests/12481) | [15-SP4](https://duck-norris.qe.suse.de/tests/12482) | [Tumbleweed](https://duck-norris.qe.suse.de/tests/12483)

### DIFF
--- a/tests/console/apache.pm
+++ b/tests/console/apache.pm
@@ -1,6 +1,6 @@
 # SUSE's Apache regression test
 #
-# Copyright 2019-2021 SUSE LLC
+# Copyright 2019-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: apache2 apache2-mod_php72 php72-php72-curl
@@ -22,6 +22,9 @@ use version_utils qw(is_sle is_jeos);
 sub run {
     select_serial_terminal;
 
+    # On SLES12-SP5 apache2 and apache2-tls13 are supported
+    my $apache2 = get_var('APACHE2_PKG', "apache2");
+
     # installation of docs and manpages is excluded in zypp.conf
     # enable full package installation, and clean up previous apache2 deployment
     if (is_jeos) {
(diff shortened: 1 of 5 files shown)


dzedro:
I assume and hope this setup is only on 12-SP5.
APACHE2_PKG is IMO unhandy and confusing, either test install apache2-tls13 or apache2 or both packages and does test them. AFAICS apache test can test both versions and other tests can do the same or just continue with apache2.
With APACHE2_PKG, we will have two instances of apache test ?

rfan1:
If my understanding is correct, the current logic can only test one `apache2` version，not both, correct?

If we need to test both `apache2` and `apache2-tls13`, we may have to create a new job, or create a new test module.

grisu48:
> If my understanding is correct, the current logic can only test one `apache2` version，not both, correct?
> 
> If we need to test both `apache2` and `apache2-tls13`, we may have to create a new job, or create a new test module.

Correct, I will add a second job after this PR. Introducting the `APACHE2_PKG` is the necessary requirement for doing so.

tests/console/apache.pm
```diff
@@ -33,13 +36,24 @@ sub run {
     assert_script_run 'echo "index" > /srv/www/htdocs/index.html';
 
     # Ensure apache2 is installed and stopped
-    if (script_run('rpm -q apache2') == 0) {
-        systemctl('stop apache2');
+    if (is_sle("=12-SP5")) {
+        if ($apache2 eq 'apache2-tls13') {
+            zypper_call 'rm apache2';
```
rfan1, line 41: Why do we need to remove the package here?
It has been installed at https://duck-norris.qe.suse.de/tests/12473#

rfan1: Seems remove and install the same packages :

https://duck-norris.qe.suse.de/tests/12473#step/apache/6
https://duck-norris.qe.suse.de/tests/12473#step/apache/8

grisu48: Strictly speaking it's not necessary, but this is to ensure that the correct package is installed and I think it's good to leave it there for this reason.

rfan1: Fine, one more question, do we need to remove `apache2*-utils` at the same time?

grisu48: Perhaps, but I think we're better off in not doing the `zypper rm` in the first place. Let me try to tidy this block a bit ... Will poke you when done.

grisu48: I removed the whole block, now we just install the corresponding apache package. This would cause a package conflict in case a conflicting apache2 installation is already present, so we should be safe.

Let's add the safeguard only if necessary to keep the codebase tidy.

rfan1 (approved):
LGTM

PR #16773, opened 2023-04-04
Replace GUI host installation with autoyast for sle15

tonyyuan1:
Create a autoyast template for sles15spx in order for product virtuallzation test to switch from GUI installation to autoyast

https://progress.opensuse.org/issues/112508

- Related ticket: https://progress.opensuse.org/issues/112508
- Needles: no
- Verification run: 
- sle15sp5 xen AMD: http://10.67.183.130/tests/3493
- sle15sp5 kvm: http://10.67.183.130/tests/3492
- sle15sp4 kvm: http://10.67.183.130/tests/3505#
 - gi-guest_developing-on-host_developing-kvm: https://openqa.suse.de/tests/11130833
 - uefi-gi-guest_developing-on-host_developing-xen: https://openqa.suse.de/tests/11132392 
- gi-guest_sles15sp4-on-host_developing-kvm: https://openqa.suse.de/tests/11132406
- prj2_host_upgrade_sles15sp4_to_developing_xen: https://openqa.suse.de/tests/11132409
- prj4_guest_upgrade_sles15sp4_on_sles15sp4-xen: https://openqa.suse.de/tests/11136940
- uefi-gi-guest_developing-on-host_sles15sp4: https://openqa.suse.de/tests/11137310#
- virt-guest-migration-developing-from-developing-to-developing-kvm-src:
 - https://openqa.suse.de/tests/11137008
- https://openqa.suse.de/tests/11137009
- sev-es-gi-guest_developing-on-host_developing-kvm: https://openqa.suse.de/tests/11146649
- virt-guest-migration-sles15sp4-from-sles15sp4-to-developing-xen-src: https://openqa.suse.de/tests/11154029

### DIFF
--- a/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
@@ -1,0 +1,268 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <suse_register>
+    <do_registration config:type="boolean">true</do_registration>
+    <reg_server>{{SCC_URL}}</reg_server>
+    <reg_code>{{SCC_REGCODE}}</reg_code>
+    <install_updates config:type="boolean">true</install_updates>
+    % if (keys %$addons) {
+    <addons config:type="list">
+      % while (my ($key, $addon) = each (%$addons)) {
+      <addon>
+        <name><%= $addon->{name} %></name>
+        <version><%= $addon->{version} %></version>
+        <arch><%= $addon->{arch} %></arch>
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sles')) {
+        <reg_code><%= $get_var->('SCC_REGCODE_WE') %></reg_code>
+        % }
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        % }
+        % if ($key eq 'rt') {
+        <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+        % }
+        % if ($key eq 'ltss') {
+        <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
+        % }
+        <release_type/>
+      </addon>
+      % }
+    </addons>
+    %}
+  </suse_register>
+  <!--add-on>
+    <add_on_products config:type="list">
+      % my $n =0;
+      % for my $repo (@$repos) {
+      <listentry>
+        <media_url><%= $repo %></media_url>
+        <alias>TEST_<%= $n++ %></alias>
+      </listentry>
+      % }
+    </add_on_products>
+  </add-on-->
+  <bootloader>
+    <global>
+      <activate>true</activate>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --speed=115200 --unit=<%= $get_var->('SERIALDEV') =~ s/ttyS//r %> --word=8 --parity=no --stop=1</serial>
+      <timeout config:type="integer">15</timeout>
+      <trusted_grub>false</trusted_grub>
+      <terminal>console</terminal>
+      % if ($check_var->('SYSTEM_ROLE', 'xen')) {
+      <xen_append>splash=silent quiet console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %> <%= defined $bmwqemu::vars{"OPT_KERNEL_PARAMS"} ? $bmwqemu::vars{"OPT_KERNEL_PARAMS"} : "" %></xen_append>
+      <xen_kernel_append>console=com2,115200 vga=gfx-1024x768x16 loglvl=all guest_loglvl=all sync_console</xen_kernel_append>
+      % } else {
+      <append>splash=silent console=<%= $get_var->('SERIALDEV') %>,115200 console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %> <%= defined $bmwqemu::vars{"OPT_KERNEL_PARAMS"} ? $bmwqemu::vars{"OPT_KERNEL_PARAMS"} : "" %></append>
+      % }
+    </global>
+    <loader_type>grub2</loader_type>
+  </bootloader>
+  <general>
+    <ask-list config:type="list"/>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+      <final_reboot config:type="boolean">true</final_reboot>
+    </mode>
+    <proposals config:type="list"/>
+    <signature-handling>
+      <accept_file_without_checksum config:type="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
+      <accept_unknown_gpg_key config:type="boolean">true</accept_unknown_gpg_key>
+      <accept_unsigned_file config:type="boolean">true</accept_unsigned_file>
+      <accept_verification_failed config:type="boolean">true</accept_verification_failed>
+      <import_gpg_key config:type="boolean">true</import_gpg_key>
+    </signature-handling>
+    <storage>
+      <start_multipath config:type="boolean">false</start_multipath>
+    </storage>
+  </general>
+  <kdump>
+    <add_crash_kernel config:type="boolean">true</add_crash_kernel>
+    <crash_kernel>201M,high</crash_kernel>
+    <crash_xen_kernel>201M\&lt;4G</crash_xen_kernel>
+  </kdump>
+  <keyboard>
+    <keymap>english-us</keymap>
+  </keyboard>
+  <language>
+    <language>en_US</language>
+    <languages/>
+  </language>
+  <login_settings/>
+  <networking>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+    <backend>wicked</backend>
+    <dns>
+      <dhcp_hostname config:type="boolean">true</dhcp_hostname>
+      <nameservers config:type="list">
+        <nameserver>10.162.0.1</nameserver>
+<!--gonzo and a another machine can not get nameserver from local dns server after autoyast installation. Put a static nameserver option above. If nameserver is retrived from dhcp they will be merged. Otherwise use the static -->
+      </nameservers>
+      <resolv_conf_policy>auto</resolv_conf_policy>
+    </dns>
+    <virt_bridge_proposal config:type="boolean">true</virt_bridge_proposal>
+  </networking>
+  <ntp-client>
+    <ntp_policy>auto</ntp_policy>
+    <ntp_servers config:type="list">
+      <ntp_server>
+        <address>europe.pool.ntp.org</address>
+        <iburst config:type="boolean">true</iburst>
+        <offline config:type="boolean">false</offline>
+      </ntp_server>
+    </ntp_servers>
+    <ntp_sync>systemd</ntp_sync>
+  </ntp-client>
+  <partitioning config:type="list">
+  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745' 'openqaipmi5' => 'wwn-0x5000c5008711f2fc'};
+  % my $hostname = (split(/\./, $get_var->("SUT_IP")))[0];
+  % my $device_id = defined($wwn->{$hostname}) ? '/dev/disk/by-id/' . $wwn->{$hostname} : ''; 
+    <drive>
+      <device><%= $device_id %></device>
+      <disklabel>gpt</disklabel>
+      <enable_snapshots config:type="boolean">false</enable_snapshots>
+      <initialize config:type="boolean">true</initialize>
+      <partitions config:type="list">
+        <partition>
+          <create config:type="boolean">true</create>
+          <create_subvolumes config:type="boolean">true</create_subvolumes>
+          <filesystem config:type="symbol">btrfs</filesystem>
+          <format config:type="boolean">true</format>
+          <mount>/</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">131</partition_id>
+        </partition>
+        <partition>
+          <create config:type="boolean">true</create>
+          <filesystem config:type="symbol">xfs</filesystem>
+          <format config:type="boolean">true</format>
+          <mount>/var/lib/libvirt/images/</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">131</partition_id>
+          <resize config:type="boolean">false</resize>
+          <size>120G</size>
+        </partition>
+      </partitions>
+      <type config:type="symbol">CT_DISK</type>
+      <use>all</use>
+    </drive>
+  </partitioning>
+  <report>
+    <errors>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">0</timeout>
+    </errors>
+    <messages>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </messages>
+    <warnings>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </warnings>
+    <yesno_messages>
+      <log config:type="boolean">true</log>
+      <show config:type="boolean">true</show>
+      <timeout config:type="integer">10</timeout>
+    </yesno_messages>
+  </report>
+  <services-manager>
+    <default_target>multi-user</default_target>
+    <services>
+      <disable config:type="list">
+        <service>firewalld</service>
+      </disable>
+      <enable config:type="list">
+        <service>sshd</service>
+      </enable>
+    </services>
+  </services-manager>
+  <software>
+    <install_recommended config:type="boolean">true</install_recommended>
+    <instsource/>
+    <packages config:type="list">
+      <package>dhcp-client</package>
+      <package>guestfs-tools</package>
+      <package>nmap</package>
+    </packages>
+    <products config:type="list">
+      <product><%= uc $get_var->('SLE_PRODUCT') %></product>
+    </products>
+    <patterns config:type="list">
+      % my $patterns = $check_var->('SYSTEM_ROLE', 'kvm') ? ['base', 'enhanced_base', 'kvm_server', 'kvm_tools'] : ['base', 'enhanced_base', 'xen_server', 'xen_tools'];
+      % for my $pattern (@$patterns) {
+      <pattern><%= $pattern %></pattern>
+      % }
+    </patterns>
+  </software>
+  <ssh_import>
+    <copy_config config:type="boolean">false</copy_config>
+    <import config:type="boolean">false</import>
+  </ssh_import>
+  <timezone>
+    <hwclock>UTC</hwclock>
+    <timezone>UTC</timezone>
+  </timezone>
+  <users config:type="list">
+    <user>
+      <authorized_keys config:type="list"/>
+      <encrypted config:type="boolean">true</encrypted>
+      <fullname>bernhard</fullname>
+      <gid>100</gid>
+      <home>/home/bernhard</home>
+      <password_settings>
+        <expire/>
+        <flag/>
+        <inact/>
+        <max>99999</max>
+        <min>0</min>
+        <warn>7</warn>
+      </password_settings>
+      <shell>/bin/bash</shell>
+      <uid>1000</uid>
+      <user_password>$6$QSJG7nUenDgf$UcRX/Q6YWM2/S55RJ8Mgvu93oKEDlOs.7/rOTMYjjxXQPiZ/rq4ogx5EpkksVRZXoEF0kX8FeEa.gOTxl1l/t0</user_password>
+      <username>bernhard</username>
+    </user>
+    <user>
+      <authorized_keys config:type="list"/>
+      <encrypted config:type="boolean">true</encrypted>
+      <fullname>root</fullname>
+      <gid>0</gid>
+      <home>/root</home>
+      <password_settings>
+        <expire/>
+        <flag/>
+        <inact/>
+        <max/>
+        <min/>
+        <warn/>
+      </password_settings>
+      <shell>/bin/bash</shell>
+      <uid>0</uid>
+      <user_password>$6$0bUrc6YvA/qw$h1Z3pzadaxmc/KgcHRSEcYoU1ShVNymoXBaRAQZJ4ozVhTbCvdAMbGQrQAAX7cC9cLRybhsvDio3kBX/IB3xj/</user_password>
+      <username>root</username>
+    </user>
+  </users>
+  <scripts>
+    <post-scripts config:type="list">
+      % if ($check_var->('SYSTEM_ROLE', 'xen') && $check_var->('XEN_DEFAULT_BOOT_IS_SET', 1)) {
+      <script>
+        <filename>default_xen_boot.sh</filename>
+        <interpreter>/bin/bash -x</interpreter>
+        <source><![CDATA[
+grub2-set-default 2
+]]>
+        </source>
+      </script>
+      % }
+      <script/>
+    </post-scripts>
+  </scripts>
+</profile>
--- a/products/sle/main.pm
+++ b/products/sle/main.pm
@@ -849,19 +849,16 @@ elsif (get_var("VIRT_AUTOTEST")) {
     }
     else {
         if (!is_s390x) {
+            loadtest "autoyast/prepare_profile" if get_var("AUTOYAST_PREPARE_PROFILE");
             load_boot_tests();
             if (get_var("AUTOYAST")) {
                 loadtest "autoyast/installation";
-                loadtest "virt_autotest/reboot_and_wait_up_normal";
             }
             else {
                 load_inst_tests();
-                loadtest "virt_autotest/login_console";
             }
         }
-        else {
-            loadtest "virt_autotest/login_console";
-        }
+        loadtest "virt_autotest/login_console";
         loadtest "virt_autotest/install_package";
         loadtest "virt_autotest/update_package";
         loadtest "virt_autotest/reset_partition";
--- a/tests/virt_autotest/libvirt_extend_storage_lvm.pm
+++ b/tests/virt_autotest/libvirt_extend_storage_lvm.pm
@@ -69,7 +69,7 @@ sub prepare_lvm_storage_pool_source {
     # Use a unused hard disk for LVM volumes
     $dev = "/dev/";
     foreach my $disk (@disks) {
-        if (script_run("set -o pipefail;lsblk -rnoPKNAME,MOUNTPOINT | grep -i $disk | awk \'{print \$2}\'") ne '0') {
+        if (script_run("set -o pipefail;findmnt -n -o SOURCE / | grep $disk") != 0) {
             $lvm_disk_name = $dev . $disk;
             last;
         }


Julie-CAO:
I did not look into the detail of autoyast profile, just viewed the process of new host installation. The new autoyast installation run smoothly and faster than needle matching, it looks good. The concerns/thoughts come into my mind:

- the SOL console is still in use during host installation. I understand it is for detecting error and reboot.
- the faults of sol console are its unstability and messy screen, especially during needle matching of pxe menu and grub2 menu, our test sporatically failed at reboot and login steps.
-  If we could change to use ssh console, the faults of sol console can be avoided. Installation error still could be caught but we have to try to detect reboot. It would be prefect if the needle of a reboot could work in test.
- I noticed the host is rebooted twice after installation, after installation and in reboot_and_waitup_normal seperately. The first reboot is not required, it is here I guess it is for sake of detecting a reboot, or reboot to xen is not supported in install.pm. Could a new needle of reboot avoid it, or other means? We try to avoid needing matching, especially for the pxe/grub menu.
- And one thing we'd be aware is when we will change to use 'IPXE' boot to install shortly, no pxe menu will appear any more. Will the current installation still work?
- About xen host, the new setting XEN_DEFAULT_BOOT_IS_SET, which was introduced a few days ago, is expected to to supported in autoyast profile. please refer to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16632.

tonyyuan1:
> I did not look into the detail of autoyast profile, just viewed the process of new host installation. The new autoyast installation run smoothly and faster than needle matching, it looks good. 
Thanks!

The concerns/thoughts come into my mind:
> 
> * the SOL console is still in use during host installation. I understand it is for detecting error and reboot.
> * the faults of sol console are its unstability and messy screen, especially during needle matching of pxe menu and grub2 menu, our test sporatically failed at reboot and login steps.
> * If we could change to use ssh console, the faults of sol console can be avoided. Installation error still could be caught but we have to try to detect reboot. It would be prefect if the needle of a reboot could work in test.

The installation.pm has worked well with maintenance test. I recently have  performed 20+ installation tests in your product test  environment in OSD and  have  not encountered issue with sol  console.  It's working for me so far. 

> * I noticed the host is rebooted twice after installation, after installation and in reboot_and_waitup_normal seperately. The first reboot is not required, it is here I guess it is for sake of detecting a reboot, or reboot to xen is not supported in install.pm. Could a new needle of reboot avoid it, or other means? We try to avoid needing matching, especially for the pxe/grub menu.

one reboot between install stage 1 and stage 2.   A reboot after stage 2  which is impotent to get clean login screen to match a needle to finish the installation. 
It does not need to match "pxe/grub menu" needle in autoyast installation. 
"reboot_and_waitup_normal" is just used for switching to xen hypervisor here.
Booting to xen is not supported in installation.pm.  However, it can be done in autoyast profile.

> * And one thing we'd be aware is when we will change to use 'IPXE' boot to install shortly, no pxe menu will appear any more. Will the current installation still work?
I don't know:). We can't implement it until we get all of specific requirements and technical detail.  After we get them we will be able to do backward compatibility with all what we have.   We can't do "forward compatibility" for future thing.  


> * About xen host, the new setting XEN_DEFAULT_BOOT_IS_SET, which was introduced a few days ago, is expected to to supported in autoyast profile. please refer to [To set the default boot to Xen during host installation where grub2 is modified #16632](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16632).

 it can be done in autoyast profile.

Thank you for your comments, Julie.  Feel free to ask again if you still have question.

Julie-CAO:
> > * the SOL console is still in use during host installation. I understand it is for detecting error and reboot.
> > * the faults of sol console are its unstability and messy screen, especially during needle matching of pxe menu and grub2 menu, our test sporatically failed at reboot and login steps.
> > * If we could change to use ssh console, the faults of sol console can be avoided. Installation error still could be caught but we have to try to detect reboot. It would be prefect if the needle of a reboot could work in test.
> 
> The installation.pm has worked well with maintenance test. I recently have performed 20+ installation tests in your product test environment in OSD and have not encountered issue with sol console. It's working for me so far.
> 

The reason for switching to autoyast is not only to avoid needing matches during the installation process, but also to try to avoid the use of sol console, including the future task of switching to ipxe. So I think we should try not to use it as much as possible unless it is inevitable. The tests did not run into any probem,  which cannot be assumed that in OSD, there will be no problem because the machines are different, the environment is different, and the frequency of host installation is also different. In OSD, the frequency of install hosts is much higher than in MU test.

> > * I noticed the host is rebooted twice after installation, after installation and in reboot_and_waitup_normal seperately. The first reboot is not required, it is here I guess it is for sake of detecting a reboot, or reboot to xen is not supported in install.pm. Could a new needle of reboot avoid it, or other means? We try to avoid needing matching, especially for the pxe/grub menu.
> 
> one reboot between install stage 1 and stage 2. A reboot after stage 2 which is impotent to get clean login screen to match a needle to finish the installation. It does not need to match "pxe/grub menu" needle in autoyast installation. "reboot_and_waitup_normal" is just used for switching to xen hypervisor here. Booting to xen is not supported in installation.pm. However, it can be done in autoyast profile.

One time reboot is more prefered.

> > * And one thing we'd be aware is when we will change to use 'IPXE' boot to install shortly, no pxe menu will appear any more. Will the current installation still work?
> >   I don't know:). We can't implement it until we get all of specific requirements and technical detail.  After we get them we will be able to do backward compatibility with all what we have.   We can't do "forward compatibility" for future thing.

I don't think so. When we implement some feature, the future possibilities are supposed to be considered, especially the known upcoming projects. We look further to prevent more duplicate and re-building efforts.

> > * About xen host, the new setting XEN_DEFAULT_BOOT_IS_SET, which was introduced a few days ago, is expected to to supported in autoyast profile. please refer to [To set the default boot to Xen during host installation where grub2 is modified #16632](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16632).
> 
> it can be done in autoyast profile.

Thanks.

> Thank you for your comments, Julie. Feel free to ask again if you still have question.

tonyyuan1:
With workers in Beijing and your OSD machines NUE, I have run more than 20 autoyast installation tests with sol over the past month (including VRs above). I didn't see any sol problem. I don't understand what sol caused problem in your tests unless you provide some examples.
 
Anyway, I don't have strong option what console you use. It's up to you.  installation.pm can work with almost all consoles. It uses the console that its' previous test left at.   You can simply switch to your preferred console before it starts. 

Regarding reboot, there are three.  I guess you want to discard "reboot_and_waitup_normal" and set xen to default boot in profile, right?

Julie-CAO:
> With workers in Beijing and your OSD machines NUE, I have run more than 20 autoyast installation tests with sol over the past month (including VRs above). I didn't see any sol problem. I don't understand what sol caused problem in your tests unless you provide some examples.

sol console does not always have problem, but it is not as stable as ssh console.  For each build, there are 60+ tests with host installations, there will be 100 tests including test reruns. 20 successful tests can't indicate there are not sol problems. I have not an good example right now because most of SUTs in OSD are not working well these days and scc is inaccessible.


> Anyway, I don't have strong option what console you use. It's up to you. installation.pm can work with almost all consoles. It uses the console that its' previous test left at. You can simply switch to your preferred console before it starts.

I can accept sol console only when it is inevitable. Since installation.pm can work with all consoles, I prefer to the ssh console personally. As of which console should be used, Alice and Wayne and others can give more input.

> Regarding reboot, there are three. I guess you want to discard "reboot_and_waitup_normal" and set xen to default boot in profile, right?

I did not notice there are 3 reboots. I only see the one at the end of installation.pm and the second one is in reboot_and_waitup_normal. If possible, one reboot would be better regardless it is in which module.

"set xen to default boot in profile" only with XEN_DEFAULT_BOOT_IS_SET=1 if it is capable.

tonyyuan1:
What I understand from your comments is:
The sol sometimes breaks GUI installation in your OSD tests.  You are concerned that sol would break autoyast installation which you never run, right?
If this's the case I would say that GUI installation and autoyast are quite different. My recommendation:
Keep sol and wait to see if it breaks autoyast install in your tests.  If it does, then change console of you choice. 

Reagrding xen default boot,
The logic:  
With xen installed, set xen to default boot
Without xen installed, do nothing

The var "system_role" can be used for switching. It's necessary to introduce a new var "XEN_DEFAULT_BOOT_IS_SET"?

Julie-CAO:
> What I understand from your comments is: The sol sometimes breaks GUI installation in your OSD tests. You are concerned that sol would break autoyast installation which was never run in your OSD tests, right? If this's the case I would say that GUI installation and autoyast are quite different. My recommendation (based on my tests): Keep sol and wait to see if it breaks autoyast install in your tests. If it does, then change console of you choice.

No, it is not sol console but ssh console being used to install hosts with textmode yast install wizard in OSD. sol console is used in boot_from_pxe, reboot and login. It has not been stable in these steps, so we tend to gradually abandon it. But you have instead changed from ssh console to sol console here, so I am not very in favor unless you have sufficient reasons. 


> Reagrding xen default boot, The logic: With xen installed, set xen to default boot Without xen installed, do nothing
> 
> The var "system_role" can be used for switching. It's necessary to introduce a new var "XEN_DEFAULT_BOOT_IS_SET"?

Introducing a new var "XEN_DEFAULT_BOOT_IS_SET" was decided in that PR I mentioned. If you don't agree, you can raise it up again to team, but now we expecting existing tests would not be broken by new changes since not all hosts are installed with this autoyast profile.

tonyyuan1:
> No, it is not sol console but ssh console being used to install hosts with textmode yast install wizard in OSD. sol console is used in boot_from_pxe, reboot and login. It has not been stable in these steps, so we tend to gradually abandon it. 
You are right, "yast install wizard" is not stable. 
The main reason:  it has to take a lot of  screenshots in order to compare with needles. It takes a lot traffic. 
Autoyast install is fully automatic with no graphical wizard. I take few screenshots and it just need  to compare nothing but a  "login" screenshot.

tonyyuan1:
There is a reboot between install stage 1 and 2 in  Autoyast installation. The reboot will disconnect ssh connection.

waynechen55:
Not very sure whether profile here is the same as those in folder ```data/virtualization/autoyast```. Additionally, I think autoyast file template should allow granular control over bootloader, service, software, system config and support cusotmized scripts from my personal perspective.

tonyyuan1:
> Not very sure whether profile here is the same as those in folder `data/virtualization/autoyast`. 
It's not the same.  virtualization/autoyast are profile for maintenance tests.   The profile in this PR satisfy product test needs as the some test scripts and hardware are different. 
For example:  
disk array.  xen default boot

Additionally, I think autoyast file template should allow granular control over bootloader, service, software, system config and support cusotmized scripts from my personal perspective.

What does " granular control " mean?

tonyyuan1:
VR with xen default boot
kvm, http://10.67.183.130/tests/3567
xen, http://10.67.183.130/tests/3552

waynechen55:
> > Not very sure whether profile here is the same as those in folder `data/virtualization/autoyast`.
> 
> It's not the same. virtualization/autoyast are profile for maintenance tests. The profile in this PR satisfy product test needs as the some test scripts and hardware are different. For example: disk array, xen default boot
> 
> > Additionally, I think autoyast file template should allow granular control over bootloader, service, software, system config and support cusotmized scripts from my personal perspective.
> 
> What does " granular control " mean?

For example:
1) User can use customized boot settings, something on the kernel command line. This is supported in GUI installation as well. So I think autoyast should also support this
2) User can use customized scripts, please refer to https://susedoc.github.io/doc-sle/main/single-html/SLES-autoyast/#createprofile-scripts. So user can change system settings that can not be done during installation phase.
3) User can specify which service and software to be installed during installation.

tonyyuan1:
> 1. User can use customized boot settings, something on the kernel command line. This is supported in GUI installation as well. So I think autoyast should also support this
Yes, you can do this in either bootloader section or postscript.

> 2. User can use customized scripts, please refer to https://susedoc.github.io/doc-sle/main/single-html/SLES-autoyast/#createprofile-scripts. So user can change system settings that can not be done during installation phase.

Yes
> 3. User can specify which service and software to be installed during installation.

Yes, 
Note: All above setups are able to be implemented in autoyast very easily. 

In summary, you can move any OS configuration and other setups in test modules into profile, so the logic in  test modules become very clean:  just about test case execution.

waynechen55:
> > 1. User can use customized boot settings, something on the kernel command line. This is supported in GUI installation as well. So I think autoyast should also support this
> 
> Yes, you can do this in either bootloader section or postscript.
> 
> > 2. User can use customized scripts, please refer to https://susedoc.github.io/doc-sle/main/single-html/SLES-autoyast/#createprofile-scripts. So user can change system settings that can not be done during installation phase.
> 
> Yes
> 
> > 3. User can specify which service and software to be installed during installation.
> 
> Yes, Note: All above setups are able to be implemented in autoyast very easily.
> 
> In summary, you can move any OS configuration and other setups in test modules into profile, so the logic in test modules become very clean: just about test case execution.

If you think all these are supported, could you help provide verification run for them as well ?

waynechen55:
Multiple ARCHs covered ? If yes, I think there are some places to be refined.

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,267 @@
...
+      <os_prober>false</os_prober>
+      <serial>serial --speed=115200 --unit=<%= $get_var->('SERIALDEV') =~ s/ttyS//r %> --word=8 --parity=no --stop=1</serial>
+      <timeout config:type="integer">10</timeout>
+      <trusted_grub>false</trusted_grub>
+      <terminal>console</terminal>
+      % if ($check_var->('SYSTEM_ROLE', 'xen')) {
+      <xen_append>splash=silent quiet console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %></xen_append>
+      <xen_kernel_append>dom0_max_vcpus=4 console=com2,115200 vga=gfx-1024x768x16 loglvl=all guest_loglvl=all sync_console</xen_kernel_append>
+      % } else {
+      <append>splash=silent quiet console=<%= $get_var->('SERIALDEV') %>,115200 console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %></append>
```
waynechen55, line 59: I do not think ```quiet``` is needed. Better to remove it ?

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,267 @@
...
+      % }
+    </global>
+    <loader_type>grub2</loader_type>
```
waynechen55, line 62: Will ```grub2-efi``` be an option ?

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,267 @@
...
+    <languages/>
+  </language>
+  <login_settings/>
+  <networking>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+    <backend>wicked</backend>
+    <dns>
+      <dhcp_hostname config:type="boolean">true</dhcp_hostname>
+      <nameservers config:type="list">
+        <nameserver>10.162.0.1</nameserver>
```
waynechen55, line 102: Hard-coded nameserver ?

tonyyuan1:
> > In summary, you can move any OS configuration and other setups in test modules into profile, so the logic in test modules become very clean: just about test case execution.
> 
> If you think all these are supported, could you help provide verification run for them as well ?

They are not supported by this PR. It's "autoyast" which supports them.  This PR provide a template to generate autoyast profile which control autoyast behaviors. 

The template in this PR will do what all of  the modules do before "login_console" in a test below ,  meaning delivering a installed sles15 OS with some basic configuration.  https://openqa.suse.de/tests/10941593

What I mean by "move any OS configuration and other setups in test modules into profile: 
For example:
Suppose that you want install qa_lib_virtauto package.  You simply add tag  <package>qa_lib_virtauto</package> in profile. 
Autoyast will install it for you during OS installation.

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: gonzo and a another machine ( I forgot name) could not get nameserver from local dns server after autoyast installation.  I don't know the reason so just put a static nameserver option in there.  If the machine can get nameserver info from dhcp they will be merged. If not, it will use the static.

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: I checked that all your OSD machines were bios boot.  Change "grub2" to "default" when you have machine with uefi boot set in future.

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: > I do not think `quiet` is needed. Better to remove it ?

Product QA will be users of this template. I will remove it if you don't need it.

tonyyuan1:
> Multiple ARCHs covered ? If yes, I think there are some places to be refined. And non-SLES will be supported or not ?

Julie said just need to cover x86.   
autoyast is feature of yast. yast is tool developed by SUSE.  I don't know if yast can install non-SLES os. I guest not.

data/virt_autotest/autoyast/host_15.xml.ep
waynechen55: No, we do not need this ```quiet```.

waynechen55:
All verification runs have ```SCC_ADDONS=base,sdk,serverapp,desktop,legacy``` but actually no extensions are registered, for example, [this one](http://10.67.183.130/tests/3492#step/install_package/17).  The questions are:
1. Does ```SCC_ADDONS``` still apply ? I think this question can be extended to other settings as well. There appearance in test suite should also be reflected in actual test run, otherwise they should not be included in.
2. Does your pull request work with multiple extensions registration ? For eample, base,sdk,serverapp,desktop,legacy ?

tonyyuan1:
> All verification runs have `SCC_ADDONS=base,sdk,serverapp,desktop,legacy` but actually no extensions are registered, for example, [this one](http://10.67.183.130/tests/3492#step/install_package/17). 

It was activating "package hub" in screenshot 17.  
Looking at screenshots before 17,  for instance in screenshot 16,  you will find that the "server application" and "legacy" extensions were already activated. (the yellowish ones) .

The questions are:
> 
> 1. Does `SCC_ADDONS` still apply ? I think this question can be extended to other settings as well. There appearance in test suite should also be reflected in actual test run, otherwise they should not be included in.
> 2. Does your pull request work with multiple extensions registration ? For example, base,sdk,serverapp,desktop,legacy ?

SCC_ADDONS is applied.  JFYI. The generated profile was uploaded.  Go to Logs & Assests for a file: autoyast-host_15.xml.  You'll see registration and extensions in there.

waynechen55:
Previous GUI installation upload grub file after installation finishes, but autoyast installation does not do this anymore. I think we should upload grub as well for autoyast installation.

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,267 @@
...
+    </add_on_products>
+  </add-on-->
+  <bootloader>
+    <global>
+      <activate>true</activate>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --speed=115200 --unit=<%= $get_var->('SERIALDEV') =~ s/ttyS//r %> --word=8 --parity=no --stop=1</serial>
+      <timeout config:type="integer">10</timeout>
```
waynechen55, line 52: Could you help change timeout value to 15 ?

tonyyuan1:
> 1. Previous GUI installation upload grub file and y2logs after installation finishes, for example, [grub file](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-grub.cfg) and [y2logs](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-y2logs.tar.bz2), but autoyast installation does not do this anymore. I think we should upload grub and y2logs as well for autoyast installation.

autoayst install is fully automated and done by yast.  yast is unaware of openQA.  Uploading files from SUT has to be done in a test module after installation finishes (after final reboot).  Login SUT with select_console then use  upload_log. 

it would be good to open new ticket to address it.  


> 2. Previous GUI installation supports `OPT_KERNEL_PARAMS` in module `tests/installation/edit_optional_kernel_cmd_parameters.pm`. Does your autoyast implementation support similar functionality ?
No,
I checked that "edit_optional_kernel_cmd_parameters.pm" is loaded only when testing fips with micro.  
It need add/remove kernel parameters and install additional packages. I would suggest to create another profile template for a different product rather than  put a lot of logic in one template.

data/virt_autotest/autoyast/host_15.xml.ep
alice-suse: Would you please add a comment for it? What you explain here will be good enough.

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
+<?xml version="1.0"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <suse_register>
+    <do_registration config:type="boolean">true</do_registration>
+    <reg_server>{{SCC_URL}}</reg_server>
+    <reg_code>{{SCC_REGCODE}}</reg_code>
```
alice-suse, line 7: For reg_server and reg_code, sle15sp4 host may need specific handling here.

In OSD, 15sp4 host tests use the way in testsuite gi-guest_developing-on-host_sles15sp4-kvm, a recent job link https://openqa.suse.de/tests/10941579. The test also use sle-15-SP5-Online-x86_64 medium, which has below settings. And settings `INSTALL_TO_OTHERS` and `REPO_0_TO_INSTALL` are used to guide host installation to use the 15sp4 images.

sle	15-SP5	Online	x86_64	BETA=1, medium setting is
```
BUGZILLA_URL=http://fastzilla.suse.de/short@BUGID@.xml
BUILD_HA=%BUILD%
BUILD_SDK=%BUILD%
BUILD_SES=%BUILD%
ISO_MAXSIZE=737280000
LTP_GIT_URL=https://github.com/linux-test-project/ltp
LTP_KNOWN_ISSUES=http://qam.suse.de/media/downloads/ltp_known_issues.json
QA_HEAD_REPO=http://dist.suse.de/ibs/QA:/Head/SLE-%VERSION%
SCC_REGCODE=INTERNAL-USE-ONLY-1977-6205
SCC_REGCODE_BCL=35e031a536494293
SCC_REGCODE_GEO=07db9ab0d578f087
SCC_REGCODE_HA=223378a848e109bd
SCC_REGCODE_HPC=bdb13b194f1fece2
SCC_REGCODE_LIVE=f86c0b728fbc04c1
SCC_REGCODE_LTSS_15=ba36407f77ddba37
SCC_REGCODE_RT=5ed98f8c543b55b7
SCC_REGCODE_SES=35eb00e8035bf5b0
SCC_REGCODE_SLES4SAP=2882173e156fbd43
SCC_REGCODE_WE=234a924b6c0dc761
SCC_URL=http://all-%BUILD%.proxy.scc.suse.de
SLENKINS_TESTSUITES_REPO=http://download.suse.de/ibs/Devel:/SLEnkins:/testsuites/SLE_15/
```

If SCC_URL, SCC_REGCODE is used, sle15sp4 would be using sle15sp5 registration. Better to change .

Besides, another info from Julie. Related with 12sp5 autoyast host installation ticket. For sle12sp5 host , in OSD test, the MU update is done actually by the way shown in https://openqa.suse.de/tests/10941166#step/update_package/1
.  So actually we can skip registration in sle12sp5 host autoyast installation. @Julie-CAO  Please correct me if I am wrong.

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
...
+        <reg_code><%= $get_var->('SCC_REGCODE_WE') %></reg_code>
+        % }
+        % if ($key eq 'we' and $check_var->('SLE_PRODUCT', 'sled')) {
+        <reg_code><%= $get_var->('SCC_REGCODE') %></reg_code>
+        % }
+        % if ($key eq 'rt') {
+        <reg_code><%= $get_var->('SCC_REGCODE_RT') %></reg_code>
+        % }
+        % if ($key eq 'ltss') {
+        <reg_code><%= $get_var->('SCC_REGCODE_LTSS') %></reg_code>
```
alice-suse, line 26: Similar considerations for these scc regcodes for addons.

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
...
+      </ntp_server>
+    </ntp_servers>
+    <ntp_sync>systemd</ntp_sync>
+  </ntp-client>
+  <partitioning config:type="list">
+  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b'};
+  % my $hostname = (split(/\./, $get_var->("SUT_IP")))[0];
+  % my $device_id = defined($wwn->{$hostname}) ? '/dev/disk/by-id/' . $wwn->{$hostname} : ''; 
+    <drive>
+      <device><%= $device_id %></device>
```
alice-suse, line 125: Would you please explain the purpose/benefit of setting $wwn? In OSD, we have more than 6 SUTs. Why 6 SUTs are configured here? Miss some SUT or only need to set them?

Besides, if there can be other methods/workarounds to achieve the given purpose,  it will be good to avoid giving hardware specific info in autoyast. Otherwise, once we have new machines, we will always have to update autoyast. 
If it is not avoidable, maybe moving such hardware info (wwn) to worker.ini, then extract the settings by get_var and use it here, will be better?

data/virt_autotest/autoyast/host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
...
+        </partition>
+        <partition>
+          <create config:type="boolean">true</create>
+          <filesystem config:type="symbol">xfs</filesystem>
+          <format config:type="boolean">true</format>
+          <mount>/var/lib/libvirt/images/</mount>
+          <mountby config:type="symbol">uuid</mountby>
+          <partition_id config:type="integer">131</partition_id>
+          <resize config:type="boolean">false</resize>
+          <size>100G</size>
```
alice-suse, line 147: Can we enlarge this to 160G? I remember our disks are large enough. This will give more freedom in using disk spaces.

alice-suse:
> > Multiple ARCHs covered ? If yes, I think there are some places to be refined. And non-SLES will be supported or not ?
> 
> Julie said just need to cover x86. autoyast is feature of yast. yast is tool developed by SUSE. I don't know if yast can install non-SLES os. I guest not.

@waynechen55  You are more familiar with the stability of aarch64 VT test. Do you think it is needed to switch to autoyast host installation for arm? If yes, we can open new ticket to do that, maybe not in this PR.

alice-suse:
> There is a reboot between install stage 1 and 2 in Autoyast installation. The reboot will disconnect ssh connection.

@tonyyuan1 , I guess you tried to switch to root-ssh console for autoyast installation based on @Julie-CAO  comments? Do you mean the reboot between stage 1 and 2 is a blocker from using root-ssh console?

Can it be workarounded? It will be good to really rely on root-ssh console to have high enough stability based on our experience. This is first choice. If autoyast installation code know the point of stage 1 ends and stage 2 starts, maybe it is workaroundable by switching consoles around the point? We have such usage in existing scripts, eg reboot_and_wait_up_xx.pm. 

If it is not able to workaround and sol console is used for autoyast installation, would you please share your observations? How many times do you run and on how much of our machine pool hardwares on OSD? What's the stability number/pass ratio?

data/virt_autotest/autoyast/host_15.xml.ep
Julie-CAO: > Besides, another info from Julie. Related with 12sp5 autoyast host installation ticket. For sle12sp5 host , in OSD test, the MU update is done actually by the way shown in https://openqa.suse.de/tests/10941166#step/update_package/1 . So actually we can skip registration in sle12sp5 host autoyast installation. @Julie-CAO Please correct me if I am wrong.

Yes, for tests on sle12sp5 host, "SCC_REGISTER=none" is in use for now. but I don't be against to change to do update via scc in the future.

waynechen55:
> > 1. Previous GUI installation upload grub file and y2logs after installation finishes, for example, [grub file](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-grub.cfg) and [y2logs](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-y2logs.tar.bz2), but autoyast installation does not do this anymore. I think we should upload grub and y2logs as well for autoyast installation.
> 
> autoayst install is fully automated and done by yast. yast is unaware of openQA. Uploading files from SUT has to be done in a test module after installation finishes (after final reboot). Login SUT with select_console then use upload_log.
> 
> it would be good to open new ticket to address it.

I think "post script" or similar thing can help do this ? upload grub and y2logs.

tonyyuan1:
> @tonyyuan1 , I guess you tried to switch to root-ssh console for autoyast installation based on @Julie-CAO comments? Do you mean the reboot between stage 1 and 2 is a blocker from using root-ssh console?

There must be sshd so as to create a ssh connection. 
In stage one, the linuxrc brings up a sshd service when it see the kernel parameter "sshd=1"
Stage one is installation environment, meaning that the running kernel is download from pxe and the root filesystem is ramfs.

What Autoyast does in stage one is partitioning and installing packages.

It need reboot to mount the "real" root file system and load the installed kernel to do configurations in real os environment. That's stage two.
The reboot makes sshd service (started by linurc) gone. Any external ssh connections are broken. 
After configuration is done a new sshd will started but we don't know when it starts.  so we don't know when to connect it from openqa console. 

Moreover, when the sshd is started the whole installation is close to finish. 

> 
> Can it be workarounded? It will be good to really rely on root-ssh console to have high enough stability based on our experience. This is first choice. If autoyast installation code know the point of stage 1 ends and stage 2 starts, maybe it is workaroundable by switching consoles around the point? We have such usage in existing scripts, eg reboot_and_wait_up_xx.pm.

I don't have workarounded. The complicated workaround might make more unstability than stability. 
> 
> But if it can't workaround and sol console is used for autoyast installation, would you please share your observations? How many times do you run and on how much of our machine pool hardwares on OSD? What's the stability number/pass ratio?

I have have run about 30 install tests with that 6 OSD hosts. I did not encounter any issue caused by console

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: 1. It uses SCC proxy for developing product. 15sp5 currently.
2. It uses the default: SCC,  for released product.  15sp4 and other 15spx.  It installs the latest packages during os install so you don't need update them later.  update_package might have nothing to do.
http://10.67.183.130/tests/3577

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: > Would you please explain the purpose/benefit of setting $wwn? 
In OSD, we have more than 6 SUTs. Why 6 SUTs are configured here? Miss some SUT or only need to set them?
>
> Besides, if there can be other methods/workarounds to achieve the given purpose, it will be good to avoid giving hardware specific info in autoyast. 

They are using lun from a array storage (maybe two) .  I have to use persistent device id to make sure OS is installed on first lun.  It's not bootable with OS on other luns.  see more in the ticket.

Let me know the name of missing SUTs. 

A question:  After your tests install a sles build and reboot. How you do know if the installed sles is the right build rather then you boot a previously installed os from the second disk.  

Otherwise, once we have new machines, we will always have to update autoyast. If it is not avoidable, maybe moving such hardware info (wwn) to worker.ini, then extract the settings by get_var and use it here, will be better?

If the new machine uses lun as block device you have to update autoyast.  
Lun is not a hardware of machine.
Putting them in autoyast make people understand their  purpose easily as it's used in here. 
People like to edit one file more than edit 6 files.

waynechen55:
> > > Multiple ARCHs covered ? If yes, I think there are some places to be refined. And non-SLES will be supported or not ?
> > 
> > 
> > Julie said just need to cover x86. autoyast is feature of yast. yast is tool developed by SUSE. I don't know if yast can install non-SLES os. I guest not.
> 
> @waynechen55 You are more familiar with the stability of aarch64 VT test. Do you think it is needed to switch to autoyast host installation for arm? If yes, we can open new ticket to do that, maybe not in this PR.

Created poo#128378 as subtask to track this.

waynechen55:
> > 1. Previous GUI installation upload grub file and y2logs after installation finishes, for example, [grub file](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-grub.cfg) and [y2logs](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-y2logs.tar.bz2), but autoyast installation does not do this anymore. I think we should upload grub and y2logs as well for autoyast installation.
> 
> autoayst install is fully automated and done by yast. yast is unaware of openQA. Uploading files from SUT has to be done in a test module after installation finishes (after final reboot). Login SUT with select_console then use upload_log.
> 
> it would be good to open new ticket to address it.
> 
> > 2. Previous GUI installation supports `OPT_KERNEL_PARAMS` in module `tests/installation/edit_optional_kernel_cmd_parameters.pm`. Does your autoyast implementation support similar functionality ?
> 
> No, I checked that "edit_optional_kernel_cmd_parameters.pm" is loaded only when testing fips with micro. It need add/remove kernel parameters and install additional packages. I would suggest to create another profile template for a different product rather than put a lot of logic in one template.

Created poo#128375 and poo#128381 as subtask to track.

data/virt_autotest/autoyast/host_15.xml.ep
alice-suse: Hi Tony, from job 3577, indeed 15sp4 product is installed. It is validated with wrong testsutie gi-guest_developing-on-host_developing-xen. Testsuite gi-guest_developing-on-host_sles15sp4-kvm/xen are the ones for 15sp4 host. 

> It uses the default: SCC, for released product. 15sp4 and other 15spx.

And for the testsuite settings that need changes to work in the way you describe, I guess it will require overwriting SCC_REGCODE and SCC_URL in testsuite settings? @Julie-CAO @waynechen55   Will it affect other parts of registration/update in tests, eg sle15sp5 guest registration on sle15sp4 host, host upgrade, guest upgrade tests? I guess not based on my info. But I may overlook some details.

Besides, if this won't break any others with this way with their confirmation, I agree with it. Any other testsuite changes needed? We may need you to help summarize it on per-testsuite level and share to @Julie-CAO , then she will be able to help with further update on OSD, to let all work as expected :).

data/virt_autotest/autoyast/host_15.xml.ep
alice-suse: All machines in https://gitlab.suse.de/openqa/salt-pillars-openqa/-/blob/master/openqa/workerconf.sls with worker class `64bit-ipmi` are what we use on OSD :) .

I am not very convinced by the reasons of not putting in worker.ini. But I do not insist on this. Just share my idea here.
- Still 1 file needs to be edited, workerconf.sls which is automatically deployed by salt to worker.ini . 
- LUN id seems closely related with each SUT, fine enough to put in worker.ini .

@Julie-CAO How do you handle in TW autoyast to let installation done on first disk? Will it be good to unify the method for this?

Regarding how we confirm the correct os is booted from, we have been always installing on first disk by selecting first disk during interactive os installation. So there is no os at all on second disk. No such trouble to worry :)

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: SCC_REGCODE is the same but set SCC_URL to empty string.  
 https://openqa.suse.de/tests/10985233#

I posted how I triggered  In the ticket

alice-suse:
> > @tonyyuan1 , I guess you tried to switch to root-ssh console for autoyast installation based on @Julie-CAO comments? Do you mean the reboot between stage 1 and 2 is a blocker from using root-ssh console?
> 
> There must be sshd so as to create a ssh connection. In stage one, the linuxrc brings up a sshd service when it see the kernel parameter "sshd=1" Stage one is installation environment, meaning that the running kernel is download from pxe and the root filesystem is ramfs.
> 
> What Autoyast does in stage one is partitioning and installing packages.
> 
> It need reboot to mount the "real" root file system and load the installed kernel to do configurations in real os environment. That's stage two. The reboot makes sshd service (started by linurc) gone. Any external ssh connections are broken. After configuration is done a new sshd will started but we don't know when it starts. so we don't know when to connect it from openqa console.
> 
> Moreover, when the sshd is started the whole installation is close to finish.
> 
> > Can it be workarounded? It will be good to really rely on root-ssh console to have high enough stability based on our experience. This is first choice. If autoyast installation code know the point of stage 1 ends and stage 2 starts, maybe it is workaroundable by switching consoles around the point? We have such usage in existing scripts, eg reboot_and_wait_up_xx.pm.
> 
> I don't have workaround. The complicated workaround might make more instability than stability.
> 

Thanks for the explanations. Agree.

> > But if it can't workaround and sol console is used for autoyast installation, would you please share your observations? How many times do you run and on how much of our machine pool hardwares on OSD? What's the stability number/pass ratio?
> 
> I have have run about 30 install tests with that 6 OSD hosts. I did not encounter any issue caused by console

Let me also add some side info from HANA PERF team. They has switched to ipxe+autoyast for host installation on ipmi machines. For autoyast installation, they use sol console too. Based on Dawei's observation, the historical jobs in the past several months show 99.9% pass ratio under HANA network.

Let's try sol too.

alice-suse:
> > > 1. Previous GUI installation upload grub file and y2logs after installation finishes, for example, [grub file](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-grub.cfg) and [y2logs](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-y2logs.tar.bz2), but autoyast installation does not do this anymore. I think we should upload grub and y2logs as well for autoyast installation.
> > 
> > 
> > autoayst install is fully automated and done by yast. yast is unaware of openQA. Uploading files from SUT has to be done in a test module after installation finishes (after final reboot). Login SUT with select_console then use upload_log.
> > it would be good to open new ticket to address it.
> > > 2. Previous GUI installation supports `OPT_KERNEL_PARAMS` in module `tests/installation/edit_optional_kernel_cmd_parameters.pm`. Does your autoyast implementation support similar functionality ?
> > 
> > 
> > No, I checked that "edit_optional_kernel_cmd_parameters.pm" is loaded only when testing fips with micro. It need add/remove kernel parameters and install additional packages. I would suggest to create another profile template for a different product rather than put a lot of logic in one template.
> 
> Created poo#128375 and poo#128381 as subtask to track.

@waynechen55  Without finishing above two new subtasks, in your opinion, can we switch to autoyast host installation on OSD for x86 tests? Would you please also share the main points supporting your statement for it?

data/virt_autotest/autoyast/host_15.xml.ep
alice-suse: Tony, I guess you mean this comment, https://progress.opensuse.org/issues/112508#note-15? Well, in OSD, the tests are automatically triggered, not like qem jobs which we trigger by ourselves(free to customize settings). So for OSD VT jobs, any change needed in settings, we need to figure out a solution -- updating them in testsuite or medium or wherever, in a suitable way.

waynechen55:
> > > > 1. Previous GUI installation upload grub file and y2logs after installation finishes, for example, [grub file](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-grub.cfg) and [y2logs](https://openqa.suse.de/tests/10946088/file/logs_from_installation_system-y2logs.tar.bz2), but autoyast installation does not do this anymore. I think we should upload grub and y2logs as well for autoyast installation.
> > > 
> > > 
> > > autoayst install is fully automated and done by yast. yast is unaware of openQA. Uploading files from SUT has to be done in a test module after installation finishes (after final reboot). Login SUT with select_console then use upload_log.
> > > it would be good to open new ticket to address it.
> > > > 2. Previous GUI installation supports `OPT_KERNEL_PARAMS` in module `tests/installation/edit_optional_kernel_cmd_parameters.pm`. Does your autoyast implementation support similar functionality ?
> > > 
> > > 
> > > No, I checked that "edit_optional_kernel_cmd_parameters.pm" is loaded only when testing fips with micro. It need add/remove kernel parameters and install additional packages. I would suggest to create another profile template for a different product rather than put a lot of logic in one template.
> > 
> > 
> > Created poo#128375 and poo#128381 as subtask to track.
> 
> @waynechen55 Without finishing above two new subtasks, in your opinion, can we switch to autoyast host installation on OSD for x86 tests? Would you please also share the main points supporting your statement for it?

1. Installing with autoyast is still possible even without these two subtasks. These two tasks have no impact on basic auotyast functionality and installation process
2. But grub and y2logs are important logs files that can facilitate failure investigating and debugging conveniently.
3. OPT_KERNEL_PARAMS is there to enhance host installation process with user-defined kernel parameters, for example, [kubevrit on sle micro uses this](https://openqa.suse.de/tests/10855489). So it can also be used in other scenarios as well. Not only sle-micro, but also sles.  I think this is easy to be implemented in autoyast.

tonyyuan1:
> 1. I think a "post script" or similar thing in autoyast file can help do this ? upload grub and y2logs.
>    or
To upload in "post script" need destination openqa url  which need job id and maybe a token.  complicated. 

> 2. `tests/installation/logs_from_installation_system.pm` can still be used I think, for example, after autoyast installation.

This module will change grub so can't use it

tonyyuan1:
> 2. But grub and y2logs are important logs files that can facilitate failure investigating and debugging conveniently. For new product, if host installation fails and can not boot up, I think these two files can be very useful.
I checked code a bit:
1. If the install fail the post_fail_hook in  autoyast/installation.pm will call "save_upload_y2logs". 
2. With  autoyast install the grub and kernel parameter are predefined in profile. It will never be changed. They are fixed. If a virt test fails  the  post_fail_hook  upload grub. 


> 3. OPT_KERNEL_PARAMS is there to enhance host installation process with user-defined kernel parameters, for example, [kubevrit on sle micro uses this](https://openqa.suse.de/tests/10855489). At least, this kubevirt test can not switch to use autoyast without this parameter I think. So it can also be used in other scenarios as well. Not only sle-micro, but also sles.  I think this is easy to be implemented in autoyast.

"tests/installation/edit_optional_kernel_cmd_parameters.pm"  edit grub with needles.  With this PR merged in we still can use regular UI installation, meaning that there are two install options:

1. regular UI installation
2. autoyast installation

 SO sle micro test can still use regular UI installation.

tonyyuan1:
> 2. But grub and y2logs are important logs files that can facilitate failure investigating and debugging conveniently. For new product, if host installation fails and can not boot up, I think these two files can be very useful.
I checked code a bit:
1. If the install fail the post_fail_hook in  autoyast/installation.pm will call "save_upload_y2logs". 
2. With  autoyast install the grub and kernel parameter are predefined in profile. It will never be changed. They are fixed. If a virt test fails  the  post_fail_hook  upload grub. 


> 3. OPT_KERNEL_PARAMS is there to enhance host installation process with user-defined kernel parameters, for example, [kubevrit on sle micro uses this](https://openqa.suse.de/tests/10855489). At least, this kubevirt test can not switch to use autoyast without this parameter I think. So it can also be used in other scenarios as well. Not only sle-micro, but also sles.  I think this is easy to be implemented in autoyast.

"tests/installation/edit_optional_kernel_cmd_parameters.pm"  edit grub with needles.  With this PR merged in we still can use regular UI installation, meaning that there are two install options:

1. regular UI installation
2. autoyast installation

 SO sle micro test can still use regular UI installation.

tonyyuan1:
> 2. But grub and y2logs are important logs files that can facilitate failure investigating and debugging conveniently. For new product, if host installation fails and can not boot up, I think these two files can be very useful.
I checked code a bit:
1. If the install fail the post_fail_hook in  autoyast/installation.pm will call "save_upload_y2logs". 
2. With  autoyast install the grub and kernel parameter are predefined in profile. It will never be changed. They are fixed. If a virt test fails  the  post_fail_hook  upload grub. 


> 3. OPT_KERNEL_PARAMS is there to enhance host installation process with user-defined kernel parameters, for example, [kubevrit on sle micro uses this](https://openqa.suse.de/tests/10855489). At least, this kubevirt test can not switch to use autoyast without this parameter I think. So it can also be used in other scenarios as well. Not only sle-micro, but also sles.  I think this is easy to be implemented in autoyast.

"tests/installation/edit_optional_kernel_cmd_parameters.pm"  edit grub with needles.  With this PR merged in we still can use regular UI installation, meaning that there are two install options:

1. regular UI installation
2. autoyast installation

 SO sle micro test can still use regular UI installation.

data/virt_autotest/autoyast/host_15.xml.ep
tonyyuan1: The first disk size of some machines are 223 GB.  If giving it 160, the root file system will be about 60GB. Do you want this?

alice-suse: Oh, sorry. I should have checked https://confluence.suse.com/display/qasleapac2/Virtualization+function+HWs before giving the comment. We indeed have 4 machines with 240G first disk, 1 machine with 280G, while others are much larger. 

HDYT? Lowering to 140G or 120G?

Julie-CAO:
> > I have have run about 30 install tests with that 6 OSD hosts. I did not encounter any issue caused by console
> 
> Let me also add some side info from HANA PERF team. They has switched to ipxe+autoyast for host installation on ipmi machines. For autoyast installation, they use sol console too. Based on Dawei's observation, the historical jobs in the past several months show 99.9% pass ratio under HANA network.
> 
> Let's try sol too.

Thanks. Since you all have such stable cases, I'll have a try with sol console in my TW autoyast installation. If it works as well, I'll change to use installation.pm. If not, ssh console will be used. I have tried, the 2nd stage for TW host installation is not mandatory, at lease I have not found any issue without 2nd stage in autoyast.

waynechen55:
> > 2. But grub and y2logs are important logs files that can facilitate failure investigating and debugging conveniently. For new product, if host installation fails and can not boot up, I think these two files can be very useful.
> >    I checked code a bit:
> 
>     1. If the install fail the post_fail_hook in  autoyast/installation.pm will call "save_upload_y2logs".
> 
>     2. With  autoyast install the grub and kernel parameter are predefined in profile. It will never be changed. They are fixed. If a virt test fails  the  post_fail_hook  upload grub.
> 
> 
> > 3. OPT_KERNEL_PARAMS is there to enhance host installation process with user-defined kernel parameters, for example, [kubevrit on sle micro uses this](https://openqa.suse.de/tests/10855489). At least, this kubevirt test can not switch to use autoyast without this parameter I think. So it can also be used in other scenarios as well. Not only sle-micro, but also sles.  I think this is easy to be implemented in autoyast.
> 
> "tests/installation/edit_optional_kernel_cmd_parameters.pm" edit grub with needles. With this PR merged in we still can use regular UI installation, meaning that there are two install options:
> 
>     1. regular UI installation
> 
>     2. autoyast installation
> 
> 
> SO sle micro test can still use regular UI installation.

Better provide verification examples which provide grub and y2logs. And update progress ticket as well poo#128375.

Julie-CAO:
> VR with xen default boot kvm, http://10.67.183.130/tests/3567 xen, http://10.67.183.130/tests/3552

Hi Tony, I noticed the extra reboot was removed from your tests. It good. but can we manage to keep some necessary checks as we have in the current test, such as double_check_xen_role() and check_host_health().   And set_ssh_console_timeout_before_use() if it is not configured in autoyast.

data/virt_autotest/autoyast/host_15.xml.ep
Julie-CAO: I agree to put these wwns for particular SUTs in workers.ini in salt. They are more dependent on machine itself than the autoyast profile. Given the hostname of a SUT is changed in salt, the autoyast have to be changed as well. Or if a new machine is added in salt, the autoyast has be changed well. Putting them in workers.ini and add an setting will save future maintaince effort.

tonyyuan1:
> Better provide verification examples which provide grub and y2logs. And update progress ticket as well poo#128375.

1. Updated poo#128375.
2.  I guess that you want to know if uploading grub and y2logs works.  I can't trigger host installation fail so as for it to invoke post_fail_hook.  The "autoyast/installation.pm" is maintained by yast QE. We can trust them.

grub upload example:  http://10.67.183.130/tests/3581#downloads

tonyyuan1:
> > VR with xen default boot kvm, http://10.67.183.130/tests/3567 xen, http://10.67.183.130/tests/3552
> 
> Hi Tony, I noticed the extra reboot was removed from your tests. It good. but can we manage to keep some necessary checks as we have in the current test, such as double_check_xen_role(), check_kvm_modules() and check_host_health(). And set_ssh_console_timeout_before_use() if it is not configured in autoyast.
> 
virt-install will fail if anything above get wrong.  We don't check and let virt-install check them in guest install stage in MU tests. 

The guest install run in background in Product tests, maybe some checks above are needed.  however, autoyast install are quite stable, maybe you don't need those check.

Julie-CAO:
> > > VR with xen default boot kvm, http://10.67.183.130/tests/3567 xen, http://10.67.183.130/tests/3552
> > 
> > 
> > Hi Tony, I noticed the extra reboot was removed from your tests. It good. but can we manage to keep some necessary checks as we have in the current test, such as double_check_xen_role(), check_kvm_modules() and check_host_health(). And set_ssh_console_timeout_before_use() if it is not configured in autoyast.
> 
> virt-install will fail if anything above get wrong. We don't check and let virt-install check them in guest install stage in MU tests.
> 
> The guest install run in background in Product tests, maybe some checks above are needed. however, autoyast install are quite stable, maybe you don't need those check.

These functions are not only for checking automatin stability, the main purpose is to check the system bugs or some workaround of OSD infra problems.  

check_host_health():  it is created because once there were call trace and coredumps in a fresh SLE. we need detect them before our test.
set_ssh_console_timeout_before_use(): ssh connection are broken during our tests at times. The fuction will make our test much stable. Regarless where it is emplemented, autoyast profiles or after system login, it is required.
double_check_xen_role(): it was there because we met a product bug. SLE failed to boot into xen hypervisor and went into the next advanced bootup selection or a readonly snapshot in the boot menu, but we did not notice it and took much effort to locate consequent test failures.
check_kvm_modules(): similiar to double_check_xen_role().

waynechen55:
> > Better provide verification examples which provide grub and y2logs. And update progress ticket as well poo#128375.
> 
>     1. Updated poo#128375.
> 
>     2. I guess that you want to know if uploading grub and y2logs works.  I can't trigger host installation fail so as for it to invoke post_fail_hook.  The "autoyast/installation.pm" is maintained by yast QE. We can trust them.
> 
> 
> grub upload example: http://10.67.183.130/tests/3581#downloads

I am afraid this example does not meet what is supposed to be done. Basically grub.cfg and y2logs should be uploaded when installation fails or installation succeeds but login fails, so grub.cfg and y2logs might not be available because system might be accessed. So y2logs and grub.cfg should be collected after installation is done but before reboot, this is what logs_from_installation_system.pm does in GUI installation. I think autoyast should do the same thing, otherwise it does provide equal functionality compared with before.

waynechen55:
One more thing is if newly installed system can not boot up as normal, the abnormal message should can also be printed out and captured in screenshots, for example, https://openqa.suse.de/tests/10091322#step/login_console/17.

Julie-CAO:
> One more thing is if newly installed system can not boot up as normal, the abnormal message should can also be printed out and captured in screenshots, for example, https://openqa.suse.de/tests/10091322#step/login_console/17.

I think this kind of reboot failure is supposed to be captured by `testapi::select_console("root-ssh") called at sle/lib/Utils/Backends.pm line 98`, but it is difficult to be reproduced for verifying.

tonyyuan1:
> One more thing is if newly installed system can not boot up as normal, the abnormal message should can also be printed out and captured in screenshots, for example, https://openqa.suse.de/tests/10091322#step/login_console/17.

Installation could pass but the following login could catch it.   See the testcase in the bug report.  https://bugzilla.suse.com/show_bug.cgi?id=1202641
http://openqa.qam.suse.cz/tests/45913#dependencies

In this case the login is still useful.

waynechen55:
> > One more thing is if newly installed system can not boot up as normal, the abnormal message should can also be printed out and captured in screenshots, for example, https://openqa.suse.de/tests/10091322#step/login_console/17.
> 
> I think this kind of reboot failure is supposed to be captured by `testapi::select_console("root-ssh") called at sle/lib/Utils/Backends.pm line 98`, but it is difficult to be reproduced for verifying.



> I used testcase in https://bugzilla.suse.com/show_bug.cgi?id=1202641 to show the difference of screenshot when xen boot fails.
> 
> autoyast insallation.pm: http://10.67.183.130/tests/3610#step/installation/25, it defaults to boot xen.
> 
> login_console: http://openqa.qam.suse.cz/tests/45913#step/login_console/15 , it defaults to boot normal kernel by the autoyast but login_console reboot to xen so login_console catch it.
> 
> Which one do you want?



> I used testcase in https://bugzilla.suse.com/show_bug.cgi?id=1202641 to show the difference of screenshot when xen boot fails.
> 
> autoyast insallation.pm: http://10.67.183.130/tests/3610#step/installation/25, it defaults to boot xen.
> 
> login_console: http://openqa.qam.suse.cz/tests/45913#step/login_console/15 , it defaults to boot normal kernel by the autoyast but login_console reboot to xen so login_console catch it.
> 
> Which one do you want?

What should be preserved is when system fails to boot up, the screenshot contains the failure should be kept. Previous GUI installation has such information, your autoyast should provide the equal functionality. The example is here: https://openqa.suse.de/tests/10091322#step/login_console/17. Otherwise, it would be difficult to know what is happening.

waynechen55:
> I used testcase in https://bugzilla.suse.com/show_bug.cgi?id=1202641 to show the difference of screenshot when xen boot fails.
> 
> autoyast insallation.pm: http://10.67.183.130/tests/3610#step/installation/25, it defaults to boot xen.
> 
> login_console: http://openqa.qam.suse.cz/tests/45913#step/login_console/15 , it defaults to boot normal kernel by the autoyast but login_console reboot to xen so login_console catch it.
> 
> Which one do you want?

What should be preserved is when system fails to boot up, the screenshot contains the failure should be kept. Previous GUI installation has such information, your autoyast should provide the equal functionality. The example is here: https://openqa.suse.de/tests/10091322#step/login_console/17. Otherwise, it would be difficult to know what is happening.

waynechen55:
@tonyyuan1 Your test run http://10.67.183.130/tests/3610#step/installation/28 does not detect linux-login prompt, so I think it will not preserver any screenshot about bootup. This is not good.
login_console like this http://openqa.qam.suse.cz/tests/45913#step/login_console/18 will detect linux-login prompt, so it preserves bootup information. This is much better.

waynechen55:
By the way, I think an example of  failed autoyast installation calling post_fail_hook is better. There should be a way to trigger a failed autoyast installation I think. So we can examine what has bee uploaded actually. I think this is part of verification work for this pull request.

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
...
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <serial>serial --speed=115200 --unit=<%= $get_var->('SERIALDEV') =~ s/ttyS//r %> --word=8 --parity=no --stop=1</serial>
+      <timeout config:type="integer">15</timeout>
+      <trusted_grub>false</trusted_grub>
+      <terminal>console</terminal>
+      % if ($check_var->('SYSTEM_ROLE', 'xen')) {
+      <xen_append>splash=silent quiet console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %> <%= defined $bmwqemu::vars{"OPT_KERNEL_PARAMS"} ? $bmwqemu::vars{"OPT_KERNEL_PARAMS"} : "" %></xen_append>
+      <xen_kernel_append>dom0_max_vcpus=4 console=com2,115200 vga=gfx-1024x768x16 loglvl=all guest_loglvl=all sync_console</xen_kernel_append>
```
Julie-CAO, line 57: Why is "dom0_max_vcpus=4" added at here? we don't have this parameters in our xen tests.

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
```diff
@@ -0,0 +1,268 @@
...
+      % } else {
+      <append>splash=silent console=<%= $get_var->('SERIALDEV') %>,115200 console=tty loglevel=5 <%= $check_var->('AMD', '1') ? "amd_iommu=on" : "intel_iommu=on" %> <%= defined $bmwqemu::vars{"OPT_KERNEL_PARAMS"} ? $bmwqemu::vars{"OPT_KERNEL_PARAMS"} : "" %></append>
+      % }
```
waynechen55, line 60: I think we'd better put ```console=tty``` before ```console=ttyS1,xxxx``` in the final autoyast file to ensure ttyS1 is always the active one.

tonyyuan1: Kernel can dump message to multiple devices at the same time.   tty is virtual console device. ttyS1 is serial device.

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
Julie-CAO: Ok, let's remove it and keep everything as default since it is not required.

alice-suse (approved):
LGTM

alice-suse:
If no further comments from anyone today, will merge it at EOB.

PR #16774, opened 2023-04-04
netconfig force update and add new name server if network failure

waynechen55:
* **Sometimes** SUT machine can not communicate with system outside, for example, the worker machine or openQA website. Root cause is file /etc/resolv.conf can not be populated successfully on reboot. This happens frequently recently due to unclear reasons, detailed info can be found in [poo#126188](https://progress.opensuse.org/issues/126188). 

* **In** order to mitigate the issue, using "netconfig update -f" to refresh network configuration after ssh console is opened.

* **If** netconfig update -f does not help, remove existing and add new name servers provided by worker machine.

* **Verification Runs:**
  * [guest installation on 12sp5](http://openqa.suse.de/tests/10917129)
  * [guest installation on 15sp5](http://openqa.suse.de/tests/10917134)
  * [guest installation on sle micro](http://openqa.suse.de/tests/10917132)
  * [guest installation on 12sp5 with empty /etc/resolv.conf and using netconfig update -f followed by adding nameserver into /etc/resolv.conf](http://openqa.suse.de/tests/10957562)
  * [guest installation on 15sp5 with empty /etc/resolv.conf and using netconfig update -f followed by adding nameserver into /etc/resolv.conf](http://openqa.suse.de/tests/10991605)
  * [guest installation on 15sp4 with empty /etc/resolv.conf and using netconfig update -f followed by adding nameserver into /etc/resolv.conf](http://openqa.suse.de/tests/10991606)
  * [guest installation on sle-micro with empty /etc/resolv.conf and using netconfig update -f followed by adding nameserver into /etc/resolv.conf](http://openqa.suse.de/tests/10918426)
  * [uefi guest installation on sles with empty /etc/resolv.conf and using netconfig update -f followed by adding nameserver into /etc/resolv.conf](http://openqa.suse.de/tests/10929535)
  * [host upgrade with empty /etc/resolv.conf and using netconfig update -f](https://openqa.suse.de/tests/10891525)
  * [add nameserver into /etc/resolv.conf on local SLES host](http://10.163.28.134/tests/1866)
  * [add nameserver into /etc/resolv.conf on remote SLES host](http://10.163.28.134/tests/1880)
  * [netconfig update -f followed by adding nameserver into /etc/resolv.conf on localhost](http://10.163.28.134/tests/1868)
  * [netconfig update -f followed by adding nameserver into /etc/resolv.conf on remote SLE-Micro hot](http://10.163.28.134/tests/1869)
  * [netconfig update -f followed by adding nameserver into /etc/resolv.conf on local SLE-Micro host](http://10.163.28.134/tests/1873)
  * [host network failure and test run die](http://10.163.28.134/tests/1876)

### DIFF
--- a/lib/Utils/Backends.pm
+++ b/lib/Utils/Backends.pm
@@ -18,6 +18,7 @@ use base 'Exporter';
 use Exporter;
 use testapi ':DEFAULT';
 use Utils::Architectures 'is_s390x';
+use Mojo::File 'path';
 
 use constant {
     BACKEND => [
@@ -49,6 +50,7 @@ use constant {
           set_ssh_console_timeout
           save_serial_console
           get_serial_console
+          update_ssh_console_netconfig
         )
     ]
 };
@@ -286,6 +288,52 @@ sub is_ssh_installation {
     my $videomode = get_var('VIDEOMODE', '');
     return ($videomode =~ /ssh/ ||
           (($videomode =~ /text/) && (is_pvm || is_s390x)));
+}
+
+=head2 update_ssh_console_netconfig
+
+Update the network configuration using current netconfig settings. This will help
+refresh file like /etc/resolv.conf, so associated machine will get the latest or
+corrected name servers to continue to function as normal, for example, establish 
+connection back to worker machine by using FQDN or communicate with system outside 
+using FQDN. If "netconfig update -f" can not help, adding functioning name servers
+into /etc/resolv.conf provided by worker machine. Record failure and even terminate 
+test run if host fails to communicate outside at the last. Arguments are address 
+of the system on which netconfig will operate and the number of retries.
+
+=cut
+
+sub update_ssh_console_netconfig {
+    my (%args) = @_;
+    $args{dest} //= 'localhost';
+    $args{retry} //= 3;
+
+    my $pingaddr = version_utils::is_opensuse() ? 'openqa.opensuse.org' : 'openqa.suse.de';
+    my $sshcmd = $args{dest} eq 'localhost' ? "" : "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root\@$args{dest}";
+    my $pingret = script_run("$sshcmd timeout --kill-after=1 --signal=9 20 ping -c5 $pingaddr");
+    while ($args{retry} > 0 and $pingret != 0) {
+        script_run("$sshcmd timeout --kill-after=1 --signal=9 20 netconfig update -f");
+        $pingret = script_run("$sshcmd timeout --kill-after=1 --signal=9 20 ping -c5 $pingaddr");
+        $args{retry} -= 1;
+    }
+    save_screenshot;
+    if ($pingret != 0) {
+        my $resolv_file = "/etc/resolv.conf";
+        assert_script_run("$sshcmd sed -i -r \'/^nameserver.*\$/d\' $resolv_file") if (script_run("$sshcmd cat $resolv_file | grep \"^nameserver.*\$\"") == 0);
+        my $nameserver = join '', map { "$_\\n" } grep { /^nameserver/ } split("\n", path($resolv_file)->slurp);
+        $args{dest} eq 'localhost' ? assert_script_run("sed -i -r \'\$a $nameserver\' $resolv_file") : assert_script_run("$sshcmd \"sed -i -r \'\\\$a $nameserver\' $resolv_file\"");
+        $pingret = script_run("$sshcmd timeout --kill-after=1 --signal=9 20 ping -c5 $pingaddr");
+    }
+    save_screenshot;
+    my $outputcmd = "cat /etc/resolv.conf;cat /etc/netconfig;ip neighbor;ip addr show;ip route show all";
+    $outputcmd = "$sshcmd \"$outputcmd\"" if ($args{dest} ne 'localhost');
+    if ($pingret != 0) {
+        record_info("$args{dest} network failure", script_output($outputcmd, type_command => 1, proceed_on_failure => 1), result => 'fail');
+        die "Test run can not continue due to severe communication failure" if ($args{dest} eq 'localhost');
+    }
+    else {
+        record_info("$args{dest} network intact", script_output($outputcmd, type_command => 1, proceed_on_failure => 1));
+    }
 }
 
 1;
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -13,10 +13,11 @@ use warnings;
 use File::Basename;
 use testapi;
 use Utils::Architectures;
-use Utils::Backends qw(use_ssh_serial_console is_remote_backend set_ssh_console_timeout);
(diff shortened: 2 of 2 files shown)


waynechen55:
@alice-suse @Julie-CAO @guoxuguang @nanzhg @varunkojha Please have a look.

Julie-CAO:
LGTM

tests/virt_autotest/login_console.pm
```diff
@@ -209,6 +210,7 @@ sub login_to_console {
     # double-check xen role for xen host
     double_check_xen_role if (is_xen_host and !get_var('REBOOT_AFTER_UPGRADE'));
     check_kvm_modules if is_x86_64 and is_kvm_host and !get_var('REBOOT_AFTER_UPGRADE');
+    update_ssh_console_netconfig if (is_x86_64 and (is_sle or is_sle_micro) and get_var('VIRT_AUTOTEST', ''));
```
alice-suse, line 213: If this is going to mitigate some of the connection issues to worker machines(eg granache-1) by DNS problem(or DHCP server misbehave by lack of DNS server info), will it help to add it on more test modules? I saw from your mentioned job examples, it does not always happen on the same module, eg some on update_package, while some on host_upgrade_step2_run.

waynechen55: host_upgrade_step2_run failed because it already failed in login_console after reboot, the failures can only be seen in some other modules because only those modules try to communicate outside by using FQDN. After fixing name server after reboot, there is no need to do this anymore in other modules.

alice-suse: Good finding!

alice-suse:
@okurz  Maybe you are interested to have a look?

lib/Utils/Backends.pm
```diff
@@ -288,4 +289,58 @@ sub is_ssh_installation {
...
+        assert_script_run("$sshcmd sed -i -r \'/^nameserver.*\$/d\' $resolv_file") if (script_run("$sshcmd cat $resolv_file | grep \"^nameserver.*\$\"") == 0);
+        my $nameserver = "";
+        open my $fh, '<', $resolv_file or die "Unable to open $resolv_file for input";
+        while (my $line = <$fh>) {
+            if ($line =~ /^nameserver.*$/) {
+                my $temp = $line;
+                $temp =~ s/\n//g;
+                $nameserver .= "\\n$temp";
+            }
+        }
```
okurz, line 330: Can you explain what you want to do here as I don't fully understand it. Regarding reading/writing files in Perl I suggest to use https://docs.mojolicious.org/Mojo/File#slurp but maybe we can further simplify here

waynechen55: It performs:
1. open /etc/resolv.conf file on worker machine. Main content of this file is as below:
```
### /etc/resolv.conf is a symlink to /run/netconfig/resolv.conf
### autogenerated by netconfig!
#
# Before you change this file manually, consider to define the
# static DNS configuration using the following variables in the
# /etc/sysconfig/network/config file:
#     NETCONFIG_DNS_STATIC_SEARCHLIST
#     NETCONFIG_DNS_STATIC_SERVERS
#     NETCONFIG_DNS_FORWARDER
# or disable DNS configuration updates via netconfig by setting:
#     NETCONFIG_DNS_POLICY=''
#
# See also the netconfig(8) manual page and other documentation.
#
### Call "netconfig update -f" to force adjusting of /etc/resolv.conf.
search yyyyy zzzzz
nameserver xxxxx
nameserver xxxx

```
2. Iterate over the whole file line by line
3. If a line matches pattern ```^nameserver.*$```, then record its value to ```$temp``` and remove ```\n```
4. Append each matched line to ```$nameserver``` which takes the form of a single string ```\nnameserver\nnameserver2\nnameserver3```
5. ```$nameserver``` will be used in ```sed``` command to be appended to /etc/resolv.conf on SUT machine.

waynechen55: So I think I can not use API to read all lines or bytes all together.

okurz: I still assume there is a way to make this look more readable. Anyone from @os-autoinst/tools-team has an idea?

lib/Utils/Backends.pm
```diff
@@ -288,4 +289,58 @@ sub is_ssh_installation {
...
+        assert_script_run("$sshcmd sed -i -r \'/^nameserver.*\$/d\' $resolv_file") if (script_run("$sshcmd cat $resolv_file | grep \"^nameserver.*\$\"") == 0);
+        my $nameserver = "";
+        open my $fh, '<', $resolv_file or die "Unable to open $resolv_file for input";
+        while (my $line = <$fh>) {
+            if ($line =~ /^nameserver.*$/) {
+                my $temp = $line;
+                $temp =~ s/\n//g;
+                $nameserver .= "$temp\\n";
+            }
+        }
```
kraih, line 330: With a little less code:
```perl
for my $line (split "\n", path($resolve_file)->slurp) {
  next unless $line =~ /^nameserver/;
  ...
}
```

kraih: Or as a one-liner, but i find this less readable:
```perl
my $nameservers = join '', map { "$_\\n" } grep { /^nameserver/ } split "\n", path($resolve_file)->slurp;
```

lib/Utils/Backends.pm
Martchus: One could of course use `for my $line (split("\n", path($resolv_file)->slurp)) { … }`. That's likely preferable. Note that your code lacks a `close` call (if you use `Mojo::File` you don't need to care about it).

lib/Utils/Backends.pm
waynechen55: Done. Use this one-liner.

lib/Utils/Backends.pm
waynechen55: Done. Use one-liner.

okurz:
If this is only about broken DHCP for machines within FC Basement then maybe this is not needed anymore due to fixes in the infrastructure, see https://progress.opensuse.org/issues/126188#note-29

waynechen55:
> If this is only about broken DHCP for machines within FC Basement then maybe this is not needed anymore due to fixes in the infrastructure, see https://progress.opensuse.org/issues/126188#note-29

I think this might be still useful. It gives extra information and try to use netconfig to refresh config.

kalikiana (approved):
I second Wayne here. This will probably prove valuable even if one particular case was dealt with now, and it will make the issue clearer and easier to debug.

alice-suse:
I tend to hold on the merge.
-  without the fix for some time, we can verify if the infra fix really fix the issue. 
    - If yes, this PR won't be necessary. Without it, we will be relying what DHCP server offers, which is clean network solution without patches , and thus be able to expose any new real network problems if any in future
    - If no, let's merge it to workaround the problem. Better to revert once the problem is really fixed.

So I take it as, able to mitigate certain network problems, but meanwhile hide some potential problems. I would prefer enabling it when we know what explicitly it is mitigating, rather than always having it.

waynechen55:
Closing....

PR #16778, opened 2023-04-04
Execute pattern tests in SLE Micro only

jlausuch:


Vogtinator (changes requested):
Wrong test?

DimStar77:
That's not the module that's failing... 

it's microos/patterns that fails:

Vogtinator:
You probably want to change line 194 instead

Vogtinator (approved):

PR #16781, opened 2023-04-04
Only test year_2038_detection on ALP

rfan1:
We may not able set right permission for serial console with 'ROOTONLY' system. so right now I will only schedule the test on ALP.
I will try to find a way to enable the tests on Micro-OS later.

- Verification run: https://openqa.opensuse.org/tests/3208912

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -331,7 +331,7 @@ sub load_tests {
         load_common_tests;
         load_transactional_tests unless is_zvm;
     }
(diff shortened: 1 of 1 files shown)


DimStar77 (approved):
ok as a workaround - but please make sure to get this test in TW/MicroOS rather sooner than later (factory first)

rfan1:
> ok as a workaround - but please make sure to get this test in TW/MicroOS rather sooner than later (factory first)

Sure, I will file a new ticket and enable it.

PR #16783, opened 2023-04-04
[WIP] fixfiles: fix issue when path contains spaces

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/126812
- Verification run: ..


tjyrinki (approved):
Not sure whether the word "beautiful!" is the right word to describe :), but regardless a very well found out and fixed!

DimStar77:
Rerunning snapshot 0410 which passed yesterday shows that it fails at the same place now.

PR #16785, opened 2023-04-04
Check that Azure blob md5 field is present

pdostal:
This continues the effort of b5cea54643aada0d0ca134e3a6c044a9da840ea7 (#16770).

- Verification run: [15-SP5-Azure-BYOS](https://pdostal-server.suse.cz/t4263), [15-SP5-EC2-ARM](https://pdostal-server.suse.cz/t4264)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -67,8 +67,27 @@ sub find_img {
     my $sku = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';
     $name =~ s/\.xz$//;
     $name =~ s/\.vhdfixed$/-$sku.vhd/;
+
+    my $container = $self->container;
+    my $storage_account = get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'eisleqaopenqa');
+    my $key = $self->get_storage_account_keys($storage_account);
+    my $cmd_show = "az storage blob show --account-key $key -o json " .
+      "--container-name '" . $container . "' --account-name '$storage_account' --name $name " .
+      '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"';
+    record_info('BLOB INFO', script_output($cmd_show));    # This is temporal debug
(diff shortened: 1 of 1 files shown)


lib/publiccloud/azure.pm
```diff
@@ -58,6 +58,7 @@ sub get_image_id {
     return $self->SUPER::get_image_id($img_url);
 }
 
+# asdf
```
asmorodskyi, line 61: leftover ?

lib/publiccloud/azure.pm
```diff
@@ -67,8 +68,29 @@ sub find_img {
...
+    record_info('BLOB INFO', script_output($cmd_show));    # This is temporal debug
+    my $start_time = time();
+    my $upload_time = 600;
+    my $completed = 0;
+    while (time() - $start_time < $upload_time) {
+        my $blob_info = decode_azure_json(script_output($cmd_show));
+        record_info('MD5', "Blob MD5: $blob_info->{md5}");
+        # Check that MD5 metadata exists and is not empty
+        # We don't compare the value yet, because we don't always have the downloaded.
+        $completed = 1 if ($blob_info->{md5} && length($blob_info->{md5}) > 0);
```
mpagot, line 87:
```
length($blob_info->{md5}) > 0
```

formally it is never 0, even if blob does not have the MD5, the `az storage blob show` return you the string `null` or something similar.

asmorodskyi:
generally I like it but I have doubts about this waiting loop , I have a feeling that we solving problem in wrong place . Normal there should be single job which upload image and all others should not start before it is done and we have openQA which make sure that it works like that . Let's check why we having several upload_images uploading same image in parallel ( let's discuss this in the call )

lib/publiccloud/azure.pm
pdostal: Pity that you did not raise this concern yesterday.

I cannot compare it with the md5 hash of the file. See the comment above.

The md5 hash should have always 32 characters, so please check now.

mpagot (approved):
LGTM

ricardobranco777 (approved):
LGTM

PR #16786, opened 2023-04-04
Removed waitforguestregister for BYOS images

yarunachalam:
Removed waitforguestregister for BYOS images

- Verification run: WIP


tests/publiccloud/migration.pm
```diff
@@ -27,8 +27,7 @@ sub run {
     my ($self, $args) = @_;
     select_serial_terminal();
     my $provider = $args->{my_provider};
-    my $instance = $provider->create_instance();
-    $instance->wait_for_guestregister();
+    my $instance = $provider->create_instance(check_guestregister => 0);
```
m-dati, line 30: Please why do you need to remove that check?

m-dati:
Ok, I am checking the behavior of that original issue

yarunachalam (commented):
we use only BYOS image for DMS test. we do not wait_for_guestregister

elsif ($out eq 'active') { $self->upload_log($log, log_name => $name); die "guestregister should not be active on BYOS" if (is_byos); }

m-dati:
Hi, sorry but that blocking error was Not due to the _guestregister_ check function itself: the same check worked correctly in this other [DMS-AZURE-SAP-BYOS](https://openqa.suse.de/tests/10854142/#step/migration/76) test, failed then for other reason.

Instead I fould that the issue in [10854148](https://openqa.suse.de/tests/10854148#step/migration/220) was due to an unexpected command output, caused by a previous **ssh** issue by _Identity file missing_,  that corrupted the expected checks in the loop, therefore never matching until timeout.

Infact a command `ssh -i 'identity_file'` expects that file be present, but  in the [log](https://openqa.suse.de/tests/10854148/logfile?filename=serial_terminal.txt) you will find instead all these commands (in that loop) the `-i` option has **empty ''** value:
```
timeout 90 ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i '' "ec2-user@3.142.249.196" -- 'rpm -qa cloud-regionsrv-client'
```

Anyway I will update the guestregister check to be much resilient.

PR #16788, opened 2023-04-05
Set plain text output in curl for matching on squid test

ilmanzo:
New version of `curl` has bold output by default, so to use in `validate_script_output` it needs an extra option to switch to plain text  

- Related ticket: https://progress.opensuse.org/issues/125585
- Verification runs:
  - s390x https://openqa.suse.de/tests/10856031
  - x86_64 https://openqa.suse.de/tests/10856033
  - aarch64 https://openqa.suse.de/tests/10856034


realcharmer:
LGTM

PR #16789, opened 2023-04-05
Add wait_quit_zypper in zdup

Zaoliang:
we have sporadic issue with zypper process running in background, see poo#126425, so add wait_quit_zypper can avoid this issue.

verification test run: http://10.161.8.50/tests/59


Zaoliang:
no need to change, since I cannot reproduce this issue anymore on O3

PR #16796, opened 2023-04-05
Run upstream libica's build testsuite in FIPS mode

ilmanzo:
We schedule the upstream build testsuite and a simple 'smoke' test to check libica functionality in the system. Libica is s390x only

- Related ticket: https://progress.opensuse.org/issues/120558 
- Verification run: https://openqa.suse.de/tests/10855924#


paolostivanin:
LGTM

PR #16799, opened 2023-04-05
Use default timeout for MicroOS boot in aarch64

jlausuch:
By removing it, we will be using the default one (500), which is higher than current one.
We are hitting this 300s timeout sporadically.

- Related ticket: https://progress.opensuse.org/issues/127211
- VR: http://openqa.suse.de/tests/10864238


clanig:
LGTM

PR #16807, opened 2023-04-06
Refactor spack to align with source code which is selected by get_mpi_src

b10n1k:
boost is a special case which requires LD_LIBRARY_PATH which has been removed
and spack is not setting during its installation. A workaround causes problems
with the system packages and make its fix makes the test complicated. Instead,
some other basic source code can avoid this, as it should be able to be
compiled and run without additional hacks. Although this does not decrease the
execution time. Meaning that spack still takes some time to install its
dependencies, including the openssl which was causing the issues in the
previous implementation.
    
- Verification run: https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%23spack_openssl

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -347,6 +347,24 @@ from the rest and can be exported via a network file system.
 
 After C<prepare_spack_env> run, C<spack> should be ready to build entire tool stack,
 downloading and installing all bits required for whatever package or compiler.
+
+This sub is designed to install one of the mpi implementations. Although there are
+thousands packages to be used.
+
+LD_LIBRARY_PATH is removed and spack is not exported. In case LD_LIBRARY_PATH is required
+it has to be add it in the F<.spack/modules.yaml>.
+
+=begin text
+  See bsc#1208751 for details
+=end text
+
+To do that run
+
+=begin text
+  spack config add modules:prefix_inspections:lib64:[LD_LIBRARY_PATH]
+  spack config add modules:prefix_inspections:lib:[LD_LIBRARY_PATH]
+=end text
+
 =cut
 
 sub prepare_spack_env {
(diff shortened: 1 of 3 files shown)


czerw:
I see commit message confusing, as you don't change or use any source code. It's about different environment settings
 for binaries not source code.

b10n1k:
> I see commit message confusing, as you don't change or use any source code. It's about different environment settings for binaries not source code.

i just notice i have to squish the commits. Regarding the commit, the source code is determined by a variable. I think the flow now is in such way which any source code which it doesnt have some kind of complicated requirement should work. Spack now is only install and load mpi, again based on variables

tests/hpc/spack_slave.pm
```diff
@@ -44,9 +45,10 @@ sub post_fail_hook ($self) {
     # Inside compiled modules comes as <module>-<version>-<hash>
     # Each module includes build logs under <module>-<version>-<hash>/.spack
     my $compiler_ver = script_output("gcc --version | grep -E '\\b[0-9]+\.[0-9]+\.[0-9]+\$' | awk '{print \$4}'");
+    record_info "postfail", "$compiler_ver";
     my $arch = get_var('ARCH');
     my $node = script_output('hostname');
-    tar_and_upload_log("/opt/spack/linux-sle_hpc15-$arch/gcc-$compiler_ver", "/tmp/spack_$node.tar.bz2");
+    #tar_and_upload_log("/opt/spack/linux-sle_hpc15-$arch/gcc-$compiler_ver", "/tmp/spack_$node.tar.bz2");
```
czerw, line 51: either `#` or whole line  should be removed

czerw:
> the source code is determined by a variable

This is not source code.

lib/hpcbase.pm
```diff
@@ -347,44 +347,53 @@ from the rest and can be exported via a network file system.
...
+  spack config add modules:prefix_inspections:lib64:[LD_LIBRARY_PATH]
+  spack config add modules:prefix_inspections:lib:[LD_LIBRARY_PATH]
+=end text
+
 =cut
 
 sub prepare_spack_env {
     my ($self, $mpi) = @_;
     $mpi //= 'mpich';
+    zypper_call "in libopenssl-devel", timeout => 480;
```
czerw, line 373: You can install all in one `zypper_call` together on line 384.

czerw: moreover, why it is needed? `ibopenssl-devel` is already pulled by spack as dependency

b10n1k: this is messed up from squash. I will remove it

tests/hpc/spack_slave.pm
b10n1k: i had to squash and rebase which asked me to fix some conflicts. i missed that

b10n1k: fixed and ready for review.

lib/hpcbase.pm
b10n1k: fixed and ready for review

czerw:
I really don't understand which source code you are talking about, moreover what is `get_mpi_src` and where is it used in test?

b10n1k:
> I really don't understand which source code you are talking about, moreover what is `get_mpi_src` and where is it used in test?

`get_mpi_src` returns the source code based on the HPC_LIB. it is used among all the mpi modules. This is the main change. The commit just remove the `boost` specific bits, i would say. Such as, the source can be compiled (without the previous hacks for LD_LIBRARY_PATH) and run.

PR #16812, opened 2023-04-06
[WIP] New JeOS common schedule

jlausuch:


lib/main_jeos.pm
```diff
@@ -0,0 +1,200 @@
...
+
+our @EXPORT = qw(
+  load_jeos_tests
+);
+
+sub load_jeos_boot_tests {
+    my $machine = get_required_var('MACHINE');
+
+    loadtest 'installation/bootloader_svirt' if is_svirt;
+    loadtest 'installation/bootloader_uefi' if ($machine =~ /xen-hvm|hyperv$|vmware/);
```
mloviska, line 27: Better would be to use
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/Utils/Backends.pm

lib/main_jeos.pm
```diff
@@ -0,0 +1,200 @@
...
+
+    loadtest 'jeos/firstrun';
+    loadtest 'console/verify_efi_mok' if check_var('UEFI', '1');
```
mloviska, line 30: I would load this one only in specific case.When `CHECK_MOK_IMPORT == 1`

lib/main_jeos.pm
```diff
@@ -0,0 +1,200 @@
...
+
+    if (check_var('FIRST_BOOT_CONFIG', 'cloud-init') {
+        loadtest 'installation/first_boot';
+        loadtest 'console/system_prepare';
+    } else {
+        loadtest 'jeos/firstrun';
+    }
+}
+
+sub load_common_tests {
```
mloviska, line 40: I am thinking to rename this one, something like `image_base_tests`

mloviska:
Looks very nice! Great job! :clap:

PR #16814, opened 2023-04-07
Add GitHub CODEOWNERS for Public Cloud and Containers tests

pdostal:
The [GitHub CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) file basically notifies all the people interested in particular file in case that file has been changed.

Also, if given file has code owners then at least one of them needs to approve every pull request before it can be merged.

This should be valid after we merge this pull request.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# GitHub CODEOWNERS https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+lib/publiccloud/ @pdostal @grisu48
+tests/publiccloud/ @pdostal @grisu48
+
+lib/containers/ @pdostal @grisu48 @asmorodskyi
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# GitHub CODEOWNERS https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+lib/publiccloud/ @pdostal @grisu48
```
grisu48, line 3: And what about the others? I would add at least also @asmorodskyi and anyone else who wants to.

grisu48:
Other than that, LGTM.

.github/CODEOWNERS
pdostal: It makes sense but I would like to wait for Anton.

asmorodskyi: LGTM

PR #16820, opened 2023-04-09
Do not fail if publiccloud::azure::upload_img is called for already uploaded image

pdostal:
- Previous pull requests: #16785, #16813
- Example failure: [openqa.suse.de/t10889077](https://openqa.suse.de/t10889077)
- Verification run: [pdostal-server.suse.cz/t4272](https://pdostal-server.suse.cz/t4272)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -82,7 +82,7 @@ sub find_img {
     eval {
         $image = decode_azure_json($json)->{name};
     };
-    record_info('INFO', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
     return $image;
 }
 
@@ -158,8 +158,8 @@ sub upload_img {
     $blobs =~ s/^\s+|\s+$//g;    # trim
     my @blobs = split(/\n/, $blobs);
     if (grep(/$img_name/, @blobs)) {
-        die("The upload_img() subroutine has been called even tho the $img_name blob exists. " .
-              "This means that ccloud/upload_image test module did not properly detected it.");
+        record_soft_fail('BLOB EXISTS', "The upload_img() subroutine has been called even tho the $img_name blob exists. " .
+              "This means that publiccloud/upload_image test module did not properly detect it.");
     } else {
         record_info("blobs", $blobs);
         # Note: VM images need to be a page blob type


clanig:
Thanks! LGTM.

lib/publiccloud/azure.pm
```diff
@@ -82,7 +82,7 @@ sub find_img {
     eval {
         $image = decode_azure_json($json)->{name};
     };
-    record_info('INFO', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
```
grisu48, line 85: NO NEED TO YELL!!1!

```suggestion
    record_info('Image not present', "Cannot find image $name. Need to upload it.\n$@") if ($@);
```

lib/publiccloud/azure.pm
```diff
@@ -158,8 +158,8 @@ sub upload_img {
     $blobs =~ s/^\s+|\s+$//g;    # trim
     my @blobs = split(/\n/, $blobs);
     if (grep(/$img_name/, @blobs)) {
-        die("The upload_img() subroutine has been called even tho the $img_name blob exists. " .
-              "This means that ccloud/upload_image test module did not properly detected it.");
+        record_soft_fail('BLOB EXISTS', "The upload_img() subroutine has been called even tho the $img_name blob exists. " .
```
grisu48, line 161: SAME AS ABOVE!!1!

```suggestion
        record_soft_fail('Image exists', "upload_img() called even though the $img_name blob exists.\nThis means that publiccloud/upload_image test module did not properly detect it.");
```

grisu48 (commented):
Two comments, otherwise LGTM!

lib/publiccloud/azure.pm
pdostal: It's not yelling. We use uppercase letters in those boxes.

lib/publiccloud/azure.pm
pdostal: Same as above + it's not image but blob. See line `157`: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/9a13199b5fc32abdcca93661a7a272e33a374dd7/lib/publiccloud/azure.pm#L157

lib/publiccloud/azure.pm
grisu48: Ok

lib/publiccloud/azure.pm
grisu48: Ok

grisu48 (approved):

PR #16822, opened 2023-04-10
Put timeout scale logic into YuiRestClient.pm

coolgw:
Put timeout scale logic into Wait.pm
- Related ticket: https://progress.opensuse.org/issues/125561
- Needles: na
- Verification run: https://openqa.suse.de/t10917218

### DIFF
--- a/lib/YuiRestClient/Wait.pm
+++ b/lib/YuiRestClient/Wait.pm
@@ -9,6 +9,8 @@ use DateTime;
 sub wait_until {
     my (%args) = @_;
     $args{timeout} //= 10;
+    $args{timeout_scale} //= 1;
+    $args{timeout} *= $args{timeout_scale};
     $args{interval} //= 1;
     $args{message} //= '';
 
--- a/tests/installation/product_selection/install_SLES.pm
+++ b/tests/installation/product_selection/install_SLES.pm
@@ -12,7 +12,8 @@ use warnings;
 sub run {
     my $product_selection = $testapi::distri->get_product_selection();
 
-    $product_selection->wait_product_selection_page({timeout_scale => get_var('TIMEOUT_SCALE', 1), message => 'Product Selection page is not displayed'});
+    $product_selection->wait_for_product_selection_page({timeout_scale => get_var('TIMEOUT_SCALE', 1), message => 'Product Selection page is not displayed'});
+
     $product_selection->install_product('SLES');
 }
 
--- a/lib/Installation/PerformingInstallation/PerformingInstallationController.pm
+++ b/lib/Installation/PerformingInstallation/PerformingInstallationController.pm
@@ -51,7 +51,7 @@ sub get_system_reboot_with_timeout_popup {
     return $self->{OKStopPopup};
 }
 
-sub wait_installation_popup {
+sub wait_for_installation_popup {
     my ($self, $args) = @_;
     YuiRestClient::Wait::wait_until(object => sub {
             $self->{AbstractOKPopup}->is_shown({timeout => 0});
--- a/lib/Installation/ProductSelection/ProductSelectionController.pm
+++ b/lib/Installation/ProductSelection/ProductSelectionController.pm
@@ -32,9 +32,9 @@ sub get_product_selection_page {
     return $self->{ProductSelectionPage};
 }
 
-sub wait_product_selection_page {
+sub wait_for_product_selection_page {
     my ($self, $args) = @_;
-    $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
+    $args->{timeout} = $args->{timeout} // YuiRestClient::get_timeout();
     YuiRestClient::Wait::wait_until(object => sub {
             $self->{ProductSelectionPage}->is_shown({timeout => 0});
     }, %$args);
--- a/tests/installation/performing_installation/perform_installation.pm
+++ b/tests/installation/performing_installation/perform_installation.pm
@@ -19,9 +19,9 @@ sub run {
 
     $performing_install->get_performing_installation_page();
 
-    my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
-    $performing_install->wait_installation_popup({
-            timeout => $timeout,
+    $performing_install->wait_for_installation_popup({
+            timeout => 2400,
+            timeout_scale => get_var('TIMEOUT_SCALE', 1),
             interval => 2,
             message => 'System reboot popup did not appear'});
 


lib/YuiRestClient.pm
```diff
@@ -78,6 +78,7 @@ sub set_timeout {
 sub init_app {
     my (%args) = @_;
     my $timeout = $args{timeout} || TIMEOUT;
+    $timeout *= get_var('TIMEOUT_SCALE', 1);
```
jknphy, line 81: the timeout scale is applicable to specific cases, we don't need for now a general one at the level of the app, otherwise we would bump timeout in too many places, hiding bugs.

jknphy:
the whole idea is to see if we can somehow not repeat this code in different places:
```
sub wait_product_selection_page {
    my ($self, $args) = @_;
    $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
            $self->{ProductSelectionPage}->is_shown({timeout => 0});
    }, %$args);
}
```
In perform_installation we have:
```
my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
     $performing_install->wait_installation_popup({
             timeout => $timeout,
             interval => 2,
             message => 'System reboot popup did not appear'});
```
the method could be call `wait_for` from the test.
the multiplying operation should be in `wait_until`, so we already have our place for encapsulation.
In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.

lib/YuiRestClient.pm
coolgw: Done

coolgw:
> the whole idea is to see if we can somehow not repeat this code in different places:
> 
> ```
> sub wait_product_selection_page {
>     my ($self, $args) = @_;
>     $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
>     YuiRestClient::Wait::wait_until(object => sub {
>             $self->{ProductSelectionPage}->is_shown({timeout => 0});
>     }, %$args);
> }
> ```
> 
> In perform_installation we have:
> 
> ```
> my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
>      $performing_install->wait_installation_popup({
>              timeout => $timeout,
>              interval => 2,
>              message => 'System reboot popup did not appear'});
> ```
> 
> the method could be call `wait_for` from the test accepting the necessary parameters. the multiplying operation should be in `wait_until`, so we already have our place for encapsulation. In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.

I have change the function name to wait_for with necessary page parameter. Could you help review it?
Currently if you want no duplicate wait_for function in control page, then we need add base level such as YaST::ControlPageBase, then every control page will use this as parent, what do you think?

lib/YuiRestClient/Wait.pm
```diff
@@ -3,12 +3,13 @@
...
 use strict;
 use warnings;
-
+use testapi;
 use DateTime;
 
 sub wait_until {
     my (%args) = @_;
     $args{timeout} //= 10;
+    $args{timeout} *= get_var('TIMEOUT_SCALE', 1);
```
jknphy, line 12: remember that we don't use "in general" the testapi in the framework, other than in the Test Layer, so you need to introduce a new parameter and pass it from test layer to this encapsulate component which shouldn't know anything about `get_var` function or other testapi dependencies, so it is decoupled.

tests/installation/product_selection/install_SLES.pm
```diff
@@ -12,7 +12,7 @@ use warnings;
 sub run {
     my $product_selection = $testapi::distri->get_product_selection();
 
-    $product_selection->wait_product_selection_page({timeout_scale => get_var('TIMEOUT_SCALE', 1), message => 'Product Selection page is not displayed'});
+    $product_selection->wait_for({message => 'Product Selection page is not displayed', waitpage => 'ProductSelectionPage', timeout => 10});
```
jknphy, line 15: we cannot pass the page name here, we are coupling the test layer with the specific implementation, we just pass the timeout and the timeout_scale.

jknphy:
> > the whole idea is to see if we can somehow not repeat this code in different places:
> > ```
> > sub wait_product_selection_page {
> >     my ($self, $args) = @_;
> >     $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
> >     YuiRestClient::Wait::wait_until(object => sub {
> >             $self->{ProductSelectionPage}->is_shown({timeout => 0});
> >     }, %$args);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In perform_installation we have:
> > ```
> > my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
> >      $performing_install->wait_installation_popup({
> >              timeout => $timeout,
> >              interval => 2,
> >              message => 'System reboot popup did not appear'});
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the method could be call `wait_for` from the test accepting the necessary parameters. the multiplying operation should be in `wait_until`, so we already have our place for encapsulation. In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.
> 
> I have change the function name to wait_for with necessary page parameter. Could you help review it? Currently if you want no duplicate wait_for function in control page, then we need add base level such as YaST::ControlPageBase, then every control page will use this as parent, what do you think?

As we don't even want to use Controllers and I don't find better solution than what you mentioned for now, let's postpone this base class creation, so in this ticket we only cover to add timeout_scale to wait_until.
We could rename all those methods I mentioned above to `wait_for` keeping specific page check, no need to pass as parameter. wdyt?

jknphy:
> > > > the whole idea is to see if we can somehow not repeat this code in different places:
> > > > ```
> > > > sub wait_product_selection_page {
> > > >     my ($self, $args) = @_;
> > > >     $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
> > > >     YuiRestClient::Wait::wait_until(object => sub {
> > > >             $self->{ProductSelectionPage}->is_shown({timeout => 0});
> > > >     }, %$args);
> > > > }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > In perform_installation we have:
> > > > ```
> > > > my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
> > > >      $performing_install->wait_installation_popup({
> > > >              timeout => $timeout,
> > > >              interval => 2,
> > > >              message => 'System reboot popup did not appear'});
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > the method could be call `wait_for` from the test accepting the necessary parameters. the multiplying operation should be in `wait_until`, so we already have our place for encapsulation. In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.
> > > 
> > > 
> > > I have change the function name to wait_for with necessary page parameter. Could you help review it? Currently if you want no duplicate wait_for function in control page, then we need add base level such as YaST::ControlPageBase, then every control page will use this as parent, what do you think?
> > 
> > 
> > As we don't even want to use Controllers and I don't find better solution than what you mentioned for now, let's postpone this base class creation, so in this ticket we only cover to add timeout_scale to wait_until. We could rename all those methods I mentioned above to `wait_for` keeping specific page check, no need to pass as parameter. wdyt?
> 
> If you need generic function name wait_for but without any page name as parameter, then in test code level view will make confusing. Further more, if test code need wait_for another specific page, how to name the function?

in current model we need to call it wait_for_blablab, you're right, without controllers life would be easier, because controllers might have different pages to wait for... with pages will be 1:1.

coolgw:
> > > > > the whole idea is to see if we can somehow not repeat this code in different places:
> > > > > ```
> > > > > sub wait_product_selection_page {
> > > > >     my ($self, $args) = @_;
> > > > >     $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
> > > > >     YuiRestClient::Wait::wait_until(object => sub {
> > > > >             $self->{ProductSelectionPage}->is_shown({timeout => 0});
> > > > >     }, %$args);
> > > > > }
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > In perform_installation we have:
> > > > > ```
> > > > > my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
> > > > >      $performing_install->wait_installation_popup({
> > > > >              timeout => $timeout,
> > > > >              interval => 2,
> > > > >              message => 'System reboot popup did not appear'});
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > the method could be call `wait_for` from the test accepting the necessary parameters. the multiplying operation should be in `wait_until`, so we already have our place for encapsulation. In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.
> > > > 
> > > > 
> > > > I have change the function name to wait_for with necessary page parameter. Could you help review it? Currently if you want no duplicate wait_for function in control page, then we need add base level such as YaST::ControlPageBase, then every control page will use this as parent, what do you think?
> > > 
> > > 
> > > As we don't even want to use Controllers and I don't find better solution than what you mentioned for now, let's postpone this base class creation, so in this ticket we only cover to add timeout_scale to wait_until. We could rename all those methods I mentioned above to `wait_for` keeping specific page check, no need to pass as parameter. wdyt?
> > 
> > 
> > If you need generic function name wait_for but without any page name as parameter, then in test code level view will make confusing. Further more, if test code need wait_for another specific page, how to name the function?
> 
> in current model we need to call it wait_for_blablab, you're right, without controllers life would be easier, because controllers might have different pages to wait for... with pages will be 1:1.

change for name to wait_for_blablab done

coolgw:
> the whole idea is to see if we can somehow not repeat this code in different places:
> 
> ```
> sub wait_product_selection_page {
>     my ($self, $args) = @_;
>     $args->{timeout} = YuiRestClient::get_timeout() * $args->{timeout_scale};
>     YuiRestClient::Wait::wait_until(object => sub {
>             $self->{ProductSelectionPage}->is_shown({timeout => 0});
>     }, %$args);
> }
> ```
> 
> In perform_installation we have:
> 
> ```
> my $timeout = 2400 * get_var('TIMEOUT_SCALE', 1);
>      $performing_install->wait_installation_popup({
>              timeout => $timeout,
>              interval => 2,
>              message => 'System reboot popup did not appear'});
> ```
> 
> the method could be call `wait_for` from the test accepting the necessary parameters. the multiplying operation should be in `wait_until`, so we already have our place for encapsulation. In summary we should use the same interface (method) in both places and unify this, hope that clarify better the task.

done

PR #16823, opened 2023-04-11
Unify qam-yast_self_update+15@uefi for sle15sp3 and sle15sp4 using single test module for registration

tinawang123:
Unify qam-yast_self_update+15@uefi for sle15sp3 and sle15sp4 using single test module for registration

- Related ticket: https://progress.opensuse.org/issues/123076
- Verification run: https://openqa.suse.de/tests/11029912#step/register_extensions_and_modules/3
http://openqa.suse.de/tests/11040959#step/register_extensions_and_modules/2


schedule/yast/qam-yast_self_update.yaml
```diff
@@ -32,13 +31,6 @@ schedule:
   - installation/first_boot
   - '{{efi}}'
 conditional_schedule:
-  sles_product:
-    VERSION:
-      15-SP3:
-        - installation/module_registration/add_ltss_regcode
-      15-SP4:
-        - installation/module_registration/add_additional_regcodes
-        - installation/module_registration/import_untrusted_gnpupg_key
```
jknphy, line 41: looks like we are missing the deletion of this now unused files, right?

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,32 @@
...
+    foreach my $addon (@scc_addons) {
+        if ($addon eq 'we' && get_var('SCC_REGCODE_WE', '')) {
+            my $regcode = get_var('SCC_REGCODE_WE');
+            $testapi::distri->get_module_regcode()->add_separate_we_registration_code($regcode, $timeout);
+            save_screenshot;
+            $testapi::distri->get_module_regcode()->trust_gnupg_key();
+        }
+        if ($addon eq 'ha' && get_var('SCC_REGCODE_HA', '')) {
+            my $regcode = get_var('SCC_REGCODE_HA');
+            $testapi::distri->get_module_regcode()->add_separate_ha_registration_code($regcode, $timeout);
```
jknphy, line 31: we shouldn't need now to pass any timeout if we wait before explicitly.
actually with don't process that parameter `$timeout` at all.

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,32 @@
...
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    my $timeout = 60 * get_var('TIMEOUT_SCALE', 1);
+    $testapi::distri->wait_registration_common_regcode_finished({timeout => $timeout, interval => 2, message => 'Page to insert module registration code did not appear'});
+    foreach my $addon (@scc_addons) {
+        if ($addon eq 'we' && get_var('SCC_REGCODE_WE', '')) {
```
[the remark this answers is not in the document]
jknphy: it would fail anyway, it is good practice, but it is rare settings are not there, means that the configuration is completely mess up and the test might not need to care for such a bad situation, but more for testing functionality not infrastructure/test configuration.

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,32 @@
...
+        }
+        if ($addon eq 'ha' && get_var('SCC_REGCODE_HA', '')) {
+            my $regcode = get_var('SCC_REGCODE_HA');
+            $testapi::distri->get_module_regcode()->add_separate_ha_registration_code($regcode, $timeout);
+            save_screenshot;
+        }
+        if ($addon eq 'ltss' && get_var('SCC_REGCODE_LTSS', '')) {
+            my $regcode = get_var('SCC_REGCODE_LTSS');
+            $testapi::distri->get_module_regcode()->add_separate_ltss_registration_code($regcode, $timeout);
+            save_screenshot;
```
jknphy, line 37: you don't need the final screenshot at the end of previous two `if` because it is taken by the test module in post_run_hook.

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,32 @@
...
-use testapi qw(save_screenshot get_required_var);
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    my $timeout = 60 * get_var('TIMEOUT_SCALE', 1);
+    $testapi::distri->wait_registration_common_regcode_finished({timeout => $timeout, interval => 2, message => 'Page to insert module registration code did not appear'});
+    foreach my $addon (@scc_addons) {
```
jknphy, line 22: please, try to user high-order functions, not loops, specially not in testing layer. For example:
```
my @scc_addons = ('a', 'b', 'we');
my ($addon) = grep { /we|ha|ltss/ } @scc_addons;
print "SCC_REGCODE_" . uc $addon . "\n";
```

jknphy: we can unify those 3 methods `add_separate_*_registration_code` using a regex when we try to search it with libyui-rest-api at the level of Page.

jknphy: then we pass as parameter the particular addon and the controller method will check with a grep, for example if is 'we' so it would trigger at the end `get_module_regcode()->trust_gnupg_key()` because the other two methods are exactly the same. In that way in theory we could avoid to use any loop.

tinawang123: > please, try to user high-order functions, not loops, specially not in testing layer. For example:
> 
> ```
> my @scc_addons = ('a', 'b', 'we');
> my ($addon) = grep { /we|ha|ltss/ } @scc_addons;
> print "SCC_REGCODE_" . uc $addon . "\n";
> ```

But if our case need both ha and ltss.  This method cannot handle this situation.

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,19 @@
...
 use strict;
 use warnings;
-use testapi qw(save_screenshot get_required_var);
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    $testapi::distri->wait_registration_common_regcode_finished({timeout => 60, interval => 2, message => 'Page to insert module registration code did not appear'});
```
jknphy, line 20: should be call `wait_for_separate_regcode`
`common` or `finished` do not add more information.
wdyt?

lib/Installation/ModuleRegistration/SeparateRegCodesController.pm
```diff
@@ -46,22 +46,11 @@ sub wait_registration_common_regcode_finished {
...
-}
-
-sub add_separate_ltss_registration_code {
-    my ($self, $regcode) = @_;
-    $self->get_module_regcode_page()->set_ltss_regcode($regcode);
+sub add_separate_registration_code {
+    my ($self, $addon, $regcode) = @_;
+    $self->get_module_regcode_page()->set_regcode($addon, $regcode);
     $self->get_module_regcode_page()->press_next();
+    $self->trust_gnupg_key() if ($addon eq 'we');
```
jknphy, line 53: thinking twice, doing this here is weird, it is part of the test, conditional part but not something hidden, we have done wrong this in the past, you will see many example of that in expert partitioner. Let's do it at the level of test (sorry for confusing you here).

tests/installation/module_registration/register_extensions_and_modules.pm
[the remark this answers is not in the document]
jknphy: you can separate each parameter in a line, `timeout`, `interval`, etc, to make lines a bit more shorter and readable, or at least `message` should be in a different line because it is too long, (just styling, neatpicking :) )

schedule/yast/qam-yast_self_update.yaml
jknphy: and tests/installation/module_registration/import_untrusted_gnpupg_key.pm ?

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,24 @@
...
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    $testapi::distri->wait_for_separate_regcode({
+            timeout => 60,
+            interval => 2,
+            message => 'Page to insert module registration code did not appear'});
+    foreach my $addon (@scc_addons) {
+        if ($addon =~ /we|ha|ltss/ && get_required_var('SCC_REGCODE_' . uc $addon)) {
+            my $regcode = get_var('SCC_REGCODE_' . uc $addon);
```
jknphy, line 26: we don't need to do this operation twice to get the setting, it is not necessary to test the infrastructure, you can remove it from the `if` condition and we assume it is there

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,24 @@
...
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    $testapi::distri->wait_for_separate_regcode({
+            timeout => 60,
+            interval => 2,
+            message => 'Page to insert module registration code did not appear'});
```
jknphy, line 23: s/module registration/separate registration/

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,24 @@
...
 use strict;
 use warnings;
-use testapi qw(save_screenshot get_required_var);
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
+    $testapi::distri->wait_for_separate_regcode({
```
jknphy, line 20: you need to check if the addon is in that list (with a `grep` for example, before running this expensive action, that we want to run just once, then use that result in line 25 below.

tests/installation/module_registration/register_extensions_and_modules.pm
```diff
@@ -11,12 +11,24 @@
 use base 'y2_installbase';
 use strict;
 use warnings;
-use testapi qw(save_screenshot get_required_var);
+use testapi qw(save_screenshot get_required_var get_var);
 
 sub run {
     my @scc_addons = grep($_, split(/,/, get_required_var('SCC_ADDONS')));
     $testapi::distri->get_module_registration()->register_extension_and_modules([@scc_addons]);
     save_screenshot;
```
jknphy, line 19: neat: you could use a extra blank line between logical section of the code, in general after save_screenshot as a way to style your code in general. It is much clear to read.

tests/installation/module_registration/register_extensions_and_modules.pm
tinawang123: After we get the registration page, we can input the register code. 
If don't have the page, why do we need check if the addons in the list.
The logic is, first we get the registration page, when check which addons need input the register code.
The page should be checked first.

tinawang123: After we get the registration page, we can input the register code. 
If don't have the page, why do we need check if the addons in the list.
The logic is, first we get the registration page, when check which addons need input the register code.
The page should be checked first.

jknphy: this test module is supposed to handle when there are special addons that have additional regcode and also when it is not the case, for the name that is what is does, looks generic,
But if we only use it when there are separate regcode the name needs to be changed and there is not need for a condition.

In other words, we shouldn't wait for that screen if we don't have those special addons, that is why I suggested to move code.

schedule/yast/qam-yast_self_update.yaml
tinawang123: As 15-SP4 has 'we' addon, so it need import_untrusted_gnpupg_key.
At 15-SP3, the 'we' addon has expired.

tinawang123:
Any comments?

schedule/yast/qam-yast_self_update.yaml
jknphy: are the 3 test modules above used somewhere else? if not, we need to clean up.

lib/Installation/ModuleRegistration/SeparateRegCodesController.pm
jknphy: this is not resolved, but we can do later.

tests/installation/module_registration/register_extensions_and_modules.pm
jknphy: Not solved, the test module names doesn't distinguish between the two cases, but we can do that later.

tests/installation/module_registration/register_extensions_and_modules.pm
jknphy: save_screenshot goes after the logical section, then you can introduce a blank line to separate logical section:
- logical section
- save_screenshot
- blank line
The code doesn't follow that, but it is just a suggestion, we can follow in the future.

PR #16825, opened 2023-04-11
[WIP] Add liburing testing suite

acerv:
WIP to test liburing testing suite in openQA.


tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,141 @@
...
+                    $version = "liburing-$+{version}";
+                    last;
+                }
+            }
+        }
+    }
+
+    # install dependences
+    zypper_call("in git-core");
+    zypper_call("in -t pattern devel_basis");
```
czerw, line 46: You can merge those calls into one and some increased timeout.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,141 @@
...
+    my $repository = get_var('LIBURING_REPO', 'https://github.com/axboe/liburing.git');
+    my $timeout = get_var('LIBURING_TIMEOUT', 1800);
+    my $version = get_var('LIBURING_VERSION', '');
+    my $exclude = get_var('LIBURING_EXCLUDE', '');
+    my $issues = get_var('LIBURING_KNOWNISSUES', '');
+    my @lines;
+
+    # select latest liburing version which is supported by the system
+    if ($version eq '') {
+        my $pkginfo = script_output("zypper --no-refresh info liburing2");
```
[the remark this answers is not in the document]
acerv: This is a bit tricky indeed. We need to test as much functionalities as possible with liburing2 testing suite, but we don't know what kernel io_uring patches or features we have. For this reason, code always takes the latest library available version, which is supposed to test **enough** functionalities.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
...
+                }
+            }
+        }
+    }
+
+    # install dependences
+    zypper_call("in -t pattern devel_basis git-core");
+
+    # download and compile tests
+    assert_script_run("git clone $repository");
```
czerw, line 48: `git clone --depth 1 --no-single-branch ` will be better

tests/kernel/io_uring.pm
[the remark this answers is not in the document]
pevik: Yes, `rpm -qf liburing2` is what we want. Also, I'd use put info into `record_info()` so that it stays even logs are removed from openQA.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Executes liburing testing suite
+# Maintainer: Andrea Cervesato <andrea.cervesato@suse.com>
```
czerw, line 7: Please use: `# Maintainer: Kernel QE <kernel-qa@suse.de>`

tests/kernel/io_uring.pm
acerv: We need the whole tree to provide checkout support and select different versions provided by LIBURING_VERSION variable, so unfortunately we can't use ``depth 1`` but maybe we can use ``--single-branch`` to download ``master`` only?

tests/kernel/io_uring.pm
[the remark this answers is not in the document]
czerw: How it will handle following case?
```
Information for package liburing2:
----------------------------------
Repository     : Main Repository (OSS)
Name           : liburing2
Version        : 2.3-2.2
Arch           : x86_64
Vendor         : openSUSE
Installed Size : 66.6 KiB
Installed      : Yes (automatically)
Status         : out-of-date (version 2.3-2.1 installed)
Source package : liburing-2.3-2.2.src
```
Do you see it?

```
rpm -q liburing2
liburing2-2.3-2.1.x86_64
```

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,141 @@
...
+    # select latest liburing version which is supported by the system
+    if ($version eq '') {
+        my $pkginfo = script_output("zypper --no-refresh info liburing2");
+
+        foreach my $line (split /\n/, $pkginfo) {
+            if ($line =~ "Version") {
+                my @match = split /:/, $line;
+                if ($match[1] =~ '(?P<version>2.\d+)-.*') {
+                    $version = "liburing-$+{version}";
+                    last;
```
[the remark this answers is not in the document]
pevik: I'd slightly prefer `-qi` over `-q` and put that into 

Or something like: `rpm -qi liburing2| grep Version | awk '{print $3}' # returns 2.1`.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
...
+use utils;
+
+sub run {
+    my $self = shift;
+
+    select_serial_terminal;
+
+    my $repository = get_var('LIBURING_REPO', 'https://github.com/axboe/liburing.git');
+    my $timeout = get_var('LIBURING_TIMEOUT', 1800);
+    my $version = get_var('LIBURING_VERSION', '');
```
pevik, line 24: Without the default `''` you can later have just `if ($version)`.

tests/kernel/io_uring.pm
czerw: `--no-single-branch` will make sure that all branches are get (i just verified it)

tests/kernel/io_uring.pm
czerw: You are right, not possible to merge patterns with pkg.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
...
+    # search for failed tests and known issues
+    @lines = $out =~ /Tests failed \(\d+\):.*/mg;
+    if (@lines) {
+        my @failures = $lines[0] =~ '<(?P<test>[\w-]+\.t)>';
+        my @known_issues;
+
+        if ($issues ne '') {
+            foreach my $issue (split / /, $issues) {
+                if (grep($issue, @failures)) {
+                    push(@known_issues, $issue);
```
mdoucha, line 91: Storing known issues in a testsuite setting will be a pain to maintain, not to mention that there'll be no info about why it's whitelisted. Using the existing `LTP::Whitelist` is much simpler. The modification needed to use whitelist URL from another variable is fairly small.

mdoucha (changes requested):

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
...
+        $pkgs .= " liburing2";
+    }
+
+    # install dependences
+    zypper_call("in -t pattern devel_basis");
+    zypper_call("in $pkgs");
+
+    # select latest liburing version which is supported by the system
+    if ($version eq '') {
+        $out = script_output("rpm -qi liburing2 | awk '/Version/ {print \$3}'");
```
mdoucha, line 43: Even better: `script_output('rpm -q --qf "${version}\n" liburing2');`

pevik: Why `$`? I'd expect `rpm -q --qf "%{Version}" liburing2` (even new line is not needed.

tests/kernel/io_uring.pm
acerv: But ``LTP::Whitelist`` seems to use ``LTP_KNOWN_ISSUES`` which is a specific variable for LTP. I tried to use it, but it doesn't seem to match with external framework.

acerv: Also it's not possible to use ``LTP::Whitelist`` because liburing testing suite is executed using ``make runtests`` command, which is needed in order to run **only** the tests liburing is supporting. And also, we should rewrite most of the features implemented in ``liburing/test/runtests.sh`` .

At first I just filtered out tests, but that didn't work really well, because we also need ``dmesg`` check after tests executions and to filter out only the tests which are listed inside the Makefile.

tests/kernel/io_uring.pm
mdoucha: Yes, sorry, it should be `%{version}`. Fixed. However, if you don't include the newline or some other whitespace and you have multiple package versions installed, all versions will be printed on the same line with no separator. E.g.

    1.11.21.3

instead of

    1.1
    1.2
    1.3

pevik: Make sense, but int that case it might make sense to sort and print the only one:
```
rpm -q --qf "%{Version}\n" liburing2 | sort -nr | head -1
```
(although having more versions it'd be better to determine the default one and test against it)

tests/kernel/io_uring.pm
pevik: BTW in pynfs, we just store every info in the code (not optimal, but at least it does not disappear on next openQA postgres db disaster):
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/f4992c4cae76ae0846e38c9ae3dfd4c102833378/tests/nfs/pynfs_result.pm#L29-L50

+1 to modify LTP known issues that can be use for other testsuites. We have started to use quite lot of them that it'd pay the effort.

acerv: I'm now using ``LTP::Whitelist``, but its implementation needs an adaptation to add custom known issues environment variables, because ``LTP_KNOWN_ISSUES`` and ``LTP_KNOWN_ISSUES_LOCAL`` are LTP specific.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,132 @@
...
+    @lines = $out =~ /Tests failed \(\d+\):.*/mg;
+    if (@lines) {
+        my $whitelist = LTP::WhiteList->new();
+        my $environment = {
+            product => get_var('DISTRI') . ':' . get_var('VERSION'),
+            revision => get_var('BUILD'),
+            flavor => get_var('FLAVOR'),
+            arch => get_var('ARCH'),
+            backend => get_var('BACKEND'),
+            kernel => '',
```
mdoucha, line 92: You should also set `kernel => script_output('uname -r')` in order to allow whitelisting specific kernels only. `ltp_version` could be set to the io_uring test version if we plan to build a stable test package.

It'd also be a good idea to collect this info before you start testing because when the tests fail, the system may be dead and you won't be able to query the kernel version anymore.

mdoucha: BTW, LTP::Whitelist supports skip rules which should be preferred over `LIBURING_EXCLUDE`. That's another reason to collect env info before running the tests.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,132 @@
...
+    my $timeout = get_var('LIBURING_TIMEOUT', 1800);
+    my $version = get_var('LIBURING_VERSION', '');
+    my $exclude = get_var('LIBURING_EXCLUDE', '');
+    my $pkgs = "git-core";
+    my @lines;
+    my $out;
+
+    # check if liburing2 is installed and eventually install it
+    $out = script_output("rpm -q liburing2");
+    if ($out =~ /not installed/) {
```
mdoucha, line 33: `$pkgs .= " liburing2" if script_run('rpm -q liburing2');` is probably slightly safer because it doesn't depend on current locale.

tests/kernel/io_uring.pm
```diff
@@ -0,0 +1,140 @@
...
+        harness => 'SUSE OpenQA',
+        ltp_version => $version
+    };
+
+    # run tests executables
+    my @skipped = $whitelist->list_skipped_tests($environment, 'liburing');
+    if (@skipped) {
+        my $test_exclude = join(' ', @skipped);
+        if ($exclude eq '') {
+            $test_exclude .= " " . $exclude;
```
mdoucha, line 73: `push @skipped, $exclude if $exclude;` would be cleaner but this also works. LGTM

czerw:
VR passed: https://openqa.suse.de/tests/10985240

PR #16830, opened 2023-04-12
Handle java license popup during installation

sofiasyria:
For SP4 there is an extra java license popup that breaks installation tests. This PR is adding a way to handle it in confirm_installation.pm

- Related ticket: https://progress.opensuse.org/issues/124140
- Verification run: http://falafel.suse.cz/tests/overview?build=20230411-1&distri=sle&version=15-SP4
OSD: https://openqa.suse.de/tests/10902903#next_previous

### DIFF
--- a/tests/installation/confirm_installation.pm
+++ b/tests/installation/confirm_installation.pm
@@ -9,9 +9,22 @@
 
 use strict;
 use warnings;
+use testapi;
 use base 'y2_installbase';
+use version_utils 'is_sle';
 
 sub run {
+    # For SLE 15 SP4 tests that have the Legacy module added during installation, there is
+    # additional licence popup that is handled by the following function.
+    if ((get_var("SCC_ADDONS") =~ /legacy/) && (is_sle("=15-SP4"))) {
+        my $package_license_popup = $testapi::distri->get_accept_popup_controller();
+        $package_license_popup->wait_accept_popup({
+                timeout => 3000,
+                interval => 2,
+                message => 'Accept license popup did not appear'});
+        save_screenshot;
+        $package_license_popup->accept();
+    }
     my $install_popup = $testapi::distri->get_ok_popup_controller();
     $install_popup->accept();
 }
--- a/lib/Distribution/Opensuse/Tumbleweed.pm
+++ b/lib/Distribution/Opensuse/Tumbleweed.pm
@@ -40,6 +40,7 @@ use Installation::SystemProbing::EncryptedVolumeActivationController;
 use Installation::SystemRole::SystemRoleController;
 use Installation::Popups::OKPopupController;
 use Installation::Popups::YesNoPopupController;
+use Installation::Popups::AcceptPopupController;
 use YaST::Bootloader::BootloaderSettingsController;
 use YaST::Firstboot::ConfigurationCompletedController;
 use YaST::Firstboot::HostNameController;
@@ -174,6 +175,10 @@ sub get_ok_popup_controller {
 
 sub get_yes_no_popup_controller {
     return Installation::Popups::YesNoPopupController->new();
+}
+
+sub get_accept_popup_controller {
+    return Installation::Popups::AcceptPopupController->new();
 }
 
 sub get_encrypted_volume_activation {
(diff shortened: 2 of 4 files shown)


tests/installation/confirm_installation.pm
```diff
@@ -9,9 +9,21 @@
 
 use strict;
 use warnings;
+use testapi;
 use base 'y2_installbase';
+use version_utils 'is_sle';
 
 sub run {
+    # For SLE 15 SP4 tests that have the Legacy module added during instalaltion, there is
```
jknphy, line 17: s/installation/

tests/installation/confirm_installation.pm
```diff
@@ -9,9 +9,21 @@
...
+    # additional licence popup that is handled by the following function.
+    if ((get_var("SCC_ADDONS") =~ /legacy/) && (is_sle("=15-SP4"))) {
+        my $package_license_popup = $testapi::distri->get_accept_popup_controller();
+        $package_license_popup->wait_accept_popup({
+                timeout => 3000,
```
jknphy, line 22: the confirm package popup happens intermediately, there is not need to have an explicit wait.

tests/installation/confirm_installation.pm
```diff
@@ -9,9 +9,21 @@
...
 use strict;
 use warnings;
+use testapi;
 use base 'y2_installbase';
+use version_utils 'is_sle';
 
 sub run {
+    # For SLE 15 SP4 tests that have the Legacy module added during instalaltion, there is
+    # additional licence popup that is handled by the following function.
+    if ((get_var("SCC_ADDONS") =~ /legacy/) && (is_sle("=15-SP4"))) {
```
jknphy, line 19: we need a ticket to remove this condition:
the product part of the condition via our distribution pattern and the first part just passing as parameter to a future confirm_installation_method() as we discussed in Slack.

tests/installation/confirm_installation.pm
```diff
@@ -9,9 +9,21 @@
...
+        my $package_license_popup = $testapi::distri->get_accept_popup_controller();
+        $package_license_popup->wait_accept_popup({
+                timeout => 3000,
+                interval => 2,
+                message => 'Accept license popup did not appear'});
+        $package_license_popup->accept();
```
jknphy, line 25: please add an screenshot in the middle to see the java license popup.

tests/installation/confirm_installation.pm
jknphy: can you share the link? what is the timeout and how much needs in the slower arch? 3000 is too much, and we should not abuse of this mechanism of explicit wait.

jknphy: btw your runs are done in your machine, for timeout I would recommend OSD, we might find something funny...

tests/installation/confirm_installation.pm
sofiasyria: I will open a ticket for this, but can't say that I have full understanding of your solution proposal. You want this java popup to be handled inside the confirm_installation() which will receive some arguments and act accordingly? The conditions should be inside the library?

sofiasyria: Can you please, check here: https://progress.opensuse.org/issues/127529

tests/installation/confirm_installation.pm
sofiasyria: I am running some VRs on OSD now. For the failed runs without the timeout , it might be this one: http://falafel.suse.cz/tests/901#step/confirm_package_license/2 or it has desappeared.

jknphy: that is some other failure during your development, I don't think is needed extra wait, we already wait 2 mins iirc,
`Can't call method "exist" on an undefined value`

jknphy: ok you added to the ticket (just noticed), so I'm fine if we do in the follow-up.

tests/installation/confirm_installation.pm
jknphy: I rephrased a bit but you got the idea, thanks!

jknphy:
@sofiasyria we probably needs to add to the condition the case where we have minimal patterns:
`PATTERNS: base,minimal`
https://openqa.suse.de/tests/10910263#step/confirm_installation/1

PR #16834, opened 2023-04-12
Add users in slurm compute nodes

b10n1k:
Users should appear among all the nodes of the cluster and their uid/gid should match. It seems that the uid/gid are assigned correctly.

This used to work without the creation of the users on compute nodes for SLE15SP4 and below but apparently it shouldnt. This commit just makes things right.

- Related ticket: https://progress.opensuse.org/issues/127487
- Verification run: https://aquarius.suse.cz/tests/15865

### DIFF
--- a/tests/hpc/slurm_master.pm
+++ b/tests/hpc/slurm_master.pm
@@ -225,10 +225,10 @@ sub t01_accounting() {
     my $description = 'Basic check for slurm accounting cmd';
     my $result = 0;
     my %users = (
-        'user_1' => 'Sebastian',
-        'user_2' => 'Egbert',
-        'user_3' => 'Christina',
-        'user_4' => 'Jose',
+        'user_1' => 'sebastian',
+        'user_2' => 'egbert',
+        'user_3' => 'christina',
+        'user_4' => 'jose',
     );
 
     ##Add users TODO: surely this should be abstracted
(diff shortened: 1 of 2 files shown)


tests/hpc/slurm_master.pm
```diff
@@ -225,10 +225,10 @@ sub t01_accounting() {
     my $description = 'Basic check for slurm accounting cmd';
     my $result = 0;
     my %users = (
-        'user_1' => 'Sebastian',
```
schlad, line 228: keeping upper case imo makes sense as slurm should handle that and this is a test per se - to see if indeed users will be added to the db with lower cases

tests/hpc/slurm_slave.pm
```diff
@@ -28,7 +28,18 @@ sub run ($self) {
...
+        'user_2' => 'egbert',
+        'user_3' => 'christina',
+        'user_4' => 'jose',
+    );
 
+    ##Add users TODO: surely this should be abstracted
+    script_run("useradd $users{user_1}");
+    script_run("useradd $users{user_2}");
+    script_run("useradd $users{user_3}");
+    script_run("useradd $users{user_4}");
```
schlad, line 42: Users should appear among all the nodes of the cluster and their uid/gid
should match. It seems that the uid/gid are assigned correctly.

This used to work without the creation of the users on compute nodes for
SLE15SP4 and below but apparently it shouldnt. This commit just makes things
right.

---

Where could I find docs describing this (the fact that each user shall be created on each compute node)?

b10n1k: https://slurm.schedmd.com/accounting.html#infrastructure

b10n1k:
close in favor of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16891

PR #16835, opened 2023-04-12
Common criteria: use bootloader_start for hdd creation

ilmanzo:
Previous refactoring introduced double test for `bootloader`, resulting in a wrong behavior. We schedule `bootloader_start` that takes care internally of the architecture differences. 
On the same principle we can dismiss specific UEFI schedule because `bootloader_uefi` is already managed by `bootloader_start` (see https://gitlab.suse.de/qe-security/osd-sle15-security/-/merge_requests/92) 

- Related ticket: https://progress.opensuse.org/issues/127250
- Verification runs: 
     - x86_64      https://openqa.suse.de/tests/10902939  
     - x86_64@uefi https://openqa.suse.de/tests/10902940           
     - s390x       https://openqa.suse.de/tests/10902941           
     - aarch64     https://openqa.suse.de/tests/10902942


tjyrinki (approved):
Great to get those back to working!

One CI test failed, rerunning now to see whether it was temporary or if there's eg some whitespace issue.

PR #16836, opened 2023-04-12
Ensure that zfs is not available on SLE but on openSUSE

asdil12:
https://progress.opensuse.org/issues/123100


asdil12:
https://openqa.suse.de/tests/10903311#step/grub2_test/77

asdil12:
https://openqa.opensuse.org/tests/3224299#step/zfs/30

PR #16842, opened 2023-04-13
lib/utils: Enable online repos on installations from live media if applicable

Vogtinator:
Revert the previous commits, implement it how it was meant to.

"addon repos" == "online repos" apparently, which is why I previously thought this would be the wrong approach.

- Related ticket: https://progress.opensuse.org/issues/127565
- Verification runs:
15.4 Incidents: https://openqa.opensuse.org/tests/3223992
15.4 Images (Upgrade from 15.2): https://openqa.opensuse.org/tests/3223991
15.4 Images (installation): https://openqa.opensuse.org/tests/3223999

CC @rfan1

### DIFF
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -563,7 +563,7 @@ sub load_slepos_tests {
 sub load_system_role_tests {
     # This part is relevant only for openSUSE
     if (is_opensuse) {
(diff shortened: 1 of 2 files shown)


rfan1 (approved):
LGTM

PR #16846, opened 2023-04-13
Removing soft failure  bsc#1163000 marked as WONTFIX

sofiasyria:
- Related ticket: https://progress.opensuse.org/issues/126608

- Verification run: https://openqa.suse.de/tests/10912049#next_previous
 https://openqa.suse.de/tests/10918991
https://openqa.suse.de/tests/10902982


lib/service_check.pm
```diff
@@ -228,11 +228,6 @@ service package name.
 
 sub _is_applicable {
     my ($srv_pkg_name) = @_;
-    if ($srv_pkg_name eq 'kdump' && is_s390x) {
-        # workaround for bsc#116300 on s390x
-        record_soft_failure 'bsc#1163000 - System does not come back after crash on s390x';
-        return 0;
-    }
```
jknphy, line 235: don't we need to convert it into record_info, that `record 0;` is an early exit that probably is needed.

sofiasyria: We would if the issue reported was still reproducible, which is not the case.

rakoenig:
LGTM

PR #16849, opened 2023-04-13
Bump up timeout in yast2_software_management

sofiasyria:
This issue has appeared just once and seems sporadic, therefore slightly increasing timeout should prevent it from reoccurring in the future

- Related ticket: https://progress.opensuse.org/issues/127604
- Verification run: https://openqa.suse.de/tests/10912372#next_previous


rakoenig:
LGTM.

PR #16854, opened 2023-04-13
Improve stability of squid tests by checking for network port listening

ilmanzo:
Test was flaky so we try to improve the stability with some added checks.
Also redirect `curl` stderr so it can be validated via regular expression

- Related ticket: https://progress.opensuse.org/issues/125585
- Verification runs: 
  - https://openqa.suse.de/tests/10913298 
  - https://openqa.suse.de/tests/10913299


paolostivanin:
LGTM

PR #16855, opened 2023-04-13
Fix assert_screen in apply_workaround_poo124652

rakoenig:
- Related ticket:  https://progress.opensuse.org/issues/127499
- Needles: -
- Verification run: https://openqa.suse.de/tests/10912707

- Previous PR turned out to make some  test modules less stable.

### DIFF
--- a/lib/YaST/workarounds.pm
+++ b/lib/YaST/workarounds.pm
@@ -39,7 +39,7 @@ sub apply_workaround_poo124652 {
     if (!check_screen($mustmatch, %args)) {
         record_info('poo#124652', 'poo#124652 - gtk glitch not showing dialog window decoration on openQA');
         send_key('shift-f3', wait_screen_change => 1);
-        assert_screen('style-sheet-selection-popup');
+        check_screen('style-sheet-selection-popup', 10);
         send_key('esc', wait_screen_change => 1);
         # in some verification tests this didn't work, so let's check
         if (!check_screen($mustmatch)) {


lib/YaST/workarounds.pm
```diff
@@ -39,7 +39,7 @@ sub apply_workaround_poo124652 {
     if (!check_screen($mustmatch, %args)) {
         record_info('poo#124652', 'poo#124652 - gtk glitch not showing dialog window decoration on openQA');
         send_key('shift-f3', wait_screen_change => 1);
-        assert_screen('style-sheet-selection-popup');
+        check_screen('style-sheet-selection-popup', 10);
```
jknphy, line 42: they are the same, one function through exception the other not, I think you were lucky with the VR,
or what is the difference?

rakoenig: `assert_screen` fails when the needle is not matched. That means the other approach with "maximize/unmaximize" screen will never kick in when the "shift-f3" approach does fail. `check_screen` does not fail the complete test,it just exits and the test module goes on and can handle the situation.

jknphy: I see, the problem is that I cannot see the resizing, could you please put one save_screenshot after both send_(alt-f10...) and run some verification just in case, then we can merge it.

jknphy: feel free to add those screenshot any time. Let's merge this for PublicRC.

PR #16860, opened 2023-04-14
Need recover the workaround for bsc#1129504

lemon-suse:
https://openqa.suse.de/tests/10905648# The failure means the workaround still needed.

- Related ticket: https://progress.opensuse.org/issues/126602
- Needles: N/A
- Verification run: N/A


lemon-suse:
It is a needle issue, after update needle it works. https://openqa.suse.de/tests/10918395#

PR #16867, opened 2023-04-15
WIP: PC wait-ssh with 3 checks 

m-dati:
PR for inspection only.

P.C. `wait-for-ssh` has been almost rewitten, to detect the system-up condition, running 3 differents chk types, in sequence, from a `@command` array.
This draft PR has been created to show the strucutre and identify the best command to use.

Related ticket: https://progress.opensuse.org/issues/126218 poo
Needles:none
Verification run: [3814](http://10.168.4.117/tests/3814), [3816](http://10.168.4.117/tests/3816)
More tests coming soon

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -352,11 +352,13 @@ sub wait_for_ssh {
     my $start_time = time();
     my $check_port = 1;
     my $sleep_period = $args{ignore_wrong_pubkey} ? 20 : 1;
-
-    my $ana = get_var('PUBLIC_CLOUD_WAITSSH_SYSANA', 0);
-    my $command = $ana ? 'systemd-analyze time' : 'sudo systemctl is-system-running --wait';
-    my $output;
-    my $output2;
+    #
+    my @sem = (1, 1, 1);
+    my @command = ('sudo systemctl is-system-running --wait', 'systemd-analyze time', 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"');
+    my @output = ('running|degraded', 'Startup finished in', 'Reached target ');
+    my $kostate = 'maintenance|stopping|offline|unknown';    # wrong states of command[0]
+    my $pass = 0;
+    my @out;
 
     # Looping until reaching timeout or passing two conditions :
     # - SSH port 22 is reachable
@@ -369,36 +371,41 @@ sub wait_for_ssh {
         else {
             # On boottime test we do hard reboot which may change the instance address
             script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
-            $output = $self->run_ssh_command(
-                # cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"',
-                cmd => $command,
-                proceed_on_failure => 1,
-                username => $args{username});
-            $output2 = $self->run_ssh_command(
-                cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"',
-                proceed_on_failure => 1,
-                username => $args{username});
-            if ($output2 =~ m/Reached target /) {
-                record_info("CHECK SSH J", "wait ssh journal OK " . $duration . "s\n" . "System: " . $output2);
+            # here 3 check methods 0 to 2 from @command used:
+            foreach my $j (0 .. 2) {
+                $out[$j] = $self->run_ssh_command(cmd => $command[$j], proceed_on_failure => 1, username => $args{username});
+                if ($sem[$j] && $out[$j] =~ m/$output[$j]/) {
+                    --$sem[$j];
+                    # DEBUG print
+                    if ($j == 0 && $out[$j] =~ m/degraded/) {
+                        $out[$j] = $self->run_ssh_command(cmd => 'sudo systemctl --failed', proceed_on_failure => 1, username => $args{username});
+                    }
+                    my $msg = 'wait ssh ' . (++$pass) . ' ok, ' . (time() - $start_time) . 's, System:' . "\n" . $out[$j];
+                    record_info("CHECK SSH", $msg);
+                }
+                elsif ($j == 0 && $sem[$j] && $out[$j] =~ m/$kostate/) {
+                    # check if system status ko using command 0:
+                    $sem[0] = 2;
+                    last;    # exit For.
+                }
+                elsif ($out[$j] =~ m/Permission denied \(publickey\).*/) {
+                    die "ssh permission denied (pubkey)" unless $args{ignore_wrong_pubkey};
+                }
             }
-            if (($output =~ m/running|degraded/i and !$ana) or ($output =~ m/Startup finished in/ and $ana)) {
-                # DEBUG print
-                record_info("CHECK SSH", "wait ssh OK " . $duration . "s\n" . "System: " . $output);
+            if ($pass == 3) {    # all checks passed
                 $self->journal_upload("/tmp/waitssh_journal.txt", username => $args{username});
                 return $duration;
             }
-            elsif ($output =~ m/Permission denied \(publickey\).*/) {
-                die "ssh permission denied (pubkey)" unless $args{ignore_wrong_pubkey};
-            }
-            elsif (defined $output && length $output > 0 && !$ana) {
-                last;    # loop break error
-            }
+            last if ($sem[0] > 1);    # exit While, due to command 0
         }
         sleep $sleep_period;
     }
 
     # DEBUG print
-    record_info("CHECK SSH", "wait ssh NOK " . time() - $start_time . "s\n" . "System: " . $output);
+    foreach my $j (0 .. 2) {
+        my $msg = 'wait ssh ' . ($j + 1) . ' Nok ' . (time() - $start_time) . 's, System:' . "\n" . $out[$j];
+        record_info("CHECK SSH", $msg) if $sem[$j];
+    }
     script_run("ssh  -i /root/.ssh/id_rsa -v $args{username}\@$args{public_ip} true", timeout => 360);
 
     unless ($args{proceed_on_failure}) {
@@ -422,10 +429,13 @@ sub wait_for_ssh {
 }
 
 sub journal_upload {
+    # moved this code from wait_for_ssh into THIS temporary function
+    # for faster log provisioning and code reuse, improvement and control.
     my ($self, $log, %args) = @_;
     my $usr = $args{username};
-    my $com = 'sudo journalctl -b --no-pager| grep -iE "Reached target" > ' . $log;
+    my $com = 'sudo journalctl -b --no-pager > ' . $log;
     eval {
+        return unless ($log);
         if (!get_var('PUBLIC_CLOUD_SLES4SAP')) {
             $self->run_ssh_command(cmd => $com, proceed_on_failure => 1, username => $usr, timeout => 360);
             $self->upload_log($log, failok => 1);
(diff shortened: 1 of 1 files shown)


lib/publiccloud/instance.pm
```diff
@@ -352,40 +352,65 @@ sub wait_for_ssh {
     my $start_time = time();
     my $check_port = 1;
     my $sleep_period = $args{ignore_wrong_pubkey} ? 20 : 1;
-
+    #
+    my @sem = (1, 1, 1);
+    my @out = ('') x 3;
+    my @command = ('sudo systemctl is-system-running --wait', 'systemd-analyze time', 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"');
```
pdostal, line 358: Isn't one command enough?

m-dati: Yes, indeed infact as I told above, this PR is to verify _which_ command is the best one to use for this system-up definition. I put VRs results in the same test in order to compare them. So this PR is not to be merged as it is now, but at the end of evaluation, only one will be used.

m-dati:
Please consider that the `command[0]` has 3 possible steps, based on [documentation](https://www.freedesktop.org/software/systemd/man/systemctl.html#is-system-running):  
1) preparing UP, before and during waiting system up; using `--wait` the _not yet ready_ states will not appear, while waiting;
2) valid UP detected: _running_ or _degraded_;
3) invalid UP, other states with exit >0

Therefore, in this test code, running 3 different check types, we'll find some control steps related to id=0 or command[0] only, to deal with such steps, like L377 and L395, different from the other check commands 1 and 2. In case this command will not be consider the best one for our scope, these steps will be removed too.

asmorodskyi:
generally I like the direction where it is heading . But I would like better to understand what actually going on . Algorithm seems pretty complex

m-dati:
> generally I like the direction where it is heading . But I would like better to understand what actually going on . Algorithm seems pretty complex

Yes, sorry I added a note in the poo to try explaining the process a bit: [126218#note-5](https://progress.opensuse.org/issues/126218#note-5)

lib/publiccloud/instance.pm
```diff
@@ -352,40 +352,65 @@ sub wait_for_ssh {
...
             script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
-
-            my $output = $self->run_ssh_command(
-                cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"',
-                proceed_on_failure => 1,
-                username => $args{username});
-            if ($output =~ m/Reached target.*/) {
+            # here 3 check methods 0 to 2 from @command used:
+            foreach my $j (0 .. 2) {
+                $out[$j] = $self->run_ssh_command(cmd => $command[$j], proceed_on_failure => 1, username => $args{username});
```
pdostal, line 375: @m-dati checking the output like `$self->run_ssh_command(cmd => $command[$j], proceed_on_failure => 1` is not very good approach. Instead, please try to make use of the awesome work of @grisu48 and his `$self-> ssh_assert_script_run($command[$j])` which will pass if the command exited with `0` as every well-behaved command on Linux does and fail otherwise. Alternatively, there is also `$self-> ssh_script_run()` which will return the exit code for you to process elsewhere.

m-dati: Sorry, this approach was already present in the original version, in master you can still see it. there.
Anyway in this first phase, we still have to choose which UP command to use: the best infrastructure to use could be much more clear after removed the not needed commands.

m-dati:
more VRs:
http://autobot.qa.suse.de/tests/3818 PASS
http://autobot.qa.suse.de/tests/3819 PASS
http://autobot.qa.suse.de/tests/3820 PASS
http://autobot.qa.suse.de/tests/3821 FAIL timeout (tested loop exit from L404)

lib/publiccloud/instance.pm
m-dati: I studied the functions `run_ssh_command` and `ssh_assert_script_run` and I noted that the steps performed are _almost_ identical in both, with the only exception that at the end the first one executes `script_run` while the second one executes `assert_script_run` instead.  
When it comes to `run_ssh_command` and `ssh_script_run` (without assert) the steps performed are _almost_ identical in both the two, with the first one including some parameters providing more flexibilty.

Now, referring to [note-5](https://progress.opensuse.org/issues/126218#note-5), it could occur that the exit code 0 does not fully guarantee target reached, moreover depending on the check type used.
Moreover the ok status could occurr in the loop after some time with result exit>0, that an assertion will soon terminate.  
Finally for check type n.1 I am considering to let it to pass even with an eventual 'degraded' state (exit >0), allowing the test to proceed and check afterward the eventual impact.

m-dati:
New VR on OSD: https://openqa.suse.de/tests/10939875 
Note here the wait_for_ssh executed 3 times: 
1 in prepare_instance, 
2 in patch_and_reboot -> softreboot: 2.1 fail for system DOWN , then 2.2 pass after UP.

m-dati:
In this VR [10940502](https://openqa.suse.de/tests/10940502#step/patch_and_reboot/35) 
we can see the check type n.1 `degraded` due to failed `systemd-vconsole-setup.service` in box [36](https://openqa.suse.de/tests/10940502#step/patch_and_reboot/36); however the other checks Pass normally and  doesn't provide such information

m-dati:
After analyzed the following **journal** extracted  tests logs:

https://openqa.suse.de/tests/10940502/logfile?filename=patch_and_reboot-waitssh_journal.txt (type n.1 degraded)
```
Apr 18 19:17:21 ip-172-31-16-32 systemd[1]: systemd-vconsole-setup.service: Main process exited, code=exited, status=1/FAILURE
Apr 18 19:17:56 ip-172-31-16-32 systemd[1]: Reached target Cloud-init target.
Apr 18 19:17:56 ip-172-31-16-32 systemd[1]: Startup finished in 2.018s (kernel) + 2.691s (initrd) + 32.944s (userspace) = 37.654s.
Apr 18 19:18:03 ip-172-31-16-32 systemd[1485]: Reached target Main User Target.
Apr 18 19:18:06 ip-172-31-16-32 sudo[1524]: ec2-user : PWD=/home/ec2-user ; USER=root ; COMMAND=/usr/bin/systemctl is-system-running --wait
```

https://openqa.suse.de/tests/10939875/logfile?filename=patch_and_reboot-waitssh_journal.txt (all 3 types ok)

```
Apr 18 16:07:34 ip-172-31-82-152 systemd[1]: Startup finished in 1.092s (kernel)
Apr 18 16:07:34 ip-172-31-82-152 systemd[1]: Reached target Cloud-init target.
Apr 18 16:07:47 ip-172-31-82-152 systemd[1633]: Reached target Main User Target.
Apr 18 16:07:49 ip-172-31-82-152 sudo[1672]: ec2-user : PWD=/home/ec2-user ; USER=root ; COMMAND=/usr/bin/systemctl is-system-running --wait
```

http://autobot.qa.suse.de/tests/3820/logfile?filename=run_ltp-waitssh_journal.txt  (all 3 types ok)
```
Apr 17 16:36:52 localhost systemd[1499]: Reached target Main User Target.
Apr 17 16:36:53 localhost sudo[1512]: susetest : PWD=/home/susetest ; USER=root ; COMMAND=/usr/bin/systemctl is-system-running --wait
Apr 17 16:36:55 localhost systemd[1]: Startup finished in 1.690s (kernel) + 4.623s (initrd) + 29.861s (userspace) = 36.175s.
```

http://autobot.qa.suse.de/tests/3818#step/img_proof/129  (all 3 types ok)
```
Apr 17 16:38:09 openqa-suse-de-33b14f8bc6d33dab systemd[1]: Reached target Cloud-init target.
Apr 17 16:38:09 openqa-suse-de-33b14f8bc6d33dab systemd[1]: Startup finished in 1.718s (kernel) + 5.883s (initrd) + 58.270s (userspace) = 1min 5.872s.
Apr 17 16:39:01 openqa-suse-de-33b14f8bc6d33dab sudo[1750]: azureuser : PWD=/home/azureuser ; USER=root ; COMMAND=/usr/bin/systemctl is-system-running --wait
Apr 17 16:39:01 openqa-suse-de-33b14f8bc6d33dab systemd[1741]: Reached target Main User Target.
```

we can see that the check type n.1 is present in the logs after the other types, when the system is full running, but is also able to inform about failed services, if any.

lib/publiccloud/instance.pm
```diff
@@ -352,40 +352,65 @@ sub wait_for_ssh {
...
             # On boottime test we do hard reboot which may change the instance address
             script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
-
-            my $output = $self->run_ssh_command(
-                cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"',
-                proceed_on_failure => 1,
-                username => $args{username});
-            if ($output =~ m/Reached target.*/) {
+            # here 3 check methods 0 to 2 from @command used:
+            foreach my $j (0 .. 2) {
```
asmorodskyi, line 374: I don't know maybe it is something personal problem for me , but I having hard times to "process" this code and understand what is going on here , you have 3 arrays which contains 3 pieces of 3 different flows split by array ID so to understand what is going on in each case you need to take one pieces from each array . I think if we will drop this loop and drop arrays and just write actions one by one it would be much much readable and easier to understand 

```
my $is_sys_run_cmd = 'sudo systemctl is-system-running --wait';
$out_is_sys_run = $self->run_ssh_command(cmd => $is_sys_run_cmd, proceed_on_failure => 1, username => $args{username});
if ($is_sys_run_out =~ m/degraded/) {
                    my $failed = $self->run_ssh_command(cmd => 'sudo systemctl --failed', proceed_on_failure => 1, username => $args{username});
                    record_info("CHECK SSH", 'wait ssh ' . (++$pass) . ' fail.serv.,' . (time() - $start_time) . 's, System:' . "\n" . $failed);
                }
```

lib/publiccloud/instance.pm
```diff
@@ -352,40 +352,65 @@ sub wait_for_ssh {
     my $start_time = time();
     my $check_port = 1;
     my $sleep_period = $args{ignore_wrong_pubkey} ? 20 : 1;
-
+    #
+    my @sem = (1, 1, 1);
```
asmorodskyi, line 356: I didn't fully get why we need this one ?

m-dati: Those are semaphores: like `check_port` for ssh port, each `sem[j]=1`, when condition of check-type j is detected `true`, it marks as **done** the test assuming value `sem[j]=0`, that will no more executed on next loops, and when all 3 condition are _true_, that is $pass==3, the test PASS.

m-dati: But all this structure is to run the 3 check types in each test-run, but after analyzed the last tests, I'd propose to use the only type n.1: 
`sudo systemctl is-system-running --wait`

lib/publiccloud/instance.pm
m-dati: Please consider to use the [note-5](https://progress.opensuse.org/issues/126218#note-5) for some indication. Moreover I did above an analysis of the last tests results, so it may be tha now I could use one only check type :) .

m-dati:
see PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16890

lib/publiccloud/instance.pm
pdostal: Maybe instead of loop we can use `ssh_script_retry()`?

Also only `journalctl` requires us to parse the output.
Both `systemctl is-system-running --wait` as well as `systemd-analyze time` should return `0` only if/when the system is booted.

m-dati:
Closing this PR being the work continued in the other named PR and branch

PR #16868, opened 2023-04-17
Fix needle match timeout for firefox_appearance test

rfan1:
https://progress.opensuse.org/issues/127511

At the same time, remove the loop to match the needle
- Verification run:
[QEM](https://openqa.suse.de/tests/overview?distri=sle&build=rfan0417)
[SLE15SP5](https://openqa.suse.de/tests/overview?build=rfan0417_1&version=15-SP5&distri=sle&test=firefox_desktop)
- Needles: Added via openQA webUI

### DIFF
--- a/tests/x11/firefox/firefox_appearance.pm
+++ b/tests/x11/firefox/firefox_appearance.pm
@@ -1,7 +1,7 @@
 # SUSE's openQA tests
 #
 # Copyright 2009-2013 Bernhard M. Wiedemann
-# Copyright 2012-2016 SUSE LLC
+# Copyright 2012-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: MozillaFirefox
@@ -13,7 +13,7 @@
 # - Open url "addons.mozilla.org/en-US/firefox/addon/opensuse" and check
 # - Install opensuse theme and check
 # - Exit firefox
-# Maintainer: wnereiz <wnereiz@gmail.com>
+# Maintainer: wnereiz <wnereiz@gmail.com>, QE Core <qe-core@suse.de>
 
 use strict;
 use warnings;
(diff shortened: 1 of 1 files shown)


Vogtinator:
I don't think the timeout is the issue here. The difference between the [failing](https://openqa.suse.de/tests/10901283#step/firefox_appearance/14) and [working](https://openqa.suse.de/tests/10901960#step/firefox_appearance/15) cases is the popup at the top right. That popup should probably be handled, maybe just included in the needle.

rfan1:
> I don't think the timeout is the issue here. The difference between the [failing](https://openqa.suse.de/tests/10901283#step/firefox_appearance/14) and [working](https://openqa.suse.de/tests/10901960#step/firefox_appearance/15) cases is the popup at the top right. That popup should probably be handled, maybe just included in the needle.

Thank you @Vogtinator 

I would increase the timeout at `last if check_screen 'firefox-appearance-installed', 90;`

Vogtinator:
AFAICT the test doesn't use needles in a way that works reliably. `check_screen` with a timeout is guaranteed to be racy and needs to be avoided. It's wrong in pretty much all cases. Use `assert_screen` with multiple matches instead and then act based on the matched needle, e.g.:

```
        assert_and_click qw(firefox-appearance-addto firefox-appearance-addto-permissions_requested);
        assert_screen 'firefox-appearance-installed' if match_has_tag('firefox-appearance-addto');
```

rfan1:
> AFAICT the test doesn't use needles in a way that works reliably. `check_screen` with a timeout is guaranteed to be racy and needs to be avoided. It's wrong in pretty much all cases. Use `assert_screen` with multiple matches instead and then act based on the matched needle, e.g.:
> 
> ```
>         assert_and_click qw(firefox-appearance-addto firefox-appearance-addto-permissions_requested);
>         assert_screen 'firefox-appearance-installed' if match_has_tag('firefox-appearance-addto');
> ```

I agree with you, let me revise the code

rfan1:
@Vogtinator @GraceWang571 
Can you please help review this PR again?

Vogtinator (approved):

GraceWang571 (approved):

PR #16873, opened 2023-04-17
Assert x3260 log with grep instead of needles

realcharmer:
Replace needle matching of log output with simple `assert_script_run` and `grep`. This will prevent occasional fails due to the log output not rendering properly in X11 terminal (see https://openqa.suse.de/tests/10929330#step/x3270_ssl/28). The needle can later be removed.

- Related ticket: https://progress.opensuse.org/issues/127730
- Verification run: https://openqa.suse.de/tests/10932103


paolostivanin:
LGTM

ilmanzo:
Just to be sure: can you add VR for other architectures please ?

realcharmer:
Good thinking. It revealed a failure in aarch64 due to a fast window switching before all windows could be spawned. Fixed with `wait_still_screen`. Verification runs provided for all platforms.

PR #16875, opened 2023-04-17
Increase timeout for boot log collection

grisu48:
Increase the timeout for collecting the boot-diagnostics (i.e. serial terminal) log on Azure.

- Related failure: https://openqa.suse.de/tests/10929910#step/ssh_interactive_end/31
- Verification run: https://openqa.suse.de/tests/10932333

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -387,7 +387,7 @@ sub cleanup {
 
     my $id = $args->{my_instance}->{instance_id};
 
-    script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt");
+    script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt", timeout => 120, die_on_timeout => 0);
     upload_logs("bootlog.txt", failok => 1);
 
     $self->SUPER::cleanup();


lib/publiccloud/azure.pm
```diff
@@ -387,7 +387,7 @@ sub cleanup {
 
     my $id = $args->{my_instance}->{instance_id};
 
-    script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt");
+    script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' > bootlog.txt", timeout => 120, die_on_timeout => 0);
```
pdostal, line 390:
```suggestion
    script_run("az vm boot-diagnostics get-boot-log --ids $id | jq -r '.' | tee bootlog.txt", timeout => 120, die_on_timeout => 0);
```

grisu48: I'd rather not. The log can be large, and spamming openQA with this doesn't seem to give us any obvious benefit.

pdostal: My point was that the log can contain error / debug messages. But if we proceed and upload the file even in case of failure then it's probably good.

pdostal (approved):

asmorodskyi:
LGTM !

PR #16876, opened 2023-04-17
Remove Java popup handling

sofiasyria:
On current maintenance updtaes, there is no update for java package , so the license popup does nto appear anymore and the java handling solution is causing new failures.

- Related ticket: https://progress.opensuse.org/issues/127529
- Verification runs: http://falafel.suse.cz/tests/overview?build=20230416-1&distri=sle&version=15-SP4


rakoenig:
LGTM.

PR #16877, opened 2023-04-17
Increase timeout value for startupdate

lemon-suse:
We need wait more time for 'startupdate'.

- Related ticket: https://progress.opensuse.org/issues/127715
- Needles: N/A
- Verification run: 
  http://openqa.suse.de/tests/10932788#
  http://openqa.suse.de/tests/10932789#
  http://openqa.suse.de/tests/10932790#
  http://openqa.suse.de/tests/10932791#
  http://openqa.suse.de/tests/10932793#

### DIFF
--- a/tests/installation/start_install.pm
+++ b/tests/installation/start_install.pm
@@ -31,9 +31,9 @@ sub run {
     # Also, virtual machines for testing can be really slow in this step
     my $started_timeout = get_var('LIVECD') ? 1200 : 300;
     if (is_upgrade) {
+        wait_still_screen 2;
         send_key $cmd{update};
-        sleep 1;
-        assert_screen [qw(startupdate startupdate-conflict license-popup)], 5;
+        assert_screen [qw(startupdate startupdate-conflict license-popup)], 10;
 
         while (match_has_tag("startupdate-conflict") || match_has_tag("license-popup")) {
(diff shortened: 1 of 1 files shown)


tests/installation/start_install.pm
```diff
@@ -31,9 +31,10 @@ sub run {
     # Also, virtual machines for testing can be really slow in this step
     my $started_timeout = get_var('LIVECD') ? 1200 : 300;
     if (is_upgrade) {
+        wait_still_screen 2;
         send_key $cmd{update};
         sleep 1;
```
rfan1, line 36: Why `sleep 1` here?

lemon-suse: It is the original code, I think sleep here just try to ensure the key sent can be received.

rakoenig: I also wonder about that `sleep`. That test module is running on the worker, so why do we put a sleep in here. We increased the timeout of the following `assert_screen` from 5 to 10 seconds, to what is the difference betweern sleeping and assert_screen for 10 seconds?

lemon-suse: I removed the sleep since it is not needed, we have already increased the wait time for assert_screen.

rfan1 (approved):

PR #16884, opened 2023-04-18
extra_tests_kernel: Add BCC tools test

richiejp:
Add smoke test for BCC. There are many BCC tools and they all are different. So this just selects a few which might be interesting to us.

Note that it doesn't check the output yet because I want to figure out how to properly do that with bpftrace first. Also a lot of these tools have upstream issues and are inherently flaky. We are only running them to make sure the compiler framework and kernel work.

@czerw @pevik


richiejp:
Verification run: https://openqa.suse.de/tests/10933416#

czerw (approved):

pevik (approved):

PR #16885, opened 2023-04-18
container_selinux: check that containers are using selinux

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/119827
- Verification run: https://openqa.opensuse.org/tests/3231708

### DIFF
--- a/tests/security/selinux/container_selinux.pm
+++ b/tests/security/selinux/container_selinux.pm
@@ -16,12 +16,12 @@ sub run {
     select_serial_terminal;
 
     # ensure that SELinux is enabled and in enforcing mode
-    validate_script_output("sestatus", sub { m/SELinux\ status: .*enabled.* Current\ mode: .*enforce/sx });
+    validate_script_output("sestatus", sub { m/SELinux\ status: .*enabled.* Current\ mode: .*enforcing/sx });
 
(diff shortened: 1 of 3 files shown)


tests/security/selinux/container_selinux.pm
```diff
@@ -0,0 +1,31 @@
...
+    select_serial_terminal;
+
+    # ensure that SELinux is enabled and in enforcing mode
+    validate_script_output("sestatus", sub { m/SELinux\ status: .*enabled.* Current\ mode: .*enforcing/sx });
+
+    my $image = 'registry.opensuse.org/opensuse/tumbleweed:latest';
+    my $container_name = 'test_container_selinux';
+
+    assert_script_run("podman pull $image");
+    assert_script_run("podman run --name $container_name -dt $image");
```
ilmanzo, line 25: `podman run` on a new image implicitly does a pull, so maybe it's not necessary ?
since images can be quite big, I'd insert a generous timeout here :)

paolostivanin: I'd prefer to keep those cmds separate, since once it's involving downloading and the other running a container.
Tumbleweed image is about 40MB, so 90s should always be more than enough. But I agree that it's better to be safe than sorry :) I'll put 300s.

PR #16886, opened 2023-04-18
Check if the Azure blob is uploaded correctly

pdostal:
- Verification run: [new image](https://pdostal-server.suse.cz/t4295), [existing image](https://pdostal-server.suse.cz/t4296)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -61,6 +61,7 @@ sub get_image_id {
 
 sub find_img {
     my ($self, $name) = @_;
+    my ($json, $md5, $image);
 
     return if (!$self->resource_exist());
 
@@ -69,21 +70,40 @@ sub find_img {
     $name =~ s/\.xz$//;
     $name =~ s/\.vhdfixed$/-$sku.vhd/;
 
-    my $container = $self->container;
     my $storage_account = get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'eisleqaopenqa');
     my $key = $self->get_storage_account_keys($storage_account);
-    my $cmd_show = "az storage blob show --account-key $key -o json " .
-      "--container-name '$container' --account-name '$storage_account' --name '$name' " .
-      '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"';
-    record_info('BLOB INFO', script_output($cmd_show, proceed_on_failure => 1));
 
-    my $json = script_output("az image show --resource-group " . $self->resource_group . " --name $name", 60, proceed_on_failure => 1);
-    record_info('IMG INFO', $json);
-    my $image;
-    eval {
-        $image = decode_azure_json($json)->{name};
-    };
-    record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    $json = script_output("az storage blob show --account-key $key -o json " .
+          "--container-name '$self->container' --account-name '$storage_account' --name '$name' " .
+          '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"',
+        proceed_on_failure => 1);
+    record_info('BLOB INFO', $json);
+    eval { $md5 = decode_azure_json($json)->{md5}; };
+    if ($@) {
+        record_info('BLOB NOT-FOUND', "Cannot find blob $name. Need to upload it.\n$@");
+    } elsif (!$md5 || $md5 !~ /^[a-fA-F0-9]{32}$/) {
+        record_info('INVALID', "The blob $name does not have valid md5 field.");
+        return $image;
+    }
+
+    my $arch = (check_var('PUBLIC_CLOUD_ARCH', 'arm64')) ? 'Arm64' : 'x64';
+    if ($arch eq 'Arm64') {
+        my $resource_group = $self->resource_group;
+        my $gallery = $self->image_gallery;
+        my $version = calc_img_version();
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        $json = script_output("az sig image-version show --resource-group '$resource_group' --gallery-name '$gallery' " .
+              "--gallery-image-definition '$definition' --gallery-image-version '$version'", timeout => 60 * 30);
+    } else {
+        $json = script_output("az image show --resource-group " . $self->resource_group . " --name $name", 60, proceed_on_failure => 1);
+        record_info('IMG INFO', $json);
+        eval { $image = decode_azure_json($json)->{name}; };
+        record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    }
+    record_info('IMGV INFO', $json);
+    eval { $image = decode_azure_json($json)->{name}; };
+    record_info('IMGV NOT-FOUND', "Cannot find image-version $name. Need to upload it.\n$@") if ($@);
     return $image;
 }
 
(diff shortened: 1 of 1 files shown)


lib/publiccloud/azure.pm
```diff
@@ -72,14 +73,24 @@ sub find_img {
...
+          '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"',
+        proceed_on_failure => 1);
+    record_info('BLOB INFO', $json);
+    eval {
+        $md5 = decode_azure_json($json)->{md5};
+    };
+    record_info('BLOB NOT-FOUND', "Cannot find blob $name. Need to upload it.\n$@") if ($@);
+
+    if (!$@ && (!$md5 || length($md5) != 32)) {
+        record_soft_failure("The blob $name does not have 32 md5 field of 32 characters.");
```
ricardobranco777, line 88:
```suggestion
    if (!$@ && $md5 !~ /^[a-fA-F0-9]{32}$/) {
        record_soft_failure("The blob $name does not have valid MD5 field");
```

ricardobranco777 (approved):
Just a small suggestion. Otherwise LGTM.

lib/publiccloud/azure.pm
```diff
@@ -72,14 +73,24 @@ sub find_img {
...
-    my $json = script_output("az image show --resource-group " . $self->resource_group . " --name $name", 60, proceed_on_failure => 1);
+    $json = script_output("az storage blob show --account-key $key -o json " .
+          "--container-name '$container' --account-name '$storage_account' --name '$name' " .
+          '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"',
+        proceed_on_failure => 1);
+    record_info('BLOB INFO', $json);
+    eval {
+        $md5 = decode_azure_json($json)->{md5};
+    };
+    record_info('BLOB NOT-FOUND', "Cannot find blob $name. Need to upload it.\n$@") if ($@);
```
grisu48, line 85: AFAICS this information is shown, also if the blob would exist, no?

pdostal: If blob exists there is no error in `eval` and therefore the `$@` is not present.

grisu48: Oh, there is this tiny little `if ($@);` at the end.

I'm adding this to the list of reasons why this is syntactic sin perls commits on a daily basis :facepalm:

asmorodskyi, line 85:
```
if ($@) {
record_info('BLOB NOT-FOUND', "Cannot find blob $name. Need to upload it.\n$@") 
}
elif (!defined($md5) || $md5 !~ /^[a-fA-F0-9]{32}$/) {
        record_info('INVALID', "The blob $name does not have valid md5 field.");
        return $image;
 }
```

asmorodskyi: two reasons : 
1. IMHO `!$md5` will cause "Use of uninitialized variable " perl exceptions in code which are quite annoying and distinctive when you parsing test log with hope to find root cause of the problem 
2. you have 2 if's one for "$@" and another for "!@$" IMHO it is looks better in single if/else

lib/publiccloud/azure.pm
```diff
@@ -150,32 +162,18 @@ sub upload_img {
...
+    assert_script_run('az storage blob upload --max-connections 4 --type page'
+          . " --account-name '$storage_account' --account-key '$key' --container-name '$container'"
+          . " --file '$file' --name '$img_name' --tags '$tags'", timeout => 60 * 60 * 2);
+    # After blob is uploaded we save the of it as its metadata.
+    # This is also to verify that the upload has been finished.
     my $file_md5 = script_output("md5sum $file | cut -d' ' -f1", timeout => 240);
-
-    # Check if blob already exists
-    my $container = $self->container;
-    my $blobs = script_output("az storage blob list --account-key $key --container-name '$container' --account-name '$storage_account' --query '[].name' -o tsv");
```
asmorodskyi, line 163: wait you fully drop listing of blobs ? can you elaborate on this ?

asmorodskyi (approved):
LGTM

lib/publiccloud/azure.pm
```diff
@@ -72,18 +73,45 @@ sub find_img {
...
     };
-    record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    if ($@) {
+        record_info('BLOB NOT-FOUND', "Cannot find blob $name. Need to upload it.\n$@");
+    } elsif (!$md5 || $md5 !~ /^[a-fA-F0-9]{32}$/) {
+        record_info('INVALID', "The blob $name does not have valid md5 field.");
+        return $image;
+    }
+
+    my $arch = (check_var('PUBLIC_CLOUD_ARCH', 'arm64')) ? 'Arm64' : 'x64';
```
asmorodskyi, line 92: ha funny , never noticed that before . I have seen "x86" and "x86_64" and "x86_32" but I never seen "x64" Microsoft is really doing everything in its own way ...

lib/publiccloud/azure.pm
```diff
@@ -72,18 +73,45 @@ sub find_img {
...
+    if ($arch eq 'Arm64') {
+        my $resource_group = $self->resource_group;
+        my $gallery = $self->image_gallery;
+        my $version = calc_img_version();
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        # asdf
+        $json = script_output("az sig image-version show --resource-group '$resource_group' --gallery-name '$gallery' " .
+              "--gallery-image-definition '$definition' --gallery-image-version '$version'", timeout => 60 * 30);
+        record_info('IMGV INFO', $json);
```
asmorodskyi, line 102: this lines : 
```
record_info('IMGV INFO', $json);
        eval {
            $image = decode_azure_json($json)->{name};
        };
        record_info('IMGV NOT-FOUND', "Cannot find image-version $name. Need to upload it.\n$@") if ($@);
```
are fully identical in if/else hence can be moved out from if/else

lib/publiccloud/azure.pm
asmorodskyi: ofc. this comment does not require any reaction :)

lib/publiccloud/azure.pm
```diff
@@ -212,6 +227,7 @@ sub upload_img {
         assert_script_run("az sig image-definition create --resource-group '$resource_group' --gallery-name '$gallery' " .
               "--gallery-image-definition '$definition' --os-type Linux --publisher '$publisher' --offer '$offer' --sku '$sku' " .
               "--architecture '$arch' --hyper-v-generation '$hyperv' --os-state 'Generalized'", timeout => 300);
+        # asdf
```
asmorodskyi, line 230: please remove this

asmorodskyi:
except duplicated code in if/else and mysterious comment all looks fine

lib/publiccloud/azure.pm
```diff
@@ -72,18 +73,45 @@ sub find_img {
...
+          "--container-name '$container' --account-name '$storage_account' --name '$name' " .
+          '--query="{name: name,createTime: properties.creationTime,md5: properties.contentSettings.contentMd5}"',
+        proceed_on_failure => 1);
+    record_info('BLOB INFO', $json);
     eval {
-        $image = decode_azure_json($json)->{name};
+        $md5 = decode_azure_json($json)->{md5};
     };
-    record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
+    if ($@) {
```
grisu48, line 85: Is this correct? Isn't `if ($@)` being executed, if in the previous `eval` block the json could be parsed?

grisu48 (commented):
One comment.

I think this can work, but it's difficult to test and to see, if not something breaks. We need to have a close look at possible fallout when merging. Let's do this at the beginning of the week!

lib/publiccloud/azure.pm
pdostal: See https://perldoc.perl.org/functions/eval
> If there is a syntax error or runtime error, or a [die](https://perldoc.perl.org/functions/die) statement is executed, eval returns [undef](https://perldoc.perl.org/functions/undef) in scalar context, or an empty list in list context, and [$@](https://perldoc.perl.org/perlvar#%24%40) is set to the error message. (Prior to 5.16, a bug caused [undef](https://perldoc.perl.org/functions/undef) to be returned in list context for syntax errors, but not for runtime errors.) If there was no error, [$@](https://perldoc.perl.org/perlvar#%24%40) is set to the empty string. A control flow operator like [last](https://perldoc.perl.org/functions/last) or [goto](https://perldoc.perl.org/functions/goto) can bypass the setting of [$@](https://perldoc.perl.org/perlvar#%24%40). Beware that using eval neither silences Perl from printing warnings to STDERR, nor does it stuff the text of warning messages into [$@](https://perldoc.perl.org/perlvar#%24%40). To do either of those, you have to use the [$SIG{__WARN__}](https://perldoc.perl.org/perlvar#%25SIG) facility, or turn off warnings inside the BLOCK or EXPR using no warnings 'all'. See [warn](https://perldoc.perl.org/functions/warn), [perlvar](https://perldoc.perl.org/perlvar), and [warnings](https://perldoc.perl.org/warnings).

grisu48: Ah, that's the error message, not the return code. Yep, then it works.

grisu48 (approved):

lib/publiccloud/azure.pm
```diff
@@ -72,18 +73,45 @@ sub find_img {
...
+    }
+
+    my $arch = (check_var('PUBLIC_CLOUD_ARCH', 'arm64')) ? 'Arm64' : 'x64';
+    if ($arch eq 'Arm64') {
+        my $resource_group = $self->resource_group;
+        my $gallery = $self->image_gallery;
+        my $version = calc_img_version();
+        my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION');
+        $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+        # asdf
```
grisu48, line 99: Please remove this

lib/publiccloud/azure.pm
pdostal: I don't understand, can you please help me out?

pdostal:
I should be back the next week to finally merge this.

lib/publiccloud/azure.pm
```diff
@@ -141,6 +162,7 @@ sub upload_img {
     $img_name =~ s/\.vhdfixed/-$sku.vhd/;
     my $disk_name = $img_name;
     my $storage_account = get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'eisleqaopenqa');
+    my $container = $self->container;
```
mpagot, line 165: why a dedicated variable is needed ? Why not to directly use `$self->container` ?

lib/publiccloud/azure.pm
```diff
@@ -150,32 +172,18 @@ sub upload_img {
...
     $self->create_resources($storage_account) if (!$rg_exist);
 
     my $key = $self->get_storage_account_keys($storage_account);
 
-
+    # Note: VM images need to be a page blob type
+    assert_script_run('az storage blob upload --max-connections 4 --type page'
+          . " --account-name '$storage_account' --account-key '$key' --container-name '$container'"
+          . " --file '$file' --name '$img_name' --tags '$tags'", timeout => 60 * 60 * 2);
+    # After blob is uploaded we save the of it as its metadata.
```
mpagot, line 183: `we save the of it as its metadata` 

should be

`we save the MD5 of it as its metadata`

lib/publiccloud/azure.pm
asmorodskyi: lines 99,100,101 within `if ($arch eq 'Arm64')`  and lines 104,105,106 within `else` branch are identical so no matter what arch is we will do this , so there is no point to keep the duplicated inside each branch we can just call this AFTER `if/esle` : 

```
} else {
        $json = script_output("az image show --resource-group " . $self->resource_group . " --name $name", 60, proceed_on_failure => 1);
}

record_info('IMG INFO', $json);
eval { $image = decode_azure_json($json)->{name}; };
record_info('IMG NOT-FOUND', "Cannot find image $name. Need to upload it.\n$@") if ($@);
```

mpagot (approved):
LGTM

PR #16887, opened 2023-04-18
Remove unused needle & update record_info syntax

realcharmer:
Removing unused needle. The syntax of `record_info` has also been changed to be the same as the rest in the entire document.

- Related ticket: https://progress.opensuse.org/issues/127814
- Verification run: https://openqa.suse.de/tests/10937670

Since no significant logic changes were made, one test run is sufficient (only checking syntax, not test itself).


ilmanzo:
LGTM

PR #16889, opened 2023-04-18
Add container schedule for SLEM on PC

grisu48:
Adds code to allow scheduling of container test runs for SLEM on PC.

- Related ticket: https://progress.opensuse.org/issues/110791
- Verification run: [podman](https://duck-norris.qe.suse.de/tests/12581) | [docker](https://duck-norris.qe.suse.de/tests/12583)

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -14,7 +14,7 @@ use Exporter;
 use main_common;
 use main_containers qw(load_container_tests is_container_test);
 use testapi qw(check_var get_required_var get_var set_var);
-use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp is_staging is_released is_transactional is_rt);
+use version_utils;
 use utils;
 use Utils::Architectures;
 use Utils::Backends;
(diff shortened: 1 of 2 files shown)


lib/main_micro_alp.pm
```diff
@@ -14,7 +14,7 @@ use Exporter;
 use main_common;
 use main_containers qw(load_container_tests is_container_test);
 use testapi qw(check_var get_required_var get_var set_var);
-use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp is_staging is_released is_transactional is_rt);
+use version_utils qw(is_microos is_sle_micro is_leap_micro is_alp is_staging is_released is_transactional is_rt is_public_cloud);
```
asmorodskyi, line 17: just my IMHO not necessary to follow - when list getting so big it is better just to do 
```suggestion
use version_utils;
```

lib/main_micro_alp.pm
```diff
@@ -266,7 +266,30 @@ sub load_journal_check_tests {
...
+    my $args = OpenQA::Test::RunArgs->new();
+
+    loadtest("boot/boot_to_desktop");
+    loadtest("publiccloud/prepare_instance", run_args => $args);
+    loadtest("publiccloud/registration", run_args => $args);
+    loadtest("publiccloud/ssh_interactive_start", run_args => $args);
+    loadtest("publiccloud/instance_overview", run_args => $args);
+    loadtest("publiccloud/slem_prepare", run_args => $args);
+
+    if (get_var("PUBLIC_CLOUD_CONTAINERS")) {
```
jlausuch, line 279: I would turn this block into single line:
`load_container_tests() if is_container_test`

`load_container_tests()` will fail if `CONTAINER_RUNTIME` is not defined.

grisu48: I would agree with you, if we would not need the `return 1` in the same block.

grisu48: Sorry, I was looking at the wrong block. But still, silently not executing the tests is IMHO worse than immediately dying with an error message.

Then again, those error messages are anyways replaced by the generic "something something went wrong in main.pm, please check syntax", so I guess in the end it doesn't make a difference.

Long story short: Will amend your suggestion.

grisu48: Amended, new VR: https://duck-norris.qe.suse.de/tests/12588

jlausuch (approved):

pdostal (approved):

grisu48:
Marking as draft while checking, why the registration is done with `SUSEConnect`: https://duck-norris.qe.suse.de/tests/12581#step/registration/17

asmorodskyi:
> Marking as draft while checking, why the registration is done with `SUSEConnect`: https://duck-norris.qe.suse.de/tests/12581#step/registration/17

I think registercloudguest can not work on ro FS

PR #16890, opened 2023-04-18
Public Cloud wait_for_ssh routine refactored and system-up check updated

m-dati:
Public cloud `wait_for_ssh` routine code has been refactored and updated.

The changes topics are:
* routine structure improved,split in 3 blocks,ssh-port check,system-up check,result report;
* system-up check improved using `systemctl is-system-running`, more reliable than journal inspection;
* report on final status improved, also returned when `die(croak)`, helping in debugging;
* description in head section improved, parameters meaning and usage detailed;
* start-up phases detailed in inner loop, providing precise details on the status detected;
* new option `systemup_check`, to check on-demand ssh-open only and skip system-up check;
* new option `wait_stop`, to wait until ssh-closed,instead of open,improving shutdown/reboot checks;
* new option `logs`, _journalctl_ uploaded to test logs, instead of printed on ouptput, _logs=0_ to skip;
* softreboot routine updated to use `wait_stop` option.
* slem_basic last SElinux reboot allowed to retry loop.
* create_instance error handling args added for eventual wait_for_ssh call setting.

Related ticket: poo [126218](https://progress.opensuse.org/issues/126218) 
Needles: no change
Verification run: see VRs in the next posts below.

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -334,73 +334,147 @@ sub wait_for_guestregister {
 
 =head2 wait_for_ssh
 
-    wait_for_ssh([timeout => 600] [, proceed_on_failure => 0])
+    wait_for_ssh([timeout => 600] [, proceed_on_failure => 0] [, ...])
 
-Check if the SSH port of the instance is reachable and open.
+When a remote pc instance starting, by default wait_stop param.=0(false) and 
+this routine checks until the SSH port of the remote instance is reachable and open. 
+Then by default also checks that system is up, unless systemup_check false/0.
+
+Wnen a remote pc instance is stopping in shutdown, we set input param. wait_stop=1(true),
+to expect until ssh is closed; automatic defaults systemup_check=0 and proceed_on_failure=1 applied.
+Status values of exit_code: 0 = pass; 1 = fail; 2 = fail,but on-demand can retry,till timeout.
+
+Parameters:
+ timeout => total wait timeout; default: 600.
+ wait_stop => If true waits for ssh port to become unreachable, if false waits for ssh reachable; default: false.
+ ignore_wrong_pubkey => for eventual publickey issue or unpredicted 'else' sysout, if true, lets retry in loop, false let routine fail.
+ proceed_on_failure => in case of fail, if false exit test with error, if true let calling code to continue; default: wait_stop.
+ username => default: username().
+ public_ip => default: public_ip().
+ systemup_check => If true, checks if the system is up too, instead of just checking the ssh port; default: !wait_stop.
+ logs => If true, upload journal to test logs, if false log not uploaded, to speed up check; default: true.
+
+Return:
+ duration if pass 
+ undef if fail and proceed_on_failure true, otherwise die.
 =cut
 
 sub wait_for_ssh {
     my ($self, %args) = @_;
+    # Input parameters, see description in above head2 - Parameters section:
     $args{timeout} //= 600;
-    $args{proceed_on_failure} //= 0;
+    $args{wait_stop} //= 0;
+    $args{proceed_on_failure} //= $args{wait_stop};
+    $args{systemup_check} //= not $args{wait_stop};
+    $args{logs} //= 1;
+    $args{public_ip} //= $self->public_ip();
     # DMS migration (tests/publiccloud/migration.pm) is running under user "migration"
-    # until it is not over we will recieve "ssh permission denied (pubkey)" error
-    # but it is not good reason to die early because after it will be over DMS will return normal
-    # user and error will be resolved.
+    # until it is not over we will receive "ssh permission denied (pubkey)" error
+    # but it is not good reason to die early because after it will be over
+    # DMS will return normal user and error will be resolved.
     $args{ignore_wrong_pubkey} //= 0;
     $args{username} //= $self->username();
-    $args{public_ip} //= $self->public_ip();
+    my $delay = $args{ignore_wrong_pubkey} ? 20 : 1;
     my $start_time = time();
-    my $check_port = 1;
-    my $sleep_period = $args{ignore_wrong_pubkey} ? 20 : 1;
+    my $instance_msg = "instance: $self->{instance_id}, public IP: $self->{public_ip}";
+    my ($duration, $exit_code, $sshout, $sysout);
 
-    # Looping until reaching timeout or passing two conditions :
-    # - SSH port 22 is reachable
-    # - journalctl got message about reaching one of certain targets
-    while ((my $duration = time() - $start_time) < $args{timeout}) {
-        if ($check_port) {
-            $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
-        }
-        else {
-            # On boottime test we do hard reboot which may change the instance address
-            script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
+    # Looping until SSH port 22 is reachable or timeout.
+    while (($duration = time() - $start_time) < $args{timeout}) {
+        $exit_code = script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1);
+        last if (isok($exit_code) and not $args{wait_stop});    # ssh port open ok
+        last if (not isok($exit_code) and $args{wait_stop});    # ssh port closed ok
+        sleep 1;
+    }    # endloop
 
-            my $output = $self->run_ssh_command(
-                cmd => 'sudo journalctl -b | grep -E "Reached target (Cloud-init|Default|Main User Target)"',
-                proceed_on_failure => 1,
-                username => $args{username});
-            if ($output =~ m/Reached target.*/) {
-                return $duration;
+    # exit_code is 0 when shell script is ok
+    if (isok($exit_code)) {
+        $sshout = "SSH port is open\n";
+    }
+    else {
+        $sshout = "SSH port is not open failed access\n";
+        $sshout .= "as expected by stopping: OK.\n" if $args{wait_stop};
+    }    # endif
+
+    # check also remote system is up and running:
+    if ($args{systemup_check} and isok($exit_code)) {
+        script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
+          if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME') or get_var('PUBLIC_CLOUD_CHECK_REBOOT'));
+        while (($duration = time() - $start_time) < $args{timeout}) {
+            # On boottime test we do hard reboot which may change the instance address:
+            # timeout recalculated removing consumed time until now
+            $sysout = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running',
+                timeout => $args{timeout} - $duration, proceed_on_failure => 1, username => $args{username});
+            # result check
+            if ($sysout =~ m/Permission denied \(publickey\).*/) {
+                $exit_code = 2;
+                last unless $args{ignore_wrong_pubkey};    # ondemand retry
             }
-            elsif ($output =~ m/Permission denied \(publickey\).*/) {
-                die "ssh permission denied (pubkey)" unless $args{ignore_wrong_pubkey};
+            elsif ($sysout =~ m/initializing|starting/) {    # still starting
+                $exit_code = undef;
             }
-        }
-        sleep $sleep_period;
-    }
+            elsif ($sysout =~ m/running/) {    # startup OK
+                $exit_code = 0;
+                $sysout .= "\nSystem successfully booted";
+                last;
+            }
+            elsif ($sysout =~ m/degraded/) {    # up but with failed services to collect
+                $exit_code = 0;
+                $sysout .= "\nSystem booted, but some services failed:\n" .
+                  $self->ssh_script_output(cmd => 'sudo systemctl --failed',
+                    proceed_on_failure => 1, username => $args{username});
+                last;
+            }
+            elsif ($sysout =~ m/maintenance|stopping|offline|unknown/) {
+                $exit_code = 1;
+                $sysout .= "\nCan not reach systemd target";
+                last;
+            }
+            else {    # FAIL: Connection refused or else
+                $exit_code = 2;
+                $sysout .= "\nCan not reach systemd target";
+                last unless ($args{proceed_on_failure} or $args{ignore_wrong_pubkey});    # ondemand retry until timeout
+            }    # endif
+            sleep $delay;
+        }    # end loop
 
-    script_run("ssh  -i /root/.ssh/id_rsa $args{username}\@$args{public_ip} true", timeout => 360);
-    # Debug output: We have occasional error in 'journalctl -b' - see poo#96464 - this will be removed soon.
-    # Exclude 'mr_test/saptune' test case as it will introduce random softreboot failures.
-    if (!get_var('PUBLIC_CLOUD_SLES4SAP')) {
-        $self->run_ssh_command(cmd => 'sudo journalctl -b', proceed_on_failure => 1, username => $args{username}, timeout => 360);
-    }
+        # Log upload
+        if (!get_var('PUBLIC_CLOUD_SLES4SAP') and $args{logs}) {
+            #Exclude 'mr_test/saptune' test case as it will introduce random softreboot failures.
+            $self->ssh_script_run('sudo journalctl -b --no-pager > /tmp/journalctl.log',
+                timeout => 360, proceed_on_failure => 1, username => $args{username}, quiet => 1);
+            $self->upload_log('/tmp/journalctl.log', failok => 1);
+        }    # endif
+    }    # endif
 
-    unless ($args{proceed_on_failure}) {
-        my $error_msg;
-        if ($check_port) {
-            $error_msg = sprintf("Unable to reach SSH port of instance %s with public IP:%s within %d seconds", $self->{instance_id}, $self->{public_ip},
-                $args{timeout});
-        }
-        else {
-            $error_msg = sprintf("Can not reach systemd target on instance %s with public IP:%s within %d seconds",
-                $self->{instance_id}, $self->{public_ip}, $args{timeout});
-        }
-        croak($error_msg);
-    }
+    # result display
+    $sysout .= "\nTimeout $args{timeout} sec. expired" if ($duration >= $args{timeout});
+    $instance_msg = "Check" . ($args{systemup_check} ? " SYSTEM " : " SSH ") . ($args{wait_stop} ? "DOWN" : "UP") .
+      ", $instance_msg, Duration: $duration sec.\nResult: $sshout . $sysout";
+    record_info("WAIT CHECK", $instance_msg);
+    # OK
+    return $duration if (isok($exit_code) and not $args{wait_stop});
+    # FAIL
+    croak($sshout . $sysout) unless ($args{proceed_on_failure});
+    return;    # proceed_on_failure true
+}    # end sub
 
-    return;
-}
+=head2 isok
+
+    isok($exit_code);
+
+To convert in a true or 1 value for perl tests the exit_code 0 of shell script ok.
+
+Return:
+the positive test status of a shell exit code, 
+that is true(1) and ok when its value is defined and zero: C<$x == 0>
+otherwise false(undef).
+=cut
+
+sub isok {
+    my ($x) = @_;
+    return (defined($x) and $x == 0);
+}    # end sub
 
 =head2 softreboot
 
@@ -433,9 +507,11 @@ sub softreboot {
     my $start_time = time();
 
     # wait till ssh disappear
-    while (($duration = time() - $start_time) < $args{timeout}) {
-        last unless (defined($self->wait_for_ssh(timeout => 1, proceed_on_failure => 1, username => $args{username})));
-    }
+    my $out = $self->wait_for_ssh(timeout => $args{timeout}, wait_stop => 1, username => $args{username});
+    # ok ssh port closed
+    record_info("Shutdown failed", "WARNING: while stopping the system, ssh port still open after timeout,\nreporting: $out")
+      if (defined $out);    # not ok port still open
+
     my $shutdown_time = time() - $start_time;
     die("Waiting for system down failed!") unless ($shutdown_time < $args{timeout});
     my $bootup_time = $self->wait_for_ssh(timeout => $args{timeout} - $shutdown_time,
@@ -456,6 +532,7 @@ sub softreboot {
     return ($shutdown_time, $bootup_time);
 }
 
+
 =head2 stop
 
     stop();
@@ -472,8 +549,8 @@ sub stop {
 
     start([timeout => ?]);
 
-Start the instance and check SSH connectivity. Return the number of seconds
-till the SSH port was available.
+Start the instance and wait for the system to be up.
+Returns the number of seconds till the system up and running.
 =cut
 
 sub start {
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -298,6 +298,9 @@ publiccloud::instance objects.
 C<image>         defines the image_id to create the instance.
 C<instance_type> defines the flavor of the instance. If not specified, it will load it
                      from PUBLIC_CLOUD_INSTANCE_TYPE.
+C<timeout>             Parameter to pass to instance::wait_for_ssh.
+C<ignore_wrong_pubkey> Same as timeout.
+C<proceed_on_failure>  Same as timeout.
 
 =cut
 
@@ -310,7 +313,9 @@ sub create_instances {
     foreach my $instance (@vms) {
         record_info("INSTANCE", $instance->{instance_id});
         if ($args{check_connectivity}) {
-            $instance->wait_for_ssh();
+            $instance->wait_for_ssh(timeout => $args{timeout},
+                ignore_wrong_pubkey => $args{ignore_wrong_pubkey},
+                proceed_on_failure => $args{proceed_on_failure});
             # Install server's ssh publicckeys to prevent authenticity interactions
             assert_script_run(sprintf('ssh-keyscan %s >> ~/.ssh/known_hosts', $instance->public_ip));
         }
--- a/tests/publiccloud/slem_basic.pm
+++ b/tests/publiccloud/slem_basic.pm
@@ -71,7 +71,7 @@ sub run {
     $instance->softreboot();
     $instance->run_ssh_command(cmd => 'sudo sestatus | grep disabled');
     $instance->run_ssh_command(cmd => 'sudo transactional-update -n setup-selinux');
-    $instance->softreboot();
+    $instance->softreboot(ignore_wrong_pubkey => 1);
 
     # SElinux and logging tests
     $instance->run_ssh_command(cmd => 'sudo sestatus | grep enabled');


pdostal:
@m-dati you already opened 3 pull requests for one feature. Is this really needed?

m-dati:
> @m-dati you already opened 3 pull requests for one feature. Is this really needed?

This last PR is the synthesis of all the work done in the previous PRs, where I studied the behaviour of 3 check types all together, sharing the results and outcomes.

So, after last results I had in the previous PR/branch, in this last branch I'm including the only proposed check-command, I removed the other 2 commands, so the code is different and shorter than the previous branch and ready for kind review, discussion and eventual merge.

lib/publiccloud/instance.pm
```diff
@@ -364,28 +366,40 @@ sub wait_for_ssh {
...
+            }
+            elsif ($out =~ m/degraded/) {
+                my $failed = $self->run_ssh_command(cmd => 'sudo systemctl --failed', proceed_on_failure => 1, username => $args{username});
+                record_info("CHECK SSH", 'wait ssh - system degraded, ' . (time() - $start_time) . 's, System:' . "\n" . $failed);
+                $self->journal_upload("/tmp/waitssh_journal.txt", username => $args{username});
                 return $duration;
             }
-            elsif ($output =~ m/Permission denied \(publickey\).*/) {
+            elsif ($out =~ m/maintenance|stopping|offline|unknown/) {
+                # check if system status ko
```
asmorodskyi, line 384: maybe want to log some record info here with what exactly status was ?

lib/publiccloud/instance.pm
```diff
@@ -401,6 +415,24 @@ sub wait_for_ssh {
...
+    my $com = 'sudo journalctl -b --no-pager > ' . $log;
+    eval {
+        return unless ($log);
+        if (!get_var('PUBLIC_CLOUD_SLES4SAP')) {
+            $self->run_ssh_command(cmd => $com, proceed_on_failure => 1, username => $usr, timeout => 360);
+            $self->upload_log($log, failok => 1);
+        }
+    };
+    if ($@) {
+        print "ERR: $@\n";
```
asmorodskyi, line 432: there are native openQA's ways to log something into openQA logs please use them instead print 
```suggestion
        bmwqemu::diag("ERR: $@\n");
```

asmorodskyi:
generally looks good ! I like this version

asmorodskyi:
> I does not make sense to me.
> 
> The `sudo systemctl is-system-running --wait` command is **waiting** for boot to be somehow finished, so it should not be inside the `while` loop.
> 
> What I imagine is something like:
> 
> ```perl
> my $duration = time();
> script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout});
> my $boot = $self->ssh_script_run('sudo systemctl is-system-running --wait`, timeout => $args{timeout});
> $self->ssh_script_run('sudo journalctl -b --no-pager > /tmp/boot.log');
> $self->upload_log('/tmp/boot.log', failok => 1);
> die('The system didn't boot correctly, please check the logs") if ($boot != 0);
> ```

mmm and what about  : 
```
if ($out =~ 'running') {
}
elif ($out =~ '..........') {
}
else {
}
```
is it fully drop from your suggestion ? or where will it go ? ( IMHO it is needed part ) 

But yeah I agree that while loop does not match with `--wait` flag .... but also we can just drop wait flag and keep the loop ?

pdostal:
Yes, it's either fully booted (return code `0`) or degraded (other return code) or the `sudo systemctl is-system-running --wait` would be still running (in which case we can kill it using `timeout` command. See [man](https://manpages.debian.org/stretch/systemd/systemctl.1.en.html#System_Commands) page.

m-dati:
Than you much for your comments @pdostal @asmorodskyi.
I reply here:

About the [loop topic](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16890#issuecomment-1515340272), indeed this _check type_, as explained in the [poo note-5](https://progress.opensuse.org/issues/126218#note-5) item 1 and documentation, by using `--wait` option, it is quite wating until one of these 3 possible final states occurr (exclusively):
1) the boot is ok system _running_ exit=0, or
2) it resulted _degraded_, exit>0, meaning that some services didn't start, or
3) some wrong states occurred, exit>0.

So, we can avoid that `while` loop and use the timeout for unsuccessful cases, I will update the code accordingly.  

But about the item n.2, as I explained in the previous PR, it could still pass and allow the test to proceed and the eventual impact be analyzed afterward in the test. Therefore here the ouptut is intercepted as _string_ and not as assertion on exit status number otherwise using `ssh_assert_script_run` only n.1 pass and n.2 and n.3 would be terminated.

lib/publiccloud/instance.pm
m-dati: ok thank you, wiil update it

lib/publiccloud/instance.pm
m-dati: The line checking `/Permission denied.../`, inherited by the previous code, indeed it could occurr when some ssh issue could prevent the access to the remote session, therefore I left it there.

m-dati: Instead the line checking the `maintenance|...` is the 3rd status possible, resuting after the `--wait`, below discussed.

m-dati:
Kind note: 
I noted that the function `run_ssh_command` _could be_ used in place of the other `ssh_script...`, as a single unified function to have either return _code_ or _stdout_ (switching by `rc_only`), if I/we could apply few code fixes, basically in the `timeout == 0` case. 
But this would require much more checks and VRs, being `run_ssh_command` widely used in many other modules. It could be interesting a separated discussion.

lib/publiccloud/instance.pm
asmorodskyi: ok

asmorodskyi:
> Kind note: I noted that the function `run_ssh_command` _could be_ used in place of the other `ssh_script_`_run_ or _output_ functions, as a single unified function to have either return _code_ or _stdout_ (switching `rc_only`), if we could apply few code fixes, basically in the `timeout == 0` case. Evenmore we could unify _assert_script_run_ there.
> 
> But this would require much more checks and VRs, being `run_ssh_command` widely used in many other modules. It could be interesting a separated discussion.

Yes sure indeed every call of `ssh_script_run` , `ssh_script_output` , `ssh_assert_script_run` can be replaced with `run_ssh_command` .  But I would recommend keep using `ssh_script_run` , `ssh_script_output` , `ssh_assert_script_run` . On other  hand I can not come up with strong arguments WHY , probably it is just matter of taste ,   @grisu48  you as author of 3 functions mentioned above might want to comment on that ? Do you have strong arguments against `run_ssh_command` ?

lib/publiccloud/instance.pm
```diff
@@ -350,55 +350,50 @@ sub wait_for_ssh {
     $args{username} //= $self->username();
     $args{public_ip} //= $self->public_ip();
     my $start_time = time();
-    my $check_port = 1;
     my $sleep_period = $args{ignore_wrong_pubkey} ? 20 : 1;
-
+    my $retry = ($args{timeout} > $sleep_period) ? int($args{timeout} / $sleep_period) : 1;
+    my $xaccess;
```
asmorodskyi, line 355:
```suggestion
    my ($xaccess, $out);
```

lib/publiccloud/instance.pm
```diff
@@ -350,55 +350,50 @@ sub wait_for_ssh {
...
+    $xaccess = ssh_script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry);
+    if ($xaccess == 0) {
+        # helps in boottime tests,when hard reboot could change the instance address:
+        ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
+        # check system is running:
+        $out = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running --wait', proceed_on_failure => 1, username => $args{username});
+        if ($out =~ m/running/) {    # startup ok
+            $status = 'OK';
+        }
+        elsif ($out =~ m/degraded/) {    # failed services colletion
```
asmorodskyi, line 370:
```suggestion
        elsif ($out =~ m/degraded/) {    # failed services collection
```

lib/publiccloud/instance.pm
```diff
@@ -350,55 +350,50 @@ sub wait_for_ssh {
...
     # - SSH port 22 is reachable
-    # - journalctl got message about reaching one of certain targets
-    while ((my $duration = time() - $start_time) < $args{timeout}) {
-        if ($check_port) {
-            $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
+    # - is-system-running ok
+    $xaccess = ssh_script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry);
+    if ($xaccess == 0) {
+        # helps in boottime tests,when hard reboot could change the instance address:
+        ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
```
asmorodskyi, line 364: wait but if we got here there is no problem with ssh keys , isn't this call needs to be done in `else` branch ?

lib/publiccloud/instance.pm
```diff
@@ -350,55 +350,50 @@ sub wait_for_ssh {
...
-            $error_msg = sprintf("Unable to reach SSH port of instance %s with public IP:%s within %d seconds", $self->{instance_id}, $self->{public_ip},
-                $args{timeout});
-        }
-        else {
-            $error_msg = sprintf("Can not reach systemd target on instance %s with public IP:%s within %d seconds",
-                $self->{instance_id}, $self->{public_ip}, $args{timeout});
-        }
-        croak($error_msg);
+    # result display
+    record_info("CHECK SSH", 'wait ssh ' . (time() - $start_time) . 's, System:' . $out . "\n" . $status);
```
asmorodskyi, line 381: mmm I assume this line should hardly depend on what we have in `$out` and `$status`  , otherwise we will get `Use of uninitialized variable` perl's exceptions . e.g. when `$xaccess` is non-zero value `$out` will be undefined . I think we need to move this string generation inside `if($xaccess == 0){}else{}` and in each certain case generate different content for this string with use of different variables . e.g. : 
```
my $check_ssh_msg;
if($xaccess == 0)
{
......
 $check_ssh_msg = 'wait ssh ' . (time() - $start_time) . 's, System:' . $out . "\n" . $status;
}
else
{
 $check_ssh_msg = 'Failed to reach port'
}
record_info("CHECK SSH", $check_ssh_msg);

**Comment by asmorodskyi** [comment]
nice work @m-dati  I like direction where this PR is going but there is still some work needs to be done

**Comment by asmorodskyi** [comment]
@mpagot  , @lilyeyes  this might affect you also , what VRs @m-dati  can do to check your part is working ok ?

File: lib/publiccloud/instance.pm (hunk 5 above)
Comment (m-dati) [reply lib/publiccloud/instance.pm]: This command is in the ok branch of [master version](https://github.com/m-dati/os-autoinst-distri-opensuse/blob/8dcf744faa8c124b55b217378a3bf505bb09f126/lib/publiccloud/instance.pm#L365), probably due to some old test IP-changed issue

File: lib/publiccloud/instance.pm (hunk 6 above)
Comment (m-dati) [reply lib/publiccloud/instance.pm:386]: I have updated the code in order to have always a coherent report

File: lib/publiccloud/instance.pm (hunk 5 above)
Comment (asmorodskyi) [reply lib/publiccloud/instance.pm]: ok than

File: lib/publiccloud/instance.pm (hunk 6 above)
Comment (asmorodskyi) [reply lib/publiccloud/instance.pm:386]: please push this changes

Comment (m-dati) [reply lib/publiccloud/instance.pm]: pushed a more fix. I also added `$info` var too, but as it is now, we could also use only `out` and `status`.

**Comment by m-dati** [comment]
VR
https://openqa.suse.de/tests/10964972
https://openqa.suse.de/tests/10964966
https://openqa.suse.de/tests/10964940

**Comment by m-dati** [comment]
New VRs:
https://openqa.suse.de/tests/10965397
https://openqa.suse.de/tests/10965400
https://openqa.suse.de/tests/10965398
https://openqa.suse.de/tests/10965399

**Comment by m-dati** [comment]
I noted that here using `script_retry`,  each `nc` [retry](https://openqa.suse.de/tests/10965397#step/prepare_instance/90) is dispalyed in the test Details boxes, while in the original code with a local loop, those attempts only in the serial log were present.

**Comment by lilyeyes** [comment]
> @mpagot , @lilyeyes this might affect you also , what VRs @m-dati can do to check your part is working ok ?

Thanks, https://openqa.suse.de/tests/10959776 this case can be used to verify code changes.

File: lib/publiccloud/instance.pm [hunk 7]
```diff
@@ -350,55 +352,58 @@ sub wait_for_ssh {
...
     # Looping until reaching timeout or passing two conditions :
     # - SSH port 22 is reachable
-    # - journalctl got message about reaching one of certain targets
-    while ((my $duration = time() - $start_time) < $args{timeout}) {
-        if ($check_port) {
-            $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
+    # - is-system-running ok
+    $xaccess = script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry);
+    $duration = time() - $start_time;    # elapsed time until access
+    if (is_exit_ok($xaccess)) {
```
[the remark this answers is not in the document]
pdostal: You would have to add `, die => 0` to `script_retry()` to get it's exit code but I think it's not needed because of above comment.

lib/publiccloud/utils.pm
```diff
@@ -174,6 +175,12 @@ sub is_embargo_update {
     return 0;
 }
 
+# Transforms input result exit-code ok(=0) into boolean value for test ok(=1)
+sub is_exit_ok {
+    my ($ret) = @_;
+    return ($ret == 0) ? 1 : 0;
+}
```
pdostal, line 182: This is interesting idea, but I'd move it to separate PR and to `lib/utils.pm`.

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,58 @@ sub wait_for_ssh {
...
+    my $out = '#';
+    my ($xaccess, $duration);
     # Looping until reaching timeout or passing two conditions :
     # - SSH port 22 is reachable
-    # - journalctl got message about reaching one of certain targets
-    while ((my $duration = time() - $start_time) < $args{timeout}) {
-        if ($check_port) {
-            $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
+    # - is-system-running ok
+    $xaccess = script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry);
```
pdostal, line 363:
```suggestion
    script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry);
```

pdostal:
```suggestion
    script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry, fail_msg => "Something went horribly wrong, the port 22 seems to be dead.");
```

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,58 @@ sub wait_for_ssh {
...
-                $self->{instance_id}, $self->{public_ip}, $args{timeout});
-        }
-        croak($error_msg);
-    }
+    # result display
+    record_info("CHECK SSH", 'wait ssh ' . $duration . 's, System: ' . $out);
 
-    return;
+    # Log upload
+    if (!get_var('PUBLIC_CLOUD_SLES4SAP') && (is_exit_ok($xaccess))) {
```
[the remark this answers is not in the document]
m-dati: Yes, removing the $xaccess check, but still not for SAP due to old comment in L385

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,47 @@ sub wait_for_ssh {
...
+    # Looping until SSH port 22 is reachable or timeout:
+    script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry, fail_msg => "Unable to reach SSH port of " . $instance_msg);
+    # ssh OK:
+    $duration = time() - $start_time;    # time after ssh ok
+
+    # help in boottime tests,when hard reboot could change the instance address:
+    $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
+
+    # check remote system is running, looping by --wait option:
+    $out = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running --wait', ($args{timeout} - $duration), proceed_on_failure => 1, username => $args{username});
```
pdostal, line 370: Just initialise the variables only when they're needed.
```suggestion
    my $out = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running --wait', ($args{timeout} - $duration), proceed_on_failure => 1, username => $args{username});
```

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,47 @@ sub wait_for_ssh {
...
-
-    return;
+    # Log upload
+    if (!get_var('PUBLIC_CLOUD_SLES4SAP')) {
+        #Exclude 'mr_test/saptune' test case as it will introduce random softreboot failures.
+        $self->ssh_script_run('sudo journalctl -b --no-pager > /tmp/journalctl.log', proceed_on_failure => 1, username => $args{username}, timeout => 360);
+        $self->upload_log('/tmp/journalctl.log', failok => 1);
+    }
+    # result display
+    record_info("CHECK SSH", 'wait ssh ' . $duration . 's, System: ' . $out);
```
pdostal, line 390:
```suggestion
    record_info('CHECK SSH', "Duration: $duration seconds\nStatus: $out");
```

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,47 @@ sub wait_for_ssh {
...
-            }
-            elsif ($output =~ m/Permission denied \(publickey\).*/) {
-                die "ssh permission denied (pubkey)" unless $args{ignore_wrong_pubkey};
-            }
-        }
-        sleep $sleep_period;
+    my $retry = ($args{timeout} > $sleep_period) ? int($args{timeout} / $sleep_period) : 1;
+    my $status = 0;
+    my $out = '#';
+    my $instance_msg = sprintf("instance %s with public IP:%s within %d seconds", $self->{instance_id}, $self->{public_ip}, $args{timeout});
```
pdostal, line 359: Maybe you can move this down to the `record_info`?

m-dati: no, needed in lines 362 and 382

pdostal: I feel like there is too many unnecessary variables.

lib/publiccloud/instance.pm
```diff
@@ -350,55 +352,47 @@ sub wait_for_ssh {
...
+    # Looping until SSH port 22 is reachable or timeout:
+    script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22', timeout => $args{timeout}, delay => $sleep_period, retry => $retry, fail_msg => "Unable to reach SSH port of " . $instance_msg);
+    # ssh OK:
+    $duration = time() - $start_time;    # time after ssh ok
+
+    # help in boottime tests,when hard reboot could change the instance address:
+    $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts") if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
+
+    # check remote system is running, looping by --wait option:
+    $out = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running --wait', timeout => ($args{timeout} - $duration), proceed_on_failure => 1, username => $args{username});
```
asmorodskyi, line 370: `timeout => ($args{timeout} - $duration)` - can we get negative value here ? and what would be the behavior of `ssh_script_output`

lib/publiccloud/instance.pm
```diff
@@ -336,6 +336,8 @@ sub wait_for_guestregister {
     wait_for_ssh([timeout => 600] [, proceed_on_failure => 0])
 
 Check if the SSH port of the instance is reachable and open.
+Also check system UP
+
 =cut
 
 sub wait_for_ssh {
```
asmorodskyi, line 343: I would suggest to rename a function so it will be more accurate for what is going on below . we not just "wait for ssh" anymore ...

asmorodskyi:
description of this function needs to be updated due to this PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L475 ( we not returning time needed to port become available but time for whole boot process is over )

m-dati:
> description of this function needs to be updated due to this PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L475 ( we not returning time needed to port become available but time for whole boot process is over )

Should we consider here introduction of any additional parameter in order to use pass either ssh-only or system up too?

lib/publiccloud/instance.pm
m-dati: But renaming this function it'd mean also to change name in many many modules using it now: should we do it the same?

asmorodskyi:
> > description of this function needs to be updated due to this PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L475 ( we not returning time needed to port become available but time for whole boot process is over )
> 
> Asking as hypotesis, @asmorodskyi / all, should we consider here introduction of any additional parameter in order to return time either when ssh-only pass or full system up too? Then decision when/where to use it, up to the process logic.

yeah I had same thoughts , so I think yeah we need such param so user may use this function just to check that port is available

lib/publiccloud/instance.pm
m-dati: ok :) then changing there the name too will help me to take the count of updates

lib/publiccloud/instance.pm
m-dati: Note: the `timeout = 0` usually in our functions **deactivates** the timeout

m-dati:
> > > description of this function needs to be updated due to this PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/publiccloud/instance.pm#L475 ( we not returning time needed to port become available but time for whole boot process is over )
> > 
> > 
> > Asking as hypotesis, @asmorodskyi / all, should we consider here introduction of any additional parameter in order to return time either when ssh-only pass or full system up too? Then decision when/where to use it, up to the process logic.
> 
> yeah I had same thoughts , so I think yeah we need such param so user may use this function just to check that port is available

Indeed moreover, about `softreboot` the [going down](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/7daed94b9452547896aa7025c7fb808436dadfdc/lib/publiccloud/instance.pm#L436) phase would need for `ssh check` only, to _fail_, while the `start` function needs for the full `up`. That is, it should be then udated to use this new option

mpagot (approved):
LGTM

pdostal:
What about separating `wait_for_ssh()` and `wait_for_boot()`?

m-dati:
> What about separating `wait_for_ssh()` and `wait_for_boot()`? @asmorodskyi @m-dati

Personally and from mantaining and using point of view I consider useful do not increase granularity and have a function with buttons to enable/disable similar properties instead to have many separated functions doing "similar" things

pdostal:
> > What about separating `wait_for_ssh()` and `wait_for_boot()`? @asmorodskyi @m-dati
> 
> Personally and from mantaining and using point of view I consider useful do not increase granularity and have a function with buttons to enable/disable similar properties instead to have many separated functions doing "similar" things

One subroutine would be checking for open port and the other for systemd status. Those are not the same things.

For me it's having one complicated subroutine with sevaral variables and conditions versus have two small, clean and well-defined subroutines.

m-dati:
> > > What about separating `wait_for_ssh()` and `wait_for_boot()`? @asmorodskyi @m-dati
> > 
> > 
> > Personally and from mantaining and using point of view I consider useful do not increase granularity and have a function with buttons to enable/disable similar properties instead to have many separated functions doing "similar" things
> 
> One subroutine would be checking for open port and the other for systemd status. Those are not the same things.
> 
> For me it's having one complicated subroutine with sevaral variables and conditions versus have two small, clean and well-defined subroutines.

I'd like also to kindly point out that actually in our tests I noted much more cases needing both ssh and system up checks than only ssh checks and generally few or no case needing system up check _only_. 
So, by splitting those functionalities, we'd have many cases to update adding 2 function calls instead on 1 only, without a real benefit, but increasing the code complexity.

asmorodskyi (approved):
Well done Maurizio !

asmorodskyi:
> > > > What about separating `wait_for_ssh()` and `wait_for_boot()`? @asmorodskyi @m-dati
> > > 
> > > 
> > > Personally and from mantaining and using point of view I consider useful do not increase granularity and have a function with buttons to enable/disable similar properties instead to have many separated functions doing "similar" things
> > 
> > 
> > One subroutine would be checking for open port and the other for systemd status. Those are not the same things.
> > For me it's having one complicated subroutine with sevaral variables and conditions versus have two small, clean and well-defined subroutines.
> 
> I'd like also to kindly point out that actually in our tests I noted much more cases needing both ssh and system up checks than only ssh checks and generally few or no case needing system up check _only_. So, by splitting those functionalities, we'd have many cases to update adding 2 function calls instead on 1 only, without a real benefit, but increasing the code complexity.

Agree with Maurizio . And also in anycase I would move discussion about split out of this PR this **might** be next step but for this PR this will be too much already

asmorodskyi:
Now you need to do A LOT of VRs Maurizio as it is quite big change

pdostal (approved):
Let's proceed then 🚀

m-dati:
CI unit test point out needs for some fix.

asmorodskyi:
first thing you need to rebase on latest master and solve conflict !

```
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16890/conflicts) or the  to resolve conflicts.
Conflicting files
lib/publiccloud/instance.pm
```

m-dati:
> first thing you need to rebase on latest master and solve conflict !
> 
> ```
> This branch has conflicts that must be resolved
> Use the [web editor](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16890/conflicts) or the  to resolve conflicts.
> Conflicting files
> lib/publiccloud/instance.pm
> ```
Yes, usually I always rebase before pushing: it may be that last time I forgot some step: checking

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,59 @@ sub wait_for_ssh {
...
-    # - SSH port 22 is reachable
-    # - journalctl got message about reaching one of certain targets
-    while ((my $duration = time() - $start_time) < $args{timeout}) {
-        if ($check_port) {
-            $check_port = 0 if (script_run('nc -vz -w 1 ' . $self->{public_ip} . ' 22', quiet => 1) == 0);
+    my $instance_msg = sprintf("instance %s with public IP:%s within %d seconds", $self->{instance_id}, $self->{public_ip}, $args{timeout});
+    my $status = 0;
+    # Looping until SSH port 22 is reachable or timeout:
+    script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22',
+        timeout => $args{timeout}, delay => ($args{ignore_wrong_pubkey} ? 20 : 1), retry => 1, fail_msg => "Unable to reach SSH port of " . $instance_msg);
```
alvarocarvajald, line 360: I'd use either `"Unable to reach SSH port of $instance_msg"` or `'Unable to reach SSH port of ' . $instance_msg'`

asmorodskyi:
https://openqa.suse.de/tests/10992955
https://openqa.suse.de/tests/10993136
https://openqa.suse.de/tests/10993072
https://openqa.suse.de/tests/10990977
https://openqa.suse.de/tests/10989416
https://openqa.suse.de/tests/10989444
https://openqa.suse.de/tests/10989465
https://openqa.suse.de/tests/10989461
https://openqa.suse.de/tests/10976881
https://openqa.suse.de/tests/10976886
https://openqa.suse.de/tests/10976897
https://openqa.suse.de/tests/10992769
https://openqa.suse.de/tests/10992770
https://openqa.suse.de/tests/10992757
https://openqa.suse.de/tests/10992758

please clone ALL this jobs with your branch , in some cases you can even get failures in random places but only should be important for you that this failures not related to `wait_for_ssh`

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,59 @@ sub wait_for_ssh {
...
+    # ssh OK.
+
+    my $duration = time() - $start_time;
+
+    # help in boottime tests,when hard reboot could change the instance address:
+    my $out = $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
```
alvarocarvajald, line 366: Why assing this to `$out` if `$out` content is not checked before overwriting it in L371 and L378?

alvarocarvajald (commented):
A couple of comments from my part, but in general LGTM

lib/publiccloud/instance.pm
m-dati: done

lib/publiccloud/instance.pm
m-dati: done

m-dati:
Due to the behavior of `script _retry` and the `proceed_on_failure` opt. for ssh, a more updated is needed.

lib/publiccloud/instance.pm
```diff
@@ -433,7 +447,7 @@ sub softreboot {
 
     # wait till ssh disappear
     while (($duration = time() - $start_time) < $args{timeout}) {
-        last unless (defined($self->wait_for_ssh(timeout => 1, proceed_on_failure => 1, username => $args{username})));
+        last unless (defined($self->wait_for_ssh(timeout => 1, proceed_on_failure => 1, username => $args{username}, systemup_check => 0)));
```
pdostal, line 450: This doesn't make sense to me.

m-dati: Reading the comment in L448, we can see that it is important instead, for the `stop` phase.
During _stopping_ phase of a `soft reboot` we have the reverse process than the access and ssh is the last service that we can interrogate and will no more answer, while during the start it will be the first one to answer. 
It is like when you're exiting from a room, waiting until the port is closed.

m-dati: Reading the comment in L448, we can see that it is important instead, for the `stop` phase.
During _stopping_ phase of a `soft reboot` we have the reverse process than the access and ssh is the last service that we can interrogate and will no more answer when stop completed, therefore we have also `proceed_on_failure => 1`,but no need to check also the system up; during the start phase instead, ssh will be the first one to answer. 
It is like when a party, beers, music, all finished and you're exiting from that house: after that the door (ssh) is closed, you guess that party is over.

m-dati: Reading the comment in L448, we can see that it is important instead, for the `stop` phase.
During _stopping_ phase of a `soft reboot` we have the reverse process than the start and access and ssh is the last service that we can interrogate and will no more answer when stop completed, therefore we have also `proceed_on_failure => 1`,but no need to check also the system up; during the start phase instead, ssh will be the first one to answer. 
It is like when a party, beers, music, all finished and you're exiting from that house: after that the door (ssh) is closed, you guess that party is over.

pdostal: I see, we call `$self->wait_for_ssh()` below on line 454.

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
+    # Looping until SSH port 22 is reachable or timeout.
+    # Consider we that the args{timeout} is the Total timeout,
+    # while the script_retry's timeout is for each command retry,
+    # however we use the nc command with internal '-w 1' sec. timeout.
+    # So, we calculate the retry number of the check + delay, consuming the Total timeout,
+    # but with 10sec timeout as upper limit of the command run:
+    my $delay = $args{ignore_wrong_pubkey} ? 20 : 1;
+    my $status = script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22',
+        timeout => 10, delay => $delay, retry => int($args{timeout} / ($delay + 1)) + 1,
+        fail_msg => "Unable to reach SSH port of $instance_msg", die => !$args{proceed_on_failure});
```
pdostal, line 367:
```suggestion
    my $status = script_retry('nc -vz -w 1 ' . $self->{public_ip} . ' 22',
        timeout => 10, delay => 5, retry => int($args{timeout} / 5),
        fail_msg => "Unable to reach SSH port of $instance_msg", die => !$args{proceed_on_failure});
```

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
+    # ssh OK or proceed_on_failure
+    my $duration = time() - $start_time;
+    $status = 0 unless (defined($status) and $args{timeout} > $duration);
+
+    # help in boottime tests,when hard reboot could change the instance address:
+    $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
+      if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
```
pdostal, line 374:
```suggestion
    # On boottime test we do hard reboot which may change the instance address
    $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
      if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
```
Do we really hard reboot also in perf test?

lib/publiccloud/instance.pm
pdostal: The IP address is printed by `terraform apply`.

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
+
+    my $out;
+    if ($status) {
+        $out = "ssh access OK";
```
pdostal, line 378:
```suggestion
        $out = "SSH access OK.\n";
```

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
-                $args{timeout});
+    # check also remote system is running, looping by --wait option:
+    if ($args{systemup_check} and $status) {
+        # timeout recalculated removing consumed time until now
+        $out = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running --wait',
+            timeout => $args{timeout} - $duration, proceed_on_failure => 1, username => $args{username});
+
+        $duration = time() - $start_time;    # time update after check result
+        if ($out =~ m/running/) {    # startup OK
+            $status = 1;
```
pdostal, line 391: Add
```perl
$out .= 'System successfully booted.';
```

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
+            $self->ssh_script_run('sudo journalctl -b --no-pager > /tmp/journalctl.log',
+                timeout => 360, proceed_on_failure => 1, username => $args{username});
+            $self->upload_log('/tmp/journalctl.log', failok => 1);
         }
-        croak($error_msg);
     }
-
-    return;
+    # result display
+    record_info("CHECK " . ($args{systemup_check} ? "SYSTEM" : "SSH"), "Duration: $duration sec.\nResult: $out");
```
pdostal, line 410: Add
```perl
croak('Booting timeout exceeded!') if ($duration > $args{timeout});
```

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
+    $status = 0 unless (defined($status) and $args{timeout} > $duration);
+
+    # help in boottime tests,when hard reboot could change the instance address:
+    $self->ssh_script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
+      if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME'));
+
+    my $out;
+    if ($status) {
+        $out = "ssh access OK";
     }
```
pdostal, line 379:
```suggestion
    } else {
        $out = "SSH port is not open.\n";
    }
```

lib/publiccloud/instance.pm
asmorodskyi: ok

lib/publiccloud/instance.pm
```diff
@@ -349,56 +351,68 @@ sub wait_for_ssh {
...
-            elsif ($output =~ m/Permission denied \(publickey\).*/) {
-                die "ssh permission denied (pubkey)" unless $args{ignore_wrong_pubkey};
-            }
-        }
-        sleep $sleep_period;
+    my $instance_msg = sprintf("instance %s with public IP:%s within %d seconds", $self->{instance_id}, $self->{public_ip}, $args{timeout});
+
+    # Looping until SSH port 22 is reachable or timeout.
+    # Consider we that the args{timeout} is the Total timeout,
+    # while the script_retry's timeout is for each command retry,
```
asmorodskyi, line 360:
```suggestion
    # while the script_retry's timeout defines timeout of each retry iteration,
```

lib/publiccloud/instance.pm
asmorodskyi: `$args{ignore_wrong_pubkey}` does not have any effect in your suggestion

lib/publiccloud/instance.pm
m-dati: In `start` and `softreboot` subs we still use `wait_for_ssh`, even during for the `shutdown` phase inside `softreboot`, as discussed [above](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16890#discussion_r1181114476). Therefore there could be still the risk of such issue: isn't it?
Eventually the comment could be better updated accordingly

m-dati: Probably, discussion apart, even the 'hard' `stop` sub should include such call, like in the above comment on `softreboot`

lib/publiccloud/instance.pm
m-dati: Already performed in `script_retry` by *fail_msg*

asmorodskyi: why ? isn't this will fail execution ? what about `$args{proceed_on_failure}` in such case ?

lib/publiccloud/instance.pm
m-dati: I think that `delay 5` is too much, we use wait option `-w 1` in `nc` that already implements an internal loop, so instead better to increase it to `-w 5`.

I mantained some settings of the old code, as in master, but not clear utility of `ignore_wrong_pubkey` opt., having already `proceed_on_failure`, 
neither clear why *delay* (old name sleep_period) should be so big 20s when set  true.

lib/publiccloud/instance.pm
m-dati: That information is already present in the ending `croak` call, at/near L420, and the proper message will be composed inside the previous `if` branches.

lib/publiccloud/instance.pm
```diff
@@ -346,59 +348,74 @@ sub wait_for_ssh {
...
+    # So, we calculate the retry number of the check + delay, consuming the Total timeout,
+    # but with 10sec timeout as upper limit of the command run:
+    my $delay = $args{ignore_wrong_pubkey} ? 20 : 1;
+
+    my $status = script_retry('nc -vz -w 4 ' . $self->{public_ip} . ' 22',
+        timeout => 10, delay => $delay, retry => int($args{timeout} / ($delay + 4)) + 1, quiet => 1,
+        fail_msg => "Unable to reach SSH port of $instance_msg in $args{timeout} seconds.", die => !$args{proceed_on_failure});
+    # ssh OK or proceed_on_failure
+    my $duration = time() - $start_time;
+    $status = (defined($status) and $status == 0) ? 1 : 0;    # shell-code to perl-test convertion
```
alvarocarvajald, line 371: No need for the ternary operator there.
```
$status = (defined($status) and $status == 0);
```
Is enough.

Also, there's a typo in the comment: it should be `conversion`.

lib/publiccloud/instance.pm
```diff
@@ -346,59 +348,74 @@ sub wait_for_ssh {
...
+        }
+        else {    # FAIL: $sysout ~ maintenance|stopping|offline|unknown or no pubkey or else
+            $status = 0;
+            $sysout .= "\nCan not reach systemd target on $instance_msg";
         }
-        else {
-            $error_msg = sprintf("Can not reach systemd target on instance %s with public IP:%s within %d seconds",
-                $self->{instance_id}, $self->{public_ip}, $args{timeout});
+        # Log upload
+        if (!get_var('PUBLIC_CLOUD_SLES4SAP') and $status > 0) {
```
alvarocarvajald, line 404: `$status` is being treated as a boolean elsewhere. I'd go here for readability with:
```
if (!get_var('PUBLIC_CLOUD_SLES4SAP') and $status) {
```

m-dati: Yes thanks, it is a typo due to a previous plan to use values < 0, then abandoned.

lib/publiccloud/instance.pm
m-dati: Thanks

m-dati:
First VR after last update: 
sle-micro-15-SP4-Micro-5-4-BYOS-EC2-x86_64
https://openqa.suse.de/tests/11019168

lib/publiccloud/instance.pm
```diff
@@ -346,59 +355,88 @@ sub wait_for_ssh {
     # until it is not over we will recieve "ssh permission denied (pubkey)" error
     # but it is not good reason to die early because after it will be over DMS will return normal
     # user and error will be resolved.
-    $args{ignore_wrong_pubkey} //= 0;
+    #$args{ignore_wrong_pubkey} //= $args{proceed_on_failure};
```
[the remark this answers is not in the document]
m-dati: yes, it is in test migration.pm only

lib/publiccloud/instance.pm
```diff
@@ -333,9 +333,18 @@ sub wait_for_guestregister {
...
+
+Check if the SSH port of the remote instance is reachable and open, 
+then by default also check that system is up.
+
+Parameters:
+ timeout => total wait timeout; default 600;
+ proceed_on_failure => false=exit with error, true=continue to run; def. false;
+ username => def. username();
+ public_ip => def. public_ip();
+ systemup_check => true=after ssh also check system up, false=only ssh port check; def. true;
```
grisu48, line 346:
```suggestion
 systemup_check => If true, checks if the system is up, instead of just checking the ssh port. Default: true
```

lib/publiccloud/instance.pm
```diff
@@ -333,9 +333,18 @@ sub wait_for_guestregister {
...
-
-Check if the SSH port of the instance is reachable and open.
+    wait_for_ssh([timeout => 600] [, proceed_on_failure => 0] [, ...])
+
+Check if the SSH port of the remote instance is reachable and open, 
+then by default also check that system is up.
+
+Parameters:
+ timeout => total wait timeout; default 600;
+ proceed_on_failure => false=exit with error, true=continue to run; def. false;
```
grisu48, line 343:
```suggestion
 proceed_on_failure => false=exit with error, true=continue to run; default false;
```

Please no lazy abbreviations in documentation :slightly_smiling_face:

lib/publiccloud/instance.pm
```diff
@@ -346,59 +355,88 @@ sub wait_for_ssh {
     # until it is not over we will recieve "ssh permission denied (pubkey)" error
     # but it is not good reason to die early because after it will be over DMS will return normal
     # user and error will be resolved.
-    $args{ignore_wrong_pubkey} //= 0;
+    #$args{ignore_wrong_pubkey} //= $args{proceed_on_failure};
     $args{username} //= $self->username();
     $args{public_ip} //= $self->public_ip();
+    $args{systemup_check} //= 1;    # check also system up, defult:true; when false,systemup check not executed,only ssh checked.
```
grisu48, line 361:
```suggestion
    $args{systemup_check} //= 1;    # If true, check if the system state is up, otherwise only the ssh port will be checked for availability. Default: true
```

lib/publiccloud/instance.pm
```diff
@@ -346,59 +355,88 @@ sub wait_for_ssh {
...
+    # args{timeout} is the Total timeout,
+    # while the script_retry's timeout defines timeout of each retry iteration,
+    # however we use the nc command with option '-w 3' seconds of internal timeout.
+    # So, we calculate the retry number of the '-w' + delay consuming the Total timeout,
+    # but with 5sec timeout as upper limit of the command run:
+
+    my $exit_code = script_retry('nc -vz -w 3 ' . $self->{public_ip} . ' 22',
+        timeout => 5, delay => 1, retry => int($args{timeout} / 4) + 1, quiet => 1,
+        fail_msg => "Unable to reach SSH port of $instance_msg in $args{timeout} seconds.", die => !$args{proceed_on_failure});
+    # ssh OK or proceed_on_failure true: script ok has exit_code==0
```
grisu48, line 376: I don't fully understand this comment, please check if my suggestion makes sense:

```suggestion
    $exit_code is 0 if ssh is reachable
```

I don't understand what `proceed_on_failure true` means in this context.

lib/publiccloud/instance.pm
```diff
@@ -346,59 +355,88 @@ sub wait_for_ssh {
...
+
+    # Looping until SSH port 22 is reachable or timeout.
+    # args{timeout} is the Total timeout,
+    # while the script_retry's timeout defines timeout of each retry iteration,
+    # however we use the nc command with option '-w 3' seconds of internal timeout.
+    # So, we calculate the retry number of the '-w' + delay consuming the Total timeout,
+    # but with 5sec timeout as upper limit of the command run:
+
+    my $exit_code = script_retry('nc -vz -w 3 ' . $self->{public_ip} . ' 22',
+        timeout => 5, delay => 1, retry => int($args{timeout} / 4) + 1, quiet => 1,
```
grisu48, line 374: The retry counter is off in my opinion. I think you wanted to do

```suggestion
        timeout => 5, delay => 1, retry => int($args{timeout} / 6) + 1, quiet => 1,
```

Why do I think this is wrong? See the following example:

    timeout = 120
    retry = (120/4)+1 = 31

But now, If we execute 31x a timeout of 5 seconds plus a delay of 1, then we get 186, which is not the expected timeout of 120.

If you want to quit after `timeout` seconds, then you need to calculate the retry counter by taking into account the full length of one retry interval, which is `timeout + delay`, not `timeout - delay`.

lib/publiccloud/instance.pm
```diff
@@ -471,8 +509,8 @@ sub stop {
 
     start([timeout => ?]);
 
-Start the instance and check SSH connectivity. Return the number of seconds
-till the SSH port was available.
+Start the instance and check SSH connectivity then system is UP.
```
grisu48, line 512:
```suggestion
Start the instance and wait for the system to be up
```

grisu48 (commented):
Some minor documentation comments plus one mistake in the calculation of the retry counts in my opinion.

m-dati:
Sorry @grisu48 Felix, I didn't pushed yet the new commit :) , having been code many changes, but I will apply your suggestion to comments where present, before to push it :)

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+Wnen a remote pc instance is stopping in shutdown, we set input param. wait_stop=1(true),
+to expect until ssh is closed; automatic defaults systemup_check=0 and proceed_on_failure=1 applied.
+ 
+Parameters:
+ timeout => total wait timeout; default 600.
+ wait_stop => If true waits for ssh stopping, also auto-disabling system up check, otherwise false waits for ssh starting; default false.
+ proceed_on_failure => If false exit with error, if true continue to run; default the value assumed by wait_stop.
+ username => def. username().
+ public_ip => def. public_ip().
+ systemup_check => If true, checks if the system is up, instead of just checking the ssh port; default negated value assumed by wait_stop.
```
grisu48, line 351:
```suggestion
 systemup_check => If true, checks if the system is up, instead of just checking the ssh port; default: !wait_stop
```

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+    $args{wait_stop} //= 0;    # see head2 - Parameters: wait_stop
+    $args{proceed_on_failure} //= $args{wait_stop};    # see head2 - Parameters: proceed_on_failure
+    $args{systemup_check} //= not $args{wait_stop};    # see head2 - Parameters: systemup_check
+    $args{logs} //= 1;    # see head2 - Parameters: logs
+    $args{public_ip} //= $self->public_ip();
     # DMS migration (tests/publiccloud/migration.pm) is running under user "migration"
-    # until it is not over we will recieve "ssh permission denied (pubkey)" error
-    # but it is not good reason to die early because after it will be over DMS will return normal
-    # user and error will be resolved.
+    # until it is not over we will receieve "ssh permission denied (pubkey)" error
```
grisu48, line 368:
```suggestion
    # until it is not over we will receive "ssh permission denied (pubkey)" error
```

lib/publiccloud/instance.pm
```diff
@@ -432,9 +502,10 @@ sub softreboot {
     my $start_time = time();
 
     # wait till ssh disappear
-    while (($duration = time() - $start_time) < $args{timeout}) {
-        last unless (defined($self->wait_for_ssh(timeout => 1, proceed_on_failure => 1, username => $args{username})));
-    }
+    my $out = $self->wait_for_ssh(timeout => $args{timeout}, wait_stop => 1, username => $args{username});
+    # ok ssh port closed
+    record_info("STOP NOK", "$out") if (defined $out);    # not ok port still open
```
grisu48, line 507: ?????????????

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+When a remote pc instance starting, by default wait_stop param.=0(false) and 
+this routine checks until the SSH port of the remote instance is reachable and open. 
+Then by default also checks that system is up, unless systemup_check false/0.
+
+Wnen a remote pc instance is stopping in shutdown, we set input param. wait_stop=1(true),
+to expect until ssh is closed; automatic defaults systemup_check=0 and proceed_on_failure=1 applied.
+ 
+Parameters:
+ timeout => total wait timeout; default 600.
+ wait_stop => If true waits for ssh stopping, also auto-disabling system up check, otherwise false waits for ssh starting; default false.
```
grisu48, line 347:
```suggestion
 wait_stop => If true waits for ssh to become unreachable. Default false.
```

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+ proceed_on_failure => If false exit with error, if true continue to run; default the value assumed by wait_stop.
```
grisu48, line 348:
```suggestion
 proceed_on_failure => If false exit with error, if true continue to run; default: wait_stop
```

lib/publiccloud/instance.pm
m-dati: In the case we check the ssh for **halting** systems,  we expect the $out is `undef`. The previous code didn't consider this check, and here I only provide a debug info in the case the system _still answering_ after timeout, that is $out is defined, instead.

lib/publiccloud/instance.pm
m-dati: text updated

lib/publiccloud/instance.pm
m-dati: code changed in last commit

m-dati:
2 VR on migration:

sle-12-SP4-DMS-EC2-SAP-BYOS https://openqa.suse.de/tests/11065195 PASS

sle-12-SP4-DMS-AZURE-BYOS https://openqa.suse.de/tests/11065193 PASS

lib/publiccloud/instance.pm
grisu48: Or just update the `STOP NOK` with something that everybody can understand without reading the code.

m-dati:
PR notes applied, general comments and softreboot stop phase info updated.
Following here the VR executed

lib/publiccloud/instance.pm
m-dati: done

lib/publiccloud/instance.pm
```diff
@@ -432,9 +502,11 @@ sub softreboot {
     my $start_time = time();
 
     # wait till ssh disappear
-    while (($duration = time() - $start_time) < $args{timeout}) {
-        last unless (defined($self->wait_for_ssh(timeout => 1, proceed_on_failure => 1, username => $args{username})));
-    }
+    my $out = $self->wait_for_ssh(timeout => $args{timeout}, wait_stop => 1, username => $args{username});
+    # ok ssh port closed
+    record_info("STOP NOK", "WARNING: while stopping the system, ssh port still open after timeout,\nreporting: $out")
```
grisu48, line 507:
```suggestion
    record_info("Shutdown failed", "WARNING: while stopping the system, ssh port still open after timeout,\nreporting: $out")
```

I assume NOK means not-ok. Let's please not use slang, which might be confusing.

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+            }
+            elsif ($sysout =~ m/initializing|starting/) {    # still starting
+                $exit_code = undef;
+            }
+            elsif ($sysout =~ m/running/) {    # startup OK
+                $exit_code = 0;
+                $sysout .= "\nSystem successfully booted";
+                last;
+            }
+            elsif ($sysout =~ m/degraded/) {    # up but with failed services to collect
```
grisu48, line 417: I think all of those `elsif` are not needed. We just want to check if the system is up. Additional checks (e.g. failing services) are not meant to take place at this point.

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
-            $error_msg = sprintf("Can not reach systemd target on instance %s with public IP:%s within %d seconds",
-                $self->{instance_id}, $self->{public_ip}, $args{timeout});
-        }
-        croak($error_msg);
+    # exit_code is 0 when shell script is ok
+    if (isok($exit_code)) {
+        $sshout = "SSH port is open\n";
     }
+    else {
+        $sshout = "SSH port is not open failed access\n";
```
grisu48, line 391: If the ssh port is closed, does it even make sense to continue? If not, let's just return here then we can stop carrying the `sshout` message around, which makes the whole flow simpler.

grisu48 (changes requested):
This PR is big and it introduces many new checks, which are not motivated. Given that this is part of the core code of our publiccloud handling, we should really reconsider if this is a good approach.

I would really advise to split this PR up into multiple smaller ones. AFAICS a suitable approach would be the following:

1. Replace line 368 (and follow-ups) with replacing the old `journalctl` with the new `systemctl is-system-running`
2. Change the loop (line 359) to incorporate the new behaviour
3. If needed, add additional checks

This PR is large, and changes the logic in many places. This will break things and changes at this level affect ALL publiccloud test runs for all squads. We need to be very careful here.

lib/publiccloud/instance.pm
m-dati: no, please, those checks do are needed, because the phases are different if before , during or after. As it is now, it works well, I already did many different tests

lib/publiccloud/instance.pm
m-dati: If the ssh port is closed, we collect here the information, the sys-up loop is skipped and we go down to the _result display_ section, where result appears in a test detail box then **die** (croak) if not `proceed_on_failure`

lib/publiccloud/instance.pm
```diff
@@ -333,73 +333,143 @@ sub wait_for_guestregister {
...
+        script_run("ssh-keyscan $args{public_ip} | tee -a ~/.ssh/known_hosts")
+          if (get_var('PUBLIC_CLOUD_PERF_COLLECT') or get_var('PUBLIC_CLOUD_CHECK_BOOT_TIME') or get_var('PUBLIC_CLOUD_CHECK_REBOOT'));
+        while (($duration = time() - $start_time) < $args{timeout}) {
+            # On boottime test we do hard reboot which may change the instance address:
+            # timeout recalculated removing consumed time until now
+            $sysout = $self->ssh_script_output(cmd => 'sudo systemctl is-system-running',
+                timeout => $args{timeout} - $duration, proceed_on_failure => 1, username => $args{username});
+            # result check
+            if ($sysout =~ m/Permission denied \(publickey\).*/) {
+                $exit_code = 2;
```
alvarocarvajald, line 406: Any reason for the different exit codes (0, 1, 2, undef)? They're set, but never used save by `isok()` which only checks for defined and 0.

alvarocarvajald (approved):
Just one question. LGTM in general.

lib/publiccloud/instance.pm
m-dati: It is basically to allow to distinguish that status error in the loop, 
because 2 occurrs for `pubkey` issue, particular case (conditioned if `ignore` is true or die) that could change state after some retries; 
similar for `undef` that is not really an issue, but a transitory status expecting changes after retries (unconditioned, never die).

m-dati: Indeed after this clarification , the `else` status could be 2 too :) (because allowed to retry)

m-dati:
Commit message updated with changes details:
```
* routine structure improved,split in 3 blocks,ssh-port check,system-up check,result report;
* system-up check improved using 'systemctl is-system-running',more reliable than journal inspection;
* report on final status improved,also returned when die(croak),helping in debugging;
* description in head section improved,parameters meaning and usage detailed;
* start-up phases detailed in inner loop,providing precise details on the status detected;
* new option `systemup_check`,to check on-demand ssh-open only and skip system-up check;
* new option `wait_stop`,to wait until ssh-closed,instead of open,improving shutdown/reboot checks;
* new option `logs`,journalctl uploaded to test logs,instead of printed on ouptput,logs=0 to skip;
* softreboot routine updated to use wait_stop option.
```
Note that in `systemctl is-system-running`, the `--wait` option has been _removed_, result checked in loop, because option available in systemd version >= v240, while SLE 12-SP4 / 5, 15-SP1 / 2 images tests have version < v240.

m-dati:
poo [126218](https://progress.opensuse.org/issues/126218 ) **PAUSED**

pdostal:
@m-dati please squeeze the commits, update the pull request description and title.

pdostal:
Also please leave one empty line between the commit title and commit message body.

m-dati:
New VR tests on curent branch:

- **PC Latest**: 
  - sle-15-SP5-Azure-BYOS-x86_64-Build0317-publiccloud_consoletests@az_Standard_B2s
https://openqa.suse.de/tests/11200940 
{
https://openqa.suse.de/tests/11200940#step/prepare_instance/190
https://openqa.suse.de/tests/11200940#step/journalctl/86
https://openqa.suse.de/tests/11200940#step/journalctl/101
}
PASS

  - sle-15-SP5-EC2-x86_64-Build0314-publiccloud_consoletests@ec2_t2.large
https://openqa.suse.de/tests/11200942 
{
https://openqa.suse.de/tests/11200942#step/prepare_instance/197
https://openqa.suse.de/tests/11200942#step/journalctl/86
https://openqa.suse.de/tests/11200942#step/journalctl/101
}
PASS

  - sle-15-SP5-GCE-x86_64-Build0277-publiccloud_consoletests@gce_n1_standard_2
https://openqa.suse.de/tests/11200943 
{
https://openqa.suse.de/tests/11200943#step/prepare_instance/246
https://openqa.suse.de/tests/11200943#step/journalctl/86
https://openqa.suse.de/tests/11200943#step/journalctl/101
}
PASS
===

m-dati:
Second VRs session

- **PC Maint.Test Repo / Upates**:

  - sle-15-SP4-AZURE-Basic-Updates-aarch64-Build20230525-1-publiccloud_img_proof@64bit
https://openqa.suse.de/tests/11200980 PASS
{
https://openqa.suse.de/tests/11200980#step/prepare_instance/143
https://openqa.suse.de/tests/11200980#step/patch_and_reboot/40 
https://openqa.suse.de/tests/11200980#step/patch_and_reboot/82
}

  - sle-12-SP5-EC2-Updates-x86_64-Build20230525-1-publiccloud_consoletests@64bit
https://openqa.suse.de/tests/11201283 PASS
{
https://openqa.suse.de/tests/11201283#step/prepare_instance/185
https://openqa.suse.de/tests/11201283#step/patch_and_reboot/40
https://openqa.suse.de/tests/11201283#step/patch_and_reboot/66
https://openqa.suse.de/tests/11201283#step/journalctl/73
https://openqa.suse.de/tests/11201283#step/journalctl/88 
}

  - sle-15-SP4-EC2-CHOST-BYOS-Updates-x86_64-Build20230525-1-publiccloud_containers@64bit 
https://openqa.suse.de/tests/11201284 PASS
{
https://openqa.suse.de/tests/11201284#step/prepare_instance/121
https://openqa.suse.de/tests/11201284#step/patch_and_reboot/40
https://openqa.suse.de/tests/11201284#step/patch_and_reboot/74
}

  - sle-15-SP4-EC2-BYOS-Updates-aarch64-Build20230525-1-publiccloud_consoletests@64bit
https://openqa.suse.de/tests/11200973 PASS
{
https://openqa.suse.de/tests/11200973#step/prepare_instance/113
https://openqa.suse.de/tests/11200973#step/patch_and_reboot/40
https://openqa.suse.de/tests/11200973#step/patch_and_reboot/74
https://openqa.suse.de/tests/11200973#step/journalctl/86
https://openqa.suse.de/tests/11200973#step/journalctl/107
}

  - sle-15-SP4-GCE-BYOS-Updates-x86_64-Build20230525-1-publiccloud_containers@64bit
https://openqa.suse.de/tests/11200974 PASS 
{
https://openqa.suse.de/tests/11200974#step/prepare_instance/98
https://openqa.suse.de/tests/11200974#step/patch_and_reboot/40
https://openqa.suse.de/tests/11200974#step/patch_and_reboot/74
}

  - sle-15-SP3-GCE-BYOS-Updates-x86_64-Build20230525-1-publiccloud_consoletests@64bit
https://openqa.suse.de/tests/11201274 PASS
{
https://openqa.suse.de/tests/11201274#step/prepare_instance/98
https://openqa.suse.de/tests/11201274#step/patch_and_reboot/40
https://openqa.suse.de/tests/11201274#step/patch_and_reboot/74
}

  - sle-15-SP4-GCE-CHOST-BYOS-Updates-aarch64-Build20230524-1-publiccloud_containers@64bit
https://openqa.suse.de/tests/11200982
https://openqa.suse.de/tests/11200982#step/prepare_instance/98 PASS wait_for_sssh
https://openqa.suse.de/tests/11200982#step/transfer_repos/321 FAIL in transfer_repos
_FAILURE Not caused by this PR._

m-dati:
VRs session n.3:

- **Latest HA SAP**:

   - sle-15-SP5-Azure-SAP-BYOS-x86_64-Build0227-sles4sap_gnome_saptune_delete_rename@az_Standard_E8s_v3 - 
https://openqa.suse.de/tests/11201371   PASS
{
https://openqa.suse.de/tests/11201371#step/qesap_terraform/343
https://openqa.suse.de/tests/11201371#step/mr_test/50
https://openqa.suse.de/tests/11201371#step/mr_test/57
}

  - sle-15-SP5-EC2-SAP-BYOS-x86_64-Build0227-sles4sap_gnome_saptune_notes@ec2_r4.8xlarge - 
https://openqa.suse.de/tests/11201372 PASS
{
https://openqa.suse.de/tests/11201372#step/qesap_terraform/385
https://openqa.suse.de/tests/11201372#step/mr_test/50
https://openqa.suse.de/tests/11201372#step/mr_test/57 
}

  - sle-15-SP5-GCE-SAP-x86_64-Build0227-sles4sap_gnome_saptune_notes@gce_n1_highmem_8 - 
https://openqa.suse.de/tests/11201373 PASS
{
https://openqa.suse.de/tests/11201373#step/qesap_terraform/442
https://openqa.suse.de/tests/11201373#step/mr_test/50
https://openqa.suse.de/tests/11201373#step/mr_test/57
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/27
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/34
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/53
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/60
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/92
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/99
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/118
https://openqa.suse.de/tests/11201373#step/1_saptune_notes/125
.../ all other softreboots PASS OK.
}


- **SLEM 5.4 on P.C**:

  - sle-micro-5.4-BYOS-Azure-x86_64-Build0113-slem_containers@64bit
https://openqa.suse.de/tests/11201376  FAIL - NOT DUE TO THIS PR.
{
https://openqa.suse.de/tests/11201376#step/prepare_instance/190 OK
https://openqa.suse.de/tests/11201376#step/slem_prepare/16 OK
https://openqa.suse.de/tests/11201376#step/slem_prepare/37 OK
https://openqa.suse.de/tests/11201376#step/podman_netavark/62 OK
https://openqa.suse.de/tests/11201376#step/podman_netavark/83 OK
https://openqa.suse.de/tests/11201376#step/podman_netavark/93 FAILED for needle
no candidate needle with tag(s) 'user-console' matched
}

  - sle-micro-5.4-BYOS-GCE-ARM-x86_64-Build0113-slem_containers@64bit
https://openqa.suse.de/tests/11201377  FAIL – MORE INFO ADDED
{
https://openqa.suse.de/tests/11201377#step/prepare_instance/90 Port not open: see info timeout.
https://openqa.suse.de/tests/11201377#step/prepare_instance/92 FAIL [croak in L458](https://github.com/m-dati/os-autoinst-distri-opensuse/blob/d91b6ed444061982f50c6a106ae0933e039b38da/lib/publiccloud/instance.pm#L458)
}

  - sle-micro-5.4-BYOS-EC2-x86_64-Build0113-slem_containers@64bit
https://openqa.suse.de/tests/11201378 PASS
{
https://openqa.suse.de/tests/11201378#step/prepare_instance/165
https://openqa.suse.de/tests/11201378#step/slem_prepare/16
https://openqa.suse.de/tests/11201378#step/slem_prepare/31
https://openqa.suse.de/tests/11201378#step/podman_netavark/62
https://openqa.suse.de/tests/11201378#step/podman_netavark/77
}

- **Migration**:

  - sle-12-SP4-DMS-GCE-SAP-BYOS-x86_64-Build0201-GCE_SP4_SAP_x86_64_DMS_DEMO@64bit
https://openqa.suse.de/tests/11201379 PASS
{
https://openqa.suse.de/tests/11201379#step/migration/64
https://openqa.suse.de/tests/11201379#step/migration/158
https://openqa.suse.de/tests/11201379#step/migration/608
}

  - sle-12-SP5-DMS-EC2-SAP-BYOS-x86_64-Build0199-EC2_SP5_SAP_x86_64_DMS_DEMO @64bit
https://openqa.suse.de/tests/11201311 FAIL - MORE INFO ADDED
{
[https://openqa.suse.de/tests/11201311#step/migration/6](https://openqa.suse.de/tests/11201311#step/migration/61) wait_for_ssh routine provides more infos.
https://openqa.suse.de/tests/11201311#step/migration/63 FAIL identity file
} FAILURE Not caused by this PR that instead added more details on test issue.

m-dati:
Restarting this PR activity, I provided above 3 sessions of Verification tests executed today, for kind check and approval.

ricardobranco777 (approved):
LGTM

asmorodskyi:
I will merge it on Monday , but also @grisu48  needs to remove his veto before ...

jlausuch (approved):

m-dati:
Last commit for rebase refresh only

m-dati:
Here last VRs tests runs for SLEM5 on P.C.:

- sle-micro-5.4,3 on P.C.
  - sle-micro-5.4-BYOS-Azure-x86_64-Build0117-slem_basic  
  https://openqa.suse.de/tests/11211389  PASS
  - sle-micro-5.4-BYOS-GCE-ARM-x86_64-Build0117-slem_basic  
  https://openqa.suse.de/tests/11211396  PASS
  - sle-micro-5.4-BYOS-GCE-ARM-x86_64-Build0117-slem_containers  
 https://openqa.suse.de/tests/11211390  PASS
  - sle-micro-5.4-BYOS-EC2-x86_64-Build0117-slem_basic  
 https://openqa.suse.de/tests/11211391 PASS
  - sle-micro-5.3-BYOS-Azure-x86_64-Build0377-slem_basic
  https://openqa.suse.de/tests/11211404 PASS

m-dati:
VRs after last commit:
https://openqa.suse.de/tests/11211720 PASS
https://openqa.suse.de/tests/11211722 PASS
https://openqa.suse.de/tests/11211723 PASS

m-dati:
VRs after last commit:
- sle-15-SP5-GCE-x86_64-Build0277-publiccloud_consoletests.02 @gce_n1_standard_2
  https://openqa.suse.de/tests/11211735 PASS
- sle-micro-5.4-BYOS-Azure-x86_64-Build0117-slem_basic
  https://openqa.suse.de/tests/11211720 PASS
- sle-12-SP4-DMS-GCE-SAP-BYOS-x86_64-Build0201-GCE_SP4_SAP_x86_64_DMS_DEMO
  https://openqa.suse.de/tests/11211722 PASS
- sle-12-SP5-EC2-Updates-x86_64-Build20230525-1-publiccloud_consoletests
  https://openqa.suse.de/tests/11211723 PASS

grisu48 (approved):

PR #16891, opened 2023-04-19
Switch to users terminal to run srun without --uid flags

b10n1k:
As the `--uid` is causing issues and it is not a realistic user case, we have
to run commands as those users instead as root. For this the users needs to
exists on the compute nodes.

- Verification run: https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=b10n1k%2Fos-autoinst-distri-opensuse%23refactor_slurm


tests/hpc/slurm_master.pm
```diff
@@ -483,4 +505,15 @@ sub post_fail_hook ($self) {
     upload_logs('/var/log/slurmctld.log');
 }
 
+sub create_user_account ($username) {
+    return 1 unless $username;
```
pevik, line 509: I'd expect you need to do `my $username = shift;`

tests/hpc/slurm_master.pm
```diff
@@ -232,10 +232,9 @@ sub t01_accounting() {
     );
 
     ##Add users TODO: surely this should be abstracted
```
pevik, line 234: You left here that ugly formatted TODO.

tests/hpc/slurm_master.pm
```diff
@@ -267,12 +266,29 @@ sub t01_accounting() {
...
-
-    script_run("srun --uid=$users{user_3} --account=UNI_Y_Biology -N 3 date");
-    script_run("srun --uid=$users{user_4} --account=UNI_Y_Physics -N 3 hostname");
+    my $current_user = $testapi::username;
+    $testapi::username = $users{user_1};
+    my $prompt = $testapi::username . '@' . get_required_var('HOSTNAME') . ':~> ';
+    record_info "$testapi::username", "ok";
+    select_user_serial_terminal($prompt);
+    script_run("srun --account=UNI_X_Math -w slave-node00,slave-node01 date");
+    type_string("su - $users{user_2}", lf => 1);
```
pevik, line 275: I wonder if it helped you to have more user serial terminals. Because with `su -` it does not matter whether the previous account was `root` or non-root UID.

tests/hpc/slurm_slave.pm
```diff
@@ -28,6 +28,18 @@ sub run ($self) {
...
+        'user_2' => 'Egbert',
+        'user_3' => 'Christina',
+        'user_4' => 'Jose',
+    );
+
+    ##Add users TODO: surely this should be abstracted
+    script_run("useradd -m $users{user_1}");
+    script_run("useradd -m $users{user_2}");
+    script_run("useradd -m $users{user_3}");
+    script_run("useradd -m $users{user_4}");
```
pevik, line 42: Why not something like you do above?
```
foreach my $key (keys %{users}) {
	create_user_account($users{$key});
}
```

tests/hpc/slurm_master.pm
b10n1k: I need input from seba for that. its his comment and i am not sure what he had in mind.

tests/hpc/slurm_master.pm
b10n1k: i think it would help. But i dont know how to add this functionality

tests/hpc/slurm_slave.pm
b10n1k: Done

tests/hpc/slurm_master.pm
b10n1k: in fact i dont even need this check due to `-signatures` the function will complain when the signature is not invoked with `$username`. i will remove it

b10n1k: Done

schlad (approved):

tests/hpc/slurm_master.pm
```diff
@@ -483,4 +505,14 @@ sub post_fail_hook ($self) {
...
 }
 
+sub create_user_account ($username) {
+    script_run("useradd -m $username");
+    type_string("passwd $username", lf => 1);
+    wait_serial("New password:");
+    type_string("$testapi::password", lf => 1);
+    wait_serial("Retype new password:");
+    type_string("$testapi::password", lf => 1);
+    return 0;
```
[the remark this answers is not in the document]
b10n1k: Done

czerw (approved):
LGTM

tests/hpc/slurm_master.pm
pevik: `VIRTIO_CONSOLE_NUM=N` could be the way. `VIRTIO_CONSOLE_NUM=2`  creates one root serial console. `VIRTIO_CONSOLE_NUM=3` creates 2 root serial consoles. Maybe I could add the same functionality for user serial consoles. But they would login as `bernhart`, thus you would still need to use `su -`.

PR #16892, opened 2023-04-19
Ensure 'Add' button clicked in Network Settings

lemon-suse:
We need ensure 'Add' button clicked in Network Settings, and there is refreshment issue for ' yast2-network-settings_overview_hardware-dialog_device-type_bridge', so add workaround for it.

- Related ticket: https://progress.opensuse.org/issues/127607
- Needles: N/A
- Verification run:  https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=lemon-suse%2Fos-autoinst-distri-opensuse%23ensure-add-button-clicked-network-setting&flavor=Online
 job for the glitch issue:  https://openqa.suse.de/tests/10971316#step/yast2_network_settings/18

### DIFF
--- a/tests/yast2_gui/yast2_network_settings.pm
+++ b/tests/yast2_gui/yast2_network_settings.pm
@@ -17,6 +17,7 @@ use warnings;
 use testapi;
 use y2_module_basetest 'is_network_manager_default';
 use version_utils 'is_sle';
+use YaST::workarounds;
 
 sub run {
 
@@ -57,12 +58,14 @@ sub run {
     unless (is_network_manager_default) {
         send_key $cmd{overview_tab};
         assert_screen 'yast2-network-settings_overview';
+        wait_still_screen 2;
         send_key $cmd{add_device};
         # Older than 15-SP2 versions require two steps to select device type: expand the list and then select the option.
         # On 15-SP2 it is a list of radio buttons, so only one step with selecting the radio is needed.
         if (is_sle('<15-SP2')) {
             assert_and_click 'yast2-network-settings_overview_hardware-dialog_device-type';
         }
+        apply_workaround_poo124652('yast2-network-settings_overview_hardware-dialog_device-type_bridge');
         assert_and_click 'yast2-network-settings_overview_hardware-dialog_device-type_bridge';
         send_key $cmd{next};
         assert_screen 'yast2-network-settings_overview_network-card-setup';


jknphy:
LGTM other than we need to show 1 link where we face the glitch issue.

tests/yast2_gui/yast2_network_settings.pm
```diff
@@ -57,12 +58,14 @@ sub run {
     unless (is_network_manager_default) {
         send_key $cmd{overview_tab};
         assert_screen 'yast2-network-settings_overview';
+        wait_still_screen 2;
```
coolgw, line 61: I would like to use assert_screen_change { send_key xx } try make sure page open or send_key_until_needlematch.
My guess after assert_screen the screen should already still?

tests/yast2_gui/yast2_network_settings.pm
```diff
@@ -57,12 +58,14 @@ sub run {
...
         send_key $cmd{add_device};
         # Older than 15-SP2 versions require two steps to select device type: expand the list and then select the option.
         # On 15-SP2 it is a list of radio buttons, so only one step with selecting the radio is needed.
         if (is_sle('<15-SP2')) {
             assert_and_click 'yast2-network-settings_overview_hardware-dialog_device-type';
         }
+        apply_workaround_poo124652('yast2-network-settings_overview_hardware-dialog_device-type_bridge');
```
coolgw, line 68: I am not sure we need this, the page not fully show?

tests/yast2_gui/yast2_network_settings.pm
lemon-suse: I have different opinion on this, the 'add' button may not work at that point which cause send key but next screen not shown, so we need wait to make sure the 'add' button work then send key.

See the latest VR, I have run 10 times and no such issue found any more.

tests/yast2_gui/yast2_network_settings.pm
lemon-suse: See the job in description for the glitch issue, the page stop there and no refreshment, so I add the workaround there, while this is a sporistic issue.

jknphy: thanks @lemon-suse I had the same doubt, now is clear, the glitch hits again and again.

tests/yast2_gui/yast2_network_settings.pm
jknphy: assert_screen doesn't warranted that, the best way to fix that is with libyui-rest-api and we are not investing now on the scope of this ticket, so I guess this solution is fine.

PR #16895, opened 2023-04-19
[WIP] kernel/run_ltp.pm: Improve timeout capabilities

pevik:
Add new variables to specify timeout for particular arch or test only.

* `LTP_TIMEOUT_test` (e.g.  `LTP_TIMEOUT_zram01`) Test specific timeout in sec (overrides `LTP_TIMEOUT`).

* `LTP_TIMEOUT_MUL` Multiplies the timeout. Also exported as environment variable for LTP (originally LTP variable).

* `LTP_TIMEOUT_MUL_arch` (e.g. `LTP_TIMEOUT_MUL_aarch64`) Multiplicator for specific arch (overrides `LTP_TIMEOUT_MUL`, also exported as `LTP_TIMEOUT_MUL` for LTP).

=> I would add `LTP_TIMEOUT_MUL_aarch64=3` to aarch64 workers we use or to aarch64 media type.
Also add test specific timeout to testsuites, e.g. `LTP_TIMEOUT_zram01=1800`.

Also it's a question if use somehow `TIMEOUT_SCALE` (used in `ltp_syscalls_baremetal` in http://openqa.qam.suse.cz/tests/52619#settings).

Verification run: TODO

### DIFF
--- a/tests/kernel/run_ltp.pm
+++ b/tests/kernel/run_ltp.pm
@@ -16,6 +16,7 @@ use Time::HiRes qw(clock_gettime CLOCK_MONOTONIC);
 use serial_terminal;
 use Mojo::File 'path';
 use Mojo::JSON;
+use LTP::utils;
 use LTP::WhiteList;
 require bmwqemu;
 
@@ -291,10 +292,16 @@ sub run {
     my ($self, $tinfo) = @_;
     die 'Need LTP_COMMAND_FILE to know which tests to run' unless $tinfo && $tinfo->runfile;
     my $runfile = $tinfo->runfile;
-    my $timeout = get_var('LTP_TIMEOUT') || 900;
+    my $test = $tinfo->test;
+
+    # default timeout
+    my $timeout = (get_var('LTP_TIMEOUT') || 900);
+    # test specific timeout, e.g. LTP_TIMEOUT_zram01
+    $timeout = (get_var('LTP_TIMEOUT_' . $test->{name})) || $timeout;
+    $timeout *= get_ltp_mul();
+
     my $is_posix = $runfile =~ m/^\s*openposix\s*$/i;
     my $test_result_export = $tinfo->test_result_export;
-    my $test = $tinfo->test;
     my %env = %{$test_result_export->{environment}};
 
     $env{retval} = 'undefined';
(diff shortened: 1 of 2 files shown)


pevik:
@richiejp FYI

czerw:
Different timeout per architecture really makes sense. Timeout per test is nice addition,  but maybe, isn't enough to set timeout per test suite ?

pevik:
mdoucha asked:
* _Is there some specific case where we need per-test timeout override?_ => IMHO at least `ltp_kernel_misc` for `zram01`, ), but maybe not that important (arch specific timeout is IMHO more important.
* _Job variable names must be uppercase_ => at least UI allowed me to create these vars, I'll test if it works. But sure, I can use upper case to avoid potential problems in the future.

tests/kernel/run_ltp.pm
```diff
@@ -405,7 +411,21 @@ This overrides LTP_COMMAND_PATTERN.
...
+=head2 LTP_TIMEOUT_test (e.g.  LTP_TIMEOUT_zram01)
+
+Test specific timeout in sec (overrides LTP_TIMEOUT).
+
+=head2 LTP_TIMEOUT_MUL
+
+Multiplies the timeout. Also exported as environment variable for LTP
+(originally LTP variable).
+
+=head2 LTP_TIMEOUT_MUL_arch (e.g. LTP_TIMEOUT_MUL_aarch64)
```
czerw, line 425: I think all should be upper case.

tests/kernel/run_ltp.pm
```diff
@@ -291,10 +291,16 @@ sub run {
     my ($self, $tinfo) = @_;
     die 'Need LTP_COMMAND_FILE to know which tests to run' unless $tinfo && $tinfo->runfile;
     my $runfile = $tinfo->runfile;
-    my $timeout = get_var('LTP_TIMEOUT') || 900;
+    my $test = $tinfo->test;
+
+    # default timeout
+    my $timeout = (get_var('LTP_TIMEOUT') || 900);
+    # test specific timeout, e.g. LTP_TIMEOUT_zram01
+    $timeout = (get_var('LTP_TIMEOUT_' . $test->{name})) || $timeout;
```
czerw, line 299: we can have here upper case issue

tests/kernel/run_ltp.pm
pevik: Sure, I'll change it. BTW it'd be interesting to see that documented.

pevik: You both were right, variable is not set => need to be upper case.

pevik:
> Different timeout per architecture really makes sense. Timeout per test is nice addition, but maybe, isn't enough to set timeout per test suite ?

OK, let's add only timeout per arch, can come back to per test timeout in the future.

mdoucha:
We can already do per-arch timeout on the medium level if it can be set independently of LTP_ENV.

czerw:
> We can already do per-arch timeout on the medium level if it can be set independently of LTP_ENV.

That's true, we don't need separate variable per arch.

pevik:
Well, the only 

> > We can already do per-arch timeout on the medium level if it can be set independently of LTP_ENV.
> 
> That's true, we don't need separate variable per arch.

Please what is the name of variable? What am I missing? ATM the test timeout is set `my $timeout = get_var('LTP_TIMEOUT') || 900;`.

mdoucha:
> > > We can already do per-arch timeout on the medium level if it can be set independently of LTP_ENV.
> Please what is the name of the variable? What am I missing?

Medium variables have to be set separately for each arch.

pevik:
> It means that we just need to set LTP_TIMEOUT in medium settings for aarch.

Sure, it will work as a quick solution, thus ok to close this. It's just that because of need to have single testsuite running longer all tests will be running longer. It's nt a problem LTP tests which use new API, because they timeout themselves in reasonable time, the problem could be with old tests (buggy old test in syscalls will run long before timeout and take time to other tests).

PR #16902, opened 2023-04-19
containers: ensure k3s has the default serviceaccount

ilausuch:
When k3s is installed (utils::install_k3s) it needs some time to be ready to accept the creation of pods.

- Related ticket: https://progress.opensuse.org/issues/127955
- Verification run: http://openqa.suse.de/tests/10948319

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -62,6 +62,8 @@ sub install_k3s {
     script_run("mkdir -p ~/.kube");
     script_run("rm -f ~/.kube/config");
     assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml ~/.kube/config");
+    sleep 60;
+    script_retry("kubectl get serviceaccount default -o name", delay => 60, retry => 3);
     record_info('k3s', "k3s installed");
 }
 


lib/containers/k8s.pm
```diff
@@ -62,6 +62,7 @@ sub install_k3s {
     script_run("mkdir -p ~/.kube");
     script_run("rm -f ~/.kube/config");
     assert_script_run("ln -s /etc/rancher/k3s/k3s.yaml ~/.kube/config");
+    script_retry("kubectl get serviceaccount default -o name", delay => 20, retry => 10);
```
jlausuch, line 65: I would increase the delay and decrease the retries, wdyt?

jlausuch (approved):

lib/containers/k8s.pm
grisu48: If this is happening often, I would say that adding a `sleep 60` before the first try might be a good idea.

grisu48 (approved):
one suggestions but LGTM!

lib/containers/k8s.pm
ilausuch: added

PR #16909, opened 2023-04-20
extra_tests_kernel/bpftrace: Avoid ending SSH session with EOT

richiejp:
On backends using a jump host (e.g. spvm) sending EOT kills the intermediate SSH connection. So use the file download feature instead.

Also kill the background script once we are finished.

Waiting for verification: https://openqa.suse.de/tests/10953267


richiejp:
New verification: https://openqa.suse.de/tests/10953859

richiejp:
X86_64 as well: https://openqa.suse.de/tests/10953995

PR #16910, opened 2023-04-20
Improve the check of the PowerShell screen

pablo-herranz:
The PowerShell screen sometimes stays at the background and creates an incorrect scenario, because the commands are not being typed

- Related ticket: https://progress.opensuse.org/issues/126725
- Verification run:
  - https://openqa.suse.de/tests/10953266

### DIFF
--- a/tests/wsl/firstrun.pm
+++ b/tests/wsl/firstrun.pm
@@ -8,6 +8,7 @@
 
 use Mojo::Base qw(windowsbasetest);
 use testapi;
+use utils qw(enter_cmd_slow);
 use version_utils qw(is_sle);
 use wsl qw(is_sut_reg is_fake_scc_url_needed);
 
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
LGTM

tests/wsl/firstrun.pm
```diff
@@ -165,17 +166,17 @@ sub run {
...
 
     is_fake_scc_url_needed || become_root;
     assert_script_run 'cd ~';
     assert_script_run "zypper ps";
-    enter_cmd 'exit';
+    enter_cmd_slow "exit\n";
     sleep 3;
     save_screenshot;
-    is_fake_scc_url_needed || enter_cmd 'exit';
+    is_fake_scc_url_needed || enter_cmd_slow "exit\n";
```
mloviska, line 179: Isn't `enter_cmd_slow` always hitting the enter/newline?

pablo-herranz: Not sure, so I added it just in case... :man_shrugging:

PR #16911, opened 2023-04-20
Podman pods: Increase k3s timeout and uninstall it afterwards

pdostal:
- Related pull request: os-autoinst/opensuse-jobgroups#308
- Related ticket: [poo#124601](https://progress.opensuse.org/issues/124601), [poo#128069](https://progress.opensuse.org/issues/128069)
- Verification run: [Leap 15.5](https://pdostal-server.suse.cz/tests/4309)

### DIFF
--- a/tests/containers/podman_pods.pm
+++ b/tests/containers/podman_pods.pm
@@ -22,6 +22,7 @@ sub run {
     my ($self, $args) = @_;
     select_serial_terminal;
     my $podman = $self->containers_factory('podman');
+    $self->{podman} = $podman;
 
     record_info('Prep', 'Get kube yaml');
     assert_script_run('curl ' . data_url('containers/hello-kubic.yaml') . ' -o hello-kubic.yaml');
@@ -83,13 +84,26 @@ sub run {
             install_k3s();
             record_info('Test', 'kube apply');
             assert_script_run('podman kube apply --kubeconfig ~/.kube/config -f pod.yaml');
-            assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod');
+            assert_script_run('kubectl wait --timeout=240s --for=condition=Ready pod/testing-pod', timeout => 260);
             validate_script_output('kubectl exec testing-pod -- cat /etc/os-release', sub { m/SUSE Linux Enterprise Server/ });
-            uninstall_k3s();    # prevent k3s from hogging too much memory and interfere with other test runs.
         }
     }
+}
 
-    $podman->cleanup_system_host();
+sub cleanup {
+    my ($self) = @_;
+    $self->{podman}->cleanup_system_host();
+    uninstall_k3s();    # prevent k3s from hogging too much memory and interfere with other test runs.
+}
+
+sub post_run_hook {
+    my ($self) = @_;
+    $self->cleanup();
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    $self->cleanup();
 }
 
 1;


tests/containers/podman_pods.pm
```diff
@@ -81,11 +81,17 @@ sub run {
         install_k3s();
         record_info('Test', 'kube apply');
         assert_script_run('podman kube apply --kubeconfig ~/.kube/config -f pod.yaml');
-        assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod');
+        assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod', timeout => 240);
         validate_script_output('kubectl exec testing-pod -- cat /etc/os-release', sub { m/SUSE Linux Enterprise Server/ });
+        uninstall_k3s();
```
grisu48, line 86: I would put this after `$podman->cleanup_system_host();` or even better into a cleanup routine, which is executed also on test failures.

ricardobranco777 (approved):
LGTM

tests/containers/podman_pods.pm
```diff
@@ -81,11 +81,17 @@ sub run {
         install_k3s();
         record_info('Test', 'kube apply');
         assert_script_run('podman kube apply --kubeconfig ~/.kube/config -f pod.yaml');
-        assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod');
+        assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod', timeout => 240);
```
danishprakash, line 84:
```suggestion
        assert_script_run('kubectl wait --for=condition=Ready pod/testing-pod --timeout=240s');
```

nit: If doing so would not hamper the test flow

pdostal: The default for `assert_script_run()` is `90`. Should I use both `--timeout=240s` as well as `, timeout => 245`?

grisu48 (approved):
LGTM, if the VRs are passing!

tests/containers/podman_pods.pm
```diff
@@ -83,12 +84,31 @@ sub run {
...
+
+sub post_fail_hook {
+    my ($self) = @_;
+    $self->cleanup();
+}
 
-    $podman->cleanup_system_host();
+sub test_flags {
+    # Always rollback due to k3s
+    return {always_rollback => 1};
```
mloviska, line 111: Why? you are doing a cleanup anyway

tests/containers/podman_pods.pm
pdostal: Quite the oposit. The `kubectl` must end before `assert_script_run()`. Otherwise we won't know if the `kubectl` was just slow or if it had an issue.

danishprakash: Right, but what I meant was, if the default value for kubectl is something like 60s and you expect it to complete within ~120s (assuming this because you've set the wrappers to 240s), it would actually be detrimental to your workflow because nowhere would this 60s be present in this snippet. 

But looks like the default behavior is to check once and not wait[1] which I guess has been working fine for you in the past so you _could_ either overlook this suggestion or add it for posterity just to make the expectations clear.

[1] - https://github.com/kubernetes/kubectl/blob/b3d63dd09bcd124f198954e2cf8a43b3c68de207/pkg/cmd/wait/wait.go#L147

tests/containers/podman_pods.pm
grisu48: Pavel's idea was to ensure that the VM is in a well-defined state and we believe that installing `k3s` is rather intrusive. So, rolling back to a state where `k3s` is not installed might be desirable to prevent side-effects from that installation.

tests/containers/podman_pods.pm
pdostal: I hope that right now it's clear in the code. The `kubectl wait` waits for `240s` and `assert_script_run` for `260` so there is enough time for the `kubectl wait` to properly exit.

Thank you for your hint!

tests/containers/podman_pods.pm
pdostal: Removed. Thanks for the hint Martin.

tests/containers/podman_pods.pm
danishprakash: That sounds perfect!

PR #16916, opened 2023-04-20
LTP Whitelist: Add support for other job variables

mdoucha:
LTP Whitelist implementation is generic enough for non-LTP use but the LTP_KNOWN_ISSUES job variable is specific to run_ltp test module. Allow creating new LTP::Whitelist instances with custom URL passed as constructor argument.

The LTP_KNOWN_ISSUES_LOCAL variable was replaced with private class attribute cache but kept in case some tests use it explicitly.

The new argument will be used in PR #16825.

- Related ticket: N/A
- Needles: N/A
- Verification run: Pending

### DIFF
--- a/lib/LTP/WhiteList.pm
+++ b/lib/LTP/WhiteList.pm
@@ -18,13 +18,18 @@ use Mojo::UserAgent;
 use Mojo::File 'path';
 use YAML::PP;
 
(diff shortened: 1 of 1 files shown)


pevik (approved):
Thank you for a quick implementation!

PR #16919, opened 2023-04-21
move more create_hdd tests to use autoyast

rfan1:
https://progress.opensuse.org/issues/127472

- Verification run: 
[qam-regression-installation-SLED](https://openqa.suse.de/tests/10937683)
[mru-install-minimal-with-addons](https://progress.opensuse.org/issues/127472#:~:text=SLED%0Amru%2Dinstall%2Dminimal%2Dwith%2Daddons%0Amru%2Dinstall%2Ddesktop%2Dwith%2Daddons)
[mru-install-desktop-with-addons](https://openqa.suse.de/tests/overview?build=rfan0420&distri=sle)


rfan1:
> I can't recall what is the point of all the symlinks ? It's always same profile why not just use directly e.g. `12-common_base_installation.xml.ep` ? Minimal install test is using `PATTERNS=base,minimal` why the rename of the profile ?

I used softlink for Minimal install test before :) then tried to use the same style.

I agree with you on this point, as far as I know we can use the same `sle12` or `sle15` autoyast profile.  

Let me revise the code!

rfan1:
> > I can't recall what is the point of all the symlinks ? It's always same profile why not just use directly e.g. `12-common_base_installation.xml.ep` ? Minimal install test is using `PATTERNS=base,minimal` why the rename of the profile ?
> 
> I used softlink for Minimal install test before :) then tried to use the same style.
> 
> I agree with you on this point, as far as I know we can use the same `sle12` or `sle15` autoyast profile.
> 
> Let me revise the code!

Now, I use single autoyast file and remove all soft links there. however, I need to add the setting to job group.

dzedro:
Yes we have to update tests, but at the end it's simpler than always creating symlinks.

dzedro (approved):
I would first change test ay profiles to 12 or 15 common profile and then merge this, even when there is failure, we can just restart the test and after this is merged the test will work. Otherwise we have to clone job and fix the ay profile path.
We could have one yaml, because this yamls are all same, test just behave differently based on test settings, but that can be done later. :star_struck:

rfan1:
> I would first change test ay profiles to 12 or 15 common profile and then merge this, even when there is failure, we can just restart the test and after this is merged the test will work. Otherwise we have to clone job and fix the ay profile path. We could have one yaml, because this yamls are all same, test just behave differently based on test settings, but that can be done later. star_struck

Thanks for the comments, done!

foursixnine (approved):

schedule/qam/common/15-common_base_installation_ay.yaml
```diff
@@ -0,0 +1,25 @@
+---
+name: 12-common_base_installation_autoyast
```
paolostivanin, line 2: shouldn't this be called `15-...` ? @rfan1

rfan1: YES, YES

rfan1: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16996

PR #16920, opened 2023-04-21
Use matching version of upstream testsuite

ilmanzo:
The openQA module was pulling the latest version of testsuite from upstream git repository.

This was conflicting with the packaged version, giving test failures.

- Related ticket: https://progress.opensuse.org/issues/128057
- Verification run: https://openqa.suse.de/tests/10954352 

NOTE: this test is currently relevant only for aarch64.


paolostivanin:
once the CI issue is fixed, LGTM

PR #16921, opened 2023-04-21
Podman pods: As installation is conditionally disabled the uninstallation should be as well

pdostal:
- Related ticket: [poo#124601](https://progress.opensuse.org/issues/124601), [poo#128069](https://progress.opensuse.org/issues/128069)
- Previous pull request: #16911
- Verification run:


DimStar77 (approved):
looks reasonable to me

PR #16927, opened 2023-04-21
ALP test profile for automated installation of Bedrock

rakoenig:
- Related tickets: https://progress.opensuse.org/issues/126857
  (also included parts of): https://progress.opensuse.org/issues/127349
- Needles: 3 new needles created with needle editor
- [Verification runs for aarch64 and x86_64 on O3](https://openqa.opensuse.org/tests/overview?distri=alp&version=1.0.0&build=4.26&groupid=96)

- Using `agama.auto` boot parameter to launch the automatic installation
- Currently using a shell skript to execute agama commands

### DIFF
--- a/schedule/yast/opensuse/agama_autoinst.yaml
+++ b/schedule/yast/opensuse/agama_autoinst.yaml
@@ -1,0 +1,10 @@
+---
+name: agama_auto
+description: >
+  First test for autoinstallation wit agama
+vars:
+  AGAMA_AUTO: yam/agama/auto/alp.sh
+schedule:
+  - yam/agama/auto/url_prepare
+  - installation/bootloader_start
+  - yam/agama/auto/auto
--- a/tests/yam/agama/auto/auto.pm
+++ b/tests/yam/agama/auto/auto.pm
@@ -1,0 +1,23 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: First installation using D-Installer current CLI (only for development purpose)
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+use testapi;
+
+
+sub run {
+
+    assert_screen('alp-installer-ui', 120);
+
+    assert_screen('alp-installing', 60);
+
+    assert_screen('bedrock-login', 960);
+}
+
+1;
--- a/tests/yam/agama/auto/url_prepare.pm
+++ b/tests/yam/agama/auto/url_prepare.pm
@@ -1,0 +1,19 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: First installation using D-Installer current CLI (only for development purpose)
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+use testapi;
+use utils;
+
+sub run {
+    my $path = data_url(get_var('AGAMA_AUTO'));
+    set_var('EXTRABOOTPARAMS', "agama.auto=\"$path\"");
+}
+
+1;
--- a/data/yam/agama/auto/alp.sh
+++ b/data/yam/agama/auto/alp.sh
@@ -1,0 +1,8 @@
+set -ex
+
+/usr/bin/agama config set software.product=ALP-Bedrock
+/usr/bin/agama config set user.userName=joe user.password=doe
+/usr/bin/agama config set root.password=nots3cr3t
+/usr/bin/sleep 30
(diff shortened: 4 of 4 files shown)


schedule/yast/opensuse/agama_autoinst.yaml
```diff
@@ -0,0 +1,10 @@
+---
```
jknphy, line 1: let's create `schedule/yam/` and not differenciate yet between opensuse or sle, we can see later how it goes, so :
`schedule/yam/agama_auto`
as discussed offline, with can use just `auto` instead of `autoinst` is shorter and in the right folder is well understood IMO.

schedule/yast/opensuse/agama_autoinst.yaml
```diff
@@ -0,0 +1,10 @@
...
+name: agama_auto
+description: >
+  First test for autoinstallation wit agama
```
jknphy, line 4: s/wit/with/
Please don't call it first, just states what does in the description prepare blab, boot, perform auto-installation blablab.

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
...
+use strict;
+use warnings;
+
+use testapi;
+
+
+sub run {
+
+    assert_screen('alp-installer-ui', 120);
+
```
jknphy, line 17: please remove this lines in the middle of the commands.

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: First installation using D-Installer current CLI (only for development purpose)
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+use testapi;
```
jknphy, line 11: better to import only 'assert_screen' from testapi, not the whole one so we minimize the use of testapi as pattern.

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
+## Copyright 2023 SUSE LLC
```
jknphy, line 1: probably we just need `tests/yam/agama/auto.pm` for now or we plan to do more version of this, I guess this would represent some sort o general mechanism, but I would not discard it wouldn't the only one, up to you.

tests/yam/agama/auto/url_prepare.pm
```diff
@@ -0,0 +1,19 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: First installation using D-Installer current CLI (only for development purpose)
```
jknphy, line 4: Summary needs to be updated

tests/yam/agama/auto/url_prepare.pm
```diff
@@ -0,0 +1,19 @@
+## Copyright 2023 SUSE LLC
```
jknphy, line 1: this doesn't need a new test module, it can be implemented inside `bootloader_start` we don't need additional method just for this, in fact the dud is resolved in similar way I think. We can add this for follow-up ticket.

schedule/yast/opensuse/agama_autoinst.yaml
```diff
@@ -0,0 +1,10 @@
...
+vars:
+  AGAMA_AUTO: yam/agama/auto/alp.sh
```
jknphy, line 6: please move this to job group, so the same schedule could be reuse with different scripts.

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
...
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+use testapi;
+
+
+sub run {
+
+    assert_screen('alp-installer-ui', 120);
```
jknphy, line 16: we shouldn't call it "installer", The Installer is the sle/opensuse one, this has a unique name, Agama,  I would just match "Agama" text on top:
https://openqa.opensuse.org/tests/3237762#step/auto/1

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
...
+
+    assert_screen('alp-installing', 60);
```
jknphy, line 18: suggestion: `agama-installing`

tests/yam/agama/auto/auto.pm
```diff
@@ -0,0 +1,23 @@
...
+
+    assert_screen('bedrock-login', 960);
```
jknphy, line 20: to reuse for other products, we might just need `welcome to` text in the needle and a needle 'welcome-to` to signal finalization of auto-installation, wdyt?

jknphy:
I will merge this, and as my comments are more related with code organization @lemon-suse volunteer to do it.

PR #16929, opened 2023-04-24
Update for alp auto-installation

lemon-suse:
We need update for alp auto-installation following comments on PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16927

- Related ticket: https://progress.opensuse.org/issues/127349
- Needles: tags: agama-main-page, agama-installing, welcome-to, merged in OSD.
- Verification run: https://openqa.opensuse.org/tests/3241681#  x86_64
                                https://openqa.opensuse.org/tests/3241920#  aarch64


schedule/yam/agama_auto.yaml
```diff
@@ -0,0 +1,7 @@
+---
+name: agama_auto
+description: >
+  Prepare url for agama.auto boot parameter, boot, perform auto-installation for autoinstallation with agama.
```
jknphy, line 4: we can remove "for autoinstallation" in this sentence.

tests/installation/bootloader_start.pm
```diff
@@ -63,6 +63,11 @@ sub run {
         # As normally we also need `bootloader` for this scenario
         return if get_var('MEDIACHECK');
     }
+    # Prepare for agama.auto
```
jknphy, line 66: this code should be in `lib/bootloader_setup.pm`

jknphy:
regarding needles:
- better not include icon in needle, just the text as the icon might change more frequently that the font or the aliment between them: https://openqa.opensuse.org/tests/3241583#step/auto/2
- To make the needle useful for other products, we could just check Install button instead of specific product:
https://openqa.opensuse.org/tests/3241583#step/auto/1

jknphy:
For VR you need to test in both architectures.

schedule/yam/agama_auto.yaml
lemon-suse: Updated.

tests/installation/bootloader_start.pm
lemon-suse: Updated. VR is running.

lib/bootloader_setup.pm
```diff
@@ -505,6 +505,10 @@ sub bootmenu_default_params {
         }
 
     }
+    if (get_var('AGAMA_AUTO')) {
+        my $path = data_url(get_var('AGAMA_AUTO'));
+        set_var('EXTRABOOTPARAMS', "agama.auto=\"$path\"");
```
jknphy, line 510: I guess if EXTRABOOTPARAMS is already set, we will overwrite or perhaps is running before, worth a comment about it if this functionality is limited at the moment, wdyt?

jknphy:
> For VR you need to test in both architectures.

remember that you don't need to use the Playwright live-cd, but the [ALP one](https://build.opensuse.org/package/binaries/YaST:Head:Agama/agama-live:ALP/images)

lib/bootloader_setup.pm
jknphy: seems fine :525 handle those, so should be ok.

lemon-suse:
VRs updated, both for x86_64 and aarch64.

PR #16937, opened 2023-04-25
hpc: Fix few typos in mpi tests 

Avinesh:


czerw:
Commit msg could be better, something like "hpc: Fix typos in mpi tests".

Avinesh:
> Commit msg could be better, something like "hpc: Fix typos in mpi tests".

Updated, Thank you!

PR #16939, opened 2023-04-25
Increase reboot timeout value for Public Cloud

lilyeyes:
Fix patch_and_reboot failure: reboot need more wait time, for some special instances such as "ec2_r5b.metal" the reboot needs about 15 minutes.

From the serial_terminal log of VR you can see the reboot need at least 15 minutes:
```
  Reboot scheduled for Tue 2023-04-25 10:51:49 UTC, use 'shutdown -c' to cancel.   
  ...   
  Apr 25 11:06:10 vmhana01 systemd[2491]: Reached target Main User Target.   
```

[TEAM-7834](https://jira.suse.com/browse/TEAM-7834) - Fix mr_test failed on "patch_and_reboot" on AWS "r5b.metal" instance type

- Related ticket: https://jira.suse.com/browse/TEAM-7834
- Needles: NA
- Verification run: http://openqa.nue.suse.com/t10980695

### DIFF
--- a/tests/publiccloud/patch_and_reboot.pm
+++ b/tests/publiccloud/patch_and_reboot.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2019-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: zypper
@@ -30,7 +30,7 @@ sub run {
 
     ssh_fully_patch_system($remote);
 
-    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 600));
+    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 1200));
 }
 
 sub test_flags {


tests/publiccloud/patch_and_reboot.pm
```diff
@@ -30,7 +30,7 @@ sub run {
 
     ssh_fully_patch_system($remote);
 
-    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 600));
+    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 1200));
```
pdostal, line 33:
```suggestion
    $args->{my_instance}->softreboot(timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', (get_var('PUBLIC_CLOUD_SLES4SAP') ? 1200 : 600));
```

asmorodskyi: not even that , instead of tweaking default value we can just define a variable in certain case . this is the whole point of having variable here :) otherwise we can just drop variable and simply always tweak value in place :)

lilyeyes: I will withdraw this PR and using PUBLIC_CLOUD_REBOOT_TIMEOUT instead.
Thanks [asmorodskyi](https://github.com/asmorodskyi) your comment makes sense

PR #16941, opened 2023-04-26
Using Agama to perform auto-installation of Tumbleweed

coolgw:
Add alp auto script for tw
- Related ticket: https://progress.opensuse.org/issues/128183
- Needles: na
- Verification run: 
https://openqa.opensuse.org/tests/3247743  (sh)
https://openqa.opensuse.org/t3247732 (json, current blocked by suspect issue)

### DIFF
--- a/tests/yam/agama/auto.pm
+++ b/tests/yam/agama/auto.pm
@@ -14,7 +14,9 @@ use testapi 'assert_screen';
 sub run {
     assert_screen('agama-main-page', 120);
     assert_screen('agama-installing', 60);
-    assert_screen('welcome-to', 960);
+
+    my @tags = ("welcome-to", "login");
+    assert_screen \@tags, 960;
 }
 
 1;
--- a/data/yam/agama/auto/alp_tw.json
+++ b/data/yam/agama/auto/alp_tw.json
@@ -1,0 +1,14 @@
+{
+   "software": {
+      "product": "Tumbleweed"
+   },
+   "user": {
+      "fullName": "Jane Doe",
+      "password": "doe",
+      "userName": "joe"
+   },
+   "root": {
+      "password": "nots3cr3t"
+   }
+}
+


jknphy:
Please mention somewhere in the PR title description a commit that we are using Agama to perform auto-installation of Tumbleweed, there is no ALP involved here (other than the same tooling for ALP can be used for existing distris).

tests/yam/agama/auto.pm
```diff
@@ -8,13 +8,16 @@ use base 'y2_installbase';
...
-use testapi 'assert_screen';
+use testapi;
 
 
 sub run {
     assert_screen('agama-main-page', 120);
     assert_screen('agama-installing', 60);
-    assert_screen('welcome-to', 960);
+
+    my $install_product = get_var('INSTALL_PRODUCT', 'Bedrock');
```
jknphy, line 18: we are not going to create magic settings in the future, actually we will probably delete most of all of them.
In this case we end up in grub, we could just have an array of arrayref for multiple tags, we don't need to be strict on this test module, we can be more strict and fail in the next module that will be created to login the system or do something on the installed system, in that way we gain as well to not use conditional code in tests.

data/yam/agama/auto/alp-tw.sh
```diff
@@ -0,0 +1,8 @@
+set -ex
+
+/usr/bin/agama config set software.product=Tumbleweed
```
jknphy, line 3: we agreed to use a different format of .sh for each product, please pick json or jsonnet so we can cover all of them with the automation.

tests/yam/agama/auto.pm
```diff
@@ -8,13 +8,15 @@ use base 'y2_installbase';
 use strict;
 use warnings;
 
-use testapi 'assert_screen';
+use testapi;
```
lemon-suse, line 11: We'd better to include functions which used, so recover it will be ok.

data/yam/agama/auto/alp-tw.json
```diff
@@ -0,0 +1,14 @@
+{
```
jknphy, line 1: let's use always underscore, `alp_tw.json`

data/yam/agama/auto/alp-tw.sh
jknphy: let's delete the file, we want to run the same scenario for both archs.

tests/yam/agama/auto.pm
jknphy: yes, please, do not import the whole library, so we limit the number of function to use, better style.

jknphy:
> Add alp auto script for tw
> 
> * Related ticket: https://progress.opensuse.org/issues/128183
> * Needles: na
> * Verification run:
>   https://openqa.opensuse.org/tests/3247743  (sh)
>   https://openqa.opensuse.org/t3247732 (json, current blocked by suspect issue)

what is the issue with json? is it reported somewhere? issues in agama repo or bug? does it happen in latest build?

data/yam/agama/auto/alp-tw.sh
coolgw: why need delete the file? And this file should be used for both archs.

jknphy: because we talked offline that we will use for all arch installing the same product the same file and we will use different format for other products, we used for installing Bedrock with Agama .sh, so we don't repeat the same here, we use jsonnet or json. We don't need the very same scenario using json and using the script, they are functionally the same.

coolgw:
> > Add alp auto script for tw
> > 
> > * Related ticket: https://progress.opensuse.org/issues/128183
> > * Needles: na
> > * Verification run:
> >   https://openqa.opensuse.org/tests/3247743  (sh)
> >   https://openqa.opensuse.org/t3247732 (json, current blocked by suspect issue)
> 
> what is the issue with json? is it reported somewhere? issues in agama repo or bug? does it happen in latest build?
https://bugzilla.suse.com/show_bug.cgi?id=1210941

PR #16942, opened 2023-04-26
Update autoyast profile for textmode installation

chcao:
Update the autoyast profile of textmode installation on spvm, the
pattern list is not the default pattern list, so add them; Under
the 'services-manager', the default_target should be 'multi-user'
for textmode.

- Related ticket: https://progress.opensuse.org/issues/127598
- Needles: n/a
- Verification run: 
- https://openqa.suse.de/tests/10991561
- https://openqa.suse.de/tests/10991563
- https://openqa.suse.de/tests/10991562
- Related MR: https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/500

### DIFF
--- a/data/autoyast_sle15/autoyast_sle_powervm.xml.ep
+++ b/data/autoyast_sle15/autoyast_sle_powervm.xml.ep
@@ -74,7 +74,7 @@
     <ntp_policy>auto</ntp_policy>
   </ntp-client>
   <services-manager config:type="map">
-    <default_target>graphical</default_target>
+    <default_target>multi-user</default_target>
     <services config:type="map">
       <enable config:type="list">
         <service>firewalld</service>
@@ -95,6 +95,9 @@
       <pattern>basesystem</pattern>
       <pattern>enhanced_base</pattern>
       <pattern>minimal_base</pattern>
+      <pattern>x11</pattern>
+      <pattern>x11_yast</pattern>
+      <pattern>yast2_basis</pattern>
     </patterns>
     <products config:type="list">
       <product>SLES</product>


rfan1:
@chcao threre is already one for yaml schedule, you can just link it
os-autoinst-distri-opensuse/schedule/functional/autoyast_create_hdd/:[0]# cat autoyast_create_hdd_textmode.yaml
name: autoyast_create_hdd_textmode
description:    >
  Test performs autoyast installation to generate qcow2 images used by dependent functional tests.
conditional_schedule:
  svirt_upload_assets:
    ARCH:
      s390x:
        - shutdown/svirt_upload_assets
    VIRSH_VMM_FAMILY:
      xen:
        - shutdown/svirt_upload_assets
  version_specific:
    VERSION:
      'Tumbleweed':
        - update/zypper_clear_repos
        - console/zypper_ar
        - console/zypper_ref
vars:
  DESKTOP: textmode
schedule:
  - autoyast/prepare_profile
  - autoyast/installation
  - installation/first_boot
  - console/system_prepare
  - '{{version_specific}}'
  - console/hostname
  - shutdown/grub_set_bootargs
  - shutdown/cleanup_before_shutdown
  - shutdown/shutdown
  - '{{svirt_upload_assets}}'

data/autoyast_sle15/autoyast_sle_powervm.xml.ep
```diff
@@ -74,7 +74,7 @@
     <ntp_policy>auto</ntp_policy>
   </ntp-client>
   <services-manager config:type="map">
-    <default_target>graphical</default_target>
+    <default_target>multi-user</default_target>
```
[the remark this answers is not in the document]
lemon-suse: Another thinking, gnome is not supported on powerVM, so we won't set DESKTOP=gnome on powerVM. I think we can update the name for the profile, such as autoyast_sle_powervm_textmode.xml.ep so people won't feel confused. WDYT?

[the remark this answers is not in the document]
jknphy: gnome is include, it is not that is not supported, it is more like we don't have the infrastructure to test it differently and maybe IBM is not using desktop at all, so we are fine with textmode in pvm.

data/autoyast_sle15/autoyast_sle_powervm.xml.ep
```diff
@@ -88,13 +88,17 @@
     % foreach (values %$addons) {
         <package><%= lc($_->{name}) %>-release</package>
     % }
+      <package>firewalld</package>
```
jknphy, line 91: this is really strange, how existing job was able to enable firewall but wasn't necessary to add the package?
perhaps firewalld is installed by default with AutoYaST when you don't specify it explicitly?

data/autoyast_sle15/autoyast_sle_powervm.xml.ep
```diff
@@ -88,13 +88,17 @@
...
     </packages>
     <patterns config:type="list">
       <pattern>apparmor</pattern>
       <pattern>base</pattern>
       <pattern>basesystem</pattern>
       <pattern>enhanced_base</pattern>
       <pattern>minimal_base</pattern>
+      <pattern>x11</pattern>
```
jknphy, line 99: I can see here more patterns than in https://openqa.suse.de/tests/11023045#step/disable_grub_timeout/1
we don't need to use .ep template to optimize from the beginning, the most important is that will be equivalent image created.

chcao: Do you mean no need to add x11 here? The x11 pattern is one of the pattern from X Window System.

data/autoyast_sle15/autoyast_sle_powervm.xml.ep
chcao: I just noticed that it may related with x11 pattern, the tigervnc package is under x11 pattern, x11 was not added when I met the problem. I remove the firewalld package from the ay profile and let's see the result.

jknphy:
LGTM, let's merge
Could you please take a look to comments in the MR?

chcao:
> LGTM, let's merge Could you please take a look to comments in the MR?

Thanks, I updated the testsuit in MR based on the comments.

PR #16949, opened 2023-04-26
Add system_prepare to extra_test_{gnome,textmode} schedule

DimStar77:
At least on PPC, the serial device (dev/hvc0) does not retain permission
post reboot and we fail to write to it as normal user.

system_prepare takes care of setting the right permissions for us.

- Related ticket: https://progress.opensuse.org/issues/123791
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/3246598


Avinesh (approved):
looks fine!

PR #16951, opened 2023-04-27
Use bootloader_start only on s390x and ppc64le

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/127664
- Verification runs:
  - a
  - b


ilmanzo:
Very good ! 👍🏻

PR #16957, opened 2023-04-28
xfstests: Support nfs test on multi machine

lansuse:
Support xfstests run nfs test on multi machine. Set XFSTESTS_NFS_SERVER=1 to setup nfs server, and set PARALLEL_WITH for nfs client.

- Related ticket: https://progress.opensuse.org/issues/127478
- Needles: N/A
- Verification run: 
- https://openqa.suse.de/tests/11052962 (nfs server)
- https://openqa.suse.de/tests/11052963 (nfs client)
- https://openqa.suse.de/tests/10999022 (nfs test in one server)
- https://openqa.suse.de/tests/10999020 (generic test without nfs)

### DIFF
--- a/tests/xfstests/partition.pm
+++ b/tests/xfstests/partition.pm
@@ -21,10 +21,15 @@ use serial_terminal 'select_serial_terminal';
 use filesystem_utils qw(str_to_mb parted_print partition_num_by_type mountpoint_to_partition
   partition_table create_partition remove_partition format_partition);
 use File::Basename;
+use lockapi;
+use mmapi;
+use mm_network;
+use nfs_common;
 
 my $INST_DIR = '/opt/xfstests';
 my $CONFIG_FILE = "$INST_DIR/local.config";
 my $NFS_VERSION = get_var('XFSTESTS_NFS_VERSION', '4.1');
+my $NFS_SERVER_IP;
 
 # Number of SCRATCH disk in SCRATCH_DEV_POOL, other than btrfs has only 1 SCRATCH_DEV, xfstests specific
 sub partition_amount_by_homesize {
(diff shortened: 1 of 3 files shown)


tests/xfstests/partition.pm
```diff
@@ -332,8 +338,23 @@ sub run {
...
+            install_dependencies_nfs;
+            server_configure_network($self);
+            setup_nfs_server("$NFS_VERSION");
+            mutex_create('xfstests_nfs_server_ready');
+            wait_for_children;
+        }
+        elsif (get_var('PARALLEL_WITH')) {
+            setup_static_mm_network('10.0.2.102/24');
+            assert_script_run('mkdir -p /nfs/test /nfs/scratch');
+            $NFS_SERVER_IP = '10.0.2.101';
```
pevik, line 351: Having these IP set as perl constants somewhere instead of copy paste them would help searching them / understanding they are related.

tests/xfstests/run.pm
```diff
@@ -433,6 +436,13 @@ sub run {
     %BLACKLIST = (%BLACKLIST, %tests_needto_exclude);
 
     test_prepare;
+
+    # wait until nfs service is ready
+    if (get_var('PARALLEL_WITH')) {
+        mutex_wait('xfstests_nfs_server_ready');
+        script_retry('ping -c3 10.0.2.101', delay => 15, retry => 12);
```
schlad, line 443: retry 12 feels generous; if mutex_wait works as it should, the retry should not be needed, right?
Anyway -c3 is there to have some lag time, so -c3 and retry 12 feels a bit steep imho

schlad (approved):
I would personally not do it this way since you crafted the nfs mm setup in a very xfstests specific way; I mean the mm nfs set-up is tightly coupled with the xfstests test runner.
But overall I think this looks good for the purpose and could help us with better testing.

tests/xfstests/run.pm
pevik: Agree with 12 being too much. 
But IMHO `-c1` is enough, because wih `-c3` `ping` will fail if even just one of the three packets does not arrive, see:

[`man ping(8)`](https://linux.die.net/man/8/ping):
> If ping does not receive any reply packets at all it will exit with code 1. If a packet count and deadline are both specified, and fewer than count packets are received by the time the deadline has arrived, it will also exit with code 1.

Also, I double checked the code (https://github.com/iputils/iputils/blob/b3ba3ddcb4af52fb10d42d02cceb6fde9854e626/ping/ping_common.c#L935):
```
return (!rts->nreceived || (rts->deadline && rts->nreceived < rts->npackets));
```

`rts->nreceived < rts->npackets` is what we talk about.

PR #16961, opened 2023-04-28
Podman_pods: Do not test kube apply in staging

pdostal:
- Failing test: https://openqa.opensuse.org/tests/3251297#step/podman_pods/140
- Verification run: https://openqa.opensuse.org/tests/3251356

### DIFF
--- a/tests/containers/podman_pods.pm
+++ b/tests/containers/podman_pods.pm
@@ -15,7 +15,7 @@ use testapi;
 use utils qw(script_retry);
 use containers::utils qw(check_min_runtime_version);
 use serial_terminal 'select_serial_terminal';
-use version_utils qw(is_sle is_opensuse);
+use version_utils qw(is_sle is_opensuse is_staging);
 use containers::k8s qw(install_k3s uninstall_k3s);
 
 sub run {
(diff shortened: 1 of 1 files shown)


tests/containers/podman_pods.pm
```diff
@@ -78,7 +78,7 @@ sub run {
         validate_script_output('podman pod ps', sub { !m/testing-pod/ });
         validate_script_output('podman ps', sub { !m/testing-pod-container/ });
 
-        if (check_min_runtime_version('4.4.0')) {
+        if (check_min_runtime_version('4.4.0') && !is_staging) {
```
grisu48, line 81: Ok, but if this is a temporary fix, please add a comment that reminds us to work on a proper solution.

jlausuch: I don't think it's temporary. If we can't install certain packages cause in staging we don't have full repos, we can live without this test in staging.

jlausuch (approved):

tests/containers/podman_pods.pm
pdostal: Agree with Jose.

PR #16965, opened 2023-05-01
Add preparation step for SELinux

grisu48:
Eanble SELinux testing for SLEM on PC. We add additional preparation setps to install and configure SELinux, as well as to allow the ssh tunnel port to make sure we have access to openQA.

This PR enabled the code part. Job group settings will be added after merge.

- Related ticket: https://progress.opensuse.org/issues/110791
- Verification runs:
* [SLEM 5.1 slem_selinux](https://duck-norris.qe.suse.de/tests/12919)
* [SLEM 5.1 containers_selinux](https://duck-norris.qe.suse.de/tests/12917)
* [SLEM 5.2 slem_selinux](https://duck-norris.qe.suse.de/tests/12926)
* [SLEM 5.2 containers_selinux](https://duck-norris.qe.suse.de/tests/12922)
* [SLEM 5.3 slem_selinux](https://duck-norris.qe.suse.de/tests/12916)
* [SLEM 5.3 containers_selinux](https://duck-norris.qe.suse.de/tests/12918)
* [SLEM 5.4 default](https://duck-norris.qe.suse.de/tests/12924)
* [SLEM 5.4 containers](https://duck-norris.qe.suse.de/tests/12925)

### DIFF
--- a/lib/transactional.pm
+++ b/lib/transactional.pm
@@ -23,6 +23,7 @@ use version_utils;
 use utils 'reconnect_mgmt_console';
 use Utils::Backends;
 use Utils::Architectures;
+use publiccloud::instances;
 
 our @EXPORT = qw(
   process_reboot
@@ -74,6 +75,12 @@ sub process_reboot {
     $args{trigger} //= 0;
     $args{automated_rollback} //= 0;
     $args{expected_grub} //= 1;
+
+    if (is_public_cloud) {
+        my $instance = publiccloud::instances::get_instance();
+        $instance->softreboot();    # Handled re-establishing of the required ssh tunnel and consoles
+        return;
+    }
 
     # Switch to root-console as we need VNC to check for grub and for login prompt
     my $prev_console = current_console();
--- a/tests/transactional/enable_selinux.pm
+++ b/tests/transactional/enable_selinux.pm
@@ -13,26 +13,46 @@ use strict;
 use warnings;
 use testapi;
 use transactional qw(process_reboot);
-use version_utils qw(is_sle_micro);
+use version_utils qw(is_sle_micro is_public_cloud);
 use Utils::Systemd qw(systemctl);
 use Utils::Logging 'save_and_upload_log';
+use transactional qw(trup_call check_reboot_changes process_reboot);
 
 sub run {
     select_console 'root-console';
+
+    # Until bsc#1211058 is resolved, we cannot enable SELinux via `transactional-update setup-selinux`.
+    if (is_sle_micro('=5.2')) {
+        record_soft_failure("bsc#1211058 Enabling SELinux broken via transactional-update setup-selinux");
+        return;
+    }
+
     my $audit_log = "/var/log/audit/audit.log";
 
     # auditd should be enabled
     systemctl 'is-enabled auditd';
 
     # install and enable SELinux if not done by default
-    if (script_run('test -d /sys/fs/selinux && test -e /etc/selinux/config') != 0) {
-        assert_script_run('transactional-update setup-selinux');
+    if (script_run('test -d /sys/fs/selinux && test -e /etc/selinux/config && getenforce | grep -i "enforcing" >/dev/null') != 0) {
+        trup_call('setup-selinux');
         upload_logs('/var/log/transactional-update.log');
         save_and_upload_log('rpm -qa', 'installed_pkgs.txt');
-        process_reboot(trigger => 1);
+        check_reboot_changes;
+        if (is_public_cloud) {
+            # Additional packages required for semanage
+            trup_call('pkg install policycoreutils-python-utils');
+            check_reboot_changes;
+            # allow ssh tunnel port (to openQA)
+            my $upload_port = get_required_var('QEMUPORT') + 1;
+            assert_script_run("semanage port -a -t ssh_port_t -p tcp $upload_port");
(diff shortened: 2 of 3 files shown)


tests/publiccloud/slem_prepare.pm
```diff
@@ -26,6 +26,24 @@ sub run {
         trup_call("pkg install $runtime toolbox");
         $instance->softreboot();
     }
+    if (get_var("PUBLIC_CLOUD_SELINUX")) {
+        # Enable SELinux
+        record_info("SELinux", "Enabling SELinux on this instance");
+        trup_call("pkg in policycoreutils-python-utils");
+        $instance->softreboot();
+        # This adds 'security=selinux selinux=1 enforcing=1' to GRUB_CMDLINE_LINUX_DEFAULT in /etc/default/grub
+        assert_script_run('sed \'s/^\(GRUB_CMDLINE_LINUX_DEFAULT="\)\(.*\)\("\)$/\1\2 security=selinux selinux=1 enforcing=1"/g\' -i /etc/default/grub');
```
jlausuch, line 35: SLEM uses the following command to enable selinux:
`transactional-update setup-selinux`
I would prefer to use the official way so we can also test this command which comes as part of the product.

tests/transactional/enable_selinux.pm
```diff
@@ -29,10 +39,26 @@ sub run {
         assert_script_run('transactional-update setup-selinux');
         upload_logs('/var/log/transactional-update.log');
         save_and_upload_log('rpm -qa', 'installed_pkgs.txt');
-        process_reboot(trigger => 1);
+        reboot;
+    }
+
+    # On PublicCloud we need to enable SELinux via kernel boot parameter
```
jlausuch, line 45: but this is what `transactional-update setup-selinux` does for you, or isn't not the case for PC?

grisu48: Nope.

jlausuch (approved):

tests/transactional/enable_selinux.pm
```diff
@@ -26,13 +27,25 @@ sub run {
...
+            check_reboot_changes;
+            # allow ssh tunnel port (to openQA)
+            my $upload_port = get_required_var('QEMUPORT') + 1;
+            assert_script_run("semanage port -a -t ssh_port_t -p tcp $upload_port");
+            process_reboot(trigger => 1);
+        }
     }
 
     assert_script_run('selinuxenabled');
+    validate_script_output("getenforce", sub { m/Enforcing/ });
```
jlausuch, line 46: Please test this change also in SLE Micro 5.1 and 5.2, where the default was not enforcing.

grisu48:
A large bunch of VRs restarted now.

grisu48:
New VRs started.

tests/transactional/enable_selinux.pm
```diff
@@ -25,14 +26,26 @@ sub run {
     systemctl 'is-enabled auditd';
 
     # install and enable SELinux if not done by default
-    if (script_run('test -d /sys/fs/selinux && test -e /etc/selinux/config') != 0) {
-        assert_script_run('transactional-update setup-selinux');
+    if (script_run('test -d /sys/fs/selinux && test -e /etc/selinux/config && getenforce | grep -i "enforcing" >/dev/null') != 0) {
```
jlausuch, line 29: why do you add this command `getenforce | grep -i "enforcing"` ?

jlausuch: 5.2 is enabled in `permissive` mode by default. Why forcing it to re-enable it?

grisu48:
We need to exclude this on SLEM 5.2 due to bsc#1211058

tests/transactional/enable_selinux.pm
grisu48: Because on SLEM 5.1 SELinux is installed in permissive mode and we need to enable it in enforcing mode.

lib/transactional.pm
```diff
@@ -75,6 +76,12 @@ sub process_reboot {
     $args{automated_rollback} //= 0;
     $args{expected_grub} //= 1;
 
+    if (is_public_cloud) {
+        my $instance = publiccloud::instances::get_instance();
+        $instance->softreboot();    # Handled re-establishing of the required ssh tunnel and consoles
+        return;
+    }
```
pdostal, line 83:
```suggestion
    process_reboot(trigger => 1) if (is_public_cloud());
```

grisu48: And the return?

pdostal: Yesss

tests/transactional/enable_selinux.pm
jlausuch: No.
- 5.1 is not enabled by default. And enabling it results in enforcing mode.
- 5.2 and 5.3 enabled in permissive mode. `trup_call selinux` will go to enforcing -> bsc#1211058
- 5.4 enabled in enforcing mode.

grisu48: We need to discuss this. I see that we have different opinions on what should be achieved with this test run.

grisu48: We agreed that if SELinux is enabled, then the `enable_selinux` test should set it to `enforcing`.

PR #16970, opened 2023-05-02
Containers: Increase the debbug information for k3s 

ilausuch:
- Related ticket: https://progress.opensuse.org/issues/128354
- Verification run: http://openqa.suse.de/tests/11024155

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -71,7 +71,11 @@ sub install_k3s {
     # Await k3s to be ready and exists the default service account
     script_retry("kubectl get serviceaccount default -o name", delay => 60, retry => 3);
     # Await k3s to be ready and the api is accessible
+    script_retry("kubectl get namespaces", delay => 60, retry => 3);
     script_retry("kubectl get pods --all-namespaces", delay => 60, retry => 3);
+    record_info("k3s api resources", script_output("kubectl api-resources"));
+    assert_script_run("kubectl auth can-i 'create' 'pods'");
+    assert_script_run("kubectl auth can-i 'create' 'deployments'");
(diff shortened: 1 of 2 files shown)


lib/containers/k8s.pm
```diff
@@ -72,6 +72,11 @@ sub install_k3s {
     script_retry("kubectl get serviceaccount default -o name", delay => 60, retry => 3);
     # Await k3s to be ready and the api is accessible
     script_retry("kubectl get pods --all-namespaces", delay => 60, retry => 3);
+    assert_script_run("systemctl is-active k3s");
```
grisu48, line 75: If k3s is not active, the previous command can not succeed, so this one is not necessary IMHO.

lib/containers/k8s.pm
```diff
@@ -72,6 +72,11 @@ sub install_k3s {
...
+    assert_script_run("kubectl api-resources");
```
grisu48, line 76: If this is for debugging purposes, I would consider putting it into a `record_info`, similar to lines 80 and 81.

lib/containers/k8s.pm
ilausuch: I suspect that sometimes the k3s fall down, so I would like to keep this or move to later. what do you think?

ilausuch: Ok, removed. We have the sytemctl status k3s in case of failure

lib/containers/k8s.pm
ilausuch: Done

jlausuch (approved):

mloviska (approved):

PR #16973, opened 2023-05-03
Peering for Azure IBSMirror in openqa

BillAnastasiadis:
Appropriate address ranges need to be fed to terraform, in order to avoid multiple concurrent openqa jobs trying to initiate a peering with the Azure IBS Mirror while having the same address space, which would make the peering fail.

- Related ticket: https://jira.suse.com/browse/TEAM-7704
- Verification run: 
http://openqaworker15.qa.suse.cz/tests/153220
http://openqaworker15.qa.suse.cz/tests/153221

The above 2 jobs ran in parallel, with vnet and subnet address ranges assigned using the algorithm in function `qesap_get_az_address_range`. The peering was successful and both were additionally confirmed using `az network vnet peering list`. Fail is not related to the peering.


data/sles4sap/qe_sap_deployment/mr_test_azure_maintenance.yaml
```diff
@@ -0,0 +1,38 @@
...
+#   tests/sles4sap/publiccloud/qesap_ansible.pm
+
+provider: 'azure'
+apiver: 3
+terraform:
+  provider: 'azure'
+  variables:
+    # GENERAL VARIABLES #
+    vnet_address_range: "%VNET_ADDRESS_RANGE%"
+    subnet_address_range: "%SUBNET_ADDRESS_RANGE%"
```
mpagot, line 17: `vnet_address_range` and `subnet_address_range` are the two new variables

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+=head3 qesap_get_az_address_range
+
+Query the existing maintenance deployments for azure
+and determine a name with a free adress range number
+for the current deployment
+
+=cut
+
+sub qesap_get_az_address_range {
+    my $worker = get_var("WORKER_INSTANCE");
```
mpagot, line 761: I suggest to move `get_var("WORKER_INSTANCE")` in the caller. Add an integer argument here named `slot`

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+
+=item B<VERBOSE> - wether output will be verbose or not
+
+=back
+=cut
+
+sub qesap_az_vnet_peering {
+    my (%args) = @_;
+    croak 'Missing mandatory source_group argument' unless $args{source_group};
+    croak 'Missing mandatory source_vnet argument' unless $args{source_vnet};
```
mpagot, line 799: can we think about an `az` query to obtain `vnet` from the `resource_group` ? So reduce the input arguments from 4 to 2?
```
az network vnet list -g ${MY_AZ_GROUP} --query "[0].name" -o tsv
```

Something like this (take care that I'm not sure about the zero in `[0].name`

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+    }
+
+    my $az_installed = script_run("which az");
+    if ($az_installed != 0) {
+        record_info("AZURE MISSING", "Azure command \"az\" doesn't seem installed\n");
+        return 1;
+    }
+
+    my $source_vnet_id = script_output("az network vnet show --resource-group $args{source_group} --name $args{source_vnet} --query id --output tsv");
+    record_info("[M] source vnet ID: $source_vnet_id\n") if $args{verbose};
```
mpagot, line 815: maybe you can accumulate all of these `record_info` to a single one at the end of the function.

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+
+    my $target_vnet_id = script_output("az network vnet show --resource-group $args{target_group} --name $args{target_vnet} --query id --output tsv");
+    record_info("[M] target vnet ID: $target_vnet_id\n") if $args{verbose};
+
+    my $peering_name = "$args{source_vnet}-$args{target_vnet}";
+
+    my $result = script_run("az network vnet peering create --name $peering_name --resource-group $args{source_group} --vnet-name $args{source_vnet} --remote-vnet $target_vnet_id --allow-vnet-access --output $az_output");
+    if ($result == 0) {
```
mpagot, line 823: I think you can remove this if/else and just convert `script_run` to `assert_script_run`

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+        record_info("PEERING SUCCESS (source)", "[M] Peering from $args{source_group}.$args{source_vnet} server was successful\n") if $args{verbose};
+    } else {
+        record_info("PEERING FAIL (source)", "[E:$result] Unable to setup peering from the $args{source_group}.$args{source_vnet}\n");
+        return $result;
+    }
+
+    $result = script_run("az network vnet peering create --name $peering_name --resource-group $args{target_group} --vnet-name $args{target_vnet} --remote-vnet $source_vnet_id --allow-vnet-access --output $az_output");
```
mpagot, line 830: In order to reduce code repetition, you can create a loop over a 2 element list to compose and call `az network vnet peering create`

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+    if ($result == 0) {
+        record_info("PEERING SUCCESS (target)", "[M] Peering from $args{target_group}.$args{target_vnet} server was successful\n") if $args{verbose};
+    } else {
+        record_info("PEERING FAIL (target)", "[E:$result] Unable to setup peering from $args{target_group}.$args{target_vnet}\n");
+        return $result;
+    }
+
+    if ($args{verbose}) {
```
mpagot, line 838: I think you can remove the if condition and always call `az network vnet peering show` using `assert_script_run`

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,135 @@ sub qesap_get_az_resource_group {
...
+        $result = script_run("az network vnet peering show --name $peering_name --resource-group $args{target_group} --vnet-name $args{target_vnet} --output $az_output");
+        if ($result != 0) {
+            record_info("PEERING STATUS FAIL", "[E:$result] Unable to get peering status!\n");
+            return $result;
+        } else {
+            record_info("PEERING STATUS SUCCESS");
+        }
+    }
+
+    return 0;
```
mpagot, line 849: if you move to `assert_script_run` this function does not need to return anything

lib/qesapdeployment.pm
BillAnastasiadis: We probably can - but, this will be replaced by a script soon, right? 

So we could probably do that in the script, to avoid the double effort. Better sooner than later (maybe even expand the scope of this ticket to add the script as well?)

lib/qesapdeployment.pm
BillAnastasiadis: True, but as above, I think it's better to do all that directly in the external script that we're eventually going to use... what do you think?

mpagot (approved):
LGTM

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+sub qesap_calculate_az_address_range {
+    my %args = @_;
+    croak 'Missing mandatory slot argument' unless $args{slot};
+
+    die "Invalid 'slot' argument - valid values are 1-8192" if ($args{slot} > 8192 || $args{slot} < 1);
+
+    my $offset = ($args{slot} - 1) * 8;
+
+    # addresses are of the form 10.ip2.ip3.0/21 and /24 respectively
+    #ip2 gets incremented when is >=256
```
alvarocarvajald, line 776: s/when is/when it is/

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+    my $ip2 = int($offset / 256);
```
alvarocarvajald, line 777: No need to convert this into an integer, when you're passing it below in L781 & L782 to `sprintf()` using `%d`  ... `%d` is an integer casting and will do the conversion:
```
# perl
my $t = 0.9999;
print sprintf('string created with sprintf. [%d] was a float', $t), "\n";
string created with sprintf. [0] was a float
```

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+    #ip3 gets incremented by 8 until it's >=256, then it resets
+    my $ip3 = $offset % 256;
+
+    my $vnet_address_range = sprintf("10.%d.%d.0/21", $ip2, $ip3);
+    my $subnet_address_range = sprintf("10.%d.%d.0/24", $ip2, $ip3);
+
+    return (
+        vnet_address_range => $vnet_address_range,
+        subnet_address_range => $subnet_address_range,
+    );
```
alvarocarvajald, line 787: Suggestion:

1. Remove L781-L783.
2. Replace L784-L787 with:
```
    return (
        vnet_address_range => sprintf("10.%d.%d.0/21", $ip2, $ip3),
        subnet_address_range => sprintf("10.%d.%d.0/24", $ip2, $ip3),
    )
```
No need for the extra variables IMO.

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+    my $peering_name = "$args{source_vnet}-$args{target_vnet}";
+
+    assert_script_run("az network vnet peering create --name $peering_name --resource-group $args{source_group} --vnet-name $args{source_vnet} --remote-vnet $target_vnet_id --allow-vnet-access --output table");
+    record_info("PEERING SUCCESS (source)", "[M] Peering from $args{source_group}.$args{source_vnet} server was successful\n") if $args{verbose};
+
+    assert_script_run("az network vnet peering create --name $peering_name --resource-group $args{target_group} --vnet-name $args{target_vnet} --remote-vnet $source_vnet_id --allow-vnet-access --output table");
+    record_info("PEERING SUCCESS (target)", "[M] Peering from $args{target_group}.$args{target_vnet} server was successful\n") if $args{verbose};
+
+    enter_cmd("echo \"[M] Peering status:\n\"");
+    assert_script_run("az network vnet peering show --name $peering_name --resource-group $args{target_group} --vnet-name $args{target_vnet} --output table");
```
alvarocarvajald, line 825: I'm guessing these 2 lines above will leave something like this in the serial terminal:
```
install# echo "[M] Peering status:
"
[M] Peering status:

install# az network vnet peering show --name <PEERING NAME> --resource-group <RES GRP> --vnet-name <TGT VNET> --output table
```
And then the output of the az-cli command.

Perhaps it would look better from a review standpoint instead to do:
```
record_info '[M] Peering Status',
    script_output("az network vnet peering show --name $peering_name --resource-group $args{target_group} --vnet-name $args{target_vnet} --output table");
```
Or something similar.

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+
+=back
+=cut
+
+sub qesap_delete_az_peering {
+    my (%args) = @_;
+    croak 'Missing mandatory source_group argument' unless $args{source_group};
+    croak 'Missing mandatory source_vnet argument' unless $args{source_vnet};
+    croak 'Missing mandatory target_group argument' unless $args{target_group};
+    croak 'Missing mandatory target_vnet argument' unless $args{target_vnet};
```
alvarocarvajald, line 851: Suggestion for L848-L851:
```
foreach (qw(source_group source_vnet target_group target_vnet)) {
    croak "Missing mandatory $_ argument" unless $args{$_}
}
```

alvarocarvajald (commented):
Some comments and suggestions.

BTW, this PR only adds the new functions but not their use in the test modules. Will this be done in a new PR?

BillAnastasiadis:
> Some comments and suggestions.
> 
> BTW, this PR only adds the new functions but not their use in the test modules. Will this be done in a new PR?

We decided with Michele that since this ticket is exclusively about the address range problem, I should only leave the relevant stuff in the pr. I have the modules to create and destroy the peering ready (I used them for the VRs), so it would be good if I immediately hop on an IBSM implementation ticket after this and add those as well. Wdyt?

lib/qesapdeployment.pm
```diff
@@ -32,6 +32,7 @@ use warnings;
 use Carp qw(croak);
 use Mojo::JSON qw(decode_json);
 use YAML::PP;
+use List::Util qw(any);
```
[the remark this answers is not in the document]
BillAnastasiadis: Probably leftover from previous implementation, will make sure and remove

lib/qesapdeployment.pm
BillAnastasiadis: Even nicer catch!

BillAnastasiadis: It *may* be good for visibility to someone trying to understand the code, though

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
     return $result;
 }
 
+=head3 qesap_calculate_az_address_range
```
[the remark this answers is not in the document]
BillAnastasiadis: I'm not sure that aws and gcp will have the same problem with the address ranges. I vaguely remember someone mentioning they may not. @mpagot what's your take on this?

Even if it is going to be used there as well, I don't know if the address ranges are going to be in the same format for gcp and aws. More info is needed for this one. Maybe Michele know more

lib/qesapdeployment.pm
BillAnastasiadis: True. This is mostly me trying to imitate the original bash script by Jan exactly. Will implement your solution.

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,112 @@ sub qesap_get_az_resource_group {
...
+    croak 'Missing mandatory source_group argument' unless $args{source_group};
+    croak 'Missing mandatory source_vnet argument' unless $args{source_vnet};
+    croak 'Missing mandatory target_group argument' unless $args{target_group};
+    croak 'Missing mandatory target_vnet argument' unless $args{target_vnet};
+
+    my $peering_name = "$args{source_vnet}-$args{target_vnet}";
+    my $source_cmd = "az network vnet peering delete --resource-group $args{source_group} --vnet-name $args{source_vnet} -n $peering_name";
+    my $target_cmd = "az network vnet peering delete --resource-group $args{target_group} --vnet-name $args{target_vnet} -n $peering_name";
+    assert_script_run($source_cmd);
+    assert_script_run($target_cmd);
```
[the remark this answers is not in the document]
BillAnastasiadis: Ok, on it!

alvarocarvajald:
> We decided with Michele that since this ticket is exclusively about the address range problem, I should only leave the relevant stuff in the pr. I have the modules to create and destroy the peering ready (I used them for the VRs), so it would be good if I immediately hop on an IBSM implementation ticket after this and add those as well. Wdyt?

Makes sense.

lib/qesapdeployment.pm
```diff
@@ -745,4 +749,143 @@ sub qesap_get_az_resource_group {
...
+=item B<SOURCE_VNET> - vnet of source
+
+=item B<TARGET_GROUP> - resource group of target
+
+=item B<TARGET_VNET> - vnet of target
+
+=back
+=cut
+
+sub qesap_delete_az_peering {
```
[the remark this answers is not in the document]
BillAnastasiadis: The function doesn't look at openqa variables, it expects the calling code to provide the vnet and RG. So shouldn't I leave the vnet calculation to the calling code?

What do you mean in a loop? Should I run the peering destruction command more than one time for each connection?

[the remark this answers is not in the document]
mpagot: can we obtain vnet like ... ?

```
    $args{source_vnet} = script_output('az network vnet list --resource-group ' . $args{source_group} . ' --query "[0].name" -o tsv');
    $args{target_vnet} = script_output('az network vnet list --resource-group ' . $args{target_group} . ' --query "[0].name" -o tsv');
```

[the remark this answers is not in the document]
BillAnastasiadis: yes of course. I was thinking that this would be done by the code that calls this function, but it might as well happen here. On it!

lib/qesapdeployment.pm
```diff
@@ -745,4 +748,118 @@ sub qesap_get_az_resource_group {
...
+    my (%args) = @_;
+    foreach (qw(source_group target_group)) {
+        croak "Missing mandatory $_ argument" unless $args{$_};
+    }
```
alvarocarvajald, line 846: Awww. You dropped the other 2 arguments and now the loop seems overkill for just 2 args. :P

Just to satisfy my OCD (and nothing else, feel free to ignore): L844-L846 and L801-L802 are doing mostly the same, but using different approaches. Could you use the same in both?

alvarocarvajald (approved):
LGTM.

Just an OCD-related comment, but feel free to ignore.

lib/qesapdeployment.pm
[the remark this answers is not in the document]
BillAnastasiadis: Done

pdostal:
We will need probably something similar in QE-C.

PR #16975, opened 2023-05-03
Containers: Retry k3s installation script download

ilausuch:
Sometimes the download for k3s installation fails, this commit retry in this case

- Related ticket: https://progress.opensuse.org/issues/128570
- Verification run: http://openqa.suse.de/tests/11025411

### DIFF
--- a/lib/containers/k8s.pm
+++ b/lib/containers/k8s.pm
@@ -51,7 +51,9 @@ sub install_k3s {
 
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io  -o install_k3s.sh", timeout => 180, delay => 60, retry => 3);
+        assert_script_run("INSTALL_K3S_SKIP_START=true sh install_k3s.sh --disable=metrics-server");
+        script_run("rm -f install_k3s.sh");
     }
 
     if (get_var('REGISTRY')) {


mloviska:
The delay feels a bit longer than necessary to me, but otherwise LGTM

mloviska (approved):

jlausuch (approved):

lib/containers/k8s.pm
```diff
@@ -48,7 +48,7 @@ sub install_k3s {
         assert_script_run("export INSTALL_K3S_CHANNEL=" . get_var('K3S_CHANNEL')) if (get_var('K3S_CHANNEL'));
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180, delay => 60, retry => 3);
```
grisu48, line 51: I would split the command into two stages, and only retry the download but not the installation step itself.

ilausuch: ah yes, very interesting

lib/containers/k8s.pm
```diff
@@ -51,7 +51,9 @@ sub install_k3s {
 
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io > install_k3s.sh", timeout => 180, delay => 60, retry => 3);
+        assert_script_run("INSTALL_K3S_SKIP_START=true sh install_k3s.sh --disable=metrics-server");
```
jlausuch, line 55: Does it work like this? Don't you need to `chmod +x` to `install_k3s.sh`?

lib/containers/k8s.pm
```diff
@@ -51,7 +51,9 @@ sub install_k3s {
 
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io > install_k3s.sh", timeout => 180, delay => 60, retry => 3);
```
jlausuch, line 54: Or optionally, `curl -sfL https://get.k3s.io -o install_k3s.sh`

lib/containers/k8s.pm
ilausuch: Is not necessary because we are launching using sh, that is the described way

lib/containers/k8s.pm
ilausuch: I changed this

lib/containers/k8s.pm
ilausuch: solved

lib/containers/k8s.pm
```diff
@@ -51,7 +51,9 @@ sub install_k3s {
 
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io  -o install_k3s.sh", timeout => 180, delay => 60, retry => 3);
+        assert_script_run("INSTALL_K3S_SKIP_START=true sh install_k3s.sh --disable=metrics-server");
+        assert_script_run("rm -f install_k3s.sh");
```
grisu48, line 56:
```suggestion
        script_run("rm -f install_k3s.sh");
```

No need to assert here, but also not a blocker

grisu48 (approved):
LGTM!

lib/containers/k8s.pm
grisu48: Feel free to include this suggestion or ignore it. Both ways is fine by me.

lib/containers/k8s.pm
```diff
@@ -51,7 +51,9 @@ sub install_k3s {
 
         # github.com/k3s-io/k3s#5946 - The kubectl delete namespace helm-ns-413 command freezes and does nothing
         # Note: The install script starts a k3s-server by default, unless INSTALL_K3S_SKIP_START is set to true
-        assert_script_run("curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_START=true sh -s - --disable=metrics-server", timeout => 180);
+        script_retry("curl -sfL https://get.k3s.io  -o install_k3s.sh", timeout => 180, delay => 60, retry => 3);
+        assert_script_run("INSTALL_K3S_SKIP_START=true sh install_k3s.sh --disable=metrics-server");
```
jlausuch, line 55: I think we need to increase the timeout in this command: 
https://openqa.suse.de/tests/11030338#step/run_container_in_k3s/7

PR #16869, opened 2023-04-17
Test ovn3 on sles15sp5

rfan1:
https://progress.opensuse.org/issues/127580

- Verification run: http://openqa.suse.de/tests/10931806
- **SLE15SP5 feature only**
Please help review the MR below as well
**https://gitlab.suse.de/qsf-u/qa-sle-functional-userspace/-/merge_requests/160**

PR #16976, opened 2023-05-03
Switch to openvswitch3* packages for ovs-server test on sles 15 sp5

Zaoliang:
we have now openvswitch3* packages and current packages of openvswitch belong to repository SLE-Module-Legacy15-SP5-Pool, that is why zypper_call failed

see https://progress.opensuse.org/issues/127799 for more details

### DIFF
--- a/tests/console/ovs_client.pm
+++ b/tests/console/ovs_client.pm
@@ -27,7 +27,7 @@ use serial_terminal 'select_serial_terminal';
 use lockapi;
 use utils;
 use console::ovs_utils;
-
+use version_utils;
 
 my $server_ip = "10.0.2.101";
 my $client_ip = "10.0.2.102";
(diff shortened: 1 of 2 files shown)


Zaoliang:
https://openqa.suse.de/tests/11024375
https://openqa.suse.de/tests/11024376

looks good with new packages for ovs_server and ovs_client

Zaoliang:
test runs after pr update:

https://openqa.suse.de/tests/11025434
https://openqa.suse.de/tests/11025435

rfan1:
I would suggest using job setting to control the pkg version, like:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16869

rfan1:
> I would suggest using job setting to control the pkg version, like: #16869

We may need to test openvswitch3 on TW later, then use job setting is easier

Zaoliang:
actually I think this is a replacement for openvswitch which should not be tested anymore (except maintenance for any changes from Legacy repository). 
Factory first policy should be followed, so tumbleweed will get new packages soon.
But let try with variables.md at first before I can say which way is easier :)

Zaoliang:
I think another way with variables.md and yaml schedule and code changes is better complicated. Since this is a replacement  of packages to test, I keep this way.

PR #16980, opened 2023-05-04
Add assert option in firefox_open_url

dzedro:
- Related ticket: https://progress.opensuse.org/issues/128336
- Verification run: https://openqa.suse.de/tests/overview?build=firefox&distri=sle
https://openqa.suse.de/tests/11024894#step/firefox_smoke/20 retry example

### DIFF
--- a/lib/x11test.pm
+++ b/lib/x11test.pm
@@ -607,7 +607,7 @@ sub start_clean_firefox {
     }
 
     # get rid of the reader & tracking pop-up once, first test should have milestone flag
-    $self->firefox_open_url('eu.httpbin.org/html');
+    $self->firefox_open_url('eu.httpbin.org/html', assert_loaded_url => 'firefox-urls_protocols-http');
     wait_still_screen(3);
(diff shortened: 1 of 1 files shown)


lib/x11test.pm
```diff
@@ -607,7 +607,7 @@ sub start_clean_firefox {
     }
 
     # get rid of the reader & tracking pop-up once, first test should have milestone flag
-    $self->firefox_open_url('eu.httpbin.org/html');
+    $self->firefox_open_url('eu.httpbin.org/html', assert_loaded_url => 'firefox-urls_protocols-http');
```
foursixnine, line 610: I think that at some point we should move this to an internal thing (like the html5 test page) or run an instance of it ourselves (out of the scope of this pr tho)? wdyt?

foursixnine (approved):

lib/x11test.pm
dzedro: Yes for sure

PR #16981, opened 2023-05-04
Remove ubi7 container images

jlausuch:
They now need authorization to be pulled
https://progress.opensuse.org/issues/128624


DimStar77 (approved):
Seems to make sense (assuming this is not just a random error on redhat's servers)

jlausuch:
> Seems to make sense (assuming this is not just a random error on redhat's servers)

No, I tried to pull them locally. ubi8 and ubi9 works, but ubi7 asks for authorization. Anyway, I will try to investigate why this is suddenly happening and if there is a way to still pull them.

jlausuch:
> > Seems to make sense (assuming this is not just a random error on redhat's servers)
> 
> No, I tried to pull them locally. ubi8 and ubi9 works, but ubi7 asks for authorization. Anyway, I will try to investigate why this is suddenly happening and if there is a way to still pull them.

Actually, now the only one failing is `ubi-init`... so it might be temporal issue, yes... let's wait some time.

DimStar77:
FYI: this also fails in Staging - so this blocks TW actually from progressing (in prod, error _can_ be ignored, in staging, they are fatal)

https://openqa.opensuse.org/tests/3262900#step/podman_3rd_party_images/820

jlausuch:
VR removing only `ubi7/ubi-init`: https://openqa.opensuse.org/tests/3263302#details

Vogtinator (approved):

Vogtinator:
Change looks good, just the commit + MR subject is wrong now

jlausuch:
> Change looks good, just the commit + MR subject is wrong now

amended

grisu48 (approved):

PR #16989, opened 2023-05-04
HPC: Fix expansion of $mpi in assert_script_run

b10n1k:
The variable is called inside the function which setups spack. Side note that this used to work so far until recently. Which sle15sp3 began to fail. So this should fix the issue with the missing libraries but the flow needs some investigation.

- Related ticket: https://progress.opensuse.org/issues/128597
- Verification run: 
https://openqa.suse.de/tests/overview?version=15-SP3&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%23spack_expansion

https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%23spack_expansion


czerw:
Just please  add `hpc:` prefix to commit msg. It is quite generic and could confuse others.

PR #16993, opened 2023-05-04
New test for trento about unregister

hadeskun:
New test for Trento that unregister the secondary node for a hana cluster and check the status of the cluster on Trento.

Ticket number TRNT-921

Verification run: https://openqaworker15.qa.suse.cz/tests/181279#

### DIFF
--- a/lib/trento.pm
+++ b/lib/trento.pm
@@ -36,6 +36,7 @@ use YAML::PP;
 use utils qw(script_retry);
 use testapi;
 use qesapdeployment;
+use Carp;
 
 use Exporter 'import';
 
@@ -51,6 +52,7 @@ our @EXPORT = qw(
   cluster_hdbadm
   cluster_trento_net_peering
   cluster_wait_status
+  cluster_wait_status_by_regex
   get_trento_ip
   deploy_vm
   trento_acr_azure
@@ -821,6 +823,45 @@ sub cluster_wait_status {
             join("\n\n", "Output : $show_attr",
                 'status{vmhana01} : ' . $status{vmhana01},
                 'status{vmhana02} : ' . $status{vmhana02},
+                "done : $done"));
+    }
+    die "Timeout waiting for the change" if !$done;
+}
+
+=head3 cluster_wait_status_by_regex
+
+Remotely run 'SAPHanaSR-showAttr' in a loop on $host, wait output that matches regular expression
+=over 3
+
+=item B<HOST> - Ansible name or filter for the remote host where to run 'SAPHanaSR-showAttr'
+
+=item B<TIMEOUT> - Max time to retry. Die if timeout
+
+=item B<REGULAR_EXPRESSION> - Regular expression to match the text to find
+
+=back
+=cut
+
+sub cluster_wait_status_by_regex {
+    my ($host, $regular_expression, $timeout) = @_;
+    $timeout //= bmwqemu::scale_timeout(300);
+    my $_monitor_start_time = time();
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $done = 0;
+    croak 'No regular expression provided' unless (ref $regular_expression eq 'Regexp');
+    while ((time() - $_monitor_start_time <= $timeout) && (!$done)) {
+        sleep 30;
+        my $show_attr = qesap_ansible_script_output(
+            cmd => 'SAPHanaSR-showAttr',
+            provider => $prov,
+            host => $host,
+            root => 1);
+
+        for my $line (split("\n", $show_attr)) {
+            $done = 1 if ($line =~ $regular_expression);
+        }
+        record_info("SAPHanaSR-showAttr",
+            join("\n\n", "Output : $show_attr",
                 "done : $done"));
     }
     die "Timeout waiting for the change" if !$done;
--- a/schedule/sles4sap/trento/trento.yml
+++ b/schedule/sles4sap/trento/trento.yml
@@ -17,6 +17,7 @@ schedule:
     - sles4sap/trento/cluster_deploy
     - sles4sap/trento/install_agents
     - sles4sap/trento/test_trento_web
+    - sles4sap/trento/test_hana_unregister_node
     - sles4sap/trento/test_hana_stop
     - sles4sap/trento/test_hana_restore_stopped
     - sles4sap/trento/destroy
--- a/tests/sles4sap/trento/test_hana_unregister_node.pm
+++ b/tests/sles4sap/trento/test_hana_unregister_node.pm
@@ -1,0 +1,84 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: Trento unregister the secondary node
+# Maintainer: QE-SAP <qe-sap@suse.de>, Michele Pagot <michele.pagot@suse.com>
+
+use strict;
+use warnings;
+use Mojo::Base 'publiccloud::basetest';
+use base 'consoletest';
+use testapi;
+use serial_terminal 'select_serial_terminal';
+use qesapdeployment;
+use trento;
+
+sub run {
+    my ($self) = @_;
+    select_serial_terminal;
+
+    my $secondary_host = 'vmhana02';
+    cluster_print_cluster_status($secondary_host);
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, qr/global.+true/, 900);
+
+    # Stop hana instance
+    cluster_hdbadm($secondary_host, 'HDB stop');
+    cluster_wait_status_by_regex($secondary_host, qr/miky\s+SFAIL/, 900);
+
+    # Unregistered the hana database and started
+    my $cmd_unregister = join(' ', 'hdbnsutil',
+        '-sr_unregister',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, $cmd_unregister);
+    cluster_hdbadm($secondary_host, 'HDB start');
+
+    # Register the secondary node again with the primary node
+    my $cmd_register = join(' ', 'hdbnsutil',
+        '-sr_register',
+        '--force_full_replica',
+        '--remoteHost=vmhana01',
+        '--remoteInstance=00',
+        '--replicationMode=sync',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, 'HDB stop');
+    cluster_wait_status_by_regex($secondary_host, qr/miky\s+SFAIL/, 900);
+    cluster_hdbadm($secondary_host, $cmd_register);
+    cluster_hdbadm($secondary_host, 'HDB start');
+
+    # Remove the cluster from maintenance mode
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=false",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, qr/global.+false/, 900);
+    cluster_wait_status_by_regex($secondary_host, qr/miky\s+SOK/, 900);
+    qesap_ansible_cmd(cmd => "sudo crm resource cleanup msl_SAPHana_HDB_HDB00",
+        provider => $prov,
+        filter => $secondary_host);
+
+    my $cypress_test_dir = "/root/test/test";
+    enter_cmd "cd $cypress_test_dir";
+    cypress_test_exec($cypress_test_dir, 'unregister', bmwqemu::scale_timeout(900));
+    trento_support();
+    trento_collect_scenarios('test_hana_unregister');
+}
+
+sub post_fail_hook {
+    my ($self) = shift;
+    qesap_upload_logs();
+    if (!get_var('TRENTO_EXT_DEPLOY_IP')) {
+        k8s_logs(qw(web runner));
+        trento_support();
+        trento_collect_scenarios('test_hana_unregister');
+        az_delete_group();
+    }
+    cluster_destroy();
+    $self->SUPER::post_fail_hook;
+}
+
+1;


tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,82 @@
...
+use serial_terminal 'select_serial_terminal';
+use utils 'script_retry';
+use qesapdeployment;
+use trento;
+
+sub run {
+    my ($self) = @_;
+    select_serial_terminal;
+
+    my $secondary_host = 'vmhana02';
```
mpagot, line 21: For Trento, that is Azure only, this one is fine. 

In other scenarios it does not work https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16744

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,82 @@
...
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);    
+
+    # Stop hana instance
+    $cmd_stop = join(' ', 'HDB', 'stop');
+    cluster_hdbadm($secondary_host, $cmd_stop);
```
mpagot, line 32: I think that for a such simple command you can use
```
cluster_hdbadm($secondary_host, 'HDB stop');
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,82 @@
...
+
+    # Unregistered the hana database and started
+    my $cmd_unregister = join(' ', 'hdbnsutil',
+        '-sr_unregister',
+        '--name=miky');
+    my $cmd_start = join(' ', 'HDB', 'start');        
+    cluster_hdbadm($secondary_host, $cmd_unregister);
+    cluster_hdbadm($secondary_host, $cmd_start);
```
mpagot, line 40: Like before, directly place the command string in this one too

tests/sles4sap/trento/test_hana_unregister_node.pm
mpagot: Is it not some time needed here to let Trento web to keep up with the stop state?

tests/sles4sap/trento/test_hana_unregister_node.pm
mpagot: Also here, do we not need to wait for Trento to keep up?

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok master thx for the info

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,82 @@
...
+sub run {
+    my ($self) = @_;
+    select_serial_terminal;
+
+    my $secondary_host = 'vmhana02';
+    cluster_print_cluster_status($secondary_host);
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
```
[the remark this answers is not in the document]
hadeskun: line 55:
# Remove the cluster from maintenance mode
    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=false",

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: wait applied

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: wait applied

lib/trento.pm
```diff
@@ -826,6 +828,43 @@ sub cluster_wait_status {
...
+            cmd => 'SAPHanaSR-showAttr',
+            provider => $prov,
+            host => $host,
+            root => 1);
+
+        my %status = ();
+        for my $line (split("\n", $show_attr)) {
+            $status{$1} = $line if ($line =~ $regular_expression);
+        }
+        my $hash_string = Dumper(\%status);
```
mpagot, line 854: I'm not sure about it. Is there a way to do the same without to use Data::Dumper?

schedule/sles4sap/trento/trento.yml
```diff
@@ -17,6 +17,7 @@ schedule:
     - sles4sap/trento/cluster_deploy
     - sles4sap/trento/install_agents
     - sles4sap/trento/test_trento_web
+    - sles4sap/trento/test_hana_unregister_node
     - sles4sap/trento/test_hana_stop
-    - sles4sap/trento/test_hana_restore_stopped
+    - sles4sap/trento/test_hana_restore_stopped
```
mpagot, line 22: Some spaces to be removed

lib/trento.pm
```diff
@@ -826,6 +828,43 @@ sub cluster_wait_status {
...
+        my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+        my $show_attr = qesap_ansible_script_output(
+            cmd => 'SAPHanaSR-showAttr',
+            provider => $prov,
+            host => $host,
+            root => 1);
+
+        my %status = ();
+        for my $line (split("\n", $show_attr)) {
+            $status{$1} = $line if ($line =~ $regular_expression);
```
mpagot, line 852: I think that this line:

```
$status{$1} = $line if ($line =~ $regular_expression);
```

and in particular `$1` only works if user provides a regexp using groups. There's not validation about it. can we may just use a list for `status`?

Looking at the code ... do we really need both $regular_expression and $text_to_find ?

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+# Summary: Trento unregister the secondary node
+# Maintainer: QE-SAP <qe-sap@suse.de>, Michele Pagot <michele.pagot@suse.com>
+
+use strict;
+use warnings;
+use Mojo::Base 'publiccloud::basetest';
+use base 'consoletest';
+use testapi;
+use serial_terminal 'select_serial_terminal';
+use utils 'script_retry';
```
mpagot, line 13: seems this file is not using `script_retry`

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    # Stop hana instance
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'stop'));
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
+
+    # Unregistered the hana database and started
+    my $cmd_unregister = join(' ', 'hdbnsutil',
+        '-sr_unregister',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, $cmd_unregister);
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'start'));
```
mpagot, line 40: Mind to change to 
```
cluster_hdbadm($secondary_host,  'HDB start');
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+
+    # Register the secondary node again with the primary node
+    my $cmd_register = join(' ', 'hdbnsutil',
+        '-sr_register',
+        '--force_full_replica',
+        '--remoteHost=vmhana01',
+        '--remoteInstance=00',
+        '--replicationMode=sync',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'stop'));
```
mpagot, line 51:
```
cluster_hdbadm($secondary_host,  'HDB stop');
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
+    cluster_hdbadm($secondary_host, $cmd_register);
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'start'));
```
mpagot, line 54:
```
cluster_hdbadm($secondary_host,  'HDB start');
```

lib/trento.pm
```diff
@@ -826,6 +828,43 @@ sub cluster_wait_status {
     die "Timeout waiting for the change" if !$done;
 }
 
+=head3 cluster_wait_status_by_regex
+
+Remotly run 'SAPHanaSR-showAttr' in a loop on $host, wait output that match using regular expression
+=cut
+
+sub cluster_wait_status_by_regex {
+    my ($host, $timeout, $regular_expression, $text_to_find) = @_;
```
alvarocarvajald, line 837: This is related to the comment from @mpagot in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16993/files#r1197935655

Some of the parameters are clear (`$host`, `$timeout`, `$text_to_find`), but the third one isn't, so either rename it so its purpose it's clear, or add a summary of what each is used for as suggested by Michele.

Additionally, are any of the parameters mandatory or can they be dropped? The line below seems to imply that `$timeout` is not always mandatory as it has a default value, but then if `$host`, `$regular_expression` and `$text_to_find` are not mandatory, there is no check in the code verifying that they actually contain something.

I suppose that if either `$regular_expression` or `$text_to_find` are empty or undef, then the function will not do anything.

lib/trento.pm
```diff
@@ -826,6 +828,43 @@ sub cluster_wait_status {
...
+    $timeout //= bmwqemu::scale_timeout(300);
+    my $_monitor_start_time = time();
+    my $done;
+    while ((time() - $_monitor_start_time <= $timeout) && (!$done)) {
+        sleep 30;
+        my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
```
alvarocarvajald, line 843: This line can be moved before the `while()`. I don't expect the setting to change inside the loop.

lib/trento.pm
alvarocarvajald, line 852: What happens if `$regular_expression` does not have a capture group? In that case `$1` will be undef and the assignment into `%status` could get into problems. In fact, I don't see any capture groups in the regexps used in the module. How is this working?

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+true/, "true");
+
+    # Stop hana instance
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'stop'));
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
```
alvarocarvajald, line 33: Here's another instance, this time with `SFAIL`.

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+
+    # Unregistered the hana database and started
+    my $cmd_unregister = join(' ', 'hdbnsutil',
+        '-sr_unregister',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, $cmd_unregister);
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'start'));
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
alvarocarvajald, line 41: And another with `SOK`

lib/trento.pm
alvarocarvajald, line 854: I don't understand the logic behind this block between lines 851 & 854. Basically you parse the lines from the showAttr output so that only the ones that match the `$regular_expression` go into `%status`, but then you don't do anything with the structure created in `%status` and instead turn it into a string with `Dumper` where you look for the other expression.

Why not check directly in the showAttr output if you're going to check a string with a regexp?

From my other comments, I can see that the expressions in use are:

```
m/global.+true/, "true"
m/miky\s+SFAIL/, "SFAIL"
m/global.+false/, "false"
m/miky\s+SOK/, "SOK"
```

Going back to my 2 regular expressions question, seems to me the whole thing can be accomplished by either matching directly into `$show_attr` with a more precise multi-line regular expression, or - better yet - dropping `$text_to_find` and simply doing `$done = 1` when you see a line that matches to `$regular_expression`. No need to create the `%status` hash, or to then turn it into a string using `Dumper`.

What you're doing would be akin to doing the following in python:
```
import re
status = {}
for line in show_attr.split("\n"):
    result = re.match(regular_expression, line)
    if result:
        status[result.group(1)] = line
if re.match(text_to_find, str(status)):
    done = 1
else
    done = 0
```
Notice that nothing is done with `status` except turning it into a string? It's defined as a dict, but it could easily be a list, a concatenated string or some other data type. Same issue is present in the Perl code.

alvarocarvajald (changes requested):

alvarocarvajald:
There are 10 commits, and only 3 file changes. Please squash the commits. Additionally, commits such as https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16993/commits/3b85f28861066aed01b924599cd7e157e2c43195 and similar which are merge commits, should not be part of the PR.

hadeskun:
> There are 10 commits, and only 3 file changes. Please squash the commits. Additionally, commits such as [3b85f28](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/3b85f28861066aed01b924599cd7e157e2c43195) and similar which are merge commits, should not be part of the PR.

lib/trento.pm
hadeskun: yes I will remove the Data::Dumper

schedule/sles4sap/trento/trento.yml
hadeskun: thx for the comment, will fix this

lib/trento.pm
```diff
@@ -826,6 +828,43 @@ sub cluster_wait_status {
     die "Timeout waiting for the change" if !$done;
 }
 
+=head3 cluster_wait_status_by_regex
+
+Remotly run 'SAPHanaSR-showAttr' in a loop on $host, wait output that match using regular expression
+=cut
```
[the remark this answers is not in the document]
hadeskun: thx for the comment, adding doc

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: thx for the comment, will remove script_retry

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+true/, "true");
+
+    # Stop hana instance
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'stop'));
```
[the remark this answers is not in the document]
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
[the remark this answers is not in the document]
hadeskun: removed

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+
+    my $secondary_host = 'vmhana02';
+    cluster_print_cluster_status($secondary_host);
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+true/, "true");
```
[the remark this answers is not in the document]
hadeskun: good point using just one

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    # Register the secondary node again with the primary node
+    my $cmd_register = join(' ', 'hdbnsutil',
+        '-sr_register',
+        '--force_full_replica',
+        '--remoteHost=vmhana01',
+        '--remoteInstance=00',
+        '--replicationMode=sync',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'stop'));
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
```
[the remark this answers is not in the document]
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    cluster_hdbadm($secondary_host, $cmd_register);
+    cluster_hdbadm($secondary_host, join(' ', 'HDB', 'start'));
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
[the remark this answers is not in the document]
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+
+    # Remove the cluster from maintenance mode
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=false",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+false/, "false");
```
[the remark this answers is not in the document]
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,87 @@
...
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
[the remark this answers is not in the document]
hadeskun: refactored

lib/trento.pm
hadeskun: ok

lib/trento.pm
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: refactored

lib/trento.pm
hadeskun: thx for the comment, refactored

alvarocarvajald:
Seems merge pipelines are failing on lines with `croak`:
```
String found where operator expected at /__w/os-autoinst-distri-opensuse/os-autoinst-distri-opensuse/lib/trento.pm line 850, near "croak "Not regular expression provided""
```
You will need to add `use Carp;` to the module. I thought it was there before.

lib/trento.pm
```diff
@@ -826,6 +827,52 @@ sub cluster_wait_status {
...
+=back
+=cut
+
+sub cluster_wait_status_by_regex {
+    my ($host, $timeout, $regular_expression) = @_;
+    $timeout //= bmwqemu::scale_timeout(300);
+    my $_monitor_start_time = time();
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    my $done;
+    croak "Not regular expression provided" if ($regular_expression eq "") ;
```
alvarocarvajald, line 850: Text for `croak` would be better as `'No regular expression provided'` (single quotes, and `No` instead of `Not`).

lib/trento.pm
```diff
@@ -826,6 +827,52 @@ sub cluster_wait_status {
...
+    while ((time() - $_monitor_start_time <= $timeout) && (!$done)) {
```
mpagot, line 851: `$done` used here but not initialized

alvarocarvajald: > `$done` used here but not initialized

No need to. It's defined in L849 without initialization, so it's `undef` which is `false` and `!$done` would be `true`.

lib/trento.pm
```diff
@@ -826,6 +827,52 @@ sub cluster_wait_status {
...
+        my @status = ();
+        for my $line (split("\n", $show_attr)) {
+            push @status , $line if ($line =~ $regular_expression);
+        }
+        if (scalar @status > 0) {
+            $done = 1;
+        }
+        else {
+            $done = 0;
+        }
```
alvarocarvajald, line 868: While code looks better with the `@status` list instead of the `%status` hash, I still don't see the need for it.

Basically `$done` will be set to one if there are elements in the list or to zero if the list is empty. List is declared empty and populated whenever the loop finds a line which matches `$regular_expression`. Then, why not simply do:
```
        for my $line (split("\n", $show_attr)) {
            $done = 1 if ($line =~ $regular_expression);
        }
```
And remove `@status` completely?

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+
+    my $secondary_host = 'vmhana02';
+    cluster_print_cluster_status($secondary_host);
+
+    # Set secondary node in maintenance mode
+    my $prov = get_required_var('PUBLIC_CLOUD_PROVIDER');
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=true",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+true/, "true");
```
alvarocarvajald, line 28: Now that the function receives 1 regexp instead of 2, I guess this call will need to be changed to:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/global.+true/);
```
Regarding `qr//` vs. `m//`, sorry I did not comment on this before. Using `m//` will make Perl try to match the regexp at the moment of the call to whatever is in `$_` and then pass the result of that operation to the function. This is not what you want ... what you want to do is pass the regexp to the function. This can be accomplished with `qr//` (quote regexp).

See more details here: https://perldoc.perl.org/perlop#Regexp-Quote-Like-Operators

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+
+    # Stop hana instance
+    cluster_hdbadm($secondary_host, 'HDB stop');
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
```
alvarocarvajald, line 32: Same comment as before.
```
cluster_wait_status_by_regex($secondary_host, 900, qr/miky\s+SFAIL/);
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+
+    # Unregistered the hana database and started
+    my $cmd_unregister = join(' ', 'hdbnsutil',
+        '-sr_unregister',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, $cmd_unregister);
+    cluster_hdbadm($secondary_host, 'HDB start');
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
alvarocarvajald, line 40: Same comment as before:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/miky\s+SOK/);
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+    # Register the secondary node again with the primary node
+    my $cmd_register = join(' ', 'hdbnsutil',
+        '-sr_register',
+        '--force_full_replica',
+        '--remoteHost=vmhana01',
+        '--remoteInstance=00',
+        '--replicationMode=sync',
+        '--name=miky');
+    cluster_hdbadm($secondary_host, 'HDB stop');
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SFAIL/, "SFAIL");
```
alvarocarvajald, line 51: Same comment as before:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/miky\s+SFAIL/);
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+    cluster_hdbadm($secondary_host, $cmd_register);
+    cluster_hdbadm($secondary_host, 'HDB start');
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
alvarocarvajald, line 54: Same comment as before:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/miky\s+SOK/);
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+
+    # Remove the cluster from maintenance mode
+    qesap_ansible_cmd(cmd => "sudo crm configure property maintenance-mode=false",
+        provider => $prov,
+        filter => $secondary_host);
+    cluster_wait_status_by_regex($secondary_host, 900, m/global.+false/, "false");
```
alvarocarvajald, line 60: Same comment as before:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/global.+false/);
```

tests/sles4sap/trento/test_hana_unregister_node.pm
```diff
@@ -0,0 +1,86 @@
...
+    cluster_wait_status_by_regex($secondary_host, 900, m/miky\s+SOK/, "SOK");
```
alvarocarvajald, line 61: Same comment as before:
```
cluster_wait_status_by_regex($secondary_host, 900, qr/miky\s+SOK/);
```

alvarocarvajald (changes requested):

alvarocarvajald:
> Seems merge pipelines are failing on lines with `croak`:
> 
> ```
> String found where operator expected at /__w/os-autoinst-distri-opensuse/os-autoinst-distri-opensuse/lib/trento.pm line 850, near "croak "Not regular expression provided""
> ```
> 
> You will need to add `use Carp;` to the trento library. I thought it was there before.

BTW, since the `croak` / `Carp` error was not seen on a verification run, I suppose no VR was triggered after the latest changes. Please run a VR after you apply the next changes to confirm test code is working.

lib/trento.pm
alvarocarvajald: Another thing that can be done to confirm the function is getting a regular expression in that variable would be to do this:
```
croak 'No regular expression provided' unless (ref $regular_expression eq 'Regexp');
```

lib/trento.pm
hadeskun: thx for the comment

lib/trento.pm
```diff
@@ -826,6 +827,52 @@ sub cluster_wait_status {
     die "Timeout waiting for the change" if !$done;
 }
 
+=head3 cluster_wait_status_by_regex
+
+Remotly run 'SAPHanaSR-showAttr' in a loop on $host, wait output that match using regular expression
```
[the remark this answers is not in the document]
hadeskun: ok

lib/trento.pm
hadeskun: thx for the help

lib/trento.pm
hadeskun: good tip thx

lib/trento.pm
```diff
@@ -826,6 +827,52 @@ sub cluster_wait_status {
...
+
+=item B<TIMEOUT> - Max time to retry. Die if timeout
+
+=item B<REGULAR_EXPRESSION> - Regular expression to match the text to find
+
+=back
+=cut
+
+sub cluster_wait_status_by_regex {
+    my ($host, $timeout, $regular_expression) = @_;
```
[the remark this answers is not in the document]
hadeskun: refactored

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: thx for the comment, refactored.

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

tests/sles4sap/trento/test_hana_unregister_node.pm
hadeskun: ok

lib/trento.pm
```diff
@@ -826,6 +828,46 @@ sub cluster_wait_status {
...
+    croak 'No regular expression provided' unless (ref $regular_expression eq 'Regexp');
+    while ((time() - $_monitor_start_time <= $timeout) && (!$done)) {
+        sleep 30;
+        my $show_attr = qesap_ansible_script_output(
+            cmd => 'SAPHanaSR-showAttr',
+            provider => $prov,
+            host => $host,
+            root => 1);
+
+        my @status = ();
```
alvarocarvajald, line 860: This line can be removed as the variable is no longer used.

alvarocarvajald (approved):
Only one comment from my side. In general LGTM.

Let's see with a VR if it works as expected, and then we can merge.

lib/trento.pm
hadeskun: good catch thx

mpagot (approved):
LGTM

PR #16994, opened 2023-05-04
Recover only yast logs in Agama auto-install

JRivrain:
See https://progress.opensuse.org/issues/128489

    Adding a post_fail_hook directly in the module to use the default
    installer password and just upload y2logs, as it seems like there
    is nothing that does only that in the libs.

VRs
https://openqa.opensuse.org/tests/overview?distri=alp&build=JRivrain%2Fos-autoinst-distri-opensuse%2316994&version=1.0.0
https://openqa.opensuse.org/tests/overview?version=agama-0.9&build=JRivrain%2Fos-autoinst-distri-opensuse%2316994&distri=alp

### DIFF
--- a/lib/yam/agama/agama_base.pm
+++ b/lib/yam/agama/agama_base.pm
@@ -1,0 +1,23 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: base class for Agama tests
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package yam::agama::agama_base;
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi 'select_console';
+use y2_base 'save_upload_y2logs';
+use Utils::Logging 'save_and_upload_log';
+
+sub post_fail_hook {
+    my ($self) = @_;
+    $testapi::password = 'linux';
+    select_console 'root-console';
+    y2_base::save_upload_y2logs($self, skip_logs_investigation => 1);
+    save_and_upload_log('journalctl -u agama-auto', "/tmp/agama-auto-log.txt");
+}
+
+1;
--- a/tests/yam/agama/auto.pm
+++ b/tests/yam/agama/auto.pm
@@ -4,7 +4,7 @@
 # Summary: First installation using D-Installer current CLI (only for development purpose)
 # Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
 
-use base 'y2_installbase';
+use base yam::agama::agama_base;
 use strict;
 use warnings;
 


tests/yam/agama/auto.pm
```diff
@@ -7,16 +7,25 @@
...
 sub run {
     assert_screen('agama-main-page', 120);
     assert_screen('agama-installing', 60);
+    return 42;
 
     my @tags = ("welcome-to", "login");
     assert_screen \@tags, 960;
 }
 
+sub post_fail_hook {
```
jknphy, line 24: post_fail_hook should be a new base class as suggested in the ticket, we can have our own new base class that we can use for agama to retrieve logs. With Agama we don't need to inherit all the mess with retrieving those too many logs that  are not useful, we can add progressively more logs depending on the bugs we will find in this new product.

tests/yam/agama/auto.pm
```diff
@@ -7,16 +7,25 @@
...
 use warnings;
+use serial_terminal;
 
 use testapi 'assert_screen';
 
 
 sub run {
     assert_screen('agama-main-page', 120);
     assert_screen('agama-installing', 60);
+    return 42;
```
jknphy, line 18: please try to communicate which code serves to simulate failures to distinguish it from the one to be merged.

tests/yam/agama/auto.pm
```diff
@@ -7,16 +7,25 @@
...
 
     my @tags = ("welcome-to", "login");
     assert_screen \@tags, 960;
 }
 
+sub post_fail_hook {
+    my ($self) = shift;
+    $testapi::password = 'linux';
+    select_serial_terminal;
```
jknphy, line 27: there were some complains in progress ticket about not knowing what is the failure due to so many screens (but it is openQA design, no doubt), could you please check how would look with normal root console all this block here  https://openqa.opensuse.org/tests/3265003#step/auto/55 ?

lib/Agama/agama_base.pm
```diff
@@ -0,0 +1,20 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: First installation using D-Installer current CLI (only for development purpose)
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Agama::agama_base;
```
jknphy, line 7: I know there are some that start by capital letter, but for this one, it is just a logical separation, I think it should `agama::agama_base`. Like "containers".

lib/Agama/agama_base.pm
```diff
@@ -0,0 +1,20 @@
...
+use base 'y2_base';
```
jknphy, line 8: could we use `opensusebasetest`? even in y2_base there are plenty of sle specific stuff.

lib/Agama/agama_base.pm
jknphy: btw it should be under `lib/yam/agama` for each main folder in the repo, schedule, lib, tests, etc, we should have our yam folder and put there our best stuff.

lib/Agama/agama_base.pm
JRivrain: does not work with opensusebasetest, not sure why http://waaa-amazing.suse.cz/tests/18326

lib/yam/agama/agama_base.pm
```diff
@@ -0,0 +1,20 @@
...
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use y2_base 'save_upload_y2logs';
+
+sub post_fail_hook {
+    $testapi::password = 'linux';
+    select_console 'root-console';
+    y2_base::save_upload_y2logs;
```
coolgw, line 17: Could you also collect journalctl log in post_fail_hook?
script_run('journalctl -u agama-auto');

Another question is, when system is installed correctly with password nots3** , then this post_fail_hook will failed. How to handle this kind of situation?

lib/yam/agama/agama_base.pm
```diff
@@ -0,0 +1,20 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: base class for Agama tests
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package yam::agama::agama_base;
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
```
lemon-suse, line 11: It's better to just include function used in the lib, for this only 'select_console'.

lib/yam/agama/agama_base.pm
coolgw: BTW my PR link is:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16999

JRivrain: > Could you also collect journalctl log in post_fail_hook? script_run('journalctl -u agama-auto');
> 
> Another question is, when system is installed correctly with password nots3** , then this post_fail_hook will failed. How to handle this kind of situation? such as this case https://openqa.opensuse.org/tests/3290466#step/auto/18

save_upload_y2logs already uploads all the output from journalctl, so it should include as well include agama-auto unit's logs, I assume. For the second question, it's a good point, we probably need to set this in the module, I'm checking if that works.

lib/yam/agama/agama_base.pm
JRivrain: I assume agama_base will become bigger and will probably use more things from testapi. But for now you're right.

tests/yam/agama/auto.pm
```diff
@@ -4,19 +4,29 @@
...
 
 sub run {
+    $post_fail_pass = 'linux';
     assert_screen('agama-main-page', 120);
     assert_screen('agama-installing', 60);
 
     my @tags = ("welcome-to", "login");
     assert_screen \@tags, 960;
+    if (match_has_tag("login") || match_has_tag("welcome-to")) {
+        $post_fail_pass = 'nots3cr3t';
```
JRivrain, line 23: ideally we should set it to undef (so it would use default) and not use LIVECD or LIVETEST variables.

coolgw: Why the initial case set LIVECD or LIVETEST ?

JRivrain: I suppose it's because Agama runs on a livecd, but those variables were here for older livecds that were pretty different

lib/yam/agama/agama_base.pm
JRivrain: So the answer to that is now, after install we will switch to a different module with a different base. Agama is the installer, so it should only be used in the installer.

coolgw (approved):

PR #16999, opened 2023-05-05
Login properly into the systems installed by Agama and verify OS

coolgw:
Login properly into the systems installed by Agama and verify OS

- Related ticket: https://progress.opensuse.org/issues/128492
- Needles: na
- Verification run: https://openqa.opensuse.org/t3307952

### DIFF
--- a/lib/yam/agama/agama_base.pm
+++ b/lib/yam/agama/agama_base.pm
@@ -23,4 +23,8 @@ sub post_fail_hook {
     save_and_upload_log('journalctl -u agama-auto', "/tmp/agama-auto-log.txt");
 }
 
+sub post_run_hook {
+    $testapi::password = 'nots3cr3t';
+}
+
 1;
--- a/tests/yam/agama/validate_product.pm
+++ b/tests/yam/agama/validate_product.pm
@@ -1,0 +1,28 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Validate that the product installed by agama via /etc/os-release
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base "consoletest";
+use strict;
+use warnings;
+use testapi;
+use Config::Tiny;
+use Test::Assert ':all';
+use scheduler 'get_test_suite_data';
+
+sub run {
+    select_console 'root-console';
+
+    my $os_release_name_expected = get_test_suite_data()->{os_release_name};
+
+    my $os_release_output = script_output('cat /etc/os-release');
+    my $os_release_name = Config::Tiny->read_string($os_release_output)->{_}->{NAME};
+
+    assert_equals("\"" . $os_release_name_expected . "\"", $os_release_name, 'Wrong product NAME in /etc/os-release');
+}
+
+1;
--- a/schedule/yam/agama_auto.yaml
+++ b/schedule/yam/agama_auto.yaml
@@ -5,3 +5,4 @@ description: >
 schedule:
   - installation/bootloader_start
   - yam/agama/auto
+  - yam/agama/validate_product
--- a/test_data/yam/agama_auto_bedrock.yaml
+++ b/test_data/yam/agama_auto_bedrock.yaml
@@ -1,0 +1,2 @@
+---
+os_release_name: ALP Bedrock
--- a/test_data/yam/agama_auto_micro.yaml
+++ b/test_data/yam/agama_auto_micro.yaml
@@ -1,0 +1,2 @@
+---
+os_release_name: ALP Micro
--- a/test_data/yam/agama_auto_tumbleweed.yaml
+++ b/test_data/yam/agama_auto_tumbleweed.yaml
@@ -1,0 +1,2 @@
+---
+os_release_name: openSUSE Tumbleweed


tests/yam/agama/login.pm
```diff
@@ -0,0 +1,23 @@
+package login;
+
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: login console
+# Maintainer: Wei Gao <wegao@suse.com>
```
rakoenig, line 9: Shouldn't that line be `# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>` ?

jknphy (commented):
As discussed offline, let's check the product that we install for each case.
You could cherry-pick commits to make easier the review, so it is clear what are your commits a what the commits you need from other tasks.

coolgw:
> As discussed offline, let's check the product that we install for each case. You could cherry-pick commits to make easier the review, so it is clear what are your commits a what the commits you need from other tasks.

I have moved https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16994/files into this PR and make a small update.

coolgw:
> As discussed offline, let's check the product that we install for each case. You could cherry-pick commits to make easier the review, so it is clear what are your commits a what the commits you need from other tasks.

I have integrate Jonathan's PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16994/files and make some update on top directly.

lib/yam/agama/agama_base.pm
```diff
@@ -0,0 +1,27 @@
...
+
+sub post_fail_hook {
+    $testapi::password = 'linux';
+    select_console 'root-console';
+    y2_base::save_upload_y2logs;
+    save_and_upload_log('journalctl -u agama-auto', "/tmp/agama-auto-log.txt");
+}
+
+sub post_run_hook {
+    #In lib/main_common.pm setting LIVETEST & LIVECD will trigger logic which set password to NULL
```
JRivrain, line 22: What about not using these variables at the first place?

jknphy:
Please, rebase.

tests/console/check_os_release.pm
```diff
@@ -62,6 +62,15 @@ sub run {
         $checker{PRETTY_NAME} = $checker{NAME};
     }
 
+    if (is_alp) {
```
jknphy, line 65: this code is overly complicated, we don't need to use it, just use a library to read this kind of files and then compare from test data that value with the value read by the library, sort of 3 lines of code and useful for each product to install.
Example of use of the library is in `lib/YaST/workarounds.pm` using use `Config::Tiny;`

lib/yam/agama/agama_base.pm
```diff
@@ -20,4 +20,10 @@ sub post_fail_hook {
     save_and_upload_log('journalctl -u agama-auto', "/tmp/agama-auto-log.txt");
 }
 
+sub post_run_hook {
+    # In lib/main_common.pm setting LIVETEST & LIVECD will trigger logic which set password to NULL
```
jknphy, line 24: I asked @rakoenig to create a pre_run_hook to set `linux` as password.
The fact that the live settings set to undef the password is not really the reason, the reason is that agama has `linux` password because is based on Cockpit which has that password, and here the reason is that we need to change password to the one for the installed system when the installation succeeds. 

Therefore you could remove this comments, it might confuse us, because when alp will be in OBS we might need to implement something from scratch for livecd that will not have those settings at all.

tests/yam/agama/validate_product_installed_agama.pm
```diff
@@ -0,0 +1,28 @@
+# SUSE's openQA tests
```
jknphy, line 1: it is not specific from agama, we could use this validation somewhere else.
I would suggest `tests/yam/validations/validate_product.pm`
and in the future we could unify with `tests/console/validate_product_installed_SLES.pm`

data/yam/agama/auto/agama_auto_tumbleweed.yaml
```diff
@@ -0,0 +1,3 @@
+---
+test_data:
+  product: openSUSE Tumbleweed
```
jknphy, line 3: you are confusing here the path, look at this example:
https://openqa.suse.de/tests/11164022#settings
path should under `test_data/yam/`
And you can already include the rest of the products in other files.

tests/yam/agama/validate_product_installed_agama.pm
```diff
@@ -0,0 +1,28 @@
...
+use warnings;
+use testapi;
+use Config::Tiny;
+use Test::Assert ':all';
+use scheduler 'get_test_suite_data';
+
+sub run {
+    select_console 'root-console';
+
+    my ($product) = @{get_test_suite_data()->{test_data}}{product};
```
jknphy, line 20: when you use the right path (as suggested above) you don't need to access `test_data` hash.

data/yam/agama/auto/agama_auto_tumbleweed.yaml
jknphy, line 3: we might change the way to validate the product, the hash key should be:
`os_release_name` which is what you compare later, it might be very subtle ...but in my experience it is a mess with test data if not named properly and sometimes we have found almost similar values for similar things, like imaging there is a name and name_blabla, which is the same but lower letter and you also want to compare it, so then using exact name of what you compare in test data is faster to understand the test.

tests/yam/agama/validate_product_installed_agama.pm
```diff
@@ -0,0 +1,28 @@
...
+
+    my $os_release_output = script_output('cat /etc/os-release');
+    my $os_release_name = Config::Tiny->read_string($os_release_output)->{_}->{NAME};
```
jknphy, line 23: if you follow suggestion above, you might end up with two variables with `$os_release_name` so the way we usually disambiguate both is to use: `$os_release_name_current` and `$os_release_name_expected`. Also removing `_current` should be readable, because the enphasis in on having `_expected` in the name of variable, so it looks like proper testing :) (something that is missing in many places and in many validation in the repo, the future reader will thank us)

lib/yam/agama/agama_base.pm
```diff
@@ -20,4 +20,8 @@ sub post_fail_hook {
     save_and_upload_log('journalctl -u agama-auto', "/tmp/agama-auto-log.txt");
 }
 
+sub post_run_hook {
```
JRivrain, line 23: We no longer need this as long as we don't use the "LIVETEST" variable, the default password will be used.

coolgw: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17136 set password in pre_run_hook, so you need reset password in post hook now.

tests/yam/agama/validate_product.pm
```diff
@@ -0,0 +1,28 @@
...
+use warnings;
+use testapi;
+use Config::Tiny;
+use Test::Assert ':all';
+use scheduler 'get_test_suite_data';
+
+sub run {
+    select_console 'root-console';
+
+    my ($os_release_name_expected) = get_test_suite_data()->{os_release_name};
```
jknphy, line 20: it is not a list, just `my $os_release_name_expected`.

jknphy:
LGTM besides that minor thing, could you please run verification for 3 products?

lib/yam/agama/agama_base.pm
JRivrain: So it was decided to use "LIVETEST" in the end ? Why ?

coolgw:
> LGTM besides that minor syntax thing, could you please run verification for the 3 products?
-tw VR already done .
-micro &  bedrock will tested by another ticket(@hjluo)
-leap is blocked by issue currently

jknphy:
Please, tidy your code and we can proceed to merge.

coolgw:
> Please, tidy your code and we can proceed to merge.

Done.

PR #17012, opened 2023-05-08
Disable low space warning when testing SAP installation wizard

alvarocarvajald:
Depending on the disk size used during testing, the SAP Installation Wizard test module can cause a Low Disk Space warning to be shown on gnome which causes troubles when detecting with `assert_screen()` that the wizard has finished, as it checks for a clean gnome desktop. This commit adds the expected HANA filesystems to the ignore-paths key in `org.gnome.settings-daemon.plugins.housekeeping` so these paths are ignored and the warning not shown.

- Related ticket: N/A
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/11064820#step/wizard_hana_install/21


emiura (commented):
LGTM

PR #17013, opened 2023-05-08
WIP Fix issue with openvswitch3 packages for update test as well

Zaoliang:
see poo#128855


Zaoliang:
Keep this change for now and wait for next update test and see which repo provide ovs and from which repo.

Zaoliang:
let me keep this for a while because it can make sense for code change in future.

PR #17019, opened 2023-05-09
Prevent hot-corner invocation by mouse movement

realcharmer:
The test opened overview mode by moving the cursor to `0,0`, which invoked the Gnome hot-corner feature. This mouse movement is now redundant and can be removed.

- Related ticket: https://progress.opensuse.org/issues/128960
- Verification run: https://openqa.suse.de/tests/11072699


paolostivanin:
nice catch! LGTM

PR #17036, opened 2023-05-10
Use python36-devel on SLES12-SP5

grisu48:
On SLES12-SP5 we need to use `python36-devel` instead of `python3-devel` to provide the required dependencies for BCI-Test.

- Related ticket: https://progress.opensuse.org/issues/128900
- Verification run: [12-SP5](https://openqa.suse.de/tests/11087357) | [15-SP4](https://openqa.suse.de/tests/11087419)


jlausuch (approved):

ricardobranco777 (approved):
LGTM

PR #17037, opened 2023-05-10
Add tag for OpenStack images

jlausuch:
https://progress.opensuse.org/issues/129041

VR: https://openqa.suse.de/tests/11087684

### DIFF
--- a/lib/publiccloud/openstack.pm
+++ b/lib/publiccloud/openstack.pm
@@ -74,6 +74,7 @@ sub upload_img {
     assert_script_run("openstack image create"
           . " --disk-format qcow2"
           . " --container-format bare"
+          . " --tag openqa"
           . " --file $file $img_name", timeout => 60 * 60);
 
     my $image_id = $self->find_img($img_name);


mloviska (approved):

asmorodskyi:
this is nice to have at least something , but maybe we should also add `openqa_ttl` and use it as instruction for PCW for how long this entity should be used ? In such a case you can have "debugging" cases protected from deletion out-of-the-box ( you just set openqa_tll bigger and PCW ignore entity longer )

jlausuch:
> this is nice to have at least something , but maybe we should also add `openqa_ttl` and use it as instruction for PCW for how long this entity should be used ? In such a case you can have "debugging" cases protected from deletion out-of-the-box ( you just set openqa_tll bigger and PCW ignore entity longer )

Yes, we can do that.  The problem is that we could do it only for images... not for VMs..
In any case, we discussed yesterday that something like 2-3 days TTL by default is ok for ALL resources.  We are not "paying" for resources as in Public Clouds.

PR #17039, opened 2023-05-10
Implement YaST Control Center Page

rakoenig:
- Related ticket: https://progress.opensuse.org/issues/122218
- Verification runs:
  - [Verification run based on agama auto installation](https://openqa.opensuse.org/tests/3318470)
  - [Verification run in the Development ALP Bedrock group based on a qcow image](https://openqa.opensuse.org/tests/3323997)

### DIFF
--- a/lib/Yam/ControlCenterPage.pm
+++ b/lib/Yam/ControlCenterPage.pm
@@ -23,12 +23,11 @@ sub init {
     return $self;
 }
 
-sub get_control_center_page {
+sub _get_control_center_page {
     my ($self) = @_;
     die 'YaST Control Center is not displayed' unless $self->{btn_run}->exist();
     return $self;
 }
-
 
 # private methods collection
 
@@ -49,7 +48,7 @@ sub _apply_workaround_bsc1206966 {
 
 sub _select_software {
     my ($self) = @_;
-    $self->{slb_groups}->select('Software');
+    $self->_get_control_center_page()->{slb_groups}->select('Software');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -81,7 +80,7 @@ sub _select_software_repositories {
 
 sub _select_system {
     my ($self) = @_;
-    $self->{slb_groups}->select('System');
+    $self->_get_control_center_page()->{slb_groups}->select('System');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -120,7 +119,7 @@ sub _select_services_manager {
 
 sub _select_hardware {
     my ($self) = @_;
-    $self->{slb_groups}->select('Hardware');
+    $self->_get_control_center_page()->{slb_groups}->select('Hardware');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -137,7 +136,7 @@ sub _select_system_keyboard_layout {
 
 sub _select_network_services {
     my ($self) = @_;
-    $self->{slb_groups}->select('Network Services');
+    $self->_get_control_center_page()->{slb_groups}->select('Network Services');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -169,7 +168,7 @@ sub _select_iscsi_initiator {
 
 sub _select_security_and_users {
     my ($self) = @_;
-    $self->{slb_groups}->select('Security and Users');
+    $self->_get_control_center_page()->{slb_groups}->select('Security and Users');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -186,7 +185,7 @@ sub _select_user_and_group_management {
 
 sub _select_support {
     my ($self) = @_;
-    $self->{slb_groups}->select('Support');
+    $self->_get_control_center_page()->{slb_groups}->select('Support');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -198,7 +197,7 @@ sub _select_release_notes {
 
 sub _select_miscellaneous {
     my ($self) = @_;
-    $self->{slb_groups}->select('Miscellaneous');
+    $self->_get_control_center_page()->{slb_groups}->select('Miscellaneous');
     $self->_apply_workaround_bsc1206966();
     return $self;
 }
@@ -220,7 +219,7 @@ sub _select_vendor_driver {
 
 sub _run {
     my ($self) = @_;
-    $self->{btn_run}->click();
+    $self->_get_control_center_page()->{btn_run}->click();
 }
 
 # public methods
@@ -347,7 +346,6 @@ sub open_user_and_group_management {
 
 sub open_release_notes {
     my ($self) = @_;
-    $self->get_control_center_page();
     $self->_select_support();
     $self->_select_release_notes();
 }
@@ -372,8 +370,7 @@ sub open_vendor_driver {
 
 sub quit {
     my ($self) = @_;
-    $self->get_control_center_page();
-    $self->{btn_quit}->click();
+    $self->_get_control_center_page()->{btn_quit}->click();
 }
 
 
--- a/lib/Yam/ControlCenterPage.pm
+++ b/lib/Yam/ControlCenterPage.pm
@@ -1,0 +1,380 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Control Center
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::ControlCenterPage;
+use parent 'Yam::PageBase';
+use strict;
+use warnings;
+use testapi;    # needed because of bsc1206966
+
+sub init {
+    my $self = shift;
+    $self->{lbl_yast_cc} = $self->{app}->label({label => 'YaST Control Center'});
+    $self->{slb_groups} = $self->{app}->selectionbox({id => 'groups'});
+    $self->{slb_progs} = $self->{app}->selectionbox({id => 'progs'});
+    $self->{btn_help} = $self->{app}->button({id => 'help'});
+    $self->{btn_run} = $self->{app}->button({id => 'run'});
+    $self->{btn_quit} = $self->{app}->button({id => 'quit'});
+    return $self;
+}
+
+sub get_control_center_page {
+    my ($self) = @_;
+    die 'YaST Control Center is not displayed' unless $self->{btn_run}->exist();
+    return $self;
+}
+
+
+# private methods collection
+
+sub _apply_workaround_bsc1206966 {
+    my ($self) = @_;
+    record_soft_failure('bsc#1206966 - Item selection with REST API does not update contents');
+    my $currentgroup = $self->{slb_groups}->selected_items();
+    for (1 .. 4) { send_key 'tab'; }
+    if ($currentgroup eq 'Software') {
+        send_key 'down';
+        send_key 'up';
+    }
+    else {
+        send_key 'up';
+        send_key 'down';
+    }
+}
+
+sub _select_software {
+    my ($self) = @_;
+    $self->{slb_groups}->select('Software');
+    $self->_apply_workaround_bsc1206966();
+    return $self;
+}
+
+sub _select_online_update {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Online Update');
+}
+
+sub _select_software_management {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Software Management');
+}
+
+sub _select_add_on_products {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Add-On Products');
+}
+
+sub _select_media_check {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Media Check');
+}
+
+sub _select_software_repositories {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Software Repositories');
+}
+
+sub _select_system {
+    my ($self) = @_;
+    $self->{slb_groups}->select('System');
+    $self->_apply_workaround_bsc1206966();
+    return $self;
+}
+
+sub _select_boot_loader {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Boot Loader');
+}
+
+sub _select_date_and_time {
+    my ($self) = @_;
+    return $self;
+    $self->{slb_progs}->select('Date and Time');
+}
+
+sub _select_kernel_kdump {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Kernel Kdump');
+}
+
+sub _select_language {
+    my ($self) = @_;
+    return $self;
+    $self->{slb_progs}->select('Language');
+}
+
+sub _select_network_settings {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Network Settings');
+}
+
+sub _select_services_manager {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Services Manager');
+}
+
+sub _select_hardware {
+    my ($self) = @_;
+    $self->{slb_groups}->select('Hardware');
+    $self->_apply_workaround_bsc1206966();
+    return $self;
+}
+
+sub _select_printer {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Printer');
+}
+
+sub _select_system_keyboard_layout {
+    my ($self) = @_;
+    $self->{slb_progs}->select('System Keyboard Layout');
+}
+
+sub _select_network_services {
+    my ($self) = @_;
+    $self->{slb_groups}->select('Network Services');
+    $self->_apply_workaround_bsc1206966();
+    return $self;
+}
+
+sub _select_hostnames {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Hostnames');
+}
+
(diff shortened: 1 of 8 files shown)


lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Control Center
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::ControlCenter;
```
jknphy, line 9: s/ControlCenter/ControlCenterPage

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+    $self->{lbl_yast_cc} = $self->{app}->label({label => 'YaST Control Center'});
+    $self->{slb_groups} = $self->{app}->selectionbox({id => 'groups'});
+    $self->{slb_progs} = $self->{app}->selectionbox({id => 'progs'});
+    $self->{btn_help} = $self->{app}->button({id => 'help'});
+    $self->{btn_run} = $self->{app}->button({id => 'run'});
+    $self->{btn_quit} = $self->{app}->button({id => 'quit'});
+    return $self;
+}
+
+sub get_launcher_page {
```
jknphy, line 26: As mention in the past, let's use the names on the screen instead of cook ours, it is always less confusing (ie what does it mean a launcher for each person).
`get_control_center_page`

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Control Center
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::ControlCenter;
+use parent 'YaST::PageBase';
+use strict;
+use warnings;
+use testapi;    # needed because of workaround_bsc1206966
```
jknphy, line 13: name of the function might change:
`needed because of bsc#1206966`

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Control Center
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::ControlCenter;
+use parent 'YaST::PageBase';
```
jknphy, line 10: let's duplicate this page for now under Yam, I would like to not have dependencies with previous code, so it is easier to follow for everyone, so let's create Yam::PageBase to inherit from.

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+    my ($self) = @_;
+    $self->{slb_progs}->select('Media Check');
+}
+
+sub _select_software_repositories {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Software Repositories');
+}
+
+
```
jknphy, line 79: extra line, let's not do artificial separation.
I also would remove the private and public lines of comments, I don't think they help much, but without comments the separation between methods should be the same.
It is clear by the workaround added there that there is another section.

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+    return $self;
+}
+
+sub get_launcher_page {
+    my ($self) = @_;
+    die 'YaST Control Center is not displayed' unless $self->{lbl_yast_cc}->exist();
+    return $self;
+}
+
+sub workaround_bsc1206966 {
```
jknphy, line 32: for uniformity, verb+noun format:
`apply_workaround_bsc1206966`.

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+    my ($self) = @_;
+    my $currentgroup = $self->{slb_groups}->selected_items();
```
jknphy, line 34: missing soft-failure here.

lib/Yam/ControlCenter.pm
```diff
@@ -0,0 +1,374 @@
...
+    $self->{slb_progs}->select('Vendor Driver');
+}
+
+
+sub _run {
+    my ($self) = @_;
+    $self->{btn_run}->click();
+}
+
+# methods
```
jknphy, line 222: they are all methods, it is a way to express that some are public and some are not, but this is very from other programming languages, so I would not bother to put a comment, or if you think is helpful,
`public methods`, `private methods` for example.

jknphy (commented):
The skeleton looks good to me, you could mention in the PR title commit that this is not a "lib" is a "Page".
You are in good shape to proceed for implementing the remaining part, so now we need to access for real other pages, and you have to decide if it is too much work (probably it is) just access to 1 or 2 should be fine, so you can create a test that use this Page.

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,24 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
```
jknphy, line 3: 2023

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,24 @@
...
+# Summary: Start containerized YaST Control Center with ncurses
+#
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'basetest';
+use strict;
+use warnings;
+use testapi;
+
+my $podman_cmd = "podman container runlabel run registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/yast-mgmt-ncurses-test-api";
```
jknphy, line 15: should be inside the `run` otherwise I think it is global variable, we probably want all together, so it is easier to refactor.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,368 @@
...
+    return $self;
+}
+
+sub get_control_center_page {
+    my ($self) = @_;
+    die 'YaST Control Center is not displayed' unless $self->{lbl_yast_cc}->exist();
+    return $self;
+}
+
+sub apply_workaround_bsc1206966 {
```
jknphy, line 32: you could add `# workarounds` to be consistent with comments above this method name.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,368 @@
...
+    $self->{lbl_yast_cc} = $self->{app}->label({label => 'YaST Control Center'});
+    $self->{slb_groups} = $self->{app}->selectionbox({id => 'groups'});
+    $self->{slb_progs} = $self->{app}->selectionbox({id => 'progs'});
+    $self->{btn_help} = $self->{app}->button({id => 'help'});
+    $self->{btn_run} = $self->{app}->button({id => 'run'});
+    $self->{btn_quit} = $self->{app}->button({id => 'quit'});
+    return $self;
+}
+
+sub get_control_center_page {
```
jknphy, line 26: I missed something important on the first draft, I can see now that this method is actually sort of private (we don't use it from the test in general but can be used as public as a workaround).
I actually would recommend to name is as `_get_control_center_page`, include in private, and then the most important, we add it in front of each other action we do, example:

```
sub _select_systemd_journal {
    my ($self) = @_;
    get_control_center_page()->{slb_progs}->select('Systemd Journal');
}
```

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,24 @@
...
+
+sub run {
+    select_console('root-console');
+    my $yui_params = get_var('YUI_PARAMS');
+    enter_cmd($yui_params . " " . $podman_cmd);
+    my $control_center = $testapi::distri->get_control_center()->get_control_center_page();
```
jknphy, line 21: from the test you only specify the action, not the getter, so here you will call run directly.
You just need to store the variable $control_center to make multiple calls to a couple of modules for example.

tests/yam/agama/auto.pm
```diff
@@ -17,6 +17,7 @@ sub run {
 
     my @tags = ("welcome-to", "login");
     assert_screen \@tags, 960;
+    $testapi::password = "nots3cr3t";
```
jlausuch, line 20: isn't this already the predefined value in the testapi?

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,27 @@
...
+# Summary: Start containerized YaST Control Center with ncurses
+#
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+use testapi;
+
+
```
jknphy, line 15: extra blank line

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,27 @@
...
+sub run {
+
```
jknphy, line 17: extra blank line

products/alp/main.pm
```diff
@@ -18,7 +18,15 @@ my $distri = testapi::get_required_var('CASEDIR') . '/lib/susedistribution.pm';
 require $distri;
 testapi::set_distribution(DistributionProvider->provide());
 
-return 1 if load_yaml_schedule;
+# Containerized YaST needs proper libyui initialization
+if (load_yaml_schedule) {
+    if (YuiRestClient::is_libyui_rest_api) {
+        YuiRestClient::set_libyui_backend_vars;
```
jlausuch, line 24: Why is this block needed here? Can't you just use `YAML_SCHEDULE` variable in the jobs?

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,370 @@
...
+    _select_miscellaneous();
+    _select_vendor_driver();
+    _run();
+}
+
+sub quit {
+    my ($self) = @_;
+    $self->{btn_quit}->click();
+}
+
```
jknphy, line 368: extra blank line

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,27 @@
...
+    my $podman_cmd = "podman container runlabel run registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/yast-mgmt-ncurses-test-api";
+
+    select_console('root-console');
+    my $yui_params = get_var('YUI_PARAMS');
+    enter_cmd($yui_params . " " . $podman_cmd);
+    sleep(30);
+    my $control_center = $testapi::distri->get_control_center()->get_control_center_page();
```
jknphy, line 24: here we just need an specific wait as suggested above.

tests/yam/yast_container/yast2_control_center.pm
```diff
@@ -0,0 +1,27 @@
...
+
+
+sub run {
+
+    my $podman_cmd = "podman container runlabel run registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/yast-mgmt-ncurses-test-api";
+
+    select_console('root-console');
+    my $yui_params = get_var('YUI_PARAMS');
+    enter_cmd($yui_params . " " . $podman_cmd);
+    sleep(30);
```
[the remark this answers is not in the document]
jknphy: it is a draft, suggested solution above:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17039/files#r1202107010

schedule/yam/control_center_container.yaml
```diff
@@ -0,0 +1,11 @@
+---
+name: control_center_container
+description: >
+  Install system and then launch the container with the YaST2 control center.
+vars:
+  YUI_REST_API: 1
```
jknphy, line 6: could you please set this at the job group level? we have this historically but doesn't make sense.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,370 @@
...
+        send_key 'down';
+    }
+}
+
+# private methods collection
+
+sub _select_software {
+    my ($self) = @_;
+    $self->{slb_groups}->select('Software');
+    apply_workaround_bsc1206966();
```
jknphy, line 54: you are forgetting to return the `$self` for each single call as we still stay in the screen.
Chaining calls is quite nice, but I'm not sure if the libyui-rest-api logs are good enough, the problem was that the failing line was confusing in an error (but after stack trace was introduced might not be an issue, we need to revisit that).
In case of doubt we could follow the same design, and return the `$self`, it might be useful to trigger help menu and do it as one-liner :)

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,428 @@
...
+sub _select_media_check {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Media Check');
+    return $self;
+}
+
+sub _select_software_repositories {
+    my ($self) = @_;
+    $self->{slb_progs}->select('Software Repositories');
+    return $self;
```
jknphy, line 85: you only need `$self` after methods that click on menus on the left, "software", "system", "hardware", etc, not the submenus, once you click in submenu you are in a different screen, so  you cannot chain any command.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,428 @@
...
+        send_key 'down';
+    }
+}
+
+# private methods collection
+
+sub _select_software {
+    my ($self) = @_;
+    $self->{slb_groups}->select('Software');
+    apply_workaround_bsc1206966();
```
jknphy, line 54: also it is a method class, so should be `$self->apply_workaround_bsc1206966()`.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,428 @@
...
+
+sub get_control_center_page {
+    my ($self) = @_;
+    die 'YaST Control Center is not displayed' unless $self->{btn_run}->exist();
+    return $self;
+}
+
+# workarounds
+
+sub apply_workaround_bsc1206966 {
```
jknphy, line 34: let's make it private as well, `_apply*`

lib/Yam/ReleaseNotesPage.pm
```diff
@@ -0,0 +1,34 @@
...
+
+sub get_release_notes_page {
+    my ($self) = @_;
+    die 'Release notes dialog is not displayed' unless $self->{cmb_language}->exist();
+    return $self;
+}
+
+sub close {
+    my ($self) = @_;
+    $self->get_release_notes_page();
```
jknphy, line 30: one line as usual:
`$self->get_release_notes_page()->{btn_close}->click()`

lib/Yam/ReleaseNotesPage.pm
```diff
@@ -0,0 +1,34 @@
...
+
+sub init {
+    my $self = shift;
+    $self->{cmb_language} = $self->{app}->combobox({id => 'lang'});
+    $self->{btn_help} = $self->{app}->button({id => 'help'});
+    $self->{btn_close} = $self->{app}->button({id => 'next'});
+    return $self;
+}
+
+sub get_release_notes_page {
```
jknphy, line 22: as discussed because it is internal, let's use `_get_release_notes_page` to clear differentiate the interface.
Same for `ControlCenterPage`.

tests/yam/yast_modules/control_center.pm
```diff
@@ -0,0 +1,31 @@
...
+use testapi qw(select_console enter_cmd wait_still_screen save_screenshot get_var);
+
+sub run {
+    my $podman_cmd = "podman container runlabel run registry.opensuse.org/suse/alp/workloads/tumbleweed_containerfiles/suse/alp/workloads/yast-mgmt-ncurses-test-api";
+    select_console('root-console');
+    enter_cmd(get_var('YUI_PARAMS') . " $podman_cmd");
+    my $control_center = $testapi::distri->get_control_center();
+    $control_center->open_release_notes();
+    my $release_notes = $testapi::distri->get_release_notes();
+    wait_still_screen(2);
```
jknphy, line 22: please remove those `wait_still_screen` those are just sleep, for now with save_screenshot is enough and if we don't get the right screen, we can improve on that later. We need test that are easy to read, not full of hacks :)

Just in case is possible, is it possible to check some other control that would ensure to be ready to take the screenshot? the waiting mechanism should be done through the libyui-rest-api not the test api, only use test-api for screnshots, nothing else.

lib/Yam/ControlCenterPage.pm
```diff
@@ -0,0 +1,380 @@
...
+    return $self;
+}
+
+sub get_control_center_page {
+    my ($self) = @_;
+    die 'YaST Control Center is not displayed' unless $self->{btn_run}->exist();
+    return $self;
+}
+
+
```
jknphy, line 32: blank line.

schedule/yam/control_center_container.yaml
jknphy: we can do this later.

PR #17044, opened 2023-05-11
Fix nfs-server exportfs timeout issue

lemon-suse:
The clear_console in serial terminal is unexpected which blocked the system and not returned which caused exportfs timeout.

- Related ticket: https://progress.opensuse.org/issues/128648
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=lemon-suse%2Fos-autoinst-distri-opensuse%23fix-nfs-server-exports-timeout-issue&flavor=Online&arch=aarch64&arch=x86_64

### DIFF
--- a/lib/nfs_common.pm
+++ b/lib/nfs_common.pm
@@ -257,8 +257,6 @@ sub config_service {
 
 sub start_service {
     my ($rw, $ro) = @_;
-    # Back on the console
-    clear_console;
 
     # Server is up and running, client can use it now!
     check_nfs_ready($rw, $ro);


jknphy:
Have you tried to run the same test today without any code change? it might be green all the time because there is not networking issue. I'm not sure how it is related that the clear command would produce the problem.

Vogtinator (approved):
LGTM. Mostly because that `clear_console` is useless anyway, this is used in a serial_terminal context

jknphy (approved):

PR #17045, opened 2023-05-11
Let warewulf controller boot when controller is ready

b10n1k:
Add mutex to remove controllers rebooting until the controller is ready. Fix also the tear down operations which destroy the barriers.

Previously, as we implementated the loop, it could have finished before the
controller is ready to boot the nodes or the nodes would have to wait even if
the server was done.

- Related ticket: https://progress.opensuse.org/issues/128003
- Verification run: https://aquarius.suse.cz/tests/16626

### DIFF
--- a/tests/hpc/ww4_await_pxe_install.pm
+++ b/tests/hpc/ww4_await_pxe_install.pm
@@ -11,31 +11,24 @@
 use Mojo::Base qw(hpcbase), -signatures;
 use testapi;
 use lockapi;
+use mmapi;
 use utils;
 use mm_tests;
 use POSIX 'strftime';
 
 sub run ($self) {
-    my $i = 0;
-
-    while ($i < 10) {
-        check_screen('pxe-start');
-        record_info "pxe", "$i";
-        sleep(1);
+    if (check_screen('pxe-start')) {
         send_key "ctrl-B";
-        sleep(20);
-        type_string("reboot", lf => 1);
         save_screenshot();
-        $i++;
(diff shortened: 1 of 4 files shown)


tests/hpc/ww4_await_pxe_install.pm
```diff
@@ -11,24 +11,22 @@
...
+        record_info "pxe", "wait for controller";
     }
-
+    mutex_wait "ww4_ready";
+    barrier_wait('WWCTL_READY');
+    record_info 'WWCTL_READY', strftime("\%H:\%M:\%S", localtime);
+    type_string("reboot", lf => 1);
+    save_screenshot();
     while (1) {
         unless (check_screen('ww4-ready')) {
```
czerw, line 31: I think you can also remove this loop and just use `assert_screen 'ww4-ready', 180;`. If I understand, loop just checks screen every second till we get to login screen.  We don't screenshots or `record_info` from this loop.

schlad (changes requested):
Before looking at the code, please amend the commit to make the English correct:
"We previous implementation..."

tests/hpc/ww4_await_pxe_install.pm
b10n1k: Done. https://aquarius.suse.cz/tests/16636#step/ww4_await_pxe_install/8

schlad (approved):
looks good imo

czerw (approved):
LGTM, thanks for simplification of boot section.

PR #17050, opened 2023-05-11
Do not pipe output into a pager for wicked service

tinawang123:
Do not pipe output into a pager for wicked service

- Related ticket: https://progress.opensuse.org/issues/128150
- Verification run: https://openqa.suse.de/tests/11146505#step/wicked/1

### DIFF
--- a/tests/autoyast/wicked.pm
+++ b/tests/autoyast/wicked.pm
@@ -24,10 +24,11 @@ use strict;
 use warnings;
 use base 'consoletest';
 use testapi;
+use utils 'systemctl';
 
 sub run {
     # https://en.opensuse.org/openSUSE:Bugreport_wicked
-    enter_cmd "systemctl status wickedd.service";
+    systemctl 'status wickedd.service';
     enter_cmd "echo `wicked show all |cut -d ' ' -f 1` END | tee /dev/$serialdev";
     my $iflist = wait_serial("END", 10);
     # For poo#70453, to filter network link from mixed info of the output of wicked cmd


rakoenig:
LGTM.

chcao (approved):
LGTM

tests/autoyast/wicked.pm
```diff
@@ -27,7 +27,7 @@ use testapi;
 
 sub run {
     # https://en.opensuse.org/openSUSE:Bugreport_wicked
-    enter_cmd "systemctl status wickedd.service";
+    enter_cmd "systemctl status --no-pager wickedd.service";
```
lemon-suse, line 30: For me this looks like a workaround while not found the root cause. 
We'd better to know why wicked will not return, maybe this is a bug, or something wrong for the network setting in the profile.

tinawang123: I just fixed the ticket. For the root cause, we need check later.

jknphy: same that it does `lib/Utils/Systemd.pm` in `sub systemcl` we should use that functions for most of the cases btw.
if the status pipe out to page is normal that the automation cannot work.
now...why doesn't ellipsize lines like before, it might be that the defaults have changed ...
we should resolve that question before merging IMO.

tinawang123: Updated, use 'systemctl' function, It can work. 
Thanks for your suggestion.

rfan1:
@tinawang123 seems your change causes failure at leap test:
https://openqa.opensuse.org/tests/3299196#step/wicked/4

jknphy:
> @tinawang123 seems your change causes failure at leap test: https://openqa.opensuse.org/tests/3299196#step/wicked/4

for a dead service the command return '3' and when using the library we need to ignore failures to have similar behavior that `enter_cmd`.
As recommended, @tinawang123 it is always good practice to read the code of the function libraries we use.

PR #17052, opened 2023-05-11
Adjust schedules to avoid processing beta popup

chcao:
Adjust schedules to avoid processing beta popup when the product will not contain it in GMC in all the job groups owned by Yam.

- Related ticket: https://progress.opensuse.org/issues/129020
- Needles: n/a
- Verification run: n/a

### DIFF
--- a/schedule/yast/modify_existing_partition/modify_existing_partition@s390x.yaml
+++ b/schedule/yast/modify_existing_partition/modify_existing_partition@s390x.yaml
@@ -5,10 +5,15 @@ description: >
   create_hdd test suite.
 vars:
   YUI_REST_API: 1
+conditional_schedule:
+  access_beta_distribution:
+    BETA:
+      1:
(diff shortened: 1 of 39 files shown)


schedule/yast/modify_existing_partition/modify_existing_partition@s390x.yaml
```diff
@@ -5,10 +5,15 @@ description: >
   create_hdd test suite.
 vars:
   YUI_REST_API: 1
+conditional_schedule:
```
jknphy, line 8: we don't use conditional schedule, we tried to avoid this solution a time ago, please just remove the test module for now.

chcao: ok, I get it, I will submit a new PR, so I close this one, thanks.

PR #17058, opened 2023-05-12
Check Python3 module for 3.11 instead of 3.10

tinawang123:
python 3.10 is not shipped with SLE15 SP5 any more as it's been upgraded to python 3.11

- Related ticket: https://progress.opensuse.org/issues/128690
- Verification run: https://openqa.suse.de/tests/11111616#step/check_migration_features/29
  https://openqa.suse.de/tests/11130644#step/check_migration_features/27

### DIFF
--- a/tests/console/check_migration_features.pm
+++ b/tests/console/check_migration_features.pm
@@ -40,20 +40,20 @@ sub check_python2_dropped {
 # SLE-23610: Python3 module
 # test steps:
 # 1) activate the python3 module
-# 2) install the python310 package
-# 3) check python3.10's version which should be 3.10.X
+# 2) install the python311 package
+# 3) check python3.11's version which should be 3.11
 # 4) check python3's version
-# 5) check python310's lifecycle
+# 5) check python311's lifecycle
 sub check_python3_module {
     record_info('SLE-23610', 'Check Python3 Module');
     my $OS_VERSION = script_output("grep VERSION_ID /etc/os-release | cut -c13- | head -c -2");
     my $ARCH = get_required_var('ARCH');
     assert_script_run("SUSEConnect -p sle-module-python3/$OS_VERSION/$ARCH", timeout => 180);
-    zypper_call("se python310");
-    zypper_call("in python310");
-    assert_script_run("python3.10 --version | grep Python | grep 3.10.");
+    zypper_call("se python311");
+    zypper_call("in python311");
+    assert_script_run("python3.11 --version | grep Python | grep 3.11.");
     assert_script_run("python3 --version | grep Python | grep 3.6.");
-    assert_script_run("zypper lifecycle python310");
+    assert_script_run("zypper lifecycle python311");
 }
 
 # function to check all the features after migration


sofiasyria:
LGTM . If you have time, please run a VR for the other two tests: https://openqa.suse.de/tests/overview?arch=&flavor=&machine=&test=&modules=check_migration_features&module_re=&distri=sle&build=96.1&version=15-SP5&groupid=478#

tests/console/check_migration_features.pm
```diff
@@ -40,20 +40,25 @@ sub check_python2_dropped {
...
 sub check_python3_module {
     record_info('SLE-23610', 'Check Python3 Module');
     my $OS_VERSION = script_output("grep VERSION_ID /etc/os-release | cut -c13- | head -c -2");
     my $ARCH = get_required_var('ARCH');
     assert_script_run("SUSEConnect -p sle-module-python3/$OS_VERSION/$ARCH", timeout => 180);
-    zypper_call("se python310");
-    zypper_call("in python310");
-    assert_script_run("python3.10 --version | grep Python | grep 3.10.");
+    my $python_version = get_var('PYTHON_VERSIONS', '3.11');
+    my $sub_version = substr($python_version, 2);
```
b10n1k, line 53: Not a big thing. But why not set `PYTHON_VERSIONS` directly to `311` for simplification?

tinawang123: Just follow the python version format to set this.

lemon-suse: I agree with Yiannis, set PYTHON_VERSIONS to 311 will make the code more simple, then we do not need so complcated code in line 54 & 55 to judge the python 311 package exist or not. WDYT?

tests/console/check_migration_features.pm
```diff
@@ -40,20 +40,25 @@ sub check_python2_dropped {
...
+    my @python31x_results = split "\n", script_output("zypper se python3$sub_version | awk -F \'|\' \'/python3$sub_version/ {gsub(\" \", \"\"); print \$2}\'");
+    if ((scalar(@python31x_results) == 0) || ($python31x_results[0] ne "python3$sub_version")) {
+        die("Python $python_version not found in the repositories\n\n");
+    }
+    zypper_call("install python3$sub_version");
```
lemon-suse, line 58: why not use 'in' instead of 'install'?

tests/console/check_migration_features.pm
```diff
@@ -40,20 +40,25 @@ sub check_python2_dropped {
...
+    assert_script_run("python3.$sub_version --version | grep Python | grep 3.$sub_version.");
```
lemon-suse, line 59: Suggest to update to 
python3.$sub_version --version | grep "Python"."3.$sub_version."

lemon-suse:
The PR title is too long, could we set it as 'Check Python3 module for 3.11 instead of 3.10' ? WDYT?

tests/console/check_migration_features.pm
tinawang123: Between 'Python' and version, it may have space. 
Use this will be more flexible.

tests/console/check_migration_features.pm
tinawang123: I prefer to keep use my method. It will be more readable.
For python version, we called it 2.6 or 3.10 but not 311. So I think my method is 'user friendly'.

tests/console/check_migration_features.pm
lemon-suse: You can try whether this works, on my local ENV this works. And this can check what you said together.

tests/console/check_migration_features.pm
jknphy: for me it is different three hundred eleven than three point eleven, it is never a good idea to set values artificially for the comfort of the developer, always better to be meaningful from outside.
But I agree this code could be better :), but it would need then a setting `PYTHON3_SUBVERSION` I guess. I will remember for future refactor that something smells here, thanks for the input.

jknphy: I agree that this code is overly complicated. the only difference is '10' or '11'. You most likely just need a condition to set the value of a variable `my $python3_subversion`.

tinawang123: Updated description.

jknphy: the description was a minor issue, the code is too complicated and at least 3 contributors thinks the same.
could you please address it?  Basically the suggestion is that there is not need for more code, and just pass '10' or '11'.

tinawang123: If only change 10 to 11, but python version changed to 3.22 or 3.12 later, we need change code again.
If use my method,  we can only change setting to update python version later.

tinawang123: I don't think the code too complicated. Before we only set 'zypper_call("se python310")'. Now I handled this error message and die the test module if the expected python version cannot get.

tests/console/check_migration_features.pm
```diff
@@ -40,20 +40,25 @@ sub check_python2_dropped {
...
     assert_script_run("SUSEConnect -p sle-module-python3/$OS_VERSION/$ARCH", timeout => 180);
-    zypper_call("se python310");
-    zypper_call("in python310");
-    assert_script_run("python3.10 --version | grep Python | grep 3.10.");
+    my $python_version = get_var('PYTHON_VERSIONS', '3.11');
+    my $sub_version = substr($python_version, 2);
+    my @python31x_results = split "\n", script_output("zypper se python3$sub_version | awk -F \'|\' \'/python3$sub_version/ {gsub(\" \", \"\"); print \$2}\'");
+    if ((scalar(@python31x_results) == 0) || ($python31x_results[0] ne "python3$sub_version")) {
+        die("Python $python_version not found in the repositories\n\n");
+    }
```
lemon-suse, line 57: Why not just zypper_call("se python3$sub_version"); ? 
We can have a variable to replace  the " python3$sub_version", such as my $python_pkg = "python3"$sub_version;

lemon-suse, line 57: Why not just zypper_call("se python3$sub_version"); ? 
We can have a variable to replace  the " python3$sub_version", such as my $python_pkg = "python3"$sub_version;

rakoenig:
LGTM.

lemon-suse (approved):

tests/console/check_migration_features.pm
lemon-suse: Following the conclusion of review meeting, use hardcode of python version to avoid complicated code.

PR #17070, opened 2023-05-15
[WIP] kernel/nfs: Add mutex to wait for barrier creation

richiejp:
Currently there is a race between support server and the NFS client/server. So add a mutex to wait for the barriers and support server to be setup.

No validation yet because it fails for another reason.

@schlad


richiejp:
Ah I see now there is a mutex in boot, but it wasn't enabled in my config.

PR #17071, opened 2023-05-15
Fix timeout while booting ppc64le

emiura:
ppc64le is taking more than 60 seconds to boot, therefore is not able to match bootloader screen.

Failure:
https://openqa.suse.de/tests/11131033

VR:
https://openqa.suse.de/tests/11131517

Fixes:
TEAM-7973 ppc64le is taking too long to show grub screen


bs-suse (approved):
LGTM

PR #17098, opened 2023-05-18
STIG remediation on autoyast installation

ilmanzo:
We introduce a new test that installs a system using autoyast, configured under all requirements (separate audit partition, encrypted disks, password on boot loader) and provide STIG remediation on first boot.
Testsuite is not yet scheduled, this PR is prerequisite for testing and scheduling on final openQA job groups.

- Related ticket: https://progress.opensuse.org/issues/122962
- Verification runs:
  - 15sp4 https://openqa.suse.de/tests/11149252
  - 15sp5 https://openqa.suse.de/tests/11149251


tests/security/grub_auth/grub_authentication.pm
```diff
@@ -0,0 +1,27 @@
...
+#
+# Maintainer: QE Security <none@suse.de>
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use base 'consoletest';
+
+my $user = 'root';
```
paolostivanin, line 14: you could use `$testapi::username` and `$testapi::password`

ilmanzo: seems standard test username is not allowed to unlock 'grub', we need root here. Updated also the test reliability adding a needle to ensure the correct screen is displayed

tjyrinki (approved):
Looks good. The logs seem similar to my manual installation, in other words lots of "failures" too which at some point (later) we may want to check what should be expected stig remediation results.

PR #17100, opened 2023-05-19
Not pipe out the information to pager for wicked service

tinawang123:
Not pipe out the information to pager for wicked service

- Verification run:  https://openqa.opensuse.org/tests/3301737#step/wicked/1


jknphy:
could you please always set the progress ticket(s) related? it is easier to navigate and review. thx

PR #17105, opened 2023-05-19
[qesap] Reuse infrastructure from previous test

lpalovsky:
This PR allows reuse infrastructure created by qesap-deployment project between 
multiple tests. Running test with parameter QESAP_DEPLOYMENT_EXPORT=1 disables 
cleanup and exports data from testrun. Next test run with 
QESAP_DEPLOYMENT_IMPORT=<test id> will connect to the existing infrastructure. 
This is meant mostly for development or keeping infrastructure for bug 
reporting.

- Related ticket: https://jira.suse.com/browse/TEAM-6902
- Needles: N/A
- Verification runs: 
Without export (cleanup should happen): https://openqaworker15.qa.suse.cz/tests/179826#
Export infra (Cleanup skipped): https://openqaworker15.qa.suse.cz/tests/179825#
Import infra (Cleanup skipped, hosts have same IPs): https://openqaworker15.qa.suse.cz/tests/179827#step/verify_infrastructure/33

### DIFF
--- a/data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
+++ b/data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
@@ -14,7 +14,7 @@ terraform:
     # GENERAL VARIABLES #
     az_region: '%PUBLIC_CLOUD_REGION%'
     admin_user: 'cloudadmin'
-    deployment_name: '%PUBLIC_CLOUD_RESOURCE_GROUP%'
+    deployment_name: '%QESAP_DEPLOYMENT_NAME%'
     public_key: '~/.ssh/id_rsa.pub'
     private_key: '~/.ssh/id_rsa'
     os_image: "%PUBLIC_CLOUD_IMAGE_ID%"
--- a/lib/qesapdeployment.pm
+++ b/lib/qesapdeployment.pm
@@ -37,6 +37,8 @@ use publiccloud::utils qw(get_credentials);
 use mmapi 'get_current_job_id';
 use testapi;
 use Exporter 'import';
+use Scalar::Util 'looks_like_number';
+use File::Basename;
 
 my @log_files = ();
 
@@ -74,6 +76,8 @@ our @EXPORT = qw(
   qesap_az_vnet_peering_delete
   qesap_add_server_to_hosts
   qesap_calculate_deployment_name
+  qesap_export_instances
+  qesap_import_instances
 );
 
 =head1 DESCRIPTION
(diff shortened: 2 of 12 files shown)


lib/sles4sap_publiccloud.pm
lpalovsky, line 1: Only moving "create_instance_data" from "tests/sles4sap/publiccloud/qesap_terraform.pm"

data/sles4sap/qe_sap_deployment/sles4sap_azure_generic.yaml
```diff
@@ -14,7 +14,7 @@ terraform:
     # GENERAL VARIABLES #
     az_region: '%PUBLIC_CLOUD_REGION%'
     admin_user: 'cloudadmin'
-    deployment_name: '%PUBLIC_CLOUD_RESOURCE_GROUP%'
+    deployment_name: '%QESAP_DEPLOYMENT_NAME%'
```
mpagot, line 17: Seems like I forget this one in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17079

Thanks

lib/qesapdeployment.pm
```diff
@@ -955,5 +959,55 @@ sub qesap_add_server_to_hosts {
         provider => $prov);
 }
 
+=head3 qesap_import_instances
+
+    Downloads assets required for re-using infrastructure from previously exported test.
+    qesap_import_instances(<$test_id>)
+
+=over 2
```
mpagot, line 967: `=over 1`

lib/qesapdeployment.pm
```diff
@@ -955,5 +959,55 @@ sub qesap_add_server_to_hosts {
...
+
+=item B<$test_id> - OpenQA test ID from a test previously run with "QESAP_DEPLOYMENT_IMPORT=1" and infrastructure still being up and running
+
+=back
+=cut
+
+sub qesap_import_instances {
+    my ($test_id) = @_;
+    die("OpenQA test ID must be a number. parameter 'QESAP_DEPLOYMENT_IMPORT' must contain ID of previously exported test")
```
mpagot, line 976: `Parameter` with capital `P`

lib/sles4sap_publiccloud.pm
```diff
@@ -453,4 +457,40 @@ sub cloud_file_content_replace() {
...
+
+sub create_instance_data {
+    my $provider = shift;
+    my $class = ref($provider);
+    die "Unexpected class type [$class]" unless ($class =~ /^publiccloud::(azure|ec2|gce)/);
+    my @instances = ();
+    my $inventory_file = qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER'));
+    my $ypp = YAML::PP->new;
+    my $raw_file = script_output("cat $inventory_file");
+    my $inventory_data = $ypp->load_string($raw_file)->{all}{children};
```
mpagot, line 475: Does not this block strictly depend on `qe-sap-deployment` and its Ansible inventory format?


```
my $inventory_file = qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER'));
    my $ypp = YAML::PP->new;
    my $raw_file = script_output("cat $inventory_file");
    my $inventory_data = $ypp->load_string($raw_file)->{all}{children};
```  

and maybe even more code below than this. What about to create a qesapdeployment.pm method to internally read the inventory and to return a perl data structure of some sort to be used by the double for loop below, without the need of that double loop code to know too much about the Ansible inventory format?

mpagot: I think that this code can reuse the same "data decoupling" concept, and qesapdeployment.pm method

https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/eb838e004ca8dadfec0bcd0d6baca884ada8ad3b#diff-d537e7676e4de934957a302dd9662d07237c473a02581e3fa2ca4065a6471c6dR129

lib/qesapdeployment.pm
lpalovsky: Thanks. Fixed.

lib/qesapdeployment.pm
lpalovsky: Thanks, fixed.

lib/sles4sap_publiccloud.pm
lpalovsky: You mean some function inside qesapdeplyoment.pm that will directly 'translate' inventory.yaml into perl structure?
Yeah it can be done. I would rather do it in another PR/Jira task not to delay this one. As you can see this function was only moved elsewhere and was already existing.

mpagot: Yes it is https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/eb838e004ca8dadfec0bcd0d6baca884ada8ad3b#diff-d537e7676e4de934957a302dd9662d07237c473a02581e3fa2ca4065a6471c6dR129 code moved in a lib file. 

I'm fine to eventually look for any possible improvement in a next different PR

mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM.

Just 1 comment and 1 question from my end.

Comment: I would change a bunch of double quotes to single quotes, as many of the strings between the double quotes do not contain variables or single quotes.

Question: regarding cleanup. Clean up of the cloud infra is not done by the original test as it will export it to be used in upcoming tests ... but since the terraform state is not shared, I don't think cleanup will work in subsequent tests either ... so intention is for the cleanup to be manual? Or something else?

lpalovsky:
@alvarocarvajald 

`Comment: I would change a bunch of double quotes to single quotes, as many of the strings between the double quotes do not contain variables or single quotes.`

Can you point me to some example?
I think the ones you mean were not escaping attempts but to have literal quote in for example 'record_info' output. 

`Question: regarding cleanup. Clean up of the cloud infra is not done by the original test as it will export it to be used in upcoming tests ... but since the terraform state is not shared, I don't think cleanup will work in subsequent tests either ... so intention is for the cleanup to be manual? Or something else?`

Yes, correct. The idea was mostly for non standard cases like development or giving access to developer to check in case of a bug report. And manually delete it afterwards.
I think it could be possible to recreate qesap deployment part but it might be more complicated. Especially the storing data between tests so they do not expose anything sensitive.

lpalovsky:
> 

Uuuuh... didn't know that about quotation difference. Going to read some perlmonks, fix it and will be back !
Thanks.

alvarocarvajald (approved):
LGTM

PR #17107, opened 2023-05-19
selinux: move ALP scheduling to yaml

paolostivanin:
- Related ticket: https://progress.opensuse.org/issues/126785
- Related pr: https://github.com/os-autoinst/opensuse-jobgroups/pull/330
- Verification runs:
  - bedrock: https://openqa.opensuse.org/tests/3301987
  - micro: https://openqa.opensuse.org/tests/3302004
  - tw: https://openqa.opensuse.org/tests/3302518
  - 15sp4: https://openqa.suse.de/tests/11164195

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -227,27 +227,6 @@ sub load_fips_tests {
     loadtest 'fips/mozilla_nss/nss_smoke' unless is_microos;
 }
 
-sub load_selinux_tests {
-    loadtest 'security/selinux/selinux_setup';
-    loadtest 'security/selinux/sestatus';
-    # ALP has selinux enabled and in enforcing mode by default
-    loadtest 'security/selinux/selinux_smoke' unless is_alp;
-    loadtest 'security/selinux/enforcing_mode_setup' unless is_alp;
-    loadtest 'security/selinux/semanage_fcontext';
-    loadtest 'security/selinux/semanage_boolean';
-    loadtest 'security/selinux/fixfiles';
-    loadtest 'security/selinux/print_se_context';
-    loadtest 'security/selinux/audit2allow';
-    loadtest 'security/selinux/semodule';
-    loadtest 'security/selinux/setsebool';
-    loadtest 'security/selinux/restorecon';
-    loadtest 'security/selinux/chcon';
-    loadtest 'security/selinux/chcat';
-    loadtest 'security/selinux/set_get_enforce';
-    loadtest 'security/selinux/selinuxexeccon';
-}
-
-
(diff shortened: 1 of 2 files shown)


jlausuch:
why?

paolostivanin:
it's more readable, maintainable and also would fix the issue where non related tests are scheduled within this test (e.g. year_2038_detection)

paolostivanin:
thanks Jose, that's indeed a valid point. We'll review this internally again, and then we'll decide how to move forward (yaml or perl).

tjyrinki (approved):
As discussed before, we can decide how we want to maintain these, but now Jose has explained his view in more detail. People in our team have historically prefered the YAML approach, so this can go in from my point of view unless wanting to backtrack based on the reasoning showed here.

Both approaches have their pros and cons. The main_micro_alp.pm is simpler to look at this point, but on the other it also includes main_common which we have tried to clean since it feels for us hard to follow. With newcomers coming on board (always), understanding the YAML route seems lower learning curve. But that is just personal point of view.

pdostal:
Hello Timo,

> As discussed before, we can decide how we want to maintain these, but now Jose has explained his view in more detail. People in our team have historically prefered the YAML approach, so this can go in from my point of view unless wanting to backtrack based on the reasoning showed here.

I was a big supporter of YAML schedule and I adopted it early. Then I probably gave some hard time to my teammates when we decided to switch to `lib/main_publiccloud.pm` but pretty please, consider this approach. We have huge job groups and we have hundreds of them. 

I of course agree that `lib/main_common.pm` is a mess, but having separate files for each product helps a lot. Perl is really easier to look work with then YAML in this case.

> Both approaches have their pros and cons. The main_micro_alp.pm is simpler to look at this point, but on the other it also includes main_common which we have tried to clean since it feels for us hard to follow. With newcomers coming on board (always), understanding the YAML route seems lower learning curve. But that is just personal point of view.

first of all, you have good point, the `lib/main_alp.pm` really includes `lib/main_common.pm` to my surprise. But it's in my opinion wrong. I suggest the following:
```diff
- use main_common;
+ use main_common qw(loadtest);
```

cc: @jlausuch

jlausuch:
> first of all, you have good point, the `lib/main_alp.pm` really includes `lib/main_common.pm` to my surprise. But it's in my opinion wrong. I suggest the following:
> 
> ```diff
> - use main_common;
> + use main_common qw(loadtest);
> ```
> 
> cc: @jlausuch

Yes, +1 to that change.

paolostivanin:
we discussed it internally, and we agreed to keep the scheduling the in pm file, at least for now.

Closing this.

PR #17108, opened 2023-05-19
Support server dDNS fix on Tumbleweed

richiejp:
Dynamic DNS seems to be broken on Tumbleweed because the  NAMED_CONF_INCLUDE_FILES variable was obsoleted in `/etc/sysconfig/named`

It's not clear if the below fix will break other versions. Also the NFS commit is only partially related; it seems that NetworkManager does not renew the DHCP release like wicked, this needs more investigation.

### DIFF
--- a/tests/support_server/setup.pm
+++ b/tests/support_server/setup.pm
@@ -143,13 +143,16 @@ sub setup_dns_server {
     $setup_script .= qq@
         sed -i -e '/^NETCONFIG_DNS_FORWARDER=/ s/=.*/="bind"/' \\
                -e '/^NETCONFIG_DNS_FORWARDER_FALLBACK=/ s/yes/no/' /etc/sysconfig/network/config
-        sed -i '/^NAMED_CONF_INCLUDE_FILES=/ s/=.*/="openqa.zones"/' /etc/sysconfig/named
         sed -i 's|#forwarders.*;|include "/etc/named.d/forwarders.conf";|' /etc/named.conf
         sed -i 's|#dnssec-validation .*;|dnssec-validation no;|' /etc/named.conf
+
+        echo -e '\ninclude "/etc/named.d/openqa.zones";' >> /etc/named.conf
         curl -f -v $named_url/openqa.zones > /etc/named.d/openqa.zones
+        chown :named /etc/named.d/openqa.zones
+
         curl -f -v $named_url/openqa.test.zone > /var/lib/named/master/openqa.test.zone
         curl -f -v $named_url/2.0.10.in-addr.arpa.zone > /var/lib/named/master/2.0.10.in-addr.arpa.zone
-        chown named:named /var/lib/named/master
+        chown -R named:named /var/lib/named/master
     @;
 
     # Allow RPZ overrides - poo#32290


pdostal (approved):

czerw:
Please make sure that it will work with old 12-sp3 support server image on osd https://openqa.suse.de/tests/11164297

tests/kernel/before_nfs_test.pm
```diff
@@ -11,13 +11,22 @@ use testapi;
...
     select_serial_terminal;
-    systemctl 'stop ' . $self->firewall;
+    systemctl '--now disable ' . $self->firewall;
     set_hostname(get_var("HOSTNAME", "susetest"));
+
+    power_action('reboot');
+    $self->wait_boot;
+    select_serial_terminal;
+
+    assert_script_run('ip addr');
```
b10n1k, line 27: it seems you print just info here. i suggest to just replace `assert_script_run` with `script_run`

richiejp:
Attempting verification with the linked HPC job: https://openqa.suse.de/tests/11176370

tests/kernel/before_nfs_test.pm
richiejp: I will change this in a future NFS PR, I have removed it for now.

richiejp:
Verification on o3 TW: https://openqa.opensuse.org/tests/3307520

This test doesn't actually use dDNS AFAICT, but the DNS role is enabled. Nothing on o3 appears to use dynamic DNS (it's broken after all).

richiejp:
Verifications passed

czerw (approved):

pevik:
> Please make sure that it will work with old 12-sp3 support server image on osd https://openqa.suse.de/tests/11164297

@czerw Would you please post a link what job should be checked? I haven't found supportserver in [12-SP3](https://openqa.suse.de/tests/overview?distri=sle&version=12-SP3&build=%3A29047%3Acronie&groupid=106), nor even in [12-SP5](https://openqa.suse.de/tests/overview?distri=sle&version=12-SP5&build=%3A29047%3Acronie&groupid=282).

richiejp:
> 
> @czerw Would you please post a link what job should be checked? I haven't found supportserver

See the HDD variable on the support_server job. It is using a static image.

b10n1k (approved):

czerw:
> > @czerw Would you please post a link what job should be checked? I haven't found supportserver
> 
> See the HDD variable on the support_server job. It is using a static image.

Yes, this config is same for few last years :)

alvarocarvajald:
This seems to have broken HA Migration tests with a support server running on 12-SP3. See: https://openqa.suse.de/tests/12931513#step/setup/37

In particular, this line https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/support_server/setup.pm#L149, in the case of migrations is adding `include "/etc/named.d/openqa.zones";` to `/etc/named.conf` while that file already has a `include "/etc/named.conf.include";` and `/etc/named.conf.include` it's also including `/etc/named.d/openqa.zones`. This is what causes named to fail to start in the HA migration tests.

I will try to come up with a solution for our failures, but if you have a suggestion it would be appreciated.

alvarocarvajald:
I'm submitting this as a fix: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/18236

Will tag you there for reviews.

PR #17116, opened 2023-05-22
Deal with no bounce back on powerKVM

lemon-suse:
We have workaround the situation when no bounce back happened on powerkvm, but it is not work when online_migration_setup loaded after bootloader module. Because when showed 'display manager' but test still used wait_boot -> boot_local_disk which will match from the point of bootloader. 

We need ensure if no bounce back happend it won't go back to grub2 for the test process, and go ahead, that handle_grub shouldn't be called anymore after dispaly-mananger selected in bootloader module.

- Related ticket: https://progress.opensuse.org/issues/129649
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/11175598#step/online_migration_setup/1
  VR to avoid regression: http://openqa.suse.de/tests/11181404

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -881,6 +881,9 @@ sub wait_boot {
         #
         $self->handle_pxeboot(bootloader_time => $bootloader_time, pxemenu => 'pxe-custom-kernel', pxeselect => 'pxe-custom-kernel-selected');
     }
+    # When no bounce back on power KVM, we need skip bootloader process and go ahead when 'displaymanager' matched.
+    elsif ((check_screen('displaymanager', 5)) && (get_var('OFW'))) {
+    }
     else {
         assert_screen([qw(virttest-pxe-menu qa-net-selection prague-pxe-menu pxe-menu)], 600) if (uses_qa_net_hardware() || get_var("PXEBOOT"));
         $self->handle_grub(bootloader_time => $bootloader_time, in_grub => $in_grub);


rakoenig:
LGTM

sofiasyria:
LGTM too

jknphy:
@lemon-suse I would expect tones of verification for this change and additional reviewers from other squad to approve it.
`sub wait_boot` is a function very shared in the repo. wdyt?

do you think it would worth to put some comment above the change explaining why this is needed? I hardly can understand why is the reason for the change reading the first comment (without screenshots I mean).

lemon-suse:
> @lemon-suse I would expect tones of verification for this change and additional reviewers from other squads to approve it. `sub wait_boot` is a function very shared in the repo. wdyt?
> 
> Do you think it would worth to put some comment above the change explaining why this is needed? I hardly can understand why is the reason for the change reading the first comment (without screenshots I mean).

NP, thanks for review, I have added comment.

okurz:
bonce?

lemon-suse:
> bonce?

It's a long story, I just continue to use this word to describe the thing there:

#######
sub boot_local_disk {
    if (get_var('OFW')) {
        # TODO use bootindex to properly boot from disk when first in boot order is cd-rom
        wait_screen_change { send_key 'ret' };
        # Currently the bootloader would bounce back to inst-bootmenu screen after pressing 'ret'
        # on 'local' menu-item, we have to check it and send 'ret' again to make booting properly

jknphy:
please, s/bonce/bounce/g

lemon-suse:
> please, s/bonce/bounce/g

Updated, thanks.

mloviska:
Hello, I think this has broken SLEM micro image testing. Currently the wait_boot code is not able to catch GRUB2 screen due to very short time out.

https://openqa.suse.de/tests/11285491#step/disk_boot/1

lemon-suse:
Reverted this PR.

PR #17118, opened 2023-05-22
Add PackageHub on HPC autoyast

b10n1k:
PackageHub is required to install warewulf on 15SP4. I edited aarch64 template as well even if warewulf is not scheduled there yet. (in fact is in the dev group)

- Verification run: https://openqa.suse.de/tests/11176838# + https://openqa.suse.de/tests/11176855#step/ww4/64 (although fails later and i havent checked why yet)


czerw (changes requested):
Why we need phub? Did you check it with Egbert?

czerw:
Not needed.

PR #17122, opened 2023-05-22
Add basic installation test for Elemental

ldevulder:
This PR adds a basic installation test for Elemental ISO.

Elemental is a very small OS based on SLEM for Rancher intended to be used with Rancher Manager to provision edge cluster. See [Elemental docs](https://elemental.docs.rancher.com) for more information.

This test will be used mainly by QAM to validate SLEM fixes pushed to Elemental image. More complete E2E tests are done in the [Elemental GitHub project](https://github.com/rancher/elemental).

- Related ticket: N/A
- Needles: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1641
- Add new product in OBS: https://gitlab.suse.de/openqa/openqa-trigger-from-ibs-plugin/-/merge_requests/128 and https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/868
- Job Group definition: https://github.com/rancher/elemental/pull/878

Verification runs:
- http://ldevulder.udp.ovpn1.nue.suse.de/tests/104

### DIFF
--- a/tests/elemental/installation.pm
+++ b/tests/elemental/installation.pm
@@ -1,0 +1,98 @@
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Test boot of Elemental OS
+# Maintainer: elemental@suse.de
+
+use base 'opensusebasetest';
+use strict;
+use warnings;
+
+use testapi;
+use power_action_utils qw(power_action);
+use serial_terminal qw(select_serial_terminal);
+use utils qw(file_content_replace);
+
+sub run {
+    my ($self) = @_;
+    my $cloudconfig = "/tmp/cloud-config.yaml";
+    my %boot_entries = (
+        active => {
+            tag => 'elemental-bootmenu-default'
+        },
+        passive => {
+            tag => 'elemental-bootmenu-fallback'
+        },
+        recovery => {
+            tag => 'elemental-bootmenu-recovery'
+        }
+    );
+
+
+    # Wait for GRUB
+    $self->wait_grub(bootloader_time => 120);
+    send_key('ret', wait_screen_change => 1);
+    wait_still_screen(timeout => 90);
+    save_screenshot();
+
+    ## No GUI, easier and quicker to use the serial console
+    select_serial_terminal();
+
+    # Add a simple cloud-config
+    my $rootpwd = script_output('openssl passwd -6 ' . get_var('TEST_PASSWORD'));
+    assert_script_run('curl ' . data_url('elemental/cloud-config.yaml') . ' -o ' . $cloudconfig);
+    file_content_replace($cloudconfig, '%TEST_PASSWORD%' => $rootpwd);
+
+    # Install Elemental OS on HDD
+    assert_script_run('elemental install /dev/vda --cloud-init ' . $cloudconfig);
+
+    # Reboot after installation
+    power_action('reboot', keepconsole => 1, textmode => 1);
+
+    # Use new root password
+    $testapi::password = get_var('TEST_PASSWORD');
+
+    # Loop on all entries to test them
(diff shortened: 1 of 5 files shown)


okurz:
https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/868 merged

tests/elemental/installation.pm
```diff
@@ -0,0 +1,94 @@
...
+    $self->wait_grub(bootloader_time => 120);
+    send_key('ret', wait_screen_change => 1);
+    wait_still_screen(timeout => 90);
+    save_screenshot();
+
+    ## No GUI, easier and quicker to use the serial console
+    select_serial_terminal();
+
+    # Add a simple cloud-config
+    my $rootpwd = script_output('openssl passwd -6 ' . get_var('PASSWORD'));
```
alvarocarvajald, line 41: No idea exactly how Elemental is supposed to work, so this may be a non-issue, but if by any chance this can leave infrastructure in the cloud you want to be protected, it may be better to use something like:
```
get_var('_SECRET_PASSWORD');
```
As it will not be published as part of the test results.

Usually what I do is `get_var('_SECRET_PASSWORD', get_var('PASSWORD'))` so any of the two can work.

alvarocarvajald (approved):
Code LGTM.

Couldn't see the verification run as it seems to be tied to your machine connected to the VPN perhaps?

I only have one comment regarding the **PASSWORD** setting which may be irrelevant.

tests/elemental/installation.pm
ldevulder: > Couldn't see the verification run as it seems to be tied to your machine connected to the VPN perhaps?

Yes, I will anyway run a new VR before merging to validate again.

As for the password it's only the root password generated live for the deployed OS, nothing secret. But maybe I can use a different variable name to be less generic.

ldevulder: I now remember why I used `PASSWORD` variable: it's because I have a default password set in the ISO for the first boot (set to `ros`) and to use this one for the `select_serial_console` function I need to set the value in `PASSWORD`.

It is defined in `set_defaults_for_username_and_password` function available in `lib/main_common.pm` to be able to use it.

So, as for the boot-after-installation test I need also to be able to connect I have to define a root password with `cloud-init` and so to simplify my test code I simply use the same `ros` value after installation. This is why `PASSWORD` is used.

ldevulder: But as this is not so good to use the same password I will push a modification to use a new one after OS installation.

ldevulder: Done!

tests/elemental/installation.pm
```diff
@@ -0,0 +1,97 @@
...
+    wait_still_screen(timeout => 90);
+    save_screenshot();
+
+    ## No GUI, easier and quicker to use the serial console
+    select_serial_terminal();
+
+    # Add a simple cloud-config
+    my $rootpwd = script_output('openssl passwd -6 ' . get_var('TEST_PASSWORD'));
+    assert_script_run('curl ' . data_url('elemental/cloud-config.yaml') . ' -o ' . $cloudconfig);
+    assert_script_run("sed -i 's|%TEST_PASSWORD%|$rootpwd|g' $cloudconfig");
```
alvarocarvajald, line 43: There's `file_content_replace()` in `lib/utils.pm` which does this.

alvarocarvajald (approved):
LGTM. Only a small suggestion.

tests/elemental/installation.pm
ldevulder: Yes I forgot this one... I already used it back in the day on HANA test :D

ricardobranco777 (approved):
LGTM!

PR #17128, opened 2023-05-23
Add libyui widget debug info

lemon-suse:
For libyui automate test, we need get the widget id firstly, add this libyui widget debug info to save time for manually get widget id.

- Related ticket: N/A
- Needles: N/A
- Verification run: http://openqa.suse.de/tests/11181589/logfile?filename=yui-log.txt

### DIFF
--- a/lib/YuiRestClient/Http/HttpClient.pm
+++ b/lib/YuiRestClient/Http/HttpClient.pm
@@ -6,6 +6,7 @@ use warnings;
 
 use Mojo::UserAgent;
 use YuiRestClient::Logger;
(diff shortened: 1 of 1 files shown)


jknphy:
we should properly analyze that and have a ticket to find a good solution, changing the framework is never easy task or quick PR, for example I don't know how this behaves when you have wait_until or a failure in the middle of performing installation, etc. I would recommend you to close this and file a ticket.

lemon-suse:
Closed the PR and file a ticket to track it https://progress.opensuse.org/issues/129709.

PR #17133, opened 2023-05-23
WIP wicked: enable valgrind for debugging

cfconrad:


lib/wickedbase.pm
```diff
@@ -117,6 +119,104 @@ sub assert_wicked_state {
...
+    if (index($val, ",")) {
+        @tmp = split(/,/, $val);
+    } else {
+        @tmp = $val;
+    }
+    foreach my $v (@tmp) {
+        $v =~ s/\.service$//;
+        if (grep { /^$v$/ } @all) {
+            push @enable, $v;
+        } elsif ($v =~ /^1|all$/i) {
```
asmorodskyi, line 144: what "/^1|" part means ?

lib/wickedbase.pm
```diff
@@ -117,6 +119,104 @@ sub assert_wicked_state {
...
+
+    my @services = $self->valgrind_get_services();
+    return 0 if (!@services);
+
+    record_info("valgrind enable", "services: @services\ncommand: $valgrind_cmd");
+
+    foreach my $service (@services) {
+        my $service_file = "/etc/systemd/system/$service.service";
+
+        assert_script_run("systemctl cat $service > $service_file");
```
asmorodskyi, line 178: I don't understand this line , isn't you output'ing  into service file what was already there ?

lib/wickedbase.pm
```diff
@@ -117,6 +119,104 @@ sub assert_wicked_state {
...
+        assert_script_run(sprintf(q(sed -i -E 's@^(ExecStart=)(.*)$@\1%s \2@' '%s'),
```
asmorodskyi, line 179: I think this line requires some comments explaining why you are sure that ExecStart will have exactly this content and maybe one example of expected content

lib/wickedbase.pm
```diff
@@ -117,6 +119,104 @@ sub assert_wicked_state {
...
+        assert_script_run("rm -f $logfile");
+    }
+}
+
+sub valgrind_postrun {
+    my $self = shift;
+
+    my @services = $self->valgrind_get_services();
+    foreach my $service (@services) {
+        my $logfile = "/var/log/valgrind_$service.log";
```
asmorodskyi, line 206: haven't found upload of this log files anywhere . I think we should do this

asmorodskyi:
code looks good but please add more **documentation** : 
1.  add variables which you created into variables.md 
2. add perldoc to functions which you have created

lib/wickedbase.pm
cfconrad: I didn't upload it, as I put it as `record_info()` if there is something relevant in it: http://openqa.wicked.suse.de/tests/136483#step/before_test/44

I think upload is to much here.

lib/wickedbase.pm
cfconrad: Exactly. But it is a different directory. This is the directory where "custom" changes can be made. The RPM put it in a system directory like `/lib/systemd/system/` and I made my changes in `/etc/systemd/system/`  , see: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Unit%20File%20Load%20Path

lib/wickedbase.pm
cfconrad: if the value is `1` or `all`, then valgind get's enabled for every wicked related binary.

lib/wickedbase.pm
cfconrad: What we do here, we add a prefix (the valgrind command) to the executable of the service. The only thing we expect is `ExecStart=` which must be there, see: https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart= 

More details:
 * We use `q()` as string quotes to allow `'` and special chars like "@" in it.
 * We build a more complex string with `sprintf("%s", $var)` to make the reading of the `sed` command more easy.
 * The `sed` call replace the line starting with `ExecStart=<command>` with `ExecStart=<valgrind-cmd> <command>`
   *  we use `@` as regex delemiter, to allow `/` 
   * `\1` and `\2` in the replace are the findings of group 1 and 2, where group 1 is always `ExecStart=`

Maybe as comment:
```
# Add valgrind command prefix to `ExecStart=` in service file
```

asmorodskyi: > Maybe as comment:

> # Add valgrind command prefix to `ExecStart=` in service file

yes please

lib/wickedbase.pm
asmorodskyi: up to you here , but from my personal experience very often in such cases when you have "something relevant" you want to see more context and get info what was happening before / after this relevant line  , so maybe we can upload full log when something relevant was found ? 

But don't want to insists so I am resolving this thread keeping this on your good will

asmorodskyi:
also you need to rebase your branch

lib/wickedbase.pm
cfconrad: But that is what you get in the `record_info()` it contain the full file not only the regex match.
So if we upload the file, we do not get more info, just from a other place...

lib/wickedbase.pm
cfconrad: done

lib/wickedbase.pm
asmorodskyi: ah ok

asmorodskyi (approved):
LGTM

PR #17135, opened 2023-05-23
Set host name when NetworkManager is active

richiejp:
Firstly restarting the network service doesn't seem to cause NetworkManager to renew the DHCP lease. If we use `nmcli` instead then it doesn't wait for the lease to be renewed. So we have to wait for DNS to come back or perhaps use ping.

I'm not sure whether we should just switch to Wicked or if these are bugs with the NetworkManager setup.

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1094,7 +1094,24 @@ sub set_hostname {
     assert_script_run "uname -n|grep $hostname";
     systemctl 'status network.service';
     save_screenshot;
-    assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+
+    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        assert_script_run 'nmcli -w 30 network off';
+        assert_script_run 'nmcli -w 30 network on';
+
+        for (my $i = 0; $i < 30; $i++) {
+            my $state = script_output 'nmcli -w 30 networking connectivity check';
+
+            last if $state =~ /full/;
+
+            sleep 1;
+        }
+    } else {
+        assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+    }
+
+    print_ip_info;
+    script_run("dig +short $hostname.openqa.test");
 }
 
 =head2 assert_and_click_until_screen_change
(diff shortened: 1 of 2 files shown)


lib/utils.pm
```diff
@@ -1078,7 +1097,22 @@ sub set_hostname {
...
     save_screenshot;
-    assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+
+    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        assert_script_run 'nmcli -w 30 network off';
+        assert_script_run 'nmcli -w 30 network on';
+
+        # Despite -w above, NM does not wait for the DHCP lease
+        if (get_var('USE_SUPPORT_SERVER', 0)) {
+            sleep 5;
```
Vogtinator, line 1107: You can do `nmcli net conn check` instead. That also waits for connectivity.

richiejp: Thanks that is much better. I have included a loop in case it does not wait for operations in progress. I don't see any indication in the man page it would wait, but it seems to.

richiejp:
Tumbleweed verification: https://openqa.opensuse.org/tests/3313949 (just checking for regressions)

richiejp:
SLES: https://openqa.suse.de/tests/11188700 (again just checking for regressions, this uses Wicked)

lib/utils.pm
```diff
@@ -1078,7 +1094,25 @@ sub set_hostname {
     assert_script_run "uname -n|grep $hostname";
     systemctl 'status network.service';
     save_screenshot;
-    assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+
+    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        # NetworkManager appears to do nothing when network.service is restarted on TW 20230514
```
czerw, line 1099: is it bug or feature?

richiejp: I have created a bug to check this: https://bugzilla.opensuse.org/show_bug.cgi?id=1211677

schlad (approved):

czerw (approved):

richiejp:
Verification looks OK. Once this is merged the NFS tests can be scheduled on TW.

https://github.com/os-autoinst/opensuse-jobgroups/pull/332

DimStar77:
As weird as it sounds: this change is responsible for https://openqa.opensuse.org/tests/3315793# (seen across multiple stagings)

A test run against a PR reverting this (on top of os-autoinst-distri-master/main) shows success: https://openqa.opensuse.org/tests/3315904

PR #17136, opened 2023-05-23
Set password in pre_run_hook for agama_base

rakoenig:
- basically the livecd settings clear the password (and that might change once we move to IBS) but once we are in agama we should match password with the system to linux, and when we finish without failure we should match password to running system password.


JRivrain:
> * basically the livecd settings clear the password (and that might change once we move to IBS) but once we are in agama we should match password with the system to linux, and when we finish without failure we should match password to running system password.

This is the reason why we decided not to use that variable! Now we need to set it again in post_run_hook...

PR #17145, opened 2023-05-24
Fix and expand SBD delay code

lpalovsky:
This PR fixes sbd delay calculation formula used by "on premise" tests and allows 
this function to be used by HANA SR as well. It adds as well ability to change 
the SBD_DELAY_START value for public cloud based tests.

- Related ticket: https://jira.suse.com/browse/TEAM-7914
- Needles: N/A
- Verification run: 

15SP5 default value on prem:
SBD 2nodes: https://openqa.suse.de/tests/11216985#step/check_after_reboot/32
Diskless SBD 3 nodes: https://openqa.suse.de/tests/11216991#step/check_after_reboot/30

Public cloud:
Timeout default:
15SP5: https://openqa.suse.de/tests/11217582#step/Crash_site_b-primary/22
15SP4: https://openqaworker15.qa.suse.cz/tests/184194#step/Crash_site_a-primary/22
15SP3: https://openqaworker15.qa.suse.cz/tests/184195#step/Crash_site_a-primary/22

Timeout set to "150s" or "120s":
15SP5: https://openqa.suse.de/tests/11217304#step/Crash_site_a-primary/41
15SP4: https://openqaworker15.qa.suse.cz/tests/184196#step/Crash_site_a-primary/24
15SP3: https://openqaworker15.qa.suse.cz/tests/184212#step/Crash_site_a-primary/24

Timeout "yes":
15SP5:  http://mordor.suse.cz/tests/6041#step/Crash_site_a-primary/25
15SP4: https://openqaworker15.qa.suse.cz/tests/184645#step/Crash_site_a-primary/25
15SP3: https://openqaworker15.qa.suse.cz/tests/184644#step/Crash_site_b-primary/24

Timeout "no":
** Failures are expected since SBD timeout is 0.
15SP5: https://openqa.suse.de/tests/11217337#step/Crash_site_a-primary/25
15SP4: https://openqaworker15.qa.suse.cz/tests/184200#step/Crash_site_a-primary/24
15SP3: https://openqaworker15.qa.suse.cz/tests/184201#step/Crash_site_a-primary/24

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -19,12 +19,19 @@ use lockapi;
 use isotovideo;
 use x11utils 'ensure_unlocked_desktop';
 use Utils::Logging 'export_logs';
+use Carp qw(croak);
+use Data::Dumper;
 
 our @EXPORT = qw(
   $crm_mon_cmd
   $softdog_timeout
   $join_timeout
   $default_timeout
+  $corosync_token
+  $corosync_consensus
+  $sbd_watchdog_timeout
+  $sbd_delay_start
+  $pcmk_delay_max
   exec_csync
   add_file_in_csync
   get_cluster_name
@@ -90,11 +97,17 @@ Extension (HA or HAE) tests.
 =back
 
 =cut
+
 our $crm_mon_cmd = 'crm_mon -R -r -n -N -1';
 our $softdog_timeout = bmwqemu::scale_timeout(60);
 our $prev_console;
 our $join_timeout = bmwqemu::scale_timeout(60);
 our $default_timeout = bmwqemu::scale_timeout(30);
+our $corosync_token = q@corosync-cmapctl | awk -F " = " '/runtime.config.totem.token\s/ {print $2/1000}'@;
+our $corosync_consensus = q@corosync-cmapctl | awk -F " = " '/runtime.config.totem.consensus\s/ {print $2/1000}'@;
+our $sbd_watchdog_timeout = q@grep -oP '(?<=^SBD_WATCHDOG_TIMEOUT=)[[:digit:]]+' /etc/sysconfig/sbd@;
+our $sbd_delay_start = q@grep -oP '(?<=^SBD_DELAY_START=)([[:digit:]]+|yes|no)+' /etc/sysconfig/sbd@;
+our $pcmk_delay_max = q@crm resource param stonith-sbd show pcmk_delay_max| sed 's/[^0-9]*//g'@;
 
 # Private functions
 sub _just_the_ip {
@@ -968,55 +981,73 @@ sub activate_ntp {
 
 =head2 calculate_sbd_start_delay
 
-  calculate_sbd_start_delay();
+  calculate_sbd_start_delay(\%sbd_parameters);
 
 Calculates start time delay after node is fenced.
 Prevents cluster failure if fenced node restarts too quickly.
 Delay time is used either if specified in sbd config variable "SBD_DELAY_START"
 or calculated:
-"corosync_token + corosync_consensus + SBD_WATCHDOG_TIMEOUT * 2"
+"corosync token timeout + consensus timeout + pcmk_delay_max + msgwait"
 Variables 'corosync_token' and 'corosync_consensus' are converted to seconds.
+For diskless SBD pcmk_delay_max is set to static 30s.
+
+%sbd_parameters = {
+    'corosync_token' => <runtime.config.totem.token>,
+    'corosync_consensus' => <runtime.config.totem.consensus>,
+    'sbd_watchdog_timeout' => <SBD_WATCHDOG_TIMEOUT>,
+    'sbd_delay_start' => <SBD_DELAY_START>,
+    'pcmk_delay_max' => <pcmk_delay_max>
+}
+
+If C<%sbd_parameters> argument is omitted, then function will
+try to obtain the values from the configuration files.
+
 =cut
 
 sub calculate_sbd_start_delay {
-    my %params;
-    my $default_wait = 35 * get_var('TIMEOUT_SCALE', 1);
-
-    %params = (
-        'corosync_token' => script_output("corosync-cmapctl | awk -F \" = \" '/config.totem.token\\s/ {print \$2}'"),
-        'corosync_consensus' => script_output("corosync-cmapctl | awk -F \" = \" '/totem.consensus\\s/ {print \$2}'"),
-        'sbd_watchdog_timeout' => script_output("awk -F \"=\" '/SBD_WATCHDOG_TIMEOUT/ {print \$2}' /etc/sysconfig/sbd"),
-        'sbd_delay_start' => script_output("awk -F \"=\" '/SBD_DELAY_START/ {print \$2}' /etc/sysconfig/sbd")
+    my ($sbd_parameters) = @_;
+    my %params = (ref($sbd_parameters) eq 'HASH') ? %$sbd_parameters : (
+        'corosync_token' => script_output($corosync_token),
+        'corosync_consensus' => script_output($corosync_consensus),
+        'sbd_watchdog_timeout' => script_output($sbd_watchdog_timeout),
+        'sbd_delay_start' => script_output($sbd_delay_start),
+        'pcmk_delay_max' => get_var('USE_DISKLESS_SBD') ? 30 :
+          script_output($pcmk_delay_max)
     );
 
+    my $default_wait = 35 * get_var('TIMEOUT_SCALE', 1);
+
+    record_info('SBD Params', Dumper(\%params));
+
     # if delay is false return 0sec wait
-    if ($params{'sbd_delay_start'} == 'no' || $params{'sbd_delay_start'} == 0) {
-        record_info("SBD start delay", "SBD delay disabled in config file");
+    if (grep /^$params{'sbd_delay_start'}$/, qw(no 0)) {
+        record_info('SBD start delay', 'SBD delay disabled either in /etc/sysconfig/sbd or by provided function arguments');
         return 0;
     }
 
     # if delay is only true, calculate according to default equation
-    if ($params{'sbd_delay_start'} == 'yes' || $params{'sbd_delay_start'} == 1) {
+    if (grep /^$params{'sbd_delay_start'}$/, qw(yes 1)) {
         for my $param_key (keys %params) {
-            if (!looks_like_number($params{$param_key})) {
-                record_soft_failure("SBD start delay",
-                    "Parameter '$param_key' returned non numeric value:\n$params{$param_key}");
-                return $default_wait;
-            }
-            my $sbd_delay_start_time =
-              $params{'corosync_token'} / 1000 +
-              $params{'corosync_consensus'} / 1000 +
-              $params{'sbd_watchdog_timeout'} * 2;
-            record_info("SBD start delay", "SBD delay calculated: $sbd_delay_start_time");
-            return ($sbd_delay_start_time);
+            croak("Parameter '$param_key' returned non numeric value: $params{$param_key}\n
+                This might indicate test issue or unexpected HA configuration value.")
+              if !looks_like_number($params{$param_key}) and $param_key ne 'sbd_delay_start';
         }
+        my $sbd_delay_start_time =
+          $params{'corosync_token'} +
+          $params{'corosync_consensus'} +
+          $params{'pcmk_delay_max'} +
+          $params{'sbd_watchdog_timeout'} * 2 +    # msgwait = sbd_watchdog_timeout * 2
+          30;    # wait time should be greater than formula therefore adding 30s
+        record_info('SBD start delay', "SBD delay calculated: $sbd_delay_start_time");
+        return ($sbd_delay_start_time);
     }
 
     # if sbd_delay_stat is specified by number explicitly
     if (looks_like_number($params{'sbd_delay_start'})) {
-        record_info("SBD start delay", "Specified explicitly in config: $params{'sbd_delay_start'}");
+        record_info('SBD start delay', "Specified explicitly in config: $params{'sbd_delay_start'}");
         return $params{'sbd_delay_start'};
     }
+    return $default_wait;
 }
 
 =head2 check_iscsi_failure
--- a/lib/sles4sap_publiccloud.pm
+++ b/lib/sles4sap_publiccloud.pm
@@ -16,13 +16,14 @@ use base 'publiccloud::basetest';
 use strict;
 use warnings FATAL => 'all';
 use Exporter 'import';
+use Scalar::Util 'looks_like_number';
 use publiccloud::utils;
 use publiccloud::provider;
 use testapi;
 use List::MoreUtils qw(uniq);
 use utils 'file_content_replace';
 use Carp qw(croak);
-use hacluster '$crm_mon_cmd';
+use hacluster;
 use qesapdeployment;
 use YAML::PP;
 use publiccloud::instance;
@@ -44,6 +45,7 @@ our @EXPORT = qw(
   wait_for_pacemaker
   cloud_file_content_replace
   setup_sbd_delay
+  sbd_delay_formula
   create_instance_data
 );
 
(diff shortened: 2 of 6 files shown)


lib/sles4sap_publiccloud.pm
```diff
@@ -434,16 +437,46 @@ sub wait_for_pacemaker {
...
+=head2 sbd_delay_formula
+    sbd_delay_formula();
+
+
+=cut
+
+sub sbd_delay_formula() {
+    my ($self) = @_;
+    my %params = (
+        'corosync_token' => $self->run_cmd(cmd => "corosync-cmapctl | awk -F \" = \" '/runtime.config.totem.token\\s/ {print \$2 /1000}'"),
```
mpagot, line 467: is this code portion similar to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17145/files#diff-f1dc32f4e8e57e2753021dd0a58d9cbd5cc8a10f9f919280ca8e75ab6d9ed9b1R986 ?

t/11_hacluster.t
```diff
@@ -0,0 +1,58 @@
+use strict;
+use warnings;
+use testapi;
```
mpagot, line 3: try not to include `testapi`

t/11_hacluster.t
```diff
@@ -0,0 +1,58 @@
...
+        '1' => 55,
+        'no' => 0,
+        '0' => 0,
+        '120' => 120,
+    );
+
+    my $hacluster = Test::MockModule->new('hacluster', no_auto => 1);
+    $hacluster->redefine(record_info => sub { note(join(' ', 'RECORD_INFO -->', @_)); });
+    $hacluster->redefine(record_soft_failure => sub { note(join(' ', 'RECORD_SOFT_FAILURE -->', @_)); });
+    $hacluster->redefine(script_output => sub { note(join(' ', 'SCRIPT_OUTPUT -->', @_)); });
```
mpagot, line 31: Why this `script_output` does not need to return a string?

lib/sles4sap_publiccloud.pm
lpalovsky: Yep. The difference is that PC uses "run_cmd" and on prem tests are doing "script_output". 
I guess I could bind that somehow together, but didn't wanna complicate things and let the function accept the values.

t/11_hacluster.t
lpalovsky: Yes. Thanks and fixed

t/11_hacluster.t
```diff
@@ -0,0 +1,58 @@
+use strict;
+use warnings;
+use testapi;
+use Test::More;
+use Test::Exception;
+use Test::MockModule;
+use hacluster;
+use Scalar::Util 'looks_like_number';
+
+my $sbd_delay;
```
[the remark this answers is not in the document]
lpalovsky: Fixed.

lib/hacluster.pm
```diff
@@ -974,49 +976,55 @@ Calculates start time delay after node is fenced.
...
-"corosync_token + corosync_consensus + SBD_WATCHDOG_TIMEOUT * 2"
+"corosync token timeout + consensus timeout + pcmk_delay_max + msgwait"
 Variables 'corosync_token' and 'corosync_consensus' are converted to seconds.
+For diskless SBD pcmk_delay_max is set to static 30s.
 =cut
 
 sub calculate_sbd_start_delay {
-    my %params;
+    my ($sbd_parameters) = @_;
+    my %params = $sbd_parameters ? %$sbd_parameters : (
```
mpagot, line 986: What is the idea of %params? What is the idea of all the `corosync-cmapctl` calls?
What I understand is that `calculate_sbd_start_delay` can work with `sbd_parameters` fully provided by the caller or by using values currently in use in the SUT. 
Is it right?

lib/hacluster.pm
```diff
@@ -974,49 +976,55 @@ Calculates start time delay after node is fenced.
 Prevents cluster failure if fenced node restarts too quickly.
 Delay time is used either if specified in sbd config variable "SBD_DELAY_START"
 or calculated:
-"corosync_token + corosync_consensus + SBD_WATCHDOG_TIMEOUT * 2"
+"corosync token timeout + consensus timeout + pcmk_delay_max + msgwait"
 Variables 'corosync_token' and 'corosync_consensus' are converted to seconds.
+For diskless SBD pcmk_delay_max is set to static 30s.
```
mpagot, line 981: Does it makes sense to describe the expected `sbd_parameters` structure here?

lib/sles4sap_publiccloud.pm
```diff
@@ -434,16 +437,46 @@ sub wait_for_pacemaker {
...
 =cut
 
 sub setup_sbd_delay() {
-    my ($self, $set_delay) = @_;
-    my $delay = $set_delay || "no";
-    record_info("SBD delay", "Setting SBD delay to: $delay");
+    my ($self) = @_;
+    my $delay = get_var('HA_SBD_START_DELAY');
+    $delay =~ s/[^ye]s//g;
+    if (get_var('HA_SBD_START_DELAY') eq '') { record_info('SBD delay', 'Skipping, parameter without value'); return; }
```
mpagot, line 448: I propose to split it in 3 lines

lib/sles4sap_publiccloud.pm
```diff
@@ -434,16 +437,46 @@ sub wait_for_pacemaker {
      "yes" - sets default SBD value which is calculated from a formula
      "<number of seconds>" - sets sepcific delay in seconds
```
mpagot, line 438: `sepcific` --> `specific`

t/12_sles4sap_publicccloud.t
```diff
@@ -0,0 +1,31 @@
...
+
+    for my $value (@failok_values) {
+        set_var('HA_SBD_START_DELAY', $value);
+        dies_ok { $self->setup_sbd_delay() } "Test expected failing 'HA_SBD_START_DELAY' value: $value";
+    }
+
+    for my $value (@passing_values) {
+        set_var('HA_SBD_START_DELAY', $value);
+        ok($self->setup_sbd_delay(), "Test 'HA_SBD_START_DELAY' passing values: $value");
+    }
```
mpagot, line 28: Please call `set_var('HA_SBD_START_DELAY', undef);` at the end

t/11_hacluster.t
lpalovsky: I think it should not be required here since unit test provides mock values through input variable, and the parts using 'script_output' are not executed.

lib/hacluster.pm
lpalovsky: Yes. The 'calculate_sbd_start_delay' was previously used by on prem tests only so it would get the required HA parameters directly from SUT using 'script_output'. The %params was the collection of gathered parameters. 
Since public cloud tests cannot use 'script_output' I created the option to provide those values by caller.

lib/sles4sap_publiccloud.pm
lpalovsky: done

lib/sles4sap_publiccloud.pm
lpalovsky: fixed

t/12_sles4sap_publicccloud.t
lpalovsky: Yes, good catch. Thanks.

lib/hacluster.pm
lpalovsky: Definitely. I just updated the code.

mpagot (approved):
LGTM

alvarocarvajald:
Looking at the VR before looking at the code, it would seem like in order to have tests passing in different versions of SLES in Azure, we require different SBD_DELAY_START settings.

I see 15-SP5 passed with keeping the default, 15-SP4 passed with 150s, and 15-SP3 passed with `yes` which calculated the delay to 66s.

Wonder if same behavior is observed in AWS/EC2.

t/11_hacluster.t
```diff
@@ -0,0 +1,57 @@
...
+
+    my $corosync_token_original = $sbd_delay_params{'corosync_token'};
+    $sbd_delay_params{'corosync_token'} = 'asdf';
+    $sbd_delay_params{'sbd_delay_start'} = 'yes';
+
+    dies_ok { calculate_sbd_start_delay(\%sbd_delay_params) } "Test should die with unexpected values";
+    $sbd_delay_params{'corosync_token'} = $corosync_token_original;
+};
+
+done_testing;
```
alvarocarvajald, line 57: Thanks for this! I think I will create tickets to extend unit testing for `lib/hacluster`

lib/hacluster.pm
```diff
@@ -974,49 +976,64 @@ Calculates start time delay after node is fenced.
...
-        'corosync_consensus' => script_output("corosync-cmapctl | awk -F \" = \" '/totem.consensus\\s/ {print \$2}'"),
-        'sbd_watchdog_timeout' => script_output("awk -F \"=\" '/SBD_WATCHDOG_TIMEOUT/ {print \$2}' /etc/sysconfig/sbd"),
-        'sbd_delay_start' => script_output("awk -F \"=\" '/SBD_DELAY_START/ {print \$2}' /etc/sysconfig/sbd")
-    );
+    record_info('SBD Params', Dumper(\%params));
 
     # if delay is false return 0sec wait
-    if ($params{'sbd_delay_start'} == 'no' || $params{'sbd_delay_start'} == 0) {
-        record_info("SBD start delay", "SBD delay disabled in config file");
+    if (grep /^$params{'sbd_delay_start'}$/, qw(no 0)) {
```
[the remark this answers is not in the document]
lpalovsky: Hmm... good point. I think perl code will poop itself since $params{'sbd_delay_start'} does not exist. 
I am thinking about putting the undefined value into the  'yes' branch and if the value is not provided, delay will be calculated from parameters. 
https://github.com/lpalovsky/os-autoinst-distri-opensuse/blob/7412a5ff3fa15bc5f07ff4aa577ddc533af0edb3/lib/hacluster.pm#L1015
I guess that would apply as well in case if sbd config file does not include the parameter as well. 
And this means I need to update unit test too.

lib/hacluster.pm
```diff
@@ -974,49 +976,64 @@ Calculates start time delay after node is fenced.
...
+    'pcmk_delay_max' => <pcmk_delay_max>
+}
+
 =cut
 
 sub calculate_sbd_start_delay {
-    my %params;
+    my ($sbd_parameters) = @_;
+    my %params = $sbd_parameters ? %$sbd_parameters : (
+        'corosync_token' => script_output("corosync-cmapctl | awk -F \" = \" '/runtime.config.totem.token\\s/ {print \$2\/1000}'"),
```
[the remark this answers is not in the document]
lpalovsky: That is some serious black magic...
I like it. :D

lpalovsky:
@alvarocarvajald 
`Looking at the VR before looking at the code, it would seem like in order to have tests passing in different versions of SLES in Azure, we require different SBD_DELAY_START settings.

I see 15-SP5 passed with keeping the default, 15-SP4 passed with 150s, and 15-SP3 passed with yes which calculated the delay to 66s.

Wonder if same behavior is observed in AWS/EC2.`

Hmm... that is  complicated... why wouldn't it be ? :)
Still need to check what is what but it appears that different releases have different values directly from isntallation:
15SP3 Azure - value is set to yes, which means it is calculated by formula, no need to touch from our side
15SP4 Azure - value is set to 216 as a static number - still a default value and no need to touch it from our end
15SP5 Azure - same as 216
15SP5 on prem - 71 static value

So we should start with those values and change them after test fails only with param HA_SBD_START_DELAY.
I think that so far all the predefined values were working.

lib/sles4sap_publiccloud.pm
```diff
@@ -432,18 +435,51 @@ sub wait_for_pacemaker {
...
+=head2 sbd_delay_formula
+    sbd_delay_formula();
+
+
+=cut
+
+sub sbd_delay_formula() {
+    my ($self) = @_;
+    my %params = (
+        'corosync_token' => $self->run_cmd(cmd => "corosync-cmapctl | awk -F \" = \" '/runtime.config.totem.token\\s/ {print \$2 /1000}'"),
```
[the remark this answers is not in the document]
lpalovsky: Yep. as you mentioned in other comment I added those command as global ones in hacluster.pm and exported them:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/537dcd65f0173dd087114d3b38dc88adbedcaeee/lib/hacluster.pm#L106

I am running two VRs now, one for PC and one for on prem. 
https://openqa.suse.de/tests/11237258#live
https://openqaworker15.qa.suse.cz/tests/185484#live

lib/sles4sap_publiccloud.pm
```diff
@@ -432,18 +435,51 @@ sub wait_for_pacemaker {
...
 
 =cut
 
 sub setup_sbd_delay() {
-    my ($self, $set_delay) = @_;
-    my $delay = $set_delay || "no";
-    record_info("SBD delay", "Setting SBD delay to: $delay");
+    my ($self) = @_;
+    my $delay = get_var('HA_SBD_START_DELAY');
+    $delay =~ s/[^ye]s//g;
```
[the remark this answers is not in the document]
lpalovsky: Dear diary, today I have learned what are "Lookaround assertions" :)

Fixed couple of things:
- fixed regex 
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1f3f1365d1475ebf0e1e2db41eb96be3d8f7fbe7/lib/sles4sap_publiccloud.pm#L447

-  setup_sbd_delay will return value that was changed
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1f3f1365d1475ebf0e1e2db41eb96be3d8f7fbe7/lib/sles4sap_publiccloud.pm#L458

-  unit test is now checking changed value as well to ensure regex is correct
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1f3f1365d1475ebf0e1e2db41eb96be3d8f7fbe7/t/12_sles4sap_publicccloud.t#L33

lib/sles4sap_publiccloud.pm
```diff
@@ -16,16 +16,18 @@ use base 'publiccloud::basetest';
...
 use testapi;
 use List::MoreUtils qw(uniq);
 use utils 'file_content_replace';
 use Carp qw(croak);
-use hacluster '$crm_mon_cmd';
+use hacluster qw(calculate_sbd_start_delay $crm_mon_cmd);
 use qesapdeployment;
 use YAML::PP;
 use publiccloud::instance;
+use Data::Dumper;
```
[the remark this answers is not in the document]
lpalovsky: Oh yes, you are correct. Probably leftover from debugging.

lib/sles4sap_publiccloud.pm
```diff
@@ -432,18 +435,51 @@ sub wait_for_pacemaker {
 
      "no" - do not set and turn off SBD delay time
      "yes" - sets default SBD value which is calculated from a formula
-     "<number of seconds>" - sets sepcific delay in seconds
```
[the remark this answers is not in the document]
lpalovsky: Fixed.

lib/sles4sap_publiccloud.pm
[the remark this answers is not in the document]
lpalovsky: Both tests finished. PC failed but because of the timeout.

lib/sles4sap_publiccloud.pm
```diff
@@ -432,18 +435,51 @@ sub wait_for_pacemaker {
...
+    sbd_delay_formula();
+
+
+=cut
+
+sub sbd_delay_formula() {
+    my ($self) = @_;
+    my %params = (
+        'corosync_token' => $self->run_cmd(cmd => "corosync-cmapctl | awk -F \" = \" '/runtime.config.totem.token\\s/ {print \$2 /1000}'"),
+        'corosync_consensus' => $self->run_cmd(cmd => "corosync-cmapctl | awk -F \" = \" '/runtime.config.totem.consensus\\s/ {print \$2 /1000}'"),
```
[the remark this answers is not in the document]
lpalovsky: Done

lib/sles4sap_publiccloud.pm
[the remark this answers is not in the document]
alvarocarvajald: Fantastic! Looks more clear, and now the regexp does what is supposed to. Thanks!

lib/sles4sap_publiccloud.pm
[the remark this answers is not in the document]
alvarocarvajald: > Both tests finished. PC failed but because of the timeout.

Seems on prem test failed with a syntax error: https://openqa.suse.de/tests/11237258#step/check_after_reboot/25

Haven't looked at the code yet, so don't know if you already fixed it. Just leaving the note here just in case.

lib/hacluster.pm
```diff
@@ -968,55 +981,70 @@ sub activate_ntp {
...
+For diskless SBD pcmk_delay_max is set to static 30s.
+
+%sbd_parameters = {
+    'corosync_token' => <runtime.config.totem.token>,
+    'corosync_consensus' => <runtime.config.totem.consensus>,
+    'sbd_watchdog_timeout' => <SBD_WATCHDOG_TIMEOUT>,
+    'sbd_delay_start' => <SBD_DELAY_START>,
+    'pcmk_delay_max' => <pcmk_delay_max>
+}
+
```
alvarocarvajald, line 1001: Suggestion to add:
```
If C<%sbd_parameters> argument is omitted, then function will
try to obtain the values from the configuration files.
```

lib/sles4sap_publiccloud.pm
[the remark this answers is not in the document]
lpalovsky: Ah sorry ... forgotten to update the VR link https://openqa.suse.de/tests/11242888
It's fixed now.

lib/sles4sap_publiccloud.pm
```diff
@@ -424,26 +426,61 @@ sub wait_for_pacemaker {
...
 =cut
 
 sub setup_sbd_delay() {
-    my ($self, $set_delay) = @_;
-    my $delay = $set_delay || "no";
-    record_info("SBD delay", "Setting SBD delay to: $delay");
+    my ($self) = @_;
+    my $delay = get_var('HA_SBD_START_DELAY');
+    $delay =~ s/(?<![ye])s//g;
+    if (get_var('HA_SBD_START_DELAY') eq '') {
```
alvarocarvajald, line 449: Why `get_var('HA_SBD_START_DELAY')` and not `$delay`?

t/11_hacluster.t
```diff
@@ -0,0 +1,57 @@
...
+        '120' => 120,
+    );
+
+    my $hacluster = Test::MockModule->new('hacluster', no_auto => 1);
+    $hacluster->redefine(record_info => sub { note(join(' ', 'RECORD_INFO -->', @_)); });
+    $hacluster->redefine(record_soft_failure => sub { note(join(' ', 'RECORD_SOFT_FAILURE -->', @_)); });
+    $hacluster->redefine(script_output => sub { note(join(' ', 'SCRIPT_OUTPUT -->', @_)); });
+
+    for my $input_value (keys %value_vs_expected) {
+        print "$input_value \n";
```
alvarocarvajald, line 33: I don't think this `print` is needed.

alvarocarvajald (approved):
LGTM.

I left a couple of suggestions: feel free to ignore the `print` one, but please add the documentation one. I also left a question, but that part of the code looks good as it is, so it was simply curiosity.

I suppose we can merge this on Monday.

t/11_hacluster.t
lpalovsky: Yes. Sorry... it was just a debug leftover. Removed.

lib/sles4sap_publiccloud.pm
lpalovsky: I am not sure what does get_var() return if the variable is not defined. I was afraid that it will return 0 instead of empty variable or undef and in that case function would change sbd value to literal 0. 
But now that I have unit tests I can check it in to time !

Will be back :D

lpalovsky: And it works. Added some more values to the unit test as well.

alvarocarvajald: `get_var()` takes 2 arguments, the name of the setting, and a default value. If no default value is provided, and the setting is not defined, then it will return `undef`.

PR #17150, opened 2023-05-25
Add licence popup check for confirm_installation

sofiasyria:
When there is a java update, we see a popup during confirm installation page display that breaks the test. This PR implements a conditional check that should handle the popup if it appears. If not, it just confirms and proceeds to installation.
The code is not yet tested for a java popup case, since we don't have the update available at this point.

- Related ticket: https://progress.opensuse.org/issues/127529
- Verification runs: 
15-SP5 : https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=sofiasyria%2Fos-autoinst-distri-opensuse%23java_popup
15-SP4  :  https://openqa.suse.de/tests/overview?distri=sle&version=15-SP4&build=sofiasyria%2Fos-autoinst-distri-opensuse%23java_popup


lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
...
+
+package Installation::ConfirmInstallation;
+use Installation::Popups::AcceptPopup;
+use Installation::Popups::OKPopup;
+use strict;
+use warnings;
+use testapi qw(save_screenshot record_info);
+use YuiRestClient;
+
+sub new {
```
jknphy, line 17: you can inherit from `lib/YaST/PageBase.pm` to avoid writting this again.

lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
...
+
+# Summary: Handles Confirm Installation Page.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Installation::ConfirmInstallation;
+use Installation::Popups::AcceptPopup;
+use Installation::Popups::OKPopup;
+use strict;
+use warnings;
+use testapi qw(save_screenshot record_info);
```
jknphy, line 14: We don't include testapi in the framework (unless there is not other way to interact with the screen, very specific corner cases), please remove it, you have to do this operations from test layer.

lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles Confirm Installation Page.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Installation::ConfirmInstallation;
```
jknphy, line 9: Either is a controller or is a page, so we need to name it, I guess in this case:
`Installation::ConfirmInstallationPage`.

lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
...
+sub new {
+    my ($class, $args) = @_;
+    my $self = bless {
+        app => $args->{app}
+    }, $class;
+    return $self->init();
+}
+
+sub init {
+    my $self = shift;
```
jknphy, line 26: this is very hard to maintain, you are doing POMs joining two popups in one page, in POM you get some part of the screen or all the screen, but not two screens that sometimes one happen and sometimes happens both...whoever needs to change this code would have a lot of doubts. It would be wiser to do the condition at the level of test layer (in other words, when I recommend to not use conditional logic in test layer, I mean that I even recommend less to use condition logic in the framework, it is like shooting yourself in the foot).

Confirm installation is just one thing, you are dealing with something that should be called `ConfirmAllBeforeInstallation`, which is quite blurry for this corner case, which is a popup that only happens in our automation in the way that we wrongly test the update, it is nto even a functional thingy, notice you need those screenshot in the middle here because of that, you can do the screenshot at the level of test.

In summary I would expect to see this conditional logic well visible in `tests/installation/confirm_installation.pm`

lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
...
+    $self->{popup_OK} = Installation::Popups::OKPopup->new({app => YuiRestClient::get_app()});
+    $self->{popup_accept} = Installation::Popups::AcceptPopup->new({app => YuiRestClient::get_app()});
+    $self->{rct_license} = $self->{app}->richtext({id => "lic"});
+    $self->{lbl_license} = $self->{app}->label({debug_label => "Confirm Package License: java-1_..."});
+    return $self;
+}
+
+sub accept_license_if_shown {
+    my ($self) = @_;
+    if ($self->{rct_license}->exist({timeout => 10})) {
```
jknphy, line 36: wait 10 seconds 99.999 of the time that a job in product or maintenance run, except for that single one time that we got this strange java popup 1 per year? I don't think it is a good idea...

I would prefer to click on confirm installation if it is found and if it is not, then trigger that logic to detect if there is java popup, and if there is confirm installation afterward, wdyt?

lib/Installation/ConfirmInstallation.pm
```diff
@@ -0,0 +1,55 @@
...
+        save_screenshot();
+        $self->{popup_accept}->press_accept();
+        my $info = get_label_text();
+        record_info("License information", $info);
```
jknphy, line 40: with a screenshot from test layer we probably don't need this (just guessing, because I think with don't have a way to see that at the moment right?)

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -201,6 +202,10 @@ sub get_dasd_disk_management {
     return Installation::DiskActivation::DASDDiskManagementController->new();
 }
 
+sub get_confirm_installation {
+    return Installation::ConfirmInstallation->new({app => YuiRestClient::get_app()});
```
jknphy, line 206: you don't need to pass the app if we get it in that base class as suggested in below comment.

lib/Installation/ConfirmInstallation.pm
sofiasyria: Done

lib/Installation/ConfirmInstallation.pm
sofiasyria: I am still using this kind of logic, only because the accept button, like the ok button, can be seen in other Pages etc. I could easily include the accept button in the LicencePopup, instead of having it in a separate Popup, but in case we find it elsewhere, we have to repeat the code. If you think it's uncommon and not worth the trouble, I'll implement it in LicencePopup

lib/Installation/ConfirmInstallation.pm
sofiasyria: Done

lib/Installation/ConfirmInstallation.pm
sofiasyria: Done

lib/Distribution/Opensuse/Tumbleweed.pm
sofiasyria: I have changed the design quite a bit and now I am  calling LicencePopup from the test layer, so I think now is necessary. Right?

tests/installation/confirm_installation_maintenance.pm
```diff
@@ -0,0 +1,29 @@
...
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    my $install_popup = $testapi::distri->get_ok_popup_controller();
+    my $license_popup = $testapi::distri->get_license_popup();
+    # First check is for the confirm installation popup in order to save time as
+    # license popup will rarely appear.
+    if ($install_popup->get_ok_popup()) {
```
jknphy, line 19: There is an statement `$install_popup->accept();
` written twice:
```
unless ($install_popup->get_ok_popup()) {
        save_screenshot();
        $license_popup->accept();
}
$install_popup->accept();

```

lib/Distribution/Opensuse/Tumbleweed.pm
```diff
@@ -201,6 +197,10 @@ sub get_dasd_disk_management {
     return Installation::DiskActivation::DASDDiskManagementController->new();
 }
 
+sub get_license_popup {
+    return Installation::Popups::LicensePopup->new({app => YuiRestClient::get_app()});
```
jknphy, line 201: it is not needed to be passed, the base class take care of having an `app` instance for th object.

lib/Installation/Popups/LicensePopup.pm
```diff
@@ -0,0 +1,35 @@
...
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles License Popup.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Installation::Popups::LicensePopup;
+use parent 'YaST::PageBase';
+use strict;
+use warnings;
+use Installation::Popups::AcceptPopup;
```
jknphy, line 13: I got lost here, I don't understand why there is a popup inside another pop ...
Both are different containers, they might have some part repeated, but in this case doesn't worth to do composition.

lib/Installation/Popups/LicensePopup.pm
```diff
@@ -0,0 +1,35 @@
...
+    return $self;
+}
+
+sub accept_license_popup {
+    my ($self) = @_;
+    $self->{popup_accept}->press_accept();
+    return $self;
+}
+
+sub get_license_popup {
```
jknphy, line 29: this getter method usually goes after the constructor and the init.

lib/Installation/Popups/LicensePopup.pm
```diff
@@ -0,0 +1,35 @@
...
+sub init {
+    my $self = shift;
+    $self->{popup_accept} = Installation::Popups::AcceptPopup->new({app => YuiRestClient::get_app()});
+    $self->{rct_license} = $self->{app}->richtext({id => "lic"});
+    return $self;
+}
+
+sub accept_license_popup {
+    my ($self) = @_;
+    $self->{popup_accept}->press_accept();
```
jknphy, line 25: now we don't use controller for this, so you need to call the getter before each call:
`$self->get_license_popup()->{popup_accept}->press_accept();` to know that the element is really there when you click it.
To distinguish methods that belong to the interface, (public accesible from outside) we can name the private one with underscore, like in this case `_get_license_popup`.

tests/installation/confirm_installation_maintenance.pm
jknphy, line 19: but this method `get_ok_pop` will die if doesn't find it, so you need to create a new one:
`is_ok_popup_visible()` which return 1 or 0.

tests/installation/confirm_installation_maintenance.pm
```diff
@@ -0,0 +1,29 @@
...
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    my $install_popup = $testapi::distri->get_ok_popup_controller();
+    my $license_popup = $testapi::distri->get_license_popup();
+    # First check is for the confirm installation popup in order to save time as
```
jknphy, line 17: we could include the example of the Java popup, if you have saved some screenshot, to be more specific, although the solution is generic, which I don't like much, because I would like that we have control of each license, but ok, for the sake of not fixing the test and not need to fix the automation, let's assume that any non-free license at the end introduced as addon should be accepted. (actually you could also say that, that this happens only when when we introduce a maintenance update that has non-free licence as an additional addon/product during the installation).

lib/Installation/Popups/LicensePopup.pm
sofiasyria: The idea was to create something like AbstractOKPopup.pm that could be used in other cases. But I assume that this is not found so often to worth a separate .pm

lib/Installation/Popups/AcceptPopup.pm
```diff
@@ -3,28 +3,32 @@
...
 use warnings;
-
-sub new {
-    my ($class, $args) = @_;
-    my $self = bless {
-        app => $args->{app}
-    }, $class;
-    return $self->init();
-}
+use YuiRestClient;
```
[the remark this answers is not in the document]
sofiasyria: I thought we need this for click() and exist() to work. But it works without it.

tests/installation/confirm_installation_maintenance.pm
sofiasyria: I can change the comments to make it clear that any license popup will be accepted here.

lib/Installation/Popups/LicensePopup.pm
sofiasyria: done

tests/installation/confirm_installation_maintenance.pm
```diff
@@ -0,0 +1,28 @@
...
+use base 'y2_installbase';
+
+sub run {
+    my $install_popup = $testapi::distri->get_ok_popup_controller();
+    my $license_popup = $testapi::distri->get_license_popup();
+    # The first check is for the confirm installation popup, in order to save time as
+    # extra license popups will rarely appear.
+    unless ($install_popup->is_ok_popup_visible()) {
+        save_screenshot();
+        $license_popup->accept();
```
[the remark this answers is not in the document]
jknphy: I would have opt for a code without conditions here...one method that returns a handler to the page that detects, and then call `accept()` with it. ie: Javascript implements `race` to get the fastest thing to appear in the screen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

```
# tumbleweed.pm
sub get_any_popup {
...
}

# confirm_installation_maintenance.pm
my $any_popup = $testapi::distri->get_any_popup();
my $popup = $any_popup.race([$install_popup, $license_popup]);
$popup->accept();

# AnyPopup.pm
sub race {
 # calls to is_shown for each page passed, method try/catched or without die and win the best.
}
```

But it is just an idea in pseudocode, you don't need to go that path, we can accept your existing code.

lib/Installation/Popups/LicensePopup.pm
```diff
@@ -0,0 +1,34 @@
...
+sub get_license_popup {
+    my ($self) = @_;
+    die "Licence popup is not shown" unless $self->{rct_license}->exist();
+    return $self;
+}
+
+sub accept_license_popup {
+    my ($self) = @_;
+    $self->get_license_popup()->{btn_accept}->click();
+    return $self;
```
jknphy, line 31: no need to return the $self if the object goes out of use/visibility, no chained call could be done.

lib/Installation/Popups/OKPopupController.pm
```diff
@@ -40,4 +40,9 @@ sub accept {
     $self->get_ok_popup()->press_ok();
 }
 
+sub is_ok_popup_visible {
+    my ($self) = @_;
+    return ($self->{OKPopup}->is_shown() ? 1 : 0);
```
jknphy, line 45: why to re-map to 1 or 0, `is_shown` always returns 1 or 0  using `exist` from `lib/YuiRestClient/Widget/Base.pm`, right?

jknphy (approved):
LGTM, just some ideas for future how we could enhance it and some minor things.

tests/installation/confirm_installation_maintenance.pm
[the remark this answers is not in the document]
sofiasyria: I have been checking this for a while, there is a race option in Mojo library : https://docs.mojolicious.org/Mojo/Promise which I could possibly use.
What I don't like is 
1) in case the java popup appears we need to accept both popups, in case it doesn't we accept only the second, so it's not either one or the other
2) I assume underneath the race functions there are quite a few if statements, so our code might look more elegant but on computing level doesn't seem like a big gain. 

Do we cover the expectations better? The current code in the PR expects to confirm installation, or if the license popup appears to fist accept it before the installation confirmation. In case of a race condition, we set the same expectations, no? So, I will not go forward with this now, but we can keep it in mind for the future.

tests/installation/confirm_installation_maintenance.pm
```diff
@@ -0,0 +1,28 @@
...
+#          that will appear before the confirm installation popup.
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+
+sub run {
+    my $install_popup = $testapi::distri->get_ok_popup_controller();
```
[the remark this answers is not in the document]
sofiasyria: this is in the pre-existing code, I would have to change all following:
`falafel@falafel:/var/lib/openqa/tests/os-autoinst-distri-opensuse> grep -r get_ok_popup_controller
lib/Distribution/Opensuse/Tumbleweed.pm:sub get_ok_popup_controller {
tests/installation/validate_beta_message.pm:    my $actual_beta_text = $testapi::distri->get_ok_popup_controller()->get_text();
tests/installation/access_beta_distribution.pm:    $testapi::distri->get_ok_popup_controller()->accept();
tests/installation/confirm_installation.pm:    my $install_popup = $testapi::distri->get_ok_popup_controller();
tests/installation/confirm_installation_maintenance.pm:    my $install_popup = $testapi::distri->get_ok_popup_controller();
`

tests/installation/confirm_installation_maintenance.pm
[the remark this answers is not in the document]
jknphy: oh, cool, nice finding! mmm that opens a door, but would require to change lower layer of our own framework.
1 is the logic of the application, it is how it is, there is not accept all or something like that. About 2, complexity is always there, there is no magic way to get rid of it, code designs just move this complexity from one place to another. Relying in a solid tested "race" method is always more maintainable than creating our own hacky if/else condition each time (and sometimes in a different way each time).
No problem to go forward with your current implementation,
Could you please file a ticket with that idea for Mojo promises?

tests/installation/confirm_installation_maintenance.pm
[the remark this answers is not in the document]
jknphy: yes, just if you want, it is not part of the ticket, just to clean up a bit.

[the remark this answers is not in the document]
sofiasyria: I'll make a separate pr because I would like to run more VRs for this

lib/Installation/Popups/OKPopupController.pm
sofiasyria: I had tried and it would always go inside the unless condition, but I will try again.

sofiasyria: seems like it works now for tests that the popup doesn't appear as well :+1:

PR #17153, opened 2023-05-25
Make TW test more stable and faster

Julie-CAO:
1. type key again to ensure to enter the particular ipxe entry.
2.  reduce timeout of grub2 menu of guest to prevent recieving keys from outside accidently
3.  decrease host installation time(ipxe_install + autoyast_install + login_console) from 30 min to 20 min by optimizing some needle matching
4. stop guest installation when dry run failed. (it will be droppend when the PR is ready to merge)
5. remove `<hiddenmenu>` attribute as it is not found in TW/SLE15SP4/SLE12SP5 autoyast guide.
6. deactivate the 2nd network interface to avoid networking issues
7. add udev rule to make br0 have same MAC with em1 to be able to get an IP from dhcp server

Related ticket: 
https://progress.opensuse.org/issues/130108
https://progress.opensuse.org/issues/130111

Verification run: 
// item 1 is not verified in this test because they have not been reproduced successully.
[virt-guest-installation-kvm](https://openqa.opensuse.org/tests/3330864)    //successful one. to verify change item 2, 3, 5
[dry run failed](https://openqa.opensuse.org/tests/3345200)  //to verify item 4
[dry fun failed](https://openqa.opensuse.org/tests/3345811#)  //to verify item4
[login successful on rebel:5](https://openqa.opensuse.org/tests/3352726)  //to verify item 6 & 7
[kvm test on rebel:5](https://openqa.opensuse.org/tests/3352909)  //PASS
[xen test on rebel:6](https://openqa.opensuse.org/tests/3352901)  //Fail as expected

### DIFF
--- a/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
@@ -68,6 +68,13 @@
   <networking>
     <keep_install_network t="boolean">true</keep_install_network>
     <backend>network_manager</backend>
+    <interfaces t="list">
+      <interface>
+        <bootproto>dhcp</bootproto>
+        <name>p3p1</name>
+        <startmode>manual</startmode>
+      </interface>
+    </interfaces>
   </networking>
   <software>
     <products t="list">
@@ -125,10 +132,10 @@
     <init-scripts t="list">
       <script>
         <source><![CDATA[
-sshd_config_file="/etc/ssh/sshd_config.d/permit_root_login.conf"
-echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > $sshd_config_file
-sshd_config_file="/etc/ssh/sshd_config.d/keep_alive.conf"
-echo -e "TCPKeepAlive yes\nClientAliveInterval 60\nClientAliveCountMax 480" > $sshd_config_file
+sshd_config_dir="/etc/ssh/sshd_config.d"
+echo -e "PermitRootLogin yes\nPubkeyAuthentication yes\nPasswordAuthentication yes" > ${sshd_config_dir}/permit_root_login.conf
+echo -e "TCPKeepAlive yes\nClientAliveInterval 60\nClientAliveCountMax 480" > ${sshd_config_dir}/keep_alive.conf
+echo -e "[Match]\nDriver=bridge\n\n[Link]\nMACAddressPolicy=none" > /etc/systemd/network/98-default-bridge.link
 ]]>
         </source>
       </script>
--- a/lib/concurrent_guest_installations.pm
+++ b/lib/concurrent_guest_installations.pm
@@ -112,6 +112,8 @@ sub install_guest_instances {
         else {
             $guest_instances{$_}->guest_installation_run(@_);
         }
+        # Abort current guest installation if dry run failed
+        next if $guest_instances{$_}->{guest_installation_result} eq 'FAILED';
         if ($guest_instances{$_}->has_noautoconsole_for_sure) {
             assert_screen('text-logged-in-root');
             $guest_instances{$_}->do_attach_guest_installation_screen_without_session;
--- a/tests/installation/ipxe_install.pm
+++ b/tests/installation/ipxe_install.pm
@@ -153,12 +153,10 @@ sub run {
         save_screenshot;
         ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
         assert_screen('o3-ipxe-menu', 300);
-        if (get_var('HOST_INSTALL_AUTOYAST')) {
-            is_kvm_host ? send_key 'k' : send_key 'z';
-        }
-        else {
-            send_key 't';
-        }
+        my $key = get_var('HOST_INSTALL_AUTOYAST') ? (is_kvm_host ? 'k' : 'z') : 't';
+        send_key "$key";
+        # try one more time as sometimes sending key does not take effect
+        send_key "$key" if check_screen('o3-ipxe-menu');
         assert_screen([qw(load-linux-kernel load-initrd)], 240);
         # Loading initrd spend much time(fg. 10-15 minutes to Beijing SUT)
         # Downloading from O3 became much more quick, some needles may not be caught.
@@ -170,12 +168,13 @@ sub run {
             record_info("TW install?", "Pls make sure it is Tumbleweed that is being installed.", result => 'softfail');
         }
         assert_screen([qw(network-config-created loading-installation-system)], 60);
+        return if get_var('AUTOYAST');
         wait_still_screen(stilltime => 12, similarity_level => 60, timeout => 30) unless check_screen('sshd-server-started', timeout => 60);
         save_screenshot;
     }
 
     # when we don't use autoyast, we need to also load the right test modules to perform the remote installation
-    if (get_var('AUTOYAST') and !get_var('VIRT_AUTOTEST')) {
+    if (get_var('AUTOYAST')) {
         # make sure to wait for a while befor changing the boot device again, in order to not change it too early
         sleep 120;
     }
--- a/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
+++ b/data/virt_autotest/guest_unattended_installation_files/opensuse_tumbleweed_kvm_hvm_guest_x86_64.xml
@@ -6,7 +6,6 @@
       <append>console=tty console=ttyS0,115200n8 mitigations=auto</append>
       <cpu_mitigations>auto</cpu_mitigations>
       <gfxmode>auto</gfxmode>
-      <hiddenmenu>false</hiddenmenu>
       <os_prober>false</os_prober>
       <secure_boot>false</secure_boot>
       <serial>serial --unit=0 --speed=115200 --parity=no --word=8</serial>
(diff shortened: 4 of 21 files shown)


lib/concurrent_guest_installations.pm
```diff
@@ -112,6 +112,8 @@ sub install_guest_instances {
         else {
             $guest_instances{$_}->guest_installation_run(@_);
         }
+        # abort current guest installation if dry run failed
+        return if $self->{guest_installation_result} eq 'FAILED';
```
varunkojha, line 116: Are we providing a clearer indication that the installation is being aborted due to a failure?

Julie-CAO:
@alice-suse @xguo @waynechen55  @varunkojha and others, Welcome review!

lib/concurrent_guest_installations.pm
alice-suse: To serve your purpose, which should be checked, `$guest_instances{$_}->{guest_installation_result}` or `$self->{guest_installation_result}`?

What's more important , when multiple vms are being installed, this change will stop further handling for the problematic vm and also LATTER vms. Is it intended too? If I remember correctly, one of the key behavior of concurrent guest installation is to let other vms' installation continue while some fail. That's what are preferred.

At last, suggest to do more careful&comprehensive verifications on this, especially different vm states for multiple vm installation case.

Julie-CAO: > Are we providing a clearer indication that the installation is being aborted due to a failure? 

For guest installation, we don't abort the test if some guests fail. we'd like to continue with other guests and consequent feature tests.

> After this will it throw an exception and terminate the execution or stop the test here?

No, I planed to skip the current guest installation and continue with other guests, but obviously the code is not correct at here. I will correct it.

Julie-CAO: > To serve your purpose, which should be checked, `$guest_instances{$_}->{guest_installation_result}` or `$self->{guest_installation_result}`?

I think you are correct. I need look into it closely.

> What's more important , when multiple vms are being installed, this change will stop further handling for the problematic vm and also LATTER vms. Is it intended too? If I remember correctly, one of the key behavior of concurrent guest installation is to let other vms' installation continue while some fail. That's what are preferred.

I planed to abort the installation for current guest only, but the codes are not correct at here. I am thinking of it should be 'next' instead of 'return'...

> At last, suggest to do more careful&comprehensive verifications on this, especially different vm states for multiple vm installation case.

I will put this change into next PR. Because tests can be run with one PR only, I usually put all changes in it. Especially this kind of code fix to failures which don't always reproduce. They can be merged until they got enough verification runs.

I will mark this commit as WIP and drop it when the PR is going be merged right now. And since the PR for xen test  has been merged, it is capable to verify it on xen tests.

tests/installation/ipxe_install.pm
```diff
@@ -153,12 +153,10 @@ sub run {
...
-        if (get_var('HOST_INSTALL_AUTOYAST')) {
-            is_kvm_host ? send_key 'k' : send_key 'z';
-        }
-        else {
-            send_key 't';
-        }
+        my $key = get_var('HOST_INSTALL_AUTOYAST') ? (is_kvm_host ? 'k' : 'z') : 't';
+        send_key "$key";
+        # try one more time as sometimes sending key does not take effect
+        send_key "$key" if check_screen('o3-ipxe-menu');
```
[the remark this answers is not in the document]
Julie-CAO: The failure happens casually. I have not gotten a chance to verify this change, so it is hard to say if 2 times retry is ok. This is just an attempt, and we can make further improvements based on the actual situation in the future.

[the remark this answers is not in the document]
varunkojha: That makes sense, Looks good to me

lib/concurrent_guest_installations.pm
```diff
@@ -112,6 +112,8 @@ sub install_guest_instances {
         else {
             $guest_instances{$_}->guest_installation_run(@_);
         }
+        # abort current guest installation if dry run failed
+        next if $guest_instances{$_}->{guest_installation_result} eq 'FAILED';
```
waynechen55, line 116: This is not needed here, because there is already the following lines before this line:
```
if ($guest_instances{$_}->{guest_installation_result} ne '') {
    next;
}
```

okurz (commented):
Please rephrase the git commit message to be more specific to the affected component as you are not generally making "Tumbleweed tests more stable". How about

"virtualization: Improve stability in Tumbleweed tests"

lib/concurrent_guest_installations.pm
Julie-CAO: the result of dry run if after its else statement:
```
112         else {
113             $guest_instances{$_}->guest_installation_run(@_);
114         }

```
fg. https://openqa.opensuse.org/tests/3339451#step/unified_guest_installation/440

As the dry run already fails, the test need not attach or detach guest's screen.

Julie-CAO:
> Please rephrase the git commit message to be more specific to the affected component as you are not generally making "Tumbleweed tests more stable". How about
> 
> "virtualization: Improve stability in Tumbleweed tests"

Ok, no problem. I'll add 'virtualization" to make it more specific.

tests/installation/ipxe_install.pm
```diff
@@ -170,12 +168,13 @@ sub run {
...
         }
         assert_screen([qw(network-config-created loading-installation-system)], 60);
+        return if get_var('AUTOYAST');
         wait_still_screen(stilltime => 12, similarity_level => 60, timeout => 30) unless check_screen('sshd-server-started', timeout => 60);
         save_screenshot;
     }
 
     # when we don't use autoyast, we need to also load the right test modules to perform the remote installation
-    if (get_var('AUTOYAST') and !get_var('VIRT_AUTOTEST')) {
+    if (get_var('AUTOYAST')) {
```
b10n1k, line 177: does this still make sense here? this condition wont never run due to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17153/files#diff-f8329f3474ec62c7e8b241c5be029319db687b91e770ec1f2327dc96ed9c0a97R171, right?

Julie-CAO: No, `return if get_var('AUTOYAST');` is in the condition block `if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {`.
So the PR does not change the logic of tests without `VIRT_AUTOTEST=1`

Julie-CAO:
one more commit to fix the http download problem by deactivating p3p1 and set udev rule to make bridge have a dependent MAC. more verify runs are updated in PR description field.

data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
```diff
@@ -68,6 +68,13 @@
   <networking>
     <keep_install_network t="boolean">true</keep_install_network>
     <backend>network_manager</backend>
+    <interfaces t="list">
+      <interface>
+        <bootproto>dhcp</bootproto>
+        <name>p3p1</name>
+        <startmode>manual</startmode>
```
alice-suse, line 75: Why manual start, not auto? If manual is selected, which part of code will start it? I didn't find it in the PR.

tests/installation/ipxe_install.pm
```diff
@@ -170,12 +168,13 @@ sub run {
...
         # make sure to wait for a while befor changing the boot device again, in order to not change it too early
         sleep 120;
     }
```
alice-suse, line 180: Just a reminder, with so many if else to differentiate the handling for tumbleweed, the readability of this file is decreasing. Maybe it is good to consider if another way is an option for future similar case, namely `inheriting from ipxe_install, then rewrite sub run`

alice-suse (commented):
I reviewed again, and left some minor comments. Others look good :+1:

data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
Julie-CAO: this interface is deactivated deleberately. It brings troubles. but it cannot be disabled entirely because it is needed by future sriov tests. The default interface is 'em1' which is used in our regular test.

tests/installation/ipxe_install.pm
Julie-CAO: This PR is just to stablize the TW tests, so to optimizing test structure is not considered, but it can be included in future test.

lib/concurrent_guest_installations.pm
```diff
@@ -112,6 +112,8 @@ sub install_guest_instances {
         else {
             $guest_instances{$_}->guest_installation_run(@_);
         }
+        # abort current guest installation if dry run failed
```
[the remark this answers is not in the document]
Julie-CAO: Yes

data/virt_autotest/host_unattended_installation_files/autoyast_opensuse_kvm_sshd.xml
alice-suse: I see, thanks for explanations. It is helpful to add a comment for this?

Julie-CAO: I am not sure if a comment is welcome in an autoyast profile so I describe it in commit msg and PR description.

alice-suse: I just checked the commit msg. It is good enough!  BTW, adding comment in autoyast is welcome too, format is <!--> , but not need to comment here again, since it is well mentioned.

Julie-CAO:
@alice-suse The PR is ready for merge. With the host autoyast change, we've gotten a few green tests https://openqa.opensuse.org/tests/3356774#next_previous

PR #17158, opened 2023-05-25
Revert "Set host name when NetworkManager is active"

DimStar77:
Testing a theory - this seems to be a candidate for a failure seen in Stagings

Reverts os-autoinst/os-autoinst-distri-opensuse#17135

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -77,7 +77,6 @@ our @EXPORT = qw(
   exec_and_insert_password
   shorten_url
   reconnect_mgmt_console
-  print_ip_info
   set_hostname
   show_tasks_in_blocked_state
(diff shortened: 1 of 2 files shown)


DimStar77:
To all our surprised, the theory is confirmed by openQA: the change is indeed responsible for the failure seen in the KDE tests.

The running theory is that the running kde session loses its network setup and gets a new one, which in turn triggers an internal session reset

richiejp:
I don't think there will be any harm in reverting as we have not scheduled anything yet, thanks!

richiejp:
Specifically it is caused when `nmcli network off` is executed. Setting it on does not cause it which is the opposite of what I would expect.

PR #17162, opened 2023-05-26
Make sure that apparmor is disabled on HPC

b10n1k:
As we are shifting towards selinux HPC team suggested that we do not need to be testing against apparmor, which there are not expectations for it to work without issues. This commit ensure that Apparmor is turned off, however we still testing installation where Apparmor comes as dependency of the diveded system roles

- Related ticket: https://progress.opensuse.org/issues/127670
- Verification run: 
https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2317162
https://openqa.suse.de/tests/overview?distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2317162&version=15-SP3


czerw (approved):

Avinesh (approved):
looks good!

PR #17163, opened 2023-05-26
Containers: use httpd image instead of fedora/apache

jlausuch:
- Related ticket: https://progress.opensuse.org/issues/129910
VRs:
- https://openqa.suse.de/tests/11199934
- https://openqa.suse.de/tests/11200094
- https://openqa.suse.de/tests/11200095

### DIFF
--- a/tests/containers/podman_netavark.pm
+++ b/tests/containers/podman_netavark.pm
@@ -75,7 +75,7 @@ sub run {
         subnet_v6 => 'fd00::1:8:0/112'
     };
     my $ctr1 = {
-        image => get_required_var('REGISTRY') . '/fedora/apache',
+        image => registry_url('httpd'),
         name => 'apache_ctr',
         ip => '10.90.0.8',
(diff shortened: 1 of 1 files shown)


ricardobranco777 (commented):
Why use Docker Hub?  Do we have a subscription to avoid the rate-limits?  Otherwise we may have failing tests because of this.

We can use our own registries (registry.suse.de or registry.opensuse.org) to fetch images for all architectures.

jlausuch:
> Why use Docker Hub? Do we have a subscription to avoid the rate-limits? Otherwise we may have failing tests because of this.
> 
> We can use our own registries (registry.suse.de or registry.opensuse.org) to fetch images for all architectures.

I don't understand your question. We are not using docker hub anywhere :)

ricardobranco777:
> 

OK.  I thought the removal of the REGISTRY variable would default to Docker Hub.

PR #17147, opened 2023-05-24
zypper_lifecycle: Skip EOS python2 module

dzedro:
- Related ticket: https://progress.opensuse.org/issues/129026
- Verification run:
https://openqa.suse.de/tests/11190286 x86_64
https://openqa.suse.de/tests/11190287 s390x
https://openqa.suse.de/tests/11190288 aarch64

PR #17164, opened 2023-05-26
Drop python2 module support on 15sp1-ltss

skliu:
quick PR

- Needles: none
- Verification run: https://openqa.suse.de/tests/11200130

### DIFF
--- a/tests/console/zypper_lifecycle.pm
+++ b/tests/console/zypper_lifecycle.pm
@@ -37,7 +37,7 @@ our $date_re = qr/[0-9]{4}-[0-9]{2}-[0-9]{2}/;
 
 sub lifecycle_output_check {
     my $output = shift;
-    if (is_sle('=15-sp1') && $output =~ /Python 2 Module.*2021-07-31/) {
+    if (is_sle('=15-sp1') && $output =~ /Python 2 Module.*2021-07-3(0|1)/) {
         record_info 'poo#129026';
         return;
     }


jknphy:
Please fix it in this way, relaxing condition for example:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17147/files
`if (is_sle('=15-sp1') && $output =~ /Python 2 Module.*2021-07-3(0|1)/) {`

skliu:
> Please fix it in this way, relaxing condition for example: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17147/files `if (is_sle('=15-sp1') && $output =~ /Python 2 Module.*2021-07-3(0|1)/) {`

Done.

tests/console/zypper_lifecycle.pm
```diff
@@ -37,7 +37,8 @@ our $date_re = qr/[0-9]{4}-[0-9]{2}-[0-9]{2}/;
 
 sub lifecycle_output_check {
     my $output = shift;
-    if (is_sle('=15-sp1') && $output =~ /Python 2 Module.*2021-07-31/) {
+
```
jknphy, line 40: please remove extra line.
We new code we can separate logically parts, but for existing code shared I don't recommend it, just change the minimum number of lines and characters :)

skliu: Done.

dzedro:
I don't know to what date it can change, maybe expect any digit ?

skliu:
> I don't know to what date it can change, maybe expect any digit ?

In fact, I think we can just match `Python 2 Module` and ignore all the date digit since we drop supporting python2 on sle15sp1.

dzedro:
> > I don't know to what date it can change, maybe expect any digit ?
> 
> In fact, I think we can just match `Python 2 Module` and ignore all the date digit since we drop supporting python2 on sle15sp1.

Yes, I was thinking to ignore this for LTSS, at some point this will happen on each product which get LTSS, IMO.

jknphy:
LGTM, we can relax further the condition, the VR seems to take too much time.

dzedro (approved):

lemon-suse (approved):

PR #17170, opened 2023-05-29
set_hostname: Renew DHCP lease when using NetworkManager

richiejp:
Second attempt at updating the DHCP lease to cause a dDNS update.

Previous attempt was reverted because it caused KDE Wallet to pop-up:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17135

It appears that setting the device down and up again renews the DHCP lease, but does not cause a KDE wallet pop-up. Another option would be to setup the KDE wallet, but it seems likely there will be more issues on other desktops. A yet more direct approach is to trigger a DHCP request without touching the device status. However I have not found a way to do this on our setup.

Waiting for verification: https://openqa.opensuse.org/tests/3321302#


richiejp:
The above is not the right job for verification, this is what failed before: https://openqa.opensuse.org/tests/3321382

richiejp:
Verification passed.

@czerw @DimStar77 @Vogtinator

lib/utils.pm
```diff
@@ -1078,7 +1093,26 @@ sub set_hostname {
...
-    assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+
+    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        my $conn_id = script_output 'basename $(ls -1 /etc/NetworkManager/system-connections | head -n 1) .nmconnection';
+
+        assert_script_run 'nmcli device down ' . $conn_id;
+        assert_script_run 'nmcli device up ' . $conn_id;
+
+        for (my $i = 0; $i < 30; $i++) {
+            my $state = script_output 'nmcli -w 30 networking connectivity check';
```
b10n1k, line 1104: i think the delay here is a bit overestimated. which would cause unnecessary slowness. Likely this will take 1 or 2 iterations.  if we have to iterate 30 times over it, would this make sense to set nmcli to 5 or 10?

richiejp: So far it has always taken one iteration. I guess if it has not connected after a few seconds then it is probably not going to connect.

richiejp:
New verification run: https://openqa.opensuse.org/tests/3323769

richiejp:
Verification is passing

czerw:
@DimStar77   we would like to merge this one today.  If you see anything wrong like with previous PR, just let us know.

DimStar77:
> @DimStar77 we would like to merge this one today. If you see anything wrong like with previous PR, just let us know.

out of curiosity: what's the difference from the previous one causing issues? They look the same to me.


In any case, I'm okay with merging, but please keep in mind that we're just before the weekend, which means if it fails, there is basically no one to reach out to and I'd have to revert (again) - but seeing the VR having passed, this should be all right

richiejp:
> if I see the diff right, last time we tried network off/on, this time device down/up, right?

Yeah, we just restart the device and this does not appear to trigger whatever dbus thing was getting kwallet excited. We could leave it until Monday I guess @czerw ?

czerw:
> > if I see the diff right, last time we tried network off/on, this time device down/up, right?
> 
> Yeah, we just restart the device and this does not appear to trigger whatever dbus thing was getting kwallet excited. We could leave it until Monday I guess @czerw ?

Sure, no stress.

DimStar77:
😢 https://openqa.opensuse.org/tests/3338322#step/hostname/24

rfan1:
Seems test failed on SLE as well:
https://openqa.suse.de/tests/11260461#step/hostname/26

lib/utils.pm
```diff
@@ -1078,7 +1093,26 @@ sub set_hostname {
     assert_script_run "uname -n|grep $hostname";
     systemctl 'status network.service';
     save_screenshot;
-    assert_script_run "if systemctl -q is-active network.service; then systemctl reload-or-restart network.service; fi";
+
+    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        my $conn_id = script_output 'basename $(ls -1 /etc/NetworkManager/system-connections | head -n 1) .nmconnection';
```
DimStar77, line 1098: maybe rather use something like `nmcli device show | grep GENERAL.DEVICE | grep -v "\Wlo$"` ?

richiejp: I had already written something slightly different in #17221. I guess the main thing is they both use `nmcli` to get the device list

PR #17175, opened 2023-05-30
Automate auto-installation which adds user with password

coolgw:
Automate auto-installation which adds user with password
- Related ticket: https://progress.opensuse.org/issues/129739
- Needles: na
- VR: 

TW: 
https://openqa.opensuse.org/t3330457
micro:
https://openqa.opensuse.org/t3330458
bedrock:
https://openqa.opensuse.org/t3330459


tests/yam/validate/validate_user.pm
```diff
@@ -0,0 +1,33 @@
...
+
+sub run {
+    my $user = "test";
+    select_serial_terminal();
+    assert_script_run "useradd -m $user";
+    assert_script_run "echo '$user:$testapi::password' | chpasswd";
+    type_string("su - $user", lf => 1);
+
+    # Validate new alp user with password
+    $user = "bernhard";
```
rfan1, line 23: how about `$testapi::user`? :)

data/yam/agama/auto/alp_add_user.sh
```diff
@@ -0,0 +1,8 @@
+set -ex
```
jknphy, line 1: I noticed naming under `data/yam/agama/auto/` is wrong.
`alp_tw.json` -> `default_tumbleweed.json`
`alp_tw.sh` -> `default_tumbleweed.sh`
`alp.sh` -> `alp_bedrock.sh`
Could you apply that change?

For this one should be `/data/yam/agama/auto/tumbleweed_add_user.sh`
In other words, alp shouldn't be mention at all because it is not connected in any way.

tests/yam/validate/validate_user.pm
```diff
@@ -0,0 +1,33 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Validate alp user with password
```
jknphy, line 6: it is just an user installed by agama, you can drop "alp" as well here.

schedule/yam/agama_auto_add_user.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
jknphy, line 1: we can reuse the yamls created for agama_auto from previous task, adding user is a default.

tests/yam/validate/validate_user.pm
```diff
@@ -0,0 +1,33 @@
...
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base "consoletest";
+use strict;
+use warnings;
+use testapi;
+use serial_terminal qw(select_serial_terminal);
+
+sub run {
+    my $user = "test";
```
jknphy, line 16: user should be added properly by agama, we could just login to check that the password is fine.

coolgw: The user is root after we call select_serial_terminal(), if you switch root to other user, no need input any password.
That's the reason for current logic, root-> test->bernhard.

tests/yam/validate/validate_user.pm
coolgw: Thanks for comments, current code will be recontructed later.

tests/yam/agama/auto.pm
```diff
@@ -25,6 +25,8 @@ sub run {
 
     my @tags = ("welcome-to", "login");
     assert_screen \@tags, 960;
+
+    reset_consoles;
```
jknphy, line 29: this should belong to the `post_run_hook`

data/yam/agama/auto/alp_bedrock.sh
```diff
@@ -1,7 +1,7 @@
 set -ex
```
jknphy, line 1: might help to call them: `default_alp_bedrock` and `default_alp_micro` as you did for tw, to distinguish them from further work.

jknphy:
LGTM other than those minor things.

PR #17179, opened 2023-05-30
Integrate OpenBLAS in mpi module

b10n1k:
Those are some changes which take place to run test against OpenBLAS library. The source code doesnt use openMP but it is a smoke test of the library. Likely it could run in a single machine test. Added also some record_info to increase visibility on the UI.

- Related ticket: https://progress.opensuse.org/issues/120801


Avinesh (approved):
LGTM

PR #17182, opened 2023-05-30
Add new disk encryption tests

realcharmer:
New security tests for encrypted installation testing with separate `/boot` partition and running in FIPS mode. So far it doesn't cover testing on `ppc64le`.

The tests are derived from YaST team and patched to run in FIPS mode, since some platforms require special steps in such case. This was done after consulting with the YaST team.

- Related tickets:
    - https://progress.opensuse.org/issues/113162
    - https://progress.opensuse.org/issues/116281
- Verification runs:
    - s390x: https://openqa.suse.de/tests/11211502
    - x86_64: https://openqa.suse.de/tests/11217049
    - aarch64: https://openqa.suse.de/tests/11217050

### DIFF
--- a/schedule/security/encryption/lvm_encrypt_separate_boot.yaml
+++ b/schedule/security/encryption/lvm_encrypt_separate_boot.yaml
@@ -1,0 +1,39 @@
+---
+name: lvm-encrypt-separate-boot
+description: >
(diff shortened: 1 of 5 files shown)


ilmanzo:
looks good! Let's wait for the VR to finish and then we can merge

PR #17194, opened 2023-05-31
Public Cloud: Fix Perl syntax in azure image handling code

pdostal:
- Related: #16886

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -74,7 +74,7 @@ sub find_img {
     my $key = $self->get_storage_account_keys($storage_account);
 
     $json = script_output("az storage blob show --account-key $key -o json " .
-          "--container-name '$self->container' --account-name '$storage_account' --name '$name' " .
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM

lib/publiccloud/azure.pm
```diff
@@ -176,12 +176,12 @@ sub upload_img {
 
     # Note: VM images need to be a page blob type
     assert_script_run('az storage blob upload --max-connections 4 --type page'
-          . " --account-name '$storage_account' --account-key '$key' --container-name '$self->container'"
+          . " --account-name '$storage_account' --account-key '$key' --container-name '${self->container}'"
           . " --file '$file' --name '$img_name' --tags '$tags'", timeout => 60 * 60 * 2);
```
m-dati, line 180: Kind comment: could `single quote` here prevent the variable expansion?

m-dati: I meant, external double quote let expand, but internal single quote doesn't: 
https://www.geeksforgeeks.org/perl-quoted-interpolated-and-escaped-strings/

PR #17197, opened 2023-05-31
Public Cloud: Add -overwrite to azure::upload_img

pdostal:
- Related ticket: #16886

When was the blob uploaded incorrectly, we need to overwrite it:
```
ErrorCode:BlobAlreadyExists
If you want to overwrite the existing one, please add --overwrite in your command.[0m
```


asmorodskyi (approved):
LGTM

asmorodskyi (approved):
LGTM

PR #17201, opened 2023-06-01
test_running: Fail test if no openQA jobs are running

kalikiana:
Note the echo preventing $ret from ever being effective.

See also https://progress.opensuse.org/issues/58007#note-4

See: https://progress.opensuse.org/issues/130201

### DIFF
--- a/tests/openqa/osautoinst/test_running.pm
+++ b/tests/openqa/osautoinst/test_running.pm
@@ -10,9 +10,11 @@ use strict;
 use warnings;
 use base "consoletest";
 use testapi;
+use utils 'zypper_call';
 
 sub run {
-    assert_script_run ' ret=false; for i in {1..5} ; do openqa-client jobs state=running | grep --color -z running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ] ; echo $? ', 300;
+    zypper_call('in ack');
+    assert_script_run q{ret=false; for i in {1..5} ; do openqa-cli api jobs state=running state=done | ack --passthru --color 'running|done' && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]}, 300;
 }
 
 sub post_fail_hook {


Martchus (changes requested):
But what if the test has already finished at the point we hit this line? Or what if it is still assigned and on the next check 30 seconds later it has already finished? I don't think we should expect that we can always catch the test in the running state.

Martchus (approved):
Ok, I think this works now.

tests/openqa/osautoinst/test_running.pm
```diff
@@ -12,7 +12,8 @@ use base "consoletest";
 use testapi;
 
 sub run {
-    assert_script_run ' ret=false; for i in {1..5} ; do openqa-client jobs state=running | grep --color -z running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ] ; echo $? ', 300;
+    assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
```
baierjan, line 15: Maybe it could be a good idea to move the installation somewhere else, where we can do retries / error-catching as zypper call tends to make the code less stable.

tests/openqa/osautoinst/test_running.pm
```diff
@@ -12,7 +12,8 @@ use base "consoletest";
...
+    assert_script_run 'ret=false; for i in {1..5} ; do openqa-cli api jobs state=running state=done | ack --passthru --color 'running|done' && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
```
perlpunk, line 16: The single quotes break the string.
```suggestion
    assert_script_run q{ret=false; for i in {1..5} ; do openqa-cli api jobs state=running state=done | ack --passthru --color 'running|done' && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]}, 300;
```

tests/openqa/osautoinst/test_running.pm
kalikiana: Not sure about moving it. There's calls all over the test distri anyway - but we can use `zypper_call` of course which will retry up to 5 times among other things.

perlpunk (approved):

kalikiana:
Verification run: https://openqa.opensuse.org/tests/3332303#step/test_running/5

Note that this doesn't address the failure in **test_results** which is due to the match area being off screen.

okurz:
> Verification run: https://openqa.opensuse.org/tests/3332303#step/test_running/5
> 
> Note that this doesn't address the failure in **test_results** which is due to the match area being off screen.

I created
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17212
to address this

PR #17203, opened 2023-06-01
Allow both Maintenance updates as well as PTFs in Public Cloud

pdostal:
* Verification runs: **TBD**


pdostal:
Continuing in #17211

PR #17206, opened 2023-06-02
Add arch_only key for extra repos

dzedro:
Right now the extra customer repos are only for x86_64, other architectures are failing as they can't add not existing repo

- Related ticket:  #17196
- Verification run:
https://openqa.suse.de/tests/11234928 aarch64 jeos
https://openqa.suse.de/tests/11234920 aarch64 jeos
https://openqa.suse.de/tests/11234918 s390x sle
https://openqa.suse.de/tests/11234919 aarch64 sle

PR #17208, opened 2023-06-02
Fix extra repo arch_only condition

dzedro:
- Related ticket: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17206#pullrequestreview-1457195497
- Verification run: https://openqa.suse.de/tests/overview?&distri=sle&build=arch_only

### DIFF
--- a/lib/qam.pm
+++ b/lib/qam.pm
@@ -83,15 +83,21 @@ sub add_repo_if_not_present {
 # https://progress.opensuse.org/issues/90522
 sub add_extra_customer_repositories {
     my $arch = get_var('ARCH');
-    my @repo_list = (
-        {cond => '=12-SP2', name => '12-SP2-LTSS-ERICSSON-Updates', uri => "http://dist.suse.de/ibs/SUSE/Updates/SLE-SERVER/12-SP2-LTSS-ERICSSON/$arch/update/"},
-        {cond => '=12-SP3', name => '12-SP3-LTSS-TERADATA-Updates', uri => "http://dist.suse.de/ibs/SUSE/Updates/SLE-SERVER/12-SP3-LTSS-TERADATA/$arch/update/"},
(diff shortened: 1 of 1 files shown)


mdoucha:
All 4 repos are x86_64 only. Maybe it'd be better to have separate arrays for each arch?

    my %repo_list = (
        x86_64 => [{cond => ..., name => ..., uri => ...}],
        ...
    );

lib/qam.pm
```diff
@@ -83,15 +83,20 @@ sub add_repo_if_not_present {
...
+            {cond => '=15-SP4', name => '15-SP4-ERICSSON-Updates', uri => 'http://dist.suse.de/ibs/SUSE/Updates/SLE-Product-SLES/15-SP4-ERICSSON/x86_64/update/'},
+        ],
+        s390x => [],
+        ppc64le => [],
+        aarch64 => [],
     );
 
-    for my $repo (@repo_list) {
-        add_repo_if_not_present($repo->{uri}, $repo->{name}) if is_sle($repo->{cond}) && defined($repo->{arch_only}) eq 'x86_64';
+    for my $repo (@{$repo_list{$arch}}) {
```
mdoucha, line 98: It might be a good idea to add `return unless $repo_list{$arch};`  just in case somebody tries to run tests on some exotic arch. Otherwise LGTM.

mdoucha (approved):

lib/qam.pm
dzedro: I would rather use die, return will be unnoticed.

dzedro: Added. :+1:

mdoucha: It should be added *before* the loop to prevent an error from casting `undef` to array.

dzedro:
@mdoucha Thanks for looking on the previous PR, it was bad. :face_with_spiral_eyes:

lib/qam.pm
dzedro: :face_in_clouds: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17217

PR #17212, opened 2023-06-05
openQA: Fix finished results hidden out of screen

okurz:
1 job has been created:
 - opensuse-Tumbleweed-DVD-x86_64-Build20230604-openqa_bootstrap@64bit -> https://openqa.opensuse.org/tests/3336705

https://openqa.opensuse.org/tests/3336705#step/test_results/14 shows how the openQA is detected on a scrolled down page despite the job actually still running.

### DIFF
--- a/tests/openqa/webui/test_results.pm
+++ b/tests/openqa/webui/test_results.pm
@@ -46,7 +46,11 @@ sub run {
     handle_notify_popup;
     assert_screen 'openqa-tests';
     assert_and_click 'openqa-tests';
-    assert_and_click 'openqa-job-minimalx';
+    # At this point the openQA job might still be running or already finished.
+    # Ensure to show finished results at the bottom of the screen whenever the
+    # page finished loading
+    send_key_until_needlematch 'openqa-job-minimalx', 'end';
+    click_lastmatch;
     assert_and_click 'openqa-job-details';
     assert_screen 'openqa-testresult', 600;
 }


perlpunk (approved):

tests/openqa/webui/test_results.pm
```diff
@@ -46,7 +46,11 @@ sub run {
     handle_notify_popup;
     assert_screen 'openqa-tests';
     assert_and_click 'openqa-tests';
-    assert_and_click 'openqa-job-minimalx';
+    # At this point the openQA job might still be running or already finished.
+    # Ensure to show finished results at the bottom of the screen whenever the
+    # page finished loading
+    send_key_until_needlematch 'openqa-job-minimalx', 'end';
```
DimStar77, line 52: IIRC, send)key_until_needlematch presses the key before checking the current screen, right? so in essence this carries the risk that minor UI changes makes the text disappear at the top of the screen already

okurz: yes, that could happen. As https://openqa.opensuse.org/tests/3336705#step/test_results/14 shows we just still see the "running" result. a "minor UI change" could of course fail the test but we can't know all ways that tests would fail on UI changes so I wouldn't bother. An alternative I considered was to try to wait for the browser to finish loading the page but I wouldn't trust that to work consistently either. The other only other alternative I could think of is to change openQA itself to e.g. jump to an anchor, waste less screen estate, collapse empty sections, etc.

DimStar77:
@okurz do we still want to merge this? technically, not wrong, but also not really required (as there was an underlying error uncovered by this)

okurz:
> @okurz do we still want to merge this? technically, not wrong, but also not really required (as there was an underlying error uncovered by this)

I think it is required to fix sporadic failures that also happened in the past and might happen again

kalikiana (approved):

PR #17227, opened 2023-06-06
prepare_instance updated to retry on error in wait_for_ssh

m-dati:
P.C. module `prepare_instance` updated to allow check for system up to _retry_ in case of _publickey_ or _unknown_ errors in `wait_for_ssh`

- Related ticket: https://progress.opensuse.org/issues/130444
- Needles: No change
- Verification run: see next posts

### DIFF
--- a/tests/publiccloud/prepare_instance.pm
+++ b/tests/publiccloud/prepare_instance.pm
@@ -37,6 +37,7 @@ sub run {
     my $provider = $self->provider_factory();
     my %instance_args;
     $instance_args{check_connectivity} = 1;
+    $instance_args{ignore_wrong_pubkey} //= 0;    # parameter set for wait_for_ssh: 0 = stop on publickey error.
     $instance_args{use_extra_disk} = {size => $additional_disk_size, type => $additional_disk_type} if ($additional_disk_size > 0);
     $args->{my_provider} = $provider;
     my $instance = $provider->create_instance(%instance_args);
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -356,7 +356,7 @@ sub wait_for_ssh {
     # DMS will return normal user and error will be resolved.
     $args{ignore_wrong_pubkey} //= 0;
     $args{username} //= $self->username();
-    my $delay = $args{ignore_wrong_pubkey} ? 20 : 1;
+    my $delay = $args{ignore_wrong_pubkey} ? 5 : 1;
     my $start_time = time();
     my $instance_msg = "instance: $self->{instance_id}, public IP: $self->{public_ip}";
     my ($duration, $exit_code, $sshout, $sysout);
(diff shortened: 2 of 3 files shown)


m-dati:
VR on target code:
sle-15-SP5-GCE-x86_64-Build0291-publiccloud_consoletests @gce_n1_standard_2
[11265441](https://openqa.suse.de/tests/11265441) PASS

m-dati:
I'd propose to also decrease the sleep time to 5 in [L377](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/197c68aa19f69a155485b67e1af01489dd8cfdaa/lib/publiccloud/instance.pm#L377)

ricardobranco777 (approved):
LGTM

m-dati:
In the last commit I also changed the sleep period in case of ignore_wrong_pubkey from 20 to 5  to increase the checks

tests/publiccloud/prepare_instance.pm
```diff
@@ -37,10 +37,11 @@ sub run {
     my $provider = $self->provider_factory();
     my %instance_args;
     $instance_args{check_connectivity} = 1;
+    $instance_args{ignore_wrong_pubkey} = 1;
     $instance_args{use_extra_disk} = {size => $additional_disk_size, type => $additional_disk_type} if ($additional_disk_size > 0);
     $args->{my_provider} = $provider;
     my $instance = $provider->create_instance(%instance_args);
-    $instance->wait_for_guestregister();
+    # $instance->wait_for_guestregister(); # call removed for refactoring see its head description
```
grisu48, line 44: I don't understand why this was removed. What is the "head description"?

m-dati: yes sorry too simplified text. referring to [this L322](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/1845b47443877b6c32a2ec5f2c2dc75ee07e6220/lib/publiccloud/instance.pm#L322)

grisu48: Ah ok, but why not just fully remove it if it is not needed?

grisu48: ok

asmorodskyi (changes requested):
initial problem which this patch is trying to fix is : 
1. sporadic 
2. has unknown root cause 

before merging this we need certainly find root cause of the issue . Only than we will be able to decide if it is right solution to a problem or not . And I assume that it can not be right approach . Currently we have single test where we set `ignore_wrong_pubkey=1` and  this case is when VM doing chroot during test execution.

m-dati:
I executed VR tests with this patched branch, but also I restarted the previous failed tests in the official [Build group on master.](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=0293&groupid=275).
Both sessions (PR and master) were successful, but one only case of rerun yesterday on master returned a newly the `Permission Denied` error. Today same test [rerun](https://openqa.suse.de/tests/11270608) _pass_ normally, instead.

m-dati:
VR Latest GCE:  
sle-15-SP5-GCE-BYOS-x86_64-Build0303-publiccloud_smoketests.04@gce_n1_highcpu_2
https://openqa.suse.de/tests/11392657

asmorodskyi (approved):
LGTM

PR #17228, opened 2023-06-07
Integrate yupdate script in openQA in order to get e2e playwright tests from a different repo

coolgw:
Integrate yupdate script in openQA in order to get e2e playwright tests from a different repo

- Related ticket: https://progress.opensuse.org/issues/130000
- Needles: na
- Verification run: https://openqa.opensuse.org/tests/3346380

### DIFF
--- a/lib/yam/agama/patch_agama_base.pm
+++ b/lib/yam/agama/patch_agama_base.pm
@@ -1,0 +1,24 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: base class for Patch Agama tests
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package yam::agama::patch_agama_base;
+use base 'opensusebasetest';
+use strict;
+use warnings;
+use testapi;
+use y2_base 'save_upload_y2logs';
+
+sub pre_run_hook {
+    $testapi::password = 'linux';
+}
+
+sub post_fail_hook {
+    my ($self) = @_;
+    select_console 'root-console';
+    y2_base::save_upload_y2logs($self, skip_logs_investigation => 1);
+}
+
+1;
--- a/tests/yam/agama/patch_agama.pm
+++ b/tests/yam/agama/patch_agama.pm
@@ -1,0 +1,23 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: This module use yupdate patch the Agama on Live Medium
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base yam::agama::patch_agama_base;
+use strict;
+use warnings;
+use testapi;
+
+sub run {
+    assert_screen('agama_product_selection', 120);
+
+    select_console 'root-console';
+
+    my ($repo, $branch) = split /#/, get_required_var('YUPDATE_GIT');
+    assert_script_run("yupdate patch $repo $branch", timeout => 60);
+
+    select_console 'installation';
+}
+
+1;
--- a/schedule/yam/agama/agama_installation.yaml
+++ b/schedule/yam/agama/agama_installation.yaml
@@ -4,4 +4,5 @@ description: >
   Playwright test on agama
 schedule:
   - installation/bootloader_start
+  - yam/agama/patch_agama
   - yam/agama/agama_installation
--- a/variables.md
+++ b/variables.md
@@ -215,6 +215,7 @@ YUI_SERVER | string | | libyui REST API server name or ip address.
 YUI_START_PORT | integer | 39000 | Sets starting port for the libyui REST API, on qemu VNC port is then added to this port not to have conflicts.
 YUI_REST_API | boolean | false | Is used to setup environment for libyui REST API, as some parameters have to be set before the VM is started.
 YUI_PARAMS | string | | libyui REST API params required to open YaST modules
+YUPDATE_GIT | string | | Github link used by yast help script yupdate, format is repo#branch such as yast/agama#main.
 ZDUP | boolean | false | Prescribes zypper dup scenario.
 ZDUPREPOS | string | | Comma separated list of repositories to be added/used for zypper dup call, defaults to SUSEMIRROR or attached media, e.g. ISO.
 ZFCP_ADAPTERS | string | | Comma separated list of available ZFCP adapters in the machine (usually 0.0.fa00 and/or 0.0.fc00)


lemon-suse:
LGTM, just the title confused me, I think this is an example how to use yupdate to get e2e playwright tests from a different repo.

tests/yam/agama/patch_agama.pm
```diff
@@ -0,0 +1,35 @@
...
+use testapi;
+use serial_terminal 'select_serial_terminal';
+
+sub run {
+    assert_screen('agama_product_selection', 120);
+
+    select_serial_terminal;
+
+    my ($repo, $branch) = split /#/, get_var('YUPDATE_GIT');
+    if ($repo && $branch) {
```
jknphy, line 19: please leave the control of error to yupdate itself, not need for extra logic, if I would need to read the message "Patch agama SKIP" I wouldn't know what do to...or what does it mean, I would still need to check the patch command (it is repo wrong? it is branch wrong? did I forget to pass this settting?) Besides when doing condition in Perl it is better to not use if/else structure and do something like:
`record_info ("...") unless ($repo && $branch)` which safe spaces, but as  I said, I'd prefer not to have error control, because you have to cover too many cases to help the user, and the patch itself have some defaults values afir.

tests/yam/agama/patch_agama.pm
```diff
@@ -0,0 +1,35 @@
...
+        record_info("Patch agama SKIP!");
+    }
+}
+
+sub pre_run_hook {
+    $testapi::password = 'linux';
+}
+
+sub post_fail_hook {
+    upload_logs('./test-results/test-The-main-page-has-the-Agama-title-chromium/trace.zip');
```
jknphy, line 32: could we create a base class like we did for auto-installation? these methods should not be in test layer.

tests/yam/agama/patch_agama.pm
```diff
@@ -0,0 +1,35 @@
...
+
+sub run {
+    assert_screen('agama_product_selection', 120);
+
+    select_serial_terminal;
+
+    my ($repo, $branch) = split /#/, get_var('YUPDATE_GIT');
+    if ($repo && $branch) {
+        assert_script_run("yupdate patch $repo $branch", timeout => 60);
+        assert_script_run("playwright test --trace on --project chromium --config /usr/share/e2e-agama-playwright/ test.spec", timeout => 120);
```
jknphy, line 21: patching should not include test execution, this should go in a different test module,
something like `agama_installation` but more like `agama_dummy_test`, then in other PR, when we have `agama_installation_lvm` we can delete this test. But yeah, definitely needs to be separated from the patching.

tests/yam/agama/patch_agama.pm
```diff
@@ -0,0 +1,35 @@
...
+use base "consoletest";
+use strict;
+use warnings;
+use testapi;
+use serial_terminal 'select_serial_terminal';
+
+sub run {
+    assert_screen('agama_product_selection', 120);
+
+    select_serial_terminal;
```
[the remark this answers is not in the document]
coolgw: serial console will give more screen indeed but it will not change current tty console, if you switch to root console then you should do select_console(installation) back, otherwise the next module agama_installation will failed, that's why i select serial console finally.

tests/yam/agama/patch_agama.pm
coolgw: The "yudate patch" command will not give any error or message if you have no any parameter, that's why we need handle this.
localhost:~ # yupdate patch
localhost:~ # echo $?
0

And i think we better not use unless if we need "if else" scenario. See below discussion on unless:
https://stackoverflow.com/questions/3048726/usage-of-if-versus-unless-for-perl-conditionals

Record_info give some confusing, i will update it.

tests/yam/agama/patch_agama.pm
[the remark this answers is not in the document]
jknphy: I'd prefer to switch back, to have too many screenshot seems to be a problem for developers trying to figure out what the test does. For one extra line of code one can try to do something with the feedback we get.

tests/yam/agama/patch_agama.pm
jknphy: The unless discussion goes for each programming languages, it wasn't my point, my point was to give the error/info ealier: `record_info ("...") if !($repo && $branch)` instead of having the structure `if { ... } else { ... }` save lines and characters. It is used a lot in the repo, but it is just code style to look more like Perl and less like C.

We can use yupdate without branch as it default to master afir, so it doesn't make sense to control both parameters mandatory.
A simple solution is just to use `get_required_var`, due to we are going to patch always the livecd, test will be in a different repo than the agama project, where the 3 test for CI are packaged in the livecd. In other words, it will not be skip by design.

lib/yam/agama/patch_agama_base.pm
```diff
@@ -0,0 +1,24 @@
+## Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Summary: base class for Patch Agama tests
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package yam::agama::patch_agama_base;
```
jknphy, line 7: this still looks strange, we should be able to use some base cases for agama, but the current one we use for autoyast includes few things we don't need.
Could you please create a new ticket to sort this out?

tests/yam/agama/patch_agama.pm
jknphy: ignore previous comment here, thinking twice we should not repeat this test module, probably just pass a variable to use it with the test we want, Lemon wanted to do initially and now seems that makes sense.
could you create a setting that we can pass [here](https://github.com/lemon-suse/os-autoinst-distri-opensuse/blob/df94211770533cbc89d2c8317c02abe70e9d2acb/tests/yam/agama/agama_installation.pm)

lib/yam/agama/patch_agama_base.pm
coolgw: https://progress.opensuse.org/issues/130573

tests/yam/agama/patch_agama.pm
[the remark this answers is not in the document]
coolgw: If you want select_console 'installation' can take effect you need more setting such as VIDEOMODE to text.
But following failure will happen if you set VIDEOMODE=text.
https://openqa.opensuse.org/tests/3346366#step/bootloader_start/14

Additional info:
lib/susedistribution.pm:490:        $self->add_console('installation', 'tty-console', {tty => check_var('VIDEOMODE', 'text') ? 1 : 7});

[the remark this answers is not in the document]
coolgw: So i prefer still using select_serial_terminal, what do you think?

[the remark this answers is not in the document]
jknphy: never checked that code, didn't know there was those difficulties, could you please create a ticket as well for handling this difficulty.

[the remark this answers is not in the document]
coolgw: Sorry my misleading, the VIDEOMODE should not be setted since agama's UI tty is on 7.  
So i rebase the code again.  We do not need new ticket for investigation this.

PR #17221, opened 2023-06-06
Fix nmcli failures

richiejp:
https://progress.opensuse.org/issues/130408

Caused by: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17170

PR #17230, opened 2023-06-07
utils/set_hostname: Avoid killing the SSH connection on zVM

richiejp:
Fix s390x zVM disconnect: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17221#issuecomment-1579313403

As suggested by @Soulofdestiny it could be supported using s3270 console commands. However I don't think support server will currently work on this setup anyway which is what motivates these changes in the first place.

@czerw Are we likely to have the same issue on svirt or spvm? They seem to connect to the hypervisor over ssh not the SUT itself?

@DimStar77


richiejp:
Waiting for verification: https://openqa.opensuse.org/tests/3341937

czerw:
> @czerw Are we likely to have the same issue on svirt or spvm? They seem to connect to the hypervisor over ssh not the SUT itself?

Probably yes, but there is no powervm backend o3. I even didn't find svirt. Our tests on osd use wicked only so far,  can't recommend amy job at the moment. Maybe @foursixnine, are you aware in your job groups about any pvm/svirt jobs which use networkmanager?

foursixnine:
> > @czerw Are we likely to have the same issue on svirt or spvm? They seem to connect to the hypervisor over ssh not the SUT itself?
> 
> Probably yes, but there is no powervm backend o3. I even didn't find svirt. Our tests on osd use wicked only so far, can't recommend any job at the moment. Maybe @foursixnine, are you aware in your job groups about any pvm/svirt jobs which use networkmanager?

So ALP will uses network manager, so yes we will hit it.

foursixnine (commented):
Maybe you can try using the `reconnect_mgmt_console` thing?

richiejp:
> Waiting for verification: https://openqa.opensuse.org/tests/3341937

Ooops, seems to be missing the casedir. I added spvm and restarted: https://openqa.opensuse.org/tests/3344592

> So ALP will uses network manager, so yes we will hit it.

I think for ALP and new products in general we need clarification on how systemd/NM should behave and if this is a product bug we are working around:
https://bugzilla.opensuse.org/show_bug.cgi?id=1211677

In theory we should not have to restart the network to update DHCP at all. 

> Maybe you can try using the `reconnect_mgmt_console` thing?

Do we need this for current tests? That is, do we run multi-machine tests with support server dDNS on any backend other than QEMU? Possibly we will want it on baremetal, but I would wait until we try using it before making changes like this.

lib/utils.pm
```diff
@@ -1095,6 +1095,11 @@ sub set_hostname {
     save_screenshot;
 
     if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+        if (is_s390x || is_spvm || is_pvm_hmc) {
```
rfan1, line 1098: `is_pvm = is_spvm || is_pvm_hmc` IMO:)

rfan1: btw, I am not sure if `ipmi` can be considered as well

richiejp: hhmm, yes, I think I'll just try and avoid having to check for all these things.

richiejp:
This is getting complicated and it's all to make support server work, which AFAICT only works on QEMU for now. So let's try just checking for support server before doing any workaround.

lib/utils.pm
```diff
@@ -1094,7 +1094,7 @@ sub set_hostname {
     systemctl 'status network.service';
     save_screenshot;
 
-    if (systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
+    if (get_var('USE_SUPPORT_SERVER') && systemctl('is-active NetworkManager', ignore_failure => 1) == 0) {
```
b10n1k, line 1097: this is a lib module and i am concerned about the fact that such a function has a depended condition to `USE_SUPPORT_SERVER`. there is no quarantine that multimachine jobs will use a supportserver. 
What test case does this consider?
is it possible to avoid this and make the function job-variable-free?

richiejp:
Verification passed. I don't have an opinion on whether it would be better to check for the QEMU backend instead. I think the main thing is that the workaround is only implemented in cases where it is needed because it has a lot of side effects.

richiejp:
I restarted those jobs with `is_qemu` to check if it has the same effect.

richiejp:
https://bugzilla.suse.com/show_bug.cgi?id=1212144

czerw:
@DimStar77  we will merge this one today.

richiejp:
@czerw I merged the commits

PR #17235, opened 2023-06-07
Additional Reboot after migration

yarunachalam:
Add additional reboot and validate it is reachable after the boot https://jira.suse.com/browse/CSE-457

- Verification run:(https://openqa.suse.de/tests/11272482#step/migration/781)

### DIFF
--- a/tests/publiccloud/migration.pm
+++ b/tests/publiccloud/migration.pm
@@ -34,6 +34,7 @@ sub run {
     my $pc_rpm = get_var('PUBLIC_CLOUD_DMS_PC_RPM');
     my $pc_act_location = get_var('PUBLIC_CLOUD_PC_ACT_LOCATION');
     my $act_rpm = get_var('PUBLIC_CLOUD_DMS_ACT_RPM');
+    my $add_boot = get_var('PUBLIC_CLOUD_DMS_ADD_REBOOT');
     my $act_url = "$pc_act_location" . "$act_rpm";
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
A comment. Otherwise LGTM.

tests/publiccloud/migration.pm
```diff
@@ -118,6 +119,17 @@ sub run {
     record_info('Migrated Version', $migrated_version);
 
     die("Wrong version: expected: " . $target_version . ", got " . $migrated_version) if ($migrated_version ne $target_version);
+
+    if ($add_boot) {
+        record_info('system reboots after succesfull migration');
+        my ($shutdown_time, $startup_time) = $instance->softreboot(
+            timeout => get_var('PUBLIC_CLOUD_REBOOT_TIMEOUT', 400),
+            ignore_wrong_pubkey => 1
```
asmorodskyi, line 127: you should not use it in this case , we not expecting "wrong pubkey" error here , so if it will happen we should fail

tests/publiccloud/migration.pm
```diff
@@ -118,6 +119,17 @@ sub run {
...
+        );
+
+        my $product_version = $instance->run_ssh_command(cmd => 'sudo cat /etc/os-release');
```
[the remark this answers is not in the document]
yarunachalam: done

tests/publiccloud/migration.pm
yarunachalam: done

PR #17245, opened 2023-06-09
Fix small iscsi server mistake

BillAnastasiadis:
Tests fail in `cluster_add_repos`, seemingly because even though only the hana nodes are amended with the test repository and the iscsi server is excluded, it is not excluded in the next loop which `zypper ref`s. That leads to a crash due to no repositories existing in the server.

- Related ticket: https://jira.suse.com/browse/TEAM-7535
- Verification run:


alvarocarvajald (approved):
LGTM

mpagot (approved):
LGTM

PR #17247, opened 2023-06-09
Deal with no bounce back on powerKVM

lemon-suse:
We have workaround the situation when no bounce back happened on powerkvm, but it is not work when online_migration_setup loaded after bootloader module. Because when showed 'display manager' but test still used wait_boot -> boot_local_disk which will match from the point of bootloader.

We need ensure if no bounce back happend it won't go back to grub2 for the test process, and go ahead, that handle_grub shouldn't be called anymore after dispaly-mananger selected in bootloader module.

- Related ticket: https://progress.opensuse.org/issues/129649
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/11286593#
   VR for SLEM micro image testing: http://openqa.suse.de/tests/11286602#
   VR for powerVM: http://openqa.suse.de/tests/11354720#details

### DIFF
--- a/lib/opensusebasetest.pm
+++ b/lib/opensusebasetest.pm
@@ -881,6 +881,9 @@ sub wait_boot {
         #
         $self->handle_pxeboot(bootloader_time => $bootloader_time, pxemenu => 'pxe-custom-kernel', pxeselect => 'pxe-custom-kernel-selected');
     }
+    # When no bounce back on power KVM, we need skip bootloader process and go ahead when 'displaymanager' matched.
+    elsif (get_var('OFW') && (check_screen('displaymanager', 5))) {
+    }
     else {
         assert_screen([qw(virttest-pxe-menu qa-net-selection prague-pxe-menu pxe-menu)], 600) if (uses_qa_net_hardware() || get_var("PXEBOOT"));
         $self->handle_grub(bootloader_time => $bootloader_time, in_grub => $in_grub);


lib/opensusebasetest.pm
```diff
@@ -881,6 +881,9 @@ sub wait_boot {
         #
         $self->handle_pxeboot(bootloader_time => $bootloader_time, pxemenu => 'pxe-custom-kernel', pxeselect => 'pxe-custom-kernel-selected');
     }
+    # When no bounce back on power KVM, we need skip bootloader process and go ahead when 'displaymanager' matched.
+    elsif (get_var('OFW') && (check_screen('displaymanager', 5))) {
```
jknphy, line 885: why don't you put the check_screen inside the `elsif` ? otherwise you have an empty block.

lemon-suse: I just hope won't affect other conditions, if put the check_screen inside elsif, when check_screen failed, there is no chance to run other if branches. So just see the condition of OFW and check_screen as a whole condition and won't affect other conditions to be checked when it failed.

tinawang123:
Could you please verify one case for powerVM.

lemon-suse:
> Could you please verify one case for powerVM.

Updated the VR in PR description.

jknphy:
if not more feedback, let's give it a try again.

PR #17250, opened 2023-06-10
Add Github Shellcheck workflow

pdostal:
There are many bash scripts in this repository and Shellcheck would be useful.

My plan is to enable this workflow but exclude most of the directories so people can individually choose to "opt-in".

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,19 @@
+name: shellcheck
+on:
+  push:
+    paths:
+      - 'data/publiccloud/**'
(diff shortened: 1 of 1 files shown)


data/publiccloud/x
grisu48, line 1: What is this empty file doing here?

pdostal: Sorry, that was just for debugging.

ricardobranco777 (approved):
LGTM

grisu48 (approved):

okurz (commented):
+1

PR #17255, opened 2023-06-12
Add peering cleanup in base module

BillAnastasiadis:
Add peering destruction to the `clenup` function of `sles4sap_publiccloud_basetest`.
Part of the https://jira.suse.com/browse/TEAM-7535 saga.

- Related ticket: https://jira.suse.com/browse/TEAM-7535
- Verification run: 
FAILING: http://openqaworker15.qa.suse.cz/tests/188749
PASSING: http://openqaworker15.qa.suse.cz/tests/188748 (well, should be passing but there's an unrelated registration error)
NO PEERING: http://openqaworker15.qa.suse.cz/tests/188747

### DIFF
--- a/lib/sles4sap_publiccloud_basetest.pm
+++ b/lib/sles4sap_publiccloud_basetest.pm
@@ -14,6 +14,7 @@ use warnings FATAL => 'all';
 use Exporter 'import';
 use testapi;
 use qesapdeployment;
+use publiccloud::utils;
 
 our @EXPORT = qw(cleanup);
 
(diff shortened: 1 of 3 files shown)


lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -28,6 +28,17 @@ sub cleanup {
         # Skip cleanup if ansible inventory is not present (deployment could not have been done without it)
         next if (script_run 'test -f ' . qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER')));
 
+        record_info('Cleanup', `Executing peering cleanup (if peering is present)`);
+        my $rg = qesap_az_get_resource_group();
```
mpagot, line 32: It has to be in a `if` like:
```
if check_var('PUBLIC_CLOUD_PROVIDER', 'AZURE')
```

alvarocarvajald (approved):
LGTM

lpalovsky:
LGTM

lib/sles4sap_publiccloud_basetest.pm
mpagot: I propose to add an if here for the moment, I'd suggest to have at least 2  implementation before to think about how to re-factor to a unit function

BillAnastasiadis: done, also added check for `IS_MAINTENANCE` and pasted new VRs (still running, but the last one completed and it behaves as expected)

lib/sles4sap_publiccloud_basetest.pm
```diff
@@ -28,6 +29,20 @@ sub cleanup {
         # Skip cleanup if ansible inventory is not present (deployment could not have been done without it)
         next if (script_run 'test -f ' . qesap_get_inventory(get_required_var('PUBLIC_CLOUD_PROVIDER')));
 
+
+        if (is_azure && check_var('IS_MAINTENANCE', 1)) {
```
mpagot, line 33: It is not ok to have this within the `for my $command ('ansible', 'terraform') {` loop. It can be done before the loop.

PR #17263, opened 2023-06-14
ALP: fix repo url

mloviska:
ALP M1 testing

### DIFF
--- a/tests/transactional/host_config.pm
+++ b/tests/transactional/host_config.pm
@@ -34,8 +34,8 @@ sub run {
 
     if (is_alp) {
         # Add Core repo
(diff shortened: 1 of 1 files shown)


czerw:
This part was broken on o3 several time and we habe another round on osd now.

Unfortunately, this breaks ability to add custom core REPO as it purely add repositories synced to openQA. For example current REPO_SLE_ALP_MICRO=ALP-Micro-1.0-x86_64-Build4.1-Media1 which ads something which we really don't need.

We need to add core repo which is used to build ALP https://download.suse.de/ibs/SUSE:/ALP:/Source:/Standard:/Core:/1.0:/Build/standard/ . This was original purpose of this line.

Can we revert this?

PR #17267, opened 2023-06-14
Use logs_from_installation_system on HPC installation

b10n1k:
This PR allows logs_from_installation_system to collect logs from HPC autoyast installation. I think this module should be applicable to be used without modification but the installation console which used here was an obstacle. When we use installation shell first we have to login and then we have to exit, instead of return to it.
The changes do not provide a general solution but it just makes it work for the HPC AY installation particular.

- Related ticket: https://progress.opensuse.org/issues/124032
- https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1645
- Verification run: https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=b10n1k%2Fos-autoinst-distri-opensuse%2317267

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -783,6 +783,10 @@ sub activate_console {
             # login as root, who does not have a password on Live-CDs
             wait_screen_change { enter_cmd "root" };
         }
+        elsif (check_screen('tty2-selected')) {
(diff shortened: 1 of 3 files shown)


czerw:
HPC effort is lowered. There is no planned work in this area.

PR #17274, opened 2023-06-15
Give more tries for script_retry for journalctl module

Zaoliang:
we have sporadic issue with journalctl on Tumbleweed. One possible reason is performance. So increase re-tries for script_retry and check wether this is enough to make the test stabile.

see https://progress.opensuse.org/issues/129304


Zaoliang:
VRs:
no failure of 150 test runs, only 1 softfail for race condition.

https://openqa.opensuse.org/tests/3359675#next_previous

foursixnine (approved):

PR #17280, opened 2023-06-16
Create two modules to replace upgrade_select

JRivrain:
2 new modules added and update of yaml schedule migration_offline_media_x86_64.yaml
See https://progress.opensuse.org/issues/130078.

VR https://openqa.suse.de/tests/11560584

### DIFF
--- a/lib/Yam/SelectForUpdatePage.pm
+++ b/lib/Yam/SelectForUpdatePage.pm
@@ -1,0 +1,32 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handle page that allows to select system/partition to be upgraded.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::SelectForUpdatePage;
+use parent 'Yam::PageBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->{btn_next} = $self->{app}->button({id => 'next'});
+    $self->{tbl_select_for_update} = $self->{app}->table({id => "partition"});
+    return $self;
+}
+
+sub _get_select_for_update {
+    my ($self) = @_;
+    die "Select update page not displayed" unless $self->{tbl_select_for_update}->exist();
+    return $self;
+}
+
+sub next {
+    my ($self) = @_;
+    $self->_get_select_for_update()->{btn_next}->click();
+}
+
+1;
--- a/tests/installation/upgrade/select_for_update.pm
+++ b/tests/installation/upgrade/select_for_update.pm
@@ -1,0 +1,17 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles installer page for handling upgrade of partition with previous installation.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    $testapi::distri->get_select_for_update()->next();
+}
+
+1;
--- a/lib/Distribution/Opensuse/Tumbleweed.pm
+++ b/lib/Distribution/Opensuse/Tumbleweed.pm
@@ -41,6 +41,8 @@ use Installation::SystemRole::SystemRoleController;
 use Installation::Popups::OKPopupController;
 use Installation::Popups::YesNoPopupController;
 use Installation::Popups::LicensePopup;
+use Yam::PreviouslyUsedReposPage;
+use Yam::SelectForUpdatePage;
 use YaST::Bootloader::BootloaderSettingsController;
 use YaST::Firstboot::ConfigurationCompletedController;
 use YaST::Firstboot::HostNameController;
@@ -233,6 +235,14 @@ sub get_kdump_fadump_startup {
 
 sub get_restart_info {
     return YaST::RestartInfoPage->new({app => YuiRestClient::get_app()});
+}
+
+sub get_select_for_update {
+    Yam::SelectForUpdatePage->new();
+}
+
+sub get_previously_used_repos {
+    Yam::PreviouslyUsedReposPage->new();
 }
 
 1;
--- a/lib/Yam/PreviouslyUsedReposPage.pm
+++ b/lib/Yam/PreviouslyUsedReposPage.pm
@@ -1,0 +1,32 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handle page that allows to select system/partition to be upgraded.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::PreviouslyUsedReposPage;
+use parent 'Yam::PageBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->{btn_next} = $self->{app}->button({id => 'next'});
+    $self->{tbl_previously_used_repos} = $self->{app}->table({id => "\"table_of_repos\""});
+    return $self;
+}
+
+sub _get_previously_used_repos {
+    my ($self) = @_;
+    die "Previous repositories page not shown" unless $self->{tbl_previously_used_repos}->exist();
+    return $self;
+}
+
+sub next {
+    my ($self) = @_;
+    $self->_get_previously_used_repos()->{btn_next}->click();
+}
+
+1;
--- a/schedule/migration/migration_offline_media_x86_64.yaml
+++ b/schedule/migration/migration_offline_media_x86_64.yaml
@@ -4,6 +4,7 @@ description:    |
 vars:
   ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
   UPGRADE_TARGET_VERSION: '%VERSION%'
+  YUI_REST_API: '1'
 schedule:
   - migration/version_switch_origin_system
   - boot/boot_to_desktop
@@ -17,8 +18,11 @@ schedule:
   - migration/reboot_to_upgrade
   - migration/version_switch_upgrade_target
   - installation/bootloader
+  - installation/setup_libyui
   - installation/welcome
-  - installation/upgrade_select
+  - installation/upgrade/select_for_update
+  - installation/licensing/accept_license
+  - installation/upgrade/accept_remove_previous_repos
   - installation/scc_registration
   - installation/addon_products_sle
   - installation/resolve_dependency_issues
--- a/tests/installation/upgrade/accept_remove_previous_repos.pm
+++ b/tests/installation/upgrade/accept_remove_previous_repos.pm
@@ -1,0 +1,17 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handles page for detection of previously used repositories.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    $testapi::distri->get_previously_used_repos()->next();
+}
+
+1;


tests/installation/upgrade/remove_previous_repo.pm
```diff
@@ -0,0 +1,19 @@
...
+# Summary: Handles page for detection of previously used repositories.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    my $previous_repo = $testapi::distri->get_previous_repo();
+    $previous_repo->get_previous_repo();
```
JRivrain, line 15: Only an assertion here, selecting the row would toggle "Remove" to "Keep" otherwise.

lib/Yam/Installation/Upgrade/PreviousRepoPage.pm
```diff
@@ -0,0 +1,27 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handle page that allows to select system/partition to be upgraded.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::Installation::Upgrade::PreviousRepoPage;
```
jknphy, line 9: we don't need folders under `lib/Yam/` for now. Let's reorganize later and create all the pages at the same level.

lib/Yam/Installation/Upgrade/PreviousRepoPage.pm
```diff
@@ -0,0 +1,27 @@
...
+use parent 'Installation::Navigation::NavigationBase';
```
jknphy, line 10: we are going to avoid inheritance at all costs for new designs regarding UI elements, but we can inherit here from PageBase and avoid to pass `app => ...` in `lib/Distribution/Opensuse/Tumbleweed.pm` for structural reasons, For compose UI elements we will use composition, but for now we can just remap `press_next`.
It might be some cases where the button needs to be enable or we have to wait for something, so we will specifically wait for that thing in the test level or we would put in its own page the bottom part with the help the next etc. The previous design was not really transparent IMO, the POM is exclusively a repository for locators, the composition or inheritance of POM is another problem to solve always in each design, so as far as we are not really sure, we can just make those POM very trivial following the pattern by the book.

lib/Yam/Installation/Upgrade/SelectUpdatePage.pm
```diff
@@ -0,0 +1,32 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Handle page that allows to select system/partition to be upgraded.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Yam::Installation::Upgrade::SelectUpdatePage;
```
jknphy, line 9: as at the end we will end up with many pages and we already have in the screen quite nice wording in English for selecting name I would recommend to use literally what we find there for the purpose of potential disambiguation, it helps a lot to orientate in the code when everyone is following, in this case "SelectForUpdatePage".

lib/Yam/Installation/Upgrade/SelectUpdatePage.pm
```diff
@@ -0,0 +1,32 @@
...
+use parent 'Installation::Navigation::NavigationBase';
+use strict;
+use warnings;
+
+sub init {
+    my ($self) = @_;
+    $self->SUPER::init();
+    $self->{tbl_select_origin} = $self->{app}->table({id => "partition"});
```
jknphy, line 17: s/tbl_select_origin/tbl_partition_or_system_to_update/
we should avoid to use the internal ids, and always develop the POMs thinking from PoV of what the user can see in the screen.

lib/Yam/Installation/Upgrade/SelectUpdatePage.pm
```diff
@@ -0,0 +1,32 @@
...
+    return $self;
+}
+
+sub get_select_origin {
```
jknphy, line 21: now we have all the logic in one page, without controllers, what we are doing is to preceed with `_` to distinguish private method that public interface and we also will put all the public method first.
So this one can be move down and be called `_get_select_for_update_page` (which will never be called from outside).

lib/Yam/Installation/Upgrade/SelectUpdatePage.pm
```diff
@@ -0,0 +1,32 @@
...
+    my ($self) = @_;
+    die "Select update page not displayed" unless $self->{tbl_select_origin}->exist();
+    return $self;
+}
+
+sub select_origin_system {
+    my ($self, $origin) = @_;
+    $self->get_select_origin->{tbl_select_origin}->select(value => $origin);
```
jknphy, line 29: don't forget for consistency (although Perl doesn't require it) to add `()` after each function.

tests/installation/upgrade/select_upgrade_target_sle.pm
```diff
@@ -0,0 +1,22 @@
+# SUSE's openQA tests
```
jknphy, line 1: `this could be select_partitiong_or_system_to_update.pm`
We can take a look later, before adding `_sle` or something like that, for another ticket.

JRivrain: It's because the module it replaces is also used in opensuse, so unless we use some logic or external test data, we'll probably want that soon.

lib/Yam/Installation/Upgrade/PreviousRepoPage.pm
JRivrain: I assume "remap press_next" means to put it in each page ? So we'll repeat the same in each page ?

JRivrain: I think this will likely cause the issues that made us iplement this in the past, and we'll have to apply the same fix in every single module, no ?

tests/installation/upgrade/select_upgrade_target_sle.pm
JRivrain: Therefore, I don't see why to create a ticket later for what we will most likely do later. I keep it for now, please confirm if you really want that, then I'll remove the _sle.

jknphy: My recommendation is to use the text on the screen everywhere, name of classes, name of modules, etc instead of introducing our own wording. so `select_partitiong_or_system_to_update.pm` is the most suitable name.
There is extensive vocabulary here if we want to do better (but still will not match with the screen, having more trouble for maintenance): https://documentation.suse.com/sles/15-SP4/single-html/SLES-upgrade/index.html
In the vocabulary there is migration target which is the system to be migrated to, but in this case in the name is confusing because we are selecting to system to migrate from, we should use Zero imagination/creativity for names and follow the screens given our own multi-language/culture squad characteristics :) my two cents.

JRivrain: Ok, so in this case would we prefer some logic in the module based on variables, or some test data in the yaml schedule ? I thought this was the current approach, are we going back to test data ?

JRivrain: Or, the specific string is in the page, and we create a different page for opensuse ?

lib/Yam/Installation/Upgrade/PreviousRepoPage.pm
jknphy: yes, it is simple enough to repeat this code for next to be acceptable for now (now we are rewriting to not use controllers and it is an opportunity to start to make it better).

A better option is to have a page called Navigation for mapping Help, Abort, Back, Next.
In Java or in Perl composition can only be done by re-mapping, so you still need a call to a method ->next() which use a method from object Navigation. The problem is that this doesn't work if we detach the object to click next for example from the object that check if the page is there, when we click and the page is not there we would fail and the ribbon might be used with different pages. Then it would require something more elaborated.

We need some common interface method for all pages, so instead of `get_select_origin` all the methods like this should be called the same like `page_loaded()`. Then when you initialize the Navigation in one page, you need to construct this object passing it and instance of the class where is included `$self`,  here there is a draft for that:

```
package XX::YY::SelectForUpdatePage
$self->{Navigation} = XXX::YYY::Navigation->new($self);

sub press_next {
    my ($self) = @_;
    $self->{Navigation}->next();
}

package XXX::YYY::Navigation
sub new {
    my ($class, $args) = @_;
    $self->{base} = $args->{base};

sub next {
    my ($self) = @_;
    $self->{base}->page_loaded();
    $self->{btn_ok}->click();
}
```

We can always leave this point open for future implementation and go for the easy thing to do which is repeat the code for the next button, but we should avoid inheritance (for adding functionality, for structural reason is fine).

tests/installation/upgrade/select_upgrade_target_sle.pm
JRivrain: Why to tackle later when we can do it right away? I don't think we should never re-use code for Sle in Leap, they're 99% the same thing nowadays, so soon or late, we'll want to do something like this. With this particular example in mind, what would you see as the best way to select a different string in the same module ?

1- with test data ?
2- with logic based on variable (eg "if is_sle") ?
3- using Distribution pattern and a specific page (eg YAm/sle/SelectgForUpgradePage.pm) ?
4- by appending to the name "_sle" at the end of the module's name and using schedule ?
5- by just pressing next and not asserting what we select ?

Or something else ? We are changing the way we handle product variations regularly. In the end we have quite some mess, and the doc is obsoleted each time. Maybe we should decide something and stick to it once and for all, so we could document it and stop having the same discussion every 6 months ;-).

I fear you may choose option 5, but I like to actually assert what we select when possible.

lib/Yam/Installation/Upgrade/SelectUpdatePage.pm
JRivrain: I'd rather the libyui id instead (tbl_partition), as in the case of https://openqa.suse.de/tests/11385576#step/remove_previous_repo/1 we would end-up with tbl_these_repositories_were_found_on_the_system_you_are_upgrading...

jknphy: you could shorten those strings, but I have seen really wrong one, Rainer has faced afir, and myself, internal fields usually don't help, long names are really clear, in this case `tbl_repositories_found` (if there is not other system you are upgrading), just an example, still follows the screen which help us with English and avoid our own personal choices for wording.

tests/installation/upgrade/remove_previously_used_repos.pm
```diff
@@ -0,0 +1,19 @@
+# SUSE's openQA tests
```
jknphy, line 1: as we don't validate any removal (as we discuss it could be done with test data in an extended test suite), to keep the current test modules name convention, `accept_previously _used_repos`.

tests/installation/upgrade/remove_previously_used_repos.pm
```diff
@@ -0,0 +1,19 @@
...
+# Summary: Handles page for detection of previously used repositories.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    my $previous_repo = $testapi::distri->get_previously_used_repos();
+    $previous_repo->get_previously_used_repos();
```
jknphy, line 15: test should not contain this getter, this is just a internal, we should show in the test only the actions and validations.  Actually it should be called `_get_previously_used_repos` (in the future we can use a generic name for all the pages, an interface method, is_load or something like that, but not at this point).

Easy to fix if you just call `get_select_for_update` in `press_next` as we used to do for each method in Controllers.

tests/installation/upgrade/select_for_update.pm
```diff
@@ -0,0 +1,19 @@
...
+# Summary: Handles installer page for handling upgrade of partition with previous installation.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use base 'y2_installbase';
+use strict;
+use warnings;
+
+sub run {
+    my $select_update = $testapi::distri->get_select_for_update();
+    $select_update->get_select_for_update();
```
jknphy, line 15: same here.

jknphy (approved):

tests/installation/upgrade/remove_previously_used_repos.pm
JRivrain: The default action of this module is to remove the repos, not to accept it, that would be confusing. I could rename it to just previously _used_repos otherwise.But I think the verb should reflect the action that is done, the convention is to use a verb describing what the module does, in this case, not "accept".

jknphy: my point is that there is not action from the user, defaults are not actions of user, are actions of the products and from the point of view of testing/qe we are more interested on seeing them as an user.
"remove_*" would imply that we remove something but we don't. As we want to re-use this test module for accepting anything that is there, more rows, more removal, etc, we are not validating anything, we are accepting what is in the screen, we didn't call "accept_defaults_*" because it was too long, but we have been following that format for long time, "accept_<name-of-the-page>.pm". In other words, in this case we don't have a verb, is an exception, we need to invent something else like "accept", "next", "go_forward", or something like that.

JRivrain: I get the idea, but in that particular case it's really confusing. I propose accept_remove...

lib/Yam/SelectForUpdatePage.pm
```diff
@@ -0,0 +1,32 @@
...
+    return $self;
+}
+
+sub _get_select_for_update {
+    my ($self) = @_;
+    die "Select update page not displayed" unless $self->{tbl_select_for_update}->exist();
+    return $self;
+}
+
+sub press_next {
```
jknphy, line 27: press_next was what we used in pages with controllers but now there are controlleres, pressing is not important, `next` would suffice.

rakoenig (approved):
LGTM

PR #17288, opened 2023-06-19
Retry script_output with sanity check

lpalovsky:
SAP/HA tests are currently failing because of kernel messages creeping into 
'script_output' output. This PR adds "script_output_retry_check" subroutine 
which checks result against expected regex and retries command couple of times 
upon failure. 

- Related ticket: https://jira.suse.com/browse/TEAM-8076
- Needles: N/A
- Verification run: 
Replicated issue: https://openqa.suse.de/tests/11406510#step/check_after_reboot/31
Issue did not occur: http://openqa.suse.de/tests/11406907#step/check_after_reboot/31

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -71,6 +71,7 @@ our @EXPORT = qw(
   test_flags
   is_not_maintenance_update
   activate_ntp
+  script_output_retry_check
   calculate_sbd_start_delay
   check_iscsi_failure
 );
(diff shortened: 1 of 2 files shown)


emiura:
LGTM

lib/hacluster.pm
```diff
@@ -979,6 +980,44 @@ sub activate_ntp {
...
+C<$regex_string> regular expression to check output against.
+C<$retry> number of retries. Defaults to C<5>.
+C<$sleep> sleep time between retries. Defaults to C<10s>.
+C<$ignore_failure> do not kill the test upon failure.
+
+  Example: script_output_retry_check(cmd=>'hostname', regex_string=>'^node01$', retry=>'100', sleep=>'60', ignore_failure=>'1');
+
+=cut
+
+sub script_output_retry_check {
```
alvarocarvajald, line 1000: Wonder if it would make more sense to move this to `lib/utils.pm` next to the other similar functions.

lib/hacluster.pm
```diff
@@ -979,6 +980,44 @@ sub activate_ntp {
...
+    my %args = @_;
+    my $cmd = $args{cmd} // die('No command specified.');
+    my $regex = $args{regex_string} // die('Regex input missing');
+    my $retry = $args{retry} // 5;
+    my $sleep = $args{sleep} // 10;
+    my $ignore_failure = $args{ignore_failure} // "0";
+    my $result;
+
+    foreach (1 .. $retry) {
+        $result = script_output($cmd);
```
alvarocarvajald, line 1010: You could also do:
```
$result = script_output($cmd, %args)
```
And the extra args you receive in this function can be passed to the inner function.

alvarocarvajald (approved):
LGTM.

Just a minor suggestion and a question.

srinidhiranganna (approved):
LGTM

PR #17293, opened 2023-06-20
Doubling the timeout for seahorse-sshkey-inhibit.

feri:
openQA logs revelaed that 3 seconds is often too short for this dialog to appear and then the button to be clicked.

6 seconds seems to be enough for this test to pass.

Commit fixes poo#130150

- Related ticket: https://progress.opensuse.org/issues/130150


realcharmer:
Gret job, we can merge after CI finishes.

PR #17294, opened 2023-06-20
Use 15-SP4 as base container as suggested in bsc#1212429

tjyrinki:
- Related ticket: https://progress.opensuse.org/issues/130919
- Verification run:
Unable to test properly, as with the change it fails in a different way:
https://openqa.suse.de/tests/11396340#step/sssd_openldap_functional/93

..which is also how it fails with openqa-clone-custom-git-refspec pointing to a repository which does not have a change to the test
https://openqa.suse.de/tests/11396848#step/sssd_openldap_functional/93

So currently not sure how to properly rerun it, but regardless it has been reported that 15-sp3 containers will no longer be available so a switch to 15-SP4 should be done.

### DIFF
--- a/tests/console/sssd_openldap_functional.pm
+++ b/tests/console/sssd_openldap_functional.pm
@@ -36,13 +36,13 @@ sub run {
     }
     zypper_call("in sssd sssd-ldap openldap2-client sshpass docker");
     systemctl('enable --now docker');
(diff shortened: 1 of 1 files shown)


realcharmer:
LGTM

tjyrinki:
(revert if more trouble than the status quo)

PR #17297, opened 2023-06-21
Make sure we are logged out before continuing

ggardet:
- Related ticket: https://progress.opensuse.org/issues/120010
- Verification run: https://openqa.opensuse.org/tests/3370979

### DIFF
--- a/tests/x11/multi_users_dm.pm
+++ b/tests/x11/multi_users_dm.pm
@@ -80,8 +80,8 @@ sub run {
     send_key 'ctrl-d';
     # logout user
     handle_logout;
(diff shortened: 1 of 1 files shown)


Vogtinator:
Looking at the failure it's because of dbus-activated systemd services. The systemd user instance might stay around for a while after logout, so this might not be a reliable detection method either.

ggardet:
> Looking at the failure it's because of dbus-activated systemd services. The systemd user instance might stay around for a while after logout, so this might not be a reliable detection method either.

Making sure to be logged out, should be at least an improvement. 
We could merge this PR and monitor next tests to check if this is enough or if it needs more fixes. WDYT?

PR #17020, opened 2023-05-09
WIP: Add peering creation and deletion in modules

BillAnastasiadis:
This pr adds a new module for installing update repos to the cluster and deleting the peering with the IBSMirror once everything's ready. It also adds the peering deletion in the post fail hooks of relevant modules. 

- Related ticket: https://jira.suse.com/browse/TEAM-7487
- Verification run: 
post_fail_hook: http://openqaworker15.qa.suse.cz/tests/153895#step/cluster_add_repos/59
normal peering delete: http://openqaworker15.qa.suse.cz/tests/154790#step/cluster_add_repos/62

(failure not related to the pr in the second case)

PR #17300, opened 2023-06-21
Change mr_test_azure config to be more generic

mpagot:
Add VNET_ADDRESS_RANGE and SUBNET_ADDRESS_RANGE to the qesap config used in mr_tests that does not need network peering to IBS. Change the test module to populate these settings only for Azure tests and not only for IS_MAINTENANCE tests. This is a double step: 1. remove an `if` test that in any case will break when extending the test to AWS, 2. Implement slot IPs also for tests that does not need to perform net peering, it is useless but also harmless.

Related ticket: TEAM-7487 as follow up of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17020/files

Verification run:
- mr_test with 15sp5 image from IBS (_uri config) https://openqa.suse.de/tests/11406964 https://openqa.suse.de/tests/11407265
- mr_test maintenance http://openqaworker15.qa.suse.cz/tests/192293

### DIFF
--- a/data/sles4sap/qe_sap_deployment/mr_test_azure.yaml
+++ b/data/sles4sap/qe_sap_deployment/mr_test_azure.yaml
@@ -13,6 +13,8 @@ terraform:
   provider: 'azure'
   variables:
     # GENERAL VARIABLES #
+    vnet_address_range: "%VNET_ADDRESS_RANGE%"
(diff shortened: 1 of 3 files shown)


BillAnastasiadis (approved):

alvarocarvajald (approved):
LGTM

lilyeyes (commented):
LGTM.

PR #17302, opened 2023-06-22
Patch and Reboot for Maintenance hanasr jobs

BillAnastasiadis:
This solves the problem of `qesap_ansible` failing if `patch_and_reboot` runs before it.

The problem was in the ssh settings: `patch_and_reboot` runs a couple of commands via ssh on the node, and due to the ssh settings and some problem this created a multiplexed ssh connection that broke after the command was executed, leaving a stale ssh socket file behind. Later, ansible ssh command tried to re-use that socket file which lead to failure.

Adding `-o ControlMaster=no` to the ssh command in `patch_and_reboot` disables the ssh multiplexing for that command, which leads to the job succeeding.

- Related ticket: https://jira.suse.com/browse/TEAM-8035
- Verification run: http://openqaworker15.qa.suse.cz/tests/193006


mpagot (approved):
LGTM

alvarocarvajald (approved):
LGTM.

Verification Run looks beautiful :P

PR #17304, opened 2023-06-22
Set WONTFIX to soft failure in cc tests on 15-SP3

realcharmer:
This prevents the openqa_review bot from reopening [poo#116683](https://progress.opensuse.org/issues/116683), since the issue will not be fixed upstream.


paolostivanin:
LGTM

PR #17306, opened 2023-06-22
susedistribution: bootloader_setup: Add user-sut-serial

pevik:
user-serial is svirt serial console for user. Adding a link helps switching between root and non root user in `select_serial_terminal()`. This commit is similar to 4a89bfb0f.

Verification run:
with updated os-autoinst https://github.com/os-autoinst/os-autoinst/pull/2330 (API version 39)
- https://openqa.suse.de/tests/overview?build=sut-serial-for-user
- https://openqa.opensuse.org/tests/overview?build=sut-serial-for-user (please ignore s390x, because virtio console test does not work on z/VM)
- https://openqa.suse.de/tests/11441410#step/zypper_lifecycle/1 (sle-15-SP4-Server-DVD-Updates-s390x-Buildlifecycle-qam-minimal+base@s390x-kvm-sle12, thanks to @dzedro)

Ticket: [poo#131258](https://progress.opensuse.org/issues/131258)

NOTE: WIP is due waiting API version 39 being installed to all workers.

### DIFF
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -19,7 +19,7 @@ use version_utils qw(is_microos is_sle_micro is_jeos is_leap is_sle is_tumblewee
 use mm_network;
 use Utils::Backends;
 
-use backend::svirt qw(SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT SERIAL_CONSOLE_DEFAULT_DEVICE SERIAL_CONSOLE_DEFAULT_PORT);
+use backend::svirt qw(SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT SERIAL_CONSOLE_DEFAULT_DEVICE SERIAL_CONSOLE_DEFAULT_PORT SERIAL_USER_TERMINAL_DEFAULT_DEVICE SERIAL_USER_TERMINAL_DEFAULT_PORT);
 
 our @EXPORT = qw(
   add_custom_grub_entries
@@ -1134,13 +1134,22 @@ sub zkvm_add_pty {
             target_port => SERIAL_CONSOLE_DEFAULT_PORT
         });
 
-    # sut-serial (serial terminal: emulation of QEMU's virtio console for svirt)
+    # ssh-virtsh-serial for root (serial terminal: emulation of QEMU's virtio console for svirt)
     $svirt->add_pty(
         {
             pty_dev => SERIAL_TERMINAL_DEFAULT_DEVICE,
             pty_dev_type => 'pty',
             target_type => 'virtio',
             target_port => SERIAL_TERMINAL_DEFAULT_PORT
+        });
+
+    # ssh-virtsh-serial for user (serial terminal: emulation of QEMU's virtio console for svirt)
+    $svirt->add_pty(
+        {
+            pty_dev => SERIAL_USER_TERMINAL_DEFAULT_DEVICE,
+            pty_dev_type => 'pty',
+            target_type => 'virtio',
+            target_port => SERIAL_USER_TERMINAL_DEFAULT_PORT
         });
 }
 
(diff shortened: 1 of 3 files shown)


pevik:
@dzedro FYI

lib/bootloader_setup.pm
```diff
@@ -21,6 +22,10 @@ use Utils::Backends;
 
 use backend::svirt qw(SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT SERIAL_CONSOLE_DEFAULT_DEVICE SERIAL_CONSOLE_DEFAULT_PORT);
 
+if (isotovideo::get_version() >= 39) {
```
pevik, line 25: Not sure why CI complains about `Use of uninitialized value in numeric ge (>=)` here :(. @dzedro any idea?

lib/susedistribution.pm
```diff
@@ -18,11 +18,17 @@ use utils qw(
...
 use Utils::Backends;
-use backend::svirt qw(SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT);
+
 use Cwd;
 use autotest 'query_isotovideo';
 use isotovideo;
 
+use backend::svirt qw(SERIAL_TERMINAL_DEFAULT_DEVICE SERIAL_TERMINAL_DEFAULT_PORT);
+
+if (isotovideo::get_version() >= 39) {
```
pevik, line 28: And here also CI complains about `Use of uninitialized value in numeric ge (>=)` :(.

dzedro: I would not use `if (isotovideo::get_version() >= 39)`, just define it as everything else, we can wait until workers are updated, it's not like we can have this only on some workers.

pevik: Other option would be to detect if `eval "use use backend::svirt qw(SERIAL_USER_TERMINAL_DEFAULT_DEVICE SERIAL_USER_TERMINAL_DEFAULT_PORT)";` works. But ok, if it's not a problem, I'll just define new variables and we will wait till workers are updated. Thanks!

dzedro: Ye, but IMO we don't need to complicate it.

pevik:
@dzedro The failure in isotovideo `main: "SERIAL_USER_TERMINAL_DEFAULT_DEVICE" is not exported by the backend::svirt module` means that CI is using old os-autoinst. I wonder when it's going to be updated.

dzedro (approved):

lib/susedistribution.pm
```diff
@@ -476,10 +478,15 @@ sub init_consoles {
...
         set_var('SVIRT_VNC_CONSOLE', 'sut');
     } else {
-        # sut-serial (serial terminal: emulation of QEMU's virtio console for svirt)
+        # sut-serial for root (serial terminal: emulation of QEMU's virtio console for svirt)
         $self->add_console('root-sut-serial', 'ssh-virtsh-serial', {
                 pty_dev => SERIAL_TERMINAL_DEFAULT_DEVICE,
                 target_port => SERIAL_TERMINAL_DEFAULT_PORT});
+
+        # sut-serial for user (serial terminal: emulation of QEMU's virtio console for svirt)
+        $self->add_console('sut-serial', 'ssh-virtsh-serial', {
```
dzedro, line 487: Why not `user-sut-serial` ?

pevik:
It looks like the problem was solved after I rebased. Therefore testing in https://openqa.opensuse.org/tests/overview?build=sut-serial-for-user and https://openqa.suse.de/tests/overview?build=sut-serial-for-user.

lib/susedistribution.pm
pevik: Good point, renamed.

pevik:
@dzedro IMHO it's working now(e.g. https://openqa.suse.de/tests/11529483#step/zypper_lifecycle/1 and others), thus I could remove WIP, so that it could be merged. WDYT?
I also need to upload `serial_terminal_user.txt`, but that can be a separate effort (in different git repository anyway).

dzedro:
> @dzedro IMHO it's working now(e.g. https://openqa.suse.de/tests/11529483#step/zypper_lifecycle/1 and others), thus I could remove WIP, so that it could be merged. WDYT? I also need to upload `serial_terminal_user.txt`, but that can be a separate effort (in different git repository anyway).

I agree, let's merge this.

PR #17309, opened 2023-06-23
Cover unexpected values appearing in SBD calculation

lpalovsky:
Module 'check_after_reboot' currently fails because of unexpected values 
appearing in SBD delay formula. This PR converts unexpected float parameters to 
int. It also sets default value in case 'pcmk_delay_max' is not obtainable. 

- Related ticket: https://jira.suse.com/browse/TEAM-8104
- Needles: N/A
- Verification run: 
http://openqa.suse.de/tests/11443306#step/check_after_reboot/35
http://openqa.suse.de/tests/11443643#step/check_after_reboot/36

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -104,8 +104,8 @@ our $softdog_timeout = bmwqemu::scale_timeout(60);
 our $prev_console;
 our $join_timeout = bmwqemu::scale_timeout(60);
 our $default_timeout = bmwqemu::scale_timeout(30);
-our $corosync_token = q@corosync-cmapctl | awk -F " = " '/runtime.config.totem.token\s/ {print $2/1000}'@;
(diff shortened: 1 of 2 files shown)


emiura:
LGTM

bs-suse:
TLDR LGTM

The changes you made are looking good to me and I like that you moved the parameter collection into the `collect_sbd_delay_parameters` function.

Briefly I was thinking about rounding instead of `int()`, but I looked that up and now I'm fine with your way with that as well.

srinidhiranganna (approved):
LGTM

lpalovsky:
@bs-suse Yea I wanted to round at first as well, but I found only the printf method which would introduce more funny characters into the command.

alvarocarvajald (commented):
LGTM

PR #17310, opened 2023-06-23
Public Cloud: Switch Azure x86_64 disks to compute galleries  * Add V1/V2 to image definition names.

pdostal:
* Both x86 as well as Arm64 images are now managed in image galleries.
* Image definition name now contains architecture and SKU as it needs to be unique.
* Image definition offer now contains architecture as it needs to be unique.
* The architecture is set in `PUBLIC_CLOUD_ARCH` and that can be improved in the future.

- Related pull requests: #17375, [gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1266](https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1266)
- Related ticket: [poo#114517](https://progress.opensuse.org/issues/114517)
- Scheduling changes:
  * Upload: `PUBLIC_CLOUD_REGION=westeurope,westus PUBLIC_CLOUD_AZURE_OFFER=`
  * Others: `PUBLIC_CLOUD_AZURE_OFFER= PUBLIC_CLOUD_IMAGE_URI=auto`
- Verification run:
  * gen1 x86: [upload_img](https://pdostal-server.suse.cz/tests/4677), [img_proof](https://pdostal-server.suse.cz/tests/4678)
  * gen2 Arm64: [upload_img](https://pdostal-server.suse.cz/tests/4683), [img_proof](https://pdostal-server.suse.cz/tests/4684)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -66,9 +66,9 @@ sub find_img {
     return if (!$self->resource_exist());
 
     ($name) = $name =~ m/([^\/]+)$/;
-    my $sku = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';
+    my $gen = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';
     $name =~ s/\.xz$//;
-    $name =~ s/\.vhdfixed$/-$sku.vhd/;
+    $name =~ s/\.vhdfixed$/-$gen.vhd/;
 
     my $storage_account = get_var('PUBLIC_CLOUD_STORAGE_ACCOUNT', 'eisleqaopenqa');
     my $key = $self->get_storage_account_keys($storage_account);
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
LGTM

asmorodskyi (changes requested):
there is a mess around use of `$sku` variable . In line 154 ( right side numbering ) sku is defined as `my $sku = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';`  so sku will be V1 or V2 which is one meaning and here it will be gen2 or gen1 which is a bit different story . I suggest : 
1. extract all places where we generating definition into dedicated function `get_definition`
2. use different name for holding V1/V2 ( I am not sure how to call it but it is not sku )

asmorodskyi:
> Code changes look good to me, but we will need to have at least one full scenario tested (i.e. `upload_img` and one of the subsequent test runs e.g. `publiccloud_containers` or `img_proof`). I would expect that we need some changes with the image determination there as well, if we switch over to compute galleries.

Fully agree with Felix here - changes are LGTM ( except story with definition generation see comment above )  but we certainly need to cover more cases to prove that it works properly : 
1. upload of image in two different regions 
2. run tests on uploaded images in two different regions

pdostal:
Ouf `PUBLIC_CLOUD_AZURE_OFFER` must be unique for each architecture because of the following error:
```
Code: OperationNotAllowed
Message: Gallery image: 'SLE-AZURE-BYOS-15-SP5-V2' identified by (publisher: 'qe-c', offer: 'sles15sp5-byos', sku: 'gen2') already exists. Choose a different publisher, offer, sku combination.
```

lib/publiccloud/azure.pm
```diff
@@ -66,9 +66,9 @@ sub find_img {
     return if (!$self->resource_exist());
 
     ($name) = $name =~ m/([^\/]+)$/;
-    my $sku = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';
+    my $gen = (check_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2')) ? 'V2' : 'V1';
     $name =~ s/\.xz$//;
-    $name =~ s/\.vhdfixed$/-$sku.vhd/;
+    $name =~ s/\.vhdfixed$/-$gen.vhd/;
```
asmorodskyi, line 71:
```suggestion
    $name =~ s/\.vhdfixed.xz$/-$gen.vhd/;
```
will allow also drop line above ? :thinking:

pdostal: What if we decide to provide uncompressed image? I mean, it's corner case but it may happen (maybe in development or in manual yet openQA-assisted testing)?

asmorodskyi: not that I get you point , if we will decide to do so we will need to do extra PR in both version , right ? And my suggestion not over complicate works needs to be done in such case . But it is not a big deal if you like to keep it let's keep it

lib/publiccloud/azure_client.pm
```diff
@@ -42,6 +42,8 @@ sub init {
           . '}');
     $self->az_login();
     assert_script_run("az account set --subscription \$ARM_SUBSCRIPTION_ID");
+    #$self->region =~ s/,.*$//g;
```
asmorodskyi, line 45: can you please remove this ?

pdostal:
Please do not merge yet. I need to get rid of the `PUBLIC_CLOUD_AZURE_OFFER` required variable.

asmorodskyi:
> Please do not merge yet. I need to get rid of the `PUBLIC_CLOUD_AZURE_OFFER` required variable.

I think the best way to address that is to convert PR into a draft

pdostal:
Please merge this together with [gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1266](https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1266).

PR #17312, opened 2023-06-23
mr_test get the os_image from PUBLIC_CLOUD_IMAGE_ID

mpagot:
Change the qe-sap-deployment used by mr_test in AWS to get the value for the Terraform variable os_image from the PUBLIC_CLOUD_IMAGE_ID setting. It will be in the form of the ami- and no more the image name.

- Related ticket: TEAM-8098
- Verification run:


mpagot:
We do not want it for the moment

PR #17314, opened 2023-06-25
Add retries on autofs client to sync data with server

rfan1:
https://progress.opensuse.org/issues/131291
At the same time, remove a mutex lock which is not required.

- Verification run: 
[autofs](https://openqa.suse.de/tests/11437911#dependencies)


dzedro (approved):

DeepthiYV (approved):
LGTM

PR #17320, opened 2023-06-26
prepare_instance update in P.C. to handle ssh connectivity errors

m-dati:
publiccloud `prepare_instance` update to handle ssh connectivity errors.

As followup of previous PR#17227, in that PR at end we simply applied key rescan refresh only, skipping the retry on error.
But,as explained in the poo [note 11](https://progress.opensuse.org/issues/130444#note-11) the last new tests errors should be likely due to a rush condition of protocol checks in the multiple types of keys. In fact in those `journal` logs appeared 3 keys checks, where RSA is ok, but being last in the attempts sequence of instance VM, those tests failed due to stop at-first-error.

Therefore, the here reproposed retry on error should solve the issue.

- Related ticket: https://progress.opensuse.org/issues/130444
- Needles: none
- Verification run: TBD see next posts.

### DIFF
--- a/tests/publiccloud/prepare_instance.pm
+++ b/tests/publiccloud/prepare_instance.pm
@@ -37,7 +37,7 @@ sub run {
     my $provider = $self->provider_factory();
     my %instance_args;
     $instance_args{check_connectivity} = 1;
-    $instance_args{ignore_wrong_pubkey} //= 0;    # parameter set for wait_for_ssh: 0 = stop on publickey error.
+    $instance_args{ignore_wrong_pubkey} //= 1; # parameter set for wait_for_ssh: 0 = stop on publickey error; 1 = retry to connect until other events or timeout.
     $instance_args{use_extra_disk} = {size => $additional_disk_size, type => $additional_disk_type} if ($additional_disk_size > 0);
     $args->{my_provider} = $provider;
     my $instance = $provider->create_instance(%instance_args);


tests/publiccloud/prepare_instance.pm
```diff
@@ -37,7 +37,7 @@ sub run {
     my $provider = $self->provider_factory();
     my %instance_args;
     $instance_args{check_connectivity} = 1;
-    $instance_args{ignore_wrong_pubkey} //= 0;    # parameter set for wait_for_ssh: 0 = stop on publickey error.
+    $instance_args{ignore_wrong_pubkey} //= 1; # parameter set for wait_for_ssh: 0 = stop on publickey error; 1 = retry to connect until other events or timeout.
```
asmorodskyi, line 40: except one exception where on instance reboot there is chroot happening we not expecting public key change and before it never was an issue do you know why now we should accept it ?

m-dati: In the poo note-11 I explained what the failed tests logs are showing.

Let we take the failed test https://openqa.suse.de/tests/11401949.

After ssh port ok, The [ssh keyscan](https://openqa.suse.de/tests/11401949#step/prepare_instance/75) provides 3 keys: ed25519, rsa, nistp256.
In the next [ssh run](https://openqa.suse.de/tests/11401949#step/prepare_instance/83) we have as result a "strange temporay" error result `kex_exchange_identification: read: Connection reset by peer`; _strange_ because this error has been found _always present_ even in _passed_ tests, _temporary_ because it does not affect the next key resolution, see in log `Reloading OpenSSH Daemon` then `Accepted publickey`.

In the [journal]( https://openqa.suse.de/tests/11401949/file/prepare_instance-journalctl.log), we see the key checks sequence provided by the instance:
```
...
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1614]: error: kex_exchange_identification: Connection closed by remote host
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1615]: Connection closed by 195.135.221.27 port 46952 [preauth]
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1617]: Connection closed by 195.135.221.27 port 46926 [preauth]
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1619]: Connection closed by 195.135.221.27 port 46942 [preauth]
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1616]: Unable to negotiate with 195.135.221.27 port 46960: no matching host key type found. Their offer: sk-ecdsa-sha2-nistp256@openssh.com [preauth]
Jun 21 00:40:01 opetnqa-suse-de-de92d346dfd3bcf6 sshd[1618]: Unable to negotiate with 195.135.221.27 port 46962: no matching host key type found. Their offer: sk-ssh-ed25519@openssh.com [preauth]
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 systemd[1]: Startup finished in 5.261s (kernel) + 3.124s (initrd) + 20.611s (userspace) = 28.997s.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 systemd[1]: Reloading OpenSSH Daemon.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1583]: Received SIGHUP; restarting.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1583]: Server listening on 0.0.0.0 port 22.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1583]: Server listening on :: port 22.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 systemd[1]: Reloaded OpenSSH Daemon.
Jun 21 00:40:01 openqa-suse-de-de92d346dfd3bcf6 sshd[1630]: Accepted publickey for susetest from 195.135.221.27 port 46984 ssh2: RSA SHA256:MqNeSY1qN7676L1LFG+VYw/LcQRLZAFHlEtOskViQHM
...
```
The Perl test driver `waith_for_ssh` in this case is set to die at first error, therefore the first `Unable to negotiate...` leads to fail.

pdostal: Isn't that "strange" error just the master connection being killed because of the server has been restarted?

pdostal: Maybe something like this could help?
```bash
ssh -O exit sut
```

m-dati: It is in the create instance phase, the first access to the new instance, so isn't a case of server restart

m-dati: but anyway I saw thet error present in every P.C. test journal once, passed or failed

m-dati: Also a more thing to note is that line `... systemd[1]: Startup finished in...` that is quite the result of the `systemd-analyze` command: it means that the instance is up and running.

grisu48: @m-dati Please see https://progress.opensuse.org/issues/130444#note-12 - I think this is just a simple race condition, and retrying instead of ignoring would be the better approach.

m-dati: Yes, thanks in fact this PR means to retry when that errors occur and [re-loop to next checks](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/cfc785052e30e616046851ac0a7ce57c9a4dce16/lib/publiccloud/instance.pm#L388).

grisu48: In that case, I think this is a valid solution.

For my future self: Please note that `ignore_wrong_pubkey` does not mean that the ssh authentication is weakened, it just means that the call does not fail immediately. It effectively means to retry on failures related to ssh public key errors. We do not ignore them, but re-try.

grisu48 (approved):

ricardobranco777 (approved):
LGTM.  Hope this simple change fixes stuff.

tests/publiccloud/prepare_instance.pm
m-dati: OK,thanks.
Anyway, adding more complexity, but could it be useful to upload a more log, where to store all resulted failure lines (pubkey ,...), before to go next retry and notify at end in the report info something like: 'other errors retries, see log xxx' ?
Otherwise no log, but a simpler _other_errors_retries_  counter :-) ?

grisu48:
> **so I think we need to fully drop concept of wrong public key early exit from the function**

See https://progress.opensuse.org/issues/130444#note-12 - while the instance is booting this error sometimes arises. We cannot simply drop this option without proper replacement, which will be doing just exactly the same.

asmorodskyi:
> > **so I think we need to fully drop concept of wrong public key early exit from the function**
> 
> See https://progress.opensuse.org/issues/130444#note-12 - while the instance is booting this error sometimes arises. We cannot simply drop this option without proper replacement, which will be doing just exactly the same.

please read more carefully what I suggested . we drop `if ($sysout =~ m/Permission denied \(publickey\).*/) `  AND `else {    # FAIL: Connection refused or else` so in case of ANY errors we keep looping until timeout . which will perfectly cover any raise conditions which you mention in your message

asmorodskyi (approved):
LGTM with note that work on my suggestion will continue in different PR

PR #17322, opened 2023-06-26
Wait for 'zypper purge-kernels' done in update_minimal test

rfan1:
https://progress.opensuse.org/issues/131213
After installing new kernel package, system will try to remove the old ones via 'zypper -n purge-kernels', it may block other zypper processes.

- Verification run: https://openqa.suse.de/tests/11451470

### DIFF
--- a/tests/qam-minimal/update_minimal.pm
+++ b/tests/qam-minimal/update_minimal.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2016-2018 SUSE LLC
+# Copyright 2016-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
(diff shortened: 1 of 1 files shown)


dzedro:
There is this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/zypper.pm#L32

rfan1:
> There is this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/zypper.pm#L32

Glad to know, thanks much!

rfan1:
> > There is this https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/zypper.pm#L32
> 
> Glad to know, thanks much!

https://openqa.suse.de/tests/11451470

dzedro (approved):

PR #17325, opened 2023-06-27
Introduce obsproject test in python

michaelgrifalconi:
Introduce obsproject python test:

Introduce new openQA test written in Python, in a new directory structure to make more visible the squad which is responsible for it.
We should clarify if this structure is ok or not (maybe x11 dir inside qe-core?)

Test header missing, I guess we should use the same as we do on perl tests

- Related ticket: https://progress.opensuse.org/issues/109891
- Needles: already created
- Verification run: https://openqa.opensuse.org/tests/3382602

### DIFF
--- a/schedule/functional/extra_tests_gnome.yaml
+++ b/schedule/functional/extra_tests_gnome.yaml
@@ -26,7 +26,6 @@ conditional_schedule:
                 - x11/steam
                 - x11/wine
                 - x11/chrome
-                - qe-core/x11/openbroadcastersoftware.py
     python_liblouis:
         DISTRI:
             sle:
--- a/schedule/functional/extra_tests_gnome.yaml
+++ b/schedule/functional/extra_tests_gnome.yaml
@@ -26,7 +26,7 @@ conditional_schedule:
                 - x11/steam
                 - x11/wine
                 - x11/chrome
-                - python/qe-core/openbroadcastersoftware.py
+                - qe-core/x11/openbroadcastersoftware.py
     python_liblouis:
         DISTRI:
             sle:
--- a/schedule/functional/extra_tests_gnome.yaml
+++ b/schedule/functional/extra_tests_gnome.yaml
@@ -26,6 +26,7 @@ conditional_schedule:
                 - x11/steam
                 - x11/wine
                 - x11/chrome
+                - python/qe-core/openbroadcastersoftware.py
     python_liblouis:
         DISTRI:
             sle:
--- a/tests/python/qe-core/openbroadcastersoftware.py
+++ b/tests/python/qe-core/openbroadcastersoftware.py
@@ -1,0 +1,35 @@
+from testapi import *
+
+def run(self):
+
+    perl.require("serial_terminal")
+    perl.require("utils")
+    perl.require("x11utils")
+
+
+    perl.serial_terminal.select_serial_terminal()
+    perl.utils.zypper_call("in flatpak")
+    assert_script_run("flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo")
+    assert_script_run("flatpak install -y com.obsproject.Studio",300)
+
+    select_console('x11')
+    perl.x11utils.turn_off_gnome_screensaver()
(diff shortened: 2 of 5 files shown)


tests/python/qe-core/openbroadcastersoftware.py
```diff
@@ -0,0 +1,35 @@
...
+
+    perl.require("serial_terminal")
+    perl.require("utils")
+    perl.require("x11utils")
+
+
+    perl.serial_terminal.select_serial_terminal()
+    perl.utils.zypper_call("in flatpak")
+    assert_script_run("flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo")
+    assert_script_run("flatpak install -y com.obsproject.Studio",300)
```
foursixnine, line 13: Any particular reason why using flatpak vs using the obs from repos?

schedule/functional/extra_tests_gnome.yaml
```diff
@@ -26,6 +26,7 @@ conditional_schedule:
                 - x11/steam
                 - x11/wine
                 - x11/chrome
+                - python/qe-core/openbroadcastersoftware.py
```
foursixnine, line 29: Let's go with the change you proposed for this `tests/qe-core/x11/openbroadcastersoftware` 

Is the `.py` intentional?

foursixnine:
One thing I see is that there are no calls that pass named arguments, for instance:

`assert_script_run("dd if=/dev/zero of=$hard_disk_name count=1M", timeout => 1500, fail_message => "Failed to wipe hard disk clean on $hard_disk_name"`

schedule/functional/extra_tests_gnome.yaml
michaelgrifalconi: Will do! `.py` is very intentional! By default we assume no extension means `.pm` files. So for python the extension is mandatory

michaelgrifalconi:
> One thing I see is that there are no calls that pass named arguments, for instance:
> 
> `assert_script_run("dd if=/dev/zero of=$hard_disk_name count=1M", timeout => 1500, fail_message => "Failed to wipe hard disk clean on $hard_disk_name"`

Great point, that is a tricky thing to spot.

Have a look at the `x11_start_program` function here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/01d28e9e4ca03bed18ddf0432523a8506d2ddc57/lib/susedistribution.pm#L278 

And a random example of it being used in perl:  here https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/01d28e9e4ca03bed18ddf0432523a8506d2ddc57/lib/y2_mm_common.pm#L30

We are doing the same thing here `perl.x11utils.x11_start_program("flatpak run com.obsproject.Studio", "target_match", "obsproject-wizard")` .

foursixnine:
waiting for WIP to be removed

foursixnine:
@michaelgrifalconi one thing, before adding it into the main stuff,  let's have this run for a build on the Devel jobgroup

tests/python/qe-core/openbroadcastersoftware.py
michaelgrifalconi: I haven't found official repos, see https://software.opensuse.org/package/obs-studio and https://github.com/obsproject/obs-studio/wiki/install-instructions#unofficial-builds

foursixnine:
@michaelgrifalconi can you check the CI?

michaelgrifalconi:
Ok I duplicated the extra_test_gnome schedule to add obs test in development. After is merged I will create a test suite and add it on the development job group on o3

foursixnine (approved):

michaelgrifalconi:
VR before addin the schedule to dev job: https://openqa.opensuse.org/tests/3410682#

alvarocarvajald:
Sorry I took so long to read this and couldn't comment before merging: this is awesome! 

It would be nice if there weren't those `wait_serial()` failures in one of the steps in the Python module. For example:

https://openqa.opensuse.org/tests/3410682#step/openbroadcastersoftware/14
https://openqa.opensuse.org/tests/3382602#step/openbroadcastersoftware/14

But these seem to come from the `perl.x11utils.turn_off_gnome_screensaver()` call, so out of the scope of this PR. :)

All in all, awesome.

PR #17328, opened 2023-06-28
Set milestone test flag to agama_base.pm

coolgw:
Set milestone test flag to agama_base.pm

- Related ticket: https://progress.opensuse.org/issues/131483
- Needles: na
- Verification run: https://openqa.opensuse.org/tests/3420599#

### DIFF
--- a/lib/Yam/agama/agama_base.pm
+++ b/lib/Yam/agama/agama_base.pm
@@ -33,4 +33,8 @@ sub post_run_hook {
     ensure_serialdev_permissions;
 }
 
+sub test_flags {
+    return {fatal => 1};
+}
+
 1;


sofiasyria:
LGTM

rakoenig:
LGTM

lib/yam/agama/agama_base.pm
```diff
@@ -33,4 +33,8 @@ sub post_run_hook {
     ensure_serialdev_permissions;
 }
 
+sub test_flags {
+    return {fatal => 1, milestone => 1};
```
lemon-suse, line 37: I think we don't need to set milestone since we haven't such requirement for a snapshot here.

coolgw: I just use this flag to stop failed case, not used for make snapshot here.

lemon-suse: If just want to stop when failed, we can just set 'fatal=1', 'milestone' is for a snapshot for the current test environment.

jknphy: afir milestone shouldn't be needed. seems that PR need rebase now.

coolgw: @jknphy PR arleady rebase and VR done.

PR #17331, opened 2023-06-29
Close pop-up windows after updates installed in 'updates_packagekit_gpk' test

rfan1:
https://progress.opensuse.org/issues/130468
https://bugzilla.opensuse.org/show_bug.cgi?id=1212680

Close pop-up windows if any after updates installed.

VR: https://openqa.opensuse.org/tests/overview?&build=rfan0703&distri=opensuse

### DIFF
--- a/tests/update/updates_packagekit_gpk.pm
+++ b/tests/update/updates_packagekit_gpk.pm
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2016-2019 SUSE LLC
+# Copyright 2016-2023 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
 # Package: PackageKit gnome-packagekit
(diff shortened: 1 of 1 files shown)


zhaocongjia:
Richard and I have been debugging this problem, and tested the solution. LGTM.

rfan1:
Need to run VRs on both TW and SLE

dzedro:
I'm not 100%, but I guess this failure does happen when authentication is not confirmed with password, there is no video and the code is so bad that it will not stop on authentication pop-up. I will confirm it in https://openqa.opensuse.org/tests/3392831

dzedro (approved):

DimStar77:
https://openqa.opensuse.org/tests/3404576#step/consoletest_finish/11 seems to be a fallout from this change

in updates_packagekit_gpk, the last action is 'pressing alt f4' even though there is no popup to close. on XFCE this triggers a logout dialog - the tests then go straight to text console for more testing and at the point we get back to graphical mode, the logout dialog is still present - but unexpected

PR #17332, opened 2023-06-29
Improve summary of journal_check

clanig:
This PR just improves the summary of what the module journal_check does do.

### DIFF
--- a/tests/console/journal_check.pm
+++ b/tests/console/journal_check.pm
@@ -3,7 +3,9 @@
 # Copyright 2017-2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
-# Summary: Basic journal tests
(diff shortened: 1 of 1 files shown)


tests/console/journal_check.pm
```diff
@@ -3,7 +3,9 @@
 # Copyright 2017-2021 SUSE LLC
 # SPDX-License-Identifier: FSFAP
 
-# Summary: Basic journal tests
+# Summary: Check system journal for errors given a list of known patterns
+#          referring to known bugs. This module will fail in case an unknown
+#          message was found.
```
grisu48, line 8: I would add, that it also fails in case a failing systemd service is detected (see Lines 97 and following)

clanig: You are right. Thanks for pointing that out. I have added it in the newly pushed version.

PR #17341, opened 2023-06-30
[WIP]Add test info for kvm mitigation function

museQi:
Add some info for kvm mitigation function test.
- Related ticket: https://progress.opensuse.org/issues/130841?issue_count=130&issue_position=12&next_issue_id=130667&prev_issue_id=130847
- Verification run:http://10.67.129.7/tests/2234/file/autoinst-log.txt
                               http://10.67.129.7/tests/2287/logfile?filename=autoinst-log.txt

### DIFF
--- a/lib/Mitigation.pm
+++ b/lib/Mitigation.pm
@@ -201,6 +201,7 @@ sub read_msr {
 
 sub vulnerabilities {
     my $self = shift;
+    diag "Check if $self->{'name'} is affected.";
     if ($self->read_cpuid_edx() & $self->CPUID()) {
         if ($self->read_msr() & $self->MSR()) {
             record_info("$self->{'name'} Not Affected", "This machine needn't be tested.");
@@ -246,6 +247,7 @@ sub lscpu {
 #This function will finish testing in default status.
 #As out of box testing. and clean up all mitigations parameters.
 sub check_default_status {
+    diag "Check mitigation default status.";
     my $self = shift;
     assert_script_run('cat /proc/cmdline');
     if (ref($self->{parameter}) ne 'ARRAY') {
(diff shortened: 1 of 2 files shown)


museQi:
@XGWang0 @zyuhu,please help me review.

lib/Mitigation.pm
```diff
@@ -246,6 +247,7 @@ sub lscpu {
 #This function will finish testing in default status.
 #As out of box testing. and clean up all mitigations parameters.
 sub check_default_status {
+    diag "check mitigation default status.";
```
zyuhu, line 250: Since we're adding logs, think we could have the format uniform, like all logs first letter uppercase

lib/Mitigation.pm
```diff
@@ -274,6 +276,7 @@ sub check_default_status {
 #Check cpu flags exist or not.
 #when $cmd is off, the match is inverted.
 sub check_cpu_flags {
+    diag "check the cpu flags.";
```
zyuhu, line 279: same with line 250 comment, below same comment will skip, please check.

lib/Mitigation.pm
```diff
@@ -201,6 +201,7 @@ sub read_msr {
 
 sub vulnerabilities {
     my $self = shift;
+    diag "Check if $self->{'name'} is affected.";
```
zyuhu, line 204: is that more convenient to add function or method name here?

XGWang0:
I think you should add the info including the status, not the pure message to record the code is ran.

museQi:
> I think you should add the info including the status, not the pure message to record the code is ran.

Almost all the functions uses the assert_script_run,So about the status can be seen by this.

PR #17347, opened 2023-07-03
console/valgrind: Call cachegrind with --cache-sim=yes

Vogtinator:
Valgrind 3.21 disables cache-sim by default. We check that its output is present, so just enable it explicitly again.

- Related ticket: https://progress.opensuse.org/issues/129622
- Verification run: http://10.168.4.168/tests/1323

### DIFF
--- a/tests/console/valgrind.pm
+++ b/tests/console/valgrind.pm
@@ -102,7 +102,7 @@ sub run {
     assert_present($output, 'totals: ', "'callgrind totals' mismatch");
 
     # cachegrind tool checks
-    assert_script_run('valgrind --tool=cachegrind --cachegrind-out-file="cachegrind.out" ./valgrind-test');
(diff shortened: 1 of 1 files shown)


tests/console/valgrind.pm
```diff
@@ -102,7 +102,7 @@ sub run {
     assert_present($output, 'totals: ', "'callgrind totals' mismatch");
 
     # cachegrind tool checks
-    assert_script_run('valgrind --tool=cachegrind --cachegrind-out-file="cachegrind.out" ./valgrind-test');
+    assert_script_run('valgrind --tool=cachegrind --cachegrind-out-file="cachegrind.out" --cache-sim=yes ./valgrind-test');
```
DimStar77, line 105: is this compatible with all valgrind versions we are running tests with? (think SLE12 SP Maintenance, if there is any valgrind tests)

Vogtinator: Probably. It's already in a commit from 2009.

PR #17349, opened 2023-07-03
Move SBD delay calculation before hana stop

lpalovsky:
There is and issue in HanaSR secondary crash test where 'crm' command is called 
after the cluster is already down on that node. This cause command and 
eventually test to fail. This PR moves the command before stopping the database.

**- Related ticket:** https://jira.suse.com/browse/TEAM-8033
**- Needles: N/A**
**- Verification run:** 
 - 15SP2 and 15SP3 VRs are not included due to another problem and test does not reach changed code. 
15SP4: https://openqaworker15.qa.suse.cz/tests/197281#
15SP5: https://openqaworker15.qa.suse.cz/tests/197276#

### DIFF
--- a/tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
+++ b/tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
@@ -4,6 +4,10 @@
 # SPDX-License-Identifier: FSFAP
 # Maintainer: QE-SAP <qe-sap@suse.de>
 # Summary: Test module for performing database stop using various methods on secondary HANA database site.
+#
+# Parameters:
+# HA_SBD_START_DELAY (optional) - Sets SBD start delay in /etc/sysconfig/sbd
+# DB_ACTION (optional) - Action to be done on the database to simulate failure - check lib/sles4sap_publiccloud "stop_hana" function
 
 use base 'sles4sap_publiccloud_basetest';
 use strict;
@@ -20,6 +24,7 @@ sub run {
     # $site_b = $instance of secondary instance located in $run_args->{$instances}
     croak('site_b is missing or undefined in run_args') if (!$run_args->{site_b});
     my $site_b = $run_args->{site_b};
+    my $sbd_delay;
     select_serial_terminal;
 
     # Switch to control Site B (currently replica mode)
@@ -35,9 +40,9 @@ sub run {
     my $db_action = get_var('DB_ACTION', $run_args->{hana_test_definitions}{$self->{name}});
     croak('Database action unknown or not defined.') if ($db_action !~ /^(stop|kill|crash)$/);
 
-    if (($db_action eq 'crash') && get_var('HA_SBD_START_DELAY')) {
+    if (($db_action eq 'crash') && defined(get_var('HA_SBD_START_DELAY'))) {
         # Setup sbd delay in case of crash OS to prevent cluster starting too quickly after reboot.
-        $self->setup_sbd_delay(get_var('HA_SBD_START_DELAY'));
+        $self->setup_sbd_delay();
         record_info('Crash DB', "Crashing OS on Site B ('$site_b->{instance_id}')");
     }
     else {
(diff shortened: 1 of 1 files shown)


tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
```diff
@@ -35,22 +36,29 @@ sub run {
     my $db_action = get_var('DB_ACTION', $run_args->{hana_test_definitions}{$self->{name}});
     croak('Database action unknown or not defined.') if ($db_action !~ /^(stop|kill|crash)$/);
 
-    if (($db_action eq 'crash') && get_var('HA_SBD_START_DELAY')) {
+    if (($db_action eq 'crash') && defined(get_var('HA_SBD_START_DELAY'))) {
```
mpagot, line 39: Is `defined` needed here? I can only see one other usage of this pattern in this repo

```
tests/jeos/firstrun.pm:
  
194:     if (is_generalhw && !defined(get_var('GENERAL_HW_VNC_IP'))) {
```

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
```diff
@@ -35,22 +36,29 @@ sub run {
...
         # Setup sbd delay in case of crash OS to prevent cluster starting too quickly after reboot.
-        $self->setup_sbd_delay(get_var('HA_SBD_START_DELAY'));
+        $self->setup_sbd_delay();
```
mpagot, line 41: Can this  crash test works also without `HA_SBD_START_DELAY` ?

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
lpalovsky: Let me find some older runs... the test was considering this as false and would set the SBD delay to 0sec. 
Or let me run the test once again without it, what will happen.

mpagot (approved):
LGTM

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
lpalovsky: Thank you. Very good catch. In case of Public cloud tests this is done by:
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/fc55bfe85aa448966df178ad0928821e5509dae5/lib/sles4sap_publiccloud.pm#L449
You can see it here for example: https://openqaworker15.qa.suse.cz/tests/196654#step/Crash_site_a-primary/24
Delay was set to 120s, normally it is 66s I think.

However on prem tests are missing this. I will create a ticket.

BillAnastasiadis (approved):
LGTM

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
lpalovsky: TEAM-8151

lpalovsky: Correction. THe On prem tests do not yet have feature to set the delay value. I will change the ticket for that. So it's not a bug but a new feature, kind of...

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
alvarocarvajald: `defined` is important here depending on what the module should do when `get_var('HA_SBD_START_DELAY')` returns 0 or the empty string. Both 0 and the empty string will return true if evaluated with `defined` and false in other cases.

I see `setup_sbd_delay()` will skip empty string, but in the case of 0, it will try to set it.

tests/sles4sap/publiccloud/hana_sr_test_secondary.pm
alvarocarvajald: > Correction. THe On prem tests do not yet have feature to set the delay value. I will change the ticket for that. So it's not a bug but a new feature, kind of...

Thanks!

PR #17350, opened 2023-07-04
Update agama_installation to use e2e-agama-playwright

lemon-suse:
Update agama_installation to use own project repo of e2e-agama-playwright and add upload trace to post_fail_hook.

- Related ticket: N/A
- Needles:  N/A
- Verification run:  https://openqa.opensuse.org/tests/3406485#

### DIFF
--- a/tests/yam/agama/agama_installation.pm
+++ b/tests/yam/agama/agama_installation.pm
@@ -23,13 +23,17 @@ sub run {
     $testapi::password = 'linux';
     select_console 'root-console';
 
-    assert_script_run('RUN_INSTALLATION=1 playwright test --trace on --project chromium --config /usr/share/agama-playwright take_screenshots', timeout => 600);
-
+    assert_script_run('RUN_INSTALLATION=1 playwright test --trace on --project chromium --config /usr/share/e2e-agama-playwright take_screenshots', timeout => 600);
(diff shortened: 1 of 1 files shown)


tests/yam/agama/agama_installation.pm
```diff
@@ -23,13 +23,16 @@ sub run {
...
+    assert_script_run('RUN_INSTALLATION=1 playwright test --trace on --project chromium --config /usr/share/e2e-agama-playwright take_screenshots', timeout => 600);
 
-    upload_logs('./test-results/take_screenshots-The-Installer-installs-the-system-chromium/trace.zip');
     assert_script_run('reboot');
     assert_screen('grub2', 120);
     my @tags = ("welcome-to", "login");
     assert_screen(\@tags, 300);
 }
 
+sub post_fail_hook {
```
jknphy, line 34: could you please convert it in a post_run_hook, so we have always at this development phase more info even when passes ?

lemon-suse: It will fail if I move the upload cmd to post_run_hook, https://openqa.opensuse.org/tests/3404625#step/agama_installation/9 

I added the upload trace cmd after the take_screenshots test, then we can get the trace log whether it pass or fail, WDYT?

lemon-suse: VR updated.

PR #17359, opened 2023-07-04
Boottime parameters defaults added in instance-measure code

m-dati:
Description:
   Boottime parameters defaults added in P.C.instance::measure,store code.
   By default now boottime measures collected and saved at instance creation.

References:
- Needles: None.
- Verification run: To Be Done

### DIFF
--- a/lib/publiccloud/instance.pm
+++ b/lib/publiccloud/instance.pm
@@ -580,7 +580,9 @@ Set PUBLIC_CLOUD_PERF_COLLECT true or >0, to activate boottime measurements.
 
 sub measure_boottime() {
     my ($self, $instance, $type) = @_;
-    return 0 unless (get_var('PUBLIC_CLOUD_PERF_COLLECT'));
+    my $data_collect = get_var('PUBLIC_CLOUD_PERF_COLLECT', 1);
+
+    return 0 unless ($data_collect);
 
     my $ret = {
         kernel_release => undef,
@@ -619,22 +621,24 @@ sub measure_boottime() {
 
 Save data collected with measure_boottime in a DB;
 Mainly stored on a remote InfluxDB on a Grafana server.
-
+To activate boottime push, shall be available results and
+  PUBLIC_CLOUD_PERF_PUSH_DATA true/not 0 and
+  _SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN defined
 =cut
 
 sub store_boottime_db() {
     my ($self, $results) = @_;
-    return unless (get_var('PUBLIC_CLOUD_PERF_PUSH_DATA') && $results);
-
-    my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
-    my $token = get_var('_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN');
-    unless ($url && $token) {
-        record_info("WARN", "PUBLIC_CLOUD_PERF_DB_URI or _SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');
-        return 0;
-    }
-
+    my $data_push = get_var('PUBLIC_CLOUD_PERF_PUSH_DATA', 1);
+    my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI', 'http://publiccloud-ng.qa.suse.de:8086');
     my $org = get_var('PUBLIC_CLOUD_PERF_DB_ORG', 'qec');
     my $db = get_var('PUBLIC_CLOUD_PERF_DB', 'perf_2');
+    my $token = get_var('_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN');
+
+    return unless ($results && $data_push);
+    unless ($token) {
+        record_info("WARN", "_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN is missing ", result => 'fail');
+        return 0;
+    }
 
     my $tags = {
         instance_type => get_var('PUBLIC_CLOUD_INSTANCE_TYPE'),
--- a/variables.md
+++ b/variables.md
@@ -304,12 +304,12 @@ PUBLIC_CLOUD_IMG_PROOF_EXCLUDE | string | "" | Tests to be excluded by img-proof
 PUBLIC_CLOUD_INSTANCE_TYPE | string | "" | Specify the instance type. Which instance types exists depends on the CSP. (default-azure: Standard_A2, default-ec2: t2.large )
 PUBLIC_CLOUD_LTP | boolean | false | If set, the run_ltp test module is added to the job.
 PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE | boolean | false | Do not remove the instance when the test fails.
-PUBLIC_CLOUD_PERF_COLLECT | boolean | false | Enable `boottime` measures collection, at end of `create_instance` routine.
-PUBLIC_CLOUD_PERF_DB_URI | string | "" | If set, the bootup times get stored in the influx database. (e.g. PUBLIC_CLOUD_PERF_DB_URI=http://publiccloud-ng.qe.suse.de:8086)
-PUBLIC_CLOUD_PERF_DB | string | "perf_2" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the bucket in which the performance metrics are stored
-PUBLIC_CLOUD_PERF_DB_ORG | string | "qec" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the organization in which the performance metrics are stored
-_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN | string | "" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this required variable is the access token
-PUBLIC_CLOUD_PERF_PUSH_DATA | boolean | "" | If set to `1`, then the test run will push it's metrics to the InfluxDB.
+PUBLIC_CLOUD_PERF_COLLECT | boolean | 1 | To enable `boottime` measures collection, at end of `create_instance` routine.
+PUBLIC_CLOUD_PERF_DB_URI | string | "http://publiccloud-ng.qe.suse.de:8086" | bootup time measures get pushed to this Influx database url.
+PUBLIC_CLOUD_PERF_DB | string | "perf_2" | defines the bucket in which the performance metrics are stored on PUBLIC_CLOUD_PERF_DB_URI
+PUBLIC_CLOUD_PERF_DB_ORG | string | "qec" | defines the organization in which the performance metrics are stored on PUBLIC_CLOUD_PERF_DB_URI
+_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN | string | "" | this required variable is the token to access PUBLIC_CLOUD_PERF_DB_URI (defined in `salt workerconf`)
+PUBLIC_CLOUD_PERF_PUSH_DATA | boolean | 1 | To enable the test to push it's metrics to the InfluxDB, when PUBLIC_CLOUD_PERF_COLLECT true.
 PUBLIC_CLOUD_PERF_THRESH_CHECK | boolean | "" | If set to `1` or any not empty value, then the test run will _also_ execute the thresholds check on the collected metrics. By _default_ that check is _Not executed_.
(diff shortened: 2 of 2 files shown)


lib/publiccloud/instance.pm
```diff
@@ -580,7 +580,9 @@ Set PUBLIC_CLOUD_PERF_COLLECT true or >0, to activate boottime measurements.
 
 sub measure_boottime() {
     my ($self, $instance, $type) = @_;
-    return 0 unless (get_var('PUBLIC_CLOUD_PERF_COLLECT'));
+    my $data_collect = get_var('PUBLIC_CLOUD_PERF_COLLECT', 1);
```
grisu48, line 583:
```suggestion
    return 0 unless (get_var('PUBLIC_CLOUD_PERF_COLLECT', 1));
```

It's a style question but I'd argue this is trivial enough that we don't need an additional variable.

lib/publiccloud/instance.pm
```diff
@@ -619,23 +621,25 @@ sub measure_boottime() {
...
+  PUBLIC_CLOUD_PERF_PUSH_DATA true/not 0 and
+  _SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN and PUBLIC_CLOUD_PERF_DB_URI defined
 =cut
 
 sub store_boottime_db() {
     my ($self, $results) = @_;
-    return unless (get_var('PUBLIC_CLOUD_PERF_PUSH_DATA') && $results);
-
-    my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI');
+    my $data_push = get_var('PUBLIC_CLOUD_PERF_PUSH_DATA', 1);
```
grisu48, line 631: I'd say the old style was better because it immediately returns if not enabled.

```suggestion
    return unless (get_var('PUBLIC_CLOUD_PERF_PUSH_DATA', 1) && $results);
```

Plus, you could replace the next following `get_var` calls with `get_required_var` - if the data push is enabled, the performance database variables should be treated as required variables IMHO.

variables.md
```diff
@@ -304,12 +304,12 @@ PUBLIC_CLOUD_IMG_PROOF_EXCLUDE | string | "" | Tests to be excluded by img-proof
 PUBLIC_CLOUD_INSTANCE_TYPE | string | "" | Specify the instance type. Which instance types exists depends on the CSP. (default-azure: Standard_A2, default-ec2: t2.large )
 PUBLIC_CLOUD_LTP | boolean | false | If set, the run_ltp test module is added to the job.
 PUBLIC_CLOUD_NO_CLEANUP_ON_FAILURE | boolean | false | Do not remove the instance when the test fails.
-PUBLIC_CLOUD_PERF_COLLECT | boolean | false | Enable `boottime` measures collection, at end of `create_instance` routine.
-PUBLIC_CLOUD_PERF_DB_URI | string | "" | If set, the bootup times get stored in the influx database. (e.g. PUBLIC_CLOUD_PERF_DB_URI=http://publiccloud-ng.qe.suse.de:8086)
+PUBLIC_CLOUD_PERF_COLLECT | boolean | 1 | 0:disable, 1:enable `boottime` measures collection, at end of `create_instance` routine.
```
grisu48, line 307:
```suggestion
PUBLIC_CLOUD_PERF_COLLECT | boolean | 1 | enables `boottime` measurement collection, at end of `create_instance` routine.
```

IMHO this is unnecessary verbosity

variables.md
```diff
@@ -304,12 +304,12 @@ PUBLIC_CLOUD_IMG_PROOF_EXCLUDE | string | "" | Tests to be excluded by img-proof
...
+PUBLIC_CLOUD_PERF_DB_URI | string | "http://publiccloud-ng.qe.suse.de:8086" | If set, the bootup times get stored in the influx database. (e.g. PUBLIC_CLOUD_PERF_DB_URI=http://publiccloud-ng.qe.suse.de:8086)
 PUBLIC_CLOUD_PERF_DB | string | "perf_2" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the bucket in which the performance metrics are stored
 PUBLIC_CLOUD_PERF_DB_ORG | string | "qec" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this variable defines the organization in which the performance metrics are stored
 _SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN | string | "" | If `PUBLIC_CLOUD_PERF_DB_URI` is defined, this required variable is the access token
-PUBLIC_CLOUD_PERF_PUSH_DATA | boolean | "" | If set to `1`, then the test run will push it's metrics to the InfluxDB.
+PUBLIC_CLOUD_PERF_PUSH_DATA | boolean | 1 | 0:disable, 1:enable the test to push it's metrics to the InfluxDB, when PUBLIC_CLOUD_PERF_COLLECT true.
```
grisu48, line 312:
```suggestion
PUBLIC_CLOUD_PERF_PUSH_DATA | boolean | 1 | enable the test to push it's metrics to the InfluxDB, when PUBLIC_CLOUD_PERF_COLLECT true.
```

IMHO this is unnecessary verbosity

lib/publiccloud/instance.pm
m-dati: I expected this effect, but I preferred to leave more visible the parameter handling by separating the value assignment from usage, helpful in case of explicit FALSE/0 assignments to disable.

lib/publiccloud/instance.pm
m-dati: same comment as above for data collect

variables.md
m-dati: ehm if someone prefers to use `empty` or undefined '',  willing to disable all, it will not happen obtaining default=1. This verbosity I considered necessary to avoid future misunderstanding

lib/publiccloud/instance.pm
asmorodskyi: > It's a style question but I'd argue this is trivial enough that we don't need an additional variable.
by removing variable you removing ability to disable this flow . While I want it to be default behavior I still would like to have ability to switch it off in some certain cases

lib/publiccloud/instance.pm
asmorodskyi: > Plus, you could replace the next following get_var calls with get_required_var - if the data push is enabled, the performance database variables should be treated as required variables IMHO.

the whole point of this PR is to remove need of defining this variables elsewhere . Maurizio did everything as expected

lib/publiccloud/instance.pm
```diff
@@ -619,23 +621,25 @@ sub measure_boottime() {
...
+    my $url = get_var('PUBLIC_CLOUD_PERF_DB_URI', 'http://publiccloud-ng.qa.suse.de:8086');
+    my $org = get_var('PUBLIC_CLOUD_PERF_DB_ORG', 'qec');
+    my $db = get_var('PUBLIC_CLOUD_PERF_DB', 'perf_2');
     my $token = get_var('_SECRET_PUBLIC_CLOUD_PERF_DB_TOKEN');
+
+    return unless ($results && $data_push);
     unless ($url && $token) {
```
asmorodskyi, line 638: this unless is not relevant anymore

variables.md
m-dati: As above, I here suggested to explicitly assign zero when we want to disable, avoiding use of `undefined`.

lib/publiccloud/instance.pm
m-dati: Sorry, what of the 2?
the first one still is needed, when results not produced for any error or when we wanted to disable measures

lib/publiccloud/instance.pm
m-dati: get_required_var would cause test to die, unwanted here

asmorodskyi: and also `get_required_var` not allow to provide default value which major idea of this PR

variables.md
asmorodskyi: and here I will agree with Felix , this is something obvious or should/can be added to ALL variables in this list

lib/publiccloud/instance.pm
asmorodskyi: ok , so I was wrong we can not and should not fully remove this `unless` .

`$token` is still may be absent . And actually this is how we are handling clone of job into none-production openQA instance . So let's keep unless

BUT we should drop `$url` from it ( from now on it will be ALWAYS defined )

m-dati: Yes, me too, I was thinking that url check no more needed.

lib/publiccloud/instance.pm
grisu48: I don't understand, but as this is just a style suggestion, just ignore it.

lib/publiccloud/instance.pm
grisu48: Also here: I don't understand, but as this is just a style suggestion, just ignore it.

variables.md
m-dati: done

variables.md
m-dati: done

lib/publiccloud/instance.pm
asmorodskyi: arrrgh ! :roll_eyes:  :see_no_evil:  :speak_no_evil: . and only NOW I realized what Felix was suggesting :D :D :D Felix the problem that we have TWO variables here one is perl variable and another is openQA variable . Maybe it was obvious to you that you referring to perl one but at least me get that you asking to remove **openQA one**  and of course I totally against this . But about removing **perl variable** it make sense yeah +1 from my side . 

Not sure what was Maurizio point :thinking:  Maurizio did you get into same trap as I do ?

grisu48: Please don't bother. This is just a style suggestion, not worth discussion it for more than 10 seconds.

Just leave it as it is (or change it), it is just a suggestion.

m-dati:
Prepared some VRs, Boottime measures behaved as expected by defaults or settings:

- cloned (https://openqa.suse.de/tests/11506033) 
default collect + push db
sle-15-SP5-EC2-x86_64-Build0364-publiccloud_consoletests @ec2_t2.large - 
https://openqa.suse.de/tests/11512912  
  - PASS boottime collected and pushed

- cloned (https://openqa.suse.de/tests/11505984) 
set explicit collect & do-Not-push db:
PUBLIC_CLOUD_PERF_COLLECT=1 PUBLIC_CLOUD_PERF_PUSH_DATA=0 
sle-15-SP5-GCE-x86_64-Build0324-publiccloud_smoketests @gce_n2d_standard_2 - 
https://openqa.suse.de/tests/11512914  
  - PASS boottime collected but not pushed

- cloned (https://openqa.suse.de/tests/11508989) 
set empty val to collect expecting assumend 1 & do-Not-push db:
PUBLIC_CLOUD_PERF_COLLECT='' PUBLIC_CLOUD_PERF_PUSH_DATA=0
sle-15-SP4-GCE-Updates-x86_64-Build20230704-1-publiccloud_containers - 
https://openqa.suse.de/tests/11512915
  - PASS boottime collected but not pushed

m-dati:
Note please, that with defaults enabling by code, also all the `P.C. Latest <CSP>` test groups Azure [EC2](https://openqa.suse.de/group_overview/274) [GCE](https://openqa.suse.de/group_overview/275) all tests will start to collect and push data to db, against planned to start with 15-SP6, therefore those test groups yml setting shall be tweaked to set `PUBLIC_CLOUD_PERF_COLLECT=0`.

lib/publiccloud/instance.pm
m-dati: I got that it was a style preference and sorry for keeping my position because this is my style, also because with or without variable in case of exit the timing is not so different, but 'with' it, the code legibility is improved and is less spagetti-style :-)

asmorodskyi:
> Note please, that with defaults enabling by code, also all the `P.C. Latest <CSP>` test groups Azure [EC2](https://openqa.suse.de/group_overview/274) [GCE](https://openqa.suse.de/group_overview/275) all tests will start to collect and push data to db, against planned to [start with 15-SP6](https://progress.opensuse.org/issues/129328), therefore those test groups yml setting shall be tweaked to set `PUBLIC_CLOUD_PERF_COLLECT=0`.

let's just leave it as is . If it is fail it does not matter we don't care about this job groups till start of development of SP6

m-dati:
Thank you for merging!
About `Latest` groups, no failure should occur for this change, only more data pushed, but it's all ok. :-)

PR #17366, opened 2023-07-06
Reduce needle to match grub2 on aarch64 for agama

lemon-suse:
Sometimes it can't match grub2 in a short time on aarch64 for agama, we reduce needle to fix it.

- Related ticket: N/A
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/overview?arch=aarch64&flavor=&machine=&test=&modules=&module_re=&distri=alp&build=lemon-suse%2Fos-autoinst-distri-opensuse%23debug-agama-auto-install-issue&version=agama-2.1-staging#

### DIFF
--- a/tests/installation/bootloader_uefi.pm
+++ b/tests/installation/bootloader_uefi.pm
@@ -91,7 +91,14 @@ sub run {
     # By pressing a random key, we stop the grub timeout
     send_key 'backspace' if (is_jeos && is_aarch64);
 
-    assert_screen([qw(bootloader-shim-import-prompt bootloader-grub2)], $bootloader_timeout);
+    if (get_var('VERSION') =~ /agama/) {
+        # For agama test, it is too short time to match the grub2(10s), so we create
+        # a new needle to avoid too much needles loaded.
+        assert_screen("bootloader-grub2-agama", $bootloader_timeout);
+    }
+    else {
+        assert_screen([qw(bootloader-shim-import-prompt bootloader-grub2)], $bootloader_timeout);
+    }
     if (match_has_tag("bootloader-shim-import-prompt")) {
         send_key "down";
         send_key "ret";
@@ -117,7 +124,7 @@ sub run {
         # random magic numbers
         send_key_until_needlematch('inst-onupgrade', 'down', 11, 3);
     }
-    else {
+    elsif (get_var('VERSION') !~ /agama/) {
         if (get_var("PROMO") || get_var('LIVETEST') || get_var('LIVECD')) {
             send_key_until_needlematch("boot-live-" . get_var("DESKTOP"), 'down', 11, 3);
         }


tests/installation/bootloader_uefi.pm
```diff
@@ -117,7 +119,7 @@ sub run {
         # random magic numbers
         send_key_until_needlematch('inst-onupgrade', 'down', 11, 3);
     }
-    else {
+    elsif (get_var('VERSION') !~ /agama/) {
```
jknphy, line 122: you might break the logic for another livecd, like tumbleweed, not sure why you need this change...

jknphy:
> Sometimes it can't match grub2 in a short time on aarch64 for agama, we reduce needle to fix it.
> 
> * Related ticket: N/A
> * Needles: N/A
> * Verification run: https://openqa.opensuse.org/tests/3409480#step/bootloader_start/2

we should run at least 10 VR to be sure this work (just in case we need special needle for agama and exclude current to possible needle matches, not just one like you did).

tests/installation/bootloader_uefi.pm
lemon-suse: Sometimes it can't match the boot-live-gnome in short time, https://openqa.opensuse.org/tests/3406032#step/bootloader_start/3  I add condition to avoid it to match boot-live-gnome.

lemon-suse:
> > Sometimes it can't match grub2 in a short time on aarch64 for agama, we reduce needle to fix it.
> > 
> > * Related ticket: N/A
> > * Needles: N/A
> > * Verification run: https://openqa.opensuse.org/tests/3409480#step/bootloader_start/2
> 
> we should run at least 10 VR to be sure this work (just in case we need special needle for agama and exclude current to possible needle matches, not just one like you did).

I have run it many times, but haven't added to PR. I will run 10 times more for it.

jknphy:
looks like it doesn't always work, https://openqa.opensuse.org/tests/3411362#step/bootloader_start/2
still there are 8 seconds and the screen is blocked, we might need to try to use a new needle there, instead of excluding one of those two, meaning, exclude both and use new one. make sense?

lemon-suse:
> looks like it doesn't always work, https://openqa.opensuse.org/tests/3411362#step/bootloader_start/2
> 
> ```
> [2023-07-07T06:31:05.633488+01:00] [warn] [pid:5742] !!! backend::baseclass::check_asserted_screen: check_asserted_screen took 1.05 seconds for 42 candidate needles - make your needles more specific
> [2023-07-07T06:31:05.633751+01:00] [debug] [pid:5742] no match: 179.0s, best candidate: bootloader-grub2-leap-15.5-timeout-uefi-20230219 (0.24)
> ...
> ```
> 
> still there are 8 seconds and the screen is blocked, we might need to try to use a new needle there, instead of excluding one of those two, meaning, exclude both and use new one. make sense?

Updated.

tests/installation/bootloader_uefi.pm
```diff
@@ -91,7 +91,15 @@ sub run {
     # By pressing a random key, we stop the grub timeout
     send_key 'backspace' if (is_jeos && is_aarch64);
 
-    assert_screen([qw(bootloader-shim-import-prompt bootloader-grub2)], $bootloader_timeout);
+    my @needles = qw(bootloader-grub2 bootloader-shim-import-prompt);
```
jknphy, line 94: That is not good code style, to remove things that you don't need to add in the first place. Those two command grep do not compensate to avoid the `else` statement. what do you think about this?

```
    if (get_var('VERSION') =~ /agama/) {
     # comment
      assert_screen(\@needles, 'bootloader-grub2-agama');
    }
    else {
      assert_screen([qw(bootloader-shim-import-prompt bootloader-grub2)], $bootloader_timeout);
    }
```

jknphy (approved):

tests/installation/bootloader_uefi.pm
lemon-suse: Updated.

PR #17368, opened 2023-07-06
xfstests: Support ocfs2 test

SeroSun:
Init ocfs2 test support. The plan is in related ticket. This is first step to test ocfs2 in 1 node without cluster setting. 

- Related ticket: https://progress.opensuse.org/issues/128918
- Verification run: 
- test ocfs2/001: http://10.67.133.133/tests/241
- test generic/001-100: http://10.67.133.133/tests/245#step/generate_report/423
- (the fail is expected, to check if new code would solve hung issue, fail tests and skip tests)

### DIFF
--- a/lib/filesystem_utils.pm
+++ b/lib/filesystem_utils.pm
@@ -275,12 +275,24 @@ sub format_partition {
     if ($filesystem =~ /ext4/) {
         $options = $args{options} // "-F";
     }
+    elsif ($filesystem =~ /ocfs2/) {
+        $options = $args{options} // "-F --fs-features=local --fs-feature-level=max-features";
+    }
     else {
         $options = $args{options} // "-f";
     }
     script_run("umount -f $part");
     sleep 1;
-    assert_script_run("mkfs.$filesystem $options $part");
+    if ($filesystem =~ /ocfs2/) {
+        # mkfs.ocfs2 will still require input y even you used -F
+        background_script_run("mkfs.$filesystem $options $part");
+        sleep 1;
+        script_run('y');
+        wait_still_screen(10, 60);
+    }
+    else {
+        assert_script_run("mkfs.$filesystem $options $part");
+    }
 }
 
 =head2 df_command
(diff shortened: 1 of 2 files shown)


lansuse:
LGTM

lib/filesystem_utils.pm
```diff
@@ -275,12 +275,21 @@ sub format_partition {
...
+    }
     else {
         $options = $args{options} // "-f";
     }
     script_run("umount -f $part");
     sleep 1;
-    assert_script_run("mkfs.$filesystem $options $part");
+    if ($filesystem =~ /ocfs2/) {
+        # mkfs.ocfs2 will still require input y even you used -F
+        assert_script_run("y | mkfs.$filesystem $options $part");
```
lansuse, line 288: I guess there should be `echo y | mkfs.$filesystem $options $part`,
or else
`yes | mkfs.$filesystem $options $part`
But IMO `yes` is better, because it output 'y' repeatedly.

lansuse (changes requested):

lib/filesystem_utils.pm
SeroSun: The input only requires y here,`mkfs.ocfs2 -F --fs-features=local --fs-feature-level=max-features /home/test_dev ;
mkfs.ocfs2 1.8.7
Overwriting existing ocfs2 partition.
Proceed (y/N): `

SeroSun: change to echo is better, I'll change this. It will solve the problem like this http://10.67.133.133/tests/249#step/run/478

SeroSun: I tried with echo and script_run can't solve this problem in http://10.67.133.133/tests/249#step/run/478. I'll revert this to the previous background_script_run and type y solution.

schlad (changes requested):
Could you amend the git commit, so it will explain what is ocfs2

SeroSun:
> Could you amend the git commit, so it will explain what is ocfs2

Done

schlad (approved):

PR #17370, opened 2023-07-06
[WIP] kselftests support

acerv:
kselftests support for openQA permits to run kernel kselftests, using kirk as the main executor.

- Related ticket: https://progress.opensuse.org/issues/130781
- Related ticket: https://progress.opensuse.org/issues/130784

### DIFF
--- a/tests/kernel/bpftrace.pm
+++ b/tests/kernel/bpftrace.pm
@@ -71,17 +71,11 @@ sub run {
       old/tcpdrop.bt);
 
     foreach my $t (@assert_tests) {
-        my $ret = script_run("timeout --preserve-status -s SIGINT --kill-after=10 10 bpftrace $tools_dir/$t");
-
-        if ($ret == 130) {
-            record_info('timeout', "'bpftrace $t' did not handle SIGINT; system was probably too slow to attach probes");
-        } elsif ($ret) {
-            die "'bpftrace $t' failed";
-        }
+        assert_script_run(qq%echo -e "\\ninterval:s:5 { exit(); }" | cat $tools_dir/$t - | bpftrace -%);
     }
 
     foreach my $t (@tests) {
-        script_run("timeout --preserve-status -s SIGINT --kill-after=10 10 bpftrace $tools_dir/$t");
+        script_run(qq%echo -e "\\ninterval:s:5 { exit(); }" | cat $tools_dir/$t - | bpftrace -%);
     }
 
     my $case_dir = get_required_var('CASEDIR');


tests/kernel/kselftests.pm
```diff
@@ -0,0 +1,70 @@
...
+    # download linux source code
+    zypper_call("in -y git");
+    assert_script_run("git clone -q --single-branch -b $branch --depth 1 $repo");
+
+    # install build tools and compile tests
+    my $suite = get_required_var('KSELFTESTS_SUITE');
+    my $root = "/root/linux/tools/testing/selftests";
+
+    zypper_call("in -t pattern devel_basis");
+    assert_script_run("make -C $root/$suite", timeout => 1800);
```
richiejp, line 34: Everything up to this point is likely unusable for the BPF tests at least. We should use the package (when it is available) which doesn't need a generic module, we can just put it in kselftests_bpf_install.pm or whatever. It's likely that a bunch of quirks will need to be handled during install.

tests/kernel/kselftests.pm
```diff
@@ -0,0 +1,70 @@
...
+            result => 'softfail'
+        );
+
+        set_var('KIRK_SKIP', "$test_exclude");
+    }
+
+    # setup kirk framework before calling it
+    set_var('KIRK_FRAMEWORK', "kselftests:root=$root");
+    set_var('KIRK_SUITE', "$suite");
+}
```
richiejp, line 68: Looks good, possibly this could be a generic module or library function (called from the specific install module) just to populate the skip list.

schedule/kernel/kselftests.yaml
```diff
@@ -0,0 +1,8 @@
+name:          kselftests
+description:    >
+    Generic scheduler for Linux kselftests
+schedule:
+    - boot/boot_to_desktop
+    - kernel/kselftests
+    - kernel/kirk
+    - shutdown/shutdown
```
richiejp, line 8: I don't think either of these yaml files are needed, unless the kirk one is for testing kirk itself?

tests/kernel/kirk.pm
```diff
@@ -0,0 +1,70 @@
...
+use strict;
+use testapi;
+use serial_terminal 'select_serial_terminal';
+
+our $result_file = 'result.json';
+
+sub run
+{
+    my ($self) = @_;
+    my $repo = get_var('KIRK_REPO', 'https://github.com/acerv/kirk.git');
```
richiejp, line 21: In general this looks good. Kirk should really be in a package, but I would not volunteer to do that.

tests/kernel/kselftests.pm
acerv: Yes, now code is cgroups kselftests specific.

tests/kernel/kselftests.pm
acerv: I thought about that, but never dig into it. There's a risk we are going to use kirk for multiple modules and for each one of them we are going to define KIRK_SOMETHING

schedule/kernel/kselftests.yaml
acerv: We never know :)

tests/kernel/kselftests.pm
acerv: Also, the idea is that once we have a `zypper in` we can get rid of all modules for specific kselftests installations and to use a unique variables telling what's the suite we want to install.

richiejp:
In summary the core kirk module looks solid, so I'd like that to be merged and the rest can be dropped for now. This is just for convenience so that the bpf tests and anything else being based off of kirk can spin-off from master instead of some other dev branch.

acerv:
> In summary the core kirk module looks solid, so I'd like that to be merged and the rest can be dropped for now. This is just for convenience so that the bpf tests and anything else being based off of kirk can spin-off from master instead of some other dev branch.

I can create a PR for that then

richiejp:
> I can create a PR for that then

Thanks! https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17374

PR #17372, opened 2023-07-06
wicked: add regression test for boo#1212806

cfconrad:
- Verification run: 
  - http://dante.wicked.suse.de/t160 (Tumbleweed: show error case)
  - http://dante.wicked.suse.de/t158 (Tumbleweed: with fix)
  - http://dante.wicked.suse.de/t215 (LEAP 15.5)
  - http://dante.wicked.suse.de/t214 (SLE-15-SP5)
  - http://dante.wicked.suse.de/t211 (SLE-12-SP5)

### DIFF
--- a/lib/wickedbase.pm
+++ b/lib/wickedbase.pm
@@ -535,7 +535,12 @@ Used to syncronize the wicked tests for SUT and REF creating the corresponding m
 sub do_barrier {
     my ($self, $type) = @_;
     my $barrier_name = 'test_' . $self->{name} . '_' . $type;
-    barrier_wait($barrier_name);
+    barrier_wait({name => $barrier_name, check_dead_job => 1});
+
+    # This is to mitigate the problem, that if a parallel job is running in the
+    # barrier_wait() poll loop, while this job finished. This would lead to a
+    # failure on the other side.
+    $self->{last_barrier_wait_call} = time;
 }
 
 =head2 setup_vlan
@@ -984,6 +989,17 @@ sub post_run {
     $self->check_logs() unless $self->{skip_check_logs_on_post_run};
     $self->check_coredump();
     $self->upload_wicked_logs('post');
+
+    if (check_var('IS_WICKED_REF')) {
+        my $time_since_barrier_wait = time - ($self->{last_barrier_wait_call} // 0);
+        if ($time_since_barrier_wait < lockapi::POLL_INTERVAL) {
+            my $seconds = lockapi::POLL_INTERVAL - $time_since_barrier_wait;
+            #see https://github.com/os-autoinst/os-autoinst/issues/2340
+            bmwqemu::diag("If the parallel job might wait in barrier_wait() poll loop," .
+                  " we should not finish this parent job to early! sleep $seconds seconds");
+            sleep $seconds;
+        }
+    }
 }
 
 sub pre_run_hook {
--- a/tests/wicked/basic/sut/t10_peerip_and_static_route.pm
+++ b/tests/wicked/basic/sut/t10_peerip_and_static_route.pm
@@ -1,0 +1,40 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Package: wicked wicked-service
+# Summary: Regression test for boo#1212806 (Default route is missing
+#          when gateway has a host route)
+#
+# Maintainer: Clemens Famulla-Conrad <cfamullaconrad@suse.de>
+
+
+use Mojo::Base 'wickedbase';
+use testapi;
+
+sub run {
+    my ($self, $ctx) = @_;
+
+    my $ip = $self->get_ip(type => 'host');
+    my $ifc = $ctx->iface();
+    $self->write_cfg('/etc/sysconfig/network/ifcfg-' . $ifc, <<EOT);
+BOOTPROTO='static'
+STARTMODE='auto'
+IPADDR='136.243.128.214'
+REMOTE_IPADDR='192.168.128.1'
+EOT
+    $self->write_cfg('/etc/sysconfig/network/ifroute-' . $ifc, <<EOT);
+default 192.168.128.1 - $ifc
+EOT
+
+    $self->wicked_command('ifup', $ifc);
+    $self->assert_wicked_state(iface => $ifc);
+    validate_script_output("ip -4 route show", sub { m/default via 192\.168\.128\.1 dev $ifc/ });
+}
+
+sub test_flags {
+    return {always_rollback => 1};
+}
+
+1;


lib/wickedbase.pm
```diff
@@ -535,7 +535,8 @@ Used to syncronize the wicked tests for SUT and REF creating the corresponding m
 sub do_barrier {
     my ($self, $type) = @_;
     my $barrier_name = 'test_' . $self->{name} . '_' . $type;
-    barrier_wait($barrier_name);
+    barrier_wait({name => $barrier_name, check_dead_job => 1});
+    $self->{fix_barrier_wait} = time;
```
asmorodskyi, line 539:
```suggestion
    $self->{barrier_wait_start_at} = time;
```

lib/wickedbase.pm
```diff
@@ -984,6 +985,12 @@ sub post_run {
     $self->check_logs() unless $self->{skip_check_logs_on_post_run};
     $self->check_coredump();
     $self->upload_wicked_logs('post');
+
+    my $time_since_barrier_wait = time - ($self->{fix_barrier_wait} // 0);
+    if ($time_since_barrier_wait < lockapi::POLL_INTERVAL) {
+        bmwqemu::diag("If the other parallel JOB might wait in barrier_wait(), we should not quit this job to early!");
```
asmorodskyi, line 991:
```suggestion
        bmwqemu::diag("Waiting for parallel job to finish, so barrier will be properly released");
```

lib/wickedbase.pm
```diff
@@ -984,6 +985,12 @@ sub post_run {
     $self->check_logs() unless $self->{skip_check_logs_on_post_run};
     $self->check_coredump();
     $self->upload_wicked_logs('post');
+
+    my $time_since_barrier_wait = time - ($self->{fix_barrier_wait} // 0);
```
asmorodskyi, line 989: first of all I am not fully understanding what is going on here , so please explain and I think it will make sense to put at least part of your explanation here so future us or someone who will have luck to support this code after us will have some hints

tests/wicked/basic/sut/t10_peerip_and_static_route.pm
```diff
@@ -0,0 +1,40 @@
...
+    $self->wicked_command('ifup', $ifc);
+    $self->assert_wicked_state(iface => $ifc);
+    validate_script_output("ip -4 route show", sub { m/default via 192\.168\.128\.1 dev $ifc/ });
+}
+
+sub test_flags {
+    return {always_rollback => 1};
+}
+
+1;
```
asmorodskyi, line 40: sometimes I think about wicked tests "what a mess " . But when I see something like this I realizing how awesome they are at the same time :) 

BEAUTIFUL

asmorodskyi:
to summarize my review - all my concerns goes into documentation . Code is perfect

tests/wicked/basic/sut/t10_peerip_and_static_route.pm
cfconrad: Yes and even the this, by quite complex setup.

What I like on this test is, that you see from the testmodule perl file the config directly, without looking into some data folder and do string replacements...

lib/wickedbase.pm
cfconrad: ack

lib/wickedbase.pm
cfconrad: I need to explain this `fix_barrier_wait`, because what you wrote sounds more like the opposite of what I tried to achieve.

This is the situation. 
```
┌──────────────────────────────────────────────────────────────────────────────────────────────────┐
│  GOOD CASE                                                                                       │
│                                                                                                  │
│           Child                                            Parent                                │
│              │                                                 │                                 │
│              │                                                 │                                 │
│              ├────────────►barrier_wait(a)                     │                                 │
│              │                      │                          │                                 │
│              │                      │                          │                                 │
│              │                      │                          │                                 │
│              │                      │                          │                                 │
│              │          api::check_barrier(a)                  │                                 │
│              │            -> not released (409)                ├───────────────────┐             │
│              │                      │                          │                   ▼             │
│              │                      │                          │         Parent::barrier_wait(a) │
│              │                sleep 5                          │                   ▼             │
│              │                      │                          │         api::check_barrier(a)   │
│              │                      │                          │           -> successful   (200) │
│              │                      │                          │                   │             │
│              │                      │                          │◄──────────────────┘             │
│              │                      │                          ▼                                 │
│              │                      │                    do_some_more_tests()                    │
│              │                      │                          │                                 │
│              │          api::check_barrier(a)                  │                                 │
│              │            -> Successful (200)                  │                                 │
│              │                      │                          │                                 │
│              │◄─────────────────────┘                          │                                 │
│              ▼                                                 ▼                                 │
│        Job finished successful                           Job Finished                            │
└──────────────────────────────────────────────────────────────────────────────────────────────────┘
                                                      fix_barrier_wait




┌──────────────────────────────────────────────────────────────────────────────────────────────────┐
│   BAD CASE                                                                                       │
│                                                                                                  │
│           Child                                            Parent                                │
│              │                                                 │                                 │
│              │                                                 │                                 │
│              ├────────────►barrier_wait(a)                     │                                 │
│              │                      │                          │                                 │
│              │                      │                          │                                 │
│              │           api::check_barrier(a)                 ├───────────────────┐             │
│              │             -> not released (409)               │                   │             │
│              │                      │                          │                   ▼             │
│              │                      │                          │            fix_barrier_wait     │
│              │                sleep 5                          │         Parent::barrier_wait(a) │
│              │                      │                          │                   ▼             │
│              │                      │                          │         api::check_barrier(a)   │
│              │                      │                          │           -> successful   (200) │
│              │                      │                          │                   │             │
│              │                      │                          │◄──────────────────┘             │
│              │                      │                          │                                 │
│              │                      │                          ▼                                 │
│              │          api::check_barrier(a)            Job Finished                            │
│              │            -> Lock-Owner-Finished (410)                                           │
│              │               -> die()                                                            │
│              │                     │                                                             │
│              │                     │                                                             │
│              │◄────────────────────┘                                                             │
│              ▼                                                                                   │
│        Job failed unexpected                                                                     │
└──────────────────────────────────────────────────────────────────────────────────────────────────┘
```
## Explanation of `BAD CASE`:
The client job call `barrier_wait()`. In barrier wait is a loop which is calling the `api::check_barrier` call (url is `barrier/$name`). The return value could be

* 200 => barrier released
* 409 => you need to wait - not all expected jobs are here
* 410 => something unexpected happen - cancel with error

The first call of `api::check_barrier()` return 409, because the Parent didn't called barrier_wait() till now.
Then Parent is calling `barrier_wait()` and get's 200. This was the last call in the Parent job, so the Parent job immediately finished.
By default Child process poll the state of barrier every 5 seconds. With this, the next call of `api::check_barrier()` get's 409, because the parent process already finished. which lead to an unexpected die().

=> solution, simply finish not 5seconds after the last barrier_wait() call.

cfconrad: I raised this issue also here: https://github.com/os-autoinst/os-autoinst/issues/2340

lib/wickedbase.pm
```diff
@@ -984,6 +989,14 @@ sub post_run {
     $self->check_logs() unless $self->{skip_check_logs_on_post_run};
     $self->check_coredump();
     $self->upload_wicked_logs('post');
+
+    my $time_since_barrier_wait = time - ($self->{last_barrier_wait_call} // 0);
```
asmorodskyi, line 993: this code is relevant ONLY for parent job , right ? Any chance that with current implementation child will also execute it ?

cfconrad: we can simply guard it with `IS_WICKED_REF == 1`

mloviska:
Hola guys, can you please have a look on https://openqa.suse.de/tests/11609292#step/t10_peerip_and_static_route/98

asmorodskyi:
> Hola guys, can you please have a look on https://openqa.suse.de/tests/11609292#step/t10_peerip_and_static_route/98

rollback was missing https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17441

PR #17375, opened 2023-07-06
Fix failure in repetitive upload_image calls

grisu48:
Adds the architecture to the image definition name to allow x86_64 and aarch64 images in parallel.
Also prevent a failure due to multiple image definitions, if the given image is already defined (see related failure above)

- Related failure: https://openqa.suse.de/tests/11610662
- Related ticket: https://progress.opensuse.org/issues/133013
- Verification runs:
* [x86_64 (new image)](https://duck-norris.qe.suse.de/tests/13363#step/upload_image/97) | [x86_64 (reusing existing image)](https://duck-norris.qe.suse.de/tests/13365#step/upload_image/97)
* [aarch64 (new image)](https://duck-norris.qe.suse.de/tests/13364#step/upload_image/97) | [aarch64 (using existing image)](https://duck-norris.qe.suse.de/tests/13366#step/upload_image/97)
* [slem-basic](https://duck-norris.qe.suse.de/tests/13367) | [slem_containers](https://duck-norris.qe.suse.de/tests/13369)

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -23,6 +23,11 @@ has container => 'sle-images';
 has image_gallery => 'test_image_gallery';
 has lease_id => undef;
 
+# Default SKU/HyperV generation, if not explicitly defined.
+# Currently too many places still rely on gen1 as default, this should be changed soon (poo#133079)
+# This trigger is a temporary solution to make this transition easier and can be removed when poo#133079 is being resolved.
+my $default_sku = 'gen1';
+
 sub init {
     my ($self) = @_;
     $self->SUPER::init();
@@ -59,6 +64,24 @@ sub get_image_id {
     return $self->SUPER::get_image_id($img_url);
 }
 
+=head2 generate_azure_image_definition
+
+    my $definition = $self->generate_azure_image_definition();
+
+Generated the Azure Image name from the job settings. If present, it takes the PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION setting.
+If not present it generated the image definition name based on distri, version, flavor and SKU.
+
+Note: Image definitions needs to be distinct names and can only serve one architecture!
+
+Example: 'SLE-MICRO-5.4-BYOS-AZURE-X86_64-GEN1'
+=cut
+
+sub generate_azure_image_definition {
+    my $arch = (get_var('PUBLIC_CLOUD_ARCH', get_required_var('ARCH'))) eq 'x86_64' ? 'x64' : 'Arm64';
+    my $definition = get_required_var('DISTRI') . '-' . get_required_var('VERSION') . '-' . get_required_var('FLAVOR') . '-' . $arch . '-' . get_var('PUBLIC_CLOUD_AZURE_SKU', $default_sku);
+    return get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+}
+
 sub find_img {
     my ($self, $name) = @_;
     my ($json, $md5, $image);
@@ -90,8 +113,7 @@ sub find_img {
     my $resource_group = $self->resource_group;
     my $gallery = $self->image_gallery;
     my $version = calc_img_version();
-    my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_var('PUBLIC_CLOUD_ARCH', 'x86_64') . '-' . get_required_var('VERSION') . '-' . $gen;
-    $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc($definition));
+    my $definition = $self->generate_azure_image_definition();
     $json = script_output("az sig image-version show --resource-group '$resource_group' --gallery-name '$gallery' " .
           "--gallery-image-definition '$definition' --gallery-image-version '$version'", proceed_on_failure => 1, timeout => 60 * 30);
     record_info('IMGV INFO', $json);
@@ -142,6 +164,22 @@ sub calc_img_version {
     return "$build.$kiwi";
 }
 
+sub get_image_definition {
+    # Check if the given image definition, identified by the tuple (publisher, offer and sku) is present in the given image gallery
+    # returns the name of the found image definition or undef if not found
+    my ($self, $resource_group, $gallery, $publisher, $offer, $sku) = @_;
+
+    my $definitions = script_output("az sig image-definition list --resource-group '$resource_group' --gallery-name '$gallery'");
+    return undef unless ($definitions);
+    my $json_data = decode_azure_json($definitions);
+    foreach my $def (@$json_data) {
+        my $identifier = $def->{identifier};
+        next unless (defined $identifier);
+        return $def->{name} if ($identifier->{publisher} eq $publisher && $identifier->{offer} eq $offer && $identifier->{sku} eq $sku);
+    }
+    return undef;
+}
+
 sub upload_img {
     my ($self, $file) = @_;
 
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
LGTM

lib/publiccloud/azure.pm
```diff
@@ -215,12 +233,20 @@ sub upload_img {
...
-              "--gallery-image-definition '$definition' --os-type Linux --publisher '$publisher' --offer '$offer' --sku '$sku' " .
-              "--architecture '$arch' --hyper-v-generation '$hyperv' --os-state 'Generalized'", timeout => 300);
+        record_info("img-defs", "Existing image definitions:\n$images");
+
+        # Create new definition only, if there is no previous definition present
+        my $definition = $self->get_image_definition($resource_group, $gallery, $publisher, $offer, $sku);
+        if (defined $definition) {
+            record_info("use img-def", "Using found image definitions:\n$definition");
+        } else {
+            $definition = get_var("PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION", uc(get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION') . '-' . $sku));
```
asmorodskyi, line 243: why we not calling here `gen_image_definition` ?

asmorodskyi:
I like the change but before sign it I would like to get answer on one question ...

lib/publiccloud/azure.pm
```diff
@@ -150,6 +155,21 @@ sub calc_img_version {
     return "$build.$kiwi";
 }
 
+sub get_image_definition {
```
asmorodskyi, line 158:
```suggestion
sub get_image_definition_from_azure {
```

lib/publiccloud/azure.pm
```diff
@@ -59,6 +59,12 @@ sub get_image_id {
     return $self->SUPER::get_image_id($img_url);
 }
 
+sub gen_image_definition {
```
asmorodskyi, line 62:
```suggestion
sub generate_image_definition {
```

lib/publiccloud/azure.pm
```diff
@@ -59,6 +59,12 @@ sub get_image_id {
...
+    # Note: PUBLIC_CLOUD_AZURE_SKU is something like 'aarch64-gen2'
+    my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION') . '-' . get_var('PUBLIC_CLOUD_AZURE_SKU', 'gen2');
```
asmorodskyi, line 64: SKU needs to be required

asmorodskyi:
```suggestion
    my $definition = get_required_var('DISTRI') . '-' . get_required_var('FLAVOR') . '-' . get_required_var('VERSION') . '-' . '-' . get_var('PUBLIC_CLOUD_ARCH', 'x86_64') . '-' . get_required_var('PUBLIC_CLOUD_AZURE_SKU');
```

lib/publiccloud/azure.pm
grisu48: You would rename a internal subroutine in `azure.pm` from `get_image_definition` to `get_image_definition_from_azure` ?

lib/publiccloud/azure.pm
grisu48: ok

lib/publiccloud/azure.pm
grisu48: ok

lib/publiccloud/azure.pm
grisu48: good catch!

grisu48:
Blocked for now because my openQA instance is down.

m-dati:
LGTM

lib/publiccloud/azure.pm
grisu48: Renamed to `generate_azure_image_definition` because now it's also used outside the `azure.pm` module.

asmorodskyi: ok , agree

asmorodskyi: BUT currently we have `generate_azure_image_definition` and `get_image_definition`  : 
1. "azure"  in `generate_azure_image_definition` should be removed with same reasoning , right ?
2.  We have 2 functions one is "getting" image definition and another "generating" it . :thinking:  I don't have something certain to suggest here but I can just state that naming is a bit confusing ...

lib/publiccloud/azure.pm
```diff
@@ -59,6 +64,24 @@ sub get_image_id {
...
+Generated the Azure Image name from the job settings. If present, it takes the PUBLIC_CLOUD_AZURE_IMAGE_DEFINITION setting.
+If not present it generated the image definition name based on distri, version, flavor and SKU.
+
+Note: Image definitions needs to be distinct names and can only serve one architecture!
+
+Example: 'SLE-MICRO-5.4-BYOS-AZURE-X86_64-GEN1'
+=cut
+
+sub generate_azure_image_definition {
+    my $sku = get_required_var('ARCH') . '-' . $default_sku;    # PUBLIC_CLOUD_AZURE_SKU is something like 'aarch64-gen2'
```
asmorodskyi, line 80: `'aarch64-gen2'` - from where this coming ? I just checked our job group definitions and I failed to find anywhere `PUBLIC_CLOUD_AZURE_SKU=aarch64-gen2`

lib/publiccloud/azure.pm
asmorodskyi: ok after reading function description I would say it should be something like `get_image_definition_from_gallery` , WDYT ?

lib/publiccloud/azure.pm
pdostal: For image definition the `publisher - offer - sku` combination needs to be unique.
That's why it's generated and it does not matter to pass it as variable in this scenario.

On the other hand when we're launching already published images it makes sense to provide those variables.

lib/publiccloud/azure.pm
```diff
@@ -142,6 +164,22 @@ sub calc_img_version {
...
+sub get_image_definition {
+    # Check if the given image definition, identified by the tuple (publisher, offer and sku) is present in the given image gallery
+    # returns the name of the found image definition or undef if not found
+    my ($self, $resource_group, $gallery, $publisher, $offer, $sku) = @_;
+
+    my $definitions = script_output("az sig image-definition list --resource-group '$resource_group' --gallery-name '$gallery'");
+    return undef unless ($definitions);
+    my $json_data = decode_azure_json($definitions);
+    foreach my $def (@$json_data) {
+        my $identifier = $def->{identifier};
```
asmorodskyi, line 176: any chance of getting "Use of uninitialized variable" errors here ? maybe we need to do `next unless (defined $def->{identifier});` first and only than trying to assign value to variable ?  "Use of uninitialized variable" errors in autoinst log are very annoying they distract your attention but usually they not have any real problem behind

lib/publiccloud/azure.pm
grisu48: Since the gallery is the default and there are no other means/variants I would argue that unnecessarily verbose. Or do we expect there to be multiple variations of `get_image_definition` in the foreseeable future?

lib/publiccloud/azure.pm
asmorodskyi: my point that comment is stating that `PUBLIC_CLOUD_AZURE_SKU` can have value like `aarch64-gen2` but I never heard/seen something like that . So or **comment** has false statement than it needs to be removed/updated or I don't know something about `PUBLIC_CLOUD_AZURE_SKU` than I need to update my knowledge

grisu48: Exactly what Pavel says. It comes from the requirement of Azure to identify images via the tuple `(publisher, offer, sku)` and arguably the architecture is best fitted into the sku and not into `publisher` or `offer`.

asmorodskyi: from my knowledge `PUBLIC_CLOUD_AZURE_SKU` can have ONLY two possible values "gen1" or "gen2" . I never have seen architecture inside sku . Am I wrong ?

lib/publiccloud/azure.pm
grisu48: Theoretically yes, practically no because you should get a well-defined data structure from an API call.

pdostal (approved):

lib/publiccloud/azure.pm
```diff
@@ -201,16 +237,25 @@ sub upload_img {
...
+    record_info("img-defs", "Existing image definitions:\n$images");
+    my $arch = (get_var('PUBLIC_CLOUD_ARCH', 'x86_64')) eq 'x86_64' ? 'x64' : 'Arm64';
+
+    # Create new definition only, if there is no previous definition present
+    my $definition = $self->get_image_definition($resource_group, $gallery, $publisher, $offer, $sku);
+    if (defined $definition) {
+        record_info("use img-def", "Using found image definitions:\n$definition");
+    } else {
+        $definition = $self->generate_azure_image_definition();
+        my $gen = ($sku =~ "gen2" ? "V2" : "V1");
```
asmorodskyi, line 252: so `'aarch64-gen2'` will give us `V1` ? :thinking:

grisu48: `=~` but I know, perl is hard sometimes :wink:

lib/publiccloud/azure.pm
grisu48: I answered your question "any chance of getting "Use of uninitialized variable" errors here ?" with "Theoretically yes, practically no because you should get a well-defined data structure from an API call.".

Then I also said that I added a check in the next line that even in the unlikely event of that identifier not being present, the code is still behaving well-defined.

From my point of view this addresses your concern entirely, what am I missing?

lib/publiccloud/azure.pm
grisu48: Can you please make a concrete suggestion on how we should address this then?

pdostal: I am not honestly sure. If we leave `SKU` as `gen1|gen2` then the `offer will need to be something like `sle-15-SP5-byos-aarch64`.

pdostal: But the `$sku` here is anyways used only in this function which returns the image definiton name.

pdostal: And also this function (`generate_azure_image_definition`) is not currently used.

grisu48: > And also this function (`generate_azure_image_definition`) is not currently used.

It's used 2 times in `azure.pm` and one time in `provider.pm`.

pdostal: Sorry. My browser is somehow not in shape.

lib/publiccloud/azure.pm
asmorodskyi: ok

asmorodskyi (approved):
LGTM

grisu48:
VR's are passing, ready for merge

PR #17379, opened 2023-07-07
Remove version dependencies in fips setup

jlausuch:
These packages/versions were part of the fips certification but they no longer are. NSS won't be certified package, so these dependencies are not needed.

VR: https://openqa.suse.de/tests/11533058#live


msmeissn:
change looks good to me

PR #17388, opened 2023-07-10
Add kselftests support

acerv:
This patch aims to add kselftests support to openQA. This is only an initial support that works with cgroup testing. In order to enable other kselftests, we need to adapt the code and to install kselftests from other sources, rather than guessing a Makefile for each kselftests folder.

- Related ticket: https://progress.opensuse.org/issues/130784

### DIFF
--- a/tests/kernel/kselftests.pm
+++ b/tests/kernel/kselftests.pm
@@ -1,0 +1,71 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+#
+# Summary: Executes kselftests
+# Maintainer: Kernel QE <kernel-qa@suse.de>
+
+use base 'opensusebasetest';
+use testapi qw(get_var get_required_var set_var);
+use utils;
+use strict;
+use testapi;
+use warnings;
+use serial_terminal 'select_serial_terminal';
+use LTP::WhiteList;
+
+sub run
+{
+    my ($self) = @_;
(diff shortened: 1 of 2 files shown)


schlad (approved):

pevik (approved):

tests/kernel/kselftests.pm
```diff
@@ -0,0 +1,71 @@
...
+{
+    my ($self) = @_;
+    my $repo = get_var('LINUX_REPO', 'https://github.com/torvalds/linux');
+    my $branch = get_var('LINUX_BRANCH', 'master');
+
+    select_serial_terminal;
+
+    # download linux source code
+    zypper_call("in -y git");
+    assert_script_run("git clone -q --single-branch -b $branch --depth 1 $repo");
```
richiejp, line 28: FYI, possibly this is quite slow (at least last time I tried this). Instead I uploaded the tar ball as an asset and just unpack it on the SUT, i.e.

```perl
    assert_script_run('curl -o kernel.tar.xz -sS ' . data_url('ASSET_0'));
    assert_script_run('tar -xaf kernel.tar.xz && ls && cd linux-*');
```

pevik: Very good point. But OTOH if https://openqa.suse.de/tests/11406900 is verification run of the current code, whole testsuite was running only 03:41 and kaselftest.pm part only 1m 4s.

acerv: Once we have tests packages, we should get rid of source code compile anyway

PR #17398, opened 2023-07-11
wicked: add tests for dupplicated address detection (DAD)

cfconrad:
This PR add tests for DAD in auto4, static and dhcp4 scenario.

### Verification runs
```
[  742s] Target: SLE_12_SP5@x86_64
[  742s] OpenQA URL: http://dante.wicked.suse.de/tests/overview?distri=sle&version=12-SP5&build=DAD_tests
[  742s] Name               State    URL                                
[  742s] ----------------   ------   --------------------------------   
[  742s] Name               State    URL                                
[  742s] ----               -----    ---                                
[  742s] wicked_basic_ref   passed   http://dante.wicked.suse.de/t543   
[  742s] wicked_basic_sut   passed   http://dante.wicked.suse.de/t544   
[  742s] 
[  742s] Target: SLE_15_SP5@x86_64
[  742s] OpenQA URL: http://dante.wicked.suse.de/tests/overview?distri=sle&version=15-SP5&build=DAD_tests
[  742s] Name               State    URL                                
[  742s] ----------------   ------   --------------------------------   
[  742s] Name               State    URL                                
[  742s] ----               -----    ---                                
[  742s] wicked_basic_ref   passed   http://dante.wicked.suse.de/t545   
[  742s] wicked_basic_sut   passed   http://dante.wicked.suse.de/t546   
[  742s] 
[  742s] Target: openSUSE_Leap_15.5@x86_64
[  742s] OpenQA URL: http://dante.wicked.suse.de/tests/overview?distri=opensuse&version=15.5&build=DAD_tests
[  742s] Name               State    URL                                
[  742s] ----------------   ------   --------------------------------   
[  742s] Name               State    URL                                
[  742s] ----               -----    ---                                
[  742s] wicked_basic_ref   passed   http://dante.wicked.suse.de/t547   
[  742s] wicked_basic_sut   passed   http://dante.wicked.suse.de/t548   
[  742s] 
[  742s] Target: openSUSE_Tumbleweed@x86_64
[  742s] OpenQA URL: http://dante.wicked.suse.de/tests/overview?distri=opensuse&version=Tumbleweed&build=DAD_tests
[  742s] Name               State    URL                                
[  742s] ----------------   ------   --------------------------------   
[  742s] Name               State    URL                                
[  742s] ----               -----    ---                                
[  742s] wicked_basic_ref   passed   http://dante.wicked.suse.de/t549   
[  742s] wicked_basic_sut   passed   http://dante.wicked.suse.de/t550   
```


lib/wickedbase.pm
```diff
@@ -668,6 +671,24 @@ sub upload_wicked_logs {
...
+
+  do_barrier_create(<barrier_postfix> [, <test_name>] )
+
+Create a barier which can be later used to syncronize the wicked tests for SUT and REF.
+This function can be called statically. In this case the C<test_name> parameter is 
+mandatory.
+
+=cut
+sub do_barrier_create {
+    my ($self, $type, $test_name) = ref $_[0] ? @_ : (undef, @_);
```
asmorodskyi, line 684: "/me shooting myself into the head while jumping from the cliff and keeping note - perl make me do so "

lib/wickedbase.pm
```diff
@@ -1109,6 +1130,43 @@ sub check_coredump {
     }
 }
 
+sub container_runtime {
```
asmorodskyi, line 1133: @jlausuch , @pdostal , @mloviska , @ricardobranco777 , @grisu48  -  WDYT about this ? is there chance to use more our existing codebase for containers ? 

In short : wicked tests need to use some utility . it was decided to wrap it into container . So we not testing containers here just using container as a tool for testing ... But I still have doubts if this functions should be in wickedbase and that we not duplicating logic which already exists in containers tests

tests/wicked/basic/sut/t09_hotplug.pm
```diff
@@ -1,6 +1,6 @@
 # SUSE's openQA tests
 #
-# Copyright 2019 SUSE LLC
+# Copyright 2023 SUSE LLC
```
asmorodskyi, line 3:
```suggestion
# Copyright 2019-2023 SUSE LLC
```

lib/wickedbase.pm
jlausuch: At a first glance, I don't have any objection keeping what Clemens has done here. It's very specific for this use case and he's already using functions from container libraries (install_docker, etc). Moving this to container libraries would require some small re-work in those libraries, not sure if it makes sense.

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+    scapy.sniff(filter="arp",prn=lambda x: handle_packet(x,ctx), iface=interface)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(prog='arp-tool', description='Test utility to for various ARP fake tasks')
+    subparsers = parser.add_subparsers(dest='tool')
+
+    defend_arg = subparsers.add_parser('defend', help='This tool can be used to send on each "HOW-HAS-IP" request a claim RESPONSE. Used to fake address in use cases.')
+    defend_arg.add_argument('interface', help="The network interface where we listen and send the ARP reply")
+    defend_arg.add_argument('network', default="169.254.0.0/16", help="The network which we asume to claim for us!")
```
ricardobranco777, line 51:
```suggestion
    defend_arg.add_argument('network', default="169.254.0.0/16", type=network, help="The network which we asume to claim for us!")
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+import ipaddress
+import time
+import argparse
+import sys
+
+
+def defend(interface, network, count):
+    myMAC = scapy.get_if_hwaddr(interface)
+    counter = 0
+    ipaddress.ip_network(network)
```
ricardobranco777, line 12:
```suggestion
```

We could make `argparse` check for a valid network with this function and the suggestion shown later.

```
def network(string):
    return ipaddress.ip_network(string)
```

This way, the script would fail with:
`arp-tool.py: error: argument --network: invalid network value: 'xxx'
`

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+
+        print("RCV: " + packet.__repr__())
+        reply = scapy.ARP(op=2,  hwsrc=myMAC, psrc=packet.pdst, hwdst=packet[scapy.ARP].hwsrc, pdst="0.0.0.0")
+        reply = scapy.Ether(dst="ff:ff:ff:ff:ff:ff", src=myMAC) / reply
+
+        scapy.sendp(reply, iface=ctx['ifc'], verbose=False)
+        print("SND: " + reply.__repr__())
+        print("");
+
+        ctx['pkt_count'] = ctx['pkt_count'] + 1
```
ricardobranco777, line 34:
```suggestion
        ctx['pkt_count'] += 1
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+        if(ctx['max_count'] > 0 and ctx['pkt_count'] >= ctx['max_count']):
+                sys.exit(0)
+        return
+
+    my_macs=list(filter(lambda x: x != '00:00:00:00:00:00', [scapy.get_if_hwaddr(i) for i in scapy.get_if_list()]))
+    ctx = dict( pkt_count = 0, max_count = count, net = network, ifc = interface, ignore_src = my_macs);
```
ricardobranco777, line 40:
```suggestion
    ctx = dict( pkt_count = 0, max_count = count, net = str(network), ifc = interface, ignore_src = my_macs);
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(prog='arp-tool', description='Test utility to for various ARP fake tasks')
+    subparsers = parser.add_subparsers(dest='tool')
+
+    defend_arg = subparsers.add_parser('defend', help='This tool can be used to send on each "HOW-HAS-IP" request a claim RESPONSE. Used to fake address in use cases.')
+    defend_arg.add_argument('interface', help="The network interface where we listen and send the ARP reply")
+    defend_arg.add_argument('network', default="169.254.0.0/16", help="The network which we asume to claim for us!")
+    defend_arg.add_argument('--count', default=1, type=int, help="The number fakes we will send out")
```
ricardobranco777, line 52:
```suggestion
    defend_arg.add_argument('count', default=1, type=int, help="The number fakes we will send out")
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+
+    kwargs = vars(parser.parse_args())
+    tool = kwargs.pop('tool')
+    if (tool is None):
```
ricardobranco777, line 56:
```suggestion
    if tool is None:
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+        scapy.sendp(reply, iface=ctx['ifc'], verbose=False)
+        print("SND: " + reply.__repr__())
+        print("");
+
+        ctx['pkt_count'] = ctx['pkt_count'] + 1
+        if(ctx['max_count'] > 0 and ctx['pkt_count'] >= ctx['max_count']):
+                sys.exit(0)
+        return
+
+    my_macs=list(filter(lambda x: x != '00:00:00:00:00:00', [scapy.get_if_hwaddr(i) for i in scapy.get_if_list()]))
```
ricardobranco777, line 39:
```suggestion
    my_macs = {scapy.get_if_hwaddr(i) for i in scapy.get_if_list()} ^ {'00:00:00:00:00:00'}
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+        print("RCV: " + packet.__repr__())
+        reply = scapy.ARP(op=2,  hwsrc=myMAC, psrc=packet.pdst, hwdst=packet[scapy.ARP].hwsrc, pdst="0.0.0.0")
+        reply = scapy.Ether(dst="ff:ff:ff:ff:ff:ff", src=myMAC) / reply
+
+        scapy.sendp(reply, iface=ctx['ifc'], verbose=False)
+        print("SND: " + reply.__repr__())
+        print("");
+
+        ctx['pkt_count'] = ctx['pkt_count'] + 1
+        if(ctx['max_count'] > 0 and ctx['pkt_count'] >= ctx['max_count']):
```
ricardobranco777, line 35:
```suggestion
        if ctx['max_count'] > 0 and ctx['pkt_count'] >= ctx['max_count']:
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+    myMAC = scapy.get_if_hwaddr(interface)
+    counter = 0
+    ipaddress.ip_network(network)
+
+    def handle_packet(packet, ctx):
+        if packet[scapy.ARP].op != 1: 
+            return
+
+        if ipaddress.ip_address(packet.pdst) not in ipaddress.ip_network(ctx['net']):
+            print("IGN: Ignore incomming arp request with pdst=" + packet.pdst)
```
ricardobranco777, line 19:
```suggestion
            print("IGN: Ignore incoming arp request with pdst=" + packet.pdst)
```

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
...
+            return
+
+        if str(packet.src) in ctx['ignore_src']:
+            print("IGN: Ignore incomming arp request by MAC with src=" + packet.src)
```
ricardobranco777, line 23:
```suggestion
            print("IGN: Ignore incoming arp request by MAC with src=" + packet.src)
```

ricardobranco777 (approved):
Some small suggestions. Otherwise LGTM.

lib/wickedbase.pm
cfconrad: because this function can be called `statically` like
```perl
wickedbase::do_barrier_create($type, $testname);
```
And as a member method:
```perl
my $wb = wickedbase->new();
$wb->do_barrier_create($type, $testname);
```
The difference is, that the later will get the first `$self` parameter, while the first one don't.

data/wicked/arp-tool.py
cfconrad: wow nice, thx!

data/wicked/arp-tool.py
cfconrad: why is this needed? I would guess I can store any kind of object in the dict.

ricardobranco777: > why is this needed? I would guess I can store any kind of object in the dict.

Because we're using the value returned by the `network` function that I suggested.

lib/wickedbase.pm
asmorodskyi: and why we need to support both ways ?

lib/wickedbase.pm
asmorodskyi: ok , than let's keep it as "implementer" have chosen

data/wicked/arp-tool.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
asmorodskyi: ignore it

data/wicked/arp-tool.py
ricardobranco777: It's just sets. I tried in my machine and it works. XD

lib/wickedbase.pm
cfconrad: we have it here: 
 * https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/35980bf112836ad1799e68348235a410422e7601/tests/wicked/locks_init.pm#L21
where we use the static way and with this we need to pass the "testname".
But we also can use it within a test-case, where we don't need to pass the testname again:
 * https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/35980bf112836ad1799e68348235a410422e7601/tests/wicked/basic/ref/t12_DAD_dhcp4_detect.pm#L15

data/wicked/arp-tool.py
cfconrad: And I like it, thx for the pointer!

cfconrad:
> Wow! Looks like a lot of work! :)
> 
> Can you please make sure this works on all SP versions in maintenance as well as Tumbleweed tests (as it's part of the basic suite)?

I have VR's for all needed sles versions: http://dante.wicked.suse.de/tests/overview?build=DAD_tests

*BUT* there is a issue which I didn't solved till now. If I try to run these tests on osd, they fail. I think it's again that bridge/GRE network setup. In our servers, we do not use workers on different hosts, as we encountered unreliable network issues.
E.g. https://openqa.suse.de/tests/11573763#step/t12_DAD_dhcp4_detect/85 ...
When I connect to the running test via vnc and try these things manually, they behaves as expected.

asmorodskyi (approved):
LGTM

asmorodskyi:
> > Wow! Looks like a lot of work! :)
> > Can you please make sure this works on all SP versions in maintenance as well as Tumbleweed tests (as it's part of the basic suite)?
> 
> I have VR's for all needed sles versions: http://dante.wicked.suse.de/tests/overview?build=DAD_tests
> 
> _BUT_ there is a issue which I didn't solved till now. If I try to run these tests on osd, they fail. I think it's again that bridge/GRE network setup. In our servers, we do not use workers on different hosts, as we encountered unreliable network issues. E.g. https://openqa.suse.de/tests/11573763#step/t12_DAD_dhcp4_detect/85 ... When I connect to the running test via vnc and try these things manually, they behaves as expected.

I think we need to find a way to run wicked tests in osd **without** GRE protocol ( make sure that both instances are running on same worker ) :thinking:

pdostal:
> I think we need to find a way to run wicked tests in osd **without** GRE protocol ( make sure that both instances are running on same worker ) 🤔

Each worker should have it's own unique `WORKER_CLASS` for that reason.

cfconrad:
plz consider this link for better comparison of the latest changes: https://github.com/os-autoinst/os-autoinst-distri-opensuse/compare/e728299eb..9544cd7e1ee255af58e93c94fc3c5041a375277c 
I did a rebase on upstream/master in between, which scrambled the diff

cfconrad:
with latest version I produced green bubbles in osd and ci... plz check again, thx

PR #17400, opened 2023-07-12
Support install hosts via iPXE for virt and hana test

Julie-CAO:
- Schedule installation/ipxe_install.pm for tests with IPXE=1.
- Add more kernel and linuxrc arguments in bootscript for virt and hana tests.
- Finally don't touch the prefixing in autoyast profile path with SUT_IP in lib/autoyast.pm as it will be fixed in another PR. It breaks IPXE autoyast installation, fg. http://openqa.suse.de/tests/11712594#step/installation/1. 
- Prevent warnings `Use of uninitialized value...`, fg. https://openqa.suse.de/tests/11712594/logfile?filename=autoinst-log.txt
- "Boot flag" is not required anymore after vh015 firmwire upgrade
- Support passing "EXTRA_PXE_CMDLINE" to bootscripts.
- Enable sshd during autoyast installation for sake of debugging and post_fail_hook().
- Allow to install other SLE release with '+MIRROR_HTTP' as always.
- Be able to install TW with installer DVD outside O3. fg.[TW in my own openqa server](http://10.67.129.27/tests/77)
- Differentiate iPXE server on baremental-support service and static
  iPXE server by introducing a new setting, IPXE_STATIC(O3 and OSD share same ipxe menu)
- Add hard disk's wwns of two machines to autoyast profiles for virt test s in a seperate commit

Related ticket: https://progress.opensuse.org/issues/128060
Needles MR: https://gitlab.suse.de/openqa/os-autoinst-needles-sles/-/merge_requests/1647

Verification run: 
[xen tests with autoyast](http://10.67.129.27/tests/66)
[sle12sp5 with autoyast](http://10.67.129.27/tests/65)
[sle12sp5 xen without autoyast](http://10.67.129.27/tests/64)
[kvm without autoyast on OSD](https://openqa.suse.de/tests/11626483#)
[kvm with autoyast on OSD](https://openqa.suse.de/tests/11895292)
[prj2 tests with autoyast](http://10.67.129.27/tests/67)
[TW kvm test on O3](https://openqa.opensuse.org/tests/3522659)
[UEFI test with autoyast](http://10.67.129.27/tests/70)
[TW in my own openqa server](http://10.67.129.27/tests/77)

### DIFF
--- a/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
@@ -118,7 +118,7 @@
     <ntp_sync>systemd</ntp_sync>
   </ntp-client>
   <partitioning config:type="list">
-  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745', 'openqaipmi5' => 'wwn-0x5000c5008711f2fc'};
+  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745', 'openqaipmi5' => 'wwn-0x5000c5008711f2fc', 'ph052' => 'wwn-0x5000c500e28a91bb', 'vh015' => 'wwn-0x50000f0a47804240', 'vh016' => 'wwn-0x50000f0a47803ea0'};
   % my $hostname = (split(/\./, $get_var->("SUT_IP")))[0];
   % my $device_id = defined($wwn->{$hostname}) ? '/dev/disk/by-id/' . $wwn->{$hostname} : ''; 
     <drive>
--- a/lib/main_common.pm
+++ b/lib/main_common.pm
@@ -425,6 +425,9 @@ sub load_boot_tests {
         loadtest "boot/boot_from_pxe";
         set_var("DELAYED_START", get_var("PXEBOOT"));
     }
+    elsif (get_var("IPXE")) {
+        loadtest "installation/ipxe_install";
+    }
     else {
         loadtest "installation/data_integrity" if data_integrity_is_applicable;
         loadtest "installation/bootloader" unless load_bootloader_s390x();
--- a/tests/installation/ipxe_install.pm
+++ b/tests/installation/ipxe_install.pm
@@ -48,7 +48,7 @@ sub poweron_host {
 sub set_pxe_boot {
     while (1) {
         my $stdout = ipmitool('chassis bootparam get 5');
-        last if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/;
+        last if $stdout =~ m/Force PXE/;
         diag "setting boot device to pxe";
         my $options = get_var('IPXE_UEFI') ? 'options=efiboot' : '';
         ipmitool("chassis bootdev pxe ${options}");
@@ -63,8 +63,8 @@ sub set_bootscript {
     my $url = "$http_server/v1/bootscript/script.ipxe/$ip";
     my $arch = get_required_var('ARCH');
     my $autoyast = get_var('AUTOYAST', '');
-    my $regurl = get_var('SCC_URL');
-    my $console = get_var('IPXE_CONSOLE');
+    my $regurl = get_var('SCC_URL', '');
+    my $console = get_var('IPXE_CONSOLE', '');
     my $mirror_http = get_required_var('MIRROR_HTTP');
 
     # trim all strings from variables to get rid of bogus whitespaces
@@ -87,19 +87,46 @@ sub set_bootscript {
         $initrd .= "/boot/$arch/loader/initrd";
     }
 
+    if (get_var('SUT_NETDEVICE') and !is_tumbleweed) {
+        my $interface = get_var('SUT_NETDEVICE');
+        $install .= "?device=$interface ifcfg=$interface=dhcp4 ";
+    }
 
     my $cmdline_extra;
     $cmdline_extra .= " regurl=$regurl " if $regurl;
     $cmdline_extra .= " console=$console " if $console;
 
+    # Support passing both EXTRA_PXE_CMDLINE to bootscripts
+    $cmdline_extra .= get_var('EXTRA_PXE_CMDLINE') . ' ' if get_var('EXTRA_PXE_CMDLINE');
     $cmdline_extra .= " root=/dev/ram0 initrd=initrd textmode=1" if check_var('IPXE_UEFI', '1');
 
     if ($autoyast ne '') {
-        $cmdline_extra .= " autoyast=$autoyast ";
+        $cmdline_extra .= " autoyast=$autoyast sshd=1 sshpassword=$testapi::password ";
     } else {
-        $cmdline_extra .= " sshd=1 vnc=1 VNCPassword=$testapi::password sshpassword=$testapi::password ";    # trigger default VNC installation
+        if (check_var('VIDEOMODE', 'text')) {
+            $cmdline_extra .= " ssh=1 sshpassword=$testapi::password ";    # trigger ssh-text installation
+        }
+        else {
+            $cmdline_extra .= " vnc=1 VNCPassword=$testapi::password ";    # trigger default VNC installation
+        }
     }
-    $cmdline_extra .= ' plymouth.enable=0 ';
+    $cmdline_extra .= " plymouth.enable=0 ";
+
+    # Extra options for virtualization tests with ipmi backend
+    if (get_var('VIRT_AUTOTEST') || get_var('HANA_PERF')) {
+        $cmdline_extra .= " video=1024x768 vt.color=0x07 " if check_var('VIDEOMODE', 'text');
+        # Support either IPXE_CONSOLE=ttyS1,115200 or SERIALDEV=ttyS1
+        my $serial_dev;
+        if (get_var('IPXE_CONSOLE')) {
+            get_var('IPXE_CONSOLE') =~ /^(\w+)/;
+            $serial_dev = $1;
+        }
+        else {
+            $serial_dev = get_var('SERIALDEV', 'ttyS1');
+            $cmdline_extra .= " console=$serial_dev,115200 ";
+        }
+        $cmdline_extra .= " Y2DEBUG=1 linuxrc.log=/dev/$serial_dev linuxrc.core=/dev/$serial_dev linuxrc.debug=4,trace reboot_timeout=0";
+    }
 
     my $bootscript = <<"END_BOOTSCRIPT";
 #!ipxe
@@ -124,6 +151,11 @@ END_BOOTSCRIPT
 
     my $response = HTTP::Tiny->new->request('POST', $url, {content => $bootscript, headers => {'content-type' => 'text/plain'}});
     diag "$response->{status} $response->{reason}\n";
+
+    # Comment out but keep the block for possible debug purpose.
+    #diag "\nThe bootscript from http server is: \n";
+    #$response = HTTP::Tiny->new->get($url);
+    #print "$response->{content}\n";
 }
 
 sub set_bootscript_hdd {
@@ -141,6 +173,21 @@ END_BOOTSCRIPT
     diag "$response->{status} $response->{reason}\n";
 }
 
+sub enter_o3_ipxe_boot_entry {
+    ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
+    assert_screen('o3-ipxe-menu', 210);
+    my $key = get_var('HOST_INSTALL_AUTOYAST') ? (is_kvm_host ? 'k' : 'z') : 't';
+    send_key "$key";
+    # try one more time as sometimes sending key does not take effect
+    send_key "$key" if check_screen('o3-ipxe-menu');
+    # confirm the dialog asking for user modifications to kernel, cmdline and initrd
+    # set limit to 10 in case some keys don't go through
+    for (1 .. 10) {
+        last if check_screen([qw(load-linux-kernel load-initrd)], 3);
+        send_key "ret";
+    }
+}
+
 
 sub run {
     my $self = shift;
@@ -148,7 +195,7 @@ sub run {
     poweroff_host;
 
     #virtualization tests use a static ipxe configuration file in O3
-    set_bootscript unless get_var('VIRT_AUTOTEST') && is_tumbleweed;
+    set_bootscript unless get_var('IPXE_STATIC');
 
     set_pxe_boot;
 
@@ -156,35 +203,24 @@ sub run {
 
     select_console 'sol', await_console => 0;
 
-    #use ipxe bootloader in O3
-    #it is static menu and choose the TW entry to start installation
-    if (get_var('VIRT_AUTOTEST') && is_tumbleweed) {
-        save_screenshot;
-        ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
-        assert_screen('o3-ipxe-menu', 300);
-        my $key = get_var('HOST_INSTALL_AUTOYAST') ? (is_kvm_host ? 'k' : 'z') : 't';
-        send_key "$key";
-        # try one more time as sometimes sending key does not take effect
-        send_key "$key" if check_screen('o3-ipxe-menu');
-
-        # confirm the dialog asking for user modifications to kernel, cmdline and initrd
-        # set limit to 10 in case some keys don't go through
-        for (1 .. 10) {
-            last if check_screen([qw(load-linux-kernel load-initrd)], 3);
-            send_key "ret";
-        }
-
+    # Print screenshots for ipxe boot process
+    if (get_var('VIRT_AUTOTEST') || get_var('HANA_PERF')) {
+        #it is static menu and choose the TW entry to start installation
+        enter_o3_ipxe_boot_entry if get_var('IPXE_STATIC');
         assert_screen([qw(load-linux-kernel load-initrd)], 240);
         # Loading initrd spend much time(fg. 10-15 minutes to Beijing SUT)
         # Downloading from O3 became much more quick, some needles may not be caught.
-        check_screen([qw(start-tw-install network-config-created loading-installation-system)], 360);
-        if (match_has_tag("start-tw-install")) {
-            record_info("TW install", "Start with installing Tumbleweed...");
+        check_screen([qw(start-tw-install start-sle-install network-config-created)], 60);
+        if (match_has_tag('start-tw-install')) {
+            record_info("Install TW", "Start installing Tumbleweed...");
+        }
+        elsif (match_has_tag('start-sle-install')) {
+            record_info("Install SLE", "Start installing SLE...");
         }
         else {
-            record_info("TW install?", "Pls make sure it is Tumbleweed that is being installed.", result => 'softfail');
+            record_info("Install others?", "Pls make sure the product that is expected to be installed.");
         }
-        assert_screen([qw(network-config-created loading-installation-system)], 60);
+        assert_screen([qw(network-config-created loading-installation-system sshd-server-started autoyast-installation)], 300);
         return if get_var('AUTOYAST');
         wait_still_screen(stilltime => 12, similarity_level => 60, timeout => 30) unless check_screen('sshd-server-started', timeout => 60);
         save_screenshot;
@@ -197,8 +233,8 @@ sub run {
     }
     else {
         my $ssh_vnc_wait_time = 1500;
-        #for TW virtualization test, 15 minutes is enough to load installation system, 75 minutes is too long
-        $ssh_vnc_wait_time = 300 if get_var('VIRT_AUTOTEST') && is_tumbleweed;
+        #for virtualization test, 9 minutes is enough to load installation system, 75 minutes is too long
+        $ssh_vnc_wait_time = 180 if get_var('VIRT_AUTOTEST');
         my $ssh_vnc_tag = eval { check_var('VIDEOMODE', 'text') ? 'sshd' : 'vnc' } . '-server-started';
         my @tags = ($ssh_vnc_tag);
         if (check_screen(\@tags, $ssh_vnc_wait_time)) {
--- a/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_12.xml.ep
+++ b/data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_12.xml.ep
@@ -112,7 +112,7 @@
     </dns>
   </networking>
   <partitioning config:type="list">
-  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745'};
+  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745', 'ph052' => 'wwn-0x5000c500e28a91bb', 'vh015' => 'wwn-0x50000f0a47804240', 'vh016' => 'wwn-0x50000f0a47803ea0'};
   % my $hostname = (split(/\./, $get_var->("SUT_IP")))[0];
   % my $device_id = defined($wwn->{$hostname}) ? '/dev/disk/by-id/' . $wwn->{$hostname} : '';
     <drive>
--- a/tests/autoyast/installation.pm
+++ b/tests/autoyast/installation.pm
@@ -129,7 +129,7 @@ sub run {
     # So push a needle to check upcoming reboot on zVM that is a way to indicate the stage done
     push @needles, 'autoyast-stage1-reboot-upcoming' if is_s390x || (is_pvm && !is_upgrade);
     # Similar situation over IPMI backend, we can check against PXE menu
-    push @needles, qw(prague-pxe-menu qa-net-selection) if is_ipmi;
+    push @needles, qw(prague-pxe-menu qa-net-selection) if is_ipmi and !get_var('IPXE');
     # Import untrusted certification for SMT
     push @needles, 'untrusted-ca-cert' if get_var('SMT_URL');
     # Workaround for removing package error during upgrade
@@ -354,7 +354,7 @@ sub run {
     # match openSUSE Welcome dialog on matching distros
     push(@needles, 'opensuse-welcome') if opensuse_welcome_applicable;
     # There will be another reboot for IPMI backend
-    push @needles, qw(prague-pxe-menu qa-net-selection) if is_ipmi;
+    push @needles, qw(prague-pxe-menu qa-net-selection) if is_ipmi and !get_var('IPXE');
     until (match_has_tag('reboot-after-installation')
           || match_has_tag('opensuse-welcome'))
     {


tests/installation/ipxe_install.pm
```diff
@@ -95,12 +99,26 @@ sub set_bootscript {
     $cmdline_extra .= " root=/dev/ram0 initrd=initrd textmode=1" if check_var('IPXE_UEFI', '1');
 
     if ($autoyast ne '') {
-        $cmdline_extra .= " autoyast=$autoyast ";
+	    #julie        $cmdline_extra .= " autoyast=$autoyast ";
+        $cmdline_extra .= " autoyast=$autoyast sshd=1 sshpassword=$testapi::password ";
```
Dawei-Pang, line 103: During autoyast installation, no interactive is required, I think we need not sshd and sshpassword for boot cmdline, how about you?

tests/installation/ipxe_install.pm
```diff
@@ -15,7 +15,7 @@ use testapi;
 use bmwqemu;
 use ipmi_backend_utils;
 use version_utils qw(is_upgrade is_tumbleweed);
-use bootloader_setup 'prepare_disks';
+use bootloader_setup qw(prepare_disks bootmenu_default_params specific_bootmenu_params);
```
Dawei-Pang, line 18: Shall we think about remove `bootmenu_default_params` `specific_bootmenu_params` in `use`, and do not touch `bootloader_setup.pm`?

Julie-CAO: I am managing to append these arguments directly.

lib/autoyast.pm
```diff
@@ -713,9 +713,13 @@ sub upload_profile {
...
 
-    if (check_var('IPXE', '1')) {
-        $path = get_required_var('SUT_IP') . $path;
-    }
+    # Don't know why the 'if' statement is required
+    # it fails to save the autoyast profile to the job URL on the worker
+    # fg. http://openqa.suse.de/tests/11712594#step/installation/1
+    # so comment out this part
+    # if (check_var('IPXE', '1')) {
+    #     $path = get_required_var('SUT_IP') . $path;
```
Dawei-Pang, line 721: Reproduced the issue when including IPXE=1, it generates the `$path` likes`${SUT_IP}autoyast_hanaperf/mu.xml`, but no code loading autoinst.xml following it.
I agree to remove this code block.

tests/installation/ipxe_install.pm
```diff
@@ -95,12 +99,26 @@ sub set_bootscript {
     $cmdline_extra .= " root=/dev/ram0 initrd=initrd textmode=1" if check_var('IPXE_UEFI', '1');
```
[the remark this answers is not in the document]
Julie-CAO: Ok. done.

tests/installation/ipxe_install.pm
```diff
@@ -97,10 +101,23 @@ sub set_bootscript {
...
+        else {
+            $cmdline_extra .= " sshd=1 vnc=1 VNCPassword=$testapi::password ";    # trigger default VNC installation
+        }
+        $cmdline_extra .= "sshpassword=$testapi::password ";
     }
     $cmdline_extra .= ' plymouth.enable=0 ';
 
+    # Extra options for virtualization tests with ipmi backend
+    if (check_var('VIDEOMODE', 'text')) {
+        my $serial_dev = get_var('SERIALDEV', 'ttyS1');
```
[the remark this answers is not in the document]
Julie-CAO: There is no CONSOLE but SEREAL_DEV in virt tests. I am thinking of change to use CONSOLE or support both of them.

tests/installation/ipxe_install.pm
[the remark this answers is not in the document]
Dawei-Pang: Thanks!

tests/installation/ipxe_install.pm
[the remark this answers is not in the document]
Dawei-Pang: I suggest supporting both of them, thanks!

tests/installation/ipxe_install.pm
```diff
@@ -48,7 +48,7 @@ sub poweron_host {
 sub set_pxe_boot {
     while (1) {
         my $stdout = ipmitool('chassis bootparam get 5');
-        last if $stdout =~ m/Boot Flag Valid[\d\D]*Force PXE/;
+        last if $stdout =~ m/Force PXE/;
```
mdoucha, line 51: If you drop `Boot Flag Valid` from the regex, the function will accept a state where IPMI boot override is disabled and the SUT will boot according to BIOS settings, ignoring the `Force PXE` flag.

mdoucha (dismissed):

tests/installation/ipxe_install.pm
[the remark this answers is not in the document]
Julie-CAO: yes. find that `SEREAL_DEV` is a mandatory setting for virt tests, so support both of them.

tests/installation/ipxe_install.pm
Julie-CAO: To check `Boot Flag Valid` was added by me in a previous PR because we had a machine that only worked with this flag. We upgraded the BIOS and firmwire of this machine days ago, after that I found Boot Flag always remains 'Invalid' which stopped the test to be continued, fg. http://10.67.129.27/tests/28/logfile?filename=autoinst-log.txt

And it seems that the flag does not effect the tests anymore. I will do more verification runs on more machines.

mdoucha: OK, never mind then.

Julie-CAO: wait a moment... I might speak too early. it seems that something wrong with the machine, it boot from pxe to install SLE repeatedly...   I am going to figure out why its Boot Flag keeps 'Invalid' firstly.

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
```diff
@@ -118,7 +118,7 @@
     <ntp_sync>systemd</ntp_sync>
   </ntp-client>
   <partitioning config:type="list">
-  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745', 'openqaipmi5' => 'wwn-0x5000c5008711f2fc'};
+	  % my $wwn = {'quinn' => 'wwn-0x5000c50099db2117', 'kermit-1' => 'wwn-0x500a075119406ab6', 'gonzo-1' => 'wwn-0x500a075119406aa6', 'fozzie-1' => 'wwn-0x55cd2e414f1f16f1', 'scooter-1' => 'wwn-0x55cd2e414f1760e2', 'amd-zen3-gpu-sut1-1' => 'wwn-0x500a075133755d4b', 'ix64ph1075' => 'wwn-0x5000c5004f25d745', 'openqaipmi5' => 'wwn-0x5000c5008711f2fc', '10.67.130.52' => 'wwn-0x5000c500e28a91bb', '10.67.131.15' => 'wwn-0x50000f0a47804240'};
```
alice-suse, line 121: Why not use domain name here? Better than direct ip?

lib/autoyast.pm
alice-suse: Will it change the path for kernel team? Better to involve kernel team PO for review when PR is ready.

tests/installation/ipxe_install.pm
```diff
@@ -87,19 +87,44 @@ sub set_bootscript {
...
+            $cmdline_extra .= " vnc=1 VNCPassword=$testapi::password ";    # trigger default VNC installation
+        }
+    }
+    $cmdline_extra .= ' sshd=1 sshpassword=$testapi::password plymouth.enable=0 ';
+
+    # Extra options for virtualization tests with ipmi backend
+    if (get_var('VIRT_AUTOTEST') || get_var('HANA_PERF')) {
+        $cmdline_extra .= " vga=791 video=1024x768 vt.color=0x07";
+        my $serial_dev;
+        if (get_var('IPXE_CONSOLE')) {
```
alice-suse, line 118: What's the background for differentiating IPXE_CONSOLE and SERIALDEV? It seems both are to speciy the real serial console?

tests/installation/ipxe_install.pm
```diff
@@ -141,6 +171,21 @@ END_BOOTSCRIPT
     diag "$response->{status} $response->{reason}\n";
 }
 
+sub enter_o3_ipxe_boot_entry {
+    ipmitool('chassis power reset') unless check_screen([qw(o3-ipxe-menu ipxe-boot-failure)], 180);
+    assert_screen('o3-ipxe-menu', 300);
```
alice-suse, line 176: With default TIME_SCALE 3, it wil give 900 seconds(15 minutes) to asset ipxe menu. Is it really needed? How about fail earlier?

tests/installation/ipxe_install.pm
```diff
@@ -156,27 +201,24 @@ sub run {
...
+        if (match_has_tag('start-tw-install')) {
+            record_info("Install TW", "Start installing Tumbleweed...");
+        }
+        elsif (match_has_tag('start-sle-install')) {
+            record_info("Install SLE", "Start installing SLE...");
         }
         else {
-            record_info("TW install?", "Pls make sure it is Tumbleweed that is being installed.", result => 'softfail');
+            record_info("Install others?", "Pls make sure the product that is expected to be installed.");
         }
```
alice-suse, line 220: Do we need to add the  needles per sles product? If not, that's fine. Otherwise maybe a little expensive just for record_info.

alice-suse:
Thanks @Julie-CAO  for the great work. I left only some minor comments.

lib/autoyast.pm
```diff
@@ -713,9 +713,13 @@ sub upload_profile {
     my $profile = $args{profile};
     my $path = $args{path};
 
-    if (check_var('IPXE', '1')) {
-        $path = get_required_var('SUT_IP') . $path;
```
frankenmichl, line 717: We should agree with @mdoucha if we want to keep this or not, or maybe add another variable to control this?

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
Julie-CAO: yes domaine name is better. 

With autoyast installation, we must change the settings in workers.ini if `SUT_IP=10.67.130.15` to `SUT_IP=vh015.qa2.suse.asia`.

tests/installation/ipxe_install.pm
Julie-CAO: yes, IPXE_CONSOLE was introduced by ipxe installation and is being used by kernel team. SERIALDEV is being used by virt team and it is mandatory as it is referenced in host autoyast profiles. So we have to support both of them.

Julie-CAO: usually these value are valid for them:
```
IPXE_CONSOLE=ttyS1,115200
SERIALDEV=ttyS1
```

alice-suse: ok, maybe adding a comment to explain them will be useful for reference

tests/installation/ipxe_install.pm
Julie-CAO: it depends on the restart time including power off and power on. I cannot get an acurate time from a test case, https://openqa.opensuse.org/tests/3446335#step/ipxe_install/22, but the time seems to be shorter than 900s. Let's reduce to 210 and keep eyes on it in future tests.

tests/installation/ipxe_install.pm
Julie-CAO: I have added needles for TW and SLE seperately, no need to add needles for each SLE release.

lib/autoyast.pm
Julie-CAO: I see it is moved to another place in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583. I'll not leave this part to your team as you are working on it. In my a few verify tests, these lines are not required at here.

data/virt_autotest/host_unattended_installation_files/autoyast/dev_host_15.xml.ep
Julie-CAO: I added this task to https://progress.opensuse.org/issues/134261

Julie-CAO:
Hello everyone, the code change is ready. The descripton and verification runs are updated. Welcome review! @frankenmichl @mdoucha @Dawei-Pang @alice-suse @guoxuguang @waynechen55 @nanzhg @tonyyuan1 @RoyCai7 @tbaev @varunkojha @JoyceNa @sudarshannm @rmaliska

lib/autoyast.pm
Julie-CAO: I keep this change in this PR for a moment for sake of verifying tests on openqa.

waynechen55:
Does this support aarch64 as well ?

tests/installation/ipxe_install.pm
```diff
@@ -87,19 +95,46 @@ sub set_bootscript {
...
+        }
+        else {
+            $cmdline_extra .= " vnc=1 VNCPassword=$testapi::password ";    # trigger default VNC installation
+        }
+    }
+    $cmdline_extra .= " plymouth.enable=0 ";
+
+    # Extra options for virtualization tests with ipmi backend
+    if (get_var('VIRT_AUTOTEST') || get_var('HANA_PERF')) {
+        $cmdline_extra .= " vga=791 video=1024x768 vt.color=0x07";
```
waynechen55, line 125: Some options, like ```vga``` is already deprecated.

Julie-CAO: Are these 3 parameters all deprecated? I am not aware of it. I'll do some reseach and tests before removing them.

Julie-CAO: Are these 3 parameters all deprecated? I am not aware of it. I'll do some reseach and tests before removing them.

Julie-CAO:
> Does this support aarch64 as well ?

Aarch64 boot with UEFI instead of legacy BIOS? UEFI should be supported, but about aarch64 I am not sure. I guess there should be a new ticket to handle it seperately.

tests/installation/ipxe_install.pm
Julie-CAO: `vga=xx` is deprecated in 32-bit linux and Grub2. `video=xx` and `vt.colar=xx` are still in use.  I removed `vga=791`.

tests/installation/ipxe_install.pm
varunkojha: seems to depend on the trade-off between the potential security concerns of including passwords in the command line and the practical benefits of having SSH access for troubleshooting purposes during non-interactive AutoYaST installations. What happens if we remove it => A hook might need SSH access to collect important logs, I hope the understanding is correct.

lib/main_common.pm
```diff
@@ -425,6 +425,9 @@ sub load_boot_tests {
         loadtest "boot/boot_from_pxe";
         set_var("DELAYED_START", get_var("PXEBOOT"));
     }
+    elsif (get_var("IPXE")) {
+        loadtest "installation/ipxe_install";
+    }
```
waynechen55, line 430: So ```ipxe_install``` has never been scheduled before in ```main_common.pm``` ?

Julie-CAO: No, it is usually loaded in job scheduler yaml files directly in test suites.

tests/installation/ipxe_install.pm
```diff
@@ -70,11 +70,19 @@ sub set_bootscript {
     # trim all strings from variables to get rid of bogus whitespaces
     $url =~ s/^\s+|\s+$//g;
     $arch =~ s/^\s+|\s+$//g;
-    $autoyast =~ s/^\s+|\s+$//g;
-    $regurl =~ s/^\s+|\s+$//g;
-    $console =~ s/^\s+|\s+$//g;
+    $autoyast =~ s/^\s+|\s+$//g if $autoyast;
+    $regurl =~ s/^\s+|\s+$//g if $regurl;
+    $console =~ s/^\s+|\s+$//g if $console;
```
waynechen55, line 75: I do not think these three ```if``` are very necessary. Keep these line as what they were is enough.

tests/installation/ipxe_install.pm
Julie-CAO: Yes, you get it right. This secrect of ssh login is not so sensible as the IPMI access.

tests/installation/ipxe_install.pm
Julie-CAO: They bring `uninitialized value ...` warnings in auto-inst.txt. To remove this kind of warning and make logs clean follows https://progress.opensuse.org/issues/120750.

waynechen55: ```$autoyast``` already initialized by ```get_var('AUTOYAST', '');```. So I think you just need
```
my $regurl = get_var('SCC_URL', '');
my $console = get_var('IPXE_CONSOLE', '');
```
instead of using ```if```.

waynechen55:
I think you need some verifications with developing product, namely 15-SP6 to verify it also works with developing product.

tests/installation/ipxe_install.pm
waynechen55: Using ssh allows user to login anyway. I think we should keep it.

tests/installation/ipxe_install.pm
```diff
@@ -141,50 +181,54 @@ END_BOOTSCRIPT
...
+
 
 sub run {
     my $self = shift;
 
     poweroff_host;
 
     #virtualization tests use a static ipxe configuration file in O3
-    set_bootscript unless get_var('VIRT_AUTOTEST') && is_tumbleweed;
+    set_bootscript unless get_var('VIRT_AUTOTEST') and is_tumbleweed and get_var('OPENQA_URL') =~ "openqa.opensuse.org";
```
waynechen55, line 206: According to your comment, is_tumbleweed should be removed from ```unless``` ?

tests/installation/ipxe_install.pm
Julie-CAO: https://en.opensuse.org/SDB:Linuxrc
```
ssh=1  --  This parameter enables access to linuxrc via SSH when performing the installation with YaST in text mode or via X11 forwarding. Use ssh -X root@hostname or ssh -Y root@hostname for X11 forwarding.

sshd=1  --- This parameter enables login to the installation system while the installation is running.
```

Julie-CAO:
> 1. I think you need some verifications with developing product, namely 15-SP6 to verify it also works with developing product.
Verifiying 15-SP6 will be done during it is deployment. They are seperate tasks. Introduing unmature sp6 tests will bring unexpected troubles. It is not smart.

> 2. What a failed ipxe install looks like ? Enough information available for user to do follow-up ?
Here is a recent failed case with incorrect kernel path: http://10.67.129.27/tests/85
screenshots + logs + videos will help locating failures.

tests/installation/ipxe_install.pm
Julie-CAO: I don't know if ALP behaves same way as Tumbleweed, so just to be limited to TW.

tests/installation/ipxe_install.pm
waynechen55: Yes. I was wondering whether ```ssh=1``` works with auotyast.

tests/installation/ipxe_install.pm
Julie-CAO: Yes, it works. thanks.

tests/installation/ipxe_install.pm
Julie-CAO: Not sure. Let's see in future tests.  'ssh=1' is not propriate, because some tests also run on sol console or on VNC graphic mode.

tests/installation/ipxe_install.pm
```diff
@@ -75,6 +75,14 @@ sub set_bootscript {
     $console =~ s/^\s+|\s+$//g;
     $mirror_http =~ s/^\s+|\s+$//g;
 
+    # Install other release rather than the VERSION
+    if (get_var('REPO_0_TO_INSTALL')) {
+        my $openqa_url = get_required_var('OPENQA_URL');
+        $openqa_url = 'http://' . $openqa_url unless $openqa_url =~ /http:\/\//;
```
mdoucha, line 81: This condition will result in `$openqa_url = 'http://https://openqa.opensuse.org'`

mdoucha (changes requested):

tests/installation/ipxe_install.pm
```diff
@@ -124,6 +159,11 @@ END_BOOTSCRIPT
 
     my $response = HTTP::Tiny->new->request('POST', $url, {content => $bootscript, headers => {'content-type' => 'text/plain'}});
     diag "$response->{status} $response->{reason}\n";
```
Dawei-Pang, line 161: Suggest to add the code as following to checking if IPXE_HTTPSERVER works
```    die "Fail to post IPXE_HTTPSERVER\n" unless $response->{success};```

**Comment by waynechen55** [comment]
> > 1. I think you need some verifications with developing product, namely 15-SP6 to verify it also works with developing product.
> 
> Verifiying 15-SP6 will be done during its deployment. They are seperate tasks. Introducing unmature sp6 tests will bring unexpected troubles. It is not smart.

I do not think this is ready to be merged without verification with 15-SP6.

File: lib/autoyast.pm (hunk 3 above)
Comment (Julie-CAO) [reply lib/autoyast.pm]: Finally I made this change limited to virt tests and hana tests only.

File: tests/installation/ipxe_install.pm (hunk 16 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm:81]: Yes, good catch. thanks. fixed.

File: tests/installation/ipxe_install.pm (hunk 17 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm:161]: Added. Trying out to verify...

File: tests/installation/ipxe_install.pm (hunk 16 above)
Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: The correct fix is this:

    $mirror_http = autoinst_url("/assets/repo/" . get_var("REPO_0_TO_INSTALL"));

File: tests/installation/ipxe_install.pm (hunk 17 above)
Comment (mdoucha) [reply tests/installation/ipxe_install.pm:161]: This is fixed in #17583. We could merge that first and rebase this on top.

File: tests/installation/ipxe_install.pm (hunk 16 above)
Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: But more importantly, why do you even add the `REPO_0_TO_INSTALL` variable to do the exact same thing as `MIRROR_HTTP`?

File: tests/installation/ipxe_install.pm (hunk 15 above)
Comment (mdoucha) [reply tests/installation/ipxe_install.pm:207]: This should be `set_bootscript unless get_var('IPXE_STATIC'); or something like that. Then set `IPXE_STATIC` whereever it's needed. Checking `DISTRI` or `OPENQA_URL` is really terrible approach.

Comment (frankenmichl) [reply tests/installation/ipxe_install.pm:207]: And please always keep in mind the baremetal support service has no persistency for it's data. If, for any reason, the service restarts, it will lose all the static configuration. Would be a better solution to always set your bootscript if you don't want to risk boot failures in that case.

Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm:207]: > And please always keep in mind the baremetal support service has no persistency for it's data. If, for any reason, the service restarts, it will lose all the static configuration. Would be a better solution to always set your bootscript if you don't want to risk boot failures in that case.

Thanks. There is no baremetal support service deployed or be accessible on O3. There is only a static iPXE menu shared by tests.

Comment (frankenmichl) [reply tests/installation/ipxe_install.pm:207]: Okay! If you need the service on o3 just reach out, I bet we can do something with that.

File: tests/installation/ipxe_install.pm (hunk 17 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm:161]: Sounds great. I remove the change from my PR.

File: tests/installation/ipxe_install.pm (hunk 15 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm:207]: Thank you, Michael. I did some research before my tests are deployed in O3 with iPXE. The dhcp server in O3 is dnsmasq, it seems that the http redirection for designated machines are not supported. All the tests have to share the same iPXE configurations. In addition, accessing internal SUSE network(IP likes 10.xx) is not allowed on O3. So finally we choosed the static iPXE menu.

File: tests/installation/ipxe_install.pm (hunk 16 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm]: > The correct fix is this:
> 
> ```
> $mirror_http = autoinst_url("/assets/repo/" . get_var("REPO_0_TO_INSTALL"));
> ```

Just tried it out, but the path is not correct. It is the job space in worker instead of the http path in openqa repo.

```
https://openqa.suse.de:20523/T8bHv8ZC0urZFd92/assets/repo/fixed/SLE-12-SP5-Server-DVD-x86_64-GM-DVD1

```

I need:
```
https://openqa.suse.de/assets/repo/fixed/SLE-12-SP5-Server-DVD-x86_64-GM-DVD1
```

Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: Ah, right. Normal `REPO_N` would be copied to the worker and then accessed via `autoinst_url()`. But the `_TO_INSTALL` suffix prevents that here.

On the other hand, when the worker is physically located far away from the OpenQA server, this approach will stress the network a lot. But we also have several mirror hosts located near workers which are supposed to be used throught the `MIRROR_HTTP` variable. Why not use that?

Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm]: `MIRROR_HTTP` is for installing the current products, fg.sle-15-SP5. but in virt tests, we need install different products, such as sle-12-sp5, sle15-sp4. For these group of tests, `REPO_0_TO_INSTALL` is introduced to flag which product will be installed.

The baremetal machines locate in different locations and VLAN.  medium in openqa server is a good place to access, and it is easier to maintain.

Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: > `MIRROR_HTTP` is for installing the current products, fg.sle-15-SP5. but in virt tests, we need install different products, such as sle-12-sp5, sle15-sp4. For these group of tests, `REPO_0_TO_INSTALL` is introduced to flag which product will be installed.

All repos available on OSD are also available on the dist mirrors, e.g. sle12-sp5 installer repo you've used as an example above (SLE-12-SP5-Server-DVD-x86_64-GM-DVD1) would be `http://[nearest dist mirror]/install/SLP/SLE-12-SP5-Server-GM/x86_64/DVD1/`. In the worst case, you can set `MIRROR_HTTP` to the OSD repo URL.

> The baremetal machines locate in different locations and VLAN. medium in openqa server is a good place to access, and it is easier to maintain. `REPO_0_TO_INSTALL=fixed/SLE-xx` has been widely used in virt tests, code change should be compitible with all existing tests.

You can set `MIRROR_HTTP` in the medium like this:

    MIRROR_HTTP=http://%MIRROR_SERVER%/install/SLP/SLE-12-SP5-Server-GM/x86_64/DVD1/

and then set `MIRROR_SERVER` to the preferred mirror hostname on each worker.

Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm]: We cannot set MIRROR_HTTP in the medium for each test suite because they are in same job group and run with same medium. All the tests are triggered at same time with same MIRROR_HTTP in one medium. ie. We install sle12sp5 in a sle15sp6 job group.  As far as I know, MIRROR_HTTP is passed in the command to trigger an entire job group(or even a product), it will override the configurations in test suite level.

Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: You can set `+MIRROR_HTTP=...` in the testsuite or job entry in the jobgroup YAML. The `+` will override the setting from medium and scheduled product.

Comment (mdoucha) [reply tests/installation/ipxe_install.pm]: See for example the ALP config in https://openqa.suse.de/admin/job_templates/498 where `+HDD_1` is used to override `HDD_1` set by `isos post`.

Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm]: Yes, `+MIRROR_HTTP=` works well in virt tests with non-developing versions. In addition, setting it in test suites is more flexable. fg. http://10.67.129.27/tests/94

File: tests/installation/ipxe_install.pm (hunk 15 above)
Comment (Julie-CAO) [reply tests/installation/ipxe_install.pm]: > This should be `set_bootscript unless get_var('IPXE_STATIC');` or something like that. Then set `IPXE_STATIC` whereever it's needed. Checking `DISTRI` or `OPENQA_URL` is really terrible approach.

I noticed that OSD began to use the same static iPXE menu on O3. `get_var('IPXE_STATIC')` is more approriate. thanks.

**Comment by Julie-CAO** [comment]
the new verification run: https://openqa.opensuse.org/tests/3537864#details

PR is ready. Please help merge.

**Comment by alice-suse** [comment]
@mdoucha  Hi Martin, there is change request from you. Is it old one or active one? I am considering to merge if it is resolved.

**Review by alice-suse — APPROVED** [review]
LGTM

File: tests/installation/ipxe_install.pm [hunk 18]
```diff
@@ -141,50 +173,54 @@ END_BOOTSCRIPT
...
+    assert_screen('o3-ipxe-menu', 210);
+    my $key = get_var('HOST_INSTALL_AUTOYAST') ? (is_kvm_host ? 'k' : 'z') : 't';
+    send_key "$key";
+    # try one more time as sometimes sending key does not take effect
+    send_key "$key" if check_screen('o3-ipxe-menu');
+    # confirm the dialog asking for user modifications to kernel, cmdline and initrd
+    # set limit to 10 in case some keys don't go through
+    for (1 .. 10) {
+        last if check_screen([qw(load-linux-kernel load-initrd)], 3);
+        send_key "ret";
```
mdoucha, line 187: Is this logic correct? This will keep sending the "Enter" keypress blindly even if no dialog shows up.

mdoucha (changes requested):

tests/installation/ipxe_install.pm
varunkojha: @mdoucha do we need a timeout?

my $max_attempts = 10;
my $timeout = 3;

for (1 .. $max_attempts) {
    if (check_screen([qw(load-linux-kernel load-initrd)], $timeout)) {
        last;
    }
    send_key "ret";
}

varunkojha: 3 secs is already there

mdoucha: No. I think the logic should look like this:

    send_key "ret" if check_screen("needle-checking-whether-the-dialog-showed-up", 30);

The jobs using the ipxe_install module usually run with 3x timeout multiplier so the key will be sent over up to 90 seconds. If the dialog doesn't show up and interactive YaST starts quickly, this'll end up advancing the installation several steps and breaking subsequent installation modules.

Julie-CAO: this logic was introduced by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17571. It is to adapt the new iPXE menu. the new iPXE menu need user's input to to start the command. "ret" will be sent until it is confirmed that the bootscript has been executed.

see this screenshot: https://openqa.opensuse.org/tests/3537864#step/ipxe_install/6

"ret" keys are required in the 'edit' mode, actually finally 3 "ret" keys are typed in this test.

```
[2023-08-31T11:28:52] no match: -0.7s, best candidate: ipxe_install-o3-ipxe-menu-20220811 (0.00)
[2023-08-31T11:28:52] >>> testapi::_check_backend_response: match=o3-ipxe-menu timed out after 0 (check_screen)
[2023-08-31T11:28:52] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:186 called testapi::check_screen
[2023-08-31T11:28:52] <<< testapi::check_screen(mustmatch=[
    "load-linux-kernel",
    "load-initrd"
  ], timeout=3)
[2023-08-31T11:28:53] no match: 8.2s, best candidate: ipxe_install-load-initrd-20220629 (0.00)
...
[2023-08-31T11:29:02] no match: -0.8s, best candidate: ipxe_install-load-linux-kernel-20220629 (0.79)
[2023-08-31T11:29:021] >>> testapi::_check_backend_response: match=load-initrd,load-linux-kernel timed out after 9 (check_screen)
[2023-08-31T11:29:02] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:187 called testapi::send_key
[2023-08-31T11:29:02] <<< testapi::send_key(key="ret", wait_screen_change=0)
[2023-08-31T11:29:03] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:186 called testapi::check_screen
[2023-08-31T11:29:03] <<< testapi::check_screen(mustmatch=[
    "load-linux-kernel",
    "load-initrd"
  ], timeout=3)
[2023-08-31T11:29:03] no match: 8.7s, best candidate: ipxe_install-load-initrd-20220629 (0.00)
...
[2023-08-31T11:29:12] no match: -0.3s, best candidate: ipxe_install-load-linux-kernel-20220629 (0.79)
[2023-08-31T11:29:12] >>> testapi::_check_backend_response: match=load-initrd,load-linux-kernel timed out after 9 (check_screen)
[2023-08-31T11:29:12] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:187 called testapi::send_key
[2023-08-31T11:29:12] <<< testapi::send_key(key="ret", wait_screen_change=0)
[2023-08-31T11:29:13] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:186 called testapi::check_screen
[2023-08-31T11:29:131] <<< testapi::check_screen(mustmatch=[
    "load-linux-kernel",
    "load-initrd"
  ], timeout=3)
...
[2023-08-31T11:29:22] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:187 called testapi::send_key
[2023-08-31T11:29:22] <<< testapi::send_key(key="ret", wait_screen_change=0)
[2023-08-31T11:29:23] tests/installation/ipxe_install.pm:209 called ipxe_install::enter_o3_ipxe_boot_entry -> tests/installation/ipxe_install.pm:186 called testapi::check_screen
[2023-08-31T11:29:23] <<< testapi::check_screen(mustmatch=[
    "load-linux-kernel",
    "load-initrd"
  ], timeout=3)
[2023-08-31T11:29:23] no match: 8.7s, best candidate: ipxe_install-load-initrd-20220629 (0.00)
[2023-08-31T11:29:25] no match: 6.7s, best candidate: ipxe_install-load-initrd-20220629 (0.00)
[2023-08-31T11:29:26] >>> testapi::_handle_found_needle: found ipxe_install-load-initrd-20220629, similarity 1.00 @ 87/386
```

waynechen55:
> 

@Julie-CAO Could you help make a verification run with 15-SP6, build of which just works as normal.

Julie-CAO:
> > 
> 
> @Julie-CAO Could you help make a verification run with 15-SP6, build of which just works as normal

sle15sp6 with autoyast would fail until https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583 is merged.

tests/installation/ipxe_install.pm
Julie-CAO: the logic works now. Let's fix it in another PR when we find it bring troubles.

waynechen55:
> > > 
> > 
> > 
> > @Julie-CAO Could you help make a verification run with 15-SP6, build of which just works as normal
> 
> sle15sp6 with autoyast would fail until #17583 is merged.

If this is the case, I do not think we need to rush merge this pull request. This should only be merged after pr#17583, otherwise, it will block all test runs.

Julie-CAO:
> If this is the case, I do not think we need to rush merge this pull request. This pull request should only be merged after [pr#17583](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583), otherwise it will block all test runs. After [pr#17583](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583) is merged, you can help do a verification with 15-SP6 to make sure merging this pull request will not block test runs.

No, it won't block sle15sp6 tests with this PR merged because the machines are still booting from PXE on OSD. 

That PR and my PR edited the same file, it would be better that my PR go in first and that PR go in next, and my needles PR got merged too. Finally we switch the workers on OSD to iPXE.

waynechen55:
> > If this is the case, I do not think we need to rush merge this pull request. This pull request should only be merged after [pr#17583](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583), otherwise it will block all test runs. After [pr#17583](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17583) is merged, you can help do a verification with 15-SP6 to make sure merging this pull request will not block test runs.
> 
> No, it won't block sle15sp6 tests with this PR merged because the machines are still booting from PXE on OSD.
> 
> That PR and my PR edited the same file, it would be better that my PR go in first and that PR go in next, and my needles PR got merged too. Finally we switch the workers on OSD to iPXE.

I think you can pick a worker later to have a try with iPXE and autoyast first before full depoyment.

Julie-CAO:
> > No, it won't block sle15sp6 tests with this PR merged because the machines are still booting from PXE on OSD.
> > That PR and my PR edited the same file, it would be better that my PR go in first and that PR go in next, and my needles PR got merged too. Finally we switch the workers on OSD to iPXE.
> 
> I think you can pick a worker later to have a try with iPXE and autoyast first before full depoyment.

Thanks. we had a pilot worker, scooter, is booting from iPXE. It is out of our worker pool now.

tests/installation/ipxe_install.pm
alice-suse: It seems to me @mdoucha comment is more about accurately handle that operating of sending ret and make test more efficient than blindly trying, which makes sense IMO. @Julie-CAO Do you think this can be improved in future PR by using some needle like needle-checking-whether-the-dialog-showed-up? To me, this is a good thing but minor and not directly related with the main scope of the PR, better to merge it first and improve that later.

If no against comment today, will merge first.

Julie-CAO: Yes, this part was about the new iPXE menu just introduced days ago. It is out of the scope of my PR. And it involves the tests in O3 only. I don't have plan to improve it in this PR.

About if it will be improved in the future, we need monitor more tests before give a solution, because 3 more needles("needle-checking-whether-the-dialog-showed-up")  from SOL console are required. If it is more efficient actually, we need more considerations.

mdoucha: OK, future PR is good enough.

mdoucha (approved):

alice-suse:
Thanks  all for the review, merge it now.

PR #17401, opened 2023-07-12
Add lvm installation test

lemon-suse:
We need add lvm installation test, so add related yaml schedule file and test module for it.

- Related ticket: https://progress.opensuse.org/issues/130054
- Needles: N/A
- Verification run: https://openqa.opensuse.org/tests/3422144#
                                https://openqa.opensuse.org/tests/3422062

### DIFF
--- a/schedule/yam/agama/agama_lvm.yaml
+++ b/schedule/yam/agama/agama_lvm.yaml
@@ -1,0 +1,9 @@
+---
+name: agama_lvm
+description: >
+  Playwright use lvm test on agama
+schedule:
+  - installation/bootloader_start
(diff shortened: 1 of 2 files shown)


schedule/yam/agama/agama_lvm_installation.yaml
```diff
@@ -0,0 +1,9 @@
+---
+name: agama_lvm_installation
```
jknphy, line 2: we can remove "installation" from all names, due to agama only used for installation.

lemon-suse: updated code and VR.

PR #17421, opened 2023-07-14
Enable existing libyui-rest-api modules in migration 

JRivrain:
1a71a734586260e13d889341da429964075309cf  Enable existing libyui-rest-api modules in migration 

See https://progress.opensuse.org/issues/131048

Updates of the yaml schedule schedule/migration/migration_offline_media_x86_64.yaml

50ffb933603d2c3123ef4090109062e674c69efe Modify InstallationSettingsPage to work in migration

The string "SSH port will be" is not visible if ssh is not enabled, so
the subroutine is_loaded_completely can fail in this case.
On the other hand "Boot Loader Type" is always present, so we can use it
in all cases.

f10b0bee1ec13dc19bd00df959552156530bfe69 Add a module to accept keyboard layout

VR https://openqa.suse.de/tests/overview?version=15-SP5&distri=sle&build=JRivrain%2Fos-autoinst-distri-opensuse%2317421

### DIFF
--- a/schedule/migration/migration_offline_media_x86_64.yaml
+++ b/schedule/migration/migration_offline_media_x86_64.yaml
@@ -24,7 +24,7 @@ schedule:
   - installation/upgrade/accept_remove_previous_repos
   - installation/addon_products_sle
   - installation/resolve_dependency_issues
-  - installation/bootloader_settings/disable_boot_menu_timeout_navigating_tabs
+  - installation/bootloader_settings/disable_boot_menu_timeout
   - installation/launch_installation
   - installation/confirm_installation
   - installation/performing_installation/perform_installation
@@ -32,4 +32,3 @@ schedule:
   - installation/reboot_after_installation
   - installation/grub_test
   - installation/first_boot
-  
--- a/schedule/migration/migration_offline_media_x86_64.yaml
+++ b/schedule/migration/migration_offline_media_x86_64.yaml
@@ -1,6 +1,6 @@
 name: migration_offline_x86_64.yaml
 description: |
-  This is for offline migraiton tests on x86_64.
+  This is for offline migration tests on x86_64.
 vars:
   ORIGIN_SYSTEM_VERSION: '%HDDVERSION%'
   UPGRADE_TARGET_VERSION: '%VERSION%'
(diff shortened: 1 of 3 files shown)


schedule/migration/migration_offline_media_x86_64.yaml
```diff
@@ -26,10 +26,10 @@ schedule:
   - installation/scc_registration
   - installation/addon_products_sle
   - installation/resolve_dependency_issues
-  - installation/installation_overview
-  - installation/disable_grub_timeout
-  - installation/start_install
-  - installation/await_install
+  - installation/bootloader_settings/disable_boot_menu_timeout
```
jknphy, line 29: and modules like registration or welcome? looks like there are more modules that could be enabled. If it is not possible, please file ticket to covert to libyui-rest-api.

JRivrain: then there is resolve_dependency_issues, we can probably get rid of it, but I was not sure, as ths is migraton, it could actually happen to be useful sometimes (?)

tinawang123:
LGTM

schedule/migration/migration_offline_media_x86_64.yaml
```diff
@@ -1,6 +1,6 @@
 name: migration_offline_x86_64.yaml
```
lemon-suse, line 1: Suggest to create a new schedule file for libyui test, ex: schedule/migration/migration_offline_media_x86_64_libyui.yaml

sofiasyria: Why? If the old schedule is used nowhere, why keep it around?

sofiasyria:
LGTM too

schedule/migration/migration_offline_media_x86_64.yaml
lemon-suse: I'd reserve the non-libyui test for comparation, anyway, Jonathon, WDYT?

JRivrain: The point of this MR is to get rid of needle-based modules because they require more maintenance, if we keep both, we need even more maintenance, so I'm rather against, also that would require an additional MR. restapi is already a proven concept.

lemon-suse: Sounds good.

lemon-suse (approved):

PR #17422, opened 2023-07-14
Fix serial terminal prompt handling

mdoucha:
The expected console prompt was switched inconsistently and could get out of sync with the actual SUT shell with the right sequence of console switching function calls. Move the automatic prompt setup logic to `susedistribution::activate_console()` and `susedistribution::console_selected()`. Also make sure that manual prompt changes using the appropriate helper function will update the console prompt expected by script_run().

- Related ticket: https://progress.opensuse.org/issues/126794
- Needles: N/A
- Verification run: https://openqa.suse.de/tests/11574354

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -761,6 +761,46 @@ sub console_nr {
     return $nr;
 }
 
+=head2 prompt_for_user
+
+  prompt_for_user($username)
+
+Returns the shell prompt that should be set for the given username
+=cut
+
+sub prompt_for_user {
+    my ($self, $username) = @_;
+
+    return $username eq 'root' ? '# ' : '$ ';
+}
+
+=head2 get_console_info
+
+  get_console_info($console)
+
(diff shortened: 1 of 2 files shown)


mdoucha:
This kind of problems really shows how utterly broken the whole console subsystem is. There are 4 layers of abstraction across 2 different projects and shell handling is spread across all of them. In any sane design, 3 of those layers shouldn't even know whether there is an interactive shell running on the SUT.

lib/susedistribution.pm
```diff
@@ -761,6 +761,46 @@ sub console_nr {
     return $nr;
 }
 
+=head2 get_console_info
```
pdostal:
```suggestion
=head2 prompt_for_user
```

pdostal (approved):

schlad (approved):

b10n1k (approved):

okurz (approved):

PR #17424, opened 2023-07-14
Remove workaround for bsc1202234 and bsc121197

JRivrain:
This commit removes the workaround in just few places, but there are more workarounds in place, it will require another PR.

VR https://openqa.suse.de/tests/11567313


lib/registration.pm
```diff
@@ -900,11 +900,10 @@ sub scc_deregistration {
             add_suseconnect_product('sle-module-python2');
         }
         my $deregister_ret = script_run('SUSEConnect --de-register --debug > /tmp/SUSEConnect.debug 2>&1', 300);
-        if (defined $deregister_ret and $deregister_ret) {
-            # Cleanup and upload logs in case of de-registration failure.
-            record_soft_failure "bsc#1211970 - SUSEConnect --de-register returned code $deregister_ret";
+        if ($deregister_ret) {
+            # See git blame for previous workarounds.
             upload_logs "/tmp/SUSEConnect.debug";
-            assert_script_run('SUSEConnect --cleanup', 200);
```
lemon-suse, line 907: What do you think to reserve the cleanup action here?

JRivrain: Well that was part of the workaround, it should never be needed if the bug is fixed.

lemon-suse:
Suggest to run move VRs on other versions such as 15SP4, 15SP5 and 12SP5.

JRivrain:
> Suggest to run move VRs on other versions such as 15SP4, 15SP5 and 12SP5. Aha, I think we'd better to have a ticket to track this, there are many profiles need to be updated for this.

Right, updated VRs

lemon-suse:
It seems all the VRs are ok except this:
https://openqa.suse.de/tests/11669982#step/scc_deregistration/13 For me it seems the design for the test is wrong, MEDIA_UPGRADE=1 means the system will be de-registered before and after migration, so it doesn't need to add scc_deregistration after migration and it will cause de-register fail for not registered. 

Please clone the job and unset the SCC_DEREGISTER and run a test to verify this.

JRivrain:
> Please clone the job and unset the SCC_DEREGISTER and run a test to verify this.

Sorry I forgot to comment, that worked https://openqa.suse.de/tests/11741247

PR #17429, opened 2023-07-17
Prepare two synchronized machines to test PCW

ilausuch:
PCW is executed in a microos and is setup using ansible. This PR creates the environment for two VMs to test the ansible playbooks and selenium tests.

The microos is ready with all the needed packages.

- Related ticket: https://progress.opensuse.org/issues/130144
- Verification run: 
https://openqa.suse.de/tests/11624244
https://openqa.suse.de/tests/11624245

### DIFF
--- a/data/publiccloud/pcw/sshd_config
+++ b/data/publiccloud/pcw/sshd_config
@@ -1,0 +1,2 @@
+PubkeyAuthentication yes
+PermitRootLogin yes
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -13,6 +13,7 @@ use base 'Exporter';
 use Exporter;
 use main_common;
 use main_containers qw(load_container_tests is_container_test);
+use main_publiccloud 'load_ansible_tests';
 use testapi qw(check_var get_required_var get_var set_var);
 use version_utils;
 use utils;
@@ -278,25 +279,19 @@ sub load_slem_on_pc_tests {
 }
 
 sub load_tests {
-    # SLEM on PC
     if (is_public_cloud()) {
-        load_slem_on_pc_tests;
+        if (check_var('PUBLIC_CLOUD_ANSIBLE_CLIENT', 1) || check_var('PUBLIC_CLOUD_ANSIBLE_TARGET', 1)) {
+            # SLE-Micro Ansible target
+            load_ansible_tests();
+        } else {
+            # SLEM on PC
+            load_slem_on_pc_tests();
+        }
         return 1;
     }
 
     if (is_kernel_test()) {
         load_kernel_tests;
-        return 1;
-    }
-
-    if (check_var('PUBLIC_CLOUD_ANSIBLE_CLIENT', 1)) {
-        loadtest 'boot/boot_to_desktop';
-        loadtest 'publiccloud/ansible_client';
-        return 1;
-    }
-    elsif (check_var('PUBLIC_CLOUD_ANSIBLE_TARGET', 1)) {
-        loadtest 'microos/disk_boot';
-        loadtest 'publiccloud/ansible_target';
         return 1;
     }
 
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -289,6 +289,17 @@ sub load_tests {
         return 1;
     }
 
+    if (check_var('PUBLIC_CLOUD_ANSIBLE_CLIENT', 1)) {
+        loadtest 'boot/boot_to_desktop';
+        loadtest 'publiccloud/ansible_client';
+        return 1;
+    }
+    elsif (check_var('PUBLIC_CLOUD_ANSIBLE_TARGET', 1)) {
+        loadtest 'microos/disk_boot';
+        loadtest 'publiccloud/ansible_target';
+        return 1;
+    }
+
     if (get_var('REMOTE_TARGET')) {
         load_remote_target_tests;
         return 1;
--- a/lib/main_publiccloud.pm
+++ b/lib/main_publiccloud.pm
@@ -19,6 +19,7 @@ require bmwqemu;
 
 our @EXPORT = qw(
   load_publiccloud_tests
+  load_ansible_tests
 );
 
 sub load_maintenance_publiccloud_tests {
@@ -179,6 +180,20 @@ sub load_publiccloud_download_repos {
     loadtest 'publiccloud/download_repos';
     loadtest 'shutdown/shutdown';
 }
+
+sub load_ansible_tests {
+    if (check_var('PUBLIC_CLOUD_ANSIBLE_CLIENT', 1)) {
+        loadtest 'boot/boot_to_desktop';
+        loadtest 'publiccloud/ansible_client';
+        return 1;
+    }
+    elsif (check_var('PUBLIC_CLOUD_ANSIBLE_TARGET', 1)) {
+        loadtest 'microos/disk_boot';
+        loadtest 'publiccloud/ansible_target';
+        return 1;
+    }
+}
+
 
 =head2 load_publiccloud_tests
 
--- a/lib/network_utils.pm
+++ b/lib/network_utils.pm
@@ -60,7 +60,7 @@ sub iface {
     my ($quantity) = @_;
     $quantity ||= 1;
     # bonding_masters showing up in ppc64le jobs in 15-SP5: bsc#1210641
-    return script_output('ls /sys/class/net/ | grep -v lo | grep -v bonding_masters | head -' . $quantity);
+    return script_output('ls /sys/class/net/ | grep -v -e lo -e bonding_masters -e docker | head -' . $quantity);
 }
 
 =head2 can_upload_logs
--- a/tests/publiccloud/ansible_client.pm
+++ b/tests/publiccloud/ansible_client.pm
@@ -1,0 +1,47 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: prepare the VM to sync with the ansible_target
+# Maintainer: qa-c@suse.de
+
+use base "consoletest";
+use serial_terminal 'select_serial_terminal';
+use strict;
+use warnings;
+use testapi;
+use mmapi;
+use utils qw(zypper_call);
+use network_utils qw(setup_static_network);
+use lockapi;
+
+sub run {
+    my ($self, $args) = @_;
+    select_serial_terminal;
+
+    record_info('system', script_output('cat /etc/os-release'));
+    setup_static_network(ip => '10.0.2.15/15', gw => '10.0.2.2');
+    record_info('ip', script_output('ip a'));
+    record_info('route', script_output('ip r'));
+    assert_script_run('echo "10.0.2.20  microos" >> /etc/hosts');
+    zypper_call('in -y iputils git');
+
+    assert_script_run('mkdir /root/.ssh');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/id_rsa > /root/.ssh/id_rsa');
+    assert_script_run('chmod 600 /root/.ssh/id_rsa');
+
+    my $children = get_children();
+    my $child_id = (keys %$children)[0];
+    mutex_wait('target_is_ready', $child_id);
+
+    # Testing target is accessible
+    assert_script_run('ping -c 1 microos');
+    assert_script_run('ssh -v -o StrictHostKeyChecking=accept-new root@microos cat /etc/os-release');
+
+    mutex_create 'job_completed';
+
+    wait_for_children;
+}
+
+1;
--- a/tests/publiccloud/ansible_client.pm
+++ b/tests/publiccloud/ansible_client.pm
@@ -8,26 +8,32 @@
 
 use base "consoletest";
 use serial_terminal 'select_serial_terminal';
-use strict;
+use transactional qw(trup_call process_reboot);
+#use strict;
 use warnings;
 use testapi;
-use mmapi;
-use utils qw(zypper_call);
-use network_utils qw(setup_static_network);
 use lockapi;
+use mmapi;
+use mm_network qw(setup_static_mm_network);
+use utils qw(zypper_call);
 
 sub run {
     my ($self, $args) = @_;
     select_serial_terminal;
 
     record_info('system', script_output('cat /etc/os-release'));
-    setup_static_network(ip => '10.0.2.15/15', gw => '10.0.2.2');
+    # setup_static_network(ip => '10.0.2.15/15', gw => '10.0.2.2');
+    # record_info('ip', script_output('ip a'));
+    # record_info('route', script_output('ip r'));
+    # assert_script_run('echo "10.0.2.20  microos" >> /etc/hosts');
+    # zypper_call('in -y iputils git');
+
+    setup_static_mm_network('10.0.2.15/15');
+
     record_info('ip', script_output('ip a'));
     record_info('route', script_output('ip r'));
-    assert_script_run('echo "10.0.2.20  microos" >> /etc/hosts');
-    zypper_call('in -y iputils git');
+    script_run('ping -c 1 download.suse.de');
 
-    assert_script_run('mkdir /root/.ssh');
     assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/id_rsa > /root/.ssh/id_rsa');
     assert_script_run('chmod 600 /root/.ssh/id_rsa');
 
--- a/tests/publiccloud/ansible_target.pm
+++ b/tests/publiccloud/ansible_target.pm
@@ -1,0 +1,48 @@
+# SUSE's openQA tests
+#
+# Copyright 2023 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: prepare the VM to grant access to ansible_client
+# Maintainer: qa-c@suse.de
+
+use base "consoletest";
+use serial_terminal 'select_serial_terminal';
+use transactional qw(trup_call process_reboot);
+use strict;
+use warnings;
+use testapi;
+use lockapi;
+use mm_network qw(setup_static_mm_network);
+use utils qw(zypper_call);
+use Utils::Systemd qw(systemctl);
+
+sub run {
+    my ($self, $args) = @_;
+    select_serial_terminal;
+    record_info('system', script_output('cat /etc/os-release'));
+    record_info('device', script_output('nmcli -t device'));
+
+    setup_static_mm_network('10.0.2.20/15');
+
+    record_info('ip', script_output('ip a'));
+    record_info('route', script_output('ip r'));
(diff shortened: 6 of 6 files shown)


tests/publiccloud/pcw_client.pm
```diff
@@ -0,0 +1,37 @@
...
+    my ($self, $args) = @_;
+    select_serial_terminal;
+
+    record_info('system', script_output('cat /etc/os-release'));
+    setup_static_network(ip => '10.0.2.15/15', gw => '10.0.2.2', iface => 'eth0');
+    assert_script_run('echo "10.0.2.20  microos" >> /etc/hosts');
+    zypper_call('in -y iputils git');
+
+    assert_script_run('mkdir /root/.ssh');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/id_rsa > /root/.ssh/id_rsa');
```
pdostal, line 21: Please generate new keys and put them into `data/publiccloud/pcw/`.

data/publiccloud/pcw/sshd_config
```diff
@@ -0,0 +1,3 @@
+PubkeyAuthentication yes
+PermitRootLogin yes
+AuthorizedKeysFile .ssh/authorized_keys
```
pdostal, line 3: We use keys and that should be permitted by default. This file is not needed.

tests/publiccloud/pcw_client.pm
```diff
@@ -0,0 +1,37 @@
...
+    assert_script_run('chmod 600 /root/.ssh/id_rsa');
+
+    my $children = get_children();
+    my $child_id = (keys %$children)[0];
+    mutex_wait('target_is_ready', $child_id);
+
+    # Testing target is accessible
+    assert_script_run('ping -c 1 microos');
+    assert_script_run('ssh -v -o StrictHostKeyChecking=accept-new -o KexAlgorithms=ecdh-sha2-nistp521 root@microos hostname');
```
pdostal, line 30: Interesting solution, but I'd still prefer:
```suggestion
    assert_script_run('ssh-keyscan microos >> ~/.ssh/known_hosts');
    assert_script_run('ssh root@microos hostname');
```

tests/publiccloud/pcw_target.pm
```diff
@@ -0,0 +1,39 @@
...
+use lockapi;
+
+sub run {
+    my ($self, $args) = @_;
+    select_serial_terminal;
+    record_info('system', script_output('cat /etc/os-release'));
+    record_info('device', script_output('nmcli -t device'));
+
+    assert_script_run('nmcli con add type ethernet con-name "static-ip" ifname ens4 ipv4.method manual ipv4.addresses 10.0.2.20/15 gw4 10.0.2.2');
+    assert_script_run('nmcli con mod static-ip ipv4.dns "10.136.53.53,10.136.53.54,10.100.2.10"');
```
pdostal, line 16: There is `mm_network::configure_static_ip`, `mm_network::configure_default_gateway` and `mm_network::configure_static_dns`.

tests/publiccloud/pcw_target.pm
```diff
@@ -0,0 +1,39 @@
...
+    script_run('ping -c 1 10.0.2.15');
+    script_run('ping -c 1 download.suse.de');
+
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/authorized_keys >> /root/.ssh/authorized_keys');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/publiccloud/pcw/sshd_config >/etc/ssh/sshd_config');
+
+    assert_script_run('zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Tumbleweed/ SUSE_CA');
+    assert_script_run('zypper --gpg-auto-import-keys ref');
+    assert_script_run('transactional-update -n -c pkg install ca-certificates-suse htop iftop iotop atop telnet nmap jq git rsync wget parted tcpdump screen sqlite3 fortune iputils nginx dehydrated dehydrated-nginx podman podman-cni-config cockpit cockpit-podman toolbox python3-selinux');
+    process_reboot(trigger => 1);
```
pdostal, line 28: This is already done in Ansible.

lib/network_utils.pm
```diff
@@ -40,7 +40,8 @@ sub setup_static_network {
     $args{silent} //= 0;
     configure_static_dns(get_host_resolv_conf(), silent => $args{silent});
     assert_script_run('echo default ' . $args{gw} . ' - - > /etc/sysconfig/network/routes');
-    my $iface = iface();
+    $args{iface} //= iface();
```
asmorodskyi, line 43: usually `iface()` function perfectly detecting available interfaces

tests/publiccloud/pcw_target.pm
asmorodskyi: yes I would expect only python3 install here , all other things should happen via ansible script if we currently missing something we need to add this into ansible script not here

tests/publiccloud/pcw_client.pm
ilausuch: For somereason I had to add this option KexAlgorithms=ecdh-sha2-nistp521, else this doesn't work
I tried to add the kex algoritms on sshd_config, but seems is not working
```
KexAlgorithms ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group14-sha1,diffie-hellman-group-exchange-sha256
```

lib/network_utils.pm
ilausuch: I remember that I had to specify this, but was long time ago, I can test it without this change, to see if the problem appears again

tests/publiccloud/pcw_target.pm
ilausuch: do you mean python3-selinux? 
why python3 shouldn't be in the other repo?

tests/publiccloud/pcw_client.pm
pdostal: That should really not be needed. Please try without it with `ssh -v` and let me know the output. Or we can have a call about it.

tests/publiccloud/pcw_target.pm
asmorodskyi: > Try to remove this as a whole.

no we do need to install python on target ( it is not available by default on MicroOS and requirement by ansible ) 

> do you mean python3-selinux?

no I mean that exception installation of python package target should not be changed by the test at all , no addition of extra repos , no extra packages installation . the whole point of this test is to check that **on clean** VM this ansible playbook can create workable system. With set of this `zypper addrepo ...` / `zypper in ...`  you removing "on clean VM"

pdostal: Anton, there is no Python in the list right now. Isn't Python on MicroOS by default? And if it's not, then Ivan will find out latest.

grisu48: python3 is not present by default I think. However this things are moving rather quickly, it's anyways good to install it explicitly, just to ensure it's present.

grisu48:
Do we really need selenium? Wouldn't it be sufficient to fetch the `/health.json` from the newly installed PCW instance and check if it returns the expected status object?

```json
{"status": "ok"}
```

asmorodskyi:
> Do we really need selenium? 

1. I think this discussion goes far away from this PR . this is something what we will decide later on 
2. depending what expected coverage for the test . I think at the end we will have several tests based on that some of them will concentrate covering ansible playbook and they won't need selenium . Others will aim give better coverage of PCW functionality and here we might use selenium

lib/network_utils.pm
```diff
@@ -60,7 +60,7 @@ sub iface {
     my ($quantity) = @_;
     $quantity ||= 1;
     # bonding_masters showing up in ppc64le jobs in 15-SP5: bsc#1210641
-    return script_output('ls /sys/class/net/ | grep -v lo | grep -v bonding_masters | head -' . $quantity);
+    return script_output('ls /sys/class/net/ | grep -v lo | grep -v bonding_masters | grep -v docker | head -' . $quantity);
```
asmorodskyi, line 63: @cfconrad  WDYT about this ?

ricardobranco777:
> Do we really need selenium? Wouldn't it be sufficient to fetch the `/health.json` from the newly installed PCW instance and check if it returns the expected status object?
> 
> ```json
> {"status": "ok"}
> ```

I do believe that a Selenium test would fit nicely here, because the Selenium test in GitHub interacts directly with Django, and in this case we have Nginx + Django.

@ilausuch (or anyone planning to add such a test):
We can use the approach I introduced in HAWK that uses a Docker image for this (to avoid installing these dependencies on the OS):

- Example Dockerfile: The image is defined here: https://github.com/ClusterLabs/hawk/blob/master/e2e_test/Dockerfile
- Example code on how to run it in openQA: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/ha/check_hawk.pm
- Basic Selenium test for `pcw`: https://github.com/SUSE/pcw/blob/master/tests/test_webui.py

tests/publiccloud/pcw_client.pm
pdostal: I still don't understand how the systems are installed. If there is some 'normal user' you can SCP the public key via password authorized SSH.

lib/network_utils.pm
cfconrad: I think it's ok. Getting a docker interface here isn't expected

tests/publiccloud/ansible_client.pm
```diff
@@ -0,0 +1,38 @@
...
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/id_rsa > /root/.ssh/id_rsa');
+    assert_script_run('chmod 600 /root/.ssh/id_rsa');
+
+    my $children = get_children();
+    my $child_id = (keys %$children)[0];
+    mutex_wait('target_is_ready', $child_id);
+
+    # Testing target is accessible
+    assert_script_run('ping -c 1 microos');
+    # assert_script_run('ssh -v -o StrictHostKeyChecking=accept-new -o KexAlgorithms=ecdh-sha2-nistp521 root@microos hostname');
```
asmorodskyi, line 30: please remove commented code

lib/network_utils.pm
ilausuch: done

tests/publiccloud/ansible_target.pm
```diff
@@ -0,0 +1,40 @@
...
+    record_info('route', script_output('ip r'));
+    script_run('ping -c 1 10.0.2.15');
+    script_run('ping -c 1 download.suse.de');
+
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/authorized_keys >> /root/.ssh/authorized_keys');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/publiccloud/pcw/sshd_config >/etc/ssh/sshd_config');
+
+    assert_script_run('zypper ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Tumbleweed/ SUSE_CA');
+    assert_script_run('zypper --gpg-auto-import-keys ref');
+    assert_script_run('transactional-update -n -c pkg install ca-certificates-suse htop iftop iotop atop telnet nmap jq git rsync wget parted tcpdump screen sqlite3 fortune iputils nginx dehydrated dehydrated-nginx podman podman-cni-config cockpit cockpit-podman toolbox python3-selinux');
```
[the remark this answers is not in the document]
ilausuch: done

tests/publiccloud/pcw_client.pm
ilausuch: yes, works, I don't know why this didn't work in the past tests

tests/publiccloud/ansible_client.pm
ilausuch: done

tests/publiccloud/ansible_target.pm
```diff
@@ -0,0 +1,41 @@
...
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/publiccloud/pcw/sshd_config >/etc/ssh/sshd_config');
+
+    zypper_call('ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Tumbleweed/ SUSE_CA');
+    zypper_call('--gpg-auto-import-keys ref');
+    trup_call('pkg install ca-certificates-suse htop iftop iotop atop telnet nmap jq git rsync wget parted tcpdump screen sqlite3 fortune iputils nginx dehydrated dehydrated-nginx podman podman-cni-config cockpit cockpit-podman toolbox python3-selinux');
+    process_reboot(trigger => 1);
+
+    script_run('systemctl restart sshd');
+    sleep 10;
+    assert_script_run('systemctl status sshd');
```
ricardobranco777, line 34:
```suggestion
    systemctl('restart sshd');
    systemctl('status sshd');

```

You can `use Utils::Systemd 'systemctl';`

ricardobranco777 (approved):
LGTM. It's nice that this can be reused for other projects, like our Docker Registry, if it ever makes it.

tests/publiccloud/ansible_target.pm
```diff
@@ -0,0 +1,41 @@
...
+
+    record_info('ip', script_output('ip a'));
+    record_info('route', script_output('ip r'));
+    script_run('ping -c 1 10.0.2.15');
+    script_run('ping -c 1 download.suse.de');
+
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/authorized_keys >> /root/.ssh/authorized_keys');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/publiccloud/pcw/sshd_config >/etc/ssh/sshd_config');
+
+    zypper_call('ar http://download.suse.de/ibs/SUSE:/CA/openSUSE_Tumbleweed/ SUSE_CA');
```
asmorodskyi, line 28: the same  with the cert repo , please remove it

asmorodskyi (approved):
LGTM

lib/main_publiccloud.pm
```diff
@@ -207,6 +207,12 @@ sub load_publiccloud_tests {
             load_publiccloud_download_repos();
         } elsif (get_var('PUBLIC_CLOUD_QAM')) {
             load_maintenance_publiccloud_tests();
+        } elsif (get_var('PUBLIC_CLOUD_PCW_CLIENT')) {
+            loadtest 'boot/boot_to_desktop';
+            loadtest 'publiccloud/ansible_client';
+        } elsif (get_var('PUBLIC_CLOUD_PCW_TARGET')) {
+            loadtest 'microos/disk_boot';
```
asmorodskyi, line 214: it won't work like this see line 201

asmorodskyi:
```perl
if (check_var('PUBLIC_CLOUD_PREPARE_TOOLS', 1)) {
        load_create_publiccloud_tools_image();
    }
    elsif (check_var('PUBLIC_CLOUD_TOOLS_CLI', 1)) {
        load_publiccloud_cli_tools();
    }
elsif (get_var('PUBLIC_CLOUD_PCW_TARGET')) {
            loadtest 'microos/disk_boot';
            loadtest 'publiccloud/ansible_target';
}
    else {
        loadtest 'boot/boot_to_desktop';
        if (get_var('PUBLIC_CLOUD_MIGRATION')) {
            my $args = OpenQA::Test::RunArgs->new();
            loadtest('publiccloud/upload_image', run_args => $args);
            loadtest('publiccloud/migration', run_args => $args);
        } elsif (check_var('PUBLIC_CLOUD_DOWNLOAD_TESTREPO', 1)) {
            load_publiccloud_download_repos();
        } elsif (get_var('PUBLIC_CLOUD_QAM')) {
            load_maintenance_publiccloud_tests();
        } elsif (get_var('PUBLIC_CLOUD_PCW_CLIENT')) {
             loadtest 'publiccloud/ansible_client';
}
```

ilausuch: Yes, understood. In my wayt a previous loadtest is added

data/publiccloud/pcw/sshd_config
grisu48: > Neither of those options is needed. We log in to Larry as `root` via ssh key and look:
> 
> ```shell
> publiccloud-ng:~ # cat /etc/ssh/sshd_config
> cat: /etc/ssh/sshd_config: No such file or directory
> ```

They are needed. Look at `/etc/ssh/sshd_config.d/sshd_config.conf`, after usrmerge one should not edit `/etc/ssh/sshd_config` directly but place configuration files under `/etc/ssh/sshd_config.d/`.

Root access is denied by default unless explicitly allowed by a configuration file.

pdostal: OK. I din't know that. I'm sorry.

lib/main_micro_alp.pm
```diff
@@ -264,17 +264,27 @@ sub load_journal_check_tests {
...
-    if (get_var("PUBLIC_CLOUD_CONTAINERS")) {
-        load_container_tests() if is_container_test;
+    if (check_var('PUBLIC_CLOUD_ANSIBLE_CLIENT', 1)) {
+        loadtest 'boot/boot_to_desktop';
+        loadtest 'publiccloud/ansible_client';
+    }
+    elsif (check_var('PUBLIC_CLOUD_ANSIBLE_TARGET', 1)) {
+        loadtest 'microos/disk_boot';
+        loadtest 'publiccloud/ansible_target';
+    }
```
pdostal, line 274: I would move this out of `load_slem_on_pc_tests()`.

tests/publiccloud/ansible_client.pm
```diff
@@ -0,0 +1,47 @@
...
+
+    record_info('system', script_output('cat /etc/os-release'));
+    setup_static_network(ip => '10.0.2.15/15', gw => '10.0.2.2');
+    record_info('ip', script_output('ip a'));
+    record_info('route', script_output('ip r'));
+    assert_script_run('echo "10.0.2.20  microos" >> /etc/hosts');
+    zypper_call('in -y iputils git');
+
+    assert_script_run('mkdir /root/.ssh');
+    assert_script_run('curl -f -v ' . autoinst_url . '/data/slenkins/ssh/id_rsa > /root/.ssh/id_rsa');
```
pdostal, line 31: Please either allow password login and generate the key in the test or at least generate our own key in `data/publiccloud/`

grisu48: Let's please not have pre-generated keys as they always expose a certain security risk. Unless really really really necessary, let's generate per-test key pairs.

asmorodskyi: > Let's please not have pre-generated keys as they always expose a certain security risk. Unless really really really necessary, let's generate per-test key pairs.

IMHO it is too much . we can not apply common security rules into short living test VM running in internal network . Generating per-test key pairs coming with very high price -> in case debug is needed you can access VM only via VNC

asmorodskyi: if we really want to care about security of test VM running on internal network let's hide ssh key in worker config so it will be protected on same level as some other info which has much much more value

ilausuch:
We decided to move in an other direction. See the ticket

PR #17431, opened 2023-07-17
Create autoyast installation for textmode in yast job group

manfredi:
- Description: Create autoyast installation for textmode in _yast_ job group
- Related ticket: https://progress.opensuse.org/issues/130985
- Verification run:
  + [overview](https://openqa.suse.de/tests/overview?distri=sle&build=manfredi%2Fos-autoinst-distri-opensuse%23issues-130985&version=15-SP5)
  + [create_hdd_textmode_autoyast@ppc64le](https://openqa.suse.de/tests/overview?distri=sle&version=15-SP5&build=102.1&groupid=96)
- Related [MR#519](https://gitlab.suse.de/qsf-y/qa-sle-functional-y/-/merge_requests/519)

### DIFF
--- a/data/yam/autoyast/create_hdd_textmode_aarch64.xml
+++ b/data/yam/autoyast/create_hdd_textmode_aarch64.xml
@@ -1,0 +1,442 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <bootloader t="map">
+    <global t="map">
+      <cpu_mitigations>auto</cpu_mitigations>
+      <gfxmode>auto</gfxmode>
+      <hiddenmenu>false</hiddenmenu>
+      <os_prober>false</os_prober>
+      <secure_boot>true</secure_boot>
+      <terminal>gfxterm</terminal>
+      <timeout t="integer">-1</timeout>
+      <update_nvram>true</update_nvram>
+      <xen_kernel_append>vga=gfx-1024x768x16</xen_kernel_append>
+    </global>
+    <loader_type>grub2-efi</loader_type>
+  </bootloader>
+  <firewall t="map">
+    <default_zone>public</default_zone>
+    <enable_firewall t="boolean">true</enable_firewall>
+    <log_denied_packets>off</log_denied_packets>
+    <start_firewall t="boolean">true</start_firewall>
+    <zones t="list">
+      <zone t="map">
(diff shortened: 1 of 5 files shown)


lemon-suse:
In general, LGTM, just for the VR of ppc64le, it should be on ppc64le-spvm and please update MR for it also.

data/yam/autoyast/create_hdd_textmode_aarch64.xml
```diff
@@ -0,0 +1,442 @@
...
+          <create_subvolumes t="boolean">true</create_subvolumes>
+          <filesystem t="symbol">btrfs</filesystem>
+          <format t="boolean">true</format>
+          <mount>/</mount>
+          <mountby t="symbol">uuid</mountby>
+          <partition_id t="integer">131</partition_id>
+          <partition_nr t="integer">2</partition_nr>
+          <quotas t="boolean">true</quotas>
+          <resize t="boolean">false</resize>
+          <size>29928456192</size>
```
rfan1, line 179: How about set the size to `max`?

lemon-suse: Yes, set to max can avoid risk here than hardcode.

lemon-suse: Yes, only for root folder ('/').

data/yam/autoyast/create_hdd_textmode_ppc64le.xml
```diff
@@ -0,0 +1,467 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
lemon-suse, line 1: It is different for power kvm and power VM, so it is better to update the name, such as:
data/yam/autoyast/create_hdd_textmode_powervm.xml

rfan1: > It is different for power kvm and power VM, so it is better to update the name, such as: data/yam/autoyast/create_hdd_textmode_powervm.xml

As I know, Yast team will not test power + kvm backend any more, then you don't need to change the name.

or if kvm is still tested, I would suggest keeping a single autoyast file for kvm backend/spvm/powerhmc, there should be 2 differences for kvm backend:

1.  set the root disk to `/dev/vda`
2.  we don't need to enable sshd service in firewalld

lemon-suse: For power kvm we still need it for yam migration test. For this ticket, we have a power vm profile is enough for current yast job group test. So I think we can rename it to powervm for now, for power kvm we can work on it on another ticket, and it should be for migration test on SLE12SP5.

lemon-suse (approved):

rfan1 (approved):
LGTM

PR #17438, opened 2023-07-18
Public Cloud: Make sure the openqa_var_SERVER tag does not contain slash

pdostal:
- Verification run: [openqa.suse.de](https://openqa.suse.de/tests/11609478), [pdostal-server.suse.cz](https://pdostal-server.suse.cz/tests/4712)

### DIFF
--- a/lib/publiccloud/provider.pm
+++ b/lib/publiccloud/provider.pm
@@ -623,11 +623,14 @@ c<{"key": "value"}> where c<value> must be a string.
 sub terraform_param_tags
 {
     my ($self) = @_;
+    my $openqa_var_server = get_var('OPENQA_URL', get_var('OPENQA_HOSTNAME'));
+    # Remove the http:// https:// and/or the slash at the end
+    $openqa_var_server =~ s@^https?://|/$@@gm;
     my $tags = {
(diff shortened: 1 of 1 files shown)


ricardobranco777 (approved):
LGTM

lib/publiccloud/provider.pm
```diff
@@ -623,11 +623,13 @@ c<{"key": "value"}> where c<value> must be a string.
 sub terraform_param_tags
 {
     my ($self) = @_;
+    my $openqa_var_server = get_var('OPENQA_URL', get_var('OPENQA_HOSTNAME'));
+    $openqa_var_server =~ s@^https?://|/$@@gm;
```
grisu48, line 627: Please add a comment here for our future selves that will be thankful for it!

grisu48 (approved):

lib/publiccloud/provider.pm
asmorodskyi, line 627: keep in mind that `$openqa_var_server` can be just FQDN . I assume this will not be a problem here , right ?

pdostal: I hope not.
image

PR #17446, opened 2023-07-19
lib/main_common: CR-DVD does not need_clear_repos

Vogtinator:
This flavor uses published repos during testing.

+ Bonus improvement to a comment I had lying around in my checkout but didn't want to open a separate PR for.

- Related ticket: https://progress.opensuse.org/issues/131186
- Verification run: https://openqa.opensuse.org/tests/3440543


Vogtinator:
This adds a condition for a very specific FLAVOR and is a noop for other cases, so I'll just hit the merge button without waiting for a review this time. That makes it easier to see whether this change was enough or some other case also needs to be taken care of.

PR #17449, opened 2023-07-19
Run prune after all containers are stopped

mloviska:
`podman-prune` might run while containers are stopping.

- ticket: https://progress.opensuse.org/issues/133031
- Verification run: [sle-micro-5.3-BYOS-Azure-x86_64-Build0468-slem_containers@64bit](http://kepler.suse.cz/tests/21305#)

### DIFF
--- a/lib/containers/engine.pm
+++ b/lib/containers/engine.pm
@@ -321,6 +321,10 @@ sub cleanup_system_host {
     my ($self, $assert) = @_;
     $assert //= 1;
     $self->_engine_assert_script_run("ps -q | xargs -r " . $self->runtime . " stop", 180);
+
+    # all containers should be stopped before running prune
+    # https://github.com/containers/podman/issues/19038
+    $self->_engine_assert_script_run("rm --force --all", 120);
     $self->_engine_assert_script_run("system prune -a -f", 300);
 
     if ($assert) {


mloviska:
> The solution is to run `podman rm -f`, that will block: [containers/podman#19038 (comment)](https://github.com/containers/podman/issues/19038#issuecomment-1612638308)

Thanks, I have missed this comment. I will update the PR

lib/containers/engine.pm
```diff
@@ -321,6 +321,20 @@ sub cleanup_system_host {
...
+    # https://github.com/containers/podman/issues/19038
+    my $count = 5;
+    my $no_container;
+    do {
+        $no_container = $self->_engine_script_output('ps --noheading');
+        $count--;
+
+        if ($no_container) {
+            record_soft_failure("bsc#1212825 - podman system prune sometimes fails or podman stop doesn't wait for containers to stop");
+        }
```
ricardobranco777, line 335:
```suggestion
        if ($no_container) {
            record_soft_failure("bsc#1212825 - podman system prune sometimes fails or podman stop doesn't wait for containers to stop");
            break;
        }
```

We need a break. :)

lib/containers/engine.pm
```diff
@@ -321,6 +321,20 @@ sub cleanup_system_host {
...
+    # all containers should be stopped before running prune
+    # https://github.com/containers/podman/issues/19038
+    my $count = 5;
+    my $no_container;
+    do {
+        $no_container = $self->_engine_script_output('ps --noheading');
+        $count--;
+
+        if ($no_container) {
+            record_soft_failure("bsc#1212825 - podman system prune sometimes fails or podman stop doesn't wait for containers to stop");
```
ricardobranco777, line 334:
```suggestion
            record_soft_failure("bsc#1212825 - podman system prune sometimes fails or podman stop doesn't wait for containers to stop");
            break;
```

_Gimme a break_. XD

mloviska: :D actually, the idea was to make sure all are stopped. So the loop would if the variable is empty. Anyway, just added `rm`

ricardobranco777 (approved):
LGTM

PR #17369, opened 2023-07-06
Enhance how we open/close YaST modules in containers to detect potential problems

tinawang123:
Enhance how we open/close YaST modules in containers to detect potential problems

- Related ticket: https://progress.opensuse.org/issues/131069
- Verification run: https://openqa.suse.de/tests/11615686#step/control_center/4

PR #17450, opened 2023-07-20
Enhance how we open/close YaST modules in containers to detect potential problems

tinawang123:
Enhance how we open/close YaST modules in containers to detect potential problems

-  Related ticket: https://progress.opensuse.org/issues/131069
-  Verification run: https://openqa.suse.de/tests/11615686#step/control_center/4
   https://openqa.suse.de/tests/11623715

### DIFF
--- a/lib/y2_module_consoletest.pm
+++ b/lib/y2_module_consoletest.pm
@@ -11,24 +11,32 @@ our @EXPORT_OK = qw(yast2_console_exec);
 
 sub yast2_console_exec {
     my %args = @_;
-    die "Yast2 module has not been found among function arguments!\n" unless (defined($args{yast2_module}));
-    my $y2_start = y2_module_basetest::with_yast_env_variables($args{extra_vars}) . ' yast2 ';
-    my $module_name = 'yast2-' . $args{yast2_module} . '-status';
-    $y2_start .= (defined($args{yast2_opts})) ?
-      $args{yast2_opts} . ' ' . $args{yast2_module} :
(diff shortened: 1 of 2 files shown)


tinawang123:
Related PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17369

lib/y2_module_consoletest.pm
```diff
@@ -11,24 +11,32 @@ our @EXPORT_OK = qw(yast2_console_exec);
...
+    }
+    if (defined($args{podman})) {
+        $cmd_start = y2_module_basetest::with_yast_env_variables($args{extra_vars}) . $args{podman};
+        $module_name = 'podman-status';
+    }
+    if (!script_run($cmd_start . " echo $module_name-\$? > /dev/$serialdev", 0)) {
         return $module_name;
     } else {
-        die "Yast2 module failed to execute!\n";
+        die "$cmd_start failed to execute!\n";
```
lemon-suse, line 39: How do you think here use $cmd_start instead of 'Command' to show the cmds in log to help debug?

lemon-suse:
You have modified the logic of yast2_console_exec, so it is better to clone a yast2 module test to avoid regression, such as https://openqa.nue.suse.com/tests/11164077

lib/y2_module_consoletest.pm
lemon-suse: Seems already updated. :)

lemon-suse (approved):

PR #17455, opened 2023-07-20
Increase timeout in 'nmcli device disconnect $dev' for qemu no kvm

ggardet:
Default timeout (10 seconds) may be too short with qemu NO kvm so increase to 20s - poo#131366

- Related ticket: https://progress.opensuse.org/issues/131366
- Verification run:  https://openqa.opensuse.org/tests/3448590#step/hostname/21

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1113,7 +1113,9 @@ sub set_hostname {
                 next if ($dev eq 'lo');
                 next if !($line =~ /connected/);
 
-                assert_script_run 'nmcli device disconnect ' . $dev;
+                # Default timeout (10 seconds) may be too short with qemu NO kvm, so increase to 20s - poo#131366
+                my $nmcli = get_var('QEMU_NO_KVM') ? 'nmcli -w 20' : 'nmcli';
+                assert_script_run "$nmcli device disconnect $dev";
                 assert_script_run 'nmcli device connect ' . $dev;
             }
 


lib/utils.pm
```diff
@@ -1113,7 +1113,13 @@ sub set_hostname {
...
                 next if !($line =~ /connected/);
 
-                assert_script_run 'nmcli device disconnect ' . $dev;
+                if (get_var('QEMU_NO_KVM')) {
+                    # Default timeout (10 seconds) may be too short with qemu NO kvm, so increase to 20s - poo#131366
+                    assert_script_run 'nmcli -w 20 device disconnect ' . $dev;
+                }
+                else {
+                    assert_script_run 'nmcli device disconnect ' . $dev;
+                }
```
okurz, line 1122: That's rather ugly and hard for people to follow in all the different cases. How about simply unconditionally

```suggestion
                assert_script_run 'nmcli -w 60 device disconnect ' . $dev;
```

okurz (changes requested):

lib/utils.pm
ggardet: I suspect people would prefer to catch if default timeout is not enough in a regular case.

okurz, line 1122:
```suggestion
                # Default timeout (10 seconds) may be too short with qemu NO kvm, so increase to 20s - poo#131366
                my $nmcli = get_var('QEMU_NO_KVM') ? 'nmcli -w 20' : 'nmcli';
                assert_script_run "$nmcli device disconnect $dev";
```

okurz (changes requested):

lib/utils.pm
ggardet: Thanks for the suggestion!

okurz (approved):

PR #17457, opened 2023-07-20
Trim strings retrieved from variables

frankenmichl:
Sometimes, especially when working in development environments, you just get bogus whitespaces in variables, which will lead to non-functional tests. In order to avoid this, let's trim all strings we retrieve.


schlad (approved):
+1
I've requested the review from Petr to make sure we do not affect others' work with this patch

pevik (approved):
Very good idea, thanks!

PR #17464, opened 2023-07-21
Use spack_get_libs.sh to setup LD_LIBRARY_PATH

b10n1k:
Spack doesnt set LD_LIBRARY_PATH and libraries cant be found. Apply workaround
from bsc#1208751. This is supposed is gonna be fixed and this workaround would
be needed.

Update: Second commit overrides workaround. 

Signed-off-by: Ioannis Bonatakis ybonatakis@suse.com

https://progress.opensuse.org/issues/127481
http://aquarius.suse.cz/tests/17841 - w/o HPC_LIB
http://aquarius.suse.cz/tests/17835 - with HPC_LIB

### DIFF
--- a/lib/hpcbase.pm
+++ b/lib/hpcbase.pm
@@ -355,44 +355,43 @@ After C<prepare_spack_env> run, C<spack> should be ready to build entire tool st
 downloading and installing all bits required for whatever package or compiler.
 
 This sub is designed to install one of the mpi implementations. Although there are
-thousands packages to be used.
+thousands packages to be used. Spack will check if any mpi is installed and it will
+build the package if it is not found. In case you want to make the runtime of the
+test faster you can install C<$mpi-gnu-hpc $mpi-gnu-hpc-devel> packages in advanced.
 
 LD_LIBRARY_PATH is removed and spack is not exported. In case LD_LIBRARY_PATH is required
-it has to be add it in the F<.spack/modules.yaml>.
+it has to be added in the F<.spack/modules.yaml>.
 
 =begin text
   See bsc#1208751 for details
 =end text
 
-To do that run
+The workaround of that requires to run
 
 =begin text
   spack config add modules:prefix_inspections:lib64:[LD_LIBRARY_PATH]
   spack config add modules:prefix_inspections:lib:[LD_LIBRARY_PATH]
 =end text
 
+however the newest spack updates contain C<spack_get_libs.sh>.
+
 =cut
 
 sub prepare_spack_env {
     my ($self, $mpi) = @_;
     $mpi //= 'mpich';
-    zypper_call "in spack $mpi-gnu-hpc $mpi-gnu-hpc-devel", timeout => 1200;
+    zypper_call "in spack", timeout => 1200;
     type_string('pkill -u root');    # this kills sshd
     select_serial_terminal(0);
-    assert_script_run "module load gnu $mpi";
-    assert_script_run 'source /usr/share/spack/setup-env.sh';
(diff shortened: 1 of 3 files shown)


pevik:
From the first commit message:
> This is supposed is gonna be fixed and this workaround would be needed.

I don't understand to the sentense. Do you mean this? This is supposed is _to_ be fixed and this workaround would _not_ be needed.

lib/hpcbase.pm
```diff
@@ -355,44 +355,43 @@ After C<prepare_spack_env> run, C<spack> should be ready to build entire tool st
...
+    zypper_call "in spack", timeout => 1200;
     type_string('pkill -u root');    # this kills sshd
     select_serial_terminal(0);
-    assert_script_run "module load gnu $mpi";
-    assert_script_run 'source /usr/share/spack/setup-env.sh';
-    if (check_var('HPC_LIB', 'boost')) {
-        # Workaround for bsc#1208751
-        assert_script_run "spack config add modules:prefix_inspections:lib64:[LD_LIBRARY_PATH]";
-        assert_script_run "spack config add modules:prefix_inspections:lib:[LD_LIBRARY_PATH]";
-    }
```
pevik: Am I correct, that you add this code in previous commit and then you remove it? Why?

tests/hpc/spack_master.pm
```diff
@@ -41,7 +43,9 @@ sub run ($self) {
 
     barrier_wait('MPI_SETUP_READY');
     if (check_var('HPC_LIB', 'boost')) {
-        assert_script_run 'spack load boost';
+        assert_script_run 'source /usr/share/spack/setup-env.sh';
```
pevik, line 46: I suppose we will always use `/bin/sh` as symlink to `bash`, but `. /usr/share/spack/setup-env.sh` would work also on `dash`.

tests/hpc/spack_slave.pm
```diff
@@ -25,9 +25,13 @@ sub run ($self) {
     type_string("$testapi::password\n");
     barrier_wait('CLUSTER_PROVISIONED');
     barrier_wait('MPI_SETUP_READY');
-    assert_script_run "spack load $mpi";
-    script_run "module av";
-
+    if (check_var('HPC_LIB', 'boost')) {
+        assert_script_run 'source /usr/share/spack/setup-env.sh';
```
pevik, line 29: Also here.

b10n1k:
> From the first commit message:
> 
> > This is supposed is gonna be fixed and this workaround would be needed.
> 
> I don't understand to the sentense. Do you mean this? This is supposed is _to_ be fixed and this workaround would _not_ be needed.

I tried to do it in two steps with two consequential commits. The first has the work around. Next is the way to go as it was provided by Egbert. The gh message gets the initial commit message. Sorry for the confusion.

lib/hpcbase.pm
b10n1k: That was the solution provided initially. Later on HPC team released a script to do that. It is what the script should do which is to set LD_LIBRARY_PATH such as the library can be found

tests/hpc/spack_master.pm
b10n1k: AFAIK the first is true and we really dont use dash. lets leave it as it is.

tests/hpc/spack_slave.pm
b10n1k: As per last reply too. lets keep it as it is if there is no problem

b10n1k:
i am gonna merge it as discuss offline. i hope it will be fine

pevik:
Sure, no problem.

PR #17471, opened 2023-07-24
Bci repo

mloviska:
Obsolete and duplicate openQA BCI repo code with BCI repository test as it is
executed against fedora image (with `dnf`) in order to perform more
repoclosure check over BCI repository.

- Related ticket: https://progress.opensuse.org/issues/132935


mloviska:
VRs: https://github.com/SUSE/BCI-tests/pull/274#issuecomment-1641905220
Job group: https://gitlab.suse.de/qac/qac-openqa-yaml/-/merge_requests/1275

tests/containers/bci_test.pm
```diff
@@ -99,6 +99,9 @@ sub run {
     my $engine = $args->{runtime};
     my $bci_devel_repo = get_var('BCI_DEVEL_REPO');
     my $bci_tests_repo = get_required_var('BCI_TESTS_REPO');
+    if (my $bci_repo = get_var('REPO_BCI')) {
+        $bci_tests_repo = "http://openqa.suse.de/assets/repo/$bci_repo";
```
grisu48, line 103: Judging from the test module, this variable is actually never being used. I see the VRs are passing but are we sure, that they are using the desired BCI repo? :thinking:

mloviska: Right, good catch! Thanks

http://kepler.suse.cz/tests/21329#step/bci_test_podman/40

lib/main_containers.pm
```diff
@@ -243,9 +243,6 @@ sub load_container_tests {
                 load_image_tests_in_k8s($run_args) if (/k8s/i);
                 load_image_tests_in_openshift if (/openshift/i);
             }
-        } elsif (get_var('REPO_BCI')) {
-            loadtest 'containers/host_configuration';
-            loadtest 'containers/bci_repo';
```
jlausuch, line 248: wait, if you remove this, how is these jobs now scheduled?

mloviska: It would be scheduled by ibs sync and I have put the related variables into job group.

jlausuch: Ah ok, it's scheduled because is_container_image_test() returns 1 due to `BCI_TESTS=1`.

jlausuch (approved):

PR #17481, opened 2023-07-25
Fix condition in publiccloud::azure::get_image_definition()

pdostal:
- Related ticket: https://progress.opensuse.org/issues/133013
- Verification run: https://pdostal-server.suse.cz/tests/4773#details

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -165,6 +165,7 @@ sub get_image_definition {
     # Check if the given image definition, identified by the tuple (publisher, offer and sku) is present in the given image gallery
     # returns the name of the found image definition or undef if not found
     my ($self, $resource_group, $gallery, $publisher, $offer, $sku) = @_;
+    record_info('get_image_definition', "Searching for image definition with publisher=$publisher offer=$offer sku=$sku");
 
     my $definitions = script_output("az sig image-definition list --resource-group '$resource_group' --gallery-name '$gallery'");
     return undef unless ($definitions);
@@ -172,8 +173,12 @@ sub get_image_definition {
     foreach my $def (@$json_data) {
         my $identifier = $def->{identifier};
         next unless (defined $identifier);
-        return $def->{name} if ($identifier->{publisher} eq $publisher && $identifier->{offer} eq $offer && $identifier->{sku} eq $sku);
+        if ($identifier->{publisher} =~ /$publisher/i && $identifier->{offer} =~ /$offer/i && $identifier->{sku} =~ /$sku/i) {
+            record_info('image_definition', "Found $def->{name} image definition");
+            return $def->{name};
+        }
     }
+    record_info('no image_definition', "Did not found image definition.");
     return undef;
 }
 


ricardobranco777 (approved):
LGTM

lib/publiccloud/azure.pm
```diff
@@ -165,15 +165,20 @@ sub get_image_definition {
...
+    record_info('get_image_definition', "Searching for image definition with publisher=$publisher offer=$offer sku=$sku");
 
     my $definitions = script_output("az sig image-definition list --resource-group '$resource_group' --gallery-name '$gallery'");
     return undef unless ($definitions);
     my $json_data = decode_azure_json($definitions);
     foreach my $def (@$json_data) {
         my $identifier = $def->{identifier};
         next unless (defined $identifier);
-        return $def->{name} if ($identifier->{publisher} eq $publisher && $identifier->{offer} eq $offer && $identifier->{sku} eq $sku);
+        if ($identifier->{publisher} =~ /$publisher/i && $identifier->{offer} =~ /$offer/i && $identifier->{sku} =~ /$sku/i) {
```
asmorodskyi, line 176: can you elaborate on this ? This is images which we uploading , right ? How did that happened that we don't know exact `publisher` , `offer` and `sku` for the image that we uploaded so we need to ignore the case and search for at least something similar

asmorodskyi: seems like we hiding problems here

grisu48 (approved):

lib/publiccloud/azure.pm
asmorodskyi: yes I know but why we can not search `$identifier->{publisher} eq $publisher` but need to use `$identifier->{publisher} =~ /$publisher/i` ?

pdostal: Because the `offer` contains `%DISTRI%-%VERSION%-%FLAVOR%-%ARCH%` and isn't case sensitive. So if `%FLAVOR%` is once `BYOS-Azure` and then `BYOS-AZURE` then we have a problem.

asmorodskyi: can you point me to the case where we have same DISTRI , VERSION, ARCH the same and flavor is different and we want this two flavors to share same image ? 
I know that we have problems with different writing of AZURE/Azure but I hardly believe that there is a case where we really need to treat that as same image . 
Also even if this is really the case it is good excuse ONLY for offer , publisher and sku should be always the same and in same case

asmorodskyi: this is very important piece of code which basically deciding which exact image will be tested . Mistake done here really hard to find ( you got some image after all so no reason to throw error which will draw attention of reviewer ) , on other hand mistake done here have very high price ( we reporting that something is tested when in a fact different image was taken ) .  So relaxing search conditions here without good reason is increasing risks without good reason for this . 
And I certainly don't see ANY reason for relaxing search conditions for publisher and sku , and have big doubts about the offer

pdostal: To be honest I had to search, but here you are: In our code, we have:
```perl
my $offer = get_var("PUBLIC_CLOUD_AZURE_OFFER", get_var('DISTRI') . '-' . get_var('VERSION') . '-' . get_var('FLAVOR') . '-' . $arch);
```
which produces:
```
sle-micro-5.3-BYOS-Azure-ARM-Arm64
```
Then when we search for existing definitions, we get:
```bash
$ az sig image-definition list --resource-group 'openqa-upload' --gallery-name 'test_image_gallery' | jq -r '.[] | select(.identifier.sku=="gen2" and .identifier.publisher=="qe-c" and .identifier.offer=="sle-micro-5.3-BYOS-Azure-ARM-Arm64").identifier.offer'
sle-micro-5.3-BYOS-Azure-ARM-Arm64
```

Which does not proof what I say, so I will prepare the revert.

PR #17485, opened 2023-07-25
Check custom systemd service setup

mloviska:
Instead of checking `qemu-guest-agent.service` that requires a device to be added, try to write a custom systemd service and check whether it has executed correctly.

- Verification run: http://kepler.suse.cz/tests/21343#

### DIFF
--- a/data/jeos/g_cloudinit.cfg
+++ b/data/jeos/g_cloudinit.cfg
@@ -42,8 +42,24 @@ chpasswd:
 packages:
   - iotop
 
+write_files:
+  - content: |
+        [Unit]
+        Description=Simple test service
+
+        [Service]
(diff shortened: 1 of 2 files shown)


ricardobranco777 (approved):
Isn't the test really about `qemu-guest-agent' being enabled and started?

If not, LGTM.

mloviska:
> Isn't the test really about `qemu-guest-agent' being enabled and started?
> 

Not really, the point is the exercise cloud init + minimal-vm Cloud image. `qemu-guest-agent` was chosen before because it is pre-installed but you need to boot the VM with `/dev/virtio-ports/org.qemu.guest_agent.0` device.

> If not, LGTM.

PR #17495, opened 2023-07-26
Fix and enable user-serial-terminal on zVM and JeOS

dzedro:
- Related ticket: https://progress.opensuse.org/issues/129026
- Verification run: In Progress
https://openqa.suse.de/tests/overview?build=zypper_lifecycle SLE
https://dzedro.suse.cz/tests/overview?build=zypper_lifecycle&distri=opensuse&version=Tumbleweed TW

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -592,7 +592,7 @@ sub init_consoles {
 
             ($hostname) = $s390_params =~ /Hostname=(\S+)/;
 
-            # adds serial console for S390_ZKVM
+            # adds serial console for s390x zVM
             # NOTE: adding consoles just at the top of init_consoles() is not enough, otherwise
             # using just them would fail with:
             # ::: basetest::runtest: # Test died: Error connecting to <root@192.168.112.9>: Connection refused at /usr/lib/os-autoinst/testapi.pm line 1700.
@@ -604,6 +604,14 @@ sub init_consoles {
                         hostname => $hostname,
                         password => $testapi::password,
                         username => 'root'
+                    });
+                $self->add_console(
+                    'user-serial-ssh',
+                    'ssh-serial',
+                    {
+                        hostname => $hostname,
+                        password => $testapi::password,
+                        username => $testapi::username
                     });
             }
         }
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -441,28 +441,6 @@ sub init_consoles {
         }
     }
 
-    if (get_var('SUT_IP') || get_var('VIRSH_GUEST')) {
-        my $hostname = get_var('SUT_IP', get_var('VIRSH_GUEST'));
-
-        $self->add_console(
-            'root-serial-ssh',
-            'ssh-serial',
-            {
-                hostname => $hostname,
-                password => $testapi::password,
-                username => 'root'
-            });
-
-        $self->add_console(
-            'user-serial-ssh',
-            'ssh-serial',
-            {
-                hostname => $hostname,
-                password => $testapi::password,
-                username => $testapi::username
-            });
-    }
-
     # svirt backend, except s390x ARCH
     if (is_svirt_except_s390x) {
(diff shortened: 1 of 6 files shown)


pevik (commented):
Looks good so far, looking forward to the verification run.

lib/susedistribution.pm
```diff
@@ -605,6 +605,14 @@ sub init_consoles {
                         password => $testapi::password,
                         username => 'root'
                     });
```
pevik, line 607: Maybe adding here comment that it's for zVM and JeOS? (the previous one is for zkvm). Or not needed because all developers can use `git blame`?

dzedro:
The serial console on zypper_lifecycle test on JeOS ans s390x is IMO straight forward, but problematic was travis failure thus I merged the duplicate `add_console` code into one https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17495/commits/83abd4cdaef54eb2b7503e3724ee218657842c29 and consoletest_finish on s390x was failing. It's mystery to me how is it mostly passing and with unrelated change which makes previous test faster it will start to fail. Exception is 15-SP2 s390x-zVM where it was failing pretty often https://openqa.suse.de/tests/11715263#next_previous

lib/susedistribution.pm
```diff
@@ -576,38 +554,44 @@ sub init_consoles {
...
-            unless (get_var('SUT_IP')) {
-                $self->add_console(
-                    'root-serial-ssh',
-                    'ssh-serial',
-                    {
-                        hostname => $hostname,
-                        password => $testapi::password,
-                        username => 'root'
-                    });
-            }
```
mloviska, line 608: damn, I am wondering what else might break over here. I guess we can just try it out and be ready to revert if anything

mloviska: hm...maybe there should be more `svirt` VRs ?

dzedro: I tried o cover it will test on "each" backend, but ofc it can still break somewhere. :crossed_fingers:

dzedro: Think I tested each svirt, but I can check again.

foursixnine (approved):

lib/susedistribution.pm
dzedro: I went trough the worker configuration https://gitlab.suse.de/openqa/salt-pillars-openqa/-/blob/master/openqa/workerconf.sls
Added few tests, I could not clone some tests because e.g. ipmi workers are down I guess due to migration.
But there is one ipmi tests and I assume they should behave same.
svirt-hyperv tests are failing but this test are AFAIS very bad, e.g. bootloader is finished but in following test is still typing in grub2, behaves same wihtout my change. :face_with_spiral_eyes: 
I would merge this next week :game_die:  :pray:

dzedro: There is no `svirt-hyperv2012r2`, that is special, I found only old 15-SP5 tests on build 102.1

tests/console/consoletest_finish.pm
```diff
@@ -41,12 +42,19 @@ sub run {
...
-    $console->reset;
-
-    $console = select_console 'user-console';
-    enter_cmd "exit";    # logout
-    $console->reset;
     wait_still_screen(2);
+    select_console 'user-console';
+    enter_cmd "exit";
+    wait_still_screen(2);
+    unless (check_var('SERIAL_CONSOLE', 0) || check_var('VIRTIO_CONSOLE', 0)) {
```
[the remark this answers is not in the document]
dzedro: I just used it, same as it's used in `select_serial_terminal`, I assume if it works there then it will also here.
https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/serial_terminal.pm#L372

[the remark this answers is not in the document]
pevik: Sure, sure, it would be just cleaner design. Just a minor detail...

PR #17496, opened 2023-07-26
Revert #17481 Fix condition in publiccloud::azure::get_image_definition()

pdostal:
Reverts #17481

 * Related ticket: https://progress.opensuse.org/issues/133013

### DIFF
--- a/lib/publiccloud/azure.pm
+++ b/lib/publiccloud/azure.pm
@@ -173,7 +173,7 @@ sub get_image_definition {
     foreach my $def (@$json_data) {
         my $identifier = $def->{identifier};
         next unless (defined $identifier);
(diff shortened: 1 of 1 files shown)


lib/publiccloud/azure.pm
```diff
@@ -173,7 +173,7 @@ sub get_image_definition {
     foreach my $def (@$json_data) {
         my $identifier = $def->{identifier};
         next unless (defined $identifier);
-        if ($identifier->{publisher} =~ /$publisher/i && $identifier->{offer} =~ /$offer/i && $identifier->{sku} =~ /$sku/i) {
+        if ($identifier->{publisher} eq $publisher && $identifier->{offer} eq $offer && $identifier->{sku} eq $sku) {
```
grisu48, line 176: but ... why?

pdostal: When I explained that change yesterday to Anton I was not able to find any example.
The expression was also pretty benevolent so I decided to just leave the `record_info`'s for now.

PR #17503, opened 2023-07-28
Add new validate_repos test module

BillAnastasiadis:
Create validate_repos derived from download_repos. Add this new test module in HanaSr and mr_test scheduled when IS_MAINTENANCE is enabled

- Related ticket: https://jira.suse.com/browse/TEAM-8322
- Verification run: 
WITH EMBARGOED: http://openqaworker15.qa.suse.cz/tests/208684#step/validate_repos/46
NO REPO: http://openqaworker15.qa.suse.cz/tests/208685#step/validate_repos/26

### DIFF
--- a/lib/publiccloud/utils.pm
+++ b/lib/publiccloud/utils.pm
@@ -25,6 +25,7 @@ our @EXPORT = qw(
   deregister_addon
   define_secret_variable
   get_credentials
+  validate_repo
   is_byos
   is_ondemand
   is_ec2
@@ -131,6 +132,22 @@ sub register_openstack {
     my $cmd = "sudo SUSEConnect -r $regcode";
     $cmd .= " --url $fake_scc" if $fake_scc;
     $instance->ssh_assert_script_run(cmd => $cmd, timeout => 700, retry => 5);
+}
+
+# Validation for update repos
+sub validate_repo {
+    my ($maintrepo) = @_;
+    if ($maintrepo =~ /\/(PTF|Maintenance):\/(\d+)/g) {
+        my ($incident, $type) = ($2, $1);
+        die "We did not detect incident number for URL \"$maintrepo\". We detected \"$incident\"" unless $incident =~ /\d+/;
+        if (is_embargo_update($incident, $type)) {
+            record_info("EMBARGOED", "The repository \"$maintrepo\" belongs to embargoed incident number \"$incident\"");
+            script_run("echo 'The repository \"$maintrepo\" belongs to embargoed incident number \"$incident\"'");
+            return 0;
+        }
+        return 1;
+    }
+    die "Unexpected URL \"$maintrepo\"";
 }
 
 # Check if we are a BYOS test run
--- a/tests/publiccloud/validate_repos.pm
+++ b/tests/publiccloud/validate_repos.pm
@@ -1,0 +1,58 @@
+# SUSE's openQA tests
+#
+# Copyright 2019-2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: Download repositores from the internal server
+#
+# Maintainer: qa-c <qa-c@suse.de>
+
+use base 'consoletest';
+use registration;
+use warnings;
+use testapi;
+use strict;
+use utils;
+use publiccloud::ssh_interactive "select_host_console";
+use publiccloud::utils "validate_repo";
+
+sub run {
+    my ($self, $args) = @_;
+    select_host_console();    # select console on the host, not the PC instance
+
+    if (get_var('PUBLIC_CLOUD_SKIP_MU')) {
+        # Skip maintenance updates. This is useful for debug runs
+        record_info('Skip validation', 'Skipping maintenance update validation (triggered by setting)');
+        return;
+    } else {
+        # In Incidents there is INCIDENT_REPO instead of MAINT_TEST_REPO
+        # Those two variables contain list of repositories separated by comma
+        set_var('MAINT_TEST_REPO', get_var('INCIDENT_REPO')) unless get_var('MAINT_TEST_REPO');
+        my @repos = split(/,/, get_var('MAINT_TEST_REPO'));
+        # Failsafe: Fail if there are no test repositories, otherwise we have the wrong template link
+        my $count = scalar @repos;
+        my $check_empty_repos = get_var('PUBLIC_CLOUD_IGNORE_EMPTY_REPO', 0) == 0;
+        die "No test repositories" if ($check_empty_repos && $count == 0);
+
+        my $repo_count = 0;
+        my ($incident, $type);
+        for my $maintrepo (@repos) {
+            next unless validate_repo($maintrepo);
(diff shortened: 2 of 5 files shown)


lib/publiccloud/utils.pm
```diff
@@ -133,6 +134,19 @@ sub register_openstack {
     $instance->ssh_assert_script_run(cmd => $cmd, timeout => 700, retry => 5);
 }
 
+# Validation for update repos
+sub validate_repo {
+    my ($maintrepo) = @_;
+    my ($incident, $type) = ($2, $1) if ($maintrepo =~ /\/(PTF|Maintenance):\/(\d+)/g);
```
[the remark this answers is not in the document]
asmorodskyi: ofc. it is less likely but if we want to handle unexpected situations let's also take into account when string didn't match in such a case no point to print undef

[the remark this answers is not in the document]
asmorodskyi: ok now I see that you just moved code from old place to a function , feel free to ignore my comment as in a fact you haven't introduced this code . Nevertheless if you will tweak it IMHO it will be nice improvement

tests/publiccloud/validate_repos.pm
```diff
@@ -0,0 +1,58 @@
...
+use publiccloud::ssh_interactive "select_host_console";
+use publiccloud::utils "validate_repo";
+
+sub run {
+    my ($self, $args) = @_;
+    select_host_console();    # select console on the host, not the PC instance
+
+    if (get_var('PUBLIC_CLOUD_SKIP_MU')) {
+        # Skip maintenance updates. This is useful for debug runs
+        record_info('Skip calidation', 'Skipping maintenance update validation (triggered by setting)');
```
asmorodskyi, line 25:
```suggestion
        record_info('Skip validation', 'Skipping maintenance update validation (triggered by setting)');
```

tests/publiccloud/validate_repos.pm
```diff
@@ -0,0 +1,58 @@
...
+        my $check_empty_repos = get_var('PUBLIC_CLOUD_IGNORE_EMPTY_REPO', 0) == 0;
+        die "No test repositories" if ($check_empty_repos && $count == 0);
+
+        my $repo_count = 0;
+        my ($incident, $type);
+        for my $maintrepo (@repos) {
+            next unless validate_repo($maintrepo);
+            $repo_count++;
+        }
+        die "No usable test repositories" if ($repo_count == 0);
```
asmorodskyi, line 43: no to have a failure in openQA at this place , I suggesting to soft fail here . @pdostal  , @okurz , @grisu48 , @michaelgrifalconi ( please add more relevant people as decision is quite big actually )

asmorodskyi: to give more context - we deciding what to do when we have job triggered for maint. aggregates/incidents and contains ONLY embargoed updates which was filtered out so job is actually a fake but in case we will die here it block release when actually no issue was found

okurz: Soft failure pointing to which product issue exactly? Simple rule: soft fails must always point to an open bugzilla entry, i.e. valid product issue, for everything else you likely just want to not influence the test result.

asmorodskyi: > Soft failure pointing to which product issue exactly? Simple rule: soft fails must always point to an open bugzilla entry, i.e. valid product issue, for everything else you likely just want to not influence the test result.

generally I agree with you and also trying to follow this rule as much as I can , but unfortunately seems like we need more statuses :) . What status you suggests here ? I agree with your concerns regarding soft-failure but do you agree with mine that failing the test here is also not acceptable ? Because we will block unfairly block release of embargoed updates ( which usually the one with most of time pressure ) . 
To me looks like we choosing between bad and very bad options and question is what is worse unfairly block important updates or break golden rule of "no soft-failure without a bug " ...

BillAnastasiadis: @asmorodskyi but in cloud cases it makes sense for the job to stop, right? I mean, why deploy (and pay) when there's no (valid) update to be tested in maintenance jobs? It would be a pointless run.

That's what Michele had in mind when he proposed this - to stop the job before deploy if no (valid) updates exist to be tested. Yes, failure is not ideal since this is closer to a cancellation than failure... but softfail would allow the job to move on to deploy which would miss the point of creating this module

tests/publiccloud/validate_repos.pm
BillAnastasiadis: Thanks for caching that, fixed!

lib/publiccloud/utils.pm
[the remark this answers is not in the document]
BillAnastasiadis: Thank you for the proposal, implemented!

mpagot (approved):
LGTM

tests/publiccloud/validate_repos.pm
okurz: > generally I agree with you and also trying to follow this rule as much as I can , but unfortunately seems like we need more statuses :)

I don't think that would help

> What status you suggests here ? I agree with your concerns regarding soft-failure but do you agree with mine that failing the test here is also not acceptable?

Yes, I agree. So how about just pass the test module then? What would happen as consequence?

asmorodskyi: ok , let's fail here . Don't want to block you here

mpagot: A little bit of more context hoping to help. Code for this new test module is derived from https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/tests/publiccloud/download_repos.pm

The **download_repo** test module is implementing both validation and upload/download of the repo to test.

The **download_repo** approach in case of embargoed update is just to ignore it https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/e03233eb9ce9257d85f9383e9dfe0c72e03eb477/tests/publiccloud/download_repos.pm#L69

This new test module proposed in this PR is not mind to be used in the same test schedule of **download_repo**

We have tests schedule that cannot use the update download/upload approach used by download_repo, so we are provisioning the update to test in a totally different way. But the criteria about embargoed updates has to be applied there too. 

The scheduling is coded in  https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/e03233eb9ce9257d85f9383e9dfe0c72e03eb477/tests/sles4sap/publiccloud/hana_sr_schedule_deployment.pm#L33 

The problem with that scheduling is that we are realizing that the update is embargoed and not to be tested only later after many time from the test beginning, and only after other steps that are consuming resources in the cloud. 

This PR is our attempt to solve this issue. The idea here is to add a test module,  dedicated to the update list validation, at the first beginning of the schedule. How can we run it and avoid to schedule any further not needed test modules if all the updates are embargoed?

ricardobranco777 (approved):
LGTM

alvarocarvajald (approved):
LGTM

PR #17505, opened 2023-07-28
Add Ansible cmd to retry if Azure reboot timeout error found

rjagu1:
Add Ansible cmd to re-trigger run if it find Azure reboot time out in qesapdeployment/deploy.pm
Add Fail_ok = 1 flag in qesap_upload_crm_report to avoid log collecting failures

- Related ticket: [TEAM-8006](https://jira.suse.com/browse/TEAM-8006)
- Verification run: AZURE :  http://openqaworker15.qa.suse.cz/t208688
VR with wrong reg_code : http://openqaworker15.qa.suse.cz/tests/209145 (Expected failure)
-VR for Fail_ok flag : http://openqaworker15.qa.suse.cz/tests/209144
- AWS VR : http://openqaworker15.qa.suse.cz/t208689

### DIFF
--- a/tests/sles4sap/qesapdeployment/deploy.pm
+++ b/tests/sles4sap/qesapdeployment/deploy.pm
@@ -27,6 +27,7 @@ sub run {
         my $rec_timeout = qesap_ansible_log_find_timeout($ret[1]);
         if ($rec_timeout) {
             record_info('DETECTED ANSIBLE TIMEOUT ERROR');
+            @ret = qesap_execute(cmd => 'ansible', cmd_options => '--profile', verbose => 1, timeout => 3600);
         }
         die "'qesap.py ansible' return: $ret[0]";
     }


tests/sles4sap/qesapdeployment/deploy.pm
```diff
@@ -27,6 +27,7 @@ sub run {
         my $rec_timeout = qesap_ansible_log_find_timeout($ret[1]);
         if ($rec_timeout) {
             record_info('DETECTED ANSIBLE TIMEOUT ERROR');
+            @ret = qesap_execute(cmd => 'ansible', cmd_options => '--profile', verbose => 1, timeout => 3600);
```
mpagot, line 30: I think it is not ok as in case of Ansible failure and then Ansible pass at the retry, the `die` is called in any case

lib/qesapdeployment.pm
```diff
@@ -874,6 +874,7 @@ sub qesap_upload_crm_report {
         provider => $prov,
         filter => $host,
         host_keys_check => 1,
+        failok => 1,
```
lpalovsky, line 877: I think it might be good to have something like softfail here that will tell us about 'crm report' failing and it needs to be fixed. Maybe at the end of the test there could be a curl call to check if file is present in the results?
Something like checking  url 'https://<openqa instance>/tests/<test id>/logfile?filename=filename'
@mpagot wdyt?

mpagot (approved):
LGTM

lib/qesapdeployment.pm
mpagot: The point here is that we are sometime calling `qesap_upload_crm_report` after an Ansible failure. The failure could be so early that the cluster is not there at all. 
I'd propose not to add a softfailure but to  add a `failok` argument in the `qesap_upload_crm_report` too to let the caller to decide. I propose to do it later in a different PR. I'm also ok to fully remove this `failok` change from this PR at all and address everything in a new dedicated PR.

tests/sles4sap/qesapdeployment/deploy.pm
mpagot: ok solved as the behavior is intentional. We do not know yet what is going to happens to run Ansible for a second time in a row. So it is fine to do it but to `die` in any case.

lib/qesapdeployment.pm
BillAnastasiadis: I agree with Michele. @rjagu1 if you want to go for it in this ticket, it's as simple as adding

**qesapdeployment.pm, line 903:**
`my $failok = $args{failok} // 0;` (or 1, to have failing be ok be default)

and then on the subsequent `qesap_ansible_cmd` and `qesap_ansible_fetch_file` on lines 906 and 911, add:
`failok => $failok`

jankohoutek (approved):
LGTM

alvarocarvajald (commented):
Code LGTM.

I do have some questions though:

1. VR is failing. Is it expected?
2. After the `qesap_execute(cmd => 'ansible', ...)` retry, test module will still die in the following sentence. Is that intentional? Unfortunately Pull Request's description is not clear if the re-triggering is intended to work around some issue or simply retry but still fail.
3. Speaking of the Pull Request's description, there is no code implementing the sentence below:
> Add Fail_ok = 1 flag in qesap_upload_crm_report to avoid log collecting failures

rjagu1:
> Code LGTM.
> 
> I do have some questions though:
> 
> 1. VR is failing. Is it expected? 
_Yes ,it is to check behavior in failed cases_
> 2. After the `qesap_execute(cmd => 'ansible', ...)` retry, test module will still die in the following sentence. Is that intentional? Unfortunately Pull Request's description is not clear if the re-triggering is intended to work around some issue or simply retry but still fail.
_This is work around cmd which triggers rarely as it is in sporadic nature_
> 3. Speaking of the Pull Request's description, there is no code implementing the sentence below:
> 
> > Add Fail_ok = 1 flag in qesap_upload_crm_report to avoid log collecting failures  _Raising Dedicated PR, so here it was discarded _

PR #17512, opened 2023-08-01
Restart firewalld on 15-SP5

grisu48:
Apply the known workaround to fix not working ip-forwarding on 15-SP5.

- Related ticket: https://bugzilla.suse.com/show_bug.cgi?id=1213811
- Verification run: https://duck-norris.qe.suse.de/tests/13382

### DIFF
--- a/tests/containers/bci_test.pm
+++ b/tests/containers/bci_test.pm
@@ -65,7 +65,7 @@ sub reset_engines {
     my ($self, $current_engine) = @_;
     my ($version, $sp, $host_distri) = get_os_release;
     my $sp_version = "$version.$sp";
-    if ($sp_version =~ /15.3|15.4/) {
+    if ($sp_version =~ /15.3|15.4|15.5/) {
         # This workaround is only needed in SLE 15-SP3 and 15-SP4 (and Leap 15.3 and 15.4)
         # where we need to restart docker and firewalld before running podman, otherwise
         # the podman containers won't have access to the outside world.
--- a/tests/containers/host_configuration.pm
+++ b/tests/containers/host_configuration.pm
@@ -47,8 +47,8 @@ sub run {
     install_podman_when_needed($host_distri) if ($engine =~ 'podman');
 
     # It has been observed that after system update, the ip forwarding doesn't work.
-    # In 15.3/15.4 there is a need to restart the firewall and docker daemon.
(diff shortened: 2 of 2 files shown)


mloviska:
Nice catch, nevertheless as it is a workaround it would make sense that devs would have a look at this.

tests/containers/bci_test.pm
```diff
@@ -65,7 +65,7 @@ sub reset_engines {
     my ($self, $current_engine) = @_;
```
mloviska, line 65: I think this makes perfect sense to be moved into a `bci_prepare` rather than keeping it in `bci_test`

grisu48: We can, let's do this in a later step, we need to release the BCI containers in the pipeline first.

grisu48: After looking at it more closely, I think we should leave it in the `bci_test.pm` module, as it stops docker when podman is being tested. We cannot do this globally, as we are testing docker and podman within one test run.

grisu48: It's unfortunately needed in both test runs, but we can move it to a library function.

tests/containers/bci_test.pm
```diff
@@ -65,7 +65,7 @@ sub reset_engines {
...
     my ($version, $sp, $host_distri) = get_os_release;
     my $sp_version = "$version.$sp";
-    if ($sp_version =~ /15.3|15.4/) {
+    if ($sp_version =~ /15.3|15.4|15.5/) {
         # This workaround is only needed in SLE 15-SP3 and 15-SP4 (and Leap 15.3 and 15.4)
```
pdostal, line 69: and spř

grisu48: I'm sorry, I don't understand

mloviska: I guess he means that the comment is missing `SP5` :D

grisu48: I thought he means to add `15.6` and the ř is just a typo.

pdostal: I meant SP5 but now I see that it's there. Sorry.

tests/containers/bci_test.pm
grisu48: See https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17527

PR #17515, opened 2023-08-01
Type very slow if NVDIMM

BillAnastasiadis:
Nvdimm tests tend to sporadically fail because of typos made during typing grub cmdline. This pr uses a slower typing speed when nvdimm tests are detected.

- Related ticket: https://jira.suse.com/browse/TEAM-8111

- Verification runs: (jobs where cancelled after they passed the relevant module)
https://openqa.suse.de/tests/11731494
https://openqa.suse.de/tests/11731235
https://openqa.suse.de/tests/11731495
https://openqa.suse.de/tests/11732359
https://openqa.suse.de/tests/11732360
https://openqa.suse.de/tests/11733026

(6 back to back tests, problem does not appear)

### DIFF
--- a/tests/boot/boot_from_pxe.pm
+++ b/tests/boot/boot_from_pxe.pm
@@ -15,7 +15,7 @@ use lockapi;
 use testapi;
 use bootloader_setup qw(bootmenu_default_params specific_bootmenu_params prepare_disks sync_time);
 use registration 'registration_bootloader_cmdline';
-use utils qw(type_string_slow enter_cmd_slow);
+use utils qw(type_string_slow type_string_very_slow enter_cmd_slow enter_cmd_very_slow);
 use Utils::Backends;
 use Utils::Architectures;
 use version_utils qw(is_upgrade is_sle);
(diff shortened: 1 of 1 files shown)


mpagot (approved):
LGTM

emiura:
LGTM

alvarocarvajald (commented):
LGTM.

My only issue here, is with the use of the NVDIMM setting. NVDIMM setting is more a test setting than a worker setting; it tells the test modules to test for NVDIMM. In a way is related to the bare metal HW, but you can run tests on the bare metal HW without setting NVDIMM and they should work as well. However in that case, the changes in `boot/boot_from_pxe` will skip the very slow typing, even though they're running in the same HW.

The only worker-specific setting that I see on both NVDIMM workers is the worker class, which is `64bit-ipmi-nvdimm`, but no idea how complicated it would be to check for that instead of on *NVDIMM* as the *WORKER_CLASS* is a special setting which may contain a list.

alvarocarvajald (approved):
LGTM

PR #17516, opened 2023-08-01
Create a new module to recover version to original

manfredi:
- Description: Create a new module to recover version to original
- Related ticket: [issues/132734](https://progress.opensuse.org/issues/132734)
- Verification run: [Test result overview](https://openqa.suse.de/tests/overview?build=manfredi%2Fos-autoinst-distri-opensuse%23issues-132734&distri=sle&version=15-SP5)

### DIFF
--- a/schedule/migration/migration_offline_media_all_patterns_x86_64.yaml
+++ b/schedule/migration/migration_offline_media_all_patterns_x86_64.yaml
@@ -7,7 +7,7 @@ vars:
   YUI_REST_API: '1'
   PATTERNS: all
 schedule:
-  - migration/version_switch_origin_system
+  - migration/version_reset_origin_system
   - boot/boot_to_desktop
   - migration/suseconnect_register_system
   - migration/patch_and_reboot_system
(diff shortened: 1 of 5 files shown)


tests/migration/version_recover_origin_system.pm
```diff
@@ -0,0 +1,25 @@
+# SUSE's openQA tests
+#
+# Copyright 2017 SUSE LLC
```
lemon-suse, line 3: s/2017/2023

tests/migration/version_recover_origin_system.pm
```diff
@@ -0,0 +1,25 @@
+# SUSE's openQA tests
```
lemon-suse, line 1: Thinking about the name of version_recover_origin_system.pm , for me 'recover' is not suitable because no 'recover' thing here, I suggest to use 'set' instead of 'recover'. WDYT?

tests/migration/version_recover_origin_system.pm
manfredi: Changed

tests/migration/version_recover_origin_system.pm
lemon-suse: For the name, 'switch' is very suitable but it already used in old module, maybe 'reset' or 'set' is somehow like the meaning. Anyway, we'd better find a suitable name here.

manfredi: Changed to `migration/version_reset_origin_system`

lemon-suse (approved):

PR #17524, opened 2023-08-03
Change hard coded product names in playwright test specs

lemon-suse:
Change hard coded product names in playwright test specs, add upload trace file for tests/yam/agama/agama_full_disk_encryption.pm.

- Related ticket: https://progress.opensuse.org/issues/133562
- Related Playwright PR: https://github.com/jknphy/e2e-agama-playwright/pull/9
- Needles: N/A
- Verification run:  https://openqa.opensuse.org/tests/3478043#step/agama_lvm/5
  https://openqa.opensuse.org/tests/3478044#step/agama_full_disk_encryption/5
The failures are for https://bugzilla.suse.com/show_bug.cgi?id=1213836

### DIFF
--- a/tests/yam/agama/agama_full_disk_encryption.pm
+++ b/tests/yam/agama/agama_full_disk_encryption.pm
@@ -12,6 +12,7 @@ use warnings;
 use testapi qw(
   assert_screen
   assert_script_run
+  get_required_var
   select_console
   upload_logs
 );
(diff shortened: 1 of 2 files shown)


sofiasyria:
I am getting confused. We have three tests: lvm, full_disk_encryption and take_screenshots. For each, we have a different module that is very similar but has some differences like the base (why we have agama_base in some and y2_installbase in other?) and the test name. We could use the same base for all and have the post_fail_hook there. And we could also use the same module for all tests and just change the test name, same as you did for the product name.

lemon-suse:
> I am getting confused. We have three tests: lvm, full_disk_encryption and take_screenshots. For each, we have a different module that is very similar but has some differences like the base (why we have agama_base in some and y2_installbase in other?) and the test name. We could use the same base for all and have the post_fail_hook there. And we could also use the same module for all tests and just change the test name, same as you did for the product name. I'd suggest for now to just skip adding the post_fail_hook. Merge your code and open a follow up ticket to tidy this mess up. WDYT?

Aha, we already have the ticket, https://progress.opensuse.org/issues/132989

lemon-suse:
> > I am getting confused. We have three tests: lvm, full_disk_encryption and take_screenshots. For each, we have a different module that is very similar but has some differences like the base (why we have agama_base in some and y2_installbase in other?) and the test name. We could use the same base for all and have the post_fail_hook there. And we could also use the same module for all tests and just change the test name, same as you did for the product name. I'd suggest for now to just skip adding the post_fail_hook. Merge your code and open a follow up ticket to tidy this mess up. WDYT?
> 
> Aha, we already have the ticket, https://progress.opensuse.org/issues/132989

Removed the post_fail_hook.

PR #17525, opened 2023-08-03
Disable qemu/kvm specific test for SLE Micro RT

waynechen55:
* **SLE Micro RT** has no kvm support, so disable relevant tests, otherwise it will fail like [this](https://openqa.suse.de/tests/11713080#step/qemu/4).

* **Please** refer to [SLE Micro 5.5 Test Plan](https://confluence.suse.com/display/qasle/Master+Test+Plan+for+SLE+Micro+5.5)

*  **Verification runs:**
    * [Loaded slem_installation_virt on x86_64 DVD](http://10.163.28.134/tests/1927)
    * [Loaded slem_installation_virt on aarch64 DVD](http://10.163.28.134/tests/1931)
    * [Loaded slem_installation_virt on s390x DVD](http://10.163.28.134/tests/1932)
    * [Loaded slem_virtualization on x86_64 Default](http://10.163.28.134/tests/1933)
    * [Loaded slem_virtualization on aarch64 Default](http://10.163.28.134/tests/1934)
    * [Loaded slem_virtualizaiton on s390x Default](http://10.163.28.134/tests/1935)
    * [Loaded slem_virtualization on x86_64 Default-RT](http://10.163.28.134/tests/1926)
    OR
    * [Loaded slem_installation_virt on x86_64 DVD](http://10.67.18.153/tests/1927)
    * [Loaded slem_installation_virt on aarch64 DVD](http://10.67.18.153/tests/1931)
    * [Loaded slem_installation_virt on s390x DVD](http://10.67.18.153/tests/1932)
    * [Loaded slem_virtualization on x86_64 Default](http://10.67.18.153/tests/1933)
    * [Loaded slem_virtualization on aarch64 Default](http://10.67.18.153/tests/1934)
    * [Loaded slem_virtualizaiton on s390x Default](http://10.67.18.153/tests/1935)
    * [Loaded slem_virtualization on x86_64 Default-RT](http://10.67.18.153/tests/1926)

### DIFF
--- a/lib/main_micro_alp.pm
+++ b/lib/main_micro_alp.pm
@@ -204,8 +204,8 @@ sub load_network_tests {
 sub load_qemu_tests {
     loadtest 'microos/rebuild_initrd' if is_s390x;
     loadtest 'qemu/info';
-    loadtest 'qemu/qemu';
-    loadtest 'qemu/kvm' unless is_aarch64;
+    loadtest 'qemu/qemu' unless is_rt;
+    loadtest 'qemu/kvm' unless (is_aarch64 or is_rt);
     # qemu-linux-user package not available in SLEM
     loadtest 'qemu/user' unless (is_sle_micro || is_leap_micro);
 }


lib/main_micro_alp.pm
```diff
@@ -204,8 +204,8 @@ sub load_network_tests {
 sub load_qemu_tests {
     loadtest 'microos/rebuild_initrd' if is_s390x;
     loadtest 'qemu/info';
-    loadtest 'qemu/qemu';
-    loadtest 'qemu/kvm' unless is_aarch64;
+    loadtest 'qemu/qemu' unless is_rt;
+    loadtest 'qemu/kvm' unless (is_aarch64 or is_rt);
```
jlausuch, line 208: why not in aarch64? It seems to work fine.

jlausuch (changes requested):
Just to confirm, kvm is not supported on RT image. But it is in arm image...

lib/main_micro_alp.pm
waynechen55: I did not do anything to arm.  ```unless is aarch64``` already there.

waynechen55: It comes from here this file: schedule/sle-micro/virt.yaml

waynechen55: I double checked corresponding test in SLE Micro 5.4, there is no kvm test for aarch64 either, please refer to https://openqa.suse.de/tests/10855751

waynechen55: According to susedoc, it seems nested kvm is only supported with intel or amd cpu, please refer to https://susedoc.github.io/doc-sle/main/single-html/SLES-virtualization/#sec-vt-installation-nested-vms. There is no relevant description about how to enable this with aarch64.

mloviska: I think that nested virtualization is not supported on our arm workers that we have. This might have change by now though. Regarding RT, well the pattern is not preinstalled by default. I do not remember if it should work or not or if we should support it. In that case, it does not make much sense to run that test suite against the image.

mloviska: Generally, I would say running VMs on top of RT kernel does not seem to me like a valid user case. Or at least it is very peculiar IMHO

jlausuch: > Generally, I would say running VMs on top of RT kernel does not seem to me like a valid user case. Or at least it is very peculiar IMHO

Yes,I confirmed this. RT is not supporting virtualization.

jlausuch: > I did not do anything to arm. `unless is aarch64` already there. But aarch64 does not support nested kvm might be the reason, so there is no need to test kvm in qemu vm ? The possible reason does not originate from me, I just try my best to recall something that had ever crossed my brain.

ok, sorry, I miss-read it. Yes, it's ok then.

jlausuch (approved):

lib/main_micro_alp.pm
waynechen55: @mloviska Anything else ? Or can be merged ?

alice-suse: @jlausuch  Hi Jose, with clarification that RT is not supporting virtualization, JFYI, we will remove such test on RT in virtualization job group -- https://openqa.suse.de/tests/11768423.

alice-suse: BTW, it seems to me this PR can be merged. Will do it if no further reply before EOB today.

PR #17529, opened 2023-08-04
Avoid duplicate configuration in update_package.pm

tonyyuan1:
With host installation by autoyast, some configuration functionalities of install_package and update_package get duplicate so adding conditional statements to avoid their execution.

https://progress.opensuse.org/issues/132836

- Related ticket: https://progress.opensuse.org/issues/132836
- Needles: no
- Verification run: 
- prj2_host_upgrade_sles12sp5_to_developing_kvm, http://10.67.183.130/tests/3787
- prj2_host_upgrade_sles12sp5_to_developing_xen, http://10.67.183.130/tests/3803
- prj2_host_upgrade_sles15sp4_to_developing_xen (GUI install), https://openqa.suse.de/tests/11641701
- prj2_host_upgrade_sles15sp4_to_developing_xen, https://openqa.suse.de/tests/11637657
- prj2_host_upgrade_sles15sp4_to_developing_kvm, https://openqa.suse.de/tests/11739188
- -gi-guest_developing-on-host_developing-xen, http://10.67.183.130/tests/3775
- gi-guest_developing-on-host_developing-kvm, http://10.67.183.130/tests/3770
- uefi-gi-guest_sles15sp4-on-host_developing-xen,  https://openqa.suse.de/tests/11765772
- sev-es-gi-guest_developing-on-host_sles15sp4-kvm,  https://openqa.suse.de/tests/11777628

### DIFF
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -123,11 +123,11 @@ sub login_to_console {
         }
     }
 
-    unless (is_tumbleweed or check_screen([qw(grub2 grub1 prague-pxe-menu)], 210)) {
+    unless (is_tumbleweed or check_screen([qw(grub2 grub1 prague-pxe-menu)], get_var('AUTOYAST') && !get_var("NOT_DIRECT_REBOOT_AFTER_AUTOYAST") ? 1 : 180)) {
         ipmitool("chassis power reset");
         reset_consoles;
         select_console 'sol', await_console => 0;
-        check_screen([qw(grub2 grub1 prague-pxe-menu)], 90);
+        check_screen([qw(grub2 grub1 prague-pxe-menu)], 120);
     }
 
     # If a PXE menu will appear just select the default option (and save us the time)
--- a/tests/virt_autotest/reboot_and_wait_up.pm
+++ b/tests/virt_autotest/reboot_and_wait_up.pm
@@ -81,7 +81,9 @@ sub reboot_and_wait_up {
         reset_consoles;
 
         #wait boot finish and relogin
+        set_var('NOT_DIRECT_REBOOT_AFTER_AUTOYAST', '1');
         login_console::login_to_console($self, $reboot_timeout);
+
     }
 }
 
--- a/tests/virt_autotest/update_package.pm
+++ b/tests/virt_autotest/update_package.pm
@@ -48,10 +48,13 @@ sub update_package {
 sub run {
     my $self = shift;
     #workaroud: skip update package for registered aarch64 tests and because there are conflicts on sles15sp2 XEN
-    $self->update_package() unless (is_registered_sles && is_aarch64);
-    unless ((is_registered_sles && is_aarch64) || is_s390x) {
+    $self->update_package() unless (!!get_var('AUTOYAST') || is_registered_sles && is_aarch64);
+    unless (!!get_var('AUTOYAST') || (is_registered_sles && is_aarch64) || is_s390x) {
         set_grub_on_vh('', '', 'xen') if is_xen_host;
         set_grub_on_vh('', '', 'kvm') if is_kvm_host;
+    } else {
+        my @files_to_upload = ("/boot/grub2/grub.cfg", "/etc/default/grub");
+        upload_logs($_, failok => 1) foreach (@files_to_upload);
     }
     update_guest_configurations_with_daily_build();
 
--- a/products/sle/main.pm
+++ b/products/sle/main.pm
@@ -862,7 +862,7 @@ elsif (get_var("VIRT_AUTOTEST")) {
         loadtest "virt_autotest/install_package";
         loadtest "virt_autotest/update_package";
         loadtest "virt_autotest/reset_partition";
-        loadtest "virt_autotest/reboot_and_wait_up_normal" if get_var('REPO_0_TO_INSTALL');
+        loadtest "virt_autotest/reboot_and_wait_up_normal" if (!get_var('AUTOYAST') && get_var('REPO_0_TO_INSTALL'));
         loadtest "virt_autotest/download_guest_assets" if get_var("SKIP_GUEST_INSTALL") && is_x86_64;
     }
     if (get_var("VIRT_PRJ1_GUEST_INSTALL")) {
--- a/tests/virt_autotest/install_package.pm
+++ b/tests/virt_autotest/install_package.pm
@@ -103,7 +103,7 @@ sub install_package {
     if (get_var('GUEST_LIST', '') =~ /^win-.*/ && (is_xen_host)) { zypper_call '--no-refresh --no-gpg-checks in mkisofs' }
 
     #Subscribing packagehub from SLE 15-SP4 onwards that enables access to many useful software tools
-    virt_autotest::utils::subscribe_extensions_and_modules(reg_exts => 'PackageHub') if (is_sle('>=15-sp4') and !is_s390x);
+    virt_autotest::utils::subscribe_extensions_and_modules(reg_exts => 'PackageHub') if (!get_var('AUTOYAST') and is_sle('>=15-sp4') and !is_s390x);
 }
 
 sub run {


tests/virt_autotest/login_console.pm
```diff
@@ -121,7 +121,7 @@ sub login_to_console {
         }
     }
 
-    unless (is_tumbleweed or check_screen([qw(grub2 grub1 prague-pxe-menu)], 210)) {
+    unless (is_tumbleweed or check_screen([qw(grub2 grub1 prague-pxe-menu)], get_var('AUTOYAST') && !get_var("GUI_INSTALL") ? 1 : 210)) {
```
waynechen55, line 124: Why add this
```
get_var('AUTOYAST') && !get_var("GUI_INSTALL") ? 1 : 210
```
?

tests/virt_autotest/reboot_and_wait_up.pm
```diff
@@ -81,7 +81,9 @@ sub reboot_and_wait_up {
         reset_consoles;
 
         #wait boot finish and relogin
+        set_var('GUI_INSTALL', '1');
         login_console::login_to_console($self, $reboot_timeout);
+
```
waynechen55, line 86: Remove this empty line

tests/virt_autotest/reboot_and_wait_up.pm
```diff
@@ -81,7 +81,9 @@ sub reboot_and_wait_up {
         reset_consoles;
 
         #wait boot finish and relogin
+        set_var('GUI_INSTALL', '1');
```
waynechen55, line 84: Do we need this ```GUI_INSTALL``` ?

tests/virt_autotest/update_package.pm
```diff
@@ -48,8 +48,8 @@ sub update_package {
 sub run {
     my $self = shift;
     #workaroud: skip update package for registered aarch64 tests and because there are conflicts on sles15sp2 XEN
-    $self->update_package() unless (is_registered_sles && is_aarch64);
-    unless ((is_registered_sles && is_aarch64) || is_s390x) {
+    $self->update_package() unless (!!get_var('AUTOYAST') || is_registered_sles && is_aarch64);
```
waynechen55, line 51: Why use two exclamations ? This can be improved I think.

tests/virt_autotest/update_package.pm
```diff
@@ -48,8 +48,8 @@ sub update_package {
...
+    unless (!!get_var('AUTOYAST') || (is_registered_sles && is_aarch64) || is_s390x) {
```
waynechen55, line 52: Why use two exclamations ? This can be improved I think.

Julie-CAO (commented):
I checked that the 4 items in the part "Useless functions or test modules" described in https://progress.opensuse.org/issues/132836#note-5 have been skipped in this PR.

I see there are a few runs for prj2 tests , but more verification runs are required. Here are the list of tests we need verify(sorry for providing the list late):

- gi-guest_developing-on-host_developing-kvm
- gi-guest_developing-on-host_sles12sp5-xen
- sev-es-gi-guest_developing-on-host_sles15sp4-kvm
- uefi-gi-guest_sles15sp4-on-host_developing-xen

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: Agree, the purpose of this var is clear, but name is a little bit misleading for some situations, eg host is installed by AUTOYAST, in the afterwards reboot_and_wait_upxx, it will be set to GUI_INSTALL, which is opposite in logic.

tests/virt_autotest/update_package.pm
tonyyuan1: https://stackoverflow.com/questions/2168406/what-does-double-exclamation-point-mean

There are some usage examples in openqa source.  you can grep them.

tests/virt_autotest/reboot_and_wait_up.pm
tonyyuan1: The logic:  there's a "GUI installation mode value" which is 210.    In "afterwards reboot_and_wait_up" function  we always use this value no matter if the actual installation is GUI or autiyast.  There is no first login or second ...

What would you guys suggest to change?

tests/virt_autotest/login_console.pm
tonyyuan1: https://progress.opensuse.org/issues/128201#note-35

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: I do no think there is need to differentiate first login or second or third, for example, reboot_and_wait_up_normal can be called anywhere, and for host upgrade scenario, detecting grub after upgrade is meaningful (before host_upgrade_step3_run). 
Did you invent it or is being used widely in other tests already ?

tests/virt_autotest/update_package.pm
waynechen55: It seems ```!!``` will evaluate
```
everything greater than 0 to 1
everything smaller than 0 to 1
0 to empty
any string to 1
```
Is this what you want ?

tonyyuan1: It converts the value to a boolean.

tests/virt_autotest/reboot_and_wait_up.pm
tonyyuan1: It's a workaround. See https://progress.opensuse.org/issues/128201#note-35 and related comments in  that tickets.

waynechen55: I do not quite understand your change here. There is no need to change this part at all. And it seems that if you set GUI_INSTALL=0, then there will be actually no grub menu checking at all (because 1 second is not enough to capture anything and the system might not load grub menu at that time). I do not think this is acceptable. We use autoyast to install host which does not mean there is no need to detecting grub menu after install.

tests/virt_autotest/login_console.pm
waynechen55: I do not quite understand your change here. There is no need to change this part at all. And it seems that if you set GUI_INSTALL=0, then there will be actually no grub menu checking at all (because 1 second is not enough to capture anything and the system might not load grub menu at that time). I do not think this is acceptable. We use autoyast to install host which does not mean there is no need to detecting grub menu after install.

tests/virt_autotest/update_package.pm
waynechen55: You can print the result out, it converts 0 to empyt, and
```
everything greater than 0 to 1
everything smaller than 0 to 1
any string to 1
```

tests/virt_autotest/reboot_and_wait_up.pm
Julie-CAO: > 
> What would you guys suggest to change?

I suggest use `caller`.

waynechen55: > > What would you guys suggest to change?
> 
> I suggest use `caller`.

Maybe just add another extra line in login_console.pm, so do not modify any existing lines, for example, add following after two reset_consoles:
```
if (caller xxx and AUTOYAST and other required constraints) {
    reboot immediately
}
```
???
Straightforward and easy.

alice-suse:
I had a quick check if it is ready to merge. But several discussion threads are open. According to the comments, it seems to me not all points reach each other well. If needed, maybe some slack/teams talk can be launched  to better understand, quickly discussed, and agreed on solution?  
BTW, to me, I am fine with either solution -- caller or current way(but different name). Both have limitations, but both work.

Julie-CAO:
> I had a quick check if it is ready to merge. But several discussion threads are open. According to the comments, it seems to me not all points reach each other well. If needed, maybe some slack/teams talk can be launched to better understand, quickly discussed, and agreed on solution? BTW, to me, I am fine with either solution -- caller or current way(but different name). Both have limitations, but both work.

Since you think both are ok, the writer can choose either way. It is not a big case. If anyone is against it, please raise a talk.

tonyyuan1:
Then I set all resolved.  The PR is ready to merge if there is nothing else.

waynechen55:
> Then I set all resolved. The PR is ready to merge if there is nothing else.

It seems that this test has no grub2 menu screenshot, do you know the reason ? http://10.67.183.130/tests/3770

tonyyuan1:
I don't know why it didn't take a snapshot.  login_console starts with power reset so it has nothing to do with  autoyast installation.

This autoyast one has it. https://openqa.suse.de/tests/11739188#step/login_console/3

Why did this not have  grub either?  https://openqa.suse.de/tests/11168363#step/reboot_and_wait_up_upgrade/12

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: @tonyyuan1 If you finally decide to use current way -- using a var to differentiate, please update the variable name . See more details in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17529/files#r1286533531.

tonyyuan1: Do you mean "GUI_INSTALL"?
Note: it's just a temporary var.  What name do you recommend?

tests/virt_autotest/login_console.pm
waynechen55, line 124: I think it would be better to lower outer timeout to 180 and increase inner timeout to 120, so
```
unless (is_tumbleweed or check_screen([qw(grub2 grub1 prague-pxe-menu)], get_var('AUTOYAST') && !get_var("GUI_INSTALL") ? 1 : 180)) {
    ..........
    ..........
    check_screen([qw(grub2 grub1 prague-pxe-menu)], 120);
```

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: Yes. Any name not resulting in conflicting info is okay, since this is a workaround anyway for calling login_console,  after autoyast installation during which the grub menu already passed, but any other cases for calling login_console should keep the grub check(eg reboot_and_waitup_xx and original graphical host installation).     The name can be, for example,  KEEP_INITIAL_GRUB_CHECK.

Please note that if  only var name changed from `GUI_INSTALL`, and any other code unchanged in the PR, you will need to set the var to 1 in all testsuites that keep original graphical installation. Otherwise the first login_console after host installation, will mostly has no grub screen matching. So, IMHO, a minor improvement is to SET this var only during AUTOYAST installation, unset it afterwards. Then anywhere else won't need setting for this var at all. Up to you to decide. @Julie-CAO  FYI. Please pay attention to this point in the final OSD deployment during 15sp6.

tests/virt_autotest/update_package.pm
waynechen55, line 52: After this change, there will be no grub.cfg uploading anymore. Switching from step-by-step to autoyast install already leads to no grub.cfg/grub default uploading, it seems that they will be uploaded in ```update_pacakage.pm```. At last, even ```update_package.pm``` does not do this. I think this is a big gap compared to previous functionality.

waynechen55: The easiest way to solve the problem is as below I think:
```
unless (!!get_var('AUTOYAST') || (is_registered_sles && is_aarch64) || is_s390x) {
    xxxxxx
}
else {
    upload_logs("grub.cfg and grub default file");
}
```

tests/virt_autotest/login_console.pm
alice-suse: Please be aware that default TIMEOUT_SCALE is 3, so the real timeout value is the value * 3.

waynechen55: I think 180 and 120 is more suitable for general case. There should not be such big difference between outer and inner timeout, namely the first and second try.

tests/virt_autotest/reboot_and_wait_up.pm
Julie-CAO: Hi @alice-suse, Based on current logic, we don't need add `GUI_INSTALL` setting to any test suite. It is set in `reboot_and_wait_up.pm`. Only tests with autoyast installation will skip grub check in login_console.pm.

waynechen55: > Hi @alice-suse, Based on current logic, we don't need add `GUI_INSTALL` setting to any test suite. It is set in `reboot_and_wait_up.pm`. Tests with autoyast installation will skip grub check in login_console.pm.

1. This var ```GUI_INSTALL``` is really confusing. If author wants use variable to differentiate first boot after autoyast from other boots, a better name conveying correct message is required. For example, "reboot_after_autoyast" or similar. I do think the name of this var is very acceptable.
2. If autoyast install, there will be only grub checking once in any case, namely the inner check_screen in unless block.

tests/virt_autotest/login_console.pm
alice-suse: Agree. I suppose Tony did not do this because it is a separate improvement from autoyast installation. But since it is minor effort, doing it together is also nice choice.  I am fine with either choice.

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: > > Hi @alice-suse, Based on current logic, we don't need add `GUI_INSTALL` setting to any test suite. It is set in `reboot_and_wait_up.pm`. Tests with autoyast installation will skip grub check in login_console.pm.
> 

@Julie-CAO  Hi Julie, I hope you get my precondition for this comment -- "you will need to set the var to 1 in all testsuites **that keep original graphical installation**".


> 1. This var `GUI_INSTALL` is really confusing. If author wants use variable to differentiate first boot after autoyast from other boots, a better name conveying correct message is required, just like `UPGRADE_AFTER_REBOOT` and `REBOOT_AFTER_UPGRADE` for offline host upgrade in login_console.pm. For example, "reboot_after_autoyast" or similar. I do think the name of this var is very acceptable if author only wants to differentiate first boot after autoyast from other boots. And I do not think we should bother GUI install (literally) at all.

Agree. In https://progress.opensuse.org/issues/128201#note-35, I initially suggested a similar name ,  `FIRST_REBOOT_AFTER_AUTOYAST_INST` .  I also sugested that this var was better to be set in autoyast installation module, and unset it after the first boot grub check in login_console.

tests/virt_autotest/login_console.pm
waynechen55: @tonyyuan1 It would be better you can help update this. This is very trivial. Thanks.

tonyyuan1: No problem. Did you test to verify that the two new values work as expected?

waynechen55: When you finish all updates/fixes, you can help re-run some verifications together. Should be enough.

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: @tonyyuan1 Just summarize for you. New name suggestions for `GUI_INSTALL` -- `REBOOT_AFTER_AUTOYAST`, `FIRST_REBOOT_AFTER_AUTOYAST_INST`, or `KEEP_INITIAL_GRUB_CHECK` . Seems good to pick one from the first two options.

tonyyuan1: After reading all of comments here, It seems that:
A VAR name just can't confuse people. It might be that  the logic behind it is a bit difficult to understand.
'GUI_INSTALL' is short for "GUI installation mode" which is related to the logic.  Changing it makes it irrelevant to logic. Therefore, I add a inline comment to explain them. Let me know if it look better to you.

tests/virt_autotest/login_console.pm
tonyyuan1: I want get this ticket over with soon.  can't effort to rerun SOME.  Just one http://10.67.183.130/tests/3824

Maybe I do it with my next PR or you do it with your PR if it's not enough.

tests/virt_autotest/reboot_and_wait_up.pm
waynechen55: > After reading all of comments here, It seems that: A VAR name just can't confuse people. It might be that the logic behind it is a bit difficult to understand. 'GUI_INSTALL' is short for "GUI installation mode" which is related to the logic. Changing it makes it irrelevant to logic. Therefore, I add a inline comment to explain them. Let me know if it look better to you.

It depends on why you needs this variable. If it just differentiates first boot after autoyast from other boots, it has to be renamed.

waynechen55: @tonyyuan1 Does your ```GUI_INSTALL``` refer to previous step-by-step text installation ? If this is the case, I do not think it is very good. It can kind of mix up text mode and real graphical interface, and I think you should consider other variable names. And using ```GUI_INSTALL``` does not reflect what you are trying to solve, and it seems that using it indicates a situation of different nature.

tests/virt_autotest/update_package.pm
waynechen55: @tonyyuan1 Please also help update this part. And use failok=> 1 with upload_logs to make it safe.

tonyyuan1:
I implemented the code based on the logic about "GUI installation mode". It was tested and verified. You guys are trying to understand it by a logic about "first boot" or something.  That of cause might lead to confusing you.  
I don't care about a VAR name but my logic will not be able to explain the code well after  "GUI_INSTALL" is changed.  I don't know if you logic can explain it either. 
Anyway, I don't want us to waste time on this any more.  Do you guys still  want to change it? I will change if yes.

tests/virt_autotest/reboot_and_wait_up.pm
alice-suse: To me, it seems better with the comment explanation. 
Given that Wayne still has questions and suggestions on this, in addition to https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17529/files#r1295552604,  I will hold on the next move, until you guys make things clear to each other and get agreement. I support NECESSARY items with minor effort to be done in this PR before merge, while   others maybe not.

waynechen55:
> I implemented the code based on the logic about "GUI installation mode". It was tested and verified. You guys are trying to understand it by a logic about "first boot" or something. That of cause might lead to confusing you. I don't care about a VAR name but my logic will not be able to explain the code well after "GUI_INSTALL" is changed. I don't know if you logic can explain it either. Anyway, I don't want us to waste time on this any more. Do you guys still want to change it? I will change if yes.

Yes, I think you'd better help change it. I think ```REBOOT_AFTER_AUTOYAST``` is enough. By the way, I think Julie's suggestion about use caller is better because it is even more clear and accurate, but using ```REBOOT_AFTER_AUTOYAST``` should also suffice if you prefer to use a var. Thanks.

tests/virt_autotest/update_package.pm
tonyyuan1: This is out of scope of this ticket. I will do this time. Please try not to ask to do this in the future.

waynechen55: It falls into this pull request, because the original grub config uploading happens in ```set_grub_on_vh('', '', 'xen') ``` which is skipped after this change. It should be kept, because we already skip the first time grub menu needle matching.

waynechen55:
> > I implemented the code based on the logic about "GUI installation mode". It was tested and verified. You guys are trying to understand it by a logic about "first boot" or something. That of cause might lead to confusing you. I don't care about a VAR name but my logic will not be able to explain the code well after "GUI_INSTALL" is changed. I don't know if you logic can explain it either. Anyway, I don't want us to waste time on this any more. Do you guys still want to change it? I will change if yes.
> 
> Yes, I think you'd better help change it. I think `REBOOT_AFTER_AUTOYAST` is enough. By the way, I think Julie's suggestion about use caller is better because it is even more clear and accurate, but using `REBOOT_AFTER_AUTOYAST` should also suffice if you prefer to use a var. Thanks.

@tonyyuan1 I understand you want to maintain your current logic regarding the use of the var, so ```REBOOT_AFTER_AUTOYAST``` may break your logic because it needs to be set to 1 in the first place and then back to 0 later. And, at the same time, the name ```GUI_INSTALL``` really can not explain itself based on its usage, so after thinking twice, I think using ```NOT_A_REBOOT_AFTER_AUTOYAST``` might be easier for you, because you just only need to change the name and nothing else (keeping your logic).

tonyyuan1:
> 
> @tonyyuan1 I understand you want to maintain your current logic regarding the use of the var, so `REBOOT_AFTER_AUTOYAST` may break your logic because it needs to be set to 1 in the first place and then back to 0 later. And, at the same time, the name `GUI_INSTALL` really can not explain itself based on its usage, so after thinking twice, I think using `NOT_DIRECT_REBOOT_AFTER_AUTOYAST` might be easier for you, because you just only need to change the name and nothing else (keeping your logic).

Did you read the inline comment about my logic that I had added in tests/virt_autotest/login_console.pm? Does that clarify a bit about `GUI_INSTALL` for you?

waynechen55:
> > @tonyyuan1 I understand you want to maintain your current logic regarding the use of the var, so `REBOOT_AFTER_AUTOYAST` may break your logic because it needs to be set to 1 in the first place and then back to 0 later. And, at the same time, the name `GUI_INSTALL` really can not explain itself based on its usage, so after thinking twice, I think using `NOT_DIRECT_REBOOT_AFTER_AUTOYAST` might be easier for you, because you just only need to change the name and nothing else (keeping your logic).
> 
> Did you read the inline comment about my logic that I had added in tests/virt_autotest/login_console.pm? Does that clarify a bit about `GUI_INSTALL` for you?

Yes. Your explanation let me know how you want to use this var and I think there is nothing wrong the way in which you use it, but to be honest, it does not justify the name ```GUI_INSTALL``` from my perspective. So in order to maintain your usage and logic, I suggest ```NOT_DIRECT_REBOOT_AFTER_AUTOYAST```, only name change, light and easy. But after this long discussion, if you still insist keeping using ```GUI_INSTALL```, then you can keep it.

Julie-CAO:
I vote to change the name. The name itself is confusing. This is not a big PR but you spent much time in explaining the logic, it indicates the change brings confuse to us.

tonyyuan1:
> But I have to point out `GUI_INSTALL` will be changed from 0 to 1 in your code, which means host installation method changes in a single test literally. I think this is the direct message that this var conveys, if you can have a more proper var name, why bother using inline comment to explain its real purpose ? It looks like a sign post on road accompanied by text "rotate clockwise by 90 degree".

Alice wants that inline comment kept.  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17529#discussion_r1297939835.  What new name would you suggest if we keep  inline comment?

waynechen55:
> > But I have to point out `GUI_INSTALL` will be changed from 0 to 1 in your code, which means host installation method changes in a single test literally. I think this is the direct message that this var conveys, if you can have a more proper var name, why bother using inline comment to explain its real purpose ? It looks like a sign post on road accompanied by text "rotate clockwise by 90 degree".
> 
> Alice wants that inline comment kept. [#17529 (comment)](https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17529#discussion_r1297939835). What new name would you suggest if we keep inline comment?

VAR NAME:
NOT_DIRECT_REBOOT_AFTER_AUTOYAST

INLINE COMMENT:
NOT_DIRECT_REBOOT_AFTER_AUTOYAST to differentiate intended direct reboot from other subsequent reboots after autoyast installation, module of which reboots system automatically, because the immediate next test module queries boot process which may already finishes. All the other subsequent queries/reboots are not affected.

Julie-CAO:
> Is this ok for you, @alice-suse @Julie-CAO

Better than GUI.

tests/virt_autotest/update_package.pm
```diff
@@ -48,10 +48,13 @@ sub update_package {
...
         set_grub_on_vh('', '', 'xen') if is_xen_host;
         set_grub_on_vh('', '', 'kvm') if is_kvm_host;
+    } else {
+        my @files_to_upload = ("/boot/grub2/grub.cfg", "/etc/default/grub");
+        upload_logs($_, failok => 1) foreach (@files_to_upload);
```
waynechen55, line 57: Can we have one verification run for this log uploading ? Thanks. @tonyyuan1

tonyyuan1: I made one right before last commit. 
http://10.67.183.130/tests/3865#

waynechen55: Changes take effect.

waynechen55:
> > Then I set all resolved. The PR is ready to merge if there is nothing else.
> 
>     1. It seems that this test has no grub2 menu screenshot, do you know the reason ? http://10.67.183.130/tests/3770
> 

I found this verification run has TIMEOUT_SCALE=3, but all assert/check_screens do not multiply their timeouts accordingly. @tonyyuan1 What's wrong with this verification run ?

tonyyuan1:
> > It seems that this test has no grub2 menu screenshot, do you know the reason ? http://10.67.183.130/tests/3770
> 
> I found this verification run has TIMEOUT_SCALE=3, but all assert/check_screens do not multiply their timeouts accordingly. @tonyyuan1 What's wrong with this verification run ?

I once tried to use set_var("TIMEOUT_SCALE",1) to save some time. Maybe that was its run. Long story. see the ticket for detail.

alice-suse:
I had a final check, it seems all agreed points have been updated to latest code. Thank you, Tony. I can't tell whether verification runs have been updated(no such comment), but I suppose @tonyyuan1  has done that? If yes, then plan to merge it today. If any different opinion from any one, please raise comment.

alice-suse (approved):

alice-suse:
Won't wait further. Consider it as validation done. Merge it.

PR #17527, opened 2023-08-03
Refactor reset_container_network_if_needed

grisu48:
Extract the commonly used reset_container_network_if_needed function to avoid code duplication.

- Related PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17512
- Verification run: TBD

PR #17535, opened 2023-08-04
Revert "Refactor reset_container_network_if_needed"

Vogtinator:
Reverts os-autoinst/os-autoinst-distri-opensuse#17527

It caused https://progress.opensuse.org/issues/133808

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -22,7 +22,7 @@ use Mojo::Util 'trim';
 
 our @EXPORT = qw(test_seccomp runtime_smoke_tests basic_container_tests get_vars
   can_build_sle_base get_docker_version get_podman_version check_runtime_version
-  check_min_runtime_version container_ip container_route registry_url reset_container_network_if_needed);
+  check_min_runtime_version container_ip container_route registry_url);
 
 sub test_seccomp {
     my $no_seccomp = script_run('docker info | tee /tmp/docker_info.txt | grep seccomp');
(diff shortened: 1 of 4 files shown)


jlausuch (approved):

mloviska:
Isn't the breaking commint this one ?
https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/7d46a0d9000594ad3cc72bd4d15ab8eee0315411

Vogtinator:
> Isn't the breaking commit this one ? [7d46a0d](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/7d46a0d9000594ad3cc72bd4d15ab8eee0315411)

No, it was already broken before that commit: https://openqa.opensuse.org/tests/3479576

mloviska:
> > Isn't the breaking commit this one ? [7d46a0d](https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/7d46a0d9000594ad3cc72bd4d15ab8eee0315411)
> 
> No, it was already broken before that commit: https://openqa.opensuse.org/tests/3479576

Damn, this will break BCI tests though. Let's hope there won't be many builds over the weekend.

Vogtinator:
We could also wait for a proper fix, if it won't take too long. For Leap I already tagged all failures appropriately and TW doesn't appear to be affected, so at least the current fallout isn't severe. WDYT?

mloviska:
> We could also wait for a proper fix, if it won't take too long. For Leap I already tagged all failures appropriately and TW doesn't appear to be affected, so at least the current fallout isn't severe. WDYT?

I think reverting is a better solution for now, Felix is already out

PR #17539, opened 2023-08-07
Improve libvirt checks for login_console test module

nanzhg:
1. To improve the function 'is_monolithic_libvirtd()', adding a warning message for checking whether libvirt package installed or not. (lib/virt_autotest/utils.pm)
2. Skip libvirt service checks for kubevirt tests. (tests/virt_autotest/login_console.pm)

- Related ticket: https://progress.opensuse.org/issues/132608
- Verification run: https://openqa.suse.de/tests/11760650

### DIFF
--- a/lib/virt_autotest/utils.pm
+++ b/lib/virt_autotest/utils.pm
@@ -159,6 +159,7 @@ sub is_kvm_host {
 
 #retrun 1 if libvirt 9.0- is running which monolithic libvirtd is the default service
 sub is_monolithic_libvirtd {
+    record_info('WARNING', 'Libvirt package is not installed', result => 'fail') if (script_run('rpm -q libvirt-libs'));
     unless (is_alp) {
         return 1 if script_run('rpm -q libvirt-libs | grep -e "libs-9\.0" -e "libs-[1-8]\."') == 0;
     }
--- a/tests/virt_autotest/login_console.pm
+++ b/tests/virt_autotest/login_console.pm
@@ -65,7 +65,9 @@ sub check_kvm_modules {
     # for modular libvirt, virtqemud is expected in "loaded: active or inactive" status.
     # virtqemud.socket seems to be always in "loaded: active" status
     unless (is_monolithic_libvirtd) {
-        die 'virtqemud.socket is not running!' unless script_run("systemctl is-active virtqemud.socket") eq 0;
+        unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/ or script_run("systemctl is-active virtqemud.socket") eq 0) {
+            die 'virtqemud.socket is not running!';
+        }
     }
     record_info("KVM", "kvm modules are loaded!");
 }


Julie-CAO:
LGTM

lib/virt_autotest/utils.pm
```diff
@@ -159,6 +159,7 @@ sub is_kvm_host {
 
 #retrun 1 if libvirt 9.0- is running which monolithic libvirtd is the default service
 sub is_monolithic_libvirtd {
+    record_info('WARNING', 'Libvirt package is not installed', result => 'fail') if (script_run('rpm -q libvirt-libs'));
```
alice-suse, line 162: How will the result look like? When  is_monolithic_libvirtd is called, but no libvirt-libs is installed, how will openqa webui look like for the overall result and this record_info box?

Julie-CAO:
we also need verify on sle tests and TW tests. but since TW tests are blocked, let's have sle tests verified.

lib/virt_autotest/utils.pm
nanzhg: It will finally fail the test.
https://openqa.suse.de/tests/11758005#step/login_console/11

waynechen55: ```rpm -q libvirt-libs``` will fail on leap15.4 kvm host

tests/virt_autotest/login_console.pm
```diff
@@ -64,8 +64,10 @@ sub check_kvm_modules {
 
     # for modular libvirt, virtqemud is expected in "loaded: active or inactive" status.
     # virtqemud.socket seems to be always in "loaded: active" status
-    unless (is_monolithic_libvirtd) {
-        die 'virtqemud.socket is not running!' unless script_run("systemctl is-active virtqemud.socket") eq 0;
+    unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/) {
```
waynechen55, line 67: This issue is specific to kubevirt-tests ?

lib/virt_autotest/utils.pm
nanzhg: > I found libvirt-libs is dependency of libvirt. So it should be installed by default ?

Yes, It's the base package of libvirt.

nanzhg:
> we also need verify on sle tests and TW tests. but since TW tests are blocked, let's have sle tests verified.

It passed login_console test module with SLE, https://openqa.suse.de/tests/11765393#

lib/virt_autotest/utils.pm
waynechen55: libvirt-libs installed by default when you choose to install kvm host pattern together with SLE Micro 5.5. I verified this on local machine.

tests/virt_autotest/login_console.pm
nanzhg: Kubevirt tests should ignore the libvirt checking, its tests do not depends on libvirt.

lib/virt_autotest/utils.pm
alice-suse: > It will finally fail the test. https://openqa.suse.de/tests/11758005#step/login_console/11

Looks good, thanks

tests/virt_autotest/login_console.pm
waynechen55: I do not think the code orgnization here is good. You should skip your checking inside instead of outside, for example:
```
unless (is_monolithic_libvirtd) {
  unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/
}
```
Code block here should be grouped by checkings instead of tests.

Julie-CAO: if so, his kubevirt test will get a failed result finally.

waynechen55: He can switch the testing logic order. Put test name checking inside to skip.

nanzhg: If the code logic  goes into the function "is_monolithic_libvirtd", kubevirt tests will fail because it doesn't install any libvirt packages. That's why I put it outside of the judgment of function "is_monolithic_libvirtd".
```
sub is_monolithic_libvirtd {
    record_info('WARNING', 'Libvirt package is not installed', result => 'fail') if (script_run('rpm -q libvirt-libs'));
```

waynechen55: Why not install kvm pattern together during host installation phase ? You can just set PATTERNS=default,kvm. It is easy and no need to modify code at all.

alice-suse: @waynechen55  JFYI. For kubevirt test, it requires the host to have kvm module installed and loaded, but it does not need qemu/virtualization stake, because they will be given in the kubevirt containers. Does that help for reviewing the PR? 
If you still have further comments, feel free to raise. I suppose @nanzhg  will follow and  reply. Otherwise, maybe we can consider merging?

waynechen55:
```
unless (is_monolithic_libvirtd) {
    unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/) {
        die 'virtqemud.socket is not running!' unless script_run("systemctl is-active virtqemud.socket") eq 0;
    }
}
```
This will still fail you kubevirt test ? @nanzhg 
I just think checking here should be grouped by types of checkings instead of test names.

nanzhg: > Why not install kvm pattern together during host installation phase ? You can just set PATTERNS=default,kvm. It is easy and no need to modify code at all. The functionality already implemented which is ready to be used instantly.

I tried with the parameter PATTERNS=default,kvm, looks like it didn't take effect.
https://openqa.suse.de/tests/11768689#step/login_console/14

waynechen55: > > Why not install kvm pattern together during host installation phase ? You can just set PATTERNS=default,kvm. It is easy and no need to modify code at all. The functionality already implemented which is ready to be used instantly.
> 
> I tried with the parameter PATTERNS=default,kvm, looks like it didn't take effect. https://openqa.suse.de/tests/11768689#step/login_console/14

This will work ?
```
unless (is_monolithic_libvirtd) {
    unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/ or script_run("systemctl is-active virtqemud.socket") eq 0) {
        die 'virtqemud.socket is not running!';
    }
}
```
Better ?

nanzhg: > This will still fail you kubevirt test ? @nanzhg is_monolithic_libvirtd will return 0 or 1 and does not have assertion. I just think code here should be grouped by types of checkings instead of test names.

Yes, is_monolithic_libvirtd will return 0. When it begins to execute "is_monolithic_libvirtd", kubevirt tests will be failed as the WARNING statement will set the whole test as failed.

waynechen55: > > This will still fail you kubevirt test ? @nanzhg is_monolithic_libvirtd will return 0 or 1 and does not have assertion. I just think code here should be grouped by types of checkings instead of test names.
> 
> Yes, is_monolithic_libvirtd will return 0. When it begins to execute "is_monolithic_libvirtd", kubevirt tests will be failed as the WARNING statement will set the whole test as failed.

I do no think we should fail inside a ```is_xxx``` function. ```is_xxx``` should return true or false to let test decide what to do next. You can help make it better, but not let it fail. I do no think it is very good.

nanzhg: > This will work ?
> 
> ```
> unless (is_monolithic_libvirtd) {
>     unless (get_var('TEST_SUITE_NAME') =~ /kubevirt-tests/ or script_run("systemctl is-active virtqemud.socket") eq 0) {
>         die 'virtqemud.socket is not running!';
>     }
> }
> ```
> 
> Better ?

The first statement will fail the whole test on sle-micro. E.g. https://openqa.suse.de/tests/11758005#step/login_console/11

`unless (is_monolithic_libvirtd) {
`

nanzhg: @waynechen55 Probably I was wrong, the record_info statement will not fail the test after double checking with a whole test (http://10.67.129.96/tests/2492). I also updated the logic of check_kvm_modules function as you suggested.

waynechen55: > @waynechen55 Sorry I was wrong. After double checking with a whole test (http://10.67.129.96/tests/2492#step/login_console/11), the WARNING statement will not fail the test, it just marked info red. I also updated the logic of check_kvm_modules function as you suggested.

Looks better now

alice-suse:
It seems all comments are resolved. Plan to merge if no more comments.

lib/virt_autotest/utils.pm
waynechen55, line 162: I still think there is no need to give a "fail" result even for record_info inside a ```is_xxx``` function. For example, ```is_xxx``` function only responsible for answering "yes" or "no". If a system just need to know whether it is monolithic_libvirtd, but it gets a "fail" flag, this is not very reasonable. Any decision, including recording a failure info, should be done outside of ```is_xxx```.

Julie-CAO: this function is not a tempoary function to differentiate it is modular libvirt. it will be obsolete when modular libvirt becomes to be the default virtualization. I even had thought add a setting 'IS_MONOLITHIC_LIBVIRT=1' for hosts with older products.

We know It is not perfect function, because there are 3 possible results, 'yes', 'no' and 'no libvirt'. but it has been widely used in virt tests as we overlooked the 'no libvirt'. I think it is the best solution with minimal effort currently we can find out as SLEM 5.5 is waiting for green tests.

We can think of making it better in the future when it is really needed.

waynechen55: OK. Fine.

PR #17543, opened 2023-08-07
push boottime data updated proceed on failure option

m-dati:
In publiccloud the routine to push boottime data to InfluxDB  the _proceed on failure_ option has been updated in order to handle script_output errors and _timeout_ cases too.

- Related ticket: https://progress.opensuse.org/issues/133871
- Needles: N/A
- Verification run: TDB, see next posts

### DIFF
--- a/lib/db_utils.pm
+++ b/lib/db_utils.pm
@@ -75,7 +75,7 @@ sub influxdb_push_data {
     $out =~ s/$token/<redacted>/;
     record_info('curl POST', $out);
 
-    my $output = script_output($cmd, quiet => $args{quiet});
+    my $output = script_output($cmd, quiet => $args{quiet}, proceed_on_failure => $args{proceed_on_failure});
     my ($return_code) = $output =~ /RETURN_CODE:(\d+)/;
     unless ($return_code >= 200 && $return_code < 300) {
         my $msg = "Failed pushing data into Influx DB:\n$output\n";


lib/db_utils.pm
```diff
@@ -75,7 +75,7 @@ sub influxdb_push_data {
     $out =~ s/$token/<redacted>/;
     record_info('curl POST', $out);
 
-    my $output = script_output($cmd, quiet => $args{quiet});
+    my $output = script_output($cmd, quiet => $args{quiet}), $args{proceed_on_failure});
```
asmorodskyi, line 78:
```suggestion
    my $output = script_output($cmd, quiet => $args{quiet}, $args{proceed_on_failure});
```

asmorodskyi:
```suggestion
    my $output = script_output($cmd, quiet => $args{quiet}, proceed_on_failure => $args{proceed_on_failure});
```

grisu48: This will still not work, because `$return_code` will be undefined in the next line and then in line 80 the `unless` will be executed.

In short - this alone is insufficient, the test will still fail on timeouts.

To solve this you could use `script_run` to run `$cmd` and pipe the output to a temporary file. Then you can use `script_output` to extract the contents of that file. Like this you can check the return code and get the output.

asmorodskyi (approved):
LGTM

m-dati:
VR https://openqa.suse.de/tests/11759555#step/prepare_instance/254
I set a fake port 9999, to have failure and indeed now the test doesn't fail anymore and proceeds after expired the standard timeout for _script_output_ command [90sec].

PR #17546, opened 2023-08-07
Not die on empty repos when embargoed updates detected

asmorodskyi:


pdostal (changes requested):
I don't understand what `PUBLIC_CLOUD_NO_EMBARGOED_UPDATES` means.

asmorodskyi:
> I don't understand what `PUBLIC_CLOUD_NO_EMBARGOED_UPDATES` means.

it is a flag which signals that embargoed updates was detected in the queue , when it equals 0 ( **NO_EMBARGOED_UPDATES** is false so there are embargoed updates  )  we not dying if repos are empty

pdostal:
Can you please do `PUBLIC_CLOUD_EMBARGOED` and then if that variable is true we don't die?

m-dati:
I also agree that global boolean variables should always have **Positive Logic** (1=do it, 0=doesn't) and names reflecting this expectation, to avoid confusion in values definitions.

Moreover the `variables.md` should be updated with related variable description.

asmorodskyi:
> Can you please do `PUBLIC_CLOUD_EMBARGOED` and then if that variable is true we don't die?

Done , IMHO now `die "Empty test repositories" if (!get_var("PUBLIC_CLOUD_EMBARGOED_UPDATES_DETECTED") && $check_empty_repos && $size < 100);` line looks more messy than in mine version ( and this is why I come up with mine version in first place ) but OK let's keep it like this

asmorodskyi:
> Moreover the `variables.md` should be updated with related variable description.

variables.md is updated

pdostal (approved):
You could just ignore my suggestion but I really like this version more.

Thank you Anton.

m-dati:
Now thinking on these `set_var` actions in L78, L71, it seems to me that this one could be an _internal_ perl variable instead, being not possible, manually setting it from externally the code when desired, to keep that value and address test code flows, like all other in variables.md instead: am I wrong ?

m-dati (commented):
a

asmorodskyi:
> Now thinking on these `set_var` actions in L68, L71, it seems to me that this one could be an _internal_ perl variable instead, being not possible, manually setting it from externally the code when desired, to keep that value and address test code flows, like all other in variables.md instead: am I wrong ?

my idea was that in this way this variable can be used in any other module if we would like to do something based on the fact that current run proved to have embargoed updates

tests/publiccloud/download_repos.pm
```diff
@@ -65,8 +65,12 @@ sub run {
         my $reject = "'robots.txt,*.ico,*.png,*.gif,*.css,*.js,*.htm*'";
         my $regex = "'s390x\\/|ppc64le\\/|kernel*debuginfo*.rpm|src\\/'";
         my ($incident, $type);
+        set_var("PUBLIC_CLOUD_EMBARGOED_UPDATES_DETECTED", 0);
```
mpagot, line 68: Do we need to use settings?Is it not a variable enough?

pdostal: I didn't get what you mean.

mpagot: why not to define a local variable
```
my $embargoed_updates_detected = 0;
```

And just use it?

pdostal: Because you cannot change the local variable without changing the code.

asmorodskyi: on top of that - now this variable can me used in different `*.pm` files . If you make it local perl variable you can not read/write this state in other perl modules due to nature of how openQA run modules

PR #17550, opened 2023-08-08
Restart firewalld only if active

grisu48:
Restart the firewall only if it was previously already active to avoid (bsc#1214080)[https://bugzilla.suse.com/show_bug.cgi?id=1214080]

- Related ticket: https://progress.opensuse.org/issues/133808
- Verification run: https://openqa.opensuse.org/tests/3492898

### DIFF
--- a/lib/containers/utils.pm
+++ b/lib/containers/utils.pm
@@ -310,7 +310,7 @@ sub reset_container_network_if_needed {
             } elsif ($current_engine eq 'docker') {
                 systemctl("start docker");
             }
-            systemctl("restart firewalld");
+            systemctl("restart firewalld") if (script_run("systemctl is-active firewalld") == 0);
         }
     }
 }


ricardobranco777 (changes requested):
I would add a soft-fail with the link to the bug you opened: https://bugzilla.suse.com/show_bug.cgi?id=1214080

At the very least a comment on why this is necessary.

mloviska:
> I would add a soft-fail with the link to the bug you opened: https://bugzilla.suse.com/show_bug.cgi?id=1214080
> 
> At the very least a comment on why this is necessary.
> 
> Otherwise LGTM.

Agreed, I would also add the podman related bug

grisu48:
> I would add a soft-fail with the link to the bug you opened: https://bugzilla.suse.com/show_bug.cgi?id=1214080
> 
> At the very least a comment on why this is necessary.
> 
> Otherwise LGTM.

Agreed, it becomes obvious that we need to reference the two softfailures. Since those two bsc are not easy to trigger (require specific environment setup), I'll move them to a separate test. It does not make sense to add a softfailure everywhere where we apply those workaround because they are omnipresent.

Let's move forward with this PR and I will add corresponding softailures in an additional test run for the container engine?

ricardobranco777 (approved):
LGTM

PR #17554, opened 2023-08-09
NFS: Schedule stress-ng on NFS

schlad:


schedule/storage/nfs.yaml
```diff
@@ -15,8 +15,10 @@ conditional_schedule:
         - kernel/nfs_server
       nfs_client:
         - kernel/nfs_client
+    VERSION:
+      15-SP6:
+        - kernel/nfs_stress_ng
```
czerw, line 20: I would do it like https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/5f6c381c38b902af90b1418f4b3f3e3e2b8c9965/schedule/kernel/prepare_baremetal.yaml#L10

schedule/storage/nfs.yaml
```diff
@@ -15,8 +15,10 @@ conditional_schedule:
         - kernel/nfs_server
       nfs_client:
         - kernel/nfs_client
+    VERSION:
+      15-SP6:
```
czerw, line 19: What about tumbleweed?

schedule/storage/nfs.yaml
```diff
@@ -15,8 +15,10 @@ conditional_schedule:
...
+        - kernel/nfs_stress_ng
 schedule:
   - boot/boot_to_desktop
   - kernel/before_nfs_test
   - '{{nfstest}}'
-  - kernel/nfs_stress_ng
```
czerw, line 22: Please make separate condition for nfs_stress_ng to don't mix it with server/client logic.

czerw (approved):

czerw:
Could you please do one verification run for maintenance and one for TW please?

czerw:
Issues were related to multimachine network setup. We don't need to exclude test from running at all for maintenance,

PR #17555, opened 2023-08-09
Improve the usage of nfs when guest migration

RoyCai7:
Progress ticket: https://progress.opensuse.org/issues/133139 
Utilizing NFS share storage in a virtualization environment: 
either by creating a netfs storage pool on both source and destination hosts or by directly mounting the NFS directory on both hosts, with the current testing setup involving only mounting the NFS directory on the destination host without mounting the NFS directory on the source host.
The purpose of doing this is to ensure that the same file system is accessed during guest migration.

- Related ticket: Progress ticket: https://progress.opensuse.org/issues/133139 
- Needles: N/A
- Verification run: 
    - [Xen Guest migration](http://openqa.qam.suse.cz/tests/overview?distri=sle&version=15-SP5&build=:cailftest:live_migrat)
    - [Kvm Guest migration](http://openqa.qam.suse.cz/tests/overview?build=:cailftest:KVM:live_migrat&distri=sle&version=15-SP5&groupid=167)

### DIFF
--- a/lib/parallel_guest_migration_base.pm
+++ b/lib/parallel_guest_migration_base.pm
@@ -541,15 +541,16 @@ sub check_host_gid {
 =head2 config_host_shared_storage
 
 Configure shared nfs storage on server and mount it on client. Main arguments
-are type of shared storage, path of exported shared storage, server or client 
-and mount path on client. 
+are type of shared storage, path of exported shared storage, server or client
+, original source path on server and mount path on server and client. 
 =cut
 
 sub config_host_shared_storage {
     my ($self, %args) = @_;
     $args{type} //= 'nfs';
-    $args{exppath} //= '/var/lib/libvirt/images';
+    $args{exppath} //= '/home/virt';
     $args{role} //= 'server';
+    $args{srcpath} //= '/var/lib/libvirt/images';
     $args{mntpath} //= '/var/lib/libvirt/images';
 
     record_info("Configure host shared storage");
@@ -570,12 +571,22 @@ sub config_host_shared_storage {
             }
             $_temppath = $args{exppath};
             $_temppath =~ s/\//\\\//g;
+            # Create export path directory if it doesn't exist
+            if (script_run("test -d $_temppath") != 0) {
+                assert_script_run("mkdir -p $_temppath");
+            }
+            # Move guest images from the default image path to the export path
+            # Guest installation puts images to /var/lib/libvirt/images by default,
+            # move them to nfs export directory /home/virt, then mount /home/virt to /var/lib/libvirt/images.
+            assert_script_run("mv $args{srcpath}/* $args{exppath}/");
             assert_script_run("sed -i \'/^.*$_temppath.*\$/d\' /etc/exports");
             assert_script_run("echo \"$args{exppath} *(rw,sync,no_root_squash,no_subtree_check)\" >> /etc/exports");
             assert_script_run("exportfs -a");
             systemctl('restart nfs-server.service');
             systemctl('status nfs-server.service');
             assert_script_run("rm -f -r $args{exppath}/nfsok; touch $args{exppath}/nfsok");
+            script_run("umount $args{mntpath} || umount -f -l $args{mntpath}");
+            assert_script_run("mount -t nfs localhost:$args{exppath} $args{mntpath}");
             save_screenshot;
         }
         elsif ($args{role} eq 'client') {


lib/parallel_guest_migration_base.pm
```diff
@@ -570,12 +570,18 @@ sub config_host_shared_storage {
             }
             $_temppath = $args{exppath};
             $_temppath =~ s/\//\\\//g;
+            # Create export path directory if it doesn't exist
+            unless (-d $_temppath) {
+                assert_script_run("mkdir -p $_temppath");
+            }
+            assert_script_run("cp -r $args{mntpath}/* $args{exppath}/");
```
alice-suse, line 577: Maybe good to add a comment to explain why this is needed (sth like guest installation puts images to xx by default, move them to nfs export directory before migration, ... ).

BTW, will mv be better than cp? To avoid confusion what image is used when there is problem in nfs.

lib/parallel_guest_migration_base.pm
```diff
@@ -570,12 +570,21 @@ sub config_host_shared_storage {
             }
             $_temppath = $args{exppath};
             $_temppath =~ s/\//\\\//g;
+            # Create export path directory if it doesn't exist
+            unless (-d $_temppath) {
```
waynechen55, line 574: I am afraid this
```
unless (-d $_temppath)
```
judges whether ```$_temppath``` exist on worker machine instead of SUT machine. Maybe you can verify your test again with such path on worker machine but not on SUT machine.

lib/parallel_guest_migration_base.pm
```diff
@@ -570,12 +570,21 @@ sub config_host_shared_storage {
...
+            # move them to nfs export directory /home/virt, then mount /home/virt to /var/lib/libvirt/images.
+            assert_script_run("mv $args{mntpath}/* $args{exppath}/");
             assert_script_run("sed -i \'/^.*$_temppath.*\$/d\' /etc/exports");
             assert_script_run("echo \"$args{exppath} *(rw,sync,no_root_squash,no_subtree_check)\" >> /etc/exports");
             assert_script_run("exportfs -a");
             systemctl('restart nfs-server.service');
             systemctl('status nfs-server.service');
             assert_script_run("rm -f -r $args{exppath}/nfsok; touch $args{exppath}/nfsok");
+            assert_script_run("mount -t nfs localhost:$args{exppath} $args{mntpath}");
             save_screenshot;
```
waynechen55, line 588: In this case, the original image path is mntpath, so there is no wrong to copy images over to exppath from there. But in light of the logic of this change and in order to ensure consistency and general-purpose, I think it would be better to introduce another argument, ```$args{imgpath}```, so
```
 $args{imgpath} //= '/var/lib/libvirt/images';
......
assert_script_run("mv $args{imgpath}/* $args{exppath}/");
```

lib/parallel_guest_migration_base.pm
```diff
@@ -570,12 +570,21 @@ sub config_host_shared_storage {
...
+            # Guest installation puts images to /var/lib/libvirt/images by default,
+            # move them to nfs export directory /home/virt, then mount /home/virt to /var/lib/libvirt/images.
+            assert_script_run("mv $args{mntpath}/* $args{exppath}/");
             assert_script_run("sed -i \'/^.*$_temppath.*\$/d\' /etc/exports");
             assert_script_run("echo \"$args{exppath} *(rw,sync,no_root_squash,no_subtree_check)\" >> /etc/exports");
             assert_script_run("exportfs -a");
             systemctl('restart nfs-server.service');
             systemctl('status nfs-server.service');
             assert_script_run("rm -f -r $args{exppath}/nfsok; touch $args{exppath}/nfsok");
+            assert_script_run("mount -t nfs localhost:$args{exppath} $args{mntpath}");
```
waynechen55, line 587: Umount mntpath in the firstplace I think:
```
script_run("umount $args{mntpath} || umount -f -l $args{mntpath}");
```

lib/parallel_guest_migration_base.pm
RoyCai7: I am a little confused about it. NFS server just setup on server, not on client. So judges it on server. It is enough.

lib/parallel_guest_migration_base.pm
waynechen55: Maybe more general name for example $args{srcpath} is better.

RoyCai7: Sound good.

lib/parallel_guest_migration_base.pm
RoyCai7: I made test, no such issue.

alice-suse: `unless (-d $_temppath)`  is not executed on the test machine, no matter server or client. It runs on where the worker services runs, for your case, void.qam.suse.cz. You should use `script_run("xxx")`. This will be executed on the test machine.

RoyCai7: if (script_run("test -d $_temppath") != 0) {
}
It also works. will execute on the worker.

alice-suse: Explained to Roy offline. Should be clear now.

waynechen55:
Please also help update description (between ==head2 and ==cut) of the subroutine, you can use description as below if you think it is proper:
```
Configure shared nfs storage on server and mount it on client. Main arguments
are type of shared storage, path of exported shared storage, server or client
, original source path on server and mount path on server and client.
```

alice-suse:
It seems all comments are resolved. Consider to merge if no other comments.

PR #17557, opened 2023-08-09
Install kirk from repositories

acerv:
Now it's possible to define KIRK_FROM variable that can select kirk install from LTP repository or git.
This modification add support for transactional-update command.

- [osd verification run](https://openqa.suse.de/tests/11768749#step/test_memcontrol/1)

### DIFF
--- a/lib/hacluster.pm
+++ b/lib/hacluster.pm
@@ -1219,7 +1219,7 @@ sub cluster_status_matches_regex {
     my $previous_line = '';
 
     for my $line (split("\n", $show_cluster_status)) {
-	if (lc($line) =~ /\s?(stopped|failed|pending|blocked|starting|promoting):?/) {
+        if ($line =~ /\s?(stopped|failed|pending|blocked|starting|promoting):?/i) {
             push @resource_list, $previous_line;
             push @resource_list, $line;
         }


tests/kernel/kirk.pm
```diff
@@ -30,11 +36,31 @@ sub run
...
+    if ($install_from =~ /git/i) {
+        if (is_transactional) {
+            trup_install("git");
+        } else {
+            zypper_call("in -y git");
+        }
+
+        assert_script_run("git clone -q --single-branch -b $branch --depth 1 $repo");
+
+        $cmd .= "python3 kirk/kirk ";
```
czerw, line 51: why `.=` and not just `=`?

tests/kernel/kirk.pm
```diff
@@ -30,11 +36,31 @@ sub run
...
+    } else {
+        add_ltp_repo;
```
czerw, line 53: so we assume that kirk is in same repos as ltp, fair enough, but it should be mentioned is some doc section

tests/kernel/kirk.pm
```diff
@@ -30,11 +36,31 @@ sub run
...
     select_serial_terminal;
+    zypper_call("ref");
+
+    my $cmd = '';
+    if ($install_from =~ /git/i) {
+        if (is_transactional) {
+            trup_install("git");
+        } else {
+            zypper_call("in -y git");
+        }
```
czerw, line 47: I would do one liner `is_transactional ? trup_install("git") : zypper_call("in -y git");`, but is just different style.

tests/kernel/kirk.pm
acerv: You mean doc section inside the code or somewhere else?

tests/kernel/kirk.pm
```diff
@@ -6,19 +6,28 @@
...
 use utils;
 use strict;
 use testapi;
 use warnings;
 use serial_terminal 'select_serial_terminal';
+use version_utils qw(is_transactional);
+use transactional 'trup_install';
+
+use FindBin 1.51 qw( $RealBin );
+use lib $RealBin;
```
czerw, line 22: why is needed findbin?

lib/LTP/utils.pm
```diff
@@ -352,4 +353,67 @@ sub parse_runfiles {
...
+
+    my $run_cmd = is_transactional ? 'transactional-update -c -d --quiet run' : '';
+    for my $pkg (@pkgs) {
+        my $want_32bit = want_ltp_32bit($pkg);
+
+        record_info("LTP pkg: $pkg", script_output("$run_cmd rpm -qi $pkg | tee "
+                  . get_ltp_version_file($want_32bit)));
+        assert_script_run "find " . get_ltproot($want_32bit) .
+          q(/testcases/bin/openposix/conformance/interfaces/ -name '*.run-test' > )
+          . get_ltp_openposix_test_list_file($want_32bit);
```
czerw, line 415: `get_ltp_openposix_test_list_file` will have to be moved as well

acerv: `get_ltp_openposix_test_list_file` is already in LTP/utils.pm

PR #17563, opened 2023-08-10
Integrate Slurm versioned installation

b10n1k:
Make possible to install and test latest version of the Slurm package. As for now, Slurm is explicitly invoked. But now test should be able to get the latest available versioned Slurm in the system, install it and run the tests.

Those changes are not apply to all the test cases at the moment. Verifications run only against a basic scenario (see hpc_EPSILON) and the test requires a new job variable, SLURM_VERSIONED_TEST.

- Related ticket: https://progress.opensuse.org/issues/130838

- Verification run: 
https://aquarius.suse.cz/tests/18805# current version
https://aquarius.suse.cz/tests/18763# versioned slurm

### DIFF
--- a/lib/hpc/utils.pm
+++ b/lib/hpc/utils.pm
@@ -14,6 +14,9 @@ use utils;
 use version_utils 'is_sle';
 use serial_terminal 'select_serial_terminal';
 use version;
+use Exporter 'import';
+
+our @EXPORT = qw(get_slurm_version);
 
 sub get_mpi() {
     my $mpi = get_required_var('MPI');
@@ -27,6 +30,20 @@ sub get_mpi() {
     }
 
     return $mpi;
+}
+
+=head2 get_slurm_version
+
+ get_slurm_version();
+
+Returns the slurm package to be installed. C<SLURM_VERSION> takes the form of I<Major_Minor>.
+for instance: 23_02.
+
+=cut
+
+sub get_slurm_version {
+    my $pkg_version = shift;
+    return $pkg_version ? 'slurm_' . $pkg_version : 'slurm';
 }
 
 =head2 get_mpi_src
--- a/tests/hpc/slurm_db.pm
+++ b/tests/hpc/slurm_db.pm
@@ -13,20 +13,24 @@ use testapi;
 use serial_terminal 'select_serial_terminal';
 use lockapi;
 use utils;
+use hpc::utils 'get_slurm_version';
 use version_utils 'is_sle';
 
 sub run ($self) {
     select_serial_terminal();
     my $hostname = get_required_var("HOSTNAME");
+    my $slurm_pkg = get_slurm_version(get_var('SLURM_VERSION', ''));
 
     barrier_wait('CLUSTER_PROVISIONED');
 
     $self->prepare_user_and_group();
 
     # Install slurm
-    zypper_call("in slurm slurm-munge slurm-slurmdbd");
+    # $slurm_pkg-munge is installed explicitly since slurm_23_02
+    zypper_call("in $slurm_pkg $slurm_pkg-munge $slurm_pkg-slurmdbd");
     # install slurm-node if sle15, not available yet for sle12
-    zypper_call('in slurm-node') if is_sle '15+';
+    # $slurm_pkg-munge is installed explicitly since slurm_23_02
+    zypper_call("in $slurm_pkg-node") if is_sle '15+';
 
     my $mariadb_service = "mariadb";
     $mariadb_service = "mysql" if is_sle('<12-sp4');
--- a/tests/hpc/slurm_master.pm
+++ b/tests/hpc/slurm_master.pm
@@ -6,7 +6,7 @@
 # Summary: Slurm master node
 #    This test is setting up slurm master node and runs tests depending
 #    on the slurm cluster configuration.
-#    SLURM_VERSIONED_TEST enables installation of the versioned Slurm.
+#    SLURM_VERSION enables installation of the particular versioned Slurm.
 # Maintainer: Kernel QE <kernel-qa@suse.de>
 
 use Mojo::Base qw(hpcbase hpc::configs), -signatures;
@@ -14,10 +14,13 @@ use testapi;
 use serial_terminal qw(select_serial_terminal select_user_serial_terminal);
 use lockapi;
 use utils;
+use hpc::utils 'get_slurm_version';
 use version_utils 'is_sle';
 use Utils::Logging 'export_logs_basic';
 
 our @all_tests_results;
+
+our $slurm_pkg = get_slurm_version(get_var('SLURM_VERSION', ''));
 
 sub run_tests ($slurm_conf) {
     my $xmlfile = 'testresults.xml';
@@ -190,8 +193,8 @@ sub t08_basic() {
     my $name = 'pdsh-slurm';
     my $description = 'Basic check of pdsh-slurm over ssh';
     my $result = 0;
-
-    zypper_call('in pdsh pdsh-slurm');
+    # $slurm_pkg-munge is installed explicitly since slurm_23_02
+    zypper_call("in pdsh pdsh-$slurm_pkg");
 
     my $sinfo_nodeaddr = script_output('sinfo -a --Format=nodeaddr -h');
     my $pdsh_nodes = script_output('pdsh -R ssh -P normal /usr/bin/hostname');
@@ -421,34 +424,11 @@ sub extended_hpc_tests ($master_ip, $slave_ip) {
     parse_extra_log('XUnit', './results/TEST-hpc-test.xml');
 }
 
-our $slurm_pkg = 'slurm';
-
 sub run ($self) {
     select_serial_terminal();
     my $nodes = get_required_var('CLUSTER_NODES');
     my $slurm_conf = get_required_var('SLURM_CONF');
     my $version = get_required_var('VERSION');
-
-    if (get_var('SLURM_VERSIONED_TEST', '0') == '1') {
-        if (is_sle('<15-SP4')) {
-            record_info 'No testing', 'versioned Slurm not used';
-            return;
-        }
-        my $query_output = script_output(qq{zypper se slurm | awk '{print \$2}' | grep -E '^slurm_[0-9]{1,2}_[0-9]{1,2}\$' | uniq});
-        my @slurm_vers = split(/(\ |\n)/, $query_output);
-        my $higher_slurm_ver = 0;
-        record_info "@slurm_vers", 'slurm packages found';
-        foreach my $v (@slurm_vers) {
-            my ($slurm_version_tmp) = $v =~ /([0-9]{1,2}_[0-9]{1,2})/;
-            $slurm_version_tmp =~ s/_/\./;
-            if ($higher_slurm_ver < $slurm_version_tmp) {
-                $higher_slurm_ver = $slurm_version_tmp;
-            }
-        }
-        $higher_slurm_ver =~ s/\./_/;    # restore format
-        record_info "Latest $higher_slurm_ver", "This should be the latest/higher slurm version";
-        $slurm_pkg = 'slurm_' . $higher_slurm_ver;
-    }
 
     barrier_wait('CLUSTER_PROVISIONED');
     $self->prepare_user_and_group();
@@ -457,13 +437,9 @@ sub run ($self) {
     # provision HPC cluster, so the proper rpms are installed,
     # munge key is distributed to all nodes, so is slurm.conf
     # and proper services are enabled and started
-    if ($slurm_pkg =~ /slurm_/) {
-        # $slurm_pkg-munge is installed explicitly since slurm_23_02
-        # bsc1214094
-        zypper_call("in $slurm_pkg $slurm_pkg-munge");
-    } else {
-        zypper_call("in slurm slurm-munge slurm-torque");
-    }
+    # $slurm_pkg-munge is installed explicitly since slurm_23_02
+    zypper_call("in $slurm_pkg $slurm_pkg-munge $slurm_pkg-torque");
+    record_info script_output("rpm -q --queryformat='%{VERSION}' $slurm_pkg"), 'slurm version';
 
     if ($slurm_conf =~ /ha/) {
         $self->mount_nfs();
--- a/tests/hpc/slurm_master.pm
+++ b/tests/hpc/slurm_master.pm
@@ -5,7 +5,8 @@
 
 # Summary: Slurm master node
 #    This test is setting up slurm master node and runs tests depending
-#    on the slurm cluster configuration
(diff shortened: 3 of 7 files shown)


tests/hpc/slurm_master.pm
```diff
@@ -420,20 +421,49 @@ sub extended_hpc_tests ($master_ip, $slave_ip) {
...
     $self->generate_and_distribute_ssh();
 
     # provision HPC cluster, so the proper rpms are installed,
     # munge key is distributed to all nodes, so is slurm.conf
     # and proper services are enabled and started
-    zypper_call('in slurm slurm-munge slurm-torque');
+    if ($slurm_pkg =~ /slurm_/) {
+        # $slurm_pkg-munge is installed explicitly since slurm_23_02
+        # bsc1214094
+        zypper_call("in $slurm_pkg $slurm_pkg-munge");
```
czerw, line 463: what about `$slurm_pkg-torque`? i see it bellow in non-versioned

tests/hpc/slurm_master.pm
```diff
@@ -420,20 +421,49 @@ sub extended_hpc_tests ($master_ip, $slave_ip) {
...
+        record_info "@slurm_vers", 'slurm packages found';
+        foreach my $v (@slurm_vers) {
+            my ($slurm_version_tmp) = $v =~ /([0-9]{1,2}_[0-9]{1,2})/;
+            $slurm_version_tmp =~ s/_/\./;
+            if ($higher_slurm_ver < $slurm_version_tmp) {
+                $higher_slurm_ver = $slurm_version_tmp;
+            }
+        }
+        $higher_slurm_ver =~ s/\./_/;    # restore format
+        record_info "Latest $higher_slurm_ver", "This should be the latest/higher slurm version";
```
czerw, line 449: Does it mean that there can be multiple versioned slurms? If yes, will all be supported ?  If so, we should test each version in defined job and not search use latest one. I would use `SLURM_VERSION` variable where we can define which version to test.

tests/hpc/slurm_master.pm
b10n1k: i dont know about this package. Maybe @schlad has an idea as he was the author in first place. I cant really remember but i think `$slurm_pkg-torque` is not available.

tests/hpc/slurm_master.pm
b10n1k: This would complicate things. There was two versions. the lower one is the one which is installed by the non-version as far as i tell

tests/hpc/slurm_master.pm
czerw: Ok, lets wait for @schlad

schlad: just to summarize; we talked and agreed this thing is very much a product related thing, so @b10n1k is the one who can really check/answer this.
This is a simple matter of slurm-torque and answering if this is an rpm which is versioned/non-versioned and checking if it does install nicely in case slurm is in versioned flavor?

I proposed we could verify that manually together but we would need a cluster with provisioned slurm scheduled. 
Surely we could just search with zypper and see how that package looks like in case the slurm is versioned.

schlad (changes requested):
Please, as we talked, check if slurm-torque has the versioned flavor or only non-versioned? 
I would pay attention to this and, before introducing an if statement, I would check how slurm-torque behaves in case we have slurm versioned flavor installed in the system.

czerw:
> Please, as we talked, check if slurm-torque has the versioned flavor or only non-versioned? I would pay attention to this and, before introducing an if statement, I would check how slurm-torque behaves in case we have slurm versioned flavor installed in the system.

slurm is definitly versioned as i can see  `slurm_23_02-torque-23.02.2-3.3.5.x86_64.rpm` in maint update repository.

tests/hpc/slurm_db.pm
```diff
@@ -15,6 +15,15 @@ use lockapi;
...
 
+our $slurm_pkg;
+BEGIN {
+    if (get_var('SLURM_VERSION') eq 'slurm') {
+        $slurm_pkg = 'slurm';
+    } else {
+        $slurm_pkg = 'slurm_' . get_var('SLURM_VERSION');
+    }
+}
+
```
czerw, line 26: You duplicate same lines of code across different files. It should be function in utils.

Alos can be done in more simple way:
```
my $slurm_version;
my $slurm_pkg  = $slurm_version ? 'slurm_' . $slurm_version : 'slurm';
```

czerw (changes requested):

tests/hpc/slurm_master.pm
```diff
@@ -425,15 +437,22 @@ sub run ($self) {
     my $nodes = get_required_var('CLUSTER_NODES');
     my $slurm_conf = get_required_var('SLURM_CONF');
     my $version = get_required_var('VERSION');
-
+    if ($slurm_pkg =~ /slurm/) {
+        my $query_output = script_output(qq{zypper if slurm | grep -i version | awk '{print \$3}'});
```
czerw, line 441: I will recommend as usual what i recommend when I see awk and grep in pipe : https://blog.jpalardy.com/posts/skip-grep-use-awk/

tests/hpc/slurm_master.pm
```diff
@@ -425,15 +437,22 @@ sub run ($self) {
     my $nodes = get_required_var('CLUSTER_NODES');
     my $slurm_conf = get_required_var('SLURM_CONF');
     my $version = get_required_var('VERSION');
-
+    if ($slurm_pkg =~ /slurm/) {
```
czerw, line 440: This complicate structure is not needed at all,  just do:
```
rpm -q --queryformat='%{VERSION}' $slurm_pkg
```

variables.md
```diff
@@ -174,6 +174,7 @@ SES5_CEPH_QA_HEALTH_OK | string | | URL for repo containing ceph-qa-health-ok pa
 SKIP_CERT_VALIDATION | boolean | false | Enables linuxrc parameter to skip certificate validation of the remote source, e.g. when using self-signed https url.
 SET_CUSTOM_PROMPT | boolean | false | Set a custom, shorter prompt in shells. Saves screen space but can take time to set repeatedly in all shell sessions.
 SLE_PRODUCT | string | | Defines SLE product. Possible values: `sles`, `sled`, `sles4sap`. Is mainly used for SLE 15 installation flow.
+SLURM_VERSION | string | | Takes `slurm` to install current version, or expected version to be tested, ex 23_02. Use https://confluence.suse.com/pages/viewpage.action?pageId=1263338001 to identify version per SLE servive pack.
```
czerw, line 177: Remove confluence reference from doc, its not reachable outside SUSE.
Also I wouls just use SLURM_VERSION with version and default value, if not specified should be slurm. It wil make more simple.

tests/hpc/slurm_master.pm
```diff
@@ -189,8 +200,9 @@ sub t08_basic() {
     my $name = 'pdsh-slurm';
     my $description = 'Basic check of pdsh-slurm over ssh';
     my $result = 0;
-
-    zypper_call('in pdsh pdsh-slurm');
+    # $slurm_pkg-munge is installed explicitly since slurm_23_02
```
czerw, line 203: I think this comment is not need. It's a feature not bug. If you mention bug reference, one would think thare is something pending, but there is nothing pending in this case.

b10n1k: I thought to add it, to give awareness of the change. This change the behavior of the test if neglected, and to avoid reviewers checking again what is the difference. Another option is to add the line on the description. wdyt?

tests/hpc/slurm_db.pm
b10n1k: This block will run during early on in the compilation. As such i avoid to invoke another subroutine in runtime. i though that it will make it a bit faster and clever. 
Now i am not sure what to do with your suggestion

tests/hpc/slurm_master.pm
schlad: personally I would opt for having this comment; maybe amending line 204 to: "see this bug for more technical details: bsc1214094" could be a solution to @czerw right reservations  (If you mention bug reference, one would think thare is something pending)?

tests/hpc/slurm_master.pm
b10n1k: Done, ready for review

tests/hpc/slurm_master.pm
b10n1k: Done, ready for review

variables.md
b10n1k: i removed second sentence. review and close if all good

tests/hpc/slurm_master.pm
b10n1k: i think this can be resolved

tests/hpc/slurm_master.pm
b10n1k: this also can resolved. With the second commit this block has been removed.

tests/hpc/slurm_db.pm
b10n1k: i just notice that the CI breaks because of this, and i see why pcervinka asked for a lib function. need to look on that further

variables.md
```diff
@@ -174,6 +174,7 @@ SES5_CEPH_QA_HEALTH_OK | string | | URL for repo containing ceph-qa-health-ok pa
 SKIP_CERT_VALIDATION | boolean | false | Enables linuxrc parameter to skip certificate validation of the remote source, e.g. when using self-signed https url.
 SET_CUSTOM_PROMPT | boolean | false | Set a custom, shorter prompt in shells. Saves screen space but can take time to set repeatedly in all shell sessions.
 SLE_PRODUCT | string | | Defines SLE product. Possible values: `sles`, `sled`, `sles4sap`. Is mainly used for SLE 15 installation flow.
+SLURM_VERSION | string | | if given it installed versioned Slurm, otherwise it uses the base version. ex 23_02.
```
czerw, line 177: Current sentence doesn't make much sense. I would say `Defines slurm version (23_02) for installation. `

tests/hpc/slurm_db.pm
czerw: Just put it into hpc utils. There is no reason (bit faster and clever argument looks strange in this context) to duplicate line across files.

tests/hpc/slurm_master.pm
czerw: yes, lets keep first comment, just remove bug reference

tests/hpc/slurm_db.pm
b10n1k: Please ask me. because i feel we are talking about different things. 
In any case, changes have been made and it is ready for review

tests/hpc/slurm_master.pm
b10n1k: Done and ready for review

variables.md
b10n1k: I add this and modified the rest. i hope it makes more sense

variables.md
```diff
@@ -174,6 +174,7 @@ SES5_CEPH_QA_HEALTH_OK | string | | URL for repo containing ceph-qa-health-ok pa
 SKIP_CERT_VALIDATION | boolean | false | Enables linuxrc parameter to skip certificate validation of the remote source, e.g. when using self-signed https url.
 SET_CUSTOM_PROMPT | boolean | false | Set a custom, shorter prompt in shells. Saves screen space but can take time to set repeatedly in all shell sessions.
 SLE_PRODUCT | string | | Defines SLE product. Possible values: `sles`, `sled`, `sles4sap`. Is mainly used for SLE 15 installation flow.
+SLURM_VERSION | string | | Defines slurm version (ex: 23_02) for installation. if not set, installation uses the baseSlurm version.
```
czerw, line 177: `if not set, installation uses the baseSlurm version.` 

Please fix capitalization in `if`, it is starting new sentence.   Also `baseSlurm` needs to be fixed.

lib/hpcbase.pm
```diff
@@ -16,6 +16,14 @@ use version_utils 'is_sle';
...
 use Utils::Logging 'save_and_upload_log';
 
+our $slurm_pkg;
+
+BEGIN {
+    if (get_var('HPC', '') =~ /slurm/) {
+        $slurm_pkg = get_var('SLURM_VERSION', '') ? 'slurm_' . get_var('SLURM_VERSION') : 'slurm';
+    }
+}
+
```
czerw, line 26: Please do standard function definition (like we do everywhere else) which will have one argument for version and will return package name and not this approach.

lib/hpc/utils.pm
```diff
@@ -29,6 +32,19 @@ sub get_mpi() {
     return $mpi;
 }
 
+=head2 get_slurm_version
+
+ get_slurm_version();
+
+Returns the slurm package to be installed. C<SLURM_VERSION> takes the form of I<Major.Minor>.
+for instance: 23.02.
```
czerw, line 40: It is `23_02`, why did you suddenly change it to `23.02`?

lib/hpc/utils.pm
```diff
@@ -29,6 +32,19 @@ sub get_mpi() {
...
+
+=cut
+
+sub get_slurm_version {
```
czerw, line 44: Please use argument with slurm version.

tests/hpc/slurm_master.pm
```diff
@@ -5,19 +5,23 @@
...
 use serial_terminal qw(select_serial_terminal select_user_serial_terminal);
 use lockapi;
 use utils;
+use hpc::utils 'get_slurm_version';
 use version_utils 'is_sle';
 use Utils::Logging 'export_logs_basic';
 
 our @all_tests_results;
 
+our $slurm_pkg = get_slurm_version();
```
czerw, line 23: There is no need to have it as global variable. Use it in the code like we usually do.

czerw (changes requested):

tests/hpc/slurm_master.pm
b10n1k: it is not needed for most of the modules. but it does for `slurm_master`.  The variable is used on `t08_basic`. i think it makes sense to not touch anything here and just use the variable as public for this case.

czerw: It will be better to update `t08_basic` code with one line. I will be at one place much clear where is variable coming from.

lib/hpc/utils.pm
```diff
@@ -29,6 +32,20 @@ sub get_mpi() {
     return $mpi;
 }
 
+=head2 get_slurm_version
+
+ get_slurm_version();
+
+Returns the slurm package to be installed. C<SLURM_VERSION> takes the form of I<Major.Minor>.
```
czerw, line 39: <Major.Minor> should be <Major_Minor>.

czerw (approved):

schlad (approved):

PR #17569, opened 2023-08-11
Revert 2ee46ad8cf to solve aarch boot fail

SeroSun:
aarch64 xfstests all fail to boot in alp when using qcow2 as HDD_1. I checked the log it skip the bootloader part in the previous commit 2ee46ad8cf in tests/microos/disk_boot.pm.
Pass in x86_64: https://openqa.suse.de/tests/11732109#step/disk_boot/1
Fails in aarch64: https://openqa.suse.de/tests/11782227#step/disk_boot/4
Shall we consider reverting this behavior in aarch64 alp? I can't find which test fails before made this change.

- Related ticket: https://progress.opensuse.org/issues/134099
- Verification run: WIP

### DIFF
--- a/tests/microos/disk_boot.pm
+++ b/tests/microos/disk_boot.pm
@@ -10,7 +10,7 @@ use base "consoletest";
 use strict;
 use warnings;
 use testapi;
-use version_utils qw(is_sle_micro is_leap_micro is_alp);
+use version_utils qw(is_sle_micro is_leap_micro);
 use Utils::Architectures qw(is_aarch64);
 use microos "microos_login";
 use transactional "record_kernel_audit_messages";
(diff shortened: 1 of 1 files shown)


SeroSun:
At this moment the verify run does not run OK with CASEDIR. https://openqa.suse.de/tests/11801064 (main: Unable to clone Git repository 'https://github.com/serosun/os-autoinst-distri-opensuse/tree/yosun/xfstests_aarch' specified via CASEDIR)
I'm not sure if this could fix my aarch64 tests problem or not. I'll try to fix the verify run.

ricardobranco777:
> At this moment the verify run does not run OK with CASEDIR. https://openqa.suse.de/tests/11801064 (main: Unable to clone Git repository 'https://github.com/serosun/os-autoinst-distri-opensuse/tree/yosun/xfstests_aarch' specified via CASEDIR) I'm not sure if this could fix my aarch64 tests problem or not. I'll try to fix the verify run.

Set CASEDIR to `https://github.com/serosun/os-autoinst-distri-opensuse.git#xfstests_aarch`

mloviska:
This will break `aarch64` tests, as openQA won't be able to catch grub2 screen. Hence, the first job in your chain would be failing. In fact, in the first job you turn off grub2 timeout, maybe you can reconsider to do this, or just use a different grub2 handling module in the second job

SeroSun:
> This will break `aarch64` tests, as openQA won't be able to catch grub2 screen. Hence, the first job in your chain would be failing. In fact, in the first job you turn off grub2 timeout, maybe you can reconsider to do this, or just use a different grub2 handling module in the second job

Thank you for your input. And I checked osd, the new build didn't fail in disk_boot step. Then let me close this PR.

PR #17562, opened 2023-08-10
ntp_client.pm: reinstall chrony if pool.conf is empty

paolostivanin:
This is a workaround for bsc#1214141.

- Related ticket: https://progress.opensuse.org/issues/133763
- Verification runs:
  - https://openqa.suse.de/tests/11792552
  - https://openqa.suse.de/tests/11792553

PR #17574, opened 2023-08-12
ntp_client.pm: only apply workaround on SLE

paolostivanin:
follow-up fixes after #17562


paolostivanin:
@DimStar77 this should fix the ntp issue :smile:

PR #17575, opened 2023-08-14
Create new libyui-rest-api based test modules for ntp_config_settings

tinawang123:
Create new libyui-rest-api based test modules for ntp_config_settings

- Related ticket: https://progress.opensuse.org/issues/133001
- Verification run: https://openqa.suse.de/tests/11857438
   s390x: https://openqa.suse.de/tests/11820003#step/ntp_configuration/1
   aarch64: https://openqa.suse.de/tests/11821383#step/ntp_configuration/1

### DIFF
--- a/lib/Distribution/Sle/15_current.pm
+++ b/lib/Distribution/Sle/15_current.pm
@@ -30,6 +30,7 @@ use Installation::SystemRole::Sle::SystemRoleController;
 use Installation::ModuleRegistration::SeparateRegCodesController;
 use YaST::DNSServer::Sle::DNSServerController;
 use YaST::DNSServer::Sle::DNSServerSetupController;
+use Installation::NTPConfiguration::NTPConfigurationPage;
 
 =head2 get_license_agreement
 
@@ -96,6 +97,10 @@ sub get_dns_server {
 
 sub get_dns_server_setup {
     return YaST::DNSServer::Sle::DNSServerSetupController->new();
+}
+
+sub get_ntp_configuration {
+    return Installation::NTPConfiguration::NTPConfigurationPage->new();
 }
 
 1;
--- a/lib/Installation/NTPConfiguration/NTPConfigurationPage.pm
+++ b/lib/Installation/NTPConfiguration/NTPConfigurationPage.pm
@@ -1,0 +1,44 @@
+# SUSE's openQA tests
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act on the
+#          ntp Configuration page.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Installation::NTPConfiguration::NTPConfigurationPage;
+use parent 'Yam::PageBase';
+use strict;
+use warnings;
+
+sub init {
+    my $self = shift;
+    $self->{txb_ntp_server} = $self->{app}->textbox({id => '"Y2Caasp::Widgets::NtpServer"'});
+    $self->{btn_next} = $self->{app}->button({id => 'next'});
+    return $self;
+}
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{txb_ntp_server}->exist({timeout => 200});
+}
+
+sub get_ntp_servers {
+    my ($self) = @_;
+    $self->get_ntp_configuration_page();
+    return $self->{txb_ntp_server}->value();
+}
+
+sub get_ntp_configuration_page {
+    my ($self) = @_;
+    die "Ntp configuration page is not displayed" unless $self->is_shown();
+}
+
+sub press_next {
+    my ($self) = @_;
+    $self->get_ntp_configuration_page();
+    return $self->{btn_next}->click();
+}
+
+1;
--- a/schedule/yast/sle/flows/default_slem.yaml
+++ b/schedule/yast/sle/flows/default_slem.yaml
@@ -12,7 +12,7 @@ registration:
   - installation/registration/register_via_scc
 addons: []
 ntp_configuration:
-  - installation/ntp_config_settings
+  - yam/ntp_configuration
 authentication:
   - installation/authentication/root_simple_pwd
 software:
--- a/tests/yam/ntp_configuration.pm
+++ b/tests/yam/ntp_configuration.pm
@@ -1,0 +1,22 @@
+# SUSE's openQA tests
+#
+# Copyright 2021 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: This test module launches the installation from
+#          the installation settings page.
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use Test::Assert 'assert_matches';
+
+sub run {
+    my $ntp_settings = $testapi::distri->get_ntp_configuration()->get_ntp_servers();
+    assert_matches(qr/suse\.pool\.ntp\.org|10\.160\.0\.45\ 10\.160\.0\.44\ 10\.160\.255\.254/, $ntp_settings, "NTP setting is not correct");
+    $testapi::distri->get_ntp_configuration()->press_next();
+}
+
+1;


lib/Installation/NtpConfiguration/NtpConfigurationPage.pm
```diff
@@ -0,0 +1,42 @@
...
+    my ($self) = @_;
+    return $self->{txb_ntp_server}->exist();
+}
+
+sub get_ntp_config {
+    my ($self) = @_;
+    return $self->{txb_ntp_server}->value();
+}
+
+sub get_ntp_configuration_page {
```
sofiasyria, line 32: I don't see this function being used anywhere.

tests/installation/ntp_configuration.pm
```diff
@@ -0,0 +1,22 @@
...
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use Test::Assert ':all';
+
+sub run {
+    my $ntp_settings = $testapi::distri->get_ntp_config()->get_ntp_config();
```
sofiasyria, line 17: I would recommend to give different names to these functions. Like get_ntp_value().

tests/installation/ntp_configuration.pm
```diff
@@ -0,0 +1,22 @@
...
+
+# Summary: This test module launches the installation from
+#          the installation settings page.
+
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use Test::Assert ':all';
```
sofiasyria, line 14: you just need use Test::Assert 'assert_matches';

sofiasyria (approved):
Some minor edits recommended. Other than that LGTM.

lib/Installation/NtpConfiguration/NtpConfigurationPage.pm
tinawang123: Need it to check if we get the expected page.  Added to use this function at 'tests/installation/ntp_configuration.pm'

lib/Distribution/Sle/15_current.pm
```diff
@@ -30,6 +30,8 @@ use Installation::SystemRole::Sle::SystemRoleController;
 use Installation::ModuleRegistration::SeparateRegCodesController;
 use YaST::DNSServer::Sle::DNSServerController;
 use YaST::DNSServer::Sle::DNSServerSetupController;
+use Installation::NtpConfiguration::NtpConfigurationPage;
+use YuiRestClient;
```
jknphy, line 34: you don't need this, you can inherit lib/Yam/PageBase.pm in corresponding Page

lib/Distribution/Sle/15_current.pm
```diff
@@ -98,4 +100,8 @@ sub get_dns_server_setup {
     return YaST::DNSServer::Sle::DNSServerSetupController->new();
 }
 
+sub get_ntp_config {
```
jknphy, line 103: s/config/configuration/ (the preference is always whatever developers thought was best for wording, even if it is long), but yeah neat picking here :)

lib/Installation/NtpConfiguration/NtpConfigurationPage.pm
```diff
@@ -0,0 +1,42 @@
...
+#
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: FSFAP
+
+# Summary: The module provides interface to act on the
+#          ntp Configuration page.
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+package Installation::NtpConfiguration::NtpConfigurationPage;
+use parent 'Installation::Navigation::NavigationBase';
```
jknphy, line 11: we are avoiding to use NavigationBase, this was a bad choice in the past, please repeat the next method for now in this class.

lib/Installation/NtpConfiguration/NtpConfigurationPage.pm
jknphy: you need to run it chained in line 39.

schedule/yast/sle/flows/default_slem.yaml
```diff
@@ -12,7 +12,7 @@ registration:
   - installation/registration/register_via_scc
 addons: []
 ntp_configuration:
-  - installation/ntp_config_settings
+  - installation/ntp_configuration
```
jknphy, line 15: new stuff goes to yam folder not installation folder.

lib/Installation/NtpConfiguration/NtpConfigurationPage.pm
tinawang123: Before get the ntp server value, I need make sure the NTP page show up.
So run it chained in line 28 instead of 39.

jknphy: you need to chain the call in both places, if you check existing libyui-rest-api call you will see this strategy all  over the source code.

tests/installation/ntp_configuration.pm
jknphy: please use names in the screens, `get_npt_servers` in the test layer.

lib/Distribution/Sle/15_current.pm
jknphy: `app => YuiRestClient::get_app()}` is not needed if you use that parent class.

tests/yam/yast_modules/ntp_configuration.pm
```diff
@@ -0,0 +1,22 @@
+# SUSE's openQA tests
```
jknphy, line 1: you don't need intermediate folder 'yast_modules', we are trying to avoid big tree of folders under yam/ and it is not a YaST module.

tests/yam/yast_modules/ntp_configuration.pm
```diff
@@ -0,0 +1,22 @@
...
+# Maintainer: QE YaST and Migration (QE Yam) <qe-yam at suse de>
+
+use strict;
+use warnings;
+use base 'y2_installbase';
+use Test::Assert 'assert_matches';
+
+sub run {
+    my $ntp_settings = $testapi::distri->get_ntp_configuration()->get_ntp_value();
+    assert_matches(qr/suse\.pool\.ntp\.org|10\.160\.0\.45\ 10\.160\.0\.44\ 10\.160\.255\.254/, $ntp_settings, "NTP setting is not correct");
```
jknphy, line 18: in general we don't do this, this is a weak validation, if the servers where swap we would not detect any issue, better practice is not validate the screen in 99 % of the test suite and then have a extended test suite that does all the heavy validation via test data (without any condition, 'OR" or anything. But you can keep it like that, just pointing you to the better solution for future.

lib/Installation/NTPConfiguration/NTPConfigurationPage.pm
```diff
@@ -0,0 +1,50 @@
...
+    $self->get_ntp_configuration_page();
+    return $self->{txb_ntp_server}->value();
+}
+
+sub get_ntp_configuration_page {
+    my ($self) = @_;
+    die "Ntp configuration page is not displayed" unless $self->is_shown();
+}
+
+sub accept_ntp_config {
```
jknphy, line 45: "next" would suffice, it is redundant to use the whole name, and "accept" would be ok, but as this never changes, "next()" is very clear. But it is just a recommendation.

lib/Distribution/Sle/15_current.pm
tinawang123: I need this at here.  As we removed controller part. At here, we need get app then to get the widget information.
if remove it, it will get error: https://openqa.suse.de/tests/11848431

jknphy: you shouldn't get that issue if you add `use parent 'Yam::PageBase';` in Page, see for example ControlCenterPage.pm

lib/Installation/NTPConfiguration/NTPConfigurationPage.pm
```diff
@@ -0,0 +1,44 @@
...
+
+sub get_ntp_configuration_page {
+    my ($self) = @_;
+    die "Ntp configuration page is not displayed" unless $self->is_shown();
+}
+
+sub press_next {
+    my ($self) = @_;
+    $self->get_ntp_configuration_page();
+    return $self->{btn_next}->click();
```
jknphy, line 41: the code style that we have been using for a while is not to run this in two instructions but to chain them.

lib/Installation/NTPConfiguration/NTPConfigurationPage.pm
```diff
@@ -0,0 +1,44 @@
...
+
+sub is_shown {
+    my ($self) = @_;
+    return $self->{txb_ntp_server}->exist({timeout => 200});
+}
+
+sub get_ntp_servers {
+    my ($self) = @_;
+    $self->get_ntp_configuration_page();
+    return $self->{txb_ntp_server}->value();
```
jknphy, line 30: same here.

PR #17559, opened 2023-08-09
Change the way to populate os_image_uri

mpagot:
Related ticket: [TEAM-8211](https://jira.suse.com/browse/TEAM-8211)

Note: content of https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17549/ is useful to understand this PR

Verification run

- HanaSR with OS image from CSP catalog (so should not use code I changed) http://openqaworker15.qa.suse.cz/tests/212573
- sle-15-SP5-Azure-SAP-BYOS-x86_64-Build0290-azure_sap_byos_hanasr_sapconf  https://openqa.suse.de/tests/11787078
- sle-15-SP5-Azure-SAP-BYOS-x86_64-Build0290-sles4sap_gnome_saptune_notes https://openqa.suse.de/tests/11787084
- sle-15-SP5-Azure-SAP-BYOS-x86_64-Build0290-azure_sap_byos_qesap_saptune (qesap regression on OSD, should not be influenced by code changes in this PR) https://openqa.suse.de/tests/11787080
-

PR #17576, opened 2023-08-14
Only use get_os_vhd_uri in Azure.

mpagot:
Related ticket: [TEAM-8211](https://jira.suse.com/browse/TEAM-8211)

Note: fix for code in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17559

Verification run:
- sle-15-SP5-Azure-SAP-x86_64-Build0291-qesap_azure_saptune_uri@64bit -> https://openqa.suse.de/tests/11820775


tests/sles4sap/publiccloud/qesap_terraform.pm
```diff
@@ -110,7 +109,7 @@ sub create_hana_vars_section {
 sub run {
     my ($self, $run_args) = @_;
 
-    if (is_azure) {
```
mpagot, line 113: I decided to drop `is_azure` as it not only checks for `PUBLIC_CLOUD_PROVIDER` but also for other PC settings that are not relevant here

emiura:
LGTM

tests/sles4sap/publiccloud/qesap_terraform.pm
a-kpappas: I see now that is_azure is lying about what it checks so I retract my comment

a-kpappas (approved):

lilyeyes (commented):
LGTM.

tests/sles4sap/publiccloud/qesap_terraform.pm
alvarocarvajald: > I decided to drop `is_azure` as it not only checks for `PUBLIC_CLOUD_PROVIDER` but also for other PC settings that are not relevant here

AFAICS the is_* functions are only checking for **PUBLIC_CLOUD** and **PUBLIC_CLOUD_PROVIDER** settings. I think both settings are relevant, and dropping `is_azure` and `is_gce` for `check_var()` only makes our code harder to read.

mpagot: The `if` is there to protect the call of a `$provider` method only implemented in the `provider::azure` child class.

The `provider_factory` that we call few line before in the same test module, is also not using `is_azure` to determine that `provider::azure` has to be instantiated.

alvarocarvajald: I haven't checked so I'm not 100% sure, but it's possible that `publiccloud::basetest::provider_factory()` was implemented before any of the `is_ec2`, `is_azure` or `is_gce` functions were written, which would explain why those is_* functions were not used inside `provider_factory()`. In any case, I don't see how the implementation of a method in one of the base classes should have any bearing on what is used in the test module itself.

As I see it, dropping `is_azure` and `is_gce` from the module only accomplishes a small advantage, in that we no longer need (?) to set the **PUBLIC_CLOUD** setting; OTOH, it makes the code harder to read and encourages us to drop a setting which may be required by other functions, methods and classes.

In any case, changes are already committed and work, so discussion is moot.

PR #17581, opened 2023-08-14
Exclude python3-module from 15-SP6 AY profile

b10n1k:
This is a workaround as the `sle-module-python3` is not available and blocks the installation. This obviously needs to be addressed and be reverted asap we get the expectations/requirements.

- Related ticket: https://progress.opensuse.org/issues/134192
- Verification run: https://aquarius.suse.cz/tests/18171


czerw:
Just please update commit message to be more clear, that it is for HPC. Something like  `hpc: Exclude python3-module from 15-SP6 AY profile`

PR #17582, opened 2023-08-14
Press down on textmode installation

dzedro:
Sometimes timeout is not set to -1, the typing is somehow interrupted
Typed is only '-' and this will set timeout to 0
Pressing down will decrease the value, 0 to -1, -1 is the lowest value

- Related ticket: https://progress.opensuse.org/issues/134183
- Verification run: https://openqa.suse.de/tests/overview?build=grub2_timeout

### DIFF
--- a/tests/installation/disable_grub_timeout.pm
+++ b/tests/installation/disable_grub_timeout.pm
@@ -78,6 +78,11 @@ sub run {
     wait_still_screen(1);
     # ncurses uses blocking modal dialog, so press return is needed
     send_key 'ret' if check_var('VIDEOMODE', 'text');
+    # make sure sure value is -1 not 0
+    if (check_var('VIDEOMODE', 'text') && $timeout eq '-1') {
+        wait_still_screen(1);
+        send_key 'down';
+    }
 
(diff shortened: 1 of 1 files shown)


tests/installation/disable_grub_timeout.pm
```diff
@@ -78,6 +78,11 @@ sub run {
     wait_still_screen(1);
     # ncurses uses blocking modal dialog, so press return is needed
     send_key 'ret' if check_var('VIDEOMODE', 'text');
+    # make sure sure value is -1 not 0
+    if (check_var('VIDEOMODE', 'text') && $timeout eq '-1') {
+        wait_still_screen(1);
+        send_key 'down';
+    }
```
rfan1, line 85: Can we add a needle check here to make sure it has be set to "-1"?

dzedro: I don't want to add needle, just because IMO it's not needed.
It works, but sometimes "1" get somehow lost, as workaround pressing down will make sure the timeout is "-1".
But if by any chance this will happen again I will add needle. :wink:

PR #17583, opened 2023-08-14
iPXE bootloader fixes

mdoucha:
The `ipxe_install` bootloader currently doesn't work for AutoYaST installation jobs. Fix the issue, add iPXE support to `bootloader_start` job module and simplify iPXE boot script handling.

- Related ticket: https://progress.opensuse.org/issues/130219
- Needles: N/A
- Verification runs:
  - https://openqa.suse.de/tests/11998046 (note: `autoyast/installation` module is expected to fail after reboot due to unrelated AutoYaST profile issue)
  - https://openqa.suse.de/tests/11998047

### DIFF
--- a/lib/autoyast.pm
+++ b/lib/autoyast.pm
@@ -713,9 +713,6 @@ sub upload_profile {
     my $profile = $args{profile};
     my $path = $args{path};
 
-    if (check_var('IPXE', '1')) {
-        $path = get_required_var('SUT_IP') . $path;
-    }
     save_tmp_file($path, $profile);
     # Copy profile to ulogs directory, so profile is available in job logs
     make_path('ulogs');
@@ -843,6 +840,11 @@ sub prepare_ay_file {
     $profile = expand_version($profile);
     $profile = adjust_network_conf($profile);
     $profile = expand_variables($profile);
+
+    if (check_var('IPXE', '1')) {
+        $path = get_required_var('SUT_IP') . $path;
+    }
+
     upload_profile(profile => $profile, path => $path);
     return $path;
 }
--- a/lib/ipmi_backend_utils.pm
+++ b/lib/ipmi_backend_utils.pm
@@ -18,9 +18,10 @@ use utils;
 use power_action_utils 'prepare_system_shutdown';
 use Utils::Architectures;
 use Carp;
+use Socket;
 use virt_autotest::utils qw(is_xen_host check_port_state);
 
-our @EXPORT = qw(set_grub_on_vh switch_from_ssh_to_sol_console adjust_for_ipmi_xen set_pxe_efiboot ipmitool enable_sev_in_kernel add_kernel_options set_grub_terminal_and_timeout reconnect_when_ssh_console_broken);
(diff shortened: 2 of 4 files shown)


b10n1k (approved):

lib/autoyast.pm
```diff
@@ -843,6 +840,11 @@ sub prepare_ay_file {
     $profile = expand_version($profile);
     $profile = adjust_network_conf($profile);
     $profile = expand_variables($profile);
+
+    if (check_var('IPXE', '1')) {
+        $path = get_required_var('SUT_IP') . $path;
```
Dawei-Pang, line 845: The code in original place cannot work, may I know why we need connect get_required_var('SUT_IP') and $path? Thanks!

Dawei-Pang:
We have a similar PR: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17400/files

lib/autoyast.pm
mdoucha: Ask @frankenmichl who wrote the original code (before it was broken by refactoring). I assume that there was some concern about clobbering AutoYaST profiles generated for different IPMI machines. This is the correct way to modify the path.

Dawei-Pang: Thanks! I tried to remove the code block and verified with different IPMI machine parallel, the 3 autoyast path are unified for each machine, no conflict occurred, here is an example about running 3 machines parallel: 
http://10.67.134.70:20453/s6_JMKvZYPjJjwKb/files/autoyast_hanaperf/mu.xml
http://10.67.134.70:20463/9WepUQzXV6Hk3nCc/files/autoyast_hanaperf/mu.xml
http://10.67.134.70:20443/qz4FZiEf4FAGWmRP/files/autoyast_hanaperf/mu.xml

frankenmichl: Honestly, it has been a while, so I don't remember the details here. But I am sure it had to do with different machines of the same WORKER_CLASS.

frankenmichl (approved):
Looks good! Thanks

lib/autoyast.pm
Julie-CAO: The autoyast profile failed to be uploaded to the job space on worker for virtualization tests with `$path = get_required_var('SUT_IP') . $path;`. I planed to removed this statement from my PR(https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17400), but since you have more concers about it,  I will remove it for virt tests only.

mdoucha: Autoyast profile download fails because the path is modified in the wrong place and the download then uses the unmodified path. The fix above will work everywhere.

Julie-CAO: Great. I'll revert my change to this path in my PR#17400.

alice-suse:
I was planning to give it a merge since all good and approved, but there is warning about conflicts to be resolved. @mdoucha Can you have a look?

mdoucha:
Rebased, new VRs created. But we'll have to wait a few days for the worker machines to get back online...

frankenmichl (approved):
I like moving some code to ipmi_backend_utils, maybe we can continue with this where reasonable in the future.

PR #17589, opened 2023-08-15
Add workaround for bsc#1213991

mloviska:
RT kernel's scriptlet is failing therefore the whole snapshot is discarded and the rest of the testing is without other updates that are currently in the queue.

- Related ticket: https://progress.opensuse.org/issues/134240
- VRs
  * http://kepler.suse.cz/tests/21523#
  * http://kepler.suse.cz/tests/21522#

### DIFF
--- a/tests/transactional/install_updates.pm
+++ b/tests/transactional/install_updates.pm
@@ -16,6 +16,19 @@ use version_utils 'is_sle_micro';
 use serial_terminal;
 use utils qw(script_retry);
 
+sub soft_fail_rt_scriptlet {
+    return if (get_var('FLAVOR') !~ /rt/i);
+
+    if (script_run("grep '%post(kernel-rt-5.14.21-150400.15.46.1.x86_64) scriptlet failed, exit status 1' /var/log/zypp/history") == 0) {
+        record_soft_failure('bsc#1213991 - %post(kernel-rt-5.14.21-150400.15.40.1.x86_64) scriptlet failed');
+        select_console 'root-console';
+        trup_shell 'zypper -n update', timeout => 1800;
+    } else {
+        die "Transactional update failed with different error cause";
+    }
+}
+
+
 sub run {
     my ($self) = @_;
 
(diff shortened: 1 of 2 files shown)


tests/transactional/install_updates.pm
```diff
@@ -16,6 +16,19 @@ use version_utils 'is_sle_micro';
 use serial_terminal;
 use utils qw(script_retry);
 
+sub soft_fail_rt_scriptlet {
+    return if (get_var('FLAVOR') !~ /rt/i);
+
+    if (script_run("grep '%post(kernel-rt-5.14.21-150400.15.46.1.x86_64) scriptlet failed, exit status 1' /var/log/zypp/history") == 0) {
```
ricardobranco777, line 22: Can we avoid having the kernel version hard-coded?

mloviska: We can, but this time it is really specific to this one patch. Therefore, it appears only in slem 5.4 and 5.3. It is unfortunate that we had the gap where no updates were tested and were just released

grisu48: The softfailure should be very specific to the circumstances, which in this case is the kernel version. This is the signature for the softfailure and must be specific, not generic.

tests/transactional/install_updates.pm
```diff
@@ -34,7 +47,8 @@ sub run {
     }
     add_test_repositories;
     record_info 'Updates', script_output('zypper lu');
-    trup_call 'up', timeout => 1800;
+    my $ret = trup_call 'up', timeout => 1800, proceed_on_failure => 1;
+    soft_fail_rt_scriptlet if $ret;
```
grisu48, line 51:
```suggestion
    soft_fail_rt_scriptlet if ($ret != 0);
```

Shouldn't it be `$ret != 0` because it's a bash return code? And yes, this is confusing.

mloviska: It is the same any value > 0 is truthy

ricardobranco777 (approved):
LGTM

tests/transactional/install_updates.pm
grisu48: Ah yes that's true. I would still be explicit to avoid the same confusion as I had :sweat_smile:

grisu48 (approved):

tests/transactional/install_updates.pm
mloviska: added

PR #17593, opened 2023-08-16
[WIP] ltp: Add option to set SELinux mode for LTP job

czerw:
It is useful for troubleshooting, when we can clone the job and
easily compare results with different settings of SELinux. It is
controlled by LTP_SELINUX variable. Supported values/modes are
disabled, permissive or 0.

We could setup job twice for selected test suites, one with enabled and second with disabled SELinux. Good candidates would be syscalls and controllers.

- Related ticket: none
- Needles: none
- Verification runs:
ALP: https://openqa.suse.de/tests/overview?version=1.0&build=czerw%2Fos-autoinst-distri-opensuse%2317593&distri=alp
SLE Micro:  https://openqa.suse.de/tests/overview?version=5.5&build=czerw%2Fos-autoinst-distri-opensuse%2317593&distri=sle-micro
SLE 12:
SLE 15:
TW:

### DIFF
--- a/tests/sles4sap/publiccloud/qesap_ansible.pm
+++ b/tests/sles4sap/publiccloud/qesap_ansible.pm
@@ -14,6 +14,7 @@ use publiccloud::utils;
 use sles4sap_publiccloud;
 use qesapdeployment;
 use serial_terminal 'select_serial_terminal';
+use version_utils 'is_sle';
 
 sub test_flags {
     return {fatal => 1, publiccloud_multi_module => 1};
@@ -48,6 +49,11 @@ sub run {
         # Check ssh connection for all hosts
         $instance->wait_for_ssh;
 
+        # Update sudo package as a temporary fix for bsc#1205325
+        if (is_sle('=15-sp2')) {
+            $instance->run_ssh_command(cmd => 'sudo zypper up -y sudo');
+            record_soft_failure('bsc#1205325 - update sudo pkg');
+        }
         # Skip instances without HANA db or setup without cluster
         next if ($instance_id !~ m/vmhana/) or !$ha_enabled;
         $self->wait_for_sync();


czerw:
@acerv @Avinesh @richiejp @mdoucha @coolgw  please check

richiejp:
LGTM, but note that putting SELinux into permissive mode is not the same thing as disabling it completely. It still logs denials somewhere which is probably quite useful. However that may still result in some test failures if they rely on particular memory behavior (for e.g.)

acerv:
> LGTM, but note that putting SELinux into permissive mode is not the same thing as disabling it completely. It still logs denials somewhere which is probably quite useful. However that may still result in some test failures if they rely on particular memory behavior (for e.g.)

This is a good point. Why we don't disable SELinux instead?

czerw:
> > LGTM, but note that putting SELinux into permissive mode is not the same thing as disabling it completely. It still logs denials somewhere which is probably quite useful. However that may still result in some test failures if they rely on particular memory behavior (for e.g.)
> 
> This is a good point. Why we don't disable SELinux instead?

You can do  setenforce permissive and continue without restart. If @richiejp  thinks that permissive mode is not enough (still could impact some LTP tests), I can change it.

richiejp:
I think disabling and permissive mode serve different purposes. So you probably want both options available (eventually). It was more just an FYI for when some cgroup or memory management test randomly fails because its arbitrary thresholds are exceeded

czerw:
> I think disabling and permissive mode serve different purposes. So you probably want both options available (eventually). It was more just an FYI for when some cgroup or memory management test randomly fails because its arbitrary thresholds are exceeded

I will keep already suggested logic and will add possibility fully disable SELinux as well.

czerw:
> I will keep already suggested logic and will add possibility fully disable SELinux as well.

It is extended, it is possible use LTP_SELINUX=disabled.

tests/kernel/boot_ltp.pm
```diff
@@ -41,6 +40,25 @@ sub run {
...
+    boot_system($self);
+
+    my $ltp_selinux = get_var('LTP_SELINUX');
+    if (defined($ltp_selinux)) {
+        if ($ltp_selinux eq 'disabled') {
+            assert_script_run("sed -i -e 's|^SELINUX=.*|SELINUX=disabled|g' /etc/selinux/config");
+            power_action('reboot', textmode => 1);
+            boot_system($self);
+        } else {
+            script_run("setenforce ${ltp_selinux}");
```
Avinesh, line 58: @czerw , Don't we need to reboot if we are changing from `disabled` to `enabled` ? or will we not encounter that scenario as by default it'll be enabled?

czerw: I didn't count with that scenario, as we don't have it by default, but i can add option to enable SELinux if, is disabled.

mdoucha:
I'd suggest adding a helper function that'll modify SELinux configuration. `boot_ltp` should call this function and then reboot if return value indicates that a reboot is needed.

tests/kernel/boot_ltp.pm
```diff
@@ -80,4 +98,11 @@ sub test_flags {
 
 See run_ltp.pm.
 
+=head2 LTP_SELINUX
```
coolgw, line 101: If new variables defined, should we add description into the variables.md instead of here?

tests/kernel/boot_ltp.pm
```diff
@@ -41,6 +40,25 @@ sub run {
...
+    my $ltp_selinux = get_var('LTP_SELINUX');
+    if (defined($ltp_selinux)) {
+        if ($ltp_selinux eq 'disabled') {
+            assert_script_run("sed -i -e 's|^SELINUX=.*|SELINUX=disabled|g' /etc/selinux/config");
+            power_action('reboot', textmode => 1);
+            boot_system($self);
+        } else {
+            script_run("setenforce ${ltp_selinux}");
+        }
+        record_info("SELinux", script_output('sestatus'));
```
coolgw, line 60: I suggest move this line out of "if", then we can see status without any condition.

tests/kernel/boot_ltp.pm
```diff
@@ -41,6 +40,25 @@ sub run {
     }
 
     select_serial_terminal;
+}
+
+sub run {
+    my ($self) = @_;
+    my $cmd_file = get_var('LTP_COMMAND_FILE') || '';
```
coolgw, line 47: What's usage for $cmd_file? I can not see any other place use this var in current file. Correct me if i am wrong.

czerw: If you look carefully, this is not related to this PR and it is used in `schedule_tests`.

tests/kernel/boot_ltp.pm
czerw: No, it would fail on old systems where is not SELinux at all and I'm not interested in them.

czerw: I will try to add it there in simple way, let's see if it is good solution.

tests/kernel/boot_ltp.pm
czerw: I decided to support disabling or setting permissive mode for now. We don't need to enable it for now and it would add more complexity.

PR #17595, opened 2023-08-17
Revert "containers: Add tests for volumes"

DimStar77:
Reverts os-autoinst/os-autoinst-distri-opensuse#17584

FWIW: as the test is also enabled in Staging, this blocks Staging / Tumbleweed from accepting new changes, thus blocking entirely.


pdostal (changes requested):
Is this really needed? @ricardobranco777 is online, so he may be already work on a fix.

ricardobranco777 (commented):
It's the `CONTAINER_RUNTIME` variable that is broken as it's set to `podman,docker`

DimStar77:
> It's the `CONTAINER_RUNTIME` variable that is broken as it's set to `podman,docker`

disagreed

https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/variables.md



CONTAINER_RUNTIME | string |   | Container runtime to be used, e.g.  docker, podman, or both podman,docker. In addition, it is also used for other container tests, like  kubectl, helm, etc.
-- | -- | -- | --

mloviska:
> It's the `CONTAINER_RUNTIME` variable that is broken as it's set to `podman,docker`

No, that is a standard for sle-micro,alp and others. It can be a list in fact and we have kinda missed this in the previous PR, sorry for that!

ricardobranco777 (changes requested):
I will submit a proper fix.  No need to revert.

ricardobranco777:
Fixing here:

https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17598

DimStar77:
> I will submit a proper fix. No need to revert.

With a fix incoming in reasonable timeframes, I am fully on board to not revert this.

PR #17607, opened 2023-08-18
Swap order of kernel parameters for new grub entries

czerw:
New custom grub entries added parameters to the beginning of kernel command line. This change will add them at the end of command line to allow overwriting (by order) existing parameters.

Alternative approach to  https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17593. New solution is more simple and more powerful and extends already existing functionality.

- Related ticket:  none
- Needles: none
- Verification run: 
 GRUB_PARAM=debug_pagealloc=on;ima_policy=tcb;selinux=0;enforcing=0
adds two new options to disable SELinux, one entry will disable completely, the other one will permissive mode on
installation: https://openqa.suse.de/tests/11897495#
boot (GRUB_BOOT_NONDEFAULT= 3): https://openqa.suse.de/tests/11897572#step/boot_ltp/5

Comparison of test result:
With SELinux: https://openqa.suse.de/tests/11849701#step/memcg_subgroup_charge/8
Without SELinux (GRUB_BOOT_NONDEFAULT= 3): https://openqa.suse.de/tests/11891059#step/memcg_subgroup_charge/8

### DIFF
--- a/lib/bootloader_setup.pm
+++ b/lib/bootloader_setup.pm
@@ -151,7 +151,7 @@ sub add_custom_grub_entries {
         my $script_new_esc = $script_new =~ s~/~\\/~rg;
         assert_script_run("cp -v $script_old $script_new");
 
-        my $cmd = "sed -i -e 's/\\(args=.\\)\\(\\\$4\\)/\\1$grub_param \\2/'";
+        my $cmd = "sed -i -e 's/\\(args=.\\)\\(\\\$4[^\"]*\\)/\\1\\2 $grub_param/'";
         $cmd .= " -e 's/\\(Advanced options for %s\\)/\\1 ($grub_param)/'";
         $cmd .= " -e 's/\\(menuentry .\\\$(echo .\\\$title\\)/\\1 ($grub_param)/'";
         $cmd .= " -e 's/\\(menuentry .\\\$(echo .\\\$os\\)/\\1 ($grub_param)/' $script_new";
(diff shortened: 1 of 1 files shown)


czerw:
@pevik @richiejp @mdoucha @Avinesh @coolgw   please check

mdoucha (approved):
LGTM

lib/bootloader_setup.pm
```diff
@@ -151,7 +151,7 @@ sub add_custom_grub_entries {
         my $script_new_esc = $script_new =~ s~/~\\/~rg;
         assert_script_run("cp -v $script_old $script_new");
 
-        my $cmd = "sed -i -e 's/\\(args=.\\)\\(\\\$4\\)/\\1$grub_param \\2/'";
+        my $cmd = "sed -i -e 's/\\(args=.\\)\\(\\\$4\\)/\\1\\2 $grub_param/'";
```
coolgw, line 154: I prefer following change then you have same order of GRUB_PARAM, otherwise you will get reverse orde.
my $cmd = "sed -i -e 's/\\(args=.\\)\\(\\\$4.*\[^\"\]\\)/\\1\\2 $grub_param/'";

czerw: Could you please show the example what it would fix and which reverse order do you mean?

We are changing the line  `args="$4"`  to ` args="$4 debug_pagealloc=on"` and without this PR line would be  `args="debug_pagealloc=on $4"`

original: `args="$4"` 
old behavior:  `args="debug_pagealloc=on $4"`
new behavior: `args="$4 debug_pagealloc=on"`

mdoucha: Imagine the original string looks like this: `args="foo $4 bar"`. If `\2` doesn't include everything up to the closing quotation mark, you'll get `args="foo $4 debug_pagealloc=on bar"`

acerv:
Nice idea, LGTM

richiejp:
One potential issue with overriding parameters is the following

> The number of kernel parameters is not limited, but the length of the
> complete command line (parameters including spaces etc.) is limited to
> a fixed number of characters. This limit depends on the architecture
> and is between 256 and 4096 characters

It seems to be set to 2048 on the platforms we test on except s390 which is set in the config. I'm not sure what happens when the limit is exceeded. I suppose we are not likely to hit it unless registration and connection details are passed to systemd.

Possibly we should check the string size or make a note of the limit?

czerw:
> Possibly we should check the string size or make a note of the limit?
It is non existing issue so far and this PR is just swapping order of kernel cmd line parameters. Created follow up https://progress.opensuse.org/issues/134486.

lib/bootloader_setup.pm
czerw: Latest VR: https://openqa.suse.de/tests/11897495#step/install_ltp/184

PR #17608, opened 2023-08-18
Unschedule rootless podman test case on <=15-SP2

mloviska:
Rootless podman test case is failing with unexpected list of capablities when container is spawned with `keep-id` option on sle15-sp2. Tested with multiple kernel versions and the podman packages are same for a long time. Further investigation is needed but we should not block updates

- Verification run: http://kepler.suse.cz/tests/21542#


grisu48 (approved):

grisu48:
Please rebase and then feel free to merge.

ricardobranco777 (approved):

PR #17609, opened 2023-08-21
usr_sbin_smbd: use 'assert_and_click' instead of 'send_key'

zhaocongjia:
It is faster and more accurate to use `assert_and_click` to focus on the address bar.

- Related ticket: https://progress.opensuse.org/issues/134330
- Needles: None
- Verification run: https://openqa.opensuse.org/tests/3515713#step/usr_sbin_smbd/61


GraceWang571:
LGTM

PR #17610, opened 2023-08-21
Add new workaround for nvidia repo not reliable with new hot-key

rfan1:
https://progress.opensuse.org/issues/134423
https://bugzilla.suse.com/show_bug.cgi?id=1214234

- Verification run: 
[SLES](https://openqa.suse.de/tests/11888696)
[SLED](https://openqa.suse.de/tests/11888689)

Based on the test result, the workaround can not work perfectly on sled tests, we may need to file a new ticket.

### DIFF
--- a/lib/registration.pm
+++ b/lib/registration.pm
@@ -544,8 +544,9 @@ sub process_scc_register_addons {
             }
             elsif (match_has_tag('nvidia-validation-failed')) {
                 # nvidia repos unreliable
-                send_key 'alt-y';
-                record_soft_failure 'bsc#1144831';
+                record_soft_failure 'bsc#1214234';
+                wait_still_screen { send_key 'alt-o' };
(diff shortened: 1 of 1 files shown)


foursixnine (approved):
This may result in this failing in older sp's if they are still using old installer, but I doubt that they will face the problem as the nvidia repos are already present for released products

rfan1:
> This may result in this failing in older sp's if they are still using old installer, but I doubt that they will face the problem as the nvidia repos are already present for released products

Thanks, and you are correct, I noticed that `autoyast` uses the same workaround

`autoyast/installation.pm`

            # nvidia repositories are unstable and really not needed for anything
            record_info("NVIDIA", "NVIDIA repository is broken");
            wait_still_screen { send_key 'alt-o' };
            send_key 'alt-n';
            $num_errors++;
        }

PR #17622, opened 2023-08-22
Add workaround for verify correct capabilities

mloviska:
A container spawned with `keep-id` option lists capabilities. The original workaround has disappeared, however the issue is still tracked in the same report - https://bugzilla.suse.com/show_bug.cgi?id=1182428

- [sle15-sp2 + podman v3.4.7](http://kepler.suse.cz/tests/21561#step/rootless_podman/214)
- [sle15-sp2 + podman v2.1.1](http://kepler.suse.cz/tests/21565#)


grisu48 (approved):
LGTM! http://kepler.suse.cz/tests/21561#step/rootless_podman/214 fails, but I guess that's unrelated

mloviska:
> LGTM! http://kepler.suse.cz/tests/21561#step/rootless_podman/214 fails, but I guess that's unrelated

As of now, unrelated as I had to add manually some old podman 3.4.7 update for 15-SP2 LTSS update. It is just a demo that the root-cause is simply podman 2.1.1

